PR #1261, opened 2017-02-28
Fix pushing down wrong queries for INSERT ... SELECT queries

onderkalaci:
(This PR should be back-ported to both 6.0 and 6.1)

Fixes #1241, and since a better approach than #1243 I'm closing it.

Before this commit, in certain cases router planner allowed pushing
down JOINs that are not on the partition keys.

With @anarazel's suggestion, we change the logic to use uninstantiated
parameter. Previously, the planner was traversing on the restriction
information and once it finds the parameter, it was replacing it with
the shard range. With this commit, instead of traversing the restrict
infos, the planner explicitly checks for the equivalence of the relation
partition key with the uninstantiated parameter. If finds an equivalence,
it adds the restrictions. In this way, we have more control over the
queries that are pushed down.

PR #1268, opened 2017-03-01
Replace adding the qual logic with attribute equivalence for INSERT ... SELECT

onderkalaci:
In this PR, we aim to deduce whether each of the RTE_RELATION
is joined with at least on another RTE_RELATION on their partition keys. If each
RTE_RELATION follows the above rule, we can conclude that all RTE_RELATIONs are
joined on their partition keys.

In order to do that, we invented a new equivalence class namely:
AttributeEquivalenceClass. In very simple words, a AttributeEquivalenceClass is
identified by an unique id and consists of a list of AttributeEquivalenceMembers.

Each AttributeEquivalenceMember is designed to identify attributes uniquely within the
whole query. The necessity of this arise since varno attributes are defined within
a single level of a query. Instead, here we want to identify each RTE_RELATION uniquely
and try to find equality among each RTE_RELATION's partition key.

Whenever we find an equality clause A = B, where both A and B originates from
relation attributes (i.e., not random expressions, consts and such), we create an
AttributeEquivalenceClass to record this knowledge. If we later find another
equivalence B = C, we create another AttributeEquivalenceClass. Finally, we can
apply transitity rules and generate a new AttributeEquivalenceClass which includes
A, B and C.

Note that equality among the members are identified by the varattno and rteIdentity.

Each equality among RTE_RELATION is saved using an AttributeEquivalenceClass where
each member attribute is identified by a AttributeEquivalenceMember. In the final
step, we try generate a common attribute equivalence class that holds as much as
AttributeEquivalenceMembers whose attributes are a partition keys.

(Note that this PR fails to push down `UNIONS` correctly, but we'll handle them separately in the upcoming PRs)


anarazel:
Hi Onder,

It's a bit hard to say in the current state - will play a bit tomorrow.  But I suspect quite strongly that atm this isn't correct - I think we're going to do incorrect things with outer/anti/semijoins - there's often a reason quals haven't pushed into baserestrictinfos (that obviously is only possible in the first place for Var = Const type restrictions, but my point holds).  I'm also quite doubtful that it's "legal" to just push a qual like you're doing in AddToVarEquivalenceClass - what if the RTE_RELATION rte is inside, e.g., an anti join?

could you maybe spent an hour or two today to clean this up before I have a look?

- Andres

onderkalaci:
@anarazel Note that this PR is on top of #1261 (there are some extra tests on it, that's why I used that branch as the base).

> w. But I suspect quite strongly that atm this isn't correct - I think we're going to do incorrect things with outer/anti/semijoins - there's often a reason quals haven't pushed into baserestrictinfos (that obviously is only possible in the first place for Var = Const type restrictions, but my point holds)

I was already working to improve that. The updated PR uses `root->eq_classes` for inner joins, and uses join restriction information (via `set_join_pathlist_hook`) for other joins. The PR skips anti-joins at all and processes outer joins via the generated restriction information. (Using two different things bugs me, hoping to unify the logic soon.)

>  I'm also quite doubtful that it's "legal" to just push a qual like you're doing in AddToVarEquivalenceClass - what if the RTE_RELATION rte is inside, e.g., an anti join?

Your comment makes sense in general. However, if there are any anti-joins, the code prevents pushing it down by skipping the evaluation of the restrictions generated for the RTEs in the anti-join. Added some tests for that.

I couldn't find any other way of building a relation among the different levels of subqueries via equivalence classes. Is there a better approach to do that?

One final note here is that, I haven't considered set operations since `INSERT ... SELECT ` disallows set operations at all. Do you see any problems with implementing set operations?

(There are some issues with reference tables, two tests fail. But, they are minor issues, should be easy to fix.)

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -325,6 +356,459 @@ CreateInsertSelectRouterPlan(Query *originalQuery,
...
+	bool allRelationsJoinedOnPartitionKey = true;
+	uint32 referenceRelationCount = ReferenceRelationCount(restrictionContext);
+	uint32 totalRelationCount = list_length(restrictionContext->relationRestrictionList);
+
+	ListCell *commonEqClassCell = NULL;
+	ListCell *restrictionCell = NULL;
+	Relids commonRelids = NULL;
+
+	/*
+	 * If there is a single RTE_RELATION that is not a reference table,
```
mtuncer: this means there are no joins between non-reference tables. Comment indirectly says that however it requires some thinking to understand.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -73,6 +76,8 @@ typedef struct WalkerState
 } WalkerState;
 
 bool EnableRouterExecution = true;
+static uint32 varEquivalenceId = 1;
```
mtuncer: Eventually this would overflow. Should we reset this when we start planning a query ?

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -325,6 +356,459 @@ CreateInsertSelectRouterPlan(Query *originalQuery,
...
+						 VarEquivalenceClass **varEquivalanceClass)
+{
+	RangeTblEntry *rte = root->simple_rte_array[varToBeAdded->varno];
+
+	if (rte->rtekind == RTE_RELATION)
+	{
+		VarEquivalenceClassMember *varEqMember = palloc0(
+			sizeof(VarEquivalenceClassMember));
+		varEqMember->varattno = varToBeAdded->varattno;
+		varEqMember->varno = varToBeAdded->varno;
```
mtuncer: varno is meaningless outside the Query it was in.  Need to check how it used afterwards.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -325,6 +356,459 @@ CreateInsertSelectRouterPlan(Query *originalQuery,
...
+		{
+			RestrictInfo *rinfo = (RestrictInfo *) lfirst(restrictionInfoList);
+			OpExpr *restrictionOpExpr = NULL;
+			Expr *leftExpr = NULL;
+			Expr *rightExpr = NULL;
+			Var *leftVar = NULL;
+			Var *rightVar = NULL;
+			Expr *restrictionClause = rinfo->clause;
+
+			if (!IsA(restrictionClause, OpExpr))
```
mtuncer: how cases like
```
ON (r1.a = r1.b and false)
ON (r1.a =  r2.b or r1.c = NULL)
```
handled here.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -325,6 +356,459 @@ CreateInsertSelectRouterPlan(Query *originalQuery,
...
+ */
+static bool
+VarClassMemberEqualsToVarClass(VarEquivalenceClassMember *inputMember,
+							   VarEquivalenceClass *varEqClass)
+{
+	ListCell *classCell = NULL;
+	foreach(classCell, varEqClass->equivalentVars)
+	{
+		VarEquivalenceClassMember *memberOfClass = lfirst(classCell);
+		if (memberOfClass->varno == inputMember->varno &&
```
mtuncer: comparing varno's could be problematic here. Same relationid used in 2 different subqueries and at the same location in rtable list would have the same varno. Does it make a difference in the correctness of this function ?

src/backend/distributed/planner/multi_router_planner.c
onderkalaci: > ON (r1.a = r1.b and false)

For this one, Postgres understands that the join wouldn't yield and rows thus doesn't generate any join  restrictions. In the later planning, both relations' restriction information includes `false` and thus all shards pruned.

Example:
```SQL

INSERT INTO agg_events                                                                                                                                                                                                                                      (value_1_agg,                                                                                                                                                                                                                                               user_id)                                                                                                                                                                                                                                      SELECT SUM(value_1),                                                                                                                                                                                                                                              id                                                                                                                                                                                                                                                  FROM   (SELECT raw_events_second.user_id AS id,                                                                                                                                                                                                                           raw_events_second.value_1                                                                                                                                                                                                                           FROM   raw_events_first,                                                                                                                                                                                                                                          raw_events_second                                                                                                                                                                                                                                   WHERE  raw_events_first.value_1 = 15 AND raw_events_first.user_id = raw_events_second.user_id  and false) AS foo                                                                                                                                   GROUP  BY id;
DEBUG:  StartTransactionCommand
DEBUG:  StartTransaction
DEBUG:  name: unnamed; blockState:       DEFAULT; state: INPROGR, xid/subid/cid: 0/1/0, nestlvl: 1, children: 
DEBUG:  Skipping target shard interval 102072 since SELECT query for it pruned away
DEBUG:  Skipping target shard interval 102073 since SELECT query for it pruned away
DEBUG:  Skipping target shard interval 102074 since SELECT query for it pruned away
DEBUG:  Skipping target shard interval 102075 since SELECT query for it pruned away
DEBUG:  Skipping target shard interval 102076 since SELECT query for it pruned away
DEBUG:  Skipping target shard interval 102077 since SELECT query for it pruned away
DEBUG:  Skipping target shard interval 102078 since SELECT query for it pruned away
DEBUG:  Skipping target shard interval 102079 since SELECT query for it pruned away
DEBUG:  Skipping target shard interval 102080 since SELECT query for it pruned away
DEBUG:  Skipping target shard interval 102081 since SELECT query for it pruned away
DEBUG:  Skipping target shard interval 102082 since SELECT query for it pruned away
DEBUG:  Skipping target shard interval 102083 since SELECT query for it pruned away
DEBUG:  Skipping target shard interval 102084 since SELECT query for it pruned away
DEBUG:  Skipping target shard interval 102085 since SELECT query for it pruned away
DEBUG:  Skipping target shard interval 102086 since SELECT query for it pruned away
DEBUG:  Skipping target shard interval 102087 since SELECT query for it pruned away
DEBUG:  Skipping target shard interval 102088 since SELECT query for it pruned away
DEBUG:  Skipping target shard interval 102089 since SELECT query for it pruned away
DEBUG:  Skipping target shard interval 102090 since SELECT query for it pruned away
DEBUG:  Skipping target shard interval 102091 since SELECT query for it pruned away
DEBUG:  Skipping target shard interval 102092 since SELECT query for it pruned away
DEBUG:  Skipping target shard interval 102093 since SELECT query for it pruned away
DEBUG:  Skipping target shard interval 102094 since SELECT query for it pruned away
DEBUG:  Skipping target shard interval 102095 since SELECT query for it pruned away
DEBUG:  Skipping target shard interval 102096 since SELECT query for it pruned away
DEBUG:  Skipping target shard interval 102097 since SELECT query for it pruned away
DEBUG:  Skipping target shard interval 102098 since SELECT query for it pruned away
DEBUG:  Skipping target shard interval 102099 since SELECT query for it pruned away
DEBUG:  Skipping target shard interval 102100 since SELECT query for it pruned away
DEBUG:  Skipping target shard interval 102101 since SELECT query for it pruned away
DEBUG:  Skipping target shard interval 102102 since SELECT query for it pruned away
DEBUG:  Skipping target shard interval 102103 since SELECT query for it pruned away
DEBUG:  ProcessQuery
DEBUG:  CommitTransactionCommand
DEBUG:  CommitTransaction
DEBUG:  name: unnamed; blockState:       STARTED; state: INPROGR, xid/subid/cid: 0/1/0, nestlvl: 1, children: 
INSERT 0 0
```

> ON (r1.a =  r2.b or r1.c = NULL)

For this one, good catch I haven't considered this at all. 

The restriction is generated to be `BoolExpr` where the logic here expects `OpExpr` and thus not able to push the query. 

But, as far as I can see, we shouldn't allow OR clauses. So, the behavior seems OK to me. What do you think?

src/backend/distributed/planner/multi_router_planner.c
onderkalaci: As far as I can see, per appearance of a relation, Postgres generates an RTE and assigns a unique `RTEIndex`. 
Vars of those relations are assigned the unique rteIndex.
(Postgres also generates RTEs for subqueries etc. but that is not in the scope of this discussion.)
 
For example:

```SQL
INSERT INTO agg_events
            (user_id, value_4_agg)
SELECT
  outer_most.id, max(outer_most.value)
FROM
(
  SELECT f2.id as id, f2.v4 as value FROM
    (SELECT
          id
      FROM   (SELECT reference_table.user_id      AS id
               FROM   raw_events_first,
                      reference_table
            WHERE  raw_events_first.user_id = reference_table.user_id ) AS foo) as f
  INNER JOIN
    (SELECT v4,
          v1,
          id
    FROM   (SELECT SUM(raw_events_second.value_4) AS v4,
               SUM(raw_events_first.value_1) AS v1,
               raw_events_second.user_id      AS id
            FROM   raw_events_first,
                    raw_events_second
            WHERE  raw_events_first.user_id = raw_events_second.user_id
            GROUP  BY raw_events_second.user_id
            HAVING SUM(raw_events_second.value_4) > 10) AS foo2 ) as f2
ON (f.id = f2.id)) as outer_most
GROUP BY
  outer_most.id;
```

We got the following vars associated with relations (i.e.., the member of common var eq class for this query)

```
INFO:  Relation Name: raw_events_first Var No:1
INFO:  Relation Name: raw_events_second Var No:2
INFO:  Relation Name: raw_events_first Var No:6
```

> varno is meaningless outside the Query it was in. Need to check how it used afterwards.

Do you still think in that way?

src/backend/distributed/planner/multi_router_planner.c
onderkalaci: > comparing varno's could be problematic here. Same relationid used in 2 different subqueries and at the same location in rtable list would have the same varno. Does it make a difference in the correctness of this function ?

Please see my comment above. As far as I can see, each range table gets a unique RTEIndex.

src/include/distributed/multi_planner.h
```diff
@@ -56,6 +84,18 @@ struct MultiPlan;
 extern struct MultiPlan * GetMultiPlan(PlannedStmt *planStatement);
 extern void multi_relation_restriction_hook(PlannerInfo *root, RelOptInfo *relOptInfo,
 											Index index, RangeTblEntry *rte);
+extern void multi_special_join_restriction_hook(PlannerInfo *root,
```
onderkalaci: Ops, I was trying some other stuff and forgot this here 😄

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -325,6 +358,522 @@ CreateInsertSelectRouterPlan(Query *originalQuery,
...
+ */
+static VarEquivalenceClass *
+GenerateCommonPartitionKeyEquivalence(List *varEquivalences)
+{
+	VarEquivalenceClass *mainEquivalenceClass = NULL;
+	Bitmapset *addedEquivalenceIds = NULL;
+	uint32 eqClassSize = list_length(varEquivalences);
+	uint32 equivalenceClassIndex = 0;
+
+	mainEquivalenceClass = palloc0(sizeof(VarEquivalenceClass));
```
onderkalaci: `mainEquivalenceClass`to `commonEquivalenceClass`

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -325,6 +358,522 @@ CreateInsertSelectRouterPlan(Query *originalQuery,
...
+static Var *
+GetVarFromAssignedParam(List *parentPlannerParamList, Param *plannerParam)
+{
+	Var *assignedVar = NULL;
+	ListCell *plannerParameterCell = NULL;
+
+	Assert(plannerParam != NULL);
+
+	/* we're only interested in parameters that Postgres added for execution */
+	if (plannerParam->paramkind != PARAM_EXEC)
```
onderkalaci: Add regression tests with parametrized prepared statements. (w/wout `PARAM_EXEC` in the query)

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -325,6 +358,522 @@ CreateInsertSelectRouterPlan(Query *originalQuery,
...
+
+		if (!IsA(subqueryTargetEntry->expr, Var))
+		{
+			return;
+		}
+
+		varToBeAdded = (Var *) subqueryTargetEntry->expr;
+
+		/* Can only handle a simple Var of subquery's query level */
+		if (varToBeAdded && IsA(varToBeAdded, Var) && varToBeAdded->varlevelsup == 0)
```
onderkalaci: Can we come-up with regression tests where `varlevelsup  != 0`?

src/backend/distributed/planner/multi_router_planner.c
onderkalaci: Ok, what @mtuncer suggests makes sense. We've switched to `RTEIdentity` to overcome this problem.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -339,6 +372,549 @@ CreateInsertSelectRouterPlan(Query *originalQuery,
 
 
 /*
+ * AllRelationsJoinedOnPartitionKey expands the var equivalances gathered from
```
onderkalaci: @anarazel This function and this approach cannot pushdown queries that don't involve any joins but has more than 1 relations, namely set operations:

```SQL
INSERT INTO agg_events 
            (user_id) 
SELECT user_id 
FROM   ((SELECT user_id 
         FROM   raw_events_first) 
        UNION 
        (SELECT user_id 
         FROM   raw_events_first)) AS t;   
```

Do you see any viable approach to handle such queries?

Note that the following query can be pushed down since it involves join:

```SQL
INSERT INTO agg_events 
            (user_id) 
SELECT t.user_id 
FROM   ((SELECT user_id 
         FROM   raw_events_first) 
        UNION 
        (SELECT user_id 
         FROM   raw_events_first)) AS t JOIN raw_events_second ON t.user_id = raw_events_second.user_id; 
```

onderkalaci:
@anarazel There is only one case, where this approach fails to push down the query. I think, the same thing applies to the other approach you've shared privately with us:

```SQL
(SELECT user_id FROM users)
UNION
(SELECT user_id FROM users);
```

Basically, there is not equivalence among any relations in the query. I think we need to write some other code to detect these kind of queries given that they might be useful (and the existing subquery pushdown logic already supports that).

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -341,6 +381,706 @@ CreateInsertSelectRouterPlan(Query *originalQuery,
...
+		AttributeEquivalenceClass *currentEquivalenceClass =
+			list_nth(attributeEquivalenceList, equivalenceClassIndex);
+		ListCell *equivalenceMemberCell = NULL;
+
+		/*
+		 * This is an optimization. If we already added the same equivalence class,
+		 * we could skip it since we've already added all the relevant equivalence
+		 * members.
+		 */
+		if (bms_overlap(addedEquivalenceIds,
```
anarazel: wouldn't bms_is_member be better here?

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -341,6 +381,706 @@ CreateInsertSelectRouterPlan(Query *originalQuery,
...
+ * The algorithm followed can be summarized as below:
+ *
+ * - Per join restriction
+ *     - Per RestrictInfo of the join restriction
+ *     - Check whether the join restriction is in the form of (Var1 = Var2)
+ *         - Create an AttributeEquivalenceClass
+ *         - Add both Var1 and Var2 to the AttributeEquivalenceClass
+ *
+ * Note that this function does not deal with whether the member of restriction
+ * are partition key or not. That's handled later in the planning.
```
anarazel: Why?

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -341,6 +381,706 @@ CreateInsertSelectRouterPlan(Query *originalQuery,
...
+				continue;
+			}
+
+			leftVar = (Var *) strippedLeftExpr;
+			rightVar = (Var *) strippedRightExpr;
+
+			attributeEquivalance = palloc0(sizeof(AttributeEquivalenceClass));
+			attributeEquivalance->equivalenceId = attributeEquivalenceId++;
+
+			/* we could (probably) safely do that since we disallowed ANTI JOINs ? */
```
anarazel: Where?

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -341,6 +381,706 @@ CreateInsertSelectRouterPlan(Query *originalQuery,
...
+	ListCell *joinRestrictionCell = NULL;
+
+	foreach(joinRestrictionCell, joinRestrictionContext->joinRestrictionList)
+	{
+		JoinRestriction *joinRestriction = lfirst(joinRestrictionCell);
+		ListCell *restrictionInfoList = NULL;
+
+		foreach(restrictionInfoList, joinRestriction->joinRestrictInfoList)
+		{
+			RestrictInfo *rinfo = (RestrictInfo *) lfirst(restrictionInfoList);
```
anarazel: This is O(nrels^2), no?  With some not that small constants.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -341,6 +381,706 @@ CreateInsertSelectRouterPlan(Query *originalQuery,
...
+			{
+				continue;
+			}
+
+			leftNode = linitial(restrictionOpExpr->args);
+			rightNode = lsecond(restrictionOpExpr->args);
+
+			/* we also don't want implicit coercions */
+			strippedLeftExpr = (Expr *) strip_implicit_coercions((Node *) leftNode);
+			strippedRightExpr = (Expr *) strip_implicit_coercions((Node *) rightNode);
```
anarazel: Are there actually implicit coercions at this point?

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -341,6 +381,706 @@ CreateInsertSelectRouterPlan(Query *originalQuery,
...
+static bool
+AttributeClassContainsAttributeClassMember(AttributeEquivalenceClassMember *inputMember,
+										   AttributeEquivalenceClass *
+										   attributeEquivalenceClass)
+{
+	ListCell *classCell = NULL;
+	foreach(classCell, attributeEquivalenceClass->equivalentAttributes)
+	{
+		AttributeEquivalenceClassMember *memberOfClass = lfirst(classCell);
+		if (memberOfClass->rteIdendity == inputMember->rteIdendity &&
```
anarazel: This is O(N), and you're doing it for every attribute member - O(N^2)...

src/backend/distributed/planner/multi_router_planner.c
onderkalaci: This is an old comment and removed now. We currently do not disallow anti joins. I don't see a reason to disallow anti-joins. Do you see?

src/backend/distributed/planner/multi_router_planner.c
onderkalaci: Yes, when I create two tables that are partitioned on varchar columns, I can see that we get the following here:
```
{RELABELTYPE :arg {VAR :varno 3 :varattno 1 :vartype 1043 :vartypmod -1 :varcollid 100 :varlevelsup 0 :varnoold 3 :varoattno 1 :location 49} :resulttype 25 :resulttypmod -1 :resultcollid 100 :relabelformat 2 :location -1}
```

src/backend/distributed/planner/multi_router_planner.c
onderkalaci: Yes, I agree that there are few places where the complexity is O(n^2) on the attribute number. But, now that I switched to adding only partition keys to the attribute equivalence, I started to think that it is OK given that we'd have very few attribute members.

Do you think should we find a better way of processing the equivalences, or are you OK with it?

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -394,43 +410,19 @@ RouterModifyTaskForShardInterval(Query *originalQuery, ShardInterval *shardInter
...
-			RestrictInfo *geRestrictInfo = NULL;
-			RestrictInfo *leRestrictInfo = NULL;
-
-			OpExpr *hashedGEOpExpr = (OpExpr *) linitial(hashedOpExpressions);
-			OpExpr *hashedLEOpExpr = (OpExpr *) lsecond(hashedOpExpressions);
+		hashedOpExpressions = ShardIntervalOpExpressions(shardInterval, rteIndex);
 
-			geRestrictInfo = make_simple_restrictinfo((Expr *) hashedGEOpExpr);
-			originalBaserestrictInfo = lappend(originalBaserestrictInfo, geRestrictInfo);
+		hashedRestrictInfo = make_simple_restrictinfo((Expr *) hashedOpExpressions);
```
mtuncer: this is no longer for hashed operator expression right ?

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -924,13 +877,18 @@ MultiTaskRouterSelectQuerySupported(Query *query)
...
 		if (subquery->setOperations != NULL)
 		{
-			return DeferredError(ERRCODE_FEATURE_NOT_SUPPORTED,
-								 "set operations are not allowed in INSERT ... SELECT "
-								 "queries",
-								 NULL, NULL);
+			SetOperationStmt *setOperationStatement =
+				(SetOperationStmt *) subquery->setOperations;
+
+			if (setOperationStatement->op != SETOP_UNION)
```
mtuncer: You should revisit this when enabling UNIONS as there could me more than one set operation such ash  a U (b - c).

mtuncer: On an irrelevant subject function name ```MultiTaskRouterSelectQuerySupported``` did not mean anything to me.  This was only called from ```InsertSelectQuerySupported``` yet has the name MultiTask...

src/backend/distributed/planner/multi_planner.c
```diff
@@ -621,23 +708,65 @@ multi_relation_restriction_hook(PlannerInfo *root, RelOptInfo *relOptInfo, Index
 }
 
 
+/* GetRTEIdentity returns the identity assigned with AssignRTEIdentity. */
```
onderkalaci: done

src/backend/distributed/planner/multi_router_planner.c
onderkalaci: In this part of the code (i.e., `INSERT .. SELECT `), it is still hashed restrict info.

src/backend/distributed/planner/multi_router_planner.c
onderkalaci: I'll remove support for UNIONs in this PR

src/backend/distributed/planner/multi_planner.c
```diff
@@ -26,11 +26,14 @@
...
 #include "nodes/makefuncs.h"
 #include "nodes/nodeFuncs.h"
+#include "parser/parsetree.h"
+#include "optimizer/pathnode.h"
 #include "optimizer/planner.h"
 #include "utils/memutils.h"
 
 
 static List *relationRestrictionContextList = NIL;
+static List *joinRestrictionContextList = NIL;
```
anarazel: Maybe I'm missing something - but those are equivalent?

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -341,6 +381,706 @@ CreateInsertSelectRouterPlan(Query *originalQuery,
...
+ * AttributeEquivalenceMembers whose attributes are a partition keys.
+ *
+ * AllRelationsJoinedOnPartitionKey uses both relation restrictions and join restrictions
+ * to find as much as information that Postgres planner provides to extensions. For the
+ * details of the usage, please see GenerateAttributeEquivalencesForRelationRestrictions()
+ * and GenerateAttributeEquivalencesForJoinRestrictions()
+ *
+ * Finally, as the name of the function reveals, the function returns true if all relations
+ * are joined on their partition keys. Otherwise, the function returns false.
+ */
```
anarazel: >  Hmm, single shard is also not very accurate

How not?  `Relations` is plural, and each of those prunes down to a single shard?  This doesn't actually tests node restrictions, so that'd not be any better?

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -341,6 +381,706 @@ CreateInsertSelectRouterPlan(Query *originalQuery,
...
+													 *restrictionContext)
+{
+	List *attributeEquivalenceList = NIL;
+	ListCell *relationRestrictionCell = NULL;
+
+	foreach(relationRestrictionCell, restrictionContext->relationRestrictionList)
+	{
+		RelationRestriction *relationRestriction = lfirst(relationRestrictionCell);
+		List *equivalenceClasses = relationRestriction->plannerInfo->eq_classes;
+		ListCell *equivilanceClassCell = NULL;
```
anarazel: >  > Doesn't this mean we're duplicating a lot of eq classes

> Yes, that's true. I can add some logic to prevent adding the same attributed eq class to the final output list. However, we'd be still processing the same eq_classes.

Yea, but we'd not duplicating things repeatedly.



    As I said, I think we should move to only collecting plannerInfos.

> But, wouldn't we still need to process each plannerInfo that is associated with a relation? (Or, do we skip the plannerInfos that are equal to each other?)

Yes, but I'm not seeing the problem?

src/backend/distributed/planner/multi_colocated_subquery_planner.c
```diff
@@ -0,0 +1,783 @@
+/*
+ * multi_colocated_subquery_planner.c
```
anarazel: I'd be inclined to go for shard_pruning.c or such - that's what I'd added in my faster shard pruning stuff, and this pretty much related.

src/backend/distributed/planner/multi_planner.c
onderkalaci: Not really. `root->eq_classes` contains two separate `EQUIVALENCECLASS` es, which means that they are not known to be equal by postgres (and so we don't treat them being equal). 

However `JoinPathExtraData->restrictlist` contains an equality `OpExpr` among the corresponding `Vars`. 

That seems kind of weird to me as well. But, I couldn't figure out why this is happening. Do you have any ideas?

src/backend/distributed/planner/multi_router_planner.c
onderkalaci: > This doesn't actually tests node restrictions, so that'd not be any better?

Yes, this function is not doing any shard pruning or adding restrictions. We use this function as a helper function to decide whether to add or not to add the restrictions.

Some other suggestions:

* `SafeToPushDownSubquery()` (actually, in the `Union`s PR, I've switched to this name)
* `AddShardIntervalRestrictionsToRelations()`
* `AddShardIntervalRestrictions()`

Any more suggestions?

src/backend/distributed/planner/multi_router_planner.c
onderkalaci: > Yea, but we'd not duplicating things repeatedly.

I changed the logic to skip duplicating the elements over and over.

> Yes, but I'm not seeing the problem?

That's not a problem, just wanted to note it.

src/backend/distributed/planner/multi_colocated_subquery_planner.c
onderkalaci: For the time being, we changed to `multi_planner_utils.c`. Does that work for you?

src/backend/distributed/planner/multi_planner.c
```diff
@@ -125,24 +113,87 @@ multi_planner(Query *parse, int cursorOptions, ParamListInfo boundParams)
...
 	/* remove the context from the context list */
-	PopRestrictionContext();
+	PopPlannerRestrictionContext();
 
 	return result;
 }
 
 
 /*
+ * To be able to track individual RTEs through postgres' query
```
mtuncer: This comment and the next one do not follow function header comment style

src/backend/distributed/planner/multi_planner.c
```diff
@@ -622,51 +717,109 @@ multi_relation_restriction_hook(PlannerInfo *root, RelOptInfo *relOptInfo, Index
...
+	plannerRestrictionContextList = lcons(plannerRestrictionContext,
+										  plannerRestrictionContextList);
 
-	return restrictionContext;
+	return plannerRestrictionContext;
 }
 
 
 /*
  * CurrentRestrictionContext returns the the last restriction context from the
```
mtuncer: Function name changed

src/backend/distributed/planner/multi_planner.c
```diff
@@ -622,51 +717,109 @@ multi_relation_restriction_hook(PlannerInfo *root, RelOptInfo *relOptInfo, Index
...
+
+		copiedPlanParamList = lappend(copiedPlanParamList, copiedParamItem);
+	}
+
+	return copiedPlanParamList;
+}
+
+
+/*
+ * CreateAndPushPlannerRestrictionContext creates a new relation restriction context
```
mtuncer: function definition changed. It now creates planner restriction context

src/backend/distributed/planner/multi_planner.c
```diff
@@ -622,51 +717,109 @@ multi_relation_restriction_hook(PlannerInfo *root, RelOptInfo *relOptInfo, Index
...
 
-	return restrictionContext;
+	return relationRestrictionContext;
 }
 
 
 /*
- * PopRestrictionContext removes the most recently added restriction context from
- * context list. The function assumes the list is not empty.
+ * CurrentRestrictionContext returns the the last restriction context from the
```
mtuncer: CurrentJoinRestrictionContext

src/backend/distributed/planner/multi_planner.c
onderkalaci: Fixed

src/backend/distributed/planner/multi_planner.c
onderkalaci: Fixed

src/backend/distributed/planner/multi_planner.c
onderkalaci: Hazards of copy & pasting 😄

anarazel:
First off: The title is a bit confusing: "Remove Adding" ;)

src/backend/distributed/planner/multi_planner.c
```diff
@@ -621,23 +708,65 @@ multi_relation_restriction_hook(PlannerInfo *root, RelOptInfo *relOptInfo, Index
...
- * beginning of the context list, and returns the newly created context.
+ * CopyPlanParamList deep copies the input PlannerParamItem list and returns the newly
+ * allocated list.
+ * Note that we cannot use copyObject() function directly since there is no support for
+ * copying PlannerParamItem structs.
  */
-static RelationRestrictionContext *
-CreateAndPushRestrictionContext(void)
+static List *
+CopyPlanParamList(List *originalPlanParamList)
```
anarazel: I realize we need to replace params, but why do we need to do so in-place, rather than just in a clause passed to pruning?

src/backend/distributed/planner/multi_planner_utils.c
```diff
@@ -0,0 +1,863 @@
+/*
+ * multi_planner_utils.c
```
anarazel: Still disagree with the file name - sounds like minor helpers, rather than some fundamental infrastructure.

src/backend/distributed/planner/multi_planner_utils.c
```diff
@@ -0,0 +1,863 @@
...
+ *
+ * This file contains functions helper functions for planning
+ * queries with colocated tables and subqueries.
+ *
+ * Copyright (c) 2017-2017, Citus Data, Inc.
+ *
+ *-------------------------------------------------------------------------
+ */
+#include "postgres.h"
+#include "c.h"
```
anarazel: Don't ever include c.h yourself when postgres.h is also included. Yes, we do that in some places (where I didn't pay attention), but that shouldn't spread around.

src/backend/distributed/planner/multi_planner_utils.c
```diff
@@ -0,0 +1,863 @@
...
+	uint32 nonReferenceRelationCount = totalRelationCount - referenceRelationCount;
+	ListCell *commonEqClassCell = NULL;
+	ListCell *relationRestrictionCell = NULL;
+	Relids commonRteIdentities = NULL;
+
+	/*
+	 * If the query includes a single relation which is not a reference table,
+	 * we should not check the partition column equality.
+	 * Consider two example cases:
+	 *   (i)   The query includes only a single colocated relation
```
anarazel: That's why I'm suggested modelling this as "prunes down to single relation".

src/backend/distributed/planner/multi_planner_utils.c
```diff
@@ -0,0 +1,863 @@
...
+	 *   (ii)  A colocated relation is joined with a reference
+	 *         table where colocated relation is not joined on the partition key
```
anarazel: But what if there's more than one non-reference table involved?  We need to be safe against this anyway.

src/backend/distributed/planner/multi_planner_utils.c
```diff
@@ -0,0 +1,863 @@
...
+				(EquivalenceClass *) lfirst(equivalenceClassCell);
+
+			AttributeEquivalenceClass *attributeEquivalance =
+				AttributeEquivalenceClassForEquivalenceClass(plannerEqClass,
+															 relationRestriction);
+
+			attributeEquivalenceList =
+				AddAttributeClassToAttributeClassList(attributeEquivalenceList,
+													  attributeEquivalance);
+		}
```
anarazel: I'm still concerned with the amount of redundancy this is going to generate.  Aren't we frequently going to end up with the same thing repeated over and over here?  Yes, we're filtering some of that out later, but that's still O(n^3) in total or such.

src/backend/distributed/planner/multi_planner_utils.c
```diff
@@ -0,0 +1,863 @@
...
+										   expressionVar);
+		}
+	}
+
+	return attributeEquivalance;
+}
+
+
+/*
+ *
```
anarazel: superflous line.

src/backend/distributed/planner/multi_planner_utils.c
```diff
@@ -0,0 +1,863 @@
...
+
+	foreach(equivalenceClassMemberCell, equivalenceMemberList)
+	{
+		AttributeEquivalenceClassMember *newEqMember =
+			(AttributeEquivalenceClassMember *) lfirst(equivalenceClassMemberCell);
+
+		if (AttributeClassContainsAttributeClassMember(newEqMember, *firstClass))
+		{
+			continue;
+		}
```
anarazel: This is O(n^2), right?

src/backend/distributed/planner/multi_planner_utils.c
```diff
@@ -0,0 +1,863 @@
...
+													  AttributeEquivalenceClass *
+													  secondClass);
+
+/*
+ * AllRelationsJoinedOnPartitionKey aims to deduce whether each of the RTE_RELATION
+ * is joined with at least one another RTE_RELATION on their partition keys. If each
+ * RTE_RELATION follows the above rule, we can conclude that all RTE_RELATIONs are
+ * joined on their partition keys.
+ *
+ * The function returns true if all relations are joined on their partition keys.
```
anarazel: Are we actually guaranteeing anywhere that the relevant tables are all "co-partitioned"?

src/backend/distributed/planner/multi_planner_utils.c
onderkalaci: The filtering is happening on the `AddAttributeClassToAttributeClassList ()`, so we're not generating the same thing over and over. But, the algorithm that does the filtering is o(n^3). 

Note that this O(n^3) only includes partition keys, we're not adding all the restrictions to these lists. That's why I'm not very much concerned about that. Do you want me to work and improve that algorithm? Maybe using a hash table/bms or so?

src/backend/distributed/planner/multi_planner_utils.c
onderkalaci: Fixed

src/backend/distributed/planner/multi_planner_utils.c
onderkalaci: Yes, `ListConcatUniqueAttributeClassMemberLists ` is O(n^2).

onderkalaci:
@anarazel Do you have any further comments here?

src/backend/distributed/planner/multi_planner_utils.c
anarazel: I still think shard_pruning.c would be better (we've started to move away from multi_ prefixes btw), but I can live with that.

anarazel:
I'm still quite concerned about the complexities here.  Could you do some testing with creating queries that a) have a number of joins b) have a significant number of superfluous / redundant restrictions on partition keys / other columns (IIRC the way filtering is done still makes them have some cost).  I'm quite concerned that we'll be shipping this and suddenly a bunch of people will run into cubic complexity cases.

src/backend/distributed/planner/multi_planner.c
anarazel: By that logic, we don't need to copy, no?

src/backend/distributed/planner/relation_restriction_equivalence.c
```diff
@@ -0,0 +1,862 @@
...
+	 * If the query includes a single relation which is not a reference table,
+	 * we should not check the partition column equality.
+	 * Consider two example cases:
+	 *   (i)   The query includes only a single colocated relation
+	 *   (ii)  A colocated relation is joined with a (or multiple) reference
+	 *         table(s) where colocated relation is not joined on the partition key
+	 *
+	 * For the above two cases, we should not execute the partition column equality
+	 * algorithm.
+	 */
```
anarazel: This still needs a better explanation. Why "should" we not execute the algorithm?  Why is measuring this by the length of restrictionContext->relationRestrictionList correct?

mtuncer:
:shipit:

onderkalaci:
> We need to open a new issue to address the technical debt added by this commit, specifically the complexity codewise and algorithmically in relation_restriction_equivalence.c. We'll need to address that soon-ish.

@anarazel I opened an issue and shared some benchmark results. It seems that we're OK in terms of the actual running times. Please feel free to follow up in that [issue](https://github.com/citusdata/citus/issues/1322).

Issue #1270, opened 2017-03-02
SendRemoteCommandParams() segfaults if MultiConnection->pgConn is NULL


anarazel:
```
	PGconn *pgConn = connection->pgConn;
	bool wasNonblocking = PQisnonblocking(pgConn);
	int rc = 0;

	LogRemoteCommand(connection, command);

	/* make sure not to block anywhere */
	if (!wasNonblocking)
	{
		PQsetnonblocking(pgConn, true);
	}

	rc = PQsendQueryParams(pgConn, command, parameterCount, parameterTypes,
						   parameterValues, NULL, NULL, 0);

	/* reset nonblocking connection to its original state */
	if (!wasNonblocking)
	{
		PQsetnonblocking(pgConn, false);
	}
```

```PQisnonblocking()```, ```LogRemoteCommand()``` don't like NULL pgConns. ```PQsetnonblocking()```, ```PQsendQueryParams()``` do.

We shouldn't try to send queries over closed connections, but we apparently missing a check somewhere. A second line of defense seems quite advisable.

@sumedhpathak I think we should fix & backport this soon.

@fdr @lfittl Might be relevant for you guys.

PR #1271, opened 2017-03-02
Fix SendRemoteCommandParams() handling of a NULL MultiConnection->pgConn

anarazel:
Previously we'd segfault in PQisnonblocking() which, contrary to other
libpq calls, doesn't handle a NULL PQconn (because there'd be no
appropriate return value for that).

The bug can be reproduced by firing a router query against a host where connections don't directly fail (e.g. max_connections, or shutdown host), but fail via timeout (reproduced by kill -STOP ing postmaster).  Unfortunately it's not easy to add a test for this.

Fixes: #1270

### DIFF
--- a/src/backend/distributed/connection/remote_commands.c
+++ b/src/backend/distributed/connection/remote_commands.c
@@ -277,10 +277,21 @@ SendRemoteCommandParams(MultiConnection *connection, const char *command,
 						const char *const *parameterValues)
 {
 	PGconn *pgConn = connection->pgConn;
-	bool wasNonblocking = PQisnonblocking(pgConn);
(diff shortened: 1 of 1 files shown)


marcocitus:
Which code path calls SendRemoteCommandParams on a failed connection?

anarazel:
> Which code path calls SendRemoteCommandParams on a failed connection?

Plain router queries unfortunately...  I think we might want to do something about that as well, but it seems better to harden SendRemoteCommandParams(), given it's mirroring libpq functionality which is safe against NULL connections..

jasonmp85:
:shipit:

Issue #1257, opened 2017-02-27
Router planner support for range partitioned tables


onderkalaci:
We currently support router planner for hash distributed and reference tables. But, it seems like there is no blocker for using it for range partitioned tables as well.

Besides with #1204 we're planning to utilize router planner basics for co-located subquery pushdown planner. Thus, enabling router planner support could enable subquery pushdown for range partitioned tables.

PR #1010, opened 2016-12-02
Evaluate functions in INSERT..SELECT

marcocitus:
While writing https://www.citusdata.com/blog/2016/11/29/event-aggregation-at-scale-with-postgresql/ I ran into a number of usability issues with INSERT..SELECT. In particular, it errors out when using functions like now(), date_trunc(), the timestamptz type, or even simple string||int concatenation, which can be very inconvenient, especially for doing roll-ups.

The problem with doing function evaluation for INSERT .. SELECT is that we perform different transformation on the query trees in the planner for each shard. While we could pass the entire query tree into the plan for each task and subsequently perform function evaluation on each query tree, this can cause substantial overhead.

This PR introduces a generic list of relation<->shard mappings ("RelationShard") that gets passed down with every (router planner) task, such that the per-shard query tree can be reconstructed from the jobQuery in the router executor and applies function evaluation to the SELECT part of the INSERT .. SELECT.  The infrastructure will also allow us to deparse and thus perform function evaluation on more complex queries (e.g. SELECT) in the executor. 

The RelationShard list replaces the existing selectShardList used for INSERT..SELECT lcoking, since it contains the same information. This change reduces overhead since selectShardList was a list of ShardInterval structs, whereas RelationShard only contains a relationId and a shardId.

Part of #961

PR #1272, opened 2017-03-03
Enable router planner for queries on range partitioned table

mtuncer:
We used to exclude range partitioned tables from router planning. This checkin removes that restriction.

Mix mode queries however are still rejected by router planner. So one query either have hash partitioned (and reference) tables or, range partitioned (and reference) tables. Queries involving hash and range partitioned tables are still rejected. 

We could possibly run some of those queries if certain conditions hold, however, it may cause bad customer experience when some queries magically becomes supported and not supported. 

Fixes : #1257

### DIFF
--- a/src/backend/distributed/planner/multi_router_planner.c
+++ b/src/backend/distributed/planner/multi_router_planner.c
@@ -2660,7 +2660,7 @@ MultiRouterPlannableQuery(Query *query, RelationRestrictionContext *restrictionC
 			char partitionMethod = PartitionMethod(distributedTableId);
 
 			if (!(partitionMethod == DISTRIBUTE_BY_HASH || partitionMethod ==
-				  DISTRIBUTE_BY_NONE))
+				  DISTRIBUTE_BY_NONE || partitionMethod == DISTRIBUTE_BY_RANGE))
 			{
 				return false;
 			}
--- a/src/test/regress/expected/multi_router_planner.out
+++ b/src/test/regress/expected/multi_router_planner.out
@@ -15,10 +15,18 @@ CREATE TABLE articles_range (
 	title varchar(20) NOT NULL,
 	word_count integer
 );
+CREATE TABLE articles_append (
+	id bigint NOT NULL,
+	author_id bigint NOT NULL,
+	title varchar(20) NOT NULL,
+	word_count integer
+);
 -- Check for the existence of line 'DEBUG:  Creating router plan'
 -- to determine if router planner is used.
 -- this table is used in a CTE test
-CREATE TABLE authors_hash ( name text, id bigint );
+CREATE TABLE authors_hash ( name varchar(20), id bigint );
+CREATE TABLE authors_range ( name varchar(20), id bigint );
+CREATE TABLE authors_reference ( name varchar(20), id bigint );
 -- this table is used in router executor tests
 CREATE TABLE articles_single_shard_hash (LIKE articles_hash);
 SELECT master_create_distributed_table('articles_hash', 'author_id', 'hash');
@@ -49,6 +57,12 @@ SELECT master_create_worker_shards('articles_hash', 2, 1);
 SELECT master_create_worker_shards('articles_single_shard_hash', 1, 1);
  master_create_worker_shards 
 -----------------------------
+ 
+(1 row)
+
+SELECT create_reference_table('authors_reference');
+ create_reference_table 
+------------------------
  
 (1 row)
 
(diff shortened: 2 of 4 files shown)


src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -2660,7 +2662,19 @@ MultiRouterPlannableQuery(Query *query, RelationRestrictionContext *restrictionC
 			char partitionMethod = PartitionMethod(distributedTableId);
 
 			if (!(partitionMethod == DISTRIBUTE_BY_HASH || partitionMethod ==
-				  DISTRIBUTE_BY_NONE))
+				  DISTRIBUTE_BY_NONE || partitionMethod == DISTRIBUTE_BY_RANGE))
+			{
+				return false;
+			}
+
+			hasHashDistributedTable |= partitionMethod == DISTRIBUTE_BY_HASH;
```
onderkalaci: Do we really need this? Shard pruning logic already handles restrictions differently per distribution model. Even if we use hash and range partitioned tables in the same query, they both should be able to pruned down to a single shard.

I don't see a reason for this check also since we allow such queries on multi-shard planner. For instance, if you enable task-tracker on the test that errors out, the query yields the correct result.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -2660,7 +2662,19 @@ MultiRouterPlannableQuery(Query *query, RelationRestrictionContext *restrictionC
...
+			hasRangeDistributedTable |= partitionMethod == DISTRIBUTE_BY_RANGE;
+
+			/*
+			 * we could probably relax this, however, we should not
```
onderkalaci: (I agree your comment that we should not promote this, but, I'm more inclined to allow it.)

src/test/regress/sql/multi_router_planner.sql
```diff
@@ -706,26 +717,95 @@ SELECT author_id FROM articles_hash
...
+UPDATE pg_dist_shard SET shardminvalue = 31, shardmaxvalue=40 WHERE shardid = :shard_id;
+
+SELECT master_create_empty_shard('articles_range') as shard_id \gset
+UPDATE pg_dist_shard SET shardminvalue = 1, shardmaxvalue=10 WHERE shardid = :shard_id;
+
+SELECT master_create_empty_shard('articles_range') as shard_id \gset
+UPDATE pg_dist_shard SET shardminvalue = 11, shardmaxvalue=30 WHERE shardid = :shard_id;
+
+SELECT master_create_empty_shard('articles_range') as shard_id \gset
+UPDATE pg_dist_shard SET shardminvalue = 21, shardmaxvalue=40 WHERE shardid = :shard_id;
```
mtuncer: yes it was intentional

src/test/regress/expected/multi_router_planner.out
```diff
@@ -1617,48 +1631,229 @@ DEBUG:  Plan is router executable
...
+SELECT * FROM articles_range ar join authors_range au on (ar.author_id = au.id)
+	WHERE ar.author_id = 1 and au.id = 2;
+DEBUG:  Creating router plan
+DEBUG:  Plan is router executable
+ id | author_id | title | word_count | name | id 
+----+-----------+-------+------------+------+----
+(0 rows)
+
+-- this is a bug, it is a single shard join query but not router plannable
+SELECT * FROM articles_range ar join authors_range au on (ar.author_id = au.id)
```
mtuncer: Perhaps we should open an issue on this if it there is not one

src/backend/distributed/planner/multi_router_planner.c
mtuncer: We don't allow joins between different distribution types in our real-time planner/executor. This kind of joins are handled by repartitioning. 

I do not see a value allowing some of them in router execution. The same query could become magically supported and not supported between 2 setups (test and prod).

onderkalaci: > This kind of joins are handled by repartitioning.

Yes, what I meant was we already allow such combinations in the other planners/executors, why should we disallow it here?

> I do not see a value allowing some of them in router execution. T

I think the value here is that both planner/executor pairs behave similarly for the similar queries (in terms of partition types.)

> The same query could become magically supported and not supported between 2 setups (test and prod).

That's right. But, if you're dealing with range partitioned tables the responsibility of maintaining co-location is is always on you. If we think in that way, we shouldn't allow router planner for range partitioned tables at all.

So, I don't see it similar to the case that we've disallowed router planner hitting two shards on the same worker in [here](https://github.com/citusdata/citus/pull/1010/commits/d745d7bf70cbf12318b84d4971ddab22224f2c84) (i.e., `RelationPrunesToMultipleShards ()`)

mtuncer:
@onderkalaci requested changes are made, commits are squashed

onderkalaci:
This look good :shipit: 

 One final minor request (feel free to ignore): Could you write a single line commit message (except for the headline)? I personally don't like reading commits without any messages in the git history.

mtuncer:
done

PR #1274, opened 2017-03-08
Remove unused SendCommandToWorker

lithp:
Assigning you because I think you wrote the original code, this function does not appear to be used.

### DIFF
--- a/src/backend/distributed/transaction/worker_transaction.c
+++ b/src/backend/distributed/transaction/worker_transaction.c
@@ -33,29 +33,6 @@
 
 
 /*
- * SendCommandToWorker sends a command to a particular worker as part of the
- * 2PC.
- */
-void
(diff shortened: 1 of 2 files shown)


jasonmp85:
Hey, so `SendCommandToWorker` is used in Enterprise.

jasonmp85:
Ok, so this is blocking a merge to enterprise where I'd like to test some recent changes. `SendCommandToWorkers` looks similar, but I don't have the time to look into whether I can refactor the Enterprise caller to use it, or if it's actually equivalent.

 Reverting for now to unblock myself.

lithp:
Ouch, sorry about that, I forgot to check there. Thinking about ways to prevent this in the future, we can add a third travis check, that PRs merge build and test cleanly over there?

PR #1278, opened 2017-03-12
Execute DDL on coordinator before workers

### LINKED ISSUE #357: Consider issuing DDL commands on the master before the worker nodes
@anarazel mentioned

"Executing the DDL on the workers before doing so on the master strikes me as rather fragile. There's always going to be problems without using 2PC, but if we were to do this on the master first, we'd at least have a chance to mark the shards where the statement failed as broken. This way round there's no way of doing that."

The rationale for doing this on the worker nodes first was to verify that the command can go through on atleast one shard before we commit to doing it on the master.

However, since we can rollback on the master node in case a query fails on the first worker node, we could run it on the master first. So, there don't seem to be advantages of doing it on the workers before doing it on the master.

However, if we do execute it on the master first, we have the ability to rollback and that makes it better for certain cases.

# Plan

  - [x] Reorganize `multi_ProcessUtility` to better separate DDL mode from other modes
  - [x] Determine call sites which modify parse tree
  - [x] Refactor modifications into separate methods, as needed
  - [x] Change call order of `standard_ProcessUtility` to happen first
  - [ ] Evaluate locking semantics
  - [ ] Verify no unit tests break
  - [ ] Run against PostgreSQL tests

jasonmp85:
Fixes #357

My initial pass at this had some problems with `DROP INDEX` in particular, as dropping the local index first means we can no longer look up what relation owns the index. To address this, I removed the execution portion of all the _Process-_ methods within `multi_utility` and added a `DDLJob` struct to represent the work to eventually be done on the workers.

I renamed such methods to begin with _Plan-_ instead. Their new contract is to perform any needed modifications on the query (which they already did, despite pretending to return a modified version) and return a `DDLJob`. The (possibly) modified query executes on the coordinator, then the `DDLJob` is used to perform MX metadata syncing and `DDL` propagation.

Two design decisions:

  * The concerns of the old _Process-_ methods are myriad. I think there could be some consistency here to first verify the query for distributed execution, emit messages and warnings, and create a plan. Each sub-method does this in a slightly different order, which is confusing. I avoided modifying any of their logic to keep the review focused
  * The locks taken to look up whether targeted objects are associated with distributed relations could probably be relaxed. In fact, it's possibly only some DDL commands require evaluation before coordinator DDL execution (see my comment about `DROP INDEX` above); however, for consistency, I just moved all _Plan-_ (née _Process-_) calls above the call to `standard_ProcessUtility`

These two simplifications can be revisited if the reviewer requires.

# TODO

  - [x] Documentation pass
  - [x] Fix MX test
  - [x] Change `DDLJob *` to a `List *` of `DDLJob *`s (necessary for Enterprise's `GRANT` support)
  - [x] Rebase onto `master` when the time comes

### DIFF
--- a/src/backend/distributed/executor/multi_utility.c
+++ b/src/backend/distributed/executor/multi_utility.c
@@ -168,6 +168,7 @@ multi_ProcessUtility(Node *parsetree,
 	bool commandMustRunAsOwner = false;
 	Oid savedUserId = InvalidOid;
 	int savedSecurityContext = 0;
+	DDLJob *ddlJob = NULL;
 
 	if (IsA(parsetree, TransactionStmt))
 	{
@@ -252,8 +253,6 @@ multi_ProcessUtility(Node *parsetree,
 	 */
 	if (EnableDDLPropagation)
 	{
-		DDLJob *ddlJob = NULL;
-
 		if (IsA(parsetree, IndexStmt))
 		{
 			ddlJob = PlanIndexStmt((IndexStmt *) parsetree, queryString);
@@ -311,11 +310,6 @@ multi_ProcessUtility(Node *parsetree,
 									 "commands to worker nodes"),
 							  errhint("Connect to worker nodes directly to manually "
 									  "move all tables.")));
-		}
-
-		if (ddlJob != NULL)
-		{
-			ExecuteDistributedDDLJob(ddlJob);
 		}
 	}
 	else
@@ -379,6 +373,11 @@ multi_ProcessUtility(Node *parsetree,
 	if (commandMustRunAsOwner)
 	{
 		SetUserIdAndSecContext(savedUserId, savedSecurityContext);
+	}
+
+	if (ddlJob != NULL)
+	{
+		ExecuteDistributedDDLJob(ddlJob);
 	}
 
 	/* we run VacuumStmt after standard hook to benefit from its checks and locking */
--- a/src/backend/distributed/executor/multi_utility.c
+++ b/src/backend/distributed/executor/multi_utility.c
@@ -382,6 +382,7 @@ multi_ProcessUtility(Node *parsetree,
 		foreach(ddlJobCell, ddlJobs)
 		{
 			DDLJob *ddlJob = (DDLJob *) lfirst(ddlJobCell);
+
 			ExecuteDistributedDDLJob(ddlJob);
 		}
 	}
@@ -2362,15 +2363,14 @@ CopyGetAttnums(TupleDesc tupDesc, Relation rel, List *attnamelist)
 
 
 /*
- * ReplicateGrantStmt replicates GRANT/REVOKE command to worker nodes if the
+ * PlanGrantStmt replicates GRANT/REVOKE command to worker nodes if the
  * the statement affects distributed tables.
  *
  * NB: So far column level privileges are not supported.
  */
-void
-ReplicateGrantStmt(Node *parsetree)
+List *
+PlanGrantStmt(GrantStmt *grantStmt)
 {
-	GrantStmt *grantStmt = (GrantStmt *) parsetree;
 	StringInfoData privsString;
 	StringInfoData granteesString;
 	StringInfoData targetString;
@@ -2378,7 +2378,7 @@ ReplicateGrantStmt(Node *parsetree)
 	ListCell *granteeCell = NULL;
 	ListCell *objectCell = NULL;
 	bool isFirst = true;
-	DDLJob *ddlJob = palloc(sizeof(DDLJob));
+	List *ddlJobs = NIL;
 
 	initStringInfo(&privsString);
 	initStringInfo(&granteesString);
@@ -2392,7 +2392,7 @@ ReplicateGrantStmt(Node *parsetree)
 	if (grantStmt->targtype != ACL_TARGET_OBJECT ||
 		grantStmt->objtype != ACL_OBJECT_RELATION)
 	{
-		return;
+		return NIL;
 	}
 
 	/* deparse the privileges */
@@ -2463,6 +2463,7 @@ ReplicateGrantStmt(Node *parsetree)
 		RangeVar *relvar = (RangeVar *) lfirst(objectCell);
 		Oid relOid = RangeVarGetRelid(relvar, NoLock, false);
 		const char *grantOption = "";
+		DDLJob *ddlJob = NULL;
 
 		if (!IsDistributedTable(relOid))
 		{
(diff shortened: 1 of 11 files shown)


lithp:
This looks great! It's a little suspicious that you're changing the tests at the same time you change behavior but this test change looks reasonable.

Annoying you have to go through all this effort just to make `DROP INDEX` work. It looks like the only reason we hit the table in these `Process` functions is to check whether the object is distributed. Pulling out an `InvolvesDistributedRelation` function feels like it could require less work but that, especially for `PlanIndexStatement`, involves the refactoring in (1) which you mentioned not wanting to do.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -257,20 +253,17 @@ multi_ProcessUtility(Node *parsetree,
 	 */
 	if (EnableDDLPropagation)
 	{
-		bool isTopLevel = (context == PROCESS_UTILITY_TOPLEVEL);
-
 		if (IsA(parsetree, IndexStmt))
 		{
-			parsetree = ProcessIndexStmt((IndexStmt *) parsetree, queryString,
-										 isTopLevel);
+			ddlJob = PlanIndexStmt((IndexStmt *) parsetree, queryString);
```
marcocitus: Hm, I was kind of hoping we'd now be in a position to support things like `CREATE INDEX ON ..`  without an explicit name. What if we only keep `DropStmt` above `standard_ProcessUtility`?

jasonmp85: Was starting out consistent and can carve out modifications as needed during the review. I agree that there are probably only certain commands that need planning first but was keeping changes to mostly shuffling things around in as least invasive a way possible.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -384,6 +375,11 @@ multi_ProcessUtility(Node *parsetree,
 		SetUserIdAndSecContext(savedUserId, savedSecurityContext);
 	}
 
+	if (ddlJob != NULL)
+	{
+		ExecuteDistributedDDLJob(ddlJob);
+	}
+
 	/* we run VacuumStmt after standard hook to benefit from its checks and locking */
 	if (IsA(parsetree, VacuumStmt))
```
anarazel: I assume vacuum can't be done via ExecuteDistributedDDLJob due to transactional behaviour and such?

src/test/regress/expected/multi_mx_modifying_xacts.out
```diff
@@ -183,7 +183,7 @@ ABORT;
 -- applies to DDL
 BEGIN;
 INSERT INTO labs_mx VALUES (6, 'Bell labs_mx');
-ALTER TABLE labs_mx ADD COLUMN text motto;
+ALTER TABLE labs_mx ADD COLUMN motto text;
```
jasonmp85: Yeah I hate this thinko.

src/backend/distributed/executor/multi_utility.c
jasonmp85: It easily could be. In fact, I could probably shoehorn `COPY` in here with enough effort. Didn't know how much of that you wanted.

jasonmp85:
@anarazel two things:

  * My `GRANT` changes were still on my rebased-onto-enterprise local version. I've moved them into this PR. Pretty similar to everything else, except now the `DDLJob` to be executed lives in a list (a single incoming `GRANT` might produce multiple worker jobs)
  * Did you have any thoughts on removing or changing any of the locking? I was waiting to rip out any of it before your review, though it's certainly more than is necessary for this change. @marcocitus mentioned rejiggering things further to permit added functionality, but I think that can be done more easily after this change if needed

src/backend/distributed/executor/multi_utility.c
```diff
@@ -608,17 +611,18 @@ ProcessCopyStmt(CopyStmt *copyStatement, char *completionTag, bool *commandMustR
 
 
 /*
- * ProcessIndexStmt processes create index statements for distributed tables.
+ * PlanIndexStmt processes create index statements for distributed tables.
```
metdos: Shall we also update the comments with the names given that we don't "execute" anything here? Similarly for other Plan_XXX_Stmt functions.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -384,6 +375,18 @@ multi_ProcessUtility(Node *parsetree,
 		SetUserIdAndSecContext(savedUserId, savedSecurityContext);
 	}
 
+	if (ddlJobs != NIL)
```
metdos: Maybe a comment which notes we run the ddl on the coordinator above with standard process utility and now we will run on worker nodes.

src/backend/distributed/executor/multi_utility.c
jasonmp85: Passing on this, though it would be easy to do once #1287 is merged (since it adds a flag to ensure a particular `DDLJob` runs outside of a transaction).

src/backend/distributed/executor/multi_utility.c
```diff
@@ -673,7 +670,10 @@ ProcessIndexStmt(IndexStmt *createIndexStatement, const char *createIndexCommand
 			/* if index does not exist, send the command to workers */
 			if (!OidIsValid(indexRelationId))
 			{
-				ExecuteDistributedDDLCommand(relationId, createIndexCommand, isTopLevel);
+				ddlJob = palloc0(sizeof(DDLJob));
+				ddlJob->targetRelationId = relationId;
+				ddlJob->commandString = createIndexCommand;
+				ddlJob->taskList = DDLTaskList(relationId, createIndexCommand);
 			}
 			else if (!createIndexStatement->if_not_exists)
```
jasonmp85: Sure, removed.

src/include/distributed/multi_utility.h
```diff
@@ -14,6 +14,13 @@
 
 extern bool EnableDDLPropagation;
 
+typedef struct DDLJob
+{
+	Oid targetRelationId;
+	const char *commandString;
+	List *taskList;
+} DDLJob;
```
jasonmp85: Sure.

src/backend/distributed/executor/multi_utility.c
jasonmp85: Yes, updating all comments.

src/backend/distributed/executor/multi_utility.c
jasonmp85: I'll add a top-level command describing the workflow.

jasonmp85:
Merging based on @anarazel's review approval. I addressed all his comments and all of @metdos's minor comments as well, so getting this in to make subsequent review cleaner.

PR #1281, opened 2017-03-15
Fix access permission checks for distributed relations

metdos:
We ditched the access permission check with Custom Scan API changes. This PR brings it back.

### DIFF
--- a/src/backend/distributed/planner/multi_planner.c
+++ b/src/backend/distributed/planner/multi_planner.c
@@ -454,6 +454,9 @@ FinalizeNonRouterPlan(PlannedStmt *localPlan, MultiPlan *multiPlan,
 	finalPlan->queryId = localPlan->queryId;
 	finalPlan->utilityStmt = localPlan->utilityStmt;
 
+	/* add original range table list for access permission checks */
+	finalPlan->rtable = list_concat(finalPlan->rtable, localPlan->rtable);
+
 	return finalPlan;
 }
 
@@ -473,7 +476,7 @@ FinalizeRouterPlan(PlannedStmt *localPlan, CustomScan *customScan)
 	List *targetList = NIL;
 	List *columnNameList = NIL;
 
-	/* we will have only one range table entry */
+	/* we will have custom scan range table entry as the first one in the list */
 	int customScanRangeTableIndex = 1;
 
 	/* build a targetlist to read from the custom scan output */
@@ -513,6 +516,9 @@ FinalizeRouterPlan(PlannedStmt *localPlan, CustomScan *customScan)
 
 	remoteScanRangeTableEntry = RemoteScanRangeTableEntry(columnNameList);
 	routerPlan->rtable = list_make1(remoteScanRangeTableEntry);
+
+	/* add original range table list for access permission checks */
+	routerPlan->rtable = list_concat(routerPlan->rtable, localPlan->rtable);
 
 	routerPlan->canSetTag = true;
 	routerPlan->relationOids = NIL;
--- a/src/test/regress/sql/multi_multiuser.sql
+++ b/src/test/regress/sql/multi_multiuser.sql
@@ -1,0 +1,92 @@
+--
+-- MULTI_MULTIUSERS
+--
+-- Test user permissions.
+--
+
+ALTER SEQUENCE pg_catalog.pg_dist_shardid_seq RESTART 1420000;
+ALTER SEQUENCE pg_catalog.pg_dist_jobid_seq RESTART 1420000;
+
+SET citus.shard_replication_factor TO 1;
+SET citus.shard_count TO 2;
+
+CREATE TABLE test (id integer);
+SELECT create_distributed_table('test', 'id');
+
+CREATE USER full_access;
+CREATE USER read_access;
+CREATE USER no_access;
+
+GRANT ALL ON TABLE test TO full_access;
+GRANT SELECT ON TABLE test TO read_access;
+
+\c - - - :worker_1_port
+CREATE USER full_access;
+CREATE USER read_access;
+CREATE USER no_access;
+
+GRANT ALL ON TABLE test_1420000 TO full_access;
+GRANT SELECT ON TABLE test_1420000 TO read_access;
+
+\c - - - :worker_2_port
+CREATE USER full_access;
+CREATE USER read_access;
+CREATE USER no_access;
+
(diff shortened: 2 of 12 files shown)


src/backend/distributed/planner/multi_planner.c
```diff
@@ -295,6 +295,15 @@ CreateDistributedPlan(PlannedStmt *localPlan, Query *originalQuery, Query *query
 	resultPlan = FinalizePlan(localPlan, distributedPlan);
 
 	/*
+	 * PostgreSQL checks access permissions in standard_ExecutorStart(), but
+	 * we do these checks here just before returning from planner. After this
+	 * point, we don't have access to our local tables in our plan. Therefore,
```
onderkalaci: I think `local tables` is a bit confusing since we're using distributed relations in the next sentence. I'd rather use `original queries' range table list` or smth like that.

metdos:
@onderkalaci refused to give a ship, so I moved him to reviewers :D

onderkalaci:
This looks good to me.

Let me share our conversation here on this:

* As far as we can see, the function should be called on the `standard_ExecutorStart()`
* After the custom scan node changes, we don't have access to the `standard_ExecutorStart()` (Postgres calls it for us!)
* In `standard_ExecutorStart()` , `ExecCheckRTPerms()` is already called. However, since the distributed plan is wrapped into a `customScan` node, the checks are done on the range tables of the `customScan`.
* Thus, we need to call that function just before the execution starts with the correct range tables. And, the appropriate place seems to be the end of `CreateDistributedPlan ()`

So, it's a :shipit: except for the very minor comment. But, let's give a chance to @anarazel and/or @jasonmp85 before merging in.

src/backend/distributed/planner/multi_planner.c
metdos: Updated the comment.

anarazel:
I don't think approach is sufficient. Wit prepared statements this is going to potentially happen in a different transaction than originally.

I'm also not following:
> In standard_ExecutorStart() , ExecCheckRTPerms() is already called. However, since the distributed plan is > wrapped into a customScan node, the checks are done on the range tables of the customScan.

The custom scan doesn't have a range table...  Isn't the fix here to just retain the original rtable in FinalizeRouterPlan()?

metdos:
> I don't think approach is sufficient. Wit prepared statements this is going to potentially happen in a different transaction than originally.

Darn, prepared statements.

I verified the problem here, we can extend regression tests.
https://github.com/citusdata/citus-enterprise/commit/09974b626c44245e246d4ed0e9d0b5bb1a2db5ec#diff-65502fee3d6eec89af3b80edfe2c121bR25

> The custom scan doesn't have a range table... Isn't the fix here to just retain the original rtable in FinalizeRouterPlan()?

@anarazel, are you proposing something like this?
```C
void
CitusSelectBeginScan(CustomScanState *node, EState *estate, int eflags)
{
	CitusScanState *scanState = (CitusScanState *) node;
	MultiPlan *multiPlan = scanState->multiPlan;
	Job *workerJob = multiPlan->workerJob;
	Query *jobQuery = workerJob->jobQuery;
	List *rangeTableList = jobQuery->rtable;

	ExecCheckRTPerms(rangeTableList, true);
}
```
If yes, currently range table entries are RTE_FUNCTION kind. I need to find a solution for this.

I couldn't look into this more today because of other code reviews etc. Monday, I will look into this more.

anarazel:
@metdos No, not really. In FinalizeRouterPlan() we currently do
```
	routerPlan = makeNode(PlannedStmt);
...
	routerPlan->rtable = list_make1(remoteScanRangeTableEntry);
```
so we replace the normal PlannedStmt with one that doesn't have the original rtable.  Which prevents ExecCheckRTPerms() from doing its job.   What I'm proposing is to just copy the rtable from `localPlan` (and thus set `customScanRangeTableIndex` to `list_length(localPlan->rtable)`).  That should take care of the whole issue, no?

metdos:
> What I'm proposing is to just copy the rtable from localPlan (and thus set customScanRangeTableIndex to list_length(localPlan->rtable)). That should take care of the whole issue, no?

This sounds good, but the same problem also exists for realtime-tasktracker executors. Therefore, I need to fix it also. OOTO currently, will take a look on Monday evening.

metdos:
I fixed the issue by adding original range table list to our finalized plan for both router and non-router executors. This has a side effect on the explain output. It adds table name (remote_scan) to column references which is fine I think.

I tested these changes on enterprise repo manually and with check-enterprise. I also added a test (https://github.com/citusdata/citus-enterprise/commit/e571b26a7ab4b7587bff90d86465baa9eb17a867#diff-980c54907a09b5a859437b45282df138R42) for prepared statements in permission checks test suite which we can merge after merging this PR.

Another possible fix is adding original range table to MultiPlan and running ExecCheckRTPerms() on this range table list explicitly at the beginning of the executions (CitusSelectBeginScan).

anarazel:
I think we should have a basic permission checking test case in this repo.

metdos:
Hey @jasonmp85, @anarazel, I added the regression test.

jasonmp85:
Rebased this for myself/you.

src/test/regress/sql/multi_multiuser.sql
```diff
@@ -0,0 +1,92 @@
...
+SET citus.shard_count TO 2;
+
+CREATE TABLE test (id integer);
+SELECT create_distributed_table('test', 'id');
+
+CREATE USER full_access;
+CREATE USER read_access;
+CREATE USER no_access;
+
+GRANT ALL ON TABLE test TO full_access;
```
jasonmp85: Could probably have had slightly less noisy test output by doing this before distributing the table.

src/test/regress/sql/multi_multiuser.sql
```diff
@@ -0,0 +1,92 @@
...
+GRANT SELECT ON TABLE test TO read_access;
+
+\c - - - :worker_1_port
+CREATE USER full_access;
```
jasonmp85: Is this strictly necessary? The checks take place on the coordinator, no? I agree these statements on the worker make the test more realistic, but they're probably overkill to ensure permissions checks are happening.

sumedhpathak:
👍 Thanks Jason /Metin/Andres!

jasonmp85:
Amusingly, it appears this new test file was not run against Enterprise, where it fails. Fixing.

PR #1282, opened 2017-03-15
Physical planner changes for subquery pushdown

onderkalaci:
This PR allows the phyiscal planner to utilize `INSERT ... SELECT` logic.

This PR is on top io #1268 and strictly depends on that PR.

This commit applies the logic that exists in INSERT .. SELECT planning to the subquery pushdown changes.
    
The main algorithm is followed as:
```
       - pick an anchor relation (i.e., target relation)
       - per each target shard interval
           - add the target shard interval's shard range
             as a restriction to the relations (if all relations
             joined on the partition keys)
            - Check whether the query is router plannable per
              target shard interval.
            - If router plannable, create a task
```
Also, I realized an important issue in this PR in terms of shard pruning for router planner. 

I'll share an example which is already apparent in the regression tests (we used to create 2 tasks now we create 4 shards due to the following behavior see changed explain outputs):
Also, here are the steps to re-produce the issue [here](https://gist.github.com/onderkalaci/f421f8a07adc89502cad847b889500cb). Basically, the query hits a single shard, but router planner fails to detect that.

This might be critical if there are too many shards and the query hits only a few.

### DIFF
--- a/src/backend/distributed/planner/multi_logical_optimizer.c
+++ b/src/backend/distributed/planner/multi_logical_optimizer.c
@@ -147,16 +147,15 @@ static bool TablePartitioningSupportsDistinct(List *tableNodeList,
 static bool GroupedByColumn(List *groupClauseList, List *targetList, Var *column);
 
 /* Local functions forward declarations for subquery pushdown checks */
-static void ErrorIfContainsUnsupportedSubquery(MultiNode *logicalPlanNode);
+static void ErrorIfContainsUnsupportedSubquery(MultiNode *logicalPlanNode,
+											   PlannerRestrictionContext *
+											   plannerRestrictionContext);
 static void ErrorIfCannotPushdownSubquery(Query *subqueryTree, bool outerQueryHasLimit);
 static void ErrorIfUnsupportedTableCombination(Query *queryTree);
 static void ErrorIfUnsupportedUnionQuery(Query *unionQuery);
 static bool TargetListOnPartitionColumn(Query *query, List *targetEntryList);
 static FieldSelect * CompositeFieldRecursive(Expr *expression, Query *query);
 static bool FullCompositeFieldList(List *compositeFieldList);
-static Query * LateralQuery(Query *query);
-static bool SupportedLateralQuery(Query *parentQuery, Query *lateralQuery);
-static bool JoinOnPartitionColumn(Query *query);
 static void ErrorIfUnsupportedShardDistribution(Query *query);
 static List * RelationIdList(Query *query);
 static bool CoPartitionedTables(Oid firstRelationId, Oid secondRelationId);
@@ -189,9 +188,15 @@ static bool HasOrderByHllType(List *sortClauseList, List *targetList);
  * Third, the function pulls up the collect operators in the tree. Fourth, the
  * function finds the extended operator node, and splits this node into master
  * and worker extended operator nodes.
+ *
+ * We also pass plannerRestrictionContext to the optimizer. The context
+ * is primarily used to decide whether the subquery is safe to pushdown.
+ * If not, it helps to produce meaningful error messages for subquery
+ * pushdown planning.
  */
 void
-MultiLogicalPlanOptimize(MultiTreeRoot *multiLogicalPlan)
+MultiLogicalPlanOptimize(MultiTreeRoot *multiLogicalPlan,
+						 PlannerRestrictionContext *plannerRestrictionContext)
 {
 	bool hasOrderByHllType = false;
 	List *selectNodeList = NIL;
@@ -211,7 +216,7 @@ MultiLogicalPlanOptimize(MultiTreeRoot *multiLogicalPlan)
 	ErrorIfContainsUnsupportedAggregate(logicalPlanNode);
 
 	/* check that we can pushdown subquery in the plan */
-	ErrorIfContainsUnsupportedSubquery(logicalPlanNode);
+	ErrorIfContainsUnsupportedSubquery(logicalPlanNode, plannerRestrictionContext);
 
 	/*
 	 * If a select node exists, we use the idempower property to split the node
@@ -2805,13 +2810,15 @@ GroupedByColumn(List *groupClauseList, List *targetList, Var *column)
  * the subquery.
  */
 static void
-ErrorIfContainsUnsupportedSubquery(MultiNode *logicalPlanNode)
+ErrorIfContainsUnsupportedSubquery(MultiNode *logicalPlanNode,
+								   PlannerRestrictionContext *plannerRestrictionContext)
 {
 	Query *subquery = NULL;
 	List *extendedOpNodeList = NIL;
 	MultiTable *multiTable = NULL;
 	MultiExtendedOp *extendedOpNode = NULL;
 	bool outerQueryHasLimit = false;
+	bool restrictionEquivalenceForPartitionKeys = false;
 
 	/* check if logical plan includes a subquery */
 	List *subqueryMultiTableList = SubqueryMultiTableList(logicalPlanNode);
@@ -2822,6 +2829,18 @@ ErrorIfContainsUnsupportedSubquery(MultiNode *logicalPlanNode)
 
 	/* currently in the planner we only allow one subquery in from-clause*/
 	Assert(list_length(subqueryMultiTableList) == 1);
+
+	restrictionEquivalenceForPartitionKeys =
+		RestrictionEquivalenceForPartitionKeys(plannerRestrictionContext);
+	if (!restrictionEquivalenceForPartitionKeys)
+	{
+		ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+						errmsg("cannot pushdown the subquery since all relations are not "
+							   "joined using distribution keys"),
+						errdetail("Each relation should be joined with at least "
+								  "one another relation using distribution keys and "
+								  "equality operator.")));
+	}
 
 	multiTable = (MultiTable *) linitial(subqueryMultiTableList);
 	subquery = multiTable->subquery;
@@ -2895,7 +2914,6 @@ ErrorIfCannotPushdownSubquery(Query *subqueryTree, bool outerQueryHasLimit)
 {
 	bool preconditionsSatisfied = true;
 	char *errorDetail = NULL;
-	Query *lateralQuery = NULL;
 	List *subqueryEntryList = NIL;
 	ListCell *rangeTableEntryCell = NULL;
 
@@ -2989,43 +3007,6 @@ ErrorIfCannotPushdownSubquery(Query *subqueryTree, bool outerQueryHasLimit)
 		preconditionsSatisfied = false;
 		errorDetail = "Having qual without group by on partition column is "
 					  "currently unsupported";
-	}
-
-	/*
-	 * Check if join is supported. We check lateral joins differently, because
-	 * lateral join representation in query tree is a bit different than normal
-	 * join queries.
-	 */
-	lateralQuery = LateralQuery(subqueryTree);
-	if (lateralQuery != NULL)
-	{
-		bool supportedLateralQuery = SupportedLateralQuery(subqueryTree, lateralQuery);
-		if (!supportedLateralQuery)
-		{
-			preconditionsSatisfied = false;
-			errorDetail = "This type of lateral query in subquery is currently "
-						  "unsupported";
-		}
-	}
-	else
-	{
-		List *joinTreeTableIndexList = NIL;
-		uint32 joiningTableCount = 0;
-
-		ExtractRangeTableIndexWalker((Node *) subqueryTree->jointree,
-									 &joinTreeTableIndexList);
-		joiningTableCount = list_length(joinTreeTableIndexList);
-
-		/* if this is a join query, check if join clause is on partition columns */
-		if ((joiningTableCount > 1))
-		{
-			bool joinOnPartitionColumn = JoinOnPartitionColumn(subqueryTree);
-			if (!joinOnPartitionColumn)
-			{
-				preconditionsSatisfied = false;
-				errorDetail = "Relations need to be joining on partition columns";
-			}
-		}
 	}
 
 	/* distinct clause list must include partition column */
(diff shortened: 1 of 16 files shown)


src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -1978,183 +1987,205 @@ BuildJobTreeTaskList(Job *jobTree)
...
-			ShardInterval *shardInterval = (ShardInterval *) lfirst(shardIntervalCell);
+			subqueryTask->jobId = jobId;
+			sqlTaskList = lappend(sqlTaskList, subqueryTask);
+		}
 
-			RangeTableFragment *shardFragment = palloc0(fragmentSize);
-			shardFragment->fragmentReference = shardInterval;
-			shardFragment->fragmentType = CITUS_RTE_RELATION;
-			shardFragment->rangeTableId = tableId;
+		++taskIdIndex;
```
onderkalaci: We probably need to increment the task id inside the `if` statement.

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -1978,183 +1987,205 @@ BuildJobTreeTaskList(Job *jobTree)
...
-			break;
-		}
-	}
+	Oid relationId = 0;
+	int shardCount = 0;
+	int shardOffset = 0;
+	DistTableCacheEntry *targetCacheEntry = NULL;
+	RelationRestrictionContext *relationRestrictionContext =
+		plannerRestrictionContext->relationRestrictionContext;
+	bool allRelationsJoinedOnPartitionKey =
```
metdos: If `allRelationsJoinedOnPartitionKey` is false, we should error out, right? I think we can make this check in multi_logical_optimizer.c with other checks and keep the `SubquerySqlTaskList()` responsible for creating tasks.

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -1978,183 +1987,205 @@ BuildJobTreeTaskList(Job *jobTree)
...
+	 * Note that this has to be done on a copy, as the originals are needed
+	 * per target shard interval.
+	 */
+	foreach(restrictionCell, copiedRestrictionContext->relationRestrictionList)
+	{
+		RelationRestriction *restriction = lfirst(restrictionCell);
+		Index rteIndex = restriction->index;
+		List *originalBaserestrictInfo = restriction->relOptInfo->baserestrictinfo;
+
+		if (!allRelationsJoinedOnPartitionKey || allReferenceTables)
```
metdos: If `allRelationsJoinedOnPartitionKey` is false, we should fail below at `if (!routerPlannable)`, right? 

We also used to tell the reason why we can't push down the query.

```
DETAIL:  Relations need to be joining on partition columns
```

However, if we fail after `RouterSelectQuery ()`, we just provide a generic error message.
```
ERROR:  cannot perform distributed planning for the given query
```

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -1978,183 +1987,205 @@ BuildJobTreeTaskList(Job *jobTree)
...
+	 * or not. If we can, we also rely on the side-effects that all RTEs have been
+	 * updated to point to the relevant nodes and selectPlacementList is determined.
+	 */
+	routerPlannable = RouterSelectQuery(copiedQuery, copiedRestrictionContext,
+										&selectPlacementList, &selectAnchorShardId,
+										&relationShardList, replacePrunedQueryWithDummy);
 
-		sqlTask = CreateBasicTask(jobId, taskIdIndex, SQL_TASK, sqlQueryString->data);
-		sqlTask->dependedTaskList = dataFetchTaskList;
+	if (!routerPlannable)
```
metdos: If we error on `allRelationsJoinedOnPartitionKey`, is there a legitimate path which hits this error message?

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -417,10 +410,10 @@ RouterModifyTaskForShardInterval(Query *originalQuery, ShardInterval *shardInter
 			continue;
 		}
 
-		hashedOpExpressions = ShardIntervalOpExpressions(shardInterval, rteIndex);
-
-		hashedRestrictInfo = make_simple_restrictinfo((Expr *) hashedOpExpressions);
-		originalBaserestrictInfo = lappend(originalBaserestrictInfo, hashedRestrictInfo);
+		shardOpExpressions = ShardIntervalOpExpressions(shardInterval, rteIndex);
+		shardRestrictionList = make_simple_restrictinfo((Expr *) shardOpExpressions);
+		originalBaserestrictInfo = lappend(originalBaserestrictInfo,
```
metdos: If you're renaming variables, could you also rename `originalBaserestrictInfo` to `originalBaseRestrictInfo` :)

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -1978,183 +1987,205 @@ BuildJobTreeTaskList(Job *jobTree)
...
+		if (cacheEntry->partitionMethod == DISTRIBUTE_BY_NONE)
 		{
-			finalShardIntervalList = shardIntervalList;
+			continue;
 		}
 
-		/* if all shards are pruned away, we return an empty task list */
-		finalShardCount = list_length(finalShardIntervalList);
-		if (finalShardCount == 0)
+		if (targetShardIntervalList == NIL)
```
metdos: Do we need this check and `targetShardIntervalList` variable at all?

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -1978,183 +1987,205 @@ BuildJobTreeTaskList(Job *jobTree)
...
 		{
-			return NIL;
+			targetCacheEntry = DistributedTableCacheEntry(relationId);
+			shardCount = targetCacheEntry->shardIntervalArrayLength;
+			break;
 		}
+	}
 
-		fragmentCombinationCell = list_head(fragmentCombinationList);
+	for (shardOffset = 0; shardOffset < shardCount; shardOffset++)
```
metdos: Can we move next line from above to here before `for` loop?
`shardCount = targetCacheEntry->shardIntervalArrayLength;

Also, maybe add an `Assert(targetCacheEntry != NULL)` before it.

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -1978,183 +1987,205 @@ BuildJobTreeTaskList(Job *jobTree)
...
+	{
+		ShardInterval *targetShardInterval =
+			targetCacheEntry->sortedShardIntervalArray[shardOffset];
+		Task *subqueryTask = NULL;
 
-		foreach(shardIntervalCell, finalShardIntervalList)
+		subqueryTask = SubqueryTaskCreate(subquery, targetShardInterval,
+										  allRelationsJoinedOnPartitionKey,
+										  relationRestrictionContext, taskIdIndex);
+
```
metdos: Double empty lines.

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -1978,183 +1987,205 @@ BuildJobTreeTaskList(Job *jobTree)
...
-			{
-				List *fragmentCombination = (List *) lfirst(fragmentCombinationCell);
-				fragmentCombination = lappend(fragmentCombination, shardFragment);
 
-				/* get next fragment for the first relation list */
-				fragmentCombinationCell = lnext(fragmentCombinationCell);
-			}
-		}
+/*
+ * SubqueryTaskCreate creates a modify task by replacing the target
```
metdos: We don't create `modify` tasks here.

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -1978,183 +1987,205 @@ BuildJobTreeTaskList(Job *jobTree)
...
+	ListCell *restrictionCell = NULL;
+	Task *subqueryTask = NULL;
+	List *selectPlacementList = NIL;
+	uint64 selectAnchorShardId = INVALID_SHARD_ID;
+	List *relationShardList = NIL;
+	uint64 jobId = INVALID_JOB_ID;
+	bool routerPlannable = false;
+	bool upsertQuery = false;
+	bool replacePrunedQueryWithDummy = false;
+	bool allReferenceTables = restrictionContext->allReferenceTables;
```
metdos: Can we hit here with `allReferenceTables `? Should not it be a router query?

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -1978,183 +1987,205 @@ BuildJobTreeTaskList(Job *jobTree)
...
+	/*
+	 * Add the restriction qual parameter value in all baserestrictinfos.
+	 * Note that this has to be done on a copy, as the originals are needed
+	 * per target shard interval.
+	 */
+	foreach(restrictionCell, copiedRestrictionContext->relationRestrictionList)
+	{
+		RelationRestriction *restriction = lfirst(restrictionCell);
+		Index rteIndex = restriction->index;
+		List *originalBaserestrictInfo = restriction->relOptInfo->baserestrictinfo;
```
metdos: We modify this list, so maybe just `baseRestrictInfo`?

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -1978,183 +1987,205 @@ BuildJobTreeTaskList(Job *jobTree)
...
+	/* and generate the full query string */
+	deparse_shard_query(copiedQuery, distributedTableId, shardInterval->shardId,
+						queryString);
+	ereport(DEBUG4, (errmsg("distributed statement: %s", queryString->data)));
+
+	subqueryTask = CreateBasicTask(jobId, taskIdIndex, SQL_TASK, queryString->data);
+	subqueryTask->dependedTaskList = NULL;
+	subqueryTask->anchorShardId = selectAnchorShardId;
+	subqueryTask->taskPlacementList = selectPlacementList;
+	subqueryTask->upsertQuery = upsertQuery;
```
metdos: I would set this directly to `false` because we see what we are setting on this line differently from function calls.

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -1978,183 +1987,205 @@ BuildJobTreeTaskList(Job *jobTree)
...
+ * The function errors out if the subquery is not router select query (i.e.,
+ * subqueries with non equi-joins.).
+ */
+static Task *
+SubqueryTaskCreate(Query *originalQuery, ShardInterval *shardInterval,
+				   bool allRelationsJoinedOnPartitionKey,
+				   RelationRestrictionContext *restrictionContext,
+				   uint32 taskIdIndex)
+{
+	Query *copiedQuery = copyObject(originalQuery);
```
metdos: Maybe `taskQuery`?

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -118,8 +121,14 @@ static ArrayType * SplitPointObject(ShardInterval **shardIntervalArray,
...
-static Job * BuildJobTreeTaskList(Job *jobTree);
-static List * SubquerySqlTaskList(Job *job);
+static Job * BuildJobTreeTaskList(Job *jobTree,
+								  PlannerRestrictionContext *plannerRestrictionContext);
+static List * SubquerySqlTaskList(Job *job,
+								  PlannerRestrictionContext *plannerRestrictionContext);
+static Task * SubqueryTaskCreate(Query *originalQuery, ShardInterval *shardInterval,
+								 bool allRelationsJoinedOnPartitionKey,
+								 RelationRestrictionContext *restrictionContext,
+								 uint32 taskIdIndex);
```
metdos: Maybe just `taskId` similar to other places.

src/backend/distributed/planner/multi_physical_planner.c
metdos: I agree on this.

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -1978,183 +1987,205 @@ BuildJobTreeTaskList(Job *jobTree)
 
 
 /*
+ * TODO: update comment
```
onderkalaci: I'll remove this.

src/backend/distributed/planner/multi_physical_planner.c
onderkalaci: I'm switching to passing the restriction to the optimizer. It seems a bit weird to pass the restrictions to the optimizer, but, I agree that we should error out nicely and thus we need it.

src/backend/distributed/planner/multi_physical_planner.c
onderkalaci: Hmm, I couldn't find such a path. But, I'm not enough confident to switch to an Assert. Are you OK with keeping it as an `if`?

metdos: Yes! Then, can we add a comment noting that we keep this as the last security check?

src/backend/distributed/planner/multi_physical_planner.c
metdos: Hmm maybe just `extendedBaseRestrictInfo`, but I'm okay both way.

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -189,9 +189,14 @@ static bool HasOrderByHllType(List *sortClauseList, List *targetList);
  * Third, the function pulls up the collect operators in the tree. Fourth, the
  * function finds the extended operator node, and splits this node into master
  * and worker extended operator nodes.
+ *
+ * We also pass plannerRestrictionContext to the optimizer. The context is not
+ * used for any distributed optimization purposes. Instead, it is used to produce
+ * meaningful error messages for subquery pushdown planning.
```
metdos: Besides `meaningful `, we actually use it to decide if we can push down the subquery. Maybe we can also add this to the comment.

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -1979,182 +1987,201 @@ BuildJobTreeTaskList(Job *jobTree)
...
- * pushdown job. For this, it gets all range tables in the subquery tree, then
- * walks over each range table in the list, gets shards for each range table,
- * and prunes unneeded shards. Then for remaining shards, fragments are created
- * and merged to create fragment combinations. For each created combination, the
- * function builds a SQL task, and appends this task to a task list.
+ * pushdown job. For this, it first checks whether the query is safe to push-down.
+ * In other words, the function first checks whether all the relations that appear
+ * in the query are JOINed on their partition column. Note that the behaviour is
+ * slightly different for reference tables and queries that involve UNIONs. For the
+ * details please see AllRelationsJoinedOnPartitionKey().
```
metdos: I believe we moved this logic to checks in optimizer.

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -1979,182 +1987,201 @@ BuildJobTreeTaskList(Job *jobTree)
...
 	List *sqlTaskList = NIL;
-	List *fragmentCombinationList = NIL;
-	List *opExpressionList = NIL;
-	List *queryList = NIL;
 	List *rangeTableList = NIL;
-	ListCell *fragmentCombinationCell = NULL;
 	ListCell *rangeTableCell = NULL;
-	ListCell *queryCell = NULL;
-	Node *whereClauseTree = NULL;
 	uint32 taskIdIndex = 1; /* 0 is reserved for invalid taskId */
```
metdos: This is old code, but maybe just `taskId`.

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -1979,182 +1987,201 @@ BuildJobTreeTaskList(Job *jobTree)
...
-		opExpressionList = PartitionColumnOpExpressionList(query);
-		if (opExpressionList != NIL)
-		{
-			break;
-		}
-	}
+	/*
+	 * If we're going to create tasks for subquery pushdown, all
+	 * relations needs to be joined on the partition key.
+	 */
```
metdos: Maybe we can put `allRelationsJoinedOnPartitionKey` declaration just before the `Assert`?

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -1979,182 +1987,201 @@ BuildJobTreeTaskList(Job *jobTree)
...
-		}
+	return sqlTaskList;
+}
 
-		rangeTableIndex++;
-	}
+
+/*
+ * SubqueryTaskCreate creates a sql task by replacing the target
+ * shardInterval's boundary value when allRelationsJoinedOnPartitionKey
```
metdos: We don't have `allRelationsJoinedOnPartitionKey ` anymore here.

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -1979,182 +1987,201 @@ BuildJobTreeTaskList(Job *jobTree)
...
+	bool routerPlannable = false;
+	bool replacePrunedQueryWithDummy = false;
+	bool allReferenceTables PG_USED_FOR_ASSERTS_ONLY =
+		restrictionContext->allReferenceTables;
+	RelationRestrictionContext *copiedRestrictionContext =
+		CopyRelationRestrictionContext(restrictionContext);
+	List *shardOpExpressions = NIL;
+	RestrictInfo *shardRestrictionList = NULL;
+
+	Assert(!allReferenceTables);
```
metdos: Maybe just `Assert(!restrictionContext->allReferenceTables)`

src/test/regress/sql/multi_subquery_complex_queries.sql
```diff
@@ -0,0 +1,763 @@
...
+      WHERE "users"."value_2" = "some_recent_users"."user_id"
+        AND value_2 > 75
+      LIMIT 1) "some_users_data" ON TRUE
+   ORDER BY value_2 DESC
+   LIMIT 10) "some_users"
+  order BY value_2 DESC, user_id DESC
+  limit 10;
+
+SET citus.subquery_pushdown TO FALSE;
+SET citus.enable_router_execution TO TRUE;
```
metdos: Need a new line at the end of the file.

metdos:
@onderkalaci these changes look good. As we talked, after merging #1268 and rebasing this to master,  I will do a quick final pass. Then we can merge this to subquery branch.

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -189,9 +189,15 @@ static bool HasOrderByHllType(List *sortClauseList, List *targetList);
  * Third, the function pulls up the collect operators in the tree. Fourth, the
  * function finds the extended operator node, and splits this node into master
  * and worker extended operator nodes.
+ *
+ * We also pass plannerRestrictionContext to the optimizer. The context is not
+ * used for any distributed optimization purposes. Instead, it is used to decide
+ * whether the subquery is safe to pushdown. If not, it helps to produce meaningful
+ * error messages for subquery pushdown planning.
```
anarazel: The "is not used" bit seems quite likely to get out of date.  How about 'is primarily used to decide whether ...'?

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -2823,6 +2831,18 @@ ErrorIfContainsUnsupportedSubquery(MultiNode *logicalPlanNode)
 	/* currently in the planner we only allow one subquery in from-clause*/
 	Assert(list_length(subqueryMultiTableList) == 1);
 
+	allRelationsJoinedOnPartitionKey =
+		AllRelationsJoinedOnPartitionKey(plannerRestrictionContext);
```
anarazel: I still think that this is a misleading name in the presence of reference tables.

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -1979,182 +1987,194 @@ BuildJobTreeTaskList(Job *jobTree)
...
+	}
 
-		sqlTask->anchorShardId = AnchorShardId(fragmentCombination, anchorRangeTableId);
 
-		taskIdIndex++;
-		sqlTaskList = lappend(sqlTaskList, sqlTask);
+	/* ensure that we do not send queries where select is pruned away completely */
+	if (list_length(selectPlacementList) == 0)
+	{
+		ereport(DEBUG2, (errmsg("Skipping target shard interval %ld since "
```
anarazel: Minor message style guideline violation.

src/backend/distributed/planner/multi_logical_optimizer.c
onderkalaci: I'll try to come up with a more accurate one, considering your suggestion as well.

src/backend/distributed/planner/multi_logical_optimizer.c
onderkalaci: Makes sense

src/test/regress/output/multi_subquery.source
```diff
@@ -434,18 +450,390 @@ CREATE TYPE user_composite_type AS
...
 CREATE TYPE user_composite_type AS
 (
 	tenant_id BIGINT,
 	user_id BIGINT
 );
+CREATE FUNCTION user_composite_type_equal(user_composite_type, 
+user_composite_type) 
+returns boolean AS 'select $1.tenant_id = $2.tenant_id AND $1.user_id = $2.user_id;' language sql immutable
+returns NULL ON NULL input; 
+CREATE OR REPLACE FUNCTION user_composite_type_lt(v1 user_composite_type, v2 user_composite_type) returns boolean AS
```
metdos: Why do we create `user_composite_type_lt` twice? Similarly for following functions.

src/test/regress/output/multi_subquery.source
```diff
@@ -590,20 +977,15 @@ FROM
...
 		hasdone) AS subquery_top
 GROUP BY
 	hasdone;
-   event_average    |       hasdone       
---------------------+---------------------
- 4.0000000000000000 | Has not done paying
- 2.5000000000000000 | Has done paying
-(2 rows)
-
+ERROR:  cannot pushdown the subquery since all relations are not joined using distribution keys
```
metdos: Why do we fail on this query?

src/backend/distributed/planner/multi_logical_optimizer.c
onderkalaci: Yes, that's right

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -37,6 +37,9 @@
 #include "distributed/multi_logical_optimizer.h"
 #include "distributed/multi_logical_planner.h"
 #include "distributed/multi_physical_planner.h"
+#include "distributed/multi_planner.h"
+#include "distributed/multi_planner_utils.h"
+#include "distributed/multi_router_planner.h"
```
onderkalaci: Nice catch, removing them.

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -1979,182 +1987,194 @@ BuildJobTreeTaskList(Job *jobTree)
...
-	uint32 rangeTableIndex = 0;
-	const uint32 fragmentSize = sizeof(RangeTableFragment);
-	uint64 largestTableSize = 0;
+	Oid relationId = 0;
+	int shardCount = 0;
+	int shardOffset = 0;
+	DistTableCacheEntry *targetCacheEntry = NULL;
+	RelationRestrictionContext *relationRestrictionContext =
+		plannerRestrictionContext->relationRestrictionContext;
+	bool allRelationsJoinedOnPartitionKey PG_USED_FOR_ASSERTS_ONLY =
```
onderkalaci: I think it is worth keeping this `Assert` here. Because all the assumptions made in this function is based on the result of that function in the logical planner. So, I think it may prevent any future changes to break the assumption here.

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -1979,182 +1987,194 @@ BuildJobTreeTaskList(Job *jobTree)
...
+	/*
+	 * Use router select planner to decide on whether we can push down the query
+	 * or not. If we can, we also rely on the side-effects that all RTEs have been
+	 * updated to point to the relevant nodes and selectPlacementList is determined.
+	 */
+	routerPlannable = RouterSelectQuery(taskQuery, copiedRestrictionContext,
+										&selectPlacementList, &selectAnchorShardId,
+										&relationShardList, replacePrunedQueryWithDummy);
+
+	if (!routerPlannable)
```
onderkalaci: sure, makes sense.

src/test/regress/expected/multi_subquery_complex_queries.out
```diff
@@ -0,0 +1,870 @@
...
+--         GROUP BY "events"."user_id") as t2 on t2.user_id = t.user_id ) as second_query
+--        ON ("first_query".user_id = "second_query".user_id)) as final_query 
+--GROUP BY user_id ORDER BY cnt DESC, user_id DESC LIMIT 10;
+-- Simple LATERAL JOINs with GROUP BYs in each side
+SELECT *
+FROM
+  (SELECT "some_users_data".user_id, lastseen
+   FROM
+     (SELECT user_id,
+             Max(TIME) AS lastseen
```
onderkalaci: nothing special about it. Probably some copy-paste mistake.

src/test/regress/expected/multi_subquery_complex_queries.out
```diff
@@ -0,0 +1,870 @@
...
+      15 | Sat Mar 11 14:43:29.525317 2017
+      15 | Sat Mar 11 14:43:29.525317 2017
+      15 | Sat Mar 11 14:43:29.525317 2017
+      15 | Sat Mar 11 14:43:29.525317 2017
+      15 | Sat Mar 11 14:43:29.525317 2017
+      15 | Sat Mar 11 14:43:29.525317 2017
+      15 | Sat Mar 11 14:43:29.525317 2017
+(10 rows)
+
+-- not supported since the inner JOIN is not equi join
```
onderkalaci: > For failing use-cases, shall we use simpler queries? It would be easier to read regression tests and understand why they are failing.

I had comments on each test for why it is failing.

> I'm ok to keep current ones if updating them will take too much time. We can add an issue to think about them later.

I'm inclined to keep these complex queries as failure tests as well. One would only read the test if it somehow breaks it, so, considering the frequency of happening that, it should be fine to keep these longer tests 😄

src/test/regress/output/multi_subquery.source
onderkalaci: This PR currently breaks `UNION`s. We'll these tests in the upcoming `PR`s. I thought that removing and adding back would be too much pain. Are you ok with this?

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -2134,6 +2131,7 @@ SubqueryTaskCreate(Query *originalQuery, ShardInterval *shardInterval,
 										&selectPlacementList, &selectAnchorShardId,
 										&relationShardList, replacePrunedQueryWithDummy);
 
+	/* we don't expect to this this error but keeping it as a precaution for future changes */
```
metdos: `to this this error but keeping it` -> `this to error but we keep it`

src/test/regress/output/multi_subquery.source
metdos: Yes, that's okay. We will merge this to `subquery_main` branch anyway.

src/backend/distributed/planner/multi_physical_planner.c
metdos: I you feel more comfortable, it's fine :)

src/test/regress/expected/multi_subquery_complex_queries.out
metdos: Maybe, in future, we can revisit these tests, but currently, I'm ok to keep them as they are.

src/test/regress/expected/multi_subquery_complex_queries.out
metdos: Can we convert it to lowercase then :)

src/test/regress/output/multi_subquery.source
metdos: Thinking about this more, we can add `TODO` comments for these regression tests, so we don't miss them later.

metdos:
Finally, shall we create an issue to track regression test updates?
i. Check that we uncommented all regression tests.
ii. Make sure that all the expected tests are run correctly (TODOs).
ii. Adding different time rows to `users_table` and `events_table`. Currently, all rows have same `time` value.

onderkalaci:
>. Check that we uncommented all regression tests.

Done,

i. Make sure that all the expected tests are run correctly (TODOs).

Sure, we always compare with the local tables

ii. Adding different time rows to users_table and events_table. Currently, all rows have same time value.

Should be fixed now. One of the tests results still returns the same time values. But, that should be fine as a special case test

metdos:
> i. Make sure that all the expected tests are run correctly (TODOs).
> 
> Sure, we always compare with the local tables

I meant to check we started to run failing queries correctly, such as;
https://github.com/citusdata/citus/pull/1282/files#diff-b71347b694f44af4eb5eca5e2d1d987a

src/test/regress/output/multi_subquery.source
onderkalaci: It is not possible to forget that since we've already open PR which enables those tests: #1289 
I think to add notes only could lead to git conflicts so I'm refusing to handle this request 😄

src/test/regress/output/multi_subquery.source
onderkalaci: The reason is that we're creating the same functions on different workers and the master.

metdos: Makes sense

src/test/regress/output/multi_subquery.source
metdos: OMG 🗡

metdos:
After Travis, 🚢  to `subquery_main `

Issue #968, opened 2016-11-09
Reference Tables: Foreign Keys


onderkalaci:
Reference tables would have a single shard with replication of (# worker nodes). However, with #698 we considered to not allow foreign keys with replication factor > 1. In that sense, supporting foreign keys for reference tables become tricky. We didn't add this issue to MVP for 6.1.

The main problem with that is a concurrent arbitrary multi-shard modification could run in an order such that the reference table's CASCADE operations are blocked on them and vice versa. See the example below:

1. User runs a DELETE/UPDATE on the `reference_table`
2. The first placement of the `reference_table` gets the DELETE/UPDATE and it cascades the modification to the `shard_1` of  `colocated_table`
3. Then, user runs a concurrent master_modify_multiple_shards (or other type of multi-shard modification) on `colocated_table` on a different connection.
4. The 3rd operation is firstly executed on `shard_2` of `colocated_table`.
5. After that when `reference_table`'s second replica gets the UPDATE/DELETE, it tries to cascade to the `shard_2` of `colocated_table`. However, since the 4th operation holds lock on `shard_2` of `colocated_table`, this operation is blocked.
6. The 3rd operation continues to run on `shard_1` of `colocated_table`. However, since the 2nd operation holds lock on  `shard_1` of `colocated_table`, this operation is blocked.
7. Since both 5 and 6 are blocked, we end-up with a deadlock.

onderkalaci:
Note that we could allow foreign keys among reference tables. The reason is that they'd be single sharded and the above case does not apply for a single shard distributed table.

I'll also double check this with @marcocitus 

This is also semi-related to #962.

marcocitus:
> Note that we could allow foreign keys among reference tables. The reason is that they'd be single sharded and the above case does not apply for a single shard distributed table.

The replication factor issues would still apply, e.g. #861 (also need to take appropriate shard resource locks for cascading delete/update).

sumedhpathak:
Noting example from a contact from our docs repo:

I ran into a related issue with the next table:

```sql
CREATE TABLE my_table (
	id SERIAL NOT NULL,
	name VARCHAR,
	PRIMARY KEY (id),
	UNIQUE (name, id)
);
SELECT create_reference_table('my_table');

CREATE TABLE my_second_table (
	id SERIAL NOT NULL,
	name VARCHAR,
    my_table_id SERIAL NOT NULL,
	PRIMARY KEY (id, my_table_id),
[...]

PR #1283, opened 2017-03-16
Decouple reference table replication

### LINKED ISSUE #1200: Decouple replication of reference tables from master_add_node
We had an internal meeting related to decoupling the replication of reference tables from ``master_add_node()``. We decided to do the following:

* We're revising our upgrade policy with 6.2 onwards. Starting with this release, we'll add columns to our metadata tables (system catalogs) and force users to Alter Extension Upgrade.
* Since we can now add a column to metadata tables, we can implement this UDF in a more intuitive way (Marco's email below). We add this column and implement the UDF accordingly right after the 6.1 release. (Uninformed minor question: We will have two boolean columns to indicate status -- are these better represented as enum fields?)
* @fdr noted that he could do one of two things: (1) pick up a 6.1-cloud release that has the above system catalog changes or (2) wait until 6.2 anticipating that reference tables won't be a big problem. The first one is more ideal and we'll likely do a small release called 6.1-cloud for the Cloud.

--

I'm also copy/pasting @marcocitus' email on how we decouple metadata propagation from adding nodes in Citus MX.

Arguing with myself: While there might not be a clean way, it is pretty much necessary to do add_node before the node is running for things like auto-HA. e.g. in the original MX prototype we would first
add a node to decide whether it would be primary/secondary and in which group *before* we could start the node. In that case, there's no way we could replicate the reference tables to that node on
master_add_node or use it for anything else.

In MX there was a second step called cluster_activate_node. A node would not be used for anything before it was activated (reflected by an isactive column in pg_dist_node), and when it got activated it also got a copy of the metadata. In Citus 6.1, we opted for decoupling metadata syncing and node activation, which allowed us to make MX optional and simplified pg_dist_node syncing.

In this case, it seems that what we would really want is for the activate step to come back in order to replicate reference tables before we use a node. While a node is inactive, we can assume it to be
not yet up-and-running and would not assign any new shards to it. This avoids any situation where reference tables are under-replicated. At some point, we will also want to replicate global objects such as roles, schemas, and functions upon node activation. We probably don't want to change the API again in the next release, so it might be a mistake to make it very reference table-specific.

Strictly speaking, we would need a new column in pg_dist_node (isactive), but we've more or less decided not to do metadata changes in 6.1. With our current feature set, we could live with a situation
in which reference tables are temporarily under-replicated so we could add that later.

byucesoy:
Fixes #1200 

With this change we add an option to add a node without replicating all reference tables to that node. If a node is added with this option, we mark the node as inactive and no queries will sent to that node.

We also added two new UDFs;
 - master_activate_node(host, port):
    - marks node as active and replicates all reference tables to that node
 - master_add_inactive_node(host, port):
    - only adds node to pg_dist_node

### DIFF
--- a/src/backend/distributed/metadata/metadata_sync.c
+++ b/src/backend/distributed/metadata/metadata_sync.c
@@ -99,6 +99,15 @@ start_metadata_sync_to_node(PG_FUNCTION_ARGS)
 								"(%s,%d)", escapedNodeName, nodePort)));
 	}
 
+	if (!workerNode->isActive)
+	{
+		ereport(ERROR, (errcode(ERRCODE_OBJECT_NOT_IN_PREREQUISITE_STATE),
+						errmsg("you cannot sync metadata to an inactive node"),
+						errhint("First, activate the node with "
+								"SELECT master_activate_node(%s,%d)",
+								escapedNodeName, nodePort)));
+	}
+
 	MarkNodeHasMetadata(nodeNameString, nodePort, true);
 
 	/* generate and add the local group id's update query */
@@ -206,7 +215,7 @@ MetadataCreateCommands(void)
 	List *metadataSnapshotCommandList = NIL;
 	List *distributedTableList = DistributedTableList();
 	List *propagatedTableList = NIL;
-	List *workerNodeList = WorkerNodeList();
+	List *workerNodeList = ActiveWorkerNodeList();
 	ListCell *distributedTableCell = NULL;
 	char *nodeListInsertCommand = NULL;
 	bool includeSequenceDefaults = true;
@@ -401,24 +410,25 @@ NodeListInsertCommand(List *workerNodeList)
 
 	/* generate the query without any values yet */
 	appendStringInfo(nodeListInsertCommand,
-					 "INSERT INTO pg_dist_node "
-					 "(nodeid, groupid, nodename, nodeport, noderack, hasmetadata) "
-					 "VALUES ");
+					 "INSERT INTO pg_dist_node (nodeid, groupid, nodename, nodeport, "
+					 "noderack, hasmetadata, isactive) VALUES ");
 
 	/* iterate over the worker nodes, add the values */
 	foreach(workerNodeCell, workerNodeList)
 	{
 		WorkerNode *workerNode = (WorkerNode *) lfirst(workerNodeCell);
-		char *hasMetadaString = workerNode->hasMetadata ? "TRUE" : "FALSE";
+		char *hasMetadataString = workerNode->hasMetadata ? "TRUE" : "FALSE";
+		char *isActiveString = workerNode->isActive ? "TRUE" : "FALSE";
 
 		appendStringInfo(nodeListInsertCommand,
-						 "(%d, %d, %s, %d, %s, %s)",
+						 "(%d, %d, %s, %d, %s, %s, %s)",
 						 workerNode->nodeId,
 						 workerNode->groupId,
 						 quote_literal_cstr(workerNode->workerName),
 						 workerNode->workerPort,
 						 quote_literal_cstr(workerNode->workerRack),
-						 hasMetadaString);
+						 hasMetadataString,
+						 isActiveString);
 
 		processedWorkerNodeCount++;
 		if (processedWorkerNodeCount != workerCount)
@@ -682,6 +692,24 @@ NodeDeleteCommand(uint32 nodeId)
 					 "WHERE nodeid = %u", nodeId);
 
 	return nodeDeleteCommand->data;
+}
+
+
+/*
+ * NodeStateUpdateCommand generates a command that can be executed to update
+ * isactive column of a node in pg_dist_node table.
+ */
+char *
+NodeStateUpdateCommand(uint32 nodeId, bool isActive)
+{
+	StringInfo nodeStateUpdateCommand = makeStringInfo();
+	char *isActiveString = isActive ? "TRUE" : "FALSE";
+
+	appendStringInfo(nodeStateUpdateCommand,
+					 "UPDATE pg_dist_node SET isactive = %s "
+					 "WHERE nodeid = %u", isActiveString, nodeId);
+
+	return nodeStateUpdateCommand->data;
 }
 
 
@@ -973,7 +1001,7 @@ OwnerName(Oid objectId)
 static bool
 HasMetadataWorkers(void)
 {
-	List *workerNodeList = WorkerNodeList();
+	List *workerNodeList = ActiveWorkerNodeList();
 	ListCell *workerNodeCell = NULL;
 
 	foreach(workerNodeCell, workerNodeList)
--- a/src/backend/distributed/utils/node_metadata.c
+++ b/src/backend/distributed/utils/node_metadata.c
@@ -33,10 +33,12 @@
 #include "distributed/multi_router_planner.h"
 #include "distributed/pg_dist_node.h"
 #include "distributed/reference_table_utils.h"
+#include "distributed/resource_lock.h"
 #include "distributed/shardinterval_utils.h"
 #include "distributed/worker_manager.h"
 #include "distributed/worker_transaction.h"
 #include "lib/stringinfo.h"
+#include "storage/bufmgr.h"
 #include "storage/lock.h"
 #include "storage/fd.h"
 #include "utils/builtins.h"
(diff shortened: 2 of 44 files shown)


metdos:
@byucesoy is #1200 still relevant? If not could you close that one and add a note referring this PR?

Could you also add the issue to the PR description?

metdos:
@byucesoy, regression tests are still failing? What is the problem?

src/backend/distributed/utils/resource_lock.c
```diff
@@ -342,3 +342,42 @@ LockMetadataSnapshot(LOCKMODE lockMode)
...
 }
+
+
+/*
+ * LockNodeId returns after grabbing a lock for distribution metadata related
+ * to the specified shard, blocking if required. Any locks acquired using this
+ * method are released at transaction end.
+ */
+void
+LockNodeId(int64 nodeId, LOCKMODE lockMode)
```
jasonmp85: I don't like that we have to add a whole new lock tag just for this one pathway. I believe you could get away with simply locking the row in `pg_dist_node`. In particular, you error out if the node can't be found, so you only proceed to locking if the row exists. This should have identical semantics to something like `SELECT ... FOR UPDATE`, so could probably be implemented by simply finding the row and retaining an exclusive lock on it to the end of the operation. If the row isn't found, error out (see the call site to `TryLockNodeId`, etc.)

src/backend/distributed/utils/node_metadata.c
```diff
@@ -120,32 +150,104 @@ master_remove_node(PG_FUNCTION_ARGS)
...
 	text *nodeName = PG_GETARG_TEXT_P(0);
 	int32 nodePort = PG_GETARG_INT32(1);
+
 	char *nodeNameString = text_to_cstring(nodeName);
-	bool forceRemove = true;
-	RemoveNodeFromCluster(nodeNameString, nodePort, forceRemove);
+	Datum nodeRecord = 0;
 
-	PG_RETURN_VOID();
+	nodeRecord = ActivateNode(nodeNameString, nodePort);
```
byucesoy: I guess `nodeid` would be useful when we implement #968. I am not sure if they have non-replacable usage areas right now.

src/backend/distributed/utils/node_metadata.c
```diff
@@ -120,32 +150,104 @@ master_remove_node(PG_FUNCTION_ARGS)
...
+	DeleteAllReferenceTablePlacementsFromNode(nodeName, nodePort);
+	SetNodeState(nodeName, nodePort, NODE_STATE_INACTIVE);
+
+	hasShardPlacements = NodeHasActiveShardPlacements(nodeName, nodePort);
+
+	if (hasShardPlacements)
+	{
+		ereport(NOTICE, (errmsg("Node %s:%d has active shard placements. Some queries "
+								"may fail after this operation. Use "
+								"SELECT master_activate_node('%s', %d) to activate this "
```
byucesoy: Well, this is existing message and behavior master_disable_node. Its behavior is kind of strange, I think it was implemented to emulate removing node from pg_worker_list file. This prevents new shard creation on the node, but it still gets some queries like INSERTs.

I agree the behavior is kind of strange and the suggestion is not that good, but I try to not change them. Do you have a suggestion about it?

src/backend/distributed/utils/node_metadata.c
```diff
@@ -120,32 +150,104 @@ master_remove_node(PG_FUNCTION_ARGS)
...
- * the master node and all nodes with metadata regardless of the node having an active
- * shard placement.
- * The call to the master_remove_node should be done by the super user.
- * This function also deletes all reference table placements belong to the given node from
- * pg_dist_shard_placement, but it does not drop actual placement at the node. In the case
- * of re-adding the node, master_add_node first drops and re-creates the reference tables.
+ * master_disable_node function sets nodestate of the provided node to inactive
+ * at master node and all nodes with metadata regardless of the node having an
+ * active shard placement.
+ * The call to the master_disable_node should be done by the super user.
```
byucesoy: Fixed

src/backend/distributed/utils/node_metadata.c
```diff
@@ -87,21 +93,45 @@ master_add_node(PG_FUNCTION_ARGS)
...
+		nodeRecord = ActivateNode(nodeNameString, nodePort);
 	}
 
-	PG_RETURN_CSTRING(returnData);
+	PG_RETURN_CSTRING(nodeRecord);
+}
+
+
+/*
+ * master_prepare_node function adds a new node to the cluster as inactive node
```
byucesoy: This name come up in the e-mail discussion of the change. I also suggested `master_add_inactive_node` and `master_add(_inactive)_node_to_metadata` but we decided not use them because we use inactive and metadata terms in different places with slightly different meanings.

master_prepare_node name come from Ozgun, but I think he does not have a strong preference either. Do you have a suggestion in this matter?

Quote from mail discussion
> Both master_add_inactive_node() and master_add_(inactive)_metadata() describe our behavior in terms of other api names: add, in-active, and metadata. Rather than use mix of words that have other meanings, we use a different word for now. Say, master_prepare_node()?

src/backend/distributed/utils/node_metadata.c
```diff
@@ -87,21 +93,45 @@ master_add_node(PG_FUNCTION_ARGS)
...
+	Datum nodeRecord = AddNodeMetadata(nodeNameString, nodePort, groupId, nodeRack,
+									   hasMetadata, NODE_STATE_INACTIVE,
+									   &nodeAlreadyExists);
 
 	/*
-	 * After adding new node, if the node is not already exist, we  replicate all existing
-	 * reference tables to the new node. ReplicateAllReferenceTablesToAllNodes replicates
-	 * reference tables to all nodes however, it skips nodes which already has healthy
-	 * placement of particular reference table.
+	 * After adding new node, if the node is not already exist, we will activate
```
byucesoy: Fixed

src/backend/distributed/metadata/metadata_sync.c
```diff
@@ -686,6 +695,23 @@ NodeDeleteCommand(uint32 nodeId)
 
 
 /*
+ * NodeStateUpdateCommand generates a command that can be executed to update
+ * nodestate of a node.
+ */
+char *
+NodeStateUpdateCommand(uint32 nodeId, uint32 nodeState)
+{
+	StringInfo NodeStateUpdateCommand = makeStringInfo();
```
byucesoy: Fixed

byucesoy:
Hi @jasonmp85 

I fixed smallish issues you mentioned, did not fixed locking issue yet. Also asked you a few question about your comments. I will continue to work on this PR tommorrow.

src/backend/distributed/utils/node_metadata.c
```diff
@@ -120,32 +150,104 @@ master_remove_node(PG_FUNCTION_ARGS)
...
+	bool lockAcquired = false;
+	Datum nodeRecord = 0;
+
+	if (workerNode == NULL)
+	{
+		ereport(ERROR, (errcode(ERRCODE_OBJECT_NOT_IN_PREREQUISITE_STATE),
+						errmsg("cannot activate non-existing node")));
+	}
+
+	lockAcquired = TryLockNodeId(workerNode->nodeId, ExclusiveLock);
```
byucesoy: I quickly scan our codebase but did not find a place where we take a lock on a row. Do you know a particular place? If not, tomorrow I will debug `SELECT ... FOR UPDATE`.

byucesoy:
Hey @jasonmp85 

I removed new lock mechanism, instead now we take look on the respective row of pg_dist_node with heap_lock_tuple function.

src/backend/distributed/utils/node_metadata.c
jasonmp85: Chatted with Ozgun last week and he was OK with changing back to `master_add_inactive_node`  if we're OK with it ("if you and Burak are good with it, I'm good with it"). I said it was more intuitive to me that `master_add_inactive_node` + `master_activate_node` = `master_add_node`.

I'd like to change `master_prepare_node` to `mamaster_add_inactive_node`; can we go ahead with that?

src/backend/distributed/utils/node_metadata.c
jasonmp85: Got it; leave it as-is, then.

src/backend/distributed/utils/node_metadata.c
jasonmp85: Sure; makes sense.

jasonmp85:
If you agree that it's more intuitive that `master_add_inactive_node` + `master_activate_node` = `master_add_node`, then change `master_prepare_node` to `master_add_inactive_node`. I think it makes sense to say "add an inactive node, then activate it" rather than using the word _prepare_, which is new and requires definition (especially since it's already used a lot in PostgreSQL).

After that change (assuming you want to make it), :shipit:!

byucesoy:
Yes, I also think master_add_inactive_node is more descriptive and intuitive. I renamed the UDF.

Issue #1007, opened 2016-12-01
CREATE INDEX CONCURRENTLY is not supported for distributed tables.


saicitus:
Encountered this in 2 customer conversations. The first user uses psycopg2 as their ORM and the second uses Rails.

SQL Query which is used:
```
CREATE INDEX CONCURRENTLY
IF NOT EXISTS {table_name}_idx_a_b_c_d ON {table_name} (a, b, c, d) WHERE e='test'
```

anarazel:
@saicitus I think it'd be useful to use labels to indicate whether you think something is a bug or a feature request.

lfittl:
Just adding a note that we should also support `DROP INDEX CONCURRENTLY`, for easy trial and error for new indices.

saicitus:
I wanted to drop a note here, in last ~2 weeks, came up in 3 customer/prospect engagements.

ozgune:
We talked about this feature during our 6.2 product planning. We noted that fixing #357 also helps notably with this issue. We therefore decided to prioritize #357 together with (as a prerequisite to) this issue.

jasonmp85:
As @ozgune points out, #357 is what we're tackling first (PR today), but that's basically an immediate stepping-stone to this (and `DROP` support).

We had a meeting about this approach last week and went over the relevant pieces, both from tech feasibility and from user experience standpoints.

Because the `CONCURRENTLY` options to `CREATE`/`DROP INDEX` prohibit use within a transaction, it is possible some nodes successfully apply the command while others fail. Rather than mark failed nodes as bad (hinted at in #357), we will emit a message to the user that the index is in an incomplete state. This is analogous to the failure mode of single-node PostgreSQL, which can itself fail to create an index using `CONCURRENTLY`.

Though it can't be performed transactionally, the race condition window is small enough that it might be worthwhile to edit the catalog to mark the created index as `INVALID`, so that `psql` inspection commands (`\d` and the like) accurate reflect the partial nature of a failed index.

To back out of such a situation, users should drop and recreate the index (or just retry `DROP` in the drop case), with appropriate `IF (NOT) EXISTS` modifiers.

The work here is relatively straightforward after #357 is complete and mostly consists of verifying the worker application command's functionality on the remotes.

PR #1287, opened 2017-03-22
Support (CREATE|DROP) INDEX CONCURRENTLY

jasonmp85:
Fixes: #1007

The changes here are as follow:

  * Modify `IndexStmt` and `DropStmt` to be deparsed on the master rather than a worker. Calling `worker_apply_shard_ddl_command` is not possible as it triggers a transaction and `CONCURRENTLY` statements must run outside a transaction
  * Remove checks preventing execution of `CONCURRENTLY` statement
  * Add flag to `DDLJob` to denote when transactions are prohibited
  * Add mode to force `Task`s to run serially rather than in parallel. `CREATE INDEX CONCURRENTLY` statements conflict with one another, _even on different tables_
  * Fix existing tests

# TODO

  - [x] Documentation pass
  - [x] Add back my failure testing
  - [x] Add slightly nicer error messaging
  - [x] (Possibly) mark index as invalid if any remote commands fail
  - [x] ~~Remove `IndextStmt` and `DropStmt` support from `RelayEventExtend`, etc.~~
  - [x] MX fix

### DIFF
--- a/src/backend/distributed/executor/multi_router_executor.c
+++ b/src/backend/distributed/executor/multi_router_executor.c
@@ -870,13 +870,13 @@ ExecuteModifyTasksWithoutResults(List *taskList)
 
 
 /*
- * ExecuteSequentialTasksWithoutResults basically calls ExecuteModifyTasks in a
- * loop in order to simulate sequential execution of a list of tasks. Useful in
- * cases where issuing commands in parallel before waiting for results could
+ * ExecuteTasksSequentiallyWithoutResults basically calls ExecuteModifyTasks in
+ * a loop in order to simulate sequential execution of a list of tasks. Useful
+ * in cases where issuing commands in parallel before waiting for results could
  * result in deadlocks (such as CREATE INDEX CONCURRENTLY).
  */
-int64
-ExecuteSequentialTasksWithoutResults(List *taskList)
+void
+ExecuteTasksSequentiallyWithoutResults(List *taskList)
 {
 	ListCell *taskCell = NULL;
 
@@ -885,10 +885,8 @@ ExecuteSequentialTasksWithoutResults(List *taskList)
 		Task *task = (Task *) lfirst(taskCell);
 		List *singleTask = list_make1(task);
 
-		ExecuteModifyTasks(singleTask, false, NULL, NULL);
+		ExecuteModifyTasksWithoutResults(singleTask);
 	}
-
-	return 0;
 }
 
 
--- a/src/backend/distributed/executor/multi_router_executor.c
+++ b/src/backend/distributed/executor/multi_router_executor.c
@@ -869,6 +869,23 @@ ExecuteModifyTasksWithoutResults(List *taskList)
 }
 
 
+int64
+ExecuteSequentialTasksWithoutResults(List *taskList)
+{
+	ListCell *taskCell = NULL;
+
+	foreach(taskCell, taskList)
+	{
+		Task *task = (Task *) lfirst(taskCell);
+		List *singleTask = list_make1(task);
+
+		ExecuteModifyTasks(singleTask, false, NULL, NULL);
+	}
+
+	return 0;
+}
+
+
 /*
  * ExecuteModifyTasks executes a list of tasks on remote nodes, and
  * optionally retrieves the results and stores them in a tuple store.
--- a/src/backend/distributed/executor/multi_router_executor.c
+++ b/src/backend/distributed/executor/multi_router_executor.c
@@ -869,6 +869,12 @@ ExecuteModifyTasksWithoutResults(List *taskList)
 }
 
 
+/*
+ * ExecuteSequentialTasksWithoutResults basically calls ExecuteModifyTasks in a
+ * loop in order to simulate sequential execution of a list of tasks. Useful in
+ * cases where issuing commands in parallel before waiting for results could
+ * result in deadlocks (such as CREATE INDEX CONCURRENTLY).
+ */
 int64
 ExecuteSequentialTasksWithoutResults(List *taskList)
 {
--- a/src/backend/distributed/executor/multi_utility.c
+++ b/src/backend/distributed/executor/multi_utility.c
@@ -684,7 +684,7 @@ PlanIndexStmt(IndexStmt *createIndexStatement, const char *createIndexCommand)
 			{
 				DDLJob *ddlJob = palloc0(sizeof(DDLJob));
 				ddlJob->targetRelationId = relationId;
-				ddlJob->preventTransaction = createIndexStatement->concurrent;
+				ddlJob->concurrent = createIndexStatement->concurrent;
 				ddlJob->commandString = createIndexCommand;
 				ddlJob->taskList = IndexTaskList(relationId, createIndexStatement);
 
@@ -778,7 +778,7 @@ PlanDropIndexStmt(DropStmt *dropIndexStatement, const char *dropIndexCommand)
 		ErrorIfUnsupportedDropIndexStmt(dropIndexStatement);
 
 		ddlJob->targetRelationId = distributedRelationId;
-		ddlJob->preventTransaction = dropIndexStatement->concurrent;
+		ddlJob->concurrent = dropIndexStatement->concurrent;
 		ddlJob->commandString = dropIndexCommand;
 		ddlJob->taskList = DropIndexTaskList(distributedRelationId, distributedIndexId,
 											 dropIndexStatement);
@@ -873,7 +873,7 @@ PlanAlterTableStmt(AlterTableStmt *alterTableStatement, const char *alterTableCo
 
 	ddlJob = palloc0(sizeof(DDLJob));
 	ddlJob->targetRelationId = leftRelationId;
-	ddlJob->preventTransaction = false;
+	ddlJob->concurrent = false;
 	ddlJob->commandString = alterTableCommand;
 
 	if (rightRelationId)
@@ -1987,24 +1987,7 @@ ExecuteDistributedDDLJob(DDLJob *ddlJob)
 
 	EnsureCoordinator();
 
-	if (ddlJob->preventTransaction)
-	{
-		/* save old commit protocol to restore at xact end */
-		Assert(SavedMultiShardCommitProtocol == COMMIT_PROTOCOL_BARE);
-		SavedMultiShardCommitProtocol = MultiShardCommitProtocol;
-		MultiShardCommitProtocol = COMMIT_PROTOCOL_BARE;
-
-		if (shouldSyncMetadata)
-		{
-			List *commandList = list_make2(DISABLE_DDL_PROPAGATION,
-										   (char *) ddlJob->commandString);
-
-			SendBareCommandListToWorkers(WORKERS_WITH_METADATA, commandList);
-		}
-
-		ExecuteSequentialTasksWithoutResults(ddlJob->taskList);
-	}
-	else
+	if (!ddlJob->concurrent)
 	{
 		ShowNoticeIfNotUsing2PC();
 
@@ -2015,6 +1998,36 @@ ExecuteDistributedDDLJob(DDLJob *ddlJob)
 		}
 
 		ExecuteModifyTasksWithoutResults(ddlJob->taskList);
+	}
+	else
+	{
+		/* save old commit protocol to restore at xact end */
+		Assert(SavedMultiShardCommitProtocol == COMMIT_PROTOCOL_BARE);
+		SavedMultiShardCommitProtocol = MultiShardCommitProtocol;
+		MultiShardCommitProtocol = COMMIT_PROTOCOL_BARE;
+
+		PG_TRY();
+		{
+			if (shouldSyncMetadata)
+			{
+				List *commandList = list_make2(DISABLE_DDL_PROPAGATION,
+											   (char *) ddlJob->commandString);
+
+				SendBareCommandListToWorkers(WORKERS_WITH_METADATA, commandList);
+			}
+
+			ExecuteSequentialTasksWithoutResults(ddlJob->taskList);
+		}
+		PG_CATCH();
+		{
+			ereport(ERROR,
+					(errmsg("CONCURRENTLY-enabled index command failed"),
+					 errdetail("CONCURRENTLY-enabled index commands can fail partially, "
+							   "leaving behind an INVALID index."),
+					 errhint("Use DROP INDEX IF EXISTS to remove the invalid index, then "
+							 "retry the original command.")));
+		}
+		PG_END_TRY();
 	}
 }
 
@@ -2701,7 +2714,7 @@ PlanGrantStmt(GrantStmt *grantStmt)
 
 		ddlJob = palloc0(sizeof(DDLJob));
 		ddlJob->targetRelationId = relOid;
-		ddlJob->preventTransaction = false;
+		ddlJob->concurrent = false;
 		ddlJob->commandString = pstrdup(ddlString.data);
 		ddlJob->taskList = DDLTaskList(relOid, ddlString.data);
 
--- a/src/backend/distributed/executor/multi_utility.c
+++ b/src/backend/distributed/executor/multi_utility.c
@@ -77,7 +77,6 @@
 #include "utils/palloc.h"
 #include "utils/rel.h"
 #include "utils/relcache.h"
-#include "utils/snapmgr.h"
 #include "utils/syscache.h"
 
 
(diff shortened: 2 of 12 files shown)


metdos:
> CREATE INDEX CONCURRENTLY statements conflict with one another, even on different tables

This is interesting. Do you know the reason?

src/backend/distributed/executor/multi_utility.c
```diff
@@ -677,8 +679,9 @@ PlanIndexStmt(IndexStmt *createIndexStatement, const char *createIndexCommand)
 			{
 				DDLJob *ddlJob = palloc0(sizeof(DDLJob));
 				ddlJob->targetRelationId = relationId;
+				ddlJob->preventTransaction = createIndexStatement->concurrent;
 				ddlJob->commandString = createIndexCommand;
-				ddlJob->taskList = DDLTaskList(relationId, createIndexCommand);
+				ddlJob->taskList = IndexTaskList(relationId, createIndexStatement);
```
metdos: Would it be better to continue using worker_apply_shard_ddl_command() for regular index commands given that worker_apply_shard_ddl_command() is more native(?) and more tested?

```C
				if (createIndexStatement->concurrent)
				{
					ddlJob->taskList = IndexTaskList(relationId, createIndexStatement);
				} else {
					ddlJob->taskList = DDLTaskList(relationId, createIndexCommand);
				}
```

metdos:
CTRL + C stops "create index concurrently" immediately. We see the master table with a successful index but some of the shards on workers don't get the index yet.

metdos:
> (Possibly) mark index as invalid if any remote commands fail

This makes sense.

metdos:
Currently, this fails on synchronized MX tables because we sync metadata in a transaction via SendCommandToWorkers().

```sql
postgres=# create index concurrently index_6 on lineitem (l_orderkey);
ERROR:  CREATE INDEX CONCURRENTLY cannot run inside a transaction block
```

One option is to send metadata commands for "create index concurrently" to workers without a transaction as you did for shards, however, the failure semantics will be complex.

I think it is ok to not support this for MX tables and to throw a more meaningful error message. However, we should check this with @ozgune first.

jasonmp85:
Today I was mostly working on checking out the MX finding you posted. Replied in email to you, @ozgune, @sumedhpathak, and @marcocitus. Probably going to disable within MX for now.

You seemed to like the idea of marking the index as `invalid` if something goes wrong. That's still not perfect, since we can never have all the transactional guarantees we might like, but I think I'll add that back tomorrow.

Re `worker_apply_shard_ddl_command`, I actually think its support for `DropStmt` and `IndexStmt` should be removed. Here's my reasoning:

  * `CONCURRENTLY` doesn't prohibit any other flags, so to implement `CONCURRENTLY` using deparsing, I also had to support _all_ other flags
  * Because I support all flags, we'd have one path of code for non-`CONCURRENTLY`, and an entirely different path for `CONCURRENTLY` statements. This seems bad, since it doubles the bug surface for index features
  * I'm not even sold that `worker_apply_shard_ddl_command` is more "native", as I'd argue it's far more likely the parsenode representation changes than the SQL representation, which is pretty stable (and standardized)

So I think I'm leaning to switching to `deparse`-style for all index commands. The only open question is "what if an older coordinator node tries to call `worker_apply_shard_ddl_command` on a worker node running a newer Citus version?" IMO this is an unsupported configuration, so probably not something to worry about, but it was the only thing that occurred to me when I considered removing `IndexStmt` and `DropStmt` from `worker_apply_shard_ddl_command`.

src/backend/distributed/executor/multi_utility.c
jasonmp85: Answered in a top-level comment.

jasonmp85:
> This is interesting. Do you know the reason? 

It was _super_ hard to Google for this actually, but it turns out that `CONCURRENTLY` index creation takes place in two or three phases (see the code for more detail). Basically it creates the index, but in a state where it can't be used yet, commits, then starts building it, but doesn't allow it for certain operations yet, then commits, then flips on all operations. The docs say:

> When this option is used, PostgreSQL must perform two scans of the table, and in addition it must wait for all existing transactions that could potentially modify or use the index to terminate.

_all existing transactions that could potentially modify or use the index_. If you think about it for a sec, that actually means _all transactions_, since PostgreSQL has no clue what open transactions might do in their next command. So (I think?) it grabs `ShareLocks` on the `virtualxid` locks of all open transactions (!). This lock type isn't used for too much else (which is why Googling was hard), but transactions hold `ExclusiveLock`s on their own `virtualxid`. So a `CONCURRENTLY` creation will wait for all other transactions to complete, including other `CONCURRENTLY`s. So if I issue the commands in order, why do they get cycles?

> In a concurrent index build, the index is actually entered into the system catalogs in one transaction, then two table scans occur in two more transactions. Before each table scan, the index build must wait for existing transactions that have modified the table to terminate. After the second scan, the index build must wait for any transactions that have a snapshot (see Chapter 13) predating the second scan to terminate. Then finally the index can be marked ready for use, and the CREATE INDEX command terminates. Even then, however, the index may not be immediately usable for queries: in the worst case, it cannot be used as long as transactions exist that predate the start of the index build.

"the index build must wait for any transactions that have a snapshot predating the second scan". I think that's the phase where concurrent `CONCURRENTLY` deadlocks occur.

In short, I had been thinking of `CONCURRENTLY` as "faster" for some reason, but it's not. It has better operational characteristics (namely that `INSERT`s/`UPDATE`s, etc. can proceed), but can take much longer due to the multi-step lock acquisition. This will be compounded by the requirement that we issue the shard index creations serially.

jasonmp85:
@metdos you can read a thread about `CREATE INDEX CONCURRENTLY` locking here: http://www.postgresql-archive.org/Simultaneous-index-creates-on-different-schemas-cause-deadlock-td5753045.html

It supposedly was relaxed by https://git.postgresql.org/gitweb/?p=postgresql.git;a=commit;h=c3d09b3bd23f5f65b5eb8124a3c7592dad85a50c but I'm not seeing that in practice.

metdos:
> Today I was mostly working on checking out the MX finding you posted. Replied in email to you, @ozgune, @sumedhpathak, and @marcocitus. Probably going to disable within MX for now.

I think we should be fine to disable within MX for now. @ozgune, any objections on the product side?
> 
> You seemed to like the idea of marking the index as invalid if something goes wrong. That's still not perfect, since we can never have all the transactional guarantees we might like, but I think I'll add that back tomorrow.

My main concern is that, as a user, how will I know that my index is created completely on all shards? In the case of an error or "CTRL + C", PostgreSQL marks the index INVALID. Therefore, if we can't mark the index as INVALID, we should document this very well.

Maybe, we can also provide a udf to check the health of the index such as `verify_index_is_complete(index_name)`

> So I think I'm leaning to switching to deparse-style for all index commands.

Okay, your arguments make sense.

> The only open question is "what if an older coordinator node tries to call worker_apply_shard_ddl_command on a worker node running a newer Citus version?" IMO this is an unsupported configuration, so probably not something to worry about, but it was the only thing that occurred to me when I considered removing IndexStmt and DropStmt from worker_apply_shard_ddl_command.

Instead of removing them now, can we add depreciation comments and remove them later? This would be a problem if we start a rolling upgrade in the cluster from worker nodes.

marcocitus:
> I think we should be fine to disable within MX for now. @ozgune, any objections on the product side?

By "disabling" we mean skipping the SendCommandToWorkers, right?

metdos:
> By "disabling" we mean skipping the SendCommandToWorkers, right?

I was thinking erroring out, but skipping the SendCommandToWorkers and documenting the behavior for "create index concurrently" with MX is also fine.

metdos:
> @metdos you can read a thread about CREATE INDEX CONCURRENTLY locking here: http://www.postgresql-archive.org/Simultaneous-index-creates-on-different-schemas-cause-deadlock-td5753045.html

Thanks for pointing out this. I like the comment of @anarazel noting that "... especially as nobody seems to have complained so far other than the OP ...". I think we will second the complain :)

> It supposedly was relaxed by https://git.postgresql.org/gitweb/?p=postgresql.git;a=commit;h=c3d09b3bd23f5f65b5eb8124a3c7592dad85a50c but I'm not seeing that in practice.

For PostgreSQL, I can easily replicate the deadlock on the same table (running index creation queries concurrently on the same table) but not on different ones (running index creation queries concurrently on different tables)  - though we can replicate the deadlock on different tables with your changes as I noted in my other comment.

For Citus, I couldn't replicate the deadlock on the master with distributed tables, either on the same table or on different tables.

jasonmp85:
@metdos — Updated with many of the discussed changes. Couldn't outright remove `RelayEventExtendShardNames` or whatever because it's still called by the machinery that adds shards to existing tables.

Fixed MX, added MX tests (and noticed/fixed an MX test error), fixed up error messages, added partial failure tests, added `INVALID` marking, added function comments, etc.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -76,6 +77,7 @@
 #include "utils/palloc.h"
 #include "utils/rel.h"
 #include "utils/relcache.h"
+#include "utils/snapmgr.h"
```
metdos: Do we need to include snapmgr.h?

src/backend/distributed/transaction/worker_transaction.c
```diff
@@ -69,6 +69,58 @@ SendCommandToWorkers(TargetWorkerSet targetWorkerSet, char *command)
...
+ */
+void
+SendBareCommandListToWorkers(TargetWorkerSet targetWorkerSet, List *commandList)
+{
+	List *workerNodeList = WorkerNodeList();
+	ListCell *workerNodeCell = NULL;
+	char *nodeUser = CitusExtensionOwnerName();
+	ListCell *commandCell = NULL;
+
+	if (XactModificationLevel > XACT_MODIFICATION_NONE)
```
metdos: Do we need this? PostgreSQL does not allow index concurrently in transactions anyway. Also, this error message is not visible because of PG_TRY()/PG_CATCH().

If we ditch this, we can also run `ExecuteSequentialTasksWithoutResults` before metadata sync above and reduce the window for the mx sync problem.

metdos:
This locks COPY (expected given that we use transactions), but INSERT/UPDATE/DELETE are fine 👍

src/backend/distributed/transaction/worker_transaction.c
```diff
@@ -69,6 +69,58 @@ SendCommandToWorkers(TargetWorkerSet targetWorkerSet, char *command)
...
 
 /*
+ * SendBareCommandListToWorkers sends a list of commands to a set of target
+ * workers in serial. Commands are committed immediately: new connections are
+ * always used and no transaction block is used (hence "bare"). The connections
+ * are made as the extension owner to ensure write access to the Citus metadata
+ * tables. Primarly useful for INDEX commands using CONCURRENTLY.
+ */
+void
+SendBareCommandListToWorkers(TargetWorkerSet targetWorkerSet, List *commandList)
```
jasonmp85: Elsewhere in the code we already use `COMMIT_PROTOCOL_BARE` to describe commands sent without any sort of transaction wrapping them, so it's out of analogy to that. I might leave this the way it is.

src/backend/distributed/transaction/worker_transaction.c
jasonmp85: That's a reasonable point. I'll remove it. I'm not sure about calling this before `ExecuteSequentialTasks`, but I'll consider it.

src/include/distributed/multi_utility.h
```diff
@@ -23,6 +23,7 @@ extern bool EnableDDLPropagation;
 typedef struct DDLJob
 {
 	Oid targetRelationId;      /* oid of the target distributed relation */
+	bool concurrent;           /* related to a CONCURRENTLY index command? */
```
jasonmp85: Hm. I'll change to `concurrentIndexCmd`, then.

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -870,6 +870,29 @@ ExecuteModifyTasksWithoutResults(List *taskList)
 
 
 /*
+ * ExecuteSequentialTasksWithoutResults basically calls ExecuteModifyTasks in a
+ * loop in order to simulate sequential execution of a list of tasks. Useful in
+ * cases where issuing commands in parallel before waiting for results could
+ * result in deadlocks (such as CREATE INDEX CONCURRENTLY).
+ */
+int64
```
jasonmp85: That would make sense, yes. I copy-pasted from `ExecuteModifyTasksWithoutResults` and didn't realize I'd done that.

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -870,6 +870,29 @@ ExecuteModifyTasksWithoutResults(List *taskList)
...
+ExecuteSequentialTasksWithoutResults(List *taskList)
```
jasonmp85: I'll change it to the latter.

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -870,6 +870,29 @@ ExecuteModifyTasksWithoutResults(List *taskList)
...
+{
+	ListCell *taskCell = NULL;
+
+	foreach(taskCell, taskList)
+	{
+		Task *task = (Task *) lfirst(taskCell);
+		List *singleTask = list_make1(task);
+
+		ExecuteModifyTasks(singleTask, false, NULL, NULL);
```
jasonmp85: Sure.

src/backend/distributed/executor/multi_utility.c
jasonmp85: I previously had push/pop snapshot calls in here. Removed, so I'll remove this `#include`.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -1987,15 +1984,49 @@ ExecuteDistributedDDLJob(DDLJob *ddlJob)
...
+	else
+	{
+		/* save old commit protocol to restore at xact end */
+		Assert(SavedMultiShardCommitProtocol == COMMIT_PROTOCOL_BARE);
+		SavedMultiShardCommitProtocol = MultiShardCommitProtocol;
+		MultiShardCommitProtocol = COMMIT_PROTOCOL_BARE;
+
+		PG_TRY();
+		{
+			if (shouldSyncMetadata)
```
jasonmp85: Sure, I'll swap them. Ultimately, the coordinator node will be marked `INVALID`, which close to as good as we can do, without a bunch of custom hackery.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -1987,15 +1984,49 @@ ExecuteDistributedDDLJob(DDLJob *ddlJob)
...
+
+			ExecuteSequentialTasksWithoutResults(ddlJob->taskList);
+		}
+		PG_CATCH();
+		{
+			ereport(ERROR,
+					(errmsg("CONCURRENTLY-enabled index command failed"),
+					 errdetail("CONCURRENTLY-enabled index commands can fail partially, "
+							   "leaving behind an INVALID index."),
+					 errhint("Use DROP INDEX IF EXISTS to remove the invalid index, then "
```
jasonmp85: Sure. Does dropping an index ever take that long? I'll suggest `CONCURRENTLY`.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -134,12 +136,15 @@ static bool IsAlterTableRenameStmt(RenameStmt *renameStatement);
 static void ExecuteDistributedDDLJob(DDLJob *ddlJob);
 static void ShowNoticeIfNotUsing2PC(void);
 static List * DDLTaskList(Oid relationId, const char *commandString);
+static List * IndexTaskList(Oid relationId, IndexStmt *indexStmt);
```
jasonmp85: Sure.

jasonmp85:
Is this a :shipit: after these (minor?) things are fixed?

src/backend/distributed/executor/multi_utility.c
metdos: Aha, I think it does take that long :)

src/backend/distributed/transaction/worker_transaction.c
metdos: Just to be clear, I was referring to this comment for the order: https://github.com/citusdata/citus/pull/1287#discussion_r108928440

src/backend/distributed/transaction/worker_transaction.c
metdos: Okay.

src/include/distributed/multi_utility.h
metdos: Sounds good.

metdos:
Hey @jasonmp85, I did a final pass, it looks good. After making changes you noted above, you can 
🚢  and 🚀  it!

Issue #756, opened 2016-08-29
Bogus var on explain when ORDER BY contains aggregate


metdos:
The query below works without a problem if you run it without explain. If you run it with explain we get this error:

```
ERROR:  bogus varattno for OUTER_VAR var: 3
```

Steps to replicate the problem:

On all nodes:

``` sql
CREATE TYPE user_composite_type AS
(
    tenant_id BIGINT,
    user_id BIGINT
);
```

On the master node:

``` sql
CREATE TABLE events (
    composite_id user_composite_type,
    event_id bigint,
    event_type character varying(255),
    event_time bigint
);
SELECT master_create_distributed_table('events', 'composite_id', 'range');

SELECT master_create_empty_shard('events') AS new_shard_id
\gset
UPDATE pg_dist_shard SET shardminvalue = '(1,1)', shardmaxvalue = '(1,2000000000)'
WHERE shardid = :new_shard_id;

SELECT master_create_empty_shard('events') AS new_shard_id
\gset
UPDATE pg_dist_shard SET shardminvalue = '(1,2000000001)', shardmaxvalue = '(1,4300000000)'
WHERE shardid = :new_shard_id;

SELECT master_create_empty_shard('events') AS new_shard_id
\gset
UPDATE pg_dist_shard SET shardminvalue = '(2,1)', shardmaxvalue = '(2,2000000000)'
WHERE shardid = :new_shard_id;

SELECT master_create_empty_shard('events') AS new_shard_id
\gset
UPDATE pg_dist_shard SET shardminvalue = '(2,2000000001)', shardmaxvalue = '(2,4300000000)'
WHERE shardid = :new_shard_id;


CREATE TABLE users (
    composite_id user_composite_type,
    lastseen bigint
);
SELECT master_create_distributed_table('users', 'composite_id', 'range');

SELECT master_create_empty_shard('users') AS new_shard_id
\gset
UPDATE pg_dist_shard SET shardminvalue = '(1,1)', shardmaxvalue = '(1,2000000000)'
WHERE shardid = :new_shard_id;

SELECT master_create_empty_shard('users') AS new_shard_id
\gset
UPDATE pg_dist_shard SET shardminvalue = '(1,2000000001)', shardmaxvalue = '(1,4300000000)'
WHERE shardid = :new_shard_id;

SELECT master_create_empty_shard('users') AS new_shard_id
\gset
[...]

PR #1289, opened 2017-03-22
Subquery pushdown - Unions Phase - 1 

onderkalaci:
(This PR allows `UNION` and `UNION ALL` that are not in the top level subquery. We'll open a separate PR for that.)

In general, the PR implements the following logic: Once a `UNION` or `UNION ALL` is found, each target list element at the same target list position  is added  for each subquery that is in the `UNION`/`UNION ALL` s.

For instance, for the following query, we add `events_subquery_1.user_id`, `events_subquery_2.user_id`, `events_subquery_3.user_id` and `events_subquery_4.user_id` to the equivelence class. 

```SQL
SELECT ("final_query"."event_types") as types, count(*) AS sumOfEventType
FROM
  ( SELECT *, random()
   FROM
     ( SELECT "t"."user_id",
              "t"."time",
              unnest("t"."collected_events") AS "event_types"
      FROM
        ( SELECT "t1"."user_id",
                 min("t1"."time") AS "time",
                 array_agg(("t1"."event")
                           ORDER BY TIME ASC, event DESC) AS collected_events
         FROM (
                 (SELECT *
                  FROM
                   (SELECT "events"."user_id",
                          "events"."time",
                         0 AS event
                 FROM events_table as  "events"
                WHERE event_type IN (10, 11, 12, 13, 14, 15) ) events_subquery_1) 
                UNION 
                 (SELECT *
                  FROM
                    (SELECT "events"."user_id", "events"."time", 1 AS event
                     FROM events_table as "events"
                     WHERE event_type IN (15, 16, 17, 18, 19) ) events_subquery_2)
               UNION 
                 (SELECT *
                  FROM
                    (SELECT "events"."user_id", "events"."time", 2 AS event
                     FROM events_table as  "events"
                      WHERE event_type IN (20, 21, 22, 23, 24, 25) ) events_subquery_3)
               UNION 
                 (SELECT *
                  FROM
                    (SELECT "events"."user_id", "events"."time", 3 AS event
                     FROM events_table as "events"
                     WHERE event_type IN (26, 27, 28, 29, 30, 13)) events_subquery_4)) t1
         GROUP BY "t1"."user_id") AS t) "q" 
INNER JOIN
     (SELECT "users"."user_id"
      FROM users_table as "users"
      WHERE value_1 > 50 and value_1 < 70) AS t ON (t.user_id = q.user_id)) as final_query
GROUP BY types
ORDER BY types;
```

### DIFF
--- a/src/backend/distributed/planner/multi_logical_optimizer.c
+++ b/src/backend/distributed/planner/multi_logical_optimizer.c
@@ -42,6 +42,7 @@
 #include "parser/parse_agg.h"
 #include "parser/parse_coerce.h"
 #include "parser/parse_oper.h"
+#include "parser/parsetree.h"
 #include "utils/builtins.h"
 #include "utils/fmgroids.h"
 #include "utils/lsyscache.h"
@@ -151,8 +152,9 @@ static void ErrorIfContainsUnsupportedSubquery(MultiNode *logicalPlanNode,
 											   PlannerRestrictionContext *
 											   plannerRestrictionContext);
 static void ErrorIfCannotPushdownSubquery(Query *subqueryTree, bool outerQueryHasLimit);
+static void ErrorIfUnsupportedSetOperation(Query *subqueryTree, bool outerQueryHasLimit);
+static bool ExtractSetOperationStatmentWalker(Node *node, List **setOperationList);
 static void ErrorIfUnsupportedTableCombination(Query *queryTree);
-static void ErrorIfUnsupportedUnionQuery(Query *unionQuery);
 static bool TargetListOnPartitionColumn(Query *query, List *targetEntryList);
 static FieldSelect * CompositeFieldRecursive(Expr *expression, Query *query);
 static bool FullCompositeFieldList(List *compositeFieldList);
@@ -2945,18 +2947,7 @@ ErrorIfCannotPushdownSubquery(Query *subqueryTree, bool outerQueryHasLimit)
 
 	if (subqueryTree->setOperations)
 	{
-		SetOperationStmt *setOperationStatement =
-			(SetOperationStmt *) subqueryTree->setOperations;
-
-		if (setOperationStatement->op == SETOP_UNION)
-		{
-			ErrorIfUnsupportedUnionQuery(subqueryTree);
-		}
-		else
-		{
-			preconditionsSatisfied = false;
-			errorDetail = "Intersect and Except are currently unsupported";
-		}
+		ErrorIfUnsupportedSetOperation(subqueryTree, outerQueryHasLimit);
 	}
 
 	if (subqueryTree->hasRecursive)
@@ -3048,6 +3039,77 @@ ErrorIfCannotPushdownSubquery(Query *subqueryTree, bool outerQueryHasLimit)
 
 
 /*
+ * ErrorIfUnsupportedSetOperation is a helper function for ErrorIfCannotPushdownSubquery().
+ * It basically iterates over the subqueries that reside under the given set operations.
+ *
+ * The function also errors out for set operations INTERSECT and EXCEPT.
+ */
+static void
+ErrorIfUnsupportedSetOperation(Query *subqueryTree, bool outerQueryHasLimit)
+{
+	List *rangeTableList = subqueryTree->rtable;
+	List *rangeTableIndexList = NIL;
+	ListCell *rangeTableIndexCell = NULL;
+	List *setOperationStatementList = NIL;
+	ListCell *setOperationStatmentCell = NULL;
+
+	ExtractSetOperationStatmentWalker((Node *) subqueryTree->setOperations,
+									  &setOperationStatementList);
+	foreach(setOperationStatmentCell, setOperationStatementList)
+	{
+		SetOperationStmt *setOperation =
+			(SetOperationStmt *) lfirst(setOperationStatmentCell);
+
+		if (setOperation->op != SETOP_UNION)
+		{
+			ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+							errmsg("cannot push down this subquery"),
+							errdetail("Intersect and Except are currently unsupported")));
+		}
+	}
+
+	ExtractRangeTableIndexWalker((Node *) subqueryTree->setOperations,
+								 &rangeTableIndexList);
+	foreach(rangeTableIndexCell, rangeTableIndexList)
+	{
+		int rangeTableIndex = lfirst_int(rangeTableIndexCell);
+		RangeTblEntry *rangeTableEntry = rt_fetch(rangeTableIndex, rangeTableList);
+
+		Assert(rangeTableEntry->rtekind == RTE_SUBQUERY);
+
+		ErrorIfCannotPushdownSubquery(rangeTableEntry->subquery, outerQueryHasLimit);
+	}
+}
+
+
+/*
+ * ExtractSetOperationStatementWalker walks over a set operations statment,
+ * and finds all set operations in the tree.
+ */
+static bool
+ExtractSetOperationStatmentWalker(Node *node, List **setOperationList)
+{
+	bool walkerResult = false;
+	if (node == NULL)
+	{
+		return false;
+	}
+
+	if (IsA(node, SetOperationStmt))
+	{
+		SetOperationStmt *setOperation = (SetOperationStmt *) node;
+
+		(*setOperationList) = lappend(*setOperationList, setOperation);
+	}
+
+	walkerResult = expression_tree_walker(node, ExtractSetOperationStatmentWalker,
+										  setOperationList);
+
+	return walkerResult;
+}
+
+
+/*
  * ErrorIfUnsupportedTableCombination checks if the given query tree contains any
  * unsupported range table combinations. For this, the function walks over all
  * range tables in the join tree, and checks if they correspond to simple relations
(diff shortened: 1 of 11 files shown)


src/test/regress/expected/multi_subquery_complex_queries.out
```diff
@@ -10,165 +10,986 @@ ALTER SEQUENCE pg_catalog.pg_dist_jobid_seq RESTART 1400000;
...
+         GROUP BY "t1"."user_id") AS t) "q" 
+INNER JOIN
+     (SELECT "users"."user_id"
+      FROM users_table as "users"
+      WHERE value_1 > 50 and value_1 < 70) AS t ON (t.user_id = q.user_id)) as final_query
+GROUP BY types
+ORDER BY types;
+ERROR:  cannot pushdown the subquery since all relations are not joined using distribution keys
+DETAIL:  Each relation should be joined with at least one another relation using distribution keys and equality operator.
+-- not supported since events_subquery_2 doesn't have partition key on the target list
```
onderkalaci: Comment says not supported, but, query returns results. How is so?

src/test/regress/expected/multi_subquery_complex_queries.out
```diff
@@ -10,165 +10,986 @@ ALTER SEQUENCE pg_catalog.pg_dist_jobid_seq RESTART 1400000;
...
+                     FROM events_table as "events"
+                     WHERE event_type IN (26, 27, 28, 29, 30, 13)) events_subquery_4)) t1
+         GROUP BY "t1"."user_id") AS t) "q" 
+INNER JOIN
+     (SELECT "users"."user_id"
+      FROM users_table as "users"
+      WHERE value_1 > 50 and value_1 < 70) AS t ON (t.user_id = q.user_id)) as final_query
+GROUP BY types
+ORDER BY types;
+ERROR:  cannot pushdown the subquery since all relations are not joined using distribution keys
```
onderkalaci: I think this error message doesn't make much sense. We should error out properly saying that we don't support subqueries withing unions (or we should support it).

src/test/regress/expected/multi_subquery_complex_queries.out
```diff
@@ -866,5 +1687,115 @@ FROM
...
+                     FROM events_table as "events"
+                     WHERE event_type IN (26, 27, 28, 29, 30, 13) ) events_subquery_4) OFFSET 3 ) t1
+         GROUP BY "t1"."user_id") AS t) "q" 
+INNER JOIN
+     (SELECT "users"."user_id"
+      FROM users_table as "users"
+      WHERE value_1 > 50 and value_1 < 70) AS t ON (t.user_id = q.user_id)) as final_query
+GROUP BY types
+ORDER BY types;
+ERROR:  cannot push down this subquery
```
onderkalaci: Once `OFFSET` is moved to the inner query, we get different error messeage

src/test/regress/input/multi_subquery.source
```diff
@@ -173,12 +173,12 @@ FROM
...
-) b;
+-- TEMPORARLY DISABLE UNION ALL
+-- SELECT count(*) FROM
+-- (
+--   (SELECT l_orderkey FROM lineitem_subquery) UNION ALL
+--   (SELECT 1::bigint)
+--) b;
 
 ---
 -- TEMPORARLY DISABLE UNIONS WITHOUT JOINS
```
onderkalaci: Now, we should be able to run these queries.

src/backend/distributed/planner/multi_planner_utils.c
```diff
@@ -733,6 +782,144 @@ AddToAttributeEquivalenceClass(AttributeEquivalenceClass **attributeEquivalanceC
...
+												 SetOperationStmt *setOperation,
+												 Var *varToBeAdded,
+												 AttributeEquivalenceClass **
+												 attributeEquivalenceClass)
+{
+	Node *leftArgument = setOperation->larg;
+	Node *rightArgument = setOperation->rarg;
+
+	/* we're only interested in UNIONs */
+	if (setOperation->op != SETOP_UNION)
```
onderkalaci: This should be an Assert

src/test/regress/sql/multi_subquery_complex_queries.sql
```diff
@@ -12,168 +12,898 @@ SET citus.subquery_pushdown TO TRUE;
...
+     ( SELECT "t"."user_id",
+              "t"."time",
+              unnest("t"."collected_events") AS "event_types"
+      FROM
+        ( SELECT "t1"."user_id",
+                 min("t1"."time") AS "time",
+                 array_agg(("t1"."event")
+                           ORDER BY TIME ASC, event DESC) AS collected_events
+         FROM (
+                 (SELECT *
```
onderkalaci: Check this more @mtuncer

src/backend/distributed/planner/relation_restriction_equivalence.c
```diff
@@ -734,6 +783,144 @@ AddToAttributeEquivalenceClass(AttributeEquivalenceClass **attributeEquivalanceC
...
+	targetEntry = get_tle_by_resno(subquery->targetList, varToBeAdded->varattno);
+	if (!IsA(targetEntry->expr, Var))
+	{
+		return;
+	}
+
+	targetVar = (Var *) targetEntry->expr;
+	targetRangeTableEntry = rt_fetch(targetVar->varno, subquery->rtable);
+
+	/* we currently only support relations within UNIONs */
```
onderkalaci: I think this restriction doesn't make much sense

src/backend/distributed/planner/relation_restriction_equivalence.c
```diff
@@ -734,6 +783,144 @@ AddToAttributeEquivalenceClass(AttributeEquivalenceClass **attributeEquivalanceC
...
+	Node *leftArgument = setOperation->larg;
+	Node *rightArgument = setOperation->rarg;
+
+	/* we're only interested in UNIONs */
+	if (setOperation->op != SETOP_UNION)
+	{
+		return;
+	}
+
+	if (IsA(leftArgument, RangeTblRef))
```
mtuncer: We should use a walker to get the list of RangeTblRef and iterate on this list.

src/backend/distributed/planner/relation_restriction_equivalence.c
```diff
@@ -734,6 +783,144 @@ AddToAttributeEquivalenceClass(AttributeEquivalenceClass **attributeEquivalanceC
...
+												 SetOperationStmt *setOperation,
+												 Var *varToBeAdded,
+												 AttributeEquivalenceClass **
+												 attributeEquivalenceClass)
+{
+	Node *leftArgument = setOperation->larg;
+	Node *rightArgument = setOperation->rarg;
+
+	/* we're only interested in UNIONs */
+	if (setOperation->op != SETOP_UNION)
```
mtuncer: we already check this inside ```ErrorIfUnsupportedUnionQuery()```

src/backend/distributed/planner/relation_restriction_equivalence.c
```diff
@@ -649,33 +664,10 @@ AddToAttributeEquivalenceClass(AttributeEquivalenceClass **attributeEquivalanceC
...
-		attributeEqMember->relationId = rangeTableEntry->relid;
-
-		(*attributeEquivalanceClass)->equivalentAttributes =
-			lappend((*attributeEquivalanceClass)->equivalentAttributes,
-					attributeEqMember);
+		AddRteToAttributeEquivalenceClass(rangeTableEntry, attributeEquivalanceClass,
+										  varToBeAdded);
 	}
-	else if (rangeTableEntry->rtekind == RTE_SUBQUERY && !rangeTableEntry->inh)
+	else if (rangeTableEntry->rtekind == RTE_SUBQUERY)
```
mtuncer: Maybe move this inside to  ```AddSubqueryRteToAttributeEquivalenceClass``` or something similar as you did for ```AddRteToAttributeEquivalenceClass```

src/backend/distributed/planner/relation_restriction_equivalence.c
mtuncer: 👍  on that

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -3048,6 +3044,63 @@ ErrorIfCannotPushdownSubquery(Query *subqueryTree, bool outerQueryHasLimit)
...
+
+		Assert(rangeTableEntry->rtekind == RTE_SUBQUERY);
+
+		ErrorIfCannotPushdownSubquery(rangeTableEntry->subquery, outerQueryHasLimit);
+	}
+	else if (IsA(leftArgument, SetOperationStmt))
+	{
+		SetOperationStmt *leftUnion = (SetOperationStmt *) leftArgument;
+
+		ErrorIfUnsupportedUnionQuery(rangeTableList, leftUnion, outerQueryHasLimit);
```
onderkalaci: Why is so? I'd be in favor of processing in the format that postgres keeps these structs.

> We should also think about removing indirect recursion

Again, Postgres keeps exactly in that format. Why should we not follow it?

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -3048,6 +3044,63 @@ ErrorIfCannotPushdownSubquery(Query *subqueryTree, bool outerQueryHasLimit)
...
+						errdetail("Intersect and Except are currently unsupported")));
+	}
+
+	if (IsA(leftArgument, RangeTblRef))
+	{
+		RangeTblRef *rangeTableReference = (RangeTblRef *) leftArgument;
+		RangeTblEntry *rangeTableEntry = rt_fetch(rangeTableReference->rtindex,
+												  rangeTableList);
+
+		Assert(rangeTableEntry->rtekind == RTE_SUBQUERY);
```
onderkalaci: > Does it have to be subquery ? Can we have relation here ?

No, range table references are always to `RTE_SUBQUERY` for set operations.

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -2948,15 +2951,8 @@ ErrorIfCannotPushdownSubquery(Query *subqueryTree, bool outerQueryHasLimit)
...
-		if (setOperationStatement->op == SETOP_UNION)
-		{
-			ErrorIfUnsupportedUnionQuery(subqueryTree);
-		}
-		else
-		{
-			preconditionsSatisfied = false;
-			errorDetail = "Intersect and Except are currently unsupported";
-		}
+		ErrorIfUnsupportedUnionQuery(subqueryTree->rtable, setOperationStatement,
```
onderkalaci: Makes sense

src/backend/distributed/planner/relation_restriction_equivalence.c
onderkalaci: Makes sense, I'll do that

src/backend/distributed/planner/relation_restriction_equivalence.c
onderkalaci: > we already check this inside ErrorIfUnsupportedUnionQuery()

I realized that we call this function prior to `ErrorIfUnsupportedUnionQuery ()` to give more meaningful error message. So, we need to keep it as it is.

src/backend/distributed/planner/relation_restriction_equivalence.c
onderkalaci: Same here, I'd be in favor of following Postgres' structure as-it-is. Also, what would be our motivation of doing that?

src/backend/distributed/planner/relation_restriction_equivalence.c
onderkalaci: Should be fixed now

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -3049,6 +3045,63 @@ ErrorIfCannotPushdownSubquery(Query *subqueryTree, bool outerQueryHasLimit)
 
 
 /*
+ * ErrorIfUnsupportedUnionQuery is a helper function for ErrorIfCannotPushdownSubquery().
+ * It basically iterates over the subqueries that reside under the given set operations.
+ *
+ * The function also errors out for set operations INTERSECT and EXCEPT.
+ */
+static void
+ErrorIfUnsupportedUnionQuery(List *rangeTableList,
```
onderkalaci: > That works too. I think we should pass Query * as parameter here

That seems not feasible since it makes recursing into unions harder.

src/test/regress/input/multi_subquery.source
onderkalaci: > Now, we should be able to run these queries.

Sorry, those should be enabled on the second PR

src/backend/distributed/planner/relation_restriction_equivalence.c
mtuncer: We are working on a tree structure and adding qualified items to a list like structure. Separating search from actual processing may simplify the code. Right now we are evaluating 4 different options (left-rangetblref, left-setoperationstmt, right-rangetblref, right-setoperationstmt). What we actually need is to find out rangetblref's and call the processor function.

onderkalaci: Ok, that's fair enough. I updated the code and looks simpler.

src/test/regress/output/multi_subquery.source
```diff
@@ -1299,8 +1348,7 @@ GROUP BY
 	count_pay
 ORDER BY
 	count_pay;
-ERROR:  cannot pushdown the subquery since all relations are not joined using distribution keys
-DETAIL:  Each relation should be joined with at least one another relation using distribution keys and equality operator.
+ERROR:  bogus varattno for OUTER_VAR var: 3
```
mtuncer: huh ! Do we have this logged somewhere ?

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -3048,6 +3039,77 @@ ErrorIfCannotPushdownSubquery(Query *subqueryTree, bool outerQueryHasLimit)
...
+
+		Assert(rangeTableEntry->rtekind == RTE_SUBQUERY);
+
+		ErrorIfCannotPushdownSubquery(rangeTableEntry->subquery, outerQueryHasLimit);
+	}
+}
+
+
+/*
+ * ExtractSetOperationStatementWalker walks over a set operations statment,
```
mtuncer: it is more like set operation tree, note the typo ```statment``` --> ```statement```

src/test/regress/expected/multi_subquery_behavioral_analytics.out
```diff
@@ -77,8 +77,18 @@ FROM (
   GROUP BY  t1.user_id, hasdone_event
 ) t GROUP BY user_id, hasdone_event
 ORDER BY user_id;
-ERROR:  cannot pushdown the subquery since all relations are not joined using distribution keys
-DETAIL:  Each relation should be joined with at least one another relation using distribution keys and equality operator.
+ user_id | sum | length |   hasdone_event
```
onderkalaci: Oh nice catch

src/test/regress/expected/multi_subquery_behavioral_analytics.out
```diff
@@ -268,8 +281,18 @@ GROUP BY
   count_pay, user_id
 ORDER BY
   event_average DESC, count_pay DESC, user_id DESC;
-ERROR:  cannot pushdown the subquery since all relations are not joined using distribution keys
-DETAIL:  Each relation should be joined with at least one another relation using distribution keys and equality operator.
+ user_id |     event_average      | count_pay
```
onderkalaci: Fixed again

src/test/regress/expected/multi_subquery_complex_queries.out
onderkalaci: done

src/test/regress/expected/multi_subquery_complex_queries.out
onderkalaci: We don't get anymore. This was before we properly process the subqueries within the UNIONs

src/test/regress/output/multi_subquery.source
onderkalaci: Yes, sure we have the issue in #756 
This bug exists before this work as well.

src/backend/distributed/planner/multi_logical_optimizer.c
onderkalaci: fixed

PR #1292, opened 2017-03-23
Alter Table Add Constraint

### LINKED ISSUE #1239: Add support for Alter Table Add/Drop Constraint
We currently don't propagate Alter Table Add/Drop Constraint commands to related shards. This feature request came up in a customer conversation; and I'm opening this issue to track it.

<pre>
ALTER TABLE triggers DROP CONSTRAINT IF EXISTS triggers_pkey;
ERROR: alter table command is currently supported
DETAIL: Only ADD|DROP COLUMN, SET|DROP NOT NULL, SET|DROP DEFAULT and TYPE subcommands are supported.
</pre>

velioglu:
Fixes #1239.

Follow the same logic for create table and alter table to propagate constraint definitions to workers.


velioglu:
There is a problem on the implementation when the constraints are checked with ErrorIfNotSupportedConstraint from multi_ProcessUtility. Transaction errors out from the last check of ErrorIfNotSupportedForeignConstraint with the error message of :

ERROR:  cannot create foreign key constraint
DETAIL:  Citus Community Edition currently supports foreign key constraints only for "citus.shard_replication_factor = 1".
HINT:  Please change "citus.shard_replication_factor to 1". To learn more about using foreign keys with other replication factors, please contact us at https://citusdata.com/about/contact_us.

When I check the shard replication factor from the error it is equal to 2, yet when I control it using SHOW citus.shard_replication_factor via sql it is 1. Note that, this errors occurs with create_distributed_table also.

src/test/regress/expected/multi_alter_table_add_constraints.out
```diff
@@ -0,0 +1,131 @@
...
+-- Error out from the related shard
+INSERT INTO unique_test_table VALUES(1, 'Ahmet');
+INSERT INTO unique_test_table VALUES(1, 'Mehmet');
+ERROR:  duplicate key value violates unique constraint "unn_id_1400001"
+DETAIL:  Key (id)=(1) already exists.
+CONTEXT:  while executing command on localhost:57638
+ALTER TABLE unique_test_table DROP CONSTRAINT unn_id;
+-- Can create unique constraint over multiple columns including
+-- distribution column
+INSERT INTO unique_test_table VALUES(1, 'Mehmet');
```
byucesoy: Let's insert this just before the other insertion

src/test/regress/expected/multi_alter_table_add_constraints.out
```diff
@@ -0,0 +1,131 @@
+--
```
byucesoy: Let's also add test cases when we fail to create constraints because of;
* missing constraint name
* ADD CONSTRAINT is not only subcommand
* Constraint not on distribution column
* other...

Also let's add test for reference/append distributed tables too

src/test/regress/expected/multi_alter_table_add_constraints.out
```diff
@@ -0,0 +1,131 @@
...
+DETAIL:  Distributed relations cannot have UNIQUE, EXCLUDE, or PRIMARY KEY constraints that do not include the partition column (with an equality operator if EXCLUDE).
+ALTER TABLE products ADD CONSTRAINT p_key PRIMARY KEY(product_no);
+-- Error out from the related shard
+INSERT INTO products VALUES(1, 'product_1', 1);
+INSERT INTO products VALUES(1, 'product_1', 1);
+ERROR:  duplicate key value violates unique constraint "p_key_1400065"
+DETAIL:  Key (product_no)=(1) already exists.
+CONTEXT:  while executing command on localhost:57638
+drop table products;
+-- Check "EXCLUSION CONSTRAINT"
```
byucesoy: And also tests for NOT NULL

src/test/regress/expected/multi_alter_table_add_constraints.out
byucesoy: We should also add tests with transactions

src/backend/distributed/commands/multi_copy.c
```diff
@@ -810,10 +810,11 @@ OpenCopyConnections(CopyStmt *copyStatement, ShardConnections *shardConnections,
 
 /*
  * CanUseBinaryCopyFormat iterates over columns of the relation given in rowOutputState
- * and looks for a column whose type is array of user-defined type or composite type.
- * If it finds such column, that means we cannot use binary format for COPY, because
- * binary format sends Oid of the types, which are generally not same in master and
- * worker nodes for user-defined types.
+ * and looks for a column whose type is array of user-defined type. If it finds such
```
velioglu: oops 😃

src/test/regress/expected/multi_alter_table_add_constraints.out
```diff
@@ -0,0 +1,459 @@
...
+ALTER TABLE products ADD CONSTRAINT unn_pno UNIQUE(product_no);
+ALTER TABLE products ADD CONSTRAINT check_price CHECK(price > discounted_price);
+ALTER TABLE products ALTER COLUMN product_no SET NOT NULL;
+ALTER TABLE products ADD CONSTRAINT p_key_product PRIMARY KEY(product_no);
+-- Single shard DML command can't be located in the same transaction with above commands.
+INSERT INTO products VALUES(1,'product_1', 10, 8);
+ERROR:  single-shard DML commands must not appear in transaction blocks which contain multi-shard data modifications
+DROP TABLE products;
+ERROR:  current transaction is aborted, commands ignored until end of transaction block
+END;
```
byucesoy: Also let's have a test

```sql
BEGIN:
-- create a distributed table
-- alter table add constraint
ROLLBACK;
-- see no constraint is created in master
-- see no constraint is created in one of the worker
```

src/test/regress/sql/multi_alter_table_add_constraints.sql
```diff
@@ -0,0 +1,387 @@
...
+INSERT INTO unique_test_table_ref VALUES(1, 'Mehmet');
+
+DROP TABLE unique_test_table_ref;
+
+-- Check "UNIQUE CONSTRAINT" with append table
+CREATE TABLE unique_test_table_append(id int, name varchar(20));
+SELECT create_distributed_table('unique_test_table_append', 'id', 'append');
+
+-- We can add unique constraint on any column with reference tables
+ALTER TABLE unique_test_table_append ADD CONSTRAINT unn_name UNIQUE(name);
```
marcocitus: We should prevent it. Append-distributed tables may have overlapping shards and typically do.

src/test/regress/expected/multi_alter_table_add_constraints.out
```diff
@@ -0,0 +1,459 @@
...
+CREATE TABLE unique_test_table_append(id int, name varchar(20));
+SELECT create_distributed_table('unique_test_table_append', 'id', 'append');
+ create_distributed_table 
+--------------------------
+ 
+(1 row)
+
+-- We can add unique constraint on any column with reference tables
+ALTER TABLE unique_test_table_append ADD CONSTRAINT unn_name UNIQUE(name);
+WARNING:  table "unique_test_table_append" has a UNIQUE or EXCLUDE constraint
```
byucesoy: We discussed about this with @marcocitus. We want to prevent this behavior but also we do not want to break existing environments. If user has a append distributed table with unique constraint, we should not error out when creating new shard for this table (with COPY for example). 

It seems there are four case to consider. Correct behavior for each case should be;
- Table is not distributed yet, user creates an unique constraint, then distributes by append -> **error out during distributed table creation**
- Table is distributed(before this PR) by append and has NO unique constraint, user creates unique constraint with ALTER TABLE -> **error out during ALTER TABLE**
- Table is distributed(before this PR) by append and has unique constraint, user replicates a shard -> **propagate unique constraints to new shards**
- Table is distributed(before this PR) by append and has unique constraint, user creates new shard -> **propagate unique constraints to new shards**

@metdos we can implement this behavior with one line fix, do you have any objection adding the fix to this PR.

metdos:
> We discussed about this with @marcocitus. We want to prevent this behavior but also we do not want to break existing environments. If user has a append distributed table with unique constraint, we should not error out when creating new shard for this table (with COPY for example).
> 
> It seems there are four case to consider. Correct behavior for each case should be;
> 
> Table is not distributed yet, user creates an unique constraint, then distributes by append -> error out during distributed table creation
> Table is distributed(before this PR) by append and has NO unique constraint, user creates unique constraint with ALTER TABLE -> error out during ALTER TABLE
> Table is distributed(before this PR) by append and has unique constraint, user replicates a shard -> propagate unique constraints to new shards
> Table is distributed(before this PR) by append and has unique constraint, user creates new shard -> propagate unique constraints to new shards
> @metdos we can implement this behavior with one line fix, do you have any objection adding the fix to this PR.

We can add this as a separate commit in this PR and also add your explanation to commit description. Then, we should be okay!

src/test/regress/expected/multi_alter_table_add_constraints.out
```diff
@@ -0,0 +1,501 @@
...
+CREATE TABLE unique_test_table(id int, name varchar(20));
+SELECT create_distributed_table('unique_test_table', 'id');
+ create_distributed_table 
+--------------------------
+ 
+(1 row)
+
+-- Can only add unique constraint on distribution column (or group
+-- of columns including distribution column)
+ALTER TABLE unique_test_table ADD CONSTRAINT unn_name UNIQUE(name);
```
byucesoy: Specify that the command will error out.

src/test/regress/expected/multi_alter_table_add_constraints.out
```diff
@@ -0,0 +1,501 @@
...
+-- Error out from the related shard. Table can not have two rows with same id.
+INSERT INTO unique_test_table VALUES(1, 'Ahmet');
+INSERT INTO unique_test_table VALUES(1, 'Mehmet');
+ERROR:  duplicate key value violates unique constraint "unn_id_1400035"
+DETAIL:  Key (id)=(1) already exists.
+CONTEXT:  while executing command on localhost:57638
+ALTER TABLE unique_test_table DROP CONSTRAINT unn_id;
+-- Can create unique constraint over multiple columns which must include
+-- distribution column
+INSERT INTO unique_test_table VALUES(1, 'Mehmet');
```
byucesoy: According to comment, I expect that next example will be successfully adding a unique constraints over multiple columns, but we test failing case when there is already conflicting rows in the table. Let's keep both tests, but write separate comments for them.

src/test/regress/expected/multi_alter_table_add_constraints.out
```diff
@@ -0,0 +1,501 @@
...
+-- Command below should error out since 'name' is not a distribution column
+ALTER TABLE products_append ADD CONSTRAINT p_key PRIMARY KEY(name);
+ERROR:  cannot add UNIQUE, PRIMARY KEY or EXCLUDE constraint on "products_append" 
+DETAIL:  UNIQUE constraints, EXCLUDE constraints, and PRIMARY KEYs on append-partitioned tables cannot be enforced.
+HINT:  Consider using hash partitioning.
+ALTER TABLE products_append ADD CONSTRAINT p_key PRIMARY KEY(product_no);
+ERROR:  cannot add UNIQUE, PRIMARY KEY or EXCLUDE constraint on "products_append" 
+DETAIL:  UNIQUE constraints, EXCLUDE constraints, and PRIMARY KEYs on append-partitioned tables cannot be enforced.
+HINT:  Consider using hash partitioning.
+COPY products_append FROM STDIN DELIMITER AS ',';
```
byucesoy: No need to COPY data.

src/test/regress/expected/multi_alter_table_add_constraints.out
```diff
@@ -0,0 +1,501 @@
...
+    price numeric,
+    discounted_price numeric
+);
+SELECT create_distributed_table('products', 'product_no');
+ create_distributed_table 
+--------------------------
+ 
+(1 row)
+
+-- Can add column and table constraints
```
byucesoy: This comment does not make sense.

src/test/regress/expected/multi_alter_table_add_constraints.out
```diff
@@ -0,0 +1,501 @@
...
+ALTER TABLE products ADD CONSTRAINT p_check CHECK(price > 0);
+ALTER TABLE products ADD CONSTRAINT p_multi_check CHECK(price > discounted_price);
+-- Error out from the related shard. Table can not have row conflicting with p_multi_check
```
byucesoy: "Related shard" term does not feel right here. Lets make this (and others) something similar to;

"This query will error out because insterting row's price is less than 0.

src/test/regress/expected/multi_alter_table_add_constraints.out
```diff
@@ -0,0 +1,501 @@
...
+    price numeric,
+    discounted_price numeric
+);
+SELECT create_reference_table('products_ref');
+ create_reference_table 
+------------------------
+ 
+(1 row)
+
+-- Can add column and table constraints
```
byucesoy: Similar to above

src/test/regress/expected/multi_alter_table_add_constraints.out
```diff
@@ -0,0 +1,501 @@
...
+ 
+(1 row)
+
+-- Can only add exclusion constraint on distribution column (or group of columns
+-- including distribution column)
+-- Command below should error out since 'name' is not a distribution column
+ALTER TABLE products ADD CONSTRAINT exc_name EXCLUDE USING btree (name with =); 
+ERROR:  cannot create constraint on "products"
+DETAIL:  Distributed relations cannot have UNIQUE, EXCLUDE, or PRIMARY KEY constraints that do not include the partition column (with an equality operator if EXCLUDE).
+-- We can add composite exclusion because none of pair of rows are conflicting
```
byucesoy: We can add composite exclusion -> corrent
because none of pair of rows are conflicting -> did not understand that

src/test/regress/expected/multi_alter_table_add_constraints.out
```diff
@@ -0,0 +1,501 @@
...
+);
+SELECT create_distributed_table('products', 'product_no');
+ create_distributed_table 
+--------------------------
+ 
+(1 row)
+
+-- Can only add exclusion constraint on distribution column (or group of columns
+-- including distribution column)
+-- Command below should error out since 'name' is not a distribution column
```
byucesoy: Good example of comment :+1:

src/test/regress/expected/multi_alter_table_add_constraints.out
```diff
@@ -0,0 +1,501 @@
...
+ create_reference_table 
+------------------------
+ 
+(1 row)
+
+-- We can add exclusion constraint on any column
+ALTER TABLE products_ref ADD CONSTRAINT exc_name EXCLUDE USING btree (name with =); 
+-- We can add composite exclusion because none of pair of rows are conflicting
+ALTER TABLE products_ref ADD CONSTRAINT exc_pno_name EXCLUDE USING btree (product_no with =, name with =);
+-- Third insertion will error out, since it has the same name with second insertion
```
byucesoy: Good example of comment :+1:

src/test/regress/expected/multi_alter_table_add_constraints.out
```diff
@@ -0,0 +1,501 @@
...
+    price numeric
+);
+SELECT create_distributed_table('products_append', 'product_no','append');
+ create_distributed_table 
+--------------------------
+ 
+(1 row)
+
+-- Cannot add exclusion constraint on append table
+-- Command below should error out since 'name' is not a distribution column
```
byucesoy: Erroring out reason is not correct.

src/test/regress/expected/multi_alter_table_add_constraints.out
```diff
@@ -0,0 +1,501 @@
...
+ create_distributed_table 
+--------------------------
+ 
+(1 row)
+
+INSERT INTO products VALUES(1,'product_1', 5);
+-- Should error out since conflicts with the above single-shard data modification command.
+ALTER TABLE products ADD CONSTRAINT unn_pno UNIQUE(product_no);
+ERROR:  distributed DDL commands must not appear within transaction blocks containing single-shard data modifications
+COMMIT;
```
byucesoy: After error we ROLLBACK, not COMMIT.

src/test/regress/expected/multi_alter_table_add_constraints.out
```diff
@@ -0,0 +1,501 @@
...
+ 
+(1 row)
+
+BEGIN;
+-- Add constraints (which will be rollbacked)
+ALTER TABLE products ADD CONSTRAINT unn_pno UNIQUE(product_no);
+ALTER TABLE products ADD CONSTRAINT check_price CHECK(price > discounted_price);
+ALTER TABLE products ADD CONSTRAINT p_key_product PRIMARY KEY(product_no);
+ROLLBACK;
+-- All of them should be equal to zero
```
byucesoy: Instead of querying pg_constraint, you can just use

/d products and see there are no constraints.

src/test/regress/expected/multi_complex_expressions.out
```diff
@@ -445,15 +445,15 @@ DEBUG:  push down of limit count: 30
  l_partkey | o_custkey | l_quantity 
 -----------+-----------+------------
     154380 |       421 |      26.00
-    103981 |       325 |      26.00
+    138854 |        79 |      26.00
```
byucesoy: Why these numbers changed? I did not understand their relation with constraints on append table.

src/test/regress/expected/multi_create_table_constraints.out
```diff
@@ -6,13 +6,10 @@ ALTER SEQUENCE pg_catalog.pg_dist_jobid_seq RESTART 365000;
 --  test that Citus forbids unique and EXCLUDE constraints on append-partitioned tables.
 CREATE TABLE uniq_cns_append_tables
 (
-	partition_col integer UNIQUE,
+	partition_col integer,
```
byucesoy: According to comment above, this(and below) table is used to test that Citus forbids unique and EXCLUDE constraints on append-partitioned tables.

So I think it is better to keep UNIQUE keyword and see error message. It seems that this table is not used anywhere else.

src/test/regress/input/multi_copy.source
```diff
@@ -341,8 +341,7 @@ CREATE TABLE customer_worker_copy_append (
         c_phone char(15),
         c_acctbal decimal(15,2),
         c_mktsegment char(10),
-        c_comment varchar(117),
-                primary key (c_custkey));
+        c_comment varchar(117));
```
byucesoy: Fix the comment above.

src/test/regress/expected/multi_complex_expressions.out
velioglu: Since removing PRIMARY KEY constraints change the query plan, it effects the order of rows with the same l_quantity. I have changed the column used by ORDER BY to l_partkey, o_custkey from l_quantity to generate the same result regardless of the plan.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -133,6 +136,11 @@ static void ErrorIfDistributedAlterSeqOwnedBy(AlterSeqStmt *alterSeqStmt);
 static void ErrorIfUnsupportedTruncateStmt(TruncateStmt *truncateStatement);
 static bool OptionsSpecifyOwnedBy(List *optionList, Oid *ownedByTableId);
 static void ErrorIfDistributedRenameStmt(RenameStmt *renameStatement);
+static void ErrorIfUnsupportedAlterAddConstraintStmt(AlterTableStmt *alterTableStatement);
+static void ErrorIfNotSupportedForeignConstraint(Relation relation,
```
byucesoy: I know this is existing function, but I realized we generally used "ErrorIf**Un**supported..." instead of "ErrorIf**Not**Supporter...", could you rename this and ErrorIfNotSupportedConstraint functions.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -391,6 +399,27 @@ multi_ProcessUtility(Node *parsetree,
...
+		/*
+		 * Check the altered table's constraints using the same logic with create table.
+		 * Note that we already eliminated alter table command with more than one
+		 * sub-command with ErrorIfUnsupportedAlterTableStmt function. Therefore,
+		 * controlling the type of first(and single) sub-command of command list
+		 * is enough.
+		 */
+		if (IsA(parsetree, AlterTableStmt))
+		{
+			AlterTableStmt *alterTableStatement = (AlterTableStmt *) parsetree;
```
byucesoy: Let's add here an Assert to ensure commandList really has single sub-command, similar to [here](https://github.com/citusdata/citus/blob/master/src/backend/distributed/executor/multi_utility.c#L944)

src/test/regress/expected/multi_complex_expressions.out
```diff
@@ -439,21 +439,21 @@ FROM
 WHERE 
 	li.l_quantity > 25
 ORDER BY
-	li.l_quantity
+	li.l_partkey, o.o_custkey
```
metdos: You can do
```sql
ORDER BY
    li.l_quantity, li.l_partkey, o.o_custkey
```

src/test/regress/expected/multi_create_table_constraints.out
```diff
@@ -41,15 +32,15 @@ CREATE TABLE pk_on_non_part_col
 	other_col integer PRIMARY KEY
 );
 SELECT master_create_distributed_table('pk_on_non_part_col', 'partition_col', 'hash');
-ERROR:  cannot distribute relation: "pk_on_non_part_col"
+ERROR:  cannot create constraint on "pk_on_non_part_col"
```
metdos: Why did we change this error message? Th older one makes more sense.

src/test/regress/expected/multi_foreign_key.out
```diff
@@ -464,8 +462,7 @@ SELECT create_distributed_table('referencing_table', 'ref_id', 'hash');
 
 -- columns for the referenced table is empty
 ALTER TABLE referencing_table ADD CONSTRAINT test_constraint FOREIGN KEY(ref_id) REFERENCES referenced_table ON DELETE CASCADE;
-ERROR:  cannot create foreign key constraint because referenced column list is empty
-HINT:  Add column names to "REFERENCES" part of the statement.
+ERROR:  number of referencing and referenced columns for foreign key disagree
```
metdos: The error message seems more explanatory.

src/test/regress/expected/multi_foreign_key.out
```diff
@@ -882,8 +870,7 @@ SELECT create_reference_table('reference_table');
 (1 row)
 
 ALTER TABLE reference_table ADD CONSTRAINT fk FOREIGN KEY(referencing_column) REFERENCES referenced_local_table(id);
-ERROR:  cannot create foreign key constraint
-DETAIL:  Foreign key constraints are not allowed from or to reference tables.
+ERROR:  relation referenced_local_table is not distributed
```
metdos: This new error message also is not a bit confusing.

src/test/regress/expected/multi_create_table_constraints.out
velioglu: We check the constraints of CREATE TABLE and ALTER TABLE statements using the same part of the code. For the ALTER TABLE case, "cannot distribute relation" seems meaningless that the table has already distributed.

metdos:
@velioglu, I have a question. I see that we moved `ErrorIfNotSupportedConstraint` and `ErrorIfNotSupportedConstraint` to another file. Did we just move them or did we also make some changes on them?

src/test/regress/expected/multi_foreign_key.out
velioglu: Error message is generated by Postgres, i.e. user will be faced with the same error message running the same command without distributing tables.

`CREATE TABLE referenced_table(id int UNIQUE, test_column int, PRIMARY KEY(id, test_column));
CREATE TABLE referencing_table(id int, ref_id int);
ALTER TABLE referencing_table ADD CONSTRAINT test_constraint FOREIGN KEY(ref_id) REFERENCES referenced_table ON DELETE CASCADE;
ERROR:  number of referencing and referenced columns for foreign key disagree`

velioglu:
@metdos, we almost just moved and renamed them. Only changed some error messages to make them meaningful with both CREATE TABLE and ALTER TABLE statements and removed an unnecessary check for co-located tables in ErrorIfUnsupportedForeignConstraint.

metdos:
> @metdos, we almost just moved and renamed them. Only changed some error messages to make them meaningful with both CREATE TABLE and ALTER TABLE statements and removed an unnecessary check for co-located tables in ErrorIfUnsupportedForeignConstraint.

What about here?
https://github.com/citusdata/citus/pull/1292/files#diff-a8033caa333591e68f4fb7a599757296R2048

src/test/regress/expected/multi_foreign_key.out
velioglu: Since we check foreign table constraints after we create them, planning phase catch the not distributed table before constraint checking catch the reference table. I can get the change back by adding a check of reference tables before planning phase, yet it will break the logic of checking constraints of ALTER TABLE and CREATE TABLE statements using the same code part.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -1623,211 +1651,421 @@ ErrorIfUnsupportedAlterTableStmt(AlterTableStmt *alterTableStatement)
...
+		{
+			ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+							errmsg("cannot create foreign key constraint"),
+							errdetail("Partition column must exist both "
+									  "referencing and referenced side of the "
+									  "foreign constraint statement and it must "
+									  "be in the same ordinal in both sides.")));
+		}
+
+		multiShardedReferencingExist = IsDistributedTable(referencingTableId) &&
```
metdos: This part is confusing.  I assume `multiShardedReferencingExist` means `IsDistributedTable(referencingTableId)` and `multiShardedReferencingBeingCreated ` means `!IsDistributedTable(referencingTableId)`. `!SingleReplicatedTable(referencingTableId)` and `ShardReplicationFactor > 1` are only exist to use in if below.

What are you trying to do here? I presume you want to distinguish between `CREATE TABLE` and `ALTER TABLE` and check `shard_count > 1` according to that. If we rewrite this in this form, will it be correct?

```c
if (IsDistributedTable(referencingTableId))
  {
    \\ ALTER TABLE...
    if (!SingleReplicatedTable(referencingTableId))
    {
        singleReplicatedTable = false;
     }
}
else
{
  \\ CREATE TABLE
  if (ShardReplicationFactor > 1)
  {
       singleReplicatedTable = false;
  }
}

if(!singleReplicatedTable)
{
 \\ error out
}
```

You may also want to use `selfReferencingTable` in checks above. What is our policy for self-referencing tables? Do we support constraints on them all the time?

Does this make sense? What do you think?

velioglu: Yes, I was trying to distinguish whether the function is called for CREATE TABLE or ALTER TABLE by IsDistributedTable(), so if we rewrite it in the form you have written that will also be correct. Self-referencing tables also error out if the shard replication factor is > 1. 

What we check by multiShardedReferencingExist and multiShardedReferencingBeingCreated are not clear enough now, so I'm going to change the code to make it more readable.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -1623,211 +1649,432 @@ ErrorIfUnsupportedAlterTableStmt(AlterTableStmt *alterTableStatement)
...
+		if (IsDistributedTable(referencingTableId))
+		{
+			if (!SingleReplicatedTable(referencingTableId))
+			{
+				singleReplicatedTable = false;
+			}
+		}
+
+		/* check whether creating single replicated table with foreign constraint */
+		if (!IsDistributedTable(referencingTableId))
```
metdos: We can use `else` instead of this `if`. You can get comments inside if blocks to have symmetry.

src/test/regress/expected/multi_alter_table_add_constraints.out
```diff
@@ -0,0 +1,477 @@
...
+INSERT INTO unique_test_table VALUES(1, 'Mehmet');
+-- Can not create constraint since it conflicts with the existing data
+ALTER TABLE unique_test_table ADD CONSTRAINT unn_id UNIQUE(id);
+ERROR:  could not create unique index "unn_id_1450034"
+DETAIL:  Key (id)=(1) is duplicated.
+CONTEXT:  while executing command on localhost:57637
+-- Can create unique constraint over multiple columns which must include
+-- distribution column
+ALTER TABLE unique_test_table ADD CONSTRAINT unn_id_name UNIQUE(id, name);
+-- Error out, since tables can not have two rows woth same id and name.
```
metdos: typo: woth

src/test/regress/expected/multi_alter_table_add_constraints.out
```diff
@@ -0,0 +1,477 @@
...
+    price numeric
+);
+SELECT create_distributed_table('products', 'product_no');
+ create_distributed_table 
+--------------------------
+ 
+(1 row)
+
+ALTER TABLE products ALTER COLUMN name SET NOT NULL;
+-- Insertions will error our since both product_no and name can not have NULL value
```
metdos: typo: our

src/backend/distributed/executor/multi_utility.c
```diff
@@ -1583,7 +1610,6 @@ ErrorIfUnsupportedAlterTableStmt(AlterTableStmt *alterTableStatement)
 			case AT_DropColumn:
 			case AT_ColumnDefault:
 			case AT_AlterColumnType:
-			case AT_SetNotNull:
```
metdos: Why did move `AT_SetNotNull` to below?

src/backend/distributed/executor/multi_utility.c
```diff
@@ -391,6 +399,25 @@ multi_ProcessUtility(Node *parsetree,
 	{
 		ListCell *ddlJobCell = NULL;
 
+		if (IsA(parsetree, AlterTableStmt))
```
metdos: Should not we have this check in `ErrorIfUnsupportedAlterTableStmt`?

Looking this more, it looks like you do checks here after running master query so it becomes easier to use create table checks? Am I right?

velioglu: Yes, you are right.

metdos: Can we add a comment explaining why we have these checks here instead of earlier with other checks?

src/backend/distributed/executor/multi_utility.c
```diff
@@ -391,6 +399,25 @@ multi_ProcessUtility(Node *parsetree,
...
+			ListCell *commandCell = NULL;
+
+			foreach(commandCell, commandList)
+			{
+				AlterTableCmd *command = (AlterTableCmd *) lfirst(commandCell);
+				AlterTableType alterTableType = command->subtype;
+
+				if (alterTableType == AT_AddConstraint)
+				{
+					Assert(commandList->length <= 1);
```
velioglu: Since we checked that length of commandList is smaller than 1 on ErrorIfUnsupportedAlterTableStmt, we are sure list_length(commandList) = 1. Both ErrorIfUnsupportedAlterTableStmt and PlanAlterTableStmt check the length of command list in that way, should I fix them too ?

src/backend/distributed/executor/multi_utility.c
```diff
@@ -1532,7 +1559,7 @@ ErrorIfUnsupportedDropIndexStmt(DropStmt *dropIndexStatement)
  * ALTER TABLE ALTER COLUMN SET DATA TYPE
  * ALTER TABLE SET|DROP NOT NULL
  * ALTER TABLE SET|DROP DEFAULT
- * ALTER TABLE ADD|DROP CONSTRAINT FOREIGN
+ * ALTER TABLE ADD|DROP CONSTRAINT
```
metdos: This my mistake.  CONSTRAINT >  CONSTRAINT FOREIGN

src/backend/distributed/executor/multi_utility.c
metdos: > Since we checked that length of commandList is smaller than 1 on ErrorIfUnsupportedAlterTableStmt, we are sure list_length(commandList) = 1.

Nice!

> Both ErrorIfUnsupportedAlterTableStmt and PlanAlterTableStmt check the length of command list in that way, should I fix them too ?

I think we can change `->length` to `list_length` because `list_length` makes null check. Even though it is not required here, it's a better practice.

src/backend/distributed/executor/multi_utility.c
velioglu: I think other 4 constraints should be prevented to apply them over the partition column, but adding NOT NULL constraint over the partition column won't cause a problem.

metdos:
Looks good 🚀

PR #1294, opened 2017-03-23
Prepare for valgrind automation

### LINKED ISSUE #1039: Automated valgrind test runs
Besides the work done in #835 we also need to make the run free from spurious failures:
- due to timeouts some tests still fail, we need to make
  REMOTE_NODE_CONNECT_TIMEOUT configurable ( #912)
- due to slow queries, autovacuum analyzes some tables, causing
  different plans. This probably should be fixed by adding explicit
  analyze statements to those
- should we verify that postgres was compiled with USE_VALGRIND
  specified? Probably not worth the effort.

byucesoy:
Fixes #1039 

With this PR we made some changes to automate valgrind tests;
* We added configurable pg_ctl-timeout and connection-timeout parameters to test script.
* Since valgrind tests takes long time, it affects output of some tests. We try to stabilize the tests by adding AUTOVACUUM commands.
* We also modified parameters passed to valgrind to generate clean log file without unnecessary warnings.

### DIFF
--- a/src/test/regress/expected/multi_explain.out
+++ b/src/test/regress/expected/multi_explain.out
@@ -34,6 +34,9 @@ BEGIN
   RETURN result;
 END;
 $BODY$ LANGUAGE plpgsql;
+-- VACUMM related tables to ensure test outputs are stable
+VACUUM ANALYZE lineitem;
+VACUUM ANALYZE orders;
 -- Test Text format
 EXPLAIN (COSTS FALSE, FORMAT TEXT)
 	SELECT l_quantity, count(*) count_quantity FROM lineitem
@@ -287,12 +290,13 @@ Limit
                     ->  Limit
                           ->  Sort
                                 Sort Key: lineitem.l_quantity
-                                ->  Hash Join
-                                      Hash Cond: (lineitem.l_orderkey = orders.o_orderkey)
-                                      ->  Seq Scan on lineitem_290001 lineitem
-                                            Filter: (l_quantity < 5.0)
-                                      ->  Hash
-                                            ->  Seq Scan on orders_290008 orders
+                                ->  Merge Join
+                                      Merge Cond: (orders.o_orderkey = lineitem.l_orderkey)
+                                      ->  Index Scan using orders_pkey_290008 on orders_290008 orders
+                                      ->  Sort
+                                            Sort Key: lineitem.l_orderkey
+                                            ->  Seq Scan on lineitem_290001 lineitem
+                                                  Filter: (l_quantity < 5.0)
 -- Test insert
 EXPLAIN (COSTS FALSE)
 	INSERT INTO lineitem VALUES(1,0);
@@ -314,11 +318,9 @@ Custom Scan (Citus Router)
   ->  Task
         Node: host=localhost port=57638 dbname=regression
         ->  Update on lineitem_290000
-              ->  Bitmap Heap Scan on lineitem_290000
-                    Recheck Cond: (l_orderkey = 1)
+              ->  Index Scan using lineitem_pkey_290000 on lineitem_290000
+                    Index Cond: (l_orderkey = 1)
                     Filter: (l_partkey = 0)
-                    ->  Bitmap Index Scan on lineitem_pkey_290000
-                          Index Cond: (l_orderkey = 1)
 -- Test delete
 EXPLAIN (COSTS FALSE)
 	DELETE FROM lineitem
@@ -329,11 +331,9 @@ Custom Scan (Citus Router)
   ->  Task
         Node: host=localhost port=57638 dbname=regression
         ->  Delete on lineitem_290000
-              ->  Bitmap Heap Scan on lineitem_290000
-                    Recheck Cond: (l_orderkey = 1)
+              ->  Index Scan using lineitem_pkey_290000 on lineitem_290000
+                    Index Cond: (l_orderkey = 1)
                     Filter: (l_partkey = 0)
-                    ->  Bitmap Index Scan on lineitem_pkey_290000
-                          Index Cond: (l_orderkey = 1)
 -- Test single-shard SELECT
 EXPLAIN (COSTS FALSE)
 	SELECT l_quantity FROM lineitem WHERE l_orderkey = 5;
@@ -342,10 +342,8 @@ Custom Scan (Citus Router)
   Tasks Shown: All
   ->  Task
         Node: host=localhost port=57637 dbname=regression
-        ->  Bitmap Heap Scan on lineitem_290000 lineitem
-              Recheck Cond: (l_orderkey = 5)
-              ->  Bitmap Index Scan on lineitem_pkey_290000
-                    Index Cond: (l_orderkey = 5)
+        ->  Index Scan using lineitem_pkey_290000 on lineitem_290000 lineitem
+              Index Cond: (l_orderkey = 5)
 SELECT true AS valid FROM explain_xml($$
 	SELECT l_quantity FROM lineitem WHERE l_orderkey = 5$$);
 t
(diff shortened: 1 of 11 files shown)


byucesoy:
At the moment, I am executing this script to run valgrind test on fresh Ubuntu instace. It needs some tweaking before going into automation(like dynamically using most recent postgresql version or setting file permissions to something less permissive), but for manual testing, it is enough.

#Download and install required packages
sudo apt-get update
sudo apt-get install build-essential libreadline6 libreadline6-dev zlib1g-dev flex git libssl-dev valgrind

```sh
#Set environment variables
export LC_ALL=en_US.UTF-8
export LANG=en_US.UTF-8
export LANGUAGE=en_US.UTF-8
export PG_CONFIG=/usr/local/pgsql/bin/pg_config

#Download and install PostgreSQL
wget https://ftp.postgresql.org/pub/source/v9.6.2/postgresql-9.6.2.tar.gz
tar -xzvf postgresql-9.6.2.tar.gz
rm postgresql-9.6.2.tar.gz
cd postgresql-9.6.2/
./configure --enable-cassert --enable-debug CFLAGS="-ggdb -Og -fno-omit-frame-pointer -DUSE_VALGRIND"
make -j8 -s
sudo make install
export PATH=/usr/local/pgsql/bin:$PATH

#Download and install Citus
cd ..
git clone https://github.com/citusdata/citus.git
git checkout fix_test_outputs_for_valgrind
cd citus/
./configure
make clean
make -j8 -s
sudo make install

#Change permissions and run valgrind tests
sudo chmod 777 /usr/local/pgsql/bin/ -R
sudo chown ubuntu /usr/local/pgsql/bin/ -R
cd src/test/regress
make check-multi-vg
```

metdos:
@byucesoy, I have an uninformed question. Running "make check-multi-vg" runs regression tests and compares query results. However, how will we know if there is a memory leak?

byucesoy:
@metdos if valgrind detects a memory related problem, it will exit with exit code 128. We can detect it by looking return code of the test process.

However, there are couple of things we need to decide on;
* Relying return code to detect memory problems, means if there is a memory problem we will exit immediately and will not execute remaining tests. I think this is not desirable because if there are more than one memory problems, we will only detect one of the problems at a time. (We can configure valgrind to not exit at problems)

* At the moment we pass --leak-check=no to the valgrind. That means we are not detecting memory leaks(we still detect problems such as branching on uninitialized variable or accessing freed memory block). Do we want to continue that behavior?

* Will we automate the valgrind test runs? If so did we decide on frequency? (I guess it will be nightly) I do not have exact time but I guess a test took around 10 hours to complete (which will increase with each new test). Do we plan to keep an instance open all time for valgrind tests?

@anarazel is there a specific reason for the `--leak-check=no` and `--error-exitcode=128` parameters? Should we keep them as they are?

byucesoy:
We discussed about this on mail. Here are the 

* --leak_check=no is necessary. Because;
> most leaks in postgres are prevented via the memory context mechanism, and there's a good number of "intentional" one-off leaks. They're also usually not that critical an issue in comparison to actually undefined behaviour. 

* For automation, it is not decided yet whether we keep an big open instance for valgrind tests or use existing smaller instance to open/close bigger instance.

Remaining decision is not related with this PR but the next PR we will open. This PR is ready for review @metdos

src/test/regress/expected/multi_explain.out
```diff
@@ -34,6 +34,9 @@ BEGIN
   RETURN result;
 END;
 $BODY$ LANGUAGE plpgsql;
+-- VACUMM relates tables to ensure test outputs are stable
```
metdos: relates -> related

src/test/regress/output/multi_load_data.source
```diff
@@ -19,3 +19,6 @@ ALTER SEQUENCE pg_catalog.pg_dist_shardid_seq RESTART 290000;
 \copy lineitem_hash_part FROM '@abs_srcdir@/data/lineitem.2.data' with delimiter '|'
 \copy orders_hash_part FROM '@abs_srcdir@/data/orders.1.data' with delimiter '|'
 \copy orders_hash_part FROM '@abs_srcdir@/data/orders.2.data' with delimiter '|'
+VACUUM ANALYZE;
```
metdos: Hmm, we don't propagate this to worker nodes. Does this effect the explain output?

src/test/regress/pg_regress_multi.pl
```diff
@@ -56,6 +60,10 @@ ()
 my %functions = ();
 my %operators = ();
 my $valgrind = 0;
+my $valgrind_path = "valgrind";
```
metdos: Should we use hyphen "-", or underscore "_" in variables? We should also update prints according to this above.

src/test/regress/pg_regress_multi.pl
```diff
@@ -56,6 +60,10 @@ ()
...
+my $valgrind_log_file = "valgrind_test.log";
+my $pg_ctl_timeout = undef;
```
metdos: Shall we set some meaningful value instead of `undef`?

src/test/regress/pg_regress_multi.pl
```diff
@@ -71,6 +79,10 @@ ()
     'load-extension=s' => \@extensions,
     'server-option=s' => \@userPgOptions,
     'valgrind' => \$valgrind,
+    'valgrind-path=s' => \$valgrind_path,
```
metdos: Maybe we can use camelCase for perl variables?

src/test/regress/pg_regress_multi.pl
byucesoy: I think it is not possible to have a variable name with hyphen "-" in perl. General convention for command line parameters is to separating words with hyphen. We just map hyphened variables to underscored variables below.

src/test/regress/pg_regress_multi.pl
byucesoy: [Here](https://github.com/citusdata/citus/pull/1294/files/97838113560ee66d8c0519d35e087d875b287e62#diff-a7ce4fb174e1acce25b5bcd3121576a9R153), we only set PGCTLTIMEOUT variable if pg_ctl_timeout is not defined. I think it is necessary to keep this as-is.

metdos: That's a good point. Why did we update that line? Isn't it better to check if `$valgrind` is defined?

byucesoy: `$valgrind` can be defined without defining `$pg_ctl_timeout`, or we may want to use `$pg_ctl_timeout` for some other purpose.

metdos: Then, shall we update the comment there?

metdos:
Looks nice 🚀

anarazel:
@byucesoy Why have you removed the --error-exitcode=128 parameter (thus making it harder to automate the tests), and changed --read-var-info to off (making it harder to interpret the failures)?

src/test/regress/sql/multi_explain.sql
```diff
@@ -36,6 +36,10 @@ BEGIN
 END;
 $BODY$ LANGUAGE plpgsql;
 
+-- VACUMM related tables to ensure test outputs are stable
```
anarazel: Typo.

byucesoy:
@anarazel 

--error-exitcode=128 causes exiting from tests after first failed test. Running valgrind tests takes time and if we want to keep that parameter, we can only fix errors one by one.

--read-var-info=yes prints lots of unnecessary warning messages . It prints warnings like this more than 50.000 times, which makes reading actual error messages more difficult.
```
--19307-- warning: evaluate_Dwarf3_Expr: unhandled DW_OP_ 0xf2
```
It seems this is a bug on [valgrind side](https://bugs.kde.org/show_bug.cgi?id=339744). As far as I understand, when it tries to collect more information and cannot find this information for some reason(because of the optimizations for example), it prints this error message. It seems optimizations is not only reason for this, because I encountered same warnings even when I compiled PostgreSQL and Citus without optimizations.

* I asked about it on valgrind IRC channel and confirmed it is not critical.
* In our previous test documents, we also used it as --read-var-info=no
* Instructions in [this page](https://wiki.postgresql.org/wiki/Valgrind) does not set --read-var-info to anything and its default value is no..
* Suppressions does not work for this kind of warnings.

Thus, I decided to use --read-var-info=no

anarazel:
> --error-exitcode=128 causes exiting from tests after first failed test. Running valgrind tests takes time and if we want to keep that parameter, we can only fix errors one by one.

Right. That's fine for doing it locally, but not fine for doing it in an automated manner.

>>  --read-var-info=yes prints lots of unnecessary warning messages . It prints warnings like this more than 50.000 times, which makes reading actual error messages more difficult.
> --19307-- warning: evaluate_Dwarf3_Expr: unhandled DW_OP_ 0xf2

That warning is harmless however.

It's fine to make such calls, but you should make them as a comment in this PR...

PR #1295, opened 2017-03-23
Update README.md

ozgune:
Updated to Readme.md's Local Cluster steps to include the right numbering.


begriffs:
Good catch.

Normally in the Github markdown format if you write

```markdown
1. first one
1. another
1. the last
```

it renders as "1, 2, 3" rather than "1, 1, 1." This makes it easier to add/remove/rearrange items while editing. But it looks like in our case the list items had too many things in between which reset the counter.

Anyhow I just previewed your branch and the readme looks good there.

PR #1296, opened 2017-03-27
Allow top level subquery joins (not ready for merge yet)

mtuncer:
Our existing subquery logic requires that the top level query can only select from a single subquery and can not have joins.

This work relaxes that requirement to allow subquery joins in top level query. It is done in a way to be compatible with our existing multi tree structure. Another query structure is created and placed under the top level query such that all joins, selections, and column projections are done inside the this query and the top level query only performs grouping, ordering etc.

The newly created query needed an alias, and column names for target list. Auxiliary names are created for that. Some better naming might be required here.


onderkalaci:
@mtuncer Do we still need this PR? Or, are we ok to close since we have #1298 ?

PR #1298, opened 2017-03-29
Enable top level subquery join queries

mtuncer:
This work enables
- Top level subquery joins
- Joins between subqueries and relations
- Joins involving more than 2 range table entries

A new regression test file is added to reflect enabled test cases


src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1993,62 +2038,107 @@ SubqueryPushdownMultiPlanTree(Query *queryTree)
...
+			columnNamesList = lappend(columnNamesList, makeString(
+										  columnNameString->data));
+		}
+
 		column->varno = 1;
+		column->varattno = targetNo;
 	}
 
+	/* new query only has target entries and join tree */
+	queryCopy->targetList = subqueryTargetEntryList;
```
onderkalaci: Why are we using a copy of the original query? 
It makes really hard to follow the code given that we're actually trying to replace the query with a projection query. So, I'd suggest we make this intention clear such as only creating a new query object and getting the required fields (there may be other fields we should fill):

```C
	projectionQuery = makeNode(Query);
	projectionQuery->rtable = copyObject(queryTree->rtable);
	projectionQuery->jointree = copyObject(queryTree->jointree);
	projectionQuery->commandType = CMD_SELECT;
	projectionQuery->targetList = subqueryTargetEntryList;

```

And, I think this logic (ie., including the above column preparation) should live in another function given that it is completely separate what we're doing here.

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1993,62 +2038,107 @@ SubqueryPushdownMultiPlanTree(Query *queryTree)
...
+	 * upper level query's target list, and update varno/varattno attributes
+	 * of them accordingly.
 	 */
-	Assert(list_length(subqueryEntryList) == 1);
-
-	qualifierColumnList = pull_var_clause_default((Node *) qualifierList);
-	targetListColumnList = pull_var_clause_default((Node *) targetEntryList);
-
-	columnList = list_concat(qualifierColumnList, targetListColumnList);
+	columnList = pull_var_clause_default((Node *) targetEntryList);
```
onderkalaci: We need to consider `HAVING` clause as well. It currently crashes.

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1993,62 +2038,107 @@ SubqueryPushdownMultiPlanTree(Query *queryTree)
...
+	queryCopy->hasAggs = false;
+	queryCopy->distinctClause = NIL;
+	queryCopy->hasDistinctOn = false;
+	queryCopy->limitCount = NULL;
+	queryCopy->limitOffset = NULL;
+	queryCopy->havingQual = NULL;
+
 	/* create multi node for the subquery */
-	subqueryRangeTableEntry = (RangeTblEntry *) linitial(subqueryEntryList);
+	appendStringInfo(rteName, "citus_subquery_name");
```
onderkalaci: I'd rather move this to a macro. Also, do we need stringInfo? Why cant directly set the alias name from the macro itself?

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1993,62 +2038,107 @@ SubqueryPushdownMultiPlanTree(Query *queryTree)
...
+
+	subqueryRangeTableEntry = makeNode(RangeTblEntry);
```
onderkalaci: Should we set `rtekind` to `RTE_SUBQUERY`?

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1993,62 +2038,107 @@ SubqueryPushdownMultiPlanTree(Query *queryTree)
...
-	Assert(list_length(subqueryEntryList) == 1);
-
-	qualifierColumnList = pull_var_clause_default((Node *) qualifierList);
-	targetListColumnList = pull_var_clause_default((Node *) targetEntryList);
-
-	columnList = list_concat(qualifierColumnList, targetListColumnList);
+	columnList = pull_var_clause_default((Node *) targetEntryList);
 	foreach(columnCell, columnList)
 	{
+		TargetEntry *newTargetEntry = makeNode(TargetEntry);
```
onderkalaci: Here the variable names make it hard to follow. I couldn't exactly get what we're trying to achieve.

* Can we pick better names for `column`, `foundColumn` and `knownColumn`? I cannot follow which one is used for what purpose.
* What is `targetNo` here? Why we're incrementing it in the upper loop and setting it to `column->varattno`?
* Can't we move the first loop to a function and handle the main logic in one step? 
* `newTargetEntry` should be allocated inside the lower if statement.
* `targetNo = 1` line should be seperated by new lines.

src/test/regress/expected/multi_subquery_behavioral_analytics.out
```diff
@@ -0,0 +1,512 @@
...
+-- ALTER SEQUENCE pg_catalog.pg_dist_shardid_seq RESTART 1430000;
+-- ALTER SEQUENCE pg_catalog.pg_dist_jobid_seq RESTART 1430000;
+ 
+SET citus.subquery_pushdown TO TRUE;
+SET citus.enable_router_execution TO FALSE;
+------------------------------------
+-- Vanilla funnel query
+------------------------------------
+SELECT user_id, array_length(events_table, 1)
+FROM (
```
onderkalaci: I really love that we added these regression tests for subquery pushdown.

However, looking into the tests, they are mostly not relevant to the changes (i.e., top level subquery includes a single target entry). So, could you add some regression tests that have top level joins? For instance, mostly using the same tests but removing the top level aggregations?


For instance, removing the top level aggregations from the next test I got an error (even if I remove the UNION and use a single query on the LEFT of the join tree):
```SQL
SELECT *
  FROM (
    (
      SELECT u.user_id, 'step=>1'::text AS event, e.time
      FROM users_table AS u,
          events_table AS e
      WHERE  u.user_id = e.user_id
      AND u.user_id >= 10
      AND u.user_id <= 25
      AND e.event_type IN (100, 101, 102)
    )
    UNION
    (
      SELECT u.user_id, 'step=>2'::text AS event, e.time
      FROM users_table AS u,
         events_table AS e
      WHERE  u.user_id = e.user_id
      AND u.user_id >= 10
      AND u.user_id <= 25
      AND e.event_type IN (103, 104, 105)
    )
  ) t1 LEFT JOIN (
      SELECT DISTINCT user_id, 
        'Has done event'::TEXT AS hasdone_event
      FROM  events_table AS e
      
      WHERE  e.user_id >= 10
      AND e.user_id <= 25
      AND e.event_type IN (106, 107, 108)

  ) t2 ON (t1.user_id = t2.user_id);
```

I got the following error:
```
ERROR:  bogus varno: 3
```

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -399,13 +400,31 @@ ErrorIfQueryNotSupported(Query *queryTree)
 		errorHint = filterHint;
 	}
 
-	if (queryTree->setOperations)
+	if (queryTree->setOperations && !subqueryPushdown)
```
onderkalaci: I think we generally do not prefer to pass a boolean variable to a function and act according to that variable. Do you see an easy way of not doing it?

onderkalaci: I'm thinking that we should have this check in `ErrorIfCannotPushdownSubquery()` function. What do you think?

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -370,7 +371,7 @@ MultiPlanTree(Query *queryTree)
  * more functionality in our distributed planning.
  */
 static void
-ErrorIfQueryNotSupported(Query *queryTree)
+ErrorIfQueryNotSupported(Query *queryTree, bool subqueryPushdown)
```
onderkalaci: We talked with @mtuncer in person and decided to move this check to #1289 on `ErrorIfCannotPushdownSubquery()` function.

src/backend/distributed/planner/multi_logical_planner.c
mtuncer: Per our discussion I am reverting these to be handled when we are doing union work

src/backend/distributed/planner/multi_logical_planner.c
mtuncer: done

src/backend/distributed/planner/multi_logical_planner.c
mtuncer: done

src/backend/distributed/planner/multi_logical_planner.c
mtuncer: done for the most part.

It would have been nice if all newly introduced code to be placed in to a single function, however, since we also change queryTree, when we reassign varno/varattno fields  in target entry list and and havingQuals, I did not want to have a function with side effects.

src/backend/distributed/planner/multi_logical_planner.c
mtuncer: agreed

src/backend/distributed/planner/multi_logical_planner.c
mtuncer: appendStringInfo makes sure this string is palloc'ed.

src/backend/distributed/planner/multi_logical_planner.c
mtuncer: good catch, thanks

src/backend/distributed/planner/multi_logical_planner.c
mtuncer: done

onderkalaci:
@mtuncer Could you please rebase this branch before I start the review? Thanks!

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1993,62 +2038,118 @@ SubqueryPushdownMultiPlanTree(Query *queryTree)
...
 	qualifierList = QualifierList(queryTree->jointree);
 	ValidateClauseList(qualifierList);
 
 	/*
-	 * We disregard pulled subqueries. This changes order of range table list.
-	 * We do not allow subquery joins, so we will have only one range table
-	 * entry in range table list after dropping pulled subquery. For this reason,
-	 * here we are updating columns in the most outer query for where clause
-	 * list and target list accordingly.
+	 * We would be creating a new Query and pushing down top level query's
```
onderkalaci: I think this explanation is good enough, but, it'd be great if we can visualize the structure for quick looks such that:
```
masterQuery -- contains the top level aggregates, limit, having etc.
    lowerQuery -- contains only a single RTE, which is the worker query. The target of this query includes the necessary target list elements to the master query.
        workerQuery  -- contains the query that is being pushed to the 
```

Would you mind adding such a sketch to the end of the comment? (The explanation I've written is not very accurate, I believe you can write better one 😄 )

Lastly, could you add a one more line comment explaining why we've chosen to wrap the worker query into a single RTE? Probably, the comment should start with that sentence.

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -2061,6 +2162,19 @@ SubqueryPushdownMultiPlanTree(Query *queryTree)
 	 * in the logical optimizer.
 	 */
 	extendedOpNode = MultiExtendedOpNode(queryTree);
+
+	/*
+	 * Postgres standard planner converts having qual node to a list of and
+	 * clauses and expects havingQual to be of type List when executing the
```
onderkalaci: I think this comment needs a bit more clarification.

We should be mentioning that in some cases we use original query and in some casses we use the query that has passed through the standard planner. That's why we have this code here.

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -1475,6 +1477,7 @@ BuildSubqueryJobQuery(MultiNode *multiNode)
 	jobQuery->groupClause = groupClauseList;
 	jobQuery->limitOffset = limitOffset;
 	jobQuery->limitCount = limitCount;
+	jobQuery->havingQual = havingQual;
```
onderkalaci: Hmm, minor issue

```C
jobQuery->hasAggs = contain_agg_clause((Node *) targetList) || contain_agg_clause(havingQual);
```

Although we currently don't support having without group bys,  may it be relevant in the future?

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1993,62 +2038,118 @@ SubqueryPushdownMultiPlanTree(Query *queryTree)
...
+			newTargetEntry->expr = (Expr *) copyObject(column);
+
+			appendStringInfo(columnNameString, WORKER_COLUMN_FORMAT, resNo);
+			newTargetEntry->resname = columnNameString->data;
+
+			newTargetEntry->resjunk = false;
+			newTargetEntry->resno = resNo;
+
+			subqueryTargetEntryList = lappend(subqueryTargetEntryList, newTargetEntry);
+			columnNamesList = lappend(columnNamesList, makeString(
```
onderkalaci: Minor: I think we generally have the following format on line breaks:
```C
columnNamesList = 
    lappend(columnNamesList, makeString(columnNameString->data));

```

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1993,62 +2038,118 @@ SubqueryPushdownMultiPlanTree(Query *queryTree)
...
+	lowerQuery->targetList = subqueryTargetEntryList;
+	lowerQuery->jointree = copyObject(queryTree->jointree);
+	lowerQuery->rtable = copyObject(queryTree->rtable);
+
 	/* create multi node for the subquery */
-	subqueryRangeTableEntry = (RangeTblEntry *) linitial(subqueryEntryList);
+	appendStringInfo(rteName, "citus_subquery_name");
+
+	subqueryRangeTableEntry = makeNode(RangeTblEntry);
+	subqueryRangeTableEntry->rtekind = RTE_SUBQUERY;
```
onderkalaci: Hmm, not sure but we may need to set `subqueryRangeTableEntry ->inFromCl` to true?

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1993,62 +2038,118 @@ SubqueryPushdownMultiPlanTree(Query *queryTree)
...
 		column->varno = 1;
+		column->varattno = resNo;
 	}
 
+	/* new query only has target entries, join tree, and rtable*/
+	lowerQuery = makeNode(Query);
+	lowerQuery->commandType = queryTree->commandType;
+	lowerQuery->targetList = subqueryTargetEntryList;
+	lowerQuery->jointree = copyObject(queryTree->jointree);
+	lowerQuery->rtable = copyObject(queryTree->rtable);
```
onderkalaci: Do we need to copy `SetOperations` as well?
Also, I'm a bit worried about `querySource` field since this query involved in deparsing. Do you have any idea on that? Should we copy it as well?

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1993,62 +2038,118 @@ SubqueryPushdownMultiPlanTree(Query *queryTree)
...
 	MultiCollect *subqueryCollectNode = CitusMakeNode(MultiCollect);
 	MultiTable *subqueryNode = NULL;
-	MultiSelect *selectNode = NULL;
 	MultiProject *projectNode = NULL;
 	MultiExtendedOp *extendedOpNode = NULL;
 	MultiNode *currentTopNode = NULL;
 	RangeTblEntry *subqueryRangeTableEntry = NULL;
-	List *subqueryEntryList = SubqueryEntryList(queryTree);
-
+	Query *lowerQuery = NULL;
```
onderkalaci: I'd call this as `projectionQuery` instead of `lowerQuery`. Because, as a reader, I cannot follow which query is this lower than?

But, feel free to overlook my suggestion if doesn't make sense.

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -3081,19 +3081,6 @@ ErrorIfUnsupportedTableCombination(Query *queryTree)
 		}
 	}
 
-	if ((subqueryRangeTableCount > 0) && (relationRangeTableCount > 0))
```
mtuncer: done, for the first part.
I think this current function name is still good to use.

src/backend/distributed/planner/multi_logical_planner.c
mtuncer: This is weird, I can swear I removed those.

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -399,13 +400,31 @@ ErrorIfQueryNotSupported(Query *queryTree)
...
 
-	if (queryTree->setOperations)
+	if (queryTree->setOperations && !subqueryPushdown)
 	{
 		preconditionsSatisfied = false;
 		errorMessage = "could not run distributed query with UNION, INTERSECT, or "
 					   "EXCEPT";
 		errorHint = filterHint;
 	}
+	else if (queryTree->setOperations && subqueryPushdown)
```
mtuncer: removed (again)

src/backend/distributed/planner/multi_logical_planner.c
mtuncer: I will add the visual description too. But what you have here is not accurate.
it is more like
extendedOpNode (where master and worker queries are created from)
```
-->ProjectNode
-->-->CollectNode
-->-->-->MultiTable
-->-->-->-->lowerQuery (this is modified pushed down version of top level query
```

src/backend/distributed/planner/multi_logical_planner.c
mtuncer: ```pushedDownQuery``` ?

onderkalaci: > pushedDownQuery

Sounds good to me

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: Great!

mtuncer: more like
```
	 * Consider query
	 * 		select s1.a, sum(s2.c)
	 * 		from (some subquery) s1, (some subquery) s2
	 * 		where s1.a = s2.a
	 * 		group by s1.a
	 * 		having avg(s2.b);
	 *
	 * We want to prepare a multi tree to avoid subquery joins at top level,
	 * therefore above query is converted to an equivalent
	 * 	 	select worker_column_0, sum(worker_column_1)
	 * 	 	from (
	 * 	 	    	select s1.a as worker_column_0, s2.c as worker_column_1, s2.b as as worker_column_2
	 * 				from (some subquery) s1, (some subquery) s2
	 * 				where s1.a = s2.a
	 * 				) citus_subquery_name
	 * 		group by worker_column_0
	 * 		having avg(worker_column_2);
	 * 	After this conversion MultiTree is created as follows
	 * 	
	 * 	MultiExtendedOpNode(
	 * 		targetList : worker_column_0, sum(worker_column_1)
	 * 		groupBy : worker_column_0
	 * 		having :  avg(worker_column_2))
	 * --->MultiProject (worker_column_0, worker_column_1, worker_column_2)
	 * --->--->	MultiTable (subquery : citus_subquery_name)
	 * 
	 * Master and worker queries will be created out of this MultiTree at later stages.
```

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1993,62 +2038,118 @@ SubqueryPushdownMultiPlanTree(Query *queryTree)
...
+	 * in the upper query will be updated to have varno=1, and varattno=<resno>
+	 * of matching target entry in lower query.
 	 */
-	Assert(list_length(subqueryEntryList) == 1);
-
-	qualifierColumnList = pull_var_clause_default((Node *) qualifierList);
-	targetListColumnList = pull_var_clause_default((Node *) targetEntryList);
-
-	columnList = list_concat(qualifierColumnList, targetListColumnList);
+	columnList = pull_var_clause_default((Node *) targetEntryList);
```
mtuncer: ok

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1993,62 +2038,118 @@ SubqueryPushdownMultiPlanTree(Query *queryTree)
...
+		bool needsNewTargetEntry = true;
+		AttrNumber resNo = 1;
+
+		/*
+		 * Search if we previously processed this column. We would create a new
+		 * target entry when we encounter the column for the first time. resNo
+		 * is set to resno of the matching target entry for this column in
+		 * lower level query.
+		 */
+		foreach(addedColumnCell, addedColumnList)
```
mtuncer: I think it get complicated since I was trying to 2 things at once. I needed to create a target entry for each unique column at the same time updating varattno fields of all columns. Perhaps I should make that in multiple passes.

src/test/regress/expected/multi_subquery_behavioral_analytics.out
```diff
@@ -0,0 +1,997 @@
...
+      51 |               1 |        1 |                   1 |                1 |                  
+      85 |               6 |        6 |                   6 |                  |                  
+      73 |               4 |        4 |                   4 |                  |                  
+       0 |               3 |        3 |                   3 |                  |                  
+      10 |               2 |        2 |                   2 |                  |                  
+      13 |               2 |        2 |                   2 |                  |                  
+(15 rows)
+
+-- Same query without all limitations but uses having() to show only those submitted their credit card info
+SELECT
```
onderkalaci: Having such tests is really exciting!

But, could we also add a few failure tests such that join is not equi join or not on the partition key etc?

Also, could you test referring from several levels below for the LATERAL joins. For example, let's say `e5` refers to `e1` (The query probably won't make much sense, but, I'd like to make sure that the codes that we've written for `LATERAL`s works as expected)

src/test/regress/expected/multi_subquery_complex_queries.out
```diff
@@ -683,6 +768,45 @@ FROM
...
+   ) "some_recent_users"
+   JOIN LATERAL
+   (SELECT "users".user_id
+    FROM users_table as "users"
+    WHERE "users"."user_id" = "some_recent_users"."user_id"
+        AND users.value_2 > 200
+    LIMIT 1
+   ) "some_users_data" ON TRUE
+ORDER BY value_3 DESC
+LIMIT 10;
```
onderkalaci: To ensure getting the same results per run, could you also add `user_id` to the ORDER BY? To the others as well?

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -2197,22 +2169,87 @@ ErrorIfSubqueryJoin(Query *queryTree)
...
+											 List **columnNamesList)
+{
+	AttrNumber resNo = 1;
+
+	ListCell *columnCell = NULL;
+	foreach(columnCell, uniqueColumnList)
+	{
+		Var *column = (Var *) lfirst(columnCell);
+		TargetEntry *newTargetEntry = makeNode(TargetEntry);
+		StringInfo columnNameString = makeStringInfo();
```
onderkalaci: Maybe a single new-line after this?

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -2197,22 +2169,87 @@ ErrorIfSubqueryJoin(Query *queryTree)
...
+		newTargetEntry->expr = (Expr *) copyObject(column);
+		appendStringInfo(columnNameString, WORKER_COLUMN_FORMAT, resNo);
+		newTargetEntry->resname = columnNameString->data;
+		newTargetEntry->resjunk = false;
+		newTargetEntry->resno = resNo;
+		*subqueryTargetEntryList = lappend(*subqueryTargetEntryList, newTargetEntry);
```
onderkalaci: maybe another newline before `subqueryTargetEntryList = = lappend(*subqueryTargetEntryList, newTargetEntry)`

mtuncer: done

src/backend/distributed/planner/multi_logical_planner.c
mtuncer: done

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -2086,22 +2169,89 @@ ErrorIfSubqueryJoin(Query *queryTree)
...
+
+
+/*
+ * UpdateVarMappingsForExtendedOpNode updates varno/varattno fields of columns
+ * in columnList to point to corresponding location in uniqueColumnList. The
+ * subquery target entry list is also created in similar manner to make sure
+ * varattno field from column list matches target entry lists resno field.
+ */
+static void
+UpdateVarMappingsForExtendedOpNode(List *columnList, List *uniqueColumnList)
```
onderkalaci: I'm noting what we've talked in person. It probably would make easier to read the code once we pass `subqueryTargetEntryList` instead of `uniqueColumnList`. I think this is my final comment.

mtuncer: Sounds good, that triggered more changes and some code removal.

onderkalaci:
This looks great :shipit:

PR #1299, opened 2017-03-29
Add disable/enable trigger all support

### LINKED ISSUE #1080: Support DISABLE/ENABLE TRIGGER ALL on distributed tables
This is a missing feature for those using Citus together with Ruby on Rails, and using a framework like [database_cleaner](https://github.com/DatabaseCleaner/database_cleaner/blob/master/lib/database_cleaner/active_record/truncation.rb#L235) to truncate their tables after test suite runs. In particular, Rails has a method called [disable_referential_integrity](https://github.com/rails/rails/blob/92703a9ea5d8b96f30e0b706b801c9185ef14f0e/activerecord/lib/active_record/connection_adapters/postgresql/referential_integrity.rb#L15) that runs the following SQL on each table known to Rails:

```sql
ALTER TABLE ... DISABLE TRIGGER ALL
-- other SQL here
ALTER TABLE ... ENABLE TRIGGER ALL
```

This happens after a test run to remove the test run's data without executing triggers.

Even though we don't support triggers (https://github.com/citusdata/citus/issues/906) this causes errors when running the tests for a Rails app, since its always being called, including for tables that have no triggers.

It would be helpful if we supported this as a no-op for now, and push it down once https://github.com/citusdata/citus/issues/906 is implemented.

metdos:
We can actually push down disable/enable trigger all, just escaped them from the error check. I didn't add them to supported ALTER TABLE .. error message because they are not frequently used commands. I added regression tests to verify behavior.

Fixes #1080.

### DIFF
--- a/src/backend/distributed/executor/multi_utility.c
+++ b/src/backend/distributed/executor/multi_utility.c
@@ -1650,8 +1650,10 @@ ErrorIfUnsupportedAlterTableStmt(AlterTableStmt *alterTableStatement)
 			}
 
 			case AT_DropConstraint:
+			case AT_EnableTrigAll:
+			case AT_DisableTrigAll:
 			{
-				/* we will no perform any special check for ALTER TABLE DROP CONSTRAINT */
+				/* we will not perform any special checks for these ALTER TABLE types */
 				break;
 			}
 
--- a/src/test/regress/input/multi_alter_table_statements.source
+++ b/src/test/regress/input/multi_alter_table_statements.source
@@ -380,6 +380,45 @@ END;
 
 DROP TABLE sequence_deadlock_test;
 
+-- verify enable/disable trigger all works
+
+SET citus.shard_replication_factor TO 1;
+SET citus.shard_count TO 1;
+
+CREATE TABLE trigger_table (
+	id int,
+	value text
+);
+
+SELECT create_distributed_table('trigger_table', 'id');
+
+-- first set a trigger on a shard
+\c - - - :worker_1_port
+
+CREATE FUNCTION update_value() RETURNS trigger AS $up$
+    BEGIN
+		NEW.value := 'trigger enabled';
+		RETURN NEW;
+    END;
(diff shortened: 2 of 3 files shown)


src/backend/distributed/executor/multi_utility.c
```diff
@@ -1650,8 +1650,10 @@ ErrorIfUnsupportedAlterTableStmt(AlterTableStmt *alterTableStatement)
 			}
 
 			case AT_DropConstraint:
+			case AT_EnableTrigAll:
+			case AT_DisableTrigAll:
 			{
-				/* we will no perform any special check for ALTER TABLE DROP CONSTRAINT */
+				/* we will not perform any special checks for these ALTER TABLE types */
 				break;
 			}
```
marcocitus: I guess the `errdetail` message below `default:` is getting a little out-of-date.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -1650,8 +1650,10 @@ ErrorIfUnsupportedAlterTableStmt(AlterTableStmt *alterTableStatement)
...
 
 			case AT_DropConstraint:
+			case AT_EnableTrigAll:
+			case AT_DisableTrigAll:
```
marcocitus: Are there any other commands that we can trivially add here?

e.g. 
- `AT_EnableTrigUser`
- `AT_DisableTrigUser`
- `AT_SetLogged`
- `AT_SetUnlogged`
- `AT_ChangeOwner`

metdos:
> Are there any other commands that we can trivially add here?
> 
> e.g.
> 
> AT_EnableTrigUser
> AT_DisableTrigUser
> AT_SetLogged
> AT_SetUnlogged
> AT_ChangeOwner

Yeap, some of them also crossed my mind. Just, I was more concerned about the testing effort.

metdos:
> Looks good to me 🚢

I really like this type of volunteer code reviews, thanks :D

PR #1300, opened 2017-03-30
Change copy format of ltree

### LINKED ISSUE #1231: COPY with ltree data type column errors out
**Citus Version:** Latest 6.1 release
```
CREATE TABLE test (
  col1 ltree NOT NULL,
  col2 character varying(255) NOT NULL
);
SELECT create_reference_table('test');
```
```
\COPY test FROM 'test.csv' WITH (format CSV, header TRUE);
ERROR:  no binary output function available for type ltree
```
**Hypothesis**: The function appears to check for composite, array or user-defined types. It should check if the types have a binary output function.
https://github.com/citusdata/citus/blob/master/src/backend/distributed/commands/multi_copy.c#L934
**Suggested Fix**: Add check for the presence of binary output functions for columns in the table for which the COPY is occurring.

velioglu:
Fixes #1231.

Checks whether each column of a tuple has binary output function defined. If any of them doesn't have, fallback to non-binary protocol.

### DIFF
--- a/src/backend/distributed/commands/multi_copy.c
+++ b/src/backend/distributed/commands/multi_copy.c
@@ -93,8 +93,8 @@ static void OpenCopyConnections(CopyStmt *copyStatement,
 								ShardConnections *shardConnections, bool stopOnFailure,
 								bool useBinaryCopyFormat);
 
-static bool CanUseBinaryCopyFormat(TupleDesc tupleDescription,
-								   CopyOutState rowOutputState);
+static bool CanUseBinaryCopyFormat(TupleDesc tupleDescription);
+static bool BinaryOutputFunctionDefined(Oid typeId);
 static List * MasterShardPlacementList(uint64 shardId);
 static List * RemoteFinalizedShardPlacementList(uint64 shardId);
 
@@ -464,7 +464,7 @@ CopyToNewShards(CopyStmt *copyStatement, char *completionTag, Oid relationId)
 	copyOutState->delim = (char *) delimiterCharacter;
 	copyOutState->null_print = (char *) nullPrintCharacter;
 	copyOutState->null_print_client = (char *) nullPrintCharacter;
-	copyOutState->binary = CanUseBinaryCopyFormat(tupleDescriptor, copyOutState);
+	copyOutState->binary = CanUseBinaryCopyFormat(tupleDescriptor);
 	copyOutState->fe_msgbuf = makeStringInfo();
 	copyOutState->rowcontext = executorTupleContext;
 
@@ -809,14 +809,15 @@ OpenCopyConnections(CopyStmt *copyStatement, ShardConnections *shardConnections,
 
 
 /*
- * CanUseBinaryCopyFormat iterates over columns of the relation given in rowOutputState
- * and looks for a column whose type is array of user-defined type or composite type.
- * If it finds such column, that means we cannot use binary format for COPY, because
- * binary format sends Oid of the types, which are generally not same in master and
- * worker nodes for user-defined types.
+ * CanUseBinaryCopyFormat iterates over columns of the relation and looks for a
+ * column whose type is array of user-defined type or composite type. If it finds
+ * such column, that means we cannot use binary format for COPY, because binary
+ * format sends Oid of the types, which are generally not same in master and
+ * worker nodes for user-defined types. If the function can not detect a binary
+ * output function for any of the column, it returns false.
  */
 static bool
-CanUseBinaryCopyFormat(TupleDesc tupleDescription, CopyOutState rowOutputState)
+CanUseBinaryCopyFormat(TupleDesc tupleDescription)
 {
 	bool useBinaryCopyFormat = true;
 	int totalColumnCount = tupleDescription->natts;
@@ -828,6 +829,7 @@ CanUseBinaryCopyFormat(TupleDesc tupleDescription, CopyOutState rowOutputState)
 		Oid typeId = InvalidOid;
 		char typeCategory = '\0';
 		bool typePreferred = false;
+		bool binaryOutputFunctionDefined = false;
 
 		if (currentColumn->attisdropped)
 		{
@@ -835,6 +837,15 @@ CanUseBinaryCopyFormat(TupleDesc tupleDescription, CopyOutState rowOutputState)
 		}
 
 		typeId = currentColumn->atttypid;
+
+		/* built-in types may also don't have binary output function */
+		binaryOutputFunctionDefined = BinaryOutputFunctionDefined(typeId);
+		if (!binaryOutputFunctionDefined)
+		{
+			useBinaryCopyFormat = false;
+			break;
+		}
+
 		if (typeId >= FirstNormalObjectId)
 		{
 			get_type_category_preferred(typeId, &typeCategory, &typePreferred);
@@ -848,6 +859,32 @@ CanUseBinaryCopyFormat(TupleDesc tupleDescription, CopyOutState rowOutputState)
 	}
 
 	return useBinaryCopyFormat;
+}
+
+
+/*
+ * BinaryOutputFunctionDefined checks whether binary output function is defined
+ * for the given type.
+ */
+static bool
+BinaryOutputFunctionDefined(Oid typeId)
+{
+	Oid typeFunctionId = InvalidOid;
+	Oid typeIoParam = InvalidOid;
+	int16 typeLength = 0;
+	bool typeByVal = false;
+	char typeAlign = 0;
+	char typeDelim = 0;
+
+	get_type_io_data(typeId, IOFunc_send, &typeLength, &typeByVal,
+					 &typeAlign, &typeDelim, &typeIoParam, &typeFunctionId);
+
+	if (OidIsValid(typeFunctionId))
+	{
+		return true;
+	}
+
+	return false;
 }
 
 
@@ -1732,7 +1769,7 @@ CitusCopyDestReceiverStartup(DestReceiver *dest, int operation,
 	copyOutState->delim = (char *) delimiterCharacter;
 	copyOutState->null_print = (char *) nullPrintCharacter;
 	copyOutState->null_print_client = (char *) nullPrintCharacter;
-	copyOutState->binary = CanUseBinaryCopyFormat(inputTupleDescriptor, copyOutState);
+	copyOutState->binary = CanUseBinaryCopyFormat(inputTupleDescriptor);
 	copyOutState->fe_msgbuf = makeStringInfo();
 	copyOutState->rowcontext = GetPerTupleMemoryContext(copyDest->executorState);
 	copyDest->copyOutState = copyOutState;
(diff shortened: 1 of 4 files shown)


anarazel:
On 2017-03-30 05:11:01 -0700, Burak Velioglu wrote:
> Fixes #1231.

This doesn't look like the right fix to me. I think we rather should
check whether a binary output function exists.

marcocitus:
> This doesn't look like the right fix to me. I think we rather should check whether a binary output function exists.

What if the user-defined type is wacky like composite/array and contains OIDs? This fix seems safer.

anarazel:
On 2017-03-30 08:47:04 -0700, Marco Slot wrote:
> > This doesn't look like the right fix to me. I think we rather should check whether a binary output function exists.
> 
> What if the user-defined type is wacky like composite/array and contains OIDs? This fix seems safer.

On the other hand it's orders of magnitude slower for a bunch of things.
And there are builtin types without a binary output function.

src/backend/distributed/commands/multi_copy.c
```diff
@@ -93,8 +94,8 @@ static void OpenCopyConnections(CopyStmt *copyStatement,
 								ShardConnections *shardConnections, bool stopOnFailure,
 								bool useBinaryCopyFormat);
 
-static bool CanUseBinaryCopyFormat(TupleDesc tupleDescription,
-								   CopyOutState rowOutputState);
+static bool CanUseBinaryCopyFormat(TupleDesc tupleDescription);
+static bool BinaryOutputFunctionIsDefined(Oid typeId);
```
byucesoy: Just remove the "Is" from the function name. We generally use question form and drop the is, has,can etc. (CanUseBinaryCopyFormat is an exception, because otherwise it becomes difficult to understand what it does)

src/backend/distributed/commands/multi_copy.c
```diff
@@ -828,13 +830,23 @@ CanUseBinaryCopyFormat(TupleDesc tupleDescription, CopyOutState rowOutputState)
 		Oid typeId = InvalidOid;
 		char typeCategory = '\0';
 		bool typePreferred = false;
+		bool binaryOutputFunctionIsDefined = false;
```
byucesoy: Similarly, no need for "is"

src/backend/distributed/commands/multi_copy.c
```diff
@@ -852,6 +864,34 @@ CanUseBinaryCopyFormat(TupleDesc tupleDescription, CopyOutState rowOutputState)
...
 
 /*
+ * BinaryOutputFunctionIsDefined checks whether binary output function is defined
+ * for the given type.
+ */
+static bool
+BinaryOutputFunctionIsDefined(Oid typeId)
+{
+	HeapTuple typeTuple;
+	Form_pg_type pt;
```
byucesoy: We should initialize this and the variable above. Also we need more descriptive name than pt.

src/backend/distributed/commands/multi_copy.c
```diff
@@ -852,6 +864,34 @@ CanUseBinaryCopyFormat(TupleDesc tupleDescription, CopyOutState rowOutputState)
...
+	bool defineBinaryOutputFunction = false;
```
byucesoy: binaryOutputFunctionDefined would be more inline with the function's name.

src/test/regress/expected/multi_copy_with_ltree.out
```diff
@@ -0,0 +1,14 @@
+ALTER SEQUENCE pg_catalog.pg_dist_shardid_seq RESTART 1430000;
```
byucesoy: I think we do not need to new test file for this. You can add your tests at multi_copy.source

src/test/regress/expected/multi_copy_with_ltree.out
```diff
@@ -0,0 +1,14 @@
...
+ALTER SEQUENCE pg_catalog.pg_dist_jobid_seq RESTART 1430000;
+CREATE EXTENSION ltree;
```
byucesoy: Why do we try to create the extension here. It is already created.

src/backend/distributed/commands/multi_copy.c
```diff
@@ -852,6 +864,34 @@ CanUseBinaryCopyFormat(TupleDesc tupleDescription, CopyOutState rowOutputState)
...
+
+	pt = (Form_pg_type) GETSTRUCT(typeTuple);
+	if (pt->typisdefined && OidIsValid(pt->typsend))
+	{
+		defineBinaryOutputFunction = true;
+	}
+
+	ReleaseSysCache(typeTuple);
+	return defineBinaryOutputFunction;
+}
```
anarazel: Have you checked using get_type_io_data() instead?

src/test/regress/Makefile
```diff
@@ -46,7 +46,7 @@ check-worker: all
 	-- $(MULTI_REGRESS_OPTS) --schedule=$(citus_abs_srcdir)/worker_schedule $(EXTRA_TESTS)
 
 check-multi: all tempinstall-main
-	$(pg_regress_multi_check) --load-extension=citus \
+	$(pg_regress_multi_check) --load-extension=citus --load-extension=ltree \
```
byucesoy: We still have some other test where we need to add ltree

src/backend/distributed/commands/multi_copy.c
```diff
@@ -809,14 +810,15 @@ OpenCopyConnections(CopyStmt *copyStatement, ShardConnections *shardConnections,
...
- * CanUseBinaryCopyFormat iterates over columns of the relation given in rowOutputState
- * and looks for a column whose type is array of user-defined type or composite type.
- * If it finds such column, that means we cannot use binary format for COPY, because
- * binary format sends Oid of the types, which are generally not same in master and
- * worker nodes for user-defined types.
+ * CanUseBinaryCopyFormat iterates over columns of the relation and looks for a
+ * column whose type is array of user-defined type or composite type. If it finds
+ * such column, that means we cannot use binary format for COPY, because binary
+ * format sends Oid of the types, which are generally not same in master and
+ * worker nodes for user-defined types. Function also error outs if it can not
```
byucesoy: This function does not error out. It just returns false if a column does not have binary output function.

src/backend/distributed/commands/multi_copy.c
```diff
@@ -852,6 +864,32 @@ CanUseBinaryCopyFormat(TupleDesc tupleDescription, CopyOutState rowOutputState)
 
 
 /*
+ * BinaryOutputFunctionIsDefined checks whether binary output function is defined
```
metdos: Minor: Function name in the comment is slightly different.

src/backend/distributed/commands/multi_copy.c
```diff
@@ -852,6 +864,32 @@ CanUseBinaryCopyFormat(TupleDesc tupleDescription, CopyOutState rowOutputState)
...
+ */
+static bool
+BinaryOutputFunctionDefined(Oid typeId)
+{
+	Oid typeFunctionId = InvalidOid;
+	Oid typeIoParam = InvalidOid;
+	int16 typeLength = 0;
+	bool typeByVal = false;
+	char typeAlign = '0';
+	char typeDelim = '0';
```
metdos: Could you assign char variables directly to 0? '0' is a random character.

Such as: https://github.com/citusdata/citus/blob/33b58f8e267ceabc81a764042226b70c1a856b1c/src/backend/distributed/commands/create_distributed_table.c#L249

src/backend/distributed/commands/multi_copy.c
```diff
@@ -74,6 +74,7 @@
 #include "utils/lsyscache.h"
 #include "utils/rel.h"
 #include "utils/memutils.h"
+#include "utils/syscache.h"
```
metdos: Adding `utils/syscache.h` seems unnecessary.

src/test/regress/input/multi_copy.source
```diff
@@ -749,3 +749,15 @@ SELECT shardid, shardstate, nodename, nodeport
...
+-- Test with ltree column
+CREATE TABLE test_ltree (path ltree);
+SELECT create_reference_table('test_ltree');
+
+COPY test_ltree FROM STDIN DELIMITER AS ',';
+Top
+Top.Science
+Top.Science.Astronomy
+\.
+
```
metdos: Maybe add a simple `SELECT` to check copied data, such as;

```sql
SELECT path, nlevel(path) AS depth FROM test ORDER BY depth;
```

metdos:
> I think this PR looks good. Just fix the comment of CanUseBinaryCopyFormat
> 
> @metdos My review is done, you can have a look

Looks good, 🚢

onderkalaci:
Guys, @byucesoy @velioglu @metdos , when I rebased by PR to this one, and try running the regression tests I got `ERROR:  could not open extension control file "/usr/local/pgsql/share/extension/ltree.control": No such file or directory`

Do we now have a dependency to `ltree` extension?

velioglu:
@onderkalaci Yes, I have added the --load-extension=ltree environment variable to perl script to load the ltree extension.

metdos:
> Guys, @byucesoy @velioglu @metdos , when I rebased by PR to this one, and try running the regression tests I got ERROR: could not open extension control file "/usr/local/pgsql/share/extension/ltree.control": No such file or directory
> 
> Do we now have a dependency to ltree extension?

@onderkalaci, I have not noticed this. @velioglu, we should remove this, apparently, this makes running regression tests harder at local machines. I presume Travis installs the extension automatically for us, so we missed it?

onderkalaci:
>  I presume Travis installs the extension automatically for us, so we missed it?

I think so.

> @velioglu, we should remove this, apparently, this makes running regression tests harder at local machines 

How does `hll` extension installed? Should we follow the same pattern here as well?

metdos:
> How does hll extension installed? Should we follow the same pattern here as well?

Regression test itself creates it, but still requires to load related files (hll.control etc).

byucesoy:
ltree is part of PostgreSQL/contrib. It is already in your filesystem but to run tests, you only need to compile and install it.

However, I think it is possible to remove that. ltree creates the type as follows:

```sql
CREATE TYPE ltree (
	INTERNALLENGTH = -1,
	INPUT = ltree_in,
	OUTPUT = ltree_out,
	STORAGE = extended
);
```

So it is not composite type, thus it passes binary output usage checks(before this PR), we can also create a type similar to this in our tests and use it instead.

metdos:
@velioglu I think you can replace ltree with one of the next types.

```SQL
postgres=# select oid, typname, typtype, typinput, typoutput, typreceive, typsend from pg_type where typreceive = '-'::regproc and typtype = 'b';
 oid  |  typname  | typtype |  typinput   |  typoutput   | typreceive | typsend
------+-----------+---------+-------------+--------------+------------+---------
  210 | smgr      | b       | smgrin      | smgrout      | -          | -
 1033 | aclitem   | b       | aclitemin   | aclitemout   | -          | -
 3642 | gtsvector | b       | gtsvectorin | gtsvectorout | -          | -
(3 rows)
```

PR #1301, opened 2017-03-30
Subquery pushdown - Unions Phase - 2

onderkalaci:
(The PR currently is based on a branch which is not rebased to its parent branch. Thus, we see some older commits. As we progress to review this patch, I'll rebase all PRs accordingly):

    Add top level union support
    
    This commit adds support for UNION/UNION ALL subqueries that are
    in the following form:
    
         .... (Q1 UNION Q2 UNION ...) as union_query ....
    
    In other words, Citus supports allow top level
    unions being wrapped into aggregations queries
    and/or simple projection queries that only selects
    some fields from the lower level queries.

### DIFF
--- a/src/backend/distributed/planner/multi_logical_optimizer.c
+++ b/src/backend/distributed/planner/multi_logical_optimizer.c
@@ -150,7 +150,8 @@ static bool GroupedByColumn(List *groupClauseList, List *targetList, Var *column
 /* Local functions forward declarations for subquery pushdown checks */
 static void ErrorIfContainsUnsupportedSubquery(MultiNode *logicalPlanNode,
 											   PlannerRestrictionContext *
-											   plannerRestrictionContext);
+											   plannerRestrictionContext,
+											   Query *originalQuery);
 static void ErrorIfCannotPushdownSubquery(Query *subqueryTree, bool outerQueryHasLimit);
 static void ErrorIfUnsupportedSetOperation(Query *subqueryTree, bool outerQueryHasLimit);
 static bool ExtractSetOperationStatmentWalker(Node *node, List **setOperationList);
@@ -191,14 +192,15 @@ static bool HasOrderByHllType(List *sortClauseList, List *targetList);
  * function finds the extended operator node, and splits this node into master
  * and worker extended operator nodes.
  *
- * We also pass plannerRestrictionContext to the optimizer. The context
- * is primarily used to decide whether the subquery is safe to pushdown.
+ * We also pass plannerRestrictionContext and originalQuery to the optimizer.
+ * These are primarily used to decide whether the subquery is safe to pushdown.
  * If not, it helps to produce meaningful error messages for subquery
  * pushdown planning.
  */
 void
 MultiLogicalPlanOptimize(MultiTreeRoot *multiLogicalPlan,
-						 PlannerRestrictionContext *plannerRestrictionContext)
+						 PlannerRestrictionContext *plannerRestrictionContext,
+						 Query *originalQuery)
 {
 	bool hasOrderByHllType = false;
 	List *selectNodeList = NIL;
@@ -218,7 +220,8 @@ MultiLogicalPlanOptimize(MultiTreeRoot *multiLogicalPlan,
 	ErrorIfContainsUnsupportedAggregate(logicalPlanNode);
 
 	/* check that we can pushdown subquery in the plan */
-	ErrorIfContainsUnsupportedSubquery(logicalPlanNode, plannerRestrictionContext);
+	ErrorIfContainsUnsupportedSubquery(logicalPlanNode, plannerRestrictionContext,
+									   originalQuery);
 
 	/*
 	 * If a select node exists, we use the idempower property to split the node
@@ -2807,20 +2810,22 @@ GroupedByColumn(List *groupClauseList, List *targetList, Var *column)
 
 /*
  * ErrorIfContainsUnsupportedSubquery extracts subquery multi table from the
- * logical plan and uses helper functions to check if we can push down subquery
- * to worker nodes. These helper functions error out if we cannot push down the
- * the subquery.
+ * logical plan using plannerRestrictionContext and the original query. It uses
+ * some helper functions to check if we can push down subquery to worker nodes.
+ * These helper functions error out if we cannot push down the subquery.
  */
 static void
 ErrorIfContainsUnsupportedSubquery(MultiNode *logicalPlanNode,
-								   PlannerRestrictionContext *plannerRestrictionContext)
+								   PlannerRestrictionContext *plannerRestrictionContext,
+								   Query *originalQuery)
 {
 	Query *subquery = NULL;
 	List *extendedOpNodeList = NIL;
 	MultiTable *multiTable = NULL;
 	MultiExtendedOp *extendedOpNode = NULL;
 	bool outerQueryHasLimit = false;
-	bool restrictionEquivalenceForPartitionKeys = false;
+	RelationRestrictionContext *relationRestrictionContext =
+		plannerRestrictionContext->relationRestrictionContext;
 
 	/* check if logical plan includes a subquery */
 	List *subqueryMultiTableList = SubqueryMultiTableList(logicalPlanNode);
@@ -2832,9 +2837,26 @@ ErrorIfContainsUnsupportedSubquery(MultiNode *logicalPlanNode,
 	/* currently in the planner we only allow one subquery in from-clause*/
 	Assert(list_length(subqueryMultiTableList) == 1);
 
-	restrictionEquivalenceForPartitionKeys =
-		RestrictionEquivalenceForPartitionKeys(plannerRestrictionContext);
-	if (!restrictionEquivalenceForPartitionKeys)
+	/*
+	 * We're checking two things here:
+	 *    (i)   If the query contains a top level union, ensure that all leaves
+	 *          return the partition key at the same position
+	 *    (ii)  Else, check whether all relations joined on the partition key or not
+	 */
+	if (ContainsUnionSubquery(originalQuery))
+	{
+		if (!SafeToPushdownUnionSubquery(relationRestrictionContext))
+		{
+			ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+							errmsg("cannot pushdown the subquery since all leaves of "
+								   "the UNION does not include partition key at the "
+								   "same position"),
+							errdetail("Each leaf query of the UNION should return "
+									  "partition key at the same position on its "
+									  "target list.")));
+		}
+	}
+	else if (!RestrictionEquivalenceForPartitionKeys(plannerRestrictionContext))
 	{
 		ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
 						errmsg("cannot pushdown the subquery since all relations are not "
--- a/src/backend/distributed/planner/multi_physical_planner.c
+++ b/src/backend/distributed/planner/multi_physical_planner.c
@@ -2008,14 +2008,6 @@ SubquerySqlTaskList(Job *job, PlannerRestrictionContext *plannerRestrictionConte
 	DistTableCacheEntry *targetCacheEntry = NULL;
 	RelationRestrictionContext *relationRestrictionContext =
 		plannerRestrictionContext->relationRestrictionContext;
-	bool restrictionEquivalenceForPartitionKeys PG_USED_FOR_ASSERTS_ONLY =
-		RestrictionEquivalenceForPartitionKeys(plannerRestrictionContext);
-
-	/*
-	 * If we're going to create tasks for subquery pushdown, all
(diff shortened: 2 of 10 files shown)


src/backend/distributed/planner/multi_planner.c
```diff
@@ -117,6 +117,8 @@ multi_planner(Query *parse, int cursorOptions, ParamListInfo boundParams)
 
 		if (needsDistributedPlanning)
 		{
+			relationRestrictionContext->parseTree = copyObject(originalQuery);
```
onderkalaci: I think we should move this line to `CreateAndPushPlannerContexts ()` to make it more readable

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -2832,9 +2832,9 @@ ErrorIfContainsUnsupportedSubquery(MultiNode *logicalPlanNode,
 	/* currently in the planner we only allow one subquery in from-clause*/
 	Assert(list_length(subqueryMultiTableList) == 1);
 
-	restrictionEquivalenceForPartitionKeys =
-		RestrictionEquivalenceForPartitionKeys(plannerRestrictionContext);
-	if (!restrictionEquivalenceForPartitionKeys)
+	safeToPushdownSubquery =
+		SafeToPushDownSubquery(plannerRestrictionContext);
+	if (!safeToPushdownSubquery)
```
onderkalaci: I don't like this generic error message. We should differentiate unions and joins, at least,

src/backend/distributed/planner/relation_restriction_equivalence.c
```diff
@@ -1052,3 +1259,93 @@ AttributeEquivalancesAreEqual(AttributeEquivalenceClass *firstAttributeEquivalan
...
+static bool
+TopLevelUnionQuery(Query *queryTree)
+{
+	List *rangeTableList = queryTree->rtable;
+	Node *setOperations = queryTree->setOperations;
+	RangeTblEntry *rangeTableEntry = NULL;
+	Query *subqueryTree = NULL;
+
+	/* don't allow joines on top of unions */
+	if (list_length(rangeTableList) > 1)
```
mtuncer: I am not following this. 
`query->rtable `has nothing to do with joins.  joined range table entries can be found from iterating through `query->tree`

mtuncer: I wonder if we can support 
```sql
select user_id from users_table where user_id < 10
union
select user_id from users_table where user_id < 10;
```

type queries. This work does not do you see a blocker on supporting this ?

src/backend/distributed/planner/relation_restriction_equivalence.c
```diff
@@ -1052,3 +1259,93 @@ AttributeEquivalancesAreEqual(AttributeEquivalenceClass *firstAttributeEquivalan
...
+
+		if (setOperationStatement->op != SETOP_UNION)
+		{
+			return false;
+		}
+
+		return true;
+	}
+
+	return TopLevelUnionQuery(subqueryTree);
```
mtuncer: I did not like recursion here.  I think this is used to traverse query tree to find out the first query where there is either set operation or join. 

Can you move search logic to another walker like function and remove  recursion here ?

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -2053,6 +2053,7 @@ SubquerySqlTaskList(Job *job, PlannerRestrictionContext *plannerRestrictionConte
 		subqueryTask = SubqueryTaskCreate(subquery, targetShardInterval,
 										  relationRestrictionContext, taskIdIndex);
 
+
```
onderkalaci: Unnecessary new line

src/backend/distributed/planner/relation_restriction_equivalence.c
```diff
@@ -113,6 +122,187 @@ static void ListConcatUniqueAttributeClassMemberLists(AttributeEquivalenceClass
...
+		{
+			Var *targetVar = (Var *) lfirst(translatedVarCell);
+
+			childAttrNumber++;
+
+			if (targetVar->varno == relationRteIndex &&
+				targetVar->varattno == relationPartitionKey->varattno)
+			{
+				*partitionKeyIndex = childAttrNumber;
+
```
onderkalaci: Can't we use `targetVar ->varattno`? That seems more accurate to me.

src/backend/distributed/planner/relation_restriction_equivalence.c
onderkalaci: > type queries. This work does not, do you see a blocker on supporting this ?

I theory, we can support that. But, I don't want to deal with !subquery code-path in this PR/release. What do you think?

src/backend/distributed/planner/relation_restriction_equivalence.c
onderkalaci: Nope, how we use seems OK. We only need to ensure that it is actually a `Var`

src/backend/distributed/planner/relation_restriction_equivalence.c
mtuncer: I think it won't bring any more complexity, we would just need to move down in the query funnel until there is a set operation or join.

onderkalaci: > I think it won't bring any more complexity, we would just need to move down in the query funnel until there is a set operation or join.

I still think that we should tackle it separately, maybe with another PR on top of this?

onderkalaci: Hmm, it could be that the name of the function is a bit confusing. This function is looking for a query tree such that the query may have as many upper level queries above the `UNION` query, unless none of them is a join.

Is the function comment not clear enough for that?

src/backend/distributed/planner/relation_restriction_equivalence.c
```diff
@@ -113,6 +119,193 @@ static void ListConcatUniqueAttributeClassMemberLists(AttributeEquivalenceClass
...
+}
+
+
+/*
+ * FindTranslatedVar iterates on the appendRelList and tries to find a translated
+ * child var identified by the relation id and the relation rte index. The function
+ * returns NULL if it cannot find a translated var.
+ */
+static Var *
+FindTranslatedVar(List *appendRelList, Oid relationOid, Index relationRteIndex,
```
mtuncer: Is it the Var was already translated, and we are trying to find translation ? 

Function could be better named. Could you rephrase explanation.

src/backend/distributed/planner/relation_restriction_equivalence.c
```diff
@@ -113,6 +119,193 @@ static void ListConcatUniqueAttributeClassMemberLists(AttributeEquivalenceClass
...
+
+/*
+ * SafeToPushdownTopLevelUnion returns true if all the relations are returns
+ * partition keys in the same ordinal position.
+ *
+ * Note that the function expects (and asserts) the input query to be a top
+ * level union query defined by TopLevelUnionQuery().
+ */
+bool
+SafeToPushdownTopLevelUnion(RelationRestrictionContext *restrictionContext)
```
mtuncer: perhaps SafeToPushdownSubqueryUnion  or SafeToPushdownUnionSubquery

src/backend/distributed/planner/relation_restriction_equivalence.c
```diff
@@ -1052,3 +1261,93 @@ AttributeEquivalancesAreEqual(AttributeEquivalenceClass *firstAttributeEquivalan
...
+	{
+		TargetEntry *targetEntry = (TargetEntry *) lfirst(targetEntryCell);
+		Expr *targetExpression = targetEntry->expr;
+
+		partitionKeyTargetAttrIndex++;
+
+		if (!targetEntry->resjunk &&
+			IsPartitionColumn(targetExpression, relationPlannerParseQuery))
+		{
+			Var *targetColumn = (Var *) targetExpression;
```
mtuncer: Are you sure it is a Var here ?

src/include/distributed/multi_planner_utils.h
```diff
@@ -0,0 +1,21 @@
...
+ *-------------------------------------------------------------------------
+ */
+
+#ifndef MULTI_PLANNER_UTILS_H
+#define MULTI_PLANNER_UTILS_H
+
+#include "distributed/multi_planner.h"
+
+
+extern bool SafeToPushDownSubquery(PlannerRestrictionContext *plannerRestrictionContext);
```
mtuncer: This is only used inside multi_router_planner.c , declared in multi_planner_utils.h, and defined in relation_restriction_equivalance.c

It looks weird.

src/include/distributed/relation_restriction_equivalence.h
```diff
@@ -15,8 +15,12 @@
 #include "distributed/multi_planner.h"
 
 
+extern bool SafeToPushDownSubquery(PlannerRestrictionContext *plannerRestrictionContext,
```
mtuncer: no need for additional declaration. it is inside  a header file now

mtuncer:
you should check when a partition column is repeated inside target list
```sql
 select count(*) from (
        select user_id, user_id from users_table 
    union 
        select user_id, 2 from users_table) u;
```
works, but 
```sql
 select count(*) from (
        select user_id, user_id from users_table
    union 
        select 2, user_id  from users_table) u;
```
does not

onderkalaci:
> you should check when a partition column is repeated inside target list

I'm happy to leave this as a known issue 😄  That would a very rare case, wouldn't it? Also, the solution is very simple, re-order you target lists such that partition match is at the first index?

onderkalaci:
I'm happy to note this in the `SafeToPushdownUnionSubquery` function comment?

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -2832,9 +2837,26 @@ ErrorIfContainsUnsupportedSubquery(MultiNode *logicalPlanNode,
...
+			ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+							errmsg("cannot pushdown the subquery since all leaves of "
+								   "the UNION does not include partition key at the "
+								   "same position"),
+							errdetail("Each leaf query of the UNION should return "
+									  "partition key at the same position on its "
+									  "target list.")));
+		}
+	}
+	else if (!RestrictionEquivalenceForPartitionKeys(plannerRestrictionContext))
```
mtuncer: Does this mean we do not need to check`RestrictionEquivalenceForPartitionKeys` if the query contains top level union ?

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -310,7 +312,7 @@ CreateInsertSelectRouterPlan(Query *originalQuery,
 		modifyTask = RouterModifyTaskForShardInterval(originalQuery, targetShardInterval,
 													  relationRestrictionContext,
 													  taskIdIndex,
-													  restrictionEquivalenceForPartitionKeys);
+													  safeToPushDownSubquery);
```
mtuncer: we are passing  variable `safeToPushDownSubquery ` in place of `allRelationsJoinedOnPartitionKey` perhaps we should change function api here

src/backend/distributed/planner/relation_restriction_equivalence.c
```diff
@@ -1052,3 +1250,104 @@ AttributeEquivalancesAreEqual(AttributeEquivalenceClass *firstAttributeEquivalan
...
+		return false;
+	}
+
+	subqueryTree = rangeTableEntry->subquery;
+	setOperations = subqueryTree->setOperations;
+	if (setOperations != NULL)
+	{
+		SetOperationStmt *setOperationStatement = (SetOperationStmt *) setOperations;
+
+		if (setOperationStatement->op != SETOP_UNION)
```
mtuncer: may be a single line comment here stating set operation type in whole tree is checked elsewhere

src/test/regress/expected/multi_subquery_toplevel_union.out
```diff
@@ -0,0 +1,777 @@
+--
```
mtuncer: since we are steering away from 'toplevel_union' we can use multi_subquery_union

src/backend/distributed/planner/multi_logical_optimizer.c
onderkalaci: Yes, that's true, we don't need to check that.

src/backend/distributed/planner/multi_router_planner.c
onderkalaci: Good catch, I'll update it

src/backend/distributed/planner/relation_restriction_equivalence.c
onderkalaci: sure, that makes sense/

src/test/regress/expected/multi_subquery_toplevel_union.out
onderkalaci: I'll update it.

PR #1302, opened 2017-03-31
 Support expressions in the partition column in INSERTs

### LINKED ISSUE #687: Support evaluating functions on master to determine partition column value
**Use case:** I have a distributed `customers` table with a `customer_id` column as SERIAL/UUID. `customer_id` is also my partition column.

My intention is for Citus to generate the `customer_id` (using a sequence or `uuid_generate_v4()`) and return it back to me using `RETURNING` on insertion.

Right now we're failing with `ERROR:  values given for the partition column must be constants or constant expressions`.

The current workaround is to have the client generate the ID, which works easy enough for UUIDs, but is rather cumbersome for sequences/SERIAL.

Full example for SERIAL:

``` sql
CREATE TABLE example_serial(id serial, name text, created_at timestamp);
SELECT master_create_distributed_table('example_serial', 'id', 'hash');
SELECT master_create_worker_shards('example_serial', 16, 1);

INSERT INTO example_serial (name, created_at) VALUES ('My test', now());
-- ERROR:  values given for the partition column must be constants or constant expressions

INSERT INTO example_serial (id, name, created_at) VALUES (nextval('example_serial_id_seq'::regclass), 'My test', now());
-- ERROR:  values given for the partition column must be constants or constant expressions

INSERT INTO example_serial (id, name, created_at) VALUES (1, 'My test', now());
-- INSERT 0 1
```

Full example for uuid:

``` sql
CREATE TABLE example_uuid(id uuid default uuid_generate_v4(), name text, created_at timestamp);
SELECT master_create_distributed_table('example_uuid', 'id', 'hash');
SELECT master_create_worker_shards('example_uuid', 16, 1);

INSERT INTO example_uuid (name, created_at) VALUES ('My test', now());
-- ERROR:  values given for the partition column must be constants or constant expressions

INSERT INTO example_uuid (id, name, created_at) VALUES (uuid_generate_v4(), 'My test', now());
-- ERROR:  values given for the partition column must be constants or constant expressions

INSERT INTO example_uuid (id, name, created_at) VALUES ('fa5f4b4b-6c97-47ee-9f84-fa702e01c98c', 'My test', now());
-- INSERT 0 1
```

marcocitus:
This PR defers shard pruning to the executor for INSERTs that contain an expression in the partition column until after function evaluation. In that case, `TargetShardIntervalForModify` now returns `NULL`. 
For simplicity, the planner still generates a task for the insert in the planner, but leaves assigning the shard ID and shard placements to the executor. 

A more general approach would be to generate the list of tasks after function evaluation, which would allow us to defer shard pruning for other commands as well, but that would have required more code changes and is not possible for all commands nor are there many use-cases for it, so it didn't seem worthwhile.

All the shard pruning logic for INSERTs has been encapsulated in `FindShardForInsert`, such that it can be called both from `TargetShardIntervalForModify` and from the executor.

We no longer generate an error if the partition column contains a non-Const, but still generate an error when it contains a parameter, since that will cause our planner to automagically pick a plan that has no parameters.

Fixes #687

### DIFF
--- a/src/backend/distributed/executor/multi_router_executor.c
+++ b/src/backend/distributed/executor/multi_router_executor.c
@@ -74,6 +74,7 @@ bool EnableDeadlockPrevention = true;
 
 /* functions needed during run phase */
 static void ReacquireMetadataLocks(List *taskList);
+static void AssignInsertTaskShardId(Query *jobQuery, List *taskList);
 static void ExecuteSingleModifyTask(CitusScanState *scanState, Task *task,
 									bool expectResults);
 static void ExecuteSingleSelectTask(CitusScanState *scanState, Task *task);
@@ -87,7 +88,6 @@ static List * TaskShardIntervalList(List *taskList);
 static void AcquireExecutorShardLock(Task *task, CmdType commandType);
 static void AcquireExecutorMultiShardLocks(List *taskList);
 static bool RequiresConsistentSnapshot(Task *task);
-static void ProcessMasterEvaluableFunctions(Job *workerJob, PlanState *planState);
 static void ExtractParametersFromParamListInfo(ParamListInfo paramListInfo,
 											   Oid **parameterTypes,
 											   const char ***parameterValues);
@@ -406,7 +406,11 @@ RequiresConsistentSnapshot(Task *task)
 
 
 /*
- * CitusModifyBeginScan checks the validity of the given custom scan node and
+ * CitusModifyBeginScan first evaluates expressions in the query and then
+ * performs shard pruning in case the partition column in an insert was
+ * defined as a function call.
+ *
+ * The function also checks the validity of the given custom scan node and
  * gets locks on the shards involved in the task list of the distributed plan.
  */
 void
@@ -415,7 +419,23 @@ CitusModifyBeginScan(CustomScanState *node, EState *estate, int eflags)
 	CitusScanState *scanState = (CitusScanState *) node;
 	MultiPlan *multiPlan = scanState->multiPlan;
 	Job *workerJob = multiPlan->workerJob;
+	Query *jobQuery = workerJob->jobQuery;
 	List *taskList = workerJob->taskList;
+	bool deferredPruning = workerJob->deferredPruning;
+
+	if (workerJob->requiresMasterEvaluation)
+	{
+		PlanState *planState = &(scanState->customScanState.ss.ps);
+
+		ExecuteMasterEvaluableFunctions(jobQuery, planState);
+
+		if (deferredPruning)
+		{
+			AssignInsertTaskShardId(jobQuery, taskList);
+		}
+
+		RebuildQueryStrings(jobQuery, taskList);
+	}
 
 	/*
 	 * If we are executing a prepared statement, then we may not yet have obtained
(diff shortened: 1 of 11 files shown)


ozgune:
@marcocitus -- This is exciting! I had two curiosity questions.

If you think of the higher level design options, do you view them as these two mentioned in your PR?

1. Defer ``INSERT`` statement shard pruning to the executor stage, and
2. Defer ``INSERT / UPDATE / DELETE`` statements to the executor stage

I also had a question for the future. If we support multi-row INSERTs or multi-shard DELETEs in upcoming releases, would our higher level approach (deferring shard pruning to the executor) work or could be made to cleanly work?

marcocitus:
> 1. Defer INSERT statement shard pruning to the executor stage, and
> 2. Defer INSERT / UPDATE / DELETE statements to the executor stage

Yes. Given that 1 is pretty simple it'll be easy to support commands other than single-row INSERT when the need arises and we understand the requirements.

For update/delete, deferring pruning seems irrelevant for now, since it would only apply to queries like `UPDATE hash_distributed_table SET x = 1 WHERE partition_column = now()`, which seem unlikely to occur. In the future, we may want to support deferred pruning for multi-shard update/delete on range-partitioned tables (`DELETE FROM range_distributed_table WHERE created_at <= now() - interval '1 week'`).

It is much harder to check whether we a) can defer shard pruning and b) need to defer shard pruning for an update/delete. Given that it's both hard and irrelevant, I didn't want to go there :).

> If we support multi-row INSERTs or multi-shard DELETEs in upcoming releases, would our higher level approach (deferring shard pruning to the executor) work or could be made to cleanly work?

To implement multi-row inserts, we can either reuse the COPY logic or write a function that can break down a (multi-row) insert into a list of tasks. Once such a function exists, it'll be trivial to adjust the deferred shard pruning to (re)generate the task list.

sumedhpathak:
@lithp, just checking-in, are you reviewing these changes? We decided to prioritize these for the 6.2 cycle. Thanks!

lithp:
@sumedhpathak yep, I have a pending review I'm working on. I'll push it soon!

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -428,6 +449,91 @@ CitusModifyBeginScan(CustomScanState *node, EState *estate, int eflags)
...
+
+	/*
+	 * If we deferred shard pruning to the executor, then we still need to assign
+	 * shard placements. We do this after acquiring the metadata locks to ensure
+	 * we can't get stale metadata. At some point, we may want to load all
+	 * placement metadata here.
+	 */
+	if (deferredPruning)
+	{
+		AssignTaskPlacements(workerJob);
```
lithp: Can't this just be: `workerJob->taskList = FirstReplicaAssignTaskList(workerJob->taskList)`?

src/test/regress/sql/multi_modifications.sql
```diff
@@ -418,3 +418,15 @@ SELECT master_create_worker_shards('app_analytics_events', 4, 1);
 INSERT INTO app_analytics_events VALUES (DEFAULT, 101, 'Fauxkemon Geaux') RETURNING id;
 INSERT INTO app_analytics_events (app_id, name) VALUES (102, 'Wayz') RETURNING id;
 INSERT INTO app_analytics_events (app_id, name) VALUES (103, 'Mynt') RETURNING *;
+
+DROP TABLE app_analytics_events;
+
+-- again with serial in the partition column
+CREATE TABLE app_analytics_events (id serial, app_id integer, name text);
+SELECT create_distributed_table('app_analytics_events', 'id');
```
lithp: Maybe add a test which includes an expression like `2+2`.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1519,7 +1521,7 @@ ModifyQuerySupported(Query *queryTree)
 			}
 
 			if (commandType == CMD_INSERT && targetEntryPartitionColumn &&
-				!IsA(targetEntry->expr, Const))
+				ExpressionContainsParam((Node *) targetEntry->expr))
```
marcocitus: Not quite, `EvaluateNodeIfReferencesFunction` doesn't always evaluate parameters. You're right that this could be supported now, and it could eventually replace our prepared statement hack, but I'd rather do that in a separate PR.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -2066,41 +2106,162 @@ TargetShardIntervalForModify(Query *query)
...
+
+		restrictClauseList = list_make1(equalityExpr);
+
+		shardIntervalList = LoadShardIntervalList(distributedTableId);
 		prunedShardList = PruneShardList(distributedTableId, tableId, restrictClauseList,
 										 shardIntervalList);
 	}
 
 	prunedShardCount = list_length(prunedShardList);
 	if (prunedShardCount != 1)
```
marcocitus: I was thinking it's actually easier to understand without the mixing and matching of messages depending on the command type. I do feel like this function is a little strangely structured now that the insert logic is delegated to `FindShardForInsert`. I'm introducing a `FindShardForUpdateOrDelete` to make it more symmetric.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -2231,7 +2374,6 @@ static List *
 QueryRestrictList(Query *query)
 {
 	List *queryRestrictList = NIL;
-	CmdType commandType = query->commandType;
 	Oid distributedTableId = ExtractFirstDistributedTableId(query);
```
marcocitus: Fixed.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -2066,41 +2106,162 @@ TargetShardIntervalForModify(Query *query)
...
+/*
+ * FindShardForInsert finds the shard interval for an INSERT query or
+ * NULL if the partition column value is defined as an expression that
+ * still needs to be evaluated.
+ */
+ShardInterval *
+FindShardForInsert(Query *query)
+{
+	Oid distributedTableId = ExtractFirstDistributedTableId(query);
+	DistTableCacheEntry *cacheEntry = DistributedTableCacheEntry(distributedTableId);
```
marcocitus: Since this is a public function that is also called from the executor, it seems slightly nicer to avoid putting the burden on the caller here.

marcocitus: I removed some other redundant `ExtractFirstDistributedTableId` calls.

lithp:
Okay, did another pass and ran some more tests and this looks good! :+1: Though the rebase looks like it'll require a small change in the tests

PR #1306, opened 2017-04-04
Disallow subqueries without a relation in the range table list

onderkalaci:
This commit disallows subqueries without relation in the range table
list. This commit is only applied for subquery pushdown and INSERT..
SELECT queries. In other words, we do not add this limitation for single
table re-partition subqueries.

The reasoning behind this limitation is that if we allow pushing down
such queries, the result would include (shardCount * expectedResults)
where in a non distributed world the result would be (expectedResult)
only.

### DIFF
--- a/src/backend/distributed/planner/multi_router_planner.c
+++ b/src/backend/distributed/planner/multi_router_planner.c
@@ -871,6 +871,15 @@ MultiTaskRouterSelectQuerySupported(Query *query)
 
 		Assert(subquery->commandType == CMD_SELECT);
 
+		/* pushing down rtes without relations yields (shardCount * expectedRows) */
+		if (subquery->rtable == NIL)
+		{
+			return DeferredError(ERRCODE_FEATURE_NOT_SUPPORTED,
+								 "Subqueries without relations are not allowed in "
+								 "INSERT ... SELECT queries",
+								 NULL, NULL);
+		}
+
 		/* pushing down limit per shard would yield wrong results */
 		if (subquery->limitCount != NULL)
 		{
--- a/src/test/regress/expected/multi_insert_select_non_pushable_queries.out
+++ b/src/test/regress/expected/multi_insert_select_non_pushable_queries.out
@@ -660,3 +660,30 @@ FROM
   WHERE users_table.value_1 < 50;
 ERROR:  cannot perform distributed planning for the given modification
 DETAIL:  Select query cannot be pushed down to the worker.
+-- not supported since one of the queries doesn't have a relation
+INSERT INTO agg_results (user_id, agg_time, value_2_agg)
+SELECT
+    user_id,
+    user_lastseen,
+    array_length(event_array, 1)
+FROM (
+    SELECT
+        user_id,
+        max(u.time) as user_lastseen,
+        array_agg(event_type ORDER BY u.time) AS event_array
+    FROM (
+        SELECT user_id, time, value_3 as val_3
+        FROM users_table
+        WHERE
+        user_id >= 10 AND user_id <= 70 AND
+        users_table.value_1 > 10 AND users_table.value_1 < 12
+        ) u LEFT JOIN LATERAL (
+          SELECT event_type, time
+          FROM events_table, (SELECT 1 as x) as f
+          WHERE user_id = u.user_id AND
+          events_table.event_type > 10 AND events_table.event_type < 12
+        ) t ON true
+        GROUP BY user_id
+) AS shard_union
+ORDER BY user_lastseen DESC;
+ERROR:  Subqueries without relations are not allowed in INSERT ... SELECT queries
--- a/src/test/regress/sql/multi_insert_select_non_pushable_queries.sql
+++ b/src/test/regress/sql/multi_insert_select_non_pushable_queries.sql
@@ -649,3 +649,30 @@ FROM
     ) temp 
   ON users_table.user_id = temp.user_id 
   WHERE users_table.value_1 < 50;
+
+-- not supported since one of the queries doesn't have a relation
+INSERT INTO agg_results (user_id, agg_time, value_2_agg)
+SELECT
+    user_id,
+    user_lastseen,
+    array_length(event_array, 1)
+FROM (
+    SELECT
+        user_id,
+        max(u.time) as user_lastseen,
+        array_agg(event_type ORDER BY u.time) AS event_array
+    FROM (
+        SELECT user_id, time, value_3 as val_3
+        FROM users_table
+        WHERE
+        user_id >= 10 AND user_id <= 70 AND
+        users_table.value_1 > 10 AND users_table.value_1 < 12
+        ) u LEFT JOIN LATERAL (
+          SELECT event_type, time
+          FROM events_table, (SELECT 1 as x) as f
+          WHERE user_id = u.user_id AND
+          events_table.event_type > 10 AND events_table.event_type < 12
+        ) t ON true
+        GROUP BY user_id
+) AS shard_union
+ORDER BY user_lastseen DESC;
--- a/src/test/regress/sql/multi_subquery_behavioral_analytics.sql
+++ b/src/test/regress/sql/multi_subquery_behavioral_analytics.sql
@@ -1285,6 +1285,66 @@ SELECT * FROM run_command_on_workers('DROP FUNCTION array_index(ANYARRAY, ANYELE
 ORDER BY 1,2;
 DROP FUNCTION array_index(ANYARRAY, ANYELEMENT);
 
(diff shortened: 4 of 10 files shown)


src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -838,6 +838,15 @@ MultiTaskRouterSelectQuerySupported(Query *query)
 
 		Assert(subquery->commandType == CMD_SELECT);
 
+		/* pushing down rtes without relations yields (shardCount * expectedRows) */
+		if (subquery->rtable == NIL)
+		{
+			return DeferredError(ERRCODE_FEATURE_NOT_SUPPORTED,
```
metdos: Is this a fix for INSERT ... SELECT queries? If yes, shall we do that separately?

src/test/regress/expected/multi_insert_select_non_pushable_queries.out
```diff
@@ -660,3 +660,32 @@ FROM
...
+    user_id,
+    user_lastseen,
+    array_length(event_array, 1)
+FROM (
+    SELECT
+        user_id,
+        max(u.time) as user_lastseen,
+        array_agg(event_type ORDER BY u.time) AS event_array
+    FROM (
+
```
metdos: Empty line. Also, we can fix query format a bit.

metdos:
Adding back disappeared comment:

This looks good. I have two comments;

- We also change the point where we error out for INSERT ... SELECT queries. As long as I tested, we already detect this problem at another place and error out with a different message. Do we want to make this change here? (I prefer this approach to old one)
- Could you go over regression tests, and fix some indentation issues?

src/backend/distributed/planner/multi_router_planner.c
onderkalaci: I moved this to a separate commit. Does that work?

src/test/regress/expected/multi_insert_select_non_pushable_queries.out
```diff
@@ -660,3 +660,31 @@ FROM
...
+FROM (
+    SELECT
+        user_id,
+        max(u.time) as user_lastseen,
+        array_agg(event_type ORDER BY u.time) AS event_array
+    FROM (
+        SELECT user_id, time, value_3 as val_3
+        FROM users_table
+        WHERE
+        user_id >= 10 AND
```
metdos: why user_id's are on different lines but value_1's are on same line :)

src/test/regress/sql/multi_insert_select_non_pushable_queries.sql
```diff
@@ -649,3 +649,31 @@ FROM
...
+FROM (
+    SELECT
+        user_id,
+        max(u.time) as user_lastseen,
+        array_agg(event_type ORDER BY u.time) AS event_array
+    FROM (
+        SELECT user_id, time, value_3 as val_3
+        FROM users_table
+        WHERE
+        user_id >= 10 AND
```
metdos: why user_id's are on different lines but value_1's are on same line :)

src/test/regress/sql/multi_subquery_behavioral_analytics.sql
```diff
@@ -1285,6 +1285,59 @@ SELECT * FROM run_command_on_workers('DROP FUNCTION array_index(ANYARRAY, ANYELE
 ORDER BY 1,2;
 DROP FUNCTION array_index(ANYARRAY, ANYELEMENT);
 
+-- a not supported query due to constant range table entry
+SELECT count(*) as subquery_count
+FROM (
+  SELECT 
+      user_id
+    FROM
+    users_table
```
metdos: I couldn't see any pattern on above 3 lines.

src/test/regress/sql/multi_subquery_behavioral_analytics.sql
```diff
@@ -1285,6 +1285,59 @@ SELECT * FROM run_command_on_workers('DROP FUNCTION array_index(ANYARRAY, ANYELE
...
+  WHERE
+    (value_1 = '5' OR value_1 = '13')
+  GROUP BY user_id
```
metdos: we have user_id on the same line with GROUP BY, but for SELECT we use the next line.

src/test/regress/sql/multi_subquery_behavioral_analytics.sql
```diff
@@ -1285,6 +1285,59 @@ SELECT * FROM run_command_on_workers('DROP FUNCTION array_index(ANYARRAY, ANYELE
...
+  SELECT
+    1 as user_id
+  ) AS b 
+  ON a.user_id = b.user_id 
+WHERE b.user_id IS NULL
+GROUP BY a.user_id;
+
+-- this is slightly different, we use RTE_VALUEs here
+SELECT Count(*) AS subquery_count 
+FROM   (SELECT user_id
```
metdos: More than one space.

src/test/regress/sql/multi_subquery_behavioral_analytics.sql
```diff
@@ -1285,6 +1285,59 @@ SELECT * FROM run_command_on_workers('DROP FUNCTION array_index(ANYARRAY, ANYELE
...
+  ON a.user_id = b.user_id 
+WHERE b.user_id IS NULL
+GROUP BY a.user_id;
+
+-- this is slightly different, we use RTE_VALUEs here
+SELECT Count(*) AS subquery_count 
+FROM   (SELECT user_id 
+        FROM   users_table 
+        WHERE  ( value_1 = '5' 
+                  OR value_1 = '13' )
```
metdos: Looking to pattern here ` OR value_1 = '13' ) ` should continue on the line above.

src/test/regress/sql/multi_subquery_complex_queries.sql
```diff
@@ -1786,5 +1786,91 @@ FROM     (
...
+  ( SELECT *, random()
+   FROM
+     ( SELECT "t"."user_id",
+              "t"."time",
+              unnest("t"."collected_events") AS "event_types"
+      FROM
+        ( SELECT "t1"."user_id",
+                 min("t1"."time") AS "time",
+                 array_agg(("t1"."event")
+                           ORDER BY TIME ASC, event DESC) AS collected_events
```
metdos: ORDER BY TIME indented too much.

src/test/regress/sql/multi_subquery_behavioral_analytics.sql
onderkalaci: I don't think we have a convention for this in our tests?

src/test/regress/sql/multi_subquery_complex_queries.sql
onderkalaci: But that `ORDER BY` belongs to `array_agg`, so should be fine?

src/test/regress/sql/multi_subquery_behavioral_analytics.sql
metdos: I think we don't have a formal one, but having symmetry and consistent spacing makes it easier to read and understand queries. It doesn't need to be strict as much as our coding conventions.

src/test/regress/sql/multi_subquery_complex_queries.sql
metdos: Ohh, maybe we should put them in one line :)

src/test/regress/sql/multi_subquery_behavioral_analytics.sql
onderkalaci: Should be fixed now?

src/test/regress/sql/multi_subquery_behavioral_analytics.sql
```diff
@@ -1285,6 +1285,66 @@ SELECT * FROM run_command_on_workers('DROP FUNCTION array_index(ANYARRAY, ANYELE
...
+GROUP BY a.user_id;
+
+-- same with INNER JOIN
+SELECT count(*) as subquery_count
+FROM (
+  SELECT 
+      user_id
+    FROM
+    users_table
+  WHERE
```
metdos: Unaligned lines.

src/test/regress/sql/multi_subquery_union.sql
```diff
@@ -605,5 +605,73 @@ FROM
...
+  (SELECT (random() * 100)::int)
+) b;
+
+-- we don't support subqueries without relations
+SELECT 
+  user_id, value_3
+FROM 
+(
+  (SELECT value_3, user_id FROM events_table where event_type IN (1, 2, 3, 4, 5))
+    UNION ALL
```
metdos: first union indented but not others

PR #1309, opened 2017-04-06
Error out on parametrized SQL functions

### LINKED ISSUE #981: SQL functions(called by SELECT * FROM) with parameters throw wrong results
```
CREATE TABLE lineitem(id integer, org_id integer);
select master_create_distributed_table('lineitem','org_id','hash');
select master_create_worker_shards('lineitem',16,1);

CREATE OR REPLACE FUNCTION count_rows(org_id_val integer)
RETURNS TABLE (a bigint)
AS $$
  SELECT count(*) AS count_val from lineitem where org_id=org_id_val;
$$ LANGUAGE SQL STABLE;
```
The below test throws wrong results. The query doesn't run on the distributed table.
```
select * from count_rows(1);
 a

---
 0
```

byucesoy:
Fixes #981 

Before this commit, we were erroring out for queries containing parameterized SQL functions
like 'SELECT parameterized_sql_query(value)' as we should, however, we were returning wrong
results for queries like 'SELECT * FROM parameterized_sql_query(value)'. With this commit
we started to error out on such queries too.

### DIFF
--- a/src/backend/distributed/planner/multi_planner.c
+++ b/src/backend/distributed/planner/multi_planner.c
@@ -126,6 +126,22 @@ multi_planner(Query *parse, int cursorOptions, ParamListInfo boundParams)
 	/* remove the context from the context list */
 	PopPlannerRestrictionContext();
 
+	/*
+	 * In some cases, for example; parameterized SQL functions, we may miss that
+	 * there is a need for distributed planning. Such cases only become clear after
+	 * standart_planner performs some modifications on parse tree. In such cases
+	 * we will simply error out.
+	 */
+	if (!needsDistributedPlanning && NeedsDistributedPlanning(parse))
+	{
+		ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+						errmsg("cannot perform distributed planning on this "
+							   "query because parameterized queries for SQL "
+							   "functions referencing distributed tables are "
+							   "not supported"),
+						errhint("Consider using PL/pgSQL functions instead.")));
+	}
+
 	return result;
 }
 
@@ -334,7 +350,7 @@ CreateDistributedPlan(PlannedStmt *localPlan, Query *originalQuery, Query *query
 						  "could not create distributed plan",
 						  "Possibly this is caused by the use of parameters in SQL "
 						  "functions, which is not supported in Citus.",
-						  "Consider using PLPGSQL functions instead.");
+						  "Consider using PL/pgSQL functions instead.");
 	}
 
 	/*
--- a/src/test/regress/expected/multi_sql_function.out
+++ b/src/test/regress/expected/multi_sql_function.out
@@ -34,15 +34,6 @@ CREATE FUNCTION sql_test_no_4() RETURNS bigint AS '
 		o_custkey = c_custkey AND
 		o_orderkey = l_orderkey;
 ' LANGUAGE SQL;
-CREATE FUNCTION sql_test_no_6(integer) RETURNS bigint AS  $$
-	SELECT
-		count(*)
-	FROM
-		orders, lineitem
-	WHERE
-		o_orderkey = l_orderkey AND
-		l_suppkey > $1;
-$$ LANGUAGE SQL RETURNS NULL ON NULL INPUT;
 SET citus.task_executor_type TO 'task-tracker';
 SET client_min_messages TO INFO;
 -- now, run plain SQL functions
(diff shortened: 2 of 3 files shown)


src/backend/distributed/planner/multi_planner.c
```diff
@@ -127,6 +127,18 @@ multi_planner(Query *parse, int cursorOptions, ParamListInfo boundParams)
 			result = CreateDistributedPlan(result, originalQuery, parse,
 										   boundParams, restrictionContext);
 		}
+		else
+		{
+			needsDistributedPlanning = NeedsDistributedPlanning(parse);
```
metdos: I think we can do this check after PG_TRY()/PG_CATCH() blocks just before returning. 

We can also use just one if block such as;

```C
	if (!needsDistributedPlanning && NeedsDistributedPlanning(parse))
	{
		ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
						errmsg("cannot perform distributed planning on this "
							   "query because parametrized queries for SQL "
							   "functions referencing are not supported."),
						errhint("Consider using PLPGSQL functions instead.")));
	}
```

I think this is nicer because it shows in one line that it was not distributed but now it returns distributed so something is wrong :)

byucesoy: @metdos  fixed

src/backend/distributed/planner/multi_planner.c
```diff
@@ -138,6 +138,21 @@ multi_planner(Query *parse, int cursorOptions, ParamListInfo boundParams)
 	/* remove the context from the context list */
 	PopRestrictionContext();
 
+	/*
+	 * In some cases, for example; parametrized SQL functions, we may miss that
```
metdos: parametrized -> parameterized

src/backend/distributed/planner/multi_planner.c
```diff
@@ -138,6 +138,21 @@ multi_planner(Query *parse, int cursorOptions, ParamListInfo boundParams)
...
+	 * there is a need for distributed planning. Such cases only become clear after
+	 * standart_planner performs some modifications on parse tree. In such cases
+	 * we will simply error out.
+	 */
+	if (!needsDistributedPlanning && NeedsDistributedPlanning(parse))
+	{
+		ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+						errmsg("cannot perform distributed planning on this "
+							   "query because parametrized queries for SQL "
+							   "functions referencing are not supported."),
```
metdos: Maybe: referencing distributed tables are not supported yet

src/backend/distributed/planner/multi_planner.c
```diff
@@ -138,6 +138,21 @@ multi_planner(Query *parse, int cursorOptions, ParamListInfo boundParams)
...
+						errhint("Consider using PLPGSQL functions instead.")));
```
metdos: PLPGSQL -> PL/pgSQL

src/test/regress/expected/multi_sql_function.out
```diff
@@ -319,7 +319,31 @@ SELECT * FROM prepare_table ORDER BY key, value;
    6 |      
 (8 rows)
 
+-- test running parametrized SQL funtion
```
metdos: replace all `parametrized` with `parameterized`

src/test/regress/sql/multi_sql_function.sql
```diff
@@ -140,7 +140,24 @@ SELECT non_partition_parameter_delete_sql(62);
 -- check after deletes
 SELECT * FROM prepare_table ORDER BY key, value;
 
+-- test running parametrized SQL funtion
```
metdos: replace all `parametrized` with `parameterized`

src/backend/distributed/planner/multi_planner.c
```diff
@@ -148,9 +148,10 @@ multi_planner(Query *parse, int cursorOptions, ParamListInfo boundParams)
 	{
 		ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
 						errmsg("cannot perform distributed planning on this "
-							   "query because parametrized queries for SQL "
-							   "functions referencing are not supported."),
-						errhint("Consider using PLPGSQL functions instead.")));
+							   "query because parameterized queries for SQL "
+							   "functions referencing distributed tables are "
+							   "not supported."),
```
metdos: Minor. We don't put a dot at the end of `errmsg`.

Issue #1249, opened 2017-02-21
Add support for renaming Distributed Tables


sumedhpathak:
E.g.

```SQL
ALTER TABLE distributors RENAME TO suppliers;
```

lfittl:
Just to add a quick note that this came up with a current customer, as a maintenance operation they wanted to run.

The motivations for this was/is that their Rails application changed, and they adjusted the model name in the ORM, and would like to have the underlying table name match the new model name.

Altonymous:
This is also very helpful when you need to change the distribution key on a table and want to do an inline replace of the table..

As an example:
`BEGIN TRANSACTION;

CREATE TABLE web_logs_new (
...
);

SELECT create_distributed_table('web_logs_new', 'ip');

ALTER TABLE web_logs RENAME TO web_logs_domain_dist;
ALTER TABLE web_logs_new RENAME TO web_logs;

END TRANSACTION;`

marcocitus:
Fixed by #1939

PR #1312, opened 2017-04-06
Enable distributed ALTER TABLE ... RENAME COLUMN

### LINKED ISSUE #633: Add Support for Renaming Columns for Distributed Tables
Currently, we do not support renaming distributed tables or their objects, which is a requested feature. 

We error out with the following message:

```
postgres=# ALTER TABLE test_keyval RENAME COLUMN val TO value;
ERROR:  renaming distributed tables or their objects is currently unsupported
```

@samay-sharma noted that merging #618 would make it easier to add support for renaming distributed tables and their columns. We already capture RENAME statements so adding support for it should also be pretty straight-forward.

This issue can be tracked under #25 (DDL Propagation v2).

jasonmp85:
Pretty straightforward. Had some concerns about locking, but due to the fact that all distributed operations use either some level of deparsing or need to enumerate column names, they all block during any concurrent column renames (due to the AccessExclusive lock).

In addition, I had some misgivings about permitting renames of the distribution column, but nothing bad comes from just allowing them.

Finally, I tried to trigger any sort of error using prepared statements and could not trigger any errors not also exhibited by plain PostgreSQL tables.

Fixes #633

### DIFF
--- a/src/backend/distributed/executor/multi_utility.c
+++ b/src/backend/distributed/executor/multi_utility.c
@@ -112,6 +112,7 @@ static List * PlanDropIndexStmt(DropStmt *dropIndexStatement,
 								const char *dropIndexCommand);
 static List * PlanAlterTableStmt(AlterTableStmt *alterTableStatement,
 								 const char *alterTableCommand);
+static List * PlanRenameStmt(RenameStmt *renameStmt, const char *renameCommand);
 static Node * WorkerProcessAlterTableStmt(AlterTableStmt *alterTableStatement,
 										  const char *alterTableCommand);
 static List * PlanAlterObjectSchemaStmt(AlterObjectSchemaStmt *alterObjectSchemaStmt,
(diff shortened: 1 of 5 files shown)


onderkalaci:
> By only removing ErrorIfDistributedRenameStmt() call, I managed to rename distributed tables.

Even if we can achieve that with a very small change, I'd be more comfortable if we do it with a separate PR.

jasonmp85:
@byucesoy, I think making that change requires thinking a bit more about concurrent operations, i.e. whether it's always safe to change a distributed relation's name during other operations. This PR solely addresses #633, which was scheduled for this cycle, whereas #1249 (I believe) was not.

Are there any other blockers to merging this, then?

jasonmp85:
(I guess the important question is: can this be a :shipit:, and if not, what do I need to do?)

byucesoy:
I think, there are no other blockers for this to be :shipit:

sumedhpathak:
@jasonmp85 This PR is approved. Anything else needs to be done here?

PR #1314, opened 2017-04-08
[WIP] Add pg_dist_statements view

lfittl:
This adds the necessary backend for a "pg_dist_statements" view to Citus, that one can JOIN against pg_stat_statements based on a queryid, in order to determine which executor a certain query used.

The primary use case for this is multi-tenant customers that want to understand how many cross-shard queries are running on their production system, and what their performance is.

This needs significant cleanup, but I thought I'd post it here for sharing my work in progress.

The output looks like this:

```
citus=> SELECT * FROM pg_dist_statements;
 executor  | userid | dbid  |  queryid   |                   query                    | calls | total_time | min_time | max_time | mean_time | stddev_time | rows 
-----------+--------+-------+------------+--------------------------------------------+-------+------------+----------+----------+-----------+-------------+------
 real-time |  16384 | 16385 |  436114840 | SELECT * FROM test;                        |     1 |      8.669 |    8.669 |    8.669 |     8.669 |           0 |    1
 router    |  16384 | 16385 |  948709032 | INSERT INTO test (id, test) VALUES (?, ?); |     1 |     40.991 |   40.991 |   40.991 |    40.991 |           0 |    1
 router    |  16384 | 16385 | 2615345884 | SELECT * FROM test wHERE id = ?;           |     2 |      2.946 |    0.677 |    2.269 |     1.473 |       0.796 |    2
(3 rows)
```

Remaining TODO I'm aware of:

* [ ] Figure out naming / unify things (pg_dist_statement vs StatStatementExecutors, etc)
* [ ] Add regression tests
* [ ] Review the code for unnecessary logic, a good chunk of this has been copied from pg_stat_statements
* [ ] Correctly handle the case where pg_stat_statements is not loaded (set max size to 0 and do nothing)
* [ ] Figure out where to place the CREATE VIEW piece that depends on pg_stat_statements (this could potentially be left to the user / created by default on Citus Cloud)
* [ ] **Nice to have:** Keep a connection counter (i.e. number of newly started connections) inside Citus for each query execution, and track that in pg_dist_statements as well - that way we can determine more easily how problematic a certain query is, and whether there are connection leaks

In progress reviews and thoughts welcome of course :)

### DIFF
--- a/src/backend/distributed/Makefile
+++ b/src/backend/distributed/Makefile
@@ -10,7 +10,7 @@ EXTVERSIONS = 5.0 5.0-1 5.0-2  \
 	5.2-1 5.2-2 5.2-3 5.2-4 \
 	6.0-1 6.0-2 6.0-3 6.0-4 6.0-5 6.0-6 6.0-7 6.0-8 6.0-9 6.0-10 6.0-11 6.0-12 6.0-13 6.0-14 6.0-15 6.0-16 6.0-17 6.0-18 \
 	6.1-1 6.1-2 6.1-3 6.1-4 6.1-5 6.1-6 6.1-7 6.1-8 6.1-9 6.1-10 6.1-11 6.1-12 6.1-13 6.1-14 6.1-15 6.1-16 6.1-17 \
-	6.2-1 6.2-2
+	6.2-1 6.2-2 6.2-3
 
 # All citus--*.sql files in the source directory
 DATA = $(patsubst $(citus_abs_srcdir)/%.sql,%.sql,$(wildcard $(citus_abs_srcdir)/$(EXTENSION)--*--*.sql))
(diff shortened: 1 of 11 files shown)


onderkalaci:
Just adding a note for reference: Although requires some more work, here is the [prototype](https://github.com/citusdata/citus-enterprise/tree/pg_stat_dist_improved) for adding the tenant ids to the `pg_dist_statements` view.

lfittl:
Quick note - when we pick this up we'll also need to take care of the change in Postgres 11 that makes queryid 64-bit instead of 32-bit: https://www.postgresql.org/message-id/flat/CA%2BTgmobG_Kp4cBKFmsznUAaM1GWW6hhRNiZC0KjRMOOeYnz5Yw%40mail.gmail.com#CA+TgmobG_Kp4cBKFmsznUAaM1GWW6hhRNiZC0KjRMOOeYnz5Yw@mail.gmail.com

marcocitus:
Replaced by #2235

PR #1315, opened 2017-04-10
Change the behaviour of subquery pushdown flag

onderkalaci:
This commit changes the behaviour of the citus.subquery_pushdown flag.
Before this commit, the flag is used to enable subquery pushdown logic. But,
with this commit, that behaviour is enabled by default. In other words, the
flag is now useless. We prefer to keep the flag since we don't want to break
the backward compatibility. Also, we may consider using that flag for other
purposes in the next commits.


metdos:
Having two commits, one for moving functions and one for updating them, would make code reviewer's (me) job easier :)

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -158,6 +181,795 @@ MultiLogicalPlanCreate(Query *originalQuery, Query *queryTree)
...
+	bool outerQueryHasLimit = false;
+	DeferredErrorMessage *error = NULL;
+	bool safeToPushDownSubquery = false;
+
+	if (originalQuery->limitCount != NULL)
+	{
+		outerQueryHasLimit = true;
+	}
+
+	safeToPushDownSubquery = SafeToPushDownSubquery(plannerRestrictionContext);
```
metdos: We don't use a deferred message here because SafeToPushDownSubquery() returns false only for joins, right? At least, can we add a comment for that?

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -158,6 +181,795 @@ MultiLogicalPlanCreate(Query *originalQuery, Query *queryTree)
...
+			TargetListOnPartitionColumn(subqueryTree, distinctTargetEntryList);
+		if (!distinctOnPartitionColumn)
+		{
+			preconditionsSatisfied = false;
+			errorDetail = "Distinct on columns without partition column is "
+						  "currently unsupported";
+		}
+	}
+
+	/* finally check and error out if not satisfied */
```
metdos: /* finally check and return deferred error if not satisfied */

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -158,6 +181,795 @@ MultiLogicalPlanCreate(Query *originalQuery, Query *queryTree)
...
+
+
+/*
+ * ErrorIfUnsupportedUnionQuery is a helper function for ErrorIfCannotPushdownSubquery().
+ * It basically iterates over the subqueries that reside under the given set operations.
+ *
+ * The function also errors out for set operations INTERSECT and EXCEPT.
+ */
+static DeferredErrorMessage *
+ErrorIfUnsupportedUnionQuery(List *rangeTableList,
```
metdos: Shall we also rename this function?

metdos:
@onderkalaci, I provided some feedback, but I will do other passes after we rebase this and solve merge conflicts.

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -158,6 +181,795 @@ MultiLogicalPlanCreate(Query *originalQuery, Query *queryTree)
...
+ * In order to do create the logical plan, we follow the algorithm below:
+ *    -  If subquery pushdown planner can plan the query
+ *        -  We're done, we create the multi plan tree and return
+ *    -  Else
+ *       - If the query is not eligible for single table repartition subquery planning
+ *            - Throw the error that the subquery pushdown planner generated
+ *       - If it is eligible for single table repartition subquery planning
+ *            - Check for the errors for single table repartition subquery planning
+ *                - If no errors found, we're done. Create the multi plan and return
+ *                - If found errors, throw it
```
metdos: I will prefer erroring out with  `subquery pushdown` message because we expect most of our users to be real-time use case users?

src/backend/distributed/planner/multi_logical_planner.c
metdos: Thinking about this a bit more, we can also use a deferred error here. Do we have any strong reason not to use deferred error here?

onderkalaci: I've not made this to defer error since we're using the same function for `Insert ... select` and there we currently don't error out as we do here. I think we could do it in a separate PR which improves `insert ..select` error messages. What do you think?

metdos: Some options;
i. Add a comment and explain behavior
ii. Wrap this in another function which returns a deferred error
iii. Return a deferred error within if block.

I slightly prefer iii->ii->i. I'm okay with i.

However, I think the name of `SafeToPushDownSubquery` is misleading (not in this PR but overall subqueries).
i. It does not completely check subquery push down.
ii. It does two things:
a. Partition column check for joins.
b. Union checks

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -641,13 +1486,27 @@ ErrorIfSubqueryNotSupported(Query *subqueryTree)
...
-		ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
-						errmsg("cannot perform distributed planning on this query"),
-						errdetail("%s", errorDetail)));
+		return DeferredError(ERRCODE_FEATURE_NOT_SUPPORTED,
+							 "cannot perform distributed planning on this query",
+							 errorDetail, NULL);
 	}
+
+	/* continue with the inner subquery */
+	rangeTableEntry = linitial(subqueryTree->rtable);
```
onderkalaci: We should use the first rte index in `joinTreeTableIndexList `, not `linitial`

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -158,6 +180,810 @@ MultiLogicalPlanCreate(Query *originalQuery, Query *queryTree)
...
+ */
+static bool
+SingleRelationRepartitionSubquery(Query *queryTree)
+{
+	List *rangeTableList = queryTree->rtable;
+	RangeTblEntry *rangeTableEntry = NULL;
+	Query *subqueryTree = NULL;
+
+	/* don't allow joines on top of unions */
+	if (list_length(rangeTableList) > 1)
```
onderkalaci: We should use the index walker instead of `list_length`

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: Well, that would mean that we remove all the single repartition error messages. I think keeping it as it is might be better. Otherwise, someone who is trying to run single repartition subquery would always see subquery pushdown error messages.

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -158,6 +180,810 @@ MultiLogicalPlanCreate(Query *originalQuery, Query *queryTree)
...
+		outerQueryHasLimit = true;
+	}
+
+	/*
+	 * We're checking two things here:
+	 *    (i)   If the query contains a top level union, ensure that all leaves
+	 *          return the partition key at the same position
+	 *    (ii)  Else, check whether all relations joined on the partition key or not
+	 */
+	if (ContainsUnionSubquery(originalQuery))
```
onderkalaci: Adding @metdos old comment: Why don't we defer this errors as well?

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -158,6 +181,795 @@ MultiLogicalPlanCreate(Query *originalQuery, Query *queryTree)
...
+
+
+/*
+ * ContainsUnsupportedSubquery iterates on the query's subquery entry list
+ * and uses helper functions to check if we can push down subquery
+ * to worker nodes. These helper functions returns a deferred error out
+ * if we cannot push down the subquery.
+ */
+static DeferredErrorMessage *
+ContainsUnsupportedSubquery(Query *originalQuery,
```
onderkalaci: We don't have `SafeToPushDownSubquery ()` anymore, we've renamed it to `RestrictionEquivalenceForPartitionKeys()` in the previous PRs.

`ContainsUnsupportedSubquery()` already existed in the main branch, I wanted to be consistent with it. I'll now update it to `DeferErrorIfContainsUnsupportedSubquery()` to make it consistent with the other error functions. Later, we can discuss to rename all together to prefix with `Check..` or so.

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: > Which error message? Subquery pushdown or single table repartition?
  
The sentence starts with `Check for the errors for single table repartition subquery planning`. I think using it again may look unnecessary?

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -56,11 +60,29 @@ typedef MultiNode *(*RuleApplyFunction) (MultiNode *leftNode, MultiNode *rightNo
...
+																	  originalQuery,
+																	  PlannerRestrictionContext
+																	  *
+																	  plannerRestrictionContext);
+static DeferredErrorMessage * DeferErrorIfUnsupportedFilters(Query *subquery);
+static bool EqualOpExpressionLists(List *firstOpExpressionList,
+								   List *secondOpExpressionList);
+static DeferredErrorMessage * DeferErrorIfCannotPushdownSubquery(Query *subqueryTree,
+																 bool outerQueryHasLimit);
+static DeferredErrorMessage * ErrorIfUnsupportedUnionQuery(Query *queryTree,
```
metdos: ErrorIfUnsupportedUnionQuery -> naming pattern is diferent.

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -158,6 +182,823 @@ MultiLogicalPlanCreate(Query *originalQuery, Query *queryTree)
...
 
 /*
+ * MultiSubqueryPlanTree gets the query objects and returns logical plan
+ * for subqueries.
+ *
+ * We currently have two different code paths for creating logic plan for subqueries:
+ *   (i) subquery pushdown
+ *   (ii) single relation repartition subquery
+ *
+ * In order to do create the logical plan, we follow the algorithm below:
```
metdos: In order to create -> drop "do"

src/backend/distributed/planner/multi_logical_planner.c
metdos: > The sentence starts with Check for the errors for single table repartition subquery planning. I think using it again may look unnecessary?

Ohh, that one was a note to me to explain differently, but I forgot to update the comment.

metdos: > Well, that would mean that we remove all the single repartition error messages. I think keeping it as it is might be better. Otherwise, someone who is trying to run single repartition subquery would always see subquery pushdown error messages.

The reverse is also correct. Someone who is trying to run subquery pushdown query with a single table will see errors related repartition subquery planning. My argument is that for our use cases, it is more probable that somebody will want to run a subquery pushdown query than a  single table repartition subquery.

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -158,6 +182,823 @@ MultiLogicalPlanCreate(Query *originalQuery, Query *queryTree)
...
+	 */
+	subqueryPushdownError = DeferErrorIfContainsUnsupportedSubquery(originalQuery,
+																	plannerRestrictionContext);
+	if (!subqueryPushdownError)
+	{
+		multiQueryNode = SubqueryPushdownMultiPlanTree(originalQuery);
+	}
+	else if (subqueryPushdownError)
+	{
+		int singleRelationRepartitionSubquery =
```
metdos: This variable should be `bool`.

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -158,6 +182,823 @@ MultiLogicalPlanCreate(Query *originalQuery, Query *queryTree)
...
+			RaiseDeferredErrorInternal(subqueryPushdownError, ERROR);
+		}
+
+		subqueryEntryList = SubqueryEntryList(queryTree);
+		subqueryRangeTableEntry = (RangeTblEntry *) linitial(subqueryEntryList);
+		Assert(subqueryRangeTableEntry->rtekind == RTE_SUBQUERY);
+
+		subqueryTree = subqueryRangeTableEntry->subquery;
+
+		repartitionQueryError = DeferErrorIfSubqueryNotSupported(subqueryTree);
```
metdos: There are both `DeferErrorIfContainsUnsupportedSubquery` and `DeferErrorIfSubqueryNotSupported`. I can't tell the difference from names.

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: > The reverse is also correct. Someone who is trying to run subquery pushdown query with a single table will see errors related repartition subquery planning.

That seems much less frequent to me. Because considering our use cases, if you're to write queries for subquery pushdown, you'd very likely to have multiple `RTE_RELATION`s in the queries. In that case, we'd error out using subquery pushdown errors. For example, all the queries we have in our production customers and `Raw Queries` document have multiple `RTE_RELATION`s. What do you think?

Also, I'm really not comfortable with removing all error messages for repartition subqueries.

The main source of the problem is that we have two separate planners for subqueries, but, we cannot fix it now 😄

metdos: I can think of union/union all queries on a single table. However, If you're comfortable with this approach, I'm okay to continue with this given that you know the example queries better than me.

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -158,6 +182,823 @@ MultiLogicalPlanCreate(Query *originalQuery, Query *queryTree)
...
+
+	return multiQueryNode;
+}
+
+
+/*
+ * SingleRelationRepartitionSubquery returns true if it is eligible single
+ * repartition query planning in the sense that:
+ *   - None of the levels of the subquery contains a join
+ *   - Only a single RTE_RELATION exists
```
metdos: "Only a single RTE_RELATION exists"

Does this also cover the joins?

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -158,6 +182,823 @@ MultiLogicalPlanCreate(Query *originalQuery, Query *queryTree)
...
+	ExtractRangeTableIndexWalker((Node *) queryTree->jointree,
+								 &rangeTableIndexList);
+	if (list_length(rangeTableIndexList) != 1)
+	{
+		return false;
+	}
+
+	rangeTableIndex = linitial_int(rangeTableIndexList);
+	rangeTableEntry = rt_fetch(rangeTableIndex, rangeTableList);
+	if (rangeTableEntry->rtekind == RTE_RELATION)
```
metdos: A suggestion. Would it be more readable like this?;

```c
	if (rangeTableEntry->rtekind == RTE_RELATION)
	{
		return true;
	}
        else if (rangeTableEntry->rtekind == RTE_SUBQUERY)
	{
                Query *subqueryTree = rangeTableEntry->subquery;
		return SingleRelationRepartitionSubquery(subqueryTree);
	}

	return false;



```

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: We don't support repartition subqueries with joins. And, yes that check is intended for returning false for joins

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: Yes, your suggestion looks nicer

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -259,14 +1100,14 @@ MultiPlanTree(Query *queryTree)
...
+
 		subqueryRangeTableEntry = (RangeTblEntry *) linitial(subqueryEntryList);
 		subqueryTree = subqueryRangeTableEntry->subquery;
 
-		/* check if subquery satisfies preconditons */
-		ErrorIfSubqueryNotSupported(subqueryTree);
-
-		/* check if subquery has joining tables */
-		ErrorIfSubqueryJoin(subqueryTree);
+		/* ensure if subquery satisfies preconditons */
```
metdos: typo: preconditons

src/backend/distributed/planner/multi_logical_planner.c
metdos: I know these names come from the previous code, it really bothered me and confused me (mostly likely I wrote both functions 😄 ) My final proposal :)
`DeferErrorIfContainsUnsupportedSubqueryPushdown` -> `DeferErrorIfUnsupportedSubqueryPushdown` (to make it shorter)
`DeferErrorIfSubqueryNotSupported` -> `DeferErrorIfUnsupportedSubqueryRepartition`

What do you think?

src/backend/distributed/planner/multi_logical_planner.c
metdos: I meant "Only a single RTE_RELATION exists" should deduce that we don't have joins. I'm okay to keep the first line, just noting to see I missed anything?

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: > I can think of union/union all queries on a single table. However, If you're comfortable with this approach, I'm okay to continue with this given that you know the example queries better than me.

Each occurrence of a relation generates a separate `RTE_RELATION`, and also since single-repartition queries do not support `UNIONs`, we'd error out using pushdown errors.

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: Both suggestions make sense, I'll update them.

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: Oh, sorry I misunderstood your comment. This function checks for 3 things (comment updated in the PR as well): 

1. No joins at all
2. Only a single `RTE_RELATION`, which means only a single table name is specified on the whole query 
3. No subqueries in WHERE clause

> I meant "Only a single RTE_RELATION exists" should deduce that we don't have joins. I'm okay to keep the first line, just noting to see I missed anything?

In fact, that comment is intended for the second item in the list. Is that better now?

src/test/regress/expected/multi_mx_router_planner.out
```diff
@@ -864,8 +864,8 @@ SELECT * FROM (
 	(SELECT * FROM articles_hash_mx WHERE author_id = 1)
 	UNION
 	(SELECT * FROM articles_hash_mx WHERE author_id = 2)) uu;
-ERROR:  cannot perform distributed planning on this query
-DETAIL:  Subqueries without group by clause are not supported yet
+ERROR:  cannot push down this subquery
+DETAIL:  Currently all leaf queries need to have same filters on partition column
```
metdos: I though that we ditched this kind of leaf query checks. Is this specific to union queries?

src/test/regress/output/multi_subquery.source
```diff
@@ -141,8 +144,8 @@ FROM
 		lineitem_subquery
 	GROUP BY
 		l_suppkey) AS order_counts;
-ERROR:  cannot push down this subquery
-DETAIL:  Group by list without partition column is currently unsupported
+ERROR:  cannot use real time executor with repartition jobs
+HINT:  Set citus.task_executor_type to "task-tracker".
```
metdos: Shall we set task-tracker and run this query?

src/test/regress/expected/multi_mx_router_planner.out
onderkalaci: We have an item to remove those checks, but that's a `should have task`.

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -602,14 +1443,34 @@ ErrorHintRequired(const char *errorHint, Query *queryTree)
...
+
+	/*
+	 * Extract all range table indexes from the join tree. Note that sub-queries
+	 * that get pulled up by PostgreSQL don't appear in this join tree.
+	 */
+	ExtractRangeTableIndexWalker((Node *) subqueryTree->jointree,
+								 &joinTreeTableIndexList);
+
+	joiningRangeTableCount = list_length(joinTreeTableIndexList);
+	if (joiningRangeTableCount > 1)
```
metdos: Can we hit this check? Maybe we can replace this with an Assert?

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -641,13 +1502,28 @@ ErrorIfSubqueryNotSupported(Query *subqueryTree)
...
+
+	/* continue with the inner subquery */
+	rangeTableIndex = linitial_int(joinTreeTableIndexList);
+	rangeTableEntry = rt_fetch(rangeTableIndex, subqueryTree->rtable);
+	if (rangeTableEntry->rtekind == RTE_RELATION)
+	{
+		return NULL;
+	}
+
+
```
metdos: Double new lines.

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -158,6 +183,822 @@ MultiLogicalPlanCreate(Query *originalQuery, Query *queryTree)
...
+	 */
+	subqueryPushdownError = DeferErrorIfUnsupportedSubqueryPushdown(originalQuery,
+																	plannerRestrictionContext);
+	if (!subqueryPushdownError)
+	{
+		multiQueryNode = SubqueryPushdownMultiPlanTree(originalQuery);
+	}
+	else if (subqueryPushdownError)
+	{
+		bool singleRelationRepartitionSubquery =
```
metdos: I would move this to before if block where we use the `singleRelationRepartitionSubquery`.

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -158,6 +183,822 @@ MultiLogicalPlanCreate(Query *originalQuery, Query *queryTree)
...
+	}
+
+	return false;
+}
+
+
+/*
+ * DeferErrorIfContainsUnsupportedSubqueryPushdown iterates on the query's subquery
+ * entry list and uses helper functions to check if we can push down subquery
+ * to worker nodes. These helper functions returns a deferred error out
```
metdos: out in "error out" is extra

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: You're right, we shouldn't hit this. I'll use an `Assert.

onderkalaci: Sorry, we could hit this line. This is basically necessary to generate the error message for that case.

src/test/regress/output/multi_subquery.source
onderkalaci: Probably not? This test became not relevant to this file since we're removing the flag.  We could even consider removing this test?

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: We use `SingleRelationRepartitionSubquery` to decide on whether to print the errors we've generated for subquery pushdown. If we don't produce error for subquery pushdown, we still execute this code and check whether Citus supports re-partition subqueries or not.

metdos: But, should not `SingleRelationRepartitionSubquery` return false on joins, and we raise deferred error for subquery pushdown?

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -158,6 +183,822 @@ MultiLogicalPlanCreate(Query *originalQuery, Query *queryTree)
...
+
+/*
+ * SingleRelationRepartitionSubquery returns true if it is eligible single
+ * repartition query planning in the sense that:
+ *   - None of the levels of the subquery contains a join
+ *   - Only a single RTE_RELATION exists, which means only a single table
+ *     name is specified on the whole query
+ *   - No sublinks exists in the subquery
+ *
+ * Note that the caller should still call DeferErrorIfSubqueryNotSupported() to ensure
```
metdos: We renamed `DeferErrorIfSubqueryNotSupported()`

src/test/regress/expected/multi_subquery_complex_queries.out
```diff
@@ -2025,8 +2024,8 @@ FROM     (
                  ) user_id 
 ORDER BY 1, 
          2 limit 10;
-ERROR:  cannot push down this subquery
-DETAIL:  Window functions are currently unsupported
+ERROR:  cannot perform distributed planning on this query
```
metdos: Now error message does not match with the test comment. Maybe, we can add group by to the subquery to error out on window function?

metdos:
This looks good 🚢  (a minor comment on regression tests)

Some notes;
- `DeferErrorIf` prefix makes already long error functions a bit longer, but I couldn't find a better alternative.
- `DeferErrorIfUnsupportedSubqueryPushdown` calls `DeferErrorIfCannotPushdownSubquery`, and their naming are similar, again I don't have better alternatives. These names are historical and better naming needs more refactoring.
- In future, for any PR if we are moving functions around files, it would be better to have move operations in a separate commit to make code review easier.

onderkalaci:
> (a minor comment on regression tests)

I'll tackle many issues with the regression tests on a separate PR.

> In future, for any PR if we are moving functions around files, it would be better to have move operations in a separate commit to make code review easier.

Yes, I think that's a good idea. Sorry for that!

PR #1316, opened 2017-04-10
Log cross-shard queries

### LINKED ISSUE #1308: Add session-level GUC to log all cross-shard queries
One of the issues that frequently comes up in customers migrating to Citus, especially in a multi-tenant model, is understanding if there are any cross-shard (i.e. realtime/task-tracker executor) queries in a particular part of the customer's application code.

Right now one has to manually check all executed SQL and verify that the tenant_id is present on the queries.

Ideally we'd have a GUC (e.g. `citus.log_dist_statement`, defaulting to `off`) on the coordinator that can be set on a per-session basis to something like `cross_shard` or `multi-shard`.

When enabled all cross-shard queries should be output to the logfile, similar to how `log_statement` is output:

```
2014-11-24 22:17:15 CET LOG:  statement: SELECT * FROM objects WHERE id = 123;
```

---

**Stretch goal:**

We could consider including the duration in the output as well, or even naming this `citus.log_min_duration_multi_shard_statement` and make it work similarly to `log_min_duration_statement` (i.e. `-1` is the default and logs nothing, `0` logs everything, `500` logs every cross-shard/multi-shard query that is slower than 500ms).

This is not required however, since the customer mostly cares whether they missed any cross-shard queries in their app migration work (i.e. forgot to add tenant_id to queries), and would only use this in development.

velioglu:
Fixes #1308 

Inform user about cross shard queries according to the log level defined by citus.log_dist_statement. Uses the same log levels with error level codes.

### DIFF
--- a/src/backend/distributed/executor/multi_server_executor.c
+++ b/src/backend/distributed/executor/multi_server_executor.c
@@ -29,6 +29,7 @@
 int RemoteTaskCheckInterval = 100; /* per cycle sleep interval in millisecs */
 int TaskExecutorType = MULTI_EXECUTOR_REAL_TIME; /* distributed executor type */
 bool BinaryMasterCopyFormat = false; /* copy data from workers in binary format */
+int MultiTaskQueryLogLevel = MULTI_TASK_QUERY_INFO_OFF; /* multi-task query log level */
 
 
 /*
@@ -56,6 +57,15 @@ JobExecutorType(MultiPlan *multiPlan)
 	{
 		ereport(DEBUG2, (errmsg("Plan is router executable")));
 		return MULTI_EXECUTOR_ROUTER;
+	}
+
+	/* if it is not a router executable plan, inform user according to the log level */
+	if (MultiTaskQueryLogLevel != MULTI_TASK_QUERY_INFO_OFF)
+	{
+		ereport(MultiTaskQueryLogLevel, (errmsg("multi-task query about to be executed"),
+										 errhint("Queries are split to multiple tasks "
+												 "if they have to be split into several"
+												 " queries on the workers.")));
 	}
 
 	if (executorType == MULTI_EXECUTOR_REAL_TIME)
--- a/src/backend/distributed/planner/multi_router_planner.c
+++ b/src/backend/distributed/planner/multi_router_planner.c
@@ -34,6 +34,7 @@
 #include "distributed/multi_logical_optimizer.h"
 #include "distributed/multi_physical_planner.h"
 #include "distributed/multi_router_planner.h"
+#include "distributed/multi_server_executor.h"
 #include "distributed/listutils.h"
 #include "distributed/citus_ruleutils.h"
 #include "distributed/relation_restriction_equivalence.h"
@@ -341,6 +342,15 @@ CreateInsertSelectRouterPlan(Query *originalQuery,
 		}
 
 		++taskIdIndex;
+	}
+
+	if (MultiTaskQueryLogLevel != MULTI_TASK_QUERY_INFO_OFF &&
+		list_length(sqlTaskList) > 1)
+	{
+		ereport(MultiTaskQueryLogLevel, (errmsg("multi-task query about to be executed"),
+										 errhint("Queries are split to multiple tasks "
+												 "if they have to be split into several"
+												 " queries on the workers.")));
 	}
 
 	/* Create the worker job */
--- a/src/backend/distributed/shared_library_init.c
+++ b/src/backend/distributed/shared_library_init.c
@@ -92,6 +92,16 @@ static const struct config_enum_entry multi_shard_commit_protocol_options[] = {
 	{ NULL, 0, false }
 };
 
+static const struct config_enum_entry multi_task_query_log_level_options[] = {
+	{ "off", MULTI_TASK_QUERY_INFO_OFF, false },
+	{ "debug", DEBUG2, false },
+	{ "log", LOG, false },
+	{ "notice", NOTICE, false },
+	{ "warning", WARNING, false },
+	{ "error", ERROR, false },
+	{ NULL, 0, false }
+};
+
 /* *INDENT-ON* */
 
 
@@ -610,6 +620,16 @@ RegisterCitusConfigVariables(void)
 					 "The random policy picks all workers randomly."),
 		&ShardPlacementPolicy,
 		SHARD_PLACEMENT_ROUND_ROBIN, shard_placement_policy_options,
+		PGC_USERSET,
+		0,
+		NULL, NULL, NULL);
+
+	DefineCustomEnumVariable(
+		"citus.multi_task_query_log_level",
+		gettext_noop("Sets the level of multi task query execution log messages"),
+		NULL,
+		&MultiTaskQueryLogLevel,
+		MULTI_TASK_QUERY_INFO_OFF, multi_task_query_log_level_options,
 		PGC_USERSET,
 		0,
 		NULL, NULL, NULL);
--- a/src/include/distributed/multi_server_executor.h
+++ b/src/include/distributed/multi_server_executor.h
@@ -36,6 +36,8 @@
 #define JOB_CLEANUP_QUERY "SELECT task_tracker_cleanup_job("UINT64_FORMAT ")"
 #define JOB_CLEANUP_TASK_ID INT_MAX
 
+#define MULTI_TASK_QUERY_INFO_OFF 0  /* do not log multi-task queries */
+
 
 /* Enumeration to track one task's execution status */
 typedef enum
@@ -184,6 +186,7 @@ extern int RemoteTaskCheckInterval;
 extern int MaxAssignTaskBatchSize;
 extern int TaskExecutorType;
 extern bool BinaryMasterCopyFormat;
+extern int MultiTaskQueryLogLevel;
 
 
 /* Function declarations for distributed execution */
--- a/src/test/regress/expected/multi_cross_shard.out
+++ b/src/test/regress/expected/multi_cross_shard.out
@@ -1,0 +1,191 @@
+--
+-- MULTI_CROSS_SHARD
+--
+-- Tests to log cross shard queries according to error log level
+--
+-- Create a distributed table and add data to it
+CREATE TABLE multi_task_table
+(
+	id int, 
+	name varchar(20)
+);
+SELECT create_distributed_table('multi_task_table', 'id');
+ create_distributed_table 
+--------------------------
+ 
+(1 row)
+
+INSERT INTO multi_task_table VALUES(1, 'elem_1');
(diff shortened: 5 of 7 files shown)


src/test/regress/expected/multi_cross_shard.out
```diff
@@ -0,0 +1,51 @@
...
+SELECT * FROM t1;
+ id |  name  
+----+--------
+  1 | elem_1
+  3 | elem_3
+  2 | elem_2
+(3 rows)
+
+-- Get warning messages with the log level 'warning'
+SET citus.log_dist_statement TO warning;
```
metdos: Default error level is `notice`, you don't need to set it to `warning`.
https://www.postgresql.org/docs/9.6/static/runtime-config-logging.html#RUNTIME-CONFIG-SEVERITY-LEVELS

src/backend/distributed/executor/multi_server_executor.c
```diff
@@ -29,6 +29,7 @@
 int RemoteTaskCheckInterval = 100; /* per cycle sleep interval in millisecs */
 int TaskExecutorType = MULTI_EXECUTOR_REAL_TIME; /* distributed executor type */
 bool BinaryMasterCopyFormat = false; /* copy data from workers in binary format */
+int CrossShardLogType = CROSS_SHARD_INFO_OFF; /* cross shard log type */
```
metdos: Rename to `CrossShardLogLevel`?

src/include/distributed/multi_server_executor.h
```diff
@@ -104,6 +104,16 @@ typedef enum
 	CONNECT_ACTION_CLOSED = 2
 } ConnectAction;
 
+/* Enumeration that defines the different log types of cross shard queries */
+typedef enum
+{
+	CROSS_SHARD_INFO_NONE = 0,
+	CROSS_SHARD_INFO_OFF = DEBUG5 - 1, /* TODO : May change it */
```
metdos: This is hacky. Also, we can ditch `CROSS_SHARD_INFO_NONE ` and have `CROSS_SHARD_INFO_OFF ` as the default one.

src/backend/distributed/executor/multi_server_executor.c
```diff
@@ -58,6 +59,9 @@ JobExecutorType(MultiPlan *multiPlan)
 		return MULTI_EXECUTOR_ROUTER;
 	}
 
+	/* if it is not a router executable plan, inform user according to the log level */
```
metdos: We can wrap this in an `if` block;

```
if(CrossShardLogType != CROSS_SHARD_INFO_OFF)
{
....

```

src/backend/distributed/shared_library_init.c
```diff
@@ -613,6 +621,16 @@ RegisterCitusConfigVariables(void)
 		0,
 		NULL, NULL, NULL);
 
+	DefineCustomEnumVariable(
+		"citus.log_dist_statement",
```
metdos: We can use the same name `citus.cross_shard_log_level`

marcocitus:
I don't think we refer to these as "cross shard queries" anywhere, only in #1308, which I found by Googling for "citus cross-shard queries" :). Could we name this GUC something else? (e.g. multi-shard queries / parallel queries / whatever @samay-sharma would call it)

samay-sharma:
@marcocitus : We have been using the word "cross-shard" pretty often now-a-days. I'm not sure where it started, probably from us saying "these operations span across shards".

I don't think we've put thought into what to name these kind of queries. Now that I think of it, cross-shard does seem a bit misleading. I think a more apt alternative would be "multi-shard queries".

How does citus.multi_shard_query_log_level sound for the config parameter? @lfittl , @saicitus , any objections / alternate suggestions on naming?

Also, semi-related, will a multi-shard INSERT / SELECT get logged with this? The reason I ask is that we recently noticed that INSERT SELECT says "router" in the explain plan with 32 tasks. It uses router but is multi-shard, so I think it should get logged but don't know if this change addresses that.

anarazel:
> How does citus.multi_shard_query_log_level sound for the config parameter? @lfittl , @saicitus , any objections / alternate suggestions on naming?

Isn't that also quite misleading? Given that router can query multiple
shards without any issue.  Maybe we should name it multi_task_query or
such?

samay-sharma:
Hmm, good point. I think task is a concept we don't introduce to our users. So, I'd be little hesitant to add a config with that name. Can we use a higher level but accurate term for this (I can't think of anything)?

Also, this brings us to a more higher level semi-related question around communicating Citus's SQL support with router. Queries going to a single-shard isn't accurate (because of co-located tables), single-node also isn't completely accurate (because single node can have multiple tasks whose results need to be combined). I won't hijack this conversation with discussion on this, but just something to think about and make more concrete.

marcocitus:
At least "Task" is something we use in EXPLAIN output, so that's quite visible to users. When you run `EXPLAIN` on a query it tells you whether or not it's a multi-task query.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -169,6 +170,11 @@ CreateModifyPlan(Query *originalQuery, Query *query,
 {
 	if (InsertSelectQuery(originalQuery))
 	{
+		if (MultiTaskQueryLogLevel != MULTI_TASK_QUERY_INFO_OFF)
```
metdos: An InsertSelectQuery can have only one task, so you also need to check task count on this. You can move this check into  CreateInsertSelectRouterPlan() and also check task count.

src/backend/distributed/executor/multi_server_executor.c
```diff
@@ -58,6 +59,13 @@ JobExecutorType(MultiPlan *multiPlan)
...
+	if (MultiTaskQueryLogLevel != MULTI_TASK_QUERY_INFO_OFF)
+	{
+		ereport(MultiTaskQueryLogLevel, (errmsg("this query is multi-task query, that "
```
anarazel: There's a missing 'a' here, but I'd rephrase a bit more aggressively.  How about something roughly like`errmsg("multi-task about to be executed"), errhint("Queries use multiple tasks if they have to be split into several queries on the workers.")`

src/backend/distributed/shared_library_init.c
```diff
@@ -92,6 +92,15 @@ static const struct config_enum_entry multi_shard_commit_protocol_options[] = {
 	{ NULL, 0, false }
 };
 
+static const struct config_enum_entry multi_task_query_options[] = {
+	{ "off", MULTI_TASK_QUERY_INFO_OFF, false },
+	{ "log", MULTI_TASK_QUERY_INFO_LOG, false },
+	{ "notice", MULTI_TASK_QUERY_INFO_NOTICE, false },
+	{ "warning", MULTI_TASK_QUERY_INFO_WARNING, false },
+	{ "error", MULTI_TASK_QUERY_INFO_ERROR, false },
+	{ NULL, 0, false }
```
anarazel: I'd also allow some debug log level.

src/include/distributed/multi_server_executor.h
```diff
@@ -104,6 +104,16 @@ typedef enum
...
 
+/* Enumeration that defines the different log types of cross shard queries */
+typedef enum
+{
+	MULTI_TASK_QUERY_INFO_OFF = 0,
+	MULTI_TASK_QUERY_INFO_LOG = LOG,
+	MULTI_TASK_QUERY_INFO_NOTICE = NOTICE,
+	MULTI_TASK_QUERY_INFO_WARNING = WARNING,
+	MULTI_TASK_QUERY_INFO_ERROR = ERROR
+} MultiTaskQueryLogLevels;
```
anarazel: Seems a bit superflous to copy the definitions of LOG etc here, but doesn't matter much.

src/test/regress/expected/multi_cross_shard.out
```diff
@@ -0,0 +1,51 @@
+--
+-- MULTI_CROSS_SHARD
+--
+-- Tests to log cross shard queries according to error log level
+--
+ALTER SEQUENCE pg_catalog.pg_dist_shardid_seq RESTART 1440000;
+ALTER SEQUENCE pg_catalog.pg_dist_jobid_seq RESTART 1440000;
```
anarazel: No need for these resets here, no shard / job ids are being displayed.

src/test/regress/expected/multi_cross_shard.out
```diff
@@ -0,0 +1,51 @@
...
+  2 | elem_2
+(3 rows)
+
+SELECT AVG(id) AS avg_id FROM multi_task_tt;
+NOTICE:  this query is multi-task query, that spanning across shards
+       avg_id       
+--------------------
+ 2.0000000000000000
+(1 row)
+
```
anarazel: I think this also needs a test-case that ERRORs out.

src/test/regress/multi_schedule
```diff
@@ -33,6 +33,7 @@ test: multi_load_data
 test: multi_insert_select_behavioral_analytics_create_table
 test: multi_insert_select_behavioral_analytics_basics multi_insert_select_behavioral_analytics_single_shard_queries multi_insert_select_non_pushable_queries
 test: multi_insert_select
+test: multi_cross_shard
```
anarazel: Any chance you could find a parallel group to fit this in?  Without aggressively looking for that our test times will grow way too quickly.

src/test/regress/sql/multi_cross_shard.sql
```diff
@@ -0,0 +1,64 @@
...
+SET citus.multi_task_query_log_level TO error;
+INSERT INTO summary_table SELECT id, SUM(order_count) FROM raw_table GROUP BY id;
+
+-- Should have only three rows
+SET citus.multi_task_query_log_level to DEFAULT;
+SELECT * FROM summary_table;
+
+DROP TABLE multi_task_tt;
+DROP TABLE raw_table;
+DROP TABLE summary_table;
```
metdos: We should also add queries for task_tracker, router-select and single task insert-into select.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -343,6 +344,15 @@ CreateInsertSelectRouterPlan(Query *originalQuery,
 		++taskIdIndex;
 	}
 
+	if (list_length(sqlTaskList) > 1 && MultiTaskQueryLogLevel !=
```
metdos: If we put `MultiTaskQueryLogLevel != MULTI_TASK_QUERY_INFO_OFF`, these would be more readable.

```
if (list_length(sqlTaskList) > 1 && 
MultiTaskQueryLogLevel != MULTI_TASK_QUERY_INFO_OFF)
```
```
if (MultiTaskQueryLogLevel != MULTI_TASK_QUERY_INFO_OFF && 
   (list_length(sqlTaskList) > 1
```

metdos:
Looks good, squash and 🚀

PR #1318, opened 2017-04-11
Change checks with built-in type (omit ltree)

velioglu:
This commit checks copying data to the table which has a column without binary output function.


metdos:
For context: https://github.com/citusdata/citus/pull/1300#issuecomment-293186757

anarazel:
@velioglu @metdos  Any reason this was not backported?  The label on the issue seems to indicate it should have been?

PR #1320, opened 2017-04-11
Support UPDATE/DELETE with parameterised partition column qual

### LINKED ISSUE #1319: Prepared UPDATE/DELETE errors out after 5th invocation
To reproduce:
```
PREPARE foo(int,int) AS UPDATE test SET y = $2 WHERE x = $1;
EXECUTE foo(1, 2);
EXECUTE foo(1, 2);
EXECUTE foo(1, 2);
EXECUTE foo(1, 2);
EXECUTE foo(1, 2);
EXECUTE foo(1, 2);
ERROR:  cannot run UPDATE command which targets multiple shards
HINT:  Consider using an equality filter on partition column "x" to target a single shard. If you'd like to run a multi-shard operation, use master_modify_multiple_shards().
```

The issue is that `TargetShardIntervalForModify` throws an error if it cannot prune down to 1 shard, which it cannot if the partition column value is defined as a parameter. However, if it instead returned a `DeferredError` up the chain, then the planner would pick the plan without a parameter.

marcocitus:
Fixes #1319 by setting a `planningError` if `TargetShardIntervalForModify` fails to prune down to 1 shard instead of throwing an error. Our regression tests already covered the unsupported cases, but the expected output contained errors. We should backport this since it's a hole in the expected prepared statement support of 6.1.

@metdos assigning you as our prepared statement expert :), but feel free to hand it off.

### DIFF
--- a/src/backend/distributed/planner/multi_router_planner.c
+++ b/src/backend/distributed/planner/multi_router_planner.c
@@ -100,8 +100,9 @@ static bool MasterIrreducibleExpressionWalker(Node *expression, WalkerState *sta
 static char MostPermissiveVolatileFlag(char left, char right);
 static bool TargetEntryChangesValue(TargetEntry *targetEntry, Var *column,
 									FromExpr *joinTree);
-static Task * RouterModifyTask(Query *originalQuery, Query *query);
-static ShardInterval * TargetShardIntervalForModify(Query *query);
+static Task * RouterModifyTask(Query *originalQuery, ShardInterval *shardInterval);
+static ShardInterval * TargetShardIntervalForModify(Query *query,
+													DeferredErrorMessage **planningError);
 static List * QueryRestrictList(Query *query);
 static bool FastShardPruningPossible(CmdType commandType, char partitionMethod);
 static Const * ExtractInsertPartitionValue(Query *query, Var *partitionColumn);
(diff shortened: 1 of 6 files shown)


sumedhpathak:
@metdos Could you review this change? Thanks!

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -2101,6 +2114,7 @@ TargetShardIntervalForModify(Query *query)
 		StringInfo errorHint = makeStringInfo();
 		const char *targetCountType = NULL;
 		bool showHint = false;
+		StringInfo errorMessage = makeStringInfo();
```
metdos: I would declare `errorMessage` just before `errorHint `.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -2141,10 +2155,14 @@ TargetShardIntervalForModify(Query *query)
...
 		showHint = errorHint->len > 0;
 
-		ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
-						errmsg("cannot run %s command which targets %s shards",
-							   commandName, targetCountType),
-						showHint ? errhint("%s", errorHint->data) : 0));
+		appendStringInfo(errorMessage, "cannot run %s command which targets %s shards",
+						 commandName, targetCountType);
+
+		*planningError = DeferredError(ERRCODE_FEATURE_NOT_SUPPORTED,
```
metdos:
```
(*planningError) = DeferredError(ERRCODE_FEATURE_NOT_SUPPORTED, ...)
```

src/test/regress/expected/multi_prepare_sql.out
```diff
@@ -847,9 +843,7 @@ SELECT * FROM prepare_table ORDER BY key, value;
    0 |      
    0 |      
    0 |      
-   6 |    60
-   6 |      
-(8 rows)
+(6 rows)
```
metdos: While trying to understand why this table changed on this file without any other update, I noticed that there is a copy/paste mistake here and we use a table from another regression test. I'm pushing a commit to fix it.

marcocitus:
Thanks for the review and the fix up!

anarazel:
@marcocitus Any reason this was not backported?  The label seems to indicate it should?

PR #1321, opened 2017-04-12
Support query parameters in combination with function evaluation

### LINKED ISSUE #983: Prepared inserts via JDBC fail when containing VARCHAR.
Step to reproduce:
```
create table t
(
  i character varying(20),
  j character varying(20),
  k timestamp without time zone default now(),
  constraint t_pkey primary key (i)
);
select * from create_distributed_table('t', 'i');
prepare fooplan (text) as insert into t values($1);
execute fooplan ('bug');
```

encountered error:
```
no value found for parameter 1
```

I dig a bit further. It seems passing default econtext is not covering all cases,
```
    /*
     * And evaluate it.
     *
     * It is OK to use a default econtext because none of the ExecEvalExpr()
     * code used in this situation will use econtext.  That might seem
     * fortuitous, but it's not so unreasonable --- a constant expression does
     * not depend on context, by definition, n'est ce pas?
     */
    const_val = ExecEvalExprSwitchContext(exprstate,
                                          GetPerTupleExprContext(estate),
                                          &const_is_null, NULL);
```
which leads to
```
static Datum
ExecEvalParamExtern(ExprState *exprstate, ExprContext *econtext,
                    bool *isNull, ExprDoneCond *isDone)
{
    Param      *expression = (Param *) exprstate->expr;
    int         thisParamId = expression->paramid; <---- this is 1
    ParamListInfo paramInfo = econtext->ecxt_param_list_info; <---- this is NULL
```

### LINKED ISSUE #1304: Timestamp column with default causes JDBC error
If I have a sharded table that looks like this:

    create table uuidsample (id uuid,
                         other_id bigint,
                         name varchar(50) not null,
                         namespace varchar(50) not null,
                         create_date timestamptz not null default now());

And some JDBI code like this:

```
// line breaks added for readability
@SqlUpdate("insert into uuidsample(id, other_id, name, namespace) 
         values (:id, :other_id, :name, :namespace) on conflict do nothing")
int createUUID(@Bind("id") UUID id, @Bind("other_id") long otherId, 
        @Bind("name") String name, @Bind("namespace") String namespace);
```

Attempting to insert a row gives me this error:

```
    org.skife.jdbi.v2.exceptions.UnableToExecuteStatementException: org.postgresql.util.PSQLException: 
ERROR: no value found for parameter 3 [statement:"insert into uuidsample(id, other_id, name, namespace) values (:id, :other_id, :name, :namespace) on conflict do nothing", 
located:"insert into uuidsample(id, other_id, name, namespace) values (:id, :other_id, :name, :namespace) on conflict do nothing", 
rewritten:"insert into uuidsample(id, other_id, name, namespace) values (?, ?, ?, ?) on conflict do nothing", 
arguments:{ positional:{}, named:{name:'asdf',namespace:'ns',id:4ac7d33f-a963-4fcc-b20c-17ad143524a6,other_id:1}, finder:[]}]
```

But getting rid of the create_date column causes the error to go away. This code works fine with plain old unsharded postgres table of same construction.

Thoughts? I have uploaded the code used to repro this to https://github.com/trasa/citus_default_date_bug/

### LINKED ISSUE #1317: Table's DEFAULT now() not working with stored procedures
I have the following table (example) with DEFAULT now() for the time stamp fields.

```sql
CREATE TABLE example (
    key text NOT NULL,
    seq int4 NOT NULL,
    created_at timestamptz NOT NULL DEFAULT now(),
    updated_at timestamptz NOT NULL DEFAULT now(),
    PRIMARY KEY (key, seq)

);

SELECT create_distributed_table('example', 'key');
```

I want to use the following stored procedure to insert data into the above table.

```sql
CREATE OR REPLACE FUNCTION example_fails_1(key text) RETURNS VOID AS
$BODY$
    #variable_conflict use_variable
    DECLARE
        max_seq int4;
    BEGIN
        SELECT MAX(seq) INTO max_seq
        FROM example
        WHERE example.key = key;

        IF max_seq IS NULL THEN
            max_seq := 0;
        END IF;

        INSERT INTO example(key, seq) VALUES (key, max_seq + 1);
    END;
$BODY$
LANGUAGE plpgsql;
```

When I call this stored procedure I get the following error:
```
problem=# select * from example_fails_1('key');
ERROR:  no value found for parameter 3
CONTEXT:  SQL statement "INSERT INTO example(key, seq) VALUES (key, max_seq + 1)"
PL/pgSQL function example_fails_1(text) line 14 at SQL statement
```

Apparently it is missing parameter 3 which could be the column `created_at` so I created the following stored procedure that sets `created_at` and `updated_at` using `now()`: 

```sql
CREATE OR REPLACE FUNCTION example_fails_2(key text) RETURNS VOID AS
$BODY$
    #variable_conflict use_variable
    DECLARE
        max_seq int4;
    BEGIN
        SELECT MAX(seq) INTO max_seq
        FROM example
        WHERE example.key = key;

        IF max_seq IS NULL THEN
            max_seq := 0;
        END IF;

        INSERT INTO example(key, seq, created_at, updated_at) VALUES (key, max_seq + 1, now(), now());
    END;
$BODY$
LANGUAGE plpgsql;
```

Unfortunately, when I call this stored procedure I get the same error:

```
problem=# select * from example_fails_2('key');
ERROR:  no value found for parameter 3
CONTEXT:  SQL statement "INSERT INTO example(key, seq, created_at, updated_at) VALUES (key, max_seq + 1, now(), now())"
PL/pgSQL function example_fails_2(text) line 14 at SQL statement
```

However, when I create a variable `now` and assign `now()` to it and use that variable it works. Stored procedure shown below:

```sql
CREATE OR REPLACE FUNCTION example_succeeds(key text) RETURNS VOID AS
$BODY$
    #variable_conflict use_variable
    DECLARE
        max_seq int4;
        now timestamptz = now();
    BEGIN
        SELECT MAX(seq) INTO max_seq
        FROM example
        WHERE example.key = key;

        IF max_seq IS NULL THEN
            max_seq := 0;
        END IF;

        INSERT INTO example(key, seq, created_at, updated_at) VALUES (key, max_seq + 1, now, now);
    END;
$BODY$
LANGUAGE plpgsql;
```

Is this a bug or am I doing something wrong? I'm using PostgreSQL 9.6 with Citus 8c5d0f686b5f72674e1a376424af5048603b919f (April 4th, 2017).

marcocitus:
Currently, prepared statements which require function evaluation (e.g. have a default `now()`) don't work because the function evaluation logic throws an error when it runs into an unbound parameter.

This change passes down the PlanState from the executor into the function evaluation logic to be able to evaluate parameters.

Fixes #1317
Fixes #1304 
Fixes #983

### DIFF
--- a/src/backend/distributed/utils/citus_clauses.c
+++ b/src/backend/distributed/utils/citus_clauses.c
@@ -22,11 +22,21 @@
 #include "utils/datum.h"
 #include "utils/lsyscache.h"
 
-static Node * PartiallyEvaluateExpression(Node *expression);
-static Node * EvaluateNodeIfReferencesFunction(Node *expression);
-static Node * PartiallyEvaluateExpressionMutator(Node *expression, bool *containsVar);
+
+typedef struct FunctionEvaluationContext
+{
+	PlanState *planState;
+	bool containsVar;
+} FunctionEvaluationContext;
+
+
+/* private function declarations */
+static Node * PartiallyEvaluateExpression(Node *expression, PlanState *planState);
+static Node * EvaluateNodeIfReferencesFunction(Node *expression, PlanState *planState);
+static Node * PartiallyEvaluateExpressionMutator(Node *expression,
+												 FunctionEvaluationContext *context);
 static Expr * citus_evaluate_expr(Expr *expr, Oid result_type, int32 result_typmod,
-								  Oid result_collation);
+								  Oid result_collation, PlanState *planState);
 
 
 /*
@@ -88,7 +98,7 @@ RequiresMasterEvaluation(Query *query)
  * any sub-expressions which don't include Vars.
  */
 void
-ExecuteMasterEvaluableFunctions(Query *query)
+ExecuteMasterEvaluableFunctions(Query *query, PlanState *planState)
 {
 	CmdType commandType = query->commandType;
 	ListCell *targetEntryCell = NULL;
@@ -99,7 +109,8 @@ ExecuteMasterEvaluableFunctions(Query *query)
 
 	if (query->jointree && query->jointree->quals)
 	{
-		query->jointree->quals = PartiallyEvaluateExpression(query->jointree->quals);
+		query->jointree->quals = PartiallyEvaluateExpression(query->jointree->quals,
+															 planState);
 	}
 
 	foreach(targetEntryCell, query->targetList)
@@ -114,11 +125,13 @@ ExecuteMasterEvaluableFunctions(Query *query)
 
 		if (commandType == CMD_INSERT && !insertSelectQuery)
 		{
-			modifiedNode = EvaluateNodeIfReferencesFunction((Node *) targetEntry->expr);
+			modifiedNode = EvaluateNodeIfReferencesFunction((Node *) targetEntry->expr,
+															planState);
 		}
 		else
 		{
-			modifiedNode = PartiallyEvaluateExpression((Node *) targetEntry->expr);
+			modifiedNode = PartiallyEvaluateExpression((Node *) targetEntry->expr,
+													   planState);
 		}
 
 		targetEntry->expr = (Expr *) modifiedNode;
@@ -133,14 +146,14 @@ ExecuteMasterEvaluableFunctions(Query *query)
 			continue;
 		}
 
-		ExecuteMasterEvaluableFunctions(rte->subquery);
+		ExecuteMasterEvaluableFunctions(rte->subquery, planState);
 	}
 
 	foreach(cteCell, query->cteList)
 	{
 		CommonTableExpr *expr = (CommonTableExpr *) lfirst(cteCell);
 
-		ExecuteMasterEvaluableFunctions((Query *) expr->ctequery);
+		ExecuteMasterEvaluableFunctions((Query *) expr->ctequery, planState);
 	}
 }
 
@@ -150,10 +163,11 @@ ExecuteMasterEvaluableFunctions(Query *query)
  * doesn't show up in the parameter list.
  */
 static Node *
-PartiallyEvaluateExpression(Node *expression)
+PartiallyEvaluateExpression(Node *expression, PlanState *planState)
 {
-	bool unused;
-	return PartiallyEvaluateExpressionMutator(expression, &unused);
+	FunctionEvaluationContext globalContext = { planState, false };
+
+	return PartiallyEvaluateExpressionMutator(expression, &globalContext);
 }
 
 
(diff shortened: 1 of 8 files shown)


src/backend/distributed/utils/citus_clauses.c
```diff
@@ -133,14 +146,16 @@ ExecuteMasterEvaluableFunctions(Query *query)
 			continue;
 		}
 
-		ExecuteMasterEvaluableFunctions(rte->subquery);
+		ExecuteMasterEvaluableFunctions(rte->subquery,
```
lithp: Unnecessary newline

src/backend/distributed/utils/citus_clauses.c
```diff
@@ -133,14 +146,16 @@ ExecuteMasterEvaluableFunctions(Query *query)
...
+										planState);
 	}
 
 	foreach(cteCell, query->cteList)
 	{
 		CommonTableExpr *expr = (CommonTableExpr *) lfirst(cteCell);
 
-		ExecuteMasterEvaluableFunctions((Query *) expr->ctequery);
+		ExecuteMasterEvaluableFunctions((Query *) expr->ctequery,
```
lithp: Unnecessary newline

src/backend/distributed/utils/citus_clauses.c
```diff
@@ -133,14 +146,16 @@ ExecuteMasterEvaluableFunctions(Query *query)
 			continue;
 		}
 
-		ExecuteMasterEvaluableFunctions(rte->subquery);
+		ExecuteMasterEvaluableFunctions(rte->subquery,
+										planState);
```
lithp: Unnecessary newline

src/backend/distributed/utils/citus_clauses.c
```diff
@@ -317,18 +340,23 @@ citus_evaluate_expr(Expr *expr, Oid result_type, int32 result_typmod,
 	 * Prepare expr for execution.  (Note: we can't use ExecPrepareExpr
 	 * because it'd result in recursively invoking eval_const_expressions.)
 	 */
-	exprstate = ExecInitExpr(expr, NULL);
+	exprstate = ExecInitExpr(expr, planState);
+
+	if (planState != NULL)
+	{
+		/* use executor's context to pass down parameters */
+		econtext = planState->ps_ExprContext;
```
lithp: It's insane how much code you had to touch to make this one change!

src/backend/distributed/utils/citus_clauses.c
```diff
@@ -317,18 +340,23 @@ citus_evaluate_expr(Expr *expr, Oid result_type, int32 result_typmod,
...
 	 * And evaluate it.
-	 *
-	 * It is OK to use a default econtext because none of the ExecEvalExpr()
-	 * code used in this situation will use econtext.  That might seem
-	 * fortuitous, but it's not so unreasonable --- a constant expression does
-	 * not depend on context, by definition, n'est ce pas?
 	 */
-	const_val = ExecEvalExprSwitchContext(exprstate,
-										  GetPerTupleExprContext(estate),
+	const_val = ExecEvalExprSwitchContext(exprstate, econtext,
```
lithp: This newline is no longer required

src/include/distributed/multi_executor.h
```diff
@@ -31,6 +31,7 @@ typedef struct CitusScanState
 	MultiPlan *multiPlan;             /* distributed execution plan */
 	MultiExecutorType executorType;   /* distributed executor type */
 	bool finishedRemoteScan;          /* flag to check if remote scan is finished */
+	EState *estate;                   /* executor state */
```
lithp: This is unused

PR #1323, opened 2017-04-13
Subquery pushdown - main branch

onderkalaci:
We check-in the subquery pushdown work to this branch. I'm opening this PR to see the whole change at once.

In the final step, we'll check this branch to the master.


metdos:
We added lots of regression tests, I have some comments on regression tests before merging this to the master branch. We have chatted with @onderkalaci on these in other PRs, but have not come to a conclusion.

i. We have multiple new tests. Possibly(?), some of them test similar code paths. Do we want to spend some time to find overlapping tests and eliminate them?

ii. We have main complex queries, and their multiple versions with minor tweaks to test mostly unsupported cases. Do we want to keep these slightly modified complex queries for unsupported cases, or do we want to simplify them?

iii. I think we don't have a coding convention for regression tests. However, personally, I would like to have a better format for regressions tests, especially for complex ones to understand queries with less effort.

Obviously, there is a time trade-off, but I wanted to note these issues and make an informed decision on these. @onderkalaci, please add any comments you have.

onderkalaci:
To start with, I'm not in favor of removing any of the tests here -- at least for non-failure scenarios.

> We added lots of regression tests, I have some comments on regression tests before merging this to the master branch

Could you also share your motivation for asking changes on the regression tests?

> i. We have multiple new tests. Possibly(?), some of them test similar code paths. Do we want to spend some time to find overlapping tests and eliminate them?

Not really. We've added 3 different files:

* `multi_subquery_behavioral_analytics.sql`: This file contains the queries that the private `Raw Queries: Expanding SQL Coverage` file includes. There might be queries that execute the same code-paths over and over. However, keeping them here (i) makes me feel more confident (ii) constitutes some kind of documentation (iii) Each query has a small explanation on it so that anyone can learn some popular query structure for behavioral analytics.

* `multi_subquery_complex_queries.sql`: This file has queries that we support in the production for some of the existing customers. It is basically a regression test file which aims to ensure that the new changes won't break production customers. I think we can consider finding a better name for the file. Also, as far as I can see, each query has a different structure. Although they look similar, each has a different structure such as: 
   * `..UNION..UNION..INNER JOIN..`
   * `..UNION..UNION..LEFT JOIN JOIN..`
   * `..UNION ALL ..UNION ALL ..INNER JOIN..`
   * `..UNION ALL ..UNION ALL ..LEFT JOIN JOIN..`

So, the sense of duplication comes from things such as: (i) The same query with `UNION` and `UNION ALL`s given that we plan them differently (ii) Same structure with `INNER` and `LEFT JOINs` (iii) Same queries where target list entries are shuffled. -- We need to test this given that subquery planning uses the positions of target entries. (iv) Failure scenarios with small changes such as `=` become '!=' or so.

* `multi_subquery_union.sql`: This file is intended for `UNION` queries that do not involve any `JOIN`s, but only set operations. This file also includes some of the remaining queries from `Raw Queries` document and some of the existing customer production queries. And the sense of repeated queries is similar to the above item. 

And we also have `multi_subquery.input`, which I wanted to keep because it executes composite key tests and varchar tests etc.

> ii. We have main complex queries, and their multiple versions with minor tweaks to test mostly unsupported cases. Do we want to keep these slightly modified complex queries for unsupported cases, or do we want to simplify them?

I think each failure case has a single line comment that explains why the query fails. Thus, I don't find them hard to follow. 

Do you find them hard to follow? Also, do you see value in simplifying such queries?

There might be some overlapping test cases which execute the similar failure cases. We can consider removing those. But, given that planning & execution times of such queries is very low, I'm OK with keeping them. 

We may consider moving such queries to the end of the test files?

(A note on the `UNION`/`UNION ALL`failure cases: Given that they form a tree, having a different number of leaf queries triggers different code paths. This is one another reason for long queries.)

> iii. I think we don't have a coding convention for regression tests. However, personally, I would like to have a better format for regressions tests, especially for complex ones to understand queries with less effort.

Unfortunately, we don't have a convention for SQL queries. But, I don't think that the queries are hard to follow just because of their format. They are hard to follow because they are long and complex. Unless we have a convention, it is subjective to say `better format`. 

I think the main thing with the queries is that they are not very consistent. I'll go over the tests and try to fix the indentation issues and make all tests look more consistent. (after we check-in the last PR)

velioglu:
With @onderkalaci we get the benchmark results of main branch and subquery_main branch on AWS (c4.4xlarge) with 2 workers. Running query was the 2nd query of the Heap. Note that SC represents the shard counter and following 4 tables show the running time of SubquerySQLTaskList, Multi Physical Plan Create, Create Distributed Plan and multi_planner. (Time is given in the sec.msec format)

|  SubquerySQL TaskList Time | SC = 16 | SC = 32 | SC = 64 | SC = 128 | SC = 256 | SC = 512 | SC = 1024 |
|---------------------------------|---------|---------|---------|----------|----------|----------|-----------|
| Master Branch                   | 00.001  | 00.003  | 00.006  | 00.012   | 00.022   | 00.052   | 00.106    |
| Subquery Branch                 | 00.005  | 00.014  | 00.042  | 00.150   | 00.560   | 02.148   | 08.395    |


| Multi Physical Plan Create Time | SC = 16 | SC = 32 | SC = 64 | SC = 128 | SC = 256 | SC = 512 | SC = 1024 |
|---------------------------------|---------|---------|---------|----------|----------|----------|-----------|
| Master Branch                   | 00.001  | 00.003  | 00.006  | 00.013   | 00.022   | 00.054   | 00.112    |
| Subquery Branch                 | 00.005  | 00.014  | 00.043  | 00.150   | 00.560   | 02.150   | 08.400    |

| Create Distributed Plan Time | SC = 16 | SC = 32 | SC = 64 | SC = 128 | SC = 256 | SC = 512 | SC = 1024 |
|------------------------------|---------|---------|---------|----------|----------|----------|-----------|
| Master Branch                | 00.002  | 00.004  | 00.007  | 00.014   | 00.026   | 00.061   | 00.125    |
| Subquery Branch              | 00.005  | 00.014  | 00.044  | 00.152   | 00.564   | 02.157   | 08.414    |

| multi_planner Time | SC = 16 | SC = 32 | SC = 64 | SC = 128 | SC = 256 | SC = 512 | SC = 1024 |
|--------------------|---------|---------|---------|----------|----------|----------|-----------|
| Master Branch      | 00.002  | 00.004  | 00.008  | 00.015   | 00.026   | 00.061   | 00.126    |
| Subquery Branch    | 00.006  | 00.015  | 00.044  | 00.153   | 00.564   | 02.158   | 08.414    |

PR #1324, opened 2017-04-13
Replace native hash function with worker_hash

### LINKED ISSUE #1279: INSERT INTO SELECT fails if the partition column doesn't have a native hash function
**How to replicate**

```sql
CREATE TABLE test
(
    id BIGINT,
    time DATE
);

CREATE TABLE test_second
(
    id BIGINT,
    time DATE
);

SELECT create_distributed_table('test', 'time');
SELECT create_distributed_table('test_second', 'time');

INSERT INTO test_second SELECT * FROM test;
```

Fails with 
```sql
postgres=# INSERT INTO test_second SELECT * FROM test;
ERROR:  function pg_catalog.hashint4(date) does not exist
```

**Reason:** PostgreSQL does not have a native hash function for every data type. As a result, if the partition key is one of them like date, bytea the worker query fails. 

**Fix:** We had the same problem in tenant isolation. For a proper fix, we can replace hash_xxx function calls with [worker_hash()](https://github.com/citusdata/citus/blob/4a84ee5512f94bab15d38377abdf646e016586b3/src/backend/distributed/master/master_split_shards.c#L61) function. Check the enterprise repo for example usage!

velioglu:
Fixes #1279.

Since partition columns without native hash function error out while using them with INSERT INTO ... SELECT queries, native hash function is replaced by worker_hash function.

### DIFF
--- a/src/backend/distributed/utils/metadata_cache.c
+++ b/src/backend/distributed/utils/metadata_cache.c
@@ -26,8 +26,10 @@
 #include "commands/extension.h"
 #include "commands/trigger.h"
 #include "distributed/colocation_utils.h"
+#include "distributed/citus_ruleutils.h"
 #include "distributed/master_metadata_utility.h"
 #include "distributed/metadata_cache.h"
+#include "distributed/multi_logical_optimizer.h"
 #include "distributed/pg_dist_local_group.h"
 #include "distributed/pg_dist_node.h"
 #include "distributed/pg_dist_partition.h"
@@ -99,6 +101,7 @@ static Oid distShardPlacementNodeidIndexId = InvalidOid;
 static Oid distTransactionRelationId = InvalidOid;
 static Oid distTransactionGroupIndexId = InvalidOid;
 static Oid extraDataContainerFuncId = InvalidOid;
+static Oid workerHashFunctionId = InvalidOid;
 
 /* Citus extension version variables */
 bool EnableVersionChecks = true; /* version checks are enabled */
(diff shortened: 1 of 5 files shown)


onderkalaci:
> hashint4 function is replaced by worker_hash functio

A minor note on the title: I think this definition does not look accurate enough. `INSERT .. SELECT` does not necessarily uses `inthash4`, it uses `partition_key_hash_func` when necessary.

sumedhpathak:
Is this only for Insert...Select queries? Or are we replacing the hash-function for all hash-partitioned types?

metdos:
> Is this only for Insert...Select queries? Or are we replacing the hash-function for all hash-partitioned types?

This is just for Insert...Select queries. We use same hash-functions but they will be resolved at worker nodes.

src/test/regress/sql/multi_insert_select.sql
```diff
@@ -1574,6 +1574,29 @@ INSERT INTO text_table (part_col) SELECT val FROM text_table;
...
+(
+    id BIGINT,
+    time DATE
+);
+
+SELECT create_distributed_table('test', 'time');
+SELECT create_distributed_table('test_second', 'time');
+
+INSERT INTO test VALUES(1, '11-11-1980');
+INSERT INTO test_second SELECT * FROM test;
```
metdos: Maybe, we can use more meaningful names for temporary tables;

`test ` -> `raw_table`
`test_second ` -> `summary_table`

Then;
```sql
INSERT INTO summary_table SELECT time, count(*) FROM raw_table;
```

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -1452,6 +1455,28 @@ CitusExtraDataContainerFuncId(void)
 }
 
 
+/* return oid of the worker_hash function */
+Oid
+CitusWorkerHashFunctionId(void)
+{
+	Oid citusExtensionOid = InvalidOid;
+	Oid citusSchemaOid = InvalidOid;
+	char *citusSchemaName = 0;
```
metdos: This should be NULL because this is a pointer.

metdos: Actually, we can get these declarations into if block (they are only used there) and merge them with definitions.

src/test/regress/sql/multi_insert_select.sql
```diff
@@ -1574,6 +1574,29 @@ INSERT INTO text_table (part_col) SELECT val FROM text_table;
...
+(
+    id BIGINT,
+    time DATE
+);
+
+SELECT create_distributed_table('raw_table', 'time');
+SELECT create_distributed_table('summary_table', 'time');
+
+INSERT INTO raw_table VALUES(1, '11-11-1980');
+INSERT INTO summary_table SELECT * FROM raw_table;
```
metdos: This looks more meaningful I think;

```sql
INSERT INTO summary_table SELECT time, count(*) FROM raw_table GROUP BY time;
```

So we also need to update the `summary_table` table.
```sql
CREATE TABLE summary_table
 (
          time DATE,
          count BIGINT
 );
```

metdos:
Looks good 🚀

PR #1326, opened 2017-04-17
Fix node copy error

byucesoy:
Instead of directly returning heap tuple obtained from heap scan we return copied version of it.

### DIFF
--- a/src/backend/distributed/utils/node_metadata.c
+++ b/src/backend/distributed/utils/node_metadata.c
@@ -639,6 +639,7 @@ GetNodeTuple(char *nodeName, int32 nodePort)
 	ScanKeyData scanKey[scanKeyCount];
 	SysScanDesc scanDescriptor = NULL;
 	HeapTuple heapTuple = NULL;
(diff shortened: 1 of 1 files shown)


metdos:
🚢

ozgune:
@byucesoy -- I had a quick question. What was our motivation for this change? **If** we fixed an issue with this change, is there an easy regression test that would have captured the issue?

If not, please overlook my comment.

byucesoy:
Hey @ozgune, before this PR, GetHeapTuple function searches for a HeapTuple via heap scan and directly returns the found HeapTuple to caller. However, PostgreSQL memory management deallocates the memory for that HeapTuple. To prevent reading deallocated memory, we now send copy of found HeapTuple.

It is very similar to what we did [here](https://github.com/citusdata/citus/blob/master/src/backend/distributed/utils/metadata_cache.c#L2301).

metdos:
@ozgune, we could not add a regression test because this is a memory related issue. I and @velioglu  have seen local regression tests to fail before this fix, but Travis tests were passing without any problem.

Issue #1284, opened 2017-03-17
Improve router planner coverage for queries which involve range filter


onderkalaci:
We've recently enabled router planner to support range partitioned tables with #1272 . That led the opportunity of range filter on the query hit to a single shard (i.e., this issue does not apply to hash partitioned tables).

In plain words, the issue is that range filters on the partition keys are not distributed to the joining tables. 

The issue is related to #1204. Without this, we may generate unnecassary worker queries in some cases. 

```SQL

CREATE TABLE events (
	user_id int,
	event_id bigint,
	event_type character varying(255),
	event_time bigint
);
SELECT master_create_distributed_table('events', 'user_id', 'range');

SELECT master_create_empty_shard('events') AS new_shard_id
\gset
UPDATE pg_dist_shard SET shardminvalue = 1, shardmaxvalue = 100
WHERE shardid = :new_shard_id;

SELECT master_create_empty_shard('events') AS new_shard_id
\gset
UPDATE pg_dist_shard SET shardminvalue = 101, shardmaxvalue = 200
WHERE shardid = :new_shard_id;

\COPY events FROM STDIN WITH CSV
10,20001,click,1472807012
10,20002,submit,1472807015
10,20003,pay,1472807020
11,20010,click,1472807022
12,20011,click,1472807023
13,20012,submit,1472807025
15,20013,pay,1472807030
110,20014,click,1472807032
111,20015,click,1472807033
111,20016,click,1472807034
111,20017,submit,1472807035
\.

CREATE TABLE users (
	user_id int,
	lastseen bigint
);
SELECT master_create_distributed_table('users', 'user_id', 'range');

SELECT master_create_empty_shard('users') AS new_shard_id
\gset
UPDATE pg_dist_shard SET shardminvalue = 1, shardmaxvalue = 100
WHERE shardid = :new_shard_id;

SELECT master_create_empty_shard('users') AS new_shard_id
\gset
UPDATE pg_dist_shard SET shardminvalue = 101, shardmaxvalue = 200
WHERE shardid = :new_shard_id;

\COPY users FROM STDIN WITH CSV
10,1472807115
11,1472807215
111,1472807315
\.


-- the query hits single shard
-- and router plannable since the filter on user_id on the users
-- table is equality (users.user_id = 5)
SELECT
		user_id,
		array_agg(event_type) AS events
	FROM
		(SELECT
			users.user_id,
			event_type,
			events.event_time
		FROM
			users,
			events
		WHERE
[...]

Issue #1245, opened 2017-02-20
Get rid of unnecessary calls to RouterSelectTask() for INSERT ... SELECT queries


onderkalaci:
In the current implementation of `INSERT ... SELECT` queries, the following algorithm is followed:

```python

for targetShardInterval in shardIntervalsOfInsertedTable:

     InstantiateQualToTargetShardInterval()
     call RouterSelectTask()
          call ShardPruning() on all shards
```

The above algorithm does not scale for high-shard counts. We should probably implement this more efficient.

One possible approach could be (as prototyped in #1240 ):

```python

prunedShardList = PruneShardList()

for targetShardInterval in shardIntervalsOfInsertedTable:
    
     if targetShardInterval not in prunedShardList:
         continue;
     
    InstantiateQualToTargetShardInterval()   

    call RouterSelectTask() with input shard list as prunedShardList (instead of the whole shards)
          call ShardPruning() with input shard list as prunedShardList (instead of the whole shards)
```

This could dramatically decrease the planning time with high shard counts where the query hits only few shards.

onderkalaci:
Note that we've fixed the issue for `SELECT` queries, but, keeping it open because it is not fixed for `INSERT ... SELECT` queries.

PR #1331, opened 2017-04-18
Faster Shard Pruning Implementation

### LINKED ISSUE #1305: Improve shard pruning performance for deployments with 10K+ shards
When a query comes to the Citus coordinator node, the coordinator applies partition and join pruning logic. A few Citus deployments have over 100K+ shards, and our current partition pruning logic may introduce planning overhead on the order of seconds.

This issue tracks reducing that planning time for Citus deployments that have large shard counts.

anarazel:
Fixes #1305

To fundamentally get better improvements we have to:

1. Add `metadata_cache.c` infrastructure to detect non-overlapping shards.  Simple enough.
2. Split analysis of predicate into a separate step (only executed once for all shards) from the per-shard proof.  This is by far the most complicated part of this change. (cf. `PrunableExpressions()`)
3. For equal restrictions, use binary-search (or direct hash-bucket computation) for hash-partitioned and for range/append partitioned tables without overlapping shard (cf `FastShardPruning()` and `FindShardIntervalIndex()` calls in `PruneOne()`)
4. For `<`, `<=`, `>`, `>=` restrictions on range/append partitioned tables without overlapping shards, use binary search on boundaries. Everything inside the resulting range survives. (cf. `PruneWithBoundaries()`)
5. Implement a faster exhaustive search for all other cases (cf. `ExhaustivePrune()`).

2) is the complex bit here.  Previously we relied on `predicate_refuted_by()` to check for expressions.  As that, as outlined in #1305, is too slow, we have to come up with a different approach.  What I've implemented here is an analysis step (prototyped in `PrunableExpressions()` that builds a list of `PruningInstance`.  Each of these pruning instances is one ANDed list of conditions that can be pruned by, multiple `PruningInstance`s are effectively ORed.

To make pruning fast, each `PruningInstance` has to be fully prunable on its own, without resorting to recursion like `predicate_refuted_by` does. The trick here is to convert all ANDed expressions (be it explicit ANDs or implicitly ANDed ones at the top level) into one `PruningInstance`; and create a separate `PruningInstance` for each ORed expression.   *This is somewhat less powerful* than what the predicate logic does, because it means that expressions like WHERE (composite_key >= 'a' OR composite_key >= 'b') AND (composite_key <= 'c' OR composite_key <= 'd') will prune less than before.  I think this is perfectly OK in practice, because such restrictions seem unlikely to be used successfully in practice, but this definitely requires some though.  We can add a fallback if necessary.

The basic algorithm to build all such `PruningInstance`s is to use a `expression_tree_walker` to iterate over the restrictions. As long as there are no ORs a single `PruningInstance` is built in depth-first manner.  If an OR is encountered (or possibly later a `ScalarArrayOpExpr`), that fact is memorized once for each ORed expression, but no immediate recursion into is performed.  Instead the rest of the expression is processed.   Then, after the tree has fully been processed, all 'partial' `PruningInstance`s are performed.  That's so that e.g. expressions like `WHERE (key >= 'a'  OR key >= 'b') AND (key <= 'c')` result in two Pruning instances (namely `key >= 'a' AND key <'c'` and ' key >= 'b' AND key <'c'`).   If the processing of the expression tree results in one (or more) PruningInstances without a valid restriction (e.g. because there were no restriction on the partition keys, or a type of expression we can't handle yet), `PruneShardList` has to return all shards.

There's a good chunk of work remaining here, but I think the basic approach is pretty sound.  Even when only using the exhaustive search method, I've seen speedups over 10x, with binary-search based 100x over the previous implementations with large numbers (100k+) of shards, more when more shards were used.

### DIFF
--- a/src/backend/distributed/planner/shard_pruning.c
+++ b/src/backend/distributed/planner/shard_pruning.c
@@ -1,0 +1,1319 @@
+/*-------------------------------------------------------------------------
+ *
+ * shard_pruning.c
+ *   Shard pruning related code.
+ *
+ * The goal of shard pruning is to find a minimal (super)set of shards that
+ * need to be queried to find rows matching the expression in a query.
+ *
+ * In PruneShards, we first compute a simplified disjunctive normal form (DNF)
+ * of the expression as a list of pruning instances. Each pruning instance
+ * contains all AND-ed constraints on the partition column. An OR expression
+ * will result in two or more new pruning instances being added for the
+ * subexpressions. The "parent" instance is marked isPartial and ignored
+ * during pruning.
+ *
+ * We use the distributive property for constraints of the form P AND (Q OR R)
+ * to rewrite it to (P AND Q) OR (P AND R) by copying constraints from parent
+ * to "child" pruning instances. However, we do not distribute nested
+ * expressions. While (P OR Q) AND (R OR S) is logically equivalent to (P AND
+ * R) OR (P AND S) OR (Q AND R) OR (Q AND S), in our implementation it becomes
+ * P OR Q OR R OR S. This is acceptable since this will always result in a
+ * superset of shards.  If this proves to be a issue in practice, a more
+ * complete algorithm could be implemented.
+ *
+ * We then evaluate each non-partial pruning instance in the disjunction
+ * through the following, increasingly expensive, steps:
+ *
+ * 1) If there is a constant equality constraint on the partition column, and
+ *    no overlapping shards exist, find the shard interval in which the
+ *    constant falls
+ *
+ * 2) If there is a hash range constraint on the partition column, find the
+ *    shard interval matching the range
+ *
+ * 3) If there are range constraints (e.g. (a > 0 AND a < 10)) on the
+ *    partition column, find the shard intervals that overlap with the range
+ *
+ * 4) If there are overlapping shards, exhaustively search all shards that are
+ *    not excluded by constraints
+ *
+ * Finally, the union of the shards found by each pruning instance is
+ * returned.
+ *
+ * Copyright (c) 2014-2017, Citus Data, Inc.
+ *
+ *-------------------------------------------------------------------------
+ */
+#include "postgres.h"
+
+#include "distributed/shard_pruning.h"
+
+#include "access/nbtree.h"
+#include "catalog/pg_am.h"
+#include "catalog/pg_collation.h"
+#include "catalog/pg_type.h"
+#include "distributed/metadata_cache.h"
+#include "distributed/multi_planner.h"
+#include "distributed/multi_join_order.h"
+#include "distributed/multi_physical_planner.h"
+#include "distributed/shardinterval_utils.h"
+#include "distributed/pg_dist_partition.h"
+#include "distributed/worker_protocol.h"
+#include "nodes/nodeFuncs.h"
+#include "optimizer/clauses.h"
+#include "utils/catcache.h"
+#include "utils/lsyscache.h"
+#include "utils/memutils.h"
+
+/*
+ * A pruning instance is a set of ANDed constraints on a partition key.
+ */
+typedef struct PruningInstance
+{
+	/* Does this instance contain any prunable expressions? */
+	bool hasValidConstraint;
+
+	/*
+	 * This constraint never evaluates to true, i.e. pruning does not have to
+	 * be performed.
+	 */
+	bool evaluatesToFalse;
+
+	/*
+	 * Constraints on the partition column value.  If multiple values are
+	 * found the more restrictive one should be stored here.  Even in case of
+	 * a hash-partitioned table, actual column-values are stored here, *not*
+	 * hashed values.
+	 */
+	Const *lessConsts;
+	Const *lessEqualConsts;
+	Const *equalConsts;
+	Const *greaterEqualConsts;
+	Const *greaterConsts;
+
+	/*
+	 * Constraint using a pre-hashed column value. The constant will store the
+	 * hashed value, not the original value of the restriction.
+	 */
+	Const *hashedEqualConsts;
+
+	/*
+	 * Types of constraints not understood.  We could theoretically try more
+	 * expensive methods of pruning if any such restrictions are found.
+	 *
+	 * TODO: any actual use for this? Right now there seems little point.
+	 */
+	List *otherRestrictions;
+
+	/*
+	 * Has this PruningInstance been added to
+	 * ClauseWalkerContext->pruningInstances? This is not done immediately,
+	 * but the first time a constraint (independent of us being able to handle
+	 * that constraint) is found.
+	 */
+	bool addedToPruningInstances;
+
+	/*
+	 * When OR clauses are found, the non-ORed part (think of a < 3 AND (a > 5
+	 * OR a > 7)) of the expression is stored in one PruningInstance which is
+	 * then copied for the ORed expressions.  The original is marked as
+	 * isPartial, to avoid it being used for pruning.
+	 */
+	bool isPartial;
+} PruningInstance;
+
+
+/*
+ * Partial instances that need to be finished building.  This is used to
+ * collect all ANDed restrictions, before looking into ORed expressions.
+ */
+typedef struct PendingPruningInstance
+{
+	PruningInstance *instance;
+	Node *continueAt;
+} PendingPruningInstance;
+
+
+/*
+ * Data necessary to perform a single PruneShards().
+ */
+typedef struct ClauseWalkerContext
+{
+	Var *partitionColumn;
+	char partitionMethod;
+
+	/* ORed list of pruning targets */
+	List *pruningInstances;
+
+	/*
+	 * Partially built PruningInstances, that need to be completed by doing a
+	 * separate PrunableExpressionsWalker() pass.
+	 */
+	List *pendingInstances;
+
+	/* PruningInstance currently being built, all elegible constraints are added here */
+	PruningInstance *currentPruningInstance;
+
+	/*
+	 * Information about function calls we need to perform. Re-using the same
+	 * FunctionCallInfoData, instead of using FunctionCall2Coll, is often
+	 * cheaper.
+	 */
+	FunctionCallInfoData compareValueFunctionCall;
+	FunctionCallInfoData compareIntervalFunctionCall;
+} ClauseWalkerContext;
+
+static void PrunableExpressions(Node *originalNode, ClauseWalkerContext *context);
+static bool PrunableExpressionsWalker(Node *originalNode, ClauseWalkerContext *context);
+static void AddPartitionKeyRestrictionToInstance(ClauseWalkerContext *context,
+												 OpExpr *opClause, Var *varClause,
+												 Const *constantClause);
+static void AddHashRestrictionToInstance(ClauseWalkerContext *context, OpExpr *opClause,
+										 Var *varClause, Const *constantClause);
+static PruningInstance * CopyPartialPruningInstance(PruningInstance *sourceInstance);
+static List * ShardArrayToList(ShardInterval **shardArray, int length);
+static List * DeepCopyShardIntervalList(List *originalShardIntervalList);
+static int PerformValueCompare(FunctionCallInfoData *compareFunctionCall, Datum a,
+							   Datum b);
+static int PerformCompare(FunctionCallInfoData *compareFunctionCall);
+
+static List * PruneOne(DistTableCacheEntry *cacheEntry, ClauseWalkerContext *context,
+					   PruningInstance *prune);
+static List * PruneWithBoundaries(DistTableCacheEntry *cacheEntry,
+								  ClauseWalkerContext *context,
+								  PruningInstance *prune);
+static List * ExhaustivePrune(DistTableCacheEntry *cacheEntry,
+							  ClauseWalkerContext *context,
+							  PruningInstance *prune);
+static int UpperShardBoundary(Datum partitionColumnValue,
+							  ShardInterval **shardIntervalCache,
+							  int shardCount, FunctionCallInfoData *compareFunction,
+							  bool includeMin);
+static int LowerShardBoundary(Datum partitionColumnValue,
+							  ShardInterval **shardIntervalCache,
+							  int shardCount, FunctionCallInfoData *compareFunction,
+							  bool includeMax);
+
+
+/*
+ * PruneShards returns all shards from a distributed table that cannot be
+ * proven to be eliminated by whereClauseList.
+ *
+ * For reference tables, the function simply returns the single shard that the
+ * table has.
+ */
+List *
+PruneShards(Oid relationId, Index rangeTableId, List *whereClauseList)
+{
+	DistTableCacheEntry *cacheEntry = DistributedTableCacheEntry(relationId);
+	char partitionMethod = cacheEntry->partitionMethod;
+	ClauseWalkerContext context = { 0 };
+	ListCell *pruneCell;
+	List *prunedList = NIL;
+	bool foundRestriction = false;
+
+	/* always return empty result if WHERE clause is of the form: false (AND ..) */
+	if (ContainsFalseClause(whereClauseList))
+	{
+		return NIL;
+	}
+
+	/* short circuit for reference tables */
+	if (partitionMethod == DISTRIBUTE_BY_NONE)
+	{
+		prunedList = ShardArrayToList(cacheEntry->sortedShardIntervalArray,
+									  cacheEntry->shardIntervalArrayLength);
+		return DeepCopyShardIntervalList(prunedList);
+	}
+
+
+	context.partitionMethod = partitionMethod;
+	context.partitionColumn = PartitionColumn(relationId, rangeTableId);
+	context.currentPruningInstance = palloc0(sizeof(PruningInstance));
+
+	if (cacheEntry->shardIntervalCompareFunction)
+	{
+		/* initiate function call info once (allows comparators to cache metadata) */
+		InitFunctionCallInfoData(context.compareIntervalFunctionCall,
+								 cacheEntry->shardIntervalCompareFunction,
+								 2, DEFAULT_COLLATION_OID, NULL, NULL);
+	}
+	else
+	{
+		ereport(ERROR, (errmsg("shard pruning not possible without "
+							   "a shard interval comparator")));
+	}
+
+	if (cacheEntry->shardColumnCompareFunction)
+	{
+		/* initiate function call info once (allows comparators to cache metadata) */
+		InitFunctionCallInfoData(context.compareValueFunctionCall,
+								 cacheEntry->shardColumnCompareFunction,
+								 2, DEFAULT_COLLATION_OID, NULL, NULL);
+	}
+	else
+	{
+		ereport(ERROR, (errmsg("shard pruning not possible without "
+							   "a partition column comparator")));
+	}
+
+	/* Figure out what we can prune on */
+	PrunableExpressions((Node *) whereClauseList, &context);
+
+	/*
+	 * Prune using each of the PrunableInstances we found, and OR results
+	 * together.
+	 */
+	foreach(pruneCell, context.pruningInstances)
+	{
+		PruningInstance *prune = (PruningInstance *) lfirst(pruneCell);
+		List *pruneOneList;
+
+		/*
+		 * If this is a partial instance, a fully built one has also been
+		 * added. Skip.
+		 */
+		if (prune->isPartial)
+		{
+			continue;
+		}
+
+		/*
+		 * If the current instance has no prunable expressions, we'll have to
+		 * return all shards.  No point in continuing pruning in that case.
+		 */
+		if (!prune->hasValidConstraint)
+		{
+			foundRestriction = false;
+			break;
+		}
+
+		/*
+		 * Similar to the above, if hash-partitioned and there's nothing to
+		 * prune by, we're done.
+		 */
+		if (context.partitionMethod == DISTRIBUTE_BY_HASH &&
+			!prune->evaluatesToFalse && !prune->equalConsts && !prune->hashedEqualConsts)
+		{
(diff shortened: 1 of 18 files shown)


anarazel:
@marcocitus I'd welcome some input about the basic approach, especially the behavioural change of being able to prune a bit less.

@sumedhpathak You'll have to assign reviewers and future owners of the topic ;)

anarazel:
With customer's range-partitioned usecase, comparing the plan (not execution) times. 18  queries fail before & after because we lack the details of a few functions.

Before:
```
Time: 66.156 ms
postgres[20250][1]=# SELECT time_explain IS NULL AS executable, count(*), SUM(time_explain), AVG(time_explain) FROM timing_before GROUP BY 1; 
┌────────────┬───────┬────────────────┬─────────────────┐
│ executable │ count │      sum       │       avg       │
├────────────┼───────┼────────────────┼─────────────────┤
│ f          │    73 │ 00:01:46.61029 │ 00:00:01.460415 │
│ t          │    18 │ (null)         │ (null)          │
└────────────┴───────┴────────────────┴─────────────────┘
(2 rows)


```
After
```
postgres[20250][1]=# SELECT time_explain IS NULL AS executable, count(*), SUM(time_explain), AVG(time_explain) FROM timing_after GROUP BY 1;
┌────────────┬───────┬─────────────────┬────────────────┐
│ executable │ count │       sum       │      avg       │
├────────────┼───────┼─────────────────┼────────────────┤
│ f          │    73 │ 00:00:02.973268 │ 00:00:00.04073 │
│ t          │    18 │ (null)          │ (null)         │
└────────────┴───────┴─────────────────┴────────────────┘
(2 rows)
```

with exhaustive pruning instead of doing binary search:
```
┌────────────┬───────┬─────────────────┬────────────────┐
│ executable │ count │       sum       │      avg       │
├────────────┼───────┼─────────────────┼────────────────┤
│ f          │    73 │ 00:00:20.872922 │ 00:00:00.28593 │
│ t          │    18 │ (null)          │ (null)         │
└────────────┴───────┴─────────────────┴────────────────┘
```

onderkalaci:
I'm doing some benchmarks on `INSERT ... SELECT` with 32K shards on c4.4xlarge. Although the planner iterates on each shard and calls `RouterSelectTask()` (which calls `PruneShardList()`), the results look really impressive:

(Note that we only measure time passed during `CreateInsertSelectRouterPlan()`)

`insert into test_table select * from test_table;` -> `849 msec`
`insert into test_table select * from test_table where a = 5;` -> `310 msec`
`insert into test_table select * from test_table where a = 5 OR a = 6 OR a = 7;` -> `443 msec`
`insert into test_table select * from test_table where a != 5;` -> `794 msec`

Since we're going to replace the logic to generate subqueries with the approach we have for `Insert .. SELECT`, I'd expect to see som increase in the total planning time unless we fix #1245 (as far as I can see).

Any comments for the above?

Btw, could we fix #1284 within this PR?

anarazel:
@onderkalaci Those are timing post the patch?  What are the pre-patch values like?

> Btw, could we fix #1284 within this PR?

Ad-hoc I'd say that's an unrelated issue, but I've not looked into it much.

onderkalaci:
> @onderkalaci Those are timing post the patch? What are the pre-patch values like?

Those are the post patch. The pre-patch values are so high that (`> 10 minutes` ), I didn't try to measure them.

src/backend/distributed/utils/shard_pruning.c
```diff
@@ -0,0 +1,1065 @@
...
+	}
+
+	/*
+	 * For hashing, there's no further pruning methods available.  Because
+	 * triangle inequality doesn't hold for hashed values, we can't prune
+	 * based on < and >.
+	 */
+	if (context->partitionMethod == DISTRIBUTE_BY_HASH)
+	{
+		return NIL;
```
marcocitus: Shouldn't this return all shards?

It seems like we could only get here if we have overlapping shards in a hash-partitioned table, might be worth eliminating that case in `PruneShardList` and adding an assert here.

anarazel:
On 2017-04-26 14:46:05 -0700, Marco Slot wrote:
> 
> 
> 
> > +	 * better to pre-compute that by checking a PruningInstance's
> +	 * less[Equal]Consts/greater[Equal]Consts against equalConsts.
> +	 */
> +	if (shardInterval)
> +	{
> +		return list_make1(shardInterval);
> +	}
> +
> +	/*
> +	 * For hashing, there's no further pruning methods available.  Because
> +	 * triangle inequality doesn't hold for hashed values, we can't prune
> +	 * based on < and >.
> +	 */
> +	if (context->partitionMethod == DISTRIBUTE_BY_HASH)
> +	{
> +		return NIL;
> 
> Shouldn't this return all shards?
> 
> It seems like we could only get here if we have overlapping shards in a hash-partitioned table, might be worth eliminating that case in `PruneShardList` and adding an assert here.

Right, I turned this into an Assert already.  Don't look a the code in
this PR too much yet (hence the Prototype:), I'm about to push a
noticeably cleaned up version, that also fixes a few bugs.

anarazel:
@marcocitus Ok, I think this is now in a reasonably reviewable state (thus removed Prototype: from title)  For now contains changes to-be-merged separately in #1355, and will be simplified somewhat by #1353.

src/backend/distributed/utils/shard_pruning.c
```diff
@@ -0,0 +1,1239 @@
...
+	if (cacheEntry->shardIntervalCompareFunction)
+	{
+		/* initiate function call info once (allows comparators to cache metadata) */
+		InitFunctionCallInfoData(context.compareIntervalFunctionCall,
+								 cacheEntry->shardIntervalCompareFunction,
+								 2, DEFAULT_COLLATION_OID, NULL, NULL);
+	}
+	else
+	{
+		ereport(ERROR, (errmsg("frak1")));
```
anarazel: Err, this obviously should have a smarter error message - Or actually probably just an assert.

src/test/regress/expected/multi_prune_shard_list.out
```diff
@@ -69,21 +69,21 @@ SELECT prune_using_single_value('pruning', NULL);
...
 (1 row)
 
--- but if both values are on the same shard, should get back that shard
+-- even if both values are on the same shard, a value can't be equal to two others
 SELECT prune_using_both_values('pruning', 'tomato', 'rose');
  prune_using_both_values 
 -------------------------
- {800002}
+ {}
 (1 row)
```
anarazel: This deserves a look - for me this is quite obviously correct, but input would be welcome.

src/backend/distributed/utils/shard_pruning.c
```diff
@@ -0,0 +1,1245 @@
...
+ * are pruned upon, and builds a list of ORed constraints for pruning. This
+ * happens in PrunableExpressions().  The reason for keeping this separate
+ * from actually pruning away shards is that it has to only be done once for
+ * all shards.
+ *
+ * In the second phase, the shards are pruned using the constraints built in
+ * the first phase.  Each constraint (a set of ANDed constraints) is evaluated
+ * by PruneOne(), and the results of all results is then ORed/unioned
+ * together.  Depending on the types of constraints found, and the type of
+ * partitioning used, different methods of pruning are employed.  Ranging from
```
marcocitus: It would be nice if this documentation was a bit more explicit about what we're doing.

The goal of shard pruning is to find a minimal (super)set of shards that need to be queried to find rows matching the expression in a query.

In `PruneShards`, we first compute a simplified disjunctive normal form (DNF) of the expression as a list of pruning instances. Each pruning instance contains all AND-ed constraints on the partition column. An OR expression will result in two or more new pruning instances for the subexpression.

We use the distributive property for constraints of the form P AND (Q OR R) to rewrite it to (P AND Q) OR (P AND R). However, we do not distribute nested expressions. While (P OR Q) AND (R OR S) is logically equivalent to (P AND R) OR (P AND S) OR (Q AND R) OR (Q AND S), in our implementation it becomes P OR Q OR R OR S. This is acceptable since this will always result in a superset.

We then evaluate each pruning instance in the disjunction through the following steps:

1. If there is a constant equality constraint on the partition column, find the shard interval in which the constant falls
2. If there is a hash range constraint on the partition column, find the shard interval matching the range
3. If there are range constraints on the partition column, find the shard intervals that overlap with the range
4. If there are overlapping shards, exhaustively search all shards that are not excluded by constraints

Finally, we take the union of the shards found by each pruning instance.

src/backend/distributed/utils/shard_pruning.c
```diff
@@ -0,0 +1,1245 @@
...
+	while (context->pendingInstances != NIL)
+	{
+		PendingPruningInstance *instance =
+			(PendingPruningInstance *) linitial(context->pendingInstances);
+		PruningInstance *oldPrune = instance->instance;
+		PruningInstance *newPrune = palloc0(sizeof(PruningInstance));
+
+		if (oldPrune)
+		{
+			memcpy(newPrune, oldPrune, sizeof(PruningInstance));
```
marcocitus: This probably deserves some more explanation. e.g. implement the distributive property: (P AND (Q || R)) `oldPrune` is a partial pruning instance which contains `P` which is copied into two new pruning instances, one for `P AND Q` and one for `P AND R`.

src/backend/distributed/utils/shard_pruning.c
```diff
@@ -0,0 +1,1239 @@
...
+	 * Constraint using a pre-hashed column value. The constant will store the
+	 * hashed value, not the original value of the restriction.
+	 */
+	Const *hashedEqualConsts;
+
+	/*
+	 * Types of constraints not understood.  We could theoretically try more
+	 * expensive methods of pruning if any such restrictions are found.
+	 *
+	 * TODO: any actual use for this? Right now there seems little point.
```
anarazel: Right - I was more thinking whether we should keep the field for the moment, given it's not used atm.

src/backend/distributed/utils/shard_pruning.c
```diff
@@ -0,0 +1,1239 @@
...
+typedef struct ClauseWalkerContext
+{
+	Var *partitionColumn;
+	char partitionMethod;
+
+	/* ORed list of pruning targets */
+	List *pruningInstances;
+
+	/*
+	 * Partially built PruningInstances, that need to be completed. XXX
```
anarazel: Think I wanted to expand the comment a bit, or rather reference a comment explaining why partial instances are needed.

src/backend/distributed/utils/shard_pruning.c
```diff
@@ -0,0 +1,1239 @@
...
+						if (!prune->equalConsts)
+						{
+							prune->equalConsts = constantClause;
+						}
+						else if (PerformValueCompare(&context->compareValueFunctionCall,
+													 constantClause->constvalue,
+													 prune->equalConsts->constvalue) != 0)
+						{
+							/* key can't be equal to two values */
+							prune->alwaysFalse = true;
```
anarazel: We're checking for comparator() != 0 above?  That precludes such a problem, right?

src/backend/distributed/utils/shard_pruning.c
```diff
@@ -0,0 +1,1239 @@
...
+
+	Assert(shardCount != 0);
+
+	/* setup partitionColumnValue argument once */
+	compareFunction->arg[0] = partitionColumnValue;
+	compareFunction->argnull[0] = false;
+
+	while (lowerBoundIndex < upperBoundIndex)
+	{
+		int middleIndex = (lowerBoundIndex + upperBoundIndex) / 2; /* XXX: overflow? */
```
anarazel: Why?  Practically it'll currently not because you'll grow very bored before creating close to enough shards, but if upperBoundIndex is something like 2**30, and lowerBoundIndex is increased to 2**29 by one probe, you'll get an overflow, no?

src/backend/distributed/utils/shard_pruning.c
```diff
@@ -0,0 +1,1239 @@
...
+		shardIndex = FindShardIntervalIndex(prune->hashedEqualConsts->constvalue,
+											sortedShardIntervalArray,
+											cacheEntry->shardIntervalArrayLength,
+											context->partitionMethod,
+											cacheEntry->shardIntervalCompareFunction,
+											cacheEntry->hasUniformHashDistribution);
+
+		if (shardIndex == INVALID_SHARD_INDEX)
+		{
+			return NIL;
```
anarazel: 0 shards is the only case I think.

src/backend/distributed/utils/shard_pruning.c
```diff
@@ -0,0 +1,1239 @@
...
+	 * less[Equal]Consts/greater[Equal]Consts against equalConsts.
+	 */
+	if (shardInterval)
+	{
+		return list_make1(shardInterval);
+	}
+
+	/*
+	 * Should never get here for hashing, either we've filtered down to zero
+	 * or one shard, and returned.
```
anarazel: Hm?  That shouldn't be possible for hash, right?

marcocitus: I agree, but why do we call `HasOverlappingShardInterval` then?

(we should just get rid of that function)

src/backend/distributed/utils/shard_pruning.c
```diff
@@ -0,0 +1,1239 @@
...
+	/* always return empty result if WHERE clause is of the form: false (AND ..) */
+	if (ContainsFalseClause(whereClauseList))
+	{
+		return NIL;
+	}
+
+	/* short circuit for reference tables */
+	if (partitionMethod == DISTRIBUTE_BY_NONE)
+	{
+		return ShardArrayToList(cacheEntry->sortedShardIntervalArray,
```
anarazel: Hm. You have a point with the deep copy - but I think it'd make more sense to do that one layer up.  Otherwise we'll have to do a lot more work when ORing two PruneOne results.

src/backend/distributed/utils/shard_pruning.c
```diff
@@ -0,0 +1,1239 @@
...
+	 * added to the list, so don't use foreach().
+	 */
+	while (context->pendingInstances != NIL)
+	{
+		PendingPruningInstance *instance =
+			(PendingPruningInstance *) linitial(context->pendingInstances);
+		PruningInstance *oldPrune = instance->instance;
+		PruningInstance *newPrune = palloc0(sizeof(PruningInstance));
+
+		if (oldPrune)
```
anarazel: Never these days... Removed.

src/backend/distributed/utils/shard_pruning.c
anarazel: Hope you don't mind cribbing some of that directly into the header...

src/backend/distributed/utils/shard_pruning.c
marcocitus: A wrapper could work, or making a copy before returning.

src/backend/distributed/utils/shard_pruning.c
```diff
@@ -0,0 +1,1239 @@
...
+#include "utils/lsyscache.h"
+#include "utils/memutils.h"
+
+/*
+ * A pruning instance is a set of ANDed constraints on a partition key.
+ */
+typedef struct PruningInstance
+{
+	/*
+	 * This constraint never evaluates to false, i.e. pruning does not have to
```
anarazel: done (next push)

src/backend/distributed/utils/shard_pruning.c
```diff
@@ -0,0 +1,1239 @@
...
+	 * be performed.
+	 */
+	bool alwaysFalse;
```
anarazel: done (next push)

src/backend/distributed/utils/shard_pruning.c
```diff
@@ -0,0 +1,1239 @@
...
+	 */
+	List *otherRestrictions;
+
+	/*
+	 * Has this PruningInstance been added to
+	 * ClauseWalkerContext->pruningInstances? This is not done immediately,
+	 * but the first time a constraint (independent of us being able to handle
+	 * that constraint) is found.
+	 */
+	bool added;
```
anarazel: done (next push)

src/backend/distributed/utils/shard_pruning.c
```diff
@@ -0,0 +1,1239 @@
...
+	 * OR a > 7)) of the expression is stored in one PruningInstance which is
+	 * then copied for the ORed expressions.  The original is marked as
+	 * isPartial, to avoid it being used for pruning.
+	 */
+	bool isPartial;
+
+	/*
+	 * Does this instance contain any prunable expressions.
+	 */
+	bool hasValidConstraint;
```
anarazel: done (next push)

src/backend/distributed/utils/shard_pruning.c
```diff
@@ -0,0 +1,1239 @@
...
+	}
+
+	/* found no valid restriction, build list of all shards */
+	if (!foundRestriction)
+	{
+		return ShardArrayToList(cacheEntry->sortedShardIntervalArray,
+								cacheEntry->shardIntervalArrayLength);
+	}
+
+	return prunedList;
```
anarazel: Done (next push) in PruneShards(), not PruneOne().

src/backend/distributed/utils/shard_pruning.c
```diff
@@ -0,0 +1,1239 @@
...
+			/*
+			 * Found a restriction on the partition column itself. Update the
+			 * current constraint with the new information.
+			 */
+
+			btreeIntepretationList =
+				get_op_btree_interpretation(operatorExpression->opno);
+			foreach(btreeInterpretationCell, btreeIntepretationList)
+			{
+				OpBtreeInterpretation *btreeIntepretation =
```
anarazel: done (next push)

src/backend/distributed/utils/shard_pruning.c
anarazel: Added something along those lines, although I split it between PrunableExpressions() and the relevant bit in PrunableExpressionsWalker().

src/backend/distributed/utils/shard_pruning.c
```diff
@@ -0,0 +1,1239 @@
...
+					(OpBtreeInterpretation *) lfirst(btreeInterpretationCell);
+
+				switch (btreeIntepretation->strategy)
```
anarazel: I tried this, and it imo doesn't seem to come out easier to understand.  Unless you feel strongly, I'd rather not go for it.

src/test/regress/expected/multi_prune_shard_list.out
marcocitus: Seems fine.

src/backend/distributed/planner/shard_pruning.c
```diff
@@ -0,0 +1,1267 @@
...
+	 * added to the list, so don't use foreach().
+	 */
+	while (context->pendingInstances != NIL)
+	{
+		PendingPruningInstance *instance =
+			(PendingPruningInstance *) linitial(context->pendingInstances);
+		PruningInstance *oldPrune = instance->instance;
+		PruningInstance *newPrune = palloc0(sizeof(PruningInstance));
+
+
```
marcocitus: Extra newline

src/backend/distributed/planner/shard_pruning.c
```diff
@@ -0,0 +1,1267 @@
...
+		 * context->pendingInstances why construction of the PruningInstance
+		 * was deferred.
+		 *
+		 * To make the new PruningInstance useful for pruning, we have to
+		 * unset it being partial - if necessary it'll be marked so again by
+		 * PrunableExpressionsWalker().
+		 */
+		memcpy(newPrune, oldPrune, sizeof(PruningInstance));
+		newPrune->addedToPruningInstances = false;
+		newPrune->isPartial = false;
```
marcocitus: For readability, it might makes sense to move the 4 lines above to something like an `InitializeWithSameConstraints` function.

src/backend/distributed/planner/shard_pruning.c
```diff
@@ -0,0 +1,1267 @@
...
+			 */
+
+			btreeInterpretationList =
+				get_op_btree_interpretation(operatorExpression->opno);
+			foreach(btreeInterpretationCell, btreeInterpretationList)
+			{
+				OpBtreeInterpretation *btreeInterpretation =
+					(OpBtreeInterpretation *) lfirst(btreeInterpretationCell);
+
+				switch (btreeInterpretation->strategy)
```
marcocitus: It would be nicer to move this block into a separate function (see previous comment).

src/backend/distributed/planner/shard_pruning.c
```diff
@@ -0,0 +1,1267 @@
...
+		{
+			return list_make1(sortedShardIntervalArray[shardIndex]);
+		}
+	}
+
+	/*
+	 * FIXME: We could perform ExhaustivePrune() on the individual
+	 * shardInterval, in case there's both an = and a < / >? It'd probably be
+	 * better to pre-compute that by checking a PruningInstance's
+	 * less[Equal]Consts/greater[Equal]Consts against equalConsts.
```
marcocitus: This comment doesn't help me understand why we return below. Maybe we can say "We found a shard based on an equality constraint. We might be able to prune further if there are also range constraints, but this is atypical and we can save on planning cost by returning earlier."

src/backend/distributed/planner/shard_pruning.c
anarazel: Done, with a different name though.

src/backend/distributed/planner/shard_pruning.c
anarazel: Done, into two different functions for the magic col, and plain constraint cases.

src/backend/distributed/planner/shard_pruning.c
anarazel: Better?

anarazel:
@marcocitus Thanks for the careful review!

sumedhpathak:
Thanks @anarazel @marcocitus !!
🍾

PR #1332, opened 2017-04-18
Set default value of isactive to true

### LINKED ISSUE #1330: citus--6.2-2--6.2-3.sql upgrade breaks existing setups by marking all nodes inactive
Because the update script simply does
```
ALTER TABLE pg_dist_node ADD isactive bool;
```
the `isactive` column will be set to NULL for all pre-existing rows, which `TupleToWorkerNode()` will blithely ignore (because HeapTupleHasNulls() only checks for existing columns, not for ones later added), which'll end up with `workerNode->isActive` being set to false.

byucesoy:
Fixes #1330 

With this change, we set to default value of isactive column to true so that upgrading users' all nodes will be marked as active to not break their environment.

### DIFF
--- a/src/backend/distributed/citus--6.2-2--6.2-3.sql
+++ b/src/backend/distributed/citus--6.2-2--6.2-3.sql
@@ -2,7 +2,7 @@
 
 SET search_path = 'pg_catalog';
 
-ALTER TABLE pg_dist_node ADD isactive bool;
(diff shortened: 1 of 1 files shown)


byucesoy:
Yes, I also manually tested upgrading from 6.1 to 6.2-3 and from 6.2-2 to 6.2-3.

PR #1333, opened 2017-04-18
Evaluate expressions on the LIMIT clause for subquery pushdown

onderkalaci:
Subquery pushdown uses orignal query, the LIMIT and OFFSET clauses
are not evaluated. However, logical optimizer expects these expressions
are already evaluated by the standard planner. This commit manually
evaluates the functions on the logical planner for subquery pushdown.

### DIFF
--- a/src/backend/distributed/planner/multi_logical_planner.c
+++ b/src/backend/distributed/planner/multi_logical_planner.c
@@ -19,6 +19,7 @@
 #include "catalog/pg_am.h"
 #include "catalog/pg_class.h"
 #include "commands/defrem.h"
+#include "distributed/citus_clauses.h"
 #include "distributed/colocation_utils.h"
 #include "distributed/metadata_cache.h"
 #include "distributed/multi_logical_optimizer.h"
@@ -2991,6 +2992,17 @@ SubqueryPushdownMultiPlanTree(Query *queryTree)
 		extendedOpNode->havingQual =
 			(Node *) make_ands_implicit((Expr *) extendedOpNode->havingQual);
 	}
+
+	/*
+	 * Postgres standard planner evaluates expressions in the LIMIT/OFFSET clauses.
+	 * Since we're using original query here, we should manually evaluate the
+	 * expression on the LIMIT and OFFSET clauses. Note that logical optimizer
+	 * expects those clauses to be already evaluated.
+	 */
+	extendedOpNode->limitCount =
+		PartiallyEvaluateExpression(extendedOpNode->limitCount, NULL);
+	extendedOpNode->limitOffset =
+		PartiallyEvaluateExpression(extendedOpNode->limitOffset, NULL);
 
 	SetChild((MultiUnaryNode *) extendedOpNode, currentTopNode);
 	currentTopNode = (MultiNode *) extendedOpNode;
--- a/src/backend/distributed/utils/citus_clauses.c
+++ b/src/backend/distributed/utils/citus_clauses.c
@@ -31,7 +31,6 @@ typedef struct FunctionEvaluationContext
 
 
 /* private function declarations */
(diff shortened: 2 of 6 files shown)


src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -2992,6 +2993,30 @@ SubqueryPushdownMultiPlanTree(Query *queryTree)
...
+		PartiallyEvaluateExpression(extendedOpNode->limitOffset, NULL);
+
+	/*
+	 * If either LIMIT or OFFSET contains a parameter, PartiallyEvaluateExpression()
+	 * fails to evaluate them and we should error out sanely.
+	 */
+	if ((extendedOpNode->limitCount && !IsA(extendedOpNode->limitCount, Const)) ||
+		(extendedOpNode->limitOffset && !IsA(extendedOpNode->limitOffset, Const)))
+	{
+		ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
```
mtuncer: I do not feel comfortable having this error message here. But I admit that there is probably no good place for this either. Perhaps we could put the error message to where we actually require them to be of type Const.

mtuncer: While on that, we should have separate limit clause and limit offset errors, and error message could be re-written like
"unsupported limit expression" or "unsupported offset expression". We do not need to write a whole big sentence.

src/backend/distributed/utils/citus_clauses.c
```diff
@@ -162,7 +161,7 @@ ExecuteMasterEvaluableFunctions(Query *query, PlanState *planState)
  * Walks the expression evaluating any node which invokes a function as long as a Var
  * doesn't show up in the parameter list.
  */
-static Node *
+Node *
```
mtuncer: Please check this with @marcocitus

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: >  Perhaps we could put the error message to where we actually require them to be of type Const.

That's a good idea

src/backend/distributed/utils/citus_clauses.c
onderkalaci: I briefly asked him and his response was: `yes, if there are no parameters`
That's why I added the parametrized test.

src/test/regress/sql/multi_subquery_behavioral_analytics.sql
```diff
@@ -1345,4 +1345,123 @@ FROM (SELECT
...
+    SELECT u.user_id, e.event_type::text AS event, e.time
+    FROM users_table AS u,
+         events_table AS e
+    WHERE u.user_id = e.user_id
+      AND e.event_type IN (100, 101, 102)
+  ) t
+  GROUP BY user_id
+) q
+ORDER BY 2 DESC, 1
+LIMIT volatile_func_test() + (ROW(1,2,NULL) < ROW(1,3,0))::int OFFSET volatile_func_test() + volatile_func_test();
```
mtuncer: this is kind of extreme but should be fine

onderkalaci:
@mtuncer I applied your feedback, anything remaining for shipt?

src/test/regress/expected/multi_subquery_behavioral_analytics.out
```diff
@@ -1672,4 +1672,147 @@ WHERE b.user_id IS NULL
 GROUP BY a.user_id;  
 ERROR:  cannot push down this subquery
 DETAIL:  Table expressions other than simple relations and subqueries are currently unsupported
+-- same query withou LIMIT/OFFSET returns 30 row
```
mtuncer: s/withou/without

PR #1334, opened 2017-04-18
master_citus_tools.c: PVS-Studio: CWE-476 (NULL Pointer Dereference)

SvyatoslavRazmyslov:
We have found and fixed a weakness using PVS-Studio tool. PVS-Studio is a static code analyzer for C, C++ and C#: https://www.viva64.com/en/pvs-studio/

We suggests having a look at the emails, sent from @pvs-studio.com. 

Analyzer warning: [V595](https://www.viva64.com/en/w/V595/) The 'errorMessage' pointer was utilized before it was verified against nullptr. Check lines: 438, 447. master_citus_tools.c 438


marcocitus:
Thanks, this is fixed in master.

PR #1335, opened 2017-04-18
colocation_utils.c: PVS-Studio: underflow of the buffer 'values'

SvyatoslavRazmyslov:
We have found and fixed a bug using PVS-Studio tool. PVS-Studio is a static code analyzer for C, C++ and C#: https://www.viva64.com/en/pvs-studio/

We suggests having a look at the emails, sent from @pvs-studio.com.

Analyzer warning: [V512](https://www.viva64.com/en/w/V512/) A call of the 'memset' function will lead to underflow of the buffer 'values'. colocation_utils.c 641


marcocitus:
Thanks, fixed this in master.

PR #1340, opened 2017-04-19
Better format subquery regression tests

onderkalaci:
Only some formatting changes, no code-reviewer needed.

This PR does not make the regression tests look `perfect`. But at least they are now more consistent and readable. @metdos Feel free to have a look at the changes.


onderkalaci:
I only removed a single regression test given that it was exactly the same with the test that is above it.

PR #1341, opened 2017-04-19
Require subquery_pushdown when limit is used in subquery

mtuncer:
Using limit in subqueries may cause returning incorrect results. Therefore we allow limits in subqueries only if user explicitly set subquery_pushdown flag.

In short:
- limits in subqueries is not supported by default
- setting subquery_pushdown flag enables limit
- if a query contains a limit clause anywhere in inner queries, then it must have limit clause in outermost query

### DIFF
--- a/src/backend/distributed/planner/multi_logical_planner.c
+++ b/src/backend/distributed/planner/multi_logical_planner.c
@@ -71,9 +71,11 @@ static DeferredErrorMessage * DeferErrorIfUnsupportedFilters(Query *subquery);
 static bool EqualOpExpressionLists(List *firstOpExpressionList,
 								   List *secondOpExpressionList);
 static DeferredErrorMessage * DeferErrorIfCannotPushdownSubquery(Query *subqueryTree,
-																 bool outerQueryHasLimit);
+																 bool
+																 outerMostQueryHasLimit);
 static DeferredErrorMessage * DeferErrorIfUnsupportedUnionQuery(Query *queryTree,
-																bool outerQueryHasLimit);
+																bool
+																outerMostQueryHasLimit);
 static bool ExtractSetOperationStatmentWalker(Node *node, List **setOperationList);
 static DeferredErrorMessage * DeferErrorIfUnsupportedTableCombination(Query *queryTree);
 static bool TargetListOnPartitionColumn(Query *query, List *targetEntryList);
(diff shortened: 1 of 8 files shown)


src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -571,16 +573,26 @@ DeferErrorIfCannotPushdownSubquery(Query *subqueryTree, bool outerQueryHasLimit)
 		errorDetail = "Offset clause is currently unsupported";
 	}
 
-	if (subqueryTree->limitCount && !outerQueryHasLimit)
+	/*
+	 * Limit is partially supported when SubqueryPushdown is set.
+	 * The outermost query must have a limit clause.
+	 */
+	if (subqueryTree->limitCount && !SubqueryPushdown)
```
onderkalaci: Hmm, shall we add a `errhnt` for this case? What do you think?

onderkalaci: Also, in the master branch, we error out `LIMIT`s whenever there is no LIMIT on the outside query. But, here we skip that if the flag is set?

Is that an informed decision? Thinking about, this change makes more sense to me. What are your thoughtson this?

mtuncer: > Hmm, shall we add a errhnt for this case? What do you think?

errhint would require exposing `citus.subquery_pushdown`. I think we never publicly mentioned that so far.

mtuncer: > Also, in the master branch, we error out LIMITs whenever there is no LIMIT on the outside query. But, here we skip that if the flag is set?

I did not understand this comment.  here we check
1 -  inner query has limit and subquery_pushdown is not set ---> error
2 - inner query has limit, pushdown_flag is set (implicit due to else if), and outermost query does not have limit ---> error

onderkalaci: > implicit due to else if

I think that confused me a bit. Can't we make each as a separate `if` statements as we have throughout the function?

```C
	if (subqueryTree->limitCount && !SubqueryPushdown)
	{
		errorDetail = "Limit in subquery is currently unsupported";
	}
	 
    if (subqueryTree->limitCount && SubqueryPushdown && !outerMostQueryHasLimit)
	{
		errorDetail = "Limit in subquery without limit in the outermost query is "
					  "unsupported";
	}


```

PR #1342, opened 2017-04-20
Subqueries in WHERE clause

onderkalaci:
Support for subqueries in WHERE clause
    
    This commit enables subqueries in WHERE clause to be pushed down
    by the subquery pushdown logic.
    
    The support covers:
      - Correlated subqueries with IN, NOT IN, EXISTS, NOT EXISTS,
        operator expressions such as (>, <, =, ALL, ANY etc.)
      - Non-correlated subqueries with (partition_key) IN (SELECT partition_key ..)
        (partition_key) =ANY (SELECT partition_key ...)
    
    Note that this commit heavily utilizes the attribute equivalence logic introduced
    in the 1cb6a34ba8430fb2db3c481ac110923cecc0d12a. In general, this commit mostly
    adjusts the logical planner not to error out on the subqueries in WHERE clause.


src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -2199,16 +2304,6 @@ IsSelectClause(Node *clause)
 	Var *firstColumn = NULL;
 	Index firstColumnTableId = 0;
 	bool isSelectClause = true;
-	NodeTag nodeTag = nodeTag(clause);
```
onderkalaci: Ops, we should add these lines back. It was remaining from some of the tests.

onderkalaci: I moved these checks to a separate function given that I don't think it is meaningful to error out for sublinks here. Does that work?

onderkalaci:
The PR contains two separate commits for ease of the reviewer. The first PR enables subqueries in where clause and the second PR applies the error checks on those subqueries.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -2947,8 +2947,6 @@ CopyRelationRestrictionContext(RelationRestrictionContext *oldContext)
 
 		/* not copyable, but readonly */
 		newRestriction->plannerInfo = oldRestriction->plannerInfo;
-		newRestriction->prunedShardIntervalList =
```
onderkalaci: This might go to another commit.

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -2993,11 +2993,10 @@ ExtractQueryWalker(Node *node, List **queryList)
 		Query *query = (Query *) node;
 
 		(*queryList) = lappend(*queryList, query);
-		walkerResult = query_tree_walker(query, ExtractQueryWalker, queryList,
-										 QTW_EXAMINE_RTES);
+		return query_tree_walker(query, ExtractQueryWalker, queryList, 0);
```
onderkalaci: Shall we add `QTW_IGNORE_CTE_SUBQUERIES`,

metdos: Can we hit here for CTEs? We should already error for CTEs, right?

metdos: Do we want to drop QTW_EXAMINE_RTES here?

onderkalaci: We're not dropping that. Instead, if there are no flag (i.e., 0), the walker traverses every part of the input query. The change is required to traverse where clause in addition to the rtable.

onderkalaci: No, we can't hit. We already error out. So, let's keep this as it is?

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -676,41 +746,20 @@ DeferErrorIfCannotPushdownSubquery(Query *subqueryTree, bool outerMostQueryHasLi
 							 errorDetail, NULL);
 	}
 
-	/* recursively do same check for subqueries of this query */
```
metdos: If we make this function non-recursive, we should also update the function comments.

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1248,11 +1279,11 @@ ErrorIfQueryNotSupported(Query *queryTree)
 	const char *filterHint = "Consider using an equality filter on the distributed "
 							 "table's partition column.";
 
-	if (queryTree->hasSubLinks)
+	if (queryTree->hasSubLinks && SublinkList(queryTree) == NIL)
 	{
 		preconditionsSatisfied = false;
 		errorMessage = "could not run distributed query with subquery outside the "
-					   "FROM clause";
+					   "FROM and WHERE clauses";
```
metdos: Where can be subqueries outside of FROM and WHERE clauses? Is this only target list?

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1752,11 +1789,37 @@ ValidateClauseList(List *clauseList)
 	{
 		Node *clause = (Node *) lfirst(clauseCell);
 
-		bool selectClause = IsSelectClause(clause);
-		bool joinClause = IsJoinClause(clause);
-		bool orClause = or_clause(clause);
+		/* produce a better error message for sublinks */
```
metdos: Is this a TODO?

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1752,11 +1789,37 @@ ValidateClauseList(List *clauseList)
...
+		ErrorIfSublink(clause);
```
metdos: Do we need this function? We can just do
```c
if (IsSublinkClause(clause))
{
error
}
```

src/backend/distributed/planner/multi_logical_optimizer.c
metdos: yeap

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: No that's not. Below we error out only saying `unsupported clause type`. But for sublinks we can produce better message

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: Ok, I'll update the pr

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: Yes, on the target list such as:
```SQL
  SELECT a.title AS name, (SELECT a2.id FROM articles_single_shard_hash_mx a2 WHERE a.id = a2.id  LIMIT 1)
  						 AS special_price FROM articles_hash_mx a;
```

onderkalaci:
> no flag does not include QTW_EXAMINE_RTES

But, the code you've shared is for `range_table_walker ()`. I think we should look at [`query_tree_walker()`](https://github.com/postgres/postgres/blob/8f0530f58061b185dc385df42e62d78a18d4ae3e/src/backend/nodes/nodeFuncs.c#L2267), not?

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -376,13 +435,20 @@ DeferErrorIfUnsupportedSubqueryPushdown(Query *originalQuery,
...
 	}
 
-	subqueryEntryList = SubqueryEntryList(originalQuery);
-	foreach(rangeTableEntryCell, subqueryEntryList)
-	{
-		RangeTblEntry *rangeTableEntry = lfirst(rangeTableEntryCell);
-		Query *subquery = rangeTableEntry->subquery;
+	/*
+	 * We first extract all the queries that appear in the orignal query. Later,
+	 * we delete the original query given that error rules does not apply to the
```
onderkalaci: makes sense

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1810,6 +1873,10 @@ JoinClauseList(List *whereClauseList)
  * subqueries but differently from the outermost query, they are run on a copy
  * of parse tree and changes do not get persisted as modifications to the original
  * query tree.
+ *
+ * Also this function adds SubLinks to the baseQualifierList when they appear on
```
metdos: Did we change anything on this function?

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: Oh, good point

src/backend/distributed/planner/multi_logical_planner.c
metdos: If we are sure that no sublinks will be passed into this function, I think we can remove this check.

metdos:
> no flag does not include QTW_EXAMINE_RTES
> But, the code you've shared is for range_table_walker (). I think we should look at query_tree_walker(), not?

`query_tree_walker()` calls `range_table_walker()`
https://github.com/postgres/postgres/blob/8f0530f58061b185dc385df42e62d78a18d4ae3e/src/backend/nodes/nodeFuncs.c#L2299

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: We didn't change anything. But, wanted to explicitly mention that if sublinks exist in the `WHERE` clause, they'll be added to the qualifiers list.

src/backend/distributed/planner/multi_router_planner.c
metdos: Do we want to set this to NULL? Also, we can take this to another commit as you noted.

src/test/regress/sql/multi_subquery_behavioral_analytics.sql
```diff
@@ -1600,5 +1618,47 @@ FROM (
...
+         events_table AS e
+    WHERE test_join_function_2(u.user_id, e.user_id)
+  ) t
+  GROUP BY user_id
+) q
+ORDER BY 2 DESC, 1;
+
+-- note that the following query has joins on the partition keys
+-- however we fail to push down it due to the function call on the
+-- where clause. We probably need to relax that check
```
metdos: Both tests seem to test joins via functions. If there are not different, we can drop one. If you would like to keep both, you can move these comments above.

src/test/regress/expected/multi_verify_no_subquery.out
```diff
@@ -1,18 +0,0 @@
---
--- MULTI_VERIFY_NO_SUBQUERY
```
metdos: Do we support all of these now?

src/test/regress/expected/multi_insert_select.out
```diff
@@ -142,7 +142,7 @@ FROM
   raw_events_first
 WHERE
   user_id < 0;
-NOTICE:  evaluating on master
+ERROR:  Subqueries without relations are not allowed in INSERT ... SELECT queries
```
onderkalaci: I think it should be fine. The test is not very meaningful and RTEs without relations may lead to errors. This seems a bugfix to me 😄

src/test/regress/expected/multi_insert_select.out
```diff
@@ -1106,23 +1106,7 @@ INSERT INTO agg_events
...
-DEBUG:  predicate pruning for shardId 13300000
-DEBUG:  predicate pruning for shardId 13300001
-DEBUG:  predicate pruning for shardId 13300003
-DEBUG:  distributed statement: INSERT INTO public.agg_events_13300010 AS citus_table_alias (user_id, value_1_agg) WITH sub_cte AS (SELECT 1) SELECT user_id, (SELECT sub_cte."?column?" FROM sub_cte) FROM public.raw_events_first_13300002 raw_events_first WHERE ((worker_hash(user_id) >= 0) AND (worker_hash(user_id) <= 1073741823))
-DEBUG:  predicate pruning for shardId 13300000
-DEBUG:  predicate pruning for shardId 13300001
-DEBUG:  predicate pruning for shardId 13300002
-DEBUG:  distributed statement: INSERT INTO public.agg_events_13300011 AS citus_table_alias (user_id, value_1_agg) WITH sub_cte AS (SELECT 1) SELECT user_id, (SELECT sub_cte."?column?" FROM sub_cte) FROM public.raw_events_first_13300003 raw_events_first WHERE ((worker_hash(user_id) >= 1073741824) AND (worker_hash(user_id) <= 2147483647))
-DEBUG:  Plan is router executable
+ERROR:  Subqueries without relations are not allowed in INSERT ... SELECT queries
```
onderkalaci: Same here: I think it should be fine. The test is not very meaningful and RTEs without relations may lead to errors. This seems a bugfix to me 😄

src/test/regress/expected/multi_mx_router_planner.out
```diff
@@ -456,15 +456,6 @@ DEBUG:  Plan is router executable
 SELECT * FROM articles_hash_mx, position('om' in 'Thomas') WHERE author_id = 1 or author_id = 2;
 ERROR:  could not run distributed query with complex table expressions
 HINT:  Consider using an equality filter on the distributed table's partition column.
--- subqueries are not supported in WHERE clause in Citus
```
onderkalaci: We have a bunch of tests that covers these cases. I really don't see value in keeping these tests. Do you?

src/test/regress/expected/multi_mx_router_planner.out
```diff
@@ -538,7 +529,7 @@ HINT:  Set citus.task_executor_type to "task-tracker".
 -- subqueries are not supported in SELECT clause
 SELECT a.title AS name, (SELECT a2.id FROM articles_single_shard_hash_mx a2 WHERE a.id = a2.id  LIMIT 1)
 						 AS special_price FROM articles_hash_mx a;
-ERROR:  could not run distributed query with subquery outside the FROM clause
+ERROR:  could not run distributed query with subquery outside the FROM and WHERE clauses
```
onderkalaci: But, all the error messages in `ErrorIfQueryNotSupported()` starts with `could not`. I don't think it is a good idea to only change this? We probably should do it later?

src/test/regress/expected/multi_sql_function.out
```diff
@@ -326,6 +332,11 @@ ERROR:  could not create distributed plan
 DETAIL:  Possibly this is caused by the use of parameters in SQL functions, which is not supported in Citus.
 HINT:  Consider using PL/pgSQL functions instead.
 CONTEXT:  SQL function "test_parameterized_sql_function" statement 1
+SELECT test_parameterized_sql_function_in_subquery_where(1);
```
onderkalaci: Coming in #1356 😄

src/test/regress/expected/multi_subquery_behavioral_analytics.out
```diff
@@ -546,62 +546,103 @@ ORDER BY user_lastseen DESC, user_id;
 
 ------------------------------------
 -- Count the number of distinct users_table who are in segment X and Y and Z
--- This query will be supported when we have subqueries in where clauses.
 ------------------------------------
-SELECT DISTINCT user_id
+SELECT user_id
```
onderkalaci: > i. Do we support distinct? Do we want to keep a regression with distinct?

Not, Citus does not support DISTINCT on the top level query at all.

> ii. Shall we group by another column other than user_id to show that we don't need to group by user_id?

I wanted to mimic `DISTINCT` via the `group by`. These queries are coming from `Raw Queries` document and I feel like sticking to those queries is slightly better for documentation?

src/test/regress/expected/multi_subquery_behavioral_analytics.out
```diff
@@ -684,18 +736,32 @@ WHERE
               users_table
              WHERE
    	          value_1 = 15 AND value_2 > 25
-            );
-ERROR:  could not run distributed query with join types other than INNER or OUTER JOINS
-HINT:  Consider joining tables on partition column and have equal filter on joining columns.
+            )
```
onderkalaci: Sure, I'll update it

src/test/regress/expected/multi_verify_no_subquery.out
onderkalaci: I don't think so. But, the tests were here to make sure that we error out on whenever we got subqueries in where clause. 
Now that the ones that we don't support would error out saying they are not joined on the partition key. And, we have a good coverage on the other test file for this.

src/test/regress/sql/multi_subquery_behavioral_analytics.sql
onderkalaci: I think keeping the tests is useful given that in the above one there is actually no join on the partition keys and join is done via the function. In the below one there is actually join on the partition key but we still error out (which we'll probably consider removing in the future).

> you can move these comments above.

Sure, I can do that.

src/backend/distributed/planner/multi_router_planner.c
onderkalaci: > Do we want to set this to NULL? 

Yes, we do. The reason is that we don't have a copy function for `shardInterval`  and this is really not required to copy for our purposes.

I'll move it to another commit while I refactor the PR

metdos:
> I think it should be fine. The test is not very meaningful and RTEs without relations may lead to errors. This seems a bugfix to me 😄

This test has an objective to test that we evaluate functions on the master node. See: https://github.com/citusdata/citus/commit/11031bcf5513ddca3be012cb70e906e61f7b08c9

Still, if you would like to change this, could you check with @marcocitus?

src/test/regress/expected/multi_mx_router_planner.out
metdos: okay

onderkalaci:
> This test has an objective to test that we evaluate functions on the master node. See: 11031bc

Yes, but we have plenty of more meaningful master function evaluation tests for `INSERT .. SELECT`. I think if we actually support `CTE`s, we could consider evaluating functions on CTEs. 

@marcocitus Does that sound reasonable to you as well?

src/test/regress/expected/multi_subquery_behavioral_analytics.out
metdos: Okay

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1248,11 +1277,11 @@ ErrorIfQueryNotSupported(Query *queryTree)
 	const char *filterHint = "Consider using an equality filter on the distributed "
 							 "table's partition column.";
 
-	if (queryTree->hasSubLinks)
+	if (queryTree->hasSubLinks && SublinkList(queryTree) == NIL)
```
metdos: Could you add a comment noting that why we can have hasSubLinks set but SublinkList(queryTree) empty?

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1752,11 +1787,43 @@ ValidateClauseList(List *clauseList)
...
 
-		bool selectClause = IsSelectClause(clause);
-		bool joinClause = IsJoinClause(clause);
-		bool orClause = or_clause(clause);
+		/* produce a better error message for sublinks */
+		if (IsSublinkClause(clause))
+		{
+			ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+							errmsg("cannot perform distributed planning on this query"),
+							errdetail(
```
metdos: We can fix this (citus_indent) side effect by keeping the line under 90 characters.

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1752,11 +1787,43 @@ ValidateClauseList(List *clauseList)
...
+
+
+/*
+ * ValidateSubqueryPushdownClauseList walks over the given list of clauses,
+ * and checks that we can recognize all the clauses. This function ensures
+ * that we do not drop an unsupported clause type on the floor, and thus
+ * prevents erroneous results.
+ *
+ * Note that this function is slightly different than ValidateClauseList(),
+ * additionally allowing subkinks.
```
metdos: subkinks -> sublinks

src/backend/distributed/planner/multi_logical_planner.c
metdos: Also, this message is not completely correct, right? We support some subqueries in where-clause.

src/test/regress/expected/multi_insert_select.out
```diff
@@ -672,11 +662,7 @@ INSERT INTO agg_events
...
-DEBUG:  distributed statement: INSERT INTO public.agg_events_13300009 AS citus_table_alias (user_id, value_1_agg) WITH sub_cte AS (SELECT 1) SELECT user_id, (SELECT sub_cte."?column?" FROM sub_cte) FROM public.raw_events_first_13300001 raw_events_first WHERE ((worker_hash(user_id) >= '-1073741824'::integer) AND (worker_hash(user_id) <= '-1'::integer))
-DEBUG:  distributed statement: INSERT INTO public.agg_events_13300010 AS citus_table_alias (user_id, value_1_agg) WITH sub_cte AS (SELECT 1) SELECT user_id, (SELECT sub_cte."?column?" FROM sub_cte) FROM public.raw_events_first_13300002 raw_events_first WHERE ((worker_hash(user_id) >= 0) AND (worker_hash(user_id) <= 1073741823))
-DEBUG:  distributed statement: INSERT INTO public.agg_events_13300011 AS citus_table_alias (user_id, value_1_agg) WITH sub_cte AS (SELECT 1) SELECT user_id, (SELECT sub_cte."?column?" FROM sub_cte) FROM public.raw_events_first_13300003 raw_events_first WHERE ((worker_hash(user_id) >= 1073741824) AND (worker_hash(user_id) <= 2147483647))
-DEBUG:  Plan is router executable
+ERROR:  Subqueries without relations are not allowed in INSERT ... SELECT queries
```
metdos: We say "-- We do support some CTEs" in the comment but we error out here. I'm okay not supporting this. Remove test/update comment?

src/test/regress/expected/multi_subquery_in_where_clause.out
```diff
@@ -0,0 +1,584 @@
+--
+-- multi subquery in where queries aims to expand existing subquery pushdown
+-- regression tests to cover more cases specifically subqueries in WHERE clause
+-- the tables that are used depends to multi_insert_select_behavioral_analytics_create_table.sql
+--
+-- We don't need shard id sequence here, so commented out to prevent conflicts with concurrent tests
+-- ALTER SEQUENCE pg_catalog.pg_dist_shardid_seq RESTART 1430000;
```
metdos: We can just delete these I guess.

src/test/regress/expected/multi_subquery_in_where_clause.out
```diff
@@ -0,0 +1,584 @@
...
+	user_id;
+ user_id 
+---------
+      13
+      17
+      23
+      25
+(4 rows)
+
+-- the following query doesn't have a meaningul result
```
metdos: meaningul  -> meaningful

src/test/regress/expected/multi_subquery_in_where_clause.out
```diff
@@ -0,0 +1,584 @@
...
+      42 |      55
+      42 |     471
+      51 |     758
+      72 |     897
+      82 |     691
+      95 |     951
+(7 rows)
+
+-- similar to the above query
+-- the following query doesn't have a meaningul result
```
metdos: meaningul -> meaningful

src/test/regress/expected/multi_subquery_in_where_clause.out
```diff
@@ -0,0 +1,584 @@
...
+	SELECT
+	  user_id
+	 FROM (
+	  SELECT
+	 	 subquery_1.user_id, count_pay
+	  FROM
+	  (
+	    (SELECT
+	      users_table.user_id,
+	      'action=>1'AS event,
```
metdos: 'action=>1'AS need space before AS

src/test/regress/expected/multi_subquery_in_where_clause.out
metdos: This repeats, better to make a search for it :)

src/test/regress/expected/multi_subquery_in_where_clause.out
```diff
@@ -0,0 +1,584 @@
...
+              events_table  
+           WHERE 
+              users_table.user_id = events_table.user_id AND event_type = 50
+           GROUP BY
+              user_id
+           OFFSET 3
+          );
+ERROR:  cannot push down this subquery
+DETAIL:  Offset clause is currently unsupported
+-- we can detect errors even if they appear in WHERE -> FROM -> WHERE
```
metdos: Didn't get what this comment means :)

src/test/regress/expected/multi_subquery_in_where_clause.out
onderkalaci: I'm surprised that it works like this.

metdos:
Looks good 🚢

Issue #1337, opened 2017-04-18
check-vanilla in test suite crashes Citus


byucesoy:
Running `make check-vanilla` in `citus/src/test/regression` causes Citus to crash. You need to compile postgresql from source to be able run these tests.

It seems that this #1268 breaks the tests, but I did not dig into it.

metdos:
Related to #1034.

anarazel:
@onderkalaci looks like uninitialized memory usage, possibly due to memory context resets:
```
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51                                                                                                                                            
#1  0x00007fdcc107c3fa in __GI_abort () at abort.c:89                                                                                                                                                                
#2  0x000000000092a387 in ExceptionalCondition (conditionName=0xae4eb8 "!(((list) == ((List *) ((void *)0)) || ((((const Node*)((list)))->type) == T_List)))", errorType=0xae4d99 "FailedAssertion",                 
    fileName=0xae4d60 "/home/andres/src/postgresql-9.5/src/backend/nodes/list.c", lineNumber=133) at /home/andres/src/postgresql-9.5/src/backend/utils/error/assert.c:54                                             
#3  0x00000000006c4644 in lappend (list=0x1159ee8, datum=0x1159ed8) at /home/andres/src/postgresql-9.5/src/backend/nodes/list.c:133                                                                                  
#4  0x00007fdcc0abb098 in multi_join_restriction_hook (root=0x11983d0, joinrel=0x1159698, outerrel=0x119e890, innerrel=0x119ebb8, jointype=JOIN_INNER, extra=0x7ffebbb889d0)                                         
    at /home/andres/src/citusdb/src/backend/distributed/planner/multi_planner.c:674                                                                                                                                  
#5  0x000000000070f0be in add_paths_to_joinrel (root=0x11983d0, joinrel=0x1159698, outerrel=0x119e890, innerrel=0x119ebb8, jointype=JOIN_INNER, sjinfo=0x7ffebbb88a60, restrictlist=0x1159a18)                       
    at /home/andres/src/postgresql-9.5/src/backend/optimizer/path/joinpath.c:222                                                                                                                                     
#6  0x00000000007116e6 in make_join_rel (root=0x11983d0, rel1=0x119e890, rel2=0x119ebb8) at /home/andres/src/postgresql-9.5/src/backend/optimizer/path/joinrels.c:754                                                
[...]

Issue #1336, opened 2017-04-18
check-isolation tests in test suite crashes Citus


byucesoy:
#Running `make check-isolation` in `citus/src/test/regression` causes Citus to crash. You need to compile postgresql from source to be able run these tests and you may also need to compile `postgresql/src/test/isolation` (which is not compiled automatically).

It seems that this #1113 breaks the tests, but I did not dig into it.

metdos:
Related to #1034.

anarazel:
@onderkalaci Assigning to you, since it's been triggered by one of your PRs. This is blocking some other, not super urgent, work...

lithp:
Here's the stacktrace:

```
#2  0x000000000081b4ec in ExceptionalCondition (conditionName=conditionName@entry=0x7fe075ecc63f "!(!foundInCache)", 
    errorType=errorType@entry=0x7fe075ebdf2f "FailedAssertion", fileName=fileName@entry=0x7fe075ecc52e "utils/metadata_cache.c", lineNumber=lineNumber@entry=748)
    at assert.c:54
#3  0x00007fe075ea5f72 in BuildCachedShardList (cacheEntry=cacheEntry@entry=0x2ac2640) at utils/metadata_cache.c:748
#4  0x00007fe075ea658d in BuildDistTableCacheEntry (cacheEntry=cacheEntry@entry=0x2ac2640) at utils/metadata_cache.c:577
#5  0x00007fe075ea66ce in LookupDistTableCacheEntry (relationId=relationId@entry=16642) at utils/metadata_cache.c:487
#6  0x00007fe075ea670b in IsDistributedTable (relationId=relationId@entry=16642) at utils/metadata_cache.c:195
#7  0x00007fe075e79f6a in ErrorIfUnsupportedTruncateStmt (truncateStatement=truncateStatement@entry=0x2a80740) at executor/multi_utility.c:2229
#8  0x00007fe075e7c556 in multi_ProcessUtility (parsetree=0x2a80740, 
    queryString=0x2a7f728 "\n    SELECT pg_sleep(20);\n    DEALLOCATE all;\n    TRUNCATE test_table;\n    PREPARE insertone AS INSERT INTO test_table VALUES(1, 1);\n    PREPARE insertall AS INSERT INTO test_table SELECT test_id, dat"..., context=PROCESS_UTILITY_QUERY, params=0x0, dest=0x2ab8ef0, completionTag=0x7ffedc8569e0 "")
[...]

PR #1343, opened 2017-04-20
Use custom compiled PostgreSQL in Travis for merge commits

### LINKED ISSUE #1034: Run Travis test against custom compiled postgres
That will:
- allow to run isolationtester tests as part of the normal tests (#917)
- allow testing against assertions (#208)
- allow running standard postgres tests as part of our tests (#1017)
- in theory it'd allow to run valgrind tests, but they take too long for travis (3-5h)

Possibly more.

@jasonmp85 ^

byucesoy:
Fixes #1034 

With this commit, we start to use custom compiled PostgreSQL builds in Travis
for merge commits. This allows us to run isolation tests and PostgreSQL's own
regression tests along with our regression tests in Travis.

Since manually compiling PostgreSQL takes more time and we also add new tests,
we only enable running these tests on merge commits.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,11 +8,13 @@ env:
   global:
     # GitHub API token for citus-bot
     - secure: degV+qb2xHiea7E2dGk/WLvmYjq4ZsBn6ZPko+YhRcNm2GRXRaU3FqMBIecPtsEEFYaL5GwCQq/CgBf9aQxgDQ+t2CrmtGTtI9AGAbVBl//amNeJOoLe6QvrDpSQX5pUxwDLCng8cvoQK7ZxGlNCzDKiu4Ep4DUWgQVpauJkQ9nHjtSMZvUqCoI9h1lBy9Mxh7YFfHPW2PAXCqpV4VlNiIYF84UKdX3MXKLy9Yt0JBSNTWLZFp/fFw2qNwzFvN94rF3ZvFSD7Wp6CIhT6R5/6k6Zx8YQIrjWhgm6OVy1osUA8X7W79h2ISPqKqMNVJkjJ+N8S4xuQU0kfejnQ74Ie/uJiHCmbW5W2TjpL1aU3FQpPsGwR8h0rSeHhJAJzd8Ma+z8vvnnQHDyvetPBB0WgA/VMQCu8uEutyfYw2hDmB2+l2dDwkViaI7R95bReAGrpd5uNqklAXuR7yOeArz0ZZpHV0aZHGcNBxznMaZExSVZ5DVPW38UPn7Kgse8BnOWeLgnA1hJVp6CmBCtu+hKYt+atBPgRbM8IUINnKKZf/Sk6HeJIJZs662jD8/X93vFi0ZtyV2jEKJpouWw8j4vrGGsaDzTEUcyJgDqZj7tPJptM2L5B3BcFJmkGj2HO3N+LGDarJrVBBSiEjhTgx4NnLiKZnUbMx547mCRg2akk2w=
+    # During pull request builds (not push builds), use custom-built PostgreSQL
+    - USE_CUSTOM_PG="${TRAVIS_PULL_REQUEST_SHA}"
   matrix:
     - PGVERSION=9.5
     - PGVERSION=9.6
 before_install:
-  - git clone -b v0.5.1 --depth 1 https://github.com/citusdata/tools.git
+  - git clone -b v0.6.1 --depth 1 https://github.com/citusdata/tools.git
   - sudo make -C tools install
   - setup_apt
   - curl https://install.citusdata.com/community/deb.sh | sudo bash
@@ -20,7 +22,9 @@ before_install:
 install:
   - install_uncrustify
   - install_pg
-  - sudo apt-get install -y "postgresql-${PGVERSION}-hll=2.10.1.citus-1"
+  - install_custom_pg
+  # download and install HLL manually, as custom builds won't satisfy deps
+  - apt-get download "postgresql-${PGVERSION}-hll=2.10.1.citus-1" && sudo dpkg --force-confold --force-confdef --force-all -i *hll*.deb
 before_script: citus_indent --quiet --check
 script: CFLAGS=-Werror pg_travis_multi_test check
 after_success:
--- a/src/test/regress/isolation_schedule
+++ b/src/test/regress/isolation_schedule
@@ -1,4 +1,4 @@
-test: isolation_add_node_vs_reference_table_operations
+ignore: isolation_add_node_vs_reference_table_operations
 
 # tests that change node metadata should precede 
 # isolation_cluster_management such that tests 
--- a/src/test/regress/pg_regress_multi.pl
+++ b/src/test/regress/pg_regress_multi.pl
@@ -467,7 +467,8 @@ if ($vanillatest)
     $ENV{PGPORT} = $masterPort;
     $ENV{PGUSER} = $user;
 
-    system("make -C $postgresBuilddir/src/test/regress installcheck-parallel")
+    system("make -C $postgresBuilddir/src/test/regress installcheck-parallel") == 0
+    or die "Could not run vanilla tests";
 }
 elsif ($isolationtester)
 {


byucesoy:
Tests are failing right now because of #1336 and #1337. Other than that, this PR is ready for review. However it makes sense to review [this PR](https://github.com/citusdata/tools/pull/10) first, because this PR depends on that PR to work properly. After merging that PR, we should change [this line](https://github.com/citusdata/citus/pull/1343/files#diff-354f30a63fb0907d4ad57269548329e3R16) to point correct branch.

.travis.yml
```diff
@@ -8,11 +8,13 @@ env:
...
     # GitHub API token for citus-bot
     - secure: degV+qb2xHiea7E2dGk/WLvmYjq4ZsBn6ZPko+YhRcNm2GRXRaU3FqMBIecPtsEEFYaL5GwCQq/CgBf9aQxgDQ+t2CrmtGTtI9AGAbVBl//amNeJOoLe6QvrDpSQX5pUxwDLCng8cvoQK7ZxGlNCzDKiu4Ep4DUWgQVpauJkQ9nHjtSMZvUqCoI9h1lBy9Mxh7YFfHPW2PAXCqpV4VlNiIYF84UKdX3MXKLy9Yt0JBSNTWLZFp/fFw2qNwzFvN94rF3ZvFSD7Wp6CIhT6R5/6k6Zx8YQIrjWhgm6OVy1osUA8X7W79h2ISPqKqMNVJkjJ+N8S4xuQU0kfejnQ74Ie/uJiHCmbW5W2TjpL1aU3FQpPsGwR8h0rSeHhJAJzd8Ma+z8vvnnQHDyvetPBB0WgA/VMQCu8uEutyfYw2hDmB2+l2dDwkViaI7R95bReAGrpd5uNqklAXuR7yOeArz0ZZpHV0aZHGcNBxznMaZExSVZ5DVPW38UPn7Kgse8BnOWeLgnA1hJVp6CmBCtu+hKYt+atBPgRbM8IUINnKKZf/Sk6HeJIJZs662jD8/X93vFi0ZtyV2jEKJpouWw8j4vrGGsaDzTEUcyJgDqZj7tPJptM2L5B3BcFJmkGj2HO3N+LGDarJrVBBSiEjhTgx4NnLiKZnUbMx547mCRg2akk2w=
+    # Bash command below checks if latest commit is a merge commit, if it is we will use custom compiled PostgreSQL
+    - USE_CUSTOM_PG_BUILD=$(git rev-list --merges HEAD~1..HEAD)
   matrix:
     - PGVERSION=9.5
     - PGVERSION=9.6
 before_install:
-  - git clone -b v0.5.1 --depth 1 https://github.com/citusdata/tools.git
+  - git clone -b support_for_custom_compiled_postgres --depth 1 https://github.com/citusdata/tools.git
```
anarazel: So, before merging, we'll have to change this back, right? I.e. the changes have to be merged into tools first?

.travis.yml
```diff
@@ -8,11 +8,13 @@ env:
   global:
     # GitHub API token for citus-bot
     - secure: degV+qb2xHiea7E2dGk/WLvmYjq4ZsBn6ZPko+YhRcNm2GRXRaU3FqMBIecPtsEEFYaL5GwCQq/CgBf9aQxgDQ+t2CrmtGTtI9AGAbVBl//amNeJOoLe6QvrDpSQX5pUxwDLCng8cvoQK7ZxGlNCzDKiu4Ep4DUWgQVpauJkQ9nHjtSMZvUqCoI9h1lBy9Mxh7YFfHPW2PAXCqpV4VlNiIYF84UKdX3MXKLy9Yt0JBSNTWLZFp/fFw2qNwzFvN94rF3ZvFSD7Wp6CIhT6R5/6k6Zx8YQIrjWhgm6OVy1osUA8X7W79h2ISPqKqMNVJkjJ+N8S4xuQU0kfejnQ74Ie/uJiHCmbW5W2TjpL1aU3FQpPsGwR8h0rSeHhJAJzd8Ma+z8vvnnQHDyvetPBB0WgA/VMQCu8uEutyfYw2hDmB2+l2dDwkViaI7R95bReAGrpd5uNqklAXuR7yOeArz0ZZpHV0aZHGcNBxznMaZExSVZ5DVPW38UPn7Kgse8BnOWeLgnA1hJVp6CmBCtu+hKYt+atBPgRbM8IUINnKKZf/Sk6HeJIJZs662jD8/X93vFi0ZtyV2jEKJpouWw8j4vrGGsaDzTEUcyJgDqZj7tPJptM2L5B3BcFJmkGj2HO3N+LGDarJrVBBSiEjhTgx4NnLiKZnUbMx547mCRg2akk2w=
+    # Bash command below checks if latest commit is a merge commit, if it is we will use custom compiled PostgreSQL
+    - USE_CUSTOM_PG_BUILD=$(git rev-list --merges HEAD~1..HEAD)
```
anarazel: If you'd add --count, then this'd always be set to 0 or 1, that seems easier to deal with.

.travis.yml
byucesoy: Yes, exactly. I changed this line temporarily to see tests are running as expected. I will change it before merging this PR.

jasonmp85: Open a PR in `citusdata/tools`, please, as I think we're starting to get quite a few branches open over there and we should sweep them up and make a new tag at some point.

.travis.yml
jasonmp85: This seems post-facto… Travis will build like this _after_ a PR merge, right? Why not use a Travis environment variable to detect whether this is a PR build and build _that_ using this config.

When you push the branch, Travis'll build the HEAD of that branch normally…

When you open a PR, Travis sets certain environment variables and builds a merge commit between the target and subject branches of the PR; we could do that build with the increased scrutiny.

jasonmp85: The environment variables are defined [here](https://docs.travis-ci.com/user/environment-variables/). `TRAVIS_PULL_REQUEST` might be useful. If it's _false_, run normally, otherwise, do this custom build.

.travis.yml
byucesoy: There is a [PR](https://github.com/citusdata/citus/pull/1343) in tools repo. I will change this line once it is merged.

jasonmp85: Heh that link points back here.

.travis.yml
jasonmp85: The other `TRAVIS_PULL` variables are better, as they are entirely blank when not a merge build, facilitating better use in conditionals.

jasonmp85:
@byucesoy for expediency's sake, I implemented some of the feedback I was going to have about these two reviews, mostly around simplicity. I don't think the build was passing before, but the PostgreSQL 9.6 build appears to work now; unfortunately, there is some sort of `libpq` version problem with the PostgreSQL 9.5 build.

@anarazel the error is `/usr/lib/postgresql/9.6/bin/psql: symbol lookup error: /usr/lib/postgresql/9.6/bin/psql: undefined symbol: PQsetErrorContextVisibility`, which looks like a 9.6 `psql` is being used with a 9.5 `libpq` or something. Any clue what could be happening?

anarazel:
On 2017-05-04 22:20:56 -0700, Jason Petersen wrote:
> @anarazel the error is `/usr/lib/postgresql/9.6/bin/psql: symbol lookup error: /usr/lib/postgresql/9.6/bin/psql: undefined symbol: PQsetErrorContextVisibility`, which looks like a 9.6 `psql` is being used with a 9.5 `libpq` or something. Any clue what could be happening?

Hum, that's a weird one.  Did you perchance disable installing the usual
packages we install?  Because we should get a very new libpq from the
system...

jasonmp85:
I worked around it by adding a `apt-get install -y libpq5` after the custom PG install. I believe the `nuke_pg` phase might remove most traces of PostgreSQL from the system, so I don't know what `libpq` we were being left with (is one installed by `make install`?)

Adding that gets the tests passing, but the `hll` fails with [an even weirder error](https://travis-ci.org/citusdata/citus/jobs/228994792#L2579). I have two test variants for "HLL present" and "HLL absent", but this is some sort of in-between state (like the control and extension files are present but it can't find the `so`?)

Testing out skipping the HLL install for now.

jasonmp85:
OK, so there are general problems with the tests themselves (removing the HLL install got me past most of the tests). [This build](https://travis-ci.org/citusdata/citus/jobs/229011007) shows parts of the vanilla tests failing but exits with status zero anyhow. We need the right status to have Travis know about a failure, and without that status, the scripts don't print out the `regression.diffs` file.

Can we verify that, locally, PostgreSQL 9.5 doesn't have problems? And that if it does, `make` exits with a non-zero status?

jasonmp85:
OK, so the problem is now about the `CONTEXT` lines in the 9.5 output. See [here](https://travis-ci.org/citusdata/citus/jobs/229014817#L3322-L3338). I can never remember how to read these diffs, but I'm assuming this has something to do with when PL/pgSQL does or doesn't print CONTEXT changing in 9.6. We've had something similar before; can't remember what caused or fixed it…

byucesoy:
There are already PostgreSQL installations(from 9.2 to 9.6) in travis. I guess when we compile PostgreSQL from source, travis uses new binaries but old libraries(In this case new binaries are 9.5 binaries and old libraries are 9.6 libraries).

I added `sudo apt-get --purge -y remove postgresql-*` to nuke_pg to completely remove all existing PostgreSQL artifacts.

Also I modified the $PATH to point correct directory. Before that `which psql` returned `/usr/bin/psql` which is not the location of the binaries we compiled.

jasonmp85:
Thanks for finding a solution; I added back the HLL install and everything looks :100:. Shipping.

jasonmp85:
Oops; I was mistaken. Was looking at the push build, not the pull request one. It has the same failure as before, but now in _both_ 9.6 and 9.5 (before was only 9.5).

PR #1348, opened 2017-04-25
Add view regression tests for increased subquery coverage

mtuncer:
- joins between views and tables
- joins between views
- union/union all queries involving views
- limit queries
- explain

### DIFF
--- a/src/test/regress/expected/multi_view.out
+++ b/src/test/regress/expected/multi_view.out
@@ -230,6 +230,23 @@ SELECT l_suppkey, count(*) FROM
 	GROUP BY l_suppkey ORDER BY 2 DESC, 1 LIMIT 5;
 ERROR:  cannot perform distributed planning on this query
 DETAIL:  Subqueries without group by clause are not supported yet
+-- repartition query on view with single table subquery
+CREATE VIEW supp_count_view AS SELECT * FROM (SELECT l_suppkey, count(*) FROM lineitem_hash_part GROUP BY 1) s1;
+SELECT * FROM supp_count_view ORDER BY 2 DESC, 1 LIMIT 10;
+ l_suppkey | count 
+-----------+-------
+      6104 |     8
+      1868 |     6
+      5532 |     6
+      5849 |     6
+      6169 |     6
+      6669 |     6
+      6692 |     6
+      7703 |     6
+      7869 |     6
+      8426 |     6
+(10 rows)
+
 SET citus.task_executor_type to DEFAULT;
 -- create a view with aggregate
 CREATE VIEW lineitems_by_shipping_method AS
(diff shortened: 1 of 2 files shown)


src/test/regress/expected/multi_view.out
```diff
@@ -268,3 +268,514 @@ SELECT * FROM  lineitems_by_orderkey WHERE l_orderkey = 100;
...
+HINT:  Consider using an equality filter on the distributed table's partition column.
+-- it is supported if it is a router query
+SELECT count(*) FROM events_table et WHERE et.user_id IN (SELECT user_id FROM recent_selected_users WHERE user_id = 90);
+ count 
+-------
+   109
+(1 row)
+
+-- expected this to work but it did not
+(SELECT user_id FROM recent_users)
```
onderkalaci: Why do you expect this to work? The UNION is not even in a subquery (and inside a view)? But even wrapping that into a top level query wouldn't help because then the leaf queries would include subqueries.

src/test/regress/expected/multi_view.out
```diff
@@ -268,3 +268,514 @@ SELECT * FROM  lineitems_by_orderkey WHERE l_orderkey = 100;
...
+		(SELECT user_id FROM recent_users) 
+		UNION
+		(SELECT user_id FROM selected_users) ) u
+	WHERE user_id < 15 AND user_id > 10;
+ count 
+-------
+     4
+(1 row)
+
+-- expected this to work but it does not
```
onderkalaci: Yes, you're right this may seem to be expected to work considering that the same query works with the relations. However, this is not a bug, I'd call this as a limitation. 

I think we should improve top level union/union all support in the next release and ensure that this test pass.

onderkalaci:
Oh, and I think it'd be great to have a `CREATE VIEW AS ... (A query that has subquery)`

src/test/regress/expected/multi_view.out
mtuncer: When this query is parsed it looks like
```
Query
    rtable
        rte (subquery 1)
        rte (subquery 2)
    setOperations (rte-1 union rte-2)
    joinTree (NULL)
```

I read the next comment now. I think we should keep this around for future reference

src/test/regress/expected/multi_view.out
```diff
@@ -268,3 +268,514 @@ SELECT * FROM  lineitems_by_orderkey WHERE l_orderkey = 100;
...
+      66
+      71
+      74
+      86
+      87
+      90
+     100
+(14 rows)
+
+-- outer join inside a subquery
```
mtuncer: it is a left join in inner query

src/test/regress/expected/multi_view.out
```diff
@@ -268,3 +268,514 @@ SELECT * FROM  lineitems_by_orderkey WHERE l_orderkey = 100;
...
+      39 |    96
+      71 |    95
+      74 |    93
+       6 |    89
+      58 |    87
+      50 |    79
+     100 |    55
+(14 rows)
+
+-- the same query with a left join however, it would still generate the same result
```
mtuncer: now it is left join.

mtuncer:
Addressed comments, rebase, squashed

PR #1349, opened 2017-04-25
Fix check-vanilla tests

### LINKED ISSUE #1337: check-vanilla in test suite crashes Citus

byucesoy:
Fixes #1337 

It semms that GEQO optimizations, when it is set to on, create their own memory context
and free it after when it is no longer necessary. In join multi_join_restriction_hook
we allocate our variables in the CurrentMemoryContext, which is GEQO's memory context
if it is active. To prevent deallocation of our variables when GEQO's memory context is
freed, we started to allocate memory fo these variables in separate MemoryContext.

PS: I did not add tests because we will cover check-vanilla automatically with the merge of #1343

### DIFF
--- a/src/backend/distributed/planner/multi_planner.c
+++ b/src/backend/distributed/planner/multi_planner.c
@@ -72,8 +72,7 @@ static PlannedStmt * FinalizeRouterPlan(PlannedStmt *localPlan, CustomScan *cust
 static void CheckNodeIsDumpable(Node *node);
 static List * CopyPlanParamList(List *originalPlanParamList);
 static PlannerRestrictionContext * CreateAndPushPlannerRestrictionContext(void);
-static RelationRestrictionContext * CurrentRelationRestrictionContext(void);
-static JoinRestrictionContext * CurrentJoinRestrictionContext(void);
+static PlannerRestrictionContext * CurrentPlannerRestrictionContext(void);
 static void PopPlannerRestrictionContext(void);
 static bool HasUnresolvedExternParamsWalker(Node *expression, ParamListInfo boundParams);
 
@@ -658,20 +657,40 @@ multi_join_restriction_hook(PlannerInfo *root,
 							JoinType jointype,
 							JoinPathExtraData *extra)
 {
-	JoinRestrictionContext *joinContext = NULL;
-	JoinRestriction *joinRestriction = palloc0(sizeof(JoinRestriction));
+	PlannerRestrictionContext *plannerRestrictionContext = NULL;
+	JoinRestrictionContext *joinRestrictionContext = NULL;
+	JoinRestriction *joinRestriction = NULL;
+	MemoryContext restrictionsMemoryContext = NULL;
+	MemoryContext oldMemoryContext = NULL;
 	List *restrictInfoList = NIL;
 
-	restrictInfoList = extra->restrictlist;
-	joinContext = CurrentJoinRestrictionContext();
-	Assert(joinContext != NULL);
+	/*
+	 * Use a memory context that's guaranteed to live long enough, could be
+	 * called in a more shorted lived one (e.g. with GEQO).
+	 */
+	plannerRestrictionContext = CurrentPlannerRestrictionContext();
+	restrictionsMemoryContext = plannerRestrictionContext->memoryContext;
+	oldMemoryContext = MemoryContextSwitchTo(restrictionsMemoryContext);
 
+	/*
+	 * We create a copy of restrictInfoList because it may be created in a memory
+	 * context which will be deleted when we still need it, thus we create a copy
+	 * of it in our memory context.
+	 */
+	restrictInfoList = copyObject(extra->restrictlist);
+
+	joinRestrictionContext = plannerRestrictionContext->joinRestrictionContext;
+	Assert(joinRestrictionContext != NULL);
+
+	joinRestriction = palloc0(sizeof(JoinRestriction));
 	joinRestriction->joinType = jointype;
 	joinRestriction->joinRestrictInfoList = restrictInfoList;
 	joinRestriction->plannerInfo = root;
 
-	joinContext->joinRestrictionList =
-		lappend(joinContext->joinRestrictionList, joinRestriction);
+	joinRestrictionContext->joinRestrictionList =
+		lappend(joinRestrictionContext->joinRestrictionList, joinRestriction);
+
+	MemoryContextSwitchTo(oldMemoryContext);
 }
 
 
@@ -684,7 +703,10 @@ void
 multi_relation_restriction_hook(PlannerInfo *root, RelOptInfo *relOptInfo, Index index,
 								RangeTblEntry *rte)
 {
-	RelationRestrictionContext *restrictionContext = NULL;
+	PlannerRestrictionContext *plannerRestrictionContext = NULL;
+	RelationRestrictionContext *relationRestrictionContext = NULL;
+	MemoryContext restrictionsMemoryContext = NULL;
+	MemoryContext oldMemoryContext = NULL;
 	RelationRestriction *relationRestriction = NULL;
 	DistTableCacheEntry *cacheEntry = NULL;
 	bool distributedTable = false;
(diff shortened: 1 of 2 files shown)


anarazel:
On 2017-04-25 03:30:05 -0700, Burak Yücesoy wrote:
> It semms that GEQO optimizations, when it is set to on, create their own memory context
> and free it after when it is no longer necessary. In join multi_join_restriction_hook
> we allocate our variables in the CurrentMemoryContext, which is GEQO's memory context
> if it is active. To prevent deallocation of our variables when GEQO's memory context is
> freed, we started to allocate them in MessageContext.

I don't think MessageContext is the right thing here.  Planning a query
can be invoked from a lot of different places (think e.g. UDFs), and
MessageContext is going to be largely independent.   How about just
catching the CurrentMemoryContext in CreateAndPushPlannerRestrictionContext()?

byucesoy:
I modified this PR to use CurrentMemoryContext which was active during CreateAndPushPlannerRestrictionContext call, as Andres suggested.

This PR is ready for review.

src/backend/distributed/planner/multi_planner.c
```diff
@@ -658,11 +659,29 @@ multi_join_restriction_hook(PlannerInfo *root,
 							JoinType jointype,
 							JoinPathExtraData *extra)
 {
+	JoinRestriction *joinRestriction = NULL;
 	JoinRestrictionContext *joinContext = NULL;
-	JoinRestriction *joinRestriction = palloc0(sizeof(JoinRestriction));
 	List *restrictInfoList = NIL;
+	MemoryContext restrictionsMemoryContext = CurrentRestrictionsMemoryContext();
+	MemoryContext oldMemoryContext = CurrentMemoryContext;
```
anarazel: It doesn't make sense to assign CurrentMemoryContext here, you're overwriting it below, so it's plainly misleading. I know we have that pattern in a bunch of places, but it's wrong.

src/backend/distributed/planner/multi_planner.c
```diff
@@ -658,11 +659,29 @@ multi_join_restriction_hook(PlannerInfo *root,
...
-	restrictInfoList = extra->restrictlist;
+	/*
+	 * Before reaching this point, some code blocks (for example GEQO optimizations)
+	 * may switch to their own memory context. If we allocate the memory here and the
+	 * lappend calls below in their context, allocated memory will be freed with that
+	 * code block's memory context. We need longer lifespan for these variables, so we
+	 * allocate them in the CurrentRestrictionsMemoryContext(), which is set to
+	 * CurrentMemoryContext at CreateAndPushPlannerRestrictionContext.
+	 */
+	oldMemoryContext = MemoryContextSwitchTo(restrictionsMemoryContext);
```
anarazel: I'd actually shorten that to "Use a memory context that's guaranteed to live long enough, could be called in a more shorted lived one (e.g. with GEQO).".

src/backend/distributed/planner/multi_planner.c
```diff
@@ -701,6 +724,16 @@ multi_relation_restriction_hook(PlannerInfo *root, RelOptInfo *relOptInfo, Index
...
 
+	/*
+	 * Before reaching this point, some code blocks (for example GEQO optimizations)
+	 * may switch to their own memory context. If we allocate the memory here and the
+	 * lappend calls below in their context, allocated memory will be freed with that
+	 * code block's memory context. We need longer lifespan for these variables, so we
+	 * allocate them in the CurrentRestrictionsMemoryContext(), which is set to
+	 * CurrentMemoryContext at CreateAndPushPlannerRestrictionContext.
+	 */
+	oldMemoryContext = MemoryContextSwitchTo(restrictionsMemoryContext);
```
anarazel: Shouldn't duplicate larger comments, how about referring to the other hook instead?

src/backend/distributed/planner/multi_planner.c
```diff
@@ -837,6 +874,27 @@ CurrentJoinRestrictionContext(void)
...
+
+	Assert(plannerRestrictionContextList != NIL);
+
+	plannerRestrictionContext =
+		(PlannerRestrictionContext *) linitial(plannerRestrictionContextList);
+
+	restrictionsMemoryContext = plannerRestrictionContext->restrictionsMemoryContext;
+
+	return restrictionsMemoryContext;
+}
```
anarazel: Why is this function needed?   The callers already use CurrentJoinRestrictionContext()/CurrentRelationRestrictionContext().

I think it'd make more sense to replace CurrentJoinRestrictionContext()/CurrentRelationRestrictionContext() with CurrentRestrictionContext(), and then have the individual functions access ->relationRestrictionContext/joinRestrictionContext/restrictionsMemoryContext;

@onderkalaci Why did you add those separate functions?

onderkalaci: There is no specific reason, it seemed easier to read the code. 

I'm ok to remove those functions and use `CurrentRestrictionContext ()` as you specified.

byucesoy:
@anarazel just using one function to get PlannerRestrictionContext makes sense. I added that function and removed the other ones. I also made the other changes you mentioned.

@anarazel, @lithp do you guys have any other comment?

src/backend/distributed/planner/multi_planner.c
```diff
@@ -658,20 +657,40 @@ multi_join_restriction_hook(PlannerInfo *root,
...
+	restrictionsMemoryContext = plannerRestrictionContext->memoryContext;
+	oldMemoryContext = MemoryContextSwitchTo(restrictionsMemoryContext);
+	joinRestriction = palloc0(sizeof(JoinRestriction));
+
+	/*
+	 * We create a copy of restrictInfoList because it may be created in a memory
+	 * context which will be deleted when we still need it, thus we create a copy
+	 * of it in our memory context.
+	 */
+	restrictInfoList = list_copy(extra->restrictlist);
```
marcocitus: We probably need a deep copy here.

marcocitus:
Pushed a commit that changes it to a deep copy and rebased.

src/backend/distributed/planner/multi_planner.c
```diff
@@ -658,20 +657,40 @@ multi_join_restriction_hook(PlannerInfo *root,
...
+
+	/*
+	 * We create a copy of restrictInfoList because it may be created in a memory
+	 * context which will be deleted when we still need it, thus we create a copy
+	 * of it in our memory context.
+	 */
+	restrictInfoList = copyObject(extra->restrictlist);
+
+	joinRestrictionContext = plannerRestrictionContext->joinRestrictionContext;
+	Assert(joinRestrictionContext != NULL);
```
lithp: What does this `Assert` win us? We're just going to segfault in a few lines anyway aren't we?

src/backend/distributed/planner/multi_planner.c
```diff
@@ -658,20 +657,40 @@ multi_join_restriction_hook(PlannerInfo *root,
...
-	joinContext = CurrentJoinRestrictionContext();
-	Assert(joinContext != NULL);
+	/*
+	 * Use a memory context that's guaranteed to live long enough, could be
+	 * called in a more shorted lived one (e.g. with GEQO).
+	 */
+	plannerRestrictionContext = CurrentPlannerRestrictionContext();
+	restrictionsMemoryContext = plannerRestrictionContext->memoryContext;
+	oldMemoryContext = MemoryContextSwitchTo(restrictionsMemoryContext);
+	joinRestriction = palloc0(sizeof(JoinRestriction));
```
lithp: This is a little confusingly placed, it's not really related to the memory context switching. It probably belongs above line 686, where the fields of `joinRestriction` are set.

src/backend/distributed/planner/multi_planner.c
```diff
@@ -695,11 +717,19 @@ multi_relation_restriction_hook(PlannerInfo *root, RelOptInfo *relOptInfo, Index
...
+	plannerRestrictionContext = CurrentPlannerRestrictionContext();
+	restrictionsMemoryContext = plannerRestrictionContext->memoryContext;
+	oldMemoryContext = MemoryContextSwitchTo(restrictionsMemoryContext);
+
 	distributedTable = IsDistributedTable(rte->relid);
 	localTable = !distributedTable;
 
-	restrictionContext = CurrentRelationRestrictionContext();
-	Assert(restrictionContext != NULL);
+	relationRestrictionContext = plannerRestrictionContext->relationRestrictionContext;
```
lithp: This initialization might be better-placed on line 750, right above where it's first used. And I'm not sure this `Assert` serves any purpose either, it's either a crash here or a crash on line 751.

src/backend/distributed/planner/multi_planner.c
```diff
@@ -795,44 +829,20 @@ CreateAndPushPlannerRestrictionContext(void)
 
 
 /*
- * CurrentRelationRestrictionContext returns the the last restriction context from the
- * relationRestrictionContext list.
+ * CurrentRestrictionContext returns the the last PlannerRestrictionContext from the
+ * plannerRestrictionContextList list.
```
lithp: The comments in this file are rather inconsistent but this is definitely not the last one in the list! Maybe "returns the most recently added PlannerRestrictionContext from the"

marcocitus:
@lithp Makes sense, made the improvements.

PR #1350, opened 2017-04-25
Fix VPATH builds broken in 087d8427e3791.

anarazel:
1) Generated files reside in the build directory, not the source
   directory.
2) As a generated file is now included in the build, add it to the
   include path (-I)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,14 +13,14 @@ include Makefile.global
 all: extension
 
 # build extension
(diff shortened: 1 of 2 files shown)


jasonmp85:
Also, sorry about this in the first place; didn't raise it during review as it didn't appear to be a regression to me. Turns out it also broke nightly builds.

PR #1351, opened 2017-04-26
Remove Pruning Debug Output

anarazel:
A followup PR will optimize shard-pruning so we don't linearly walk across all shards in a lot of cases.  That changes a *lot* of debug output, and will likely continue to change further.  Therefore it seems unrealistic to rely on the `predicate pruning for shardId ...` debug messages.

Those messages also have the issue that they output shardids to a lot of places where it's not needed, preventing parallelization of tests.

I've added a second commit, that tries to re-add coverage removed by the removal of these debug messages by using explain - that's not perfect, but seems to give a reasonable coverage?

### DIFF
--- a/src/test/regress/output/multi_subquery.source
+++ b/src/test/regress/output/multi_subquery.source
@@ -373,14 +373,14 @@ SELECT master_create_worker_shards('subquery_pruning_varchar_test_table', 4, 1);
 
 SET citus.subquery_pushdown TO TRUE;
 SET client_min_messages TO DEBUG2;
-SELECT * FROM 
-	(SELECT count(*) FROM subquery_pruning_varchar_test_table WHERE a = 'onder' GROUP BY a) 
+SELECT * FROM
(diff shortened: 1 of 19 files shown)


src/test/regress/output/multi_subquery.source
```diff
@@ -376,19 +376,13 @@ SET client_min_messages TO DEBUG2;
...
 
 SELECT * FROM 
 	(SELECT count(*) FROM subquery_pruning_varchar_test_table WHERE 'eren' = a GROUP BY a)
 AS foo;
-DEBUG:  predicate pruning for shardId 270005
-DEBUG:  predicate pruning for shardId 270007
-DEBUG:  predicate pruning for shardId 270008
  count 
 -------
 (0 rows)
```
marcocitus: Very unrelated, but this file has some `EXPLAIN` commands without `COSTS OFF` that might be worth fixing.

PR #1353, opened 2017-04-26
Refactor COPY to not directly use cache entry

### LINKED ISSUE #1338: Crash in CitusCopyFrom
Debugged this with @anarazel:

```
(gdb) bt
#0  FunctionCall1Coll (flinfo=0x7f0837c65fb8, collation=0, arg1=41301184) at fmgr.c:1297
#1  0x00007f082c821202 in FindShardInterval (partitionColumnValue=<optimized out>, shardIntervalCache=0x7f0837c5cbe0, shardCount=32, partitionMethod=104 'h', compareFunction=<optimized out>, hashFunction=<optimized out>, useBinarySearch=0 '\000')
    at utils/shardinterval_utils.c:248
#2  0x00007f082c7e0b0a in CopyToExistingShards (completionTag=0x7fff3ae56530 "", copyStatement=0x27130f8) at commands/multi_copy.c:465
#3  CitusCopyFrom (copyStatement=0x27130f8, completionTag=0x7fff3ae56530 "") at commands/multi_copy.c:182
#4  0x00007f082c7ed070 in ProcessCopyStmt (commandMustRunAsOwner=<optimized out>, completionTag=0x7fff3ae56530 "", copyStatement=0x27130f8) at executor/multi_utility.c:588
#5  multi_ProcessUtility (parsetree=0x27130f8, 
    queryString=0x26e6e70 "COPY "..., 
    context=PROCESS_UTILITY_TOPLEVEL, params=0x0, dest=0x2713458, completionTag=0x7fff3ae56530 "") at executor/multi_utility.c:230
#6  0x00007f082c134fae in pgss_ProcessUtility (parsetree=0x27130f8, 
    queryString=0x26e6e70 "COPY "...,
    context=PROCESS_UTILITY_TOPLEVEL, params=0x0, dest=0x2713458, completionTag=0x7fff3ae56530 "") at pg_stat_statements.c:982
#7  0x00000000006e0afc in PortalRunUtility (portal=0x2643c70, utilityStmt=0x27130f8, isTopLevel=1 '\001', setHoldSnapshot=<optimized out>, dest=0x2713458, completionTag=0x7fff3ae56530 "") at pquery.c:1193
#8  0x00000000006e19fb in PortalRunMulti (portal=0x2643c70, isTopLevel=1 '\001', setHoldSnapshot=0 '\000', dest=0x2713458, altdest=0x2713458, completionTag=0x7fff3ae56530 "") at pquery.c:1349
#9  0x00000000006e2164 in PortalRun (portal=0x2643c70, count=9223372036854775807, isTopLevel=1 '\001', dest=0x2713458, altdest=0x2713458, completionTag=0x7fff3ae56530 "") at pquery.c:815
#10 0x00000000006de6f1 in exec_simple_query (
    query_string=0x26e6e70 "COPY "...)
    at postgres.c:1094
#11 0x00000000006dffc8 in PostgresMain (argc=<optimized out>, argv=<optimized out>, dbname=0x26858a8 "citus", username=<optimized out>) at postgres.c:4076
#12 0x00000000006809c5 in BackendRun (port=0x2681190) at postmaster.c:4271
#13 BackendStartup (port=0x2681190) at postmaster.c:3945
#14 ServerLoop () at postmaster.c:1701
#15 PostmasterMain (argc=<optimized out>, argv=<optimized out>) at postmaster.c:1309
#16 0x0000000000608b48 in main (argc=3, argv=0x26246b0) at main.c:228

...
(gdb)
#2  0x00007f082c7e0b0a in CopyToExistingShards (completionTag=0x7fff3ae56530 "", copyStatement=0x27130f8) at commands/multi_copy.c:465
465	in commands/multi_copy.c
(gdb) p *cacheEntry->hashFunction 
$10 = {fn_addr = 0x4949d0 <hashtext>, fn_oid = 400, fn_nargs = 1, fn_strict = 1 '\001', fn_retset = 0 '\000', fn_stats = 2 '\002', fn_extra = 0x0, fn_mcxt = 0x269c7c0, fn_expr = 0x0}
(gdb) p cacheEntry 
$11 = (DistTableCacheEntry *) 0x271a038
```

Quoting @anarazel "Cache entry has been rebuilt and thats why the pointer to the hash function is pointing to freed memory"

jasonmp85:
If the cache is invalidated during a COPY, the reference will not point to valid memory. By deep copying the fields we actually need, we can avoid that.

Fixes #1338

### DIFF
--- a/src/backend/distributed/commands/multi_copy.c
+++ b/src/backend/distributed/commands/multi_copy.c
@@ -1753,12 +1753,6 @@ CitusCopyDestReceiverStartup(DestReceiver *dest, int operation,
 	/* keep the table metadata to avoid looking it up for every tuple */
 	copyDest->tableMetadata = cacheEntry;
 
-	/* determine whether to use binary search */
-	if (partitionMethod != DISTRIBUTE_BY_HASH || !cacheEntry->hasUniformHashDistribution)
-	{
-		copyDest->useBinarySearch = true;
-	}
-
 	if (cacheEntry->replicationModel == REPLICATION_MODEL_2PC)
 	{
 		CoordinatedTransactionUse2PC();
@@ -1835,18 +1829,9 @@ CitusCopyDestReceiverReceive(TupleTableSlot *slot, DestReceiver *dest)
 {
 	CitusCopyDestReceiver *copyDest = (CitusCopyDestReceiver *) dest;
 
-	DistTableCacheEntry *tableMetadata = copyDest->tableMetadata;
-	char partitionMethod = tableMetadata->partitionMethod;
 	int partitionColumnIndex = copyDest->partitionColumnIndex;
 	TupleDesc tupleDescriptor = copyDest->tupleDescriptor;
 	CopyStmt *copyStatement = copyDest->copyStatement;
-
-	int shardCount = tableMetadata->shardIntervalArrayLength;
-	ShardInterval **shardIntervalCache = tableMetadata->sortedShardIntervalArray;
-
-	bool useBinarySearch = copyDest->useBinarySearch;
-	FmgrInfo *hashFunction = tableMetadata->hashFunction;
-	FmgrInfo *compareFunction = tableMetadata->shardIntervalCompareFunction;
 
 	HTAB *shardConnectionHash = copyDest->shardConnectionHash;
 	CopyOutState copyOutState = copyDest->copyOutState;
@@ -1907,10 +1892,7 @@ CitusCopyDestReceiverReceive(TupleTableSlot *slot, DestReceiver *dest)
 	 * For reference table, this function blindly returns the tables single
 	 * shard.
 	 */
-	shardInterval = FindShardInterval(partitionColumnValue, shardIntervalCache,
-									  shardCount, partitionMethod,
-									  compareFunction, hashFunction,
-									  useBinarySearch);
+	shardInterval = FindShardInterval(partitionColumnValue, copyDest->tableMetadata);
 	if (shardInterval == NULL)
 	{
 		ereport(ERROR, (errcode(ERRCODE_OBJECT_NOT_IN_PREREQUISITE_STATE),
--- a/src/backend/distributed/utils/shardinterval_utils.c
+++ b/src/backend/distributed/utils/shardinterval_utils.c
@@ -23,9 +23,7 @@
 #include "utils/memutils.h"
 
 
-static int FindShardIntervalIndex(Datum searchedValue, ShardInterval **shardIntervalCache,
-								  int shardCount, char partitionMethod,
-								  FmgrInfo *compareFunction, bool useBinarySearch);
+static int FindShardIntervalIndex(Datum searchedValue, DistTableCacheEntry *cacheEntry);
 static int SearchCachedShardInterval(Datum partitionColumnValue,
 									 ShardInterval **shardIntervalCache,
 									 int shardCount, FmgrInfo *compareFunction);
@@ -188,12 +186,7 @@ ShardIndex(ShardInterval *shardInterval)
 	Datum shardMinValue = shardInterval->minValue;
 
 	DistTableCacheEntry *cacheEntry = DistributedTableCacheEntry(distributedTableId);
-	ShardInterval **shardIntervalCache = cacheEntry->sortedShardIntervalArray;
-	int shardCount = cacheEntry->shardIntervalArrayLength;
 	char partitionMethod = cacheEntry->partitionMethod;
-	FmgrInfo *compareFunction = cacheEntry->shardIntervalCompareFunction;
-	bool hasUniformHashDistribution = cacheEntry->hasUniformHashDistribution;
-	bool useBinarySearch = false;
 
 	/*
 	 * Note that, we can also support append and range distributed tables, but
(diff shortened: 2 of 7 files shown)


src/backend/distributed/commands/multi_copy.c
```diff
@@ -1751,7 +1752,24 @@ CitusCopyDestReceiverStartup(DestReceiver *dest, int operation,
 	LockShardListResources(shardIntervalList, ShareLock);
 
 	/* keep the table metadata to avoid looking it up for every tuple */
-	copyDest->tableMetadata = cacheEntry;
+	copyDest->partitionMethod = cacheEntry->partitionMethod;
+	copyDest->shardCount = cacheEntry->shardIntervalArrayLength;
+	copyDest->shardIntervalCache = (ShardInterval **)
+								   PointerArrayFromList(shardIntervalList);
```
marcocitus: It might be worth adding a comment on why we're calling `PointerArrayFromList` instead of using `cacheEntry->sortedShardIntervalArray` to prevent a sleepy developer from optimising this step away in the future.

anarazel:
On 2017-04-26 10:53:30 -0700, Marco Slot wrote:
> 
> 
> 
> > @@ -1751,7 +1752,24 @@ CitusCopyDestReceiverStartup(DestReceiver *dest, int operation,
>  	LockShardListResources(shardIntervalList, ShareLock);
>  
>  	/* keep the table metadata to avoid looking it up for every tuple */
> -	copyDest->tableMetadata = cacheEntry;
> +	copyDest->partitionMethod = cacheEntry->partitionMethod;
> +	copyDest->shardCount = cacheEntry->shardIntervalArrayLength;
> +	copyDest->shardIntervalCache = (ShardInterval **)
> +								   PointerArrayFromList(shardIntervalList);
> 
> It might be worth adding a comment on why we're calling `PointerArrayFromList` instead of using `cacheEntry->sortedShardIntervalArray` to prevent a sleepy developer from optimising this step away in the future.

Maybe I am sleep deprived, but what does this gain us?  The underlying
ShardIntervals matter too, no?

- Andres

jasonmp85:
> Maybe I am sleep deprived, but what does this gain us?  The underlying ShardIntervals matter too, no?

They do, which is why they were copied by the call to `LoadShardIntervalList`. Not very explicit, I know.

anarazel:
On 2017-04-26 11:06:53 -0700, Jason Petersen wrote:
> > Maybe I am sleep deprived, but what does this gain us?  The underlying ShardIntervals matter too, no?
> 
> They do, which is why they were copied by the call to `LoadShardIntervalList`. Not very explicit, I know.

Why do you need to copy the list then?

jasonmp85:
> Why do you need to copy the list then?

Because it needs to be an array?

anarazel:
I don't think this is the right fix.  Why don't we just instead reference the variables from the DistTableCacheEntry->XXX, instead of storing them in local vars?   That'd solve the issue as well, and'll likely be easier to backport too, besides being more performant.

jasonmp85:
I'm not sure about the more performant part? The copy calls I added only occur once, at the beginning of the `COPY`. It does fix the problem (just verified by manually adding an invalidation call in `CitusCopyDestReceiverReceive`).

src/backend/distributed/utils/shardinterval_utils.c
```diff
@@ -273,10 +254,14 @@ FindShardInterval(Datum partitionColumnValue, ShardInterval **shardIntervalCache
...
 static int
-FindShardIntervalIndex(Datum searchedValue, ShardInterval **shardIntervalCache,
-					   int shardCount, char partitionMethod, FmgrInfo *compareFunction,
-					   bool useBinarySearch)
+FindShardIntervalIndex(Datum searchedValue, DistTableCacheEntry *cacheEntry)
 {
+	ShardInterval **shardIntervalCache = cacheEntry->sortedShardIntervalArray;
+	int shardCount = cacheEntry->shardIntervalArrayLength;
+	char partitionMethod = cacheEntry->partitionMethod;
+	FmgrInfo *compareFunction = cacheEntry->shardIntervalCompareFunction;
```
marcocitus: I wonder if it's worth passing down the cache entry to `SearchCachedShardInterval` as well. 

(It also would be the most technically correct solution. It's very unlikely, but a compare function call could trigger a cache invalidation if it calls into some Citus logic that does a cache lookup. That would then cause the next compare function in the binary search to segfault.)

src/backend/distributed/utils/shardinterval_utils.c
```diff
@@ -273,10 +254,14 @@ FindShardInterval(Datum partitionColumnValue, ShardInterval **shardIntervalCache
...
+	bool useBinarySearch = (partitionMethod != DISTRIBUTE_BY_HASH ||
+							!cacheEntry->hasUniformHashDistribution);
```
anarazel: I think this change then essentially makes FastShardPruning obsolete, no?  I think we should just remove it in that case?

jasonmp85: Sure; running tests on that change and pushing up.

jasonmp85:
OK, added that change and rebased.

jasonmp85:
@anarazel anything blocking :shipit:?

PR #1354, opened 2017-04-26
Skip exhaustive test in CoPartitionedTables() if declared colocated.

anarazel:
That's considerably cheaper.


src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -3992,6 +3991,22 @@ CoPartitionedTables(Oid firstRelationId, Oid secondRelationId)
 
 	Assert(comparisonFunction != NULL);
 
+	/*
+	 * Check if both tables are marked as colocated - if so, we know they're
+	 * colocated.
```
marcocitus: "Check if the tables have the same colocation ID - ..." is more accurate.

PR #1355, opened 2017-04-26
Preliminary, metadata related, changes required for faster shard pruning.

anarazel:


src/backend/distributed/utils/metadata_cache.c
```diff
@@ -933,6 +934,52 @@ HasUninitializedShardInterval(ShardInterval **sortedShardIntervalArray, int shar
...
+		/* only called if !hasUninitializedShardInterval */
+		Assert(lastShardInterval->minValueExists && lastShardInterval->maxValueExists);
+		Assert(curShardInterval->minValueExists && curShardInterval->maxValueExists);
+
+		comparisonDatum = CompareCall2(shardIntervalSortCompareFunction,
+									   lastShardInterval->maxValue,
+									   curShardInterval->minValue);
+		comparisonResult = DatumGetInt32(comparisonDatum);
+
+		if (comparisonResult >= 0)
```
marcocitus: Should this be `comparisonResult > 0`? Ranges are inclusive.

anarazel:
@marcocitus One thing I'm slightly "concerned" about, is that this'll require a comparator for composite types used for hash-partitioned tables.  That seems pretty fair - and apparently there's no production users we know of that use this.

src/backend/distributed/utils/metadata_cache.c
anarazel: Hm.  But [1, 3] and [3, 5] - which is the case at hand, right? - should be marked as overlapping, no?

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -933,6 +934,52 @@ HasUninitializedShardInterval(ShardInterval **sortedShardIntervalArray, int shar
 
 
 /*
+ * HasOverlappingShardInterval determines whether the given list of sorted
+ * shards has overlapping ranges.
+ */
+static bool
+HasOverlappingShardInterval(ShardInterval **shardIntervalArray,
+							int shardIntervalArrayLength,
+							FmgrInfo *shardIntervalSortCompareFunction)
```
marcocitus: Maybe just name it `shardIntervalSortCompareFunction` for consistency.

src/include/distributed/metadata_cache.h
```diff
@@ -49,7 +50,15 @@ typedef struct
 	int shardIntervalArrayLength;
 	ShardInterval **sortedShardIntervalArray;
 
-	FmgrInfo *shardIntervalCompareFunction; /* NULL if no shard intervals exist */
+	/* comparator for partition column's type, NULL if DISTRIBUTE_BY_NONE */
+	FmgrInfo *shardValueCompareFunction;
```
marcocitus: This suggests there is a "shard value". Maybe `partitionColumnValueCompareFunction` or `partitionValueCompareFunction`.

src/backend/distributed/utils/metadata_cache.c
marcocitus: Ah right, read this the wrong way around.

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -933,6 +934,52 @@ HasUninitializedShardInterval(ShardInterval **sortedShardIntervalArray, int shar
 
 
 /*
+ * HasOverlappingShardInterval determines whether the given list of sorted
+ * shards has overlapping ranges.
+ */
+static bool
+HasOverlappingShardInterval(ShardInterval **shardIntervalArray,
```
marcocitus: How much do we need this function? We typically assume hash- and range-partitioned tables do not have overlapping shards. Is this mainly to distinguish append-partitioned tables that happen to not have overlapping shards?

anarazel: I don't think it's absolutely critical to have it, but I felt uncomfortable assuming that everyone using range partitioning was careful enough about their boundaries.  That it ends up working for append partitioned tables is a nice benefit.

src/include/distributed/metadata_cache.h
anarazel: I like ColumnValue, will update.

src/backend/distributed/utils/metadata_cache.c
anarazel: Hm? I'm not seeing any difference?

marcocitus: d'oh. `shardIntervalCompareFunction` without the `Sort`

src/backend/distributed/utils/metadata_cache.c
marcocitus: I think the main distinction between range- and append-partitioning is that ranges are distinct. It would be better to enforce this during shard creation. I'd prefer if we can trim down the code a bit and also lower the cost of cache invalidations with high shard counts.

anarazel: >  It would be better to enforce this during shard creation.

Right - but I think the users of range partitioning we have in practice actually manipulate shard state themselves. With the current state of out-of-the-box range-partitioning usability, it seems unrealistic to argue they just get to keep the pieces.  If we were able to have a constraint that enforces the boundaries (exclusion constraints would, but we'd have to manipulate catalogs differently), I'd be more welcoming on that front.

 > lower the cost of cache invalidations with high shard counts.

Actually measured that - in comparison to the loading of the shards (way to expensive) and the sorting (also fairly expensive), it doesn't show up in any sort of significant way. In my demo case, on an optimized build, it goes from 1102ms to 1089ms (avg of 3), neutering the check.

I've some fixes that'd make loading faster, but that's for another day/release.

I think it'll be very annoying to debug if we use boundary pruning with overlapping shards for some reason, so I'd rather err on being a bit paranoid.

marcocitus: We'll need to fix a lot more code if we stop assuming distinct ranges, so this seems pretty arbitrary

anarazel: Oh well, I think we'd be better off with the code, but I don't think it's that important. Dropping.

anarazel:
As discussed, I've re-added the overlaps test since we'd some range partitioned regression tests failing.  I've also added an error message for bad hash partitioned tables - triggered in our tests, but only in something with a comment ~"should error"

anarazel:
Closing, will merge as part of #1331

PR #1356, opened 2017-04-26
Add support for parametrized execution for subquery pushdown

onderkalaci:
Distributed query planning for subquery pushdown is done on the original
query. This prevents the usage of external parameters on the execution.
To overcome this, we manually replace the parameters on the original
query.

### DIFF
--- a/src/backend/distributed/planner/multi_logical_planner.c
+++ b/src/backend/distributed/planner/multi_logical_planner.c
@@ -132,6 +132,7 @@ static MultiNode * ApplyCartesianProduct(MultiNode *leftNode, MultiNode *rightNo
  * Local functions forward declarations for subquery pushdown. Note that these
  * functions will be removed with upcoming subqery changes.
  */
+static Node * ResolveExternalParams(Node *inputNode, ParamListInfo boundParams);
 static MultiNode * MultiSubqueryPlanTree(Query *originalQuery,
 										 Query *queryTree,
 										 PlannerRestrictionContext *
@@ -152,6 +153,12 @@ static MultiTable * MultiSubqueryPushdownTable(Query *subquery);
  * plan and adds a root node to top of it. The  original query is only used for subquery
  * pushdown planning.
  *
+ * In order to support external parameters for the queries where planning
+ * is done on the original query, we need to replace the external parameters
+ * manually. To achive that for subquery pushdown planning, we pass boundParams
+ * to this function. We need to do that since Citus currently unable to send
+ * parameters to the workers on the execution.
+ *
  * We also pass queryTree and plannerRestrictionContext to the planner. They
  * are primarily used to decide whether the subquery is safe to pushdown.
  * If not, it helps to produce meaningful error messages for subquery
@@ -159,7 +166,8 @@ static MultiTable * MultiSubqueryPushdownTable(Query *subquery);
  */
 MultiTreeRoot *
 MultiLogicalPlanCreate(Query *originalQuery, Query *queryTree,
-					   PlannerRestrictionContext *plannerRestrictionContext)
+					   PlannerRestrictionContext *plannerRestrictionContext,
+					   ParamListInfo boundParams)
 {
 	MultiNode *multiQueryNode = NULL;
 	MultiTreeRoot *rootNode = NULL;
@@ -176,6 +184,9 @@ MultiLogicalPlanCreate(Query *originalQuery, Query *queryTree,
 	 */
 	if (SubqueryEntryList(queryTree) != NIL || SublinkList(originalQuery) != NIL)
 	{
+		originalQuery = (Query *) ResolveExternalParams((Node *) originalQuery,
+														boundParams);
+
 		multiQueryNode = MultiSubqueryPlanTree(originalQuery, queryTree,
 											   plannerRestrictionContext);
 	}
@@ -189,6 +200,99 @@ MultiLogicalPlanCreate(Query *originalQuery, Query *queryTree,
 	SetChild((MultiUnaryNode *) rootNode, multiQueryNode);
 
 	return rootNode;
+}
+
+
+/*
+ * ResolveExternalParams replaces the external parameters that appears
+ * in the query with the corresponding entries in the boundParams.
+ *
+ * Note that this function is inspired by eval_const_expr() on Postgres.
+ * We cannot use that function because it requires access to PlannerInfo.
+ */
+static Node *
+ResolveExternalParams(Node *inputNode, ParamListInfo boundParams)
+{
+	/* consider resolving external parameters only when boundParams exists */
+	if (!boundParams)
+	{
+		return inputNode;
+	}
+
+	if (inputNode == NULL)
+	{
+		return NULL;
+	}
+
+	if (IsA(inputNode, Param))
+	{
+		Param *paramToProcess = (Param *) inputNode;
+		ParamExternData *correspondingParameterData = NULL;
+		int numberOfParameters = boundParams->numParams;
+		int parameterId = paramToProcess->paramid;
+		int16 typeLength = 0;
+		bool typeByValue = false;
+		Datum constValue = 0;
+		bool paramIsNull = false;
+		int parameterIndex = 0;
+
+		if (paramToProcess->paramkind != PARAM_EXTERN)
+		{
+			return inputNode;
+		}
+
+		if (parameterId < 0)
+		{
+			return inputNode;
+		}
+
+		/* parameterId starts from 1 */
+		parameterIndex = parameterId - 1;
+		if (parameterIndex >= numberOfParameters)
+		{
+			return inputNode;
+		}
+
+		correspondingParameterData = &boundParams->params[parameterIndex];
+
+		if (!(correspondingParameterData->pflags & PARAM_FLAG_CONST))
+		{
+			return inputNode;
+		}
+
+		get_typlenbyval(paramToProcess->paramtype, &typeLength, &typeByValue);
+
+		paramIsNull = correspondingParameterData->isnull;
+		if (paramIsNull)
+		{
+			constValue = 0;
+		}
+		else if (typeByValue)
+		{
+			constValue = correspondingParameterData->value;
+		}
+		else
+		{
+			/*
+			 * Out of paranoia ensure that datum lives long enough,
+			 * although bind params currently should always live
+			 * long enough.
+			 */
+			constValue = datumCopy(correspondingParameterData->value, typeByValue,
+								   typeLength);
+		}
+
+		return (Node *) makeConst(paramToProcess->paramtype, paramToProcess->paramtypmod,
+								  paramToProcess->paramcollid, typeLength, constValue,
+								  paramIsNull, typeByValue);
+	}
+	else if (IsA(inputNode, Query))
+	{
+		return (Node *) query_tree_mutator((Query *) inputNode, ResolveExternalParams,
+										   boundParams, 0);
+	}
+
+	return expression_tree_mutator(inputNode, ResolveExternalParams, boundParams);
 }
 
 
--- a/src/test/regress/expected/multi_subquery_misc.out
+++ b/src/test/regress/expected/multi_subquery_misc.out
@@ -1,0 +1,312 @@
+-- multi subquery pushdown misc aims to test subquery pushdown queries with
+--   (i)   Prepared statements
+--   (ii)  PL/PGSQL functions
+--   (iii) SQL functions
+-- the tables that are used depends to multi_insert_select_behavioral_analytics_create_table.sql
+-- We don't need shard id sequence here, so commented out to prevent conflicts with concurrent tests
+SET citus.enable_router_execution TO false;
+PREPARE prepared_subquery_1 AS
+SELECT
+    user_id,
+    user_lastseen,
+    array_length(event_array, 1)
+FROM (
+    SELECT
+        user_id,
+        max(u.time) as user_lastseen,
+        array_agg(event_type ORDER BY u.time) AS event_array
+    FROM (
+        SELECT user_id, time
+        FROM users_table
+        WHERE
+        user_id >= 10 AND
+        user_id <= 70 AND
+        users_table.value_1 > 10 AND users_table.value_1 < 12
+        ) u LEFT JOIN LATERAL (
+          SELECT event_type, time
+          FROM events_table
+          WHERE user_id = u.user_id AND
+          events_table.event_type > 10 AND events_table.event_type < 12
+        ) t ON true
+        GROUP BY user_id
+) AS shard_union
+ORDER BY user_lastseen DESC, user_id;
+EXECUTE prepared_subquery_1;
+ user_id |          user_lastseen          | array_length 
+---------+---------------------------------+--------------
+      12 | Sun Jan 19 01:49:20.372688 2014 |            1
+      20 | Sat Jan 18 14:25:31.817903 2014 |            1
+      42 | Thu Jan 16 07:08:02.651966 2014 |            1
+      56 | Tue Jan 14 12:11:47.27375 2014  |            1
+      57 | Mon Jan 13 14:53:50.494836 2014 |            1
+      65 | Sun Jan 12 03:14:26.810597 2014 |            1
+(6 rows)
+
+PREPARE prepared_subquery_2(int, int) AS
+SELECT
+    user_id,
+    user_lastseen,
+    array_length(event_array, 1)
+FROM (
+    SELECT
+        user_id,
+        max(u.time) as user_lastseen,
+        array_agg(event_type ORDER BY u.time) AS event_array
+    FROM (
+        SELECT user_id, time
+        FROM users_table
+        WHERE
+        user_id >= $1 AND
+        user_id <= $2 AND
+        users_table.value_1 > 10 AND users_table.value_1 < 12
+        ) u LEFT JOIN LATERAL (
+          SELECT event_type, time
+          FROM events_table
+          WHERE user_id = u.user_id AND
+          events_table.event_type > 10 AND events_table.event_type < 12
+        ) t ON true
+        GROUP BY user_id
+) AS shard_union
+ORDER BY user_lastseen DESC, user_id;
+-- should be fine with more than five executions
+EXECUTE prepared_subquery_2(10, 70);
+ user_id |          user_lastseen          | array_length 
+---------+---------------------------------+--------------
+      12 | Sun Jan 19 01:49:20.372688 2014 |            1
+      20 | Sat Jan 18 14:25:31.817903 2014 |            1
+      42 | Thu Jan 16 07:08:02.651966 2014 |            1
+      56 | Tue Jan 14 12:11:47.27375 2014  |            1
+      57 | Mon Jan 13 14:53:50.494836 2014 |            1
+      65 | Sun Jan 12 03:14:26.810597 2014 |            1
+(6 rows)
+
+EXECUTE prepared_subquery_2(10, 70);
+ user_id |          user_lastseen          | array_length 
+---------+---------------------------------+--------------
+      12 | Sun Jan 19 01:49:20.372688 2014 |            1
+      20 | Sat Jan 18 14:25:31.817903 2014 |            1
+      42 | Thu Jan 16 07:08:02.651966 2014 |            1
+      56 | Tue Jan 14 12:11:47.27375 2014  |            1
+      57 | Mon Jan 13 14:53:50.494836 2014 |            1
+      65 | Sun Jan 12 03:14:26.810597 2014 |            1
+(6 rows)
+
+EXECUTE prepared_subquery_2(10, 70);
+ user_id |          user_lastseen          | array_length 
+---------+---------------------------------+--------------
+      12 | Sun Jan 19 01:49:20.372688 2014 |            1
+      20 | Sat Jan 18 14:25:31.817903 2014 |            1
+      42 | Thu Jan 16 07:08:02.651966 2014 |            1
+      56 | Tue Jan 14 12:11:47.27375 2014  |            1
+      57 | Mon Jan 13 14:53:50.494836 2014 |            1
+      65 | Sun Jan 12 03:14:26.810597 2014 |            1
+(6 rows)
+
+EXECUTE prepared_subquery_2(10, 70);
+ user_id |          user_lastseen          | array_length 
+---------+---------------------------------+--------------
+      12 | Sun Jan 19 01:49:20.372688 2014 |            1
+      20 | Sat Jan 18 14:25:31.817903 2014 |            1
+      42 | Thu Jan 16 07:08:02.651966 2014 |            1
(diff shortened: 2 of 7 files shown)


src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -153,14 +155,24 @@ static MultiTable * MultiSubqueryPushdownTable(Query *subquery);
...
+ *
  * We also pass queryTree and plannerRestrictionContext to the planner. They
  * are primarily used to decide whether the subquery is safe to pushdown.
  * If not, it helps to produce meaningful error messages for subquery
  * pushdown planning.
  */
 MultiTreeRoot *
 MultiLogicalPlanCreate(Query *originalQuery, Query *queryTree,
-					   PlannerRestrictionContext *plannerRestrictionContext)
+					   PlannerRestrictionContext *plannerRestrictionContext,
```
onderkalaci: I considered not passing `boundParams` here. However, that would make things more complicated things on the `CreateDistributedPlan()` function by adding the checks there as well. But, I'm open to suggestions here.

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -194,6 +214,92 @@ MultiLogicalPlanCreate(Query *originalQuery, Query *queryTree,
...
 
 /*
+ * ReplaceExternParamsOnOriginalQuery replaces the external parameters that appears
+ * on the query with the corresponding entries on the boundParams.
+ *
+ * Note that this function is inspired by eval_const_expr() on Postgres. We failed to
+ * use that function given that it requires access to PlannerInfo.
+ */
+static Node *
+ReplaceExternParamsOnOriginalQuery(Node *inputNode, ParamListInfo boundParams)
```
marcocitus: I wonder if we could use `ExecuteMasterEvaluableFunctions` instead.

onderkalaci: I think we probably don't want to use `ExecuteMasterEvaluableFunctions ()` given that we don't want to evaluate other stuff except for the PARAMs on the `originalQuery`.

I've updated the function comment (which I was doing even before your comment 😄 )

marcocitus: Makes sense, though it kind of sucks to have 4 (!) different ways of handling prepared statements though: deferred error, `ReplaceExternParamsOnOriginalQuery`, `ExecuteMasterEvaluableFunctions`, send parameters in router executor.

I think we need to do some significant refactoring after 6.2 to make the planner accessible for future generations. We should generally decouple planning from pruning.

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -152,14 +154,24 @@ static MultiTable * MultiSubqueryPushdownTable(Query *subquery);
  * plan and adds a root node to top of it. The  original query is only used for subquery
  * pushdown planning.
  *
+ * In order to support external parameters for the queries where planning
+ * is done on the original query, we need to replace the external parameters
+ * manually. To achive that for subquery pushdown planning, we pass boundParams
+ * to this function.
```
anarazel: Worth mentioning that we only need to do so, because we currently are unable to send parameters to workers?

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -152,14 +154,24 @@ static MultiTable * MultiSubqueryPushdownTable(Query *subquery);
...
+ *
+ * Note that although router planner uses original query, the above does not
+ * apply given that we send the parameters to the workers explicitly on the
+ * execution.
+ *
  * We also pass queryTree and plannerRestrictionContext to the planner. They
  * are primarily used to decide whether the subquery is safe to pushdown.
  * If not, it helps to produce meaningful error messages for subquery
  * pushdown planning.
```
anarazel: Then this could be deleted, too.

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -193,6 +213,94 @@ MultiLogicalPlanCreate(Query *originalQuery, Query *queryTree,
 
 
 /*
+ * ReplaceExternParamsOnOriginalQuery replaces the external parameters that appears
+ * on the query with the corresponding entries on the boundParams.
```
anarazel: s/on the query/in the query/ s/on the boundParams/in boundParams/

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -193,6 +213,94 @@ MultiLogicalPlanCreate(Query *originalQuery, Query *queryTree,
...
+ *
+ * Note that this function is inspired by eval_const_expr() on Postgres. We failed to
```
anarazel: s/failed to/cannot/, /given that/because/

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -193,6 +213,94 @@ MultiLogicalPlanCreate(Query *originalQuery, Query *queryTree,
...
+ * use that function given that it requires access to PlannerInfo and does evaluations
+ * on other parts of the query.
```
anarazel: Don't see how the latter is an argument?

src/test/regress/sql/multi_subquery_misc.sql
```diff
@@ -0,0 +1,157 @@
...
+--   (i)   Prepared statements
+--   (ii)  PL/PGSQL functions
+--   (iii) SQL functions
+
+-- the tables that are used depends to multi_insert_select_behavioral_analytics_create_table.sql
+
+-- We don't need shard id sequence here, so commented out to prevent conflicts with concurrent tests
+-- ALTER SEQUENCE pg_catalog.pg_dist_shardid_seq RESTART 1430000;
+-- ALTER SEQUENCE pg_catalog.pg_dist_jobid_seq RESTART 1430000;
+SET citus.enable_router_execution TO false;
```
anarazel: It seems a bit weird to have commented out ALTER SEQUENCE statements here.

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: I thought the next paragraph explains that?

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: For example, if the query includes a `Const` expression such as:
```SQL
SELECT 3+5, user_id FROM ..
```

`eval_const_expr ()` replaces that with
```SQL
SELECT 8, user_id FROM ..
```

which we don't want because it is in the original query? (Similarly for immutable functions etc)

anarazel:
On 2017-04-28 10:25:04 -0700, Önder Kalacı wrote:
> onderkalaci commented on this pull request.
> 
> 
> 
> > @@ -152,14 +154,24 @@ static MultiTable * MultiSubqueryPushdownTable(Query *subquery);
>   * plan and adds a root node to top of it. The  original query is only used for subquery
>   * pushdown planning.
>   *
> + * In order to support external parameters for the queries where planning
> + * is done on the original query, we need to replace the external parameters
> + * manually. To achive that for subquery pushdown planning, we pass boundParams
> + * to this function.
> 
> I thought the next paragraph explains that? 

Well, they explain it w/ a reference to router, I'd rather explain it on
its own terms.


> > @@ -193,6 +213,94 @@ MultiLogicalPlanCreate(Query *originalQuery, Query *queryTree,
>  
>  
>  /*
> + * ReplaceExternParamsOnOriginalQuery replaces the external parameters that appears
> + * on the query with the corresponding entries on the boundParams.
> + *
> + * Note that this function is inspired by eval_const_expr() on Postgres. We failed to
> + * use that function given that it requires access to PlannerInfo and does evaluations
> + * on other parts of the query.
> 
> For example, if the query includes a `Const` expression such as:
> ```SQL
> SELECT 3+5, user_id FROM ..
> ```
> 
> `eval_const_expr ()` replaces that with
> ```SQL
> SELECT 8, user_id FROM ..
> ```
> 
> which we don't want because it is in the original query? (Similarly for immutable functions etc)

So, what'd be the problem with that?  It's fine not to repeat all that
on the workers...

- Andres

onderkalaci:
> So, what'd be the problem with that?  It's fine not to repeat all that
on the workers...

My only concern is that we're referring the query as original query, and after `check_eval_expr()` it would slightly diverge from being an original query. But, I can re-word the comment if you wish.

anarazel:
On 2017-04-28 10:37:43 -0700, Önder Kalacı wrote:
> > So, what'd be the problem with that?  It's fine not to repeat all that
> on the workers...
> 
> My only concern is that we're referring the query as original query, and after `check_eval_expr()` it would slightly diverge from being an original query. But, I can re-word the comment if you wish.

original really refers to not being processed by standard_planner() -
that's the crucial property, because it munges very heavily with the
structure, building constructs that aren't necessarily trivially
representable as SQL anymore.

- Andres

onderkalaci:
> original really refers to not being processed by standard_planner() -
that's the crucial property, because it munges very heavily with the
structure, building constructs that aren't necessarily trivially
representable as SQL anymore.

That makes sense, removed the related comment

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -130,6 +130,8 @@ static MultiNode * ApplyCartesianProduct(MultiNode *leftNode, MultiNode *rightNo
  * Local functions forward declarations for subquery pushdown. Note that these
  * functions will be removed with upcoming subqery changes.
  */
+static Node * ReplaceExternParamsInOriginalQuery(Node *inputNode,
```
anarazel: I don't think it's a good idea to say "originalQuery" here.  This is a pretty general facility.  Maybe "ResolveExternalParams" or "InstantiateExternalParamsAsConsts"?

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -148,14 +150,25 @@ static MultiTable * MultiSubqueryPushdownTable(Query *subquery);
...
  *
+ * In order to support external parameters for the queries where planning
+ * is done on the original query, we need to replace the external parameters
+ * manually. To achive that for subquery pushdown planning, we pass boundParams
+ * to this function. We need to do that since Citus currently unable to send
+ * parameters to the workers on the execution.
+ *
+ * Note that although router planner uses original query, the above does not
+ * apply given that we send the parameters to the workers explicitly on the
+ * execution.
```
anarazel: I'd just delete that graf.

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -186,6 +207,93 @@ MultiLogicalPlanCreate(Query *originalQuery, Query *queryTree,
...
+ * ReplaceExternParamsInOriginalQuery replaces the external parameters that appears
+ * in the query with the corresponding entries in the boundParams.
+ *
+ * Note that this function is inspired by eval_const_expr() on Postgres. We cannot
+ * use that function because it requires access to PlannerInfo.
+ */
+static Node *
+ReplaceExternParamsInOriginalQuery(Node *inputNode, ParamListInfo boundParams)
+{
+	int parameterIndex = 0;
```
anarazel: Why's this declared at the top scope?

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -186,6 +207,93 @@ MultiLogicalPlanCreate(Query *originalQuery, Query *queryTree,
...
+
+	if (inputNode == NULL)
+	{
+		return NULL;
+	}
+
+	if (IsA(inputNode, Param))
+	{
+		Param *paramToProcess = (Param *) inputNode;
+		ParamExternData correspondingParameterData;
```
anarazel: Shouldn't this be a pointer instead?  This way we're currently copying the whole struct.

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -186,6 +207,93 @@ MultiLogicalPlanCreate(Query *originalQuery, Query *queryTree,
...
+
+		if (paramToProcess->paramkind != PARAM_EXTERN)
+		{
+			return inputNode;
+		}
+
+		if (parameterId < 0)
+		{
+			return inputNode;
+		}
```
anarazel: Optional: If you'd move the rest of the current scope into something like
```
if (paramToProcess->paramkind != PARAM_EXTERN && parameterId < 0)
{
}
```
you could move most  variable declarations inside, directly defining them, instead of using some pointless assignment first.

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -186,6 +207,93 @@ MultiLogicalPlanCreate(Query *originalQuery, Query *queryTree,
...
+		{
+			constValue = 0;
+		}
+		else if (typeByValue)
+		{
+			constValue = correspondingParameterData.value;
+		}
+		else
+		{
+			constValue = datumCopy(correspondingParameterData.value, typeByValue,
```
anarazel: Hm. Actually. Why do we need to copy the datum?  We can just reference it in the const I think?

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: > ResolveExternalParams vs InstantiateExternalParamsAsConsts

`ResolveExternalParams()` sounds good to me.

(Also, as one of your previous comments suggested, I think I've misused `originalQuery` name/concept slightly wrong.)

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: That seems wrong

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: Ok, that probably makes more sense. That is also how it is used in `eval_const_expressions_mutator()`.

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: But, `eval_const_expressions_mutator()` copies the `Datum`. I'd be more comfortable with sticking that, do you?

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: sure

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -176,6 +184,13 @@ MultiLogicalPlanCreate(Query *originalQuery, Query *queryTree,
 	 */
 	if (SubqueryEntryList(queryTree) != NIL || SublinkList(originalQuery) != NIL)
 	{
+		/* consider resolving external parameters only when boundParams exists */
+		if (boundParams)
+		{
+			originalQuery = (Query *) ResolveExternalParams((Node *) originalQuery,
+															boundParams);
+		}
```
anarazel: Hm, it'd be a bit nicer if we'd move the if into ResolveExternalParams.

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -193,6 +208,90 @@ MultiLogicalPlanCreate(Query *originalQuery, Query *queryTree,
...
+		{
+			constValue = 0;
+		}
+		else if (typeByValue)
+		{
+			constValue = correspondingParameterData->value;
+		}
+		else
+		{
+			constValue = datumCopy(correspondingParameterData->value, typeByValue,
```
anarazel: Maybe add a comment like `/* out of paranoia ensure that datum lives long enough, although bind params currently should always live long enough */`?  Because I don't see otherwise why it's needed?

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -152,14 +153,21 @@ static MultiTable * MultiSubqueryPushdownTable(Query *subquery);
...
+ * to this function. We need to do that since Citus currently unable to send
+ * parameters to the workers on the execution.
+ *
  * We also pass queryTree and plannerRestrictionContext to the planner. They
  * are primarily used to decide whether the subquery is safe to pushdown.
  * If not, it helps to produce meaningful error messages for subquery
  * pushdown planning.
  */
 MultiTreeRoot *
 MultiLogicalPlanCreate(Query *originalQuery, Query *queryTree,
```
anarazel: Perhaps we should rename 'original' at some point to? Perhaps 'untransformed'?

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -193,6 +208,90 @@ MultiLogicalPlanCreate(Query *originalQuery, Query *queryTree,
...
+	{
+		Param *paramToProcess = (Param *) inputNode;
+		ParamExternData *correspondingParameterData = NULL;
+		int numberOfParameters = boundParams->numParams;
+		int parameterId = paramToProcess->paramid;
+		int16 typeLength = 0;
+		bool typeByValue = false;
+		Datum constValue = 0;
+		bool paramIsNull = false;
+		int parameterIndex = 0;
```
anarazel: Why don't you initialize parameterIndex fully up here?

src/test/regress/expected/multi_subquery_misc.out
```diff
@@ -0,0 +1,265 @@
...
+			    WHERE
+			    	short_list.user_id = ma.user_id and ma.value_1 < $1 and short_list.event_type < 50
+			    ) temp
+			  ON users_table.user_id = temp.user_id
+			  WHERE 
+			    users_table.value_1 < $2;
+
+END;
+$$ LANGUAGE plpgsql;
+-- run more than 5 times
```
anarazel: s/more than 5 times/enough times (6+) to actually use prepared statements/

src/test/regress/sql/multi_subquery_misc.sql
```diff
@@ -0,0 +1,155 @@
...
+WHERE user_id IN (SELECT user_id FROM users_table WHERE value_1 >= $4 AND value_1 <= $3)
+    AND user_id IN (SELECT user_id FROM users_table WHERE value_1 >= $5 AND value_1 <= $6)
+    AND user_id IN (SELECT user_id FROM users_table WHERE  value_1 >= $1 AND value_1 <= $2)
+GROUP BY 
+  user_id
+ORDER BY
+  user_id DESC
+  LIMIT 5;
+
+EXECUTE prepared_subquery_3(50, 60, 20, 10, 30, 40);
```
anarazel: Why only once?

src/test/regress/sql/multi_subquery_misc.sql
```diff
@@ -0,0 +1,155 @@
...
+	  WHERE 
+	    users_table.value_1 < $2;
+$$ LANGUAGE SQL;
+
+-- should error out
+SELECT sql_subquery_test(5,5);
+
+DEALLOCATE prepared_subquery_1;
+DEALLOCATE prepared_subquery_2;
+DEALLOCATE prepared_subquery_3;
```
anarazel: That seems superflous?

src/test/regress/sql/multi_subquery_misc.sql
```diff
@@ -0,0 +1,155 @@
...
+
+DROP FUNCTION plpgsql_subquery_test(int, int);
+DROP FUNCTION sql_subquery_test(int, int);
+
+SET citus.enable_router_execution TO true;
```
anarazel: That's really not useful.  I know we do that in a bunch of places, but it's not a good thing to repeat.  GUCs don't survive longer than a single session (at most a single test file, lest \c is used, then shorter).

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: I don't have any strong opinions on this, but, I initialized it below since the function could return before using it at all. (i.e., see if `parameterId` < 0) return inputNode).

Also, adding the comment (i.e., `parameterId starts from 1`) seemed a good fit once we initialize `parameterIndex `.
Do you want me to update it?

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: makes sense, I'll add the comment

src/test/regress/sql/multi_subquery_misc.sql
onderkalaci: I though we already have the tests with >5 times executed. This is mostly to ensure that it works with subqueries in WHERE clause as well. 
But, thinking more, I'll repeat this 6 times as well.

src/backend/distributed/planner/multi_logical_planner.c
anarazel: > I don't have any strong opinions on this, but, I initialized it below since the function could return before using it at all. (i.e., see if parameterId < 0) return inputNode).

Compilers can do that automatically.

>  Do you want me to update it?

You choose...

src/test/regress/sql/multi_subquery_misc.sql
onderkalaci: Oh, I didn't know that. I'll remove it.

src/test/regress/sql/multi_subquery_misc.sql
onderkalaci: Again not a big deal, but looks better with `DROP FUNCTION`s below?

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: Makes sense.

As a side question, can't we implement the prepared statement support in general as we do it here? (i..e, Move `ResolveExternalParams()` before `standard_planner()` call?)

onderkalaci:
@anarazel I've updated the PR according to your comments. Anything blocking to merge this?

Thanks!

src/backend/distributed/planner/multi_logical_planner.c
anarazel: > As a side question, can't we implement the prepared statement support in general as we do it here? (i..e, Move ResolveExternalParams() before standard_planner() call?)

No, the parameters aren't generally available, only when a "custom" plan is chosen, not when a generic one is chosen.   In the cases where they are, that's pretty much how it already works, except it's done by standard_planner().

src/test/regress/sql/multi_subquery_misc.sql
anarazel: Why?   I'm ok with the DROP FUNCTIONs, because they're a persistence resource, but deallocating things uselessly is a bad pattern.

anarazel:
On 2017-05-08 23:52:03 -0700, Önder Kalacı wrote:
> I don't have any strong opinions on this, but, I initialized it below since the function could return before using it at all. (i.e., see if `parameterId` < 0) return inputNode).

Compilers can do that automatically...

> Also, adding the comment (i.e., `parameterId starts from 1`) seemed a good fit once we initialize `parameterIndex `.
> Do you want me to update it?

Both is fine with me.

PR #1357, opened 2017-04-27
Add missing regression test output files to .gitignore

marcocitus:


mtuncer:
this was already done in subquery_main https://github.com/citusdata/citus/pull/1323

lithp:
Doing them as separate PRs feels slightly cleaner and will get this merged faster. And while this PR won't cause any additional conflicts the other PR is out of date and needs to be updated anyway, those two lines can be removed at the same time.

PR #1360, opened 2017-04-27
Don't change parse tree of DDL commands

### LINKED ISSUE #1358: PL/pgSQL function to create regular Postgres table with index errors out on its second call
**Reproducing the issue:**
```
CREATE OR REPLACE FUNCTION public.trying_to_index_me()
RETURNS VOID  AS
$BODY$
    BEGIN
    CREATE TABLE public.table_to_index ( 
        id INTEGER NOT NULL,    
        this_id UUID NOT NULL,
        that_id smallint NOT NULL,
        CONSTRAINT idx_table_to_index_unique
            UNIQUE (id,this_id,that_id)     
    ); 
    CREATE INDEX idx_table_to_index_thisthat ON public.table_to_index(this_id,that_id);  
    DROP TABLE public.table_to_index;
END;
$BODY$ LANGUAGE plpgsql;
```
**First time you run the UDF**
```
SELECT public.trying_to_index_me();
 trying_to_index_me
--------------------

(1 row)
```
**Second time you run the UDF in the same client connection**
```
SELECT public.trying_to_index_me();
ERROR:  schema "" does not exist
CONTEXT:  SQL statement "CREATE INDEX idx_table_to_index_thisthat ON public.table_to_index(this_id,that_id)"
PL/pgSQL function trying_to_index_me() line 10 at SQL statement
```

**Source from which this issue was created**
http://stackoverflow.com/questions/40247599/postgresql-error-schema-does-not-exist-on-temporary-table

**Analysis:**
The PL/pgSQL ends up reusing the IndexStmt because it always uses prepared statements.
We explicitly set the schema name in the statement in:
https://github.com/citusdata/citus/blob/master/src/backend/distributed/executor/multi_utility.c#L785

marcocitus:
Previously we scribbled on the parse tree in ProcessUtility, which may be used across multiple statements/transactions by PL/pgSQL.

Fixes #1358

### DIFF
--- a/src/backend/distributed/executor/multi_utility.c
+++ b/src/backend/distributed/executor/multi_utility.c
@@ -241,6 +241,9 @@ multi_ProcessUtility(Node *parsetree,
 
 	if (IsA(parsetree, CopyStmt))
 	{
+		/* copy parse tree since we might scribble on it to fix the schema name */
+		parsetree = copyObject(parsetree);
+
 		parsetree = ProcessCopyStmt((CopyStmt *) parsetree, completionTag,
 									&commandMustRunAsOwner);
 
@@ -272,6 +275,9 @@ multi_ProcessUtility(Node *parsetree,
 	{
 		if (IsA(parsetree, IndexStmt))
 		{
+			/* copy parse tree since we might scribble on it to fix the schema name */
+			parsetree = copyObject(parsetree);
+
 			ddlJobs = PlanIndexStmt((IndexStmt *) parsetree, queryString);
 		}
 
@@ -781,7 +787,13 @@ PlanIndexStmt(IndexStmt *createIndexStatement, const char *createIndexCommand)
 
 		isDistributedRelation = IsDistributedTable(relationId);
 
-		/* ensure future lookups hit the same relation */
+		/*
+		 * Before we do any further processing, fix the schema name to make sure
+		 * that a (distributed) table with the same name does not appear on the
+		 * search path in front of the current schema. We do this even if the
+		 * table is not distributed, since a distributed table may appear on the
+		 * search path by the time postgres starts processing this statement.
+		 */
 		namespaceName = get_namespace_name(RelationGetNamespace(relation));
 		createIndexStatement->relation->schemaname = namespaceName;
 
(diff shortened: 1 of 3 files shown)


anarazel:
IIRC we have the same pattern in a couple other places?

marcocitus:
> IIRC we have the same pattern in a couple other places?

The only other place where we scribble on the parse tree is when creating a foreign key, in which case we set `constraint->skip_validation = true;`, but that is intentional and harmless. There are a few places in `multi_utility.c` where we create a new DDL command that we scribble on, which are unrelated.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -781,20 +783,22 @@ PlanIndexStmt(IndexStmt *createIndexStatement, const char *createIndexCommand)
 
 		isDistributedRelation = IsDistributedTable(relationId);
 
-		/* ensure future lookups hit the same relation */
-		namespaceName = get_namespace_name(RelationGetNamespace(relation));
-		createIndexStatement->relation->schemaname = namespaceName;
-
-		heap_close(relation, NoLock);
```
anarazel: I don't think it's correct to remove this before the isDistributedRelation check - by the time the statement gets actually executed on the master, somebody could have created a distributed table somewhere in the search path.

marcocitus: Hm, true. Expanded the comment to not make the same mistake again.

jasonmp85:
Is there anything blocking :shipit: of this? If we wanna backport it, window's closing.

marcocitus:
I moved the `copyObject` call into the process utility hook. It's only done for `CopyStmt` and `IndexStmt` since we don't actually fix the schema name in any other statement. I'll open a separate issue for that.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -2569,20 +2581,12 @@ CreateIndexTaskList(Oid relationId, IndexStmt *indexStmt)
...
 	ListCell *shardIntervalCell = NULL;
-	Oid schemaId = get_rel_namespace(relationId);
-	char *schemaName = get_namespace_name(schemaId);
 	StringInfoData ddlString;
 	uint64 jobId = INVALID_JOB_ID;
 	int taskId = 1;
 
 	initStringInfo(&ddlString);
 
-	/* set statement's schema name if it is not set already */
```
jasonmp85: Was this just redundant, then?

marcocitus: Yes, it's always set in line 798.

PR #1361, opened 2017-04-28
Send explain queries with savepoints

### LINKED ISSUE #1344: EXPLAIN ANALYZE with DML executes write twice
I think this used to just throw an error:

```
EXPLAIN ANALYZE INSERT INTO test (x) VALUES (5);
                                                   QUERY PLAN                                                   
----------------------------------------------------------------------------------------------------------------
 Custom Scan (Citus Router)  (cost=0.00..0.00 rows=0 width=0) (actual time=5.292..5.292 rows=0 loops=1)
   Task Count: 1
   Tasks Shown: All
   ->  Task
         Node: host=localhost port=9700 dbname=postgres
         ->  Insert on test_102951  (cost=0.00..0.01 rows=1 width=8) (actual time=0.058..0.058 rows=0 loops=1)
               ->  Result  (cost=0.00..0.01 rows=1 width=8) (actual time=0.001..0.002 rows=1 loops=1)
             Planning time: 0.321 ms
             Execution time: 0.099 ms
 Planning time: 0.177 ms
 Execution time: 5.357 ms
(11 rows)

postgres=# SELECT * FROM test;
 x | y 
---+---
 5 |  
 5 |  
(2 rows)
```

metdos:
With this commit, we started to send explain queries within a savepoint. After
running explain query, we rollback to savepoint. This saves us from side effects
of EXPLAIN ANALYZE on DML queries.

Fixes #1344

### DIFF
--- a/src/backend/distributed/connection/remote_commands.c
+++ b/src/backend/distributed/connection/remote_commands.c
@@ -14,6 +14,7 @@
 
 #include "distributed/connection_management.h"
 #include "distributed/remote_commands.h"
+#include "lib/stringinfo.h"
 #include "miscadmin.h"
 #include "storage/latch.h"
 
@@ -322,6 +323,38 @@ int
 SendRemoteCommand(MultiConnection *connection, const char *command)
 {
 	return SendRemoteCommandParams(connection, command, 0, NULL, NULL);
+}
+
+
+/*
+ * ReadFirstColumnAsText reads the first column of result tuples from the given
+ * PGresult struct and returns them in a StringInfo list.
+ */
+List *
+ReadFirstColumnAsText(PGresult *queryResult)
+{
+	List *resultRowList = NIL;
+	const int columnIndex = 0;
+	int64 rowIndex = 0;
+	int64 rowCount = 0;
+
+	ExecStatusType status = PQresultStatus(queryResult);
+	if (status == PGRES_TUPLES_OK)
+	{
+		rowCount = PQntuples(queryResult);
+	}
+
+	for (rowIndex = 0; rowIndex < rowCount; rowIndex++)
+	{
+		char *rowValue = PQgetvalue(queryResult, rowIndex, columnIndex);
+
+		StringInfo rowValueString = makeStringInfo();
+		appendStringInfoString(rowValueString, rowValue);
+
+		resultRowList = lappend(resultRowList, rowValueString);
+	}
+
+	return resultRowList;
 }
 
 
--- a/src/backend/distributed/planner/multi_explain.c
+++ b/src/backend/distributed/planner/multi_explain.c
@@ -8,6 +8,7 @@
  */
 
 #include "postgres.h"
+#include "libpq-fe.h"
 #include "miscadmin.h"
 
 #include "access/xact.h"
@@ -20,6 +21,7 @@
 #include "commands/tablecmds.h"
 #include "optimizer/cost.h"
 #include "distributed/citus_nodefuncs.h"
+#include "distributed/connection_management.h"
 #include "distributed/multi_client_executor.h"
 #include "distributed/multi_executor.h"
 #include "distributed/multi_explain.h"
@@ -29,6 +31,8 @@
 #include "distributed/multi_physical_planner.h"
 #include "distributed/multi_planner.h"
 #include "distributed/multi_server_executor.h"
+#include "distributed/remote_commands.h"
+#include "distributed/placement_connection.h"
 #include "distributed/worker_protocol.h"
 #include "lib/stringinfo.h"
 #include "nodes/plannodes.h"
@@ -288,15 +292,60 @@ RemoteExplain(Task *task, ExplainState *es)
 	remotePlan = (RemoteExplainPlan *) palloc0(sizeof(RemoteExplainPlan));
 	explainQuery = BuildRemoteExplainQuery(task->queryString, es);
 
+	/*
+	 * Use a coordinated transaction to ensure that we open a transaction block
+	 * such that we can set a savepoint.
+	 */
+	BeginOrContinueCoordinatedTransaction();
+
 	for (placementIndex = 0; placementIndex < placementCount; placementIndex++)
 	{
 		ShardPlacement *taskPlacement = list_nth(taskPlacementList, placementIndex);
-		char *nodeName = taskPlacement->nodeName;
-		uint32 nodePort = taskPlacement->nodePort;
+		MultiConnection *connection = NULL;
+		PGresult *queryResult = NULL;
+		int connectionFlags = 0;
+		int executeResult = 0;
 
 		remotePlan->placementIndex = placementIndex;
-		remotePlan->explainOutputList = ExecuteRemoteQuery(nodeName, nodePort,
-														   NULL, explainQuery);
+
+		connection = GetPlacementConnection(connectionFlags, taskPlacement, NULL);
+
+		/* try other placements if we fail to connect this one */
+		if (PQstatus(connection->pgConn) != CONNECTION_OK)
+		{
+			continue;
+		}
+
+		RemoteTransactionBeginIfNecessary(connection);
+
+		/*
+		 * Start a savepoint for the explain query. After running the explain
+		 * query, we will rollback to this savepoint. This saves us from side
+		 * effects of EXPLAIN ANALYZE on DML queries.
+		 */
+		ExecuteCriticalRemoteCommand(connection, "SAVEPOINT citus_explain_savepoint");
+
+		/* run explain query */
+		executeResult = ExecuteOptionalRemoteCommand(connection, explainQuery->data,
+													 &queryResult);
+		if (executeResult != 0)
+		{
+			PQclear(queryResult);
+			ForgetResults(connection);
+
+			continue;
+		}
+
+		/* read explain query results */
+		remotePlan->explainOutputList = ReadFirstColumnAsText(queryResult);
+
+		PQclear(queryResult);
+		ForgetResults(connection);
+
+		/* rollback to the savepoint */
+		ExecuteCriticalRemoteCommand(connection,
+									 "ROLLBACK TO SAVEPOINT citus_explain_savepoint");
+
 		if (remotePlan->explainOutputList != NIL)
 		{
 			break;
--- a/src/test/regress/expected/multi_explain.out
+++ b/src/test/regress/expected/multi_explain.out
@@ -669,3 +669,20 @@ Custom Scan (Citus Router)  (cost=0.00..0.00 rows=0 width=0)
         Node: host=localhost port=57637 dbname=regression
         ->  Index Scan using lineitem_pkey_290000 on lineitem_290000 lineitem  (cost=0.28..11.83 rows=3 width=5)
               Index Cond: (l_orderkey = 5)
+-- test explain in a transaction with alter table to test we use right connections
+BEGIN;
+CREATE TABLE explain_table(id int);
+SELECT create_distributed_table('explain_table', 'id');
+
+ALTER TABLE explain_table ADD COLUMN value int;
+NOTICE:  using one-phase commit for distributed DDL commands
+HINT:  You can enable two-phase commit for extra safety with: SET citus.multi_shard_commit_protocol TO '2pc'
+EXPLAIN (COSTS FALSE) SELECT value FROM explain_table WHERE id = 1;
+Custom Scan (Citus Router)
+  Task Count: 1
+  Tasks Shown: All
+  ->  Task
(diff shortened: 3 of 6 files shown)


metdos:
I couldn't add regression tests because EXPLAIN ANALYZE prints time even though I switch off costs and time.

```sql
postgres=# EXPLAIN (ANALYZE TRUE, VERBOSE FALSE, COSTS FALSE, BUFFERS FALSE, TIMING FALSE, FORMAT TEXT) INSERT INTO public.explain_table_102070 (id) VALUES (995);
                       QUERY PLAN
--------------------------------------------------------
 Insert on explain_table_102070 (actual rows=0 loops=1)
   ->  Result (actual rows=1 loops=1)
 Planning time: 0.144 ms
 Execution time: 0.086 ms
(4 rows)
```

Here are some tests I run;

```SQL
CREATE TABLE explain_table (id int, value int);
SELECT create_distributed_table('explain_table', 'id');

-- test explain analyze with DML queries
EXPLAIN ANALYZE INSERT INTO explain_table VALUES (1, 0);
SELECT * FROM explain_table ORDER BY id, value;

EXPLAIN ANALYZE UPDATE explain_table SET value = 1 WHERE id = 1;
SELECT * FROM explain_table ORDER BY id, value;

-- DELETE is a bit tricky because we first DELETE rows then run the explain query
EXPLAIN ANALYZE DELETE FROM explain_table WHERE id = 1;
SELECT * FROM explain_table ORDER BY id, value;

-- test explain within a transaction
BEGIN;
EXPLAIN ANALYZE INSERT INTO explain_table VALUES (1, 0);
COMMIT;
SELECT * FROM explain_table ORDER BY id, value;

BEGIN;
EXPLAIN ANALYZE INSERT INTO explain_table VALUES (2, 0);
ROLLBACK;
SELECT * FROM explain_table ORDER BY id, value;

BEGIN;
INSERT INTO explain_table VALUES (3, 0);
EXPLAIN ANALYZE INSERT INTO explain_table VALUES (3, 0);
EXPLAIN ANALYZE INSERT INTO explain_table VALUES (3, 0);
INSERT INTO explain_table VALUES (3, 0);
COMMIT;
SELECT * FROM explain_table ORDER BY id, value;

DROP TABLE explain_table;
```

I run a couple of next scripts in parallel to test concurrency.
```bash
for i in `seq 1 1000`;
do
    /usr/local/pgsql/bin/psql -d postgres -c  "EXPLAIN ANALYZE INSERT INTO explain_table VALUES ($i);"
done
```

metdos:
@anarazel, could you take a look when you have time?

src/backend/distributed/connection/remote_commands.c
```diff
@@ -326,6 +328,38 @@ SendRemoteCommand(MultiConnection *connection, const char *command)
...
+
+	ExecStatusType status = PQresultStatus(queryResult);
+	if (status == PGRES_TUPLES_OK)
+	{
+		rowCount = PQntuples(queryResult);
+	}
+
+	for (rowIndex = 0; rowIndex < rowCount; rowIndex++)
+	{
+		char *rowValue = MultiClientGetValue(queryResult, rowIndex, columnIndex);
```
marcocitus: This is just a thin (deprecated) wrapper around PQgetvalue, so maybe just call that directly.

src/backend/distributed/planner/multi_explain.c
```diff
@@ -293,10 +296,60 @@ RemoteExplain(Task *task, ExplainState *es)
...
+
+		/*
+		 * Use a coordinated transaction to skip sending BEGIN in an already
+		 * existing transaction.
+		 */
+		BeginOrContinueCoordinatedTransaction();
+		RemoteTransactionBeginIfNecessary(connection);
+
+		/* start a savepoint for explain */
+		executeResult = ExecuteOptionalRemoteCommand(connection,
```
marcocitus: Why optional? We just want to roll back on failure.

marcocitus: Never mind, we don't :)

src/backend/distributed/planner/multi_explain.c
```diff
@@ -293,10 +296,60 @@ RemoteExplain(Task *task, ExplainState *es)
...
-		remotePlan->explainOutputList = ExecuteRemoteQuery(nodeName, nodePort,
-														   NULL, explainQuery);
+
+		connection = GetNodeConnection(connectionFlags, nodeName, nodePort);
+
+		/*
+		 * Use a coordinated transaction to skip sending BEGIN in an already
+		 * existing transaction.
+		 */
+		BeginOrContinueCoordinatedTransaction();
```
marcocitus: Maybe move this to before the loop.

src/backend/distributed/planner/multi_explain.c
```diff
@@ -293,10 +296,60 @@ RemoteExplain(Task *task, ExplainState *es)
...
+		MultiConnection *connection = NULL;
+		PGresult *queryResult = NULL;
+		int connectionFlags = SESSION_LIFESPAN;
+		int executeResult = 0;
 
 		remotePlan->placementIndex = placementIndex;
-		remotePlan->explainOutputList = ExecuteRemoteQuery(nodeName, nodePort,
-														   NULL, explainQuery);
+
+		connection = GetNodeConnection(connectionFlags, nodeName, nodePort);
```
marcocitus: We should probably use GetPlacementConnection here. We might create a deadlock if we run an EXPLAIN on the wrong connection after a DDL.

src/backend/distributed/connection/remote_commands.c
metdos: Good point!

src/backend/distributed/planner/multi_explain.c
marcocitus: On the other hand, the code would be a bit cleaner here if we used ExecuteCriticalRemoteCommand. If we add `if (connection->pgConn) continue;` after `GetPlacementConnection` then we still get failover if the worker is down.

src/backend/distributed/connection/remote_commands.c
```diff
@@ -326,6 +327,38 @@ SendRemoteCommand(MultiConnection *connection, const char *command)
 
 
 /*
+ * ReadRemoteQueryResults reads first columns of result tuples from the given
+ * PGresult struct and returns them in a StringInfo list.
+ */
+List *
+ReadRemoteQueryResults(PGresult *queryResult)
```
marcocitus: To be more descriptive, maybe something like `ReadColumnAsText`

src/backend/distributed/planner/multi_explain.c
```diff
@@ -310,25 +310,20 @@ RemoteExplain(Task *task, ExplainState *es)
...
 		}
 
+		RemoteTransactionBeginIfNecessary(connection);
+
+		/* start a savepoint for explain */
+		ExecuteCriticalRemoteCommand(connection, "SAVEPOINT citus_explain_savepoint");
+
 		/* run explain query */
-		queryResult = NULL;
 		executeResult = ExecuteOptionalRemoteCommand(connection, explainQuery->data,
```
marcocitus: Probably makes sense to use ExecuteCriticalRemoteCommand here too.

src/backend/distributed/planner/multi_explain.c
```diff
@@ -288,15 +292,56 @@ RemoteExplain(Task *task, ExplainState *es)
...
+	BeginOrContinueCoordinatedTransaction();
+
 	for (placementIndex = 0; placementIndex < placementCount; placementIndex++)
 	{
 		ShardPlacement *taskPlacement = list_nth(taskPlacementList, placementIndex);
-		char *nodeName = taskPlacement->nodeName;
-		uint32 nodePort = taskPlacement->nodePort;
+		MultiConnection *connection = NULL;
+		PGresult *queryResult = NULL;
+		int connectionFlags = SESSION_LIFESPAN;
```
marcocitus: This should be 0.

src/backend/distributed/planner/multi_explain.c
```diff
@@ -288,15 +292,56 @@ RemoteExplain(Task *task, ExplainState *es)
...
+
+		/* try other placements if we fail to connect this one */
+		if (PQstatus(connection->pgConn) != CONNECTION_OK)
+		{
+			continue;
+		}
+
+		RemoteTransactionBeginIfNecessary(connection);
+
+		/* start a savepoint for explain */
```
marcocitus: Maybe elaborate on why we set a savepoint.

metdos:
@marcocitus, do you have any further comments?

anarazel:
Could you add a test that does something like
```
BEGIN;
CREATE TABLE whatnot(parcol int primary key);
create_distributed_table(...)
EXPLAIN SELECT count(*) FROM whatnot;
```
to make sure we use the right connection?

marcocitus:
`create_distributed_table` doesn't use the connection API.

marcocitus:
Hm, well, maybe if you put in some data first and then do an explain analyze to check the row count. The data loading does use the connection API.

anarazel:
On 2017-04-28 10:30:09 -0700, Marco Slot wrote:
> `create_distributed_table` doesn't use the connection API.

GAH, we should fix that, right.

marcocitus:
I guess this works:

```
BEGIN;
EXPLAIN (ANALYZE, COSTS FALSE) SELECT * FROM test WHERE x = 1;
                                     QUERY PLAN                                     
------------------------------------------------------------------------------------
 Custom Scan (Citus Router) (actual time=0.848..0.849 rows=1 loops=1)
   Task Count: 1
   Tasks Shown: All
   ->  Task
         Node: host=localhost port=9701 dbname=postgres
         ->  Seq Scan on test_103138 test (actual time=0.004..0.004 rows=1 loops=1)
               Filter: (x = 1)
             Planning time: 0.221 ms
             Execution time: 0.019 ms
 Planning time: 1.087 ms
 Execution time: 0.946 ms
(11 rows)

```

marcocitus:
(this shouldn't be a regression test, given planning/execution time)

marcocitus:
ok, this appears to work and would deadlock if it picked the wrong connection:

```

postgres=# BEGIN;
BEGIN
Time: 0.315 ms
postgres=# ALTER TABLE test ADD COLUMN z int;
NOTICE:  using one-phase commit for distributed DDL commands
HINT:  You can enable two-phase commit for extra safety with: SET citus.multi_shard_commit_protocol TO '2pc'
ALTER TABLE
Time: 68.545 ms
postgres=# EXPLAIN (COSTS OFF) SELECT z FROM test WHERE x = 1; 
                       QUERY PLAN                       
--------------------------------------------------------
 Custom Scan (Citus Router)
   Task Count: 1
   Tasks Shown: All
   ->  Task
         Node: host=localhost port=9701 dbname=postgres
         ->  Seq Scan on test_103234 test
               Filter: (x = 1)
(7 rows)

Time: 2.477 ms
```

metdos:
@marcocitus, @anarazel, I added the regression test and squashed commits. If this looks good, I will merge it.

src/test/regress/expected/multi_explain.out
```diff
@@ -669,3 +669,21 @@ Custom Scan (Citus Router)  (cost=0.00..0.00 rows=0 width=0)
...
+EXPLAIN (COSTS FALSE) SELECT value FROM explain_table WHERE id = 1;
+Custom Scan (Citus Router)
+  Task Count: 1
+  Tasks Shown: All
+  ->  Task
+        Node: host=localhost port=57637 dbname=regression
+        ->  Index Scan using explain_table_pkey_570001 on explain_table_570001 explain_table
+              Index Cond: (id = 1)
+COMMIT;
+DROP TABLE explain_table;
```
marcocitus: Would be fine to just ABORT here.

src/test/regress/expected/multi_explain.out
```diff
@@ -669,3 +669,21 @@ Custom Scan (Citus Router)  (cost=0.00..0.00 rows=0 width=0)
...
+ALTER TABLE explain_table ADD COLUMN value int;
+NOTICE:  using one-phase commit for distributed DDL commands
+HINT:  You can enable two-phase commit for extra safety with: SET citus.multi_shard_commit_protocol TO '2pc'
+EXPLAIN (COSTS FALSE) SELECT value FROM explain_table WHERE id = 1;
+Custom Scan (Citus Router)
+  Task Count: 1
+  Tasks Shown: All
+  ->  Task
+        Node: host=localhost port=57637 dbname=regression
+        ->  Index Scan using explain_table_pkey_570001 on explain_table_570001 explain_table
```
marcocitus: I wonder if this can every be a sequential scan due to the small table size. Could consider removing the primary key.

PR #1363, opened 2017-04-28
Fix crash in isolation tests

lithp:
- There was a crash when the table a shardid belonged to changed during
  a session. Instead of crashing (a failed assert) we now throw an error
- Update the isolation test which was crashing to no longer exercise
  that code path
- Add a regression test to check that the error is thrown

Fixes https://github.com/citusdata/citus/issues/1336


src/backend/distributed/utils/metadata_cache.c
```diff
@@ -745,7 +745,13 @@ BuildCachedShardList(DistTableCacheEntry *cacheEntry)
 
 		shardEntry = hash_search(DistShardCacheHash, &shardInterval->shardId, HASH_ENTER,
 								 &foundInCache);
-		Assert(!foundInCache);
+		if (foundInCache)
+		{
+			ereport(ERROR, (errmsg("a shard has changed the table it belongs to within "
+								   "this session"),
+							errhint("try again in a new session")));
```
anarazel: Hints are supposed to be actual sentences:
https://www.postgresql.org/docs/devel/static/error-style-guide.html
maybe something like "Reconnect and try again."

Also, I'd be a bit less precise about the error message, there could be other reasons.  How about "cached metadata for relation %u, shard %u is inconsistent"?

lithp:
@anarazel pushed an update

PR #1376, opened 2017-05-02
Skip exhaustive test in CoPartitionedTables() if declared colocated

onderkalaci:
That's considerably cheaper.

(Note that due to some merge conflicts, we accidentally undo the changes in #1354 and applying back to the master.)


metdos:
🚢

PR #1377, opened 2017-05-02
Make minor tweaks to prepare for PostgreSQL 10

jasonmp85:
These changes are anything that I deemed "safely" immediately mergeable to move us towards PostgreSQL 10 compatibility. There are no new version-specific tests, no `#if` blocks based on version, etc., just changes that work in PostgreSQL 9.x that _also_ work well in PostgreSQL 10.

The commit history has full details: each commit will have a small description of what the change is and why it was made.


src/backend/distributed/worker/worker_data_fetch_protocol.c
```diff
@@ -482,6 +482,7 @@ worker_apply_sequence_command(PG_FUNCTION_ARGS)
 	/* run the CREATE SEQUENCE command */
 	ProcessUtility(commandNode, commandString, PROCESS_UTILITY_TOPLEVEL,
 				   NULL, None_Receiver, NULL);
+	CommandCounterIncrement();
```
lithp: What's specific to PG 10 about this? And, is this something that should include a test?

lithp:
This looks pretty painstaking, thanks for going through all this.

src/backend/distributed/worker/worker_data_fetch_protocol.c
jasonmp85: Read the commit message; basically we create a sequence and then expect it to be visible. In PostgreSQL 10 it isn't yet, and there could be two reasons why:

  * We made some other call in Citus that pre-PostgreSQL 10 called CCI but no longer does, implicitly relying on command counter increment to make our change visible
  * Due to [changes in how PostgreSQL's stores and accesses sequence attributes](https://www.postgresql.org/message-id/flat/CAB7nPqSJB1h535AcQ%2BZOyqM0Sg6VORf0hL_Vrw96uvajR03hQg%40mail.gmail.com#CAB7nPqSJB1h535AcQ+ZOyqM0Sg6VORf0hL_Vrw96uvajR03hQg@mail.gmail.com), the newly created sequence was previously visible, but now that we have to access it through `SearchSysCache` we must increment the command counter to see it

jasonmp85: At any rate, adding the CCI call fixes it (I think the latter hypothesis is the cause), and @anarazel said we should probably have had it in the first place.

jasonmp85:
> This looks pretty painstaking

Just wait till you see #1379…

src/backend/distributed/worker/worker_data_fetch_protocol.c
lithp: Thanks for the thorough explanation :+1:, too bad this doesn't change behavior on 9.6 so we can't add a test for it.

PR #1382, opened 2017-05-03
Fix locking in master_drop_all_shards / master_apply_delete_command

### LINKED ISSUE #727: master_apply_delete_command() calls don't block each other
Create multiple shards

``` sql
SET citus.shard_max_size to '100MB';
COPY lineitem from '/home/ubuntu/tpch_2_13_0/lineitem.tbl' WITH DELIMITER '|';
```

Run **two** master_apply_delete_command() concurrently

``` sql
SELECT * from master_apply_delete_command('DELETE FROM lineitem');
```

Observe that one of the calls will fail;

```
ERROR:  could not find valid entry for relationId: 34972 and shard 110613
```

Minor Edit: Clarify that we run **two** master_apply_delete_command() concurrently.

### LINKED ISSUE #761: Direct use of SET_LOCKTAG_ADVISORY in master_apply_delete_command() a bad idea
master_apply_delete_command() does, without further docs that I can see

```
    /* acquire lock */
    SET_LOCKTAG_ADVISORY(lockTag, MyDatabaseId, relationId, 0, 0);
    LockAcquire(&lockTag, ExclusiveLock, sessionLock, dontWait);
```

there's no actual document about what that that lock actually tries to accomplish, and what the semantics are. I think this should use the infrastructure from resource_lock.[ch], and wrap the lock acquisition in a function with a name that actually explains what's being locked here.

### LINKED ISSUE #1381: Truncate fails if shards are dropped concurrently
Looks like truncate command pre-fetches shards to be dropped before it was blocked by concurrent transaction.  Shards are already dropped and committed by the time the first worker receives truncate command. The command fails and returns shard does not exist error.

## Steps
- 2 concurrent connections are opened. 
- connection 1 starts a transaction and issues drop all shards command
- before connection 1 commits, connection 2 issues truncate table command.

### Connection 1
```sql
begin;
select master_drop_all_shards('drop_test'::regclass, 'public', 'drop_test');
commit;
```

### Connection 2
```sql
truncate table drop_test;
```

Expected
- connection 2 to be blocked until connection 1 commits
- truncate command to complete silently since there is nothing left to truncate

Result
- connection 2 to be blocked until connection 1 commits ✅ 
- truncate command returned error stating some shard cannot be found ❌ 
```
ERROR:  relation "public.drop_test_102040" does not exist
```

marcocitus:
This PR adds the appropriate table locks to `master_drop_all_shards` and `master_apply_delete_command` and ensures that they can safely be executed concurrently through isolation tests.

It does not prevent `master_apply_delete_command` from running concurrently with `SELECT` since users might rely on the current behaviour. The long-term solution there is to take shard metadata locks on `SELECT`.

Fixes #1381 
Fixes #727
Fixes #761

### DIFF
--- a/src/test/regress/specs/isolation_drop_shards.spec
+++ b/src/test/regress/specs/isolation_drop_shards.spec
@@ -1,0 +1,71 @@
+setup
+{
+	CREATE TABLE append_table (test_id integer NOT NULL, data text);
+	SELECT create_distributed_table('append_table', 'test_id', 'append');
+
+	SELECT 1 FROM (
+		SELECT min(master_create_empty_shard('append_table')) FROM generate_series(1,16)
+	) a;
+}
+
+teardown
+{
+	DROP TABLE append_table;
+}
+
+session "s1"
+
+step "s1-begin"
+{
+	BEGIN;
+}
+
+step "s1-apply-delete-command"
+{
+	SELECT master_apply_delete_command('DELETE FROM append_table');
+}
+
+step "s1-drop-all-shards"
+{
+	SELECT master_drop_all_shards('append_table', 'public', 'append_table');
+}
+
+step "s1-commit"
+{
+	COMMIT;
+}
+
+session "s2"
+
+step "s2-truncate"
+{
+	TRUNCATE append_table;
+}
+
+step "s2-apply-delete-command"
+{
+	SELECT master_apply_delete_command('DELETE FROM append_table');
+}
+
+step "s2-drop-all-shards"
+{
+	SELECT master_drop_all_shards('append_table', 'public', 'append_table');
+}
+
+step "s2-select"
+{
+	SELECT * FROM append_table;
+}
+
+permutation "s1-begin" "s1-drop-all-shards" "s2-truncate" "s1-commit"
+permutation "s1-begin" "s1-drop-all-shards" "s2-apply-delete-command" "s1-commit"
+permutation "s1-begin" "s1-drop-all-shards" "s2-drop-all-shards" "s1-commit"
+permutation "s1-begin" "s1-drop-all-shards" "s2-select" "s1-commit"
+
+permutation "s1-begin" "s1-apply-delete-command" "s2-truncate" "s1-commit"
+permutation "s1-begin" "s1-apply-delete-command" "s2-apply-delete-command" "s1-commit"
+permutation "s1-begin" "s1-apply-delete-command" "s2-drop-all-shards" "s1-commit"
+
+# We can't verify master_apply_delete_command + SELECT since it blocks on the
+# the workers, but this is not visible on the master, meaning the isolation
+# test cannot proceed.
--- a/src/test/regress/specs/isolation_drop_shards.spec
+++ b/src/test/regress/specs/isolation_drop_shards.spec
@@ -20,6 +20,11 @@ step "s1-begin"
 	BEGIN;
 }
 
+step "s1-truncate"
+{
+	TRUNCATE append_table;
+}
+
 step "s1-apply-delete-command"
 {
 	SELECT master_apply_delete_command('DELETE FROM append_table');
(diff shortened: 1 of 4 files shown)


src/test/regress/specs/isolation_drop_shards.spec
```diff
@@ -0,0 +1,60 @@
...
+{
+    SELECT master_apply_delete_command('DELETE FROM append_table');
+}
+
+step "s2-drop-all-shards"
+{
+    SELECT master_drop_all_shards('append_table', 'public', 'append_table');
+}
+
+permutation "s1-begin" "s1-apply-delete-command" "s2-truncate" "s1-commit"
```
onderkalaci: This looks neat! I didn't know that it is so easy to write isolation tests.

src/test/regress/specs/isolation_drop_shards.spec
```diff
@@ -0,0 +1,60 @@
...
+step "s2-apply-delete-command"
+{
+    SELECT master_apply_delete_command('DELETE FROM append_table');
+}
+
+step "s2-drop-all-shards"
+{
+    SELECT master_drop_all_shards('append_table', 'public', 'append_table');
+}
+
```
anarazel: Maybe add a read-access test too?  Possibly once with and once without xact?

marcocitus: OK. We can do it for `master_drop_all_shards`, though not for `master_apply_delete_command`, since SELECTs are allowed to proceed and then block on the worker, which means the isolation test doesn't proceed.

src/test/regress/specs/isolation_drop_shards.spec
```diff
@@ -0,0 +1,71 @@
...
+permutation "s1-begin" "s1-drop-all-shards" "s2-truncate" "s1-commit"
+permutation "s1-begin" "s1-drop-all-shards" "s2-apply-delete-command" "s1-commit"
+permutation "s1-begin" "s1-drop-all-shards" "s2-drop-all-shards" "s1-commit"
+permutation "s1-begin" "s1-drop-all-shards" "s2-select" "s1-commit"
+
+permutation "s1-begin" "s1-apply-delete-command" "s2-truncate" "s1-commit"
+permutation "s1-begin" "s1-apply-delete-command" "s2-apply-delete-command" "s1-commit"
+permutation "s1-begin" "s1-apply-delete-command" "s2-drop-all-shards" "s1-commit"
+
+# We can't verify master_apply_delete_command + SELECT since it blocks on the
```
metdos: Shall we also a set of permutations start with truncate?

src/test/regress/specs/isolation_drop_shards.spec
```diff
@@ -0,0 +1,71 @@
...
+
+step "s2-select"
+{
+	SELECT * FROM append_table;
+}
+
+permutation "s1-begin" "s1-drop-all-shards" "s2-truncate" "s1-commit"
+permutation "s1-begin" "s1-drop-all-shards" "s2-apply-delete-command" "s1-commit"
+permutation "s1-begin" "s1-drop-all-shards" "s2-drop-all-shards" "s1-commit"
+permutation "s1-begin" "s1-drop-all-shards" "s2-select" "s1-commit"
```
metdos: I don't know why - but I can't see results for this line (s2-select) in the expected file.

metdos: Aha. The reason is that the expected file is wrong (at least it fails at my local tests). We haven't merged this https://github.com/citusdata/citus/pull/1343, so Travis does not test isolation tests yet.

metdos:
Hey @marcocitus, code changes look good. I tested the failing scenarios. I updated the outdated expected file and also added truncate first isolation tests - feel free to update/squash them.

Overall, I think we can 🚢  this.

marcocitus:
@metdos Thanks!

PR #1385, opened 2017-05-04
Refactor COPY to not directly use cache entry (v6.1)

jasonmp85:
Backport of the fix from #1353.


jasonmp85:
FWIW I don't want to use the "Merge" strategy to close this PR, but rather the "Rebase and merge" one, to better emulate cherry-picking this stuff onto `release-6.1`.

PR #1387, opened 2017-05-04
Fix pushing down wrong INSERT ... SELECT queries (v6.1)

jasonmp85:
Backport of the fix from #1261


onderkalaci:
Thanks! I was planning to do it now.

jasonmp85:
No problem; I thought there were conflicts in the C code, but after looking over it it was straightforward. There were conflicts throughout the test file, but I just ran tests, copied the output, and diffed against master's version of the test output. All changes were related to standard `DEBUG` output that had changed, nothing meaningful, so I think it's good.

PR #1389, opened 2017-05-04
Improve subquery pushdown regression tests

### LINKED ISSUE #1373: Improve subquery_multi regression test file
* Create composite keys related comparison functions using native postgresql functions. See examples on `multi_data_types.sql`
* Use citus tools to create functions on the workers/master
* Move table creation/data loading to another file to get rid of input/output files.

onderkalaci:
- Use native postgres function for composite key btree functions (see `multi_data_types.sql`)
  - Use Citus tools to create functions on the workers
  - Move explain tests to multi_explain.sql (get rid of .out _0.out files)
  - Get rid of input/output files for multi_subquery.sql by moving table creations

Fixes #1373


onderkalaci:
@mtuncer Would you mind reviewing this PR which only includes moving/refactoring of regression tests? Thanks!

src/test/regress/output/multi_insert_select_behavioral_analytics_create_table.source
```diff
@@ -62,3 +62,465 @@ CREATE INDEX is_index3 ON users_table(value_1);
...
+);
+-- do it on the workers as well
+SELECT run_command_on_workers('
+
+	CREATE TYPE user_composite_type AS
+	(
+		tenant_id BIGINT,
+		user_id BIGINT
+	);
+');
```
anarazel: Not for this, but I think we should have a function that executes stuff both on master and workers...

src/test/regress/input/multi_insert_select_behavioral_analytics_create_table.source
```diff
@@ -41,3 +41,425 @@ CREATE INDEX is_index4 ON events_table(event_type);
...
+
+CREATE FUNCTION cmp_user_composite_type_function(user_composite_type, user_composite_type) RETURNS int
+LANGUAGE 'internal'
+AS 'btrecordcmp'
+IMMUTABLE
+RETURNS NULL ON NULL INPUT;
+
+-- do it on the workers as well
+SELECT run_command_on_workers('
+
```
anarazel: You should use `$f$worker_sql$f$` or such, then you don't need to manually double quotes with `''`

src/test/regress/input/multi_insert_select_behavioral_analytics_create_table.source
```diff
@@ -41,3 +41,425 @@ CREATE INDEX is_index4 ON events_table(event_type);
...
+	CREATE FUNCTION cmp_user_composite_type_function(user_composite_type, user_composite_type) RETURNS int
+	LANGUAGE ''internal''
+	AS ''btrecordcmp''
+	IMMUTABLE
+	RETURNS NULL ON NULL INPUT;
+');
```
anarazel: I'd really create a wrapper function for `run_command_on_workers()` that runs both on master and worker.  Should be like three lines
```sql
CREATE FUNCTION run_command_on_master_and_workers(p_sql text)
RETURNS void LANGUAGE plpgsql AS $$
BEGIN
     EXECUTE p_sql;
     PERFORM run_command_on_workers(p_sql);
END;$$;
```

lfittl: FWIW, something like `run_command_on_coordinator_and_workers` could also be useful for customers when creating e.g. ENUM types for use in table definitions.

src/test/regress/output/multi_insert_select_behavioral_analytics_create_table.source
```diff
@@ -62,3 +62,370 @@ CREATE INDEX is_index3 ON users_table(value_1);
 CREATE INDEX is_index4 ON events_table(event_type);
 CREATE INDEX is_index5 ON users_table(value_2);
 CREATE INDEX is_index6 ON events_table(value_2);
+-- create a helper function to create types/functions on each node
+CREATE FUNCTION run_command_on_master_and_workers(p_sql text)
```
metdos: Isn't this same with `run_command_on_coordinator_and_workers ()`? Maybe we can move it to a more central place such citus tools or another base regression tests and use it?

https://github.com/citusdata/citus/blob/40ebd93be3b0140155dc4d24836033d0d917b542/src/test/regress/sql/multi_schema_support.sql#L742

onderkalaci: I don't want to create such dependencies among the tests. 

I think in one day we should add this function to `citus_tools`. Also, if we one day decide to improve the regression tests such that they all share some common tests which are responsible for creating tables/types/functions etc., I'd move this function.

metdos: We already use it in two different places. Why create a technical debt while we can fix it easily now?

Also having different names for the same functionality is not right.

onderkalaci: > We already use it in two different places. Why create a technical debt while we can fix it easily now?

I don't think that it fixes anything. If I were to use the same function here, I'd have created an order dependency between that test and this test. The proper fix is to add this function to `citustools`, which I don't want to do on this PR.

> Also having different names for the same functionality is not right.

That was intentional. I gave a separate name to prevent error in case we run the tests concurrently.

metdos:
> That was intentional. I gave a separate name to prevent error in case we run the tests concurrently.

I think this is very hacky :)

src/test/regress/input/multi_behavioral_analytics_create_table.source
```diff
@@ -41,3 +41,328 @@ CREATE INDEX is_index4 ON events_table(event_type);
...
+	o_orderdate date not null,
+	o_orderpriority char(15) not null,
+	o_clerk char(15) not null,
+	o_shippriority integer not null,
+	o_comment varchar(79) not null,
+	PRIMARY KEY(o_orderkey) );
+SELECT master_create_distributed_table('orders_subquery', 'o_orderkey', 'range');
+
+SET citus.enable_router_execution TO 'false';
+
```
mtuncer: extra spaces should be reduced

src/test/regress/input/multi_behavioral_analytics_create_table.source
```diff
@@ -41,3 +41,328 @@ CREATE INDEX is_index4 ON events_table(event_type);
...
+
+
+-- Check that we don't crash if there are not any shards.
+
```
mtuncer: remove extra space

src/test/regress/input/multi_behavioral_analytics_create_table.source
```diff
@@ -41,3 +41,328 @@ CREATE INDEX is_index4 ON events_table(event_type);
...
+SET citus.shard_max_size TO "1MB";
+
+\copy lineitem_subquery FROM '@abs_srcdir@/data/lineitem.1.data' with delimiter '|'
+\copy lineitem_subquery FROM '@abs_srcdir@/data/lineitem.2.data' with delimiter '|'
+
+\copy orders_subquery FROM '@abs_srcdir@/data/orders.1.data' with delimiter '|'
+\copy orders_subquery FROM '@abs_srcdir@/data/orders.2.data' with delimiter '|'
+
+DROP FUNCTION run_command_on_master_and_workers(p_sql text);
```
mtuncer: needs a newline at the end

src/test/regress/sql/multi_explain.sql
```diff
@@ -126,6 +126,224 @@ EXPLAIN (COSTS FALSE, VERBOSE TRUE)
 	GROUP BY l_quantity
 	HAVING l_quantity > (100 * random());
 
+
```
mtuncer: remove extra space

src/test/regress/input/multi_behavioral_analytics_create_table.source
```diff
@@ -41,3 +41,328 @@ CREATE INDEX is_index4 ON events_table(event_type);
...
+	OPERATOR 3 = (user_composite_type, user_composite_type),
+	OPERATOR 4 >= (user_composite_type, user_composite_type),
+	OPERATOR 5 > (user_composite_type, user_composite_type),
+	
+	FUNCTION 1 cmp_user_composite_type_function(user_composite_type, user_composite_type);
+$f$);
+
+SELECT run_command_on_master_and_workers($f$
+	
+	CREATE OPERATOR CLASS cats_2_op_fam_class
```
mtuncer: shall we worry about dropping these types, functions, and operators after subquery tests ?

src/test/regress/sql/multi_subquery.sql
```diff
@@ -0,0 +1,521 @@
...
+   b int
+);
+
+SELECT master_create_distributed_table('subquery_pruning_varchar_test_table', 'a', 'hash');
+SELECT master_create_worker_shards('subquery_pruning_varchar_test_table', 4, 1);
+
+SET client_min_messages TO DEBUG2;
+
+SELECT * FROM
+	(SELECT count(*) FROM subquery_pruning_varchar_test_table WHERE a = 'onder' GROUP BY a)
```
mtuncer: can we use made up names instead of real names here ?

src/test/regress/input/multi_behavioral_analytics_create_table.source
onderkalaci: Hmm, that makes sense.

src/test/regress/sql/multi_explain.sql
jasonmp85: How did you manage to add this file?!? It's in `.gitignore`…

src/test/regress/sql/multi_subquery.sql
```diff
@@ -0,0 +1,538 @@
...
+		'Has done paying'::TEXT AS hasdone
+	FROM
+		events
+	WHERE
+		events.composite_id >= '(1, -9223372036854775808)'::user_composite_type AND
+		events.composite_id <= '(1, 9223372036854775807)'::user_composite_type AND
+		event_type = 'pay') AS subquery_2
+	ON
+		subquery_1.composite_id = subquery_2.composite_id
+	GROUP BY
```
jasonmp85: I tried to excise all unordered queries in #1377, but it appears this change introduces quite a few more. In the future, try to find something to `ORDER` queries by so their output is deterministic (this query changes in PostgreSQL 10).

onderkalaci: Ah, sorry for the trouble. 

In general, we tried to add `ORDER BYs` to all tests for subqueries. However, these tests come from pre `v5.0` and it seems I forgot to refactor them.

PR #1185, opened 2017-02-01
Use CustomScan API for query execution

anarazel:
Reason we're working this:
- removing the temp table for real-time queries, which is good for efficiency and for running queries on standbys
- basis for allowing to execute multiple remote queries within one query (i.e. pull to master style queries)
- fix PREPARE + explain
- remove code

Behavioural changes:
- explain now happens inside the custom scan's explain callback. That's good because it removes a bunch of code and add features. But it also means the explain output changes - we're now getting called "within" the master query for real-time/tt queries, i.e. the remote explain is now where the sequential scan of the remote query used to be.  That's hard to avoid without complicating things, *and* seems to be the right thing to anyway, because we'll soon have multiple distributed queries within one query.
- execution now happens during the actual execution phase, not earlier. This means we should be able to make pg_stat_statements work (if it doesn't already)
- no temp tables anymore :) (which means pg matviews are now supported...) 

Todo
- [x] Verify pg_stat_statements now works (i.e. doesn't show "(temp table creation)" and such, and that the times are meaningful) 
- [x] Verify query execution now happens during the 'execute' phase, not the 'bind' phase as before.  A good way to test that is to use pgbench with -M extended, and use log_min_duration_statement=0
- [x] Do some simple benchmarking with a query returning a large number of rows (e.g. just COPY lineitem TO '/dev/null'), to make sure we're not regressing
- [x] Concurrent pgbench for quick router queries.
- [x] Basic testing
- [x] Make sure we still acquire all the right relation level locks

Fixes #457 
Fixes #372 
Fixes #713

PR #1402, opened 2017-05-08
Perform INSERT..SELECT via coordinator if it cannot be pushed down

marcocitus:
This PR adds a second implementation of `INSERT INTO distributed_table SELECT ...` that is used if the query cannot be pushed down. The basic idea is to execute the `SELECT` query separately and pass the results into the distributed table using a `CopyDestReceiver`, which is also used for `COPY` and `create_distributed_table`. When planning the `SELECT`, we go through planner hooks again, which means the `SELECT` can also be a distributed query.

The planner has to change the original SELECT query in several ways to make it run correctly: 
- It adds outer-most CTEs to the subquery and corrects `ctelevelsup` where necessary
- Calls `ReorderInsertSelectTargetLists`
- Adds casts to the SELECT target list where necessary, since COPY does not cast the input

The final plan for a `INSERT INTO ... SELECT` query is similar to that of a router query: A top-level Custom Scan node containing a `MultiPlan`. The `MultiPlan` contains the target relation, insert target list, and the select query. I considered introducing a new type of plan (not `MultiPlan`), but found the maintenance burden too high. I also considered passing a `PlannedStmt` instead of a `Query`, but this isn't supported in 9.5.

Execution is straight-forward. The Custom Scan plans and executes the `SELECT` query, sending tuples into a `CopyDestReceiver`.

There is some reshuffling of error messages going on, since we fall through the distributed INSERT...SELECT planner if it errors out, but in most unsupported cases it lands on a very similar error from the subquery pushdown logic, so this seems acceptable.

EXPLAIN is supported, but EXPLAIN ANALYZE is not because preventing double execution was a lot more complicated in this case.

Some things you can do with INSERT ... SELECT:

Generate random data:
```
INSERT INTO events
SELECT s, now(), random() FROM generate_series(1,1000000) s;
```

Ingest from local staging table:
```
WITH ingest_rows AS (
  DELETE FROM staging_table RETURNING *
)
INSERT INTO events
SELECT * FROM ingest_rows;
```

Import data via dblink:
```
INSERT INTO events
SELECT * FROM dblink('host=remote-host', 'SELECT * FROM events')
AS events(user_id int, event_time timestamptz, value double precision);
```

Re-partition a table in MX:
```
SELECT run_command_on_placements('partitioned_by_x', $$
  INSERT INTO partitioned_by_y SELECT * FROM %I
$$);
```

Non-co-located roll-ups:
```
INSERT INTO total_by_date
SELECT event_time::date, count(*) FROM events
WHERE event_time BETWEEN '...' AND '...' GROUP BY event_time::date;
```
etc. 

When selecting from local data sources, this implementation can stream results into the table as they become available. For selecting from a distributed table, you have to be careful that all the query results are collected on the master first.


metdos:
> SELECT * FROM staging_table;

Minor typo: staging_table -> ingest_rows?

marcocitus:
> Minor typo: staging_table -> ingest_rows?

Oops, corrected

src/backend/distributed/executor/insert_select_executor.c
```diff
@@ -0,0 +1,142 @@
...
+
+	if (!scanState->finishedRemoteScan)
+	{
+		EState *executorState = scanState->customScanState.ss.ps.state;
+		MultiPlan *multiPlan = scanState->multiPlan;
+		Query *selectQuery = multiPlan->insertSelectQuery;
+		List *insertTargetList = multiPlan->insertTargetList;
+		Oid targetRelationId = multiPlan->targetRelationId;
+
+		ereport(DEBUG1, (errmsg("Collecting INSERT ... SELECT results locally")));
```
marcocitus: What about MX?

src/backend/distributed/executor/insert_select_executor.c
```diff
@@ -0,0 +1,142 @@
...
+	long count = FETCH_ALL;
+
+	/* create a new portal for executing the query */
+	portal = CreateNewPortal();
+
+	/* don't display the portal in pg_cursors, it is for internal use only */
+	portal->visible = false;
+
+	/* plan the subquery, this may be another distributed query */
+	queryPlan = pg_plan_query(query, 0, params);
```
marcocitus: > What about passing CURSOR_OPT_PARALLEL_OK to pg_plan_query? That seems similar to this discussion?

Good idea!

> Also, shall we obey the citus.binary_master_copy_format flag here?

That should happen automatically if selecting from a distributed table.

src/backend/distributed/executor/insert_select_executor.c
marcocitus: It seems there are still a few errors with "via coordinator", those should be "locally" as well I think.

src/backend/distributed/executor/insert_select_executor.c
```diff
@@ -0,0 +1,142 @@
...
+
+		ereport(DEBUG1, (errmsg("Collecting INSERT ... SELECT results locally")));
+
+		ExecuteSelectIntoRelation(targetRelationId, insertTargetList, selectQuery,
+								  executorState);
+
+		scanState->finishedRemoteScan = true;
+	}
+
+	resultSlot = ReturnTupleFromTuplestore(scanState);
```
marcocitus: Yeh, it should be possible, though tricky since we need to evaluate the expression in the RETURNING part so it didn't seem very interesting cost/benefit-wise.

src/backend/distributed/executor/insert_select_executor.c
```diff
@@ -0,0 +1,142 @@
...
+
+	/* don't display the portal in pg_cursors, it is for internal use only */
+	portal->visible = false;
+
+	/* plan the subquery, this may be another distributed query */
+	queryPlan = pg_plan_query(query, 0, params);
+
+	PortalDefineQuery(portal,
+					  NULL,
+					  "",
```
marcocitus: I couldn't see a way to get the query from a CustomScanState.

src/backend/distributed/executor/insert_select_executor.c
marcocitus: ok, can't make up my mind :), but for consistency I'll go with `coordinator` for now.

marcocitus:
> Users might get some confusing errors (i.e., copy in the context of insert .. select), but, that seems unlikely to prevent I guess?

> BEGIN;
>
> insert into test_dist select * from test_dist ;
> insert into test_dist select * from test_local  ;
> ERROR:  distributed copy operations must not appear in transaction blocks containing

We can safely remove that error message since the COPY logic will pick the right connections using `GetPlacementConnection` after the first INSERT..SELECT.

I opened a separate PR for this:
https://github.com/citusdata/citus/pull/1441

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -2930,6 +2941,12 @@ InsertSelectQuery(Query *query)
 	/* ensure that there is a query */
 	Assert(IsA(subqueryRte->subquery, Query));
 
+	insertRte = ExtractInsertRangeTableEntry(query);
+	if (!IsDistributedTable(insertRte->relid))
```
marcocitus: changed the function name and clarified in the comment

src/backend/distributed/planner/insert_select_planner.c
```diff
@@ -0,0 +1,215 @@
+/*-------------------------------------------------------------------------
+ *
+ * insert_select_planner.c
```
marcocitus: I think that's a worthwhile change, but it would blow up the PR quite a bit. Would you be ok with doing that as a separate PR?

src/backend/distributed/planner/insert_select_planner.c
```diff
@@ -0,0 +1,215 @@
...
+															columnTypeMod,
+															COERCION_EXPLICIT,
+															COERCE_IMPLICIT_CAST,
+															-1);
+
+			selectTargetEntry->expr = typeCast;
+		}
+	}
+
+	multiPlan->insertSelectQuery = subquery;
```
marcocitus: We modify the subquery, so the original INSERT..SELECT is no longer valid. Using it as a container for insertTargetList seems like it might cause bugs at some point.

There's also a possibility that we might use a different `targetRelationId` at some point. I'm thinking of cases in which there is an `ON CONFLICT` and under the covers we first copy into a temporary distributed table and then do a distributed `INSERT ... SELECT ... ON CONFLICT ...`.

src/backend/distributed/executor/insert_select_executor.c
```diff
@@ -0,0 +1,150 @@
...
+	portal->visible = false;
+
+#if (PG_VERSION_NUM >= 90600)
+	cursorOptions = CURSOR_OPT_PARALLEL_OK;
+#endif
+
+	/* plan the subquery, this may be another distributed query */
+	queryPlan = pg_plan_query(query, cursorOptions, params);
+
+	PortalDefineQuery(portal,
```
onderkalaci: I've got a question and a note here. 
It seems when we use `Portal`, it doesn't create a separate session for the `subquery`:

```SQL
SET citus.shard_replication_factor = 1;
SET citus.shard_count = 4;


CREATE TABLE events_table (user_id int, time timestamp, event_type int, value_2 int, value_3 float, value_4 bigint);
SELECT create_distributed_table('events_table', 'user_id');
COPY events_table_local FROM '/Users/onderkalaci/Documents/citus_code/citus/src/test/regress/data/events_table.data' WITH CSV;


CREATE TABLE agg_results (user_id int, value_1_agg int, value_2_agg int, value_3_agg float, value_4_agg bigint, agg_time timestamp);
SELECT create_distributed_table('agg_results', 'user_id');

INSERT INTO agg_results (value_4_agg, user_id) SELECT user_id, value_3 FROM events_table WHERE user_id = 5;
DEBUG:  StartTransactionCommand
DEBUG:  StartTransaction
DEBUG:  name: unnamed; blockState:       DEFAULT; state: INPROGR, xid/subid/cid: 0/1/0, nestlvl: 1, children: 
DEBUG:  cannot perform distributed INSERT INTO ... SELECT because the partition columns in the source table and subquery do not match
DETAIL:  The data type of the target table's partition column should exactly match the data type of the corresponding simple column reference in the subquery.
DEBUG:  ProcessQuery
DEBUG:  Creating router plan
ERROR:  cannot establish new placement connection for DML when other connections exist for colocated placements

```
Do you foresee any other such cases? 

(As far as I can see, `SPI` wouldn't be applicable here. )

src/backend/distributed/planner/insert_select_planner.c
```diff
@@ -0,0 +1,234 @@
...
+
+
+/*
+ * ErrorIfCoordinatorInsertSelectUnsupported returns an error if executing an
+ * INSERT ... SELECT command by pulling results of the SELECT to the coordinator
+ * is unsupported because it uses RETURNING, ON CONFLICT, or an append-distributed
+ * table.
+ */
+static DeferredErrorMessage *
+ErrorIfCoordinatorInsertSelectUnsupported(Query *insertSelectQuery)
```
onderkalaci: Very minor: For subquery pushdown, we used the prefix `DeferErrorIf..`. But, pick whichever you want.

src/include/distributed/multi_physical_planner.h
```diff
@@ -221,6 +221,11 @@ typedef struct MultiPlan
 	Query *masterQuery;
 	bool routerExecutable;
 
+	/* INSERT ... SELECT via coordinator only */
+	Query *insertSelectQuery;
```
onderkalaci: Minor discussion: When I see `insertSelectQuery`, I feel that it covers both the distributed and the coordinator one. So, thinking to suggest renaming the field to `coordinatorInsertSelectQuery` (or so), but that's a very long name for a variable too. So, just sharing my thoughts :)

src/backend/distributed/executor/insert_select_executor.c
marcocitus: We can't perform the subquery in a separate session since we wouldn't see uncommitted writes and the subquery may perform modifications that we don't want to commit immediately. Note that SPI also doesn't start a separate session, it's essentially just a wrapper around the `Portal` functions.

I can't seem to reproduce that error, but there will be transactions which cannot be executed due to the way we assign connections, similar to COPY. Over time we can hopefully reduce these.

src/backend/distributed/planner/insert_select_planner.c
```diff
@@ -0,0 +1,234 @@
...
+		selectOutputType = columnVar->vartype;
+
+		/*
+		 * If the type in the target list does not match the type of the column,
+		 * we need to cast to the column type. PostgreSQL would do this
+		 * automatically during the insert, but we're passing the SELECT
+		 * output directly to COPY.
+		 */
+		if (columnType != selectOutputType)
+		{
```
onderkalaci: Minor: Maybe move `TargetEntry *selectTargetEntry = (TargetEntry *) lfirst(selectTargetCell);` to here? That'd seem a bit confusing but probably more convinient wrt the conventions?

src/backend/distributed/planner/insert_select_planner.c
```diff
@@ -0,0 +1,234 @@
...
+
+	insertRte = ExtractInsertRangeTableEntry(insertSelectQuery);
+	if (PartitionMethod(insertRte->relid) == DISTRIBUTE_BY_APPEND)
+	{
+		return DeferredError(ERRCODE_FEATURE_NOT_SUPPORTED,
+							 "INSERT ... SELECT into an append-distributed table is "
+							 "not supported", NULL, NULL);
+	}
+
+	return NULL;
```
onderkalaci: I think we need to error out if the target table is a reference table and the node is not the coordinator node.

src/backend/distributed/executor/insert_select_executor.c
```diff
@@ -0,0 +1,150 @@
...
+ */
+static void
+ExecuteSelectIntoRelation(Oid targetRelationId, List *insertTargetList,
+						  Query *selectQuery, EState *executorState)
+{
+	ParamListInfo paramListInfo = executorState->es_param_list_info;
+
+	ListCell *insertTargetCell = NULL;
+	List *columnNameList = NIL;
+	bool stopOnFailure = false;
```
onderkalaci: I think we need to error on inserts to reference table otherwise replicas may diverge (or it marks one of the replicas of reference table as `INVALID`):
```C
	partitionMethod = PartitionMethod(tableId);
	if (partitionMethod == DISTRIBUTE_BY_NONE)
	{
		stopOnFailure = true;
	}
```

src/include/distributed/insert_select_executor.h
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * Declarations for public functions and types related to executing
+ * INSERT..SELECT commands.
+ *
+ * Copyright (c) 2017, Citus Data, Inc.
+ *
+ *-------------------------------------------------------------------------
+ */
+
+#pragma once
```
onderkalaci: It seems we don't use `pragma once` any other place? Instead, we use `ifndef ..` (Same comment for the planner header file as well.)

src/backend/distributed/executor/insert_select_executor.c
```diff
@@ -0,0 +1,150 @@
...
+	queryPlan = pg_plan_query(query, cursorOptions, params);
+
+	PortalDefineQuery(portal,
+					  NULL,
+					  "",
+					  "SELECT",
+					  list_make1(queryPlan),
+					  NULL);
+
+	PortalStart(portal, params, eflags, GetActiveSnapshot());
```
marcocitus: That's an annoying bug. This is caused by the way in which Citus handles calls to `nextval` in `SELECT` queries. Arguably those should not be pushed down since the sequence might not even exist on the worker. This seems like something we should fix independently of the INSERT...SELECT implementation. For now I guess we could error out if there is a `nextval` call in the insert target list and the SELECT needs distributed planning.

src/backend/distributed/planner/insert_select_planner.c
marcocitus: isn't doing `lfirst` at the start of `foreach` also a convention though? :)

src/backend/distributed/planner/insert_select_planner.c
marcocitus: This happens in `CitusCopyDestReceiverStartup`.

src/backend/distributed/executor/insert_select_executor.c
marcocitus: Makes sense, added.

onderkalaci:
> INSERT INTO events
SELECT * FROM dblink('host=remote-host', 'SELECT * FROM events')
AS events(user_id int, event_time timestamptz, value double precision);

Btw, it is very easy to achieve similar things with `postgres_fdw` 👍

src/backend/distributed/planner/insert_select_planner.c
```diff
@@ -0,0 +1,253 @@
...
+	bool walkIsComplete = false;
+	if (node == NULL)
+	{
+		return false;
+	}
+
+	if (IsA(node, RangeTblEntry))
+	{
+		RangeTblEntry *rangeTableEntry = (RangeTblEntry *) node;
+		if (rangeTableEntry->ctename != NULL &&
```
marcocitus: it's not the CTE itself but a reference to a CTE

marcocitus:
> I know you've already explained this to me in person, but I forgot the details. Could you please explain our motivation and the behavior once more?

CTEs can be nested inside of any subquery. Range table entries that reference CTEs therefore specify which subquery they refer to (e.g. 2 subqueries up), since CTE names don't have to be unique. The `DecrementCTELevelsWalker` function exists because we go from having:

```
WITH cte1 AS (....) INSERT INTO distributed_table WITH cte2 AS (....) SELECT ... FROM cte1, cte2 ...;
```

to:
```
WITH cte1 AS (....), cte2 AS (....) SELECT ... FROM cte1, cte2 ...;
```

i.e. the "level" of `cte1` changes from being part of the `INSERT` to being part of the `SELECT`. The `ctelevelsup` (read: number of subquery levels up to find CTE) in all the range table entry generated for `FROM cte1` needs to be changed from 1 to 0. Otherwise, it points to a non-existent level and the query will error out.

A limitation of the current approach is that we don't handle CTEs with a name that occurs both in the outer-level and in the top-level SELECT, though it seems unlikely that the user will type this (there's actually no way to refer to the outer-CTE):

```
WITH cte1 AS (....) INSERT INTO test WITH cte1 AS (....) SELECT * FROM cte1;
ERROR:  could not find plan for CTE "cte1"
```

I also considered wrapping the subquery in another subquery and attaching the outer-CTEs to the subquery, but that was more complicated and didn't seem worth it for the 1 edge case.

src/test/regress/expected/multi_mx_modifications.out
```diff
@@ -108,10 +108,6 @@ DELETE FROM limit_orders_mx WHERE id = 246 AND placed_at = current_timestamp::ti
 INSERT INTO limit_orders_mx VALUES (DEFAULT), (DEFAULT);
 ERROR:  cannot perform distributed planning for the given modification
 DETAIL:  Multi-row INSERTs to distributed tables are not supported.
--- INSERT ... SELECT ... FROM commands are unsupported from workers
-INSERT INTO limit_orders_mx SELECT * FROM limit_orders_mx;
```
marcocitus: I think I removed it because it was causing constraint violations. Added some MX tests below.

src/backend/distributed/planner/insert_select_planner.c
```diff
@@ -0,0 +1,253 @@
...
+					rangeTableEntry->ctelevelsup--;
+					break;
+				}
+			}
+		}
+	}
+	else if (IsA(node, Query))
+	{
+		/*
+		 * One level deeper, so a ctelevelsup referring to a top-level CTE will
```
marcocitus: Yes, CTEs can be defined on every subquery, so every subquery is a new level.

onderkalaci:
@marcocitus Thanks for the detailed explanation for the `CTEs`. So, given that we're implementing a relatively complex logic, shall we add some regression tests such as the followings (which works perfectly now, but, lets include in case anyone breaks in the future):  

(The following query may not makes sense, but, syntatically correct. So, feel free to write more meaningful test 😄 )
```SQL
WITH 
cte1 AS
(

	SELECT * FROM events_local_1 WHERE user_id IN (SELECT user_id FROM events_local_2 WHERE user_id > 15)

), 
cte2 AS
(

	SELECT * FROM cte1 WHERE user_id IN (SELECT user_id FROM events_local_2 WHERE user_id > 50)

), 
cte3 AS
(

	SELECT * FROM cte2 WHERE user_id IN (SELECT user_id FROM events_local_2 WHERE user_id > 500)

)
INSERT INTO events_local_4 

	WITH cte4 AS (SELECT * FROM cte3 WHERE user_id > 999)
	SELECT cte3.* FROM cte3, cte4 WHERE cte3.user_id = cte4.user_id;
```

src/backend/distributed/planner/insert_select_planner.c
```diff
@@ -0,0 +1,246 @@
...
+	multiPlan->insertSelectSubquery = subquery;
+	multiPlan->insertTargetList = insertSelectQuery->targetList;
+	multiPlan->targetRelationId = targetRelationId;
+
+	return multiPlan;
+}
+
+
+/*
+ * ErrorIfCoordinatorInsertSelectUnsupported returns an error if executing an
```
onderkalaci: Very very minor: function name and first word of the comment doesn't match

src/backend/distributed/planner/insert_select_planner.c
```diff
@@ -0,0 +1,246 @@
...
+
+	if (IsA(node, RangeTblEntry))
+	{
+		RangeTblEntry *rangeTableEntry = (RangeTblEntry *) node;
+		if (rangeTableEntry->ctename != NULL &&
+			rangeTableEntry->ctelevelsup == context->topLevelsUp)
+		{
+			ListCell *cteCell = NULL;
+
+			foreach(cteCell, context->topLevelCTEList)
```
onderkalaci: Very minor -- feel free to overlook since the following comment might be personal:

 shall we use `insertCTEList` (or smth like that) instead of `topLevelCTEList`? Initially, at this point of the code, I became confused since this is a walker function and `topLevel` seemed top level of the recursion.

marcocitus:
> Would it make sense to change them all to locally as you used previously at some points?

I think locally is also a bit confusing, since it might mean local to the worker(s). I'm not particularly happy about "Coordinator", but I can live with it for now. Added a comment to `CreateCoordinatorInsertSelectPlan`.

marcocitus:
@onderkalaci I decided to replace `DecrementCTELevelsWalker` and wrap the SELECT into a subquery in `insert_select_planner.c` because the edge cases got a little weird. I also addressed #1458 in this PR since it was an issue for INSERT ... SELECT as well. Could you have another look?

src/backend/distributed/executor/insert_select_executor.c
```diff
@@ -0,0 +1,157 @@
...
+
+	partitionMethod = PartitionMethod(targetRelationId);
+	if (partitionMethod == DISTRIBUTE_BY_NONE)
+	{
+		stopOnFailure = true;
+	}
+
+	/* build the list of column names for the COPY statement */
+	foreach(insertTargetCell, insertTargetList)
+	{
```
onderkalaci: While I was randomly testing the changes, I hit an issue which is not straightforward to understand (at least for me). So, I'm posting here in case you have any ideas.

```SQL

CREATE TABLE agg_results (user_id int, value_1_agg int, value_2_agg int, value_3_agg float, value_4_agg bigint, agg_time timestamp);
SELECT create_distributed_table('agg_results', 'user_id');
CREATE TABLE users_table_local (user_id int, time timestamp, value_1 int, value_2 int, value_3 float, value_4 bigint);

INSERT INTO users_table_local (user_id) VALUES (1);

INSERT INTO agg_results  (value_4_agg, user_id)
(
	SELECT value_4::float, user_id FROM users_table_local
)
UNION
(
	SELECT value_4, user_id FROM users_table_local
);
ERROR:  the partition column of table public.agg_results should have a value

```

src/backend/distributed/planner/insert_select_planner.c
```diff
@@ -0,0 +1,223 @@
...
+	if (multiPlan->planningError != NULL)
+	{
+		return multiPlan;
+	}
+
+	ReorderInsertSelectTargetLists(insertSelectQuery, insertRte, subqueryRte);
+
+	if (list_length(insertSelectQuery->cteList) > 0)
+	{
+		/*
```
onderkalaci: This logic seems complicated enough to live in a separate function. Could you move it to another function if it is easy enough to do?

src/backend/distributed/executor/insert_select_executor.c
marcocitus: would it make sense to change "should have a value" to "cannot be NULL"?

onderkalaci: But the problem is that `user_id` has a value. I suspect this is happening something related to `ReorderInsertSelectTargetList`

src/test/regress/expected/multi_explain_0.out
```diff
@@ -940,3 +940,38 @@ Custom Scan (Citus Router)
         ->  Seq Scan on explain_table_570001 explain_table
               Filter: (id = 1)
 ROLLBACK;
+-- test explain with local INSERT ... SELECT
```
onderkalaci: Shall we have an EXPLAIN test with `CTE`s? Since we're manually changing the query, `EXPLAIN` could ensure that we're doing things correctly.

src/backend/distributed/planner/insert_select_planner.c
```diff
@@ -0,0 +1,223 @@
...
+														   selectAlias, false, true);
+		outerQuery->rtable = list_make1(newRangeTableEntry);
+
+		/* set the FROM expression in the subquery */
+		newRangeTableRef = makeNode(RangeTblRef);
+		newRangeTableRef->rtindex = 1;
+		outerQuery->jointree = makeFromExpr(list_make1(newRangeTableRef), NULL);
+
+		/* create a target list that matches the SELECT */
+		foreach(selectTargetCell, subquery->targetList)
```
onderkalaci: I think we should also consider the `Vars` in `havingQual`

src/backend/distributed/executor/insert_select_executor.c
marcocitus: It seems `ReorderInsertSelectTargetList` doesn't handle top-level unions properly. In case of a set operation we now also wrap the SELECT in a subquery and then reorder using the outer query to avoid having to reorder a top-level union.

src/test/regress/expected/multi_explain_0.out
marcocitus: Makes sense, added.

src/backend/distributed/planner/insert_select_planner.c
```diff
@@ -0,0 +1,254 @@
...
+		return multiPlan;
+	}
+
+	selectQuery = selectRte->subquery;
+
+	/*
+	 * Wrap the SELECT as a subquery if the INSERT...SELECT has CTEs or the SELECT
+	 * has top-level set operations.
+	 *
+	 * We could simply wrap all queries, but that might create a subquery that is
```
onderkalaci: Hopefully one day I'll update `ReorderInsertSelectTargetList()` to handle top level union 😄

src/backend/distributed/planner/insert_select_planner.c
```diff
@@ -0,0 +1,254 @@
...
+											   selectTargetEntry->resname,
+											   selectTargetEntry->resjunk);
+
+		newTargetList = lappend(newTargetList, newSelectTargetEntry);
+	}
+
+	outerQuery->targetList = newTargetList;
+
+	return outerQuery;
+}
```
onderkalaci: Missing new line at the end of the file

src/backend/distributed/planner/multi_explain.c
```diff
@@ -710,3 +715,73 @@ ExplainYAMLLineStarting(ExplainState *es)
...
 }
+
+
+/*
+ * CoordinatorInsertSelectExplainScan is a custom scan explain callback function
+ * which is used to print explain information of a Citus plan for an INSERT INTO
+ * distributed_table SELECT ... query that is evaluated on the coordinator.
+ */
+void
+CoordinatorInsertSelectExplainScan(CustomScanState *node, List *ancestors,
```
jasonmp85: The bottom of this file is all copy-pasted functions from `explain.c`, so this function doesn't belong here (and putting it here omits it from `citus_indent` consideration).

src/backend/distributed/planner/multi_explain.c
```diff
@@ -710,3 +715,73 @@ ExplainYAMLLineStarting(ExplainState *es)
...
+	ExplainCloseGroup("Select Query", "Select Query", false, es);
+}
+
+
+/*
+ * CitusExplainOneQuery is simply a duplicate of ExplainOneQuery in explain.c, which
+ * is static.
+ */
+static void
+CitusExplainOneQuery(Query *query, IntoClause *into, ExplainState *es,
```
jasonmp85: We copied a bunch of other `static` `Explain*` functions as the end of this file. Why prefix this one with `Citus-`? Is it meaningfully different?

jasonmp85: Actually, didn't we work hard to _remove_ the copy-pasted version of `ExplainOneQuery` in #1185? Why does it come back now?

marcocitus: > Is it meaningfully different?

ExplainOneQuery also handles utility commands (mainly `EXECUTE`). `CitusExplainOneQuery` does not handle utility commands, since it's specifically meant for certain (locally executed) subqueries.

> Actually, didn't we work hard to remove the copy-pasted version of ExplainOneQuery in #1185? Why does it come back now?

To explain the SELECT part of an INSERT ... SELECT, which separately goes through the top-level planner hooks and should go through the top-level EXPLAIN hook.

anarazel: Well, that wasn't the primary motiviation for #1185 .

But nevertheless, it seems unnecessary here. Why don't we plan that query once, and store it in the custom scan node?  Then we don't replan (which could yield a different plan!), and the need for this ought to be gone?

jasonmp85: Didn't mean primary motivation, was a side-effect, but I do like deleting copy-pasted code from PostgreSQL. Maybe we can look into what you suggest in a bit (for now I've simply made @marcocitus's changes version-aware in 42140bae37b903c29efae340bfbb06385e56c58e).

src/backend/distributed/planner/multi_explain.c
marcocitus: Makes sense. I've added it as a separate commit to PR #1469.

src/backend/distributed/planner/multi_explain.c
marcocitus: > Why don't we plan that query once, and store it in the custom scan node? Then we don't replan (which could yield a different plan!), and the need for this ought to be gone?

Main reason is that it PlannedStmt can't be serialised in 9.5.

anarazel: > > Why don't we plan that query once, and store it in the custom scan node? Then we don't replan (which could yield a different plan!), and the need for this ought to be gone?

> Main reason is that it PlannedStmt can't be serialised in 9.5.

We're about to a) drop 9.5 b) remove the serialization, so I'd not go there just for that reason.

jasonmp85: How soon are we doing a? I don't think it can reasonably be done before PostgreSQL 10 is out.

jasonmp85:
Added the backport label since 2cd358ad1ae05662b33dd173d5b694c65ed1a0e1 ported back to 6.2.

onderkalaci:
> Added the backport label since 2cd358a ported back to 6.2.

But, we're not going to backport the whole PR, right?

jasonmp85:
no, just that commit

Issue #1400, opened 2017-05-08
Concurrent calls to master_remove_node fails with “tuple concurrently updated” error


mtuncer:
### Steps
- Connection 1
  - Starts transaction
  - Calls master_remove_node
  - Does not commit;
- Connection 2
  - Starts transaction
  - Calls master_remove_node
  - Blocked
- Connection 1 commits
- Connection 2 gets unblocked fails with “Error: tuple concurrently updated”

Same behavior is observed when master_activate_node and master_remove_node is called concurrently where activate is called first.

mtuncer:
Clarification : functions are called for the same node

sumedhpathak:
@mtuncer Isn't this expected behavior? Or are we expecting something different to happen? Thanks.

marcocitus:
Definitely not the expected error message. 

The table locking in `node_metadata.c` needs work.

Issue #1397, opened 2017-05-08
Concurrent create_distributed_table and master_remove_node fails 


velioglu:
To reproduce : 

- Connection 1
```
create table tt1(id int);
begin;
    select create_distributed_table('tt1', 'id');
```

- Connection 2
` select master_remove_node('host-name', port);`

- Connection 1
`end;`

```
Insert into tt1 values(5);
ERROR:  failed to assign 1 task(s) to worker nodes
```

metdos:
@velioglu, could you also add the expected behavior?

velioglu:
`master_remove_node` supposed to delete nodes without any active shard on it only. Otherwise, it deletes the tuple from `pg_dist_node` with the given node-name and node port, but shard placements with the same name and port are still there in the `pg_dist_shard` table. So, the expected behaviour would be preventing `master_remove_node` from removing the node.

metdos:
The fix here is simple actually, getting a lock on `pg_dist_shard` in master_remove_node() solves the problem.

sumedhpathak:
@metdos Are we planning to fix this in 6.2 then? Thanks.

metdos:
> @metdos Are we planning to fix this in 6.2 then? Thanks.

Not necessarily. After talking with @ozgune, we inclined to not prioritize it. However, if we would like to fix it, I can create the PR and have somebody review it tomorrow.

marcocitus:
Note that this has been fixed on master, though it might be worth adding some isolation tests.

PR #1403, opened 2017-05-08
Accept invalidations during COPY command

### LINKED ISSUE #1392: Copying to reference table causes data inconsistency when a new node is added to cluster concurrently
Master_add_node called first, copy blocked, after copy is unblocked it only copies to old nodes but not to the newly added node.

## Prerequisites:
Cluster has 3 nodes, 1 ready to add
A reference table is created (r1)
Some data is ready to copy (/tmp/r1.txt)

### Connection 1:
```sql
begin;
select * from master_add_node('localhost', 9703);
commit;
```
### Connection 2:
```sql
begin;
copy r1 from '/tmp/r1.txt';
commit;
```
```
Connection 1, executes master_add_node but does not commit. 
Connection 2, executes copy, gets blocked
Connection 1 commits;
Connection 1 commits;
```
### Result:
All 4 workers contain reference table, but the newly added one does not receive any data. Verified by running
```sql
select run_command_on_placements('r1', 'select count(*) from %s');
```

onderkalaci:
Fixes #1392 . 

    Accept invalidations during COPY command to prevent inconsistent data among placements
    
    COPY command is blocked on the shard metadata lock and once it acquires the lock it doesn't check the cache invalidations. In some
    cases this could lead to data inconsistency among the shard replicas
    if there exists concurrent shard placement addition.

### DIFF
--- a/src/backend/distributed/commands/multi_copy.c
+++ b/src/backend/distributed/commands/multi_copy.c
@@ -72,6 +72,7 @@
 #include "nodes/makefuncs.h"
 #include "tsearch/ts_locale.h"
 #include "utils/builtins.h"
+#include "utils/inval.h"
 #include "utils/lsyscache.h"
 #include "utils/rel.h"
 #include "utils/memutils.h"
@@ -1750,6 +1751,14 @@ CitusCopyDestReceiverStartup(DestReceiver *dest, int operation,
 	/* prevent concurrent placement changes and non-commutative DML statements */
 	LockShardListMetadata(shardIntervalList, ShareLock);
 	LockShardListResources(shardIntervalList, ShareLock);
+
+	/*
+	 * We might have some concurrent metadata changes. In order to get the changes,
+	 * we first need to accept the cache invalidation messages and then access the
+	 * cache entry to trigger the cache entry rebuild.
+	 */
+	AcceptInvalidationMessages();
+	cacheEntry = DistributedTableCacheEntry(tableId);
 
 	/* keep the table metadata to avoid looking it up for every tuple */
 	copyDest->tableMetadata = cacheEntry;
--- a/src/test/regress/expected/isolation_add_node_vs_copy.out
+++ b/src/test/regress/expected/isolation_add_node_vs_copy.out
@@ -8,11 +8,11 @@ step s1-begin:
     BEGIN;
 
 step s1-add-second-worker: 
-   	SELECT master_add_node('localhost', 57638);
+   	SELECT nodename, nodeport, isactive FROM master_add_node('localhost', 57638);
 
-master_add_node
+nodename       nodeport       isactive       
 
-(4,4,localhost,57638,default,f,t)
+localhost      57638          t              
 step s2-begin: 
     BEGIN;
 
@@ -27,12 +27,17 @@ step s2-commit:
     COMMIT;
 
 step s2-print-content: 
-	SELECT run_command_on_placements('test_reference_table', 'select count(*) from %s');
+	SELECT 
+		nodeport, success, result 
+	FROM 
+		run_command_on_placements('test_reference_table', 'select count(*) from %s')
+	ORDER BY
+		nodeport;
 
-run_command_on_placements
+nodeport       success        result         
 
-(localhost,57637,102197,t,5)
-(localhost,57638,102197,t,5)
+57637          t              5              
+57638          t              5              
 step s1-remove-second-worker: 
    	SELECT master_remove_node('localhost', 57638);
 
@@ -43,11 +48,11 @@ step s1-begin:
     BEGIN;
 
 step s1-add-second-worker: 
-   	SELECT master_add_node('localhost', 57638);
+   	SELECT nodename, nodeport, isactive FROM master_add_node('localhost', 57638);
 
-master_add_node
+nodename       nodeport       isactive       
 
-(5,5,localhost,57638,default,f,t)
+localhost      57638          t              
 step s2-begin: 
     BEGIN;
 
@@ -62,12 +67,17 @@ step s2-commit:
     COMMIT;
 
 step s2-print-content: 
-	SELECT run_command_on_placements('test_reference_table', 'select count(*) from %s');
+	SELECT 
+		nodeport, success, result 
+	FROM 
+		run_command_on_placements('test_reference_table', 'select count(*) from %s')
+	ORDER BY
+		nodeport;
 
-run_command_on_placements
+nodeport       success        result         
 
-(localhost,57637,102197,t,10)
-(localhost,57638,102197,t,10)
-master_add_node
+57637          t              10             
+57638          t              10             
+nodename       nodeport       isactive       
 
-(5,5,localhost,57638,default,f,t)
+localhost      57638          t              
--- a/src/test/regress/expected/isolation_add_node_vs_copy.out
+++ b/src/test/regress/expected/isolation_add_node_vs_copy.out
@@ -1,0 +1,73 @@
+Parsed test spec with 2 sessions
+
+starting permutation: s1-begin s1-add-second-worker s2-begin s2-copy-to-reference-table s1-commit s2-commit s2-print-content s1-remove-second-worker s1-begin s1-add-second-worker s2-begin s2-copy-to-reference-table s1-commit s2-commit s2-print-content
+create_reference_table
+
+               
+step s1-begin: 
+    BEGIN;
+
+step s1-add-second-worker: 
+   	SELECT master_add_node('localhost', 57638);
+
+master_add_node
+
+(4,4,localhost,57638,default,f,t)
+step s2-begin: 
+    BEGIN;
+
+step s2-copy-to-reference-table: 
+	COPY test_reference_table FROM PROGRAM 'echo "1\n2\n3\n4\n5"';
+ <waiting ...>
+step s1-commit: 
+    COMMIT;
+
+step s2-copy-to-reference-table: <... completed>
+step s2-commit: 
+    COMMIT;
+
+step s2-print-content: 
+	SELECT run_command_on_placements('test_reference_table', 'select count(*) from %s');
+
+run_command_on_placements
+
+(localhost,57637,102197,t,5)
+(localhost,57638,102197,t,5)
+step s1-remove-second-worker: 
+   	SELECT master_remove_node('localhost', 57638);
+
+master_remove_node
+
+               
+step s1-begin: 
+    BEGIN;
+
+step s1-add-second-worker: 
+   	SELECT master_add_node('localhost', 57638);
+
+master_add_node
+
+(5,5,localhost,57638,default,f,t)
+step s2-begin: 
+    BEGIN;
+
+step s2-copy-to-reference-table: 
+	COPY test_reference_table FROM PROGRAM 'echo "1\n2\n3\n4\n5"';
+ <waiting ...>
+step s1-commit: 
+    COMMIT;
+
+step s2-copy-to-reference-table: <... completed>
+step s2-commit: 
+    COMMIT;
+
+step s2-print-content: 
+	SELECT run_command_on_placements('test_reference_table', 'select count(*) from %s');
+
+run_command_on_placements
+
+(localhost,57637,102197,t,10)
+(localhost,57638,102197,t,10)
+master_add_node
+
+(5,5,localhost,57638,default,f,t)
--- a/src/test/regress/isolation_schedule
+++ b/src/test/regress/isolation_schedule
@@ -2,3 +2,4 @@ test: isolation_cluster_management
 test: isolation_dml_vs_repair
 test: isolation_concurrent_dml
 test: isolation_drop_shards
+test: isolation_add_node_vs_copy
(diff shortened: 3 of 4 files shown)


src/backend/distributed/commands/multi_copy.c
```diff
@@ -1751,6 +1752,12 @@ CitusCopyDestReceiverStartup(DestReceiver *dest, int operation,
 	LockShardListMetadata(shardIntervalList, ShareLock);
 	LockShardListResources(shardIntervalList, ShareLock);
 
+	/*
+	 * We might have some concurrent metadata changes and we need to accept the cache
+	 * invalidation messages to learn about the changes.
+	 */
+	AcceptInvalidationMessages();
```
marcocitus: We should reload the cache entry below, i.e. `cacheEntry = DistributedTableCacheEntry(...);`.

marcocitus: ok, tracing through the code path, it seems we call `OpenCopyConnections` ->  `MasterShardPlacementList` -> `FinalizedShardPlacementList` -> `ShardPlacementList` -> `LookupShardCacheEntry` -> `LookupDistTableCacheEntry` which reloads the cache entry, so reloading here is not strictly necessary.

marcocitus: I wonder whether it makes sense to always do `AcceptInvalidationMessages()` before checking whether a cache entry is valid in `metadata_cache.c`.

onderkalaci: Hmm, we're planning to backport this PR. So, I'd be in favor of discussing that later. But your suggestion seems a pretty useful approach if `AcceptInvalidationMessages ()` is a cheap function to call.

(For reloading the cache entry, I was also thinking to add to make the code more readable. Also, we're setting `copyDest->tableMetatada = cacheEntry` just 1 line below. So, it makes sense to trigger the cache invalidation there.)

marcocitus: > For reloading the cache entry, I was also thinking to add to make the code more readable. 

Yep, seems like a good idea. Note that the `copyDest->tableMetatada = cacheEntry` is completely safe since the `cacheEntry` pointer always stays the same.

What's slightly more suspicious is that we do `cacheEntry->replicationModel == REPLICATION_MODEL_2PC` right after a possible invalidation. Theoretically, we might see a stale replication model here :).

anarazel:
Isn't the proper fix to add `AcceptInvalidationMessages()` to `LockShardListResources` etc? That's how it works for postgres...

marcocitus:
> Isn't the proper fix to add AcceptInvalidationMessages() to LockShardListResources etc?

Should be `LockShardListMetadata`. `LockShardListResources` is unrelated to metadata.

anarazel:
> Should be LockShardListMetadata. LockShardListResources is unrelated to metadata.

Well, I simply mean *all* our locking functions.   Most of similar locking infrastructure, like e.g. `LockRelationOid()` (called by `heap_open()` via `relation_open()`) have a block like:
```C
	/*
	 * Now that we have the lock, check for invalidation messages, so that we
	 * will update or flush any stale relcache entry before we try to use it.
	 * RangeVarGetRelid() specifically relies on us for this.  We can skip
	 * this in the not-uncommon case that we already had the same type of lock
	 * being requested, since then no one else could have modified the
	 * relcache entry in an undesirable way.  (In the case where our own xact
	 * modifies the rel, the relcache update happens via
	 * CommandCounterIncrement, not here.)
	 */
	if (res != LOCKACQUIRE_ALREADY_HELD)
		AcceptInvalidationMessages();
```

onderkalaci:
@marcocitus Assigning to you for the review. You can easily fail the added isolation test by removing the source code changes.

I'll open a separate issue to track these discussions.

src/test/regress/expected/isolation_add_node_vs_copy.out
```diff
@@ -0,0 +1,73 @@
...
+   	SELECT master_add_node('localhost', 57638);
+
+master_add_node
+
+(4,4,localhost,57638,default,f,t)
+step s2-begin: 
+    BEGIN;
+
+step s2-copy-to-reference-table: 
+	COPY test_reference_table FROM PROGRAM 'echo "1\n2\n3\n4\n5"';
```
metdos: Nice solution for copy :)

src/test/regress/expected/isolation_add_node_vs_copy.out
```diff
@@ -0,0 +1,73 @@
...
+step s2-copy-to-reference-table: <... completed>
+step s2-commit: 
+    COMMIT;
+
+step s2-print-content: 
+	SELECT run_command_on_placements('test_reference_table', 'select count(*) from %s');
+
+run_command_on_placements
+
+(localhost,57637,102197,t,10)
```
onderkalaci: Hmm, printing these stuff is fragile. Any ideas?

mtuncer:
``` sql
select success, result from run_command_on_placements(...) order by shardid;
```

src/test/regress/specs/isolation_add_node_vs_copy.spec
```diff
@@ -0,0 +1,71 @@
...
+	
+    CREATE TABLE test_reference_table (test_id integer);
+    SELECT create_reference_table('test_reference_table');
+}
+
+teardown
+{
+    DROP TABLE IF EXISTS test_reference_table CASCADE;
+
+	SELECT master_add_node('localhost', 57637);
```
marcocitus: Should this be remove_node?

src/test/regress/specs/isolation_add_node_vs_copy.spec
```diff
@@ -0,0 +1,71 @@
...
+	
+	SELECT master_add_node('localhost', 57637);
+	
+    CREATE TABLE test_reference_table (test_id integer);
+    SELECT create_reference_table('test_reference_table');
+}
+
+teardown
+{
+    DROP TABLE IF EXISTS test_reference_table CASCADE;
```
marcocitus: IF EXISTS seems redundant.

src/test/regress/specs/isolation_add_node_vs_copy.spec
```diff
@@ -0,0 +1,71 @@
+setup
+{
+	truncate pg_dist_shard_placement;
+	truncate pg_dist_shard;
+	truncate pg_dist_partition;
+	truncate pg_dist_colocation;
+	truncate pg_dist_node;
```
marcocitus: It seems better to handle this by cleaning up in the `teardown` section, which seems to already happen?

src/test/regress/specs/isolation_add_node_vs_copy.spec
```diff
@@ -0,0 +1,71 @@
...
+step "s2-print-content"
+{
+	SELECT run_command_on_placements('test_reference_table', 'select count(*) from %s');
+}
+
+# verify that copy gets the invalidation and re-builts its metadata cache
+# note that we need to run the same test twice to ensure that metadata is cached
+# otherwise the test would be useless since the cache would be empty and the 
+# metadata data is gathered from the tables directly
+permutation "s1-begin" "s1-add-second-worker" "s2-begin" "s2-copy-to-reference-table" "s1-commit" "s2-commit" "s2-print-content" "s1-remove-second-worker" "s1-begin" "s1-add-second-worker" "s2-begin" "s2-copy-to-reference-table" "s1-commit" "s2-commit" "s2-print-content"
```
marcocitus: Are there other permutations we should include?

src/test/regress/isolation_schedule
```diff
@@ -2,3 +2,4 @@ test: isolation_cluster_management
 test: isolation_dml_vs_repair
 test: isolation_concurrent_dml
 test: isolation_drop_shards
+test: isolation_add_node_vs_copy
```
marcocitus: We probably want to do this test before the `isolation_cluster_management` test and do a `SELECT master_remove_node(nodename, nodeport) FROM pg_dist_node` in the `teardown`

anarazel:
I'm a -0.75 for merging something like this, rather than doing it properly like #1404

marcocitus:
That might expose cases in which we don't properly handle free-d cache entry data, so backporting it to 6.1 might be something we'll regret. We could merge this into release-6.1 and do it properly in master/6.2.

anarazel:
> That might expose cases in which we don't properly handle free-d cache entry data, so backporting it to 6.1 might be something we'll regret. We could merge this into release-6.1 and do it properly in master/6.2.

Most of those probably are vulnerable already, because many postgres functions (like heap_open etc), will already allow to trigger invals.  Not sure that this is likely to trigger cases that aren't vulnerable already.   But I can live with the split 6.1/6.2 approach.

onderkalaci:
> That might expose cases in which we don't properly handle free-d cache entry data, so backporting it to 6.1 might be something we'll regret. We could merge this into release-6.1 and do it properly in master/6.2.

Isn't that a bit late to check-in such a change to 6.2? We're almost done with the release testing.

src/test/regress/expected/isolation_add_node_vs_copy.out
onderkalaci: Thanks, nice to learn that!

marcocitus:
Well, we're release testing and still finding bugs in the release. This problem is likely to affect other areas as well. e.g. 2 concurrent `master_copy_shard_placement` calls (?). There are ~20 places where we call these lock functions and then read from the metadata cache without accepting invalidations. There are at least dozens of concurrency bugs, many of which we probably haven't tested. While they're unlikely to occur individually, the number of Citus deployments is such that they are somewhat likely to start affecting users.

Minor note: From a software engineering perspective, it makes more sense to me to put `AcceptInvalidationMessages()` just before we check `isValid` in `LookupDistTableCacheEntry` and `LookupShardCacheEntry`, otherwise it's just a weird side-effect of a locking function.

anarazel:
On 2017-05-10 01:32:35 -0700, Marco Slot wrote:
> Minor note: From a software engineering perspective, it makes more sense to me to put `AcceptInvalidationMessages()` just before we check `isValid` in `LookupDistTableCacheEntry` and `LookupShardCacheEntry`, otherwise it's just a weird side-effect of a locking function.

Well, not really.  The reason for putting them around locking is that
that's the point where the set of shards (or whatever) can't change due
to concurrent activity anymore.  I.e. after the lock you actually can be
sure that you'll have a consistent view of things.

- Andres

marcocitus:
> The reason for putting them around locking is that that's the point where the set of shards (or whatever) can't change due to concurrent activity anymore.

It technically does the right thing, but it's a side effect that is unrelated to locking itself, which is a poor software engineering practice.

src/test/regress/specs/isolation_add_node_vs_copy.spec
marcocitus: Yikes, that's not good. This seems like a separate problem related to #1400 and #1397.

src/test/regress/specs/isolation_add_node_vs_copy.spec
```diff
@@ -0,0 +1,72 @@
...
+    CREATE TABLE test_reference_table (test_id integer);
+    SELECT create_reference_table('test_reference_table');
+}
+
+# ensure that both nodes exists for the remaining of the isolation tests
+teardown
+{
+	DROP TABLE test_reference_table;
+	SELECT nodename, nodeport, isactive FROM master_add_node('localhost', 57637);
+	SELECT nodename, nodeport, isactive FROM master_add_node('localhost', 57638);
```
marcocitus: How about we do:
```
SELECT master_remove_node(nodename, nodeport) FROM pg_dist_node;
```

and put this test at the top of `isolation_schedule`?

onderkalaci: We could do that, but now the test is independent of its place. Isn't the current structure better?

marcocitus: If it was independent of place then it shouldn't have to remove a node at the start. That seems like a strange pattern.

To keep things simple, I suggest having 2 types of tests: tests that call master_add_node/master_remove_node and tests that don't. The former type should be before `isolation_cluster_management`, start with an empty `pg_dist_node`, and are not parallelisable. The latter type should be after `isolation_cluster_management` and are mostly parallelisable.

onderkalaci:
> I'm a -0.75 for merging something like this, rather than doing it properly like #1404

So, we're currently working on properly fixing it on #1406 . For now, we're freezing this PR and moving the regression test over there. If we consider not backporting #1406 to 6.1, we can consider this PR again.

sumedhpathak:
@onderkalaci Should we close this PR now that #1406 is done?

onderkalaci:
> @onderkalaci Should we close this PR now that #1406 is done?

This fixes a rare event, could only happen via concurrent `COPY` with replicate node when reference tables exist or repair/move placement on 6.1. 

The proper fix #1406 (checked-in to 6.2) touches literally all operations on Citus. Thus, although #1406 PR is very small, it doesn't seem a good candidate for backporting to 6.1.

We could consider backporting this PR, which touches only `COPY` code-path. Given that 6.2 is out and this is a rare event, I can live with this as a known issue, so closing the PR.

Any comments here?

PR #1405, opened 2017-05-09
Ensure all preceding writes are visible in data migration

### LINKED ISSUE #1398: Concurrent creation of a distributed table from non-empty table and copy into a same table result in empty table
### Setup:
- A local table with some data in it
- An external file to load data from

### Connection 1:
- Starts a transaction,
- issues copy to local table
- does not commit

### Connection 2:
- Starts a transaction
- Issues create distributed table table (blocked)
---
When Connection 1 commits, connection2 gets unblocked and creates a distributed table.

However the distributed table does not contain any data. Expect it to contain both old, and recently copied in data.

Removed the row in pg_dist_partition to make the table local again, verified copied data is still there.

It is somewhat related to #1392

marcocitus:
Fixes #1398

`create_distributed_table` neglects to push a new snapshot after obtaining the lock on the relation, which means it might not see some concurrent writes. Added the code to update the snapshot and isolation test.

This feature is being introduced as part of 6.2 and should be fixed before releasing 6.2.

### DIFF
--- a/src/backend/distributed/commands/create_distributed_table.c
+++ b/src/backend/distributed/commands/create_distributed_table.c
@@ -762,6 +762,17 @@ CopyLocalDataIntoShards(Oid distributedRelationId)
 	/* take an ExclusiveLock to block all operations except SELECT */
 	distributedRelation = heap_open(distributedRelationId, ExclusiveLock);
 
+	/*
+	 * All writes have finished, make sure that we can see them by using the
+	 * latest snapshot. We use GetLatestSnapshot instead of
+	 * GetTransactionSnapshot since the latter would not reveal all writes
+	 * in serializable or repeatable read mode. Note that subsequent reads
+	 * from the distributed table would reveal those writes, temporarily
+	 * violating the isolation level. However, this seems preferable over
+	 * dropping the writes entirely.
+	 */
+	PushActiveSnapshot(GetLatestSnapshot());
+
 	/* get the table columns */
 	tupleDescriptor = RelationGetDescr(distributedRelation);
 	slot = MakeSingleTupleTableSlot(tupleDescriptor);
@@ -827,6 +838,8 @@ CopyLocalDataIntoShards(Oid distributedRelationId)
 	ExecDropSingleTupleTableSlot(slot);
 	FreeExecutorState(estate);
 	heap_close(distributedRelation, NoLock);
+
+	PopActiveSnapshot();
 }
 
 
--- a/src/test/regress/specs/isolation_data_migration.spec
+++ b/src/test/regress/specs/isolation_data_migration.spec
@@ -1,0 +1,73 @@
+setup
+{
+	CREATE TABLE migration_table (test_id integer NOT NULL, data text);
+}
+
+teardown
+{
+	DROP TABLE migration_table;
+}
+
+session "s1"
+
+step "s1-begin"
+{
+	BEGIN;
+}
+
+step "s1-begin-serializable"
+{
+	BEGIN TRANSACTION ISOLATION LEVEL SERIALIZABLE;
+	SELECT 1;
+}
+
+step "s1-create_distributed_table"
+{
+	SELECT create_distributed_table('migration_table', 'test_id');
+}
+
+step "s1-commit"
+{
+	COMMIT;
+}
+
+session "s2"
+
+step "s2-begin"
+{
+	BEGIN;
+}
+
+step "s2-copy"
+{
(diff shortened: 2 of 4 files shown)


anarazel:
Note that this'd probably still not get a correct answer if the transaction were to use repeatable-read or serializable snapshot isolation.  It's probably fine not to care, but I wonder if we should take note somewhere?

marcocitus:
I guess the fix would be to error out in a transaction block when not in read committed mode?

marcocitus:
No wait, that wouldn't help since the snapshot comes from the outer SELECT. Any other ways? (error out completely?)

anarazel:
I think just using `GetLatestSnapshot()` instead of `GetTransactionSnapshot()` should do the trick here.  That'll give a recent MVCC snapshot, even in repeatable-read/serializable.

marcocitus:
Ah, nice! Switched to `GetLatestSnapshot` and added an isolation test.

src/test/regress/specs/isolation_data_migration.spec
```diff
@@ -0,0 +1,73 @@
...
+
+step "s1-begin"
+{
+    BEGIN;
+}
+
+step "s1-begin-serializable"
+{
+    BEGIN TRANSACTION ISOLATION LEVEL SERIALIZABLE;
+	SELECT 1;
```
metdos: Why do we need `SELECT` here?

metdos:
@anarazel, @marcocitus are there any developer documentation about snapshots other than snapmgr.c itself? I couldn't find any readme in that folder.

@marcocitus, regressions look good - I just added a question on  s1-begin-serializable.

src/test/regress/specs/isolation_data_migration.spec
marcocitus: We don't get a snapshot until the first statement, so if we don't do anything before then, we'll see all preceding writes regardless of whether we're in serializable mode or not. If we're in serializable mode, the second statement will normally see the same copy of the database as the first statement. In read committed mode, we'd writes that have been committed since.

marcocitus:
https://www.postgresql.org/docs/current/static/transaction-iso.html gives a fairly comprehensive explanation of the different isolation levels.

src/test/regress/specs/isolation_data_migration.spec
metdos: So `SELECT 1` is to get a fix snapshot? Becuase `SELECT 1` does not update anything, so I don't see the difference between serializable mode and read committed mode. Maybe we can talk more on this tomorrow morning :)

metdos:
> https://www.postgresql.org/docs/current/static/transaction-iso.html gives a fairly comprehensive explanation of the different isolation levels.

Thanks, I read it also before and it is useful. I was more curious about how did you get the idea of using `PushActiveSnapshot()/PopActiveSnapshot()` other than reading PostgreSQL source code?

marcocitus:
Reading PostgreSQL source code :)

anarazel:
On 2017-05-10 07:49:59 -0700, Metin Döşlü wrote:
> +step "s1-begin-serializable"
> +{
> +    BEGIN TRANSACTION ISOLATION LEVEL SERIALIZABLE;
> +	SELECT 1;
> 
> Why do we need `SELECT` here?

The transaction snapshots in repateable read/serializable is only
acquired at the first command in a transaction.

src/test/regress/specs/isolation_data_migration.spec
```diff
@@ -0,0 +1,73 @@
...
+}
+
+step "s2-commit"
+{
+    COMMIT;
+}
+
+step "s2-select"
+{
+	SELECT * FROM migration_table;
```
anarazel: Realize this has only a single row for now, but it seems better to start with an ORDER BY here.

src/backend/distributed/commands/create_distributed_table.c
```diff
@@ -762,6 +762,9 @@ CopyLocalDataIntoShards(Oid distributedRelationId)
 	/* take an ExclusiveLock to block all operations except SELECT */
 	distributedRelation = heap_open(distributedRelationId, ExclusiveLock);
 
+	/* all writes have finished, make sure we can see them */
+	PushActiveSnapshot(GetLatestSnapshot());
```
anarazel: I'd append something like 'we can see them, even when using a transaction snapshot' - It's probably not obvious for the casual reader that GetTransactionSnapshot() would mean something different than GetLatestSnapshot().

PR #1406, opened 2017-05-10
Accept invalidation messages before accessing the metadata cache

### LINKED ISSUE #1280: Metadata is not refreshed after taking shard metadata locks
We call `LockShardDistributionMetadata` in various places to protect against concurrent modifications of the placement list (e.g. `master_move_shard_placement`, `master_copy_shard_placement`). We do so before reading the placement metadata, expecting that the metadata may have changed once the lock is released, but typically after reading the list of shards from the metadata, since that is how we learn which shard ID to lock. 

In b813b392 we started caching placement metadata immediately when calling `DistributedTableCacheEntry` to simplify dealing with co-located shard placements. However, we now read and cache the placement metadata before taking the appropriate locks and return it from functions such as ShardPlacementList after obtaining the lock.

It's a little tricky to reproduce, but this seems to do it somewhat reliably for me. 
```
Session 1:
SET citus.shard_count TO 2;
UPDATE test SET y = 5 WHERE x = 5;
BEGIN;
SELECT count(*) FROM test WHERE x = 5;

Session 2:
 SELECT * FROM pg_dist_shard_placement WHERE shardid = get_shard_id_for_distribution_column('test', 5);
 shardid | shardstate | shardlength | nodename  | nodeport | placementid
---------+------------+-------------+-----------+----------+-------------
  102180 |          1 |           0 | localhost |     9700 |         219
(1 row)
BEGIN;
SELECT master_move_shard_placement(102180, 'localhost', 9700, 'localhost', 9701);

Session 1:
UPDATE test SET y = 5 WHERE x = 5;

Session 2:
COMMIT:

Session 1:
WARNING:  relation "public.test_102180" does not exist
```

For replication factor > 1, this could lead to inconsistent shard placements.

We should maybe explicitly `AcceptInvalidationMessages()` prior to reading from the metadata.

### LINKED ISSUE #1392: Copying to reference table causes data inconsistency when a new node is added to cluster concurrently

### LINKED ISSUE #1404: Consider accepting cache invalidations on lock acqusitions
With #1392, we realized that Citus could sometimes skip getting the cache invalidations during a command planning/execution, which may lead to working on stale metadata.

I'm opening this issue to track the fix. Also, moving some of the earlier conversations:

- I wonder whether it makes sense to always do AcceptInvalidationMessages() before checking whether a cache entry is valid in metadata_cache.c. (  @marcocitus )

- Isn't the proper fix to add AcceptInvalidationMessages() to LockShardListResources etc? That's how it works for postgres... ( @anarazel )

- Well, I simply mean all our locking functions. Most of similar locking infrastructure, like e.g. LockRelationOid() (called by heap_open() via relation_open()) have a block like:
```
	/*
	 * Now that we have the lock, check for invalidation messages, so that we
	 * will update or flush any stale relcache entry before we try to use it.
	 * RangeVarGetRelid() specifically relies on us for this.  We can skip
	 * this in the not-uncommon case that we already had the same type of lock
	 * being requested, since then no one else could have modified the
	 * relcache entry in an undesirable way.  (In the case where our own xact
	 * modifies the rel, the relcache update happens via
	 * CommandCounterIncrement, not here.)
	 */
	if (res != LOCKACQUIRE_ALREADY_HELD)
		AcceptInvalidationMessages();
```   
( @anarazel )

```
We should generally be following this order:

1. Take a lock on the table that prevents concurrent changes to shard metadata (pg_dist_shard) and table properties (pg_dist_partition)
2. Accept invalidation messages
3. Load table and shard metadata
4. Take a lock on the shard ID that prevents concurrent changes to shard placement metadata (pg_dist_shard_placement)
5. Accept invalidation messages
6. Load placement metadata

There is a question around what type of lock we should use in #1 and before changing shard metadata. We usually want to avoid blocking writes during operations like master_create_empty_shard, master_apply_delete_command or tenant isolation. We currently don't have such locks, which means concurrent queries might land on an already dropped shard if their metadata is stale, which doesn't break correctness. A friendlier, though slightly less performance alternative would be to take a more aggressive lock in these operations just before committing, such that no other transaction is past step #1 when the metadata change becomes visible.
``` 
@marcocitus 

I personally also think that applying Postgres' way of handling invalidations make sense where we accept invalidations once any session acquires lock.

Also, one more note from a conversation with @marcocitus is that the reason that this is not observed with concurrent INSERT / shard rebalancer might be that while deparsing query we open heap, which accepts the invalidations. (So, we luckily do not get into this situation. This is probably valid for various other cases.)

onderkalaci:
This commit is crucial to prevent stale metadata reads from the
cache. Without this commit, some of the operations may use stale
metadata which could end up with various bugs such as crashes,
inconsistent/lost data etc.

As an example, consider that a COPY operation is blocked on shard
metadata lock. Another concurrent session updates the metadata and
invalidates the cache. However, since Citus doesn't accept invalidations,
COPY continues with the stale metadata once it acquires the lock.

With this commit, we make sure that invalidation messages are accepted
just before accessing the metadata cache and preventing any operation to
use stale metadata.

Fixes #1404 
Fixes #1392 
Fixes #1280

### DIFF
--- a/src/test/regress/isolation_schedule
+++ b/src/test/regress/isolation_schedule
@@ -1,4 +1,10 @@
+test: isolation_add_node_vs_reference_table_operations
+
+# tests that change node metadata should precede 
+# isolation_cluster_management such that tests 
+# that come later can be parallelized
 test: isolation_cluster_management
-test: isolation_dml_vs_repair
+
+test: isolation_dml_vs_repair isolation_copy_placement_vs_copy_placement
 test: isolation_concurrent_dml isolation_data_migration
-test: isolation_drop_shards
+test: isolation_drop_shards isolation_copy_placement_vs_modification 
--- a/src/test/regress/specs/isolation_copy_placement_vs_copy_placement.spec
+++ b/src/test/regress/specs/isolation_copy_placement_vs_copy_placement.spec
@@ -1,0 +1,71 @@
+# we use 5 as the partition key value through out the test
+# so setting the corresponding shard here is useful
+setup
+{	
+	SET citus.shard_count TO 2;
+	SET citus.shard_replication_factor TO 2;
+	CREATE TABLE test_hash_table (x int, y int);
+	SELECT create_distributed_table('test_hash_table', 'x');
+	
+	SELECT get_shard_id_for_distribution_column('test_hash_table', 5) INTO selected_shard_for_test_table;
+}
+
+teardown
+{
+	DROP TABLE test_hash_table;
+	DROP TABLE selected_shard_for_test_table;
+}
+
+session "s1"
+
+# since test_hash_table has rep > 1 simple select query doesn't hit all placements
+# hence not all placements are cached
+# but with copy all placements are cached
+step "s1-load-cache"
+{
+	COPY test_hash_table FROM PROGRAM 'echo "1,1\n2,2\n3,3\n4,4\n5,5"' WITH CSV;
+}
+
+step "s1-repair-placement"
+{
+	SELECT master_copy_shard_placement((SELECT * FROM selected_shard_for_test_table), 'localhost', 57637, 'localhost', 57638);
+}
+
+session "s2"
+
+step "s2-begin"
+{
+	BEGIN;
+}
+
+step "s2-set-placement-inactive"
+{
+	UPDATE pg_dist_shard_placement SET shardstate = 3 WHERE shardid IN (SELECT * FROM selected_shard_for_test_table) AND nodeport = 57638;
+}
+
+step "s2-repair-placement"
+{
+	SELECT master_copy_shard_placement((SELECT * FROM selected_shard_for_test_table), 'localhost', 57637, 'localhost', 57638);
+}
+
+# since test_hash_table has rep > 1 simple select query doesn't hit all placements
+# hence not all placements are cached
+# but with copy all placements are cached
+step "s2-load-cache"
+{
+	COPY test_hash_table FROM PROGRAM 'echo "1,1\n2,2\n3,3\n4,4\n5,5"' WITH CSV;
+}
+
+step "s2-commit"
+{
+	COMMIT;
+}
+
+# two concurrent shard repairs on the same shard
+# note that "s1-repair-placement" errors out but that is expected
+# given that "s2-repair-placement" succeeds and the placement is
+# already repaired
+permutation "s1-load-cache" "s2-load-cache" "s2-set-placement-inactive" "s2-begin" "s2-repair-placement" "s1-repair-placement" "s2-commit"
+
+# the same test without the load caches
+permutation "s2-set-placement-inactive" "s2-begin" "s2-repair-placement" "s1-repair-placement" "s2-commit"
--- a/src/test/regress/specs/isolation_copy_placement_vs_modification.spec
+++ b/src/test/regress/specs/isolation_copy_placement_vs_modification.spec
@@ -1,0 +1,127 @@
+# we use 5 as the partition key value through out the test
+# so setting the corresponding shard here is useful
+setup
+{	
+	SET citus.shard_count TO 2;
+	SET citus.shard_replication_factor TO 2;
+	CREATE TABLE test_table (x int, y int);
+	SELECT create_distributed_table('test_table', 'x');
+	
+	SELECT get_shard_id_for_distribution_column('test_table', 5) INTO selected_shard;
+}
+
+teardown
+{
+	DROP TABLE test_table;
+	DROP TABLE selected_shard;
+}
+
+session "s1"
+
+step "s1-begin"
+{
+    BEGIN;
+}
+
+# since test_table has rep > 1 simple select query doesn't hit all placements
+# hence not all placements are cached 
+step "s1-load-cache"
+{
+	TRUNCATE test_table;
+}
+
+step "s1-insert"
+{
+	INSERT INTO test_table VALUES (5, 10);
+}
+
+step "s1-update"
+{
+	UPDATE test_table SET y = 5 WHERE x = 5;
+}
+
+step "s1-delete"
+{
+	UPDATE test_table SET y = 5 WHERE x = 5;
+}
+
+step "s1-select"
+{
+	SELECT count(*) FROM test_table WHERE x = 5;
+}
+
+step "s1-ddl"
+{
+	CREATE INDEX test_table_index ON test_table(x);
+}
+
+step "s1-copy"
+{
+	COPY test_table FROM PROGRAM 'echo "1,1\n2,2\n3,3\n4,4\n5,5"' WITH CSV;
+}
+
+step "s1-commit"
+{
+	COMMIT;
+}
+
+session "s2"
+
+step "s2-begin"
+{
+	BEGIN;
+}
+
+step "s2-set-placement-inactive"
+{
+	UPDATE pg_dist_shard_placement SET shardstate = 3 WHERE shardid IN (SELECT * FROM selected_shard) AND nodeport = 57638;
+}
+
+step "s2-repair-placement"
+{
+	SELECT master_copy_shard_placement((SELECT * FROM selected_shard), 'localhost', 57637, 'localhost', 57638);
+}
+
+step "s2-commit"
+{
+	COMMIT;
+}
+
+step "s2-print-content"
+{
+	SELECT 
+		nodeport, success, result 
+	FROM 
+		run_command_on_placements('test_table', 'select y from %s WHERE x = 5')
+	WHERE
+		shardid IN (SELECT * FROM selected_shard)
+	ORDER BY
+		nodeport;
+}
+
+step "s2-print-index-count"
+{
+	SELECT 
+		nodeport, success, result 
+	FROM 
+		run_command_on_placements('test_table', 'select count(*) from pg_indexes WHERE tablename = ''%s''')
+	ORDER BY
+		nodeport;
+}
+
+# repair a placement while concurrently performing an update/delete/insert/copy
+# note that at some points we use "s1-select" just after "s1-begin" given that BEGIN
+# may invalidate cache at certain cases
+permutation "s1-load-cache" "s1-insert" "s1-begin" "s1-select" "s2-set-placement-inactive" "s2-begin" "s2-repair-placement" "s1-update" "s2-commit" "s1-commit" "s2-print-content"
+permutation "s1-load-cache" "s1-insert" "s1-begin" "s1-select" "s2-set-placement-inactive" "s2-begin" "s2-repair-placement" "s1-delete" "s2-commit" "s1-commit" "s2-print-content"
+permutation "s1-load-cache" "s1-begin" "s1-select" "s2-set-placement-inactive" "s2-begin" "s2-repair-placement" "s1-insert" "s2-commit" "s1-commit" "s2-print-content"
+permutation "s1-load-cache" "s1-begin" "s1-select" "s2-set-placement-inactive" "s2-begin" "s2-repair-placement" "s1-copy" "s2-commit" "s1-commit" "s2-print-content"
+permutation "s1-load-cache" "s1-begin" "s1-select" "s2-set-placement-inactive" "s2-begin" "s2-repair-placement" "s1-ddl" "s2-commit" "s1-commit" "s2-print-index-count"
+
+
+# the same tests without loading the cache at first
+permutation "s1-insert" "s1-begin" "s1-select" "s2-set-placement-inactive" "s2-begin" "s2-repair-placement" "s1-update" "s2-commit" "s1-commit" "s2-print-content"
+permutation "s1-insert" "s1-begin" "s1-select" "s2-set-placement-inactive" "s2-begin" "s2-repair-placement" "s1-delete" "s2-commit" "s1-commit" "s2-print-content"
+permutation "s1-begin" "s1-select" "s2-set-placement-inactive" "s2-begin" "s2-repair-placement" "s1-insert" "s2-commit" "s1-commit" "s2-print-content"
+permutation "s1-begin" "s1-select" "s2-set-placement-inactive" "s2-begin" "s2-repair-placement" "s1-copy" "s2-commit" "s1-commit" "s2-print-content"
+permutation "s1-begin" "s1-select" "s2-set-placement-inactive" "s2-begin" "s2-repair-placement" "s1-ddl" "s2-commit" "s1-commit" "s2-print-index-count"
--- a/src/backend/distributed/utils/metadata_cache.c
+++ b/src/backend/distributed/utils/metadata_cache.c
@@ -385,14 +385,23 @@ LookupShardCacheEntry(int64 shardId)
 
 		recheck = true;
 	}
-	else if (!shardEntry->tableEntry->isValid)
+	else
 	{
 		/*
-		 * The cache entry might not be valid right now. Reload cache entry
-		 * and recheck (as the offset might have changed).
+		 * We might have some concurrent metadata changes. In order to get the changes,
+		 * we first need to accept the cache invalidation messages.
 		 */
-		LookupDistTableCacheEntry(shardEntry->tableEntry->relationId);
-		recheck = true;
+		AcceptInvalidationMessages();
+
+		if (!shardEntry->tableEntry->isValid)
+		{
+			/*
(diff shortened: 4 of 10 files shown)


onderkalaci:
Depending on where we merge #1403, we may need to move the regression test to here. And potentially add some more tests.

marcocitus:
Thanks, the fix looks great.

It would be nice to add a whole bunch of isolation tests. I'm doing the same for the pg_dist_node locking issues.
https://github.com/citusdata/citus/compare/fix_node_locking?expand=1

marcocitus:
Should we also add a call to `GetWorkerNodeHash`?

src/test/regress/specs/isolation_add_node_vs_copy_reference_table.spec
```diff
@@ -0,0 +1,88 @@
+# initially remove all existing nodes and then
+# add single one of the nodes for the purpose of the test
+setup
+{	
+	SELECT master_remove_node(nodename, nodeport) FROM pg_dist_node;
```
marcocitus: I think it's ok to assume tests do a good job and leave an empty `pg_dist_node`. Removing this will help make sure.

src/test/regress/specs/isolation_add_node_vs_copy_reference_table.spec
```diff
@@ -0,0 +1,88 @@
...
+
+	SELECT nodename, nodeport, isactive FROM master_add_node('localhost', 57637);
```
marcocitus: Spacing seems slight off with lines below.

src/test/regress/specs/isolation_add_node_vs_copy_reference_table.spec
```diff
@@ -0,0 +1,88 @@
...
+step "s1-commit"
+{
+    COMMIT;
+}
+
+session "s2"
+
+step "s2-load-metadata-cache"
+{
+	COPY test_reference_table FROM PROGRAM 'echo "1\n2\n3\n4\n5"';
```
marcocitus: Is the COPY necessary?

src/test/regress/specs/isolation_add_node_vs_copy_reference_table.spec
```diff
@@ -0,0 +1,88 @@
...
+
+	SELECT count(*) FROM test_reference_table LIMIT 0;
```
marcocitus: Why LIMIT 0? Seems like something we might one day optimise away.

src/test/regress/specs/isolation_add_node_vs_copy_reference_table.spec
```diff
@@ -0,0 +1,88 @@
...
+	ORDER BY
+		nodeport;
+}
+
+
+# verify that copy gets the invalidation and re-builts its metadata cache
+# note that we need to run  "s1-load-metadata-cache" and "s2-load-metadata-cache" 
+# to ensure that metadata is cached otherwise the test would be useless since 
+# the cache would be empty and the metadata data is gathered from the tables directly
+permutation "s1-load-metadata-cache" "s2-load-metadata-cache" "s1-begin" "s1-add-second-worker" "s2-begin" "s2-copy-to-reference-table" "s1-commit" "s2-commit" "s2-print-content"
```
marcocitus: `"s2-begin"` and `"s2-commit"` are redundant, maybe `"s1-load-metadata-cache"` as well

marcocitus:
Also fixes #1280.

anarazel:
I'm doubtful this moves us forward architecturally: Without sane locking this doesn't improve the situation, with sane locking this shouldn't be necessary. Isn't this more likely to just hide bugs in locking protocols than anything?

src/test/regress/specs/isolation_add_node_vs_copy_reference_table.spec
onderkalaci: Yeah, that all makes sense.

src/test/regress/specs/isolation_add_node_vs_copy_reference_table.spec
onderkalaci: Just because we don't need the output. But, removing this line anyway.

src/test/regress/specs/isolation_add_node_vs_copy_reference_table.spec
onderkalaci: I think COPY is necessary especially because the table is reference table where COPY accesses all placement metadata. I added a comment for clarification. Removed `SELECT` query.

src/test/regress/specs/isolation_add_node_vs_copy_reference_table.spec
onderkalaci: Makes sense.

src/test/regress/isolation_schedule
```diff
@@ -1,4 +1,10 @@
+test: isolation_add_node_vs_copy_reference_table
+
+# the tests that changes node metadata should proceed
+# isolation_cluster_management so that the tests that
+# come later can be parallelized
```
marcocitus: grammar nazi: "tests that change node metadata should precede isolation_cluster_management such that tests that come later can be parallelized"

marcocitus:
Fortunately, the locking around this is already quite sane. The shard metadata locks were introduced long before the placement metadata cache. They seem like separate concepts.

> Isn't this more likely to just hide bugs in locking protocols than anything?

Technically, it really doesn't matter whether we accept invalidations just after getting a lock or just before checking cache validity. It just needs to be anywhere in between.

The reason for not putting it in the locking function is that it appears as magic to anyone who is not aware that of the fact that locking functions do that. Putting the call to `AcceptInvalidationMessages()` into the relevant function where we use the result of `AcceptInvalidationMessages()` makes it abundantly clear to readers where and how invalidation is  performed, which helps avoid bugs.

onderkalaci:
> I'm doubtful this moves us forward architecturally: Without sane locking this doesn't improve the situation, with sane locking this shouldn't be necessary. Isn't this more likely to just hide bugs in locking protocols than anything?

I cannot see any locking (except for locking the metadata cache itself) could solve this problem. So for instance, could you briefly explain a locking mechanism that could solve #1392 where one session accesses a stale metadata cache?

(The reason that I cannot come up with any locking is the following: Citus mostly acquires locks on (list of)`shardId`/`tableId`s, which means that in order to acquire the lock, the cache is already accessed and built. Thus, once a session is blocked, there is always the possibility that the cache is invalidated by a concurrent session. Are you considering changing the high-level locking or caching approach in Citus ?)

src/test/regress/specs/isolation_add_node_vs_copy_reference_table.spec
```diff
@@ -0,0 +1,69 @@
+# the test expects to have zero nodes in pg_dist_node at the beginning
+# add single one of the nodes for the purpose of the test
+setup
+{	
+	SELECT nodename, nodeport, isactive FROM master_add_node('localhost', 57637);
```
marcocitus: Alignment seems a little off from other lines.

src/test/regress/specs/isolation_add_node_vs_copy_reference_table.spec
```diff
@@ -0,0 +1,69 @@
...
+step "s1-commit"
+{
+    COMMIT;
+}
+
+session "s2"
+
+# COPY accesses all shard/placement metadata, so should be enough for
+# loading the cache
+step "s2-load-metadata-cache"
```
marcocitus: Do we need a separate item for this? It seems identical to the one below.

src/test/regress/specs/isolation_add_node_vs_copy_reference_table.spec
```diff
@@ -0,0 +1,69 @@
...
+	ORDER BY
+		nodeport;
+}
+
+
+# verify that copy gets the invalidation and re-builts its metadata cache
+# note that we need to run  "s1-load-metadata-cache" and "s2-load-metadata-cache" 
+# to ensure that metadata is cached otherwise the test would be useless since 
+# the cache would be empty and the metadata data is gathered from the tables directly
+permutation "s2-load-metadata-cache" "s1-begin" "s1-add-second-worker" "s2-copy-to-reference-table" "s1-commit" "s2-print-content"
```
marcocitus: Would be nice to also test:
```
permutation "s2-load-metadata-cache" "s2-begin" "s2-copy-to-reference-table" "s1-add-second-worker" "s2-commit" "s2-print-content"
```

and maybe repeat them without caching.

src/test/regress/specs/isolation_add_node_vs_copy_reference_table.spec
onderkalaci: Not strictly necessary, but seemed better to give a separate explicit name. Do you prefer using the same step for both?

marcocitus: Keeping it is fine.

src/test/regress/expected/isolation_add_node_vs_copy_reference_table.out
```diff
@@ -0,0 +1,152 @@
...
+	COPY test_reference_table FROM PROGRAM 'echo "1\n2\n3\n4\n5"';
+
+step s1-add-second-worker: 
+	SELECT nodename, nodeport, isactive FROM master_add_node('localhost', 57638);
+ <waiting ...>
+step s2-commit: 
+	COMMIT;
+ <waiting ...>
+step s2-commit: <... completed>
+step s1-add-second-worker: <... completed>
```
onderkalaci: Hmm, it that going to be solved with your PR? Shall we keep it as it is here now so that once we fix, we see that output changes?

marcocitus: Not sure. Yes, let's keep the test and open an issue.

onderkalaci:
Does anyone have any favorite isolation tests to add to his PR?

marcocitus:
Maybe 2 concurrent calls to master_copy_shard_placement on the same shard.

src/test/regress/isolation_schedule
```diff
@@ -1,4 +1,11 @@
...
+
+# tests that change node metadata should precede 
+# isolation_cluster_management such that tests 
+# that come later can be parallelized
 test: isolation_cluster_management
+
 test: isolation_dml_vs_repair
 test: isolation_concurrent_dml isolation_data_migration
 test: isolation_drop_shards
+test: isolation_copy_placement_vs_modification
```
marcocitus: Can this be run in parallel with other tests?

src/test/regress/specs/isolation_copy_placement_vs_modification.spec
```diff
@@ -0,0 +1,106 @@
...
+		nodeport, success, result 
+	FROM 
+		run_command_on_placements('test_table', 'select y from %s WHERE x = 5')
+	WHERE
+		shardid IN (SELECT * FROM selected_shard)
+	ORDER BY
+		nodeport;
+}
+
+permutation "s1-load-cache" "s1-insert" "s1-begin" "s1-select" "s2-set-placement-inactive" "s2-begin" "s2-repair-placement" "s1-update" "s2-commit" "s1-commit" "s2-print-content"
```
marcocitus: Brief one-line explanations of the tests would be nice, since it's not very readable, e.g.

```
# repair a placement while concurrently performing an update
```

src/test/regress/specs/isolation_copy_placement_vs_modification.spec
```diff
@@ -0,0 +1,106 @@
...
+permutation "s1-load-cache" "s1-insert" "s1-begin" "s1-select" "s2-set-placement-inactive" "s2-begin" "s2-repair-placement" "s1-delete" "s2-commit" "s1-commit" "s2-print-content"
```
marcocitus: `"s1-begin"`, `"s1-select"`, and "s1-commit" don't seem to add much, concurrency-testing wise. It would be nice to keep these tests as simple as possible.

src/test/regress/isolation_schedule
onderkalaci: Yes, I'll do that

src/test/regress/specs/isolation_copy_placement_vs_modification.spec
onderkalaci: I haven't verified the reason yet, but to re-produce #1280, we need those. Am I wrong?

marcocitus: Hmz, you're right, that is a little mysterious. Any idea why the preceding select matters?

onderkalaci: I think in some cases `BEGIN` triggers invalidation.

src/test/regress/specs/isolation_copy_placement_vs_copy_placement.spec
```diff
@@ -0,0 +1,71 @@
...
+}
+
+# two concurrent shard repairs on the same shard
+# note that "s1-repair-placement" errors out but that is expected
+# given that "s2-repair-placement" succeeds and the placement is
+# already repaired
+permutation "s1-load-cache" "s2-load-cache" "s2-set-placement-inactive" "s2-begin" "s2-repair-placement" "s1-repair-placement" "s2-commit"
+
+# the same test without the load caches
+permutation "s2-set-placement-inactive" "s2-begin" "s2-repair-placement" "s1-repair-placement" "s2-commit"
```
metdos: Minor: The end of the line is missing.

src/test/regress/specs/isolation_copy_placement_vs_modification.spec
metdos: Shall we add this explanation to the comments of the test for the future readers?

onderkalaci:
> Should we also add a call to GetWorkerNodeHash?

Hmm, you're right. I think we should add there as well. Do you have any further comments on this? I'm adding it right now. @marcocitus

src/test/regress/specs/isolation_copy_placement_vs_modification.spec
```diff
@@ -0,0 +1,125 @@
...
+{
+	SELECT 
+		nodeport, success, result 
+	FROM 
+		run_command_on_placements('test_table', 'select count(*) from pg_indexes WHERE tablename = ''%s''')
+	ORDER BY
+		nodeport;
+}
+
+# repair a placement while concurrently performing an update/delete/insert/copy
```
metdos: For all of these tests, we first run repair-placement and then run DML/DDL. Shall we replicate tests also with the reverse order?

onderkalaci: I think once we add more tests from the release test to here, we should add such tests.

However, here the main goal is to make sure that concurrent operations see the concurrent placement additions. On the other hand, the tests you suggest aims to ensure correct locking among concurrent operations, which we should add anyway.

If you have strong opinions on adding those tests now, I can do that.

metdos: I was thinking that concurrent placement additions also should see concurrent operations. Also, why not :) 

If you wish I can add them, and you can squash them into one comment.

metdos: I added tests at my local, but I think you're right. We can add them separately later. I will keep them :)

metdos:
All new tests look good, I think you can squash them.

jasonmp85:
I suppose we already released `6.1.1`, but how important is this backport-wise?

jasonmp85:
Did anyone run the isolation tests in 9.5? They fail.

jasonmp85:
Ok, third question 😁… is #1403 invalid now?

PR #1407, opened 2017-05-10
Fix pg_dist_node locking

### LINKED ISSUE #1397: Concurrent create_distributed_table and master_remove_node fails

### LINKED ISSUE #1400: Concurrent calls to master_remove_node fails with “tuple concurrently updated” error

### LINKED ISSUE #1401: Concurrent calls to master_remove_node and master_disable_node fails with “could not obtain lock on row in relation pg_dist_node” error
### Steps
- Connection 1
  - Starts transaction
  - Calls master_remove_node
  - Does not commit;
- Connection 2
  - Starts transaction
  - Calls master_disable_node
  - Blocked
- Connection 1 commits
- Connection 2 gets unblocked fails with “Error: could not obtain lock on row in relation pg_dist_node””

Same behavior is observed when master_remove_node and master_activate_node is called concurrently where remove is called first.

Related to #1400

marcocitus:
This PR adds slightly saner locking to node metadata changes by taking an exclusive lock on `pg_dist_node` changes, and a row share lock on `create_distributed_table` and `create_reference_table`, which cannot tolerate concurrent `pg_dist_node` changes, and an access share lock for normal reads. Recommendations for better lock combinations welcome.

It also adds isolation tests to test concurrent usage of pg_dist_node-changing functions, as well as pg_dist_node changes that run concurrently with `create_distributed_table`, `create_reference_table`, or insert into a reference table.

It's not meant to be a comprehensive solution for all pg_dist_node accesses, just to make it somewhat sane.

Fixes #1397 
Fixes #1400
Fixes #1401

### DIFF
--- a/src/backend/distributed/utils/node_metadata.c
+++ b/src/backend/distributed/utils/node_metadata.c
@@ -219,6 +219,9 @@ master_disable_node(PG_FUNCTION_ARGS)
 
 	CheckCitusVersion(ERROR);
 
+	/* take an exclusive lock on pg_dist_node to serialize pg_dist_node changes */
+	LockRelationOid(DistNodeRelationId(), ExclusiveLock);
+
 	workerNode = FindWorkerNodeAnyCluster(nodeName, nodePort);
 	if (workerNode == NULL)
 	{
@@ -351,28 +354,12 @@ PrimaryNodeForGroup(uint32 groupId, bool *groupContainsNodes)
 static Datum
 ActivateNode(char *nodeName, int nodePort)
 {
-	Relation pgDistNode = heap_open(DistNodeRelationId(), RowExclusiveLock);
-	HeapTuple heapTuple = GetNodeTuple(nodeName, nodePort);
-	CommandId commandId = GetCurrentCommandId(true);
-	LockTupleMode lockTupleMode = LockTupleExclusive;
-	LockWaitPolicy lockWaitPolicy = LockWaitError;
-	bool followUpdates = false;
-	Buffer buffer = 0;
-	HeapUpdateFailureData heapUpdateFailureData;
-
 	WorkerNode *workerNode = NULL;
 	bool isActive = true;
 	Datum nodeRecord = 0;
 
-	if (heapTuple == NULL)
-	{
-		ereport(ERROR, (errmsg("could not find valid entry for node \"%s:%d\"",
-							   nodeName, nodePort)));
-	}
-
-	heap_lock_tuple(pgDistNode, heapTuple, commandId, lockTupleMode, lockWaitPolicy,
-					followUpdates, &buffer, &heapUpdateFailureData);
-	ReleaseBuffer(buffer);
+	/* take an exclusive lock on pg_dist_node to serialize pg_dist_node changes */
+	LockRelationOid(DistNodeRelationId(), ExclusiveLock);
 
 	SetNodeState(nodeName, nodePort, isActive);
 
@@ -384,8 +371,6 @@ ActivateNode(char *nodeName, int nodePort)
 	}
 
 	nodeRecord = GenerateNodeTuple(workerNode);
-
-	heap_close(pgDistNode, NoLock);
 
 	return nodeRecord;
 }
@@ -400,7 +385,7 @@ Datum
 master_initialize_node_metadata(PG_FUNCTION_ARGS)
 {
 	ListCell *workerNodeCell = NULL;
-	List *workerNodes = NULL;
+	List *workerNodes = NIL;
 	bool nodeAlreadyExists = false;
 
 	/* nodeRole and nodeCluster don't exist when this function is caled */
@@ -409,7 +394,15 @@ master_initialize_node_metadata(PG_FUNCTION_ARGS)
 
 	CheckCitusVersion(ERROR);
 
+	/*
+	 * This function should only ever be called from the create extension
+	 * script, but just to be sure, take an exclusive lock on pg_dist_node
+	 * to prevent concurrent calls.
+	 */
+	LockRelationOid(DistNodeRelationId(), ExclusiveLock);
+
 	workerNodes = ParseWorkerNodeFileAndRename();
+
 	foreach(workerNodeCell, workerNodes)
 	{
 		WorkerNode *workerNode = (WorkerNode *) lfirst(workerNodeCell);
@@ -629,6 +622,9 @@ RemoveNodeFromCluster(char *nodeName, int32 nodePort)
 	EnsureCoordinator();
 	EnsureSuperUser();
 
+	/* take an exclusive lock on pg_dist_node to serialize pg_dist_node changes */
+	LockRelationOid(DistNodeRelationId(), ExclusiveLock);
+
 	workerNode = FindWorkerNodeAnyCluster(nodeName, nodePort);
 	if (workerNode == NULL)
 	{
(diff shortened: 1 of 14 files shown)


anarazel:
Caveat: I'm uncaffeinated as of now

I wonder whether we need a slightly more granular locking protocol.  It seems like it'll be useful to be able to update details of pg_dist_node rows, without changing the topology of the cluster.  Quite possibly deferrable to later, and it's not obvious to me how to do that using existing locklevels.  But it seems something we should think of first.

marcocitus:
I agree, though things are quite bad right now. For example, we're taking an AccessExclusiveLock on the table, and then releasing it immediately. It would be good to first make it work, and then improve it.

anarazel:
> I agree, though things are quite bad right now. For example, we're taking an AccessExclusiveLock on the table, and then releasing it immediately. It would be good to first make it work, and then improve it.

Yea, I agree it's reasonable to work incrementally. Still think a picture of where we're want to go is good.

src/test/regress/specs/isolation_add_remove_node.spec
```diff
@@ -0,0 +1,134 @@
...
+{
+    SELECT master_remove_node(nodename, nodeport) FROM pg_dist_node;
+}
+
+session "s1"
+
+step "s1-begin"
+{
+    BEGIN;
+}
```
anarazel: Minor: Inconsistent indentation.

src/test/regress/specs/isolation_add_remove_node.spec
```diff
@@ -0,0 +1,134 @@
...
+
+# disable and activate an active node from 2 transactions, should be ok
+permutation "s1-add-node-1" "s1-begin" "s1-disable-node-1" "s2-activate-node-1" "s1-commit" "s1-show-nodes"
+# activate and disable an active node node from 2 transactions, should be ok
+permutation "s1-add-node-1" "s1-begin" "s1-activate-node-1" "s2-disable-node-1" "s1-commit" "s1-show-nodes"
+
+# disable and activate an inactive node from 2 transactions, should be ok
+permutation "s1-add-inactive-1" "s1-begin" "s1-disable-node-1" "s2-activate-node-1" "s1-commit" "s1-show-nodes"
+# activate and disable an inactive node node from 2 transactions, should be ok
+permutation "s1-add-inactive-1" "s1-begin" "s1-activate-node-1" "s2-disable-node-1" "s1-commit" "s1-show-nodes"
```
anarazel: Can we add a permutation or two that rolls back node addition? That certainly looked borked with the row-level locking, seems like a good idea to make sure it stays sane.

src/test/regress/specs/isolation_create_table_add_remove_node.spec
```diff
@@ -0,0 +1,101 @@
...
+	SELECT 1 FROM master_add_node('localhost', 57637);
+	SELECT * FROM master_get_active_worker_nodes() ORDER BY node_name, node_port;
+}
+
+teardown
+{
+	DROP TABLE IF EXISTS dist_table;
+
+    SELECT master_remove_node(nodename, nodeport) FROM pg_dist_node;
+}
```
anarazel: More nitpicky indentation stuff.

src/backend/distributed/utils/node_metadata.c
```diff
@@ -593,7 +595,7 @@ AddNodeMetadata(char *nodeName, int32 nodePort, int32 groupId, char *nodeRack,
 static void
 SetNodeState(char *nodeName, int32 nodePort, bool isActive)
 {
-	Relation pgDistNode = heap_open(DistNodeRelationId(), AccessShareLock);
+	Relation pgDistNode = heap_open(DistNodeRelationId(), RowExclusiveLock);
 	TupleDesc tupleDescriptor = RelationGetDescr(pgDistNode);
```
marcocitus: Yeh, we should maybe backport fixing this particular lock.

src/backend/distributed/utils/node_metadata.c
```diff
@@ -700,7 +701,7 @@ GenerateNodeTuple(WorkerNode *workerNode)
 	nodeDatum = HeapTupleGetDatum(heapTuple);
 
 	/* close the relation */
```
marcocitus: Removed.

src/test/regress/specs/isolation_add_remove_node.spec
marcocitus: Added some rollback tests to each of the test files.

onderkalaci:
A note on this PR. With #1413, we added an output file for 9.5 since there is a minor output/behavioral change on the isolation tester. But, we could probably be able to get rid of that extra file after this PR checked-in. So, could you please check that once this is rebased to the master?

onderkalaci:
Just a note: `LockMetadataSnapshot` acquires `AccessExclusiveLock` on `DistNodeRelationId()`, which you may consider fixing as well.

marcocitus:
> Just a note: LockMetadataSnapshot acquires AccessExclusiveLock on DistNodeRelationId(), which you may consider fixing as well.

It seems that we don't actually call that function so it can be removed.

mtuncer:
@marcocitus This branch does not build as it is due to changes in postgres master. Also there are merge conflicts that need resolving.

mtuncer:
While we are on it we should also fix locking semantics between create empty shard and remove node.

- create table tt1(a int);
- select master_create_distributed_table('tt1', 'a', 'append');

Connection 1:
```sql
begin;
select master_create_empty_shard('tt1');
select * from pg_dist_shard_placement;
```
Here record the node_name/node_port of newly created shard
Connection 2;
```sql
select master_remove_node(node_name, node_port)
```

Connection 1
```sql
commit;
```

Result:
- pg_dist_shard HAS the shard  information
- pg_dist_shard_placement DOES NOT have the shard information
- pg_dist_placement HAS the placement group information
- shard table is created in the removed node

Now you can not drop the table since there is a dangling shard information.

```sql
postgres=# drop table tt1;
ERROR:  the metadata is inconsistent
DETAIL:  there is a placement in group 2 but there are no nodes in that group
CONTEXT:  SQL statement "SELECT master_drop_all_shards(v_obj.objid, v_obj.schema_name, v_obj.object_name)"
PL/pgSQL function citus_drop_trigger() line 21 at PERFORM
```

src/backend/distributed/utils/node_metadata.c
```diff
@@ -678,18 +680,20 @@ AddNodeMetadata(char *nodeName, int32 nodePort, int32 groupId, char *nodeRack,
...
 	 */
-	LockRelationOid(DistNodeRelationId(), ShareRowExclusiveLock);
+	LockRelationOid(DistNodeRelationId(), ExclusiveLock);
 
 	/* check if the node already exists in the cluster */
 	workerNode = FindWorkerNode(nodeName, nodePort);
 	if (workerNode != NULL)
 	{
-		*nodeAlreadyExists = true;
+		/* fill return data and return */
```
mtuncer: rebase artifact, needs fixing

src/test/regress/specs/isolation_create_table_vs_add_remove_node.spec
```diff
@@ -0,0 +1,108 @@
...
+
+step "s2-create-table-1"
+{
+	SET citus.shard_count TO 4;
+	SET citus.shard_replication_factor TO 1;
+	CREATE TABLE dist_table (x int, y int);
+	SELECT create_distributed_table('dist_table', 'x');
+}
+
+step "s2-create-table-2"
```
mtuncer: we should also have append version of this. I added a failing test case to comments

marcocitus:
Fixed `master_create_empty_shard`

PR #1412, opened 2017-05-15
Send correct and quoted owner name while propagating schema creation

### LINKED ISSUE #1411: master_add_node fails when a different user have created a reference table in a different schema
Steps to reproduce:
1. Setup two node cluster
2. Create the following reference table in a schema as a regular user:
 ```CREATE SCHEMA tutorial;

CREATE TABLE tutorial.users (
  user_id uuid NOT NULL DEFAULT uuid_generate_v4(),
  email text NOT NULL UNIQUE,
  encrypted_password text NOT NULL ,
  created_at timestamptz NOT NULL DEFAULT now(),
  updated_at timestamptz NOT NULL DEFAULT now(),
  PRIMARY KEY (user_id)
);
```
3. Call master_add_node

Error on coordinator:
```
ERROR:  must be member of role "postgres"
CONTEXT:  while executing command on ec2-35-166-240-133.us-west-2.compute.amazonaws.com:5432
STATEMENT:  SELECT master_add_node('ec2-35-166-240-133.us-west-2.compute.amazonaws.com', 5432)
```

Error on worker:
```
ERROR:  must be member of role "postgres"
STATEMENT:  SELECT worker_apply_shard_ddl_command (102008, 'CREATE SCHEMA IF NOT EXISTS tutorial AUTHORIZATION postgres')
``` 

 ```citus=> \dn
   List of schemas
   Name   |  Owner   
----------+----------
 citus    | postgres
 public   | postgres
 tutorial | citus
(3 rows)
```
note that the schema is owned by citus user, not postgres user - but we incorrectly try to create it with the postgres user as owner

byucesoy:
Fixes #1411 

We incorrectly tried to use relation cache to find particular schema's owner and
when we cannot find the schema in the relation cache(i.e always), we automatically
used current user as the schema's owner. This means we always created schemas in
the data nodes with current user. With this patch we started to use namespace
cache to find schemas and create schemas with correct owner. We also start to quote
owner name while generating schema creation query.

### DIFF
--- a/src/test/regress/expected/multi_schema_support.out
+++ b/src/test/regress/expected/multi_schema_support.out
@@ -1084,3 +1084,62 @@ SET search_path TO public;
 ALTER TABLE test_schema_support.nation_hash SET SCHEMA public;
 WARNING:  not propagating ALTER ... SET SCHEMA commands to worker nodes
 HINT:  Connect to worker nodes directly to manually change schemas of affected objects.
+-- we will use this function in next test
+CREATE FUNCTION run_command_on_coordinator_and_workers(p_sql text)
+RETURNS void LANGUAGE plpgsql AS $$
+BEGIN
+     EXECUTE p_sql;
+     PERFORM run_command_on_workers(p_sql);
+END;$$;
+-- test schema propagation with user other than current user
+SELECT run_command_on_coordinator_and_workers('CREATE USER "test-user"');
+NOTICE:  not propagating CREATE ROLE/USER commands to worker nodes
+HINT:  Connect to worker nodes directly to manually create all necessary users and roles.
+CONTEXT:  SQL statement "CREATE USER "test-user""
+PL/pgSQL function run_command_on_coordinator_and_workers(text) line 3 at EXECUTE
+ run_command_on_coordinator_and_workers 
+----------------------------------------
+ 
+(1 row)
+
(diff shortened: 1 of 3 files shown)


src/test/regress/expected/multi_schema_support.out
```diff
@@ -1084,3 +1084,46 @@ SET search_path TO public;
 ALTER TABLE test_schema_support.nation_hash SET SCHEMA public;
 WARNING:  not propagating ALTER ... SET SCHEMA commands to worker nodes
 HINT:  Connect to worker nodes directly to manually change schemas of affected objects.
+-- test schema propagation with user other than current user
+CREATE USER "test-user";
```
onderkalaci: Maybe use `run_command_on_workers` ? Or even consider @anarazel 's suggestion [here](https://github.com/citusdata/citus/pull/1389#discussion_r115418487)? If you add such a function, you could consider adding that to a test file such that others can also use it. 

(In 7.0, we could add another commit to `citus_tools` so that we support `run_command_on_master_and_workers()` and remove this function here)

onderkalaci: > run_command_on_master_and_workers

And we prefer `coordinator` over `master`  😄  `run_command_on_coordinator_and_workers()` should be better

marcocitus:
Fix makes sense, after addressing Onder's suggestion this should be ready to ship.

PR #1413, opened 2017-05-15
Add 9.5 output file for isolation test

onderkalaci:
With this commit, we add one additional regression test output file which
has some output syntax differences with its 9.6 equivalence.

Note to the reviewer: The output difference is introduced to Postgres isolation tester on this [commit](https://github.com/postgres/postgres/commit/38f8bdcac4982215beb9f65a19debecaf22fd470#diff-da8e149e3cc96e8758cc82a484f81acaR547) to 9.6 release. From the changed test perspective, the outcome is the same where a deadlock is detected. But, in 9.5 the second session is not reported to be in `waiting` state whereas in 9.6 it is reported to be waiting.


metdos:
Looks good, verified that this solves the problem with 9.5. After adding a note about deadlocks we have discussed, this is good to 🚢

onderkalaci:
> After adding a note about deadlocks we have discussed, this is good to 


Added the note [there](https://github.com/citusdata/citus/pull/1407#issuecomment-301459710)

PR #1416, opened 2017-05-15
Use cached PostgreSQL build to reduce testing time

byucesoy:
With this PR, we started to cache custom compiled PostgreSQL builds. If there
are no new commits to the related PostgreSQL branches, we will use already
compiled binaries to reduce testing time.

We still need to compile PostgreSQL for each PR at least once, but subsequent
tests will use cached PostgreSQL build. That is because cache of each branch and
each PR is different. Actually, Travis uses parent branch's cache if current branch's
cache is not created yet, but we do not compile PostgreSQL for branches, we only
compile it for pull requests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,10 @@
 sudo: required
 dist: trusty
 language: c
-cache: apt
+cache:
+  apt: true
+  directories:
+  - /home/travis/postgresql
 branches:
   except: [ /^open-.*$/ ]
 env:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -14,7 +17,7 @@ env:
     - PGVERSION=9.5
     - PGVERSION=9.6
 before_install:
-  - git clone -b v0.6.1 --depth 1 https://github.com/citusdata/tools.git
+  - git clone -b use_cached_postgresql_build --depth 1 https://github.com/citusdata/tools.git
```
byucesoy: After merging [this PR](https://github.com/citusdata/tools/pull/16), we will change this line to clone correct tag.

marcocitus: Can that PR be merged now?

byucesoy: It just needs to be reviewed.

jasonmp85:
The `0.6.2` tag in the tools repo isn't part of `master`, which is where tags live. It's not even part of `develop`. It's part of a now-unnamed feature branch. This is bad `git` hygiene. Tags shouldn't be created lightly.

I have two branches I'm merging to `develop`, whereupon I'll merge `develop` to `master` and release a `0.6.2` myself, overwriting the old tag.

PR #1420, opened 2017-05-17
Bump extension and configure PACKAGE versions

jasonmp85:
Actually getting this done before the next dev cycle begins.


jasonmp85:
@byucesoy added you as reviewer since you suggested doing this as part of release. 7.0 is our next release, so I updated to that.

PR #1425, opened 2017-05-18
Fix aggressive error out behavior

### LINKED ISSUE #1422: Confusing ALTER EXTENSION citus UPDATE behaviour
The new `ALTER EXTENSION citus UPDATE` behaviour is quite confusing. 

An error is thrown even during non-citus commands. This is very dangerous, especially if the `ALTER EXTENSION` has some reason to error out (e.g. new function already exists).

```
postgres=# \dx
ERROR:  loaded Citus library version differs from installed extension version
DETAIL:  Loaded library requires 6.2, but the installed extension version is 6.1-17.
HINT:  Run ALTER EXTENSION citus UPDATE and try again.
```

It gives especially strange results when the workers have not yet been fully upgraded.

```
postgres=# ALTER EXTENSION citus update;
ALTER EXTENSION

postgres=# SELECT * FROM test WHERE x = 1;
WARNING:  loaded Citus library version differs from installed extension version
DETAIL:  Loaded library requires 6.2, but the installed extension version is 6.1-17.
HINT:  Run ALTER EXTENSION citus UPDATE and try again.
ERROR:  could not receive query results

# DROP TABLE test;
ERROR:  loaded Citus library version differs from installed extension version
DETAIL:  Loaded library requires 6.2, but the installed extension version is 6.1-17.
HINT:  Run ALTER EXTENSION citus UPDATE and try again.
SQL statement "SELECT master_drop_all_shards(v_obj.objid, v_obj.schema_name, v_obj.object_name)"
PL/pgSQL function citus_drop_trigger() line 21 at PERFORM

# ALTER EXTENSION citus UPDATE ;
NOTICE:  version "6.2-4" of extension "citus" is already installed
ALTER EXTENSION

```

Due to the way errors are propagated, the error messages make it seem as though the error comes from the coordinator, while it actually comes from the worker. It's not entirely clear why the worker even throws this error.

### LINKED ISSUE #1424: Upgrade from 5.2 -> 6.2 crashes
Repro steps:
- Install PG 9.5
- git checkout release-5.2
- make install
- create cluster, create a table, 
- master_create_distributed_table/master_create_worker_shards.
- load data
- git checkout release-6.2
- make install
- restart PG
- alter extension citus update (Crashes).

The error happens here: https://github.com/citusdata/citus/blob/master/src/backend/distributed/utils/reference_table_utils.c#L448. DistributedTableCacheEntry() returns NULL because Citus hasn't been loaded (its in an upgrade step), and we try to access that.

The call trace is:
```
0   citus.so 0x000000010a0231b9 ReferenceTableOidList + 105 (reference_table_utils.c:450)
1   citus.so 0x000000010a023089 ReplicateAllReferenceTablesToNode + 25 (reference_table_utils.c:128)
2   citus.so 0x000000010a021209 ActivateNode + 153 (node_metadata.c:382)
3   citus.so 0x000000010a021b8f master_initialize_node_metadata + 1071 (node_metadata.c:269)
```

byucesoy:
Fixes #1422 
Fixes #1424 

With this PR we made two main changes each in their respective commits.

We first fixed the crash occurring upgrading Citus 5.2 to Citus 6.2. To do that we remove ActivateNode call in master_initialize_node_metadata. In fact, we did not need to activate the nodes added via master_initialize_node_metadata anyway. Because this function is called **once** at either;

* upgrading from 5.2 to 6.0 and there are no reference tables or isactive column in Citus 5.2

or

* during extension creation and there are no reference tables created at that moment

The second commit is to only error out for distributed queries in case of version mismatch. Previously, we errored out on almost all kinds of queries because when there is a version mismatch it is not safe to use distributed cache which expects metadata tables in particular structure. With this change, we added a function to read pg_dist_partition metadata table directly without using distributed cache. In this function, we try to minimize our assumptions about pg_dist_partition. We only expect the name of the table, the position of the logicalrelid column and the name of the index on the logicalrelid column stays same across versions.

### DIFF
--- a/src/backend/distributed/commands/create_distributed_table.c
+++ b/src/backend/distributed/commands/create_distributed_table.c
@@ -110,6 +110,7 @@ master_create_distributed_table(PG_FUNCTION_ARGS)
 	bool requireEmpty = true;
 
 	EnsureCoordinator();
+	CheckCitusVersion(ERROR);
 
 	if (ReplicationModel != REPLICATION_MODEL_COORDINATOR)
 	{
@@ -147,6 +148,7 @@ create_distributed_table(PG_FUNCTION_ARGS)
 	char *colocateWithTableName = NULL;
 
 	EnsureCoordinator();
+	CheckCitusVersion(ERROR);
 
 	/* guard against a binary update without a function update */
 	if (PG_NARGS() >= 4)
@@ -233,13 +235,17 @@ static void
 CreateReferenceTable(Oid relationId)
 {
 	uint32 colocationId = INVALID_COLOCATION_ID;
-	List *workerNodeList = ActiveWorkerNodeList();
-	int replicationFactor = list_length(workerNodeList);
+	List *workerNodeList = NIL;
+	int replicationFactor = 0;
 	char *distributionColumnName = NULL;
 	bool requireEmpty = true;
 	char relationKind = 0;
 
 	EnsureCoordinator();
+	CheckCitusVersion(ERROR);
+
+	workerNodeList = ActiveWorkerNodeList();
+	replicationFactor = list_length(workerNodeList);
 
 	/* if there are no workers, error out */
 	if (replicationFactor == 0)
--- a/src/backend/distributed/executor/multi_utility.c
+++ b/src/backend/distributed/executor/multi_utility.c
@@ -99,7 +99,6 @@ struct DropRelationCallbackState
 
 
 /* Local functions forward declarations for deciding when to perform processing/checks */
-static bool SkipCitusProcessingForUtility(Node *parsetree);
 static bool IsCitusExtensionStmt(Node *parsetree);
 
 /* Local functions forward declarations for Transmit statement */
@@ -185,22 +184,28 @@ multi_ProcessUtility(Node *parsetree,
 	Oid savedUserId = InvalidOid;
 	int savedSecurityContext = 0;
 	List *ddlJobs = NIL;
-	bool skipCitusProcessing = SkipCitusProcessingForUtility(parsetree);
+	bool checkExtensionVersion = false;
 
-	if (skipCitusProcessing)
+	if (IsA(parsetree, TransactionStmt))
 	{
-		bool checkExtensionVersion = IsCitusExtensionStmt(parsetree);
-
+		/*
+		 * Transaction statements (e.g. ABORT, COMMIT) can be run in aborted
+		 * transactions in which case a lot of checks cannot be done safely in
+		 * that state. Since we never need to intercept transaction statements,
+		 * skip our checks and immediately fall into standard_ProcessUtility.
+		 */
 		standard_ProcessUtility(parsetree, queryString, context,
 								params, dest, completionTag);
 
-		if (EnableVersionChecks && checkExtensionVersion)
-		{
-			ErrorIfUnstableCreateOrAlterExtensionStmt(parsetree);
-		}
-
 		return;
 	}
+
+	checkExtensionVersion = IsCitusExtensionStmt(parsetree);
+	if (EnableVersionChecks && checkExtensionVersion)
+	{
+		ErrorIfUnstableCreateOrAlterExtensionStmt(parsetree);
+	}
+
 
 	if (!CitusHasBeenLoaded())
 	{
@@ -443,63 +448,6 @@ multi_ProcessUtility(Node *parsetree,
 		VacuumStmt *vacuumStmt = (VacuumStmt *) parsetree;
 
 		ProcessVacuumStmt(vacuumStmt, queryString);
-	}
-}
-
-
-/*
- * SkipCitusProcessingForUtility simply returns whether a given utility should
- * bypass Citus processing and checks and be handled exclusively by standard
- * PostgreSQL utility processing. At present, CREATE/ALTER/DROP EXTENSION,
- * ABORT, COMMIT, ROLLBACK, and SET (GUC) statements are exempt from Citus.
- */
-static bool
-SkipCitusProcessingForUtility(Node *parsetree)
-{
-	switch (parsetree->type)
-	{
-		/*
-		 * In the CitusHasBeenLoaded check, we compare versions of loaded code,
-		 * the installed extension, and available extension. If they differ, we
-		 * force user to execute ALTER EXTENSION citus UPDATE. To allow this,
-		 * CREATE/DROP/ALTER extension must be omitted from Citus processing.
-		 */
-		case T_DropStmt:
-		{
-			DropStmt *dropStatement = (DropStmt *) parsetree;
-
-			if (dropStatement->removeType != OBJECT_EXTENSION)
-			{
-				return false;
-			}
-		}
-
-		/* no break, fall through */
-
-		case T_CreateExtensionStmt:
-		case T_AlterExtensionStmt:
-
-		/*
-		 * Transaction statements (e.g. ABORT, COMMIT) can be run in aborted
-		 * transactions in which case a lot of checks cannot be done safely in
-		 * that state. Since we never need to intercept transaction statements,
-		 * skip our checks and immediately fall into standard_ProcessUtility.
-		 */
-		case T_TransactionStmt:
-
-		/*
-		 * Skip processing of variable set statements, to allow changing the
-		 * enable_version_checks GUC during testing.
-		 */
-		case T_VariableSetStmt:
-		{
-			return true;
-		}
-
-		default:
-		{
-			return false;
-		}
 	}
 }
 
@@ -1474,8 +1422,8 @@ ErrorIfUnstableCreateOrAlterExtensionStmt(Node *parsetree)
 		 * from the citus.control file. In case a new default is available, we
 		 * will force a compatibility check of the latest available version.
 		 */
-		availableExtensionVersion = NULL;
-		ErrorIfAvailableVersionMismatch();
+		citusVersionKnownCompatible = false;
+		CheckCitusVersion(ERROR);
 	}
 }
 
--- a/src/backend/distributed/utils/metadata_cache.c
+++ b/src/backend/distributed/utils/metadata_cache.c
@@ -106,8 +106,7 @@ static Oid workerHashFunctionId = InvalidOid;
 /* Citus extension version variables */
 bool EnableVersionChecks = true; /* version checks are enabled */
 
-char *availableExtensionVersion = NULL;
-static char *installedExtensionVersion = NULL;
+bool citusVersionKnownCompatible = false;
 
 /* Hash table for informations about each partition */
 static HTAB *DistTableCacheHash = NULL;
@@ -130,6 +129,7 @@ static ScanKeyData DistShardScanKey[1];
 
 
 /* local function forward declarations */
+static bool IsDistributedTableViaCatalog(Oid relationId);
 static ShardCacheEntry * LookupShardCacheEntry(int64 shardId);
 static DistTableCacheEntry * LookupDistTableCacheEntry(Oid relationId);
 static void BuildDistTableCacheEntry(DistTableCacheEntry *cacheEntry);
@@ -142,7 +142,8 @@ static bool HasUniformHashDistribution(ShardInterval **shardIntervalArray,
 									   int shardIntervalArrayLength);
 static bool HasUninitializedShardInterval(ShardInterval **sortedShardIntervalArray,
 										  int shardCount);
-static void ErrorIfInstalledVersionMismatch(void);
+static bool CheckInstalledVersion(int elevel);
+static bool CheckAvailableVersion(int elevel);
 static char * AvailableExtensionVersion(void);
 static char * InstalledExtensionVersion(void);
 static bool HasOverlappingShardInterval(ShardInterval **shardIntervalArray,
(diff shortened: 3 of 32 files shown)


src/backend/distributed/utils/metadata_cache.c
```diff
@@ -194,13 +195,68 @@ IsDistributedTable(Oid relationId)
...
-	cacheEntry = LookupDistTableCacheEntry(relationId);
+	/*
+	 * We will not use cache if there is a version mismatch, because while
+	 * creating the cache, we assume our metadata tables are in particular
+	 * structure. However, in version mismatch, this is not guaranteed, thus
+	 * we will read from pg_dist_partition table and we will only assume
+	 * specific details which are most likely do not change in long term.
+	 */
+	installedVersion = InstalledExtensionVersion();
+	if (!MajorVersionsCompatible(installedVersion, CITUS_EXTENSIONVERSION))
```
anarazel: This doesn't look good to me, InstalledExtensionVersion() is far from cheap.  I think you're going to have to integrate this with the state kept in CitusHasBeenLoaded(), i.e. once determined that things are compatible, it needs to be cached.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -185,23 +184,29 @@ multi_ProcessUtility(Node *parsetree,
 	Oid savedUserId = InvalidOid;
 	int savedSecurityContext = 0;
 	List *ddlJobs = NIL;
-	bool skipCitusProcessing = SkipCitusProcessingForUtility(parsetree);
```
byucesoy: I removed this function, because in the version we error out almost all queries, we added this to whitelist some of the DDLs like CREATE/ALTER/DROP extension... Now since we do not error out non-distributed queries, we do not need this.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -185,23 +184,29 @@ multi_ProcessUtility(Node *parsetree,
...
+	bool checkExtensionVersion = false;
 
-	if (skipCitusProcessing)
+	if (IsA(parsetree, TransactionStmt))
```
byucesoy: This check was originally here but we moved it to SkipCitusProcessingForUtility when we implemented it. I added it back.

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -194,13 +195,68 @@ IsDistributedTable(Oid relationId)
...
+	Relation pgDistPartition = heap_open(DistPartitionRelationId(), AccessShareLock);
+
+	ScanKeyInit(&scanKey[0], Anum_pg_dist_partition_logicalrelid,
+				BTEqualStrategyNumber, F_OIDEQ, ObjectIdGetDatum(relationId));
+
+	scanDescriptor = systable_beginscan(pgDistPartition,
+										DistPartitionLogicalRelidIndexId(),
+										indexOK, NULL, scanKeyCount, scanKey);
+
+	partitionTuple = systable_getnext(scanDescriptor);
```
anarazel: I'm a bit doubtful that it's a good idea to rely on DistPartitionRelationId(), DistPartitionLogicalRelidIndexId() etc here. But we can possibly defer that concern to a later point.

src/backend/distributed/commands/create_distributed_table.c
```diff
@@ -301,6 +301,9 @@ ConvertToDistributedTable(Oid relationId, char *distributionColumnName,
 	char relationKind = 0;
 	Var *distributionColumn = NULL;
 
+	/* we will not continue if there is a version mismatch */
+	ErrorIfInstalledVersionMismatch();
```
byucesoy: If there is a version mismatch and the table is distributed, we error out at IsDistributedTable and IsDistributedTable covers most of the cases. But here we also should error out of table is not distributed.

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -435,28 +491,14 @@ DistributedTableCacheEntry(Oid distributedRelationId)
 {
 	DistTableCacheEntry *cacheEntry = NULL;
 
-	/*
-	 * Can't be a distributed relation if the extension hasn't been loaded
-	 * yet. As we can't do lookups in nonexistent tables, directly return NULL
-	 * here.
-	 */
-	if (!CitusHasBeenLoaded())
+	if (!IsDistributedTable(distributedRelationId))
```
anarazel: This *doubles* the lookup cost in the common case that this actually is a distributed table. Again, I don't think we should do this for the normal case.

src/backend/distributed/commands/create_distributed_table.c
anarazel: This is fairly late though, at that point we'll already have written to the catalog and such (which could cause issues w/ a mismatch).  That's not sufficient - I think this needs to be around the places where we currently check EnsureCoordinator();

anarazel:
Oh, and I'd argue this should add some tests to multi_extension.sql.

src/backend/distributed/commands/create_distributed_table.c
```diff
@@ -301,6 +301,10 @@ ConvertToDistributedTable(Oid relationId, char *distributionColumnName,
 	char relationKind = 0;
 	Var *distributionColumn = NULL;
 
+	/* we will not continue if there is a version mismatch */
+	ErrorIfAvailableVersionMismatch();
+	ErrorIfInstalledVersionMismatch();
```
anarazel: I think your previous approach of combining these into one function made sense.

src/backend/distributed/utils/metadata_cache.c
byucesoy: InstalledExtensionVersion already caches the result and we set it to NULL in case of [CREATE/ALTER EXTENSION](https://github.com/citusdata/citus/blob/master/src/backend/distributed/utils/metadata_cache.c#L1072)

anarazel: Nevertheless, this is doing string parsing, still makes the function a lot slower in the common cached case.

src/backend/distributed/commands/create_distributed_table.c
```diff
@@ -111,6 +111,10 @@ master_create_distributed_table(PG_FUNCTION_ARGS)
 
 	EnsureCoordinator();
 
+	/* we will not continue if there is a version mismatch */
+	ErrorIfAvailableVersionMismatch();
+	ErrorIfInstalledVersionMismatch();
```
anarazel: Please combine into one wrapper function.

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -1119,30 +1205,101 @@ ErrorIfAvailableVersionMismatch(void)
...
+static bool
+AvailableVersionCompatible(void)
+{
+	char *availableVersion = NULL;
+	MemoryContext oldMemoryContext = NULL;
+
+	/* if we cached the result before, return it */
+	if (availableVersionCompatible != NULL)
+	{
+		return *availableVersionCompatible;
```
anarazel: Type mismatch.

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -1119,30 +1205,101 @@ ErrorIfAvailableVersionMismatch(void)
...
+	/* we will cache the result of comparision to prevent string parsing */
+	if (CacheMemoryContext == NULL)
+	{
+		CreateCacheMemoryContext();
+	}
+
+	oldMemoryContext = MemoryContextSwitchTo(CacheMemoryContext);
+
+	availableVersionCompatible = palloc0(sizeof(bool));
+	*availableVersionCompatible = MajorVersionsCompatible(availableVersion,
```
anarazel: A pointer here is pretty awkward - why don't you instead just redo the check as long as availableVersionCompatible == false, and set it to true the first time it passes?

src/backend/distributed/utils/metadata_cache.c
byucesoy: I cached the results of MajorVersionsCompatible :+1:

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -1070,16 +1116,59 @@ CitusHasBeenLoaded(void)
...
+			{
+				ereport(ERROR, (errmsg("loaded Citus library version differs from "
+									   "installed extension version"),
+								errdetail("Loaded library requires %s, but the installed "
+										  "extension version is %s.", CITUS_MAJORVERSION,
+										  InstalledExtensionVersion()),
+								errhint("Run ALTER EXTENSION citus UPDATE and try "
+										"again.")));
+			}
+		}
```
anarazel: Still think a single caching variable (like `bool citusVersionsCompatible`) would be better.  Set to false by default and when extension dropped, set to true when version matches.  As long as it's false do the expensive check, otherwise the slow stuff.

byucesoy: We show different error messages based on which version differs from loaded libraries' version. I am not sure whether we should combine them(?)

src/backend/distributed/utils/metadata_cache.c
byucesoy: Umm, how? Anyway I will change this to regular bool, not bool pointer.

src/backend/distributed/utils/metadata_cache.c
byucesoy: Makes sense.

src/backend/distributed/utils/metadata_cache.c
anarazel: What I'm saying is that you shouldn't individually cache the results of the different relevant versions, but instead have a single cache: citusVersionsCompatible. If that's false, always perform both checks in an uncached manner.

byucesoy: I think I got it. It is late here and my cognitive abilities are non-existent :smile:  It is best if I continue this tomorow.

byucesoy:
I also go forward and add checks for almost all of our UDFs for version mismatch. These UDFs should have been call CitusHasBeenLoaded. Was there a specific reson to not call it? I added this change as separate commit so we can drop if there is a reason that I am not aware of.

To not miss any UDF, I scanned citus--7.0-1.sql file, and here are some UDFs I did not add the checks explicitly;

No Citus metadata access (As far as I realized):
* citus_extradata_container
* column_name_to_column
* column_to_column_name
* lock_shard_metadata
* lock_shard_resources
* shard_name
* worker_find_block_local_path
* worker_foreign_file_path

Not C functions (What should we do with them?):
* citus_drop_trigger
* citus_truncate_trigger
* create_insert_proxy_for_table
* distributed_tables_colocated
* run_command_on_colocated_placements
* run_command_on_placements
* run_command_on_shards
* run_command_on_workers

Enterprise features (We should open separate PR in enterprise for these):
* isolate_tenant_to_new_shard
* master_move_shard_placement

I added checks to these, but I am not sure:
* master_dist_local_group_cache_invalidate
* master_dist_node_cache_invalidate
* master_dist_partition_cache_invalidate
* master_dist_placement_cache_invalidate
* master_dist_shard_cache_invalidate

src/backend/distributed/executor/multi_utility.c
```diff
@@ -1474,8 +1422,8 @@ ErrorIfUnstableCreateOrAlterExtensionStmt(Node *parsetree)
 		 * from the citus.control file. In case a new default is available, we
 		 * will force a compatibility check of the latest available version.
 		 */
-		availableExtensionVersion = NULL;
-		ErrorIfAvailableVersionMismatch();
+		citusVersionKnownCompatible = false;
+		CheckCitusVersion(ERROR);
```
anarazel: I guess this is intended to protect against somebody installing a new version while connected, and then trying to perform an update?

src/backend/distributed/worker/worker_merge_protocol.c
```diff
@@ -81,6 +81,7 @@ worker_merge_files_into_table(PG_FUNCTION_ARGS)
 	/* we should have the same number of column names and types */
 	int32 columnNameCount = ArrayObjectCount(columnNameObject);
 	int32 columnTypeCount = ArrayObjectCount(columnTypeObject);
+
```
anarazel: Not a problem, but why was this line changed?

src/backend/distributed/worker/worker_partition_protocol.c
```diff
@@ -112,6 +112,7 @@ worker_range_partition_table(PG_FUNCTION_ARGS)
 
 	/* first check that array element's and partition column's types match */
 	Oid splitPointType = ARR_ELEMTYPE(splitPointObject);
+
```
anarazel: Not a problem, but why was this line changed?

src/test/regress/sql/multi_extension.sql
```diff
@@ -78,6 +78,71 @@ ALTER EXTENSION citus UPDATE TO '6.1-15';
...
+        SELECT sum(shard_size) INTO table_size
+        FROM (
+                SELECT max(shardlength) AS shard_size
+                FROM pg_dist_shard JOIN pg_dist_shard_placement USING (shardid)
+                WHERE logicalrelid = table_name AND shardstate = 1
+                GROUP BY shardid
+        ) shard_sizes;
+
+        RETURN table_size;
+END;
```
anarazel: @byucesoy What does this test, and whats the benefit of actually having a meaningful function body?

src/test/regress/sql/multi_extension.sql
```diff
@@ -78,6 +78,71 @@ ALTER EXTENSION citus UPDATE TO '6.1-15';
 ALTER EXTENSION citus UPDATE TO '6.1-16';
 ALTER EXTENSION citus UPDATE TO '6.1-17';
 ALTER EXTENSION citus UPDATE TO '6.2-1';
+
+-- Test non-distributed queries work even in version mismatch
```
anarazel: Not sure I like this being part of the series of all update steps here..., but I don't feel strongly about it.

src/test/regress/sql/multi_extension.sql
byucesoy: It causes next ALTER EXTENSION citus UPDATE TO '6.2-2'; fail. Because we also create a function with same name. It tests we can continue to operate when altering extension failed. I added it because of [this issue](https://github.com/citusdata/citus/issues/1422).

jasonmp85: Yeah, it's a dummy test function… we should excise the body.

src/test/regress/sql/multi_extension.sql
```diff
@@ -78,6 +78,71 @@ ALTER EXTENSION citus UPDATE TO '6.1-15';
...
+4
+\.
+
+-- Test INSERT
+INSERT INTO version_mismatch_table(column1) VALUES(5);
+ 
+-- Test SELECT
+SELECT * FROM version_mismatch_table ORDER BY column1;
+
+-- Test SELECT from pg_catalog
```
jasonmp85: Can you issue a `\l` here instead? That's what causes problems with `pg_ctlcluster`, so maybe having it explicitly in the test could be good.

src/test/regress/sql/multi_extension.sql
jasonmp85: I'm fine with it, though after this patch goes in maybe it can live in my version tester matrix instead of here.

jasonmp85: (in fact that's where I think it should live after this is released, otherwise this'll be a bit weird as more version upgrade scripts are added and whatnot)

src/backend/distributed/executor/multi_utility.c
jasonmp85: Yeah, that was the intent of the _in case a new default is available_ comment.

byucesoy: Yes, basically we do not want users to run ALTER EXTENSION UPDATE before restarting the database first.

src/backend/distributed/worker/worker_merge_protocol.c
byucesoy: Oops, it remained from a change I revert back.  I will fix this with the test change.

jasonmp85:
This change looks good to me; tested it some over the weekend and didn't
have any trouble. I have no blocking feedback for :shipit: so feel free to
test more over in Istanbul.

--
*Jason Petersen*
Software Engineer | Citus Data
303.736.9255
jason@citusdata.com

On May 19, 2017, at 5:22 PM, Burak Yücesoy <notifications@github.com> wrote:

*@byucesoy* commented on this pull request.
------------------------------

In src/backend/distributed/worker/worker_merge_protocol.c
<https://github.com/citusdata/citus/pull/1425#discussion_r117592108>:

> @@ -81,6 +81,7 @@ worker_merge_files_into_table(PG_FUNCTION_ARGS)
 	/* we should have the same number of column names and types */
 	int32 columnNameCount = ArrayObjectCount(columnNameObject);
 	int32 columnTypeCount = ArrayObjectCount(columnTypeObject);
+

Oops, it remained from a change I revert back. I will fix this with the
test change.

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/citusdata/citus/pull/1425#discussion_r117592108>, or mute
the thread

src/test/regress/sql/multi_extension.sql
byucesoy: \l also prints collation and ctype information which is likely to be change among different systems. (For example my laptop uses tr_TR.UTF-8 but our servers uses en_US.UTF-8). Since you said you have no blocking feedback, I am skipping this one.

src/test/regress/sql/multi_extension.sql
byucesoy: I put the these tests after all upgrade steps.

jasonmp85:
> Enterprise features (We should open separate PR in enterprise for these):
> * isolate_tenant_to_new_shard
> * master_move_shard_placement

Uh, did we?

PR #1427, opened 2017-05-22
Fix version checks

byucesoy:
While testing upgrading, we encountered two issues. This PR fixes those issues;

* During version update, we called CheckInstalledVersion, which obviously fails because during version update it is expected to have version mismatch.

* During ` citus--5.2-4--6.0-1.sql` [update](https://github.com/citusdata/citus/blob/master/src/backend/distributed/citus--5.2-4--6.0-1.sql), we alter  `pg_dist_partition` to have `regclass` instead of `Oid` for `logicalrelid` column. That operation leads to rebuild the index defined on the `logicalrelid`. In some cases, Citus reads the stale `Oid` for the index, which breaks almost all operations on the database since Citus errors out unexpectedly.  We decided to read index `Oid` from the catalog instead of cache to solve this problem only for version mismatch checks.

Both of these problems can be reproduced by upgrading 5.2 -> 6.2. To reproduce the first problem do not specify version in ALTER EXTENSION;

Instead of;
~~`ALTER EXTENSION citus UPDATE;`~~
use;
`ALTER EXTENSION citus UPDATE TO `6.2-4`;`

To reproduce the second problem, there should not be any distributed tables in the system. After successful extension upgrade, just run non-distributed command or `\d`. That problem is easily re-producible on the workers.

### DIFF
--- a/src/backend/distributed/utils/metadata_cache.c
+++ b/src/backend/distributed/utils/metadata_cache.c
@@ -511,6 +511,11 @@ LookupDistTableCacheEntry(Oid relationId)
 		return NULL;
 	}
 
+	if (DistTableCacheHash == NULL)
+	{
+		InitializeDistTableCache();
+	}
+
 	/*
 	 * If the version is not known to be compatible, perform thorough check,
 	 * unless such checks are disabled.
@@ -536,11 +541,6 @@ LookupDistTableCacheEntry(Oid relationId)
 			/* incompatible, can't access cache, so return before doing so */
 			return NULL;
 		}
-	}
-
-	if (DistTableCacheHash == NULL)
-	{
-		InitializeDistTableCache();
 	}
 
 	cacheEntry = hash_search(DistTableCacheHash, hashKey, HASH_ENTER, &foundInCache);
--- a/src/backend/distributed/utils/metadata_cache.c
+++ b/src/backend/distributed/utils/metadata_cache.c
@@ -106,7 +106,7 @@ static Oid workerHashFunctionId = InvalidOid;
 /* Citus extension version variables */
 bool EnableVersionChecks = true; /* version checks are enabled */
 
-bool citusVersionKnownCompatible = false;
+static bool citusVersionKnownCompatible = false;
 
 /* Hash table for informations about each partition */
 static HTAB *DistTableCacheHash = NULL;
@@ -143,7 +143,6 @@ static bool HasUniformHashDistribution(ShardInterval **shardIntervalArray,
 static bool HasUninitializedShardInterval(ShardInterval **sortedShardIntervalArray,
 										  int shardCount);
 static bool CheckInstalledVersion(int elevel);
-static bool CheckAvailableVersion(int elevel);
 static char * AvailableExtensionVersion(void);
 static char * InstalledExtensionVersion(void);
 static bool HasOverlappingShardInterval(ShardInterval **shardIntervalArray,
@@ -1181,13 +1180,15 @@ CheckCitusVersion(int elevel)
  * this function logs an error with the specified elevel and returns false,
  * otherwise it returns true.
  */
-static bool
+bool
 CheckAvailableVersion(int elevel)
 {
 	char *availableVersion = NULL;
 
-	Assert(CitusHasBeenLoaded());
-	Assert(EnableVersionChecks);
+	if (!EnableVersionChecks)
+	{
+		return true;
+	}
 
 	availableVersion = AvailableExtensionVersion();
 
--- a/src/backend/distributed/executor/multi_utility.c
+++ b/src/backend/distributed/executor/multi_utility.c
@@ -1419,11 +1419,9 @@ ErrorIfUnstableCreateOrAlterExtensionStmt(Node *parsetree)
 	{
 		/*
 		 * No version was specified, so PostgreSQL will use the default_version
-		 * from the citus.control file. In case a new default is available, we
-		 * will force a compatibility check of the latest available version.
+		 * from the citus.control file.
 		 */
-		citusVersionKnownCompatible = false;
-		CheckCitusVersion(ERROR);
+		CheckAvailableVersion(ERROR);
 	}
 }
 
--- a/src/include/distributed/metadata_cache.h
+++ b/src/include/distributed/metadata_cache.h
@@ -18,7 +18,6 @@
 #include "utils/hsearch.h"
 
 extern bool EnableVersionChecks;
-extern bool citusVersionKnownCompatible;
 
 /*
  * Representation of a table's metadata that is frequently used for
@@ -80,6 +79,7 @@ extern void CitusInvalidateRelcacheByShardId(int64 shardId);
 
 extern bool CitusHasBeenLoaded(void);
 extern bool CheckCitusVersion(int elevel);
+extern bool CheckAvailableVersion(int elevel);
 bool MajorVersionsCompatible(char *leftVersion, char *rightVersion);
 
 /* access WorkerNodeHash */
--- a/src/test/regress/expected/multi_extension.out
+++ b/src/test/regress/expected/multi_extension.out
@@ -169,7 +169,31 @@ SET citus.enable_version_checks TO 'true';
 DROP FUNCTION citus_table_size(regclass);
 SET citus.enable_version_checks TO 'false';
 ALTER EXTENSION citus UPDATE TO '6.2-2';
+-- Test updating to the latest version without specifying the version number
+ALTER EXTENSION citus UPDATE;
 -- re-create in newest version
 DROP EXTENSION citus;
 \c
 CREATE EXTENSION citus;
+-- test cache invalidation in workers
+\c - - - :worker_1_port
+-- this will initialize the cache
+\d
+      List of relations
+ Schema | Name | Type | Owner 
+--------+------+------+-------
+(0 rows)
+
+DROP EXTENSION citus;
+SET citus.enable_version_checks TO 'false';
+CREATE EXTENSION citus VERSION '5.2-4';
+SET citus.enable_version_checks TO 'true';
+-- during ALTER EXTENSION, we should invalidate the cache
+ALTER EXTENSION citus UPDATE;
+-- if cache is invalidated succesfull, this \d should work without any problem
+\d
+      List of relations
+ Schema | Name | Type | Owner 
+--------+------+------+-------
+(0 rows)
+
--- a/src/test/regress/sql/multi_extension.sql
+++ b/src/test/regress/sql/multi_extension.sql
@@ -153,7 +153,25 @@ DROP FUNCTION citus_table_size(regclass);
 SET citus.enable_version_checks TO 'false';
 ALTER EXTENSION citus UPDATE TO '6.2-2';
 
+-- Test updating to the latest version without specifying the version number
+ALTER EXTENSION citus UPDATE;
+
 -- re-create in newest version
 DROP EXTENSION citus;
 \c
 CREATE EXTENSION citus;
+
+-- test cache invalidation in workers
+\c - - - :worker_1_port
+
+-- this will initialize the cache
+\d
+DROP EXTENSION citus;
+SET citus.enable_version_checks TO 'false';
+CREATE EXTENSION citus VERSION '5.2-4';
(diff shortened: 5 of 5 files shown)


src/backend/distributed/utils/metadata_cache.c
```diff
@@ -1510,6 +1517,27 @@ DistPartitionRelationId(void)
...
+ */
+static Oid
+DistPartitionRelationIdViaCatalog(void)
+{
+	const char *relationName = "pg_dist_partition";
+	Oid distPartitionRelationId = get_relname_relid(relationName, PG_CATALOG_NAMESPACE);
+
+	if (distPartitionRelationId == InvalidOid)
+	{
+		ereport(ERROR, (errmsg("cache lookup failed for %s, called to early?",
```
onderkalaci: I think the error messages should be re-worded. 

Also, I'd write another function which calls `get_relname_relid ` and does the check. Later, `DistPartitionRelationIdViaCatalog ` and `DistPartitionLogicalRelidIndexIdViaCatalog ` calls that function.

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -1181,13 +1184,17 @@ CheckCitusVersion(int elevel)
...
  */
-static bool
+bool
 CheckAvailableVersion(int elevel)
 {
 	char *availableVersion = NULL;
 
-	Assert(CitusHasBeenLoaded());
-	Assert(EnableVersionChecks);
+	if (citusVersionKnownCompatible ||
```
onderkalaci: I'm not able to follow this change. `CheckCitusVersion()` has the exact same `if` statement. Do we actually need this?

byucesoy: This check is for calling CheckAvailableVersion directly in ErrorIfUnstableCreateOrAlterExtensionStmt. We changed ErrorIfUnstableCreateOrAlterExtensionStmt to call this function instead of CheckCitusVersions, because CheckCitusVersions also perform checks for installed version, which we should not perform during ALTER EXTENSION (Because during  ALTER EXTENSION, it is expected that version will mismatch).

anarazel:
This seems like the wrong approach to me:
0. Some of this seems easily testable, and thus should be added as tests. 
1.  We add two functions with only minor differences
2. This doesn't solve the issue that the cache might not be coherent afterwards, no?

Why isn't the code at the tail end of InvalidateDistRelationCacheCallback() taking care of this?

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -2730,6 +2769,24 @@ TupleToShardInterval(HeapTuple heapTuple, TupleDesc tupleDescriptor, Oid interva
  * CachedRelationLookup performs a cached lookup for the relation
  * relationName, with the result cached in *cachedOid.
  */
+static Oid
```
jasonmp85: The function comment for this is dangerously wrong.

sumedhpathak:
@byucesoy Could you clarify the scenario in which both the issues happen?
For the first issue, if I *don't* specify the version, but just do `ALTER EXTENSION citus UPDATE`, then it errors? Is this from 5.2 -> 6.2 only? Or even for 6.1 -> 6.2?

For the second issue, if I have distributed tables in the system, then there is no issue? This then seems less severe, since i wouldn't expect an extension update without there being any distributed tables a common scenario?

marcocitus:
I cherry-picked these on top of release-6.2.

Trying to upgrade from 6.0 to 6.2:
```
postgres=# ALTER EXTENSION citus UPDATE;
ERROR:  loaded Citus library version differs from installed extension version
DETAIL:  Loaded library requires 6.2, but the installed extension version is 6.0-17.
HINT:  Run ALTER EXTENSION citus UPDATE and try again.
STATEMENT:  ALTER EXTENSION citus UPDATE;
ERROR:  loaded Citus library version differs from installed extension version
DETAIL:  Loaded library requires 6.2, but the installed extension version is 6.0-17.
HINT:  Run ALTER EXTENSION citus UPDATE and try again.
```

marcocitus:
Ah, never mind, was still using current release-6.2, without these changes.

marcocitus:
Not introduced by this PR, but `IsCitusExtensionStmt` should probably check whether an `ALTER EXTENSION` is doing an `UPDATE` or another command (`ADD`/`DROP`/`SET SCHEMA`).

https://www.postgresql.org/docs/current/static/sql-alterextension.html

src/backend/distributed/utils/metadata_cache.c
marcocitus: Doesn't `CitusHasBeenLoaded()` return false during `ALTER EXTENSION` because of `creating_extension`?

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -2727,6 +2766,25 @@ TupleToShardInterval(HeapTuple heapTuple, TupleDesc tupleDescriptor, Oid interva
...
+ * from cache is not safe.
+ */
+static Oid
+RelationLookupViaCatalog(const char *relationName)
+{
+	Oid relationId = get_relname_relid(relationName, PG_CATALOG_NAMESPACE);
+
+	if (relationId == InvalidOid)
+	{
+		ereport(ERROR, (errmsg("relation lookup failed for %s", relationName)));
```
marcocitus: Maybe "relation \"%s\" does not exist".

src/backend/distributed/utils/metadata_cache.c
byucesoy: We perform this check before PostgreSQL start to process ALTER EXTENSION. So creating_extension will be false at that moment and we will not incorrectly exit early.

marcocitus: Thanks, makes sense.

marcocitus: Might be worth adding this in a comment.

metdos:
> For the second issue, if I have distributed tables in the system, then there is no issue? This then seems less severe, since i wouldn't expect an extension update without there being any distributed tables a common scenario?

@byucesoy 

i. Could you add explicit steps to create this problem?

ii. Even if we create distributed tables on the master, there will not be distributed tables on the workers. Does this problem replicable if we create distributed tables on the master and upgrade worker nodes afterward?

anarazel:
@metdos @byucesoy Could you reply to https://github.com/citusdata/citus/pull/1427#issuecomment-303145268

metdos:
> Why isn't the code at the tail end of InvalidateDistRelationCacheCallback() taking care of this?

@anarazel, @byucesoy works on this and will write a more detailed reply, but the short answer is that we don't know.

byucesoy:
@anarazel 

> 0.  Some of this seems easily testable, and thus should be added as tests. 
I will ad tests merging this PR.

> 1.  We add two functions with only minor differences
This is fixed, but we can remove even these two functions when we understand the real problem.

> 2. This doesn't solve the issue that the cache might not be coherent afterwards, no?
Yes, it just "patches" the problem which I am not fan of. I am trying to understand why InvalidateDistRelationCacheCallback does not solve this issue, for some weird reason we do not reset distPartitionLogicalRelidIndexId while updating from 5.2-4 to 6.2. 

Exact reproduction steps;
* Checkout release-5.2 branch, compile, install
* Create a cluster(Do not create distributed tables)
* Checkout this branch, remove latest commit(git reset --hard 974b27b61b3385be0a8dbab8726f4588d60404bf), compile, install
* Restart server
* Connect to worker node
* Run \d
* ALTER EXTENSION citus UPDATE;
* Run \d --> Errors out

marcocitus:
It's because `LookupDistTableCacheEntry` returns before calling `InitializeDistTableCache` and therefore  `CacheRegisterRelcacheCallback` doesn't get called.

marcocitus:
Moving the call to `InitializeDistTableCache` to be above the version check seems fine.

byucesoy:
@marcocitus 
Thank you for the finding the root cause. I checked and verified this is the real reason and moving InitializeDistTableCache to be above the version check solves the problem. InitializeDistTableCache does not try to read our metadata tables, so it is safe to use before version checks.

I double checked PostgreSQL has different statement types for `ALTER EXTENSION ... ADD/DROP` and `ALTER EXTENSION ... SET SCHEMA`.

For `ALTER EXTENSION ... ADD/DROP`, statement type is T_AlterExtensionContentsStmt
For `ALTER EXTENSION ... SET SCHEMA`, statement type is T_AlterObjectSchemaStmt

I think this was the reason for not checking for them in the first place but I did not remembered initially when you first raised this issue.

@anarazel 

I think what @marcocitus stated is the root cause of this problem. I modified the PR as he suggested. I also added a test case for first case, but second case is little difficult to test in regression tests. I have an idea in my mind to be able to test it in the regression tests and will implement it today. I think this would address all your concerns.

@sumedhpathak 
Both issues are introduced with this [PR](https://github.com/citusdata/citus/pull/1425), because we made some significant changes in our logic. So they are not exist in previous versions. I also wrote reproduction steps of second problem in my previous comment, which is also solved.

sumedhpathak:
@anarazel Could you take a look and evaluate the proposed fix?

jasonmp85:
@sumedhpathak FWIW I've been lurking/evaluating this patch myself and am much happier with the change using @marcocitus's suggestion. Since @anarazel rejected the original change, I guess he needs to sign off on this new one, but I'm happy with the root cause analysis and how it has now been addressed.

I do want to know about that second test case though, as I haven't seen explicit repro instructions here.

sumedhpathak:
Thanks @jasonmp85 . Agreed that @anarazel should review since he's requested changes. If it looks good, we'll be good to test and package tomorrow.

byucesoy:
@anarazel I also added the test for second case.

@jasonmp85 reproduction steps of second case is in the bottom of [this comment](https://github.com/citusdata/citus/pull/1427#issuecomment-303404129).

src/backend/distributed/utils/metadata_cache.c
byucesoy: On second thought, I think this does no affect ALTER EXTENSION, but it causes exit early for CREATE EXTENSION for the same reason. We perform this check before extension creation. I removed this checks.

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -1181,13 +1180,22 @@ CheckCitusVersion(int elevel)
...
 {
 	char *availableVersion = NULL;
 
-	Assert(CitusHasBeenLoaded());
-	Assert(EnableVersionChecks);
+	/*
+	 * Normally CitusHasBeenLoaded() would return false because creating_extension
+	 * is set to true during ALTER EXTENSION ... UPDATE by PostgreSQL. However we hit
+	 * this function before PostgreSQL processes ALTER EXTENSION citus UPDATE. Therefore
+	 * we will not incorrectly return early because of the check below.
```
marcocitus: This comment is a bit hard to follow because `CitusHasBeenLoaded` doesn't get called anymore. I think we can just omit it now.

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -1181,13 +1180,16 @@ CheckCitusVersion(int elevel)
...
  */
-static bool
+bool
 CheckAvailableVersion(int elevel)
 {
 	char *availableVersion = NULL;
 
-	Assert(CitusHasBeenLoaded());
-	Assert(EnableVersionChecks);
+	if (citusVersionKnownCompatible ||
```
anarazel: Why are we checking the cache here?  It seems like it'd be better if we avoided that - in that case it seems like we could avoid resetting citusVersionKnownCompatible in multi_utility and make it private.  That seems like it'd be a significant improvement.

byucesoy: I think this will only help to eliminate for the first non-cached version check after CREATE/ALTER EXTENSION. However, it is nice that we can make citusVersionKnownCompatible private. I changed it.

onderkalaci:
I did some manual testing ( before and even after [this](https://github.com/citusdata/citus/pull/1427#discussion_r118271235) change), all looks good. @byucesoy

marcocitus:
As discussed with @sumedhpathak, merging it to get the package ready today.

Issue #1101, opened 2017-01-06
Convert various UDFs to new connection API


marcocitus:
Adding list of individual items to keep track.
- [x] master_append_table_to_shard (Brian) PR: https://github.com/citusdata/citus/pull/1149
- [x] master_expire_table_cache
- [x] master_run_on_worker  (@anarazel) PR : https://github.com/citusdata/citus/pull/863
- [x] DropShards (Burak) PR: #1136 
- [ ] WorkerCreateShard
- [x] master_update_shard_statistics (Brian) PR: https://github.com/citusdata/citus/pull/1125
- [x] master_copy_shard_placement + master_move_shard_placement(enterprise)
a related issue: https://github.com/citusdata/shard_rebalancer/issues/49
- [x] recover_prepared_transactions (@marcocitus) PR : https://github.com/citusdata/citus/pull/1109
- [x] ReplicateShardToAllWorkers

We have a number of internal functions that should start using the new connection & transaction management API. Doing so enables transaction safety (correct visibility & no deadlock when used in a transaction block), allows cancellation & rollback, avoids connection leaks, and will allow us to remove redundant code.

Currently, we have the following low-level functions that open connections for modifying or query shard placements with the callers listed below each function:

-  `ExecuteRemoteCommand` 
    - `master_expire_table_cache`
    - DropShards (`master_apply_delete_command`, `master_drop_all_shards`)
-  `ExecuteRemoteQuery`
    - `master_append_table_to_shard`
    - WorkerCreateShard (`create_distributed_table`, `master_create_worker_shards`, `master_create_empty_shard`)
-  `WorkerShardStats` 
    - `master_update_shard_statistics`
-  `ExecuteRemoteQueryOrCommand` and `ExecuteCommandsInParallelAndStoreResults`
    - `master_run_on_worker`
-  `SendCommandListToWorkerInSingleTransaction` (optional)
    - `master_copy_shard_placement`
    - ReplicateShardToAllWorkers (`master_add_node`, `upgrade_to_reference_table`)
- `RecoverWorkerTransactions` 
   - `recover_prepared_transactions`

`master_run_on_worker` should be converted to use the low-level connection API, but transaction safety is not currently in scope.

`SendCommandListToWorkerInSingleTransaction` already uses the low-level functions in the new connection API, but for transaction safety we might consider using the placement API from #1079. Alternatively, we can leave these as they are, since they're unlikely to appear in the same transaction as DDL/DML.

[...]

PR #1436, opened 2017-05-29
Convert create distributed table to new api with commits

velioglu:
Part of #1101 

Convert create_distributed_table UDF to new connection API. In order to do that, each connection is created using the new connection API. Creating sequences, schemas and append distributed tables are remained as non-transactional, because creating them transactionally causes deadlock problem.

### DIFF
--- a/src/backend/distributed/commands/create_distributed_table.c
+++ b/src/backend/distributed/commands/create_distributed_table.c
@@ -41,6 +41,7 @@
 #include "distributed/pg_dist_colocation.h"
 #include "distributed/pg_dist_partition.h"
 #include "distributed/reference_table_utils.h"
+#include "distributed/remote_commands.h"
 #include "distributed/worker_protocol.h"
 #include "distributed/worker_transaction.h"
 #include "executor/executor.h"
@@ -71,7 +72,7 @@ int ReplicationModel = REPLICATION_MODEL_COORDINATOR;
 static void CreateReferenceTable(Oid distributedRelationId);
 static void ConvertToDistributedTable(Oid relationId, char *distributionColumnName,
 									  char distributionMethod, char replicationModel,
-									  uint32 colocationId, bool requireEmpty);
+									  uint32 colocationId);
 static char LookupDistributionMethod(Oid distributionMethodOid);
 static Oid SupportFunctionForColumn(Var *partitionColumn, Oid accessMethodId,
 									int16 supportFunctionNumber);
@@ -85,6 +86,11 @@ static List * TupleDescColumnNameList(TupleDesc tupleDescriptor);
 #if (PG_VERSION_NUM >= 100000)
 static bool RelationUsesIdentityColumns(TupleDesc relationDesc);
 #endif
+
+static void EnsureSchemaExistsOnAllNodes(Oid relationId);
+static void EnsureLocalTableEmpty(Oid relationId);
+static void EnsureTableNotDistributed(Oid relationId);
+static void EnsureIsTableId(Oid relationId);
 
 /* exports for SQL callable functions */
 PG_FUNCTION_INFO_V1(master_create_distributed_table);
@@ -108,8 +114,9 @@ master_create_distributed_table(PG_FUNCTION_ARGS)
 
 	char *distributionColumnName = text_to_cstring(distributionColumnText);
 	char distributionMethod = LookupDistributionMethod(distributionMethodOid);
-	bool requireEmpty = true;
 
+	EnsureTableNotDistributed(distributedRelationId);
+	EnsureLocalTableEmpty(distributedRelationId);
 	EnsureCoordinator();
 	CheckCitusVersion(ERROR);
 
@@ -125,7 +132,7 @@ master_create_distributed_table(PG_FUNCTION_ARGS)
 
 	ConvertToDistributedTable(distributedRelationId, distributionColumnName,
 							  distributionMethod, REPLICATION_MODEL_COORDINATOR,
-							  INVALID_COLOCATION_ID, requireEmpty);
+							  INVALID_COLOCATION_ID);
 
 	PG_RETURN_VOID();
 }
@@ -182,7 +189,8 @@ create_distributed_table(PG_FUNCTION_ARGS)
 	/* if distribution method is not hash, just create partition metadata */
 	if (distributionMethod != DISTRIBUTE_BY_HASH)
 	{
-		bool requireEmpty = true;
+		EnsureTableNotDistributed(relationId);
+		EnsureLocalTableEmpty(relationId);
 
 		if (ReplicationModel != REPLICATION_MODEL_COORDINATOR)
 		{
@@ -193,7 +201,7 @@ create_distributed_table(PG_FUNCTION_ARGS)
 
 		ConvertToDistributedTable(relationId, distributionColumnName,
 								  distributionMethod, REPLICATION_MODEL_COORDINATOR,
-								  INVALID_COLOCATION_ID, requireEmpty);
+								  INVALID_COLOCATION_ID);
 		PG_RETURN_VOID();
 	}
 
@@ -239,7 +247,6 @@ CreateReferenceTable(Oid relationId)
 	List *workerNodeList = NIL;
 	int replicationFactor = 0;
 	char *distributionColumnName = NULL;
-	bool requireEmpty = true;
 	char relationKind = 0;
 
 	EnsureCoordinator();
@@ -262,15 +269,19 @@ CreateReferenceTable(Oid relationId)
 	relationKind = get_rel_relkind(relationId);
 	if (relationKind == RELKIND_RELATION)
 	{
-		requireEmpty = false;
+		EnsureTableNotDistributed(relationId);
+	}
+	else
+	{
+		EnsureTableNotDistributed(relationId);
+		EnsureLocalTableEmpty(relationId);
 	}
 
 	colocationId = CreateReferenceTableColocationId();
 
 	/* first, convert the relation into distributed relation */
 	ConvertToDistributedTable(relationId, distributionColumnName,
-							  DISTRIBUTE_BY_NONE, REPLICATION_MODEL_2PC, colocationId,
-							  requireEmpty);
+							  DISTRIBUTE_BY_NONE, REPLICATION_MODEL_2PC, colocationId);
 
 	/* now, create the single shard replicated to all nodes */
 	CreateReferenceTableShard(relationId);
@@ -288,19 +299,17 @@ CreateReferenceTable(Oid relationId)
 /*
  * ConvertToDistributedTable converts the given regular PostgreSQL table into a
  * distributed table. First, it checks if the given table can be distributed,
- * then it creates related tuple in pg_dist_partition. If requireEmpty is true,
- * this function errors out when presented with a relation containing rows.
+ * then it creates related tuple in pg_dist_partition.
  *
  * XXX: We should perform more checks here to see if this table is fit for
  * partitioning. At a minimum, we should validate the following: (i) this node
  * runs as the master node, (ii) table does not make use of the inheritance
- * mechanism, (iii) table does not own columns that are sequences, and (iv)
- * table does not have collated columns.
+ * mechanism and (iii) table does not have collated columns.
  */
 static void
 ConvertToDistributedTable(Oid relationId, char *distributionColumnName,
 						  char distributionMethod, char replicationModel,
-						  uint32 colocationId, bool requireEmpty)
+						  uint32 colocationId)
 {
 	Relation relation = NULL;
 	TupleDesc relationDesc = NULL;
@@ -366,17 +375,6 @@ ConvertToDistributedTable(Oid relationId, char *distributionColumnName,
 								  "... AS IDENTITY.")));
 	}
 #endif
-
-	/* check that table is empty if that is required */
-	if (requireEmpty && !LocalTableEmpty(relationId))
-	{
-		ereport(ERROR, (errcode(ERRCODE_INVALID_TABLE_DEFINITION),
-						errmsg("cannot distribute relation \"%s\"",
-							   relationName),
-						errdetail("Relation \"%s\" contains data.",
-								  relationName),
-						errhint("Empty your table before distributing it.")));
-	}
 
 	/*
 	 * Distribution column returns NULL for reference tables,
@@ -618,7 +616,7 @@ CreateHashDistributedTable(Oid relationId, char *distributionColumnName,
 	uint32 colocationId = INVALID_COLOCATION_ID;
 	Oid sourceRelationId = InvalidOid;
 	Oid distributionColumnType = InvalidOid;
-	bool requireEmpty = true;
+	bool useExclusiveConnection = false;
 	char relationKind = 0;
 
 	/* get an access lock on the relation to prevent DROP TABLE and ALTER TABLE */
@@ -666,12 +664,26 @@ CreateHashDistributedTable(Oid relationId, char *distributionColumnName,
 	relationKind = get_rel_relkind(relationId);
 	if (relationKind == RELKIND_RELATION)
 	{
-		requireEmpty = false;
+		EnsureTableNotDistributed(relationId);
+		useExclusiveConnection = IsTransactionBlock() || !LocalTableEmpty(relationId);
+	}
+	else
+	{
+		EnsureTableNotDistributed(relationId);
+		EnsureLocalTableEmpty(relationId);
 	}
 
 	/* create distributed table metadata */
 	ConvertToDistributedTable(relationId, distributionColumnName, DISTRIBUTE_BY_HASH,
-							  ReplicationModel, colocationId, requireEmpty);
+							  ReplicationModel, colocationId);
+
+	/*
+	 * Ensure schema exists on each worker node. We can not run this function
+	 * transactionally, since we may create shards over separate sessions and
+	 * shard creation depends on the schema being present and visible from all
+	 * sessions.
+	 */
+	EnsureSchemaExistsOnAllNodes(relationId);
 
 	/* create shards */
 	if (sourceRelationId != InvalidOid)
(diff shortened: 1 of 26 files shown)


src/backend/distributed/master/master_create_shards.c
```diff
@@ -241,7 +271,7 @@ CreateShardsWithRoundRobinPolicy(Oid distributedTableId, int32 shardCount,
  * the source relation.
  */
 void
-CreateColocatedShards(Oid targetRelationId, Oid sourceRelationId)
+CreateColocatedShards(Oid targetRelationId, Oid sourceRelationId, bool emptyTable)
```
marcocitus: Same comment as above: I think the name `emptyTable` holds no particular relevance to this function. Maybe `useExclusiveConnections`.

src/test/regress/sql/multi_mx_metadata.sql
```diff
@@ -159,4 +159,4 @@ SELECT count(*) FROM pg_dist_transaction;
 -- Confirm that transactions were correctly rolled forward
 \c - - - :worker_1_port
 SELECT count(*) FROM pg_tables WHERE tablename = 'should_abort';
-SELECT count(*) FROM pg_tables WHERE tablename = 'should_commit';
+SELECT count(*) FROM pg_tables WHERE tablename = 'should_commit';
```
marcocitus: spurious change

src/test/regress/expected/multi_table_ddl.out
```diff
@@ -122,11 +122,10 @@ ALTER SEQUENCE testserialtable_id_seq OWNED BY testserialtable.id;
 -- verify sequence was created on worker
```
marcocitus: comment is stale

src/backend/distributed/master/master_stage_protocol.c
```diff
@@ -427,31 +433,58 @@ CreateShardPlacements(Oid relationId, int64 shardId, List *ddlEventList,
...
+ * connection if the command is run inside the transaction or the local table
+ * has a data on it. Function does NOT work transactionally for append
+ * distributed tables.
  */
-bool
+MultiConnection *
 WorkerCreateShard(Oid relationId, char *nodeName, uint32 nodePort,
 				  int shardIndex, uint64 shardId, char *newShardOwner,
-				  List *ddlCommandList, List *foreignConstraintCommandList)
+				  List *ddlCommandList, List *foreignConstraintCommandList,
```
marcocitus: This function is bulky with 11 parameters and contains a lot of caller-dependent logic. What if we just pass it a `MultiConnection *connection` and do the connection handling outside of this function?

src/backend/distributed/master/master_stage_protocol.c
```diff
@@ -466,12 +499,14 @@ WorkerCreateShard(Oid relationId, char *nodeName, uint32 nodePort,
...
 		}
 
-		queryResultList = ExecuteRemoteQuery(nodeName, nodePort, newShardOwner,
-											 applyDDLCommand);
-		if (queryResultList == NIL)
+		if (createInSeparateTransaction)
 		{
-			shardCreated = false;
-			break;
+			ExecuteRemoteQuery(nodeName, nodePort, newShardOwner,
```
marcocitus: We'd rather avoid calls to `ExecuteRemoteQuery`. If you open a new connection (using `FORCE_NEW_CONNECTION`) and don't call `RemoteTransactionBeginIfNecessary`, then the statements be executed outside of a transaction block if you just do `ExecuteCriticalRemoteCommand`.

src/backend/distributed/master/master_create_shards.c
```diff
@@ -409,11 +456,20 @@ CreateReferenceTableShard(Oid distributedTableId)
...
+	connectionList = CreateShardPlacements(distributedTableId, shardId, ddlCommandList,
+										   relationOwner, workerNodeList,
+										   workerStartIndex,
+										   replicationFactor, emptyTable,
+										   createInSeperateTransaction);
+
+	foreach(connectionCell, connectionList)
+	{
+		MultiConnection *connection = (MultiConnection *) lfirst(connectionCell);
+		UnclaimConnection(connection);
```
marcocitus: These calls to `UnclaimConnection` look out-of-place.

marcocitus: For readability, consider structuring the code as something like:
```
void
CreateShardsOnWorkers(List *newShardList, ..., bool useExclusiveConnections, bool createInSubtransaction)
{
    List *claimedConnectionList = NIL;
    int connectionFlags = 0;

    if (createInSubtransaction)
    {
        connectionFlags = FORCE_NEW_CONNECTION;
    }

    foreach(newShardCell, newShardList)
    {
        ...
        for (...)
        {
            ...
            connection = GetPlacementConnection(connectionFlags, ...);
            if (useExclusiveConnections)
            {
                /* use separate connection per placement to allow subsequent COPY */
                ClaimConnectionExlusively(connection);
                claimedConnectionsList = lappend(claimedConnectionsList, connection);
            }
            if (!createInSubtransaction)
            {
                RemoteTransactionBeginIfNecessary(connection);
            }
            ...
            WorkerCreateShardPlacement(connection, ...);
            ...
        }
    }

    foreach(connectionCell, claimedConnectionList)
    {
        ...
        UnclaimConnection(connection);
    }
}
```

That would make the overall procedure much easier to follow, since all the connection handling logic is in one place.

src/backend/distributed/master/master_stage_protocol.c
```diff
@@ -522,16 +556,18 @@ WorkerCreateShard(Oid relationId, char *nodeName, uint32 nodePort,
...
 						 referencedShardId, escapedReferencedSchemaName, escapedCommand);
 
-		queryResultList = ExecuteRemoteQuery(nodeName, nodePort, newShardOwner,
-											 applyForeignConstraintCommand);
-		if (queryResultList == NIL)
+		if (createInSeparateTransaction)
 		{
-			shardCreated = false;
-			break;
+			ExecuteRemoteQuery(nodeName, nodePort, newShardOwner,
```
marcocitus: We want to phase out `ExecuteRemoteQuery` since it relies on legacy APIs and opens a new connection for every command (!). If you send a command over a connection (e.g. using `ExecuteCriticalRemoteCommand`) without having sent `BEGIN` then it will commit immediately.

src/backend/distributed/commands/create_distributed_table.c
```diff
@@ -682,6 +699,45 @@ CreateHashDistributedTable(Oid relationId, char *distributionColumnName,
...
+		appendStringInfo(applySchemaCreationDDL, "%s", createSchemaDDL);
+
+		foreach(workerNodeCell, workerNodeList)
+		{
+			WorkerNode *workerNode = (WorkerNode *) lfirst(workerNodeCell);
+			char *nodeName = workerNode->workerName;
+			uint32 nodePort = workerNode->workerPort;
+			bool commandExecuted = false;
+
+			commandExecuted = ExecuteRemoteCommand(nodeName, nodePort,
```
marcocitus: I realise `ExecuteRemoteCommand` has a convenient signature, but it calls legacy `MultiClient*` functions which we'd like to phase out. Could maybe use `SendCommandListToWorkerInSingleTransaction` here or, since it's a single command, `GetNodeUserDatabaseConnection(FORCE_NEW_CONNECTION, ...)` + `ExecuteCriticalRemoteCommand`, or a new wrapper for that.

src/backend/distributed/commands/create_distributed_table.c
```diff
@@ -601,6 +609,7 @@ CreateHashDistributedTable(Oid relationId, char *distributionColumnName,
 	Oid sourceRelationId = InvalidOid;
 	Oid distributionColumnType = InvalidOid;
 	bool requireEmpty = true;
+	bool *emptyTable = palloc(sizeof(bool));
```
marcocitus: Just use `bool emptyTable = false`

src/backend/distributed/commands/create_distributed_table.c
```diff
@@ -653,7 +662,14 @@ CreateHashDistributedTable(Oid relationId, char *distributionColumnName,
 
 	/* create distributed table metadata */
 	ConvertToDistributedTable(relationId, distributionColumnName, DISTRIBUTE_BY_HASH,
-							  ReplicationModel, colocationId, requireEmpty);
+							  ReplicationModel, colocationId, requireEmpty, emptyTable);
```
marcocitus: Use `&emptyTable` to pass by pointer.

src/backend/distributed/commands/create_distributed_table.c
```diff
@@ -350,7 +357,8 @@ ConvertToDistributedTable(Oid relationId, char *distributionColumnName,
 	}
 
 	/* check that table is empty if that is required */
-	if (requireEmpty && !LocalTableEmpty(relationId))
+	*emptyTable = LocalTableEmpty(relationId);
```
marcocitus: Since we need this information later. Can we establish whether the table is empty prior to calling `ConvertToDistributedTable` and just pass the result as we do with `CreateShardsWithRoundRobinPolicy`?

src/backend/distributed/master/master_create_shards.c
```diff
@@ -88,7 +93,7 @@ master_create_worker_shards(PG_FUNCTION_ARGS)
  */
 void
 CreateShardsWithRoundRobinPolicy(Oid distributedTableId, int32 shardCount,
-								 int32 replicationFactor)
+								 int32 replicationFactor, bool emptyTable)
```
marcocitus: I think the name `emptyTable` holds no particular relevance to this function. Maybe `useExclusiveConnections`.

src/backend/distributed/master/master_stage_protocol.c
```diff
@@ -466,12 +499,14 @@ WorkerCreateShard(Oid relationId, char *nodeName, uint32 nodePort,
...
+							   applyDDLCommand);
+		}
+		else
+		{
+			ExecuteCriticalRemoteCommand(connection, applyDDLCommand->data);
```
marcocitus: I wonder whether we should keep the failover logic in `master_create_empty_shard`, since we have users that rely on that.

src/backend/distributed/commands/create_distributed_table.c
velioglu: In order to pass emptiness information to ConvertToDistributedTable, LocalTableEmpty(with IsDistributedTable and related relation opening functions) should be called just before each ConvertToDistributedTable. Should I create a new function for these operations, since calling them separately seems confusing to me.

marcocitus: You could remove the `requireEmpty` flag and introduce an `EnsureLocalTableEmpty()` function for callers that require the table to empty.

src/test/regress/expected/multi_create_shards.out
```diff
@@ -48,8 +48,10 @@ DETAIL:  Distributed relations must not specify the WITH (OIDS) option in their
 ALTER TABLE table_to_distribute SET WITHOUT OIDS;
 -- use an index instead of table name
 SELECT master_create_distributed_table('table_to_distribute_pkey', 'id', 'hash');
-ERROR:  cannot distribute relation: table_to_distribute_pkey
-DETAIL:  Distributed relations must be regular or foreign tables.
+ERROR:  "table_to_distribute_pkey" is an index
+LINE 1: SELECT EXISTS (SELECT 1 FROM public.table_to_distribute_pkey...
+                                     ^
+QUERY:  SELECT EXISTS (SELECT 1 FROM public.table_to_distribute_pkey)
```
marcocitus: I suppose it might be slightly nicer if we do something like a `EnsureTableId(relationId);` first.

src/test/regress/expected/multi_create_table.out
```diff
@@ -518,3 +512,97 @@ SELECT create_distributed_table('orders_hash_part', 'o_orderkey');
...
+
+ROLLBACK;
+-- Table should not exist on the worker node
+\c - - - :worker_1_port
+\d rollback_table*
+\c - - - :master_port
+-- Again there should be no table on the worker node
+BEGIN;
+SELECT create_distributed_table('rollback_table','id');
+ERROR:  relation "rollback_table" does not exist
```
marcocitus: Did you mean to do another `CREATE TABLE rollback_table(id int, name varchar(20));`?

src/backend/distributed/master/master_stage_protocol.c
```diff
@@ -362,96 +364,146 @@ CheckDistributedTable(Oid relationId)
...
+				ClaimConnectionExclusively(connection);
+				claimedConnectionList = lappend(claimedConnectionList, connection);
+			}
+
+			RemoteTransactionBeginIfNecessary(connection);
+			MarkRemoteTransactionCritical(connection);
 		}
+		else
+		{
+			connection = GetPlacementConnection(FORCE_NEW_CONNECTION, shardPlacement,
```
marcocitus: Something that concerns me about this change is that it changes the failure-behaviour of `master_create_empty_shard`. Older Citus users have append-distributed tables with replication factor >1 and expect that they can take down a worker node without repercussions. All operations on append-distributed tables (`SELECT`, `master_create_empty_shard`, `master_append_table_to_shard`, `COPY`) are robust to node failures.

It might make sense to handle  `master_create_empty_shard` through a separate function that's similar to the original `CreateShardPlacements`. In that function, you could use `GetNodeUserDatabaseConnection` to create the placement and if the connection fails, we try the next worker node. Once the connection is established we can insert the shard placement and call `WorkerCreateShard`.

marcocitus:
It would be good to add some more regression tests with commands immediately after the `create_distributed_table` in the same transaction, including `DROP TABLE`, `SELECT * FROM table`, `INSERT INTO table ...`, both with commit and rollback (and checking result).

src/backend/distributed/master/master_stage_protocol.c
velioglu: Then, the operation going to error out if the coordinator can connect to the worker but couldn't run the shard creation commands on it successfully, right ?

marcocitus: Correct. There's probably not as much value in failing over if a SQL command errors out, since that only happens if there is something wrong with the database (rare, and probably not something we want to hide) or the connection fails immediately after establishing it (very rare). If not doing it keeps the code simpler then that seems preferable.

src/backend/distributed/master/master_stage_protocol.c
```diff
@@ -431,27 +423,129 @@ CreateShardPlacements(Oid relationId, int64 shardId, List *ddlEventList,
 
 
 /*
+ * InsertShardPlacementRows inserts shard placements to the metadata table on
+ * the coordinator node. Then, returns the list of added shard placements.
+ */
+List *
+InsertShardPlacementRows(Oid relationId, int64 shardId, List *workerNodeList,
```
velioglu: I think functions of master_create_shards.c consists of higher level callers, so remaining that function in master_stage_protocol.c is better idea.

marcocitus:
We can continue with this PR now that #1455 is merged.

onderkalaci:
@velioglu This should be unblocked now

src/test/regress/expected/multi_create_table.out
```diff
@@ -552,3 +549,180 @@ SELECT relpersistence FROM pg_class WHERE relname LIKE 'unlogged_table_%';
...
+\d tt1_360495
+  Table "public.tt1_360495"
+ Column |  Type   | Modifiers 
+--------+---------+-----------
+ id     | integer | 
+
+\c - - - :master_port
+DROP TABLE tt1;
+-- It is known that drop table is not allowed in the same transaction with
+-- create_distributed_table
```
marcocitus: Comment is no longer accurate

src/test/regress/expected/multi_create_table.out
```diff
@@ -360,16 +360,13 @@ INSERT INTO data_load_test VALUES (243, 'world');
...
 HINT:  Empty your table before distributing it.
 SELECT create_distributed_table('data_load_test', 'col1', 'range');
 ERROR:  cannot distribute relation "data_load_test"
-DETAIL:  Relation "data_load_test" contains data.
 HINT:  Empty your table before distributing it.
 -- table must be empty when using master_create_distributed_table (no shards created)
 SELECT master_create_distributed_table('data_load_test', 'col1', 'hash');
 ERROR:  cannot distribute relation "data_load_test"
-DETAIL:  Relation "data_load_test" contains data.
 HINT:  Empty your table before distributing it.
```
marcocitus: Might be nice to keep this detail or add it to the error message.

marcocitus:
The regression tests currently fail on PG10 because there is a `\d`, which gives different output on PG10. We should no longer use `\d` in tests, instead `multi_test_helpers.sql` defines some views we can use.

src/test/regress/expected/multi_create_table.out
```diff
@@ -650,8 +653,7 @@ COMMIT;
 \c - - - :master_port
 DROP TABLE tt1;
 DROP TABLE tt2;
--- It is known that creating a table with master_create_empty_shard is not
--- transactional operation.
+-- It is known that creating a table with master_create_empty_shard is transactional
```
marcocitus: Is it? I thought we only made `create_distributed_table` transactional. `master_create_empty_shard` has a bunch of legacy applications, so it might be better to keep it non-transactional.

src/test/regress/sql/multi_create_table.sql
```diff
@@ -403,15 +402,38 @@ COMMIT;
...
+-- Tests with create_distributed_table & DDL & DML commands
+
+-- Test should pass since GetPlacementListConnection can provide connections
+-- in this order of execution
+CREATE TABLE sample_table(id int);
+SELECT create_distributed_table('sample_table','id');
+
+BEGIN;
+CREATE TABLE stage_table (LIKE sample_table);
+\COPY stage_table FROM stdin; -- Note that this operation is a local copy
```
marcocitus: can we do it as a distributed copy as well?

velioglu:
@marcocitus new tests are added and \d is omitted.

src/test/regress/sql/multi_create_table.sql
```diff
@@ -293,3 +292,176 @@ SELECT * FROM master_get_table_ddl_events('unlogged_table');
...
+\copy rollback_table from stdin delimiter ','
+1, 'name_1'
+2, 'name_2'
+3, 'name_3'
+\.
+ROLLBACK;
+
+-- Table should not exist on the worker node
+\c - - - :worker_1_port
+SELECT "Column", "Type", "Modifiers" FROM table_desc WHERE relid = (SELECT oid FROM pg_class WHERE relname LIKE 'public.rollback_table%');
```
marcocitus: the relname in pg_class does not include the schema

src/test/regress/sql/multi_create_table.sql
```diff
@@ -293,3 +292,176 @@ SELECT * FROM master_get_table_ddl_events('unlogged_table');
...
+\c - - - :master_port
+
+BEGIN;
+CREATE TABLE rollback_table(id int, name varchar(20));
+SELECT create_distributed_table('rollback_table','id');
+\copy rollback_table from stdin delimiter ','
```
marcocitus: Might be worth throwing in a create index afterwards.

Issue #1437, opened 2017-05-31
Foreign keys lead to deadlocks while rebalancing


anarazel:
When rebalancing a table with foreign keys we'll frequently get into unprincipled deadlocks, i.e. deadlocks that aren't due to user logic, but due to our code.

Steps to reproduce:

schema:
```
DROP TABLE IF EXISTS test_table CASCADE;
CREATE TABLE test_table (test_id integer primary key, data text, data2 text);
SELECT create_distributed_table('test_table', 'test_id', 'hash');

DROP TABLE IF EXISTS test_table2 CASCADE;
CREATE TABLE test_table2 (test_id integer NOT NULL references test_table , data text, data2 text);
SELECT create_distributed_table('test_table2', 'test_id', 'hash');
postgres[32630][1]=# SELECT get_shard_id_for_distribution_column('test_table', 8);
┌──────────────────────────────────────┐
│ get_shard_id_for_distribution_column │
├──────────────────────────────────────┤
│                               102144 │
└──────────────────────────────────────┘
(1 row)

postgres[28742][1]=# SELECT * FROM pg_dist_shard WHERE shardid = 102144;
┌──────────────┬─────────┬──────────────┬───────────────┬───────────────┐
│ logicalrelid │ shardid │ shardstorage │ shardminvalue │ shardmaxvalue │
├──────────────┼─────────┼──────────────┼───────────────┼───────────────┤
│ test_table   │  102144 │ t            │ -2147483648   │ -2013265921   │
└──────────────┴─────────┴──────────────┴───────────────┴───────────────┘
(1 row)
```

To reproduce the issue easily I currently use gdb. In one session I set a breakpoint like:
```
(gdb) b LockAcquire if lockmode == RowExclusiveLock
Breakpoint 1 at 0x7ca30a: file /home/andres/src/postgresql-9.5/src/backend/storage/lmgr/lock.c, line 672.
```
and then run 
```
INSERT INTO test_table VALUES(8, 1, 2);
```
once that breakpoint hits, I let it acquire just that lock, by making the function return:
```
Breakpoint 1, LockAcquire (locktag=0x7ffee5ce7100, lockmode=3, sessionLock=0 '\000', dontWait=0 '\000')
    at /home/andres/src/postgresql-9.5/src/backend/storage/lmgr/lock.c:672
672             return LockAcquireExtended(locktag, lockmode, sessionLock, dontWait, true);
(gdb) fin
[...]

PR #1438, opened 2017-05-31
Don't take a table lock in ForeignConstraintGetReferencedTableId

marcocitus:
We should add some isolation tests, but those can only be added in enterprise.

See #1437


anarazel:
Merging and backporting.

PR #1439, opened 2017-06-01
Rewrite psql-dependent tests for PostgreSQL 10

jasonmp85:
Given the large number of tests which depend on `\d`'s output (which adds two columns in PostgreSQL 10), it didn't make sense to add test variants for every test using `\d`.

Instead, I wrote a couple of views which largely mimic `\d`'s output for tables. That leaves the use of `\d` for indexes, constraints, and foreign keys.

The `\d` output doesn't change for indexes, so I left those as-is when possible. Other places, tests don't have any real need to _care_ about the actual definition of an index or table, just i.e. "whether a shard was created" or "whether an index was dropped". In those "existence" tests, it was sufficient to just use `\dt` or `\di`, which basically list objects without showing their definitions.

Finally, certain tests verified constraint and foreign key definitions. I wrote a view that prints names and definitions of `CHECK` constraints or `FOREIGN KEY`s for a provided `relid`, which covers those uses.

Additionally, three of four places had entirely wrong assertions for tests; they would have a comment like "verify the shard has not been dropped", then an `\d` invocation that returned no results. Many of these were just incorrect tests that assumed a particular shard identifier that wasn't actually in use for that distributed table. For these I corrected the test.

The `citusdata/tools` repository still has the `pg10_hack` which hardcodes use of a pre-`10` `psql`; this needs to be removed, though I also need to update the Pull Request build to be capable of building the `10beta1` PostgreSQL source tree (for the new assert-enabled and isolationtest builds). I'll follow up on that after standup.


src/test/regress/sql/multi_extension.sql
```diff
@@ -175,3 +175,131 @@ ALTER EXTENSION citus UPDATE;
...
 \d
+
+\c - - - :master_port
+
+CREATE VIEW table_fkeys AS
+SELECT r.conname AS "Constraint",
+       pg_catalog.pg_get_constraintdef(r.oid, true) AS "Definition",
+       conrelid AS "relid"
+FROM pg_catalog.pg_constraint r
+WHERE r.contype = 'f';
```
anarazel: I'm fine with this, but it might be worthwhile to look at information_schema.referential_constraints, since that's guaranteed to stay compatible.

src/test/regress/sql/multi_extension.sql
```diff
@@ -175,3 +175,131 @@ ALTER EXTENSION citus UPDATE;
...
+SELECT a.attname AS "name",
+    pg_catalog.format_type(a.atttypid, a.atttypmod) AS "type",
+    (SELECT substring(pg_catalog.pg_get_expr(d.adbin, d.adrelid) for 128)
+     FROM pg_catalog.pg_attrdef d
+     WHERE d.adrelid = a.attrelid AND d.adnum = a.attnum AND a.atthasdef) AS "default",
+    a.attnotnull AS "notnull",
+    a.attrelid AS "relid"
+FROM pg_catalog.pg_attribute a
+WHERE a.attnum > 0 AND NOT a.attisdropped
+ORDER BY a.attnum;
```
anarazel: It'd be better to refer to information_schema.columns here.  Fine with having table_attrs as a view over that, but that way we won't have to update it several times over the next releases.

src/test/regress/sql/multi_extension.sql
```diff
@@ -175,3 +175,131 @@ ALTER EXTENSION citus UPDATE;
...
+       )) AS "Modifiers",
+	   "relid"
+FROM table_attrs;
+
+CREATE VIEW table_checks AS
+SELECT r.conname AS "Constraint",
+       pg_catalog.pg_get_constraintdef(r.oid, true) AS "Definition",
+       conrelid AS "relid"
+FROM pg_catalog.pg_constraint r
+WHERE r.contype = 'c';
```
anarazel: Repeat the same information_schema complaint for all of the rest of the file.

Additionally I think it'd be better if we were to hoist run_command_on_master_and_workers() into some early file and used that here instead of repeating the definition.

src/test/regress/sql/multi_extension.sql
jasonmp85: FWIW I based all these off of the commands that `psql` prints when using `\d` commands with `ECHO_HIDDEN` set to on. Do those change that frequently?

anarazel:
On 2017-06-05 14:43:25 -0700, Jason Petersen wrote:
> FWIW I based all these off of the commands that `psql` prints when using `\d` commands with `ECHO_HIDDEN` set to on. Do those change that frequently?

Yes.  Look at the number of ifdefs in describe.c.

src/test/regress/sql/multi_extension.sql
anarazel: Yes, it does.  See the number of if's in psql's describe.c.

jasonmp85:
> Yes.  Look at the number of ifdefs in describe.c.

Contra that, the output of my queries is identical in `9.5`, `9.6`, and `10` and so if it changes in `11` we can always just change the view definition. Since this works in all three versions now, do we really need to block on future-proofing it against hypothetical changes? We're already in a better place given that all tests now use this view rather than `\d`. We can update the view later when things break and hopefully not have to touch any tests?

src/test/regress/sql/multi_extension.sql
jasonmp85: > it'd be better if we were to hoist `run_command_on_master_and_workers`

Sure, I can do that.

anarazel:
On 2017-06-05 15:05:35 -0700, Jason Petersen wrote:
> Contra that, the output of my queries is identical in `9.5`, `9.6`,
> and `10` and so if it changes in `11` we can always just change the
> view definition. Since this works in all three versions now, do we
> really need to block on future-proofing it against hypothetical
> changes? We're already in a better place given that all tests now use
> this view rather than `\d`. We can update the view later when things
> break and hopefully not have to touch any tests?

I'd still do that, while you have the context.  Will be harder for
somebody later.

src/test/regress/sql/multi_extension.sql
jasonmp85: So it turns out that mimicking this output using `information_schema.referential_constraints` is fairly hairy. Here's what I've got going right now, and it's still got several things wrong with it and a bug when two constraints share the same name:

```sql
CREATE VIEW table_fkey_cols AS
SELECT rc.constraint_name AS "name",
       kcu.column_name AS "column_name",
       uc_kcu.column_name AS "refd_column_name",
       format('%I.%I', kcu.table_schema, kcu.table_name)::regclass::oid AS relid,
       format('%I.%I', uc_kcu.table_schema, uc_kcu.table_name)::regclass::oid AS refd_relid
FROM information_schema.referential_constraints rc,
     information_schema.key_column_usage kcu,
     information_schema.key_column_usage uc_kcu
WHERE rc.constraint_schema = kcu.constraint_schema AND
      rc.constraint_name = kcu.constraint_name AND
      rc.unique_constraint_schema = uc_kcu.constraint_schema AND
      rc.unique_constraint_name = uc_kcu.constraint_name;

CREATE VIEW table_fkeys AS
SELECT name AS "Constraint",
       format('FOREIGN KEY (%s) REFERENCES %s(%s)',
              string_agg(DISTINCT quote_ident(column_name), ', '),
              string_agg(DISTINCT refd_relid::regclass::text, ', '),
              string_agg(DISTINCT quote_ident(refd_column_name), ', ')) AS "Definition",
       "relid"
FROM table_fkey_cols
GROUP BY (name, relid);
```

jasonmp85: Actually, the problem I'm having is I can't find the referenced column in a foreign key created against a column which already has a unique index. The following isn't 100% complete (it's missing schema declarations), but it's enough to demonstrate:

```sql
CREATE TABLE mx_test_schema_1.mx_table_1 (col1 int UNIQUE, col2 text);
CREATE TABLE mx_test_schema_2.mx_table_2 (col1 int, col2 text);
CREATE UNIQUE INDEX mx_index_4 ON mx_test_schema_2.mx_table_2(col1);

ALTER TABLE 
	mx_test_schema_1.mx_table_1 
ADD CONSTRAINT 
	mx_fk_constraint 
FOREIGN KEY 
	(col1)
REFERENCES
	mx_test_schema_2.mx_table_2(col1);

SELECT * FROM information_schema.referential_constraints;
--  constraint_catalog | constraint_schema |     constraint_name      | unique_constraint_catalog | unique_constraint_schema |   unique_constraint_name    | match_option | update_rule | delete_rule 
-- --------------------+-------------------+--------------------------+---------------------------+--------------------------+-----------------------------+--------------+-------------+-------------
--  regression         | mx_test_schema_1  | mx_fk_constraint         |                           |                          |                             | NONE         | NO ACTION   | NO ACTION

```

jasonmp85: OK, changing a previous `UNIQUE INDEX` to a `UNIQUE CONSTRAINT` makes my `VIEW` work again.

Still, since the initial implementation only read the names and OIDs of the check and foreign-key constraints, I think the risk of using the `pg_catalog` tables is low.

jasonmp85: (FWIW we're still left without a way to verify that the `NOT VALID` flag is propagated to worker foreign constraints).

jasonmp85:
@anarazel I think this addresses all comments. If you can give this a :shipit:, I'll merge this into the PostgreSQL 10 branch for easier management when I do this rebase in a sec.

src/test/regress/expected/multi_test_helpers.out
```diff
@@ -0,0 +1,86 @@
+-- File to create functions and helpers needed for subsequent tests
+-- create a helper function to create objects on each node
+CREATE FUNCTION run_command_on_master_and_workers(p_sql text)
+RETURNS void LANGUAGE plpgsql AS $$
+BEGIN
+     EXECUTE p_sql;
+     PERFORM run_command_on_workers(p_sql);
+END;$$;
```
anarazel: @samay-sharma @saicitus I think it'd be a good idea to just add a function like this to Citus. Do you agree?

jasonmp85: I'm a little worried it could make bad things easy.

samay-sharma: @anarazel : Yeah, I agree. It hasn't been a huge pain to ask them to do this in two steps, but it would definitely be an add-on.

anarazel: > I'm a little worried it could make bad things easy.

I'm not sure how it'd make things worse than they're today?

PR #1441, opened 2017-06-06
Allow COPY after a multi-shard command

marcocitus:
Minor enhancement that came up in #1402.

This removes the `XactModificationLevel` check at the start of COPY that was made redundant by consistently using `GetPlacementConnection`.


src/test/regress/sql/multi_modifying_xacts.sql
```diff
@@ -162,6 +162,16 @@ INSERT INTO labs VALUES (6, 'Bell Labs');
 \.
 COMMIT;
 
+-- COPY cannot be performed if multiple shards were modified over the same connection
```
marcocitus: Added

Issue #1394, opened 2017-05-06
Deadlock Handling (Parent Issue)


anarazel:
This is issue is just the high-level overview issue.  Individual sub-tasks have/will get their own tickets.  Please edit the top-level description to integrate agreed upon changes (and just make plain correction/refinements), so the overall design is visible in one-place.

A number of issues like:
- #1131 - UPDATE on multiple shards natively works 
- #1130 - DELETE on multiple shards natively works 
- #455 - Add support for multi-row INSERT statements
- #1328 Expand transaction coverage for queries that touch reference and distributed tables 
- #please-expand

are essentially blocked because we currently have neither detection nor handling of distributed deadlocks.  So far that's mostly handled by prohibiting commands / combinations of commands that could trigger deadlocks (causing usability issues) or by making them acquire more heavyweight locks than strictly necessary (causing concurrency / practicability issues).  There's several ways that such deadlocks can still be triggered, however.

[ INSERT DISTRIBUTED DEADLOCK REFERENCE/EXPLANATION?]

We're planning to implement two approaches to handle deadlocks:
1. Detect distributed deadlocks, and abort them.  This requires communication between nodes, and will be primarily important for deadlocks triggered by in-fortune sequencing of actions by the user (e.g. separate UPDATE statements within a transaction, that are not well ordered).  This'll trigger after a configurable timeout, and will usually detect deadlocks that already lasted several seconds.
1. Implement Wound/Wait for commands that opt into it.  Wound-wait, very much simplified, will cancel a "younger" transaction if an "older" requires a resource.  That allows for very fast (heuristic) deadlock detection, guarantees global progress, and is fair. Particularly useful for multi-row commands that can reasonably expected to occasionally experience deadlocks.  Depending on the command and the state in which it is sent (first in xact), the canceled command can transparently be retried.  By retaining the original transaction id, there's a guarantee of eventual progress in cases where retries are possible.

Subtasks of this are:

1. Make everything in Citus cancelable, otherwise we'll not be able to handle deadlocks by aborting (#1100, #1121)
1. Code that sends multiple commands over various connections in parallel, needs to be properly non-blocking (rather than as currently often done, be implemented by asynchronously sending commands to all, then synchronously waiting for results 1-by-1).  Otherwise additional and unnecessary deadlocks can be introduced.  There's also a good chunk in performance benefit in doing so.
[...]

PR #1447, opened 2017-06-12
Rough Prototype: Distributed Deadlock Detection

anarazel:
This is a very basic implementation of distributed deadlock detection as sketched out in #1394.

In my very limited tests it seems to be working reasonably well, so I think the concept has at least some merit.  There's obviously still a lot left to do, this is definitely not even close to being mergeable.  Most commit message have at least some detail.

As an example:
```
Setup:
CREATE TABLE test_table (test_id integer primary key, data text, data2 text);
SELECT create_distributed_table('test_table', 'test_id', 'hash');
INSERT INTO test_table VALUES (1);
INSERT INTO test_table VALUES (5);
postgres[18016][1]=# SELECT get_shard_id_for_distribution_column('test_table', 1);
┌──────────────────────────────────────┐
│ get_shard_id_for_distribution_column │
├──────────────────────────────────────┤
│                               102041 │
└──────────────────────────────────────┘
(1 row)

Time: 0.931 ms
postgres[18016][1]=# SELECT get_shard_id_for_distribution_column('test_table', 2);
┌──────────────────────────────────────┐
│ get_shard_id_for_distribution_column │
├──────────────────────────────────────┤
│                               102064 │
└──────────────────────────────────────┘
(1 row)

S1:
BEGIN;
UPDATE test_table SET data = '' WHERE test_id = 1; 

S2:
BEGIN;
UPDATE test_table SET data = '' WHERE test_id = 5;

S1:
UPDATE test_table SET data = '' WHERE test_id = 5;

S2:
UPDATE test_table SET data = '' WHERE test_id = 1;
ERROR:  57014: canceling statement due to user request
LOCATION:  ProcessInterrupts, postgres.c:2988
```

### DIFF
--- a/src/backend/distributed/Makefile
+++ b/src/backend/distributed/Makefile
@@ -11,7 +11,7 @@ EXTVERSIONS = 5.0 5.0-1 5.0-2  \
 	6.0-1 6.0-2 6.0-3 6.0-4 6.0-5 6.0-6 6.0-7 6.0-8 6.0-9 6.0-10 6.0-11 6.0-12 6.0-13 6.0-14 6.0-15 6.0-16 6.0-17 6.0-18 \
 	6.1-1 6.1-2 6.1-3 6.1-4 6.1-5 6.1-6 6.1-7 6.1-8 6.1-9 6.1-10 6.1-11 6.1-12 6.1-13 6.1-14 6.1-15 6.1-16 6.1-17 \
 	6.2-1 6.2-2 6.2-3 6.2-4 \
-	7.0-1 7.0-2
+	7.0-1 7.0-2 7.0-3
 
 # All citus--*.sql files in the source directory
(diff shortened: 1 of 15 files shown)


anarazel:
@marcocitus @onderkalaci I've an approach that works to override the default error message (basically putting information about the error in shared memory, and overwriting it in the log emission handler. Not pretty, but works). For the shared memory I think it'd be best to integrate that into the shared memory segment that we use to store the distributed transaction id - it seems pointless to have several such arrays.  Does that make sense to you two?  If so I think it'd make sense to delay working on that item until the distributed transaction id infrastructure is finished (cc: @sumedhpathak).

anarazel:
@onderkalaci There seems little of value left in this PR, should we close it and delete the branch?

onderkalaci:
> @onderkalaci There seems little of value left in this PR, should we close it and delete the branch?

I'm closing the PR, but will keep the branch for one more week 😄

PR #1448, opened 2017-06-13
Use placement connection to drop shards instead of node connection

velioglu:
In order to call drop shards with create_distributed_table in the same transaction, connect to nodes with placement connection rather than node connection.

### DIFF
--- a/src/test/regress/expected/multi_transactional_drop_shards.out
+++ b/src/test/regress/expected/multi_transactional_drop_shards.out
@@ -197,9 +197,6 @@ SELECT master_create_empty_shard('transactional_drop_shards');
 BEGIN;
 INSERT INTO transactional_drop_shards VALUES (1);
 DROP TABLE transactional_drop_shards;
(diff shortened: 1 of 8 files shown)


src/test/regress/expected/multi_transactional_drop_shards.out
```diff
@@ -242,7 +239,11 @@ Table "public.transactional_drop_shards_1410005"
 BEGIN;
 INSERT INTO transactional_drop_shards VALUES (1);
 SELECT master_apply_delete_command('DELETE FROM transactional_drop_shards');
-ERROR:  shard drop operations must not appear in transaction blocks containing other distributed modifications
+ master_apply_delete_command 
+-----------------------------
+                           1
+(1 row)
+
```
marcocitus: Could you also add a test for DDL + DROP table and maybe COPY + DROP table?

marcocitus: Maybe also create_distributed_table+DROP table since that's mentioned in the PR description :)

PR #1452, opened 2017-06-14
Add citus_create_restore_point for taking distributed snapshots

### LINKED ISSUE #1451: Add a function for creating a consistent, distributed restore point
If we recover a Citus database by recovering from an independent set of restore points or WAL timestamps, then the recovered database may not be consistent. In the worst case, DDL commands may have only been applied on some nodes, in which case the distributed table in the recovered database is no longer functional. More commonly, reference table writes may not have been applied to all replicas, or data from COPY could be partially missing.

Commit order may also differ in the recovered database, which may violate causal consistency. If the application performs write B only after committing write A, it can normally derive the fact that write A succeeded from the fact that write B succeeded, but in the recovered database this may not be the case if write A and B are on different nodes.

For the purpose of making periodic, consistent backups or forking a cluster, we should add a function (e.g. `citus_create_restore_point`) that creates a consistent restore point across all the nodes for consistent PITR/forking.

marcocitus:
This change adds a `citus_create_restore_point` function to Citus, which aims to create a consistent snapshot of the distributed database by briefly blocking writes to distributed tables and then calling `pg_create_restore_point` with the same restore point name on all nodes.

As with `pg_create_restore_point`, every restore point should have a unique name that consists of up to 64 characters:
https://www.postgresql.org/docs/current/static/functions-admin.html

The implementation tries to minimise the time writes are blocked by establishing connections to nodes upfront. It also sends a BEGIN to ensure that we actually have a connection to a remote postgres backend, and not to an intermediate pgbouncer in transaction-pooling mode that could introduce further delays. It doesn't actually send commit to workers after the `pg_create_restore_point`, since it's not required for `pg_create_restore_point` to work and would add another round-trip.

While the run-time is typically short, `citus_create_restore_point` still needs to wait for all pending writes to distributed tables to finish. In the mean time, writes to distributed tables that come behind it will block. Before running `citus_create_restore_point`, it is a good idea to first do `SET lock_timeout TO '1s'` in case there are long-running write transactions which would cause `citus_create_restore_point` to block all writes for a long time. I did some testing with single-statement DML in pgbench, and `citus_create_restore_point` took 20-30ms on a heavily loaded system, of which the first 5-10ms is time spent on connection establishment (not blocking).

Limitations: `citus_create_restore_point` does not guarantee that external changes (including those made by `run_command_on_*`) or changes made through workers in MX are consistent in the snapshot. It is still useful for MX in the sense that it prevents partially applied DDL or partially applied reference table writes. For MX, it's recommended to run `start_metadata_sync_to_node` for every node after recovering an MX cluster.

Fixes #1451

### DIFF
--- a/src/backend/distributed/master/citus_create_restore_point.c
+++ b/src/backend/distributed/master/citus_create_restore_point.c
@@ -1,0 +1,205 @@
+/*-------------------------------------------------------------------------
+ *
+ * citus_create_restore_point.c
+ *
+ * UDF for creating a consistent restore point across all nodes.
+ *
+ * Copyright (c) 2017, Citus Data, Inc.
+ *
+ *-------------------------------------------------------------------------
+ */
+
+#include "postgres.h"
+#include "libpq-fe.h"
+
+#include "access/xlog.h"
+#include "access/xlog_internal.h"
+#include "catalog/pg_type.h"
+#include "distributed/connection_management.h"
+#include "distributed/master_metadata_utility.h"
+#include "distributed/metadata_cache.h"
+#include "distributed/remote_commands.h"
+#include "nodes/pg_list.h"
+#include "storage/lmgr.h"
+#include "storage/lock.h"
+#include "utils/builtins.h"
+#include "utils/pg_lsn.h"
+
+
+#define CREATE_RESTORE_POINT_COMMAND "SELECT pg_catalog.pg_create_restore_point($1::text)"
+
+
+/* local functions forward declarations */
+static List * OpenConnectionsToAllNodes(void);
+static void BlockAllDistributedWrites(void);
+static void CreateRemoteRestorePoints(char *restoreName, List *connectionList);
+
+
+/* exports for SQL callable functions */
+PG_FUNCTION_INFO_V1(citus_create_restore_point);
+
+
+/*
+ * citus_create_restore_point blocks writes to distributed tables and then
+ * runs pg_create_restore_point on all nodes. This creates a consistent
+ * restore point under the assumption that there are no other writers
+ * than the coordinator.
+ */
+Datum
+citus_create_restore_point(PG_FUNCTION_ARGS)
+{
+	text *restoreNameText = PG_GETARG_TEXT_P(0);
+	char *restoreNameString = NULL;
+	XLogRecPtr localRestorePoint = InvalidXLogRecPtr;
+	List *connectionList = NIL;
+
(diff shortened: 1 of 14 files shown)


src/backend/distributed/master/citus_create_restore_point.c
```diff
@@ -0,0 +1,199 @@
...
+
+/*
+ * BlockAllDistributedWrites blocks all modifications to distributed tables
+ * by taking ShareRowExclusive locks on all distributed tables.
+ */
+static void
+BlockAllDistributedWrites(void)
+{
+	ListCell *distributedTableCell = NULL;
+	List *distributedTableList = DistributedTableList();
```
aamederen: Would it be required to sort this list, before taking locks on them, to avoid deadlocks?

src/backend/distributed/master/citus_create_restore_point.c
```diff
@@ -0,0 +1,199 @@
...
+	/* DANGER: finish as quickly as possible after this */
+	BlockAllDistributedWrites();
+
+	/* do local restore point first to bail out early if something goes wrong */
+	localRestorePoint = XLogRestorePoint(restoreNameString);
+
+	/* run pg_create_restore_point on all nodes */
+	CreateRemoteRestorePoints(restoreNameString, connectionList);
+
+	PG_RETURN_LSN(localRestorePoint);
```
marcocitus: Just for consistency with `pg_create_restore_point`.

aamederen:
Did some tests and it looked good on Cloud

src/backend/distributed/master/citus_create_restore_point.c
marcocitus: It doesn't really help unless all transactions always modify tables in the same order.

src/backend/distributed/master/citus_create_restore_point.c
```diff
@@ -0,0 +1,205 @@
...
+#include "storage/lmgr.h"
+#include "storage/lock.h"
+#include "utils/builtins.h"
+#include "utils/pg_lsn.h"
+
+
+#define CREATE_RESTORE_POINT_COMMAND "SELECT pg_catalog.pg_create_restore_point($1::text)"
+
+
+/* Local functions forward declarations */
```
aamederen: Should be all lowercase

aamederen:
Apart from the minor lowercase issue, this PR looks fine :shipit:

PR #1453, opened 2017-06-14
Shard placements don't hardcode nodename and nodeport

### LINKED ISSUE #1449: pg_dist_shard_placement changes for 7.0
This issue supersedes #1414, which contains the motivation and some discussion.

Time estimate:
- 3 days

Tasks:
- [x] Rename `pg_dist_shard_placement_placementid_seq` to `pg_dist_placement_placementid_seq`
- [x] Create a `pg_dist_placement` table.
    ```sql
    CREATE TABLE pg_dist_placement(
      placementid bigint not null default nextval('pg_dist_placement...'),
      shardid bigint not null,
      shardstate integer not null,
      shardlength bigint not null,
      groupid integer not null
    )
    ```
- [x] Create a function to lookup the group for a (nodename, nodeport) pair
- [x] Fill `pg_dist_placement` with the previous contents of `pg_dist_shard_placement`.
- [x] Drop `pg_dist_shard_placement` and replace it with a view (which joins against pg_dist_node)
    ```sql
    CREATE VIEW pg_dist_shard_placement AS
    SELECT shardid, shardstate, shardlength, placementid, nodename, nodeport
    FROM pg_dist_placement placement INNER JOIN pg_dist_node node ON (
        placement.groupid = node.groupid AND node.noderole = 'primary'
    );
    ```
    (This will be slightly different depending on whether `noderole` is added before or after this)
- [x] Update `metadata_sync.c` to sync `pg_dist_placement` rather than `pg_dist_shard_placement`.
- [x] `UpdateShardPlacementState` is a self-contained refactor.
- [x] `TupleToShardPlacement` should be updated to use `pg_dist_shard`.
- [x] `ShardPlacement` should be updated to match `pg_dist_placement` (`groupid` in lieu of `node{name,port}`)
  - If this doesn't happen then the future callers of `DeleteShardPlacementRow` don't have a `groupid` to pass it.
- [x] `InsertShardPlacementRow` and callers need to be updated.
- [x] `DeleteShardPlacementRow` and callers need to use `groupid` rather than `nodename`/`nodeport`.
- [x] `NodeHasActiveShardPlacements` needs to read from `pg_dist_node` and `pg_dist_placement`. Ideally it could read from the cached metadata?
- [x] `BuildShardPlacementList` should read from `pg_dist_placement`. Unclear whether it should return raw placements or directly do the joining with `pg_dist_node`.
- [x] `BuildCachedShardList` needs a refactor. We can probably get away with keeping the same data in the metadata cache, but when populating it we need to join against the `WorkerNodeHash` and find the primary node. Possibly we only store the primary node when building the `WorkerNodeHash`?
- [ ] Make sure invalidation still works, that we invalidate the cached shard placement when either the `pg_dist_node` or `pg_dist_placement` row is invalidated.
- [x] More changes to `metadata_cache.c`, functions like `DistShardPlacementRelationId` should use from the new table.
- [ ] Update `DeleteAllReferenceTablePlacementsFromNode` and all comments in `reference_table_utils` which reference `pg_dist_shard_placement`.
- [x] Drop `pg_dist_shard_placement_nodeid_index` and `DistShardPlacementNodeidIndexId` (unused)
- [x] Some comments in `worker_drop_protocol.c` should be changed to refer to the new table name
- [x] Also do something about `pg_dist_shard_placement_placementid_index`
- [ ] `master_disable_node` should become `maste_disable_group`? Same with `master_activate_node`.

lithp:
Fixes #1449


src/backend/distributed/connection/placement_connection.c
```diff
@@ -559,6 +559,8 @@ StartColocatedPlacementConnection(uint32 flags, ShardPlacement *placement,
 	 */
 	if (placement->partitionMethod != DISTRIBUTE_BY_HASH)
 	{
+		// TODO: This is where we place a function mapping groupid -> nodename, nodeport
+		// flags can later tell us whether secondaries are okay
```
lithp: @marcocitus My current plan is to remove nodename and nodeport from `ShardPlacement`. That would mean that here we lookup which node to actually connect to for a given group. Do you think that's reasonable considering a future where we might want to add logic which returns the preferred node given some information on the current query/transaction?

anarazel:
On 2017-06-14 09:46:14 -0700, Brian Cloutier wrote:
> lithp commented on this pull request.
> 
> 
> 
> > @@ -559,6 +559,8 @@ StartColocatedPlacementConnection(uint32 flags, ShardPlacement *placement,
>  	 */
>  	if (placement->partitionMethod != DISTRIBUTE_BY_HASH)
>  	{
> +		// TODO: This is where we place a function mapping groupid -> nodename, nodeport
> +		// flags can later tell us whether secondaries are okay
> 
> @marcocitus My current plan is to remove nodename and nodeport from `ShardPlacement`. That would mean that here we lookup which node to actually connect to for a given group. Do you think that's reasonable considering a future where we might want to add logic which returns the preferred node given some information on the current query/transaction?

I'm quite doubtful that's a good idea, we look that information up
frequently and in a good number of places.  What's the reason for
removing it?

lithp:
@anarazel Because `pg_shard_placement` won't have `nodename` and `nodeport`, so `ShardPlacement` won't be a good representation of the row it represents. 

An alternative is to do the extra lookup in `TupleToShardPlacement` but there it doesn't know whether we're inside a transaction or what query we're running, so in the future it won't have a good way of knowing which node to use when filling the placement.

Another alternative is to keep the fields but only populate them when needed, but that gets complicated an error prone, we have to keep track of whether the fields have been populated to a given `ShardPlacement`.

anarazel:
On 2017-06-14 10:10:35 -0700, Brian Cloutier wrote:
> @anarazel Because `pg_shard_placement` won't have `nodename` and `nodeport`, so `ShardPlacement` won't be a good representation of the row it represents. 

I think it's perfectly reasonable to have different in-memory and
on-disk datastructures, when they've different goals.

> An alternative is to do the extra lookup in `TupleToShardPlacement`
> but there it doesn't know whether we're inside a transaction or what
> query we're running, so in the future it won't have a good way of
> knowing which node to use when filling the placement.

I don't think that's right.  We shouldn't use the same placement for
different nodes depending on the current configuration, i.e. the
selection should be among different ShardPlacement, not anything
else. We want to be able to cache the ShardPlacements and use them for
purposes of identifying which connection has done what.

> Another alternative is to keep the fields but only populate them when
> needed, but that gets complicated an error prone, we have to keep
> track of whether the fields have been populated to a given
> `ShardPlacement`.

Yea, that seems worse.

- Andres

src/backend/distributed/connection/placement_connection.c
marcocitus: Can't the nodename/nodeport in ShardPlacement just reflect the name and port of the current primary of the group? Renaming it could make sense. Removing it seems impractical given how many code paths rely on those variables.

lithp:
@marcocitus Okay, the tests are passing. There's still some cleanup to do here but it's ready for some preliminary looking at.

marcocitus:
> By messing with pg_dist_node it's easy to get the database into a state where it refuses to do much work. If you run any command which attempts to read a shard placement it'll error out if the relevant groupid can't be found in pg_dist_node.

In the past we've treated these placements as inactive. Can we continue to do the same and not set the nodename/nodeport?

src/backend/distributed/citus--7.0-1--7.0-2.sql
```diff
@@ -0,0 +1,86 @@
...
+  ON DELETE TO pg_dist_shard_placement DO INSTEAD
+  DELETE FROM pg_dist_placement WHERE placementid=OLD.placementid;
+
+CREATE RULE pg_dist_shard_placement_update AS
+  ON UPDATE TO pg_dist_shard_placement DO INSTEAD
+  UPDATE pg_dist_placement
+    SET shardid = NEW.shardid, shardstate = NEW.shardstate,
+        shardlength = NEW.shardlength,
+	groupid = citus.find_groupid_for_node(NEW.nodename, NEW.nodeport)
+    WHERE placementid = NEW.placementid;
```
marcocitus: should this be `OLD.placementid` in case someone is updating the placement ID?

src/backend/distributed/master/master_metadata_utility.c
```diff
@@ -681,42 +711,56 @@ BuildShardPlacementList(ShardInterval *shardInterval)
...
+		TupleDesc tupleDescriptor = RelationGetDescr(pgPlacement);
 
 		ShardPlacement *placement = TupleToShardPlacement(tupleDescriptor, heapTuple);
+
+		uint32 groupId = placement->groupId;
+		WorkerNode *worker = NodeForGroup(groupId);
+
+		if (worker == NULL)
+		{
+			ereport(ERROR, (errmsg("the metadata is inconsistent"),
```
marcocitus: What if we treat this placement as inactive instead? That's what we've generally done if a node is not in pg_dist_node / pg_worker_list.conf. This may require some changes to code paths that handle inactive placements.

src/backend/distributed/utils/node_metadata.c
```diff
@@ -478,10 +529,17 @@ RemoveNodeFromCluster(char *nodeName, int32 nodePort)
...
 
-	DeleteNodeRow(nodeName, nodePort);
-
 	DeleteAllReferenceTablePlacementsFromNode(nodeName, nodePort);
 
+	hasAnyShardPlacements = NodeHasShardPlacements(nodeName, nodePort,
+												   onlyConsiderActivePlacements);
+	if (hasAnyShardPlacements)
+	{
+		ereport(ERROR, (errmsg("you cannot remove a node which has shard placements")));
```
marcocitus: Why consider inactive placements? Also, Were we planning to delete inactive shard placements?

src/test/regress/expected/multi_repair_shards.out
```diff
@@ -1,4 +1,6 @@
 ALTER SEQUENCE pg_catalog.pg_dist_shardid_seq RESTART 820000;
+SELECT groupid AS worker_2_group FROM pg_dist_node WHERE nodeport=:worker_2_port \gset
+SELECT groupid AS worker_1_group FROM pg_dist_node WHERE nodeport=:worker_1_port \gset
```
marcocitus: I would be kind of nice if we just set :worker_1_group to 1, and :worker_2_group to 2 for all tests and would then be able to use `SELECT master_add_node('localhost', :worker_1_port, group_id := :worker_1_group)` such that the group IDs are always known.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+### citus v7.0 (in development) ###
+
+* Replaces pg_dist_shard_placement metadata table with pg_dist_placement
```
marcocitus: We usually only add changelog entries for releases.

marcocitus:
> Add parameters to master_add_node(), allowing specifying a groupid

Yes.

> Replace master_disable_node() and master_activate_node() with master_disable_group() and master_disable_group()?

I don't think we need to change them right now. It's probably worth writing down what active vs. inactive means in the context of primaries vs. secondaries.

> Make it harder to mess up the metadata, disallow removing nodes if there are any placements?

I've been wondering whether we can add a foreign key from pg_dist_placement to pg_dist_node, but this probably won't work when we add node roles.

src/backend/distributed/citus--7.0-1--7.0-2.sql
lithp: Good question, I'll play with it and find out

src/backend/distributed/utils/node_metadata.c
lithp: This is linked to the decision above, if we allow removing nodes while there are still placements then removing a node with an inactive placement gets the system into a state where queries start erroring out.

src/test/regress/expected/multi_repair_shards.out
lithp: `:worker_1_group` might work but the second worker's group changes a ton, tests keep removing and adding the second node and it gets a new group every time.

CHANGELOG.md
lithp: I wasn't sure what to do here, I can remove this change but it seems preferable for commits to update the changelog when they touch something user facing? Having a useful git blame sounds nice this makes it less likely we'll forget to add something to the log when doing the release.

lithp:
> I've been wondering whether we can add a foreign key from pg_dist_placement to pg_dist_node, but this probably won't work when we add node roles.

This would work for now but won't work once groupid isn't unique. Once we have multiple nodes per group this could be made to work by adding a `pg_dist_group` table both `pg_dist_node` and `pg_dist_placement` have foreign keys to but that sounded like it might be too complicated if all we're getting is metadata that's easier to keep consistent.

src/backend/distributed/citus--7.0-1--7.0-2.sql
lithp: I wrote this assuming that it would disallow changing the `placementid` but it's worth making sure that's true.

src/backend/distributed/utils/node_metadata.c
lithp: Also, deleting inactive shard placements would be nice, I really wish we could make that foreign key! Without it reliably deleting them is difficult.

src/backend/distributed/master/master_metadata_utility.c
lithp: I considered two options:
- Don't return any ShardPlacements without a matching node, the other functions don't even see them
- Leave an empty nodeName/nodePort. 

The first one seemed nice but difficult to revert if we later had a function which wanted inactive shard placements.

The second one seemed bad for the same reasons removing nodeName/nodePort from ShardPlacement seemed bad, it potentially changes a lot of call sites!

I'm looking into it now, coming up with a list of functions which need to be changed.

src/backend/distributed/utils/node_metadata.c
marcocitus: Maybe add a `DeletePlacementsFromGroup(int groupId)` function that does a scan on groupid and performs `simple_heap_delete`s.

src/test/regress/expected/multi_repair_shards.out
marcocitus: but if we could specify it in `master_add_node` we could fix that, right?

(this is entirely optional)

CHANGELOG.md
marcocitus: That does sound reasonable. Could you check with @jasonmp85 on the right way to do this?

(iirc, the build scripts parse the changelog and are fairly strict about versioning and such)

src/backend/distributed/master/master_metadata_utility.c
lithp: Writing a longer comment with examples but the short version is that leaving ShardPlacements with empty nodeName/nodePort is error-prone and requires a lot of changes. Error-prone because there's no indication that `nodeName` might be `NULL` and did you know `strncmp(str, NULL, n)` is undefined?

If being so trigger-happy about erroring out isn't okay and we want to change these call sites, I'd feel more comfortable if we made a `GroupShardPlacement` and a `NodeShardPlacement`, this function returns `GroupShardPlacements` and later functions (such as `FinalizedShardPlacementList`) resolve them to `NodeShardPlacements` when required.

CHANGELOG.md
jasonmp85: I've been wanting to do this for a while. This `CHANGELOG` actually isn't parsed by anything (that's the `debian/changelog` file and `%changelog` in RPM spec files).

Make the header say `### citus v7.0.0 (Unreleased) ###` and you're good to go.

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -301,10 +301,12 @@ LoadShardInterval(uint64 shardId)
 
 
 /*
- * LoadShardPlacement returns the, cached, metadata about a shard placement.
+ * LoadShardPlacement returns the cached shard placement metadata
  *
  * The return value is a copy of the cached ShardPlacement struct and may
  * therefore be modified and/or freed.
+ *
+ * CAUTION: The ShardPlacements returned by this function may have a NULL nodeName
```
marcocitus: What's the reason we don't want to error out in this function?

lithp: There's only one caller and it doesn't touch the nodeName

src/backend/distributed/utils/node_metadata.c
lithp: @marcocitus I added a call to a newely created `DeleteAllShardPlacementesFromGroup` here which seems to work but surfaces a new error later on:

```
 DROP TABLE cluster_management_test;
+WARNING:  could not find any shard placements for shardId 1220001
+CONTEXT:  SQL statement "SELECT master_drop_all_shards(v_obj.objid, v_obj.schema_name, v_obj.object_name)"
+PL/pgSQL function citus_drop_trigger() line 21 at PERFORM
```

I think it's probably not safe to also delete any `ShardInterval`s which have been abandoned?

src/test/regress/expected/multi_repair_shards.out
lithp: I'm planning on addressing this, and adding `groupid` to master_add_node(), in the PR where we add `noderole`.

lithp:
@marcocitus I addressed the smaller comments and rebased and I think this is ready for you to look at again:

- I added a `GroupShardPlacement` but kept the existing `ShardPlacement` as-is to reduce the size of this PR. I also pushed the point where it resolves from when the cache is built to `ShardPlacementList`.

- I make a branch which deletes inactive placements when you remove nodes, but that breaks other code-paths because it's not possible to have orphaned shards (shards without placements). I hesitated to cascade to the orphaned shards because that sounds very difficult to recover from as a user. This PR's current behavior, where you can't remove nodes if there are any inactive placements, isn't so bad?

- I still have to double-check that the update rule on pg_dist_shard_placement doesn't let you change the placementid.

src/include/distributed/master_metadata_utility.h
```diff
@@ -69,14 +69,32 @@ typedef struct ShardInterval
...
+	RelayFileState shardState;
+	uint32 groupId;
+} GroupShardPlacement;
+
+
+/* A GroupShardPlacement which has had some extra data resolved */
 typedef struct ShardPlacement
 {
+	/*
+	 * careful, the rest of the code assumes this exactly matches GroupShardPlacement
```
marcocitus: The compiler could still mess things up by packing the struct. It would be safer to copy fields individually.

src/test/regress/expected/multi_metadata_sync.out
```diff
@@ -101,8 +101,8 @@ HINT:  Connect to worker nodes directly to manually change schemas of affected o
...
 ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
- TRUNCATE pg_dist_node
  SELECT worker_drop_distributed_table(logicalrelid) FROM pg_dist_partition
+ TRUNCATE pg_dist_node
```
marcocitus: What happens if this is above `worker_drop_distributed_table`?

src/test/regress/expected/multi_cluster_management.out
```diff
@@ -354,7 +352,7 @@ BEGIN;
 SELECT master_add_node('localhost', :worker_2_port);
            master_add_node           
 -------------------------------------
- (14,14,localhost,57638,default,f,t)
+ (13,13,localhost,57638,default,f,t)
```
marcocitus: to make the tests more robust to change, it would be nice to hide the groupid where possible. e.g. `SELECT 1 FROM master_add_node(...)`

src/backend/distributed/citus--7.0-1--7.0-2.sql
```diff
@@ -0,0 +1,86 @@
+/* citus--7.0-1--7.0-2.sql */
+
```
marcocitus: Should we do the rename of the groupid and nodeid sequences here?

src/test/regress/expected/multi_mx_modifications.out
```diff
@@ -446,6 +446,15 @@ END;
 COMMIT
 -- verify interaction of default values, SERIAL, and RETURNING
 \set QUIET on
+-- make sure this test always returns the same output no matter which tests have run
+SELECT last_value, min_value, max_value FROM app_analytics_events_mx_id_seq \gset
+ALTER SEQUENCE app_analytics_events_mx_id_seq NO MINVALUE NO MAXVALUE;
+SELECT setval('app_analytics_events_mx_id_seq'::regclass, 3940649673949184);
+      setval      
+------------------
+ 3940649673949184
```
marcocitus: Does this value have some special significance? or could we have used anything?

src/include/distributed/master_metadata_utility.h
anarazel: The compiler isn't allowed to pack structs without explicit instructions, would be an ABI violation (and cause crashes on a lot of platforms).

anarazel: But if you actually want to express this, it should probably just embed the other struct?

src/test/regress/expected/multi_mx_modifications.out
lithp: Nope, this is just the value which was already being used, by doing this I could leave the rest of the test alone and prove that this big change didn't change any behavior.

src/backend/distributed/citus--7.0-1--7.0-2.sql
lithp: Maybe that should be a separate PR? They seem super easy to rebase onto each other I can open another PR for it soon.

src/test/regress/expected/multi_cluster_management.out
lithp: Okay, good point.

src/test/regress/expected/multi_metadata_sync.out
lithp: I should try it again now that I've pushed back resolving the shard placement, but it used to fail because the placements didn't have associated nodes. Actually it probably still will, because we'll want to read out the node the placements are on in order to delete them.

src/include/distributed/master_metadata_utility.h
lithp: > But if you actually want to express this, it should probably just embed the other struct?

I thought about that but that requires an even bigger change, have to go change every place those fields are accessed. It's probably the right answer though, I'll push another commit

lithp:
@marcocitus For squashing, should I break this into multiple commits or squash it into one big one?

marcocitus:
If this can be separated into a small number of meaningful commits that would be nice.

src/backend/distributed/citus--7.0-2--7.0-3.sql
```diff
@@ -0,0 +1,86 @@
...
+  ON DELETE TO pg_dist_shard_placement DO INSTEAD
+  DELETE FROM pg_dist_placement WHERE placementid=OLD.placementid;
+
+CREATE RULE pg_dist_shard_placement_update AS
+  ON UPDATE TO pg_dist_shard_placement DO INSTEAD
+  UPDATE pg_dist_placement
+    SET shardid = NEW.shardid, shardstate = NEW.shardstate,
+        shardlength = NEW.shardlength,
+	groupid = citus.find_groupid_for_node(NEW.nodename, NEW.nodeport)
+    WHERE placementid = NEW.placementid;
```
marcocitus: Should change this to `OLD.placementid`. Updating the `placementid` gives a weird `UPDATE 0`.

src/backend/distributed/connection/placement_connection.c
```diff
@@ -240,7 +240,7 @@ StartPlacementConnection(uint32 flags, ShardPlacement *placement, const char *us
 		userName = freeUserName = CurrentUserName();
 	}
 
-	key.placementId = placement->placementId;
+	key.placementId = placement->placement.placementId;
```
marcocitus: We're introducing a lot of ambiguous naming in this last commit. I'm kind of inclined to omit the commit for the moment, merge this PR, and then maybe do the embedding a little more structurally (e.g. rename ShardPlacement to NodeShardPlacement and rename variables accordingly). What do you think?

src/backend/distributed/citus--7.0-1--7.0-2.sql
lithp: Fixed the original commit. Annoyingly the `UPDATE` still goes through when you attempt to update placementid but the tuple is unchanged.

src/backend/distributed/connection/placement_connection.c
lithp: That seems reasonable, about to do a giant rebase where I squash I'll leave it out

lithp:
Assuming travis likes it this is all yours!

src/backend/distributed/citus--7.0-1--7.0-2.sql
```diff
@@ -1,6 +1,5 @@
 /* citus--7.0-1--7.0-2.sql */
 
-/* redefine shard_name as STRICT */
```
marcocitus: this seems unnecessary

src/backend/distributed/citus--7.0-2--7.0-3.sql
```diff
@@ -0,0 +1,86 @@
...
+  ON DELETE TO pg_dist_shard_placement DO INSTEAD
+  DELETE FROM pg_dist_placement WHERE placementid=OLD.placementid;
+
+CREATE RULE pg_dist_shard_placement_update AS
+  ON UPDATE TO pg_dist_shard_placement DO INSTEAD
+  UPDATE pg_dist_placement
+    SET shardid = NEW.shardid, shardstate = NEW.shardstate,
+        shardlength = NEW.shardlength,
+	groupid = citus.find_groupid_for_node(NEW.nodename, NEW.nodeport)
+    WHERE placementid = OLD.placementid;
```
marcocitus: it's also missing a `SET placementid = NEW.placementid`.

src/backend/distributed/master/master_metadata_utility.c
```diff
@@ -834,25 +863,24 @@ InsertShardPlacementRow(uint64 shardId, uint64 placementId,
...
 	heapTuple = heap_form_tuple(tupleDescriptor, values, isNulls);
 
-	CatalogTupleInsert(pgDistShardPlacement, heapTuple);
+	CatalogTupleInsert(pgDistPlacement, heapTuple);
 
 	CitusInvalidateRelcacheByShardId(shardId);
 
 	CommandCounterIncrement();
-	heap_close(pgDistShardPlacement, RowExclusiveLock);
+	heap_close(pgDistPlacement, RowExclusiveLock);
```
marcocitus: not this PR's fault, but we should keep the `RowExclusiveLock` (replace with `NoLock`)

src/backend/distributed/master/master_metadata_utility.c
```diff
@@ -1045,73 +1073,54 @@ DeleteShardRow(uint64 shardId)
...
+	simple_heap_delete(pgDistPlacement, &heapTuple->t_self);
 	systable_endscan(scanDescriptor);
 
 	CitusInvalidateRelcacheByShardId(shardId);
 
 	CommandCounterIncrement();
-	heap_close(pgDistShardPlacement, RowExclusiveLock);
-
-	return placementId;
+	heap_close(pgDistPlacement, RowExclusiveLock);
```
marcocitus: same here

src/backend/distributed/citus--7.0-2--7.0-3.sql
```diff
@@ -0,0 +1,86 @@
...
+CREATE RULE pg_dist_shard_placement_delete AS
+  ON DELETE TO pg_dist_shard_placement DO INSTEAD
+  DELETE FROM pg_dist_placement WHERE placementid=OLD.placementid;
+
+CREATE RULE pg_dist_shard_placement_update AS
+  ON UPDATE TO pg_dist_shard_placement DO INSTEAD
+  UPDATE pg_dist_placement
+    SET shardid = NEW.shardid, shardstate = NEW.shardstate,
+        shardlength = NEW.shardlength,
+	groupid = citus.find_groupid_for_node(NEW.nodename, NEW.nodeport)
```
anarazel: Using RULEs is a *seriously* bad idea.  You're going to run into a lot of multiple-evaluation hazards.  I *STRONGLY* advise against doing this.  Have you tried to make these views plainly updatable? It does look possible to do so?  And if not, you should instead use `INSTEAD` triggers.

src/backend/distributed/citus--7.0-2--7.0-3.sql
```diff
@@ -0,0 +1,86 @@
...
+  RENAME TO pg_dist_placement_placementid_seq;
+
+ALTER TABLE pg_catalog.pg_dist_shard_placement
+  ALTER COLUMN placementid SET DEFAULT nextval('pg_catalog.pg_dist_placement_placementid_seq');
+
+CREATE TABLE citus.pg_dist_placement (
+  placementid BIGINT NOT NULL default nextval('pg_dist_placement_placementid_seq'::regclass),
+  shardid BIGINT NOT NULL,
+  shardstate INT NOT NULL,
+  shardlength BIGINT NOT NULL,
```
anarazel: Why's this NOT NULL?

src/backend/distributed/master/master_metadata_utility.c
```diff
@@ -221,37 +221,64 @@ DistributedTableSizeOnWorker(WorkerNode *workerNode, Oid relationId, char *sizeQ
 }
 
 
+List *
```
anarazel: Missing comment.

src/backend/distributed/master/master_metadata_utility.c
```diff
@@ -575,12 +601,13 @@ ShardLength(uint64 shardId)
...
- * NodeHasActiveShardPlacements returns whether any active shards are placed on this node
+ * NodeHasShardPlacements returns whether any active shards are placed on the group
+ * this node is a part of.
  */
 bool
-NodeHasActiveShardPlacements(char *nodeName, int32 nodePort)
+NodeHasShardPlacements(char *nodeName, int32 nodePort, bool onlyConsiderActivePlacements)
 {
-	const int scanKeyCount = 3;
+	const int scanKeyCount = (onlyConsiderActivePlacements ? 2 : 1);
```
anarazel: I'd move this down to the if block considering onlyConsiderActivePlacements - unnecessary to have to look back up for it...

src/backend/distributed/citus--7.0-1--7.0-2.sql
lithp: Oops, I must have accidentally removed this during a rebase.

src/backend/distributed/citus--7.0-2--7.0-3.sql
lithp: Okay, replaced it with a trigger. Making it simply updatable isn't possible because I'm using a JOIN and if I replace the JOIN with subselects `INSERT` isn't possible if you specify `nodename` and `nodeport`, which you're always going to want to do.

src/backend/distributed/citus--7.0-2--7.0-3.sql
lithp: No idea but it's been that way [for forever](https://github.com/citusdata/citus/blame/master/src/backend/distributed/citus.sql#L59).

lithp: And [we assume](https://github.com/citusdata/citus/blob/master/src/backend/distributed/master/master_metadata_utility.c?utf8=%E2%9C%93#L741) `NOT NULL`

src/backend/distributed/master/master_metadata_utility.c
lithp: Agree that looking up is annoying but the function is pretty small so that's not a big concern yet, and if I moved this down what how would I size `ScanKeyData scanKey[]`? I suppose I could use a `maxScanKeyCount` but the danger of that getting out of sync seems higher than the inconvenience of not immediately knowing what the value of `scanKeyCount` is.

lithp:
- Renamed all functions which return `ShardPlacement` or `List<ShardPlacement>`
- Fixed locking
- Added missing comment
- Replaced RULE with TRIGGER

Ready for another look!

src/backend/distributed/master/master_metadata_utility.c
```diff
@@ -589,25 +621,28 @@ NodeHasActiveShardPlacements(char *nodeName, int32 nodePort)
 	SysScanDesc scanDescriptor = NULL;
 	ScanKeyData scanKey[scanKeyCount];
```
anarazel: That's not going to work on all compilers - you can't size arrays by a variable amount.

src/backend/distributed/master/master_metadata_utility.c
```diff
@@ -681,76 +716,75 @@ BuildShardPlacementList(ShardInterval *shardInterval)
...
 
-		ShardPlacement *placement = TupleToShardPlacement(tupleDescriptor, heapTuple);
 		shardPlacementList = lappend(shardPlacementList, placement);
 
 		heapTuple = systable_getnext(scanDescriptor);
 	}
 
 	systable_endscan(scanDescriptor);
-	heap_close(pgShardPlacement, AccessShareLock);
+	heap_close(pgPlacement, AccessShareLock);
```
anarazel: NoLock.

src/backend/distributed/master/master_metadata_utility.c
```diff
@@ -589,25 +621,28 @@ NodeHasActiveShardPlacements(char *nodeName, int32 nodePort)
...
+	scanDescriptor = systable_beginscan(pgPlacement,
+										DistPlacementGroupidIndexId(), indexOK,
 										NULL, scanKeyCount, scanKey);
 
 	heapTuple = systable_getnext(scanDescriptor);
 	hasFinalizedPlacements = HeapTupleIsValid(heapTuple);
 
 	systable_endscan(scanDescriptor);
-	heap_close(pgShardPlacement, AccessShareLock);
+	heap_close(pgPlacement, AccessShareLock);
```
anarazel: NoLock.

src/test/regress/expected/multi_extension.out
```diff
@@ -112,6 +112,7 @@ ALTER EXTENSION citus UPDATE TO '6.2-3';
 ALTER EXTENSION citus UPDATE TO '6.2-4';
 ALTER EXTENSION citus UPDATE TO '7.0-1';
 ALTER EXTENSION citus UPDATE TO '7.0-2';
+ALTER EXTENSION citus UPDATE TO '7.0-3';
```
anarazel: I think we need a test testing that the moving of placements from one table to the other works.

src/backend/distributed/citus--7.0-2--7.0-3.sql
lithp: If we're using `heap_getattr` why do we care about the width of the struct?

src/backend/distributed/master/master_metadata_utility.c
lithp: Variable-length arrays are apparently kosher as of ISO c99?

src/backend/distributed/citus--7.0-2--7.0-3.sql
```diff
@@ -0,0 +1,97 @@
...
+      UPDATE pg_dist_placement
+        SET shardid = NEW.shardid, shardstate = NEW.shardstate,
+            shardlength = NEW.shardlength, placementid = NEW.placementid,
+            groupid = citus.find_groupid_for_node(NEW.nodename, NEW.nodeport)
+        WHERE placementid = OLD.placementid;
+      RETURN NEW;
+    ELSIF (TG_OP = 'INSERT') THEN
+      INSERT INTO pg_dist_placement
+        (placementid, shardid, shardstate, shardlength, groupid)
+	VALUES (COALESCE(NEW.placementid, nextval('pg_dist_placement_placementid_seq')),
```
lithp: I didn't realize you could do that this is cool. It's a little cleaner too

src/backend/distributed/master/master_metadata_utility.c
lithp: I think the reason for this change to `NoLock` is that callers of `NodeHasShardPlacements` are going to take action based on the result and it would be nice if we didn't `DROP pg_dist_placement` or make any drastic changes while this transaction is still running?

This makes sense, but now I'm confused why `heap_close` includes a parameter to drop locks, it seems you would _never_ want to drop locks until the transaction ends.

lithp:
Does this mean that these are also wrong? They all release their `RowExclusiveLock`s

- [InsertShardRow](https://github.com/citusdata/citus/blob/master/src/backend/distributed/master/master_metadata_utility.c#L809)
- [DeletePartitionRow](https://github.com/citusdata/citus/blob/master/src/backend/distributed/master/master_metadata_utility.c#L963)
- [DeleteShardRow](https://github.com/citusdata/citus/blob/master/src/backend/distributed/master/master_metadata_utility.c#L1005)

marcocitus:
> Does this mean that these are also wrong? 

Yes, all wrong.

> DistributedTableSize also releases an AcessShareLock, which matches this pattern, but I can't think of a bug it would cause.

That's a case where it seems fine.

lithp:
Ready to be looked at

- Fixed all the exclusive locks being released early in metadata_utility
- Rebased onto master
- Added a test for upgrading existing data to the pg_dist_placement
- Used a default column on the pg_dist_shard_placement view

lithp:
@anarazel @marcocitus 

I believe I've addressed all comments, since you've both approved the changes I'll rebase and merge this tomorrow if I don't hear back with more comments.

src/backend/distributed/citus--7.0-2--7.0-3.sql
```diff
@@ -0,0 +1,97 @@
...
+      UPDATE pg_dist_placement
+        SET shardid = NEW.shardid, shardstate = NEW.shardstate,
+            shardlength = NEW.shardlength, placementid = NEW.placementid,
+            groupid = citus.find_groupid_for_node(NEW.nodename, NEW.nodeport)
+        WHERE placementid = OLD.placementid;
+      RETURN NEW;
+    ELSIF (TG_OP = 'INSERT') THEN
+      INSERT INTO pg_dist_placement
+        (placementid, shardid, shardstate, shardlength, groupid)
+	VALUES (NEW.placementid, NEW.shardid, NEW.shardstate, NEW.shardlength,
```
marcocitus: Alignment is a little off

src/backend/distributed/utils/node_metadata.c
```diff
@@ -439,7 +488,7 @@ ReadWorkerNodes()
 	}
 
 	systable_endscan(scanDescriptor);
-	heap_close(pgDistNode, AccessExclusiveLock);
+	heap_close(pgDistNode, AccessShareLock);
```
marcocitus: NoLock

marcocitus:
Rebased this on top of the placement connection changes, changed `FindShardPlacementOnNode` to `FindShardPlacementOnGroup` and updated `BuildPlacementSelectList` to use `groupId`.

Merging this to prevent further conflicts.

jasonmp85:
This merge (broke enterprise)[https://travis-ci.com/citusdata/citus-enterprise/jobs/82724867]. The `CHANGELOG` change causes a merge conflict, but after that's fix (which is what the above job does), compilation fails entirely.

lithp:
@jasonmp85 I've opened a PR with fixes: https://github.com/citusdata/citus-enterprise/pull/174

Issue #1273, opened 2017-03-06
Unify Citus executors


marcocitus:
Proposal for a new executor to replace all other executors, but with the initial goal of replacing the real-time and multi-shard router executors with one that addresses most of their shortcomings.

## Why a new executor?

As we're expanding our SQL support for multi-shard queries (#1204), we expect Citus to power more applications that query across shards. However, in current Citus workloads the multi-shard executors have several problems:

- Not transactional - #1119
- Excessive number of connections, TIME_WAIT issues (real-time, multi-shard router)
- Might not be able to complete due to number of shards (multi-shard router) - #709
- No connection reuse (real-time, multi-shard router)
- Does not consider query time / data imbalances (real-time, multi-shard router)
- Requires complex pgbouncer configuration (real-time, multi-shard router)
- No query parameter support - #1179
- Deployment issues (task-tracker) - #564, requires .pgpass on workers
- Poor performance (task-tracker)
- Not well-understood (task-tracker)
- Implementation issues (mainly task-tracker) - #716, #255, #353, #69
- Need to manually select executor - #444
- Technical debt - #292, #294, #296, #1121, #1260

We currently have 4 different query executors: router, real-time, task-tracker, and multi-shard router (DDL/INSERT..SELECT/master_modify_multiple_shards), all with different properties. While this could be considered a way of addressing different trade-offs for different types of queries, the main trade-off for most of them was expediency. Actual technical differences could be represented through a small number of parameters in a more general executor.

## General Approach 

The unified executor establishes a small connection pool to each worker at the start of a query and assigns tasks to connections as they become available. Each connection has its own queue for tasks that need to be assigned to it because of preceding writes in the same transaction, and other tasks are added to a general queue. Once a connection is established and optionally completes BEGIN, it can either take a task from its own queue, or one from the general queue, and keeps doing so until all tasks are done.

## Advantages

By using small, possibly cached connection pools, we can reduce or even eliminate connection churn from multi-shard SELECT queries. We can also support SELECT and DDL with a much larger number of concurrent queries and much larger shard count, which also enables new sharding approaches (e.g. range-distribution, multi-level sharding). We can also allow users to adjust the pool size to increase or decrease resources used by a particular query.

[...]

PR #1455, opened 2017-06-16
Rework connection API to remove transaction restrictions

### LINKED ISSUE #965: Transactions with DDL on a reference table and INSERT INTO ... SELECT deadlocks
Though it doesn't look a common pattern, the following scenario leads to a deadlock:

```SQL

SET citus.shard_count = 4;
SET citus.shard_replication_factor = 2;

CREATE TABLE raw_events_first (user_id int, time timestamp, value_1 int, value_2 int, value_3 float, value_4 bigint, UNIQUE(user_id, value_1));
SELECT create_distributed_table('raw_events_first', 'user_id');

CREATE TABLE agg_events (user_id int, value_1_agg int, value_2_agg int, value_3_agg float, value_4_agg bigint, agg_time timestamp, UNIQUE(user_id, value_1_agg));
SELECT create_distributed_table('agg_events', 'user_id');;

-- create the reference table as well
CREATE TABLE reference_table (user_id int);

-- start a new tranasction
BEGIN;

-- alter the reference table
ALTER TABLE reference_table ADD COLUMN test int;

-- this command is blocked
INSERT INTO agg_events (user_id)
SELECT 
    raw_events_first.user_id 
FROM 
    raw_events_first, reference_table 
WHERE 
    raw_events_first.user_id = reference_table.user_id;
```

The reason for that is the reference table is not colocated with the raw_events_first. Thus, we don't send the query from the same connection as the DDL command. In the end, both connections wait for each other.

Note that, we fixed this for DDL commands on colocated table + INSERT INTO ... SELECT with #939.

### LINKED ISSUE #1107: Remove XactModificationLevel and rely on placement API instead
When we introduced transactions in Citus 5.2, we initially opened up the possibility for self deadlock as remote transaction are kept open until the master commits and commands in the same transaction that are performed over separate connections could block waiting for the still-open transactions. 

As a quick, rudimentary solution we introduced the global XactModificationLevel variable, which prevents incompatible commands from appearing the same transaction. For example, a DML command would set the XactModificationLevel to `XACT_MODIFICATION_DATA`. Other DML commands are still allowed if XactModificationLevel is `XACT_MODIFICATION_DATA` since they would use the same connection(s) to the worker(s), but DDL commands are not since they would open new connections and perform a conflicting command. On the other hand, DDL sets the XactModificationLevel to `XACT_MODIFICATION_MULTI_SHARD`, which allows other DDL commands to follow since they would use the same set of connections, but not DML.

This solution requires the developer to consider the appropriate XactModificationLevel to check for and set for every remote command and it is overly restrictive. For example, we could allow DML and DDL to appear in the same transaction block as long as they use the same connection for the same (set of co-located) shard placements.

With #1079 we are introducing logic to find a connection based on a placement ID. It already prevents most situations in which self-deadlocks could occur. Once we resolve #873, we should be in a good position to remove the XactModificationLevel restrictions and allow a wider range of transactions.

### LINKED ISSUE #1442: Copy may not be visible in join with reference table
When joining with a reference table after a `COPY` into a distributed table, we may pick a connection that cannot see the effects of the `COPY`.

```
CREATE TABLE distributed_table (
  x int,
  y int
);
SELECT create_distributed_table('distributed_table', 'x');

CREATE TABLE reference_table (
  y int,
  z int
);
INSERT INTO reference_table VALUES (2,9);

BEGIN;
-- opens two separate connections to node 
\copy distributed_table FROM STDIN WITH CSV
2,2
4,2
\.

-- Uses first connection to node which coincidentally wrote the row with x = 2
SELECT * FROM distributed_table JOIN reference_table USING (y) WHERE x = 2;
 y | x | z 
---+---+---
 2 | 2 | 9
(1 row)

-- Also uses first connection to node, does not see the row with x = 4
SELECT * FROM distributed_table JOIN reference_table USING (y) WHERE x = 4;
 y | x | z 
---+---+---
(0 rows)
END;
```

The reason is that we use the reference table shard as the anchor shard and thus pick the same connection in both cases, whereas the row with `x = 4` is only visible over the second connection.

More generally, we're not properly enforcing connection reuse when reference tables are involved, because we're not considering all placements queried by a SELECT.

marcocitus:
This PR introduces a `GetPlacementListConnection` function that can be used to reliably get a connection for a query that accesses more than one placement. `GetPlacementListConnection` takes a `ShardPlacementAccess` list which contains the placement and the operation to perform (SELECT, DML, DDL).

`GetPlacementListConnection` will error out if: 
- A placement (or one of its co-located placements) was modified over a connection that cannot be used. This is typically due to DML+DDL because the DDL executor might open a connection for the first shard, claim it exclusively, and then it turns out it actually needs it for the second shard.
- A placement (or one of its co-located placements) was read over a connection that cannot be used and we are about to perform DDL on it. This is typically due to performing DDL on a reference table after a join with a distributed table.
- Placements (or co-located placements) in the access list were modified over more than one connection. This is typically due to joining a distributed table with a reference  table after modifying both.

Otherwise, `GetPlacementListConnection` returns:
- A connection that previously modified one or more of the placements in the access list, or one of their co-located placements.
- or, an (arbitrary) connection that previously read from one of the placements
- or, the cached connection to the node
- or, a new connection to the node

The router executor uses `GetPlacementListConnection` for both single-shard (mainly relevant for SELECT), and multi-shard (mainly relevant for INSERT ... SELECT) queries. It should be trivial to extend this to UPDATE/DELETE with SELECT. 

The PR also removes the distinction between `XACT_MODIFICATION_DATA` and `XACT_MODIFICATION_MULTI_SHARD`, which was still necessary due to issues that arise when using reference tables and distributed tables in the same transaction.

To simplify the code, I removed the notion of `connectionReferences` and replaced it with a `hasSecondaryConnections` flag that indicates whether any additional connections read from a placement.

This still needs a few more regression tests to test new capabilities.

Fixes #1442
Fixes #1107 
Fixes #965

### DIFF
--- a/src/backend/distributed/connection/placement_connection.c
+++ b/src/backend/distributed/connection/placement_connection.c
@@ -341,8 +341,10 @@ StartPlacementListConnection(uint32 flags, List *placementAccessList,
 
 			ereport(ERROR,
 					(errcode(ERRCODE_ACTIVE_SQL_TRANSACTION),
-					 errmsg("cannot perform DDL on a placement which has been read over "
-							"multiple connections")));
+					 errmsg(
+						 "cannot perform DDL on placement %ld, which has been read over "
+						 "multiple connections",
+						 placement->placementId)));
 		}
 		else if (accessType == PLACEMENT_ACCESS_DDL && colocatedEntry != NULL &&
 				 colocatedEntry->hasSecondaryConnections)
@@ -358,8 +360,9 @@ StartPlacementListConnection(uint32 flags, List *placementAccessList,
 
 			ereport(ERROR,
 					(errcode(ERRCODE_ACTIVE_SQL_TRANSACTION),
-					 errmsg("cannot perform DDL on a placement if a co-located placement "
-							"has been read over multiple connections")));
+					 errmsg("cannot perform DDL on placement %ld since a co-located "
+							"placement has been read over multiple connections",
+							placement->placementId)));
 		}
 		else if (foundModifyingConnection)
 		{
@@ -411,8 +414,9 @@ StartPlacementListConnection(uint32 flags, List *placementAccessList,
 
 			ereport(ERROR,
 					(errcode(ERRCODE_ACTIVE_SQL_TRANSACTION),
-					 errmsg("cannot establish new placement connection when DDL has "
-							"been executed on existing placement connection")));
+					 errmsg("cannot establish a new connection for placement %ld, since "
+							"DDL has been executed on a connection that is in use",
+							placement->placementId)));
 		}
 		else if (placementConnection->hadDML)
 		{
@@ -432,8 +436,9 @@ StartPlacementListConnection(uint32 flags, List *placementAccessList,
 
 			ereport(ERROR,
 					(errcode(ERRCODE_ACTIVE_SQL_TRANSACTION),
-					 errmsg("cannot establish new placement connection when DML has "
-							"been executed on existing placement connection")));
+					 errmsg("cannot establish a new connection for placement %ld, since "
+							"DML has been executed on a connection that is in use",
+							placement->placementId)));
 		}
 		else if (accessType == PLACEMENT_ACCESS_DDL)
 		{
@@ -450,7 +455,7 @@ StartPlacementListConnection(uint32 flags, List *placementAccessList,
 
 			ereport(ERROR,
 					(errcode(ERRCODE_ACTIVE_SQL_TRANSACTION),
-					 errmsg("cannot perform parallel DDL command because multiple "
+					 errmsg("cannot perform a parallel DDL command because multiple "
 							"placements have been accessed over the same connection")));
 		}
 		else
--- a/src/backend/distributed/connection/placement_connection.c
+++ b/src/backend/distributed/connection/placement_connection.c
@@ -14,7 +14,9 @@
 #include "access/hash.h"
 #include "distributed/connection_management.h"
 #include "distributed/hash_helpers.h"
+#include "distributed/master_protocol.h"
 #include "distributed/metadata_cache.h"
+#include "distributed/multi_planner.h"
 #include "distributed/placement_connection.h"
 #include "utils/hsearch.h"
 #include "utils/memutils.h"
@@ -47,15 +49,12 @@ typedef struct ConnectionReference
 	bool hadDML;
 	bool hadDDL;
 
-	/*
-	 * Membership in ConnectionPlacementHashEntry->connectionReferences or
-	 * ColocatedPlacementsHashEntry->connectionReferences.
-	 */
-	dlist_node placementNode;
-
 	/* membership in MultiConnection->referencedPlacements */
 	dlist_node connectionNode;
 } ConnectionReference;
+
+
+struct ColocatedPlacementsHashEntry;
 
 
 /*
@@ -85,11 +84,14 @@ typedef struct ConnectionPlacementHashEntry
 	/* did any remote transactions fail? */
 	bool failed;
 
-	/* list of connections to remote nodes */
-	dlist_head connectionReferences;
+	/* primary connection used to access the placement */
+	ConnectionReference *primaryConnection;
 
-	/* if non-NULL, connection executing DML/DDL*/
-	ConnectionReference *modifyingConnection;
+	/* are any other connections reading from the placements? */
+	bool hasSecondaryConnections;
+
+	/* entry for the set of co-located placements */
+	struct ColocatedPlacementsHashEntry *colocatedEntry;
 
 	/* membership in ConnectionShardHashEntry->placementConnections */
 	dlist_node shardNode;
@@ -137,11 +139,11 @@ typedef struct ColocatedPlacementsHashEntry
 {
 	ColocatedPlacementsHashKey key;
 
-	/* list of connections to remote nodes */
-	dlist_head connectionReferences;
+	/* primary connection used to access the co-located placements */
+	ConnectionReference *primaryConnection;
 
-	/* if non-NULL, connection executing DML/DDL*/
-	ConnectionReference *modifyingConnection;
+	/* are any other connections reading from the placements? */
+	bool hasSecondaryConnections;
 }  ColocatedPlacementsHashEntry;
 
 static HTAB *ColocatedPlacementsHash;
@@ -172,19 +174,10 @@ typedef struct ConnectionShardHashEntry
 static HTAB *ConnectionShardHash;
 
 
-static MultiConnection * StartColocatedPlacementConnection(uint32 flags,
-														   ShardPlacement *placement,
-														   const char *userName);
-static ConnectionReference * CheckExistingPlacementConnections(uint32 flags,
-															   ConnectionPlacementHashEntry
-															   *placementEntry,
-															   const char *userName);
-static ConnectionReference * CheckExistingColocatedConnections(uint32 flags,
-															   ColocatedPlacementsHashEntry
-															   *connectionEntry,
-															   const char *userName);
+static ConnectionPlacementHashEntry * FindOrCreatePlacementEntry(
+	ShardPlacement *placement);
 static bool CanUseExistingConnection(uint32 flags, const char *userName,
-									 ConnectionReference *connectionReference);
+									 ConnectionReference *placementConnection);
 static void AssociatePlacementWithShard(ConnectionPlacementHashEntry *placementEntry,
 										ShardPlacement *placement);
 static bool CheckShardPlacements(ConnectionShardHashEntry *shardEntry);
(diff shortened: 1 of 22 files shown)


src/test/regress/sql/multi_mx_modifying_xacts.sql
```diff
@@ -150,12 +150,6 @@ INSERT INTO labs_mx VALUES (6, 'Bell labs_mx');
 SELECT count(*) FROM researchers_mx WHERE lab_id = 6;
 ABORT;
 
--- applies to DDL
-BEGIN;
-INSERT INTO labs_mx VALUES (6, 'Bell labs_mx');
-ALTER TABLE labs_mx ADD COLUMN motto text;
```
marcocitus: This now falls back to the `operation is not allowed on this node` messages, which is already covered by other tests.

src/backend/distributed/planner/multi_explain.c
```diff
@@ -328,17 +328,12 @@ RemoteExplain(Task *task, ExplainState *es)
...
-
-			continue;
+			/* read explain query results */
+			remotePlan->explainOutputList = ReadFirstColumnAsText(queryResult);
 		}
 
-		/* read explain query results */
-		remotePlan->explainOutputList = ReadFirstColumnAsText(queryResult);
-
 		PQclear(queryResult);
```
anarazel: That seems fairly independent?

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -642,6 +662,61 @@ ExecuteSingleSelectTask(CitusScanState *scanState, Task *task)
 
 
 /*
+ * BuildPlacementSelectList builds a list of SELECT placement accesses
+ * which can be used to call StartPlacementListConnection or
+ * GetPlacementListConnection.
+ */
+List *
+BuildPlacementSelectList(char *nodeName, int nodePort, List *relationShardList)
```
anarazel: Shouldn't this also be used in GetModifyConnections() / ExecuteSingleModifyTask() / ExecuteSingleSelectTask()?

I'd have commented on those functions, if github would allow to do so... What on earth is the point of not allowing to comment on unchanged pieces of code?

src/backend/distributed/connection/placement_connection.c
```diff
@@ -229,272 +222,373 @@ GetPlacementConnection(uint32 flags, ShardPlacement *placement, const char *user
...
-		if (flags & FOR_DDL)
+		else if (placementConnection->hadDML)
 		{
-			/* would deadlock otherwise */
 			ereport(ERROR,
 					(errcode(ERRCODE_ACTIVE_SQL_TRANSACTION),
-					 errmsg("cannot establish new placement connection for DDL when "
-							"other connections exist")));
+					 errmsg("cannot establish new placement connection when DML has "
+							"been executed on existing placement connection")));
```
anarazel: Wonder if we should, at a later point, distinguish between read-modify-write DML and write DML.  The latter could be run a good chunk more cases.

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -642,6 +662,61 @@ ExecuteSingleSelectTask(CitusScanState *scanState, Task *task)
...
+		List *shardPlacementList = NIL;
+		ShardPlacement *placement = NULL;
+		ShardPlacementAccess *placementAccess = NULL;
+		bool missingOK = false;
+
+		shardPlacementList = FinalizedShardPlacementList(relationShard->shardId);
+
+		placement = SearchShardPlacementInList(shardPlacementList, nodeName, nodePort,
+											   missingOK);
+		if (placement == NULL)
```
anarazel: Getting the final list, which will first build the full list, and then searching in it isn't exactly elegant & causes a good chunk of allocations...

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -642,6 +662,61 @@ ExecuteSingleSelectTask(CitusScanState *scanState, Task *task)
...
+ * and access type.
+ */
+static ShardPlacementAccess *
+CreatePlacementAccess(ShardPlacement *placement, ShardPlacementAccessType accessType)
+{
+	ShardPlacementAccess *placementAccess = NULL;
+
+	placementAccess = (ShardPlacementAccess *) palloc0(sizeof(ShardPlacementAccess));
+	placementAccess->placement = placement;
+	placementAccess->accessType = PLACEMENT_ACCESS_SELECT;
```
anarazel: The function has a accessType parameter, which seems to be ignored in favor of  PLACEMENT_ACCESS_SELECT?

src/backend/distributed/connection/placement_connection.c
```diff
@@ -47,17 +49,14 @@ typedef struct ConnectionReference
 	bool hadDML;
 	bool hadDDL;
 
-	/*
-	 * Membership in ConnectionPlacementHashEntry->connectionReferences or
-	 * ColocatedPlacementsHashEntry->connectionReferences.
-	 */
-	dlist_node placementNode;
```
anarazel: I'm not quite sure why you changed this to achieve the larger goal?  It still seems useful to get other connections that touched a placement?

src/backend/distributed/executor/multi_router_executor.c
marcocitus: Added it to `GetModifyConnections()`, which is called by `ExecuteSingleModifyTask`

src/backend/distributed/connection/placement_connection.c
marcocitus: To simplify the implementation of `GetPlacementListConnection`. There is no reason to keep track of all connections that accessed a placement right now.

I guess we could still shove it in a list and not use it...

anarazel: FWIW, yes, I think that'd be a bit better.

onderkalaci:
Just a note that there is a dependency among #1467 -> #1436  -> This PR (#1455). That's why I decided to review this now and enable the waiting PRs as soon as possible.

src/backend/distributed/connection/placement_connection.c
```diff
@@ -229,272 +222,373 @@ GetPlacementConnection(uint32 flags, ShardPlacement *placement, const char *user
...
+			 */
 			ereport(ERROR,
 					(errcode(ERRCODE_ACTIVE_SQL_TRANSACTION),
-					 errmsg("cannot establish new placement connection when DML has been "
-							"executed on existing placement connection")));
+					 errmsg("cannot perform DDL on a placement if a co-located placement "
+							"has been read over multiple connections")));
 		}
-		else
+		else if (foundModifyingConnection)
```
marcocitus: There are several improvements we can make to increase the set of multi-statement transactions that are supported, which avoids these situations:

#1482 - enable single DML / COPY + DDL/COPY (should aim to fix this in 7.0)
#1273 - enable multiple DML + DDL
#1484 - enable multiple DML + COPY

We cannot support your last example. 

I don't see an easy way to error out in sometimes-unsupported scenarios. I think we'd need to start keeping track of things at a table level. If non-co-located tables have been modified over multiple connections then we should error out if a query reads from both of those tables.

src/backend/distributed/connection/placement_connection.c
```diff
@@ -229,272 +222,373 @@ GetPlacementConnection(uint32 flags, ShardPlacement *placement, const char *user
...
+	 *
+	 * If one or more of the placements have been modified in this transaction, then
+	 * use the connection that performed the write. If placements have been written
+	 * over multiple connections or the connection is not available, error out.
+	 *
+	 * If placements have only been read in this transaction, then use the last
+	 * suitable connection found for a placement in the placementAccessList.
 	 */
-	if (placementEntry->modifyingConnection)
+	foreach(placementAccessCell, placementAccessList)
```
marcocitus: Not sure that's a good idea. There could well be multiple connections in a single statement outside of a transaction block (e.g. in a UDF call). Fortunately, iterating through a list of size 1 is probably not a performance bottleneck. :)

src/test/regress/expected/multi_modifying_xacts.out
```diff
@@ -383,16 +381,29 @@ ORDER BY nodeport;
...
 \copy labs from stdin delimiter ','
 ROLLBACK;
--- but not if COPY precedes ALTER TABLE
 BEGIN;
 \copy labs from stdin delimiter ','
 ALTER TABLE labs ADD COLUMN motto text;
-ERROR:  distributed DDL commands must not appear within transaction blocks containing single-shard data modifications
+ABORT;
+-- cannot perform DDL once a connection is used for multiple shards
+BEGIN;
```
onderkalaci: Just noting: I think we don't strictly need that the connection is used for multiple shards before the `DDL`. If we're to with the shard that is not the first shard, we still end up with the error.

To do that, simply remove the first `SELECT` query.

onderkalaci:
@anarazel 

> I think one question will be how much we want to use this new infrastructure before merging...

With this PR, as far as I can see, all router queries,  multi-shard modifications,`COPY`, `DROP shards` and `DDL`s use this infrastructure. That should be all we could currently support, right? 

Also, as far as I can see (and a brief discussion with Marco), Real-time/ task-tracker becomes more applicable with this infrastructure.

src/backend/distributed/connection/placement_connection.c
```diff
@@ -229,272 +222,373 @@ GetPlacementConnection(uint32 flags, ShardPlacement *placement, const char *user
...
+StartPlacementListConnection(uint32 flags, List *placementAccessList,
+							 const char *userName)
 {
-	dlist_iter it;
+	char *freeUserName = NULL;
+	bool foundModifyingConnection = false;
+	ListCell *placementAccessCell = NULL;
+	List *placementEntryList = NIL;
+	ListCell *placementEntryCell = NULL;
+	MultiConnection *connection = NULL;
```
marcocitus: `connection` is not so much specific to 1 particular placement, but is to be used for all placements in the list, unlike `placementConnection`. Maybe something like `chosenConnection`? Could also do `placementListConnection` maybe.

src/backend/distributed/connection/placement_connection.c
```diff
@@ -229,272 +222,373 @@ GetPlacementConnection(uint32 flags, ShardPlacement *placement, const char *user
...
+			/*
+			 * There are now multiple connections that read from the placement
+			 * and DDL commands are forbidden.
+			 */
+			placementEntry->hasSecondaryConnections = true;
+
+			if (placementEntry->colocatedEntry != NULL)
+			{
+				/* we also remember this for co-located placements */
+				placementEntry->hasSecondaryConnections = true;
```
marcocitus: Good catch!

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -601,13 +602,7 @@ ExecuteSingleSelectTask(CitusScanState *scanState, Task *task)
...
 	ListCell *taskPlacementCell = NULL;
 	char *queryString = task->queryString;
-
-	if (XactModificationLevel == XACT_MODIFICATION_MULTI_SHARD)
-	{
-		ereport(ERROR, (errcode(ERRCODE_ACTIVE_SQL_TRANSACTION),
-						errmsg("single-shard query may not appear in transaction blocks "
-							   "which contain multi-shard data modifications")));
-	}
+	List *relationShardList = task->relationShardList;
```
marcocitus: Makes sense.

src/backend/distributed/connection/placement_connection.c
```diff
@@ -229,272 +222,373 @@ GetPlacementConnection(uint32 flags, ShardPlacement *placement, const char *user
...
-			/* would deadlock otherwise */
 			ereport(ERROR,
 					(errcode(ERRCODE_ACTIVE_SQL_TRANSACTION),
-					 errmsg("cannot establish new placement connection for DDL when "
-							"other connections exist")));
+					 errmsg("cannot establish new placement connection when DML has "
+							"been executed on existing placement connection")));
 		}
-		else if (flags & FOR_DML)
+		else if (accessType == PLACEMENT_ACCESS_DDL)
```
marcocitus: At this point, there can only have been reads to co-located shard placements (otherwise `hadDML` would have been true), and DML does not conflict with reads.

src/test/regress/expected/multi_modifying_xacts.out
marcocitus: True, I'm increasingly convinced we need to fix #1482 soon.

marcocitus:
Added a bunch more regression tests to exercise all scenarios, clarified error messages, added the asserts.

src/backend/distributed/connection/placement_connection.c
```diff
@@ -229,272 +222,439 @@ GetPlacementConnection(uint32 flags, ShardPlacement *placement, const char *user
...
+		if (placement->partitionMethod == DISTRIBUTE_BY_HASH ||
+			placement->partitionMethod == DISTRIBUTE_BY_NONE)
+		{
+			ColocatedPlacementsHashKey key;
+			ColocatedPlacementsHashEntry *colocatedEntry = NULL;
+
+			strcpy(key.nodeName, placement->nodeName);
+			key.nodePort = placement->nodePort;
+			key.colocationGroupId = placement->colocationGroupId;
+			key.representativeValue = placement->representativeValue;
```
marcocitus: I guess mathematically it doesn't matter much. `int` would seem a bit easier for debugging purposes as it matches pg_dist_shard, while `uint` is a bit easier to reason about.

src/backend/distributed/transaction/multi_shard_transaction.c
```diff
@@ -69,9 +72,24 @@ OpenTransactionsToAllShardPlacements(List *shardIntervalList, int connectionFlag
...
+			/* can only open connections for DDL and DML commands */
+			Assert(task->taskType == DDL_TASK);
+
+			accessType = PLACEMENT_ACCESS_DDL;
+		}
+
 		foreach(placementCell, shardPlacementList)
 		{
 			ShardPlacement *shardPlacement = (ShardPlacement *) lfirst(placementCell);
+			ShardPlacementAccess placementModification;
```
marcocitus: In this case it's easy to allocate on the stack, which is always simpler and more efficient.

marcocitus:
> Could you please confirm the above before merging it?

Correct.

PR #1457, opened 2017-06-16
Fix crash during local CREATE INDEX CONCURRENTLY

### LINKED ISSUE #1456: No-name CREATE INDEX CONCURRENTLY against local tables crashes
As reported by @onderkalaci [in this comment](https://github.com/citusdata/citus/pull/1379#issuecomment-307027946):

```SQL

CREATE TABLE foo(id integer, name text);

CREATE INDEX CONCURRENTLY on foo(id);
-- server closed the connection unexpectedly
-- 	This probably means the server terminated abnormally
-- 	before or while processing the request.
-- The connection to the server was lost. Attempting reset: Failed.
-- !>

```

Backtrace:
```
Program received signal EXC_BAD_ACCESS, Could not access memory.
Reason: KERN_INVALID_ADDRESS at address: 0x0000000000000000
0x00007fff9a0fd132 in strlen ()
(gdb) bt
#0  0x00007fff9a0fd132 in strlen ()
#1  0x00000001025fb98c in hashname (fcinfo=0x7fff5d65d7b0) at hashfunc.c:148
#2  0x0000000102c17669 in DirectFunctionCall1Coll (func=0x1025fb960 <hashname>, collation=0, arg1=0) at fmgr.c:717
#3  0x0000000102be6c98 in CatalogCacheComputeHashValue (cache=0x7fdd748368d0, nkeys=2, cur_skey=0x7fff5d65dcd0) at catcache.c:209
#4  0x0000000102be6143 in SearchCatCache (cache=0x7fdd748368d0, v1=0, v2=2200, v3=0, v4=0) at catcache.c:1130
#5  0x0000000102c01b94 in SearchSysCache (cacheId=45, key1=0, key2=2200, key3=0, key4=0) at syscache.c:1109
#6  0x0000000102c01ce3 in GetSysCacheOid (cacheId=45, key1=0, key2=2200, key3=0, key4=0) at syscache.c:1187
#7  0x0000000102c0547c in get_relname_relid (relname=0x0, relnamespace=2200) at lsyscache.c:1685
#8  0x00000001033556cd in PostProcessUtility (parsetree=0x7fdd74009e00) at executor/multi_utility.c:3063
#9  0x0000000103353e1a in multi_ProcessUtility10 (pstmt=0x7fdd7400a1c0, queryString=0x7fdd74009240 "CREATE INDEX CONCURRENTLY on prange1_2007_local(time);", context=PROCESS_UTILITY_TOPLEVEL, params=0x0, queryEnv=0x0, dest=0x7fdd7400a2b8, completionTag=0x7fff5d65e3d0 "") at executor/multi_utility.c:439
#10 0x0000000102a5b50c in ProcessUtility (pstmt=0x7fdd7400a1c0, queryString=0x7fdd74009240 "CREATE INDEX CONCURRENTLY on prange1_2007_local(time);", context=PROCESS_UTILITY_TOPLEVEL, params=0x0, queryEnv=0x0, dest=0x7fdd7400a2b8, completionTag=0x7fff5d65e3d0 "") at utility.c:353
#11 0x0000000102a5ae08 in PortalRunUtility (portal=0x7fdd73044440, pstmt=0x7fdd7400a1c0, isTopLevel=1 '\001', setHoldSnapshot=0 '\0', dest=0x7fdd7400a2b8, completionTag=0x7fff5d65e3d0 "") at pquery.c:1179
#12 0x0000000102a5a016 in PortalRunMulti (portal=0x7fdd73044440, isTopLevel=1 '\001', setHoldSnapshot=0 '\0', dest=0x7fdd7400a2b8, altdest=0x7fdd7400a2b8, completionTag=0x7fff5d65e3d0 "") at pquery.c:1325
#13 0x0000000102a59644 in PortalRun (portal=0x7fdd73044440, count=9223372036854775807, isTopLevel=1 '\001', run_once=1 '\001', dest=0x7fdd7400a2b8, altdest=0x7fdd7400a2b8, completionTag=0x7fff5d65e3d0 "") at pquery.c:800
#14 0x0000000102a54ecc in exec_simple_query (query_string=0x7fdd74009240 "CREATE INDEX CONCURRENTLY on prange1_2007_local(time);") at postgres.c:1105
#15 0x0000000102a54091 in PostgresMain (argc=1, argv=0x7fdd74003590, dbname=0x7fdd74003410 "postgres", username=0x7fdd740033e8 "onderkalaci") at postgres.c:4075
#16 0x000000010299c0fe in BackendRun (port=0x7fdd72e01aa0) at postmaster.c:4334
#17 0x000000010299b4f3 in BackendStartup (port=0x7fdd72e01aa0) at postmaster.c:4006
#18 0x000000010299a4f5 in ServerLoop () at postmaster.c:1743
#19 0x0000000102997d00 in PostmasterMain (argc=3, argv=0x7fdd72c03640) at postmaster.c:1351
#20 0x00000001028bb1df in main (argc=3, argv=0x7fdd72c03640) at main.c:228
(gdb) c
Continuing.
```

jasonmp85:
Fixes #1456

`PostProcessUtility` does not correctly handle `CREATE INDEX` commands which lack an explicit index name. It was designed to be called to aid in the more complex processing needed for the never-transactional nature of `CREATE INDEX CONCURRENTLY` in a distributed context where transactions might otherwise be very useful.

Unfortunately, it was being called for _all_ DDL commands, not just distributed ones. Distributed `CREATE INDEX` calls go through our feature checks and those lacking an explicit index name error out earlier, but _local_ `CREATE INDEX` commands do not hit this check.

For that reason, if a `CREATE INDEX` command is local, lacks an explicit name, and has the `CONCURRENTLY` flag, we'd reach `PostProcessUtility`, which crashes when called with a `CREATE INDEX` command lacking an explicit name.

Ensuring that the original intent of that method is respected (i.e. that it is only called for distributed DDL commands) avoids the problem altogether.

This is present in `6.2`, so we should consider backporting it (the change is minimal).

### DIFF
--- a/src/backend/distributed/executor/multi_utility.c
+++ b/src/backend/distributed/executor/multi_utility.c
@@ -395,7 +395,11 @@ multi_ProcessUtility(Node *parsetree,
 	standard_ProcessUtility(parsetree, queryString, context,
 							params, dest, completionTag);
 
-	PostProcessUtility(parsetree);
+	/* don't run post-process code for local commands */
+	if (ddlJobs != NIL)
+	{
+		PostProcessUtility(parsetree);
+	}
 
 	if (commandMustRunAsOwner)
 	{
--- a/src/test/regress/expected/multi_index_statements.out
+++ b/src/test/regress/expected/multi_index_statements.out
@@ -92,6 +92,10 @@ CREATE INDEX IF NOT EXISTS lineitem_orderkey_index on index_test_hash(a);
 NOTICE:  relation "lineitem_orderkey_index" already exists, skipping
 -- Verify that we can create indexes concurrently
 CREATE INDEX CONCURRENTLY lineitem_concurrently_index ON lineitem (l_orderkey);
+-- Verify that no-name local CREATE INDEX CONCURRENTLY works
+CREATE TABLE local_table (id integer, name text);
+CREATE INDEX CONCURRENTLY ON local_table(id);
+DROP TABLE local_table;
 -- Verify that all indexes got created on the master node and one of the workers
 SELECT * FROM pg_indexes WHERE tablename = 'lineitem' or tablename like 'index_test_%' ORDER BY indexname;
  schemaname |    tablename     |             indexname              | tablespace |                                                      indexdef                                                       
--- a/src/test/regress/sql/multi_index_statements.sql
+++ b/src/test/regress/sql/multi_index_statements.sql
@@ -66,6 +66,11 @@ CREATE INDEX IF NOT EXISTS lineitem_orderkey_index on index_test_hash(a);
 -- Verify that we can create indexes concurrently
 CREATE INDEX CONCURRENTLY lineitem_concurrently_index ON lineitem (l_orderkey);
 
+-- Verify that no-name local CREATE INDEX CONCURRENTLY works
+CREATE TABLE local_table (id integer, name text);
+CREATE INDEX CONCURRENTLY ON local_table(id);
+DROP TABLE local_table;
+
 -- Verify that all indexes got created on the master node and one of the workers
 SELECT * FROM pg_indexes WHERE tablename = 'lineitem' or tablename like 'index_test_%' ORDER BY indexname;
 \c - - - :worker_1_port


lithp:
Why didn't testing catch this? This feels like something `check-vanilla` would have caught.

lithp:
To answer my own question, this (and Onder's `create table test (a int, b date) PARTITION BY RANGE(b);`) only crashes if `CREATE EXTENSION` has been called. Presence of the citus library isn't enough.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -395,7 +395,11 @@ multi_ProcessUtility(Node *parsetree,
 	standard_ProcessUtility(parsetree, queryString, context,
 							params, dest, completionTag);
 
-	PostProcessUtility(parsetree);
+	/* don't run post-process code for local commands */
+	if (ddlJobs != NIL)
+	{
+		PostProcessUtility(parsetree);
```
marcocitus: Is there any reason not to move this into the `if (ddlJobs != NIL)` block below?

jasonmp85: Expand the surrounding context and you'll see that this area potentially runs as a different user. Presently that's only in the `COPY` case, I believe, so it isn't important, but it seemed to make sense to have the "local callback" code run as that user as well if other local commands did.

I can move it, but when initially writing it it seemed more "correct" to have all local logic run as the special user. If it's unimportant to you (or whoever wants to review this), no biggie.

jasonmp85:
> this (and Onder's `create table test (a int, b date) PARTITION BY RANGE(b);`) only crashes if `CREATE EXTENSION` has been called. Presence of the citus library isn't enough.

@lithp is this an argument for adding a `CREATE EXTENSION` to the beginning of the `check-vanilla` run? Is that possible without breaking test output?

jasonmp85:
@lithp oh, now I see #1460 was created for this.

anarazel:
On 2017-06-19 11:28:02 -0700, Jason Petersen wrote:
> > this (and Onder's `create table test (a int, b date) PARTITION BY RANGE(b);`) only crashes if `CREATE EXTENSION` has been called. Presence of the citus library isn't enough.
> 
> @lithp is this an argument for adding a `CREATE EXTENSION` to the beginning of the `check-vanilla` run? Is that possible without breaking test output?

It does break it, because of our warnings around CREATE
ROLE/USER/DATABASE.  Should probably add an option to silence those.

- Andres

jasonmp85:
> Please backport at the same time

Will do.

jasonmp85:
Experimenting with putting the (GitHub-project–qualified) SHA of my change and this pull request number directly into the backported commit message for better (?) tracking.

PR #1461, opened 2017-06-21
Add automatically started per-database Maintenance Worker

### LINKED ISSUE #1421: DROP EXTENSION citus fails in SQL_DROP trigger
Set up some dummy SQL_DROP trigger that with a statement that goes through the planner hook:
```
CREATE TABLE public.unrelated_table (x int, y int);

CREATE OR REPLACE FUNCTION cleanup()
 RETURNS event_trigger
 LANGUAGE plpgsql
AS $function$                         
BEGIN
    /* go into citus planner hook */
   DELETE FROM public.unrelated_table;
END;                      
$function$;        
CREATE EVENT TRIGGER cleanup_trigger ON SQL_DROP
EXECUTE PROCEDURE cleanup();
```

Then in new session, `DROP EXTENSION citus` attempts to access `pg_dist_partition` (OID 51219) after it was dropped:
```
DROP EXTENSION citus; 
ERROR:  could not open relation with OID 51219
CONTEXT:  SQL statement "DELETE FROM public.unrelated_table"
PL/pgSQL function cleanup() line 3 at SQL statement
```

However, it succeeds the second time:
```
# DROP EXTENSION citus;
DROP EXTENSION
#
```

anarazel:
This'll be used for deadlock detection, 2pc recovery and other similar tasks.

Note that the two first commits fix various issues with cache maintenance.  If needed I can split them out into a separate PR.

See also #1394 /  #1447.

Fixes #1421

### DIFF
--- a/src/backend/distributed/utils/maintenanced.c
+++ b/src/backend/distributed/utils/maintenanced.c
@@ -1,0 +1,412 @@
+/*-------------------------------------------------------------------------
+ *
+ * maintenanced.c
+ *	  Background worker run for each citus using database in a postgres
+ *    cluster.
+ *
+ * This file provides infrastructure for launching exactly one a background
+ * worker for every database in which citus is used.  That background worker
+ * can then perform work like deadlock detection, prepared transaction
+ * recovery, and cleanup.
+ *
+ * Copyright (c) 2017, Citus Data, Inc.
+ *
+ *-------------------------------------------------------------------------
+ */
+
+#include "postgres.h"
+
+
+#include "miscadmin.h"
+#include "pgstat.h"
+
+#include "access/xact.h"
+#include "libpq/pqsignal.h"
+#include "distributed/maintenanced.h"
+#include "distributed/metadata_cache.h"
+#include "postmaster/bgworker.h"
+#include "storage/ipc.h"
+#include "storage/proc.h"
+#include "storage/latch.h"
+#include "storage/lwlock.h"
+#include "tcop/tcopprot.h"
+
+
+/*
+ * Shared memory data for all maintenance workers.
+ */
+typedef struct MaintenanceDaemonControlData
+{
+	/*
+	 * Lock protecting the shared memory state.  This is to be taken when
+	 * looking up (shared mode) or inserting (exclusive mode) per-database
+	 * data in dbHash.
+	 */
+	int trancheId;
+	LWLockTranche lockTranche;
+	LWLock lock;
+
+	/*
+	 * Hash-table of workers, one entry for each database with citus
+	 * activated.
+	 */
+	HTAB *dbHash;
+} MaintenanceDaemonControlData;
+
+
+/*
+ * Per database worker state.
+ */
+typedef struct MaintenanceDaemonDBData
+{
+	/* hash key: database to run on */
+	Oid databaseOid;
+
+	/* information: which user to use */
+	Oid userOid;
+	bool daemonStarted;
+	Latch *latch; /* pointer to the background worker's latch */
+} MaintenanceDaemonDBData;
+
+
+static shmem_startup_hook_type prev_shmem_startup_hook = NULL;
+static MaintenanceDaemonControlData *MaintenanceDaemonControl = NULL;
+
+static volatile sig_atomic_t got_SIGHUP = false;
+
+static void MaintenanceDaemonSigHupHandler(SIGNAL_ARGS);
+static size_t MaintenanceDaemonShmemSize(void);
+static void MaintenanceDaemonShmemInit(void);
+static void MaintenanceDaemonErrorContext(void *arg);
+
+/*
+ * InitializeMaintenanceDaemon, called at server start, is responsible for
+ * requesting shared memory and related infrastructure required by maintenance
+ * daemons.
+ */
+void
+InitializeMaintenanceDaemon(void)
+{
+	RequestAddinShmemSpace(MaintenanceDaemonShmemSize());
+
+	prev_shmem_startup_hook = shmem_startup_hook;
+	shmem_startup_hook = MaintenanceDaemonShmemInit;
+}
+
+
+/*
+ * InitializeMaintenanceDaemonBackend, called at backend start and
+ * configuration changes, is responsible for starting a per-database
+ * maintenance worker if necessary.
+ */
+void
+InitializeMaintenanceDaemonBackend(void)
+{
+	MaintenanceDaemonDBData *dbData = NULL;
+	Oid extensionOwner = CitusExtensionOwner();
+	bool found;
+
+	LWLockAcquire(&MaintenanceDaemonControl->lock, LW_EXCLUSIVE);
+
+	dbData = (MaintenanceDaemonDBData *) hash_search(MaintenanceDaemonControl->dbHash,
+													 &MyDatabaseId,
+													 HASH_ENTER_NULL, &found);
+
+	if (dbData == NULL)
+	{
+		/* FIXME: better message, reference relevant guc in hint */
+		ereport(ERROR, (errmsg("ran out of database slots")));
+	}
+
+	if (!found || !dbData->daemonStarted)
+	{
+		BackgroundWorker worker;
+		BackgroundWorkerHandle *handle = NULL;
+		int pid = 0;
+
+		dbData->userOid = extensionOwner;
+
+		memset(&worker, 0, sizeof(worker));
+
+		snprintf(worker.bgw_name, BGW_MAXLEN,
+				 "Citus Maintenance Daemon: %u/%u",
+				 MyDatabaseId, extensionOwner);
+
+		/* request ability to connect to target database */
+		worker.bgw_flags = BGWORKER_SHMEM_ACCESS | BGWORKER_BACKEND_DATABASE_CONNECTION;
+
+		/*
+		 * No point in getting started before able to run query, but we do
+		 * want to get started on Hot-Stanby standbys.
+		 */
+		worker.bgw_start_time = BgWorkerStart_ConsistentState;
+
+		/*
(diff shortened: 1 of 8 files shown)


lithp:
Neat, the cache changes are nice to finally have!

Why did you add a worker-per-database rather than have a global worker?

marcocitus:
> Why did you add a worker-per-database rather than have a global worker?

A background worker can only interact with one database.

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -2221,6 +2253,20 @@ GetLocalGroupId(void)
 
 
 /*
+ * RegisterLocalGroupIdCacheCallbacks registers the callbacks required for the
+ * worker node cache.  It's separate from InitializeWorkerNodeCache so the
+ * callback can be registered early, before the tables exist.
```
marcocitus: Comment is copied from `RegisterWorkerNodeCacheCallbacks` and needs to be adapted.

src/include/distributed/shared_library_init.h
```diff
@@ -0,0 +1,6 @@
+#ifndef SHARED_LIBRARY_INIT_H
+#define SHARED_LIBRARY_INIT_H
```
marcocitus: I wish we were using `#pragma once`

src/backend/distributed/utils/maintenanced.c
```diff
@@ -0,0 +1,380 @@
...
+	{
+		/* FIXME: better message, reference relevant guc in hint */
+		ereport(ERROR, (errmsg("ran out of database slots")));
+	}
+
+	if (!found || !dbData->daemonStarted)
+	{
+		BackgroundWorker worker;
+		BackgroundWorkerHandle *handle;
+		int pid;
```
marcocitus: Should have `= NULL` and `= 0` per convention

src/include/distributed/maintenanced.h
```diff
@@ -0,0 +1,18 @@
+/*-------------------------------------------------------------------------
+ *
+ * maintenanced.h
+ *
+ * Copyright (c) 2017, Citus Data, Inc.
+ *
+ *-------------------------------------------------------------------------
```
marcocitus: Should also be in `shared_library_init.h`, and maybe have some meaningful comment.

src/backend/distributed/utils/maintenanced.c
```diff
@@ -0,0 +1,380 @@
...
+	HASHCTL hashInfo;
+	int hashFlags = 0;
+
+	LWLockAcquire(AddinShmemInitLock, LW_EXCLUSIVE);
+
+	MaintenanceDaemonControl =
+		(MaintenanceDaemonControlData *) ShmemInitStruct("Citus Maintenance Daemon",
+														 MaintenanceDaemonShmemSize(),
+														 &alreadyInitialized);
+	if (!alreadyInitialized)
```
marcocitus: When is this already initialised?

marcocitus:
Currently a `pg_ctl restart` gives the following log message:
```
LOG:  worker process: Citus Maintenance Daemon: 13322 (PID 1547) exited with exit code 1
```

anarazel:
> Currently a pg_ctl restart gives the following log message:
> LOG:  worker process: Citus Maintenance Daemon: 13322 (PID 1547) exited with exit code 1

Right - that's to be expected?

src/include/distributed/maintenanced.h
anarazel: Not sure what you mean here? Remove maintenanced.h?

src/backend/distributed/utils/maintenanced.c
anarazel: On EXEC_BACKEND style builds (primarily windows), and crash restarts.

marcocitus:
> Right - that's to be expected?

Why not exit cleanly?

src/backend/distributed/utils/maintenanced.c
marcocitus: Might be worth adding a comment

src/include/distributed/maintenanced.h
marcocitus: sorry, put 2 comments in 1.

1: maintenance.d should have a meaningful comment at the top
2: shared_library_init.h should have a copyright notice.

anarazel:
> Why not exit cleanly?

Two reasons: exit(1) is the "traditional" response to SIGTERM in all other processes. Secondly, exit(0) unregisters a bgworker, i.e. it'll not be restarted anymore, which is not what we'd want to do as a general response to SIGTERM.

marcocitus:
> exit(1) is the "traditional" response to SIGTERM in all other processes. 

Except postgres background processes, which all exit(0).

> Secondly, exit(0) unregisters a bgworker, Secondly, exit(0) unregisters a bgworker, i.e. it'll not be restarted anymore, which is not what we'd want to do as a general response to SIGTERM.

Why not stop on SIGTERM and allow restart via `InitializeMaintenanceDaemonBackend`?

anarazel:
> Except postgres background processes, which all exit(0).

Well there postmaster has special logic to handle exits of such processes, so that's not applicable here.  The applicable comparison is to bgworkers and plain backends.

> >  Secondly, exit(0) unregisters a bgworker, i.e. it'll not be restarted anymore, which is not what we'd want to do as a general response to SIGTERM.

> Why not stop on SIGTERM and allow restart via InitializeMaintenanceDaemonBackend?

Because then it'd not be started unless new connections are established?  And even then there'd be a race condition around the maintenanced quitting and new connections being established.   Just seems to add complications for no gain.

marcocitus:
> Because then it'd not be started unless new connections are established? 

Which is the case anyway.

I'm not concerned about convention here. It's just a bad and confusing user experience as it's indistinguishable from a crash.

anarazel:
>  > Because then it'd not be started unless new connections are established?
> Which is the case anyway.

Huh? No, it gets restarted automatically after 5 secs.


> I'm not concerned about convention here. It's just a bad and confusing user experience as it's indistinguishable from a crash.

exit(1) is never a crash?

marcocitus:
> Huh? No, it gets restarted automatically after 5 secs.

When postgres starts then "it'd not be started unless new connections are established".

> exit(1) is never a crash?

I'm calling an error that causes the background worker to have to restart a crash here.

anarazel:
I'm really not folllowing.  Right now maintenanced gets restarted both after restarts due to manual SIGTERM and due to ERROR/FATALs it raises internally.  That seems to be the right behaviour for something kind of critical functionality-wise.   You seem to be advocating that we change either or both of these to not-restart and instead rely on maintenanced being restarted when there's a new connection to a citus using database?

marcocitus:
This also fixes #1421

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -81,27 +82,39 @@ typedef struct ShardCacheEntry
...
+	Oid distShardPlacementNodeidIndexId;
+	Oid distTransactionRelationId;
+	Oid distTransactionGroupIndexId;
+	Oid extraDataContainerFuncId;
+	Oid workerHashFunctionId;
+	Oid extensionOwner;
+};
+
+
+static struct CachedData CachedData;
```
anarazel: Done (as of next push).

src/backend/distributed/utils/metadata_cache.c
anarazel: Done (as of next push).

src/include/distributed/shared_library_init.h
anarazel: That's a holy war I'm not going to get involved into on either side ;)

src/backend/distributed/utils/maintenanced.c
anarazel: Done (as of next push).

src/backend/distributed/utils/maintenanced.c
anarazel: Done (as of next push).

src/include/distributed/maintenanced.h
anarazel: Done (as of next push).

anarazel:
> This also fixes #1421

Heh, nice accident.

anarazel:
I've updated the patch to address the points you had, including adding an error context.  It's right now just saying
```
	errcontext("Citus maintenance daemon for database %u user %u",
			   myDbData->databaseOid, myDbData->userOid);
```

src/backend/distributed/utils/maintenanced.c
```diff
@@ -0,0 +1,412 @@
...
+		BackgroundWorkerHandle *handle = NULL;
+		int pid = 0;
+
+		dbData->userOid = extensionOwner;
+
+		snprintf(worker.bgw_name, BGW_MAXLEN,
+				 "Citus Maintenance Daemon: %u/%u",
+				 MyDatabaseId, extensionOwner);
+
+		memset(&worker, 0, sizeof(worker));
```
marcocitus: This needs to be above `snprintf(worker.bgw_name, ...`.

marcocitus:
I'm just going to push a fix for that `multi_reference_table` test that keeps failing on travis to master.

src/backend/distributed/utils/maintenanced.c
anarazel: That, uh, is a very fair complaint.  WTH?

anarazel:
Thanks for the review!

anarazel:
Thanks for the review!

src/include/distributed/shared_library_init.h
jasonmp85: Is there anyone on the other side?

PR #1462, opened 2017-06-21
Remove unused line, @arguments was set but never used

lithp:
Assigning @anarazel because you wrote the line I'm removing. Did you intend to use it?

### DIFF
--- a/src/test/regress/pg_regress_multi.pl
+++ b/src/test/regress/pg_regress_multi.pl
@@ -454,7 +454,6 @@ my $startTime = time();
 # Finally run the tests
 if ($vanillatest)
 {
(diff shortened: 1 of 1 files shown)


src/test/regress/pg_regress_multi.pl
```diff
@@ -462,7 +462,6 @@ END
 # Finally run the tests
 if ($vanillatest)
 {
-    push(@arguments, "--schedule=${postgresSrcdir}/parallel_schedule");
```
marcocitus: I'm guessing the intention here was to directly start pg_regress below, which would allow us to add `--load-extension=citus`.

PR #1463, opened 2017-06-21
Add some utility functions for partitioned tables

onderkalaci:
This commit is intended to be a base for supporting declarative partitioning
on distributed tables. Here we add the following utility functions and their
unit tests:

  * Very basic functions including differnentiating partitioned tables and
    partitions, listing the partitions
  * Generating the PARTITION BY (expr) and adding this to the DDL events
    of partitioned tables
  * Ability to generate text representations of the ranges for partitions
  * Ability to generate the `ALTER TABLE parent_table ATTACH PARTITION
    partition_table FOR VALUES value_range`
  * Ability to apply add shard ids to the above command using
    `worker_apply_inter_shard_ddl_command()`
  * Ability to generate `ALTER TABLE parent_table DETACH PARTITION`


src/backend/distributed/test/partitioning_utils.c
```diff
@@ -0,0 +1,97 @@
...
+ */
+Datum
+generate_alter_table_attach_partition_command(PG_FUNCTION_ARGS)
+{
+	char *command = NULL;
+
+#if (PG_VERSION_NUM >= 100000)
+	command = GenerateAlterTableAttachPartitionCommand(PG_GETARG_OID(0));
+# else
+	command = "";
```
marcocitus: I guess we could just do `command = ""` at the start and omit the `# else`.

src/backend/distributed/test/partitioning_utils.c
```diff
@@ -0,0 +1,97 @@
...
+		appendStringInfoString(resultRelationNames, get_rel_name(partitionOid));
+
+		if (!firstPartition)
+		{
+			appendStringInfoString(resultRelationNames, ",");
+		}
+		firstPartition = true;
+	}
+# else
+	appendStringInfoString(resultRelationNames, "");
```
marcocitus: This seems redundant.

src/backend/distributed/test/partitioning_utils.c
```diff
@@ -0,0 +1,97 @@
...
+	{
+		Oid partitionOid = lfirst_oid(partitionOidCell);
+
+		appendStringInfoString(resultRelationNames, get_rel_name(partitionOid));
+
+		if (!firstPartition)
+		{
+			appendStringInfoString(resultRelationNames, ",");
+		}
+		firstPartition = true;
```
marcocitus: Maybe move into else, or use `if (resultRelationNames->len > 0)`

src/backend/distributed/utils/citus_ruleutils.c
```diff
@@ -302,14 +303,24 @@ pg_get_tableschemadef_string(Oid tableRelationId, bool includeSequenceDefaults)
 	relationName = generate_relation_name(tableRelationId, NIL);
 
 	relationKind = relation->rd_rel->relkind;
-	if (relationKind != RELKIND_RELATION && relationKind != RELKIND_FOREIGN_TABLE)
+#if (PG_VERSION_NUM >= 100000)
+	if (!(relationKind == RELKIND_RELATION || relationKind == RELKIND_FOREIGN_TABLE ||
+		  relationKind == RELKIND_PARTITIONED_TABLE))
+#else
+	if (!(relationKind == RELKIND_RELATION || relationKind == RELKIND_FOREIGN_TABLE))
+#endif
```
marcocitus: This looks a bit messy.

Not sure, maybe something like:

```
isTable = relationKind == RELKIND_RELATION || relationKind == RELKIND_FOREIGN_TABLE;

isTable = isTable || relationKind == RELKIND_PARTITIONED_TABLE;
#endif

if (!isRelation)
{
    ...
```

src/backend/distributed/utils/citus_ruleutils.c
```diff
@@ -302,14 +303,24 @@ pg_get_tableschemadef_string(Oid tableRelationId, bool includeSequenceDefaults)
...
 	{
 		ereport(ERROR, (errcode(ERRCODE_WRONG_OBJECT_TYPE),
-						errmsg("%s is not a regular or foreign table", relationName)));
+						errmsg("%s is not a regular or foreign or partitioned table",
+							   relationName)));
```
marcocitus: `regular, foreign, or partitioned`

src/backend/distributed/utils/citus_ruleutils.c
```diff
@@ -337,7 +348,15 @@ pg_get_tableschemadef_string(Oid tableRelationId, bool includeSequenceDefaults)
 	{
 		Form_pg_attribute attributeForm = tupleDescriptor->attrs[attributeIndex];
 
-		if (!attributeForm->attisdropped && attributeForm->attinhcount == 0)
+		/*
+		 * We disregard the inhereted attributes (i.e., attinhcount > 0) here. The
+		 * reasoning behind this is that Citus implements declarative partitioning
+		 * by creating the partitions first and then sending
+		 * "ALTER TABLE parent_table ATTACH PARTITION .." command. This may not play
+		 * well with regular inhereted tables, which isn't a big concern from Citus'
```
marcocitus: inhereted -> inherited

src/backend/distributed/utils/citus_ruleutils.c
```diff
@@ -337,7 +348,15 @@ pg_get_tableschemadef_string(Oid tableRelationId, bool includeSequenceDefaults)
 	{
 		Form_pg_attribute attributeForm = tupleDescriptor->attrs[attributeIndex];
 
-		if (!attributeForm->attisdropped && attributeForm->attinhcount == 0)
```
marcocitus: Do we error out somewhere on inherited tables that are not partitioned tables?

src/test/regress/expected/multi_partitioning_utils.out
```diff
@@ -0,0 +1,338 @@
...
+DECLARE
+	attach_partition_command text := '';
+	command_result text := '';
+
+BEGIN
+   -- first generate the command
+   SELECT public.generate_alter_table_attach_partition_command(partition_name) INTO attach_partition_command;
+
+   -- later detach the same partition
+   EXECUTE 'ALTER TABLE ' || parent_table_name || ' DETACH PARTITION ' || partition_name || ';';
```
marcocitus: Would it make sense to also write a C function for this? since we will want to be able to detach as well?

src/test/regress/multi_schedule
```diff
@@ -22,6 +22,11 @@ test: multi_table_ddl
 test: multi_name_lengths
 test: multi_metadata_access
 
+# ---
+# Tests for partitioning support
+# ---
+test: multi_partitioning_utils
```
marcocitus: Can we parallelise this with something else?

src/test/regress/sql/multi_partitioning_utils.sql
```diff
@@ -0,0 +1,218 @@
+ALTER SEQUENCE pg_catalog.pg_dist_shardid_seq RESTART 1620000;
+ALTER SEQUENCE pg_catalog.pg_dist_jobid_seq RESTART 1620000;
```
marcocitus: Let's not set these if we don't need them. We don't even use `pg_dist_jobid_seq` anymore.

src/test/regress/sql/multi_partitioning_utils.sql
```diff
@@ -0,0 +1,218 @@
...
+CREATE TABLE list_partitioned (col1 NUMERIC, col2 NUMERIC, col3 VARCHAR(10)) PARTITION BY LIST (col1) ;
+
+SELECT generate_partition_information('list_partitioned');
+SELECT master_get_table_ddl_events('list_partitioned');
+SELECT drop_and_recreate_partitioned_table('list_partitioned');
+
+CREATE TABLE list_partitioned_1 PARTITION OF list_partitioned FOR VALUES IN (100, 101, 102, 103, 104);
+
+SELECT generate_alter_table_attach_partition_command('list_partitioned_1');
+
```
marcocitus: Probably subjective, but I kind of like when we DROP tables at the end.

src/test/regress/sql/multi_partitioning_utils.sql
```diff
@@ -0,0 +1,218 @@
...
+	command text := '';
+
+BEGIN
+   -- first generate the command
+   CREATE TABLE partitioned_table_create_commands AS SELECT master_get_table_ddl_events(parent_table_name::text);
+
+   -- later detach the same partition
+   EXECUTE 'DROP TABLE ' || parent_table_name::text || ';';
+
+   FOR command IN SELECT * FROM partitioned_table_create_commands
```
marcocitus: Could just do `SELECT master_get_table_ddl_events(parent_table_name::text)` in here.

src/test/regress/expected/multi_partitioning_utils_0.out
```diff
@@ -0,0 +1,291 @@
+ALTER SEQUENCE pg_catalog.pg_dist_shardid_seq RESTART 1620000;
```
marcocitus: Could we add the postgres version (see `multi_explain.sql`):
```
-- print major version to make version-specific tests clear
SELECT substring(version(), '\d+(?:\.\d+)?') AS major_version;
 major_version
---------------
 9.6
(1 row)
```

src/backend/distributed/utils/citus_ruleutils.c
onderkalaci: Done

src/test/regress/sql/multi_partitioning_utils.sql
onderkalaci: Added

src/backend/distributed/utils/citus_ruleutils.c
onderkalaci: No, we don't. Currently what happens is that we allow creating those tables, but, the inheritance hierarchy is not created on the workers. Do you prefer erroring out for such cases?

src/test/regress/sql/multi_partitioning_utils.sql
marcocitus: Ah, makes sense. Probably makes more sense to use an array, but this is also fine.

src/backend/distributed/utils/citus_ruleutils.c
marcocitus: I think it would be good to avoid confusing anyone when we announce partitioning support.

onderkalaci: Added explicit checks for inherited tables.

src/backend/distributed/utils/multi_partitioning_utils.c
```diff
@@ -0,0 +1,314 @@
...
+	return partitionList;
+}
+
+
+/*
+ * GenereateDetachPartitionCommand gets a partition table and returns
+ * "ALTER TABLE parent_table DETACH PARTITION partitionName" command.
+ */
+char *
+GenereateDetachPartitionCommand(Oid partitionTableId)
```
marcocitus: Genereate -> Generate

src/backend/distributed/utils/multi_partitioning_utils.c
```diff
@@ -0,0 +1,314 @@
...
+{
+	Relation pgInherits = NULL;
+	SysScanDesc scan = NULL;
+	ScanKeyData key[1];
+	HeapTuple inheritsTuple = NULL;
+	bool tableInherits = false;
+
+	pgInherits = heap_open(InheritsRelationId, AccessShareLock);
+
+	ScanKeyInit(&key[0], Anum_pg_inherits_inhrelid,
```
marcocitus: If you search for `Anum_pg_inherits_inhparent` like `find_inheritance_children` does. You can use the index and only need to check the existence of a single record (e.g. `if (systable_getnext(scan) != NULL) tableInherits = true`)

onderkalaci: But, that function looks for `Anum_pg_inherits_inhparent` and we're looking for `Anum_pg_inherits_inhrelid` and the index is on `Anum_pg_inherits_inhparent`.

Also, since this is a function which is supposed to be called very infrequently I preferred to keep it simpler and not tried to utilize the index.

src/backend/distributed/utils/multi_partitioning_utils.c
onderkalaci: Fixed

src/backend/distributed/utils/multi_partitioning_utils.c
marcocitus: Ok, I get a little confused by the name. Could we call it `IsChildTable`? Should we also check for `IsParentTable`? We can't really distribute either.

onderkalaci: > Ok, I get a little confused by the name. Could we call it IsChildTable?

Hmm, that's probably a better name. I'll update it.

>  Should we also check for IsParentTable? We can't really distribute either.

Yes, that's a good idea.

src/backend/distributed/utils/citus_ruleutils.c
```diff
@@ -459,14 +477,43 @@ pg_get_tableschemadef_string(Oid tableRelationId, bool includeSequenceDefaults)
...
+{
+	char relationKind = relation->rd_rel->relkind;
+	bool supportedRelationKind = (relationKind == RELKIND_RELATION || relationKind ==
+								  RELKIND_FOREIGN_TABLE);
+#if (PG_VERSION_NUM >= 100000)
+	supportedRelationKind = supportedRelationKind || relationKind ==
+							RELKIND_PARTITIONED_TABLE;
+#endif
+
+	/* Citus doesn't support bare inhereted tables (i.e., not a partition or partitioned table) */
```
marcocitus: inhereted -> inherited

src/backend/distributed/utils/citus_ruleutils.c
```diff
@@ -494,10 +542,18 @@ pg_get_tablecolumnoptionsdef_string(Oid tableRelationId)
...
+#if (PG_VERSION_NUM >= 100000)
+	supportedRelationKind = supportedRelationKind || relationKind ==
+							RELKIND_PARTITIONED_TABLE;
+#endif
+
+	if (!supportedRelationKind)
 	{
 		ereport(ERROR, (errcode(ERRCODE_WRONG_OBJECT_TYPE),
-						errmsg("%s is not a regular or foreign table", relationName)));
+						errmsg("%s is not a regular or foreign table or partitioned",
```
marcocitus: Might make sense to error out a little earlier in `create_distributed_table`, but seems ok for now.

src/backend/distributed/utils/multi_partitioning_utils.c
```diff
@@ -0,0 +1,362 @@
...
+	pgInherits = heap_open(InheritsRelationId, AccessShareLock);
+
+	ScanKeyInit(&key[0], Anum_pg_inherits_inhparent,
+				BTEqualStrategyNumber, F_OIDEQ,
+				ObjectIdGetDatum(relationId));
+
+	scan = systable_beginscan(pgInherits, InheritsParentIndexId, true,
+							  NULL, 1, key);
+
+	while ((inheritsTuple = systable_getnext(scan)) != NULL)
```
marcocitus: Doing `if (systable_getnext(scan) != NULL) tableInherited = true)` should be sufficient here.

src/backend/distributed/utils/multi_partitioning_utils.c
```diff
@@ -0,0 +1,362 @@
...
+	pgInherits = heap_open(InheritsRelationId, AccessShareLock);
+
+	ScanKeyInit(&key[0], Anum_pg_inherits_inhrelid,
+				BTEqualStrategyNumber, F_OIDEQ,
+				ObjectIdGetDatum(relationId));
+
+	scan = systable_beginscan(pgInherits, InvalidOid, false,
+							  NULL, 1, key);
+
+	while ((inheritsTuple = systable_getnext(scan)) != NULL)
```
marcocitus: Doing `if (systable_getnext(scan) != NULL) tableInherited = true)` should be sufficient here.

The scan key already makes sure we only get tuples where the value in the `Anum_pg_inherits_inhrelid` column is `relationId`.

src/test/regress/multi_schedule
marcocitus: I could live with it, but the more we can put together the better.

src/backend/distributed/utils/multi_partitioning_utils.c
onderkalaci: Makes sense

PR #1465, opened 2017-06-21
[WIP] Add support for UPDATE ... FROM

### LINKED ISSUE #1127: UPDATE ... FROM support for queries that touch a single machine
We have two customers who asked if we could support the ``UPDATE ... FROM (SELECT ...)`` construct for queries that modify data in a single machine. For example:

<pre>
UPDATE foobar
SET
       distinct_viewers = metrics.viewers
FROM
       (SELECT contactable_id foobar_id,
        count(DISTINCT (CASE WHEN event_type='View' then visitor_token end)) as viewers
        FROM events
        WHERE account_id=12 group by contactable_id) metrics
WHERE
   id=metrics.foobar_id AND
   account_id=12;
</pre>

metdos:
This is one way to implement this, and possibly not the best one. I will first add more regression tests and see limits of this approach and later will revisit the approach itself.

Fixes #1127

### DIFF
--- a/src/backend/distributed/planner/multi_router_planner.c
+++ b/src/backend/distributed/planner/multi_router_planner.c
@@ -202,6 +202,7 @@ CreateSingleTaskRouterPlan(Query *originalQuery, Query *query,
 	Task *task = NULL;
 	List *placementList = NIL;
 	MultiPlan *multiPlan = CitusMakeNode(MultiPlan);
+	bool updateFromQuery = UpdateFromQuery(query);
 
 	multiPlan->operation = query->commandType;
 
@@ -211,7 +212,20 @@ CreateSingleTaskRouterPlan(Query *originalQuery, Query *query,
 		modifyTask = true;
 	}
 
-	if (modifyTask)
+	if (updateFromQuery)
+	{
+		DeferredErrorMessage *planningError = NULL;
+		planningError = ModifyQuerySupported(query);
+
+		if (planningError != NULL)
+		{
+			multiPlan->planningError = planningError;
+			return multiPlan;
+		}
+
+		task = RouterSelectTask(originalQuery, restrictionContext, &placementList);
+	}
+	else if (modifyTask)
 	{
 		Oid distributedTableId = ExtractFirstDistributedTableId(originalQuery);
 		ShardInterval *targetShardInterval = NULL;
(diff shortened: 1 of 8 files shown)


src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -211,7 +212,20 @@ CreateSingleTaskRouterPlan(Query *originalQuery, Query *query,
...
+		DeferredErrorMessage *planningError = NULL;
+		planningError = ModifyQuerySupported(query);
+
+		if (planningError != NULL)
+		{
+			multiPlan->planningError = planningError;
+			return multiPlan;
+		}
+
+		task = RouterSelectTask(originalQuery, restrictionContext, &placementList);
```
marcocitus: Is it possible to send all UPDATE (and maybe DELETE) logic through this function? (which should probably be renamed)

metdos: I think so, just gonna need to test a bit more.

marcocitus:
Apart from the deparsing bit, is there any notable difference between DELETE and UPDATE?

src/backend/distributed/planner/multi_router_planner.c
metdos: Hey @marcocitus, it turned out to be a bit more complex to send all UPDATE (and maybe DELETE) logic through RouterSelectTask. It seems possible, I did some part in a hacky way on [this branch ](https://github.com/citusdata/citus/commits/update_with_subquery). The problem is that other parts of code are highly coupled with effects of `RouterModifyTask()`, so using `RouterSelectTask()` requires more refactoring, also testing, than my initial thought.

Some of the considerations are;
- Support of prepared statements.
- The error message on unsupported cases - `RouterSelectTask()` needs refactoring to support the previous granularity.
- Evaluation of stable functions on the master.

Given these, I'm inclined to support UPDATE with subqueries as a special case, and use `RouterSelectTask()` just for them. What do you think?

metdos:
Hey @marcocitus, it turned out to be a bit more complex to send all UPDATE (and maybe DELETE) logic through RouterSelectTask. It seems possible, I did some part in a hacky way on [this branch ](https://github.com/citusdata/citus/commits/update_with_subquery). The problem is that other parts of code are highly coupled with effects of `RouterModifyTask()`, so using `RouterSelectTask()` requires more refactoring, also testing, than my initial thought.

Some of the considerations are;
- Support of prepared statements.
- The error message on unsupported cases - `RouterSelectTask()` needs refactoring to support the previous granularity.
- Evaluation of stable functions on the master.

Given these, I'm inclined to support UPDATE with subqueries as a special case, and use `RouterSelectTask()` just for them. What do you think?

marcocitus:
> Given these, I'm inclined to support UPDATE with subqueries as a special case, and use RouterSelectTask() just for them. What do you think?

I think that's likely to lead to bugs and unexpected behaviour. We will want prepared statement support and function evaluation to be consistent for all UPDATEs.

A few weeks back I worked on refactoring the UPDATE planning logic to make zero-task UPDATEs possible, and to open up the path to multi-task UPDATEs (and multi-row INSERTs) as well. I kind of got stuck on 0-task SELECTs, which is what I was originally trying to solve in order to simplify https://github.com/citusdata/citus/pull/1455 and decided not to spend more time on it.

However, the change also includes some simplifications for prepared statements and function evaluation that are likely to be useful for this PR. 

Do you think this would help resolve the issue?
https://github.com/citusdata/citus/pull/1474

metdos:
Closing this in favor of #1483.

Issue #1459, opened 2017-06-19
Support declarative partitioning on distributed tables


onderkalaci:
Postgres 10 will come with declarative partitioning [support](https://www.postgresql.org/docs/10/static/ddl-partitioning.html#ddl-partitioning-declarative).

To achieve #573, one of the alternatives is to support declarative partitioning on distributed tables. (See the other alternative [here](https://github.com/citusdata/citus/issues/573?_pjax=%23js-repo-pjax-container#issuecomment-303075836).) Note that this will build the infrastructure to fulfill the requirements of #573.

We've had some internal discussions and design notes on this [document](https://docs.google.com/document/d/11TsIbS60oUiPKKt6mj3JWxwyjPW_S9iF_Rj1778LhRc/edit). Here, I'll move the necessary parts:

- [x] Add utility functions that the other parts of the implementation will benefit
- [x]  Creating partitioned distributed tables:
    - [x]  Creating the partitioning hierarchy before distributing the relations and then distributing the relations
   - [x] Creating partitions of already distributed relations
   - [x] Creating regular PostgreSQL tables and ATTACHING to a distributed table
   - [x] Creating distributed tables and ATTACHING to another distributed table
   - [x] DETACHING partitions from a distributed table
   - [x] Erroring out on unsupported operations
- [x] Metadata: Make sure that each partition is a distributed table and the partitioning hierarchy is formed on the shards as well

- [x] Locking on the coordinator: Given that both partitions and parent table is accessible by the users, it makes sense to elaborate lock acquisitions on the master to make sure that we’re good. See below for the discussion:

    - [x]  Shard metadata locks: Make sure that parent and children are co-located. Then, co-location logic would handle the rest for us.

    - [x]  Shard resource locks: Whenever a child's lock is acquired, make sure that the parent's corresponding shard resource lock is also acquired.

    - [x]  Relation Lock on query execution: This is tricky. Whenever a modification happens on a parent table, Postgres acquires relation lock (i.e., AccessShareLock) on both the parent and all children. However, that is happening during the standard_executor(), which Citus skips. The relevant locks are already held on the shards by Postgres. However, the question is that do we actually need to acquire the relation locks on the coordinator?

    - [x]  Relation Locks on DDLs: Since Citus first executes the DDLs on the coordinator, all the necessary locks are acquired. Thus, we don't need to implement anything.

- [x] Shard rebalancer considers the partitioning hierarchy. 
- [x] Tenant isolation considers the partitioning hierarchy.
[...]

PR #1467, opened 2017-06-22
Creating distributed partitioned tables [WIP]

onderkalaci:
Currently only the following is supported:
```SQL
CREATE TABLE prange1(id int, time date) PARTITION BY RANGE (time);
 
-- create its partitions
CREATE TABLE prange1_2006 PARTITION OF prange1 FOR VALUES FROM ('2006-01-01') TO ('2007-01-01');
CREATE TABLE prange1_2007 PARTITION OF prange1 FOR VALUES FROM ('2007-01-01') TO ('2008-01-01');

SELECT create_distributed_table('prange1', 'id');

-- partitions can be created later on as well
CREATE TABLE prange1_2008 PARTITION OF prange1 FOR VALUES FROM ('2008-01-01') TO ('2009-01-01');

-- also regular tables can be attached to parent tables
CREATE TABLE prange1_2009(id int, time date);
ALTER TABLE prange1 ATTACH PARTITION prange1_2009 FOR VALUES FROM ('2009-01-01') TO ('2010-01-01');


```

Other ways of creating distributed tables mentioned on #1459 will be added to this PR.

### DIFF
--- a/src/backend/distributed/commands/create_distributed_table.c
+++ b/src/backend/distributed/commands/create_distributed_table.c
@@ -78,9 +78,6 @@ static char LookupDistributionMethod(Oid distributionMethodOid);
 static Oid SupportFunctionForColumn(Var *partitionColumn, Oid accessMethodId,
 									int16 supportFunctionNumber);
 static bool LocalTableEmpty(Oid tableId);
-static void CreateHashDistributedTable(Oid relationId, char *distributionColumnName,
-									   char *colocateWithTableName,
-									   int shardCount, int replicationFactor);
 static Oid ColumnType(Oid relationId, char *columnName);
 static void CopyLocalDataIntoShards(Oid destinationDistributedRelationId, List *
 									sourceLocalRelationList);
@@ -618,7 +615,7 @@ CreateTruncateTrigger(Oid relationId)
 /*
  * CreateHashDistributedTable creates a hash distributed table.
  */
-static void
+void
 CreateHashDistributedTable(Oid relationId, char *distributionColumnName,
 						   char *colocateWithTableName, int shardCount,
 						   int replicationFactor)
--- a/src/backend/distributed/commands/create_distributed_table.c
+++ b/src/backend/distributed/commands/create_distributed_table.c
@@ -29,6 +29,7 @@
 #include "commands/defrem.h"
 #include "commands/extension.h"
 #include "commands/trigger.h"
+#include "distributed/citus_ruleutils.h"
 #include "distributed/colocation_utils.h"
 #include "distributed/distribution_column.h"
 #include "distributed/master_metadata_utility.h"
@@ -38,6 +39,7 @@
 #include "distributed/multi_copy.h"
 #include "distributed/multi_logical_planner.h"
 #include "distributed/multi_utility.h"
+#include "distributed/multi_partitioning_utils.h"
 #include "distributed/pg_dist_colocation.h"
 #include "distributed/pg_dist_partition.h"
 #include "distributed/reference_table_utils.h"
@@ -80,7 +82,8 @@ static void CreateHashDistributedTable(Oid relationId, char *distributionColumnN
 									   char *colocateWithTableName,
 									   int shardCount, int replicationFactor);
 static Oid ColumnType(Oid relationId, char *columnName);
-static void CopyLocalDataIntoShards(Oid relationId);
+static void CopyLocalDataIntoShards(Oid destinationDistributedRelationId, List *
+									sourceLocalRelationList);
 static List * TupleDescColumnNameList(TupleDesc tupleDescriptor);
 #if (PG_VERSION_NUM >= 100000)
 static bool RelationUsesIdentityColumns(TupleDesc relationDesc);
@@ -280,7 +283,7 @@ CreateReferenceTable(Oid relationId)
 	/* copy over data for regular relations */
 	if (relationKind == RELKIND_RELATION)
 	{
-		CopyLocalDataIntoShards(relationId);
+		CopyLocalDataIntoShards(relationId, list_make1_oid(relationId));
 	}
 }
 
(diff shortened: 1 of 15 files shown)


src/backend/distributed/commands/create_distributed_table.c
```diff
@@ -810,53 +899,66 @@ CopyLocalDataIntoShards(Oid distributedRelationId)
...
 	copyDest->rStartup(copyDest, 0, tupleDescriptor);
 
-	/* begin reading from local table */
-	scan = heap_beginscan(distributedRelation, GetActiveSnapshot(), 0, NULL);
-
-	oldContext = MemoryContextSwitchTo(GetPerTupleMemoryContext(estate));
-
-	while ((tuple = heap_getnext(scan, ForwardScanDirection)) != NULL)
+	/* iterate on the local relations, copy each into the destination relation */
+	foreach(sourceRelationCell, sourceRelationList)
```
onderkalaci: I tried that first, but hit an error on the connection management (as far as I remember it was the following error):
```
SELECT create_distributed_table('partitioning_test', 'id');
NOTICE:  Copying data from local table...
NOTICE:  Copying data from local table...
ERROR:  cannot establish new placement connection when DML has been executed on existing placement connection for a colocated placement
```

That's why I switched to this method.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -1029,6 +1104,30 @@ PlanAlterTableStmt(AlterTableStmt *alterTableStatement, const char *alterTableCo
...
+			Oid relationId = get_relname_relid(relationName,
+											   get_namespace_oid(schemaName, false));
+			Oid parentId = leftRelationId;
+
+			/*
+			 * Do not generate tasks if relation is not distributed and the parent
+			 * is distributed. Because, we'll manually convert the relation into
+			 * distribtued relation and co-locate with its parent.
+			 */
+			if (!IsDistributedTable(relationId) && IsDistributedTable(parentId))
```
onderkalaci: Yes, that's in the roadmap for this PR. And, your suggestion makes sense.

src/backend/distributed/commands/create_distributed_table.c
marcocitus: I think that only happens if you alternate between `CreateHashDistributedTable` and `CopyLocalDataIntoShards`, but not if you do all `CreateHashDistributedTable` calls first. Burak's PR should also address this.

onderkalaci: That's right, it happened while code was alternating between the two functions. So, it'd probably make sense to add a new parameter to `CreateHashDistributedTable ` such as `loadLocalData` or so and skip loading the data for partitions in the first iteration of creating the distributed tables. Is that OK?

> Burak's PR should also address this.

I realized that we're changing lots of common codes with that PR.  Did we pause on that PR now? Which one do you think would go into master first?

onderkalaci: > Burak's PR should also address this.

@marcocitus Well, yes without that PR, I still get the error. Could you please help me understand what is happening?

[Here](https://github.com/citusdata/citus/compare/partition_distributed_table_create_table...connection_error?expand=1) is the roughly the changes that I'm planning to make on top of this branch (i.e., change `CopyLocalDataIntoShards ` to be as it is and do not alternate between copy & DDL). When you swith to that branch, you'll see the following:

```SQL

CREATE TABLE partitioning_test(id int, time date) PARTITION BY RANGE (time);
 
-- create its partitions
CREATE TABLE partitioning_test_2009 PARTITION OF partitioning_test FOR VALUES FROM ('2009-01-01') TO ('2010-01-01');
CREATE TABLE partitioning_test_2010 PARTITION OF partitioning_test FOR VALUES FROM ('2010-01-01') TO ('2011-01-01');

-- load some data and distribute tables
INSERT INTO partitioning_test VALUES (1, '2009-06-06');
INSERT INTO partitioning_test VALUES (2, '2010-07-07');
INSERT INTO partitioning_test VALUES (3, '2009-06-07');
INSERT INTO partitioning_test VALUES (4, '2010-07-08');
INSERT INTO partitioning_test_2009 VALUES (5, '2009-09-09');
INSERT INTO partitioning_test_2010 VALUES (6, '2010-03-03');
INSERT INTO partitioning_test_2009 VALUES (7, '2009-09-09');
INSERT INTO partitioning_test_2010 VALUES (8, '2010-03-03');

SET citus.shard_count TO 4;

-- this should error out given that parent of the partition is not distributed
SELECT create_distributed_table('partitioning_test_2010', 'id');

-- this should suceed
SELECT create_distributed_table('partitioning_test', 'id');

ERROR:  cannot establish new placement connection when DML has been executed on existing placement connection for a colocated placement
```

However, when I try similar stuff (without partitioning, using transaction blocks explicitly), the similar things works:
```SQL
CREATE TABLE test_0 (a int, b int);
INSERT INTO test_0 SELECT s, s FROM generate_series(0, 100000) s;
SET citus.shard_count TO 4;
SELECT create_distributed_table('test_0', 'a');

CREATE TABLE test_1 (a int, b int);
CREATE TABLE test_2 (a int, b int);

INSERT INTO test_1 SELECT s, s FROM generate_series(0, 100000) s;
COPY test_1 TO '/tmp/data.csv' WITH CSV;

BEGIN;

SET citus.shard_count TO 4;
SELECT create_distributed_table('test_1', 'a');
SELECT create_distributed_table('test_2', 'a');

CREATE INDEX i1 ON test_1 (a);
CREATE INDEX i2 ON test_2 (a);

COPY test_1 FROM '/tmp/data.csv' WITH CSV;


COPY test_2 FROM '/tmp/data.csv' WITH CSV;

```

So, do you have any ideas on `ERROR:  cannot establish new placement connection when DML has been executed on existing placement connection for a colocated placement` ?

Shall I wait for Burak's PR for that?

marcocitus: `CreateHashDistributedTable` calls `ExecuteRemoteQuery` to create shards, which calls `MultiClientConnectStart` to open a connection. `MultiClientConnectStart` is not allowed after a modification (e.g. CopyLocalDataIntoShards) since it might not use the right placement connection.

However, you should be able to call `CreateHashDistributedTable` for all tables, and then `CopyLocalDataIntoShards` for all tables.

onderkalaci: > However, you should be able to call CreateHashDistributedTable for all tables, and then CopyLocalDataIntoShards for all tables.

Yes, I did exactly the above on this [branch](https://github.com/citusdata/citus/compare/partition_distributed_table_create_table...connection_error?expand=1) but the behavior is still the same. I'll debug more and share my understanding on what's happening.

byucesoy:
Closing this in favor of #1509

PR #1468, opened 2017-06-22
Add weird column name test for create_distributed_table

### LINKED ISSUE #1458: Double quoted column names fails if there is data in table before distributing
If there is data inside a table with double quoted column names, `create_distributed_table()` and `create_reference_table()` fail.

```sql
CREATE TABLE test_table
(
  id uuid NOT NULL,
  tenant_id uuid NOT NULL,
  "fileHost" character varying(255)
);

insert into test_table values ('a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11', 'a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11', 'test');

SELECT create_distributed_table('test_table', 'tenant_id');
ERROR:  column "filehost" of relation "test_table_102307" does not exist
```

```sql

CREATE TABLE public.tenant
(
  id uuid NOT NULL,
  name character varying(255),
  "group" character varying(255)
);

insert into tenant values ('a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11', 'test', 'test');

SELECT create_reference_table('tenant');
ERROR:  syntax error at or near "group"
```

Note that if you distribute empty table and then insert data, it works fine.

This issue came across in citus users group.

marcocitus:
Fixes #1458

This was already fixed by #1402, but this PR adds a regression test specifically for `create_distributed_table`

### DIFF
--- a/src/test/regress/expected/multi_create_table.out
+++ b/src/test/regress/expected/multi_create_table.out
@@ -490,7 +490,7 @@ INSERT INTO data_load_test VALUES (243, 'world');
 DROP TABLE data_load_test;
 END;
 -- Test data loading after dropping a column
-CREATE TABLE data_load_test (col1 int, col2 text, col3 text);
+CREATE TABLE data_load_test (col1 int, col2 text, col3 text, "CoL4"")" int);
 INSERT INTO data_load_test VALUES (132, 'hello', 'world');
 INSERT INTO data_load_test VALUES (243, 'world', 'hello');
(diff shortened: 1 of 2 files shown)


metdos:
Looks good, also tested with examples in #1458. 🚢

jasonmp85:
Since the fix was backported, might make sense to backport the test as well. Adding the label in a bit.

PR #1469, opened 2017-06-22
Move INSERT ... SELECT planning logic into one place

marcocitus:
As discussed in #1402, this PR moves all INSERT ... SELECT planning logic to `insert_select_planner.c` without changing any code.


src/backend/distributed/planner/insert_select_planner.c
```diff
@@ -252,3 +501,743 @@ WrapSubquery(Query *subquery)
...
+	return modifyTask;
+}
+
+
+/*
+ * InsertSelectQueryNotSupported returns NULL if the INSERT ... SELECT query
+ * is supported, or a description why not.
+ */
+static DeferredErrorMessage *
+InsertSelectQuerySupported(Query *queryTree, RangeTblEntry *insertRte,
```
onderkalaci: Minor, feel free to ignore:

Given that we've two different implementations, shall we rename it to `DefereErrorIfDistributedInsertSelectQueryNoySupported` or so?

src/include/distributed/insert_select_planner.h
```diff
@@ -17,13 +17,22 @@
...
 #include "nodes/parsenodes.h"
 #include "nodes/plannodes.h"
 
 
+extern bool InsertSelectIntoDistributedTable(Query *query);
 extern MultiPlan * CreateCoordinatorInsertSelectPlan(Query *originalQuery);
+extern Query * ReorderInsertSelectTargetLists(Query *originalQuery,
+											  RangeTblEntry *insertRte,
+											  RangeTblEntry *subqueryRte);
 extern void CoordinatorInsertSelectExplainScan(CustomScanState *node, List *ancestors,
```
onderkalaci: `CoordinatorInsertSelectExplainScan ` is defined in `multi_explain.c`. Is that intentional? Or shall we move declaration to `multi_explain.h` or so?

marcocitus: It should be in multi_explain.c, since it's comparable to `CitusExplainScan` and calls static explain functions that need to remain static (due to name collisions with postgres).  Strangely, `CitusExplainScan` is declared in `multi_executor.h`

src/backend/distributed/planner/insert_select_planner.c
marcocitus: I've switched to `CoordinatorInsertSelectSupported` and `DistributedInsertSelectSupported` since the alternatives were impractically long.

Issue #883, opened 2016-10-18
Master-Issue: Duplicated code from PostgreSQL that can be removed:


anarazel:
This issue is supposed to track code duplication between postgres and similar code smells. Once an individual issue is being addressed a separate item for it should be created and then closed with a pull request.

After 9.5 support has been dropped:
- [x] simplify/deduplicate MasterIrreducibleExpressionWalker() using check_functions_in_node() (See #1478)
- [x] copy of readfuncs.c and related code in citus_readfuncs.c/citus_outfuncs.c (See #1470)

mtuncer:
both subtasks are merged
closing the issue

Issue #703, opened 2016-08-04
Evaluate reducing serialization cost during query planning


onderkalaci:
We already have some discussion around this issue in citus-mx [repo](https://github.com/citusdata/citus-mx/issues/14). We're not sure to check-in this feature immediately, see the discussion below:

The observation here is that it takes long time to  serialize/de-serialize target list entries as the target list grows. However, router executor has nothing to do with the target list elements that we have in the workerJob except for evaluating functions on the master node.

So, we could discuss getting rid of the target list for the router jobs which has performance benefits.

I created a branch called performance/reduce_serialization_cost and did some tests on citus with 1 + 2 c4.2xlarge cluster where the shard size and replication factor is fixed (128 and 1). On each test, I changed the column*\* count to see the effect of the change. Below are the results (inserts per second):

|  | master | performance/reduce_serialization_cost |
| --- | --- | --- |
| 1 Column: | 34090 | 36200 |
| 16 Columns: | 18184 | 26101 |
| 128 Columns: | 7020 | 13200 |
| 512 Columns: | 1088 | 2488 |
| 1024 Columns: | 485 | 963 |

On the average, (excluding 1 column case), we get **1.5x** to **2x** improvement. 

There is a PR which we've not merged, but a representative solution for the fix: #648 

But, I should also note that the above PR conflicts with [Evaluate functions on master](https://github.com/citusdata/citus/pull/488), so, we should find slightly different way of implementing this optimization.

Also, some important notes from @anarazel:
-  Btw, the citus branch for 9.6 support that I have, gets entirely rid of the (de-)serialization; by using the 9.6 support for extensible nodes.
- I think it'll quite possibly cause issues for prepared statements for router queries. There we need to keep the Query around, and re-generate SQL after replacing parameters. After I mention that this change is scoped with router plans: There it'll possibly become problematic when we support prepared params on the partitioning key.
- I'm a bit suspicious of the general approach. I think it's better to get rid of excessive copying, and then do relatively fragile optimizations like this one. But on the other hand, this is a fairly major win, for few lines of code.
- Because it breaks the semantics of a well defined data structure. Again, I think it might be worthwhile in this case, but I think we should be hesitant to do that in more places.

onderkalaci:
@metdos / @anarazel I think this is a good reason to stop serialization/deserialization 😄  Are we close to this after the custom scans?

metdos:
[...]

Issue #120, opened 2016-02-03
Allow Custom Aggregates


sumedhpathak:
We currently have a hard-coded white-list for aggregates. For commutative/associative aggregates, we overload the `sum` function definition. We'd like to allow people to define their own transition/final functions and define them for use and pushdown in Citus.

lithp:
This just came up in a call with a client

lithp:
This appeared again in engage@

ozgune:
This feature request recently came up in citus-users (and a related email thread to engage@). I'm capturing it here for reference purposes: https://groups.google.com/forum/#!topic/citus-users/5Jd2zbxAs3Q

lithp:
This came up again in the form of #1419

lithp:
We should be able to support both of these cases:
- Any aggregate of the form `f: a, a -> a`
- On pg 9.6 and higher, any aggregate which sets a combinefunc (Has a "[partial mode](https://www.postgresql.org/docs/9.6/static/functions-aggregate.html)" )

Unfortunately, there are useful aggregates this doesn't include:
- `[hll_union_agg](https://github.com/aggregateknowledge/postgresql-hll/blob/master/hll--2.10.0.sql#L465)`
- `jsonb_agg`

lithp:
- Both `hll_union_agg` and `jsonb_agg` support can be added by the user by creating a custom aggregate which includes a combinefunc.
- Any functions which use the `internal` transtype must specify a `serialfunc` and `deserialfunc`
- We still need to special-case count-distinct but code which specifially handles "count", "min", "max", "avg", etc, can be removed.

lithp:
Tasks:
- [ ] WorkerAggregateExpressionList no longer needs to special case `AGGREGATE_AVERAGE`, averages can simply be pushed down

lithp:
- [ ] Test that this works for #1419, we should be able to automatically pickup and pushdown enum_smaller

lithp:
(Note: this is all wrong and this change is a lot more work than I expected, going to delete these comments and add a new one with a more fleshed out proposal)

lithp:
**Proposal**

Create two aggregates;
- `worker_run_agg(aggregate oid, anyelement value)`,
-  and `coord_run_agg(aggregate oid, anyelement value)`.

[...]

PR #1470, opened 2017-06-23
Remove 9.5 support.

anarazel:
This removes 9.5 specific code files, 9.5 conditional code, and the now unneeded CitusNode* wrappers.

This includes parts of #883, namely the duplicated/copied readfuncs/read.c code.


src/test/regress/expected/multi_join_order_additional.out
```diff
@@ -2,13 +2,6 @@
 -- MULTI_JOIN_ORDER_ADDITIONAL
 --
 ALTER SEQUENCE pg_catalog.pg_dist_shardid_seq RESTART 650000;
--- print whether we're running on 9.5 to make version-specific tests clear
-SELECT substring(version(), '\d+(?:\.\d+)?') = '9.5' AS is_95;
- is_95 
--------
- f
```
jasonmp85: You removed the test file that is _not_ `9.5`?

jasonmp85: Oh. Now I see what happened here. Nevermind!

anarazel: Phew, you shocked me for a second.

src/include/distributed/citus_nodefuncs.h
```diff
@@ -47,19 +35,10 @@ extern void RegisterNodes(void);
...
-#else
-#define READFUNC_ARGS void
-#define READFUNC_RET Node *
-#endif
 
-#if (PG_VERSION_NUM >= 90600)
 #define OUTFUNC_ARGS StringInfo str, const struct ExtensibleNode *raw_node
-#else
-#define OUTFUNC_ARGS StringInfo str, const Node *raw_node
-#endif
```
anarazel: FWIW, we could remove OUTFUNC_ARGS etc now, but I left that for a separate PR.

jasonmp85: > left that for a separate PR

Maybe a separate commit? It's a find-and-replace, basically, right?

anarazel: > Maybe a separate commit?
Can, but not enthusiastic.

> It's a find-and-replace, basically, right?

Yea, pretty much.It makes the lines a good bit longer tho.

src/backend/distributed/planner/multi_planner.c
```diff
@@ -444,7 +444,7 @@ SerializeMultiPlan(MultiPlan *multiPlan)
 	char *serializedMultiPlan = NULL;
 	Const *multiPlanData = NULL;
 
-	serializedMultiPlan = CitusNodeToString(multiPlan);
```
jasonmp85: Hooray! 💀 to the duplicative `Citus`-prefixed functions!

jasonmp85:
This looks pretty good. Can you do a quick audit of `gitignore` and `gitattributes` to see whether lines can be removed? Are there any documentation references to `9.5`? Code comments?

I'm running some tests with code coverage locally to see whether there is any unexercised code that can be removed (i.e. maybe 9.5-specific code that isn't immediately apparent because it lacks any proximate `PG_VERSION_NUM` check, but isn't just untested code). Otherwise, I think this looks pretty straightforward.

anarazel:
> This looks pretty good. Can you do a quick audit of gitignore and gitattributes to see whether lines can be removed?

Good catch. Fix pushed.

> Are there any documentation references to 9.5?

I'm not sure what doc references there could be in the citus repo?

> Code comments?

There's one remaining 9.5 reference, but I think that can stay:
```
	 * Look through contain_mutable_functions_walker or future PG's equivalent for new
	 * node types before bumping this version number to fix compilation; e.g. for any
	 * PostgreSQL after 9.5, see check_functions_in_node.
```

jasonmp85:
> I'm not sure what doc references there could be in the citus repo?

`CONTRIBUTING.md`, apparently…

> There's one remaining 9.5 reference

The language around _or future PG's equivalent_ and _PostgreSQL after 9.5_ is extraneous now and should just be removed; can you rewrite it to just directly reference the function implementations we need to check in PostgreSQL 9.6 and 10? The rest is just noise with 9.5 gone.

anarazel:
>>  I'm not sure what doc references there could be in the citus repo?
> CONTRIBUTING.md, apparently…

Ah, right - I'd only searched src/{include,backend} for references, because there's  "9.5" strings in some of the test output, ignoring files would have been better.  Also found one more reference in pg_regress_multi.pl.  Pushed fixes for these.

> The language around or future PG's equivalent and PostgreSQL after 9.5 is extraneous now and should just be removed; can you rewrite it to just directly reference the function implementations we need to check in PostgreSQL 9.6 and 10? The rest is just noise with 9.5 gone.

Seems like pointless diff noise to me.

jasonmp85:
OK, so I've audited other unused code and including here for completeness…

# Things Enabled by Removing 9.5

  * Kill off our `RangeTblFunction` hack in favor of always using custom nodes (tracked in #703)
  * Remove some `ruleutils` modifications after that
  * Remove whitelisted distributed aggregates and use parallel aggregate functions (tracked in #120)

# Dead Code

Not sure if always dead or now dead because of 9.5 changes. Checked rebalancer/enterprise for calls and found none:

 * `LowestShardIntervalById`
 * `UnlockShardResource`
 * `LockMetadataSnapshot`
 * `OpExpressionContainsColumn`

# Deprecated Paths

The workerlist file parsing code and many functions within `worker_node_manager` seem overdue for killing off.

jasonmp85:
> Seems like pointless diff noise to me.

Having read that comment a couple of times and always having to make it to the very end to actually get the information I'm after, I disagree (people are going to read the comment more often than look at `git blame`), but not a big deal; would just be nice to have precisely the info we need for the versions we support, and have it frontloaded in that sentence.

No other feedback after my `lcov` runs.

anarazel:
> Not sure if always dead or now dead because of 9.5 changes. Checked rebalancer/enterprise for calls and found none:

Don't think any of these are because of the 9.5 removal.

> LowestShardIntervalById

1585c0232277e0a02952105116f398df037993cb

Seems not unlikely to gain a new caller again.

> UnlockShardResource

Don't think we ever had a caller, that's more completeness I think. I'd keep it.

> LockMetadataSnapshot

92c756700831392aa901ec67636548a72e58b490

Hm. Unclear. @marcocitus ?

> OpExpressionContainsColumn

d399f395f7c409fc43a1affcd37a127557e1bb5c

I think this one should be removed.

anarazel:
Thanks for the review!

Issue #1472, opened 2017-06-29
rollback & connection end handling doesn't cancel / abort reliably.


anarazel:
There's two bugs here:

1. If a statement is run within a coordinated transaction (e.g. DML has been issued in transaction) transaction rollback will call ForgetResults() on the connection, to be able to send the `ROLLBACK;`. Unfortunately that'll currently frequently actually wait till the underlying statement is completed. 

1. When a connections are closed at the end of a statement that failed, e.g. due to a cancel interrupt, it'll frequently continue to run on the worker. The reason for that is that we simply `PQfinish()` the connection, which doesn't guarantee that the statement /backend running on the worker finishes immediately. It'll once the backend tries to write (e.g. results) or read (e.g. next statement) from the network, but for some statements that could be a long while.

I noticed this when working on deadlock detection, where we need to cancel ongoing statements in a reliable manner.

anarazel:
Any opinions about backporting? It's not an entirely trivial fix, and I've not heard loud complaints.

Issue #1100, opened 2017-01-06
Remove direct usage of PQexec/PQgetResult/... to avoid blocking interrupts


anarazel:
It's bad to block interrupts when interacting with remote servers, because that can require restarting the server when encountering e.g. a distributed deadlock.   Instead SendRemoteCommand(), GetRemoteCommandResult() etc. should be used.

This is partially blocked by #873

Issue #1121, opened 2017-01-12
Remove sections of code blocking interrupts


anarazel:
A good bit of code recently learned to cope with interrupts - the connection mgmt/transaction mgmt infrastructure makes that a lot easier. Some code still blocks because of libpq calls (#1100), but there's also sections of code where we explicitly avoid interrupts:
- [ ] task tracker
- [ ] real-time
- [ ] placement creation (```CreateShardsWithRoundRobinPolicy()```)
- [ ] master_append_table_to_shard() (bad, can take a long time)
- [ ] DropShards()  (bad, can take a long time)

PR #1473, opened 2017-06-29
Fix issues around statement cancellation and connection closure.

anarazel:
Fixes: #1472
Fixes: #1100
See: #1121

### DIFF
--- a/src/backend/distributed/connection/connection_management.c
+++ b/src/backend/distributed/connection/connection_management.c
@@ -640,6 +640,13 @@ StartConnectionEstablishment(ConnectionHashKey *key)
 	connection->pgConn = PQconnectStartParams(keywords, values, false);
 	connection->connectionStart = GetCurrentTimestamp();
 
+	/*
+	 * To avoid issues with interrupts not getting caught all our connections
+	 * are managed in a non-blocking manner. remote_commands.c provides
+	 * wrappers emulating blocking behaviour.
+	 */
+	PQsetnonblocking(connection->pgConn, true);
+
 	return connection;
 }
 
--- a/src/backend/distributed/connection/connection_management.c
+++ b/src/backend/distributed/connection/connection_management.c
@@ -718,8 +718,7 @@ AfterXactHostConnectionHandling(ConnectionHashEntry *entry, bool isCommit)
 			PQstatus(connection->pgConn) != CONNECTION_OK ||
 			PQtransactionStatus(connection->pgConn) != PQTRANS_IDLE)
 		{
-			PQfinish(connection->pgConn);
-			connection->pgConn = NULL;
+			ShutdownConnection(connection);
 
 			/* unlink from list */
 			dlist_delete(iter.cur);
--- a/src/backend/distributed/connection/connection_management.c
+++ b/src/backend/distributed/connection/connection_management.c
@@ -402,6 +402,40 @@ CloseConnectionByPGconn(PGconn *pqConn)
 
 
 /*
+ * ShutdownConnection, if necessary cancels the currently running statement,
+ * and then closes the underlying libpq connection.  The MultiConnection
+ * itself is left intact.
+ *
+ * NB: Cancelling a statement requires network IO, and currently is not
+ * interruptible. Unfortunately libpq does not provide a non-blocking
+ * implementation of PQcancel(), so we don't have much choice for now.
+ */
+void
+ShutdownConnection(MultiConnection *connection)
+{
+	/*
+	 * Only cancel statement if there's currently one running, and the
+	 * connection is in an OK state.
+	 */
+	if (PQstatus(connection->pgConn) == CONNECTION_OK &&
+		PQtransactionStatus(connection->pgConn) == PQTRANS_ACTIVE)
+	{
+		char errorMessage[256] = { 0 };
+		PGcancel *cancel = PQgetCancel(connection->pgConn);
+
+		if (!PQcancel(cancel, errorMessage, sizeof(errorMessage)))
+		{
+			ereport(WARNING, (errmsg("could not cancel connection: %s",
+									 errorMessage)));
+		}
+		PQfreeCancel(cancel);
+	}
+	PQfinish(connection->pgConn);
+	connection->pgConn = NULL;
+}
+
+
+/*
  * FinishConnectionListEstablishment is a wrapper around FinishConnectionEstablishment.
  * The function iterates over the multiConnectionList and finishes the connection
  * establishment for each multi connection.
--- a/src/backend/distributed/connection/remote_commands.c
+++ b/src/backend/distributed/connection/remote_commands.c
@@ -23,6 +23,7 @@
 /* GUC, determining whether statements sent to remote nodes are logged */
 bool LogRemoteCommands = false;
 
+static bool FinishConnectionIO(MultiConnection *connection, bool raiseInterrupts);
 
 /* simple helpers */
 
@@ -367,7 +368,7 @@ ReadFirstColumnAsText(PGresult *queryResult)
  * an error.
  *
  * If raiseInterrupts is false and an interrupt arrives that'd otherwise raise
- * an error, GetRemotecommandResult returns NULL, and the transaction is
+ * an error, GetRemoteCommandResult returns NULL, and the transaction is
  * marked as having failed. While that's not a perfect way to signal failure,
  * callers will usually treat that as an error, and it's easy to use.
  *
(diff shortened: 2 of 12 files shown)


anarazel:
If you want to see the effects of not patching, run the added isolationtests without the fix, but be prepared to wait a bit, it'll take 10ksec ;)

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -276,10 +276,22 @@ StartRemoteTransactionAbort(MultiConnection *connection)
...
 	 */
-	ForgetResults(connection);
+	if (transaction->transactionState == REMOTE_TRANS_PREPARING ||
+		transaction->transactionState == REMOTE_TRANS_PREPARED)
+	{
+		ForgetResults(connection);
+	}
+	else if (!NonblockingForgetResults(connection))
+	{
+		ShutdownConnection(connection);
```
marcocitus: This sets `pgConn` to NULL and then we go into `SendRemoteCommand`, which returns 0. This looks a little dodgy. Could we maybe do `MarkRemoteTransactionFailed(connection, dontRaiseErrors);` here and return?

src/backend/distributed/connection/remote_commands.c
```diff
@@ -75,6 +75,90 @@ ForgetResults(MultiConnection *connection)
...
+		/* clearing would require blocking IO, return */
+		if (PQisBusy(pgConn))
+		{
+			break;
+		}
+
+		result = PQgetResult(pgConn);
+		if (PQresultStatus(result) == PGRES_COPY_IN)
+		{
+			PQputCopyEnd(connection->pgConn, NULL);
```
marcocitus: If we get here we're going to return false and close the connection, since we're almost always going to be busy in the next iteration. What if we just break here?

Minor note: When using cstore_fdw, `PQputCopyEnd` would trigger a "commit".

src/backend/distributed/connection/remote_commands.c
```diff
@@ -75,6 +75,90 @@ ForgetResults(MultiConnection *connection)
...
+		return false;
+	}
+
+	while (true)
+	{
+		PGresult *result = NULL;
+
+		CHECK_FOR_INTERRUPTS();
+
+		if (!wasNonblocking)
```
marcocitus: These annoy me a bit, since we now have this set / unset in 7 different places. Ideally, we would just set our sockets to non-blocking mode, though I'm guessing that will mess up copy. To simplify the code, could we maybe add a flag in `MultiConnection` that says whether the connection is blocking so we can get rid of the `wasNonBlocking` checks? That would also simplify asynchronous commit.

src/backend/distributed/connection/remote_commands.c
```diff
@@ -75,6 +75,90 @@ ForgetResults(MultiConnection *connection)
...
 /*
+ * NonblockingForgetResults clears a connection from pending activity if doing
+ * so does not require network IO. Returns true if successful, false
+ * otherwise.
+ */
+bool
+NonblockingForgetResults(MultiConnection *connection)
+{
+	PGconn *pgConn = connection->pgConn;
+	bool wasNonblocking = PQisnonblocking(pgConn);
```
marcocitus: Might be useful to check for `pgConn == null`

src/backend/distributed/connection/remote_commands.c
```diff
@@ -75,6 +75,90 @@ ForgetResults(MultiConnection *connection)
...
+	if (PQstatus(pgConn) != CONNECTION_OK)
+	{
+		return false;
+	}
+
+	while (true)
+	{
+		PGresult *result = NULL;
+
+		CHECK_FOR_INTERRUPTS();
```
marcocitus: Given that this loop doesn't block, do we need `CHECK_FOR_INTERRUPTS` inside the loop? It would simplify the code if we can't error out here (e.g. we can move `PQsetnonblocking` outside the loop).

src/backend/distributed/connection/connection_management.c
```diff
@@ -402,6 +402,40 @@ CloseConnectionByPGconn(PGconn *pqConn)
 
 
 /*
+ * ShutdownConnection, if necessary cancels the currently running statement,
+ * and then closes the underlying libpq connection.  The MultiConnection
+ * itself is left intact.
+ *
+ * NB: Cancelling a statement requires network IO, and currently is not
+ * interruptible. Unfortunately libpq does not provide a non-blocking
+ * implementation of PQcancel(), so we don't have much choice for now.
```
anarazel: Yea, it really sucks.  I'm going to try to get a better implementation of this into postgres, but it'll be a while before we can use it.  It's a very different code-path from normal libpq, because it needs to be able to run from a signal handler, so it might be suitable to be copied into our code at some point.

src/backend/distributed/connection/remote_commands.c
anarazel: Good catch.

src/backend/distributed/connection/remote_commands.c
anarazel: Yes, I think we need to. Consider the case that the other side is sending results faster than we can process them...

src/backend/distributed/connection/remote_commands.c
anarazel: >  Ideally, we would just set our sockets to non-blocking mode, though I'm guessing that will mess up copy. 

Yes, it'd currently upset copy, and citus-tools.  Once we've converted them to be interruptible, we should be good to go however.

> To simplify the code, could we maybe add a flag in MultiConnection that says whether the connection is blocking so we can get rid of the wasNonBlocking checks?

Wouldn't that mean we'd have to reset it in a bunch of places that are currently unrelated?  I think I'd rather get rid of the last few places that require blocking sockets and then always put them into nonblocking mode.

src/backend/distributed/connection/remote_commands.c
anarazel: That works for me too.

src/backend/distributed/connection/remote_commands.c
anarazel: > Wouldn't that mean we'd have to reset it in a bunch of places that are currently unrelated? I think I'd rather get rid of the last few places that require blocking sockets and then always put them into nonblocking mode.

That's indeed the case, but I chose to "solve" this by simply getting rid by converting all other callers.

anarazel:
I've expanded the scope of this a bit...  Do you want me to split the first half (before the bug fixes) into a separate PR?

src/backend/distributed/transaction/remote_transaction.c
marcocitus: Maybe do:
```
    if (transaction->transactionState == REMOTE_TRANS_PREPARING ||
       transaction->transactionState == REMOTE_TRANS_PREPARED)
    {
        StringInfoData command;

        /* await PREPARE TRANSACTION results, closing the connection would leave it dangling */
        ForgetResults(connection);

 	initStringInfo(&command);
 	appendStringInfo(&command, "ROLLBACK PREPARED '%s'",
                                   transaction->preparedName); 
        ...
    }
    else
    {
        if (!NonblockingForgetResults(connection))
        {
            /* can't sent ROLLBACK yet because worker is still busy, close the connection instead */
            ShutdownConnection(connection);
        }
        if (!SendRemoteCommand(connection, "ROLLBACK"))
        {
            ....
        }
        else
        {
            ...
        }
    }
```

src/backend/distributed/executor/multi_client_executor.c
```diff
@@ -854,23 +857,6 @@ MultiClientWait(WaitInfo *waitInfo)
 
 
 /*
- * ClearRemainingResults reads result objects from the connection until we get
- * null, and clears these results. This is the last step in completing an async
- * query.
- */
-static void
-ClearRemainingResults(MultiConnection *connection)
```
marcocitus: To Ozgun's point, this was probably a better name than `ForgetResults`. Would be nice to use this name instead, but can skip over that for now.

anarazel: "Ozgun's point" - which are poing are you referring to?

src/backend/distributed/transaction/remote_transaction.c
anarazel: Yea, that works.

src/backend/distributed/executor/multi_client_executor.c
marcocitus: In the deadlock meeting Ozgun brought up that `ForgetResults` is a strange name.

anarazel:
Thanks for the review @marcocitus !

src/backend/distributed/connection/remote_commands.c
```diff
@@ -395,106 +447,159 @@ GetRemoteCommandResult(MultiConnection *connection, bool raiseInterrupts)
...
 		/* if reading fails, there's not much we can do */
 		if (PQconsumeInput(pgConn) == 0)
 		{
-			failed = true;
-			break;
+			return false;
+		}
+		if (PQisBusy(pgConn))
+		{
+			waitFlags |= WL_SOCKET_READABLE;
```
marcocitus: You just made me realise I overlooked this :). We should always check for reads even if the socket is not busy, since the server could be sending notices and get blocked on them, meaning it won't accepts writes until we read.

PR #1474, opened 2017-07-03
Execute UPDATE/DELETE statements with 0 shards

### LINKED ISSUE #127: Invalidate cached plans for PREPARE statements.
When metadata changes, we need to invalidate the cached plans. The reason is that when metadata changes, we need to build new MultiPlan according to the updated metadata. For instance, currenltly if we add new shards, prepared statements would not be able to get data from those shards.

### LINKED ISSUE #1230: Citus Query Planner is unable to handle 1=0 conditions
Citus version: 6.0-17

`explain  DELETE FROM "my_model" WHERE "my_model"."my_shard_key" = 27;
```
                                                          QUERY PLAN

Distributed Query
   Executor: Router
   Task Count: 1
   Tasks Shown: All
   ->  Task
         Node: host=ec2-host.compute-1.amazonaws.com   port=5432 dbname=citus
         ->  Delete on my_model_102737  (cost=4.20..13.67 rows=6 width=6)
               ->  Bitmap Heap Scan on my_model_102737  (cost=4.20..13.67 rows=6 width=6)
                     Recheck Cond: (my_shard_key = 27)
                     ->  Bitmap Index Scan on index_my_model_on_shard_key_102737  (cost=0.00..4.20 rows=6 width=0)
                           Index Cond: (my_shard_key = 27)
```
(11 rows)

`explain  DELETE FROM "my_model" WHERE "my_model"."my_shard_key" = 27 and 1=0;`
`ERROR:  distributed modifications must target exactly one shard`

Adding 1=0 makes the query planner unable to see that it has a valid shard_key.

Unfortunately 1=0 is weird byproduct of rails ActiveRecord code when trying to:
`MyModel.where(id: []).delete_all`, 
so ideally this could be a no-op.

marcocitus:
This PR adds execution logic for `UPDATE` and `DELETE` commands that prune down to 0 shard, which currently error out.

The router planner currently makes a lot of assumptions about a job containing exactly 1 task. This PR does considerable refactoring to be able to generate a list of tasks for a job in either the planner or the executor (depending on whether the partition column value is known at planning time), which also paves the way for supporting multi-row INSERT and multi-shard UPDATE/DELETE.

There are 2 subtle behavioural changes that resolve some outstanding issues:
- Placement lists are now always loaded in the executor, after acquiring metadata locks. This resolves a race condition where a prepared statement could be executed with a stale placement list. It also prevents prepared DML statements that run during a shard copy from throwing an error.
- Parameters in modifications are now always evaluated on the master. This simplifies handling of parameters and functions, which was strangely inconsistent and prone to bugs (e.g. previously parameters were evaluated on the master if the query contained a stable function, but pushed down otherwise).

Note: I wrote this a few weeks back and got a little stuck with 0-task SELECT queries and didn't want to spend more time on it (plus it seemed invasive). However, it seems like it might be useful for #1465 and other ongoing work.

Fixes #1230
Fixes #127

### DIFF
--- a/src/backend/distributed/executor/multi_executor.c
+++ b/src/backend/distributed/executor/multi_executor.c
@@ -158,8 +158,8 @@ RouterCreateScan(CustomScan *scan)
 
 	isModificationQuery = IsModifyMultiPlan(multiPlan);
 
-	/* check if this is a single shard query */
-	if (list_length(taskList) == 1)
+	/* check whether query has at most one shard */
+	if (list_length(taskList) <= 1)
 	{
 		if (isModificationQuery)
 		{
--- a/src/backend/distributed/executor/multi_router_executor.c
+++ b/src/backend/distributed/executor/multi_router_executor.c
@@ -73,7 +73,7 @@ bool AllModificationsCommutative = false;
 bool EnableDeadlockPrevention = true;
 
 /* functions needed during run phase */
-static void ReacquireMetadataLocks(List *taskList);
+static void AcquireMetadataLocks(List *taskList);
 static ShardPlacementAccess * CreatePlacementAccess(ShardPlacement *placement,
 													ShardPlacementAccessType accessType);
 static void ExecuteSingleModifyTask(CitusScanState *scanState, Task *task,
@@ -100,22 +100,12 @@ static bool ConsumeQueryResult(MultiConnection *connection, bool failOnError,
 
 
 /*
- * ReacquireMetadataLocks re-acquires the metadata locks that are normally
- * acquired during planning.
- *
- * If we are executing a prepared statement, then planning might have
- * happened in a separate transaction and advisory locks are no longer
- * held. If a shard is currently being repaired/copied/moved, then
- * obtaining the locks will fail and this function throws an error to
- * prevent executing a stale plan.
- *
- * If we are executing a non-prepared statement or planning happened in
- * the same transaction, then we already have the locks and obtain them
- * again here. Since we always release these locks at the end of the
- * transaction, this is effectively a no-op.
+ * AcquireMetadataLocks acquires metadata locks on each of the anchor
+ * shards in the task list to prevent a shard being modified while it
+ * is being copied.
  */
 static void
-ReacquireMetadataLocks(List *taskList)
+AcquireMetadataLocks(List *taskList)
 {
 	ListCell *taskCell = NULL;
 
@@ -130,28 +120,7 @@ ReacquireMetadataLocks(List *taskList)
 	{
 		Task *task = (Task *) lfirst(taskCell);
 
-		/*
-		 * Only obtain metadata locks for modifications to allow reads to
-		 * proceed during shard copy.
-		 */
-		if (task->taskType == MODIFY_TASK &&
-			!TryLockShardDistributionMetadata(task->anchorShardId, ShareLock))
-		{
-			/*
-			 * We could error out immediately to give quick feedback to the
-			 * client, but this might complicate flow control and our default
-			 * behaviour during shard copy is to block.
-			 *
-			 * Block until the lock becomes available such that the next command
-			 * will likely succeed and use the serialization failure error code
-			 * to signal to the client that it should retry the current command.
-			 */
-			LockShardDistributionMetadata(task->anchorShardId, ShareLock);
-
-			ereport(ERROR, (errcode(ERRCODE_T_R_SERIALIZATION_FAILURE),
-							errmsg("prepared modifications cannot be executed on "
-								   "a shard while it is being copied")));
-		}
+		LockShardDistributionMetadata(task->anchorShardId, ShareLock);
 	}
 }
 
@@ -457,7 +426,7 @@ CitusModifyBeginScan(CustomScanState *node, EState *estate, int eflags)
 	}
 
 	/* prevent concurrent placement changes */
-	ReacquireMetadataLocks(taskList);
+	AcquireMetadataLocks(taskList);
 
 	/* assign task placements */
 	workerJob->taskList = FirstReplicaAssignTaskList(taskList);
--- a/src/backend/distributed/executor/multi_router_executor.c
+++ b/src/backend/distributed/executor/multi_router_executor.c
@@ -74,7 +74,6 @@ bool EnableDeadlockPrevention = true;
 
 /* functions needed during run phase */
 static void ReacquireMetadataLocks(List *taskList);
-static void AssignInsertTaskShardId(Query *jobQuery, List *taskList);
 static ShardPlacementAccess * CreatePlacementAccess(ShardPlacement *placement,
 													ShardPlacementAccessType accessType);
 static void ExecuteSingleModifyTask(CitusScanState *scanState, Task *task,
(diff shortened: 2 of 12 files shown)


src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1106,83 +1104,151 @@ TargetEntryChangesValue(TargetEntry *targetEntry, Var *column, FromExpr *joinTre
...
+{
+	uint32 rangeTableId = 1;
+	Var *partitionColumn = NULL;
+	Expr *partitionValueExpr = NULL;
 
-		/* grab shared metadata lock to stop concurrent placement additions */
-		LockShardDistributionMetadata(shardId, ShareLock);
+	if (query->commandType != CMD_INSERT)
+	{
+		/* we assume UPDATE/DELETE is always prunable */
```
metdos: I'm changing the logic which calls here for UPDATE/DELETE. However, I didn't get the why we assume "UPDATE/DELETE is always prunable here".

metdos:
@jasonmp85, @marcocitus 

I think this PR looks good, I have some questions/notes above. To summarize;

- Why CanShardPrune() only applies to INSERTs?
- Do we need to set executorState->es_param_list_info to NULL?
- Shall we add explain queries with zero-shard modifications?

jasonmp85:
Hey @metdos, I'll get around to some updates on this after updating #1517, which interacts quite a bit with this and I'm making the base of this PR. Thanks for the comments; I'll try to get to all of them tomorrow!

src/backend/distributed/planner/multi_router_planner.c
marcocitus: Mainly that we currently always perform shard pruning for UPDATE/DELETE in the planner. 

If there is no partition column filter or a non-constant filter (e.g. function), then shard pruning simply returns all shards, and this could in the future become a multi-shard UPDATE/DELETE. For INSERT things are a bit different since we really need to prune down to a single shard.

We could at some point decide to defer pruning for UPDATE/DELETE if there is a parameter or function in the partition column filter. In that case we'd change this function.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1106,83 +1104,151 @@ TargetEntryChangesValue(TargetEntry *targetEntry, Var *column, FromExpr *joinTre
...
+
+/*
+ * CreateJob returns a new Job for the given query.
+ */
+static Job *
+CreateJob(Query *query)
+{
+	Job *job = NULL;
+
+	job = CitusMakeNode(Job);
```
jasonmp85: Sure, I can do that but I'm not 100% on that being a style thing we follow…

src/backend/distributed/executor/multi_executor.c
```diff
@@ -159,7 +159,7 @@ RouterCreateScan(CustomScan *scan)
 	isModificationQuery = IsModifyMultiPlan(multiPlan);
 
 	/* check if this is a single shard query */
-	if (list_length(taskList) == 1)
+	if (list_length(taskList) <= 1)
```
jasonmp85: Fixed.

src/test/regress/sql/multi_simple_queries.sql
```diff
@@ -82,6 +82,7 @@ INSERT INTO articles_single_shard VALUES (50, 10, 'anjanette', 19519);
 
 -- zero-shard modifications should fail
 UPDATE articles SET title = '' WHERE author_id = 1 AND author_id = 2;
```
jasonmp85: Fixed comment.

src/backend/distributed/planner/multi_router_planner.c
jasonmp85: In short (I think), applying a `DELETE` or `UPDATE` on a larger set of shards than is strictly necessary should not have any ill effects (overhead, locking, etc. aside). The pruning is an optimization.

But if you apply an `INSERT` to too many shards, you're in for trouble.

So I'm guessing the comment means mainly "even if we can't prune, we can just return all shards". This would error out now presumably anyhow (though as @marcocitus pointed out, that need not be true in the near future).

metdos: Makes sense

src/backend/distributed/planner/multi_router_planner.c
jasonmp85: Ok, so one of the fields was omitted and one was duplicated, so yeah, this was a good suggestion.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1192,17 +1258,100 @@ TargetShardIntervalForModify(Oid distributedTableId, Query *query,
...
-	return shardInterval;
+	return taskList;
+}
+
+
+/*
+ * CreateTask returns a new Task with the given type.
+ */
+static Task *
+CreateTask(TaskType taskType)
```
jasonmp85: Sure, why not (though in most cases `palloc0` should be doing the work for us?)

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1401,17 +1536,24 @@ FindShardForUpdateOrDelete(Query *query, DeferredErrorMessage **planningError)
...
 
 		return NULL;
 	}
-
-	return (ShardInterval *) linitial(prunedShardList);
+	else if (prunedShardCount == 0)
+	{
+		return NULL;
+	}
+	else
```
jasonmp85: That is more our style. Fixed.

metdos:
> palloc0 should be doing the work for us?

I suppose.

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -426,75 +394,36 @@ CitusModifyBeginScan(CustomScanState *node, EState *estate, int eflags)
 	if (workerJob->requiresMasterEvaluation)
 	{
 		PlanState *planState = &(scanState->customScanState.ss.ps);
+		EState *executorState = planState->state;
 
 		ExecuteMasterEvaluableFunctions(jobQuery, planState);
 
+		/* we've already processed parameters */
+		executorState->es_param_list_info = NULL;
```
jasonmp85: @marcocitus if you have a quick answer, let me know but I think the following applies…

I believe that if we don't set this to `NULL`, then within either of the `Execute*ModifyTask(s)` functions, we'd find parameter info and serialize and send them over the wire. I'm not sure whether this actually causes any bad behavior, but it will definitely be wasted effort if we've already evaluated parameters and baked them into the SQL.

However, this is just my impression from reading. I think the comment deserves some explanation.

jasonmp85:
Added in the `EXPLAIN` tests and rebased, addressed code/style comments, and fixed any conflicts.

I believe the only outstanding issue is your second question. I'm not 100% sure on my interpretation, so maybe @marcocitus can chime in on why this line is necessary (I'm assuming it's an optimization, not a correctness thing?) and I can update the comment with my understanding (or add tests if it is indeed a correctness thing).

src/backend/distributed/executor/multi_router_executor.c
marcocitus: Yes, this was to avoid redundantly sending the values even though they're already in the deserialised query.

metdos: I see. Then, I think this PR looks and we can merge this.

marcocitus:
Expanded the comment, merging to unblock other PRs.

PR #1235, opened 2017-02-16
Add noderole to dist node, most functions only return primary nodes

lithp:
Wasn't able to add a partial unique index on (groupid, noderole) in order to ensure one primary per group because of [reasons](https://github.com/postgres/postgres/blob/master/src/backend/catalog/indexing.c#L114).

PR #1476, opened 2017-07-03
Add nodeRole column to pg_dist_node

lithp:
Rebasing #1235 was going to be a lot of work so I'm going through and re-implementing it

Current behavior:
- `master_add_node` and `master_add_inactive_node` now accept a `noderole` parameter, which defaults to `primary`.
- If the specified group already has a `primary` node then the call errors out
- When resolving `GroupShardPlacement`s into `ShardPlacement`s only `primary` nodes are considered

Remaining work:
- Add tests:
  - [x] It's not possible to add more than one primary to a group
  - [x] You get a useful error message when running a query and there's no primary in the requested group
  - [x] It's possible to add secondaries to a group
  - [x] It's possible to add multiple secondaries to a group
  - [x] secondaries are never picked when doing things like creating shards on workers
- Copy commits over from previous PR which exclude secondary nodes from some functions:
  - [x] [WorkerGetLiveNodeCount](https://github.com/citusdata/citus/pull/1235/commits/8f94bd3d14e112f84b0ad581a33b4c82b7669a6a)
  - [x] [WorkerGetRandomCandidateNode](https://github.com/citusdata/citus/pull/1235/commits/8260fcddbdce084798c09cc487eb0c3c01fe439c)
  - [x] [WorkerGetFirstLocalCandidateNode](https://github.com/citusdata/citus/pull/1235/commits/e39ac1821f7bfc081dc08b8d5226c3b5d8862623)
  - [x] [WorkerNodeList](https://github.com/citusdata/citus/pull/1235/commits/b081fb0efa650da7f75a876c52f4c489a0b34cee)
- [x] Once again look into making a unique constraint for one-primary-per-group. Where else can we try to enforce it? An insert/update trigger?
- [x] Check that you're allowed to remove nodes as long as they're not the last node in the group
- [x] Is it okay that `TupleToWorkerNode` doesn't check that `nodeRole` exists like [here](https://github.com/citusdata/citus/pull/1235/commits/f79d7a1f4d40cafabb8b51ef83fb56699693c286#diff-951979618a176dba2130cdf086a564abR916)?
- [ ] Add a trigger to enforce that you can't remove nodes which still have placements?
- [x] What about `CreateReferenceTable`?
- [x] Does `WorkerGetRandomCandidateNode` still do the right thing?

### DIFF
--- a/src/backend/distributed/citus--7.0-3--7.0-4.sql
+++ b/src/backend/distributed/citus--7.0-3--7.0-4.sql
@@ -22,4 +22,5 @@ CREATE OR REPLACE FUNCTION get_all_active_transactions(OUT database_id oid, OUT 
 	AS 'MODULE_PATHNAME', $$get_all_active_transactions$$;
  COMMENT ON FUNCTION get_all_active_transactions(OUT database_id oid, OUT process_id int, OUT initiator_node_identifier int4, OUT transaction_number int8, OUT transaction_stamp timestamptz)
      IS 'returns distributed transaction ids of active distributed transactions';
+
 RESET search_path;
--- a/src/backend/distributed/citus--7.0-4--7.0-5.sql
+++ b/src/backend/distributed/citus--7.0-4--7.0-5.sql
@@ -29,6 +29,7 @@ CREATE OR REPLACE VIEW pg_catalog.pg_dist_shard_placement AS
 CREATE OR REPLACE FUNCTION citus.pg_dist_node_trigger_func()
 RETURNS TRIGGER AS $$
   BEGIN
+    /* AddNodeMetadata also takes out a ShareRowExclusiveLock */
     LOCK TABLE pg_dist_node IN SHARE ROW EXCLUSIVE MODE;
     IF (TG_OP = 'INSERT') THEN
       IF NEW.noderole = 'primary'
--- a/src/backend/distributed/citus--7.0-4--7.0-5.sql
+++ b/src/backend/distributed/citus--7.0-4--7.0-5.sql
@@ -1,0 +1,97 @@
+/* citus--7.0-4--7.0-5.sql */
+
+SET search_path = 'pg_catalog';
+
+CREATE TYPE pg_catalog.noderole AS ENUM (
+  'primary',     -- node is available and accepting writes
+  'secondary',   -- node is available but only accepts reads
+  'unavailable' -- node is in recovery or otherwise not usable
+-- adding new values to a type inside of a transaction (such as during an ALTER EXTENSION
+-- citus UPDATE) isn't allowed in PG 9.6, and only allowed in PG10 if you don't use the
+-- new values inside of the same transaction. You might need to replace this type with a
+-- new one and then change the column type in pg_dist_node. There's a list of
+-- alternatives here:
+-- https://stackoverflow.com/questions/1771543/postgresql-updating-an-enum-type/41696273
+);
+
+ALTER TABLE pg_dist_node ADD COLUMN noderole noderole NOT NULL DEFAULT 'primary';
+
+-- we're now allowed to have more than one node per group
+ALTER TABLE pg_catalog.pg_dist_node DROP CONSTRAINT pg_dist_node_groupid_unique;
+
+-- so make sure pg_dist_shard_placement only returns writable placements
+CREATE OR REPLACE VIEW pg_catalog.pg_dist_shard_placement AS
+  SELECT shardid, shardstate, shardlength, nodename, nodeport, placementid
+  FROM pg_dist_placement placement INNER JOIN pg_dist_node node ON (
+    placement.groupid = node.groupid AND node.noderole = 'primary'
+  );
+
+CREATE OR REPLACE FUNCTION citus.pg_dist_node_trigger_func()
+RETURNS TRIGGER AS $$
+  BEGIN
+    LOCK TABLE pg_dist_node IN SHARE ROW EXCLUSIVE MODE;
+    IF (TG_OP = 'INSERT') THEN
+      IF NEW.noderole = 'primary'
+          AND EXISTS (SELECT 1 FROM pg_dist_node WHERE groupid = NEW.groupid AND
+                                                       noderole = 'primary' AND
+                                                       nodeid <> NEW.nodeid) THEN
+        RAISE EXCEPTION 'there cannot be two primary nodes in a group';
+      END IF;
+      RETURN NEW;
+    ELSIF (TG_OP = 'UPDATE') THEN
+      IF NEW.noderole = 'primary'
+           AND EXISTS (SELECT 1 FROM pg_dist_node WHERE groupid = NEW.groupid AND
+                                                        noderole = 'primary' AND
+                                                        nodeid <> NEW.nodeid) THEN
+         RAISE EXCEPTION 'there cannot be two primary nodes in a group';
+      END IF;
+      RETURN NEW;
+    END IF;
+  END;
+$$ LANGUAGE plpgsql;
+
+CREATE TRIGGER pg_dist_node_trigger
+  BEFORE INSERT OR UPDATE ON pg_dist_node
+  FOR EACH ROW EXECUTE PROCEDURE citus.pg_dist_node_trigger_func();
+
+DROP FUNCTION master_add_node(text, integer);
+CREATE FUNCTION master_add_node(nodename text,
+                                nodeport integer,
+                                groupid integer default 0,
+                                noderole noderole default 'primary',
+                                OUT nodeid integer,
+                                OUT groupid integer,
+                                OUT nodename text,
+                                OUT nodeport integer,
+                                OUT noderack text,
+                                OUT hasmetadata boolean,
+                                OUT isactive bool,
+                                OUT noderole noderole)
+  RETURNS record
+  LANGUAGE C STRICT
+  AS 'MODULE_PATHNAME', $$master_add_node$$;
+COMMENT ON FUNCTION master_add_node(nodename text, nodeport integer,
+                                    groupid integer, noderole noderole)
+  IS 'add node to the cluster';
+
+DROP FUNCTION master_add_inactive_node(text, integer);
+CREATE FUNCTION master_add_inactive_node(nodename text,
+                                         nodeport integer,
+                                         groupid integer default 0,
+                                         noderole noderole default 'primary',
+                                         OUT nodeid integer,
+                                         OUT groupid integer,
+                                         OUT nodename text,
+                                         OUT nodeport integer,
+                                         OUT noderack text,
+                                         OUT hasmetadata boolean,
+                                         OUT isactive bool,
+                                         OUT noderole noderole)
+  RETURNS record
+  LANGUAGE C STRICT
+  AS 'MODULE_PATHNAME',$$master_add_inactive_node$$;
+COMMENT ON FUNCTION master_add_inactive_node(nodename text,nodeport integer,
+                                             groupid integer, noderole noderole)
+  IS 'prepare node by adding it to pg_dist_node';
+
+RESET search_path;
--- a/src/backend/distributed/master/worker_node_manager.c
+++ b/src/backend/distributed/master/worker_node_manager.c
@@ -40,6 +40,7 @@ int MaxWorkerNodesTracked = 2048;    /* determines worker node hash table size *
 
 
 /* Local functions forward declarations */
+static WorkerNode * WorkerGetNodeWithName(const char *hostname);
 static char * ClientHostAddress(StringInfo remoteHostStringInfo);
 static WorkerNode * FindRandomNodeNotInList(HTAB *WorkerNodesHash,
 											List *currentNodeList);
@@ -276,7 +277,7 @@ ClientHostAddress(StringInfo clientHostStringInfo)
  * WorkerGetNodeWithName finds and returns a node from the membership list that
  * has the given hostname. The function returns null if no such node exists.
  */
-WorkerNode *
+static WorkerNode *
 WorkerGetNodeWithName(const char *hostname)
 {
 	WorkerNode *workerNode = NULL;
--- a/src/backend/distributed/master/worker_node_manager.c
+++ b/src/backend/distributed/master/worker_node_manager.c
@@ -42,8 +42,8 @@ int MaxWorkerNodesTracked = 2048;    /* determines worker node hash table size *
 /* Local functions forward declarations */
 static WorkerNode * WorkerGetNodeWithName(const char *hostname);
 static char * ClientHostAddress(StringInfo remoteHostStringInfo);
-static WorkerNode * FindRandomNodeNotInList(HTAB *WorkerNodesHash,
-											List *currentNodeList);
+static List * PrimaryNodesNotInList(List *currentList);
+static WorkerNode * FindRandomNodeFromList(List *candidateWorkerNodeList);
 static bool OddNumber(uint32 number);
 static bool ListMember(List *currentList, WorkerNode *workerNode);
 
@@ -54,9 +54,8 @@ static bool ListMember(List *currentList, WorkerNode *workerNode);
  */
 
 /*
- * WorkerGetRandomCandidateNode takes in a list of worker nodes, and then allocates
- * a new worker node. The allocation is performed by randomly picking a worker node
- * which is not in currentNodeList.
+ * WorkerGetRandomCandidateNode accepts a list of WorkerNode's and returns a random
+ * primary node which is not in that list.
  *
  * Note that the function returns null if the worker membership list does not
  * contain enough nodes to allocate a new worker node.
@@ -69,16 +68,11 @@ WorkerGetRandomCandidateNode(List *currentNodeList)
 	uint32 tryCount = WORKER_RACK_TRIES;
 	uint32 tryIndex = 0;
 
-	HTAB *workerNodeHash = GetWorkerNodeHash();
-
-	/*
-	 * We check if the shard has already been placed on all nodes known to us.
-	 * This check is rather defensive, and has the drawback of performing a full
-	 * scan over the worker node hash for determining the number of live nodes.
-	 */
 	uint32 currentNodeCount = list_length(currentNodeList);
-	uint32 liveNodeCount = WorkerGetLiveNodeCount();
-	if (currentNodeCount >= liveNodeCount)
+	List *candidateWorkerNodeList = PrimaryNodesNotInList(currentNodeList);
+
+	/* we check if the shard has already been placed on all nodes known to us */
+	if (list_length(candidateWorkerNodeList) == 0)
 	{
 		return NULL;
 	}
@@ -86,7 +80,7 @@ WorkerGetRandomCandidateNode(List *currentNodeList)
 	/* if current node list is empty, randomly pick one node and return */
 	if (currentNodeCount == 0)
 	{
-		workerNode = FindRandomNodeNotInList(workerNodeHash, NIL);
+		workerNode = FindRandomNodeFromList(candidateWorkerNodeList);
 		return workerNode;
 	}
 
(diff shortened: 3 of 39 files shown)


lithp:
Currently `master_add_node` has a default `nodeRole` of `unspecified` and allows it into the database. It should either:
- No longer be `STRICT` and check for a null nodeRole and fill it with `primary`
- Detect a `unspecified` nodeRole and replace it with `primary`.

Actually, it can probably accept a default of `primary`. I had thought there was some value in having Citus automatically decide whether a node should be a primary or not but probably the user always already knows which one it should be?

lithp:
Add tests:
- It's not possible to add more than one primary to a group
- You get a useful error message when running a query and there's no primary in the requested group
- It's possible to add multiple secondaries to a group

lithp:
Onder had a good point that `ALTER TYPE ... ADD VALUE` isn't allowed within a transaction by Citus. PG 10 added partial support for it, here are some notes for future reference:
- https://www.postgresql.org/message-id/4075.1459088427%40sss.pgh.pa.us (this entire thread)
- https://github.com/postgres/postgres/commit/15bc038f9bcd1a9af3f625caffafc7c20322202d
- https://stackoverflow.com/questions/1771543/postgresql-updating-an-enum-type/41696273

lithp:
To get the Oid of the enum type, maybe it makes more sense to call `makeTypeNameFromNameList` and `typenameTypeId` as found [here](https://github.com/citusdata/citus-ha/blob/master/src/replication_state.c#L37).

lithp:
@marcocitus I think this is ready for you to look at. I have a few questions around `isactive` and reference tables. This comes up in two functions:

- master_add_node probably doesn't work right now if you try to add a secondary while there are any reference tables. I can add a check which skips replication if you're not adding a primary, but then do we set `isactive` on the secondary? Maybe we set it to whatever `isactive` is on the primary?

- master_disable_node removes all reference tables from the group if you're removing a secondary. Which is totally wrong, but I'm not sure what the correct behavior is. Maybe we should no-op? Maybe we should error out if you try to run `master_disable_node` on anything but a primary?

I think the right way forward is to, eventually, add a `pg_dist_group` which keeps track of `isactive`. For now though that's a big change, so for now we just ignore whatever `isactive` is on the secondaries and require that things like auto-HA are careful to correctly set them when failing over?

src/backend/distributed/citus--7.0-3--7.0-4.sql
```diff
@@ -0,0 +1,82 @@
+/* citus--7.0-3--7.0-4.sql */
+
+SET search_path = 'pg_catalog';
+
+CREATE TYPE noderole AS ENUM (
```
marcocitus: Maybe add a comment that we need to replace the type if we want to add new roles.

src/test/regress/expected/multi_unsupported_worker_operations.out
```diff
@@ -220,8 +220,8 @@ SELECT master_add_node('localhost', 5432);
 ERROR:  operation is not allowed on this node
 HINT:  Connect to the coordinator and run it again.
 SELECT * FROM pg_dist_node WHERE nodename='localhost' AND nodeport=5432;
- nodeid | groupid | nodename | nodeport | noderack | hasmetadata | isactive 
---------+---------+----------+----------+----------+-------------+----------
+ nodeid | groupid | nodename | nodeport | noderack | hasmetadata | isactive | noderole 
+--------+---------+----------+----------+----------+-------------+----------+----------
 (0 rows)
```
marcocitus: Maybe switch to using count(*) instead of selecting all columns.

src/backend/distributed/metadata/metadata_sync.c
```diff
@@ -417,15 +417,35 @@ NodeListInsertCommand(List *workerNodeList)
...
+		if (workerNode->nodeRole != InvalidOid)
+		{
+			Datum nodeRoleOidDatum = ObjectIdGetDatum(workerNode->nodeRole);
+			Datum nodeRoleStringDatum = DirectFunctionCall1(enum_out, nodeRoleOidDatum);
+			nodeRoleString = DatumGetCString(nodeRoleStringDatum);
+		}
+		else
+		{
+			/*
+			 * Hack because master_initialize_node_metadata calls this before nodeRole
```
marcocitus: Is there a way to avoid this hack? (e.g. not calling `NodeListInsertCommand` if there are no workers with metadata in `master_initialize_node_metadata`). Workers with metadata cannot exist at the time `master_initialize_node_metadata` gets called in the extension script.

src/backend/distributed/master/worker_node_manager.c
```diff
@@ -342,70 +340,50 @@ ActiveWorkerNodeList(void)
...
-	 * assuming that the number of elements in the hash is N. We then get to
-	 * this random position by iterating over the worker hash. Please note that
-	 * the random seed has already been set by the postmaster when starting up.
-	 */
-	workerPosition = (random() % workerNodeCount) + 1;
-	hash_seq_init(&status, WorkerNodesHash);
-
-	for (workerIndex = 0; workerIndex < workerPosition; workerIndex++)
-	{
-		workerNode = (WorkerNode *) hash_seq_search(&status);
```
marcocitus: well, this was weird.

marcocitus:
Don't we also need to redefine `pg_dist_shard_placement` now?

lithp:
@marcocitus What do you mean by redefine `pg_dist_shard_placement`?

src/backend/distributed/utils/reference_table_utils.c
```diff
@@ -382,13 +382,13 @@ CreateReferenceTableColocationId()
...
- * DeleteAllReferenceTablePlacementsFromNode function iterates over list of reference
+ * DeleteAllReferenceTablePlacementsFromGroup function iterates over list of reference
  * tables and deletes all reference table placements from pg_dist_placement table
- * for given worker node. However, it does not modify replication factor of the colocation
+ * for given group. However, it does not modify replication factor of the colocation
  * group of reference tables. It is caller's responsibility to do that if it is necessary.
  */
 void
-DeleteAllReferenceTablePlacementsFromNode(char *workerName, uint32 workerPort)
+DeleteAllReferenceTablePlacementsFromGroup(uint32 groupId)
```
marcocitus: Maybe Group->NodeGroup.

I've been thinking that we've started using the word Group outside of the context of node-specific functions, and it might cause some confusion in the future.

src/include/distributed/master_metadata_utility.h
```diff
@@ -115,8 +115,7 @@ extern void CopyShardInterval(ShardInterval *srcInterval, ShardInterval *destInt
 extern void CopyShardPlacement(ShardPlacement *srcPlacement,
 							   ShardPlacement *destPlacement);
 extern uint64 ShardLength(uint64 shardId);
-extern bool NodeHasShardPlacements(char *nodeName, int32 nodePort,
-								   bool onlyLookForActivePlacements);
+extern bool GroupHasShardPlacements(uint32 groupId, bool onlyLookForActivePlacements);
```
marcocitus: Group->NodeGroup
onlyLookForActivePlacements->onlyConsiderActivePlacements

marcocitus:
> @marcocitus What do you mean by redefine pg_dist_shard_placement?

`pg_dist_shard_placement` is a view defined as: 
```
CREATE OR REPLACE VIEW pg_catalog.pg_dist_shard_placement AS
 SELECT placement.shardid,
    placement.shardstate,
    placement.shardlength,
    node.nodename,
    node.nodeport,
    placement.placementid
   FROM pg_dist_placement placement
     JOIN pg_dist_node node ON placement.groupid = node.groupid
```

This will currently include both secondaries and primaries. Not sure if that's the intended behaviour (it could be).

src/backend/distributed/utils/reference_table_utils.c
lithp: What else have we used the word Group for? Is there still confusion when you take the entire name into account? We're deleting reference tables placements from a group that seems pretty unambiguous and this method name in particular is already quite long.

marcocitus: We don't have groups of anything other than nodes yet, but group is such a generic term that it might confuse new readers.

I think maybe `GroupHasShardPlacements` would be slightly more clear as `NodeGroupHasShardPlacements`.

src/backend/distributed/utils/node_metadata.c
```diff
@@ -908,6 +990,19 @@ InsertNodeRow(int nodeid, char *nodeName, int32 nodePort, uint32 groupId, char *
 	values[Anum_pg_dist_node_hasmetadata - 1] = BoolGetDatum(hasMetadata);
 	values[Anum_pg_dist_node_isactive - 1] = BoolGetDatum(isActive);
 
+	if (nodeRole == InvalidOid)
+	{
+		/*
+		 * TODO: Make sure this will either error out or use the default once the
+		 * column exists
+		 */
+		isNulls[Anum_pg_dist_node_noderole - 1] = true;
```
marcocitus: is this meaningful, or could we just let it fall through to `ObjectIdGetDatum`?

src/backend/distributed/utils/node_metadata.c
```diff
@@ -620,27 +684,44 @@ AddNodeMetadata(char *nodeName, int32 nodePort, int32 groupId, char *nodeRack,
 		}
 	}
 
+	/* if nodeRole hasn't been added yet there's a constraint for one-node-per-group */
+	if (nodeRole != InvalidOid)
+	{
+		primaryRole = PrimaryNodeRoleId();
+		if (nodeRole == primaryRole)
```
lithp: Yeah, what's our convention here? We usually keep it separated but sometimes inline; I've assumed that it's to aid using gdb and tried to break it out whenever a future debugger might want to know the intermediate value, should I be using a different metric?

marcocitus: That's a good metric in general. I think here it'd be a little overkill, but consider the comment optional.

lithp:
Believe this last set of changes covers all the feedback

src/backend/distributed/utils/node_metadata.c
```diff
@@ -620,27 +683,43 @@ AddNodeMetadata(char *nodeName, int32 nodePort, int32 groupId, char *nodeRack,
...
+	primariesWithMetadata = CountPrimariesWithMetadata();
+	if (primariesWithMetadata != 0)
+	{
+		List *workerNodeList = list_make1(workerNode);
+		char *nodeInsertCommand = NodeListInsertCommand(workerNodeList);
+		SendCommandToWorkers(WORKERS_WITH_METADATA, nodeInsertCommand);
+	}
 
-	heap_close(pgDistNode, AccessExclusiveLock);
+	heap_close(pgDistNode, NoLock);
```
marcocitus: Should be careful with this. The function takes an access exclusive lock on pg_dist_node and now keeps it until commit, which blocks all queries. Maybe take an ExclusiveLock.

lithp:
@marcocitus I just realized my squash made my last few changes difficult to review, sorry about that! I made sure the node functions play nicely with `isactive` and secondaries. [Here](https://github.com/citusdata/citus/pull/1476/files#diff-39f91ee6a73924e721dea75995fac24dR455) and [here](https://github.com/citusdata/citus/pull/1476/files#diff-5dedd53aec4942d4a95f3edcb25f3456R60) are the relevant tests.

src/backend/distributed/citus--7.0-4--7.0-5.sql
```diff
@@ -0,0 +1,95 @@
...
+
+CREATE OR REPLACE FUNCTION citus.pg_dist_node_trigger_func()
+RETURNS TRIGGER AS $$
+  BEGIN
+    IF (TG_OP = 'INSERT') THEN
+	IF NEW.noderole = 'primary'
+	   AND EXISTS (SELECT 1 FROM pg_dist_node WHERE groupid = NEW.groupid AND
+		                                        noderole = 'primary' AND
+						        nodeid <> NEW.nodeid) THEN
+	  RAISE EXCEPTION 'there cannot be two primary nodes in a group';
```
marcocitus: Without some kind of table lock, there is a race condition here when two queries try to add a primary key at the same time.

src/test/regress/expected/multi_cluster_management.out
```diff
@@ -180,6 +180,35 @@ SELECT 1 FROM master_add_node('localhost', :worker_2_port);
 (1 row)
 
 UPDATE pg_dist_placement SET shardstate=1 WHERE groupid=:worker_2_group;
+-- when there is no primary we should get a pretty error
+UPDATE pg_dist_node SET noderole = 'secondary' WHERE nodeport=:worker_2_port;
+SELECT * FROM cluster_management_test;
+ERROR:  group 3 does not have a primary node
+DETAIL:  the primary node may be in the process of failing over
```
marcocitus: Maybe "node group 3 does not have a primary", since it's not entirely clear what "group" refers to.

The DETAIL is perhaps a little premature.

src/test/regress/expected/multi_cluster_management.out
```diff
@@ -164,7 +164,7 @@ SELECT shardid, shardstate, nodename, nodeport FROM pg_dist_shard_placement WHER
 
 -- try to remove a node with only inactive placements and see that removal still fails
 SELECT master_remove_node('localhost', :worker_2_port); 
-ERROR:  you cannot remove a node which has shard placements
+ERROR:  you cannot remove the primary node of a group which has shard placements
```
marcocitus: maybe "primary node of a group" -> "primary of a node group", since it's not entirely clear what "group" refers to.

lithp:
Good points. I tried my hardest to come up with a better way but think the lock is by far the easiest answer, and you won't often be modifying `pg_dist_node` so serializing writes is annoying but reasonable.

lithp:
Ready for you again.

src/backend/distributed/citus--7.0-4--7.0-5.sql
```diff
@@ -0,0 +1,96 @@
...
+CREATE OR REPLACE VIEW pg_catalog.pg_dist_shard_placement AS
+SELECT shardid, shardstate, shardlength, nodename, nodeport, placementid
+FROM pg_dist_placement placement INNER JOIN pg_dist_node node ON (
+  placement.groupid = node.groupid AND node.noderole = 'primary'
+);
+
+CREATE OR REPLACE FUNCTION citus.pg_dist_node_trigger_func()
+RETURNS TRIGGER AS $$
+  BEGIN
+    LOCK TABLE pg_dist_node IN SHARE ROW EXCLUSIVE MODE;
```
marcocitus: Should we take the same lock in AddNodeMetadata then? (and not release it)

src/backend/distributed/utils/node_metadata.c
```diff
@@ -620,27 +706,43 @@ AddNodeMetadata(char *nodeName, int32 nodePort, int32 groupId, char *nodeRack,
 		}
```
marcocitus: The code above seems odd and would forbid explicitly specifying a group.

src/backend/distributed/utils/node_metadata.c
```diff
@@ -620,27 +706,43 @@ AddNodeMetadata(char *nodeName, int32 nodePort, int32 groupId, char *nodeRack,
...
+	primariesWithMetadata = CountPrimariesWithMetadata();
+	if (primariesWithMetadata != 0)
+	{
+		List *workerNodeList = list_make1(workerNode);
+		char *nodeInsertCommand = NodeListInsertCommand(workerNodeList);
+		SendCommandToWorkers(WORKERS_WITH_METADATA, nodeInsertCommand);
+	}
 
-	heap_close(pgDistNode, AccessExclusiveLock);
+	heap_close(pgDistNode, NoLock);
```
marcocitus: We should take a ShareRowExclusive lock in the `heap_open`.

src/backend/distributed/utils/node_metadata.c
```diff
@@ -907,6 +1010,7 @@ InsertNodeRow(int nodeid, char *nodeName, int32 nodePort, uint32 groupId, char *
 	values[Anum_pg_dist_node_noderack - 1] = CStringGetTextDatum(nodeRack);
 	values[Anum_pg_dist_node_hasmetadata - 1] = BoolGetDatum(hasMetadata);
 	values[Anum_pg_dist_node_isactive - 1] = BoolGetDatum(isActive);
+	values[Anum_pg_dist_node_noderole - 1] = ObjectIdGetDatum(nodeRole);
 
 	/* open shard relation and insert new tuple */
 	pgDistNode = heap_open(DistNodeRelationId(), AccessExclusiveLock);
```
marcocitus: RowExclusiveLock

src/backend/distributed/utils/node_metadata.c
lithp: True, I left the code how it was originally written. If I had to guess, the motivation was to make sure that not specifying always worked. By doing this we guarantee that `nextval(pg_dist_groupid_seq)` always returns an unused groupid.

lithp:
I went through node_metadata and made some changes:
- Never release locks
- `AddNodeMetadata` and `InsertNodeRow` now take out `ShareRowExclusiveLock` so they'll conflict with the trigger
- Functions which modify the table now take out `RowExclusiveLock` instead of `AccessShareLock`. (Specifically, `ActivateNode` and `SetNodeState`)
-  `DeleteNodeRow` no longer takes `AccessExclusiveLock`, it only needs `RowExclusive`.

src/backend/distributed/utils/node_metadata.c
```diff
@@ -674,21 +674,15 @@ AddNodeMetadata(char *nodeName, int32 nodePort, int32 groupId, char *nodeRack,
...
-		/* fill return data and return */
-		returnData = GenerateNodeTuple(workerNode);
-
-		/* close the heap */
-		heap_close(pgDistNode, AccessExclusiveLock);
-
 		*nodeAlreadyExists = true;
-
-		PG_RETURN_DATUM(returnData);
+		goto finish;
```
marcocitus: goto is the best way to get there ;)

maybe just use `LockRelationOid(DistNodeRelationId(), ShareRowExclusiveLock)` and avoid the `heap_close` (and the `goto`).

src/backend/distributed/utils/node_metadata.c
```diff
@@ -1013,7 +1007,7 @@ InsertNodeRow(int nodeid, char *nodeName, int32 nodePort, uint32 groupId, char *
 	values[Anum_pg_dist_node_noderole - 1] = ObjectIdGetDatum(nodeRole);
 
 	/* open shard relation and insert new tuple */
-	pgDistNode = heap_open(DistNodeRelationId(), AccessExclusiveLock);
+	pgDistNode = heap_open(DistNodeRelationId(), ShareRowExclusiveLock);
```
marcocitus: Not very consequential, but we could probably just use `RowExclusiveLock` here since there's nothing in this function that specifically requires a `ShareRowExclusiveLock` (the caller already has one anyway).

lithp:
Rebased onto https://github.com/citusdata/citus/pull/1507 so that needs to be merged first.

src/backend/distributed/utils/node_metadata.c
lithp: That's super convenient, good point!

src/backend/distributed/citus--7.0-4--7.0-5.sql
```diff
@@ -0,0 +1,97 @@
...
+    /* AddNodeMetadata also takes out a ShareRowExclusiveLock */
+    LOCK TABLE pg_dist_node IN SHARE ROW EXCLUSIVE MODE;
+    IF (TG_OP = 'INSERT') THEN
+	IF NEW.noderole = 'primary'
+	   AND EXISTS (SELECT 1 FROM pg_dist_node WHERE groupid = NEW.groupid AND
+		                                        noderole = 'primary' AND
+						        nodeid <> NEW.nodeid) THEN
+	  RAISE EXCEPTION 'there cannot be two primary nodes in a group';
+	END IF;
+	RETURN NEW;
```
marcocitus: Alignment looks weird here (mix of tabs and spaces?)

src/backend/distributed/utils/node_metadata.c
```diff
@@ -572,37 +660,30 @@ RemoveNodeFromCluster(char *nodeName, int32 nodePort)
...
+	uint32 primariesWithMetadata = 0;
 
 	EnsureCoordinator();
 	EnsureSuperUser();
 
 	*nodeAlreadyExists = false;
 
 	/* acquire a lock so that no one can do this concurrently */
-	pgDistNode = heap_open(DistNodeRelationId(), AccessExclusiveLock);
+	/* specifically, pg_dist_node_trigger_func also takes out a ShareRowExclusiveLock */
```
marcocitus: Maybe merge the comments

PR #1477, opened 2017-07-04
Remove serialization/deserialization of multiplan node

mtuncer:
Serialization/Deserialization of ```MultiPlan```s take significantly long time when plans get large. In terms of columns in the relations, and number of worker queries.

We can use ExtensibleNode mechanism introduced in PG 9.6 to avoid MultiPlan serialization.
This work is almost %100 based on @anarazel`s prototype.

Fixes : #703

### DIFF
--- a/src/backend/distributed/executor/multi_server_executor.c
+++ b/src/backend/distributed/executor/multi_server_executor.c
@@ -170,7 +170,8 @@ InitTaskExecution(Task *task, TaskExecStatus initialTaskExecStatus)
 	uint32 nodeCount = list_length(task->taskPlacementList);
 	uint32 nodeIndex = 0;
 
-	TaskExecution *taskExecution = palloc0(sizeof(TaskExecution));
+	TaskExecution *taskExecution = CitusMakeNode(TaskExecution);
+
 	taskExecution->jobId = task->jobId;
 	taskExecution->taskId = task->taskId;
 	taskExecution->nodeCount = nodeCount;
@@ -235,7 +236,7 @@ CleanupTaskExecution(TaskExecution *taskExecution)
 	pfree(taskExecution->taskStatusArray);
 	pfree(taskExecution->connectionIdArray);
 	pfree(taskExecution->fileDescriptorArray);
-	memset(taskExecution, 0, sizeof(TaskExecution));
+	pfree(taskExecution);
 }
 
 
--- a/src/backend/distributed/planner/multi_planner.c
+++ b/src/backend/distributed/planner/multi_planner.c
@@ -70,13 +70,12 @@ static PlannedStmt * CreateDistributedPlan(PlannedStmt *localPlan, Query *origin
 										   plannerRestrictionContext);
 static void AssignRTEIdentities(Query *queryTree);
 static void AssignRTEIdentity(RangeTblEntry *rangeTableEntry, int rteIdentifier);
-static Node * SerializeMultiPlan(struct MultiPlan *multiPlan);
-static MultiPlan * DeserializeMultiPlan(Node *node);
 static PlannedStmt * FinalizePlan(PlannedStmt *localPlan, MultiPlan *multiPlan);
 static PlannedStmt * FinalizeNonRouterPlan(PlannedStmt *localPlan, MultiPlan *multiPlan,
 										   CustomScan *customScan);
 static PlannedStmt * FinalizeRouterPlan(PlannedStmt *localPlan, CustomScan *customScan);
 static void CheckNodeIsDumpable(Node *node);
+static Node * CheckNodeCopyAndSerialization(Node *node);
 static List * CopyPlanParamList(List *originalPlanParamList);
 static PlannerRestrictionContext * CreateAndPushPlannerRestrictionContext(void);
 static PlannerRestrictionContext * CurrentPlannerRestrictionContext(void);
@@ -412,60 +411,17 @@ CreateDistributedPlan(PlannedStmt *localPlan, Query *originalQuery, Query *query
 MultiPlan *
 GetMultiPlan(CustomScan *customScan)
 {
+	Node *node = NULL;
 	MultiPlan *multiPlan = NULL;
 
 	Assert(list_length(customScan->custom_private) == 1);
 
-	multiPlan = DeserializeMultiPlan(linitial(customScan->custom_private));
+	node = (Node *) linitial(customScan->custom_private);
+	Assert(CitusIsA(node, MultiPlan));
 
-	return multiPlan;
-}
+	node = CheckNodeCopyAndSerialization(node);
 
-
-/*
- * SerializeMultiPlan returns the string representing the distributed plan in a
- * Const node.
- *
- * Note that this should be improved for 9.6+, we we can copy trees efficiently.
- * I.e. we should introduce copy support for relevant node types, and just
- * return the MultiPlan as-is for 9.6.
- */
-static Node *
-SerializeMultiPlan(MultiPlan *multiPlan)
-{
-	char *serializedMultiPlan = NULL;
-	Const *multiPlanData = NULL;
-
-	serializedMultiPlan = nodeToString(multiPlan);
-
-	multiPlanData = makeNode(Const);
-	multiPlanData->consttype = CSTRINGOID;
-	multiPlanData->constlen = strlen(serializedMultiPlan);
-	multiPlanData->constvalue = CStringGetDatum(serializedMultiPlan);
-	multiPlanData->constbyval = false;
-	multiPlanData->location = -1;
-
-	return (Node *) multiPlanData;
-}
-
-
-/*
- * DeserializeMultiPlan returns the deserialized distributed plan from the string
- * representation in a Const node.
- */
-static MultiPlan *
-DeserializeMultiPlan(Node *node)
-{
-	Const *multiPlanData = NULL;
-	char *serializedMultiPlan = NULL;
-	MultiPlan *multiPlan = NULL;
-
-	Assert(IsA(node, Const));
-	multiPlanData = (Const *) node;
-	serializedMultiPlan = DatumGetCString(multiPlanData->constvalue);
-
-	multiPlan = (MultiPlan *) stringToNode(serializedMultiPlan);
-	Assert(CitusIsA(multiPlan, MultiPlan));
+	multiPlan = (MultiPlan *) node;
 
 	return multiPlan;
 }
@@ -521,7 +477,7 @@ FinalizePlan(PlannedStmt *localPlan, MultiPlan *multiPlan)
 		}
 	}
 
-	multiPlanData = SerializeMultiPlan(multiPlan);
+	multiPlanData = (Node *) multiPlan;
 
 	customScan->custom_private = list_make1(multiPlanData);
 	customScan->flags = CUSTOMPATH_SUPPORT_BACKWARD_SCAN;
@@ -662,6 +618,36 @@ CheckNodeIsDumpable(Node *node)
 #ifdef USE_ASSERT_CHECKING
 	char *out = nodeToString(node);
 	pfree(out);
+#endif
+}
+
+
+/*
+ * CheckNodeCopyAndSerialization checks copy/dump/read functions
+ * for nodes and returns copy of the input.
+ *
+ * It is only active when assertions are enabled, otherwise it returns
+ * the input directly. We use this to confirm that our serialization
+ * and copy logic produces the correct plan during regression tests.
+ *
+ * It does not check string equality on node dumps due to differences
+ * in some Postgres types.
+ */
+static Node *
+CheckNodeCopyAndSerialization(Node *node)
+{
+#ifdef USE_ASSERT_CHECKING
+	char *out = nodeToString(node);
+	Node *deserializedNode = (Node *) stringToNode(out);
+	Node *nodeCopy = copyObject(deserializedNode);
+	char *outCopy = nodeToString(nodeCopy);
+
+	pfree(out);
+	pfree(outCopy);
+
+	return nodeCopy;
+#else
+	return node;
 #endif
 }
 
--- a/src/backend/distributed/utils/citus_copyfuncs.c
+++ b/src/backend/distributed/utils/citus_copyfuncs.c
@@ -1,0 +1,272 @@
+/*-------------------------------------------------------------------------
+ *
+ * citus_copyfuncs.c
+ *    Citus specific node copy functions
+ *
+ * Portions Copyright (c) 1996-2014, PostgreSQL Global Development Group
+ * Portions Copyright (c) 1994, Regents of the University of California
+ * Portions Copyright (c) 2012-2017, Citus Data, Inc.
+ *
+ *-------------------------------------------------------------------------
+ */
+#include "postgres.h"
+
+
+#include "distributed/citus_nodefuncs.h"
+#include "distributed/multi_server_executor.h"
+#include "utils/datum.h"
+
+
+/*
+ * Macros to simplify copying of different kinds of fields.  Use these
+ * wherever possible to reduce the chance for silly typos.  Note that these
+ * hard-wire the convention that the local variables in a Copy routine are
+ * named 'newnode' and 'from'.
+ */
+static inline Node *
+CitusSetTag(Node *node, int tag)
+{
+	CitusNode *citus_node = (CitusNode *) node;
+	citus_node->citus_tag = tag;
+	return node;
+}
+
+
+#define DECLARE_FROM_AND_NEW_NODE(nodeTypeName) \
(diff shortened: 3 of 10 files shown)


src/backend/distributed/planner/multi_planner.c
```diff
@@ -412,60 +410,15 @@ CreateDistributedPlan(PlannedStmt *localPlan, Query *originalQuery, Query *query
...
-
-	Assert(IsA(node, Const));
-	multiPlanData = (Const *) node;
-	serializedMultiPlan = DatumGetCString(multiPlanData->constvalue);
-
-	multiPlan = (MultiPlan *) stringToNode(serializedMultiPlan);
-	Assert(CitusIsA(multiPlan, MultiPlan));
+	multiPlan = (MultiPlan *) node;
 
 	return multiPlan;
```
anarazel: I think to get the coverage up we should do something similar to `CheckNodeIsDumpable()`, namely to, when asserts are enabled, the objects. This change also reduces the coverage of existing CheckNodeIsDumpable() calls.

src/backend/distributed/utils/citus_copyfuncs.c
```diff
@@ -0,0 +1,239 @@
...
+	CITUS_COPY_LOCALS(DeferredErrorMessage);
+
+	COPY_SCALAR_FIELD(code);
+	COPY_STRING_FIELD(message);
+	COPY_STRING_FIELD(detail);
+	COPY_STRING_FIELD(hint);
+	COPY_STRING_FIELD(filename);
+	COPY_SCALAR_FIELD(linenumber);
+	COPY_STRING_FIELD(functionname);
+}
```
anarazel: I wrote these in a few minutes, did you check whether I've got the fields to be copied right in these functions?

mtuncer: Yes I checked, there were some fields missing, added them.

There is only one that is questioable
 ```	COPY_NODE_FIELD(taskExecution);```

TaskExecution is not a node field. It is a plain struct  defined [here](https://github.com/citusdata/citus/blob/master/src/include/distributed/multi_server_executor.h#L117). ReadFuncs and OutFuncs also use node functions for this one.

src/backend/distributed/planner/multi_planner.c
mtuncer: Do you mean adding ifdefs inside unused  (ShardInterval, MapMergeJob) copy functions ?

src/backend/distributed/utils/citus_copyfuncs.c
anarazel: Huh. That seems weird / buggy. I think the right fix would be to make TaskExecution a node field too?  Are we currently hitting this code path anywhere?

mtuncer: We just got lucky.

taskExecution field is always null inside OutTask() and CopyNodeTask() functions.  ReadTask() function is never called at all.

mtuncer: @anarazel  I made TaskExecution a node as you suggested. But due to how it is declared, I had to include multi_server_executor.h at multiple places. I did not like it.

src/backend/distributed/planner/multi_planner.c
anarazel: No, not at all. That'd have no benefits, and downsides (no compiler checks).  What I want is that we *force* copying *and* out/readfuncs usage when asserts are enabled. By calling a function which at least copies nodes, and writes them out. Potentially also reads them back in.

src/backend/distributed/utils/citus_copyfuncs.c
anarazel: > @anarazel I made TaskExecution a node as you suggested. But due to how it is declared, I had to include multi_server_executor.h at multiple places. I did not like it.

Don't see the issue with that?

src/include/distributed/citus_nodefuncs.h
```diff
@@ -47,6 +49,7 @@ extern READFUNC_RET ReadMapMergeJob(READFUNC_ARGS);
 extern READFUNC_RET ReadShardPlacement(READFUNC_ARGS);
 extern READFUNC_RET ReadRelationShard(READFUNC_ARGS);
 extern READFUNC_RET ReadTask(READFUNC_ARGS);
+READFUNC_RET ReadTaskExecution(READFUNC_ARGS);
```
anarazel: missing extern.

src/backend/distributed/planner/multi_planner.c
```diff
@@ -667,6 +624,30 @@ CheckNodeIsDumpable(Node *node)
...
+ * for nodes.
+ *
+ * It does not check string equality on node dumps due to differences
+ * in some Postgres types.
+ */
+static void
+CheckNodeCopyAndSerialization(Node *node)
+{
+#ifdef USE_ASSERT_CHECKING
+#if 0
```
anarazel: Huh?

src/backend/distributed/utils/citus_readfuncs.c
```diff
@@ -331,6 +357,40 @@ ReadTask(READFUNC_ARGS)
...
+
+
+    READ_ENUM_ARRAY(taskStatusArray, count, TaskExecStatus);
+    READ_ENUM_ARRAY(transmitStatusArray, count, TransmitExecStatus);
+    READ_INT_ARRAY(connectionIdArray, count);
+    READ_INT_ARRAY(fileDescriptorArray, count);
+
+
+	READ_INT64_FIELD(connectStartTime);
+	READ_UINT_FIELD(currentNodeIndex);
```
anarazel: Weird indentation.

src/backend/distributed/planner/multi_planner.c
```diff
@@ -70,13 +70,12 @@ static PlannedStmt * CreateDistributedPlan(PlannedStmt *localPlan, Query *origin
...
 static void AssignRTEIdentities(Query *queryTree);
 static void AssignRTEIdentity(RangeTblEntry *rangeTableEntry, int rteIdentifier);
-static Node * SerializeMultiPlan(struct MultiPlan *multiPlan);
-static MultiPlan * DeserializeMultiPlan(Node *node);
 static PlannedStmt * FinalizePlan(PlannedStmt *localPlan, MultiPlan *multiPlan);
 static PlannedStmt * FinalizeNonRouterPlan(PlannedStmt *localPlan, MultiPlan *multiPlan,
 										   CustomScan *customScan);
 static PlannedStmt * FinalizeRouterPlan(PlannedStmt *localPlan, CustomScan *customScan);
 static void CheckNodeIsDumpable(Node *node);
+static void CheckNodeCopyAndSerialization(Node *node);
```
anarazel: I don't understand why we now have both of these? We should only have the latter (doing both).

src/backend/distributed/planner/multi_planner.c
```diff
@@ -667,6 +624,30 @@ CheckNodeIsDumpable(Node *node)
...
 
 /*
+ * CheckNodeCopyAndSerialization checks copy/dump/read functions
+ * for nodes.
+ *
+ * It does not check string equality on node dumps due to differences
+ * in some Postgres types.
+ */
+static void
+CheckNodeCopyAndSerialization(Node *node)
```
anarazel: I think we should return the copy, and use that - that enforces the copy is actually correct.

src/backend/distributed/planner/multi_planner.c
mtuncer: debugging leftover 🤕

src/backend/distributed/planner/multi_planner.c
mtuncer: sounds good

src/backend/distributed/planner/multi_planner.c
mtuncer: removed CheckNodeIsDumpable

mtuncer: Added CheckNodeIsDumpable back.  It was also called for MultiTree nodes which only have dump function.

src/backend/distributed/utils/citus_readfuncs.c
mtuncer: done

mtuncer:
@anarazel I rebased it to current master. Coverage is somewhat low because TaskExecution field is always NULL when copy is needed.

mtuncer:
@anarazel could you take a look. It will make porting this to earlier version easier.

src/backend/distributed/planner/multi_planner.c
```diff
@@ -667,6 +623,35 @@ CheckNodeIsDumpable(Node *node)
 
 
 /*
+ * CheckNodeCopyAndSerialization checks copy/dump/read functions
+ * for nodes and returns copy of the input.
+ *
+ * It is only active when assertions are enabled, otherwise it returns
+ * the input directly.
```
marcocitus: Maybe add some note on why we're doing this rather silly thing. e.g.

"we use this to confirm that our serialization and copy logic produces the correct plan during regression tests"

src/backend/distributed/utils/citus_copyfuncs.c
```diff
@@ -0,0 +1,287 @@
...
+	do { \
+		Size _size = (sz); \
+		newnode->fldname = palloc(_size); \
+		memcpy(newnode->fldname, from->fldname, _size); \
+	} \
+	while (0)
+
+/* Copy a parse location field (for Copy, this is same as scalar case) */
+#define COPY_LOCATION_FIELD(fldname) \
+	(newnode->fldname = from->fldname)
```
marcocitus: we don't use some of these

src/backend/distributed/utils/citus_copyfuncs.c
```diff
@@ -0,0 +1,287 @@
...
+
+	arrayLength = from->sortedShardIntervalArrayLength;
+
+	/* now build & read sortedShardIntervalArray */
+	newnode->sortedShardIntervalArray =
+		(ShardInterval **) palloc(arrayLength * sizeof(ShardInterval *));
+
+	for (i = 0; i < arrayLength; ++i)
+	{
+		/* can't use COPY_NODE_FIELD, no field names */
```
marcocitus: maybe introduce a `COPY_NODE_ARRAY`?

src/backend/distributed/utils/citus_outfuncs.c
```diff
@@ -418,6 +445,34 @@ OutTask(OUTFUNC_ARGS)
 
 
 void
+OutTaskExecution(OUTFUNC_ARGS)
+{
+	int count = 0;
+	WRITE_LOCALS(TaskExecution);
+	WRITE_NODE_TYPE("TASKEXECUTION");
+
+	count = node->nodeCount;
```
marcocitus: `count` is a bit generic. Maybe we can just use `node->nodeCount` directly in `WRITE_ENUM_ARRAY`?

src/backend/distributed/utils/citus_readfuncs.c
```diff
@@ -119,6 +126,23 @@ CitusSetTag(Node *node, int tag)
...
+		for (i = 0; i < count; i++ ) \
+		{ \
+			token = pg_strtok(&length);		/* get field value */ \
+			local_node->fldname[i] = (enumtype) atoi(token); \
+		} \
+    } \
+	token = pg_strtok(&length);   /* skip ) */ \
+	(void) token
+
+#define READ_INT_ARRAY(fldname, count) READ_ENUM_ARRAY(fldname, count, int32)
```
marcocitus: This definition would make slightly more sense if it was the other way around.

src/backend/distributed/utils/citus_copyfuncs.c
```diff
@@ -0,0 +1,287 @@
...
+static inline Node *
+CitusSetTag(Node *node, int tag)
+{
+	CitusNode *citus_node = (CitusNode *) node;
+	citus_node->citus_tag = tag;
+	return node;
+}
+
+
+#define CITUS_COPY_LOCALS(nodeTypeName) \
```
mtuncer: ok

src/backend/distributed/utils/citus_copyfuncs.c
mtuncer: unused ones are removed

src/backend/distributed/utils/citus_copyfuncs.c
mtuncer: done

src/backend/distributed/utils/citus_outfuncs.c
mtuncer: ok

src/backend/distributed/utils/citus_outfuncs.c
```diff
@@ -418,6 +445,34 @@ OutTask(OUTFUNC_ARGS)
...
+
+	WRITE_UINT64_FIELD(jobId);
+	WRITE_UINT_FIELD(taskId);
+
+	WRITE_UINT_FIELD(nodeCount);
+
+    WRITE_ENUM_ARRAY(taskStatusArray, count);
+    WRITE_ENUM_ARRAY(transmitStatusArray, count);
```
mtuncer: done

src/backend/distributed/utils/citus_readfuncs.c
mtuncer: I could not find a better naming. Type cast in following line
```c
local_node->fldname[i] = (enumtype) atoi(token); \
```
requires that type is present (enum type or any signed integer).  Therefore we can't use READ_INT_ARRAY code inside READ_ENUM_ARRAY. reading int array is just a special case of reading a generic integer type with specific type in mind.

mtuncer:
@marcocitus could you make another pass. thanks

src/backend/distributed/executor/multi_server_executor.c
```diff
@@ -235,7 +236,7 @@ CleanupTaskExecution(TaskExecution *taskExecution)
 	pfree(taskExecution->taskStatusArray);
 	pfree(taskExecution->connectionIdArray);
 	pfree(taskExecution->fileDescriptorArray);
-	memset(taskExecution, 0, sizeof(TaskExecution));
+	pfree(taskExecution);
```
anarazel: Note this is not equivalent - have you checked the callers that there's no references to the task execution itself after this?

src/backend/distributed/utils/citus_copyfuncs.c
```diff
@@ -0,0 +1,272 @@
...
+	citus_node->citus_tag = tag;
+	return node;
+}
+
+
+#define DECLARE_FROM_AND_NEW_NODE(nodeTypeName) \
+	nodeTypeName * newnode = (nodeTypeName *) \
+							 CitusSetTag((Node *) target_node, T_ ## nodeTypeName); \
+	nodeTypeName *from = (nodeTypeName *) source_node
+
```
anarazel: Inconsistent spacing around the `*`.

src/backend/distributed/utils/citus_readfuncs.c
```diff
@@ -331,6 +357,38 @@ ReadTask(READFUNC_ARGS)
 
 
 READFUNC_RET
+ReadTaskExecution(READFUNC_ARGS)
+{
+	READ_LOCALS(TaskExecution);
+	int count = 0;
+
+	READ_UINT64_FIELD(jobId);
```
anarazel: If you want I think it'd be ok to replace the read function with an error throwing one - outputting is useful for debugging, but there's little reason to allow deserialization.

src/backend/distributed/executor/multi_server_executor.c
mtuncer: This function is called towards the end of ```MultiRealTimeExecute()``` and ```TrackerCleanupResources()```

There are no references left pointing here.

src/backend/distributed/utils/citus_copyfuncs.c
mtuncer: citus_indent should catch those things

src/backend/distributed/utils/citus_readfuncs.c
mtuncer: ok

src/backend/distributed/utils/citus_copyfuncs.c
mtuncer: correction: it was the citus_indent who formatted this way

metdos:
@anarazel it looks like you approved this PR but also had minor comments. Are you okay with merging this or do you want to take a final look?

anarazel:
> @anarazel it looks like you approved this PR but also had minor comments. Are you okay with merging this or do you want to take a final look?

Ok with merging this after the changes are done, unless questions arise while addressing them.

PR #1478, opened 2017-07-04
Replace duplicate code and call check_functions_in_node

mtuncer:
MasterIrreducibleExpressionWalker has a copied code from function check_functions_in_node() which was available with PG 9.6+. 

Now PG 9.5 support is dropped we can remove duplicate code and directly call check_functions_in_node().

Partial fix for #883

### DIFF
--- a/src/backend/distributed/planner/multi_router_planner.c
+++ b/src/backend/distributed/planner/multi_router_planner.c
@@ -90,7 +90,7 @@ static MultiPlan * CreateSingleTaskRouterPlan(Query *originalQuery,
 static bool MasterIrreducibleExpression(Node *expression, bool *varArgument,
 										bool *badCoalesce);
 static bool MasterIrreducibleExpressionWalker(Node *expression, WalkerState *state);
-static char MostPermissiveVolatileFlag(char left, char right);
+static bool MasterIrreducibleExpressionFunctionChecker(Oid func_id, void *context);
 static bool TargetEntryChangesValue(TargetEntry *targetEntry, Var *column,
 									FromExpr *joinTree);
 static Task * RouterModifyTask(Oid distributedTableId, Query *originalQuery,
@@ -833,6 +833,7 @@ MasterIrreducibleExpressionWalker(Node *expression, WalkerState *state)
 	char volatileFlag = 0;
 	WalkerState childState = { false, false, false };
 	bool containsDisallowedFunction = false;
+	bool hasVolatileFunction PG_USED_FOR_ASSERTS_ONLY = false;
 
 	if (expression == NULL)
 	{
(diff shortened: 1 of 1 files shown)


src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1024,6 +930,18 @@ MasterIrreducibleExpressionWalker(Node *expression, WalkerState *state)
 }
 
 
+static bool
+MasterIrreducibleExpressionFunctionChecker(Oid func_id, void *context)
```
anarazel: Doesn't declaring this directly as `bool *volatileContext` work?

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1024,6 +930,18 @@ MasterIrreducibleExpressionWalker(Node *expression, WalkerState *state)
 }
 
 
+static bool
```
anarazel: comment?

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1024,6 +930,18 @@ MasterIrreducibleExpressionWalker(Node *expression, WalkerState *state)
...
+MasterIrreducibleExpressionFunctionChecker(Oid func_id, void *context)
+{
+	char volatileFlag = func_volatile(func_id);
+	char *volatileContext = (char *) context;
+
+	*volatileContext = MostPermissiveVolatileFlag(*volatileContext, volatileFlag);
+
+	return (volatileFlag == PROVOLATILE_VOLATILE);
+}
```
anarazel: Hm, there's no other `MostPermissiveVolatileFlag()` callers. And this looks complicated.  I think I'd just check if `*volatileContext` is true and return, and only otherwise check if `func_volatile() == PROVOLATILE_VOLATILE``

mtuncer: ok

src/backend/distributed/planner/multi_router_planner.c
mtuncer: volatileContext is a tri-state variable.
1. volatile
2. stable 
3. else (immutable)

Walker functions checks (1) if it has volatile function, (2) if it has stable function. Only a boolean value is not enough for this comparison.

The caller ```check_functions_in_node()``` does not call this function again once it returns true.  So the context  variable can not be volatile ever. We can use this value to indicate if it has a stable function or not. 

So the fix would be replacing ```volatileFlag``` in walker function (```MasterIrreducibleExpressionWalker```) with something like ```mostPermissiveFuntionIsStable```

I did not like this approach though.

src/backend/distributed/planner/multi_router_planner.c
mtuncer: ok

src/backend/distributed/planner/multi_router_planner.c
mtuncer: > Doesn't declaring this directly as bool *volatileContext work?

No it does not.  See definitions in nodeFuncs.h

```c
typedef bool (*check_function_callback) (Oid func_id, void *context);

...

extern bool check_functions_in_node(Node *node, check_function_callback checker,
				    void *context);

extern bool expression_tree_walker(Node *node, bool (*walker) (),
				   void *context);
```

```expression_tree_walker``` allows functions with arbitrary ```coercible``` type but ```check_functions_in_node``` does not due to difference in functions' declaration.

**Comment by mtuncer** [comment]
@anarazel ping

File: src/backend/distributed/planner/multi_router_planner.c [hunk 4]
```diff
@@ -883,123 +884,28 @@ MasterIrreducibleExpressionWalker(Node *expression, WalkerState *state)
...
-	{
-		/*
-		 * to exercise this you need to use a type with a STABLE/VOLATILE intype or
-		 * outtype.
-		 */
-		CoerceViaIO *expr = (CoerceViaIO *) expression;
-		Oid iofunc;
-		Oid typioparam;
-		bool typisvarlena;
+	StaticAssertStmt(PG_VERSION_NUM < 100100, "When porting to a newer PG this section "
```
anarazel: Think we can remove this now.

mtuncer: done

PR #1481, opened 2017-07-04
Wait for commit/abort/prepare results asynchronously

marcocitus:
This PR introduces an API for waiting for results on multiple connections asynchronously and uses it for `COMMIT`, `ROLLBACK` and `PREPARE`. This prevents lock releases on one node from having to wait for lock releases on another node, which could form a distributed deadlock without an explicit edge in the lock graph.

A caller first creates a `MultiConnectionWaiter` for a set of connections and then repeatedly calls `AwaitResultsOnConnections`, which waits for results using a `WaitEventSet`, and returns all connections whose results are ready and connections that failed in a list. Returning a list of connections rather than a single connection (which would be simpler) minimises the number of calls to `CreateWaitEventSet` under load. Connections are returned in a list exactly once. When all connections have been returned, an empty list is returned.

To keep track of which connections were already returned, slots belonging to connections that are returned are overwritten with a connection from `currentOffset` and `currentOffset` is increased, such that the pending connections are always at the end of the connections array in `MultiConnectionWaiter` and the `currentOffset` points to the first connection that is still pending.

### DIFF
--- a/src/backend/distributed/connection/remote_commands.c
+++ b/src/backend/distributed/connection/remote_commands.c
@@ -23,7 +23,12 @@
 /* GUC, determining whether statements sent to remote nodes are logged */
 bool LogRemoteCommands = false;
 
+
 static bool FinishConnectionIO(MultiConnection *connection, bool raiseInterrupts);
+static WaitEventSet * BuildWaitEventSet(MultiConnection **allConnections,
+										int totalConnectionCount,
+										int pendingConnectionsStartIndex);
+
 
 /* simple helpers */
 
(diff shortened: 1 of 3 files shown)


src/backend/distributed/connection/remote_commands.c
```diff
@@ -531,3 +555,238 @@ GetRemoteCommandResult(MultiConnection *connection, bool raiseInterrupts)
...
+	 * event masks. This should be revisited in future postgres versions.
+	 */
+	waitEventSet = CreateWaitEventSet(CurrentMemoryContext, pendingConnectionCount + 2);
+
+	for (connectionIndex = pendingConnectionsStartIndex;
+		 connectionIndex < connectionCount; connectionIndex++)
+	{
+		MultiConnection *connection = allConnections[connectionIndex];
+		int socket = PQsocket(connection->pgConn);
+		int eventMask = WL_SOCKET_READABLE;
```
anarazel: Why are we unconditionally waiting for reads?

src/backend/distributed/connection/remote_commands.c
```diff
@@ -531,3 +555,238 @@ GetRemoteCommandResult(MultiConnection *connection, bool raiseInterrupts)
...
+
+		int sendStatus = PQflush(connection->pgConn);
```
anarazel: Shouldn't things like this handle errors? It's quite possible to detect connection closure here - and a closed connection will always cause the wait-set not to wait at all...

src/backend/distributed/connection/remote_commands.c
marcocitus: You should always check for reads, otherwise the server might get blocked on sending notices, warnings, etc.

src/backend/distributed/connection/remote_commands.c
marcocitus: Well, that was what I was thinking: The code got simpler by letting errors fall through to `WaitEventSetWait`.

We can't easily skip over connections here and then still go into `WaitEventSetWait` since in that case the event position is no longer aligned with the connection index and we'd need some other way of getting the index out of the event. However, we could skip waiting altogether in case of an error and that would also allow checking for `PGINVALID_SOCKET`. Will change it.

src/backend/distributed/connection/remote_commands.c
```diff
@@ -531,3 +555,257 @@ GetRemoteCommandResult(MultiConnection *connection, bool raiseInterrupts)
...
+#if (PG_VERSION_NUM >= 100000)
+		eventCount = WaitEventSetWait(waitEventSet, timeout, waiter->events,
+									  pendingConnectionCount, WAIT_EVENT_CLIENT_READ);
+#else
+		eventCount = WaitEventSetWait(waitEventSet, timeout, waiter->events,
+									  pendingConnectionCount);
+#endif
+		if (eventCount == 0)
+		{
+			/* timeout, try again */
```
anarazel: In the current coding it's not clear to me why we've a timeout?

src/backend/distributed/connection/remote_commands.c
```diff
@@ -531,3 +555,257 @@ GetRemoteCommandResult(MultiConnection *connection, bool raiseInterrupts)
...
+					/* send failed, done with this connection */
+					connectionIsReady = true;
+				}
+				else if (sendStatus == 0)
+				{
+					/* done writing, only wait for read events */
+					ModifyWaitEvent(waitEventSet, connectionIndex, WL_SOCKET_READABLE,
+									NULL);
+				}
+			}
```
marcocitus: `PQputCopyData` is way beyond scope until the `WaitEventSet` issue is resolved.

src/backend/distributed/connection/remote_commands.c
```diff
@@ -531,3 +555,257 @@ GetRemoteCommandResult(MultiConnection *connection, bool raiseInterrupts)
...
+			readyConnectionList = lappend(readyConnectionList, connection);
+
+			/* reuse the slot; move a non-ready connection from the front */
+			allConnections[connectionIndex] = allConnections[waiter->currentOffset];
+
+			waiter->currentOffset++;
+		}
+	}
+
+	FreeWaitEventSet(waitEventSet);
```
marcocitus: Hmz, good point.

src/backend/distributed/connection/remote_commands.c
anarazel: I don't see why that is? It seems perfectly reasonable to want to wait till a command is sent, but hasn't returned yet.  I think this basically is a sign that the abstraction what to wait for ain't quite right.

marcocitus: > I don't see why that is?

Creating a new wait event set for every row in `COPY` would be a bad idea.

> It seems perfectly reasonable to want to wait till a command is sent, but hasn't returned yet. 

What problem are you trying to solve?

src/backend/distributed/connection/remote_commands.c
```diff
@@ -531,3 +555,257 @@ GetRemoteCommandResult(MultiConnection *connection, bool raiseInterrupts)
...
+		{
+			connectionReady[connectionIndex] = false;
+			AddWaitEventToSet(waitEventSet, eventMask, socket, NULL, (void *) connection);
+		}
+		else
+		{
+			connectionReady[connectionIndex] = true;
+			foundReadyConnections = true;
+		}
+	}
```
marcocitus: Seems unnecessary given that `WaitEventSetWait` will return immediately if there is data to read. I'd prefer to keep the code simpler.

src/backend/distributed/connection/remote_commands.c
anarazel: > Creating a new wait event set for every row in COPY would be a bad idea.

You really don't need to? Given that we're, for now at least, ok with "unnecessarily" waiting for reads even if it has some dangers, you can just switch the sockets to wait for reads only in the copy case?

marcocitus:
d'oh, regression test warning order is no longer deterministic :)

anarazel:
> d'oh, regression test warning order is no longer deterministic :)

Hah, that's kinda annoying. Nothing really forces us to output them in the same order we get from the wire however.  Change things so we wait for all of them, and only then output?

marcocitus:
Added a wrapper that returns when all connections are done.

metdos:
Hey @marcocitus, @anarazel, what are our plans for this PR?

src/backend/distributed/connection/remote_commands.c
```diff
@@ -652,3 +656,224 @@ FinishConnectionIO(MultiConnection *connection, bool raiseInterrupts)
...
+					allConnections[connectionIndex] = allConnections[pendingConnectionsStartIndex];
+					pendingConnectionsStartIndex++;
+				}
+			}
+
+		}
+
+		if (waitEventSet != NULL)
+		{
+			FreeWaitEventSet(waitEventSet);
```
anarazel: probably good policy to reset waitEventSet to NULL afterwards, both here and below. Quite possible that somebody adds code that can error out afterwards...

src/backend/distributed/connection/remote_commands.c
```diff
@@ -652,3 +656,224 @@ FinishConnectionIO(MultiConnection *connection, bool raiseInterrupts)
...
+ */
+static WaitEventSet *
+BuildWaitEventSet(MultiConnection **allConnections, int totalConnectionCount,
+				  int pendingConnectionsStartIndex)
+{
+	int pendingConnectionCount = totalConnectionCount - pendingConnectionsStartIndex;
+	WaitEventSet *waitEventSet = NULL;
+	int connectionIndex = 0;
+
+	waitEventSet = CreateWaitEventSet(CurrentMemoryContext, pendingConnectionCount + 2);
```
anarazel: Comment on the +2 probably would be good.

anarazel: Alternatively just move the WL_POSTMASTER_DEATH / WL_LATCH_SET cases up here, documenting it via that?

marcocitus: the event indexes align a bit more nicely with the connection array if WL_POSTMASTER_DEATH and WL_LATCH_SET are at the ended, added a comment.

PR #1483, opened 2017-07-05
Add support for router UPDATEs and DELETEs with subqueries and joins

### LINKED ISSUE #1127: UPDATE ... FROM support for queries that touch a single machine

metdos:
Add support for router UPDATEs and DELETEs with subqueries and joins

Fixes #1127

TODO List:

- [x] Changing error messages: Need to modify RouterSelectJob() to return also deferred messages.
- [x] Support for zero shard cases: Need to refactor RouterSelectJob()
- [x] Make function evaluation more robust: https://github.com/citusdata/citus/pull/1483/commits/a396c0127ad3301633b0ce1546f9a0174c7bbcbd
- [x] Make setting replication model more robust: https://github.com/citusdata/citus/pull/1483/commits/92e2743939f341d17e9d58aded0cfba16810d21a
- [x] PostgreSQL 10 support: changing ruleutils should be fine
- [x] Add regression tests without SELECT but with JOINS (?)
- [x] Fix modify_multiple_shards() problem: Regressions are a bit changed.
- [x] Replication model checks (reference + distributed) + regression tests
- [x] Rename *Select functions and refactor code
- [x] Also support DELETE with the same code path
- [x] Performance tests for UPDATE and DELETE 
- [x] Minor refactoring: style, naming, writing subfunctions

### DIFF
--- a/src/backend/distributed/master/master_modify_multiple_shards.c
+++ b/src/backend/distributed/master/master_modify_multiple_shards.c
@@ -150,7 +150,9 @@ master_modify_multiple_shards(PG_FUNCTION_ARGS)
 
 	if (modifyQuery->commandType != CMD_UTILITY)
 	{
-		DeferredErrorMessage *error = ModifyQuerySupported(modifyQuery);
+		bool multiShardQuery = true;
+		DeferredErrorMessage *error = ModifyQuerySupported(modifyQuery, multiShardQuery);
+
 		if (error)
 		{
 			RaiseDeferredError(error, ERROR);
--- a/src/backend/distributed/planner/deparse_shard_query.c
+++ b/src/backend/distributed/planner/deparse_shard_query.c
@@ -90,11 +90,25 @@ RebuildQueryStrings(Query *originalQuery, List *taskList)
 			}
 		}
 
-		deparse_shard_query(query, relationId, task->anchorShardId,
-							newQueryString);
+		/*
+		 * For INSERT queries, we only have one relation to update, so we can
+		 * use deparse_shard_query(). For UPDATE and DELETE queries, we may have
+		 * subqueries and joins, so we use relation shard list to update shard
+		 * names and call pg_get_query_def() directly.
+		 */
+		if (query->commandType == CMD_INSERT)
+		{
+			deparse_shard_query(query, relationId, task->anchorShardId, newQueryString);
+		}
+		else
+		{
+			List *relationShardList = task->relationShardList;
+			UpdateRelationToShardNames((Node *) query, relationShardList);
 
-		ereport(DEBUG4, (errmsg("distributed statement: %s",
-								newQueryString->data)));
+			pg_get_query_def(query, newQueryString);
+		}
+
+		ereport(DEBUG4, (errmsg("distributed statement: %s", newQueryString->data)));
 
 		task->queryString = newQueryString->data;
 	}
--- a/src/backend/distributed/planner/insert_select_planner.c
+++ b/src/backend/distributed/planner/insert_select_planner.c
@@ -437,12 +437,12 @@ RouterModifyTaskForShardInterval(Query *originalQuery, ShardInterval *shardInter
 	uint64 jobId = INVALID_JOB_ID;
 	List *insertShardPlacementList = NULL;
 	List *intersectedPlacementList = NULL;
-	bool routerPlannable = false;
 	bool upsertQuery = false;
 	bool replacePrunedQueryWithDummy = false;
 	bool allReferenceTables = restrictionContext->allReferenceTables;
 	List *shardOpExpressions = NIL;
 	RestrictInfo *shardRestrictionList = NULL;
+	DeferredErrorMessage *planningError = NULL;
 
 	/* grab shared metadata lock to stop concurrent placement additions */
 	LockShardDistributionMetadata(shardId, ShareLock);
@@ -489,15 +489,15 @@ RouterModifyTaskForShardInterval(Query *originalQuery, ShardInterval *shardInter
 	replacePrunedQueryWithDummy = false;
 
 	/*
-	 * Use router select planner to decide on whether we can push down the query
-	 * or not. If we can, we also rely on the side-effects that all RTEs have been
-	 * updated to point to the relevant nodes and selectPlacementList is determined.
+	 * Use router planner to decide on whether we can push down the query or not.
+	 * If we can, we also rely on the side-effects that all RTEs have been updated
+	 * to point to the relevant nodes and selectPlacementList is determined.
 	 */
-	routerPlannable = RouterSelectQuery(copiedSubquery, copiedRestrictionContext,
-										&selectPlacementList, &selectAnchorShardId,
-										&relationShardList, replacePrunedQueryWithDummy);
+	planningError = PlanRouterQuery(copiedSubquery, copiedRestrictionContext,
+									&selectPlacementList, &selectAnchorShardId,
+									&relationShardList, replacePrunedQueryWithDummy);
 
-	if (!routerPlannable)
+	if (planningError)
 	{
 		ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
 						errmsg("cannot perform distributed planning for the given "
--- a/src/backend/distributed/planner/multi_router_planner.c
+++ b/src/backend/distributed/planner/multi_router_planner.c
@@ -93,33 +93,30 @@ static bool MasterIrreducibleExpressionWalker(Node *expression, WalkerState *sta
 static bool MasterIrreducibleExpressionFunctionChecker(Oid func_id, void *context);
 static bool TargetEntryChangesValue(TargetEntry *targetEntry, Var *column,
 									FromExpr *joinTree);
-static Job * RouterModifyJob(Query *originalQuery, Query *query,
+static Job * RouterInsertJob(Query *originalQuery, Query *query,
 							 DeferredErrorMessage **planningError);
 static void ErrorIfNoShardsExist(DistTableCacheEntry *cacheEntry);
 static bool CanShardPrune(Oid distributedTableId, Query *query);
-static List * RouterInsertTaskList(Query *query, DistTableCacheEntry *cacheEntry,
-								   DeferredErrorMessage **planningError);
-static List * RouterUpdateOrDeleteTaskList(Query *query, DistTableCacheEntry *cacheEntry,
-										   DeferredErrorMessage **planningError);
 static Job * CreateJob(Query *query);
 static Task * CreateTask(TaskType taskType);
 static ShardInterval * FindShardForInsert(Query *query, DistTableCacheEntry *cacheEntry,
 										  DeferredErrorMessage **planningError);
-static ShardInterval * FindShardForUpdateOrDelete(Query *query,
-												  DistTableCacheEntry *cacheEntry,
-												  DeferredErrorMessage **planningError);
-static List * QueryRestrictList(Query *query, char partitionMethod);
 static Expr * ExtractInsertPartitionValue(Query *query, Var *partitionColumn);
-static Job * RouterSelectJob(Query *originalQuery,
-							 RelationRestrictionContext *restrictionContext,
-							 bool *queryRoutable);
+static Job * RouterJob(Query *originalQuery,
+					   RelationRestrictionContext *restrictionContext,
+					   DeferredErrorMessage **planningError);
 static bool RelationPrunesToMultipleShards(List *relationShardList);
-static List * TargetShardIntervalsForSelect(Query *query,
-											RelationRestrictionContext *restrictionContext);
+static List * TargetShardIntervalsForRouter(Query *query,
+											RelationRestrictionContext *restrictionContext,
+											bool *multiShardQuery);
 static List * WorkersContainingAllShards(List *prunedShardIntervalsList);
 static bool MultiRouterPlannableQuery(Query *query,
 									  RelationRestrictionContext *restrictionContext);
 static DeferredErrorMessage * ErrorIfQueryHasModifyingCTE(Query *queryTree);
+static bool UpdateOrDeleteQuery(Query *query);
+static RangeTblEntry * GetUpdateOrDeleteRTE(List *rangeTableList);
+static bool UpdateOrDeleteRTE(RangeTblEntry *rangeTableEntry);
+static bool SelectsFromDistributedTable(List *rangeTableList);
 #if (PG_VERSION_NUM >= 100000)
 static List * get_all_actual_clauses(List *restrictinfo_list);
 #endif
@@ -161,16 +158,28 @@ CreateModifyPlan(Query *originalQuery, Query *query,
 {
 	Job *job = NULL;
 	MultiPlan *multiPlan = CitusMakeNode(MultiPlan);
+	bool multiShardQuery = false;
 
 	multiPlan->operation = query->commandType;
 
-	multiPlan->planningError = ModifyQuerySupported(query);
+	multiPlan->planningError = ModifyQuerySupported(query, multiShardQuery);
 	if (multiPlan->planningError != NULL)
 	{
 		return multiPlan;
 	}
 
-	job = RouterModifyJob(originalQuery, query, &multiPlan->planningError);
+	if (UpdateOrDeleteQuery(query))
+	{
+		RelationRestrictionContext *restrictionContext =
+			plannerRestrictionContext->relationRestrictionContext;
+
+		job = RouterJob(originalQuery, restrictionContext, &multiPlan->planningError);
+	}
+	else
+	{
+		job = RouterInsertJob(originalQuery, query, &multiPlan->planningError);
+	}
+
 	if (multiPlan->planningError != NULL)
 	{
 		return multiPlan;
@@ -204,7 +213,6 @@ CreateSingleTaskRouterPlan(Query *originalQuery, Query *query,
 						   RelationRestrictionContext *restrictionContext)
 {
 	Job *job = NULL;
-	bool queryRoutable = false;
 	MultiPlan *multiPlan = CitusMakeNode(MultiPlan);
 
 	multiPlan->operation = query->commandType;
@@ -216,8 +224,8 @@ CreateSingleTaskRouterPlan(Query *originalQuery, Query *query,
 		return multiPlan;
 	}
 
-	job = RouterSelectJob(originalQuery, restrictionContext, &queryRoutable);
-	if (!queryRoutable)
+	job = RouterJob(originalQuery, restrictionContext, &multiPlan->planningError);
+	if (multiPlan->planningError)
 	{
 		/* query cannot be handled by this planner */
 		return NULL;
@@ -455,7 +463,7 @@ ExtractInsertRangeTableEntry(Query *query)
  * features, otherwise it returns an error description.
  */
 DeferredErrorMessage *
-ModifyQuerySupported(Query *queryTree)
+ModifyQuerySupported(Query *queryTree, bool multiShardQuery)
 {
 	Oid distributedTableId = ExtractFirstDistributedTableId(queryTree);
 	uint32 rangeTableId = 1;
@@ -479,9 +487,18 @@ ModifyQuerySupported(Query *queryTree)
 	 */
 	if (queryTree->hasSubLinks == true)
 	{
-		return DeferredError(ERRCODE_FEATURE_NOT_SUPPORTED,
-							 "subqueries are not supported in distributed modifications",
-							 NULL, NULL);
+		/*
+		 * We support UPDATE and DELETE with subqueries unless they are multi
+		 * shard queries.
+		 */
+		if (!UpdateOrDeleteQuery(queryTree) || multiShardQuery)
+		{
+			return DeferredError(ERRCODE_FEATURE_NOT_SUPPORTED,
+								 "cannot perform distributed planning for the given "
+								 "modifications",
+								 "Subqueries are not supported in distributed "
+								 "modifications.", NULL);
+		}
 	}
 
 	/* reject queries which include CommonTableExpr */
@@ -542,6 +559,17 @@ ModifyQuerySupported(Query *queryTree)
 		}
 		else
 		{
+			char *rangeTableEntryErrorDetail = NULL;
+
+			/*
+			 * We support UPDATE and DELETE with subqueries and joins unless
+			 * they are multi shard queries.
+			 */
+			if (UpdateOrDeleteQuery(queryTree) && !multiShardQuery)
+			{
+				continue;
+			}
+
 			/*
 			 * Error out for rangeTableEntries that we do not support.
 			 * We do not explicitly specify "in FROM clause" in the error detail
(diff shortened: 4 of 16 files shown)


metdos:
@marcocitus, I figured out why `pg_get_query_def()` fails during `RebuildQueryStrings()` in executor part. We update rte type to RTE_FUNCTION from RTE_RELATION. During serialization and deserialization, this causes relation id field to disappear.  `pg_get_query_def()` needs to find relation from relid and find attribute names, so we see errors like `ERROR:  cache lookup failed for attribute 4 of relation 0`. Adding relid field to _outRangeTblEntry and _readRangeTblEntry for RTE_FUNCTION solved the problem. Obviously, we can not change PostgreSQL code. 

As a quick check I retrieved relation id from the relation name here:https://github.com/citusdata/citus/pull/1483/commits/a396c0127ad3301633b0ce1546f9a0174c7bbcbd and it worked.

I suppose another option would be to pass relation id as the fifth argument through `SetRangeTblExtraData`.

src/backend/distributed/planner/deparse_shard_query.c
```diff
@@ -90,8 +90,15 @@ RebuildQueryStrings(Query *originalQuery, List *taskList)
...
-		deparse_shard_query(query, relationId, task->anchorShardId,
-							newQueryString);
+		if (query->commandType == CMD_UPDATE)
+		{
+			pg_get_query_def(query, newQueryString);
+		}
+		else
+		{
+			deparse_shard_query(query, relationId, task->anchorShardId,
+								newQueryString);
```
marcocitus: What if you do:

````
        else if (query->commandType == CMD_UPDATE)
        {   
            List *relationShardList = task->relationShardList;
            
            UpdateRelationToShardNames((Node *) query, relationShardList);
        }

        deparse_shard_query(query, relationId, task->anchorShardId,
```

metdos: Are you proposing this to relation id problem I described above? If yes, this does not solve the problem by itself because deparse_shard_query() uses the same path. We called UpdateRelationToShardNames() through RouterSelectJob() path which already updated RTE types. This was not a problem for RouterModifyJob() which does not call UpdateRelationToShardNames() if it requires reevaluation in the executor. I did a similar hack with this commit: https://github.com/citusdata/citus/pull/1483/commits/bc04774d6e36c83e9cbd8a86176c1ea2515d850c

I will look to unify and polish this path.

metdos:
Another problem is in `if (deferredPruning)`  block of `CitusModifyBeginScan`. Here we call `RouterModifyTaskList` and we can not call `RouterSelectQuery` because we don't have access to restriction context here (executor). One alternative solution is to pass restriction context to executor via MultiPlan when `deferredPruning` is set.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -161,16 +161,28 @@ CreateModifyPlan(Query *originalQuery, Query *query,
...
+	if (multiPlan->operation == CMD_UPDATE || multiPlan->operation == CMD_DELETE)
+	{
+		RelationRestrictionContext *restrictionContext =
+			plannerRestrictionContext->relationRestrictionContext;
+		job = RouterSelectJob(originalQuery, restrictionContext,
+							  &multiPlan->planningError);
+	}
+	else
+	{
+		job = RouterModifyJob(originalQuery, query, &multiPlan->planningError);
```
marcocitus: Should maybe be `RouterInsertJob` now.

src/backend/distributed/master/master_modify_multiple_shards.c
```diff
@@ -150,7 +150,9 @@ master_modify_multiple_shards(PG_FUNCTION_ARGS)
 
 	if (modifyQuery->commandType != CMD_UTILITY)
 	{
-		DeferredErrorMessage *error = ModifyQuerySupported(modifyQuery);
+		bool onlySimpleUpdates = true;
```
marcocitus: I'd maybe just call this `multiShardUpdate` or `multiShardQuery` to make its meaning clear.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1640,52 +1669,108 @@ ExtractInsertPartitionValue(Query *query, Var *partitionColumn)
...
+				DistributedTableCacheEntry(relationShard->relationId);
+
+			if (replicationModel == REPLICATION_MODEL_INVALID)
+			{
+				replicationModel = cacheEntry->replicationModel;
+			}
+			else if (replicationModel != cacheEntry->replicationModel)
+			{
+				ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+								errmsg("cannot perform an UPDATE on tables with "
```
marcocitus: Do we care that much? I think we only care about the replication model of the placement that is being modified.

src/backend/distributed/planner/multi_router_planner.c
metdos: I was thinking the same :)

src/backend/distributed/master/master_modify_multiple_shards.c
metdos: multiShardQuery sounds good

src/backend/distributed/planner/multi_router_planner.c
metdos: Hmm, right. Then, shall we just error on select distributed and update reference?

marcocitus: Yup, that makes sense.

metdos:
@marcocitus, I'm done with high-level changes. I will pause refactoring(style, naming, writing subfunctions) mostly because of dependency on #1474 and conflicts with #1517.

I'm planning to continue in this order
1. Run a quick benchmark to see if performance of UPDATE/DELETE is effected
2. Start reviewing #1474 
3. Then return back to this PR

Does this make sense? In the meantime, could you make a high-level code review?

src/backend/distributed/planner/deparse_shard_query.c
```diff
@@ -90,11 +90,23 @@ RebuildQueryStrings(Query *originalQuery, List *taskList)
 			}
 		}
 
-		deparse_shard_query(query, relationId, task->anchorShardId,
-							newQueryString);
+		/*
+		 * UPDATE and DELETE queries can include subqueries, so treat them
+		 * differently.
```
marcocitus: Maybe clarify what's different between `deparse_shard_query` and `pg_get_query_def`. Might also be something we should address.

src/backend/distributed/planner/insert_select_planner.c
```diff
@@ -493,11 +493,12 @@ RouterModifyTaskForShardInterval(Query *originalQuery, ShardInterval *shardInter
 	 * or not. If we can, we also rely on the side-effects that all RTEs have been
 	 * updated to point to the relevant nodes and selectPlacementList is determined.
 	 */
-	routerPlannable = RouterSelectQuery(copiedSubquery, copiedRestrictionContext,
-										&selectPlacementList, &selectAnchorShardId,
-										&relationShardList, replacePrunedQueryWithDummy);
+	RouterQuery(copiedSubquery, copiedRestrictionContext,
```
marcocitus: Maybe `PlanRouterQuery` is a more appropriate name.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1640,66 +1494,150 @@ ExtractInsertPartitionValue(Query *query, Var *partitionColumn)
 /* RouterSelectJob builds a Job to represent a single shard select query */
 static Job *
 RouterSelectJob(Query *originalQuery, RelationRestrictionContext *restrictionContext,
```
marcocitus: Maybe `PlanRouterJob`, also update the comment to say that it handles UPDATE and DELETE as well.

src/test/regress/expected/multi_mx_modifications.out
```diff
@@ -174,7 +174,7 @@ ERROR:  common table expressions are not supported in distributed modifications
 -- cursors are not supported
 DELETE FROM limit_orders_mx WHERE CURRENT OF cursor_name;
 ERROR:  cannot run DELETE command which targets multiple shards
-HINT:  Consider using an equality filter on partition column "id" to target a single shard. If you'd like to run a multi-shard operation, use master_modify_multiple_shards().
+HINT:  Consider using an equality filter on partition column to target a single shard. If you'd like to run a multi-shard operation, use master_modify_multiple_shards().
```
marcocitus: is there a way to keep the partition column name here?

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1640,66 +1494,150 @@ ExtractInsertPartitionValue(Query *query, Var *partitionColumn)
...
+	{
+		task = CreateTask(ROUTER_TASK);
+	}
+	else
+	{
+		char replicationModel = REPLICATION_MODEL_INVALID;
+		bool distributedSelect = false;
+		char modificationPartitionMethod = 0;
+
+		/* XXX: todo refactor into a function */
```
marcocitus: agreed :)

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1640,66 +1494,150 @@ ExtractInsertPartitionValue(Query *query, Var *partitionColumn)
...
+
+			if (relationId == InvalidOid)
+			{
+				continue;
+			}
+
+			if ((ACL_UPDATE & rangeTableEntry->requiredPerms) ||
+				(ACL_DELETE & rangeTableEntry->requiredPerms))
+			{
+				requiresUpdateOrDeletePermissions = true;
```
marcocitus: this loop would be a lot clearer if it was broken down in a few steps / separate loops,

e.g. thinking out lout:
```
modificationRTE = GetModificationRTE(rangeTableList);
modificationTableCacheEntry = DistributedTableCacheEntry(rangeTableEntry->relid);

modificationPartitionMethod = cacheEntry->partitionMethod;
if (modificationPartitionMethod == DISTRIBUTE_BY_NONE && SelectsFromDistributedTable(rangeTableList))
{
    ereport(ERROR, ... );
}

task = CreateTask(MODIFY_TASK);
task->replicationModel = cacheEntry->replicationModel;
```

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -479,9 +482,15 @@ ModifyQuerySupported(Query *queryTree)
 	 */
 	if (queryTree->hasSubLinks == true)
 	{
-		return DeferredError(ERRCODE_FEATURE_NOT_SUPPORTED,
-							 "subqueries are not supported in distributed modifications",
-							 NULL, NULL);
+		/* we have an exception for UPDATE queries if multiShardQuery is set false */
+		if ((commandType != CMD_UPDATE && commandType != CMD_DELETE) || multiShardQuery)
```
marcocitus: Could maybe introduce an `IsUpdateOrDelete(queryTree)` function for cleanliness.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1640,66 +1494,150 @@ ExtractInsertPartitionValue(Query *query, Var *partitionColumn)
...
+	foreach(rangeTableCell, rangeTableList)
+	{
+		RangeTblEntry *rangeTableEntry = (RangeTblEntry *) lfirst(rangeTableCell);
+		bool requiresUpdateOrDeletePermissions = false;
+
+		/* XXX: todo refactor into a function */
+		if ((ACL_UPDATE & rangeTableEntry->requiredPerms) ||
+			(ACL_DELETE & rangeTableEntry->requiredPerms))
+		{
+			requiresUpdateOrDeletePermissions = true;
```
marcocitus: Could also use `GetModificationRTE(rangeTableList)` here.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1640,66 +1494,150 @@ ExtractInsertPartitionValue(Query *query, Var *partitionColumn)
...
+		}
+
+		if (rangeTableEntry->rtekind == RTE_SUBQUERY && requiresUpdateOrDeletePermissions)
```
marcocitus: This requires more elaboration in a comment. 

neat solution though :)

metdos:
I have seen some performance loss (around %8-9) with this branch on UPDATE/DELETE on a quick test with pgbench (examples are in the release test document 32 shards/replication count is 1). The reason, most likely, is that now we use the same path with router select. One option would be to use older code path for simple UPDATE/DELETE commands but this would increase the cost of code maintenance.

This branch:
```
UPDATE
tps = 4903.296459 (including connections establishing)
tps = 4904.893971 (excluding connections establishing)

DELETE
tps = 4919.072987 (including connections establishing)
tps = 4920.434646 (excluding connections establishing)
```
Master branch at the time being:
```
UPDATE
tps = 5348.763625 (including connections establishing)
tps = 5350.687116 (excluding connections establishing)

DELETE
tps = 5410.122127 (including connections establishing)
tps = 5412.155743 (excluding connections establishing)
```

marcocitus:
Pity, though I'd rather take the code simplification and see where we can optimise. Could you run perf to see what the main bottlenecks are?

metdos:
> Pity, though I'd rather take the code simplification and see where we can optimise. Could you run perf to see what the main bottlenecks are?

@marcocitus, it seems that the results I get above were intermittent. I updated the comment https://github.com/citusdata/citus/pull/1483#issuecomment-318045343 and it seems there is no regression.

Though I run perf and it seems most of the time spend on PostgreSQL functions. I couldn't pinpoint a place to improve on a quick look. Attaching for reference:

This is the command I run;
```bash
sudo perf record -u ec2-user -g -s -- /home/ec2-user/citus-installation/bin/pgbench -U ec2-user -c 256 -j 32 -f update.sql -n -t 500 -P 5
```
This is the report;
```
sudo perf report --sort comm,dso
```

Samples: 115K of event 'cpu-clock', Event count (approx.): 28954500000
  Children      Self  Command   Shared Object                                                                                                                                                                                                                                  ◆
+   64.07%     0.00%  postgres  [unknown]                                                                                                                                                                                                                                      ▒
-   45.64%    19.99%  postgres  libc-2.17.so                                                                                                                                                                                                                                   ▒
   + 15.01% __libc_send                                                                                                                                                                                                                                                        ▒
   + 3.70% epoll_ctl                                                                                                                                                                                                                                                           ▒
   + 3.48% epoll_create1                                                                                                                                                                                                                                                       ▒
     1.40% 0                                                                                                                                                                                                                                                                   ▒
     1.39% __GI_____strtoull_l_internal                                                                                                                                                                                                                                        ▒
   + 1.25% __libc_recv                                                                                                                                                                                                                                                         ▒
     1.14% __memcpy_ssse3_back                                                                                                                                                                                                                                                 ▒
     1.02% __GI_____strtoll_l_internal                                                                                                                                                                                                                                         ▒
     0.86% __memcmp_sse4_1                                                                                                                                                                                                                                                     ▒
   + 0.70% __epoll_wait_nocancel                                                                                                                                                                                                                                               ▒
-   44.60%    42.80%  postgres  postgres                                                                                                                                                                                                                                       ▒
     7.62% pg_strtok                                                                                                                                                                                                                                                           ▒
     4.31% hash_search_with_hash_value                                                                                                                                                                                                                                         ▒
     3.03% 0                                                                                                                                                                                                                                                                   ▒
     1.39% AllocSetAlloc                                                                                                                                                                                                                                                       ▒
   + 1.20% 0x2002d50                                                                                                                                                                                                                                                           ▒
     1.10% nodeRead                                                                                                                                                                                                                                                            ▒
     0.69% parseNodeString                                                                                                                                                                                                                                                     ▒
   + 0.63% 0x2002f08                                                                                                                                                                                                                                                           ▒
     0.61% SearchCatCache                                                                                                                                                                                                                                                      ▒
-   28.00%    28.00%  postgres  [kernel.kallsyms]                                                                                                                                                                                                                              ▒
   - 5.87% 0x2003c68                                                                                                                                                                                                                                                           ▒
      + 3.49% __libc_send                                                                                                                                                                                                                                                      ▒
      + 1.04% epoll_ctl                                                                                                                                                                                                                                                        ▒
      + 0.87% epoll_create1                                                                                                                                                                                                                                                    ▒
   + 1.88% __libc_send                                                                                                                                                                                                                                                         ▒
     1.62% 0                                                                                                                                                                                                                                                                   ▒
     0.55% 0x20f1978                                                                                                                                                                                                                                                           ▒
-    6.66%     5.73%  postgres  citus.so                                                                                                                                                                                                                                       ▒
     0.85% 0                                                                                                                                                                                                                                                                   ▒
     0.51% 0x74fcb                                                                                                                                                                                                                                                             ▒
-    3.46%     3.39%  postgres  libpq.so.5.9                                                                                                                                                                                                                                   ▒
   + 0.56% 0x7f003130322e                                                                                                                                                                                                                                                      ▒
   + 0.55% 0x7f003933312e                                                                                                                                                                                                                                                      ▒
     0.13%     0.05%  postgres  ld-2.17.so                                                                                                                                                                                                                                     ▒
     0.03%     0.01%  postgres  libcrypto.so.1.0.1k                                                                                                                                                                                                                            ▒
     0.01%     0.00%  postgres  [vdso]                                                                                                                                                                                                                                         ▒
     0.01%     0.00%  postgres  libssl.so.1.0.1k                                                                                                                                                                                                                               ▒
     0.01%     0.01%  postgres  libm-2.17.so                                                                                                                                                                                                                                   ▒
     0.01%     0.00%  postgres  libresolv-2.17.so                                                                                                                                                                                                                              ▒
     0.01%     0.00%  postgres  libselinux.so.1                                                                                                                                                                                                                                ▒
     0.00%     0.00%  postgres  libkeyutils.so.1.5                                                                                                                                                                                                                             ▒
     0.00%     0.00%  postgres  libpthread-2.17.so                                                                                                                                                                                                                             ▒
     0.00%     0.00%  postgres  libk5crypto.so.3.1                                                                                                                                                                                                                             ▒
     0.00%     0.00%  postgres  libcom_err.so.2.1                                                                                                                                                                                                                              ▒
     0.00%     0.00%  postgres  libkrb5.so.3.3                                                                                                                                                                                                                                 ▒
     0.00%     0.00%  postgres  libgssapi_krb5.so.2.2                                                                                                                                                                                                                          ▒
     0.00%     0.00%  postgres  libdl-2.17.so                                                                                                                                                                                                                                  ▒
     0.00%     0.00%  postgres  libnss_files-2.17.so                                                                                                                                                                                                                           ▒
     0.00%     0.00%  postgres  libz.so.1.2.8                                                                                                                                                                                                                                  ▒
     0.00%     0.00%  postgres  libkrb5support.so.0.1                                                                                                                                                                                                                          ▒
     0.00%     0.00%  postgres  librt-2.17.so

anarazel:
>This is the report;
>```
>sudo perf report --sort comm,dso
>```
>
>Samples: 115K of event 'cpu-clock', Event count (approx.): 28954500000
>Children      Self  Command   Shared Object                            
>                                                                      ◆
>+   64.07%     0.00%  postgres  [unknown]                              

This suggests that you don't have full debugging symbols installed...

Andres
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

metdos:
> This suggests that you don't have full debugging symbols installed...

@anarazel, good point. I updated the results https://github.com/citusdata/citus/pull/1483#issuecomment-318359160. With debugging symbols, there is a %25 decrease in TPS. Do you think this is expected?

src/test/regress/expected/multi_mx_modifications.out
metdos: Yes, there is one now :)

src/backend/distributed/planner/deparse_shard_query.c
metdos: Here we can call UpdateRelationToShardNames() on UPDATE and DELETEs and then run deparse_shard_query() without calling pg_get_query_def(). However, I think this a bit confusing, because shard id used by deparse_shard_query() becomes a non-effective operation and reading this would be a bit confusing. I'm updating the comment.

metdos:
Hey @marcocitus, this PR is ready for another pass.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1554,84 +1440,254 @@ ExtractInsertPartitionValue(Query *query, Var *partitionColumn)
...
 	bool shardsPresent = false;
+	uint64 shardId = INVALID_SHARD_ID;
 
 	*placementList = NIL;
 
+	/*
+	 * If prunedRelationShardList is NULL then it means a relation has more
+	 * than one shard left after pruning.
+	 *
+	 * TODO: The way TargetShardIntervalsForRouter represents multiple shards
```
marcocitus: Are you planning to resolve this Todo?

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1554,84 +1440,254 @@ ExtractInsertPartitionValue(Query *query, Var *partitionColumn)
...
 	}
 
 	job = CreateJob(originalQuery);
 
+	ExtractRangeTableEntryWalker((Node *) originalQuery, &rangeTableList);
+	updateOrDeleteRTE = GetUpdateOrDeleteRTE(rangeTableList);
+
+	/*
+	 * If all of the shards are pruned, we replace the relation RTE into
+	 * subquery RTE that returns no results. However, this is not usefull
```
marcocitus: Typo: useful

src/backend/distributed/planner/multi_router_planner.c
metdos: This was not part of this PR, but I can do that as a separate commit.

metdos: I added a commit to fix this. I think I can also squash that one given that I also modified other parts of this function.

metdos:
Hey @marcocitus, I addressed your comments. Also added regression tests for DELETE.

metdos:
> 🚢 though maybe ping @jasonmp85 before merging, since there's likely some conflict with #1517.

@jasonmp85, any preference? One thing to note, #1509 also depends on this PR. I don't mind rebasing after you, but timewise it's better to merge this soon.

jasonmp85:
@metdos if you're ready to merge this, go ahead. You're cycling off the workday and I'm in SF right now so I can deal with the rebase problems today, whereas you'd have to wait until tomorrow. Also, I just dealt with a substantial (100+) number of them from #1474, so I can easily re-apply my changes after you merge this.

jasonmp85:
In fact, to unblock myself and avoid doing work I'll have to redo later, I'll probably merge this if you don't get back to me in the next 30m or so (sorry I didn't respond earlier, we had stand up).

metdos:
> @metdos if you're ready to merge this, go ahead. You're cycling off the workday and I'm in SF right now so I can deal with the rebase problems today, whereas you'd have to wait until tomorrow. Also, I just dealt with a substantial (100+) number of them from #1474, so I can easily re-apply my changes after you merge this.

Sounds good!

PR #1488, opened 2017-07-10
Fix conflicting locks in VACUUM and INSERT

byucesoy:
Before this change, we used ShareLock to acquire lock on distributed tables while
running VACUUM. This makes VACUUM and INSERT block each other. With this change we
changed lock mode from ShareLock to ShareUpdateExclusiveLock, which does not conflict
with the locks INSERT acquire.

We also removed LockRelationDistributionMetadata, which is not used frequently and confusing.

### DIFF
--- a/src/test/regress/expected/isolation_insert_vs_vacuum.out
+++ b/src/test/regress/expected/isolation_insert_vs_vacuum.out
@@ -1,0 +1,36 @@
+Parsed test spec with 2 sessions
+
+starting permutation: s1-begin s1-insert s2-vacuum-analyze s1-commit
+create_distributed_table
+
+               
+step s1-begin: 
+    BEGIN;
+
+step s1-insert: 
+    INSERT INTO test_insert_vacuum VALUES(1, 1);
+
+step s2-vacuum-analyze: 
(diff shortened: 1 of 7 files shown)


marcocitus:
Looks good, but could we add an isolation test for `BEGIN; INSERT ...;` vs `VACUUM`?

src/test/regress/expected/isolation_insert_vs_vacuum.out
```diff
@@ -0,0 +1,36 @@
...
+               
+step s1-begin: 
+    BEGIN;
+
+step s1-insert: 
+    INSERT INTO test_insert_vacuum VALUES(1, 1);
+
+step s2-vacuum-full: 
+    VACUUM FULL test_insert_vacuum;
+ <waiting ...>
```
marcocitus: We should probably add a comment to the `LockRelationOid` call that we may take a stronger lock in the `ProcessUtility` in case of vacuum full.

byucesoy:
@sumedhpathak, Yes. I tagged it as backport but did not backpoerted it to 6.2.3 yet. Since we will start packaging of 6.2.3 this week, I am planning to backport it before starting to packaging.

PR #1489, opened 2017-07-10
Introduce distributed transaction ids

onderkalaci:
This commit adds distributed transaction id infrastructure in
    the scope of distributed deadlock detection.
    
    In general, the distributed transaction id consists of a tuple
    in the form of: `(databaseId, initiatorNodeIdentifier, transactionId,
    timestamp)`.
    
    Briefly, we add a shared memory block on each node, which holds some
    information per backend (i.e., an array `BackendData backends[MaxBackends]`).
    Later, on each coordinated transaction, Citus sends
    `SELECT assign_distributed_transaction_id()` right after `BEGIN`.
    For that backend on the worker, the distributed transaction id is set to
    the values assigned via the function call.
    
    The aim of the above is to correlate the transactions on the coordinator
    to the transactions on the worker nodes.

### DIFF
--- a/src/backend/distributed/citus--7.0-3--7.0-4.sql
+++ b/src/backend/distributed/citus--7.0-3--7.0-4.sql
@@ -1,0 +1,25 @@
+/* citus--7.0-3--7.0-4.sql */
+
+SET search_path = 'pg_catalog';
+
+CREATE FUNCTION assign_distributed_transaction_id(initiator_node_identifier int4, transaction_number int8, transaction_stamp timestamptz)
+     RETURNS void
+     LANGUAGE C STRICT
+     AS 'MODULE_PATHNAME',$$assign_distributed_transaction_id$$;
+ COMMENT ON FUNCTION assign_distributed_transaction_id(initiator_node_identifier int4, transaction_number int8, transaction_stamp timestamptz)
+     IS 'Only intended for internal use, users should not call this. The function sets the distributed transaction id';
+
+CREATE OR REPLACE FUNCTION get_current_transaction_id(OUT database_id oid, OUT process_id int, OUT initiator_node_identifier int4, OUT transaction_number int8, OUT transaction_stamp timestamptz)
+     RETURNS RECORD
+     LANGUAGE C STRICT
+     AS 'MODULE_PATHNAME',$$get_current_transaction_id$$;
+ COMMENT ON FUNCTION get_current_transaction_id(OUT database_id oid, OUT process_id int, OUT initiator_node_identifier int4, OUT transaction_number int8, OUT transaction_stamp timestamptz)
+     IS 'returns the current backend data including distributed transaction id';
+
+CREATE OR REPLACE FUNCTION get_all_active_transactions(OUT database_id oid, OUT process_id int, OUT initiator_node_identifier int4, OUT transaction_number int8, OUT transaction_stamp timestamptz)
+	RETURNS SETOF RECORD
+	LANGUAGE C STRICT
+	AS 'MODULE_PATHNAME', $$get_all_active_transactions$$;
+ COMMENT ON FUNCTION get_all_active_transactions(OUT database_id oid, OUT process_id int, OUT initiator_node_identifier int4, OUT transaction_number int8, OUT transaction_stamp timestamptz)
+     IS 'returns distributed transaction ids of active distributed transactions';
+RESET search_path;
--- a/src/backend/distributed/connection/remote_commands.c
+++ b/src/backend/distributed/connection/remote_commands.c
@@ -337,10 +337,11 @@ ExecuteOptionalRemoteCommand(MultiConnection *connection, const char *command,
 
 
 /*
- * SendRemoteCommand is a PQsendQuery wrapper that logs remote commands, and
- * accepts a MultiConnection instead of a plain PGconn.  It makes sure it can
+ * SendRemoteCommandParams is a PQsendQueryParams wrapper that logs remote commands,
+ * and accepts a MultiConnection instead of a plain PGconn. It makes sure it can
  * send commands asynchronously without blocking (at the potential expense of
- * an additional memory allocation).
+ * an additional memory allocation). The command string can only include a single
+ * command since PQsendQueryParams() supports only that.
  */
 int
 SendRemoteCommandParams(MultiConnection *connection, const char *command,
@@ -372,14 +373,33 @@ SendRemoteCommandParams(MultiConnection *connection, const char *command,
 
 /*
  * SendRemoteCommand is a PQsendQuery wrapper that logs remote commands, and
- * accepts a MultiConnection instead of a plain PGconn.  It makes sure it can
+ * accepts a MultiConnection instead of a plain PGconn. It makes sure it can
  * send commands asynchronously without blocking (at the potential expense of
- * an additional memory allocation).
+ * an additional memory allocation). The command string can include multiple
+ * commands since PQsendQuery() supports that.
  */
 int
 SendRemoteCommand(MultiConnection *connection, const char *command)
 {
-	return SendRemoteCommandParams(connection, command, 0, NULL, NULL);
+	PGconn *pgConn = connection->pgConn;
+	int rc = 0;
+
+	LogRemoteCommand(connection, command);
+
+	/*
+	 * Don't try to send command if connection is entirely gone
+	 * (PQisnonblocking() would crash).
+	 */
+	if (!pgConn)
+	{
+		return 0;
+	}
+
+	Assert(PQisnonblocking(pgConn));
+
+	rc = PQsendQuery(pgConn, command);
+
+	return rc;
 }
 
 
--- a/src/backend/distributed/transaction/backend_data.c
+++ b/src/backend/distributed/transaction/backend_data.c
@@ -1,0 +1,489 @@
+/*-------------------------------------------------------------------------
+ *
+ * backend_data.c
+ *
+ *  Infrastructure for managing per backend data that can efficiently
+ *  accessed by all sessions.
+ *
+ * Copyright (c) 2017, Citus Data, Inc.
+ *
+ *-------------------------------------------------------------------------
+ */
+
+#include "postgres.h"
+#include "miscadmin.h"
+
+#include "funcapi.h"
+#include "access/htup_details.h"
+#include "catalog/pg_type.h"
+#include "datatype/timestamp.h"
+#include "distributed/backend_data.h"
+#include "distributed/listutils.h"
+#include "distributed/metadata_cache.h"
+#include "distributed/transaction_identifier.h"
+#include "nodes/execnodes.h"
+#include "storage/ipc.h"
+#include "storage/lwlock.h"
+#include "storage/proc.h"
+#include "storage/spin.h"
+#include "storage/s_lock.h"
+#include "utils/timestamp.h"
+
+
+/*
+ * Each backend's data reside in the shared memory
+ * on the BackendManagementShmemData.
+ */
+typedef struct BackendManagementShmemData
+{
+	int trancheId;
+#if (PG_VERSION_NUM >= 100000)
+	NamedLWLockTranche namedLockTranche;
+#else
+	LWLockTranche lockTranche;
+#endif
+	LWLock lock;
+
+	/*
+	 * We prefer to use an atomic integer over sequences for two
+	 * reasons (i) orders of magnitude performance difference
+	 * (ii) allowing read-only replicas to be able to generate ids
+	 */
+	pg_atomic_uint64 nextTransactionNumber;
+
+	BackendData backends[FLEXIBLE_ARRAY_MEMBER];
+} BackendManagementShmemData;
+
+
+static shmem_startup_hook_type prev_shmem_startup_hook = NULL;
+static BackendManagementShmemData *backendManagementShmemData = NULL;
+static BackendData *MyBackendData = NULL;
+
+
+static void BackendManagementShmemInit(void);
+static size_t BackendManagementShmemSize(void);
+
+
+PG_FUNCTION_INFO_V1(assign_distributed_transaction_id);
+PG_FUNCTION_INFO_V1(get_current_transaction_id);
+PG_FUNCTION_INFO_V1(get_all_active_transactions);
+
+
+/*
+ * assign_distributed_transaction_id updates the shared memory allocated for this backend
+ * and sets initiatorNodeIdentifier, transactionNumber, timestamp fields with the given
+ * inputs. Also, the function sets the database id and process id via the information that
+ * Postgres provides.
+ *
+ * This function is only intended for internal use for managing distributed transactions.
+ * Users should not use this function for any purpose.
+ */
+Datum
+assign_distributed_transaction_id(PG_FUNCTION_ARGS)
+{
+	CheckCitusVersion(ERROR);
+
+	/* MyBackendData should always be avaliable, just out of paranoia */
+	if (!MyBackendData)
+	{
+		ereport(ERROR, (errmsg("backend is not ready for distributed transactions")));
+	}
+
+	SpinLockAcquire(&MyBackendData->mutex);
+
+	/* if an id is already assigned, release the lock and error */
+	if (MyBackendData->transactionId.initiatorNodeIdentifier != 0)
+	{
+		SpinLockRelease(&MyBackendData->mutex);
+
+		ereport(ERROR, (errmsg("the backend has already been assigned a "
+							   "transaction id")));
+	}
+
+	MyBackendData->databaseId = MyDatabaseId;
+
+	MyBackendData->transactionId.initiatorNodeIdentifier = PG_GETARG_INT32(0);
+	MyBackendData->transactionId.transactionNumber = PG_GETARG_INT64(1);
+	MyBackendData->transactionId.timestamp = PG_GETARG_TIMESTAMPTZ(2);
+
+	SpinLockRelease(&MyBackendData->mutex);
+
+	PG_RETURN_VOID();
+}
+
+
+/*
+ * get_current_transaction_id returns a tuple with (databaseId, processId,
+ * initiatorNodeIdentifier, transactionNumber, timestamp) that exists in the
+ * shared memory associated with this backend. Note that if the backend
+ * is not in a transaction, the function returns uninitialized data where
+ * transactionNumber equals to 0.
+ */
+Datum
+get_current_transaction_id(PG_FUNCTION_ARGS)
+{
+	TupleDesc tupleDescriptor = NULL;
+	HeapTuple heapTuple = NULL;
+
+	const int attributeCount = 5;
+	Datum values[attributeCount];
+	bool isNulls[attributeCount];
+
+	DistributedTransactionId *distributedTransctionId = NULL;
+
+	CheckCitusVersion(ERROR);
+
+	/* build a tuple descriptor for our result type */
+	if (get_call_result_type(fcinfo, NULL, &tupleDescriptor) != TYPEFUNC_COMPOSITE)
+	{
+		elog(ERROR, "return type must be a row type");
+	}
+
+	/* MyBackendData should always be avaliable, just out of paranoia */
+	if (!MyBackendData)
+	{
+		ereport(ERROR, (errmsg("backend is not ready for distributed transactions")));
+	}
+
+	distributedTransctionId = GetCurrentDistributedTransctionId();
+
+	memset(values, 0, sizeof(values));
+	memset(isNulls, false, sizeof(isNulls));
+
+	/* first two fields do not change for this backend, so get directly */
+	values[0] = ObjectIdGetDatum(MyDatabaseId);
+	values[1] = Int32GetDatum(MyProcPid);
+
+	values[2] = Int32GetDatum(distributedTransctionId->initiatorNodeIdentifier);
+	values[3] = UInt64GetDatum(distributedTransctionId->transactionNumber);
+
+	/* provide a better output */
+	if (distributedTransctionId->initiatorNodeIdentifier != 0)
+	{
+		values[4] = TimestampTzGetDatum(distributedTransctionId->timestamp);
+	}
+	else
+	{
+		isNulls[4] = true;
+	}
+
+	heapTuple = heap_form_tuple(tupleDescriptor, values, isNulls);
+
+	PG_RETURN_DATUM(HeapTupleGetDatum(heapTuple));
+}
+
+
+/*
+ * get_all_active_transactions returns all the avaliable information about all
+ * the active backends.
+ */
+Datum
+get_all_active_transactions(PG_FUNCTION_ARGS)
+{
+	ReturnSetInfo *returnSetInfo = (ReturnSetInfo *) fcinfo->resultinfo;
+	TupleDesc tupleDescriptor = NULL;
+	Tuplestorestate *tupleStore = NULL;
+	MemoryContext perQueryContext = NULL;
+	MemoryContext oldContext = NULL;
+
+	int backendIndex = 0;
+
+	const int attributeCount = 5;
+	Datum values[attributeCount];
+	bool isNulls[attributeCount];
+
+	CheckCitusVersion(ERROR);
+
+	/* check to see if caller supports us returning a tuplestore */
+	if (returnSetInfo == NULL || !IsA(returnSetInfo, ReturnSetInfo))
+	{
+		ereport(ERROR,
+				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+				 errmsg("set-valued function called in context " \
+						"that cannot accept a set")));
+	}
+
+	if (!(returnSetInfo->allowedModes & SFRM_Materialize))
+	{
+		ereport(ERROR,
+				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+				 errmsg("materialize mode required, but it is not " \
+						"allowed in this context")));
+	}
+
+	/* build a tuple descriptor for our result type */
+	if (get_call_result_type(fcinfo, NULL, &tupleDescriptor) != TYPEFUNC_COMPOSITE)
+	{
+		elog(ERROR, "return type must be a row type");
+	}
+
+	perQueryContext = returnSetInfo->econtext->ecxt_per_query_memory;
+
+	oldContext = MemoryContextSwitchTo(perQueryContext);
+
+	tupleStore = tuplestore_begin_heap(true, false, work_mem);
+	returnSetInfo->returnMode = SFRM_Materialize;
+	returnSetInfo->setResult = tupleStore;
+	returnSetInfo->setDesc = tupleDescriptor;
+
+	MemoryContextSwitchTo(oldContext);
+
+	/*
+	 * We don't want to initialize memory while spinlock is held so we
+	 * prefer to do it here. This initialization is done only for the first
+	 * row.
+	 */
+	memset(values, 0, sizeof(values));
+	memset(isNulls, false, sizeof(isNulls));
+
+	/* we're reading all the backend data, take a lock to prevent concurrent additions */
+	LWLockAcquire(AddinShmemInitLock, LW_SHARED);
+
+	for (backendIndex = 0; backendIndex < MaxBackends; ++backendIndex)
+	{
+		BackendData *currentBackend =
+			&backendManagementShmemData->backends[backendIndex];
+
+		SpinLockAcquire(&currentBackend->mutex);
+
+		/* we're only interested in active backends */
+		if (currentBackend->transactionId.transactionNumber == 0)
+		{
+			SpinLockRelease(&currentBackend->mutex);
+			continue;
+		}
+
+		values[0] = ObjectIdGetDatum(currentBackend->databaseId);
+		values[1] = Int32GetDatum(ProcGlobal->allProcs[backendIndex].pid);
+		values[2] = Int32GetDatum(currentBackend->transactionId.initiatorNodeIdentifier);
+		values[3] = UInt64GetDatum(currentBackend->transactionId.transactionNumber);
+		values[4] = TimestampTzGetDatum(currentBackend->transactionId.timestamp);
+
+		SpinLockRelease(&currentBackend->mutex);
+
+		tuplestore_putvalues(tupleStore, tupleDescriptor, values, isNulls);
+
+		/*
+		 * We don't want to initialize memory while spinlock is held so we
+		 * prefer to do it here. This initialization is done for the rows
+		 * starting from the second one.
+		 */
+		memset(values, 0, sizeof(values));
+		memset(isNulls, false, sizeof(isNulls));
+	}
+
+	LWLockRelease(AddinShmemInitLock);
+
+	/* clean up and return the tuplestore */
+	tuplestore_donestoring(tupleStore);
+
+	PG_RETURN_VOID();
+}
+
+
+/*
+ * InitializeBackendManagement requests the necessary shared memory
+ * from Postgres and sets up the shared memory startup hook.
+ */
+void
+InitializeBackendManagement(void)
+{
+	/* allocate shared memory */
+	RequestAddinShmemSpace(BackendManagementShmemSize());
+
+	prev_shmem_startup_hook = shmem_startup_hook;
+	shmem_startup_hook = BackendManagementShmemInit;
+}
+
+
+/*
+ * BackendManagementShmemInit is the callback that is to be called on shared
+ * memory startup hook. The function sets up the necessary shared memory
+ * segment for the backend manager.
+ */
+static void
+BackendManagementShmemInit(void)
+{
+	bool alreadyInitialized = false;
+
+	/* we may update the shmem, acquire lock exclusively */
+	LWLockAcquire(AddinShmemInitLock, LW_EXCLUSIVE);
+
+	backendManagementShmemData =
+		(BackendManagementShmemData *) ShmemInitStruct(
+			"Backend Management Shmem",
+			BackendManagementShmemSize(),
+			&alreadyInitialized);
+
+	if (!alreadyInitialized)
+	{
+		int backendIndex = 0;
+		char *trancheName = "Backend Management Tranche";
+
+#if (PG_VERSION_NUM >= 100000)
+		NamedLWLockTranche *namedLockTranche =
+			&backendManagementShmemData->namedLockTranche;
+
+#else
+		LWLockTranche *lockTranche = &backendManagementShmemData->lockTranche;
+#endif
+
+		/* start by zeroing out all the memory */
+		memset(backendManagementShmemData, 0,
+			   BackendManagementShmemSize());
+
+#if (PG_VERSION_NUM >= 100000)
+		namedLockTranche->trancheId = LWLockNewTrancheId();
+
+		LWLockRegisterTranche(namedLockTranche->trancheId, trancheName);
+		LWLockInitialize(&backendManagementShmemData->lock,
+						 namedLockTranche->trancheId);
+#else
+		backendManagementShmemData->trancheId = LWLockNewTrancheId();
+
+		/* we only need a single lock */
+		lockTranche->array_base = &backendManagementShmemData->lock;
+		lockTranche->array_stride = sizeof(LWLock);
+		lockTranche->name = trancheName;
+
+		LWLockRegisterTranche(backendManagementShmemData->trancheId, lockTranche);
+		LWLockInitialize(&backendManagementShmemData->lock,
+						 backendManagementShmemData->trancheId);
+#endif
+
+		/* start the distributed transaction ids from 1 */
+		pg_atomic_init_u64(&backendManagementShmemData->nextTransactionNumber, 1);
+
+		/*
+		 * We need to init per backend's spinlock before any backend
+		 * starts its execution.
+		 */
+		for (backendIndex = 0; backendIndex < MaxBackends; ++backendIndex)
+		{
+			SpinLockInit(&backendManagementShmemData->backends[backendIndex].mutex);
+		}
+	}
+
+	LWLockRelease(AddinShmemInitLock);
+
+	if (prev_shmem_startup_hook != NULL)
+	{
+		prev_shmem_startup_hook();
+	}
+}
+
+
+/*
+ * BackendManagementShmemSize returns the size that should be allocated
+ * on the shared memory for backend management.
+ */
+static size_t
+BackendManagementShmemSize(void)
+{
+	Size size = 0;
+
+	size = add_size(size, sizeof(BackendManagementShmemData));
+	size = add_size(size, mul_size(sizeof(BackendData), MaxBackends));
+
+	return size;
+}
+
+
+/*
+ *  InitializeBackendData is called per backend and does the
+ *  required initialization.
+ */
+void
+InitializeBackendData(void)
+{
+	MyBackendData = &backendManagementShmemData->backends[MyProc->pgprocno];
+
+	Assert(MyBackendData);
+
+	SpinLockAcquire(&MyBackendData->mutex);
+
+	MyBackendData->databaseId = MyDatabaseId;
+	MyBackendData->transactionId.initiatorNodeIdentifier = 0;
+	MyBackendData->transactionId.transactionNumber = 0;
+	MyBackendData->transactionId.timestamp = 0;
+
+	SpinLockRelease(&MyBackendData->mutex);
+}
+
+
+/*
+ * UnSetDistributedTransactionId simply acquires the mutex and resets the backend's
+ * distributed transaction data in shared memory to the initial values.
+ */
+void
+UnSetDistributedTransactionId(void)
+{
+	/* backend does not exist if the extension is not created */
+	if (MyBackendData)
+	{
+		SpinLockAcquire(&MyBackendData->mutex);
+
+		MyBackendData->databaseId = 0;
+		MyBackendData->transactionId.initiatorNodeIdentifier = 0;
+		MyBackendData->transactionId.transactionNumber = 0;
+		MyBackendData->transactionId.timestamp = 0;
+
+		SpinLockRelease(&MyBackendData->mutex);
+	}
+}
+
+
+/*
+ * GetCurrentDistributedTransctionId reads the backend's distributed transaction id and
+ * returns a copy of it.
+ */
+DistributedTransactionId *
+GetCurrentDistributedTransctionId(void)
+{
+	DistributedTransactionId *currentDistributedTransactionId =
+		(DistributedTransactionId *) palloc(sizeof(DistributedTransactionId));
+
+	SpinLockAcquire(&MyBackendData->mutex);
+
+	currentDistributedTransactionId->initiatorNodeIdentifier =
+		MyBackendData->transactionId.initiatorNodeIdentifier;
+	currentDistributedTransactionId->transactionNumber =
+		MyBackendData->transactionId.transactionNumber;
+	currentDistributedTransactionId->timestamp =
+		MyBackendData->transactionId.timestamp;
+
+	SpinLockRelease(&MyBackendData->mutex);
+
+	return currentDistributedTransactionId;
+}
+
+
+/*
+ * AssignDistributedTransactionId generates a new distributed transaction id and
+ * sets it for the current backend. It also sets the databaseId and
+ * processId fields.
+ *
+ * This function should only be called on BeginCoordinatedTransaction(). Any other
+ * callers is very likely to break the distributed transction management.
+ */
+void
+AssignDistributedTransactionId(void)
+{
+	pg_atomic_uint64 *transactionNumberSequence =
+		&backendManagementShmemData->nextTransactionNumber;
+
+	uint64 nextTransactionNumber = pg_atomic_fetch_add_u64(transactionNumberSequence, 1);
+	int localGroupId = GetLocalGroupId();
+	TimestampTz currentTimestamp = GetCurrentTimestamp();
+
+	SpinLockAcquire(&MyBackendData->mutex);
+
+	MyBackendData->databaseId = MyDatabaseId;
+
+	MyBackendData->transactionId.initiatorNodeIdentifier = localGroupId;
+	MyBackendData->transactionId.transactionNumber =
+		nextTransactionNumber;
+	MyBackendData->transactionId.timestamp = currentTimestamp;
+
+	SpinLockRelease(&MyBackendData->mutex);
+}
--- a/src/backend/distributed/transaction/remote_transaction.c
+++ b/src/backend/distributed/transaction/remote_transaction.c
@@ -15,6 +15,7 @@
 #include "miscadmin.h"
 
 #include "access/xact.h"
+#include "distributed/backend_data.h"
 #include "distributed/connection_management.h"
 #include "distributed/metadata_cache.h"
 #include "distributed/remote_commands.h"
@@ -32,12 +33,16 @@ static void WarnAboutLeakedPreparedTransaction(MultiConnection *connection, bool
 
 /*
  * StartRemoteTransactionBeging initiates beginning the remote transaction in
- * a non-blocking manner.
+ * a non-blocking manner. The function sends "BEGIN" followed by
+ * assign_distributed_transaction_id() to assign the distributed transaction
+ * id on the remote node.
  */
 void
 StartRemoteTransactionBegin(struct MultiConnection *connection)
 {
 	RemoteTransaction *transaction = &connection->remoteTransaction;
+	StringInfo beginAndSetDistributedTransactionId = makeStringInfo();
+	DistributedTransactionId *distributedTransactionId = NULL;
 
 	Assert(transaction->transactionState == REMOTE_TRANS_INVALID);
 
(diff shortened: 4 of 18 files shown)


anarazel:
> Shall we call UnSetDistributedTransactionId() on PREPARE TRANSACTION? Probably not?
> I don't see any problems with 2PC apart from the above question. Does anyone see?

We probably want to - once a transaction is prepared successfully it can't be waiting for anything anymore.

> How are we going to detect deadlocks that are formed without coordinated transactions? (given that we do not set transaction ids for those transactions.)

Because there's no explicit transaction on the workers, and because there's only a single statement, it's not really possible to be the "cause" of a distributed deadlock. It can be part of a wait graph, but only a deadlock if there's "real" distributed transactions involved (that's why we've avoided issues with distributed deadlocks so far - we've not allowed multiple statements in a transaction spanning multiple nodes).  Which means it's ok not to assign transaction identifiers on workers, that means such statements will never be cancelled, but we can detect the deadlocks involving them.    Makes sense?

src/backend/distributed/master/master_stage_protocol.c
```diff
@@ -26,6 +26,7 @@
 #include "catalog/namespace.h"
 #include "distributed/colocation_utils.h"
 #include "distributed/connection_management.h"
+#include "distributed/coordinated_transaction_management.h"
```
anarazel: I'm not a big fan of this rename...

src/backend/distributed/transaction/distributed_transaction_management.c
```diff
@@ -0,0 +1,433 @@
...
+
+	pg_atomic_uint64 nextTransactionId;
+
+	DistributedTransactionBackendData sessions[FLEXIBLE_ARRAY_MEMBER];
+} DistributedTransactionShmemData;
+
+
+static shmem_startup_hook_type prev_shmem_startup_hook = NULL;
+static DistributedTransactionShmemData *distributedTransactionShmemData = NULL;
+static DistributedTransactionBackendData *DistributedTransactionBackend = NULL;
```
anarazel: I find My* a lot clearer (and follows precedent) - just "backend" could be any backend, not just the current one.

src/backend/distributed/transaction/distributed_transaction_management.c
```diff
@@ -0,0 +1,433 @@
...
+slock_t BackendDistributedTransactionMutex = '\0';
```
anarazel: Huh, why do we have a non-static spinlock here? Did you introduce a single spinlock covering everything? That'd be a bad idea. Also, how does a spinlock in per-process memory even work?

src/backend/distributed/transaction/distributed_transaction_management.c
```diff
@@ -0,0 +1,433 @@
...
+static uint64
+GetNextLocalTransactionIdFromShmem(void)
+{
+	pg_atomic_uint64 *transactionIdSequence =
+		&distributedTransactionShmemData->nextTransactionId;
+	uint64 nextTransactionId = 0;
+
+	nextTransactionId = pg_atomic_fetch_add_u64(transactionIdSequence, 1);
+
+	return nextTransactionId;
```
anarazel: Personally this'd be a lot easier to read if you'd just skip all the local variables - they don't help...

src/include/distributed/distributed_transaction_management.h
```diff
@@ -0,0 +1,52 @@
...
+
+/*
+ * Each backend's active distributed transaction information is tracked via
+ * DistributedTransactionBackendData on the shared memory.
+ */
+typedef struct DistributedTransactionBackendData
+{
+	Oid databaseId;
+	DistributedTransactionId transactionId;
+} DistributedTransactionBackendData;
```
anarazel: The spinlock should live here, per-backend, rather than in process local memory.

src/backend/distributed/transaction/distributed_transaction_management.c
```diff
@@ -0,0 +1,433 @@
...
+ * and sets initiatorNodeIdentifier, transactionId, timestamp fields with the given
+ * inputs. Also, the function sets the database id via MyDatabaseId variable provided
+ * by "misadmmin.h".
+ */
+Datum
+assign_distributed_transaction_id(PG_FUNCTION_ARGS)
+{
+	CheckCitusVersion(ERROR);
+
+	/* TODO: Is assert OK? */
```
anarazel: Should be, yes. Perhaps make it an ereport(ERROR) instead, just out of paranoia?

src/backend/distributed/transaction/distributed_transaction_management.c
```diff
@@ -0,0 +1,433 @@
...
+
+
+/*
+ * get_distributed_transaction_id returns a tuple with (databaseId, initiatorNodeIdentifier,
+ * transactionId, timestamp) that exists in the shared memory associated with this backend.
+ */
+Datum
+get_distributed_transaction_id(PG_FUNCTION_ARGS)
+{
+	Datum distributedTransactionId = 0;
```
anarazel: Should check version.

src/backend/distributed/transaction/distributed_transaction_management.c
```diff
@@ -0,0 +1,433 @@
...
+	}
+
+	return activeTransactionList;
+}
+
+
+/*
+ * GenerateDistributedTransactionIdTuple returns a datum of psudo-generated heaptuple
+ * for the distributed transaction id provided by the parameters.
+ */
```
anarazel: What's a "psudo (sic)-generated heaptuple"?

src/backend/distributed/transaction/distributed_transaction_management.c
```diff
@@ -0,0 +1,433 @@
...
+
+	heapTuple = heap_form_tuple(tupleDescriptor, values, isNulls);
+	distributedTransactionIdDatum = HeapTupleGetDatum(heapTuple);
+
+	return distributedTransactionIdDatum;
+}
+
+
+/*
+ * GenerateDistributedTransactionTupleDesc returns a psudo tuple descriptor
```
anarazel: ^ - there's nothing pseudo at play here.

src/include/distributed/distributed_transaction_management.h
```diff
@@ -0,0 +1,52 @@
...
+
+
```
anarazel: I think I'd rather move these structs to a field like 'DistBackendData' or such, and in a similarly named file. We'll have to extend the data stored here pretty soon (within the deadlock detection project, to emit useful error messages), and we shouldn't have parallel arrays for that.

src/backend/distributed/transaction/distributed_transaction_management.c
onderkalaci: Ops, I agree that this is a critical bug. It was intended to be `static`. But, I'll move it to backend data structure, as you suggested.

anarazel: Even if it were static, a spinlock that's not in shared memory (and static variables aren't) doesn't actually do anything really interesting (it serves as a memory barrier, but that's it).

src/backend/distributed/transaction/distributed_transaction_management.c
```diff
@@ -0,0 +1,433 @@
...
+GenerateNextDistributedTransactionId(void)
+{
+	DistributedTransactionId *nextDistributedTransactionId =
+		(DistributedTransactionId *) palloc(sizeof(DistributedTransactionId));
+
+	/*
+	 * FIXME: Once we allow running queries from secondaries, we should generate
+	 * an id which is unique to the node, not to the group as we do here.
+	 */
+	nextDistributedTransactionId->initiatorNodeIdentifier = (uint64) GetLocalGroupId();
```
marcocitus: If we expect to change this it'd be nice if there was a `GetLocalNodeId()`.

src/backend/distributed/transaction/distributed_transaction_management.c
onderkalaci: The check (`CheckCitusVersion `) is already below.

onderkalaci:
@marcocitus @anarazel I've rebased #1447 onto this branch (ended up `distributed_id_and_deadlock_detection` branch), which successfully detects deadlocks and errors out one of the distributed transactions.

src/backend/distributed/connection/remote_commands.c
```diff
@@ -337,10 +337,12 @@ ExecuteOptionalRemoteCommand(MultiConnection *connection, const char *command,
...
 
 /*
- * SendRemoteCommand is a PQsendQuery wrapper that logs remote commands, and
+ * SendRemoteCommand is a PQsendQueryParams wrapper that logs remote commands, and
  * accepts a MultiConnection instead of a plain PGconn.  It makes sure it can
  * send commands asynchronously without blocking (at the potential expense of
- * an additional memory allocation).
+ * an additional memory allocation). Note that since the function utilizes
+ * PQsendQueryParams() as the asynchronous command processing API, the command string
+ * could only include a single command.
```
anarazel: s/could/can/. I'd also just shorten the whole sentence to 'The command string can only include a single command', the reasoning isn't really relevant. If you want to hint at it, I'd just refer to `PQsendQueryParams` rather than including 'as the asynchronous command processing API' which isn't helpful.

src/backend/distributed/connection/remote_commands.c
```diff
@@ -374,12 +376,32 @@ SendRemoteCommandParams(MultiConnection *connection, const char *command,
  * SendRemoteCommand is a PQsendQuery wrapper that logs remote commands, and
  * accepts a MultiConnection instead of a plain PGconn.  It makes sure it can
  * send commands asynchronously without blocking (at the potential expense of
- * an additional memory allocation).
+ * an additional memory allocation). Note that since the function utilizes
+ * PQsendQuery() as the asynchronous command processing API, the command string
+ * could include multiple commands.
  */
```
anarazel: Adapt from above.

src/backend/distributed/test/distributed_transaction_id.c
```diff
@@ -0,0 +1,117 @@
...
+	{
+		SRF_RETURN_DONE(functionContext);
+	}
+}
+
+
+/*
+ * CompareDistributedTransactionIds is a test function, which only sorts elements
+ * wrt transactionId field only.
+ */
```
anarazel: Why?

src/backend/distributed/transaction/coordinated_transaction_management.c
```diff
@@ -73,6 +74,9 @@ BeginCoordinatedTransaction(void)
 	}
 
 	CurrentCoordinatedTransactionState = COORD_TRANS_STARTED;
+
+	/* this is crucial for managing distributed transaction ids */
+	GenerateAndSetNextDistributedTransactionId();
```
anarazel: That comment doesn't actually add anything...  I'd maybe rename the function to AssignDistributedTransactionId() - the "next" bit is superflous to know about at this level.

src/backend/distributed/transaction/coordinated_transaction_management.c
```diff
@@ -20,9 +20,10 @@
 #include "access/twophase.h"
 #include "access/xact.h"
 #include "distributed/connection_management.h"
+#include "distributed/coordinated_transaction_management.h"
+#include "distributed/distributed_transaction_management.h"
```
anarazel: I think we should just leave this in one header, I don't see what we gain by this split.

src/backend/distributed/transaction/distributed_transaction_management.c
```diff
@@ -0,0 +1,483 @@
...
+
+	PG_RETURN_DATUM(distributedTransactionIdDatum);
+}
+
+
+/*
+ * GetAllActiveDistributedTransactions returns all the active distributed
+ * transactions on the shared memory.
+ *
+ * Note that the function could not gurantee a global consistent result, the function
```
anarazel: s/could not/can/ or s/could not/does not/

src/backend/distributed/transaction/distributed_transaction_management.c
```diff
@@ -0,0 +1,483 @@
...
+	for (sessionIndex = 0; sessionIndex < MaxBackends; ++sessionIndex)
+	{
+		DistributedTransactionBackendData *currentBackend =
+			&distributedTransactionShmemData->sessions[sessionIndex];
+
+		if (currentBackend && currentBackend->transactionId.transactionId != 0)
+		{
+			DistributedTransactionBackendData *activeBackend =
+				(DistributedTransactionBackendData *) palloc(
+					sizeof(DistributedTransactionBackendData));
```
anarazel: This looks neigh on unreadable.

src/backend/distributed/transaction/distributed_transaction_management.c
```diff
@@ -0,0 +1,483 @@
...
+					   "initiatorNodeIdentifier", INT8OID, -1, 0);
+	TupleDescInitEntry(distributedTransactionTupleDesc, (AttrNumber) 3, "transactionId",
+					   INT8OID, -1, 0);
+	TupleDescInitEntry(distributedTransactionTupleDesc, (AttrNumber) 4, "timestamp",
+					   TIMESTAMPTZOID, -1, 0);
+
+	/* let Postgres learn about this tuple descriptor */
+	BlessTupleDesc(distributedTransactionTupleDesc);
+
+	return distributedTransactionTupleDesc;
```
anarazel: We shouldn't need this - if you instead define the SRF the way I did in my prototype for the dump* functions, then you get a system provided one.

src/backend/distributed/transaction/distributed_transaction_management.c
```diff
@@ -0,0 +1,483 @@
...
+
+	Assert(MyDistributedTransactionBackend);
+
+	MyDistributedTransactionBackend->databaseId = 0;
+	MyDistributedTransactionBackend->transactionId.initiatorNodeIdentifier = 0;
+	MyDistributedTransactionBackend->transactionId.transactionId = 0;
+	MyDistributedTransactionBackend->transactionId.timestamp = 0;
+
+	/* we'll use mutex while accessing the backend shmem */
+	SpinLockInit(&MyDistributedTransactionBackend->mutex);
```
anarazel: This isn't a good idea - this should be initialized *once* in `DistributedTransactionManagementShmemInit()`. Consider what otherwise happens if somebody is currently looking at this entry - you're destroying the spinlock's internal state.   You also should hold the spinlock here while resetting the data.

src/backend/distributed/transaction/distributed_transaction_management.c
```diff
@@ -0,0 +1,483 @@
...
+}
+
+
+/*
+ * GenerateNextDistributedTransactionId returns a new distributed transaction id with
+ * the current timestamp and next local transaction id.
+ */
+static DistributedTransactionId *
+GenerateNextDistributedTransactionId(void)
+{
```
anarazel: I think the API split between `GenerateAndSetNextDistributedTransactionId()`, `GenerateNextDistributedTransactionId()`, `GetNextLocalTransactionIdFromShmem()` is a bad idea. You're essentially redoing all the work done in both, adding an extra palloc and a bunch of copying to the mix.  I think we should just collapse all these back into one function.

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -32,12 +33,17 @@ static void WarnAboutLeakedPreparedTransaction(MultiConnection *connection, bool
 
 /*
  * StartRemoteTransactionBeging initiates beginning the remote transaction in
- * a non-blocking manner.
+ * a non-blocking manner. The function sends "BEGIN" followed by a function call
+ * (i.e., assign_distributed_transaction_id()) to assign the distributed transaction
+ * id on the remote node.
```
anarazel: I'd just remove the "followed by a function call" and unparenthesize the actual function - no info added.

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -51,8 +57,24 @@ StartRemoteTransactionBegin(struct MultiConnection *connection)
...
-						   "BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED"))
+	beginCommand = "BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED";
+
+	/*
+	 * Append BEGIN and assign_distributed_transaction_id() statements into a single command
+	 * and send both in one step. The reason is purely performance, we don't want
+	 * seperate roundtrips for these two statements.
+	 */
+	appendStringInfo(beginAndSetDistributedTransactionId, "%s;", beginCommand);
+
```
anarazel: `beginCommand` as a separate string seems fairly useless here (and you're removing a const, because "" literals are const, but `beginCommand` is not). Just put into the `appendStringInfo()` call?

src/include/distributed/distributed_transaction_management.h
anarazel: > I couldn't get this request, so I'll wait to change this until I fully understand what you suggest.

I don't think the struct should a) have transaction in the name b) be in a file with transaction in the name. We're going to add more per-backend data, so we shouldn't tie this to transactions.

> Also, I was excepting that the error message goes to DistributedTransactionShmemData struct?

Hm? How would that work - there can be multiple transactions being canceled at the same time...

src/backend/distributed/transaction/distributed_transaction_management.c
anarazel: > Given that there is no table with such a tuple, I mentioned that as pseudo-generated heaptuple. Is that an inaccurate terminology?

Imo yes. It's not something that exists. It's just a heaptuple.

anarazel: generated: All heap tuples are "generated"
pseudo: It can be returned to the caller and the user, so it's not really pseudo or some such. It's not faking being a tuple.

src/backend/distributed/transaction/distributed_transaction_management.c
onderkalaci: Hmm, that'd cost us to iterate `MaxBackend` times on `shmem` initialization. But, that should be fine.

Well, I was thinking that it should be OK to init here since the backend has just started and no other callers would exist. However, deadlock detection logic might try to read it and we could be in a race situation.

I'm also removing `LWLockAcquire(&distributedTransactionShmemData->lock, LW_EXCLUSIVE);` and its complement, since they seem useless here

src/backend/distributed/test/distributed_transaction_id.c
```diff
@@ -0,0 +1,167 @@
...
+
+PG_FUNCTION_INFO_V1(get_all_active_distributed_transaction_ids);
+
+
+/*
+ * get_all_active_distributed_transaction_ids returns all the
+ * active distributed transaction ids.
+ */
+Datum
+get_all_active_distributed_transaction_ids(PG_FUNCTION_ARGS)
```
marcocitus: Could we create this as part of the extension? This seems like an incredibly useful function.

onderkalaci:
> Because there's no explicit transaction on the workers, and because there's only a single statement, it's not really possible to be the "cause" of a distributed deadlock. It can be part of a wait graph, but only a deadlock if there's "real" distributed transactions involved (that's why we've avoided issues with distributed deadlocks so far - we've not allowed multiple statements in a transaction spanning multiple nodes). Which means it's ok not to assign transaction identifiers on workers, that means such statements will never be cancelled, but we can detect the deadlocks involving them. Makes sense?

Yes, all makes sense. I couldn't come up with an example (with `fkey/partitioning` etc as well) where a distributed deadlock cannot be resolved because of single statement queries.

src/backend/distributed/test/distributed_transaction_id.c
onderkalaci: We can do that. I'll update the PR tomorrow.

src/backend/distributed/transaction/distributed_transaction_management.c
```diff
@@ -0,0 +1,431 @@
...
+		 * We should init per backend's spinlock before any backend
+		 * starts its execution.
+		 */
+		for (backendIndex = 0; backendIndex < MaxBackends; ++backendIndex)
+		{
+			slock_t backendMutex =
+				distributedTransactionShmemData->sessions[backendIndex].mutex;
+
+			SpinLockInit(&backendMutex);
+		}
```
anarazel: This looks quite wrong - you're *copying* the spinlock into local memory, and then you're initializing that copy. IOW, this does't do anything lasting.

src/backend/distributed/transaction/distributed_transaction_management.c
```diff
@@ -0,0 +1,431 @@
...
+	if (!(returnSetInfo->allowedModes & SFRM_Materialize))
+	{
+		ereport(ERROR,
+				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+				 errmsg("materialize mode required, but it is not " \
+						"allowed in this context")));
+	}
+
+	/* build a tuple descriptor for our result type */
+	if (get_call_result_type(fcinfo, NULL, &tupleDescriptor) != TYPEFUNC_COMPOSITE)
```
anarazel: I don't quite understand why you'd copied the SRF code, rather than just the `get_call_result_type()` bit, which should be all that you need?

src/backend/distributed/transaction/distributed_transaction_management.c
```diff
@@ -0,0 +1,431 @@
...
+	values[0] = ObjectIdGetDatum(MyDatabaseId);
+	values[1] = UInt64GetDatum(distributedTransctionId->initiatorNodeIdentifier);
+	values[2] = UInt64GetDatum(distributedTransctionId->transactionId);
+	values[3] = TimestampTzGetDatum(distributedTransctionId->timestamp);
+
+	tuplestore_putvalues(tupleStore, tupleDescriptor, values, isNulls);
+
+	/* clean up and return the tuplestore */
+	tuplestore_donestoring(tupleStore);
+
```
anarazel: I think you should just `heap_form_tuple()` with the result tupdesc and return the result.

src/backend/distributed/connection/remote_commands.c
```diff
@@ -337,10 +337,11 @@ ExecuteOptionalRemoteCommand(MultiConnection *connection, const char *command,
...
 
 /*
- * SendRemoteCommand is a PQsendQuery wrapper that logs remote commands, and
- * accepts a MultiConnection instead of a plain PGconn.  It makes sure it can
+ * SendRemoteCommandParams is a PQsendQueryParams wrapper that logs remote commands,
+ * and accepts a MultiConnection instead of a plain PGconn. It makes sure it can
  * send commands asynchronously without blocking (at the potential expense of
- * an additional memory allocation).
+ * an additional memory allocation). The command string can only include a single
+ * command since PQsendQueryParams() supports that.
```
anarazel: supports *only* that?

src/backend/distributed/test/distributed_transaction_id.c
```diff
@@ -0,0 +1,167 @@
...
+ *
+ * This file contains functions to exercise distributed transaction id
+ * functionality within Citus.
+ *
+ * Copyright (c) 2017, Citus Data, Inc.
+ *
+ *-------------------------------------------------------------------------
+ */
+#include "postgres.h"
+#include "c.h"
```
anarazel: don't include c.h in backend files.

src/backend/distributed/test/distributed_transaction_id.c
```diff
@@ -0,0 +1,167 @@
...
+	ReturnSetInfo *returnSetInfo = (ReturnSetInfo *) fcinfo->resultinfo;
+	TupleDesc tupleDescriptor = NULL;
+	Tuplestorestate *tupleStore = NULL;
+	MemoryContext perQueryContext = NULL;
+	MemoryContext oldContext = NULL;
+
+	List *activeTransactionList = NULL;
+	ListCell *activeTransactionCell = NULL;
+
+	int attributeCount = 4;
```
anarazel: Should be const if you use it to size arrays.

src/backend/distributed/test/distributed_transaction_id.c
anarazel: I assume you're going to to that anyway, but to be sure: If you move, it it shouldn't be in test/* ;)

src/backend/distributed/test/distributed_transaction_id.c
```diff
@@ -0,0 +1,167 @@
...
+	foreach(activeTransactionCell, activeTransactionList)
+	{
+		DistributedTransactionBackendData *backendData =
+			(DistributedTransactionBackendData *) lfirst(activeTransactionCell);
+		DistributedTransactionId *distributedTransctionId = &backendData->transactionId;
+
+		memset(values, 0, sizeof(values));
+		memset(isNulls, false, sizeof(isNulls));
+
+		values[0] = ObjectIdGetDatum(backendData->databaseId);
```
anarazel: If it's for more than just testing, we should expose the pid too.

src/backend/distributed/transaction/coordinated_transaction_management.c
anarazel: I think the split isn't actually gaining us clarity, and coordinated and distributed transactions are pretty much the same thing, and what currently resides in distributed_transaction_management.h isn't really what the name says. So I don't think the split makes much sense as is. I think I'd be ok with introducing new 'distributed_transaction_identifier' files or such.

src/backend/distributed/transaction/distributed_transaction_management.c
```diff
@@ -0,0 +1,431 @@
...
+#else
+	LWLockTranche lockTranche;
+#endif
+	LWLock lock;
+
+	/*
+	 * We prefer to use an atomic integer over sequences for two
+	 * reasons (i) orders of magnitude performance difference
+	 * (ii) allowing read-only replicas to be able to generate ids
+	 */
```
anarazel: I personally think the argument normally would rather be the other way round: Why do we need a relatively expensive persistent object to store something, if not needed... ;)

src/backend/distributed/transaction/distributed_transaction_management.c
```diff
@@ -0,0 +1,431 @@
...
+PG_FUNCTION_INFO_V1(assign_distributed_transaction_id);
+PG_FUNCTION_INFO_V1(get_distributed_transaction_id);
+
+
+/*
+ * assign_distributed_transaction_id updates the shared memory allocated for this backend
+ * and sets initiatorNodeIdentifier, transactionId, timestamp fields with the given
+ * inputs. Also, the function sets the database id via MyDatabaseId variable provided
+ * by "misadmmin.h".
+ */
```
anarazel: Do we need to document somehow, perhaps by adding _internal or such into the name, that this never should/needs to be called by the user?

src/backend/distributed/transaction/distributed_transaction_management.c
```diff
@@ -0,0 +1,431 @@
...
+
+	/* MyDistributedTransactionBackend should always be avaliable, just out of paranoia */
+	if (!MyDistributedTransactionBackend)
+	{
+		ereport(ERROR, (errmsg("backend is not ready for distributed transactions")));
+	}
+
+	SpinLockAcquire(&MyDistributedTransactionBackend->mutex);
+
+	MyDistributedTransactionBackend->databaseId = MyDatabaseId;
```
anarazel: Should we error out if one already is assigned (can do the read without spinlock, nobody else can write)?

src/backend/distributed/transaction/distributed_transaction_management.c
```diff
@@ -0,0 +1,431 @@
...
+{
+	int sessionIndex = 0;
+	List *activeTransactionList = NIL;
+
+	for (sessionIndex = 0; sessionIndex < MaxBackends; ++sessionIndex)
+	{
+		DistributedTransactionBackendData *currentBackend =
+			&distributedTransactionShmemData->sessions[sessionIndex];
+
+		if (currentBackend && currentBackend->transactionId.transactionId != 0)
```
anarazel: How can `curBackend` be NULL here? Given this is a pre-allocated array of the right size that ought to be impossible.

src/backend/distributed/transaction/distributed_transaction_management.c
```diff
@@ -0,0 +1,431 @@
...
+/*
+ * DistributedTransactionManagementShmemInit is the callback that is to be called on
+ * shared memory startup hook. The function sets up the necessary shared memory segment
+ * for the distributed transaction manager.
+ */
+static void
+DistributedTransactionManagementShmemInit(void)
+{
+	bool alreadyInitialized = false;
+	char *trancheName = "Distributed Transaction Management";
```
anarazel: String constants are constant. You shouldn't assign them to a non-const qualified pointer (it's disallowed in later versions of C++ too)

src/backend/distributed/transaction/distributed_transaction_management.c
```diff
@@ -0,0 +1,431 @@
...
+		LWLockInitialize(&distributedTransactionShmemData->lock,
+						 distributedTransactionShmemData->trancheId);
+#endif
+
+		/* start the distributed transaction ids from 1 */
+		pg_atomic_init_u64(&distributedTransactionShmemData->nextTransactionId, 1);
+
+		/*
+		 * We should init per backend's spinlock before any backend
+		 * starts its execution.
```
anarazel: Pedantic, but it's not "should" it's "need".

src/backend/distributed/transaction/distributed_transaction_management.c
```diff
@@ -0,0 +1,431 @@
...
+
+/*
+ * UnSetDistributedTransactionId simply acquires the mutex and resets
+ * the backend's distributed transaction data in shared memory to the
+ * initial values.
+ */
+void
+UnSetDistributedTransactionId(void)
+{
+	if (MyDistributedTransactionBackend)
```
anarazel: Do we actually need this guard?

src/backend/distributed/transaction/distributed_transaction_management.c
```diff
@@ -0,0 +1,431 @@
...
+AssignDistributedTransactionId(void)
+{
+	pg_atomic_uint64 *transactionIdSequence =
+		&distributedTransactionShmemData->nextTransactionId;
+
+	SpinLockAcquire(&MyDistributedTransactionBackend->mutex);
+
+	MyDistributedTransactionBackend->databaseId = MyDatabaseId;
+	MyDistributedTransactionBackend->transactionId.initiatorNodeIdentifier =
+		(uint64) GetLocalGroupId();
```
anarazel: You shouldn't have any function calls while holding a spinlock. Especially not one that reads from the catalogs and can error out. If there were some error (e.g. just a cancel interrupt) then the error would be thrown and the spinlock not released.  Same w/ GetCurrentTimestamp(). I'd even move the pg_atomic_fetch_add out of the spinlock protected section.

src/include/distributed/distributed_transaction_management.h
```diff
@@ -0,0 +1,52 @@
...
+ *     the distributed transaction
+ *  -  transactionId: A locally unique identifier assigned for the distributed
+ *     transaction on the node that initiated the distributed transaction
+ *  -  timestamp: The current timestamp of distributed transaction initiation
+ *
+ */
+typedef struct DistributedTransactionId
+{
+	uint64 initiatorNodeIdentifier;
+	uint64 transactionId;
```
anarazel: transactionNumber?

anarazel:
@onderkalaci See https://github.com/citusdata/citus/pull/1447/commits/edb8f4e2c52003effd59bf9594547d1cac7da164 for additional uses of per-backend data in the prototype.  Note that that commit currently only has a bool deadlocked, but we'd probably place a string there.  I also suspect we'll grow more cases where we want to override errors, not necessarily related to transaction handling. Therefore I'd rather not have "transaction" in the name of those structs / variables.

src/backend/distributed/transaction/distributed_transaction_management.c
onderkalaci: Ah, the stupid error happened after I removed the extra functions, otherwise, I'm aware of those issues :/

src/backend/distributed/transaction/distributed_transaction_management.c
onderkalaci: I added comments to both here and UDF comment. I like the name as it is. If you feel strong about changing it, let me know.

src/backend/distributed/transaction/distributed_transaction_management.c
onderkalaci: > can do the read without spinlock, nobody else can write

I couldn't follow this. Could you clarify a bit more?

onderkalaci:
> @onderkalaci See edb8f4e for additional uses of per-backend data in the prototype. Note that that commit currently only has a bool deadlocked, but we'd probably place a string there. I also suspect we'll grow more cases where we want to override errors, not necessarily related to transaction handling. Therefore I'd rather not have "transaction" in the name of those structs / variables.

I've update the PR accordingly (which almost become a new PR 😄 ). Is that better now?

src/backend/distributed/transaction/distributed_transaction_management.c
anarazel: What I'm thinking is that we don't want to allow a distributed tx to be assigned when one is already assigned. What I'm trying to say above is that you can do the "already assigned and error if so" before acquiring the spinlock, because nobody else could write to it.

src/include/distributed/backend_data.h
```diff
@@ -0,0 +1,55 @@
+/*
+ * backend_data.h
```
anarazel: I like backend_data.h, but I'd consider moving some of the functions and datastructures into transaction_identifier.h or such (`DistributedTransactionId`, `UnSetBackendData`, `GetCurrentDistributedTransctionId`)

src/include/distributed/backend_data.h
```diff
@@ -0,0 +1,55 @@
...
+	slock_t mutex;
+	DistributedTransactionId transactionId;
+} BackendData;
+
+
+extern void InitializeBackendManagement(void);
+extern void InitializeBackendData(void);
+extern void UnSetBackendData(void);
+extern DistributedTransactionId * GetCurrentDistributedTransctionId(void);
+extern void AssignBackendData(void);
```
anarazel: I think `UnSetBackendData`, and  `AssignBackendData` should still refer to transaction identifiers.

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -51,8 +56,23 @@ StartRemoteTransactionBegin(struct MultiConnection *connection)
 	 * side might have been changed, and that would cause problematic
 	 * behaviour.
 	 */
-	if (!SendRemoteCommand(connection,
-						   "BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED"))
+	appendStringInfo(beginAndSetDistributedTransactionId, "%s;",
+					 "BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED");
```
anarazel: Don't quite understand why you're using %s here at all, I'd just go for `appendStringInfoString()`.

src/backend/distributed/transaction/backend_data.c
```diff
@@ -0,0 +1,546 @@
...
+ */
+void
+AssignBackendData(void)
+{
+	pg_atomic_uint64 *transactionNumberSequence =
+		&backendManagementShmemData->nextTransactionNumber;
+
+	uint64 nextTransactionNumber = pg_atomic_fetch_add_u64(transactionNumberSequence, 1);
+	TimestampTz currentTimestamp = GetCurrentTimestamp();
+	uint64 localGroupId = (uint64) GetLocalGroupId();
```
anarazel: (absolutely inconsequential, feel entirely free to discard: personally I'd do the GetLocalGroupId() first - makes the timestamp a tiny bit more accurate).

src/include/distributed/backend_data.h
```diff
@@ -0,0 +1,55 @@
...
+ *  -  initiatorNodeIdentifier: A unique identifier of the node that initiated
+ *     the distributed transaction
+ *  -  transactionNumber: A locally unique identifier assigned for the distributed
+ *     transaction on the node that initiated the distributed transaction
+ *  -  timestamp: The current timestamp of distributed transaction initiation
+ *
+ */
+typedef struct DistributedTransactionId
+{
+	uint64 initiatorNodeIdentifier;
```
anarazel: I know that that's "my fault" but given that we otherwise define node identifiers to be 32 bit, we should probably fix the type here.

src/include/distributed/backend_data.h
```diff
@@ -0,0 +1,55 @@
...
+
+
+/*
+ * Each backend's active distributed transaction information is tracked via
+ * BackendData on the shared memory.
+ */
+typedef struct DistributedTransactionBackendData
+{
+	Oid databaseId;
+	int processId;
```
anarazel: We don't need to store that, it's already available in `ProcGlobal->allProcs[$offset]->pid`.

src/backend/distributed/transaction/distributed_transaction_management.c
onderkalaci: Yes, I'll add error check. See how I implement the check, I think it is safer in case someone adds another function to write it in the future?

src/include/distributed/backend_data.h
anarazel: No, I don't agree.  Storing duplicated information in shared memory for minor conveniences is not a good idea.   I'd just roll in `GetAllActiveBackends()` into its sole caller (where a lot of assignments are repeated anyway). In that case the access to `PGPROC` is trivial.

src/backend/distributed/transaction/backend_data.c
```diff
@@ -0,0 +1,558 @@
...
+	values[0] = ObjectIdGetDatum(MyDatabaseId);
+	values[1] = Int32GetDatum(MyProcPid);
+
+	values[2] = Int32GetDatum(distributedTransctionId->initiatorNodeIdentifier);
+	values[3] = UInt64GetDatum(distributedTransctionId->transactionNumber);
+	values[4] = TimestampTzGetDatum(distributedTransctionId->timestamp);
+
+	heapTuple = heap_form_tuple(tupleDescriptor, values, isNulls);
+
+	PG_RETURN_DATUM(HeapTupleGetDatum(heapTuple));
```
anarazel: Or just `PG_RETURN_HEAPTUPLEHEADER()`

src/backend/distributed/transaction/backend_data.c
```diff
@@ -0,0 +1,558 @@
...
+
+	tupleStore = tuplestore_begin_heap(true, false, work_mem);
+	returnSetInfo->returnMode = SFRM_Materialize;
+	returnSetInfo->setResult = tupleStore;
+	returnSetInfo->setDesc = tupleDescriptor;
+
+	MemoryContextSwitchTo(oldContext);
+
+	activeBackendList = GetAllActiveBackends();
+	activeBackendList = SortList(activeBackendList, CompareDistributedTransactionIds);
```
anarazel: FWIW, we shouldn't be using lists for things that we just sort directly afterwards. The size of the array here is perfectly bound to MaxBackends...

src/backend/distributed/transaction/backend_data.c
```diff
@@ -0,0 +1,558 @@
...
+ *
+ * Note that the function does not gurantee a global consistent result, the function
+ * only locks each active backend while copying it.
+ */
+static List *
+GetAllActiveBackends(void)
+{
+	int backendIndex = 0;
+	List *activeBackendList = NIL;
+
```
anarazel: This really should take the lwlocks protecting the list of backends...

src/backend/distributed/transaction/backend_data.c
```diff
@@ -0,0 +1,558 @@
...
+	size = add_size(size, sizeof(BackendManagementShmemData));
+	size = add_size(size, mul_size(sizeof(BackendData), MaxBackends));
+
+	return size;
+}
+
+
+/*
+ *  InitializeBackendData is called per backend and does the
+ *  required initilizations.
```
anarazel: s/initilizations/initialization/

src/include/distributed/backend_data.h
anarazel: Hm. I don't quite see why?

src/include/distributed/backend_data.h
```diff
@@ -0,0 +1,37 @@
+/*
+ * backend_data.h
```
anarazel: Add comment explaining what the file is about.

src/backend/distributed/transaction/backend_data.c
onderkalaci: Ok, so you suggest (i) inline `GetAllActiveBackends` into `get_all_active_transactions()`  (ii) protect via `lwlock`s.

Then, shall we use `tuplestore_putvalues` while the `lwlock` is held? Otherwise, the logic to fill `tupleStore` will become more complicated.

anarazel:
On 2017-07-17 09:01:07 +0000, Önder Kalacı wrote:
> Ok, so you suggest (i) inline `GetAllActiveBackends` into `get_all_active_transactions()`  (ii) protect via `lwlock`s.

Right. Probably filling an array, because you want to sort.


> Then, shall we use `tuplestore_putvalues` while the `lwlock` is held? Otherwise, the logic to fill `tupleStore` will become more complicated.

I'd be fine with that, but I think the sorting prevents that anyway?

onderkalaci:
> I'd be fine with that, but I think the sorting prevents that anyway?

Because this function seems to be useful for testing/debugging, I'm planning to remove sorting. In the regression tests, I'll add explicit `ORDER BY` to make results consistent

src/include/distributed/backend_data.h
onderkalaci: Because it deals with backend data?

src/backend/distributed/transaction/backend_data.c
onderkalaci: I'm already removing the list &  sorting

anarazel:
> Because this function seems to be useful for testing/debugging, I'm planning to remove sorting. In the regression tests, I'll add explicit ORDER BY to make results consistent

Yea, that seems a good bit better.

src/backend/distributed/transaction/backend_data.c
```diff
@@ -0,0 +1,480 @@
...
+
+
+PG_FUNCTION_INFO_V1(assign_distributed_transaction_id);
+PG_FUNCTION_INFO_V1(get_current_transaction_id);
+PG_FUNCTION_INFO_V1(get_all_active_transactions);
+
+
+/*
+ * assign_distributed_transaction_id updates the shared memory allocated for this backend
+ * and sets initiatorNodeIdentifier, transactionId, timestamp fields with the given
```
marcocitus: transactionId -> transactionNumber

src/include/distributed/backend_data.h
```diff
@@ -0,0 +1,39 @@
...
+
+#include "datatype/timestamp.h"
+#include "distributed/transaction_identifier.h"
+#include "nodes/pg_list.h"
+#include "storage/s_lock.h"
+
+
+/*
+ * Each backend's active distributed transaction information is tracked via
+ * BackendData on the shared memory.
```
anarazel: s/on the/in/

src/include/distributed/transaction_identifier.h
```diff
@@ -0,0 +1,39 @@
...
+ *    Data structure for distributed transaction id and related function
+ *    declarations.
+ *
+ * Copyright (c) 2017, Citus Data, Inc.
+ *
+ *-------------------------------------------------------------------------
+ */
+
+#ifndef TRANSACTION_IDENTIFIER_H_
+#define TRANSACTION_IDENTIFIER_H_
```
anarazel: Curious: Why are you adding the trailing `_`?

src/backend/distributed/transaction/backend_data.c
```diff
@@ -0,0 +1,480 @@
...
+	memset(isNulls, false, sizeof(isNulls));
+
+	/* first two fields do not change for this backend, so get directly */
+	values[0] = ObjectIdGetDatum(MyDatabaseId);
+	values[1] = Int32GetDatum(MyProcPid);
+
+	values[2] = Int32GetDatum(distributedTransctionId->initiatorNodeIdentifier);
+	values[3] = UInt64GetDatum(distributedTransctionId->transactionNumber);
+	values[4] = TimestampTzGetDatum(distributedTransctionId->timestamp);
+
```
anarazel: I think we should return NULL when there's no distributed transaction in progress. Showing a random date in 1999 is a bit confusing / not helpful.

src/backend/distributed/transaction/backend_data.c
```diff
@@ -0,0 +1,480 @@
...
+	PG_RETURN_VOID();
+}
+
+
+/*
+ * InitializeBackendManagement requests the necessary shared memory
+ * from Postgres and sets up the shared memory startup hook.
+ */
+void
+InitializeBackendManagement(void)
```
anarazel: Taste (you choose): I'd name these `*BackendData` rather than `*BackendManagement`.

src/backend/distributed/citus--7.0-3--7.0-4.sql
```diff
@@ -0,0 +1,25 @@
...
+     IS 'Only intended for internal use, users should not call this. The function sets the distributed transaction id';
+
+CREATE OR REPLACE FUNCTION get_current_transaction_id(OUT databaseId oid, OUT processId int, OUT initiatorNodeIdentifier int4, OUT transactionNumber int8, OUT transactionStamp timestamptz)
+     RETURNS RECORD
+     LANGUAGE C STRICT
+     AS 'MODULE_PATHNAME',$$get_current_transaction_id$$;
+ COMMENT ON FUNCTION get_current_transaction_id(OUT databaseId oid, OUT processId int, OUT initiatorNodeIdentifier int4, OUT transactionNumber int8, OUT transactionStamp timestamptz)
+     IS 'returns the current backend data including distributed transaction id';
+
+CREATE OR REPLACE FUNCTION get_all_active_transactions(OUT databaseId oid, OUT processId int, OUT initiatorNodeIdentifier int4, OUT transactionNumber int8, OUT transactionStamp timestamptz)
```
marcocitus: we use underscores for separating words in parameter names

src/backend/distributed/citus--7.0-3--7.0-4.sql
```diff
@@ -0,0 +1,25 @@
...
+	RETURNS SETOF RECORD
+	LANGUAGE C STRICT
+	AS 'MODULE_PATHNAME', $$get_all_active_transactions$$;
+ COMMENT ON FUNCTION get_all_active_transactions(OUT databaseId oid, OUT processId int, OUT initiatorNodeIdentifier int4, OUT transactionNumber int8, OUT transactionStamp timestamptz)
+     IS 'returns all the active backends data including distributed transaction ids';
```
marcocitus: "all the active backends data" doesn't seem useful to the user. Maybe "returns distributed transaction ids of active distributed transactions"

src/backend/distributed/transaction/backend_data.c
```diff
@@ -0,0 +1,480 @@
+/*-------------------------------------------------------------------------
+ *
+ * backend_data.c
+ *
+ *  Infrastructure for managing per backend data for distributed transaction
```
anarazel: I'd remove the "distributed transaction management" bit - we'll have non-transaction related stuff in there soon. Maybe "... per backend data that can efficiently accessed by all sessions" or similar?

src/backend/distributed/transaction/backend_data.c
```diff
@@ -0,0 +1,480 @@
...
+
+	/* MyBackendData should always be avaliable, just out of paranoia */
+	if (!MyBackendData)
+	{
+		ereport(ERROR, (errmsg("backend is not ready for distributed transactions")));
+	}
+
+	SpinLockAcquire(&MyBackendData->mutex);
+
+	/* if an id is already assigned, relase the lock and error */
```
marcocitus: relase -> release

src/backend/distributed/transaction/backend_data.c
```diff
@@ -0,0 +1,480 @@
...
+
+	for (backendIndex = 0; backendIndex < MaxBackends; ++backendIndex)
+	{
+		BackendData *currentBackend =
+			&backendManagementShmemData->backends[backendIndex];
+
+		SpinLockAcquire(&currentBackend->mutex);
+
+		/* we're only interested in active backends */
+		if (currentBackend->transactionId.transactionNumber == 0)
```
marcocitus: For readability and prettiness, you could consider adding `IsInDistributedTransaction` to this PR:
https://github.com/citusdata/citus/pull/1495/files#diff-330566a321cdbe3ff78b455379575c5cR743

src/include/distributed/transaction_identifier.h
onderkalaci: Eclipse is adding that automatically, I'll remove both.

src/backend/distributed/transaction/backend_data.c
onderkalaci: makes sense 😃

src/backend/distributed/citus--7.0-3--7.0-4.sql
onderkalaci: I haven't realized that. I'll update

src/backend/distributed/citus--7.0-3--7.0-4.sql
onderkalaci: done

PR #1495, opened 2017-07-17
Add functions to dump local and global wait edges

marcocitus:
This currently includes the changes in #1489.

This PR adds the `dump_local_wait_edges` and `dump_global_wait_edges` function from #1447 with some cleanup and fixes around the way lock groups are handled.


anarazel:
@marcocitus Could you rebase, now that #1489 is merged?

marcocitus:
Yep, rebased.

src/backend/distributed/transaction/lock_graph.c
```diff
@@ -0,0 +1,772 @@
...
+static void
+AddWaitEdge(WaitGraph *waitGraph, BackendData *waitingXactData,
+			PGPROC *waitingProc, BackendData *blockingXactData,
+			PGPROC *blockingProc, bool isBlockingProcessWaiting)
+{
+	WaitEdge *curEdge = AllocWaitEdge(waitGraph);
+
+	curEdge->waitingPid = waitingProc->pid;
+
+	if (IsInDistributedTransaction(waitingXactData))
```
onderkalaci: Hmm, when a transaction is prepared, we `unset` the backend data. In that case, wouldn't we see a distributed transaction as a local one in the graph? Would that cause any problems?

src/backend/distributed/transaction/lock_graph.c
```diff
@@ -0,0 +1,772 @@
...
+	{
+		LWLockRelease(LockHashPartitionLockByIndex(partitionNum));
+	}
+
+	UnlockDistributedTransactionSharedMemory();
+}
+
+
+/*
+ * AddEdgesForLockWaits adds an edge to the wait graph for every lock that
```
onderkalaci: Also, I think it is worth mentioning that the caller needs to hold the shared memory locks? (And thus do not change the function to do crazy things 😄 )

src/backend/distributed/transaction/lock_graph.c
```diff
@@ -0,0 +1,772 @@
...
+ * that originate from the given source node.
+ */
+static WaitGraph *
+BuildWaitGraphForSourceNode(int sourceNodeId)
+{
+	WaitGraph *waitGraph = NULL;
+	int curBackend = 0;
+	bool visited[MaxBackends];
+	PROCStack remaining;
+
```
marcocitus: but sourceNodeId is almost always 0 (the group ID of the coordinator).

src/backend/distributed/transaction/lock_graph.c
```diff
@@ -0,0 +1,554 @@
...
+ */
+Datum
+dump_local_wait_edges(PG_FUNCTION_ARGS)
+{
+	int32 sourceNodeId = PG_GETARG_INT32(0);
+
+	WaitGraph *waitGraph = BuildWaitGraphForSourceNode(sourceNodeId);
+	ReturnWaitGraph(waitGraph, fcinfo);
+
+	return (Datum) 0;
```
marcocitus: The `return (Datum) 0;` seems to be the common pattern for SRFs in postgres, even though `PG_RETURN_VOID()` is the same. I presume it's because `PG_RETURN_VOID()` is meant for void functions.

src/backend/distributed/transaction/lock_graph.c
marcocitus: The terminology gets a bit tricky here. Postgres source code usually refers to a "lockable object" or "lock object", and a process can hold and wait for a number of locks on the same lockable object.

src/backend/distributed/transaction/lock_graph.c
marcocitus: We don't really care about prepared transactions. Even though they can block other processes they cannot themselves be blocked on a lock so by definition they are not part of a deadlock. Of course, they can hold the whole world, which is why we should regularly call `recover_prepared_transactions`.

src/backend/distributed/transaction/lock_graph.c
```diff
@@ -0,0 +1,554 @@
...
+
+		currentBackendData = GetBackendDataForProc(currentProc);
+
+		/*
+		 * Only start searching from distributed transactions originating on the source
+		 * node. Other deadlocks may exist, but the source node can only resolve those
+		 * that involve its own transactions.
+		 */
+		if (sourceNodeId != currentBackendData->transactionId.initiatorNodeIdentifier ||
+			!IsInDistributedTransaction(currentBackendData))
```
marcocitus: Moved to backend_data.c

src/backend/distributed/transaction/lock_graph.c
```diff
@@ -0,0 +1,772 @@
...
+		}
+
+		remaining.procs[remaining.procCount++] = currentProc;
+	}
+
+	while (remaining.procCount > 0)
+	{
+		PGPROC *waitingProc = remaining.procs[--remaining.procCount];
+
+		/*
```
marcocitus: Updated the comment.

src/backend/distributed/transaction/lock_graph.c
```diff
@@ -0,0 +1,772 @@
...
+
+
+/*
+ * AddEdgesForWaitQueue adds an edge to the wait graph for processes in front
+ * of waitingProc in the wait queue that are trying to acquire a conflicting
+ * lock.
+ */
+static void
+AddEdgesForWaitQueue(WaitGraph *waitGraph, PGPROC *waitingProc, PROCStack *remaining)
+{
```
marcocitus: Waiting PROCLOCKs have 0 as their holdMask (since they don't hold anything), which means that we can't check for conflict.

There's probably a way to simplify postgres though :)

src/backend/distributed/transaction/backend_data.c
```diff
@@ -487,3 +487,57 @@ AssignDistributedTransactionId(void)
...
+
+/*
+ * UnlockDistributedTransactionSharedMemory releases the light-weight lock
+ * on the shared memory segment for distributed transaction information.
+ */
+void
+UnlockDistributedTransactionSharedMemory(void)
+{
+	LWLockRelease(&backendManagementShmemData->lock);
+}
```
anarazel: For me these wrappers are an abstraction that increases cost at no gains. The cost is small tho...

src/backend/distributed/transaction/lock_graph.c
```diff
@@ -0,0 +1,752 @@
...
+	remaining.procs = (PGPROC **) palloc(sizeof(PGPROC *) * MaxBackends);
+	remaining.procCount = 0;
+
+	LockLockData();
+
+	/*
+	 * Build lock-graph.  We do so by first finding all procs which we are
+	 * interested in (originating on our source system, and blocked).  Once
+	 * those are collected, do depth first search over all procs blocking
+	 * those.  To avoid redundantly dropping procs, keep track of which procs
```
anarazel: "dropping procs" sounds a bit weird.

src/backend/distributed/transaction/lock_graph.c
```diff
@@ -0,0 +1,752 @@
...
+
+	/* the lock for which this process is waiting */
+	LOCK *waitLock = waitingProc->waitLock;
+
+	/* determine the conflict mask for the lock level used by the process */
+	LockMethod lockMethodTable = GetLocksMethodTable(waitLock);
+	int conflictMask = lockMethodTable->conflictTab[waitingProc->waitLockMode];
+
+	/* iterate through the queue of processes holding the lock */
+	SHM_QUEUE *procLocks = &(waitLock->procLocks);
```
anarazel: the parens here are a bit weird.

src/backend/distributed/transaction/lock_graph.c
```diff
@@ -0,0 +1,752 @@
...
+	while (procLock != NULL)
+	{
+		PGPROC *blockingProc = procLock->tag.myProc;
+
+		/* a proc never blocks itself or any other lock group member */
+		if (!IsSameLockGroup(waitingProc, blockingProc))
+		{
+			BackendData *blockingBackendData = GetBackendDataForProc(blockingProc);
+
+			if ((procLock->holdMask & conflictMask) != 0)
```
anarazel: Add comment about this test - not necessarily obvious.

src/backend/distributed/transaction/lock_graph.c
anarazel: > There's probably a way to simplify postgres though :)

Yea - a lot of this code is really crummy and *old*.

src/backend/distributed/transaction/lock_graph.c
```diff
@@ -0,0 +1,752 @@
...
+	/* determine the conflict mask for the lock level used by the process */
+	LockMethod lockMethodTable = GetLocksMethodTable(waitLock);
+	int conflictMask = lockMethodTable->conflictTab[waitingProc->waitLockMode];
+
+	/* iterate through the wait queue */
+	PROC_QUEUE *waitQueue = &(waitLock->waitProcs);
+	int queueSize = waitQueue->size;
+	PGPROC *currentProc = (PGPROC *) waitQueue->links.next;
+
+	while (queueSize-- > 0 && currentProc != waitingProc)
```
anarazel: Add comment saying that once we reached ourself, we've processed all locks in front of the queue?

src/test/regress/specs/isolation_dump_local_wait_edges.spec
```diff
@@ -0,0 +1,87 @@
...
+
+step "dist13-begin"
+{
+    BEGIN;
+    SELECT assign_distributed_transaction_id(13, 1, '2017-01-01 00:00:00+0');
+}
+
+step "dist13-update"
+{
+	UPDATE local_table SET y = 3 WHERE x = 1;
```
anarazel: Differently indented...

src/test/regress/specs/isolation_dump_local_wait_edges.spec
```diff
@@ -0,0 +1,87 @@
...
+}
+
+# Distributed transaction blocked by another distributed transaction
+permutation "dist11-begin" "dist13-begin" "dist11-update" "dist13-update" "detector-dump-wait-edges" "dist11-abort" "dist13-abort"
+
+# Distributed transaction blocked by a regular transaction
+permutation "local-begin" "dist13-begin" "local-update" "dist13-update" "detector-dump-wait-edges" "local-abort" "dist13-abort"
+
+# Distributed transaction blocked by a regular transaction blocked by a distributed transaction
+permutation "dist11-begin" "local-begin" "dist13-begin" "dist11-update" "local-update" "dist13-update" "detector-dump-wait-edges" "dist11-abort" "local-abort" "dist13-abort"
```
anarazel: I was thinking for a moment that it would be good to have a bunch of test with different kind of locks (weird cases like w/ concurrently), but I can't come up with a scenario where it'd matter...

src/test/regress/specs/isolation_dump_global_wait_edges.spec
```diff
@@ -0,0 +1,85 @@
...
+        waiting_transaction_num,
+        blocking_transaction_num,
+        blocking_transaction_waiting;
+}
+
+# Distributed transaction blocked by another distributed transaction
+permutation "s1-begin" "s2-begin" "s1-update" "s2-update" "detector-dump-wait-edges" "s1-abort" "s2-abort"
+
+# Distributed transaction blocked by another distributed transaction blocked by another distributed transaction
+permutation "s1-begin" "s2-begin" "s3-begin" "s1-update" "s2-update" "s3-update" "detector-dump-wait-edges" "s1-abort" "s2-abort" "s3-abort"
```
anarazel: Would be good to have a local transaction in the mix - but probably too hard to make it work...

src/backend/distributed/transaction/lock_graph.c
onderkalaci: I think the only thing that annoys me in this PR is the need to have two separate iterations for doing two very similar tasks (i.e., `AddEdgesForWaitQueue()` - `AddEdgesForLockWaits ()`). 

In order to simplify things, I've a suggestion:

* `waitLock->procLocks` holds every backend that is associated with this lock. So iterate only on that.
    * For Lock Waits, do what we're already doing `procLock->holdMask & conflictMask != 0`
    * For Wait queue, do the following: `procLock->myProc->waitLock == waitLock` && `procLock->myProc->waitLockMode & conflictMask != 0`

Does the above work?

marcocitus: Not really, since that would also produce wait edges for proclocks that are behind `waitingProc` in the lock queue. Only `waitQueue` contains the true order of the lock queue, which may also be modified by postgres' deadlock detection.

src/test/regress/specs/isolation_dump_global_wait_edges.spec
marcocitus: yeh, I think it might make sense to start extending the isolation testing framework as suggested in https://github.com/citusdata/citus/issues/1499

marcocitus:
@onderkalaci I've cleaned up some of the redundancy between `AddEdgesForLockWaits` and `AddEdgesForWaitQueue` and added some extra comments as to why they are separate. Does this work for you?

src/backend/distributed/transaction/lock_graph.c
onderkalaci: >  AddEdgesForLockWaits adds an edge to the wait graph for every lock that waitingProc is waiting for.

I still think that this comment needs some clarification. We're only iterating through `waitingProc->waitLock-> procLocks`. So isn't it `Adds an edge to the wait graph for every process that holds the lock (that waitingProc is waiting on) in conflicting manner`

Or something with better words?

src/backend/distributed/transaction/lock_graph.c
```diff
@@ -0,0 +1,752 @@
...
+	ListCell *connectionCell = NULL;
+
+	/* open connections in parallel */
+	foreach(workerNodeCell, workerNodeList)
+	{
+		WorkerNode *workerNode = (WorkerNode *) lfirst(workerNodeCell);
+		char *nodeName = workerNode->workerName;
+		int nodePort = workerNode->workerPort;
+		MultiConnection *connection = NULL;
+		int connectionFlags = 0;
```
marcocitus: We're not trying to optimise for latency here, keeping connections from every node to every other node open seems costlier than opening new connections every few seconds.

src/test/regress/expected/isolation_dump_global_wait_edges.out
```diff
@@ -0,0 +1,85 @@
...
+    FROM
+        dump_global_wait_edges()
+    ORDER BY
+        waiting_transaction_num,
+        blocking_transaction_num,
+        blocking_transaction_waiting;
+
+waiting_transaction_numblocking_transaction_numblocking_transaction_waiting
+
+193            192            f
```
marcocitus: Yeh, I thought about coming up with some CTE that maps values consistently, but it gets pretty complicated.

src/backend/distributed/transaction/lock_graph.c
```diff
@@ -0,0 +1,752 @@
...
+		WaitEdge *curEdge = &waitGraph->edges[curEdgeNum];
+
+		memset(values, 0, sizeof(values));
+		memset(nulls, 0, sizeof(nulls));
+
+		values[0] = Int32GetDatum(curEdge->waitingPid);
+		values[1] = Int32GetDatum(curEdge->waitingNodeId);
+		if (curEdge->waitingTransactionNum != 0)
+		{
+			values[2] = Int64GetDatum(curEdge->waitingTransactionNum);
```
marcocitus: This mostly determines how the user sees the output of `dump_*_wait_edges`. I think NULL is slightly prettier because it usually shows up as empty.

src/backend/distributed/transaction/lock_graph.c
```diff
@@ -0,0 +1,769 @@
...
+
+	/* check to see if caller supports us returning a tuplestore */
+	if (rsinfo == NULL || !IsA(rsinfo, ReturnSetInfo))
+	{
+		ereport(ERROR,
+				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+				 errmsg(
+					 "set-valued function called in context that cannot accept a set")));
+	}
+	if (!(rsinfo->allowedModes & SFRM_Materialize))
```
marcocitus: Not doing this could cause crashes if the function is called from a context that doesn't support results returned using a tuple store, although I don't think such contexts exist right now. It's more a convention.

src/backend/distributed/transaction/lock_graph.c
```diff
@@ -0,0 +1,752 @@
...
+
+
+/*
+ * ParseBoolField parses a bool from a remote result or returns false if the
+ * result is NULL.
+ */
+static bool
+ParseBoolField(PGresult *result, int rowIndex, int colIndex)
+{
+	char *resultString = NULL;
```
marcocitus: Updated this to more explicitly implement boolean parsing.

src/backend/distributed/transaction/lock_graph.c
onderkalaci: Then, shall we use `FORCE_NEW_CONNECTION` to make this behavior consistent with every usage pattern? 

Because currently we may or may not open a new connection depending on the cached connections.

marcocitus: > Because currently we may or may not open a new connection depending on the cached connections.

That seems like desirable behaviour. We don't want to explicitly persist connections in the maintenance daemon, but if you're going to call this from a Citus session and there are existing connections, then why not use them?

src/backend/distributed/transaction/backend_data.c
```diff
@@ -501,3 +501,57 @@ CurrentDistributedTransactionNumber(void)
 
 	return MyBackendData->transactionId.transactionNumber;
 }
+
+
+/*
+ * LockBackendDataSharedMemory locks the shared memory segment used to keep
```
onderkalaci: Why do we want to prevent new backends?

src/backend/distributed/transaction/lock_graph.c
onderkalaci: I don't have strong opinions, just to be consistent and prevent any future changes (to connection mngmnt) from breaking our desirable behavior here.

src/test/regress/specs/isolation_dump_global_wait_edges.spec
onderkalaci: > yeh, I think it might make sense to start extending the isolation testing framework as suggested in #1499

I'm also excited to write / see new tests with it.

src/backend/distributed/transaction/backend_data.c
marcocitus: Mostly to avoid using spinlocks everywhere in `dump_local_wait_edges`. Will clarify the comment a bit.

PR #1497, opened 2017-07-19
Fix crash when removing an inactive node while there are reference tables

lithp:


lithp:
I think this is something we should backport?

src/backend/distributed/utils/reference_table_utils.c
```diff
@@ -406,25 +406,29 @@ DeleteAllReferenceTablePlacementsFromNode(char *workerName, uint32 workerPort)
...
 		Oid referenceTableId = lfirst_oid(referenceTableCell);
 		List *placements = GroupShardPlacementsForTableOnGroup(referenceTableId,
 															   workerGroup);
-		GroupShardPlacement *placement = (GroupShardPlacement *) linitial(placements);
-
-		uint64 shardId = placement->shardId;
-		uint64 placementId = placement->placementId;
+		if (list_length(placements) == 0)
+		{
+			/* this happens if the node was previously disabled */
```
marcocitus: What if we just skip `DeleteAllReferenceTablePlacementsFromNode` altogether for disabled nodes?

lithp: That wouldn't fix the situation where someone manually messed with the metadata and deleted all the placements for a reference table on a given group.

marcocitus: There are probably many ways to induce crashes by manually messing with metadata or catalog tables, but ok.

jasonmp85:
Wait, why did you remove the "backport" label? We use that to look for issues that have been backported, it's not just a status marker of what needs to be backported? Have you done this to other issues, too?

lithp:
This is the only one I can remember removing. I looked at the code and
realized this didn't actually need to be backported.

I didn't look again just now but if I remember right, while this problem
existed, we threw an error earlier in the call so this crash never occured.

jasonmp85:
Ah, ok. Comment that, then; the last comment on the thread was that it should be backported, followed by a removal of the label.

jasonmp85:
(I'll make sure it doesn't need a quick backport, and then re-remove the label if not).

PR #1500, opened 2017-07-19
Explicit switch/case fall-throughs to avoid compiler warnings.

pykello:
Currently when compiling Citus we get couple of warnings related to implicit switch/case fall-throughs. To avoid having to go to the code on every compile and making sure that the fall-through was intentional, this change marks them as intentional in the code so compiler doesn't warn about them.

### DIFF
--- a/src/backend/distributed/relay/relay_event_utility.c
+++ b/src/backend/distributed/relay/relay_event_utility.c
@@ -164,6 +164,7 @@ RelayEventExtendNames(Node *parseTree, char *schemaName, uint64 shardId)
 			 * performed on CreateStmt should be done here too, we simply *fall
 			 * through* to avoid code repetition.
 			 */
+			__attribute__((fallthrough));
 		}
 
 		case T_CreateStmt:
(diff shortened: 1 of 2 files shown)


lithp:
Which compiler are you using, this is a GCC thing? It seems to break in the version of GCC we've got running on travis

pykello:
GCC 7.1.1. I think -Wimplicit-fallthrough was added in GCC 7. See https://developers.redhat.com/blog/2017/03/10/wimplicit-fallthrough-in-gcc-7/

According to https://gcc.gnu.org/gcc-7/changes.html, to suppress that warning we could either use __attribute__(fallthrough), which didn't seem to work for earlier GCC versions, or a `/* fallthrough */` comment just before the following `case`. My second commit does the latter, which seems to fix the travis GCC error, but fails citus_indent.

marcocitus:
You probably already know this, but `make reindent` fixes indentation automatically.

pykello:
@marcocitus I didn't know that. Thanks.

PR #1501, opened 2017-07-19
Make maxTaskStringSize configurable

### LINKED ISSUE #716: Individual Task Query while using tasktracker executor cannot exceed 12kb
When the user runs a `SELECT` query and has the executor type set to `task-tracker`, the master node breaks up the query into query fragments and sends query fragments to worker nodes.  On each worker, there's a daemon process. The daemon process assigns individual query fragments to itself to execute. There is a limit on the length of these query fragments, they cannot exceed 12kb. (edited)
If the individual task query exceeds 12kb, then we get this error:

```
WARNING: could not receive query results from <worker-ip>:5432
DETAIL: Client error: task call string exceeds maximum assignable length
```

All the individual task queries are stored in shared memory and hence we put hard limit on what length the individual task query can be.

mtuncer:
taskCallStringSize determines the size of worker query string.
It was originally hard coded to a specific value. This has caused
issues at some users. Since it determines initial shared memory
allocation, we did not want to set it to an arbitrary higher number.
Instead made it configurable.

This commit introduces a new GUC variable max_task_string_size

Changes in this variable requires restart to be in effect.

Fixes #716

### DIFF
--- a/src/backend/distributed/worker/task_tracker.c
+++ b/src/backend/distributed/worker/task_tracker.c
@@ -53,6 +53,7 @@
 int TaskTrackerDelay = 200;       /* process sleep interval in millisecs */
 int MaxRunningTasksPerNode = 16;  /* max number of running tasks */
 int MaxTrackedTasksPerNode = 1024; /* max number of tracked tasks */
+int MaxTaskStringSize = 12288; /* max size of a worker task call string in bytes */
 WorkerTasksSharedStateData *WorkerTasksSharedState; /* shared memory state */
 
 static shmem_startup_hook_type prev_shmem_startup_hook = NULL;
@@ -390,7 +391,7 @@ TrackerCleanupJobSchemas(void)
 		cleanupTask->assignedAt = HIGH_PRIORITY_TASK_TIME;
 		cleanupTask->taskStatus = TASK_ASSIGNED;
 
-		strlcpy(cleanupTask->taskCallString, JOB_SCHEMA_CLEANUP, TASK_CALL_STRING_SIZE);
+		strlcpy(cleanupTask->taskCallString, JOB_SCHEMA_CLEANUP, MaxTaskStringSize);
 		strlcpy(cleanupTask->databaseName, databaseName, NAMEDATALEN);
 
 		/* zero out all other fields */
@@ -532,7 +533,7 @@ TaskTrackerShmemSize(void)
 
 	size = add_size(size, sizeof(WorkerTasksSharedStateData));
 
-	hashSize = hash_estimate_size(MaxTrackedTasksPerNode, sizeof(WorkerTask));
+	hashSize = hash_estimate_size(MaxTrackedTasksPerNode, WORKER_TASK_SIZE);
 	size = add_size(size, hashSize);
 
 	return size;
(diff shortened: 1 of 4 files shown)


src/backend/distributed/worker/task_tracker.c
```diff
@@ -662,9 +664,10 @@ SchedulableTaskList(HTAB *WorkerTasksHash)
 
 	for (queueIndex = 0; queueIndex < tasksToScheduleCount; queueIndex++)
 	{
-		WorkerTask *schedulableTask = (WorkerTask *) palloc0(sizeof(WorkerTask));
-		schedulableTask->jobId = schedulableTaskQueue[queueIndex].jobId;
-		schedulableTask->taskId = schedulableTaskQueue[queueIndex].taskId;
+		WorkerTask *schedulableTask = (WorkerTask *) palloc0(WORKER_TASK_SIZE);
+		WorkerTask *queuedTask = WORKER_TASK_AT(schedulableTaskQueue, queueIndex);
```
mtuncer: Not use if this is the best way to access array items. Open to suggestions

src/include/distributed/task_tracker.h
```diff
@@ -72,21 +75,28 @@ typedef enum
...
 	char userName[NAMEDATALEN]; /* user to use for local backend connection */
 	int32 connectionId;     /* connection id to local backend */
 	uint32 failureCount;    /* number of task failures */
+	char taskCallString[FLEXIBLE_ARRAY_MEMBER]; /* query or function call string */
 } WorkerTask;
 
+#define WORKER_TASK_SIZE (offsetof(WorkerTask, taskCallString) + \
+						  MaxWorkerTaskCallStringSize)
+#define WORKER_TASK_AT(workerTasks, index) \
+	((WorkerTask *) (((char *) (workerTasks)) + (index) * WORKER_TASK_SIZE))
```
mtuncer: Working around pointer arithmetic warning.

samay-sharma:
hey @mtuncer : The values look good to me. Could we rename the guc to something shorter?

Suggestions: max_task_string_size. We generally don't have master / worker in guc names. We document that info in docs and in the description. Also, I don't know if "TaskCall" is something which users would understand.

Also, could we add a hint to increase this config to the place where the user would previously get "task call string exceeds maximum assignable length". Also, maybe rephrase current message to "task string exceeds .... length [size]"

mtuncer:
@anarazel can you take a look

src/include/distributed/task_tracker.h
```diff
@@ -72,21 +72,28 @@ typedef enum
...
+ * Since taskCallString is dynamically sized use WORKER_TASK_SIZE instead of
+ * sizeof(WorkerTask). Use WORKER_TASK_AT to reference an item in WorkerTask array.
  */
 typedef struct WorkerTask
 {
 	uint64 jobId;      /* job id (upper 32-bits reserved); part of hash table key */
 	uint32 taskId;     /* task id; part of hash table key */
 	uint32 assignedAt; /* task assignment time in epoch seconds */
 
-	char taskCallString[TASK_CALL_STRING_SIZE]; /* query or function call string */
```
pykello: Also remove ```#define TASK_CALL_STRING_SIZE``` from task_tracker.h?

PR #1503, opened 2017-07-20
Fix drop table - create_distributed_table deadlock

### LINKED ISSUE #1502: Concurrent create_distributed_table and drop table might deadlock if tables are colocated
It is the reason behind the deadlock problem on the master branch. We have faced with it while running the parallel regression tests `multi_prepare_plsql.sql` and `multi_prepare_sql.sql`

In order to reproduce it, you need to create distributed table `colocated_table_1` while dropping another distributed table `colocated_table_2`, which is the only table colocated with `colocated_table_1`. The rease of the deadlock is : 
 
`CreateHashDistributedTable` takes(C-1) an `ExclusiveLock` on the `DistColocationRelationId()`, then `CreateColocatedShards` takes(C-2) an `AccessShareLock` on the ID of `colocated_table_2`, since it is the only table colocated with `colocated_table_1`.

`master_drop_all_shards` takes(D-1) an `AccessExlusiveLock` on the ID of `colocated_table_2` and `citus_drop_trigger` function takes(D-2) `RowExclusiveLock` on `DistColocationRelationId()` because it deletes the row from colocation table after checking there is no more table exist for this colocation group.

If these operations run in the order of C-1,D-1,C-2,D-2 then deadlock occurs.

In order to solve this issue, I have prepared a [PR](https://github.com/citusdata/citus/pull/1503) (thanks to Marco :)) which removes the delete operation within `citus_drop_trigger`. As a side effect, we may have dangling colocation groups on the colocation table now. PR also changes the type of lock on the `DistColocationRelationId()`to let concurrent `create_distributed_table` operations.

velioglu:
Fixes #1502 

Solves the deadlock while running create_distributed_table and drop table on colocated tables. Also makes concurrent create_distributed_tables possible.


src/backend/distributed/citus--7.0-4--7.0-5.sql
```diff
@@ -0,0 +1,46 @@
+/* citus--7.0-4--7.0-5.sql */
+
+SET search_path = 'pg_catalog';
+
+CREATE OR REPLACE FUNCTION pg_catalog.citus_drop_trigger()
```
marcocitus: It would be useful to add a comment on what changed.

src/backend/distributed/commands/create_distributed_table.c
```diff
@@ -626,6 +626,11 @@ CreateHashDistributedTable(Oid relationId, char *distributionColumnName,
 	 * Get an exclusive lock on the colocation system catalog. Therefore, we
 	 * can be sure that there will no modifications on the colocation table
 	 * until this transaction is committed.
```
marcocitus: The "until this transaction is committed" part of this comment is no longer accurate. 

I think the important point to get across is that we take an `ExclusiveLock` in case the colocation group does not exist yet and we want to avoid concurrent `create_distributed_table` calls creating 2 separate colocation groups with the same parameters. Then if the colocation group already exists we can relax the locking since we're not going to modify `pg_dist_colocation`, and otherwise we keep it such that concurrent `create_distributed_table` calls will block and then see the new colocation group.

src/backend/distributed/commands/create_distributed_table.c
```diff
@@ -644,6 +649,9 @@ CreateHashDistributedTable(Oid relationId, char *distributionColumnName,
 		}
 		else
 		{
+			/* to allow concurrent run with another create_distributed_table */
+			heap_close(pgDistColocation, ExclusiveLock);
+			heap_open(DistColocationRelationId(), AccessShareLock);
```
marcocitus: I don't think we need to get an `AccessShareLock` on `pg_dist_colocation` here, this would only protect us against the extension getting dropped, but there are various other protections against that.

src/backend/distributed/commands/create_distributed_table.c
```diff
@@ -644,6 +649,9 @@ CreateHashDistributedTable(Oid relationId, char *distributionColumnName,
...
 			sourceRelationId = ColocatedTableId(colocationId);
```
marcocitus: It's worth adding a comment that `ColocatedTableId` takes a `ShareLock` on `pg_dist_partition` such that no tables can be dropped.

marcocitus:
> We should be reading from a consistent snapshot of the metadata

Actually, an invalidation would cause us to use a new catalog snapshot when we do things like `LockShardListMetadata`. I'm beginning to think we need some kind of loop here, i.e. if we get blocked on the AccessShare lock on the source relation, read from pg_dist_partition again until there's either no source relation with the colocation ID or we can get the AccessShare lock without blocking.

marcocitus:
@velioglu I rebased and fixed up this PR a bit.

velioglu:
@marcocitus I rebased the branch and updated the citus subversion. Beside, tested your changes and looks right to me.

marcocitus:
Alright z merging.

furkansahin:
When I try to run the tests parallel, I hit this issue in PG 9.6 + Citus 7.1 in travis. I reproduced the same in my local with PG10 + Citus 7.1. However, this happens 1 in 5-6 regression test runs, so I couldn't catch it always. If I can catch, I will add my own logs here. The failing tests are multi_prepare_sql and multi_prepare_plsql in travis.
The output can be seen in [this link](https://travis-ci.org/citusdata/citus/jobs/304694352#L1358) as
```
--- /home/travis/build/citusdata/citus/src/test/regress/expected/multi_prepare_sql.out	2017-11-20 11:43:32.097741792 +0000
+++ /home/travis/build/citusdata/citus/src/test/regress/results/multi_prepare_sql.out	2017-11-20 11:45:52.530045882 +0000
@@ -1002,25 +1002,24 @@
 CREATE TABLE http_request (
   site_id INT,
   ingest_time TIMESTAMPTZ DEFAULT now(),
   url TEXT,
   request_country TEXT,
   ip_address TEXT,
   status_code INT,
   response_time_msec INT
 );
 SELECT create_distributed_table('http_request', 'site_id');
- create_distributed_table 
---------------------------
- 
-(1 row)
-
+ERROR:  deadlock detected
+DETAIL:  Process 15967 waits for AccessShareLock on relation 21321 of database 16384; blocked by process 16050.
+Process 16050 waits for RowExclusiveLock on relation 19689 of database 16384; blocked by process 15967.
+HINT:  See server log for query details.
 PREPARE FOO AS INSERT INTO http_request (
   site_id, ingest_time, url, request_country,
   ip_address, status_code, response_time_msec
 ) VALUES (
   1, clock_timestamp(), 'http://example.com/path', 'USA',
   inet '88.250.10.123', 200, 10
 );
 EXECUTE foo;
 EXECUTE foo;
 EXECUTE foo;
--- /home/travis/build/citusdata/citus/src/test/regress/expected/multi_prepare_plsql.out	2017-11-20 11:43:32.097741792 +0000
+++ /home/travis/build/citusdata/citus/src/test/regress/results/multi_prepare_plsql.out	2017-11-20 11:45:52.269912918 +0000
@@ -1177,20 +1177,23 @@
  key |   1
  key |   2
  key |   3
  key |   4
  key |   5
  key |   6
 (6 rows)
 
 DROP FUNCTION insert_with_max(text);
 DROP TABLE func_parameter_test;
+ERROR:  canceling the transaction since it was involved in a distributed deadlock
+CONTEXT:  SQL statement "SELECT master_drop_distributed_table_metadata(v_obj.objid, v_obj.schema_name, v_obj.object_name)"
+PL/pgSQL function citus_drop_trigger() line 23 at PERFORM
 -- test prepared DDL, mainly to verify we don't mess up the query tree
 SET citus.multi_shard_commit_protocol TO '2pc';
 CREATE TABLE prepare_ddl (x int, y int);
 SELECT create_distributed_table('prepare_ddl', 'x');
  create_distributed_table 
 --------------------------
  
 (1 row)
 
 CREATE OR REPLACE FUNCTION ddl_in_plpgsql()
```

marcocitus:
Do you have a deadlock path from the server log?

marcocitus:
It seems like the comment I posted on July 20th still applies and we kind of forgot about it.

PR #1504, opened 2017-07-20
Add transaction number to 2PC identifiers

marcocitus:
As more operations are starting to use 2PC, we've been getting reports of `pg_dist_transaction` records overflowing, since the name currently consist only of process ID and a process-local counter, and the process ID can get reused rather quickly.
https://github.com/citusdata/citus_docs/issues/417

The proper fix is to periodically call `recover_prepared_transactions` from the maintenance daemon, but even then we still want to guarantee that records don't overflow between `recover_prepared_transactions` calls.

This change ensures that `pg_dist_transaction` records cannot overlap in between restarts by including the distributed transaction number in the name.

### DIFF
--- a/src/backend/distributed/transaction/remote_transaction.c
+++ b/src/backend/distributed/transaction/remote_transaction.c
@@ -64,7 +64,7 @@ StartRemoteTransactionBegin(struct MultiConnection *connection)
 	 * and send both in one step. The reason is purely performance, we don't want
 	 * seperate roundtrips for these two statements.
 	 */
-	distributedTransactionId = GetCurrentDistributedTransctionId();
+	distributedTransactionId = GetCurrentDistributedTransactionId();
 	appendStringInfo(beginAndSetDistributedTransactionId,
 					 "SELECT assign_distributed_transaction_id(%d, %ld, '%s')",
 					 distributedTransactionId->initiatorNodeIdentifier,
--- a/src/backend/distributed/transaction/remote_transaction.c
+++ b/src/backend/distributed/transaction/remote_transaction.c
@@ -20,6 +20,7 @@
 #include "distributed/metadata_cache.h"
 #include "distributed/remote_commands.h"
 #include "distributed/remote_transaction.h"
+#include "distributed/transaction_identifier.h"
 #include "distributed/transaction_management.h"
 #include "distributed/transaction_recovery.h"
 #include "distributed/worker_manager.h"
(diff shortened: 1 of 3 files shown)


src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -856,22 +857,44 @@ CheckTransactionHealth(void)
...
  */
 static void
 Assign2PCIdentifier(MultiConnection *connection)
 {
-	static uint64 sequence = 0;
+	/* local sequence number used to distinguish different connections */
+	static uint32 connectionNumber = 0;
+
+	/* transaction identifier that is unique across processes */
+	DistributedTransactionId *distributedTransactionId =
```
onderkalaci: Hmm, `GetCurrentDistributedTransactionId()` is not a very cheap function to call (it does malloc, acquires the backend spinlock).

Although we can optimize reading the `transactionNumber` (skipping palloc) , I'm OK with the current approach given that this function is not supposed to be called too often.

(Or even we could add another `pg_atomic_uint64` to the shared memory and use it only for this purpose, which breaks the semantics you've mentioned on the function comment and complicates things a lot.)

What do you think?

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -856,22 +857,44 @@ CheckTransactionHealth(void)
...
+		GetCurrentDistributedTransactionId();
+
 	snprintf(connection->remoteTransaction.preparedName, NAMEDATALEN,
-			 "citus_%d_%d_"UINT64_FORMAT, GetLocalGroupId(),
-			 MyProcPid, sequence++);
+			 "citus_%u_%u_"UINT64_FORMAT "_%u", GetLocalGroupId(), MyProcPid,
```
onderkalaci: Pretty minor: `GetLocalGroupId ()` returns `int` but we use `%u`. Might be worth adding the reason on the function comment while explaining on the `source group` section?

src/backend/distributed/transaction/remote_transaction.c
marcocitus: I did that out of laziness / simplicity, but I think we don't really need the lock, since the value is only going to be changed by the current process. I could add a function to read the transactionNumber directly, adding a comment that it should only be used from the same session.

onderkalaci: Yes, that makes sense.

anarazel: OTOH, in comparison to actually doing a transaction prepare, a uncontended spinlock acquisition  - and that's what we're going to nearly always get here - is like 50 cycles, and a palloc in an already used memory context is like 120. In comparison to the locks and fsync triggered by actually preparing the transaction it shouldn't be noticeable.

lfittl:
@marcocitus If we end up making another 6.2 we should also backport this, we've had one customer report thusfar of this, as well as I've been running into this with the test suite in activerecord-multi-tenant.

marcocitus:
@lfittl That's going to be a bit tricky given that we don't have transaction numbers yet in 6.2.

Issue #1030, opened 2016-12-09
Transaction recovery must not run during prepare phase of other transactions


marcocitus:
We currently don't prevent `SELECT recover_prepared_transactions()` from running while other backends might be sending `PREPARE TRANSACTION..` commands. This means that `recover_prepared_transactions` might see prepared transactions on a worker without (visible) records in `pg_dist_transaction`. In that case, the transaction recovery logic will conclude that the transactions in question must have been aborted and try to roll them back, even though they might have to be committed. 

In MX this was addressed by taking a lock at the start of a distributed transaction, but we missed this change during MX integration:
https://github.com/citusdata/citus-mx/blob/masterless/main/src/backend/distributed/transaction/commit_protocol.c#L53

It may also be better to take the lock just before the first `PREPARE TRANSACTION..` command is sent.

The impact of this is currently low as transaction recovery is only enabled for metadata syncing, but will be high once we implement #1023 as it could cause data loss.

marcocitus:
This doesn't seem to be an issue anymore since `RecoverPreparedTransactions` blocks `LogTransactionRecord` and `LogTransactionRecord` is called prior to sending `PREPARE TRANSACTION ...`.

ChenHuajun:
when `SELECT recover_prepared_transactions()` was called before some workers had execute "PREPARE TRANSACTON ..." and others are not in a concurrent transaction, data loss would occurs.

I have reproduced this scene with gdb.

1. set breakpoint in StartRemoteTransactionPrepare()
2. update a reference table
3. continue when enter StartRemoteTransactionPrepare() for worker1
4. execute `SELECT recover_prepared_transactions()`

        [postgres@sndsdevsdd03 ~]$ psql testdb -p9700 -c "select recover_prepared_transactions()"
        NOTICE:  recovered a prepared transaction on localhost:9701
         recover_prepared_transactions 
        -------------------------------
                                     1
        (1 row)
5. continue when enter StartRemoteTransactionPrepare() for worker2

        testdb=# insert into ref_tb values(1,'c');
        WARNING:  prepared transaction with identifier "citus_0_8250_8" does not exist
        WARNING:  failed to commit transaction on localhost:9701
        WARNING:  failed to roll back prepared transaction 'citus_0_8250_8'
        HINT:  Run "COMMIT PREPARED 'citus_0_8250_8'" on localhost:9701
        INSERT 0 1

insert successed in worker2

[...]

PR #1505, opened 2017-07-21
Lock transaction metadata before issue PREPARE TRANSACTION

ChenHuajun:
Lock transaction metadata before issue `PREPARE TRANSACTION` to all relevant remote nodes to prevent transaction recovery(`SELECT recover_prepared_transactions()`) from getting inconsistent pending transactions list, and could cause data loss.

Please see #1030 for detail.

### DIFF
--- a/src/backend/distributed/transaction/remote_transaction.c
+++ b/src/backend/distributed/transaction/remote_transaction.c
@@ -23,6 +23,7 @@
 #include "distributed/transaction_management.h"
 #include "distributed/transaction_recovery.h"
 #include "distributed/worker_manager.h"
+#include "storage/lmgr.h"
 #include "utils/hsearch.h"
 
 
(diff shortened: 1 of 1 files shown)


marcocitus:
We already take a `RowExclusiveLock` before adding a record to `pg_dist_transaction`. However, we immediately release the lock by doing `heap_close(pgDistTransaction, RowExclusiveLock)`.
https://github.com/citusdata/citus/blob/master/src/backend/distributed/transaction/transaction_recovery.c#L108

The proper fix would be to change `RowExclusiveLock` to `NoLock` in the call to `heap_close`.

ChenHuajun:
Good idea! could you fix this bug? I can run my test again later.

marcocitus:
The fix is on this branch: https://github.com/citusdata/citus/tree/fix_transaction_recovery

ChenHuajun:
I had run my test for https://github.com/citusdata/citus/tree/fix_transaction_recovery, the result is ok.

PR #1506, opened 2017-07-21
Function evaluation descends into expression trees

marcocitus:
While working on #1487 I found that we don't descend into expressions when evaluating functions in DML commands, which means we skip evaluation for certain expressions.

### DIFF
--- a/src/backend/distributed/utils/citus_clauses.c
+++ b/src/backend/distributed/utils/citus_clauses.c
@@ -100,12 +100,10 @@ RequiresMasterEvaluation(Query *query)
 void
 ExecuteMasterEvaluableFunctions(Query *query, PlanState *planState)
 {
-	CmdType commandType = query->commandType;
 	ListCell *targetEntryCell = NULL;
 	ListCell *rteCell = NULL;
 	ListCell *cteCell = NULL;
 	Node *modifiedNode = NULL;
-	bool insertSelectQuery = InsertSelectIntoDistributedTable(query);
 
 	if (query->jointree && query->jointree->quals)
 	{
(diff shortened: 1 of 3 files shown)


lithp:
This diff also makes the test you added pass:

```
diff --git a/src/backend/distributed/utils/citus_clauses.c b/src/backend/distributed/utils/citus_clauses.c
index 2f8514d..b1254a2 100644
--- a/src/backend/distributed/utils/citus_clauses.c
+++ b/src/backend/distributed/utils/citus_clauses.c
@@ -100,12 +100,10 @@ RequiresMasterEvaluation(Query *query)
 void
 ExecuteMasterEvaluableFunctions(Query *query, PlanState *planState)
 {
-       CmdType commandType = query->commandType;
        ListCell *targetEntryCell = NULL;
        ListCell *rteCell = NULL;
        ListCell *cteCell = NULL;
        Node *modifiedNode = NULL;
-       bool insertSelectQuery = InsertSelectIntoDistributedTable(query);
 
        if (query->jointree && query->jointree->quals)
        {
@@ -123,16 +121,8 @@ ExecuteMasterEvaluableFunctions(Query *query, PlanState *planState)
                        continue;
                }
 
-               if (commandType == CMD_INSERT && !insertSelectQuery)
-               {
-                       modifiedNode = EvaluateNodeIfReferencesFunction((Node *) targetEntry->expr,
-                                                                                                                       planState);
-               }
-               else
-               {
-                       modifiedNode = PartiallyEvaluateExpression((Node *) targetEntry->expr,
-                                                                                                          planState);
-               }
+               modifiedNode = PartiallyEvaluateExpression((Node *) targetEntry->expr,
+                                                                                                  planState);
 
                targetEntry->expr = (Expr *) modifiedNode;
        }
```

src/backend/distributed/utils/citus_clauses.c
```diff
@@ -269,7 +269,18 @@ EvaluateNodeIfReferencesFunction(Node *expression, PlanState *planState)
 		}
 	}
 
-	return expression;
+	/* descend into expression tree (e.g. evaluate ARRAY[now(),now()]) */
```
lithp: This is a little weird because `PartiallyEvaluateExpresionMutator` already recurses with the help of `expression_tree_mutator`. It's also a bit strange that `ExecuteMasterEvaluableFunctions` takes care in picking out which parts of the query should be mutated but here we're indiscriminately calling `query_tree_mutator`. I commented with a diff which also makes the test pass and seems more in line with how the code already works.

marcocitus:
Your diff makes sense, applied it.

PR #1507, opened 2017-07-21
Fix PG 10 build, UNBOUNDED partitions now have different syntax

lithp:
Update code and tests to match the changes made in pg's d363d42

### DIFF
--- a/src/backend/distributed/utils/ruleutils_10.c
+++ b/src/backend/distributed/utils/ruleutils_10.c
@@ -5796,8 +5796,10 @@ get_rule_expr(Node *node, deparse_context *context,
 							Const	   *val;
 
 							appendStringInfoString(buf, sep);
-							if (datum->infinite)
-								appendStringInfoString(buf, "UNBOUNDED");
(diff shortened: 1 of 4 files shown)


onderkalaci:
Reflects the changes on the [commit](https://github.com/postgres/postgres/commit/d363d42bb9a4399a0207bd3b371c966e22e06bd3) went into postgres.

After updating the outputs of of pg 9.6 🚢 

(Also, you probably need to flush the cache for the `push` build to get the fresh master branch)

lithp:
Not sure what's happening here. It looks like the PR is tested using PG-10 source, but the branch is tested using the nightly package.  These two are out of date so one passes but the other fails. I guess we have to wait until this commit makes it into a nightly before we can merge?

lithp:
@onderkalaci okay, I restarted the build. It looks like they build packages more often than nightly! I was just a little too fast on this one, the package is updated now and the build passes. Any last objections before I merge?

lithp:
For future reference, here's the tree travis uses, these dates can tell you when the package was last updated: https://apt.postgresql.org/pub/repos/apt/dists/trusty-pgdg-testing/10/

PR #1508, opened 2017-07-21
Add Progress Tracking Infrastructure

aamederen:
This change adds a general purpose infrastructure to log and monitor
process about long running progresses. It uses
`pg_stat_get_progress_info` infrastructure, introduced with PostgreSQL
9.6 and used for tracking `VACUUM` commands.

This patch only handles the creation of a memory space in dynamic shared
memory, putting its info in `pg_stat_get_progress_info`, fetching the
progress monitors on demand and finalizing the progress tracking.

**NOTE and HELP REQUEST** Isolation Tests are WIP and not working because currently end of a command releases the allocated DSM segment. This makes it hard to test this infrastructure by itself in an automated fashion. It might be required to remove the tests altogether, if we cannot find a way to persist the allocated DSM segments at the end of a command. 

In other words:
```
postgres=# SELECT create_progress(1337, 8);                                               
WARNING:  01000: dynamic shared memory leak: segment 1443405848 still referenced
LOCATION:  PrintDSMLeakWarning, resowner.c:1268
 create_progress
-----------------

(1 row)
```

The mentioned segment is released/removed after the WARNING is shown.


marcocitus:
> NOTE and HELP REQUEST Isolation Tests are WIP and not working because currently end of a command releases the allocated DSM segment. This makes it hard to test this infrastructure by itself in an automated fashion. It might be required to remove the tests altogether, if we cannot find a way to persist the allocated DSM segments at the end of a command.

End of the command or end of the transaction? You typically have to use transaction blocks in isolation tests.

src/backend/distributed/progress/multi_progress.c
```diff
@@ -0,0 +1,287 @@
...
+
+	monitorSize = sizeof(ProgressMonitorHeader) + stepSize * stepCount;
+	dsmSegment = dsm_create(monitorSize, DSM_CREATE_NULL_IF_MAXSEGMENTS);
+
+	if (dsmSegment == NULL)
+	{
+		ereport(WARNING,
+				(errmsg(
+					 "Couldn't create a dynamic shared memory segment to keep track of "
+					 "progress of the current command.")));
```
marcocitus: `errmsg` convention is to use `could not create ...` and no dot at the end.

Maybe shorten the lines here such that `errmsg` fits on the same line as `ereport`.

src/backend/distributed/progress/multi_progress.c
```diff
@@ -0,0 +1,287 @@
...
+
+
+/*
+ * FinalizeProgressMonitor releases the dynamic memory segment of the current progress
+ * and removes the process from pg_stat_get_progress_info() output.
+ */
+void
+FinalizeProgressMonitor(void)
+{
+	dsm_segment *segment = dsm_find_mapping(currentProgressDSMHandle);
```
marcocitus: For consistency `segment`->`dsmSegment`

src/backend/distributed/test/progress_utils.c
```diff
@@ -0,0 +1,149 @@
...
+
+		for (stepIndex = 0; stepIndex < header->stepCount; stepIndex++)
+		{
+			uint64 *step = steps + stepIndex;
+
+			Datum values[2];
+			bool nulls[2];
+
+			MemSet(values, 0, sizeof(values));
+			MemSet(nulls, 0, sizeof(nulls));
```
marcocitus: we mostly use `memset` instead of `MemSet`

src/include/distributed/multi_progress.h
```diff
@@ -0,0 +1,43 @@
...
+/* Undefined on PostgreSQL <= 9.6 */
+#define DSM_HANDLE_INVALID 0
+extern Datum pg_stat_get_progress_info(PG_FUNCTION_ARGS);
+#endif
+
+typedef struct ProgressMonitorHeader
+{
+	uint64 processId;
+	int stepCount;
+} ProgressMonitorHeader;
```
marcocitus: Could we call this `ProgressMonitorData` and include the steps? or alternatively `ProgressMonitorData` could contain a `ProgressMonitorHeader` and steps.

src/backend/distributed/progress/multi_progress.c
```diff
@@ -0,0 +1,287 @@
...
+	}
+}
+
+
+/*
+ * FinalizeProgressMonitor releases the dynamic memory segment of the current progress
+ * and removes the process from pg_stat_get_progress_info() output.
+ */
+void
+FinalizeProgressMonitor(void)
```
marcocitus: To clarify which progress monitor, maybe `FinalizeMyProgressMonitor`.

src/test/regress/specs/isolation_progress_monitoring.spec
```diff
@@ -0,0 +1,74 @@
+setup
+{
+    CREATE FUNCTION create_progress(bigint, bigint)
+	RETURNS void
+	AS 'citus'
```
marcocitus: alignment is a little off in this file

src/backend/distributed/progress/multi_progress.c
aamederen: Thanks, it's been a long time :) fixed them.

src/backend/distributed/progress/multi_progress.c
```diff
@@ -0,0 +1,287 @@
...
+
+	if (dsmSegment == NULL)
+	{
+		dsmSegment = dsm_attach(currentProgressDSMHandle);
+	}
+
+	if (dsmSegment == NULL)
+	{
+		header = NULL;
+		steps = NULL;
```
aamederen: Makes sense to include steps and header data in a `ProgressMonitorData` struct

src/backend/distributed/progress/multi_progress.c
aamederen: Changed names to `FinalizeCurrentProgressMonitor` and `GetCurrentProgressMonitor`

src/backend/distributed/progress/multi_progress.c
aamederen: changed all `segment`s to `dsmSegment`

src/backend/distributed/test/progress_utils.c
aamederen: for consistency, I have changed it to `memset` but I wonder why do we use it instead of `MemSet`. Postgres code says `MemSet` is faster.

src/include/distributed/multi_progress.h
aamederen: Makes sense, made the changes.

src/test/regress/specs/isolation_progress_monitoring.spec
aamederen: Should be ok now

aamederen:
I tried to explain why I cannot do isolation tests in a common language: SQL :)

```
CREATE FUNCTION create_progress(bigint, bigint)
RETURNS void
AS 'citus'
LANGUAGE C STRICT;

CREATE FUNCTION update_progress(bigint, bigint)
RETURNS void
AS 'citus'
LANGUAGE C STRICT;

CREATE FUNCTION finish_progress()
RETURNS void
AS 'citus'
LANGUAGE C STRICT;

CREATE OR REPLACE FUNCTION show_progress(bigint)
RETURNS TABLE(step int, progress bigint)
AS 'citus'
LANGUAGE C STRICT;

CREATE OR REPLACE FUNCTION some_operation()
	RETURNS VOID AS $$
BEGIN
	PERFORM create_progress(1337, 5);
	PERFORM pg_sleep(3);
	PERFORM update_progress(2, 78);
	PERFORM pg_sleep(2);
	PERFORM update_progress(3, 37);
	PERFORM pg_sleep(2);
	PERFORM update_progress(4, 22);
	PERFORM pg_sleep(3);
	PERFORM finish_progress();
END;
$$ LANGUAGE 'plpgsql';


-- In a separate session for monitoring:
SELECT * FROM show_progress(1337) \watch 1


-- works with no problems and progress is shown on the monitoring session
SELECT some_operation();

-- Here is the problem
BEGIN;
SELECT create_progress(1337, 5); -- DSM Leak Warning
SELECT pg_sleep(3);
SELECT update_progress(2, 78); -- Errors out
ROLLBACK;
```
So basically, if `create_progress` function is called in a transaction, then it releases the DSM segment and also gives a warning about dsm leakage. But, if all these commands are wrapped in a pl/pgsql function, then that dsm segment is not released at the end of command and everything works fine.

I am not sure how can I write isolation tests with this behavior. Any ideas @marcocitus ?

marcocitus:
You could make your some_operation function block on a lock held by another session. Isolation tests continue with the next step as soon as a step gets blocked on a lock.

aamederen:
@marcocitus thanks, I have updated the PR and added proper isolation tests.

src/test/regress/specs/isolation_progress_monitoring.spec
```diff
@@ -19,6 +19,22 @@ setup
...
+		PERFORM create_progress(command_type, 2);
+		PERFORM pg_advisory_lock(lockid);
+		PERFORM update_progress(0, progress);
+		PERFORM pg_advisory_lock(lockid+1);
+		PERFORM update_progress(1, progress);
+		PERFORM pg_advisory_lock(lockid+2);
+		PERFORM finish_progress();
+		PERFORM pg_advisory_unlock(lockid);
+		PERFORM pg_advisory_unlock(lockid+1);
+		PERFORM pg_advisory_unlock(lockid+2);
```
marcocitus: These locks are making me a little dizzy :). Comments on what's going on would be helpful.

You could use `pg_advisory_xact_lock` here to avoid having to unlock.

src/backend/distributed/test/progress_utils.c
```diff
@@ -0,0 +1,147 @@
...
+
+
+Datum
+update_progress(PG_FUNCTION_ARGS)
+{
+	uint64 step = PG_GETARG_INT64(0);
+	uint64 newValue = PG_GETARG_INT64(1);
+
+	ProgressMonitorData *monitor = GetCurrentProgressMonitor();
+	uint64 *steps = (uint64 *) monitor->steps;
```
marcocitus: It's probably not necessary to check for NULL here since it's a test, but might still be useful in case someone copies this pattern.

src/backend/distributed/progress/multi_progress.c
```diff
@@ -0,0 +1,285 @@
...
+				void *monitorAddress = dsm_segment_address(dsmSegment);
+				ProgressMonitorData *monitor =
+					MonitorDataFromDSMSegmentAddress(monitorAddress);
+				*attachedDSMSegments = lappend(*attachedDSMSegments, dsmSegment);
+				monitorList = lappend(monitorList, monitor);
+			}
+			else
+			{
+				ereport(WARNING,
+						(errmsg("operation details of a command couldn't be fetched")));
```
marcocitus: not sure I understand this error message

src/backend/distributed/progress/multi_progress.c
```diff
@@ -0,0 +1,285 @@
...
+			if (dsmSegment == NULL)
+			{
+				dsmSegment = dsm_attach(dsmHandle);
+			}
+
+			if (dsmSegment != NULL)
+			{
+				void *monitorAddress = dsm_segment_address(dsmSegment);
+				ProgressMonitorData *monitor =
+					MonitorDataFromDSMSegmentAddress(monitorAddress);
```
marcocitus: The code above looks suspiciously similar to `GetCurrentProgressMonitor`, is there a way to combine them?

src/test/regress/specs/isolation_progress_monitoring.spec
aamederen: Yeah, that's a lot of locks but to be able to control each step, I added these locks. Not sure if it's possible to do with less locks. 

I added some comments about the rationale of the test scenario and also converted the ones inside the `sample_operation` function to `pg_advisory_xact_lock`s.

src/test/regress/specs/isolation_progress_monitoring.spec
```diff
@@ -27,48 +43,74 @@ teardown
...
+{
+	SELECT pg_advisory_unlock(12);
+	SELECT pg_advisory_unlock(22);
+	SELECT pg_advisory_unlock(32);
+}
+
+session "monitor"
 
-step "s3-get-progress-info"
+step "progress"
```
aamederen: Makes sense

src/backend/distributed/test/progress_utils.c
aamederen: Right, basically I was imagining the usage in shard rebalancer would be a proper reference but it shouldn't hurt to do some checks here.

src/backend/distributed/progress/multi_progress.c
aamederen: Right, I changed `MonitorDataFromDSMSegmentAddress` to reuse this logic.

src/backend/distributed/progress/multi_progress.c
aamederen: With the other changes, I think this message is obsolete so I removed it.

PR #1509, opened 2017-07-24
Add support for distributed partitioned tables

byucesoy:
Part of #1459 

With this PR, Citus starts to support distributed partitioned tables.

We support all possible ways to create distributed partitioned tables. These are;
- Distributing already created partitioning hierarchy
- CREATE TABLE ... PARTITION OF a distributed_table
- ALTER TABLE distributed_table ATTACH PARTITION non_distributed_table
- ALTER TABLE distributed_table ATTACH PARTITION distributed_table

We support DETACHing partitions from partitioned tables and propagating TRUNCATE and DDL commands to distributed partitioned tables. We also support DML queries, subqueries, JOINs, UNIONs etc.

This PR also refactors some parts of distributed table creation logic.

### DIFF
--- a/src/backend/distributed/commands/create_distributed_table.c
+++ b/src/backend/distributed/commands/create_distributed_table.c
@@ -38,6 +38,7 @@
 #include "distributed/metadata_sync.h"
 #include "distributed/multi_copy.h"
 #include "distributed/multi_logical_planner.h"
+#include "distributed/multi_partitioning_utils.h"
 #include "distributed/multi_utility.h"
 #include "distributed/pg_dist_colocation.h"
 #include "distributed/pg_dist_partition.h"
@@ -262,7 +263,6 @@ create_reference_table(PG_FUNCTION_ARGS)
 						errdetail("There are no active worker nodes.")));
 	}
 
-
 	CreateDistributedTable(relationId, distributionColumn, DISTRIBUTE_BY_NONE,
 						   colocateWithTableName, viaDeprecatedAPI);
 
@@ -277,7 +277,8 @@ create_reference_table(PG_FUNCTION_ARGS)
  * This functions contains all necessary logic to create distributed tables. It
  * perform necessary checks to ensure distributing the table is safe. If it is
  * safe to distribute the table, this function creates distributed table metadata,
- * creates shards and copies local data to shards.
+ * creates shards and copies local data to shards. This function also handles
+ * partitioned tables by distributing its partitions as well.
  *
  * viaDeprecatedAPI boolean flag is not optimal way to implement this function,
  * but it helps reducing code duplication a lot. We hope to remove that flag one
@@ -353,6 +354,21 @@ CreateDistributedTable(Oid relationId, Var *distributionColumn, char distributio
 	else if (distributionMethod == DISTRIBUTE_BY_NONE)
 	{
 		CreateReferenceTableShard(relationId);
+	}
+
+	/* if this table is partitioned table, distribute its partitions too */
+	if (PartitionedTable(relationId))
+	{
+		List *partitionList = PartitionList(relationId);
+		ListCell *partitionCell = NULL;
+
+		foreach(partitionCell, partitionList)
+		{
+			Oid partitionRelationId = lfirst_oid(partitionCell);
+			CreateDistributedTable(partitionRelationId, distributionColumn,
+								   distributionMethod, colocateWithTableName,
+								   viaDeprecatedAPI);
+		}
 	}
 
 	/* copy over data for hash distributed and reference tables */
@@ -566,6 +582,7 @@ EnsureRelationCanBeDistributed(Oid relationId, Var *distributionColumn,
 	Relation relation = NULL;
 	TupleDesc relationDesc = NULL;
 	char *relationName = NULL;
+	Oid parentRelationId = InvalidOid;
 
 	EnsureTableOwner(relationId);
 	EnsureTableNotDistributed(relationId);
@@ -623,6 +640,81 @@ EnsureRelationCanBeDistributed(Oid relationId, Var *distributionColumn,
 					 errdatatype(distributionColumn->vartype),
 					 errdetail("Partition column types must have a comparison function "
 							   "defined to use range partitioning.")));
+		}
+	}
+
+	if (PartitionTable(relationId))
+	{
+		parentRelationId = PartitionParentOid(relationId);
+	}
+
+	/* partitions cannot be distributed if their parent is not distributed */
+	if (PartitionTable(relationId) && !IsDistributedTable(parentRelationId))
+	{
+		char *relationName = get_rel_name(relationId);
+		char *parentRelationName = get_rel_name(parentRelationId);
+
+		ereport(ERROR, (errmsg("cannot distribute relation \"%s\" which is partition of "
+							   "\"%s\"", relationName, parentRelationName),
+						errdetail("Citus does not support distributing partitions "
+								  "if their parent is not distributed table."),
+						errhint("Distribute the partitioned table \"%s\" instead.",
+								parentRelationName)));
+	}
+
+	/*
+	 * These checks are mostly for partitioned tables not partitions because we prevent
+	 * distributing partitions directly in the above check. However, partitions can still
+	 * reach this point because, we call CreateDistributedTable for partitions if their
+	 * parent table is distributed.
+	 */
+	if (PartitionedTable(relationId))
+	{
+		/* we cannot distribute partitioned tables with master_create_distributed_table */
+		if (viaDeprecatedAPI)
+		{
+			ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+							errmsg("distributing partitioned tables in only supported "
+								   "with create_distributed_table UDF")));
+		}
+
+		/* distributing partitioned tables in only supported for hash-distribution */
+		if (distributionMethod != DISTRIBUTE_BY_HASH)
+		{
+			ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+							errmsg("distributing partitioned tables in only supported "
+								   "for hash-distributed tables")));
+		}
+
+		/* we currently don't support  partitioned tables for replication factor > 1 */
+		if (ShardReplicationFactor > 1)
+		{
+			ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+							errmsg("distributing partitioned tables with replication "
+								   "factor greater than 1 is not supported")));
+		}
+
+		/* we currently don't support MX tables to be distributed partitioned table */
+		if (replicationModel == REPLICATION_MODEL_STREAMING)
+		{
+			ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+							errmsg("distributing partitioned tables which uses "
+								   "streaming replication is not supported")));
+		}
+
+		/* we don't support distributing tables with multi-level partitioning */
+		if (PartitionTable(relationId))
+		{
+			char *relationName = get_rel_name(relationId);
+			Oid parentRelationId = PartitionParentOid(relationId);
+			char *parentRelationName = get_rel_name(parentRelationId);
+
+			ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+							errmsg("distributing multi-level partitioned tables "
+								   "is not supported"),
+							errdetail("Relation \"%s\" is partitioned table itself and "
+									  "it is also partition of relation \"%s\".",
+									  relationName, parentRelationName)));
 		}
 	}
 
@@ -1015,7 +1107,9 @@ RegularTable(Oid relationId)
 /*
  * CopyLocalDataIntoShards copies data from the local table, which is hidden
  * after converting it to a distributed table, into the shards of the distributed
- * table.
+ * table. For partitioned tables, this functions returns without copying the data
+ * because we call this function for both partitioned tables and its partitions.
+ * Returning early saves us from copying data to workers twice.
  *
  * This function uses CitusCopyDestReceiver to invoke the distributed COPY logic.
  * We cannot use a regular COPY here since that cannot read from a table. Instead
@@ -1056,6 +1150,17 @@ CopyLocalDataIntoShards(Oid distributedRelationId)
 
 	/* take an ExclusiveLock to block all operations except SELECT */
 	distributedRelation = heap_open(distributedRelationId, ExclusiveLock);
+
+	/*
+	 * Skip copying from partitioned tables, we will copy the data from
+	 * partition to partition's shards.
+	 */
+	if (PartitionedTable(distributedRelationId))
+	{
+		heap_close(distributedRelation, NoLock);
+
+		return;
+	}
 
 	/*
 	 * All writes have finished, make sure that we can see them by using the
--- a/src/backend/distributed/executor/multi_utility.c
+++ b/src/backend/distributed/executor/multi_utility.c
@@ -42,6 +42,7 @@
 #include "distributed/metadata_sync.h"
 #include "distributed/multi_copy.h"
 #include "distributed/multi_join_order.h"
+#include "distributed/multi_partitioning_utils.h"
 #include "distributed/multi_planner.h"
 #include "distributed/multi_router_executor.h"
 #include "distributed/multi_router_planner.h"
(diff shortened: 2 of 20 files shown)


mtuncer:
Would this still create distributed but non partitioned cstore tables ?

byucesoy:
@mtuncer yes.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -461,6 +462,81 @@ multi_ProcessUtility(PlannedStmt *pstmt,
...
+													parentMissingOk);
+
+			if (IsDistributedTable(parentRelationId))
+			{
+				bool missingOk = false;
+				Oid relationId = RangeVarGetRelid(createStatement->relation, NoLock,
+												  missingOk);
+				Var *parentDistributionColumn = DistPartitionKey(parentRelationId);
+				char parentDistributionMethod = DISTRIBUTE_BY_HASH;
+				uint32 parentColocationId = TableColocationId(parentRelationId);
```
onderkalaci: Minor: Assert `parentColocationId != InvalidOid`. Otherwise, we'd endup in a strange situation

src/backend/distributed/executor/multi_utility.c
```diff
@@ -461,6 +462,81 @@ multi_ProcessUtility(PlannedStmt *pstmt,
...
+		{
+			AlterTableCmd *alterTableCommand = (AlterTableCmd *) lfirst(commandCell);
+
+			/*
+			 * If a partition is being attached to a distributed table, we will distribute
+			 * attached table as well.
+			 */
+			if (alterTableCommand->subtype == AT_AttachPartition)
+			{
+				Oid relationId = AlterTableLookupRelation(alterTableStatement, NoLock);
```
onderkalaci: minor: `relationId ` to `partitionedRelationId` ?

src/backend/distributed/executor/multi_utility.c
```diff
@@ -461,6 +462,81 @@ multi_ProcessUtility(PlannedStmt *pstmt,
...
+				PartitionCmd *partitionCommand = (PartitionCmd *) alterTableCommand->def;
+				bool partitionMissingOk = false;
+				Oid partitionRelationId = RangeVarGetRelid(partitionCommand->name, NoLock,
+														   partitionMissingOk);
+
+				if (!IsDistributedTable(relationId) &&
```
onderkalaci: We should never be able to hit this check since we prevent it on the table creation. I think keeping this check is OK, but maybe add a single line comment saying that we don't expect to this here but keeping it out of paranoia or so ?

ozgune:
@byucesoy, @onderkalaci, @marcocitus -- I had an uninformed question on naming. We previously used the nouns / verbs "distribution" and "partitioning" interchangeably in our code base.

With our upcoming changes, the word "partition" may refer to two separate concepts: one that relates to sharding / distribution in Citus and another one that relates to partitioning in Postgres.

Do you think naming could cause confusion for people reading the code?

byucesoy:
@ozgune I was thinking about the same thing. This would be a good chance(as separate PR) to rename functions/variables/comments which uses "partition" in the meaning of "distribution".

Also it is user-facing/breaking change but what do you guys think about changing name of the pg_dist_partition as well? Personally, I am not comfortable with changing name of the catalog table but I wanted to get your opinions.

src/backend/distributed/commands/create_distributed_table.c
```diff
@@ -348,40 +472,55 @@ ConvertToDistributedTable(Oid relationId, char *distributionColumnName,
...
 	}
-#endif
 
-	/*
-	 * Distribution column returns NULL for reference tables,
-	 * but it is not used below for reference tables.
-	 */
-	distributionColumn = BuildDistributionKeyFromColumnName(relation,
-															distributionColumnName);
+	if (PartitionedTable(relationId))
```
byucesoy: MX check is added. There are no partition table check because we prevent distributing partitions directly in the check above.

src/backend/distributed/executor/multi_utility.c
byucesoy: In here (and above) I preferred using relationId for the relation which is the subject of the query (and used parentRelationId or partitionRelationId for the other table (whichever is applicable).

For this case query is ALTER TABLE partitioned_table ATTACH PARTITION partition_table. Thus I used relationId for partitioned table. Since we have partitionRelationId in this code block. It might be confusing since both variable names are very similar.

src/backend/distributed/executor/multi_utility.c
byucesoy: This check is for ALTER TABLE. We can hit this check if user first distributed the table then tries to make it partition of non-distributed table.

src/backend/distributed/commands/create_distributed_table.c
```diff
@@ -398,37 +537,205 @@ ConvertToDistributedTable(Oid relationId, char *distributionColumnName,
...
+	Var *sourceDistributionColumn = DistPartitionKey(sourceRelationId);
+	Oid sourceDistributionColumnType = InvalidOid;
+
+	/* reference tables have NULL distribution column */
+	if (sourceDistributionColumn != NULL)
+	{
+		sourceDistributionColumnType = sourceDistributionColumn->vartype;
+	}
+
+	if (sourceDistributionMethod != DISTRIBUTE_BY_HASH)
```
byucesoy: Fixed, it will be on the PR with the next batch of fixes.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -461,6 +462,91 @@ multi_ProcessUtility(PlannedStmt *pstmt,
...
 #endif
 
+#if (PG_VERSION_NUM >= 100000)
+	if (IsA(parsetree, CreateStmt))
+	{
+		CreateStmt *createStatement = (CreateStmt *) parsetree;
+
+		if (createStatement->inhRelations != NIL && createStatement->partbound != NULL)
+		{
+			RangeVar *parentRelation = linitial(createStatement->inhRelations);
```
byucesoy: For partitioned tables, yes we are sure.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -461,6 +462,91 @@ multi_ProcessUtility(PlannedStmt *pstmt,
...
+			}
+		}
+	}
+	else if (IsA(parsetree, AlterTableStmt))
+	{
+		AlterTableStmt *alterTableStatement = (AlterTableStmt *) parsetree;
+		List *commandList = alterTableStatement->cmds;
+		ListCell *commandCell = NULL;
+
+		foreach(commandCell, commandList)
```
byucesoy: In fact, we allow multiple commands at this point, because we do not deal with DDL propogating. We cannot put Assert here, because error checks for multiple commands is in the later stage of codepath.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -461,6 +462,91 @@ multi_ProcessUtility(PlannedStmt *pstmt,
...
+				PartitionCmd *partitionCommand = (PartitionCmd *) alterTableCommand->def;
+				bool partitionMissingOk = false;
+				Oid partitionRelationId = RangeVarGetRelid(partitionCommand->name, NoLock,
+														   partitionMissingOk);
+
+				/*
+				 * If user first distributes the table then tries to attach it to non
+				 * distributed table, we error out.
+				 */
+				if (!IsDistributedTable(relationId) &&
```
byucesoy: I added very detailed comments to explain all four cases and how we handle them. We cannot put an Assert for distributed/distributed case, because related checks in later stage of codepath.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -1035,6 +1121,43 @@ PlanAlterTableStmt(AlterTableStmt *alterTableStatement, const char *alterTableCo
...
+
+			/*
+			 * We only support ALTER TABLE ATTACH PARTITION, if it is only subcommand of
+			 * ALTER TABLE. It was already checked in ErrorIfUnsupportedAlterTableStmt.
+			 */
+			Assert(list_length(commandList) <= 1);
+
+			rightRelationId = RangeVarGetRelid(partitionCommand->name, NoLock, false);
+
+			/*
```
byucesoy: We already return for non-supported cases. See [here](https://github.com/citusdata/citus/pull/1509/files#diff-a8033caa333591e68f4fb7a599757296R1111) and [here](https://github.com/citusdata/citus/pull/1509/files#diff-a8033caa333591e68f4fb7a599757296R1172)

src/backend/distributed/executor/multi_utility.c
```diff
@@ -1035,6 +1121,43 @@ PlanAlterTableStmt(AlterTableStmt *alterTableStatement, const char *alterTableCo
...
+		}
+		else if (alterTableType == AT_DetachPartition)
+		{
+			PartitionCmd *partitionCommand = (PartitionCmd *) command->def;
+
+			/*
+			 * We only support ALTER TABLE DETACH PARTITION, if it is only subcommand of
+			 * ALTER TABLE. It was already checked in ErrorIfUnsupportedAlterTableStmt.
+			 */
+			Assert(list_length(commandList) <= 1);
```
byucesoy: Same as above.

src/backend/distributed/relay/relay_event_utility.c
```diff
@@ -622,6 +621,7 @@ AppendShardIdToName(char **name, uint64 shardId)
 	{
 		snprintf(extendedName, NAMEDATALEN, "%s%s", (*name), shardIdAndSeparator);
 	}
+
```
byucesoy: I think citus_indent put this new line. I am not sure why it is not added before (probably it is related with my version of citus_indent) but it kinda makes sense.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -1035,6 +1121,43 @@ PlanAlterTableStmt(AlterTableStmt *alterTableStatement, const char *alterTableCo
 				constraint->skip_validation = true;
 			}
 		}
+#if (PG_VERSION_NUM >= 100000)
+		else if (alterTableType == AT_AttachPartition)
```
byucesoy: Which part of the logic should be moved into separate function? It is very short logic, do you really wanto move it to another function?

src/test/regress/expected/multi_partitioning.out
```diff
@@ -0,0 +1,831 @@
...
+  1 | 06-09-2009
+  3 | 09-11-2009
+ 13 | 01-04-2009
+ 14 | 01-05-2009
+ 15 | 02-03-2009
+ 19 | 02-04-2009
+(6 rows)
+
+-- test master_modify_multiple_shards which fails in workers (updated value is outside of partition bounds)
+SELECT master_modify_multiple_shards('UPDATE partitioning_test_2009 SET time = time + INTERVAL ''6 month''');
```
onderkalaci: Maybe add one more test with a single statement update on the partitioned table that tries to move a row to a non-existing partition?

src/test/regress/expected/multi_partitioning.out
```diff
@@ -0,0 +1,831 @@
...
+-- test add PRIMARY KEY
+-- add PRIMARY KEY to partitioned table - this will error out
+ALTER TABLE partitioning_test ADD CONSTRAINT partitioning_primary PRIMARY KEY (id);
+ERROR:  primary key constraints are not supported on partitioned tables
+LINE 1: ALTER TABLE partitioning_test ADD CONSTRAINT partitioning_pr...
+                                          ^
+-- ADD PRIMARY KEY to partition
+ALTER TABLE partitioning_test_2009 ADD CONSTRAINT partitioning_2009_primary PRIMARY KEY (id);
+-- see PRIMARY KEY is created
+SELECT Constraint FROM table_checks WHERE relid = 'partitioning_test_2011'::regclass ORDER BY 1;
```
onderkalaci: This test fails, is the syntax wrong? `Constraint` -> `"Constraint"` ?

src/test/regress/expected/multi_partitioning.out
```diff
@@ -0,0 +1,831 @@
...
+ id         | integer
+ new_column | integer
+ time       | date
+(3 rows)
+
+-- COPY in transaction
+BEGIN;
+COPY partitioning_test FROM STDIN WITH CSV;
+-- see the data is loaded to shards
+SELECT * FROM partitioning_test WHERE id >= 22 ORDER BY 1;
```
onderkalaci: The test fails, only router selects work after `COPY`. So, `SELECT * FROM partitioning_test WHERE id = 22 ORDER BY 1;`  , `SELECT * FROM partitioning_test WHERE id = 23 ORDER BY 1;` etc should work.

src/test/regress/expected/multi_partitioning.out
```diff
@@ -0,0 +1,831 @@
...
+-- see rollback is successful
+SELECT * FROM partitioning_test WHERE id >= 22 ORDER BY 1;
+ id | time | new_column 
+----+------+------------
+(0 rows)
+
+-- DML in transaction
+BEGIN;
+-- INSERT in transaction
+INSERT INTO partitioning_test VALUES(25, '2010-02-02');
```
onderkalaci: It'd be nice to have DML tests with mixing partitions and partitioned tables. For instance:
1. `Insert into partitioned table`
1. `Insert into partition table with the same key`
2. `Insert into partition table with the same key`
3. `Copy into partition `
4 `Copy into partitioned`
5. `Some router selects to show that everything is OK`

src/test/regress/expected/multi_partitioning.out
```diff
@@ -0,0 +1,831 @@
...
+-- see DROPped partitions content is not accessible from partitioning_test;
+SELECT * FROM partitioning_test WHERE time >= '2010-01-01' AND time < '2011-01-01' ORDER BY 1;
+ id | time | new_column 
+----+------+------------
+(0 rows)
+
+-- test DROP partitioned table
+DROP TABLE partitioning_test;
+-- dropping the parent should CASCADE to the children as well - these should error out
+SELECT name, type FROM table_attrs WHERE relid = 'partitioning_test'::regclass ORDER BY 1;
```
onderkalaci: Minor, but I think instead of `table_attrs `, we should query `pg_dist_partition`. The failures look a bit confusing

src/test/regress/expected/multi_partitioning.out
```diff
@@ -0,0 +1,831 @@
...
+(3 rows)
+
+-- COPY in transaction
+BEGIN;
+COPY partitioning_test FROM STDIN WITH CSV;
+-- see the data is loaded to shards
+SELECT * FROM partitioning_test WHERE id >= 22 ORDER BY 1;
+ERROR:  cannot open new connections after the first modification command within a transaction
+ROLLBACK;
+-- see rollback is successful
```
onderkalaci: One more test suggestion, which should be a very common case:

```SQL
BEGIN.
DETACH partition
DROP partition;
COMMIT;
```

src/backend/distributed/planner/multi_planner.c
```diff
@@ -177,12 +187,43 @@ AssignRTEIdentities(Query *queryTree)
...
+		 * (i.e. we do not want to expand them to their partitions). To do this
+		 * we set each distributed partitioned table's inh flag to appropriate
+		 * value before and after dropping to the standart_planner.
+		 */
+		if (IsDistributedTable(rangeTableEntry->relid) &&
+			PartitionedTable(rangeTableEntry->relid))
+		{
+			rangeTableEntry->inh = setPartitionedTablesInherited;
+
+			if (setPartitionedTablesInherited)
```
byucesoy: Unfortunatelly, yes. Otherwise, Postgres does not call our join restriction hook where it supposed to call.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -777,6 +803,126 @@ ProcessCopyStmt(CopyStmt *copyStatement, char *completionTag, bool *commandMustR
 
 
 /*
+ * ProcessCreateStmt checks if user creates the table as a partition of a
+ * distributed table. In that case, we distribute partition as well. Since
+ * the table is a partition, CreateDistributedTable will attach it to its
+ * parent table automatically after distributing it.
+ */
+static void
+ProcessCreateStmt(CreateStmt *createStatement)
```
onderkalaci: This name seems generic. Maybe change it to `ProcessCreatePartitionoOfStmt`?

src/backend/distributed/executor/multi_utility.c
```diff
@@ -777,6 +803,126 @@ ProcessCopyStmt(CopyStmt *copyStatement, char *completionTag, bool *commandMustR
...
+ * perform both of these operations. Thus, we will not propagate ALTER TABLE
+ * ... ATTACH PARTITION command to workers.
+ *
+ * Parent is distributed, partition is distributed: Partition is already
+ * distributed, we only need to attach it to its parent in workers. Attaching
+ * operation will be performed via propagating this ALTER TABLE ... ATTACH
+ * PARTITION command to workers.
+ */
+static void
+ProcessAlterTableStmt(AlterTableStmt *alterTableStatement)
```
onderkalaci: This name is also a bit generic. Maybe `ProcessAlterTableAttachPartitionStmt` ?

Issue #1499, opened 2017-07-19
Deadlock Detection Testing Framework - Proposal Two: High Level


lithp:
An example test looks like:

```
setup
{
  CREATE TABLE isolation (a int unique);
  SELECT create_distributed_table('isolation', 'a');
}

teardown
{
  DROP TABLE isolation;
}

session "s1"

step "s1-begin"
{
  BEGIN;
}

step "s1-insert-1"
{
  INSERT INTO isolation (a) VALUES (1);
}

step "s1-insert-2"
{
  INSERT INTO isolation (a) VALUES (2);
}

session "s2"

step "s2-begin"
{
  BEGIN;
}

step "s2-insert-1"
{
  INSERT INTO isolation (a) VALUES (1);
}

step "s2-insert-2"
{
  INSERT INTO isolation (a) VALUES (2);
}

permutation "s1-begin" "s2-begin" "s1-update-1" "s2-update-2" "s2-update-2" "s1-update-1"
```

This acts just like isolation tester but, when the step blocks, instead of polling `pg_blocking_pids`, it calls a combination of `pg_blocking_pids` (to catch local locks) and `dump_global_wait_edges` and checks if the relevant session is waiting on a lock anywhere in the cluster.

Pro:
- This tests things from the user's perspective and makes it easy to create tests.

Con:
- This dosn't give you a ton of flexibility for creating complicated wait-for graphs. In particular, there's no easy way to simulate sessions which are local to a worker which take out locks.

lithp:
Maybe `run_command_on_worker` could be used to simulate commands which are local to the worker.

PR #1512, opened 2017-07-24
Make isolation tester cluster aware

lithp:
Do some black magic to let isolation tester correctly switch sessions when the current session is waiting on a lock on a worker node.

Depends on `dump_global_wait_edges()`

Solves #1499

### DIFF
--- a/src/test/regress/specs/isolation_replace_wait_function.spec
+++ b/src/test/regress/specs/isolation_replace_wait_function.spec
@@ -1,0 +1,47 @@
+# check that replace_isolation_tester_func correctly replaces the functions isolation
+# tester uses while searching for locks. If those functions aren't correctly replaced
+# this test will timeout, since isolation tester will never notice that s2 is blocked
+# by s1 on a lock it's taken out on one of the workers
+
+setup
+{
+  SELECT citus.replace_isolation_tester_func();
+  SELECT citus.refresh_isolation_tester_prepared_statement();
+
+  CREATE TABLE test_locking (a int unique);
+  SELECT create_distributed_table('test_locking', 'a');
+}
+
+teardown
+{
+  DROP TABLE test_locking;
+  SELECT citus.restore_isolation_tester_func();
+}
+
+session "s1"
+
+step "s1-insert-1"
+{
+  BEGIN;
+  INSERT INTO test_locking (a) VALUES (1);
+}
+
+step "s1-finish"
+{
+  COMMIT;
+}
+
+session "s2"
+
+step "s2-insert"
+{
+  BEGIN;
+  INSERT INTO test_locking (a) VALUES (1);
+}
+
+step "s2-finish"
+{
+  COMMIT;
+}
+
+permutation "s1-insert-1" "s2-insert" "s1-finish" "s2-finish"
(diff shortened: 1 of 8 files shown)


lithp:
Updated this to work on both PG10 and PG9.6

marcocitus:
This is awesome :)

anarazel:
> Do some black magic to let isolation tester correctly switch sessions when the current session is waiting on a lock on a worker node.


> This is awesome :)

Indeed!

src/test/regress/specs/isolation_replace_wait_function.spec
```diff
@@ -0,0 +1,69 @@
+# allow isolation tester to recognize locks on workers
+
+session "s1"
+
+step "pg-96-rename"
+{
+  -- break this into a separate step because each step can be at most 1024 chars
```
anarazel: Should fix that...

src/test/regress/specs/isolation_replace_wait_function.spec
```diff
@@ -0,0 +1,69 @@
...
+      SELECT transaction_number INTO mTransactionNum FROM get_all_active_transactions() WHERE process_id = pTarget;
+
+      -- pg doesn't think we're blocked locally, so check whether we're blocked globally
+      RETURN EXISTS (
+        SELECT 1 FROM dump_global_wait_edges() WHERE waiting_transaction_num = mTransactionNum
+      );
+    END;
+  $$ LANGUAGE plpgsql;
+}
+
```
anarazel: Wonder if it would be better to maintain the code for this in the extension, and just redirect the function here. That way we'd avoid the length limits and maintaining the expected/alternate files.

src/test/regress/specs/isolation_replace_wait_function.spec
```diff
@@ -0,0 +1,69 @@
...
+      IF pg_catalog.old_pg_isolation_test_session_is_blocked(pTarget, pSources) THEN
+        RETURN true;
+      END IF;
+
+      SELECT transaction_number INTO mTransactionNum FROM get_all_active_transactions() WHERE process_id = pTarget;
+
+      -- pg doesn't think we're blocked locally, so check whether we're blocked globally
+      RETURN EXISTS (
+        SELECT 1 FROM dump_global_wait_edges() WHERE waiting_transaction_num = mTransactionNum
+      );
```
anarazel: Have you checked whether this makes the tests noticeably slower? Wouldn't be surprised if it were. If so we could perhaps check whether the coordinator statement had already been running for a bit...

lithp: I also thought it would but it seems to increase run time of the entire suite by 0.1sec

Interestingly, listening to a youtube video in the background increases run time by 1.3 sec, nearly 12%!

This seems reasonable, all we need to do is open a connection to each of the local workers and run a quick c function, though it might get slower if we build complicated graphs.

lithp:
Okay, I've moved the functions into the Citus extension, now individual tests can decide whether to check for distributed locks.

src/test/regress/specs/isolation_replace_wait_function.spec
```diff
@@ -0,0 +1,53 @@
...
+{
+  SELECT citus.replace_isolation_tester_func();
+
+  CREATE TABLE test_locking (a int unique);
+  SELECT create_distributed_table('test_locking', 'a');
+
+  -- above we replaced the function isolation tester uses, however isolation tester has
+  -- already created a prepared statement with the old function. By changing the
+  -- search_path we force a re-parse which will pick up the new definition
+  SET search_path = 'citus';
```
marcocitus: Is there a way in which we could achieve this as part of `replace_isolation_tester_func`?

lithp: Yep, looks like there is! Although, I broke it into a separate function on the off chance you want to do the replacing outside of an isolationtester context.

marcocitus: works for me

lithp:
I'll keep rebasing this on `dump_local_wait_edges` and merge once that PR has been merged.

marcocitus:
That PR has been merged.

PR #1515, opened 2017-07-25
Remove shared memory locks for managing backends

onderkalaci:
As we realize, the locking around shared memory segment isn't 
strictly necessary. 

Per-backend spinlocks already protect each backend's data.
Those locks don't provide a global consistent result while 
reading all backends' data (i.e., after a backend's data is 
read, it could be updated).

However the above is not a problem for deadlock detection. In the
worst case, we'd end-up with edges that do not exist anymore. The 
deadlock detection should already be ready for such cases.

### DIFF
--- a/src/backend/distributed/transaction/backend_data.c
+++ b/src/backend/distributed/transaction/backend_data.c
@@ -227,9 +227,6 @@ get_all_active_transactions(PG_FUNCTION_ARGS)
 	memset(values, 0, sizeof(values));
 	memset(isNulls, false, sizeof(isNulls));
 
-	/* we're reading all the backend data, take a lock to prevent concurrent additions */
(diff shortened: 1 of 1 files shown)


onderkalaci:
@anarazel Any thoughts on this PR?

@marcocitus realized that we used the shared memory locks while a new backend starts on #1447. However, we've changed that on #1489 . So, we don't seem to need those locks anymore.

onderkalaci:
As @anarazel noted in a private chat 
```
I'd *strongly* advise against that course.
Individual backends write to the memory on every single xact, that's *bound* to become a bottleneck if you make it a single lock.
The split I originally had aimed for in the prototype was to use the lwlock for entering / leaving the status array, and the spinlock for contents of individual entries.
Could use lwlocks for individual entries too, but that'd be a bit more work.
```

I'm closing this PR on behalf of #1516

PR #1516, opened 2017-07-26
Improve locking semantics for the backend management

onderkalaci:
We use the backend shared memory lock for preventing
new backends to be part of a new distributed transaction
or an existing backend to leave a distributed transaction
while we're reading the all backends' data.

The primary goal is to provide consistent view of the
current distributed transactions while doing the
deadlock detection.

**Important Note:** I considered using the the lock on backend initialization as proposed in #1447. However, I think it makes more sense to use the locks only when a backend enters/leaves a distributed transaction instead of doing it on the backend initialization. This seems (i) more convenient to achieve the goal (i.e., consistent reads of txs) (ii) plays better with cached connections). Any comments?

### DIFF
--- a/src/backend/distributed/transaction/backend_data.c
+++ b/src/backend/distributed/transaction/backend_data.c
@@ -89,6 +89,12 @@ assign_distributed_transaction_id(PG_FUNCTION_ARGS)
 		ereport(ERROR, (errmsg("backend is not ready for distributed transactions")));
 	}
 
+	/*
+	 * Note that we don't need to lock shared memory (i.e., LockBackendSharedMemory()) here
+	 * since this function is executed after AssignDistributedTransactionId() issued on the
+	 * initiator node, which already takes the required lock to enforce the consistency.
+	 */
+
 	SpinLockAcquire(&MyBackendData->mutex);
 
 	/* if an id is already assigned, release the lock and error */
@@ -236,8 +242,8 @@ get_all_active_transactions(PG_FUNCTION_ARGS)
 	memset(values, 0, sizeof(values));
 	memset(isNulls, false, sizeof(isNulls));
 
-	/* we're reading all the backend data, take a lock to prevent concurrent additions */
-	LWLockAcquire(AddinShmemInitLock, LW_SHARED);
+	/* we're reading all distributed transactions, prevent new backends */
+	LockBackendSharedMemory(LW_SHARED);
 
 	for (backendIndex = 0; backendIndex < MaxBackends; ++backendIndex)
 	{
@@ -272,7 +278,7 @@ get_all_active_transactions(PG_FUNCTION_ARGS)
 		memset(isNulls, false, sizeof(isNulls));
 	}
 
-	LWLockRelease(AddinShmemInitLock);
+	UnlockBackendSharedMemory();
 
 	/* clean up and return the tuplestore */
 	tuplestore_donestoring(tupleStore);
@@ -400,6 +406,8 @@ InitializeBackendData(void)
 
 	Assert(MyBackendData);
 
+	LockBackendSharedMemory(LW_EXCLUSIVE);
+
 	SpinLockAcquire(&MyBackendData->mutex);
 
 	MyBackendData->databaseId = MyDatabaseId;
@@ -408,6 +416,8 @@ InitializeBackendData(void)
 	MyBackendData->transactionId.timestamp = 0;
 
 	SpinLockRelease(&MyBackendData->mutex);
+
+	UnlockBackendSharedMemory();
 }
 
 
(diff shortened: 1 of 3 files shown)


src/backend/distributed/transaction/backend_data.c
```diff
@@ -476,6 +516,8 @@ AssignDistributedTransactionId(void)
 	int localGroupId = GetLocalGroupId();
 	TimestampTz currentTimestamp = GetCurrentTimestamp();
 
+	LockBackendSharedMemory(LW_EXCLUSIVE);
```
onderkalaci: Well, this prevents two concurrent distributed transactions to start. That's bad. Shall I swap `LW_EXCLUSIVE ` and `LW_SHARED`? That probably makes more sense

anarazel:
I think this isn't right. We really really shouldn't use a single lock across all sessions for stuff that's happening constantly. You're building for contention here. I don't know why you just do the enter/leave bit?

onderkalaci:
> I think this isn't right. We really really shouldn't use a single lock across all sessions for stuff that's happening constantly. 

But, we're acquiring a share lock when entering/leaving the transaction? 

If we don't use a single lock, how are we going to block things happening?

> You're building for contention here. I don't know why you just do the enter/leave bit?

I thought I did it. Entering/leaving **transactions** seem more accurate to me as I noted

> Important Note: I considered using the the lock on backend initialization as proposed in #1447. However, I think it makes more sense to use the locks only when a backend enters/leaves a distributed transaction instead of doing it on the backend initialization. This seems (i) more convenient to achieve the goal (i.e., consistent reads of txs) (ii) plays better with cached connections).

anarazel:
> But, we're acquiring a share lock when entering/leaving the transaction?

Still a single memory location that's going to be contended.

> If we don't use a single lock, how are we going to block things happening?

The important bit is just to to prevent entire sessions from leaving / reentering.

marcocitus:
> The important bit is just to to prevent entire sessions from leaving / reentering.

Why is that important? We don't do anything particularly interesting with the shared memory segment when a backend is created other than resetting its backend data, which is the same thing we do at the end of a transaction.

marcocitus:
fwiw, I think we can live without the lwlock altogether since all backend data accesses go through spin locks and it's not strictly necessary for the backend data to remain the same during `dump_local_wait_edges`. It does make it easier to reason about, since it avoids situations in which we start searching from an backend because it's in a distributed transaction, but then it's no longer in a distributed transaction by the time we add the edge. This should be harmless, but it's a little messy.

anarazel:
> Why is that important? We don't do anything particularly interesting with the shared memory segment when a backend is created other than resetting its backend data, which is the same thing we do at the end of a transaction.

You could e.g. end up with another backend with the same pid in the same or another slot. Which'd make for a lot of confusion.    And I think it's very likely that we're going to end up having more data in the array, making a sensible locking regimen more important.

onderkalaci:
>  The important bit is just to to prevent entire sessions from leaving / reentering.

I'm confused. Isn't our goal to provide a consistent view of distributed transactions when deadlock detection kicks in? If so, what's the point in acquiring locks on backend initialization? Shouldn't we bother about distributed transaction initialization / termination?

I'd be very pleased if you explain your suggestion (i.e., which lock should we acquire at which point)? Because I really cannot understand how to achieve our goal without having a single lock.

anarazel:
On 2017-07-26 07:18:20 -0700, Önder Kalacı wrote:
> >  The important bit is just to to prevent entire sessions from leaving / reentering.
>
> I'm confused. Isn't our goal to provide a consistent view of distributed transactions when deadlock detection kicks in? If so, what's the point in acquiring locks on backend initialization? Shouldn't we bother about distributed transaction initialization / termination?

The goal is to have a view that we can make sense of. If we have a pid
reused because a backend disconnected and connected again (w/ same pid),
the we can't really make sense because we can see multiple wait edges
for the same pid.  It's fine if we see a distributed transaction vanish
during the scan, or appear to the scan, because they cannot be part of
the deadlock and we'll not see them to be as such. But if we see the
same pid in multiple places (as a local non-distributed xact), we can
get into trouble, because there's additional edges that could form a
cycle.

We should:
- acquire lwlock whenever entering/leaving the array, whenever we want
  to prevent that, and when manipulating state in the control struct.
- acquire spinlock whenever manipulating/viewing the state of an
  individual's backend entry.

onderkalaci:
> - acquire lwlock whenever entering/leaving the array, whenever we want
  to prevent that, and when manipulating state in the control struct.

* entering the array -> Init backend -> LW_SHARED not to prevent concurrent init backends
* leaving the array -> Where is that happening?
* Prevent That -> Read transactions from shmem - dump locks
*  manipulating state in the control struct -> We currently don't have such thing?

I updated the PR as above.

marcocitus:
> we can't really make sense because we can see multiple wait edges
for the same pid.

marcocitus:
> we can't really make sense because we can see multiple wait edges for the same pid.

Since we take all partition locks I don't really see how that could happen (and even if it did, it wouldn't be problematic for deadlock detection).

metdos:
@anarazel, do you have more comments/requests on this PR?

src/backend/distributed/transaction/backend_data.c
```diff
@@ -400,6 +406,8 @@ InitializeBackendData(void)
 
 	Assert(MyBackendData);
 
+	LockBackendSharedMemory(LW_SHARED);
```
anarazel: This should be exclusive, otherwise it seems to falsify the comments above `LockBackendSharedMemory()`?

onderkalaci: But if we do it exclusive, we'd prevent concurrent backend initializations.  That's why we keep here shared and acquire the exclusive one forming the graph?

PR #1517, opened 2017-07-26
Enable multi-row INSERTs

### LINKED ISSUE #455: Add support for multi-row INSERT statements
Currently, statements like the below error out in Citus:

INSERT INTO counters VALUES ('a','02-01-01',1),('b','01-01-01',2);
DETAIL:  Multi-row INSERTs to distributed tables are not supported.

Issuing them like below works:

INSERT INTO counters VALUES ('a', '02-01-01',1);
INSERT INTO counters VALUES('b',01-01-01',2);

However, it seems that certain ORMs use the syntax specified above.

# Task List

Broken down by planner, executor, and testing…

## Planning Changes (two days)

 - [x] Remove check preventing multi-row `INSERT`s (0 days)
 - [x] Add logic to group rows by shard (1 day)
 - [x] Change `TargetShardIntervalForModify` to return list of shard/query structs (0.5 days)
 - [x] Change `RouterModifyTask` to expect that as input and produce multiple tasks (0.5 days)

## Executor Changes (one day)

 - [x] Call above logic from `CitusModifyBeginScan` when deferred pruning needed (0 days)
 - [x] Use changes from #1455 to call multi-shard modification code in multi-row case (1 day)

## Testing Changes (1.5 days)

 - [x] Add use of multi-row `INSERT` to reference table, modifying xacts, and insert select test files (0.25 days)
 - [ ] Add isolation tester tests for multi-row feature (1 day)
 - [x] Test interaction with deadlock detection branch (0.25 days)

jasonmp85:
Fixes #455

This is a pretty substantial refactoring of the existing modify path within the router executor and planner. In particular, we now hunt for all VALUES range table entries in `INSERT` statements and group the rows contained therein by shard identifier. These rows are stashed away for later in "ModifyRoute" elements. During deparse, the appropriate RTE is extracted from the Query and its values list is replaced by these rows before any SQL is generated.

In this way, we can create multiple Tasks, but only one per shard, to piecemeal execute a multi-row `INSERT`. The execution of jobs containing such tasks now exclusively go through the "multi-router executor" which was previously used for e.g. `INSERT INTO ... SELECT`.

By piggybacking onto that executor, we participate in ongoing transactions, get rollback-ability, etc. In short order, the only remaining use of the "single modify" router executor will be for bare single-row INSERT statements (i.e. those not in a transaction).

This change appropriately handles deferred pruning as well as master-evaluated functions.

### DIFF
--- a/src/backend/distributed/executor/multi_executor.c
+++ b/src/backend/distributed/executor/multi_executor.c
@@ -66,7 +66,7 @@ static CustomExecMethods RouterSingleModifyCustomExecMethods = {
 	.ExplainCustomScan = CitusExplainScan
 };
 
-/* not static to enable swapping in multi-modify logic during router execution */
+/* not static to enable reference by multi-modify logic in router execution */
 CustomExecMethods RouterMultiModifyCustomExecMethods = {
 	.CustomName = "RouterMultiModifyScan",
 	.BeginCustomScan = CitusModifyBeginScan,
--- a/src/backend/distributed/executor/multi_executor.c
+++ b/src/backend/distributed/executor/multi_executor.c
@@ -66,7 +66,8 @@ static CustomExecMethods RouterSingleModifyCustomExecMethods = {
 	.ExplainCustomScan = CitusExplainScan
 };
 
-static CustomExecMethods RouterMultiModifyCustomExecMethods = {
+/* not static to enable swapping in multi-modify logic during router execution */
+CustomExecMethods RouterMultiModifyCustomExecMethods = {
 	.CustomName = "RouterMultiModifyScan",
 	.BeginCustomScan = CitusModifyBeginScan,
 	.ExecCustomScan = RouterMultiModifyExecScan,
--- a/src/backend/distributed/executor/multi_router_executor.c
+++ b/src/backend/distributed/executor/multi_router_executor.c
@@ -419,6 +419,11 @@ CitusModifyBeginScan(CustomScanState *node, EState *estate, int eflags)
 				RaiseDeferredError(planningError, ERROR);
 			}
 
+			if (list_length(taskList) > 1)
+			{
+				node->methods = &RouterMultiModifyCustomExecMethods;
+			}
+
 			workerJob->taskList = taskList;
 		}
 
@@ -428,7 +433,7 @@ CitusModifyBeginScan(CustomScanState *node, EState *estate, int eflags)
 	/* prevent concurrent placement changes */
 	AcquireMetadataLocks(taskList);
 
-	/* assign task placements */
+	/* modify tasks are always assigned using first-replica policy */
 	workerJob->taskList = FirstReplicaAssignTaskList(taskList);
 }
 
--- a/src/backend/distributed/planner/multi_router_planner.c
+++ b/src/backend/distributed/planner/multi_router_planner.c
@@ -71,6 +71,28 @@
 #include "catalog/pg_proc.h"
 #include "optimizer/planmain.h"
 
+/* intermediate value for INSERT processing */
+typedef struct InsertValues
+{
+	Expr *partitionValueExpr; /* partition value provided in INSERT row */
+	List *rowValues;          /* full values list of INSERT row, possibly NIL */
+	int64 shardId;            /* target shard for this row, possibly invalid */
+} InsertValues;
+
+
+/*
+ * A ModifyRoute encapsulates the the information needed to route modifications
+ * to the appropriate shard. For a single-shard modification, only one route
+ * is needed, but in the case of e.g. a multi-row INSERT, lists of these values
+ * will help divide the rows by their destination shards, permitting later
+ * shard-and-row-specific extension of the original SQL.
+ */
+typedef struct ModifyRoute
+{
+	int64 shardId;        /* identifier of target shard */
+	List *rowValuesLists; /* for multi-row INSERTs, list of rows to be inserted */
+} ModifyRoute;
+
 
 typedef struct WalkerState
 {
@@ -99,9 +121,6 @@ static void ErrorIfNoShardsExist(DistTableCacheEntry *cacheEntry);
 static bool CanShardPrune(Oid distributedTableId, Query *query);
 static Job * CreateJob(Query *query);
 static Task * CreateTask(TaskType taskType);
-static ShardInterval * FindShardForInsert(Query *query, DistTableCacheEntry *cacheEntry,
-										  DeferredErrorMessage **planningError);
-static Expr * ExtractInsertPartitionValue(Query *query, Var *partitionColumn);
 static Job * RouterJob(Query *originalQuery,
 					   RelationRestrictionContext *restrictionContext,
 					   DeferredErrorMessage **planningError);
@@ -110,6 +129,9 @@ static List * TargetShardIntervalsForRouter(Query *query,
 											RelationRestrictionContext *restrictionContext,
 											bool *multiShardQuery);
 static List * WorkersContainingAllShards(List *prunedShardIntervalsList);
+static List * BuildRoutesForInsert(Query *query, DeferredErrorMessage **planningError);
+static List * GroupInsertValuesByShardId(List *insertValuesList);
+static List * ExtractInsertValuesList(Query *query, Var *partitionColumn);
 static bool MultiRouterPlannableQuery(Query *query,
 									  RelationRestrictionContext *restrictionContext);
 static DeferredErrorMessage * ErrorIfQueryHasModifyingCTE(Query *queryTree);
@@ -120,6 +142,8 @@ static bool SelectsFromDistributedTable(List *rangeTableList);
 #if (PG_VERSION_NUM >= 100000)
 static List * get_all_actual_clauses(List *restrictinfo_list);
 #endif
+static int CompareInsertValuesByShardId(const void *leftElement,
+										const void *rightElement);
 
 
 /*
@@ -471,7 +495,6 @@ ModifyQuerySupported(Query *queryTree, bool multiShardQuery)
 	bool isCoordinator = IsCoordinator();
 	List *rangeTableList = NIL;
 	ListCell *rangeTableCell = NULL;
-	bool hasValuesScan = false;
 	uint32 queryTableCount = 0;
 	bool specifiesPartitionValue = false;
 	ListCell *setTargetCell = NULL;
@@ -555,7 +578,7 @@ ModifyQuerySupported(Query *queryTree, bool multiShardQuery)
 		}
 		else if (rangeTableEntry->rtekind == RTE_VALUES)
 		{
-			hasValuesScan = true;
+			/* do nothing, this type is supported */
 		}
 		else
 		{
@@ -624,24 +647,6 @@ ModifyQuerySupported(Query *queryTree, bool multiShardQuery)
 								 "Joins are not supported in distributed "
 								 "modifications.", NULL);
 		}
-	}
-
-	/* reject queries which involve multi-row inserts */
-	if (hasValuesScan)
-	{
-		/*
-		 * NB: If you remove this check you must also change the checks further in this
-		 * method and ensure that VOLATILE function calls aren't allowed in INSERT
-		 * statements. Currently they're allowed but the function call is replaced
-		 * with a constant, and if you're inserting multiple rows at once the function
-		 * should return a different value for each row.
-		 */
-		return DeferredError(ERRCODE_FEATURE_NOT_SUPPORTED,
-							 "cannot perform distributed planning for the given"
-							 " modification",
-							 "Multi-row INSERTs to distributed tables are not "
-							 "supported.",
-							 NULL);
 	}
 
 	if (commandType == CMD_INSERT || commandType == CMD_UPDATE ||
@@ -1143,7 +1148,8 @@ CanShardPrune(Oid distributedTableId, Query *query)
 {
 	uint32 rangeTableId = 1;
 	Var *partitionColumn = NULL;
-	Expr *partitionValueExpr = NULL;
+	List *insertValuesList = NIL;
+	ListCell *insertValuesCell = NULL;
 
 	if (query->commandType != CMD_INSERT)
 	{
@@ -1158,14 +1164,19 @@ CanShardPrune(Oid distributedTableId, Query *query)
 		return true;
 	}
 
-	partitionValueExpr = ExtractInsertPartitionValue(query, partitionColumn);
-	if (IsA(partitionValueExpr, Const))
+	/* get full list of partition values and ensure they are all Consts */
+	insertValuesList = ExtractInsertValuesList(query, partitionColumn);
+	foreach(insertValuesCell, insertValuesList)
 	{
-		/* can do shard pruning if the partition column is constant */
-		return true;
+		InsertValues *insertValues = (InsertValues *) lfirst(insertValuesCell);
+		if (!IsA(insertValues->partitionValueExpr, Const))
+		{
+			/* can't do shard pruning if the partition column is not constant */
+			return false;
+		}
 	}
 
-	return false;
+	return true;
 }
 
 
@@ -1198,8 +1209,9 @@ ErrorIfNoShardsExist(DistTableCacheEntry *cacheEntry)
 List *
 RouterInsertTaskList(Query *query, DeferredErrorMessage **planningError)
 {
-	ShardInterval *shardInterval = NULL;
-	Task *modifyTask = NULL;
+	List *insertTaskList = NIL;
+	List *modifyRouteList = NIL;
+	ListCell *modifyRouteCell = NULL;
 
 	Oid distributedTableId = ExtractFirstDistributedTableId(query);
 	DistTableCacheEntry *cacheEntry = DistributedTableCacheEntry(distributedTableId);
@@ -1208,26 +1220,30 @@ RouterInsertTaskList(Query *query, DeferredErrorMessage **planningError)
 
 	Assert(query->commandType == CMD_INSERT);
 
-	shardInterval = FindShardForInsert(query, cacheEntry, planningError);
-
+	modifyRouteList = BuildRoutesForInsert(query, planningError);
 	if (*planningError != NULL)
 	{
 		return NIL;
 	}
 
-	/* an INSERT always routes to exactly one shard */
-	Assert(shardInterval != NULL);
-
-	modifyTask = CreateTask(MODIFY_TASK);
-	modifyTask->anchorShardId = shardInterval->shardId;
-	modifyTask->replicationModel = cacheEntry->replicationModel;
-
-	if (query->onConflict != NULL)
+	foreach(modifyRouteCell, modifyRouteList)
 	{
-		modifyTask->upsertQuery = true;
+		ModifyRoute *modifyRoute = (ModifyRoute *) lfirst(modifyRouteCell);
+
+		Task *modifyTask = CreateTask(MODIFY_TASK);
+		modifyTask->anchorShardId = modifyRoute->shardId;
+		modifyTask->replicationModel = cacheEntry->replicationModel;
+		modifyTask->rowValuesLists = modifyRoute->rowValuesLists;
+
+		if (query->onConflict != NULL)
+		{
+			modifyTask->upsertQuery = true;
+		}
+
+		insertTaskList = lappend(insertTaskList, modifyTask);
 	}
 
-	return list_make1(modifyTask);
+	return insertTaskList;
 }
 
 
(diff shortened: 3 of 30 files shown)


metdos:
Just noting that I'm also working on modify path especially on UPDATE/DELETE, and there will be some conflicts with this PR #1483. It seems that most likely I will need to rebase :)

src/backend/distributed/executor/multi_executor.c
```diff
@@ -66,7 +66,8 @@ static CustomExecMethods RouterSingleModifyCustomExecMethods = {
 	.ExplainCustomScan = CitusExplainScan
 };
 
-static CustomExecMethods RouterMultiModifyCustomExecMethods = {
+/* not static to enable swapping in multi-modify logic during router execution */
+CustomExecMethods RouterMultiModifyCustomExecMethods = {
```
marcocitus: It seems the only distinction between `RouterSingleModifyCustomExecMethods` and `RouterMultiModifyCustomExecMethods` is in the `ExecCustomScan` and the only metric used to choose between them is the number of tasks. What if we collapse them into one `ExecMethods` with a `RouterModifyExecScan` that calls the right function based on the task count?

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -429,10 +429,19 @@ CitusModifyBeginScan(CustomScanState *node, EState *estate, int eflags)
 		PlanState *planState = &(scanState->customScanState.ss.ps);
 
 		ExecuteMasterEvaluableFunctions(jobQuery, planState);
+		ExecuteTaskListFunctions(taskList, planState);
```
marcocitus: Shouldn't it be sufficient to evaluate functions in the `jobQuery` and then break down the jobQuery into multiple tasks?

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1015,73 +1005,116 @@ TargetEntryChangesValue(TargetEntry *targetEntry, Var *column, FromExpr *joinTre
...
+		modifyTask->rowValuesLists = NIL;
+
+		if (query->onConflict != NULL)
+		{
+			RangeTblEntry *rangeTableEntry = NULL;
 
-		/* set the flag */
-		modifyTask->upsertQuery = true;
+			/* set the flag */
+			modifyTask->upsertQuery = true;
```
marcocitus: not this PR's fault, but this comment doesn't seem very helpful

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1015,73 +1005,116 @@ TargetEntryChangesValue(TargetEntry *targetEntry, Var *column, FromExpr *joinTre
...
-			Alias *alias = makeAlias(CITUS_TABLE_ALIAS, NIL);
-			rangeTableEntry->alias = alias;
+			int64 shardId = modifyRoute->shardId;
+
+			/* grab shared metadata lock to stop concurrent placement additions */
+			LockShardDistributionMetadata(shardId, ShareLock);
+
+			modifyTask->anchorShardId = shardId;
+			modifyTask->valuesRTEIndex = modifyRoute->partitionvarno;
+			modifyTask->rowValuesLists = modifyRoute->rowValuesLists;
```
marcocitus: Can we avoid storing the `partitionvarno` and `rowValuesList` in `Task` by passing it to `UpdateTaskQueryString`? After we do the deparsing in `UpdateTaskQueryString` we don't need the routes anymore. When we defer pruning to the executor we rebuild the list of routes anyway.

I guess the case in which storing them might be beneficial is when the partition column values are known, but some other columns contain function calls. In that case we can build the list of tasks, but we cannot deparse them yet. On the other hand, we're deparsing them on the line below, so we'll actually do deparsing twice, which is really expensive. It seems simpler and equally performant to just build the tasks in the executor after function evaluation in case there are function calls.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1384,35 +1551,74 @@ ExtractFirstDistributedTableId(Query *query)
...
+
 	if (targetEntry == NULL)
 	{
 		ereport(ERROR, (errcode(ERRCODE_NULL_VALUE_NOT_ALLOWED),
 						errmsg("cannot perform an INSERT without a partition column "
 							   "value")));
 	}
 
-	return targetEntry->expr;
+	/* We've got a multi-row INSERT. Extract all VALUES */
```
marcocitus: Might help elaborating a bit on how multi-row INSERTs are represented in postgres.

src/backend/distributed/utils/citus_clauses.c
```diff
@@ -141,12 +146,39 @@ ExecuteMasterEvaluableFunctions(Query *query, PlanState *planState)
...
+		if (rte->rtekind == RTE_SUBQUERY)
 		{
-			continue;
+			ExecuteMasterEvaluableFunctions(rte->subquery, planState);
+		}
+		else if (rte->rtekind == RTE_VALUES)
+		{
+			ListCell *exprListCell = NULL;
+
+			foreach(exprListCell, rte->values_lists)
```
marcocitus: Would be nice to move this into its own function.

src/backend/distributed/utils/citus_clauses.c
```diff
@@ -141,12 +146,39 @@ ExecuteMasterEvaluableFunctions(Query *query, PlanState *planState)
...
+				ListCell *exprCell = NULL;
+
+				foreach(exprCell, exprList)
+				{
+					Expr *expr = (Expr *) lfirst(exprCell);
+					Node *modifiedNode = NULL;
+
+					if (commandType == CMD_INSERT && !insertSelectQuery)
+					{
+						modifiedNode = EvaluateNodeIfReferencesFunction((Node *) expr,
```
marcocitus: Small note: we'll probably stop special casing INSERTs here (#1506). It seems to be a performance hack, which actually causes some bugs / limitations.

src/test/regress/sql/multi_modifications.sql
```diff
@@ -146,8 +146,10 @@ DELETE FROM limit_orders WHERE id = 246 AND bidder_id = (random() * 1000);
 -- (the cast to timestamp is because the timestamp_eq_timestamptz operator is stable)
 DELETE FROM limit_orders WHERE id = 246 AND placed_at = current_timestamp::timestamp;
 
--- commands with multiple rows are unsupported
-INSERT INTO limit_orders VALUES (DEFAULT), (DEFAULT);
+-- commands with multiple rows are supported
+INSERT INTO limit_orders VALUES (2037, 'GOOG', 5634, now(), 'buy', random()),
+                                (2038, 'GOOG', 5634, now(), 'buy', random()),
+                                (2039, 'GOOG', 5634, now(), 'buy', random());
```
marcocitus: would be good to add some regression tests in various types of transactions, prepared statements, and upserts.

src/include/distributed/multi_router_planner.h
```diff
@@ -21,6 +21,14 @@
 #include "nodes/parsenodes.h"
 
 
+typedef struct ModifyRoute
```
marcocitus: Could use a comment on what we use this for.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1015,73 +1005,116 @@ TargetEntryChangesValue(TargetEntry *targetEntry, Var *column, FromExpr *joinTre
...
+			/* setting an alias simplifies deparsing of UPSERTs */
+			rangeTableEntry = linitial(query->rtable);
+			if (rangeTableEntry->alias == NULL)
+			{
+				Alias *alias = makeAlias(CITUS_TABLE_ALIAS, NIL);
+				rangeTableEntry->alias = alias;
+			}
+		}
+
+		if (modifyRoute != NULL)
```
marcocitus: Having a partially populated task that we complete in the executor was a bit of the hack to get function evaluation in the partition column. It probably makes more sense to leave the route / task list empty when shard pruning is not yet possible and rebuild the task list in the executor.

e.g.
https://github.com/citusdata/citus/pull/1474/files#diff-4c4f0976727d06684376ecb22e8f5362R1123
https://github.com/citusdata/citus/pull/1474/files#diff-94b1a72e077cfc917cafbafd2a8b80ccR409

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1104,154 +1137,280 @@ TargetShardIntervalForModify(Oid distributedTableId, Query *query,
...
- * needs to be applied to multiple or no shards.
+ * BuildRouteForUpdateOrDelete finds the shard interval in which an UPDATE or
+ * DELETE command should be applied and builds a ModifyRoute object for routing
+ * the command. If the query involves multiple or no shards, the function will
+ * set planningError instead.
  */
-static ShardInterval *
-FindShardForUpdateOrDelete(Query *query, DeferredErrorMessage **planningError)
+static ModifyRoute *
+BuildRouteForUpdateOrDelete(Query *query, DeferredErrorMessage **planningError)
```
jasonmp85: Yeah it was mostly a consistency thing. I didn't want the two paths (`INSERT` vs `UPDATE`/`DELETE`) to have different return types. It only returns one route right now but it makes the rest of the code flow better. Is there anything you wanted me to do here in the code?

src/backend/distributed/planner/multi_router_planner.c
jasonmp85: Sure.

src/backend/distributed/executor/multi_executor.c
jasonmp85: Hm. I'm not sure that's correct, at least considering deadlock detection-adjacent features. I was led to believe that those changes were going to permit arbitrary multi-shard/multi-statement transactions. In that case, we want to (a) use the multi-task executor (and its transaction machinery) _even_ in cases where the modification hits only a single shard, or (b) use the single-shard path for single-shard modifications outside of a transaction block.

Does that make sense? I'm hesitant to make the suggested change until the other case is cleared up.

src/include/distributed/multi_router_planner.h
jasonmp85: Sure.

src/test/regress/sql/multi_modifications.sql
jasonmp85: Agreed, was working on them last week but haven't gotten them to a pushable state yet. Will do today (SF time).

src/backend/distributed/utils/citus_clauses.c
jasonmp85: Ack; fixed here, too.

jasonmp85:
Hey, so @marcocitus I was attempting to deal with #1474 by rebasing it on top of my changes, but some of your comments here sound as though you'd rather me rebase _this_ PR on top of #1474. Which would make more sense?

I think each independently trying to do some of the things you're describing and independently merging to `master` will be needlessly confusing/complicated.

src/backend/distributed/utils/citus_clauses.c
jasonmp85: Sure, sounds good.

src/backend/distributed/executor/multi_executor.c
jasonmp85: Leaving this as-is, at least for the time being.

src/backend/distributed/planner/multi_router_planner.c
jasonmp85: Leaving as-is.

src/backend/distributed/executor/multi_router_executor.c
jasonmp85: I guess? But isn't that, uh, planning? The lines between planning and execution are getting real blurry here. The planning phase (when possible) prunes and groups all rows by their destination shards. Those groups become tasks, whose regrouped rows we use to generate the final SQL.

What you suggest I could see, but it would drive a wedge between what we do in planning and what we do in execution (i.e. if fully plannable ahead of time, we'd create these tasks like I do now, but if not, we'd build things directly from the Query at execution).

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -429,10 +429,19 @@ CitusModifyBeginScan(CustomScanState *node, EState *estate, int eflags)
...
 
 		if (deferredPruning)
 		{
-			AssignInsertTaskShardId(jobQuery, taskList);
+			AssignInsertJobShardIds(workerJob);
```
jasonmp85: Thinking I'm going to need to either merge #1474 to `master` first or rebase _this_ on top of _that_ to satisfy these suggestions.

jasonmp85:
(Alternatively, if #1474 is quickly merged to master I can just rebase atop that rather than keep both PRs open).

src/backend/distributed/planner/multi_router_planner.c
jasonmp85: This comment is :trophy:-winning in its redundancy is what it is.

src/backend/distributed/planner/multi_router_planner.c
jasonmp85: I think we could, but what would you propose be done about `RebuildShardQueries` or whatever in the executor? Right here we have both values at hand in a loop, but there we just have access to a possibly-updated list of `Task`s. Having the tasks each know about how to find and replace the rows it needs interpolated in the original query works well in that function. Thoughts?

src/backend/distributed/executor/multi_router_executor.c
marcocitus: I think we should generate tasks when we're ready to generate the correct number of tasks, which might only be at execution time (e.g. function in the partition column). I guess there is an important distinction between queries with a function call in the partition column, and queries with a constant in the partition column, but a function call in another column. In the former case we cannot generate the correct number of tasks at planning time (use `ExecuteMasterEvaluableFunctions(jobQuery, planState)`?), while in the latter case we can (use `ExecuteTaskListFunctions(taskList, planState)`?), though I don't know if we get enough performance benefit out of building the task list upfront to justify the additional function.

In any case we shouldn't do function evaluation twice, that would give unexpected results (e.g. `nextval` getting called twice).

src/backend/distributed/executor/multi_executor.c
marcocitus: We're not doing anything about multi-shard update/delete for the moment, but I guess there could be a future in which multiple functions differ. Fine leaving this as is.

src/backend/distributed/planner/multi_router_planner.c
marcocitus: Fair point, I guess we could *always* rebuild the task list and then do deparsing on-the-fly, but for creating (and preparing) plans with a partially populated task list we need these fields.

jasonmp85:
Rebased and got this to compile, but it's a bit of mess logic-wise (there were > 100 conflicts in the router). Pushing it up here even though it definitely is not in a usable state; I just want a nice GitHub view of the changes for a sec, then I'll be doing some modifications to property apply your suggestions in re: using new functions/flow from #1474…

jasonmp85:
OK @marcocitus, I've made the fewest changes atop this as possible to use the things you added in #1474. Because of the separation of concerns throughout various functions, it made sense to at minimum keep the values lists within the `Task`. Besides, I think there is something to be said for that case where non-partition columns need master evaluation so pruning can be accomplished just not deparsing until last-minute.

I moved the `valuesRTEIndex` into `Job` as it's a single value for any given multi-row query.

If you can give this one more look and organization/style/flow feedback, I'll put the final touches on it tomorrow (and flesh out the tests as I've been promising for some time), and maybe we can get this to merge by Wednesday morning (Pacific). Thanks!

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1925,6 +1786,356 @@ IntersectPlacementList(List *lhsPlacementList, List *rhsPlacementList)
...
+	 * expression from each list within that RTE, we will extract the partition
+	 * values for each row within the multi-row INSERT.
+	 */
+	if (IsA(targetEntry->expr, Var))
+	{
+		Var *partitionVar = (Var *) targetEntry->expr;
+		RangeTblEntry *referencedRTE = NULL;
+		ListCell *valuesListCell = NULL;
+
+		*partitionvarno = partitionVar->varno;
```
marcocitus: This seems pretty cheap to compute from a Query *. Instead of storing it in the `Job` and passing it as a parameter to several functions, could we call something like `PartitionColumnTargetListIndex(query)` when needed?

jasonmp85: I'm so glad you agree because I hated adding it to all the functions. Removing.

marcocitus:
Code looks good. With the additional regression tests this is a :ship:.

One thing to note is that multi-shard INSERTs will  use the aggressive shard resource locks that we currently rely on to prevent deadlocks between multi-shard commands, which means we only allow one multi-row INSERT per shard at a time, i.e. multi-row INSERTs only run concurrently if they don't modify the same shards.

With deadlock detection, we could consider just using standard shard resource locks to ensure we don't update replicas in a different order, and allowing concurrent multi-shard commands.

The main downside of removing the locks is that some concurrency scenarios will throw errors where they currently just work (albeit slowly), and you're more likely to hit connection limits due to using connection per placement.

jasonmp85:
Hey @marcocitus, I addressed your latest feedback and fixed some issues that arose around the rebase on top of @metdos's changes. I think everything in the code itself is addressed now. Can you give the C code a final look (if you're back in the office) and let me know how it looks? I know I still haven't gotten the tests added but I'll be able to get to that tomorrow and wondered if I could get a provisional :shipit: pending tests for various cases. Between meetings, the rebase, and a breaking change I hadn't noticed around `copyObject` I didn't get to them today.

marcocitus:
Yep, consider my previous messages a provisional :shipit:

src/backend/distributed/executor/multi_executor.c
anarazel: This seems like a bad idea. What if there's other queries in-progress that refer to `RouterMultiModifyCustomExecMethods`? With cursors there could be many custom scan nodes referencing `RouterMultiModifyCustomExecMethods`.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1994,6 +2024,129 @@ IntersectPlacementList(List *lhsPlacementList, List *rhsPlacementList)
...
+ * during Task execution.
+ */
+static List *
+GroupInsertValuesByShardId(List *insertValuesList)
+{
+	ModifyRoute *route = NULL;
+	ListCell *insertValuesCell = NULL;
+	List *modifyRouteList = NIL;
+
+	insertValuesList = SortList(insertValuesList, CompareInsertValuesByShardId);
```
anarazel: Is that a good idea? `qsort`, which is what `SortList` uses, isn't guaranteed to be stable - in fact . Therefore sorting like this has a good chance of removing the ordering of individual rows, which in turn can cause problems because the ordering might have been intention for deadlock avoidance purposes.
Seems better to have a `shardId->List hashmap`, a `shardOffset` indexed array of lists, or a `bsearch()`able array (not list!) of shards (think I'd go with the last variant), that things get split over. As that makes the loop body simpler, I don't think you'll even end up with more code.  As an aside, it avoids the ridiculous practice of turning a list into an array, then back into a list ;)

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -2170,3 +2323,31 @@ get_all_actual_clauses(List *restrictinfo_list)
...
+ */
+static int
+CompareInsertValuesByShardId(const void *leftElement, const void *rightElement)
+{
+	InsertValues *leftValue = *((InsertValues **) leftElement);
+	InsertValues *rightValue = *((InsertValues **) rightElement);
+	int64 leftShardId = leftValue->shardId;
+	int64 rightShardId = rightValue->shardId;
+
+	/* we compare 64-bit integers, instead of casting their difference to int */
```
anarazel: That comment imo is cargo culting. The previous practice was buggy, so I don't see why it makes sense to enshrine it in such comments.

src/backend/distributed/executor/multi_executor.c
anarazel: Err, misunderstood what you meant with "swapping in multi-modify logic" - thought you were replacing an individual callback in it. Sorry for the noise.  Why not just say 'so it can be referenced by the multi-modify' logic?

jasonmp85: I meant it can be swapped in, i.e. the compound verb "swap in". I realized later there was a garden path parse of this phrasing, but held off modifying it until someone complained. Looks like I'll change it now!

src/backend/distributed/planner/multi_router_planner.c
marcocitus: Ordering guarantees are lost as soon as the insert spans multiple shards. For a single shard it might be useful.

src/backend/distributed/planner/multi_router_planner.c
jasonmp85: Removing the comment.

src/backend/distributed/planner/multi_router_planner.c
jasonmp85: Good/subtle catch. I didn't realize that users might intentionally craft the order of multi-row INSERTs to avoid unnecessary deadlocks. Adding a secondary sort key (list index) for now to add stability to the algorithm. Fixing in a better way tomorrow morning.

jasonmp85: I still think it's worthwhile that if two rows are sorted in a particular order in the initial statement that they maintain that order if sent to the same shard, so fixed it.

PR #1518, opened 2017-07-26
Add nodecluster role (currently inoperative) to pg_dist_node

lithp:
In the interest of making this easier review I decided to create a master ticket for follower clusters, #1528, and submit three different PRs. This one only adds the `nodecluster` column, it doesn't add any functionality.

Currently `nodecluster` is `text` as that matches `workername` and `noderack`. I think, however, that a better type would be `varchar(256)`, as that's how it's handled internally, and currently names above that length are silently clipped (I believe).

### DIFF
--- a/src/backend/distributed/citus--7.0-8--7.0-9.sql
+++ b/src/backend/distributed/citus--7.0-8--7.0-9.sql
@@ -1,0 +1,70 @@
+/* citus-7.0-8--7.0-9 */
+
+SET search_path = 'pg_catalog';
+
+ALTER TABLE pg_dist_node ADD COLUMN nodecluster name NOT NULL DEFAULT 'default';
+
+DROP FUNCTION master_add_node(text, integer, integer, noderole);
+CREATE FUNCTION master_add_node(nodename text,
+                                nodeport integer,
+                                groupid integer default 0,
+                                noderole noderole default 'primary',
+                                nodecluster name default 'default',
+                                OUT nodeid integer,
+                                OUT groupid integer,
+                                OUT nodename text,
+                                OUT nodeport integer,
+                                OUT noderack text,
+                                OUT hasmetadata boolean,
+                                OUT isactive bool,
+                                OUT noderole noderole,
+                                OUT nodecluster name)
+  RETURNS record
+  LANGUAGE C STRICT
+  AS 'MODULE_PATHNAME', $$master_add_node$$;
+COMMENT ON FUNCTION master_add_node(nodename text, nodeport integer,
+                                    groupid integer, noderole noderole, nodecluster name)
+  IS 'add node to the cluster';
+
+DROP FUNCTION master_add_inactive_node(text, integer, integer, noderole);
+CREATE FUNCTION master_add_inactive_node(nodename text,
+                                         nodeport integer,
+                                         groupid integer default 0,
+                                         noderole noderole default 'primary',
+                                         nodecluster name default 'default',
+                                         OUT nodeid integer,
+                                         OUT groupid integer,
+                                         OUT nodename text,
+                                         OUT nodeport integer,
+                                         OUT noderack text,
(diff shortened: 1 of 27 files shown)


marcocitus:
>  pg_dist_shard_placement only returns primaries. When run in the secondary cluster should it return the secondary nodes?

I think that would be a reasonable start, we can fine tune things later. This will require a bunch of testing.

src/backend/distributed/citus--7.0-8--7.0-9.sql
```diff
@@ -0,0 +1,73 @@
+/* citus-7.0-8--7.0-9 */
+
+SET search_path = 'pg_catalog';
+
+ALTER TABLE pg_dist_node ADD COLUMN nodecluster text NOT NULL DEFAULT 'default';
```
marcocitus: should we use varchar(...) here to prevent names that are too long?

or, might make sense to test what happens if you put in long names (either through INSERT or master_add_node).

src/include/distributed/worker_manager.h
```diff
@@ -44,6 +46,7 @@ typedef struct WorkerNode
 	bool hasMetadata;                   /* node gets metadata changes */
 	bool isActive;                      /* node's state */
 	Oid nodeRole;                       /* the node's role in its group */
+	char nodeCluster[WORKER_LENGTH];    /* the cluster the node is a part of */
```
marcocitus: Might make sense to use NAMEDATALEN here, just in case we ever get crazy and want to create some table, schema, or role with the same name.

src/test/regress/expected/multi_metadata_sync.out
```diff
@@ -1129,9 +1129,9 @@ SELECT create_distributed_table('mx_table', 'a');
 
 \c - postgres - :master_port
 SELECT master_add_node('localhost', :worker_2_port);
```
marcocitus: Should this be SELECT 1 as well?

lithp:
@marcocitus When I switched it to `NAMEDATALEN` the test for an extra-long cluster name exposed a crash. That was scary enough that I changed the column to a varchar. Probably workername and workerrack should eventually also be varchar, or at least have some checks that the received text is shorter than `WORKER_NAME`.

This was a big enough change that I don't want to merge it until you've had a look, even though you've approved it.

src/test/regress/expected/multi_cluster_management.out
```diff
@@ -470,5 +474,36 @@ UPDATE pg_dist_node SET noderole = 'primary'
...
+
+-- check that super-long cluster names are not allowed
+SELECT master_add_node('localhost', 8887, groupid => :worker_1_group, noderole => 'secondary', nodecluster=>
+	'thisisasixtyfourcharacterstringrepeatedfourtimestomake256chars.'
+	'thisisasixtyfourcharacterstringrepeatedfourtimestomake256chars.'
+	'thisisasixtyfourcharacterstringrepeatedfourtimestomake256chars.'
+	'thisisasixtyfourcharacterstringrepeatedfourtimestomake256chars.'
+	'overflow'
+);
+ERROR:  value too long for type character varying(60)
```
marcocitus: curious error

src/backend/distributed/citus--7.0-8--7.0-9.sql
```diff
@@ -0,0 +1,73 @@
+/* citus-7.0-8--7.0-9 */
+
+SET search_path = 'pg_catalog';
+
+ALTER TABLE pg_dist_node ADD COLUMN nodecluster varchar(64) NOT NULL DEFAULT 'default';
```
marcocitus: Actually, probably better to use the `name` type here, which ensures that the characters will fit in an array of NAMEDATALEN.

lithp:
Changed the type to `name`. Awaiting a final review and I'll squash the finally get this in!

PR #1519, opened 2017-07-26
Avoid connections that accessed non-colocated placements in multi-shard commands

marcocitus:
When we execute a multi-shard command we call `GetPlacementListConnection` for every shard placement. However, if a subset of placements was modified earlier in the transaction and then `GetPlacementListConnection` is called for a placement that hasn't been modified yet, then it may return a connection that already modified a placement. Since we strictly need a connection per placement for multi-shard commands, the command would then error out.

This change fixes the issue by checking whether the connection was already used for a non-co-located placement in the same table or colocation group, and opening a new connection if that was the case.

### DIFF
--- a/src/backend/distributed/commands/multi_copy.c
+++ b/src/backend/distributed/commands/multi_copy.c
@@ -763,7 +763,7 @@ OpenCopyConnections(CopyStmt *copyStatement, ShardConnections *shardConnections,
 		ShardPlacement *placement = (ShardPlacement *) lfirst(placementCell);
 		char *nodeUser = CurrentUserName();
 		MultiConnection *connection = NULL;
-		uint32 connectionFlags = FOR_DML;
+		uint32 connectionFlags = FOR_DML | CONNECTION_PER_PLACEMENT;
 		StringInfo copyCommand = NULL;
 		PGresult *result = NULL;
 
--- a/src/backend/distributed/connection/placement_connection.c
+++ b/src/backend/distributed/connection/placement_connection.c
@@ -12,6 +12,7 @@
 #include "postgres.h"
 
 #include "access/hash.h"
+#include "distributed/colocation_utils.h"
 #include "distributed/connection_management.h"
 #include "distributed/hash_helpers.h"
 #include "distributed/master_protocol.h"
@@ -48,6 +49,10 @@ typedef struct ConnectionReference
 	 */
 	bool hadDML;
 	bool hadDDL;
+
+	/* colocation group of the placement, if any */
+	uint32 colocationGroupId;
+	uint32 representativeValue;
 
 	/* membership in MultiConnection->referencedPlacements */
 	dlist_node connectionNode;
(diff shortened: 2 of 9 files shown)


src/backend/distributed/connection/placement_connection.c
```diff
@@ -625,11 +658,21 @@ FindOrCreatePlacementEntry(ShardPlacement *placement)
 				void *conRef = MemoryContextAllocZero(TopTransactionContext,
 													  sizeof(ConnectionReference));
 
+				ConnectionReference *connectionReference = (ConnectionReference *) conRef;
+
+				/*
+				 * Store the co-location group information such that we can later
+				 * determine whether a connection accessed non-co-located placements
+				 * of the same co-location group.
```
mtuncer: "non-co-located placements of the same co-location group. " ?

src/backend/distributed/commands/multi_copy.c
```diff
@@ -763,7 +763,7 @@ OpenCopyConnections(CopyStmt *copyStatement, ShardConnections *shardConnections,
 		ShardPlacement *placement = (ShardPlacement *) lfirst(placementCell);
 		char *nodeUser = CurrentUserName();
 		MultiConnection *connection = NULL;
-		uint32 connectionFlags = FOR_DML;
+		uint32 connectionFlags = FOR_DML | CONNECTION_PER_PLACEMENT;
```
mtuncer: all select queries could be too much. but we should be fine with "copy to" type queries that also use select * on a single relation

src/backend/distributed/connection/placement_connection.c
```diff
@@ -483,13 +490,35 @@ StartPlacementListConnection(uint32 flags, List *placementAccessList,
...
+		char *nodeName = placement->nodeName;
+		int nodePort = placement->nodePort;
 
 		/*
 		 * No suitable connection in the placement->connection mapping, get one from
 		 * the node->connection pool.
 		 */
-		chosenConnection = StartNodeConnection(flags, placement->nodeName,
-											   placement->nodePort);
+		chosenConnection = StartNodeConnection(flags, nodeName, nodePort);
```
mtuncer: this function is ~350 lines long. we should consider refactoring it

src/backend/distributed/connection/placement_connection.c
```diff
@@ -695,6 +738,34 @@ CanUseExistingConnection(uint32 flags, const char *userName,
 
 
 /*
+ * AccessedNonColocatedPlacement returns true if the connection accessed another
+ * placement in the same colocation group with a different representative value,
+ * meaning it's not strictly colocated.
+ */
+static bool
+AccessedNonColocatedPlacement(MultiConnection *connection, ShardPlacement *placement)
```
mtuncer: Can you pick a different name ? 
This makes me think it is about deciding if the transaction had accessed any non co-located placement whereas it is about accessing a different placement of a a relation in the same co-location group.

src/backend/distributed/commands/multi_copy.c
marcocitus: copy ... to ... goes through the real-time executor, which does not use the connection API, so it would not use the right connections and thus would not see preceding writes, or would form a self-deadlock in case of DDL.

src/backend/distributed/connection/placement_connection.c
marcocitus: Moved the main connection finding logic into a separate `FindPlacementListConnection` function.

PR #1522, opened 2017-07-27
master_activate_node UDF also returns noderole

lithp:
Fix bug introduced by https://github.com/citusdata/citus/pull/1476

Before this PR master_activate_node fails when used as a range table:

```
-SELECT master_activate_node('localhost', :worker_2_port);
-           master_activate_node            
--------------------------------------------
- (3,3,localhost,57638,default,f,t,primary)
-(1 row)
-
+SELECT isactive FROM master_activate_node('localhost', :worker_2_port);
+ERROR:  function return row and query-specified return row do not match
+DETAIL:  Returned row contains 8 attributes, but query expects 7.
```

### DIFF
--- a/src/test/regress/multi_schedule
+++ b/src/test/regress/multi_schedule
@@ -64,7 +64,8 @@ test: multi_task_assignment_policy multi_cross_shard
 test: multi_utility_statements
 test: multi_dropped_column_aliases
 test: multi_binary_master_copy_format
-test: multi_prepare_sql multi_prepare_plsql
+test: multi_prepare_sql
+test: multi_prepare_plsql
 test: multi_sql_function
 test: multi_view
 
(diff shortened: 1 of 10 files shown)


src/backend/distributed/citus--7.0-4--7.0-5.sql
```diff
@@ -95,4 +95,21 @@ COMMENT ON FUNCTION master_add_inactive_node(nodename text,nodeport integer,
                                              groupid integer, noderole noderole)
   IS 'prepare node by adding it to pg_dist_node';
 
+DROP FUNCTION master_activate_node(text, integer);
+CREATE FUNCTION master_activate_node(nodename text,
```
marcocitus: We usually don't modify existing SQL files, since some users build from `master` and missing a change would force them to drop the extension. Probably doesn't affect anyone, but may better to add a new SQL file?

lithp:
Okay, bumped the version instead of updating the upgrade script

src/test/regress/multi_schedule
```diff
@@ -64,7 +64,8 @@ test: multi_task_assignment_policy multi_cross_shard
 test: multi_utility_statements
 test: multi_dropped_column_aliases
 test: multi_binary_master_copy_format
-test: multi_prepare_sql multi_prepare_plsql
+test: multi_prepare_sql
+test: multi_prepare_plsql
```
marcocitus: I guess #1503 should undo this change

PR #1523, opened 2017-07-27
Convert the global wait edges to adjacency list

onderkalaci:
Gradually getting towards distributed deadlock detection. I prefer to continue with smaller commits, that's why I'm opening this piece of the deadlock detection now.

In this commit, we add ability to convert the global wait edges into adjacency list with the following format:
 `[transactionId] = [transactionNode->waitsFor {list of waiting transaction nodes}]`

### DIFF
--- a/src/backend/distributed/test/distributed_deadlock_detection.c
+++ b/src/backend/distributed/test/distributed_deadlock_detection.c
@@ -1,0 +1,144 @@
+/*-------------------------------------------------------------------------
+ *
+ * test/src/distributed_deadlock_detection.c
+ *
+ * This file contains functions to exercise distributed deadlock detection
+ * related lower level functionality.
+ *
+ * Copyright (c) 20167, Citus Data, Inc.
+ *
+ *-------------------------------------------------------------------------
+ */
+#include "postgres.h"
+#include "funcapi.h"
+#include "miscadmin.h"
+
+#include "access/hash.h"
+#include "distributed/backend_data.h"
+#include "distributed/distributed_deadlock_detection.h"
+#include "distributed/hash_helpers.h"
+#include "distributed/listutils.h"
+#include "distributed/lock_graph.h"
+#include "distributed/metadata_cache.h"
+#include "distributed/transaction_identifier.h"
+#include "nodes/pg_list.h"
+#include "utils/hsearch.h"
+#include "utils/timestamp.h"
+
+
+static char * WaitsForToString(List *waitsFor);
+
+
+PG_FUNCTION_INFO_V1(get_adjacency_list_wait_graph);
+
+
+/*
+ * get_adjacency_list_wait_graph returns the wait graph in adjacency list format. For the
+ * details see BuildAdjacencyListForWaitGraph().
+ *
+ * This function is mostly useful for testing and debugging purposes.
+ */
+Datum
+get_adjacency_list_wait_graph(PG_FUNCTION_ARGS)
+{
+	ReturnSetInfo *returnSetInfo = (ReturnSetInfo *) fcinfo->resultinfo;
+	TupleDesc tupleDescriptor = NULL;
+	Tuplestorestate *tupleStore = NULL;
+	MemoryContext perQueryContext = NULL;
+	MemoryContext oldContext = NULL;
+
+	WaitGraph *waitGraph = NULL;
+	HTAB *adjacencyList = NULL;
+	HASH_SEQ_STATUS status;
+	TransactionNode *transactionNode = NULL;
+
+	const int attributeCount = 2;
+	Datum values[attributeCount];
+	bool isNulls[attributeCount];
+
+	CheckCitusVersion(ERROR);
+
+	/* check to see if caller supports us returning a tuplestore */
+	if (returnSetInfo == NULL || !IsA(returnSetInfo, ReturnSetInfo))
+	{
+		ereport(ERROR,
+				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+				 errmsg("set-valued function called in context " \
+						"that cannot accept a set")));
+	}
+
+	if (!(returnSetInfo->allowedModes & SFRM_Materialize))
+	{
+		ereport(ERROR,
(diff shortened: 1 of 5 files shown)


src/backend/distributed/transaction/distributed_deadlock_detection.c
```diff
@@ -0,0 +1,303 @@
...
+
+PG_FUNCTION_INFO_V1(get_flattened_wait_graph);
+
+
+/*
+ * get_flattened_wait_graph returns the flattened wait graph. This function
+ * is mostly useful for testing and debugging purposes.
+ */
+Datum
+get_flattened_wait_graph(PG_FUNCTION_ARGS)
```
marcocitus: I don't think this UDF adds any value for the user over `dump_global_wait_edges` (also, using a comma-separated string instead of an array would be kind of strange). Could we move it to a test file?

src/backend/distributed/transaction/distributed_deadlock_detection.c
```diff
@@ -0,0 +1,303 @@
...
+/*
+ * FlattenWaitGraph converts a wait graph which is represented as a set of
+ * wait edges between all backends in the Citus cluster, to a set of nodes,
+ * one for each distributed transaction.
+ *
+ * The returned structure is an adjacency list with the following format:
+ * [transactionId] = [transactionNode->waitsFor {list of waiting transaction nodes}]
+ */
+static HTAB *
+FlattenWaitGraph(WaitGraph *waitGraph)
```
marcocitus: We could name this something more descriptive, thinking out loud: `BuildDistributedTransactionWaitsForHash`. 

Also wondering: Should this be in `lock_graph.c`? In that case we could even make WaitGraph private.

src/backend/distributed/transaction/distributed_deadlock_detection.c
```diff
@@ -0,0 +1,303 @@
...
+	}
+
+	return distributedTransactionHash;
+}
+
+
+/*
+ * GetTransactionNode searches distributedTransactionHash for the given
+ * given transactionId. If the transaction is not found, the corresponding
+ * transactionNode is initialized to be NULL.
```
marcocitus: "the corresponding transactionNode is initialized to be NULL." -> "a new transaction node with the given transaction identifier is added".

onderkalaci:
Just a note on testing these PRs if anyone interested. 

I find it easy to test on the local tables with

```
ALTER SYSTEM set deadlock_timeout TO `1200s`;
SELECT pg_reload_conf(); /* not sure if really needed */

-- session 1
BEGIN;
Select assign_distributed_transaction_id(0, 1, now());
... 

-- session 2
BEGIN;
Select assign_distributed_transaction_id(0, 2, now());


...

-- session n
BEGIN;
Select assign_distributed_transaction_id(0, n, now());
...

```

Which simulates distributed transactions and allows us to see deadlocks on the graph.

src/backend/distributed/transaction/distributed_deadlock_detection.c
```diff
@@ -0,0 +1,303 @@
...
+
+/*
+ * DistributedTransactionIdCompare compares DistributedTransactionId's a and b
+ * and returns -1 if a < b, 1 if a > b, 0 if they are equal.
+ *
+ * DistributedTransactionId are first compared by their timestamp, then transaction
+ * number, then node identifier.
+ */
+static int
+DistributedTransactionIdCompare(const void *a, const void *b, Size keysize)
```
marcocitus: These could maybe go into a utility file, since I imagine we'll want to compare distributed transaction IDs in other places.

src/backend/distributed/transaction/distributed_deadlock_detection.c
```diff
@@ -0,0 +1,303 @@
...
+}
+
+
+/*
+ * GetTransactionNode searches distributedTransactionHash for the given
+ * given transactionId. If the transaction is not found, the corresponding
+ * transactionNode is initialized to be NULL.
+ */
+static TransactionNode *
+GetTransactionNode(HTAB *distributedTransactionHash,
```
marcocitus: Ozgun would probably argue that this should be `GetOrCreateTransactionNode` :)

src/backend/distributed/transaction/distributed_deadlock_detection.c
```diff
@@ -0,0 +1,193 @@
...
+ *        blocking transactions in the adjacency list
+ *          - if not found, add new node(s) to the list
+ *      - Add blocking transaction to the waiting transaction's waitFor
+ *        list
+ *
+ *  The format of the adjacency list becomes the following:
+ *      [transactionId] = [transactionNode->waitsFor {list of waiting transaction nodes}]
+ */
+HTAB *
+BuildAdjacencyListForWaitGraph(WaitGraph *waitGraph)
```
marcocitus: Maybe List -> Lists

src/backend/distributed/test/distributed_deadlock_detection.c
```diff
@@ -0,0 +1,144 @@
...
+
+	MemoryContextSwitchTo(oldContext);
+
+	waitGraph = BuildGlobalWaitGraph();
+	adjacencyList = BuildAdjacencyListForWaitGraph(waitGraph);
+
+	/* iterate on all nodes */
+	hash_seq_init(&status, adjacencyList);
+
+	while ((transctionNode = (TransactionNode *) hash_seq_search(&status)) != 0)
```
marcocitus: transction -> transaction

src/include/distributed/distributed_deadlock_detection.h
```diff
@@ -0,0 +1,35 @@
...
+typedef struct TransactionNode
+{
+	DistributedTransactionId transactionId;
+
+	/* list of TransactionNode that this distributed transaction is waiting for */
+	List *waitsFor;
+} TransactionNode;
+
+
+HTAB * BuildAdjacencyListForWaitGraph(WaitGraph *waitGraph);
```
marcocitus: +extern

PR #1524, opened 2017-07-28
Add a regression test for citus.max_task_string_size

mtuncer:
regression test for previously introduced citus.max_task_string_size GUC.


src/test/regress/expected/multi_task_string_size.out
```diff
@@ -0,0 +1,239 @@
...
+(1 row)
+
+-- setting can not be changed on runtime
+SET citus.max_task_string_size TO 20000;
+ERROR:  parameter "citus.max_task_string_size" cannot be changed without restarting the server
+-- error message may vary between executions
+-- hiding warning and error message
+-- no output means the query has failed
+SET client_min_messages to FATAL;
+SELECT u.* FROM wide_table u JOIN wide_table v ON (u.long_column_002 = v.long_column_003);
```
marcocitus: I think the only thing we're really testing here is whether we correctly error out when the string is very long. We would perhaps just add a query with a very long string in the target list to one of the existing task-tracker tests.

marcocitus: Thinking about it more: can we make Citus to throw an error early?

mtuncer: There are few caveats here. This error is thrown from worker when a query with long text is received. Worker checks the query size and compares it against **local** setting do decide if it needs to throw error.  Setting in coordinator/master is not relevant here.

Can we enforce some GUC variables to have the same value in all/some nodes in the cluster ?

Issue #866, opened 2016-10-11
Warnings show up in schema-only pg_dump due to dependency loops


lfittl:
Splitting this out from #348 per a discussion with @ozgune and @anarazel.

When running a schema-only pg_dump, as some migration frameworks do (e.g. `rake db:migrate` with Rails/ActiveRecord), the following warnings appear:

```
  pg_dump: [sorter] WARNING: could not resolve dependency loop among these items:
  pg_dump: [sorter]   TABLE impressions  (ID 240 OID 19654)
  pg_dump: [sorter] WARNING: could not resolve dependency loop among these items:
  pg_dump: [sorter]   TABLE clicks  (ID 241 OID 19661)
```

**These are harmless** since we can still successfully dump the schema on the master, but people have been confused by this a couple of times.

The fix for this seems to be rather involved, and might require some upstream Postgres changes - for now one should just ignore these.

anarazel:
The issue is actually easily solvable.  The issue is that we've a dependency from the sharding expression to the table, to prevent the partition column from being dropped.  You'd get
```
ERROR:  2BP01: cannot drop table test_table2 column test_id because other objects depend on it
DETAIL:  table test_table2 depends on table test_table2 column test_id
HINT:  Use DROP ... CASCADE to drop the dependent objects too.
LOCATION:  reportDependentObjects, dependency.c:985
Time: 5.596 ms
``` 
but we also have another check these days:
```
ERROR:  XX000: cannot execute ALTER TABLE command involving partition column
LOCATION:  ErrorIfUnsupportedAlterTableStmt, multi_utility.c:1700
Time: 4.988 ms
```
so we can just drop the expression dependency.

anarazel:
If we want to keep the dependency, we could alternatively have a dependency from pg_dist_partition to the expression.  That'd allow us to drop the check.

PR #1527, opened 2017-07-28
Remove distributed tables' dependency on distribution key columns.

pykello:
We already check that we cannot drop distribution key columns in ErrorIfUnsupportedAlterTableStmt() at multi_utility.c, so we don't need to have distributed table to distribution key column dependency to avoid dropping of distribution key column.

Furthermore, having this dependency causes some warnings in pg_dump --schema-only (See #866), like the following, which are not desirable:

```
  pg_dump: [sorter] WARNING: could not resolve dependency loop among these items:
  pg_dump: [sorter]   TABLE impressions  (ID 240 OID 19654)
  pg_dump: [sorter] WARNING: could not resolve dependency loop among these items:
  pg_dump: [sorter]   TABLE clicks  (ID 241 OID 19661)
```

This change also allows drop of distribution keys when `SET citus.enable_ddl_propagation to false;`, which is the expected behaviour. Regression tests are updated accordingly.

TESTING: Manual. Perl tap tests which are used to test client tools like pg_dump aren't enabled for extensions yet, so I'm postponing automated tests to after we have tap tests for extensions.

### DIFF
--- a/src/backend/distributed/executor/multi_utility.c
+++ b/src/backend/distributed/executor/multi_utility.c
@@ -1783,15 +1783,38 @@ ErrorIfUnsupportedAlterTableStmt(AlterTableStmt *alterTableStatement)
 
 /*
  * ErrorIfDropPartitionColumn checks if any subcommands of the given alter table
- * command is a DROP COLUMN command which drops the partition column. If there is
- * such a subcommand, this function errors out.
+ * command is a DROP COLUMN command which drops the partition column of a distributed
+ * table. If there is such a subcommand, this function errors out.
  */
 static void
 ErrorIfAlterDropsPartitionColumn(AlterTableStmt *alterTableStatement)
 {
+	LOCKMODE lockmode = 0;
+	Oid leftRelationId = InvalidOid;
+	bool isDistributedRelation = false;
 	List *commandList = alterTableStatement->cmds;
 	ListCell *commandCell = NULL;
 
+	/* first check whether a distributed relation is affected */
+	if (alterTableStatement->relation == NULL)
+	{
+		return;
+	}
+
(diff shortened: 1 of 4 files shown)


marcocitus:
If you drop a distribution column using enable_ddl_propagation off and then use the table, what happens?

pykello:
In our multitenant tutorial example, when I drop campaigns.company_id and continue using the table it seems to work find and send worker queries. I am not sure why and how. I am continuing to do more tests.

What is the expected behaviour when enable_ddl_propagation is off? If the expected behaviour is to not be able to drop distribution column, shouldn't we check it using ErrorIfUnsupportedAlterTableStmt()?

pykello:
I'll add the check for drop distribution column and re-submit for review.

sumedhpathak:
@pykello Is this ready for review? Thanks!

pykello:
@sumedhpathak yes. @marcocitus please take a look.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -1806,6 +1782,91 @@ ErrorIfUnsupportedAlterTableStmt(AlterTableStmt *alterTableStatement)
...
+	tuple = SearchSysCacheAttName(relationId, alterColumnName);
+	if (HeapTupleIsValid(tuple))
+	{
+		Form_pg_attribute targetAttr = (Form_pg_attribute) GETSTRUCT(tuple);
+
+		/* reference tables do not have partition column, so allow them */
+		if (partitionColumn != NULL &&
+			targetAttr->attnum == partitionColumn->varattno)
+		{
+			ereport(ERROR, (errmsg("cannot execute ALTER TABLE command "
```
mtuncer: I would be happier if we could display a separate error for drop

PR #1529, opened 2017-08-01
Deadlock detection main

onderkalaci:
This is a WIP, opened to get some early feedback. The PR detects distributed deadlocks and kills one of the participating backends on the coordinator.

Things to complete before merge:

* [x] Add citus.deadlock_timeout GUC, as 1+ float factor of postgres' deadlock timeout
* [x] Discuss removing citus.enable_deadlock_prevention (I couldn't see any discussions on removing the GUC, are we planning to keep it?)
* [x] Add more regression tests, currently there is only one.
* [x] Add more meaningful error messages, currently the only error message is `ERROR:  canceling statement due to user request` (See Andres' prototype on how to implement it)
* [x] Add some debug messages as Postgres does
* [x] Make sure that deadlock detection works on MX as expected (See Marco's `deadlock_detection` branch)
* [x] Lessen the debug messages
* [ ] Do some benchmarks
* [ ] Consider failure cases a bit more. Be creative to generate test cases.
* [x] Split into smaller commits
* [x] Add some regression tests for replication > 1, deadlock includes a local transaction and multi-shard queries.
* [ ] MX tests.


src/backend/distributed/transaction/distributed_deadlock_detection.c
```diff
@@ -25,11 +25,247 @@
...
+ * The function goes over all the backends, checks for the backend with
+ * the same transaction number as the given transaction node.
+ */
+static void
+AssicateDistributedTransactionWithBackendProc(TransactionNode *transactionNode)
+{
+	int backendIndex = 0;
+
+	for (backendIndex = 0; backendIndex < MaxBackends; ++backendIndex)
+	{
```
anarazel: This seems like a really expensive approach. Why do we need this? In my prototype that certainly wasn't necessary?  We already have the PID in the earlier represenation, we should just avoid throwing it away...

src/backend/distributed/transaction/distributed_deadlock_detection.c
```diff
@@ -25,11 +25,247 @@
...
+}
+
+
+/*
+ * GetSortedDistributedTransactionIdList gets an adjaceny list and
+ * returns a list where each element is a node in the adjacency list.
+ *
+ * The list is then sorted via DistributedTransactionIdCompare (i.e., sort
+ * by timestamp -> transaction number -> node id).
+ */
```
anarazel: You can't return a list and then sort it...

src/backend/distributed/transaction/distributed_deadlock_detection.c
```diff
@@ -25,11 +25,247 @@
...
+
+/*
+ * CheckForDistributedDeadlocks is the entry point for detecing
+ * distributed deadlocks.
+ *
+ * In plain words, the function first builds a wait graph by
+ * adding the wait edges from the local node and then adding the
+ * remote wait edges to form a global wait graph. Later, the wait
+ * graph is converted into an adjacency list for more efficient
+ * searches. Finally, a DFS is done on the adjacency list.
```
anarazel: I don't really like the phrasing here - we're really converting it from an edge list to adjacency lists (plural, it's multiple lists). Both are representations of the graph.

src/backend/distributed/transaction/distributed_deadlock_detection.c
```diff
@@ -25,11 +25,247 @@
...
+	ListCell *transactionIdCell = NULL;
+
+	/*
+	 * We need to iterate on all transactions since it is not guranteed that the
+	 * transactions form a connected graph.
+	 *
+	 * We use sorted list to get (i) predictable results (ii) kill the youngest
+	 * transaction (i.e., if a DDL continues for 1 hour and deadlocks with a
+	 * SELECT continues for 10 msec, we prefer to kill the SELECT).
+	 */
```
onderkalaci: Makes sense, I think it'll be trivial to print the participating transactions with that.

src/backend/distributed/transaction/distributed_deadlock_detection.c
onderkalaci: I removed sorting at all, followed your above suggestion

src/backend/distributed/transaction/distributed_deadlock_detection.c
```diff
@@ -25,11 +25,247 @@
...
+PG_FUNCTION_INFO_V1(check_distributed_deadlocks);
+
+
+/*
+ * check_distributed_deadlocks is the external API for manually
+ * checking for distributed deadlocks. For the details, see
+ * CheckForDistributedDeadlocks().
+ */
+Datum
+check_distributed_deadlocks(PG_FUNCTION_ARGS)
```
anarazel: Nah, I was joking. I like the name (FWIW, it's a jokey reference to both Woody Guthrie's Guitar sticker and the Laundry files book series), but it's probably not quite appropriate...

src/backend/distributed/transaction/distributed_deadlock_detection.c
anarazel: >> This seems like a really expensive approach.
> That's right. But, we're doing it only after the deadlock detected. So, should be acceptable?

Once for all processes though. We really should check for deadlocks of multiple initiating processes, not just one.

> So, with my understanding, this is necessary to find the pid on the coordinator process.

> As far as I can see, we had a similar loop here. I think the necessity arises because while we form the waitGraph we treat remote and local wait edges equally, which makes hard to find the pid on the coordinator node.

I don't buy this - we only can find processes that originate on the current node, and we can just assign all pids initially for all of them.  I just kept that around for all "start" nodes, and I think that makes sense.

src/backend/distributed/transaction/distributed_deadlock_detection.c
```diff
@@ -25,11 +25,264 @@
...
+	HTAB *adjacencyList = BuildAdjacencyListsForWaitGraph(waitGraph);
+	HASH_SEQ_STATUS status;
+	TransactionNode *transactionNode = NULL;
+
+	/*
+	 * We iterate on transaction nodes and search for deadlocks where the
+	 * starting node is the given transaction node.
+	 */
+	hash_seq_init(&status, adjacencyList);
+	while ((transactionNode = (TransactionNode *) hash_seq_search(&status)) != 0)
```
anarazel: Am I missing something (I'm quite tired and I'm drinking a glass of nice red wine) or are you now check for deadlocks starting at *any* node in the graph, rather than just "initiating" nodes as I'd done in my POC? I don't think that's right.

src/backend/distributed/transaction/distributed_deadlock_detection.c
```diff
@@ -25,11 +25,264 @@
...
+
+		/* don't need to revisit the node again */
+		if (waitingTransactionNode->transactionVisited)
+		{
+			continue;
+		}
+
+		waitingTransactionNode->transactionVisited = true;
+
+		if (CheckDeadlockForTransactionNodeRecurse(startingTransactionNode,
```
anarazel: I think it's a really bad to convert this to recursion - the risk of stack overflow is quite real. Why did you change this?

src/backend/distributed/transaction/distributed_deadlock_detection.c
```diff
@@ -25,11 +25,264 @@
...
+			}
+
+			/* we should find the backend */
+			Assert(youngestTransaction->initiatorProc != NULL);
+
+			KillBackendDueToDeadlock(youngestTransaction->initiatorProc);
+
+			hash_seq_term(&status);
+
+			return true;
```
anarazel: I don't think we should return here. We really should search for multiple deadlocks in one go, otherwise we might never get ones that happen to be later in the hash space, if there's continually new ones "early" in the hash space.

Now that's not entirely trivial because we'll often find a deadlock coming from two different starting nodes. I think that means we need to keep a bit more state in the hash table elements. We could: a) just not reset the visited field when we found a deadlock, and move the reset to the end of `CheckDeadlockForTransactionNode`. That would avoid finding the same deadlock from different ends. b) keep a separate 'inDeadlock' field that's set when find a deadlock that then triggers a new `CheckDeadlockForTransactionNode` to fail when it re-finds such a node. c) more complex: remove cycles from the graph.

src/backend/distributed/transaction/distributed_deadlock_detection.c
onderkalaci: Hmm, that's right. My intention was to keep track `deadlockPath `, which seemed simpler with this method.

onderkalaci: I couldn't come up with an easy solution fo a non-recursive traversal that also finds the path. 

Any suggestions? Shall I ignore finding the path for now and kill the youngest as in the first version of the PR (i.e., sort the distributed transaction ids wrt time, then iterate over them while searching for the deadlock)?

anarazel:
On 2017-08-04 13:56:06 +0000, Önder Kalacı wrote:
> I couldn't come up with an easy solution fo a non-recursive traversal that also finds the path. 

When pushing a new node on the stack of to-be-visited-nodes, also push a
second element that denotes the currently visited path? Then that
shouldn't be too hard to keep track off?

- Andres

src/backend/distributed/transaction/distributed_deadlock_detection.c
marcocitus: On the other hand, it does keep the code a lot simpler and understandable if we just return here, and  we'll always kill another deadlock in 100ms, so we'll still be able to get rid of them at ~10/sec. I don't know if hundreds of deadlocks forming simultaneously is common enough to justify the extra code complexity.

src/backend/distributed/transaction/distributed_deadlock_detection.c
onderkalaci: Yes, I broke that and will fix in the next commit.

src/backend/distributed/transaction/distributed_deadlock_detection.c
onderkalaci: Same here, I'd rather do smth like the following as marco suggested in the daemon:

```C
		/*
		 * If we find any deadlocks, run the distributed deadlock detection
		 * more often since it is likely that that are other deadlocks needs to
		 * be resolved.
		 */
		if (foundDeadlock)
		{
			timeout = 100.0;
		}
		else
		{
			timeout = DistributedDeadlockDetectionTimeoutFactor * DeadlockTimeout;
		}

		StartTransactionCommand();
		foundDeadlock = CheckForDistributedDeadlocks();
		CommitTransactionCommand();
```

onderkalaci:
@anarazel @marcocitus I've pushed some changes. 

The PR is not very well tested yet, but I think ready for the second pass of review as I continue to test and add some more regression tests (see my WIP testing [document](https://docs.google.com/document/d/1yx4VYCeZOmL64ure9riSnQeSGlVXPnSY-rgPUHpHPJM/edit)).

 Sometimes the newly added isolation tests fail because deadlock detection may kick in before the next step is executed. I'm trying to make test outputs more consistent.

onderkalaci:
> Sometimes the newly added isolation tests fail because deadlock detection may kick in before the next step is executed. I'm trying to make test outputs more consistent.

Also, I just realized that some MX tests fail (due to xact participant changes). I'll fix them tomorrow.

src/backend/distributed/shared_library_init.c
```diff
@@ -711,6 +769,35 @@ RegisterCitusConfigVariables(void)
...
+
+/*
+ * We don't want our users to set this value to less than 1.0. We intend that only for testing.
+ */
+static void
+WarningForDeadlockTimeout(double newval, void *extra)
+{
+	if (newval < 1.0)
+	{
+		ereport(WARNING, (errcode(ERRCODE_WARNING_DEPRECATED_FEATURE),
```
onderkalaci: change `errcode`

src/backend/distributed/citus--7.0-8--7.0-9.sql
```diff
@@ -0,0 +1,12 @@
+/* citus--7.0-8--7.0-9 */
+
+SET search_path = 'pg_catalog';
+
+CREATE FUNCTION check_distributed_deadlocks()
```
marcocitus: why do we need this UDF?

onderkalaci: Hmm, initially it helped me for testing & debugging purposes. But, you're right, we don't need to expose this I think.

src/backend/distributed/transaction/distributed_deadlock_detection.c
anarazel: I don't really buy the complexity argument, it doesn't seem that hard. But I think we can move this to later, it's not critical right now.

src/backend/distributed/citus--7.0-8--7.0-9.sql
```diff
@@ -0,0 +1,12 @@
...
+RETURNS BOOL
+LANGUAGE 'c' STRICT
+AS $$MODULE_PATHNAME$$, $$check_distributed_deadlocks$$;
+COMMENT ON FUNCTION check_distributed_deadlocks()
+IS 'does a distributed deadlock check, if a deadlock found kills one of the participating backends and returns true ';
+
+RESET search_path;
```
anarazel: missing end-of-line

src/backend/distributed/citus--7.0-8--7.0-9.sql
anarazel: Hm, might be interesting for deadlock check isolation test reproducability? We could disable / move to infinity the background check and manually trigger the check in a separate session.

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -838,26 +834,6 @@ GetModifyConnections(Task *task, bool markCritical, bool noNewTransactions)
 		multiConnection = GetPlacementListConnection(connectionFlags, placementAccessList,
 													 NULL);
 
-		/*
```
anarazel: @marcocitus was wondering - on a call maybe? - whether we should force 2PC to on for transactions that span multiple nodes. What do you guys think?

src/backend/distributed/shared_library_init.c
```diff
@@ -188,6 +196,32 @@ _PG_init(void)
 
 
 /*
+ * multi_log_hook intercepts postgres log commands. We use this to show the user a
+ * meaningful error message when a process receives a SIGINT from the distributed
+ * deadlock detector.
```
anarazel: I think this is too specific. Maybe something like "We use this to override postgres error messages when they're not specific enough"? And then in the *body* describe the specific override? Because I'd be surprised if there weren't future ones coming...

src/backend/distributed/shared_library_init.c
```diff
@@ -188,6 +196,32 @@ _PG_init(void)
...
+ */
+static void
+multi_log_hook(ErrorData *edata)
+{
+	if (edata->elevel == ERROR && edata->sqlerrcode == ERRCODE_QUERY_CANCELED &&
+		MyBackendGotKilledDueToDeadlock())
```
anarazel: Well, it formally didn't get killed yet, but ;). And it's also not really killing, more cancelling - the session continues to live afterwards after all?

src/backend/distributed/shared_library_init.c
```diff
@@ -188,6 +196,32 @@ _PG_init(void)
...
+		oldContext = MemoryContextSwitchTo(ErrorContext);
+
+		edata->sqlerrcode = ERRCODE_T_R_DEADLOCK_DETECTED;
+		edata->message = "distributed deadlock detected and to resolve the deadlock "
+						 "this backend is cancelled";
+		edata->detail = "Set citus.log_distributed_deadlock_detection to 'on' "
+						"and check server log for details.";
+
+		MemoryContextSwitchTo(oldContext);
+	}
```
anarazel: What does that MemoryContextSwitchTo do, given you just referring to constants?

src/backend/distributed/shared_library_init.c
```diff
@@ -327,6 +361,18 @@ RegisterCitusConfigVariables(void)
 		NULL, NULL, NULL);
 
 	DefineCustomBoolVariable(
+		"citus.log_distributed_deadlock_detection",
+		gettext_noop("Log distributed deadlock detection related processing in "
+					 "the server log"),
```
anarazel: Hm, do we really need this? Or is this only temporarily for debugging?

src/backend/distributed/shared_library_init.c
```diff
@@ -361,6 +407,18 @@ RegisterCitusConfigVariables(void)
 		0,
 		NULL, NULL, NULL);
 
+	DefineCustomRealVariable(
+		"citus.distributed_deadlock_detection_factor",
+		gettext_noop("Sets the time to wait before checking for distributed "
+					 "deadlocks. The value is multiplied with Postgres' "
+					 "deadlock_timeout setting."),
```
anarazel: Isn't it the other way round? I know that multiplication is commutative here, but still, deadlock_timeout is the setting with the unit...

src/backend/distributed/shared_library_init.c
anarazel: Why do we allow this in the first place? This seems unnecessary?

src/backend/distributed/transaction/distributed_deadlock_detection.c
```diff
@@ -24,12 +26,344 @@
...
+	 * starting node is the given transaction node.
+	 */
+	hash_seq_init(&status, adjacencyLists);
+	while ((transactionNode = (TransactionNode *) hash_seq_search(&status)) != 0)
+	{
+		bool deadlockFound = false;
+		List *deadlockPath = NIL;
+
+		/* we're only interested in finding deadlocks originating from this node */
+		if (transactionNode->transactionId.initiatorNodeIdentifier != GetLocalGroupId())
```
anarazel: Should store GetLocalGroupId() in a local variable, lest it's called quite a number of times.

src/backend/distributed/transaction/distributed_deadlock_detection.c
```diff
@@ -24,12 +26,344 @@
...
+
+				AssocateDistributedTransactionWithBackendProc(currentNode);
+
+				LogTransactionNode(currentNode);
+
+				if (timestamptz_cmp_internal(currentTimestamp, youngestTimestamp) == 1)
+				{
+					youngestTransaction = currentNode;
+				}
+			}
```
anarazel: Does this logic guarantee it's the youngest participant on the current node, rather than from another node?

src/backend/distributed/transaction/distributed_deadlock_detection.c
```diff
@@ -24,12 +26,344 @@
...
+
+			/* we should find the backend */
+			Assert(youngestTransaction->initiatorProc != NULL);
+
+			KillBackendDueToDeadlock(youngestTransaction->initiatorProc);
```
anarazel: I think this really should be `CancelTransactionDueToDeadlock`

src/backend/distributed/transaction/distributed_deadlock_detection.c
```diff
@@ -24,12 +26,344 @@
...
+/*
+ * RemoveElementsUptoNode gets a transaction list and a transactionNode.
+ * The function searches for the given transactionNode in the list. If it
+ * finds any, the elements up to the last found node (including it) are
+ * removed from the list. If not any found, the list returned as it is.
+ */
+static List *
+RemoveElementsUptoNode(List *waitingTransactionList,
+					   TransactionNode *transactionNode)
+{
```
anarazel: This seems quite complicated. I don't quite get why you're not emulating what the callstack previously did in the recursive version?  You can have a very similar logic to maintain the list of nodes leading to the cycle?  A bit simplified: Constantly keep track of the "depth" of the search, by putting that into the thing you push onto the 'please-visit-me' list. In parallel keep an array of 'MaxBackends' size. Whenever visiting the next node, fill the current depth with the to-be-visited node.  And voila, you should end up with the firsth depth entries of that array describing the cycle?

src/backend/distributed/shared_library_init.c
```diff
@@ -327,6 +361,18 @@ RegisterCitusConfigVariables(void)
 		NULL, NULL, NULL);
 
 	DefineCustomBoolVariable(
+		"citus.log_distributed_deadlock_detection",
+		gettext_noop("Log distributed deadlock detection related processing in "
+					 "the server log"),
+		NULL,
+		&LogDistributedDeadlockDetection,
+		false,
+		PGC_USERSET,
```
anarazel: I don't think `USERSET` is the right category, setting it in a session won't have an effect. I think it should be `PGC_SIGHUP`, no?

src/backend/distributed/executor/multi_router_executor.c
marcocitus: Would be a good idea. I think we can honour multi_shard_commit_protocol as well.

src/backend/distributed/citus--7.0-8--7.0-9.sql
marcocitus: It's currently not used in the isolation tests

src/backend/distributed/shared_library_init.c
```diff
@@ -711,6 +769,35 @@ RegisterCitusConfigVariables(void)
 
 
 /*
+ * Inform the users about the depraced flag.
```
marcocitus: deprecated

src/backend/distributed/shared_library_init.c
```diff
@@ -188,6 +196,32 @@ _PG_init(void)
...
+	if (edata->elevel == ERROR && edata->sqlerrcode == ERRCODE_QUERY_CANCELED &&
+		MyBackendGotKilledDueToDeadlock())
+	{
+		MemoryContext oldContext = NULL;
+
+		oldContext = MemoryContextSwitchTo(ErrorContext);
+
+		edata->sqlerrcode = ERRCODE_T_R_DEADLOCK_DETECTED;
+		edata->message = "distributed deadlock detected and to resolve the deadlock "
+						 "this backend is cancelled";
```
marcocitus: backend -> transaction

src/backend/distributed/shared_library_init.c
onderkalaci: Yes, I'll update the functions & comments.

src/test/regress/expected/multi_mx_explain.out
```diff
@@ -360,10 +360,8 @@ Custom Scan (Citus Router)
   Tasks Shown: All
   ->  Task
         Node: host=localhost port=57638 dbname=regression
-        ->  Bitmap Heap Scan on lineitem_mx_1220055 lineitem_mx
-              Recheck Cond: (l_orderkey = 5)
-              ->  Bitmap Index Scan on lineitem_mx_pkey_1220055
-                    Index Cond: (l_orderkey = 5)
+        ->  Index Scan using lineitem_mx_pkey_1220055 on lineitem_mx_1220055 lineitem_mx
```
onderkalaci: I have no idea why this test was failing. But, it consistently produces the same output for now.

src/backend/distributed/citus--7.0-9--7.0-10.sql
```diff
@@ -0,0 +1,13 @@
+/* citus--7.0-8--7.0-9 */
+
+SET search_path = 'pg_catalog';
+
+CREATE FUNCTION check_distributed_deadlocks()
+RETURNS BOOL
+LANGUAGE 'c' STRICT
+AS $$MODULE_PATHNAME$$, $$check_distributed_deadlocks$$;
+COMMENT ON FUNCTION check_distributed_deadlocks()
+IS 'does a distributed deadlock check, if a deadlock found kills one of the participating backends and returns true ';
```
onderkalaci: `kills` -> `cancels`

src/backend/distributed/utils/maintenanced.c
```diff
@@ -258,6 +264,28 @@ CitusMaintenanceDaemonMain(Datum main_arg)
...
+			CommitTransactionCommand();
+		}
+
+		/*
+		 * If we find any deadlocks, run the distributed deadlock detection
+		 * more often since it is likely that that are other deadlocks needs to
+		 * be resolved. Thus, we use 1/20 of the calculated value. With the default
+		 * values (i.e., deadlock_timeout 1 seconds,
+		 * citus.distributed_deadlock_detection_factor 2), we'd be able to cancel
+		 * ~10 distributed deadlocks.
```
onderkalaci: `per second`

anarazel:
(Gah, github once more doesn't allow me to reply to that thread. I hate you github)

>  Without this, it'd be really hard to debug anything related to deadlock detection. I'd keep it if you're not > strongly against it. I'm planning to output fewer things in the upcoming commits.

> Also, it's GUC_NO_SHOW_ALL so shouldn't matter for users?

It's less the GUC, than a lot of code that's going to age horribly.

anarazel:
> However, in order to disable deadlock detection, we either need to introduce one more flag (which seems not a good idea) or use this one. I pick 1000 as the value which disables the distributed deadlock detection. Any thoughts?

I'd rather go with -1 then, and leave the check hook in place and disallow values that are smaller than 1.0, except for -1.

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -839,22 +834,17 @@ GetModifyConnections(Task *task, bool markCritical, bool noNewTransactions)
...
 		/*
-		 * If already in a transaction, disallow expanding set of remote
-		 * transactions. That prevents some forms of distributed deadlocks.
+		 * If we're expanding the nodes that participate in the distributed
+		 * transaction, conform to MultiShardCommitProtocol.
 		 */
-		if (noNewTransactions)
+		if (MultiShardCommitProtocol == COMMIT_PROTOCOL_2PC &&
+			InCoordinatedTransaction() &&
+			XactModificationLevel == XACT_MODIFICATION_DATA)
```
anarazel: Why the `XactModificationLevel` check here? Also, isn't this going to trigger for transactions that only touch a single shard? That might be costly for some rails etc users that use transactions for trivial one statement writes.

anarazel:
>> Should store GetLocalGroupId() in a local variable, lest it's called quite a number of times.

> GetLocalGroupId() already does that, should be enough?

Still an external function call etc.

src/backend/distributed/transaction/distributed_deadlock_detection.c
```diff
@@ -24,12 +26,335 @@
...
+			TransactionNode *waitForTransaction =
+				(TransactionNode *) lfirst(currentWaitForCell);
+
+			/*
+			 * We add both the waiting node and its parent to simplify finding the
+			 * deadlock path.
+			 */
+			waitingTransactionNodes =
+				lcons(waitingTransactionNode, waitingTransactionNodes);
+
```
anarazel: Huh, old code?

src/backend/distributed/transaction/distributed_deadlock_detection.c
```diff
@@ -24,12 +26,335 @@
...
+	/* iterate on the children of the starting node and set their depths */
+	foreach(waitingTransactionCell, waitingTransactionNodes)
+	{
+		TransactionNode *waitingTransactionNode =
+			(TransactionNode *) lfirst(waitingTransactionCell);
+
+		waitingTransactionNode->currentStackDepth =
+			startingTransactionNode->currentStackDepth + 1;
+	}
```
anarazel: I don't think this is quite right. We shouldn't keep the stack depth in the TransactionNode, it should be part of the `waitingTransactionNodes` list (which I think would better be called to `todoList`, or `queudNodes` or something - everything is waiting here, so `waitingTransactionNodes` isn't very descriptive.

What I'm basically suggesting is that the "todo list" keeps not just a pointer to the node, but instead a pair ala `(pointerToNode, stackDepthWhenQueueing)`. Storing the depth inside the node isn't correct - a node can be queued from multiple places in the graph...

src/backend/distributed/transaction/distributed_deadlock_detection.c
```diff
@@ -24,12 +26,335 @@
...
+ * stack to the deadlockPath.
+ */
+static void
+FindDeadlockPath(TransactionNode *cycledTransactionNode,
+				 TransactionNode **transactionNodeStack,
+				 List **deadlockPath)
+{
+	int deadlockStackDepth = cycledTransactionNode->currentStackDepth;
+	int stackIndex = 0;
+
```
anarazel: Think it'd be good to reset *deadlockPath to NIL first.

src/backend/distributed/transaction/distributed_deadlock_detection.c
```diff
@@ -24,12 +26,335 @@
...
+}
+
+
+/*
+ * FindDeadlockPath gets the transaction node on which the cycle is found. Later,
+ * we add the depth number of elements of the cycled node from the transaction
+ * stack to the deadlockPath.
+ */
+static void
+FindDeadlockPath(TransactionNode *cycledTransactionNode,
```
anarazel: Rename to `BuildDeadlockPathList` or such? There's not much finding involved now.

src/backend/distributed/transaction/distributed_deadlock_detection.c
anarazel: If you want I can push something showing what I mean?

src/backend/distributed/transaction/distributed_deadlock_detection.c
```diff
@@ -24,12 +26,335 @@
...
+
+	return false;
+}
+
+
+/*
+ * FindDeadlockPath gets the transaction node on which the cycle is found. Later,
+ * we add the depth number of elements of the cycled node from the transaction
+ * stack to the deadlockPath.
+ */
```
anarazel: The comment describes the manner, rather than the result. I think it'd be much more helpful if it said "* returns a list of transactions involved in a distributed deadlock (i.e. a cycle in the graph)" or such.

src/backend/distributed/utils/maintenanced.c
```diff
@@ -258,6 +264,28 @@ CitusMaintenanceDaemonMain(Datum main_arg)
...
+		if (DistributedDeadlockDetectionTimeoutFactor != 1000)
+		{
+			StartTransactionCommand();
+			foundDeadlock = CheckForDistributedDeadlocks();
+			CommitTransactionCommand();
+		}
+
+		/*
+		 * If we find any deadlocks, run the distributed deadlock detection
+		 * more often since it is likely that that are other deadlocks needs to
```
anarazel: s/likely/quite possible/?

src/test/regress/pg_regress_multi.pl
```diff
@@ -236,6 +236,15 @@ sub revert_replace_postgres
 push(@pgOptions, '-c', "citus.shard_replication_factor=2");
 push(@pgOptions, '-c', "citus.node_connection_timeout=${connectionTimeout}");
 
+if($isolationtester)
+{
```
anarazel: This needs explanation.

src/backend/distributed/utils/maintenanced.c
onderkalaci: Just to understand better, why do we think in that way? I'll bring this up on our meeting to understand more.

src/backend/distributed/executor/multi_router_executor.c
onderkalaci: > Also, isn't this going to trigger for transactions that only touch a single shard? 

No, that's not. We set `XACT_MODIFICATION_DATA ` after the execution is done, at the end of `ExecuteSingleModifyTask()`. Thus, it helps to understand that we've already executed another command before this. 

Also, note that we use the same approach on the master.

Some execution outputs from this PR:

```SQL

-- single statement, without any transactions
insert into test VALUES (7,7);
DEBUG:  distributed statement: INSERT INTO public.test_102016 (a, b) VALUES (7, 7)
DEBUG:  Creating router plan
DEBUG:  assigned task 0 to node localhost:9702
LOG:  issuing INSERT INTO public.test_102016 (a, b) VALUES (7, 7)
INSERT 0 1


-- single statement with a tx block
postgres=# BEGIN;
BEGIN
postgres=# insert into test VALUES (7,7);
DEBUG:  distributed statement: INSERT INTO public.test_102016 (a, b) VALUES (7, 7)
DEBUG:  Creating router plan
DEBUG:  assigned task 0 to node localhost:9702
LOG:  issuing BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED;SELECT assign_distributed_transaction_id(0, 3, '2017-08-10 11:15:11.944518+03')
LOG:  issuing INSERT INTO public.test_102016 (a, b) VALUES (7, 7)
INSERT 0 1
postgres=# COMMIT;
LOG:  issuing COMMIT
COMMIT

-- set 2pc, multiple statements that go to the same node
postgres=# SET citus.multi_shard_commit_protocol TO '2pc';
SET
postgres=# BEGIN;
BEGIN
postgres=# insert into test VALUES (7,7);
DEBUG:  distributed statement: INSERT INTO public.test_102016 (a, b) VALUES (7, 7)
DEBUG:  Creating router plan
DEBUG:  assigned task 0 to node localhost:9702
LOG:  issuing BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED;SELECT assign_distributed_transaction_id(0, 4, '2017-08-10 11:16:19.871937+03')
LOG:  issuing INSERT INTO public.test_102016 (a, b) VALUES (7, 7)
INSERT 0 1
postgres=# insert into test VALUES (7,7);
DEBUG:  distributed statement: INSERT INTO public.test_102016 (a, b) VALUES (7, 7)
DEBUG:  Creating router plan
DEBUG:  assigned task 0 to node localhost:9702
LOG:  issuing INSERT INTO public.test_102016 (a, b) VALUES (7, 7)
INSERT 0 1
postgres=# COMMIT;
LOG:  issuing COMMIT
COMMIT

-- set 2pc, multiple statements that go to the different nodes
postgres=# SET citus.multi_shard_commit_protocol TO '2pc';

postgres=# BEGIN;
BEGIN
postgres=# insert into test VALUES (7,7);
DEBUG:  distributed statement: INSERT INTO public.test_102016 (a, b) VALUES (7, 7)
DEBUG:  Creating router plan
DEBUG:  assigned task 0 to node localhost:9702
LOG:  issuing BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED;SELECT assign_distributed_transaction_id(0, 5, '2017-08-10 11:16:49.273187+03')
LOG:  issuing INSERT INTO public.test_102016 (a, b) VALUES (7, 7)
INSERT 0 1
postgres=# insert into test VALUES (8,7);
DEBUG:  distributed statement: INSERT INTO public.test_102008 (a, b) VALUES (8, 7)
DEBUG:  Creating router plan
DEBUG:  assigned task 0 to node localhost:9700
LOG:  issuing BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED;SELECT assign_distributed_transaction_id(0, 5, '2017-08-10 11:16:49.273187+03')
LOG:  issuing INSERT INTO public.test_102008 (a, b) VALUES (8, 7)
INSERT 0 1
postgres=# COMMIT;
LOG:  issuing PREPARE TRANSACTION 'citus_0_78762_5_0'
LOG:  issuing PREPARE TRANSACTION 'citus_0_78762_5_1'
LOG:  issuing COMMIT PREPARED 'citus_0_78762_5_0'
LOG:  issuing COMMIT PREPARED 'citus_0_78762_5_1'
COMMIT

-- and finally single statement with 2PC set

postgres=# BEGIN;
BEGIN
postgres=# insert into test VALUES (8,7);
DEBUG:  distributed statement: INSERT INTO public.test_102008 (a, b) VALUES (8, 7)
DEBUG:  Creating router plan
DEBUG:  assigned task 0 to node localhost:9700
LOG:  issuing BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED;SELECT assign_distributed_transaction_id(0, 6, '2017-08-10 11:17:21.274619+03')
LOG:  issuing INSERT INTO public.test_102008 (a, b) VALUES (8, 7)
INSERT 0 1
postgres=# COMMIT;
LOG:  issuing COMMIT
COMMIT

```
(Sorry for so long comment 😄 )

src/backend/distributed/transaction/distributed_deadlock_detection.c
onderkalaci: I had it at some point but lost at one of the commits. I'll add that again.

src/backend/distributed/transaction/distributed_deadlock_detection.c
```diff
@@ -24,12 +26,346 @@
...
+/*
+ * AssocateDistributedTransactionWithBackendProc gets a transaction node
+ * and searches the corresponding backend. Once found, transactionNodes'
+ * initiatorProc is set to it.
+ *
+ * The function goes over all the backends, checks for the backend with
+ * the same transaction number as the given transaction node.
+ */
+static void
+AssocateDistributedTransactionWithBackendProc(TransactionNode *transactionNode)
```
marcocitus: Assocate -> Associate

src/backend/distributed/transaction/backend_data.c
```diff
@@ -566,3 +573,58 @@ GetBackendDataForProc(PGPROC *proc, BackendData *result)
...
+ * MyBackendGotCancelledDueToDeadlock returns whether the current distributed
+ * transaction was cancelled due to a deadlock. If the backend is not in a
+ * distributed transaction, the function returns false.
+ */
+bool
+MyBackendGotCancelledDueToDeadlock(void)
+{
+	bool cancelledDueToDeadlock = false;
+
+	SpinLockAcquire(&MyBackendData->mutex);
```
anarazel: Check for NULL.

src/backend/distributed/transaction/backend_data.c
```diff
@@ -566,3 +573,58 @@ GetBackendDataForProc(PGPROC *proc, BackendData *result)
...
+/*
+ * CancelTransactionDueToDeadlock cancels the input proc and also marks the backend
+ * data with this information.
+ */
+void
+CancelTransactionDueToDeadlock(PGPROC *proc)
+{
+	BackendData *backendData = &backendManagementShmemData->backends[proc->pgprocno];
+
+	SpinLockAcquire(&backendData->mutex);
```
anarazel: Check for NULL.

src/backend/distributed/transaction/distributed_deadlock_detection.c
```diff
@@ -24,12 +26,352 @@
...
+CheckForDistributedDeadlocks(void)
+{
+	WaitGraph *waitGraph = BuildGlobalWaitGraph();
+	HTAB *adjacencyLists = BuildAdjacencyListsForWaitGraph(waitGraph);
+	HASH_SEQ_STATUS status;
+	TransactionNode *transactionNode = NULL;
+	int edgeCount = waitGraph->edgeCount;
+
+	/* avoid calling GetLocalGroup() many times by caching for the session */
+	if (localGroupId == -1)
```
anarazel: Only store on function local level.

src/include/distributed/transaction_identifier.h
```diff
@@ -21,6 +21,10 @@
  *
  *  -  initiatorNodeIdentifier: A unique identifier of the node that initiated
  *     the distributed transaction
+ *  -  transcationStartedOnNode: Set to true only for the transactions initialized on
```
anarazel: s/transcation/transaction/

anarazel: Maybe rename to transactionOriginator or transactionMaster or somesuch? Because StartedOnNode could as well be taken to be true for all mx queries reaching the local node...

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -844,22 +839,17 @@ GetModifyConnections(Task *task, bool markCritical, bool noNewTransactions)
 													 NULL);
 
 		/*
-		 * If already in a transaction, disallow expanding set of remote
-		 * transactions. That prevents some forms of distributed deadlocks.
+		 * If we're expanding the nodes that participate in the distributed
```
anarazel: "expanding the nodes" sounds a bit weird, maybe "expanding the set of nodes"?

src/backend/distributed/shared_library_init.c
anarazel: As I wrote somewhere else, I'm less concerned about the guc, then about rotting code.

src/backend/distributed/transaction/backend_data.c
```diff
@@ -566,3 +573,70 @@ GetBackendDataForProc(PGPROC *proc, BackendData *result)
...
+/*
+ * CancelTransactionDueToDeadlock cancels the input proc and also marks the backend
+ * data with this information.
+ */
+void
+CancelTransactionDueToDeadlock(PGPROC *proc)
+{
+	BackendData *backendData = &backendManagementShmemData->backends[proc->pgprocno];
+
+	/* backend might not have started */
```
anarazel: 'not started' doesn't seem right. Maybe 'might not have used citus yet and thus not initialized backend data'?

src/backend/distributed/transaction/backend_data.c
```diff
@@ -566,3 +573,70 @@ GetBackendDataForProc(PGPROC *proc, BackendData *result)
...
+ * MyBackendGotCancelledDueToDeadlock returns whether the current distributed
+ * transaction was cancelled due to a deadlock. If the backend is not in a
+ * distributed transaction, the function returns false.
+ */
+bool
+MyBackendGotCancelledDueToDeadlock(void)
+{
+	bool cancelledDueToDeadlock = false;
+
+	/* backend might not have started */
```
anarazel: see above.

src/backend/distributed/transaction/distributed_deadlock_detection.c
```diff
@@ -24,12 +26,352 @@
 #include "utils/hsearch.h"
 #include "utils/timestamp.h"
 
+/* GUC, determining whether debug messages for deadlock detection sent to LOG */
+bool LogDistributedDeadlockDetection = false;
+
+/* we access local group quite heavily, thus cache it via a local variable */
+static int localGroupId = -1;
```
anarazel: Please move to function local.

src/backend/distributed/transaction/distributed_deadlock_detection.c
```diff
@@ -24,12 +26,352 @@
...
+ *
+ * In plain words, the function first builds a wait graph by
+ * adding the wait edges from the local node and then adding the
+ * remote wait edges to form a global wait graph. Later, the wait
+ * graph is converted into another graph representation (adjacency
+ * lists) for more efficient searches. Finally, a DFS is done on
+ * the adjacency lists. Finding a cycle in the graph unveils a
+ * distributed deadlock. Upon finding a deadlock, the youngest
+ * participant backend is cancelled.
+ *
```
anarazel: Probably worth remarking on the complexity of the algorithm somewhere here? It's essentially worst case `O((|local-backends| + |max-locks|) * |local-backends|)`, or maybe easier to say, it's `O(N)` for each transaction that's checked for deadlocks, and there can be multiple ones that need to be checked.

src/backend/distributed/utils/maintenanced.c
```diff
@@ -248,7 +253,8 @@ CitusMaintenanceDaemonMain(Datum main_arg)
 	{
 		int rc;
 		int latchFlags = WL_LATCH_SET | WL_TIMEOUT | WL_POSTMASTER_DEATH;
-		int timeout = 10000; /* wake up at least every so often */
+		double timeout = 10000.0; /* use this if the deadlock detection is disabled */
```
anarazel: Why is this a double? The timeout is specified as an int/long, so this doesn't seem right.

onderkalaci: Doing the calculations with `double` seems better because we might divide the value? 

I added explicit cast on `WaitLatch` and a single line comment. Does that work?

src/include/distributed/transaction_identifier.h
onderkalaci: I liked the `transactionOriginator`

onderkalaci:
@anarazel I updated the PR, added some more tests, the remaining items are

src/backend/distributed/utils/maintenanced.c
anarazel: Still find it ugly, but with the cast it's at least explicit ;)

anarazel:
TL;DR: No performance issues found.

I've done a bit of benchmarking. Comparing the following:
- pgbench readonly, with just selects. No difference above the noise. None expected - there are no distributed transactions (just router executable queries)
- pgbench transactionless writes (txs don't work in master)
- pgbench "stock" writes (doesn't work in master)
- pgbench "insert only" (to assuage the minimal case with transaction identifiers)

The results so far show that we don't seem to have introduced a measurable/relevant performance penalty (got 0.05, but that's way below the run-to-run noise).   Even for the post-deadlock stuff I've not seen the deadlock detection itself be a bottleneck. There's considerably latency issues when there's a lot of deadlocks, but that's to be expected. They can to some degree be addressed by reducing the deadlock_timeout.  In the longer run it might be worthwhile to have a bit more adaptive pacing of deadlock checks.

src/backend/distributed/shared_library_init.c
```diff
@@ -195,6 +203,29 @@ _PG_init(void)
...
+	/*
+	 * Show the user a meaningful error message when a backend is cancelled
+	 * by the distributed deadlock detection.
+	 */
+	if (edata->elevel == ERROR && edata->sqlerrcode == ERRCODE_QUERY_CANCELED &&
+		MyBackendGotCancelledDueToDeadlock())
+	{
+		edata->sqlerrcode = ERRCODE_T_R_DEADLOCK_DETECTED;
+		edata->message = "distributed deadlock detected and to resolve the deadlock "
+						 "this transaction is cancelled";
```
anarazel: Think this message could be improved by rephrasing it to something like "canceling transaction due to involvement in distributed deadlock".

src/backend/distributed/shared_library_init.c
```diff
@@ -195,6 +203,29 @@ _PG_init(void)
...
+		edata->detail = "Set citus.log_distributed_deadlock_detection to 'on' "
+						"and check server log for details.";
```
anarazel: Not a fan of this hint. For one cloud customers won't be able to set it, for another it's output even if it's already on.  It's also a GUC you explicitly advocated as not being user visible.  I think we should either remove the detail, or replace it with something that just says "Check server log for details.", and  a followup improvement always log useful things for user diagnostics, rather than development diagnostics.

src/backend/distributed/shared_library_init.c
anarazel: I think this is severely conflating two different things. User targeted and developer targeted logging. It's sure not very friendly to users atm, and those are going to be the main consumers of this.

src/backend/distributed/transaction/distributed_deadlock_detection.c
anarazel: FWIW, I think this is something that should be revised in 7.1.

src/include/distributed/distributed_deadlock_detection.h
```diff
@@ -26,10 +26,29 @@ typedef struct TransactionNode
...
 	List *waitsFor;
+
+	/* backend that is on the initiator node */
+	PGPROC *initiatorProc;
+
+	bool transactionVisited;
 } TransactionNode;
 
+/* used only for finding the deadlock cycle path */
+typedef struct QueuedTransactionNode
```
anarazel: Why is this in the header?

anarazel:
@onderkalaci Could you create issues for the remaining items? I have:
- logging
- tests for mx
- cleanup of backend association

but there might be some others?

anarazel:
@onderkalaci I see you have a cleaned up branch with properly split commits, you plan to push this to this PR and then merge this one?

src/backend/distributed/shared_library_init.c
onderkalaci: I'm removing the detail for now. I think it makes sense that we should show this after we add more useful things to users.

src/include/distributed/distributed_deadlock_detection.h
onderkalaci: Nothing special just forgot here. I'll move it to the source.

onderkalaci:
> @onderkalaci I see you have a cleaned up branch with properly split commits, you plan to push this to this PR and then merge this one?

I already did this, so after addressing your final comments & opening the issue, I'll merge this. (I'll also add some regression tests with reference tables, just FYI)

src/backend/distributed/shared_library_init.c
onderkalaci: I updated to `canceling the transaction since it has involved in a distributed deadlock`, feel free to comment here so I can update it back.

sumedhpathak:
🎆

src/backend/distributed/transaction/distributed_deadlock_detection.c
```diff
@@ -25,11 +27,355 @@
...
+	/*
+	 * We iterate on transaction nodes and search for deadlocks where the
+	 * starting node is the given transaction node.
+	 */
+	hash_seq_init(&status, adjacencyLists);
+	while ((transactionNode = (TransactionNode *) hash_seq_search(&status)) != 0)
+	{
+		bool deadlockFound = false;
+		List *deadlockPath = NIL;
+		TransactionNode *transactionNodeStack[edgeCount];
```
lithp: I'm super late here but I think this is subtly wrong. You count the number of edges and use that as a proxy for the total number of nodes which could exist. However, it's possible for there to be one more node than there are edges. e.x. `A -> B -> C`. When that happens we write past the end of the array.

src/backend/distributed/transaction/distributed_deadlock_detection.c
```diff
@@ -25,11 +27,355 @@
...
+	foreach(currentWaitForCell, transactionNode->waitsFor)
+	{
+		TransactionNode *waitForTransaction =
+			(TransactionNode *) lfirst(currentWaitForCell);
+		QueuedTransactionNode *queuedNode = palloc0(sizeof(QueuedTransactionNode));
+
+		queuedNode->transactionNode = waitForTransaction;
+		queuedNode->currentStackDepth = currentStackDepth;
+
+		*toBeVisitedNodes = lappend(*toBeVisitedNodes, queuedNode);
```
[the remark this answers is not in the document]
onderkalaci: Why is that? `D` won't be in the `deadlockPath` that we form. `D` would be removed from the `transactionNodeStack` given that `B` and `D` would have the same `stackDepth` and we do the following: `transactionNodeStack[currentStackDepth] = currentTransactionNode;`

If `D` isn't in the `deadlockPath` it is not going to be killed at all.

Also, the deadlock path would be `A->B>C->A`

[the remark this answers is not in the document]
onderkalaci: Oh, wait. I see now what you mean. That's unfortunate that we use `lappend` instead of `lcons` in `PrependOutgoingNodesToQueue()`.

(Well, I'm pretty sure that I've used `lcons` at some point, but, I cannot remember why it's `lappend` now. Let me think some more before opening the PR for the fix)

src/backend/distributed/transaction/distributed_deadlock_detection.c
onderkalaci: > I'm super late here but I think this is subtly wrong.

That's true as well. We should allocate `edgeCount + 1` elements in the array. I think the likelihood of this happening is low, but, we should definitely fix it and backport.

src/backend/distributed/transaction/distributed_deadlock_detection.c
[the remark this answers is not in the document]
jasonmp85: FWIW this was reported as a new issue in #1960 and fixed in #1961.

Issue #1528, opened 2017-07-31
Support Follower Clusters


lithp:
Work to be done
- [x] (#1518) Add `pg_dist_node.nodecluster` column
  - Give every node a cluster which it is a part of, default to 'default' and make sure no existing behavior changes.
- [x] (#1530) Add `citus.cluster_name` GUC, and all functions only see nodes which are a part of the current cluster
- [x] (#1540) Add `citus.use_secondary_nodes`, which allows the coordinator in the follower cluster to use the nodes in the follower cluster
- [x] (#1542) Consider adding a `master_add_secondary_node` function
- [x] (#1543) Consider adding a test that you can run queries from a streaming replica?

Questions

- [x] Should you be allowed to add `primary` nodes to non-`default` clusters?

lfittl:
@lithp @ozgune Can this issue be closed, or is there still work required to support follower clusters?

lithp:
As far as I know it can be closed.

PR #1530, opened 2017-08-01
Add citus.cluster_name GUC

lithp:
No part of Citus sees any node which is not in the current cluster.

(A subcomponent of #1528, depends on #1518)

- [x] I think `AddNodeMetadata` will allow you to add duplicate nodes, because `FindWorkerNode` uses the worker node hash
- [x] Add test that primaries in other clusters aren't used when running queries
- [x] Add test that you can remove nodes in other clusters
- [x] Add test that you cannot add primary nodes to the non-default cluster

After this PR:
- I'm not sure if it makes sense to be able to toggle `isActive` on nodes in other clusters but this PR preserves that ability.

### DIFF
--- a/src/backend/distributed/master/master_metadata_utility.c
+++ b/src/backend/distributed/master/master_metadata_utility.c
@@ -174,8 +174,6 @@ DistributedTableSize(Oid relationId, char *sizeQuery)
 
 	ErrorIfNotSuitableToGetSize(relationId);
 
-	LockRelationOid(DistNodeRelationId(), AccessShareLock);
-
 	workerNodeList = ActivePrimaryNodeList();
 
 	foreach(workerNodeCell, workerNodeList)
--- a/src/backend/distributed/master/master_metadata_utility.c
+++ b/src/backend/distributed/master/master_metadata_utility.c
@@ -158,7 +158,6 @@ static uint64
 DistributedTableSize(Oid relationId, char *sizeQuery)
 {
 	Relation relation = NULL;
-	Relation pgDistNode = NULL;
 	List *workerNodeList = NULL;
 	ListCell *workerNodeCell = NULL;
 	uint64 totalRelationSize = 0;
@@ -175,7 +174,7 @@ DistributedTableSize(Oid relationId, char *sizeQuery)
 
 	ErrorIfNotSuitableToGetSize(relationId);
 
-	pgDistNode = heap_open(DistNodeRelationId(), AccessShareLock);
+	LockRelationOid(DistNodeRelationId(), AccessShareLock);
 
 	workerNodeList = ActivePrimaryNodeList();
 
@@ -187,7 +186,6 @@ DistributedTableSize(Oid relationId, char *sizeQuery)
 		totalRelationSize += relationSizeOnNode;
 	}
 
-	heap_close(pgDistNode, NoLock);
 	heap_close(relation, AccessShareLock);
 
 	return totalRelationSize;
--- a/src/backend/distributed/shared_library_init.c
+++ b/src/backend/distributed/shared_library_init.c
@@ -670,6 +670,16 @@ RegisterCitusConfigVariables(void)
 		0,
 		NULL, NULL, NULL);
 
+	DefineCustomStringVariable(
+		"citus.cluster_name",
+		gettext_noop("Which cluster this node is a part of"),
+		NULL,
+		&CurrentCluster,
+		"default",
+		PGC_SU_BACKEND,
+		0,
+		NULL, NULL, NULL);
+
 	DefineCustomBoolVariable(
 		"citus.enable_version_checks",
 		gettext_noop("Enables version checks during CREATE/ALTER EXTENSION commands"),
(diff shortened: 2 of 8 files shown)


src/backend/distributed/master/master_metadata_utility.c
```diff
@@ -175,7 +174,7 @@ DistributedTableSize(Oid relationId, char *sizeQuery)
 
 	ErrorIfNotSuitableToGetSize(relationId);
 
-	pgDistNode = heap_open(DistNodeRelationId(), AccessShareLock);
+	LockRelationOid(DistNodeRelationId(), AccessShareLock);
 
 	workerNodeList = ActivePrimaryNodeList();
```
marcocitus: It might make sense to put the lock into `ActivePrimaryNodeList`. I can't really come up with a reason not do so.

src/backend/distributed/utils/node_metadata.c
```diff
@@ -835,8 +872,16 @@ GetNodeTuple(char *nodeName, int32 nodePort)
 	heapTuple = systable_getnext(scanDescriptor);
 	if (!HeapTupleIsValid(heapTuple))
 	{
-		ereport(ERROR, (errmsg("could not find valid entry for node \"%s:%d\"",
-							   nodeName, nodePort)));
+		if (raiseError)
+		{
+			ereport(ERROR, (errmsg("could not find valid entry for node \"%s:%d\"",
```
marcocitus: is this something we can leave to the caller when NULL is returned?

lithp: That's probably better, I'm not happy that we've now duplicated the error message but I think you're right that the code is cleaner now.

src/backend/distributed/master/master_metadata_utility.c
lithp: Good point, although I pushed it all the way down to WorkerNodeHash, which I think covers every entrypoint to the metadata cache.

marcocitus:
The `current` feels slightly weird to me, as if it will change later on, what if we do `citus.cluster_name`?

lithp:
@marcocitus 

You're right, `current_cluster` is a bad name. Some other alternatives:

- `citus.cluster_name`
- `citus.node_cluster`
- `citus.cluster_to_use`
- `citus.which_node_cluster_to_query`

I'm kind of leaning towards `citus.which_node_cluster_to_query`

marcocitus:
> citus.which_node_cluster_to_query

It's descriptive, but it has some strange conceptual implications (e.g. coordinator is not part of the cluster, but rather queries it, but then what about MX?). I think either `cluster_name` or `node_cluster` works for me.

lithp:
Renamed to `cluster_name`.

src/backend/distributed/shared_library_init.c
```diff
@@ -670,6 +670,16 @@ RegisterCitusConfigVariables(void)
 		0,
 		NULL, NULL, NULL);
 
+	DefineCustomStringVariable(
+		"citus.cluster_name",
+		gettext_noop("Which cluster this coordinator sends queries to"),
```
marcocitus: maybe "sends queries to" -> "is part of"?

src/backend/distributed/shared_library_init.c
```diff
@@ -670,6 +670,16 @@ RegisterCitusConfigVariables(void)
...
+		NULL,
+		&CurrentCluster,
+		"default",
+		PGC_USERSET,
```
lithp: I'm probably overthinking this, but how does `BACKEND` sound?

Going through the options in `guc.h`:
- `POSTMASTER` options can only be changed by restarting the cluster
- `SIGHUP` options can be changed by sending `HUP` to the postmaster/backend
- `BACKEND` can be changed but remains constant from the point of view of any given backend. It must be changed by editing the config file or specifying it in `PGOPTIONS` at startup.

It seems that,
+ `POSTMASTER` perfectly fits the use-case we currently envision, you setup a coordinator in a different cluster and let it run, but it might be too inflexible? It's harder to test this if you have to reconfigure and restart the cluster to change `cluster_name`.
+ The backends accept the `SIGHUP` before processing every packet so it's possible for the value of `citus.cluster_name` to change between planning and execution (Between the `Bind` and `Execute` messages) That sounds too difficult to have to think about while coding.
+ However, `BACKEND` maintains a constant cluster during the lifetime of any backend, and lets you change the cluster future backends use. This might be a useful operation for administrators? I can see it being useful for testing Citus.

If you're worried about users being able to set it, we can use `SU_BACKEND` which disables the startup packet option unless you're a superuser.

marcocitus:
Fine with SU_BACKEND

src/backend/distributed/shared_library_init.c
```diff
@@ -670,6 +670,16 @@ RegisterCitusConfigVariables(void)
 		0,
 		NULL, NULL, NULL);
 
+	DefineCustomStringVariable(
+		"citus.cluster_name",
+		gettext_noop("Which cluster this coordinator is a part of"),
```
marcocitus: I think maybe just use node instead of coordinator

src/backend/distributed/shared_library_init.c
```diff
@@ -670,6 +670,16 @@ RegisterCitusConfigVariables(void)
...
+		NULL,
+		&CurrentCluster,
+		"default",
+		PGC_USERSET,
```
marcocitus: SU_BACKEND seems fine

PR #1535, opened 2017-08-03
start_metadata_sync_to_node doesn't work with secondary nodes

### LINKED ISSUE #1533: start_metadata_sync_to_node() doesn't check whether the node is a secondary
When called with a secondary node this throws a confusing error:

```
SELECT groupid AS worker_1_group FROM pg_dist_node WHERE nodeport = :worker_1_port \gset
SELECT master_add_node('localhost', 8888, groupid => :worker_1_group, noderole => 'secondary');
              master_add_node               
--------------------------------------------
 (4,1,localhost,8888,default,f,t,secondary)
(1 row)

SELECT start_metadata_sync_to_node('localhost', 8888);
WARNING:  connection error: localhost:8888
DETAIL:  no connection to the server

ERROR:  failure on connection marked as essential: localhost:8888
SELECT stop_metadata_sync_to_node('localhost', 8888);
 stop_metadata_sync_to_node 
----------------------------
 
(1 row)
```

lithp:
Fixes #1533

### DIFF
--- a/src/backend/distributed/metadata/metadata_sync.c
+++ b/src/backend/distributed/metadata/metadata_sync.c
@@ -111,6 +111,15 @@ start_metadata_sync_to_node(PG_FUNCTION_ARGS)
 
 	MarkNodeHasMetadata(nodeNameString, nodePort, true);
 
+	if (!WorkerNodeIsPrimary(workerNode))
+	{
+		/*
+		 * If this is a secondary node we can't actually sync metadata to it; we assume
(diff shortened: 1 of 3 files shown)


marcocitus:
Should we also be setting the hasmetadata flag on secondaries when a primary gets metadata to make sure this state is preserved after a failover? Or should we handle this at failover time?

lithp:
@marcocitus Good point, it just flips the flag now.

PR #1539, opened 2017-08-08
Add Savepoint Support

### LINKED ISSUE #1493: Add support for SAVEPOINT / ROLLBACK TO SAVEPOINT
The following pattern will currently not work with Citus, since we don't support rolling back to a savepoint in distributed transactions:

```
citus=> CREATE TABLE my_savepoint_test(id int);
CREATE TABLE
citus=> SELECT create_distributed_table('my_savepoint_test', 'id');
 create_distributed_table 
--------------------------
 
(1 row)

citus=> BEGIN;
BEGIN
citus=> SAVEPOINT my_savepoint;
SAVEPOINT
citus=> INSERT INTO my_savepoint_test(id) VALUES (123);
INSERT 0 1
citus=> ROLLBACK TO SAVEPOINT my_savepoint;
ROLLBACK
citus=> COMMIT;
ERROR:  cannot ROLLBACK TO SAVEPOINT in transactions which modify distributed tables
```

This is a common pattern in some frameworks (e.g. ActiveRecord), in particular for unit tests where savepoints are used to model nested transactions (i.e. the unit test has one big wrapping transaction, and explicit transactions in the application code are modeled as savepoints)

Whilst good to have both for single-shard and cross-shard use cases, for now this mostly seems to be a problem for the typical multi-tenant application use case where only a single shard (or tenant) is modified in the larger wrapping transaction.

pykello:
This change adds support for SAVEPOINT, ROLLBACK TO SAVEPOINT, and RELEASE SAVEPOINT.

When transaction connections are not established yet, the savepoints are kept in a stack and sent to the worker when the connection is later established. After establishing connections, savepoint commands are sent as they arrive.

This change fixes #1493 .


src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -112,6 +114,19 @@ FinishRemoteTransactionBegin(struct MultiConnection *connection)
 	{
 		Assert(PQtransactionStatus(connection->pgConn) == PQTRANS_INTRANS);
 	}
+
+	if (!transaction->transactionFailed)
```
anarazel: Move to statement built in `StartRemoteTransactionBegin()`, to avoid additional roundtrips.

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -820,6 +835,112 @@ CoordinatedRemoteTransactionsAbort(void)
...
+	dlist_iter iter;
+
+	dlist_foreach(iter, &InProgressTransactions)
+	{
+		MultiConnection *connection = dlist_container(MultiConnection, transactionNode,
+													  iter.cur);
+		StringInfo command = makeStringInfo();
+		appendStringInfo(command, "SAVEPOINT savepoint_%u", subId);
+		RemoteSubTransactionCommand(connection, command);
+	}
```
anarazel: I'd like split this into Start/Finish* as well, to avoid adding latencies for all connections. I.e. in Start* send the `SAVEPOINT` to all asynchronously, and then only wait for all of them together.

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -820,6 +835,112 @@ CoordinatedRemoteTransactionsAbort(void)
 
 
 /*
+ * CoordinatedRemoteTransactionsSavepointBegin sends the SAVEPOINT command for
+ * the given sub-transaction id to all connections participating in the current
+ * transaction.
+ */
+void
+CoordinatedRemoteTransactionsSavepointBegin(SubTransactionId subId)
```
anarazel: Naming-wise I think this should be `RemoteSavepointBegin` or maybe `RemoteTransactionSavepointBegin`, rather than the current name. The worker transaction stuff doesn't have the Coordinated* prefix.

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -26,7 +26,9 @@
 #include "distributed/worker_manager.h"
 #include "utils/hsearch.h"
 
-
```
anarazel: I'd not remove this newline ;)

src/test/regress/expected/multi_modifying_xacts.out
```diff
@@ -119,8 +119,7 @@ EXCEPTION
         RAISE NOTICE 'caught not_null_violation';
 END $$;
 NOTICE:  caught not_null_violation
-COMMIT;
-ERROR:  cannot ROLLBACK TO SAVEPOINT in transactions which modify distributed tables
+ROLLBACK;
 \set VERBOSITY default
```
anarazel: @jasonmp85 Do you recall what you were trying to test? It's a bit weird that we're catching & ignoring an error here...

anarazel: I think this means that the savepoint rollback logic needs to reset ->transactionFailed too (when it's been "triggered" inside the relevant savepoint(s)).

jasonmp85: This is exercising the implicit savepoints created by the PL/pgSQL's `BEGIN`/`EXCEPTION`/`END` syntax, I think. The test above tests an explicit `SAVEPOINT` and this tests an implicit one.

pykello:
@anarazel can you please take a look at this to see what I've missed (specially in functions at remote_transactions.c I suspect there should be some checks that I migh have missed), while I work on adding some more tests and adding comments to functions?

Thanks.

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -67,11 +82,21 @@ StartRemoteTransactionBegin(struct MultiConnection *connection)
 	 */
 	distributedTransactionId = GetCurrentDistributedTransactionId();
 	appendStringInfo(beginAndSetDistributedTransactionId,
-					 "SELECT assign_distributed_transaction_id(%d, %ld, '%s')",
+					 "SELECT assign_distributed_transaction_id(%d, %ld, '%s');",
 					 distributedTransactionId->initiatorNodeIdentifier,
 					 distributedTransactionId->transactionNumber,
 					 timestamptz_to_str(distributedTransactionId->timestamp));
 
+	/* append queued savepoints for this transaction */
```
anarazel: s/queued/in-progress/?

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -820,6 +845,247 @@ CoordinatedRemoteTransactionsAbort(void)
...
+{
+	dlist_iter iter;
+
+	/* asynchronously send SAVEPOINT */
+	dlist_foreach(iter, &InProgressTransactions)
+	{
+		MultiConnection *connection = dlist_container(MultiConnection, transactionNode,
+													  iter.cur);
+		RemoteTransaction *transaction = &connection->remoteTransaction;
+		if (transaction->lastUnfailedSubXact <= subId &&
```
anarazel: The function header needs to clearly explain this. Also, should we really reset it before having succeeded sending the rollback?

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -820,6 +845,247 @@ CoordinatedRemoteTransactionsAbort(void)
...
+		{
+			continue;
+		}
+
+		FinishRemoteTransactionSavepointRollback(connection, subId);
+	}
+}
+
+
+static void
```
anarazel: Comments...

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -820,6 +845,247 @@ CoordinatedRemoteTransactionsAbort(void)
...
+	}
+
+	/* and wait for the results */
+	dlist_foreach(iter, &InProgressTransactions)
+	{
+		MultiConnection *connection = dlist_container(MultiConnection, transactionNode,
+													  iter.cur);
+		RemoteTransaction *transaction = &connection->remoteTransaction;
+
+		if (transaction->transactionFailed)
```
anarazel: Right now you seem to be duplicating these checks in the `*RemoteTransactionSavepoint*` functions.

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -820,6 +845,247 @@ CoordinatedRemoteTransactionsAbort(void)
...
+		{
+			continue;
+		}
+
+		FinishRemoteTransactionSavepointRelease(connection, subId);
```
anarazel: Hm, we don't maintain / update >lastUnfailedSubXact here, is that ok?

src/include/distributed/remote_transaction.h
```diff
@@ -61,6 +61,8 @@ typedef struct RemoteTransaction
 	/* failed in current transaction */
 	bool transactionFailed;
 
+	SubTransactionId lastUnfailedSubXact;
```
anarazel: "unfailed" is a bit unwieldy - perhaps just 'alive' or 'successful?

src/backend/distributed/transaction/remote_transaction.c
pykello: Yes, this is ok. We are just interested in the last "SAVEPOINT" command that successfully executed before transaction failure. Whether that savepoint was released or not doesn't matter for us.

The key point to understanding this is that first sub-transaction always gets subId=2 by postgres, and n-th sub-transaction always gets subId=n+1. If our last successful sub-transaction is SAVEPOINT m, then rolling back to this sub-transaction or any sub-transaction with lesser id recovers the transaction from failure. But rolling back to sub-transactions with greater id doesn't recover the transaction because those SAVEPOINTs happened when we already had a failure.

Sure, there might be some of these savepoints that have been already released, but we don't need to worry about it since postgres already catches this before control reaches our callback in transaction_management.c.

So we don't need to keep an exact stack of what is active or released to handle rolling back.

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -820,6 +845,269 @@ CoordinatedRemoteTransactionsAbort(void)
...
+StartRemoteTransactionSavepointRollback(MultiConnection *connection,
+										SubTransactionId subId)
+{
+	StringInfo savepointCommand = makeStringInfo();
+	appendStringInfo(savepointCommand, "ROLLBACK TO SAVEPOINT savepoint_%u", subId);
+
+	if (!SendRemoteCommand(connection, savepointCommand->data))
+	{
+		ReportConnectionError(connection, WARNING);
+		MarkRemoteTransactionFailed(connection, true);
```
anarazel: Is it actually ok to accept interrupts here in general? I'm doubtful? That'd potentially lead to errors being raised again and and again, because the transaction abort code would be called to clean up the error raised here.

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -820,6 +845,269 @@ CoordinatedRemoteTransactionsAbort(void)
...
+FinishRemoteTransactionSavepointRollback(MultiConnection *connection, SubTransactionId
+										 subId)
+{
+	RemoteTransaction *transaction = &connection->remoteTransaction;
+
+	PGresult *result = GetRemoteCommandResult(connection, true);
+	if (!IsResponseOK(result))
+	{
+		ReportResultError(connection, result, WARNING);
+		MarkRemoteTransactionFailed(connection, true);
```
anarazel: See above.

src/test/regress/expected/multi_subtransactions.out
```diff
@@ -0,0 +1,125 @@
...
+SAVEPOINT s1;
+SAVEPOINT s2;
+INSERT INTO artists VALUES (7, NULL);
+ERROR:  null value in column "name" violates not-null constraint
+SAVEPOINT s3;
+ERROR:  current transaction is aborted, commands ignored until end of transaction block
+ROLLBACK TO SAVEPOINT s3;
+ERROR:  no such savepoint
+COMMIT;
+-- Clean-up
```
anarazel: Think it'd be good to have a few replication factor > 1 tests here, especially ones where statements fail on only one of the workers.

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -67,11 +82,21 @@ StartRemoteTransactionBegin(struct MultiConnection *connection)
...
 
+	/* append in-progress savepoints for this transaction */
+	activeSubXacts = ActiveSubXacts();
+	transaction->lastSuccessfulSubXact = TopSubTransactionId;
+	foreach(subIdCell, activeSubXacts)
+	{
+		SubTransactionId subId = lfirst_int(subIdCell);
+		appendStringInfo(beginAndSetDistributedTransactionId,
+						 "SAVEPOINT savepoint_%u;", subId);
+		transaction->lastSuccessfulSubXact = subId;
```
anarazel: Shouldn't this only be sent after we successfully run this?

pykello: Done.
(I was assuming if the whole beginAndSetDistributedTransactionId isn't successful then we don't need lastSuccessfulSubXact at all. Am I missing something?)

src/backend/distributed/transaction/remote_transaction.c
pykello: Based on our internal discussion, I added conditions for this at begin/release but not at rollback.

src/include/distributed/remote_transaction.h
```diff
@@ -106,4 +119,9 @@ extern void CoordinatedRemoteTransactionsPrepare(void);
 extern void CoordinatedRemoteTransactionsCommit(void);
 extern void CoordinatedRemoteTransactionsAbort(void);
 
+/* remote savepoint commands */
+extern void RemoteTransactionsSavepointBegin(SubTransactionId subId);
```
anarazel: I think it'd be better to stay in line with the naming from the functions just above.

src/backend/distributed/transaction/remote_transaction.c
anarazel: Ping!

src/backend/distributed/transaction/remote_transaction.c
anarazel: Maybe I'm missing something, but I don't see any adjustments here?

anarazel: The problem is that this can throw an error, while we're processing transaction abort, interrupting transaction abort. Not good.

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -820,6 +847,285 @@ CoordinatedRemoteTransactionsAbort(void)
...
+													  iter.cur);
+		RemoteTransaction *transaction = &connection->remoteTransaction;
+		if (transaction->transactionFailed)
+		{
+			continue;
+		}
+
+		StartRemoteTransactionSavepointBegin(connection, subId);
+	}
+
```
marcocitus: After rebase, should probably use `WaitForAllConnections` in these functions.

src/backend/distributed/transaction/remote_transaction.c
pykello: done.

src/backend/distributed/transaction/remote_transaction.c
pykello: done.

src/include/distributed/remote_transaction.h
pykello: done

src/backend/distributed/transaction/remote_transaction.c
pykello: done.

src/test/regress/sql/multi_subtransactions.sql
```diff
@@ -0,0 +1,161 @@
...
+-- Don't recover, and commit
+BEGIN;
+SAVEPOINT s1;
+INSERT INTO researchers VALUES (NULL, 10, 'Raymond Smullyan');
+RELEASE SAVEPOINT s1;
+SAVEPOINT s2;
+COMMIT;
+
+SELECT * FROM researchers WHERE lab_id=10;
+
```
anarazel: Could you add a few tests that test "implicit" savepoints via plpgsql exception blocks?

src/test/regress/sql/multi_subtransactions.sql
```diff
@@ -0,0 +1,161 @@
...
+ROLLBACK TO SAVEPOINT s3;
+COMMIT;
+
+-- ===================================================================
+-- Tests for replication factor > 1
+-- ===================================================================
+
+CREATE TABLE researchers (
+	id bigint NOT NULL,
+    lab_id int NOT NULL,
```
anarazel: weird indent.

src/test/regress/sql/multi_subtransactions.sql
pykello: Done. Please take a look.

src/test/regress/expected/multi_subtransactions.out
```diff
@@ -202,6 +202,58 @@ SELECT * FROM researchers WHERE lab_id=10;
  12 |     10 | Edsger Dijkstra
 (1 row)
 
+-- Implicit savepoints via pl/pgsql exceptions
+BEGIN;
+DO $$
+BEGIN
+	INSERT INTO researchers VALUES (15, 10, 'Eric S. Raymond');
+  INSERT INTO researchers VALUES (NULL, 10, 'Raymond Smullyan');
+EXCEPTION
```
anarazel: You're mixing spaces with tabs...

src/test/regress/expected/multi_subtransactions.out
```diff
@@ -202,6 +202,58 @@ SELECT * FROM researchers WHERE lab_id=10;
...
+----+--------+-----------------
+ 12 |     10 | Edsger Dijkstra
+(1 row)
+
+BEGIN;
+DO $$
+BEGIN
+	INSERT INTO researchers VALUES (15, 10, 'Eric S. Raymond');
+  RAISE EXCEPTION plpgsql_error;
+EXCEPTION
```
anarazel: here too.

src/test/regress/expected/multi_subtransactions.out
```diff
@@ -202,6 +202,58 @@ SELECT * FROM researchers WHERE lab_id=10;
...
+----+--------+-----------------
+ 12 |     10 | Edsger Dijkstra
+(1 row)
+
+BEGIN;
+DO $$
+BEGIN
+	INSERT INTO researchers VALUES (15, 10, 'Eric S. Raymond');
+  INSERT INTO researchers VALUES (NULL, 10, 'Raymond Smullyan');
+EXCEPTION
```
anarazel: dito.

src/test/regress/sql/multi_subtransactions.sql
```diff
@@ -107,7 +107,7 @@ COMMIT;
 
 CREATE TABLE researchers (
 	id bigint NOT NULL,
-    lab_id int NOT NULL,
+  lab_id int NOT NULL,
 	name text NOT NULL
 );
```
anarazel: think here too.

src/test/regress/sql/multi_subtransactions.sql
```diff
@@ -156,6 +156,46 @@ COMMIT;
...
+BEGIN;
+DO $$
+BEGIN
+	INSERT INTO researchers VALUES (15, 10, 'Eric S. Raymond');
+  INSERT INTO researchers VALUES (NULL, 10, 'Raymond Smullyan');
+EXCEPTION
+    WHEN not_null_violation THEN
+        RAISE EXCEPTION not_null_violation; -- rethrow it
+END $$;
+COMMIT;
```
anarazel: Can you add a test where we insert something after catching an error?

onderkalaci:
> This change fixes #1539 .

I think this should be #1493 ?

sumedhpathak:
🍾

sumedhpathak:
@pykello Ok to merge then?

PR #1540, opened 2017-08-08
Allow reads from secondaries

lithp:
An item in #1528

Behavior:
- When `citus.use_secondary_nodes = 'never'` behavior is identical.
- When `citus.use_secondary_nodes = 'always'` all modification DML are disallowed.
- When `citus.use_secondary_nodes = 'always'` the planner and executor only see secondary nodes., and use them to handle `SELECT` queries.
- When `citus.use_secondary_nodes = 'always'` all utility commands which call `ActivePrimaryNodeList` error out.

### DIFF
--- a/src/backend/distributed/executor/multi_server_executor.c
+++ b/src/backend/distributed/executor/multi_server_executor.c
@@ -71,7 +71,9 @@ JobExecutorType(MultiPlan *multiPlan)
 												 " queries on the workers.")));
 	}
 
-	workerNodeList = ActivePrimaryNodeList();
+	Assert(multiPlan->operation == CMD_SELECT);
+
+	workerNodeList = ActiveReadableNodeList();
 	workerNodeCount = list_length(workerNodeList);
 	taskCount = list_length(job->taskList);
 	tasksPerNode = taskCount / ((double) workerNodeCount);
--- a/src/backend/distributed/master/worker_node_manager.c
+++ b/src/backend/distributed/master/worker_node_manager.c
@@ -309,6 +309,19 @@ ActivePrimaryNodeCount(void)
 
 
 /*
+ * ActiveReadableNodeCount returns the number of groups with a node we can read from.
+ */
+uint32
+ActiveReadableNodeCount(void)
+{
+	List *workerNodeList = ActiveReadableNodeList();
+	uint32 liveWorkerCount = list_length(workerNodeList);
+
+	return liveWorkerCount;
+}
+
+
+/*
  * ActivePrimaryNodeList returns a list of all the active primary nodes in workerNodeHash
  */
 List *
@@ -318,6 +331,8 @@ ActivePrimaryNodeList(void)
 	WorkerNode *workerNode = NULL;
 	HTAB *workerNodeHash = GetWorkerNodeHash();
 	HASH_SEQ_STATUS status;
+
+	EnsureModificationsCanRun();
 
 	hash_seq_init(&status, workerNodeHash);
 
@@ -329,6 +344,42 @@ ActivePrimaryNodeList(void)
 			memcpy(workerNodeCopy, workerNode, sizeof(WorkerNode));
 			workerNodeList = lappend(workerNodeList, workerNodeCopy);
 		}
+	}
+
+	return workerNodeList;
+}
+
+
+/*
+ * ActiveReadableNodeList returns a list of all nodes in workerNodeHash we can read from.
+ */
+List *
+ActiveReadableNodeList(void)
+{
+	List *workerNodeList = NIL;
+	WorkerNode *workerNode = NULL;
+	HTAB *workerNodeHash = GetWorkerNodeHash();
+	HASH_SEQ_STATUS status;
+
+	hash_seq_init(&status, workerNodeHash);
+
+	while ((workerNode = hash_seq_search(&status)) != NULL)
+	{
+		WorkerNode *workerNodeCopy;
+
+		if (!workerNode->isActive)
+		{
+			continue;
+		}
+
+		if (!WorkerNodeIsReadable(workerNode))
+		{
+			continue;
+		}
+
+		workerNodeCopy = palloc0(sizeof(WorkerNode));
+		memcpy(workerNodeCopy, workerNode, sizeof(WorkerNode));
+		workerNodeList = lappend(workerNodeList, workerNodeCopy);
 	}
 
 	return workerNodeList;
--- a/src/backend/distributed/utils/metadata_cache.c
+++ b/src/backend/distributed/utils/metadata_cache.c
@@ -58,6 +58,9 @@
 #include "utils/typcache.h"
 
 
+/* user configuration */
+int ReadFromSecondaries = USE_SECONDARY_NODES_NEVER;
+
 /*
  * ShardCacheEntry represents an entry in the shardId -> ShardInterval cache.
  * To avoid duplicating data and invalidation logic between this cache and the
@@ -186,6 +189,7 @@ static ShardInterval * TupleToShardInterval(HeapTuple heapTuple,
 static void CachedRelationLookup(const char *relationName, Oid *cachedOid);
 static ShardPlacement * ResolveGroupShardPlacement(
 	GroupShardPlacement *groupShardPlacement, ShardCacheEntry *shardEntry);
+static WorkerNode * LookupNodeForGroup(uint32 groupid);
 
 
 /* exports for SQL callable functions */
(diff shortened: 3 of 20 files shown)


src/backend/distributed/master/worker_node_manager.c
```diff
@@ -332,6 +332,12 @@ ActivePrimaryNodeList(void)
 	HTAB *workerNodeHash = GetWorkerNodeHash();
 	HASH_SEQ_STATUS status;
 
+	if (ReadFromSecondaries == READ_FROM_SECONDARIES_ALWAYS)
+	{
+		ereport(ERROR, (errmsg("writing to nodes is not currently allowed"),
+						errdetail("citus.read_from_secondaries is set to 'always'")));
```
marcocitus: maybe we should just check for `RecoveryInProgress()`

lithp: I think you need both?

For the current use-case `RecoveryInProgress()` is enough, but if you ever want to toggle `read_from_secondaries` on the primary coordinator this should continue to disallow writes?

lithp: I'm starting to think you're right but there's one weird behavior: Say that, if you're on a primary coordinator you can continue to do all the things you could previously do. This means once you flip `read_from_secondaries` all your reads go to different nodes than your writes.

lithp:
@marcocitus right now things are in a pretty halfway state but a few questions have come up (all of the following assumes that `read_from_secondaries` is `always`. As far as I know behavior is identical while `read_from_secondaries` is `never`)
- Currently router-executor `SELECT `queries go to secondaries, as they should.
- Currently router-executor `INSERT` queries also go to secondaries
  - Maybe we should error out if you try to do any kind of modification? If you're not using synchronous replication you'll have weird inconsistencies.
- Currently real-time `SELECT` queries error out, maybe we should support them?

lithp:
@marcocitus An easy answer to avoid many of these questions is to disallow `SET read_from_secondaries TO 'always'` unless we're in recovery mode.

lithp:
After going through the code it seems it would be rather hard to make `SELECT` queries go to secondaries while `INSERT` queries continue to use primaries. I think these two options make the most sense:

- Add a check which prevents `read_from_secondaries` from being set while we're not in recovery mode. - Add checks to the planner which error out if any modification queries happen while `read_from_secondaries` is set.
  - This brings up the question about that to do about metadata changes which don't touch worker nodes

marcocitus:
> Add a check which prevents read_from_secondaries from being set while we're not in recovery mode.

What happens after promotion?

lithp:
> What happens after promotion?

A very good point, and this is also difficult to test. Considering what the other options are.

lithp:
@marcocitus Okay, no longer a WIP, waiting for your feedback. I put the current behavior into the issue description.

src/backend/distributed/master/worker_node_manager.c
```diff
@@ -319,6 +332,8 @@ ActivePrimaryNodeList(void)
 	HTAB *workerNodeHash = GetWorkerNodeHash();
 	HASH_SEQ_STATUS status;
 
+	EnsureCanRunModifications();
```
marcocitus: Just for grammar, `EnsureModificationsCanRun`?

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -480,6 +492,60 @@ ResolveGroupShardPlacement(GroupShardPlacement *groupShardPlacement,
...
+	{
+		ereport(ERROR, (errmsg("the metadata is inconsistent"),
+						errdetail("there is a placement in group %u but "
+								  "there are no nodes in that group", groupId)));
+	}
+
+	switch (ReadFromSecondaries)
+	{
+		case READ_FROM_SECONDARIES_NEVER:
+			ereport(ERROR, (errmsg("node group %u does not have a primary node",
```
marcocitus: missing brackets

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -480,6 +492,60 @@ ResolveGroupShardPlacement(GroupShardPlacement *groupShardPlacement,
...
+								   groupId)));
+		case READ_FROM_SECONDARIES_ALWAYS:
+			ereport(ERROR, (errmsg("node group %u does not have a secondary node",
+								   groupId)));
+		default:
+			ereport(FATAL, (errmsg("this should never happen")));
```
marcocitus: maybe "unrecognised read_from_secondaries value" to help the person who will be adding a new option.

src/test/regress/expected/multi_read_from_secondaries.out
```diff
@@ -0,0 +1,61 @@
...
+---
+ 1
+ 2
+(2 rows)
+
+-- insert into is definitely not allowed
+INSERT INTO dest_table (a, b)
+  SELECT a, b FROM source_table;
+ERROR:  writing to worker nodes is not currently allowed
+DETAIL:  citus.read_from_secondaries is set to 'always'
```
marcocitus: the name of this GUC makes these errors confusing. Maybe `use_secondary_nodes`?

lithp: Good point, renamed it

src/backend/distributed/utils/metadata_cache.c
lithp: Wow, I somehow never noticed we use brackets, we really like verbosity!

src/backend/distributed/executor/multi_server_executor.c
```diff
@@ -71,7 +71,12 @@ JobExecutorType(MultiPlan *multiPlan)
 												 " queries on the workers.")));
 	}
 
-	workerNodeList = ActivePrimaryNodeList();
+	if (multiPlan->operation != CMD_SELECT)
+	{
+		ereport(FATAL, (errmsg("we assume only SELECT queries hit this point")));
```
marcocitus: Maybe "JobExecutorType should only be called for SELECT queries"

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -480,6 +492,68 @@ ResolveGroupShardPlacement(GroupShardPlacement *groupShardPlacement,
...
+{
+	WorkerNode *workerNode = NULL;
+	HASH_SEQ_STATUS status;
+	HTAB *workerNodeHash = GetWorkerNodeHash();
+	bool foundAnyNodes = false;
+
+	hash_seq_init(&status, workerNodeHash);
+
+	while ((workerNode = hash_seq_search(&status)) != NULL)
+	{
```
marcocitus: This feels a bit inefficient, since we do this for every query for every placement. Is there a way to make this faster?

src/backend/distributed/executor/multi_server_executor.c
lithp: I don't think that's true though, I think `JobExecutorType` is always called, I just want to be able to `Assert` that by the time control reaches here we're only playing with `SELECT`.

src/backend/distributed/utils/metadata_cache.c
lithp: This is no slower than the call to `PrimaryNodeForGroup` which was already there. And it is inefficient to walk through the set of all worker nodes but our biggest customer only has 50 so I don't think this will have a large impact. But, there are a few options:
- Build a second WorkerNodeHash which maps from `groupid` to a list of nodes.
- Store resolved shards in the metadata cache and expire the cache whenever `pg_dist_node` changes. This means we only have to pay the cost on the first query, but has the problem from earlier where we error out if any group is missing a worker, even though we might not need the unresolvable placement.
- Push the resolution deeper into the planner / executors, to make sure we only run resolution on the shards we actually need to run it on.
- Somewhere in `ShardPlacementList` we can build a hash from groupid -> worker node, and pass it into `ResolveGroupShardPlacement`. This won't share between queries but we'll only scan the workernodehash once per query, instead of once per shard, and isn't a lot of extra coding.

src/backend/distributed/executor/multi_server_executor.c
marcocitus: Ok, but maybe just use `Assert` to do `Assert`

src/backend/distributed/utils/metadata_cache.c
lithp: Should I look in to one of these? The first one, making a second hash which stores the worker nodes keyed by group, seems most promising.

marcocitus: On the one hand I don't want to overcomplicate this, on the other hand these hash iterations have a tendency to show up in profiling. At some point we were planning to cache the worker list as an array. Is that something we can do easily?

marcocitus: (if there isn't I'm fine merging this change as is, though we should probably do some profiling)

lithp: Opened a PR: https://github.com/citusdata/citus/pull/1547

marcocitus:
(some squashing might be in order)

PR #1542, opened 2017-08-09
Add master_add_secondary_node() UDF

lithp:
Part of #1528

### DIFF
--- a/src/backend/distributed/Makefile
+++ b/src/backend/distributed/Makefile
@@ -11,7 +11,7 @@ EXTVERSIONS = 5.0 5.0-1 5.0-2  \
 	6.0-1 6.0-2 6.0-3 6.0-4 6.0-5 6.0-6 6.0-7 6.0-8 6.0-9 6.0-10 6.0-11 6.0-12 6.0-13 6.0-14 6.0-15 6.0-16 6.0-17 6.0-18 \
 	6.1-1 6.1-2 6.1-3 6.1-4 6.1-5 6.1-6 6.1-7 6.1-8 6.1-9 6.1-10 6.1-11 6.1-12 6.1-13 6.1-14 6.1-15 6.1-16 6.1-17 \
 	6.2-1 6.2-2 6.2-3 6.2-4 \
-	7.0-1 7.0-2 7.0-3 7.0-4 7.0-5 7.0-6 7.0-7 7.0-8 7.0-9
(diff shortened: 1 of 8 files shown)


src/backend/distributed/citus--7.0-9--7.0-10.sql
```diff
@@ -0,0 +1,27 @@
...
+                                          nodecluster name default 'default',
+                                          OUT nodeid integer,
+                                          OUT groupid integer,
+                                          OUT nodename text,
+                                          OUT nodeport integer,
+                                          OUT noderack text,
+                                          OUT hasmetadata boolean,
+                                          OUT isactive bool,
+                                          OUT noderole noderole,
+                                          OUT nodecluster name)
```
marcocitus: I wonder whether we should keep repeating past mistakes (i.e. not just returning a boolean here), but inconsistent signatures feels slightly worse.

lithp: In this case it turned out to be kind of useful, it made testing a little easier but I think you're right both that we should return a boolean, whether the node was added (and didn't already exist) and that the change should be a single PR which changes them all at once.

src/backend/distributed/Makefile
```diff
@@ -11,7 +11,7 @@ EXTVERSIONS = 5.0 5.0-1 5.0-2  \
 	6.0-1 6.0-2 6.0-3 6.0-4 6.0-5 6.0-6 6.0-7 6.0-8 6.0-9 6.0-10 6.0-11 6.0-12 6.0-13 6.0-14 6.0-15 6.0-16 6.0-17 6.0-18 \
 	6.1-1 6.1-2 6.1-3 6.1-4 6.1-5 6.1-6 6.1-7 6.1-8 6.1-9 6.1-10 6.1-11 6.1-12 6.1-13 6.1-14 6.1-15 6.1-16 6.1-17 \
 	6.2-1 6.2-2 6.2-3 6.2-4 \
-	7.0-1 7.0-2 7.0-3 7.0-4 7.0-5 7.0-6 7.0-7 7.0-8 7.0-9
+	7.0-1 7.0-2 7.0-3 7.0-4 7.0-5 7.0-6 7.0-7 7.0-8 7.0-9 7.0-10
```
onderkalaci: ah, you're adding one more UDF, I'm tired of rebasing these 😄

lithp: Heh, sorry, you're right it's pretty annoying. I tried playing with giving versions names like `7.0-nodecluster`, then you have to do a lot less while you rebase, but that leads to a different set of problems.

PR #1543, opened 2017-08-09
Add make target for testing follower clusters

lithp:
Part of #1528, just a framework we can later add more comprehensive tests to.

### DIFF
--- a/src/test/regress/expected/multi_follower_select_statements.out
+++ b/src/test/regress/expected/multi_follower_select_statements.out
@@ -1,0 +1,87 @@
+\c - - - :master_port
+-- do some setup
+SELECT 1 FROM master_add_node('localhost', :worker_1_port);
+ ?column? 
+----------
+        1
+(1 row)
+
+SELECT 1 FROM master_add_node('localhost', :worker_2_port);
+ ?column? 
+----------
+        1
+(1 row)
+
+CREATE TABLE the_table (a int, b int);
+SELECT create_distributed_table('the_table', 'a');
+ create_distributed_table 
+--------------------------
+ 
+(1 row)
+
+INSERT INTO the_table (a, b) VALUES (1, 1);
+INSERT INTO the_table (a, b) VALUES (1, 2);
+-- connect to the follower and check that a simple select query works, the follower
+-- is still in the default cluster and will send queries to the primary nodes
+\c - - - :follower_master_port
+SELECT * FROM the_table;
+ a | b 
+---+---
+ 1 | 1
+ 1 | 2
+(2 rows)
+
+-- now, connect to the follower but tell it to use secondary nodes. There are no
+-- secondary nodes so this should fail.
+-- (this is :follower_master_port but substitution doesn't work here)
+\c "port=9070 dbname=regression options='-c\ citus.use_secondary_nodes=always'"
+SELECT * FROM the_table;
+ERROR:  node group 2 does not have a secondary node
+-- add the secondary nodes and try again, the SELECT statement should work this time
+\c - - - :master_port
+SELECT 1 FROM master_add_node('localhost', :follower_worker_1_port,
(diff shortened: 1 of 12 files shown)


src/test/regress/pg_regress_multi.pl
```diff
@@ -279,13 +300,19 @@ sub revert_replace_postgres
 print $fh "#!/bin/bash\n";
 print $fh "exec psql ";
 print $fh "--variable=master_port=$masterPort ";
+print $fh "--variable=follower_port=$followerCoordPort ";
```
marcocitus: Maybe use follower_master_port.

lithp:
Renamed the variable and also added that changing nodecluster works

src/test/regress/expected/multi_follower_select_statements.out
```diff
@@ -0,0 +1,105 @@
...
+ pg_reload_conf 
+----------------
+ t
+(1 row)
+
+\c - - - -
+-- there are no secondary nodes in this cluster, so this should fail!
+SELECT * FROM the_table;
+ERROR:  the metadata is inconsistent
+DETAIL:  there is a placement in group 2 but there are no nodes in that group
```
marcocitus: In case the user ever gets here, could we move the detail into the error? Details don't show up in many ORMs.

Also, "group 2" -> "node group 2"

src/test/regress/sql/multi_follower_select_statements.sql
```diff
@@ -0,0 +1,69 @@
...
+\c - - - -
+
+-- now that we've added secondaries this should work
+SELECT * FROM the_table;
+
+-- okay, now let's play with nodecluster. If we change the cluster of our follower node
+-- queries should stat failing again, since there are no worker nodes in the new cluster
+
+-- cluster_name is also SU_BACKEND, so we need to reconnect after changing it
+ALTER SYSTEM SET citus.cluster_name TO 'second-cluster';
```
marcocitus: This seems a bit risky since it also affects subsequent tests.

lithp: Yeah, maybe the test needs to be renamed. I imagine this being kind of like `multi_cluster_management` which calls things like `master_add_node` and gets the cluster ready for later tests.

Since this is stuck behind a `check-follower-cluster` do you think future tests will want to play with a follower which isn't configured to be in a different cluster and read from secondaries?

marcocitus: Maybe make a proper follower_setup test then that also sets the GUC on the workers (for MX).

src/test/regress/expected/multi_follower_select_statements.out
lithp: Opened a PR for it, I'll rebase and update this test after it's merged: https://github.com/citusdata/citus/pull/1546

lithp:
Alright, added a task-tracker test and it also works on PG9.6

src/test/regress/sql/multi_follower_task_tracker.sql
```diff
@@ -0,0 +1,22 @@
...
+INSERT INTO tab (a, b) VALUES (1, 1);
+INSERT INTO tab (a, b) VALUES (1, 2);
+
+\c - - - :follower_master_port
+
+SET citus.task_executor_type TO 'real-time';
+SELECT * FROM tab;
+SET citus.task_executor_type TO 'task-tracker';
+SELECT * FROM tab;
+
```
marcocitus: It'd be nice to exercise re-partitioning. I'd expect that to fail.

lithp:
Something like grouping by a non-partition column? Won't that go through
the task tracker?

marcocitus:
joins by non-partition-column

marcocitus:
Erroring out on task-tracker queries makes sense. 

I added `check-follower-cluster` to `make check` and changed the port numbers for followers to avoid them being in use on travis.

PR #1544, opened 2017-08-09
Acquire proper locks for partitioned table operations

byucesoy:
Part of #1459

We need to acquire some locks if we are dealing with partitioned tables
to prevent deadlock on worker nodes. With this change we start to acquire;

- shard resource lock on parent table's colocated shard before acquiring lock on partition's shard.
- relation lock on all partitions of parent table on query execution.

### DIFF
--- a/src/backend/distributed/executor/insert_select_executor.c
+++ b/src/backend/distributed/executor/insert_select_executor.c
@@ -14,8 +14,10 @@
 #include "distributed/insert_select_planner.h"
 #include "distributed/multi_copy.h"
 #include "distributed/multi_executor.h"
+#include "distributed/multi_partitioning_utils.h"
 #include "distributed/multi_physical_planner.h"
 #include "distributed/multi_planner.h"
+#include "distributed/resource_lock.h"
 #include "distributed/transaction_management.h"
 #include "executor/executor.h"
 #include "nodes/execnodes.h"
@@ -59,6 +61,16 @@ CoordinatorInsertSelectExecScan(CustomScanState *node)
 		Oid targetRelationId = multiPlan->targetRelationId;
 
 		ereport(DEBUG1, (errmsg("Collecting INSERT ... SELECT results on coordinator")));
+
+		/*
+		 * If we are dealing with partitioned table, we also need to lock its
+		 * partitions. Here we only lock targetRelation, we acquire necessary
+		 * locks on selected tables during execution of those select queries.
+		 */
+		if (PartitionedTable(targetRelationId))
+		{
+			LockPartitionRelations(targetRelationId, RowExclusiveLock);
+		}
 
 		ExecuteSelectIntoRelation(targetRelationId, insertTargetList, selectQuery,
 								  executorState);
--- a/src/backend/distributed/executor/multi_router_executor.c
+++ b/src/backend/distributed/executor/multi_router_executor.c
@@ -34,6 +34,7 @@
 #include "distributed/master_metadata_utility.h"
 #include "distributed/metadata_cache.h"
 #include "distributed/multi_executor.h"
+#include "distributed/multi_partitioning_utils.h"
 #include "distributed/multi_physical_planner.h"
 #include "distributed/multi_planner.h"
 #include "distributed/multi_router_executor.h"
@@ -98,6 +99,7 @@ static bool StoreQueryResult(CitusScanState *scanState, MultiConnection *connect
 							 bool failOnError, int64 *rows);
 static bool ConsumeQueryResult(MultiConnection *connection, bool failOnError,
 							   int64 *rows);
+static LOCKMODE LockModeForModifyTask(Task *task);
 
 
 /*
@@ -434,6 +436,12 @@ CitusModifyBeginScan(CustomScanState *node, EState *estate, int eflags)
 	/* prevent concurrent placement changes */
 	AcquireMetadataLocks(taskList);
 
+	/*
+	 * We are taking locks on partitions of partitioned tables. These locks are
+	 * necessary for locking tables that appear in the SELECT part of the query.
+	 */
+	LockPartitionsInRelationList(multiPlan->relationIdList, AccessShareLock);
+
 	/* modify tasks are always assigned using first-replica policy */
 	workerJob->taskList = FirstReplicaAssignTaskList(taskList);
 }
@@ -518,6 +526,9 @@ RouterSelectExecScan(CustomScanState *node)
 		MultiPlan *multiPlan = scanState->multiPlan;
 		Job *workerJob = multiPlan->workerJob;
 		List *taskList = workerJob->taskList;
+
+		/* we are taking locks on partitions of partitioned tables */
+		LockPartitionsInRelationList(multiPlan->relationIdList, AccessShareLock);
 
 		if (list_length(taskList) > 0)
 		{
@@ -682,6 +693,9 @@ ExecuteSingleModifyTask(CitusScanState *scanState, Task *task, bool expectResult
 	char *queryString = task->queryString;
 	bool taskRequiresTwoPhaseCommit = (task->replicationModel == REPLICATION_MODEL_2PC);
 
+	ShardInterval *shardInterval = LoadShardInterval(task->anchorShardId);
+	Oid relationId = shardInterval->relationId;
+
 	/*
 	 * Modifications for reference tables are always done using 2PC. First
 	 * ensure that distributed transaction is started. Then force the
@@ -711,6 +725,17 @@ ExecuteSingleModifyTask(CitusScanState *scanState, Task *task, bool expectResult
 	 * table) and start a transaction (when in a transaction).
 	 */
 	connectionList = GetModifyConnections(task, taskRequiresTwoPhaseCommit);
+
+	/*
+	 * If we are dealing with a partitioned table, we also need to lock its
+	 * partitions.
+	 */
+	if (PartitionedTable(relationId))
+	{
+		LOCKMODE lockMode = LockModeForModifyTask(task);
+
+		LockPartitionRelations(relationId, lockMode);
+	}
 
 	/* prevent replicas of the same shard from diverging */
 	AcquireExecutorShardLock(task, operation);
@@ -947,6 +972,7 @@ ExecuteModifyTasks(List *taskList, bool expectResults, ParamListInfo paramListIn
 	int64 totalAffectedTupleCount = 0;
 	ListCell *taskCell = NULL;
 	Task *firstTask = NULL;
+	ShardInterval *firstShardInterval = NULL;
 	int connectionFlags = 0;
 	List *affectedTupleCountList = NIL;
 	HTAB *shardConnectionHash = NULL;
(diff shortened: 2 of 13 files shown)


byucesoy:
This PR is ready for review. @marcoslot would you like to have a look?

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -963,6 +974,19 @@ ExecuteModifyTasks(List *taskList, bool expectResults, ParamListInfo paramListIn
...
+	{
+		Task *task = (Task *) lfirst(taskCell);
+		ShardInterval *shardInterval = LoadShardInterval(task->anchorShardId);
+		Oid relationId = shardInterval->relationId;
+
+		/*
+		 * If we are dealing with partitioned table, we also need to lock its
+		 * partitions.
+		 */
+		LockPartitionRelationsIfPartitioned(relationId, RowExclusiveLock);
```
marcocitus: I think it would make more sense if the `if (PartitionTable(...)` statement was around the loop and the function was called `LockPartitionRelations`. The same goes for `LockParentRelationShardResources`. Now we do a bunch of work for now reason when we're dealing with a non-partitioned table.

src/backend/distributed/utils/resource_lock.c
```diff
@@ -316,3 +318,84 @@ LockRelationShardResources(List *relationShardList, LOCKMODE lockMode)
...
+ * colocated shard of parent tables of given shards' relation if given shard
+ * belongs to a partition.
+ */
+void
+LockParentRelationShardResourcesIfPartition(List *relationShardList, LOCKMODE lockMode)
+{
+	ListCell *relationShardCell = NULL;
+
+	/* lock shards in a consistent order to prevent deadlock */
+	relationShardList = SortList(relationShardList, CompareRelationShards);
```
marcocitus: does this ordering still apply after we do child shard ID -> parent shard ID conversion?

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -710,6 +715,12 @@ ExecuteSingleModifyTask(CitusScanState *scanState, Task *task, bool expectResult
 										  taskRequiresTwoPhaseCommit,
 										  startedInTransaction);
 
+	/*
+	 * If we are dealing with partitioned table, we also need to lock its
+	 * partitions.
+	 */
+	LockPartitionRelationsIfPartitioned(relationId, RowExclusiveLock);
```
marcocitus: Question: does postgres take AccessShareLock on the children when querying the parent?

byucesoy: For SELECTs yes it does. For INSERT/DELETE/UPDATE it only takes RowExclusiveLock.

marcocitus: Should we be taking it for SELECTs as well then?

src/backend/distributed/utils/resource_lock.c
byucesoy: Hmm, not necessarily. I think we should first extract the co-located shard ids on parent tables, then sort them and acquire lock on them.

src/backend/distributed/executor/multi_router_executor.c
byucesoy: But how can we know whether an element in the list belongs to a partitioned table without doing all that work:

onderkalaci: Is that guaranteed that all tasks belong to the same relation (i.e., anchorShardId's are always  belong to the same relation)?

I think even if that is true for today, having the loop seems better in the long term in case we'd have tasks that belong to different tables?

src/backend/distributed/executor/multi_router_executor.c
onderkalaci: I also think that we should take the locks. However, doing it in this PR would make things a lot complicated. We'd need to add `relationShardList` to `Job` and acquire the locks on the executor (or smth like that). 

I think we should open an issue and acquire the locks for `SELECT`s properly on the coordinator node. (Now that we acquire `AccessShareLock` just because deparsing acquires it.) @marcocitus What do you think?

src/backend/distributed/executor/multi_router_executor.c
marcocitus: It seems safe to assume all tasks modify the same relation.

src/backend/distributed/executor/multi_router_executor.c
onderkalaci: Or maybe we on `CreateDistributedPlan()` we add `relationList` to `MultiPlan` using the `relationRestrictions`. That seems simpler?

src/backend/distributed/executor/multi_router_executor.c
onderkalaci: Cool then let's simplify the loop here @byucesoy

src/backend/distributed/executor/multi_router_executor.c
marcocitus: We could extract the list of relations from the job query.

byucesoy: Ok, great :+1:

src/backend/distributed/utils/resource_lock.c
```diff
@@ -316,3 +318,121 @@ LockRelationShardResources(List *relationShardList, LOCKMODE lockMode)
...
+
+	if (parentShardIntervalList == NIL)
+	{
+		return;
+	}
+
+	/* lock shards in a consistent order to prevent deadlock */
+	parentShardIntervalList = SortList(parentShardIntervalList,
+									   CompareShardIntervalsById);
+	foreach(parentShardIntervalCell, parentShardIntervalList)
```
marcocitus: Could probably insert a new line here.

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -311,7 +318,8 @@ AcquireExecutorMultiShardLocks(List *taskList)
 			 * and therefore prevents other modifications from running
 			 * concurrently.
 			 */
-
+			LockParentRelationShardResourcesIfPartition(task->relationShardList,
```
marcocitus: I suppose we should never be doing anything here since we don't support RF >1, right?

onderkalaci:
> Small note: We should consider removing the multi-shard resource locks now that we have deadlock detection.

I think we should remove them all in the next release. Does that work?

src/test/regress/expected/multi_partitioning.out
```diff
@@ -923,15 +923,340 @@ ERROR:  no partition of relation "multi_column_partitioning_1660068" found for r
...
+-- see the locks on parent table
+SELECT
+    locktype,
+    mode
+FROM
+    pg_locks AS l JOIN pg_dist_shard AS s
+ON
+    l.objid = s.shardid
+WHERE
+    logicalrelid = 'partitioning_locks'::regclass
```
onderkalaci: It'd be nice to see the locks on the `partitioning_locks_2009 ` and `partitioning_locks_2010`.


Could you add the `logicalrelid`::regclass to the output and add `WHERE logicalrelid IN (`..` , '...')` to see those? And remove the query below?

src/test/regress/expected/multi_partitioning.out
```diff
@@ -923,15 +923,340 @@ ERROR:  no partition of relation "multi_column_partitioning_1660068" found for r
...
+ORDER BY
+    1, 2;
+ locktype |           mode           
+----------+--------------------------
+ advisory | ShareUpdateExclusiveLock
```
onderkalaci: Same here, using a single query could simplify reading the output.

src/backend/distributed/utils/resource_lock.c
```diff
@@ -316,3 +318,69 @@ LockRelationShardResources(List *relationShardList, LOCKMODE lockMode)
...
+		if (PartitionedTable(relationId))
+		{
+			LockPartitionRelations(relationId, lockmode);
+		}
+	}
+}
+
+
+/*
+ * LockPartitionRelations acquires relation lock on all partitions of given
```
byucesoy: I added the comment but do we really need Assert? In the PartitionList function we call, we already error out if given table is not a partitioned table.

onderkalaci: I didin't realize that. Then, we don't need the assert

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -296,6 +297,11 @@ AcquireExecutorMultiShardLocks(List *taskList)
 			lockMode = ExclusiveLock;
 		}
 
+		/*
+		 * If we are dealing with a partition we are also taking locks on parent table
+		 * to prevent deadlocks on concurrent operations on a partition and its parent
```
onderkalaci: very minor: missing period

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -434,6 +440,13 @@ CitusModifyBeginScan(CustomScanState *node, EState *estate, int eflags)
 	/* prevent concurrent placement changes */
 	AcquireMetadataLocks(taskList);
 
+	/*
+	 * We are taking locks on partitions of partitioned tables. These locks are
+	 * necessary for locking selected tables of router plannable INSERT/SELECT
```
onderkalaci: A suggestion, feel free to ignore:`These locks are necessary for locking tables that appear in the SELECT part of the query`

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -712,6 +731,15 @@ ExecuteSingleModifyTask(CitusScanState *scanState, Task *task, bool expectResult
 	 */
 	connectionList = GetModifyConnections(task, taskRequiresTwoPhaseCommit);
 
+	/*
+	 * If we are dealing with partitioned table, we also need to lock its
```
onderkalaci: `with partitioned` -> `with a partitioned `

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -958,13 +987,23 @@ ExecuteModifyTasks(List *taskList, bool expectResults, ParamListInfo paramListIn
 		return 0;
 	}
 
+	/*
+	 * In multi shard modification, we expect that all tasks operates on
+	 * same relation, so it is enough to acquire a lock on first relation's
```
onderkalaci: I think `on the first relation's partitions.` is not accurate? It should be `on the first task's anchor relation` or so?

`same relation` -> `the same relation`

src/backend/distributed/utils/resource_lock.c
```diff
@@ -316,3 +318,69 @@ LockRelationShardResources(List *relationShardList, LOCKMODE lockMode)
 		}
 	}
 }
+
+
+/*
+ * LockParentShardResourceIfPartition checks whether given shard belongs to a
```
onderkalaci: Pretty minor comments: 
`whether given shard` -> `whether the given shard `

`If it is, ` -> `If it does` ?

src/test/regress/expected/multi_partitioning.out
```diff
@@ -923,15 +923,326 @@ ERROR:  no partition of relation "multi_column_partitioning_1660068" found for r
...
+COMMIT;
+-- test shard resource locks with master_modify_multiple_shards
+BEGIN;
+SELECT master_modify_multiple_shards('UPDATE partitioning_locks_2009 SET time = ''2009-03-01''');
+ master_modify_multiple_shards 
+-------------------------------
+                             0
+(1 row)
+
+-- see the locks on parent table
```
onderkalaci: comment is obsolete

src/test/regress/expected/multi_partitioning.out
```diff
@@ -923,15 +923,324 @@ ERROR:  no partition of relation "multi_column_partitioning_1660068" found for r
...
+COMMIT;
+-- test locks on task-tracker SELECT
+SET citus.task_executor_type TO 'task-tracker';
+BEGIN;
+SELECT * FROM partitioning_locks AS pl1 JOIN partitioning_locks AS pl2 ON pl1.id = pl2.ref_id ORDER BY 1, 2;
+ id | ref_id | time | id | ref_id | time 
+----+--------+------+----+--------+------
+(0 rows)
+
+SELECT relation::regclass, locktype, mode FROM pg_locks WHERE relation::regclass::text LIKE 'partitioning_locks%' ORDER BY 1, 2, 3;
```
onderkalaci: I'd be a bit more comfortable if you do `AND pid = pg_backend_pid()` to the `where` clause on all such queries.

src/backend/distributed/executor/insert_select_executor.c
```diff
@@ -60,6 +62,16 @@ CoordinatorInsertSelectExecScan(CustomScanState *node)
 
 		ereport(DEBUG1, (errmsg("Collecting INSERT ... SELECT results on coordinator")));
 
+		/*
+		 * If we are dealing with partitioned table, we also need to lock its
```
onderkalaci: very minor `with partitioned` -> `with a partitioned`
`acquire necessary` -> `acquire the necessary`

PR #1547, opened 2017-08-10
Replace WorkerNodeHash with WorkerNodeList

lithp:
Here's a PR which replaces the hash with a list, which should be a fair bit faster to iterate over.

This slows down `FindWorkerNode`, which was previously able to search the hash by key, but speeds up literally every other access to the cached worker node set.

The most notable caller of `FindWorkerNode` is `ActivePlacementList` (called by `GreedyAssignTaskList`) which will now iterate over the entire list of `WorkerNode`s for every shard placement.

I'm not sure what the best answer is for all these tests which have changed. Probably some of them can have `ORDER BY` added to them but that won't make this PR any smaller, just future ones, and I'm not sure how many times we'll mess with something this deep.

### DIFF
--- a/src/backend/distributed/master/worker_node_manager.c
+++ b/src/backend/distributed/master/worker_node_manager.c
@@ -274,24 +274,20 @@ ClientHostAddress(StringInfo clientHostStringInfo)
 static WorkerNode *
 WorkerGetNodeWithName(const char *hostname)
 {
-	WorkerNode *workerNode = NULL;
-	HASH_SEQ_STATUS status;
-	HTAB *workerNodeHash = GetWorkerNodeHash();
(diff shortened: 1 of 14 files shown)


lithp:
Before I go fix the tests wonder if you think this is the right path to go down.

marcocitus:
This may be relevant if the hash iteration shows up in profiling, but we can defer it for now.

I think we should use an array rather than a list.

marcocitus:
I kind of regret that we haven't been more careful about groupid assignment. It would be nice if we were able to use a groupid-indexed array.

lithp:
Oh, now that you mention the array I have devaju too :sweat_smile:

PR #1549, opened 2017-08-10
These tests often deadlock in travis, run them in series

lithp:


marcocitus:
Can you give a sample output? This definitely doesn't seem like the right fix :).

jasonmp85:
What's wrong with this as a fix? I've seen bad interactions between these two tests in Travis as well. I guess we could work to make them work in parallel, but this fixes the problem for now (important since we're about to branch).

marcocitus:
These tests are definitely not supposed to deadlock. The right fix is to make sure there is no deadlock.

https://github.com/citusdata/citus/pull/1503 got merged after this PR was opened, so it's possible that this is no longer an issue.

I'd like to see a sample output generated from the current master branch before we do anything with this PR.

marcocitus:
Haven't seen this happen, I think it was fixed by #1503.

PR #1559, opened 2017-08-14
Add version check to the maintenance daemon

### LINKED ISSUE #1557: Distributed deadlock detection crashes when upgrading from old version
When upgrading from an older version of Citus to 7.0, then the first user-issued command will activate the maintenance daemon, which as part of distributed deadlock detection calls `ReadWorkerNodes`, which then tries to read a non-existent column from `pg_dist_node`, which causes a segmentation fault.

The maintenance daemon should not run if the catalog is behind.

onderkalaci:
Fixes #1557 

We should prevent running the daemon if there
is a major version change. Otherwise, the daemon
may access to obsolete metadata catalog tables.

### DIFF
--- a/src/backend/distributed/utils/maintenanced.c
+++ b/src/backend/distributed/utils/maintenanced.c
@@ -266,7 +266,16 @@ CitusMaintenanceDaemonMain(Datum main_arg)
 		if (DistributedDeadlockDetectionTimeoutFactor != -1.0)
 		{
 			StartTransactionCommand();
-			foundDeadlock = CheckForDistributedDeadlocks();
+
+			/*
+			 * We don't want to run the deadlock checks if there exists
+			 * any version mistmatch.
+			 */
+			if (CheckCitusVersion(DEBUG1))
+			{
+				foundDeadlock = CheckForDistributedDeadlocks();
+			}
+
 			CommitTransactionCommand();
 
 			/*


src/backend/distributed/utils/maintenanced.c
```diff
@@ -246,6 +246,17 @@ CitusMaintenanceDaemonMain(Datum main_arg)
 	/* make worker recognizable in pg_stat_activity */
 	pgstat_report_appname("Citus Maintenance Daemon");
 
+	/* before doing the actual jobs, check any version mismatches */
+	StartTransactionCommand();
+
+	if (!CheckCitusVersion(LOG))
+	{
+		/* return code of 1 requests worker restart */
+		proc_exit(1);
```
marcocitus: This'll spam the log a bit. We could just keep spinning in the main loop, but skip deadlock detection until the Citus version matches.

onderkalaci: > This'll spam the log a bit

Hmm, that makes sense. But still I'd prefer to do the check before the loop starts. I'll think more and try to come up with a better solution.

marcocitus: An alternative would be to not start the daemon unless `CheckCitusVersion` by splitting up `StartupCitusBackend()`.

src/backend/distributed/utils/maintenanced.c
```diff
@@ -266,7 +267,16 @@ CitusMaintenanceDaemonMain(Datum main_arg)
 		if (DistributedDeadlockDetectionTimeoutFactor != -1.0)
 		{
 			StartTransactionCommand();
-			foundDeadlock = CheckForDistributedDeadlocks();
+
+			/*
+			 * We don't want to run the deadlock checks if there exists
+			 * any version mistmatch.
+			 */
+			if (CheckCitusVersion(reportLevel))
```
marcocitus: DEBUG1 seems fine. We really don't care about running deadlock detection when the version is behind, since there's no way the user can start a query (hopefully).

marcocitus: I'd have a sight preference for just putting the (constant) log level in here.

PR #1560, opened 2017-08-15
Lock on pg_dist_node while dropping extension

### LINKED ISSUE #1558: Deadlock while trying to drop extension
Ran into a deadlock between the maintenance daemon and `DROP EXTENSION`.

```
# DROP EXTENSION citus CASCADE;
ERROR:  deadlock detected
DETAIL:  Process 9258 waits for AccessExclusiveLock on relation 164112 of database 13322; blocked by process 9259.
Process 9259 waits for AccessShareLock on relation 164148 of database 13322; blocked by process 9258.
HINT:  See server log for query details.
```

Process 9258 is trying to drop the extension.
Process 9259 is the maintenance daemon.
Table 164112 is `pg_dist_node`
Table 164148 is `pg_dist_local_group`

velioglu:
Fixes #1558 

This PR fixes the deadlock between maintenance daemon and drop extension by getting a lock on pg_dist_node.

### DIFF
--- a/src/backend/distributed/executor/multi_utility.c
+++ b/src/backend/distributed/executor/multi_utility.c
@@ -617,12 +617,26 @@ IsCitusDropExtensionStmt(DropStmt *dropStatement)
 	{
 		foreach(objectListCell, objectList)
 		{
-			List *objectNames = (List *) lfirst(objectListCell);
-			ListCell *objectName = NULL;
+			Node *objectNode = (Node *) lfirst(objectListCell);
 
-			foreach(objectName, objectNames)
+			if (IsA(objectNode, List))
 			{
-				Value *extensionName = (Value *) lfirst(objectName);
+				List *objectNames = (List *) objectNode;
+				ListCell *objectName = NULL;
+
+				foreach(objectName, objectNames)
+				{
+					Value *extensionName = (Value *) lfirst(objectName);
+
+					if (strcmp(strVal(extensionName), "citus") == 0)
+					{
+						return true;
+					}
+				}
+			}
+			else
+			{
+				Value *extensionName = (Value *) objectNode;
 
 				if (strcmp(strVal(extensionName), "citus") == 0)
 				{
--- a/src/backend/distributed/executor/multi_utility.c
+++ b/src/backend/distributed/executor/multi_utility.c
@@ -340,23 +340,19 @@ multi_ProcessUtility(PlannedStmt *pstmt,
 	if (IsA(parsetree, DropStmt))
 	{
 		DropStmt *dropStatement = (DropStmt *) parsetree;
+		bool dropCitus = IsCitusDropExtensionStmt(dropStatement);
 
-		if (dropStatement->removeType == OBJECT_EXTENSION)
+		/*
+		 * We need to take a lock on pg_dist_node while dropping the
+		 * citus extension. Otherwise, deadlock between dropping
+		 * extension and deadlock detection daemon may occur. We take
+		 * the lock here because standard process utility does not take
+		 * locks on dependent tables of an extension in a deterministic
+		 * manner.
+		 */
+		if (dropCitus)
 		{
-			bool dropCitus = IsCitusDropExtensionStmt(dropStatement);
-
-			/*
-			 * We need to take a lock on pg_dist_node while dropping the
-			 * citus extension. Otherwise, deadlock between dropping
-			 * extension and deadlock detection daemon may occur. We take
-			 * the lock here because standard process utility does not take
-			 * locks on dependent tables of an extension in a deterministic
-			 * manner.
-			 */
-			if (dropCitus)
-			{
-				LockRelationOid(DistNodeRelationId(), AccessExclusiveLock);
-			}
+			LockRelationOid(DistNodeRelationId(), AccessExclusiveLock);
 		}
 	}
 
@@ -617,18 +613,21 @@ IsCitusDropExtensionStmt(DropStmt *dropStatement)
 	List *objectList = dropStatement->objects;
 	ListCell *objectListCell = NULL;
 
-	foreach(objectListCell, objectList)
+	if (dropStatement->removeType == OBJECT_EXTENSION)
 	{
-		List *objectNames = (List *) lfirst(objectListCell);
-		ListCell *objectName = NULL;
-
-		foreach(objectName, objectNames)
+		foreach(objectListCell, objectList)
 		{
-			Value *extensionName = (Value *) lfirst(objectName);
+			List *objectNames = (List *) lfirst(objectListCell);
+			ListCell *objectName = NULL;
 
-			if (strcmp(strVal(extensionName), "citus") == 0)
+			foreach(objectName, objectNames)
 			{
-				return true;
+				Value *extensionName = (Value *) lfirst(objectName);
+
+				if (strcmp(strVal(extensionName), "citus") == 0)
+				{
+					return true;
+				}
 			}
 		}
 	}
--- a/src/backend/distributed/executor/multi_utility.c
+++ b/src/backend/distributed/executor/multi_utility.c
@@ -336,6 +336,17 @@ multi_ProcessUtility(PlannedStmt *pstmt,
 		ErrorIfUnsupportedTruncateStmt((TruncateStmt *) parsetree);
 	}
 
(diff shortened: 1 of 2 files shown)


src/backend/distributed/executor/multi_utility.c
```diff
@@ -336,6 +336,17 @@ multi_ProcessUtility(PlannedStmt *pstmt,
...
 	}
 
+	/* prevent any operation on pg_dist_node while dropping the extension */
+	if (IsA(parsetree, DropStmt))
+	{
+		DropStmt *dropStatement = (DropStmt *) parsetree;
+
+		if (dropStatement->removeType == OBJECT_EXTENSION)
+		{
+			LockRelationOid(DistNodeRelationId(), AccessExclusiveLock);
```
marcocitus: Should we also check that the extension being dropped is Citus?

src/backend/distributed/executor/multi_utility.c
```diff
@@ -336,6 +336,17 @@ multi_ProcessUtility(PlannedStmt *pstmt,
 		ErrorIfUnsupportedTruncateStmt((TruncateStmt *) parsetree);
 	}
 
+	/* prevent any operation on pg_dist_node while dropping the extension */
```
marcocitus: This requires a bit more elaboration: Deadlock detection starts by taking an `AccessShareLock` on `pg_dist_node` followed by `pg_dist_local_group`, but `DROP EXTENSION` may take `AccessExclusiveLock` on those table in a different order, which could lead to unexpected deadlocks.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -336,6 +337,29 @@ multi_ProcessUtility(PlannedStmt *pstmt,
 		ErrorIfUnsupportedTruncateStmt((TruncateStmt *) parsetree);
 	}
 
+	if (IsA(parsetree, DropStmt))
+	{
+		DropStmt *dropStatement = (DropStmt *) parsetree;
+
+		if (dropStatement->removeType == OBJECT_EXTENSION)
```
marcocitus: maybe move this condition into `IsCitusDropExtensionStmt` as well

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -2348,10 +2348,16 @@ GetWorkerNodeHash(void)
...
 	 * Simulate a SELECT from pg_dist_node, ensure pg_dist_node doesn't change while our
-	 * caller is using WorkerNodeHash.
+	 * caller is using GetWorkerNodeHash.
 	 */
 	LockRelationOid(DistNodeRelationId(), AccessShareLock);
 
+	/* make sure that citus extension still exists */
+	if (!CitusHasBeenLoaded())
+	{
+		ereport(ERROR, (errmsg("citus extension has been dropped")));
```
marcocitus: For this PR, maybe omit this change. I think we need something more structural.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -336,6 +337,25 @@ multi_ProcessUtility(PlannedStmt *pstmt,
...
+		 * We need to take a lock on pg_dist_node while dropping the
+		 * citus extension. Otherwise, deadlock between dropping
+		 * extension and deadlock detection daemon may occur. We take
+		 * the lock here because standard process utility does not take
+		 * locks on dependent tables of an extension in a deterministic
+		 * manner.
+		 */
+		if (dropCitus)
+		{
+			LockRelationOid(DistNodeRelationId(), AccessExclusiveLock);
```
marcocitus: This should happen before we  call postgres' utility hook, not after.

velioglu: As I understood, we call if after the hook calls only if the statement is a transaction statement or the Citus has not been loaded yet (which seems acceptable).

marcocitus: Sorry! You're right.

marcocitus:
It's probably better to go through SQL_DROP trigger, but this seems ok for now.

anarazel:
Hm, wouldn't it be better to acquire locks in the deadlock code conditionally? And just return false if only one of the locks could be acquired?

onderkalaci:
> Hm, wouldn't it be better to acquire locks in the deadlock code conditionally? And just return false if only one of the locks could be acquired?

We'll do that in the daemon with a follow-up PR. The plan is:

* On `DROP DATABASE`, remove the database from `dbHash` on the `MaintenanceDaemonControl`
* On `DROP DATABASE`  at the point `if (!myDbData)` just exit with `0` instead of erroring out

* To check the existence of the citus extension on the daemon, conditionally access to `pg_dist_node`. If not, just skip the deadlock detection and wait until the next iteration.

Does that make sense to you?

src/backend/distributed/executor/multi_utility.c
```diff
@@ -336,6 +337,25 @@ multi_ProcessUtility(PlannedStmt *pstmt,
...
+	if (IsA(parsetree, DropStmt))
+	{
+		DropStmt *dropStatement = (DropStmt *) parsetree;
+		bool dropCitus = IsCitusDropExtensionStmt(dropStatement);
+
+		/*
+		 * We need to take a lock on pg_dist_node while dropping the
+		 * citus extension. Otherwise, deadlock between dropping
+		 * extension and deadlock detection daemon may occur. We take
+		 * the lock here because standard process utility does not take
```
onderkalaci: I'd mention a bit more explicitly that `pg_dist_node` is already going to be dropped before starting the sentence `We take ...`

marcocitus:
It probably makes sense to use `RangeVarGetRelIdExtended` to simultaneously check whether `pg_dist_node` exists and get an `AccessShareLock` on it.

anarazel:
On 2017-08-16 00:43:45 -0700, Önder Kalacı wrote:
> > Hm, wouldn't it be better to acquire locks in the deadlock code conditionally? And just return false if only one of the locks could be acquired?
> 
> We'll do that in the daemon with a follow-up PR. The plan is:

Why do we need the locking hacks from this PR then?


> * On `DROP DATABASE`, remove the database from `dbHash` on the `MaintenanceDaemonControl`

Hm.

> * On `DROP DATABASE`  at the point `if (!myDbData)` just exit with `0` instead of erroring out

That seems ok.

Greetings,

Andres Freund

onderkalaci:
> Why do we need the locking hacks from this PR then?

To my understanding,

Deadlock detection acquires locks on `pg_dist_node` and `pg_dist_local_group`. If we take the `AccessExclusiveLock` on `pg_dist_node` here, we'd only need to conditionally acquire lock on `pg_dist_node` on the daemon. However, if we don't take the lock here, we'd need to conditionally acquire both on `pg_dist_node` and `pg_dist_local_group`. And, as we access more metadata tables, we'd need to conditionally acquire locks for those as well.

anarazel:
On 2017-08-16 08:25:28 +0000, Önder Kalacı wrote:
> > Why do we need the locking hacks from this PR then?
> 
> To my understanding,
> 
> Deadlock detection acquires locks on `pg_dist_node` and `pg_dist_local_group`. If we take the `AccessExclusiveLock` on `pg_dist_node` here, we'd only need to conditionally acquire lock on `pg_dist_node` on the daemon. However, if we don't take the lock here, we'd need to conditionally acquire both on `pg_dist_node` and `pg_dist_local_group`. And, as we access more metadata tables, we'd need to conditionally acquire locks for those as well.

Acquiring the necessarily locks conditionally upfront in deadlock
detection seems preferrable to me to the proposed hack. But I'm ok with
doing it differently.

marcocitus:
In the auto-recovery PR I added a function to check whether a table exists and obtain an AccessShareLock on a non-blocking manner. It's not very pretty because `RangeVarGetRelidExtended` throws an error if the lock cannot be obtained:
https://github.com/citusdata/citus/pull/1574/files#diff-165406aafb69addc03fd1397f9545a6aR388

sumedhpathak:
@velioglu ping on this one?

anarazel:
> @velioglu ping on this one?

I think it's superseded by the approach in #1584 that Marco implemented based on my "hackety hack" robust approach ;)

metdos:
superseded by #1584

PR #1569, opened 2017-08-15
Initialize backend data on startup

onderkalaci:
We see some random problems on the distributed
deadlock detection. The problem may arise
due to non-initialized memory.

Any comments @marcocitus ?


marcocitus:
isn't this already done by https://github.com/citusdata/citus/pull/1569/files#diff-f70dc55fa89802a35653ac19718a228eR340 ?

onderkalaci:
> isn't this already done by https://github.com/citusdata/citus/pull/1569/files#diff-f70dc55fa89802a35653ac19718a228eR340 ?

Ops, that's true.

PR #1580, opened 2017-08-16
Add error detail if query contains both local and distributed tables

velioglu:
Add error detail if query has both local and distributed tables, since Citus supports `INSERT INTO dist_table SELECT ... FROM local_table` queries right now.


src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -2864,7 +2864,10 @@ NeedsDistributedPlanning(Query *queryTree)
 	if (hasLocalRelation && hasDistributedRelation)
 	{
 		ereport(ERROR, (errmsg("cannot plan queries which include both local and "
-							   "distributed relations")));
+							   "distributed relations"),
+						errdetail(
+							"Only exception is INSERT INTO dist_table SELECT ... FROM "
```
marcocitus: Hmz, is there a way to only display a relevant error when the query is specifically `INSERT INTO local_table SELECT ... FROM distributed_table ...`?

velioglu: By returning the reason of false result from `InsertSelectInfoDistributedTable` function , yes.

marcocitus: Either that or create a new function.

I think always showing this `errdetail` in all cases might only add to confusion.

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -2863,6 +2863,11 @@ NeedsDistributedPlanning(Query *queryTree)
 
 	if (hasLocalRelation && hasDistributedRelation)
 	{
+		if (InsertSelectIntoLocalTable(queryTree))
+		{
+			ereport(ERROR, (errmsg("cannot plan query, Citus only supports INSERT INTO "
+								   "dist_table SELECT ... FROM local_table")));
```
marcocitus: This is not quite accurate, we also support all sorts of other SELECTs (e.g. `FROM dist_table`). Maybe just say that `INSERT INTO local_table SELECT ... FROM dist_table...` is not supported?

src/test/regress/expected/multi_insert_select_non_pushable_queries.out
```diff
@@ -1,5 +1,15 @@
 ------------------------------------
 ------------------------------------
+-- Insert into local table
+------------------------------------
+------------------------------------
```
marcocitus: This commenting style is a little unusual.

src/backend/distributed/planner/insert_select_planner.c
```diff
@@ -80,11 +81,51 @@ static void CastSelectTargetList(List *selectTargetList, Oid targetRelationId,
...
+			return true;
+		}
+	}
+
+	return false;
+}
+
+
+bool
+InsertSelectIntoLocalTable(Query *query)
```
marcocitus: These need function comments.

metdos:
@velioglu, is this PR ready for another round of review?

velioglu:
@metdos No, not yet. I'm planning to update the PR today.

src/test/regress/expected/multi_insert_select_non_pushable_queries.out
velioglu: I used that commenting style to be consistent with the other comments in this file.

velioglu:
@marcocitus It is ready for the review. #1632 will solve the errors on Travis.

jasonmp85:
The Travis build for PostgreSQL 11 doesn't affect the health check here. The problem was the 9 and 10 builds were failing because new versions were released so the `initdb` version didn't match the `postgres` one. I had assumed the `enable-depends` flag would have fixed that, but I guess not? Clearing the cache and re-building didn't seem to fix anything…

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -2863,6 +2863,11 @@ NeedsDistributedPlanning(Query *queryTree)
 
 	if (hasLocalRelation && hasDistributedRelation)
 	{
+		if (InsertSelectIntoLocalTable(queryTree))
+		{
+			ereport(ERROR, (errmsg("cannot plan INSERT INTO local_table SELECT ... FROM"
```
jasonmp85: I think this should be reworded as _cannot `INSERT` rows from a distributed query into a local table_. The `dist_table` and `local_table` identifiers are ad hoc and will be misleading appearing in SQL like that.

jasonmp85:
After re-triggering the build it appears there is an actual test failure here.

metdos:
@velioglu, could you rebase and merge this?

src/test/regress/expected/multi_colocation_utils.out
```diff
@@ -266,11 +266,11 @@ SELECT shards_colocated(1300000, 1300020);
 (1 row)
 
 -- check co-located table list
-SELECT UNNEST(get_colocated_table_array('table1_group1'))::regclass;
+SELECT UNNEST(get_colocated_table_array('table1_group1'))::regclass ORDER BY 1;
```
jasonmp85: Uuuuggghhh 👎… we now have this change in three separate branches. How did this get checked in in the first place?!?

Lemme merge #1667 and you can rebase on that and remove this commit.

velioglu: Got the shipit before your PR I guess. Waiting for your merge.

PR #1584, opened 2017-08-16
Maintenance daemon ensures that the extension is valid

### LINKED ISSUE #1558: Deadlock while trying to drop extension

### LINKED ISSUE #1586: Dropping extension causes maintenanced to throw errors indefinitely
If the extension is dropped, the maintenance daemon keeps throwing an error and restarting every 5 seconds:

```
LOG:  starting maintenance daemon on database 13322 user 10
CONTEXT:  Citus maintenance daemon for database 13322 user 10
ERROR:  cache lookup failed for pg_dist_node, called too early?
CONTEXT:  Citus maintenance daemon for database 13322 user 10
LOG:  worker process: Citus Maintenance Daemon: 13322/10 (PID 13960) exited with exit code 1
```

onderkalaci:
Fixes #1558 
Fixes #1586

With this commit, we added some checks for ensuring that
the extension exists before doing the deadlock detection.

By checking and acquiring locks on the two tables that the
deadlock detection uses, we prevent any deadlocks between
`DROP EXTENSION citus` and the daemon.

### DIFF
--- a/src/backend/distributed/utils/maintenanced.c
+++ b/src/backend/distributed/utils/maintenanced.c
@@ -21,6 +21,8 @@
 #include "pgstat.h"
 
 #include "access/xact.h"
+#include "catalog/pg_extension.h"
+#include "commands/extension.h"
 #include "libpq/pqsignal.h"
 #include "distributed/distributed_deadlock_detection.h"
 #include "distributed/maintenanced.h"
@@ -29,6 +31,7 @@
 #include "storage/ipc.h"
 #include "storage/proc.h"
 #include "storage/latch.h"
+#include "storage/lmgr.h"
 #include "storage/lwlock.h"
 #include "tcop/tcopprot.h"
 
@@ -85,6 +88,8 @@ static void MaintenanceDaemonSigHupHandler(SIGNAL_ARGS);
 static size_t MaintenanceDaemonShmemSize(void);
 static void MaintenanceDaemonShmemInit(void);
 static void MaintenanceDaemonErrorContext(void *arg);
+static bool LockCitusExtension(void);
+
 
 /*
  * InitializeMaintenanceDaemon, called at server start, is responsible for
@@ -272,11 +277,12 @@ CitusMaintenanceDaemonMain(Datum main_arg)
 		{
 			StartTransactionCommand();
 
-			/*
-			 * We don't want to run the deadlock checks if there exists
-			 * any version mistmatch.
-			 */
-			if (CheckCitusVersion(DEBUG1))
+			if (!LockCitusExtension())
+			{
+				ereport(DEBUG1, (errmsg("could not lock the citus extension, "
+										"skipping deadlock detection")));
+			}
+			else if (CheckCitusVersion(DEBUG1) && CitusHasBeenLoaded())
 			{
 				foundDeadlock = CheckForDistributedDeadlocks();
 			}
@@ -471,4 +477,36 @@ MaintenanceDaemonErrorContext(void *arg)
 	MaintenanceDaemonDBData *myDbData = (MaintenanceDaemonDBData *) arg;
 	errcontext("Citus maintenance daemon for database %u user %u",
 			   myDbData->databaseOid, myDbData->userOid);
+}
+
+
+/*
+ * LockCitusExtension acquires a lock on the Citus extension or returns
+ * false if the extension does not exist or is being dropped.
+ */
+static bool
+LockCitusExtension(void)
+{
+	Oid recheckExtensionOid = InvalidOid;
+
+	Oid extensionOid = get_extension_oid("citus", true);
+	if (extensionOid == InvalidOid)
+	{
+		/* citus extension does not exist */
+		return false;
+	}
+
+	LockDatabaseObject(ExtensionRelationId, extensionOid, 0, AccessShareLock);
+
+	/*
+	 * The extension may have been dropped and possibly recreated prior to
+	 * obtaining a lock. Check whether we still get the expected OID.
+	 */
+	recheckExtensionOid = get_extension_oid("citus", true);
+	if (recheckExtensionOid != extensionOid)
+	{
+		return false;
+	}
+
+	return true;
 }
--- a/src/backend/distributed/utils/maintenanced.c
+++ b/src/backend/distributed/utils/maintenanced.c
@@ -267,6 +267,15 @@ CitusMaintenanceDaemonMain(Datum main_arg)
 		CHECK_FOR_INTERRUPTS();
 
 		/*
+		 * XXX: We clear the metadata cache before every iteration because otherwise
+		 * it might contain stale OIDs. It appears that in some cases invalidation
+		 * messages for a DROP EXTENSION may arrive during deadlock detection and
+		 * this causes us to cache a stale pg_dist_node OID. We'd actually expect
+		 * all invalidations to arrive after obtaining a lock in LockCitusExtension.
+		 */
+		ClearMetadataOIDCache();
+
+		/*
 		 * Perform Work.  If a specific task needs to be called sooner than
 		 * timeout indicates, it's ok to lower it to that value.  Expensive
 		 * tasks should do their own time math about whether to re-run checks.
--- a/src/backend/distributed/utils/metadata_cache.c
+++ b/src/backend/distributed/utils/metadata_cache.c
@@ -2694,8 +2694,18 @@ InvalidateDistRelationCacheCallback(Datum argument, Oid relationId)
 	 */
 	if (relationId != InvalidOid && relationId == MetadataCache.distPartitionRelationId)
 	{
-		memset(&MetadataCache, 0, sizeof(MetadataCache));
+		ClearMetadataOIDCache();
 	}
+}
+
+
+/*
+ * ClearMetadataOIDCache resets all the cached OIDs and the extensionLoaded flag.
+ */
+void
+ClearMetadataOIDCache(void)
+{
+	memset(&MetadataCache, 0, sizeof(MetadataCache));
 }
 
 
--- a/src/include/distributed/metadata_cache.h
+++ b/src/include/distributed/metadata_cache.h
@@ -86,6 +86,7 @@ extern List * DistTableOidList(void);
 extern List * ShardPlacementList(uint64 shardId);
 extern void CitusInvalidateRelcacheByRelid(Oid relationId);
 extern void CitusInvalidateRelcacheByShardId(int64 shardId);
+extern void ClearMetadataOIDCache(void);
 
 extern bool CitusHasBeenLoaded(void);
 extern bool CheckCitusVersion(int elevel);


src/backend/distributed/utils/maintenanced.c
```diff
@@ -345,6 +355,35 @@ CitusMaintenanceDaemonMain(Datum main_arg)
...
+MetadataTableAccessible(const char *tableName)
+{
+	OverrideSearchPath *overridePath = GetOverrideSearchPath(CurrentMemoryContext);
+	Oid relationId = InvalidOid;
+	bool lockAcquired = false;
+
+	/* only search for pg_catalog */
+	overridePath->schemas = NIL;
+	overridePath->addCatalog = true;
+	PushOverrideSearchPath(overridePath);
```
marcocitus: Maybe do something like:
```
            Oid         namespaceId;

            /* use exact schema given */
            namespaceId = LookupExplicitNamespace(relation->schemaname, true);
            if (!OidIsValid(namespaceId))
                return false;
```

src/backend/distributed/utils/maintenanced.c
```diff
@@ -270,8 +276,15 @@ CitusMaintenanceDaemonMain(Datum main_arg)
 			/*
 			 * We don't want to run the deadlock checks if there exists
 			 * any version mistmatch.
+			 *
+			 * We also want to make sure that the extension is available by checking
+			 * the existence of the required metadata tables and AccessShareLocks are
+			 * acquired on them.
```
marcocitus: maybe "and AccessShareLocks are acquired on them" -> "and whether they can be locked for reads. Deadlock detection is skipped if the tables are locked (e.g. when the extension is being dropped) or no longer exist."

marcocitus:
So far, the only way I found to completely prevent the deadlock detector from throwing errors is to clear the cache before every iteration. There appears to be something slightly dodgy about the way we do cache invalidation on DROP EXTENSION.

To handle DROP EXTENSION, we listen for invalidations of `pg_dist_partition`. If we get such an invalidation, we clear the metadata cache. When the deadlock detector calls `CitusHasBeenLoaded()` after the invalidation, `DistPartitionRelationId()` may still resolve `pg_dist_partition` to the old (pre-DROP) OID. However, it will never get another invalidation with that OID, which means all the OIDs already stored in the metadata cache, including the one of `pg_dist_node`, will remain stale.

src/backend/distributed/utils/maintenanced.c
```diff
@@ -257,6 +263,13 @@ CitusMaintenanceDaemonMain(Datum main_arg)
 		CHECK_FOR_INTERRUPTS();
 
 		/*
+		 * We clear the metadata cache before every iteration because otherwise
+		 * it might contain stale OIDs if deadlock detection runs during a DROP
+		 * EXTENSION followed by a CREATE EXTENSION.
+		 */
+		ClearMetadataCache();
```
onderkalaci: I think this is acceptable given that we're clearing a small amount of memory and doing two catalog reads (i.e., `pg_dist_node` and `pg_dist_logcal_group`) back to fill the required metadata.

My only comment would be that the function name is a bit strong, when I first saw it, I expected that we'd clear the distributed table caches as well.  However, I think that's not related to this PR, because the name of the struct we're clearing is `MetadataCacheData`. Probably not the best alternative, but maybe `ClearCatalogMetadataCache()` or so? (feel free to ignore this)

src/backend/distributed/utils/maintenanced.c
onderkalaci: updated the comment

src/backend/distributed/utils/maintenanced.c
```diff
@@ -345,6 +366,42 @@ CitusMaintenanceDaemonMain(Datum main_arg)
...
+	Oid relationId = get_relname_relid(tableName, PG_CATALOG_NAMESPACE);
+	if (OidIsValid(relationId))
+	{
+		bool lockAcquired = false;
+
+		lockAcquired = ConditionalLockRelationOid(relationId, AccessShareLock);
+		if (lockAcquired)
+		{
+			/*
+			 * We obtained the lock, but the relation may have just been dropped.
```
onderkalaci: Is that really possible? I think after the metadata clear, we shouldn't need that? How could someone drop a table if we're holding a lock?
I'm planning to remove this part if you're OK? (I also tested without this, and all looks good)

src/backend/distributed/utils/maintenanced.c
anarazel: Why do you think this is required and correct to do this here?

onderkalaci: > Why do you think this is required 

@marcocitus's above [comment](https://github.com/citusdata/citus/pull/1584#issuecomment-322817270) explains it.

We realized this while running consecutive `DROP extension & CREATE extension` commands as [here](https://gist.github.com/onderkalaci/b69d3ec169d2b52a207bd2ded011896a), we'd see some   `ERROR:  could not open relation with OID 449464` kind of errors 

> and correct to do this here 

Well, it's not wrong but has some performance drawbacks :)

anarazel: > We realized this while running consecutive DROP extension & CREATE extension commands as here, we'd see some ERROR: could not open relation with OID 449464 kind of errors

Right. But that's just hiding a bug.  In fact, I can reproduce at least one bug around this without maintenanced active (opening PR in a couple minutes).

anarazel:
>  To handle DROP EXTENSION, we listen for invalidations of pg_dist_partition. If we get such an invalidation, we clear the metadata cache. When the deadlock detector calls CitusHasBeenLoaded() after the invalidation, DistPartitionRelationId() may still resolve pg_dist_partition to the old (pre-DROP) OID.

"May" as in - you see a path were that happens, or you're observing it?  That's obviously not supposed to happen, given that we're supposed to have done a `ClearMetadataCache();` at that point.

I've found a bug when creating/dropping extensions, related to cache invalidation (opening PR), but it's on the shard rather than the `DistPartitionRelationId()` level.

I've recreated the extension a couple hundred times, and can't observe that.  I found other issues (we don't drop shards when DROP EXTENSION citus CASCADE - sounds familiar, so maybe I just rediscovered), but none resulting in the above error messages.

src/backend/distributed/utils/maintenanced.c
marcocitus: I because, at least in the background worker, `get_relname_relid` might still return the pre-DROP OID after the invalidation of `pg_dist_partition`.

src/backend/distributed/utils/maintenanced.c
```diff
@@ -345,6 +366,42 @@ CitusMaintenanceDaemonMain(Datum main_arg)
...
 /*
+ * MetadataTableAccessible returns whether the given metadata table exists
+ * and can be locked using an AccessShareLock.
+ */
+static bool
+MetadataTableAccessible(const char *tableName)
+{
+	bool tableAccessible = false;
+
+	Oid relationId = get_relname_relid(tableName, PG_CATALOG_NAMESPACE);
```
anarazel: Why not `RangeVarGetRelidExtended(..., missing_ok = true)`?

marcocitus: What do you mean by trylock callback?

We tried `RangeVarGetRelidExtended` at first, but realised it would throw an error if it could not obtain a lock. We then tried putting it in PG_TRY, but that turned out to be a bad idea in a background worker.

src/backend/distributed/utils/maintenanced.c
anarazel: Hm. There's no cache invalidation and no locking preventing the extension from being dropped and recreated inbetween get_relname_relid() and the relation actually being opened. So that chance always exists, given how the code is currently written - independent of the background daemon.  And that window is widened because get_relname_relid() itself doesn't do invalidation.  But even without that, it could always change after that function returned - you're going to need locks to prevent that.

src/backend/distributed/utils/maintenanced.c
anarazel: > What do you mean by trylock callback?

Instead of having RangeVarGetRelidExtended acquire the lock, acquire it in the callback. But there's plenty uglyness around it, so it's not really good.

> We then tried putting it in PG_TRY, but that turned out to be a bad idea in a background worker.

Hm. That's where you got the above error messages?

src/backend/distributed/utils/maintenanced.c
marcocitus: Sure, that's what `MetadataTableAccessible` solves for `pg_dist_node` and `pg_dist_local_group`. We make sure we can lock and open the relation prior to actually using it, and if not we skip distributed deadlock detection.

However, `pg_dist_partition` is never opened in the background worker, we only call `get_relname_relid` to get the OID and then expect to later get an invalidation for that OID when the extension is dropped. That assumption doesn't hold, which means our cache gets stale.

anarazel:
Hm, what do you guys think of "just" acquiring an object lock on the extension while a deadlock check is ongoing? Drop acquires that before it starts doing the actual dropping of the elements of the extension.  Then we don't need to worry about conditional lock acquisition, deadlocks on pg_dist_*, cache invalidation race conditions between lookups and action based on those lookups?

The only race condition would be between the looking up of the extension oid and the locking, but that seems far more manageable.

src/backend/distributed/utils/maintenanced.c
marcocitus: > Instead of having RangeVarGetRelidExtended acquire the lock, acquire it in the callback. But there's plenty uglyness around it, so it's not really good.

We figured the same and decided to just do what `RangeVarGetRelidExtended` does manually.

> Hm. That's where you got the above error messages from valgrind?

Yeh, that's an unrelated error being thrown outside of the PG_TRY block, but it seems the longjmp got kind of messed up after the PG_TRY.

anarazel:
> Hm, what do you guys think of "just" acquiring an object lock on the extension while a deadlock check is ongoing?

Something like https://github.com/citusdata/citus/commit/cb97474a13cf2420119de201476c922fbedc9cba

This approach seems a lot more robust to me, because we don't have to care about all the relations we're touching etc.

src/backend/distributed/utils/maintenanced.c
anarazel: > However, pg_dist_partition is never opened in the background worker, we only call get_relname_relid to get the OID and then expect to later get an invalidation for that OID when the extension is dropped. That assumption doesn't hold, which means our cache gets stale.

Hm, I don't actually see the problem here. `CitusHasBeenLoaded()` intentionally does a `DistPartitionRelationId()` to make sure the oid is cached, `AvailableExtensionVersion()` makes sure caches are initialized. Relache callback are called regardless of the relation previously being opened in the current backend or not.  Hence it shouldn't matter whether `pg_dist_partition` has been opened in maintenanced beforehand.  What am I missing?

anarazel: (that's with the locking I suggest present, there's race conditions without a lock on either p_d_p or the extension, but that's a different mechanism than you suggest)

marcocitus: The problem is that `CitusHasBeenLoaded()` may get the old `pg_dist_partition` OID from `get_relname_relid` after it has already been invalidated. After that, no more invalidations for that OID arrive, meaning `InvalidateDistRelationCacheCallback` never clears the OID cache because it never matches the oid in `MetadataCache.distPartitionRelationId`.

anarazel: > The problem is that CitusHasBeenLoaded() may get the old pg_dist_partition OID from get_relname_relid after it has already been invalidated. 

With the locking I've prototyped? The `LockDatabaseObject()` does an `AcceptSharedInvalidationMessages()`, which'll trigger all pending invalidations. And as that's done before  `CitusHasBeenLoaded()` I can't see how this happens?  Or are you alleging some as of yet undiagnosed citus independent bug in cache invalidation in postgres?

anarazel:
I'm giving up on this for the night. Think something like the above should be integrated for the deadlock issue.  I can't reliably reproduce the issue Marco reported, and I don't believe we have a coherent theory yet...

sumedhpathak:
> This approach seems a lot more robust to me
Commit message: "Hackkety hack hack."  😄

anarazel:
On 2017-08-17 20:50:42 +0000, Sumedh Pathak wrote:
> > This approach seems a lot more robust to me
> Commit message: "Hackkety hack hack."  😄 

Hey, it was 2AM ;) - and I just wanted to show what I was thinking of ;)

src/backend/distributed/utils/maintenanced.c
```diff
@@ -262,6 +267,14 @@ CitusMaintenanceDaemonMain(Datum main_arg)
 		CHECK_FOR_INTERRUPTS();
 
 		/*
+		 * We clear the metadata cache before every iteration because otherwise
```
anarazel: Can we put an XXX here and open an issue? Because something here is very much not right with this.

src/backend/distributed/utils/maintenanced.c
```diff
@@ -272,11 +285,10 @@ CitusMaintenanceDaemonMain(Datum main_arg)
...
 			StartTransactionCommand();
 
-			/*
-			 * We don't want to run the deadlock checks if there exists
-			 * any version mistmatch.
-			 */
-			if (CheckCitusVersion(DEBUG1))
+			/* to avoid the extension being dropped while we're using it */
+
+			if (LockCitusExtension() && CheckCitusVersion(DEBUG1) &&
```
anarazel: A debug mesage around here saying that we're skipping deadlock detection would be helpful.

Issue #1579, opened 2017-08-16
Cannot drop database when citus installed


onderkalaci:
The issue happens because the background daemon is already connected to the database, we should probably kill that process first.

```
-- connect to the main database
psql postgres
CREATE DATABASE test;

-- connect to the test database
psql test
CREATE EXTENSION citus;

-- issue a command to make sure that bg worker starts
\dx

-- connect back to the main database

postgres=# drop database test;
ERROR:  database "test" is being accessed by other users
DETAIL:  There is 1 other session using the database.
```

onderkalaci:
This seems tricky to fix. 

The problem is that we want to terminate the background daemon only if the `DROP DATABASE` command succeeds.

We cannot terminate the daemon after the process utility since it'd definitely fail as described in the issue description due to the daemon's connection to the database (see `BackgroundWorkerInitializeConnectionByOid()`).

If we remove the db from the maintenance daemon before the standard process utility, we might terminate the daemon before actually the db is dropped (which may fail due to connections or other stuff).

Any ideas on how to do this without any hacks?

marcocitus:
> The problem is that we want to terminate the background daemon only if the DROP DATABASE command succeeds.

Do we care that much? We could just let it terminate and remove it from the hash. If the `DROP DATABASE` rolls back and there are more commands, it will be restarted.

sumedhpathak:
@onderkalaci Can we close this issue now?

onderkalaci:
Yes, this is resolved with #1592. Closing the issue.

PR #1585, opened 2017-08-16
Maintenance daemon dies peacefully when it gets lost finding itself

marcocitus:
When postgres crashes, the maintenance daemon is automatically restarted, but its state in shared memory is no longer there. In that case, it currently throws a "got lost finding myself" errors and gets in an infinite restart loop.

This PR changes the error to `proc_exit(0)` which stops the background worker and avoids further restarts. A new maintenance daemon will be started as soon as `CitusHasBeenLoaded()` is called.


onderkalaci:
I'm OK with merging this in since that is an improvement, but this doesn't solve #1579. I'll open a PR for that which seems to be in 7.0 ?

PR #1591, opened 2017-08-17
Fix possible shard cache incoherency.

anarazel:
When a table and it's shards are dropped, and afterwards the same
shard identifiers are reused, e.g. due to a DROP & CREATE EXTENSION,
the old entry in the shard cache and the required entry in the shard
cache might be for different tables.

Force invalidation for both old and new table to fix.

### DIFF
--- a/src/backend/distributed/utils/metadata_cache.c
+++ b/src/backend/distributed/utils/metadata_cache.c
@@ -642,11 +642,18 @@ LookupShardCacheEntry(int64 shardId)
 
 		if (!shardEntry->tableEntry->isValid)
 		{
+			Oid oldRelationId = shardEntry->tableEntry->relationId;
+			Oid currentRelationId = LookupShardRelation(shardId);
+
 			/*
-			 * The cache entry might not be valid right now. Reload cache entry
-			 * and recheck (as the offset might have changed).
+			 * The relation OID to which the shard belongs could have changed,
+			 * most notably when the extension is dropped and a shard ID is
+			 * reused. Reload the cache entries for both old and new relation
+			 * ID and then look up the shard entry again.
 			 */
-			LookupDistTableCacheEntry(shardEntry->tableEntry->relationId);
+			LookupDistTableCacheEntry(oldRelationId);
+			LookupDistTableCacheEntry(currentRelationId);
+
 			recheck = true;
 		}
 	}


byucesoy:
:tada: I encountered with this bug and I was trying to find its reason. This is my test case;

```
ALTER SEQUENCE pg_catalog.pg_dist_shardid_seq RESTART 110000;
CREATE TABLE append_copy(id integer, data text, int_data int);
SELECT create_distributed_table('append_copy', 'id', 'append');
COPY append_copy FROM PROGRAM 'echo 0, a, 0\\n1, b, 1\\n2, c, 2\\n3, d, 3\\n4, e, 4' WITH CSV;

DROP TABLE IF EXISTS append_copy CASCADE;

ALTER SEQUENCE pg_catalog.pg_dist_shardid_seq RESTART 110000;
CREATE TABLE append_copy(id integer, data text, int_data int);
SELECT create_distributed_table('append_copy', 'id', 'append');
COPY append_copy FROM PROGRAM 'echo 0, a, 0\\n1, b, 1\\n2, c, 2\\n3, d, 3\\n4, e, 4' WITH CSV;
```
It used to error out like this;

```
ERROR:  could not find valid entry for shard 110000
```

After this PR it says;

```
ERROR:  cached metadata for shard 110000 is inconsistent
HINT:  Reconnect and try again.
```
I guess this is expected behavior? Is there a way to fix this completely?

onderkalaci:
> I guess this is the expected behavior? Is there a way to fix this completely?

I think that is the expected behavior. This is a very edge case and seems acceptable to re-connect.

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -642,11 +642,17 @@ LookupShardCacheEntry(int64 shardId)
...
+
 			/*
-			 * The cache entry might not be valid right now. Reload cache entry
-			 * and recheck (as the offset might have changed).
+			 * The cache entry might not be valid right now. Reload cache
+			 * entry and recheck. The offset might have changed, or more
+			 * gravely, this shard might now belong to a different
+			 * table. Access old relation and new relation to rebuild.
 			 */
+			LookupDistTableCacheEntry(oldRelationId);
```
marcocitus: Do we also need to do this if `oldRelationId == shardEntry->tableEntry->relationId`?

onderkalaci: > Do we also need to do this if oldRelationId == shardEntry->tableEntry->relationId

Probably in very few cases for now (looking where we invalidate the cache for distributed tables):

* Change colocation group via `mark_tables_colocated`
* Create / drop shards on `append/range`

I wouldn't bother about them much.

anarazel: Doing this twice superflously is nearly free anyway...

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -642,11 +642,17 @@ LookupShardCacheEntry(int64 shardId)
 
 		if (!shardEntry->tableEntry->isValid)
 		{
+			Oid oldRelationId = LookupShardRelation(shardId);
```
marcocitus: Why is this `oldRelationId`? This returns what's currently in `pg_dist_shard`.

src/backend/distributed/utils/metadata_cache.c
marcocitus: Still, I'd add an `if (shardEntry->tableEntry->relationId != oldRelationId)` to communicate that this is a rather exceptional case.

(or `currentRelationId` instead of `oldRelationId`)

anarazel:
On 2017-08-23 04:43:36 -0700, Marco Slot wrote:
> 
> Is there a way to also resolve the "metadata is inconsistent" error?
> 
> > @@ -642,11 +642,17 @@ LookupShardCacheEntry(int64 shardId)
>  
>  		if (!shardEntry->tableEntry->isValid)
>  		{
> +			Oid oldRelationId = LookupShardRelation(shardId);
> 
> Why is this `oldRelationId`? This returns what's currently in `pg_dist_shard`.

But we know the cache entry is out of date. How about renaming it to stale*?

marcocitus:
> But we know the cache entry is out of date. How about renaming it to stale*?

isn't `shardEntry->tableEntry->relationId` the stale relation OID? `LookupShardRelation` doesn't read from the cache.

sumedhpathak:
@anarazel ping on this one?

sumedhpathak:
Speaking with @anarazel we decided this isn't necessary for 7.0, since it isn't very critical. We can add this during our patch release mid-cycle. Removed the 7.0 milestone from this issue.

PR #1592, opened 2017-08-17
Terminate bg worker on drop database

onderkalaci:
I'm working on #1579.

My goal is the following:

* On `DROP DATABASE`, remove the db from the hash and terminate the bg worker before the standard utility kicks in
   * If standard utility succeeds we're done.
   * If fails, we'll re-start the bg worker since we've already removed that.

However, this doesn't work as expected. I couldn't make `TerminateBackgroundWorker () && WaitForBackgroundWorkerShutdown()` actually terminates the worker. This ends up not terminating the session to the database being dropped and getting the same error
```
ERROR:  database "test" is being accessed by other users
DETAIL:  There is 1 other session using the database.
```

Also,  `WaitForBackgroundWorkerShutdown()` really long time. Debugging shows that it actually returns `BGWH_STOPPED`, but on `ps` I can see that worker is not terminated.

Any comments on the approach or the problems I face? What am I missing here?

### DIFF
--- a/src/backend/distributed/utils/maintenanced.c
+++ b/src/backend/distributed/utils/maintenanced.c
@@ -24,9 +24,11 @@
 #include "catalog/pg_extension.h"
 #include "commands/extension.h"
 #include "libpq/pqsignal.h"
+#include "catalog/namespace.h"
 #include "distributed/distributed_deadlock_detection.h"
 #include "distributed/maintenanced.h"
 #include "distributed/metadata_cache.h"
+#include "nodes/makefuncs.h"
 #include "postmaster/bgworker.h"
 #include "storage/ipc.h"
 #include "storage/proc.h"
@@ -73,6 +75,7 @@ typedef struct MaintenanceDaemonDBData
 	/* information: which user to use */
 	Oid userOid;
 	bool daemonStarted;
+	pid_t workerPid;
 	Latch *latch; /* pointer to the background worker's latch */
 } MaintenanceDaemonDBData;
 
@@ -170,6 +173,7 @@ InitializeMaintenanceDaemonBackend(void)
 		}
 
 		dbData->daemonStarted = true;
+		dbData->workerPid = 0;
 		LWLockRelease(&MaintenanceDaemonControl->lock);
 
 		WaitForBackgroundWorkerStartup(handle, &pid);
@@ -225,10 +229,18 @@ CitusMaintenanceDaemonMain(Datum main_arg)
 		 */
 		proc_exit(0);
 	}
-	LWLockRelease(&MaintenanceDaemonControl->lock);
 
+	/* from this point, DROP DATABASE will attempt to kill the worker */
+	myDbData->workerPid = MyProcPid;
+
+	/* wire up signals */
+	pqsignal(SIGTERM, die);
+	pqsignal(SIGHUP, MaintenanceDaemonSigHupHandler);
+	BackgroundWorkerUnblockSignals();
 
 	myDbData->latch = MyLatch;
+
+	LWLockRelease(&MaintenanceDaemonControl->lock);
 
 	/*
 	 * Setup error context so log messages can be properly attributed. Some of
@@ -242,10 +254,6 @@ CitusMaintenanceDaemonMain(Datum main_arg)
 	errorCallback.previous = error_context_stack;
 	error_context_stack = &errorCallback;
 
-	/* wire up signals */
-	pqsignal(SIGTERM, die);
-	pqsignal(SIGHUP, MaintenanceDaemonSigHupHandler);
-	BackgroundWorkerUnblockSignals();
 
 	elog(LOG, "starting maintenance daemon on database %u user %u",
 		 databaseOid, myDbData->userOid);
@@ -286,6 +294,14 @@ CitusMaintenanceDaemonMain(Datum main_arg)
 		{
 			StartTransactionCommand();
 
+			/*
+			 * We skip the deadlock detection if citus extension
+			 * is not accessible.
+			 *
+			 * Similarly, we skip to run the deadlock checks if
+			 * there exists any version mismatch or the extension
+			 * is not fully created yet.
+			 */
 			if (!LockCitusExtension())
 			{
 				ereport(DEBUG1, (errmsg("could not lock the citus extension, "
@@ -518,4 +534,34 @@ LockCitusExtension(void)
 	}
 
 	return true;
+}
+
+
+/*
+ * StopMaintenanceDaemon stops the maintenance daemon for the
+ * given database and removes it from the maintenance daemon
+ * control hash.
+ */
+void
+StopMaintenanceDaemon(Oid databaseId)
+{
+	bool found = false;
+	MaintenanceDaemonDBData *dbData = NULL;
+	pid_t workerPid = 0;
+
+	LWLockAcquire(&MaintenanceDaemonControl->lock, LW_EXCLUSIVE);
+
+	dbData = (MaintenanceDaemonDBData *) hash_search(MaintenanceDaemonControl->dbHash,
+													 &databaseId, HASH_REMOVE, &found);
+	if (found)
+	{
+		workerPid = dbData->workerPid;
+	}
+
+	LWLockRelease(&MaintenanceDaemonControl->lock);
+
+	if (workerPid > 0)
+	{
+		kill(workerPid, SIGTERM);
+	}
 }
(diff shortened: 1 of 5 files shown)


src/backend/distributed/utils/maintenanced.c
```diff
@@ -158,7 +184,7 @@ InitializeMaintenanceDaemonBackend(void)
 		memcpy(worker.bgw_extra, &extensionOwner, sizeof(Oid));
 		worker.bgw_notify_pid = MyProcPid;
 
-		if (!RegisterDynamicBackgroundWorker(&worker, &handle))
+		if (!RegisterDynamicBackgroundWorker(&worker, &dbData->workerHandle))
```
anarazel: That can't work - the pointer is set to process local memory. So other processes can't access it after this.

marcocitus:
I guess the background worker functions expect there to be a process that keeps tracks of the handles (the parallel query group leader).

We could maybe just kill the maintenance daemon directly. We've effectively been doing this in the regression tests by calling `pg_terminate_backend` and it hasn't been a problem. I pushed a commit.

src/backend/distributed/utils/maintenanced.c
```diff
@@ -238,7 +247,6 @@ CitusMaintenanceDaemonMain(Datum main_arg)
 	error_context_stack = &errorCallback;
 
 	/* wire up signals */
-	pqsignal(SIGTERM, die);
```
anarazel: Why did you move this? Until the BackgroundWorkerUnblockSignals signals are blocked...

onderkalaci:
> We could maybe just kill the maintenance daemon directly.

This was my next thing to try.

One note on this approach is that if `DROP DATABASE ` fails due to other connections, we'd already have killed the daemon.  And until a new backend starts, we wouldn't be able to start the daemon on the existing backends (since the existing backends would always see `MetadataCache.extensionLoaded` as true, they'll never start the bg worker again).

I think there is not an easy way of fixing that by invalidating cache etc. We could may be not stop bg worker in the first place,  manually call `dropdb()` on `StopMaintenanceDaemon ` and then skip `standardProcessUtility`. 

Or, we can choose to live with that, which doesn't seem too bad to me.

anarazel:
On 2017-08-18 07:52:53 +0000, Önder Kalacı wrote:
> Or, we can choose to live with that, which doesn't seem too bad to me.

+1. The database was about to be dropped after all.

src/backend/distributed/utils/maintenanced.c
```diff
@@ -275,6 +283,9 @@ CitusMaintenanceDaemonMain(Datum main_arg)
 			/*
 			 * We don't want to run the deadlock checks if there exists
 			 * any version mistmatch.
+			 *
+			 * We also want to make sure that the extension is avaliable by checking
```
onderkalaci: a minor note: I think this is obsolete and will conflict with #1584.

src/test/regress/expected/multi_extension.out
```diff
@@ -287,15 +287,6 @@ FROM test.maintenance_worker();
...
 \c regression
-SELECT datname,
-    pg_terminate_backend(pid)
-FROM test.maintenance_worker('another');
- datname | pg_terminate_backend 
----------+----------------------
- another | t
-(1 row)
-
 DROP DATABASE another;
```
onderkalaci: Shall we run 

```
 SELECT datname,
     datname = 'another',		    
     usename = (SELECT extowner::regrole::text FROM pg_extension WHERE extname = 'citus'
 FROM test.maintenance_worker();	

```
to make sure that the daemon is dead?

src/backend/distributed/utils/maintenanced.c
```diff
@@ -86,6 +89,7 @@ static size_t MaintenanceDaemonShmemSize(void);
 static void MaintenanceDaemonShmemInit(void);
 static void MaintenanceDaemonErrorContext(void *arg);
 
+
```
onderkalaci: extra new line

onderkalaci: Oh not, just fixing it

marcocitus:
We might want to unblock signals before releasing the lwlock. That way we're guaranteed that the signal handler is set up when we do a kill.

src/backend/distributed/utils/maintenanced.c
anarazel: Ping, why are we moving the pqsignal around?

anarazel:
> We might want to unblock signals before releasing the lwlock. That way we're guaranteed that the signal handler is set up when we do a kill.

Hm? A blocked signal arriving while blocked will be delivered when the signals are unblocked, including using the right handler.

marcocitus:
> Hm? A blocked signal arriving while blocked will be delivered when the signals are unblocked, including using the right handler.

We'd like to the process to die as soon as the signal arrives though.

onderkalaci:
@marcocitus should be ready for you to look

src/backend/distributed/utils/maintenanced.c
```diff
@@ -243,7 +251,6 @@ CitusMaintenanceDaemonMain(Datum main_arg)
 	error_context_stack = &errorCallback;
 
 	/* wire up signals */
-	pqsignal(SIGTERM, die);
 	pqsignal(SIGHUP, MaintenanceDaemonSigHupHandler);
 	BackgroundWorkerUnblockSignals();
```
marcocitus: This two lines should move up to the other pqsignal call.

src/backend/distributed/utils/maintenanced.c
```diff
@@ -286,6 +293,14 @@ CitusMaintenanceDaemonMain(Datum main_arg)
 		{
 			StartTransactionCommand();
 
+			/*
+			 * We skip the deadlock detection if citus extension
+			 * is not accessible.
+			 *
+			 * Similarly, we skip to run the deadlock checks if
+			 * there exists any version mistmatch or the extension
```
marcocitus: mismatch

src/test/regress/expected/multi_extension.out
```diff
@@ -287,15 +288,47 @@ FROM test.maintenance_worker();
...
+    RETURNS pg_stat_activity
+    LANGUAGE plpgsql
+AS $$
+DECLARE
+   activity record;
+   tryCount int := 0;
+BEGIN
+    LOOP
+        SELECT * INTO activity FROM pg_stat_activity
+    WHERE application_name = 'Citus Maintenance Daemon' AND datname = p_dbname;
```
marcocitus: alignment is off

src/test/regress/expected/multi_extension.out
```diff
@@ -287,15 +288,47 @@ FROM test.maintenance_worker();
...
-SELECT datname,
-    pg_terminate_backend(pid)
-FROM test.maintenance_worker('another');
- datname | pg_terminate_backend 
----------+----------------------
- another | t
+CREATE SCHEMA test_deamon;
+-- we have to create the same function on the regression database as well
+-- note that this function is slightly different then the above, where
+-- the function tries for 5 times then exits if no deamon is found
```
marcocitus: why not just loop until the daemon is gone?

onderkalaci: Makes sense

src/backend/distributed/utils/maintenanced.c
onderkalaci: I'm very consistent about making typos 😄

src/test/regress/sql/multi_extension.sql
```diff
@@ -262,15 +262,52 @@ BEGIN
...
+        IF activity.pid IS NULL THEN
+            RETURN true;
+        ELSE
+            PERFORM pg_sleep(0.1);
+            PERFORM pg_stat_clear_snapshot();
+        END IF;
+        tryCount = tryCount + 1;
+        IF tryCount > 5 THEN
+            RETURN false;
+        END IF;
```
marcocitus: Do we still need the tries?

onderkalaci: Not for this specific test, but might be relevant if the `SIGTERM` is sent to the worker when it is not accepting signals?

marcocitus: Do we have a bug that allows that? 

It seems like we can only have two cases:
1) `StopMaintenanceDaemon` runs before the worker acquires the lwlock. The worker sees its not in the hash and calls `proc_exit(0)`.
2) `StopMaintenanceDaemon` runs after the worker acquires the lwlock. The signals and `workerPid` set up and when StopMaintenanceDaemon calls `kill`, the worker calls `die`.

onderkalaci: > Do we have a bug that allows that?

That's right, I got it now. We should probably not have the loop to test the above as well.

PR #1595, opened 2017-08-17
Improve deadlock detection

onderkalaci:
Some minor improvements


onderkalaci:
I cannot understand why code-coverage fails. I'll push a fake commit to re-start it.

PR #1598, opened 2017-08-18
Fix a crash on zero-shard tables

### LINKED ISSUE #1593: Select query on empty append distributed tables with where dist_col < or > crashes
My Setup;
* PostgreSQL REL_10_STABLE
* Citus Community 7.0-14

When I just create a basic table as;
```
CREATE TABLE test(id INT, uname TEXT);
```

and append distribute it on a column which I can use basic comparison operators (<, <=, >=, >). For example
```
SELECT create_distributed_table('test', 'id', 'append');
```
and run a select query which has a where clause at the end of the distribution column such as;
```
SELECT * FROM test WHERE id < 6;
```
Postgres server just crashes.

When I debug it I see that the exception is thrown on the line 1059 of the class shard_pruning.c
```
Assert(shardCount != 0);
```
Here the shardCount is 0 as expected since there is no data and the query hits no shard.

metdos:
Fixes #1593

### DIFF
--- a/src/backend/distributed/planner/shard_pruning.c
+++ b/src/backend/distributed/planner/shard_pruning.c
@@ -216,11 +216,18 @@ List *
 PruneShards(Oid relationId, Index rangeTableId, List *whereClauseList)
 {
 	DistTableCacheEntry *cacheEntry = DistributedTableCacheEntry(relationId);
+	int shardCount = cacheEntry->shardIntervalArrayLength;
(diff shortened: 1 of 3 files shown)


metdos:
@furkansahin could you also test changes?

furkansahin:
I tested and couldn't catch a problem. Looks fine. Thanks for the fix.

PR #1600, opened 2017-08-18
Add alias for target in multi-row INSERTs

### LINKED ISSUE #1596: INSERT INTO … RETURNING ... results in error for multi-row inserts.
Following succeeds:

```sql
INSERT INTO researchers
 VALUES
 (2, 1, 'Knuth Donald')
 RETURNING id;
```

But following generates error:

```sql
INSERT INTO researchers
 VALUES
 (2, 1, 'Knuth Donald'),
 (10, 6, 'Lamport Leslie')
 RETURNING id;

ERROR:  missing FROM-clause entry for table "researchers"
CONTEXT:  while executing command on localhost:9702
```

```researchers``` is created in the same way as in multi_modifying_xacts.sql.

### PR DESCRIPTION
**Pull request by jasonmp85**
Fixes #1596

This is necessary for multi-row INSERTs for the same reasons we use it in e.g. UPSERTs: if the range table list has more than one entry, then PostgreSQL's deparse logic requires that vars be prefixed by the name of their corresponding range table entry. This of course doesn't affect single-row INSERTs, but since multi-row INSERTs have a VALUE RTE, they were affected.

The piece of ruleutils which builds range table names wasn't modified to handle shard extension; instead UPSERT/INSERT INTO ... SELECT added an alias to the RTE. When present, this alias is favored. Doing the same in the multi-row INSERT case fixes RETURNING for such commands.

### DIFF
--- a/src/backend/distributed/planner/deparse_shard_query.c
+++ b/src/backend/distributed/planner/deparse_shard_query.c
@@ -82,11 +82,15 @@ RebuildQueryStrings(Query *originalQuery, List *taskList)
 
 			UpdateRelationToShardNames((Node *) copiedSubquery, relationShardList);
 		}
-		else if (task->upsertQuery)
+		else if (task->upsertQuery || valuesRTE != NULL)
 		{
 			RangeTblEntry *rangeTableEntry = NULL;
 
-			/* setting an alias simplifies deparsing of UPSERTs */
+			/*
+			 * Always an alias in UPSERTs and multi-row INSERTs to avoid
+			 * deparsing issues (e.g. RETURNING might reference the original
+			 * table name, which has been replaced by a shard name).
+			 */
 			rangeTableEntry = linitial(query->rtable);
 			if (rangeTableEntry->alias == NULL)
 			{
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: src/backend/distributed/planner/deparse_shard_query.c [hunk 1]
```diff
@@ -164,10 +164,19 @@ UpdateTaskQueryString(Query *query, Oid distributedTableId, RangeTblEntry *value
...
 		Assert(valuesRTE->rtekind == RTE_VALUES);
 
 		oldValuesLists = valuesRTE->values_lists;
 		valuesRTE->values_lists = task->rowValuesLists;
+
+		/* setting an alias simplifies deparsing of RETURNING */
+		if (distributedRte->alias == NULL)
+		{
+			Alias *alias = makeAlias(CITUS_TABLE_ALIAS, NIL);
+			distributedRte->alias = alias;
```
marcocitus: For upsert and INSERT..SELECT we set the alias in `RebuildQueryStrings` before calling `UpdateTaskQueryString`. Would it make sense to set the alias in `RebuildQueryStrings` for multi-row INSERTs as well? 

(we should probably clean this up and do it for all INSERTs)

PR #1602, opened 2017-08-21
Increase coverage of isolation tests - Part 2

byucesoy:
With this PR we add isolation tests for

- COPY to reference table vs. other operations
- COPY to partitioned table vs. other operations
- Multi row INSERTs vs other operations
- INSERT/SELECT vs. other operations
- UPSERT vs. other operations
- DELETE vs. other operations
- TRUNCATE vs. other operations
- DROP vs. other operations
- DDL vs. other operations

other operations consist of basic SQL operations (like SELECT,
INSERT, DELETE, UPSERT, COPY TRUNCATE, CREATE INDEX) as well
as some Citus functionalities (like master_modify_multiple_shards,
master_apply_delete_command, citus_total_relation_size etc.)

### DIFF
--- a/src/test/regress/specs/isolation_ddl_vs_all.spec
+++ b/src/test/regress/specs/isolation_ddl_vs_all.spec
@@ -1,0 +1,101 @@
+#
+# How we organize this isolation test spec, is explained at README.md file in this directory.
+#
+
+# create range distributed table to test behavior of DDL in concurrent operations
+setup
+{
+	SELECT citus.replace_isolation_tester_func();
+	SELECT citus.refresh_isolation_tester_prepared_statement();
+
+	SET citus.shard_replication_factor TO 1;
+	CREATE TABLE ddl_hash(id integer, data text);
+	SELECT create_distributed_table('ddl_hash', 'id');
+}
+
+# drop distributed table
+teardown
+{
+	DROP TABLE IF EXISTS ddl_hash CASCADE;
+
+	SELECT citus.restore_isolation_tester_func();
+}
+
+# session 1
+session "s1"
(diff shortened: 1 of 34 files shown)


src/test/regress/specs/isolation_ddl_vs_all.spec
```diff
@@ -0,0 +1,99 @@
+#
+# How we organize this isolation test spec, is explained at README.md file in this directory.
+#
+
+# create range distributed table to test behavior of DDL in concurrent operations
+setup
+{
+	SELECT citus.replace_isolation_tester_func();
```
mtuncer: This is done only for few tests. Do we need to do those for other isolation tests too ?

src/test/regress/specs/isolation_insert_select_vs_all.spec
```diff
@@ -0,0 +1,97 @@
+#
+# How we organize this isolation test spec, is explained at README.md file in this directory.
+#
+
+# create range distributed table to test behavior of INSERT/SELECT in concurrent operations
+setup
+{
+	SET citus.shard_replication_factor TO 1;
+	CREATE TABLE insert_select_append(id integer, data text);
+	SELECT create_distributed_table('insert_select_append', 'id', 'append');
```
mtuncer: we are testing self inserts on single shard append distributed tables. Is this viable scenario ?

mtuncer: I would also test with multi-shard and co-located tables.

mtuncer:
rebase, squash 🚢

PR #1603, opened 2017-08-21
Avoid overflowing PROCStack in BuildWaitGraphForSourceNode

marcocitus:
Currently, we may add the same process to the stack of remaining processes multiple times during the depth-first search in `BuildWaitGraphForSourceNode`, because we only inspect the visited status after a process has been added to the array. This means the size of the array (`MaxBackends`) may not be sufficient. 

After this change, the check is performed upfront to avoid adding the same process multiple times.

### DIFF
--- a/src/backend/distributed/transaction/lock_graph.c
+++ b/src/backend/distributed/transaction/lock_graph.c
@@ -32,12 +32,14 @@
 
 /*
  * PROCStack is a stack of PGPROC pointers used to perform a depth-first search
- * through the lock graph.
+ * through the lock graph. It also keeps track of which processes have been
+ * added to the stack to avoid visiting the same process multiple times.
  */
 typedef struct PROCStack
 {
 	int procCount;
 	PGPROC **procs;
+	bool *procAdded;
 } PROCStack;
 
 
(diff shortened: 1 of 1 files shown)


src/backend/distributed/transaction/lock_graph.c
```diff
@@ -706,6 +687,23 @@ AllocWaitEdge(WaitGraph *waitGraph)
...
+ * in the depth-first search, unless it was already added.
+ */
+static void
+AddProcToVisit(PROCStack *remaining, PGPROC *proc)
+{
+	if (remaining->procAdded[proc->pgprocno])
+	{
+		return;
+	}
+
```
onderkalaci: I'd add an `Assert (remaining->procCount < MaxBackends)`

src/backend/distributed/transaction/lock_graph.c
```diff
@@ -453,33 +453,14 @@ BuildWaitGraphForSourceNode(int sourceNodeId)
...
-		 * case of a local deadlock.
-		 */
-		if (visitedProcs[waitingProc->pgprocno])
-		{
-			continue;
-		}
-
-		visitedProcs[waitingProc->pgprocno] = true;
-
-		/* only blocked processes result in wait edges */
```
onderkalaci: Why are we removing this check? That seems not related to me?

marcocitus: It's fairly redundant, but we can keep it.

src/backend/distributed/transaction/lock_graph.c
marcocitus: added

PR #1606, opened 2017-08-21
Consider dropped columns that precede the partition column in COPY

### LINKED ISSUE #1605: COPY fails with NULL partition column error after ALTER TABLE even if partition column is specified.
**Steps to reproduce**

```
DROP TABLE newtable;                                                                         
CREATE TABLE newtable (col1 int, col2 int, col3 int, col4 int);                              
SELECT create_distributed_table('newtable','col3');                                          
ALTER TABLE newtable drop column col1;                                                       
\d newtable                                                                                  
COPY newtable (col2,col3,col4) from STDIN with CSV;                                          
,1,                                                                                          
,2,                                                                                          
\.

ERROR:  the partition column of table public.newtable should have a value
CONTEXT:  COPY newtable, line 1: ",1,"
```

If I try to create another table from scratch like this table, then the same copy goes through.

```
DROP TABLE newtable1;                                                                        
CREATE TABLE newtable1 (like newtable);                                                      
SELECT create_distributed_table('newtable1','col3');                                         
COPY newtable1 (col2,col3,col4) from STDIN with CSV;                                         
,1,                                                                                          
,2,                                                                                          
\.
```

Could it be because we have the old varattno in metadata and that has changed because of alter table? (Just asking -- random theory)

```
select * from pg_dist_partition where logicalrelid = 'newtable'::regclass or logicalrelid = 'newtable1'::regclass;
 logicalrelid | partmethod |                                                        partkey                                                         | colocationid | repmodel 
--------------+------------+------------------------------------------------------------------------------------------------------------------------+--------------+----------
 newtable1    | h          | {VAR :varno 1 :varattno 2 :vartype 23 :vartypmod -1 :varcollid 0 :varlevelsup 0 :varnoold 1 :varoattno 2 :location -1} |            1 | s
 newtable     | h          | {VAR :varno 1 :varattno 3 :vartype 23 :vartypmod -1 :varcollid 0 :varlevelsup 0 :varnoold 1 :varoattno 3 :location -1} |            1 | s
(2 rows)
```

marcocitus:
This change fixes #1605 by putting the burden of determining the partition column index on the caller. 

`COPY` and `create_distributed_table` generate tuples with NULL values in dropped columns in table-order even if the dropped column does not appear in the target list. After this change, we simply use the attribute number of the partition column - 1 as the partition column index.

`INSERT...SELECT` does not add dropped columns to tuples, but may not follow table order. It therefore looks up the attribute number of each column in the target list by name to determine whether it is the partition column. This logic remains the same, but is moved to the call site.

### DIFF
--- a/src/backend/distributed/commands/create_distributed_table.c
+++ b/src/backend/distributed/commands/create_distributed_table.c
@@ -1149,6 +1149,8 @@ CopyLocalDataIntoShards(Oid distributedRelationId)
 	List *columnNameList = NIL;
 	Relation distributedRelation = NULL;
 	TupleDesc tupleDescriptor = NULL;
+	Var *partitionColumn = NULL;
+	int partitionColumnIndex = INVALID_PARTITION_COLUMN_INDEX;
 	bool stopOnFailure = true;
 
 	EState *estate = NULL;
@@ -1189,6 +1191,13 @@ CopyLocalDataIntoShards(Oid distributedRelationId)
 	slot = MakeSingleTupleTableSlot(tupleDescriptor);
 	columnNameList = TupleDescColumnNameList(tupleDescriptor);
 
+	/* determine the partition column in the tuple descriptor */
+	partitionColumn = PartitionColumn(distributedRelationId, 0);
+	if (partitionColumn != NULL)
+	{
+		partitionColumnIndex = partitionColumn->varattno - 1;
+	}
+
 	/* initialise per-tuple memory context */
 	estate = CreateExecutorState();
 	econtext = GetPerTupleExprContext(estate);
@@ -1196,8 +1205,9 @@ CopyLocalDataIntoShards(Oid distributedRelationId)
 
 	copyDest =
 		(DestReceiver *) CreateCitusCopyDestReceiver(distributedRelationId,
-													 columnNameList, estate,
-													 stopOnFailure);
+													 columnNameList,
+													 partitionColumnIndex,
+													 estate, stopOnFailure);
 
 	/* initialise state for writing to shards, we'll open connections on demand */
 	copyDest->rStartup(copyDest, 0, tupleDescriptor);
--- a/src/backend/distributed/commands/multi_copy.c
+++ b/src/backend/distributed/commands/multi_copy.c
@@ -295,6 +295,8 @@ CopyToExistingShards(CopyStmt *copyStatement, char *completionTag)
 	bool *columnNulls = NULL;
 	int columnIndex = 0;
 	List *columnNameList = NIL;
+	Var *partitionColumn = NULL;
+	int partitionColumnIndex = INVALID_PARTITION_COLUMN_INDEX;
 	TupleTableSlot *tupleTableSlot = NULL;
 
 	EState *executorState = NULL;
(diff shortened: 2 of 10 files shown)


src/backend/distributed/commands/multi_copy.c
```diff
@@ -1716,8 +1726,8 @@ CopyFlushOutput(CopyOutState cstate, char *start, char *pointer)
  * a distributed table.
  */
```
onderkalaci: Might be worth adding some comments on the difference between `INSERT..SELECT` and `COPY`, and we could maybe add the comment here while explaining why we pass `partitionColumnIndex `?

src/backend/distributed/commands/create_distributed_table.c
```diff
@@ -1189,15 +1191,23 @@ CopyLocalDataIntoShards(Oid distributedRelationId)
 	slot = MakeSingleTupleTableSlot(tupleDescriptor);
 	columnNameList = TupleDescColumnNameList(tupleDescriptor);
 
+	/* determine the partition column */
```
onderkalaci: This might be a bit confusing to the readers (at first I was surprised as well thinking about `COPY table (col5, col2)` type of syntax where the user could easily change the order of the columns).

So, we're (and should be) doing this because Postgres adds dropped columns to the `slot->tts_values` anyway, right?  (Some debugging verified that for me)

It'd be nice to add a comment for that.

src/backend/distributed/commands/multi_copy.c
```diff
@@ -322,6 +324,14 @@ CopyToExistingShards(CopyStmt *copyStatement, char *completionTag)
 	tupleTableSlot->tts_values = columnValues;
 	tupleTableSlot->tts_isnull = columnNulls;
 
+	/* determine the partition column index */
```
onderkalaci: Same here, please consider adding a comment.

src/test/regress/expected/multi_insert_select.out
```diff
@@ -2401,21 +2401,28 @@ SELECT user_id, value_4 FROM test_view ORDER BY user_id, value_4;
...
 SELECT value_4, user_id FROM raw_events_second LIMIT 5;
-SELECT user_id, value_4 FROM raw_events_first ORDER BY user_id;
- user_id | value_4 
----------+---------
-       3 |       1
-       6 |       2
-       9 |       3
-      12 |       4
-      15 |       5
+SELECT * FROM drop_col_table ORDER BY col2, col3;
```
onderkalaci: Maybe add `WHERE col2 = ..` kind of query as well as you've already noted above `-- make sure the tuple went to the right shard`

src/test/regress/output/multi_copy.source
```diff
@@ -1050,6 +1050,25 @@ SELECT create_distributed_table('tt1','id');
 \copy tt1 from STDIN;
 DROP TABLE tt1;
 END;
+-- Test dropping a column in front of the partition column
+CREATE TABLE newtable (col1 int, col2 int, col3 int, col4 int);
```
onderkalaci: Very minor, but `newtable` seems too generic, maybe `drop_copy_test_table` or so? (Yes, we have worse table names as well, so feel free to ignore)

src/test/regress/input/multi_copy.source
```diff
@@ -781,6 +781,17 @@ SELECT create_distributed_table('tt1','id');
 DROP TABLE tt1;
 END;
 
+-- Test dropping a column in front of the partition column
```
onderkalaci: I can see that the code doesn't deal with the columns after the partition column. However, to guard against any future issue, shall we add a test after dropping `col4` as well?

onderkalaci: Also, might be good to try to COPY `NULL` values after drop column.

Issue #1604, opened 2017-08-21
Multi-row INSERT statements have poor performance


ozgune:
Citus 7.0 enables multi-row INSERT statements (#455). Our current implementation (#1517) relies on the multi-shard update executor.

The multi-shard update executor isn't designed for performance. In particular, if we have a multi-row INSERT statement that touches multiple shards, then we're going to open a new connection to each shard. For example, a multi-row INSERT statement that inserts 100 values and touches 32 shards, will open a total of 32 connections. This behavior results in poor performance for multi-row INSERT statements.

We should change our multi-row INSERT logic to re-use cached connections. I'm including an email thread from @marcocitus below.

---

Multi-row INSERTs currently use the same infrastructure as DDL and DML commands. For INSERTs that span multiple shards, that means we open a new connection for every shard placement and use aggressive locks to prevent deadlocks and connection explosions.

I performed 100-row INSERTs using pgbench over a varying number of shards by tweaking partition column values.

1 shard: 1900 tps = 190,000 rows/sec (17ms avg)
2 shards: 246 tps = 24,600 rows/sec  (131ms avg)
4 shards: 175 tps = 17,500 rows/sec (182ms avg)
8 shards: 132 tps = 13,200 rows/sec (240ms avg)
16 shards: 92 tps = 9,200 rows/sec (351ms avg)
32 shards: 58 tps = 5,800 rows/sec  (555ms avg)

Measured using 32 clients/threads over a 30 second period on a cluster with 2+1 r3.large nodes.

While the multi-shard locks are no longer necessary for avoiding deadlocks now that we have deadlock detection, removing the locks does not substantially improve throughput and causes connection errors due to the number of connections. The bottleneck is the number of connections.

To address throughput at high shard counts, we'll need to write a new executor that can reuse connections for multiple tasks within the same statement, e.g.: https://github.com/citusdata/citus/issues/1273

---

Hey @marcocitus, @jasonmp85,

These performance numbers are quite informative -- thanks! I had a question.

When you're doing an INSERT into Citus, we can re-use open connections and send INSERTs to shards that way. When we receive a multi-row INSERT, could we leverage the same infrastructure?

That is, can we take a multi-row INSERT and turn that into multiple single-row INSERT statements?

---

Yep, we could write code that does that. I'd expect it to perform better in most cases.

marcocitus:
It appears sequential execution (see #1608) gives substantially better throughput and latency, and is only a minimal code change:

[...]

PR #1608, opened 2017-08-22
Execute multi-row INSERTs sequentially

marcocitus:
This PR changes multi-row INSERTs to use sequential execution, over a single connection per worker. The current approach uses the multi-shard execution infrastructure for parallel DDL, but this causes poor performance (see #1604) due to the high number of connections.

Latency-wise, sequential execution is not the optimal way to execute a multi-row INSERT, but it gives substantially lower latency and higher throughput than the current approach.

Before:

1 shard: 1900 tps = 190,000 rows/sec (17ms avg)
2 shards: 246 tps = 24,600 rows/sec (131ms avg)
4 shards: 175 tps = 17,500 rows/sec (182ms avg)
8 shards: 132 tps = 13,200 rows/sec (240ms avg)
16 shards: 92 tps = 9,200 rows/sec (351ms avg)
32 shards: 58 tps = 5,800 rows/sec (555ms avg)

After:

1 shard: 2081 tps = 208,100 rows/sec (15ms avg)
2 shards: 1484 tps = 148,400 rows/sec (22ms avg)
4 shards: 1221 tps = 122,100 rows/sec (26ms avg)
8 shards: 893 tps = 89,300 rows/sec (36ms avg)
16 shards: 622 tps = 62,200 rows/sec (51ms avg)
32 shards: 385 tps = 38,500 rows/sec (83ms avg)

### DIFF
--- a/src/backend/distributed/executor/multi_executor.c
+++ b/src/backend/distributed/executor/multi_executor.c
@@ -58,17 +58,16 @@ static CustomExecMethods TaskTrackerCustomExecMethods = {
 	.ExplainCustomScan = CitusExplainScan
 };
 
-static CustomExecMethods RouterSingleModifyCustomExecMethods = {
-	.CustomName = "RouterSingleModifyScan",
+static CustomExecMethods RouterSequentialModifyCustomExecMethods = {
+	.CustomName = "RouterSequentialModifyScan",
 	.BeginCustomScan = CitusModifyBeginScan,
-	.ExecCustomScan = RouterSingleModifyExecScan,
+	.ExecCustomScan = RouterSequentialModifyExecScan,
 	.EndCustomScan = CitusEndScan,
 	.ReScanCustomScan = CitusReScan,
 	.ExplainCustomScan = CitusExplainScan
 };
 
-/* not static to enable reference by multi-modify logic in router execution */
-CustomExecMethods RouterMultiModifyCustomExecMethods = {
+static CustomExecMethods RouterMultiModifyCustomExecMethods = {
 	.CustomName = "RouterMultiModifyScan",
 	.BeginCustomScan = CitusModifyBeginScan,
 	.ExecCustomScan = RouterMultiModifyExecScan,
@@ -100,6 +99,7 @@ static CustomExecMethods CoordinatorInsertSelectCustomExecMethods = {
 static void PrepareMasterJobDirectory(Job *workerJob);
 static void LoadTuplesIntoTupleStore(CitusScanState *citusScanState, Job *workerJob);
 static Relation StubRelation(TupleDesc tupleDescriptor);
+static bool IsMultiRowInsert(Query *query);
 
 
 /*
@@ -165,7 +165,7 @@ RouterCreateScan(CustomScan *scan)
 	{
 		if (isModificationQuery)
 		{
-			scanState->customScanState.methods = &RouterSingleModifyCustomExecMethods;
+			scanState->customScanState.methods = &RouterSequentialModifyCustomExecMethods;
 		}
 		else
 		{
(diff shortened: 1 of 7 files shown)


sumedhpathak:
After chatting with @ozgune we decided to include this in 7.0 as well, given the performance benefits. @jasonmp85 Could you review the change?

src/backend/distributed/executor/multi_executor.c
```diff
@@ -174,8 +173,20 @@ RouterCreateScan(CustomScan *scan)
 	}
 	else
 	{
+		Query *jobQuery = workerJob->jobQuery;
+		CmdType commandType = jobQuery->commandType;
+
 		Assert(isModificationQuery);
-		scanState->customScanState.methods = &RouterMultiModifyCustomExecMethods;
+
+		if (commandType == CMD_INSERT && !InsertSelectIntoDistributedTable(jobQuery))
```
anarazel: Hm, framing this as a ! is a bit harder to understand.

src/backend/distributed/executor/multi_executor.c
```diff
@@ -174,8 +173,20 @@ RouterCreateScan(CustomScan *scan)
...
+		{
+			/* multi-row INSERT is executed sequentially */
+			scanState->customScanState.methods = &RouterSingleModifyCustomExecMethods;
```
anarazel: Perhaps this'd make more sense with s/Single/Sequential/?

src/test/regress/expected/multi_modifying_xacts.out
```diff
@@ -221,10 +221,12 @@ INSERT INTO researchers VALUES (10, 6, 'Lamport Leslie');
 ERROR:  cannot establish a new connection for placement 1200003, since DML has been executed on a connection that is in use
 CONTEXT:  COPY researchers, line 2: "10,6,Lesport Lampie"
 ROLLBACK;
--- but it is allowed after a multi-row insert
+-- COPY cannot be performed after a multi-row INSERT that uses one connection
```
anarazel: can we add a test for the copy; multi-row-insert case?

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -719,7 +717,7 @@ ExecuteSingleModifyTask(CitusScanState *scanState, Task *task, bool expectResult
 	 * customers already use functions that touch multiple shards from within
 	 * a function, so we'll ignore functions for now.
 	 */
-	if (IsTransactionBlock())
+	if (IsTransactionBlock() || multipleTasks)
```
anarazel: Hm, arguably that'd be better done on the callsite? But I don't care that much...

src/test/regress/expected/multi_modifying_xacts.out
marcocitus: added

src/backend/distributed/executor/multi_executor.c
```diff
@@ -175,14 +175,53 @@ RouterCreateScan(CustomScan *scan)
...
+	if (commandType != CMD_INSERT)
+	{
+		return false;
+	}
+
+	foreach(rteCell, query->rtable)
+	{
+		RangeTblEntry *rte = (RangeTblEntry *) lfirst(rteCell);
+
+		if (rte->rtekind == RTE_VALUES)
```
anarazel: This probably needs explanatory comment: This is an accurate check only because `transformInsertStmt()` optimizes away single-row `VALUES()` in an `INSERT`. That's not necessarily obvious imo.

src/backend/distributed/executor/multi_executor.c
```diff
@@ -175,14 +175,53 @@ RouterCreateScan(CustomScan *scan)
 	else
 	{
 		Assert(isModificationQuery);
-		scanState->customScanState.methods = &RouterMultiModifyCustomExecMethods;
+
+		if (IsMultiRowInsert(workerJob->jobQuery))
+		{
+			/* multi-row INSERT is executed sequentially */
```
anarazel: "... for performance reasons"?

src/backend/distributed/executor/multi_executor.c
marcocitus: added a comment to the function description

PR #1612, opened 2017-08-23
Prevent maintanince deamon crashes due to dead processes

### LINKED ISSUE #1611: Deadlock detection should not try to cancel an already dead process
If after the distributed deadlock detection decides to cancel
a backend, the backend has been terminated/killed/cancelled
externally, we might be accessing to a NULL pointer. 

This came up in the feature testing for distributed deadlock 
detection as discussed [in this internal document](https://docs.google.com/a/citusdata.com/document/d/1yx4VYCeZOmL64ure9riSnQeSGlVXPnSY-rgPUHpHPJM/edit?usp=drive_web)

onderkalaci:
Fixes #1611

If after the distributed deadlock detection decides to cancel
a backend, the backend has been terminated/killed/canceled
externally, we might be accessing to a NULL pointer. This commit
prevents that case by ignoring the current distributed deadlock.

### DIFF
--- a/src/backend/distributed/transaction/distributed_deadlock_detection.c
+++ b/src/backend/distributed/transaction/distributed_deadlock_detection.c
@@ -50,7 +50,7 @@ static void BuildDeadlockPathList(QueuedTransactionNode *cycledTransactionNode,
 								  TransactionNode **transactionNodeStack,
 								  List **deadlockPath);
 static void ResetVisitedFields(HTAB *adjacencyList);
-static void AssociateDistributedTransactionWithBackendProc(TransactionNode *
+static bool AssociateDistributedTransactionWithBackendProc(TransactionNode *
 														   transactionNode);
 static TransactionNode * GetOrCreateTransactionNode(HTAB *adjacencyList,
 													DistributedTransactionId *
@@ -149,7 +149,7 @@ CheckForDistributedDeadlocks(void)
 														&deadlockPath);
 		if (deadlockFound)
 		{
-			TransactionNode *youngestTransaction = transactionNode;
+			TransactionNode *youngestAliveTransaction = NULL;
 			ListCell *participantTransactionCell = NULL;
 
 			/*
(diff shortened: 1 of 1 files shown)


src/backend/distributed/transaction/distributed_deadlock_detection.c
```diff
@@ -175,25 +175,29 @@ CheckForDistributedDeadlocks(void)
...
 					youngestTransaction = currentNode;
 				}
 			}
 
-			/* we should find the backend */
-			Assert(youngestTransaction->initiatorProc != NULL);
+			/* there was a deadlock but the process already gone */
+			if (youngestTransaction->initiatorProc == NULL)
+			{
+				return false;
```
marcocitus: could we just continue with deadlock detection?

onderkalaci: actually, yes we can.

src/backend/distributed/transaction/distributed_deadlock_detection.c
```diff
@@ -175,25 +175,32 @@ CheckForDistributedDeadlocks(void)
...
 
-			/* we should find the backend */
-			Assert(youngestTransaction->initiatorProc != NULL);
+			/*
+			 * There was a deadlock but the process already gone,
+			 * continue searching for the other distributed deadlocks.
+			 */
+			if (youngestTransaction->initiatorProc == NULL)
+			{
+				continue;
```
marcocitus: Minor: I'd probably invert the if statement (i.e.  `if (youngestTransaction != NULL) { ...; return true }`)

src/backend/distributed/transaction/distributed_deadlock_detection.c
```diff
@@ -175,25 +175,32 @@ CheckForDistributedDeadlocks(void)
...
 				}
 			}
 
-			/* we should find the backend */
-			Assert(youngestTransaction->initiatorProc != NULL);
+			/*
+			 * There was a deadlock but the process already gone,
+			 * continue searching for the other distributed deadlocks.
+			 */
+			if (youngestTransaction->initiatorProc == NULL)
```
marcocitus: `initiatorProc` may or may not get set on the original `transactionNode` in `AssociateDistributedTransactionWithBackendProc`, which makes this code a bit hard to follow.

What if we initialise `youngestTransaction` as `NULL` at the start and only set it if  `AssociateDistributedTransactionWithBackendProc` returned true? Maybe also rename it to `youngestRunningTransaction`.

src/backend/distributed/transaction/distributed_deadlock_detection.c
```diff
@@ -175,32 +175,46 @@ CheckForDistributedDeadlocks(void)
...
-			/* we should find the backend */
-			Assert(youngestTransaction->initiatorProc != NULL);
-
-			CancelTransactionDueToDeadlock(youngestTransaction->initiatorProc);
-			LogCancellingBackend(youngestTransaction);
+			/*
+			 * We found the deadlock and its associated proc is still
+			 * alive.
+			 */
+			if (youngestTransaction && youngestTransaction->initiatorProc != NULL)
```
marcocitus: why would `initiatorProc` be NULL?

onderkalaci: Hmm, that can never be `NULL`, but it could have `pid < 0`, which should be fine since we already have a `warning` (i.e., we can remove the second part of the `if` statement?)

marcocitus: >  we can remove the second part of the if statement?)

yep, you could also add an assert inside the `if`.

onderkalaci: > yep, you could also add an assert inside the if.

I've had many assertion crashes, so I'll prefer not to add this time 😄 😄

marcocitus: Well, in this case it would segfault in `CancelTransactionDueToDeadlock`, but in any case, it doesn't seem like this could happen anymore.

PR #1613, opened 2017-08-23
Fix multi-row INSERT with RETURNING on reference tables

### LINKED ISSUE #1610: INSERT … RETURNING for multi-row inserts on reference tables errors out
We fixed issue #1596, but it appears INSERTs on reference tables behave differently and we still don't add the alias.

```
postgres=# CREATE TABLE test (x int primary key, y int);
CREATE TABLE
postgres=# SELECT create_reference_table('test');
 create_reference_table 
------------------------
 
(1 row)

postgres=# INSERT INTO test VALUES (1,2), (5,1) RETURNING *;
ERROR:  missing FROM-clause entry for table "test"
```

Note that this appears to be specific to reference tables, RETURNING does work on distributed tables, even when they have a single shard.

marcocitus:
Fixes #1610.

I first tried returning a `valuesRTE` for reference tables in `ExtractDistributedInsertValuesRTE`, but then I realised we don't set the `rowValuesList` for reference table inserts. I decided to set the `rowValuesList` in for multi-row inserts on reference tables as well, such that multi-row INSERT on distributed tables and reference tables are handled in the same way, which seems less prone to bugs.

### DIFF
--- a/src/backend/distributed/planner/multi_router_planner.c
+++ b/src/backend/distributed/planner/multi_router_planner.c
@@ -1854,6 +1854,7 @@ BuildRoutesForInsert(Query *query, DeferredErrorMessage **planningError)
 	{
 		int shardCount = 0;
 		List *shardIntervalList = LoadShardIntervalList(distributedTableId);
+		RangeTblEntry *valuesRTE = NULL;
 		ShardInterval *shardInterval = NULL;
 		ModifyRoute *modifyRoute = NULL;
 
@@ -1867,7 +1868,17 @@ BuildRoutesForInsert(Query *query, DeferredErrorMessage **planningError)
 		modifyRoute = palloc(sizeof(ModifyRoute));
 
 		modifyRoute->shardId = shardInterval->shardId;
-		modifyRoute->rowValuesLists = NIL;
+
+		valuesRTE = ExtractDistributedInsertValuesRTE(query);
+		if (valuesRTE != NULL)
+		{
+			/* add the values list for a multi-row INSERT */
+			modifyRoute->rowValuesLists = valuesRTE->values_lists;
+		}
+		else
+		{
+			modifyRoute->rowValuesLists = NIL;
+		}
 
 		modifyRouteList = lappend(modifyRouteList, modifyRoute);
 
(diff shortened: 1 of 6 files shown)


src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1987,6 +1998,39 @@ BuildRoutesForInsert(Query *query, DeferredErrorMessage **planningError)
...
 
 /*
+ * ExtractDistributedInsertValuesRTE does precisely that. If the provided
+ * query is not an INSERT, or if the INSERT does not have a VALUES RTE
+ * (i.e. it is not a multi-row INSERT), this function returns NULL.
+ * If all those conditions are met, an RTE representing the multiple values
+ * of a multi-row INSERT is returned.
+ */
+RangeTblEntry *
+ExtractDistributedInsertValuesRTE(Query *query)
```
jasonmp85: The logic here is now identical to `IsMultiRowInsert`. Can you change the sole caller of it to instead call this and just check whether the return value is `NULL`?

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1867,7 +1868,17 @@ BuildRoutesForInsert(Query *query, DeferredErrorMessage **planningError)
 		modifyRoute = palloc(sizeof(ModifyRoute));
 
 		modifyRoute->shardId = shardInterval->shardId;
-		modifyRoute->rowValuesLists = NIL;
```
jasonmp85: If it's now an invariant that all `modifyRoutes` have a non-`NULL` `rowValuesList`, can you add an `Assert` somewhere to ensure we catch malformed values correctly?

jasonmp85:
Rebased this for you because why not.

src/backend/distributed/planner/multi_router_planner.c
marcocitus: Yep, added it it in deparse_shard_query.c where it gets used.

src/backend/distributed/planner/multi_router_planner.c
marcocitus: Changed the implementation of `IsMultiRowInsert` to `return ExtractDistributedInsertValuesRTE(query) != NULL` to preserve readability and the function comment.

marcocitus:
Thanks for the review! Hope you're feeling better.

PR #1616, opened 2017-08-23
Don't error out if deadlock detection fails to connect to worker

### LINKED ISSUE #1590: Maintenance daemon stops querying workers if one of them dies
This has come up during failure testing.

After we issue the first query we start to see following  in worker logs  every 2-3 seconds
```
SELECT * FROM dump_local_wait_edges($1)
```

If one of the workers is stopped, worker logs no longer displays that message but we got following in master logs every 5 seconds
```
2017-08-17 10:19:33.967 +03 [4494] LOG:  starting maintenance daemon on database 12558 user 10
2017-08-17 10:19:33.967 +03 [4494] CONTEXT:  Citus maintenance daemon for database 12558 user 10
2017-08-17 10:19:33.987 +03 [4494] ERROR:  connection error: localhost:9700
2017-08-17 10:19:33.987 +03 [4494] DETAIL:  no connection to the server
	
2017-08-17 10:19:33.987 +03 [4494] CONTEXT:  Citus maintenance daemon for database 12558 user 10
2017-08-17 10:19:33.988 +03 [3946] LOG:  worker process: Citus Maintenance Daemon: 12558/10 (PID 4494) exited with exit code 1
```

marcocitus:
Fixes #1590

Change `ERROR` to `WARNING` in `BuildGlobalWaitGraph` and use the remaining information to perform deadlock detection.

### DIFF
--- a/src/backend/distributed/transaction/lock_graph.c
+++ b/src/backend/distributed/transaction/lock_graph.c
@@ -134,7 +134,7 @@ BuildGlobalWaitGraph(void)
 		querySent = SendRemoteCommand(connection, command);
 		if (querySent == 0)
 		{
-			ReportConnectionError(connection, ERROR);
+			ReportConnectionError(connection, WARNING);
 		}
 	}
 
@@ -151,7 +151,8 @@ BuildGlobalWaitGraph(void)
 		result = GetRemoteCommandResult(connection, raiseInterrupts);
 		if (!IsResponseOK(result))
 		{
-			ReportResultError(connection, result, ERROR);
+			ReportResultError(connection, result, WARNING);
+			continue;
 		}
 
 		rowCount = PQntuples(result);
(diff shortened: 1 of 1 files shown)


src/backend/distributed/transaction/lock_graph.c
```diff
@@ -134,7 +134,7 @@ BuildGlobalWaitGraph(void)
 		querySent = SendRemoteCommand(connection, command);
 		if (querySent == 0)
 		{
-			ReportConnectionError(connection, ERROR);
+			ReportConnectionError(connection, WARNING);
```
onderkalaci: Pretty minor, maybe add `continue;` here as well in case anyone adds some code below?

onderkalaci:
(Also, to increase the coverage, you might consider adding a fake worker and ensuring that deadlock detection finds the deadlock on the other nodes while `WARNING` for the fake node. But doesn't seem strictly necessary to add such test)

PR #1618, opened 2017-08-23
Update build instructions in CONTRIBUTING.md

begriffs:
While building master today on Ubuntu I discovered that I now needed two more libraries installed.

I haven't tested whether all the libraries in the list are actually still needed, but the build works for me when they, plus the two new ones, are present.


jasonmp85:
This looks fine to me and makes sense from what I remember when I set up build boxes lately.

PR #1621, opened 2017-08-24
Allow default columns in multi-row INSERTs

### LINKED ISSUE #1620: multi row insert fails when used with serial column
Table is created like below
```sql
create table tr1(id bigserial, a int);
select create_distributed_table('tr1', 'id');
```
Then tried single row insert
```sql 
 insert into tr1(a) values (1);
```
It worked fine, however when I tried to run with multiple values like

```sql
insert into tr1(a) values (3), (5);
```
I got an error thrown from worker
```sql
WARNING:  syntax error at or near ")"
```
worker log is
```
ERROR:  syntax error at or near ")" at character 68
STATEMENT:  INSERT INTO public.tr1_102041 AS citus_table_alias (id, a) VALUES ()
```

Values part is not populated if partition column is of serial type

marcocitus:
Default columns in multi-row `INSERT` commands are represented as `Const` or expressions in the target list, while values in the `VALUES` section are represented with a `Var` in the target list that refers to an column index in `values_lists`. The current implementation only considers the latter which results incorrect queries on shards when columns with default values are defined.

This change adds default expressions to the `values_lists` of the `originalQuery` early on in planning. This is not the prettiest solution, since the resulting `Query *` is not entirely valid, but it can be deparsed and it avoids a complete redesign. The change also clears up some assumptions around the partition column being a `Var`, which might not be the case.

I first tried a solution which only the `rowValuesLists` were expanded to leave the query untouched, but the main issue with that approach is that `rowValuesLists` may get generated after function evaluation, in which case default expressions are not evaluated (apart from some other issues like needing to evaluate expressions in `rowValuesLists` separately).

Fixes #1620.

### DIFF
--- a/src/backend/distributed/planner/multi_router_planner.c
+++ b/src/backend/distributed/planner/multi_router_planner.c
@@ -133,6 +133,7 @@ static List * WorkersContainingAllShards(List *prunedShardIntervalsList);
 static List * BuildRoutesForInsert(Query *query, DeferredErrorMessage **planningError);
 static List * GroupInsertValuesByShardId(List *insertValuesList);
 static List * ExtractInsertValuesList(Query *query, Var *partitionColumn);
+static int GetTargetListEntryIndexByResno(List *targetList, int resno);
 static bool MultiRouterPlannableQuery(Query *query,
 									  RelationRestrictionContext *restrictionContext);
 static DeferredErrorMessage * ErrorIfQueryHasModifyingCTE(Query *queryTree);
@@ -1080,10 +1081,35 @@ RouterInsertJob(Query *originalQuery, Query *query, DeferredErrorMessage **plann
 	Job *job = NULL;
 	bool requiresMasterEvaluation = false;
 	bool deferredPruning = false;
+	bool isMultiRowInsert = false;
 
-	if (!CanShardPrune(distributedTableId, query))
+	RangeTblEntry *valuesRTE = ExtractDistributedInsertValuesRTE(originalQuery);
+	if (valuesRTE != NULL)
 	{
-		/* there is a non-constant in the partition column, cannot prune yet */
+		/*
+		 * We expand the values_lists to contain all default expressions
+		 * from the target list. By doing this early on, in the original
+		 * query, we can later evaluate default expressions for each
+		 * individual row and then perform shard pruning.
+		 */
+		valuesRTE->values_lists = ExpandValuesLists(originalQuery->targetList,
+													valuesRTE->values_lists);
+
+		isMultiRowInsert = true;
+	}
+
+	if (isMultiRowInsert || !CanShardPrune(distributedTableId, query))
+	{
+		/*
+		 * If there is a non-constant (e.g. parameter, function call) in the partition
+		 * column of the INSERT then we defer shard pruning until the executor where
+		 * these values are known.
+		 *
+		 * XXX: We also defer pruning for multi-row INSERTs because of some current
+		 * limitations with the way multi-row INSERTs are handled. Most notably, we
+		 * don't evaluate functions in task->rowValuesList. Therefore we need to
+		 * perform function evaluation before we can run RouterInsertTaskList.
+		 */
 		taskList = NIL;
 		deferredPruning = true;
 
@@ -2031,6 +2057,50 @@ ExtractDistributedInsertValuesRTE(Query *query)
 
 
 /*
+ * ExpandValuesLists expands VALUES lists by building new lists
+ * that include expressions from the target list for columns
+ * with a default expression.
+ */
+List *
+ExpandValuesLists(List *targetList, List *valuesLists)
+{
+	ListCell *valuesListCell = NULL;
+	List *expandedValuesLists = NIL;
+	ListCell *targetEntryCell = NULL;
+
+	foreach(valuesListCell, valuesLists)
+	{
+		List *valuesList = (List *) lfirst(valuesListCell);
+		List *expandedValuesList = NIL;
+
+		foreach(targetEntryCell, targetList)
+		{
+			TargetEntry *targetEntry = (TargetEntry *) lfirst(targetEntryCell);
+			Expr *targetExpr = targetEntry->expr;
+
+			if (IsA(targetExpr, Var))
+			{
+				/* expression from the VALUES section */
+				Var *targetListVar = (Var *) targetExpr;
+				targetExpr = list_nth(valuesList, targetListVar->varattno - 1);
+			}
+			else
+			{
+				/* copy the column's default expression */
+				targetExpr = copyObject(targetExpr);
+			}
+
+			expandedValuesList = lappend(expandedValuesList, targetExpr);
+		}
+
+		expandedValuesLists = lappend(expandedValuesLists, expandedValuesList);
+	}
+
+	return expandedValuesLists;
+}
+
+
+/*
  * IntersectPlacementList performs placement pruning based on matching on
  * nodeName:nodePort fields of shard placement data. We start pruning from all
  * placements of the first relation's shard. Then for each relation's shard, we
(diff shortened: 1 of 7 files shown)


jasonmp85:
Looking into this.

jasonmp85:
Can you expound on _ the main issue with that approach is that `rowValuesLists` may get generated after function evaluation, in which case default expressions are not evaluated_? I at one point had some code to iterate over `rowValuesLists` and evaluate them there (rather than as part of the `Query`, but after a refactor I removed it as unnecessary. I'm not entirely clear on the approaches you explored here.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -2138,10 +2208,11 @@ static List *
...
 {
 	List *insertValuesList = NIL;
-	TargetEntry *targetEntry = get_tle_by_resno(query->targetList,
-												partitionColumn->varattno);
+	RangeTblEntry *valuesRTE = NULL;
 
-	if (targetEntry == NULL)
+	int partitionColumnIndex = GetTargetListEntryIndexByResno(query->targetList,
+															  partitionColumn->varattno);
+	if (partitionColumnIndex == -1)
```
jasonmp85: I'm assuming this is in response to _The change also clears up some assumptions around the partition column being a Var, which might not be the case_. When is that not the case? This review's description could use a little expansion (in lieu of which I'll probably have to re-investigate this issue and print out query structures myself to see what you're alluding to).

marcocitus:
The main issue is described in the first paragraph of the PR.

src/backend/distributed/planner/multi_router_planner.c
marcocitus: Indeed, This particular change is because the partition column's target entry may contain an arbitrary expression which does not refer to values_lists. However, after values_lists expansion we know that the index of the target entry matches the column index in the values_lists.

jasonmp85:
> The main issue is described in the first paragraph of the PR.

Right; I'd read the first sentence like five times and am still unclear precisely what it describes… I'll print out some `Query` trees and try to see what you mean.

marcocitus:
If you do
```
CREATE TABLE test (x int, y serial);
INSERT INTO test (x) VALUES (3), (4);
```
then there will be 
- a target list entry for column `x` which has its `expr` set to a `Var` which refers to the column index in `values_lists`
- a target list entry for column `y` which has its `expr` set to the default expression (`nextval(...)`) instead of a `Var` and it will not have a value in `values_lists`

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1080,10 +1081,35 @@ RouterInsertJob(Query *originalQuery, Query *query, DeferredErrorMessage **plann
...
+		 * query, we can later evaluate default expressions for each
+		 * individual row and then perform shard pruning.
+		 */
+		valuesRTE->values_lists = ExpandValuesLists(originalQuery->targetList,
+													valuesRTE->values_lists);
+
+		isMultiRowInsert = true;
+	}
+
+	if (isMultiRowInsert || !CanShardPrune(distributedTableId, query))
```
jasonmp85: This bookkeeping boolean is more confusing than simply inlining the `valuesRTE != NULL` in this `if` statement. Can you switch to that?

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -2031,6 +2057,50 @@ ExtractDistributedInsertValuesRTE(Query *query)
...
+		foreach(targetEntryCell, targetList)
+		{
+			TargetEntry *targetEntry = (TargetEntry *) lfirst(targetEntryCell);
+			Expr *targetExpr = targetEntry->expr;
+
+			if (IsA(targetExpr, Var))
+			{
+				/* expression from the VALUES section */
+				Var *targetListVar = (Var *) targetExpr;
+				targetExpr = list_nth(valuesList, targetListVar->varattno - 1);
```
jasonmp85: This loop's performance isn't going to be great, right? You're looping over the entire target list to process each row, and when you hit a var, you use `list_nth` (linear access) to find a node to copy.

src/backend/distributed/planner/multi_router_planner.c
jasonmp85: Also, couldn't `CanShardPrune` just incorporate this check? Because—after this PR—wouldn't it be true that multi-row `INSERT`s cannot be shard-pruned?

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1080,10 +1081,35 @@ RouterInsertJob(Query *originalQuery, Query *query, DeferredErrorMessage **plann
...
+	}
+
+	if (isMultiRowInsert || !CanShardPrune(distributedTableId, query))
+	{
+		/*
+		 * If there is a non-constant (e.g. parameter, function call) in the partition
+		 * column of the INSERT then we defer shard pruning until the executor where
+		 * these values are known.
+		 *
+		 * XXX: We also defer pruning for multi-row INSERTs because of some current
```
jasonmp85: I think this is a pretty big caveat worth mentioning up front in the PR. The tradeoff here is that now _no_ multi-row inserts are prunable in the planning phase, even if they contain only constants. That seems like a tradeoff worth calling out.

jasonmp85: (I mean, there _is_ an `XXX` here, but I don't know the semantics of that other than "read this").

jasonmp85:
Hey, so I haven't had a ton of time to test it, but I came up with an alternate approach when reading this review… what if we just add the non-`Var` TLE expressions to each `values_lists` list and replace the `tle->expr` with a `Var` we create to appropriately reference the new `values_lists` entries?

I didn't yet copy your (much appreciated!) tests into my PR, but it's in #1622. I talked to @sumedhpathak about the hairiness of the multi-row `DEFAULT` representation and said I'd leave commentary on this review in addition to a counterproposal.

You can choose which you wish to merge, but given the change in behavior (around pruning-at-planning-time) in this PR, I think I favor #1622 (unless you know of any showstopper reasons why that approach is flawed). If you could move your tests into #1622 and brush it up, that should unblock us (I believe).

src/backend/distributed/planner/multi_router_planner.c
marcocitus: We didn't do shard pruning in the planning phase before this PR either since CanShardPrune returns false if there is a non-Const (e.g. false) in the target entry of the partition column.

jasonmp85: The multi-row INSERT PR [changed that](https://github.com/citusdata/citus/blob/6a35c2937cd06f2604d9c59d701f520788ade9fb/src/backend/distributed/planner/multi_router_planner.c#L1167-L1177)… It uses the more abstract `ExtractInsertValuesList` to get a list of all values (whether they be sourced from `values_lists` or the target list) and iterates until it finds a non-`Const`.

marcocitus:
> I'm not very keen on ripping out plantime prunability for all multi-row INSERTs (including those with all constants). It seems like a performance improvement we might want to suggest to someone at some point.

I don't think many users will have prepared multi-row `INSERT` statements with a single constant in the partition column that are used >5 times, which is the only case in which you get some (tiny) performance benefit. Most use cases we've heard want to use multi-row INSERTs for bulk loading an arbitrary number of rows across shards.

marcocitus:
> In addition, I have a hunch that the worst-case performance of ExpandValuesLists will be pretty bad. The common case (non-DEFAULT columns) here falls through to list_nth, which is called for every target entry for every row. I could be seeing ghosts, but I think that will hurt performance considerably (which we just tried to improve).

Agreed, this can be optimised by converting the list into an array to avoid the cost of `list_nth`.

src/backend/distributed/planner/multi_router_planner.c
jasonmp85: I kind of wonder if there isn't some way to rework this to operate in-place (injecting a cell into an extant list?) rather than rebuilding the lists. You're already destructively reassigning the `values_lists` field at the end of it all anyhow.

jasonmp85:
I'm fine with either PR at this point. I see what is needed to fix #1622 (which I like for being slightly less invasive and more consistent) and see you can probably get the `list_nth` cost to go away (loop is still `O(rows * columns)` with no default values). The prepare thing I would have liked called out earlier, but I see you thought they were never preparable in the first place (I think my changes to `CanShardPrune` meant they were at least before #1621).

So I guess brush up one or the other with the feedback from each and :shipit:

marcocitus:
I liked the idea of properly rebuilding the target list upfront. I combined our 2 PRs.

We still cannot do shard pruning in the planner, because we cannot build `rowValuesLists` before we do function evaluation. I'll open an issue for that.

@mtuncer could you have a final look?

src/backend/distributed/planner/multi_router_planner.c
marcocitus: Ah true, but this didn't work correctly when there is an expression in a non-partition column, since we wouldn't evaluate it on the master. The query may still work because the worker can run functions like `now()`, though not sequences.

marcocitus:
Thanks for the late night review and testing!

PR #1622, opened 2017-08-25
Ensure multi-row INSERT TL is all-Var

### LINKED ISSUE #1620: multi row insert fails when used with serial column

jasonmp85:
The changes here are mostly explained in the function comment for the new `NormalizeInsertTargetList` function, which ensures bare `DEFAULT` values are "denormalized" into each values list and replaced by a `Var` reference to those new elements.

Repeated here for ease of reference:

```c
/*
 * NormalizeInsertTargetList ensures all elements of multi-row INSERT target
 * lists are Vars. In multi-row INSERTs, most target list entries contain a Var
 * expression pointing to a position within the values_lists field of a VALUES
 * RTE, but non-NULL DEFAULT columns are handled differently. Instead of adding
 * the DEFAULT to each row, a single expression encoding the DEFAULT appears
 * in the target list. For consistency, we move these expressions into values
 * lists and replace them with an appropriately constructed Var.
 */
```

Fixes #1620

Intended as an alternative to #1621

### DIFF
--- a/src/backend/distributed/planner/multi_router_planner.c
+++ b/src/backend/distributed/planner/multi_router_planner.c
@@ -130,6 +130,7 @@ static List * TargetShardIntervalsForRouter(Query *query,
 											RelationRestrictionContext *restrictionContext,
 											bool *multiShardQuery);
 static List * WorkersContainingAllShards(List *prunedShardIntervalsList);
+static void NormalizeInsertTargetList(Query *query);
 static List * BuildRoutesForInsert(Query *query, DeferredErrorMessage **planningError);
 static List * GroupInsertValuesByShardId(List *insertValuesList);
 static List * ExtractInsertValuesList(Query *query, Var *partitionColumn);
@@ -1219,6 +1220,8 @@ RouterInsertTaskList(Query *query, DeferredErrorMessage **planningError)
 
 	Assert(query->commandType == CMD_INSERT);
 
+	NormalizeInsertTargetList(query);
+
 	modifyRouteList = BuildRoutesForInsert(query, planningError);
 	if (*planningError != NULL)
 	{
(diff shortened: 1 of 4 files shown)


jasonmp85:
OK, yeah, so I only developed on PostgreSQL 10. My bad. 😊

It's worth mentioning that PostgreSQL enforces an invariant that every list within `values_lists` must have the same length, so rather than using the (apparently new) types/mods/coll lists within RTE, you can just check the first `values_lists` element to get the starting point for the `varno` used in the synthetic var.

marcocitus:
From the new regression tests:

Column orders seems to be incorrect (name and app_id mixed up):
```
-- Test multi-row insert with serial in the partition column
INSERT INTO app_analytics_events (app_id, name)
VALUES (104, 'Wayz'), (105, 'Mynt') RETURNING *;
WARNING:  column "app_id" is of type integer but expression is of type text
HINT:  You will need to rewrite or cast the expression.
CONTEXT:  while executing command on localhost:57638
```

Serial not evaluated on coordinator for reference tables:
```
-- Test multi-row insert with serial in a reference table
INSERT INTO app_analytics_events (app_id, name)
VALUES (104, 'Wayz'), (105, 'Mynt') RETURNING *;
ERROR:  relation "public.app_analytics_events_id_seq" does not exist
```

Additional manual test where shard pruning is possible, but master evaluation required:
```
CREATE TABLE test (x int, y serial);
INSERT INTO test VALUES (3), (4);
WARNING:  relation "public.test_y_seq" does not exist
```

jasonmp85:
Bah I wrongly used `varattno` to reference the ordinal position of the expression I intended in each `values_list`. That's wrong, apparently the tle->`values_list`-position correspondence is direct? I think this might have something to do with deparsing, TBH.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1219,6 +1220,8 @@ RouterInsertTaskList(Query *query, DeferredErrorMessage **planningError)
 
 	Assert(query->commandType == CMD_INSERT);
 
+	NormalizeInsertTargetList(query);
```
marcocitus: If we're doing shard pruning in the planner, then we'll evaluate the expressions in `values_lists` in the executor in `ExecuteMasterEvaluableFunctions`, but during deparsing we'll use the unevaluated expressions from the `rowValuesLists`.

If we're doing shard pruning in the executor, then we're already post-`ExecuteMasterEvaluableFunctions` at this point. `ExecuteMasterEvaluableFunctions` will incorrectly evaluate the target list entries once, and then move the resulting constants to the `values_lists`.

jasonmp85: Yeah I realized that right after I pushed this up. I still am not 100% certain what you're describing is entirely accurate (largely due to the use of a lot of in-place evaluation in this area) but it would certainly be much clearer to pull out the normalize method I added (assuming it is changed to be correct) and call it before doing `ExecuteMasterEvaluableFunctions`.

marcocitus: You'll also want to be careful that during planning time we work with a rewritten query (with constant expressions evaluated), while during execution time we work with the original query (with expressions evaluated by our own logic).

marcocitus:
> Bah I wrongly used varattno to reference the ordinal position of the expression I intended in each values_list. That's wrong, apparently the tle->values_list-position correspondence is direct? I think this might have something to do with deparsing, TBH.

I suffered quite a bit on this change yesterday ;)

Yes, deparsing just spits out the target list in the order in which they appear and the values_lists in the order in which they appear, which may not result in a valid query since there may be more target list items than values. Hence my slightly inefficient approach of reconstructing the lists in target list order.

src/backend/distributed/planner/multi_router_planner.c
jasonmp85: I think there are several places here the distinction between `originalQuery` and the planned one are blurred and it's leading to some confusion, i.e. I think I removed some code thinking there were deep references from the `Task` values lists into the actual values of a `Query`, such that evaluating the latter (in place) would update the former. If that ever held, it certainly doesn't anymore.

Basically I think either (a) the bit that I added to evaluate `values_lists` expressions in `ExecuteMasterEvaluableFunctions` should be removed, or (b) `values_lists` should be replaced with `NIL` before that is called. In addition, we should just evaluate the values lists for each task separately.

So:

  * Fix normalization such that `Var` order in the target list matches 1—1 the values in `values_list` (either by injecting the non-`Var` nodes where they belong into `values_lists` entries or by putting our reconstructed `Var`-based `TLE` on the end of the `TLE`)
  * Ensure master evaluation does not evaluate the `values_lists` in the query
  * Instead directly evaluate the `values_lists` that we need to, as we are executing each row

I suppose this improvement could be left until after release…

jasonmp85:
> Hence my slightly inefficient approach of reconstructing the lists in target list order

I think something similar to what happens in `rewriteTargetListIU` is what we ultimately need…

```c
	/*
	 * We process the normal (non-junk) attributes by scanning the input tlist
	 * once and transferring TLEs into an array, then scanning the array to
	 * build an output tlist.  This avoids O(N^2) behavior for large numbers
	 * of attributes.
	 *
	 * Junk attributes are tossed into a separate list during the same tlist
	 * scan, then appended to the reconstructed tlist.
	 */
```

Basically `s/Junk/non-Var-based/` in the above and we have our scenario. We want to partition the TLE into two TLEs, preserving order. Those which were originally `Var`s (and thus already match `values_lists` list elements) and those which aren't (and thus need to have new `values_lists` entries added and a `Var` constructed). At the end of the algorithm, we append the latter list to the former and we're good. I thought I could get away with `attno` but that was wrong.

src/backend/distributed/planner/multi_router_planner.c
marcocitus: This is why I settled on a hack.  :)

jasonmp85:
Hm, and looking back at that loop, I think it's `n(col^2 * inserted_rows)` in the case of having absolutely no `DEFAULT`/omitted values to deal with?

jasonmp85:
> Yes, deparsing just spits out the target list in the order in which they appear and the values_lists in the order in which they appear.

Huh. What if we just slap a sort in that normalization thing to sort the target list by the `attrno` within each `Var`?

marcocitus:
> Huh. What if we just slap a sort in that normalization thing to sort the target list by the attrno within each Var?

That could make sense, since it would be cheaper than rebuilding the `values_lists` after a certain number of rows. Not sure if the target list needs to match anything else (e.g. `RETURNING`?).

marcocitus:
These changes have largely been incorporated into #1621

PR #1532, opened 2017-08-02
Subquery pushdown on reference tables

velioglu:
Support reference tables with subquery pushdown. 

In order to use reference tables with subqueries, equality between any column of the reference table and partition column of hash distributed table must exist. Beside, reference tables can not be used at the outer part of left/right/full outer joins and as part of any set operation.

Issue #1204, opened 2017-02-03
Expand SQL coverage to pushdown all multi-shard queries with colocated tables and subselects


ozgune:
Citus currently provides full SQL coverage for the multi-tenant use case.

For the operational analytics use case, when the user creates multiple colocated tables and has complex subselects, we provide them with rules in how they need to write their queries. This issue tracks expanding our SQL coverage to include all SQL queries that touch colocated tables and subselects.

We're tracking example customer queries in two internal documents (Raw Queries: Expanding SQL Coverage, Summary: Expanding SQL Coverage). I'm opening this issue to track our progress on this task.

onderkalaci:
@ozgune We only support co-location for hash distributed tables. Is the scope of this issue hash-distributed tables? In other words, existing logic allows range distributed tables and what is our goal for it?

onderkalaci:
As we're about to start the implementation, I'd like to list the items we need to implement:

- [ ] Improve physical planner to use INSERT ... SELECT logic
  - [ ] Use the logic in the prototype [branch](https://github.com/citusdata/citus/tree/subquery_pushdown_proper)
  - [ ] Use the improved shard pruning logic in the other branch (see #1245)
  - [ ] Use joininfo in restrictions (see #1244)
  - [ ] Also ensure that these changes do not break anything in router planner given that we're adding a parameter to restrict information.

- [ ]  Subquery pushdown flag
    - [ ] We should remove the subquery pushdown flag. Basically, we should have a mechanism to fall back to regular multi-planner if the subquery is not pushable. 
         - [ ] This is mostly important for single table repartition subqueries to work. But, also consider the interactions with views as well.
- [ ] Unsupported Table Combinations / Having multiple range table entries in the top level subqueries
    - [ ] Currently, subquery pushdown disallows JOINing subqueries with relations. Similarly, disallows joining more than two tables/subqueries as well. `ErrorIfUnsupportedTableCombination()` seems completely irrelevant with the new approach. We should be able to JOIN any number of tables/subqueries.
    - [ ] In the existing implementation, we disallow having more than one target list entry in the top level subquery. See the lines `subqueryRangeTableEntry = (RangeTblEntry *) linitial(subqueryEntryList);` in the logical planner. We should definitely relax this constraint and allow multiple range table entries in the top level subquery.

- [ ] Subqueries without aggregates/GROUP bys
[...]

PR #1628, opened 2017-08-25
Expand subquery pushdown for reference tables

onderkalaci:
With #1532, we've enabled reference tables on subquery 
pushdown. However, soon we've realized that the 
implementation had a notable restriction such that the user
is required to have join reference tables with hash tables
on the partition keys. That limitation is not related to
correctness. 

This change relaxes the restrictions put on the reference
tables with subquery pushdown.

With this commit, we allow joining hash distributed tables
with reference tables using non-partition keys. We also allow
non-equi joins in that case.

### DIFF
--- a/src/backend/distributed/planner/multi_logical_optimizer.c
+++ b/src/backend/distributed/planner/multi_logical_optimizer.c
@@ -2880,6 +2880,17 @@ FindReferencedTableColumn(Expr *columnExpression, List *parentQueryList, Query *
 		return;
 	}
 
+	/*
+	 * We currently don't support finding partition keys in the subqueries
+	 * that references from outer subqueries. For example, in corrolated
+	 * subqueries in WHERE clause, we don't support use of partition keys
+	 * in the subquery that is referred from the outer query.
+	 */
+	if (candidateColumn->varlevelsup > 0)
+	{
+		return;
+	}
+
 	rangeTableEntryIndex = candidateColumn->varno - 1;
 	rangeTableEntry = list_nth(rangetableList, rangeTableEntryIndex);
 
@@ -3096,7 +3107,14 @@ PartitionColumnOpExpressionList(Query *query)
 		rangeTableEntryIndex = candidatePartitionColumn->varno - 1;
 		rangeTableEntry = list_nth(rangetableList, rangeTableEntryIndex);
 
-		Assert(rangeTableEntry->rtekind == RTE_RELATION);
+		/*
+		 * We currently don't support checking for equality when user refers
+		 * to a column from the JOIN instead of the relation.
+		 */
+		if (rangeTableEntry->rtekind != RTE_RELATION)
+		{
+			continue;
+		}
 
 		relationId = rangeTableEntry->relid;
 		partitionColumn = DistPartitionKey(relationId);
--- a/src/backend/distributed/planner/multi_logical_planner.c
+++ b/src/backend/distributed/planner/multi_logical_planner.c
@@ -31,6 +31,7 @@
 #include "distributed/worker_protocol.h"
 #include "nodes/makefuncs.h"
 #include "nodes/nodeFuncs.h"
+#include "nodes/relation.h"
 #include "optimizer/clauses.h"
 #include "optimizer/prep.h"
 #include "optimizer/tlist.h"
@@ -69,6 +70,8 @@ static DeferredErrorMessage * DeferErrorIfUnsupportedSubqueryPushdown(Query *
 																	  PlannerRestrictionContext
 																	  *
 																	  plannerRestrictionContext);
+static DeferredErrorMessage * DeferErrorIfUnsupportedSublinkAndReferenceTable(
+	Query *queryTree);
 static DeferredErrorMessage * DeferErrorIfUnsupportedFilters(Query *subquery);
 static bool EqualOpExpressionLists(List *firstOpExpressionList,
 								   List *secondOpExpressionList);
@@ -87,6 +90,7 @@ static MultiNode * MultiPlanTree(Query *queryTree);
 static void ErrorIfQueryNotSupported(Query *queryTree);
 static bool HasUnsupportedReferenceTableJoin(
 	PlannerRestrictionContext *plannerRestrictionContext);
+static bool ShouldRecurseForReferenceTableJoinChecks(RelOptInfo *relOptInfo);
 static bool HasUnsupportedJoinWalker(Node *node, void *context);
 static bool ErrorHintRequired(const char *errorHint, Query *queryTree);
 static DeferredErrorMessage * DeferErrorIfUnsupportedSubqueryRepartition(Query *
@@ -96,8 +100,8 @@ static bool HasOuterJoin(Query *queryTree);
 static bool HasOuterJoinWalker(Node *node, void *maxJoinLevel);
 static bool HasComplexJoinOrder(Query *queryTree);
 static bool HasComplexRangeTableType(Query *queryTree);
-static bool RelationInfoHasReferenceTable(PlannerInfo *plannerInfo,
-										  RelOptInfo *relationInfo);
+static bool RelationInfoContainsReferenceTable(PlannerInfo *plannerInfo,
+											   RelOptInfo *relationInfo);
 static void ValidateClauseList(List *clauseList);
 static void ValidateSubqueryPushdownClauseList(List *clauseList);
 static bool ExtractFromExpressionWalker(Node *node,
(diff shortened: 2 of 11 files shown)


onderkalaci:
@marcocitus It's ready for the review. The commits are a bit granular to let you review easier. I'll adjust them to be more meaningful after the review.

marcocitus:
Does the outer join check sometimes trigger too early?

```
CREATE TABLE dist (x int, y int, z int);
CREATE TABLE coloc (x int, y int, z int);
CREATE TABLE ref (a int, b int, z int);

SELECT create_distributed_table('dist','x');
SELECT create_distributed_table('coloc','x');

SELECT * FROM (SELECT x, max(dist.y) FROM dist LEFT JOIN coloc USING (x) LEFT JOIN ref ON (a < x) GROUP BY x) u;
ERROR:  cannot pushdown the subquery
DETAIL:  There exist a reference table in the outer part of the outer join
```

It seems the reference table is the most inner part.

marcocitus:
Not this PR's fault, but the warnings get a bit weird:
```
postgres=# SELECT * FROM dist WHERE x = (SELECT a FROM ref WHERE b > 63);
WARNING:  more than one row returned by a subquery used as an expression
....
the same warning 30 times
....
WARNING:  more than one row returned by a subquery used as an expression
```

onderkalaci:
> Not this PR's fault, but the warnings get a bit weird:

Postgres also returns the same error on non-distributed tables. That's why we didn't spend much time on it.

marcocitus:
> Postgres also returns the same error on non-distributed tables. That's why we didn't spend much time on it.

I'm not suggesting fixing it, just something I ran into.

marcocitus:
Is this something we could support?

```
postgres=# SELECT * FROM dist LEFT JOIN ref USING (z) WHERE x IN (SELECT x FROM coloc WHERE y > 63);
ERROR:  cannot pushdown the subquery
DETAIL:  Reference tables are not allowed in FROM clause when the query has subqueries in WHERE clause
```

jasonmp85:
> I'm not suggesting fixing it, just something I ran into.

Can we slap a limit on the clause inside the parenthesis to fix this?

onderkalaci:
> Is this something we could support?

Yes, in fact these type of queries (i.e., hash distributed tables have equi join), we can support. However, that seems hard to implement. Do you think is that something we should support?

marcocitus:
Apart from the `dist LEFT JOIN coloc .. LEFT JOIN ref` issue this all looks good to me.

onderkalaci:
> Can we slap a limit on the clause inside the parenthesis to fix this?


For Postgres yes, but Citus doesn't support pushing down `LIMIT`s in the subqueries 😄

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1479,16 +1531,61 @@ HasUnsupportedReferenceTableJoin(PlannerRestrictionContext *plannerRestrictionCo
...
+ *
  * Since relation ids of relationInfo indexes to the range table entry list of
  * planner info, planner info is also passed.
+ *
+ * Join relation infos are ignored since we're already going to process each of them
+ * seperately.
  */
 static bool
-RelationInfoHasReferenceTable(PlannerInfo *plannerInfo, RelOptInfo *relationInfo)
+RelationInfoContainsReferenceTable(PlannerInfo *plannerInfo, RelOptInfo *relationInfo)
```
marcocitus: I don't see a big difference between `Has` and `Contains` here. Maybe we could be more descriptive about the problem we're trying to address, along the lines of `RelationInfoResultIsSharded`.

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1479,16 +1531,61 @@ HasUnsupportedReferenceTableJoin(PlannerRestrictionContext *plannerRestrictionCo
...
+	}
+
+	/*
+	 * Note that we treat the same query where relations appear in subqueries
+	 * differently. (i.e., use SELECT * FROM r; instead of r)
+	 *
+	 * In that case, to relax some restrictions, we do the following optimization:
+	 * If the subplan (i.e., plannerInfo corresponding to the subquery) contains any
+	 * joins, we skip reference table checks keeping in mind that the join is already
+	 * going to be processed seperately. This optimization should suffice for many
```
marcocitus: I feel like these `if` statements don't really belong in this function. The caller should decide whether or not to check a `RelationInfo` for reference tables and that's where the reasons why should be documented, otherwise code gets obscure.

Maybe in `HasUnsupportedReferenceTableJoin` we can do something like:
```
if (IsJoinRelationInfo(outerrel))
{
    /* description of why we skip this */
    continue;
}

if (IsSimpleSubquery(outerrel))
{
    /* description of why we skip this */
    continue;
}

if (joinType == JOIN_SEMI || joinType == JOIN_ANTI || joinType == JOIN_LEFT)
{
    if (!RelationInfoResultIsSharded(plannerInfo, outerrel))
    {
        return true;
    }
}
````

File: src/backend/distributed/planner/multi_logical_optimizer.c [hunk 3]
```diff
@@ -2880,6 +2880,12 @@ FindReferencedTableColumn(Expr *columnExpression, List *parentQueryList, Query *
 		return;
 	}
 
+	/* this function doesn't support finding the RTEs above this query */
+	if (candidateColumn->varlevelsup > 0)
+	{
+		return;
```
marcocitus: I would appreciate a slightly more elaborate comment on what the effect of this is.

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -3096,7 +3102,11 @@ PartitionColumnOpExpressionList(Query *query)
 		rangeTableEntryIndex = candidatePartitionColumn->varno - 1;
 		rangeTableEntry = list_nth(rangetableList, rangeTableEntryIndex);
 
-		Assert(rangeTableEntry->rtekind == RTE_RELATION);
+		/* when user refers to a column from the JOIN instead of the relation */
```
marcocitus: I can see that the user did, but why can we skip it? :)

marcocitus:
> For Postgres yes, but Citus doesn't support pushing down LIMITs in the subqueries

Is that a restriction we can easily lift on subqueries in the WHERE clause?

src/backend/distributed/planner/multi_logical_optimizer.c
onderkalaci: updated the comment

src/backend/distributed/planner/multi_logical_optimizer.c
onderkalaci: Good point 😄  

Clarified the comment. But basically its not very trivial to implement and the code path is obsolete that we should remove (see #1204 sub-item: `Filters on the LEAF queries`).

So, I didn't find it very useful to fix that for now.

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: These all make sense, I started in that way but somehow ended up with the current look. The trickier part was `FULL JOIN` where the code becomes a bit complicated.

Anyway, I'll try to shape it better.

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: After all the cleanup, I still would like to stick to `RelationInfoContainsReferenceTable()` since it makes more sense now?

onderkalaci:
> Edit: Probably not, unless we know that the result is the same on all nodes (e.g. there is a deterministic order by).

Yes, we need to apply few restrictions on the `ORDER BY` + `LIMIT` combinations. @mtuncer had some initial [discussions](https://docs.google.com/document/d/1ciGfqQGGE4LbLSG9n_hFCmBJIlDLvw_ceM9mlyMaXO8/edit), which we should continue at some point .

onderkalaci:
Hmm, I thought [these](https://github.com/citusdata/citus/pull/1628/files#diff-cd2db87c163e1bb47309f6ff2d9de731R45), [these](https://github.com/citusdata/citus/pull/1628/files#diff-cd2db87c163e1bb47309f6ff2d9de731R107) 
 and [these](https://github.com/citusdata/citus/pull/1628/files#diff-cd2db87c163e1bb47309f6ff2d9de731R124) tests are one of the simplest to write. (See consecutive tests in each link -- mostly one supported - one unsupported test pattern) 

Are they OK or should I try to write smaller ones?

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -2880,6 +2880,17 @@ FindReferencedTableColumn(Expr *columnExpression, List *parentQueryList, Query *
 		return;
 	}
 
+	/*
+	 * We currently don't support finding partition keys in the subqueries
+	 * that references from outer subqueries. For example, in corrolated
+	 * subqueries in WHERE clause, we don't support use of partition keys
+	 * in the subquery that is refered from the outer query.
```
onderkalaci: type `refered` -> `referred `

onderkalaci:
@marcocitus Just an additional note: I've forced all outer joins including non-subqueries to go through this planner then run the regression tests. The results show that there is a very few minor differences among both planners, which makes me more comfortable about introducing a new `multi shard push-down` planner which is responsible for planning all such queries and removing some codes in the future.

PR #1630, opened 2017-08-28
Bump Citus version

byucesoy:
With this PR, we start development of Citus 7.1. We only changed configure.in file and test outputs, configure file is generated automatically.


metdos:
Looks good 🚢

PR #1633, opened 2017-09-01
Get PostgreSQL 11 build passing

jasonmp85:
Updated tupleDesc->attrs references, added a compatibility header,
brought over a ruleutils file for PostgreSQL 11, and updated tests
which were printing the major version (now they just print whether
the version is above nine or not).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -28,4 +28,5 @@ configure				-whitespace
 src/backend/distributed/utils/citus_outfuncs.c -citus-style
 src/backend/distributed/utils/ruleutils_96.c -citus-style
 src/backend/distributed/utils/ruleutils_10.c -citus-style
+src/backend/distributed/utils/ruleutils_11.c -citus-style
 src/include/distributed/citus_nodes.h -citus-style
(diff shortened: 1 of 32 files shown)


jasonmp85:
🤕 Looks like this has broken again with recent PostgreSQL 11 changes.

pykello:
Getting some compile errors, which are probably because of recent postgres changes:

```commands/transmit.c: In function ‘FileOpenForTransmit’:
commands/transmit.c:169:13: error: too many arguments to function ‘PathNameOpenFile’
  fileDesc = PathNameOpenFile((char *) filename, fileFlags, fileMode);
             ^~~~~~~~~~~~~~~~
In file included from commands/transmit.c:22:0:
/home/hadi/postgres-11/include/server/storage/fd.h:66:13: note: declared here
 extern File PathNameOpenFile(const char *fileName, int fileFlags);
```

jasonmp85:
@pykello the PostgreSQL 11 build is still failing, but it's now because of a broken change related to resizing dynamic memory segments (see [this thread](https://www.postgresql.org/message-id/flat/1002664500.12301802.1471008223422.JavaMail.yahoo%40mail.yahoo.com#1002664500.12301802.1471008223422.JavaMail.yahoo@mail.yahoo.com)). Can you give this another look?

PR #1644, opened 2017-09-12
Set --leak-check to yes to detect memory leaks

byucesoy:
Previously we decided to set this to false because there are good number of
intentional, one-off memory leaks in PostgreSQL. However this option makes us
miss our memory leaks, so we are changing it back to --leak-check=yes.

### DIFF
--- a/src/test/regress/pg_regress_multi.pl
+++ b/src/test/regress/pg_regress_multi.pl
@@ -183,7 +183,7 @@ exec $valgrindPath \\
     --quiet \\
     --suppressions=${postgresSrcdir}/src/tools/valgrind.supp \\
     --trace-children=yes --track-origins=yes --read-var-info=no \\
-    --leak-check=no \\
+    --leak-check=yes \\
     --error-markers=VALGRINDERROR-BEGIN,VALGRINDERROR-END \\
     --log-file=$valgrindLogFile \\
     $bindir/postgres.orig \\


anarazel:
On 2017-09-12 17:55:55 +0000, Burak Yücesoy wrote:
> Previously we decided to set this to false because there are good number of
> intentional, one-off memory leaks in PostgreSQL. However this option makes us
> miss our memory leaks, so we are changing it back to --leak-check=yes.
> You can view, comment on, or merge this pull request online at:

I seriously doubt this'll add much besides more output and longer
runtimes - the majority of detected leaks will not be actual leaks, the
majority of actual leaks will not be detected, because the memory
context still has pointers to them.

byucesoy:
Is there a way to detect memory leaks even if memory context has pointers to them?

anarazel:
On 2017-09-12 18:13:48 +0000, Burak Yücesoy wrote:
> Is there a way to detect memory leaks even if memory context has pointers to them?

Not with valgrind, no. And it'd not be really useful anyway - the memory
hasn't actually leaked, and will be freed after all.  I don't know what
kind of reports you/@ozgune are thinking about, but in most cases the
problem is more that the wrong kind of context is used, or that it's not
reset.

One thing that can be useful is to trigger
MemoryContextStats(TopMemoryContext) from gdb or such.

- Andres

byucesoy:
Thanks for the info.

@metdos @ozgune Do you still want this change?

metdos:
Hey @anarazel, for the context, the intention was not merge this PR, but to run it once to see if we are missing anything with other not usefull error messages. Looking into your comments, it seems that this effort is futile.

byucesoy:
I'm closing this PR because branch itself is enough for testing.

PR #1645, opened 2017-09-12
Copy MultiPlan to avoid reuse across prepared statements

### LINKED ISSUE #1631: clock_timestamp() doesn't return latest timestamp for PREPARE'd statements.
To reproduce:

```SQL
postgres=# PREPARE FOO AS INSERT INTO http_request (
        site_id, ingest_time, url, request_country,
        ip_address, status_code, response_time_msec
    ) VALUES (
        1, clock_timestamp(), 'http://example.com/path', 'USA',
        inet '88.250.10.123', 200, 10
    );

postgres=# execute foo;
INSERT 0 1
postgres=# execute foo;
INSERT 0 1
postgres=# execute foo;
INSERT 0 1

postgres=# select count(*), ingest_time from http_request where site_id = 1 group by ingest_time;
 count |          ingest_time          
-------+-------------------------------
    10 | 2017-08-30 18:27:51.854037+00
(1 row)
```

As per PostgreSQL's definition:
_clock_timestamp() returns the actual current time, and therefore its value changes even within a single SQL command._

As per @anarazel its because we replace/evaluate the functions during the prepare phase, and reuse the value during execute.

marcocitus:
In Citus 7 we replaced the plan serialization/deserialization infrastructure with the extensible node infrastructure. However, for prepared statements the same MultiPlan gets used multiple times. Since we scribble on the plan during execution (function evaluation, placement lookup, ...), this leaves pointers to free-d memory and causes undefined behaviour. This went undetected in regression tests because we do actually copy the plan in assert-enabled builds. After this change we always copy the plan.

Fixes #1631

### DIFF
--- a/src/backend/distributed/planner/multi_planner.c
+++ b/src/backend/distributed/planner/multi_planner.c
@@ -467,7 +467,14 @@ GetMultiPlan(CustomScan *customScan)
 
 	node = CheckNodeCopyAndSerialization(node);
 
-	multiPlan = (MultiPlan *) node;
+	/*
+	 * When using prepared statements the same plan gets reused across
+	 * multiple statements and transactions. We make several modifications
+	 * to the MultiPlan during execution such as assigning task placements
+	 * and evaluating functions and parameters. These changes should not
+	 * persist, so we always work on a copy.
+	 */
+	multiPlan = (MultiPlan *) copyObject(node);
 
 	return multiPlan;
 }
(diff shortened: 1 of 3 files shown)


marcocitus:
Added a regression test just to make sure we cover the original issue, although the problem won't trigger in assert-enabled builds, and usually won't in non-assert-enabled builds. I can reproduce the problem reliably on EC2 using Amazon Linux, PostgreSQL 9.6.5 from PGDG, and citus_70_96 from package cloud.

jasonmp85:
> I can reproduce the problem reliably on EC2 using Amazon Linux, PostgreSQL 9.6.5 from PGDG, and citus_70_96 from package cloud

Nice. Looking into this.

PR #1648, opened 2017-09-12
Wait for I/O to finish after PQputCopyData

### LINKED ISSUE #1647: Copy does not wait for all data to be sent over socket
`PutRemoteCopyData` currently returns immediately if it managed to enqueue some data in the libpq buffer:
https://github.com/citusdata/citus/blob/master/src/backend/distributed/connection/remote_commands.c#L513

For large COPY volumes these buffers can get very big before the data is sent, causing excessive memory usage.

marcocitus:
We previously returned immediately in `PutRemoteCopyData` if `PQputCopyData` returned 1 (success), but 1 only means that the data was added to the internal libpq buffer. Because we continue with the next row immediately after, this could cause a large build-up of data in libpq buffers if the data is sent quickly enough.

This PR fixes the issue by calling `FinishConnectionIO` even if `PQputCopyData` reutrns 1, to wait until the socket becomes writable.

Fixes #1647

### DIFF
--- a/src/backend/distributed/connection/remote_commands.c
+++ b/src/backend/distributed/connection/remote_commands.c
@@ -497,6 +497,7 @@ PutRemoteCopyData(MultiConnection *connection, const char *buffer, int nbytes)
 {
 	PGconn *pgConn = connection->pgConn;
 	int copyState = 0;
+	bool allowInterrupts = true;
 
 	if (PQstatus(pgConn) != CONNECTION_OK)
 	{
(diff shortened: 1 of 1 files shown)


src/backend/distributed/connection/remote_commands.c
```diff
@@ -544,21 +544,12 @@ PutRemoteCopyEnd(MultiConnection *connection, const char *errormsg)
...
+	if (copyState == -1)
 	{
 		return false;
 	}
-	else
-	{
-		bool allowInterrupts = true;
-		return FinishConnectionIO(connection, allowInterrupts);
-	}
+
```
anarazel: "see PutRemoteCopyData()"?

src/backend/distributed/connection/remote_commands.c
```diff
@@ -506,21 +507,19 @@ PutRemoteCopyData(MultiConnection *connection, const char *buffer, int nbytes)
...
+	/*
+	 * PQputCopyData may have queued up part of the data even if it managed
+	 * to send some of it succesfully. We provide back pressure by waiting
+	 * until the socket is writable to prevent the internal libpq buffers
+	 * from growing excessively.
```
anarazel: If you want you could add a comment saying that for performance reasons we could in future reduce the frequency of these pushbacks / tie them to the amount of dirty data. That ought to be a bit faster on systems with fast network (small packets aren't good for performance)...

PR #1652, opened 2017-09-14
Ensure schema exists on reference table creation

### LINKED ISSUE #1526: Creating reference tables and distributed tables in a transaction on a schema leads to self-deadlock
Steps to produce:
```SQL
BEGIN;
CREATE SCHEMA sc;
CREATE TABLE sc.ref(a int);
insert into sc.ref SELECT s FROM generate_series(0, 10000) s;
SELECT create_reference_table('sc.ref');


CREATE TABLE sc.hash(a int);
insert into sc.hash SELECT s FROM generate_series(0, 10000) s;
SELECT create_distributed_table('sc.hash', 'a');

```

I suspect that the reason is that `EnsureSchemaExistsOnAllNodes()` is not called on `create_reference_table()`. I'm assigning @velioglu since he's the author of `EnsureSchemaExistsOnAllNodes()`.

Note that the following works just fine:

```SQL
BEGIN;
CREATE SCHEMA sc;
CREATE TABLE sc.hash(a int);
insert into sc.hash SELECT s FROM generate_series(0, 10000) s;
SELECT create_distributed_table('sc.hash', 'a');

CREATE TABLE sc.ref(a int);
insert into sc.ref SELECT s FROM generate_series(0, 10000) s;
SELECT create_reference_table('sc.ref');
```

onderkalaci:
Fixes #1526.

If the schema doesn't exist on the workers, create it.

It seems worthwhile to backport to 7.0. The bug introduced in this release.

### DIFF
--- a/src/test/regress/expected/multi_create_table.out
+++ b/src/test/regress/expected/multi_create_table.out
@@ -873,5 +873,60 @@ SELECT count(*) FROM tt1;
      6
 (1 row)
 
+-- the goal of the following test is to make sure that
+-- both create_reference_table and create_distributed_table
+-- calls creates the schemas without leading to any deadlocks
+-- first create reference table, then hash distributed table
+BEGIN;
+CREATE SCHEMA sc;
+CREATE TABLE sc.ref(a int);
+insert into sc.ref SELECT s FROM generate_series(0, 100) s;
+SELECT create_reference_table('sc.ref');
+NOTICE:  Copying data from local table...
+ create_reference_table 
+------------------------
+ 
+(1 row)
+
+CREATE TABLE sc.hash(a int);
+insert into sc.hash SELECT s FROM generate_series(0, 100) s;
+SELECT create_distributed_table('sc.hash', 'a');
+NOTICE:  Copying data from local table...
+ create_distributed_table 
+--------------------------
+ 
+(1 row)
+
+COMMIT;
+-- first create hash distributed table, then reference table
+BEGIN;
+CREATE SCHEMA sc2;
+CREATE TABLE sc2.hash(a int);
+insert into sc2.hash SELECT s FROM generate_series(0, 100) s;
+SELECT create_distributed_table('sc2.hash', 'a');
+NOTICE:  Copying data from local table...
+ create_distributed_table 
+--------------------------
+ 
+(1 row)
+
+CREATE TABLE sc2.ref(a int);
+insert into sc2.ref SELECT s FROM generate_series(0, 100) s;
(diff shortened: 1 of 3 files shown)


src/test/regress/expected/multi_create_table.out
```diff
@@ -873,5 +873,30 @@ SELECT count(*) FROM tt1;
...
+CREATE TABLE sc.ref(a int);
+insert into sc.ref SELECT s FROM generate_series(0, 100) s;
+SELECT create_reference_table('sc.ref');
+NOTICE:  Copying data from local table...
+ create_reference_table 
+------------------------
+ 
+(1 row)
+
+CREATE TABLE sc.hash(a int);
```
marcocitus: Does it make sense to put this in a separate transaction?

marcocitus: Oh, and separate schema names :)

onderkalaci: Hmm, I'm a bit confused by your comment. My goal here is the following:

1. `create_reference_table()` creates `sc` on all nodes, in a separate transaction
2. `create_reference_table()` creates placements and just before that issues create schema command on the node
3. `create_distributed_table()` creates (if not exists) `sc` on all nodes, in a separate transaction
4. `create_distributed_table()` creates shard placements and just before that issues create schema command on the node

This patch adds the item `1` since otherwise `2` and `3` deadlocks -- (since 2 happens in the coordinated tx whereas 3 happens seperately). I think we already have a bunch of other tests that tests the creation of the schemas on separate transactions.

(We also already have the tests in the opposite order, but I could add here for completeness if you want)

marcocitus: My thinking was: if we accidentally remove `EnsureSchemaExistsOnAllNodes` in `create_distributed_table` then this test won't catch that because the call to `create_reference_table` already makes sure that the schema exists. For that we should have a `BEGIN; CREATE SCHEMA ...; SELECT create_distributed_table(...); END;` test.

You're right that without the call to `create_distributed_table` we wouldn't notice if the `EnsureSchemaExistsOnAllNodes` in `create_reference_table` had gone missing. I missed that, since it's rather subtle that this doesn't become a problem unless you do a `create_distributed_table`. It might be worth adding a comment on why we're calling `create_distributed_table`.

onderkalaci: > It might be worth adding a comment on why we're calling create_distributed_table.

Makes sense, I'll add the comment

PR #1653, opened 2017-09-15
Allow pushing down GROUP BYs when at least there is one distribution

onderkalaci:
column in the target list

### DIFF
--- a/src/backend/distributed/planner/multi_logical_planner.c
+++ b/src/backend/distributed/planner/multi_logical_planner.c
@@ -1105,12 +1105,14 @@ TargetListOnPartitionColumn(Query *query, List *targetEntryList)
 
 		FindReferencedTableColumn(targetExpression, NIL, query, &relationId, &column);
 
-		/* if the expression belongs to reference table directly returns false */
+		/*
+		 * If the expression belongs to a reference table continue searching for
+		 * other partition keys.
+		 */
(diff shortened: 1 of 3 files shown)


src/test/regress/expected/multi_subquery_complex_reference_clause.out
```diff
@@ -1180,6 +1194,19 @@ SELECT foo.user_id FROM
 ) as foo;
 ERROR:  cannot push down this subquery
 DETAIL:  Distinct on columns without partition column is currently unsupported
+-- supported since the distinct on contains at least one hash table
```
marcocitus: hash -> distributed ?

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1105,12 +1105,14 @@ TargetListOnPartitionColumn(Query *query, List *targetEntryList)
 
 		FindReferencedTableColumn(targetExpression, NIL, query, &relationId, &column);
 
-		/* if the expression belongs to reference table directly returns false */
+		/*
+		 * If the expression belongs to a reference table continue searching for
+		 * other partition keys.
+		 */
 		if (IsDistributedTable(relationId) && PartitionMethod(relationId) ==
```
marcocitus: We should have an `IsReferenceTable` for this.

PR #1656, opened 2017-09-15
Basic usage statistics collection.

pykello:
If compiled with libcurl, adds citus.enable_statistics_collection GUC variable. If statistics collection is enabled,  sends basic usage data to Citus servers every 24 hours. But doesn't send anything in the first 10 minutes so we don't collect stats for CI instances.


src/backend/distributed/utils/statistics_collection.c
```diff
@@ -0,0 +1,155 @@
...
+	uname(&unameData);
+
+	appendStringInfo(fields, "citus_version=%s", CITUS_VERSION);
+	appendStringInfo(fields, "&table_count=" UINT64_FORMAT, roundedDistTableCount);
+	appendStringInfo(fields, "&cluster_size=" UINT64_FORMAT, roundedClusterSize);
+	appendStringInfo(fields, "&worker_node_count=%u", workerNodeCount);
+	appendStringInfo(fields, "&os_name=%s&os_release=%s&hwid=%s",
+					 unameData.sysname, unameData.release, unameData.machine);
+
+	SendHttpPostRequest("http://localhost:5000/collect_stats", fields->data);
```
pykello: I'll replace this with a public https URL.

marcocitus: Is there a timeout on this?

pykello: @marcocitus: just added a timeout of 5s.

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -0,0 +1,157 @@
...
+
+		PG_TRY();
+		{
+			Datum distTableSizeDatum = DirectFunctionCall1(citus_table_size,
+														   ObjectIdGetDatum(relationId));
+			clusterSize += DatumGetInt64(distTableSizeDatum);
+		}
+		PG_CATCH();
+		{
+			FlushErrorState();
```
marcocitus: We should probably make sure we're back on the right memory context after this (in case `citus_table_size` throws an error while the memory context is changed)

src/backend/distributed/utils/maintenanced.c
```diff
@@ -271,6 +276,17 @@ CitusMaintenanceDaemonMain(Datum main_arg)
 		int latchFlags = WL_LATCH_SET | WL_TIMEOUT | WL_POSTMASTER_DEATH;
 		double timeout = 10000.0; /* use this if the deadlock detection is disabled */
 		bool foundDeadlock = false;
+		/* collect statistics every 24 hours */
+		const double statistics_collection_interval = 24.0 * 3600.0;
```
jasonmp85: I think our style would be to `#define` this.

src/backend/distributed/utils/maintenanced.c
```diff
@@ -271,6 +276,17 @@ CitusMaintenanceDaemonMain(Datum main_arg)
...
+		/*
+		 * ... but don't collect in first 30 minutes, to not report short living
+		 * instances like in CI tests.
+		 */
+		const double statistics_collection_start = 30.0 * 60.0;
```
jasonmp85: This seems like a bit of a heuristic. Are we sure we want to keep it? I thought we discussed not doing this.

If we do it, I'd prefer `statsCollectionQuietPeriod`, since `start` can be confusing in time-related code.

src/backend/distributed/utils/maintenanced.c
```diff
@@ -271,6 +276,17 @@ CitusMaintenanceDaemonMain(Datum main_arg)
...
+		time_t current_time = time(NULL);
+		double since_boot = difftime(current_time, boot_time);
```
jasonmp85: `secondsSinceStartup`, maybe?

src/backend/distributed/utils/maintenanced.c
```diff
@@ -271,6 +276,17 @@ CitusMaintenanceDaemonMain(Datum main_arg)
...
+		double since_last_statistics_collection = difftime(current_time,
```
jasonmp85: `secondsSinceLastStatsCollection`. When dealing with time (unless it's a `time_t`), I find putting units in the variables useful (since some functions might use millis and some might use seconds, etc.)

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -0,0 +1,157 @@
...
+#include "access/xact.h"
+#include "citus_version.h"
+#include "distributed/metadata_cache.h"
+#include "distributed/statistics_collection.h"
+#include "distributed/worker_manager.h"
+#include "lib/stringinfo.h"
+
+bool EnableStatisticsCollection = true; /* send basic usage statistics to Citus */
+
+static uint64_t NextPow2(uint64_t n);
```
jasonmp85: Why use `uint64_t` rather than PostgreSQL's types? We don't (AFAIK) use C99 types, instead using the PostgreSQL wrappers for them (`uint64`, `int64`, etc.)

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -0,0 +1,157 @@
...
+ * size rounded to next power of 2, worker node count, and uname data.
+ */
+void
+CollectBasicUsageStatistics(void)
+{
+	List *distributedTables = NIL;
+	uint64_t roundedDistTableCount = 0;
+	uint64_t roundedClusterSize = 0;
+	uint32_t workerNodeCount = 0;
+	const long timeout_seconds = 5;
```
jasonmp85: Probably also should `#define` this.

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -0,0 +1,157 @@
...
+	struct utsname unameData;
```
jasonmp85: Can you zero out this memory? Either using initializer syntax (not the GNU extension, though), or with a memory function?

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -0,0 +1,157 @@
...
+	StringInfo fields = makeStringInfo();
```
jasonmp85: Normally this wouldn't matter, because the memory context would be flushed after each query… but this allocation will leak memory over time, no?

Can you either declare a `StringInfoData` on the stack or whatever (hunt around PostgreSQL: it has a pattern for this), or add a call to `pfree`?

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -0,0 +1,157 @@
...
+
+	StartTransactionCommand();
+	distributedTables = DistributedTableList();
```
jasonmp85: Again I'm a bit worried about allocations performed here. Maybe it would be best to create a temporary memory context during stats collection and then just clear it at the end?

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -0,0 +1,157 @@
...
+	workerNodeCount = ActivePrimaryNodeCount();
+	CommitTransactionCommand();
+
+	uname(&unameData);
+
+	appendStringInfo(fields, "citus_version=%s", CITUS_VERSION);
+	appendStringInfo(fields, "&table_count=" UINT64_FORMAT, roundedDistTableCount);
+	appendStringInfo(fields, "&cluster_size=" UINT64_FORMAT, roundedClusterSize);
+	appendStringInfo(fields, "&worker_node_count=%u", workerNodeCount);
+	appendStringInfo(fields, "&os_name=%s&os_release=%s&hwid=%s",
```
jasonmp85: None of this data is properly escaped… an ampersand in any of these fields would screw up your submission.

But that brings me to a larger point: can you submit the data as JSON instead? There's a PostgreSQL function that lets you pass in one array of field names and another array of field values and get a JSON object back. You can print it to a string and just POST it. You'll need to set the mime type to `application/json` or whatever, though.

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -0,0 +1,157 @@
...
+	curl_global_init(CURL_GLOBAL_DEFAULT);
+	curl = curl_easy_init();
+	if (curl)
+	{
+		curl_easy_setopt(curl, CURLOPT_URL, url);
+		curl_easy_setopt(curl, CURLOPT_POSTFIELDS, postFields);
+		curl_easy_setopt(curl, CURLOPT_TIMEOUT, timeout_seconds);
+
+		curlCode = curl_easy_perform(curl);
+		if (curlCode == CURLE_OK)
```
jasonmp85: Should we log if it's not OK?

In particular, I think we should do something if it's a `400`-class error. It makes sense to ignore `500`s if the server is having problems, and other types of errors (e.g. not being able to contact whatsoever), but if we get back a `400` that means our code is wrong.

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -0,0 +1,157 @@
...
+	return result;
+}
+
+
+/*
+ * SendHttpPostRequest sends a HTTP/HTTPS POST request to the given URL with the
+ * given POST fields.
+ */
+static bool
+SendHttpPostRequest(const char *url, const char *postFields, long timeout_seconds)
```
jasonmp85: What's the point of the return value here?

If the first request fails, that's all we try for an entire day, right? We always update the most recent submission timestamp, even if no data was sent?

pykello:
@jasonmp85 thanks for the review. Please take another look.

jasonmp85:
@pykello can you respond to my comments? If they're code-related, note whether/how you addressed the code change. If they're not code-related, they're probably questions I just wanted to discuss around how this stuff is all implemented/done.

pykello:
Sorry, I had forgot to submit my comments.

config/general.m4
```diff
@@ -0,0 +1,149 @@
+# config/general.m4
```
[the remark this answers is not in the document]
jasonmp85: Apparently not in Citus, no. We might want to rectify that, I guess?

Debian is very strict about file copyright metadata and I found those paragraphs in our `debian/copyright` file.

Our RPM package has no such accommodation. Probably best to add in Citus. Can you open an issue for that?

config/general.m4
```diff
@@ -0,0 +1,149 @@
...
+[
+if test "[$]pgac_arg_to_variable([$1], [$2])" = yes; then
+  m4_default([$7], :)
+m4_ifval([$8],
+[else
+  $8
+])[]dnl
+fi
+])[]dnl
+])# PGAC_ARG_OPTARG
```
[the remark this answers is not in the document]
jasonmp85: Sounds good; just wanted to understand.

Also today I learned `m4` means `macro`, similar to `i18n` or `l10n` for _internationalization_ and _localization_, respectively.

src/backend/distributed/shared_library_init.c
```diff
@@ -785,6 +786,18 @@ RegisterCitusConfigVariables(void)
 		0,
 		NULL, NULL, NULL);
 
+#if HAVE_LIBCURL
```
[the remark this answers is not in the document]
jasonmp85: What were your feelings about this concern?

src/backend/distributed/utils/statistics_collection.c
jasonmp85: 👍

src/backend/distributed/utils/statistics_collection.c
jasonmp85: :+1:

src/backend/distributed/utils/statistics_collection.c
jasonmp85: Sounds good.

src/backend/distributed/utils/statistics_collection.c
jasonmp85: OK, seems like an OK thing to start with (though we run the risk of self-DDOSing every minute if we get a lot of traffic and are down).

src/backend/distributed/utils/statistics_collection.c
jasonmp85: That's good, but for better extensibility later it might be better to avoid the string-building pattern you've gone with. I was thinking of either constructing an ad hoc record and using `composite_to_json` to print, or alternating keys and values in an `ANY` array and using `json_build_object`. Can you refactor to one of those?

src/include/distributed/statistics_collection.h
```diff
@@ -0,0 +1,26 @@
...
+ *
+ *-------------------------------------------------------------------------
+ */
+
+#ifndef STATISTICS_COLLECTION_H
+#define STATISTICS_COLLECTION_H
+
+#if HAVE_LIBCURL
+
+#define STATS_COLLECTION_URL "http://localhost:5000/collect_stats"
```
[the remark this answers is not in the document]
jasonmp85: Sounds good. Can always override it at compile time if we need to experiment.

src/backend/distributed/shared_library_init.c
```diff
@@ -866,3 +887,25 @@ NormalizeWorkerListPath(void)
...
+StatisticsCollectionGucCheckHook(bool *newval, void **extra, GucSource source)
+{
+#if HAVE_LIBCURL
+	return true;
+#else
+
+	/* if libcurl is not installed, only accept false */
+	if (*newval)
+	{
+		ereport(WARNING, (errmsg("cannot set EnableStatisticsCollection to false "
```
jasonmp85: Firstly: Shouldn't this say _true_? Or maybe _can only set EnableStatisticsCollection to false_?

Secondly: I believe you should be using `GUC_check_errcode(ERRCODE_FEATURE_NOT_SUPPORTED)` here to set the error code and `GUC_check_errdetail` to set the detail and then returning `false` rather than raising an error.

Thirdly: Maybe just go with a detail message of _Citus was compiled without libcurl support_? I believe the message may be auto-set based on you returning false.

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -0,0 +1,213 @@
...
+#include "distributed/worker_manager.h"
+#include "lib/stringinfo.h"
+#include "utils/json.h"
+
+bool EnableStatisticsCollection = true; /* send basic usage statistics to Citus */
+
+static uint64 NextPow2(uint64 n);
+static uint64 ClusterSize(List *distributedTableList);
+static bool SendHttpPostJsonRequest(const char *url, const char *postFields, long
+									timeout_seconds);
```
jasonmp85: Still have a snake-case variable name here (should be `timeoutSeconds`).

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -0,0 +1,213 @@
...
+
+/* WarnIfSyncDNS warns if libcurl is compiled with synchronous DNS. */
+void
+WarnIfSyncDNS(void)
+{
+	curl_version_info_data *version_info = curl_version_info(CURLVERSION_NOW);
```
jasonmp85: Should be `versionInfo`

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -0,0 +1,213 @@
...
+	if (!(version_info->features & CURL_VERSION_ASYNCHDNS))
+	{
+		ereport(WARNING, (errmsg("your current libcurl version doesn't support "
+								 "asynchronous DNS, which might cause unexpected "
+								 "delays in database's operation"),
```
jasonmp85: Change to _delays in the operation of Citus_.

jasonmp85: Change to _delays in the operation of Citus_.

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -0,0 +1,213 @@
...
+	StartTransactionCommand();
+	distributedTables = DistributedTableList();
+	roundedDistTableCount = NextPow2(list_length(distributedTables));
+	roundedClusterSize = NextPow2(ClusterSize(distributedTables));
+	workerNodeCount = ActivePrimaryNodeCount();
+	CommitTransactionCommand();
+
+	uname(&unameData);
+
+	appendStringInfoString(fields, "{\"citus_version\": ");
```
jasonmp85: I still find this pretty hard to read and worry it will be tough to extend. I was envisioning you'd use `composite_to_json` with a tuple we construct in another method (would be clear pairing of column names and field values, and supports nested composite types), or `json_build_object` (could declare like:

```c
{ key, value,
  key, value,
  key, value }
```

and would be clearer what was being generated)

jasonmp85: After looking over all the options, it's probably best to use `composite_to_json` if we want to have nested fields, but for now let's go with `json_build_object` with an array of our field names and values.

config/general.m4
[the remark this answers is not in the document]
pykello: Opened #1689.

config/general.m4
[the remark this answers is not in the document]
pykello: Cool, thanks. I didn't know what ```m4``` stands for.

src/backend/distributed/shared_library_init.c
pykello: Done.

src/backend/distributed/utils/statistics_collection.c
pykello: Sorry for still missing this. Done.

src/backend/distributed/utils/statistics_collection.c

src/backend/distributed/utils/statistics_collection.c

pykello:
I still need to fix the ```codecov/patch``` tests.

src/backend/distributed/shared_library_init.c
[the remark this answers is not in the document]
pykello: I agree with you. Already made the change.

src/backend/distributed/utils/statistics_collection.c
pykello: Ok. Keeping it this way for now. I'll look into making this more readable, maybe using ```json_build_object```.

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -0,0 +1,215 @@
...
+	appendStringInfo(fields, ",\"table_count\": " UINT64_FORMAT, roundedDistTableCount);
+	appendStringInfo(fields, ",\"cluster_size\": " UINT64_FORMAT, roundedClusterSize);
+	appendStringInfo(fields, ",\"worker_node_count\": %u", workerNodeCount);
+	appendStringInfoString(fields, ",\"os_name\": ");
+	escape_json(fields, unameData.sysname);
+	appendStringInfoString(fields, ",\"os_release\": ");
+	escape_json(fields, unameData.release);
+	appendStringInfoString(fields, ",\"hwid\": ");
+	escape_json(fields, unameData.machine);
+	appendStringInfoString(fields, ",\"api_version\": ");
```
jasonmp85: Oh, I somehow missed this in a previous review.

The version is not part of the request. It's metadata.

There are [various arguments](https://www.troyhunt.com/your-api-versioning-is-wrong-which-is/) about different ways of versioning an API, but I don't think "include it in the request" is usually considered at all.

If you're using e.g. Django, [this library](http://www.django-rest-framework.org/api-guide/versioning/) appears to support many of the common methods. I looped in some of the cloud folk for their opinion on this topic…

Basically a lot of the chatter online is "headers are most semantic because they don't pollute the URL". But they're also touchier, harder to test with e.g curl, more verbose, have issues with badly configured caching proxies/CDNs, etc. I think we should just put a version in our URL as a top-level path component.

Which brings up my final point: the `STATS_COLLECTION_URL` will probably host a version-check API at some point as well. It would be best to `#define` the host name and add a parameter to `SendHttpPostJsonRequest` to allow passing an arbitrary path.

To sum up, in this case I think we should be `POST`ing our JSON to `/v1/usage_reports` (since the thing you're creating is a usage report, so that naming is in line with RESTy practices for creating a new object over HTTP).

jasonmp85:
@pykello I didn't notice the version being sent as part of the JSON before… Can you change the code to have a `v1` as the first URL path component?

Furthermore, I think the `#define` should be for the hostname only; our HTTP methods should accept a path so we can use them to call more than one API eventually. This might also help us in Travis; we could easily set up a mock server at `localhost` and use a `-D` compiler flag to override the define for the hostname.

As far as the coverage goes; I don't think we have a clear picture of how we'll test this in Travis at the moment, so I'm OK passing on that for now. Fix the API stuff and :shipit:.

pykello:
@jasonmp85 can you please take a look at https://github.com/citusdata/citus/pull/1656/commits/7b9ffec7845000d8907f36bc34ac9123e7612d5a?

This:
- Puts version in URL instead of the json payload
- Fixes compilation with ```--without-libcurl``` (```EnableStatisticsCollection``` should be declared and defined even without libcurl)
- Fixes crashing when ```CREATE EXTENSION citus;```. This starts maintenanced while ```CitusHasBeenLoaded()``` returns false, which causes ```DistributedTableList()``` to crash. I added a check to skip statistics collection when Citus has not been loaded yet.

src/backend/distributed/utils/maintenanced.c
```diff
@@ -304,7 +304,12 @@ CitusMaintenanceDaemonMain(Datum main_arg)
 			 secondsSincePrevStatsCollection >= STATISTICS_COLLECTION_RETRY_INTERVAL))
 		{
 #if HAVE_LIBCURL
-			if (EnableStatisticsCollection)
+			bool citusHasBeenLoaded = false;
```
jasonmp85: Is there a strong reason not to move this all the way to the top of the `for` loop, such that if Citus is not loaded we simply `continue`?

jasonmp85: (i.e. is there a requirement that the maintenance loop do something even when Citus hasn't yet been loaded?)

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -25,6 +23,8 @@
 
 bool EnableStatisticsCollection = true; /* send basic usage statistics to Citus */
```
jasonmp85: Oh, heh. Good catch :-P

jasonmp85:
Only two final questions:

  * Can the `CitusLoaded` check occur somewhere higher up? Would be nice to know everything in the loop only runs with Citus loaded
  * Can you change your API path to be a plural noun? I suggested `usage_reports`, but something else could be suitable. Just think _when I `POST` to this endpoint, I am creating a…_ and fill in the blank. _usage report_ made sense to me, but if you have a better name I'd be open to it. I just don't want to have arbitrary verbs as endpoints

src/backend/distributed/utils/maintenanced.c
pykello: I couldn't find anything that we needed in maintenance loop when Citus is not loaded yet. Moved the check to top of the ```for``` loop as you suggested.

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -83,12 +83,10 @@ CollectBasicUsageStatistics(void)
 	escape_json(fields, unameData.release);
 	appendStringInfoString(fields, ",\"hwid\": ");
 	escape_json(fields, unameData.machine);
-	appendStringInfoString(fields, ",\"api_version\": ");
-	escape_json(fields, STATS_COLLECTION_API_VERSION);
 	appendStringInfoString(fields, "}");
 
-	return SendHttpPostJsonRequest(STATS_COLLECTION_URL, fields->data,
-								   HTTP_TIMEOUT_SECONDS);
+	return SendHttpPostJsonRequest(STATS_COLLECTION_HOST "/v1/collect_stats",
```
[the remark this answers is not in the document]
pykello: Makes sense. Done.

src/include/distributed/statistics_collection.h
```diff
@@ -0,0 +1,26 @@
...
+
+#ifndef STATISTICS_COLLECTION_H
+#define STATISTICS_COLLECTION_H
+
+/* Config variables managed via guc.c */
+extern bool EnableStatisticsCollection;
+
+#if HAVE_LIBCURL
+
+#define STATS_COLLECTION_HOST "https://citus-statistics.herokuapp.com"
```
marcocitus: can we get a citusdata.com domain name for this?

pykello: Sure, I'll do this in later commits.

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -0,0 +1,213 @@
...
+ *	  Anonymous reports and statistics collection.
+ *
+ * Copyright (c) 2017, Citus Data, Inc.
+ *
+ *-------------------------------------------------------------------------
+ */
+
+#include "postgres.h"
+
+#include <curl/curl.h>
```
jasonmp85: Per #1693, this needs to be moved inside the `#if` below…

PR #1661, opened 2017-09-18
Trim trailing characters & copy the message returned by `PQerrorMessage()`

### LINKED ISSUE #1660: ReportResultError() unexpectedly crashes
It looks like we've some memory issues around `ReportResultError()`:

```
Core was generated by `postgres: bgworker: Citus Maintenance Daemon: 16385/10  '.
Program terminated with signal 11, Segmentation fault.
#0  ReportResultError (connection=0x1487678, result=0x0, elevel=19) at connection/remote_commands.c:240
240    connection/remote_commands.c: No such file or directory.
Missing separate debuginfos, use: debuginfo-install postgresql96-server-9.6.4-1PGDG.rhel6.x86_64
(gdb) bt
#0  ReportResultError (connection=0x1487678, result=0x0, elevel=19) at connection/remote_commands.c:240
#1  0x00007f436766b972 in BuildGlobalWaitGraph () at transaction/lock_graph.c:154
#2  0x00007f436766ae1a in CheckForDistributedDeadlocks () at transaction/distributed_deadlock_detection.c:123
#3  0x00007f43676745ab in CitusMaintenanceDaemonMain (main_arg=<optimized out>) at utils/maintenanced.c:312
#4  0x0000000000659c0f in StartBackgroundWorker () at bgworker.c:742
#5  0x0000000000664c34 in do_start_bgworker (rw=<optimized out>) at postmaster.c:5612
#6  maybe_start_bgworkers () at postmaster.c:5809
#7  0x0000000000665815 in sigusr1_handler (postgres_signal_arg=<optimized out>) at postmaster.c:4990
#8  <signal handler called>
#9  0x00007f4371125733 in __select_nocancel () at ../sysdeps/unix/syscall-template.S:81
#10 0x000000000046e3da in ServerLoop () at postmaster.c:1683
#11 0x0000000000666946 in PostmasterMain (argc=argc@entry=3, argv=argv@entry=0x13946b0) at postmaster.c:1327
#12 0x000000000046f8a2 in main (argc=3, argv=0x13946b0) at main.c:228
```

Where the segfault happens at [remote_command.c@240](https://github.com/citusdata/citus/blob/master/src/backend/distributed/connection/remote_commands.c#L240)

and the local variables at that point:

```
(gdb) print messagePrimary
$1 = 0x7f4367409182 "connection pointer is NULL\n"
(gdb) print lastNewlineIndex
$2 = 0x7f436740919c "\n"
```

The segfault at that line makes me think of some memory contexts issues.

As I dig into this, I realized that with Postgres 10, the returned string of `PQerrorMessage()` is trimmed and copied via `pchomp()` with this [commit](https://github.com/postgres/postgres/commit/2ed193c904679a533d5e26a27c97119793bcae52). We should probably do that as well.

onderkalaci:
Fixes #1660.

I couldn't find a way to add regression tests that show how it fixes #1660. Any ideas?

### DIFF
--- a/src/backend/distributed/connection/remote_commands.c
+++ b/src/backend/distributed/connection/remote_commands.c
@@ -18,6 +18,7 @@
 #include "lib/stringinfo.h"
 #include "miscadmin.h"
 #include "storage/latch.h"
+#include "utils/palloc.h"
 
 
 /* GUC, determining whether statements sent to remote nodes are logged */
@@ -191,7 +192,7 @@ ReportConnectionError(MultiConnection *connection, int elevel)
 	int nodePort = connection->port;
 
 	ereport(elevel, (errmsg("connection error: %s:%d", nodeName, nodePort),
-					 errdetail("%s", PQerrorMessage(connection->pgConn))));
+					 errdetail("%s", pchomp(PQerrorMessage(connection->pgConn)))));
 }
 
 
@@ -229,16 +230,7 @@ ReportResultError(MultiConnection *connection, PGresult *result, int elevel)
 		 */
 		if (messagePrimary == NULL)
 		{
-			char *lastNewlineIndex = NULL;
-
-			messagePrimary = PQerrorMessage(connection->pgConn);
-			lastNewlineIndex = strrchr(messagePrimary, '\n');
-
-			/* trim trailing newline, if any */
-			if (lastNewlineIndex != NULL)
-			{
-				*lastNewlineIndex = '\0';
-			}
+			messagePrimary = pchomp(PQerrorMessage(connection->pgConn));
 		}
 
 		ereport(elevel, (errcode(sqlState), errmsg("%s", messagePrimary),
@@ -255,6 +247,28 @@ ReportResultError(MultiConnection *connection, PGresult *result, int elevel)
 	}
 	PG_END_TRY();
 }
+
+
+/* *INDENT-OFF* */
+#if (PG_VERSION_NUM < 100000)
+
+/*
+ * Make copy of string with all trailing newline characters removed.
+ */
+char *
+pchomp(const char *in)
+{
+	size_t		n;
+
+	n = strlen(in);
+	while (n > 0 && in[n - 1] == '\n')
+		n--;
+	return pnstrdup(in, n);
+}
+
+#endif
+
+/* *INDENT-ON* */
 
 
 /*
--- a/src/backend/distributed/utils/colocation_utils.c
+++ b/src/backend/distributed/utils/colocation_utils.c
@@ -112,6 +112,9 @@ get_colocated_shard_array(PG_FUNCTION_ARGS)
 	Oid arrayTypeId = OIDOID;
 	int colocatedShardIndex = 0;
 
+	/* sort to get consistent output */
+	colocatedShardList = SortList(colocatedShardList, CompareShardIntervalsById);
+
 	foreach(colocatedShardCell, colocatedShardList)
 	{
 		ShardInterval *colocatedShardInterval = (ShardInterval *) lfirst(
--- a/src/test/regress/expected/multi_colocation_utils.out
+++ b/src/test/regress/expected/multi_colocation_utils.out
@@ -266,39 +266,39 @@ SELECT shards_colocated(1300000, 1300020);
 (1 row)
 
 -- check co-located table list
-SELECT UNNEST(get_colocated_table_array('table1_group1'))::regclass;
+SELECT UNNEST(get_colocated_table_array('table1_group1'))::regclass ORDER BY 1;
     unnest     
 ---------------
- table2_group1
  table1_group1
+ table2_group1
 (2 rows)
 
-SELECT UNNEST(get_colocated_table_array('table5_groupX'))::regclass;
+SELECT UNNEST(get_colocated_table_array('table5_groupX'))::regclass ORDER BY 1;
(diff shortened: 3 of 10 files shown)


src/backend/distributed/connection/remote_commands.c
```diff
@@ -257,6 +252,32 @@ ReportResultError(MultiConnection *connection, PGresult *result, int elevel)
 }
 
 
+#if (PG_VERSION_NUM < 100000)
```
jasonmp85: Can you add the `/* *INDENT-OFF* */` and `/* *INDENT-ON* */` comments around this so `citus_indent` skips them? Re-copy the PostgreSQL code exactly so it's easier to diff in the future. It's fine if it doesn't match our style so long as it's copied from PostgreSQL.

jasonmp85:
Did you figure out exactly why #1660 is happening? I looked over it, and it appears the `result` is `NULL`. The default behavior of `PQerrorMessage` when provided a `NULL` is to use a static string as the error message. That string'll be store in the text area of the program, so attempting to write to it will cause an error… it's _not_ a memory context issue.

Attempting to trim a newline by replacing it with a `\0` is handy when we're guaranteed to be working with writeable memory, but the `pchomp` solution (while slower in that case) is better for the general case because it will copy the unwriteable/text-area memory into a new region.

I think the above is subtle enough to warrant a comment somewhere (or maybe in the commit message).

src/backend/distributed/connection/remote_commands.c
```diff
@@ -191,7 +195,7 @@ ReportConnectionError(MultiConnection *connection, int elevel)
 	int nodePort = connection->port;
 
 	ereport(elevel, (errmsg("connection error: %s:%d", nodeName, nodePort),
-					 errdetail("%s", PQerrorMessage(connection->pgConn))));
+					 errdetail("%s", pchomp(PQerrorMessage(connection->pgConn)))));
```
jasonmp85: FWIW `PQerrorMessage` is also invoked in `MultiClientSendQuery`, `ReportCopyError`, and `StoreErrorMessage`. I'm not sure which of those go on to `ereport` using the string they find, but it's worth checking.

src/test/regress/expected/multi_router_planner.out
```diff
@@ -2152,7 +2152,6 @@ BEGIN;
 INSERT INTO failure_test VALUES (1, 1);
 WARNING:  connection error: localhost:57638
 DETAIL:  no connection to the server
-
```
jasonmp85: This looks so much better.

onderkalaci:
> . That string'll be store in the text area of the program, so attempting to write to it will cause an error… it's not a memory context issue.

Oh, yes. I already realized that. My first impression (while writing #1660) was that it looks like a memory context issue. I should have updated the issue accordingly or mention it here. I'll add a clarifying comment as you suggested.

src/backend/distributed/connection/remote_commands.c
onderkalaci: Makes sense

src/backend/distributed/connection/remote_commands.c
onderkalaci: Yeah, that makes sense.


Actually, I changed all of them. Although the error messages have slightly changes for `StoreErrorMessage ()` [citus_tools], I think this is safer.

Are you OK with that?

onderkalaci:
@jasonmp85 Should be ready for you to take another look. Thanks!

src/backend/distributed/utils/colocation_utils.c
```diff
@@ -112,6 +112,9 @@ get_colocated_shard_array(PG_FUNCTION_ARGS)
 	Oid arrayTypeId = OIDOID;
 	int colocatedShardIndex = 0;
 
+	/* sort to get consistent output */
+	colocatedShardList = SortList(colocatedShardList, CompareShardIntervalsById);
```
jasonmp85: Heh. Is this necessary to do _here_? Or can it be done in SQL in the tests?

src/test/regress/expected/multi_colocation_utils.out
```diff
@@ -266,39 +266,39 @@ SELECT shards_colocated(1300000, 1300020);
 (1 row)
 
 -- check co-located table list
-SELECT UNNEST(get_colocated_table_array('table1_group1'))::regclass;
+SELECT UNNEST(get_colocated_table_array('table1_group1'))::regclass ORDER BY 1;
```
jasonmp85: I have a similar commit (517ac89e312b9e5c7569e3edceebd88883c0716b) in `fix_pg_11`, but I guess I'll merge this and remove my commit, since they're almost identical.

jasonmp85:
For posterity's sake, I realized last night I called this part of the binary the "text" segment, but the text segment is actually the part which contains executable code. I meant `.data`, which contains constants, etc. It's been a long time since my complier/architecture/language courses :-|

src/backend/distributed/connection/remote_commands.c
jasonmp85: Yeah; that all makes sense.

src/backend/distributed/utils/colocation_utils.c
onderkalaci: Could be done with doing `UNNEST` & `ORDER BY`. But this seemed to cover wider queries with a single touch.

Issue #1659, opened 2017-09-18
coordinator node crashes while creating distributed table when there is an older worker node


mtuncer:
Hit this when investigating something else
- build citus version 7.0-2 from label 04fe3f0 (git checkout 04fe3f0)
- prepare a worker node, and install citus
- build up to date citus version  7.0-15  from label c6ec493 (git checkout c6ec493)
- prepare a master node, and install citus
- create a table (create table t1(a int, b int))
- distribute t1 on a (select create_distributed_table('t1', 'a'))

coordinator first gives following error then, crashes with assert
```
ERROR:  function assign_distributed_transaction_id(integer, integer, unknown) does not exist at character 57
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
STATEMENT:  BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED;SELECT assign_distributed_transaction_id(0, 2, '2017-09-18 14:10:06.160085-07');


TRAP: FailedAssertion("!(PQtransactionStatus(connection->pgConn) == PQTRANS_INTRANS)", File: "transaction/remote_transaction.c", Line: 140)
```

Same steps result in just assert when coordinator and worker are placed on different nodes.
assert is thrown  [here](https://github.com/citusdata/citus/blob/master/src/backend/distributed/transaction/remote_transaction.c#L140)

It looks like we are sending two commands there and only checking success state of the first command.

mtuncer:
Proposed fix is

implement a logic in ```FinishRemoteTransactionBegin()``` similar to ```ForgetResults()```. 
i.e. keep looping on ```GetRemoteCommandResult()``` until results from all commands previously sent in ```StartRemoteTransactionBegin()```  are consumed.

something like
```c
while (true)
{
    PGresult *result = GetRemoteCommandResult(connection, raiseErrors);

    if (result == NULL)
    {
        break;
    }

    if (!IsResponseOK(result))
    {
        ReportResultError(connection, result, WARNING);
        MarkRemoteTransactionFailed(connection, raiseErrors);
    }
    else if (transaction->transactionState == REMOTE_TRANS_STARTING)
    {
        transaction->transactionState = REMOTE_TRANS_STARTED;
        transaction->lastSuccessfulSubXact = transaction->lastQueuedSubXact;
    }
    PQclear(result);
}
```

ozgune:
@mtuncer / @onderkalaci -- We seem to have two issues here.

[...]

PR #1662, opened 2017-09-19
Prevent crash when remote transaction start fails

mtuncer:
We sent multiple commands to worker when starting a transaction.
Previously we only checked the result of the first command that
is transaction 'BEGIN' which always succeeds. Any failure on
following commands were not checked.

With this commit, we make sure all command results are checked.
If there is any error we report the first error found.

Fixes : #1659

### DIFF
--- a/src/backend/distributed/connection/remote_commands.c
+++ b/src/backend/distributed/connection/remote_commands.c
@@ -56,6 +56,8 @@ IsResponseOK(PGresult *result)
  *
  * Note that this might require network IO. If that's not acceptable, use
  * NonblockingForgetResults().
+ *
+ * ClearResults is variant of this function which can also raise errors.
  */
 void
 ForgetResults(MultiConnection *connection)
@@ -78,6 +80,52 @@ ForgetResults(MultiConnection *connection)
 		}
 		PQclear(result);
 	}
+}
+
+
+/*
+ * ClearResults clears a connection from pending activity,
+ * returns true if all pending commands return success. It raises
+ * error if raiseErrors flag is set, any command fails and transaction
+ * is marked critical.
+ *
+ * Note that this might require network IO. If that's not acceptable, use
+ * NonblockingForgetResults().
+ */
+bool
+ClearResults(MultiConnection *connection, bool raiseErrors)
+{
+	bool success = true;
+
+	while (true)
+	{
+		PGresult *result = GetRemoteCommandResult(connection, raiseErrors);
+		if (result == NULL)
+		{
+			break;
+		}
+
+		/*
+		 * End any pending copy operation. Transaction will be marked
+		 * as failed by the following part.
+		 */
+		if (PQresultStatus(result) == PGRES_COPY_IN)
+		{
+			PQputCopyEnd(connection->pgConn, NULL);
+		}
+
+		if (!IsResponseOK(result))
+		{
+			ReportResultError(connection, result, WARNING);
+			MarkRemoteTransactionFailed(connection, raiseErrors);
+
+			success = false;
+		}
+
+		PQclear(result);
+	}
+
+	return success;
 }
 
 
--- a/src/backend/distributed/transaction/remote_transaction.c
+++ b/src/backend/distributed/transaction/remote_transaction.c
@@ -115,25 +115,17 @@ void
 FinishRemoteTransactionBegin(struct MultiConnection *connection)
 {
 	RemoteTransaction *transaction = &connection->remoteTransaction;
-	PGresult *result = NULL;
-	const bool raiseErrors = true;
+	bool clearSuccessful = true;
+	bool raiseErrors = true;
 
 	Assert(transaction->transactionState == REMOTE_TRANS_STARTING);
 
-	result = GetRemoteCommandResult(connection, raiseErrors);
-	if (!IsResponseOK(result))
-	{
-		ReportResultError(connection, result, WARNING);
-		MarkRemoteTransactionFailed(connection, raiseErrors);
-	}
-	else
+	clearSuccessful = ClearResults(connection, raiseErrors);
+	if (clearSuccessful)
 	{
 		transaction->transactionState = REMOTE_TRANS_STARTED;
 		transaction->lastSuccessfulSubXact = transaction->lastQueuedSubXact;
 	}
-
-	PQclear(result);
-	ForgetResults(connection);
 
 	if (!transaction->transactionFailed)
 	{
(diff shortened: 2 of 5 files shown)


ozgune:
@mtuncer -- Could we add a regression test to this pull request?

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -115,25 +115,41 @@ void
...
+			transaction->transactionState = REMOTE_TRANS_STARTED;
+			transaction->lastSuccessfulSubXact = transaction->lastQueuedSubXact;
+			break;
+		}
+		else if (result == NULL)
+		{
+			break;
+		}
+
+		if (!IsResponseOK(result) && !failed)
```
marcocitus: I think it would be better to move the error checks in `ForgetResults`, since there doesn't seem to be a case in which we wouldn't want to check for failure there. We may have to add a `raiseErrors` flag.

metdos:
@mtuncer, could you make requested changes by @ozgune and @marcocitus and ping @marcocitus  when the PR is ready for the next round? Thanks.

src/backend/distributed/transaction/remote_transaction.c
mtuncer: per our discussion, will make `ForgetResults()` to return boolean to indicate if anything went wrong when forgetting.

mtuncer:
updated the code, added an odd looking regression tests

@marcocitus it is all yours

src/backend/distributed/connection/remote_commands.c
```diff
@@ -82,6 +82,51 @@ ForgetResults(MultiConnection *connection)
...
+		if (result == NULL)
+		{
+			break;
+		}
+
+		if (PQresultStatus(result) == PGRES_COPY_IN)
+		{
+			PQputCopyEnd(connection->pgConn, NULL);
+
+			/* TODO: mark transaction as failed, once we can. */
```
marcocitus: I guess we're actually doing that now.

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -115,25 +115,18 @@ void
...
 
-	result = GetRemoteCommandResult(connection, raiseErrors);
-	if (!IsResponseOK(result))
-	{
-		ReportResultError(connection, result, WARNING);
-		MarkRemoteTransactionFailed(connection, raiseErrors);
-	}
-	else
+	clearSuccessful = ClearRemainingResults(connection);
+
```
marcocitus: extra whitespace

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -115,25 +115,18 @@ void
...
 	Assert(transaction->transactionState == REMOTE_TRANS_STARTING);
 
-	result = GetRemoteCommandResult(connection, raiseErrors);
-	if (!IsResponseOK(result))
-	{
-		ReportResultError(connection, result, WARNING);
-		MarkRemoteTransactionFailed(connection, raiseErrors);
-	}
-	else
+	clearSuccessful = ClearRemainingResults(connection);
```
marcocitus: It seems like it should actually be called `ClearResults`, since there isn't a remaining here. It's all results.

src/backend/distributed/connection/remote_commands.c
```diff
@@ -82,6 +82,51 @@ ForgetResults(MultiConnection *connection)
...
 /*
+ * ClearRemainingResults clears a connection from pending activity,
+ * returns true if all pending commands return success. It raises
+ * error if any command fails and transaction is deemed critical.
+ *
+ * Note that this might require network IO. If that's not acceptable, use
+ * NonblockingForgetResults().
+ */
+bool
+ClearRemainingResults(MultiConnection *connection)
```
marcocitus: To future-proof this function. What if we add the raiseErrors flag as a parameter?

src/backend/distributed/connection/remote_commands.c
mtuncer: No we don't. Shall we ?

src/backend/distributed/transaction/remote_transaction.c
mtuncer: done

src/backend/distributed/transaction/remote_transaction.c
mtuncer: fixed

src/backend/distributed/connection/remote_commands.c
mtuncer: ok

src/backend/distributed/connection/remote_commands.c
marcocitus: We do in the lines below?

mtuncer: no we don't. It is else if

mtuncer: removed else now

src/test/regress/expected/multi_extension.out
```diff
@@ -329,3 +329,39 @@ FROM
...
+CREATE EXTENSION citus;
+-- drop function assign_distributed_transaction_id
+DELETE FROM pg_depend where objid = 'assign_distributed_transaction_id'::regproc::oid;
+DROP FUNCTION pg_catalog.assign_distributed_transaction_id(initiator_node_identifier integer, transaction_number bigint, transaction_stamp timestamp with time zone);
+\c - - - :master_port
+-- create_distributed_table command should fail
+CREATE TABLE t1(a int, b int);
+SELECT create_distributed_table('t1', 'a');
+ERROR:  replication_factor (2) exceeds number of worker nodes (1)
+HINT:  Add more worker nodes or try again with a lower replication factor.
```
marcocitus: This doesn't actually test whether the BEGIN fails.

mtuncer: correct. It replication factor was 1 in my local test which led to this.
fixed

src/test/regress/sql/multi_extension.sql
```diff
@@ -307,3 +307,40 @@ FROM
...
+\c another
+CREATE EXTENSION citus;
+
+SELECT FROM master_add_node('localhost', :worker_1_port);
+
+\c - - - :worker_1_port
+CREATE EXTENSION citus;
+
+-- drop function assign_distributed_transaction_id
+DELETE FROM pg_depend where objid = 'assign_distributed_transaction_id'::regproc::oid;
```
marcocitus: Might be slightly cleaner to rename the function.

mtuncer: yes it makes  more sense. renamed.

mtuncer: for some reason travis did not like `rename to` command.

marcocitus: you didn't include argument types

PR #1665, opened 2017-09-21
Fix self-deadlock in create_distributed_table caused by long table names

### LINKED ISSUE #1664: In a transaction block, creating a distributed table whose name exceeds 64 characters may create a distributed deadlock
The following code block works as expected in Citus 6.2-4:

<pre>
BEGIN;
CREATE TABLE tmp_a191e16d9d53aeca7616674831a561af_client_session_opened (
    foo text NOT NULL
);
SELECT create_distributed_table('tmp_a191e16d9d53aeca7616674831a561af_client_session_opened', 'foo');
</pre>

In Citus 7.0-15, the same code block results in a transaction that hangs. The distributed deadlock detection code ultimately aborts the transaction. I see the following error on the coordinator node:

<pre>
ERROR:  
CONTEXT:  while executing command on citus_worker_1:5432
</pre>

PostgreSQL truncates table names that exceed 64 characters. Citus does the same for shard names. However, we didn't change any logic in Citus 7 related to name truncation (or adding a hash into the shard name).

The issue here could be related to how Citus manages transactions and connections. The error message above is a symptom and we should look to identify its root cause.

marcocitus:
Fixes #1664

### DIFF
--- a/src/backend/distributed/relay/relay_event_utility.c
+++ b/src/backend/distributed/relay/relay_event_utility.c
@@ -613,11 +613,17 @@ AppendShardIdToName(char **name, uint64 shardId)
 	shardIdAndSeparatorLength = strlen(shardIdAndSeparator);
 
 	/*
-	 * If *name strlen is < (NAMEDATALEN - shardIdAndSeparatorLength),
-	 * it is safe merely to append the separator and shardId.
+	 * If *name strlen is < (NAMEDATALEN - shardIdAndSeparatorLength - 1),
+	 * it is safe merely to append the separator and shardId. We truncate
+	 * down to 63 instead of 64 characters because postgres generates an
+	 * array type named _<relation name truncated to 63 characters> during
+	 * table creation, while postgres solves naming conflicts by adding
+	 * more underscores, the lock it takes on the type name would prevent
+	 * us from creating multiple shard placements of a table with a long
+	 * name concurrently.
 	 */
 
-	if (nameLength < (NAMEDATALEN - shardIdAndSeparatorLength))
+	if (nameLength < (NAMEDATALEN - shardIdAndSeparatorLength - 1))
 	{
 		snprintf(extendedName, NAMEDATALEN, "%s%s", (*name), shardIdAndSeparator);
 	}
@@ -652,7 +658,7 @@ AppendShardIdToName(char **name, uint64 shardId)
 		longNameHash = hash_any((unsigned char *) (*name), nameLength);
 		multiByteClipLength = pg_mbcliplen(*name, nameLength, (NAMEDATALEN -
 															   shardIdAndSeparatorLength -
-															   10));
+															   10 - 1));
 		snprintf(extendedName, NAMEDATALEN, "%.*s%c%.8x%s",
 				 multiByteClipLength, (*name),
 				 SHARD_NAME_SEPARATOR, longNameHash,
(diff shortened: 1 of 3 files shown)


jasonmp85:
Quick question about this review…

What happens if I create and distribute one of these now-impossible tables in Citus 6.2 (or any pre-7.0 version), then upgrade? With these changes to `AppendShardIdToName`, our code will no longer be able to find shards for the table, right?

src/backend/distributed/relay/relay_event_utility.c
```diff
@@ -613,11 +613,17 @@ AppendShardIdToName(char **name, uint64 shardId)
 	shardIdAndSeparatorLength = strlen(shardIdAndSeparator);
 
 	/*
-	 * If *name strlen is < (NAMEDATALEN - shardIdAndSeparatorLength),
-	 * it is safe merely to append the separator and shardId.
+	 * If *name strlen is < (NAMEDATALEN - shardIdAndSeparatorLength - 1),
+	 * it is safe merely to append the separator and shardId. We truncate
+	 * down to 63 instead of 64 characters because postgres generates an
+	 * array type named _<relation name truncated to 63 characters> during
```
jasonmp85: `makeArrayTypeName` appears to be the function that actually does this, by the way. Maybe add a reference to that…

It uses a function called `truncate_identifier` that might be useful to clean up the multi-byte truncation code we added, since it has an ominous comment about _avoid[ing] using `%.*s` here because it can misbehave if the data is not valid in what libc thinks is the prevailing encoding_.

jasonmp85:
(FWIW, while looking into this, I realized the first-class partitioned tables do not trigger a type creation. Presumably they all share a type. If Citus sent a single `CREATE TYPE` statement, then used `CREATE TABLE OF` to create the shards, we could sidestep this issue.

Additionally, PostgreSQL actually _continually prepends underscores so long as the type conflicts with an existing type_. So there's no theoretical limit to how much this might truncate our shard names when creating table names.

I think we might need to create our shards with this new scheme (to ensure the types are created in a way that is not truncated), and then _rename_ them to the old schema in the same transaction (I don't believe the types are renamed during that operation).

jasonmp85:
Nevermind; just checked. `ALTER TABLE RENAME` does indeed rename the relation's composite/array types.

marcocitus:
You're right, this approach might break a lot of existing users.

The only other approach I can think of is to not use the connection-per-placement approach in `create_distributed_table` when the table name has the maximum length. That means a subsequent COPY or CREATE INDEX will somewhat mysteriously error out. A more extreme version would be to just error out when running `create_distributed_table` with the maximum table name length in a transaction block.

jasonmp85:
> That means a subsequent `COPY` or `CREATE INDEX` will somewhat mysteriously error out

Sorry, why would those error out? Do `CREATE INDEX` and `COPY` have similar problems? I can see why `COPY` might, if it creates shards and results in new types, but `CREATE INDEX` shouldn't have any problems, right?

marcocitus:
> Sorry, why would those error out? Do CREATE INDEX and COPY have similar problems? I can see why COPY might, if it creates shards and results in new types, but CREATE INDEX shouldn't have any problems, right?

If multiple shard placements were created over a single connection then there's no way we could do a COPY into all these shards at the same time over that connection (without implementing a new form of multiplexed COPY). DDL currently also requires one connection per placement.

jasonmp85:
How crazy/bad would it be to migrate shards during `ALTER EXTENSION` when users upgrade?

jasonmp85:
Or (sorry!) bring back `shardalias`? 🙈

marcocitus:
> How crazy/bad would it be to migrate shards during ALTER EXTENSION when users upgrade?

Not sure whether we can expect all workers to be reachable during an upgrade, but in principle it seems possible (and a rather big task).

marcocitus:
Closing this PR for the moment, since it's not the right fix.

PR #1668, opened 2017-09-26
Add window function support for SUBQUERY PUSHDOWN and INSERT INTO SELECT

furkansahin:
After merging of this PR, Citus will support window functions partially.
We are using our pushdown ability to implement the window function support. Window functions are a complex feature and we support them only if they have a Partition By on the distribution column.
Partition By on distribution column makes it eligible to push the query down and the same theory applies for the insert select as well.

In the implementation part, we check any subquery with window function if it has a partition by clause on distribution column.
Subquery regression tests:
* Partition By on distribution column without group by for two different syntax. (using Over [Window functions definition] or using WINDOW keyword)
* Partition By on distribution key with a group by on distribution key in the subquery.
* Partition by on distribution key in subquery with a group by on another key in top level query.
* Window function in joins (using the same window function which is partitioned on distribution key)
* Multiple window functions in the same subquery (both are partitioned on the distribution key).
* Multiple window functions in join (both are partitioned on the distribution key).
* Multiple window functions in join (both are partitioned on the distribution key) with a group by at the top level query on the distribution key.
* Group by and Partition by in the sub query on the distribution column and another group by on another column in the top level query.
* Aggregate function in the partition by with the distribution key together with a group by on the distribution key and another group by at the top level query.
* Partition by on a non-distribution key.
* Window functions without partition by but Order by.
* Two window functions on a join which one has partition by on distribution column and the other not.
* Group by has the distribution key but not the window function.
* Creating a window function from a previously defined one on the distribution key and using in a sub query which is in another subquery. The upper level subquery has another window function which has partition by on the distribution key.
* Overriding an unsupported window function and using it in a subquery. (which is again not supported because Postgres does not allow to modify the partition by part of a window function)
* Group by has more columns including the distribution key but also the partition by has the distribution key.
* Group by has more columns including the distribution key but also the partition by has multiple keys including the distribution key.
* Window functions with partition by on distribution column in the subqueries in WHERE.
* Subquery in where with unsupported window functions.
* Window function on the distribution column in an aggregate function (which is not supported as expected).
* A couple of more Window functions in nested subqueries.

The same test scenarios are added to the insert select as well.

### DIFF
--- a/src/backend/distributed/planner/insert_select_planner.c
+++ b/src/backend/distributed/planner/insert_select_planner.c
@@ -776,6 +776,7 @@ MultiTaskRouterSelectQuerySupported(Query *query)
 {
 	List *queryList = NIL;
 	ListCell *queryCell = NULL;
+	StringInfo errorDetail = NULL;
 
 	ExtractQueryWalker((Node *) query, &queryList);
 	foreach(queryCell, queryList)
@@ -797,7 +798,7 @@ MultiTaskRouterSelectQuerySupported(Query *query)
 		if (subquery->limitCount != NULL)
 		{
 			return DeferredError(ERRCODE_FEATURE_NOT_SUPPORTED,
-								 "LIMIT clauses are not allowed in distirbuted INSERT "
+								 "LIMIT clauses are not allowed in distributed INSERT "
 								 "... SELECT queries",
 								 NULL, NULL);
 		}
@@ -811,17 +812,34 @@ MultiTaskRouterSelectQuerySupported(Query *query)
 								 NULL, NULL);
 		}
 
-		/*
-		 * We could potentially support window clauses where the data is partitioned
-		 * over distribution column. For simplicity, we currently do not support window
-		 * clauses at all.
-		 */
-		if (subquery->windowClause != NULL)
+		/* group clause list must include partition column */
+		if (subquery->groupClause)
 		{
-			return DeferredError(ERRCODE_FEATURE_NOT_SUPPORTED,
-								 "window functions are not allowed in distributed "
-								 "INSERT ... SELECT queries",
-								 NULL, NULL);
+			List *groupClauseList = subquery->groupClause;
+			List *targetEntryList = subquery->targetList;
+			List *groupTargetEntryList = GroupTargetEntryList(groupClauseList,
+															  targetEntryList);
+			bool groupOnPartitionColumn = TargetListOnPartitionColumn(subquery,
+																	  groupTargetEntryList);
+			if (!groupOnPartitionColumn)
+			{
+				return DeferredError(ERRCODE_FEATURE_NOT_SUPPORTED,
+									 "Group by list without distribution column is "
+									 "not allowed  in distributed INSERT ... "
+									 "SELECT queries",
+									 NULL, NULL);
+			}
+		}
+
+		/*
+		 * We support window functions when the window function
+		 * is partitioned on distribution column.
+		 */
+		if (subquery->windowClause && !SafeToPushdownWindowFunction(subquery,
+																	&errorDetail))
+		{
+			return DeferredError(ERRCODE_FEATURE_NOT_SUPPORTED, errorDetail->data, NULL,
+								 NULL);
 		}
 
 		if (subquery->setOperations != NULL)
--- a/src/backend/distributed/planner/multi_logical_planner.c
+++ b/src/backend/distributed/planner/multi_logical_planner.c
@@ -83,7 +83,7 @@ static DeferredErrorMessage * DeferErrorIfUnsupportedUnionQuery(Query *queryTree
 																outerMostQueryHasLimit);
 static bool ExtractSetOperationStatmentWalker(Node *node, List **setOperationList);
 static DeferredErrorMessage * DeferErrorIfUnsupportedTableCombination(Query *queryTree);
-static bool TargetListOnPartitionColumn(Query *query, List *targetEntryList);
+static bool WindowPartitionOnDistributionColumn(Query *query);
 static FieldSelect * CompositeFieldRecursive(Expr *expression, Query *query);
 static bool FullCompositeFieldList(List *compositeFieldList);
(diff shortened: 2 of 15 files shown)


metdos:
@furkansahin, could you add a description which explains what we added and what are the limitations?

furkansahin:
@metdos done.

metdos:
> Partition By on distribution column makes it eligible to push the query down and the same theory applies for the insert select as well.

Can we run multi-shard queries if group by on the partition column? We support multi-shard queries in `insert select` queries.

onderkalaci:
> Can we run multi-shard queries if group by on the partition column? We support multi-shard queries in insert select queries.

Yes, we do it for both subquery pushdown and `INSERT/SELECT` queries.

sumedhpathak:
@furkansahin Apart from the regression tests, was there any other manual testing you did for this PR? Could you add that to the review itself? Thanks!

furkansahin:
@sumedhpathak I wrote some small tests when I first started to work on window functions but at the end they evolved to the regression tests. So I don't have their basic versions stored but they are covered in the regression tests. I am editing the PR with a list of the cases we cover in tests.

src/test/regress/expected/multi_subquery_window_functions.out
```diff
@@ -0,0 +1,745 @@
...
+-- Using previously defined supported window function on distribution key
+SELECT * FROM (
+  SELECT
+    user_id, date_trunc('day', time) as time, sum(rank) OVER w2
+  FROM (
+    SELECT DISTINCT
+      user_id as user_id, time, rank() over w1
+    FROM users_table
+    WINDOW
+      w AS (PARTITION BY user_id),
```
onderkalaci: Could you briefly explain the semantics of this usage where you define a window function using another window function. 

For instance, what would be the behavior if there were `ORDER BY` in `w`, or `w1` also includes `user_id` etc? 

 (In general, is the overriding used for changing the `ORDER BY`s or can we also change the `PARTITION BY`)?

src/test/regress/sql/multi_insert_select_window.sql
```diff
@@ -0,0 +1,594 @@
+--
+-- multi subquery window function covers the tests
```
onderkalaci: Need to update the comment, it is copy&paste from the other test.

And also could you make sure that we have the same format for these type of comments that appear in the beginning of the tests? An example from `multi_router_planner.sql`, but we have plenty.

(Please follow the same syntax for the other regression test file added in this PR)

```

-- test router planner functionality for single shard select queries
```

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1146,6 +1177,41 @@ TargetListOnPartitionColumn(Query *query, List *targetEntryList)
 
 
 /*
+ * WindowPartitionOnDistributionColumn checks if window function is partitioned over
+ * distribution column. This method returns False if your window function does not
+ * have a partition by clause or it does not include the distribution column.
+ */
+bool
+WindowPartitionOnDistributionColumn(Query *query)
```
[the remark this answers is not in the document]
jasonmp85: Is this codified somewhere? I've always done that myself, but didn't know we had a preference one way or the other.

If it is useful, I have a super hacked-up topological sort wrapper around `cflow` that can suggest a "good" ordering.

src/backend/distributed/planner/insert_select_planner.c
```diff
@@ -772,17 +772,57 @@ MultiTaskRouterSelectQuerySupported(Query *query)
...
-		if (subquery->windowClause != NULL)
+		if (subquery->windowClause)
 		{
-			return DeferredError(ERRCODE_FEATURE_NOT_SUPPORTED,
-								 "window functions are not allowed in distributed "
-								 "INSERT ... SELECT queries",
-								 NULL, NULL);
+			ListCell *windowClauseCell = NULL;
+			List *windowClauseList = subquery->windowClause;
+			bool hasPartitionClause = true;
```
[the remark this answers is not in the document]
furkansahin: You are right. It was there since the code part was copy-pasted from multi_logical_planner.c and their error out implementation is different.

src/test/regress/expected/multi_subquery_window_functions.out
furkansahin: Citing from [Postgres documentation](https://www.postgresql.org/docs/10/static/sql-select.html)
> If an existing_window_name is specified it must refer to an earlier entry in the WINDOW list; the new window copies its partitioning clause from that entry, as well as its ordering clause if any. In this case the new window cannot specify its own PARTITION BY clause, and it can specify ORDER BY only if the copied window does not have one. The new window always uses its own frame clause; the copied window must not specify a frame clause.

So, if an earlier window function is referenced to create a new one, user is not able to modify partition by or order by clauses if they are already defined.

src/test/regress/expected/multi_insert_select_window.out
```diff
@@ -0,0 +1,965 @@
...
+-- ===================================================================
+TRUNCATE agg_results;
+CREATE VIEW subq AS
+SELECT
+  DISTINCT user_id, rank() OVER (PARTITION BY user_id ORDER BY value_1)
+FROM
+  users_table
+GROUP BY
+  user_id, value_1
+HAVING count(*) > 1;
```
onderkalaci: We should get the statistics & truncate after the first query as well

src/test/regress/expected/multi_subquery_window_functions.out
```diff
@@ -0,0 +1,1049 @@
...
+(10 rows)
+
+-- test with reference table partitioned on columns from both
+SELECT *
+FROM
+(
+    SELECT
+      DISTINCT user_id, it_name, count(id) OVER (PARTITION BY user_id, id)
+    FROM
+      users_table, users_ref_test_table
```
onderkalaci: This query doesn't make sense at all. Could you add `WHERE users_table.value_2 = users_ref_test_table. k_no ` or such.

PR #1670, opened 2017-09-27
Skip relation extension locks on distributed deadlock detection

### LINKED ISSUE #1655: Distributed deadlock detection may produce false positives
I observed a failing Travis test on the master branch. Rerunning it, it passed the tests. Talking with @onderkalaci, we should not see any deadlock on single commands, so this is something we should take a look.

Note that, so far we observed this on;
```sql
ALTER TABLE table_name ADD CONSTRAINT constraint_name UNIQUE/PRIMARY
```

Example outputs from Travis;

```
--- /home/travis/build/citusdata/citus/src/test/regress/expected/multi_alter_table_add_constraints.out	2017-09-15 14:14:39.562676004 +0000
+++ /home/travis/build/citusdata/citus/src/test/regress/results/multi_alter_table_add_constraints.out	2017-09-15 14:17:55.286796601 +0000
@@ -120,25 +120,23 @@
 -- Insert row which will conflict with the next unique constraint command
 INSERT INTO unique_test_table VALUES(1, 'Mehmet');
 -- Can not create constraint since it conflicts with the existing data
 ALTER TABLE unique_test_table ADD CONSTRAINT unn_id UNIQUE(id);
 ERROR:  could not create unique index "unn_id_1450035"
 DETAIL:  Key (id)=(1) is duplicated.
 -- Can create unique constraint over multiple columns which must include
 -- distribution column
 ALTER TABLE unique_test_table ADD CONSTRAINT unn_id_name UNIQUE(id, name);
+ERROR:  canceling the transaction since it has involved in a distributed deadlock
 -- Error out, since tables can not have two rows with same id and name.
 INSERT INTO unique_test_table VALUES(1, 'Mehmet');
-ERROR:  duplicate key value violates unique constraint "unn_id_name_1450035"
-DETAIL:  Key (id, name)=(1, Mehmet) already exists.
-CONTEXT:  while executing command on localhost:57638
```

Another example from @velioglu 
```
ALTER TABLE products ADD CONSTRAINT p_key PRIMARY KEY(name);
ERROR:  cannot create constraint on "products"
DETAIL:  Distributed relations cannot have UNIQUE, EXCLUDE, or PRIMARY KEY constraints that do not include the partition column (with an equality operator if EXCLUDE).
ALTER TABLE products ADD CONSTRAINT p_key PRIMARY KEY(product_no);
NOTICE:  using one-phase commit for distributed DDL commands
HINT:  You can enable two-phase commit for extra safety with: SET citus.multi_shard_commit_protocol TO '2pc'
+ERROR:  canceling the transaction since it has involved in a distributed deadlock
INSERT INTO products VALUES(1, 'product_1', 1);
-- Should error out, since we are trying to add a new row having a value on p_key column
-- conflicting with the existing row.
INSERT INTO products VALUES(1, 'product_1', 1);
-ERROR:  duplicate key value violates unique constraint "p_key_1450001"
-DETAIL:  Key (product_no)=(1) already exists.
-CONTEXT:  while executing command on localhost:57638
ALTER TABLE products DROP CONSTRAINT p_key;
+ERROR:  constraint "p_key" of relation "products" does not exist
```

onderkalaci:
Fixes some parts of #1655 (i.e., fixes issues related to `COPY`)

We should skip if the process blocked on the relation
extension since those locks are hold for a short duration
while the relation is actually extended on the disk and
released as soon as the extension is done. Thus, recording
such waits on our lock graphs could yield detecting wrong
distributed deadlocks.

Assume that two concurrent `COPY` run in parallel. If we were to add the relation extension waits (as it is the case in the `master`), the two `COPY` are very likely to have wait edges to each other due to that.

## Open Question:

* Shall we move `IsProcessWaitingForRelationExtension ()` into `IsProcessWaiting()`, or is it better to have separate functions? To me, having two functions looks better. Any thoughts on this?

## To see that the PR fixes that, one can follow the instructions:

* Create a distributed table 
* Create some indexes (since indexes also requires relation extension locks, this makes the problem more visible)
* Generate a file with random data
* Run many concurrent `COPY`
* Decrease the `deadlock_timeout` to a very low value

SQL:

```SQL
CREATE TABLE test (a int, b text, c text, d text, e text);
 CREATE INDEX i1 ON public.test USING btree (a, b, c, d, e);
CREATE INDEX i2 ON test (a);
CREATE INDEX i3 ON test (a,b);
CREATE INDEX i4 ON test (b, c);
CREATE INDEX i5 ON test (b, d);
 CREATE INDEX i6 ON test (b, e);
CREATE INDEX i7 ON test (c, e);
CREATE INDEX i8 ON test (c,d, e);
CREATE INDEX i9 ON test (d, e);
SELECT create_distributed_table('test','a');

-- hit only few shards to make it more the problem even more visible
insert into test SELECT i%5,md5(random()::text), md5(random()::text), md5(random()::text), md5(random()::text) FROM generate_series(0, 10000) i;

 COPY test TO '/tmp/data.csv' WITH CSV;

ALTER SYSTEM SET deadlock_timeout TO '1ms';
SELECT pg_reload_conf();
```

Run 15 concurrent `COPY`:
```
while true
do
	psql -f f.sql &
	psql -f f.sql &
        ...
	psql -f f.sql &
	wait
done
```
where f.sql: ` COPY test FROM '/tmp/data.csv' WITH CSV;`

### DIFF
--- a/src/backend/distributed/transaction/lock_graph.c
+++ b/src/backend/distributed/transaction/lock_graph.c
@@ -49,6 +49,7 @@ static bool ParseBoolField(PGresult *result, int rowIndex, int colIndex);
 static TimestampTz ParseTimestampTzField(PGresult *result, int rowIndex, int colIndex);
 static void ReturnWaitGraph(WaitGraph *waitGraph, FunctionCallInfo fcinfo);
 static WaitGraph * BuildLocalWaitGraph(void);
+static bool IsProcessWaitingForRelationExtension(PGPROC *proc);
 static void LockLockData(void);
 static void UnlockLockData(void);
 static void AddEdgesForLockWaits(WaitGraph *waitGraph, PGPROC *waitingProc,
@@ -447,6 +448,12 @@ BuildLocalWaitGraph(void)
 			continue;
 		}
 
+		/* skip if the process is blocked for relation extension */
+		if (IsProcessWaitingForRelationExtension(currentProc))
+		{
+			continue;
+		}
+
 		AddProcToVisit(&remaining, currentProc);
 	}
 
@@ -456,6 +463,12 @@ BuildLocalWaitGraph(void)
 
 		/* only blocked processes result in wait edges */
 		if (!IsProcessWaitingForLock(waitingProc))
+		{
+			continue;
+		}
+
+		/* skip if the process is blocked for relation extension */
+		if (IsProcessWaitingForRelationExtension(waitingProc))
 		{
 			continue;
 		}
@@ -476,6 +489,36 @@ BuildLocalWaitGraph(void)
 	UnlockLockData();
 
 	return waitGraph;
+}
+
+
+/*
+ * IsProcessWaitingForRelationExtension returns true if the given PROC
+ * waiting on relation extension lock.
+ *
+ * In general for the purpose of distributed deadlock detection, we should
+ * skip if the process blocked on the relation extension. Those locks are
+ * held for a short duration while the relation is actually extended on
+ * the disk and released as soon as the extension is done, even before the
+ * execution of the command that triggered the extension finishes. Thus,
+ * recording such waits on our lock graphs could yield detecting wrong
+ * distributed deadlocks.
+ */
+static bool
+IsProcessWaitingForRelationExtension(PGPROC *proc)
+{
+	PROCLOCK *waitProcLock = NULL;
+	LOCK *waitLock = NULL;
+
+	if (proc->waitStatus != STATUS_WAITING)
+	{
+		return false;
+	}
+
+	waitProcLock = proc->waitProcLock;
+	waitLock = waitProcLock->tag.myLock;
+
+	return waitLock->tag.locktag_type == LOCKTAG_RELATION_EXTEND;
 }
 
 
@@ -550,9 +593,14 @@ AddEdgesForLockWaits(WaitGraph *waitGraph, PGPROC *waitingProc, PROCStack *remai
 	{
 		PGPROC *currentProc = procLock->tag.myProc;
 
-		/* skip processes from the same lock group and ones that don't conflict */
+		/*
+		 * Skip processes from the same lock group, processes that don't conflict,
+		 * and processes that are waiting on a relation extension lock, which
+		 * will be released shortly.
+		 */
 		if (!IsSameLockGroup(waitingProc, currentProc) &&
-			IsConflictingLockMask(procLock->holdMask, conflictMask))
+			IsConflictingLockMask(procLock->holdMask, conflictMask) &&
+			!IsProcessWaitingForRelationExtension(currentProc))
 		{
 			AddWaitEdge(waitGraph, waitingProc, currentProc, remaining);
 		}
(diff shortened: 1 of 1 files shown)


onderkalaci:
I don't see an easy way of adding test to this PR. I'm open to any suggestion.

metdos:
@saicitus I suppose this was the issue which one of the prospects observed?

metdos:
> I don't see an easy way of adding test to this PR. I'm open to any suggestion.

Not in this PR, but as a general idea. Shall we collect scripts like you added above and make run them in a test suite?

src/backend/distributed/transaction/lock_graph.c
```diff
@@ -447,6 +448,18 @@ BuildLocalWaitGraph(void)
 			continue;
 		}
 
+		/*
+		 * We should skip if the process blocked on the relation extension since
+		 * those locks are hold for a short duration while the relation is actually
```
marcocitus: hold -> held

src/backend/distributed/transaction/lock_graph.c
```diff
@@ -590,9 +641,13 @@ AddEdgesForWaitQueue(WaitGraph *waitGraph, PGPROC *waitingProc, PROCStack *remai
 	{
 		int awaitMask = LOCKBIT_ON(currentProc->waitLockMode);
 
-		/* skip processes from the same lock group and ones that don't conflict */
+		/*
+		 * Skip processes from the same lock group and ones that don't conflict
+		 * and waiting on the relation extension.
```
marcocitus: -> Skip processes from the same lock group, processes that don't conflict, and processes that are waiting on a relation extension lock, which will be released shortly.

src/backend/distributed/transaction/lock_graph.c
```diff
@@ -447,6 +448,18 @@ BuildLocalWaitGraph(void)
 			continue;
 		}
 
+		/*
+		 * We should skip if the process blocked on the relation extension since
+		 * those locks are hold for a short duration while the relation is actually
+		 * extended on the disk and released as soon as the extension is done. Thus,
```
marcocitus: I'd move a bunch of this down in the function comment, so it's easy to find.

jasonmp85:
> I don't see an easy way of adding test to this PR. I'm open to any suggestion.

Make a group of 15 concurrent tests and have them all do copies 😈 ?

src/backend/distributed/transaction/lock_graph.c
onderkalaci: Makes sense!

src/backend/distributed/transaction/lock_graph.c
onderkalaci: Done

onderkalaci:
> Make a group of 15 concurrent tests and have them all do copies 😈 ?

😄  The thing is that even with that (plus setting the `deadlock_timeout` to 1msec) does not always end up with deadlocks :) We should probably need ~50 of them 😄

onderkalaci:
> @saicitus I suppose this was the issue which one of the prospects observed?

That's right

onderkalaci:
> Not in this PR, but as a general idea. Shall we collect scripts like you added above and make run them in a test suite?

I think we should follow the idea of adding scale testing to the test automation (or at least to the testing document and do it manually). Once we do that, I was already planning to propose even higher scale tests than this one. But, in any case, such a test would have caught 2 important issues in 7.0, so I'm more inclined to add that soon in this release.

PR #1684, opened 2017-10-03
Remove separate citus.binary_worker_copy_format regression tests

marcocitus:
We once introduced a `citus.binary_worker_copy_format` flag to use the binary COPY format when shipping files between workers. We don't enable this by default because it's not supported for some data types (e.g. arrays of custom types). Since we didn't have/know a way of changing a setting on the workers, we had a separate set of tests for this with the flag enabled on both master and workers.

Given that are regression tests are getting unbearably slow, I think we should remove these tests and set the flag somewhere in the regular regression tests. This saves around 45 seconds on my desktop.

We could also consider removing or at least disabling the task-tracker-extra tests, which would save another 70 seconds.


anarazel:
On 2017-10-03 15:47:20 +0000, Marco Slot wrote:
> We once introduced a `citus.binary_worker_copy_format` flag to use the binary COPY format when shipping files between workers. We don't enable this by default because it's not supported for some data types (e.g. arrays of custom types). Since we didn't have/know a way of changing a setting on the workers, we had a separate set of tests for this with the flag enabled on both master and workers.

FWIW, this should just use the COPY data load logic and auto-enable
itself with compatible types.

PR #1686, opened 2017-10-04
Add select distinct, and distinct on support

### LINKED ISSUE #625: Add support for DISTINCT queries.
Currently we throw this error:

```
select distinct(column_name) from table_name;
DETAIL:  Distinct clause is currently unsupported
```

Even though this query is supported and returns the same thing:

`SELECT column_name FROM table_name GROUP BY column_name`

### LINKED ISSUE #1307: Support for DISTINCT ON clause on citus
<pre>
SELECT DISTINCT ON (part_col) col1,col2,col3
FROM TABLE;
ERROR:  could not run distributed query with DISTINCT clause
</pre>

This is push-downable, but still it throws an error.
Came up in 2 user conversations. In one of these conversations, **this is the only / major blocker.**

mtuncer:
This work adds select distinct support for real-time queries, distinct on top level query of subqueries, distinct on insert into select queries.

Following syntax is supported
```sql
select distinct column_1, column_2, ... from [distributed_table | reference_table];

select distinct on (partition_column) some_other_columns from [distributed_table];

select distinct on (non_partition_column) some_other_columns from [distributed_table | reference_table];
```

distinct/distinct on clause on subqueries are also supported
```sql
select distinct [subquery_target_columns] from (subquery)
```
Subquery is any subquery supported by Citus.

Distinct clauses with aggregates are also supported, however use of distinct queries with aggregates does not make much sense since there is also group by dictating the distinctness of the query target list.

```sql
select distinct a, sum(b)
from some_table
group by a;
```

Due to definition of ```distinct on``` it is recommended that it is alway used by an order by clause since ordering of rows may change in distributed environment.

```sql
select distinct on(a) a, b
from some_table;
```

will return different results than
```sql
select distinct on(a) a, b
from some_table
order by a, b;
```
Notice that PostgreSQL requires that order by clause includes distinct on clauses in the beginning of order list.

```sql
select distinct on (a,b) a,b,c
from some_table
order by a,b,c;
````
is valid but

```sql
select distinct on (a,b) a,b,c
from some_table
order by a, c, b;
````
 is not.

Distinct targets can be of anything that PostgreSQL allows which include but not limited to columns, constants, expressions, function calls.

distinct/distinct on is also supported in insert into .. select queries. Since this queries require that that insert target list to include partition column in correct location, select distinct has to include partition column. The query would work as expected.

However, if the query contains select distinct on without partition column inside on clause, the query would still work, but it would pull all data to coordinator node to perform distinct operation, then it would distributed data back to worker nodes. Therefore it should be used with caution.

Fixes #1307
Fixes #625 

Testing
- select on a single table on range/append/hash partitioned tables
- select on a single reference table (becomes router plannable)
- select on a co-located joins
- select on hash partitioned table vs reference table
- select on re-partition join on hash tables
- select distinct on top level query in subqueries (with + without join)
- select distinct on inner query in subqueries (with + without join)
- insert into select distinct on single table and co-located joins
- insert into select distinct when distinct is inside inner subquery
- all of above with select distinct on()

### DIFF
--- a/src/backend/distributed/planner/insert_select_planner.c
+++ b/src/backend/distributed/planner/insert_select_planner.c
@@ -30,6 +30,7 @@
 #include "optimizer/clauses.h"
 #include "optimizer/planner.h"
 #include "optimizer/restrictinfo.h"
+#include "optimizer/tlist.h"
 #include "optimizer/var.h"
 #include "parser/parsetree.h"
 #include "parser/parse_coerce.h"
@@ -53,6 +54,7 @@ static DeferredErrorMessage * DistributedInsertSelectSupported(Query *queryTree,
 															   RangeTblEntry *subqueryRte,
 															   bool allReferenceTables);
 static DeferredErrorMessage * MultiTaskRouterSelectQuerySupported(Query *query);
+static bool HasUnsupportedDistinctOn(Query *query);
 static DeferredErrorMessage * InsertPartitionColumnMatchesSelect(Query *query,
 																 RangeTblEntry *insertRte,
 																 RangeTblEntry *
@@ -777,6 +779,7 @@ MultiTaskRouterSelectQuerySupported(Query *query)
 	List *queryList = NIL;
 	ListCell *queryCell = NULL;
 	StringInfo errorDetail = NULL;
+	bool hasUnsupportedDistinctOn = false;
 
 	ExtractQueryWalker((Node *) query, &queryList);
 	foreach(queryCell, queryList)
@@ -865,19 +868,49 @@ MultiTaskRouterSelectQuerySupported(Query *query)
 		}
 
 		/*
-		 * We cannot support DISTINCT ON clauses since it could be on a non-partition column.
-		 * In that case, there is no way that Citus can support this.
+		 * We don't support DISTINCT ON clauses on non-partition columns.
 		 */
-		if (subquery->hasDistinctOn)
+		hasUnsupportedDistinctOn = HasUnsupportedDistinctOn(subquery);
+		if (hasUnsupportedDistinctOn)
 		{
 			return DeferredError(ERRCODE_FEATURE_NOT_SUPPORTED,
-								 "DISTINCT ON clauses are not allowed in distributed "
-								 "INSERT ... SELECT queries",
+								 "DISTINCT ON (non-partition column) clauses are not "
+								 "allowed in distributed INSERT ... SELECT queries",
 								 NULL, NULL);
 		}
 	}
 
 	return NULL;
+}
+
+
+/*
+ * HasUnsupportedDistinctOn returns true if the query has distinct on and
+ * distinct targets do not contain partition column.
+ */
+static bool
+HasUnsupportedDistinctOn(Query *query)
+{
+	ListCell *distinctCell = NULL;
+
+	if (!query->hasDistinctOn)
+	{
+		return false;
+	}
+
+	foreach(distinctCell, query->distinctClause)
+	{
+		SortGroupClause *distinctClause = lfirst(distinctCell);
+		TargetEntry *distinctEntry = get_sortgroupclause_tle(distinctClause,
+															 query->targetList);
+
+		if (IsPartitionColumn(distinctEntry->expr, query))
+		{
+			return false;
+		}
+	}
+
+	return true;
 }
 
 
--- a/src/backend/distributed/planner/multi_logical_optimizer.c
+++ b/src/backend/distributed/planner/multi_logical_optimizer.c
@@ -1298,6 +1298,8 @@ MasterExtendedOpNode(MultiExtendedOp *originalOpNode)
 	masterExtendedOpNode->targetList = newTargetEntryList;
 	masterExtendedOpNode->groupClauseList = originalOpNode->groupClauseList;
 	masterExtendedOpNode->sortClauseList = originalOpNode->sortClauseList;
+	masterExtendedOpNode->distinctClause = originalOpNode->distinctClause;
+	masterExtendedOpNode->hasDistinctOn = originalOpNode->hasDistinctOn;
 	masterExtendedOpNode->limitCount = originalOpNode->limitCount;
 	masterExtendedOpNode->limitOffset = originalOpNode->limitOffset;
 	masterExtendedOpNode->havingQual = newHavingQual;
@@ -1787,6 +1789,7 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
 	WorkerAggregateWalkerContext *walkerContext =
 		palloc0(sizeof(WorkerAggregateWalkerContext));
 	Index nextSortGroupRefIndex = 0;
+	bool queryHasAggregates = false;
 
 	walkerContext->repartitionSubquery = false;
 	walkerContext->expressionList = NIL;
@@ -1826,6 +1829,7 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
 			WorkerAggregateWalker((Node *) originalExpression, walkerContext);
 
 			newExpressionList = walkerContext->expressionList;
+			queryHasAggregates = true;
 		}
 		else
 		{
@@ -1924,6 +1928,15 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
 
 	workerExtendedOpNode = CitusMakeNode(MultiExtendedOp);
 	workerExtendedOpNode->targetList = newTargetEntryList;
+	workerExtendedOpNode->distinctClause = NIL;
+	workerExtendedOpNode->hasDistinctOn = false;
+
+	if (!queryHasAggregates)
+	{
+		workerExtendedOpNode->distinctClause = originalOpNode->distinctClause;
+		workerExtendedOpNode->hasDistinctOn = originalOpNode->hasDistinctOn;
+	}
+
 	workerExtendedOpNode->groupClauseList = groupClauseList;
 
 	/* if we can push down the limit, also set related fields */
@@ -3356,10 +3369,10 @@ WorkerLimitCount(MultiExtendedOp *originalOpNode,
 
 /*
  * WorkerSortClauseList first checks if the given extended node contains a limit
- * that can be pushed down. If it does, the function then checks if we need to
- * add any sorting and grouping clauses to the sort list we push down for the
- * limit. If we do, the function adds these clauses and returns them. Otherwise,
- * the function returns null.
+ * or hasDistinctOn that can be pushed down. If it does, the function then
+ * checks if we need to add any sorting and grouping clauses to the sort list we
+ * push down for the limit. If we do, the function adds these clauses and
+ * returns them. Otherwise, the function returns null.
  */
 static List *
 WorkerSortClauseList(MultiExtendedOp *originalOpNode,
@@ -3370,8 +3383,8 @@ WorkerSortClauseList(MultiExtendedOp *originalOpNode,
 	List *sortClauseList = originalOpNode->sortClauseList;
 	List *targetList = originalOpNode->targetList;
 
-	/* if no limit node, no need to push down sort clauses */
-	if (originalOpNode->limitCount == NULL)
+	/* if no limit node and no hasDistinctOn, no need to push down sort clauses */
+	if (originalOpNode->limitCount == NULL && !originalOpNode->hasDistinctOn)
 	{
 		return NIL;
 	}
--- a/src/backend/distributed/planner/multi_logical_planner.c
+++ b/src/backend/distributed/planner/multi_logical_planner.c
@@ -1805,13 +1805,6 @@ ErrorIfQueryNotSupported(Query *queryTree)
 		errorHint = filterHint;
 	}
 
-	if (queryTree->distinctClause)
-	{
-		preconditionsSatisfied = false;
-		errorMessage = "could not run distributed query with DISTINCT clause";
-		errorHint = filterHint;
-	}
-
 	if (queryTree->groupingSets)
 	{
 		preconditionsSatisfied = false;
@@ -2850,6 +2843,8 @@ MultiExtendedOpNode(Query *queryTree)
 	extendedOpNode->limitCount = queryTree->limitCount;
 	extendedOpNode->limitOffset = queryTree->limitOffset;
 	extendedOpNode->havingQual = queryTree->havingQual;
+	extendedOpNode->distinctClause = queryTree->distinctClause;
+	extendedOpNode->hasDistinctOn = queryTree->hasDistinctOn;
 
 	return extendedOpNode;
 }
--- a/src/backend/distributed/planner/multi_master_planner.c
+++ b/src/backend/distributed/planner/multi_master_planner.c
@@ -30,6 +30,37 @@
 #include "utils/syscache.h"
 
 
+static List * MasterTargetList(List *workerTargetList);
+static PlannedStmt * BuildSelectStatement(Query *masterQuery, List *masterTargetList,
+										  CustomScan *remoteScan);
+static Agg * BuildAggregatePlan(Query *masterQuery, Plan *subPlan);
+static Plan * BuildDistinctPlan(Query *masterQuery, Plan *subPlan);
+
+
+/*
+ * MasterNodeSelectPlan takes in a distributed plan and a custom scan node which
+ * wraps remote part of the plan. This function finds the master node query
+ * structure in the multi plan, and builds the final select plan to execute on
+ * the tuples returned by remote scan on the master node. Note that this select
+ * plan is executed after result files are retrieved from worker nodes and
+ * filled into the tuple store inside provided custom scan.
+ */
+PlannedStmt *
+MasterNodeSelectPlan(MultiPlan *multiPlan, CustomScan *remoteScan)
+{
+	Query *masterQuery = multiPlan->masterQuery;
+	PlannedStmt *masterSelectPlan = NULL;
+
+	Job *workerJob = multiPlan->workerJob;
+	List *workerTargetList = workerJob->jobQuery->targetList;
+	List *masterTargetList = MasterTargetList(workerTargetList);
+
+	masterSelectPlan = BuildSelectStatement(masterQuery, masterTargetList, remoteScan);
+
+	return masterSelectPlan;
+}
+
+
 /*
  * MasterTargetList uses the given worker target list's expressions, and creates
  * a target target list for the master node. This master target list keeps the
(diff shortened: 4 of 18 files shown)

### CODE REVIEW

File: src/backend/distributed/planner/multi_master_planner.c [hunk 1]
```diff
@@ -151,6 +152,58 @@ BuildAggregatePlan(Query *masterQuery, Plan *subPlan)
...
+		Var *column = (Var *) lfirst(columnCell);
+		column->varno = OUTER_VAR;
+	}
+
+	/*
+	 * Create group by plan with HashAggregate if all distinct
+	 * members are hashable, and query does not have distinct on
+	 * and query does not already have sort plan on distinct clauses.
+	 */
+	if (distinctClausesHashable)
```
metdos: In `create_distinct_paths`, PostgreSQL first tries `sort-based implementations of DISTINCT`, if not possible tries `hash-based implementations of DISTINCT`. Do we have a specific reason to try hash-based one first?

src/backend/distributed/planner/multi_master_planner.c
```diff
@@ -151,6 +152,58 @@ BuildAggregatePlan(Query *masterQuery, Plan *subPlan)
...
+		plan = (Plan *) make_agg(aggregateTargetList, NIL, AGG_HASHED,
+								 AGGSPLIT_SIMPLE, distinctClauseCount,
+								 distinctColumnIdArray,
+								 distinctColumnOpArray, NIL, NIL,
+								 rowEstimate, plan);
+	}
+	else
+	{
+		plan = (Plan *) make_sort_from_sortclauses(masterQuery->distinctClause, plan);
+		plan = (Plan *) make_unique_from_distinctclause(plan,
```
metdos: Can we use directly `make_unique_from_sortclauses` here instead of writing our own version `make_unique_from_distinctclause` ?

metdos: Ohh, this `make_unique_from_distinctclause` is exactly same `make_unique_from_sortclauses`? I think we would like to keep original names with PostgreSQL, such as we do in `ruleutils_10.c`

src/backend/distributed/planner/multi_master_planner.c
mtuncer: when I looked at plans created by postgres, they always used HashAgg when distinct is used. Therefore I prioritized HashAgg over Unique/Sort.

src/backend/distributed/planner/multi_master_planner.c
mtuncer: Ok, renamed it back to  ```make_unique_from_sortclauses```

metdos:
@mtuncer Apart from the regression tests, was there any other manual testing you did for this PR? Could you add that to the review itself? Thanks!

src/backend/distributed/planner/multi_master_planner.c
```diff
@@ -151,6 +152,58 @@ BuildAggregatePlan(Query *masterQuery, Plan *subPlan)
...
+
+	/*
+	 * Create group by plan with HashAggregate if all distinct
+	 * members are hashable, and query does not have distinct on
+	 * and query does not already have sort plan on distinct clauses.
+	 */
+	if (distinctClausesHashable)
+	{
+		/* finally create the plan */
+		const long rowEstimate = 10; /* this is taken from postgresql */
```
metdos: Just curious, where did you get this? I couldn't find the place.

src/backend/distributed/planner/multi_master_planner.c
```diff
@@ -203,10 +257,43 @@ BuildSelectStatement(Query *masterQuery, List *masterTargetList, CustomScan *rem
 		topLevelPlan = &remoteScan->scan.plan;
 	}
 
+	/*
+	 * distinct on() requires sort + unique plans. If already has order by
+	 * clause we append distinct clauses to the end of it. Postgresql requires
+	 * that if both distinct on() and order by exists, ordering is first done
```
metdos: This says distinct ordering is done first, but we add it to the end. Isn't this a contradiction, or am I missing something?

src/backend/distributed/planner/multi_master_planner.c
```diff
@@ -234,6 +330,55 @@ BuildSelectStatement(Query *masterQuery, List *masterTargetList, CustomScan *rem
 
 
 /*
+ * make_unique_from_sortclauses creates and returns a unique node
+ * from provided distinct clause list.
+ * The functions is copied from postgresql function
+ * with the same name from optimizer/plan/createplan.c.
+ */
+static Unique *
+make_unique_from_sortclauses(Plan *lefttree, List *distinctList)
```
metdos: Thinking about this a bit more, I think, we should move this to separate file similar to what we did for ruleutils.c.  Source code also says that external use of `make_sort_from_sortclauses` is deprecated, so wherever they move it from` planmain.h`, we can also put it to the same file. What do you think?

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1805,11 +1809,12 @@ ErrorIfQueryNotSupported(Query *queryTree)
 		errorHint = filterHint;
 	}
 
-	if (queryTree->distinctClause)
+	if (queryTree->distinctClause && !VerifyDistinctClauseSupported(queryTree))
```
metdos: Can we refactor this similar to others here such as;
```C
hasUnsupportedDistinctClause = HasUnsupportedDistinctClause(queryTree);
if (hasUnsupportedDistinctClause)
```

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1865,6 +1870,47 @@ ErrorIfQueryNotSupported(Query *queryTree)
...
 
 
+/*
+ * VerifyDistinctClauseSupported returns true is provided query has supported
+ * distinct clause. Only target entries with Var, Const, Aggref types are supported
+ * at the moment.
+ * TODO : function support should be evaluated. immutable or stable functions can
+ * also be supported.
+ *
+ * TODO : Better error message is necessary at the caller.
```
metdos: Should we fix this?

src/test/regress/expected/multi_select_distinct.out
```diff
@@ -0,0 +1,771 @@
...
+ l_orderkey | strpos 
+------------+--------
+          1 |      1
+          2 |      1
+          3 |      1
+          4 |      1
+          5 |      1
+(5 rows)
+
+-- function calls with non-const input are supported
```
metdos: * are not

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1865,6 +1870,47 @@ ErrorIfQueryNotSupported(Query *queryTree)
 }
 
 
+/*
+ * VerifyDistinctClauseSupported returns true is provided query has supported
+ * distinct clause. Only target entries with Var, Const, Aggref types are supported
+ * at the moment.
+ * TODO : function support should be evaluated. immutable or stable functions can
```
metdos: What is supported and what is not supported? Could you be a bit more descriptive?

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1865,6 +1870,47 @@ ErrorIfQueryNotSupported(Query *queryTree)
...
+	{
+		SortGroupClause *distinctGroupClause = (SortGroupClause *) lfirst(distinctCell);
+		Index sortGroupRef = distinctGroupClause->tleSortGroupRef;
+
+		foreach(targetCell, targetList)
+		{
+			TargetEntry *targetEntry = (TargetEntry *) lfirst(targetCell);
+			if (targetEntry->ressortgroupref == sortGroupRef &&
+				!IsA(targetEntry->expr, Var) &&
+				!IsA(targetEntry->expr, Aggref) &&
```
metdos: What does `Aggref ` mean? Do we have an example of this in regression tests?

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1756,6 +1757,9 @@ ErrorIfQueryNotSupported(Query *queryTree)
 						   "equal filter on joining columns.";
 	const char *filterHint = "Consider using an equality filter on the distributed "
 							 "table's partition column.";
+	const char *distinctHint = "Consider using columns in distinct expression, "
```
metdos: Do we advise here to use router planner?

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -1924,6 +1928,13 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
 
 	workerExtendedOpNode = CitusMakeNode(MultiExtendedOp);
 	workerExtendedOpNode->targetList = newTargetEntryList;
+
+	if (!queryHasAggregates && originalOpNode->distinctClause)
```
metdos: 1. If this block returns false, `workerExtendedOpNode->distinctClause` and `workerExtendedOpNode->hasDistinctOn` stays uninitialized.
2. Do we need the `if` block, can't we just assign values from `originalOpNode`? What is the problem with `queryHasAggregates`?
3. Do we need `copyObject ` for copyObject?

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -625,6 +627,8 @@ BuildJobQuery(MultiNode *multiNode, List *dependedJobList)
 	{
 		MultiExtendedOp *extendedOp = (MultiExtendedOp *) linitial(extendedOpNodeList);
 		targetList = copyObject(extendedOp->targetList);
+		distinctClause = copyObject(extendedOp->distinctClause);
```
metdos: Why do we need `copyObject ` here?

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -699,6 +703,9 @@ BuildJobQuery(MultiNode *multiNode, List *dependedJobList)
 	jobQuery->limitCount = limitCount;
 	jobQuery->havingQual = havingQual;
 	jobQuery->hasAggs = contain_agg_clause((Node *) targetList);
+	jobQuery->distinctClause = distinctClause;
+	jobQuery->hasDistinctOn = hasDistinctOn;
+
```
metdos: extra new line

src/test/regress/sql/multi_select_distinct.sql
```diff
@@ -0,0 +1,237 @@
...
+SELECT DISTINCT l_shipmode FROM lineitem_hash_part ORDER BY 1 DESC;
+
+-- distinct with multiple columns 
+SELECT DISTINCT l_orderkey, o_orderdate
+	FROM lineitem_hash_part JOIN orders_hash_part ON (l_orderkey = o_orderkey) 
+	WHERE l_orderkey < 10
+	ORDER BY l_orderkey;
+
+-- distinct on partition column with aggregate
+-- this is the same as non-distinct version due to group by
```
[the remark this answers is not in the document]
metdos: Maybe without 'distinct keyword', or something like that.

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -1458,6 +1470,16 @@ BuildSubqueryJobQuery(MultiNode *multiNode)
 	/* build the where clause list using select predicates */
 	whereClauseList = QuerySelectClauseList(multiNode);
 
+	hasAggregates = contain_agg_clause((Node *) targetList) ||
```
metdos: I think our coding conventions is more like;
```c
if (contain_agg_clause((Node *) targetList) || contain_agg_clause((Node *) havingQual))
{
    hasAggregates = true;
}
```

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1865,6 +1870,47 @@ ErrorIfQueryNotSupported(Query *queryTree)
...
+	Assert(query->distinctClause);
+
+	foreach(distinctCell, distinctList)
+	{
+		SortGroupClause *distinctGroupClause = (SortGroupClause *) lfirst(distinctCell);
+		Index sortGroupRef = distinctGroupClause->tleSortGroupRef;
+
+		foreach(targetCell, targetList)
+		{
+			TargetEntry *targetEntry = (TargetEntry *) lfirst(targetCell);
```
metdos: Can we use `get_sortgroupclause_tle()` here?

src/backend/distributed/planner/multi_master_planner.c
mtuncer: I think I was not clear there.
I did not mean time ordering there. Requirement is that, if there is a distinct on clause, and order by clause. Order by clause MUST contain distinct on clauses in the beginning. Otherwise it would give an error.

Changed the comment a bit.

src/backend/distributed/planner/multi_master_planner.c
mtuncer: I agree, we should have some placeholder for such copy/paste functions. Do you know our existing approach here ? ```ruleutils``` is very specific, we should have more general file perhaps. There is a risk that we may have to differentiate on version numbers.

@marcocitus  may have an idea here.

src/backend/distributed/planner/multi_physical_planner.c
mtuncer: we can remove it

src/backend/distributed/planner/multi_physical_planner.c
mtuncer: removed

src/backend/distributed/planner/multi_physical_planner.c
mtuncer: I was only copying code from below and setting it to a local variable. 
made the suggested change

src/test/regress/sql/multi_select_distinct.sql
```diff
@@ -0,0 +1,237 @@
...
+-- function calls with non-const input are supported
+SELECT DISTINCT l_orderkey, strpos(l_shipmode, 'A') FROM lineitem_hash_part;
+
+-- distinct on partition column
+-- verify counts match with respect to count(distinct)
+CREATE TEMP TABLE temp_orderkeys AS SELECT DISTINCT l_orderkey FROM lineitem_hash_part;
+SELECT COUNT(*) FROM temp_orderkeys;
+SELECT COUNT(DISTINCT l_orderkey) FROM lineitem_hash_part;
+
+SELECT DISTINCT l_partkey FROM lineitem_hash_part WHERE l_orderkey < 5 order by 1;
```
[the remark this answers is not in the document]
mtuncer: yes it is other than additional filter, changed the above query to use partition column

src/test/regress/sql/multi_select_distinct.sql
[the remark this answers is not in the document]
mtuncer: ok

src/test/regress/sql/multi_select_distinct.sql
```diff
@@ -0,0 +1,237 @@
...
+	
+-- explain the query to see actual plan
+EXPLAIN (COSTS FALSE)
+	SELECT DISTINCT l_partkey, count(*)
+		FROM lineitem_hash_part
+		GROUP BY 1 
+		HAVING count(*) > 2
+		ORDER BY 1;
+
+-- following is a bug needs to be fixed
```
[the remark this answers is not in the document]
mtuncer: I found this later in the testing. I need to pay closer look. It should not be difficult. perhaps 0.5 day

src/backend/distributed/planner/multi_master_planner.c
```diff
@@ -203,10 +257,43 @@ BuildSelectStatement(Query *masterQuery, List *masterTargetList, CustomScan *rem
...
+	 */
+	if (masterQuery->hasDistinctOn)
+	{
+		ListCell *distinctCell = NULL;
+		foreach(distinctCell, masterQuery->distinctClause)
+		{
+			SortGroupClause *singleDistinctClause = lfirst(distinctCell);
+			Index sortGroupRef = singleDistinctClause->tleSortGroupRef;
+
+			if (!get_sortgroupref_clause_noerr(sortGroupRef, sortClauseList))
```
metdos: `get_sortgroupref_clause_noerr()` returns a pointer. Shall we do a NULL check instead of `!`?

src/backend/distributed/planner/multi_master_planner.c
```diff
@@ -151,6 +152,58 @@ BuildAggregatePlan(Query *masterQuery, Plan *subPlan)
 }
 
 
+static Plan *
+BuildDistinctPlan(Query *masterQuery, Plan *topLevelPlan)
+{
+	bool distinctClausesHashable = true;
+	List *distinctClauseList = masterQuery->distinctClause;
+	Plan *plan = topLevelPlan;
+	List *aggregateTargetList = copyObject(masterQuery->targetList);
```
metdos: Is `aggregate` a proper prefix here?

src/backend/distributed/planner/multi_master_planner.c
```diff
@@ -151,6 +152,58 @@ BuildAggregatePlan(Query *masterQuery, Plan *subPlan)
 }
 
 
+static Plan *
+BuildDistinctPlan(Query *masterQuery, Plan *topLevelPlan)
+{
+	bool distinctClausesHashable = true;
+	List *distinctClauseList = masterQuery->distinctClause;
+	Plan *plan = topLevelPlan;
```
metdos: I think no need to rename this and introduce another variable name. Shall we call this as subPlan and use it similar to `BuildAggregatePlan`.

src/backend/distributed/planner/insert_select_planner.c
```diff
@@ -882,6 +883,39 @@ MultiTaskRouterSelectQuerySupported(Query *query)
...
+DistinctClauseContainsPartitionColumn(Query *query)
+{
+	ListCell *distinctCell = NULL;
+
+	if (!query->hasDistinctOn)
+	{
+		return false;
+	}
+
+	Assert(query->distinctClause);
```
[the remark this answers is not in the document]
metdos: Then this Assert is unnecessary. Also, are we supposed to check only distinct on clauses here? Then, maybe we can rename the function name.

src/backend/distributed/planner/multi_master_planner.c
```diff
@@ -152,6 +153,71 @@ BuildAggregatePlan(Query *masterQuery, Plan *subPlan)
 
 
 /*
+ * BuildDistinctPlan create an returns plan for distinct. Depending on
```
metdos: creates and returns a plan ...

src/test/regress/sql/multi_select_distinct.sql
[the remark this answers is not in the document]
metdos: Could you spend a bit more time and see how complex the fix? The workaround is just to drop distinct as long as I see, but it would be nice to run it as it is. (Also making sure that it doesn't break something else)

src/backend/distributed/planner/insert_select_planner.c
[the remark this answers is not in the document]
mtuncer: Renaming function to ```HasUnsupportedDistinctOn```

src/test/regress/expected/multi_select_distinct.out
mtuncer: done

src/backend/distributed/planner/multi_logical_optimizer.c
mtuncer: 1. they are initialized to NIL, and false respectively by ```CitusMakeNode()``` few lines above.
2. correct.  It has the same effect. 
3. we can get by without copyObject, removing.

metdos:
> they are initialized to NIL, and false respectively by CitusMakeNode() few lines above.

`CitusMakeNode()` calls `palloc0fast`, do you mean that? I can't see NIL and false.

src/backend/distributed/planner/multi_master_planner.c
```diff
@@ -152,6 +153,71 @@ BuildAggregatePlan(Query *masterQuery, Plan *subPlan)
...
+	List *aggregateTargetList = copyObject(masterQuery->targetList);
+	List *aggregateColumnList = pull_var_clause_default((Node *) aggregateTargetList);
+	ListCell *columnCell = NULL;
+
+	Assert(masterQuery->distinctClause);
+	Assert(!masterQuery->hasDistinctOn);
+
+	/*
+	 * make_agg function requires that columns in aggregate target list
+	 * has valid varno which is OUTER_VAR in our case. We should
```
metdos: I couldn't see this from `make_agg`. How did you make this inference?

metdos: Could this be the reason?
https://github.com/citusdata/citus/blob/e202c51fec5408f4539e4eecb76327285654e508/src/backend/distributed/planner/multi_master_planner.c#L107

src/backend/distributed/planner/multi_master_planner.c
```diff
@@ -152,6 +153,71 @@ BuildAggregatePlan(Query *masterQuery, Plan *subPlan)
...
+
+	/*
+	 * Create group by plan with HashAggregate if all distinct
+	 * members are hashable, and query does not have distinct on
+	 * and query does not already have sort plan on distinct clauses.
+	 */
+	distinctClausesHashable = grouping_is_hashable(distinctClauseList);
+	if (distinctClausesHashable)
+	{
+		/* finally create the plan */
```
metdos: we can drop this comment or move to before the `make_agg`

src/backend/distributed/planner/multi_logical_planner.c

src/backend/distributed/planner/multi_master_planner.c
mtuncer: ok

src/backend/distributed/planner/multi_master_planner.c
mtuncer: dropped the prefix

src/backend/distributed/planner/multi_master_planner.c
mtuncer: removed the line

src/backend/distributed/planner/multi_master_planner.c
mtuncer: ok, replaced with null check

src/backend/distributed/planner/multi_master_planner.c
```diff
@@ -203,10 +257,43 @@ BuildSelectStatement(Query *masterQuery, List *masterTargetList, CustomScan *rem
...
+	}
+	else if (masterQuery->distinctClause)
+	{
+		Plan *distinctPlan = BuildDistinctPlan(masterQuery, topLevelPlan);
+		topLevelPlan = distinctPlan;
+	}
+
 	/* (3) add a sorting plan if needed */
-	if (masterQuery->sortClause)
+	if (sortClauseList)
```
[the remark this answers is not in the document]
mtuncer: done

src/backend/distributed/planner/multi_master_planner.c
```diff
@@ -217,6 +304,15 @@ BuildSelectStatement(Query *masterQuery, List *masterTargetList, CustomScan *rem
 		topLevelPlan = (Plan *) sortPlan;
 	}
 
+	if (masterQuery->hasDistinctOn)
```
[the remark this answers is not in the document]
mtuncer: added

src/test/regress/sql/multi_select_distinct.sql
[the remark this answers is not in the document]
mtuncer: Yes, dropping distinct resolves it

metdos:
Do we have regression tests including reference tables? Could you add some?

metdos:
We are not supporting distinct on subfields of `jsonb` columns. There is an example of this in the product sheet. How hard to support this?

```sql
select distinct on (repo->'id') repo->'id', created_at 
    from github_events 
    order by repo->'id', created_at 
    limit 10;

ERROR:  could not run distributed query with DISTINCT clause on non-column expressions
HINT:  Consider using columns in distinct expression, or using equality filter on the distributed table's partition column.
```

Example data schema and data
https://docs.citusdata.com/en/v7.0/tutorials/real-time-analytics-tutorial.html?highlight=github_events

src/test/regress/expected/multi_select_distinct.out
```diff
@@ -0,0 +1,790 @@
...
+EXPLAIN (COSTS FALSE)
+	SELECT DISTINCT l_partkey, l_suppkey
+		FROM lineitem_hash_part
+		WHERE l_shipmode = 'AIR' AND l_orderkey < 100
+		ORDER BY 1, 2;
+                                             QUERY PLAN                                              
+-----------------------------------------------------------------------------------------------------
+ Sort
+   Sort Key: remote_scan.l_partkey, remote_scan.l_suppkey
+   ->  HashAggregate
```
metdos: Just noting an observation. Here, PostgreSQL picks Sort + Unique, while we pick HashAggregate.

src/backend/distributed/planner/multi_master_planner.c
mtuncer: moving the function into a new file  ```postgres_planning_functions.c```

src/backend/distributed/planner/multi_master_planner.c
mtuncer: ok

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -1924,6 +1928,15 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
 
 	workerExtendedOpNode = CitusMakeNode(MultiExtendedOp);
 	workerExtendedOpNode->targetList = newTargetEntryList;
+	workerExtendedOpNode->distinctClause = NIL;
```
metdos: Very very minor. Feel free to ignore.

I would keep `workerExtendedOpNode->groupClauseList = groupClauseList;` just after `workerExtendedOpNode->targetList = newTargetEntryList;`.

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -3356,10 +3369,10 @@ WorkerLimitCount(MultiExtendedOp *originalOpNode,
 
 /*
  * WorkerSortClauseList first checks if the given extended node contains a limit
- * that can be pushed down. If it does, the function then checks if we need to
- * add any sorting and grouping clauses to the sort list we push down for the
- * limit. If we do, the function adds these clauses and returns them. Otherwise,
- * the function returns null.
+ * or hasDistinctOn that can be pushed down.  If it does, the function then
```
metdos: Minor. Added extra space before `If`

src/backend/distributed/planner/multi_master_planner.c
```diff
@@ -29,6 +29,7 @@
 #include "utils/rel.h"
 #include "utils/syscache.h"
 
+extern Unique * make_unique_from_sortclauses(Plan *lefttree, List *distinctList);
```
metdos: We don't put `extern`s into .c files. Can we move this into `multi_planner.h`

src/backend/distributed/planner/multi_master_planner.c
```diff
@@ -152,6 +153,75 @@ BuildAggregatePlan(Query *masterQuery, Plan *subPlan)
 
 
 /*
+ * BuildDistinctPlan creates an returns plan for distinct. Depending on
```
metdos: Minor. `creates and returns a plan`

src/test/regress/sql/multi_select_distinct.sql
```diff
@@ -4,23 +4,30 @@
...
+
+-- column expressions are supported
+SELECT DISTINCT l_orderkey, l_shipinstruct || l_shipmode FROM lineitem_hash_part ORDER BY 2 , 1 LIMIT 5;
 
 -- function calls with const input are supported
 SELECT DISTINCT l_orderkey, strpos('AIR', 'A') FROM lineitem_hash_part ORDER BY 1,2 LIMIT 5;
 
--- function calls with non-const input are not supported
-SELECT DISTINCT l_orderkey, strpos(l_shipmode, 'A') FROM lineitem_hash_part;
+-- function calls with non-const input are  supported
```
metdos: Minor. Extra empty space.

src/backend/distributed/planner/multi_master_planner.c
metdos: Any update on this? Could you update the comment?

src/backend/distributed/planner/multi_master_planner.c
```diff
@@ -152,6 +153,75 @@ BuildAggregatePlan(Query *masterQuery, Plan *subPlan)
...
+ * availability of hash function it chooses HashAgg over Sort/Unique
+ * plans.
+ * This function has a potential performance issue since we blindly set
+ * Plan nodes without looking at cost. We might need to revisit this
+ * if we have performance issues with select distinct queries.
+ */
+static Plan *
+BuildDistinctPlan(Query *masterQuery, Plan *subPlan)
```
metdos: This function should go to below the `BuildSelectStatement()` where it is used.

src/backend/distributed/planner/multi_master_planner.c
```diff
@@ -203,10 +274,45 @@ BuildSelectStatement(Query *masterQuery, List *masterTargetList, CustomScan *rem
...
+	 * example order by a,b,a,a,b,c is translated to equivalent order by a,b,c.
+	 *
+	 * If the query has distinct clause but not distinct on, we first create
+	 * distinct plan that is either HashAggreate or Sort + Unique plans depending
+	 * on hashable property of columns in distinct clause. If there is order by
+	 * clause, it is handled after distinct planning.
+	 */
+	if (masterQuery->hasDistinctOn)
+	{
+		ListCell *distinctCell = NULL;
```
metdos: Thinking about this a bit more, do we need this? PostgreSQL requires distinct column to be added to order by clause if there is order by. If there is no order by, then the order is not important. Can't we just replace whole block with;

```
	if (masterQuery->distinctClause && !masterQuery->hasDistinctOn)
	{
		Plan *distinctPlan = BuildDistinctPlan(masterQuery, topLevelPlan);
		topLevelPlan = distinctPlan;
	}
```

mtuncer: Unique requires ordering, otherwise it won't work. Keeping this as it is per our discussion

metdos:
Looks good, just two minor comments.

It's very nice that we support all kind of distinct queries :)

🚀

Issue #1663, opened 2017-09-19
%I doesn't work with run_command_on_shards


samay-sharma:
Steps to reproduce:

```
create table test_table (col1 int, col2 int);
select create_distributed_table('test_table' , 'col1');
select run_command_on_shards('test_table', 'update %I set col1 = 1 where col2 = 2');
```

```
                          run_command_on_shards                           
---------------------------------------------------------------------------
 (102724,f,"ERROR:  relation ""public.test_table_102724"" does not exist")
 (102725,f,"ERROR:  relation ""public.test_table_102725"" does not exist")
 (102726,f,"ERROR:  relation ""public.test_table_102726"" does not exist")
 (102727,f,"ERROR:  relation ""public.test_table_102727"" does not exist")
 (102728,f,"ERROR:  relation ""public.test_table_102728"" does not exist")
 (102729,f,"ERROR:  relation ""public.test_table_102729"" does not exist")
 (102730,f,"ERROR:  relation ""public.test_table_102730"" does not exist")
 (102731,f,"ERROR:  relation ""public.test_table_102731"" does not exist")
 (102732,f,"ERROR:  relation ""public.test_table_102732"" does not exist")
 (102733,f,"ERROR:  relation ""public.test_table_102733"" does not exist")
 (102734,f,"ERROR:  relation ""public.test_table_102734"" does not exist")
 (102735,f,"ERROR:  relation ""public.test_table_102735"" does not exist")
 (102736,f,"ERROR:  relation ""public.test_table_102736"" does not exist")
 (102737,f,"ERROR:  relation ""public.test_table_102737"" does not exist")
 (102738,f,"ERROR:  relation ""public.test_table_102738"" does not exist")
 (102739,f,"ERROR:  relation ""public.test_table_102739"" does not exist")
 (102740,f,"ERROR:  relation ""public.test_table_102740"" does not exist")
 (102741,f,"ERROR:  relation ""public.test_table_102741"" does not exist")
 (102742,f,"ERROR:  relation ""public.test_table_102742"" does not exist")
 (102743,f,"ERROR:  relation ""public.test_table_102743"" does not exist")
 (102744,f,"ERROR:  relation ""public.test_table_102744"" does not exist")
 (102745,f,"ERROR:  relation ""public.test_table_102745"" does not exist")
 (102746,f,"ERROR:  relation ""public.test_table_102746"" does not exist")
 (102747,f,"ERROR:  relation ""public.test_table_102747"" does not exist")
[...]

PR #1687, opened 2017-10-04
Omit public schema from shard_name output

marcocitus:
@samay-sharma does this work for fixing #1663?

### DIFF
--- a/src/backend/distributed/relay/relay_event_utility.c
+++ b/src/backend/distributed/relay/relay_event_utility.c
@@ -720,7 +720,15 @@ shard_name(PG_FUNCTION_ARGS)
 
 	schemaId = get_rel_namespace(relationId);
 	schemaName = get_namespace_name(schemaId);
-	qualifiedName = quote_qualified_identifier(schemaName, relationName);
+
+	if (strncmp(schemaName, "public", NAMEDATALEN) == 0)
+	{
+		qualifiedName = (char *) quote_identifier(relationName);
+	}
+	else
+	{
+		qualifiedName = quote_qualified_identifier(schemaName, relationName);
+	}
 
 	PG_RETURN_TEXT_P(cstring_to_text(qualifiedName));
 }
--- a/src/test/regress/expected/isolation_add_node_vs_reference_table_operations.out
+++ b/src/test/regress/expected/isolation_add_node_vs_reference_table_operations.out
@@ -187,7 +187,7 @@ step s2-print-index-count:
 	SELECT 
 		nodeport, success, result 
 	FROM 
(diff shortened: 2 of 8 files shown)


metdos:
@samay-sharma any comments on this?

samay-sharma:
@marcocitus : Wanted to clarify on the behavior on %I. Does this mean that %I will work with all tables?

i.e. If there is a public schema, it'll not add the public that hence it'll work. Also, if there is a schema then it'll still work with the right schema.

If that's the behavior, then it looks good to me. Otherwise, could you clarify the behavior on %I?

marcocitus:
> i.e. If there is a public schema, it'll not add the public that hence it'll work. Also, if there is a schema then it'll still work with the right schema.

No, the underlying `format` function does not quote schema names separately, unfortunately. If there is a schema, it will be encoded as `"myschema.mytable"`, which will throw an error. We could implement our own format function, but that'll be 1-2 weeks of work.

This PR unbreaks existing usages of `run_command_on_*` within the `public` schema, to avoid breaking applications of users that upgrade.

furkansahin:
@samay-sharma if you are fine with this solution, the PR looks good to me.

jasonmp85:
> We could implement our own format function, but that'll be 1-2 weeks of work.

I still think it wouldn't be terribly hard to add our own format modifier that means "replace this token with quoted-schema-dot-quoted-relation". We'd just scan for it and replace it with `%I.%I` and then use the typical format method to supply schema, relation name.

This could preserve existing behavior while providing a way forward for users who always want `"schema"."relationname"` behavior. one–two weeks for that improvement seems a high estimate.

samay-sharma:
hey @marcocitus : Hmm. So, to summarize, %I would work for public schemas (i.e. existing %I usage would still work) but for tables in another schema we'd tell users to use %s. And that would schema qualify the table names.

If this is accurate, then it sounds fair to me. We can merge this PR and then update the docs and ask users with schemas to use %s. We can also add such an example to that section.

marcocitus:
> I still think it wouldn't be terribly hard to add our own format modifier that means "replace this token with quoted-schema-dot-quoted-relation". We'd just scan for it and replace it with %I.%I and then use the typical format method to supply schema, relation name.

Some users have scripts that use syntax like `%1$s` to be able to refer to the shard name in multiple places. I think it makes sense to come up with our own format (maybe $shardname?), but it probably requires some careful design and we need to make sure not to break everyone's hacky scripts.

marcocitus:
> Hmm. So, to summarize, %I would work for public schemas (i.e. existing %I usage would still work) but for tables in another schema we'd tell users to use %s. And that would schema qualify the table names.

Correct. You should also use %s if your table name contains strange characters, otherwise quotes will get added twice, but that's probably not super common.

metdos:
@marcocitus shall we merge this?

Issue #1131, opened 2017-01-13
UPDATE on multiple shards natively works


ozgune:
When a user has the real-time analytics use case, modification commands (e.g. record.update!(field: 'value') in #595) might open transactions involving multiple tables or shards. Right now, when running in a transaction, any DML command that opens a connection to a second node will error out.

In our documentation, we ask our users to call ``master_modify_multiple_shards`` to update records across multiple shards. We talked about natively integrating this into ``UPDATE`` statements. One question however was related to users' expectations: since this operation touches across multiple shards, concurrent ``UPDATE`` commands need to run in serial to avoid deadlocks. Do we want to inform the user about this behavior? (maybe through a ``NOTICE`` message)

lithp:
Some more questions:

- `master_modify_multiple_shards` isn't exactly performant, especially when using 2PC (soon to be the default?)
- `master_modify_multiple_shards`, when using 2PC, doesn't do any kind of automatic recovery after node-failure.
- `master_modify_multiple_shards` has the deadlock issue you already described
- If you have a replication factor > 1 and the query fails on a placement , most `UPDATE` queries will still attempts to modify the remaining placements. `master_modify_multiple_shards` doesn't care and fails the query if any of the placements fail

Will users understand that there are two kinds of queries, both called `UPDATE`, which happen through entirely different mechanisms, and which you should treat very differently?

--

If our eventual plan is to make Community absorb all of MX, this makes that eventual transition more difficult. Supporting `UPDATE` across multiple shards is a distributed transaction which we can _kind of_ cheaply get away with because we have a single master. Once that goes away this gets... tricky. Are we okay with that?

(All of the above also applies to `DELETE`, and therefore #1130)

marcocitus:
I think the scope of this issue is mainly to provide the basic UX/compatibility for multi-shard UPDATE/DELETE, with some mechanism for informing the user about how much it sucks for short writes.

[...]

PR #1690, opened 2017-10-10
Support multi shard update/delete queries

### LINKED ISSUE #1130: DELETE on multiple shards natively works
When users run a ``DELETE`` command that touches multiple shards, Citus currently displays an error message.  In #905, we introduced an improvement to our error messages to instruct the user to call ``master_modify_multiple_shards()`` for hash distributed tables, and ``master_apply_delete_command()`` for append distributed ones.

This issue tracks natively intercepting the ``DELETE`` command and having it call Citus' UDFs under the covers. We have two use cases for this scenario:

1. DELETE for expiring time-series data: Users who shard their data by tenant_id or entity_id (hash distribution) call DELETE to expire old events data. Users may also choose to call DELETE on append distributed tables, but this is likely out of scope.
2. DELETE a single record for a given tenant: Users can call this in one of two ways.
    1. DELETE ... WHERE tenant_id = 97 AND rec_id = 1; Citus already supports this.
    2. DELETE ... WHERE rec_id = 1; In this case, the user intends to do the above, but forgets to include tenant_id. As a result, they may run into small concurrency issues due to how we implement ``master_modify_multiple_shards()``. (Do they also accidentally delete data in this case?)

In summary, Citus could provide native DELETE support over multiple shards by calling ``master_modify_multiple_shards``. This may introduce small concurrency issues for users who make mistakes and forget to include the shard distribution column in their query.

velioglu:
Fixes #1130 and #1131.

Support multi shard update/delete (MSUD) plans by default. Since we have deadlock detection mechanism, users can send these queries using sql commands rather than master_modify_multiple_shards UDF. 

- [X] Let MSUD queries in router planner by removing related checks & asserts.
- [X] Create multiple tasks for MSUD queries while still creating a single task for select queries via router planner.
- [X] Prevent FROM clauses with MSUD queries.
- [x] Prevent volatile functions with MSUD queries.
- [x] Add GUC variable to select connection type.
- [x] Add support for sequential connections for MSUD queries.
- [x] Log MSUD queries according to MultiTaskQueryLogLevel.
- [x] Add regression tests (especially for node based connections and within transaction queries)

### DIFF
--- a/src/backend/distributed/planner/deparse_shard_query.c
+++ b/src/backend/distributed/planner/deparse_shard_query.c
@@ -54,7 +54,11 @@ RebuildQueryStrings(Query *originalQuery, List *taskList)
 		Task *task = (Task *) lfirst(taskCell);
 		Query *query = originalQuery;
 
-		if (task->insertSelectQuery)
+		if (UpdateOrDeleteQuery(query) && list_length(taskList))
+		{
+			query = copyObject(originalQuery);
+		}
+		else if (task->insertSelectQuery)
 		{
 			/* for INSERT..SELECT, adjust shard names in SELECT part */
 			RangeTblEntry *copiedInsertRte = NULL;
--- a/src/backend/distributed/planner/multi_planner.c
+++ b/src/backend/distributed/planner/multi_planner.c
@@ -37,6 +37,7 @@
 
 
 static List *plannerRestrictionContextList = NIL;
+int MultiTaskQueryLogLevel = MULTI_TASK_QUERY_INFO_OFF; /* multi-task query log level */
 
 /* create custom scan methods for separate executors */
 static CustomScanMethods RealTimeCustomScanMethods = {
@@ -284,6 +285,56 @@ IsModifyCommand(Query *query)
 
 
 /*
+ * IsMultiShardModifyPlan returns true if the given plan was generated for
+ * multi shard update or delete query.
+ */
+bool
+IsMultiShardModifyPlan(MultiPlan *multiPlan)
+{
+	if (IsUpdateOrDelete(multiPlan) && IsMultiTaskPlan(multiPlan))
+	{
+		return true;
+	}
+
+	return false;
+}
+
+
+/*
+ * IsMultiTaskPlan returns true if job contains multiple tasks.
+ */
+bool
+IsMultiTaskPlan(MultiPlan *multiPlan)
+{
+	Job *workerJob = multiPlan->workerJob;
+
+	if (workerJob != NULL && list_length(workerJob->taskList) > 1)
+	{
+		return true;
+	}
+
+	return false;
+}
+
+
+/*
+ * IsUpdateOrDelete returns true if the query performs update or delete.
+ */
+bool
+IsUpdateOrDelete(MultiPlan *multiPlan)
+{
+	CmdType commandType = multiPlan->operation;
+
+	if (commandType == CMD_UPDATE || commandType == CMD_DELETE)
+	{
+		return true;
+	}
+
+	return false;
+}
+
+
+/*
  * IsModifyMultiPlan returns true if the multi plan performs modifications,
  * false otherwise.
  */
@@ -436,9 +487,11 @@ CreateDistributedPlan(PlannedStmt *localPlan, Query *originalQuery, Query *query
 
 	/*
 	 * As explained above, force planning costs to be unrealistically high if
-	 * query planning failed (possibly) due to prepared statement parameters.
+	 * query planning failed (possibly) due to prepared statement parameters or
+	 * if it is planned as a multi shard modify query.
 	 */
-	if (distributedPlan->planningError && hasUnresolvedParams)
+	if ((distributedPlan->planningError || IsMultiShardModifyPlan(distributedPlan)) &&
+		hasUnresolvedParams)
 	{
 		/*
 		 * Arbitraryly high cost, but low enough that it can be added up
@@ -527,6 +580,20 @@ FinalizePlan(PlannedStmt *localPlan, MultiPlan *multiPlan)
 		{
 			customScan->methods = &DelayedErrorCustomScanMethods;
 			break;
+		}
+	}
+
+	if (IsMultiTaskPlan(multiPlan))
+	{
+		/* if it is not a single task executable plan, inform user according to the log level */
+		if (MultiTaskQueryLogLevel != MULTI_TASK_QUERY_INFO_OFF)
+		{
+			ereport(MultiTaskQueryLogLevel, (errmsg(
+												 "multi-task query about to be executed"),
+											 errhint(
+												 "Queries are split to multiple tasks "
+												 "if they have to be split into several"
+												 " queries on the workers.")));
 		}
 	}
 
--- a/src/backend/distributed/planner/multi_router_planner.c
+++ b/src/backend/distributed/planner/multi_router_planner.c
@@ -2,8 +2,8 @@
  *
  * multi_router_planner.c
  *
- * This file contains functions to plan single shard queries
- * including distributed table modifications.
+ * This file contains functions to plan multiple shard queries without any
+ * aggregation step including distributed table modifications.
  *
  * Copyright (c) 2014-2016, Citus Data, Inc.
  *
@@ -137,7 +137,6 @@ static List * ExtractInsertValuesList(Query *query, Var *partitionColumn);
 static bool MultiRouterPlannableQuery(Query *query,
 									  RelationRestrictionContext *restrictionContext);
 static DeferredErrorMessage * ErrorIfQueryHasModifyingCTE(Query *queryTree);
-static bool UpdateOrDeleteQuery(Query *query);
 static RangeTblEntry * GetUpdateOrDeleteRTE(List *rangeTableList);
 static bool UpdateOrDeleteRTE(RangeTblEntry *rangeTableEntry);
 static bool SelectsFromDistributedTable(List *rangeTableList);
@@ -146,7 +145,13 @@ static List * get_all_actual_clauses(List *restrictinfo_list);
 #endif
 static int CompareInsertValuesByShardId(const void *leftElement,
 										const void *rightElement);
-
+static uint64 GetInitialShardId(List *relationShardList);
+static List * SingleShardSelectTaskList(Query *query, List *relationShardList,
+										List *placementList, uint64 shardId);
+static List * SingleShardModifyTaskList(Query *query, List *relationShardList,
+										List *placementList, uint64 shardId);
+static List * MultiShardModifyTaskList(Query *originalQuery, List *relationShardList,
+									   bool requiresMasterEvaluation);
 
 /*
  * CreateRouterPlan attempts to create a router executor plan for the given
@@ -250,7 +255,9 @@ CreateSingleTaskRouterPlan(Query *originalQuery, Query *query,
 		return multiPlan;
 	}
 
+	/* we cannot have multi shard update/delete query via this code path */
 	job = RouterJob(originalQuery, restrictionContext, &multiPlan->planningError);
+
 	if (multiPlan->planningError)
 	{
 		/* query cannot be handled by this planner */
@@ -516,11 +523,21 @@ ModifyQuerySupported(Query *queryTree, bool multiShardQuery)
 		 */
 		if (!UpdateOrDeleteQuery(queryTree) || multiShardQuery)
 		{
+			StringInfo errorHint = makeStringInfo();
+			DistTableCacheEntry *cacheEntry = DistributedTableCacheEntry(
+				distributedTableId);
+			char *partitionKeyString = cacheEntry->partitionKeyString;
+			char *partitionColumnName = ColumnNameToColumn(distributedTableId,
+														   partitionKeyString);
+
+			appendStringInfo(errorHint,
+							 "Consider using an equality filter on partition column \"%s\" to target a single shard.",
+							 partitionColumnName);
+
 			return DeferredError(ERRCODE_FEATURE_NOT_SUPPORTED,
-								 "cannot perform distributed planning for the given "
-								 "modifications",
-								 "Subqueries are not supported in distributed "
-								 "modifications.", NULL);
+								 "subqueries are not supported in modifications across "
+								 "multiple shards",
+								 errorHint->data, NULL);
 		}
 	}
 
@@ -602,8 +619,20 @@ ModifyQuerySupported(Query *queryTree, bool multiShardQuery)
 			 */
 			if (rangeTableEntry->rtekind == RTE_SUBQUERY)
 			{
-				rangeTableEntryErrorDetail = "Subqueries are not supported in"
-											 " distributed modifications.";
+				StringInfo errorHint = makeStringInfo();
+				DistTableCacheEntry *cacheEntry = DistributedTableCacheEntry(
+					distributedTableId);
+				char *partitionKeyString = cacheEntry->partitionKeyString;
+				char *partitionColumnName = ColumnNameToColumn(distributedTableId,
+															   partitionKeyString);
+
+				appendStringInfo(errorHint, "Consider using an equality filter on "
+											"partition column \"%s\" to target a single shard.",
+								 partitionColumnName);
+
+				return DeferredError(ERRCODE_FEATURE_NOT_SUPPORTED, "subqueries are not "
+																	"supported in modifications across multiple shards",
+									 errorHint->data, NULL);
 			}
 			else if (rangeTableEntry->rtekind == RTE_JOIN)
 			{
@@ -741,6 +770,14 @@ ModifyQuerySupported(Query *queryTree, bool multiShardQuery)
 								 "RETURNING clause",
 								 NULL, NULL);
 		}
+
+		if (queryTree->jointree->quals != NULL &&
+			nodeTag(queryTree->jointree->quals) == T_CurrentOfExpr)
+		{
+			return DeferredError(ERRCODE_FEATURE_NOT_SUPPORTED,
+								 "cannot run DML queries with cursors", NULL,
+								 NULL);
+		}
 	}
 
 	if (commandType == CMD_INSERT && queryTree->onConflict != NULL)
@@ -836,7 +873,7 @@ ModifyQuerySupported(Query *queryTree, bool multiShardQuery)
  * UpdateOrDeleteQuery checks if the given query is an UPDATE or DELETE command.
  * If it is, it returns true otherwise it returns false.
  */
-static bool
+bool
 UpdateOrDeleteQuery(Query *query)
 {
 	CmdType commandType = query->commandType;
@@ -1326,14 +1363,15 @@ ExtractFirstDistributedTableId(Query *query)
 }
 
 
-/* RouterJob builds a Job to represent a single shard select/update/delete query */
+/*
+ * RouterJob builds a Job to represent a single shard select/update/delete and
+ * multiple shard update/delete queries.
+ */
 static Job *
 RouterJob(Query *originalQuery, RelationRestrictionContext *restrictionContext,
 		  DeferredErrorMessage **planningError)
 {
 	Job *job = NULL;
-	Task *task = NULL;
-	StringInfo queryString = makeStringInfo();
 	uint64 shardId = INVALID_SHARD_ID;
 	List *placementList = NIL;
 	List *relationShardList = NIL;
@@ -1341,6 +1379,7 @@ RouterJob(Query *originalQuery, RelationRestrictionContext *restrictionContext,
 	bool replacePrunedQueryWithDummy = false;
 	bool requiresMasterEvaluation = false;
 	RangeTblEntry *updateOrDeleteRTE = NULL;
+	bool isMultiShardModifyQuery = false;
 
 	/* router planner should create task even if it deosn't hit a shard at all */
 	replacePrunedQueryWithDummy = true;
@@ -1350,7 +1389,8 @@ RouterJob(Query *originalQuery, RelationRestrictionContext *restrictionContext,
 
 	(*planningError) = PlanRouterQuery(originalQuery, restrictionContext,
 									   &placementList, &shardId, &relationShardList,
-									   replacePrunedQueryWithDummy);
+									   replacePrunedQueryWithDummy,
+									   &isMultiShardModifyQuery);
 	if (*planningError)
 	{
 		return NULL;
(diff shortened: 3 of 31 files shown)


src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1374,41 +1390,87 @@ RouterJob(Query *originalQuery, RelationRestrictionContext *restrictionContext,
...
+
+			task = CitusMakeNode(Task);
+			task->jobId = jobId;
+			task->taskId = taskId++;
+			task->taskType = MODIFY_TASK;
+			task->queryString = shardQueryString->data;
+			task->dependedTaskList = NULL;
+			task->replicationModel = REPLICATION_MODEL_INVALID;
+			task->anchorShardId = shardId;
+			task->taskPlacementList = FinalizedShardPlacementList(shardId);
```
marcocitus: We generally shouldn't assign placements in the planner. We do this in `CitusModifyBeginScan`.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1374,41 +1390,87 @@ RouterJob(Query *originalQuery, RelationRestrictionContext *restrictionContext,
...
+				shardPlacementCell);
+			ShardInterval *shardInterval = LoadShardInterval(shardPlacement->shardId);
+			Oid relationId = shardInterval->relationId;
+			uint64 shardId = shardInterval->shardId;
+			StringInfo shardQueryString = makeStringInfo();
+			Task *task = NULL;
+
+			deparse_shard_query(originalQuery, relationId, shardId, shardQueryString);
+
+			task = CitusMakeNode(Task);
```
marcocitus: can probably use `CreateTask(MODIFY_TASK)` here

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1374,41 +1390,87 @@ RouterJob(Query *originalQuery, RelationRestrictionContext *restrictionContext,
...
+		{
+			ShardPlacement *shardPlacement = (ShardPlacement *) lfirst(
+				shardPlacementCell);
+			ShardInterval *shardInterval = LoadShardInterval(shardPlacement->shardId);
+			Oid relationId = shardInterval->relationId;
+			uint64 shardId = shardInterval->shardId;
+			StringInfo shardQueryString = makeStringInfo();
+			Task *task = NULL;
+
+			deparse_shard_query(originalQuery, relationId, shardId, shardQueryString);
```
marcocitus: I kind of consider this function deprecated. Is there a way to use `UpdateRelationToShardNames` and `pg_get_query_def` here?

src/test/regress/expected/multi_modifications.out
```diff
@@ -984,12 +968,10 @@ SELECT * FROM summary_table ORDER BY id;
 -- unsupported multi-shard updates
 UPDATE summary_table SET average_value = average_query.average FROM (
 	SELECT avg(value) AS average FROM raw_table) average_query;
-ERROR:  cannot run UPDATE command which targets multiple shards
-HINT:  Consider using an equality filter on partition column "id" to target a single shard. If you'd like to run a multi-shard operation, use master_modify_multiple_shards().
+ERROR:  support only single distributed table with multi shard modify commands
 UPDATE summary_table SET average_value = average_value + 1 WHERE id =
   (SELECT id FROM raw_table WHERE value > 100);
-ERROR:  cannot run UPDATE command which targets multiple shards
-HINT:  Consider using an equality filter on partition column "id" to target a single shard. If you'd like to run a multi-shard operation, use master_modify_multiple_shards().
```
marcocitus: I think the old error and hint are quite informative. 

Maybe we could say:
```
ERROR: cannot run UPDATE command with a subselect that targets multiple shards
HINT: Consider using an equality filter on partition column "id" to target a single shard.
```

velioglu:
@marcocitus I have updated the PR with changes on the executor part (and some small bugs in the planner part). I'm going to add regression tests as a last step, that's why WIP tag is still there. I will also update deparse_shard_query to pg_get_query_def with regression tests. You can review the planner/executor logic now.

src/backend/distributed/shared_library_init.c
```diff
@@ -733,6 +739,16 @@ RegisterCitusConfigVariables(void)
 		0,
 		NULL, NULL, NULL);
 
+	DefineCustomEnumVariable(
+		"citus.multi_shard_modify_connection_type",
```
marcocitus: maybe `multi_shard_modify_executor`

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1842,6 +1943,41 @@ WorkersContainingAllShards(List *prunedShardIntervalsList)
 }
 
 
+static List *
+WorkersContainingModifyShards(List *prunedShardIntervalsList)
```
marcocitus: This function needs a comment

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1842,6 +1943,41 @@ WorkersContainingAllShards(List *prunedShardIntervalsList)
...
+
+	foreach(prunedShardIntervalCell, prunedShardIntervalsList)
+	{
+		List *shardIntervalList = (List *) lfirst(prunedShardIntervalCell);
+		ListCell *shardIntervalCell = NULL;
+
+		if (shardIntervalList == NIL)
+		{
+			continue;
+		}
```
marcocitus: This `if` block seems redundant, since the `foreach` below will be a noop if `shardIntervalList` is `NIL` (empty lists are represented as `NIL`).

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1682,14 +1732,73 @@ PlanRouterQuery(Query *originalQuery, RelationRestrictionContext *restrictionCon
...
+		ListCell *shardIntervalCell = NULL;
+
+		foreach(shardIntervalCell, relationShardList)
+		{
+			ShardInterval *shardInterval = (ShardInterval *) lfirst(shardIntervalCell);
+			Oid relationId = shardInterval->relationId;
+
+			relationIds = list_append_unique_oid(relationIds, relationId);
+
+			if (list_length(relationIds) > 1)
```
marcocitus: Using a list here seems a little confusing. I think it would make more sense to declare an `Oid firstRelationId = InvalidOid;` variable, and then:

```C
if (firstRelationId == InvalidOid) 
{
    firstRelationId = relationId;
}
else if (relationId != firstRelationId)
{
    return true;
}
```

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1589,30 +1626,34 @@ PlanRouterQuery(Query *originalQuery, RelationRestrictionContext *restrictionCon
...
+			relationShard->relationId = shardInterval->relationId;
+			relationShard->shardId = shardInterval->shardId;
 
-		*relationShardList = lappend(*relationShardList, relationShard);
+			/*
+			 * We need anchor shard id for compute tasks. Also, note that we have
+			 * single shard interval for select queries at this point. Therefore,
+			 * condition below should run once only for router select queries.
+			 */
+			if (!(*multiShardModifyQuery))
```
marcocitus: Could we use an additional local variable  `multiShardModifyQuery` to make this a little more readable.?

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -211,7 +217,23 @@ CreateModifyPlan(Query *originalQuery, Query *query,
 		return multiPlan;
 	}
 
-	ereport(DEBUG2, (errmsg("Creating router plan")));
+	if (*multiShardModifyQuery)
+	{
+		/* although the plan is a router one, we still touch multiple shards */
+		if (MultiTaskQueryLogLevel != MULTI_TASK_QUERY_INFO_OFF)
+		{
+			ereport(MultiTaskQueryLogLevel, (errmsg(
```
marcocitus: it would be nice if we could keep these checks in one place. For real-time SELECT it's in `JobExecutorType`, though that seems a little weird as well. Maybe move both to `FinalizePlan`?

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1374,41 +1405,78 @@ RouterJob(Query *originalQuery, RelationRestrictionContext *restrictionContext,
...
+	 */
+	if (*multiShardModifyQuery)
 	{
-		task = CreateTask(ROUTER_TASK);
+		List *taskList = NIL;
+		ListCell *relationShardCell = NULL;
+		int taskId = 1;
+
+		/* lock metadata before getting placement lists */
+		LockShardListMetadata(placementList, ShareLock);
```
marcocitus: Can remove this, since we already do it in the executor.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1626,54 +1655,64 @@ PlanRouterQuery(Query *originalQuery, RelationRestrictionContext *restrictionCon
...
 	}
 	else
 	{
-		/*
-		 * For INSERT ... SELECT, this query could be still a valid for some other target
-		 * shard intervals. Thus, we should return empty list if there aren't any matching
-		 * workers, so that the caller can decide what to do with this task.
-		 */
-		return NULL;
+		workerList = WorkersContainingModifyShards(prunedRelationShardList);
```
marcocitus: What's the point of this code? We don't use the result as far as I can tell (`placementList`) isn't used for multi-shard router queries in `RouterJob`. Is this something we can omit?

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1589,30 +1614,34 @@ PlanRouterQuery(Query *originalQuery, RelationRestrictionContext *restrictionCon
...
-		relationShard->relationId = shardInterval->relationId;
-		relationShard->shardId = shardInterval->shardId;
+			/*
+			 * We need anchor shard id for compute tasks. Also, note that we have
+			 * single shard interval for select queries at this point. Therefore,
+			 * condition below should run once only for router select queries.
+			 */
+			if (!localMultiShardModifyQuery)
+			{
+				shardId = relationShard->shardId;
```
marcocitus: I find it a bit hard to follow how the `if (!lockMultiShardModifyQuery)` relates to the comment. What if we extract the shardId separately? (e.g. introduce a `GetInitialShardId` function).

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1626,54 +1655,64 @@ PlanRouterQuery(Query *originalQuery, RelationRestrictionContext *restrictionCon
 	 * still run the query but the result will be empty. We create a dummy shard
 	 * placement for the first active worker.
 	 */
-	if (shardsPresent)
+	if (!localMultiShardModifyQuery)
```
marcocitus: This code block is becoming a bit unwieldy. Could we do something like:

```
if (isMultiShardModifyQuery)
{
    *multiShardModifyQuery  = true;
    return NULL;
}
```

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1510,76 +1564,47 @@ PlanRouterQuery(Query *originalQuery, RelationRestrictionContext *restrictionCon
...
+				return planningError;
+			}
 
-			appendStringInfo(errorHint, "Consider using an equality filter on "
-										"partition column \"%s\" to target a "
-										"single shard. If you'd like to run a "
-										"multi-shard operation, use "
-										"master_modify_multiple_shards().",
-							 partitionColumnName);
+			localMultiShardModifyQuery = true;
```
marcocitus: This function gets a little messy due to changes that are specific to multi-shard update/delete below. Do you think it'd be possible to split up this function and call the relevant part, e.g.

```
if (multiShardQuery)
{
    if (commandType == CMD_SELECT)
    {
        ...
        return planningError;
    }

    PlanMultiShardModifyQuery(...);
}
else
{
    PlanSingleShardQuery(...);
}
```

src/backend/distributed/planner/deparse_shard_query.c
```diff
@@ -100,7 +100,16 @@ RebuildQueryStrings(Query *originalQuery, List *taskList)
 
 		ereport(DEBUG4, (errmsg("query before rebuilding: %s", task->queryString)));
 
-		UpdateTaskQueryString(query, relationId, valuesRTE, task);
+		/* we need to use copy of the query if it is a multi shard update/delete query */
+		if (UpdateOrDeleteQuery(query) && list_length(taskList))
```
marcocitus: The problem this seems to solve is that we need to copy the query tree when there are multiple tasks, such that we start from the original every time.

What if we do this at the start?

```
/* when there are multiple tasks, start from a fresh copy of the query tree */
if (list_length(taskList) > 0)
{
    query = copyObject(originalQuery);
}
```

then in the `INSERT...SELECT` logic we can remove the `query = copyObject(originalQuery);` line.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1628,7 +1660,7 @@ PlanRouterQuery(Query *originalQuery, RelationRestrictionContext *restrictionCon
 	 */
 	if (shardsPresent)
 	{
-		workerList = WorkersContainingAllShards(prunedRelationShardList);
+		workerList = WorkersContainingSelectShards(prunedRelationShardList);
```
marcocitus: This rename seems unnecessary.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1589,23 +1619,22 @@ PlanRouterQuery(Query *originalQuery, RelationRestrictionContext *restrictionCon
...
 
-		/* add relation to shard mapping */
-		relationShard = CitusMakeNode(RelationShard);
-		relationShard->relationId = shardInterval->relationId;
-		relationShard->shardId = shardInterval->shardId;
-
-		*relationShardList = lappend(*relationShardList, relationShard);
+	if (isMultiShardModifyQuery)
+	{
+		*multiShardModifyQuery = isMultiShardModifyQuery;
```
marcocitus: maybe `= true;`

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1682,14 +1715,45 @@ PlanRouterQuery(Query *originalQuery, RelationRestrictionContext *restrictionCon
...
+		List *prunedShardList = (List *) lfirst(prunedRelationShardListCell);
+		ListCell *shardIntervalCell = NULL;
+
+		/* no shard is present or all shards are pruned out case will be handled later */
+		if (prunedShardList == NIL)
+		{
+			continue;
+		}
+
+		foreach(shardIntervalCell, prunedShardList)
```
marcocitus: Maybe use `linitial` as in the the original code.

src/test/regress/expected/multi_shard_update_delete.out
```diff
@@ -0,0 +1,236 @@
...
+------------------------
+ 
+(1 row)
+
+INSERT INTO users_reference_copy_table SELECT * FROM users_reference_table;
+-- Run multi shard updates and deletes without transaction on hash distributed tables
+SELECT SUM(value_1) FROM users_copy_table;
+   sum   
+---------
+ 4998894
```
marcocitus: Is there a way to get some more predictable outputs? I have no way of telling whether the outputs look right.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -211,7 +213,10 @@ CreateModifyPlan(Query *originalQuery, Query *query,
 		return multiPlan;
 	}
 
-	ereport(DEBUG2, (errmsg("Creating router plan")));
+	if (*multiShardModifyQuery == false)
+	{
+		ereport(DEBUG2, (errmsg("Creating router plan")));
```
marcocitus: Why don't we want to show this debug message? Explain will still show this query as using `Citus Router`. This complicates the signature of `RouterJob`, so not sure if it's worthwhile.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1374,41 +1396,80 @@ RouterJob(Query *originalQuery, RelationRestrictionContext *restrictionContext,
...
+			task->taskId = taskId++;
+			task->anchorShardId = relationShard->shardId;
+			task->queryString = shardQueryString->data;
+			task->relationShardList = relationShardList;
+
+			taskList = lappend(taskList, task);
+		}
+
+
+		job->taskList = taskList;
```
marcocitus: This function is still not very readable.

Maybe something like:
```
if (originalQuery->commandType == CMD_SELECT)
{
    taskList = SingleShardSelectTaskList(originalQuery, relationShardList, placementList);
}
else if (*multiShardModifyQuery) // <-- can be a regular variable after removing the parameter
{
    taskList = MultiShardModifyTaskList(originalQuery, relationShardList);
}
else
{
    taskList = SingleShardModifyTaskList(originalQuery, relationShardList);
}
```

I'm also wondering whether we can merge the latter two.

src/test/regress/expected/multi_modifications.out
```diff
@@ -984,12 +968,12 @@ SELECT * FROM summary_table ORDER BY id;
 -- unsupported multi-shard updates
 UPDATE summary_table SET average_value = average_query.average FROM (
 	SELECT avg(value) AS average FROM raw_table) average_query;
-ERROR:  cannot run UPDATE command which targets multiple shards
-HINT:  Consider using an equality filter on partition column "id" to target a single shard. If you'd like to run a multi-shard operation, use master_modify_multiple_shards().
+ERROR:  cannot perform distributed planning for the given modifications
+DETAIL:  Subqueries are not supported in distributed modifications.
```
marcocitus: We're regressing on the quality of the error messages here.

Let's aim for:
```
ERROR: subqueries are not supported in modifications across multiple shards
HINT: Consider using an equality filter on partition column "id" to target a single shard.
```

src/backend/distributed/planner/multi_planner.c
```diff
@@ -436,9 +473,10 @@ CreateDistributedPlan(PlannedStmt *localPlan, Query *originalQuery, Query *query
 
 	/*
 	 * As explained above, force planning costs to be unrealistically high if
-	 * query planning failed (possibly) due to prepared statement parameters.
+	 * query planning failed (possibly) due to prepared statement parameters or
+	 * if it is planned as a multi shard modify query.
 	 */
-	if (distributedPlan->planningError && hasUnresolvedParams)
+	if ((distributedPlan->planningError || multiShardModifyQuery) && hasUnresolvedParams)
```
marcocitus: Passing a simple boolean flag as a pointer all the way down the planner tree is a little ugly. Is there a function we can write that can tell whether a query is a `MultiShardModifyPlan(distributedPlan)`?

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1374,42 +1420,126 @@ RouterJob(Query *originalQuery, RelationRestrictionContext *restrictionContext,
...
+		Query *copiedQuery = copyObject(originalQuery);
+		Task *task = CreateTask(MODIFY_TASK);
+		StringInfo shardQueryString = makeStringInfo();
+
+		if (!requiresMasterEvaluation)
 		{
-			ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
-							errmsg("cannot perform select on a distributed table "
-								   "and modify a reference table")));
+			UpdateRelationToShardNames((Node *) copiedQuery, relationShardList);
```
marcocitus: We could move `Query *copiedQuery = copyObject(originalQuery);` here such that we don't have to copy the query tree unless we need it.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1374,42 +1420,126 @@ RouterJob(Query *originalQuery, RelationRestrictionContext *restrictionContext,
...
+			pg_get_query_def(copiedQuery, shardQueryString);
 		}
 
-		task = CreateTask(MODIFY_TASK);
-		task->replicationModel = modificationTableCacheEntry->replicationModel;
+		task->taskId = taskId++;
+		task->anchorShardId = relationShard->shardId;
+		task->queryString = shardQueryString->data;
```
marcocitus: We should probably only allocate a string info if we have an actual query and otherwise leave it as NULL.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1501,85 +1631,54 @@ SelectsFromDistributedTable(List *rangeTableList)
...
-			partitionColumnName = ColumnNameToColumn(updateOrDeleteRTE->relid,
-													 partitionKeyString);
-
-			appendStringInfo(errorHint, "Consider using an equality filter on "
-										"partition column \"%s\" to target a "
-										"single shard. If you'd like to run a "
-										"multi-shard operation, use "
-										"master_modify_multiple_shards().",
-							 partitionColumnName);
+			prunedRelationShardList = NIL;
```
marcocitus: Not sure why we're setting this local variable to NIL just before returning.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1674,6 +1775,7 @@ PlanRouterQuery(Query *originalQuery, RelationRestrictionContext *restrictionCon
 		UpdateRelationToShardNames((Node *) originalQuery, *relationShardList);
```
marcocitus: What do you think of moving this line to `SingleShardSelectTaskList` and `SingleShardModifyTaskList`? We also do this in `MultiShardModifyTaskList` for multi-shard UPDATE/DELETE. It might also make sense to define a little function that first does `UpdateRelationToShardNames` and then `pg_get_query_def`.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -2330,6 +2451,7 @@ ExtractInsertValuesList(Query *query, Var *partitionColumn)
  * by setting citus.enable_router_execution flag to false.
  */
 static bool
+/* TODO Does the command and implementation agree with each other ? */
```
marcocitus: Did we address this comment?

src/test/regress/expected/multi_explain_0.out
```diff
@@ -747,6 +747,83 @@ t
 SELECT true AS valid FROM explain_json($$
 	SELECT avg(l_linenumber) FROM lineitem WHERE l_orderkey > 9030$$);
 t
+-- Test multi shard update
+EXPLAIN (COSTS FALSE)
+	UPDATE lineitem
+	SET l_suppkey = 12;
```
marcocitus: maybe add one where we have a `WHERE partcol = 1 OR partcol = 2` to see if pruning works.

src/test/regress/expected/multi_modifications.out
```diff
@@ -984,12 +968,12 @@ SELECT * FROM summary_table ORDER BY id;
 -- unsupported multi-shard updates
 UPDATE summary_table SET average_value = average_query.average FROM (
 	SELECT avg(value) AS average FROM raw_table) average_query;
-ERROR:  cannot run UPDATE command which targets multiple shards
-HINT:  Consider using an equality filter on partition column "id" to target a single shard. If you'd like to run a multi-shard operation, use master_modify_multiple_shards().
+ERROR:  subqueries are not supported in modifications across multiple shards
+DETAIL:  Consider using an equality filter on partition column "id" to target a single shard.
```
marcocitus: Perfect, thanks.

src/backend/distributed/shared_library_init.c
```diff
@@ -733,6 +739,16 @@ RegisterCitusConfigVariables(void)
 		0,
 		NULL, NULL, NULL);
 
+	DefineCustomEnumVariable(
+		"citus.multi_shard_modify_executor",
```
marcocitus: I think executor might cause confusion with router vs. real-time vs. task-tracker. Maybe `multi_shard_modify_mode`, or something that's more in line with postgres GUCs, e.g.: `enable_parallel_modification` (make it a boolean that defaults to true).

src/test/regress/sql/multi_shard_update_delete.sql
```diff
@@ -0,0 +1,271 @@
...
+SELECT SUM(value_3) FROM users_copy_table;
+
+-- Check with kind of constraints
+UPDATE users_copy_table SET value_3 = 1 WHERE user_id = 20 or true;
+SELECT SUM(value_3) FROM users_copy_table;
+UPDATE users_copy_table SET value_3 = 0 WHERE user_id = 20 and false;
+SELECT SUM(value_3) FROM users_copy_table;
+
+-- Run multi shard updates with prepared statements
+PREPARE foo_plan(int,int) AS UPDATE users_copy_table SET value_1 = $1, value_3 = $2;
```
marcocitus: is there a way to verify that we're not accidentally running a multi-shard update when we have a prepared statement with `WHERE user_id = $1`?

src/test/regress/sql/multi_shard_update_delete.sql
```diff
@@ -0,0 +1,271 @@
...
+SET citus.shard_replication_factor to 1;
+CREATE TABLE tt1(id int, col_2 int) partition by range (col_2);
+CREATE TABLE tt1_510 partition of tt1 for VALUES FROM (5) to (10);
+CREATE TABLE tt1_1120 partition of tt1 for VALUES FROM (11) to (20);
+INSERT INTO tt1 VALUES(1,11);
+INSERT INTO tt1 VALUES(3,15);
+INSERT INTO tt1 VALUES(5,17);
+INSERT INTO tt1 VALUES(6,19);
+INSERT INTO tt1 VALUES(8,17);
+INSERT INTO tt1 VALUES(2,12);
```
marcocitus: maybe 

`INSERT INTO tt1 VALUES (1,11), (3,15), (5,17), (6,19), (8,17), (2,12);`

src/test/regress/sql/multi_shard_update_delete.sql
```diff
@@ -0,0 +1,271 @@
...
+
+-- Run multi shard updates and deletes without transaction on hash distributed tables
+SELECT COUNT(*) FROM users_copy_table;
+UPDATE users_copy_table SET value_1 = 1;
+SELECT SUM(value_1) FROM users_copy_table;
+
+SELECT COUNT(*) FROM users_copy_table WHERE user_id = 20 or user_id = 8 or user_id = 60;
+SELECT SUM(value_2) FROM users_copy_table WHERE user_id = 20 or user_id = 8 or user_id = 60;
+UPDATE users_copy_table SET value_2 = value_2 + 1 WHERE user_id = 20 or user_id = 8 or user_id = 60;
+SELECT SUM(value_2) FROM users_copy_table WHERE user_id = 20 or user_id = 8 or user_id = 60;
```
marcocitus: These queries are dizzying and the contents of the table kind of depends on other tests. Could we do something that will be simpler for future readers to follow? Maybe with some manual data.

src/backend/distributed/planner/multi_router_planner.c
velioglu: We may completely remove this function. Modify commands were already handled by CreateInsertSelectPlan and CreateModifyPlan. Also, we check whether a SELECT query is router plannable with CreateSingleTaskRouterPlan. We may also remove TODO declared in the CreateRouterPlan.

src/backend/distributed/planner/multi_router_planner.c
velioglu: In that case, we need to change the implementation of INSERT INTO .. SELECT queries and subqueries too.

src/test/regress/expected/multi_explain_0.out
velioglu: Done

src/test/regress/sql/multi_shard_update_delete.sql
velioglu: Done

src/test/regress/expected/multi_explain_0.out
```diff
@@ -747,6 +747,101 @@ t
...
+        Node: host=localhost port=57637 dbname=regression
+        ->  Update on lineitem_290007 lineitem
+              ->  Seq Scan on lineitem_290007 lineitem
+	
+EXPLAIN (COSTS FALSE)
+	UPDATE lineitem
+	SET l_suppkey = 12
+	WHERE l_orderkey = 1 OR l_orderkey = 0;
+Custom Scan (Citus Router)
+  Task Count: 1
```
marcocitus: 1 is a bit suspicious, do we have an example that prunes to 2 shards?

src/test/regress/multi_mx_schedule
```diff
@@ -26,7 +26,7 @@ test: multi_mx_tpch_query7_nested multi_mx_ddl
 test: multi_mx_repartition_udt_prepare
 test: multi_mx_repartition_join_w1 multi_mx_repartition_join_w2 multi_mx_repartition_udt_w1 multi_mx_repartition_udt_w2
 test: multi_mx_metadata 
-test: multi_mx_modifications multi_mx_transaction_recovery
+#test: multi_mx_modifications multi_mx_transaction_recovery
```
marcocitus: should uncomment and add an MX test

marcocitus:
After fixing the 2 tests and squashing the commits, this is ready to :ship:

src/backend/distributed/planner/multi_router_planner.c
marcocitus: Well, I'm not sure a future reader will be able to parse this TODO, so maybe remove the "and it has a filter to reduce number of shard pairs to one, all shard pairs are located on the same node" bit from the function comment and remove this TODO.

src/test/regress/specs/isolation_multi_shard_modify_vs_all.spec
```diff
@@ -0,0 +1,125 @@
...
+permutation "s1-begin" "s1-update_value_1_of_1_or_3" "s2-begin" "s2-update_value_1_of_5_or_7" "s1-commit" "s2-commit" "s2-select" 
+permutation "s1-begin" "s1-update_value_1_of_1_or_3" "s2-begin" "s2-update_value_1_of_1_or_3" "s1-commit" "s2-commit" "s2-select" 
+permutation "s1-begin" "s1-update_all_value_1" "s2-begin" "s2-insert-to-table" "s1-commit" "s2-commit" "s2-select" 
+permutation "s1-begin" "s1-update_all_value_1" "s2-begin" "s2-insert-into-select" "s1-commit" "s2-commit" "s2-select"
+
+# test with parallel connections
+permutation "s1-begin" "s1-change_connection_mode_to_sequential" "s1-update_all_value_1" "s2-begin" "s2-change_connection_mode_to_sequential" "s2-update_all_value_1" "s1-commit" "s2-commit" "s2-select"
+permutation "s1-begin" "s1-change_connection_mode_to_sequential" "s1-update_value_1_of_1_or_3" "s2-begin" "s2-change_connection_mode_to_sequential" "s2-update_value_1_of_1_or_3" "s1-commit" "s2-commit" "s2-select"
+permutation "s1-begin" "s1-change_connection_mode_to_sequential" "s1-update_value_1_of_1_or_3" "s2-begin" "s2-change_connection_mode_to_sequential" "s2-update_value_1_of_5_or_7" "s1-commit" "s2-commit" "s2-select"
```
marcocitus: It's useful to add a comment briefly describing the test, and what your expectation is for the outcome of the test.

e.g. "sequential multi-shard UPDATEs should not block each other"

src/test/regress/expected/isolation_multi_shard_modify_vs_all.out
```diff
@@ -0,0 +1,355 @@
...
+
+step s1-update_value_1_of_1_or_3: 
+	UPDATE users_test_table SET value_1 = 5 WHERE user_id = 1 or user_id = 3;
+
+step s2-begin: 
+	BEGIN;
+
+step s2-update_value_1_of_5_or_7: 
+	UPDATE users_test_table SET value_1 = 4 WHERE user_id = 5 or user_id = 7;
+ <waiting ...>
```
marcocitus: this `<waiting ...>` is slightly surprising since they don't hit the same user IDs. I guess they have a shard in common?

src/test/regress/expected/isolation_multi_shard_modify_vs_all.out
```diff
@@ -0,0 +1,317 @@
...
+6              21             11             25             
+3              23             8              25             
+7              27             12             18             
+5              35             10             17             
+4              42             9              23             
+step s1-commit: 
+    COMMIT;
+
+step s2-select: 
+	SELECT * FROM users_test_table ORDER BY value_1;
```
marcocitus: looks like value_1 is always 3 and therefore the output might change

Issue #1666, opened 2017-09-22
Procedure for changed IP on worker


taybin:
Situation: Citus cluster with master and workers is shutdown.  Backups for cluster nodes (including master) are restored on a different set of machines.  At this point, pg_dist_node is incorrect because the IPs are different.  Also the citus background task is detecting that it can't communicate with the other nodes and restarting the process because the connection is null.

I was able to fix pg_dist_node by UPDATEing the rows with the new IPs, but I had to make several attempts for each change because my client would get disconnected after each change by the background task.

Is there a better procedure I can use for this scenario?  Perhaps by calling `master_disable_node()` on each worker first?  The documentation wasn't clear on what would happen to shards existing on a disabled node.

Running 7.0.1

taybin:
I was able to work around this by using /etc/hosts to maintain a mapping to the IP.  I am still curious about the behavior of `master_disable_node()` though.

ozgune:
Hey @taybin -- we may prioritize a user-defined function that enables users to update their pg_dist_node metadata.

I had a clarification question on your original comment.

>> At this point, pg_dist_node is incorrect because the IPs are different. Also the citus background task is detecting that it can't communicate with the other nodes and restarting the process because the connection is null.

Could you elaborate a bit more on the behavior you observed here? When you change the IP address, does the coordinator node shut down and restart?

taybin:
Yes.  That is exactly what would happen.

I'm actually seeing something very similar to #1636 after doing a backup restore, where the coordinator node claims that it can't communicate with the workers, although queries go through successfully.  This causes it to fill the logs with "no connection to server", "connection error: worker1".

Restarting the coordinator postgres instance doesn't resolve the issue and neither does restarting the worker nodes.

I tried disabling the nodes via `master_disable_node()`, but that didn't fix the issue and then citus wouldn't let me reactivate the nodes, complaining about "could not find any healthy placement for shard 102008".

At the moment, I'm guessing this is happening because I didn't run `master_add_node()`, the `pg_dist_node` table was populated by the backup restore and `master_add_node()` does something special that short circuited by the table already being populated.  So I'm looking for something that will preserve the data, but still allow me to give the maintenance daemon a kick to get it working again.

taybin:
I figured out the issue.  Part configuration error, but maybe also a problem with the maintenance daemon.

[...]

Issue #1636, opened 2017-09-07
Maintenance daemon may skip cache invalidations


onderkalaci:
To start with, this issue does not lead to any critical issues except for generating many `WARNINGs` on the coordinator's log file. Once the coordinator is restarted, the `WARNINGs` go away. But seems worthwhile to backport to 7.0.

Steps:
1. Connect to the coordinator and remove a node
```
SELECT master_remove_node('localhost', 9702);
```

2. Check the coordinator's log file, you'll see lots of these warnings:
```
CONTEXT:  Citus maintenance daemon for database 12669 user 10
WARNING:  no connection to the server
CONTEXT:  while executing command on localhost:9702
	Citus maintenance daemon for database 12669 user 10
WARNING:  connection error: localhost:9702
DETAIL:  no connection to the server
```

The reason: 

With #1584 we've added a logic to reset the metadata OID caches on each iteration of deadlock detection. (See the reason and long discussions for that in the PR.)

That seems to lead this weird behavior. Once a node is deleted from `pg_dist_node`, the maintenance daemon gets a cache invalidation message, however, it might skip the invalidation since the OID for `pg_dist_node` might be `InvalidOid` at the time the invalidation happens. See the gdb output for better understanding (add break point to `InvalidateNodeRelationCacheCallback` and later remove a node while gdb attached to the daemon):

```
Breakpoint 1, InvalidateNodeRelationCacheCallback (argument=0, relationId=16494) at utils/metadata_cache.c:2766
2766		if (relationId == InvalidOid || relationId == MetadataCache.distNodeRelationId)
(gdb) p relationId
$1 = 16494
Current language:  auto; currently minimal
(gdb) p MetadataCache.distNodeRelationId
$2 = 0
```
(Node that `16494` is `pg_dist_node`'s `OID`)

(Thanks @furkansahin for reporting this)

anarazel:
Two possibilities:
- replace ClearMetadataOIDCache() with InvalidateSystemCaches()
- move workerNodeHashValid, LocalGroupId into the MetadataCache struct

PR #1691, opened 2017-10-10
Add master_update_node function

lithp:
Add function to allow manually changing the nodename/nodeport of nodes without calling raw SQL.

Some questions:
- Is `master_move_node` the right name? Some other alternatives:
  - `master_change_node_location`
  - `master_update_node`
- Is this the right signature? Maybe we should pass in the node's group, I think that's the way our API is slowly evolving?
- Is the locking too aggressive? This prevents any concurrent distributed queries, but I'm pretty sure we're careful about cache invalidations, and citus can handle the ip changing mid-query.
  - The user we have in mind wants to use this for managing failovers. A failover of the worker shouldn't block all queries to the rest of the nodes. Maybe we should only take out a lock on the row we're changing (happens automatically anyway), and update Citus to take locks on those rows before it opens any connections?


src/backend/distributed/citus--7.0-15--7.0-16.sql
```diff
@@ -0,0 +1,24 @@
+/* citus--7.0-15--7.0-16 */
+
+SET search_path = 'pg_catalog';
+
+CREATE OR REPLACE FUNCTION master_move_node(node_name text,
```
marcocitus: I think `master_update_node` seems more appropriate.

src/backend/distributed/citus--7.0-15--7.0-16.sql
```diff
@@ -0,0 +1,24 @@
...
+                                            node_port int,
+                                            new_node_name text,
+                                            new_node_port int)
+RETURNS void
+AS $$
+BEGIN
+  LOCK pg_dist_node IN ACCESS EXCLUSIVE MODE;
+
+  UPDATE pg_dist_node
```
marcocitus: Could we write this function in C for easier maintenance? (e.g. if there's a bug we want to backport then this becomes a whole lot easier if we don't need to redefine the function).

src/backend/distributed/citus--7.0-15--7.0-16.sql
```diff
@@ -0,0 +1,24 @@
+/* citus--7.0-15--7.0-16 */
+
+SET search_path = 'pg_catalog';
+
+CREATE OR REPLACE FUNCTION master_move_node(node_name text,
+                                            node_port int,
```
marcocitus: It'd be nicer to just use the node ID, since that makes the function idempotent and allows it to be called without knowledge of the current nodename and nodeport. That makes it easier to integrate in an external failover solution.

lithp: Okay, good points.

src/backend/distributed/citus--7.0-15--7.0-16.sql
lithp: Sure thing.

src/backend/distributed/citus--7.0-15--7.0-16.sql
lithp: I probably don't understand what you're saying because I don't see how this makes backports harder. To redefine the function we add an extension update file, that doesn't sound hard, and for something this small SQL is much easier to read and modify.

marcocitus: Say we release this function as part of 7.1. We start working on 7.2 and add new functions in citus--7.1-1--7.2-1.sql. Now we discover we really needed to do something extra in master_update_node to prevent problems and decide to backport to 7.1. We define a citus--7.1-1--7.1-2.sql, release a new package, and ask users to run ALTER EXTENSION after the install (otherwise they'll have downtime due to version checks). Now when it comes to upgrading from 7.1 to 7.2 we potentially need to consider two different upgrade paths. One with the patch and one without. We'll need to add an additional citus--7.1-2--7.2-1.sql file to enable upgrading after applying the patch. Every new patch adds yet another upgrade path we need to consider. If someone upgrades without the patch, postgres will search for the shortest path among the upgrade files and will thus pick citus--7.1-1--7.2-1.sql, skipping the patch. So now we have to put the function somewhere else in the upgrade path for those users. Then there are also users who run from master...

If master_update_node was a C function then we could just release a new binary. We'd also make it much easier to find the current version and do version control.

jasonmp85: If we _really_ want to write functions in PL/pgSQL or SQL, we can always spin up the required machinery and execute the body of our function based on some static string or whatnot. This static string could be updated in a patch update.

But @marcocitus' right otherwise… as much as I want to push for maintainability and using the simplest language for the job, the reality is we don't want to ever have an `ALTER EXTENSION` within a patch update; they're for major updates only; patches should require a simple package update and restart.

lithp:
I'm confused by this test error, I get the same error when testing on master.

lithp:
Opened https://github.com/citusdata/citus/issues/1694

src/backend/distributed/Makefile
```diff
@@ -169,6 +169,8 @@ $(EXTENSION)--7.0-14.sql: $(EXTENSION)--7.0-13.sql $(EXTENSION)--7.0-13--7.0-14.
 	cat $^ > $@
 $(EXTENSION)--7.0-15.sql: $(EXTENSION)--7.0-14.sql $(EXTENSION)--7.0-14--7.0-15.sql
 	cat $^ > $@
+$(EXTENSION)--7.0-16.sql: $(EXTENSION)--7.0-15.sql $(EXTENSION)--7.0-15--7.0-16.sql
```
marcocitus: We should be in 7.1 now.

src/backend/distributed/utils/node_metadata.c
```diff
@@ -454,6 +456,103 @@ ActivateNode(char *nodeName, int nodePort)
...
+	bool replace[Natts_pg_dist_node];
+
+	pgDistNode = heap_open(DistNodeRelationId(), RowExclusiveLock);
+	tupleDescriptor = RelationGetDescr(pgDistNode);
+
+	ScanKeyInit(&scanKey[0], Anum_pg_dist_node_nodeid,
+				BTEqualStrategyNumber, F_INT8EQ, Int32GetDatum(nodeId));
+
+	scanDescriptor = systable_beginscan(pgDistNode, InvalidOid, indexOK,
+										NULL, scanKeyCount, scanKey);
```
marcocitus: We should be able to use the `pg_dist_node_pkey` index here.

src/backend/distributed/utils/node_metadata.c
```diff
@@ -454,6 +456,103 @@ ActivateNode(char *nodeName, int nodePort)
...
+	 * - Ensure buggy code in Citus doesn't cause failures when the nodename/nodeport of
+	 *   a node changes mid-query
+	 * - Provide fencing during failover, after this function returns all connections
+	 *   will use the new node location.
+	 *
+	 * Drawback:
+	 * - This function blocks until all previous queries have finished. This means that
+	 *   long-running queries will prevent failover.
+	 */
+	LockRelationOid(DistNodeRelationId(), ExclusiveLock);
```
marcocitus: I think this should be `AccessExclusiveLock` to actually provide fencing. Even writes only take an `AccessShareLock` on `pg_dist_node`

lithp: Yes, you're right

src/backend/distributed/Makefile
lithp: Updated

lithp:
Addressed all comments and rebased onto #1703 to make the diff smaller

metdos:
@lithp, I suppose this covers the feature request in #1666?

lithp:
@metdos Yes and no. This is indeed a "procedure for changing the IP of a worker" but it's just an official way of running the SQL taybin was already running. I think #1666 should be treated as a bug report, this UDF will likely crash in the exact same places taybin's UPDATE's are crashing.

marcocitus:
The crash in #1666 is probably a combination of #1636 and #1660.

Issue #1256, opened 2017-02-24
COUNT DISTINCT on/group-by partition column in Join query errors out


saicitus:
**Query with COUNT DISTINCT on partition_column:**
```
select count(distinct(t1.partition_column)) from table1 t1 join table2 t2 
on t1.partition_column = t2.partition_column;
```
**Query with COUNT DISTINCT on non_partition_column and group by partition_column:**
```
select t1.partition_column, count(distinct(t1.non_partition_column)) from table1 t1 join table2 t2 
on t1.partition_column = t2.partition_column GROUP BY  t1.partition_column;
```
**Error:**
```
ERROR:  cannot compute aggregate (distinct)
DETAIL:  table partitioning is unsuitable for aggregate (distinct)
HINT:  You can load the hll extension from contrib packages and enable distinct approximations.
```
Both the above queries are push-downable to the worker shards.

**Use approximations to solve the issue:**
```
set citus.count_distinct_error_rate to 0.0001;
select count(distinct(t1.partition_column)) from table1 t1 join table2 t2 
on t1.partition_column = t2.partition_column;
select t1.partition_column, count(distinct(t1.non_partition_column)) from table1 t1 join table2 t2 
on t1.partition_column = t2.partition_column GROUP BY  t1.partition_column;
```
Above query runs.

lithp:
What exactly is the problem here? Current this ticket reports that our error message says the correct thing and helps the user run their queries :)

marcocitus:
That the query can be pushed down as we do with count distinct queries without joins, but we error out instead of pushing it down due to a minor technical issue.

Issue #1021, opened 2016-12-07
expand count distinct support to join queries


mtuncer:
We do support count distinct on a single table if the distinct operation is on the partition column, so that
```sql
select count(distinct l_orderkey) from lineitem;
```

is a valid query.

However, we do not support count distinct on a join even if it is on the partition column. We should be able to support
```sql
select count(distinct l_orderkey) from lineitem join orders on (l_orderkey = o_orderkey);
```

We should be verifying
- joined tables are co-located
- join is on the partition column (somewhat implied since it is not a re-partition join)
- count distinct operation is on the partition column of either table.

PR #1692, opened 2017-10-11
 Count distinct feature

mtuncer:
Expands count distinct coverage by allowing more cases. We used to support count distinct only if we can push down distinct aggregate to worker query i.e. the count distinct clause was on the partition column of the table, or there was a grouping on the partition column.

Now we can support
- non-partition columns, with or without grouping on partition column
- partition, and non partition column in the same query
- having clause
- single table subqueries
- insert into select queries
- join queries where count distinct is on partition, or non-partition column.

We  first try to push down aggregate to worker query (original case), if we can't then we modify worker query to return distinct columns to coordinator node. We do that by adding distinct column targets to group by clauses. Then we perform count distinct operation on the coordinator node.

Note that, this approach has some performance implications as mentioned [here](https://github.com/citusdata/citus_docs/issues/476#issuecomment-337478187)

This work should reduce the cases where HLL is used as it can address anything that HLL can. However, if we start having performance issues due to very large number rows, then we can recommend hll use.

Fixes : #1256 
Fixes: #1021 

Following tests are performed
- count distinct on partition column of append, range, hash partitioned table
- count distinct on non partition column of reference, append, range, hash partitioned table
- mixed count/count distinct on partition/non partition columns
- count distinct on having clause
- count distinct on single table repartition subqueries
- count distinct on partition/non-partition column of join queries
- count distinct with filter
- count distinct in insert into select queries (these has to have partition column on select target and group by on partition column, thus existing behavior was already fine)

### DIFF
--- a/src/backend/distributed/planner/multi_logical_optimizer.c
+++ b/src/backend/distributed/planner/multi_logical_optimizer.c
@@ -62,15 +62,15 @@ double CountDistinctErrorRate = 0.0; /* precision of count(distinct) approximate
 
 typedef struct MasterAggregateWalkerContext
 {
-	bool repartitionSubquery;
 	AttrNumber columnId;
+	bool pullDistinctColumns;
 } MasterAggregateWalkerContext;
 
 typedef struct WorkerAggregateWalkerContext
 {
-	bool repartitionSubquery;
 	List *expressionList;
 	bool createGroupByClause;
+	bool pullDistinctColumns;
 } WorkerAggregateWalkerContext;
 
 
@@ -117,7 +117,9 @@ static void ApplyExtendedOpNodes(MultiExtendedOp *originalNode,
 								 MultiExtendedOp *masterNode,
 								 MultiExtendedOp *workerNode);
 static void TransformSubqueryNode(MultiTable *subqueryNode, List *tableNodeList);
-static MultiExtendedOp * MasterExtendedOpNode(MultiExtendedOp *originalOpNode);
+static MultiExtendedOp * MasterExtendedOpNode(MultiExtendedOp *originalOpNode,
+											  bool groupedByDisjointPartitionColumn,
+											  List *tableNodeList);
 static Node * MasterAggregateMutator(Node *originalNode,
 									 MasterAggregateWalkerContext *walkerContext);
 static Expr * MasterAggregateExpression(Aggref *originalAggregate,
@@ -126,7 +128,14 @@ static Expr * MasterAverageExpression(Oid sumAggregateType, Oid countAggregateTy
 									  AttrNumber *columnId);
 static Expr * AddTypeConversion(Node *originalAggregate, Node *newExpression);
 static MultiExtendedOp * WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
-											  bool groupedByDisjointPartitionColumn);
+											  bool groupedByDisjointPartitionColumn,
+											  List *tableNodeList);
+static bool HasNonPartitionColumnDistinctAgg(List *targetEntryList, Node *havingQual,
+											 List *tableNodeList);
+static bool ShouldPullDistinctColumn(bool repartitionSubquery,
+									 bool groupedByDisjointPartitionColumn,
+									 bool hasNonPartitionColumnDistinctAgg);
+static bool PartitionColumnInTableList(Var *column, List *tableNodeList);
 static bool WorkerAggregateWalker(Node *node,
 								  WorkerAggregateWalkerContext *walkerContext);
 static List * WorkerAggregateExpressionList(Aggref *originalAggregate,
@@ -135,6 +144,7 @@ static AggregateType GetAggregateType(Oid aggFunctionId);
 static Oid AggregateArgumentType(Aggref *aggregate);
 static Oid AggregateFunctionOid(const char *functionName, Oid inputType);
 static Oid TypeOid(Oid schemaId, const char *typeName);
+static SortGroupClause * CreateSortGroupClause(Var *column);
 
 /* Local functions forward declarations for count(distinct) approximations */
 static char * CountDistinctHashFunctionName(Oid argumentType);
@@ -261,9 +271,12 @@ MultiLogicalPlanOptimize(MultiTreeRoot *multiLogicalPlan)
 	groupedByDisjointPartitionColumn = GroupedByDisjointPartitionColumn(tableNodeList,
 																		extendedOpNode);
 
-	masterExtendedOpNode = MasterExtendedOpNode(extendedOpNode);
+	masterExtendedOpNode = MasterExtendedOpNode(extendedOpNode,
+												groupedByDisjointPartitionColumn,
+												tableNodeList);
 	workerExtendedOpNode = WorkerExtendedOpNode(extendedOpNode,
-												groupedByDisjointPartitionColumn);
+												groupedByDisjointPartitionColumn,
+												tableNodeList);
 
 	ApplyExtendedOpNodes(extendedOpNode, masterExtendedOpNode, workerExtendedOpNode);
 
@@ -1168,9 +1181,12 @@ TransformSubqueryNode(MultiTable *subqueryNode, List *tableNodeList)
 	MultiNode *collectChildNode = ChildNode((MultiUnaryNode *) collectNode);
 	bool groupedByDisjointPartitionColumn =
 		GroupedByDisjointPartitionColumn(tableNodeList, extendedOpNode);
-	MultiExtendedOp *masterExtendedOpNode = MasterExtendedOpNode(extendedOpNode);
+	MultiExtendedOp *masterExtendedOpNode =
+		MasterExtendedOpNode(extendedOpNode, groupedByDisjointPartitionColumn,
+							 tableNodeList);
 	MultiExtendedOp *workerExtendedOpNode =
-		WorkerExtendedOpNode(extendedOpNode, groupedByDisjointPartitionColumn);
+		WorkerExtendedOpNode(extendedOpNode, groupedByDisjointPartitionColumn,
+							 tableNodeList);
 	MultiPartition *partitionNode = CitusMakeNode(MultiPartition);
 	List *groupClauseList = extendedOpNode->groupClauseList;
 	List *targetEntryList = extendedOpNode->targetList;
@@ -1232,7 +1248,9 @@ TransformSubqueryNode(MultiTable *subqueryNode, List *tableNodeList)
  * worker nodes' results.
  */
 static MultiExtendedOp *
-MasterExtendedOpNode(MultiExtendedOp *originalOpNode)
+MasterExtendedOpNode(MultiExtendedOp *originalOpNode,
+					 bool groupedByDisjointPartitionColumn,
+					 List *tableNodeList)
 {
 	MultiExtendedOp *masterExtendedOpNode = NULL;
 	List *targetEntryList = originalOpNode->targetList;
@@ -1244,14 +1262,22 @@ MasterExtendedOpNode(MultiExtendedOp *originalOpNode)
 	MultiNode *childNode = ChildNode((MultiUnaryNode *) originalOpNode);
 	MasterAggregateWalkerContext *walkerContext = palloc0(
 		sizeof(MasterAggregateWalkerContext));
+	bool hasNonPartitionColumnDistinctAgg = false;
+	bool repartitionSubquery = false;
 
 	walkerContext->columnId = 1;
-	walkerContext->repartitionSubquery = false;
 
 	if (CitusIsA(parentNode, MultiTable) && CitusIsA(childNode, MultiCollect))
 	{
-		walkerContext->repartitionSubquery = true;
+		repartitionSubquery = true;
 	}
+
+	hasNonPartitionColumnDistinctAgg = HasNonPartitionColumnDistinctAgg(targetEntryList,
+																		originalHavingQual,
+																		tableNodeList);
+	walkerContext->pullDistinctColumns =
+		ShouldPullDistinctColumn(repartitionSubquery, groupedByDisjointPartitionColumn,
+								 hasNonPartitionColumnDistinctAgg);
 
 	/* iterate over original target entries */
 	foreach(targetEntryCell, targetEntryList)
@@ -1266,7 +1292,6 @@ MasterExtendedOpNode(MultiExtendedOp *originalOpNode)
 		{
 			Node *newNode = MasterAggregateMutator((Node *) originalExpression,
 												   walkerContext);
-
 			newExpression = (Expr *) newNode;
 		}
 		else
@@ -1379,7 +1404,7 @@ MasterAggregateExpression(Aggref *originalAggregate,
 
 	if (aggregateType == AGGREGATE_COUNT && originalAggregate->aggdistinct &&
 		CountDistinctErrorRate == DISABLE_DISTINCT_APPROXIMATION &&
-		walkerContext->repartitionSubquery)
+		walkerContext->pullDistinctColumns)
 	{
 		Aggref *aggregate = (Aggref *) copyObject(originalAggregate);
 		List *varList = pull_var_clause_default((Node *) aggregate);
@@ -1769,13 +1794,14 @@ AddTypeConversion(Node *originalAggregate, Node *newExpression)
  * function to create aggregates for the worker nodes. Also, the function checks
  * if we can push down the limit to worker nodes; and if we can, sets the limit
  * count and sort clause list fields in the new operator node. It provides special
- * treatment for count distinct operator if it is used in repartition subqueries.
- * Each column in count distinct aggregate is added to target list, and group by
- * list of worker extended operator.
+ * treatment for count distinct operator if it is used in repartition subqueries
+ * or on non-partition columns. Each column in count distinct aggregate is added
+ * to target list, and group by list of worker extended operator.
  */
 static MultiExtendedOp *
 WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
-					 bool groupedByDisjointPartitionColumn)
+					 bool groupedByDisjointPartitionColumn,
+					 List *tableNodeList)
 {
 	MultiExtendedOp *workerExtendedOpNode = NULL;
 	MultiNode *parentNode = ParentNode((MultiNode *) originalOpNode);
@@ -1790,27 +1816,36 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
 		palloc0(sizeof(WorkerAggregateWalkerContext));
 	Index nextSortGroupRefIndex = 0;
 	bool queryHasAggregates = false;
+	bool enableLimitPushdown = true;
+	bool hasNonPartitionColumnDistinctAgg = false;
+	bool repartitionSubquery = false;
 
-	walkerContext->repartitionSubquery = false;
 	walkerContext->expressionList = NIL;
+
+	/* find max of sort group ref index */
+	foreach(targetEntryCell, targetEntryList)
+	{
+		TargetEntry *targetEntry = (TargetEntry *) lfirst(targetEntryCell);
+		if (targetEntry->ressortgroupref > nextSortGroupRefIndex)
+		{
+			nextSortGroupRefIndex = targetEntry->ressortgroupref;
+		}
+	}
+
+	/* next group ref index starts from max group ref index + 1 */
+	nextSortGroupRefIndex++;
 
 	if (CitusIsA(parentNode, MultiTable) && CitusIsA(childNode, MultiCollect))
 	{
-		walkerContext->repartitionSubquery = true;
-
-		/* find max of sort group ref index */
-		foreach(targetEntryCell, targetEntryList)
-		{
-			TargetEntry *targetEntry = (TargetEntry *) lfirst(targetEntryCell);
-			if (targetEntry->ressortgroupref > nextSortGroupRefIndex)
-			{
-				nextSortGroupRefIndex = targetEntry->ressortgroupref;
-			}
-		}
-
-		/* next group ref index starts from max group ref index + 1 */
-		nextSortGroupRefIndex++;
+		repartitionSubquery = true;
 	}
+
+	hasNonPartitionColumnDistinctAgg = HasNonPartitionColumnDistinctAgg(targetEntryList,
+																		havingQual,
+																		tableNodeList);
+	walkerContext->pullDistinctColumns =
+		ShouldPullDistinctColumn(repartitionSubquery, groupedByDisjointPartitionColumn,
+								 hasNonPartitionColumnDistinctAgg);
 
 	/* iterate over original target entries */
 	foreach(targetEntryCell, targetEntryList)
@@ -1821,6 +1856,7 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
 		ListCell *newExpressionCell = NULL;
 		bool hasAggregates = contain_agg_clause((Node *) originalExpression);
 
+		/* reset walker context */
 		walkerContext->expressionList = NIL;
 		walkerContext->createGroupByClause = false;
 
@@ -1852,26 +1888,19 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
 			 */
 			if (IsA(newExpression, Var) && walkerContext->createGroupByClause)
 			{
-				Var *column = (Var *) newExpression;
-				Oid lessThanOperator = InvalidOid;
-				Oid equalsOperator = InvalidOid;
-				bool hashable = false;
-				SortGroupClause *groupByClause = makeNode(SortGroupClause);
-
-				get_sort_group_operators(column->vartype, true, true, true,
-										 &lessThanOperator, &equalsOperator, NULL,
-										 &hashable);
-				groupByClause->eqop = equalsOperator;
-				groupByClause->hashable = hashable;
-				groupByClause->nulls_first = false;
-				groupByClause->sortop = lessThanOperator;
+				Var *column = (Var *) newTargetEntry->expr;
+				SortGroupClause *groupByClause = CreateSortGroupClause(column);
+				newTargetEntry->ressortgroupref = nextSortGroupRefIndex;
 				groupByClause->tleSortGroupRef = nextSortGroupRefIndex;
 
 				groupClauseList = lappend(groupClauseList, groupByClause);
-
-				newTargetEntry->ressortgroupref = nextSortGroupRefIndex;
-
 				nextSortGroupRefIndex++;
+
+				/*
+				 * If we introduce new columns accompanied by a new group by clause,
+				 * than pushing down limits will cause incorrect results.
+				 */
+				enableLimitPushdown = false;
 			}
 
 			if (newTargetEntry->resname == NULL)
@@ -1921,6 +1950,23 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
 			newTargetEntry->resjunk = false;
 			newTargetEntry->resno = targetProjectionNumber;
 
+			if (IsA(newExpression, Var) && walkerContext->createGroupByClause)
+			{
+				Var *column = (Var *) newTargetEntry->expr;
+				SortGroupClause *groupByClause = CreateSortGroupClause(column);
+				newTargetEntry->ressortgroupref = nextSortGroupRefIndex;
+				groupByClause->tleSortGroupRef = nextSortGroupRefIndex;
+
+				groupClauseList = lappend(groupClauseList, groupByClause);
+				nextSortGroupRefIndex++;
+
+				/*
+				 * If we introduce new columns accompanied by a new group by clause,
+				 * than pushing down limits will cause incorrect results.
+				 */
+				enableLimitPushdown = false;
+			}
+
 			newTargetEntryList = lappend(newTargetEntryList, newTargetEntry);
 			targetProjectionNumber++;
 		}
@@ -1939,11 +1985,14 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
 
 	workerExtendedOpNode->groupClauseList = groupClauseList;
 
-	/* if we can push down the limit, also set related fields */
-	workerExtendedOpNode->limitCount = WorkerLimitCount(originalOpNode,
-														groupedByDisjointPartitionColumn);
-	workerExtendedOpNode->sortClauseList =
-		WorkerSortClauseList(originalOpNode, groupedByDisjointPartitionColumn);
+	if (enableLimitPushdown)
+	{
+		/* if we can push down the limit, also set related fields */
+		workerExtendedOpNode->limitCount = WorkerLimitCount(originalOpNode,
+															groupedByDisjointPartitionColumn);
+		workerExtendedOpNode->sortClauseList =
+			WorkerSortClauseList(originalOpNode, groupedByDisjointPartitionColumn);
+	}
 
 	/*
 	 * If grouped by a partition column whose values are shards have disjoint sets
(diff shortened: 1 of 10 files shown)


src/test/regress/input/multi_agg_distinct.source
```diff
@@ -115,4 +113,11 @@ SELECT l_orderkey, count(distinct l_partkey) INTO range_results FROM lineitem_ra
 -- they should return the same results
 SELECT * FROM hash_results h, range_results r WHERE h.l_orderkey = r.l_orderkey AND h.count != r.count;
 
+-- agg(distinct) is allowed if we group by non-partition column
```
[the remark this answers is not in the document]
pykello: Is this allowed only when `agg(distinct partition_column)`? If yes, please update the comments. If no, please add some tests where distinct is not on partition column.

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -1959,6 +2013,139 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
...
+}
+
+
+/*
+ * CanPullDistinctColumn returns true if distinct aggregate should pull
+ * individual columns from worker to master and evaluate aggregate operation
+ * at master.
+ */
+static bool
+CanPullDistinctColumn(bool groupedByDisjointPartitionColumn,
```
pykello: From comments it seems that this should be named `ShouldPullDistinctColumn`.

src/test/regress/input/multi_agg_distinct.source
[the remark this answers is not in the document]
mtuncer: only count distinct is supported. array_agg(distinct) is also not supported.
updated the comments, added new test cases.

mtuncer:
I run some numbers on performance
I had a table t1(a int, b int, c int) with 15M rows with each item has about 1M distinct values. Table is hash partitioned among a with 4 shards to 4 workers.

select count(distinct a) from t1 --> took ~8 seconds in which all time is spent on workers
select count(distinct b) from t1 --> took ~23 seconds (worker : 17, coordinator: 6)
select count(distinct a), count(distinct b), count(distinct c) --> 105 secs (w: 41 c: 64)
select a, count(distinct b) from t1 group by a --> took ~22 seconds (w: 17.5, c: 4.5)
select b, count(distinct c) from t1 group by b --> took ~83 seconds (w: 35, c: 48)

Then I enabled hll by setting error rate to 0.01;
select count(distinct a) from t1 --> dropped to 2.8 secs from 8 (almost all on worker)
select count(distinct b) from t1 --> dropped to 2.8 secs from 8 (almost all on worker)
select count(distinct a), count(distinct b), count(distinct c) --> 5 secs (almost all on worker)
select a, count(distinct b) from t1 group by a --> took ~376 seconds (w: 28, c: 348)
select b, count(distinct c) from a1 group by b --> took ~415 seconds (w: 40, c: 375)

Note: coordinator time also contains time spend during data transfers, and serialize/deserialize operations.

I should mention that these tests are performed on a memory restricted device, therefore the last 2 hll tests might not be representative as it caused postgres to use more memory than available and caused memory swapping to disk.

metdos:
> I run some numbers on performance

@mtuncer, could you also add other manual tests you run in the PR description.

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -1959,6 +2009,139 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
...
+}
+
+
+/*
+ * ShouldPullDistinctColumn returns true if distinct aggregate should pull
+ * individual columns from worker to master and evaluate aggregate operation
+ * at master.
+ */
+static bool
+ShouldPullDistinctColumn(bool groupedByDisjointPartitionColumn,
```
metdos: This function is a bit confusing. Could you describe in which conditions we want to pull columns to master and run aggregate just on the master node? Maybe also refactor a bit?

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -2355,6 +2538,30 @@ GroupedByDisjointPartitionColumn(List *tableNodeList, MultiExtendedOp *opNode)
 }
 
 
+static SortGroupClause *
+createGroupByClause(TargetEntry *targetEntry, Index nextSortGroupRefIndex)
```
metdos: Missing comments, starts with lower case.

metdos: Maybe rename to `CreateSortGroupClause()`

metdos: Can we do this on the caller side, it puts more functionality than the function name sounds
```
targetEntry->ressortgroupref = nextSortGroupRefIndex;
```

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -2661,8 +2868,14 @@ ErrorIfUnsupportedAggregateDistinct(Aggref *aggregateExpression,
 	}
 	else if (distinctSupported)
 	{
-		bool supports = TablePartitioningSupportsDistinct(tableNodeList, extendedOpNode,
-														  distinctColumn);
+		bool supports = true;
+
+		if (aggregateType != AGGREGATE_COUNT)
```
metdos: we can replace this by updating the else if
```c
else if (distinctSupported && aggregateType != AGGREGATE_COUNT)
```

metdos: I think we can modify whole this block into this one;

```c
	if (distinctSupported)
	{
		Var *distinctColumn = AggregateDistinctColumn(aggregateExpression);

		if (distinctColumn == NULL)
		{
			/*
			 * If the query has a single table, and table is grouped by partition column,
			 * then we support count distincts even distinct column can not be identified.
			 */
			distinctSupported = TablePartitioningSupportsDistinct(tableNodeList,
																  extendedOpNode,
																  distinctColumn);
			if (!distinctSupported)
			{
				errorDetail = "aggregate (distinct) on complex expressions is unsupported";
			}
		}
		else if (aggregateType != AGGREGATE_COUNT)
		{
			bool supports = TablePartitioningSupportsDistinct(tableNodeList, extendedOpNode,
															 distinctColumn);
			if (!supports)
			{
				distinctSupported = false;
				errorDetail = "table partitioning is unsuitable for aggregate (distinct)";
			}
		}
	}
```

src/backend/distributed/planner/multi_logical_optimizer.c
metdos: And is this `PullDistinctColums` or `PullColumns` because if `repartitionSubquery` we just return true.

mtuncer: it pulls only distinct columns, if it is repartition query it should pull columns regardless of partition type or group by.

mtuncer: added pull cases

src/backend/distributed/planner/multi_logical_optimizer.c
mtuncer: makes sense fixed

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -1775,7 +1810,8 @@ AddTypeConversion(Node *originalAggregate, Node *newExpression)
  */
```
metdos: We should also extend the comment here for count distinct on non-distribution columns "treatment for count distinct operator if it is used in repartition subqueries"

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -1826,10 +1867,19 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
...
+			walkerContext->pullDistinctColumns = ShouldPullDistinctColumn(
+				groupedByDisjointPartitionColumn, repartitionSubquery,
+				hasNonPartitionColumnDistinctAgg,
+				(Node *) originalExpression, tableNodeList);
+
 			WorkerAggregateWalker((Node *) originalExpression, walkerContext);
 
 			newExpressionList = walkerContext->expressionList;
 			queryHasAggregates = true;
+			if (walkerContext->createGroupByClause)
```
metdos: explaining why this is the case with a comment and putting a new line before the `if` would be nice

src/backend/distributed/planner/multi_master_planner.c
```diff
@@ -281,6 +285,24 @@ BuildAggregatePlan(Query *masterQuery, Plan *subPlan)
 		column->varno = OUTER_VAR;
 	}
 
+	allTargetList = copyObject(masterQuery->targetList);
```
metdos: Can we put this block into a function `HasDistinctAggregate()`?

src/backend/distributed/planner/multi_master_planner.c
```diff
@@ -292,8 +314,16 @@ BuildAggregatePlan(Query *masterQuery, Plan *subPlan)
 			ereport(ERROR, (errmsg("grouped column list cannot be hashed")));
 		}
 
-		/* switch to hashed aggregate strategy to allow grouping */
-		aggregateStrategy = AGG_HASHED;
+		if (hasDistinctAggregate)
```
metdos: Can we explain with a comment why do we need to use `AGG_SORTED` for distinct aggregate.

src/backend/distributed/planner/multi_master_planner.c
```diff
@@ -351,12 +383,25 @@ BuildDistinctPlan(Query *masterQuery, Plan *subPlan)
 		column->varno = OUTER_VAR;
 	}
 
+	foreach(targetCell, targetList)
```
metdos: Can we use `HasDistinctAggregate()` here?

src/backend/distributed/planner/multi_logical_optimizer.c
mtuncer: done

src/backend/distributed/planner/multi_logical_optimizer.c
mtuncer: added comments, moved to the group creation block below

src/test/regress/input/multi_agg_distinct.source
```diff
@@ -115,4 +113,16 @@ SELECT l_orderkey, count(distinct l_partkey) INTO range_results FROM lineitem_ra
 -- they should return the same results
 SELECT * FROM hash_results h, range_results r WHERE h.l_orderkey = r.l_orderkey AND h.count != r.count;
 
+-- count(distinct) is allowed if we group by non-partition column
+SELECT l_partkey, count(distinct l_orderkey) INTO hash_results_np FROM lineitem_hash GROUP BY l_partkey;
+SELECT  l_partkey, count(distinct l_orderkey) INTO range_results_np FROM lineitem_range GROUP BY l_partkey;
```
metdos: extra empty space

src/test/regress/output/multi_agg_distinct.source
```diff
@@ -73,12 +73,14 @@ SELECT p_partkey, count(distinct l_orderkey) FROM lineitem_range, part
...
-ERROR:  cannot compute aggregate (distinct)
-DETAIL:  table partitioning is unsuitable for aggregate (distinct)
-HINT:  You can load the hll extension from contrib packages and enable distinct approximations.
+ count 
+-------
+ 11661
+(1 row)
+
+-- Check that we don't support complex expressions.
 SELECT count(distinct (l_orderkey + 1)) FROM lineitem_range;
```
metdos: We can leave this out of the scope, but I'm just curious about why we can't do this. Isn't this similar to what we did for distinct support on non-partition columns?

src/test/regress/output/multi_complex_count_distinct.source
```diff
@@ -142,6 +274,45 @@ SELECT
       12417 |     3
 (10 rows)
 
+-- group by on non-partition column
```
metdos: it would be useful to have also count(distinct)'s without a filter on the same query such as;


```
SELECT                                                                                                                                                                                                                                                                 l_suppkey, count(DISTINCT l_partkey) FILTER (WHERE l_shipmode = 'AIR'), count(distinct l_partkey), count(distinct l_shipdate)                                                                                                                                                 FROM lineitem                                                                                                                                                                                                                                                                   GROUP BY l_suppkey                                                                                                                                                                                                                                                              ORDER BY 2 DESC, 1 DESC                                                                                                                                                                                                                                                         LIMIT 10;
```

metdos:
@mtuncer could you also add explain regression tests to see we're running expected type of queries on both the master node and worker nodes.

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -117,7 +117,9 @@ static void ApplyExtendedOpNodes(MultiExtendedOp *originalNode,
 								 MultiExtendedOp *masterNode,
 								 MultiExtendedOp *workerNode);
 static void TransformSubqueryNode(MultiTable *subqueryNode, List *tableNodeList);
-static MultiExtendedOp * MasterExtendedOpNode(MultiExtendedOp *originalOpNode);
+static MultiExtendedOp * MasterExtendedOpNode(MultiExtendedOp *originalOpNode, bool
```
metdos: visual: we can take bool to the second line

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -1244,28 +1265,41 @@ MasterExtendedOpNode(MultiExtendedOp *originalOpNode)
...
 	foreach(targetEntryCell, targetEntryList)
 	{
 		TargetEntry *originalTargetEntry = (TargetEntry *) lfirst(targetEntryCell);
 		TargetEntry *newTargetEntry = copyObject(originalTargetEntry);
 		Expr *originalExpression = originalTargetEntry->expr;
 		Expr *newExpression = NULL;
-
 		bool hasAggregates = contain_agg_clause((Node *) originalExpression);
+
+
```
metdos: Two new lines. I think we can keep `bool hasAggregates` and `if (hasAggregates)` together.

src/backend/distributed/planner/multi_master_planner.c
mtuncer: done

src/backend/distributed/planner/multi_master_planner.c
mtuncer: yes we can

src/test/regress/input/multi_agg_distinct.source
mtuncer: fixed

src/test/regress/output/multi_agg_distinct.source
mtuncer: we should be able to allow that if needed in the future.
It requires evaluation expression inside distinct clause. Once we reliably do that, we could also allow more complex expressions like ```case ... when ...``` and even function calls.

I haven't check the case with approximate distinct using hll. This single column restriction might be coming from there.

src/test/regress/output/multi_complex_count_distinct.source
mtuncer: added

src/backend/distributed/planner/multi_logical_optimizer.c
mtuncer: citus_indent, fixed

src/backend/distributed/planner/multi_logical_optimizer.c
mtuncer: ok

src/backend/distributed/planner/multi_master_planner.c
```diff
@@ -351,12 +383,25 @@ BuildDistinctPlan(Query *masterQuery, Plan *subPlan)
...
+		}
+	}
+
 	/*
 	 * Create group by plan with HashAggregate if all distinct
 	 * members are hashable, Otherwise create sort+unique plan.
 	 */
 	distinctClausesHashable = grouping_is_hashable(distinctClauseList);
-	if (distinctClausesHashable)
+	if (distinctClausesHashable && !containstDistinctAggregate)
```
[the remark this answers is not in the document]
mtuncer: ok

src/backend/distributed/planner/multi_master_planner.c
```diff
@@ -281,6 +283,8 @@ BuildAggregatePlan(Query *masterQuery, Plan *subPlan)
 		column->varno = OUTER_VAR;
 	}
 
+	hasDistinctAggregate = HasDistinctAggregate(masterQuery);
```
metdos: We can move declaration and assignment of `hasDistinctAggregate ` to inside `if (groupColumnCount > 0)` where it is only used.

src/backend/distributed/planner/multi_master_planner.c
```diff
@@ -292,8 +296,21 @@ BuildAggregatePlan(Query *masterQuery, Plan *subPlan)
 			ereport(ERROR, (errmsg("grouped column list cannot be hashed")));
 		}
 
-		/* switch to hashed aggregate strategy to allow grouping */
-		aggregateStrategy = AGG_HASHED;
+		if (hasDistinctAggregate)
+		{
+			/*
```
metdos: generally, we put comments outside of the if instead of inside

src/backend/distributed/planner/multi_master_planner.c
```diff
@@ -292,8 +296,21 @@ BuildAggregatePlan(Query *masterQuery, Plan *subPlan)
...
+			 * Postgres hash aggregate does not support distinct aggregate/order by
```
metdos: I think this is not just `order by` but `order by with aggregate`?

src/backend/distributed/planner/multi_master_planner.c
mtuncer: sounds good

src/backend/distributed/planner/multi_master_planner.c
mtuncer: ok

src/backend/distributed/planner/multi_master_planner.c
mtuncer: ok

src/backend/distributed/planner/multi_master_planner.c
```diff
@@ -346,6 +333,39 @@ BuildAggregatePlan(Query *masterQuery, Plan *subPlan)
...
+{
+	List *targetVarList = NIL;
+	List *havingVarList = NIL;
+	List *allColumnList = NIL;
+	ListCell *allColumnCell = NULL;
+
+	targetVarList = pull_var_clause((Node *) masterQuery->targetList,
+									PVC_INCLUDE_AGGREGATES);
+	havingVarList = pull_var_clause(masterQuery->havingQual, PVC_INCLUDE_AGGREGATES);
+	allColumnList = list_concat(targetVarList, havingVarList);
```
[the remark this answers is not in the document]
mtuncer: ok

mtuncer:
@metdos there are 2 possible issues
- explain fails on a query that runs fine without explain
- there is a difference in one explain query output.

I think the first one is related our general explain issue. I will look into the second one (most likely drop or modify that test)

metdos:
> I think the first one is related our general explain issue.

If this is the case, could open an issue or comment on an existing one with the example query.

src/backend/distributed/planner/multi_master_planner.c
```diff
@@ -351,12 +402,15 @@ BuildDistinctPlan(Query *masterQuery, Plan *subPlan)
...
+
 	/*
 	 * Create group by plan with HashAggregate if all distinct
-	 * members are hashable, Otherwise create sort+unique plan.
+	 * members are hashable, and not containing distinct aggregate.
+	 * Otherwise create sort+unique plan.
 	 */
 	distinctClausesHashable = grouping_is_hashable(distinctClauseList);
-	if (distinctClausesHashable)
+	if (distinctClausesHashable && !containstDistinctAggregate)
```
metdos: I think we can rename `containstDistinctAggregate` to `hasDistinctAggregate` in order to be consistent.

Visual. I would move `distinctClausesHashable = grouping_is_hashable(distinctClauseList);` below `HasDistinctAggregate ()` above.

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -136,6 +147,8 @@ static Oid AggregateArgumentType(Aggref *aggregate);
 static Oid AggregateFunctionOid(const char *functionName, Oid inputType);
 static Oid TypeOid(Oid schemaId, const char *typeName);
```
metdos: I think we don't need this newline

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -1959,6 +2021,145 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
...
+static bool
+HasNonPartitionColumnDistinctAgg(List *targetEntryList, Node *havingQual,
+								 List *tableNodeList)
+{
+	List *targetVarList = pull_var_clause((Node *) targetEntryList,
+										  PVC_INCLUDE_AGGREGATES);
+	List *havingVarList = pull_var_clause((Node *) havingQual, PVC_INCLUDE_AGGREGATES);
+	List *aggregateCheckList = list_concat(targetVarList, havingVarList);
+	ListCell *aggregateCheckCell = NULL;
+	foreach(aggregateCheckCell, aggregateCheckList)
```
metdos: We can have new line here

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -1959,6 +2021,145 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
...
+		ListCell *varCell = NULL;
+		bool isPartitionColumn = false;
+
+		if (IsA(targetNode, Var))
+		{
+			continue;
+		}
+
+		Assert(IsA(targetNode, Aggref));
+		targetAgg = (Aggref *) targetNode;
```
metdos: may have a new line before Assert

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -1959,6 +2021,145 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
...
+		if (targetAgg->aggdistinct == NIL)
+		{
+			continue;
+		}
+
+		varList = pull_var_clause_default((Node *) targetAgg->args);
+
```
metdos: we can delete the newline

metdos: actually, do we need the NIL check? `foreach` should take care of it.

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -1959,6 +2021,145 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
...
+		{
+			continue;
+		}
+
+		foreach(varCell, varList)
+		{
+			Node *targetVar = (Node *) lfirst(varCell);
+			Assert(IsA(targetVar, Var));
+
+			isPartitionColumn = ExpressionContainsPartitionColumn(
```
metdos: Visual. maybe put `ExpressionContainsPartitionColumn` to next line

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -1232,7 +1251,9 @@ TransformSubqueryNode(MultiTable *subqueryNode, List *tableNodeList)
  * worker nodes' results.
  */
 static MultiExtendedOp *
-MasterExtendedOpNode(MultiExtendedOp *originalOpNode)
+MasterExtendedOpNode(MultiExtendedOp *originalOpNode, bool
```
metdos: `bool` is missing the variable on this line. Could you check for other examples and also fix them?

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -1959,6 +2021,145 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
...
+	List *targetColumnList = pull_var_clause_default((Node *) expression);
+	Var *firstColumn = NULL;
+	ListCell *tableNodeCell = NULL;
+
+	if (targetColumnList == NIL)
+	{
+		return false;
+	}
+
+	firstColumn = (Var *) linitial(targetColumnList);
```
metdos: Can `targetColumnList` have more than one colum(Var)?

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -1959,6 +2021,145 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
...
+
+
+/*
+ * ExpressionContainsPartitionColumn returns true if provided expression
+ * contains partition column reference from provided table node list.
+ * It also returns false if a column is partition column of an append
+ * distributed table.
+ */
+static bool
+ExpressionContainsPartitionColumn(Node *expression, List *tableNodeList)
```
metdos: Maybe, rename to `HasPartitionColumn`

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -1959,6 +2021,145 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
...
+ * at master.
+ *
+ * Pull cases are:
+ * - repartition subqueries
+ * - query has count distinct on a non-partition column on at least one target
+ * - count distinct is on a non-partition column and query is not
+ *   grouped on partition column
+ */
+static bool
+ShouldPullDistinctColumn(bool groupedByDisjointPartitionColumn,
```
metdos: I find a bit harder to follow this code path, based on on your comment what about refactoring like this (also reordered parameters):

```c
static bool
ShouldPullDistinctColumn(Node *expression, List *tableNodeList,
						bool repartitionSubquery,
						bool groupedByDisjointPartitionColumn,
						bool hasNonPartitionColumnDistinctAgg)
{
	if (repartitionSubquery)
	{
		return true;
	}

	if (!groupedByDisjointPartitionColumn)
	{
		bool hasPartitionColumn = false;

		if (hasNonPartitionColumnDistinctAgg)
		{
			return true;
		}

		hasPartitionColumn = ExpressionContainsPartitionColumn(expression, tableNodeList);
		if (!hasPartitionColumn)
		{
			return true;
		}
	}

	return false;
}
```

metdos: Also, is this correct? Can we even simply this?

src/backend/distributed/planner/multi_master_planner.c
mtuncer: ok

src/backend/distributed/planner/multi_logical_optimizer.c
mtuncer: removed

src/backend/distributed/planner/multi_logical_optimizer.c
mtuncer: ok

src/backend/distributed/planner/multi_logical_optimizer.c
mtuncer: we can remove it

src/backend/distributed/planner/multi_logical_optimizer.c
mtuncer: done

src/backend/distributed/planner/multi_logical_optimizer.c
mtuncer: ok

src/backend/distributed/planner/multi_logical_optimizer.c
mtuncer: It should not. We error out if count distinct has more than one column.

metdos: then maybe we can rename `firstColumn` to `distinctColumn`

mtuncer: sometimes it can :(
when there is a non-distinct aggregate

mtuncer: It should be more like, ```HasPartitionColumnOnly```

mtuncer: For our purpose : we are only interested in value of this function if a target expression is count distinct. In this case count(distinct ..) can except only columns. So we should be fine as soon as the function is called within this context.

mtuncer: updated the function comment. kept the ```firstColumn``` name the same

src/backend/distributed/planner/multi_logical_optimizer.c
mtuncer: looks okay to me

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -1959,6 +2023,146 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
...
+
+
+/*
+ * ShouldPullDistinctColumn returns true if distinct aggregate should pull
+ * individual columns from worker to master and evaluate aggregate operation
+ * at master.
+ *
+ * Pull cases are:
+ * - repartition subqueries
+ * - query has count distinct on a non-partition column on at least one target
```
metdos: Thinking about this a bit more, are not case-2 and case-3 same? Can we replace the second block with just;
```
	if (hasNonPartitionColumnDistinctAgg && !groupedByDisjointPartitionColumn)
	{
		return true;
	}
```

mtuncer: not exactly the same
more like
```c
if (groupedByDisjointPartitionColumn)
{
    return false;
}

if (hasNonPartitionColumnDistinctAgg && !groupedByDisjointPartitionColumn)
{
    return true;
}

hasPartitionColumn = HasPartitionColumn(expression, tableNodeList);
if (!hasPartitionColumn)
{
    return true;
}
return false;

```

metdos: What I am missing is that.
i. Do we suppose to call here for just count(distinct)?
ii. If yes, if `hasNonPartitionColumnDistinctAgg` is false, then `hasPartitionColumn` must be true?

mtuncer: correct.  ```hasPartitionColumn``` and ``hasNonPartitionColumnDistinctAgg``` are not independent.

```!hasPartitionColumn``` implies ```hasNonPartitionColumnDistinctAgg```

We should be able remove ```hasPartitionColumn`` checks

File: src/backend/distributed/planner/multi_master_planner.c [hunk 32]
```diff
@@ -287,13 +288,28 @@ BuildAggregatePlan(Query *masterQuery, Plan *subPlan)
 	/* if we have grouping, then initialize appropriate information */
 	if (groupColumnCount > 0)
 	{
+		bool hasDistinctAggregate = HasDistinctAggregate(masterQuery);
+
 		if (!grouping_is_hashable(groupColumnList))
```
metdos: It seems that this is (although it is old code) unnecessary. We can run `AGG_SORTED` if `grouping_is_hashable(groupColumnList)` is `false`. Could you refactor the code?

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -1168,9 +1181,12 @@ TransformSubqueryNode(MultiTable *subqueryNode, List *tableNodeList)
 	MultiNode *collectChildNode = ChildNode((MultiUnaryNode *) collectNode);
 	bool groupedByDisjointPartitionColumn =
 		GroupedByDisjointPartitionColumn(tableNodeList, extendedOpNode);
-	MultiExtendedOp *masterExtendedOpNode = MasterExtendedOpNode(extendedOpNode);
+	MultiExtendedOp *masterExtendedOpNode = MasterExtendedOpNode(extendedOpNode,
```
metdos: Minor visual. It would be better if we start `MasterExtendedOpNode` on the next line.

src/backend/distributed/planner/multi_master_planner.c
mtuncer: added checks for ```sortableness```!

src/backend/distributed/planner/multi_logical_optimizer.c
mtuncer: ok

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -1959,6 +2008,127 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
 
 
 /*
+ * HasNonPartitionColumnDistinctAgg returns true if target entry or having qualifier
+ * has non-partition column reference in aggregate definition. Note that, it only
```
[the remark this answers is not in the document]
mtuncer: done

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -1959,6 +2008,128 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
...
+	{
+		return true;
+	}
+
+	if (groupedByDisjointPartitionColumn)
+	{
+		return false;
+	}
+
+	if (!groupedByDisjointPartitionColumn && hasNonPartitionColumnDistinctAgg)
```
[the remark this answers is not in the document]
mtuncer: neatness yes, readability not so much.  I find it harder to follow that way.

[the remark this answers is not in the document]
mtuncer:
```c
	if (groupedByDisjointPartitionColumn)
	{
		return false;
	}
	else if (!groupedByDisjointPartitionColumn && hasNonPartitionColumnDistinctAgg)
	{
		return true;
	}
```

Code Complete 2, Chapter 19 Page 448 seem to like this approach better

metdos:
Looks good 🚢

src/backend/distributed/planner/multi_master_planner.c
```diff
@@ -287,13 +288,37 @@ BuildAggregatePlan(Query *masterQuery, Plan *subPlan)
...
+		 * Postgres hash aggregate strategy does not support distinct aggregates
+		 * in group and order by with aggregate operations.
+		 * see nodeAgg.c:build_pertrans_for_aggref(). In that case we use
+		 * sorted agg strategy, otherwise we use hash strategy.
+		 */
+		if (!groupingIsHashable || hasDistinctAggregate)
+		{
+			if (!groupingIsSortable)
+			{
+				ereport(ERROR, (errmsg("grouped column list must cannot be sorted"),
```
furkansahin: We might want to reword this error message. I think the right error message should be `grouped column list cannot be sorted`.

PR #1695, opened 2017-10-12
Check for updates.

pykello:
Sends a request to ```/v1/releases/latest?flavor=community``` once a day, which returns a response similar to ```{"version": "7.1.0", "major": 7, "minor": 1, "patch": 0}```. Then compares it with current Citus version, and if the latest release is newer, logs a ```LOG``` message.

### DIFF
--- a/src/backend/distributed/utils/maintenanced.c
+++ b/src/backend/distributed/utils/maintenanced.c
@@ -325,6 +325,13 @@ CitusMaintenanceDaemonMain(Datum main_arg)
 
 			if (statsCollectionSuccess)
 			{
+				/*
+				 * Checking for updates only when collecting statistics succeeds,
+				 * so we don't log update messages twice when we retry statistics
+				 * collection in a minute.
+				 */
+				CheckForUpdates();
+
 				nextStatsCollectionTime =
 					TimestampTzPlusMilliseconds(GetCurrentTimestamp(),
 												STATS_COLLECTION_TIMEOUT_MILLIS);
--- a/src/backend/distributed/utils/statistics_collection.c
+++ b/src/backend/distributed/utils/statistics_collection.c
@@ -34,6 +34,7 @@ PG_FUNCTION_INFO_V1(citus_server_id);
 #include "distributed/statistics_collection.h"
 #include "distributed/worker_manager.h"
 #include "lib/stringinfo.h"
+#include "utils/builtins.h"
 #include "utils/json.h"
 #include "utils/jsonb.h"
 
@@ -41,10 +42,19 @@ PG_FUNCTION_INFO_V1(citus_server_id);
 #include "utils/fmgrprotos.h"
 #endif
 
+static size_t CheckForUpdatesCallback(char *contents, size_t size, size_t count,
+									  void *userData);
+static bool JsonbFieldInt32(Jsonb *jsonb, const char *fieldName, int32 *result);
+static bool JsonbFieldStr(Jsonb *jsonb, const char *fieldName, StringInfo result);
 static uint64 NextPow2(uint64 n);
 static uint64 DistributedTablesSize(List *distTableOids);
-static bool SendHttpPostJsonRequest(const char *url, const char *postFields, long
-									timeoutSeconds);
+static bool UrlEncode(StringInfo buf, const char *str);
+static bool SendHttpPostJsonRequest(const char *url, const char *postFields,
+									long timeoutSeconds);
+static bool SendHttpGetJsonRequest(const char *url, long timeoutSeconds,
+								   curl_write_callback responseCallback);
+static bool PerformHttpRequest(CURL *curl);
+
 
 /* WarnIfSyncDNS warns if libcurl is compiled with synchronous DNS. */
 void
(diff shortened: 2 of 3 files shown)


src/backend/distributed/utils/statistics_collection.c
```diff
@@ -175,14 +305,23 @@ SendHttpPostJsonRequest(const char *url, const char *jsonObj, long timeoutSecond
...
+		if (responseCallback != NULL)
+		{
+			curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, responseCallback);
+		}
+
 		curlCode = curl_easy_perform(curl);
 		if (curlCode == CURLE_OK)
 		{
-			int httpCode = 0;
+			int64 httpCode = 0;
```
pykello: This fixes some out of bound memory access by the latter ```curl_easy_getinfo()```

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -86,7 +97,129 @@ CollectBasicUsageStatistics(void)
...
+ * CheckForUpdatesCallback receives the response for the request sent by
+ * CheckForUpdates(). It processes the response, and if there is a newer release
+ * of Citus available, logs a LOG message. This function returns 0 if there are
+ * any errors in the received response, which means we didn't consume the data.
+ * Otherwise, it returns (size * count) which means we consumed all of the data.
+ */
+static size_t
+CheckForUpdatesCallback(char *contents, size_t size, size_t count, void *userData)
+{
+	PG_TRY();
```
pykello: ```jsonb_in``` and ```JsonbFieldInt()``` (through ```pg_atoi```) can throw errors, so I wrapped everything inside a big ```PG_TRY()```.

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -86,7 +97,129 @@ CollectBasicUsageStatistics(void)
...
+
+		StringInfo responseNullTerminated = makeStringInfo();
+		appendBinaryStringInfo(responseNullTerminated, contents, size * count);
+		responseJson = DirectFunctionCall1(jsonb_in,
+										   CStringGetDatum(responseNullTerminated->data));
+
+		if (!JsonbFieldExists(responseJson, "version") ||
+			!JsonbFieldExists(responseJson, "major") ||
+			!JsonbFieldExists(responseJson, "minor") ||
+			!JsonbFieldExists(responseJson, "patch"))
```
pykello: These are to make sure ```jsonb_object_field_text``` doesn't return ```NULL```. We cannot use ```DirectFunctionCall``` for null-returning functions.

pykello: ```jsonb_object_field_text``` also works for object or array fields, but it returns null for ```null``` json values. If handling ```null``` values aren't easy in C code, should we skip handling them for now and leave it as TODO unless they cause crashes/etc.?

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -86,8 +100,138 @@ CollectBasicUsageStatistics(void)
...
+
+	return success;
+}
+
+
+/* CheckForUpdates queries Citus servers for newer releases of Citus. */
+void
+CheckForUpdates(void)
+{
+	SendHttpRequest(
```
pykello: Ignoring the return value. I don't know how we can use for it as timing of check for updates is same as statistics collection and I thought success/failure of statistics collection should rule there.

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -86,8 +100,138 @@ CollectBasicUsageStatistics(void)
...
+/*
+ * JsonbFieldInt returns integer value of the given field in the given JSONB
+ * object. Caller should have checked that the field exists before calling this
+ * function. If field value cannot be converted to an integer, throws an error.
+ */
+static int
+JsonbFieldInt(Datum jsonb, const char *fieldName)
+{
+	StringInfo fieldStr = JsonbFieldStr(jsonb, fieldName);
+	return pg_atoi(fieldStr->data, fieldStr->len, '\0');
```
jasonmp85: This is super circuitous: `JsonbFieldStr` gets the underlying integer, then converts it to text, then we're making a C string, then we're calling `pg_atoi`. Is there some way to use `findJsonbValueFromContainer` to accomplish what we want here?

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -86,8 +100,138 @@ CollectBasicUsageStatistics(void)
...
+	return success;
+}
+
+
+/* CheckForUpdates queries Citus servers for newer releases of Citus. */
+void
+CheckForUpdates(void)
+{
+	SendHttpRequest(
+		STATS_COLLECTION_HOST "/v1/releases/latest?edition=" CITUS_EDITION,
```
jasonmp85: Even though we (at present) control the values of `CITUS_EDITION`, I think it would be best to URL-encode this. Check out libcurl's functions to do so.

pykello:
@jasonmp85 all four comments make sense. (1) Fixed the major/patch release logic, (2) Broke HTTP/GET request function into 2 functions. Their signature is not symmetric. I thought we should just implement what we need now and refactor in future if we need to. (3)  URL encoding edition with `curl_easy_escape`. (4) Now using `findJsonbValueFromContainer` for extracting field values from json objects.

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -30,17 +30,26 @@ bool EnableStatisticsCollection = true; /* send basic usage statistics to Citus
...
+static bool JsonbFieldInt(Jsonb *jsonb, const char *fieldName, int32 *result);
+static bool JsonbFieldStr(Jsonb *jsonb, const char *fieldName, StringInfo result);
 static uint64 NextPow2(uint64 n);
 static uint64 ClusterSize(List *distributedTableList);
-static bool SendHttpPostJsonRequest(const char *url, const char *postFields, long
-									timeoutSeconds);
+static bool UrlEncode(StringInfo buf, const char *str);
+static bool SendHttpPostJsonRequest(const char *url, const char *postFields,
+									long timeoutSeconds);
+static bool SendHttpGetRequest(const char *url, long timeoutSeconds,
```
pykello: I didn't include the word `Json` in the function name since this function doesn't assume anything about response type. We could add `Accept: application/json`, but that didn't seem necessary.

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -114,8 +124,160 @@ CollectBasicUsageStatistics(void)
...
+	{
+		FlushErrorState();
+		return 0;
+	}
+	PG_END_TRY();
+
+	if (!JsonbFieldStr(responseJsonb, "version", releaseVersion) ||
+		!JsonbFieldInt(responseJsonb, "major", &releaseMajor) ||
+		!JsonbFieldInt(responseJsonb, "minor", &releaseMinor) ||
+		!JsonbFieldInt(responseJsonb, "patch", &releasePatch))
```
pykello: Since `findJsonbValueFromContainer()` can also be used to check field's existence, I thought returning `bool` for the existence of the field and setting the final parameter for result might make sense. Also, even if the field exists the value can be of incorrect type, so replacing these checks with `jsonb_exists` calls won't work. We could add functions like `JsonbIntFieldExists()`, but its body would be almost the same as a function getting the int field value, so I preferred what I did here.

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -114,8 +124,160 @@ CollectBasicUsageStatistics(void)
...
+ */
+static bool
+JsonbFieldInt(Jsonb *jsonb, const char *fieldName, int32 *result)
+{
+	JsonbValue *fieldValue = NULL;
+	JsonbValue key;
+	memset(&key, 0, sizeof(key));
+	key.type = jbvString;
+	key.val.string.len = strlen(fieldName);
+	key.val.string.val = pstrdup(fieldName);
```
pykello: I could probably cast `fieldName` to `char *` instead of `pstrdup()`, but that didn't look very good.

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -240,6 +411,71 @@ SendHttpPostJsonRequest(const char *url, const char *jsonObj, long timeoutSecond
...
+}
+
+
+/*
+ * PerformHttpRequest sends the HTTP request with the parameters set in the given
+ * curl object, and returns if it was successful or not. If the request was not
+ * successful, it may log some warnings.
+ */
+static bool
+PerformHttpRequest(CURL *curl)
```
pykello: Moved the error checking code to here to avoid much repetition between GET/POST functions.

src/backend/distributed/utils/statistics_collection.c
pykello: Right. It can also return `false` if `responseJsonb` isn't an object. It can be array, string, number, or null.

Yes, this is called synchronously and returning `0` here triggers the warning in `CheckForUpdates()`. It also triggers the `Sending HTTP request failed.` warning in `PerformHttpRequest()` with error code `"Failed writing received data to disk/application."` which means the callback function couldn't process the data.

src/backend/distributed/utils/statistics_collection.c

src/backend/distributed/utils/statistics_collection.c

src/backend/distributed/utils/maintenanced.c
```diff
@@ -326,6 +326,7 @@ CitusMaintenanceDaemonMain(Datum main_arg)
 			{
 				WarnIfSyncDNS();
 				statsCollectionSuccess = CollectBasicUsageStatistics();
+				CheckForUpdates();
```
jasonmp85: In the—admittedly weird—case where usage statistics fails but an update check succeeds, we would check for updates twice in a minute. Is that OK?

Maybe the `CheckForUpdates` call could be moved into the `statsCollectionSuccess` block (but with a suitable comment to explain why)?

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -138,6 +147,173 @@ CollectBasicUsageStatistics(void)
...
+	char *updateType = NULL;
+	MemoryContext savedContext = CurrentMemoryContext;
+
+	StringInfo responseNullTerminated = makeStringInfo();
+	appendBinaryStringInfo(responseNullTerminated, contents, size * count);
+
+	/* jsonb_in can throw errors */
+	PG_TRY();
+	{
+		responseJsonb = DatumGetJsonb(
```
jasonmp85: Again, I think this is too many operations for a line; if the line breaks like this, better to just break it up.

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -138,6 +147,173 @@ CollectBasicUsageStatistics(void)
...
+	else if (releaseMajor == citusVersionMajor &&
+			 releaseMinor == citusVersionMinor &&
+			 releasePatch > citusVersionPatch)
+	{
+		updateType = "patch";
+	}
+
+	if (updateType != NULL)
+	{
+		ereport(LOG, (errmsg("a newer %s release of Citus (%s) is available",
```
jasonmp85: Change _newer_ to _new_.

src/backend/distributed/utils/statistics_collection.c

src/backend/distributed/utils/statistics_collection.c

src/backend/distributed/utils/maintenanced.c

sumedhpathak:
@pykello Are there further open questions to merge given Jason's ship-it?

pykello:
@sumedhpathak this is based on #1726 , so we need to get that merged before I can merge this. For #1726, I'm waiting for @jasonmp85 's approval.

pykello:
Merged via #1744.

PR #1696, opened 2017-10-12
Rewrite recover_prepared_transactions to be faster, non-blocking

### LINKED ISSUE #1669: Make recover_prepared_transaction faster / not block commits
Running pgbench against #1574 shows some significant blips when `recover_prepared_transactions()` runs. 

```
$ pgbench -j 64 -c 64 -n -P 10 -T 3600 -r $PGURL
progress: 20.0 s, 2870.5 tps, lat 22.288 ms stddev 8.192
progress: 30.0 s, 2889.0 tps, lat 22.163 ms stddev 8.726
progress: 40.0 s, 2868.0 tps, lat 22.305 ms stddev 7.475
progress: 50.0 s, 2895.0 tps, lat 22.101 ms stddev 6.838
progress: 60.0 s, 2912.8 tps, lat 21.960 ms stddev 9.504
progress: 70.0 s, 1751.7 tps, lat 36.563 ms stddev 241.539
progress: 80.0 s, 2886.7 tps, lat 22.183 ms stddev 7.410
progress: 90.0 s, 2883.1 tps, lat 22.185 ms stddev 7.824
progress: 100.0 s, 2884.7 tps, lat 22.177 ms stddev 9.311
progress: 110.0 s, 2867.7 tps, lat 22.337 ms stddev 10.452
progress: 120.0 s, 2926.6 tps, lat 21.860 ms stddev 7.159
progress: 130.0 s, 1831.2 tps, lat 34.947 ms stddev 220.318
progress: 140.0 s, 2857.4 tps, lat 22.383 ms stddev 8.301
progress: 150.0 s, 2941.1 tps, lat 21.756 ms stddev 8.078
progress: 160.0 s, 2903.3 tps, lat 22.044 ms stddev 7.797
progress: 170.0 s, 2874.1 tps, lat 22.269 ms stddev 7.265
progress: 180.0 s, 2805.4 tps, lat 22.827 ms stddev 10.537
progress: 190.0 s, 1790.3 tps, lat 35.705 ms stddev 222.923
progress: 200.0 s, 2878.6 tps, lat 22.237 ms stddev 9.899
progress: 210.0 s, 2889.1 tps, lat 22.166 ms stddev 10.590
progress: 220.0 s, 2864.7 tps, lat 22.333 ms stddev 9.400
progress: 230.0 s, 2883.7 tps, lat 22.208 ms stddev 7.606
progress: 240.0 s, 2872.9 tps, lat 22.255 ms stddev 6.753
progress: 250.0 s, 1798.9 tps, lat 35.576 ms stddev 220.595
progress: 260.0 s, 2877.3 tps, lat 22.249 ms stddev 7.880
progress: 270.0 s, 2925.8 tps, lat 21.868 ms stddev 7.821
progress: 280.0 s, 2884.4 tps, lat 22.199 ms stddev 7.780
progress: 290.0 s, 2920.9 tps, lat 21.916 ms stddev 8.067
progress: 300.0 s, 2905.3 tps, lat 22.017 ms stddev 7.134
progress: 310.0 s, 1794.1 tps, lat 35.695 ms stddev 229.221
progress: 320.0 s, 2895.2 tps, lat 22.102 ms stddev 7.981
progress: 330.0 s, 2936.0 tps, lat 21.779 ms stddev 8.037
progress: 340.0 s, 2913.4 tps, lat 21.984 ms stddev 9.802
```

One way to substantially speed up `recover_prepared_transactions` is to avoid doing an index look-up for every record, but rather to delete all records for a worker node in a single pass.

Another improvement would be to get the prepared transactions list from all workers in parallel.

We should also be able to avoid blocking 2PCs by only considering prepared transactions of transactions that are not in progress.

marcocitus:
This PR is a rewrite of `RecoverWorkerTransactions` to be simpler, significantly faster, and non-blocking.

`RecoverWorkerTransactions` first builds a hash table of pending prepared transactions and a hash table of in-progress distributed transactions and then performs a single scan of `pg_dist_transaction`. For each record:
- if it is part of an active distributed transaction, skip it, which allows us to take a less aggressive lock
- if there is a pending prepared transactions on the worker, commit it
- deleting the recovery record.

Finally, `RecoverWorkerTransactions` aborts all remaining prepared transactions which are not part of an active distributed transaction.

Fixes #1669

### DIFF
--- a/src/backend/distributed/transaction/backend_data.c
+++ b/src/backend/distributed/transaction/backend_data.c
@@ -640,3 +640,45 @@ MyBackendGotCancelledDueToDeadlock(void)
 
 	return cancelledDueToDeadlock;
 }
+
+
+/*
+ * ActiveDistributedTransactionNumbers returns a list of pointers to
+ * transaction numbers of distributed transactions that are in progress.
+ */
+List *
+ActiveDistributedTransactionNumbers(void)
+{
+	List *activeTransactionNumberList = NIL;
+	int curBackend = 0;
+
+	/* build list of starting procs */
+	for (curBackend = 0; curBackend < MaxBackends; curBackend++)
+	{
+		PGPROC *currentProc = &ProcGlobal->allProcs[curBackend];
+		BackendData currentBackendData;
+		uint64 *transactionNumber = NULL;
+
+		if (currentProc->pid == 0)
+		{
+			/* unused PGPROC slot */
+			continue;
+		}
+
+		GetBackendDataForProc(currentProc, &currentBackendData);
+
+		if (!IsInDistributedTransaction(&currentBackendData))
+		{
+			/* not a distributed transaction */
+			continue;
+		}
+
+		transactionNumber = (uint64 *) palloc0(sizeof(uint64));
+		*transactionNumber = currentBackendData.transactionId.transactionNumber;
+
+		activeTransactionNumberList = lappend(activeTransactionNumberList,
+											  transactionNumber);
+	}
+
+	return activeTransactionNumberList;
+}
--- a/src/backend/distributed/transaction/backend_data.c
+++ b/src/backend/distributed/transaction/backend_data.c
@@ -644,7 +644,8 @@ MyBackendGotCancelledDueToDeadlock(void)
 
 /*
  * ActiveDistributedTransactionNumbers returns a list of pointers to
- * transaction numbers of distributed transactions that are in progress.
+ * transaction numbers of distributed transactions that are in progress
+ * and were started by the node on which it is called.
  */
 List *
 ActiveDistributedTransactionNumbers(void)
@@ -670,6 +671,12 @@ ActiveDistributedTransactionNumbers(void)
 		if (!IsInDistributedTransaction(&currentBackendData))
 		{
 			/* not a distributed transaction */
+			continue;
+		}
+
+		if (!currentBackendData.transactionId.transactionOriginator)
+		{
+			/* not a coordinator process */
 			continue;
 		}
 
--- a/src/backend/distributed/transaction/remote_transaction.c
+++ b/src/backend/distributed/transaction/remote_transaction.c
@@ -27,6 +27,9 @@
 #include "utils/hsearch.h"
 
 
+#define PREPARED_TRANSACTION_NAME_FORMAT "citus_%u_%u_"UINT64_FORMAT "_%u"
+
+
 static void StartRemoteTransactionSavepointBegin(MultiConnection *connection,
 												 SubTransactionId subId);
 static void FinishRemoteTransactionSavepointBegin(MultiConnection *connection,
@@ -1241,8 +1244,33 @@ Assign2PCIdentifier(MultiConnection *connection)
 
 	/* print all numbers as unsigned to guarantee no minus symbols appear in the name */
 	snprintf(connection->remoteTransaction.preparedName, NAMEDATALEN,
-			 "citus_%u_%u_"UINT64_FORMAT "_%u", GetLocalGroupId(), MyProcPid,
+			 PREPARED_TRANSACTION_NAME_FORMAT, GetLocalGroupId(), MyProcPid,
 			 transactionNumber, connectionNumber++);
+}
+
+
+/*
+ * ParsePreparedTransactionName parses a prepared transaction name to extract
+ * the initiator group ID, initiator process ID, distributed transaction number,
+ * and the connection number. If the transaction name does not match the expected
+ * format ParsePreparedTransactionName returns false, and true otherwise.
+ */
+bool
+ParsePreparedTransactionName(char *preparedTransactionName, int *groupId, int *procId,
+							 uint64 *transactionNumber, uint32 *connectionNumber)
+{
+	const int expectedFieldCount = 4;
+	int parsedFieldCount = 0;
+	bool nameValid = false;
+
+	parsedFieldCount = sscanf(preparedTransactionName, PREPARED_TRANSACTION_NAME_FORMAT,
+							  groupId, procId, transactionNumber, connectionNumber);
+	if (parsedFieldCount == expectedFieldCount)
+	{
+		nameValid = true;
+	}
+
+	return nameValid;
 }
 
 
--- a/src/backend/distributed/transaction/transaction_recovery.c
+++ b/src/backend/distributed/transaction/transaction_recovery.c
@@ -150,6 +150,8 @@ RecoverWorkerTransactions(WorkerNode *workerNode)
 
 	List *pendingTransactionList = NIL;
 	HTAB *pendingTransactionSet = NULL;
+	List *recheckTransactionList = NIL;
+	HTAB *recheckTransactionSet = NULL;
 
 	Relation pgDistTransaction = NULL;
 	SysScanDesc scanDescriptor = NULL;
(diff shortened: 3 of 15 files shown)


src/backend/distributed/transaction/transaction_recovery.c
```diff
@@ -182,191 +182,130 @@ RecoverWorkerTransactions(WorkerNode *workerNode)
 										 ALLOCSET_DEFAULT_MAXSIZE);
 	oldContext = MemoryContextSwitchTo(localContext);
 
-	/* find transactions that were committed, but not yet confirmed */
-	unconfirmedTransactionList = UnconfirmedWorkerTransactionsList(groupId);
-	unconfirmedTransactionList = SortList(unconfirmedTransactionList, CompareNames);
+	/* take table lock first to avoid running concurrently */
```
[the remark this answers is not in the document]
marcocitus: > Otherwise, it seems possible that we might commit an ongoing distributed transaction that has started and prepared after ActiveDistributedTransactionNumbers() and before PendingWorkerTransactionList() called?

Good point. This requires some more care.

Consider the following order:

1) P = PendingWorkerTransactionList()
2) A = ActiveDistributedTransactionNumbers() 
3) T = systable_beginscan(pgDistTransaction) - get snapshot of pg_dist_transaction

At time 2, we get a conclusive answer to which distributed transactions in P are still in progress. We therefore will never roll forward prepared transactions of in progress transactions.

However, a new transaction could have started, left prepared transactions, and committed between 1 and 2 (or 1 and 3). We wouldn't roll these prepared transactions forward since they are not in P, but we would delete their records, meaning that we would incorrectly roll them back the next time recover_prepared_transactions is called.

A solution is to get the list of prepared transactions again after obtaining the pg_dist_transaction snapshot:

4) Q = PendingWorkerTransactionList()

We then skip deletion for records that appear in Q and we resolve prepared transactions in P.

Does this make sense? Basically, anything that's in P and Q, but not in A, is known to have prepared before 1 and aborted/committed before 2, and is therefore safe to resolve after 2.

src/backend/distributed/transaction/backend_data.c
```diff
@@ -640,3 +640,45 @@ MyBackendGotCancelledDueToDeadlock(void)
...
+
+		if (currentProc->pid == 0)
+		{
+			/* unused PGPROC slot */
+			continue;
+		}
+
+		GetBackendDataForProc(currentProc, &currentBackendData);
+
+		if (!IsInDistributedTransaction(&currentBackendData))
```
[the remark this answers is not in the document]
marcocitus: Yep, makes sense.

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -1241,12 +1244,37 @@ Assign2PCIdentifier(MultiConnection *connection)
...
 
 
 /*
+ * ParsePreparedTransactionName parses a prepared transaction name to extract
+ * the initiator group ID, initiator process ID, distributed transaction number,
+ * and the connection number. If the transaction name does not match the expected
+ * format ParsePreparedTransactionName returns false, and true otherwise.
+ */
+bool
+ParsePreparedTransactionName(char *preparedTransactionName, int *groupId, int *procId,
```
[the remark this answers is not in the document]
marcocitus: I guess to avoid ever putting a dash in the name, which would require escaping.

src/backend/distributed/utils/listutils.c
```diff
@@ -102,3 +102,49 @@ DatumArrayToArrayType(Datum *datumArray, int datumCount, Oid datumTypeId)
...
+	ListCell *itemCell = NULL;
+	int flags = HASH_ELEM;
+
+	/* allocate sufficient capacity for O(1) expected look-up time */
+	int capacity = (int) (list_length(itemList) / 0.75) + 1;
+
+	/* initialise the hash table for looking up keySize bytes */
+	memset(&info, 0, sizeof(info));
+	info.keysize = keySize;
+	info.entrysize = keySize;
```
onderkalaci: Minor: Since we're not using the entries here, can't we use `0` as the size or (4 bytes to set the entry to integer `0` or `NULL`)?

src/backend/distributed/utils/listutils.c
```diff
@@ -102,3 +102,49 @@ DatumArrayToArrayType(Datum *datumArray, int datumCount, Oid datumTypeId)
...
+
+	if (!isStringList)
+	{
+		info.hash = tag_hash;
```
onderkalaci: In the function declartion comment of `tag_hash`:
```
/*
 * prototypes for functions in hashfn.c
 *
 * Note: It is deprecated for callers of hash_create to explicitly specify
 * string_hash, tag_hash, uint32_hash, or oid_hash.  Just set HASH_BLOBS or
 * not.  Use HASH_FUNCTION only when you want something other than those.
 */
```

So, maybe follow that?

src/test/regress/specs/isolation_transaction_recovery.spec
```diff
@@ -33,5 +33,5 @@ step "s2-insert"
     INSERT INTO test_transaction_recovery VALUES (1,2);
 }
 
-# Recovery and 2PCs should block each other
+# Recovery and 2PCs should not block each other
```
onderkalaci: Shall we add one more permutation for `"s1-recover"` and `"s2-recover"`?

src/backend/distributed/transaction/transaction_recovery.c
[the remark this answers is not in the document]
onderkalaci: Actually, I observed such cases on my tests:

```SQL
WARNING:  prepared transaction with identifier "citus_0_21505_69107_948" does not exist

```

src/backend/distributed/utils/listutils.c
marcocitus: In a postgres hash table, the entry always starts with the key.

src/backend/distributed/transaction/transaction_recovery.c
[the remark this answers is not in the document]
marcocitus: Good catch. Changed to always skipping prepared transactions in Q.

> P = prepared transactions on worker
> (One of the transactions commits in between)
> A = active distributed transactions (doesn't include that transaction)

Either that transaction leaves prepared transactions, which will show up in Q and we skip, or it commits them and it won't show up in Q, in which case deleting the record is fine.

src/backend/distributed/transaction/transaction_recovery.c
```diff
@@ -182,191 +184,176 @@ RecoverWorkerTransactions(WorkerNode *workerNode)
...
-	const char *leftString = *((char **) leftPointer);
-	const char *rightString = *((char **) rightPointer);
-
-	int nameCompare = strncmp(leftString, rightString, NAMEDATALEN);
-
-	return nameCompare;
-}
 
+		/* we've successfully committed the prepared transaction, remove the record */
+		simple_heap_delete(pgDistTransaction, &heapTuple->t_self);
```
onderkalaci: Nothing actionable, just noting for future reference: The perf output follows:

```
      - PostgresMain                                                                                                                                                                                                                  ▒
         - 77.59% exec_simple_query                                                                                                                                                                                                   ▒
              PortalRun                                                                                                                                                                                                               ▒
              PortalRunSelect                                                                                                                                                                                                         ▒
              ExecutorRun                                                                                                                                                                                                             ▒
              standard_ExecutorRun                                                                                                                                                                                                    ▒
              ExecutePlan                                                                                                                                                                                                             ▒
              ExecProcNodeFirst                                                                                                                                                                                                       ▒
              ExecResult                                                                                                                                                                                                              ▒
              ExecInterpExpr                                                                                                                                                                                                          ▒
              recover_prepared_transactions                                                                                                                                                                                           ▒
            - RecoverPreparedTransactions                                                                                                                                                                                             ▒
               - 77.12% RecoverWorkerTransactions                                                                                                                                                                                     ▒
                  + 31.20% simple_heap_delete                                                                                                                                                                                         ▒
                  + 29.06% systable_getnext                                                                                                                                                                                           ▒
                  + 8.08% hash_search                                                                                                                                                                                                 ▒
                  + 3.38% text_to_cstring                                                                                                                                                                                             ▒
                  + 2.64% IsTransactionInProgress                                                                                                                                                                                     ▒
                    0.92% __memcpy_ssse3                                                                                                                                                                                              ▒
                    0.56% MemoryContextDelete 
```

So, we're mostly bottlenecked on iterating and deleting from `pg_dist_transaction` which seems reasonable to me.

The largest test I've done was with 35M rows in `pg_dist_transaction`, which took ` 1064 secs` without blocking the new 2PCs on `1+2 c4.xlarge` cluster. The memory usage increased to `1GB`, but, seems OK to me.

With a smaller number of entries, the performance is obviously much better. (512000 entries in `46 secs`)

src/backend/distributed/transaction/transaction_recovery.c
```diff
@@ -182,191 +184,220 @@ RecoverWorkerTransactions(WorkerNode *workerNode)
...
+		/*
+		 * Remove the transaction from the pending list such that only transactions
+		 * that need to be aborted remain at the end.
+		 */
+		hash_search(pendingTransactionSet, transactionName, HASH_REMOVE,
+					&isTransactionPending);
 
-	return differenceList;
-}
+		if (isTransactionPending)
```
onderkalaci: Hmm, I still observe `WARNING:  prepared transaction with identifier "citus_0_31126_358610_5477" does not exist` type of warning when pgbench continue to insert into a reference table and recovery runs in parallel for every second. 

Could it be possible that some transactions appear in `pendingTransactionList ` then commits and later `activeTransactionNumberList ` doesn't include that transaction which recovery would try to commit?

I think we should lock backends with `locklockdata()` at least getting for `pendingTransactionList ` and `activeTransactionNumberList `.   Looking into the perf output, that should be a very small amount of time. What do you think?

marcocitus: > Could it be possible that some transactions appear in pendingTransactionList then commits and later activeTransactionNumberList doesn't include that transaction which recovery would try to commit?

The warnings occur during the recovery, not in the commit callback, right? I guess we could just ignore those.

> I think we should lock backends with locklockdata() at least getting for pendingTransactionList and activeTransactionNumberList. Looking into the perf output, that should be a very small amount of time. What do you think?

I'm a bit hesitant to do that, because we'd block both `assign_distributed_transaction_id` and COMMIT for the duration of a network round-trip (with possible failures, delays), instead of just blocking 2PC commit.

marcocitus: Alternatively we could skip committing prepared transactions that are not in Q.

onderkalaci: Ok, the final changes make sense to me as well. Also, the tests run smoothly now

src/backend/distributed/transaction/transaction_recovery.c
```diff
@@ -182,191 +184,221 @@ RecoverWorkerTransactions(WorkerNode *workerNode)
...
 
-		if (!nameFound)
+			continue;
+		}
+		else
 		{
 			/*
-			 * baseName is not in subtractArray and thus included
-			 * in the difference.
+			 * We found a recovery record without any prepared transaction. It
```
onderkalaci: Is that ever possible `(!foundPreparedTransactionAfterCommit && !foundPreparedTransactionBeforeCommit)`? I think not, but just wanted to ask.

marcocitus: Yes, that's actually the common case. Any 2PC that successfully completed before the call to `recover_prepared_transactions()` will have a `pg_dist_transaction` record, but no prepared transaction, since it was already committed. In that case we just delete the record.

marcocitus:
> If the transaction rate is too high, we'd probably never have a chance for auto vacuum to run? What would be our plan in that case? I think that's a separate issue from this. We should probably discuss that later.

I think it's mainly an autovacuum tuning problem. Are you using default settings? On Citus Cloud we've made autovacuum more aggressive by setting `autovacuum_vacuum_cost_delay = 0`, which means autovacuum works without pausing.

onderkalaci:
> I think it's mainly an autovacuum tuning problem. Are you using default settings? On Citus Cloud we've made autovacuum more aggressive by setting autovacuum_vacuum_cost_delay = 0, which means autovacuum works without pausing.

No, I haven't changed any config. I'll try with ` autovacuum_vacuum_cost_delay = 0`

onderkalaci:
> I think it's mainly an autovacuum tuning problem. Are you using default settings? On Citus Cloud we've made autovacuum more aggressive by setting autovacuum_vacuum_cost_delay = 0, which means autovacuum works without pausing.

So, do you think is that better idea to test the 2PC recovery with ` autovacuum_vacuum_cost_delay = 0`? In general, shall we suggest the production users this?

marcocitus:
> So, do you think is that better idea to test the 2PC recovery with autovacuum_vacuum_cost_delay = 0? In general, shall we suggest the production users this?

In general, the default setting is only suitable for spiky workloads. For benchmarks or high write volumes you should use `0`.

Issue #1697, opened 2017-10-12
intermittent crashes possibly related to stats commit


mtuncer:
Observing postmaster crashes after running few queries.
Easiest repro is to run ```\dx``` . Postmaster crashes after few runs.

It looks like it started happening after #1656 . 
setting citus.enable_statistics_collection to false seem to fix it.

onderkalaci:
Verified that related to statistics collection:

backtrace:
```
#0  0x00007fff9564c8c8 in curl_slist_free_all ()
(gdb) bt
#0  0x00007fff9564c8c8 in curl_slist_free_all ()
#1  0x000000010e0ce686 in SendHttpPostJsonRequest (url=0x10e0eddf9 "https://citus-statistics.herokuapp.com/v1/usage_reports", jsonObj=0x7f9a8b893a70 "{\"citus_version\": \"7.1devel\",\"table_count\": 4,\"cluster_size\": 2097152,\"worker_node_count\": 3,\"os_name\": \"Darwin\",\"os_release\": \"15.6.0\",\"hwid\": \"x86_64\"}", timeoutSeconds=5) at utils/statistics_collection.c:203
#2  0x000000010e0ce187 in CollectBasicUsageStatistics () at utils/statistics_collection.c:88
#3  0x000000010e0afd1c in CitusMaintenanceDaemonMain ()
#4  0x000000010d749b64 in StartBackgroundWorker () at bgworker.c:725
#5  0x000000010d75ff77 in do_start_bgworker (rw=0x7f9a89e009d0) at postmaster.c:5537
#6  0x000000010d75ae7b in maybe_start_bgworker () at postmaster.c:5712
#7  0x000000010d7595be in sigusr1_handler (postgres_signal_arg=30) at postmaster.c:4959
#8  <signal handler called>
#9  0x00007fff8d50607a in select$DARWIN_EXTSN ()
#10 0x000000010d75affc in ServerLoop () at postmaster.c:1665
#11 0x000000010d758966 in PostmasterMain (argc=3, argv=0x7f9a89c036d0) at postmaster.c:1309
#12 0x000000010d68959f in main (argc=3, argv=0x7f9a89c036d0) at main.c:228

```

PR #1698, opened 2017-10-12
Id generation for stats collection.

pykello:
- Adds ```pg_dist_metadata``` which holds tagged metadata values. 
- Adds a random value for the tag ```server_id``` when creating the extension.
- Looks up and sends ```server_id``` when collecting statistics.


mtuncer:
does it fix #1697 ?

pykello:
@mtuncer I don't know. I'll look into fixing #1697. Maybe #1695 fixes that. Initial stats collection patch had an out of bound memory access, which #1695 fixes. I'll create a separate PR for that to land it soon.

pykello:
@jasonmp85 
1. As we discussed, now I add all metadata tag/values to the json payload.
2. Bumped the version to `7.1-3`.

src/backend/distributed/citus--7.1-2--7.1-3.sql
```diff
@@ -0,0 +1,10 @@
+/* citus--7.1-2--7.1-3 */
+
+CREATE TABLE citus.pg_dist_metadata(
+    tag text NOT NULL,
```
marcocitus: Would it make sense for tag to be a primary key?

jasonmp85: :+1:

src/backend/distributed/citus--7.1-2--7.1-3.sql
```diff
@@ -0,0 +1,10 @@
...
+    value text NOT NULL
+);
+ALTER TABLE citus.pg_dist_metadata SET SCHEMA pg_catalog;
+GRANT SELECT ON pg_catalog.pg_dist_metadata TO public;
```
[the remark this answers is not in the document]
pykello: By default, user who created the extension owns the table and has all the permissions. So I think we should be OK in docker.

We probably don't need this `GRANT` for `pg_dist_metadata` since we run it in a background worker. I just added it since other `pg_dist...` tables had this `GRANT` so all users can read these tables.

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -95,8 +101,11 @@ CollectBasicUsageStatistics(void)
 	escape_json(fields, unameData.release);
 	appendStringInfoString(fields, ",\"hwid\": ");
 	escape_json(fields, unameData.machine);
+	appendStringInfo(fields, ",\"node_metadata\": %s", metadataJsonbStr);
```
pykello: I couldn't find a better API to construct json objects in C code. `json_build_object` seemed to be an option, but calling it inside C code is not very simple.

src/backend/distributed/citus--7.1-2--7.1-3.sql
```diff
@@ -0,0 +1,14 @@
+/* citus--7.1-2--7.1-3 */
+
+CREATE TABLE citus.pg_dist_node_metadata(
+    metadata jsonb
```
pykello: Using `jsonb` instead of `json` since it provides better API. Although it makes no difference in this patch.

jasonmp85: Could use a `NOT NULL` here…

src/backend/distributed/utils/statistics_collection.c
jasonmp85: Yeah that's fine.

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -3167,3 +3168,62 @@ CitusInvalidateRelcacheByShardId(int64 shardId)
...
+		Assert(!tagIsNull && !valueIsNull);
+
+		tagStr = makeStringInfo();
+		appendStringInfoString(tagStr, TextDatumGetCString(tagDatum));
+		valueStr = makeStringInfo();
+		appendStringInfoString(valueStr, TextDatumGetCString(valueDatum));
+		metadataList = lappend(metadataList, list_make2(tagStr, valueStr));
+
+		heapTuple = systable_getnext(scanDescriptor);
+	}
```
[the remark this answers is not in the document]
jasonmp85: Hm… it probably _should_ I just don't understand why it's not already. I can look at it (it's controlled by the `uncrustify` configuration in the `citusdata/tools` repo).

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -3167,3 +3168,62 @@ CitusInvalidateRelcacheByShardId(int64 shardId)
...
 	CommandCounterIncrement();
 }
+
+
+/*
+ * DistMetadataList returns list of metadata tag/value pairs stored at pg_dist_metadata.
+ * Each pair is a list with two StringInfos.
+ */
+List *
+DistMetadataList(void)
```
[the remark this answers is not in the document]
jasonmp85: Good point, I was prematurely optimizing.

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -3167,3 +3168,62 @@ CitusInvalidateRelcacheByShardId(int64 shardId)
 	/* bump command counter, to force invalidation to take effect */
 	CommandCounterIncrement();
 }
+
+
+/*
+ * DistMetadataList returns list of metadata tag/value pairs stored at pg_dist_metadata.
+ * Each pair is a list with two StringInfos.
```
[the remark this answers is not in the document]
jasonmp85: Sounds good.

src/backend/distributed/citus--7.1-2--7.1-3.sql
```diff
@@ -0,0 +1,10 @@
+/* citus--7.1-2--7.1-3 */
+
+CREATE TABLE citus.pg_dist_metadata(
+    tag text NOT NULL,
+    value text NOT NULL
+);
+ALTER TABLE citus.pg_dist_metadata SET SCHEMA pg_catalog;
+GRANT SELECT ON pg_catalog.pg_dist_metadata TO public;
+
+INSERT INTO pg_dist_metadata VALUES ('server_id', (random() * 1e18)::bigint::text);
```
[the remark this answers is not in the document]
jasonmp85: Thanks for the research on the seeding process. I'm comfortable with everything you found.

I guess I wasn't entirely clear about using `random` as it is used in the `uint` function… I didn't mean using it in SQL-land, since getting at the byte representation of e.g. a `float` in that environment is tough. I kind of meant just writing our own C function to precisely do what we want and using it.

After thinking about that some more, I realized SQL doesn't even _have_ a `uint` type. I looked at the code to generate type 4 (i.e. pure-random) UUIDs and found it was very simple. Can we just do something like this (a modified `pg_random_uuid` from `pgcrypto`)?

```c
PG_FUNCTION_INFO_V1(citus_server_id);

Datum
citus_server_id(PG_FUNCTION_ARGS)
{
	const Size uuidLength = 16;
	uint8 *randomBytes = (uint8 *) palloc(uuidLength);

	if (!pg_backend_random((char *) randomBytes, uuidLength))
	{
		ereport(ERROR, (errmsg("failed to generate server identifier")))
	}

	/*
	 * Set magic numbers for a "version 4" (pseudorandom) UUID, see
	 * http://tools.ietf.org/html/rfc4122#section-4.4
	 */
	buf[6] = (buf[6] & 0x0f) | 0x40;	/* "version" field */
	buf[8] = (buf[8] & 0x3f) | 0x80;	/* "variant" field */

	PG_RETURN_UUID_P((pg_uuid_t *) buf);
}
```

Then we know what we're getting is a good, random UUID and we can just cast it to text when inserting it to the `jsonb` value.

jasonmp85:
@pykello I put my final requests on here. Basically I wondered how hard it would be to generate a fully-random UUID using our own implementation… it's pretty easy (PostgreSQL comes with the UUID type built-in, just no functions to generate them). The UUID will make a nicer textual format anyways and this code is closer to what PostgreSQL does when it needs random integer data.

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -233,3 +237,26 @@ SendHttpPostJsonRequest(const char *url, const char *jsonObj, long timeoutSecond
 
 
 #endif /* HAVE_LIBCURL */
+
+PG_FUNCTION_INFO_V1(citus_server_id);
+
+/* citus_server_id returns a random UUID value as server identifier. */
```
jasonmp85: Can you add _modeled after PostgreSQL's pg\_random\_uuid_ to this comment?

pykello:
@jasonmp85 I just added the `NOT NULL` constraint and used the `citus_server_id()` you suggested (I didn't know `pg_backend_random()` exists. Thanks). Please take a look.

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -249,10 +250,17 @@ citus_server_id(PG_FUNCTION_ARGS)
...
 
+#if PG_VERSION_NUM >= 100000
 	if (!pg_backend_random((char *) buf, UUID_LEN))
 	{
 		ereport(ERROR, (errmsg("failed to generate server identifier")));
 	}
+#else
+	for (int bufIdx = 0; bufIdx < UUID_LEN; bufIdx++)
+	{
+		buf[bufIdx] = random() & 0xff;
```
jasonmp85: Can you add an explicit cast to `char` here: `(char) (random() & 0xFF)`?

The non-strong-random implementation of `pg_backend_random` looks like so:

```c
	for (i = 0; dst < end; i++)
	{
		uint32		r;
		int			j;

		/*
		 * pg_jrand48 returns a 32-bit integer. Fill the next 4 bytes from it.
		 */
		r = (uint32) pg_jrand48(BackendRandomShmem->seed);

		for (j = 0; j < 4 && dst < end; j++)
		{
			*(dst++) = (char) (r & 0xFF);
			r >>= 8;
		}
	}
```

So for consistency, I think we should have the cast and `FF` be uppercase, but otherwise this is good.

src/backend/distributed/citus--7.1-2--7.1-3.sql
```diff
@@ -0,0 +1,17 @@
+/* citus--7.1-2--7.1-3 */
+
+CREATE TABLE citus.pg_dist_node_metadata(
+    metadata jsonb NOT NULL
+);
+ALTER TABLE citus.pg_dist_node_metadata SET SCHEMA pg_catalog;
+GRANT SELECT ON pg_catalog.pg_dist_node_metadata TO public;
+
+CREATE FUNCTION pg_catalog.citus_server_id()
```
marcocitus: Would be kind of nice to rename this to something like citus_uuid() since it a) seems generally useful for users that don't have uuid-ossp installed, b) doesn't do a good job of returning the current server ID.

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -95,8 +109,11 @@ CollectBasicUsageStatistics(void)
 	escape_json(fields, unameData.release);
 	appendStringInfoString(fields, ",\"hwid\": ");
 	escape_json(fields, unameData.machine);
+	appendStringInfo(fields, ",\"node_metadata\": %s", metadataJsonbStr);
 	appendStringInfoString(fields, "}");
 
+	CommitTransactionCommand();
```
marcocitus: This resets the transaction memory context, so below seems like it might be a use-after-free bug. Why not follow the pattern established by deadlock detection here?

src/backend/distributed/citus--7.1-2--7.1-3.sql
jasonmp85: Hm, I kind of suggested this name to denote it as something internal-only (and it doesn't say _current_ server id, it just generates one :-P). But I see your point and could be swayed.

jasonmp85:
BTW I opened citusdata/docker#50, which uses our flexible JSON metadata value to add a `"docker": true` attribute when Docker starts up (figured this would be easier than trying to figure out what is reporting from Docker post-facto).

We might also want to tag cloud, cloudformation, and other places we have control over this stuff (running tests, just in case?).

PR #1699, opened 2017-10-12
Fix out of bound memeory access when getting HTTP response code.

### LINKED ISSUE #1697: intermittent crashes possibly related to stats commit

pykello:
Should fix #1697.


pykello:
Hey @mtuncer , @onderkalaci . I couldn't reproduce the exact crash described at #1697, but I had seen a similar crash around that code that this PR fixes. Can you please verify this patch fixes the crashes you were seeing?

mtuncer:
verified that it fixes #1697

PR #1701, opened 2017-10-12
Couple of bug-fixes for statistics collection.

### LINKED ISSUE #1693: Make with --without-libcurl fails if you don't have curl
```
utils/statistics_collection.c:13:23: fatal error: curl/curl.h: No such file or directory
```

If you don't have the curl library installed it should be possible to build, by adding the `--without-libcurl` flag.

pykello:
1.  Fixes compiling `--without-libcurl`. Previously `<curl/curl.h>` was included even if compiled `--without-libcurl`. This can fail when libcurl headers are not there. This commit guards this include by checks for `HAVE_LIBCURL`. Fixes #1693.
2. Deletes `StatsCollection` memory context after we are done with statistics reporting. Previously we left the memory context untouched, which leaked memory over time.

### DIFF
--- a/configure
+++ b/configure
@@ -3349,7 +3349,9 @@ if test "${with_libcurl+set}" = set; then :
   withval=$with_libcurl;
   case $withval in
     yes)
-      :
+
+$as_echo "#define HAVE_LIBCURL 1" >>confdefs.h
+
       ;;
     no)
       :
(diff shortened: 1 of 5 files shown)


jasonmp85:
Hey @pykello, I added one commit to this PR: 7e639050cbbf62bc1ca92ba4e022214ff80f386c

Eclipse kept shading out all the libcurl-guarded areas in my editor even though I haven't seen this happen with PostgreSQL… I realized PostgreSQL never uses `-D` and that Eclipse must not scan output for that during builds, so I switched to using a header `#define` and `#ifdef` to guard the code, which is what they seem to do more often than not.

Issue #1694, opened 2017-10-11
citus_total_relation_size error message is too verbose


lithp:
It emits `ERROR:  could not open relation with OID 23227`. This has 2 problems:
- Nobody cares about OIDs, it's not user-friendly
- It makes our regression tests brittle

The table size functions should instead accept a table name (this changes the signature of those functions but not their usage, so I think it's a safe change) and then emit an error message if that table disappears before we call `relation_open`.

jasonmp85:
> The table size functions should instead accept a table name

I'm of the strong opinion that we should use correct types when possible. `text` does not describe what this function expects.

marcocitus:
+1 on keeping regclass.

jasonmp85:
@lithp hasn't this been entirely addressed now? Can we close this?

PR #1703, opened 2017-10-13
DistributedTableSize should not emit oid when erring out

lithp:
Addresses complaint in #1694 and will fix the test errors in #1691 once merged.

### DIFF
--- a/src/backend/distributed/master/master_metadata_utility.c
+++ b/src/backend/distributed/master/master_metadata_utility.c
@@ -169,8 +169,13 @@ DistributedTableSize(Oid relationId, char *sizeQuery)
 							   " blocks which contain multi-shard data modifications")));
 	}
 
-	/* try to open relation, will error out if the relation does not exist */
-	relation = relation_open(relationId, AccessShareLock);
+	relation = try_relation_open(relationId, AccessShareLock);
+
+	if (relation == NULL)
+	{
+		ereport(ERROR,
+				(errmsg("could not compute table size: relation does not exist")));
+	}
 
 	ErrorIfNotSuitableToGetSize(relationId);
 
--- a/src/test/regress/expected/isolation_drop_vs_all.out
+++ b/src/test/regress/expected/isolation_drop_vs_all.out
@@ -174,7 +174,7 @@ step s1-drop: DROP TABLE drop_hash;
 step s2-table-size: SELECT citus_total_relation_size('drop_hash'); <waiting ...>
 step s1-commit: COMMIT;
 step s2-table-size: <... completed>
-error in steps s1-commit s2-table-size: ERROR:  could not open relation with OID 23227
+error in steps s1-commit s2-table-size: ERROR:  could not compute table size: relation does not exist
 step s2-commit: COMMIT;
 step s1-select-count: SELECT COUNT(*) FROM drop_hash;
 ERROR:  relation "drop_hash" does not exist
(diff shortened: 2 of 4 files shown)


src/backend/distributed/master/master_metadata_utility.c
```diff
@@ -169,7 +169,17 @@ DistributedTableSize(Oid relationId, char *sizeQuery)
...
+	 * relation has been removed concurrently
+	 */
+	LockRelationOid(relationId, AccessShareLock);
+
+	if (!SearchSysCacheExists1(RELOID, ObjectIdGetDatum(relationId)))
+	{
+		ereport(ERROR, (errmsg("can't find table size, relation does not exist")));
+	}
+
 	relation = relation_open(relationId, AccessShareLock);
```
marcocitus: Maybe use `try_relation_open(relationId, AccessShareLock)`, see `pg_table_size`. For consistency, what if we also return NULL like `pg_table_size`?

src/backend/distributed/master/master_metadata_utility.c
```diff
@@ -184,7 +194,7 @@ DistributedTableSize(Oid relationId, char *sizeQuery)
 		totalRelationSize += relationSizeOnNode;
 	}
 
-	heap_close(relation, AccessShareLock);
+	heap_close(relation, NoLock);
```
marcocitus: I don't think we particularly care whether the relation gets dropped immediately after calling this function. `pg_table_size` releases the lock.

jasonmp85:
I found the title of this PR quite confusing tbh…

lithp:
@marcocitus good point on try_relation_open, that's much better.
@jasonmp85 good point on the title, changed it to something easier to parse

src/backend/distributed/master/master_metadata_utility.c
```diff
@@ -169,8 +169,12 @@ DistributedTableSize(Oid relationId, char *sizeQuery)
 							   " blocks which contain multi-shard data modifications")));
 	}
 
-	/* try to open relation, will error out if the relation does not exist */
-	relation = relation_open(relationId, AccessShareLock);
+	relation = try_relation_open(relationId, AccessShareLock);
+
+	if (relation == NULL)
+	{
+		ereport(ERROR, (errmsg("can't find table size, relation does not exist")));
```
jasonmp85: I know we (probably?) don't expect to hit this often but this message has several problems re: style.

  * _can't_ should be _cannot_ (avoid contractions)
  * Even _cannot_ should probably be _could not_, to show that this might be a temporary inability (if the table were recreated, the same invocation might work); PostgreSQL's original message here used _could not_ and I don't see a strong reason to change
  * the comma is a splice

Can we change this to: _could not compute table size: relation does not exist_?

lithp:
Okay, updated the error message

src/backend/distributed/master/master_metadata_utility.c
jasonmp85: I'm also partial to this idea, but willing to punt on it to get the current PR merged so these tests stop failing every time e.g. a new function is added.

jasonmp85:
Waitasec, I thought this would fix this line too…

https://github.com/citusdata/citus/blob/e5ddaa66af5e7de46bb00bc78a9f91d1a5d77978/src/test/regress/expected/isolation_drop_vs_all.out#L216

But it doesn't? Doesn't that mean we _still_ have to update these test files every time a new function is added, i.e. the original problem is still here?

lithp:
Good point, just tagged you in a PR.

src/backend/distributed/master/master_metadata_utility.c
lithp: For future archaeologists: I've [opened a ticket](https://github.com/citusdata/citus/issues/1705) for this

PR #1704, opened 2017-10-13
Properly use #ifdef HAVE_LIBCURL.

pykello:
After #1702 `HAVE_LIBCURL` moved to `citus_version.h` instead of being a compiler `-D` flag. We need to include `citus_version.h` in every file we are checking for `HAVE_LIBCURL`.


jasonmp85:
Oh, dur. Sorry about that!

PR #1710, opened 2017-10-16
{*,}create_distributed_table doesn't emit OID

lithp:
Finish the work started in #1703


lithp:
`create_distributed_table.c` should probably be refactored to push more work into `CreateDistributedTable`, there's some unnecessary code-duplication here.

PR #1711, opened 2017-10-17
Limit number of retries for statistics collection to once a day.

pykello:
We don't want to continue retrying statistics collection every minute forever, as it can interfere with deadlock detection. This PR limits the number of retries to once every 24 hours.

### DIFF
--- a/src/backend/distributed/utils/maintenanced.c
+++ b/src/backend/distributed/utils/maintenanced.c
@@ -214,7 +214,9 @@ CitusMaintenanceDaemonMain(Datum main_arg)
 {
 	Oid databaseOid = DatumGetObjectId(main_arg);
 	MaintenanceDaemonDBData *myDbData = NULL;
-	TimestampTz nextStatsCollectionTime USED_WITH_LIBCURL_ONLY = GetCurrentTimestamp();
+	TimestampTz nextStatsCollectionTime USED_WITH_LIBCURL_ONLY =
+		TimestampTzPlusMilliseconds(GetCurrentTimestamp(), 60 * 1000);
+	bool retryStatsCollection USED_WITH_LIBCURL_ONLY = false;
 	ErrorContextCallback errorCallback;
 
 	/*
@@ -323,17 +325,25 @@ CitusMaintenanceDaemonMain(Datum main_arg)
 				statsCollectionSuccess = CollectBasicUsageStatistics();
 			}
 
-			if (statsCollectionSuccess)
+			/*
+			 * If statistics collection was successful the next collection is
+			 * 24-hours later. Also, if this was a retry attempt we don't do
+			 * any more retries until 24-hours later, so we limit number of
+			 * retries to one.
+			 */
+			if (statsCollectionSuccess || retryStatsCollection)
 			{
 				nextStatsCollectionTime =
 					TimestampTzPlusMilliseconds(GetCurrentTimestamp(),
 												STATS_COLLECTION_TIMEOUT_MILLIS);
+				retryStatsCollection = false;
 			}
 			else
 			{
 				nextStatsCollectionTime =
 					TimestampTzPlusMilliseconds(GetCurrentTimestamp(),
 												STATS_COLLECTION_RETRY_TIMEOUT_MILLIS);
+				retryStatsCollection = true;
 			}
 
 			CommitTransactionCommand();


pykello:
@ozgune added max number of retries for stats collection. Does 3 sound good?

jasonmp85:
> We don't want to continue retrying statistics collection forever.

Why? If our collection server (or the link thereto) is down for three minutes, we stop collection forever? Where was this discussed?

jasonmp85:
I could be misreading this logic; it is kinda complex…

marcocitus:
Wait, why do we have retries at all if this is a periodic task??

marcocitus:
Can we just remove the retries altogether? Thanks.

src/backend/distributed/utils/maintenanced.c
```diff
@@ -214,9 +214,12 @@ CitusMaintenanceDaemonMain(Datum main_arg)
 {
 	Oid databaseOid = DatumGetObjectId(main_arg);
 	MaintenanceDaemonDBData *myDbData = NULL;
+	ErrorContextCallback errorCallback;
+#ifdef HAVE_LIBCURL
```
jasonmp85: This is kind of annoying just for variable declarations. Does it really make sense to define these out rather than just leave the (hopefully low-impact) `difftime` call in-place?

If not, can you investigate whether the appearance of this could be improved by adding a `USED_WITH_LIBCURL_ONLY` macro similar to `PG_USED_FOR_ASSERTS_ONLY` such that when `HAVE_LIBCURL` is false it expands to `pg_attribute_unused`?

src/backend/distributed/utils/maintenanced.c
```diff
@@ -278,10 +281,13 @@ CitusMaintenanceDaemonMain(Datum main_arg)
 		int latchFlags = WL_LATCH_SET | WL_TIMEOUT | WL_POSTMASTER_DEATH;
 		double timeout = 10000.0; /* use this if the deadlock detection is disabled */
 		bool foundDeadlock = false;
+		bool citusHasBeenLoaded = false;
+#ifdef HAVE_LIBCURL
```
jasonmp85: I guess in this case you're doing more than simple declarations. I still feel there is too much state for the little amount of computation going on here (do we really need all these variables to specify "timeout with a limited number of repetitions"?), but I could be mistaken.

jasonmp85:
After looking over this code, I don't think it does what we want. It appears to _permanently_ disable stats reporting after some number of retries. That seems wrong.

And after reading the logic needed to implement this, in this way, I have to question whether we want retries at all… I might agree with @marcocitus… should we rip them out for v1?

jasonmp85:
At _minimum_, I think the code for this could be simplified by renaming `prevStatsCollection` to `nextStatsCollection`, transforming it from a timestamp in the past to a deadline in the future.

Rather than calculating a difference to determine whether to run a stats collection, you would just check if the current timestamp is after the deadline. If the stats collection succeeds, you schedule the next collection for one day in the future. If it fails, you schedule it for, say, an hour or two or three in the future. It gets rid of all the retry logic and still results in a fairly low-traffic pattern during outages.

Thoughts?

pykello:
@jasonmp85 thanks for the feedback.

1. I am now using `nextStatsCollectionTime` instead of `prevStatsCollectionTime` as you suggested.
2. I think in one of stand-ups after your feedback we thought we can limit number of retries to 1 (please correct me if I'm wrong). I implemented that.
3. I set the retry timeout to 1 minute, thinking that first failure might be because of `LockCitusExtension()` or `CitusHasBeenLoaded()` failures (this sometimes happens when I test locally). Since we want a report when users start the server, I think we shouldn't wait more than couple of minutes for the retry. What do you think?

src/backend/distributed/utils/maintenanced.c
```diff
@@ -215,6 +215,7 @@ CitusMaintenanceDaemonMain(Datum main_arg)
 	Oid databaseOid = DatumGetObjectId(main_arg);
 	MaintenanceDaemonDBData *myDbData = NULL;
 	TimestampTz nextStatsCollectionTime USED_WITH_LIBCURL_ONLY = GetCurrentTimestamp();
```
jasonmp85: What if we initialized this to `GetCurrentTimestamp` + one minute? Do you think that would reduce false positive failures due to extension/loading problems?

pykello:
On Thu, Oct 26, 2017 at 6:08 PM, Jason Petersen <notifications@github.com>
wrote:

> This logic is fine for what it does. I'm just still trying to figure out
> whether we want to group all of these together as *failure*:
>
>    - HTTP failure or other failure in actual stats collection
>    - Citus is in need of an ALTER EXTENSION UPDATE
>    - Citus has not been loaded
>    - We cannot lock the extension
>
> Some of those feel like we shouldn't throw away our "one retry" and should
> just continue checking whether the underlying condition has been fixed.
>
> In fact, I think only the HTTP failure really should count as a failure.
> What do you think?
>
I agree with you. I think we need to think more about the details of when
and how often should we retry in these situations. Since this may take some
time and back and forth, I think we should do this in next release.

marcocitus:
> I think we should do this in next release.

I'm not sure I'm on board with that. If we get something wrong here we end up throwing warnings and whatnot every minute. If we want to keep it simple why not just remove retries? At worst we'll have to wait 24 hours and that seems absolutely fine.

pykello:
Also, made a slight change: Now using `retryStatsCollection` instead `prevStatsCollectionFailed`. In the previous implementation, the condition for postponing one day was:

```if (statsCollectionSuccess || prevStatsCollectionFailed)```

in which case, if the both attempts in first day fail, then we don't do a retry in 2nd day even if the first attempt in 2nd day fails.

pykello:
On Mon, Oct 30, 2017 at 11:02 AM, Marco Slot <notifications@github.com>
wrote:

> I think we should do this in next release.
>
> I'm not sure I'm on board with that. If we get something wrong here we end
> up throwing warnings and whatnot every minute. If we want to keep it simple
> why not just remove retries? At worst we'll have to wait 24 hours and that
> seems absolutely fine.
>

We aren't throwing warnings every minute. The retry is only once per day.
If we the first attempt fails, we retry once more in a minute and then wait
until next day.

pykello:
@marcocitus, i.e. I mean we can land this PR as it is for this release, and think about differentiating between errors as Jason suggested in next release.

marcocitus:
I misread the constant. Retrying once regardless of what goes wrong seems fine.

pykello:
Merged via #1744.

PR #1712, opened 2017-10-17
Don't collect stats when there is a version mismatch.

pykello:
The following scenario can cause an Assert() crash if we don't do this:
- Install Citus v7.0-15
- Restart server & run a query to start maintenanced.
- Install Citus v7.1
- Restart server & run a query. This will tell user to upgrade.
- Type "UPDATE EXTENSION c" & press tab. maintenanced will start and crash with Assert(CitusHasBeenLoaded() && CheckCitusVersion(WARNING));
    
This change checks Citus version before calling metadata functions so the crash doesn't happen.

### DIFF
--- a/src/backend/distributed/utils/statistics_collection.c
+++ b/src/backend/distributed/utils/statistics_collection.c
@@ -65,6 +65,16 @@ CollectBasicUsageStatistics(void)
 	memset(&unameData, 0, sizeof(unameData));
 
 	StartTransactionCommand();
+	/* 
+	 * If there is a version mismatch between loaded version and available
+	 * version, metadata functions will fail. We return early to avoid crashing.
+	 * This can happen when updating the Citus extension.
+	 */
+	if (!CheckCitusVersion(LOG_SERVER_ONLY))
+	{
+		CommitTransactionCommand();
+		return false;
+	}
 	distributedTables = DistributedTableList();
 	roundedDistTableCount = NextPow2(list_length(distributedTables));
 	roundedClusterSize = NextPow2(ClusterSize(distributedTables));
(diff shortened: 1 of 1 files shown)


src/backend/distributed/utils/statistics_collection.c
```diff
@@ -65,6 +65,17 @@ CollectBasicUsageStatistics(void)
 	memset(&unameData, 0, sizeof(unameData));
 
 	StartTransactionCommand();
+
+	/*
+	 * If there is a version mismatch between loaded version and available
+	 * version, metadata functions will fail. We return early to avoid crashing.
+	 * This can happen when updating the Citus extension.
+	 */
+	if (!CheckCitusVersion(LOG_SERVER_ONLY))
```
pykello: Using `LOG_SERVER_ONLY` to not pollute test outputs.

mtuncer:
verified that without this fix installing 7.1-2 binary on top top 7.0-15 causes postgres to crash and prevents extension update. After installing this binary you can safely run postgres and update the extension.

src/backend/distributed/utils/statistics_collection.c
mtuncer: can we move this check before ```StartTransactionCommand()``` and not call ```CommitTransactionCommand()```  ?

pykello: We cannot access catalog tables outside transactions. `CheckCitusVersion()` accesses `pg_extension`.

mtuncer: ok

mtuncer:
:shipit:

jasonmp85:
Hey @pykello sorry about not catching this; I thought we added a check but I guess it was just an is loaded check? Basically for any user-facing function or metadata-accessing code, we have starting adding the version check to guard it ever since implementing our version-checking stuff.

src/backend/distributed/utils/statistics_collection.c
jasonmp85: :-P I was going to have the same question.

Issue #1685, opened 2017-10-04
INSERT/SELECT via Coordinator with ORDER BYs may crash due to some typecasts


onderkalaci:
In coordinator `INSERT/SELECT`, we cast `SELECT` target list elements to match with the `INSERT` target list elements via [CastSelectTargetList()](https://github.com/citusdata/citus/blob/master/src/backend/distributed/planner/insert_select_planner.c#L1327) where the reason is explained on the function comment.

However, with that, we skip updating the corresponding `SortGroupClause`'s `eqop` and `sortop`. In certain cases, this ends up with crashing due to applying the wrong sort/compare operators on the already type-casted target list elements.

Steps to reproduce:

```SQL
CREATE TABLE events_table (user_id int, time timestamp, value_1 numeric);
SELECT create_distributed_table('events_table', 'user_id');

INSERT INTO events_table (user_id, value_1) VALUES (1,1), (2,2);    

CREATE TABLE agg_results (user_id int, value_1_agg int);
SELECT create_distributed_table('agg_results', 'user_id');

INSERT INTO agg_results(user_id, value_1_agg)
SELECT *
FROM (
       SELECT 
          user_id, value_1
        FROM
          events_table
           
) AS ftop
ORDER BY 2 DESC, 1 DESC
LIMIT 5;
server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
The connection to the server was lost. Attempting reset: Failed.
!> 
```

Note that the same query with the correct cast works.

```SQL
INSERT INTO agg_results(user_id, value_1_agg)
SELECT *
FROM (
       SELECT 
          user_id, value_1::int
        FROM
          events_table
           
) AS ftop
ORDER BY 2 DESC, 1 DESC
LIMIT 5;
INSERT 0 5

```

Backtrace for the crash:
```
Program received signal EXC_BAD_ACCESS, Could not access memory.
Reason: KERN_INVALID_ADDRESS at address: 0x00000000000003e7
0x000000010e8b4a10 in pg_detoast_datum (datum=0x3e7) at fmgr.c:2233
2233		if (VARATT_IS_EXTENDED(datum))
(gdb) bt
#0  0x000000010e8b4a10 in pg_detoast_datum (datum=0x3e7) at fmgr.c:2233
#1  0x000000010e7d7e20 in numeric_fast_cmp (x=999, y=999, ssup=0x7fb85886d850) at numeric.c:1754
[...]

PR #1716, opened 2017-10-20
[WIP] Add coercion to COPY

lithp:
Was meant to be a solution to https://github.com/citusdata/citus/issues/1685 but might be the wrong solution, doesn't handle CoerceViaIO nodes. Opening PR to get comments and feedback before diving any deeper.

**Motivation**

For INSERT/SELECT, it's possible to build queries which include implicit casts. In postgres this is represented something like `Query(InsertQuery, SelectQuery, Coercions)`. Our current logic turns that into `COPY(SelectQuery, DestinationRelation)` by stripping out the coercions and coercing the target list entries in SelectQuery. That causes some crashes, described in #1685, because the coercions don't update the types which other parts of the query expect those target list entries to have.

**Solution**

INSERT/SELECT no longer coerces the TargetList entries and passes them off to COPY.

COPY looks at the data types of all columns in the select query, compares those with the types of the columns in the insert query, and "manually" calls the relevant conversion functions before pushing tuples into `COPY`.

**Remaining Work**

This works great for types which have a single conversion function, however there are also `COERCION_PATH_COERCEVIAIO` pairs of types which require two functions to be called in order to to the coercion. Those are not yet supported.

### DIFF
--- a/src/backend/distributed/commands/multi_copy.c
+++ b/src/backend/distributed/commands/multi_copy.c
@@ -123,6 +123,17 @@ static int64 RemoteCreateEmptyShard(char *relationName);
 static void MasterUpdateShardStatistics(uint64 shardId);
 static void RemoteUpdateShardStatistics(uint64 shardId);
 
+static void ConversionPathForTypes(Oid inputType, Oid destType, CopyCoercionData *result);
+static Oid TypeForColumnName(Oid relationId, TupleDesc tupleDescriptor, char *columnName);
+static Oid * TypeArrayFromTupleDescriptor(TupleDesc tupleDescriptor);
+static CopyCoercionData * ColumnCoercionPaths(TupleDesc destTupleDescriptor,
+											  TupleDesc inputTupleDescriptor,
+											  Oid destRelId, List *columnNameList,
+											  Oid *finalColumnTypeArray);
+static FmgrInfo * TypeOutputFunctions(uint32 columnCount, Oid *typeIdArray,
+									  bool binaryFormat);
+static Datum CoerceColumnValue(Datum inputValue, CopyCoercionData *coercionPath);
+
 /* Private functions copied and adapted from copy.c in PostgreSQL */
 static void CopySendData(CopyOutState outputState, const void *databuf, int datasize);
 static void CopySendString(CopyOutState outputState, const char *str);
@@ -598,7 +609,7 @@ CopyToNewShards(CopyStmt *copyStatement, char *completionTag, Oid relationId)
 		/* replicate row to shard placements */
 		resetStringInfo(copyOutState->fe_msgbuf);
 		AppendCopyRowData(columnValues, columnNulls, tupleDescriptor,
-						  copyOutState, columnOutputFunctions);
+						  copyOutState, columnOutputFunctions, NULL);
 		SendCopyDataToAll(copyOutState->fe_msgbuf, currentShardId,
 						  shardConnections->connectionList);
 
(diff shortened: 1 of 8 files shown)


marcocitus:
I think the approach makes sense and seems more robust than messing with the query. What do you think needs to happen for CoerceViaIO?

lithp:
@marcocitus I can't decide what the correct approach is.

In lieu of sending `columnCoerceFunctions` I could call `columnCoerceExpressions` and have `rReceive` execute the expression. That's possibly too heavy-weight and could slow INSERT/SELECT queries.

Alternatively, I could have `columnCoerceFunctions` be a small struct which encodes all the information `rReceive` needs including all the functions and domain checking it needs to run, but that's possibly too specific, and might need to be changed whenever PG does.

marcocitus:
One thing I was thinking about is that at some point we should support `INSERT INTO ... SELECT ... RETURNING` via the coordinator, which also has the requirement that the COPY logic knows how to evaluate expressions, so maybe it's not a bad precedent to set.

lithp:
I went with the little-struct route in the interest of getting this done quickly, I've spent a lot of time on what was originally just a crash-bug. Some work left to do:
- support or error out on array casts
- support or error out on domain types
- support or error out on coercions involving typemods

src/backend/distributed/commands/multi_copy.c
```diff
@@ -1953,6 +2229,9 @@ CitusCopyDestReceiverReceive(TupleTableSlot *slot, DestReceiver *dest)
 
 		/* find the partition column value */
 		partitionColumnValue = columnValues[partitionColumnIndex];
+
+		/* annoyingly this is evaluated twice, but at least we don't crash! */
```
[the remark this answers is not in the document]
onderkalaci: Also, just FYI: https://github.com/citusdata/citus/pull/1402#discussion_r119126686, in case you'd want to implement `RETURNING` 😄

sumedhpathak:
@lithp Any updates on this PR?

src/backend/distributed/commands/multi_copy.c
```diff
@@ -1217,6 +1228,209 @@ ReportCopyError(MultiConnection *connection, PGresult *result)
 }
 
 
+static void
+ConversionPathForTypes(Oid inputType, Oid destType, CopyCoercionData *result)
+{
+	Oid coercionFuncId = InvalidOid;
+	CoercionPathType coercionType;
+
+	if (destType == inputType)
```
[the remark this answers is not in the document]
lithp: I guess it doesn't really matter, since we can use whichever value we'd like here, but `COERCION_PATH_NONE` means [conversion isn't possible](https://github.com/postgres/postgres/blob/REL_10_0/src/include/parser/parse_coerce.h#L26). And technically, a type _is_ binary castable to itself, so `RELABLE` isn't the wrong thing to send.

src/backend/distributed/commands/multi_copy.c
```diff
@@ -1217,6 +1228,209 @@ ReportCopyError(MultiConnection *connection, PGresult *result)
...
+		Oid inputTupleType = inputTupleTypes[columnIndex];
+		char *columnName = lfirst(currentColumnName);
+
+		if (inputTupleType == InvalidOid)
+		{
+			/* this was a dropped column and will not be in the incoming tuples */
+			continue;
+		}
+
+		destTupleType = TypeForColumnName(destRelId, destTupleDescriptor, columnName);
```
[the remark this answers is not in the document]
lithp: That's essentially what I'm doing, except instead of pulling the attribute out of the cache, I'm using the attribute which is already stored in destTupleDescriptor. I'll still need to call `get_attnum` though, since I only have a column name! I can swap the call to `TupleDescAttr` with a call to `get_atttype`, but not sure what's it wins us.

src/backend/distributed/commands/multi_copy.c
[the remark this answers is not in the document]
lithp: IIIInteresting. Maybe it's because adding the type casts does yet another walk of the tree? I'm actually surprised, the difference shouldn't be noticeable!

lithp:
@onderkalaci I updated the tests but I'm unsure why we're no longer wrapping this subquery, I only changed things which are downstream of `WrapSubquery`.

src/test/regress/expected/multi_insert_select.out
```diff
@@ -2504,6 +2504,73 @@ SELECT * FROM drop_col_table WHERE col2 = '1';
...
+) AS ftop
+LIMIT 5;
+SELECT * FROM coerce_agg;
+ user_id | value_1_agg 
+---------+-------------
+       1 | aaaaa
+       2 | bbbbb
+(2 rows)
+
+-- TODO: Arrays! Domains! Collations!
```
marcocitus: Can you resolve this TODO?

src/test/regress/expected/multi_insert_select.out
```diff
@@ -2504,6 +2504,73 @@ SELECT * FROM drop_col_table WHERE col2 = '1';
...
+DROP TABLE coerce_events;
+DROP TABLE coerce_agg;
 DROP TABLE drop_col_table;
 DROP TABLE raw_table;
 DROP TABLE summary_table;
```
marcocitus: Random: could we put all the DROP TABLEs in a transaction block? That's just because I've started noticing that part of our regression test time goes into dropping tables and every transaction that drops a table does a scan of shared_buffers, so doing it one transaction is more efficient.

src/backend/distributed/commands/multi_copy.c
```diff
@@ -1346,6 +1566,49 @@ AppendCopyRowData(Datum *valueArray, bool *isNullArray, TupleDesc rowDescriptor,
 }
 
 
+static Datum
+CoerceColumnValue(Datum inputValue, CopyCoercionData *coercionPath)
```
marcocitus: Missing function comment

src/include/distributed/multi_copy.h
```diff
@@ -51,6 +52,15 @@ typedef struct NodeAddress
 	int32 nodePort;
 } NodeAddress;
 
+/* struct to allow rReceive to coerce tuples before sending them to workers  */
+typedef struct CopyCoercionData
+{
+	CoercionPathType coercionType;
+	FmgrInfo firstCoerceFunction;
+	FmgrInfo secondCoerceFunction;
```
marcocitus: This could use better naming. Maybe:

```
/* function to use for direct coercions*/
FmgrInfo coerceFunction;

/* input and output functions for coercion via I/O */
FmgrInfo inputFunction;
FmgrInfo outputFunction;
```

I don't think the extra memory usage is really a concern, but alternative you could use pointers.

src/backend/distributed/commands/multi_copy.c
```diff
@@ -1217,6 +1228,209 @@ ReportCopyError(MultiConnection *connection, PGresult *result)
 }
 
 
+static void
+ConversionPathForTypes(Oid inputType, Oid destType, CopyCoercionData *result)
```
marcocitus: Missing function comment.

src/backend/distributed/commands/multi_copy.c
```diff
@@ -1217,6 +1228,209 @@ ReportCopyError(MultiConnection *connection, PGresult *result)
...
+{
+	Oid coercionFuncId = InvalidOid;
+	CoercionPathType coercionType;
```
marcocitus: Maybe initialise to `COERCION_PATH_RELABELTYPE`

src/backend/distributed/commands/multi_copy.c
```diff
@@ -1217,6 +1228,209 @@ ReportCopyError(MultiConnection *connection, PGresult *result)
...
+
+	coercionType = find_coercion_pathway(destType, inputType,
+										 COERCION_EXPLICIT,
+										 &coercionFuncId);
+
+	switch (coercionType)
+	{
+		case COERCION_PATH_NONE:
+		{
+			ereport(ERROR, (errmsg("can not coerce %d to %d", inputType, destType)));
```
marcocitus: postgres uses "cannot cast" for these messages, because coercion is not a word used by mortals.

src/backend/distributed/commands/multi_copy.c
```diff
@@ -1217,6 +1228,209 @@ ReportCopyError(MultiConnection *connection, PGresult *result)
...
+
+		case COERCION_PATH_COERCEVIAIO:
+		{
+			result->coercionType = COERCION_PATH_COERCEVIAIO;
+
+			{
+				bool typisvarlena = false; /* ignored */
+				Oid iofunc = InvalidOid;
+				getTypeOutputInfo(inputType, &iofunc, &typisvarlena);
+				fmgr_info(iofunc, &(result->firstCoerceFunction));
```
marcocitus: here using inputFunction and outputFunction names would be nice

src/backend/distributed/commands/multi_copy.c
```diff
@@ -1853,9 +2116,19 @@ CitusCopyDestReceiverStartup(DestReceiver *dest, int operation,
...
-		ColumnOutputFunctions(inputTupleDescriptor, copyOutState->binary);
+	/* prepare functions to call on received tuples */
+	{
+		TupleDesc destTupleDescriptor = distributedRelation->rd_att;
+		int columnCount = inputTupleDescriptor->natts;
+		Oid *finalTypeArray = palloc0(columnCount * sizeof(Oid));
+
+		copyDest->columnCoercionPaths =
+			ColumnCoercionPaths(destTupleDescriptor, inputTupleDescriptor,
+								tableId, columnNameList, finalTypeArray);
```
marcocitus: Is it worth skipping this for regular COPY commands?

src/backend/distributed/commands/multi_copy.c
```diff
@@ -1217,6 +1228,209 @@ ReportCopyError(MultiConnection *connection, PGresult *result)
...
+
+	return coercePaths;
+}
+
+
+/*
+ * TypeOutputFunctions takes an array of types and returns an array of output functions
+ */
+static FmgrInfo *
+TypeOutputFunctions(uint32 columnCount, Oid *typeIdArray, bool binaryFormat)
```
marcocitus: Now that we have this function, could we get rid of `ColumnOutputFunctions`?

src/backend/distributed/commands/multi_copy.c
lithp: I really dislike this convention of ours, we're just throwing away useful compiler errors.

src/backend/distributed/commands/multi_copy.c
lithp: The problem is that what kind of function they are changes depending on which path is being taken. For `COERCEVIAIO` they're outputFunction and inputFunction, respectively. But for `FUNC` they're just a conversion function. 

What about wasting some memory by making the struct bigger and only using the functions which are relevant to a given path?

marcocitus: > What about wasting some memory by making the struct bigger and only using the functions which are relevant to a given path?

That's all I ever wanted.

src/include/distributed/multi_copy.h
lithp: Oh, you had the same idea. Yeah, I'll do this! An extra `FmgrInfo` per column isn't going to make a noticeable difference in memory usage.

src/backend/distributed/commands/multi_copy.c
lithp: That would make sense, how would you distinguish between the two cases? I don't think it makes a big difference, presumably the types will always match, we'll get a lot of `RELABEL_TYPE`, and receiving isn't slowed down at all, despite startup doing a little more work.

lithp:
You're 9 minutes too early :laughing: 

Take a look at the tests I just pushed! We could have better error messages, or error out sooner in some places, but there are no crashes and no wrong results.

src/backend/distributed/commands/multi_copy.c
jasonmp85: +1,000,000,000

src/backend/distributed/commands/multi_copy.c
jasonmp85: This isn't strictly true:

```
~/D/C/postgresql ❯❯❯ ack 'errmsg.*coerc'
src/backend/catalog/pg_aggregate.c
860:					 errmsg("function %s requires run-time type coercion",

src/backend/parser/parse_oper.c
480:				 errmsg("operator requires run-time type coercion: %s",

src/backend/commands/functioncmds.c
1490:					 errmsg("argument of cast function must match or be binary-coercible from source data type")));
1504:					 errmsg("return data type of cast function must match or be binary-coercible to target data type")));

src/backend/commands/prepare.c
380:					 errmsg("parameter $%d of type %s cannot be coerced to the expected type %s",
```

That said, _cannot cast_ is more common in `parse_coerce.c`, so ultimately agreed.

src/test/regress/expected/multi_insert_select.out
```diff
@@ -2504,6 +2504,162 @@ SELECT * FROM drop_col_table WHERE col2 = '1';
...
+INSERT INTO coerce_agg(user_id, value_1_agg)
+SELECT *
+FROM (
+  SELECT user_id, value_1
+  FROM coerce_events
+) AS ftop
+LIMIT 5;
+ERROR:  column "value_1_agg" is of type integer but expression is of type integer[]
+LINE 2: SELECT *
+               ^
```
marcocitus: Is this the expected output of the test?

lithp: It's a good point, I should probably get rid of that test, it's not testing anything in Citus

marcocitus:
Apart from the test and squashing, this seems ready to :ship:.

lithp:
Okay. Updated the test, squashed, and rebased onto master. Just waiting for the tests to pass.

PR #1719, opened 2017-10-20
Replace uint with uint32

lithp:
A mistake I made which breaks the build on some platforms, `uint` isn't part of any C standard.


sumedhpathak:
@lithp Seems like this is approved. Do you want to merge? Thanks.

lithp:
@sumedhpathak Yep, just did a rebase. Once the tests pass I'll merge

PR #1725, opened 2017-10-24
Block only 2PCs instead of all writes in citus_create_restore_point

marcocitus:
`citus_create_restore_point` currently blocks all writes, which gives a fully consistent restore point, but also blocks all writes until all open write transactions have finished. This behaviour is generally undesirable.

Fortunately, we can achieve the main goal of `citus_create_restore_point` by taking an exclusive lock on `pg_dist_transaction`, which blocks 2PCs just before commit, which ensures the restore point is consistent in terms of DDL, reference tables replicas, and committed distributed transactions when using 2PC. It does not block 1PCs, since those do not write to `pg_dist_transaction`, but this seems like a reasonable trade-off since 1PC fundamentally doesn't provide atomicity guarantees.

By only blocking at commit time, `citus_create_restore_point` at most has to wait until all in-progress commits finish, which is at most 1 round trip.

Adds some additional isolation tests to make sure we're blocking 2PCs.

### DIFF
--- a/src/test/regress/specs/isolation_create_restore_point.spec
+++ b/src/test/regress/specs/isolation_create_restore_point.spec
@@ -1,5 +1,6 @@
 setup
 {
+	SET citus.shard_replication_factor TO 1;
 	CREATE TABLE restore_table (test_id integer NOT NULL, data text);
 	SELECT create_distributed_table('restore_table', 'test_id');
 }
@@ -14,6 +15,7 @@ session "s1"
 step "s1-begin"
 {
 	BEGIN;
+	SET citus.multi_shard_commit_protocol TO '2pc';
 }
 
 step "s1-create-distributed"
@@ -29,7 +31,16 @@ step "s1-insert"
 
 step "s1-modify-multiple"
 {
-	SELECT master_modify_multiple_shards($$UPDATE restore_table SET data = 'world'$$);
+	UPDATE restore_table SET data = 'world';
+}
+
+step "s1-multi-statement"
+{
+	SET citus.multi_shard_commit_protocol TO '2pc';
+	BEGIN;
+	INSERT INTO restore_table VALUES (1,'hello');
+	INSERT INTO restore_table VALUES (2,'hello');
+	COMMIT;
 }
 
 step "s1-ddl"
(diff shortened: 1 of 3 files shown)


src/test/regress/specs/isolation_create_restore_point.spec
```diff
@@ -92,3 +126,15 @@ permutation "s1-begin" "s1-add-node" "s2-create-restore" "s1-commit"
...
+# verify that citus_create_restore_point is blocked by concurrent citus_create_restore_point
+permutation "s1-begin" "s1-create-restore" "s2-create-restore" "s1-commit"
+
+# verify that master_modify_multiple_shards is blocked by concurrent citus_create_restore_point
+permutation "s2-begin" "s2-create-restore" "s1-modify-multiple" "s2-commit"
+
+# verify that DDL is blocked by concurrent citus_create_restore_point
+permutation "s2-begin" "s2-create-restore" "s1-ddl" "s2-commit"
+
+# verify that multi-statement transactions iareblocked by concurrent citus_create_restore_point
```
aamederen: Typo

src/test/regress/specs/isolation_create_restore_point.spec
```diff
@@ -64,26 +85,39 @@ step "s1-commit"
...
+
 # verify that citus_create_restore_point is blocked by concurrent create_distributed_table
 permutation "s1-begin" "s1-create-distributed" "s2-create-restore" "s1-commit"
 
-# verify that citus_create_restore_point is blocked by concurrent INSERT
+# verify that citus_create_restore_point is not blocked by concurrent INSERT
 permutation "s1-begin" "s1-insert" "s2-create-restore" "s1-commit"
 
-# verify that citus_create_restore_point is blocked by concurrent master_modify_multiple_shards
+# verify that citus_create_restore_point is not blocked by concurrent master_modify_multiple_shards
```
aamederen: It took me a while to get why the `create-restore` is not blocked `modify-multiple`, 2PC transactions are not blocked until `COMMIT`. Maybe a note to the header of `citus_create_restore_point` function mentioning this would be useful for future readers.

src/test/regress/specs/isolation_create_restore_point.spec
```diff
@@ -32,6 +34,15 @@ step "s1-modify-multiple"
...
 
+step "s1-multi-statement"
+{
+	SET citus.multi_shard_commit_protocol TO '2pc';
+	BEGIN;
+	INSERT INTO restore_table VALUES (1,'hello');
+	INSERT INTO restore_table VALUES (2,'hello');
+	COMMIT;
+}
+
```
aamederen: Would it make sense to add a multi shard update/delete (not `master_modify_multiple_shards`) case as well?

Issue #1705, opened 2017-10-13
Maybe citus_total_relation_size should return NULL when the relation does not exist


lithp:
This would be consistent with `pg_table_size`, more discussion is in #1703.

ozgune:
@lithp -- The pull request related to this issue has been merged. Can we close this issue?

lithp:
@ozgune no, both of those PRs reference this issue but neither addresses it. [Here's the specific comment](https://github.com/citusdata/citus/pull/1703#discussion_r144495872) this issue refers to.

PR #1726, opened 2017-10-24
Follow the patterns used in Deadlock Detection in Stats Collection.

pykello:
This includes:

1. Wrap everything inside a `StartTransactionCommand()`/`CommitTransactionCommand()`.
This is so we can access the database. This also switches to a new memory context
and releases it, so we don't have to do our own memory management.
2. `LockCitusExtension()` so the extension cannot be dropped or created concurrently.
3. Check `CitusHasBeenLoaded() && CheckCitusVersion()` before doing any work.
4. Do not `PG_TRY()` inside a loop.

### DIFF
--- a/src/backend/distributed/utils/maintenanced.c
+++ b/src/backend/distributed/utils/maintenanced.c
@@ -214,8 +214,7 @@ CitusMaintenanceDaemonMain(Datum main_arg)
 {
 	Oid databaseOid = DatumGetObjectId(main_arg);
 	MaintenanceDaemonDBData *myDbData = NULL;
-	time_t prevStatsCollection = 0;
-	bool prevStatsCollectionFailed = false;
+	TimestampTz nextStatsCollectionTime USED_WITH_LIBCURL_ONLY = GetCurrentTimestamp();
 	ErrorContextCallback errorCallback;
 
 	/*
@@ -278,30 +277,17 @@ CitusMaintenanceDaemonMain(Datum main_arg)
 		int latchFlags = WL_LATCH_SET | WL_TIMEOUT | WL_POSTMASTER_DEATH;
 		double timeout = 10000.0; /* use this if the deadlock detection is disabled */
 		bool foundDeadlock = false;
-		time_t currentTime = time(NULL);
-		double secondsSincePrevStatsCollection = difftime(currentTime,
-														  prevStatsCollection);
-		bool citusHasBeenLoaded = false;
 
 		CHECK_FOR_INTERRUPTS();
 
-		StartTransactionCommand();
-		citusHasBeenLoaded = CitusHasBeenLoaded();
-		CommitTransactionCommand();
-
-		if (!citusHasBeenLoaded)
-		{
-			continue;
-		}
-
 		/*
-		 * XXX: We clear the metadata cache before every iteration because otherwise
-		 * it might contain stale OIDs. It appears that in some cases invalidation
-		 * messages for a DROP EXTENSION may arrive during deadlock detection and
+		 * XXX: Each task should clear the metadata cache before every iteration
+		 * by calling InvalidateMetadataSystemCache(), because otherwise it
+		 * might contain stale OIDs. It appears that in some cases invalidation
+		 * messages for a DROP EXTENSION may arrive during these tasks and
 		 * this causes us to cache a stale pg_dist_node OID. We'd actually expect
 		 * all invalidations to arrive after obtaining a lock in LockCitusExtension.
 		 */
-		InvalidateMetadataSystemCache();
 
 		/*
 		 * Perform Work.  If a specific task needs to be called sooner than
@@ -309,35 +295,54 @@ CitusMaintenanceDaemonMain(Datum main_arg)
 		 * tasks should do their own time math about whether to re-run checks.
 		 */
 
-		if (secondsSincePrevStatsCollection >= STATISTICS_COLLECTION_INTERVAL ||
-			(prevStatsCollectionFailed &&
-			 secondsSincePrevStatsCollection >= STATISTICS_COLLECTION_RETRY_INTERVAL))
-		{
 #ifdef HAVE_LIBCURL
-			if (EnableStatisticsCollection)
+		if (EnableStatisticsCollection &&
+			GetCurrentTimestamp() >= nextStatsCollectionTime)
+		{
+			bool statsCollectionSuccess = false;
+			InvalidateMetadataSystemCache();
+			StartTransactionCommand();
+
+			/*
+			 * Lock the extension such that it cannot be dropped or created
+			 * concurrently. Skip statistics collection if citus extension is
+			 * not accessible.
+			 *
+			 * Similarly, we skip statistics collection if there exists any
+			 * version mismatch or the extension is not fully created yet.
+			 */
+			if (!LockCitusExtension())
 			{
-				MemoryContext statsCollectionContext =
-					AllocSetContextCreate(CurrentMemoryContext,
-										  "StatsCollection",
-										  ALLOCSET_DEFAULT_MINSIZE,
-										  ALLOCSET_DEFAULT_INITSIZE,
-										  ALLOCSET_DEFAULT_MAXSIZE);
-				MemoryContext oldContext =
-					MemoryContextSwitchTo(statsCollectionContext);
-
-				WarnIfSyncDNS();
-				prevStatsCollectionFailed = !CollectBasicUsageStatistics();
-
-				MemoryContextSwitchTo(oldContext);
-				MemoryContextDelete(statsCollectionContext);
-				prevStatsCollection = currentTime;
+				ereport(DEBUG1, (errmsg("could not lock the citus extension, "
+										"skipping statistics collection")));
 			}
-#endif
+			else if (CheckCitusVersion(DEBUG1) && CitusHasBeenLoaded())
+			{
+				WarnIfSyncDNS();
+				statsCollectionSuccess = CollectBasicUsageStatistics();
+			}
+
+			if (statsCollectionSuccess)
+			{
+				nextStatsCollectionTime =
+					TimestampTzPlusMilliseconds(GetCurrentTimestamp(),
+												STATS_COLLECTION_TIMEOUT_MILLIS);
+			}
+			else
+			{
+				nextStatsCollectionTime =
+					TimestampTzPlusMilliseconds(GetCurrentTimestamp(),
+												STATS_COLLECTION_RETRY_TIMEOUT_MILLIS);
+			}
+
+			CommitTransactionCommand();
 		}
+#endif
 
 		/* the config value -1 disables the distributed deadlock detection  */
 		if (DistributedDeadlockDetectionTimeoutFactor != -1.0)
 		{
+			InvalidateMetadataSystemCache();
 			StartTransactionCommand();
 
 			/*
--- a/src/backend/distributed/utils/maintenanced.c
+++ b/src/backend/distributed/utils/maintenanced.c
@@ -318,6 +318,7 @@ CitusMaintenanceDaemonMain(Datum main_arg)
 			}
 			else if (CheckCitusVersion(DEBUG1) && CitusHasBeenLoaded())
 			{
+				FlushDistTableCache();
 				WarnIfSyncDNS();
 				statsCollectionSuccess = CollectBasicUsageStatistics();
 			}
--- a/src/backend/distributed/utils/metadata_cache.c
+++ b/src/backend/distributed/utils/metadata_cache.c
@@ -177,6 +177,7 @@ static void RegisterWorkerNodeCacheCallbacks(void);
 static void RegisterLocalGroupIdCacheCallbacks(void);
 static uint32 WorkerNodeHashCode(const void *key, Size keySize);
 static void ResetDistTableCacheEntry(DistTableCacheEntry *cacheEntry);
+static void CreateDistTableCache(void);
 static void InvalidateDistRelationCacheCallback(Datum argument, Oid relationId);
 static void InvalidateNodeRelationCacheCallback(Datum argument, Oid relationId);
 static void InvalidateLocalGroupIdRelationCacheCallback(Datum argument, Oid relationId);
@@ -2345,13 +2346,7 @@ InitializeDistTableCache(void)
 	DistShardScanKey[0].sk_attno = Anum_pg_dist_shard_logicalrelid;
 
 	/* initialize the per-table hash table */
-	MemSet(&info, 0, sizeof(info));
-	info.keysize = sizeof(Oid);
-	info.entrysize = sizeof(DistTableCacheEntry);
-	info.hash = tag_hash;
-	DistTableCacheHash =
-		hash_create("Distributed Relation Cache", 32, &info,
-					HASH_ELEM | HASH_FUNCTION);
+	CreateDistTableCache();
 
 	/* initialize the per-shard hash table */
 	MemSet(&info, 0, sizeof(info));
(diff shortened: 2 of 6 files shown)


pykello:
@marcocitus @onderkalaci can one of you please take a look at this and `src/backend/distributed/utils/statistics_collection.c` in general to see if I've missed anything?

@jasonmp85 can you also please take a look as someone who is familiar with the logic of statistics collection?

pykello:
While writing the test document, I noticed:
1. Based on [`man 2 uname`](http://man7.org/linux/man-pages/man2/uname.2.html), `uname(buf)` can only fail when `buf` is invalid. Maybe we should just add an `Assert(uname(&unameData) == 0)` to make it explicit that we expect it to always succeed.
2. `pg_backend_random()` can fail (e.g. when we opening `/dev/random` fails). In our current implementation of `citus_server_id()`, we do `ereport(ERROR, ...)` in this case. Since we are calling this function for `CREATE EXTENSION ...`, this can fail extension creation. We should probably just fallback to using random generation using seeds instead, and avoid throwing errors.
3. Some of metadata lookup functions can throw errors. We should catch them instead of letting `maintenanced` crash. Maybe move the `PG_TRY()` in `ClusterSize()` to `CollectBasicUsageStatistics()`?

src/backend/distributed/utils/maintenanced.c
```diff
@@ -309,31 +296,45 @@ CitusMaintenanceDaemonMain(Datum main_arg)
...
-										  ALLOCSET_DEFAULT_MINSIZE,
-										  ALLOCSET_DEFAULT_INITSIZE,
-										  ALLOCSET_DEFAULT_MAXSIZE);
-				MemoryContext oldContext =
-					MemoryContextSwitchTo(statsCollectionContext);
+		if (EnableStatisticsCollection &&
+			difftime(time(NULL), prevStatsCollection) >= statsCollectionInterval)
+		{
+			bool statsCollectionSuccess = false;
+			StartTransactionCommand();
```
marcocitus: It might make sense to call `InvalidateMetadataSystemCache()` at the start of the `if` statement just before starting the transaction, and do that again in the deadlock detection. Otherwise, we take cached OIDs from the call home into the deadlock detection, which could cause trouble.

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -128,28 +114,31 @@ ClusterSize(List *distributedTableList)
...
+		ErrorData *edata = NULL;
+		MemoryContextSwitchTo(savedContext);
+		edata = CopyErrorData();
+		FlushErrorState();
+
+		/* rethrow as WARNING */
+		edata->elevel = WARNING;
+		ThrowErrorData(edata);
+	}
+	PG_END_TRY();
```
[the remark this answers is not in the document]
onderkalaci: And also, can we add a regression test that triggers this function? Currently, the enterprise-master Travis builds fails due to low coverage. It might help to increase to coverage.

Thanks!

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -112,46 +133,41 @@ CollectBasicUsageStatistics(void)
...
 	foreach(distTableCacheEntryCell, distributedTableList)
 	{
 		DistTableCacheEntry *distTableCacheEntry = lfirst(distTableCacheEntryCell);
 		Oid relationId = distTableCacheEntry->relationId;
-		MemoryContext savedContext = CurrentMemoryContext;
 
-		PG_TRY();
+		/*
+		 * Ignore hash partitioned tables with size greater than 1, since
+		 * cstore_table_size() doesn't work on them.
```
onderkalaci: I guess `cstore_table_size` should be `citus_*` ?

src/backend/distributed/utils/maintenanced.c
```diff
@@ -278,22 +280,9 @@ CitusMaintenanceDaemonMain(Datum main_arg)
...
-		citusHasBeenLoaded = CitusHasBeenLoaded();
-		CommitTransactionCommand();
-
-		if (!citusHasBeenLoaded)
-		{
-			continue;
-		}
-
 		/*
 		 * XXX: We clear the metadata cache before every iteration because otherwise
```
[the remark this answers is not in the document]

src/backend/distributed/utils/statistics_collection.c

pykello:
On Thu, Oct 26, 2017 at 12:02 PM, Önder Kalacı <notifications@github.com>
wrote:



> There is a window between DistributedTableList() and DistributedTablesSize
> () on which if a distributed table is dropped, we'd error out on
> citus_table_size().
>
> So, to prevent that, we could consider obtaining AcessShareLock on the
> relations. But, if you think that's acceptable, I'm OK with that as well.
>
This doesn't seem too difficult to do, so I just made the change. See
DistributedTablesSize().

I don't have much experience with distributed dead locks, but do you think
any of the locks we take in stats collection can cause distributed
deadlocks? I guess no since we aren't doing any modifications here, but
thought I can ask.


> Also, to simplify things (get rid of some of the metadata cache accesses),
> I'd use DistTableOidList() function, which returns a list of Oids --
> that's what you're using anyway.
>
Done.

src/backend/distributed/utils/maintenanced.c
```diff
@@ -309,35 +296,54 @@ CitusMaintenanceDaemonMain(Datum main_arg)
...
+			 * Lock the extension such that it cannot be dropped or created
+			 * concurrently. Skip statistics collection if citus extension is
+			 * not accessible.
+			 *
+			 * Similarly, we skip statistics collection if there exists any
+			 * version mismatch or the extension is not fully created yet.
+			 */
+			if (!LockCitusExtension())
+			{
+				ereport(DEBUG1, (errmsg("could not lock the citus extension, "
```
jasonmp85: Shouldn't this be higher than `DEBUG1`? We'll never see it at that level. I could see `WARN` making sense.

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -67,36 +71,53 @@ WarnIfSyncDNS(void)
...
-	roundedDistTableCount = NextPow2(list_length(distributedTables));
-	roundedClusterSize = NextPow2(ClusterSize(distributedTables));
-	workerNodeCount = ActivePrimaryNodeCount();
-	metadataJsonbDatum = DistNodeMetadata();
-	metadataJsonbStr = DatumGetCString(DirectFunctionCall1(jsonb_out,
-														   metadataJsonbDatum));
-	uname(&unameData);
+
+	/* uname() always succeed if we pass valid buffer */
+	Assert(uname(&unameData) == 0);
```
jasonmp85: Uhhhh… _please_ do not perform meaningful operations in the body of an `Assert`! The standard PostgreSQL build has asserts disabled; this becomes a no-op!

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -112,46 +133,53 @@ CollectBasicUsageStatistics(void)
...
 		}
-		PG_CATCH();
-		{
-			FlushErrorState();
 
-			/* citus_table_size() throws an error while the memory context is changed */
-			MemoryContextSwitchTo(savedContext);
+		/*
+		 * Ignore hash partitioned tables with size greater than 1, since
+		 * citus_table_size() doesn't work on them.
```
jasonmp85: Should we open an issue to track this? Seems like a shortcoming, especially for more legacy-ish users.

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -112,46 +133,53 @@ CollectBasicUsageStatistics(void)
...
-
-		PG_TRY();
+		Oid relationId = lfirst_oid(distTableOidCell);
+
+		/*
+		 * Relations can get dropped after getting the Oid list and before we
+		 * reach here. Acquire a lock to make sure the relation is available
+		 * while we are getting its size.
+		 */
+		Relation relation = try_relation_open(relationId, AccessShareLock);
```
jasonmp85: Could we get rid of this complication if we just implemented #1705 instead?

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -112,46 +133,53 @@ CollectBasicUsageStatistics(void)
...
+		 * citus_table_size() doesn't work on them.
+		 */
+		if (PartitionMethod(relationId) == DISTRIBUTE_BY_HASH &&
+			!SingleReplicatedTable(relationId))
+		{
+			continue;
 		}
-		PG_END_TRY();
+
+		totalSize += DatumGetInt64(
```
jasonmp85: This seems like a lot of operations for a single line. Can you break out the function call to one line (and assign to a `Datum`, maybe `tableSizeDatum`) and have the addition on a second line?

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -254,19 +280,21 @@ citus_server_id(PG_FUNCTION_ARGS)
 	uint8 *buf = (uint8 *) palloc(UUID_LEN);
 
 #if PG_VERSION_NUM >= 100000
+
+	/*
+	 * If pg_backend_random() fails, fall-back to using random(). In previous
+	 * versions of postgres we don't have pg_backend_random(), so use it by
+	 * default in that case.
+	 */
 	if (!pg_backend_random((char *) buf, UUID_LEN))
```
jasonmp85: Hm; I'm not 100% sure what I think about this, but I guess we'll go with it.

src/include/distributed/maintenanced.h
```diff
@@ -13,10 +13,10 @@
 #define MAINTENANCED_H
 
 /* collect statistics every 24 hours */
-#define STATISTICS_COLLECTION_INTERVAL 86400
+#define STATS_COLLECTION_TIMEOUT_MILLIS 86400000
```
jasonmp85: Can you change to `(24 * 60 * 60 * 1000)` or something?

src/include/distributed/maintenanced.h
```diff
@@ -13,10 +13,10 @@
...
 
 /* if statistics collection fails, retry in 1 minute */
-#define STATISTICS_COLLECTION_RETRY_INTERVAL 60
+#define STATS_COLLECTION_RETRY_TIMEOUT_MILLIS 60000
```
jasonmp85: Can you change to `(60 * 1000)`?

src/backend/distributed/utils/maintenanced.c
```diff
@@ -309,35 +296,54 @@ CitusMaintenanceDaemonMain(Datum main_arg)
...
 
+			/*
+			 * Lock the extension such that it cannot be dropped or created
+			 * concurrently. Skip statistics collection if citus extension is
+			 * not accessible.
+			 *
+			 * Similarly, we skip statistics collection if there exists any
+			 * version mismatch or the extension is not fully created yet.
+			 */
+			if (!LockCitusExtension())
```
jasonmp85: What do you think about differentiating between the failure to lock the extension and the failure to submit statistics? I feel like if the extension lock fails, retrying in a minute or so makes sense; if the stats submission fails, why retry in a minute? It seems like a good way to churn away hitting our stats server if it's having a brownout.

src/backend/distributed/utils/maintenanced.c
```diff
@@ -278,26 +277,14 @@ CitusMaintenanceDaemonMain(Datum main_arg)
...
-		 * XXX: We clear the metadata cache before every iteration because otherwise
-		 * it might contain stale OIDs. It appears that in some cases invalidation
-		 * messages for a DROP EXTENSION may arrive during deadlock detection and
+		 * XXX: Each task should clear the metadata cache before every iteration
+		 * by calling InvalidateMetadataSystemCache(), because otherwise it
+		 * might contain stale OIDs. It appears that in some cases invalidation
+		 * messages for a DROP EXTENSION may arrive during these tasks and
 		 * this causes us to cache a stale pg_dist_node OID. We'd actually expect
 		 * all invalidations to arrive after obtaining a lock in LockCitusExtension.
 		 */
```
marcocitus: This call to `InvalidateMetadataSystemCache` can be removed

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -74,29 +78,46 @@ CollectBasicUsageStatistics(void)
...
 	char *metadataJsonbStr = NULL;
+	MemoryContext savedContext = CurrentMemoryContext;
 	struct utsname unameData;
+	bool metadataCollectionFailed = false;
 	memset(&unameData, 0, sizeof(unameData));
 
-	StartTransactionCommand();
+	PG_TRY();
+	{
+		distributedTables = DistributedTableList();
```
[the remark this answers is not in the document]
marcocitus: Because we clear OIDs in `InvalidateMetadataSystemCache()`, we end up ignoring regular invalidations, so the metadata cache can be arbitrarily out of date (e.g. `citus_table_size` might always fail after a rebalance).

We have some code for fully invalidating the metadata cache in `InvalidateDistRelationCacheCallback`, which we could wrap in a separate function and call before the call home code.

onderkalaci:
> I don't have much experience with distributed dead locks, but do you think
any of the locks we take in stats collection can cause distributed
deadlocks? 

No, we should be fine.

src/backend/distributed/utils/statistics_collection.c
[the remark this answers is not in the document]
onderkalaci: > We have some code for fully invalidating the metadata cache in InvalidateDistRelationCacheCallback, which we could wrap in a separate function and call before the call home code.

Makes sense to me as well. I think it'll also be useful for the future changes on the daemon.

(@pykello pinging you on this comment, in case this is lost among the others. The following code from the mentioned function should do the trick: 

```C
		DistTableCacheEntry *cacheEntry = NULL;
		HASH_SEQ_STATUS status;

		hash_seq_init(&status, DistTableCacheHash);

		while ((cacheEntry = (DistTableCacheEntry *) hash_seq_search(&status)) != NULL)
		{
			cacheEntry->isValid = false;
		}
```

src/backend/distributed/utils/statistics_collection.c

src/backend/distributed/utils/statistics_collection.c
pykello: Done at #1735.

src/backend/distributed/utils/maintenanced.c

src/backend/distributed/utils/statistics_collection.c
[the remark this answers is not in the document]
pykello: Done. Now calling `InvalidateDistRelationCache()` before statistics collection.

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -2731,6 +2723,22 @@ InvalidateDistRelationCacheCallback(Datum argument, Oid relationId)
...
+InvalidateDistRelationCache(void)
+{
+	DistTableCacheEntry *cacheEntry = NULL;
+	HASH_SEQ_STATUS status;
+
+	hash_seq_init(&status, DistTableCacheHash);
+
+	while ((cacheEntry = (DistTableCacheEntry *) hash_seq_search(&status)) != NULL)
+	{
+		cacheEntry->isValid = false;
```
marcocitus: There's a slight problem here that we never actually clean entries of dropped relations, which creates a memory leak. We've introduced these flags because invalidations can arrive in lots of different places, and callers might still be using the cache entries, so it wasn't safe to immediately free them. However, since we're calling this code explicitly before we do anything else, and only do this once or twice a day, maybe we should free the entries and do a hash_destroy.

(ideally, we wouldn't use the cache)

marcocitus: Note that this leak also exists for regular sessions, but we found this an acceptable trade-off because sessions are ultimately short-lived.

pykello: Done.

src/backend/distributed/utils/maintenanced.c
jasonmp85: Sounds good.

src/backend/distributed/utils/maintenanced.c
jasonmp85: Oh, it's `DEBUG1` in deadlock detection? OK, then no problem leaving it the same here.

marcocitus: Partly because we figured this might spam the logs in deadlock detection if a DROP EXTENSION is stuck. If you really want to see it you can set `log_min_messages` to DEBUG1 in postgresql.conf. I think we can just leave it as DEBUG1 in either case, since it's mostly interesting for debugging.

pykello:
Merged via #1744.

PR #1731, opened 2017-10-25
ALTER TABLE .. REPLICA IDENTITY support is implemented

furkansahin:
I made a small change in multi_utility.c to allow ALTER TABLE .. REPLICA IDENTITY queries to be processed just like the other Alter Table queries. 

Another change in the RelayEventExtendNames to catch if the function call is made for a ALTER TABLE REPLICA IDENTITY query and if it has USING INDEX add the shard id into the index name as well.

### DIFF
--- a/src/test/regress/input/multi_alter_table_statements.source
+++ b/src/test/regress/input/multi_alter_table_statements.source
@@ -240,9 +240,36 @@ COMMIT;
 SELECT indexname, tablename FROM pg_indexes WHERE tablename = 'lineitem_alter';
 
 -- Create single-shard table (to avoid deadlocks in the upcoming test hackery)
-CREATE TABLE single_shard_items (id integer, name text);
+CREATE TABLE single_shard_items (id integer NOT NULL, name text);
 SELECT master_create_distributed_table('single_shard_items', 'id', 'hash');
 SELECT master_create_worker_shards('single_shard_items', 1, 2);
+
+-- Verify that ALTER TABLE .. REPLICATION IDENTITY [USING INDEX]* .. works
+CREATE UNIQUE INDEX replica_idx on single_shard_items(id);
+
+SELECT relreplident FROM pg_class WHERE relname = 'single_shard_items';
+SELECT run_command_on_workers('SELECT relreplident FROM pg_class WHERE relname LIKE ''single_shard_items_%'' LIMIT 1;');
+
+ALTER TABLE single_shard_items REPLICA IDENTITY nothing;
+SELECT relreplident FROM pg_class WHERE relname = 'single_shard_items';
+SELECT run_command_on_workers('SELECT relreplident FROM pg_class WHERE relname LIKE ''single_shard_items_%'' LIMIT 1;');
+
+ALTER TABLE single_shard_items REPLICA IDENTITY full;
+SELECT relreplident FROM pg_class WHERE relname = 'single_shard_items';
+SELECT run_command_on_workers('SELECT relreplident FROM pg_class WHERE relname LIKE ''single_shard_items_%'' LIMIT 1;');
+
+ALTER TABLE single_shard_items REPLICA IDENTITY USING INDEX replica_idx;
+SELECT relreplident FROM pg_class WHERE relname = 'single_shard_items';
+SELECT run_command_on_workers('SELECT relreplident FROM pg_class WHERE relname LIKE ''single_shard_items_%'' LIMIT 1;');
+
+ALTER TABLE single_shard_items REPLICA IDENTITY default, REPLICA IDENTITY USING INDEX replica_idx, REPLICA IDENTITY nothing;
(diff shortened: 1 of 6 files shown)


src/test/regress/output/multi_alter_table_statements.source
```diff
@@ -552,6 +552,95 @@ SELECT master_create_worker_shards('single_shard_items', 1, 2);
  
 (1 row)
 
+-- Verify that ALTER TABLE .. REPLICATION IDENTITY [USING INDEX]* .. works
```
onderkalaci: One more test could be creating a Postgres table, creating unique index, setting replica identity and then creating the distributed table to see that Citus honors the replica identity on distributed table creation.

src/test/regress/output/multi_alter_table_statements.source
```diff
@@ -552,6 +552,95 @@ SELECT master_create_worker_shards('single_shard_items', 1, 2);
...
+ALTER TABLE single_shard_items ALTER COLUMN id SET NOT NULL;
```
onderkalaci: Why do we execute `ALTER TABLE single_shard_items ALTER COLUMN id SET NOT NULL;` here?

src/test/regress/sql/multi_colocation_utils.sql
```diff
@@ -392,8 +392,8 @@ CREATE TABLE table2_group_none ( id int );
 SELECT create_distributed_table('table2_group_none', 'id', colocate_with => 'NONE');
 
 -- check metadata to see colocation groups are created successfully
-SELECT * FROM pg_dist_colocation 
-    WHERE colocationid >= 1 AND colocationid < 1000 
+SELECT * FROM pg_dist_colocation
```
onderkalaci: I wouldn't touch any unrelated white spaces. Could you undo unrelated changes?

src/test/regress/output/multi_alter_table_statements.source
```diff
@@ -552,6 +552,95 @@ SELECT master_create_worker_shards('single_shard_items', 1, 2);
...
+
+\c - - - :worker_1_port
+SELECT relreplident FROM pg_class WHERE relname = 'single_shard_items_220015';
+ relreplident 
+--------------
+ i
+(1 row)
+
+\c - - - :master_port
+ALTER TABLE single_shard_items REPLICA IDENTITY default, REPLICA IDENTITY USING INDEX replica_idx, REPLICA IDENTITY nothing;
```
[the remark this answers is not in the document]
furkansahin: I wanted to show when we query multiple replica identities, the last one is issued in postgres and the behaviour is the same in Citus. I'm also adding the one you suggested.

src/test/regress/output/multi_alter_table_statements.source
furkansahin: If the question is "Is it necessary the column to be not null" the answer is: Indexes cannot be used as replica identity if the column is nullable.
But if the question is "isn't it better we make the column not null when we are creating the table?", the answer is yes. You are right, I'm moving the definition to there.

src/test/regress/output/multi_alter_table_statements.source
onderkalaci: Let's leave this for #1733

src/test/regress/input/multi_alter_table_statements.source
```diff
@@ -240,10 +240,37 @@ COMMIT;
...
+
+ALTER TABLE single_shard_items REPLICA IDENTITY USING INDEX replica_idx;
+SELECT relreplident FROM pg_class WHERE relname = 'single_shard_items';
+SELECT run_command_on_workers('SELECT relreplident FROM pg_class WHERE relname LIKE ''single_shard_items_%'' LIMIT 1;');
+
+ALTER TABLE single_shard_items REPLICA IDENTITY default, REPLICA IDENTITY USING INDEX replica_idx, REPLICA IDENTITY nothing;
+SELECT relreplident FROM pg_class WHERE relname = 'single_shard_items';
+SELECT run_command_on_workers('SELECT relreplident FROM pg_class WHERE relname LIKE ''single_shard_items_%'' LIMIT 1;');
+
+ALTER TABLE single_shard_items ADD COLUMN test_col int, REPLICA IDENTITY full;
```
marcocitus: I kind of wonder if we should error or warn out here. It seems too tempting to actually use this.

jasonmp85: It's as bad an idea for us as it is for PostgreSQL, so I say let people shoot their feet if they want I guess?

onderkalaci: > It's as bad an idea for us as it is for PostgreSQL, so I say let people shoot their feet if they want I guess?

Yes, that's what I thought as well. Btw, if the `UPDATE` / `DELETE` rate is really low, this could be useful.

PR #1734, opened 2017-10-27
create_distributed_table minds the replica identity

### LINKED ISSUE #1733: create_distributed_table() should not ignore replica identity
Currently, when a local table has replica identity and then distributed, Citus ignores the replica identity. We should fix this, a preliminary work is [here](https://github.com/citusdata/citus/compare/create_table_replica_identity)

Sample steps:
```SQL
 CREATE TABLE test (a int PRIMARY KEY);
ALTER TABLE test REPLICA IDENTITY USING INDEX test_pkey ;
postgres=# \d test
                Table "public.test"
 Column |  Type   | Collation | Nullable | Default 
--------+---------+-----------+----------+---------
 a      | integer |           | not null | 
Indexes:
    "test_pkey" PRIMARY KEY, btree (a) REPLICA IDENTITY

 SELECT create_distributed_table('test', 'a');
 create_distributed_table 
--------------------------
 
(1 row)

postgres=# \c - - - 9700
You are now connected to database "postgres" as user "onderkalaci" via socket in "/tmp" at port "9700".
\d test_1022
test_102232 test_102235 test_102238 test_102241 test_102244 test_102247 test_102250 test_102253 test_102256 test_102259 test_102262
 \d test_102238
            Table "public.test_102238"
 Column |  Type   | Collation | Nullable | Default 
--------+---------+-----------+----------+---------
 a      | integer |           | not null | 
Indexes:
    "test_pkey_102238" PRIMARY KEY, btree (a)

```

furkansahin:
This PR makes changes in the master_node_protocol.c to get the ALTER TABLE .. REPLICA IDENTITY queries if the replica identity is defined.
fixes #1733

### DIFF
--- a/src/backend/distributed/master/master_node_protocol.c
+++ b/src/backend/distributed/master/master_node_protocol.c
@@ -69,9 +69,8 @@ int ShardReplicationFactor = 1; /* desired replication factor for shards */
 int ShardMaxSize = 1048576;     /* maximum size in KB one shard can grow to */
 int ShardPlacementPolicy = SHARD_PLACEMENT_ROUND_ROBIN;
 
-
+static List * GetTableReplicaIdentityCommand(Oid relationId);
 static Datum WorkerNodeGetDatum(WorkerNode *workerNode, TupleDesc tupleDescriptor);
-
 
 /* exports for SQL callable functions */
 PG_FUNCTION_INFO_V1(master_get_table_metadata);
@@ -474,6 +473,7 @@ GetTableDDLEvents(Oid relationId, bool includeSequenceDefaults)
 	List *tableDDLEventList = NIL;
 	List *tableCreationCommandList = NIL;
 	List *indexAndConstraintCommandList = NIL;
+	List *replicaIdentityEvents = NIL;
 
 	tableCreationCommandList = GetTableCreationCommands(relationId,
 														includeSequenceDefaults);
@@ -482,7 +482,42 @@ GetTableDDLEvents(Oid relationId, bool includeSequenceDefaults)
 	indexAndConstraintCommandList = GetTableIndexAndConstraintCommands(relationId);
 	tableDDLEventList = list_concat(tableDDLEventList, indexAndConstraintCommandList);
 
+	replicaIdentityEvents = GetTableReplicaIdentityCommand(relationId);
+	tableDDLEventList = list_concat(tableDDLEventList, replicaIdentityEvents);
+
 	return tableDDLEventList;
+}
+
+
+/*
+ * GetTableReplicaIdentityCommand returns the list of DDL commands to
+ * (re)define the replica identity choice for a given table.
+ */
+static List *
+GetTableReplicaIdentityCommand(Oid relationId)
+{
+	List *replicaIdentityCreateCommandList = NIL;
+	char *replicaIdentityCreateCommand = NULL;
+
+	/*
+	 * We skip non-relations because postgres does not support
+	 * ALTER TABLE .. REPLICA IDENTITY on non-relations.
+	 */
+	char relationKind = get_rel_relkind(relationId);
+	if (relationKind != RELKIND_RELATION)
+	{
+		return NIL;
+	}
+
+	replicaIdentityCreateCommand = pg_get_replica_identity_command(relationId);
+
+	if (replicaIdentityCreateCommand)
+	{
+		replicaIdentityCreateCommandList = lappend(replicaIdentityCreateCommandList,
+												   replicaIdentityCreateCommand);
+	}
+
+	return replicaIdentityCreateCommandList;
 }
 
 
--- a/src/backend/distributed/utils/citus_ruleutils.c
+++ b/src/backend/distributed/utils/citus_ruleutils.c
@@ -1066,3 +1066,49 @@ contain_nextval_expression_walker(Node *node, void *context)
 	}
 	return expression_tree_walker(node, contain_nextval_expression_walker, context);
 }
+
+
+/*
+ * pg_get_replica_identity_command function returns the required ALTER .. TABLE
+ * command to define the replica identity.
+ */
+char *
+pg_get_replica_identity_command(Oid tableRelationId)
+{
+	Relation relation = NULL;
+	StringInfo buf = makeStringInfo();
+	char *relationName = NULL;
+	char replicaIdentity = 0;
+
+	relation = heap_open(tableRelationId, AccessShareLock);
+
+	replicaIdentity = relation->rd_rel->relreplident;
+
+	relationName = generate_qualified_relation_name(tableRelationId);
+
+	if (replicaIdentity == REPLICA_IDENTITY_INDEX)
+	{
+		Oid indexId = RelationGetReplicaIndex(relation);
+
+		if (OidIsValid(indexId))
+		{
+			appendStringInfo(buf, "ALTER TABLE %s REPLICA IDENTITY USING INDEX %s ",
+							 relationName,
+							 quote_identifier(get_rel_name(indexId)));
+		}
+	}
+	else if (replicaIdentity == REPLICA_IDENTITY_NOTHING)
+	{
+		appendStringInfo(buf, "ALTER TABLE %s REPLICA IDENTITY NOTHING",
+						 relationName);
+	}
+	else if (replicaIdentity == REPLICA_IDENTITY_FULL)
+	{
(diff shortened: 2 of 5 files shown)


onderkalaci:
Could you please refer to the issue that this PR fixes?

src/backend/distributed/master/master_node_protocol.c
```diff
@@ -482,10 +485,128 @@ GetTableDDLEvents(Oid relationId, bool includeSequenceDefaults)
...
+ * to the replicaIdentityCreateCommandList.
+ */
+static void
+AppendReplicaIdentityCommands(List **replicaIdentityCreateCommandList,
+							  char replicaIdentity, Oid relationId)
+{
+	StringInfo buf = NULL;
+	char *relationName = NULL;
+
+	buf = makeStringInfo();
```
onderkalaci: Why don't you initialize at the declaration point?

src/test/regress/sql/multi_create_table.sql
```diff
@@ -516,7 +516,62 @@ SELECT create_reference_table('sc2.ref');
...
+COMMIT;
+SELECT run_command_on_workers($$SELECT relreplident FROM pg_class join information_schema.tables AS tables ON (pg_class.relname=tables.table_name) WHERE relname LIKE 'alter_replica_table_%' AND table_schema='sc6' LIMIT 1$$);
+
 DROP TABLE tt1;
 DROP TABLE tt2;
 DROP SCHEMA sc CASCADE;
 DROP SCHEMA sc2 CASCADE;
+DROP SCHEMA sc3 CASCADE;
+DROP SCHEMA sc4 CASCADE;
+DROP SCHEMA sc5 CASCADE;
```
onderkalaci: Don't forget to drop `sc6`

src/backend/distributed/master/master_node_protocol.c
```diff
@@ -482,10 +485,128 @@ GetTableDDLEvents(Oid relationId, bool includeSequenceDefaults)
...
+	{
+		return NIL;
+	}
+
+	/*
+	 * Set search_path to NIL so that all objects outside of pg_catalog will be
+	 * schema-prefixed. pg_catalog will be added automatically when we call
+	 * PushOverrideSearchPath(), since we set addCatalog to true;
+	 */
+	overridePath = GetOverrideSearchPath(CurrentMemoryContext);
```
onderkalaci: We'd only need this pattern when we deparse any query. So, it should go down to `AppendReplicaIdentityCommands ()` function.

src/backend/distributed/master/master_node_protocol.c
```diff
@@ -482,10 +485,128 @@ GetTableDDLEvents(Oid relationId, bool includeSequenceDefaults)
...
+ */
+static void
+AppendReplicaIdentityCommands(List **replicaIdentityCreateCommandList,
+							  char replicaIdentity, Oid relationId)
+{
+	StringInfo buf = NULL;
+	char *relationName = NULL;
+
+	buf = makeStringInfo();
+	relationName = generate_qualified_relation_name(relationId);
```
marcocitus: Might be worth splitting up this function into one that takes an OID and that takes a relation name. That way we could use this with shard names in the future.

src/backend/distributed/master/master_node_protocol.c
```diff
@@ -482,10 +485,128 @@ GetTableDDLEvents(Oid relationId, bool includeSequenceDefaults)
...
+	 * schema-prefixed. pg_catalog will be added automatically when we call
+	 * PushOverrideSearchPath(), since we set addCatalog to true;
+	 */
+	overridePath = GetOverrideSearchPath(CurrentMemoryContext);
+	overridePath->schemas = NIL;
+	overridePath->addCatalog = true;
+	PushOverrideSearchPath(overridePath);
+
+	/* open system catalog and find the table with the given relationId */
+	pgClass = heap_open(RelationRelationId, AccessShareLock);
```
marcocitus: I think you should be able to do something like:

```c
relation = heap_open(relationId, AccessShareLock);
replicaIdentity = rel->rd_rel->relreplident;
heap_close(relation, AccessShareLock);
```

src/backend/distributed/utils/citus_ruleutils.c
```diff
@@ -1066,3 +1066,73 @@ contain_nextval_expression_walker(Node *node, void *context)
 	}
 	return expression_tree_walker(node, contain_nextval_expression_walker, context);
 }
+
+
+/*
+ * pg_get_replica_identity_command function returns the required DDL command
```
onderkalaci: DDL -> Alter Table Command

src/backend/distributed/utils/citus_ruleutils.c
```diff
@@ -1066,3 +1066,73 @@ contain_nextval_expression_walker(Node *node, void *context)
...
+ * to define the replica identity.
+ */
+char *
+pg_get_replica_identity_command(Oid tableRelationId)
+{
+	Relation relation = NULL;
+	OverrideSearchPath *overridePath = NULL;
+	StringInfo buf = makeStringInfo();
+	char *relationName = NULL;
+	char replicaIdentity;
```
onderkalaci: Minor: Init the variable

src/backend/distributed/utils/citus_ruleutils.c
```diff
@@ -1066,3 +1066,50 @@ contain_nextval_expression_walker(Node *node, void *context)
...
+
+	relation = heap_open(tableRelationId, AccessShareLock);
+
+	replicaIdentity = relation->rd_rel->relreplident;
+
+	relationName = generate_qualified_relation_name(tableRelationId);
+
+	if (replicaIdentity == REPLICA_IDENTITY_INDEX)
+	{
+		Oid indexId = InvalidOid;
```
onderkalaci: What about `Oid indexId = RelationGetReplicaIndex(relation);` ?

src/test/regress/sql/multi_create_table.sql
```diff
@@ -516,7 +516,79 @@ SELECT create_reference_table('sc2.ref');
...
+
+BEGIN;
+CREATE SCHEMA sc4;
+CREATE TABLE sc4.alter_replica_table
+(
+	name text,
+	id int PRIMARY KEY
+);
+INSERT INTO sc4.alter_replica_table(id) SELECT generate_series(1,100);
+ALTER TABLE sc4.alter_replica_table REPLICA IDENTITY USING INDEX alter_replica_table_pkey;
```
onderkalaci: Could you slightly change this test to be:

```
SET search_path = 'sc4';
ALTER TABLE alter_replica_table REPLICA IDENTITY USING INDEX alter_replica_table_pkey;
SELECT create_distributed_table('alter_replica_table', 'id');
```

jasonmp85:
Don't we have some tests that _directly_ exercise the deparse logic? As in you call some test method with a DDL command and it prints the shard-extended version of that DDL? I think that would be the best at simply exercising this functionality.

onderkalaci:
> Don't we have some tests that directly exercise the deparse logic? As in you call some test method with a DDL command and it prints the shard-extended version of that DDL? I think that would be the best at simply exercising this functionality.

Do we have any such tests for `DDLs`? It'd be useful to use here as well.

jasonmp85:
> Do we have any such tests for `DDLs`? It'd be useful to use here as well.

They're in `multi_generate_ddl_commands.sql`; should be easy to add an example for this new syntax.

src/backend/distributed/master/master_node_protocol.c
```diff
@@ -69,10 +69,9 @@ int ShardReplicationFactor = 1; /* desired replication factor for shards */
 int ShardMaxSize = 1048576;     /* maximum size in KB one shard can grow to */
 int ShardPlacementPolicy = SHARD_PLACEMENT_ROUND_ROBIN;
 
-
+static List * GetTableReplicaIdentityCommand(Oid relationId);
```
onderkalaci: @hanefi realized that we're not setting the replica identities after shard moves. So, say a shard have a replica identity full, moving the shard leads to losing that information. 

That might lead to interesting bugs if a shard without a primary key is moved twice. In that case, during the second move, logical replication rejects INSERT/UPDATE/DELETEs, which is annoying.  

@hanefi is going to open an issue, I'm just wondering did we see a specific reason not to do it or just forgot. @furkansahin any memories?

PR #1736, opened 2017-10-27
 Make REPORTS_BASE_URL configurable and enable stats collection in tests. 

pykello:
Making `REPORTS_BASE_URL` configurable so we can avoid sending data to our main stats server, but be still able to run statistics collection to make sure it doesn't crash.

Also, enabling statistics collection in tests. After this change and the changes in #1726, tests pass locally.

### DIFF
--- a/configure
+++ b/configure
@@ -686,6 +686,7 @@ enable_option_checking
 with_extra_version
 enable_coverage
 with_libcurl
+with_reports_hostname
 '
       ac_precious_vars='build_alias
 host_alias
@@ -1316,6 +1317,9 @@ Optional Packages:
                           append STRING to version
   --without-libcurl       do not use libcurl for anonymous statistics
                           collection
+  --with-reports-hostname=HOSTNAME
+                          Use HOSTNAME as hostname for statistics collection
+                          and update checks
 
 Some influential environment variables:
   PG_CONFIG   Location to find pg_config for target PostgreSQL instalation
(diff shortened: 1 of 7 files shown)


configure
```diff
@@ -4112,6 +4116,36 @@ fi
 
 fi
 
+# REPORTS_BASE_URL definition
+
+
+
+# Check whether --with-reports-base-url was given.
+if test "${with_reports_base_url+set}" = set; then :
```
jasonmp85: `base_url` is sort of weird. Can we do `with_reports_hostname`?

configure
```diff
@@ -4112,6 +4116,36 @@ fi
...
+  withval=$with_reports_base_url;
+  case $withval in
+    yes)
+      as_fn_error $? "argument required for --with-reports-base-url option" "$LINENO" 5
+      ;;
+    no)
+      as_fn_error $? "argument required for --with-reports-base-url option" "$LINENO" 5
+      ;;
+    *)
+      REPORTS_BASE_URL="$withval"
```
jasonmp85: (if we do `with_reports_hostname` we'd change this to something like `"https://${withval}"`

configure
pykello: Done.

configure
pykello: Done.

pykello:
Merged via #1744.

PR #1739, opened 2017-10-30
Add constraint %s primary key using index %s

### LINKED ISSUE #1721: Implement `ALTER TABLE %s ADD CONSTRAINT %s PRIMARY KEY USING INDEX %s`
Reported by @marcocitus:

As we consider using logical replication on distributed tables, it'd be useful to set primary keys of distributed tables without blocking writes (or blocking very short time).

```SQL
CREATE UNIQUE INDEX CONCURRENTLY foo_pk_idx ON foo (x, y);
ALTER TABLE foo ADD CONSTRAINT foo_pk_idx PRIMARY KEY USING INDEX foo_pk_idx;
```

Unfortunately, this currently errors out on CItus with:
```SQL
ERROR:  index "foo_pk_idx" does not exist
```

Ideally, we'd also make this work:
```SQL
ALTER TABLE foo ADD PRIMARY KEY USING INDEX foo_pk_idx;
ERROR:  cannot create constraint without a name on a distributed table
```

furkansahin:
Makes changes on relay_event_utility.c to be able to detect the alter table .. add constraint .. using index statement. After detecting that query, it appends the shard_id to the end of the index name.
fixes #1721

### DIFF
--- a/src/backend/distributed/relay/relay_event_utility.c
+++ b/src/backend/distributed/relay/relay_event_utility.c
@@ -100,8 +100,19 @@ RelayEventExtendNames(Node *parseTree, char *schemaName, uint64 shardId)
 			{
 				AlterTableCmd *command = (AlterTableCmd *) lfirst(commandCell);
 
-				if (command->subtype == AT_AddConstraint ||
-					command->subtype == AT_DropConstraint)
+				if (command->subtype == AT_AddConstraint)
+				{
+					Constraint *constraint = (Constraint *) command->def;
+
+					if (constraint->contype == CONSTR_PRIMARY && constraint->indexname)
+					{
+						char **indexName = &(constraint->indexname);
+						AppendShardIdToName(indexName, shardId);
+					}
+
+					AppendShardIdToConstraintName(command, shardId);
+				}
+				if (command->subtype == AT_DropConstraint)
 				{
 					AppendShardIdToConstraintName(command, shardId);
 				}
--- a/src/test/regress/expected/multi_alter_table_add_constraints.out
+++ b/src/test/regress/expected/multi_alter_table_add_constraints.out
@@ -488,3 +488,131 @@ SELECT "Constraint", "Definition" FROM table_checks WHERE relid='public.products
 
 \c - - - :master_port
 DROP TABLE products;
+SET citus.shard_count to 2;
+-- Test if the ALTER TABLE %s ADD %s PRIMARY KEY %s works
+CREATE SCHEMA sc1;
+CREATE TABLE sc1.alter_add_prim_key(x int, y int);
+CREATE UNIQUE INDEX CONCURRENTLY alter_pk_idx ON sc1.alter_add_prim_key(x);
+ALTER TABLE sc1.alter_add_prim_key ADD CONSTRAINT alter_pk_idx PRIMARY KEY USING INDEX alter_pk_idx;
+SELECT create_distributed_table('sc1.alter_add_prim_key', 'x');
+ create_distributed_table 
+--------------------------
+ 
+(1 row)
+
+SELECT (run_command_on_workers($$
+    SELECT  
+        kc.constraint_name 
+    FROM 
+        information_schema.table_constraints tc join information_schema.key_column_usage kc on (kc.table_name = tc.table_name and kc.table_schema = tc.table_schema and kc.constraint_name = tc.constraint_name)
+    WHERE
+        kc.table_schema = 'sc1' and tc.constraint_type = 'PRIMARY KEY' and kc.table_name LIKE 'alter_add_prim_key_%' 
+    ORDER BY 
+    1
+    LIMIT 
+        1;
+    $$)).*
+ORDER BY 
+    1,2,3,4;
+ nodename  | nodeport | success |        result        
+-----------+----------+---------+----------------------
+ localhost |    57637 | t       | alter_pk_idx_1450234
+ localhost |    57638 | t       | alter_pk_idx_1450234
+(2 rows)
+
+CREATE SCHEMA sc2;
+CREATE TABLE sc2.alter_add_prim_key(x int, y int);
+SET search_path TO 'sc2';
+SELECT create_distributed_table('alter_add_prim_key', 'x');
+ create_distributed_table 
+--------------------------
+ 
+(1 row)
+
+CREATE UNIQUE INDEX CONCURRENTLY alter_pk_idx ON alter_add_prim_key(x);
+ALTER TABLE alter_add_prim_key ADD CONSTRAINT alter_pk_idx PRIMARY KEY USING INDEX alter_pk_idx;
+NOTICE:  using one-phase commit for distributed DDL commands
+HINT:  You can enable two-phase commit for extra safety with: SET citus.multi_shard_commit_protocol TO '2pc'
+SELECT (run_command_on_workers($$
+    SELECT  
+        kc.constraint_name 
+    FROM 
+        information_schema.table_constraints tc join information_schema.key_column_usage kc on (kc.table_name = tc.table_name and kc.table_schema = tc.table_schema and kc.constraint_name = tc.constraint_name)
+    WHERE
+        kc.table_schema = 'sc2' and tc.constraint_type = 'PRIMARY KEY' and kc.table_name LIKE 'alter_add_prim_key_%'
+    ORDER BY 
+    1
+    LIMIT 
+        1;
+    $$)).*
+ORDER BY 
+    1,2,3,4;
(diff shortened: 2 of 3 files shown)


src/backend/distributed/relay/relay_event_utility.c
```diff
@@ -100,8 +100,19 @@ RelayEventExtendNames(Node *parseTree, char *schemaName, uint64 shardId)
 			{
 				AlterTableCmd *command = (AlterTableCmd *) lfirst(commandCell);
 
-				if (command->subtype == AT_AddConstraint ||
-					command->subtype == AT_DropConstraint)
+				if (command->subtype == AT_AddConstraint)
+				{
+					Constraint *con = castNode(Constraint, command->def);
```
marcocitus: `con` -> `constraint`. 

For consistency with the code below, maybe use `(Constrant *) command->def;`

src/test/regress/expected/multi_alter_table_add_constraints.out
```diff
@@ -488,3 +488,93 @@ SELECT "Constraint", "Definition" FROM table_checks WHERE relid='public.products
 
 \c - - - :master_port
 DROP TABLE products;
+SET citus.shard_count to 2;
+-- Test if the ALTER TABLE %s ADD %s PRIMARY KEY %s works
+CREATE SCHEMA sc1;
```
onderkalaci: There is a behavior mentioned in Postgres docs which makes me think that requires further investigation.

`If a constraint name is provided then the index will be renamed to match the constraint name. Otherwise the constraint will be named the same as the index.`

This means that, if you provide and arbitrary name, you'd change the index names on the workers. See the following:

```SQL
CREATE SCHEMA sc2;
CREATE TABLE sc2.alter_add_prim_key(x int, y int);
CREATE UNIQUE INDEX CONCURRENTLY alter_pk_idx ON sc2.alter_add_prim_key(x);
SELECT create_distributed_table('sc2.alter_add_prim_key', 'x');

ALTER TABLE sc2.alter_add_prim_key ADD CONSTRAINT onder PRIMARY KEY USING INDEX alter_pk_idx;

\d sc2.alter_add_prim_key 
          Table "sc2.alter_add_prim_key"
 Column |  Type   | Collation | Nullable | Default 
--------+---------+-----------+----------+---------
 x      | integer |           | not null | 
 y      | integer |           |          | 
Indexes:
    "onder" PRIMARY KEY, btree (x)

postgres=# \c - - - 9700
You are now connected to database "postgres" as user "onderkalaci" via socket in "/tmp" at port "9700".

\d sc2.alter_add_prim_key_102113
       Table "sc2.alter_add_prim_key_102113"
 Column |  Type   | Collation | Nullable | Default 
--------+---------+-----------+----------+---------
 x      | integer |           | not null | 
 y      | integer |           |          | 
Indexes:
    "onder_102113" PRIMARY KEY, btree (x)

```

I haven't hit any issues, but, could you ensure that we're good even if the index name changes? Please check Citus regression tests and see if there are any edge cases that could lead to problems since the index name has changed. (And add a test here as well)

My second note would be on `ALTER TABLE ... DROP CONSTRAINT`. Could you both manually test and add some tests here that we're all good with `ALTER TABLE sc2.alter_add_prim_key DROP CONSTRAINT alter_pk_idx;`  (i.e., make sure that we're also good with changed index names etc.)

src/backend/distributed/relay/relay_event_utility.c
```diff
@@ -100,8 +100,19 @@ RelayEventExtendNames(Node *parseTree, char *schemaName, uint64 shardId)
...
+
+					if (con->contype == CONSTR_PRIMARY && con->indexname)
```
onderkalaci: What information `con->options` include? Do we ever need to check that? It is mentioned as /* options from WITH clause */ in the code. So, what is `WITH` clause?

onderkalaci:
> fixes #1721

Also, could you add a brief note to the PR along with the above?

src/test/regress/expected/multi_alter_table_add_constraints.out
```diff
@@ -488,3 +488,93 @@ SELECT "Constraint", "Definition" FROM table_checks WHERE relid='public.products
...
+        kc.constraint_name 
+    FROM 
+        information_schema.table_constraints tc join information_schema.key_column_usage kc on (kc.table_name = tc.table_name and kc.table_schema = tc.table_schema and kc.constraint_name = tc.constraint_name)
+    WHERE
+        kc.table_schema = 'sc2' and tc.constraint_type = 'PRIMARY KEY' and kc.table_name LIKE 'alter_add_prim_key_%' 
+    LIMIT 
+        1;
+    $$);
+          run_command_on_workers          
+------------------------------------------
```
onderkalaci: And, as the test failures show, you should add `ORDER BY` to the tests. The `ORDER BY` should be on the outer query, not the one that we sent to the workers.

src/backend/distributed/relay/relay_event_utility.c
furkansahin: `WITH` is used in postgres to define a couple of variables in indexes such as 
```SQL
CREATE UNIQUE INDEX u_ind3 on t(i) with (fillfactor=50);
```
means the 50% of the index page will be left empty. There are a couple of other options for different kind of indexes. 

When I run some queries such as;

onderkalaci: > So, there is a problem on this with Citus and there is no available tests for this right now.

Let's open an issue for that and refer to this discussion.

src/backend/distributed/relay/relay_event_utility.c
```diff
@@ -100,8 +100,19 @@ RelayEventExtendNames(Node *parseTree, char *schemaName, uint64 shardId)
 			{
 				AlterTableCmd *command = (AlterTableCmd *) lfirst(commandCell);
 
-				if (command->subtype == AT_AddConstraint ||
-					command->subtype == AT_DropConstraint)
+				if (command->subtype == AT_AddConstraint)
+				{
+					Constraint *con = (Constraint *) command->def;
```
onderkalaci: `con` -> `constraint`

src/test/regress/expected/multi_alter_table_add_constraints.out
```diff
@@ -488,3 +488,162 @@ SELECT "Constraint", "Definition" FROM table_checks WHERE relid='public.products
...
+        information_schema.table_constraints tc join information_schema.key_column_usage kc on (kc.table_name = tc.table_name and kc.table_schema = tc.table_schema and kc.constraint_name = tc.constraint_name)
+    WHERE
+        kc.table_schema = 'sc1' and tc.constraint_type = 'PRIMARY KEY' and kc.table_name LIKE 'alter_add_prim_key_%' 
+    ORDER BY
+        1
+    LIMIT 
+        1;
+    $$)
+ORDER BY 
+    1;
```
onderkalaci: `ORDER BY 1` doesn't help since all hosts are `localhost`. We should have `ORDER BY 1,2,3,4`

src/test/regress/expected/multi_alter_table_add_constraints.out
```diff
@@ -488,3 +488,162 @@ SELECT "Constraint", "Definition" FROM table_checks WHERE relid='public.products
...
+INSERT INTO sc4.alter_add_prim_key(x) SELECT generate_series(1,100);
+SET search_path TO 'sc4';
+SELECT create_distributed_table('alter_add_prim_key', 'x');
+NOTICE:  Copying data from local table...
+ create_distributed_table 
+--------------------------
+ 
+(1 row)
+
+CREATE UNIQUE INDEX CONCURRENTLY alter_pk_idx ON alter_add_prim_key(x);
```
onderkalaci: Let's add a single line comment why we're running `almost` the same test with the above.

src/test/regress/expected/multi_alter_table_add_constraints.out
```diff
@@ -488,3 +488,162 @@ SELECT "Constraint", "Definition" FROM table_checks WHERE relid='public.products
...
+    $$)
+ORDER BY 
+    1;
+          run_command_on_workers          
+------------------------------------------
+ (localhost,57637,t,alter_pk_idx_1450234)
+ (localhost,57638,t,alter_pk_idx_1450234)
+(2 rows)
+
+CREATE SCHEMA sc2;
```
onderkalaci: I think tests for `sc2` and `sc1` has no difference from Citus' perspective. Feel free to remove  `sc2` tests.

onderkalaci:
> Also, could you add a brief note of what this PR implements along with the above?

Ping on this one again

PR #1743, opened 2017-10-30
Add shard transfer mode parameter to shard copy functions

marcocitus:
Add a `shard_transfer_mode` parameter to `master_move_shard_placement` and `master_copy_shard_placement` to indicate the replication mode:

- `auto` (default) : require replica identity if logical replication is possible, otherwise use legacy behaviour (e.g. for shard repair, 9.6)
- `force_logical` : must use logical replication
- `block_writes`: must block writes


src/backend/distributed/citus--7.1-3--7.1-4.sql
```diff
@@ -0,0 +1,54 @@
...
+	source_node_port integer,
+	target_node_name text,
+	target_node_port integer,
+	do_repair bool DEFAULT true,
+	transfer_mode citus.shard_transfer_mode default 'require_replica_identity')
+RETURNS void
+LANGUAGE C STRICT
+AS 'citus', $$master_copy_shard_placement$$;
+
+COMMENT ON FUNCTION master_copy_shard_placement(shard_id bigint,
```
onderkalaci: I think my only comment would be that we're not supporting logical replication for shard repair functionality and actually ignoring the  `shard_transfer_mode ` flag. 

We could potentially error out on `master_copy_shard_placement()` when `doRepair = true` and `shard_transfer_mode != block_writes`. But, that'd break quite a few regression tests and potentially breaks the user experience. Still, it'd be useful to at least have some `LOG` messages for that.

What is your thought on that?

src/backend/distributed/citus--7.1-3--7.1-4.sql
```diff
@@ -0,0 +1,54 @@
...
+	shard_id bigint,
+	source_node_name text,
+	source_node_port integer,
+	target_node_name text,
+	target_node_port integer,
+	do_repair bool DEFAULT true,
+	transfer_mode citus.shard_transfer_mode default 'require_replica_identity')
+RETURNS void
+LANGUAGE C STRICT
+AS 'citus', $$master_copy_shard_placement$$;
```
onderkalaci: AS 'MODULE_PATHNAME' ?

src/backend/distributed/citus--7.1-3--7.1-4.sql
```diff
@@ -0,0 +1,54 @@
+/* citus--7.1-3--7.1-4 */
+
+CREATE TYPE citus.shard_transfer_mode AS ENUM (
+   'require_replica_identity',
+   'force_logical',
```
onderkalaci: I don't have any strong opinions, but, I was thinking that we'd use `force_logical_replication` which is more explicit. That said, feel free to ignore this comment.

src/backend/distributed/citus--7.1-3--7.1-4.sql
marcocitus: I've added an error when using `force_logical`. Maybe `require_replica_identity` is misnamed and should be something more generic, e.g. `auto`?

src/backend/distributed/master/master_repair_shards.c
```diff
@@ -83,6 +94,12 @@ master_copy_shard_placement(PG_FUNCTION_ARGS)
 							   "with do not repair functionality "
 							   "is only supported on Citus Enterprise")));
 	}
+	else if (shardReplicationMode == FORCE_LOGICAL)
```
onderkalaci: That seems a fair trade-off 👍

src/backend/distributed/master/master_repair_shards.c
```diff
@@ -109,6 +126,51 @@ master_move_shard_placement(PG_FUNCTION_ARGS)
...
+ */
+static char
+LookupShardReplicationMode(Oid shardReplicationModeOid)
+{
+	HeapTuple enumTuple = NULL;
+	Form_pg_enum enumForm = NULL;
+	char shardReplicationMode = 0;
+	const char *enumLabel = NULL;
+
+	enumTuple = SearchSysCache1(ENUMOID, ObjectIdGetDatum(shardReplicationModeOid));
```
onderkalaci: Minor: Might simplify the code if we use `enum_out` with the direct function call?

src/backend/distributed/citus--7.1-3--7.1-4.sql
onderkalaci: Actually, that makes sense to me. `auto` in the sense that it tries to pick the best alternative and errors under certain conditions where it cannot pick the best. In fact, `require_replica_identity` is hard to understand for users.

marcocitus: Alright, changed to `auto`

src/backend/distributed/citus--7.1-3--7.1-4.sql
marcocitus: I slightly prefer `force_logical` partly because I can't type words that long :).

src/test/regress/expected/multi_extension.out
```diff
@@ -126,6 +126,9 @@ ALTER EXTENSION citus UPDATE TO '7.0-13';
 ALTER EXTENSION citus UPDATE TO '7.0-14';
 ALTER EXTENSION citus UPDATE TO '7.0-15';
 ALTER EXTENSION citus UPDATE TO '7.1-1';
+ALTER EXTENSION citus UPDATE TO '7.1-2';
```
jasonmp85: Oh hey, apparently people were forgetting these…

I guess we need to figure out how to refactor the upgrade stuff as it's getting pretty crowded.

marcocitus: Agreed, the Makefile is also getting bizarre. Any ideas?

jasonmp85: For the Makefile, I suppose it wouldn't be tough to enumerate all the versions with a shell expansion/directory listing and create dynamic targets based on them; wasn't worth it initially but I think we passed the "worth it" stage long ago.

PR #1746, opened 2017-10-31
Only drop sequences on workers with metadata

### LINKED ISSUE #1269: DROP SEQUENCE deadlocks when using master as worker
Whenever you drop a sequence (inc. a table with a serial column), master_drop_sequences is called via the SQL_DROP trigger and sends DROP SEQUENCE IF EXISTS commands to workers *even if the sequence is not used for a distributed table*. This creates self-deadlocks when a node is its own worker.

```
SELECT master_add_node('localhost', 5432);
CREATE SEQUENCE foo;
DROP SEQUENCE foo;
... deadlock ...
```

The only resolution is to remove the node from pg_dist_node:

```
SELECT master_remove_node('localhost', 5432);
DROP SEQUENCE foo;
... no deadlock ...
```

marcocitus:
Not high priority, but I had some spare time while waiting for regression tests to finish and this has been bugging me for a while, since it kills off a legit way of using and testing Citus (coordinator = only worker). We only create sequences on workers for MX tables, but we try to drop them for all distributed tables, and this causes a deadlock when the coordinator is a worker because it tries to drop the sequence both locally and remotely.

Fixes #1269

### DIFF
--- a/src/backend/distributed/master/master_delete_protocol.c
+++ b/src/backend/distributed/master/master_delete_protocol.c
@@ -297,8 +297,8 @@ master_drop_sequences(PG_FUNCTION_ARGS)
 	{
 		appendStringInfoString(dropSeqCommand, " CASCADE");
 
-		SendCommandToWorkers(ALL_WORKERS, DISABLE_DDL_PROPAGATION);
-		SendCommandToWorkers(ALL_WORKERS, dropSeqCommand->data);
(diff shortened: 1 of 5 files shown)


mtuncer:
verified that it fixes the original issue.

mtuncer:
rebase  and 🚢

PR #1749, opened 2017-10-31
Don't try to add restrictions for reference tables in insert into select

### LINKED ISSUE #1688: Distributed INSERT...SELECT between reference tables crashes when used with stable cast
When a distributed `INSERT...SELECT` contains a cast that is implemented using a stable (rather than immutable) function such as `date_in`, we go through function evaluation and rebuild the query string. Unfortunately, `RebuildQueryStrings` blindly calls `AddShardIntervalRestrictionToSelect`, which crashes when used with reference tables.

hence:
```
CREATE TABLE ref1 (d timestamptz);
SELECT create_reference_table('ref1');
CREATE TABLE ref2 (d date);
SELECT create_reference_table('ref2');
INSERT INTO ref1 SELECT * FROM ref2;
```

crashes with:
```
TRAP: FailedAssertion("!(targetPartitionColumnVar != ((void*)0))", File: "planner/multi_router_planner.c", Line: 360)
```

metdos:
Fixes #1688

### DIFF
--- a/src/test/regress/expected/multi_insert_select.out
+++ b/src/test/regress/expected/multi_insert_select.out
@@ -2392,6 +2392,39 @@ SELECT * FROM ref_table ORDER BY user_id, value_1;
 (5 rows)
 
 DROP TABLE ref_table;
+-- Select from reference table into reference table
+CREATE TABLE ref1 (d timestamptz);
+SELECT create_reference_table('ref1');
+ create_reference_table 
+------------------------
+ 
+(1 row)
+
+CREATE TABLE ref2 (d date);
+SELECT create_reference_table('ref2');
+ create_reference_table 
+------------------------
+ 
+(1 row)
+
+INSERT INTO ref2 VALUES ('2017-10-31');
+INSERT INTO ref1 SELECT * FROM ref2;
+SELECT count(*) from ref1;
+ count 
+-------
+     1
+(1 row)
+
+-- also test with now()
+INSERT INTO ref1 SELECT now() FROM ref2;
+SELECT count(*) from ref1;
+ count 
+-------
+     2
+(1 row)
+
+DROP TABLE ref1;
+DROP TABLE ref2;
 -- Select into an append-partitioned table is not supported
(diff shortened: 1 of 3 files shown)


src/test/regress/expected/multi_insert_select.out
```diff
@@ -2392,6 +2392,47 @@ SELECT * FROM ref_table ORDER BY user_id, value_1;
...
+---------+---------
+       1 |        
+       2 |        
+       3 |        
+       4 |        
+       5 |        
+(5 rows)
+
+INSERT INTO ref_table_destination
+SELECT user_id, value_1 FROM ref_table_source;
```
marcocitus: Fix looks good, but the tests don't exercise the logic. This issue only happens when go through  function evaluation (e.g. there is a `now()` in the query, or a cast that is implemented as a stable function - see #1688)

metdos: oh right, I tempted to use existing tables in regressions. I updated the regressions.

src/test/regress/expected/multi_insert_select.out
```diff
@@ -2392,6 +2392,39 @@ SELECT * FROM ref_table ORDER BY user_id, value_1;
...
+INSERT INTO ref2 VALUES ('2017-10-31');
+INSERT INTO ref1 SELECT * FROM ref2;
+SELECT count(*) from ref1;
+ count 
+-------
+     1
+(1 row)
+
+-- also test with now()
+INSERT INTO ref1 SELECT now();
```
marcocitus: this probably goes through INSERT...SELECT on the coordinator since it is not selecting from a distributed or reference table.

metdos: hmm, that's right. Should I just drop this now() test or add as a second column?

marcocitus: something like `SELECT now() FROM ref2` should work

metdos: Done

PR #1750, opened 2017-10-31
Set shard and placement IDs in regression tests using a GUC

marcocitus:
Currently, the biggest stumbling block for parallelising regression tests is the fact that many test outputs contain shard IDs and shard IDs are generated using a sequence. If tests run in parallel the order in which they generate shard IDs is non-deterministic.

This PR adds GUCs for generating shard IDs and placement IDs that are local to the session, by incrementing the GUC from its original value. If the GUC is 0 (the default) the sequence is used.

Once we've merged this we can run many regression tests in parallel.


src/backend/distributed/master/master_node_protocol.c
```diff
@@ -291,14 +292,26 @@ master_get_new_shardid(PG_FUNCTION_ARGS)
...
-	Datum sequenceIdDatum = ObjectIdGetDatum(sequenceId);
+	text *sequenceName = NULL;
+	Oid sequenceId = InvalidOid;
+	Datum sequenceIdDatum = NULL;
 	Oid savedUserId = InvalidOid;
 	int savedSecurityContext = 0;
 	Datum shardIdDatum = 0;
 	uint64 shardId = 0;
 
+	if (NextShardId > 0)
```
metdos: Maybe we can add a comment similar to `next_shard_id`.

src/backend/distributed/shared_library_init.c
```diff
@@ -814,6 +814,36 @@ RegisterCitusConfigVariables(void)
...
+		0, 0, INT_MAX,
+		PGC_USERSET,
+		GUC_NO_SHOW_ALL,
+		NULL, NULL, NULL);
+
+	DefineCustomIntVariable(
+		"citus.next_placement_id",
+		gettext_noop("Set the next placement ID to use in placement creation."),
+		gettext_noop("Placement IDs are normally generated using a sequence. If "
+					 "next_placement _id is set to a non-zero value, placement IDs will "
```
metdos: typo in next_placement _id (space)

src/backend/distributed/master/master_node_protocol.c
```diff
@@ -351,14 +365,26 @@ master_get_new_placementid(PG_FUNCTION_ARGS)
...
-	Datum sequenceIdDatum = ObjectIdGetDatum(sequenceId);
+	text *sequenceName = NULL;
+	Oid sequenceId = InvalidOid;
+	Datum sequenceIdDatum = 0;
 	Oid savedUserId = InvalidOid;
 	int savedSecurityContext = 0;
 	Datum placementIdDatum = 0;
 	uint64 placementId = 0;
 
+	if (NextPlacementId > 0)
```
metdos: similarly maybe add a comment

src/test/regress/expected/multi_colocation_utils.out
```diff
@@ -716,93 +716,129 @@ ORDER BY
...
- table1_group_default             | 1300066 | t            |    57638 | -715827883    | 715827881
- table1_group_default             | 1300066 | t            |    57637 | -715827883    | 715827881
- table1_group_default             | 1300067 | t            |    57637 | 715827882     | 2147483647
- table1_group_default             | 1300067 | t            |    57638 | 715827882     | 2147483647
- table1_groupf                    | 1300068 | t            |    57637 |               | 
- table1_groupf                    | 1300068 | t            |    57638 |               | 
- table2_groupf                    | 1300069 | t            |    57637 |               | 
- table2_groupf                    | 1300069 | t            |    57638 |               | 
-(84 rows)
+                     logicalrelid                      | shardid | shardstorage | nodeport | shardminvalue | shardmaxvalue
```
metdos: Why didn't we set GUCs here?

src/backend/distributed/master/master_node_protocol.c
```diff
@@ -291,14 +293,26 @@ master_get_new_shardid(PG_FUNCTION_ARGS)
 uint64
 GetNextShardId()
 {
-	text *sequenceName = cstring_to_text(SHARDID_SEQUENCE_NAME);
-	Oid sequenceId = ResolveRelationId(sequenceName);
-	Datum sequenceIdDatum = ObjectIdGetDatum(sequenceId);
+	text *sequenceName = NULL;
+	Oid sequenceId = InvalidOid;
+	Datum sequenceIdDatum = NULL;
```
furkansahin: We should do `Datum sequenceIdDatum = 0;`. Otherwise, it gives a warning.

metdos:
@marcocitus, I didn't notice that you updated the PR. 🚢

src/test/regress/expected/multi_colocation_utils.out
marcocitus: Missed this one. I did skip a few because it would be too inconvenient to set the GUC (e.g. they reconnect very often). Those should remain sequential for the moment.

PR #1759, opened 2017-11-02
Allow distributed partitioned table creation in Cloud

### LINKED ISSUE #1758: Distributing a partitioned table errors out
Citus MX currently doesn't support table partitioning. We disabled partitioning for MX because of the complexity associated with running `DROP TABLE` on distributed and partitioned tables.

However, the error check that we put in for Citus MX also kicks in when we use the streaming replication. So this error check forbids partitioning a distributed table on Cloud.

```
SELECT create_distributed_table('measurement', 'city_id');
ERROR:  distributing partitioned tables which uses streaming replication is not supported
```

Should we quickly fix this for 7.1?

(also, typo in the error message)

marcocitus:
We currently error out when distributing a partitioned table when `replication_model = 'streaming'`, which is the default in Citus Cloud. The reason for this check appears to be that the MX logic for partitioned tables is not implemented. This PR makes sure we only error out in an actual MX cluster.

Fixes #1758

### DIFF
--- a/src/backend/distributed/commands/create_distributed_table.c
+++ b/src/backend/distributed/commands/create_distributed_table.c
@@ -728,11 +728,12 @@ EnsureRelationCanBeDistributed(Oid relationId, Var *distributionColumn,
 		}
 
 		/* we currently don't support MX tables to be distributed partitioned table */
-		if (replicationModel == REPLICATION_MODEL_STREAMING)
+		if (replicationModel == REPLICATION_MODEL_STREAMING &&
+			CountPrimariesWithMetadata() > 0)
 		{
 			ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
-							errmsg("distributing partitioned tables which uses "
-								   "streaming replication is not supported")));
+							errmsg("distributing partitioned tables is not supported "
+								   "with Citus MX")));
 		}
 
 		/* we don't support distributing tables with multi-level partitioning */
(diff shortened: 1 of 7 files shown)


src/backend/distributed/commands/create_distributed_table.c
```diff
@@ -728,11 +728,12 @@ EnsureRelationCanBeDistributed(Oid relationId, Var *distributionColumn,
 		}
 
 		/* we currently don't support MX tables to be distributed partitioned table */
-		if (replicationModel == REPLICATION_MODEL_STREAMING)
+		if (replicationModel == REPLICATION_MODEL_STREAMING &&
+			CountPrimariesWithMetadata() > 0)
```
onderkalaci: I'm not suggesting we do it here (or that's the best API), but to understand the function, can we write the following:

```python
IsMxCluster():
    if CountPrimariesWithMetadata() >0:
       return true
   return false
```

byucesoy:
PR looks good, tested and it works as expected. Apart from @onderkalaci's comment, I do not have anything to add. :ship:

src/backend/distributed/commands/create_distributed_table.c
marcocitus: It kind of makes sense, but I would expect `IsMxCluster` to tell me whether I'm in an MX cluster on the worker nodes as well, but it won't, since the hasmetadata flag is not reliably set on the workers.

PR #1761, opened 2017-11-06
Fix intermittent test failures on count distinct tests

mtuncer:
Added analyze to test which seem to force planner to use index


metdos:
looks good 🚢

PR #1765, opened 2017-11-07
Add check for connection failures during multishard update

### LINKED ISSUE #1763: Coordinator crashes during multi shard update when a worker node fails
This is a crash due to assert.
- run a long update or insert into select query  (tested with a 32 shard table with 12M rows)
- kill one of the workers 
- coordinator crashes with and assert
```
TRAP: FailedAssertion("!(queryOK)", File: "executor/multi_router_executor.c", Line: 1145)
```

mtuncer:
multi shard executor expected a worker query to return a data. If the worker query fails, coordinator crashes.

This work introduces additional check to see if the worker process completed, and reports connection error if an issue is encountered.

Fixes #1763

### DIFF
--- a/src/backend/distributed/executor/multi_router_executor.c
+++ b/src/backend/distributed/executor/multi_router_executor.c
@@ -1141,8 +1141,11 @@ ExecuteModifyTasks(List *taskList, bool expectResults, ParamListInfo paramListIn
 											 &currentAffectedTupleCount);
 			}
 
-			/* should have rolled back on error */
-			Assert(queryOK);
+			/* We error out if the worker fails to return a result for the query. */
+			if (!queryOK)
+			{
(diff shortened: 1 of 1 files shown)


src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -1141,8 +1141,10 @@ ExecuteModifyTasks(List *taskList, bool expectResults, ParamListInfo paramListIn
 											 &currentAffectedTupleCount);
 			}
 
-			/* should have rolled back on error */
-			Assert(queryOK);
+			if (!queryOK)
+			{
```
marcocitus: Could use a small comment saying that we error out if a query fails, but return false if the connection fails.

PR #1769, opened 2017-11-08
Update configure file

byucesoy:
If autoconf version is >= 2.70, it automatically adds runstatedir but for
older versions, autoconf does not add runstatedir option. This causes
flapping this option in and out depending on who runs autoconf. With this
commit, we stabilize autoconf output.

### DIFF
--- a/configure
+++ b/configure
@@ -661,6 +661,7 @@ infodir
 docdir
 oldincludedir
 includedir
+runstatedir
 localstatedir
 sharedstatedir
 sysconfdir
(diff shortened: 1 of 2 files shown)


jasonmp85:
Hey @byucesoy this looks fine, but I'm not really sure what it's for. Can you explain the motivation here? My copy of `autoconf` or whatever doesn't appear to generate this line, which is why it's been flapping in and out of this file, depending upon who runs `autoconf`.

byucesoy:
@jasonmp85 it's been a while but I find some time to look into this problem and find the true cause (and proper fix). See my edited commit message for details. Can you verify that your setup now also generates those lines?

jasonmp85:
Yeah, it looks like I have autoconf 2.69, the highest available from Homebrew.

byucesoy:
So this is OK to :ship: ?

jasonmp85:
Sorry lost track of it when I switch tasks. Lemme do a quick local check.

byucesoy:
Closing this PR. We had discussion with @jasonmp85 and decided to keep things as it is for now.

lithp:
It didn't work out but thanks for doing this @byucesoy! This has bothered me too.

PR #1777, opened 2017-11-09
Skip page-level locks for distributed deadlock detection

onderkalaci:
Fixes another piece of #1655. 

In general, this issue becomes apparent with many concurrent 
modifications when large GIN indexes exist on the distributed tables.

Skip page-level locks on distributed deadlock detection
    
Short-term share/exclusive page-level locks are used for
read/write access. Locks are released immediately after
each index row is fetched or inserted.
    
Since those locks may not lead to any deadlocks, it's safe
to ignore them in the distributed deadlock detection.

If you want the steps to reproduce the issue, you can reach
out to me.


onderkalaci:
> Might be worth checking whether there are any other LOCKTAGs we should ignore.

Seems like this is the last one. I'll double check.

PR #1781, opened 2017-11-09
Really run CLUSTER commands on distributed tables

### LINKED ISSUE #1752: Push down CLUSTER command on distributed tables
Citus' behavior with `CLUSTER` command is confusing. We don't error out but don't push it down, which is confusing since `CLUSTER` is meaningful if there is data on the node.

It appears that we [have the logic](https://github.com/citusdata/citus/blob/master/src/backend/distributed/relay/relay_event_utility.c#L140) for pushing it down .

metdos:
Fixes #1752


metdos:
Closing this given that we passed the code freeze, will add another PR which adds a WARNING for the time being.

PR #1782, opened 2017-11-10
Warn on CLUSTER command for distributed tables

### LINKED ISSUE #1752: Push down CLUSTER command on distributed tables

metdos:
Related to #1752

### DIFF
--- a/src/backend/distributed/executor/multi_utility.c
+++ b/src/backend/distributed/executor/multi_utility.c
@@ -568,6 +568,37 @@ multi_ProcessUtility(PlannedStmt *pstmt,
 		ProcessVacuumStmt(vacuumStmt, queryString);
 	}
 
+	/* warn for CLUSTER command on distributed tables */
+	if (IsA(parsetree, ClusterStmt))
+	{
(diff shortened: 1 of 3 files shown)


src/backend/distributed/executor/multi_utility.c
```diff
@@ -568,6 +568,37 @@ multi_ProcessUtility(PlannedStmt *pstmt,
...
+	/* warn for CLUSTER commands on distributed tables */
+	if (IsA(parsetree, ClusterStmt))
+	{
+		ClusterStmt *clusterStmt = (ClusterStmt *) parsetree;
+		bool clusterDistributedTable = false;
+
+		if (clusterStmt->relation == NULL)
+		{
+			/* CLUSTER all */
+			clusterDistributedTable = true;
```
marcocitus: maybe `showPropagationWarning`

PR #1785, opened 2017-11-10
Make real-time executor work in transactions (and fix pg_partman)

marcocitus:
The real-time executor currently always opens new connections to SELECT from shards, which means it cannot see uncommitted writes and thus real-time SELECT after write errors out. It also does not register which shard placement it SELECTs from, which means DDL after real-time SELECT errors out.

This PR makes the real-time executor open its connections through `StartPlacementListConnection` to make sure it can be part of a transaction. For example, the following now works (recently came up in Slack):

```
postgres=# BEGIN;
BEGIN
Time: 0.165 ms
postgres=# SELECT count(*) FROM facilities;
 count 
-------
     0
(1 row)

Time: 70.727 ms
postgres=# INSERT INTO facilities (name) VALUES ('new-facility');
INSERT 0 1
Time: 1.436 ms
postgres=# SELECT count(*) FROM facilities;
 count 
-------
     1
(1 row)

Time: 7.623 ms
postgres=# END;
COMMIT
```

Importantly, it also enables most pg_partman functionality.

This mostly needs a lot of regression tests and some failure tests.

### DIFF
--- a/src/backend/distributed/commands/multi_copy.c
+++ b/src/backend/distributed/commands/multi_copy.c
@@ -821,9 +821,18 @@ OpenCopyConnections(CopyStmt *copyStatement, ShardConnections *shardConnections,
 		ShardPlacement *placement = (ShardPlacement *) lfirst(placementCell);
 		char *nodeUser = CurrentUserName();
 		MultiConnection *connection = NULL;
-		uint32 connectionFlags = FOR_DML | CONNECTION_PER_PLACEMENT;
+		uint32 connectionFlags = FOR_DML;
 		StringInfo copyCommand = NULL;
 		PGresult *result = NULL;
+
+		/*
+		 * Make sure we use a separate connection per placement for hash-distributed
+		 * tables in order to allow multi-shard modifications in the same transaction.
+		 */
+		if (placement->partitionMethod == DISTRIBUTE_BY_HASH)
+		{
+			connectionFlags |= CONNECTION_PER_PLACEMENT;
+		}
 
 		connection = GetPlacementConnection(connectionFlags, placement, nodeUser);
 
--- a/src/backend/distributed/executor/multi_client_executor.c
+++ b/src/backend/distributed/executor/multi_client_executor.c
@@ -22,6 +22,7 @@
 #include "distributed/connection_management.h"
 #include "distributed/multi_client_executor.h"
 #include "distributed/multi_server_executor.h"
+#include "distributed/placement_connection.h"
 #include "distributed/remote_commands.h"
 
 #include <errno.h>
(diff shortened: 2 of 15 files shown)


lfittl:
For what its worth I think this will also fix some rare issues we see in multi-tenant use cases where a cross-shard query happens in a complex transaction.

src/test/regress/sql/multi_real_time_transaction.sql
```diff
@@ -0,0 +1,147 @@
+SET citus.next_shard_id TO 1610000;
+ALTER SEQUENCE pg_catalog.pg_dist_jobid_seq RESTART 1610000;
```
marcocitus: This sequence is never used by Citus.

src/test/regress/sql/multi_real_time_transaction.sql
```diff
@@ -0,0 +1,147 @@
...
+
+SET citus.shard_count TO 4;
```
marcocitus: This is already the default shard count

src/test/regress/sql/multi_real_time_transaction.sql
```diff
@@ -0,0 +1,147 @@
+SET citus.next_shard_id TO 1610000;
```
marcocitus: It might make sense to do a 
```
CREATE SCHEMA multi_real_time_transaction;
SET search_path TO 'multi_real_time_transaction';
...
DROP SCHEMA multi_real_time_transaction
```

src/backend/distributed/executor/multi_real_time_executor.c
```diff
@@ -348,13 +352,24 @@ ManageTaskExecution(Task *task, TaskExecution *taskExecution,
...
 			 * On task failure, we close the connection. We also reset our execution
 			 * status assuming that we might fail on all other worker nodes and come
 			 * back to this failed node. In that case, we will retry the same fetch
 			 * and compute task(s) on this node again.
 			 */
 			int32 connectionId = connectionIdArray[currentIndex];
+			MultiConnection *connection = MultiClientGetConnection(connectionId);
+
+			/*
+			 * If this function was previously marked as critical (e.g. it was used
```
marcocitus: made a typo: function - > connection

src/backend/distributed/executor/multi_client_executor.c
```diff
@@ -175,6 +176,55 @@ MultiClientConnectStart(const char *nodeName, uint32 nodePort, const char *nodeD
...
+ * the connection id. Otherwise, it reports connectio error and returns
+ * INVALID_CONNECTION_ID.
+ */
+int32
+MultiClientPlacementConnectStart(List *placementAccessList, const char *userName)
+{
+	MultiConnection *connection = NULL;
+	ConnStatusType connStatusType = CONNECTION_OK;
+	int32 connectionId = AllocateConnectionId();
+	int connectionFlags = CONNECTION_PER_PLACEMENT; /* no cached connections for now */
```
marcocitus: We probably don't want to set the `CONNECTION_PER_PLACEMENT` if we're not yet in a coordinated transaction. Otherwise it would prevent connection reuse.

src/backend/distributed/executor/multi_client_executor.c
```diff
@@ -247,6 +305,31 @@ MultiClientDisconnect(int32 connectionId)
...
+MultiClientReleaseConnection(int32 connectionId)
+{
+	MultiConnection *connection = NULL;
+	const int InvalidPollingStatus = -1;
+
+	Assert(connectionId != INVALID_CONNECTION_ID);
+	connection = ClientConnectionArray[connectionId];
+	Assert(connection != NULL);
+
+	UnclaimConnection(connection);
```
marcocitus: We could consider closing the connection here if we're not (yet) in a coordinated transaction to keep the current behaviour.

marcocitus: I think it would be better to close here when not in a coordinated transaction.

src/backend/distributed/commands/multi_copy.c
```diff
@@ -821,10 +821,15 @@ OpenCopyConnections(CopyStmt *copyStatement, ShardConnections *shardConnections,
...
 		char *nodeUser = CurrentUserName();
 		MultiConnection *connection = NULL;
-		uint32 connectionFlags = FOR_DML | CONNECTION_PER_PLACEMENT;
+		uint32 connectionFlags = FOR_DML;
 		StringInfo copyCommand = NULL;
 		PGresult *result = NULL;
 
+		if(placement->partitionMethod == DISTRIBUTE_BY_HASH)
+		{
+			connectionFlags |= CONNECTION_PER_PLACEMENT;
```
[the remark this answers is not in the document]
velioglu: Handling range distributed tables like append distributed tables looks logical to me.

src/backend/distributed/executor/multi_client_executor.c
```diff
@@ -323,7 +323,14 @@ MultiClientReleaseConnection(int32 connectionId)
 	connection = ClientConnectionArray[connectionId];
 	Assert(connection != NULL);
 
-	UnclaimConnection(connection);
+	if (!InCoordinatedTransaction())
+	{
+		MultiClientDisconnect(connectionId);
+	}
+	else
```
marcocitus: This could use a comment about the reasoning behind it.

src/backend/distributed/executor/multi_real_time_executor.c
```diff
@@ -88,6 +89,7 @@ MultiRealTimeExecute(Job *job)
 
 	workerNodeList = ActiveReadableNodeList();
 	workerHash = WorkerHash(workerHashName, workerNodeList);
+	BeginOrContinueCoordinatedTransaction();
```
marcocitus: Why do we need this?

src/backend/distributed/commands/multi_copy.c
```diff
@@ -821,10 +821,15 @@ OpenCopyConnections(CopyStmt *copyStatement, ShardConnections *shardConnections,
 		ShardPlacement *placement = (ShardPlacement *) lfirst(placementCell);
 		char *nodeUser = CurrentUserName();
 		MultiConnection *connection = NULL;
-		uint32 connectionFlags = FOR_DML | CONNECTION_PER_PLACEMENT;
+		uint32 connectionFlags = FOR_DML;
 		StringInfo copyCommand = NULL;
 		PGresult *result = NULL;
 
+		if (placement->partitionMethod == DISTRIBUTE_BY_HASH)
```
marcocitus: This could use a comment.

```
/*
 * Make sure we use a separate connection per placement for hash-distributed
 * tables in order to allow multi-shard modifications in the same transaction.
 */
```

PR #1790, opened 2017-11-13
Add stub UDFs to run pg_upgrade flawlessly

### LINKED ISSUE #1788: pg_upgrade fails on clusters that ran  6.1 or below
Seems at some point the function `master_get_round_robin_candidate_nodes` was removed. The only extension control file to reflect this is `citus--6.1-17--6.2-1.sql`.

However, it fails to drop the function, and succeeds, because it is written like this:

```sql
DROP FUNCTION IF EXISTS master_get_round_robin_candidate_nodes();
```

This statement succeeds, because there is no such function signature, nor has there ever been. What *does* exist is the function with the signature `master_get_round_robin_candidate_nodes(bigint)`, which lives on after the control file has run.

To confirm this, on a 7.0 installation, if one tries to call this function, predictably, it fails:

```
ERROR:  could not find function "master_get_round_robin_candidate_nodes" in file "/usr/pgsql-9.6/lib/citus.so"
```

That's fine, as it is dead code...until you try to pg_upgrade.

pg_upgrade attempts to resolve every symbol of LANGUAGE C functions in their corresponding .sos. And that fails, here, and the result is pg_upgrade exits with a failure.

Suggested actions:

* throw in a stub symbol/function into new point releases of the .sos that simply satisfies pg_upgrade, and otherwise does nothing.

* drop the function correctly in migrations to new versions so the stub can eventually go away.

velioglu:
Fixes #1788.

Add stub UDF implementation to run pg_upgrade flawlessly.

### DIFF
--- a/src/backend/distributed/master/master_node_protocol.c
+++ b/src/backend/distributed/master/master_node_protocol.c
@@ -78,6 +78,9 @@ PG_FUNCTION_INFO_V1(master_get_table_ddl_events);
 PG_FUNCTION_INFO_V1(master_get_new_shardid);
 PG_FUNCTION_INFO_V1(master_get_new_placementid);
 PG_FUNCTION_INFO_V1(master_get_active_worker_nodes);
+PG_FUNCTION_INFO_V1(master_get_round_robin_candidate_nodes);
+PG_FUNCTION_INFO_V1(master_stage_shard_row);
+PG_FUNCTION_INFO_V1(master_stage_shard_placement_row);
 
 
 /*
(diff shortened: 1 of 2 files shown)


velioglu:
@marcocitus Should we give a warning if user calls this UDF ?

marcocitus:
> @marcocitus Should we give a warning if user calls this UDF ?

I think an error would make sense.

src/backend/distributed/master/master_node_protocol.c
```diff
@@ -371,6 +374,57 @@ GetNextPlacementId(void)
...
+
+/*
+ * master_stage_shard_placement_row is a stub UDF to make pg_upgrade
+ * work flawlessly while upgrading servers from 6.1. This implementation
+ * will be removed after the UDF dropped on the sql side properly.
+ */
+Datum
+master_stage_shard_placement_row(PG_FUNCTION_ARGS)
+{
+	Datum stubDatum = 0;
```
marcocitus: These variables, as well as the return, seem redundant.

metdos:
I and @velioglu tested 9.6/6.1 -> 9.6/7.1 -> 10/7.1 path on different machines.

jasonmp85:
Um… shouldn't this come with a `sql` file to actually drop the functions with the syntax they actually need to be dropped? We'll probably want to use `DROP IF EXISTS`, since people who never had a pre-`6.2` version will not have the functions at all.

metdos:
> Um… shouldn't this come with a sql file to actually drop the functions with the syntax they actually need to be dropped? We'll probably want to use DROP IF EXISTS, since people who never had a pre-6.2 version will not have the functions at all.

Added an issue #1800

jasonmp85:
> Added an issue #1800

I really think this should have gone in `7.1.0`; our policy of not having SQL migrations means it won't be fixed until `7.2.0`…

metdos:
> I really think this should have gone in 7.1.0; our policy of not having SQL migrations means it won't be fixed until 7.2.0…

As not fixed, you mean we will have stub functions, right? I think it's a bit late for 7.1.0.

PR #1793, opened 2017-11-13
Rename CreateDirectory to CitusCreateDirectory

lithp:


src/backend/distributed/worker/worker_partition_protocol.c
```diff
@@ -602,7 +602,7 @@ DirectoryExists(StringInfo directoryName)
 
 /* Creates a new directory with the given directory name. */
 void
-CreateDirectory(StringInfo directoryName)
+CitusCreateDirectory(StringInfo directoryName)
```
mtuncer: Are we going to put ```Citus``` prefix to our general purpose functions ? 

I suggest we have a common consensus on this one as we might do this multiple times as we go.

jasonmp85: FWIW I was always in favor of using visibility modifiers to whitelist symbols that we actually need to be visible.

mtuncer:
@ozgune any opinion on this ?

src/backend/distributed/worker/worker_partition_protocol.c
marcocitus: > Are we going to put Citus prefix to our general purpose functions ?

Seems reasonable when there is a name collision. We did this for a long time with, e.g. `CitusNodeToString`.

ozgune: I agree with @marcocitus on this. We'll only need to make this change for three functions.

PR #1794, opened 2017-11-13
Rename RemoveDirectory -> CitusRemoveDirectory

lithp:


src/backend/distributed/worker/worker_partition_protocol.c
```diff
@@ -674,7 +674,7 @@ RemoveDirectory(StringInfo filename)
 			fullFilename = makeStringInfo();
 			appendStringInfo(fullFilename, "%s/%s", directoryName, baseFilename);
 
-			RemoveDirectory(fullFilename);
+			CitusRemoveDirectory(fullFilename);
```
mtuncer: Are we going to use ```Citus``` prefix for going forward ?

jasonmp85: I'm assuming each of these symbols has a conflict with a similar call in Windows, since that's what @lithp is working on…

lithp: @mtuncer sent you an email

PR #1798, opened 2017-11-14
isolation test coverage improvement

furkansahin:
This PR is created to increase the coverage of the isolation tests by adding some of the concurrency tests such as;
* Two multi-shard update commands affecting the same rows [ADDED]
    - Sequential mode
    - Parallel Mode
* Two multi-shard update commands affecting different rows [ADDED]
    - Sequential mode
    - Parallel Mode
* master_append_table_to_shard, master_append_table_to_shard [ADDED]
* master_append_table_to_shard, master_apply_delete_command  [HELP IS NEEDED]
    - master_append_table_to_shard first
    - master_apply_delete_command first
* master_apply_delete_command, master_apply_delete_command [ADDED]
* create_distributed_table, create_distributed_table [ADDED]

### DIFF
--- a/src/test/regress/expected/isolation_master_append_table.out
+++ b/src/test/regress/expected/isolation_master_append_table.out
@@ -17,7 +17,7 @@ step s1-master_append_table_to_shard:
 
 master_append_table_to_shard
 
-0.0266667      
+0.213333       
 step s2-master_append_table_to_shard: 
 
    	SELECT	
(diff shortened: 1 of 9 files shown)


onderkalaci:
Before I give some feedback, could you please update the PR title and/or description and/or commit message? Actually, I cannot follow the necessity of the PR and what we actually want to solve.

Thanks!

src/test/regress/specs/isolation_master_append_table.spec
```diff
@@ -0,0 +1,97 @@
+setup
+{	
+  	CREATE TABLE table_to_append(id int);
+  	CREATE TABLE table_to_be_appended(id int);
+  	CREATE TABLE table_to_distribute(id int);
+
+  	SELECT create_distributed_table('table_to_append', 'id', 'append');
+  	INSERT INTO table_to_be_appended VALUES(1),(2),(3),(4),(5),(6);
+  	INSERT INTO table_to_distribute SELECT generate_series(1,100);
```
onderkalaci: Well, I think we should move the unrelated tests to separate files and give names accordingly. That's important for several reasons (i) Run them in parallel (yes, let's run those files in parallel:) ) (ii) Expand the tests in the future (which is very likely, I already suggested two new tests in this PR) (iii) The existing tests follow that pattern

So,

file1: isolation_master_append_table
file2: isolation_create_distributed_table
file3: isolation_master_apply_delete

Does that sound good to you as well?

src/test/regress/expected/isolation_master_append_table.out
```diff
@@ -0,0 +1,150 @@
...
+step s1-commit: 
+    COMMIT;
+
+step s2-create_distributed_table: <... completed>
+error in steps s1-commit s2-create_distributed_table: ERROR:  table "table_to_distribute" is already distributed
+step s2-commit: 
+	COMMIT;
+
+
+starting permutation: s1-begin s2-begin s1-master_append_table_to_shard s2-master_apply_delete_command s1-commit s2-commit
```
onderkalaci: Let's remove this from the isolation tests, we need to understand the problem here first. It could be related to false positive deadlocks we see with #1803 given that we're touching local files.

I'll be looking into this in detail.

src/test/regress/expected/isolation_master_append_table.out
```diff
@@ -0,0 +1,150 @@
...
+	WHERE
+		'table_to_append'::regclass::oid = logicalrelid;
+ <waiting ...>
+step s1-commit: 
+    COMMIT;
+
+step s2-master_append_table_to_shard: <... completed>
+master_append_table_to_shard
+
+0.0266667
```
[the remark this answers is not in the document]
furkansahin: WIP

src/test/regress/specs/isolation_master_append_table.spec
furkansahin: Yes, I was actually thinking about doing the same. I will update accordingly

onderkalaci: >file2: isolation_create_distributed_table

So, for example, we could add concurrent COPY to local table and create_distributed table. Which should be a valuable test

furkansahin: Do you have any suggestion on putting the concurrent apply_delete vs. append table tests under which file?

furkansahin: nvm, I am removing these tests.

src/test/regress/expected/isolation_master_append_table.out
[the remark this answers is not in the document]
furkansahin: Obviously, the reason was that the input size was so negligible and it couldn't express (or the VACUUM did not start yet) After making the input size larger, the results changed.

furkansahin:
@onderkalaci I added the tests you want but the output is different than the expectation. master_apply_delete_command takes an Exclusive lock on the relation. Hence, doesn't matter if the queries touch different shards, the later query is blocked.

Is this a reason to open a issue? I would argue this is the expected behavior since we are trying to avoid inserts into the same shard while we drop it. However, in that case, it would worth to examine more and try to take a lock on the shard and not the relation.

PR #1806, opened 2017-11-14
Don't spam the log with deadlock messages

lithp:
When building postgres you can pass a flag to configure: `'CFLAGS=-DEXEC_BACKEND'`, which emulates the Windows forking behavior. On `EXEC_BACKEND` builds of postgres we emit this deadlock detection message every time a new backend is started. It makes the logs pretty useless!


onderkalaci:
I think the right solution here is to get rid of the `citus.enable_deadlock_prevention` at all given that its useless and we only wanted to keep it around for few cycles to not break users' existing scripts etc.

I think 7.2 should be OK to remove the GUC.

marcocitus:
Fixed by #1876

lithp:
It might make sense to also remove the flag!

PR #1808, opened 2017-11-15
Bump Citus version to 7.2devel

marcocitus:


metdos:
Looks good 🚢

jasonmp85:
🥂 for getting this done quickly this cycle, since we always forget about it :-P

PR #1809, opened 2017-11-15
Get rid of some of the false positive distributed deadlocks

onderkalaci:
Fixes some parts of #1655.

Not sure if that's high priority enough to backport since I haven't actually hit this issue. This can be considered as a bugfix that fixes some theoretical bugs.

This change doesn't include the locks with `LOCKTAG_TUPLE` since the logic behind that is rather complex and I'm not sure that they are safe to exclude.

### DIFF
--- a/src/backend/distributed/transaction/lock_graph.c
+++ b/src/backend/distributed/transaction/lock_graph.c
@@ -496,6 +496,10 @@ BuildLocalWaitGraph(void)
  * IsProcessWaitingForSafeOperations returns true if the given PROC
  * waiting on relation extension locks, page locks or speculative locks.
  *
+ * The function also returns true if the waiting process is an autovacuum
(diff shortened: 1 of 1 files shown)


src/backend/distributed/transaction/lock_graph.c
```diff
@@ -509,17 +513,25 @@ IsProcessWaitingForSafeOperations(PGPROC *proc)
...
 	PROCLOCK *waitProcLock = NULL;
 	LOCK *waitLock = NULL;
+	PGXACT *pgxact = NULL;
 
 	if (proc->waitStatus != STATUS_WAITING)
 	{
 		return false;
 	}
 
+	pgxact = &ProcGlobal->allPgXact[proc->pgprocno];
```
marcocitus: Maybe add a short comment on what `allPgXact` contains.

PR #1810, opened 2017-11-15
Reduce default shard count in regression tests to improve performance

furkansahin:
Shard count is decreased from 32 to 4
Important note: The major change happened in multi_colocation_utils tests since I had to change the shard_count of one block of tables from 4 to 8 which had the replication factor as 2 and this resulted to affect the output of some queries. However, it is all handled.

This PR gives us ~22.25% gain in time for us according to my experiments. You can see my measurements from these gists ([before](https://gist.github.com/furkansahin/70c9aac9091706e10456fc7f30eec2ff), [after](https://gist.github.com/furkansahin/0d1e71b37017021f1d521b825ba03fe3)). Keep in mind that each time is the time for the next test vertically. 
You can see the total times at the end of both files.
Before: 164 sec (Thanks to my monstrous mac)
After: 127 sec (Thanks to lower shard count)
The effect is obvious in travis tests too.

In addition to the time gain, lower shard count will give us a chance to parallelize the tests more by decreasing the number of connections we use. 

Ps. In case anyone interested, I used the tool called [gnomon](https://github.com/paypal/gnomon) to print the times next to the tests.

### DIFF
--- a/src/test/regress/multi_schedule
+++ b/src/test/regress/multi_schedule
@@ -4,7 +4,7 @@
 # Regression tests that exercise distributed planning/execution functionality.
 #
 # All new regression tests are expected to be run by this schedule. Tests that
-# do not set specific task executor type should also be added to 
+# do not set specific task executor type should also be added to
 # multi_task_tracker_extra_schedule.
 #
 # Note that we use variant comparison files to test version dependent regression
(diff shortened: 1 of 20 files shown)


src/test/regress/multi_schedule
```diff
@@ -30,14 +30,9 @@ test: multi_read_from_secondaries
...
-test: multi_master_protocol
-test: multi_load_data
-
-test: multi_behavioral_analytics_create_table
-test: multi_behavioral_analytics_basics multi_behavioral_analytics_single_shard_queries multi_insert_select_non_pushable_queries
-test: multi_insert_select multi_insert_select_window
-test: multi_shard_update_delete
+test: multi_create_table_constraints multi_master_protocol multi_load_data multi_behavioral_analytics_create_table
+
+test: multi_behavioral_analytics_basics multi_behavioral_analytics_single_shard_queries multi_insert_select_non_pushable_queries multi_insert_select multi_insert_select_window multi_shard_update_delete
```
marcocitus: Could we do the parallelisation in a separate PR? This one is already quite huge, so it'd be easier to split them in order to get things merged.

furkansahin: Sure

PR #1811, opened 2017-11-15
EXEC_BACKEND support

lithp:
To test this you'll have to do two things:

1. Build an `EXEC_BACKEND` version of postgres. You need to re-configure postgres. Here's the command I use:

    ```
    ./configure '--enable-cassert' '--enable-debug' 'CFLAGS=-ggdb -Og -g3 -fno-omit-frame-pointer -DEXEC_BACKEND' '--prefix' '/home/brian/Work/pg-10-exec-backend'
    ```

    the important part is `CFLAGS=-DEXEC_BACKEND`.

2. Disable ASLR. Postgres always keeps its shared memory segments in the same place and uses absolute addresses within that shared memory. This is _usually_ fine, but occasionally that address conflicts with the address of the new process and postgres fails to attach the shared memory segments and the backend fails and the Citus regression test fails.

    Disable it by running this command as root:

    `echo 0 >/proc/sys/kernel/randomize_va_space`

    Don't forget to reset it (it was probably `2` previously) when you're done testing!

### DIFF
--- a/src/backend/distributed/worker/task_tracker.c
+++ b/src/backend/distributed/worker/task_tracker.c
@@ -100,11 +100,16 @@ TaskTrackerRegister(void)
 {
 	BackgroundWorker worker;
 
-	/* organize and register initialization of required shared memory */
-	RequestAddinShmemSpace(TaskTrackerShmemSize());
-
 	prev_shmem_startup_hook = shmem_startup_hook;
 	shmem_startup_hook = TaskTrackerShmemInit;
+
+	if (IsUnderPostmaster)
+	{
+		return;
+	}
+
+	/* organize and register initialization of required shared memory */
+	RequestAddinShmemSpace(TaskTrackerShmemSize());
 
 	/* and that the task tracker is started as background worker */
 	memset(&worker, 0, sizeof(worker));
--- a/src/backend/distributed/worker/task_tracker.c
+++ b/src/backend/distributed/worker/task_tracker.c
@@ -56,6 +56,9 @@ int MaxTrackedTasksPerNode = 1024; /* max number of tracked tasks */
 int MaxTaskStringSize = 12288; /* max size of a worker task call string in bytes */
 WorkerTasksSharedStateData *WorkerTasksSharedState; /* shared memory state */
 
+/* Hash table shared by the task tracker and task tracker protocol functions */
+HTAB *TaskTrackerTaskHash = NULL; /* shared memory */
+
 static shmem_startup_hook_type prev_shmem_startup_hook = NULL;
 
 /* Flags set by interrupt handlers for later service in the main loop */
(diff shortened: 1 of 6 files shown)


jasonmp85:
Plz keep git branches up-to-date with rebasing, not merging `master` into them (that results in a super-ugly/tangled history).

lithp:
@jasonmp85 It makes the history here ugly but if you choose "rebase and merge" when pulling into master the merge commits disappear.

jasonmp85:
Thing about _rebase and merge_ is… it's not a merge. It puts all these commits on the master branch. Probably shouldn't do that.

mtuncer:
I can't bring up postgres process in mac with this setting (ASLR is different in mac)

mtuncer:
verified it running under ubuntu

jasonmp85:
> I can't bring up postgres process in mac with this setting (ASLR is different in mac)

As in PostgreSQL straight-up won't start? If we're passing a custom `CFLAGS` value anyway, might as well disable PIE, right?

This script from Chromium seems capable of doing so on an already-built executable, but since our test requires a custom flag anyhow, maybe we can disable ASLR/PIE as part of compilation? It is hard to find any information about recent macOS' implementation of this behavior.

At any rate, here's [the script](https://src.chromium.org/viewvc/chrome/trunk/src/build/mac/change_mach_o_flags.py?view=markup).

This covers how one can create processes with ASLR already disabled (maybe PostgreSQL should have something inline like this?)… http://reverse.put.as/2011/08/11/how-gdb-disables-aslr-in-mac-os-x-lion/

src/backend/distributed/worker/task_tracker.c
```diff
@@ -56,6 +56,9 @@ int MaxTrackedTasksPerNode = 1024; /* max number of tracked tasks */
 int MaxTaskStringSize = 12288; /* max size of a worker task call string in bytes */
 WorkerTasksSharedStateData *WorkerTasksSharedState; /* shared memory state */
 
+/* Hash table shared by the task tracker and task tracker protocol functions */
+HTAB *TaskTrackerTaskHash; /* shared memory */
```
mtuncer: initializing it to NULL makes me feel better

src/backend/distributed/worker/task_tracker.c
```diff
@@ -98,22 +101,28 @@ TaskTrackerRegister(void)
 	BackgroundWorker worker;
 
 	/* organize and register initialization of required shared memory */
-	RequestAddinShmemSpace(TaskTrackerShmemSize());
+	if (!IsUnderPostmaster)
+	{
+		RequestAddinShmemSpace(TaskTrackerShmemSize());
+	}
 
 	prev_shmem_startup_hook = shmem_startup_hook;
```
mtuncer: perhaps move these 2 lines above ```if (!IsUnderPostmaster)``` and return from function ```if (IsUnderPostmaster)```

mtuncer:
don't forget to rebase/squash

Issue #1804, opened 2017-11-14
Recursive planning and execution using pull-push


marcocitus:
Most of Citus’ limitations in terms SQL support lie in subqueries and CTEs. Very often, the CTE or subquery itself is a query that could be planned and executed by on its own. One way to rapidly expand our SQL support is to recognise subqueries that cannot be planned using existing planners and plan and execute them independently, writing the results to an intermediate data structure that is then used in subsequent steps.

For instance, a query for the distinct number of sessions on the top 20 most visited pages:

```sql
SELECT
  page_id, count(distinct session_id)
FROM
  visits
WHERE
  page_id IN (SELECT page_id FROM visits GROUP BY page_id ORDER BY count(*) DESC LIMIT 20) 
GROUP BY
  page_id;
```

The LIMIT in the subquery prohibits subquery pushdown. If the table is not sharded by page_id, then the GROUP BY also prohibits subquery pushdown. However, the subquery itself can be executed as a distributed query:

```sql
SELECT page_id FROM visits GROUP BY page_id ORDER BY count(*) DESC LIMIT 20
```

We can execute this query separately and write the results to an intermediate data structure that is made available on the workers. In the outer query, the subquery is replaced by the intermediate result, e.g.:

```sql
SELECT
  page_id, count(distinct session_id)
FROM
  visits
WHERE
  page_id IN (SELECT * FROM intermediate_result(...) AS res(page_id int)) 
GROUP BY
  page_id;
```

A downside of recursive planning is that it prescribes a certain execution order, while an optimiser that considers the whole plan might find a more efficient alternative. However, it puts us in a position where we can gradually improve the underlying planners, which would reduce the number of recursions.

## Pull and push intermediate result

A simple implementation of recursive planning that would significantly enhance our SQL support is to execute unsupported subqueries and CTEs as regular queries on the coordinator and then push the results back into all relevant worker nodes and treat the result as a reference table that can be used in joins. This approach covers queries which require a merge step such as the LIMIT query above, as well as subqueries on local tables and functions.

A more sophisticated approach is to treat the intermediate result as a distributed table, which is sharded based on how the query result is joined. This improves efficiency and support for outer joins, but has additional implementation complexity.

We may want to use intermediate files instead of intermediate tables, because we cannot create intermediate tables in follower clusters. For small result sets (e.g. the LIMIT example), we could also consider including the results in the SQL query of the next step as VALUES or as an array parameter.

marcocitus:
### Task breakdown

[...]

PR #1814, opened 2017-11-16
Rename MultiPlan to DistributedPlan

marcocitus:
We currently use the name MultiPlan to refer to a distributed query plan. This is mostly for historical reasons (it refers to multi-relational algebra). We now plan a lot of things that have nothing to do with multi-relational algebra and in newer code we've started referring to it as distributed plans.

As part of #1804 we're going to have a new type of plan that is executed by running multiple distributed or local subplans. To avoid naming confusing, now seemed like a good time to rename MultiPlan to DistributedPlan.

This doesn't rename "multi" across the board, e.g. we still have `MultiNode`, which refers to actual multi-relational algebra and usage of the `multi_` prefix in filenames.


src/backend/distributed/utils/citus_outfuncs.c
```diff
@@ -173,9 +173,9 @@ OutMultiTreeRoot(OUTFUNC_ARGS)
 
 
 void
-OutMultiPlan(OUTFUNC_ARGS)
+OutDistributedPlan(OUTFUNC_ARGS)
 {
-	WRITE_LOCALS(MultiPlan);
+	WRITE_LOCALS(DistributedPlan);
 
 	WRITE_NODE_TYPE("MULTIPLAN");
```
mtuncer: perhaps change this to DistributedPlan too ? It may trigger a lot more changes if we want to move from ```MultiTree``` or ```MultiNode``` notation to ```DistributedTree/Node``` notation.

marcocitus:
> static MultiNode * MultiPlanTree(Query *queryTree);

Those are unrelated to `MultiPlan` and so were always a little misnamed. Changing them to `MultiNodeTree`.

src/include/distributed/distributed_planner.h
```diff
@@ -1,6 +1,6 @@
 /*-------------------------------------------------------------------------
  *
- * multi_planner.h
+ * distributed_planner.h
```
mtuncer: we should also change ```#define MULTI_PLANNER_H```   few lines below

mtuncer: and at the end of the file

PR #1816, opened 2017-11-17
Relax the checks for ensuring distribution columns in the target list

### LINKED ISSUE #1815: Feature regression on subquery pushdowns with reference tables from Citus 7 to 7.1
@saicitus just reported that three queries that used to work in Citus 7, but that #no longer do on Citus 7.1. All queries are related to subquery pushdowns in reference tables in Citus.

@saicitus will share more details in the upcoming days.

<pre>
SELECT
    DBE.id,
    DBE.event,
    DBE.userid,
    EXTRACT(EPOCH FROM FSE.timestamp) AS my_time,
    DBE.pointid,
    NULL AS info

FROM database_event AS DBE

INNER JOIN (
    SELECT DISTINCT test.id
    FROM testpoint AS test
    WHERE test.unique_id IN ('00cfdd')
) AS TEST_NAMES ON TEST_NAMES.id = DBE.pointid

ORDER BY DBE.my_time DESC
LIMIT 20
OFFSET 0;
</pre>

The `database_event` table is distributed on `pointid`. The `testpoint` table is a reference table.

In Citus 7, this query executes nicely. In Citus 7.1, we see the following error message:

<pre>
psql:q_distinct.sql:31: ERROR:  cannot push down this subquery
DETAIL:  Distinct on columns without partition column is currently unsupported
</pre>

Removing the `DISTINCT` clause also doesn't help. We then get this error message in Citus 7.1

<pre>
psql:q.sql:31: ERROR:  cannot push down this subquery
DETAIL:  Group by list without partition column is currently unsupported
</pre>

onderkalaci:
Fixes #1815

With this commit, we allow pushing down subqueries with only
reference tables where GROUP BY or DISTINCT clause or Window
functions include only columns from reference tables.

### DIFF
--- a/src/backend/distributed/planner/multi_logical_planner.c
+++ b/src/backend/distributed/planner/multi_logical_planner.c
@@ -84,6 +84,8 @@ static DeferredErrorMessage * DeferErrorIfUnsupportedUnionQuery(Query *queryTree
 static bool ExtractSetOperationStatmentWalker(Node *node, List **setOperationList);
 static DeferredErrorMessage * DeferErrorIfUnsupportedTableCombination(Query *queryTree);
 static bool WindowPartitionOnDistributionColumn(Query *query);
+static bool AllTargetExpressionsAreColumnReferences(List *targetEntryList);
+static bool RangeTableListContainsOnlyReferenceTables(List *rangeTableList);
 static FieldSelect * CompositeFieldRecursive(Expr *expression, Query *query);
 static bool FullCompositeFieldList(List *compositeFieldList);
 static MultiNode * MultiPlanTree(Query *queryTree);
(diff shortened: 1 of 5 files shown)


src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1234,10 +1236,121 @@ TargetListOnPartitionColumn(Query *query, List *targetEntryList)
...
+			if (PartitionMethod(relationId) != DISTRIBUTE_BY_NONE)
+			{
+				return false;
+			}
+		}
+		else
+		{
+			return false;
+		}
+	}
```
pykello: I wonder if the following style (which is the preferred style in some other companies) is still unacceptable in Citus:

```C
        if (rangeTableEntry->rtekind != RTE_RELATION ||
		    !IsDistributedTable(rangeTableEntry->relid) ||
			PartitionMethod(rangeTableEntry->relid) != DISTRIBUTE_BY_NONE)
		{
			return false;
		}
```

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1234,10 +1236,121 @@ TargetListOnPartitionColumn(Query *query, List *targetEntryList)
...
+
+	foreach(rangeTableCell, rangeTableList)
+	{
+		RangeTblEntry *rangeTableEntry = (RangeTblEntry *) lfirst(rangeTableCell);
+
+		if (rangeTableEntry->rtekind == RTE_RELATION)
+		{
+			Oid relationId = rangeTableEntry->relid;
+			bool distributedTable = IsDistributedTable(relationId);
+
```
mtuncer: extra space

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1234,10 +1236,121 @@ TargetListOnPartitionColumn(Query *query, List *targetEntryList)
...
+ * The function returns false for range table entries that are not relations.
+ *
+ * Note that the function doesn't recurse into subqueries, returns false when
+ * a subquery is found.
+ */
+static bool
+RangeTableListContainsOnlyReferenceTables(List *rangeTableList)
+{
+	ListCell *rangeTableCell = NULL;
+
```
mtuncer: extra space

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1234,10 +1236,121 @@ TargetListOnPartitionColumn(Query *query, List *targetEntryList)
...
+			return true;
+		}
+	}
+
+	return false;
+}
+
+
+/*
+ * RangeTableListContainsOnlyReferenceTables returns true if all relation range
```
mtuncer: possibly rephrase **if all relation range  tables entries are reference tables.** to **if all range table entries are reference tables.**

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1234,10 +1236,121 @@ TargetListOnPartitionColumn(Query *query, List *targetEntryList)
...
+ * elements in the target entry list belong to an outer query (for
+ * example the query is a sublink and references to another query
+ * in the from list).
+ *
+ * The function also returns true if any of the  target entries is not
+ * a column itself. This might be too restrictive, but, given that we're
+ * handling a very specific type of queries, that seems acceptable for now.
+ */
+static bool
+TargetListContainsNotSupportedEntries(List *targetEntryList)
```
mtuncer: I did not like this name. It would be better if we name this function as what it actually does. Implementation does not look like it care about being supported or not supported.

It is more like
```c
bool AllTargetExpressionsAreColumnReferences(List *targetEntryList)
```

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1234,10 +1236,121 @@ TargetListOnPartitionColumn(Query *query, List *targetEntryList)
 		}
 	}
 
+	/*
+	 * We could stil behave as if the target list is on partition column if
+	 * all range table entries are reference tables and there is no column
+	 * references to any outer queries.
+	 */
```
mtuncer: what if we have const expressions or functions ?

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1234,10 +1236,121 @@ TargetListOnPartitionColumn(Query *query, List *targetEntryList)
...
+
+		if (IsA(strippedColumnExpression, Var))
+		{
+			candidateColumn = (Var *) strippedColumnExpression;
+		}
+		else if (IsA(strippedColumnExpression, FieldSelect))
+		{
+			FieldSelect *compositeField = (FieldSelect *) strippedColumnExpression;
+			Expr *fieldExpression = compositeField->arg;
+
```
mtuncer: extra space

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1234,10 +1236,121 @@ TargetListOnPartitionColumn(Query *query, List *targetEntryList)
...
+	ListCell *targetEntryCell = NULL;
+
+	foreach(targetEntryCell, targetEntryList)
+	{
+		TargetEntry *targetEntry = lfirst(targetEntryCell);
+		Var *candidateColumn = NULL;
+
+		Expr *strippedColumnExpression = (Expr *) strip_implicit_coercions(
+			(Node *) targetEntry->expr);
+
```
mtuncer: extra space

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: We currently support columns only for hash distributed tables. Although it is possible to support those expressions on reference tables, I don't think they are very useful for now. Given that we'd want to backport this change, I'm inclined to keep the changes smaller.

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: Makes sense, updating the name.

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: Hmm, Is this really an extra space?

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: Oh, sure we should do that

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: I'm not resisting to change them, but, is this an extra space? I thought we have those new lines.

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: @pykello You wrote the coding conventions document so you can guide us towards new conventions :smiley_cat:

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: I'd like to keep this because we've more examples like this compared to your suggestion?

mtuncer:
rebase and 🚢

PR #1817, opened 2017-11-20
Regression parallelization

furkansahin:
This is the first part of our regression test parallelization process. Provides ~5-10% gain depending on the machine.

### DIFF
--- a/src/test/regress/expected/multi_explain.out
+++ b/src/test/regress/expected/multi_explain.out
@@ -452,10 +452,10 @@ Aggregate
                                 Sort Key: ((users.composite_id).tenant_id), ((users.composite_id).user_id)
                                 ->  Hash Join
                                       Hash Cond: (users.composite_id = events.composite_id)
-                                      ->  Seq Scan on users_1400029 users
+                                      ->  Seq Scan on users_1400033 users
                                             Filter: ((composite_id >= '(1,-9223372036854775808)'::user_composite_type) AND (composite_id <= '(1,9223372036854775807)'::user_composite_type))
                                       ->  Hash
-                                            ->  Seq Scan on events_1400025 events
+                                            ->  Seq Scan on events_1400029 events
                                                   Filter: ((event_type)::text = ANY ('{click,submit,pay}'::text[]))
 -- Union and left join subquery pushdown
 EXPLAIN (COSTS OFF)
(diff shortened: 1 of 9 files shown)


src/test/regress/multi_schedule
```diff
@@ -28,37 +28,26 @@ test: multi_read_from_secondaries
 # ----------
 # The following distributed tests depend on creating a partitioned table and
 # uploading data to it.
+# 
+# Tests for partitioning support
```
onderkalaci: The comment/title doesn't seem correct?

src/test/regress/expected/multi_explain.out
```diff
@@ -452,10 +452,10 @@ Aggregate
                                 Sort Key: ((users.composite_id).tenant_id), ((users.composite_id).user_id)
                                 ->  Hash Join
                                       Hash Cond: (users.composite_id = events.composite_id)
-                                      ->  Seq Scan on users_1400029 users
+                                      ->  Seq Scan on users_1400033 users
```
marcocitus: What triggered this change?

src/test/regress/sql/multi_insert_select_window.sql
```diff
@@ -2,9 +2,7 @@
 -- test insert select functionality for window functions
 -- ===================================================================
 
-TRUNCATE agg_results;
-
-INSERT INTO agg_results (user_id, agg_time, value_2_agg)
+INSERT INTO agg_results_window (user_id, agg_time, value_2_agg)
```
marcocitus: We could consider using different schema and search paths for different tests.

src/test/regress/expected/multi_explain.out
furkansahin: This happened after I create another table called agg_results_window for the window function tests. I added this table because this test was using the same table with another test which they run in parallel right now.

src/test/regress/sql/multi_insert_select_window.sql
furkansahin: This was the general attitude on multi_behavioral_analytics tests, so I just continued but I can apply what you suggest to all of the tests that I run parallel?

onderkalaci: Makes sense, let's consider it for the next changes. But, this change should be OK for now?

marcocitus: It probably makes sense, that way you can also just drop the schema at the end of the test and that would even speed up things a bit further (due to DROP TABLE in separate transactions being slow).

marcocitus: Yes, we can merge this.

PR #1818, opened 2017-11-21
Remove an ifdef surrounding strtoull

lithp:
We use strtoull in a few other places and do not wrap the calls with
ifdef, this commit makes us more consistent. Since we probably want to
require that you have strtoull it makes sense to fail at compile time,
not when you try to use the function.

### DIFF
--- a/src/backend/distributed/worker/worker_data_fetch_protocol.c
+++ b/src/backend/distributed/worker/worker_data_fetch_protocol.c
@@ -784,19 +784,14 @@ ExtractShardId(const char *tableName)
 	}
 	shardIdString++;
 
-#ifdef HAVE_STRTOULL
 	errno = 0;
-	shardId = strtoull(shardIdString, &shardIdStringEnd, 0);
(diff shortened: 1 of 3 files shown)


src/backend/distributed/worker/worker_data_fetch_protocol.c
```diff
@@ -784,7 +784,6 @@ ExtractShardId(const char *tableName)
 	}
 	shardIdString++;
 
-#ifdef HAVE_STRTOULL
```
mtuncer: shall we use ```pg_strtouint64``` instead ?

mtuncer: perhaps we should go replace all strtoull with pg_strtouint64.

lithp:
strtoull is already plenty portable, it works on Windows, but here you are

jasonmp85:
If `pg_strtouint64` is a nice wrapper that already exists, I say definitely use it.

PR #1820, opened 2017-11-21
Subqueries with immutable functions go through subquery pushdown

marcocitus:
This change makes sure that when we plug a `(SELECT * FROM read_records_file(...) AS ..)` into a query it will actually go through subquery pushdown where it will be nicely treated and not get flattened and abused in join order planner.


marcocitus:
I'm going to close this PR for the moment, since I ended up going with a slightly different approach in the prototype (explicitly check for an intermediate results function).

PR #1825, opened 2017-11-22
Register custom scans

onderkalaci:
While working with `sql_monster` (#1804), we realized that Postgres complains about not knowing about Citus' custom scan nodes. It appears that we should explicitly register the nodes.

The PR includes some refactoring on the custom scan related codes given that they were scattered and it was a bit hard to follow the code.

### DIFF
--- a/src/backend/distributed/executor/citus_custom_scan.c
+++ b/src/backend/distributed/executor/citus_custom_scan.c
@@ -1,0 +1,298 @@
+/*-------------------------------------------------------------------------
+ *
+ * citus_custom_scan.c
+ *
+ * Definitions of custom scan methods for all executor types.
+ *
+ * Copyright (c) 2012-2017, Citus Data, Inc.
+ *-------------------------------------------------------------------------
+ */
+#include "postgres.h"
+
+#include "miscadmin.h"
+
+#include "commands/copy.h"
+#include "distributed/citus_custom_scan.h"
+#include "distributed/insert_select_executor.h"
+#include "distributed/insert_select_planner.h"
+#include "distributed/multi_server_executor.h"
+#include "distributed/multi_router_executor.h"
+#include "distributed/multi_router_planner.h"
+#include "distributed/worker_protocol.h"
+#include "executor/executor.h"
+#include "nodes/makefuncs.h"
+#include "utils/memutils.h"
+#include "utils/rel.h"
+
+
+/* functions for creating custom scan nodes */
+static Node * RealTimeCreateScan(CustomScan *scan);
+static Node * TaskTrackerCreateScan(CustomScan *scan);
+static Node * RouterCreateScan(CustomScan *scan);
+static Node * CoordinatorInsertSelectCreateScan(CustomScan *scan);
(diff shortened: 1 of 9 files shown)


src/backend/distributed/executor/multi_task_tracker_executor.c
```diff
@@ -3004,3 +3007,37 @@ TrackerHashDisconnect(HTAB *taskTrackerHash)
...
+	CitusScanState *scanState = (CitusScanState *) node;
+	TupleTableSlot *resultSlot = NULL;
+
+	if (!scanState->finishedRemoteScan)
+	{
+		DistributedPlan *distributedPlan = scanState->distributedPlan;
+		Job *workerJob = distributedPlan->workerJob;
+
+		/* we are taking locks on partitions of partitioned tables */
+		LockPartitionsInRelationList(distributedPlan->relationIdList, AccessShareLock);
```
marcocitus: I've been wondering, why isn't this in `CitusSelectBeginScan`?

src/backend/distributed/executor/citus_custom_scan.c
```diff
@@ -0,0 +1,431 @@
...
+	.BeginCustomScan = CitusSelectBeginScan,
+	.ExecCustomScan = CoordinatorInsertSelectExecScan,
+	.EndCustomScan = CitusEndScan,
+	.ReScanCustomScan = CitusReScan,
+	.ExplainCustomScan = CoordinatorInsertSelectExplainScan
+};
+
+
+/*
+ * Let PostgreSQL know about Citus' custom scan nodes.
```
marcocitus: "RegisterCitusCustomScanMethods lets PostgreSQL know ..."

src/backend/distributed/executor/citus_custom_scan.c
```diff
@@ -0,0 +1,431 @@
...
+/*
+ * Load data collected by real-time or task-tracker executors into the tuplestore
+ * of CitusScanState. For that, we first create a tuple store, and then copy the
+ * files one-by-one into the tuple store.
+ *
+ * Note that in the long term it'd be a lot better if Multi*Execute() directly
+ * filled the tuplestores, but that's a fair bit of work.
+ */
+void
+LoadTuplesIntoTupleStore(CitusScanState *citusScanState, Job *workerJob)
```
marcocitus: Does this function really belong here? It seems more of a generic executor utility function  (multi_server_executor?).

For example, we'll use part of this logic for the `read_records_file` function.

onderkalaci: Makes sense, moving to `multi_executor` because there is an interdependency among`multi_server_executor ` and `citus_custom_scan`

src/backend/distributed/executor/multi_task_tracker_executor.c
onderkalaci: I saw it in your PR, so didn't want to interfere with your changes. Also, since this commit only moves files rather than modifying, I'd like to keep it as it is.

Issue #58, opened 2016-02-02
We currently require all join-clauses between tables to be equi-joins. Thus if I write something like: WHERE table1.id = table2.id AND table1.date < table2.date, this fails. We should allow the ' join if we have an equality join in an AND clause.


Jhanbanan:
We currently require all join-clauses between tables to be equi-joins. Thus if I write something like: WHERE table1.id = table2.id AND table1.date < table2.date, this fails. We should allow the ' join if we have an equality join in an AND clause.

lithp:
This recently came up in an email from a customer

robin900:
👍 . Actually, +100.

ozgune:
I'm copy/pasting the following from a conversation on Slack.

Citus is rejecting the <@ operator as an unsupported clause type for joins

 ```select b.created_at as dt, b.fk_account_id, tool,
             case when pm.fk_plan_id is null then 1 else pm.fk_plan_id end as fk_plan_id, b.duration_ms
             from
             ls.account_broadcasts b
             join ls.account_plan_membership pm
             on b.fk_account_id = pm.fk_account_id and b.created_at <@ pm.during```

Note that it’s an equi-join on the dist column (fk_account_id) but with an extra condition on the join and these two tables have same shard count.

Even if i tweak the query to move the `b.created_at <@ pm.during` to the WHERE clause, same error. and I notice that the location of `b.created_at = upper(pm.during)` which is a dummy clause for debugging here, doesn’t matter: it produces the same multi plan.

--

The two most common pattersn for us, when there is a compound join condition, are:
  
1) `a JOIN b on a.dist_column = b.dist_column AND a.some_tstz <@ b.some_tstzrange`
2) `a JOIN b on a.dist_column = b.dist_column AND a.tstz < b.tstz`
with a rare 3) `a JOIN b on a.dist_column = b.dist_column AND a.ctid < b.ctid` but we can work around the ctid problem.

My understanding, after reviewing the citus code, is that it would take somewhat significant refactoring to evaluate the list of clauses all at once, instead of `IsJoinClause()` called in series on a `clauseList`.

I note that the dummy join `a.fk_account_id = b.fk_account_id AND b.created_at = lower(a.during)` works merely because IsJoinClause returns true for both clauses.

They use `=` and left and right expr reference different tables (different `->varno`).
[...]

PR #1827, opened 2017-11-23
Enable non equi joins in subquery pushdown

onderkalaci:
Enable #58 for subquery pushdown.

Subquery pushdown planning is based on relation restriction
equivalence. This brings us the opportunity to allow any
other joins as long as there is an already equi join between
the distributed tables.

We already allow that for joins with reference tables and
this commit allows that for joins among distributed tables.


src/test/regress/expected/multi_subquery_behavioral_analytics.out
```diff
@@ -2040,6 +2040,21 @@ CREATE FUNCTION test_join_function_2(integer, integer) RETURNS bool
     LANGUAGE SQL
     IMMUTABLE
     RETURNS NULL ON NULL INPUT;
+SELECT run_command_on_workers($f$
```
marcocitus: You could use `run_command_on_master_and_workers` here.

PR #1830, opened 2017-11-25
Use RowExclusiveLock shard resource lock in COPY

marcocitus:
Seems COPY is taking a shard resource lock that conflicts with multi-shard UPDATE/DELETE/INSERT...SELECT statements, but it should only do so when replication factor > 1.


src/backend/distributed/commands/multi_copy.c
```diff
@@ -2079,7 +2079,12 @@ CitusCopyDestReceiverStartup(DestReceiver *dest, int operation,
 
 	/* prevent concurrent placement changes and non-commutative DML statements */
 	LockShardListMetadata(shardIntervalList, ShareLock);
-	LockShardListResources(shardIntervalList, ShareLock);
+
+	/*
+	 * Prevent concurrent UPDATE/DELETE on replication factor >1
+	 * (see multi_router_executor.c)
```
pykello: More specific address can help, e.g. "`see AcquireExecutorMultiShardLocks() at multi_router_executor.c`".

src/test/regress/expected/isolation_hash_copy_vs_all.out
```diff
@@ -97,13 +97,12 @@ create_distributed_table
 step s1-initialize: COPY hash_copy FROM PROGRAM 'echo 0, a, 0\\n1, b, 1\\n2, c, 2\\n3, d, 3\\n4, e, 4' WITH CSV;
 step s1-begin: BEGIN;
 step s1-copy: COPY hash_copy FROM PROGRAM 'echo 5, f, 5\\n6, g, 6\\n7, h, 7\\n8, i, 8\\n9, j, 9' WITH CSV;
-step s2-insert-select: INSERT INTO hash_copy SELECT * FROM hash_copy; <waiting ...>
+step s2-insert-select: INSERT INTO hash_copy SELECT * FROM hash_copy;
```
pykello: Do we have any tests that checks that we'll wait when replication factor > 1? If not, we should add some.

pykello:
@sumedhpathak @marcocitus 
I'm trying to address my feedback here and make this PR mergeable.

Last night, I was having troubles because my local `echo` requires `-e` to convert `\n` to newline, while isolation tests assume POSIX `echo` which doesn't require that.

After I work around that, I'm not getting the same output for `isolation_hash_copy_vs_all` test. I'm still getting some of the `<waiting ...>`, `<... completed>` messages locally, which we don't get at travis. I'm looking into that.

marcocitus:
I don't see any problem with the isolation tests. Are you sure the installed the right branch?

pykello:
@marcocitus recompiling citus & postgres fixed the problem. I'm not sure what I was doing wrong. This patch looks fine to me, I also did some manual testing. Merging this in interest of time, I'll add the tests in another PR.

byucesoy:
@pykello, while merging a PR net time, could you please use "Merge Pull Request" option in github UI. It adds PR number to the merge commit's title which makes tracking merged PRs/commits easier while creating packages.

PR #1831, opened 2017-11-27
Fix hard coded formatting strings for 64 bit numbers

mtuncer:
We have recently had issues with Citus Windows port regarding number format. Formatting for 64 bit numbers are different between linux/mac and Windows.

Postgres provides OS agnosting formatting macros for formatting 64 bit numbers. Replaced %ld %lu with INT64_FORMAT and UINT64_FORMAT respectively.

Also found some incorrect usages of formatting flags and fixed them.

### DIFF
--- a/src/backend/distributed/commands/create_distributed_table.c
+++ b/src/backend/distributed/commands/create_distributed_table.c
@@ -1261,13 +1261,13 @@ CopyLocalDataIntoShards(Oid distributedRelationId)
 
 		if (rowsCopied % 1000000 == 0)
 		{
-			ereport(DEBUG1, (errmsg("Copied %ld rows", rowsCopied)));
+			ereport(DEBUG1, (errmsg("Copied " UINT64_FORMAT " rows", rowsCopied)));
 		}
 	}
 
 	if (rowsCopied % 1000000 != 0)
 	{
-		ereport(DEBUG1, (errmsg("Copied %ld rows", rowsCopied)));
+		ereport(DEBUG1, (errmsg("Copied " UINT64_FORMAT " rows", rowsCopied)));
 	}
 
 	MemoryContextSwitchTo(oldContext);
--- a/src/backend/distributed/commands/multi_copy.c
+++ b/src/backend/distributed/commands/multi_copy.c
@@ -1156,7 +1156,7 @@ SendCopyDataToPlacement(StringInfo dataBuffer, int64 shardId, MultiConnection *c
 	if (!PutRemoteCopyData(connection, dataBuffer->data, dataBuffer->len))
 	{
 		ereport(ERROR, (errcode(ERRCODE_IO_ERROR),
-						errmsg("failed to COPY to shard %ld on %s:%d",
+						errmsg("failed to COPY to shard " INT64_FORMAT " on %s:%d",
 							   shardId, connection->hostname, connection->port),
 						errdetail("failed to send %d bytes %s", dataBuffer->len,
 								  dataBuffer->data)));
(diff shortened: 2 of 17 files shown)


src/backend/distributed/commands/multi_copy.c
```diff
@@ -1177,8 +1177,9 @@ EndRemoteCopy(int64 shardId, List *connectionList, bool stopOnFailure)
 			if (stopOnFailure)
 			{
 				ereport(ERROR, (errcode(ERRCODE_IO_ERROR),
-								errmsg("failed to COPY to shard %ld on %s:%d",
-									   shardId, connection->hostname, connection->port)));
+								errmsg(
+									"failed to COPY to shard " INT64_FORMAT " on %s:%d",
+									shardId, connection->hostname, connection->port)));
```
lithp: The extra line break is a little unfortunate :cry: Maybe do something like this?

selection_343

src/backend/distributed/master/master_metadata_utility.c
```diff
@@ -311,8 +311,9 @@ ShardIntervalsOnWorkerGroup(WorkerNode *workerNode, Oid relationId)
 			if (metadataLock == false)
 			{
 				ereport(WARNING, (errcode(ERRCODE_LOCK_NOT_AVAILABLE),
-								  errmsg("lock is not acquired, size of shard %ld "
-										 "will be ignored", shardId)));
+								  errmsg(
+									  "lock is not acquired, size of shard " UINT64_FORMAT
+									  " will be ignored", shardId)));
```
lithp: What if we raise the column limit to 100 chars :)

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -651,7 +651,8 @@ BuildPlacementSelectList(uint32 groupId, List *relationShardList)
 		placement = FindShardPlacementOnGroup(groupId, relationShard->shardId);
 		if (placement == NULL)
 		{
-			ereport(ERROR, (errmsg("no active placement of shard %ld found on group %d",
```
lithp: Wow, in some places `shardid` is signed, and in other places `shardid` is unsigned. We should fix this some day!

src/backend/distributed/commands/create_distributed_table.c
```diff
@@ -1261,13 +1261,13 @@ CopyLocalDataIntoShards(Oid distributedRelationId)
 
 		if (rowsCopied % 1000000 == 0)
 		{
-			ereport(DEBUG1, (errmsg("Copied %ld rows", rowsCopied)));
```
lithp: When I look at the diff this looks like you made a mistake, even though what you're really doing is fixing a mistake.

For future readers, maybe make a note in the commit message that you also fixed some inconsistent usage of `%ld` and `%lu`?

lithp:
Actually, sorry, I just noticed the codecov result. I know your patch didn't change this, but it probably makes sense to make this part of the code cleaner while we're here, and add some tests for these error messages?

src/backend/distributed/executor/multi_router_executor.c
jasonmp85: I've been saying this for like three years.

jasonmp85: (Because we store it in a PostgreSQL table, and because PostgreSQL doesn't technically have unsigned types in SQL, I say it should be signed everywhere).

src/backend/distributed/commands/multi_copy.c
jasonmp85: I had the same reaction, so :+1:

mtuncer: done

src/backend/distributed/executor/multi_router_executor.c
mtuncer: opened #1849

src/backend/distributed/commands/create_distributed_table.c
mtuncer: added to commit log

PR #1833, opened 2017-11-27
Refactor relation restriction equivalence checks to be more granular for subqueries

onderkalaci:
In subquery pushdown, we first ensure that each relation is joined with at least on another relation on the partition keys. That's fine given that the decision is binary: pushdown the query at all or not.

With #1804, we're considering another approach where if the query is not safe to pushdown, we replace _some_ parts of the query with a function call. Thus, we need the ability to understand which part(s) of the subquery is safe to pushdown.

This PR aims to provide the infrastructure for doing that.

Commits are broken down into smaller changes for ease of understanding. 

This is WIP in the sense that we'd want to test the changes with `sql_monster` before merging it.

### DIFF
--- a/src/backend/distributed/planner/multi_logical_planner.c
+++ b/src/backend/distributed/planner/multi_logical_planner.c
@@ -85,6 +85,17 @@ static DeferredErrorMessage * DeferErrorIfUnsupportedSubqueryPushdown(Query *
 																	  PlannerRestrictionContext
 																	  *
 																	  plannerRestrictionContext);
+static RelationRestrictionContext * FilterRelationRestrictionContext(
+	RelationRestrictionContext *relationRestrictionContext,
+	Relids
+	queryRteIdentities);
+static JoinRestrictionContext * FilterJoinRestrictionContext(
+	JoinRestrictionContext *joinRestrictionContext, Relids
+	queryRteIdentities);
+static bool RangeTableArrayContainsAnyRTEIdentities(RangeTblEntry **rangeTableEntries, int
+													rangeTableArrayLength, Relids
+													queryRteIdentities);
+static Relids QueryRteIdentities(Query *queryTree);
 static DeferredErrorMessage * DeferErrorIfUnsupportedSublinkAndReferenceTable(
 	Query *queryTree);
 static DeferredErrorMessage * DeferErrorIfUnsupportedFilters(Query *subquery);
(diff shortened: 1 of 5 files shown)


src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -616,6 +634,205 @@ DeferErrorIfUnsupportedSubqueryPushdown(Query *originalQuery,
...
+
+	/* simple_rte_array starts from 1, see plannerInfo struct */
+	for (rteIndex = 1; rteIndex < rangeTableCount; ++rteIndex)
+	{
+		RangeTblEntry *rangeTableEntry = rangeTableEntries[rteIndex];
+		List *rangeTableRelationList = NULL;
+		ListCell *rteRelationCell = NULL;
+
+		/* get list of all range tables in subquery tree */
+		ExtractRangeTableRelationWalker((Node *) rangeTableEntry,
```
marcocitus: This is neat, though slightly confusing. Worth mentioning that the list contains `rangeTableEntry` itself, plus any RTEs that occur in subqueries below the RTE.

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -534,8 +548,12 @@ DeferErrorIfUnsupportedSubqueryPushdown(Query *originalQuery,
 	ListCell *subqueryCell = NULL;
 	List *subqueryList = NIL;
 	DeferredErrorMessage *error = NULL;
-	RelationRestrictionContext *relationRestrictionContext =
-		plannerRestrictionContext->relationRestrictionContext;
+	RelationRestrictionContext *relationRestrictionContext = NULL;
+
+	/* we're only interested in the restriction context for the input queryTree */
+	plannerRestrictionContext =
+		FilterPlannerRestrictionForQuery(plannerRestrictionContext, originalQuery);
```
marcocitus: This call is not needed yet. We could declare `FilterPlannerRestrictionForQuery` in a header file and not use it yet.

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -616,6 +634,205 @@ DeferErrorIfUnsupportedSubqueryPushdown(Query *originalQuery,
...
+
+	ListCell *joinRestrictionCell = NULL;
+
+	foreach(joinRestrictionCell, joinRestrictionContext->joinRestrictionList)
+	{
+		JoinRestriction *joinRestriction =
+			(JoinRestriction *) lfirst(joinRestrictionCell);
+		RangeTblEntry **rangeTableEntries =
+			joinRestriction->plannerInfo->simple_rte_array;
+		int rangeTableCount = joinRestriction->plannerInfo->simple_rel_array_size;
```
marcocitus: The `simple_rel_array_size` is actually the number of range tables +1, so `rangeTableCount` might not be appropriate.

onderkalaci: That's interesting, but, I copied the example codes from postgres (i.e., [here](https://doxygen.postgresql.org/allpaths_8c.html#a007cfb04fab7231c9b4ef6b96dafa90b)) where they do what we're doing here. There are few more examples like this.

marcocitus:
Commit message seems inverted

onderkalaci:
> Commit message seems inverted

I'm planning to have a single commit with a complete explanation. The coverage decreased `0.24` without calling the function. It seems OK for now?

marcocitus:
Yep, I'm ok with that.

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: > The simple_rel_array_size is actually the number of range tables +1, so rangeTableCount might not be appropriate.

Btw, we're starting the iteration from `1` in `RangeTableArrayContainsAnyRTEIdentities()`, so should be fine?

marcocitus: Yes, I was only commenting on the inaccurate variable name. (Count instead of Size)

PR #1834, opened 2017-11-27
Subqueries containing functions go through subquery pushdown

marcocitus:
This change makes sure that when we plug a `(SELECT * FROM read_records_file(...) AS ..)` into a query it will actually go through subquery pushdown. Otherwise, it might get rewritten to `read_records_file(...) AS ..` and go into the logical planner, where it gets treated poorly.

### DIFF
--- a/src/backend/distributed/planner/multi_logical_planner.c
+++ b/src/backend/distributed/planner/multi_logical_planner.c
@@ -159,6 +159,9 @@ static MultiNode * ApplyCartesianProduct(MultiNode *leftNode, MultiNode *rightNo
  * Local functions forward declarations for subquery pushdown. Note that these
  * functions will be removed with upcoming subqery changes.
  */
+static bool ShouldUseSubqueryPushDown(Query *originalQuery, Query *rewrittenQuery);
+static bool IsFunctionRTE(Node *node);
+static bool FindNodeCheck(Node *node, bool (*check)(Node *));
 static Node * ResolveExternalParams(Node *inputNode, ParamListInfo boundParams);
 static MultiNode * SubqueryMultiNodeTree(Query *originalQuery,
 										 Query *queryTree,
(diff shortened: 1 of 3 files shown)


src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -229,6 +222,97 @@ MultiLogicalPlanCreate(Query *originalQuery, Query *queryTree,
 }
 
 
+/*
+ * ShouldUseSubqueryPushDown determines whether it's desirable to use
+ * subquery pushdown to plan the query based on the original and
+ * rewritten query.
+ */
+static bool
```
onderkalaci: I like to have a function for this purpose. We can move things to pushdown planner one by one. For example first, we add `HAS_OUTER_JOIN` here etc and gradually utilize pushdown abilities.

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -229,6 +222,97 @@ MultiLogicalPlanCreate(Query *originalQuery, Query *queryTree,
...
+	{
+		return false;
+	}
+
+	if (check(node))
+	{
+		return true;
+	}
+
+	if (IsA(node, RangeTblEntry))
```
onderkalaci: Some notes/thoughts on `range_table_walker ()`: In postgres source, it is used very few times. I think because if you have `query_tree_walker ()` and always start with a `Query` node, you don't need `range_table_walker ()`. It mostly seems useful if you could directly call the function with `rte` parameters.

That said, I don't see any downside of keeping it here, just wanted to shard my understanding on the function.

src/backend/distributed/planner/multi_logical_planner.c
marcocitus: Yep, that's a great idea, though we should probably figure out what causes the failures in https://github.com/citusdata/citus/pull/1824

src/backend/distributed/planner/multi_logical_planner.c
marcocitus: Actually, the right thing to do here is `return false`, because `query_tree_walker` already calls `range_table_walker` and it's being called redundantly here.

marcocitus: you're right that calling this function with an RTE would cause it to not descend, but you can call it `range_table_walker(rte, FindNodeCheck, check, QTW_EXAMINE_RTES)` directly in that case

PR #1835, opened 2017-11-27
Round-robin over worker nodes for 0-shard router queries

marcocitus:
When a router query prunes down to 0 shards, we generate a dummy placement which represents an arbitrary worker node (the first primary in `pg_dist_node`). Given that we'll start using this feature to plan queries on intermediate result files, we want to distribute the load across workers using round-robin.

### DIFF
--- a/src/backend/distributed/planner/multi_router_planner.c
+++ b/src/backend/distributed/planner/multi_router_planner.c
@@ -1629,6 +1629,8 @@ PlanRouterQuery(Query *originalQuery, RelationRestrictionContext *restrictionCon
 				List **placementList, uint64 *anchorShardId, List **relationShardList,
 				bool replacePrunedQueryWithDummy, bool *multiShardModifyQuery)
 {
+	static uint32 zeroShardQueryRoundRobin = 0;
+
 	bool isMultiShardQuery = false;
 	List *prunedRelationShardList = NIL;
 	DeferredErrorMessage *planningError = NULL;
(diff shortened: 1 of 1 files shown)


src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1629,6 +1629,8 @@ PlanRouterQuery(Query *originalQuery, RelationRestrictionContext *restrictionCon
 				List **placementList, uint64 *anchorShardId, List **relationShardList,
 				bool replacePrunedQueryWithDummy, bool *multiShardModifyQuery)
 {
+	static int zeroShardQueryRoundRobin = 0;
```
metdos: Can we have overflow problem if the server lives long enough?

pykello: Yeah, I think this at least should be unsigned so `%` doesn't return negative index when overflow happens.

marcocitus: Yep, good catch! Fixed it.

metdos:
It seems you also hit pg_dist_transaction problem :)
```
 SELECT count(*) FROM pg_dist_transaction;
  count 
 -------
-     4
+     6
 (1 row)
```

marcocitus:
> It seems you also hit pg_dist_transaction problem :)

yeh, we shouldn't be running `multi_mx_transaction_recovery` in parallel with `multi_mx_modifications`.

marcocitus:
Fixed up the master branch and rebased.

metdos:
I think you can 🚢

PR #1836, opened 2017-11-27
Ensure VACUUM/ANALYZE stays local when unsupported or DDL prop disabled

### LINKED ISSUE #1832: citus.enable_ddl_propagation=off does not work for ANALYZE
After setting `citus.enable_ddl_propagation=off`, `ANALYZE distributed_tables` gives a warning and says it is not pushing down it, but it actually pushes it.

The same issue also exists in 7.0.3 release.

Steps to replicate;
```sql
set citus.enable_ddl_propagation=off;
analyze github_events;
```

You can check logs at workers to see effects.

jasonmp85:
Apparently these two blocks have been incorrect for nearly a year…

Fixes #1832

### DIFF
--- a/src/test/regress/expected/multi_utilities.out
+++ b/src/test/regress/expected/multi_utilities.out
@@ -210,6 +210,45 @@ begin
     extract(epoch from clock_timestamp() - start_time);
 end
 $$ language plpgsql;
+\c - - - :worker_2_port
+CREATE MATERIALIZED VIEW prevcounts AS
+SELECT analyze_count, vacuum_count FROM pg_stat_user_tables
+WHERE relname='dustbunnies_990001';
+-- create function that sleeps until those counters increment
+create function wait_for_stats() returns void as $$
+declare
+  start_time timestamptz := clock_timestamp();
+  analyze_updated bool;
+  vacuum_updated bool;
+begin
+  -- we don't want to wait forever; loop will exit after 10 seconds
+  for i in 1 .. 100 loop
+
+    -- check to see if analyze has been updated
+    SELECT (st.analyze_count >= pc.analyze_count + 1) INTO analyze_updated
+      FROM pg_stat_user_tables AS st, pg_class AS cl, prevcounts AS pc
+     WHERE st.relname='dustbunnies_990001' AND cl.relname='dustbunnies_990001';
+
+     -- check to see if vacuum has been updated
+    SELECT (st.vacuum_count >= pc.vacuum_count + 1) INTO vacuum_updated
+      FROM pg_stat_user_tables AS st, pg_class AS cl, prevcounts AS pc
+     WHERE st.relname='dustbunnies_990001' AND cl.relname='dustbunnies_990001';
+
+    exit when analyze_updated or vacuum_updated;
+
+    -- wait a little
+    perform pg_sleep(0.1);
+
+    -- reset stats snapshot so we can test again
+    perform pg_stat_clear_snapshot();
+
+  end loop;
+
+  -- report time waited in postmaster log (where it won't change test output)
+  raise log 'wait_for_stats delayed % seconds',
+    extract(epoch from clock_timestamp() - start_time);
+end
+$$ language plpgsql;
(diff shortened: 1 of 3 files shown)


jasonmp85:
I'm not intending to own this fix myself, just gave it a quick look to see what could be causing it. I dunno why these two blocks were missing `return false` statements, but it seems as though they should have them.

Someone else can test and figure out how to add integration tests for these cases. In addition, it might be best to refactor this initialize `isSupported` to `false` and `return isSupported` at the end somehow.

jasonmp85:
OK, so I did the refactoring thing myself; whoever gets this into release (maybe support oncall) should do the testing bit (I haven't done any and didn't modify or add any tests). This is just my best guess at what causes the original bug.

metdos:
@jasonmp85, can we call this as a prototype? Is it accurate?

sumedhpathak:
The PR title is a bit unclear. This ensures that supported vacuum statements don't get propagated to workers when the enable_ddl_propagation flag is false right? Apart from Vacuum, is there is a list of other statements which we should verify?

jasonmp85:
@sumedhpathak you're right; the title was super unclear. The PostgreSQL structure is `VacuumStmt` even if the statement is an `ANALYZE`, which is the instigating bug, not `VACUUM`. I'll update the description and see whether I think there's anything else to verify.

src/test/regress/expected/multi_utilities.out
```diff
@@ -210,6 +210,45 @@ begin
     extract(epoch from clock_timestamp() - start_time);
 end
 $$ language plpgsql;
+\c - - - :worker_2_port
```
jasonmp85: At some point it's probably worth checking whether PostgreSQL changes their own way of doing this; this logic was copied from their own tests.

src/test/regress/expected/multi_utilities.out
```diff
@@ -317,12 +356,129 @@ WHERE tablename = 'dustbunnies_990002' ORDER BY attname;
...
+(2 rows)
+
+SELECT run_command_on_workers($$SELECT pg_stat_get_analyze_count(tablename::regclass) from pg_tables where tablename LIKE 'dustbunnies_%' limit 1$$);
+ run_command_on_workers 
+------------------------
+ (localhost,57637,t,3)
+ (localhost,57638,t,3)
+(2 rows)
+
+-- should not create the index in workers
```
jasonmp85: Do we have _any_ evidence that index creation/removal, truncation, etc. are not propagated? We shouldn't just add tests without reason (otherwise these test files get bloated and impossible to follow).

Since this PR is about `VACUUM`/`ANALYZE`, can you remove the index/truncate tests?

jasonmp85:
Actually I think the tests are testing _too_ much. This patch is concerned with `VACUUM`/`ANALYZE` and there are a bunch of added tests for `CREATE INDEX`/`DROP INDEX`/`TRUNCATE`, which are unaffected by the logic changes here.

IMO, the tests should be cut back to just be `VACUUM` or `ANALYZE` verification while `enable_ddl_propagation` is `false`, then it's a :shipit:.

metdos:
> Actually I think the tests are testing too much.

I think too much testing does not hurt :) @lfittl noted that if we had the same problem with TRUNCATE, it may cause to lose data, so I vote to keep tests even though they are semirelated with PR.

Shall we have another code reviewer given that @jasonmp85 was the owner of the code changes? @sumedhpathak?

pykello:
@metdos I am reviewing this.

jasonmp85:
:-1: for _Too much testing does not hurt_.

src/test/regress/expected/multi_utilities.out
furkansahin: Postgres keeps using the same logic. When I check the runtime of tests, this add +20sec (10 for each worker) since there is no analyze run in workers. Should we decrease the time or is there any other way you can think about?

jasonmp85: I think it's safest to just do what PostgreSQL does for this; it's too fiddly for us to try to re-solve on our own.

PR #1839, opened 2017-11-29
Support UNION with joins in the subqueries

marcocitus:
`SafeToPushdownUnionSubquery` currently returns false if a subquery has a relation that does not also have its partition key in the target list, which mostly prevents joins within a top-level UNION. 

After this change, `SafeToPushdownUnionSubquery` considers equivalences based on joins, such that it returns true for unions containing joins on partition column with the partition column of any table in the same place across unions.

At the moment, this enables joins within a `SELECT * FROM ((..subquery..) UNION (..subquery..)) u` query, and after #1804 this will enable full UNION support.

### DIFF
--- a/src/backend/distributed/planner/multi_logical_planner.c
+++ b/src/backend/distributed/planner/multi_logical_planner.c
@@ -633,8 +633,6 @@ DeferErrorIfUnsupportedSubqueryPushdown(Query *originalQuery,
 	ListCell *subqueryCell = NULL;
 	List *subqueryList = NIL;
 	DeferredErrorMessage *error = NULL;
-	RelationRestrictionContext *relationRestrictionContext =
-		plannerRestrictionContext->relationRestrictionContext;
 
 	if (originalQuery->limitCount != NULL)
 	{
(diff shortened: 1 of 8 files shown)


src/backend/distributed/planner/relation_restriction_equivalence.c
```diff
@@ -192,7 +199,7 @@ SafeToPushdownUnionSubquery(RelationRestrictionContext *restrictionContext)
 			/* union does not have partition key in the target list */
 			if (partitionKeyIndex == 0)
 			{
-				return false;
+				continue;
```
onderkalaci: We'd probably want to remove the above check `if (PartitionMethod(relationId) == DISTRIBUTE_BY_NONE)`, but not in this PR I guess?

src/backend/distributed/planner/relation_restriction_equivalence.c
```diff
@@ -222,7 +229,7 @@ SafeToPushdownUnionSubquery(RelationRestrictionContext *restrictionContext)
 		 */
 		if (partitionKeyIndex == InvalidAttrNumber)
 		{
-			return false;
+			continue;
```
onderkalaci: The comment is now obselete

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -649,7 +647,7 @@ DeferErrorIfUnsupportedSubqueryPushdown(Query *originalQuery,
 	 */
 	if (ContainsUnionSubquery(originalQuery))
 	{
-		if (!SafeToPushdownUnionSubquery(relationRestrictionContext))
+		if (!SafeToPushdownUnionSubquery(plannerRestrictionContext))
```
onderkalaci: The error message could be very confusing now :/ Would you mind changing it so that it also mentions about the partition key joins?

src/backend/distributed/planner/relation_restriction_equivalence.c
```diff
@@ -236,14 +243,26 @@ SafeToPushdownUnionSubquery(RelationRestrictionContext *restrictionContext)
...
 		}
 
 		AddToAttributeEquivalenceClass(&attributeEquivalance, relationPlannerRoot,
 									   varToBeAdded);
 	}
 
-	return EquivalenceListContainsRelationsEquality(list_make1(attributeEquivalance),
+	relationRestrictionAttributeEquivalenceList =
```
onderkalaci: Given that these code paths are slightly complex to the readers, could you add a comment here? I think a variation of the PR description should be sufficient.

PR #1840, opened 2017-11-29
Remove filter checks on leaf queries

marcocitus:
We currently error out if leaf queries in a UNION have different filters, this originated from the old subquery pushdown logic. With the current logic that checks for equivalences between tables, this restriction became obsolete and can be removed. This PR removes the error checks.


src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -703,12 +700,6 @@ DeferErrorIfUnsupportedSubqueryPushdown(Query *originalQuery,
 		{
 			return error;
 		}
-
-		error = DeferErrorIfUnsupportedFilters(subquery);
```
onderkalaci: This check also applies for non-set operations as well. So, we should add something like the following:
```SQL

SELECT *
FROM
  (SELECT *
   FROM
     (SELECT user_id,
             sum(value_2) AS counter
      FROM events_table
      WHERE user_id = 16
      GROUP BY user_id) AS foo,
     (SELECT user_id,
             sum(value_2) AS counter
      FROM events_table
      WHERE user_id = 15
      GROUP BY user_id) AS bar
   WHERE foo.user_id = bar.user_id ) AS baz;
```

(Well actually the above query is pretty useless, but, the following which is slightly more useful does seem to work on the master branch, which is probably a bug)
```SQL

SELECT *
FROM
  (SELECT *
   FROM
     (SELECT user_id,
             sum(value_2) AS counter
      FROM events_table
      WHERE user_id = 16
        OR user_id = 17
      GROUP BY user_id) AS foo,

     (SELECT user_id,
             sum(value_2) AS counter
      FROM events_table
      WHERE user_id = 15
        OR user_id = 16
      GROUP BY user_id) AS bar
   WHERE foo.user_id = bar.user_id ) AS baz;


```

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -976,129 +967,6 @@ DeferErrorIfUnsupportedSublinkAndReferenceTable(Query *queryTree)
...
-	partitionColumn = PartitionColumn(relationId, 0);
-
-	ExtractQueryWalker((Node *) subquery, &queryList);
-	foreach(queryCell, queryList)
-	{
-		Query *query = (Query *) lfirst(queryCell);
-		List *opExpressionList = NIL;
-		List *newOpExpressionList = NIL;
-
-		bool leafQuery = LeafQuery(query);
```
onderkalaci: `LeafQuery()` is also not used anymore, so remove more codes!

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -976,129 +967,6 @@ DeferErrorIfUnsupportedSublinkAndReferenceTable(Query *queryTree)
...
-			continue;
-		}
-
-		opExpressionList = PartitionColumnOpExpressionList(query);
-		if (opExpressionList == NIL)
-		{
-			continue;
-		}
-
-		newOpExpressionList = ReplaceColumnsInOpExpressionList(opExpressionList,
```
onderkalaci: `ReplaceColumnsInOpExpressionList` is also not used anymore

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -976,129 +967,6 @@ DeferErrorIfUnsupportedSublinkAndReferenceTable(Query *queryTree)
...
-		List *opExpressionList = NIL;
-		List *newOpExpressionList = NIL;
-
-		bool leafQuery = LeafQuery(query);
-		if (!leafQuery)
-		{
-			continue;
-		}
-
-		opExpressionList = PartitionColumnOpExpressionList(query);
```
onderkalaci: `PartitionColumnOpExpressionList` is not used at all as well

src/test/regress/expected/multi_subquery_union.out
```diff
@@ -23,6 +23,23 @@ LIMIT 5;
        2 |       4
 (5 rows)
 
+-- can use different filters on partition columns
+SELECT *
+FROM (
```
onderkalaci: `IN` is not checked. So, this test works in the master.  (`OR` also doesn't work). Seems like a bug in the previous implementation.

I think the test that started working is enough for `UNIONs`, maybe use `UNION ALL` with a `user_id=X` filters for completeness?

PR #1841, opened 2017-11-30
Send begin in real-time executor when in a coordinated transaction

marcocitus:
After this PR we send BEGIN and assign transaction IDs in the real-time executor if we are in a transaction block and did not yet send BEGIN on the connections.

This is necessary for #1829 to work, since it relies on the distributed transaction ID being known to the `read_intermediate_results` function when it is called from a SELECT on the worker. In addition, it avoids committing writes prematurely in case any of the functions called by the SELECT make changes to the database. Finally, it will also be necessary for repeatable read.

We skip data fetches when in a transaction block, because they are problematic inside a distributed transaction. For instance, many tasks might try to fetch the same reference table shard, and take the same advisory lock. As a result, the deadlock detector sees a distributed transaction that's blocked waiting for itself and kills it. This would be a problem when using the real-time executor in transactions independently of whether we send a BEGIN or not. Since we are planning to remove the data fetch code and it's not normally used, this seemed like an acceptable trade-off.


velioglu:
🚢

PR #1843, opened 2017-11-30
Test that COPY blocks UPDATE/DELETE/INSERT...SELECT when rep factor 2.

pykello:
In #1830 we made sure that COPY doesn't block UPDATE/DELETE/INSERT...SELECT when replication factor is 1, but we didn't have any tests to check that it blocks when replication factor is 2.


pykello:
@marcocitus can you please take a quick look at this? It just adds few tests to test your changes when replication factor is >1.

PR #1850, opened 2017-12-04
Sql monster tests

furkansahin:


src/test/regress/expected/multi_with.out
```diff
@@ -0,0 +1,870 @@
...
+)
+SELECT 
+  * 
+FROM
+  users_events
+ORDER BY
+  1, 2, 3, 4, 5, 6
+LIMIT
+  20;
+ERROR:  relation local_users is not distributed
```
marcocitus: Might be worth also having a test case in which `local_users` is queried in a CTE by itself, since that is supported.

src/test/regress/expected/multi_with.out
```diff
@@ -0,0 +1,870 @@
...
+      non_colocated_join
+    on (colocated_join.event_type = non_colocated_join.event_type)
+)
+SELECT
+  sum(uid), sum(event_type)
+FROM
+  users_events;
+  sum   |  sum   
+--------+--------
+ 972702 | 648967
```
marcocitus: go Citus!

marcocitus: if you find queries that error out, could you also add those to the regression tests?

furkansahin: Sure!

marcocitus:
@furkansahin  Could you rebase on the `ctes` branch?

furkansahin:
could not resolve some problems related to a previous rebase. deleting this pr and opening a new one after rebasing on the `ctes`

PR #1852, opened 2017-12-04
Treat recurring tuples as reference tables for GROUP BY checks

onderkalaci:
In subquery pushdown, we expect users to include distribution columns in `GROUP BY` clauses. Subqueries with reference tables was an exception, where we allow `GROUP BY`s without the distribution column (#1816). Recurring tuples are equivalent to reference tables, so allow it as well. 

This is relevant for #1804.

### DIFF
--- a/src/backend/distributed/planner/multi_logical_planner.c
+++ b/src/backend/distributed/planner/multi_logical_planner.c
@@ -109,7 +109,8 @@ static bool ExtractSetOperationStatmentWalker(Node *node, List **setOperationLis
 static DeferredErrorMessage * DeferErrorIfUnsupportedTableCombination(Query *queryTree);
 static bool WindowPartitionOnDistributionColumn(Query *query);
 static bool AllTargetExpressionsAreColumnReferences(List *targetEntryList);
-static bool RangeTableListContainsOnlyReferenceTables(List *rangeTableList);
+static bool FindNodeCheckInRangeTableList(List *rtable, bool (*check)(Node *));
+static bool IsDistributedTableRTE(Node *node);
 static FieldSelect * CompositeFieldRecursive(Expr *expression, Query *query);
 static bool FullCompositeFieldList(List *compositeFieldList);
 static MultiNode * MultiNodeTree(Query *queryTree);
@@ -1510,12 +1511,12 @@ TargetListOnPartitionColumn(Query *query, List *targetEntryList)
 
 	/*
 	 * We could still behave as if the target list is on partition column if
-	 * all range table entries are reference tables and all target expressions
-	 * are column references to the given query level.
+	 * all range table entries are reference tables or intermediate results,
+	 * and all target expressions are column references to the given query level.
 	 */
 	if (!targetListOnPartitionColumn)
 	{
-		if (RangeTableListContainsOnlyReferenceTables(query->rtable) &&
+		if (!FindNodeCheckInRangeTableList(query->rtable, IsDistributedTableRTE) &&
 			AllTargetExpressionsAreColumnReferences(targetEntryList))
 		{
 			targetListOnPartitionColumn = true;
@@ -1580,40 +1581,51 @@ AllTargetExpressionsAreColumnReferences(List *targetEntryList)
 
 
 /*
- * RangeTableListContainsOnlyReferenceTables returns true if all range table
- * entries are reference tables.
+ * FindNodeCheckInRangeTableList finds a node for which the check
+ * function returns true.
  *
- * The function returns false for range table entries that are not relations.
- *
- * Note that the function doesn't recurse into subqueries, returns false when
- * a subquery is found.
+ * FindNodeCheckInRangeTableList relies on FindNodeCheck() but only
+ * considers the range table entries.
  */
 static bool
-RangeTableListContainsOnlyReferenceTables(List *rangeTableList)
+FindNodeCheckInRangeTableList(List *rtable, bool (*check)(Node *))
 {
-	ListCell *rangeTableCell = NULL;
-	foreach(rangeTableCell, rangeTableList)
+	return range_table_walker(rtable, FindNodeCheck, check, QTW_EXAMINE_RTES);
+}
+
+
+/*
+ * IsDistributedTableRTE gets a node and returns true if the node
+ * is a range table relation entry that points to a distributed
+ * relation (i.e., excluding reference tables).
+ */
+static bool
+IsDistributedTableRTE(Node *node)
+{
+	RangeTblEntry *rangeTableEntry = NULL;
+	Oid relationId = InvalidOid;
+
+	if (node == NULL)
 	{
-		RangeTblEntry *rangeTableEntry = (RangeTblEntry *) lfirst(rangeTableCell);
+		return false;
+	}
 
-		if (rangeTableEntry->rtekind == RTE_RELATION)
-		{
-			Oid relationId = rangeTableEntry->relid;
+	if (!IsA(node, RangeTblEntry))
+	{
+		return false;
+	}
 
-			if (!IsDistributedTable(relationId))
-			{
-				return false;
-			}
+	rangeTableEntry = (RangeTblEntry *) node;
+	if (rangeTableEntry->rtekind != RTE_RELATION)
+	{
+		return false;
+	}
 
-			if (PartitionMethod(relationId) != DISTRIBUTE_BY_NONE)
-			{
-				return false;
-			}
-		}
-		else
-		{
-			return false;
-		}
+	relationId = rangeTableEntry->relid;
+	if (!IsDistributedTable(relationId) ||
+		PartitionMethod(relationId) == DISTRIBUTE_BY_NONE)
+	{
+		return false;
 	}
 
 	return true;
--- a/src/test/regress/expected/multi_subquery_complex_reference_clause.out
+++ b/src/test/regress/expected/multi_subquery_complex_reference_clause.out
@@ -1361,6 +1361,186 @@ ORDER BY 1 DESC
 LIMIT 4;
 ERROR:  cannot push down this subquery
 DETAIL:  Distinct on columns without partition column is currently unsupported
+-- test the read_intermediate_result() for GROUP BYs
+BEGIN;
+ 
+SELECT broadcast_intermediate_result('squares', 'SELECT s, s*s FROM generate_series(1,200) s');
+ broadcast_intermediate_result 
+-------------------------------
+                           200
+(1 row)
+
+-- single appereance of read_intermediate_result
+SELECT 
+  DISTINCT user_id 
+FROM 
+  users_table 
+JOIN 
+(SELECT 
+  max(res.val) as mx 
+FROM 
+  read_intermediate_result('squares', 'binary') AS res (val int, val_square int) 
+GROUP BY res.val_square) squares
+ ON (mx = user_id)
+ORDER BY 1
+LIMIT 5;
+ user_id 
+---------
+       1
+       2
+       3
+       4
+       5
+(5 rows)
+
+-- similar to the above, with DISTINCT on intermediate result
+SELECT DISTINCT user_id
+FROM users_table
+JOIN
+  (SELECT DISTINCT res.val AS mx
+   FROM read_intermediate_result('squares', 'binary') AS res (val int, val_square int)) squares ON (mx = user_id)
+ORDER BY 1
+LIMIT 5;
+ user_id 
+---------
+       1
+       2
+       3
+       4
+       5
+(5 rows)
+
+-- single appereance of read_intermediate_result but inside a subquery
+SELECT 
+  DISTINCT user_id 
+FROM 
+  users_table 
+JOIN (
+  SELECT *,random() FROM (SELECT 
+    max(res.val) as mx 
+  FROM 
+      (SELECT val, val_square FROM read_intermediate_result('squares', 'binary') AS res (val int, val_square int)) res 
+  GROUP BY res.val_square) foo)
+squares
+ ON (mx = user_id)
+ORDER BY 1
+LIMIT 5;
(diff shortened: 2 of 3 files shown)


src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1610,6 +1610,16 @@ RangeTableListContainsOnlyReferenceTables(List *rangeTableList)
...
+		else if (rangeTableEntry->rtekind == RTE_FUNCTION)
+		{
+			List *functionList = rangeTableEntry->functions;
+
+			if (!(list_length(functionList) == 1 &&
+				  ContainsReadIntermediateResultFunction(linitial(functionList))))
+			{
+				return false;
+			}
+		}
```
marcocitus: Would it make sense to add the other types of recurring tuples as well?

onderkalaci: I thought about that, but, I guess:

*  `RECURRING_TUPLES_EMPTY_JOIN_TREE` is not applicable here? Without `FROM` clause, I wasn't able to write a query with `GROUP BY`s.
* `RECURRING_TUPLES_FUNCTION`: I thought there is not much value on implementing that, but let me add that since people using Citus in various ways that I cannot think of 😄 

I'll add them now

marcocitus: > is not applicable here? 

This walker function may not be the right thing to do then. You could easily have `(SELECT 1 AS x) ref` as one of the items in the join tree.

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1610,6 +1611,28 @@ RangeTableListContainsOnlyReferenceTables(List *rangeTableList)
...
+			}
+			else if (!contain_mutable_functions((Node *) functionList))
+			{
+				continue;
+			}
+			else
+			{
+				return false;
+			}
+		}
```
marcocitus: I wonder, could we call `HasRecurringTuples` on each RTE or combine those functions somehow?

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1510,12 +1510,12 @@ TargetListOnPartitionColumn(Query *query, List *targetEntryList)
...
 	 * We could still behave as if the target list is on partition column if
-	 * all range table entries are reference tables and all target expressions
-	 * are column references to the given query level.
+	 * all range table entries are reference tables or intermediate results,
+	 * and all target expressions are column references to the given query level.
 	 */
 	if (!targetListOnPartitionColumn)
 	{
-		if (RangeTableListContainsOnlyReferenceTables(query->rtable) &&
+		if (RangeTableListContainsOnlyRecurringTuples(query->rtable) &&
```
marcocitus: The `read_intermediate_result` function is typically wrapped in a subquery. Would we still encounter it in the `rtable`?

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: `HasRecurringTuples` goes recursively down into the node, which is not intended at this point. Because, we're already going deep into the query tree as needed with the one level above function.

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: `TargetListOnPartitionColumn ()` is called for every `Query` node in the subquery. So, The following should also work:
```SQL
BEGIN;
 
SELECT broadcast_intermediate_result('squares', 'SELECT s, s*s FROM generate_series(1,200) s');

-- single appereance of read_intermediate_result
SELECT 
  DISTINCT user_id 
FROM 
  users_table 
JOIN (
	SELECT *,random() FROM (SELECT 
	  max(res.val) as mx 
	FROM 
  		read_intermediate_result('squares', 'binary') AS res (val int, val_square int) 
	GROUP BY res.val_square) foo)
squares
 ON (mx = user_id)
ORDER BY 1
LIMIT 5;
```

Is that what you mean? If so, I can add a regression test for it.

marcocitus: I mean that the way in which `read_intermediate_result` is used is:
```
SELECT 
  DISTINCT user_id 
FROM 
  users_table 
JOIN (
SELECT *,random() FROM (SELECT 
  max(res.val) as mx 
FROM 
  (SELECT val, val_square FROM read_intermediate_result('squares', 'binary') AS res (val int, val_square int)) res
GROUP BY res.val_square) foo)
squares
 ON (mx = user_id)
ORDER BY 1
LIMIT 5;
DETAIL:  Group by list without partition column is currently unsupported
```

onderkalaci: Hmm, could we actually end up with such query? As far as I see & test, we keep recursively planning until it we get rid of `GROUP BY`s?
```SQL
SELECT DISTINCT user_id
FROM users_table
JOIN
  (SELECT *,
          random()
   FROM
     (SELECT max(res.val) AS mx
      FROM
        (SELECT users_table.value_2 AS val,
                users_table.value_2*19 AS val_square
         FROM users_table
         GROUP BY 1,
                  2) res
      GROUP BY res.val_square) foo) squares ON (mx = user_id)
ORDER BY 1 ;
DEBUG:  replacing subquery SELECT value_2 AS val, (value_2 * 19) AS val_square FROM public.users_table GROUP BY value_2, (value_2 * 19) --> SELECT val, val_square FROM read_records_file('base/pgsql_job_cache/10_0_8877_1_0.data'::text, 'binary'::text) read_records_file(val integer, val_square integer)
DEBUG:  replacing subquery SELECT max(val) AS mx FROM (SELECT read_records_file.val, read_records_file.val_square FROM read_records_file('base/pgsql_job_cache/10_0_8877_1_0.data'::text, 'binary'::text) read_records_file(val integer, val_square integer)) res GROUP BY val_square --> SELECT mx FROM read_records_file('base/pgsql_job_cache/10_0_8877_1_1.data'::text, 'binary'::text) read_records_file(mx integer)
```

marcocitus: Man, recursive planning is pretty cool :), though I guess we'd avoid the recursion if we'd know how to handle the GROUP BY here?

onderkalaci: Or, why do we need this PR then 😄  I guess, it's still OK to have this for completeness in any way.

> though I guess we'd avoid the recursion if we'd know how to handle the GROUP BY here?

Yes, it probably makes sense. So, the rule would be `if (RangeTableListContainsOnlyRecurringTuples) -> stop recursion`

onderkalaci: > Yes, it probably makes sense. 

Ah, but that'd complicate the code here.

marcocitus: Not sure what you mean by `stop recursion`. This code gets called by `DeferErrorIfCannotPushdownSubquery`, which gets called by `ShouldRecursivelyPlanSubquery`. If this code decides that there is no problem with the GROUP BY on the subquery then we do not recursively plan that subquery.

onderkalaci: > though I guess we'd avoid the recursion if we'd know how to handle the GROUP BY here?

Yes, we could. But, this function is called for every `Query`in the input query. To understand the above case, we'd also need to recursively go into the `RTE_SUBQUERY`s in, which seems to complicate things. That's why I'd probably keep it simple for now. What do you think?

marcocitus: well, recursive planning is much more expensive than descending into the query again here, maybe add a TODO.

onderkalaci: Actually, I'm not considering the cost here. 

What I'm more concerned is that subquery pushdown error checks (i.e., `DeferErrorIfCannotPushdownSubquery()`) are done for every query, and I thought we've avoided any recursion within the checks. 

But, now I remembered that we had to recursively do the reference table checks to prevent pushing down cases like:

```SQL
SELECT * FROM
(SELECT user_id FROM users_ref_table) as u1 LEFT JOIN
(SELECT user_id FROM users_table) as u2 ON (u1.user_id = u2.user_id);
```

I think the query we discuss has very similar characteristics and seems like it'd be OK to call a function name `HasOnlyRecurringTuples()`, which is similar to `HasRecurringTuples` but returns true only when all the `RTEs` recurring tuples (i.e., Recursive version of `RangeTableListContainsOnlyRecurringTuples ()`).

Does that make sense to you as well? If yes, I can add a `TODO` comment here.

marcocitus: > Does that make sense to you as well?

Yes, ok. We're kind of missing such a function.

onderkalaci: I'll add that to this PR

onderkalaci: I added the function and the test you've mentioned above. See `HasNonRecurringTuples()` function, it is slightly different than what I expected at first. But, seems simpler to implement.

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1580,43 +1580,33 @@ AllTargetExpressionsAreColumnReferences(List *targetEntryList)
...
  */
 static bool
-RangeTableListContainsOnlyReferenceTables(List *rangeTableList)
+HasNonRecurringTuples(Node *node)
 {
+	List *rangeTableList = NULL;
 	ListCell *rangeTableCell = NULL;
+
+	/* extract range table entries for simple relations only */
+	ExtractRangeTableRelationWalker((Node *) node, &rangeTableList);
```
marcocitus: Maybe:

```
bool
ContainsDistributedTableRTE(Node *node)
{
    return FindNodeCheck(node, IsDistributedTableRTE);
}

bool
IsDistributedTableRTE(Node *node)
{
    RangeTblEntry *rangeTableEntry = NULL;
    Oid relationId = InvalidOid;

    if (!IsA(node, RangeTblEntry))
    {
        return false;
    }

    rangeTableEntry = (RangeTblEntry *) node;
    if (rangeTableEntry->rtekind != RTE_RELATION)
    {
        return false;
    }

    relationId = rangeTableEntry->relid;
    if (!IsDistributedTable(relationId) ||
        PartitionMethod(relationId) == DISTRIBUTE_BY_NONE)
    {
        return false;
    }

    return true;
}
```

(copied from sql_monster)

onderkalaci: > Instead of constructing a big long list and then potentially only looking at the first item, maybe we could do:

makes sense

onderkalaci: Ah, that required some minor change to `FindNodeCheck()` given that if it doesn't start with a `Query *`, it doesn't recurse into the rte. But, at this point, we pass a `List *rtable`, which I've added a commit to fix

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -316,8 +317,7 @@ FindNodeCheck(Node *node, bool (*check)(Node *))
 
 	if (IsA(node, RangeTblEntry))
 	{
-		/* query_tree_walker descends into RTEs */
-		return false;
+		return range_table_walker(list_make1(node), FindNodeCheck, check, 0);
```
marcocitus: Hmz, `query_tree_walker` also calls `range_table_walker`, so this would create a large amount of redundant work.

onderkalaci: Not that much given that we return as soon as we find the node we're looking for? 
I'm ok with both implementations, just pick one 😄

marcocitus: Yes, but it might take a while to find it, or maybe we won't find it. Can't we just pass a query to `ContainsDistributedTableRTE`?

marcocitus: thinking about it a bit more, maybe we should have a different walker that does not invoke `query_tree_walker` and only considers rtable?

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: Well, I think I prefer to leave this as is for now even it adds some overhead. It's much simpler than adding a new walker or modifying the existing ones.

marcocitus: After this discussion, it appears to me that `ExtractRangeTableRelationWalker` has exponential complexity and produces duplicates.

onderkalaci:
> After this discussion, it appears to me that ExtractRangeTableRelationWalker has exponential complexity (each subtree is revisited) and can produce an exponential number of duplicates. We should maybe fix that first.

Ops, I haven't seen this comment before I push my changes. I think that's a separate issue and should be handled separately? If that's OK, I can open a new issue for that with tag 7.1?

src/test/regress/expected/multi_subquery_complex_reference_clause.out
```diff
@@ -1361,6 +1361,169 @@ ORDER BY 1 DESC
...
+SELECT 
+  DISTINCT user_id 
+FROM 
+  users_table 
+JOIN 
+(SELECT 
+  max(res.val) as mx 
+FROM 
+  read_intermediate_result('squares', 'binary') AS res (val int, val_square int) 
+GROUP BY res.val_square) squares
```
marcocitus: Maybe add one more regression test with a DISTINCT.

onderkalaci: added

onderkalaci:
> I like the new approach.

Actually, this approach was already instructed in `FindNodeCheck`'s function comment. I've missed that for a long time :/

PR #1854, opened 2017-12-05
Increase sleep time in a regression test to give Valgrind tests enough time

metdos:
I also manually tested on Valgrind to see that if 1 sec is enough.

### DIFF
--- a/src/test/regress/expected/multi_transaction_recovery.out
+++ b/src/test/regress/expected/multi_transaction_recovery.out
@@ -250,7 +250,8 @@ SELECT pg_reload_conf();
  t
 (1 row)
 
-SELECT pg_sleep(0.1);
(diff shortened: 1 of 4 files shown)


src/test/regress/expected/multi_transaction_recovery.out
```diff
@@ -250,7 +250,7 @@ SELECT pg_reload_conf();
  t
 (1 row)
 
-SELECT pg_sleep(0.1);
+SELECT pg_sleep(1);
```
marcocitus: We also do this in multi_mx_transaction_recovery

metdos: @marcocitus added there too

metdos:
> Could use a comment saying why we need so much sleep

done

PR #1855, opened 2017-12-05
Regression tests for CTEs

furkansahin:
all of the CTE tests are failing after moving the tests from sql_monster to this branch


onderkalaci:
One high level observation I had is when you add `DISTINCT` to the target list of repartition queries, the execution times drops noticably. So, might make sense to add `DISTINCT` some of the queries to get better performance for the tests. (Sure we don't want all queries to run with `DISTINCT`)

src/test/regress/expected/multi_with_prepare.out
```diff
@@ -0,0 +1,339 @@
+-- prepared statements
+PREPARE prepared_test_1 AS
```
marcocitus: can you add some parameters to the prepared statements?

furkansahin: Ah, it was planned though, but then it was not supported yesterday yet :D Adding them too

marcocitus:
> all of the CTE tests are failing after moving the tests from sql_monster to this branch

Yep, but I think these are a reasonable set of queries to keep and we expect them to start working soon.

For the moment, we should also have some simpler queries that work on the current branch. Some expectations worth testing:

- data modifying statements in CTEs should error out
- recursive CTEs should error out
- Nested CTEs should not be able to reference outer CTEs
- we should not be able to put a subquery with a CTE outside the FROM or WHERE clause
- we should be able to use a CTE in the FROM clause
- we should be able to use a CTE in a subquery in the FROM clause
- we should be able to use a CTE in the WHERE clause
- [cte reference] UNION distributed_table should error out
- [cte reference] UNION [cte reference] should work
- [cte reference] LEFT JOIN distributed_table should error out
- CTEs prior to INSERT...SELECT via the coordinator should work
- CTEs prior to any other modification should error out
- CTEs should be able to use local queries
- CTEs should be able to use router queries
- CTEs should be able to use real-time queries
- CTEs should be able to use task-tracker queries
- CTEs should not be able to terminate (the last SELECT) in a local query
- CTEs should be able to terminate a router query
- CTEs should be able to terminate a real-time query
- CTEs should be able to terminate a task-tracker query

src/test/regress/sql/multi_with_basics.sql
```diff
@@ -0,0 +1,201 @@
+-- CTE cannot be outside of FROM/WHERE clause
+WITH cte_basic AS (
+	SELECT * FROM users_table
+)
+SELECT cte_basic, user_id FROM users_table;
```
marcocitus: I think the more relevant example is `SELECT (SELECT user_id FROM cte_basic), user_id FROM users_table` since this is something that would work in postgres, but not in Citus.

src/test/regress/expected/multi_with_nested.out
```diff
@@ -0,0 +1,458 @@
...
+),
+nested2 AS (
+  WITH nested2_1 AS (
+    SELECT * FROM nested1_1
+  )
+  SELECT * FROM nested2_1
+)
+SELECT * FROM nested2;
+ERROR:  relation "nested2" does not exist
+LINE 3:     SELECT * FROM nested2
```
marcocitus: note when it says `LINE ...` it's a postgres syntax error, which never hits any Citus code.

src/test/regress/sql/multi_with_basics.sql
```diff
@@ -0,0 +1,201 @@
...
+
+-- CTEs in FROM should work
+WITH cte AS (
+	SELECT user_id, value_2 from users_table WHERE user_id = 1 ORDER BY 2 LIMIT 5
+)
+SELECT * FROM cte;
+
+-- create and use CTE in FROM should work
+SELECT * FROM (WITH cte AS (
+		SELECT user_id, value_2 from users_table WHERE user_id = 1 ORDER BY 2 LIMIT 5
```
marcocitus: Note that this query and others that have `user_id = 1` work because they are router queries and do not hit the recursive planner.

src/test/regress/expected/multi_with_join.out
```diff
@@ -396,4 +404,193 @@ LIMIT
...
+EXPLAIN (COSTS false, VERBOSE true)
+WITH cte AS (
+  SELECT * FROM users_table
+)
+SELECT * FROM cte join reference_table ON cte.user_id + 1 = reference_table.user_id;
+                                                                                                             QUERY PLAN                                                                                                             
+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+ Custom Scan (Citus Router)
+   Output: remote_scan.user_id, remote_scan."time", remote_scan.value_1, remote_scan.value_2, remote_scan.value_3, remote_scan.value_4, remote_scan.user_id_1
+   ->  Distributed Subplan
```
marcocitus: The explain output is going to be pretty noisy and will change randomly. We could maybe use some SQL feature that is only available in the router planner to verify that it's a router plan, alternatively we could turn on debug output.

src/test/regress/expected/multi_with_nested.out
furkansahin: Then I am removing all of the tests related to access from inner to outer CTEs which are all ended up with this error.

src/test/regress/expected/multi_with_basics.out
```diff
@@ -0,0 +1,367 @@
+-- CTE cannot be outside of FROM/WHERE clause
+WITH cte AS (
```
onderkalaci: Well, it looks like we support this query if the subquery returns one rows. So, maybe use this example and update the comment? (I've no idea why people could use subqueries in the target list )
```SQL

 WITH cte AS (   SELECT user_id FROM users_table WHERE value_2 IN (1, 2) LIMIT 1  )
SELECT (SELECT * FROM cte);
```

marcocitus:
Thanks for all your work on the regression tests! These have been integrated into #1853 .

PR #1856, opened 2017-12-05
Avoid deadlock with DROP TABLE in ColocatedTableId

### LINKED ISSUE #1502: Concurrent create_distributed_table and drop table might deadlock if tables are colocated

marcocitus:
This PR changes the lock on `pg_dist_partition` from `ShareLock` to `AccessShareLock` in `ColocatedTableId` to avoid getting into deadlocks with `DROP TABLE`, and adds an additional check to confirm the table still exists.

When dropping a distributed table, `DROP TABLE` always takes an `AccessExclusiveLock` on the table first (mostly outside of Citus' control), and then we take a `RowExclusiveLock` on `pg_dist_partition` to remove the corresponding row.

In `ColocatedTableId`, which is called from `create_distributed_table`, we first take a `ShareLock` on `pg_dist_partition` to prevent tables from being dropped concurrently, and then take an `AccessShareLock` on a table with the given colocation ID to prevent the table from being dropped while we're still reading its shard metadata in order to co-locate the new distributed table with the existing distributed table.

Because these operations take conflicting locks in the opposite order, we often see deadlocks in Travis tests and users have noticed this as well. 

By changing the `ShareLock` on `pg_dist_partition` to an `AccessShareLock`, it is no longer conflicting with the `RowExclusiveLock` in `DROP TABLE`. However, the consequence is that a table might be dropped in between getting a `pg_dist_partition` snapshot and taking a lock on the OID of table. We therefore check whether the table still exists after obtaining a lock on the OID. If it does not exist, we try another table with the same colocation ID until all tables have been exhausted.

If all the tables with the colocation ID turn out to have been dropped, we return an InvalidOid, which means `create_distributed_table` uses round-robin placement policy. When the co-location group is empty, there is a very small risk that two concurrent `create_distributed_table` calls will place shards differently. We should expand the lock in `ColocationIdForNewTable` to also apply to empty co-location groups. However, this problem is unrelated to the `DROP TABLE` deadlock and already exists when a co-location group is empty.

Fixes #1502

### DIFF
--- a/src/backend/distributed/utils/colocation_utils.c
+++ b/src/backend/distributed/utils/colocation_utils.c
@@ -963,25 +963,46 @@ ColocatedTableId(Oid colocationId)
 	ScanKeyInit(&scanKey[0], Anum_pg_dist_partition_colocationid,
 				BTEqualStrategyNumber, F_INT4EQ, ObjectIdGetDatum(colocationId));
 
-	/* do not allow any tables to be dropped while we read from pg_dist_partition */
-	pgDistPartition = heap_open(DistPartitionRelationId(), ShareLock);
+	pgDistPartition = heap_open(DistPartitionRelationId(), AccessShareLock);
 	tupleDescriptor = RelationGetDescr(pgDistPartition);
 	scanDescriptor = systable_beginscan(pgDistPartition,
 										DistPartitionColocationidIndexId(),
 										indexOK, NULL, scanKeyCount, scanKey);
 
 	heapTuple = systable_getnext(scanDescriptor);
-	if (HeapTupleIsValid(heapTuple))
+	while (HeapTupleIsValid(heapTuple))
 	{
+		Relation colocatedRelation = NULL;
+
 		colocatedTableId = heap_getattr(heapTuple, Anum_pg_dist_partition_logicalrelid,
 										tupleDescriptor, &isNull);
 
-		/* make sure the table isn't dropped for the remainder of the transaction */
+		/*
+		 * Make sure the relation isn't dropped for the remainder of
+		 * the transaction.
+		 */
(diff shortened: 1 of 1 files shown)


metdos:
Changes look good 🚢 . I also [tested it](https://github.com/citusdata/citus/issues/1502#issuecomment-349585928) to see if #1502 is fixed. Shall we also close it with this PR?

PR #1857, opened 2017-12-05
Use proper schema in read_intermediate_result signature

marcocitus:
This PR changes the schema of the copy format type from `citus` to `pg_catalog`. It also adds a missing `SETOF` in `read_intermediate_result`, which caused assert failures in some scenarios.

Putting an enum type in the `citus` schema is a precedent set by `create_distributed_table`, but  users generally do not have access to the `citus` schema, which creates subtle issues. On cloud, you can do `SELECT create_distributed_table('test','x','hash')`, but not `SELECT create_distributed_table('test','x','hash'::citus.distribution_type)`, because it accesses the `citus` schema.

Unfortunately, this issue becomes less subtle with `read_intermediate_results` since the deparsing logic adds casts to practically anything. Moreover, we were using `LookupTypeNameOid` to look up the type OID for some internal operations, which also goes through access control checks.


onderkalaci:
> (We generally don't update the .sql files, but seems OK for this time.)

@marcocitus decided to create a new file and I tested upgrade path again, all looks good

PR #1859, opened 2017-12-06
Strip implicit coercions on LIMIT/OFFSET in WorkerLimitCount

marcocitus:
Minor change to `WorkerLimitCount` to strip implicit coercions in LIMIT and OFFSET in `WorkerLimitCount`. This gets in the way of using an original query in the logical planner, which would allow us to skip an extra `standard_planner` call during recursive planning.


marcocitus:
Closing, since passing original queries to the logical planner requires a more thorough approach.

PR #1861, opened 2017-12-06
New guc to allow task executor swap

### LINKED ISSUE #444: Consider automatically switching to task tracker executor if query needs repartitioning
If a query requires repartitioning (distributed shuffles), users have to manually switch to the task tracker executor.

As the real-time executor cannot execute queries which require repartitioning, it might be worth always making that switch automatically.

furkansahin:
We need this change for sql_monster. I didn't think through the solution much but this seemed like working. Need to add some tests though.. 
Fixes #444

### DIFF
--- a/src/backend/distributed/executor/multi_server_executor.c
+++ b/src/backend/distributed/executor/multi_server_executor.c
@@ -29,6 +29,7 @@
 int RemoteTaskCheckInterval = 100; /* per cycle sleep interval in millisecs */
 int TaskExecutorType = MULTI_EXECUTOR_REAL_TIME; /* distributed executor type */
 bool BinaryMasterCopyFormat = false; /* copy data from workers in binary format */
+bool EnableRepartitionJoins = false;
 
 
 /*
@@ -98,13 +99,25 @@ JobExecutorType(DistributedPlan *distributedPlan)
 									  "\"task-tracker\".")));
 		}
 
-		/* if we have repartition jobs with real time executor, error out */
+		/* if we have repartition jobs with real time executor and repartition
+		 * joins are not enabled, error out. Otherwise, switch to task-tracker
+		 */
 		dependedJobCount = list_length(job->dependedJobList);
 		if (dependedJobCount > 0)
 		{
-			ereport(ERROR, (errmsg("cannot use real time executor with repartition jobs"),
-							errhint("Set citus.task_executor_type to "
-									"\"task-tracker\".")));
+			if (!EnableRepartitionJoins)
+			{
+				ereport(ERROR, (errmsg(
+									"the query contains a join that requires repartitioning"),
+								errhint("Set citus.enable_repartition_joins to on "
+										"to enable repartitioning")));
+			}
+
+			ereport(DEBUG1, (errmsg(
+								 "cannot use real time executor with repartition jobs"),
+							 errhint("Since you enabled citus.enable_repartition_joins "
+									 "Citus chose to use task-tracker.")));
+			return MULTI_EXECUTOR_TASK_TRACKER;
 		}
 	}
 	else
--- a/src/backend/distributed/shared_library_init.c
+++ b/src/backend/distributed/shared_library_init.c
@@ -746,6 +746,16 @@ RegisterCitusConfigVariables(void)
 		0,
 		NULL, NULL, NULL);
 
+	DefineCustomBoolVariable(
+		"citus.enable_repartition_joins",
+		gettext_noop("Allows Citus to use task-tracker executor when necessary."),
+		NULL,
+		&EnableRepartitionJoins,
+		false,
+		PGC_USERSET,
+		0,
+		NULL, NULL, NULL);
+
 	DefineCustomEnumVariable(
 		"citus.shard_placement_policy",
 		gettext_noop("Sets the policy to use when choosing nodes for shard placement."),
--- a/src/include/distributed/multi_server_executor.h
+++ b/src/include/distributed/multi_server_executor.h
@@ -190,6 +190,7 @@ typedef struct WorkerNodeState
 extern int RemoteTaskCheckInterval;
 extern int MaxAssignTaskBatchSize;
 extern int TaskExecutorType;
+extern bool EnableRepartitionJoins;
 extern bool BinaryMasterCopyFormat;
 extern int MultiTaskQueryLogLevel;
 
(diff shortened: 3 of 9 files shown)


src/backend/distributed/executor/multi_server_executor.c
```diff
@@ -74,37 +75,69 @@ JobExecutorType(DistributedPlan *distributedPlan)
...
+		}
+		else if (taskCount >= reasonableConnectionCount && AllowExecutorSwap)
+		{
+			ereport(WARNING, (errmsg("this query uses more file descriptors than the "
+									 "configured max_files_per_process limit"),
+							  errhint("Since you enabled citus.allow_executor_swap "
+									  "Citus chose to use task-tracker. "
+									  "Consider increasing max_files_per_process to run "
+									  "this query faster.")));
+			return MULTI_EXECUTOR_TASK_TRACKER;
```
sumedhpathak: We don't quite want this right? This *increases* the cases where we use the task-tracker executor. Before this we would warn, but still use the real-time executor. We should only chose the task-tracker executor for repartition jobs dynamically. We then probably want a separate config to turn off task-tracker entirely, which would retain current behavior.

marcocitus: Right, we should only consider switching when dependedJobCount > 0.

src/backend/distributed/shared_library_init.c
```diff
@@ -746,6 +746,16 @@ RegisterCitusConfigVariables(void)
 		0,
 		NULL, NULL, NULL);
 
+	DefineCustomBoolVariable(
+		"citus.allow_executor_swap",
```
marcocitus: This name is not very descriptive.

I'd have a preference for `task_tracker_query_log_level`, analogous to `multi_task_query_log_level`, or perhaps `repartition_query_log_level`.

An alternative would be `enable_task_tracker`, but we'd like to preserve the behaviour of `SET citus.task_executor_type TO 'task-tracker'` and it would be a little confusing if task-tracker is not enabled but you can still run queries with it.

marcocitus: Or maybe `enable_repartition_joins` with values 'on', 'off' (default).

furkansahin: changed to enable_repartition_joins

src/backend/distributed/executor/multi_server_executor.c
```diff
@@ -98,13 +99,24 @@ JobExecutorType(DistributedPlan *distributedPlan)
...
+		 */
 		dependedJobCount = list_length(job->dependedJobList);
-		if (dependedJobCount > 0)
+		if (dependedJobCount > 0 && !EnableRepartitionJoin)
 		{
 			ereport(ERROR, (errmsg("cannot use real time executor with repartition jobs"),
 							errhint("Set citus.task_executor_type to "
-									"\"task-tracker\".")));
+									"\"task-tracker\" or "
+									"citus.enable_repartition_joins to ON.")));
```
marcocitus: Just because I'm lazy and I like copy-pasting this message. Could you add a "set" after the "or"? :)

src/backend/distributed/executor/multi_server_executor.c
```diff
@@ -98,13 +99,24 @@ JobExecutorType(DistributedPlan *distributedPlan)
...
+		}
+		else if (dependedJobCount > 0 && EnableRepartitionJoin)
```
marcocitus: this block looks slightly odd, why not

```
if (dependedJobCount > 0)
{
    if (!EnableRepartitionJoin)
    {
      // throw error
    }

    return MULTI_EXECUTOR_TASK_TRACKER;
}

File: src/backend/distributed/executor/multi_server_executor.c [hunk 5]
```diff
@@ -98,13 +99,24 @@ JobExecutorType(DistributedPlan *distributedPlan)
...
+		{
+			ereport(WARNING, (errmsg(
+								  "cannot use real time executor with repartition jobs"),
```
marcocitus: This should probably just be a DEBUG message.

src/backend/distributed/executor/multi_server_executor.c
marcocitus: maybe `ON` -> `on`.

src/backend/distributed/executor/multi_server_executor.c
```diff
@@ -98,13 +99,26 @@ JobExecutorType(DistributedPlan *distributedPlan)
...
-			ereport(ERROR, (errmsg("cannot use real time executor with repartition jobs"),
-							errhint("Set citus.task_executor_type to "
-									"\"task-tracker\".")));
+			if (!EnableRepartitionJoin)
+			{
+				ereport(ERROR, (errmsg(
+									"cannot use real time executor with repartition jobs"),
+								errhint("Set citus.task_executor_type to "
+										"\"task-tracker\" or "
+										"set citus.enable_repartition_joins to on.")));
```
marcocitus: Looking at this error message again, it contains a lot of terminology, maybe we could simplify it to:

```
ERROR: the query contains a repartition join, but repartition joins are currenly disabled
HINT: Set citus.enable_repartition_joins to on to enable repartition joins.
```

(we also want people to start forgetting about `citus.task_executor_type` as we prepare for a unified executor)

marcocitus: or maybe:
```
ERROR: the query contains a join that requires repartitioning
HINT: Set citus.enable_repartition_joins to on to enable repartitioning
```

furkansahin: done

marcocitus: Cool, maybe squash into 1 commit and :ship:

sumedhpathak:
@furkansahin, @marcocitus  I am looking at this PR for documentation purposes. This PR does not remove task_executor_type. Is there a reason we still retain this parameter, if we are able to dynamically choose the task-tracker if needed for repartitioning? 

Specifically, looks like this flag is "Enable Repartition Joins If task_executor is real-time"? If I set this flag to false, but set task_executor_type to task-tracker, then I will still be able to execute repartition queries right?

Removing the config would be backwards incompatible, but should we add a warning or notice somewhere about its deprecation?

marcocitus:
We kept the GUC for backwards compatibility and to retain the ability to run regular distributed SQL queries (without re-partitioning) through task-tracker.

PR #1864, opened 2017-12-08
Reduce the frequency of FinishConnectionIO calls during COPY

marcocitus:
A while ago, we had a problem with COPY running out of memory because libpq buffers were getting to large. We then provided some back pressure by calling `FinishConnectionIO` after sending a tuple, which solved the memory problem. Unfortunately, the number of CPU cycles spent in this function hurt throughput more than anticipated. 

By only flushing once the buffer grows >8MB we get ~5x better throughput on COPY, INSERT...SELECT and moving intermediate results. We still wait for all the data to flush in `PutRemoteCopyEnd`.

### DIFF
--- a/src/backend/distributed/connection/remote_commands.c
+++ b/src/backend/distributed/connection/remote_commands.c
@@ -21,6 +21,9 @@
 #include "utils/palloc.h"
 
 
+#define MAX_PUT_COPY_DATA_BUFFER_SIZE (8 * 1024 * 1024)
+
+
 /* GUC, determining whether statements sent to remote nodes are logged */
 bool LogRemoteCommands = false;
 
@@ -580,11 +583,20 @@ PutRemoteCopyData(MultiConnection *connection, const char *buffer, int nbytes)
 	 * until the socket is writable to prevent the internal libpq buffers
 	 * from growing excessively.
 	 *
-	 * In the future, we could reduce the frequency of these pushbacks to
-	 * achieve higher throughput.
+	 * We currently allow the internal buffer to grow to 8MB before
+	 * providing back pressure based on experimentation that showed
+	 * throughput get worse at 4MB and lower due to the number of CPU
+	 * cycles spent in networking system calls.
 	 */
 
-	return FinishConnectionIO(connection, allowInterrupts);
+	connection->copyBytesWrittenSinceLastFlush += nbytes;
+	if (connection->copyBytesWrittenSinceLastFlush > MAX_PUT_COPY_DATA_BUFFER_SIZE)
+	{
+		connection->copyBytesWrittenSinceLastFlush = 0;
+		return FinishConnectionIO(connection, allowInterrupts);
+	}
+
+	return true;
 }
 
 
@@ -615,6 +627,8 @@ PutRemoteCopyEnd(MultiConnection *connection, const char *errormsg)
 	}
 
 	/* see PutRemoteCopyData() */
+
+	connection->copyBytesWrittenSinceLastFlush = 0;
 
 	return FinishConnectionIO(connection, allowInterrupts);
 }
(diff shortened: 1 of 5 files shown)


src/backend/distributed/connection/remote_commands.c
```diff
@@ -580,11 +581,18 @@ PutRemoteCopyData(MultiConnection *connection, const char *buffer, int nbytes)
...
-	 * In the future, we could reduce the frequency of these pushbacks to
-	 * achieve higher throughput.
+	 * We currently allow the internal buffer to grow to 8MB before
+	 * providing back pressure based on experimentation that showed
+	 * throughput get worse at 4MB and lower due to the number of CPU
+	 * cycles spent in networking system calls.
 	 */
 
-	return FinishConnectionIO(connection, allowInterrupts);
+	if (pgConn->outCount > 8 * 1024 * 1024)
```
ozgune: Minor question. Any thoughts on putting the 8MB into a constant variable, or adding parentheses around 8MB?

ozgune:
@metdos -- Could we throughly test and backport this change?

pykello:
@marcocitus I replace using `pgConn->outCount` with manually tracking copy byte count as @lithp suggested. This is because runtime and compile-time libpq's can be different and `pgConn->outCount` can be at different offsets.

I saw 11x performance gain on a table with a single integer column. I am doing some more tests, but please take a look in the meanwhile.

src/backend/distributed/connection/remote_commands.c
```diff
@@ -616,6 +628,8 @@ PutRemoteCopyEnd(MultiConnection *connection, const char *errormsg)
 
 	/* see PutRemoteCopyData() */
 
+	connection->copyBytesWrittenSinceLastFlush = 0;
```
marcocitus: We keep the connection open after an abort and currently we don't call `PutRemoteCopyEnd` prior to closing a connection (which causes "unexpected message type 0x58" errors, but that's a separate issue). 

We should probably reset this flag in `AfterXactHostConnectionHandling`.

marcocitus: Thinking about this more, it's probably ok since we don't cache connections if they are still in a transaction when the abort happens. Worth double checking.

pykello: From what I‌ understand,
1. If the transaction is rolled back after a *successful* `COPY FROM`, `PutRemoteCopyEnd()` has already been called, and we are good.
2. If the transaction is rolled back after an *unsuccessful* `COPY FROM`, `PQstatus()` returns `CONNECTION_BAD` so we shut-down and free the connection in `AfterXactHostConnectionHandling()`. So we are good in this case too, since we will use a new connection for the next copy.

But I think resetting copy state in `AfterXactHostConnectionHandling()` doesn't hurt, so I added it.

Also, please note that the only place that we have to reset `copyBytesWrittenSinceLastFlush` is `PutRemoteCopyData()`. Other places are mostly nice-to-haves and shouldn't impact the performance much, and doesn't impact correctness at all.

marcocitus:
The change looks good to me. 

It's semantically slightly different because now we'll always flush after 8MB even if the buffer is not growing, but that also allows us to write a test that exercises the code path deterministically and get back the code coverage.

For example:
```
-- copy >8MB to a single worker to trigger a flush
BEGIN;
CREATE UNLOGGED TABLE trigger_flush AS SELECT 1 AS a, s AS b, s AS c, s AS d, s AS e, s AS f, s AS g, s AS h FROM generate_series(1,150000) s;
SELECT create_distributed_table('trigger_flush','a');
ABORT;
```

This copies 150000*(8*8+2) = 9900000 bytes to a single worker. I've confirmed that this triggers a call to `FinishConnectionIO` in `PutRemoteCopyData`.

lithp:
Thanks for cleaning these changes up!

PR #1866, opened 2017-12-09
Allow count(distinct) in queries with a subquery

marcocitus:
We currently error out on queries with count(distinct) if there is a subquery, but it seems we no longer need that check.

### DIFF
--- a/src/test/regress/expected/multi_subquery.out
+++ b/src/test/regress/expected/multi_subquery.out
@@ -152,7 +152,7 @@ ERROR:  cannot push down this subquery
 DETAIL:  Limit in subquery without limit in the outermost query is unsupported
 -- reset the flag for next query
 SET citus.subquery_pushdown to OFF;
--- Check that we error out if the outermost query is a distinct clause.
+-- Check that we support count distinct with a subquery
 SELECT
 	count(DISTINCT a)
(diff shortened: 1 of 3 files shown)


mtuncer:
This enables count(distinct) queries on top of subqueries to go through. 

However, we can not push distinct down to worker nodes even if the count distinct is on the partition column.  This would work with sub-optimal performance.

Solution is to determine if the count(distinct) column is on the partition column of the subquery, and push count distinct to worker.

mtuncer:
hold it, taking ship back

mtuncer:
We should only allow count distincts, and error out in other aggregate distincts

src/test/regress/expected/multi_subquery.out
```diff
@@ -152,7 +152,7 @@ ERROR:  cannot push down this subquery
 DETAIL:  Limit in subquery without limit in the outermost query is unsupported
 -- reset the flag for next query
 SET citus.subquery_pushdown to OFF;
--- Check that we error out if the outermost query is a distinct clause.
+-- Check that we support count distinct with a subquery
```
mtuncer: probably add an error case for
sum(distinct) and avg(distinct)

PR #1871, opened 2017-12-13
Relax checks on recurring tuples in FROM with sublinks

marcocitus:
We aggressively error out when we see a query of the form `SELECT .. FROM .. <reference table> ... WHERE .. <distributed table>.. `, but such queries are acceptable if the reference table is also joined with a distributed table, since the tuples would still be partitioned by the join. An exception is when the join is an outer join with the reference table on the outside, but those cases are already captured by the outer join checks.

### DIFF
--- a/src/backend/distributed/planner/multi_logical_planner.c
+++ b/src/backend/distributed/planner/multi_logical_planner.c
@@ -97,8 +97,7 @@ static bool RangeTableArrayContainsAnyRTEIdentities(RangeTblEntry **rangeTableEn
 													rangeTableArrayLength, Relids
 													queryRteIdentities);
 static Relids QueryRteIdentities(Query *queryTree);
-static DeferredErrorMessage * DeferErrorIfUnsupportedSublinkAndReferenceTable(
-	Query *queryTree);
+static DeferredErrorMessage * DeferErrorIfFromClauseRecurs(Query *queryTree);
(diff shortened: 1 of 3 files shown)


src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -938,41 +940,69 @@ DeferErrorIfUnsupportedSublinkAndReferenceTable(Query *queryTree)
...
+	else if (recurType == RECURRING_TUPLES_EMPTY_JOIN_TREE)
+	{
+		return DeferredError(ERRCODE_FEATURE_NOT_SUPPORTED,
+							 "cannot pushdown the subquery",
+							 "Subqueries without FROM are not allowed in FROM "
+							 "clause when the outer query has subqueries in "
+							 "WHERE clause", NULL);
+	}
+
+	/*
```
onderkalaci: Minor: I cannot think of such a query. Can you show an example?

marcocitus: `SELECT (SELECT 1)`. It has sublinks, but rtable is empty.

onderkalaci: Thanks, I forgot that target list subqueries are also called `sublink`s

marcocitus: could also be `SELECT 1 WHERE (SELECT 1 > 0);`

Issue #1870, opened 2017-12-13
SELECT FOR UPDATE in subqueries quietly allowed


marcocitus:
Citus incorrectly allows you to execute some SELECT ... FOR UPDATE commands by wrapping the FOR UPDATE in a subquery. 

```
CREATE TABLE test (x int, y int);
SELECT create_distributed_table('test','x');

SELECT * FROM test WHERE x = 1 FOR UPDATE;
ERROR:  could not run distributed query with FOR UPDATE/SHARE commands
HINT:  Consider using an equality filter on the distributed table's partition column.

SELECT * FROM (SELECT * FROM test FOR UPDATE) foo WHERE x = 1;
 x | y 
---+---
 1 | 1
(1 row)

```

While FOR UPDATE is useful, Citus currently *does not execute FOR UPDATE commands correctly*. As long as we don't support it these commands should error out.

onderkalaci:
> SELECT * FROM test WHERE x = 1 FOR UPDATE;
ERROR:  could not run distributed query with FOR UPDATE/SHARE commands
HINT:  Consider using an equality filter on the distributed table's partition column.


Note that the error message is misleading, given that it's also a router query.

marcocitus:
SELECT..FOR UPDATE is allowed in router queries, and this seems to error out correctly for multi-shard queries.

PR #1876, opened 2017-12-15
Recursively plan subqueries that are not safe to pushdown (i.e., requires merge step)

onderkalaci:
Implements one piece of #1804.

With this commit, Citus recursively plans subqueries that
are not safe to pushdown, in other words, requires a merge
step.

The algorithm is simple: Recursively traverse the query from the bottom 
up (i.e., bottom meaning the leaf queries). On each level, check 
whether the query is safe to pushdown (or a single repartition 
subquery). If the answer is yes, do not touch that subquery. If the 
answer is no, plan the subquery separately (i.e., create a subPlan 
for it) and replace the subquery with a call to 
`read_intermediate_results(planId, subPlanId)`. During the the 
execution, run the subPlans first, and make them available to the 
next query executions.

### Some of the queries that this change allows us:
   
   * Subqueries with LIMIT
   * Subqueries with GROUP BY/DISTINCT on non-partition keys
   * Subqueries involving re-partition joins, router queries
   * Mixed usage of subqueries and CTEs (i.e., use CTEs in 
     subqueries as well). Nested subqueries as long as we 
     support the subquery inside the nested subquery.
   * Subqueries with local tables (i.e., those subqueries 
     has the limitation that they have to be leaf subqueries)

   * VIEWs on the distributed tables just works (i.e., the 
     limitations mentioned below still applies to views)

### Some of the queries that are still NOT supported:

  * Correlated subqueries that are not safe to pushdown (i.e., cross subquery references are not on partition keys i.e., `users_table.value_1`)
```SQL
SELECT count(*)
FROM users_table
WHERE user_id IN
    (SELECT user_id
     FROM events_table
     WHERE events_table.user_id = users_table.value_1
     ORDER BY 1
     LIMIT 5);
```

  * Window function on non-partition keys (i.e., `PARTITION BY event_type`)
```SQL
SELECT *
FROM
  ( SELECT user_id,
           TIME,
           rnk
   FROM
     ( SELECT *,
              rank() OVER my_win AS rnk
      FROM events_table WINDOW my_win AS (PARTITION BY event_type
                                          ORDER BY TIME DESC) ) AS foo
   ORDER BY 3 DESC, 1 DESC, 2 DESC
   LIMIT 10) AS foo;
```
  * Recursively planned subqueries or CTEs on the outer
    side of an outer join (i.e., `foo` is recursively planned)
```SQL
SELECT foo.value_2
FROM
  (SELECT users_table.value_2
   FROM users_table,
        events_table
   WHERE users_table.user_id = events_table.user_id
     AND event_type IN (1,
                        2,
                        3,
                        4)
   LIMIT 5) AS foo
LEFT JOIN
  (SELECT users_table.value_2
   FROM users_table,
        events_table
   WHERE users_table.user_id = events_table.user_id
     AND event_type IN (5,
                        6,
                        7,
                        8)) AS bar ON(foo.value_2 = bar.value_2);
```

  * Only recursively planned subqueries and CTEs in the FROM
    (i.e., not any distributed tables in the FROM) and subqueries
    in WHERE clause (i.e., `foo` is recursively planned)
```SQL
SELECT *
FROM
  ( SELECT users_table.user_id
   FROM users_table,

     (SELECT user_id
      FROM events_table) AS evs
   WHERE users_table.user_id = evs.user_id
   LIMIT 5 
) AS foo
WHERE user_id IN
    (SELECT count(*)
     FROM users_table
     GROUP BY user_id);
```

  * Subquery joins that are not on the partition columns (i.e., each
    subquery is individually joined on partition keys but not the upper
    level subquery.). `foo` and `bar` individually OK to pushdown, but 
    not the whole query
```SQL
SELECT foo.value_2
FROM
  (SELECT users_table.value_2,
          random()
   FROM users_table,
        events_table
   WHERE users_table.user_id = events_table.user_id
     AND event_type IN (1,
                        2,
                        3,
                        4)) AS foo,

  (SELECT users_table.value_2
   FROM users_table,
        events_table
   WHERE users_table.user_id = events_table.user_id
     AND event_type IN (5,
                        6,
                        7,
                        8)) AS bar
WHERE foo.value_2 = bar.value_2;

```

  * Any limitation that logical planner applies such as aggregate 
    distinct (except for count) when GROUP BY is on non-partition key,
    or array_agg with ORDER BY
```SQL
SELECT 
 	* 
 FROM 
 (
 	SELECT avg(DISTINCT value_1), random() FROM users_table GROUP BY value_2 OFFSET 3
 ) as foo;
```
  

### An example that this PR implements ( copied from #1804):

For instance, a query for the distinct number of sessions on the top 20 most visited pages:

```SQL
SELECT
  page_id, count(distinct session_id)
FROM
  visits
WHERE
  page_id IN (SELECT page_id FROM visits GROUP BY page_id ORDER BY count(*) DESC LIMIT 20) 
GROUP BY
  page_id;
```

The LIMIT in the subquery prohibits subquery pushdown. If the table is not sharded by page_id, then the GROUP BY also prohibits subquery pushdown. However, the subquery itself can be executed as a distributed query:

```SQL
SELECT page_id FROM visits GROUP BY page_id ORDER BY count(*) DESC LIMIT 20
```

We can execute this query separately and write the results to an intermediate data structure that is made available on the workers. In the outer query, the subquery is replaced by the intermediate result, e.g.:

```SQL
SELECT
  page_id, count(distinct session_id)
FROM
  visits
WHERE
  page_id IN (SELECT * FROM intermediate_result(...) AS res(page_id int)) 
GROUP BY
  page_id;
```

Remaining tasks:

- [x] Discuss removing CTE checks from `DeferError..` functions
- [x] Discuss what to do when a subquery is safe to recursively plan via router planner but not logical planner (we currently don't recursively plan that)
- [x] Discuss adding some more debug outputs
- [x] Add regression tests for `CTE`s - subqueries mixed (modifying/recursive CTEs as well)
- [x] Add regression tests for views
- [ ] Add regression tests for UNIONs
- [x] Add regression tests for not supported cases
- [x] Add regression tests for deeper recursive calls
- [x] Add regression tests for partitioned tables
- [x] Add regression tests for more complex subqueries in WHERE clause
- [x] Add regression tests for more complex subquery JOINs, aggregates, distinct aggregates (i.e., push limits of SQL)
- [x] Add regression tests for composite keys, prepared statements
- [x] Add regression tests for querying secondaries
- [x] Test the changes on Citus Cloud

### DIFF
--- a/src/backend/distributed/planner/multi_logical_planner.c
+++ b/src/backend/distributed/planner/multi_logical_planner.c
@@ -113,7 +113,6 @@ static bool IsDistributedTableRTE(Node *node);
 static FieldSelect * CompositeFieldRecursive(Expr *expression, Query *query);
 static bool FullCompositeFieldList(List *compositeFieldList);
 static MultiNode * MultiNodeTree(Query *queryTree);
-static void ErrorIfQueryNotSupported(Query *queryTree);
 static DeferredErrorMessage * DeferredErrorIfUnsupportedRecurringTuplesJoin(
 	PlannerRestrictionContext *plannerRestrictionContext);
 static bool ShouldRecurseForRecurringTuplesJoinChecks(RelOptInfo *relOptInfo);
@@ -396,12 +395,17 @@ SubqueryMultiNodeTree(Query *originalQuery, Query *queryTree,
 {
 	MultiNode *multiQueryNode = NULL;
 	DeferredErrorMessage *subqueryPushdownError = NULL;
+	DeferredErrorMessage *unsupportedQueryError = NULL;
 
 	/*
 	 * This is a generic error check that applies to both subquery pushdown
 	 * and single table repartition subquery.
 	 */
-	ErrorIfQueryNotSupported(originalQuery);
+	unsupportedQueryError = DeferErrorIfQueryNotSupported(originalQuery);
+	if (unsupportedQueryError != NULL)
+	{
+		RaiseDeferredError(unsupportedQueryError, ERROR);
+	}
 
 	/*
 	 * In principle, we're first trying subquery pushdown planner. If it fails
@@ -916,7 +920,7 @@ DeferErrorIfFromClauseRecurs(Query *queryTree)
  * the join condition must be partition columns.
  * c. If there is a distinct clause, it must be on the partition column.
  *
- * This function is very similar to ErrorIfQueryNotSupported() in logical
+ * This function is very similar to DeferErrorIfQueryNotSupported() in logical
  * planner, but we don't reuse it, because differently for subqueries we support
  * a subset of distinct, union and left joins.
  *
@@ -1760,9 +1764,14 @@ MultiNodeTree(Query *queryTree)
 	MultiProject *projectNode = NULL;
 	MultiExtendedOp *extendedOpNode = NULL;
 	MultiNode *currentTopNode = NULL;
+	DeferredErrorMessage *unsupportedQueryError = NULL;
 
 	/* verify we can perform distributed planning on this query */
-	ErrorIfQueryNotSupported(queryTree);
+	unsupportedQueryError = DeferErrorIfQueryNotSupported(queryTree);
+	if (unsupportedQueryError != NULL)
+	{
+		RaiseDeferredError(unsupportedQueryError, ERROR);
+	}
 
 	/* extract where clause qualifiers and verify we can plan for them */
 	whereClauseList = WhereClauseList(queryTree->jointree);
@@ -2220,8 +2229,8 @@ IsReadIntermediateResultFunction(Node *node)
  * the given query. The checks in this function will be removed as we support
  * more functionality in our distributed planning.
  */
-static void
-ErrorIfQueryNotSupported(Query *queryTree)
+DeferredErrorMessage *
+DeferErrorIfQueryNotSupported(Query *queryTree)
 {
 	char *errorMessage = NULL;
 	bool hasTablesample = false;
@@ -2329,10 +2338,12 @@ ErrorIfQueryNotSupported(Query *queryTree)
 	if (!preconditionsSatisfied)
 	{
 		bool showHint = ErrorHintRequired(errorHint, queryTree);
-		ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
-						errmsg("%s", errorMessage),
-						showHint ? errhint("%s", errorHint) : 0));
+		return DeferredError(ERRCODE_FEATURE_NOT_SUPPORTED,
+							 errorMessage, NULL,
+							 showHint ? errorHint : NULL);
 	}
+
+	return NULL;
 }
 
 
@@ -3847,9 +3858,14 @@ SubqueryPushdownMultiNodeTree(Query *queryTree)
 	Query *pushedDownQuery = NULL;
 	List *subqueryTargetEntryList = NIL;
 	List *havingClauseColumnList = NIL;
+	DeferredErrorMessage *unsupportedQueryError = NULL;
 
 	/* verify we can perform distributed planning on this query */
-	ErrorIfQueryNotSupported(queryTree);
+	unsupportedQueryError = DeferErrorIfQueryNotSupported(queryTree);
+	if (unsupportedQueryError != NULL)
+	{
+		RaiseDeferredError(unsupportedQueryError, ERROR);
+	}
 
 	/*
 	 * We would be creating a new Query and pushing down top level query's
--- a/src/backend/distributed/planner/multi_logical_planner.c
+++ b/src/backend/distributed/planner/multi_logical_planner.c
@@ -80,7 +80,6 @@ typedef MultiNode *(*RuleApplyFunction) (MultiNode *leftNode, MultiNode *rightNo
 static RuleApplyFunction RuleApplyFunctionArray[JOIN_RULE_LAST] = { 0 }; /* join rules */
 
 /* Local functions forward declarations */
-static bool SingleRelationRepartitionSubquery(Query *queryTree);
 static DeferredErrorMessage * DeferErrorIfUnsupportedSubqueryPushdown(Query *
 																	  originalQuery,
 																	  PlannerRestrictionContext
@@ -98,9 +97,6 @@ static bool RangeTableArrayContainsAnyRTEIdentities(RangeTblEntry **rangeTableEn
 													queryRteIdentities);
 static Relids QueryRteIdentities(Query *queryTree);
 static DeferredErrorMessage * DeferErrorIfFromClauseRecurs(Query *queryTree);
-static DeferredErrorMessage * DeferErrorIfCannotPushdownSubquery(Query *subqueryTree,
-																 bool
-																 outerMostQueryHasLimit);
 static DeferredErrorMessage * DeferErrorIfUnsupportedUnionQuery(Query *queryTree,
 																bool
 																outerMostQueryHasLimit);
@@ -108,7 +104,6 @@ static bool ExtractSetOperationStatmentWalker(Node *node, List **setOperationLis
 static DeferredErrorMessage * DeferErrorIfUnsupportedTableCombination(Query *queryTree);
 static bool WindowPartitionOnDistributionColumn(Query *query);
 static bool AllTargetExpressionsAreColumnReferences(List *targetEntryList);
-static bool FindNodeCheckInRangeTableList(List *rtable, bool (*check)(Node *));
 static bool IsDistributedTableRTE(Node *node);
 static FieldSelect * CompositeFieldRecursive(Expr *expression, Query *query);
 static bool FullCompositeFieldList(List *compositeFieldList);
@@ -472,7 +467,7 @@ SubqueryMultiNodeTree(Query *originalQuery, Query *queryTree,
  * to ensure that Citus supports the subquery. Also, this function is designed to run
  * on the original query.
  */
-static bool
+bool
 SingleRelationRepartitionSubquery(Query *queryTree)
 {
 	List *rangeTableIndexList = NULL;
@@ -930,7 +925,7 @@ DeferErrorIfFromClauseRecurs(Query *queryTree)
  * limit, we let this query to run, but results could be wrong depending on the
  * features of underlying tables.
  */
-static DeferredErrorMessage *
+DeferredErrorMessage *
 DeferErrorIfCannotPushdownSubquery(Query *subqueryTree, bool outerMostQueryHasLimit)
 {
 	bool preconditionsSatisfied = true;
@@ -1517,7 +1512,7 @@ AllTargetExpressionsAreColumnReferences(List *targetEntryList)
  * FindNodeCheckInRangeTableList relies on FindNodeCheck() but only
  * considers the range table entries.
  */
-static bool
+bool
 FindNodeCheckInRangeTableList(List *rtable, bool (*check)(Node *))
 {
 	return range_table_walker(rtable, FindNodeCheck, check, QTW_EXAMINE_RTES);
@@ -2259,8 +2254,12 @@ DeferErrorIfQueryNotSupported(Query *queryTree)
 	if (queryTree->hasWindowFuncs)
 	{
 		preconditionsSatisfied = false;
-		errorMessage = "could not run distributed query with window functions";
-		errorHint = filterHint;
+		errorMessage = "could not run distributed query because the window "
+					   "function that is used cannot be pushed down";
+		errorHint = "Window functions are supported in two ways. Either add "
+					"an equality filter on the distributed tables' partition "
+					"column or use the window functions inside a subquery with "
+					"a PARTITION BY clause containing the distribution column";
 	}
 
 	if (queryTree->setOperations)
@@ -2642,10 +2641,12 @@ HasComplexRangeTableType(Query *queryTree)
 
 		/*
 		 * Check if the range table in the join tree is a simple relation or a
-		 * subquery.
+		 * subquery or a function. Note that RTE_FUNCTIONs are handled via (sub)query
+		 * pushdown.
 		 */
 		if (rangeTableEntry->rtekind != RTE_RELATION &&
-			rangeTableEntry->rtekind != RTE_SUBQUERY)
+			rangeTableEntry->rtekind != RTE_SUBQUERY &&
+			rangeTableEntry->rtekind != RTE_FUNCTION)
 		{
 			hasComplexRangeTableType = true;
 		}
--- a/src/backend/distributed/planner/multi_physical_planner.c
+++ b/src/backend/distributed/planner/multi_physical_planner.c
@@ -2093,7 +2093,19 @@ SubquerySqlTaskList(Job *job, PlannerRestrictionContext *plannerRestrictionConte
 	 */
 	if (targetCacheEntry == NULL)
 	{
-		RangeTblEntry *rangeTableEntry = (RangeTblEntry *) linitial(rangeTableList);
+		RangeTblEntry *rangeTableEntry = NULL;
+
+		if (list_length(rangeTableList) == 0)
+		{
+			/*
+			 * User disabled the router planner and forced planner go through
+			 * subquery pushdown, but we cannot continue anymore.
+			 */
+			ereport(ERROR, (errmsg("cannot handle complex subqueries when the "
+								   "router executor is disabled")));
+		}
+
+		rangeTableEntry = (RangeTblEntry *) linitial(rangeTableList);
 		relationId = rangeTableEntry->relid;
 		targetCacheEntry = DistributedTableCacheEntry(relationId);
 	}
--- a/src/backend/distributed/planner/recursive_planning.c
+++ b/src/backend/distributed/planner/recursive_planning.c
@@ -74,19 +74,6 @@
 
 
 /*
- * RecursivePlanningContext is used to recursively plan subqueries
- * and CTEs, pull results to the coordinator, and push it back into
- * the workers.
- */
-typedef struct RecursivePlanningContext
-{
-	int level;
-	uint64 planId;
-	List *subPlanList;
-	PlannerRestrictionContext *plannerRestrictionContext;
-} RecursivePlanningContext;
-
-/*
  * CteReferenceWalkerContext is used to collect CTE references in
  * CteReferenceListWalker.
  */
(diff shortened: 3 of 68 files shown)


src/backend/distributed/planner/recursive_planning.c
```diff
@@ -73,19 +75,6 @@
...
- * and CTEs, pull results to the coordinator, and push it back into
- * the workers.
- */
-typedef struct RecursivePlanningContext
-{
-	int level;
-	uint64 planId;
-	List *subPlanList;
-	PlannerRestrictionContext *plannerRestrictionContext;
-} RecursivePlanningContext;
```
marcocitus: Maybe just define a wrapper for `RecursivelyPlanSubqueriesAndCTEs` so we can keep this here.

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -316,6 +302,211 @@ RecursivelyPlanCTEs(Query *query, RecursivePlanningContext *planningContext)
...
+		/*
+		 * Postgres can always plan queries that don't require distributed planning.
+		 * Note that we need to check this first, otherwise the calls to the many other
+		 * Citus planner functions would error our due to local relations.
+		 */
+		shouldRecursivelyPlan = true;
+	}
+	else
+	{
+		if (DeferErrorIfCannotPushdownSubquery(subquery, false) == NULL)
```
marcocitus: Maybe use `else if`

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -316,6 +302,211 @@ RecursivelyPlanCTEs(Query *query, RecursivePlanningContext *planningContext)
...
+			 */
+		}
+	}
+
+	/*
+	 * Even if we could recursively plan the subquery, we should ensure
+	 * that the subquery doesn't contain any references to the outer
+	 * queries.
+	 */
+	if (shouldRecursivelyPlan && ContainsReferencesToOuterQuery(subquery))
```
marcocitus: Maybe move this to the start of the if-else block above.

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -316,6 +302,211 @@ RecursivelyPlanCTEs(Query *query, RecursivePlanningContext *planningContext)
...
+
+	if (rangeTableEntry->relkind == RELKIND_VIEW)
+	{
+		return false;
+	}
+
+	relationId = rangeTableEntry->relid;
+	if (!IsDistributedTable(relationId))
+	{
+		return true;
```
marcocitus: Maybe use `if (IsDistributedTable(relationId)) return false;` for consistency.

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -316,6 +302,211 @@ RecursivelyPlanCTEs(Query *query, RecursivePlanningContext *planningContext)
...
+RecursivelyPlanSubquery(Query *subquery, RecursivePlanningContext *planningContext)
+{
+	DistributedSubPlan *subPlan = NULL;
+	uint64 planId = planningContext->planId;
+	int subPlanId = 0;
+
+	Query *resultQuery = NULL;
+	Query *debugQuery =
+		(log_min_messages <= DEBUG1 || client_min_messages <= DEBUG1) ?
+		copyObject(subquery) : NULL;
```
marcocitus: This is a little hard to read, could we use an `if` statement?

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -316,6 +302,211 @@ RecursivelyPlanCTEs(Query *query, RecursivePlanningContext *planningContext)
...
+			 */
+			shouldRecursivelyPlan = false;
+		}
+		else if (TaskExecutorType == MULTI_EXECUTOR_TASK_TRACKER &&
+				 SingleRelationRepartitionSubquery(subquery))
+		{
+			/* Citus could plan this subquery through re-partitioning */
+			shouldRecursivelyPlan = false;
+		}
+		else if (DeferErrorIfQueryNotSupported(subquery) == NULL)
```
marcocitus: Apart from the error messages, would we break any queries by removing this check?

If not, what are some alternatives for getting good error messages?

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -316,6 +302,211 @@ RecursivelyPlanCTEs(Query *query, RecursivePlanningContext *planningContext)
...
+ * planned or not.
+ *
+ * For the details, see the cases in the function.
+ */
+static bool
+ShouldRecursivelyPlanSubquery(Query *subquery, RecursivePlanningContext *context)
+{
+	bool shouldRecursivelyPlan = false;
+
+	if (FindNodeCheckInRangeTableList(subquery->rtable, IsLocalTableRTE))
```
marcocitus: This might recurse too aggressively, e.g.:

```
SELECT * FROM distributed_table WHERE x IN (
  SELECT y FROM local_table JOIN (SELECT y FROM other_local_table GROUP BY y) USING (y)
);
```

We'd recursively plan the inner query on the local table, but then the outer query would break.

marcocitus:
> ```sql
> SELECT
>  page_id, count(distinct session_id)
>FROM
>  visits
>WHERE
>  page_id IN (SELECT page_id FROM visits GROUP BY page_id ORDER BY count(*) DESC LIMIT 20) 
>GROUP BY
>  page_id;
>```

This query isn't actually supported by this branch ;)

src/backend/distributed/planner/recursive_planning.c
marcocitus: We should maybe set `shouldRecursivelyPlan` to false instead and handle queries on local tables later.

I think recursive planning of subqueries is a 2-step process:
- decide if the query needs to be planned recursively based on its own properties (e.g. LIMIT)
- decide if any subqueries of the query need to be planned recursively based on the properties of the query (e.g. union, left join, subqueries on local tables in a distributed table query)

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -2091,12 +2091,21 @@ SubquerySqlTaskList(Job *job, PlannerRestrictionContext *plannerRestrictionConte
...
 	 * as an anchor table.
 	 */
-	if (targetCacheEntry == NULL)
+	if (targetCacheEntry == NULL && list_length(rangeTableList) >= 1)
 	{
 		RangeTblEntry *rangeTableEntry = (RangeTblEntry *) linitial(rangeTableList);
 		relationId = rangeTableEntry->relid;
 		targetCacheEntry = DistributedTableCacheEntry(relationId);
 	}
+	else if (targetCacheEntry == NULL && list_length(rangeTableList) < 1)
```
marcocitus: This looks slightly odd. The `list_length(rangeTableList) < 1` is redundant. 

but maybe just:

```
if (targetCacheEntry == NULL)
{
    RangeTblEntry *rangeTableEntry = NULL;

    if (list_length(rangeTableList) == 0)
    {
       // error
    }

    ...
}
```

src/backend/distributed/planner/recursive_planning.c
marcocitus: or maybe

```
if (FindNodeCheckInRangeTableList(subquery->rtable, IsLocalTableRTE))
{
    return false;
}

if (ContainsReferencesToOuterQuery(subquery))
{
    return false;
}

if (DeferErrorIfCannotPushdownSubquery(subquery, false) == NULL)
{
    return false;
}

if (TaskExecutorType == MULTI_EXECUTOR_TASK_TRACKER && SingleRelationRepartitionSubquery(subquery))
{
     return false;
}

return true;
```

src/backend/distributed/planner/recursive_planning.c
marcocitus: Nvm, at the end probably makes sense given the debug message. Maybe split this into two functions in which only the outer function emits debug messages.

src/backend/distributed/planner/recursive_planning.c
onderkalaci: When we discussed about re-partition joins in the subqueries, we thought it makes more sense to subplan them immediately at the leaf query as [here](https://github.com/citusdata/citus/compare/sql_monster#diff-1fb270ce40ed6550c916670330d5c788R1611). 

I thought this is a similar approach to that. I can see that the above query would fail. But, on the other hand, I'd like to handle local queries at the leaf level so that we don't need to adjust other parts of the planner to be safe for local tables.  So I'd like to think about more before changing this point.

onderkalaci:
> This query isn't actually supported by this branch ;)

Now it works magically 😄   I'll add more complex sql queries, in the TODO list we have `Add regression tests for more complex subquery JOINs, aggregates, distinct aggregates (i.e., push limits of SQL)`

src/backend/distributed/planner/recursive_planning.c
marcocitus: Looking into this more, we should just have a `else { shouldRecrusivePlan = true; }` here and reword the error messages.

Otherwise we get into a world of trouble with things like unions in subqueries. 
https://github.com/citusdata/citus/tree/unions

src/backend/distributed/planner/recursive_planning.c
marcocitus: > When we discussed about re-partition joins in the subqueries, we thought it makes more sense to subplan them immediately at the leaf query as here.

There's no way we could support a repartition join in a subquery in any other way than recursively planning it, so we might as well do that immediately, especially because it improves the non-co-located join code. Once we have a smarter join order planner, we would also defer this decision.

> we don't need to adjust other parts of the planner to be safe for local tables

That's unavoidable for any query that has a local table in the FROM and a subquery that contains a distributed table, so keeping local tables away from the planner is not an objective.

I'm also fine with punting on supporting subqueries with local tables for now, it's not really a requirement.

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -73,19 +75,6 @@
 #include "utils/guc.h"
 
 
-/*
- * RecursivePlanningContext is used to recursively plan subqueries
- * and CTEs, pull results to the coordinator, and push it back into
- * the workers.
- */
-typedef struct RecursivePlanningContext
```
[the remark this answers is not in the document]
marcocitus: Yep, I'd prefer that since the context is just an implementation detail that has no relevance to the caller.

marcocitus:
> Discuss adding some more debug outputs

Let's do that in a separate PR to no blow this one up further.

src/backend/distributed/planner/recursive_planning.c
onderkalaci: Ah, I thought you asked for `RecursivePlanningContext ` wrapper. Sure, it makes sense to have a wrapper for `RecursivePlanningContext `

src/test/regress/expected/multi_insert_select_window.out
```diff
@@ -709,8 +709,8 @@ SELECT * FROM
 ) as foo
 LIMIT
   10;
-ERROR:  could not run distributed query with window functions
-HINT:  Consider using an equality filter on the distributed table's partition column.
+ERROR:  could not run distributed query becuase the window function that is used cannot be pushed down
+HINT:  Window functions are supported in two ways. Either add an quality filter on the distributed tables' partition columns or use the window functions inside a subquery with PARTITION BY clause containing the distribution column
```
marcocitus: quality -> equality ?

unless you mean "add a really good filter" ;)

partition columns -> partition column

with PARTITION BY clause -> with a PARTITION BY clause

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -316,6 +351,189 @@ RecursivelyPlanCTEs(Query *query, RecursivePlanningContext *planningContext)
...
+		return false;
+	}
+
+	if (IsA(node, Query))
+	{
+		Query *query = (Query *) node;
+		DeferredErrorMessage *error = NULL;
+
+		context->level += 1;
+		error = RecursivelyPlanSubqueriesAndCTEs(query, context);
```
marcocitus: This can probably use some elaboration.

```c
/*
 * First, make sure any subqueries and CTEs within this subquery are recursively planned
 * if necessary.
 */
```
and then below maybe

```c
/*
 * Recursively plan this subquery if it cannot be pushed down and is eligible for recursive planning.
 */
```

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -316,6 +351,189 @@ RecursivelyPlanCTEs(Query *query, RecursivePlanningContext *planningContext)
...
+	 * Even if we could recursively plan the subquery, we should ensure
+	 * that the subquery doesn't contain any references to the outer
+	 * queries.
+	 */
+	if (ContainsReferencesToOuterQuery(subquery))
+	{
+		if (log_min_messages <= DEBUG2 || client_min_messages <= DEBUG2)
+		{
+			elog(DEBUG2, "skipping recursive planning for the subquery since it "
+						 "contains references to outer queries");
```
marcocitus: I wonder whether it's worth showing the query here, because it's hard to know what this is about in a complex query. (consider this optional)

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -2250,8 +2254,12 @@ ErrorIfQueryNotSupported(Query *queryTree)
 	if (queryTree->hasWindowFuncs)
 	{
 		preconditionsSatisfied = false;
-		errorMessage = "could not run distributed query with window functions";
-		errorHint = filterHint;
+		errorMessage = "could not run distributed query becuase the window "
+					   "function that is used cannot be pushed down";
+		errorHint = "Window functions are supported in two ways. Either add "
```
marcocitus: (my other comment got hidden)

quality -> equality ?

unless you mean "add a really good filter" ;)

partition columns -> partition column

with PARTITION BY clause -> with a PARTITION BY clause

marcocitus:
Correlated subqueries and lateral joins give quite poor error messages, but we should maybe address that in a separate PR:

```
SELECT * FROM test a JOIN LATERAL (SELECT * FROM test WHERE y = a.y ORDER BY x LIMIT 10) b USING (x);
DETAIL:  Limit in subquery is currently unsupported
```

marcocitus:
Also an area for error message improvement:
```
postgres=# SELECT * FROM test a JOIN (SELECT x, array_agg(y ORDER BY y) FROM test GROUP BY x) foo USING (x);
 x  |  y  | array_agg 
----+-----+-----------
  8 |  64 | {64}
  1 |   1 | {0,1}
  1 |   0 | {0,1}
 10 | 100 | {100}
  5 |  25 | {25}
  4 |  16 | {16}
  7 |  49 | {49}
  3 |   9 | {9}
  6 |  36 | {36}
  2 |   4 | {4}
  9 |  81 | {81}
(11 rows)

Time: 107.155 ms
postgres=# SELECT * FROM test a JOIN (SELECT x, array_agg(y ORDER BY y) FROM test GROUP BY x ORDER BY x DESC LIMIT 10) foo USING (x);
ERROR:  array_agg with order by is unsupported
```

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: > unless you mean "add a really good filter" ;)

😄

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -118,6 +130,38 @@ static bool ContainsReferencesToOuterQueryWalker(Node *node,
 static Query * BuildSubPlanResultQuery(Query *subquery, uint64 planId, uint32 subPlanId);
 
 
+/*
+ * SubplanListForSubqueriesAndCTEs is a wrapper around RecursivelyPlanSubqueriesAndCTEs.
+ * The function returns the subplans if necessary. For the details of when/how subplans are
+ * generated, see RecursivelyPlanSubqueriesAndCTEs()
+ */
+List *
+SubplanListForSubqueriesAndCTEs(uint64 planId, Query *originalQuery,
```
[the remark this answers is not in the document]
onderkalaci: Makes sense

src/backend/distributed/planner/recursive_planning.c
onderkalaci: I already tried that but deparsing queries with outer references fails (I guess due to `pull_var_clause` at some point while deparsing)

marcocitus: ah, makes sense

onderkalaci:
> but not sure if this is worth changing to something like complex_subquery_.

So, I'll remove `multi_recursive`? I don't want to do that before I get an OK :)

marcocitus:
> So, I'll remove multi_recursive? I don't want to do that before I get an OK :)

`s/multi_recursive_subquery/recursive_subquery/` or `s/multi_recursive_subquery/complex_subquery/`.

onderkalaci:
>  Add regression tests for UNIONs

I'm leaving UNION related tests to the next PR.

marcocitus:
Would it be worth adding `CannotPushdown` checks for some router plannable SQL features?

```sql
SELECT * FROM ref TABLESAMPLE SYSTEM(10);
 x | y 
---+---
(0 rows)

SELECT * FROM test JOIN (SELECT * FROM ref TABLESAMPLE SYSTEM(10)) ref USING (x);
ERROR:  could not run distributed query which use TABLESAMPLE
```

onderkalaci:
> Would it be worth adding CannotPushdown checks for some router plannable SQL features?

I guess those errors are coming from `DeferErrorIfQueryNotSupported ()`. The error you've mentioned seems to be the only one that we should enforce for router planner. 

We prevent queries with `hasForUpdate` for router ones, but the error message is pretty confusing:
```SQL
citus=> select *  FROM users_table WHERE user_id = 15 FOR UPDATE LIMIT 3;
ERROR:  could not run distributed query with FOR UPDATE/SHARE commands
```

Where `user_id` is already partition column (which should be distribution column). So, I think we should not print the hint for that case.

Shall I implement them here?

marcocitus:
> Shall I implement them here?

Maybe not, we have a separate issue: https://github.com/citusdata/citus/issues/1870

onderkalaci:
> Add regression tests for composite keys, prepared statements

I've already added 9 test files.

Note sure to add the above tests given that with CTEs you'd already added bunch of them.  What is your preference?

src/test/regress/expected/subquery_prepared_statements.out
```diff
@@ -0,0 +1,287 @@
...
+--------------------
+ (6,4)
+ (6,3)
+ (6,2)
+ (6,1)
+ (5,4)
+(5 rows)
+
+EXECUTE subquery_prepare_without_param;
+DEBUG:  push down of limit count: 5
```
onderkalaci: It is confusing to me that we're planning this again. @marco any insight on what's going on?

onderkalaci:
> Note sure to add the above tests given that with CTEs you'd already added bunch of them. What is your preference?

I decided to add those tests as well.

src/test/regress/expected/subquery_prepared_statements.out
marcocitus: Could be that the prepared statement cache is getting invalidated.

(I wonder whether DROP SCHEMA aggressively invalidates everything)

onderkalaci:
@marcocitus I think ready for you to have a look. We'd probably prefer merging #1884 then this one?

onderkalaci:
> @marcocitus I think ready for you to have a look. We'd probably prefer merging #1884 then this one?

Or maybe the order doesn't matter, I don't know

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -316,6 +352,201 @@ RecursivelyPlanCTEs(Query *query, RecursivePlanningContext *planningContext)
...
+
+
+/*
+ * ShouldRecursivelyPlanSubquery decides whether the input subquery should be recursively
+ * planned or not.
+ *
+ * For the details, see the cases in the function.
+ */
+static bool
+ShouldRecursivelyPlanSubquery(Query *subquery, RecursivePlanningContext *context)
```
marcocitus: context is not used in this function

PR #1878, opened 2017-12-15
Log remote commands sent via MultiClientSendQuery

marcocitus:
It's quite hard to keep track of how a query with recursively planned CTEs and subqueries is being executed because the tasks of multi-shard queries can only be logged in the planner, while the task(s) of single-shard queries can only be logged in the executor. This change logs all commands that are sent via `SendRemoteCommand`.

### DIFF
--- a/src/backend/distributed/executor/multi_client_executor.c
+++ b/src/backend/distributed/executor/multi_client_executor.c
@@ -397,7 +397,7 @@ MultiClientSendQuery(int32 connectionId, const char *query)
 	connection = ClientConnectionArray[connectionId];
 	Assert(connection != NULL);
 
-	querySent = PQsendQuery(connection->pgConn, query);
+	querySent = SendRemoteCommand(connection, query);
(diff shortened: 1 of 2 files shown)


src/backend/distributed/executor/multi_client_executor.c
```diff
@@ -396,7 +396,7 @@ MultiClientSendQuery(int32 connectionId, const char *query)
 	connection = ClientConnectionArray[connectionId];
 	Assert(connection != NULL);
 
-	querySent = PQsendQuery(connection->pgConn, query);
+	querySent = SendRemoteCommand(connection, query);
```
mtuncer: we should also change the PQsendQuery in master_citus_tools.c::ExecuteCommandsInParallelAndStoreResults

marcocitus: Fixed

mtuncer:
rebase and 🚢

PR #1883, opened 2017-12-18
Intermediate result size limitation

furkansahin:
Currently for CTEs and complex subqueries we use a function called `read_intermediate_result`. What the function does is to execute the query in CTE/complex subquery, take the results to the master and broadcast it to the workers. 
Since there are many I/O, network operations and the data size can quickly get really high, the query run-time can drastically increase for huge tables.
To avoid such a situation, we introduce a new GUC called `citus.max_intermediate_result_size` which gives the ability to limit the data traffic. The default value is `1GB`. The limit is applied per sub-plan basis which can be easily understood by running an `EXPLAIN` query.

### DIFF
--- a/src/backend/distributed/executor/multi_client_executor.c
+++ b/src/backend/distributed/executor/multi_client_executor.c
@@ -24,6 +24,7 @@
 #include "distributed/multi_server_executor.h"
 #include "distributed/placement_connection.h"
 #include "distributed/remote_commands.h"
+#include "distributed/subplan_execution.h"
 
 #include <errno.h>
 #include <unistd.h>
@@ -703,7 +704,7 @@ MultiClientQueryStatus(int32 connectionId)
 
 /* MultiClientCopyData copies data from the file. */
 CopyStatus
-MultiClientCopyData(int32 connectionId, int32 fileDescriptor)
+MultiClientCopyData(int32 connectionId, int32 fileDescriptor, uint64 *returnBytesReceived)
 {
 	MultiConnection *connection = NULL;
 	char *receiveBuffer = NULL;
@@ -734,6 +735,11 @@ MultiClientCopyData(int32 connectionId, int32 fileDescriptor)
 		/* received copy data; append these data to file */
 		int appended = -1;
 		errno = 0;
+
+		if (returnBytesReceived)
+		{
+			*returnBytesReceived += receiveLength;
+		}
 
 		appended = write(fileDescriptor, receiveBuffer, receiveLength);
 		if (appended != receiveLength)
--- a/src/backend/distributed/executor/multi_real_time_executor.c
+++ b/src/backend/distributed/executor/multi_real_time_executor.c
@@ -19,7 +19,6 @@
 #include "postgres.h"
 #include "miscadmin.h"
 
-#include <sys/stat.h>
 #include <unistd.h>
 
 #include "access/xact.h"
@@ -42,7 +41,8 @@
 
 /* Local functions forward declarations */
 static ConnectAction ManageTaskExecution(Task *task, TaskExecution *taskExecution,
-										 TaskExecutionStatus *executionStatus);
+										 TaskExecutionStatus *executionStatus,
+										 DistributedExecutionStats *executionStats);
 static bool TaskExecutionReadyToStart(TaskExecution *taskExecution);
 static bool TaskExecutionCompleted(TaskExecution *taskExecution);
 static void CancelTaskExecutionIfActive(TaskExecution *taskExecution);
@@ -83,6 +83,8 @@ MultiRealTimeExecute(Job *job)
 	bool allTasksCompleted = false;
 	bool taskCompleted = false;
 	bool taskFailed = false;
+	bool sizeLimitIsExceeded = false;
+	DistributedExecutionStats executionStats = { 0 };
 
 	List *workerNodeList = NIL;
 	HTAB *workerHash = NULL;
@@ -107,7 +109,8 @@ MultiRealTimeExecute(Job *job)
 	}
 
 	/* loop around until all tasks complete, one task fails, or user cancels */
-	while (!(allTasksCompleted || taskFailed || QueryCancelPending))
+	while (!(allTasksCompleted || taskFailed || QueryCancelPending ||
+			 sizeLimitIsExceeded))
 	{
 		uint32 taskCount = list_length(taskList);
 		uint32 completedTaskCount = 0;
@@ -137,7 +140,8 @@ MultiRealTimeExecute(Job *job)
 			}
 
 			/* call the function that performs the core task execution logic */
-			connectAction = ManageTaskExecution(task, taskExecution, &executionStatus);
+			connectAction = ManageTaskExecution(task, taskExecution, &executionStatus,
+												&executionStats);
 
 			/* update the connection counter for throttling */
 			UpdateConnectionCounter(workerNodeState, connectAction);
@@ -171,6 +175,13 @@ MultiRealTimeExecute(Job *job)
 				 */
 				MultiClientRegisterWait(waitInfo, executionStatus, connectionId);
 			}
+		}
+
+		/* in case the task has intermediate results */
+		if (CheckIfSizeLimitIsExceeded(&executionStats))
+		{
+			sizeLimitIsExceeded = true;
+			break;
 		}
 
 		/*
@@ -235,7 +246,11 @@ MultiRealTimeExecute(Job *job)
 	 * user cancellation request, we can now safely emit an error message (all
 	 * client-side resources have been cleared).
 	 */
-	if (taskFailed)
+	if (sizeLimitIsExceeded)
+	{
+		ErrorSizeLimitIsExceeded();
+	}
+	else if (taskFailed)
 	{
 		ereport(ERROR, (errmsg("failed to execute task %u", failedTaskId)));
 	}
@@ -258,7 +273,8 @@ MultiRealTimeExecute(Job *job)
  */
 static ConnectAction
 ManageTaskExecution(Task *task, TaskExecution *taskExecution,
-					TaskExecutionStatus *executionStatus)
+					TaskExecutionStatus *executionStatus,
+					DistributedExecutionStats *executionStats)
 {
 	TaskExecStatus *taskStatusArray = taskExecution->taskStatusArray;
 	int32 *connectionIdArray = taskExecution->connectionIdArray;
@@ -660,9 +676,16 @@ ManageTaskExecution(Task *task, TaskExecution *taskExecution,
 			int32 connectionId = connectionIdArray[currentIndex];
 			int32 fileDesc = fileDescriptorArray[currentIndex];
 			int closed = -1;
+			uint64 bytesReceived = 0;
 
 			/* copy data from worker node, and write to local file */
-			CopyStatus copyStatus = MultiClientCopyData(connectionId, fileDesc);
+			CopyStatus copyStatus = MultiClientCopyData(connectionId, fileDesc,
+														&bytesReceived);
+
+			if (SubPlanLevel > 0)
+			{
+				executionStats->totalIntermediateResultSize += bytesReceived;
+			}
 
 			/* if worker node will continue to send more data, keep reading */
 			if (copyStatus == CLIENT_COPY_MORE)
--- a/src/backend/distributed/executor/multi_router_executor.c
+++ b/src/backend/distributed/executor/multi_router_executor.c
@@ -97,7 +97,8 @@ static void ExtractParametersFromParamListInfo(ParamListInfo paramListInfo,
 static bool SendQueryInSingleRowMode(MultiConnection *connection, char *query,
 									 ParamListInfo paramListInfo);
 static bool StoreQueryResult(CitusScanState *scanState, MultiConnection *connection,
-							 bool failOnError, int64 *rows);
+							 bool failOnError, int64 *rows,
+							 DistributedExecutionStats *executionStats);
 static bool ConsumeQueryResult(MultiConnection *connection, bool failOnError,
 							   int64 *rows);
 
@@ -577,6 +578,7 @@ ExecuteSingleSelectTask(CitusScanState *scanState, Task *task)
 	ListCell *taskPlacementCell = NULL;
 	char *queryString = task->queryString;
 	List *relationShardList = task->relationShardList;
+	DistributedExecutionStats executionStats = { 0 };
 
 	/*
 	 * Try to run the query to completion on one placement. If the query fails
@@ -638,7 +640,14 @@ ExecuteSingleSelectTask(CitusScanState *scanState, Task *task)
 		}
 
 		queryOK = StoreQueryResult(scanState, connection, dontFailOnError,
-								   &currentAffectedTupleCount);
+								   &currentAffectedTupleCount,
+								   &executionStats);
+
+		if (CheckIfSizeLimitIsExceeded(&executionStats))
+		{
+			ErrorSizeLimitIsExceeded();
+		}
+
 		if (queryOK)
 		{
 			return;
@@ -821,7 +830,7 @@ ExecuteSingleModifyTask(CitusScanState *scanState, Task *task, bool multipleTask
 		if (!gotResults && expectResults)
 		{
 			queryOK = StoreQueryResult(scanState, connection, failOnError,
-									   &currentAffectedTupleCount);
+									   &currentAffectedTupleCount, NULL);
 		}
 		else
 		{
@@ -1156,7 +1165,7 @@ ExecuteModifyTasks(List *taskList, bool expectResults, ParamListInfo paramListIn
 				Assert(scanState != NULL);
 
 				queryOK = StoreQueryResult(scanState, connection, failOnError,
-										   &currentAffectedTupleCount);
+										   &currentAffectedTupleCount, NULL);
 			}
 			else
 			{
@@ -1342,7 +1351,8 @@ ExtractParametersFromParamListInfo(ParamListInfo paramListInfo, Oid **parameterT
  */
 static bool
 StoreQueryResult(CitusScanState *scanState, MultiConnection *connection,
-				 bool failOnError, int64 *rows)
+				 bool failOnError, int64 *rows,
+				 DistributedExecutionStats *executionStats)
 {
 	TupleDesc tupleDescriptor =
 		scanState->customScanState.ss.ps.ps_ResultTupleSlot->tts_tupleDescriptor;
@@ -1442,6 +1452,12 @@ StoreQueryResult(CitusScanState *scanState, MultiConnection *connection,
 				else
 				{
 					columnArray[columnIndex] = PQgetvalue(result, rowIndex, columnIndex);
+					if (SubPlanLevel > 0)
+					{
+						executionStats->totalIntermediateResultSize += PQgetlength(result,
+																				   rowIndex,
+																				   columnIndex);
+					}
 				}
 			}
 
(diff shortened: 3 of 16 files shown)


src/backend/distributed/executor/multi_real_time_executor.c
```diff
@@ -246,6 +264,37 @@ MultiRealTimeExecute(Job *job)
...
+	int32 fileDescriptor = BasicOpenFilePerm(filename, fileFlags, fileMode);
+
+	uint64 MaxIntermediateResultInBytes = MaxIntermediateResult * 1024L;
+
+	int statOK = fstat(fileDescriptor, &stat);
+	close(fileDescriptor);
+	if (statOK == 0 && SetResultLimit && MaxIntermediateResult > -1)
+	{
+		currentIntermediateResult += (uint64) stat.st_size;
+		if (currentIntermediateResult > MaxIntermediateResultInBytes)
```
marcocitus: If we take the file size, then we're allowing up to 32GB per subquery, so we could easily fill up a disk. What if we keep a global counter across all tasks?

furkansahin: Actually, the currentIntermediateResult is externed and it is reset when the transaction is over. So in the same transaction we continue to increase it? Isn't this the same with a global counter?

marcocitus: It seems we're currently increasing `currentIntermediateResult` by the file sizes of all tasks at every iteration of the real-time executor (every `remote_task_check_interval`).

That's fixable, but it would be simpler and much more efficient to just get the number of bytes copied from MultiClientCopyData and add that to the counter at:
https://github.com/citusdata/citus/blob/master/src/backend/distributed/executor/multi_real_time_executor.c#L665

src/backend/distributed/executor/multi_task_tracker_executor.c
```diff
@@ -244,6 +245,15 @@ MultiTaskTrackerExecute(Job *job)
...
 			Assert(execTaskTracker != NULL);
 
+			/* in case the task has intermediate results */
+			if (CheckIfSizeLimitIsExceeded())
+			{
+				failedTaskId = taskExecution->taskId;
+				taskFailed = true;
+				sizeLimitIsExceeded = true;
+				break;
+			}
```
marcocitus: If we no longer show a task ID in the `errmsg`, we could move this code below the loop and just do:

```
if (CheckIfSizeLimitIsExceeded())
{
    sizeLimitIsExceeded = true;
}
```

then the `|| sizeLimitIsExceeded` you added to the while condition will make sure we end the loop.

src/backend/distributed/executor/multi_task_tracker_executor.c
```diff
@@ -428,7 +438,14 @@ MultiTaskTrackerExecute(Job *job)
 	 * If we previously broke out of the execution loop due to a task failure or
 	 * user cancellation request, we can now safely emit an error message.
 	 */
-	if (taskFailed)
+	if (taskFailed && sizeLimitIsExceeded)
+	{
+		ereport(ERROR, (errmsg("failed to execute task %u", failedTaskId),
+						errhint("the max intermediate result size is exceeded "
```
marcocitus: Maybe move this line into `errmsg`. Most clients / ORMs / etc. only expose the error message, so `failed to execute task` is not very helpful.

(yes, we do it for other errors, and that's bad)

src/backend/distributed/executor/subplan_execution.c
```diff
@@ -32,6 +37,8 @@ ExecuteSubPlans(DistributedPlan *distributedPlan)
 	List *nodeList = ActiveReadableNodeList();
 	bool writeLocalFile = false;
 
+	SetResultLimit = (subPlanList && subPlanList->length);
```
marcocitus: maybe
```c
SetResultLimit = subPlanList != NIL;
```

src/backend/distributed/executor/multi_server_executor.c
```diff
@@ -281,3 +281,30 @@ AdjustStateForFailure(TaskExecution *taskExecution)
...
+	if (!SetResultLimit)
+	{
+		return false;
+	}
+
+	MaxIntermediateResultInBytes = MaxIntermediateResult * 1024L;
+
+	if (MaxIntermediateResult > -1)
+	{
+		if (currentIntermediateResult > MaxIntermediateResultInBytes)
```
marcocitus: `MaxIntermediateResult > -1 && currentIntermediateResult > MaxIntermediateResultInBytes`

src/backend/distributed/executor/subplan_execution.c
```diff
@@ -19,6 +19,11 @@
 #include "executor/executor.h"
 
 
+int MaxIntermediateResult = 1048576; /* maximum size in KB the intermediate result can grow to */
+uint64 currentIntermediateResult = 0;
+bool SetResultLimit = false;
```
marcocitus: this sounds a bit like a function name, maybe `UseResultSizeLimit`.

src/backend/distributed/executor/subplan_execution.c
```diff
@@ -19,6 +19,11 @@
 #include "executor/executor.h"
 
 
+int MaxIntermediateResult = 1048576; /* maximum size in KB the intermediate result can grow to */
+uint64 currentIntermediateResult = 0;
```
marcocitus: since this contains a size and it's the total size across all subplans, maybe `TotalIntermediateResultSize`

src/backend/distributed/executor/multi_client_executor.c
```diff
@@ -735,6 +736,11 @@ MultiClientCopyData(int32 connectionId, int32 fileDescriptor)
 		int appended = -1;
 		errno = 0;
 
+		if (SetResultLimit)
+		{
+			currentIntermediateResult += receiveLength;
```
marcocitus: `MultiClientCopyData ` is used for things which have nothing do with `currentIntermediateResult`

It'd be nicer to add a `int* returnBytesReceived` parameter and do something like:
```
if (returnBytesReceived != NULL)
{
    *returnBytesReceived = receiveLength;
}
```

src/backend/distributed/executor/multi_real_time_executor.c
```diff
@@ -136,6 +137,15 @@ MultiRealTimeExecute(Job *job)
 				continue;
 			}
 
+			/* in case the task has intermediate results */
+			if (CheckIfSizeLimitIsExceeded())
+			{
+				failedTaskId = taskExecution->taskId;
+				taskFailed = true;
+				sizeLimitIsExceeded = true;
+				break;
```
marcocitus: We should be able to move this out of the while loop and omit the `failedTaskId = ` and `taskFailed` part.

src/backend/distributed/executor/multi_server_executor.c
```diff
@@ -281,3 +281,23 @@ AdjustStateForFailure(TaskExecution *taskExecution)
...
+	uint64 MaxIntermediateResultInBytes = 0;
+	if (!UseResultSizeLimit)
+	{
+		return false;
+	}
+
+	MaxIntermediateResultInBytes = MaxIntermediateResult * 1024L;
+
+	return (MaxIntermediateResult > -1 && TotalIntermediateResultSize >
+			MaxIntermediateResultInBytes);
```
marcocitus: looking at this again, could also structure it like:

```
if (MaxIntermediateResult == -1)
{
    return false;
}

MaxIntermediateResultInBytes = MaxIntermediateResult * 1024L;
if (TotalIntermediateResultSize < MaxIntermediateResultInBytes)
{
    return false;
}

return true;
```

we tend to be a little verbose about if-statements to make it easier to set breakpoints in gdb or add log output.

src/backend/distributed/executor/multi_server_executor.c
```diff
@@ -281,3 +281,23 @@ AdjustStateForFailure(TaskExecution *taskExecution)
...
+
+
+/*
+ * CheckIfSizeLimitIsExceeded checks if the limit is exceeded by intermediate
+ * results, if there is any.
+ */
+bool
+CheckIfSizeLimitIsExceeded()
+{
+	uint64 MaxIntermediateResultInBytes = 0;
```
marcocitus: we generally use lower case for local variables `maxIntermediateResultInBytes`

src/backend/distributed/shared_library_init.c
```diff
@@ -547,6 +548,17 @@ RegisterCitusConfigVariables(void)
 		GUC_UNIT_KB,
 		NULL, NULL, NULL);
 
+	DefineCustomIntVariable(
+		"citus.max_intermediate_result_size",
+		gettext_noop(
+			"Sets the maximum size of the intermediate results during the execution."),
+		NULL,
+		&MaxIntermediateResult,
+		1048576, -1, INT_MAX, /* max allowed size not set to MAX_KILOBYTES on purpose */
```
marcocitus: which purpose?

furkansahin: This part was copy-pasted from shard_max_size for the time being. I don't know the reason why they chose not to use it. I'd like to use MAX_KILOBYTES here. Do you see any problem with it?

marcocitus: I don't

src/backend/distributed/executor/subplan_execution.c
```diff
@@ -32,6 +37,10 @@ ExecuteSubPlans(DistributedPlan *distributedPlan)
 	List *nodeList = ActiveReadableNodeList();
 	bool writeLocalFile = false;
 
+
+	UseResultSizeLimit = UseResultSizeLimit || (subPlanList != NIL &&
```
marcocitus: Maybe use the verbose form (makes it easier to set a breakpoint in gdb).

```
if (subPlanList != NIL && MaxIntermediateResult >= 0)
{
    UseResultSizeLimit = true;
}
```

src/backend/distributed/executor/multi_real_time_executor.c
```diff
@@ -660,9 +674,16 @@ ManageTaskExecution(Task *task, TaskExecution *taskExecution,
 			int32 connectionId = connectionIdArray[currentIndex];
 			int32 fileDesc = fileDescriptorArray[currentIndex];
 			int closed = -1;
+			uint64 returnBytesReceived = 0;
```
marcocitus: Minor: Maybe `returnBytesReceived` -> `bytesReceived`. The `return` makes more sense within `MultiClientCopyData` where it is like a return value.

furkansahin:
>Open question: Should we also reset UseResultSizeLimit after finishing the subplans as to not restrict the final result size?
I guess, we automatically switched to this after making the limit per sub-plan

src/backend/distributed/executor/multi_real_time_executor.c
```diff
@@ -235,7 +246,13 @@ MultiRealTimeExecute(Job *job)
 	 * user cancellation request, we can now safely emit an error message (all
 	 * client-side resources have been cleared).
 	 */
-	if (taskFailed)
+	if (sizeLimitIsExceeded)
+	{
+		ereport(ERROR, (errmsg("the max intermediate result size is exceeded, "
+							   "consider to set citus.max_intermediate_result_size to "
```
marcocitus: We can probably share a bit more information.

```
ERROR: the intermediate result size exceeded citus.max_intermediate_result_size (currently 1048576KB)
DETAIL: Citus restricts the size of intermediate results of complex subqueries and CTEs to avoid accidentally pulling large result sets into once place.
HINT: To run the current query, set citus.max_intermediate_result_size to a higher value or -1 to disable.
```

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -534,6 +536,7 @@ RouterSelectExecScan(CustomScanState *node)
 {
 	CitusScanState *scanState = (CitusScanState *) node;
 	TupleTableSlot *resultSlot = NULL;
+	uint64 totalIntermediateResultSize = 0;
```
marcocitus: Would it make sense to keep this counter in `ExecuteSingleSelectTask`? It seems to mimic `currentAffectedTupleCount`.

src/backend/distributed/executor/subplan_execution.c
```diff
@@ -19,6 +19,10 @@
 #include "executor/executor.h"
 
 
+int MaxIntermediateResult = 1048576; /* maximum size in KB the intermediate result can grow to */
+bool UseResultSizeLimit = false;
```
marcocitus: Maybe add a comment that when this is true, we enforce intermediate result size limit in all executors.

src/backend/distributed/executor/multi_task_tracker_executor.c
```diff
@@ -96,7 +97,8 @@ static TaskExecStatus ManageTaskExecution(TaskTracker *taskTracker,
 										  Task *task, TaskExecution *taskExecution);
 static TransmitExecStatus ManageTransmitExecution(TaskTracker *transmitTracker,
 												  Task *task,
-												  TaskExecution *taskExecution);
+												  TaskExecution *taskExecution,
+												  uint64 *totalIntermediateResultSize);
```
marcocitus: Consider this optional: It'd be kind of nice if we passed down some more generic data structure (e.g. `DistributedExecutionStats *stats`) in which we keep the `totalIntermediateResultSize`, such that we can easily add other counters in the future without having to change signatures everywhere.

src/backend/distributed/executor/multi_real_time_executor.c
```diff
@@ -107,7 +109,8 @@ MultiRealTimeExecute(Job *job)
 	}
 
 	/* loop around until all tasks complete, one task fails, or user cancels */
-	while (!(allTasksCompleted || taskFailed || QueryCancelPending))
+	while (!(allTasksCompleted || taskFailed || QueryCancelPending ||
+			 sizeLimitIsExceeded))
```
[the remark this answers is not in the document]
furkansahin: For the sake of consistency, I want to keep this check in while && (add `continue;` instead of `break;` || just remove `break;`) 
Which one do you suggest?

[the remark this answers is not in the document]
marcocitus: I like `break;` because it makes it obvious that we immediately leave the loop. Could also live with just leaving both the `break` and the `|| sizeLimitIsExceeded` here.

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -1340,7 +1359,8 @@ ExtractParametersFromParamListInfo(ParamListInfo paramListInfo, Oid **parameterT
  */
 static bool
 StoreQueryResult(CitusScanState *scanState, MultiConnection *connection,
-				 bool failOnError, int64 *rows)
+				 bool failOnError, int64 *rows,
+				 DistributedExecutionStats *executionStats)
```
marcocitus: We could consider putting rows in executionStats as well.

src/backend/distributed/executor/multi_real_time_executor.c
```diff
@@ -83,6 +83,9 @@ MultiRealTimeExecute(Job *job)
 	bool allTasksCompleted = false;
 	bool taskCompleted = false;
 	bool taskFailed = false;
+	bool sizeLimitIsExceeded = false;
+	DistributedExecutionStats *executionStats = palloc0(
+		sizeof(DistributedExecutionStats));
```
marcocitus: You could also consider allocating this on the stack:

```c
DistributedExecutionStats executionStats = { 0 };
```
and then passing it by pointer to `ManageTaskExecution(..., &executionStats)`

src/backend/distributed/executor/multi_router_executor.c
furkansahin: I can do it but should we do it in this PR as a separate commit or does it deserve its own PR?

src/backend/distributed/executor/subplan_execution.c
```diff
@@ -43,13 +48,22 @@ ExecuteSubPlans(DistributedPlan *distributedPlan)
 
 		char *resultId = GenerateResultId(planId, subPlanId);
 
+		if (MaxIntermediateResult >= 0)
+		{
+			SubPlanLevel++;
```
marcocitus: I think this variable might have other uses and increasing/decreasing it costs almost nothing. Maybe remove the `if`

src/backend/distributed/transaction/transaction_management.c
```diff
@@ -213,6 +214,7 @@ CoordinatedTransactionCallback(XactEvent event, void *arg)
 			XactModificationLevel = XACT_MODIFICATION_NONE;
 			dlist_init(&InProgressTransactions);
 			CoordinatedTransactionUses2PC = false;
+			SubPlanLevel = 0;
```
marcocitus: We've not done a good job of describing what we're doing in this function. Could you add a comment here saying what `SubPlanLevel` is and why we need to reset it?

src/backend/distributed/executor/multi_task_tracker_executor.c
```diff
@@ -428,7 +441,20 @@ MultiTaskTrackerExecute(Job *job)
...
+							   "citus.max_intermediate_result_size (currently %d kB)",
+							   MaxIntermediateResult),
+						errdetail("Citus restricts the size of intermediate "
+								  "results of complex subqueries and CTEs to "
+								  "avoid accidentally pulling large result sets "
+								  "into once place."),
+						errhint("To run the current query, set "
+								"citus.max_intermediate_result_size to a higher"
+								" value or -1 to disable.")));
+	}
```
marcocitus: Maybe move this error message into a separate function.

src/backend/distributed/executor/multi_task_tracker_executor.c
```diff
@@ -1397,12 +1424,20 @@ ManageTransmitExecution(TaskTracker *transmitTracker,
...
 			int32 connectionId = TransmitTrackerConnectionId(transmitTracker, task);
 			Assert(connectionId != INVALID_CONNECTION_ID);
 
-			copyStatus = MultiClientCopyData(connectionId, fileDescriptor);
+			copyStatus = MultiClientCopyData(connectionId, fileDescriptor,
+											 &bytesReceived);
+
+			if (SubPlanLevel > 0)
+			{
+				(executionStats->totalIntermediateResultSize) += bytesReceived;
```
marcocitus: why the parentheses?

src/backend/distributed/executor/multi_server_executor.c
```diff
@@ -281,3 +281,43 @@ AdjustStateForFailure(TaskExecution *taskExecution)
...
+	{
+		return false;
+	}
+
+	return true;
+}
+
+
+void
+ErrorSizeLimitIsExceeded()
```
marcocitus: needs a function comment

src/backend/distributed/executor/multi_real_time_executor.c
```diff
@@ -660,9 +676,16 @@ ManageTaskExecution(Task *task, TaskExecution *taskExecution,
...
 			int32 fileDesc = fileDescriptorArray[currentIndex];
 			int closed = -1;
+			uint64 bytesReceived = 0;
 
 			/* copy data from worker node, and write to local file */
-			CopyStatus copyStatus = MultiClientCopyData(connectionId, fileDesc);
+			CopyStatus copyStatus = MultiClientCopyData(connectionId, fileDesc,
+														&bytesReceived);
+
+			if (SubPlanLevel > 0 && bytesReceived != 0)
```
marcocitus: The `bytesReceived != 0` looks a bit redundant (the comparison is probably about as expensive as the addition)

src/backend/distributed/shared_library_init.c
```diff
@@ -547,6 +548,17 @@ RegisterCitusConfigVariables(void)
 		GUC_UNIT_KB,
 		NULL, NULL, NULL);
 
+	DefineCustomIntVariable(
+		"citus.max_intermediate_result_size",
+		gettext_noop(
+			"Sets the maximum size of the intermediate results during the execution."),
```
marcocitus: maybe add " in KB"

furkansahin:
This PR is documented, rebased, stashed to 1 commit and approved. After Travis passes, I'm merging?

PR #1884, opened 2017-12-18
Fix join clause eq restrictions

### LINKED ISSUE #58: We currently require all join-clauses between tables to be equi-joins. Thus if I write something like: WHERE table1.id = table2.id AND table1.date < table2.date, this fails. We should allow the ' join if we have an equality join in an AND clause.

mtuncer:
We used to error out if the join clause includes filters like t1.a < t2.a even if other filter like t1.key = t2.key exists.

Recently we lifted that restriction in subquery planning by not lifting that restriction and focusing on equivalance classes provided by postgres.

This checkin forwards previously erroring out real-time queries due to join clauses to subquery planner and let it handle the join even if the query does not have a subquery.

Additionally, since we are now pushing down queries that do not have any subqueries in it. Error message looked misleading, changed error message to state query pushdown is failed.

Fixes #58

### DIFF
--- a/src/backend/distributed/planner/multi_logical_planner.c
+++ b/src/backend/distributed/planner/multi_logical_planner.c
@@ -128,7 +128,7 @@ static bool IsRecurringRTE(RangeTblEntry *rangeTableEntry,
 static bool IsRecurringRangeTable(List *rangeTable, RecurringTuplesType *recurType);
 static bool HasRecurringTuples(Node *node, RecurringTuplesType *recurType);
 static bool IsReadIntermediateResultFunction(Node *node);
-static DeferredErrorMessage * ValidateClauseList(List *clauseList);
+static DeferredErrorMessage * DeferErrorIfUnsupportedClause(List *clauseList);
 static bool ExtractFromExpressionWalker(Node *node,
 										QualifierWalkerContext *walkerContext);
 static List * MultiTableNodeList(List *tableEntryList, List *rangeTableList);
@@ -259,8 +259,12 @@ ShouldUseSubqueryPushDown(Query *originalQuery, Query *rewrittenQuery)
 		return true;
 	}
 
+	/*
+	 * Some unsupported join clauses in logical planner
+	 * may be supported by subquery pushdown planner.
+	 */
 	whereClauseList = WhereClauseList(rewrittenQuery->jointree);
-	if (ValidateClauseList(whereClauseList) != NULL)
+	if (DeferErrorIfUnsupportedClause(whereClauseList) != NULL)
 	{
 		return true;
 	}
@@ -568,11 +572,9 @@ DeferErrorIfUnsupportedSubqueryPushdown(Query *originalQuery,
 	else if (!RestrictionEquivalenceForPartitionKeys(plannerRestrictionContext))
 	{
 		return DeferredError(ERRCODE_FEATURE_NOT_SUPPORTED,
-							 "cannot pushdown the subquery since all relations are not "
-							 "joined using distribution keys",
-							 "Each relation should be joined with at least "
-							 "one another relation using distribution keys and "
-							 "equality operator.", NULL);
+							 "complex joins are only supported when all distributed tables are "
+							 "joined on their distribution columns with equal operator",
+							 NULL, NULL);
 	}
 
 	/* we shouldn't allow reference tables in the FROM clause when the query has sublinks */
@@ -1778,7 +1780,7 @@ MultiNodeTree(Query *queryTree)
 
 	/* extract where clause qualifiers and verify we can plan for them */
 	whereClauseList = WhereClauseList(queryTree->jointree);
-	unsupportedQueryError = ValidateClauseList(whereClauseList);
+	unsupportedQueryError = DeferErrorIfUnsupportedClause(whereClauseList);
 	if (unsupportedQueryError)
 	{
 		RaiseDeferredErrorInternal(unsupportedQueryError, ERROR);
@@ -2754,12 +2756,15 @@ QualifierList(FromExpr *fromExpr)
 
 
 /*
- * ValidateClauseList walks over the given list of clauses, and checks that we
- * can recognize all the clauses. This function ensures that we do not drop an
- * unsupported clause type on the floor, and thus prevents erroneous results.
+ * DeferErrorIfUnsupportedClause walks over the given list of clauses, and
+ * checks that we can recognize all the clauses. This function ensures that
+ * we do not drop an unsupported clause type on the floor, and thus prevents
+ * erroneous results.
+ *
+ * Returns a deferred error, caller is responsible for raising the error.
  */
 static DeferredErrorMessage *
-ValidateClauseList(List *clauseList)
+DeferErrorIfUnsupportedClause(List *clauseList)
 {
 	ListCell *clauseCell = NULL;
 	foreach(clauseCell, clauseList)
@@ -2920,8 +2925,8 @@ IsJoinClause(Node *clause)
 	List *argumentList = NIL;
 	Node *leftArgument = NULL;
 	Node *rightArgument = NULL;
-	List *leftColumnList = NIL;
-	List *rightColumnList = NIL;
+	Node *strippedLeftArgument = NULL;
+	Node *strippedRightArgument = NULL;
 
 	if (!IsA(clause, OpExpr))
 	{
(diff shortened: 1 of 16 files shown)


src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -561,7 +569,7 @@ DeferErrorIfUnsupportedSubqueryPushdown(Query *originalQuery,
 	else if (!RestrictionEquivalenceForPartitionKeys(plannerRestrictionContext))
 	{
 		return DeferredError(ERRCODE_FEATURE_NOT_SUPPORTED,
-							 "cannot pushdown the subquery since all relations are not "
+							 "cannot pushdown the query since all relations are not "
 							 "joined using distribution keys",
```
marcocitus: Maybe we should rephrase this message entirely, "all relations are not joined using distribution keys" is also grammatically incorrect. 

Maybe:

```
ERROR: complex joins are only supported when all distributed tables are joined on their partition columns
```

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1760,13 +1768,15 @@ MultiNodeTree(Query *queryTree)
...
+	DeferredErrorMessage *deferredError PG_USED_FOR_ASSERTS_ONLY = NULL;
 
 	/* verify we can perform distributed planning on this query */
 	ErrorIfQueryNotSupported(queryTree);
 
 	/* extract where clause qualifiers and verify we can plan for them */
 	whereClauseList = WhereClauseList(queryTree->jointree);
-	ValidateClauseList(whereClauseList);
+	deferredError = ValidateClauseList(whereClauseList);
+	Assert(deferredError == NULL);
```
marcocitus: `MultiNodeTree` is called recursively. Are we sure we can never ever reach it when there is an unsupported clause type?

src/test/regress/expected/multi_subquery.out
```diff
@@ -152,6 +152,147 @@ ERROR:  cannot push down this subquery
...
+
+-- count(distinct) queries work
+SELECT COUNT(DISTINCT l_orderkey)
+FROM
+	lineitem_subquery l
+JOIN
+	orders_subquery o
+ON (l_orderkey = o_orderkey)
+WHERE
+	(o_orderkey < l_quantity);
```
marcocitus: it might be worth trying a few different cases that `IsJoinClause` prevents,

e.g. `o_orderkey - l_orderkey > 0`, `o_orderkey + 1 = l_orderkey`

mtuncer: I am hitting
```
ERROR:  cannot perform local joins that involve expressions
DETAIL:  local joins can be performed between columns only
```

When there is a single join clause (i.e. `o_orderkey + 1 = l_orderkey`) whereas it throws another error message if we have (`o_orderkey + 1 = l_orderkey and o_orderkey < l_quantity`)
```
ERROR:  complex joins are only supported when all distributed tables are joined on their partition columns
```

marcocitus: Should we push down those cases as well?

mtuncer: done

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -134,7 +134,7 @@ static bool IsRecurringRTE(RangeTblEntry *rangeTableEntry,
 static bool IsRecurringRangeTable(List *rangeTable, RecurringTuplesType *recurType);
 static bool HasRecurringTuples(Node *node, RecurringTuplesType *recurType);
 static bool IsReadIntermediateResultFunction(Node *node);
-static void ValidateClauseList(List *clauseList);
+static DeferredErrorMessage * ValidateClauseList(List *clauseList);
```
onderkalaci: Minor: We have an unwritten rule of prefixing with `DeferError..`. So maybe `DeferErrorIfNotValidClauseList`?

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -263,6 +265,12 @@ ShouldUseSubqueryPushDown(Query *originalQuery, Query *rewrittenQuery)
 		return true;
 	}
 
+	whereClauseList = WhereClauseList(rewrittenQuery->jointree);
```
onderkalaci: Needs a comment/justification why we're allowing this through the pushdown planner. Something like in the commit message.

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -2895,8 +2907,10 @@ IsJoinClause(Node *clause)
 	List *argumentList = NIL;
 	Node *leftArgument = NULL;
 	Node *rightArgument = NULL;
-	List *leftColumnList = NIL;
-	List *rightColumnList = NIL;
+	Node *strippedLeftArgument = NULL;
+	Node *strippedRightArgument = NULL;
+	NodeTag leftArgumentType = T_Invalid;
```
onderkalaci: Minor: We probably don't need `leftArgumentType ` and `rightArgumentType `, which makes it slightly harder to follow. Can't we move `nodeTag(strippedLeftArgument)` and `nodeTag(strippedRightArgument)` to the if statements instead?

src/test/regress/expected/multi_subquery.out
onderkalaci: > Also maybe some kind of function(o.a,o.b)
 
Agreed. 

Also, maybe add explicit coercions as well such as (o_orderkey::bigint < l_quantity::int)

onderkalaci:
> s. Besides, we'd probably want to use distribution columns instead of partition columns in the message

And also mention that the join should be an equi-join.

src/test/regress/expected/multi_subquery.out
onderkalaci: For the joins check an example here: https://github.com/citusdata/citus/pull/1827/files#diff-0dcc934517678392319b744e5aa005c9R2043  
The first query is not supported, the next one is supported.

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -561,11 +569,9 @@ DeferErrorIfUnsupportedSubqueryPushdown(Query *originalQuery,
 	else if (!RestrictionEquivalenceForPartitionKeys(plannerRestrictionContext))
 	{
 		return DeferredError(ERRCODE_FEATURE_NOT_SUPPORTED,
-							 "cannot pushdown the subquery since all relations are not "
-							 "joined using distribution keys",
-							 "Each relation should be joined with at least "
-							 "one another relation using distribution keys and "
-							 "equality operator.", NULL);
+							 "complex joins are only supported when all distributed tables are "
```
onderkalaci: As I wrote somewhere else, we should probably want to use `distribution column` instead of `partition columns` to prevent any confusion Postgres partitioning vs Citus.

And, we'd also want to mention that the joins should be `equi joins / equality operator` etc.

src/backend/distributed/planner/multi_logical_planner.c
mtuncer: added

src/backend/distributed/planner/multi_logical_planner.c

src/backend/distributed/planner/multi_logical_planner.c
mtuncer: replaced them with
```c
if (IsA(strippedLeftArgument, Var) && IsA(strippedRightArgument, Var))
```

src/test/regress/expected/multi_subquery.out
mtuncer: adding that does not provide any additional value, but added.

src/test/regress/sql/multi_subquery.sql
```diff
@@ -145,6 +145,126 @@ FROM
...
+
+-- the same queries returning a non-partition column
+SELECT l_quantity
+FROM
+	lineitem_subquery l
+JOIN
+	orders_subquery o
+ON (l_orderkey = o_orderkey)
+WHERE
+	(o_orderkey < l_quantity)
```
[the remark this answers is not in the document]
mtuncer: done

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: `recursive planner` should be `pushdown planner` I think.

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -560,11 +572,9 @@ DeferErrorIfUnsupportedSubqueryPushdown(Query *originalQuery,
...
+							 "joined on their distribution columns with equal operator",
```
marcocitus: could we remove the "with equal operator" part? We also support joins on operators like IN, and it seems grammatically incorrect.

PR #1889, opened 2017-12-19
Do not reinitialise MyBackendData on cache invalidations

marcocitus:
We currently reinitialise `MyBackendData` from `CitusHasBeenLoaded()` after a cache invalidation of `pg_dist_partition`, which may cause the coordinator to lose its distributed transaction ID and assign a 0 transaction ID to new connections, which may affect the deadlock detection and affects our ability to read intermediate results on the workers.

### DIFF
--- a/src/backend/distributed/transaction/backend_data.c
+++ b/src/backend/distributed/transaction/backend_data.c
@@ -401,27 +401,39 @@ BackendManagementShmemSize(void)
 
 
 /*
- *  InitializeBackendData is called per backend and does the
- *  required initialization.
+ * InitializeBackendData initialises MyBackendData to the shared memory segment
+ * belonging to the current backend.
+ *
+ * The function is called through CitusHasBeenLoaded when we first detect that
+ * the Citus extension is present, and after any subsequent invalidation of
+ * pg_dist_partition (see InvalidateMetadataSystemCache()).
+ *
+ * We only need to initialise MyBackendData once. The only goal here is to make
+ * sure that we don't use the backend data from a previous backend with the same
+ * pgprocno. Resetting the backend data after a distributed transaction happens
+ * on COMMIT/ABORT through transaction callbacks.
  */
 void
 InitializeBackendData(void)
 {
+	if (MyBackendData != NULL)
+	{
+		/*
+		 * We already initialized MyBackendData before. We definitely should
+		 * not initialise it again, because we might be in the middle of a
+		 * distributed transaction.
+		 */
+		return;
+	}
+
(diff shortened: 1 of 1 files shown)


src/backend/distributed/transaction/backend_data.c
```diff
@@ -401,27 +401,39 @@ BackendManagementShmemSize(void)
...
+ *
+ * We only need to initialise MyBackendData once. The only goal here is to make
+ * sure that we don't use the backend data from a previous backend with the same
+ * pgprocno. Resetting the backend data after a distributed transaction happens
+ * on COMMIT/ABORT through transaction callbacks.
  */
 void
 InitializeBackendData(void)
 {
+	if (MyBackendData != NULL)
```
onderkalaci: It might have led us skipping some distributed deadlocks :/ 

`InitializeMaintenanceDaemonBackend` already has a similar logic to skip starting the daemon when called for the second time.

src/test/regress/specs/isolation_distributed_transaction_id.spec
```diff
@@ -57,6 +62,11 @@ step "s2-assign-transaction-id"
     SELECT assign_distributed_transaction_id(2, 2, '2015-01-02 00:00:00+0');
 }
 
+step "s2-vacuum"
+{
+	VACUUM FULL pg_dist_partition;
```
onderkalaci: Verified that this is failing on master.

What other commands can lead to this situation? I'd like to check that the issue we're seeing in the travis outputs will be solved by this. I only see create/drop schema-view-type-tables in the concurrent tests, which doesn't seem to lead the issue?

marcocitus: It seems it's usually triggered by autovacuum.

src/backend/distributed/transaction/backend_data.c
```diff
@@ -401,27 +401,39 @@ BackendManagementShmemSize(void)
...
 
 /*
- *  InitializeBackendData is called per backend and does the
- *  required initialization.
+ * InitializeBackendData initialises MyBackendData to the shared memory segment
+ * belonging to the current backend.
+ *
+ * The function is called through CitusHasBeenLoaded when we first detect that
+ * the Citus extension is present, and after any subsequent invalidation of
+ * the metadata.
```
onderkalaci: So, `invalidation of the metadata.` seems not accurate? That could happen on `drop extension` - `alter pg_dist_partition` or `vacuum pg_dist_partition` or any others?

PR #1893, opened 2017-12-19
Fix CTE column alias issue

### LINKED ISSUE #1885: Using column aliases causes warnings and errors
Citus does not support column aliases in any of its planners and gives weird warnings and errors instead.

Router:
```sql
SELECT a FROM test t(a,b) WHERE a = 1;
WARNING:  column "a" does not exist
ERROR:  could not receive query results
```

Real-time:
```
SELECT a FROM test t(a,b);
WARNING:  column "a" does not exist
WARNING:  column "a" does not exist
WARNING:  column "a" does not exist
WARNING:  column "a" does not exist
WARNING:  column "a" does not exist
WARNING:  column "a" does not exist
WARNING:  column "a" does not exist
WARNING:  column "a" does not exist
ERROR:  failed to execute task 50
```

Recursive:
```
WITH foo(a,b) AS (SELECT * FROM test) SELECT * FROM foo;
WARNING:  column "a" does not exist
ERROR:  could not receive query results
```

mtuncer:
We were creating intermediate query result's target names from subquery target list. Now we also check if cte re-defines its column name aliases and create intermediate result query accordingly.

Partially fixes #1885 for recursive planner (CTE)

### DIFF
--- a/src/backend/distributed/planner/recursive_planning.c
+++ b/src/backend/distributed/planner/recursive_planning.c
@@ -126,7 +126,8 @@ static bool CteReferenceListWalker(Node *node, CteReferenceWalkerContext *contex
 static bool ContainsReferencesToOuterQuery(Query *query);
 static bool ContainsReferencesToOuterQueryWalker(Node *node,
 												 VarLevelsUpWalkerContext *context);
-static Query * BuildSubPlanResultQuery(Query *subquery, uint64 planId, uint32 subPlanId);
+static Query * BuildSubPlanResultQuery(Query *subquery, List *columnAliasList,
+									   uint64 planId, uint32 subPlanId);
 
 
 /*
@@ -295,7 +296,8 @@ RecursivelyPlanCTEs(Query *query, RecursivePlanningContext *planningContext)
 		planningContext->subPlanList = lappend(planningContext->subPlanList, subPlan);
 
 		/* replace references to the CTE with a subquery that reads results */
-		resultQuery = BuildSubPlanResultQuery(subquery, planId, subPlanId);
+		resultQuery = BuildSubPlanResultQuery(subquery, cte->aliascolnames, planId,
+											  subPlanId);
 
 		foreach(rteCell, context.cteReferenceList)
 		{
@@ -528,7 +530,11 @@ RecursivelyPlanSubquery(Query *subquery, RecursivePlanningContext *planningConte
 	subPlan = CreateDistributedSubPlan(subPlanId, subquery);
 	planningContext->subPlanList = lappend(planningContext->subPlanList, subPlan);
 
-	resultQuery = BuildSubPlanResultQuery(subquery, planId, subPlanId);
+	/*
+	 * BuildSubPlanResultQuery() can optionally use provided column aliases.
+	 * We do not need to send additional alias list for subqueries.
+	 */
+	resultQuery = BuildSubPlanResultQuery(subquery, NIL, planId, subPlanId);
 
 	if (log_min_messages <= DEBUG1 || client_min_messages <= DEBUG1)
 	{
@@ -712,13 +718,15 @@ ContainsReferencesToOuterQueryWalker(Node *node, VarLevelsUpWalkerContext *conte
  *   read_intermediate_result('<planId>_<subPlanId>', '<copy format'>)
  *   AS res (<column definition list>);
  *
- * The target list and column definition list are derived from the given subquery.
+ * The target list and column definition list are derived from the given subquery
+ * and columm name alias list.
  *
  * If any of the types in the target list cannot be used in the binary copy format,
  * then the copy format 'text' is used, otherwise 'binary' is used.
  */
 static Query *
-BuildSubPlanResultQuery(Query *subquery, uint64 planId, uint32 subPlanId)
+BuildSubPlanResultQuery(Query *subquery, List *columnAliasList, uint64 planId,
+						uint32 subPlanId)
 {
 	Query *resultQuery = NULL;
 	char *resultIdString = NULL;
@@ -739,6 +747,7 @@ BuildSubPlanResultQuery(Query *subquery, uint64 planId, uint32 subPlanId)
 	int columnNumber = 1;
 	bool useBinaryCopyFormat = true;
 	Oid copyFormatId = BinaryCopyFormatId();
+	int columnAliasCount = list_length(columnAliasList);
 
 	/* build the target list and column definition list */
 	foreach(targetEntryCell, subquery->targetList)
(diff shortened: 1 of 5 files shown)


src/backend/distributed/planner/recursive_planning.c
```diff
@@ -482,13 +483,15 @@ ContainsReferencesToOuterQueryWalker(Node *node, VarLevelsUpWalkerContext *conte
...
- * The target list and column definition list are derived from the given subquery.
+ * The target list and column definition list are derived from the given subquery
+ * and the cte columm name alias list.
  *
  * If any of the types in the target list cannot be used in the binary copy format,
  * then the copy format 'text' is used, otherwise 'binary' is used.
  */
 static Query *
-BuildSubPlanResultQuery(Query *subquery, uint64 planId, uint32 subPlanId)
+BuildSubPlanResultQuery(Query *subquery, CommonTableExpr *cte, uint64 planId,
```
onderkalaci: We're going to call this function for subqueries as well. So we'd better not to pass cte here?

onderkalaci: Maybe pass `List *aliascolnames `?

src/test/regress/expected/with_basics.out
```diff
@@ -220,6 +220,52 @@ JOIN
   users_table USING (user_id);
 ERROR:  cannot push down this subquery
 DETAIL:  Aggregates without group by are currently unsupported
+-- column aliases in CTE
```
onderkalaci: Maybe also add a target list entry with expressions such as `SELECT id, id * val, val, val+2 FROM cte` ?

src/backend/distributed/planner/recursive_planning.c
mtuncer: makes sense

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -546,7 +551,16 @@ BuildSubPlanResultQuery(Query *subquery, uint64 planId, uint32 subPlanId)
 		newTargetEntry = makeNode(TargetEntry);
 		newTargetEntry->expr = (Expr *) functionColumnVar;
 		newTargetEntry->resno = columnNumber;
-		newTargetEntry->resname = columnName;
+		if (cteAliasCount >= columnNumber)
```
onderkalaci: I cannot follow what does `if (cteAliasCount >= columnNumber)` mean.  

Did we mean `if (cteAliasCount > 0)` such that alias is provided?

Also,  `cteAliasCount` should probably not contain `cte` in the name.

src/test/regress/expected/with_basics.out
```diff
@@ -220,6 +220,88 @@ JOIN
...
+  3 |       5
+  1 |       5
+  5 |       4
+(10 rows)
+
+-- computed targets from columns also work
+WITH top_ten(id, val, val_mul, val_sum) AS (
+	SELECT user_id, value_1, value_1*2, value_1 + value_2 FROM users_table ORDER BY value_1 DESC, user_id DESC LIMIT 10
+)
+SELECT * FROM top_ten ORDER BY id DESC, val_mul DESC;
```
onderkalaci: I'd even to some calculations on the final target list such as `SELECT *, id+val, val_mul - 5 FROM top_ten ..` 😄 

Btw, could you please double check that the results are the same with Postgres?

src/backend/distributed/planner/recursive_planning.c
mtuncer: renamed the variable.

apparently you can write queries like
```sql
with cte(a, b) as ( select x, y, z from t) select * from cte;
```
cte defines only 2 columns even if inner query has 3. In such cases the output looks like
```sql
 a | b | z
---+---+---
```

Therefore, we only rename the column if an alias is specified, otherwise its own name is used as before.

mtuncer: Perhaps I should write that down as a comment

onderkalaci: > cte defines only 2 columns even if inner query has 3. 

Oh, that's crazy 😄

src/test/regress/expected/with_basics.out
mtuncer: final calculations : done,
comparison with postgres results : done

src/test/regress/expected/with_basics.out
```diff
@@ -220,6 +220,88 @@ JOIN
   users_table USING (user_id);
 ERROR:  cannot push down this subquery
 DETAIL:  Aggregates without group by are currently unsupported
+-- column aliases in CTE
+WITH top_ten(id, val1) AS (
+	SELECT user_id, value_1 FROM users_table ORDER BY value_1 DESC, user_id DESC LIMIT 10
+)
+SELECT * FROM top_ten;
```
onderkalaci: I wonder if we need `ORDER BY` on the top as well? What do you think? I think given that there is a single CTE, it seems fine. Feel free to ignore this comment.

mtuncer: we have order by in another test
```sql
-- computed targets from columns and group by outside
WITH top_ten(id, val, val_mul, val_sum) AS (
	SELECT user_id, value_1, value_1*2, value_1 + value_2 FROM users_table ORDER BY value_1 DESC, user_id DESC
)
SELECT id, count(*), avg(val), max(val_mul), min(val_sum) FROM top_ten
GROUP BY id
ORDER BY 2 DESC, 1 DESC;
```

src/test/regress/expected/with_basics.out
mtuncer: added

src/test/regress/expected/with_basics.out
onderkalaci: I mean could the results be inconsistent since we skip `ORDER BY`

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -528,7 +530,7 @@ RecursivelyPlanSubquery(Query *subquery, RecursivePlanningContext *planningConte
 	subPlan = CreateDistributedSubPlan(subPlanId, subquery);
 	planningContext->subPlanList = lappend(planningContext->subPlanList, subPlan);
 
-	resultQuery = BuildSubPlanResultQuery(subquery, planId, subPlanId);
+	resultQuery = BuildSubPlanResultQuery(subquery, NIL, planId, subPlanId);
```
onderkalaci: Maybe add a single line comment explaining why we're always passing `NIL` such as 
`/*  BuildSubPlanResultQuery() preserves the column aliases (if exists) for subqueries, thus no need to provide them for the result query */`

src/test/regress/sql/with_basics.sql
```diff
@@ -156,6 +156,44 @@ FROM
 JOIN
   users_table USING (user_id);
 
+-- column aliases in CTE
+WITH top_ten(id, val1) AS (
```
onderkalaci: Well, I should have added this in #1876, but I didn't think that we could write such queries:
```SQL
 SELECT * FROM (SELECT user_id, value_1, value_2 from users_table OFFSET 0) as foo(x, y) ORDER BY 1 DESC, 2 DESC, 3 DESC LIMIT 5;
```

Would you mind adding this to `src/test/regress/sql/subquery_complex_target_list.sql` ?

src/backend/distributed/planner/recursive_planning.c
mtuncer: done

src/test/regress/expected/with_basics.out
mtuncer: it should not cause inconsistency, we have order by in inner query on all columns. Generated intermediate table should have the results in order. If that behavior changes for any reason we would know from this test

src/test/regress/sql/with_basics.sql
mtuncer: added

mtuncer: done

PR #1894, opened 2017-12-20
Support set operations

### LINKED ISSUE #359: Support UNION SQL syntax
Noted by @samay-sharma as a frequent customer facing issue.
UNIONs: I think this may be one of the most encountered SQL constructs which we don't support. Also, it is relevant in the real-time analytics context.

marcocitus:
Implement set operations (`UNION`, `UNION ALL`, `INTERSECT`, `INTERSECT ALL`, `EXCEPT`, `EXCEPT ALL`) through recursive planning. 

Fixes #359

### DIFF
--- a/src/backend/distributed/planner/multi_logical_planner.c
+++ b/src/backend/distributed/planner/multi_logical_planner.c
@@ -97,9 +97,6 @@ static bool RangeTableArrayContainsAnyRTEIdentities(RangeTblEntry **rangeTableEn
 													queryRteIdentities);
 static Relids QueryRteIdentities(Query *queryTree);
 static DeferredErrorMessage * DeferErrorIfFromClauseRecurs(Query *queryTree);
-static DeferredErrorMessage * DeferErrorIfUnsupportedUnionQuery(Query *queryTree,
-																bool
-																outerMostQueryHasLimit);
 static bool ExtractSetOperationStatmentWalker(Node *node, List **setOperationList);
 static DeferredErrorMessage * DeferErrorIfUnsupportedTableCombination(Query *queryTree);
 static bool WindowPartitionOnDistributionColumn(Query *query);
@@ -985,8 +982,7 @@ DeferErrorIfCannotPushdownSubquery(Query *subqueryTree, bool outerMostQueryHasLi
 
 	if (subqueryTree->setOperations)
 	{
-		deferredError = DeferErrorIfUnsupportedUnionQuery(subqueryTree,
-														  outerMostQueryHasLimit);
+		deferredError = DeferErrorIfUnsupportedUnionQuery(subqueryTree);
 		if (deferredError)
 		{
 			return deferredError;
@@ -1098,9 +1094,8 @@ DeferErrorIfCannotPushdownSubquery(Query *subqueryTree, bool outerMostQueryHasLi
  * DeferErrorIfUnsupportedUnionQuery is a helper function for ErrorIfCannotPushdownSubquery().
  * The function also errors out for set operations INTERSECT and EXCEPT.
  */
-static DeferredErrorMessage *
-DeferErrorIfUnsupportedUnionQuery(Query *subqueryTree,
-								  bool outerMostQueryHasLimit)
+DeferredErrorMessage *
+DeferErrorIfUnsupportedUnionQuery(Query *subqueryTree)
 {
 	List *setOperationStatementList = NIL;
 	ListCell *setOperationStatmentCell = NULL;
@@ -1531,6 +1526,17 @@ bool
 FindNodeCheckInRangeTableList(List *rtable, bool (*check)(Node *))
 {
 	return range_table_walker(rtable, FindNodeCheck, check, QTW_EXAMINE_RTES);
+}
+
+
+/*
+ * QueryContainsDistributedTableRTE determines whether the given
+ * query contains a distributed table.
+ */
+bool
+QueryContainsDistributedTableRTE(Query *query)
+{
+	return FindNodeCheck((Node *) query, IsDistributedTableRTE);
 }
 
 
--- a/src/backend/distributed/planner/recursive_planning.c
+++ b/src/backend/distributed/planner/recursive_planning.c
@@ -67,10 +67,12 @@
 #include "distributed/relation_restriction_equivalence.h"
 #include "lib/stringinfo.h"
 #include "optimizer/planner.h"
+#include "parser/parsetree.h"
 #include "nodes/nodeFuncs.h"
 #include "nodes/nodes.h"
 #include "nodes/pg_list.h"
 #include "nodes/primnodes.h"
+#include "nodes/relation.h"
 #include "utils/builtins.h"
 #include "utils/guc.h"
 
@@ -117,6 +119,10 @@ static DeferredErrorMessage * RecursivelyPlanCTEs(Query *query,
 												  RecursivePlanningContext *context);
 static bool RecursivelyPlanSubqueryWalker(Node *node, RecursivePlanningContext *context);
 static bool ShouldRecursivelyPlanSubquery(Query *subquery);
+static bool ShouldRecursivelyPlanSetOperation(Query *query,
+											  RecursivePlanningContext *context);
+static void RecursivelyPlanSetOperations(Query *query, Node *node,
+										 RecursivePlanningContext *context);
 static bool IsLocalTableRTE(Node *node);
 static void RecursivelyPlanSubquery(Query *subquery,
 									RecursivePlanningContext *planningContext);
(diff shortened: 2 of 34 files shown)


onderkalaci:
Could you please rebase? Thanks!

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -455,6 +466,97 @@ ShouldRecursivelyPlanSubquery(Query *subquery, RecursivePlanningContext *context
...
+	if (context->level == 0)
+	{
+		/*
+		 * We cannot push down top-level set operation. Recursively plan the
+		 * leaf nodes such that it becomes a router query.
+		 */
+		return true;
+	}
+
+	if (setOperations->op != SETOP_UNION)
```
onderkalaci: `DeferErrorIfUnsupportedUnionQuery ()` already does this. But, you wanted to make it more explicit?

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -455,6 +466,97 @@ ShouldRecursivelyPlanSubquery(Query *subquery, RecursivePlanningContext *context
...
+	{
+		/*
+		 * If at least one leaf query in the union is recurring, then all
+		 * leaf nodes need to be recurring.
+		 */
+		return true;
+	}
+
+	filteredRestrictionContext =
+		FilterPlannerRestrictionForQuery(context->plannerRestrictionContext, query);
```
onderkalaci: It looks like the only problematic function for local tables is  `FilterPlannerRestrictionForQuery()->  ReferenceRelationCount() -> PartitionMethod()`. In fact, we should have skipped that for local tables when we allowed local tables in subqueries.

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -455,6 +466,97 @@ ShouldRecursivelyPlanSubquery(Query *subquery, RecursivePlanningContext *context
...
+	}
+	else if (IsA(node, RangeTblRef))
+	{
+		RangeTblRef *rangeTableRef = (RangeTblRef *) node;
+		RangeTblEntry *rangeTableEntry = rt_fetch(rangeTableRef->rtindex,
+												  query->rtable);
+		Query *subquery = rangeTableEntry->subquery;
+
+		if (rangeTableEntry->rtekind == RTE_SUBQUERY &&
+			QueryContainsDistributedTableRTE(subquery))
```
onderkalaci: Do we actually need `QueryContainsDistributedTableRTE(subquery)` or is it the part you're still thinking about?

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -206,6 +212,11 @@ RecursivelyPlanSubqueriesAndCTEs(Query *query, RecursivePlanningContext *context
 	/* descend into subqueries */
 	query_tree_walker(query, RecursivelyPlanSubqueryWalker, context, 0);
 
+	if (ShouldRecursivelyPlanSetOperation(query, context))
```
onderkalaci: Would be nice to have some comment after `query_tree_walker ()` that explains when we hit here? It might not be obvious for the readers

src/backend/distributed/planner/recursive_planning.c
marcocitus: We don't need to recursively plan subqueries that don't have a distributed table (and things like `SELECT * FROM reference_table LIMIT 10` are already recursively planned at this point).

marcocitus:
I briefly explored an alternative approach where we do the unions in a separate walk in `GenerateSubplansForSubqueriesAndCTEs`, but that becomes a little messy because`DeferErrorIfCannotPushdownSubquery` returns an error for set operations, so it will already try to recursively plan the subqueries containing the unions in the first round (which is unhelpful).

src/backend/distributed/planner/recursive_planning.c
marcocitus: Changed `AddRteRelationToAttributeEquivalenceClass`  to skip local relations.

src/backend/distributed/planner/recursive_planning.c
marcocitus: Yes, the name DeferErrorIfUnsupportedUnionQuery does not suggest it gives a deferred error for intersect or except

src/test/regress/expected/set_operations.out
```diff
@@ -0,0 +1,481 @@
+CREATE SCHEMA recursive_union;
```
onderkalaci: I guess you're still working on the regression tests. But, we should probably add tests for:

* Local tables
* CTEs
* EXCEPT/INTERSECT ALL

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -437,20 +458,98 @@ ShouldRecursivelyPlanSubquery(Query *subquery)
...
+	PlannerRestrictionContext *filteredRestrictionContext = NULL;
+
+	SetOperationStmt *setOperations = (SetOperationStmt *) query->setOperations;
+	if (setOperations == NULL)
+	{
 		return false;
 	}
 
-	return true;
+	if (context->level == 0)
```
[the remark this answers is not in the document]
marcocitus: Yep, worth opening up an issue for that. It's potentially *a lot* faster.

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1529,6 +1524,17 @@ FindNodeCheckInRangeTableList(List *rtable, bool (*check)(Node *))
 }
 
 
+/*
+ * QueryContainsDistributedTableRTE determines whether the given
+ * qury contains a distributed table.
```
marcocitus: qury -> query

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1529,6 +1524,17 @@ FindNodeCheckInRangeTableList(List *rtable, bool (*check)(Node *))
...
+ */
+bool
+QueryContainsDistributedTableRTE(Query *query)
+{
+	return FindNodeCheck((Node *) query, IsDistributedTableRTE);
```
marcocitus: I'm wondering whether we should do `FindNodeCheckInRangeTableList(query->rtable, IsDistributedTableRTE)` here, like we do in other places.

src/test/regress/sql/set_operation_and_local_tables.sql
```diff
@@ -0,0 +1,102 @@
...
+	WHERE ref.a = foo.x;
+
+-- subquery union in WHERE clause without parition column equality is recursively planned including the local tables
+SELECT * FROM test a WHERE x IN (SELECT x FROM test b UNION SELECT y FROM test c UNION SELECT y FROM local_test d) ORDER BY 1,2;
+
+-- same query with subquery in where is wrapped in CTE
+SELECT * FROM test a WHERE x IN (WITH cte AS (SELECT x FROM test b UNION SELECT y FROM test c UNION SELECT y FROM local_test d) SELECT * FROM cte) ORDER BY 1,2;
+
+-- not supported since local table is joined with a set operation
+SELECT * FROM ((SELECT * FROM test) EXCEPT (SELECT * FROM test ORDER BY x LIMIT 1)) u JOIN local_test USING (x) ORDER BY 1,2;
```
marcocitus: might be worth trying this query with a join against generate_series instead of the local table.

I get:
```
ERROR:  could not run distributed query with UNION, INTERSECT, or EXCEPT
```

marcocitus:
Might be worth trying a query with a join between a union and a generate_series, since that's what got us into trouble with local tables.

e.g.
```
-- pushdownable
 SELECT * FROM ((SELECT * FROM test) UNION (SELECT * FROM test ORDER BY x)) u JOIN generate_series(1,10) x USING (x) ORDER BY 1,2;

-- recursively plan all the things
 SELECT * FROM ((SELECT * FROM test) EXCEPT (SELECT * FROM test ORDER BY x)) u JOIN generate_series(1,10) x USING (x) ORDER BY 1,2;
```

onderkalaci:
> Might be worth trying a query with a join between a set operation and a generate_series, since that's what got us into trouble with local tables.

Makes sense

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: That's a good point. With our current implementation, we'd probably be fine given that we don't allow `SELECT * FROM ref WHERE x in (SELECT a FROM dist)`. 

But, I'd slightly prefer to keep it, for now, to prevent getting into tricky bugs in the near future?

onderkalaci:
I should add tests with `VIEWs`

marcocitus:
Can't approve, but :ship: from me.

PR #1895, opened 2017-12-21
Get shard resource locks for only DMLs

### LINKED ISSUE #1891: CREATE INDEX CONCURRENTLY blocks inserts for shard_replication_factor = 2
Here is how to replicate;

```sql
CREATE TABLE test_table (id int, value numeric);

SET citus.shard_replication_factor = 2;
SELECT create_distributed_table('test_table', 'id');

-- insert enough data
INSERT INTO test_table SELECT generate_series(1,10000000), random();
```

```sql
-- run these concurrently from different psql clients
CREATE INDEX CONCURRENTLY myindex ON test_table(id);

INSERT INTO test_table VALUES (1, 1);
```

metdos:
We already take required locks on the table for DDL commands and there is no need to take shard resource locks.

Fixes #1891

### DIFF
--- a/src/backend/distributed/executor/multi_router_executor.c
+++ b/src/backend/distributed/executor/multi_router_executor.c
@@ -1063,8 +1063,15 @@ ExecuteModifyTasks(List *taskList, bool expectResults, ParamListInfo paramListIn
 		LockPartitionRelations(firstShardInterval->relationId, RowExclusiveLock);
 	}
 
-	/* ensure that there are no concurrent modifications on the same shards */
-	AcquireExecutorMultiShardLocks(taskList);
+	/*
+	 * Ensure that there are no concurrent modifications on the same
+	 * shards. For DDL commands, we already obtained the appropriate
+	 * locks in ProcessUtility.
+	 */
+	if (firstTask->taskType == MODIFY_TASK)
+	{
+		AcquireExecutorMultiShardLocks(taskList);
+	}
 
 	BeginOrContinueCoordinatedTransaction();
 


src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -1054,8 +1054,15 @@ ExecuteModifyTasks(List *taskList, bool expectResults, ParamListInfo paramListIn
...
 	}
 
-	/* ensure that there are no concurrent modifications on the same shards */
-	AcquireExecutorMultiShardLocks(taskList);
+	/*
+	 * Ensure that there are no concurrent modifications on the same
+	 * shards. For DDL commands, we already obtained the appropriate
+	 * locks in ProcessUtility.
+	 */
+	if (firstTask->taskType != DDL_TASK)
```
marcocitus: Maybe `taskType == MODIFY_TASK`? 

Might be worth mentioning that the current motivation for `AcquireExecutorMultiShardLocks` is two-fold:
- To prevent multi-shard UPDATE/DELETE/INSERT...SELECT from running concurrently with each other. This is not strictly necessary but it avoids making an excessive number of connections and creating many distributed deadlocks.
- Ensure that non-commutative writes are serialised when replication factor > 1

metdos: > Might be worth mentioning that the current motivation for AcquireExecutorMultiShardLocks is two-fold:

Do you mean here in the comment? We already have explained these in comments of `AcquireExecutorMultiShardLocks`.

metdos:
> Is there an isolation test that we can add?

I think we can't
```sql
postgres=# CREATE INDEX CONCURRENTLY myindex ON test_table(id);
ERROR:  CREATE INDEX CONCURRENTLY cannot run inside a transaction block
```

PR #1897, opened 2017-12-21
new debug output for subplans

furkansahin:
Now we also print out the query with read_intermediate_result replacements for CTE/complex subqueries when the log_min_messages or client_min_messages <= DEBUG1

### DIFF
--- a/src/test/regress/expected/multi_insert_select.out
+++ b/src/test/regress/expected/multi_insert_select.out
@@ -692,6 +692,7 @@ INSERT INTO agg_events
 DEBUG:  distributed INSERT ... SELECT can only select from distributed tables
 DEBUG:  Collecting INSERT ... SELECT results on coordinator
 DEBUG:  generating subplan 51_1 for CTE fist_table_agg: SELECT (max(value_1) + 1) AS v1_agg, user_id FROM public.raw_events_first GROUP BY user_id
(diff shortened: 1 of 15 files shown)


src/test/regress/expected/multi_insert_select.out
```diff
@@ -692,6 +692,7 @@ INSERT INTO agg_events
 DEBUG:  distributed INSERT ... SELECT can only select from distributed tables
 DEBUG:  Collecting INSERT ... SELECT results on coordinator
 DEBUG:  generating subplan 51_1 for CTE fist_table_agg: SELECT (max(value_1) + 1) AS v1_agg, user_id FROM public.raw_events_first GROUP BY user_id
+DEBUG:  Plan 51 query: SELECT user_id, v1_agg FROM (SELECT fist_table_agg.v1_agg, fist_table_agg.user_id FROM (SELECT intermediate_result.v1_agg, intermediate_result.user_id FROM read_intermediate_result('51_1'::text, 'binary'::citus_copy_format) intermediate_result(v1_agg integer, user_id integer)) fist_table_agg) citus_insert_select_subquery
```
marcocitus: I wonder if we could give users a tiny bit of context on what's going on here, e.g. "plan %d query after replacing subqueries and CTEs: "

PR #1899, opened 2017-12-21
Prototype: Recursively plan sublinks in WHERE clause except IN/ANY

marcocitus:
Recursively plan non-correlated sublinks in WHERE except IN/ANY, which might be pushdownable.

Needs regression tests.

### DIFF
--- a/src/backend/distributed/planner/multi_logical_planner.c
+++ b/src/backend/distributed/planner/multi_logical_planner.c
@@ -104,7 +104,6 @@ static bool ExtractSetOperationStatmentWalker(Node *node, List **setOperationLis
 static DeferredErrorMessage * DeferErrorIfUnsupportedTableCombination(Query *queryTree);
 static bool WindowPartitionOnDistributionColumn(Query *query);
 static bool AllTargetExpressionsAreColumnReferences(List *targetEntryList);
-static bool IsDistributedTableRTE(Node *node);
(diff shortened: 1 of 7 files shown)


onderkalaci:
What about the following query with `IN`:
```SQL
SELECT 
	count(*)
FROM
	users_table
WHERE
	value_2 IN
	(
		(SELECT user_id FROM users_table )
		UNION
	 	(SELECT user_id FROM events_table)
)
ORDER BY 1 DESC;
```

Should we be able to recursively plan the subquery?

marcocitus:
> Should we be able to recursively plan the subquery?

That's a non-co-located join issue, not something we can handle through the same code path.

marcocitus:
Random thought on this PR: 

At the moment, recursive planning primarily kicks in when you use something like a GROUP BY on a non-partition column or LIMIT. These constructs naturally reduce the size of the result set and require a merge step, which makes them good candidates for pulling to the coordinator.

The logic in the PR primarily affects subqueries that apparently passed the pushdown checks, which means they may have larger intermediate results, but probably do not require a merge step, and therefore they are good candidates for broadcasting (or repartitioning) directly between the workers.

I wonder whether we should first, or in tandem, invest in new mechanisms for copying intermediate results around.

onderkalaci:
Closing this since we're already targeting it in #2016

PR #1901, opened 2017-12-22
Error messages are updated after recursive planner

furkansahin:
After implementing the recursive planner, some of our error messages in logical planner became misleading. This PR updates these error messages accordingly.

### DIFF
--- a/src/backend/distributed/planner/multi_logical_planner.c
+++ b/src/backend/distributed/planner/multi_logical_planner.c
@@ -960,14 +960,16 @@ DeferErrorIfCannotPushdownSubquery(Query *subqueryTree, bool outerMostQueryHasLi
 	if (subqueryTree->limitOffset)
 	{
 		preconditionsSatisfied = false;
-		errorDetail = "Offset clause is currently unsupported";
+		errorDetail = "Offset clause is currently unsupported when a subquery "
+					  "references a column from another query";
 	}
 
 	/* limit is not supported when SubqueryPushdown is not set */
 	if (subqueryTree->limitCount && !SubqueryPushdown)
 	{
 		preconditionsSatisfied = false;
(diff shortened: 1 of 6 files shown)


src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -960,14 +960,16 @@ DeferErrorIfCannotPushdownSubquery(Query *subqueryTree, bool outerMostQueryHasLi
 	if (subqueryTree->limitOffset)
 	{
 		preconditionsSatisfied = false;
-		errorDetail = "Offset clause is currently unsupported";
+		errorDetail = "Offset clause is currently unsupported when there "
```
onderkalaci: I wonder which is more useful/accurate:
`Offset clause is currently unsupported when there is a reference to an outer query` vs
`Offset clause is currently unsupported when a subquery references a column from another query `

I'm still thinking about the message. Do you have any other alternatives in mind?

onderkalaci: `the subquery` or `a subquery`? I need some grammar lessons 😄

furkansahin: I was checking this in the internet and saw that T-SQL also uses a term like `outer reference` [link](http://www.databasesanddeadlanguages.info/2014/02/t-sql-error-each-group-by-expression.html]). The context was different than ours but this led me to think that ` reference to an outer query` as a term make sense. But of course I am open to suggestions

furkansahin: > `the subquery` or `a subquery`? I need some grammar lessons 😄 
I guess `a subquery` is better if we pick it

onderkalaci:
This looks good :shipit:  after (i) Squashing the commits (ii) Updating the PR title and description

PR #1903, opened 2017-12-22
Do not raise errors in the real-time executor

### LINKED ISSUE #1902: Subqueries getting stuck on abort transaction
When we run a query with subquery/CTEs and if there is a problem in the execution, we send abort through the connections and this gets stuck. These are the example queries which I can successfully reproduce the problem;
```SQL
WITH cte as (
	SELECT 
		user_id, value_2 
	from 
		events_table
)
SELECT * FROM users_table where value_2 < (
	SELECT 
		min(cte.value_2) 
	FROM 
		cte 
	WHERE 
		users_table.user_id=cte.user_id 
	GROUP BY 
		user_id, cte.value_2);
```

```SQL
SELECT
	user_id
FROM
	users_table
WHERE
	value_2 > (
		SELECT 
			DISTINCT ON (max(value_2)) value_2
		FROM
			events_table
		WHERE
			event_type = 2
		GROUP BY 
			value_2 
			HAVING min(value_2) > 0
	)
GROUP BY user_id
HAVING count(*) > 3
ORDER BY user_id
LIMIT 5;
```

These two queries both should error out by saying the subquery returns more than one row.

This is the backtrace
```
(gdb) bt
#0  0x00007fffc421d2a6 in malloc_zone_malloc ()
#1  0x00007fffc421c200 in malloc ()
#2  0x0000000101d936ce in PQmakeEmptyPGresult (conn=0x7fcf60d12850, status=PGRES_COPY_OUT) at fe-exec.c:146
#3  0x0000000101d96950 in getCopyResult (conn=0x7fcf60d12850, copytype=PGRES_COPY_OUT) at fe-exec.c:1881
#4  0x0000000101d96730 in PQgetResult (conn=0x7fcf60d12850) at fe-exec.c:1814
#5  0x0000000101c84430 in NonblockingForgetResults (connection=0x7fcf610ea098) at connection/remote_commands.c:184
#6  0x0000000101cd6403 in StartRemoteTransactionAbort (connection=0x7fcf610ea098) at transaction/remote_transaction.c:383
#7  0x0000000101cd705f in CoordinatedRemoteTransactionsAbort () at transaction/remote_transaction.c:862
#8  0x0000000101cd7c42 in CoordinatedTransactionCallback (event=XACT_EVENT_ABORT, arg=0x0) at transaction/transaction_management.c:203
#9  0x000000010128a336 in CallXactCallbacks (event=XACT_EVENT_ABORT) at xact.c:3347
#10 0x0000000101285bef in AbortTransaction () at xact.c:2591
#11 0x0000000101286ac5 in AbortCurrentTransaction () at xact.c:3021
#12 0x0000000101612a80 in PostgresMain (argc=1, argv=0x7fcf63003180, dbname=0x7fcf63003048 "postgres", username=0x7fcf61002fc8 "mehmetfurkansahin") at postgres.c:3879
#13 0x000000010156e0d9 in BackendRun (port=0x7fcf60c048c0) at postmaster.c:4405
#14 0x000000010156d493 in BackendStartup (port=0x7fcf60c048c0) at postmaster.c:4077
#15 0x000000010156c555 in ServerLoop () at postmaster.c:1755
#16 0x000000010156a0b9 in PostmasterMain (argc=3, argv=0x7fcf60c02cf0) at postmaster.c:1363
#17 0x0000000101499d89 in main (argc=3, argv=0x7fcf60c02cf0) at main.c:228
```

At some point, `NonblockingForgetResults` gets into an infinite loop.
The partial fix can be found [here](https://github.com/citusdata/citus/tree/fix_abort_stuck) by @marcoslot

marcocitus:
Fixes #1902 

In #1785 we added an error to the real-time executor if an error occurred on a connection with a critical transaction (e.g. a DDL command was performed prior to the SELECT, or an intermediate result was sent over the connection). Throwing an error means that we skip the cancellation logic in the real-time executor and go straight to the abort handler. Unfortunately, the abort handler currently cannot handle connections that have a `COPY (SELECT ...) TO STDOUT` in progress, because `NonblockingForgetResults` gets stuck in a loop. 

The `NonblockingForgetResults` function exists to avoid getting blocked on network I/O (waiting for the worker) in the abort handler, which previously caused users to sometimes have to cancel twice. This function did not consider real-time executor queries because those never ran in a transaction block prior to #1785 and therefore `StartRemoteTransactionAbort` was never called. This problem is fixed separately in #1905.

For the moment the only reliable way we have to cancel a real-time query is to drop into the `PQcancel` logic at the end of the real-time executor and perform cleanup. This PR changes the real-time executor to avoid throwing an error and instead adds a flag to the TaskExecution to immediately drop out of the execution loop when a failure occurs on a connection with a critical transaction.

A tricky part of this approach is that the real-time executor already observes the error in `MultiClientCopyData` and it does not let us distinguish between a connection/node failure and a query error, and also allows one more real-time executor round before we handle the `EXEC_TASK_FAILED` state. This causes warnings from other tasks to show up in the output as well.

In addition, because the real-time executor might reuse one of the connections that was used to copy the intermediate result to the worker, it might run a failing query inside the transaction block that created the intermediate result, which will then abort and instantly remove the intermediate result, causing errors in other transactions that are trying to open the result file.

While this PR fixes the critical issue (getting permanently stuck in an infinite loop), we may see confusing warnings like "result 25_1 does not exist" when an error occurs in the transaction that created the intermediate result.

```
# SELECT * FROM test WHERE x = (SELECT x FROM test WHERE x = 1 LIMIT 2 OFFSET 0);
WARNING:  more than one row returned by a subquery used as an expression
WARNING:  result "25_1" does not exist
WARNING:  more than one row returned by a subquery used as an expression
```

I haven't come up with a nice and simple solution to this problem yet.

### DIFF
--- a/src/backend/distributed/executor/multi_real_time_executor.c
+++ b/src/backend/distributed/executor/multi_real_time_executor.c
@@ -387,7 +387,7 @@ ManageTaskExecution(Task *task, TaskExecution *taskExecution,
 
 		case EXEC_TASK_FAILED:
 		{
-			bool raiseError = true;
+			bool raiseError = false;
 
 			/*
 			 * On task failure, we close the connection. We also reset our execution
@@ -397,12 +397,15 @@ ManageTaskExecution(Task *task, TaskExecution *taskExecution,
 			 */
 			int32 connectionId = connectionIdArray[currentIndex];
 			MultiConnection *connection = MultiClientGetConnection(connectionId);
+			bool isCritical = connection->remoteTransaction.transactionCritical;
 
 			/*
-			 * If this connection was previously marked as critical (e.g. it was used
-			 * to perform a DDL command), then throw an error. Otherwise, mark it
-			 * as failed and continue executing the query.
+			 * Mark the connection as failed in case it was already used to perform
+			 * writes. We do not error out here, because the abort handler currently
+			 * cannot handle connections with COPY (SELECT ...) TO STDOUT commands
+			 * in progress.
 			 */
+			raiseError = false;
 			MarkRemoteTransactionFailed(connection, raiseError);
 
 			MultiClientDisconnect(connectionId);
@@ -411,8 +414,16 @@ ManageTaskExecution(Task *task, TaskExecution *taskExecution,
 
 			taskStatusArray[currentIndex] = EXEC_TASK_CONNECT_START;
 
-			/* try next worker node */
-			AdjustStateForFailure(taskExecution);
+			if (isCritical)
+			{
+				/* cannot recover when error occurs in a critical transaction */
+				taskExecution->criticalErrorOccurred = true;
+			}
+			else
+			{
+				/* try next worker node */
+				AdjustStateForFailure(taskExecution);
+			}
 
 			/*
 			 * Add a delay, to avoid potentially excerbating problems by
--- a/src/backend/distributed/executor/multi_server_executor.c
+++ b/src/backend/distributed/executor/multi_server_executor.c
@@ -249,6 +249,11 @@ CleanupTaskExecution(TaskExecution *taskExecution)
 bool
 TaskExecutionFailed(TaskExecution *taskExecution)
 {
+	if (taskExecution->criticalErrorOccurred)
+	{
+		return true;
+	}
+
 	if (taskExecution->failureCount >= MAX_TASK_EXECUTION_FAILURES)
 	{
 		return true;
--- a/src/include/distributed/multi_server_executor.h
+++ b/src/include/distributed/multi_server_executor.h
@@ -145,6 +145,7 @@ struct TaskExecution
 	uint32 querySourceNodeIndex; /* only applies to map fetch tasks */
 	int32 dataFetchTaskIndex;
 	uint32 failureCount;
+	bool criticalErrorOccurred;
 };
 
 
--- a/src/test/regress/expected/subquery_and_cte.out
+++ b/src/test/regress/expected/subquery_and_cte.out
@@ -435,6 +435,42 @@ FROM
      ) as bar  
 WHERE foo.user_id = bar.user_id;
 ERROR:  recursive CTEs are not supported in distributed queries
+-- We error-out when there's an error in execution of the query. By repeating it
+-- multiple times, we increase the chance of this test failing before PR #1903.
+SET client_min_messages TO ERROR;
+DO $$
+DECLARE
+	errors_received INTEGER;
+BEGIN
+errors_received := 0;
+FOR i IN 1..3 LOOP
+	BEGIN
+		WITH cte as (
+			SELECT
+				user_id, value_2
+			from
+				events_table
+		)
+		SELECT * FROM users_table where value_2 < (
+			SELECT
+				min(cte.value_2)
+			FROM
+				cte
+			WHERE
+				users_table.user_id=cte.user_id
+			GROUP BY
+				user_id, cte.value_2);
+	EXCEPTION WHEN OTHERS THEN
+		IF SQLERRM LIKE 'failed to execute task%' THEN
+			errors_received := errors_received + 1;
+		END IF;
+	END;
+END LOOP;
+RAISE '(%/3) failed to execute one of the tasks', errors_received;
+END;
+$$;
+ERROR:  (3/3) failed to execute one of the tasks
+CONTEXT:  PL/pgSQL function inline_code_block line 29 at RAISE
 SET client_min_messages TO DEFAULT;
 DROP SCHEMA subquery_and_ctes CASCADE;
 NOTICE:  drop cascades to table users_table_local
--- a/src/test/regress/sql/subquery_and_cte.sql
+++ b/src/test/regress/sql/subquery_and_cte.sql
@@ -319,6 +319,40 @@ FROM
      ) as bar  
 WHERE foo.user_id = bar.user_id;
 
+-- We error-out when there's an error in execution of the query. By repeating it
+-- multiple times, we increase the chance of this test failing before PR #1903.
+SET client_min_messages TO ERROR;
+DO $$
+DECLARE
+	errors_received INTEGER;
+BEGIN
+errors_received := 0;
+FOR i IN 1..3 LOOP
+	BEGIN
+		WITH cte as (
+			SELECT
+				user_id, value_2
+			from
+				events_table
+		)
+		SELECT * FROM users_table where value_2 < (
+			SELECT
+				min(cte.value_2)
+			FROM
+				cte
+			WHERE
+				users_table.user_id=cte.user_id
+			GROUP BY
+				user_id, cte.value_2);
+	EXCEPTION WHEN OTHERS THEN
+		IF SQLERRM LIKE 'failed to execute task%' THEN
+			errors_received := errors_received + 1;
+		END IF;
+	END;
+END LOOP;
+RAISE '(%/3) failed to execute one of the tasks', errors_received;
+END;
+$$;
 
 SET client_min_messages TO DEFAULT;
 


pykello:
I am looking into this. I am wondering why didn't we instead make the following change, which apparently also solves the issue:

```C
diff --git a/src/backend/distributed/connection/remote_commands.c b/src/backend/distributed/connection/remote_commands.c
index e5467fbc..a09aa877 100644
--- a/src/backend/distributed/connection/remote_commands.c
+++ b/src/backend/distributed/connection/remote_commands.c
@@ -182,7 +182,8 @@ NonblockingForgetResults(MultiConnection *connection)
                }
 
                result = PQgetResult(pgConn);
-               if (PQresultStatus(result) == PGRES_COPY_IN)
+               if (PQresultStatus(result) == PGRES_COPY_IN ||
+                       PQresultStatus(result) == PGRES_COPY_OUT)
                {
                        /* in copy, can't reliably recover without blocking */
                        return false;
```

From the `/* in copy, can't reliably recover without blocking */` comment, it seems that we should have done this in the first place.

@onderkalaci do you know the reason?

I am doing more investigation.

marcocitus:
@pykello that's done in #1905

onderkalaci:
> For the moment the only reliable way we have to cancel a real-time query is to drop into the PQcancel logic at the end of the real-time executor and perform cleanup. 

To me, the above is a sufficient argument to get this PR merged. Otherwise, we'd be leaking resources which seems pretty bad :/


> , we may see confusing warnings like "result 25_1 does not exist" when an error occurs in the transaction that created the intermediate result.

Actually, that didn't disturb me at all. I think to see the `WARNING:  more than one row returned by a subquery used as an expression` should be sufficient for many people to realize that they have a problem with the query.  Also, just adding a related note that we're already giving multiple warnings for many other cases so that shouldn't also be an important aspect of this change:
```SQL
 select user_id/0 FROM users_table;
WARNING:  division by zero
WARNING:  division by zero
WARNING:  division by zero
WARNING:  division by zero
WARNING:  division by zero
WARNING:  division by zero
WARNING:  division by zero
WARNING:  division by zero
WARNING:  division by zero
WARNING:  division by zero
WARNING:  division by zero
WARNING:  division by zero
WARNING:  division by zero
ERROR:  failed to execute task 14
```

> @pykello that's done in #1905. However, we also skip some essential clean-up if we go straight into the abort handler, so we should still avoid errors.

Thanks, @marcocitus for opening these two separate PRs now it's much easier to reason about!

pykello:
I added a test for this.

1. I turned off the warnings, since they were inconsistent from execution to execution (Number of `more than one row ...` messages changed sometimes, and ` result "25_1" does not exist` errors happened sometimes).
2. The error message also had an inconsistent task number, so I just catch the exception and check that it has the expected pattern.
3. The test sometimes succeeded before this PR, so I repeat the query 3 times in a loop to increase the chance of test failing before this PR.

I spent yesterday understanding the changes and trying some alternative ways. The changes look good to me as they are.

pykello:
@onderkalaci can you please take a final look and merge the changes of they make sense?

onderkalaci:
> @onderkalaci can you please take a final look and merge the changes if they make sense?

Hey @pykello, @sumedhpathak and @metdos (and @marcocitus  if reachable 😄 )

Both #1903 and #1905 makes sense to me. The tests look good to me as well. 

However, I hit an issue (when I merge two PRs into another branch), which I haven't resolved yet and wanted to hand-off to you back.

You added the loop in the regression tests, which I've extended to run for 5000 times such as the following:

```SQL
SET client_min_messages TO WARNING;
DO $$
DECLARE
	errors_received INTEGER;
BEGIN
errors_received := 0;
FOR i IN 1..5000 LOOP
	BEGIN
		WITH cte as (
			SELECT
				user_id, value_2
			from
				events_table
		)
		SELECT * FROM users_table where value_2 < (
			SELECT
				min(cte.value_2)
			FROM
				cte
			WHERE
				users_table.user_id=cte.user_id
			GROUP BY
				user_id, cte.value_2);
	EXCEPTION WHEN OTHERS THEN
		IF SQLERRM LIKE 'failed to execute task%' THEN
			errors_received := errors_received + 1;
			RAISE WARNING '(%/5000) failed to execute one of the tasks', errors_received;
		END IF;
	END;
END LOOP;
RAISE '(%/5000) failed to execute one of the tasks', errors_received;
END;
$$;

```

If you wait until the loop finishes, we're all good.

But, if you try to cancel at a random point in the loop, we occasionally get **segmentation faults**.  Unfortunately, the stack trace of the crash doesn't reveal anything such as the following:

```
#0  __longjmp () at ../sysdeps/x86_64/__longjmp.S:67
#1  0x171530d2d8648d6c in ?? ()
Cannot access memory at address 0xc0d530d2e79b3204
```

However, when I write the query (i.e., including `BEGIN`/`END`) 5000 inside a file (i.e., get rid of `pl/pgsql`),  run them by `psql -f queries.sql`  and send a `SIGINT` at a random point, I **do not see the segmentation faults**

Could you investigate that some more? Could it be specific to `pl/pgsql`?

onderkalaci:
> Could you investigate that some more? Could it be specific to pl/pgsql?

Or, it might be a completely separate issue?

pykello:
I am looking into the issue @onderkalaci reported. It is probably related to real-time executor and these PRs since I couldn't reproduce it with task-tracker queries.

pykello:
My current guess is that this has something to do with sub-transactions (i.e. savepoints). pl/pgsql uses sub-transactions for exception handling. Also, when I trace by debug messages, I get the crash in `pl_exec.c -> exec_stmt_block()` a bit after `RollbackAndReleaseCurrentSubTransaction();` returns.

I am not sure yet, since trying some similar code block without exceptions doesn't work in Citus (I'll open a separate issue for this) 😞 : 

```sql
DO $$                   
DECLARE
  x INTEGER;
BEGIN
	FOR i IN 1..5000 LOOP
		SELECT user_id INTO x FROM users_table where value_2 < 1;
	END LOOP;
END;
$$;

ERROR:  could not create directory "base/pgsql_job_cache/master_job_119168565249": File exists
CONTEXT:  SQL statement "SELECT user_id        FROM users_table where value_2 < 1"
PL/pgSQL function inline_code_block line 6 at SQL statement
```

I am still looking into this.

marcocitus:
> ERROR:  could not create directory 

I guess we're not removing the job directory until the DO statement is done and all invocations are using the same prepared statement with the same job ID. I guess we could do some clean up in the transaction hooks, but maybe simpler to not error out if the directory exists?

pykello:
I think I found the issue. If you look at elog.h, you see that both `PG_CATCH()` and `PG_END_TRY()` reset `PG_exception_stack`. So we shouldn't return in middle of of a `PG_TRY()` block, otherwise this variable won't be reset. If we don't reset this variable, exception handling will be messed up and we will see unexpected behaviours as the one we were seeing.

We do this in `WaitForAllConnections()`. It is called by `CoordinatedRemoteTransactionsSavepointRollback()` which is called by PostgreSQL's `RollbackAndReleaseCurrentSubTransaction()` which is called by plpgsql's `exec_stmt_block()`.

When I make the following change, the problem goes away:

```C
diff --git a/src/backend/distributed/connection/remote_commands.c b/src/backend/distributed/connection/remote_commands.c
index a09aa877..aa40c475 100644
--- a/src/backend/distributed/connection/remote_commands.c
+++ b/src/backend/distributed/connection/remote_commands.c
@@ -838,8 +838,7 @@ WaitForAllConnections(List *connectionList, bool raiseInterrupts)
 													  ProcDiePending))
 					{
 						/* return immediately in case of cancellation */
-						FreeWaitEventSet(waitEventSet);
-						return;
+						break;
 					}
 
 					continue;
@@ -905,6 +904,12 @@ WaitForAllConnections(List *connectionList, bool raiseInterrupts)
 				}
 			}
 
+			if (InterruptHoldoffCount > 0 && (QueryCancelPending || ProcDiePending))
+			{
+				/* return immediately in case of cancellation */
+				break;
+			}
+
 			/* move non-ready connections to the back of the array */
 			for (connectionIndex = pendingConnectionsStartIndex;
 				 connectionIndex < totalConnectionCount; connectionIndex++)
```

I'll think a bit more about this, do more testing, and send it as a PR later today or over weekend.

As far as it is related to this PR and #1905, I think they can be merged as they are.

pykello:
Keeping the changes to `WaitForAllConnections()` for Tuesday, since I think it might have few more problems (I might be wrong, but when I read it some parts were confusing and I want to spend some more time understanding them).

Issue #1784, opened 2017-11-10
Cancellation leaks connection IDs in real-time executor


marcocitus:
In the real-time (and task-tracker) executor, we don't call `MultiClientDisconnect` on cancellation, since it's assumed that we disconnect all open connections on ABORT in the transaction callback. However, we don't reset `ClientConnectionArray` in `multi_client_executor.c` on ABORT, which means this array will eventually fill up after a large number of cancellations:

```
set citus.shard_count = 100;
CREATE TABLE test (x int, y int);
SELECT create_distributed_table('test','x');
INSERT INTO test SELECT s,s FROM generate_series(1,1000000) s;

# SELECT count(*) FROM test;
^CCancel request sent
ERROR:  canceling statement due to user request
# SELECT count(*) FROM test;
^CCancel request sent
ERROR:  canceling statement due to user request
...

# SELECT count(*) FROM test;
WARNING:  could not allocate connection in connection pool
...
WARNING:  could not allocate connection in connection pool
ERROR:  failed to execute task 106
```

marcocitus:
Closing in favour of #2526 which has a more detailed analysis.

PR #1905, opened 2017-12-22
Fix cancellation issues in the real-time executor

### LINKED ISSUE #1904: Real-time query in a transaction block might get stuck when cancelled
There's a suspicious problem in the real-time executor cancellation logic related to #1902.

We could throw an error on cancellation in:
https://github.com/citusdata/citus/blob/master/src/backend/distributed/executor/multi_client_executor.c#L932

If that error gets thrown, then in the abort handler we would hit `NonblockingForgetResults`:
https://github.com/citusdata/citus/blob/master/src/backend/distributed/transaction/remote_transaction.c#L383

Then in `NonblockingForgetResults` we do a `PQgetResult`, but on a connection that is doing a `COPY (SELECT ...) TO STDOUT` this always returns the result of `PQmakeEmptyPGresult` and is thus never `NULL`, causing an infinite loop.
https://github.com/citusdata/citus/blob/master/src/backend/distributed/connection/remote_commands.c#L106

I haven't managed to reproduce it, though manually adding an error in the real-time executor trivially causes the coordinator to get stuck in an infinite loop, e.g.

```sql
BEGIN;
-- send BEGIN on all connections
UPDATE test SET y = y + 1;

-- real-time SELECT on the connections
SELECT count(*) FROM test;
<error thrown -> infinite loop>
```

marcocitus:
Fixes #1904 

Two small fixes in addition to #1903, which fixes a different error in the real-time executor.

1. Do not run `CHECK_FOR_INTERRUPTS()` in `MultiClientWait`
2. Do not get stuck in `NonblockingForgetResults` if the connection is performing a COPY (SELECT ..) TO STDOUT

We are not in a state where we can liberally throw errors in the real-time executor. Some important clean-up happens at the end of `MultiRealTimeExecute`. Moreover, if we do throw an error we might get stuck in an infinite loop in `NonblockingForgetResults` now that real-time queries can run in a transaction block and may be followed by `ROLLBACK`. This PR defensively fixes both of those issues.

A longer-term fix is to start closing files via the resource owner framework or directly write to a tuple store, and handle in-progress real-time queries in the abort handler, but given that we're at 7.2 feature freeze, a simpler fix seemed preferable.

This may also be related to #1784.

### DIFF
--- a/src/backend/distributed/connection/remote_commands.c
+++ b/src/backend/distributed/connection/remote_commands.c
@@ -182,7 +182,8 @@ NonblockingForgetResults(MultiConnection *connection)
 		}
 
 		result = PQgetResult(pgConn);
-		if (PQresultStatus(result) == PGRES_COPY_IN)
+		if (PQresultStatus(result) == PGRES_COPY_IN ||
+			PQresultStatus(result) == PGRES_COPY_OUT)
 		{
 			/* in copy, can't reliably recover without blocking */
 			return false;
--- a/src/backend/distributed/executor/multi_client_executor.c
+++ b/src/backend/distributed/executor/multi_client_executor.c
@@ -924,15 +924,11 @@ MultiClientWait(WaitInfo *waitInfo)
 		{
 			/*
 			 * Signals that arrive can interrupt our poll(). In that case just
-			 * check for interrupts, and try again. Every other error is
-			 * unexpected and treated as such.
+			 * return. Every other error is unexpected and treated as such.
 			 */
 			if (errno == EAGAIN || errno == EINTR)
 			{
-				CHECK_FOR_INTERRUPTS();
-
-				/* maximum wait starts at max again, but that's ok, it's just a stopgap */
-				continue;
+				return;
 			}
 			else
 			{
--- a/src/test/regress/expected/multi_real_time_transaction.out
+++ b/src/test/regress/expected/multi_real_time_transaction.out
@@ -291,6 +291,64 @@ SELECT * FROM co_test_table;
 (1 row)
 
 ROLLBACK;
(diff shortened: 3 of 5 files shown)


onderkalaci:
> directly write to a tuple store,

I'd be in favour of this in the near future!

onderkalaci:
> This may also be related to #1784.

Yes, that seems to be fixed as well.

onderkalaci:
@pykello We should consider adding a regression test to this PR.

The following is a suggestion is a test, which ensures that we're canceling transactions correctly in the real-time executor:


```SQL

BEGIN;
 select user_id, pg_advisory_lock(15) FROM users_table LIMIT 5;
 ROLLBACK;

```

We're actually creating a distributed deadlock manually by acquiring the same advisory lock on all shards. 

If you run this consecutively (i.e., 10 times) on the master branch, you'd see that the transaction would block on the coordinator. But, not in this branch since we're able to rollback things correctly here.

To kick the deadlock detection faster, maybe set the deadlock timeout to much lower value (and set back to default after the test)
```
alter system set deadlock_timeout TO '1ms';
SELECT pg_reload_conf();
```
Also, during the tests maybe set the `set client_min_messages TO ERROR;` to skip some warnings.

(Thanks @marcocitus for the deadlock idea)

onderkalaci:
> Also, during the tests maybe set the set client_min_messages TO ERROR; to skip some warnings.

Actually, that may be a bad idea. I see some warnings, which we may consider opening a separate issue:


```
WARNING:  could not issue cancel request
DETAIL:  Client error: PQcancel() -- connect() failed: Socket is already connected
```

But, at least to make the test outputs consistent we could set the  set client_min_messages TO ERROR; for those tests

pykello:
@onderkalaci can you please take a final look and merge the changes if they make sense?

PR #1906, opened 2017-12-25
remove the flag

furkansahin:
WIP

references #1806

### DIFF
--- a/src/backend/distributed/executor/multi_router_executor.c
+++ b/src/backend/distributed/executor/multi_router_executor.c
@@ -73,9 +73,6 @@
 /* controls use of locks to enforce safe commutativity */
 bool AllModificationsCommutative = false;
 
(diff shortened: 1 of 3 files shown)


marcocitus:
Does this cause errors on startup if people still have the flag in their postgresopen.conf?

onderkalaci:
> Does this cause errors on startup if people still have the flag in their postgresopen.conf?

It should :) We talked with @metdos and discussed that this is better for 7.3. We've been warning the users since 7.0.  Another option could be never remove this GUC and add a comment in the code?

marcocitus:
We should only remove GUCs in major releases (8.0/PG11), since it requires upgrade steps.

onderkalaci:
> We should only remove GUCs in major releases (8.0/PG11), since it requires non-standard upgrade steps.

I missed that... So, we should close this PR

furkansahin:
This issue is postponed to major release.

PR #1907, opened 2017-12-25
Add some regression tests for recursively planned subqueries

onderkalaci:
Add some more basic regression tests, mostly for documentation purposes. The aims is to have one simple test for each case in `DeferErrorIfCannotPushdownSubquery ()`.


onderkalaci:
This PR only adds 3 basic regression test, so merging without review.

PR #1918, opened 2017-12-28
Outer joins should also use/try subquery pushdown planner if join clause is not supported

### LINKED ISSUE #1916: Relax equi join expressions for reference tables when they are in an outer join
With 7.2, we aimed to fix a bug in `IsJoinClause()` function, which led to a new bug. 

The bug was that we used to allow joins in the form of `function1(dist_table.a) = function2(dist_table.b)`, which could easily lead to wrong results. 

We aimed to allow the joins in the following form only `dist_table.a = dist_table.b`. However, we soon realized that it is safe to have the joins in the following form `dist_table.a = function2(reference_table.b)`.

[This](https://github.com/citusdata/citus/pull/1884/files#diff-cbde2c0628e90ada0a1da07530fece59R2927) is the change that led to the situation.

An example:
```SQL
CREATE TABLE test_dist (a int, b int);
SELECT create_distributed_table('test_dist', 'a');

CREATE TABLE test_ref (a int, b int);
SELECT create_reference_table('test_ref');

-- success in 7.1, fails in 7.2
SELECT *
FROM test_dist t1
LEFT JOIN test_ref t2 ON t1.a = trunc(t2.a);
```

Note that we should enforce this checks for joins between distributed tables, but, that's OK to relax the checks for reference tables.

onderkalaci:
Fixes #1916 

This change allows unsupported clauses to go through query pushdown
planner instead of erroring out as we already do for non-outer joins.

### DIFF
--- a/src/backend/distributed/planner/multi_logical_planner.c
+++ b/src/backend/distributed/planner/multi_logical_planner.c
@@ -224,7 +224,7 @@ MultiLogicalPlanCreate(Query *originalQuery, Query *queryTree,
 static bool
 ShouldUseSubqueryPushDown(Query *originalQuery, Query *rewrittenQuery)
 {
-	List *whereClauseList = NIL;
+	List *qualifierList = NIL;
 
 	/*
 	 * We check the existence of subqueries in FROM clause on the modified query
@@ -260,8 +260,8 @@ ShouldUseSubqueryPushDown(Query *originalQuery, Query *rewrittenQuery)
 	 * Some unsupported join clauses in logical planner
 	 * may be supported by subquery pushdown planner.
 	 */
-	whereClauseList = WhereClauseList(rewrittenQuery->jointree);
-	if (DeferErrorIfUnsupportedClause(whereClauseList) != NULL)
+	qualifierList = QualifierList(rewrittenQuery->jointree);
+	if (DeferErrorIfUnsupportedClause(qualifierList) != NULL)
 	{
 		return true;
 	}
--- a/src/test/regress/sql/multi_subquery.sql
+++ b/src/test/regress/sql/multi_subquery.sql
@@ -279,6 +279,59 @@ WHERE
 ORDER BY l_orderkey DESC
 LIMIT 10;
 
+-- outer joins on reference tables with functions works
+SELECT DISTINCT ON (t1.user_id) t1.user_id, t2.value_1, t2.value_2, t2.value_3
+FROM events_table t1
+LEFT JOIN users_reference_table t2 ON t1.user_id = trunc(t2.user_id)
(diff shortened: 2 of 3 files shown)


src/test/regress/sql/multi_subquery.sql
```diff
@@ -279,6 +279,59 @@ WHERE
...
+-- outer joins on distributed tables with expressions should not work
+SELECT DISTINCT ON (t1.user_id) t1.user_id, t2.value_1, t2.value_2, t2.value_3
+ FROM 
+ users_table t0 LEFT JOIN
+ events_table t1  ON t0.user_id = trunc(t1.user_id)
+ LEFT JOIN users_reference_table t2 ON t1.user_id = trunc(t2.user_id)
+ ORDER BY 1 DESC, 2 DESC, 3 DESC, 4 DESC 
+ LIMIT 5;
+
+-- inner joins on reference tables with functuions works
```
metdos: functuions -> typo

metdos: It also exists in a couple of places, would help replace all :)

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -260,7 +260,7 @@ ShouldUseSubqueryPushDown(Query *originalQuery, Query *rewrittenQuery)
 	 * Some unsupported join clauses in logical planner
 	 * may be supported by subquery pushdown planner.
 	 */
-	whereClauseList = WhereClauseList(rewrittenQuery->jointree);
+	whereClauseList = QualifierList(rewrittenQuery->jointree);
```
mtuncer: Once we are on this, I would feel better if we also change ```WhereClauseList()``` call in ```MultiNodeTree()``` to ```QualifierList()```. 

Also consider renaming variable name to ```clauseList``` or something like that because it is not where clause list any more

src/test/regress/sql/multi_subquery.sql
```diff
@@ -279,6 +279,59 @@ WHERE
...
+SELECT DISTINCT ON (t1.user_id) t1.user_id, t2.value_1, t2.value_2, t2.value_3
+FROM events_table t1
+LEFT JOIN users_table t2 ON t1.user_id > t2.user_id
+ORDER BY 1 DESC, 2 DESC, 3 DESC, 4 DESC 
+LIMIT 5;
+
+-- outer joins on reference tables with expressions should work
+SELECT DISTINCT ON (t1.user_id) t1.user_id, t2.value_1, t2.value_2, t2.value_3
+FROM events_table t1
+LEFT JOIN users_reference_table t2 ON t1.user_id = (CASE WHEN t2.user_id > 3 THEN 3 ELSE t2.user_id END)
```
mtuncer: can we have multiple columns from t2 table in case expression ?

Issue #83, opened 2016-02-02
We can not partition over column types like json which does not support aggregate functions. We should add a custom error messages for these cases.


Jhanbanan:
We can not partition over column types like json which does not support aggregate functions. We should add a custom error messages for these cases.

PR #1923, opened 2018-01-03
Don't return in PG_TRY() block when cancellations happen in WaitForConnections().

pykello:
We shouldn't return in middle of a `PG_TRY()` block because if we do, we won't reset `PG_exception_stack`, and later when a re-throw tries to jump to the jump-point which was active in this `PG_TRY()` block, it seg-faults.

We used to return in middle of `PG_TRY()` block in `WaitForConnections()` where we checked for cancellations. Whenever cancellations were caught here, Citus crashed. And example was reported by @onderkalaci at https://github.com/citusdata/citus/pull/1903#issuecomment-354454614

Another example, is we execute the following query so many times until cancellation is caught in `WaitForAllConnections()`:
```SQL
BEGIN;
SELECT user_id, pg_advisory_lock(15) FROM users_table;
COMMIT;
```

To understand the problem better, look at how `PG_TRY()`, `PG_CATCH()`, and `PG_END_TRY()` has been defined and refer to the man pages of `sigsetjmp()` and `siglongjmp()`.

I tried hard to find an automated test where the cancellation happened in `WaitForAllConnections()` frequent enough, but I failed. Manually I can reproduce it though to make sure that this change fixes it.

I think this should be back-ported, because this is nothing new, and I think it should be reproducible in previous versions where we check for cancellations in `WaitForAllConnections()`.

### DIFF
--- a/src/backend/distributed/connection/remote_commands.c
+++ b/src/backend/distributed/connection/remote_commands.c
@@ -781,6 +781,7 @@ WaitForAllConnections(List *connectionList, bool raiseInterrupts)
 
 		while (pendingConnectionsStartIndex < totalConnectionCount)
 		{
+			bool cancellationReceived = false;
 			int eventIndex = 0;
 			int eventCount = 0;
 			long timeout = -1;
@@ -837,9 +838,13 @@ WaitForAllConnections(List *connectionList, bool raiseInterrupts)
 					if (InterruptHoldoffCount > 0 && (QueryCancelPending ||
 													  ProcDiePending))
 					{
-						/* return immediately in case of cancellation */
-						FreeWaitEventSet(waitEventSet);
-						return;
+						/*
+						 * Break out of event loop immediately in case of cancellation.
+						 * We cannot use "return" here inside a PG_TRY() block since
+						 * then the exception stack won't be reset.
+						 */
+						cancellationReceived = true;
+						break;
 					}
 
 					continue;
@@ -903,6 +908,11 @@ WaitForAllConnections(List *connectionList, bool raiseInterrupts)
 					 */
 					rebuildWaitEventSet = true;
 				}
+			}
+
+			if (cancellationReceived)
+			{
+				break;
 			}
 
 			/* move non-ready connections to the back of the array */


onderkalaci:
Thanks for the detailed examination and the fix!

> We shouldn't return in middle of a PG_TRY() block because if we do, we won't reset PG_exception_stack, and later when a re-throw tries to jump to the jump-point which was active in this PG_TRY() block, it seg-faults.

That seems correct. I've read few things and [this]( http://web.eecs.utk.edu/~plank/plank/classes/cs360/360/notes/Setjmp/lecture.html) lecture notes seems one of the most comprehensive notes on `setjmp()` / `longjmp()`. See the following quotes from the notes:

> `setjmp()` saves all the registers, including the sp (`stack pointer`) and fp (`frame pointer`). What this means is that if you return from a procedure that calls `setjmp()`, then the env buffer of that `setjmp()` will no longer be valid. Why? Because that env buffer contains the `sp` and `fp` of the calling procedure. If that procedure returns, then when you restore the `sp` and `fp`, the stack will be in a different state than before, and you will have an error.

> This is a very common bug with `setjmp()` and `longjmp()` -- to use them properly, you CANNOT RETURN FROM THE PROCEDURE THAT CALLS `setjmp()`. This is sometimes called "longjmp-ing up the stack." 

Just for reference from the [man page](http://man7.org/linux/man-pages/man3/setjmp.3.html)

>   If the function which called setjmp() returns before longjmp() is called, the behavior is undefined.  Some kind of subtle or unsubtle chaos is sure to result.

(I guess that is the core dump we get is `subtle chaos` 😄 )

Also, I went over (almost) all usage of `PG_TRY()/PG_CATCH()` blocks in Postgres source, which seems to carefully avoid returning in the blocks. I think this should be something we should keep an eye on in the future and not let such code into the mater.

For the tests: 

> I tried hard to find an automated test where the cancellation happened in WaitForAllConnections() frequent enough, but I failed. Manually I can reproduce it though to make sure that this change fixes it.

I can see that how hard it would be to add regression tests without any code changes. Actually, I'm OK with not adding such tests here. 

Just for reference for future, for instance, for online rebalancer tests, we had some tricks to add regression tests, please see `ConflictOnlyWithIsolationTesting() ` and related codes in the enterprise repo. 

We should probably consider adding that to test automation, see (and feel free to expand) the issue on testing automation [#83](https://github.com/citusdata/test-automation/issues/83). In general, the tests that are easier to write with custom scripts are good candidates for test automation.

> I think this should be back-ported, because this is nothing new, and I think it should be reproducible in previous versions where we check for cancellations in WaitForAllConnections().

I agree with that as well. We could maybe wait for @marcocitus to see if he has any objections/comments to that?

src/backend/distributed/connection/remote_commands.c
```diff
@@ -837,9 +838,9 @@ WaitForAllConnections(List *connectionList, bool raiseInterrupts)
 					if (InterruptHoldoffCount > 0 && (QueryCancelPending ||
 													  ProcDiePending))
 					{
-						/* return immediately in case of cancellation */
-						FreeWaitEventSet(waitEventSet);
-						return;
+						/* break out of event loop immediately in case of cancellation */
```
onderkalaci: I think it'd be great if we add a comment on why we're not returning here immediately.

pykello:
On Wed, Jan 3, 2018 at 4:53 AM, Önder Kalacı <notifications@github.com>
wrote:

> setjmp() saves all the registers, including the sp (stack pointer) and fp
> (frame pointer). What this means is that if you return from a procedure
> that calls setjmp(), then the env buffer of that setjmp() will no longer
> be valid. Why? Because that env buffer contains the sp and fp of the
> calling procedure. If that procedure returns, then when you restore the sp
> and fp, the stack will be in a different state than before, and you will
> have an error.
>
> This is a very common bug with setjmp() and longjmp() -- to use them
> properly, you CANNOT RETURN FROM THE PROCEDURE THAT CALLS setjmp(). This
> is sometimes called "longjmp-ing up the stack."
>
>
In addition to this, PG_exception_stack is a **pointer** to setjmp's env
buffer, it is not a copy of it. Since the env buffer is allocated in stack,
its memory location can be overridden after function returns and an unreset
PG_exception_stack might point a location that's not an env buffer at all.

PR #1930, opened 2018-01-04
Prototype: Properly interpret the result of GetRemoteCommandResult()

### LINKED ISSUE #1926: Citus may not behave properly on the first command execution after a worker is stopped
In a hash distributed table, I manually stopped one of the workers that the below query would hit. I was expecting it to return an error message but it returned me empty results. The steps to reproduce the issue is below.

```SQL
create table test_2(a int, b int);
select create_distributed_table('test_2', 'a');
insert into test_2 SELECT x%307, x%1001 FROM generate_series(1,30000) as f(x);

select * from test_2 where test_2.a=100;

  a  |  b
-----+-----
 100 | 100
 100 | 407
 100 | 714
 100 |  20
 100 | 327
 100 | 634
 100 | 941
 100 | 247
 100 | 554
 100 | 861
 100 | 167
 100 | 474
 100 | 781
...
-- continues

SELECT get_shard_id_for_distribution_column('test_2', 100);
 get_shard_id_for_distribution_column
                               102376
```
I checked in which worker the shard is located in and killed it manually by the following command
```
pg_ctl -D DBw1 -l logfile stop -m i
```
After that I run the router select query again
```SQL
postgres=# select * from test_2 where test_2.a=100;
 a | b
---+---
(0 rows)
```

I doubt that this issue might be related to cache invalidation because when we run the same query again, citus gives a connection error as expected.
```SQL
postgres=# select * from test_2 where test_2.a=100;
WARNING:  01000: connection error: localhost:9701
DETAIL:  server closed the connection unexpectedly
could not send SSL negotiation packet: Broken pipe
LOCATION:  ReportConnectionError, remote_commands.c:247
ERROR:  XX000: could not receive query results
LOCATION:  ExecuteSingleSelectTask, multi_router_executor.c:657
Time: 2.116 ms
```

onderkalaci:
This is a WIP PR yet aims to fix #1926

See the [comment](https://github.com/citusdata/citus/issues/1926#issuecomment-355289153) for the details.

Note that this WIP only fixes single occurrence of the same bug, this should be expanded to fix all occurrences as mentioned in the  [comment](https://github.com/citusdata/citus/issues/1926#issuecomment-355289153).

### DIFF
--- a/src/backend/distributed/executor/multi_router_executor.c
+++ b/src/backend/distributed/executor/multi_router_executor.c
@@ -1401,7 +1401,22 @@ StoreQueryResult(CitusScanState *scanState, MultiConnection *connection,
 		bool doRaiseInterrupts = true;
 
 		PGresult *result = GetRemoteCommandResult(connection, doRaiseInterrupts);
-		if (result == NULL)
+
+		if (PQstatus(connection->pgConn) == CONNECTION_BAD)
+		{
+			if (failOnError)
+			{
+				ReportResultError(connection, result, ERROR);
+			}
+			else
+			{
+				ReportResultError(connection, result, WARNING);
+			}
+
+			commandFailed = true;
+			break;
+		}
+		else if (result == NULL)
 		{
 			break;
 		}


marcocitus:
Replaced by #2031

PR #1939, opened 2018-01-10
Add support for renaming Distributed Tables

DimCitus:
This Pull Request implements support for renaming Distributed Tables, fixing #1249. It also adds a commit to edit the regression tests and have more coverage of non-default schema (not public, not in the search_path).

It might be better for me to rebase/squash the main commit and its follow-up make reindent pass though, but will that trigger another travis-ci run?

### DIFF
--- a/src/backend/distributed/relay/relay_event_utility.c
+++ b/src/backend/distributed/relay/relay_event_utility.c
@@ -388,12 +388,37 @@ RelayEventExtendNames(Node *parseTree, char *schemaName, uint64 shardId)
 				char **oldRelationName = &(renameStmt->relation->relname);
 				char **newRelationName = &(renameStmt->newname);
 				char **objectSchemaName = &(renameStmt->relation->schemaname);
+				int newRelationNameLength;
 
 				/* prefix with schema name if it is not added already */
 				SetSchemaNameIfNotExist(objectSchemaName, schemaName);
 
 				AppendShardIdToName(oldRelationName, shardId);
 				AppendShardIdToName(newRelationName, shardId);
+
+				/*
+				 * PostgreSQL creates array types for each ordinary table, with
+				 * the same name plus a prefix of '_'.
+				 *
+				 * ALTER TABLE ... RENAME TO ... also renames the underlying
+				 * array type, and the DDL is run in parallel connections over
+				 * all the placements and shards at once. Concurrent access
+				 * here deadlocks.
+				 *
+				 * Let's provide an easier to understand error message here
+				 * than the deadlock one.
+				 *
+				 * See also https://github.com/citusdata/citus/issues/1664
+				 */
+				newRelationNameLength = strlen(*newRelationName);
+				if (newRelationNameLength >= (NAMEDATALEN - 1))
+				{
+					ereport(ERROR,
+							(errcode(ERRCODE_NAME_TOO_LONG),
+							 errmsg(
+								 "shard name %s exceeds %d characters",
+								 *newRelationName, NAMEDATALEN - 1)));
+				}
 			}
 			else if (objectType == OBJECT_COLUMN || objectType == OBJECT_TRIGGER)
 			{
--- a/src/test/regress/expected/multi_reference_table.out
+++ b/src/test/regress/expected/multi_reference_table.out
@@ -1382,12 +1382,13 @@ SELECT "Column", "Type", "Modifiers" FROM table_desc WHERE relid='public.referen
 (0 rows)
 
 \c - - - :master_port
--- as we expect, renaming and setting WITH OIDS does not work for reference tables
-ALTER TABLE reference_table_ddl RENAME TO reference_table_ddl_test;
-ERROR:  renaming distributed tables is currently unsupported
+-- as we expect, setting WITH OIDS does not work for reference tables
 ALTER TABLE reference_table_ddl SET WITH OIDS;
 ERROR:  alter table command is currently unsupported
 DETAIL:  Only ADD|DROP COLUMN, SET|DROP NOT NULL, SET|DROP DEFAULT, ADD|DROP CONSTRAINT, ATTACH|DETACH PARTITION and TYPE subcommands are supported.
+-- now test the renaming of the table, and back to the expected name
+ALTER TABLE reference_table_ddl RENAME TO reference_table_ddl_test;
+ALTER TABLE reference_table_ddl_test RENAME TO reference_table_ddl;
 -- now test reference tables against some helper UDFs that Citus provides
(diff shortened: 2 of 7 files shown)


src/backend/distributed/relay/relay_event_utility.c
```diff
@@ -644,7 +667,6 @@ AppendShardIdToName(char **name, uint64 shardId)
 	{
 		snprintf(extendedName, NAMEDATALEN, "%s%s", (*name), shardIdAndSeparator);
 	}
-
```
marcocitus: this is the line `citus_indent` is complaining about

src/test/regress/expected/multi_reference_table.out
```diff
@@ -1308,26 +1308,26 @@ SELECT logicalrelid FROM pg_dist_shard WHERE logicalrelid::regclass::text LIKE '
...
  create_reference_table 
 ------------------------
  
 (1 row)
 
 -- CREATE & DROP index and check the workers
-CREATE INDEX reference_index_1 ON reference_table_ddl(value_1);
-CREATE INDEX reference_index_2 ON reference_table_ddl(value_2, value_3);
+CREATE INDEX reference_index_1 ON reference_schema.reference_table_ddl(value_1);
+CREATE INDEX reference_index_2 ON reference_schema.reference_table_ddl(value_2, value_3);
```
marcocitus: couldn't we just use search_path here? that's the pattern followed by some other tests.

src/backend/distributed/relay/relay_event_utility.c
```diff
@@ -394,6 +394,29 @@ RelayEventExtendNames(Node *parseTree, char *schemaName, uint64 shardId)
...
+				 * than the deadlock one.
+				 *
+				 * See also https://github.com/citusdata/citus/issues/1664
+				 */
+				if (strlen(*newRelationName) >= (NAMEDATALEN - 1))
+				{
+					ereport(ERROR,
+							(errcode(ERRCODE_NAME_TOO_LONG),
+							 errmsg(
+								 "distributed identifier must be less than %d characters",
```
marcocitus: maybe: distributed identifier -> shard name

src/backend/distributed/relay/relay_event_utility.c
```diff
@@ -394,6 +394,29 @@ RelayEventExtendNames(Node *parseTree, char *schemaName, uint64 shardId)
...
+				 * array type, and the DDL is run in parallel connections over
+				 * all the placements and shards at once. Concurrent access
+				 * here deadlocks.
+				 *
+				 * Let's provide an easier to understand error message here
+				 * than the deadlock one.
+				 *
+				 * See also https://github.com/citusdata/citus/issues/1664
+				 */
+				if (strlen(*newRelationName) >= (NAMEDATALEN - 1))
```
marcocitus: A minor Citus convention (which I think derives from Code Complete 2) is to avoid function calls in expressions where possible. So the preferred pattern is:

```
newRelationNameLength = strlen(*newRelationName);
if (newRelationNameLength >= (NAMEDATALEN - 1))
{
   ...
}
```

That way we could set a break point on the `if` and see the actual length that causes the expression to evaluate to true.

src/backend/distributed/relay/relay_event_utility.c
marcocitus: could maybe also say "shard name %s exceeds %d characters"

src/test/regress/expected/multi_reference_table.out
DimCitus: Depends on what exactly we want to test. Is the test more interesting with using the search_path or without using it?

marcocitus:
would be good to also add a test on distributed tables, as those are a bit different from reference tables.

src/test/regress/expected/multi_reference_table.out
marcocitus: Both are interesting. Anyway, no strong objections to keep it this way.

DimCitus:
I added a commit on top of the previous two, in order to make it easier to follow-up in this PR. Let me know if/when it should be squashed into the first commit of this branch.

About distributed tables, we already test them in https://github.com/citusdata/citus/blob/feature/rename-table/src/test/regress/sql/multi_name_lengths.sql#L31 ; see https://github.com/citusdata/citus/blob/feature/rename-table/src/test/regress/sql/multi_name_lengths.sql#L66 ; so I think we're good and didn't add another test in the suite. Is that ok?

Other suggestions applied. `make reindent` still doesn't do much locally:

```
$ uncrustify --version
Uncrustify-0.66.1_f

$ make reindent
cd /Users/dim/dev/CitusData/citus && citus_indent --quiet
/usr/local/etc/citustools/citus-style.cfg:788 Unknown symbol 'align_number_left'
```
```

**Comment by marcocitus** [comment]
> ; so I think we're good and didn't add another test in the suite. Is that ok?

this appears to only test a failure case, we have yet to confirm in the regression tests that we can successfully rename a distributed table

**Comment by marcocitus** [comment]
@jasonmp85 any thoughts on the uncrustify error? should we update our config?

**Comment by DimCitus** [comment]
I just added some test cases in their own patch, and rebased against origin/master that changed in between, so that I could then push force here. Once review is all clear, I'll need to rebase/squash my 4 commits into 2 of them, if I understand correctly. I'm thinking doing that as a last step so that we can see the incremental revisions in the meantime.

File: src/test/regress/input/multi_alter_table_statements.source [hunk 5]
```diff
@@ -356,10 +355,30 @@ FROM
...
+\c - - - :worker_1_port
+SELECT relname FROM pg_class WHERE relname LIKE 'lineitem_renamed%';
+\c - - - :master_port
+
+-- revert it to original name
+ALTER TABLE lineitem_renamed RENAME TO lineitem_alter;
+
+-- show rename worked on one worker, too
+\c - - - :worker_1_port
+SELECT relname FROM pg_class WHERE relname LIKE 'lineitem_alter%';
```
marcocitus: It might make sense to add `ORDER BY relname` to prevent the output from changing in different environments (e.g. Travis). I think we need to change existing cases that omit the ORDER BY, but it's a good practice to follow going forward.

DimCitus: And Travis is failing right on that here. I'm only going to edit new tests in this PR, though.

DimCitus:
Oops, in a previous version of the “Tests results should have a known ordering.” I included unwanted changes. Cleaned the patch and git pushed force. Hope that's it now ;-)

marcocitus:
Yep, looks good. Squash and :ship:

PR #1944, opened 2018-01-11
Add base schedule for only running specific regression tests

marcocitus:
This change adds a base schedule to make it easier to run a specific regression test using the `EXTRA_TESTS` flag.

For example, to only run the `with_basics` and `with_join` tests you would run: 
```
make check-base -C src/test/regress/ EXTRA_TESTS="with_basics with_join"
```

which only takes 14 seconds.


metdos:
This looks great 🚢 

Would you like to send an email to dev team after merging this?

PR #1946, opened 2018-01-11
Add test that checks whether distributed transaction ID survives pg_dist_partition invalidation

onderkalaci:
We somehow not able to merge this commit in #1889, probably due to some `git rebase` issue. This PR adds the missing regression test only.

### DIFF
--- a/src/test/regress/expected/isolation_distributed_transaction_id.out
+++ b/src/test/regress/expected/isolation_distributed_transaction_id.out
@@ -77,7 +77,7 @@ step s4-get-all-transactions:
 initiator_node_identifiertransaction_numbertransaction_stamp
 
 
-starting permutation: s1-create-table s1-begin s1-insert s1-get-current-transaction-id s2-get-first-worker-active-transactions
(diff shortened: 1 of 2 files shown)


src/test/regress/expected/isolation_distributed_transaction_id.out
```diff
@@ -110,3 +110,35 @@ step s2-get-first-worker-active-transactions:
...
+
+               
+step s1-has-transaction-number: 
+    SELECT transaction_number > 0 FROM get_current_transaction_id();
+
+?column?       
+
+t              
+step s2-vacuum: 
+	VACUUM FULL pg_dist_partition;
```
marcocitus: alignment is slightly off

onderkalaci:
Hmm, did you check the travis output? It's failing on a test which I'm not familiar with. Do you think is that OK to restart the travis or do you want to look a bit?

marcocitus:
yeh, seems to be a replication race condition. We should maybe use synchronous_commit = remote_apply.

PR #1863, opened 2017-12-07
Add support for ALTER TABLE ... SET (storage_parameter = ...)

byucesoy:
Fixes #1202 

This came up couple of times in customer conversations and it is trivial
to support.

PR #1948, opened 2018-01-11
Add support for Alter Index Commands

### LINKED ISSUE #354: Add support for Alter Index Commands
Found this issue while testing index propagation. create/drop index commands are reflected to worker nodes, but alter index is not.

Renaming index on distributed tables is not supported. However, we do not give an error message and let master node rename the index locally. This leads to having different index names on master and worker nodes. You can not drop or do any distributed operations on this index afterwards.

It is not exactly a corruption since you can recover from it but I think it should be fixed.

Repro steps
uph(a int, b int) is a hash distributed table with 8 shards.

``` SQL
postgres=# create index ix on uph(a);
CREATE INDEX
postgres=# alter index ix rename to hix;
ALTER INDEX
postgres=# drop index ix;
ERROR:  index "ix" does not exist
postgres=# drop index hix;
WARNING:  could not receive query results from 10.179.173.10:5432
DETAIL:  Client error: index "hix_102091" does not exist
ERROR:  could not execute DDL command on worker node shards
```

You can recover from this by

``` sql
postgres=# alter index hix rename to ix;
ALTER INDEX
postgres=# drop index hix;
ERROR:  index "hix" does not exist
postgres=# drop index ix;
DROP INDEX
```

@samay-sharma  suggested this issue is more widespread than I observed. we should respond to 
`alter schema`, `alter tablespace`, `alter database`  commands too.

### LINKED ISSUE #1202: `ALTER TABLE [table] SET (storage_parameter = [value])` is not supported
This came up [in the context of autovacuum](https://github.com/citusdata/citus/issues/1186) where it's currently a little painful to set per-table autovacuum settings, but I can also imagine pushing down per-table `fillfactor`s being useful.

### LINKED ISSUE #1747: CREATE INDEX %s ON %s WITH ( storage_parameter = value [, ... ] ) is not supported on distributed table
`WITH` is used in postgres to define a couple of variables in indexes such as 
```SQL
CREATE UNIQUE INDEX u_ind3 on t(i) with (fillfactor=50);
```
means the 50% of the index page will be left empty. There are a couple of other options for different kind of indexes. 

When I run some queries such as;
``` SQL
CREATE TABLE test(i int);
CREATE UNIQUE INDEX u_ind ON test(i) WITH (fillfactor=50);
SELECT create_distributed_table('test', 'i');
 create_distributed_table
--------------------------

(1 row)
```
It successfully distributes and shards have the index as it is.

However, 
```SQL
CREATE TABLE test2(i int);
SELECT create_distributed_table('test2', 'i');
CREATE UNIQUE INDEX u_ind2 ON test2(i) WITH (fillfactor=50);
ERROR:  42601: syntax error at or near "OPTIONS"
LOCATION:  ReportResultError, remote_commands.c:289
Time: 106.297 ms
```

When I check the log file of the worker, I see the following
```
2017-10-31 11:47:57.032 +03 [63029] ERROR:  syntax error at or near "OPTIONS" at character 83
2017-10-31 11:47:57.032 +03 [63029] STATEMENT:  CREATE UNIQUE INDEX   u_ind2_102199 ON public.test2_102199 USING btree (i ) WITH  OPTIONS (fillfactor '50')
```
however, the master_get_table_ddl_events('test2') gives us the following;
```SQL
 CREATE UNIQUE INDEX u_ind2 ON public.test2 USING btree (i) WITH (fillfactor='50') TABLESPACE pg_default;
```

So, there is a problem on this with Citus and there is no available tests for this right now.
The problem is that we cannot create an index with `WITH` clause on distributed table. However, we do distribute a table even if it has an index with a `WITH` clause on its definition.

Thanks to @onderkalaci , I realized this while working on #1739

DimCitus:
DESCRIPTION: Adds support for ALTER INDEX (SET|RESET|RENAME TO) commands
DESCRIPTION: Adds support for setting storage parameters on distributed tables

Currently this branch and Pull Request only contains support for ALTER INDEX ... RENAME TO ... on distributed relations from Citus. Before considering a merge, we're going to see about adding other ALTER INDEX sub-commands (well, SET and RESET).

The last commands to support would then be ALTER INDEX ... SET TABLESPACE ..., and that depends on tablespace support in Citus. It's out of scope here.

In passing other bugs where fixed, so before considering a merge we'll need to split and squash the patches into a good series, to be defined. Also WIP commit messages include GitHub issue and PR numbers, this will need a clean up before merge.

Fixes #354
Fixes #1747 
Fixes #1202
See  #1863  (duplicate? sorry)

### DIFF
--- a/src/backend/distributed/executor/multi_utility.c
+++ b/src/backend/distributed/executor/multi_utility.c
@@ -139,6 +139,7 @@ static void ErrorIfUnstableCreateOrAlterExtensionStmt(Node *parsetree);
 static void ErrorIfUnsupportedIndexStmt(IndexStmt *createIndexStatement);
 static void ErrorIfUnsupportedDropIndexStmt(DropStmt *dropIndexStatement);
 static void ErrorIfUnsupportedAlterTableStmt(AlterTableStmt *alterTableStatement);
+static void ErrorIfUnsupportedAlterIndexStmt(AlterTableStmt *alterTableStatement);
 static void ErrorIfAlterDropsPartitionColumn(AlterTableStmt *alterTableStatement);
 static void ErrorIfUnsupportedSeqStmt(CreateSeqStmt *createSeqStmt);
 static void ErrorIfDistributedAlterSeqOwnedBy(AlterSeqStmt *alterSeqStmt);
@@ -366,7 +367,8 @@ multi_ProcessUtility(PlannedStmt *pstmt,
 		if (IsA(parsetree, AlterTableStmt))
 		{
 			AlterTableStmt *alterTableStmt = (AlterTableStmt *) parsetree;
-			if (alterTableStmt->relkind == OBJECT_TABLE)
+			if (alterTableStmt->relkind == OBJECT_TABLE ||
+				alterTableStmt->relkind == OBJECT_INDEX)
 			{
 				ddlJobs = PlanAlterTableStmt(alterTableStmt, queryString);
 			}
@@ -1238,6 +1240,7 @@ PlanAlterTableStmt(AlterTableStmt *alterTableStatement, const char *alterTableCo
 	LOCKMODE lockmode = 0;
 	Oid leftRelationId = InvalidOid;
 	Oid rightRelationId = InvalidOid;
+	char leftRelationKind;
 	bool isDistributedRelation = false;
 	List *commandList = NIL;
 	ListCell *commandCell = NULL;
@@ -1255,13 +1258,38 @@ PlanAlterTableStmt(AlterTableStmt *alterTableStatement, const char *alterTableCo
 		return NIL;
 	}
 
+	/*
+	 * AlterTableStmt applies also to INDEX relations, and we have support for
+	 * SET/SET storage parameters in Citus, so we might have to check for
+	 * another relation here.
+	 */
+	leftRelationKind = get_rel_relkind(leftRelationId);
+	if (leftRelationKind == RELKIND_INDEX)
+	{
+		leftRelationId = IndexGetRelation(leftRelationId, false);
+	}
+
 	isDistributedRelation = IsDistributedTable(leftRelationId);
 	if (!isDistributedRelation)
 	{
 		return NIL;
 	}
 
-	ErrorIfUnsupportedAlterTableStmt(alterTableStatement);
+	/*
+	 * The PostgreSQL parser dispatches several commands into the node type
+	 * AlterTableStmt, from ALTER INDEX to ALTER SEQUENCE or ALTER VIEW. Here
+	 * we have a special implementation for ALTER INDEX, and a specific error
+	 * message in case of unsupported sub-command.
+	 */
+	if (leftRelationKind == RELKIND_INDEX)
+	{
+		ErrorIfUnsupportedAlterIndexStmt(alterTableStatement);
+	}
+	else
+	{
+		/* this function also accepts more than just RELKIND_RELATION... */
+		ErrorIfUnsupportedAlterTableStmt(alterTableStatement);
+	}
 
 	/*
 	 * We check if there is a ADD FOREIGN CONSTRAINT command in sub commands list.
@@ -2011,9 +2039,9 @@ ErrorIfUnsupportedDropIndexStmt(DropStmt *dropIndexStatement)
 
 
 /*
- * ErrorIfUnsupportedAlterTableStmt checks if the corresponding alter table statement
- * is supported for distributed tables and errors out if it is not. Currently,
- * only the following commands are supported.
+ * ErrorIfUnsupportedAlterTableStmt checks if the corresponding alter table
+ * statement is supported for distributed tables and errors out if it is not.
+ * Currently, only the following commands are supported.
  *
  * ALTER TABLE ADD|DROP COLUMN
  * ALTER TABLE ALTER COLUMN SET DATA TYPE
@@ -2021,6 +2049,8 @@ ErrorIfUnsupportedDropIndexStmt(DropStmt *dropIndexStatement)
  * ALTER TABLE SET|DROP DEFAULT
  * ALTER TABLE ADD|DROP CONSTRAINT
  * ALTER TABLE REPLICA IDENTITY
+ * ALTER TABLE SET ()
+ * ALTER TABLE RESET ()
  */
 static void
 ErrorIfUnsupportedAlterTableStmt(AlterTableStmt *alterTableStatement)
@@ -2170,14 +2200,71 @@ ErrorIfUnsupportedAlterTableStmt(AlterTableStmt *alterTableStatement)
 				break;
 			}
 
+			case AT_SetRelOptions:  /* SET (...) */
+			case AT_ResetRelOptions:    /* RESET (...) */
+			case AT_ReplaceRelOptions:  /* replace entire option list */
+			{
+				/* this command is supported by Citus */
+				break;
+			}
+
 			default:
 			{
-				ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
-								errmsg("alter table command is currently unsupported"),
-								errdetail("Only ADD|DROP COLUMN, SET|DROP NOT NULL, "
-										  "SET|DROP DEFAULT, ADD|DROP CONSTRAINT, "
-										  "ATTACH|DETACH PARTITION and TYPE subcommands "
-										  "are supported.")));
+				ereport(ERROR,
+						(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+						 errmsg("alter table command is currently unsupported"),
+						 errdetail("Only ADD|DROP COLUMN, SET|DROP NOT NULL, "
+								   "SET|DROP DEFAULT, ADD|DROP CONSTRAINT, "
+								   "SET (), RESET (), "
+								   "ATTACH|DETACH PARTITION and TYPE subcommands "
+								   "are supported.")));
+			}
+		}
+	}
+}
+
+
+/*
+ * ErrorIfUnsupportedAlterIndexStmt checks if the corresponding alter index
+ * statement is supported for distributed tables and errors out if it is not.
+ * Currently, only the following commands are supported.
+ *
+ * ALTER INDEX SET ()
+ * ALTER INDEX RESET ()
+ */
+static void
+ErrorIfUnsupportedAlterIndexStmt(AlterTableStmt *alterTableStatement)
+{
+	List *commandList = alterTableStatement->cmds;
+	ListCell *commandCell = NULL;
+
+	/* error out if any of the subcommands are unsupported */
+	foreach(commandCell, commandList)
+	{
+		AlterTableCmd *command = (AlterTableCmd *) lfirst(commandCell);
+		AlterTableType alterTableType = command->subtype;
+
+		switch (alterTableType)
+		{
+			case AT_SetRelOptions:  /* SET (...) */
+			case AT_ResetRelOptions:    /* RESET (...) */
+			case AT_ReplaceRelOptions:  /* replace entire option list */
+			{
+				/* this command is supported by Citus */
+				break;
+			}
+
+			/* unsupported create index statements */
+			case AT_SetTableSpace:
+			default:
+			{
+				ereport(ERROR,
+						(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+						 errmsg("alter index ... set tablespace ... "
+								"is currently unsupported"),
+						 errdetail("Only RENAME TO, SET (), and RESET () "
+								   "are supported.")));
+				return; /* keep compiler happy */
 			}
 		}
 	}
--- a/src/backend/distributed/executor/multi_utility.c
+++ b/src/backend/distributed/executor/multi_utility.c
@@ -155,6 +155,7 @@ static void ErrorIfUnsupportedForeignConstraint(Relation relation,
 static char * ExtractNewExtensionVersion(Node *parsetree);
 static void CreateLocalTable(RangeVar *relation, char *nodeName, int32 nodePort);
 static bool IsAlterTableRenameStmt(RenameStmt *renameStmt);
+static bool IsIndexRenameStmt(RenameStmt *renameStmt);
 static bool AlterInvolvesPartitionColumn(AlterTableStmt *alterTableStatement,
 										 AlterTableCmd *command);
 static void ExecuteDistributedDDLJob(DDLJob *ddlJob);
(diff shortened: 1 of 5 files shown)


DimCitus:
About the latest patch, I'm using a much simpler way to get to the relation itself or the relation that the index is created on:

```c
	/*
	 * AlterTableStmt applies also to INDEX relations, and we have support for
	 * SET/SET storage parameters in Citus, so we might have to check for
	 * another relation here.
	 */
	if (get_rel_relkind(leftRelationId) == RELKIND_INDEX)
	{
		leftRelationId = IndexGetRelation(leftRelationId, false);
	}
```

I am not sure that the `PlanRenameStmt` implementation should match this one, because of the convoluted/complex logic around the cases when to support, ignore or error out depending on the commands and if the target relation is managed by Citus or not.

It might be that it was easier to get my head around the mechanisms in place in the second run though, but if we choose to simplify `PlanRenameStmt` I'd argue to keep the level of comments I added thanks to the new implementation.

src/test/regress/output/multi_alter_table_statements.source
```diff
@@ -844,6 +844,156 @@ SELECT relname FROM pg_class WHERE relname LIKE 'lineitem_alter%' ORDER BY relna
...
+ lineitem_alter_220000 | {fillfactor=40}
+ lineitem_alter_220001 | {fillfactor=40}
+ lineitem_alter_220002 | {fillfactor=40}
+ lineitem_alter_220003 | {fillfactor=40}
+ lineitem_alter_220004 | {fillfactor=40}
+ lineitem_alter_220005 | {fillfactor=40}
+ lineitem_alter_220006 | {fillfactor=40}
+ lineitem_alter_220007 | {fillfactor=40}
+ lineitem_alter_220008 | {fillfactor=40}
+ lineitem_alter_220009 |
```
marcocitus: Wonder what's happening here.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -1371,11 +1383,17 @@ PlanAlterTableStmt(AlterTableStmt *alterTableStatement, const char *alterTableCo
 static List *
 PlanRenameStmt(RenameStmt *renameStmt, const char *renameCommand)
 {
-	Oid relationId = InvalidOid;
+	Oid objRelationId = InvalidOid; /* SQL Object OID */
+	Oid tblRelationId = InvalidOid; /* Relation OID, maybe not the same. */
```
marcocitus: For readability, we prefer full words over acronyms, i.e. objectRelationId, tableRelationId.

src/test/regress/output/multi_alter_table_statements.source
```diff
@@ -844,6 +844,156 @@ SELECT relname FROM pg_class WHERE relname LIKE 'lineitem_alter%' ORDER BY relna
 (15 rows)
 
 \c - - - :master_port
+-- verify that we can set and reset storage parameters
+ALTER TABLE lineitem_alter SET(fillfactor=40);
```
marcocitus: We should also verify these commands are correctly propagated when creating new shards (or moving one), for example by running them on a local table and then running create_distributed_table.

For the commands to be applied to a new shard, master_get_table_ddl_events needs to return the appropriate command.

DimCitus:
This patch should address all the points above and in particular implement the needed support for storage options in `master_get_table_ddl_events`.

The test strange result is because of the `\copy` failure test.

Again the plan is the get the PR all-green then squash the branch to a single commit that we can merge, and I think I will have to rebase against current master, too.

src/test/regress/output/multi_alter_table_statements.source
```diff
@@ -1021,3 +1130,67 @@ SELECT relname FROM pg_class WHERE relname LIKE 'test_table_1%';
...
+-----------------------+-----------------
+ hash_dist_pkey_220033 | {fillfactor=40}
+ hash_dist_pkey_220034 | {fillfactor=40}
+ hash_dist_pkey_220035 | {fillfactor=40}
+ hash_dist_pkey_220036 | {fillfactor=40}
+(4 rows)
+
+\c - - - :master_port
+ALTER INDEX idx_lineitem_linenumber RESET(fillfactor);
+ERROR:  relation "idx_lineitem_linenumber" does not exist
```
marcocitus: should this be `hash_dist_pkey`?

src/backend/distributed/executor/multi_utility.c
```diff
@@ -2733,14 +2807,14 @@ OptionsSpecifyOwnedBy(List *optionList, Oid *ownedByTableId)
...
+ * At the moment of writing this comment, it could be either tables or indexes.
  */
 static void
 ErrorIfUnsupportedRenameStmt(RenameStmt *renameStmt)
 {
-	Assert(IsAlterTableRenameStmt(renameStmt));
-
-	if (renameStmt->renameType == OBJECT_TABCONSTRAINT)
+	if (IsAlterTableRenameStmt(renameStmt) &&
+		renameStmt->renameType == OBJECT_TABCONSTRAINT)
```
marcocitus: Is there anything stopping us from renaming constraints?

marcocitus:
Slightly strange message for unsupported ALTER INDEX:
```
ALTER INDEX test_pkey SET TABLESPACE foo;
ERROR:  alter table command is currently unsupported
```

src/backend/distributed/executor/multi_utility.c
```diff
@@ -2125,12 +2190,21 @@ ErrorIfUnsupportedAlterTableStmt(AlterTableStmt *alterTableStatement)
...
+				break;
+			}
+
 			default:
 			{
 				ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
 								errmsg("alter table command is currently unsupported"),
 								errdetail("Only ADD|DROP COLUMN, SET|DROP NOT NULL, "
 										  "SET|DROP DEFAULT, ADD|DROP CONSTRAINT, "
+										  "SET (), RESET (), "
```
marcocitus: I wonder whether we should start listing what is not supported.

src/test/regress/output/multi_alter_table_statements.source
DimCitus: oops, too much copy/pasting around those regression test files, good catch!

src/backend/distributed/executor/multi_utility.c
DimCitus: All I'm seeing at the moment in the code is this error message in src/backend/distributed/relay/relay_event_utility.c: _"unsafe object type in rename statement"_. I had guessed it's been studied and punt off before already, should we see about adding support for it now?

src/backend/distributed/executor/multi_utility.c
```diff
@@ -2125,14 +2194,49 @@ ErrorIfUnsupportedAlterTableStmt(AlterTableStmt *alterTableStatement)
...
-				ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
-								errmsg("alter table command is currently unsupported"),
-								errdetail("Only ADD|DROP COLUMN, SET|DROP NOT NULL, "
-										  "SET|DROP DEFAULT, ADD|DROP CONSTRAINT, "
-										  "ATTACH|DETACH PARTITION and TYPE subcommands "
-										  "are supported.")));
+				switch (leftRelationKind)
+				{
+					case RELKIND_RELATION:
+					{
```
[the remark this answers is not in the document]
DimCitus: Thanks for catching also this, I was mistakenly trying to minimize the footprint of the patch. I guess I was too eager to be done. Hopefully we have reviews!

src/backend/distributed/executor/multi_utility.c
```diff
@@ -2125,14 +2201,71 @@ ErrorIfUnsupportedAlterTableStmt(AlterTableStmt *alterTableStatement)
...
+		AlterTableCmd *command = (AlterTableCmd *) lfirst(commandCell);
+		AlterTableType alterTableType = command->subtype;
+
+		switch (alterTableType)
+		{
+			case AT_SetRelOptions:  /* SET (...) */
+			case AT_ResetRelOptions:    /* RESET (...) */
+			case AT_ReplaceRelOptions:  /* replace entire option list */
+			{
+				/* no special processing needs to be done by Citus here. */
```
marcocitus: I don't think this comment is accurate. Maybe `/* command is supported by Citus */`?

src/backend/distributed/executor/multi_utility.c
```diff
@@ -1385,32 +1420,71 @@ PlanRenameStmt(RenameStmt *renameStmt, const char *renameCommand)
...
 
 	/* we have no planning to do unless the table is distributed */
-	isDistributedRelation = IsDistributedTable(relationId);
+	switch (renameStmt->renameType)
+	{
+		case OBJECT_TABLE:
+		case OBJECT_COLUMN:
+		case OBJECT_TABCONSTRAINT:
+		{
+			/* the target object is our tblRelationId. */
```
marcocitus: tbl -> table

src/backend/distributed/executor/multi_utility.c
```diff
@@ -1385,32 +1420,71 @@ PlanRenameStmt(RenameStmt *renameStmt, const char *renameCommand)
...
+			tableRelationId = IndexGetRelation(objectRelationId, false);
+			break;
+		}
+
+		default:
+
+			/*
+			 * Nodes that are not supported by Citus: we pass-through to the
+			 * main PostgreSQL executor. Any Citus-supported RenameStmt
+			 * renameType must appear above in the switch, explicitely.
```
marcocitus: explicitely -> explicitly

src/backend/distributed/utils/citus_ruleutils.c
```diff
@@ -972,16 +992,23 @@ generate_qualified_relation_name(Oid relid)
...
+ * prefix is prepended to the option list, examples: OPTIONS or WITH
+ * separator separates lists elements, example: "," (comma)
+ * operator separates lists keys and values, typically either NULL or "="
  */
 static void
-AppendOptionListToString(StringInfo stringBuffer, List *optionList)
+AppendOptionListToString(StringInfo stringBuffer, List *optionList,
+						 const char *prefix,
+						 const char *separator,
+						 const char *operator)
```
marcocitus: I'm not a huge fan of adding all these parameters. This function is doing something fairly trivial, but now you need to consult the call sites to know what exactly it does. What if we just have separate functions for OPTIONS and WITH since they clearly follow a different syntax?

src/backend/distributed/executor/multi_utility.c
```diff
@@ -1385,32 +1420,71 @@ PlanRenameStmt(RenameStmt *renameStmt, const char *renameCommand)
...
+			 */
+			return NIL;
+	}
+
+	isDistributedRelation = IsDistributedTable(tableRelationId);
+
```
marcocitus: super minor style/consistency thingy: we usually don't have an empty line between the `if` statement and the result it checks.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -2125,14 +2201,71 @@ ErrorIfUnsupportedAlterTableStmt(AlterTableStmt *alterTableStatement)
...
-										  "SET|DROP DEFAULT, ADD|DROP CONSTRAINT, "
-										  "ATTACH|DETACH PARTITION and TYPE subcommands "
-										  "are supported.")));
+				ereport(ERROR,
+						(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+						 errmsg("alter table command is currently unsupported"),
+						 errdetail("Only ADD|DROP COLUMN, SET|DROP NOT NULL, "
+								   "SET|DROP DEFAULT, ADD|DROP CONSTRAINT, "
+								   "SET (), RESET (), "
+								   "ATTACH|DETACH PARTITION and TYPE subcommands "
```
marcocitus: I suppose RENAME TO is missing from this list, but not sure how long we should allow this list to grow.

src/backend/distributed/executor/multi_utility.c
marcocitus: I don't think it's needed in this PR. We might also leave it to another new hire ;).

src/backend/distributed/utils/citus_ruleutils.c
DimCitus: The alternative is to have two function with almost the same content, but it's only about 30 lines each. Let's try that and see if we like it better (in a patch revision coming soon).

src/backend/distributed/executor/multi_utility.c
DimCitus: Right. The only thing we do in “here” is erroring out on unsupported commands, after all...

DimCitus:
Ok I rebased against master's branch and squashed commits together to have a PR that looks like it could be considered for merging. It nows contains 3 patches which I think should be kept separate.

src/backend/distributed/utils/citus_ruleutils.c
```diff
@@ -1004,6 +1022,42 @@ AppendOptionListToString(StringInfo stringBuffer, List *optionList)
...
 
 
+/*
+ * AppendStorageParametersToString converts the storage parameter list to its
+ * textual format, and appends this text to the given string buffer.
+ */
+static void
+AppendStorageParametersToString(StringInfo stringBuffer, List *optionList)
+{
+	if (optionList != NIL)
```
marcocitus: We have the same `if` statement at the call site. What's the appropriate place for it?

DimCitus:
I now see that if I open the commits in GH there are some comments not reported in the main “discussion” page of the Pull Request. I'm going to go through all of them and then push again.

src/test/regress/output/multi_alter_table_statements.source
DimCitus: This is now tested too.

src/backend/distributed/executor/multi_utility.c
DimCitus: this is changed now

src/backend/distributed/executor/multi_utility.c
DimCitus: ok

src/backend/distributed/executor/multi_utility.c
DimCitus: ok

src/backend/distributed/executor/multi_utility.c
DimCitus: Or I'll do it next to close the gap. I kind of read the comment as “here be dragons”, if it's only “didn't get there yet” that's a different story... for another PR ok.

src/backend/distributed/executor/multi_utility.c
marcocitus: bump

src/backend/distributed/executor/multi_utility.c
```diff
@@ -2125,14 +2200,71 @@ ErrorIfUnsupportedAlterTableStmt(AlterTableStmt *alterTableStatement)
...
+ * ALTER INDEX RESET ()
+ */
+static void
+ErrorIfUnsupportedAlterIndexStmt(AlterTableStmt *alterTableStatement)
+{
+	List *commandList = alterTableStatement->cmds;
+	ListCell *commandCell = NULL;
+
+	/* error out if any of the subcommands are unsupported */
+	foreach(commandCell, commandList)
```
marcocitus: I suppose there can only be one subcommand? Should we make that more explicit by just taking the first item? and asserting that there is 1?

src/backend/distributed/executor/multi_utility.c
marcocitus: This has been fixed in `ErrorIfUnsupportedAlterTableStmt`, could we update it here as well? I'm still not sure what is meant by "special processing" :).

PR #1959, opened 2018-01-18
Remove variable length arrays

lithp:
VLAs aren't supported by Visual Studio.

- Remove all existing instances of VLAs.
- Add a flag, -Werror=vla, which makes gcc refuse to compile if we add
  VLAs in the future.

### DIFF
--- a/src/backend/distributed/connection/remote_commands.c
+++ b/src/backend/distributed/connection/remote_commands.c
@@ -744,15 +744,11 @@ WaitForAllConnections(List *connectionList, bool raiseInterrupts)
 	int connectionIndex = 0;
 	ListCell *connectionCell = NULL;
 
-	MultiConnection *allConnections[REMOTE_MAX_CONNECTIONS];
-	WaitEvent events[REMOTE_MAX_CONNECTIONS];
-	bool connectionReady[REMOTE_MAX_CONNECTIONS];
+	MultiConnection **allConnections =
+		palloc(totalConnectionCount * sizeof(MultiConnection *));
+	WaitEvent *events = palloc(totalConnectionCount * sizeof(WaitEvent));
+	bool *connectionReady = palloc(totalConnectionCount * sizeof(bool));
 	WaitEventSet *waitEventSet = NULL;
-
-	if (totalConnectionCount > REMOTE_MAX_CONNECTIONS)
-	{
-		ereport(ERROR, (errmsg("too many connections")));
-	}
 
 	/* convert connection list to an array such that we can move items around */
 	foreach(connectionCell, connectionList)
@@ -765,8 +761,7 @@ WaitForAllConnections(List *connectionList, bool raiseInterrupts)
 	}
 
 	/* make an initial pass to check for failed and idle connections */
-	for (connectionIndex = pendingConnectionsStartIndex;
-		 connectionIndex < totalConnectionCount; connectionIndex++)
+	for (connectionIndex = 0; connectionIndex < totalConnectionCount; connectionIndex++)
 	{
 		MultiConnection *connection = allConnections[connectionIndex];
 
@@ -953,6 +948,10 @@ WaitForAllConnections(List *connectionList, bool raiseInterrupts)
 			FreeWaitEventSet(waitEventSet);
 			waitEventSet = NULL;
 		}
+
+		pfree(allConnections);
+		pfree(events);
+		pfree(connectionReady);
 	}
 	PG_CATCH();
 	{
@@ -962,6 +961,10 @@ WaitForAllConnections(List *connectionList, bool raiseInterrupts)
 			FreeWaitEventSet(waitEventSet);
 			waitEventSet = NULL;
 		}
+
+		pfree(allConnections);
+		pfree(events);
+		pfree(connectionReady);
 
 		PG_RE_THROW();
 	}
--- a/src/backend/distributed/connection/remote_commands.c
+++ b/src/backend/distributed/connection/remote_commands.c
@@ -744,10 +744,15 @@ WaitForAllConnections(List *connectionList, bool raiseInterrupts)
 	int connectionIndex = 0;
 	ListCell *connectionCell = NULL;
 
-	MultiConnection *allConnections[totalConnectionCount];
-	WaitEvent events[totalConnectionCount];
-	bool connectionReady[totalConnectionCount];
+	MultiConnection *allConnections[REMOTE_MAX_CONNECTIONS];
+	WaitEvent events[REMOTE_MAX_CONNECTIONS];
+	bool connectionReady[REMOTE_MAX_CONNECTIONS];
 	WaitEventSet *waitEventSet = NULL;
+
+	if (totalConnectionCount > REMOTE_MAX_CONNECTIONS)
+	{
+		ereport(ERROR, (errmsg("too many connections")));
+	}
 
 	/* convert connection list to an array such that we can move items around */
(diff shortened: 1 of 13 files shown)


lithp:
Oops, looks like we've added one since I made this commit. I'll update the PR tomorrow.

lithp:
We should consider removing `./configure` from the repository. We don't change `configure.in` very often but this is just asking for merge conflicts.

lithp:
This will need to be updated once #1961 is merged.

lithp:
Okay, updated this; it's ready for review again.

byucesoy:
Just to note; we change configure.in for each release, but if you find a way to reliably reproduce runstatedir at each setup, that would be amazing. My setup (with ubuntu 16.04) also adds it, but I guess in mac computers it is not added. [Here](https://github.com/citusdata/citus/pull/1769/files) is a starting point

src/backend/distributed/transaction/distributed_deadlock_detection.c
```diff
@@ -230,25 +235,28 @@ CheckForDistributedDeadlocks(void)
...
+	transactionNodeStack[rootStackDepth] = startingTransactionNode;
 
-	PrependOutgoingNodesToQueue(startingTransactionNode, currentStackDepth,
+	PrependOutgoingNodesToQueue(startingTransactionNode, rootStackDepth,
 								&toBeVisitedNodes);
 
 	/* traverse the graph and search for the deadlocks */
 	while (toBeVisitedNodes != NIL)
 	{
+		int currentStackDepth;
```
pykello: So we have agreed upon not initializing variables?

src/backend/distributed/transaction/distributed_deadlock_detection.c
```diff
@@ -134,7 +134,12 @@ CheckForDistributedDeadlocks(void)
 	{
 		bool deadlockFound = false;
 		List *deadlockPath = NIL;
-		TransactionNode *transactionNodeStack[edgeCount + 1];
+
+		/*
+		 * since we only see nodes which are waiting or being waited upon it's not
```
pykello: Multiline comments start with capital letter.

src/backend/distributed/transaction/distributed_deadlock_detection.c
jasonmp85: 😈

jasonmp85:
> We should consider removing ./configure from the repository. We don't change configure.in very often but this is just asking for merge conflicts.

Also a :-1: for this… we made several changes here for the phone-home feature and make tweaks here-and-there every so often. We should figure out our problems with it instead (if they're indeed fixable, which last time I looked at @byucesoy's suggestion, I could not figure out how to do so).

src/backend/distributed/transaction/distributed_deadlock_detection.c
lithp: If you want to push back on this I'm down to talk about it but it's a silly convention and I'm willing to block progress until we either stop enforcing initialization to always happen or stop requiring that variables are only declared at the beginning of a scope. This is a 10 minute conversation after which we decide to stop doing the wrong thing for, and I use this word carefully, _literally_ zero cost.

jasonmp85: I'm 100% on board with the convention change and have noticed you sneaking it in inside various patches. I agree the benefits of not initializing are real and I don't know when the benefits of initializing actually exist (at least given our compiler flags, which will catch failing to initialize).

src/backend/distributed/metadata/metadata_sync.c
```diff
@@ -813,11 +813,10 @@ static void
 MarkNodeHasMetadata(char *nodeName, int32 nodePort, bool hasMetadata)
 {
 	const bool indexOK = false;
-	const int scanKeyCount = 2;
```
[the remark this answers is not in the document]
lithp: I couldn't decide what to do. I mostly wanted a constant, and a `#define` was the only way I could make one. I guess embedding the number is more consistent.

src/backend/distributed/connection/remote_commands.c
```diff
@@ -744,11 +744,16 @@ WaitForAllConnections(List *connectionList, bool raiseInterrupts)
...
-	WaitEvent events[totalConnectionCount];
-	bool connectionReady[totalConnectionCount];
+	MultiConnection *allConnections[REMOTE_MAX_CONNECTIONS];
+	WaitEvent events[REMOTE_MAX_CONNECTIONS];
+	bool connectionReady[REMOTE_MAX_CONNECTIONS];
 	WaitEventSet *waitEventSet = NULL;
 
+	if (totalConnectionCount > REMOTE_MAX_CONNECTIONS)
+	{
+		ereport(ERROR, (errmsg("too many connections")));
```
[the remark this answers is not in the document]
lithp: you're right, pushed a fix

src/backend/distributed/transaction/distributed_deadlock_detection.c
lithp: pushed a fix

lithp:
@pykello okay, pushed a fix which addresses all of your comments. I'll need to do some rebasing to cleanup this patchset but think the diff is ready to be reviewed again

lithp:
> we made several changes here for the phone-home feature and make tweaks here-and-there every so often

@jasonmp85 I'm not sure what that's arguing against. I'm not saying we remove `configure.in`! I'm suggesting we remove `./configure` and require that anyone building citus have `autoconf` installed. My prior experience with `autoconf` makes me rather disinclined to look into exactly what's happening. And there's not really a need to. Rather than try to stabilize an autogenerated file it's much easier to just not check in said file.

That said, this isn't really related to this PR. Probably I'll open an issue.

jasonmp85:
@lithp Oops! I completely misinterpreted what you said. Yeah, I suppose we could require that users have `autoconf` installed. At minimum it would force people to realize it's actually something our project uses (which I think many devs are ignorant of, having never needed to update it).

lithp:
@pykello did the squash, just waiting on your feedback :)

lithp:
@jasonmp85: opened [an issue](https://github.com/citusdata/citus/issues/1984), don't know if you have any feedback on it.

src/backend/distributed/connection/remote_commands.c
```diff
@@ -958,6 +959,10 @@ WaitForAllConnections(List *connectionList, bool raiseInterrupts)
 			waitEventSet = NULL;
 		}
 
+		pfree(allConnections);
+		pfree(events);
+		pfree(connectionReady);
```
pykello: I think I was wrong when I said we have single point of return :D We should also free these when no exception is thrown. That is, at the end of the `PG_TRY()` block before `PG_CATCH()`. I think we should probably refactor this big function as it is not very easy to read.

lithp: oops, for some reason I thought this was a `try-finally`.

lithp:
okay, added the missing pfree and rebased

lithp:
Completely agree that this function could use a refactor. It would probably be easier to understand if it used a hashtable of connections instead of an array.

src/backend/distributed/connection/remote_commands.c
lithp: Completely agree that this function could use a refactor. It would probably be easier to understand if it used a hashtable of connections instead of an array.

PR #1961, opened 2018-01-19
Prevent canceling backends that are active but not involved in the distributed deadlock

### LINKED ISSUE #1960: The deadlock detector could kill the wrong backends?
I'm not sure, I haven't run a test to confirm yet, but here's why I think this is possible: https://github.com/citusdata/citus/pull/1529/files#r162503326

onderkalaci:
DESCRIPTION: Prevent canceling backends that are not involved in the distributed deadlock

**Should be backported to 7.0, 7.1 and 7.2**

Fixes #1960 

With this fix, we traverse the graph with DFS which was originally
intended. Note that, before the fix, we traverse the graph with BFS
which might lead to killing some unrelated backend that is not
involved in the distributed deadlock.

We also add one more commit which is not very likely to cause any issues
but would prevent wrong memory accesses.

How to understand the effect of the change:

Generate @lithp 's scenario mentioned in the issue (slightly more complex), follow the steps in the order written as below in the comment starting with `step`:

```
A--B--C--A
 \
  D
```
Exact form is the following:

screen shot 2018-01-19 at 14 40 03

With the following queries (Note that `events_table` DDL can be found in the regression tests):
```SQL
-- simulate transaction (A)
BEGIN;

-- step 0
UPDATE events_table SET event_type = 1 WHERE user_id = 1;

-- step 6
UPDATE events_table SET event_type = 1;

-- simulate transaction (B)
BEGIN;

-- step 3
UPDATE events_table SET event_type = 1 WHERE user_id = 4;

-- step 4
UPDATE events_table SET event_type = 1 WHERE user_id = 3;


-- simulate transaction (C)
BEGIN;
-- step 2
UPDATE events_table SET event_type = 1 WHERE user_id = 3;

-- step 5
UPDATE events_table SET event_type = 1 WHERE user_id = 1;

-- simulate transaction (D)
BEGIN;
-- step 1
UPDATE events_table SET event_type = 1 WHERE user_id = 2;

```

Before the fix, you'd see that the deadlock detection adds `transaction D` to the deadlock path, which is actually not involved in the transaction.

To debug it: 
1. Enable distributed deadlock detection logs: `citus.log_distributed_deadlock_detection` to `ON`
2. Disable Postgres deadlock detection `ALTER SYSTEM SET deadlock_timeout TO '1h'`
3. Disable Citus deadlock detection `ALTER SYSTEM SET citus.distributed_deadlock_detection_factor TO -1`
4. On all four transactions, set the `citus.log_remote_commands` TO `on`
5. After step 6, run ` SELECT check_distributed_deadlocks();` and observe the postmaster logs w/wout the fix. You'll see that without the fix, Citus would add transaction `D` to the deadlock path, which is obviously wrong.

### DIFF
--- a/src/test/regress/specs/isolation_distributed_deadlock_detection.spec
+++ b/src/test/regress/specs/isolation_distributed_deadlock_detection.spec
@@ -97,6 +97,11 @@ step "s1-insert-ref-11"
   INSERT INTO deadlock_detection_reference VALUES (11, 11);
 }
 
+step "s1-update-2-4"
+{
+  UPDATE deadlock_detection_test SET some_val = 1 WHERE user_id = 2 OR user_id = 4;
+}
+
 step "s1-finish"
 {
   COMMIT;
@@ -122,6 +127,11 @@ step "s2-update-2"
 step "s2-update-3"
 {
   UPDATE deadlock_detection_test SET some_val = 2 WHERE user_id = 3;
+}
+
+step "s2-update-4"
+{
+  UPDATE deadlock_detection_test SET some_val = 2 WHERE user_id = 4;
 }
 
 step "s2-upsert-select-all"
@@ -249,6 +259,11 @@ step "s4-update-7"
   UPDATE deadlock_detection_test SET some_val = 4 WHERE user_id = 7;
 }
 
+step "s4-random-adv-lock"
+{
+  SELECT pg_advisory_xact_lock(8765);
+}
+
 step "s4-finish"
 {
   COMMIT;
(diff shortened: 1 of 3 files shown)


src/test/regress/specs/isolation_distributed_deadlock_detection.spec
```diff
@@ -249,6 +259,11 @@ step "s4-update-7"
   UPDATE deadlock_detection_test SET some_val = 4 WHERE user_id = 7;
 }
 
+step "s4-random-adv-lock"
+{
+  SELECT pg_advisory_xact_lock(8765);
```
marcocitus: This could use a comment

lithp:
Wow, I didn't expect for this to be fixed by the time I woke up!

I've confirmed this change fixes both issues, feel free to merge!

One small quibble, the commit message has a typo: "With this fix, we traverse the graph with DFS which was originally inteded." should be "With this fix, we traverse the graph with DFS which was originally _intended_."

PR #1963, opened 2018-01-20
Don't rely on gcc-specific features

lithp:
`#ifdef` away reliance on support for some features which are not supported by Visual Studio.

### DIFF
--- a/src/backend/distributed/worker/worker_partition_protocol.c
+++ b/src/backend/distributed/worker/worker_partition_protocol.c
@@ -616,15 +616,21 @@ CitusCreateDirectory(StringInfo directoryName)
 
 
 #ifdef WIN32
-static bool FileIsLink(char *filename, struct stat filestat)
+static bool
+FileIsLink(char *filename, struct stat filestat)
 {
 	return pgwin32_is_junction(filename);
 }
+
+
 #else
-static bool FileIsLink(char *filename, struct stat filestat)
+static bool
+FileIsLink(char *filename, struct stat filestat)
 {
 	return S_ISLNK(filestat.st_mode);
 }
+
+
 #endif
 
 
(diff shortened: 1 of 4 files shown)


src/backend/distributed/worker/worker_partition_protocol.c
```diff
@@ -614,6 +615,26 @@ CitusCreateDirectory(StringInfo directoryName)
 }
 
 
+#ifdef WIN32
+static bool
+FileIsLink(char *filename, struct stat filestat)
+{
+	DWORD attributes = GetFileAttributes(filename);
+	return attributes & FILE_ATTRIBUTE_REPARSE_POINT;
```
pykello: Why not use `pgwin32_is_junction()` instead?

src/backend/distributed/worker/worker_partition_protocol.c
```diff
@@ -614,6 +615,26 @@ CitusCreateDirectory(StringInfo directoryName)
...
+{
+	return S_ISLNK(filestat.st_mode);
+}
+
+
+#endif
+
+
 /*
  * CitusRemoveDirectory first checks if the given directory exists. If it does, the
```
pykello: I think we should delete this function and use `rmtree()` instead. They seem to do the same thing.

lithp: Added some comments to your PR. I think that's a reasonable aim for a future PR but don't know if it belongs in this one.

src/backend/distributed/worker/worker_partition_protocol.c
lithp: Nice catch. How did you find it?

pykello: I usually grep for functions/constants we use to check if postgres already uses them, and if yes, how :)

src/backend/distributed/worker/worker_partition_protocol.c
pykello: 👍

PR #1965, opened 2018-01-22
Skip JSON validation on coordinator during COPY

marcocitus:
Some Citus users report poor COPY performance when ingesting large JSON objects. The reason is that parsing and validating large JSON objects consumes a lot of CPU time. However, it is not strictly necessary for the coordinator to parse the JSON, it is only useful for validation and returning the line number of line that contains the malformed JSON.

This PR introduces a GUC to skip JSON parsing when copying into a JSONB column that is not the partition column by treating it as text within the parsing functions. For sending the row to the worker in binary copy format, we still need to adhere to the encoding used by `json_send`  and `json_recv` (version number 1 followed by text). To that end I added a citus_text_send_as_jsonb function.

On a Citus Cloud formation of 1+2*r4.xlarge this improved single-session COPY throughput by 3.2x when ingesting 3KB JSON objects with 100 fields (from 14k/s to 45k/s).

Given the substantial benefit, I propose enabling this GUC by default to always have better COPY+JSONB performance. The downside of enabling is that the line number is no longer in the detail when the JSON is malformed, since the validation now happens on the worker, but this seems only relevant for debugging.

### DIFF
--- a/src/backend/distributed/commands/multi_copy.c
+++ b/src/backend/distributed/commands/multi_copy.c
@@ -59,9 +59,11 @@
 #include "catalog/pg_type.h"
 #include "commands/copy.h"
 #include "commands/defrem.h"
+#include "distributed/listutils.h"
 #include "distributed/master_protocol.h"
 #include "distributed/metadata_cache.h"
 #include "distributed/multi_copy.h"
+#include "distributed/multi_executor.h"
 #include "distributed/multi_partitioning_utils.h"
 #include "distributed/multi_physical_planner.h"
 #include "distributed/multi_shard_transaction.h"
@@ -71,6 +73,7 @@
 #include "distributed/shard_pruning.h"
 #include "distributed/version_compat.h"
 #include "executor/executor.h"
+#include "libpq/pqformat.h"
 #include "nodes/makefuncs.h"
 #include "tsearch/ts_locale.h"
 #include "utils/builtins.h"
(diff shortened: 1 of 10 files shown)


src/backend/distributed/commands/multi_copy.c
```diff
@@ -480,6 +551,84 @@ CopyToExistingShards(CopyStmt *copyStatement, char *completionTag)
...
+		{
+			char *columnName = NameStr(currentColumn->attname);
+			ListCell *inputColumnCell = NULL;
+			bool isInputColumn = false;
+
+			foreach(inputColumnCell, inputColumnNameList)
+			{
+				char *inputColumnName = strVal(lfirst(inputColumnCell));
+
+				if (strncmp(inputColumnName, columnName, NAMEDATALEN) == 0)
```
pykello: Maybe use `namestrcmp` instead of `NameStr` + `strncmp`, i.e.:

```C
if (namestrcmp(&currentColumn->attname, inputColumnName) == 0)
```

src/backend/distributed/commands/multi_copy.c
```diff
@@ -369,39 +383,96 @@ CopyToExistingShards(CopyStmt *copyStatement, char *completionTag)
...
+			currentColumn->atttypid = TEXTOID;
+
+			if (copyOutState->binary)
+			{
+				Oid textSendAsJsonbFunctionId = CitusTextSendAsJsonbFunctionId();
+
+				/*
+				 * If we're using binary encoding between coordinator and workers
+				 * then we should honour the format expected by jsonb_recv, which
+				 * is a version number followed by text. We thefere use an output
```
pykello: type, `thefere` -> `therefore`.

src/backend/distributed/commands/multi_copy.c
```diff
@@ -369,39 +383,96 @@ CopyToExistingShards(CopyStmt *copyStatement, char *completionTag)
...
+				 * function which sends the text as if it were jsonb, namely by
+				 * prepending a version number.
+				 */
+				fmgr_info(textSendAsJsonbFunctionId, currentOutputFunction);
```
pykello: I think using `&copyDest->columnOutputFunctions[columnIndex]` directly instead of a variable might help the readability a bit. When I saw this I first expected a use of `currentOutputFunction` after this line, but after a bit of thinking I realized that we are actually modifying the array inside the `copyDest`. If we didn't use a variable, I needed less thinking.

src/backend/distributed/commands/multi_copy.c
jasonmp85: Huh; I didn't know about this function!

metdos:
Shall we also add a description to PR for the changelog?

marcocitus:
> Shall we also add a description to PR for the changelog?

Added. Thanks for the reminder!

PR #1967, opened 2018-01-22
Avoid platform-specific usage of echo in isolation tests

marcocitus:
Currently, the isolation tests use COPY .. FROM PROGRAM 'echo ...' in a way that does not work on some platforms. A command like `echo "1\n2"` will not actually print 2 lines on RHEL-like systems (Fedora, CentOS, ...). echo is a bit weird like that because it is a shell command, not a program.

When running the isolation tests, this causes a lot of unexpected errors like:
```
ERROR: extra data after last expected column
```

```
starting permutation: s1-begin s2-begin s1-master_append_table_to_shard s2-master_append_table_to_shard s1-commit s2-commit
setup failed: ERROR:  invalid input syntax for integer: "0
7
8
9
10000"
CONTEXT:  COPY table_to_append, line 1, column id: "0
7
8
9
10000"
```

I've had this problem for a while on Fedora and still have it after switching to CentOS. @pykello ran into it as well.

I went over all usages of echo in the isolation tests and it seems most of them were just for initialising the table. I've replaced those with multi-row INSERTs. For the remainder, I've replaced usages of `\n` and `\\n` with `:` followed by `| tr : "\n"` and verified that this works on Mac, RHEL-like and Debian-like systems.


src/test/regress/expected/isolation_add_node_vs_reference_table_operations.out
```diff
@@ -285,15 +285,14 @@ step s2-begin:
...
+	COPY test_reference_table FROM PROGRAM 'seq 1 5';
 
 step s1-add-second-worker: 
 	SELECT nodename, nodeport, isactive FROM master_add_node('localhost', 57638);
  <waiting ...>
 step s2-commit: 
 	COMMIT;
- <waiting ...>
-step s2-commit: <... completed>
+
```
metdos: Why is this line changed?

src/test/regress/expected/isolation_add_node_vs_reference_table_operations.out
```diff
@@ -5,7 +5,7 @@ create_reference_table
 
                
 step s2-load-metadata-cache: 
-	COPY test_reference_table FROM PROGRAM 'echo "1\n2\n3\n4\n5"';
+	COPY test_reference_table FROM PROGRAM 'seq 1 5';
```
mtuncer: ```seq``` does not work in windows. We would need to it from ```MinGW```.  Perhaps we can create a script in regression test like ```tmp_check/tmp-bin/psql``` to get this behavior.

src/test/regress/expected/isolation_copy_placement_vs_copy_placement.out
```diff
@@ -2,10 +2,10 @@ Parsed test spec with 2 sessions
 
 starting permutation: s1-load-cache s2-load-cache s2-set-placement-inactive s2-begin s2-repair-placement s1-repair-placement s2-commit
 step s1-load-cache: 
-	COPY test_hash_table FROM PROGRAM 'echo "1,1\n2,2\n3,3\n4,4\n5,5"' WITH CSV;
+	COPY test_hash_table FROM PROGRAM 'echo "1,1:2,2:3,3:4,4:5,5" | tr : "\n"' WITH CSV;
```
mtuncer: platform specific ```echo``` seems like a better option.

marcocitus:
Addressed by #2022

PR #1968, opened 2018-01-23
Small refactor, we were using incompatible types

lithp:
Tiny changes

- `NULL` is not 0, we should give functions which accept int64 0
- `Int32GetDatum` gives us a datum of width 4 but we were telling postgres to compare datums of width 8. This was crashing 32-bit builds of Windows.

### DIFF
--- a/src/backend/distributed/utils/node_metadata.c
+++ b/src/backend/distributed/utils/node_metadata.c
@@ -516,7 +516,7 @@ UpdateNodeLocation(int32 nodeId, char *newNodeName, int32 newNodePort)
 	tupleDescriptor = RelationGetDescr(pgDistNode);
 
 	ScanKeyInit(&scanKey[0], Anum_pg_dist_node_nodeid,
-				BTEqualStrategyNumber, F_INT8EQ, Int32GetDatum(nodeId));
+				BTEqualStrategyNumber, F_INT4EQ, Int32GetDatum(nodeId));
 
 	scanDescriptor = systable_beginscan(pgDistNode, DistNodeNodeIdIndexId(), indexOK,
 										NULL, scanKeyCount, scanKey);
@@ -630,7 +630,7 @@ get_shard_id_for_distribution_column(PG_FUNCTION_ARGS)
 		List *shardIntervalList = LoadShardIntervalList(relationId);
 		if (shardIntervalList == NIL)
 		{
-			PG_RETURN_INT64(NULL);
+			PG_RETURN_INT64(0);
 		}
 
 		shardInterval = (ShardInterval *) linitial(shardIntervalList);
@@ -679,7 +679,7 @@ get_shard_id_for_distribution_column(PG_FUNCTION_ARGS)
 		PG_RETURN_INT64(shardInterval->shardId);
 	}
 
-	PG_RETURN_INT64(NULL);
+	PG_RETURN_INT64(0);
 }
 
 
@@ -1052,7 +1052,7 @@ GetNodeTuple(char *nodeName, int32 nodePort)
 	ScanKeyInit(&scanKey[0], Anum_pg_dist_node_nodename,
 				BTEqualStrategyNumber, F_TEXTEQ, CStringGetTextDatum(nodeName));
 	ScanKeyInit(&scanKey[1], Anum_pg_dist_node_nodeport,
-				BTEqualStrategyNumber, F_INT8EQ, Int32GetDatum(nodePort));
+				BTEqualStrategyNumber, F_INT4EQ, Int32GetDatum(nodePort));
 	scanDescriptor = systable_beginscan(pgDistNode, InvalidOid, indexOK,
 										NULL, scanKeyCount, scanKey);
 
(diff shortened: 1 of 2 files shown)


src/backend/distributed/utils/node_metadata.c
```diff
@@ -630,7 +630,7 @@ get_shard_id_for_distribution_column(PG_FUNCTION_ARGS)
 		List *shardIntervalList = LoadShardIntervalList(relationId);
 		if (shardIntervalList == NIL)
 		{
-			PG_RETURN_INT64(NULL);
+			PG_RETURN_INT64(0);
```
pykello: I think this should be `PG_RETURN_NULL()`. I think `0` is a valid shard Id, so it doesn't mean the function didn't find any shards.

`PG_RETURN_NULL()` sets `fcinfo->isnull` to `true` before returning from function, so postgres knows this function returned null. `PG_RETURN_INT64()` doesn't set that.

src/backend/distributed/utils/node_metadata.c
```diff
@@ -679,7 +679,7 @@ get_shard_id_for_distribution_column(PG_FUNCTION_ARGS)
 		PG_RETURN_INT64(shardInterval->shardId);
 	}
 
-	PG_RETURN_INT64(NULL);
+	PG_RETURN_INT64(0);
```
pykello: Again, `PG_RETURN_NULL()`.

src/backend/distributed/utils/node_metadata.c
pykello: It seems that our shard Ids start from `102008`, but anyway I‌ think returning `NULL` for absence makes more sense than returning 0.

src/backend/distributed/utils/node_metadata.c
```diff
@@ -1052,7 +1052,7 @@ GetNodeTuple(char *nodeName, int32 nodePort)
 	ScanKeyInit(&scanKey[0], Anum_pg_dist_node_nodename,
 				BTEqualStrategyNumber, F_TEXTEQ, CStringGetTextDatum(nodeName));
 	ScanKeyInit(&scanKey[1], Anum_pg_dist_node_nodeport,
-				BTEqualStrategyNumber, F_INT8EQ, Int32GetDatum(nodePort));
+				BTEqualStrategyNumber, F_INT4EQ, Int32GetDatum(nodePort));
```
pykello: We have the exact same problem in `metadata_sync.c` -> `MarkNodeHasMetadata`. Please also fix that.

lithp:
Wow, good catch on `GetNextNodeId()`! I couldn't find your PR, so I opened one of my own: https://github.com/citusdata/citus/pull/1982 feel free to ignore it if you already have a PR!

src/backend/distributed/utils/node_metadata.c
jasonmp85: Changing this function to return `NULL` rather than `0` would be a change in a public API, no?

src/backend/distributed/utils/node_metadata.c
```diff
@@ -516,7 +516,7 @@ UpdateNodeLocation(int32 nodeId, char *newNodeName, int32 newNodePort)
 	tupleDescriptor = RelationGetDescr(pgDistNode);
 
 	ScanKeyInit(&scanKey[0], Anum_pg_dist_node_nodeid,
-				BTEqualStrategyNumber, F_INT8EQ, Int32GetDatum(nodeId));
+				BTEqualStrategyNumber, F_INT4EQ, Int32GetDatum(nodeId));
```
jasonmp85: Yikes; this was a bad/weird oversight.

src/backend/distributed/utils/node_metadata.c
lithp: I'm not sure what the right thing to do here is. @jasonmp85 is technically right, the only kind of right, this has a chance of breaking user code. I think we're doing a v7.3 next and miner versions aren't allowed to change the API?

On the other hand, the chance of breaking user code is pretty small and returning `NULL` is much cleaner.

src/backend/distributed/utils/node_metadata.c
lithp: pushed

lithp:
@pykello I've pushed an update which fixes the other instance of `INT8EQ`, nice catch!

I ended up not changing it to `PG_RETURN_NULL`. That's technically an interface change which is disallowed for minor releases. We could ask for and would probably receive an exception but there's no rush to fix it now, it's mostly a matter of purity and we are kind of rushing to get these changed merged in.

Ready again for you to review!

jasonmp85:
> I ended up not changing it to `PG_RETURN_NULL`. That's technically an interface change which is disallowed for minor releases.

At first I thought this wasn't true, since upgrading from e.g. `7.2` to `7.3` still requires "jumping tracks" in a packaging system now (i.e. you have to uninstall 7.2 and instal 7.3 and run an `ALTER EXTENSION`, unlike updating from `7.2.0` to `7.2.1`). But I reread our version policy document and I think what you said is correct: we should wait to return `NULL` until an `8.0` release. While the _process_ (i.e. commands invoked) of upgrading from `6.2` to `7.0` and `7.0` to `7.1` is the same, the expectations differ (we expect a _major_ upgrade to involve more application changes, such as a function changing its return value).

Is this being tracked in another issue?

lithp:
@jasonmp85: made an issue: https://github.com/citusdata/citus/issues/1990

PR #1971, opened 2018-01-23
Use PostgreSQL's rmtree().

pykello:
PostgreSQL already has a function for removing a directory tree, so we should use that instead of writing our own code.

### DIFF
--- a/src/backend/distributed/worker/worker_partition_protocol.c
+++ b/src/backend/distributed/worker/worker_partition_protocol.c
@@ -38,6 +38,7 @@
 #include "distributed/version_compat.h"
 #include "executor/spi.h"
 #include "mb/pg_wchar.h"
+#include "port.h"
 #include "storage/lmgr.h"
 #include "utils/builtins.h"
 #include "utils/lsyscache.h"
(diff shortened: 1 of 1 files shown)


src/backend/distributed/worker/worker_partition_protocol.c
```diff
@@ -615,87 +616,17 @@ CitusCreateDirectory(StringInfo directoryName)
...
-	}
-
-	/* we now have an empty directory or a regular file, remove it */
-	if (S_ISDIR(fileStat.st_mode))
-	{
-		removed = rmdir(filename->data);
-	}
-	else
-	{
-		removed = unlink(filename->data);
```
lithp: Not sure if this is what causing the tests to fail but this behavior isn't replicated by `rmtree`. If you pass `rmtree` a file instead of a directory it won't unlink the file.

src/backend/distributed/worker/worker_partition_protocol.c
```diff
@@ -615,87 +616,17 @@ CitusCreateDirectory(StringInfo directoryName)
...
-							errmsg("could not stat file \"%s\": %m", filename->data)));
-		}
-	}
-
-	/*
-	 * If this is a directory, iterate over all its contents and for each
-	 * content, recurse into this function. Also, make sure that we do not
-	 * recurse into symbolic links.
-	 */
-	if (S_ISDIR(fileStat.st_mode) && !S_ISLNK(fileStat.st_mode))
```
lithp: `rmtree` also doesn't reproduce this behavior. `rmtree` will follow symbolic links. I wonder why we decided not to follow them.

src/backend/distributed/worker/worker_partition_protocol.c
```diff
@@ -615,87 +616,17 @@ CitusCreateDirectory(StringInfo directoryName)
...
 {
-	struct stat fileStat;
-	int removed = 0;
-
-	int fileStated = stat(filename->data, &fileStat);
-	if (fileStated < 0)
-	{
-		if (errno == ENOENT)
-		{
-			return;  /* if file does not exist, return */
```
lithp: This simply returns, [but `rmtree` throws an error](https://github.com/postgres/postgres/blob/master/src/common/pgfnames.c#L43). This should cause at least one test to fail. And if it doesn't we should add a test which exercises this path!

Issue #1970, opened 2018-01-23
Add continuous integration for windows


lithp:
We're merging in changes which add support for windows. When that support breaks we're going to take a while to notice unless we add continuous integration. Unfortunately, Travis doesn't support it.

Some options:
- [Visual Studio Team Services Continuous Integration](https://www.visualstudio.com/team-services/continuous-integration/)
- [AppVeyor](https://github.com/marketplace/appveyor)

AppVeyor appears to have tighter github integration (it'll set the build status so we can enforce that PRs don't break Windows).

lithp:
AppVeyor looks nice:

- You create an `appveyor.yml` which contains all the configuration.
- mingw and perl and visual studio are already installed on the build runners. We might not need to install much extra.
- We'll need to create a script which clones postgres and moves our source into the contrib directory

PR #1972, opened 2018-01-23
Add appveyor-powered continuous integration for Windows

lithp:
Continuous integration for windows!

For #1970

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,119 @@
 # Citus toplevel Makefile
 
+EXTENSION = citus
+DATA = citus--7.0-15.sql
+
+
+MODULE_big = citus
+
+OBJS = src/backend/distributed/shared_library_init.o \
+     src/backend/distributed/commands/create_distributed_table.o \
+     src/backend/distributed/commands/drop_distributed_table.o \
+     src/backend/distributed/commands/multi_copy.o \
+     src/backend/distributed/commands/transmit.o \
+     src/backend/distributed/connection/connection_management.o \
+     src/backend/distributed/connection/placement_connection.o \
+     src/backend/distributed/connection/remote_commands.o \
+     src/backend/distributed/executor/citus_custom_scan.o \
+     src/backend/distributed/executor/insert_select_executor.o \
+     src/backend/distributed/executor/intermediate_results.o \
+     src/backend/distributed/executor/multi_client_executor.o \
+     src/backend/distributed/executor/multi_executor.o \
+     src/backend/distributed/executor/multi_real_time_executor.o \
+     src/backend/distributed/executor/multi_router_executor.o \
+     src/backend/distributed/executor/multi_server_executor.o \
+     src/backend/distributed/executor/multi_task_tracker_executor.o \
+     src/backend/distributed/executor/multi_utility.o \
+     src/backend/distributed/executor/subplan_execution.o \
+     src/backend/distributed/master/citus_create_restore_point.o \
+     src/backend/distributed/master/master_citus_tools.o \
+     src/backend/distributed/master/master_create_shards.o \
+     src/backend/distributed/master/master_delete_protocol.o \
+     src/backend/distributed/master/master_expire_table_cache.o \
+     src/backend/distributed/master/master_metadata_utility.o \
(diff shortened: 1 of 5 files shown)


lithp:
It works :boom:

jasonmp85:
It's still failing, though?

lithp:
Well, https://github.com/citusdata/citus/pull/1959 hasn't been merged yet :)

Makefile
```diff
@@ -1,5 +1,118 @@
 # Citus toplevel Makefile
 
+EXTENSION = citus
+DATA = citus--7.0-15.sql
+
+
+MODULE_big = citus
+
+OBJS = src/backend/distributed/shared_library_init.o \
```
jasonmp85: Hey so this is crazy but what if we generate the Makefile using a minimal Docker image we run on Windows?

windows/include/citus_version.h
```diff
@@ -0,0 +1,26 @@
+/* Citus full name as a string */
```
jasonmp85: Same for this. Maybe we just generate it using a minimal Linux image?

Makefile
lithp: IInteresting. This goes to something I was thinking about yesterday when mulling over `citus_version.h`, I wonder how portable we want this to be. I mean, I can get appveyor running perfectly by installing `autoconf` and `make` and making small changes to our existing makefiles, but I don't know how much help I should be giving other people who might want to build Citus in the future.

Presumably Docker is in the same category. Should we require anyone building Citus on Windows to have docker?

jasonmp85: I'm not saying to make users install it; maybe we should just check in `.win` versions of these files or something. I'm just saying for the purposes of making a script _we_ can run when we need to to update various files, it might be nice to script it somewhere we're comfortable with tools (i.e. Linux) and just run it against a shared-mount folder when we need to update these things in Windows.

pykello:
Is this in development or waiting for review from someone?

lithp:
@pykello oh, this is waiting on review from you!

lithp:
And also on #2002

Makefile
```diff
@@ -1,5 +1,118 @@
 # Citus toplevel Makefile
```
pykello: Why is this PR changing this?

windows/include/citus_version.h
pykello: Why is this PR changing this?

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -25,7 +25,17 @@ PG_FUNCTION_INFO_V1(citus_server_id);
 #ifdef HAVE_LIBCURL
```
pykello: Not part of this PR

src/backend/distributed/executor/multi_client_executor.c
```diff
@@ -29,9 +29,8 @@
 #include <errno.h>
```
pykello: Not part of this PR

src/backend/distributed/connection/remote_commands.c
```diff
@@ -995,13 +995,19 @@ BuildWaitEventSet(MultiConnection **allConnections, int totalConnectionCount,
 	WaitEventSet *waitEventSet = NULL;
```
pykello: Not part of this PR

src/backend/distributed/connection/connection_management.c
```diff
@@ -10,7 +10,10 @@
```
pykello: Not part of this PR

Makefile
lithp: Windows builds don't use Make, they instead use visual studio. There's a script in the postgres repo, Mkvcbuild.pm, which will parse our Makefile and generate the correct visual studio project file. Here we're getting our Makefile into a form which it understands.

Probably a better alternative would be to switch over to CMake, but that's going to be a large project.

windows/include/citus_version.h
lithp: Similar story. `citus_version.h` is generated by our Makefile. On Windows we don't use Make, so we have to hard-code it. This is very suboptimal but hard to avoid.

PR #1973, opened 2018-01-24
Window function pushdown support

furkansahin:
WIP


marcocitus:
Quick question: Are we also planning to implement non-pushdownable window functions on top of the result set?

As far as I understand a window function is applied after JOIN, WHERE, GROUP BY and HAVING (so those can be pushed down), but before ORDER BY and LIMIT.

furkansahin:
> Quick question: Are we also planning to implement non-pushdownable window functions on top of the result set?

Yes, we do.

I still continue to research and in a couple of days, we will have a meeting with you and @onderkalaci  to discuss the implementation details.

onderkalaci:
> As far as I understand a window function is applied after JOIN, WHERE, GROUP BY and HAVING (so those can be pushed down), but before ORDER BY and LIMIT (so those can no longer be pushed down in case of a non-pushdownable window function)

That seems accurate, posting the related Postgres [docs](https://www.postgresql.org/docs/10/static/queries-table-expressions.html#QUERIES-WINDOW) for reference.

onderkalaci:
@furkansahin Shall we close this PR now since we're implementing it in # 2008?

Issue #1787, opened 2017-11-10
Remove broadcast join related logic


ozgune:
Citus 5.x and earlier had the notion of "broadcast joins". In Citus 6.x and after, we replaced that logic with reference tables (#879). However, we still have logic in the distributed planner and executor for broadcast joins. (edited)

@velioglu and I walked over our code base and decided to remove this logic in three or four steps. I'm including those steps below.

PR-1a
----
1. Remove all data and code segments that execute broadcast join logic (see [comment here](https://github.com/citusdata/citus/issues/1787#issuecomment-353874206) for detailed list). The only exception here is `citus.large_table_shard_count`
2. Fix related regression tests
3. Check that the enterprise and shard rebalancer repos work as expected

PR-1b
----
4. Change `LocalJoin()` in the join order planner. This function will now check if two tables are colocated. If not, we won't execute a local join. We expect this change to impact two large append distributed tables that are joined on the distribution column, or two hash / range distributed tables that are joined on different types
5. Fix related regression tests

PR-2
----
6. Refactor `BroadcastJoin` function into a `ReferenceTableJoin` function. We no longer have to check for number of shards in a table in this function. Remove `candidateShardList` argument from all join evaluation functions
7. In `FixedJoinOrderPlanner`, we emit the following error message: `Cannot perform outer joins with broadcast joins of more than 1 shard`. Check if this error message is accurate. If so, remove the related `if / else if` block.
8. Clean up `TODOs` marked in @velioglu's local branch.

PR-3
----
9. Change `citus.large_table_shard_count` to emit an error message when set
10. Fix related regression tests
8. Update documentation. Check that we have documentation to upgrade from broadcast tables to reference tables (#1075)

PR-4
---
12. Read over task tracker executor design document and/or fix
13. Read over issue #458 and check if it has now been resolved

Once we make these changes, we're also going to resolve the following issues:
* #2027
* #1789
* #1260 
* #527 
* #638 
* #292 (only third item in the list)
* #458 (if we get to PR-4)
* #253 
* #56

jasonmp85:
Our [code coverage for that file](https://codecov.io/gh/citusdata/citus/src/master/src/backend/distributed/executor/multi_real_time_executor.c) still shows these lines being hit during testing…

We do still rely on this logic, right? So the proposal would be to entirely remove broadcast joins in favor of user-initiated reference tables for anything that might need to be on all nodes?

ozgune:
[...]

PR #1974, opened 2018-01-25
Remove broadcast pr2 (on pr1)

velioglu:
This PR applies the changes given in PR-1b step of the #1787


src/backend/distributed/planner/multi_join_order.c
```diff
@@ -1232,13 +1238,26 @@ LocalJoin(JoinOrderNode *currentJoinNode, TableEntry *candidateTable,
...
+	if (!joinOnPartitionColumns)
 	{
-		nextJoinNode = MakeJoinOrderNode(candidateTable, LOCAL_PARTITION_JOIN,
-										 currentPartitionColumn,
-										 currentPartitionMethod);
+		return NULL;
+	}
+
+	/* shard interval lists must have 1-1 matching for local joins */
+	coPartitionedTables = CoPartitionedTables(
```
onderkalaci: Well, the function comment used to be wrong, but, with this change it seems to fit into the comment 😄

src/test/regress/expected/multi_create_table.out
```diff
@@ -52,6 +52,9 @@ HINT:  Consider using hash partitioning.
  
 (1 row)
 
+-- Manually colocate tables lineitem and orders
```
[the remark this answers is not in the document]
velioglu: Has been updated

src/test/regress/expected/multi_limit_clause_approximate.out
```diff
@@ -40,7 +40,7 @@ DEBUG:  push down of limit count: 600
     199283 | 2726988572
     185925 | 2672114100
     196629 | 2622637602
-    157064 | 2614644408
+    149926 | 2606013732
```
onderkalaci: Ops. Why the test output changes?

src/test/regress/expected/multi_limit_clause_approximate.out
```diff
@@ -79,11 +79,11 @@ DEBUG:  push down of limit count: 150
  c_custkey |       c_name       | lineitem_count 
 -----------+--------------------+----------------
         43 | Customer#000000043 |             42
-       370 | Customer#000000370 |             38
-        79 | Customer#000000079 |             37
-       689 | Customer#000000689 |             36
-       472 | Customer#000000472 |             35
-       685 | Customer#000000685 |             35
+       370 | Customer#000000370 |             40
```
onderkalaci: Again, why the test output changes?

src/test/regress/expected/multi_partition_pruning.out
```diff
@@ -8,18 +8,18 @@ SET citus.next_shard_id TO 770000;
 SELECT l_orderkey, l_linenumber, l_shipdate FROM lineitem WHERE l_orderkey = 9030 or l_orderkey = 1;
```
onderkalaci: I'd add an `ORDER BY` here.  Likely to fail.

src/test/regress/output/multi_outer_join.source
```diff
@@ -53,6 +53,10 @@ SELECT master_create_distributed_table('multi_outer_join_third', 't_custkey', 'a
  
 (1 row)
 
+-- Manually colocate tables multi_outer_join_left, multi_outer_join_right and multi_outer_join_third
```
onderkalaci: I think this is ugly. Why not directly call `create_distributed_table()` and get rid of the old APIs?

src/backend/distributed/planner/multi_join_order.c
```diff
@@ -1423,7 +1456,8 @@ CartesianProduct(JoinOrderNode *currentJoinNode, TableEntry *candidateTable,
 {
 	JoinOrderNode *nextJoinNode = MakeJoinOrderNode(candidateTable, CARTESIAN_PRODUCT,
 													currentJoinNode->partitionColumn,
-													currentJoinNode->partitionMethod);
+													currentJoinNode->partitionMethod,
```
onderkalaci: I think we should explicitly pass `NULL` as we did for `DualHashPartition`. What do you think? Because we're losing the `anchor` notion anyway.

src/backend/distributed/planner/multi_join_order.c
```diff
@@ -1221,7 +1225,18 @@ LocalJoin(JoinOrderNode *currentJoinNode, TableEntry *candidateTable,
...
 	char currentPartitionMethod = currentJoinNode->partitionMethod;
+	TableEntry *currentAnchorTable = currentJoinNode->anchorTable;
 	bool joinOnPartitionColumns = false;
+	bool coPartitionedTables = false;
+
+	/*
+	 * If we previously dual-hash re-partitioned the tables for a join, we
+	 * currently don't allow local join.
+	 */
+	if (currentPartitionMethod == REDISTRIBUTE_BY_HASH)
```
onderkalaci: We should probably add `CARTESIAN_PRODUCT` here as well

src/test/regress/expected/multi_limit_clause_approximate.out
onderkalaci: ping? Is that the shard range thing?

src/test/regress/expected/multi_limit_clause_approximate.out
onderkalaci: ping?

src/test/regress/expected/multi_create_table.out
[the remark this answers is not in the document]
onderkalaci: Hmm, we're marking two append partitioned tables as colocated. This is definitely something that we don't support and seems very confusing now. 

Can't we make them `CoPartitioned` while loading the data? Is that because of the different shard ranges? If that's the case, we're doing even more hacky way. 

How hard is it to make those tables `CoPartitioned()` naturally in the sense that they have the same shard ranges and colocated placements?

src/test/regress/expected/multi_explain.out
```diff
@@ -47,8 +47,8 @@ Sort
   ->  HashAggregate
         Group Key: remote_scan.l_quantity
         ->  Custom Scan (Citus Real-Time)
-              Task Count: 8
-              Tasks Shown: One of 8
+              Task Count: 2
```
onderkalaci: What is the reason for fewer task counts? Could you remind me that?

src/test/regress/expected/multi_join_order_additional.out
```diff
@@ -77,66 +67,18 @@ CREATE TABLE customer_hash (
...
-DEBUG:  join prunable for intervals [12036,13473] and [2951,4455]
-DEBUG:  join prunable for intervals [12036,13473] and [4480,5986]
-DEBUG:  join prunable for intervals [12036,13473] and [8997,10560]
-DEBUG:  join prunable for intervals [13473,14947] and [1,1509]
-DEBUG:  join prunable for intervals [13473,14947] and [1509,4964]
-DEBUG:  join prunable for intervals [13473,14947] and [2951,4455]
-DEBUG:  join prunable for intervals [13473,14947] and [4480,5986]
-DEBUG:  join prunable for intervals [13473,14947] and [8997,10560]
-DEBUG:  join prunable for intervals [13473,14947] and [10560,12036]
+LOG:  join order: [ "lineitem" ][ broadcast join "lineitem" ]
```
onderkalaci: How could this even be possible? `join order: [ "lineitem" ][ broadcast join "lineitem" ]`. How many shards `lineitem` have?

src/backend/distributed/planner/multi_join_order.c
```diff
@@ -1232,13 +1247,25 @@ LocalJoin(JoinOrderNode *currentJoinNode, TableEntry *candidateTable,
...
+	if (!joinOnPartitionColumns)
 	{
-		nextJoinNode = MakeJoinOrderNode(candidateTable, LOCAL_PARTITION_JOIN,
-										 currentPartitionColumn,
-										 currentPartitionMethod);
+		return NULL;
 	}
 
+	/* shard interval lists must have 1-1 matching for local joins */
+	coPartitionedTables = CoPartitionedTables(currentAnchorTable->relationId, relationId);
```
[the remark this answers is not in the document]
velioglu: Added

src/test/regress/expected/multi_explain.out
velioglu: Since we changed the shard_max_size

src/test/regress/expected/multi_join_order_additional.out
velioglu: lineitem has 2 shard, so it will be fixed by the following PR.

src/test/regress/expected/multi_limit_clause_approximate.out
velioglu: Since we changed the shard_max_size

src/test/regress/expected/multi_limit_clause_approximate.out
velioglu: Since we changed the shard_max_size

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -2227,10 +2226,13 @@ ErrorIfUnsupportedShardDistribution(Query *query)
 
 
 /*
- * CoPartitionedTables checks if given two distributed tables have 1-to-1 shard
- * partitioning.
+ * CoPartitionedTables first checks for the shard count, if tables don't have
```
onderkalaci: Can't we start with something more generic sentence before givin into the details:
```CoPartitionedTables checks if given two distributed tables have 1-to-1 shard placement matching.```

File: src/include/distributed/multi_physical_planner.h [hunk 13]
```diff
@@ -326,6 +326,7 @@ extern bool TaskListMember(const List *taskList, const Task *task);
 extern List * TaskListDifference(const List *list1, const List *list2);
 extern List * AssignAnchorShardTaskList(List *taskList);
 extern List * FirstReplicaAssignTaskList(List *taskList);
+extern bool CoPartitionedTables(Oid firstRelationId, Oid secondRelationId);
```
onderkalaci: I think this shouldn't be the last function considering the definition/declaration orders

src/test/regress/input/multi_outer_join.source
```diff
@@ -163,14 +163,13 @@ SELECT * FROM master_apply_delete_command('DELETE FROM multi_outer_join_left');
 SELECT * FROM master_apply_delete_command('DELETE FROM multi_outer_join_right');
 
 -- reload shards with 1-1 matching
-\copy multi_outer_join_left FROM '@abs_srcdir@/data/customer-1-15.data' with delimiter '|'
+\copy multi_outer_join_left FROM '@abs_srcdir@/data/customer-11-20.data' with delimiter '|'
```
onderkalaci: What is the reason for this change? Is this really required?

velioglu: To make left and right tables colocated, otherwise we can't use local join.

onderkalaci:
> Maybe send an e-mail to dev - @saicitus - @samay-sharma about the changes

We decided to send the e-mail when all four PRs are merged. So, probably within the next week.

PR #1976, opened 2018-01-25
Prototype: Recursively plan subqueries in WHERE clause and (add infrastructure for non colocated joins)

onderkalaci:
This is another alternative for #1899. Also includes #1900, otherwise it'd be more confusing.

Mostly opened the PR to get some feedback from @marcocitus. The commits are granular (enough?) to follow what I've in mind.

This requires more thinking around interactions with non-colocated joins, set operations, reference tables etc. But, in general, seems to work pretty well so far. Needs some benchmarking due to increase planning times.

Some large commits are mostly code refactoring - moving.

The two main functions are `ShouldRecursivelyPlanNonColocatedJoins()` and `RecursivelyPlanNonColocatedJoins()`.

My proposal for non-colocated joins is that: Find the optimal? `RTE_IDENTITY` that is not in the equivalence class for the subquery inspected and, replace any lower level subqueries that contain the rte_relation. Recursively plan that lower level subquery. Loop until you're done.


onderkalaci:
Closing this since we're already targeting it in #2016

Issue #1799, opened 2017-11-14
Real-Time select queries might get stuck and never finish under high load


onderkalaci:
Steps to re-produce (almost every time it happens):

* 1+3 c4.4xlarge cluster
* Citus 7.1.4 - Postgres 10.0

```SQL
CREATE TYPE complex AS (
    r       double precision,
    i       double precision
);

SELECT run_command_on_workers('
CREATE TYPE complex AS (
    r       double precision,
    i       double precision
);');

CREATE TABLE test_table_1 
	(key int, 
	 occurred_at timestamp DEFAULT now(), 
	 value_1 jsonb, 
	 value_2 text[], 
	 value_3 int4range, 
	 value_4 complex NOT NULL);
SELECT create_distributed_table('test_table_1', 'key');
```

fab pg.set_config:max_wal_size,"'5GB'"
fab pg.set_config:max_connections,1000

Load some data with pgbench:

```SQL
\set aid  random(1, 100000)
\set bid  random(1, 100000)
\set cid  random(1, 100000)
\set did  random(1, 100000)

INSERT INTO test_table_1 (key, value_1, value_2, value_3, value_4) VALUES 
						 (:aid,
						  row_to_json(row(:aid,:bid, row(:cid+10, :did+500,:aid *7), row(:cid-10, :did-500,:aid *71))),
						  ARRAY[:aid::text, :bid::text, :cid::text, :did::text, 'Onder Kalaci', 'CitusData Rocks']::text[],
						  int4range(:aid, :aid + :bid),
						    (:aid, :did)::complex);
```

Then pgbench command:

` pgbench -f insert_only.sql -P 3 -c128 -j512 -t 1000 -n`

Real-time select pgbench file:
```SQL
SELECT * FROM test_table_1 LIMIT 10;
```

pgbench command:
` pgbench -f select.sql  -P 3 -c32 -j32 -t 512 -n`

At some point, some of the clients blocks with the following backtrace

```
#0  0x00007f4aa73bc9d0 in __poll_nocancel () from /lib64/libc.so.6
#1  0x00007f4a9f74dd77 in MultiClientWait (waitInfo=waitInfo@entry=0x263de98) at executor/multi_client_executor.c:822
#2  0x00007f4a9f74f343 in MultiRealTimeExecute (job=job@entry=0x25d2778) at executor/multi_real_time_executor.c:177
[...]

PR #1979, opened 2018-01-26
Handle errors that are discovered during abort

### LINKED ISSUE #1978: NonblockingForgetResults can get stuck in a loop if error is returned in abort
`NonblockingForgetResults` currently assumes that `PQgetResult` eventually returns null.
https://github.com/citusdata/citus/blob/5d7c52ffa6d5bce4d8dcca05d9918e46cafead6e/src/backend/distributed/connection/remote_commands.c#L192

but there are states in which the connection persistently returns an error result:
https://github.com/postgres/postgres/blob/REL_10_STABLE/src/interfaces/libpq/fe-exec.c#L788

ForgetResults and ClearResults seem to have a similar issue.

marcocitus:
Fixes #1978 , though I suspect there might be an issue with ForgetResults and ClearResults as well.

Also did some slight refactoring of the NonblockingForgetResults function, since the naming was poor and it seemed to miss some PQclear cases.

### DIFF
--- a/src/backend/distributed/connection/remote_commands.c
+++ b/src/backend/distributed/connection/remote_commands.c
@@ -58,7 +58,7 @@ IsResponseOK(PGresult *result)
  * ForgetResults clears a connection from pending activity.
  *
  * Note that this might require network IO. If that's not acceptable, use
- * NonblockingForgetResults().
+ * ClearResultsIfReady().
  *
  * ClearResults is variant of this function which can also raise errors.
  */
@@ -93,7 +93,7 @@ ForgetResults(MultiConnection *connection)
  * is marked critical.
  *
  * Note that this might require network IO. If that's not acceptable, use
- * NonblockingForgetResults().
+ * ClearResultsIfReady().
  */
 bool
 ClearResults(MultiConnection *connection, bool raiseErrors)
@@ -133,12 +133,12 @@ ClearResults(MultiConnection *connection, bool raiseErrors)
 
 
 /*
- * NonblockingForgetResults clears a connection from pending activity if doing
+ * ClearResultsIfReady clears a connection from pending activity if doing
  * so does not require network IO. Returns true if successful, false
  * otherwise.
  */
 bool
-NonblockingForgetResults(MultiConnection *connection)
+ClearResultsIfReady(MultiConnection *connection)
 {
 	PGconn *pgConn = connection->pgConn;
 
@@ -152,6 +152,7 @@ NonblockingForgetResults(MultiConnection *connection)
 	while (true)
 	{
 		PGresult *result = NULL;
+		ExecStatusType resultStatus;
 
 		/* just in case there's a lot of results */
 		CHECK_FOR_INTERRUPTS();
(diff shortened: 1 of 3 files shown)


src/backend/distributed/connection/remote_commands.c
```diff
@@ -152,6 +152,7 @@ NonblockingForgetResults(MultiConnection *connection)
 	while (true)
 	{
 		PGresult *result = NULL;
+		ExecStatusType resultStatus;
 
 		/* just in case there's a lot of results */
 		CHECK_FOR_INTERRUPTS();
```
pykello: `ClearResultsIfReady()` is called from `StartRemoteTransactionAbort()`. Doesn't checking for interrupts here might cause the transaction abort be aborted and leave database in a bad state? Maybe we should pass a flag to `ClearResultsIfReady` and check for interrupts only if that is true?

pykello: For example, if query is cancelled here, none of `CurrentCoordinatedTransactionState`, `InProgressTransactions`, or `XactModificationLevel` will be reset. Also some of the clients won't get the `ROLLBACK`, etc. Then the behaviour of next transaction might be unexpected.

marcocitus: Agreed, this would probably lead to PANIC and the motivation for keeping it in the comment doesn't seem very strong.

src/backend/distributed/connection/remote_commands.c
```diff
@@ -182,19 +180,31 @@ NonblockingForgetResults(MultiConnection *connection)
...
+		PQclear(result);
+
+		if (resultStatus == PGRES_COPY_IN || resultStatus == PGRES_COPY_OUT)
 		{
 			/* in copy, can't reliably recover without blocking */
 			return false;
 		}
 
-		if (result == NULL)
+		if (!(resultStatus == PGRES_SINGLE_TUPLE || resultStatus == PGRES_TUPLES_OK ||
```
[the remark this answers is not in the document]
marcocitus: No, we already called PQclear on the result.

marcocitus:
> We see pqParseInput3 in on the top in both backtraces and the line numbers we see in pqParseInput3 is different. Could the infinite loop happen there?

One of the stacktrackes points to line 67, which the start of the function, outside the loop, so it doesn't seem to be in the loop.

> Or maybe just this loop as this line is appearing in both bts? 

Hmz, it's kind of strange that we enter that loop at all. We return from `NonBlockingForgetResults` if the connection is busy after a call to `parseInput` (in `PQisBusy`). Yet somehow we end up in a `while (conn->asyncStatus == PGASYNC_BUSY)` loop in the subsequent call to `PQgetResult`. The only thing that happened in between `PQisBusy` and this loop is another call to `parseInput`, but `pqParseInput3` never sets the `asyncStatus` to `PGASYNC_BUSY`. `PQisBusy` also returns false if `conn` is `NULL`, but clearly that is not the case, since otherwise `PQgetResult` would have returned NULL.

src/backend/distributed/connection/remote_commands.c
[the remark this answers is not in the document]
onderkalaci: Oh, sure

onderkalaci:
>  Yet somehow we end up in a while (conn->asyncStatus == PGASYNC_BUSY) loop in the subsequent call to PQgetResult. The only thing that happened in between PQisBusy and this loop is another call to parseInput, but pqParseInput3 never sets the asyncStatus to PGASYNC_BUSY.

That's true and aligns with my observation as well. So, I cannot find any code-path that could generate such infinite loops with the given stack traces. 

> but pqParseInput3 never sets the asyncStatus to PGASYNC_BUSY.

I double checked that as well. There seems to be no code path in `pqParseInput3` to set the `asyncStatus` to `PGASYNC_BUSY`.

So, a question from one step back: Are we sure that they observed an infinite loop? I guess, otherwise, it'd probably be not possible to get the stack trace from `pqParseInput3 ` at line 67. 

(Also, given that the waiting time for us to kill the backend is a day, I don't suspect that it might be related to #1799 anymore.)

pykello:
@onderkalaci what do you think about backporting removal of ` CHECK_FOR_INTERRUPTS();`? I think if we try hard to cancel aborts, we will see the problem.

onderkalaci:
> I think if we try hard to cancel aborts, we will see the problem.

Actually, we weren't able to find the code-path that could lead to the infinite loop even with `CHECK_FOR_INTERRUPTS()`. It looks like, the code should always be able to get out of the loop. It'd be super-useful if you share the code-path that could lead the issue (if you already found it).

> @onderkalaci what do you think about backporting removal of CHECK_FOR_INTERRUPTS();? 

The reason that we're cautious about backporting is that we're not sure that'd solve the problem in any way. So, again, if you could point out the code-path that could lead this, we should go ahead and merge all the changes?

PR #1980, opened 2018-01-29
Convert broadcast join to reference join

velioglu:
Makes the changes given as PR-2 on issue [1787](https://github.com/citusdata/citus/issues/1787).

After this commit large_table_shard_count wont be used to
check whether broadcast join, which is renamed as reference
join, can be applied. Reference join can only be applied over
reference tables.


onderkalaci:
> After this commit large_table_shard_count wont be used to
check whether broadcast join, which is renamed as reference
join, can be applied. Reference join can only be applied over
reference tables.

Could you please add reference to the original issue?

src/test/regress/expected/multi_large_table_task_assignment.out
```diff
@@ -48,42 +48,13 @@ DEBUG:  assigned task 2 to node localhost:57637
   2985
 (1 row)
 
--- Single range repartition join, along with a join with a small table containing
```
onderkalaci: Why do we remove this test?

src/test/regress/output/multi_outer_join.source
```diff
@@ -76,20 +115,17 @@ SELECT
 	min(t_custkey), max(t_custkey)
 FROM
 	multi_outer_join_third a LEFT JOIN multi_outer_join_right b ON (r_custkey = t_custkey);
-LOG:  join order: [ "multi_outer_join_third" ][ broadcast join "multi_outer_join_right" ]
```
onderkalaci: This is one of the few cases where we regress in the test coverage. Can't we make this test also pass with reference joins?

src/backend/distributed/planner/multi_join_order.c
```diff
@@ -188,60 +176,11 @@ FixedJoinOrderList(FromExpr *fromExpr, List *tableEntryList)
 			joinClauseList = list_concat(joinClauseList, joinWhereClauseList);
 		}
 
-		/* get the sorted list of shards to check broadcast/local join possibility */
-		candidateShardList = LoadShardIntervalList(nextTable->relationId);
-
 		/* find the best join rule type */
 		nextJoinNode = EvaluateJoinRules(joinedTableList, currentJoinNode,
-										 nextTable, candidateShardList,
```
onderkalaci: @mtuncer and @marcocitus we're doing a major code refactoring for join order planning. In summary:
* For broadcast joins, rely only on reference tables (e.g., get rid of `large_table_shard_count`) 
* For local joins, rely on `CoPartitionedTables()`

The implication of those changes in `FixedJoinOrder()` seems substantial. In general, it simplifies the code a lot (as expected) and all changes make sense to me. Feel free to have a look and comment on the changes.

src/test/regress/expected/multi_create_table.out
```diff
@@ -62,7 +62,22 @@ CREATE TABLE customer (
 	c_acctbal decimal(15,2) not null,
 	c_mktsegment char(10) not null,
 	c_comment varchar(117) not null);
-SELECT master_create_distributed_table('customer', 'c_custkey', 'append');
+SELECT create_reference_table('customer');
```
onderkalaci: would you mind adding small comments on why we have both `customer ` and `customer_append` tables? Maybe a single comment at the beginning of the test to explain once?

velioglu:
We checked tpch queries to understand whether we changed the join order logic after the PRs mentioned in [1787](https://github.com/citusdata/citus/issues/1787). 

We both checked whether the selected join logics are remaining same and queries return the same result even if the local joins become a repartition join (it can be tested by changing the number of nodes to 3 in regression tests). 

Only difference between the branch before these PRs and after these PRs are that some broadcast joins (which is renamed as reference join) becomes a local join. These queries were selecting the broadcast, since the number of shards for append tables were smaller than the large_table_shard_count, so that difference is expected.

PR #1981, opened 2018-01-29
Skip call to ActiveReadableNodeList when there are no subplans

marcocitus:
Minor performance optimisation. I noticed the call to `ActiveReadableNodeList` in `ExecuteSubplans` shows up when profiling SELECT lookups in 2% of samples, but we don't need to call it when there are no subplans.


metdos:
Do we need a changelog description for this PR?

PR #1982, opened 2018-01-29
GetNextNodeId() incorrectly called PG_RETURN_DATUM

lithp:
Nice catch! Here's a PR for you.

PR #1983, opened 2018-01-29
Stabilize the output of a multi_router_planner test

lithp:
This test caused a build of https://github.com/citusdata/citus/pull/1982 to fail


lithp:
Going to just roll this into #1982

Issue #1932, opened 2018-01-05
Master plan may cause out of memory issues due to HashAggregate


byucesoy:
For real-time queries, **Postgres predicts that Custom Scan Node will return 0 rows** and creates its plans according to that. This issue becomes problematic when there is a memory heavy aggregation after custom scan node such as HLL union.

Each HLL intermediate data structure uses about 130KB of memory.  Since PostgreSQL thinks Custom Scan Node will return 0 rows, it prefers to use HashAgg. However, if the query returns ~1.5 million HLL objects, HashAgg requires about 200GB of memory.

A solution could be giving Postgres more realistic estimates about number of rows returning from Custom Scan Node, I am not sure if it is not easy thing to do.

```sql
EXPLAIN SELECT hll_cardinality(SUM(column1))::int FROM table_name GROUP BY column2;
                                     QUERY PLAN                                                                         
-------------------------------------------------------------------------------------------------
 HashAggregate  (cost=0.00..0.00 rows=0 width=0)
   Group Key: remote_scan.worker_column_2
   ->  Custom Scan (Citus Real-Time)  (cost=0.00..0.00 rows=0 width=0)
         Task Count: 64
         Tasks Shown: One of 64
         ->  Task
               Node: host=ec2-13-126-13-116.ap-south-1.compute.amazonaws.com port=5432 dbname=citus
               ->  GroupAggregate  (cost=5050.08..5050.18 rows=8 width=40)
                     Group Key: column2
                     ->  Seq Scan on table_name_102000 table_name  (cost=0.00..4364.72 rows=137072 width=25)
(10 rows)

```

mtuncer:
We prepare the plan for postgres in ```multi_master_planner.c``` see function  ```BuildSelectStatement()``` and  ```BuildAggregatePlan()```

We literally hard code the plan type there.  Perhaps we should add a check to see if there is hll aggregate and force aggregation strategy to ```AGG_SORTED``` . We used this one to plan ```select count(distinct)``` queries.

samay-sharma:
I saw a similar issue with another customer. They have a query like below where neither a nor b are the partition column and they have 250M rows in the table.

```
Select a, b, sum(c), count(*) from table group by a,b having count(*) > 1
```

Citus exhausts all the memory on the coordinator node irrespective of what work_mem we set.

byucesoy:
@mtuncer that is one possible solution but that would cause;
- not using hash aggregation even if it can fit in memory (performance penalty)
[...]

PR #1988, opened 2018-01-31
Respect enable_hashagg in the master planner

metdos:
DESCRIPTION: Adds support for respecting enable_hashagg in the master plan

Provides a workaround for #1932

### DIFF
--- a/src/backend/distributed/planner/multi_master_planner.c
+++ b/src/backend/distributed/planner/multi_master_planner.c
@@ -289,7 +289,7 @@ BuildAggregatePlan(Query *masterQuery, Plan *subPlan)
 	if (groupColumnCount > 0)
 	{
 		bool groupingIsHashable = grouping_is_hashable(groupColumnList);
-		bool groupingIsSortable = grouping_is_hashable(groupColumnList);
+		bool groupingIsSortable = grouping_is_sortable(groupColumnList);
 		bool hasDistinctAggregate = HasDistinctAggregate(masterQuery);
 
 		if (!groupingIsHashable && !groupingIsSortable)
--- a/src/backend/distributed/planner/multi_master_planner.c
+++ b/src/backend/distributed/planner/multi_master_planner.c
@@ -21,6 +21,7 @@
 #include "nodes/makefuncs.h"
 #include "nodes/nodeFuncs.h"
 #include "optimizer/clauses.h"
+#include "optimizer/cost.h"
 #include "optimizer/planmain.h"
 #include "optimizer/tlist.h"
 #include "optimizer/var.h"
@@ -303,13 +304,20 @@ BuildAggregatePlan(Query *masterQuery, Plan *subPlan)
 		 * see nodeAgg.c:build_pertrans_for_aggref(). In that case we use
 		 * sorted agg strategy, otherwise we use hash strategy.
 		 */
-		if (!groupingIsHashable || hasDistinctAggregate)
+		if (!enable_hashagg || !groupingIsHashable || hasDistinctAggregate)
 		{
+			char *messageHint = NULL;
+			if (!enable_hashagg && groupingIsHashable)
+			{
+				messageHint = "Consider setting enable_hashagg to on.";
+			}
+
 			if (!groupingIsSortable)
 			{
 				ereport(ERROR, (errmsg("grouped column list must cannot be sorted"),
 								errdetail("Having a distinct aggregate requires "
-										  "grouped column list to be sortable.")));
+										  "grouped column list to be sortable."),
+								messageHint ? errhint("%s", messageHint) : 0));
 			}
 
 			aggregateStrategy = AGG_SORTED;
@@ -418,7 +426,8 @@ BuildDistinctPlan(Query *masterQuery, Plan *subPlan)
 	 */
 	distinctClausesHashable = grouping_is_hashable(distinctClauseList);
 	hasDistinctAggregate = HasDistinctAggregate(masterQuery);
-	if (distinctClausesHashable && !hasDistinctAggregate)
+
+	if (enable_hashagg && distinctClausesHashable && !hasDistinctAggregate)
 	{
 		const long rowEstimate = 10;  /* using the same value as BuildAggregatePlan() */
 		AttrNumber *distinctColumnIdArray = extract_grouping_cols(distinctClauseList,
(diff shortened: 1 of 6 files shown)


metdos:
I see weird PostgreSQL version errors on Travis for PostgreSQL 9.6.

```
 -- print whether we're using version > 9 to make version-specific tests clear
 SHOW server_version \gset
 SELECT substring(:'server_version', '\d+')::int > 9 AS version_above_nine;
  version_above_nine 
 --------------------
- t
+ f
```

Any ideas? Maybe @jasonmp85

metdos:
@saicitus, @samay-sharma is there a way to test if this helps with the memory problem on some hll queries?

marcocitus:
> I see weird PostgreSQL version errors on Travis for PostgreSQL 9.6.

The regression tests isn't aware of versions. It just shows the diff with the minimal amount of changes, so it's comparing the 9.6 output to the 10 expected output, but it means that something about the 9.6 output changed.

src/backend/distributed/planner/multi_master_planner.c
```diff
@@ -303,7 +304,7 @@ BuildAggregatePlan(Query *masterQuery, Plan *subPlan)
 		 * see nodeAgg.c:build_pertrans_for_aggref(). In that case we use
 		 * sorted agg strategy, otherwise we use hash strategy.
 		 */
-		if (!groupingIsHashable || hasDistinctAggregate)
+		if (!enable_hashagg || !groupingIsHashable || hasDistinctAggregate)
```
marcocitus: @lithp / @mtuncer  will using this GUC cause trouble on windows?

metdos:
> It just shows the diff with the minimal amount of changes

Makes sense.

src/backend/distributed/planner/multi_master_planner.c
mtuncer: yes it causes build failure.  I do not think we should commit this change as it is.  
Is there a way to get this value by other means ? (spi etc)

metdos: Hmm, that's interesting. Don't we use similar GUC variables all around the Citus? Do we use only Citus GUC variables?

mtuncer: On second thought,  we can get this checked in. But we have to make sure that enable_hashagg is declared with PGDLLIMPORT. There are other variables declared with this in ```cost.h```

src/backend/distributed/planner/multi_master_planner.c
```diff
@@ -303,7 +304,7 @@ BuildAggregatePlan(Query *masterQuery, Plan *subPlan)
...
 		{
 			if (!groupingIsSortable)
 			{
```
mtuncer: following error message is now incorrect.  
```
if (!enable_hashagg && groupingIsHashable && !hasDistinctAggregate)
``
setting ```enable_hasagg``` may get it to work. 
Perhaps we could provide this as hint.

Comment (metdos) [reply src/backend/distributed/planner/multi_master_planner.c]: But the default for `enable_hasagg` is already on.

Comment (mtuncer) [reply src/backend/distributed/planner/multi_master_planner.c]: I meant hint could be "setting enable_hashagg to true may help ...."

File: src/backend/distributed/planner/multi_master_planner.c [hunk 3]
```diff
@@ -30,6 +30,9 @@
 #include "utils/syscache.h"
 
 
+extern PGDLLIMPORT bool enable_hashagg;
```
mtuncer: I think the right fix is to make this inside postgresql. We should open an issue to track this.

metdos: 1. We can keep this temporarily until PostgreSQL gets it. Does that work?

src/backend/distributed/planner/multi_master_planner.c
```diff
@@ -303,13 +304,15 @@ BuildAggregatePlan(Query *masterQuery, Plan *subPlan)
...
-		if (!groupingIsHashable || hasDistinctAggregate)
+		if (!enable_hashagg || !groupingIsHashable || hasDistinctAggregate)
 		{
 			if (!groupingIsSortable)
 			{
 				ereport(ERROR, (errmsg("grouped column list must cannot be sorted"),
 								errdetail("Having a distinct aggregate requires "
-										  "grouped column list to be sortable.")));
+										  "grouped column list to be sortable."),
+								errhint("Setting enable_hashagg to on may "
```
mtuncer: we should show this hint only if ```enable_hashagg``` is not set

mtuncer: try
```c
enable_hashagg?NULL:errhint(...)
```

metdos:
> using enable_hasagg breaks windows builds. We can live with it at the moment. But we should get it fixed, or provide a work around

Sent a patch to PostgreSQL

https://www.postgresql.org/message-id/CAL1dPcfa45o1dC-c4t-48v0OZE6oy4ChJhObrtkK8mzNfXqDTA%40mail.gmail.com

src/backend/distributed/planner/multi_master_planner.c
```diff
@@ -303,13 +304,21 @@ BuildAggregatePlan(Query *masterQuery, Plan *subPlan)
...
 		 * sorted agg strategy, otherwise we use hash strategy.
 		 */
-		if (!groupingIsHashable || hasDistinctAggregate)
+		if (!enable_hashagg || !groupingIsHashable || hasDistinctAggregate)
 		{
+			char *messageHint = NULL;
+			if (!enable_hashagg && groupingIsHashable)
+			{
+				messageHint = "Setting enable_hashagg to on may help "
+							  "if grouped column list is hashable.";
```
mtuncer: ```groupingIsHashable``` asserts that grouping here is hashable. We do not need to second hint line. Perhaps something like
```
"Consider setting enable_hashagg to on."
```

metdos:
The required change is merged into PostgreSQL master branch.
https://github.com/postgres/postgres/commit/935dee9ad5a8d12f4d3b772a6e6c99d245e5ad44

lithp:
@metdos I think that we should ask for this commit to be backported to at least `REL_10_STABLE`. It's currently in master, which means we won't run on Windows until version postgres 11 is realeased.

jmehnle:
@metdos, @ozgune (since you've handled technical issues with us in the past): Is there any chance this workaround can be backported to Citus 6.2?

PR #1989, opened 2018-01-31
Some code refactoring and performance improvements for restriction equivalences

onderkalaci:
We're planning to expand recursive planning for subqueries in WHERE clause and the approach we're discussing heavily depends on the restriction equivalences (#1976).

This PR does some code refactoring to make that PR smaller and adds one more commit for improving the performance of restriction equivalence checks which we're planning to call per subquery.

For example, the improvement would decrease the join restriction count from  6054 to 51 by eliminating the duplicates for this complex [query](https://gist.github.com/onderkalaci/d99117be17db56dd863105ff1ef8b7a4#file-smart_non_colocated_joins-sql-L211). Similarly, the number of join restrictions decreases from 36 to 9 for this [query](https://github.com/citusdata/citus/blob/198438978ed4f2fee4a73f9eacd11566b0d92eed/src/test/regress/sql/multi_subquery_misc.sql#L78) which involves subqueries in where clause.

I'll do some more benchmarks to see if there is any negative effect on the planning times of pushdownable queries (initial benchmarks show that no-noticeable difference).


src/backend/distributed/planner/relation_restriction_equivalence.c
```diff
@@ -433,20 +428,43 @@ RestrictionEquivalenceForPartitionKeys(PlannerRestrictionContext *
...
+}
+
+
+/*
+ * GenerateAllAttributeEquivalances gets the planner restriction context and returns
+ * the list of all attribute equivalences based on both join restrictions and relation
+ * restrictions.
+ */
+List *
+GenerateAllAttributeEquivalances(PlannerRestrictionContext *plannerRestrictionContext)
```
marcocitus: lances -> lences

src/backend/distributed/planner/relation_restriction_equivalence.c
```diff
@@ -1751,3 +1754,83 @@ QueryRteIdentities(Query *queryTree)
...
+
+		/* obviously we shouldn't treat different join types as being the same */
+		if (joinRestriction->joinType != joinRestrictionInput->joinType)
+		{
+			continue;
+		}
+
+		/*
+		 * If we're dealing with different queries, we shouldn't treat their
+		 * restrictions as bing the same.
```
marcocitus: bing -> being

src/backend/distributed/planner/relation_restriction_equivalence.c
```diff
@@ -1751,3 +1754,83 @@ QueryRteIdentities(Query *queryTree)
...
+
+		/* if we already have the same restrictions, skip */
+		if (JoinRestrictionListExistsInContext(joinRestriction, filteredContext))
+		{
+			continue;
+		}
+
+		filteredContext->joinRestrictionList =
+			lappend(filteredContext->joinRestrictionList,
+					joinRestriction);
```
marcocitus: does this need to be on 3 lines?

src/backend/distributed/planner/relation_restriction_equivalence.c
```diff
@@ -1751,3 +1754,83 @@ QueryRteIdentities(Query *queryTree)
...
+		 * restrictions as bing the same.
+		 */
+		if (joinRestriction->plannerInfo != joinRestrictionInput->plannerInfo)
+		{
+			continue;
+		}
+
+		/* if we already have the same restrictions, skip */
+		if (list_difference(joinRestriction->joinRestrictInfoList,
+							joinRestrictionInput->joinRestrictInfoList) == NIL)
```
marcocitus: I find this piece a bit hard to understand. Why is this a valid way of checking for duplicates? What's being compared here? Could maybe use a comment on why calling list_difference is sufficient.

src/backend/distributed/planner/relation_restriction_equivalence.c
```diff
@@ -1518,3 +1550,294 @@ RelationIdList(Query *query)
...
+		 * If we're dealing with different queries, we shouldn't treat their
+		 * restrictions as being the same.
+		 */
+		if (joinRestriction->plannerInfo != joinRestrictionInput->plannerInfo)
+		{
+			continue;
+		}
+
+		/*
+		 * We check whether the restrictions in joinRestriction is a super set
```
onderkalaci: Is the comment better now?

marcocitus: Yes, though it's not entirely obvious that the lists will contain the same pointers. Is there a more robust way to check for equality?

onderkalaci: Hmm, `list_difference()` calls `equal()` on each element, so we're not comparing the pointers. Is that what you're asking?

marcocitus: > Hmm, list_difference() calls equal() on each element, so we're not comparing the pointers. Is that what you're asking?

We are comparing the pointers because `equal` does not handle `JoinRestriction`. It's not entirely the obvious that this is (and always will be) the case and whether comparing pointers is sufficient.

onderkalaci: We chatted with @marcocitus on this. Actually, we're comparing `RestrictInfo `s here, where the `equal()` function is implemented.

I updated the comment, is that clear?

PR #1993, opened 2018-02-02
Fix count distinct using field select on top level query

mtuncer:
DESCRIPTION : Fixes count distinct bug on column expressions when used with subqueries

We were allowing count distict queries even if they were
not directly on columns if the query is grouped on
distribution column.

When performing these checks we were skipping subqueries
because they also perform this check in a more concise manner.
We relied on oid SUBQUERY_RELATION_ID (10000) to decide if
a given RTE relation id denotes a subquery, however, we also
use SUBQUERY_PUSHDOWN_RELATION_ID (10001) for some subqueries.

We skip both type of subqueries with this change.

### DIFF
--- a/src/backend/distributed/planner/multi_logical_optimizer.c
+++ b/src/backend/distributed/planner/multi_logical_optimizer.c
@@ -2888,6 +2888,8 @@ ErrorIfUnsupportedAggregateDistinct(Aggref *aggregateExpression,
  * AggregateDistinctColumn checks if the given aggregate expression's distinct
  * clause is on a single column. If it is, the function finds and returns that
  * column. Otherwise, the function returns null.
+ * The function expects to find a single column here, no FieldSelect or other
+ * expressions are accepted as a column.
  */
 static Var *
 AggregateDistinctColumn(Aggref *aggregateExpression)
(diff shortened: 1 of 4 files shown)


mtuncer:
note that there are 3 other places where we check ```SUBQUERY_RELATION_ID``` but not ```SUBQUERY_PUSHDOWN_RELATION_ID```.
1 - [MultiLogicalPlanOptimize](https://github.com/citusdata/citus/blob/36f049bdc5d3c1af6c5bc6bdeb4a8ab73351ac4c/src/backend/distributed/planner/multi_logical_optimizer.c#L286)
2 - [GroupedByDisjointPartitionColumn](https://github.com/citusdata/citus/blob/36f049bdc5d3c1af6c5bc6bdeb4a8ab73351ac4c/src/backend/distributed/planner/multi_logical_optimizer.c#L2504)
3 - [ChildNodeList](https://github.com/citusdata/citus/blob/54ccfb24be0684c4ecc506651f6ccc756d058e3a/src/backend/distributed/planner/multi_physical_planner.c#L1713)

onderkalaci:
> note that there are 3 other places where we check SUBQUERY_RELATION_ID but not SUBQUERY_PUSHDOWN_RELATION_ID.

Maybe open an issue with `research` tag? We probably wouldn't want to forget about them?

metdos:
> DESCRIPTION : minor fix on count distinct with subqueries

@mtuncer , could you update description in a similar format to changelog entries?
https://github.com/citusdata/citus/blob/master/CHANGELOG.md

PR #1994, opened 2018-02-02
hash based targetEntry problem solution

furkansahin:
This PR is created just to help the issue #1995 

This is a proposed solution and not yet implemented.
Currently `window_function_pushdown_support` PR uses a walker to change the target entries. Currently,  if a target entry refers to join, we replace it with the conjugate of the column from the joinaliaslist. This creates some problem for subqueries. (the more detailed problem definition can be find in #1995).

I propose that let's not flip the target entries which refers to the result of a join to joinalias but we replace the one which refers to joinalias to the result of a join. Basically what it does is that it iterates over the range table entries and creates a hashmap from the RTE_JOINs as (key: joinalias, value: join_itself). By the help of this hashmap, in our already implemented walker, if we find an expression which is in hashtable, we are going to replace it with the one in value.

### DIFF
--- a/src/backend/distributed/planner/multi_logical_planner.c
+++ b/src/backend/distributed/planner/multi_logical_planner.c
@@ -32,6 +32,7 @@
 #include "nodes/makefuncs.h"
 #include "nodes/nodeFuncs.h"
 #include "nodes/relation.h"
+#include "nodes/print.h"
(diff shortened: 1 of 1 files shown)


marcocitus:
Say what? :)

furkansahin:
@marcocitus you are so quick 😄

PR #1996, opened 2018-02-04
Cache worker node array for faster iteration

marcocitus:
DESCRIPTION: Performance improvements to reduce distributed planning time

We call `LookupNodeForGroup` whenever we look up a shard placement. This function iterates over the worker node hash, which means we spend noticable CPU time in `hash_seq_search` during query planning. This PR changes `LookupNodeForGroup` to use a cached array, which makes it much faster. Because it only stores pointers, is has negligible overhead.

After this change, pgbench throughput increased by ~5% even with just 2 worker nodes.

### DIFF
--- a/src/backend/distributed/utils/metadata_cache.c
+++ b/src/backend/distributed/utils/metadata_cache.c
@@ -2517,7 +2517,7 @@ PrepareWorkerNodeCache(void)
 static void
 InitializeWorkerNodeCache(void)
 {
-	HTAB *oldWorkerNodeHash = NULL;
+	HTAB *newWorkerNodeHash = NULL;
 	List *workerNodeList = NIL;
 	ListCell *workerNodeCell = NULL;
 	HASHCTL info;
@@ -2543,10 +2543,7 @@ InitializeWorkerNodeCache(void)
 	info.match = WorkerNodeCompare;
 	hashFlags = HASH_ELEM | HASH_FUNCTION | HASH_CONTEXT | HASH_COMPARE;
 
-	oldWorkerNodeHash = WorkerNodeHash;
-	WorkerNodeHash = hash_create("Worker Node Hash",
-								 maxTableSize,
-								 &info, hashFlags);
+	newWorkerNodeHash = hash_create("Worker Node Hash", maxTableSize, &info, hashFlags);
 
 	/* read the list from pg_dist_node */
 	workerNodeList = ReadWorkerNodes(includeNodesFromOtherClusters);
@@ -2565,7 +2562,7 @@ InitializeWorkerNodeCache(void)
 
 		/* search for the worker node in the hash, and then insert the values */
 		hashKey = (void *) currentNode;
-		workerNode = (WorkerNode *) hash_search(WorkerNodeHash, hashKey,
+		workerNode = (WorkerNode *) hash_search(newWorkerNodeHash, hashKey,
 												HASH_ENTER, &handleFound);
 
 		/* fill the newly allocated workerNode in the cache */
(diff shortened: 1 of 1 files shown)


src/backend/distributed/utils/metadata_cache.c
```diff
@@ -2463,12 +2465,49 @@ InitializeDistTableCache(void)
...
- * it returns the hash.
+ * GetWorkerNodeArray returns the worker node data as an array of pointers to the
+ * WorkerNode structs allocated in WorkerNode hash. It also sets the integer pointed
+ * to by workerNodeCountResult to the current worker node count.
+ *
+ * The array is returned from the cache, if the cache is not (yet) valid, it is first
+ * rebuilt.
+ */
+static WorkerNode **
+GetWorkerNodeArray(int *workerNodeCountResult)
```
lithp: This feels like unnecessary ceremony. Since this function is only called once, what about just inlining the `PrepareWorkerNodeCache()` call into `LookupNodeForGroup` and using `WorkerNode{Count,Array}` directly?

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -2510,9 +2547,15 @@ InitializeWorkerNodeCache(void)
...
 	long maxTableSize = (long) MaxWorkerNodesTracked;
 	bool includeNodesFromOtherClusters = false;
+	int workerNodeIndex = 0;
 
 	InitializeCaches();
 
+	if (WorkerNodeArray != NULL)
+	{
+		pfree(WorkerNodeArray);
+	}
```
lithp: It makes me nervous to call `pfree` and have such a large gap before `Alloc` is called.

- This should also set `WorkerNodeArray` to `NULL`, just in case this function accidentally calls something which uses `WorkerNodeArray`.
- I can't think of a reason why free-ing early would cause problems but this same function seems to go out of it's way to not destroy `oldWorkerNodeHash` until the very end, is that vestigial?

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -2558,6 +2605,8 @@ InitializeWorkerNodeCache(void)
 		workerNode->nodeRole = currentNode->nodeRole;
 		strlcpy(workerNode->nodeCluster, currentNode->nodeCluster, NAMEDATALEN);
 
+		WorkerNodeArray[workerNodeIndex++] = workerNode;
```
lithp: It makes me nervous to iteratively build a global variable instead of atomically swapping it from one consistent state to another but I suppose we do the same thing with `WorkerNodeHash` so this isn't any worse than the function already way.

src/backend/distributed/utils/metadata_cache.c
marcocitus: Refactored it to do all the re-initialisation at the end.

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -2570,7 +2590,16 @@ InitializeWorkerNodeCache(void)
...
+	hash_destroy(WorkerNodeHash);
+
+	if (WorkerNodeArray != NULL)
+	{
+		pfree(WorkerNodeArray);
+	}
+
+	WorkerNodeCount = newWorkerNodeCount;
+	WorkerNodeArray = newWorkerNodeArray;
+	WorkerNodeHash = newWorkerNodeHash;
```
lithp: If we were in a multi-threaded context this would be dangerous but I suppose the chance of us ever using threads is 0% so... let's ship it!

Issue #1966, opened 2018-01-22
Assert failure in FinishRemoteTransactionAbort in cancellation test


marcocitus:
I'm setting up a CentOS 7.4 VM for development.

When I run `make check` on the `master` branch, I get the following assert failure: 
```
TRAP: FailedAssertion("!(!result)", File: "transaction/remote_transaction.c", Line: 439)
```
Line: https://github.com/citusdata/citus/blob/master/src/backend/distributed/transaction/remote_transaction.c#L439  

It's triggered by a cancellation test in `multi_real_time_transaction.sql`:

```
BEGIN;
SELECT id, pg_advisory_lock(15) FROM test_table;
ERROR:  canceling the transaction since it was involved in a distributed deadlock
ROLLBACK;
BEGIN;
SELECT id, pg_advisory_lock(15) FROM test_table;
ERROR:  canceling the transaction since it was involved in a distributed deadlock
ROLLBACK;
BEGIN;
SELECT id, pg_advisory_lock(15) FROM test_table;
ERROR:  canceling the transaction since it was involved in a distributed deadlock
ROLLBACK;
BEGIN;
SELECT id, pg_advisory_lock(15) FROM test_table;
ERROR:  canceling the transaction since it was involved in a distributed deadlock
ROLLBACK;
BEGIN;
SELECT id, pg_advisory_lock(15) FROM test_table;
ERROR:  canceling the transaction since it was involved in a distributed deadlock
server closed the connection unexpectedly
    This probably means the server terminated abnormally
    before or while processing the request.
connection to server was lost
```

onderkalaci:
I've seen this on travis as well

onderkalaci:
Hit this again on Travis

jasonmp85:
I think I'm still seeing this (maybe in Windows?)

onderkalaci:
I've seen this came-up in `tpcb-like` pgbench on 1+8 c4.4xlarge cluster, with the following command ` pssh  -t 0  -h all_workers.txt -A -i -l ec2-user "pg-latest/bin/pgbench -T 4800 -j32 -c32 -n  -b tpcb-like" `. 

The bt:
```
#0  0x00007f981a747277 in raise () from /lib64/libc.so.6
#1  0x00007f981a748968 in abort () from /lib64/libc.so.6
[...]

PR #1997, opened 2018-02-04
Cache the partition column Var

marcocitus:
We currently cache the partition column string from `pg_dist_partition`, but re-parse it whenever we call `DistPartitionKey` (or `PartitionColumn`), which we do a lot and causes noticable CPU overhead. This PR adds the parsed partition column to the cache. `DistPartitionKey` now returns a copy of the partition column instead of (re-)parsing it.

### DIFF
--- a/src/backend/distributed/planner/multi_join_order.c
+++ b/src/backend/distributed/planner/multi_join_order.c
@@ -1350,8 +1350,6 @@ Var *
 DistPartitionKey(Oid relationId)
 {
 	DistTableCacheEntry *partitionEntry = DistributedTableCacheEntry(relationId);
-	Node *variableNode = NULL;
-	Var *partitionKey = NULL;
 
 	/* reference tables do not have partition column */
 	if (partitionEntry->partitionMethod == DISTRIBUTE_BY_NONE)
@@ -1359,13 +1357,7 @@ DistPartitionKey(Oid relationId)
 		return NULL;
 	}
 
-	/* now obtain partition key and build the var node */
-	variableNode = stringToNode(partitionEntry->partitionKeyString);
-
-	partitionKey = (Var *) variableNode;
-	Assert(IsA(variableNode, Var));
-
-	return partitionKey;
+	return copyObject(partitionEntry->partitionColumn);
 }
 
 
--- a/src/backend/distributed/utils/metadata_cache.c
+++ b/src/backend/distributed/utils/metadata_cache.c
@@ -843,8 +843,19 @@ BuildDistTableCacheEntry(DistTableCacheEntry *cacheEntry)
 	/* note that for reference tables partitionKeyisNull is true */
 	if (!partitionKeyIsNull)
 	{
+		Node *partitionNode = NULL;
+
 		oldContext = MemoryContextSwitchTo(CacheMemoryContext);
+
+		/* get the string representation of the partition column Var */
 		cacheEntry->partitionKeyString = TextDatumGetCString(partitionKeyDatum);
+
+		/* convert the string to a Node and ensure it is a Var */
+		partitionNode = stringToNode(cacheEntry->partitionKeyString);
+		Assert(IsA(partitionNode, Var));
+
+		cacheEntry->partitionColumn = (Var *) partitionNode;
+
 		MemoryContextSwitchTo(oldContext);
 	}
 	else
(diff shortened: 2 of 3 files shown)


src/include/distributed/metadata_cache.h
```diff
@@ -49,6 +49,7 @@ typedef struct
 
 	/* pg_dist_partition metadata for this table */
 	char *partitionKeyString;
+	Var *partitionColumn;
```
onderkalaci: I don't want to complicate things, but, do we ever need `partitionKeyString` after this change?

mtuncer:
This PR has been silent for a while. @marcocitus do we plan to merge this ?

src/include/distributed/metadata_cache.h
marcocitus: Yes, it's still used in about 6 places.

marcocitus:
Just rebased it. It's waiting for a reviewer.

metdos:
> Just rebased it. It's waiting for a reviewer.

@jasonmp85 @pykello could you take a look if have time? Thanks!

metdos:
> Just rebased it. It's waiting for a reviewer.

@marcocitus it seems that 9.6 tests are failed.

marcocitus:
> @marcocitus it seems that 9.6 tests are failed.

That's good old https://github.com/citusdata/citus/issues/1966

jasonmp85:
I'll review.

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -843,8 +843,17 @@ BuildDistTableCacheEntry(DistTableCacheEntry *cacheEntry)
...
 	{
+		Var *partitionColumn = NULL;
+
 		oldContext = MemoryContextSwitchTo(CacheMemoryContext);
+
+		/* get the string representation of the partition column Var */
 		cacheEntry->partitionKeyString = TextDatumGetCString(partitionKeyDatum);
+
+		/* convert the string to a generic Var */
+		partitionColumn = (Var *) stringToNode(cacheEntry->partitionKeyString);
```
jasonmp85: I think it's best to have the `Assert` when building the entry and remove all the `Assert`s during access.

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -886,12 +895,12 @@ BuildDistTableCacheEntry(DistTableCacheEntry *cacheEntry)
...
-		Node *partitionNode = stringToNode(cacheEntry->partitionKeyString);
-		Var *partitionColumn = (Var *) partitionNode;
 		FmgrInfo *hashFunction = NULL;
 
-		Assert(IsA(partitionNode, Var));
-		typeEntry = lookup_type_cache(partitionColumn->vartype,
+		/* partition column should have been set above */
+		Assert(IsA(cacheEntry->partitionColumn, Var));
+
+		typeEntry = lookup_type_cache(cacheEntry->partitionColumn->vartype,
```
jasonmp85: I'm not a huge fan of the double-dereference here. I'd stick with assigning to a intermediate `Var *` variable.

src/backend/distributed/planner/multi_join_order.c
```diff
@@ -1359,11 +1358,8 @@ DistPartitionKey(Oid relationId)
 		return NULL;
 	}
 
-	/* now obtain partition key and build the var node */
-	variableNode = stringToNode(partitionEntry->partitionKeyString);
-
-	partitionKey = (Var *) variableNode;
-	Assert(IsA(variableNode, Var));
+	partitionKey = copyObject(partitionEntry->partitionColumn);
```
jasonmp85: When `__typeof__` is supported, this will actually expand to `(Var *) copyObjectImpl(partitionEntry->partitionColumn)`. I think the `Assert` is kind of redundant here.

jasonmp85: (I suppose this doesn't actually ensure the thing it points to is a `Var` in the PostgreSQL sense, but looking at the impl it's still pretty redundant. I think we should `Assert` on what `stringToNode` returns and then assume this is a `Var` from then on out).

jasonmp85:
Rewriting this commit with my own feedback (😈) and merging.

jasonmp85:
OK, moved the `Assert` to when the `cacheEntry` is being built. Verified that the only place these fields have write accesses is in that function. Now when it's guaranteed to be set (i.e. when partition method is not `NONE`), then callers assume it is non-null and a valid `Var *`.

PR #2002, opened 2018-02-06
Stub uname where it does not exist

lithp:
Windows doesn't have uname

(Ignore the red check marks, that's appveyor and not relevant, the tests are passing)

### DIFF
--- a/src/backend/distributed/utils/statistics_collection.c
+++ b/src/backend/distributed/utils/statistics_collection.c
@@ -25,7 +25,17 @@ PG_FUNCTION_INFO_V1(citus_server_id);
 #ifdef HAVE_LIBCURL
 
 #include <curl/curl.h>
+#ifndef WIN32
 #include <sys/utsname.h>
+#else
+typedef struct utsname
+{
+	char sysname[65];
+	char release[65];
+	char version[65];
+	char machine[65];
+} utsname;
+#endif
 
 #include "access/xact.h"
 #include "distributed/metadata_cache.h"
@@ -57,6 +67,9 @@ static bool SendHttpPostJsonRequest(const char *url, const char *postFields,
 static bool SendHttpGetJsonRequest(const char *url, long timeoutSeconds,
 								   curl_write_callback responseCallback);
 static bool PerformHttpRequest(CURL *curl);
+#ifdef WIN32
+static int uname(struct utsname *buf);
+#endif
 
 
 /* WarnIfSyncDNS warns if libcurl is compiled with synchronous DNS. */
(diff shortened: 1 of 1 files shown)


lithp:
Looks like travis relies on a link which has fallen over

src/backend/distributed/utils/statistics_collection.c
```diff
@@ -145,20 +149,26 @@ CollectBasicUsageStatistics(void)
...
 	appendStringInfo(fields, ",\"worker_node_count\": %u", workerNodeCount);
+#ifndef WIN32
 	appendStringInfoString(fields, ",\"os_name\": ");
 	escape_json(fields, unameData.sysname);
 	appendStringInfoString(fields, ",\"os_release\": ");
 	escape_json(fields, unameData.release);
 	appendStringInfoString(fields, ",\"hwid\": ");
 	escape_json(fields, unameData.machine);
+#else
+	appendStringInfoString(fields, ",\"os_name\": \"Windows\"");
```
jasonmp85: Take a look at the [Windows System Information Functions](https://msdn.microsoft.com/en-us/library/windows/desktop/ms724953(v=vs.85).aspx). It appears maybe `GetSystemInfo` and `GetProductInfo` can give us what we need pretty easily?

jasonmp85: Nevermind, I just looked into this and it's not clear what the "right way" is, for whatever reason.

jasonmp85: I think we actually need `GetVersionEx` to get the numbers to pass to `GetProductInfo`, maybe…

jasonmp85: OK, so here's how Python does it: https://github.com/python/cpython/blob/aa0735f597b072c0eb00404c4d7df359ddc26755/Python/sysmodule.c#L1050-L1115

This just leaves the architecture. Largely academic, I suppose, but Perl's `POSIX` module has a good implementation (they have an entire `uname` shim for Windows)… https://github.com/Perl/perl5/blob/69374fe705978962b85217f3eb828a93f836fd8d/win32/win32.c#L2109-L2137

jasonmp85: Some suggested registry keys, which I was hesitant to accept as a solution, but Python seems to fall back to it (https://github.com/python/cpython/blob/fe2d5babba5d26de2093b6518316b268488187be/Lib/platform.py#L500-L545 ). It's always hard to read this stuff because you're never sure what the obsolete paths are and which are modern (Python/Perl tend to have fallback after fallback to determine this stuff, as the APIs seem to have changed both in call style and in semantics over the years).

lithp: I wrote a replacement for uname inspired by the perl one, thanks for doing this research!

lithp:
Failing because some extra commits to #2011 broke the build

lithp:
(Ignore the appveyor failure, this is because I changed the merge base, appveyor shouldn't be looking at this PR)

pykello:
Is this waiting for the appveyor PR or some action from me?

lithp:
This is waiting on #2011

PR #2003, opened 2018-02-07
Only copy distributed plan when modifying it

marcocitus:
We currently always make a deep copy of the distributed plan when we extract it from the custom scan in the executor. However, this has noticable performance impact, especially due to the cost of copying the jobQuery. The reason we started copying the plan is that we scribble on it when executing DML statements because we need to a) evaluate functions, b) do shard pruning after evaluating functions, and c) load the task placement list in a way that is safe w.r.t. concurrent rebalances. 

However, we never modify the ditributed plan for SELECTs in the executor, so we don't need to pay the price for copying it. This PR changes the executor to only copy the `DistributedPlan` before it gets modified.

### DIFF
--- a/src/backend/distributed/planner/distributed_planner.c
+++ b/src/backend/distributed/planner/distributed_planner.c
@@ -828,6 +828,9 @@ ResolveExternalParams(Node *inputNode, ParamListInfo boundParams)
 
 /*
  * GetDistributedPlan returns the associated DistributedPlan for a CustomScan.
+ *
+ * Callers should only read from the returned data structure, since it may be
+ * the plan of a prepared statement and may therefore be reused.
  */
 DistributedPlan *
 GetDistributedPlan(CustomScan *customScan)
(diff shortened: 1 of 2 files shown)


src/backend/distributed/planner/distributed_planner.c
```diff
@@ -835,16 +838,9 @@ GetDistributedPlan(CustomScan *customScan)
 	node = (Node *) linitial(customScan->custom_private);
 	Assert(CitusIsA(node, DistributedPlan));
 
-	node = CheckNodeCopyAndSerialization(node);
+	CheckNodeCopyAndSerialization(node);
```
marcocitus: I changed this line to also not return a copied plan in the regression tests. This change helped me catch a bug in my initial implementation.

mtuncer: I am okay with that.

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -835,16 +838,9 @@ GetDistributedPlan(CustomScan *customScan)
...
 
-	/*
-	 * When using prepared statements the same plan gets reused across
-	 * multiple statements and transactions. We make several modifications
-	 * to the DistributedPlan during execution such as assigning task placements
-	 * and evaluating functions and parameters. These changes should not
-	 * persist, so we always work on a copy.
-	 */
-	distributedPlan = (DistributedPlan *) copyObject(node);
+	distributedPlan = (DistributedPlan *) node;
```
mtuncer: We are now putting burden of copying this object to callers. Are we sure we want to do this ?

Function description mentions this to new callers. My worry is that, when somebody makes change in one of the callers to modify the distributed plan we may have problems.

marcocitus: True, but this optimisation is worth getting. What if we name the function something like `GetDistributedPlanPointer` to make it clear that it does not make a copy?

mtuncer: Current name is good as it stands I think. I did not like adding  ```Pointer``` suffix much.

PR #2006, opened 2018-02-12
Add support for modifying CTEs

metdos:
DESCRIPTION: Adds support for modifying CTEs


src/test/regress/expected/with_modifying.out
```diff
@@ -25,52 +25,148 @@ SELECT
...
 )
 SELECT
 	*
 FROM
 	basic_delete;
-ERROR:  data-modifying statements are not supported in the WITH clauses of distributed queries
+ user_id |              time               | value_1 | value_2 | value_3 | value_4 
+---------+---------------------------------+---------+---------+---------+---------
+       6 | Thu Nov 23 00:09:44.19812 2017  |       5 |       2 |       0 |        
+       6 | Thu Nov 23 14:43:18.024104 2017 |       3 |       2 |       5 |
```
marcocitus: These could use some ORDER BY's

metdos: Good catch, I added some, and I will also check others. I will also do a bit more testing.

metdos:
Hey @marcocitus, I added [some tests](https://github.com/citusdata/citus/pull/2006/commits/f9930a225412bfa93e7d11ce3c49b00416420fb7) and will continue to add more. Any specific tests you would like to see in regressions?

marcocitus:
Can we use the results of one modifying CTE in a DELETE, UPDATE, or INSERT...SELECT in another CTE?

metdos:
> Can we use the results of one modifying CTE in a DELETE/UPDATE (single shard with subquery), or INSERT...SELECT in another CTE?

Not sure If I understand you completely;

These are not supported;

https://github.com/citusdata/citus/blob/c19c3ef4a1ee22cd09ed7446df02f3e0105f58b7/src/test/regress/expected/with_modifying.out#L94

OR

```sql
WITH added_data AS (
	INSERT INTO modify_table VALUES (1,1), (2, 2), (3,3) RETURNING *
),
raw_data AS (
	DELETE FROM modify_table WHERE id = (SELECT min(id) FROM added_data) RETURNING *
)
INSERT INTO summary_table SELECT id, COUNT(*) AS counter FROM raw_data GROUP BY id;
ERROR:  subqueries are not supported in modifications across multiple shards
DETAIL:  Consider using an equality filter on partition column "id" to target a single shard.
```

However, this is supported;
```sql
WITH added_data AS (
	INSERT INTO modify_table VALUES (1,1), (2, 2), (3,3) RETURNING *
),
raw_data AS (
	DELETE FROM modify_table WHERE id = 1 AND val >= (SELECT min(val) FROM added_data) RETURNING *
)
INSERT INTO summary_table SELECT id, COUNT(*) AS counter FROM raw_data GROUP BY id;
```

marcocitus:
Can you add those as tests?

marcocitus:
One thing I've been wondering about is whether we should do something about using ctid in UPDATE/DELETE with a subquery, since it is kind of dangerous if you're deleting a ctid that may come from a different node.

metdos:
> One thing I've been wondering about is whether we should do something about using ctid in UPDATE/DELETE with a subquery, since it is kind of dangerous if you're deleting a ctid that may come from a different node.

What do you mean by this? As I expected PostgreSQL does not allow to update ctid;
```sql
update articles_hash set ctid = (1,1) where id = 1;
ERROR:  cannot assign to system column "ctid"
LINE 1: update articles_hash set ctid = (1,1) where id = 1;
```

marcocitus:
`DELETE FROM table WHERE ctid IN (SELECT ctid FROM table ...)` is a pretty common pattern in Postgres.

metdos:
> DELETE FROM table WHERE ctid IN (SELECT ctid FROM table ...) is a pretty common pattern in Postgres, but it becomes dangerous if the subquery is a recursively planned CTE.

i. It seems we are good for this query - I will add a regression test.
```sql
DELETE FROM articles_hash WHERE ctid IN (SELECT ctid FROM articles_hash WHERE author_id = 1);
ERROR:  subqueries are not supported in modifications across multiple shards
DETAIL:  Consider using an equality filter on partition column "author_id" to target a single shard.
```

ii. Can you think of other scenarios?

iii. Shall we error out on ctid for any case?

iv. Why even people use such queries :)

marcocitus:
Interesting, seems we actually get a weird error.

```
INSERT INTO test VALUES (2,1);

WITH c AS (
	SELECT ctid FROM test WHERE x = 2
),
d AS (
	DELETE FROM test WHERE x= 1 AND ctid IN (SELECT * FROM c ) RETURNING *
)
SELECT * FROM d;
ERROR:  unrecognized paramkind: 2
```

src/test/regress/expected/multi_router_planner.out
```diff
@@ -428,13 +428,55 @@ WITH RECURSIVE hierarchy as (
 				ce.company_id = 2))
 SELECT * FROM hierarchy WHERE LEVEL <= 2;
 ERROR:  recursive CTEs are not supported in distributed queries
--- CTE with queries other than SELECT is not supported
+-- Test router modifying CTEs
```
marcocitus: We actually still have the `ErrorIfQueryHasModifyingCTE` function in the router planner. Is this something we could get rid of such that, when possible, we'd fully push these queries down instead of doing recursive planning?

marcocitus:
although I also get that error without ctid
```
WITH c AS (                       
SELECT y FROM test WHERE x = 1
),
d AS (
DELETE FROM test WHERE x= 1 AND y IN (SELECT * FROM c ) RETURNING x
)
SELECT * FROM d;
ERROR:  unrecognized paramkind: 2
```

src/test/regress/expected/multi_router_planner.out
marcocitus: (we might have to make an exception for INSERTs)

metdos:
> WITH c AS (                       
> SELECT y FROM test WHERE x = 1
> ),
> d AS (
> DELETE FROM test WHERE x= 1 AND y IN (SELECT * FROM c ) RETURNING x
> )
> SELECT * FROM d;
> ERROR:  unrecognized paramkind: 2

Hey @marcocitus, with some help from @onderkalaci we figured out that this happens because we try to evaluate queries with `read_intermediate_result` in `ExecuteMasterEvaluableFunctions()`. 

Therefore I wanted to NOT set `requiresMasterEvaluation` on `read_intermediate_result` and I ended up replicating `contain_mutable_functions_walker` with this commit: https://github.com/citusdata/citus/pull/2006/commits/87082268a27e727979a41dfe317c55a5d0435cad. What do you think about this approach? If it is also good for you, I will clean up the code (I'm looking for pushdown delete/update queries separately).

src/test/regress/expected/multi_router_planner.out
metdos: I think we can do this, but it requires more stuff than just removing some error checks. I prefer this to be a separate project. What do you think?

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -509,6 +509,8 @@ ModifyQuerySupported(Query *queryTree, bool multiShardQuery)
 	Node *onConflictWhere = NULL;
 
 	CmdType commandType = queryTree->commandType;
+	bool hasRecursivePlanning =
+		ContainsReadIntermediateResultFunction((Node *) queryTree);
```
marcocitus: I think a query can contain an intermediate result AND another distributed table, but we don't do appropriate checks for the latter. Should this be `FindNodeCheck((Node *) queryTree, IsDistributedTableRTE)`?

metdos: Should we error out for that case? And what type of a query are you thinking? I think we error out with having joins objection such as;

```sql
WITH select_data AS (
	SELECT * FROM modify_table
),
raw_data AS (
	UPDATE modify_table SET val = 0 WHERE 
		id IN (SELECT id FROM select_data) AND 
		val IN (SELECT counter FROM summary_table)
	RETURNING id, val
)
SELECT * FROM raw_data ORDER BY val;
```

```sql
DETAIL:  Joins are not supported in distributed modifications.
```

marcocitus: Oh, that's good. I was thinking this is where we prevent joins. Do we need this check at all then? 

In other parts of the planner we treat subqueries like `(SELECT * FROM generate_series(...))` in the same way as `(SELECT * FROM read_intermediate_result(...))` and other subqueries that don't contain distributed tables. Could we do the same here? I.e. any subquery that doesn't contain distributed tables is acceptable.

metdos: Do you mean something like this?
https://github.com/citusdata/citus/commit/cca756f81b9528849127530c3bea584699660f46#diff-4c4f0976727d06684376ecb22e8f5362

However, we still need to update `contain_volatile_functions ` calls;
https://github.com/citusdata/citus/commit/cca756f81b9528849127530c3bea584699660f46#diff-4c4f0976727d06684376ecb22e8f5362R721

marcocitus: > Do you mean something like this?

I think we can actually just remove the whole `if (queryTree->hasSubLinks == true)` section, since we already check for joins later in that function. The error message changes a bit, but the error message is no longer accurate anyway.

> However, we still need to update contain_volatile_functions calls;

Yes, I think we need `contain_volatile_functions_citus` independently of this change.

marcocitus: Example of what I mean:
https://github.com/citusdata/citus/commit/267c8038cb9e0b995e03e399ccfaa0bed933f4be

This allows us to run queries like:
```sql
WITH sel AS (
  SELECT DISTINCT y FROM test
), 
up AS (
  UPDATE test SET y = 1 WHERE x IN (SELECT * FROM sel) RETURNING *
)
SELECT count(*) FROM up;
````

as well as:
```sql
UPDATE test SET y = 1 WHERE x IN (SELECT 2*s FROM generate_series(1,10) s);
```

metdos: > Example of what I mean:
> 267c803

Removing all subquery checks, we miss the upsert case. I will look into query tree after the lunch. Here is the example query:
https://github.com/citusdata/citus/pull/2006/commits/2673edca3adc94cd3215bfb572d268e07cdc1b9f

metdos:
@marcocitus, I presume we don't want to run this query;

```sql

WITH select_rows AS (
	SELECT ctid FROM modify_table WHERE id = 1
),
deleted_rows AS (
	DELETE FROM modify_table WHERE ctid IN (SELECT ctid FROM select_rows) RETURNING *
)
SELECT * FROM deleted_rows;
```

https://github.com/citusdata/citus/pull/2006/commits/7bae9625273918dc51c87cca7c526f9c0b6c95e5#diff-acbd83aee5c8f2cf60953d2666f3963fR567

marcocitus:
Preferably not, since deleting the wrong data is not fun.

src/backend/distributed/planner/multi_router_planner.c
marcocitus: Ah, makes sense, these error checks are a bit messy.

I guess you could just leave the hasSublinks checks and remove the `|| multiShardQuery` bit.

src/test/regress/expected/multi_router_planner.out
marcocitus: Makes sense.

src/backend/distributed/planner/multi_router_planner.c
marcocitus: Semi-related, the current error message is pretty weird:
```sql
INSERT INTO test VALUES (1,2) ON CONFLICT (x) DO UPDATE SET y = (SELECT count(*) FROM test);
ERROR:  subqueries are not supported in modifications across multiple shards
DETAIL:  Consider using an equality filter on partition column "x" to target a single shard.
```

metdos: > I guess you could just leave the hasSublinks checks and remove the || multiShardQuery bit.

Note that, this will make `generate_series` error out.

metdos: > Semi-related, the current error message is pretty weird:

Any suggestions which would work for all :)

marcocitus: > Note that, this will make generate_series error out.

That's because of `hasRecursive`. You don't need `hasRecursive`.

marcocitus: > Any suggestions which would work for all :)

I guess we could say: "subqueries in INSERT commands are not supported", not sure if this creates confusion w.r.t. INSERT...SELECT.

metdos:
> Preferably not, since deleting the wrong data is not fun.

This will be tricky. I attempted to error out on `var->varattno < InvalidAttrNumber` but even simple queries start to fail by claiming that varattno is -1. Such as (I added warning for debugging);

```sql
DELETE FROM raw_events_second WHERE user_id = 2;
WARNING:  var->varattno:-1
DETAIL:  Distributed modifications including system columns such as ctid are not supported.
``` 

Code changes: https://github.com/citusdata/citus/commit/383570296a767c492e064f9b48149735a85063f8 

Perhaps, I need a better walker who cares about resjunk or something else which I miss currently. Any ideas?

src/backend/distributed/planner/multi_router_planner.c
metdos: > That's because of hasRecursive. You don't need hasRecursive.

I think I do, If I remove it, we will start to error out on subqueries which include CTEs. Such as;
https://github.com/citusdata/citus/commit/534a70c7985fce154dc067d39d349bec475c6fad#diff-acbd83aee5c8f2cf60953d2666f3963fR383

metdos: > I guess we could say: "subqueries in INSERT commands are not supported", not sure if this creates confusion w.r.t. INSERT...SELECT.

But, we still error out on some UPDATE/DELETE queries.

marcocitus: > I think I do, If I remove it, we will start to error out on subqueries which include CTEs. Such as;
534a70c#diff-acbd83aee5c8f2cf60953d2666f3963fR383

I don't follow. Where do you error out? 

See https://github.com/citusdata/citus/commit/7066cd720c5537b465fc2e3d762378b2904e1ab8

metdos: I forgot to remove multishard part, now it seems good.

marcocitus:
I think you can just check for `vartype== TIDOID` using FindNodeCheck.

marcocitus:
It's probably best to only apply that check in the multi-shard DELETE/UPDATE case, since it is safe in single-shard DELETE/UPDATE and we might have users relying on that.

metdos:
> I think you can just check for vartype== TIDOID using FindNodeCheck.

We would miss other system columns like xmin. What do you think?
https://github.com/citusdata/citus/pull/2006/commits/f93d2fa6b04886937f3ad527ec4d3c7b51dbb8f5#diff-acbd83aee5c8f2cf60953d2666f3963fR576

metdos:
> It's probably best to only apply that check in the multi-shard DELETE/UPDATE case, since it is safe in single-shard DELETE/UPDATE and we might have users relying on that.

Sound good!

marcocitus:
> We would miss other system columns like xmin. What do you think?

That's fine, I think. No one deletes by xmin, whereas deleting by ctid is standard practice.

metdos:
> We would miss other system columns like xmin. What do you think?
> 
> That's fine, I think. No one deletes by xmin, whereas deleting by ctid is standard practice.



> We would miss other system columns like xmin. What do you think?
> 
> That's fine, I think. No one deletes by xmin, whereas deleting by ctid is standard practice.

Hmm, should I update the current implementation of your proposal then?

marcocitus:
> Hmm, should I update the current implementation to your proposal then?

So what I'm saying is: no one deletes by xmin, so we can allow it without risk, everyone deletes by ctid, so we shouldn't allow it.

You can implement an `IsTIDColumn` function and then use `FindNodeCheck` to check for ctid usage.

src/backend/distributed/utils/citus_clauses.c
```diff
@@ -39,6 +41,146 @@ static Node * PartiallyEvaluateExpressionMutator(Node *expression,
...
+
+	/* Recurse to check arguments */
+	if (IsA(node, Query))
+	{
+		/* Recurse into subselects */
+		return query_tree_walker((Query *) node,
+								 contain_mutable_functions_walker_citus,
+								 context, 0);
+	}
+	return expression_tree_walker(node, contain_mutable_functions_walker_citus,
```
marcocitus: Do we need a new walker or could we use `FindNodeCheck`?

metdos:
> You can implement an IsTIDColumn function and then use FindNodeCheck to check for ctid usage.

One problem with that if `targetEntry->resjunk` is set true, I need to skip the expression and its Var child, but `FindNodeCheck` also traverse children even though `check` function returns false. Am I missing something?

marcocitus:
> One problem with that if targetEntry->resjunk is set true, I need to skip the expression and its Var child

What problem does that solve?

src/backend/distributed/utils/citus_clauses.c
metdos: I don't follow. We want the logic in `contain_mutable_functions` but we don't want our custom functions to be counted for that. We also need the logic `check_functions_in_node` which we don't have in `FindNodeCheck`.

marcocitus: It doesn't make much sense that every function has the same expression_walker boilerplate, even (/especially) if we're copy-pasting code from postgres.

For example, you could have this function:
```
static bool
CitusContainsMutableFunction(Node *node)
{
	/* Check for volatile functions in node itself */
	if (check_functions_in_node(node, contain_volatile_functions_checker_citus,
								context))
	{
		return true;
	}

	if (IsA(node, NextValueExpr))
	{
		/* NextValueExpr is volatile */
		return true;
	}
#endif

	return false;
}
```

and then call `FindNodeCheck(node, CitusContainsMutableFunction)`.

metdos:
> if targetEntry->resjunk is set true, I need to skip the expression and its Var child
> 
> What problem does that solve?

Regular DELETE queries have an expression where resjunk is true and `vartype== TIDOID`.

marcocitus:
> Regular DELETE queries have an expression where resjunk is true and vartype== TIDOID.

Hmz, can you give an example?

```
DELETE FROM test
{QUERY                                                                                                                                        
   :commandType 4 
   :querySource 0 
   :canSetTag true 
   :utilityStmt <> 
   :resultRelation 1 
   :hasAggs false 
   :hasWindowFuncs false 
   :hasTargetSRFs false 
   :hasSubLinks false 
   :hasDistinctOn false 
   :hasRecursive false 
   :hasModifyingCTE false 
   :hasForUpdate false 
   :hasRowSecurity false 
   :cteList <> 
   :rtable (
      {RTE 
      :alias <> 
      :eref 
         {ALIAS 
         :aliasname test 
         :colnames ("x" "y" "version")
         }
      :rtekind 0 
      :relid 66596 
      :relkind r 
      :tablesample <> 
      :lateral false 
      :inh true 
      :inFromCl false 
      :requiredPerms 8 
      :checkAsUser 0 
      :selectedCols (b)
      :insertedCols (b)
      :updatedCols (b)
      :securityQuals <>
      }
   )
   :jointree 
      {FROMEXPR 
      :fromlist (
         {RANGETBLREF 
         :rtindex 1
         }
      )
      :quals <>
      }
   :targetList <> 
   :override 0 
   :onConflict <> 
   :returningList <> 
   :groupClause <> 
   :groupingSets <> 
   :havingQual <> 
   :windowClause <> 
   :distinctClause <> 
   :sortClause <> 
   :limitOffset <> 
   :limitCount <> 
   :rowMarks <> 
   :setOperations <> 
   :constraintDeps <> 
   :stmt_location 0 
   :stmt_len 16
   }
```

marcocitus:
Ah, we're somehow working on the rewritten query here. That's a bit odd.

Anyway, you don't need to check the target list, just the join tree.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -507,9 +562,22 @@ ModifyQuerySupported(Query *queryTree, bool multiShardQuery)
 	List *onConflictSet = NIL;
 	Node *arbiterWhere = NULL;
 	Node *onConflictWhere = NULL;
-
 	CmdType commandType = queryTree->commandType;
 
+	if (multiShardQuery)
```
marcocitus: I think there was a flaw in my reasoning here. This problem is not specific to multi-shard queries, but to intermediate results which could cause you to get a CTID from some other node / transaction. 

so I think this should be: `ContainsReadIntermediateResultFunction(queryTree)`.

That way, subqueries on single shard with ctid are still allowed, provided they don't also have intermediate results.

metdos:
> Ah, we're somehow working on the rewritten query here. That's a bit odd.
> 
> Anyway, you don't need to check the target list, just the join tree.

Ok, here is another attempt to change it. What do you think about it?
https://github.com/citusdata/citus/pull/2006/commits/7cffd77e2c2e13f01c8917f41451b72e3e357b17

src/backend/distributed/utils/citus_clauses.c
```diff
@@ -67,8 +57,8 @@ contain_volatile_functions_checker_citus(Oid func_id, void *context)
 }
 
 
-static bool
-contain_volatile_functions_walker_citus(Node *node, void *context)
+bool
+CitusContainsVolatileFunction(Node *node)
 {
 	if (node == NULL)
```
marcocitus: you can remove the `if (node == NULL)` as well

src/backend/distributed/utils/citus_clauses.c
```diff
@@ -39,6 +41,95 @@ static Node * PartiallyEvaluateExpressionMutator(Node *expression,
...
+	{
+		return false;
+	}
+
+	return (func_volatile(func_id) == PROVOLATILE_VOLATILE);
+}
+
+
+bool
+CitusContainsVolatileFunction(Node *node)
```
marcocitus: `CitusIsVolatileFunction` or `CitusIsVolatileExpression` is probably more appropriate now that the function doesn't do any walking itself.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -488,6 +490,52 @@ ExtractInsertRangeTableEntry(Query *query)
 }
 
 
+static bool
+HasTIDColumn(Node *node)
+{
+	List *queryList = NIL;
+	ListCell *queryCell = NULL;
+
+	ExtractQueryWalker(node, &queryList);
```
marcocitus: Since we only call this function with a Query *, maybe you could change it to `QueryHasTidColumn(Query *query)` and avoid the `ExtractQueryWalker` call.

metdos: I want to reach jointree at every level. How can I do that without `ExtractQueryWalker`?

marcocitus: Why do you want to do that? I thought the CTID that we wanted to ignore is only in the target list at the top level?

src/backend/distributed/utils/citus_clauses.c
```diff
@@ -39,6 +41,85 @@ static Node * PartiallyEvaluateExpressionMutator(Node *expression,
 static Expr * citus_evaluate_expr(Expr *expr, Oid result_type, int32 result_typmod,
 								  Oid result_collation, PlanState *planState);
 
+static bool contain_volatile_functions_checker_citus(Oid func_id, void *context);
+static bool contain_mutable_functions_checker_citus(Oid func_id, void *context);
+
+
+static bool
+contain_volatile_functions_checker_citus(Oid func_id, void *context)
```
marcocitus: Maybe `CitusIsVolatileFunctionIdChecker`

src/backend/distributed/utils/citus_clauses.c
```diff
@@ -39,6 +41,83 @@ static Node * PartiallyEvaluateExpressionMutator(Node *expression,
 static Expr * citus_evaluate_expr(Expr *expr, Oid result_type, int32 result_typmod,
 								  Oid result_collation, PlanState *planState);
 
+static bool CitusIsVolatileFunctionIdChecker(Oid func_id, void *context);
+static bool CitusIsMutableFunctionIdChecker(Oid func_id, void *context);
+
+
+static bool
+CitusIsVolatileFunctionIdChecker(Oid func_id, void *context)
```
marcocitus: These need function comments

src/test/regress/expected/with_modifying.out
```diff
@@ -120,8 +225,485 @@ WITH cte AS (
...
+ERROR:  cannot perform distributed planning for the given modification
+DETAIL:  Distributed modifications including system columns such as ctid on multiple shards are not supported.
+WITH added_data AS (
+	INSERT INTO modify_table VALUES (1,2), (1,6) RETURNING *
+),
+select_data AS (
+	SELECT * FROM added_data WHERE id = 1
+),
+raw_data AS (
+	DELETE FROM modify_table WHERE id = 1 AND ctid IN (SELECT ctid FROM select_data) RETURNING val
```
marcocitus: hmz, we should not allow this but somehow the query is by-passing the check.

Looking at the query tree I see something like:
```
         {SUBPLAN 
         :subLinkType 2 
         :testexpr 
            {OPEXPR 
            :opno 387 
            :opfuncid 1292 
            :opresulttype 16 
            :opretset false 
            :opcollid 0 
            :inputcollid 0 
            :args (
               {VAR 
               :varno 1 
               :varattno -1 
               :vartype 27 
               :vartypmod -1 
               :varcollid 0 
               :varlevelsup 0 
               :varnoold 1 
               :varoattno -1 
               :location 203
```

Basically, the `read_intermediate_result` calls has been stripped from the rewritten query tree. In the multi-shard case we pass the original query to `ModifyQuerySupported`. 

I tried changing to `originalQuery` in `CreateModifyPlan`, but it seems that breaks some other tests, namely:

```sql 
 -- setting the partition column value to itself is allowed
 UPDATE limit_orders SET id = 246 WHERE id = 246;
+ERROR:  modifying the partition value of rows is not allowed
 UPDATE limit_orders SET id = 246 WHERE id = 246 AND symbol = 'GM';
+ERROR:  modifying the partition value of rows is not allowed

UPDATE example SET value = CASE WHEN true THEN now() ELSE now() + interval '1 hour' END WHERE key = 1;
+ERROR:  non-IMMUTABLE functions are not allowed in CASE or COALESCE statements
```

I guess we could maybe take the CTID check out of the `ModifyQuerySupported` function or pass original query as well.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -513,7 +514,7 @@ IsTidColumn(Node *node)
  * features, otherwise it returns an error description.
  */
 DeferredErrorMessage *
-ModifyQuerySupported(Query *queryTree, bool multiShardQuery)
+ModifyQuerySupported(Query *queryTree, Query *originalQuery, bool multiShardQuery)
```
marcocitus: Maybe add a function comment on why we need both.

metdos:
@marcocitus, I'm done with my changes. Let me know if I missed anything or you have new feedback.

Issue #52, opened 2016-02-02
Implement SQL window functions for multi-shard queries that are pushdownable


ozgune:
When the window function's `PARTITION BY` column and the distribution column are the same (or inclusive), we should be able to pushdown the query as-is.

We're already doing it for the window functions that are wrapped in a subquery, such as the following is supported:

```SQL
SELECT
   user_id, time, rnk
FROM
(
  SELECT
    *, rank() OVER my_win as rnk
  FROM
    events_table
    WINDOW my_win AS (PARTITION BY user_id ORDER BY time DESC)
) as foo
ORDER BY
  3 DESC, 1 DESC, 2 DESC
LIMIT
  10;
```
However, the same query where the window function is on the top as follows fail:
```SQL
SELECT *,
       rank() OVER my_win AS rnk
FROM events_table WINDOW my_win AS (PARTITION BY user_id
                                    ORDER BY TIME DESC)
ORDER BY 3 DESC,
         1 DESC,
         2 DESC
LIMIT 10;
ERROR:  could not run distributed query because the window function that is used cannot be pushed down
HINT:  Window functions are supported in two ways. Either add an equality filter on the distributed tables' partition column or use the window functions inside a subquery with a PARTITION BY clause containing the distribution column
```

Note that Citus already supports all window functions for queries that hit single node (i.e., distribution key filter on the query)
```SQL

SELECT row_number() OVER() FROM users_table WHERE user_id = 1;
 row_number 
------------
          1
          2
```

lithp:
This came up in citus-users

lukashes:
I did not see plans of window-functions implementation in citus-users group. This plans are exist?

marcocitus:
Note that window functions are supported for SELECT queries that, based on filter conditions, hit a single shard, co-located shards, and/or reference tables, but not for SELECT queries that hit multiple shards.

ssky1:
Hi,

I am trying to get this query work, but I am getting the same error  

I need to get for each row (id) a unique key distri that's why I didn't use partition by in my windowing 

select tmp.key_distri, tmp.id from (
SELECT ROW_NUMBER() OVER (ORDER BY V.id desc) AS key_distri, 
to_char(V.date,'YYYYMMDD') AS key_filter, 
 V.* FROM V 
 )tmp

this query already work on Postgres (not a distributed system) but don't work on citus, but do work on Teradata too (with is a distributed database like Citus) !

[...]

PR #2008, opened 2018-02-13
Pushdown top level window functions

mtuncer:
DESCRIPTION: Add support for window functions that can be pushed down to worker

We can now support window functions that can be pushed down to workers. Window function must have distribution column in the partition clause to be pushed down.

Fixes : #52

### DIFF
--- a/src/backend/distributed/planner/multi_logical_optimizer.c
+++ b/src/backend/distributed/planner/multi_logical_optimizer.c
@@ -1296,7 +1296,13 @@ MasterExtendedOpNode(MultiExtendedOp *originalOpNode,
 		Expr *newExpression = NULL;
 
 		bool hasAggregates = contain_agg_clause((Node *) originalExpression);
-		if (hasAggregates)
+		bool hasWindowFunction = contain_window_function((Node *) originalExpression);
+
+		/*
+		 * if the aggregate belongs to a window function, it is not mutated, but pushed
+		 * down to worker as it is. Master query should treat that as a Var.
+		 */
+		if (hasAggregates && !hasWindowFunction)
 		{
 			Node *newNode = MasterAggregateMutator((Node *) originalExpression,
 												   walkerContext);
@@ -1858,6 +1864,9 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
 	bool hasNonPartitionColumnDistinctAgg = false;
 	bool repartitionSubquery = false;
 
+	/* only window functions that can be pushed down reach here */
+	bool pushDownWindowFunction = originalOpNode->hasWindowFuncs;
+
 	walkerContext->expressionList = NIL;
 
 	/* find max of sort group ref index */
@@ -1893,12 +1902,18 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
 		List *newExpressionList = NIL;
 		ListCell *newExpressionCell = NULL;
 		bool hasAggregates = contain_agg_clause((Node *) originalExpression);
+		bool hasWindowFunction = contain_window_function((Node *) originalExpression);
 
 		/* reset walker context */
 		walkerContext->expressionList = NIL;
 		walkerContext->createGroupByClause = false;
 
-		if (hasAggregates)
+		/*
+		 * If the expression uses aggregates inside window function contain agg
+		 * clause still returns true. We want to make sure it is not a part of
+		 * window function before we proceed.
+		 */
+		if (hasAggregates && !hasWindowFunction)
 		{
 			WorkerAggregateWalker((Node *) originalExpression, walkerContext);
 
@@ -2013,6 +2028,8 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
 	workerExtendedOpNode = CitusMakeNode(MultiExtendedOp);
 	workerExtendedOpNode->distinctClause = NIL;
 	workerExtendedOpNode->hasDistinctOn = false;
+	workerExtendedOpNode->hasWindowFuncs = originalOpNode->hasWindowFuncs;
+	workerExtendedOpNode->windowClause = originalOpNode->windowClause;
 
 	if (!queryHasAggregates)
 	{
@@ -2042,13 +2059,48 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
 										 newTargetEntryListForSortClauses);
 	}
 
+	if (workerExtendedOpNode->windowClause)
+	{
+		List *windowClauseList = workerExtendedOpNode->windowClause;
+		ListCell *windowClauseCell = NULL;
+
+		foreach(windowClauseCell, windowClauseList)
+		{
+			WindowClause *windowClause = (WindowClause *) lfirst(windowClauseCell);
+
+			List *partitionClauseTargetList =
+				GenerateNewTargetEntriesForSortClauses(originalOpNode->targetList,
+													   windowClause->partitionClause,
+													   &targetProjectionNumber,
+													   &nextSortGroupRefIndex);
+			List *orderClauseTargetList =
+				GenerateNewTargetEntriesForSortClauses(originalOpNode->targetList,
+													   windowClause->orderClause,
+													   &targetProjectionNumber,
+													   &nextSortGroupRefIndex);
+
+			newTargetEntryList = list_concat(newTargetEntryList,
+											 partitionClauseTargetList);
+			newTargetEntryList = list_concat(newTargetEntryList,
+											 orderClauseTargetList);
+		}
+	}
+
 	workerExtendedOpNode->targetList = newTargetEntryList;
 
 	/*
 	 * If grouped by a partition column whose values are shards have disjoint sets
 	 * of partition values, we can push down the having qualifier.
+	 *
+	 * When a query with subquery is provided, we can't determine if
+	 * groupedByDisjointPartitionColumn, therefore we also check if there is a
+	 * window function too. If there is a window function we would know that it
+	 * is safe to push down (i.e. it is partitioned on distribution column, and
+	 * if there is a group by, it contains distribution column).
+	 *
 	 */
-	if (havingQual != NULL && groupedByDisjointPartitionColumn)
+	if (havingQual != NULL &&
+		(groupedByDisjointPartitionColumn || pushDownWindowFunction))
 	{
 		workerExtendedOpNode->havingQual = originalOpNode->havingQual;
 	}
@@ -2067,7 +2119,10 @@ HasNonPartitionColumnDistinctAgg(List *targetEntryList, Node *havingQual,
 								 List *tableNodeList)
 {
 	List *targetVarList = pull_var_clause((Node *) targetEntryList,
-										  PVC_INCLUDE_AGGREGATES);
+										  PVC_INCLUDE_AGGREGATES |
+										  PVC_RECURSE_WINDOWFUNCS);
+
+	/* having clause can't have window functions, no need to recurse for that */
 	List *havingVarList = pull_var_clause((Node *) havingQual, PVC_INCLUDE_AGGREGATES);
 	List *aggregateCheckList = list_concat(targetVarList, havingVarList);
 
@@ -2722,7 +2777,8 @@ ErrorIfContainsUnsupportedAggregate(MultiNode *logicalPlanNode)
 	 * PVC_REJECT_PLACEHOLDERS is implicit if PVC_INCLUDE_PLACEHOLDERS isn't
 	 * specified.
 	 */
-	List *expressionList = pull_var_clause((Node *) targetList, PVC_INCLUDE_AGGREGATES);
+	List *expressionList = pull_var_clause((Node *) targetList, PVC_INCLUDE_AGGREGATES |
+										   PVC_INCLUDE_WINDOWFUNCS);
 
 	ListCell *expressionCell = NULL;
 	foreach(expressionCell, expressionList)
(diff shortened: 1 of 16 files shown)


onderkalaci:
> We can now pushdown top level window functions in subqueries

I think we probably should consider removing `subqueries` from the title/description, which might be slightly confusing. What we're doing is that we're enabling top-level pushdownable window functions, irrespective of subqueries?

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -2017,8 +2019,10 @@ HasNonPartitionColumnDistinctAgg(List *targetEntryList, Node *havingQual,
 								 List *tableNodeList)
 {
 	List *targetVarList = pull_var_clause((Node *) targetEntryList,
-										  PVC_INCLUDE_AGGREGATES);
-	List *havingVarList = pull_var_clause((Node *) havingQual, PVC_INCLUDE_AGGREGATES);
+										  PVC_INCLUDE_AGGREGATES |
+										  PVC_RECURSE_WINDOWFUNCS);
+	List *havingVarList = pull_var_clause((Node *) havingQual, PVC_INCLUDE_AGGREGATES |
```
onderkalaci: Could we have window functions in the having qual?

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -3802,6 +3818,64 @@ SubqueryPushdownMultiNodeTree(Query *queryTree)
 }
 
 
+/*
+ * FlattenJoinVars iterates over provided columnList to identify
```
onderkalaci: I think the comment is pretty clear. 

Maybe one thing to mention is that Postgres handles this during standard planner and we don't use the modified query? (i.e., just before or after the sentence `Postgres is smart about it to detect that ..`)

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -3802,6 +3818,64 @@ SubqueryPushdownMultiNodeTree(Query *queryTree)
...
+ * a join alias. Therefore the same column from a table could be referenced
+ * twice using its absolute table name (t1.a), and without table name (a).
+ * This is a problem when one of them is inside the group by clause and the
+ * other is not. Postgres is smart about it to detect that both target columns
+ * resolve to the same thing, and allows a single group by clause to cover
+ * both target entries. We want to make sure we provide correct varno/varattno
+ * values to Postgres so that it could produce valid query.
+ *
+ * Only exception is that, if a join is given an alias name, we do not want to
+ * flatten those var's.
```
onderkalaci: Maybe very briefly mention that in case joinalias exists, the inner rtes cannot be referenced? (Is that the reason, right?)

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -3802,6 +3818,64 @@ SubqueryPushdownMultiNodeTree(Query *queryTree)
...
+
+		/*
+		 * if join has an alias, it is copied over join rte. There is
+		 * no need to find the JoinExpr to check whether it has
+		 * an alias defined.
+		 */
+		columnRte = rt_fetch(column->varno, rteList);
+		if (columnRte->rtekind == RTE_JOIN && columnRte->alias == NULL)
+		{
+			PlannerInfo *root = makeNode(PlannerInfo);
```
onderkalaci: Shall we create this once?
```C

if (root == NULL)
{
... init..
}

normalizedVar = (Var *) flatten_join_alias_vars(root, (Node *) column);
```

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -3802,6 +3818,64 @@ SubqueryPushdownMultiNodeTree(Query *queryTree)
...
+			Var *normalizedVar = NULL;
+
+			root->parse = (queryTree);
+			root->planner_cxt = CurrentMemoryContext;
+			root->hasJoinRTEs = true;
+
+			normalizedVar = (Var *) flatten_join_alias_vars(root, (Node *) column);
```
onderkalaci: From Postgresql's source code, which we might consider adding here as well:
```
We use the planner's flatten_join_alias_vars
routine to do the flattening; it wants a PlannerInfo root node, which
fortunately can be mostly dummy.
```

onderkalaci:
> I think we probably should consider removing subqueries from the title/description, which might be slightly confusing. What we're doing is that we're enabling top-level pushdownable window functions, irrespective of subqueries?

Ping on this :)

onderkalaci:
> I think this looks almost ready, I'll do some more testing and add an extra commit for adding/modifying the regression tests.

We realized some critical issues with our processing of top level window functions (i.e., apply `GROUP BY` on the coordinator after the window functions applied on the workers). This is wrong and we should fix that.

So, I'm taking my words back, we're not merging this before fixing the issue (and potentially considering other cases as well)

src/backend/distributed/planner/multi_logical_optimizer.c
mtuncer: no

src/backend/distributed/planner/multi_logical_planner.c

src/backend/distributed/planner/multi_logical_planner.c
mtuncer: correct

src/backend/distributed/planner/multi_logical_planner.c
mtuncer: ok

src/backend/distributed/planner/multi_logical_planner.c
mtuncer: ok

src/backend/distributed/planner/multi_logical_optimizer.c
mtuncer: removed

mtuncer:
update on @onderkalaci comment. Now we have resolved the issue, and the PR is ready to be merged upon completion of review.

src/test/regress/expected/multi_subquery_window_functions.out
```diff
@@ -1012,5 +1012,174 @@ FROM (
 ORDER BY 2 DESC, 1 DESC
 LIMIT 5;
 ERROR:  could not run distributed query because the window function that is used cannot be pushed down
-HINT:  Window functions are supported in two ways. Either add an equality filter on the distributed tables' partition column or use the window functions inside a subquery with a PARTITION BY clause containing the distribution column
+HINT:  Window functions are supported in two ways. Either add an equality filter on the distributed tables' partition column or use the window functions with a PARTITION BY clause containing the distribution column
+-- window function at top level query
```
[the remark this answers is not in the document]
mtuncer: fixed

src/test/regress/sql/window_functions.sql
```diff
@@ -0,0 +1,412 @@
...
+	user_id,
+	avg(value_1),
+	RANK() OVER (partition by user_id order by sum(value_2))
+FROM
+	users_table
+GROUP BY user_id, value_2
+ORDER BY user_id, avg(value_1) DESC
+LIMIT 5;
+
+
```
onderkalaci: extra new lines

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -2042,13 +2052,48 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
 										 newTargetEntryListForSortClauses);
 	}
 
+	if (workerExtendedOpNode->windowClause)
```
onderkalaci: In general,  `WorkerExtendedOpNode ()` and `MasterExtendedOpNode ()` functions are becoming really, really long. 

Do you think would it make sense to refactor these functions into smaller ones? Say, can we move most this logic directly into another function (e.g., `List * GenerateNewTargetEntriesForWindowClause(windowClause, ...)` or smth like that)?

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -1997,8 +2000,15 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
 	/*
 	 * If grouped by a partition column whose values are shards have disjoint sets
 	 * of partition values, we can push down the having qualifier.
+	 *
+	 * When a query with subquery is provided, we can't determine if
```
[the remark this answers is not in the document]
mtuncer: rename the variable as ```pushDownWindowFunction```

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -2673,7 +2686,8 @@ ErrorIfContainsUnsupportedAggregate(MultiNode *logicalPlanNode)
 	 * PVC_REJECT_PLACEHOLDERS is implicit if PVC_INCLUDE_PLACEHOLDERS isn't
 	 * specified.
 	 */
-	List *expressionList = pull_var_clause((Node *) targetList, PVC_INCLUDE_AGGREGATES);
+	List *expressionList = pull_var_clause((Node *) targetList, PVC_INCLUDE_AGGREGATES |
```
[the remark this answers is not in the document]
mtuncer: aggregates in the window function are already in the target entry list. window function expressions are only references to those in target entry list. They are marked as resjunk if not explicitly stated in query target.

src/backend/distributed/planner/multi_logical_optimizer.c
mtuncer: we should do that as a part of window function pull. It too will have changes scattered to these two functions.

src/test/regress/expected/window_functions.out
```diff
@@ -0,0 +1,471 @@
...
+---------+-----
+       1 |  13
+       1 |  13
+       1 |  13
+       1 |  13
+       2 |  10
+(5 rows)
+
+-- the same query, but this time join with an alias
+SELECT
```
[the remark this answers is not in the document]
mtuncer: I relied on  you for those queries. fixed.

src/test/regress/expected/window_functions.out
```diff
@@ -0,0 +1,471 @@
...
+       4 | 113
+       2 | 111
+       5 | 109
+       3 | 109
+       4 | 106
+       2 | 106
+(10 rows)
+
+-- the same query with reference tables on the partition by doesn't work
+SELECT
```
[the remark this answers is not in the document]
mtuncer: fixed

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -1292,7 +1292,8 @@ MasterExtendedOpNode(MultiExtendedOp *originalOpNode,
 		Expr *newExpression = NULL;
 
 		bool hasAggregates = contain_agg_clause((Node *) originalExpression);
-		if (hasAggregates)
+		bool hasWindowFunction = contain_window_function((Node *) originalExpression);
```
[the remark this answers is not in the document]
mtuncer: comment added

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -3363,7 +3376,8 @@ pull_var_clause_default(Node *node)
 	 * PVC_REJECT_PLACEHOLDERS is implicit if PVC_INCLUDE_PLACEHOLDERS
 	 * isn't specified.
 	 */
-	List *columnList = pull_var_clause(node, PVC_RECURSE_AGGREGATES);
+	List *columnList = pull_var_clause(node, PVC_RECURSE_AGGREGATES |
```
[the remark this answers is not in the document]
mtuncer: Additional flag is only there to prevent pull_var_clause function to fail.  Any aggregate function (or expression) window function contains already has to be inside target entry list. Therefore  going inside window function is not required.

src/test/regress/multi_schedule
```diff
@@ -31,7 +31,7 @@ test: multi_read_from_secondaries
 # ----------
 test: multi_create_table
 test: multi_create_table_constraints multi_master_protocol multi_load_data multi_behavioral_analytics_create_table
-test: multi_behavioral_analytics_basics multi_behavioral_analytics_single_shard_queries multi_insert_select_non_pushable_queries multi_insert_select multi_insert_select_window multi_shard_update_delete
+test: multi_behavioral_analytics_basics multi_behavioral_analytics_single_shard_queries multi_insert_select_non_pushable_queries multi_insert_select multi_insert_select_window multi_shard_update_delete window_functions
```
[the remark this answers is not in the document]
mtuncer: moved window functions test to line below with 2 more tests from parallel group

src/test/regress/expected/window_functions.out
```diff
@@ -0,0 +1,915 @@
...
+       5 |     3.0000000000000000 |    1
+       6 |     1.3333333333333333 |    5
+       6 |     2.0000000000000000 |    4
+       6 |     4.0000000000000000 |    3
+       6 | 1.00000000000000000000 |    2
+       6 |     2.5000000000000000 |    1
+(32 rows)
+
+-- order by in the window function is same as avg(value_1) DESC
+SELECT
```
onderkalaci: Cool query. But, we should add `LIMIT`  to pushdown `ORDER BY`.  Could you go over the tests to see if we have any other queries like this (e.g., the above is also misses the `LIMIT`)?

src/test/regress/expected/window_functions.out
```diff
@@ -0,0 +1,915 @@
...
+       5 |     2.3333333333333333 |    5
+       5 |     1.6000000000000000 |    6
+       6 |     4.0000000000000000 |    1
+       6 |     2.5000000000000000 |    2
+       6 |     2.0000000000000000 |    3
+       6 |     1.3333333333333333 |    4
+       6 | 1.00000000000000000000 |    5
+(32 rows)
+
+-- limit is not pushed down to worker !!
```
onderkalaci: Ah, you've already noted here @mtuncer

mtuncer:
issue #2034 is opened on distinct on()

PR #2011, opened 2018-02-14
Replace poll() with select() where poll() is not available

lithp:


pykello:
High level question: Why are we not using WaitEventSets?

I think we discussed this earlier, I just wanted to refresh my memory.

src/backend/distributed/connection/connection_management.c
```diff
@@ -10,7 +10,20 @@
 
 #include "postgres.h"
 
+#undef HAVE_POLL
+#undef HAVE_POLL_H
+#undef HAVE_SYS_POLL_H
```
pykello: Are these leftovers from testing?

src/backend/distributed/connection/connection_management.c
```diff
@@ -10,7 +10,20 @@
...
+
+#ifdef HAVE_POLL_H
 #include <poll.h>
+#endif
+#ifdef HAVE_SYS_POLL_H
```
pykello: I don't see any `HAVE_SYS_POLL_H` in `pg_config.h.in`. Why are we doing this?

src/backend/distributed/connection/connection_management.c
```diff
@@ -469,6 +485,31 @@ FinishConnectionEstablishment(MultiConnection *connection)
...
+			if (pollmode == PGRES_POLLING_READING)
+			{
+				FD_SET(selectFileDescriptor, &readFileDescriptorSet);
+			}
+			else if (pollmode == PGRES_POLLING_WRITING)
+			{
+				FD_SET(selectFileDescriptor, &writeFileDescriptorSet);
+			}
+
+			pollResult = (select) (selectFileDescriptor + 1, &readFileDescriptorSet,
```
pykello: why are the parenthesis around `select`?

src/backend/distributed/connection/remote_commands.c
```diff
@@ -994,14 +994,20 @@ BuildWaitEventSet(MultiConnection **allConnections, int totalConnectionCount,
 	int pendingConnectionCount = totalConnectionCount - pendingConnectionsStartIndex;
 	WaitEventSet *waitEventSet = NULL;
 	int connectionIndex = 0;
+	const int MAX_WAIT_EVENT_COUNT = 50;
```
pykello: This should probably be a function of `FD_SETSIZE` instead of constant 50 everywhere.

lithp: Oh yeah, good point

lithp:
@pykello we're not using `WaitEventSet`s because we ran into multiple different issues with them, @mtuncer would be better able to tell you what he was seeing before we gave up and moved to `select()`, but once we switched to `select()` all of our problems went away.

src/backend/distributed/connection/connection_management.c
lithp: Hmm, probably @mtuncer was trying to make this section match [the one here](https://github.com/citusdata/citus/pull/2011/files#diff-b95ac9c2bf85402fbb5eca5edcd94302R35), but that section looks like it was originally broken, we can probably just remove `HAVE_SYS_POLL_H` entirely

src/backend/distributed/connection/connection_management.c
lithp: This might not be necessary since I think `FRONTEND` isn't defined while citus builds, but it was to make sure we don't hit [the `select` macro](https://github.com/postgres/postgres/blob/REL_10_STABLE/src/include/port/win32.h?utf8=%E2%9C%93#L374). We really just want to do our own thing and not think about whatever layer postgres tries to put between us and the syscall.

mtuncer:
@pykello @lithp we were having connection establishment issues when we are opening many connections. When we use wait event set, for some reason connection state update from remote node did not made it back to the coordinator, we kept getting invalid connection state (or something similar) when we try to proceed with the connection establishment routine. We tried some workarounds but none of them worked, therefore we rolled back to our original (and somewhat old)  poll/select logic. 

@anarazel does not agree with this and says we should find a way to use wait event set which may also include fixing this issue  in postgres side. 

At this moment, we would go with using select (preferably for windows builds only).

lithp:
I think we should probably hold off on this change and do it the right way.

lithp:
https://github.com/citusdata/citus/pull/2011/commits/b3dd772217f3869070dc9b13b453326751b85ac8 seems to have broken the build on Windows.

lithp:
@pykello okay, I think we're going to merge this after all. Friendly ping asking for a review :)

sumedhpathak:
@lithp Let's hold on merging. I sent a mail about it.

src/backend/distributed/executor/multi_client_executor.c
```diff
@@ -815,8 +814,19 @@ MultiClientCreateWaitInfo(int maxConnections)
 {
 	WaitInfo *waitInfo = palloc(sizeof(WaitInfo));
 
+	/* we subtract 2 to make room for the WL_POSTMASTER_DEATH and WL_LATCH_SET events */
+	if (maxConnections > FD_SETSIZE - 2)
+	{
+		maxConnections = FD_SETSIZE - 2;
```
pykello: Do we need to do this when poll is available?

src/backend/distributed/executor/multi_client_executor.c
```diff
@@ -853,9 +876,17 @@ MultiClientRegisterWait(WaitInfo *waitInfo, TaskExecutionStatus executionStatus,
...
+#else
+	int connectionFileDescriptor = 0;
+#endif
 
-	Assert(waitInfo->registeredWaiters < waitInfo->maxWaiters);
+	/* This is to make sure we could never register more than maxWaiters in Windows */
+	if (waitInfo->registeredWaiters >= waitInfo->maxWaiters)
+	{
+		return;
+	}
```
pykello: What would happen if a job needs more connections than `maxWaiters`? We either shouldn't allow more than `maxWaiters` connections by erroring out, or be sure that we can handle it.

src/backend/distributed/executor/multi_client_executor.c
```diff
@@ -879,6 +911,26 @@ MultiClientRegisterWait(WaitInfo *waitInfo, TaskExecutionStatus executionStatus,
...
+#else
+	connectionFileDescriptor = PQsocket(connection->pgConn);
+	if (connectionFileDescriptor > waitInfo->maxConnectionFileDescriptor)
+	{
+		waitInfo->maxConnectionFileDescriptor = connectionFileDescriptor;
+	}
+
+	if (executionStatus == TASK_STATUS_SOCKET_READ)
+	{
+		FD_SET(connectionFileDescriptor, &(waitInfo->exceptionFileDescriptorSet));
```
pykello: Exception fd set isn't same as `POLLERR`. Based on the man pages, it is more like `POLLPRI`. So we don't need to set it here. Actually, we also don't need to set `POLLERR` for `poll`, since it is ignored in `events`.

src/backend/distributed/connection/connection_management.c
```diff
@@ -469,6 +475,31 @@ FinishConnectionEstablishment(MultiConnection *connection)
...
+								   &writeFileDescriptorSet, &exceptionFileDescriptorSet,
+								   &selectTimeout);
+#endif /* HAVE_POLL */
```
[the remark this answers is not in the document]
lithp: Addressed in https://github.com/citusdata/citus/pull/2011/commits/488e5d1fdb552ede7de3b1bd96494303e8884f04

src/backend/distributed/connection/connection_management.c
```diff
@@ -10,7 +10,10 @@
 
 #include "postgres.h"
 
+#ifdef HAVE_POLL_H
 #include <poll.h>
+#endif
```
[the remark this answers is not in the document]
lithp: I'm not entirely sure, somewhere inside "postgres.h" it's being included for us. I assume [from here](https://github.com/postgres/postgres/blob/9d4649ca49416111aee2c84b7e4441a0b7aa2fac/src/include/port/win32_port.h)

src/backend/distributed/executor/multi_client_executor.c
lithp: Addressed in https://github.com/citusdata/citus/pull/2011/commits/114414d9f1f75e61476d2d13493ecb3668ab1505

src/backend/distributed/executor/multi_client_executor.c
lithp: In that case we'll have worse-than-usual performance. We'll wait until one of the `maxWaiters` connections becomes ready. When a connection outside of that set becomes ready we won't react until one of the other connections triggers. Since we go down the list and try to use every connection whenever one of them has an event I don't expect that this would have a noticeable impact on performance.

src/backend/distributed/executor/multi_client_executor.c
lithp: Addressed in https://github.com/citusdata/citus/pull/2011/commits/fd699f12534ebf78a84a0d6e77988d1a6961b0d9

src/backend/distributed/executor/multi_client_executor.c
```diff
@@ -832,14 +844,27 @@ MultiClientResetWaitInfo(WaitInfo *waitInfo)
...
 }
 
 
 /* MultiClientFreeWaitInfo frees a resources associated with a waitInfo struct. */
 void
 MultiClientFreeWaitInfo(WaitInfo *waitInfo)
 {
+#ifdef HAVE_POLL
 	pfree(waitInfo->pollfds);
+#else
```
pykello: what is the `#else` for?

src/backend/distributed/executor/multi_client_executor.c
```diff
@@ -832,14 +844,27 @@ MultiClientResetWaitInfo(WaitInfo *waitInfo)
 	waitInfo->registeredWaiters = 0;
 	waitInfo->haveReadyWaiter = false;
 	waitInfo->haveFailedWaiter = false;
+
+#ifdef HAVE_POLL
+#else
```
pykello: why not use `#ifndef` instead?

PR #2013, opened 2018-02-16
Prune shards once per relation in subquery pushdown

marcocitus:
DESCRIPTION: Faster shard pruning for subqueries

Took a shot at removing the redundant work in the subquery pushdown planner. A slight challenge is that sometimes a shard might be pruned away while its co-located shard is not. I ended up using a bitmap for which shard indexes in the sorted shard interval arrays to query and pass the shard indexes that don't get pruned away to `SubqueryTaskCreate`, which now avoids calling the router planner altogether.

Partially addresses #1245.

### DIFF
--- a/src/backend/distributed/planner/multi_physical_planner.c
+++ b/src/backend/distributed/planner/multi_physical_planner.c
@@ -33,6 +33,7 @@
 #include "distributed/citus_nodes.h"
 #include "distributed/citus_ruleutils.h"
 #include "distributed/colocation_utils.h"
+#include "distributed/deparse_shard_query.h"
 #include "distributed/master_protocol.h"
 #include "distributed/metadata_cache.h"
 #include "distributed/multi_router_planner.h"
@@ -130,7 +131,7 @@ static void ErrorIfUnsupportedShardDistribution(Query *query);
 static bool ShardIntervalsEqual(FmgrInfo *comparisonFunction,
 								ShardInterval *firstInterval,
 								ShardInterval *secondInterval);
-static Task * SubqueryTaskCreate(Query *originalQuery, ShardInterval *shardInterval,
+static Task * SubqueryTaskCreate(Query *originalQuery, int shardIndex,
 								 RelationRestrictionContext *restrictionContext,
 								 uint32 taskId);
 static List * SqlTaskList(Job *job);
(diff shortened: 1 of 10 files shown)


src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -2052,83 +2053,96 @@ SubquerySqlTaskList(Job *job, PlannerRestrictionContext *plannerRestrictionConte
...
-	 * That means all tables are reference tables and we can pick any any of them
-	 * as an anchor table.
-	 */
-	if (targetCacheEntry == NULL)
-	{
-		RangeTblEntry *rangeTableEntry = NULL;
+		if (taskRequiredForShardIndex == NULL)
+		{
+			shardCount = cacheEntry->shardIntervalArrayLength;
+			taskRequiredForShardIndex = (bool *) palloc0(shardCount);
```
onderkalaci: I think if we use a `bms` here instead of `taskRequiredForShardIndex `, we could even get rid of the following `for (shardOffset = 0; shardOffset < shardCount; shardOffset++)` which might be a useful win for very large shard counts?

src/include/distributed/master_metadata_utility.h
```diff
@@ -67,6 +67,7 @@ typedef struct ShardInterval
 	Datum minValue;     /* a shard's typed min value datum */
 	Datum maxValue;     /* a shard's typed max value datum */
 	uint64 shardId;
+	int shardIndex;
```
onderkalaci: Don't we need to add this to `CopyNodeShardInterval()` and similar related functions?

src/backend/distributed/planner/multi_physical_planner.c
mtuncer: 👍

src/include/distributed/master_metadata_utility.h
marcocitus: Added

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -2333,90 +2347,71 @@ ShardIntervalsEqual(FmgrInfo *comparisonFunction, ShardInterval *firstInterval,
  * subqueries with non equi-joins.).
  */
 static Task *
-SubqueryTaskCreate(Query *originalQuery, ShardInterval *shardInterval,
-				   RelationRestrictionContext *restrictionContext,
-				   uint32 taskId)
+SubqueryTaskCreate(Query *originalQuery, int shardIndex,
```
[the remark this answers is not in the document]
marcocitus: Fixed

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -2052,83 +2053,96 @@ SubquerySqlTaskList(Job *job, PlannerRestrictionContext *plannerRestrictionConte
...
-		RangeTblEntry *rangeTableEntry = (RangeTblEntry *) lfirst(rangeTableCell);
+		ereport(ERROR, (errmsg("cannot handle complex subqueries when the "
+							   "router executor is disabled")));
+	}
+
+	prunedRelationShardList = TargetShardIntervalsForQuery(subquery,
+														   relationRestrictionContext,
+														   &isMultiShardQuery);
+
+	forboth(prunedRelationShardCell, prunedRelationShardList,
```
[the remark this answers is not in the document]
marcocitus: Would be a bit of an ugly function with multiple values to return.

marcocitus:
> Wrt my local benchmarks, this is becoming one of the bottlenecks. So, I'm trying to come up with smth that could skip this

Since I expect we'll often query adjacent shards in the range-partitioned case, we could keep track of a min and max shardIndex and only iterate over the min-max range.

onderkalaci:
> I'm mostly trying to skip the loop in case of too many shards. (I can do some more proper benchmarks if you want before doing any changes.)


I've done some more proper benchmarks with 10K shards, query hitting < 10 shards, and it is not becoming a bottleneck. So, I'm taking my words back 😄

Issue #571, opened 2016-06-01
Feature request: json_agg support


rickyspeak:
postgres=# select json_agg(event) from events group by (uid);
ERROR:  unsupported aggregate function json_agg

For our data model this feature is pretty critical. I hope this is a pretty easy to implement.

Edit:  Forgot to mention that the shards are hash distributed on uid.  Though I doubt that matters very much for this issue.

ozgune:
@rickyspeak -- thanks for getting in touch with us. We'll look into the details of ``json_agg`` and will post an update in a few days.

Could you elaborate a bit more on your use-case? If you have an example table schema, could you share that with us too?

rickyspeak:
Thanks for getting back to me @ozgune .  We are exploring moving away from MongoDB to Postgres and the Citus sharding support is a major contributor there.  Our use-case is that we have event data in json format.  New events types with unknown fields can be added by end users, so our tables will primarily be a jsonb.  Events are coded by a uid and we want to query against many events that share the same uid and then return the collections of events that meet our query conditions. The table will just be these two columns.  Hash sharding on the uid column. With a GIN index on the event column.

For example:
uid | event
---------------
'1' | '{"a":1}'
'2' | '{'"a":1}'
'2' | '{"b":3}'

We would like to query for a uid that matches {"a":1} AND {"b":3}.  In order to do this query we'll select a good condition like {"a":1} query on that.  json_agg(event) on all the returned uid.  And then filter on the full condition {"a":1} AND {"b":3}.  Something like:
SELECT result FROM (
  SELECT json_agg(my_events.event) as result 
  FROM (
    (SELECT uid FROM my_events WHERE event @> '{"a":1}') as uids 
    LEFT JOIN my_events 
    ON my_events.uid = uids.uid 
    GROUP BY uids.uid
  )
) as intermediate
WHERE result @> '[{"a":1}]' AND  result @> '[{"b":3}]'

I'm sure there is a typo in there somewhere, but that is the gist.

At the innermost layer ideally we'd like to do multiple SELECT's on the shard key column (uid) and perform an INTERSECT on this value that is performed on the shard.

We will be exploring UPSERT models which would negate this issue, but our ingestion rate is on the order of 20-50k events per second and we've found ingestion to be a major issue and would rather stick with an INSERT only data model.

marcocitus:
[...]

PR #2015, opened 2018-02-17
Add support for json(b) aggregation

MarkusSintonen:
DESCRIPTION: Adds support for Postgres `jsonb` and `json` aggregation functions:

- `jsonb_agg`
- `json_agg`
- `jsonb_object_agg`
- `json_object_agg`

Aggregates are implemented similarly to `array_agg` where workers run the normal aggregate and the coordinator concatenates the results into the final json(b) array/object.

Feature is based on the discussion in https://github.com/citusdata/citus/issues/571.

Added test cases are based on the `array_agg`s test cases.

### DIFF
--- a/src/backend/distributed/citus--7.3-2--7.3-3.sql
+++ b/src/backend/distributed/citus--7.3-2--7.3-3.sql
@@ -1,0 +1,75 @@
+/* citus--7.3-2--7.3-3 */
+
+/*****************************************************************************
+ * Citus json aggregate helpers
+ *****************************************************************************/
+
+CREATE FUNCTION pg_catalog.citus_jsonb_concatenate(state jsonb, val jsonb)
+	RETURNS jsonb
+	LANGUAGE SQL
+AS $function$
+	SELECT CASE 
+		WHEN val IS NULL THEN state
+		WHEN jsonb_typeof(state) = 'null' THEN val
+		ELSE state || val
+	END;
+$function$;
+
+CREATE FUNCTION pg_catalog.citus_jsonb_concatenate_final(state jsonb)
+	RETURNS jsonb
+	LANGUAGE SQL
+AS $function$
+	SELECT CASE WHEN jsonb_typeof(state) = 'null' THEN NULL ELSE state END;
+$function$;
+
+CREATE FUNCTION pg_catalog.citus_json_concatenate(state json, val json)
+	RETURNS json
+	LANGUAGE SQL
+AS $function$
+	SELECT CASE 
+		WHEN val IS NULL THEN state
+		WHEN json_typeof(state) = 'null' THEN val
+		WHEN json_typeof(state) = 'object' THEN
+ 			(SELECT json_object_agg(key, value) FROM (
+		 		SELECT * FROM json_each(state)
+		 		UNION ALL
+		 		SELECT * FROM json_each(val)
+	 		) t)
+		ELSE 
+	 		(SELECT json_agg(a) FROM (
+	 			SELECT json_array_elements(state) AS a
+		 		UNION ALL
+		 		SELECT json_array_elements(val) AS a
+	 		) t)
+	END;
+$function$;
+
+CREATE FUNCTION pg_catalog.citus_json_concatenate_final(state json)
+	RETURNS json
+	LANGUAGE SQL
+AS $function$
+	SELECT CASE WHEN json_typeof(state) = 'null' THEN NULL ELSE state END;
+$function$;
+
+
+/*****************************************************************************
+ * Citus json aggregates
+ *****************************************************************************/
+
+CREATE AGGREGATE pg_catalog.jsonb_cat_agg(jsonb) (
+    SFUNC = citus_jsonb_concatenate,
+    FINALFUNC = citus_jsonb_concatenate_final,
+    STYPE = jsonb,
+    INITCOND = 'null'
+);
+COMMENT ON AGGREGATE pg_catalog.jsonb_cat_agg(jsonb)
+    IS 'concatenate input jsonbs into a single jsonb';
+    
+CREATE AGGREGATE pg_catalog.json_cat_agg(json) (
+    SFUNC = citus_json_concatenate,
+    FINALFUNC = citus_json_concatenate_final,
+    STYPE = json,
+    INITCOND = 'null'
+);
+COMMENT ON AGGREGATE pg_catalog.json_cat_agg(json)
+    IS 'concatenate input jsons into a single json';
--- a/src/test/regress/expected/multi_json_agg.out
+++ b/src/test/regress/expected/multi_json_agg.out
@@ -1,0 +1,183 @@
+--
+-- MULTI_JSON_AGG
+--
+SET citus.next_shard_id TO 520000;
+CREATE OR REPLACE FUNCTION array_sort (json)
+RETURNS json LANGUAGE SQL
+AS $$
+SELECT json_agg(value) FROM (
+	SELECT value FROM json_array_elements($1) ORDER BY value::jsonb
+) t
+$$;
+-- Check multi_cat_agg() aggregate which is used to implement json_agg()
+SELECT json_cat_agg(i) FROM 
+	(VALUES ('[1,{"a":2}]'::json), ('[null]'::json), (NULL), ('["3",5,4]'::json)) AS t(i);
+         json_cat_agg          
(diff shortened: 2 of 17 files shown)


src/backend/distributed/citus--7.3-2--7.3-3.sql
```diff
@@ -0,0 +1,75 @@
...
+
+CREATE FUNCTION pg_catalog.citus_json_concatenate(state json, val json)
+	RETURNS json
+	LANGUAGE SQL
+AS $function$
+	SELECT CASE 
+		WHEN val IS NULL THEN state
+		WHEN json_typeof(state) = 'null' THEN val
+		WHEN json_typeof(state) = 'object' THEN
+ 			(SELECT json_object_agg(key, value) FROM (
```
MarkusSintonen: This is not so elegant but postgres does not have `jsonb_concat` equivalent for `json`.

src/backend/distributed/citus--7.3-2--7.3-3.sql
```diff
@@ -0,0 +1,75 @@
...
+ * Citus json aggregate helpers
+ *****************************************************************************/
+
+CREATE FUNCTION pg_catalog.citus_jsonb_concatenate(state jsonb, val jsonb)
+	RETURNS jsonb
+	LANGUAGE SQL
+AS $function$
+	SELECT CASE 
+		WHEN val IS NULL THEN state
+		WHEN jsonb_typeof(state) = 'null' THEN val
```
marcocitus: Did you run into a specific problem that made you treat a JSONB `null` differently? Or do you think it's more conceptually sound?

src/backend/distributed/citus--7.3-2--7.3-3.sql
```diff
@@ -0,0 +1,75 @@
...
+ 			(SELECT json_object_agg(key, value) FROM (
+		 		SELECT * FROM json_each(state)
+		 		UNION ALL
+		 		SELECT * FROM json_each(val)
+	 		) t)
+		ELSE 
+	 		(SELECT json_agg(a) FROM (
+	 			SELECT json_array_elements(state) AS a
+		 		UNION ALL
+		 		SELECT json_array_elements(val) AS a
```
marcocitus: Given that this an internal function for concatenating objects and arrays it's probably not super important, but the behaviour is not consistent with jsonb_concat when val is not an array.

```
postgres=# SELECT citus_json_concatenate('[3]'::json, '3'::json);                                                                                                  
ERROR:  cannot call json_array_elements on a scalar
```

src/backend/distributed/citus--7.3-2--7.3-3.sql
marcocitus: It might be simpler to just convert to JSONB and back, which wouldn't be significantly more expensive since these functions pay the parsing/deparsing cost anyway.

src/test/regress/expected/multi_json_agg.out
```diff
@@ -0,0 +1,183 @@
+--
+-- MULTI_JSON_AGG
+--
+SET citus.next_shard_id TO 520000;
```
marcocitus: Setting the GUC is not really necessary unless you're creating new distributed tables.

src/backend/distributed/citus--7.3-2--7.3-3.sql
MarkusSintonen: Postgres `array_cat` and `jsonb_concat` treat `NULL` concatenation differently. `jsonb_concat` nullifies the value if we try to append `NULL` into to the object/array. Therefore we need to handle these cases some way or another.

src/backend/distributed/citus--7.3-2--7.3-3.sql
MarkusSintonen: Yes I did not take that into account since basically the added functions do not receive mixed types of json(b) values. They only receive arrays/objects coming from worker aggregations, not scalars. I can add explicit handling of scalars to `citus_json_concatenate` if we wish so.

src/backend/distributed/citus--7.3-2--7.3-3.sql
MarkusSintonen: We would lose consistent behaviour with vanilla `json_object_agg` since converting to `jsonb` will lose duplicated keys and ordering.

src/test/regress/expected/multi_json_agg.out
MarkusSintonen: I can clean that up

src/backend/distributed/citus--7.3-2--7.3-3.sql
marcocitus: Yep, `jsonb_concat` is unfortunately marked as `STRICT` which means that it always returns `NULL` if an argument is `NULL`, whereas `array_cat` is not. Defining a wrapper that handles NULL input is the right thing to do.

though jsonb_concat does treat 'null'::jsonb as a regular value
```sql
postgres=# SELECT jsonb_concat('null'::jsonb, '3'::jsonb);
 jsonb_concat 
──────────────
 [null, 3]
(1 row)

Time: 0.293 ms
postgres=# SELECT citus_jsonb_concatenate('null'::jsonb, '3'::jsonb);
 citus_jsonb_concatenate 
─────────────────────────
 3
(1 row)
```
so that made me wonder if there's anything special about `'null'::jsonb` in the Citus case. I realise that `'null'::jsonb` is the `INITCOND` of the aggregate so it's used to bootstrap the aggregate. On the other hand, you could have also used `NULL` as the `INITCOND` and check for `state IS NULL` here.

I think either way is fine, since the difference is philosphical and it perfectly solves the problem of distributing `jsonb_agg`, but was just making sure that I'm not missing anything.

src/backend/distributed/citus--7.3-2--7.3-3.sql
marcocitus: >  I can add explicit handling of scalars to citus_json_concatenate if we wish so.

Nah, seems like overkill.

src/backend/distributed/citus--7.3-2--7.3-3.sql
MarkusSintonen: There is something intresting going on when using `NULL` as the `INITCOND` here. It actually gets converted into `'null'::jsonb` somewhere along the way. So checking `state IS NULL` does not work. I really can not explain what happens in this case behind the scenes. So that PG aggregate initialization gets `NULL` actually converted to `'null'::jsonb`. I then chose to be explicit about the initial condition and set it to be `'null'`.

MarkusSintonen: Because of this the `'null'::jsonb` is used to detect whether anything was received via aggregate state transition. Then final function yields actual `NULL` if nothing was received (state is `'null'::jsonb`).

src/test/regress/expected/multi_json_agg.out
marcocitus: No worries, I'll merge this PR and then fix it up.

src/backend/distributed/citus--7.3-2--7.3-3.sql
```diff
@@ -0,0 +1,75 @@
...
+AS $function$
+	SELECT CASE WHEN json_typeof(state) = 'null' THEN NULL ELSE state END;
+$function$;
+
+
+/*****************************************************************************
+ * Citus json aggregates
+ *****************************************************************************/
+
+CREATE AGGREGATE pg_catalog.jsonb_cat_agg(jsonb) (
```
onderkalaci: It looks like we've broken `PARALLEL ` fields of these aggregates. Before citus created:
```SQL
postgres=# SELECT p1.oid, p1.proname
postgres-# FROM pg_proc AS p1
postgres-# WHERE provolatile = 'i' AND proparallel = 'u';
 oid | proname 
-----+---------
(0 rows)

postgres=# create extension citus ;
CREATE EXTENSION
postgres=# SELECT p1.oid, p1.proname
postgres-# FROM pg_proc AS p1
postgres-# WHERE provolatile = 'i' AND proparallel = 'u';
  oid  |    proname    
-------+---------------
 16492 | array_cat_agg
 16773 | jsonb_cat_agg
 16774 | json_cat_agg
(3 rows)
```

I hit this while running Postgresql's regression test `check-installed` when Citus is created.

Is that intended?

PR #2016, opened 2018-02-18
Support non-co-located joins between subqueries

onderkalaci:
DESCRIPTION:Support non-co-located joins between subqueries

(Psudeo-code for the changes: https://gist.github.com/onderkalaci/28d0808e14aa7563d575874847556297)

With #1804 (and related PRs), Citus gained the ability to plan subqueries that are not safe to pushdown. 

There are two high-level requirements for pushing down subqueries:

1. Individual subqueries that require a merge step (i.e., `GROUP BY` on non-distribution key, or `LIMIT` in the subquery etc). We've handled such subqueries via #1876. 

2. Combination of subqueries that are not joined on distribution keys. This PR aims to recursively plan some of such subqueries to make the whole query safe to pushdown.

This PR enables a wide range of subquery combinations. Below, I'll show some basic examples of what cases are enabled:

### `FROM` clause non colocated subquery joins:
```SQL
-- foo is individually safe to pushdown and joined on distribution keys
-- bar is individually safe to pushdown and joined on distribution keys
-- However, join among foo and bar is not on the distribution keys, so not safe to pushdown 
-- since the join is on `value_2`

    SELECT
        foo.value_2
    FROM
        (SELECT 
        	users_table.value_2 
        FROM 
        	users_table, events_table 
        WHERE 
        	users_table.user_id = events_table.user_id AND event_type IN (1,2,3,4)) as foo,
        (SELECT 
        	users_table.value_2 
        FROM 
        	users_table, events_table 
        WHERE 
        	users_table.user_id = events_table.user_id AND event_type IN (5,6,7,8)) as bar
    WHERE
        foo.value_2 = bar.value_2;

-- recursively plan bar such that the query becomes safe to pushdown
DEBUG:  generating subplan 5_1 for subquery SELECT users_table.value_2 FROM public.users_table, public.events_table WHERE ((users_table.user_id = events_table.user_id) AND (events_table.event_type = ANY (ARRAY[5, 6, 7, 8])))

```
### Non colocated subqueries in WHERE clause:
```SQL
-- the join is via `NOT IN` expression, which is not safe to pushdown
-- so recursively plan subquery in WHERE clause
    SELECT
        count(*)
    FROM
        events_table
    WHERE
        user_id
    NOT IN
        (SELECT user_id FROM events_table WHERE event_type = 2);


DEBUG:  generating subplan 9_1 for subquery SELECT user_id FROM public.events_table WHERE (event_type = 2)

```
### Individual subqueries that are not colocated (i.e., repartition joins in subqueries):
```SQL
SELECT count(*)
FROM
  (SELECT 
    users_table.user_id
   FROM 
    users_table, events_table
   WHERE 
    users_table.user_id = events_table.value_2 AND event_type IN (1,2,3,4)) AS foo,
  (SELECT users_table.user_id
   FROM users_table, events_table
   WHERE users_table.user_id = events_table.user_id AND event_type IN (5,6,7,8)) AS bar
WHERE foo.user_id = bar.user_id

-- foo is a repartition join, thus recursively plan in
DEBUG:  generating subplan 19_1 for subquery SELECT users_table.user_id FROM public.users_table, public.events_table WHERE ((users_table.user_id = events_table.value_2) AND (events_table.event_type = ANY (ARRAY[1, 2, 3, 4])))
```
 
### TODO list
- [x] More tests with CTEs, UNIONs, local tables etc
- [x]  Consider some more code/test refactoring, especially restriction eq. codes
- [x] Few comments missing and few variable names seems not very proper

Each commit is intended to include specific changes, so hopefully, the review would be easier.

### DIFF
--- a/src/backend/distributed/planner/query_colocation_checker.c
+++ b/src/backend/distributed/planner/query_colocation_checker.c
@@ -1,0 +1,314 @@
+/*-------------------------------------------------------------------------
+ *
+ * query_colocation_checker.c implements the logic for determining
+ * whether any subqueries in a given query are co-located (e.g.,
+ * distribution keys of the relations inside subqueries are equal).
+ *
+ * The main logic behind non colocated subquery joins is that we pick
+ * an anchor range table entry and check for distribution key equality
+ * of any  other subqueries in the given query. If for a given subquery,
+ * we cannot find distribution key equality with the anchor rte, we
+ * recursively plan that subquery.
+ *
+ * We also used a hacky solution for picking relations as the anchor range
+ * table entries. The hack is that we wrap them into a subquery. This is only
+ * necessary since some of the attribute equivalance checks are based on
+ * queries rather than range table entries.
+ *
+ * Copyright (c) 2018, Citus Data, Inc.
+ *-------------------------------------------------------------------------
+ */
+
+#include "postgres.h"
+
+#include "distributed/multi_logical_planner.h"
+#include "distributed/query_colocation_checker.h"
+#include "distributed/pg_dist_partition.h"
+#include "distributed/relation_restriction_equivalence.h"
+#include "nodes/makefuncs.h"
+#include "nodes/nodeFuncs.h"
+#include "parser/parsetree.h"
+#include "parser/parse_relation.h"
+#include "optimizer/planner.h"
+#include "optimizer/prep.h"
+
+
+static RangeTblEntry * AnchorRte(Query *subquery);
+static Query * WrapRteRelationIntoSubquery(RangeTblEntry *rteRelation);
+static List * UnionRelationRestrictionLists(List *firstRelationList,
+											List *secondRelationList);
+
+
+/*
+ * CreateColocatedJoinChecker is a helper function that simply calculates
+ * a ColocatedJoinChecker with the given input and returns it.
+ */
+ColocatedJoinChecker
+CreateColocatedJoinChecker(Query *subquery, PlannerRestrictionContext *restrictionContext)
+{
+	ColocatedJoinChecker colocatedJoinChecker;
+
+	RangeTblEntry *anchorRangeTblEntry = NULL;
+	Query *anchorSubquery = NULL;
+	PlannerRestrictionContext *anchorPlannerRestrictionContext = NULL;
+	RelationRestrictionContext *anchorRelationRestrictionContext = NULL;
+	List *anchorRestrictionEquivalences = NIL;
+
+	/* we couldn't pick an anchor subquery, no need to continue */
+	anchorRangeTblEntry = AnchorRte(subquery);
+	if (anchorRangeTblEntry == NULL)
+	{
+		colocatedJoinChecker.anchorRelationRestrictionList = NIL;
+
+		return colocatedJoinChecker;
+	}
+
+	if (anchorRangeTblEntry->rtekind == RTE_RELATION)
+	{
+		/*
+		 * If we get a relation as our anchor, wrap into a subquery. The only
+		 * reason that we wrap the relation into a subquery is that some of the utility
+		 * functions (i.e., FilterPlannerRestrictionForQuery()) rely on queries
+		 * not relations.
+		 */
+		anchorSubquery = WrapRteRelationIntoSubquery(anchorRangeTblEntry);
+	}
+	else if (anchorRangeTblEntry->rtekind == RTE_SUBQUERY)
+	{
+		anchorSubquery = anchorRangeTblEntry->subquery;
+	}
+	else
+	{
+		/* we don't expect any other RTE type here */
+		pg_unreachable();
+	}
+
+	anchorPlannerRestrictionContext =
+		FilterPlannerRestrictionForQuery(restrictionContext, anchorSubquery);
+	anchorRelationRestrictionContext =
+		anchorPlannerRestrictionContext->relationRestrictionContext;
+	anchorRestrictionEquivalences =
+		GenerateAllAttributeEquivalences(anchorPlannerRestrictionContext);
+
+	/* fill the non colocated planning context */
+	colocatedJoinChecker.subquery = subquery;
+	colocatedJoinChecker.subqueryPlannerRestriction = restrictionContext;
+
+	colocatedJoinChecker.anchorRelationRestrictionList =
+		anchorRelationRestrictionContext->relationRestrictionList;
+	colocatedJoinChecker.anchorAttributeEquivalences = anchorRestrictionEquivalences;
+
+	return colocatedJoinChecker;
+}
+
+
+/*
+ * AnchorRte gets a query and searches for a relation or a subquery within
+ * the join tree of the query such that we can use it as our anchor range
+ * table entry during our non colocated subquery planning.
+ *
+ * The function returns NULL if it cannot find a proper range table entry for our
+ * purposes. See the function for the details.
+ */
+static RangeTblEntry *
+AnchorRte(Query *subquery)
+{
+	FromExpr *joinTree = subquery->jointree;
+	Relids joinRelIds = get_relids_in_jointree((Node *) joinTree, false);
+	int currentRTEIndex = -1;
+	RangeTblEntry *anchorRangeTblEntry = NULL;
+
+	/*
+	 * Pick a random anchor relation or subquery (i.e., the first) for now. We
+	 * might consider picking a better rte as the anchor. For example, we could
+	 * iterate on the joinRelIds, and check which rteIndex has more distribution
+	 * key equiality with rteIndexes. For the time being, the current primitive
+	 * approach helps us in many cases.
+	 */
+	while ((currentRTEIndex = bms_next_member(joinRelIds, currentRTEIndex)) >= 0)
+	{
+		RangeTblEntry *currentRte = rt_fetch(currentRTEIndex, subquery->rtable);
+
+		/*
+		 * We always prefer distributed releations if we can find any. The
+		 * reason is that Citus is currently able to recursively plan
+		 * subqueries, but not relations.
+		 *
+		 * For the subqueries, make sure that the subquery contains at least one
+		 * distributed table and doesn't have a set operation.
+		 *
+		 * TODO: The set operation restriction might sound weird, but, the restriction
+		 * equivalance generation functions ignore set operations. We should
+		 * integrate the logic in SafeToPushdownUnionSubquery() to
+		 * GenerateAllAttributeEquivalences() such that the latter becomes aware of
+		 * the set operations.
+		 */
+		if (anchorRangeTblEntry == NULL && currentRte->rtekind == RTE_SUBQUERY &&
+			QueryContainsDistributedTableRTE(currentRte->subquery) &&
+			currentRte->subquery->setOperations == NULL &&
+			!ContainsUnionSubquery(currentRte->subquery))
+		{
+			/* found a subquery, keep it if we cannot find a relation */
+			anchorRangeTblEntry = currentRte;
+		}
+		else if (currentRte->rtekind == RTE_RELATION)
+		{
+			Oid relationId = currentRte->relid;
+
+			if (PartitionMethod(relationId) == DISTRIBUTE_BY_NONE)
+			{
+				/*
+				 * Reference tables should not be the anchor rte since they
+				 * don't have distribution key.
+				 */
+				continue;
+			}
+
+			anchorRangeTblEntry = currentRte;
+			break;
+		}
+	}
+
+	return anchorRangeTblEntry;
+}
+
+
+/*
+ * SubqueryColocated returns true if the input subquery has a distribution
+ * key equality with the anchor subquery. In other words, we refer the
+ * distribution key equality of relations as "colocation" in this context.
+ */
+bool
+SubqueryColocated(Query *subquery, ColocatedJoinChecker *checker)
+{
+	List *anchorRelationRestrictionList = checker->anchorRelationRestrictionList;
+	List *anchorAttributeEquivalances = checker->anchorAttributeEquivalences;
+
+	PlannerRestrictionContext *restrictionContext = checker->subqueryPlannerRestriction;
+	PlannerRestrictionContext *filteredPlannerContext =
+		FilterPlannerRestrictionForQuery(restrictionContext, subquery);
+	List *filteredRestrictionList =
+		filteredPlannerContext->relationRestrictionContext->relationRestrictionList;
+
+	List *unionedRelationRestrictionList = NULL;
+	RelationRestrictionContext *unionedRelationRestrictionContext = NULL;
+	PlannerRestrictionContext *unionedPlannerRestrictionContext = NULL;
+
+	/*
+	 * We merge the relation restrictions of the input subquery and the anchor
+	 * restrictions to form a temporary relation restriction context. The aim of
+	 * forming this temporary context is to check whether the context contains
+	 * distribution key equality or not.
+	 */
+	unionedRelationRestrictionList =
+		UnionRelationRestrictionLists(anchorRelationRestrictionList,
+									  filteredRestrictionList);
+
+	/*
+	 * We already have the attributeEquivalances, thus, only need to prepare
+	 * the planner restrictions with unioned relations for our purpose of
+	 * distribution key equality. Note that we don't need to calculate the
+	 * join restrictions, we're already relying on the attributeEquivalances
+	 * provided by the context.
+	 */
+	unionedRelationRestrictionContext = palloc0(sizeof(RelationRestrictionContext));
+	unionedRelationRestrictionContext->relationRestrictionList =
+		unionedRelationRestrictionList;
+
+	unionedPlannerRestrictionContext = palloc0(sizeof(PlannerRestrictionContext));
+	unionedPlannerRestrictionContext->relationRestrictionContext =
+		unionedRelationRestrictionContext;
+
+	if (!RestrictionEquivalenceForPartitionKeysViaEquivalances(
+			unionedPlannerRestrictionContext,
+			anchorAttributeEquivalances))
+	{
+		return false;
+	}
+
+	return true;
+}
+
+
+/*
+ * WrapRteRelationIntoSubquery wraps the given relation range table entry
+ * in a newly constructed "(SELECT * FROM table_name as anchor_relation)" query.
+ *
+ * Note that the query returned by this function does not contain any filters or
+ * projections. The returned query should be used cautiosly and it is mostly
+ * designed for generating a stub query.
+ */
+static Query *
+WrapRteRelationIntoSubquery(RangeTblEntry *rteRelation)
+{
+	Query *subquery = makeNode(Query);
+	RangeTblRef *newRangeTableRef = makeNode(RangeTblRef);
+	RangeTblEntry *newRangeTableEntry = NULL;
+	Var *targetColumn = NULL;
+	TargetEntry *targetEntry = NULL;
+
+	subquery->commandType = CMD_SELECT;
+
+	/* we copy the input rteRelation to preserve the rteIdentity */
+	newRangeTableEntry = copyObject(rteRelation);
+	subquery->rtable = list_make1(newRangeTableEntry);
+
+	/* set the FROM expression to the subquery */
+	newRangeTableRef = makeNode(RangeTblRef);
+	newRangeTableRef->rtindex = 1;
+	subquery->jointree = makeFromExpr(list_make1(newRangeTableRef), NULL);
+
+	/* Need the whole row as a junk var */
+	targetColumn = makeWholeRowVar(newRangeTableEntry, newRangeTableRef->rtindex, 0,
+								   false);
+
+	/* create a dummy target entry */
+	targetEntry = makeTargetEntry((Expr *) targetColumn, 1, "wholerow", true);
+
+	subquery->targetList = lappend(subquery->targetList, targetEntry);
+
+	return subquery;
+}
+
+
+/*
+ * UnionRelationRestrictionLists merges two relation restriction lists
+ * and returns a newly allocated list. The merged relation restriction
+ * list doesn't contain any duplicate elements.
+ */
+static List *
+UnionRelationRestrictionLists(List *firstRelationList, List *secondRelationList)
+{
+	RelationRestrictionContext *unionedRestrictionContext = NULL;
+	List *unionedRelationRestrictionList = NULL;
+	ListCell *relationRestrictionCell = NULL;
+	Relids rteIdentities = NULL;
+	List *allRestrictionList = NIL;
+
+	/* list_concat destructively modifies the first list, thus copy it */
+	firstRelationList = list_copy(firstRelationList);
+	allRestrictionList = list_concat(firstRelationList, secondRelationList);
+
+	foreach(relationRestrictionCell, allRestrictionList)
+	{
+		RelationRestriction *restriction =
+			(RelationRestriction *) lfirst(relationRestrictionCell);
+		int rteIdentity = GetRTEIdentity(restriction->rte);
+
+		/* already have the same rte, skip */
+		if (bms_is_member(rteIdentity, rteIdentities))
+		{
+			continue;
+		}
+
+		unionedRelationRestrictionList =
+			lappend(unionedRelationRestrictionList, restriction);
+
+		rteIdentities = bms_add_member(rteIdentities, rteIdentity);
+	}
+
+	unionedRestrictionContext = palloc0(sizeof(RelationRestrictionContext));
+	unionedRestrictionContext->relationRestrictionList = unionedRelationRestrictionList;
+
+	return unionedRelationRestrictionList;
+}
--- a/src/backend/distributed/planner/recursive_planning.c
+++ b/src/backend/distributed/planner/recursive_planning.c
@@ -116,6 +116,12 @@ typedef struct VarLevelsUpWalkerContext
 static DeferredErrorMessage * RecursivelyPlanSubqueriesAndCTEs(Query *query,
 															   RecursivePlanningContext *
 															   context);
+
+static bool ShouldRecursivelyPlanNonColocatedSubqueries(Query *subquery,
+														RecursivePlanningContext *
+														context);
+static void RecursivelyPlanNonColocatedSubqueries(Query *subquery,
+												  RecursivePlanningContext *context);
 static bool ShouldRecursivelyPlanAllSubqueriesInWhere(Query *query);
 static bool RecursivelyPlanAllSubqueries(Node *node,
 										 RecursivePlanningContext *planningContext);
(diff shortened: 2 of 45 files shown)


src/backend/distributed/planner/recursive_planning.c
```diff
@@ -252,10 +309,516 @@ RecursivelyPlanSubqueriesAndCTEs(Query *query, RecursivePlanningContext *context
...
+
+	/*
+	 * This check helps us in two ways:
+	 *   (i) We're not targeting queries that don't include subqueries at all,
+	 *       they should go through regular planning.
+	 *  (ii) Lower level subqueries are already recursively planned, so we should
+	 *       only bother non-colocated subquery joins, which only happens when
+	 *       there are subqueries.
+	 */
+	if (SubqueryEntryList(subquery) == NIL && SublinkList(subquery) == NIL)
```
marcocitus: Since we're immediately throwing away these lists and this function may get called very often, could we implement this in a way that does not allocate memory?

marcocitus:
Higher-level question: Could you describe how we can go from the current implementation to one that re-partitions the intermediate results when the result of the subquery is big?

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -252,10 +309,516 @@ RecursivelyPlanSubqueriesAndCTEs(Query *query, RecursivePlanningContext *context
...
+												 context);
+}
+
+
+/*
+ * AnchorSubqueryRte gets a query and searches for a subquery within the join tree of
+ * the query such that we can use it as our anchor subquery during our planning.
+ *
+ * The function returns NULL if it cannot find a proper range table entry for our
+ * purposes.
```
marcocitus: What's "proper"?

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -252,10 +309,516 @@ RecursivelyPlanSubqueriesAndCTEs(Query *query, RecursivePlanningContext *context
...
+	Query *anchorSubquery = NULL;
+	PlannerRestrictionContext *anchorPlannerRestrictionContext = NULL;
+	RelationRestrictionContext *anchorRelationRestrictionContext = NULL;
+	List *anchorRestrictionEquivalences = NIL;
+
+	FromExpr *joinTree = subquery->jointree;
+	PlannerRestrictionContext *plannerRestrictionContext = NULL;
+
+	/* we couldn't pick an anchor subquery, so we cannot continue */
+	anchorSubquery = AnchorSubquery(subquery);
```
marcocitus: under what conditions can we not pick an anchor subquery?

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -252,10 +309,516 @@ RecursivelyPlanSubqueriesAndCTEs(Query *query, RecursivePlanningContext *context
...
+	 * with rteIndexes. For the time being, the current primitive approach helps
+	 * us in many cases.
+	 */
+	while ((currentRTEIndex = bms_next_member(joinRelIds, currentRTEIndex)) >= 0)
+	{
+		RangeTblEntry *rte = rt_fetch(currentRTEIndex, subquery->rtable);
+
+		/*
+		 * We always prefer subqueries if we can find a suitable one for our
+		 * purposes.
```
marcocitus: why do we prefer subqueries over relations? If you have `table JOIN subquery ON (non partition column)` wouldn't that be enabled by picking the table as an anchor?

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -252,10 +309,516 @@ RecursivelyPlanSubqueriesAndCTEs(Query *query, RecursivePlanningContext *context
...
+		 *
+		 * For the subquery, make sure that our subquery contains at least one
+		 * distributed table and doesn't have a set operation.
+		 *
+		 * The set operation restriction might sound weird, but, the restriction
+		 * equivalance generation functions ignore set operations. We should
```
marcocitus: Maybe add a TODO or XXX for future archeologists

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -252,10 +309,516 @@ RecursivelyPlanSubqueriesAndCTEs(Query *query, RecursivePlanningContext *context
...
 	return NULL;
 }
 
 
+/*
+ * WrapRteRelationIntoSubquery wraps the given relation range table entry
+ * in a newly constructed "(SELECT * FROM table_name as anchor_relation)" query.
+ */
+static Query *
+WrapRteRelationIntoSubquery(RangeTblEntry *rteRelation)
```
marcocitus: could this ever lead to the table being pulled to the coordinator? (e.g. at a higher level of recursion) if that's something we do, is there a case for pushing down filters and projections here?

src/backend/distributed/planner/recursive_planning.c
onderkalaci: Sure, I can write new functions which returns immediately when it finds a sublink or subquery. But, that'd mean to almost copy both functions. 

I'm not that much concerned about the memory given that  I expect the other parts of the planning would be the bottleneck not here if the there are many subqueries in the query (i.e., the lists are lengthy).

Still, I could implement more efficient functions I you prefer to do so.

onderkalaci:
>  Consider some more code/test refactoring, especially restriction eq. codes

This is from the TODO list I've added. Just a clarification, that's actually the decision of whether stick to `AnchorSubquery` or switch to `AnchorRTE`, which requires refactoring in restriction eq. codes.

onderkalaci:
@marcocitus I applied your in-person feedback, you can check the last 4 commits. I think it simplified the code. Thanks!

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -252,10 +314,515 @@ RecursivelyPlanSubqueriesAndCTEs(Query *query, RecursivePlanningContext *context
...
+												 context);
+}
+
+
+/*
+ * CreateNonColocatedSubqueryJoinContext is a helper function that simply calculates
+ * a NonColocatedSubqueryContext with the given input and returns it.
+ */
+static NonColocatedSubqueryContext
+CreateNonColocatedSubqueryJoinContext(Query *subquery, RangeTblEntry *anchorRangeTblEntry,
```
marcocitus: This file is getting  big. Would it be possible to move some of these functions into a different file that we can eventually expand with more advanced logic? e.g. `noncolocated_join_planning.c`

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -111,17 +115,60 @@ typedef struct VarLevelsUpWalkerContext
...
 
 
+/*
+ * NonColocatedSubqueryContext is used to recursively plan subquery
+ * joins that are not on the distribution column.
+ */
+typedef struct NonColocatedSubqueryContext
+{
+	Query *subquery;
+	List *anchorAttributeEquivalances;
```
marcocitus: Equivalences

src/backend/distributed/planner/recursive_planning.c
marcocitus: Kind of. Also for readability, it'd be much nicer if there was a `ContainsSubquery(subquery)` function.

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -252,10 +314,515 @@ RecursivelyPlanSubqueriesAndCTEs(Query *query, RecursivePlanningContext *context
...
+	NonColocatedSubqueryContext nonColocatedPlanningContext;
+
+	RangeTblEntry *anchorRangeTblEntry = NULL;
+
+	FromExpr *joinTree = subquery->jointree;
+	PlannerRestrictionContext *restrictionContext = NULL;
+
+	/* we couldn't pick an anchor subquery, so we cannot continue */
+	anchorRangeTblEntry = AnchorRte(subquery);
+	if (anchorRangeTblEntry == NULL)
```
marcocitus: Is this something we could move into `CreateNonColocatedSubqueryJoinContext`, given that we already pass a `subquery`? Mostly asking because it makes it easier to move `CreateNonColocatedSubqueryJoinContext` into a different file.

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -111,17 +115,60 @@ typedef struct VarLevelsUpWalkerContext
 } VarLevelsUpWalkerContext;
 
 
+/*
+ * NonColocatedSubqueryContext is used to recursively plan subquery
+ * joins that are not on the distribution column.
+ */
+typedef struct NonColocatedSubqueryContext
```
marcocitus: Since it doesn't just concern subqueries, should we call it `NonColocatedJoinContext`? (also wondering if there is a different word than context that applies here).

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -252,10 +314,515 @@ RecursivelyPlanSubqueriesAndCTEs(Query *query, RecursivePlanningContext *context
...
+}
+
+
+/*
+ * SubqueryColocated returns true if the input subquery has a distribution
+ * key equality with the anchor subquery. In other words, we refer the
+ * distribution key equality of relations as "colocation" in this context.
+ */
+static bool
+SubqueryColocated(Query *subquery, NonColocatedSubqueryContext *context)
```
marcocitus: This is also a candidate for moving into noncolocated_join_planning.c

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -533,6 +1100,22 @@ ShouldRecursivelyPlanSubquery(Query *subquery)
 	}
 	else if (DeferErrorIfCannotPushdownSubquery(subquery, false) == NULL)
 	{
+		/*
+		 * We should do one more check for the distribution key equality.
+		 *
+		 * If the input query to the planner doesn't contain distribution key equality,
+		 * we should further check whether this individual subquery contains or not.
+		 *
+		 * If all the relations are not joined on their distribution keys for the
```
marcocitus: If all the relations are not joined -> if not all relations are joined

(the error message that says this always bugs me :) )

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -533,6 +1100,22 @@ ShouldRecursivelyPlanSubquery(Query *subquery)
...
+		 * subquery, we cannot pushdown the it, thus, recursively plan it.
```
marcocitus: "push the it..." -> "push it down and therefore recursively plan it."

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -252,10 +314,515 @@ RecursivelyPlanSubqueriesAndCTEs(Query *query, RecursivePlanningContext *context
...
+{
+	List *sublinkList = SublinkList(query);
+	ListCell *sublinkCell = NULL;
+
+	foreach(sublinkCell, sublinkList)
+	{
+		SubLink *sublink = (SubLink *) lfirst(sublinkCell);
+		Query *subselect = (Query *) sublink->subselect;
+
+		if (subselect == NULL)
```
marcocitus: is subselect ever NULL here?

src/backend/distributed/planner/relation_restriction_equivalence.c
```diff
@@ -143,6 +141,72 @@ static bool JoinRestrictionListExistsInContext(JoinRestriction *joinRestrictionI
...
+/*
+ * QueryContainsDistributionKeyEquality returns true if either
+ *    (i)  there exists join in the query and all relations joined on their
+ *         partition keys
+ *    (ii) there exists only union set operations and all relations has
+ *         partition keys in the same ordinal position in the query
+ */
+bool
+QueryContainsDistributionKeyEquality(PlannerRestrictionContext *plannerRestrictionContext,
+									 Query *originalQuery)
```
marcocitus: minor: I find "Contains" a bit confusing (does it contain 1 distribution key equality?).  Maybe `AllDistributionKeysInQueryAreEqual`.

Super minor: it seems a bit more natural/common that the query comes first and the context after.

src/backend/distributed/planner/relation_restriction_equivalence.c
```diff
@@ -399,15 +448,21 @@ FindTranslatedVar(List *appendRelList, Oid relationOid, Index relationRteIndex,
...
+ * provided externally (i.e., not NIL), the function uses it and skips generating it from
+ * the given planner context. This provides some flexibilty around how to call the function.
+ * It is especially useful if you have a pre-computed attribute equivalances, which saves
+ * us many CPU cycles.
  */
 bool
 RestrictionEquivalenceForPartitionKeys(PlannerRestrictionContext *
-									   plannerRestrictionContext)
+									   plannerRestrictionContext,
+									   List *allAttributeEquivalenceList)
```
marcocitus: I'm a bit confused by the extra parameter that completely changes the behaviour of the function. Should this be a separate function?

src/test/regress/expected/multi_mx_router_planner.out
```diff
@@ -441,71 +441,27 @@ DEBUG:  push down of limit count: 5
...
-DETAIL:  Creating dependency on merge taskId 11
-DEBUG:  pruning merge fetch taskId 10
-DETAIL:  Creating dependency on merge taskId 14
-DEBUG:  pruning merge fetch taskId 11
-DETAIL:  Creating dependency on merge taskId 14
-ERROR:  the query contains a join that requires repartitioning
-HINT:  Set citus.enable_repartition_joins to on to enable repartitioning
+ORDER BY test.word_count DESC, articles_hash.id LIMIT 5;
+ERROR:  missing FROM-clause entry for table "articles_hash"
+LINE 3: ORDER BY test.word_count DESC, articles_hash.id LIMIT 5;
```
marcocitus: parse error?

src/test/regress/expected/multi_subquery_complex_queries.out
```diff
@@ -423,8 +425,25 @@ GROUP BY
...
+DEBUG:  Plan 16 query after replacing subqueries and CTEs: SELECT event_types AS types, count(*) AS sumofeventtype FROM (SELECT q.user_id, q."time", q.event_types, t.user_id, random() AS random FROM ((SELECT t_1.user_id, t_1."time", unnest(t_1.collected_events) AS event_types FROM (SELECT t1.user_id, min(t1."time") AS "time", array_agg(t1.event ORDER BY t1."time", t1.event DESC) AS collected_events FROM (SELECT intermediate_result."time", intermediate_result.event, intermediate_result.user_id FROM read_intermediate_result('16_5'::text, 'binary'::citus_copy_format) intermediate_result("time" timestamp without time zone, event integer, user_id integer)) t1 GROUP BY t1.user_id) t_1) q JOIN (SELECT users.user_id FROM public.users_table users WHERE ((users.value_1 > 0) AND (users.value_1 < 4))) t ON ((t.user_id = q.user_id)))) final_query(user_id, "time", event_types, user_id_1, random) GROUP BY event_types ORDER BY event_types
```
marcocitus: I'm a bit scared by the amount of debug output in some of these tests and possible future maintenance pain. Is it necessary here?

marcocitus:
> DESCRIPTION: Recursively plan subqueries that lack distribution key equality

For users reading the changelog, maybe "Support non-co-located joins between subqueries"

src/backend/distributed/planner/relation_restriction_equivalence.c
onderkalaci: I liked `AllDistributionKeysInQueryAreEqual `

src/test/regress/expected/multi_mx_router_planner.out
onderkalaci: oops, fixed

src/backend/distributed/planner/recursive_planning.c
onderkalaci: Hmm, that's a fair point. 

The main problem I see is how do we pass `RecursivePlanningContext ` struct to the lower level functions? I think we don't want to make it public as well. And, without making that public, the benefit of moving into a separate file seems low since half of the functions use that.

What if we create a new header file (e.g., 'recursive_planning_typedefs.h'), which only includes the `RecursivePlanningContext` struct, and we comment that it should only be included by recursive planning source files?

src/test/regress/expected/multi_subquery_complex_queries.out
onderkalaci: Actually, I also realized that. They are mostly here to ensure that the order of planning isn't broken. And, given that each test is self-contained, I left them.

Do you have any preference?

marcocitus:
> The main problem I see is how do we pass RecursivePlanningContext struct to the lower level functions? I think we don't want to make it public as well. And, without making that public, the benefit of moving into a separate file seems low since half of the functions use that.

I think the candidates for moving are `NonColocatedJoinContext`, `CreateNonColocatedJoinContext`, `WrapRteRelationIntoSubquery`, `AnchorRte`, `SubqueryColocated`, `UnionRelationRestrictionLists`, which don't depend on `RecursivePlanningContext`. Basically, you'd encapsulate the logic for determining whether subqueries are co-located behind a simple API (2 functions), and keep the higher-level planning logic in one place.

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -252,10 +317,605 @@ RecursivelyPlanSubqueriesAndCTEs(Query *query, RecursivePlanningContext *context
...
+
+	unionedRestrictionContext = palloc0(sizeof(RelationRestrictionContext));
+	unionedRestrictionContext->relationRestrictionList = unionedRelationRestrictionList;
+
+	return unionedRelationRestrictionList;
+}
+
+
+/*
+ * RecursivelyPlanNonColocatedSubqueriesInWhere ge
```
marcocitus: truncated line

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -553,6 +1229,39 @@ ShouldRecursivelyPlanSubquery(Query *subquery)
...
+ */
+static bool
+AllDistributionKeysInSubqueryAreEqual(Query *subquery,
+									  PlannerRestrictionContext *restrictionContext)
+{
+	bool allDistributionKeysInSubqueryAreEqual = false;
+	PlannerRestrictionContext *filteredRestrictionContext = NULL;
+
+	/* we don't support distribution eq. checks for CTEs yet */
+	if (subquery->cteList != NIL)
```
marcocitus: Not super important, but I suppose you can make an exception for CTEs without distributed tables.

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -252,10 +317,605 @@ RecursivelyPlanSubqueriesAndCTEs(Query *query, RecursivePlanningContext *context
...
+	Node *queryQuals = NULL;
+
+	if (!joinTree)
+	{
+		return false;
+	}
+
+	queryQuals = joinTree->quals;
+
+	return ContainsSubqueryInWhereClauseWalker(queryQuals);
```
marcocitus: I think you can use `FindNodeCheck` here and terminate if you find a `Query`.

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -252,10 +317,605 @@ RecursivelyPlanSubqueriesAndCTEs(Query *query, RecursivePlanningContext *context
...
+	return false;
+}
+
+
+/*
+ * ContainsSubqueryInWhereClause returns true if the input query contains any subqueries
+ * in the WHERE clause.
+ */
+static bool
+ContainsSubqueryInWhereClause(Query *query)
```
marcocitus: Would it make sense to move this function to the logical planner and use it to replace `SublinkList`? There's actually no code that uses the output of `SublinkList`.

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -252,10 +317,605 @@ RecursivelyPlanSubqueriesAndCTEs(Query *query, RecursivePlanningContext *context
...
+	List *rangeTableList = query->rtable;
+	List *joinTreeTableIndexList = NIL;
+	ListCell *joinTreeTableIndexCell = NULL;
+
+	/*
+	 * Extract all range table indexes from the join tree. Note that here we
+	 * only walk over range table entries at this level and do not recurse into
+	 * subqueries.
+	 */
+	ExtractRangeTableIndexWalker((Node *) query->jointree, &joinTreeTableIndexList);
```
marcocitus: Now we're still building a list and it's a lot of (duplicate) code. Why not something like:

```
bool
JoinTreeContainsSubquery(Node *joinTreeNode, Query *query)
{
    if (joinTreeNode == NULL)
    {
        return false;
    }

    if (IsA(joinTreeNode, RangeTblRef))
    {
        RangeTblRef *rangeTableRef = (RangeTblRef *) joinTreeNode;
        RangeTblEntry *rangeTableEntry = rt_fetch(rangeTableRef->rtindex, query->rtable);
        if (rangeTableEntry->rtekind == RTE_SUBQUERY)
        {
            return true;
        }
    }

    return expression_tree_walker(joinTreeNode, JoinTreeContainsSubqueryWalker, query);
}
```

marcocitus:
> I think having context in the name complies with both Citus and Postgres conventions, no?

A few things bug me:
1) We pass down 2 different contexts in `RecursivelyPlanNonColocatedJoinWalker`
2) There are now 3 different "contexts" in this file
3) If feel like it represents some semantically meaningful properties of the (sub)query, rather than the usual catch-all context which we we only use because C lacks classes / anonymous functions

onderkalaci:
> I think the candidates for moving are NonColocatedJoinContext, CreateNonColocatedJoinContext, WrapRteRelationIntoSubquery, AnchorRte, SubqueryColocated, UnionRelationRestrictionLists, which don't depend on RecursivePlanningContext

Note that we'd need to keep `NonColocatedJoinContext` (or whatever name we give) should be in the header file. Is that what you ask for?

marcocitus:
> Note that we'd need to keep NonColocatedJoinContext (or whatever name we give) should be in the header file. Is that what you ask for?

Yes

onderkalaci:
> Yes

I didn't want to do this large refactoring twice so preferred to ask twice 😄

marcocitus:
or, if it's possible to define the struct in the C file and the only declare the type in the header, that would be even better

onderkalaci:
> or, if it's possible to define the struct in the C file and the only declare the type in the header, that would be even better

That seems hard since `RecursivelyPlanNonColocatedJoinWalker` uses `nonColocatedJoinContext->subquery->rtable;`

src/backend/distributed/planner/recursive_planning.c
onderkalaci: > Now we're still building a list and it's a lot of (duplicate) code. 

Oops, I copy & pasted the code and ignored the list

marcocitus:
> That seems hard since RecursivelyPlanNonColocatedJoinWalker uses nonColocatedJoinContext->subquery->rtable;

Well, you could just define a third function to get the subquery, but I can live without out it.

src/backend/distributed/planner/recursive_planning.c
onderkalaci: `RecursivelyPlanNonColocatedSubqueriesInWhere` uses it.

But, I can replace the call in logical planner with this one

src/backend/distributed/planner/recursive_planning.c
onderkalaci: We don't support eq. of local tables anyway. So, keeping it for now as it is?

marcocitus: I was thinking more in terms of constant CTEs and reference tables, but we can skip this.

src/test/regress/expected/non_colocated_subquery_joins.out
```diff
@@ -0,0 +1,917 @@
...
+DEBUG:  Plan 13 query after replacing subqueries and CTEs: SELECT foo.user_id FROM (SELECT users_table.user_id, events_table.event_type FROM public.users_table, public.events_table WHERE ((users_table.user_id = events_table.user_id) AND (events_table.event_type = ANY (ARRAY[1, 2, 3, 4])))) foo, (SELECT intermediate_result.user_id FROM read_intermediate_result('13_1'::text, 'binary'::citus_copy_format) intermediate_result(user_id integer)) bar WHERE ((foo.user_id = bar.user_id) AND (NOT (foo.user_id IN (SELECT intermediate_result.user_id FROM read_intermediate_result('13_2'::text, 'binary'::citus_copy_format) intermediate_result(user_id integer)))))
+ valid 
+-------
+ t
+(1 row)
+
+-- Subqueries in WHERE and FROM are mixed
+-- In this query, one of the joins in the FROM clause is not colocated and subquery in WHERE clause is not colocated
+-- similar to the above, but, this time bar is the anchor subquery
+SELECT true AS valid FROM explain_json_2($$
```
onderkalaci: @marcocitus this might be confusing to you. note that this is another edge we should consider when we implement outer joins via recursive planning. We currently error out for this query because the outer side of the semi-join is intermediate results. However, for this query, the inner side of the join also doesn't contain any distributed tables (e.g., another intermediate result).

onderkalaci:
@marcocitus Should be ready for you to look again

src/backend/distributed/planner/query_colocation_checker.c
```diff
@@ -0,0 +1,303 @@
+/*-------------------------------------------------------------------------
+ *
+ * query_colocation_checker.c implements the logic for determining
+ * whether any subqueries in a given query are co-located (e.g.,
+ * distribution keys of the relations inside subqueries are equal).
+ *
+ * Copyright (c) 2018, Citus Data, Inc.
```
marcocitus: It would be nice to add a paragraph about the details of the implementation here (mainly, checking against anchor RTE, wrapping tables in a subquery).

src/test/regress/expected/multi_mx_explain_0.out
```diff
@@ -59,7 +59,7 @@ EXPLAIN (COSTS FALSE, FORMAT TEXT)
 	SELECT l_quantity, count(*) count_quantity FROM lineitem_mx
 	GROUP BY l_quantity ORDER BY count_quantity, l_quantity;
 Sort
-  Sort Key: COALESCE((sum((COALESCE((sum(remote_scan.count_quantity))::bigint, '0'::bigint))))::bigint, '0'::bigint), remote_scan.l_quantity
+  Sort Key: COALESCE((pg_catalog.sum((COALESCE((pg_catalog.sum(remote_scan.count_quantity))::bigint, '0'::bigint))))::bigint, '0'::bigint), remote_scan.l_quantity
```
marcocitus: What's happening here?

src/test/regress/expected/multi_simple_queries_0.out
```diff
@@ -215,60 +215,7 @@ ERROR:  Complex subqueries and CTEs are not supported when task_executor_type is
...
- 44 |      16793
- 45 |        864
- 46 |      17702
- 47 |       1772
- 48 |      18610
- 49 |       2681
- 50 |      19519
-(50 rows)
-
+ERROR:  Complex subqueries and CTEs are not supported when task_executor_type is set to 'task-tracker'
```
marcocitus: Just highlighting this for future readers of the PR. We've consciously decided to stop relying on postgres planner transformations for subqueries, since we cannot deparse the resulting query trees. Therefore, some excotic cases that used to be supported through regular execution now require re-partitioning or recursive planning.

src/test/regress/expected/set_operations.out
```diff
@@ -658,22 +668,24 @@ DEBUG:  Plan 137 query after replacing subqueries and CTEs: SELECT x, y FROM rec
 
 -- correlated subquery with union in WHERE clause
 SELECT * FROM test a WHERE x IN (SELECT x FROM test b UNION SELECT y FROM test c WHERE a.x = c.x) ORDER BY 1,2;
-DEBUG:  generating subplan 141_1 for subquery SELECT x FROM recursive_union.test b
+DEBUG:  generating subplan 144_1 for subquery SELECT x FROM recursive_union.test b
+DEBUG:  skipping recursive planning for the subquery since it contains references to outer queries
+DEBUG:  skipping recursive planning for the subquery since it contains references to outer queries
```
marcocitus: a thought to consider for the future is that when there are correlated subqueries, we may want to prefer them as an anchor.  

(e.g. for situations like `<subquery partitioned by A> join <subquery partitioned by B> where <correlated subquery partitioned by B>`). We'd fail to plan this query because we'd fail to recursively plan the last subquery, but we'd succeed if it was the anchor.)

PR #2022, opened 2018-02-22
Cross-platform regression tests

lithp:


src/test/regress/sql/multi_utility_statements.sql
```diff
@@ -101,7 +101,7 @@ COPY (
 -- check copying to file
 -- (quiet off to force number of copied records to be displayed)
 \set QUIET off
-COPY nation TO '/dev/null';
+\COPY nation TO :dev_null;
```
pykello: When I run the tests, a file named `:dev_null` is created in the current directory. Maybe this variable is not passed to all test scripts.

src/test/regress/output/multi_copy.source
```diff
@@ -640,8 +640,8 @@ SELECT master_create_worker_shards('super_packed_numbers_hash', 4, 1);
  
 (1 row)
 
-COPY (SELECT 1, ROW(ROW(42, 42), ROW(42, 42))) TO '/tmp/copy_test_composite_of_composite';
-COPY super_packed_numbers_hash FROM '/tmp/copy_test_composite_of_composite';
+COPY (SELECT 1, ROW(ROW(42, 42), ROW(42, 42))) TO '@testtablespace@\copy_test_composite_of_composite';
```
pykello: Why are we using `@testtablespace@` as the temp directory? Can't we use something like `mktemp` which seems to be cross-platform?

src/test/regress/expected/isolation_add_node_vs_reference_table_operations.out
```diff
@@ -5,7 +5,7 @@ create_reference_table
 
                
 step s2-load-metadata-cache: 
-	COPY test_reference_table FROM PROGRAM 'echo "1\n2\n3\n4\n5"';
+	COPY test_reference_table FROM PROGRAM 'echo 1 && echo 2 && echo 3 && echo 4 && echo 5';
```
pykello: These files also have similar echos which have been missed from the current PR:
- isolation_create_distributed_table
- isolation_master_append_table
- isolation_master_apply_delete

src/test/regress/sql/multi_utility_statements.sql
lithp: Addressed by 9436138

src/test/regress/expected/isolation_add_node_vs_reference_table_operations.out
pykello: What do you think about changing the following files?

    isolation_create_distributed_table
    isolation_master_append_table
    isolation_master_apply_delete

pykello: Ok, I did this. Please take a look.

pykello:
@lithp why are the tests failing for this branch? Otherwise, looks good to me once the failures are resolved.

PR #2023, opened 2018-02-23
Do not use new connection in table size functions

### LINKED ISSUE #2004: Table size functions open a new connection on every call
Users with a lot of distributed tables may experience issues when running a query like `SELECT logicalrelid, pg_size_pretty(citus_total_relation_size(logicalrelid)) FROM pg_dist_partition` because the size functions open up a new connection on every call.

`DistributedTableSize` uses a FORCE_NEW_CONNECTION flag, meaning it always opens a new connection. We probably just need to remove that flag.

marcocitus:
DESCRIPTION: Fixes a bug that could cause table size functions to open a new connection on every call

Fixes #2004

### DIFF
--- a/src/backend/distributed/master/master_metadata_utility.c
+++ b/src/backend/distributed/master/master_metadata_utility.c
@@ -210,7 +210,7 @@ DistributedTableSizeOnWorker(WorkerNode *workerNode, Oid relationId, char *sizeQ
 	char *tableSizeString;
 	uint64 tableSize = 0;
 	MultiConnection *connection = NULL;
-	uint32 connectionFlag = FORCE_NEW_CONNECTION;
+	uint32 connectionFlag = 0;
 	PGresult *result = NULL;
 	int queryResult = 0;
 	List *sizeList = NIL;


onderkalaci:
Shall we add `backport` label to this PR?

marcocitus:
> Shall we add backport label to this PR?

Well, it's not a very critical bug, but it's small enough to be safe to backport.

onderkalaci:
Hmm, this seems to break the following where we try to get multiple relation sizes at one command:
```SQL
select pg_size_pretty(
citus_relation_size('public.partition_03_13_2018')+
citus_relation_size('public.partition_03_14_2018')+
citus_relation_size('public.partition_03_15_2018')+
citus_relation_size('public.partition_03_16_2018'));
```
Or even the following query returns the same error
```
 SELECT logicalrelid, pg_size_pretty(citus_total_relation_size(logicalrelid)) FROM pg_dist_partition;
```

I haven't thought in detail, but, should we force a new connection as we used to do, but, this time close it once we're done? Does that solve #2004? I guess the actual problem there was the total number of concurrently open connections, not the total number of connections opened?

```C
-    uint32 connectionFlag = 0;
+    uint32 connectionFlag = FORCE_NEW_CONNECTION;
     PGresult *result = NULL;
     int queryResult = 0;
     List *sizeList = NIL;
@@ -222,6 +222,7 @@ DistributedTableSizeOnWorker(WorkerNode *workerNode, Oid relationId, char *sizeQ
                                                            sizeQuery);
 
     connection = GetNodeConnection(connectionFlag, workerNodeName, workerNodePort);
     queryResult = ExecuteOptionalRemoteCommand(connection, tableSizeQuery->data, &result);
 
     if (queryResult != 0)
@@ -235,6 +236,8 @@ DistributedTableSizeOnWorker(WorkerNode *workerNode, Oid relationId, char *sizeQ
     tableSizeString = tableSizeStringInfo->data;
     tableSize = atol(tableSizeString);
 
+    CloseConnection(connection);
+
     return tableSize;
 }
```

onderkalaci:
> No, it's just missing a call to ForgetResults(connection);

Yes, sure makes sense

Issue #520, opened 2016-05-13
EXPLAIN errors out with "bogus varattno" when using avg() + order by


marcocitus:
Error message when using avg + order by:

```
postgres=# EXPLAIN SELECT avg(review_rating) FROM customer_reviews ORDER BY 1 DESC;
ERROR:  bogus varattno for OUTER_VAR var: 2
```

Regular average works fine:

```
postgres=# EXPLAIN SELECT avg(review_rating) FROM customer_reviews;
                                                   QUERY PLAN                                                    
-----------------------------------------------------------------------------------------------------------------
 Distributed Query into pg_merge_job_0042
   Executor: Real-Time
   Task Count: 16
   Tasks Shown: One of 16
   ->  Task
         Node: host=10.192.0.174 port=5432 dbname=postgres
         ->  Aggregate  (cost=1616.65..1616.66 rows=1 width=4)
               ->  Seq Scan on customer_reviews_102104 customer_reviews  (cost=0.00..1437.10 rows=35910 width=4)
 Master Query
   ->  Aggregate  (cost=0.01..0.02 rows=1 width=0)
         ->  Seq Scan on pg_merge_job_0042  (cost=0.00..0.00 rows=0 width=0)
(11 rows)
```

ozgune:
@marcocitus -- how complex do you think the fix is?

PR #2024, opened 2018-02-23
Fix worker query when order by aggregate is used

### LINKED ISSUE #2021: Citus may send wrong ORDER BYs if the ORDER BY contains an aggragete
This seems to a be an old bug.

Since Citus mutates aggregates, `ORDER BY`s in the aggregates might be a mutated one, which could easily lead to incorrect outputs.

```SQL
SELECT user_id,
       avg(value_1)
FROM users_table
GROUP BY user_id
ORDER BY avg(value_1)
LIMIT 2;
```

Note that worker query sorts on a different aggragete:
```SQL
SELECT user_id,
       sum(value_1) AS AVG,
       count(value_1) AS AVG
FROM users_table_102039 users_table
WHERE TRUE
GROUP BY user_id
ORDER BY (sum(value_1)) USING OPERATOR(pg_catalog.<) NULLS LAST
LIMIT '2'::bigint
```

Note that this bites us in #2008 in the following cases:

Master query:
```SQL
SELECT
	user_id, rank() OVER (PARTITION BY user_id ORDER BY avg(value_3))
FROM
	users_table
GROUP BY
	user_id
LIMIT 5;
```

Worker query, see `ORDER BY` again:
```SQL
SELECT worker_column_1 AS user_id,
       rank() OVER (PARTITION BY worker_column_1
                    ORDER BY (sum(worker_column_2))) AS rank,
                   sum(worker_column_2) AS worker_column_3,
                   count(worker_column_2) AS worker_column_4
FROM
  (SELECT users_table.user_id AS worker_column_1,
          users_table.value_3 AS worker_column_2
   FROM public.users_table_102039 users_table) worker_subquery
WHERE TRUE
GROUP BY worker_column_1
```

mtuncer:
DESCRIPTION : Fix order by on worker query when aggregate is used in order by

We push down order by to worker query when limit is specified
(with some other additional checks). If the query has an expression
on an aggregate or avg aggregate by itself, and there is an order
by on this particular target we may send wrong order by to worker
query with potential to affect query result.

The fix creates a auxilary target entry in the worker query and
uses that target entry for sorting.

Fixes #2021


mtuncer:
no it does not fix #520

mtuncer:
we should backport this too

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -3523,19 +3544,136 @@ WorkerSortClauseList(MultiExtendedOp *originalOpNode,
...
+ * to have 2 target entries, sum() and count(), if we have order by and limit
+ * on this target entry we want to (1) use correct target entry (avg), and (2)
+ * point to this particular target entry from sort clauses.
+ *
+ * Note the function modifies workerExtendedOpNodes targetList, contents
+ * of SortGroupClause items in sortClauseList, and nextSortGroupRefIndex.
+ */
+static void
+AdjustWorkerSortClausesAndTargetList(MultiExtendedOp *workerExtendedOpNode,
+									 MultiExtendedOp *originalOpNode,
```
onderkalaci: Could we pass `originalOpNode -> targetList ` instead of passing all?

> + * Note the function modifies workerExtendedOpNodes targetList, contents
+ * of SortGroupClause items in sortClauseList, and nextSortGroupRefIndex.

Hmm, also should we pass `workerExtendedOpNode->targetList` and return the modified list? I think it's acceptable to modify `nextSortGroupRefIndex` and elements of `sortClauseList `. But, I feel like it'd be more readable if we return the new targetlist from the function.

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -2026,6 +2038,15 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
 															groupedByDisjointPartitionColumn);
 		workerExtendedOpNode->sortClauseList =
 			WorkerSortClauseList(originalOpNode, groupedByDisjointPartitionColumn);
+		sortClauseList = list_concat(sortClauseList,
```
onderkalaci: Why do we need `sortClauseList ` at all? can't we operate solely on `workerExtendedOpNode->sortClauseList`?

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -3523,19 +3544,136 @@ WorkerSortClauseList(MultiExtendedOp *originalOpNode,
...
+		if (found)
+		{
+			sgClause->tleSortGroupRef = (*newIndexPtr);
+		}
+		else
+		{
+			TargetEntry *newTargetEntry = copyObject(targetEntry);
+			newTargetEntry->ressortgroupref = *nextSortGroupRefIndex;
+			sgClause->tleSortGroupRef = *nextSortGroupRefIndex;
+			newTargetEntry->resjunk = false;
```
onderkalaci: Hmm, I was thinking that we'd use `resjunk = true`. Do we actually need to pull that column from the worker? Aren't we already using the correct target entries on the coordinator?

src/test/regress/sql/multi_behavioral_analytics_basics.sql
```diff
@@ -476,3 +476,87 @@ FROM
...
+ORDER BY avg(value_1) DESC
+LIMIT 5;
+
+SELECT user_id, avg(value_1)
+FROM users_table
+GROUP BY user_id
+ORDER BY avg(value_1) DESC
+LIMIT 1;
+
+SELECT user_id, avg(value_1) + 1
```
onderkalaci: Maybe some tests with (i) target list doesn't contain the aggregate (ii) target list and order by contains different aggregates  (iii) Same aggragate with different expression `SELECT user_id, avg(value_1) FROM users_table GROUP BY user_id ORDER BY avg(value_1) * user_id DESC LIMIT 5;`

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -3523,19 +3544,136 @@ WorkerSortClauseList(MultiExtendedOp *originalOpNode,
...
+		 * on a single aggregate, or expression containing multiple aggregates.
+		 * Worker query mutates these target entries to have a naked target entry
+		 * per aggregate function. We want to use original target entries if this
+		 * the case.
+		 * If the original target expression is an avg aggref, we also want to use
+		 * original target entry.
+		 */
+		if (!IsA(targetExpr, Aggref))
+		{
+			useOriginalTargetEntry = true;
```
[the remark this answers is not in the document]
onderkalaci: Also, I guess the name should be `createNewTargetEntry` or such?

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -3523,19 +3544,136 @@ WorkerSortClauseList(MultiExtendedOp *originalOpNode,
...
+		 * just use that reference, otherwise create a new target entry in worker
+		 * target list.
+		 */
+		if (found)
+		{
+			sgClause->tleSortGroupRef = (*newIndexPtr);
+		}
+		else
+		{
+			TargetEntry *newTargetEntry = copyObject(targetEntry);
```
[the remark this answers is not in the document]
mtuncer: We will need everything inside the original query's target entry. Even if we make a new target entry, we would have to copy almost everything.

src/backend/distributed/planner/multi_logical_optimizer.c
mtuncer: I was thinking that too, but master query did not play nice with it.  It expected to see this column in worker query results.  Perhaps we should fix this altogether.

src/backend/distributed/planner/multi_logical_optimizer.c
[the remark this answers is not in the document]
mtuncer: spaces --> done

src/backend/distributed/planner/multi_logical_optimizer.c
mtuncer: It turned out to be easy fix in multi_master_planner

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -3494,8 +3515,8 @@ WorkerSortClauseList(MultiExtendedOp *originalOpNode,
 					 bool groupedByDisjointPartitionColumn)
 {
 	List *workerSortClauseList = NIL;
-	List *groupClauseList = originalOpNode->groupClauseList;
-	List *sortClauseList = originalOpNode->sortClauseList;
+	List *groupClauseList = copyObject(originalOpNode->groupClauseList);
```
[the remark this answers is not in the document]
mtuncer: right, you can't have aggs in group clause list

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -172,6 +178,11 @@ static Node * WorkerLimitCount(MultiExtendedOp *originalOpNode,
 							   bool groupedByDisjointPartitionColumn);
 static List * WorkerSortClauseList(MultiExtendedOp *originalOpNode,
 								   bool groupedByDisjointPartitionColumn);
+static void AdjustWorkerSortClausesAndTargetList(MultiExtendedOp *workerExtendedOpNode,
+												 MultiExtendedOp *originalOpNode,
+												 List *sortClauseList,
+												 Index
```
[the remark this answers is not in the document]
mtuncer: done

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -3523,19 +3544,136 @@ WorkerSortClauseList(MultiExtendedOp *originalOpNode,
...
 	return workerSortClauseList;
 }
 
 
+/*
+ * AdjustWorkerSortClausesAndTargetList goes over provided sort clause lists and
+ * make sure they still reference to a correct target entry.
+ *
+ * The function is required because we change target entry if it contains an
+ * aggregate operation, AVG operation specific to this case. A problem arises
```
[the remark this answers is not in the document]
mtuncer: fixed

src/backend/distributed/planner/multi_logical_optimizer.c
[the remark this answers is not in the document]
mtuncer: variable was already renamed.

However, reducing the logical checks and ditching the boolean flag made code less readable.

src/backend/distributed/planner/multi_logical_optimizer.c
mtuncer: we are modifying all passed arguments. Selectively returning one of them looked weird to me.

src/backend/distributed/planner/multi_logical_optimizer.c
mtuncer: it will come in handy when we need this in window function pushdown (next pr)

src/backend/distributed/planner/multi_logical_optimizer.c
onderkalaci: But, modifying 3 parameters in a single function seems weird as well. I don't think we have any other code that does it. 

Even if we have some functions which *mutates* the input, their name included *mutate*.

And, if we keep the API like this, how are we going to use it with window functions?

src/backend/distributed/planner/multi_logical_optimizer.c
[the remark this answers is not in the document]
onderkalaci: We'd only need to copy the `targetExpr`?
```C
int resno = workerTargetList->length + 1;
bool resjunk = true;
Expr *newTargetExpr = copyObject(targetExpr);
TargetEntry *newTargetEntry = makeTargetEntry(newTargetExpr, resno, "not_important_since_resjunk_resno", resjunk);

File: src/backend/distributed/planner/multi_logical_optimizer.c (hunk 1 above)
Comment (onderkalaci) [reply src/backend/distributed/planner/multi_logical_optimizer.c]: I propose the following API, which seems to be easliy applicable to window functions as well:

```C
static List *
AddRequiredSortExpressionsToTargetList(List *queryTargetList, List *sortGroupClauseList, Index *nextSortGroupRefIndex)
{
    /* 
      * modify elementsOf sortList as needed, 
       * increment nextSortGroupRefIndex as needed, 
      *return new targetList*/
     return newTargetListWithNewElements
}
```

Here:

```C
workerExtendedOpNode->targetList = AddRequiredSortExpressionsToTargetList (workerExtendedOpNode->targetList, sortClauseList, &refIndex)
```


And on window functions:
```C
targetList = AddRequiredSortExpressionsToTargetList(targetList, window->partitionClause, &refIndex);
targetList = AddRequiredSortExpressionsToTargetList(targetList, window->orderClause, &refIndex);
```

src/test/regress/sql/multi_behavioral_analytics_basics.sql
mtuncer: added some, I started to think we should also add explain tests

src/backend/distributed/planner/multi_logical_optimizer.c
mtuncer: We need to access target list of original extended op node since we are modifying the worker target list up to this point.

mtuncer: per our discussion
1 - function will generate and return list of new target entries to be appended to worker target list
2 - it will be possible to have repetitions of the same aggregate if it is also used in window function order by clause

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -3523,19 +3544,136 @@ WorkerSortClauseList(MultiExtendedOp *originalOpNode,
...
+								 &hashCtl,
+								 HASH_ELEM | HASH_BLOBS);
+
+	Assert(sortClauseList != NIL);
+
+	foreach(sortClauseCell, sortClauseList)
+	{
+		SortGroupClause *sgClause = (SortGroupClause *) lfirst(sortClauseCell);
+		TargetEntry *targetEntry = get_sortgroupclause_tle(sgClause, originalTargetList);
+		bool containsAggregate = contain_agg_clause((Node *) targetEntry->expr);
```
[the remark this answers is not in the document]
mtuncer: ok

src/test/regress/expected/multi_behavioral_analytics_basics.out
```diff
@@ -522,3 +522,210 @@ SELECT count(*), count(DISTINCT user_id), avg(user_id), avg(value_1_agg) FROM ag
      6 |     6 | 3.5000000000000000 | 0.16666666666666666667
 (1 row)
 
+-- test order by when pushing down avg
+SELECT user_id, avg(value_1)
```
onderkalaci: Minor: this test source is mostly intended for pushdownable `INSERT ... SELECT` queries. It'd be good if we can move them to somewhere else such as `multi_complex_expressions.sql`?

src/test/regress/sql/multi_behavioral_analytics_basics.sql
```diff
@@ -476,3 +476,117 @@ FROM
...
+ORDER BY 2  DESC
+LIMIT 1;
+
+SELECT user_id, sum(value_1) + sum(value_2)
+FROM users_table
+GROUP BY user_id
+ORDER BY sum(value_2)  DESC
+LIMIT 1;
+
+SELECT user_id, (100 / max(value_1))
```
onderkalaci: I think we could consider adding one test with multiple target entries in the ORDER BY that are modified. We seem to have only one for all tests?

Also, maybe a test with a join such that each element belonging to a different table?

onderkalaci: One more test idea: Aggregate inside distinct: `select distinct user_id, (avg (value_1)) FROM users_table GROUP BY user_id ORDER BY 2 DESC, 1 DESC LIMIT 10;`
(Is there a way to pushdown `DISTINCT` with aggragates)

onderkalaci:
> DESCRIPTION : Fix order by on worker query when avg() is used in order by

It looks like the description (and the PR description) doesn't mention about the cases you've fixed after `avg` fix

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -2021,13 +2024,24 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
...
 	{
+		List *newTargetEntryListForSortClauses = NIL;
+
 		/* if we can push down the limit, also set related fields */
 		workerExtendedOpNode->limitCount = WorkerLimitCount(originalOpNode,
 															groupedByDisjointPartitionColumn);
 		workerExtendedOpNode->sortClauseList =
 			WorkerSortClauseList(originalOpNode, groupedByDisjointPartitionColumn);
+
+		newTargetEntryListForSortClauses = GenerateNewTargetEntriesForSortClauses(
```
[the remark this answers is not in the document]
mtuncer: targetProjectionNumber is used to determine value of ```resno``` field.  Having an out variable can be worked out as you suggested but in any case we need to pass in something so that we can use correct resno value.

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -3523,19 +3537,101 @@ WorkerSortClauseList(MultiExtendedOp *originalOpNode,
...
+			{
+				createNewTargetEntry = true;
+			}
+		}
+
+		if (createNewTargetEntry)
+		{
+			bool resJunk = true;
+			AttrNumber nextResNo = (*targetProjectionNumber);
+			TargetEntry *newTargetEntry = makeTargetEntry(targetEntry->expr, nextResNo,
```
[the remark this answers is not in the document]
mtuncer: used copyObject(targetExpr)
we should be okay with resname.

src/test/regress/expected/multi_behavioral_analytics_basics.out
mtuncer: done

src/test/regress/expected/multi_select_distinct.out
```diff
@@ -969,7 +969,7 @@ EXPLAIN (COSTS FALSE)
                                              Node: host=localhost port=57637 dbname=regression
                                              ->  Limit
                                                    ->  Sort
-                                                         Sort Key: (array_agg(l_linenumber))
+                                                         Sort Key: (array_length(array_agg(l_linenumber), 1))
```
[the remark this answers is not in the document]
mtuncer: done

src/test/regress/sql/multi_behavioral_analytics_basics.sql
mtuncer: added

mtuncer:
I am the most likely person to make a change that would break that 😬

PR #2028, opened 2018-02-27
Fix regression for changes on REL_10_STABLE

metdos:


marcocitus:
Might make sense to temporarily add an alternative output file (`multi_index_statements_0.out`)

metdos:
> Might make sense to temporarily add an alternative output file (multi_index_statements_0.out)

Good idea!

metdos:
@marcocitus, added the file and opened an issue to keep track of it.

PR #2030, opened 2018-02-27
Add support for bool and bit aggregates

marcocitus:
DESCRIPTION: Adds support for bool_ and bit_ aggregates

bool_and, bool_or, bit_and, bit_or, and every are handled in the same way as sum; the aggregate itself is the combine function.


src/include/distributed/multi_logical_optimizer.h
```diff
@@ -105,7 +105,8 @@ static const char *const AggregateNames[] = {
 	"invalid", "avg", "min", "max",
 	"sum", "count", "array_agg",
 	"jsonb_agg", "jsonb_object_agg",
-	"json_agg", "json_object_agg"
+	"json_agg", "json_object_agg",
+	"bit_and", "bit_or", "bool_and", "bool_or", "every"
 };
```
velioglu: Should we add these aggregates to the `AggregateType` enum to not to break the rule given in the second part of the comment above ?

marcocitus:
> Should we add bit and bool aggregates as a comment to the line https://github.com/citusdata/citus/blob/bool_agg/src/backend/distributed/planner/multi_logical_optimizer.c#L1667 ?

Seems like that'll become unsustainable. The json aggregates are already missing. Removed the list.

src/include/distributed/multi_logical_optimizer.h
marcocitus: Makes sense, added them.

PR #2031, opened 2018-02-28
Improve error handling on failures

### LINKED ISSUE #1926: Citus may not behave properly on the first command execution after a worker is stopped

onderkalaci:
DESCRIPTION: Fixes a bug that could lead to unexpected results after failures on the worker nodes

Fixes #1926. 

Commit message which is good to post here as well: 
```
    This commit checks the connection status right after any IO happens
    on the socket.
    
    This is necessary since before this commit we didn't pass any information
    to the higher level functions whether we're done with the connection
    (e.g., no IO required anymore) or an error happened during the IO.
```

### DIFF
--- a/src/backend/distributed/connection/remote_commands.c
+++ b/src/backend/distributed/connection/remote_commands.c
@@ -65,24 +65,7 @@ IsResponseOK(PGresult *result)
 void
 ForgetResults(MultiConnection *connection)
 {
-	while (true)
-	{
-		PGresult *result = NULL;
-		const bool dontRaiseErrors = false;
-
-		result = GetRemoteCommandResult(connection, dontRaiseErrors);
-		if (result == NULL)
-		{
-			break;
-		}
-		if (PQresultStatus(result) == PGRES_COPY_IN)
-		{
-			PQputCopyEnd(connection->pgConn, NULL);
-
-			/* TODO: mark transaction as failed, once we can. */
-		}
-		PQclear(result);
-	}
+	ClearResults(connection, false);
 }
 
 
--- a/src/backend/distributed/connection/remote_commands.c
+++ b/src/backend/distributed/connection/remote_commands.c
@@ -106,6 +106,14 @@ ClearResults(MultiConnection *connection, bool raiseErrors)
 			MarkRemoteTransactionFailed(connection, raiseErrors);
 
 			success = false;
+
+			/* an error happened, there is nothing we can do more */
+			if (PQresultStatus(result) == PGRES_FATAL_ERROR)
+			{
+				PQclear(result);
+
+				break;
+			}
 		}
 
 		PQclear(result);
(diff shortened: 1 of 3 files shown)


src/test/regress/expected/multi_extension.out
```diff
@@ -369,8 +369,8 @@ SELECT create_distributed_table('t1', 'a');
 WARNING:  function assign_distributed_transaction_id(integer, integer, unknown) does not exist
 HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
 CONTEXT:  while executing command on localhost:57637
-ERROR:  current transaction is aborted, commands ignored until end of transaction block
-CONTEXT:  while executing command on localhost:57637
+ERROR:  connection error: localhost:57637
```
onderkalaci: Although it seems like a regression, this error message is a more accurate one.

src/backend/distributed/connection/remote_commands.c
```diff
@@ -550,6 +541,12 @@ GetRemoteCommandResult(MultiConnection *connection, bool raiseInterrupts)
 
 	if (!FinishConnectionIO(connection, raiseInterrupts))
 	{
+		/* some error(s) happened while doing the I/O, signal the callers */
+		if (PQstatus(pgConn) == CONNECTION_BAD)
+		{
+			return PQmakeEmptyPGresult(pgConn, PGRES_FATAL_ERROR);
```
marcocitus: Could we try setting the error message from `PQerrorMessage(pgConn)` here?

onderkalaci: `PQmakeEmptyPGresult ` seems to do that, so I skipped doing that

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -1446,6 +1447,12 @@ StoreQueryResult(CitusScanState *scanState, MultiConnection *connection,
 
 			commandFailed = true;
 
+			/* an error happened, there is nothing we can do more */
+			if (fatalError)
```
marcocitus: I don't think the boolean adds much, could just put `resultStatus == PGRES_FATAL_ERROR)` here

marcocitus:
@lithp this would be a good case to test in the automated failure tests

PR #2035, opened 2018-03-02
Fix incorrect limit pushdown when distinct clause is not superset of group by

### LINKED ISSUE #2034: limits are pushed down to workers when distinct clause does not contain all group by columns
Following query pushes down order by and limit clauses to workers.

```sql
SELECT
  DISTINCT ON (sum(value_1))
  SUM(value_1),
  user_id
FROM
  users_table
GROUP BY
   user_id, value_1, value_2
ORDER BY
  (sum(value_1)), 2
LIMIT 10;
```
Since distinct on comes with additional ordering/unique semantics, we should not be pushing down  limit (and order by) here.

Similar issue could be observed with window functions too. We should not push down limit when there are window function too

```sql
SELECT
  DISTINCT ON (SUM(value_1) OVER (PARTITION BY user_id)) user_id, SUM(value_2) OVER (PARTITION BY user_id)
FROM
  users_table
GROUP BY
  user_id, value_1, value_2
HAVING count(*) = 1
ORDER BY
  (SUM(value_1) OVER (PARTITION BY user_id)) , 2 DESC, 1
LIMIT
  100;
```

mtuncer:
DESCRIPTION: bugfix on incorrect limit pushdown when distinct (on) clause is not superset of group by clause

Pushing down limit and order by into workers may produce
wrong output when distinct on() clause has expressions,
aggregates, or window functions.

This checking allows pushing down of limits only if
distinct clause is a superset of group by clause. i.e. it contains  all clauses in group by.

Fixes #2034

### DIFF
--- a/src/backend/distributed/planner/multi_logical_optimizer.c
+++ b/src/backend/distributed/planner/multi_logical_optimizer.c
@@ -1860,7 +1860,9 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
 		palloc0(sizeof(WorkerAggregateWalkerContext));
 	Index nextSortGroupRefIndex = 0;
 	bool queryHasAggregates = false;
-	bool enableLimitPushdown = true;
+	bool distinctClauseSupersetofGroupClause = false;
+	bool distinctPreventsLimitPushdown = false;
+	bool createdNewGroupByClause = false;
 	bool hasNonPartitionColumnDistinctAgg = false;
 	bool repartitionSubquery = false;
 
@@ -1949,11 +1951,7 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
 				groupClauseList = lappend(groupClauseList, groupByClause);
 				nextSortGroupRefIndex++;
 
-				/*
-				 * If we introduce new columns accompanied by a new group by clause,
-				 * than pushing down limits will cause incorrect results.
-				 */
-				enableLimitPushdown = false;
+				createdNewGroupByClause = true;
 			}
 
 			if (newTargetEntry->resname == NULL)
@@ -2013,11 +2011,7 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
 				groupClauseList = lappend(groupClauseList, groupByClause);
 				nextSortGroupRefIndex++;
 
-				/*
-				 * If we introduce new columns accompanied by a new group by clause,
-				 * than pushing down limits will cause incorrect results.
-				 */
-				enableLimitPushdown = false;
+				createdNewGroupByClause = true;
 			}
 
 			newTargetEntryList = lappend(newTargetEntryList, newTargetEntry);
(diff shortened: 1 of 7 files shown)


src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -2031,7 +2032,30 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
...
 	workerExtendedOpNode->windowClause = originalOpNode->windowClause;
 
-	if (!queryHasAggregates)
+	/*
+	 * we are only pushing down distinct on, and limit (along with order by)
+	 * if it distinct on clause is on bare column. If it has expression, aggregates
+	 * or function calls, they are evaluated at worker query and returned to master
+	 * to perform necessary distinct on/order by/limit operations.
+	 */
+	if (originalOpNode->hasDistinctOn)
```
onderkalaci: Looks like we've wrong results for queries without an expression on the `DISTINCT`
```SQL
-- wrong results without expression in the DISTINCT
SELECT DISTINCT ON (value_2) value_2, user_id, avg(value_1)
FROM 
	users_table
GROUP BY 
	user_id, value_1, value_2
ORDER BY 
	1 DESC, 3 DESC, 2 DESC
LIMIT 10;
```

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -2031,7 +2031,21 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
 	workerExtendedOpNode->hasWindowFuncs = originalOpNode->hasWindowFuncs;
 	workerExtendedOpNode->windowClause = originalOpNode->windowClause;
 
-	if (!queryHasAggregates)
+	/*
+	 * we are only pushing down distinct on, and limit (along with order by)
+	 * if it distinct on clause is on bare column. If it has expression, aggregates
```
onderkalaci: `if it` -> `if`

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -2031,7 +2031,21 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
...
+	 * or function calls, they are evaluated at worker query and returned to master
+	 * to perform necessary distinct on/order by/limit operations.
+	 */
+	if (enableLimitPushdown && originalOpNode->distinctClause && groupClauseList != NIL)
```
onderkalaci: `groupClauseList != NIL` could we remove that since it is already checked in `IsGroupBySubsetOfDistinct` ?

mtuncer: ```IsGroupBySubsetOfDistinct()``` returns false if group by clause is empty which causes setting orderby/limit pushdown flags to false. We do not want that.

onderkalaci:
> Fix incorrect limit pushdown when distinct on has expressions

Title seems stale

> DESCRIPTION: bugfix on incorrect limit pushdown when distinct (on) clause differs from group by clause

The description seems too broad. Can't we use `is superset ` instead of `differs from` or such

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -2030,16 +2031,39 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
...
 	workerExtendedOpNode->windowClause = originalOpNode->windowClause;
+	workerExtendedOpNode->groupClauseList = groupClauseList;
+
+	/*
+	 * We are only pushing down distinct on, and limit (along with order by)
+	 * if distinct on clause is superset of group by clause. If there is no
+	 * grouping we are good to pushdown distinct (and orderby/limit). Query
+	 * result depends on order by expression.
+	 */
+	if (shouldPushdownOrderByLimit && originalOpNode->distinctClause &&
```
onderkalaci: Well, I think I found what disturbs me here.

We're creating a cyclic dependency among `limit pushdown` and `distinct/distinct on`. We initially look `shouldPushdownOrderByLimit ` and `DISTINCT` to decide on the new value of `shouldPushdownOrderByLimit `. Later, we use `shouldPushdownOrderByLimit ` to decide on `shouldPushdownDistinct `. 

What about if we break the dependency and go only in one direction that we always evaluate `DISTINCT/DISTINCT ON` and modify `shouldPushdownOrderByLimit` whenever necessary. 

Such as this: 

https://github.com/citusdata/citus/compare/fix_distinct_on_agg...fix_distinct_on_agg_onder?expand=1

(Note that the branch is NOT complete, I mostly wanted to share my take on the issue. We should probably iterate on the correctness and comments etc once more)

Issue #2039, opened 2018-03-05
Copying nested jsonb fields crashes


metdos:
Steps to reproduce;

Create table and types.
```sql
CREATE TYPE complex AS (r double precision, i double precision);
SELECT run_command_on_workers('CREATE TYPE complex AS (r double precision, i double precision)');
CREATE TABLE test_table (key int, occurred_at timestamp DEFAULT now(), value_1 jsonb, value_2 text[], value_3 int4range, value_4 complex NOT NULL);
SELECT create_distributed_table('test_table', 'key');
```

Put this line into a file such as `scale_test_data.csv`
```
27109	2018-03-05 09:35:19.32211	{"f1": 27109, "f2": 80313, "f3": {"f1": 56369, "f2": 8692, "f3": 189763}, "f4": {"f1": 56349, "f2": 7692, "f3": 1924739}}	{27109,80313,56359,8192,"Onder Kalaci","CitusData Rocks"}	[27109,107422)	(27109,8192)
```

Run copy and it will crash;
```
\COPY test_table FROM 'scale_test_data.csv'
```

This bug seems related the optimization in #1965. Setting `citus.skip_jsonb_validation_in_copy` to `false` fixes the problem.

Another easier way to replicate problem is to run scale tests in the test automation framework;
`fab run.pgbench_tests:scale_test.ini`
Reference: https://github.com/citusdata/test-automation#-running-scale-tests

Here is the call stack;

```
getJsonbOffset() at jsonb_util.c:133 0x10160310d	
JsonbIteratorNext() at jsonb_util.c:832 0x101603803	
JsonbToCStringWorker() at jsonb.c:475 0x1015fca40	
JsonbToCString() at jsonb.c:431 0x1015fc6b5	
jsonb_out() at jsonb.c:137 0x1015fc682	
FunctionCall1Coll() at fmgr.c:1,037 0x101713b6f	
OutputFunctionCall() at fmgr.c:1,669 0x101715b11	
AppendCopyRowData() at multi_copy.c:1,704 0x10208abaa	
CitusCopyDestReceiverReceive() at multi_copy.c:2,454 0x10208b741	
CopyToExistingShards() at multi_copy.c:523 0x10208998b	
CitusCopyFrom() at multi_copy.c:211 0x102088fbb	
ProcessCopyStmt() at multi_utility.c:831 0x1020a163d	
multi_ProcessUtility() at multi_utility.c:316 0x1020a07b8	
[...]

PR #2041, opened 2018-03-06
Change default to false for citus.skip_jsonb_validation_in_copy

metdos:
A workaround for #2039. Until we implement a proper fix, we turn off the optimization by default.

### DIFF
--- a/src/backend/distributed/shared_library_init.c
+++ b/src/backend/distributed/shared_library_init.c
@@ -519,9 +519,9 @@ RegisterCitusConfigVariables(void)
 					 "be higher. This setting does not apply if the input format is "
 					 "binary."),
 		&SkipJsonbValidationInCopy,
-		true,
+		false,
 		PGC_USERSET,
-		0,
+		GUC_NO_SHOW_ALL,
 		NULL, NULL, NULL);
 
 	DefineCustomIntVariable(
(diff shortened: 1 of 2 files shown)


src/backend/distributed/shared_library_init.c
```diff
@@ -519,7 +519,7 @@ RegisterCitusConfigVariables(void)
 					 "be higher. This setting does not apply if the input format is "
 					 "binary."),
 		&SkipJsonbValidationInCopy,
-		true,
+		false,
```
onderkalaci: I'd change such that the value is not visible users via `GUC_NO_SHOW_ALL `. What do you think?

I think we don't have anything in the citus docs related to `citus. skip_jsonb_validation_in_copy`. Could you double check? I guess we don't want to document it before we fix the issue.

metdos: This makes sense. I updated the PR. I also could not find a reference in the docs as expected.

src/backend/distributed/shared_library_init.c
```diff
@@ -519,8 +519,8 @@ RegisterCitusConfigVariables(void)
 					 "be higher. This setting does not apply if the input format is "
 					 "binary."),
 		&SkipJsonbValidationInCopy,
-		true,
-		PGC_USERSET,
+		false,
```
onderkalaci: Hmm, why do we remove `PGC_USERSET`, it seems a separate value. I guess it should be the same as the above GUC?

```
	DefineCustomBoolVariable(
		"citus.enable_router_execution",
		gettext_noop("Enables router execution"),
		NULL,
		&EnableRouterExecution,
		true,
		**PGC_USERSET,
		GUC_NO_SHOW_ALL,**
		NULL, NULL, NULL);
```

metdos: stupid mistake, fixed!

PR #2044, opened 2018-03-06
[WIP] Automated Failure Testing

lithp:
- Create a daemon which backgrounds gdb and allows us to send commands to it.

### DIFF
--- a/src/test/regress/gdb.py
+++ b/src/test/regress/gdb.py
@@ -1,0 +1,169 @@
+#!/usr/bin/env python3
+
+'''
+A daemon which communicates with gdb.
+
+Sits on a domain socket. Accepts commands from it and forwards them to gdb
+
+TODO:
+- prevent there from being multiple clients?
+'''
+
+import enum
+import fcntl
+import functools
+import os
+import re
+import select
+import signal
+import socket
+import subprocess
+import sys
+
+def spawn_gdb():
+    process = subprocess.Popen(
+        ["gdb"],
+        stdin=subprocess.PIPE, stdout=subprocess.PIPE,
+        stderr=subprocess.STDOUT,  # redirect stderr -> stdout
+
+        #  these parameters ensure we're getting TextIOWrapper's and can call readline()
+        universal_newlines=True,
+        bufsize=1  # enables line-buffering
+    )
+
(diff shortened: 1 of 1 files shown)


src/backend/distributed/citus--7.3-3--7.3-4.sql
```diff
@@ -0,0 +1,78 @@
...
+/* helper functions for automated failure testing */
+
+CREATE FUNCTION citus.run_command(text)
+RETURNS void AS $$
+  use strict;
+  use warnings;
+  use IO::Socket;
+
+  my $command = $_[0];
+  my $sockpath = '/home/brian/Work/citus/src/test/regress/gdb.sock';
```
mtuncer: can you make this a relative path ?

lithp: I think eventually it'll need to be a hostname:port tuple, otherwise this won't work on Windows.

src/test/regress/specs/isolation_failure_testing.spec
```diff
@@ -0,0 +1,37 @@
...
+session "gdb"
+
+step "gdb-break"
+{
+  WITH pid AS (
+    SELECT pid FROM pg_stat_activity WHERE application_name = 'gdb-process'
+  ) SELECT citus.gdb_attach(pid.pid) FROM pid;
+
+  SELECT citus.run_command('!break NeedsDistributedPlanning');
+  SELECT citus.run_command('continue');
```
marcocitus: The API seems a bit messy at the moment. Would it make sense to do something like:
```
SELECT citus.gdb_init();
SELECT citus.gdb_break('NeedsDistributedPlanning');
SELECT citus.gdb_continue();
```

lithp: Yeah, 100%. This API isn't very user friendly. I haven't made it nicer yet for a couple reasons:
- I'm not sure what the right API is yet, I want to write some more tests before deciding what the right set of concepts is.
- These functions are a little hard to add and iterate on. I haven't figured out how to share code between the functions so I need to copy and paste the code for connecting to a socket and modify it in all of them whenever it changes. For now just having a `run_command` is easier to develop with.

PR #2052, opened 2018-03-13
Relax assertion on transaction abort/rollback to handle failures

### LINKED ISSUE #2050: Server crashes if node fails during insert into/select due to unnecessary assertion
It was shown up during 7.3 release tests. You can reproduce it by following the steps below.

```
-- tt1 is distributed on id column
create table tt1(id int, value_1 varchar(20));
insert into tt1 select x,x::text from generate_series(1,1000000) as f(x);
select create_distributed_table('tt1','id');

-- tt2 is distributed on value_1 column and data is inserted to it from
-- tt1 using coordinator insert into/select.
create table tt2(id int, value_1 varchar(20));
select create_distributed_table('tt2','value_1');
insert into tt2 select * from tt1; 
```

If you fail one of the worker nodes during the last command, server crashes with,

```
WARNING:  connection not open
WARNING:  failed to abort 1PC transaction "" on localhost:9700
ERROR:  server closed the connection unexpectedly
```

onderkalaci:
Fixes #2050 and #1966 

In case a failure happens when a transaction is rollbacked/aborted,
we used to hit an assertion for ensuring there is no pending
activity on the connection. However, that's not true after the
changes in #2031. Thus, we've replaced the assertion with a more
generic function call to consume any pending activity, if exists.

@velioglu can you do some more failure testing on this PR?

@marcocitus It looks like it'd be good to get this change to 7.3. Does that sound OK to you as well?

### DIFF
--- a/src/backend/distributed/transaction/remote_transaction.c
+++ b/src/backend/distributed/transaction/remote_transaction.c
@@ -418,34 +418,30 @@ void
 FinishRemoteTransactionAbort(MultiConnection *connection)
 {
 	RemoteTransaction *transaction = &connection->remoteTransaction;
-	PGresult *result = NULL;
-	const bool dontRaiseErrors = false;
-	const bool isNotCommit = false;
+	const bool raiseErrors = false;
 
-	result = GetRemoteCommandResult(connection, dontRaiseErrors);
-
-	if (!IsResponseOK(result))
+	if (transaction->transactionState == REMOTE_TRANS_2PC_ABORTING)
 	{
-		ReportResultError(connection, result, WARNING);
-		MarkRemoteTransactionFailed(connection, dontRaiseErrors);
+		PGresult *result = GetRemoteCommandResult(connection, raiseErrors);
+		if (!IsResponseOK(result))
+		{
+			const bool isCommit = false;
+			WarnAboutLeakedPreparedTransaction(connection, isCommit);
+		}
 
-		if (transaction->transactionState == REMOTE_TRANS_2PC_ABORTING)
-		{
-			WarnAboutLeakedPreparedTransaction(connection, isNotCommit);
-		}
-		else
-		{
-			ereport(WARNING,
-					(errmsg("failed to abort 1PC transaction \"%s\" on %s:%d",
-							transaction->preparedName, connection->hostname,
-							connection->port)));
-		}
+		PQclear(result);
 	}
 
-	PQclear(result);
-
-	result = GetRemoteCommandResult(connection, dontRaiseErrors);
-	Assert(!result);
+	/*
+	 * Try to consume results of any in-progress commands. In the 1PC case
+	 * this is also where we consume the result of the ROLLBACK.
+	 *
+	 * If we don't succeed the connection will be in a bad state, so we close it.
+	 */
+	if (!ClearResults(connection, raiseErrors))
+	{
+		ShutdownConnection(connection);
+	}
 
 	transaction->transactionState = REMOTE_TRANS_ABORTED;
 }


velioglu:
@onderkalaci It passes the failure tests.

sumedhpathak:
@onderkalaci @marcocitus Is it still the plan to make this part of 7.3? I thought we had already begun packaging?

onderkalaci:
> @onderkalaci @marcocitus Is it still the plan to make this part of 7.3? I thought we had already begun packaging?

After talking with @marcocitus it looks like the change is not important for 7.3, so we don't need to backport. 

Also, @marcocitus thinks a broader change in this function, so we might consider closing this PR in that case.

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -444,8 +444,8 @@ FinishRemoteTransactionAbort(MultiConnection *connection)
 
 	PQclear(result);
 
-	result = GetRemoteCommandResult(connection, dontRaiseErrors);
-	Assert(!result);
+	/* the transaction is already aborted, clear any pending activity if exists */
+	ForgetResults(connection);
```
marcocitus: This entire function should perhaps just be:

```c
if (!ClearResults(connection, false))
{
    ShutdownConnection(connection);
}
```

The code above tries to give some meaningful warning if the `ABORT` fails, but it might not actually be checking the result of the abort, but rather the result of the preceding command. In the 1PC case the an abort failure has 0 consequences, but we should probably close the connection because it is in a weird state. In the 2PC case we might need to wait until transaction recovery before the locks are released. 

We could keep the `WarnAboutLeakedPreparedTransaction` since we cannot be in this situation where we send `ROLLBACK PREPARED` while another command is still in progress.

maybe something like:

```c
bool raiseErrors = false;

if (transaction->transactionState == REMOTE_TRANS_2PC_ABORTING)
{
    PGresult *result = GetRemoteCommandResult(connection, raiseErrors);
    if (!IsResponseOK(result))
    {
        bool isCommit = false;
        WarnAboutLeakedPreparedTransaction(connection, isCommit);
    }

    PQclear(result);
}

if (!ClearResults(connection, raiseErrors))
{
    ShutdownConnection(connection);
}
```

(not sure if this checks for dead connections properly)

pykello: From what I understand, @marcocitus 's suggested solution makes sense, except that (1) we should probably add the `failed to abort 1PC transaction ...` warning somewhere. (2) Set `transactionState` to `REMOTE_TRANS_ABORTED`.

I tried it on dead connections, and it seems to work.

pykello:
@marcocitus please take a look. I addressed your feedback, and did some testing.

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -418,34 +418,24 @@ void
...
+	if (transaction->transactionState == REMOTE_TRANS_2PC_ABORTING)
 	{
-		ReportResultError(connection, result, WARNING);
-		MarkRemoteTransactionFailed(connection, dontRaiseErrors);
-
-		if (transaction->transactionState == REMOTE_TRANS_2PC_ABORTING)
+		PGresult *result = GetRemoteCommandResult(connection, dontRaiseErrors);
+		if (!IsResponseOK(result))
 		{
+			const bool isNotCommit = false;
```
marcocitus: While we're changing this, we should maybe use `isCommit`. These double negations are confusing.

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -418,34 +418,24 @@ void
 FinishRemoteTransactionAbort(MultiConnection *connection)
 {
 	RemoteTransaction *transaction = &connection->remoteTransaction;
-	PGresult *result = NULL;
 	const bool dontRaiseErrors = false;
```
marcocitus: small note: we're changing this to `raiseErrors` in https://github.com/citusdata/citus/pull/2123

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -418,34 +418,24 @@ void
...
-		}
-	}
 
-	PQclear(result);
+		PQclear(result);
+	}
 
-	result = GetRemoteCommandResult(connection, dontRaiseErrors);
-	Assert(!result);
+	if (!ClearResults(connection, dontRaiseErrors))
```
marcocitus: Maybe add a little comment, e.g.:
```c
/*
 * Try to consume results of any in-progress commands. In the 1PC case
 * this is also where we consume the result of the ROLLBACK.
 *
 * If we don't succeed the connection will be in a bad state, so we close it.
 */
```

PR #2060, opened 2018-03-15
Bump citus version to 7.4devel

velioglu:


metdos:
@velioglu why are we waiting for this PR?

velioglu:
I can rebase and merge it after a review.

configure
```diff
@@ -661,7 +661,6 @@ infodir
 docdir
 oldincludedir
 includedir
-runstatedir
```
metdos: I remember that we have talked about this and you said we removed this 'runstatedir' on purpose?

metdos: Apart from this question/clarification, it looks good 🚢

velioglu: Yes, it was emerged due to running autoconf on Ubuntu and should be removed before merging, so we are removing it with this PR.

jasonmp85:
Can you also update `citus_version.h` inside the `windows` folder?

Issue #2049, opened 2018-03-12
Document & refactor logical optimizer


onderkalaci:
Recently we've hit few bugs (#2021, #2034, #2017) in the optimizer/master planner. The motivation for opening this issue that we suspect there might be more hidden bugs like the ones we've found.

The goal is to:

1. Simplify the related code
     - Better variable names, smaller functions, more comments
2. Document what is being pushed down and what is pulled to the coordinator when the query involves
    - Aggregates, DISTINCT, DISTINCT ON, HAVING, COUNT DISTINCT, LIMIT - OFFSET - ORDER BY, Window functions, Group by on distribution key / or not, subqueries (mostly pushdownable ones)
3. Improve the optimizer
    - Some more refactoring based on the previous items
    - Apply missing optimizations if any
    - Fix any issues found

onderkalaci:
The main goal of this task is to simplify the logical optimizer for two reasons:
1. Reduce time to build new features
2. Improve the product quality 

Citus supports a variety of SQL features listed below**. The optimizer picks a different approach to plan/optimize these features. 

	1.  Aggragates in the target list
	2.  Expressions in the target list
	3. `DISTINCT` / `DISTINCT ON` that is pushdownable
	4. `DISTINCT` / `DISTINCT ON` that is not pushdownable
	5. `COUNT (DISTINCT )` that is pushdownable
	6. `COUNT (DISTINCT )` that is not pushdownable
	7.  Single table re-partition subqueries
	8. `HAVING` that is pushdownable
	9. `HAVING` that is NOT pushdownable
	10. `LIMIT` / `OFFSET` that is pushdownable
	11.`LIMIT` / `OFFSET` that is not pushdownable
	12. `ORDER BY` clause 
	13. `GROUP BY` clause
	14. Window functions that are pushdownable
	15. `WHERE` clause which is always pushdownable

** Subqueries/CTEs/Set Operations and joins are excluded since they are not in our scope in this investigation.


### The worker query generation as of Citus 7.3:

1. Iterate over the original target list.
2. Add the non-aggregate target entries as-is (including 
   pushdownable window functions which include aggregates)
3. Add the aggregate target entries after mutating them 
   (see `WorkerAggregateWalker()`)
4. If the mutation decides to convert any aggregates into 
    regular columns (e.g., reduce `agg (distinct column)` to `column`),
    4.1 Add the mutated column to the GROUP BY to ensure DISTINCTness
	       on the worker
[...]

PR #2070, opened 2018-03-27
Refactor logical optimizer [Phase 1] - Unify `extendedOpNode` creation

onderkalaci:
Implements one of the subitems of #2049.

Before this commit, we had a divergence among
the creation of master/worker extended op nodes.

This commit moves the related parts into a single place
and ensures the creation of master/extended op nodes to
share a common data structure.

The PR mostly consists of code moves and API simplifications.
In other words, there is almost no logic change at all. The only
logic change is in the `TransformSubqueryNode ()` where we used
to pass much unnecessary information (e.g., the global `tableNodeList `) 
to the function, which is fixed by only using the relevant 
`tableNodeList` in the function.

### DIFF
--- a/src/include/distributed/extended_op_node_utils.h
+++ b/src/include/distributed/extended_op_node_utils.h
@@ -1,0 +1,38 @@
+/*-------------------------------------------------------------------------
+ *
+ * extended_op_node_utils.h
+ *	  General Citus planner code.
+ *
+ * Copyright (c) 2018, Citus Data, Inc.
+ *-------------------------------------------------------------------------
+ */
+
+#ifndef EXTENDED_OP_NODE_UTILS_H_
+#define EXTENDED_OP_NODE_UTILS_H_
+
+#include "distributed/multi_logical_planner.h"
+
+
+/*
+ * ExtendedOpNodeProperties is a helper structure that is used to
+ * share the common information among the worker and master extended
+ * op nodes.
+ *
+ * It is designed to be a read-only singleton object per extended op node
+ * generation and processing.
+ */
+typedef struct ExtendedOpNodeProperties
+{
+	bool groupedByDisjointPartitionColumn;
+	bool repartitionSubquery;
+	bool hasNonPartitionColumnDistinctAgg;
+	bool pullDistinctColumns;
+	bool pushDownWindowFunctions;
+} ExtendedOpNodeProperties;
+
+
+extern ExtendedOpNodeProperties BuildExtendedOpNodeProperties(MultiExtendedOp *
+															  extendedOpNode);
+
+
+#endif /* EXTENDED_OP_NODE_UTILS_H_ */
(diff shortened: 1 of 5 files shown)


src/include/distributed/extended_op_node_utils.h
```diff
@@ -0,0 +1,39 @@
...
+{
+	bool groupedByDisjointPartitionColumn;
+	bool repartitionSubquery;
+	bool hasNonPartitionColumnDistinctAgg;
+	bool pullDistinctColumns;
+	bool pushDownWindowFunctions;
+} SharedSourceForExtendedOpNodes;
+
+
+extern SharedSourceForExtendedOpNodes BuildSharedSourceForExtendedOpNodes(
```
onderkalaci: `BuildSharedSourceForExtendedOpNodes ` and `SharedSourceForExtendedOpNodes `  are probably not the best names. I'm still trying to come-up with a better alternative. Any suggestion?

mtuncer: This was the first thing to itch my eyes. 
something shorter like `ExtendedOpNodeStats` or `ExtendedOpNodeInfo` or even `ExtendedOpNodeMetadata` might be more meaningful than `SharedSource...`

onderkalaci: I intentionally made the name so long that we have to change it before merging in 😄 

 I actually considered `ExtendedOpNodeStats `. But, it might not be compatible with future changes in case we add some more stuff here? What do you think?

Also, what about `ExtendedOpNodeSummary`?

onderkalaci: > ExtendedOpNodeStats

@mtuncer I'm planning to continue with `ExtendedOpNodeStats ` which makes more sense now.

src/include/distributed/extended_op_node_utils.h
```diff
@@ -0,0 +1,37 @@
...
+ * generation and processing.
+ */
+typedef struct ExtendedOpNodeStats
+{
+	bool groupedByDisjointPartitionColumn;
+	bool repartitionSubquery;
+	bool hasNonPartitionColumnDistinctAgg;
+	bool pullDistinctColumns;
+	bool pushDownWindowFunctions;
+} ExtendedOpNodeStats;
```
marcocitus: Stats suggests this is about monitoring. Maybe Properties or Flags? 

Also, since we only have flags in this struct, I wonder whether this could be an integer.

onderkalaci: I think I'd prefer to go with the `Properties` given that it's likely we could expand this in the future for non-boolean things as well.

src/include/distributed/extended_op_node_utils.h
```diff
@@ -0,0 +1,38 @@
+/*-------------------------------------------------------------------------
+ *
+ * extended_op_node_utils.h
+ *	  General Citus planner code.
+ *
+ * Copyright (c) 2017, Citus Data, Inc.
```
marcocitus: I wonder if we should just omit years :)

PR #2073, opened 2018-03-29
Copy range filters to joined tables

mtuncer:
Postgres correctly copies equality filters on joined relations to other ones when creating relation restrictions.  This helps shard pruning.

However, it does not do the same for the range filters. Therefore we end up doing less than optimal shard pruning. This becomes an  issue when you have subquery joins, and have lots of (1000++) shards.

This commit first determines filter dependencies between shards by looking at join restrictions. Then computes a filter dependency structure after flattening out dependencies. Finally copies range filters to depended relations.

Only range filters on the distribution columns are copied, the other filters are not changed. Also, only range distributed tables are processed.  Filters on reference, or hash distributed tables are not copied.

After this work, we will perform more aggressive shard pruning, and create less number of worker queries.

- [ ] Check/rebase parent PR #2013 
- [ ] Resolve memory issues related to relation restriction creation
- [ ] Refactor relation_restriction_equivalance related part
- [ ] Update regression tests


onderkalaci:
Shall we close this PR @mtuncer ?

mtuncer:
guess we can

PR #2075, opened 2018-03-30
Implement hash-repartitioning for single repartition joins (e.g., enable single repartition joins among hash distributed tables)

### LINKED ISSUE #1044: Implement hash-repartitioning for single repartition joins
We currently don't support single repartition joins for hash-distributed tables. Instead we fall back to a dual repartition join even if one of the tables is joined on the distribution column. We do have the machinery for performing hash-repartitioning, but previously this machinery was only used for dual re-partition joins and actual hash-distributed tables did not exist at the time.

onderkalaci:
Fixes #1044.

As far as I remember, this is the first PR I opened that touches repartitioning logic this heavily. So, it'd be great to get some architectural feedback as well (cc @ozgune @sumedhpathak). And, I'll be asking @marcocitus for the actual code review anyway.

This change consists of 4 commits:

1. Change `worker_hash_partition_table()` such that the divergence between Citus planner's hashing and  `worker_hash_partition_table()` becomes the same.
2. Rename single partitioning to single range partitioning. 
3. Main commit: Add single hash repartitioning. Basically, logical planner treats single hash and range partitioning almost equally. Physical planner, on the other hand, treats single hash and dual hash repartitioning almost equally (except for `JoinPruning`).
4. Add regression tests.

I've also done some benchmarking with the queries in the regression tests. I loaded `10,000,000` rows to 1+3 `c4.4xlarge` cluster, and run the tests. Also, I run the same tests with non-distributed tables with the same amount of data. (`citus.task_tracker_delay = 10ms`)

  | Single hash repartition | Dual hash repartition (master) | Local
-- | -- | -- | --
Q1 | (00:04.601) | (00:05.507)|    (00:11.300)
Q2 |(00:04.383)  |(00:05.367)|  (00:11.141)
Q3 | (00:19.161) |(00:54.891) |  (00:57.966)
Q4 | (00:35.196)  | (00:48.712) |   (01:15.138)
Q5 |(00:59.579) | (01:04.672)  |  (00:58.391)

### Missing Things
- [ ] We ignore tenant isolation. This could lead to unequal partition buckets during single hash repartitioning. It seems too much work to fix that for a minor performance gain. What do you thing? In any case, add a regression test to enterprise repo on a table which has isolated shards.
- [ ] Do/add some more complicated tests (e.g., with subqueries/ctes, more complex join orders etc)
- [ ] Think on the `hash` + `non-hash` single repartitioning some more.
- [ ] Could we come up with faster `JoinPruning()`  for `hash` + `hash` single repartition joins? The current approach goes through a very slow pruning process.


marcocitus:
> This doesn't work with tenant isolation. The benefit of implementing that seems low given that one hand there is tenants the other is repartitioning. What do you think?

It would be good to support tenant isolation / custom ranges, because being able to run background data warehouse queries on (a follower of) your main database is  pretty interesting, but given that we're doing this "off-cycle" I don't think we should get too caught up in the details. It's already an exciting step forward.


One slightly odd/interesting thing is that on my desktop (with local workers) the joins got slower

```sql
-- before this patch
postgres=# SELECT count(*) FROM test a JOIN test2 b ON (a.x = b.y);
 count  
────────
 100000
(1 row)

Time: 3655.331 ms (00:03.655)

-- after this patch
postgres=# SELECT count(*) FROM test a JOIN test2 b ON (a.x = b.y);
 count  
────────
 100000
(1 row)

Time: 6074.905 ms (00:06.075)
```

The main reason appears to be that there are only 16 merge tasks in the dual-repartition join case while there are 32 in the single-repartition join case. With 8 shards the single-repartition join is 3x faster. This appears to be mainly be an artifact of running on a desktop and the way we choose the number of intermediate partitions.

onderkalaci:
> It would be good to support tenant isolation / custom ranges, because being able to run background data warehouse queries on (a follower of) your main database is pretty interesting,

I think the main blocker for me to implement is that `HashPartitionId()` is not aware of non-uniform shard ranges. It assumes that the ranges provided to that have a uniform hash distribution.

As far as I can see, the steps to (nicely) implement that:
* Pass actual shard ranges (instead of `shardMinValue`s only) to `worker_hash_partition_table`
* Refactor ` FindShardIntervalIndex()` such that it doesn't rely on the `DistTableCacheEntry *cacheEntry` but to shard ranges/shard count etc., and we can call it in `HashPartitionId()` for completeness/correctness.

So that seems like a large change to tackle in the next release. Is that OK to you as well? If so, I'll update `SinglePartitionJoin()` function to skip tables with `hasUniformHashDistribution`.

onderkalaci:
>  Could we come up with faster JoinPruning() for hash + hash single repartition joins? The current approach goes through a very slow pruning process.

Well, it seems possible but I'll prefer to tackle it separately. I'll open an issue to discuss whether do we need the join pruning logic at all.

onderkalaci:
@marcocitus I think the PR is ready for you now. I've preferred to run `EXPLAIN` only in the regression tests due to the performance of repartitioning. Some manual correctness tests I've done can be found here: https://gist.github.com/onderkalaci/99c5cc9275843db4385da502fd254c0a

onderkalaci:
@marcocitus I think this is ready for the review now. I've implemented - tested single repartitioning with tenant isolation as well with the queries / random data generated here: https://gist.github.com/onderkalaci/99c5cc9275843db4385da502fd254c0a

src/backend/distributed/citus--7.3-3--7.4-1.sql
```diff
@@ -3,3 +3,4 @@
 DROP FUNCTION IF EXISTS master_expire_table_cache(regclass);
 DROP FUNCTION IF EXISTS pg_catalog.worker_fetch_regular_table(text, bigint, text[], integer[]);
 DROP FUNCTION IF EXISTS pg_catalog.worker_fetch_foreign_file(text, bigint, text[], integer[]);
+
```
marcocitus: extra newline

src/backend/distributed/citus--7.4-1--7.4-2.sql
```diff
@@ -0,0 +1,14 @@
+/* citus--7.4-1--7.4-2 */
+SET search_path = 'pg_catalog';
+
+DROP FUNCTION IF EXISTS worker_hash_partition_table(bigint, integer, text, text, oid, integer);
```
marcocitus: This function might actually get used right now, which means users cannot upgrade their workers without upgrading the master at the same time, which is a common pattern. Is there a way to keep this function?

src/backend/distributed/planner/multi_join_order.c
```diff
@@ -1044,11 +1049,15 @@ SinglePartitionJoin(JoinOrderNode *currentJoinNode, TableEntry *candidateTable,
...
+	JoinRuleType currentJoinRuleType = currentJoinNode->joinRuleType;
+
+	OpExpr *joinClause = NULL;
 
 	Oid relationId = candidateTable->relationId;
 	uint32 tableId = candidateTable->rangeTableId;
+	DistTableCacheEntry *candidateCacheEntry = DistributedTableCacheEntry(relationId);
 	Var *candidatePartitionColumn = PartitionColumn(relationId, tableId);
-	char candidatePartitionMethod = PartitionMethod(relationId);
+	char candidatePartitionMethod = candidateCacheEntry->partitionMethod;
```
marcocitus: I guess you could just use `PartitionMethod` here.

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -3531,12 +3542,50 @@ ApplyLocalJoin(MultiNode *leftNode, MultiNode *rightNode,
...
+static MultiNode *
+ApplySingleHashPartitionJoin(MultiNode *leftNode, MultiNode *rightNode,
+							 Var *partitionColumn, JoinType joinType,
+							 List *applicableJoinClauses)
+{
+	MultiJoin *joinNode =
+		ApplySinglePartitionJoin(leftNode, rightNode, partitionColumn, joinType,
+								 applicableJoinClauses);
+
+	joinNode->joinRuleType = SINGLE_HASH_PARTITION_JOIN;
```
marcocitus: would it make sense for this to become a parameter?

src/backend/distributed/worker/worker_partition_protocol.c
```diff
@@ -221,6 +245,39 @@ worker_hash_partition_table(PG_FUNCTION_ARGS)
...
 
 
+/*
+ * SyntheticShardIntervalArrayForShardMinValues returns a shard interval pointer array
+ * which gets the shardMinValues from the input shardMinValues array. Note that
+ * we simply calculate shard max values by decrementing previous shard min values by one.
+ *
+ * The function only fills the min/max values of shard the intervals. Thus, should
+ * not be used for general purpose operations.
+ */
```
marcocitus: Probably the right approach here would be to introduce some intermediate data structure that `SearchCachedShardInterval` works on. If someone makes changes to `SearchCachedShardInterval` then he's unlikely to know that it gets called in a weird way.

Maybe a nice way to do this would be if ShardInterval contained a sub-struct (e.g. ShardRange) that contained `minValue` and `maxValue`. Then you could quite efficiently build an array of pointers to all the `ShardRange`s inside the `ShardInterval`s during shard pruning, and construct an array of `ShardRange`s here.

We probably don't have the time to think about mitigating potential performance implications, so this approach seems fine for now.

src/backend/distributed/citus--7.4-1--7.4-2.sql
onderkalaci: Well, it turned out that we cannot do that without code duplication.

The reason is that Citus 7.3's hash pruning is different that Citus 7.4's hash pruning in the `worker_hash_partition_table()`. So, some workers relying on Citus shard pruning and the others relying on old shard pruning implemented in `worker_hash_partition_table()` ends up wrong query results.

So, if we really want to do that, we'd end up not removing any code from the older implementation and add the new implementation. We'd also probably end-up with some not nice `if` statements, acting according to the version.

How important to not break users during the upgrades? Shall we duplicate the code or fail hash repartitioning queries during the upgrades?

src/backend/distributed/planner/multi_logical_planner.c
marcocitus: I'd have a slight preference for that

src/backend/distributed/worker/worker_partition_protocol.c
```diff
@@ -1156,29 +1253,56 @@ RangePartitionId(Datum partitionValue, const void *context)
...
 
-	/* hash functions return unsigned 32-bit integers */
-	hashDatum = FunctionCall1(hashFunction, partitionValue);
-	hashResult = DatumGetUInt32(hashDatum);
-	hashPartitionId = (hashResult % partitionCount);
+	/*
+	 * This is to provide backward compatibility. See comments in
+	 * worker_hash_partition_table().
+	 */
+	if (hashPartitionContext->deprecatedAPIcall)
```
marcocitus: it might be cleaner to have an alternative implementation of `HashPartitionId` that we pass to `FilterAndPartitionTable` when using the deprecated API, which seems acceptable since it's only 10 lines of code.

src/backend/distributed/planner/multi_logical_planner.c
onderkalaci: Ok, I remembered the reason now :) 

It'd break the `RuleApplyFunction()` signature. So, do you still prefer that if I need to change it?

src/backend/distributed/worker/worker_partition_protocol.c
```diff
@@ -1156,6 +1256,52 @@ RangePartitionId(Datum partitionValue, const void *context)
...
+	ShardInterval **syntheticShardIntervalArray =
+		hashPartitionContext->syntheticShardIntervalArray;
+	FmgrInfo *comparisonFunction = hashPartitionContext->comparisonFunction;
+	Datum hashDatum = FunctionCall1(hashFunction, partitionValue);
+	int32 hashResult = 0;
+	uint32 hashPartitionId = 0;
+
+	if (hashDatum == 0)
+	{
+		return hashPartitionId;
```
marcocitus: could just do return 0 here

onderkalaci:
>  I'm wondering is whether we should have a GUC to disable single-repartition joins in case dual-repartition joins are faster.

That's a good idea to prevent any future complaints of the performance. (We should probably spend some more time to make the repartitioning faster irrespective of this.)

`citus.enable_single_repartition_joins`, the default value is ON, `PGC_USERSET` and `GUC_NO_SHOW_ALL` (e.g., we've too many configs and this seems fairly technical to show users unless someone suffers from it) ?

onderkalaci:
>  the default value is ON

Or even maybe the default is OFF ?

marcocitus:
Maybe default off for the moment, we can turn default on once we make re-partitioning faster.

onderkalaci:
> Maybe default off for the moment, we can turn default on once we make re-partitioning faster.

Set the default to off. Only enabled in the newly added regression tests.

lithp:
This PR broke the build on Windows. Here's the appveyor output: https://ci.appveyor.com/project/lithp/citus/build/1.0.1172-7fd4383-appveyor

(unsupported_worker_operations and transactions_drop_shards were broken by a previous PR, but all the other failures didn't appear until this PR was merged)

PR #2080, opened 2018-03-30
Attempt to address planner context crashes

jasonmp85:
Both of these are a bit of a shot in the dark. In one case, we noticed a stack trace where a caller received a null pointer and attempted to dereference the memory context field (at 0x010). In the other, I saw that any error thrown from within AdjustParseTree could keep the stack from being cleaned up (presumably if we push we should always pop).

Both stack traces were collected during times of high memory pressure and locally reproducing the problem locally or otherwise has been very tricky (i.e. it hasn't been reproduced reliably at all).


marcocitus:
I guess this probably doesn't solve the problem, but not crashing would be good.

PR #2083, opened 2018-04-05
Add partitioning support to MX tables

byucesoy:
Previously, we prevented creation of partitioned tables on Citus MX.
We decided to not focus on this feature until there is a need. Since
now there are requests for this feature, we are implementing support
for partitioned tables on Citus MX.

DESCRIPTION: Adds partitioning support to MX tables

### DIFF
--- a/src/backend/distributed/executor/multi_utility.c
+++ b/src/backend/distributed/executor/multi_utility.c
@@ -170,6 +170,7 @@ static void RangeVarCallbackForDropIndex(const RangeVar *rel, Oid relOid, Oid ol
 static void CheckCopyPermissions(CopyStmt *copyStatement);
 static List * CopyGetAttnums(TupleDesc tupDesc, Relation rel, List *attnamelist);
 static void PostProcessUtility(Node *parsetree);
+static void ProcessDropTableStmt(DropStmt *dropTableStatement);
 
 
 /*
@@ -361,6 +362,11 @@ multi_ProcessUtility(PlannedStmt *pstmt,
 			if (dropStatement->removeType == OBJECT_INDEX)
 			{
 				ddlJobs = PlanDropIndexStmt(dropStatement, queryString);
+			}
+
+			if (dropStatement->removeType == OBJECT_TABLE)
+			{
+				ProcessDropTableStmt(dropStatement);
 			}
 		}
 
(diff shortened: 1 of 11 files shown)


src/backend/distributed/executor/multi_utility.c
```diff
@@ -3770,3 +3776,54 @@ PlanGrantStmt(GrantStmt *grantStmt)
...
+
+	foreach(dropTableCell, dropTableStatement->objects)
+	{
+		List *tableNameList = (List *) lfirst(dropTableCell);
+		RangeVar *tableRangeVar = makeRangeVarFromNameList(tableNameList);
+		bool missingOK = true;
+		List *partitionList = NIL;
+		ListCell *partitionCell = NULL;
+
+		Oid relationId = RangeVarGetRelid(tableRangeVar, AccessExclusiveLock, missingOK);
```
onderkalaci: `AccessExclusiveLock ` seems too strong here. Why not `AccessShareLock` ?

src/backend/distributed/executor/multi_utility.c
```diff
@@ -3770,3 +3776,54 @@ PlanGrantStmt(GrantStmt *grantStmt)
...
+
+		if (relationId == InvalidOid ||
+			!IsDistributedTable(relationId) ||
+			!ShouldSyncTableMetadata(relationId) ||
+			!PartitionedTable(relationId))
+		{
+			continue;
+		}
+
+		partitionList = PartitionList(relationId);
```
onderkalaci: Shouldn't we return if `partitionList` has zero entries?

src/backend/distributed/executor/multi_utility.c
```diff
@@ -3770,3 +3776,54 @@ PlanGrantStmt(GrantStmt *grantStmt)
...
+
+		SendCommandToWorkers(WORKERS_WITH_METADATA, DISABLE_DDL_PROPAGATION);
+
+		foreach(partitionCell, partitionList)
+		{
+			Oid partitionRelationId = lfirst_oid(partitionCell);
+			char *detachPartitionCommand = GenerateDetachPartitionCommand(
```
onderkalaci: Minor: I think we recently use the following format more:
```C
char *detachPartitionCommand = 
    GenerateDetachPartitionCommand(partitionRelationId);

File: src/backend/distributed/metadata/metadata_sync.c [hunk 4]
```diff
@@ -219,7 +219,8 @@ ShouldSyncTableMetadata(Oid relationId)
  * following queries:
  *
  * (i)   Query that populates pg_dist_node table
- * (ii)  Queries that create the clustered tables
+ * (ii)  Queries that create the clustered tables (including foreign keys,
```
onderkalaci: We don't use the term `clustered` anymore. But, let's not touch it in this commit.

src/backend/distributed/worker/worker_drop_protocol.c
```diff
@@ -22,6 +22,7 @@
 #include "distributed/distribution_column.h"
 #include "distributed/master_metadata_utility.h"
 #include "distributed/metadata_cache.h"
+#include "distributed/multi_partitioning_utils.h"
```
onderkalaci: Extra include?

src/test/regress/expected/multi_mx_create_table.out
```diff
@@ -466,16 +466,16 @@ FROM pg_dist_partition NATURAL JOIN shard_counts
...
- citus_mx_test_schema.nation_hash                       |            2 |          16 | h          | s
- citus_mx_test_schema_join_1.nation_hash                |            3 |           4 | h          | s
- citus_mx_test_schema_join_1.nation_hash_2              |            3 |           4 | h          | s
- citus_mx_test_schema_join_2.nation_hash                |            3 |           4 | h          | s
- citus_mx_test_schema.nation_hash_collation_search_path |            3 |           4 | h          | s
- citus_mx_test_schema.nation_hash_composite_types       |            3 |           4 | h          | s
- mx_ddl_table                                           |            3 |           4 | h          | s
- app_analytics_events_mx                                |            3 |           4 | h          | s
- company_employees_mx                                   |            3 |           4 | h          | s
+ citus_mx_test_schema_join_1.nation_hash                |            2 |           4 | h          | s
```
onderkalaci: What is the reason for the test changes? It seems like `ORDER BY` issue only?

src/backend/distributed/executor/multi_utility.c
```diff
@@ -3770,3 +3776,54 @@ PlanGrantStmt(GrantStmt *grantStmt)
...
+		List *tableNameList = (List *) lfirst(dropTableCell);
+		RangeVar *tableRangeVar = makeRangeVarFromNameList(tableNameList);
+		bool missingOK = true;
+		List *partitionList = NIL;
+		ListCell *partitionCell = NULL;
+
+		Oid relationId = RangeVarGetRelid(tableRangeVar, AccessExclusiveLock, missingOK);
+
+		if (relationId == InvalidOid ||
+			!IsDistributedTable(relationId) ||
```
onderkalaci: Minor: Can't we fit these 4 lines into 2 lines?

src/test/regress/expected/multi_mx_partitioning.out
```diff
@@ -0,0 +1,268 @@
+--
+-- Distributed Partitioned Table MX Tests
+--
+SET citus.next_shard_id TO 1700000;
+SET citus.shard_count TO 4;
+SET citus.shard_replication_factor TO 1;
```
onderkalaci: Thanks for the great tests!

src/backend/distributed/executor/multi_utility.c
byucesoy: DROP TABLE query can have multiple tables in it. partitionList may be empty for first table but it may have entries for next ones. Still, I added an if check to `continue` so that we will not send lots of unnecessary commands to disable DDL propagation. (We still send some unnecessary ones unfortunately. I can add an additional flag to prevent that but I guess that would be overkill and cause difficult to maintain code.)

src/test/regress/expected/multi_mx_create_table.out
byucesoy: This is just change in colocation ids.

src/backend/distributed/executor/multi_utility.c
byucesoy: We can, but I think this is more readable. If you insist I can change it.

src/backend/distributed/executor/multi_utility.c
onderkalaci: We don't have to do it, but, I don't think it'd bad to add a flag named `sentDisableDDL` and skip it if already `true`.

PR #2089, opened 2018-04-09
Fix issue preventing multiple size function calls per query

marcocitus:
DESCRIPTION: Fix issue which prevented multiple citus_table_size calls per query


metdos:
@marcocitus can we merge this?

PR #2090, opened 2018-04-10
Refactor logical optimizer [Phase 2] - Eliminate code duplication in `WorkerExtendedOpNode()`

onderkalaci:
Implements one of the subitems of #2049.

Should be reviewed after or along with #2070.

Before this commit, we had code duplication in the
 `WorkerExtendedOpNode()`. The duplication was
noticeable and any change is prone to bugs.

The PR consists of 4 commits. Each commit incrementally
fixes the problem by moving certain parts of the duplicated 
code into smaller, better-documented functions.

### DIFF
--- a/src/backend/distributed/planner/multi_logical_optimizer.c
+++ b/src/backend/distributed/planner/multi_logical_optimizer.c
@@ -129,6 +129,20 @@ static Expr * AddTypeConversion(Node *originalAggregate, Node *newExpression);
 static MultiExtendedOp * WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
 											  ExtendedOpNodeProperties *
 											  extendedOpNodeProperties);
+static void ExpandWorkerTargetEntry(List *expressionList,
+									TargetEntry *originalTargetEntry,
+									bool addToGroupByClause,
+									List **newTargetEntryList,
+									AttrNumber *targetProjectionNumber,
+									List **groupClauseList,
+									Index *nextSortGroupRefIndex);
+static Index GetNextSortGroupRef(List *targetEntryList);
+static TargetEntry * GenerateWorkerTargetEntry(TargetEntry *targetEntry,
+											   Expr *workerExpression,
+											   AttrNumber targetProjectionNumber);
+static void AppendTargetEntryToGroupClause(TargetEntry *targetEntry,
+										   List **groupClause,
+										   Index *nextSortGroupRefIndex);
 static bool WorkerAggregateWalker(Node *node,
 								  WorkerAggregateWalkerContext *walkerContext);
 static List * WorkerAggregateExpressionList(Aggref *originalAggregate,
@@ -1833,7 +1847,7 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
 	bool queryHasAggregates = false;
 	bool distinctClauseSupersetofGroupClause = false;
 	bool distinctPreventsLimitPushdown = false;
-	bool createdNewGroupByClause = false;
+	bool createNewGroupByClause = false;
 	bool groupedByDisjointPartitionColumn =
 		extendedOpNodeProperties->groupedByDisjointPartitionColumn;
 	bool pushDownWindowFunction = extendedOpNodeProperties->pushDownWindowFunctions;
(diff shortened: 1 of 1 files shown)


src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -2092,6 +2031,168 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
...
+	{
+		Expr *newExpression = (Expr *) lfirst(newExpressionCell);
+		TargetEntry *newTargetEntry = NULL;
+
+		/* generate and add the new target entry to the target list */
+		newTargetEntry =
+			GenerateWorkerTargetEntry(originalTargetEntry, newExpression,
+									  *targetProjectionNumber);
+		(*targetProjectionNumber)++;
+		(*newTargetEntryList) = lappend(*newTargetEntryList, newTargetEntry);
```
marcocitus: Since pointer dereferencing adds some mental overhead, maybe turn these into variables and set the pointers at the end?

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -2092,6 +2031,168 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
...
+ * In summary, the function gets a list of expressions, converts them to target
+ * entries and updates all the necessary fields such that the expression is correctly
+ * added to the worker query's target list.
+ *
+ * Adding an expression to the target list includes generating a target entry,
+ * appending it to the newTargetEntryList and incrementing the projection number.
+ *
+ * If the expression is requested to be added to the group clause, the entry
+ * is appended to groupClauseList and sort group reference is incremented.
+ */
```
marcocitus: Could you add descriptions of the other parameters as well? (e.g. `targetProjectionNumber`)

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -2092,6 +2031,168 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
...
+static void
+ProcessWorkerExpressionList(List *expressionList, TargetEntry *originalTargetEntry,
```
marcocitus: I find the name a bit confusing. Would something like `ExpandWorkerTargetEntry` be appropriate?

src/backend/distributed/planner/multi_logical_optimizer.c
onderkalaci: I actually have a slight preference to keep them dereferenced pointers and create that cognitive overhead. On the high level, one of the goals of these series of PRs is to create a clear distinction between the inputs and the outputs. And, it feels like these dereferened pointers helps that.

That said if you think its too disturbing, I'm happy to change them as well.

src/backend/distributed/planner/multi_logical_optimizer.c
onderkalaci: How do they look now? I'm planning to add similar comments to #2107 if you're good with them

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -2110,6 +2049,182 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
...
+ *                        to this list for each expression that is in expressionList.
+ *  - targetProjectionNumber: A pointer to the worker query target list index. Should be
+ *                            incremented by one for each target entry added to
+ *                            newTargetEntryList.
+ *  - groupClauseList: Worker query's group clause list. The function adds one entry to
+ *                     this list for any expression that is in expressionList when
+ *                     addToGroupByClause flag is true.
+ *  - nextSortGroupRefIndex: A pointer to worker query's sort group reference index.
+ *                           Should be incremented by one for each element added to
+ *                           the groupClauseList.
```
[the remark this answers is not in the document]
onderkalaci: > what is a sort group reference index? what is a target projection number?

Aren't those too much information to have on a Citus planner function comment? 

>  when do we need to add entries to the group clause list? why does a target entry have a list of expressions?

I'm actually adding those details to the functions that we add in the third [PR](https://github.com/citusdata/citus/pull/2107/files#diff-b067f8998f46f5afc9aad3fc1c35aa09R1989). It looks like expanding those comments would be nicer?

This function seems a utility function such that it only handles expanding the target list.  I think the callers should give that information?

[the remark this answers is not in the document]
marcocitus: > I'm actually adding those details to the functions that we add in the third PR. It looks like expanding those comments would be nicer?

that works

PR #2097, opened 2018-04-12
Fix crash on Windows where there is no detail

lithp:
I'm not sure why the behavior is different on Windows but unfortunately it is.

### DIFF
--- a/src/backend/distributed/commands/multi_copy.c
+++ b/src/backend/distributed/commands/multi_copy.c
@@ -1233,9 +1233,10 @@ ReportCopyError(MultiConnection *connection, PGresult *result)
 	{
 		/* probably a constraint violation, show remote message and detail */
 		char *remoteDetail = PQresultErrorField(result, PG_DIAG_MESSAGE_DETAIL);
+		bool haveDetail = remoteDetail != NULL;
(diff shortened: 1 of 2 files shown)


src/backend/distributed/commands/multi_copy.c
```diff
@@ -1234,6 +1234,11 @@ ReportCopyError(MultiConnection *connection, PGresult *result)
 		/* probably a constraint violation, show remote message and detail */
 		char *remoteDetail = PQresultErrorField(result, PG_DIAG_MESSAGE_DETAIL);
 
+		if (remoteDetail == NULL)
+		{
+			remoteDetail = "(null)";
+		}
+
 		ereport(ERROR, (errmsg("%s", remoteMessage),
 						errdetail("%s", remoteDetail)));
```
marcocitus: Postgres sometimes uses: 

``` 
(remoteDetail != NULL) ? errdetail("%s", remoteDetail) : 0,
```

that way the detail is hidden instead of showing up as "DETAIL: (null)"

lithp: Nice, that's much better.

PR #2098, opened 2018-04-12
Fix cancellation of real time queries

lithp:
Fix `multi_real_time_transaction` test on Windows, without this change it blocks forever.

### DIFF
--- a/src/backend/distributed/executor/multi_real_time_executor.c
+++ b/src/backend/distributed/executor/multi_real_time_executor.c
@@ -199,6 +199,16 @@ MultiRealTimeExecute(Job *job)
 		{
 			MultiClientWait(waitInfo);
 		}
+
+#ifdef WIN32
+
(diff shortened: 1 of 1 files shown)


src/backend/distributed/executor/multi_real_time_executor.c
```diff
@@ -199,6 +199,16 @@ MultiRealTimeExecute(Job *job)
...
 		}
+
+#ifdef WIN32
+
+		/*
+		 * Don't call CHECK_FOR_INTERRUPTS because we want to clean up after ourselves,
+		 * calling pgwin32_dispatch_queued_signals sets QueryCancelPending so we leave
+		 * the loop.
+		 */
+		pgwin32_dispatch_queued_signals();
```
pykello: We won't need this once we switch to WaitEventSets.

PR #2102, opened 2018-04-12
ANDs are being explicit for worker queries in pushdown

### LINKED ISSUE #2067: Pushing down `HAVING` with multiple `AND`ed conditions fails
```SQL
SELECT max(value_1) FROM users_table GROUP BY user_id HAVING avg(value_1) > 10 AND count(*) > 4;
....
WARNING:  syntax error at or near ","
WARNING:  syntax error at or near ","
WARNING:  syntax error at or near ","
ERROR:  failed to execute task 4
...
```

The worker query:
```SQL
COPY (SELECT max(value_1) AS max, user_id AS worker_column_2, sum(value_1) AS worker_column_3, count(value_1) AS worker_column_4, count(*) AS worker_column_5 FROM users_table_102037 users_table WHERE true GROUP BY user_id HAVING (avg(value_1) > '10'::numeric), (count(*) > 4)) TO STDOUT
```

It looks like we're missing to call `make_ands_explicit()` on the `HAVING` clause. So, this fix should be 1-line change.

@velioglu hit this issue while working on the `tableau` integration.

(Note that the same query with `OR` in the `HAVING` clause works fine).

furkansahin:
This is a fairly simple PR that changes the AND clauses in having explicit for worker queries for pushdown planner. Since, they are going to be switched back to be implicit in worker itself, we should provide them in explicit form. Otherwise, the worker errors-out saying the query syntax is wrong.

DESCRIPTION: Adds support for multiple ANDs in Having for push_down planner

fixes #2067

### DIFF
--- a/src/backend/distributed/planner/multi_logical_optimizer.c
+++ b/src/backend/distributed/planner/multi_logical_optimizer.c
@@ -2133,6 +2133,19 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
 		(groupedByDisjointPartitionColumn || pushDownWindowFunction))
 	{
 		workerExtendedOpNode->havingQual = originalOpNode->havingQual;
+
+		/*
+		 * We converted the having expression to a list in subquery pushdown
+		 * planner. However, this query cannot be parsed as it is in the worker.
+		 * We should convert this back to being explicit for worker query
+		 * so that it can be parsed when it hits to the standard planner in
+		 * worker.
+		 */
+		if (IsA(workerExtendedOpNode->havingQual, List))
+		{
+			workerExtendedOpNode->havingQual =
+				(Node *) make_ands_explicit((List *) workerExtendedOpNode->havingQual);
+		}
 	}
 
 	return workerExtendedOpNode;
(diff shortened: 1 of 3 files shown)


mtuncer:
could you add DESCRIPTION line to PR.
also update the commit message to reflect what this does.

furkansahin:
@mtuncer I tested this PR against ANDs in target entries, order bys. Couldn't find any problem with them. Do you want me to add them as tests to this PR?

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -2133,6 +2133,19 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
 		(groupedByDisjointPartitionColumn || pushDownWindowFunction))
 	{
 		workerExtendedOpNode->havingQual = originalOpNode->havingQual;
+
+		/*
+		 * Postgres standard planner converts having qual node to a list of and
```
mtuncer: We are doing this so that the worker query could be deparsed. This information (although correct) is irrelevant to why we are doing this.

furkansahin: done!

mtuncer:
🚢

PR #2103, opened 2018-04-12
Add co-placement check to CoPartitionedTable

velioglu:
After removing the shard data fetch logic with [1915](https://github.com/citusdata/citus/pull/1915) and adding co-partition check with [1919](https://github.com/citusdata/citus/pull/1919), we expect to have co-placed tables to trigger local join. In order to get the correct result for co-placement, this PR adds check to CoPartitionedTables function. Without checking co-placement, function may return true for two append distributed tables which has same shard intervals but the shards are not located in the same node.

### DIFF
--- a/src/backend/distributed/planner/multi_physical_planner.c
+++ b/src/backend/distributed/planner/multi_physical_planner.c
@@ -194,7 +194,8 @@ static StringInfo MergeTableQueryString(uint32 taskIdIndex, List *targetEntryLis
 static StringInfo IntermediateTableQueryString(uint64 jobId, uint32 taskIdIndex,
 											   Query *reduceQuery);
 static uint32 FinalTargetEntryCount(List *targetEntryList);
-
+static bool CoPlacedShardIntervals(ShardInterval *firstInterval,
+								   ShardInterval *secondInterval);
 
 /*
  * CreatePhysicalDistributedPlan is the entry point for physical plan generation. The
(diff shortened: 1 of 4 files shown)


src/test/regress/expected/non_colocated_join_order.out
```diff
@@ -0,0 +1,55 @@
...
+SET citus.enable_repartition_joins to ON;
+SELECT count(*) FROM test_table_1, test_table_2 WHERE test_table_1.id = test_table_2.id;
+LOG:  join order: [ "test_table_1" ][ single partition join "test_table_2" ]
+DEBUG:  cannot use real time executor with repartition jobs
+HINT:  Since you enabled citus.enable_repartition_joins Citus chose to use task-tracker.
+ count 
+-------
+     9
+(1 row)
+
```
onderkalaci: Maybe decrease the debug level here, `EventTriggerInvoke 19977` might change

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -2341,6 +2342,53 @@ CoPartitionedTables(Oid firstRelationId, Oid secondRelationId)
 }
 
 
+/*
+ * CoPlacedInterval checks whether the given intervals located in the same nodes.
+ */
+static bool
+CoPlacedInterval(ShardInterval *firstInterval, ShardInterval *secondInterval)
```
onderkalaci: Maybe `ShardIntervalsCoPlaced` ?

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -2341,6 +2342,53 @@ CoPartitionedTables(Oid firstRelationId, Oid secondRelationId)
...
+
+	ListCell *firstShardPlacementCell = NULL;
+
+	/* Shards must have same number of placements */
+	if (list_length(firstShardPlacementList) != list_length(secondShardPlacementList))
+	{
+		return false;
+	}
+
+	foreach(firstShardPlacementCell, firstShardPlacementList)
```
onderkalaci: Maybe the following easier to follow? 
```
firstShardPlacementList = SortList(firstShardPlacementList,
											CompareShardPlacements);
econdShardPlacementList = SortList(secondShardPlacementList,
											CompareShardPlacements);

forboth()
       if doesn't match return false

return true
```

src/test/regress/expected/non_colocated_join_order.out
```diff
@@ -0,0 +1,55 @@
+--
```
onderkalaci: This is not added to any schedule?

PR #2104, opened 2018-04-12
Refactor query pushdown related logic

velioglu:
Refactor query pushdown related logic and adjust SubquerySqlTaskList (by changing its name to QueryPushdownSqlTaskList) to make it callable by modify planner. That will used by modify queries with multiple tables.


src/backend/distributed/planner/query_pushdown_planning_utils.c
```diff
@@ -0,0 +1,1280 @@
+/*-------------------------------------------------------------------------
+ *
+ * query_pushdown_planning_utils.c
```
marcocitus: "util" is supposed to mean "utility functions", but it has a tendency to mean "pile of uncategorized stuff". 

What if we name this file `query_pushdown_planing.c` and move all the logic relating to query pushdown in here? including e.g. `SubqueryPushdownMultiNodeTree`?

velioglu: I opened the PR in this way to not to spend too much time on refactoring after talked with Onder. Today, we have discussed the scope and timing with Onder and Metin after your review, and decided to extend it to cover the changes you mentioned. I will update the PR with these changes.

src/backend/distributed/planner/query_pushdown_planning_utils.c
```diff
@@ -0,0 +1,1704 @@
...
+ *
+ * Routines for creating pushdown plans for queries.
+ *
+ * Copyright (c) 2012-2016, Citus Data, Inc.
```
onderkalaci: Hmm, I think we might use `Copyright (c) 2018, Citus Data, Inc.`. Well, I'm not sure about that 😄 Could you find someone who knows this and maybe reply here what should we do?

src/backend/distributed/planner/query_pushdown_planning_utils.c
onderkalaci: Ping: we still see `util` in the name?

I think it'd be nice if you can expand the comment here.
* What are the basic requirements for query pushdown with plain words
* When do we use it

src/backend/distributed/planner/query_pushdown_planning_utils.c
```diff
@@ -0,0 +1,1704 @@
...
+#include "distributed/query_pushdown_planning_utils.h"
+#include "distributed/relation_restriction_equivalence.h"
+#include "nodes/nodeFuncs.h"
+#include "nodes/pg_list.h"
+#include "optimizer/clauses.h"
+#include "optimizer/var.h"
+#include "parser/parsetree.h"
+
+/* Local functions forward declarations */
+static bool JoinTreeContainsSubqueryWalker(Node *joinTreeNode, void *context);
```
onderkalaci: Seems like we usually have 2 new lines before local function declarations.

src/backend/distributed/planner/query_pushdown_planning_utils.c
```diff
@@ -0,0 +1,1704 @@
...
+static DeferredErrorMessage * DeferErrorIfUnsupportedTableCombination(Query *queryTree);
+static bool ShouldRecurseForRecurringTuplesJoinChecks(RelOptInfo *relOptInfo);
+static bool RelationInfoContainsRecurringTuples(PlannerInfo *plannerInfo,
+												RelOptInfo *relationInfo,
+												RecurringTuplesType *recurType);
+static bool IsRecurringRTE(RangeTblEntry *rangeTableEntry,
+						   RecurringTuplesType *recurType);
+static bool IsRecurringRangeTable(List *rangeTable, RecurringTuplesType *recurType);
+
+
```
onderkalaci: Why do we have 3 new lines here?

src/backend/distributed/planner/query_pushdown_planning_utils.c
```diff
@@ -0,0 +1,1704 @@
...
+
+static MultiNode * SubqueryPushdownMultiNodeTree(Query *queryTree);
+static void FlattenJoinVars(List *columnList, Query *queryTree);
+static void UpdateVarMappingsForExtendedOpNode(List *columnList,
+											   List *subqueryTargetEntryList);
+static MultiTable * MultiSubqueryPushdownTable(Query *subquery);
+static List * CreateSubqueryTargetEntryList(List *columnList);
+static Task * QueryPushdownTaskCreate(Query *originalQuery, int shardIndex,
```
onderkalaci: In general, we try to follow the rule: The function definition and declaration orders should be consistent. So, if `QueryPushdownTaskCreate ` is defined later in the code, we generally declare it later (compared to the other functions obviously such as `ErrorIfUnsupportedShardDistribution `).

We follow the same rule for header file declarations as well such as `SubqueryMultiNodeTree` vs `QueryPushdownSqlTaskList`.

Since we're doing a refactoring, I'd be in favor of sticking our conventions.

src/include/distributed/multi_logical_planner.h
```diff
@@ -189,15 +205,9 @@ extern bool SubqueryPushdown;
...
 											  PlannerRestrictionContext *
 											  plannerRestrictionContext);
-extern bool JoinTreeContainsSubquery(Query *query);
-extern bool WhereClauseContainsSubquery(Query *query);
 extern bool FindNodeCheck(Node *node, bool (*check)(Node *));
 extern bool SingleRelationRepartitionSubquery(Query *queryTree);
-extern DeferredErrorMessage * DeferErrorIfCannotPushdownSubquery(Query *subqueryTree,
-																 bool
-																 outerMostQueryHasLimit);
 extern DeferredErrorMessage * DeferErrorIfUnsupportedUnionQuery(Query *queryTree);
```
onderkalaci: I think `DeferErrorIfUnsupportedUnionQuery` should go to pushdown planner as well. And, as far as I can see, `RecurringTuplesType` should go there as well (probably static in the file?).

src/include/distributed/multi_logical_planner.h
```diff
@@ -189,15 +205,9 @@ extern bool SubqueryPushdown;
 extern MultiTreeRoot * MultiLogicalPlanCreate(Query *originalQuery, Query *queryTree,
```
onderkalaci: I think `extern bool SubqueryPushdown;` should go to query pushdown as well .

src/backend/distributed/planner/query_pushdown_planning.c
```diff
@@ -0,0 +1,1926 @@
...
+
+/*
+ * QueryPushdownSqlTaskList creates a list of SQL tasks to execute the given subquery
+ * pushdown job. For this, the it is being checked whether the query is router
+ * plannable per target shard interval. For those router plannable worker
+ * queries, we create a SQL task and append the task to the task list that is going
+ * to be executed.
+ */
+List *
+QueryPushdownSqlTaskList(Query *query, uint64 jobId,
```
marcocitus: Here we're mixing physical planner logic with logical planning logic. It'd be nice to preserve the separation. Maybe we should introduce another file?

velioglu: I have moved `QueryPushdownSqlTaskList` and related functions to multi_physical_planner (as you said as an option in our meeting yesterday), otherwise I need to introduce a new file with three functions only. After https://github.com/citusdata/citus/pull/2124, this logic will be used by router_planner to create tasks for modify queries with multiple range table entries.

PR #2105, opened 2018-04-13
Citus should not consider subqueries that are removed by PostgreSQL planner

onderkalaci:
PostgreSQL removes some of the joins from the query plans given that it doesn't need it, see [ join_is_removable()](https://doxygen.postgresql.org/analyzejoins_8c_source.html#l00159).

Before this commit, Citus was trying to access a subquery which is not planned by PostgreSQL, which was resulting in a segfault.

Now we ignore such subqueries during our planning as well. Also, since PostgreSQL doesn't give the relations restrictions for the tables** that are in the subqueries that are not planned, Citus is still able to pushdown the whole query.

** `multi_relation_restriction_hook` doesn't get called for such tables


onderkalaci:
> Which versions of PostgreSQL does this apply?

Seems like a very old feature.

Note that this is different than subquery pull-up where the query tree is modified. Here, I'm referring to PostgreSQL's internal data types that it keeps for its own planning: ` PlannerInfo *`

PR #2107, opened 2018-04-16
Simplify optimizer [Phase 3]  - Move processing each part of the query into its own functions

onderkalaci:
This the last PR intended for #2049 in 7.4.

This PR should be review after (or along with) #2070 and #2090. In general, **all there PRs don't change any of the logic at all**. Instead, the goal is to:

* Get rid of any code duplication
* Incremental changes to the optimizer made it slightly hard to follow the code, improve that and make it easier to implement new features
* Simplify the code by moving each part of query processing (e.g., `DISTINCT`, `LIMIT` etc) into its own function
* Make the interaction between each part of the query more obvious (e.g., How `DISTINCT` affects `LIMIT` etc)

Next steps (probably none of them is very urgent): 

* There are few quick wins we could implement soon, see [here](https://github.com/citusdata/citus/issues/2049#issuecomment-376133840) with title `Tasks that could be quick wins for Citus optimizer (Probably not realistic for 7.4, but, keeping the list here for future reference):`
* The processing of `LIMIT` and `ORDER BY` is very dependent on each other. It seems possible to separate them to some extent.
* `LIMIT` processing depends on too many things. Can't we try to simplify that?
* There are few `TODO` comments in the code now. I added those since those were quite confusing parts of the code. Now, the reader at least can understand why those codes are there.

### DIFF
--- a/src/backend/distributed/planner/multi_logical_optimizer.c
+++ b/src/backend/distributed/planner/multi_logical_optimizer.c
@@ -75,6 +75,80 @@ typedef struct WorkerAggregateWalkerContext
 } WorkerAggregateWalkerContext;
 
 
+/*
+ * QueryTargetList encapsulates the necessary fields to form
+ * worker query's target list.
+ */
+typedef struct QueryTargetList
+{
+	List *targetEntryList; /* the list of target entries */
+	AttrNumber targetProjectionNumber; /* the index of the last entry */
+} QueryTargetList;
+
+
+/*
+ * QueryGroupClause encapsulates the necessary fields to form
+ * worker query's group by clause.
+ */
+typedef struct QueryGroupClause
+{
+	List *groupClauseList; /* the list of group clause entries */
+	Index *nextSortGroupRefIndex; /* pointer to the index of the largest sort group reference index */
+} QueryGroupClause;
+
+
+/*
+ * QueryDistinctClause encapsulates the necessary fields to form
+ * worker query's DISTINCT/DISTINCT ON parts.
+ */
+typedef struct QueryDistinctClause
+{
+	List *workerDistinctClause; /* the list of distinct clause entries */
+	bool workerHasDistinctOn;
+} QueryDistinctClause;
+
+
+/*
+ * QueryWindowClause encapsulates the necessary fields to form
+ * worker query's window clause.
+ */
+typedef struct QueryWindowClause
+{
+	List *workerWindowClauseList; /* the list of window clause entries */
+	bool hasWindowFunctions;
+	Index *nextSortGroupRefIndex; /* see QueryGroupClause */
+} QueryWindowClause;
+
+
+/*
+ * QueryOrderByLimit encapsulates the necessary fields to form
+ * worker query's order by and limit clauses. Note that we don't
+ * keep track of limit offset clause since it is incorporated
+ * into the limit clause during the processing.
+ */
+typedef struct QueryOrderByLimit
+{
+	Node *workerLimitCount;
+	List *workerSortClauseList;
+	Index *nextSortGroupRefIndex; /* see QueryGroupClause */
+} QueryOrderByLimit;
+
+
+/*
+ * OrderByLimitReference a structure that is used commonly while
+ * processing sort and limit clauses.
+ */
+typedef struct OrderByLimitReference
+{
+	bool groupedByDisjointPartitionColumn;
+	bool groupClauseIsEmpty;
+	bool sortClauseIsEmpty;
+	bool hasOrderByAggregate;
+	bool canApproximate;
+	bool hasDistinctOn;
+} OrderByLimitReference;
+
+
 /* Local functions forward declarations */
 static MultiSelect * AndSelectNode(MultiSelect *selectNode);
 static MultiSelect * OrSelectNode(MultiSelect *selectNode);
@@ -129,20 +203,49 @@ static Expr * AddTypeConversion(Node *originalAggregate, Node *newExpression);
 static MultiExtendedOp * WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
 											  ExtendedOpNodeProperties *
 											  extendedOpNodeProperties);
+static bool TargetListHasAggragates(List *targetEntryList);
+static void ProcessTargetListForWorkerQuery(List *targetEntryList,
+											ExtendedOpNodeProperties *
+											extendedOpNodeProperties,
+											QueryTargetList *queryTargetList,
+											QueryGroupClause *queryGroupClause);
+static void ProcessHavingClauseForWorkerQuery(Node *havingQual,
+											  ExtendedOpNodeProperties *
+											  extendedOpNodeProperties,
+											  Node **workerHavingQual,
+											  QueryTargetList *queryTargetList,
+											  QueryGroupClause *queryGroupClause);
+static void ProcessDistinctClauseForWorkerQuery(List *distinctClause, bool hasDistinctOn,
+												List *groupClauseList,
+												bool queryHasAggregates,
+												QueryDistinctClause *queryDistinctClause,
+												bool *distinctPreventsLimitPushdown);
+static void ProcessWindowFunctionsForWorkerQuery(List *windowClauseList,
+												 List *originalTargetEntryList,
+												 QueryWindowClause *queryWindowClause,
+												 QueryTargetList *queryTargetList);
+static void ProcessLimitOrderByForWorkerQuery(OrderByLimitReference orderByLimitReference,
+											  Node *originalLimitCount, Node *limitOffset,
+											  List *sortClauseList, List *groupClauseList,
+											  List *originalTargetList,
+											  QueryOrderByLimit *queryOrderByLimit,
+											  QueryTargetList *queryTargetList);
+static OrderByLimitReference BuildOrderByLimitReference(bool hasDistinctOn, bool
+														groupedByDisjointPartitionColumn,
+														List *groupClause,
+														List *sortClauseList,
+														List *targetList);
 static void ExpandWorkerTargetEntry(List *expressionList,
 									TargetEntry *originalTargetEntry,
 									bool addToGroupByClause,
-									List **newTargetEntryList,
-									AttrNumber *targetProjectionNumber,
-									List **groupClauseList,
-									Index *nextSortGroupRefIndex);
+									QueryTargetList *queryTargetList,
+									QueryGroupClause *queryGroupClause);
 static Index GetNextSortGroupRef(List *targetEntryList);
 static TargetEntry * GenerateWorkerTargetEntry(TargetEntry *targetEntry,
 											   Expr *workerExpression,
 											   AttrNumber targetProjectionNumber);
 static void AppendTargetEntryToGroupClause(TargetEntry *targetEntry,
-										   List **groupClause,
-										   Index *nextSortGroupRefIndex);
+										   QueryGroupClause *queryGroupClause);
 static bool WorkerAggregateWalker(Node *node,
 								  WorkerAggregateWalkerContext *walkerContext);
 static List * WorkerAggregateExpressionList(Aggref *originalAggregate,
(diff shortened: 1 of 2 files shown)


src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -74,6 +74,20 @@ typedef struct WorkerAggregateWalkerContext
 	bool pullDistinctColumns;
 } WorkerAggregateWalkerContext;
 
+/*
+ * LimitOrderByStats a structure that is used commonly while
+ * processing sort and limit clauses.
+ */
+typedef struct LimitOrderByStats
```
marcocitus: Not sure I like the name Stats, since this isn't just for monitoring.

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -1872,162 +2022,323 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
...
+ *
+ * The function also updates the workerHavingQual if the having clause can
+ * be safely pushed down to the worker query.
+ *
+ * TODO: Do we really need to pull anything to the coordinator if we're already
+ * pushing down the HAVING clause?
+ *
+ *     inputs: originalHavingQual, extendedOpNodeStats
+ *     outputs: workerHavingQual, newTargetEntryList, targetProjectionNumber,
+ *              groupClauseList, nextSortGroupRefIndex
```
marcocitus: since the function names are not that descriptive, I'm going to want descriptions for all the parameters.

onderkalaci: Are you OK with the format I propose here: https://github.com/citusdata/citus/pull/2090/files#diff-b067f8998f46f5afc9aad3fc1c35aa09R2060

onderkalaci: Actually, rebasing this PR on to the previous ones took more than I thought. 

Feel free to add comments here but I'll probably apply them after we merge the previous ones to keep the rebase efforts smaller.

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -74,6 +74,20 @@ typedef struct WorkerAggregateWalkerContext
 	bool pullDistinctColumns;
 } WorkerAggregateWalkerContext;
 
+/*
+ * LimitOrderByReference a structure that is used commonly while
+ * processing sort and limit clauses.
+ */
+typedef struct LimitOrderByReference
```
marcocitus: It reads a bit like "order by reference". Maybe swap it to `OrderByLimitReference`, since that's also the SQL order.

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -1821,41 +1879,143 @@ AddTypeConversion(Node *originalAggregate, Node *newExpression)
...
+	Node *originalLimitOffset = originalOpNode->limitOffset;
+	List *originalWindowClause = originalOpNode->windowClause;
+	List *originalDistinctClause = originalOpNode->distinctClause;
+	bool hasDistinctOn = originalOpNode->hasDistinctOn;
+
+	int originalGroupClauseLength = list_length(originalGroupClauseList);
+	bool queryHasAggregates = TargetListHasAggragates(originalTargetEntryList);
+
+	/* calculate the next sort group index based on the original target list */
+	nextSortGroupRefIndex = GetNextSortGroupRef(originalTargetEntryList);
```
marcocitus: The use of `targetProjectNumber` and `nextSortGroupRefIndex` keeps bugging me a bit. Even though they are essentially simple counters of how much stuff we added to the target list and group by/order by, it's hard to figure out what they represent from their name, or the code, or the comments. It's also a bit hard to figure out which variables belong together and what their relationships are.

I think the quickest way to resolve that is to add a comment saying what the variable represents to help the reader digest the parameters. A more structural approach could be to just move these variables into the `MultiExtendedOpNode` and passing the full `workerExtendedOpNode` in the functions below. Even though they are not useful in the physical planner, the improved simplicity/readability would be worth having some redundant fields.

src/backend/distributed/planner/multi_logical_optimizer.c
onderkalaci: Makes sense

src/backend/distributed/planner/multi_logical_optimizer.c
onderkalaci: I'll be unifying multiple comments from you:

>  It's also a bit hard to figure out which variables belong together and what their relationships are.
and
>Given the number of workerExtendedOpNode fields we pass, I'm wondering whether we shouldn't just pass workerExtendedOpNode

Thinking about your comments, I'd propose the following:

* We create few new smaller structs
* We'd be able to decrease the number of parameters passed to many functions
* We'd be able to create a relationship between the two indexes and their corresponding data structures (note that `nextSortGroupRefIndex` might be in multiple places)

```typdef struct QueryTargetList
{
List **newTargetEntryList;
AttrNumber *targetProjectionNumber;
}

typdef struct QueryGroupClause
{
List **groupClauseList;
AttrNumber *nextSortGroupRefIndex;
}

typedef struct QuerySortLimitClauses
{
Node **workerLimitCount;
List **workerSortClauseList
AttrNumber *nextSortGroupRefIndex;
}

```

@marcocitus  Since the above is a relatively large change in the PR, I'll wait confirmation from you before doing any changes.

marcocitus: I think this makes a lot of sense and is way better than my proposal.

onderkalaci: I pushed the changes. I think now it looks (much) better. What do you think?

`OrderByLimit` processing very tightly coupled, so that part still looks a bit crowded. Still, that seems acceptable for now.

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -75,6 +75,80 @@ typedef struct WorkerAggregateWalkerContext
...
+
+/*
+ * QueryWindowClause encapsulates the necessary fields to form
+ * worker query's window clause.
+ */
+typedef struct QueryWindowClause
+{
+	List **workerWindowClauseList;
+	bool *hasWindowFunctions;
+	Index *nextSortGroupRefIndex;
```
marcocitus: Why are these pointers? Can't we just modify the values if we have a pointer to the struct?

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -1877,175 +2095,363 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
...
-
 		ExpandWorkerTargetEntry(newExpressionList, originalTargetEntry,
-								createNewGroupByClause, &newTargetEntryList,
-								&targetProjectionNumber, &groupClauseList,
-								&nextSortGroupRefIndex);
+								workerAggContext->createGroupByClause,
+								queryTargetList->targetEntryList,
+								queryTargetList->targetProjectionNumber,
+								queryGroupClause->groupClauseList,
+								queryGroupClause->nextSortGroupRefIndex);
```
marcocitus: Could we pass on the structs to ExpandWorkerTargetEntry as well?

src/backend/distributed/planner/multi_logical_optimizer.c
onderkalaci: I think we could. But, my idea was to make sure that we have a singleton `workerExtendedOpNode ` and all the fields in these structs directly point to the `workerExtendedOpNode`'s fields to prevent any weird bugs in the future.

I was mostly suspicious about `nextSortGroupRefIndex ` which is shared by multiple structs.

Do you think would it be better to not have pointers here?

marcocitus: I think using a lot of pointers is prone to bugs such as segmentation faults being introduced. I like having pointers to structs that are allocated on the stack (as you did), and then modifying the fields.

src/backend/distributed/planner/multi_logical_optimizer.c
onderkalaci: Yes, makes sense. Even to one lower level function `AppendTargetEntryToGroupClause`

src/backend/distributed/planner/multi_logical_optimizer.c
onderkalaci: But in this case, they are all references to some already allocated variables:

```C
Index nextSortGroupRefIndex = 0;

... 
queryGroupClause.nextSortGroupRefIndex = &nextSortGroupRefIndex;
``` 

One minor reason that I try to push back on doing that is we'd end up with assigning the values back to some other structs in between function calls such as:

```C

/* process each part of the query in order to generate the worker query's parts */
ProcessTargetListForWorkerQuery(originalTargetEntryList, extendedOpNodeProperties,
									&queryTargetList, &queryGroupClause);

...

queryWindowClause.nextSortGroupRefIndex = queryGroupClause.nextSortGroupRefIndex;
ProcessWindowFunctionsForWorkerQuery(originalWindowClause, originalTargetEntryList,
										 &queryWindowClause, &queryTargetList);

```
Becuase those are acutally global variables within the function.

And, missing such things could lead to bugs that are hard to understand, whereas segfault are pretty simple to recognise 😃 

But, I'm OK to change if you still think that non-pointer fields are better.

onderkalaci: (I can see that using these structs on other places would prone to what you've described, but, I don't think these are common structs to use at any other place?)

marcocitus: > One minor reason that I try to push back on doing that is we'd end up with assigning the values back to some other structs in between function calls such as:

That could be a good reason to keep a pointer in case of `nextSortGroupRefIndex`, but then I again wonder whether the separation really makes sense and the functions shouldn't just be working on the `MultiExtendedOp` directly.

Anyway, I think de-pointering everything that isn't shared between structs would be acceptable.

> And, missing such things could lead to bugs that are hard to understand, whereas segfault are pretty simple to recognise 😃

and extremely annoying to customers

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -1821,41 +1925,168 @@ AddTypeConversion(Node *originalAggregate, Node *newExpression)
...
+	queryOrderByLimit.nextSortGroupRefIndex = &nextSortGroupRefIndex;
+
+	/* initialize the remaining parts to default values */
+	queryTargetList.targetEntryList = NIL;
+	queryDistinctClause.workerDistinctClause = NIL;
+	queryDistinctClause.workerHasDistinctOn = false;
+	queryWindowClause.hasWindowFunctions = false;
+	queryWindowClause.workerWindowClauseList = NIL;
+	queryOrderByLimit.workerLimitCount = NULL;
+	queryOrderByLimit.workerSortClauseList = NIL;
```
marcocitus: it might be worth just doing a `memset` to initialise to 0 instead, in case we ever forget a field.

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -1877,175 +2108,348 @@ WorkerExtendedOpNode(MultiExtendedOp *originalOpNode,
...
+ * The function also sets distinctPreventsLimitPushdown. As the name reveals,
+ * distinct could prevent pushwing down LIMIT clauses later in the planning.
+ * For the details, see the comments in the function.
+ *
+ *     inputs: distinctClause, hasDistinctOn, groupClauseList, queryHasAggregates
+ *     outputs: queryDistinctClause, distinctPreventsLimitPushdown
+ *
+ */
+static void
+PrcoessDistinctClauseForWorkerQuery(List *distinctClause, bool hasDistinctOn,
```
marcocitus: Prcoess -> Process

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -1821,41 +1925,168 @@ AddTypeConversion(Node *originalAggregate, Node *newExpression)
...
+ * for the given inputs.
+ *
+ * The function walks over the input targetEntryList. For the entries
+ * with aggregates in them, it calls the recursive aggregate walker function to
+ * create aggregates for the worker nodes. For example, the avg() is sent to
+ * the worker with two expressions count() and sum(). Thus, a single target entry
+ * might end up with multiple expressions in the worker query.
+ *
+ * The function doesn't change the aggragates in the window functions and sends them
+ * as-is. The reason is that Citus currently only supports pushiing down window
```
marcocitus: pushiing -> pushing

src/backend/distributed/planner/multi_logical_optimizer.c
onderkalaci: makes sense

PR #2111, opened 2018-04-17
Full CMake PoC (WIP)

jasonmp85:
I'm setting this aside for a bit, but it's pretty much to a working
state for now.

TODO: Add more description here.


sumedhpathak:
@jasonmp85 Maybe we should close this for now? What's the best way to save this work?

jasonmp85:
> Maybe we should close this for now? What's the best way to save this work?

I'll close this but not delete the branch.

PR #2112, opened 2018-04-18
[DNM] Cmake build appveyor

lithp:
Adding to the appveyor instructions, telling it how to build using CMake.

This runs into a problem though, I think EnterpriseDB is too much of a fork for us to build off of. `CREATE EXTENSION citus` crashes because `MaintenanceControlDaemon` is `NULL`, it appears that `MaintenanceDaemonShmemInit` is never called.

I'm calling this quits for now. The whole point was to speed up appveyor by allowing us to use a pre-built postgres. I don't believe there are any non-EDB pre-built windows packages.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,3 @@
-# FIXME The changes here are an example; haven't actually tested on AppVeyor
 # Notes:
 #   - Minimal appveyor.yml file is an empty file. All sections are optional.
 #   - Indent each level of configuration with 2 spaces. Do not use tabs!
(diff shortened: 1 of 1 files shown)


lithp:
@jasonmp85 When you built and installed on Windows, were you able to run the regression tests or run `CREATE EXTENSION`?

lithp:
@jasonmp85 convinced me that I was wrong about EnterpriseDB, and in fact our initialization wasn't being called because `_PG_init` isn't marked `PGDLLEXPORT`, so it was never being called by Postgres. I added `PGDLLEXPORT` to `_PG_init` and made a little progress: postgres now crashes immediately upon starting with an error which indicates the heap was corrupted.

sumedhpathak:
@lithp Should we close this issue?

jasonmp85:
@sumedhpathak I think it was only ever open to trigger an appveyor build. Closing…

Issue #2086, opened 2018-04-05
Our abort handling code is unstable and could cause crashes.


lithp:
By adding [one line of code](https://github.com/citusdata/citus/pull/2085/files) I was able to cause Citus to crash (this class of error actually caused crashes on Windows):

```c
ereport(ERROR, (errmsg("could not remove file \"%s\"", filename->data)));
```

Why does `ereport(ERROR` cause a crash? Because it aborts the transaction! Why does that cause a crash? Because our abort handler calls this line of code... which then aborts the transaction, which calls this line of code, which aborts the transaction, until Postgres finally gives up.

There are a couple ways to fix this:
1) Be extra careful to not `ERROR` during anything we call from `CoordinatedTransactionCallback`.
2) Add an exception handler to `CoordinatedTransactionCallback` which turns the `ERROR` into a `WARNING` and allows the abort to continue? We're already aborting anyway so we're not really ignoring the error.
3) Don't do any significant work in the transaction callbacks beside changing variables?
4) Add some kind of circuit breaker which makes sure our handler is only called once per transaction? If we've already tried to abort then don't call the functions again.

(1) is a fragile solution because we'll definitely forget and c's type system isn't useful enough to enforce it statically. At worst, it would require threading `throwOnError` parameters into every function which is ever called from `CoordinatedTransactionCallback`.

I think (2) is the best solution? But I'm not sure.

#### Reproduction instructions and gdb output:

```sql
brian=# BEGIN;
BEGIN
brian=# SELECT create_intermediate_result('squares', 'SELECT s, s*s FROM generate_series(1, 5) s');
 create_intermediate_result 
----------------------------
                          5
(1 row)

brian=# SELECT * FROM hello;
WARNING:  AbortTransaction while in ABORT state
WARNING:  AbortTransaction while in ABORT state
WARNING:  AbortTransaction while in ABORT state
ERROR:  relation "hello" does not exist
LINE 1: SELECT * FROM hello;
                      ^
ERROR:  could not remove file "base/pgsql_job_cache/10_0_1"
ERROR:  could not remove file "base/pgsql_job_cache/10_0_1"
ERROR:  could not remove file "base/pgsql_job_cache/10_0_1"
ERROR:  could not remove file "base/pgsql_job_cache/10_0_1"
PANIC:  ERRORDATA_STACK_SIZE exceeded
server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
[...]

PR #2115, opened 2018-04-18
Turn the crashes on Windows into WARNINGs

lithp:
A partial solution to https://github.com/citusdata/citus/issues/2086

On any platform `ERROR`s thrown during this block could cause Citus to crash, on Windows that `ERROR` is actually thrown (a PR to fix that root cause is incoming)

I'm not sure how to write a test for this, it'd be nice if we could inject errors but sans that, maybe there's some way I could mess with the job cache directory which would cause removing it to fail?

### DIFF
--- a/src/backend/distributed/transaction/transaction_management.c
+++ b/src/backend/distributed/transaction/transaction_management.c
@@ -183,6 +183,8 @@ CoordinatedTransactionCallback(XactEvent event, void *arg)
 
 		case XACT_EVENT_ABORT:
 		{
+			MemoryContext savedContext = CurrentMemoryContext;
+
 			/*
 			 * FIXME: Add warning for the COORD_TRANS_COMMITTED case. That
 			 * can be reached if this backend fails after the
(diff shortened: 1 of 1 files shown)


src/backend/distributed/transaction/transaction_management.c
```diff
@@ -194,7 +196,38 @@ CoordinatedTransactionCallback(XactEvent event, void *arg)
...
+
+				/* turn the ERROR into a WARNING and emit it */
+				edata->elevel = WARNING;
+				ThrowErrorData(edata);
+
+				/* leave the error handling system */
+				FlushErrorState();
+				MemoryContextSwitchTo(savedContext);
+			}
+			PG_END_TRY();
```
marcocitus: It'd be nice if we could move this whole block into a separate function. Could even use a function pointer in case we run into this problem again.

lithp: Good idea, pushed an amended commit.

src/backend/distributed/transaction/transaction_management.c
```diff
@@ -194,7 +196,17 @@ CoordinatedTransactionCallback(XactEvent event, void *arg)
...
-			RemoveIntermediateResultsDirectory();
+			{
+				/*
+				 * On Windows it's not possible to delete a file before you've closed all
+				 * handles to it (rmdir will return success but not take effect). Since
+				 * we're in an ABORT handler it's very likely that not all handles have
+				 * been closed; force them closed here before running
+				 * RemoveIntermediateResultsDirectory.
+				 */
+				AtEOXact_Files();
```
marcocitus: Could there be any ERRORs in here?

lithp: Do you mean in the call to `AtEOXact_Files()`? I don't think so, it's called by Postgres just after calling the user xact callbacks, if I remember correctly.

src/backend/distributed/transaction/transaction_management.c
```diff
@@ -194,7 +196,17 @@ CoordinatedTransactionCallback(XactEvent event, void *arg)
 			 * transaction management. Do so before doing other work, so the
 			 * callbacks still can perform work if needed.
 			 */
-			RemoveIntermediateResultsDirectory();
+			{
```
marcocitus: This extra block seems to not contribute much.

src/backend/distributed/transaction/transaction_management.c
```diff
@@ -423,3 +435,48 @@ ActiveSubXacts(void)
...
+ * everything up before control passes back to us.
+ */
+static void
+SwallowErrors(void (*func)())
+{
+	MemoryContext savedContext = CurrentMemoryContext;
+
+	PG_TRY();
+	{
+		func();
```
pykello: Assuming `func` can also do database stuff, we should start a subtransaction before `PG_TRY()` and rollback to it in `PG_CATCH()` (and release it at the end of `PG_TRY()`) so we also restore transaction state. See [CollectBasicUsageStatistics()](https://github.com/citusdata/citus/blob/release-7.3/src/backend/distributed/utils/statistics_collection.c#L104).

It doesn't do database access in the single usage in this PR, but assuming `SwallowErrors()` might be used in future, it is safer to do it.

marcocitus: This meant to be called from the abort handler, so it's probably too late to open a subtransaction.

PR #2117, opened 2018-04-19
Configure appveyor to run regression tests

lithp:
- Add install.pl to instal .sql files on Windows
- Remove a hack to PGDLLIMPORT some variables
- Add citus_version.o to the Makefile
- Fix pg_regress_multi's PATH generation on Windows
- Output regression.diffs when the tests fail
- Fix permissions in data directory, make sure postgres can play with it

Now that all the test fixes have been applied, actually run the tests!

I've based this PR on `prevent-abort-crashes` (#2115) because it depends on those changes in order for the tests to pass. The intent is for that PR to be merged first, and to then merge this one.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -49,11 +49,6 @@ install:
   # make postgres think we're a contrib module
   - git clone -b REL_10_STABLE https://github.com/postgres/postgres C:\projects\postgres
   - git -C C:\projects\postgres apply C:\projects\citus\windows\Mkvcbuild.pm.patch
-  # temporary hack: mark enable_hashagg PGDLLIMPORT
-  - git -C C:\projects\postgres cherry-pick 935dee9ad5a8d12f4d3b772a6e6c99d245e5ad44 & exit 0
-  # an even bigger hack, this file has a merge conflict so skip it
-  - git -C C:\projects\postgres reset src\include\optimizer\cost.h
-  - git -C C:\projects\postgres checkout src\include\optimizer\cost.h
   # only build Citus, don't build the other contrib modules
   - git -C C:\projects\postgres apply C:\projects\citus\windows\Mkvcbuild-minimize.patch
   - rmdir /s /q C:\projects\postgres\contrib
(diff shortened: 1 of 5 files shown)


src/test/regress/pg_regress_multi.pl
```diff
@@ -115,10 +115,13 @@ ()
 # want to find binaries if they're not in PATH.
 if (defined $bindir)
 {
-    $ENV{PATH} = ($ENV{PATH} || '').":$bindir";
+    if ($usingWindows) {
+        $ENV{PATH} = ($ENV{PATH} || '').";$bindir";
```
pykello: 1. What about `$libdir`? I think we should also use semi-colon separators for VS there?
2. Does `LD_LIBRARY_PATH` work for VS? [MSDN](https://msdn.microsoft.com/en-us/library/6y6t9esh.aspx) says it is `LINK`?

appveyor.yml
```diff
@@ -80,4 +75,18 @@ build_script:
 #---------------------------------#
 
 # disable automatic tests
```
pykello: Remove this comment?

appveyor.yml
```diff
@@ -80,4 +75,18 @@ build_script:
...
+before_test:
+  - ps: cd C:\projects\postgres\src\tools\msvc
+  - install C:\projects\pgsql
+  - cd C:\projects\postgres\contrib\citus\src\backend\distributed
+  - copy citus--*.sql C:\projects\pgsql\share\extension\
+  - perl install.pl C:\projects\pgsql
+
+test_script:
+  - ps: cd C:\projects\postgres\contrib\citus\src\test\regress
+  - perl pg_regress_multi.pl --bindir=C:\projects\pgsql\bin --load-extension=citus -- --schedule=multi_schedule
```
pykello: Are we planning to run other tests (isolation, mx, ...) in windows?

lithp: I think that makes sense to do eventually, but the regular regression tests are good enough for now.

src/test/regress/pg_regress_multi.pl
lithp: Pushed a change which correctly adds `$libdir` to `PATH` if you've set it. None of the other variables are relevant for Windows. Looks like Windows uses both `LIB` and `LINK`. I don't know why someone would want to set those though, on Windows installation works by adding everything you could need into `$prefix/lib`, for better or worse.

appveyor.yml
lithp: Good catch, fixed

PR #2121, opened 2018-04-24
Add support for (VACUUM | ANALYZE) VERBOSE

### LINKED ISSUE #1050: Implement distributed (VACUUM|ANALYZE) VERBOSE
In #1013, we implemented all options of `VACUUM` and `ANALYZE`, except `VERBOSE`.

Adding the `VERBOSE` option would require routing and displaying output from all worker nodes during `VACUUM`/`ANALYZE` execution: the implementation we chose was relatively straightforward and non-intrusive and the added code complexity of `VERBOSE` didn't seem worth it in a first pass at the `VACUUM` feature.

When a use case presents itself, this ticket will be waiting.

furkansahin:
DESCRIPTION: Add support for (VACUUM | ANALYZE) VERBOSE

This PR adds support for (VACUUM | ANALYZE) VERBOSE on distributed tables. The idea is to use the notice hook and change the log direction from logfile to the console.

The notice handler is implemented as we discussed in the mail thread `Citus Notice Handler`. For now, the default for the notices is set to DEBUG1. It is chosen consciously to not allow all of the notices to go to the console. When we set it to NOTICE or INFO, we have some weird cases we should consider. 
For example, if we set the notice level to NOTICE and run;
```
SELECT create_distributed_table('schema.table', 'i');
```

We print to the console `Schema 'schema' already exists, skipping` for each connection and I don't think this is a good user experience. So, I left it as DEBUG1.

As @onderkalaci suggested, one approach we should consider is that instead of having a static `CitusNoticeLogLevel`, we might parse the message we get from the connection and learn which level used in the worker and use the respective log level in the coordinator for reporting. However, I don't really think that this problem is related to that PR. If we want to implement this mechanism, I propose to do it with another PR.

fixes #1050

### DIFF
--- a/src/backend/distributed/connection/connection_management.c
+++ b/src/backend/distributed/connection/connection_management.c
@@ -25,6 +25,7 @@
 #include "distributed/metadata_cache.h"
 #include "distributed/hash_helpers.h"
 #include "distributed/placement_connection.h"
+#include "distributed/remote_commands.h"
 #include "distributed/version_compat.h"
 #include "mb/pg_wchar.h"
 #include "utils/hsearch.h"
@@ -41,7 +42,11 @@ static uint32 ConnectionHashHash(const void *key, Size keysize);
 static int ConnectionHashCompare(const void *a, const void *b, Size keysize);
 static MultiConnection * StartConnectionEstablishment(ConnectionHashKey *key);
 static void AfterXactHostConnectionHandling(ConnectionHashEntry *entry, bool isCommit);
+static void DefaultCitusNoticeProcessor(void *arg, const char *message);
 static MultiConnection * FindAvailableConnection(dlist_head *connections, uint32 flags);
+
+
+static int CitusNoticeLogLevel = DEFAULT_CITUS_NOTICE_LEVEL;
 
 
 /*
@@ -663,6 +668,8 @@ StartConnectionEstablishment(ConnectionHashKey *key)
 	 */
 	PQsetnonblocking(connection->pgConn, true);
 
+	SetCitusNoticeProcessor(connection);
+
 	return connection;
 }
 
@@ -765,4 +772,84 @@ AfterXactHostConnectionHandling(ConnectionHashEntry *entry, bool isCommit)
 			UnclaimConnection(connection);
 		}
 	}
+}
+
+
+/*
+ * SetCitusNoticeProcessor sets the NoticeProcessor to DefaultCitusNoticeProcessor
+ */
+void
+SetCitusNoticeProcessor(MultiConnection *connection)
+{
+	PQsetNoticeProcessor(connection->pgConn, DefaultCitusNoticeProcessor,
+						 connection);
+}
+
+
+/*
+ * SetCitusNoticeLevel is used to set the notice level for distributed
+ * queries.
+ */
+void
+SetCitusNoticeLevel(int level)
+{
+	CitusNoticeLogLevel = level;
+}
+
+
+/*
+ * UnsetCitusNoticeLevel sets the CitusNoticeLogLevel back to
+ * its default value.
+ */
+void
+UnsetCitusNoticeLevel()
+{
+	CitusNoticeLogLevel = DEFAULT_CITUS_NOTICE_LEVEL;
+}
+
+
+/*
+ * DefaultCitusNoticeProcessor is used to redirect worker notices
+ * from logfile to console.
+ */
+static void
+DefaultCitusNoticeProcessor(void *arg, const char *message)
+{
+	MultiConnection *connection = (MultiConnection *) arg;
+	char *nodeName = connection->hostname;
+	uint32 nodePort = connection->port;
+	char *chompedMessage = pchomp(message);
+	char *trimmedMessage = TrimLogLevel(chompedMessage);
+	char *level = strtok((char *) message, ":");
+
+	ereport(CitusNoticeLogLevel, (errmsg("%s", trimmedMessage),
+								  errdetail("%s from %s:%d",
+											level, nodeName, nodePort)));
+}
+
+
+/*
+ * TrimLogLevel makes a copy of the string with the leading log level
+ * and spaces removed such as
+ *      From:
+ *          INFO:  "normal2_102070": scanned 0 of 0 pages...
+ *      To:
+ *          "normal2_102070": scanned 0 of 0 pages...
+ */
+char *
+TrimLogLevel(const char *message)
+{
+	size_t n;
+
+	n = 0;
+	while (n < strlen(message) && message[n] != ':')
+	{
+		n++;
+	}
+
+	do {
+		n++;
+	} while (n < strlen(message) && message[n] == ' ');
+
+	return pstrdup(message + n);
 }
--- a/src/backend/distributed/transaction/multi_shard_transaction.c
+++ b/src/backend/distributed/transaction/multi_shard_transaction.c
@@ -81,7 +81,7 @@ OpenTransactionsForAllTasks(List *taskList, int connectionFlags)
 		else
 		{
 			/* can only open connections for DDL and DML commands */
-			Assert(task->taskType == DDL_TASK);
+			Assert(task->taskType == DDL_TASK || VACUUM_ANALYZE_TASK);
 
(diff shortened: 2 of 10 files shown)


src/backend/distributed/transaction/multi_shard_transaction.c
```diff
@@ -133,6 +135,11 @@ OpenTransactionsForAllTasks(List *taskList, int connectionFlags)
 			 * transaction to fail.
 			 */
 			MarkRemoteTransactionCritical(connection);
+
+			if (task->taskType == VACUUM_ANALYZE_TASK)
+			{
+				PQsetNoticeReceiver(connection->pgConn, &VerboseNoticeReceiver, NULL);
```
marcocitus: Does this have any side-effects for connections which are cached? (e.g. router query; VACUUM; router query with notice?).

furkansahin:
@marcocitus , this PR is ready for review

src/backend/distributed/connection/connection_management.c
```diff
@@ -725,6 +731,8 @@ AfterXactHostConnectionHandling(ConnectionHashEntry *entry, bool isCommit)
 		MultiConnection *connection =
 			dlist_container(MultiConnection, connectionNode, iter.cur);
 
+		UnsetCitusNoticeLevel();
```
marcocitus: This seems like a weird place to put this. It gets called for every connection and seems unrelated to closing connections. Should it be in the abort handler?

marcocitus: (`CoordinatedTransactionCallback`)

src/backend/distributed/connection/connection_management.c
```diff
@@ -764,3 +772,54 @@ AfterXactHostConnectionHandling(ConnectionHashEntry *entry, bool isCommit)
...
+SetCitusNoticeProcessor(MultiConnection *connection)
+{
+	StringInfo nodeAddress = makeStringInfo();
+	char *hostName = connection->hostname;
+	uint32 nodePort = connection->port;
+
+	appendStringInfo(nodeAddress, "%s:%d", hostName, nodePort);
+
+	PQsetNoticeProcessor(connection->pgConn, DefaultCitusNoticeProcessor,
+						 MemoryContextStrdup(ConnectionContext, nodeAddress->data));
```
marcocitus: Allocating it in `ConnectionContext` means this will not be freed until the end of the session. Maybe just use the `MultiConnection` as the context pointer instead of a separate string?

src/test/regress/expected/multi_utilities.out
```diff
@@ -446,7 +446,32 @@ SELECT 1 FROM citus_create_restore_point('regression-test');
...
+VACUUM (FULL, VERBOSE) dustbunnies;
+INFO:  vacuuming "public.dustbunnies"
+INFO:  "dustbunnies": found 0 removable, 0 nonremovable row versions in 0 pages
+DETAIL:  0 dead row versions cannot be removed yet.
+CPU: user: 0.00 s, system: 0.00 s, elapsed: 0.00 s.
+INFO:  FROM localhost:57638 MESSAGE: INFO:  vacuuming "public.dustbunnies_990002"
+
+INFO:  FROM localhost:57638 MESSAGE: INFO:  "dustbunnies_990002": found 0 removable, 6 nonremovable row versions in 1 pages
+DETAIL:  0 dead row versions cannot be removed yet.
+CPU: user: 0.00 s, system: 0.00 s, elapsed: 0.00 s.
```
marcocitus: This'll probably break travis a lot. I think VERBOSE is probably not something for which we can really add a regression test.

src/test/regress/expected/multi_utilities.out
```diff
@@ -446,7 +446,32 @@ SELECT 1 FROM citus_create_restore_point('regression-test');
...
+
+INFO:  FROM localhost:57637 MESSAGE: INFO:  vacuuming "public.dustbunnies_990002"
+
+INFO:  FROM localhost:57637 MESSAGE: INFO:  "dustbunnies_990002": found 0 removable, 6 nonremovable row versions in 1 pages
```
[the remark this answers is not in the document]
marcocitus: I also wonder whether we should move the hostname into CONTEXT, since that's what we do everywhere else. Or maybe not show it at all.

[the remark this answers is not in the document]
furkansahin: AFAIK CONTEXT is invoked when there is an ERROR. We do not error out for vacuum verbose. I also think that it would be beneficial to provide it for debugging purposes if we want to use this mechanism for other notices as well. So instead of not showing, and putting it into context, I want to put it into Detail. Does this make sense?

src/test/regress/expected/multi_utilities.out
furkansahin: Than I propose that we should add (VACUUM | ANALYZE) VERBOSE to our testing document consistently. Maybe auto-testing tool can handle this?

src/backend/distributed/transaction/transaction_management.c
```diff
@@ -312,6 +312,8 @@ CoordinatedTransactionCallback(XactEvent event, void *arg)
 			break;
 		}
 	}
+
+	UnsetCitusNoticeLevel();
```
marcocitus: We only need to do this on abort, since otherwise we would have followed the normal flow, so maybe put it at line 240.

marcocitus: and also in the `SUBXACT_EVENT_ABORT_SUB` handler

src/backend/distributed/connection/connection_management.c
```diff
@@ -766,3 +772,53 @@ AfterXactHostConnectionHandling(ConnectionHashEntry *entry, bool isCommit)
...
+ * from logfile to console.
+ */
+static void
+DefaultCitusNoticeProcessor(void *arg, const char *message)
+{
+	MultiConnection *connection = (MultiConnection *) arg;
+	char *nodeName = connection->hostname;
+	uint32 nodePort = connection->port;
+
+	ereport(CitusNoticeLogLevel, (errmsg("%s", message),
```
marcocitus: it seems we put an extra newline after messages. Could we remove it like we do in `ReportResultError`? 

Would also be nice to remove the "INFO: " prefix from the message, since otherwise we'll give that twice.

src/backend/distributed/connection/connection_management.c
```diff
@@ -766,3 +772,80 @@ AfterXactHostConnectionHandling(ConnectionHashEntry *entry, bool isCommit)
...
+/*
+ * TrimLogLevel makes a copy of the string with the leading log level
+ * and spaces removed such as
+ * 		From:
+ * 		 	INFO:  "normal2_102070": scanned 0 of 0 pages...
+ * 		To:
+ * 			"normal2_102070": scanned 0 of 0 pages...
+ */
+char *
+TrimLogLevel(const char *in)
```
marcocitus: Maybe `in` -> `message`

src/backend/distributed/connection/connection_management.c
```diff
@@ -766,3 +772,80 @@ AfterXactHostConnectionHandling(ConnectionHashEntry *entry, bool isCommit)
...
+DefaultCitusNoticeProcessor(void *arg, const char *message)
+{
+	MultiConnection *connection = (MultiConnection *) arg;
+	char *nodeName = connection->hostname;
+	uint32 nodePort = connection->port;
+	char *chompedMessage = pchomp(message);
+	char *trimmedMessage = TrimLogLevel(chompedMessage);
+
+	ereport(CitusNoticeLogLevel, (errmsg("%s", trimmedMessage),
+								  errdetail("while executing command on %s:%d\n",
```
marcocitus: there should not be `\n` at the end here.

One minor nitpicky thing: The "while executing command" sounds a bit weird in the middle of all the output. The phrasing is based on getting errors, so you'd see "ERROR...while executing command". Maybe we could say "message from %s:%d" or something like that. 

(if you want to go all crazy, you could include the original log level here, e.g. "info from localhost:9700")

furkansahin:
@marcocitus this PR has the same problem with #2136 . I want to merge. What do you think?

marcocitus:
I think you can go ahead.

PR #2123, opened 2018-04-24
Improve connection error reporting

marcocitus:
DESCRIPTION: Add specialised error codes for connection failures
DESCRIPTION: Improve error messages on connection failure

We had a request to add a meaningful error code for connection failures, such that they could be distinguished from other types of failure and may require immediate action (e.g. switching to a different cluster in a very high availability scenario). I used ERRCODE_CONNECTION_FAILURE, since Postgres only uses that when it loses connection to the client, meaning clients would never see such a code in other cases.

While at it, I decided to address two more common complaints about Citus' error reporting on connection failures.

We throw "failure on connection marked as essential" errors in many failure cases after showing a warning containing a description of the problem. Since ORMs typically only record errors, the user does not know what happened. I noticed we very often do `ReportConnectionError` + `MarkRemoteTransactionFailed`. I combined these into a single function which can use the original error message from the client.

Finally, we sometimes showed "connection pointer is NULL" in the detail when a connection timeout occurred because in that cased we passed a `NULL` `PGconn *` to `PQerrorMessage`. For now I removed the detail in that case.

### DIFF
--- a/src/backend/distributed/transaction/remote_transaction.c
+++ b/src/backend/distributed/transaction/remote_transaction.c
@@ -107,8 +107,9 @@ StartRemoteTransactionBegin(struct MultiConnection *connection)
 
 	if (!SendRemoteCommand(connection, beginAndSetDistributedTransactionId->data))
 	{
-		ReportConnectionError(connection, WARNING);
-		MarkRemoteTransactionFailed(connection, true);
+		const bool raiseErrors = true;
+
+		HandleRemoteTransactionConnectionError(connection, raiseErrors);
 	}
 }
 
@@ -187,7 +188,7 @@ void
 StartRemoteTransactionCommit(MultiConnection *connection)
 {
 	RemoteTransaction *transaction = &connection->remoteTransaction;
-	const bool dontRaiseError = false;
+	const bool raiseErrors = false;
 	const bool isCommit = true;
 
 	/* can only commit if transaction is in progress */
@@ -225,8 +226,7 @@ StartRemoteTransactionCommit(MultiConnection *connection)
 
 		if (!SendRemoteCommand(connection, command.data))
 		{
-			ReportConnectionError(connection, WARNING);
-			MarkRemoteTransactionFailed(connection, dontRaiseError);
+			HandleRemoteTransactionConnectionError(connection, raiseErrors);
 
 			WarnAboutLeakedPreparedTransaction(connection, isCommit);
 		}
@@ -244,8 +244,7 @@ StartRemoteTransactionCommit(MultiConnection *connection)
 			 * Failing in this state means that we don't know whether the the
 			 * commit has succeeded.
 			 */
-			ReportConnectionError(connection, WARNING);
-			MarkRemoteTransactionFailed(connection, dontRaiseError);
+			HandleRemoteTransactionConnectionError(connection, raiseErrors);
 		}
 	}
 }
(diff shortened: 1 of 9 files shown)


src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -225,8 +226,7 @@ StartRemoteTransactionCommit(MultiConnection *connection)
 
 		if (!SendRemoteCommand(connection, command.data))
 		{
-			ReportConnectionError(connection, WARNING);
-			MarkRemoteTransactionFailed(connection, dontRaiseError);
+			HandleRemoteTransactionConnectionError(connection, dontRaiseError);
```
jasonmp85: `dontRaiseError` is confusing, given all the other uses use `raiseError`; can you make it consistent? I'm not a huge fan of this `const bool` stuff anyhow (IMO everyone should be using an editor capable of showing them the function declaration), but if we must, making them positive when true and negative when false will add more confusion.

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -244,8 +244,7 @@ StartRemoteTransactionCommit(MultiConnection *connection)
 			 * Failing in this state means that we don't know whether the the
 			 * commit has succeeded.
 			 */
-			ReportConnectionError(connection, WARNING);
-			MarkRemoteTransactionFailed(connection, dontRaiseError);
+			HandleRemoteTransactionConnectionError(connection, dontRaiseError);
```
jasonmp85: Gah, there are apparently a lot of these? Can you make them consistent with the function declaration, i.e. a boolean whose value represents whether to raise errors (this looks like a boolean whose value represents whether to not raise errors).

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -639,6 +633,53 @@ RemoteTransactionsBeginIfNecessary(List *connectionList)
 }
 
 
+/*
```
jasonmp85: :confused: there's a _lot_ of similarity in these two methods and I worry about them diverging, but I suppose this is OK (apparently `ReportConnectionError` and `ReportResultError` were already broken apart in the past?)

src/backend/distributed/transaction/remote_transaction.c
marcocitus: Yep, the double negative bugged me too. Fixing.

src/backend/distributed/transaction/remote_transaction.c
marcocitus: There is, but `ReportResultError` is wildly different from `ReportConnectionError` and takes a `PGresult *`.

marcocitus:
Uff, so apparently Linux says "Name or service not known" while Windows says "Unknown host", and that's causing appveyor to fail.

jasonmp85:
> Uff, so apparently Linux says "Name or service not known" while Windows says "Unknown host"

Oh yeah, IIRC macOS/BSD has a different error message, too.

marcocitus:
It seems the failing tests already was already setting a GUC to avoid exactly this problem. Added a `\set VERBOSITY terse` to hide the detail.

jasonmp85:
You good to merge, then?

marcocitus:
Just waiting for the green light from AppVeyor.

jasonmp85:
Any clue what's up with appveyor?

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -426,6 +423,9 @@ FinishRemoteTransactionAbort(MultiConnection *connection)
 		if (!IsResponseOK(result))
 		{
 			const bool isCommit = false;
+
+			HandleRemoteTransactionResultError(connection, result, raiseErrors);
```
marcocitus: Thanks for adding this back.

jasonmp85: I didn't add it, right? It was added in your change, I just put it where it needed to go in the conflict region.

PR #2124, opened 2018-04-24
Support UPDATE/DELETE with joins and subqueries

velioglu:
DESCRIPTION: Adds support for pushdownable subqueries and joins in distributed UPDATE/DELETE

With this PR we begin to support modify queries with multiple range table entries if these queries are pushdownable. Following items have been implemented as part of the PR.

* Check whether the modify query with multiple range table entries is pushdownable in `ModifyQuerySupported` function.
* Use `QueryPushdownSqlTaskList` for modify queries with multiple range table entries. 
* Within `QueryPushdownTaskCreate`, set task's anchor shard id as the id of the target table to obtain the correct lock on relation and select tables.
* Change `insertSelectQuery` flag to `modifyMultipleTableQuery`. In addition to INSERT ... SELECT query, msud with multiple tables also access multiple table (We need to use that flag in `RequiresConsistentSnapshot` function)
* Modify queries with multiples RTEs will not be supported with `master_modify_multiple_shards` UDF.
* Instead of `RelationRestrictionContext`, `PlannerRestrictionContext` will be passed to `CreaterRouterPlan`, since we need to apply check over multiple tables for MSUD with multiple tables in the router planner path.

### DIFF
--- a/src/backend/distributed/executor/multi_router_executor.c
+++ b/src/backend/distributed/executor/multi_router_executor.c
@@ -310,8 +310,7 @@ AcquireExecutorMultiShardLocks(List *taskList)
 		/*
 		 * If the task has a subselect, then we may need to lock the shards from which
 		 * the query selects as well to prevent the subselects from seeing different
-		 * results on different replicas. In particular this prevents INSERT..SELECT
-		 * commands from having different effects on different placements.
+		 * results on different replicas.
 		 */
 
 		if (RequiresConsistentSnapshot(task))
@@ -330,18 +329,17 @@ AcquireExecutorMultiShardLocks(List *taskList)
 
 /*
  * RequiresConsistentSnapshot returns true if the given task need to take
- * the necessary locks to ensure that a subquery in the INSERT ... SELECT
- * query returns the same output for all task placements.
+ * the necessary locks to ensure that a subquery in the modify query
+ * returns the same output for all task placements.
  */
 static bool
 RequiresConsistentSnapshot(Task *task)
 {
 	bool requiresIsolation = false;
 
-	if (!task->insertSelectQuery)
+	if (!task->modifyWithSubquery)
 	{
 		/*
-		 * Only INSERT/SELECT commands currently require SELECT isolation.
 		 * Other commands do not read from other shards.
 		 */
 
--- a/src/backend/distributed/master/master_modify_multiple_shards.c
+++ b/src/backend/distributed/master/master_modify_multiple_shards.c
@@ -151,7 +151,7 @@ master_modify_multiple_shards(PG_FUNCTION_ARGS)
 	{
 		bool multiShardQuery = true;
 		DeferredErrorMessage *error =
-			ModifyQuerySupported(modifyQuery, modifyQuery, multiShardQuery);
+			ModifyQuerySupported(modifyQuery, modifyQuery, multiShardQuery, NULL);
 
 		if (error)
 		{
--- a/src/backend/distributed/planner/insert_select_planner.c
+++ b/src/backend/distributed/planner/insert_select_planner.c
@@ -44,8 +44,8 @@ static DistributedPlan * CreateDistributedInsertSelectPlan(Query *originalQuery,
 														   plannerRestrictionContext);
 static Task * RouterModifyTaskForShardInterval(Query *originalQuery,
 											   ShardInterval *shardInterval,
-											   RelationRestrictionContext *
-											   restrictionContext,
+											   PlannerRestrictionContext *
+											   plannerRestrictionContext,
 											   uint32 taskIdIndex,
 											   bool allRelationsJoinedOnPartitionKey);
 static DeferredErrorMessage * DistributedInsertSelectSupported(Query *queryTree,
@@ -252,14 +252,14 @@ CreateDistributedInsertSelectPlan(Query *originalQuery,
 		Task *modifyTask = NULL;
 
 		modifyTask = RouterModifyTaskForShardInterval(originalQuery, targetShardInterval,
-													  relationRestrictionContext,
+													  plannerRestrictionContext,
 													  taskIdIndex,
 													  allDistributionKeysInQueryAreEqual);
 
 		/* add the task if it could be created */
 		if (modifyTask != NULL)
 		{
-			modifyTask->insertSelectQuery = true;
+			modifyTask->modifyWithSubquery = true;
 
 			sqlTaskList = lappend(sqlTaskList, modifyTask);
 		}
@@ -404,7 +404,7 @@ DistributedInsertSelectSupported(Query *queryTree, RangeTblEntry *insertRte,
  */
 static Task *
 RouterModifyTaskForShardInterval(Query *originalQuery, ShardInterval *shardInterval,
-								 RelationRestrictionContext *restrictionContext,
+								 PlannerRestrictionContext *plannerRestrictionContext,
 								 uint32 taskIdIndex,
 								 bool safeToPushdownSubquery)
 {
@@ -417,8 +417,8 @@ RouterModifyTaskForShardInterval(Query *originalQuery, ShardInterval *shardInter
 	Oid distributedTableId = shardInterval->relationId;
 	DistTableCacheEntry *cacheEntry = DistributedTableCacheEntry(distributedTableId);
 
-	RelationRestrictionContext *copiedRestrictionContext =
-		CopyRelationRestrictionContext(restrictionContext);
+	PlannerRestrictionContext *copyOfPlannerRestrictionContext = palloc0(
+		sizeof(PlannerRestrictionContext));
 
 	StringInfo queryString = makeStringInfo();
 	ListCell *restrictionCell = NULL;
@@ -431,11 +431,22 @@ RouterModifyTaskForShardInterval(Query *originalQuery, ShardInterval *shardInter
 	List *intersectedPlacementList = NULL;
 	bool upsertQuery = false;
 	bool replacePrunedQueryWithDummy = false;
-	bool allReferenceTables = restrictionContext->allReferenceTables;
+	bool allReferenceTables =
+		plannerRestrictionContext->relationRestrictionContext->allReferenceTables;
 	List *shardOpExpressions = NIL;
 	RestrictInfo *shardRestrictionList = NULL;
 	DeferredErrorMessage *planningError = NULL;
 	bool multiShardModifyQuery = false;
+	List *relationRestrictionList = NIL;
+
+	copyOfPlannerRestrictionContext->relationRestrictionContext =
+		CopyRelationRestrictionContext(
+			plannerRestrictionContext->relationRestrictionContext);
+	copyOfPlannerRestrictionContext->joinRestrictionContext =
+		plannerRestrictionContext->joinRestrictionContext;
+	relationRestrictionList =
+		copyOfPlannerRestrictionContext->relationRestrictionContext->
+		relationRestrictionList;
 
 	/* grab shared metadata lock to stop concurrent placement additions */
 	LockShardDistributionMetadata(shardId, ShareLock);
@@ -444,7 +455,7 @@ RouterModifyTaskForShardInterval(Query *originalQuery, ShardInterval *shardInter
 	 * Replace the partitioning qual parameter value in all baserestrictinfos.
 	 * Note that this has to be done on a copy, as the walker modifies in place.
 	 */
-	foreach(restrictionCell, copiedRestrictionContext->relationRestrictionList)
+	foreach(restrictionCell, relationRestrictionList)
 	{
 		RelationRestriction *restriction = lfirst(restrictionCell);
 		List *originalBaseRestrictInfo = restriction->relOptInfo->baserestrictinfo;
@@ -493,7 +504,7 @@ RouterModifyTaskForShardInterval(Query *originalQuery, ShardInterval *shardInter
 	 * If we can, we also rely on the side-effects that all RTEs have been updated
 	 * to point to the relevant nodes and selectPlacementList is determined.
 	 */
-	planningError = PlanRouterQuery(copiedSubquery, copiedRestrictionContext,
+	planningError = PlanRouterQuery(copiedSubquery, copyOfPlannerRestrictionContext,
 									&selectPlacementList, &selectAnchorShardId,
 									&relationShardList, replacePrunedQueryWithDummy,
 									&multiShardModifyQuery);
--- a/src/backend/distributed/planner/multi_physical_planner.c
+++ b/src/backend/distributed/planner/multi_physical_planner.c
@@ -127,15 +127,12 @@ static ArrayType * SplitPointObject(ShardInterval **shardIntervalArray,
 static bool DistributedPlanRouterExecutable(DistributedPlan *distributedPlan);
 static Job * BuildJobTreeTaskList(Job *jobTree,
 								  PlannerRestrictionContext *plannerRestrictionContext);
-static List * QueryPushdownSqlTaskList(Query *query, uint64 jobId,
-									   RelationRestrictionContext *
-									   relationRestrictionContext,
-									   List *prunedRelationShardList, TaskType taskType);
 static void ErrorIfUnsupportedShardDistribution(Query *query);
 static Task * QueryPushdownTaskCreate(Query *originalQuery, int shardIndex,
 									  RelationRestrictionContext *restrictionContext,
 									  uint32 taskId,
-									  TaskType taskType);
+									  TaskType taskType,
+									  bool modifyRequiresMasterEvaluation);
 static bool ShardIntervalsEqual(FmgrInfo *comparisonFunction,
 								ShardInterval *firstInterval,
 								ShardInterval *secondInterval);
@@ -201,6 +198,7 @@ static StringInfo IntermediateTableQueryString(uint64 jobId, uint32 taskIdIndex,
 static uint32 FinalTargetEntryCount(List *targetEntryList);
 static bool CoPlacedShardIntervals(ShardInterval *firstInterval,
 								   ShardInterval *secondInterval);
+
 
 /*
  * CreatePhysicalDistributedPlan is the entry point for physical plan generation. The
@@ -2008,7 +2006,8 @@ BuildJobTreeTaskList(Job *jobTree, PlannerRestrictionContext *plannerRestriction
 			sqlTaskList = QueryPushdownSqlTaskList(job->jobQuery, job->jobId,
 												   plannerRestrictionContext->
 												   relationRestrictionContext,
-												   prunedRelationShardList, SQL_TASK);
+												   prunedRelationShardList, SQL_TASK,
+												   false);
 		}
 		else
 		{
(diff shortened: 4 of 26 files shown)


velioglu:
@marcocitus @onderkalaci PR is ready for the review. Beside the implementation, I have some questions :

1) As the comment of `AcquireMetadataLocks` it is stated that 
> we assume tasks to be already sorted by shard ID such that deadlocks are avoided. This is true for INSERT/SELECT, which is the only multi-shard command right now

However, after this PR modify queries with multiple range table entries also create multi-shard queries. Up to that point in the executor, tasks have not been sorted by the shard ID if we call it from modify query (not INSERT ... SELECT) stack. TaskList will be sorted with `FirstReplicaAssignTaskList` which is called just after `AcquireMetadataLocks`. So, can you see any problem if we change the order of calling these two functions in `CitusModifyBeginScan` or should I sort the taskList in the router planner ?  

2) If the modify query multiple range table entries need to be evaluated on the master, we can't evaluate the function. You can see simple query [here](https://github.com/citusdata/citus/pull/2124/files#diff-81eb343533a6009c86a1aa45a5ade03dR469). We hit that error because `Param` is provided to `ExecInitExprRec` function as `PARAM_SUBLINK` yet the comment 

> PARAM_SUBLINK: This type of Param is converted to PARAM_EXEC during planning.

Since we use the original query we still have the `Param` as sublink. How should I convert the paramkind of the param? Or should I limit the functionality of such queries in this PR ?

3) Is there any other case you want to add to the regression tests ?

onderkalaci:
>  So, can you see any problem if we change the order of calling these two functions in CitusModifyBeginScan or should I sort the taskList in the router planner ?

That seems reasonable and we should probably have done that earlier. You should maybe expand the comment `* modify tasks are always assigned using first-replica policy */` to also mention that we rely on the sorted task list as well.
```C
/* some comment */
taskList = FirstReplicaAssignTaskList(taskList);

AcquireMetadataLocks(taskList);
...

workerJob->taskList  = taskList;
```

> Is there any other case you want to add to the regression tests ?

I'll test it today and let you know if you miss any crazy test cases 😏

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1672,10 +1728,12 @@ PlanRouterQuery(Query *originalQuery, RelationRestrictionContext *restrictionCon
 	uint64 shardId = INVALID_SHARD_ID;
 	CmdType commandType = originalQuery->commandType;
 	bool isMultiShardModifyQuery = false;
+	bool isMultipleTableModifyQuery = false;
 
 	*placementList = NIL;
 	prunedRelationShardList = TargetShardIntervalsForQuery(originalQuery,
```
onderkalaci: This looks really hard to read. Can you add a newline before `TargetShardIntervalsForQuery `

onderkalaci:
> I've actually observed two issues during my initial testing. First, we seem to be not detecting outer joins properly. We should have prevented below, right?

Sorry, PostgreSQL pulls that query. It looks like we do the correct checks, but, could you please add some tests for that similar to the ones you've added for subquery + reference tables ?:
```SQL

UPDATE users_table
SET value_1 = 4
WHERE user_id IN
    (SELECT DISTINCT e2.user_id
     FROM users_table_ref
     LEFT JOIN users_table e2 ON (e2.user_id = users_table_ref.value_1)) RETURNING *;
```

onderkalaci:
Unfortunately the error messages are pretty bad for this case since we don't support recursive planning, we should somehow omit `WHEN a subquery REFERENCES a COLUMN FROM another query`:
```SQL
UPDATE users_table
SET value_1 = 4
WHERE user_id IN
    (SELECT DISTINCT e2.user_id
     FROM users_table_ref
     RIGHT JOIN users_table e2 ON (e2.user_id = users_table_ref.value_1)
     LIMIT 4) RETURNING *;

ERROR: cannot push down this subquery DETAIL:
LIMIT IN subquery IS currently unsupported WHEN a subquery REFERENCES a COLUMN
FROM another query
```

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -453,7 +458,6 @@ CitusModifyBeginScan(CustomScanState *node, EState *estate, int eflags)
 	 */
 	LockPartitionsInRelationList(distributedPlan->relationIdList, AccessShareLock);
 
-	/* modify tasks are always assigned using first-replica policy */
 	workerJob->taskList = FirstReplicaAssignTaskList(taskList);
```
onderkalaci: Why do we call `FirstReplicaAssignTaskList ` again?

src/backend/distributed/planner/insert_select_planner.c
```diff
@@ -418,7 +418,11 @@ RouterModifyTaskForShardInterval(Query *originalQuery, ShardInterval *shardInter
 	DistTableCacheEntry *cacheEntry = DistributedTableCacheEntry(distributedTableId);
 
 	RelationRestrictionContext *copiedRestrictionContext =
-		CopyRelationRestrictionContext(restrictionContext);
+		CopyRelationRestrictionContext(
+			plannerRestrictionContext->relationRestrictionContext);
+	RelationRestrictionContext *originalRestrictionContext =
```
onderkalaci: Why do we change `RouterModifyTaskForShardInterval ()` at all? I cannot see this being used? Can't we only change the parameters when this function is called?

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -687,23 +689,56 @@ ModifyQuerySupported(Query *queryTree, Query *originalQuery, bool multiShardQuer
...
-		if (!UpdateOrDeleteQuery(queryTree) || multiShardQuery)
+		/* We can not get restriction context via master_modify_multiple_shards path */
+		if (plannerRestrictionContext == NULL)
 		{
 			return DeferredError(ERRCODE_FEATURE_NOT_SUPPORTED,
-								 "cannot perform distributed planning for the given "
-								 "modification",
-								 "Joins are not supported in distributed "
-								 "modifications.", NULL);
+								 "cannot run multi shard modify query with master_modify_multiple_shard",
```
onderkalaci: Maybe `cannot run multi shard modify query with master_modify_multiple_shard when the query involves subquery or join` ? We generally try to give more useful information in the error messages.

`Send your query directly.` -> `Execute the query without using master_modify_multiple_shard()` ? (I'm not sure about this, maybe let's wait @marcocitus )

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -444,6 +442,13 @@ CitusModifyBeginScan(CustomScanState *node, EState *estate, int eflags)
 		RebuildQueryStrings(jobQuery, taskList);
 	}
 
+	/*
+	 * Modify tasks are always assigned using first-replica policy. Since tasks
+	 * are also sorted by anchor shard id within this function, we do not need
+	 * to sort them again while obtaining locks.
+	 */
+	taskList = FirstReplicaAssignTaskList(taskList);
```
marcocitus: This is too early to assign tasks to placements since we haven't obtained the metadata locks yet.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1694,8 +1752,14 @@ PlanRouterQuery(Query *originalQuery, RelationRestrictionContext *restrictionCon
 
 		Assert(UpdateOrDeleteQuery(originalQuery));
 
+		isMultipleTableModifyQuery = (list_length(
+										  plannerRestrictionContext->
+										  relationRestrictionContext->
+										  relationRestrictionList) > 1);
```
marcocitus: we usually break this down into an `if` statement

src/backend/distributed/executor/multi_router_executor.c
velioglu: Should I sort them in the planner ?

marcocitus: why do you need to sort them?

velioglu: To obtain locks in the correct order within `AcquireMetadataLocks`  (https://github.com/citusdata/citus/pull/2124#issuecomment-384171739)

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -453,8 +458,7 @@ CitusModifyBeginScan(CustomScanState *node, EState *estate, int eflags)
 	 */
 	LockPartitionsInRelationList(distributedPlan->relationIdList, AccessShareLock);
 
-	/* modify tasks are always assigned using first-replica policy */
-	workerJob->taskList = FirstReplicaAssignTaskList(taskList);
+	workerJob->taskList = taskList;
```
marcocitus: There should probably have been a comment, but changing this line breaks the rebalancer.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -687,23 +690,48 @@ ModifyQuerySupported(Query *queryTree, Query *originalQuery, bool multiShardQuer
...
+		/* We can not get restriction context via master_modify_multiple_shards path */
+		if (plannerRestrictionContext == NULL)
 		{
 			return DeferredError(ERRCODE_FEATURE_NOT_SUPPORTED,
-								 "cannot perform distributed planning for the given "
-								 "modification",
-								 "Joins are not supported in distributed "
-								 "modifications.", NULL);
+								 "cannot run multi shard modify query with master_modify_multiple_shard when the query involves subquery or join",
+								 "Execute the query without using master_modify_multiple_shard()",
```
marcocitus: _shard -> shards

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -687,23 +690,48 @@ ModifyQuerySupported(Query *queryTree, Query *originalQuery, bool multiShardQuer
 	}
 
 	/*
-	 * Reject queries which involve joins. Note that UPSERTs are exceptional for this case.
-	 * Queries like "INSERT INTO table_name ON CONFLICT DO UPDATE (col) SET other_col = ''"
-	 * contains two range table entries, and we have to allow them.
+	 * We have to allow modify queries with two range table entries, if it is pushdownable.
 	 */
 	if (commandType != CMD_INSERT && queryTableCount != 1)
```
marcocitus: As discussed on slack, the `queryTableCount != 1` does not belong here because we can have subqueries without having multiple tables. It would make sense to move it into the `if` statement below, since it does apply to `master_modify_multiple_shards`

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1694,8 +1742,16 @@ PlanRouterQuery(Query *originalQuery, RelationRestrictionContext *restrictionCon
...
+		if (list_length(
+				plannerRestrictionContext->relationRestrictionContext->
+				relationRestrictionList) > 1)
+		{
+			isMultipleTableModifyQuery = true;
+		}
+
 		planningError = ModifyQuerySupported(originalQuery, originalQuery,
-											 isMultiShardQuery);
+											 isMultipleTableModifyQuery,
```
marcocitus: As discussed on slack, we're passing the wrong variable here. `isMultipleTableModifyQuery` should be `isMultiShardQuery` and we don't need `isMultipleTableModifyQuery` here at all

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1446,8 +1477,24 @@ RouterJob(Query *originalQuery, RelationRestrictionContext *restrictionContext,
...
+
+		/* If it is a modify task with multiple tables */
+		if (list_length(
+				plannerRestrictionContext->relationRestrictionContext->
+				relationRestrictionList) > 1)
+		{
+			ListCell *taskCell = NULL;
+			foreach(taskCell, job->taskList)
+			{
+				Task *task = (Task *) lfirst(taskCell);
```
onderkalaci: Can't we move this logic into `QueryPushdownSqlTaskList ()` and apply this when the task is `ModifyTask`? That seems a better place, no?

marcocitus:
should we merge #2137 into this branch?

onderkalaci:
> should we merge #2137 into this branch?

I'd prefer not to do that for smaller/simpler reviews. We could merge them together to the master when we're all done?

velioglu:
@marcocitus I have added full outer join with USING expression check and updated the code according to your comments

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -2345,14 +2410,38 @@ QueryPushdownTaskCreate(Query *originalQuery, int shardIndex,
...
+	}
 
-	/* and generate the full query string */
-	pg_get_query_def(taskQuery, queryString);
-	ereport(DEBUG4, (errmsg("distributed statement: %s", queryString->data)));
+	subqueryTask = CreateBasicTask(jobId, taskId, taskType, NULL);
+
+	if (UpdateOrDeleteQuery(taskQuery))
+	{
+		if (HasDangerousJoinUsing(taskQuery->rtable, (Node *) taskQuery->jointree))
```
marcocitus: it seems a bit weird to do this for each individual task in the planning function, rather than once in the `ModifyQuerySupported` function.

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -2284,6 +2319,21 @@ QueryPushdownTaskCreate(Query *originalQuery, int shardIndex,
 	List *selectPlacementList = NIL;
 	uint64 jobId = INVALID_JOB_ID;
 	uint64 anchorShardId = INVALID_SHARD_ID;
+	bool modifyWithSubselect = false;
+	RangeTblEntry *resultRangeTable = NULL;
+	Oid resultRelationOid = InvalidOid;
+
+	/*
+	 * If it is a modify query with sub-select, we need to have result relation
+	 * to obtain right lock on it.
```
marcocitus: this comment is a bit misleading since we don't actually have any special logic to obtain a lock on the result relation, rather on all the other relations.

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -2314,8 +2364,21 @@ QueryPushdownTaskCreate(Query *originalQuery, int shardIndex,
...
+				if (relationId == resultRelationOid)
+				{
+					anchorShardId = shardInterval->shardId;
+				}
+			}
+			else
+			{
+				/* use a shard from a distributed table as the anchor shard */
+				anchorShardId = shardInterval->shardId;
+			}
```
marcocitus: the nesting here is a bit confusing, we could maybe make the code a bit more self-documenting
```c
else if (UpdateOrDeleteQuery(originalQuery))
{
    if (!modifyWithSubquery || relationId == resultRelationOid)
    {
        /* for UPDATE/DELETE the shard in the result relation becomes the anchor shard */
        anchorShardId = shardInterval->shardId;
    }
}
else 
{
    /* for SELECT we pick an arbitrary shard as the anchor shard */
    anchorShardId = shardInterval->shardId;
}
```
could also introduce a boolean variable to avoid calling `UpdateOrDeleteQuery`

marcocitus: Could also consider always initialising the `resultRelationOid` for `UpdateOrDeleteQuery` to further improve readability.

src/test/regress/sql/with_modifying.sql
```diff
@@ -221,13 +221,15 @@ SELECT * FROM modify_table ORDER BY id, val;
 SELECT * FROM anchor_table ORDER BY id;
 
 INSERT INTO modify_table VALUES (11, 1), (12, 2), (13, 3);
-WITH select_data AS (
+
+-- This test will be enabled after recursive planner is enabled for modify query.
+/*WITH select_data AS (
```
marcocitus: Maybe just leave the test as is, but with an output change. That'll make sure that we don't forget to re-enable it in the recursive planning PR.

src/include/distributed/multi_physical_planner.h
```diff
@@ -179,7 +179,7 @@ typedef struct Task
 	bool upsertQuery;              /* only applies to modify tasks */
 	char replicationModel;         /* only applies to modify tasks */
 
-	bool insertSelectQuery;
+	bool modifyWithSubquery;       /* true for multi-shard INSERT/UPDATE/DELETE queries */
```
marcocitus: I know we haven't added very descriptive comments in this struct, but could we add one above saying what we use this for?

src/test/regress/expected/multi_explain.out
```diff
@@ -816,6 +816,87 @@ Custom Scan (Citus Router)
...
+  Tasks Shown: All
+  ->  Task
+        Node: host=localhost port=57637 dbname=regression
+        ->  Update on lineitem_hash_part_360041 lineitem_hash_part
+              ->  Hash Join
+                    Hash Cond: (lineitem_hash_part.l_orderkey = orders_hash_part.o_orderkey)
+                    ->  Seq Scan on lineitem_hash_part_360041 lineitem_hash_part
+                    ->  Hash
+                          ->  Seq Scan on orders_hash_part_360045 orders_hash_part
+  ->  Task
```
marcocitus: do we need to show more than one task?

src/include/distributed/multi_physical_planner.h
```diff
@@ -179,7 +179,7 @@ typedef struct Task
 	bool upsertQuery;              /* only applies to modify tasks */
 	char replicationModel;         /* only applies to modify tasks */
 
-	bool insertSelectQuery;
+	bool modifyWithSubquery;
```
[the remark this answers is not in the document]
velioglu: Added

src/test/regress/expected/multi_shard_update_delete.out
```diff
@@ -361,6 +347,300 @@ NOTICE:  Copying data from local table...
...
+FROM   events_test_table_2
+WHERE  users_test_table.user_id = events_test_table_2.user_id;
+ERROR:  cannot push down this subquery
+DETAIL:  Shards of relations in subquery need to have 1-to-1 shard partitioning
+-- Should error out due to multiple row return from subquery, but we can not get this information within 
+-- subquery pushdown planner. This query will be sent to worker with recursive planner.
+DELETE FROM users_test_table
+WHERE  users_test_table.user_id = (SELECT user_id
+                                   FROM   events_test_table);
+ERROR:  complex joins are only supported when all distributed tables are joined on their distribution columns with equal operator
```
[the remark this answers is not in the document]
velioglu: We can not get partition key equality from plannerRestrictionContext for this query, as we can not get it for corresponding Select query as well. We are planning corresponding Select query recursively and get the error mentioned in the comment from worker nodes. Test result should be changed after recursive dml pr get merged.

src/backend/distributed/master/master_modify_multiple_shards.c
```diff
@@ -149,9 +149,9 @@ master_modify_multiple_shards(PG_FUNCTION_ARGS)
 
 	if (modifyQuery->commandType != CMD_UTILITY)
 	{
-		bool multiShardQuery = true;
+		bool multiShardQuery = false;
```
marcocitus: why this change? seems a bit confusing

marcocitus: I know, but this line makes no sense to me. Why can't `multiShardQuery` be true?

velioglu: Otherwise, we need to have planner restriction context for multi shard queries (not multi table) sent by `master_modify_multiple_shards`. We can remain it as it is using the last commit I have sent.

PR #2137, opened 2018-04-27
Implement recursive planning for DML statements

marcocitus:
DESCRIPTION: Support non-pushdownable subqueries and CTEs in distributed UPDATE/DELETE

It's pretty simple to implement recursive planning for DML statements on top of Burak's branch. This solves the problem that UPDATE/DELETE may contain subqueries that we don't consider to be pushdownable, which we not normally handle through recursive planning. We currently only expect to hit the error messages for non-pushdownable subqueries when there are correlated subqueries and changing them will be a challenge. It's actually much simpler to just implement recursive planning.

but....this enables *really a lot* of new SQL support and thus needs *really a lot* of tests, but throwing it out there as a suggestion (that we'll probably want anyway).

### DIFF
--- a/src/backend/distributed/planner/distributed_planner.c
+++ b/src/backend/distributed/planner/distributed_planner.c
@@ -573,6 +573,7 @@ CreateDistributedPlan(uint64 planId, Query *originalQuery, Query *query, ParamLi
 	DistributedPlan *distributedPlan = NULL;
 	MultiTreeRoot *logicalPlan = NULL;
 	List *subPlanList = NIL;
+	bool hasCtes = originalQuery->cteList != NIL;
 
 	if (IsModifyCommand(originalQuery))
 	{
@@ -670,8 +671,13 @@ CreateDistributedPlan(uint64 planId, Query *originalQuery, Query *query, ParamLi
 	 * We could simplify this code if the logical planner was capable of dealing
 	 * with an original query. In that case, we would only have to filter the
 	 * planner restriction context.
+	 *
+	 * Note that we check both for subplans and whether the query had CTEs
+	 * prior to calling GenerateSubplansForSubqueriesAndCTEs. If none of
+	 * the CTEs are referenced then there are no subplans, but we still want
+	 * to retry the router planner.
 	 */
-	if (list_length(subPlanList) > 0)
+	if (list_length(subPlanList) > 0 || hasCtes)
 	{
 		Query *newQuery = copyObject(originalQuery);
 		bool setPartitionedTablesInherited = false;
--- a/src/test/regress/expected/dml_recursive.out
+++ b/src/test/regress/expected/dml_recursive.out
@@ -1,5 +1,6 @@
 CREATE SCHEMA recursive_dml_queries;
 SET search_path TO recursive_dml_queries, public;
+SET citus.next_shard_id TO 2370000;
 CREATE TABLE recursive_dml_queries.distributed_table (tenant_id text, dept int, info jsonb);
 SELECT create_distributed_table('distributed_table', 'tenant_id');
  create_distributed_table 
@@ -274,8 +275,7 @@ DEBUG:  subqueries are not supported within INSERT queries
 HINT:  Try rewriting your queries with 'INSERT INTO ... SELECT' syntax.
 ERROR:  subqueries are not supported within INSERT queries
 HINT:  Try rewriting your queries with 'INSERT INTO ... SELECT' syntax.
--- we error out of the CTE is not referenced
--- by any part of the query (e.g., cte2 is not referenced)
+-- DML with an unreferenced SELECT CTE
 WITH cte_1 AS (
     WITH cte_2 AS (
         SELECT tenant_id as cte2_id 
@@ -294,7 +294,49 @@ WHERE distributed_table.tenant_id < cte_1.tenant_id;
 DEBUG:  common table expressions are not supported in distributed modifications
 DEBUG:  generating subplan 20_1 for CTE cte_1: WITH cte_2 AS (SELECT second_distributed_table.tenant_id AS cte2_id FROM recursive_dml_queries.second_distributed_table WHERE (second_distributed_table.dept >= 2)) UPDATE recursive_dml_queries.distributed_table SET dept = 10 RETURNING tenant_id, dept, info
 DEBUG:  common table expressions are not supported in distributed modifications
-ERROR:  common table expressions are not supported in distributed modifications
+DEBUG:  Plan 20 query after replacing subqueries and CTEs: UPDATE recursive_dml_queries.distributed_table SET dept = 5 FROM (SELECT intermediate_result.tenant_id, intermediate_result.dept, intermediate_result.info FROM read_intermediate_result('20_1'::text, 'binary'::citus_copy_format) intermediate_result(tenant_id text, dept integer, info jsonb)) cte_1 WHERE (distributed_table.tenant_id < cte_1.tenant_id)
+EXPLAIN (COSTS FALSE) WITH cte_1 AS (
+    WITH cte_2 AS (
+        SELECT tenant_id as cte2_id 
+        FROM second_distributed_table 
+        WHERE dept >= 2
+    )
+    
+    UPDATE distributed_table 
+    SET dept = 10
+    RETURNING *
+)
+UPDATE distributed_table
+SET dept = 5
+FROM cte_1
+WHERE distributed_table.tenant_id < cte_1.tenant_id;
+DEBUG:  common table expressions are not supported in distributed modifications
+DEBUG:  generating subplan 22_1 for CTE cte_1: WITH cte_2 AS (SELECT second_distributed_table.tenant_id AS cte2_id FROM recursive_dml_queries.second_distributed_table WHERE (second_distributed_table.dept >= 2)) UPDATE recursive_dml_queries.distributed_table SET dept = 10 RETURNING tenant_id, dept, info
+DEBUG:  common table expressions are not supported in distributed modifications
+DEBUG:  Plan 22 query after replacing subqueries and CTEs: UPDATE recursive_dml_queries.distributed_table SET dept = 5 FROM (SELECT intermediate_result.tenant_id, intermediate_result.dept, intermediate_result.info FROM read_intermediate_result('22_1'::text, 'binary'::citus_copy_format) intermediate_result(tenant_id text, dept integer, info jsonb)) cte_1 WHERE (distributed_table.tenant_id < cte_1.tenant_id)
+                                           QUERY PLAN                                           
+------------------------------------------------------------------------------------------------
+ Custom Scan (Citus Router)
+   ->  Distributed Subplan 22_1
+         ->  Custom Scan (Citus Router)
+               Task Count: 4
+               Tasks Shown: One of 4
+               ->  Task
+                     Node: host=localhost port=57638 dbname=regression
+                     ->  Update on distributed_table_2370000 distributed_table
+                           ->  Seq Scan on distributed_table_2370000 distributed_table
+   Task Count: 4
+   Tasks Shown: One of 4
+   ->  Task
+         Node: host=localhost port=57638 dbname=regression
+         ->  Update on distributed_table_2370000 distributed_table
+               ->  Nested Loop
+                     Join Filter: (distributed_table.tenant_id < intermediate_result.tenant_id)
+                     ->  Function Scan on read_intermediate_result intermediate_result
+                     ->  Materialize
+                           ->  Seq Scan on distributed_table_2370000 distributed_table
+(19 rows)
+
 -- we don't support updating local table with a join with
 -- distributed tables
 UPDATE 
--- a/src/test/regress/sql/dml_recursive.sql
+++ b/src/test/regress/sql/dml_recursive.sql
@@ -1,5 +1,6 @@
 CREATE SCHEMA recursive_dml_queries;
 SET search_path TO recursive_dml_queries, public;
+SET citus.next_shard_id TO 2370000;
 
 CREATE TABLE recursive_dml_queries.distributed_table (tenant_id text, dept int, info jsonb);
 SELECT create_distributed_table('distributed_table', 'tenant_id');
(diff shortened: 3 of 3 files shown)


src/backend/distributed/planner/distributed_planner.c
```diff
@@ -597,31 +576,62 @@ CreateDistributedSelectPlan(uint64 planId, Query *originalQuery, Query *query,
...
+
+		/* the functions above always return a plan, possibly with an error */
+		Assert(distributedPlan);
+
 		if (distributedPlan->planningError == NULL)
 		{
-			/* successfully created a router plan */
 			return distributedPlan;
 		}
 		else
```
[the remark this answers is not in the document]
marcocitus: > Open question: Do you see a simple way to not try recursive planning if we're sure that the query wouldn't be planned at all?

We're not doing anything special, just checking for CTEs and subqueries, which is pretty cheap and queries that don't work usually don't become a peformance bottleneck. The fact that we have a DEBUG message and an ERROR message is just an artifact from going through `RaiseDeferredError(distributedPlan->planningError, DEBUG1);`. 

We could consider immediately throwing an error if the problem cannot be resolved through recursive planning or filling in parameters.

> Another difference would be in the use of volatile functions, we'd be providing some workarounds. Do you see any problems with that?:

Not really, I guess we could allow the `value_2 * random()` in case of RF=1 (though not easily), but we can't do any workaround because it needs to be evaluated for every row. Conversely `(SELECT random())` resolves to the same value for all rows, so recursively planning it is correct.

onderkalaci:
One interesting issue I've found, not directly related to this PR, is the following:

```SQL
WITH ids_to_upsert AS 
(
	SELECT tenant_id FROM distributed_table WHERE dept > 7
)
INSERT INTO distributed_table  
       SELECT distributed_table.tenant_id FROM ids_to_upsert, distributed_table
       		WHERE  distributed_table.tenant_id = ids_to_upsert.tenant_id
       	ON CONFLICT (tenant_id) DO UPDATE SET dept = 8
 RETURNING *;

```

The reason is that we have some [hard](https://github.com/citusdata/citus/blob/master/src/backend/distributed/planner/insert_select_planner.c#L503) errors in the `INSERT .. SELECT` planning, which prevents any fallback to recursive planning.

I think we should tackle that separately, since changing that might have other consequences. 

We could consider this if we spend some time for unifying `INSERT .. SELECT` planning with query pushdown planning. I'll add a failing test for now.

marcocitus:
Agreed, unifying INSERT..SELECT logic might be a good 8.0 tech debt item.

onderkalaci:
One another limitation I hit due to executor/connection mgnmt:
```SQL
-- not a very meaningful query
-- but has three modifying CTEs along with another
-- modify statement at the end
WITH copy_to_other_table AS (
    INSERT INTO distributed_table 
        SELECT *
            FROM second_distributed_table
        WHERE dept = 3
        ON CONFLICT (tenant_id) DO UPDATE SET dept = 4
        RETURNING *
),
main_table_deleted AS (
    DELETE 
    FROM distributed_table 
    WHERE dept < 10
      AND NOT EXISTS (SELECT 1 FROM second_distributed_table 
                      WHERE second_distributed_table.dept = 1
                        AND second_distributed_table.tenant_id = distributed_table.tenant_id)
                        RETURNING *     

),
second_table_recursive AS
(
	UPDATE second_distributed_table SET dept = 51 FROM 
	(
		SELECT tenant_id FROM second_distributed_table WHERE dept != 2
	) as foo WHERE foo.tenant_id != second_distributed_table.tenant_id RETURNING second_distributed_table.*
)
INSERT INTO second_distributed_table
        SELECT *
            FROM main_table_deleted
        EXCEPT 
        SELECT *
            FROM copy_to_other_table
        EXCEPT 
        SELECT * 
        	FROM second_table_recursive;
DEBUG:  distributed INSERT ... SELECT can only select from distributed tables
DEBUG:  data-modifying statements are not supported in the WITH clauses of distributed queries
DEBUG:  generating subplan 137_1 for CTE copy_to_other_table: INSERT INTO with_modifying.distributed_table (tenant_id, dept) SELECT tenant_id, dept FROM with_modifying.second_distributed_table WHERE (dept = 3) ON CONFLICT(tenant_id) DO UPDATE SET dept = 4 RETURNING distributed_table.tenant_id, distributed_table.dept
DEBUG:  generating subplan 137_2 for CTE main_table_deleted: DELETE FROM with_modifying.distributed_table WHERE ((dept < 10) AND (NOT (EXISTS (SELECT 1 FROM with_modifying.second_distributed_table WHERE ((second_distributed_table.dept = 1) AND (second_distributed_table.tenant_id = distributed_table.tenant_id)))))) RETURNING tenant_id, dept
DEBUG:  generating subplan 137_3 for CTE second_table_recursive: UPDATE with_modifying.second_distributed_table SET dept = 51 FROM (SELECT second_distributed_table_1.tenant_id FROM with_modifying.second_distributed_table second_distributed_table_1 WHERE (second_distributed_table_1.dept <> 2)) foo WHERE (foo.tenant_id <> second_distributed_table.tenant_id) RETURNING second_distributed_table.tenant_id, second_distributed_table.dept
DEBUG:  complex joins are only supported when all distributed tables are joined on their distribution columns with equal operator
DEBUG:  generating subplan 140_1 for subquery SELECT tenant_id FROM with_modifying.second_distributed_table WHERE (dept <> 2)
DEBUG:  Plan 140 query after replacing subqueries and CTEs: UPDATE with_modifying.second_distributed_table SET dept = 51 FROM (SELECT intermediate_result.tenant_id FROM read_intermediate_result('140_1'::text, 'binary'::citus_copy_format) intermediate_result(tenant_id text)) foo WHERE (foo.tenant_id <> second_distributed_table.tenant_id) RETURNING second_distributed_table.tenant_id, second_distributed_table.dept
DEBUG:  generating subplan 137_4 for subquery SELECT main_table_deleted.tenant_id, main_table_deleted.dept FROM (SELECT intermediate_result.tenant_id, intermediate_result.dept FROM read_intermediate_result('137_2'::text, 'binary'::citus_copy_format) intermediate_result(tenant_id text, dept integer)) main_table_deleted EXCEPT SELECT copy_to_other_table.tenant_id, copy_to_other_table.dept FROM (SELECT intermediate_result.tenant_id, intermediate_result.dept FROM read_intermediate_result('137_1'::text, 'binary'::citus_copy_format) intermediate_result(tenant_id text, dept integer)) copy_to_other_table EXCEPT SELECT second_table_recursive.tenant_id, second_table_recursive.dept FROM (SELECT intermediate_result.tenant_id, intermediate_result.dept FROM read_intermediate_result('137_3'::text, 'binary'::citus_copy_format) intermediate_result(tenant_id text, dept integer)) second_table_recursive
DEBUG:  Plan 137 query after replacing subqueries and CTEs: SELECT tenant_id, dept FROM (SELECT intermediate_result.tenant_id, intermediate_result.dept FROM read_intermediate_result('137_4'::text, 'binary'::citus_copy_format) intermediate_result(tenant_id text, dept integer)) citus_insert_select_subquery
ERROR:  cannot establish a new connection for placement 371, since DML has been executed on a connection that is in use
```

As far as I can see this is not related to this PR but rather a more generic issue. So, I won't be adding this to the tests but posting here to (i) inform @marcocitus  (ii) future reference.

onderkalaci:
> Subqueries with local tables

Local tables in subquery fail due to some checks in `ModifyQuerySupported()`. Do you think would it worth fixing that? We should probably defer error if there is a non distributed table?

```SQL
-- the query is nonsense, just a valid SQL
UPDATE 
	distributed_table 
SET 
	info =  row_to_json(row(avg_tenant_id))
FROM 
(
	SELECT 
		avg(local_table.id::int) as avg_tenant_id
	FROM 
		local_table
) as foo
WHERE
	foo.avg_tenant_id::int::text = distributed_table.tenant_id
RETURNING
	distributed_table.*;
ERROR:  relation local_table is not distributed
```

bt:
```
#0  errstart (elevel=20, filename=0x104a3e9c3 "utils/metadata_cache.c", lineno=713, funcname=0x104a3eb32 "DistributedTableCacheEntry", domain=0x0) at elog.c:251
#1  0x0000000104a00d56 in DistributedTableCacheEntry (distributedRelationId=17041) at utils/metadata_cache.c:713
#2  0x00000001049dc749 in ModifyQuerySupported (queryTree=0x7f96330c4420, originalQuery=0x7f96330c4308, multiShardQuery=0 '\0', plannerRestrictionContext=0x7f96330fe390) at planner/multi_router_planner.c:601
#3  0x00000001049dc3b1 in CreateModifyPlan (originalQuery=0x7f96330c4308, query=0x7f96330c4420, plannerRestrictionContext=0x7f96330fe390) at planner/multi_router_planner.c:194
#4  0x00000001049c0128 in CreateDistributedSelectPlan (planId=289, originalQuery=0x7f96330c4308, query=0x7f96330c4420, boundParams=0x0, hasUnresolvedParams=0 '\0', plannerRestrictionContext=0x7f96330fe390) at planner/distributed_planner.c:592
#5  0x00000001049bee9e in CreateDistributedPlan (planId=289, localPlan=0x7f96331717f8, originalQuery=0x7f96330c4308, query=0x7f96330c4420, boundParams=0x0, plannerRestrictionContext=0x7f96330fe390) at planner/distributed_planner.c:493
#6  0x00000001049bea20 in distributed_planner (parse=0x7f96330c4420, cursorOptions=256, boundParams=0x0) at planner/distributed_planner.c:149
```

marcocitus:
> Do you think would it worth fixing that? We should probably defer error if there is a non distributed table?

agreed, let me fix that

onderkalaci:
We have the `ContextContainsLocalRelation(RelationRestrictionContext *restrictionContext)` if that works in `relation_restriction_eq.c`

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -534,6 +536,18 @@ ModifyQuerySupported(Query *queryTree, Query *originalQuery, bool multiShardQuer
 	Node *onConflictWhere = NULL;
 	CmdType commandType = queryTree->commandType;
 
+	RelationRestrictionContext *relationRestrictionContext =
```
onderkalaci: Ops, this fails to catch `VIEW`s because they don't appear in the restriction context.

So, using a VIEW instead of local table fails now.

src/test/regress/expected/with_transactions.out
```diff
@@ -0,0 +1,102 @@
...
+BEGIN;
+	WITH ids_inserted AS
+		(
+			INSERT INTO raw_table VALUES (11, 1000, now()) RETURNING tenant_id 
+		)
+		UPDATE raw_table SET created_at = '2001-02-10 20:00:00' WHERE tenant_id IN (SELECT tenant_id FROM ids_inserted);
+DEBUG:  common table expressions are not supported in distributed modifications
+DEBUG:  generating subplan 8_1 for CTE ids_inserted: INSERT INTO with_transactions.raw_table (tenant_id, income, created_at) VALUES (11, 1000, now()) RETURNING tenant_id
+DEBUG:  Plan 8 query after replacing subqueries and CTEs: UPDATE with_transactions.raw_table SET created_at = 'Sat Feb 10 20:00:00 2001 PST'::timestamp with time zone WHERE (tenant_id IN (SELECT ids_inserted.tenant_id FROM (SELECT intermediate_result.tenant_id FROM read_intermediate_result('8_1'::text, 'binary'::citus_copy_format) intermediate_result(tenant_id integer)) ids_inserted))
+	TRUNCATE second_raw_table;
```
onderkalaci: I realized that we've missed running `SELECT`s inside the transactions.

Could we add the following two  (or any other queries) `SELECT` statements in this transction:
```SQL

SELECT count (*) FROM second_raw_table;
SELECT count(DISTINCT created_at) FROM raw_table;
```

src/test/regress/expected/dml_recursive.out
```diff
@@ -0,0 +1,317 @@
...
+    RETURNING *
+)
+UPDATE distributed_table
+SET dept = 5
+FROM cte_1
+WHERE distributed_table.tenant_id < cte_1.tenant_id;
+DEBUG:  common table expressions are not supported in distributed modifications
+DEBUG:  generating subplan 20_1 for CTE cte_1: WITH cte_2 AS (SELECT second_distributed_table.tenant_id AS cte2_id FROM recursive_dml_queries.second_distributed_table WHERE (second_distributed_table.dept >= 2)) UPDATE recursive_dml_queries.distributed_table SET dept = 10 RETURNING tenant_id, dept, info
+DEBUG:  common table expressions are not supported in distributed modifications
+ERROR:  common table expressions are not supported in distributed modifications
```
onderkalaci: In general `common table expressions are not supported in distributed modifications` seems not a very accurate error message anymore. 

As far as I can see, it could only happen when a `SELECT` cte is not referenced at all.

Shall we update the error message accordingly? 

(I'm OK not fixing it as well since it is a very minor thing)

src/test/regress/expected/recursive_dml_with_different_planners_executors.out
```diff
@@ -0,0 +1,87 @@
...
+-- joined with a real-time query
+UPDATE 
+	distributed_table 
+SET dept = foo.dept FROM 
+	(SELECT tenant_id, dept FROM second_distributed_table WHERE dept = 1 ) as foo, 
+	(SELECT tenant_id FROM second_distributed_table WHERE dept IN (1, 2, 3, 4) OFFSET 0) as bar
+	WHERE foo.tenant_id = bar.tenant_id
+	AND distributed_table.tenant_id = bar.tenant_id; 
+DEBUG:  cannot push down this subquery
+DETAIL:  Offset clause is currently unsupported when a subquery references a column from another query
```
onderkalaci: Minor: Since we `DEBUG1`  all deferred error, it might be confusing as here:`Offset clause is currently unsupported when a subquery references a column from another query`

But, seems hard to fix this.

src/test/regress/multi_schedule
```diff
@@ -32,7 +32,7 @@ test: multi_read_from_secondaries
 test: multi_create_table
 test: multi_create_table_constraints multi_master_protocol multi_load_data multi_behavioral_analytics_create_table
 test: multi_behavioral_analytics_basics multi_behavioral_analytics_single_shard_queries multi_insert_select_non_pushable_queries multi_insert_select
-test: multi_insert_select_window multi_shard_update_delete window_functions
+test: multi_insert_select_window multi_shard_update_delete window_functions dml_recursive recursive_dml_with_different_planners_executors
```
onderkalaci: I think the only test we've intended and not added is `EXPLAIN`. So, maybe add a single explain test for completeness?

src/test/regress/expected/recursive_dml_with_different_planners_executors.out
onderkalaci: Btw, I personally prefer having the slightly off `DEBUG1` message over not having it at all.

src/test/regress/expected/dml_recursive.out
marcocitus: Fixed the underlying issue, so these queries work now.

 We may want to clean up debug messages at some point.

src/test/regress/multi_schedule
marcocitus: added

PR #2138, opened 2018-04-29
Make DROP TABLE connect as current user instead of superuser

marcocitus:
DESCRIPTION: DROP TABLE now drops shards as the currrent user instead of the superuser

Currently, when you issue a DROP TABLE on a distributed table, the coordinator connects to the workers as superuser to drop shards. The main reason is that the `citus_drop_trigger` function is a `SECURITY DEFINER` function and it needs to be because it is a PL/pgSQL function that queries `pg_dist_partition`, to which the user dropping the table may not have access.

However, this has been causing a variety of issues, especially on Cloud where users cannot connect as superuser. A `DROP TABLE` typically cannot be used inside a transaction block that access a co-located table, because it would grab locks as the current user, but then try to drop a shard as the superuser, which causes "cannot establish a new connection for placement" errors. This situation also occurs if a regular user tries to create a distributed table and drops it in the same transaction, as a way of creating a temporary distributed table. It also seems risky to let a regular user perform actions as superuser, and it means the user cannot e.g. cancel `DROP TABLE` commands on the workers.

This PR addresses the issue by re-defining `citus_drop_trigger` to not query `pg_dist_partition`, but instead having the UDFs it calls check whether the table is distributed and silently ignore calls when it's not. Permission checks will have already been performed at this point, since the trigger is called from the Postgres code that drops the object on the coordinator. If the table still exists (the user calls the function directly), permission checks are still performed.

I looked at redefining the whole trigger function in C, but it seems it's not easy to retrieve the details of the drop event in C.


metdos:
Hey @jasonmp85, if you have time, could you also review this PR, if not let me know. Thanks.

src/backend/distributed/citus--7.4-1--7.4-2.sql
```diff
@@ -0,0 +1,35 @@
+/* citus--7.4-1--7.4-2 */
```
jasonmp85: This is going to drive me to rewrite our whole upgrade script infrastructure :-|

jasonmp85: (Where _this_ == an inability to see the git history of objects we define in our upgrade scripts)

src/backend/distributed/commands/drop_distributed_table.c
```diff
@@ -38,9 +39,19 @@ master_drop_distributed_table_metadata(PG_FUNCTION_ARGS)
...
-	EnsureCoordinator();
 	CheckCitusVersion(ERROR);
 
+	/*
+	 * The SQL_DROP trigger calls this function even for tables that are
+	 * not distributed. In that case, silently ignore and return -1.
+	 */
+	if (!IsDistributedTable(relationId) || !EnableDDLPropagation)
+	{
+		PG_RETURN_INT32(-1);
```
jasonmp85: This is a void function: what does it mean to return `-1`?

jasonmp85:
This looks good, modulo the part where you added a `-1` return to a `void` function; just silently return I suppose.

lithp:
This PR broke the build on Windows. Here's the appveyor output: https://ci.appveyor.com/project/lithp/citus/build/1.0.1171-663c9b2-appveyor

The unsupported_worker_operations and transactional_drop_shards tests were broken by this PR.

Issue #2136, opened 2018-04-26
Appveyor is flaky


lithp:
`multi_real_time_transaction` sometimes fails, claiming every line has changed, and sometimes succeeds. It seems to happen randomly (restarting the build can cause it to pass).

Here's an example pass: https://ci.appveyor.com/project/lithp/citus/build/1.0.800-pg11_compat?fullLog=true
And an example fail: https://ci.appveyor.com/project/lithp/citus/build/1.0.798-pg11_compat?fullLog=true

They seem to be the exact same until the end.

lithp:
Maybe we pass or fail depending on which build worker handles the build?

lithp:
It's apparently possible to remote into the worker to diagnose problems: https://www.appveyor.com/docs/how-to/rdp-to-build-worker/

lithp:
I added a primitive hexdump to the test output. The line endings are different on `multi_real_time_transaction`, but they're also different on `with_basics`!

```
$s = Get-Content -Path .\expected\multi_real_time_transaction.out -Raw
$s.substring(0, 100) | Format-Hex
           00 01 02 03 04 05 06 07 08 09 0A 0B 0C 0D 0E 0F
00000000   53 45 54 20 63 69 74 75 73 2E 6E 65 78 74 5F 73  SET citus.next_s
00000010   68 61 72 64 5F 69 64 20 54 4F 20 31 36 31 30 30  hard_id TO 16100
00000020   30 30 3B 0A 43 52 45 41 54 45 20 53 43 48 45 4D  00;.CREATE SCHEM
00000030   41 20 6D 75 6C 74 69 5F 72 65 61 6C 5F 74 69 6D  A multi_real_tim
00000040   65 5F 74 72 61 6E 73 61 63 74 69 6F 6E 3B 0A 53  e_transaction;.S
00000050   45 54 20 73 65 61 72 63 68 5F 70 61 74 68 20 3D  ET search_path =
00000060   20 27 6D 75                                       'mu            
$s = Get-Content -Path .\results\multi_real_time_transaction.out -Raw
$s.substring(0, 100) | Format-Hex
           00 01 02 03 04 05 06 07 08 09 0A 0B 0C 0D 0E 0F
00000000   53 45 54 20 63 69 74 75 73 2E 6E 65 78 74 5F 73  SET citus.next_s
00000010   68 61 72 64 5F 69 64 20 54 4F 20 31 36 31 30 30  hard_id TO 16100
[...]

PR #2139, opened 2018-04-30
Fix appveyor newline issues

onderkalaci:
I'm not sure about this fix. I've mostly followed the discussion [here](https://help.appveyor.com/discussions/problems/1119-allow-changing-git-autocrlf-setting).

Any thoughts?

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -11,6 +11,9 @@
 
 version: 1.0.{build}-{branch}
 
+init:
+  - git config --global core.autocrlf true
+
 # Maximum number of concurrent jobs for the project
 max_jobs: 1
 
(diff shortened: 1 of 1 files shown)


lithp:
Sure, it's worth a shot! I don't expect that this will fix things, in #2136 it looked like the line endings were "wrong" for other tests too, but pg_regress_multi only failed one of the tests.

However, I really don't understand what's going on. This is an easy change and has a chance of fixing the tests, let's merge it!

onderkalaci:
ops. This failed again. So, probably it is not a good idea to merge?

onderkalaci:
This is not the right approach to fix the issue.

PR #2141, opened 2018-05-02
shard count for some of the tests are increased

furkansahin:
This PR increases shard_count in the following tests

- multi_alter_table_add_constraints.sql
- multi_foreign_key
- multi_modifications (not effected yet because it still uses `master_create_distributed_table` UDF)
- subquery_and_cte
- with_basics

### DIFF
--- a/src/test/regress/sql/multi_alter_table_add_constraints.sql
+++ b/src/test/regress/sql/multi_alter_table_add_constraints.sql
@@ -4,6 +4,7 @@
 -- Test checks whether constraints of distributed tables can be adjusted using
 -- the ALTER TABLE ... ADD CONSTRAINT ... command.
 
+SET citus.shard_count TO 32;
 SET citus.next_shard_id TO 1450000;
 SET citus.next_placement_id TO 1450000;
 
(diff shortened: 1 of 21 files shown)


mtuncer:
@furkansahin why we did this change ?

src/test/regress/sql/subquery_and_cte.sql
```diff
@@ -1,8 +1,7 @@
 -- ===================================================================
 -- test recursive planning functionality with subqueries and CTEs
 -- ===================================================================
-CREATE SCHEMA subquery_and_ctes;
-SET search_path TO subquery_and_ctes, public;
+SET search_path TO subquery_and_ctes;
```
mtuncer: perhaps you can drop subquery_and_ctes schema after the test

mtuncer: never mind, it was already there, not just in the diff

src/test/regress/sql/multi_alter_table_add_constraints.sql
```diff
@@ -417,11 +418,11 @@ SELECT "Constraint", "Definition" FROM table_checks WHERE relid='products'::regc
 
 \c - - - :worker_1_port
 
-SELECT "Constraint", "Definition" FROM table_checks WHERE relid='public.products_1450034'::regclass;
+SELECT "Constraint", "Definition" FROM table_checks WHERE relid='public.products_1450202'::regclass;
 
 \c - - - :master_port
-SET citus.next_shard_id TO 1450038;
-SET citus.next_placement_id TO 1450038;
+SET citus.next_shard_id TO 1450234;
```
mtuncer: I know they were already there, but why we have them in the first place ? can we drop it altogether ?

PR #2143, opened 2018-05-02
Create distributed table test update

furkansahin:
from master_Create_distributed_table to create_distributed_table
DESCRIPTION: tests are updated to use create_distributed_table


onderkalaci:
note that we've some tests in the enterprise repo that should be updated as well

furkansahin:
@onderkalaci I am inclined to leave it to another PR since this is already a huge one and updates most of the tests. We can update enterprise tests later, can't we?

src/test/regress/expected/multi_insert_select.out
```diff
@@ -75,19 +75,19 @@ WHERE
...
 ---------
-       1
-       5
        3
        4
-       6
+       1
+       5
        2
+       6
```
marcocitus: Seems the order is volatile.

PR #2147, opened 2018-05-03
Always prefix operators with their namespace.

pykello:
Previously we checked if an operator is in pg_catalog, and if it wasn't we prefixed it with namespace in worker queries. This can have a huge impact on performance of physical planner when using custom data types.

This happened regardless of current `search_path` config, because Citus overrides the search path in `get_query_def_extended()`. When we do so, the check for existence of the operator in current search path in `generate_operator_name()` fails for any operators outside `pg_catalog`. This means that nothing gets cached, and in the following calls we will again recheck the system tables for existence of the operators, which took an additional 40-50ms for some of the usecases we were seeing.

In this change we skip the pg_catalog check, and always prefix the operator with its namespace.


src/test/regress/expected/multi_repartition_join_planning.out
```diff
@@ -12,12 +12,16 @@ SHOW server_version \gset
 SELECT substring(:'server_version', '\d+')::int > 9 AS version_above_nine;
  version_above_nine 
 --------------------
- t
```
pykello: This and some other changes to this file don't make sense. I'll look into this.

src/test/regress/expected/multi_repartition_join_planning_0.out
```diff
@@ -12,16 +12,12 @@ SHOW server_version \gset
 SELECT substring(:'server_version', '\d+')::int > 9 AS version_above_nine;
  version_above_nine 
 --------------------
- f
```
pykello: The changes to this file don't make sense. I'll look into this.

src/test/regress/expected/multi_view.out
```diff
@@ -10,7 +10,7 @@ SHOW server_version \gset
 SELECT substring(:'server_version', '\d+')::int > 10 AS version_above_ten;
  version_above_ten 
 -------------------
- t
```
pykello: this doesn't make sense.

src/test/regress/expected/multi_view.out
```diff
@@ -797,8 +797,8 @@ RESET citus.subquery_pushdown;
 VACUUM ANALYZE users_table;
 -- explain tests
 EXPLAIN (COSTS FALSE) SELECT user_id FROM recent_selected_users GROUP BY 1 ORDER BY 1;
-                                                                 QUERY PLAN                                                                  
----------------------------------------------------------------------------------------------------------------------------------------------
```
pykello: Look into this.

jasonmp85:
(I'll review this after a phone screen I'm conducting).

onderkalaci:
> :frown: why is there so much debug output that prints SQL whyyyyy this should be restricted to a single test file 👎

I acknowledge your complaints on this.

That was a discussion for recursive queries (e.g., tests with `intermediate_result`) during the code review. The conclusion was if we don't print the worker SQL queries, we could miss any behavior changes during the planning such as the order of the subquery planning or pushdown vs pull-push planning.

Given that we've too many ways of recursive planning, we ended up with too many tests :/

jasonmp85:
> Honestly, in some of these test files, because of lack of comments, I'm not sure what exactly we are testing for.

Unfortunately that's not only limited to this file; I think the SQL-based style of testing in general suffers from this weakness, comments or no.

src/backend/distributed/utils/ruleutils_11.c
```diff
@@ -7837,45 +7836,14 @@ generate_operator_name(Oid operid, Oid arg1, Oid arg2)
...
-
-	if (p_result != NULL && oprid(p_result) == operid)
-		nspname = NULL;
-	else
-	{
-		nspname = get_namespace_name(operform->oprnamespace);
-		appendStringInfo(&buf, "OPERATOR(%s.", quote_identifier(nspname));
-	}
-
+	nspname = get_namespace_name(operform->oprnamespace);
```
jasonmp85: Is there any reason this might be `NULL`? If so, can you add an `Assert` here?

jasonmp85:
I'm assuming you reverted the changes to the version printing bits of the tests (those are in there because I got sick of not being able to tell which files corresponded to which version of PostgreSQL without running tests and manually comparing, so we have 9.6/10/11 printed in tests which differ).

This is a :shipit: if you can answer my `Assert` question (i.e. if you think it might be needed, add it, if you think it will never be hit, don't add it and just merge).

src/backend/distributed/utils/ruleutils_11.c
pykello: Unless user messes with `pg_operator` manually, this shouldn't be `NULL`. I added an assert here. Postgres' `generate_operator_name()` assumes this cannot be `NULL`. They don't have an `Assert()` here, but the following `quote_identifier(nspname)` will crash if this is `NULL`, so I guess they assume this is always not null.

metdos:
@marcocitus, @jasonmp85, @onderkalaci, @pykello should we backport this to 7.4 release?

onderkalaci:
> @marcocitus, @jasonmp85, @onderkalaci, @pykello should we backport this to 7.4 release?

I vote for yes. Because it seems a reasonable, small change. Also, this will help even more with complex queries involving lots of operators in the query. And, I think if we don't add it to 7.4 now, we'll need to do it later 😄

PR #2150, opened 2018-05-07
Check volatile functions in modify queries

velioglu:
Check volatile functions in the query tree of modify queries with multiple tables and joins.

### DIFF
--- a/src/test/regress/expected/multi_shard_update_delete_0.out
+++ b/src/test/regress/expected/multi_shard_update_delete_0.out
@@ -722,10 +722,26 @@ SET    value_2 = 5 * random()
 FROM   events_test_table
 WHERE  users_test_table.user_id = events_test_table.user_id;
 ERROR:  functions used in UPDATE queries on distributed tables must not be VOLATILE
+-- Recursive modify planner does not take care of following test because the query
+-- is fully pushdownable, yet not allowed because it would lead to inconsistent replicas.
+UPDATE users_test_table
+SET    value_2 = subquery.random FROM (SELECT user_id, random()
+                                       FROM events_test_table) subquery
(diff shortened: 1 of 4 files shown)


src/test/regress/sql/multi_shard_update_delete.sql
```diff
@@ -580,6 +580,11 @@ SET    value_2 = 5 * random()
 FROM   events_test_table
 WHERE  users_test_table.user_id = events_test_table.user_id;
 
+UPDATE users_test_table
```
marcocitus: Could use a comment, since this is a rather unusual case in the sense that recursive planning does not take care of it because the query is fully pushdownable, but not allowed because it would lead to inconsistent replicas.

it might also be interesting to add:
```sql
UPDATE users_test_table
SET    value_2 = subquery.random FROM (SELECT user_id, random()
                                       FROM events_test_table) subquery;

UPDATE users_test_table
SET    value_2 = subquery.random FROM (SELECT user_id, random()
                                       FROM events_test_table OFFSET 0) subquery
WHERE  users_test_table.user_id = subquery.user_id;
```

src/test/regress/expected/multi_shard_update_delete_0.out
```diff
@@ -766,7 +775,7 @@ DECLARE test_cursor CURSOR FOR SELECT * FROM users_test_table;
 FETCH test_cursor;
  user_id | value_1 | value_2 | value_3 
 ---------+---------+---------+---------
-       8 |       4 |      13 |       0
+       8 |       4 |       0 |       0
```
marcocitus: seems like this would be 1 about half the time (postgres casts double to int using rounding)

PR #2159, opened 2018-05-08
Fix the huge appveyor diffs

lithp:
I've finally figured out what's causing #2136. The regression tests run just fine, but `pg_regress_multi.pl` was setting an environment variable which caused `diff` to be passed the wrong flags while generating `regression.diffs`.

This means: every appveyor failure was a legitimate failure but the verbose diff output masked the error.


jasonmp85:
`diff` is sometimes a shortcut or alias to something other than the (I believe GNU) "normal" `diff`, so I wanted to test whether this would break with others (i.e. in my setup I avoided installing `msys` or any of that stuff, so I put `git-diff` on my path). It works. :shipit:!

PR #2161, opened 2018-05-09
Move call to RemoveIntermediateResultsDirectory

lithp:
Errors thrown in the COMMIT handler will cause Postgres to abort,
there's nothing it can do it abort the transaction by the time that
handler is called!

`RemoveIntermediateResultsDirectory` is problematic for two reasons:
- It has calls to ereport(ERROR which have been known to trigger
- It makes memory allocations which raise ERRORs when they fail

Once the COMMIT process has begun we don't use the intermediate results,
so it's safe to remove them a little earlier in the process. A failure
here will abort the transaction. That's pretty unnecessary, it's not
that important that we remove the results, but it's still better than a
crash.

(Meant to address https://github.com/citusdata/citus-enterprise/issues/231 and https://github.com/citusdata/citus/issues/2086)

### DIFF
--- a/src/backend/distributed/transaction/transaction_management.c
+++ b/src/backend/distributed/transaction/transaction_management.c
@@ -159,7 +159,6 @@ CoordinatedTransactionCallback(XactEvent event, void *arg)
 			 * transaction management. Do so before doing other work, so the
 			 * callbacks still can perform work if needed.
 			 */
-			RemoveIntermediateResultsDirectory();
 			ResetShardPlacementTransactionState();
 
 			if (CurrentCoordinatedTransactionState == COORD_TRANS_PREPARED)
@@ -256,6 +255,8 @@ CoordinatedTransactionCallback(XactEvent event, void *arg)
 
 		case XACT_EVENT_PRE_COMMIT:
 		{
+			RemoveIntermediateResultsDirectory();
+
 			/* nothing further to do if there's no managed remote xacts */
 			if (CurrentCoordinatedTransactionState == COORD_TRANS_NONE)
 			{


src/backend/distributed/transaction/transaction_management.c
```diff
@@ -256,6 +255,8 @@ CoordinatedTransactionCallback(XactEvent event, void *arg)
 
 		case XACT_EVENT_PRE_COMMIT:
 		{
+			RemoveIntermediateResultsDirectory();
+
```
marcocitus: We can probably move this below the if statement below to not bother non-citus transactions.

jasonmp85: Agreed; the function is basically a no-op if not in a Citus transaction but it's such a simple change to not call it needlessly I think it makes sense to move.

jasonmp85:
I think @marcocitus's feedback makes sense. I'm not terribly familiar with all cases that create intermediate results, but if they always trigger a Citus transaction state change (and it sounds like they do), I'd say move this call to right after the `break`.

lithp:
@marcocitus, I just dove into the code and I think it belongs above the break.

`RemoveIntermediateResultsDirectory` already has a guard, it checks `CreatedResultsDirectory` and no-ops if it doesn't exist, so we're not hurting anything by calling it more than necessary.

And it's true that we'll usually be inside a coordinated transaction if an intermediate result directory has been created (`create_intermediate_result` immediately calls `BeginCoordinatedTransaction`), but that feels like a malleable implementation detail that we shouldn't rely on, and it's not actually always true!

`CreateIntermediateResultsDirectory` is also called when you run `COPY "resultid" FROM STDIN WITH (format result)`. That doesn't necessarily happen inside a coordinated transaction, and if we waited until after the break then users would be able to DOS the server by sending a lot of results which fill up space and are never removed.

marcocitus:
> @marcocitus, I just dove into the code and I think it belongs above the break.

You're right. Thanks for the fix!

PR #2162, opened 2018-05-09
Alleviate OOM failures in COMMIT callback

lithp:
Previously those failures caused us to crash, postgres abort()s when it notices a failure in the COMMIT callback.

This resolves the problem I understand https://github.com/citusdata/citus-enterprise/issues/231 as being a symptom of.

### DIFF
--- a/src/backend/distributed/transaction/transaction_management.c
+++ b/src/backend/distributed/transaction/transaction_management.c
@@ -49,6 +49,9 @@ dlist_head InProgressTransactions = DLIST_STATIC_INIT(InProgressTransactions);
 /* stack of active sub-transactions */
 static List *activeSubXacts = NIL;
 
+/* some pre-allocated memory so we don't need to call malloc() during callbacks */
+MemoryContext CommitContext = NULL;
+
 /*
  * Should this coordinated transaction use 2PC? Set by
  * CoordinatedTransactionUse2PC(), e.g. if DDL was issued and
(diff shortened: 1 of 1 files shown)


src/backend/distributed/transaction/transaction_management.c
```diff
@@ -154,6 +164,17 @@ CoordinatedTransactionCallback(XactEvent event, void *arg)
...
+			/*
+			 * ERRORs thrown during XACT_EVENT_COMMIT will cause postgres to abort, at
+			 * this point enough work has been done that it's not possible to rollback.
+			 *
+			 * One possible source of errors is memory allocation failures. To minimize
+			 * the chance of those happening we've pre-allocated some memory in the
+			 * CommitContext, it has 8kb of memory that we're allowed to use.
+			 */
+			MemoryContext previousContext = CurrentMemoryContext;
+			MemoryContextSwitchTo(CommitContext);
```
marcocitus: should we also be doing this in the abort handler then?

jasonmp85:
Got set up to repro this yesterday; I'll get on actually reproing it today.

src/backend/distributed/transaction/transaction_management.c
lithp: No, because postgres [already sets aside memory](https://github.com/postgres/postgres/blob/master/src/backend/access/transam/xact.c#L990) for abort handling [and switches to it](https://github.com/postgres/postgres/blob/master/src/backend/access/transam/xact.c#L2454) when processing an abort.

jasonmp85: Ok, but why _just_ do this in `EVENT_COMMIT` and not also the other blocks here?

FWIW I think this fixes the problem from my testing. I can't give you repro instructions easily though (it'll be easier after Monday).

jasonmp85: (I guess you're going to say that in all other cases it's OK to roll back, but maybe add that detail in the comment for future readers)

lithp:
Okay, updated the comment and rebased, waiting for the tests to finish.

PR #2166, opened 2018-05-11
Fix regression tests in AppVeyor

onderkalaci:
And fix some regression test failures.

### DIFF
--- a/citus.control
+++ b/citus.control
@@ -1,6 +1,6 @@
 # Citus extension
 comment = 'Citus distributed database'
-default_version = '7.4-1'
+default_version = '7.4-3'
 module_pathname = '$libdir/citus'
 relocatable = false
 schema = pg_catalog
--- a/src/test/regress/expected/multi_agg_approximate_distinct_0.out
+++ b/src/test/regress/expected/multi_agg_approximate_distinct_0.out
@@ -87,6 +87,8 @@ ERROR:  cannot compute count (distinct) approximation
 HINT:  You need to have the hll extension loaded.
 -- Check that approximate count(distinct) works at a table in a schema other than public
 -- create necessary objects
+SET citus.next_shard_id TO 20000000;
+SET citus.next_placement_id TO 20000000;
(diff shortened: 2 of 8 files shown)


marcocitus:
Hmz, AppVeyor looks like it's still using an older schema version.

marcocitus:
there's a `citus.control` file in the root directory that still says 7.4-1

citus.control
```diff
@@ -1,6 +1,6 @@
 # Citus extension
 comment = 'Citus distributed database'
-default_version = '7.4-1'
+default_version = '7.4-3'
```
marcocitus: It'd be nice if there was only one of these, but in the absence of that we could maybe put a comment in both files pointing out that you need to change the version in the other one as well.

onderkalaci: I can add a comment, but, I think if the appveyor were not failing for some unrelated things, we could have realized issues due to version upgrades.

I think this is a great example that we should not let appveyor or any other CI to be in failing state.

marcocitus: The comment is just to save everyone who forgets this an hour.

onderkalaci: I was going to propose making `appveyor ` not optional for merging into master. Otherwise, we'll likely to get into the similar situation again?

(I can the comment as well those are not dependent 😄)

marcocitus: > I was going to propose making appveyor not optional for merging into master. Otherwise, we'll likely to get into the similar situation again?

Yep, seems fine.

> (I can add the comment as well those are not dependent 😄)

Yep, it's just kind of hard to figure out that the schema is stale (we should maybe add it to the output in `multi_extension`)

onderkalaci:
There are few more failing tests due to minor output changes. I'll try to fix those soon (I'll be OOTO on monday&tuesday) unless someone else wants to fix those 😈

citus.control
lithp: I've opened https://github.com/citusdata/citus/pull/2173 to help us remember to update `citus.control`.

PR #2167, opened 2018-05-11
Add a debug message with distribution column value

marcocitus:
DESCRIPTION: Show the distribution column value for router queries in a debug message

This PR adds a detail to the "Plan is router executable" debug message with the distribution column value. This is not in itself very useful, but adds some useful infrastructure for implementing landlord, which requires the ability to extract the distribution column value from a distributed plan. Doing so requires some careful code surgery, so putting this up for feedback.

### DIFF
--- a/src/test/regress/expected/multi_router_planner.out
+++ b/src/test/regress/expected/multi_router_planner.out
@@ -127,6 +127,7 @@ DEBUG:  Plan is router executable
 SELECT * FROM articles_hash WHERE author_id = 10 AND id = 50;
 DEBUG:  Creating router plan
 DEBUG:  Plan is router executable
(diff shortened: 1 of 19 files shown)


src/test/regress/expected/multi_router_planner.out
```diff
@@ -862,6 +898,7 @@ SELECT * FROM (
 ORDER BY id;
 DEBUG:  Creating router plan
 DEBUG:  Plan is router executable
+DETAIL:  distribution column value: 3
```
mtuncer: this is a multi-tenant query. We should not be picking arbitrary one here. 
Plus this query gives different results in mac vs linux

marcocitus:
bug fixes look good to me!

PR #2170, opened 2018-05-15
Bump Citus to 7.5devel

velioglu:


lithp:
@mtuncer I mean, kind of. We could require that you copy it yourself before running the build, and tell appveyor to do so.

mtuncer:
Is it possible to make it post-build action such that it would be created when the build succeeds ?

It was needed when installing citus from contrib folder.

lithp:
It's not ideal but I've opened https://github.com/citusdata/citus/pull/2173 to help us remember to update `citus.control`.

Unfortunately we don't have much control over the build process, we rely on `Mkvcbuild.pm`. I suppose we could add another patch which changes `Mkvcbuild.pm` and does the copy for us?

jasonmp85:
FWIW `Makefile`s _must_ use tabs, not spaces. If this PR was generated by a tool, the tool needs to be fixed to use tabs.

jasonmp85:
Aaaaand the `citus_version.h` file is also out of date…

PR #2171, opened 2018-05-15
Fix tests that fail under PG11

mtuncer:
Few PRs that broke PG11 tests are merged sınce we allowed PG11 failures during 7.4 development. Perhaps we should make PG11 tests mandatory from now on.


mtuncer:
@jasonmp85 You had some ideas on getting ```version_compat.h``` into better shape. We can also think about that.

mtuncer:
@jasonmp85 this is PR is node defunct. just closed. no further work is needed here.

PR #2173, opened 2018-05-15
Ensure the two copies of citus.control are kept in sync

lithp:
1. Leave a comment reminding they must be kept in sync
2. Add a check to travis which diffs the two files and confirms they're identical

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ matrix:
   allow_failures:
     - env: PGVERSION=11
 before_install:
-  - git clone -b v0.7.6 --depth 1 https://github.com/citusdata/tools.git
+  - git clone -b add-citus-control-check --depth 1 https://github.com/citusdata/tools.git
   - sudo make -C tools install
   - setup_apt
   - curl https://install.citusdata.com/community/deb.sh | sudo bash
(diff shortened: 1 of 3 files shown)


citus.control
```diff
@@ -1,5 +1,5 @@
 # Citus extension
-comment = 'Citus distributed database'
+omment = 'Citus distributed database'
```
lithp: Added this to show that it fails the build

.travis.yml
```diff
@@ -22,7 +22,7 @@ matrix:
   allow_failures:
     - env: PGVERSION=11
 before_install:
-  - git clone -b v0.7.6 --depth 1 https://github.com/citusdata/tools.git
+  - git clone -b add-citus-control-check --depth 1 https://github.com/citusdata/tools.git
```
lithp: I'll update this one https://github.com/citusdata/tools/pull/33 is merged

jasonmp85:
#2111 makes this go away, FWIW 😈…

lithp:
Yes, cmake would be dope

jasonmp85:
Eh, I'm not a fan of this. It's extra complication and bound to be incomplete in other ways later (in fact it already _is_ in the sense that it doesn't check `citus_version.h`, which was also wrong in the offending original PR).

I think we should hold a powwow about Windows support during build time and whether we want to continue the CMake work because without it we're going to keep needing hacks like this.

lithp:
I'm all for fixing the core problem, the fact that we use multiple build systems, but I don't see why that's a reason to block this PR. I'm not introducing any extra complication, just adding some safety to the complication that we already have. And sure, this doesn't solve all problems, but it does solve a problem!

jasonmp85:
Dropping appveyor support (#2541) means we're going with _[dropping] Windows from our build matrix_, so this PR is obsolete.

PR #2175, opened 2018-05-16
[WIP] Topn custom agg support

furkansahin:
This PR implements Top-N aggregate function support in Citus.


furkansahin:
Currently we hit to the issue #2184 with this PR. The aggregates are implemented but the `topn` function is not supported.

This issue cannot be resolved with this PR. First, we should decide how to solve it. 
1- Solve the issue #2184 
2- Implement another `topn` function which would return an already known type such as jsonb with the top n elements in TopN repository. (We will either introduce a new function for this or break the backward compatibility)
3- Implement a custom type for TopN and return it when `topn` is called. (We will either introduce a new function for this or break the backward compatibility)
4- Leave it as it is and require users to make `topn` call as `(topn(agg, 10)).*`.

PR #2176, opened 2018-05-17
Increase deadlock timeout so we get fewer signals

### LINKED ISSUE #2172: multi_real_time_transaction is flakey under appveyor
An example failure:

```
+++ C:/projects/postgres/contrib/citus/src/test/regress/results/multi_real_time_transaction.out	2018-05-15 10:09:43.041906600 -0700
@@ -298,23 +298,25 @@
 SELECT pg_reload_conf();
  pg_reload_conf 
 ----------------
  t
 (1 row)
 
 BEGIN;
 SELECT id, pg_advisory_lock(15) FROM test_table;
 ROLLBACK;
 BEGIN;
 SELECT id, pg_advisory_lock(15) FROM test_table;
-ERROR:  canceling the transaction since it was involved in a distributed deadlock
 ROLLBACK;
 BEGIN;
 SELECT id, pg_advisory_lock(15) FROM test_table;
 ROLLBACK;
 BEGIN;
 SELECT id, pg_advisory_lock(15) FROM test_table;
 ROLLBACK;
 BEGIN;
```

lithp:
Fixes #2172, as per Onder's suggestion.

@onderkalaci assigning you because [you originally suggested](https://github.com/citusdata/citus/pull/1905#issuecomment-354298684) the test that I'm changing. I don't really understand what it's for so I'm afraid that I'm making the test useless, but if you're okay with it I'm okay with it :)

### DIFF
--- a/src/test/regress/expected/multi_real_time_transaction.out
+++ b/src/test/regress/expected/multi_real_time_transaction.out
@@ -294,7 +294,7 @@ ROLLBACK;
 -- Test cancelling behaviour. See https://github.com/citusdata/citus/pull/1905.
 -- Repeating it multiple times to increase the chance of failure before PR #1905.
 SET client_min_messages TO ERROR;
-alter system set deadlock_timeout TO '1ms';
+alter system set deadlock_timeout TO '50ms';
 SELECT pg_reload_conf();
  pg_reload_conf 
 ----------------
--- a/src/test/regress/expected/multi_real_time_transaction_0.out
+++ b/src/test/regress/expected/multi_real_time_transaction_0.out
@@ -302,7 +302,7 @@ ROLLBACK;
 -- Test cancelling behaviour. See https://github.com/citusdata/citus/pull/1905.
 -- Repeating it multiple times to increase the chance of failure before PR #1905.
 SET client_min_messages TO ERROR;
-alter system set deadlock_timeout TO '1ms';
+alter system set deadlock_timeout TO '50ms';
 SELECT pg_reload_conf();
  pg_reload_conf 
 ----------------
--- a/src/test/regress/sql/multi_real_time_transaction.sql
+++ b/src/test/regress/sql/multi_real_time_transaction.sql
@@ -179,7 +179,7 @@ ROLLBACK;
 -- Test cancelling behaviour. See https://github.com/citusdata/citus/pull/1905.
 -- Repeating it multiple times to increase the chance of failure before PR #1905.
 SET client_min_messages TO ERROR;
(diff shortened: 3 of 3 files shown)


lithp:
All the tests passed, I'm restarting the appveyor build to see if we can reproduce it flaking out.

onderkalaci:
>  so I'm afraid that I'm making the test useless, but if you're okay with it I'm okay with it :)

Nope, the tests are still very useful. The only thing is that it takes  `50ms` to find the next distributed deadlock instead of `1ms`. So, since there are 5 consecutive tests rely on distributed deadlock detection, the test time would probably increase ~ `5 * 50ms * 2`, where 2 is the value of [citus.distributed_deadlock_detection_factor](https://docs.citusdata.com/en/v7.3/develop/api_guc.html?highlight=distributed_deadlock_detection_factor#citus-distributed-deadlock-detection-factor-floating-point).


> I don't really understand what it's for

1. Citus finds a distributed deadlock
2. Citus decides to cancel one of the processes and sends the signal
3. Citus finds the same deadlock again since the deadlock check timeout is too low
4. The process that Citus has picked on step 2 is canceled 
5. Citus decides to cancel the same backend as in 2 for the deadlock is found at step 3
6. Citus decides to cancel the same backend again, which is actually a bug. So, Citus cancels `ROLLBACK` or other commands that comes after the `SELECT` in the tests.

So, in plain word, the bug can be framed as: `We could find the same deadlock more than once when the timeout is too low and try to cancel the same backend incorrectly for the deadlocks detected later.` We could probably check postgresql's source code to see how they handle this issue.

onderkalaci:
>  We could probably check postgresql's source code to see how they handle this issue.

Oh, I think that's not a problem for PostgreSQL since the deadlock detection is not a seperate bg worker. Instead, the processes are checked for the deadlocks while [`ProcSleep()`](https://github.com/postgres/postgres/blob/9d4649ca49416111aee2c84b7e4441a0b7aa2fac/src/backend/storage/lmgr/proc.c#L1037)

lithp:
> We could find the same deadlock more than once when the timeout is too low and try to cancel the same backend incorrectly for the deadlocks detected later.

Oh, this is a great description of what was happening in #2172, but I meant I don't understand what this test was originally written to test :)

onderkalaci:
>  but I meant I don't understand what this test was originally written to test.

We've had some bugs around query cancellation when a real-time query is executed (#1905). After the fix, we came up with this hacky test. Actually, by adding ` pg_advisory_lock(15)` to the target list, we create a self-distributed deadlock. And the deadlock detection cancels the query automatically 😄

PR #2180, opened 2018-05-17
(WIP) Internals Monitor

pykello:
Being able to monitor internal data structures of Citus is useful for (1) understanding how Citus works, (2) Debugging.

- [x] `citus_connections_hash()` for monitoring `ConnectionHash` in `connection_management.c`.
- [x] `citus_connection_placements_hash()` for monitoring `ConnectionPlacementHash` in `placement_connection.c`.
- [x] List of queries sent over a connection.
- [x] transaction modification level.

### DIFF
--- a/src/backend/distributed/citus--7.4-2--7.4-3.sql
+++ b/src/backend/distributed/citus--7.4-2--7.4-3.sql
@@ -11,3 +11,65 @@ COMMENT ON FUNCTION worker_hash_partition_table(bigint, integer, text, text, oid
     IS 'hash partition query results';
 
 RESET search_path;
+
+CREATE FUNCTION citus_connections_hash()
+RETURNS TABLE (
+    hostname TEXT,
+    port INT,
+    "user" TEXT,
+    database TEXT,
+    socket INT,
+    sessionLifespan BOOLEAN,
+    claimedExclusively BOOLEAN,
+    connectionStart Timestamp WITH TIME ZONE,
+    transactionState TEXT,
+    transactionCritical BOOLEAN,
+    transactionFailed BOOLEAN,
+    preparedName TEXT,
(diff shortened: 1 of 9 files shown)


src/backend/distributed/citus--7.4-2--7.4-3.sql
```diff
@@ -10,4 +10,88 @@ COMMENT ON FUNCTION worker_hash_partition_table(bigint, integer, text, text, oid
                                                 anyarray)
     IS 'hash partition query results';
 
+CREATE FUNCTION citus_connections_hash()
```
marcocitus: Could we put these in the citus schema and make sure regular users cannot call them?

src/backend/distributed/connection/remote_commands.c
```diff
@@ -339,11 +340,29 @@ pchomp(const char *in)
...
 {
+	MemoryContext oldContext;
+	RemoteCommandLogRecord *logRecord;
+
 	if (!LogRemoteCommands)
 	{
 		return;
 	}
 
+	oldContext = MemoryContextSwitchTo(TopMemoryContext);
```
marcocitus: I usually just do:
```
SET client_min_messages TO LOG;
SET citus.log_remote_commands TO on;
```

pykello:
> Hey Hadi, do you think it is worth revisiting this PR? It makes it easier to learn about Citus internals.

I think having this will be useful. At the time I was doing this, my impression was that current data structures in connection managements are a bit overcomplicated and hard to understand, so we might want to simplify them a bit before moving this forward. I'll look at it again and come up with a plan.

pykello:
Now that I have better understanding of Citus internals, I'm closing this in favour of a future plan to create a better internals monitor tool.

PR #2186, opened 2018-05-24
Hll aggregate support

furkansahin:
This PR adds support for the aggregate functions in postgresql-hll to be known by Citus as supported aggregates and we create the master aggregate according to the given aggregate function.


src/test/regress/sql/custom_aggregate_support.sql
```diff
@@ -0,0 +1,61 @@
...
+
+
+SELECT EXTRACT(MONTH FROM day) AS month, hll_cardinality(hll_union_agg(unique_users))
+FROM daily_uniques
+WHERE day >= '2018-06-23' AND day <= '2018-07-01'
+GROUP BY 1 
+ORDER BY 1;
+
+-- These are going to be supported after window function support
+SELECT day, #hll_union_agg(unique_users) OVER seven_days
```
velioglu: Considering tests above, we may also use hll_cardinality instead of # operator within this and the following test.

src/test/regress/expected/custom_aggregate_support.out
```diff
@@ -0,0 +1,104 @@
+--
+-- CUSTOM_AGGREGATE_SUPPORT
+--
+-- Create HLL extension
+CREATE EXTENSION HLL;
+ERROR:  extension "hll" already exists
```
marcocitus: I think we need an alternative output for when you don't have HLL installed.

furkansahin:
@marcocitus Seems like all the tests passed, if you don't have any objection, I would love to merge this before any other PR (to avoid appveyor...)

mtuncer:
@furkansahin @marcocitus @velioglu  DESCRIPTION ?

PR #2189, opened 2018-05-28
Make sure that sequential DDL opens a single connection to each node

### LINKED ISSUE #2188: Multi-shard sequential modification doesn't use 2PC when not executed in a transaction block
When Citus uses seq. mode for multi-shard modifications (e.g., set citus.multi_shard_modify_mode TO 'sequential';), it doesn't use 2PC when not executed in a transaction.

Note that when executed in TX, we enable 2PC, see the related comment:

```
		/*
		 * If we're expanding the set nodes that participate in the distributed
		 * transaction, conform to MultiShardCommitProtocol.
		 */
```

Example:
```SQL
set citus.multi_shard_commit_protocol TO '2pc';
set citus.multi_shard_modify_mode TO 'sequential';
update users_table set value_1 = value_1 + 1;
....

LOG:  issuing COMMIT
LOG:  issuing COMMIT
LOG:  issuing COMMIT
```

onderkalaci:
Fixes #2188 

Another attempt to implement #2185. 

We preferred using the code-path that executes single task router
queries (e.g., ExecuteSingleModifyTask()) in order not to invent
a new executor that is only applicable for DDL commands that require
sequential execution.

Adding the regression tests is not very simple given that the only current user of the code `CREATE INDEX CONCURRENTLY` cannot be executed inside a transaction. Once we implement foreign keys tasks, I'll add the tests. (Now, you can simply enable `log_connections` on the worker and see the connections)

### DIFF
--- a/src/backend/distributed/executor/multi_router_executor.c
+++ b/src/backend/distributed/executor/multi_router_executor.c
@@ -728,7 +728,9 @@ CreatePlacementAccess(ShardPlacement *placement, ShardPlacementAccessType access
 
 /*
  * ExecuteSingleModifyTask executes the task on the remote node, retrieves the
- * results and stores them, if RETURNING is used, in a tuple store.
+ * results and stores them, if RETURNING is used, in a tuple store. The function
+ * can execute both DDL and DML tasks. When a DDL task is passed, the function
+ * does not expect scanState to be present.
  *
  * If the task fails on one of the placements, the function reraises the
  * remote error (constraint violation in DML), marks the affected placement as
@@ -739,9 +741,9 @@ static void
 ExecuteSingleModifyTask(CitusScanState *scanState, Task *task, bool multipleTasks,
 						bool expectResults)
 {
-	CmdType operation = scanState->distributedPlan->operation;
-	EState *executorState = scanState->customScanState.ss.ps.state;
-	ParamListInfo paramListInfo = executorState->es_param_list_info;
+	CmdType operation = CMD_UNKNOWN;
+	EState *executorState = NULL;
+	ParamListInfo paramListInfo = NULL;
 	List *taskPlacementList = task->taskPlacementList;
 	List *connectionList = NIL;
 	ListCell *taskPlacementCell = NULL;
@@ -753,9 +755,28 @@ ExecuteSingleModifyTask(CitusScanState *scanState, Task *task, bool multipleTask
 
 	char *queryString = task->queryString;
 	bool taskRequiresTwoPhaseCommit = (task->replicationModel == REPLICATION_MODEL_2PC);
+	bool failOnError = false;
 
 	ShardInterval *shardInterval = LoadShardInterval(task->anchorShardId);
 	Oid relationId = shardInterval->relationId;
+
+	if (multipleTasks)
+	{
+		taskRequiresTwoPhaseCommit |= (MultiShardCommitProtocol == COMMIT_PROTOCOL_2PC);
+	}
+
+	/* we don't allow failures to happen with CREATE INDEX CONCURRENTLY */
+	if (MultiShardCommitProtocol == COMMIT_PROTOCOL_BARE)
+	{
+		failOnError = true;
+	}
+
+	if (task->taskType == MODIFY_TASK)
+	{
+		operation = scanState->distributedPlan->operation;
+		executorState = scanState->customScanState.ss.ps.state;
+		paramListInfo = executorState->es_param_list_info;
+	}
 
 	/*
 	 * Modifications for reference tables are always done using 2PC. First
@@ -772,7 +793,8 @@ ExecuteSingleModifyTask(CitusScanState *scanState, Task *task, bool multipleTask
 	/*
 	 * Get connections required to execute task. This will, if necessary,
 	 * establish the connection, mark as critical (when modifying reference
-	 * table) and start a transaction (when in a transaction).
+	 * table or multi-shard command) and start a transaction (when in a
+	 * transaction).
 	 */
 	connectionList = GetModifyConnections(task, taskRequiresTwoPhaseCommit);
 
@@ -788,8 +810,11 @@ ExecuteSingleModifyTask(CitusScanState *scanState, Task *task, bool multipleTask
 		LockPartitionRelations(relationId, RowExclusiveLock);
 	}
 
-	/* prevent replicas of the same shard from diverging */
-	AcquireExecutorShardLock(task, operation);
+	if (task->taskType == MODIFY_TASK)
+	{
+		/* prevent replicas of the same shard from diverging */
+		AcquireExecutorShardLock(task, operation);
+	}
 
 	/* try to execute modification on all placements */
 	forboth(taskPlacementCell, taskPlacementList, connectionCell, connectionList)
@@ -797,7 +822,6 @@ ExecuteSingleModifyTask(CitusScanState *scanState, Task *task, bool multipleTask
 		ShardPlacement *taskPlacement = (ShardPlacement *) lfirst(taskPlacementCell);
 		MultiConnection *connection = (MultiConnection *) lfirst(connectionCell);
 		bool queryOK = false;
-		bool failOnError = false;
 		int64 currentAffectedTupleCount = 0;
 
 		if (connection->remoteTransaction.transactionFailed)
@@ -821,7 +845,7 @@ ExecuteSingleModifyTask(CitusScanState *scanState, Task *task, bool multipleTask
 		}
 
 		/* if we're running a 2PC, the query should fail on error */
-		failOnError = taskRequiresTwoPhaseCommit;
+		failOnError |= taskRequiresTwoPhaseCommit;
 
 		if (multipleTasks && expectResults)
 		{
@@ -848,6 +872,9 @@ ExecuteSingleModifyTask(CitusScanState *scanState, Task *task, bool multipleTask
 		 */
 		if (!gotResults && expectResults)
 		{
+			/* DDL tasks don't need to store any results */
+			Assert(task->taskType != DDL_TASK);
+
 			queryOK = StoreQueryResult(scanState, connection, failOnError,
 									   &currentAffectedTupleCount, NULL);
 		}
@@ -897,7 +924,10 @@ ExecuteSingleModifyTask(CitusScanState *scanState, Task *task, bool multipleTask
 	/* if some placements failed, ensure future statements don't access them */
 	MarkFailedShardPlacements();
 
-	executorState->es_processed += affectedTupleCount;
+	if (executorState)
+	{
+		executorState->es_processed += affectedTupleCount;
+	}
 
 	if (IsTransactionBlock())
 	{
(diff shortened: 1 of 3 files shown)


src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -753,10 +755,29 @@ ExecuteSingleModifyTask(CitusScanState *scanState, Task *task, bool multipleTask
...
 
 	ShardInterval *shardInterval = LoadShardInterval(task->anchorShardId);
 	Oid relationId = shardInterval->relationId;
 
+	if (multipleTasks)
+	{
+		taskRequiresTwoPhaseCommit |= (MultiShardCommitProtocol == COMMIT_PROTOCOL_2PC);
+	}
+
+	/* we don't allow failures to happen with CREATE INDEX CONCURRENTLY */
```
onderkalaci: I had to do that because we error out immediately with `COMMIT_PROTOCOL_BARE `, so wanted to stick to that (and pass the tests).

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -753,10 +755,29 @@ ExecuteSingleModifyTask(CitusScanState *scanState, Task *task, bool multipleTask
...
+	if (MultiShardCommitProtocol == COMMIT_PROTOCOL_BARE)
+	{
+		failOnError = true;
+	}
+
+	if (task->taskType == MODIFY_TASK)
```
onderkalaci: `scanState ` is deeply embedded into the function (see `StoreQueryResult()`). So, I ended-up this ugly `if` statement. I could do `scanState != NULL`, any comments/preference here?

src/backend/distributed/executor/multi_router_executor.c
marcocitus: Don't we always error out for DDL Commands? Should we check for DDL commands inside the loop?

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -753,10 +759,29 @@ ExecuteSingleModifyTask(CitusScanState *scanState, Task *task, bool multipleTask
...
+	if (scanState)
+	{
+		operation = scanState->distributedPlan->operation;
+		executorState = scanState->customScanState.ss.ps.state;
+		paramListInfo = executorState->es_param_list_info;
+	}
+
 	/*
 	 * Modifications for reference tables are always done using 2PC. First
 	 * ensure that distributed transaction is started. Then force the
```
marcocitus: There are some confusing interactions between `multipleTasks`, `failOnError`, `taskRequiresTwoPhaseCommit`, and (deeper down) `markCritical`. This could use some clean-up.

I wonder whether we should move the block below to the relevant caller.

Whether we're using 2PC seems in no way relevant to the execution of a single task, so it's kind of surprising that we have a `taskRequiresTwoPhaseCommit` variable in here. The fact that we don't deactivate a shard placement when a write to a reference table fails (*fail on error*) also seems somewhat independent from the fact that we execute it in a 2PC.

src/backend/distributed/executor/multi_router_executor.c
onderkalaci: I think this is related to the comment you've done below. We use `COMMIT_PROTOCOL_2PC` to decide `failOnError`. But, `CREATE INDEX CONCURRENTLY` uses `COMMIT_PROTOCOL_BARE ` and to enable `failOnError` behavior, I added this hack.

src/backend/distributed/executor/multi_router_executor.c
marcocitus: I was thinking something along the lines of:
https://github.com/citusdata/citus/commit/c135a291bb4cd3a17596d4eb4779879e3b899a43

This lays out the different failure handling cases in `RouterSequentialModifyExecScan`.

marcocitus:
Note that this PR doesn't actually affect regular DDL commands yet.

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -505,6 +533,33 @@ RouterSequentialModifyExecScan(CustomScanState *node)
...
+
+	if (taskList == NIL)
+	{
+		return false;
+	}
+
+	task = (Task *) linitial(taskList);
+	if (task->replicationModel == REPLICATION_MODEL_2PC)
+	{
+		return true;
```
marcocitus: Maybe `if (task->replicationModel != REPLICATION_MODEL_2PC) return false` for consistency.

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -848,6 +892,9 @@ ExecuteSingleModifyTask(CitusScanState *scanState, Task *task, bool multipleTask
 		 */
 		if (!gotResults && expectResults)
 		{
+			/* DDL tasks don't need to store any results */
+			Assert(task->taskType != DDL_TASK);
```
marcocitus: is this universally true? We wouldn't propagate them, but e.g. EXECUTE commands may return results.

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -1019,23 +1077,58 @@ ExecuteModifyTasksWithoutResults(List *taskList)
...
+	bool expectResults = false;
+	int64 affectedTupleCount = 0;
+	bool failOnError = true;
+	bool taskListRequires2PC = TaskListRequires2PC(taskList);
+
+	/*
+	 * When there are multiple tasks, start a coordinated transaction unless
+	 * requirest other (e.g., COMMIT_PROTOCOL_BARE) and use 2PC when enabled.
+	 */
+	if (multipleTasks)
```
marcocitus: Maybe start with a :
```
if (MultiShardCommitProtocol == COMMIT_PROTOCOL_BARE)
{
    /* we don't run CREATE INDEX CONCURRENTLY in a distributed transaction */
}
else
{
    ...
}
```

to simplify the code below

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -475,16 +475,43 @@ RouterSequentialModifyExecScan(CustomScanState *node)
...
+			 * in case of failures.
+			 *
+			 * For reference tables, we always failOnError since we absolutely want to avoid
+			 * marking any placements invalid.
+			 *
+			 * We also cannot handle faulures when there is RETURNING and there are more than
+			 * one task to execute.
+			 */
+			if (taskListRequires2PC ||
+				(multipleTasks && MultiShardCommitProtocol == COMMIT_PROTOCOL_2PC))
```
marcocitus: I think we could move this particular check into `TaskListRequires2PC`.

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -1019,23 +1077,58 @@ ExecuteModifyTasksWithoutResults(List *taskList)
...
+			BeginOrContinueCoordinatedTransaction();
+
+			if (MultiShardCommitProtocol == COMMIT_PROTOCOL_2PC ||
+				taskListRequires2PC)
+			{
+				CoordinatedTransactionUse2PC();
+			}
+		}
+	}
+	else if (taskListRequires2PC && MultiShardCommitProtocol != COMMIT_PROTOCOL_BARE)
```
onderkalaci: there is a bug here, I'll fix it.

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -505,6 +532,55 @@ RouterSequentialModifyExecScan(CustomScanState *node)
...
+	if (task->replicationModel == REPLICATION_MODEL_2PC)
+	{
+		return true;
+	}
+
+	/*
+	 * Some tasks don't set replicationModel thus we rely on
+	 * the anchorShardId as well replicationModel.
+	 *
+	 * TODO: Do we ever need replicationModel in the Task structure?
```
onderkalaci: I actually intended to remove `task->replicationModel `. But, (in the future) could we ever need a task where the `anchorShardId ` doesn't belong to a reference table but we need `2PC` for that task?

marcocitus: As a general observation: Metadata lookups during execution tend to get us into trouble so it's better to put it in the plan.

I could see 2PC being a replication model for a distributed table.

onderkalaci: > though not sure that really applies here.

Actually, we'll hit this metadata read on any sequential modification query that are not targeting a reference table. But, given that we call `LoadShardInterval()` in the `ExecuteSingleModifyTask()`, I thought `ReferenceTableShardId ()` wouldn't add any measurable overhead.

I'm ignoring the overhead for `DDL`s, which seems OK.

(We hit `LookupShardCacheEntry()` in both functions, which is the actual lookup)

src/test/regress/sql/sequential_ddl.sql
```diff
@@ -0,0 +1,114 @@
...
+SET citus.multi_shard_modify_mode TO 'sequential';
+SELECT recover_prepared_transactions();
+CREATE INDEX ref_test_seq_index ON ref_test(a);
+SELECT tx_count = worker_count FROM (SELECT count(*) as  tx_count FROM pg_dist_transaction WHERE gid LIKE 'citus_%_' || pg_backend_pid() || '%_%') as s1,  (SELECT count(*) as worker_count FROM pg_dist_node WHERE noderole = 'primary') as s2;
+
+-- reference tables should always use 2PC
+SET citus.multi_shard_modify_mode TO 'parallel';
+SELECT recover_prepared_transactions();
+CREATE INDEX ref_test_seq_index_2 ON ref_test(a);
+SELECT tx_count = worker_count FROM (SELECT count(*) as  tx_count FROM pg_dist_transaction WHERE gid LIKE 'citus_%_' || pg_backend_pid() || '%_%') as s1,  (SELECT count(*) as worker_count FROM pg_dist_node WHERE noderole = 'primary') as s2;
```
marcocitus: Is there a way to make these more readable?

onderkalaci: Initially I started with a single example, now ended up with many. So, lots of copy&paste.

Maybe wrapping the relevant queries into a functions makes it better?
```SQL
CREATE OR REPLACE FUNCTION distributed_2PCs_are_equal_to_worker_count()
    RETURNS bool AS
$$
DECLARE
    result bool;
BEGIN
    SELECT tx_count = worker_count FROM (SELECT count(*) as  tx_count FROM pg_dist_transaction WHERE gid LIKE 'citus_%_' || pg_backend_pid() || '%_%') as s1,  (SELECT count(*) as worker_count FROM pg_dist_node WHERE noderole = 'primary') as s2 INTO result;
    RETURN result;
END;
$$
LANGUAGE 'plpgsql' IMMUTABLE;

CREATE OR REPLACE FUNCTION distributed_2PCs_are_equal_to_placement_count()
    RETURNS bool AS
$$
DECLARE
    result bool;
BEGIN
    SELECT count(*) = current_setting('citus.shard_count')::bigint * current_setting('citus.shard_replication_factor')::bigint FROM pg_dist_transaction WHERE gid LIKE 'citus_%_' || pg_backend_pid() || '%_%'
     INTO result;
    RETURN result;
END;
$$
LANGUAGE 'plpgsql' IMMUTABLE;

```

PR #2190, opened 2018-05-29
Add node_conninfo GUC and related logic

jasonmp85:
To support more flexible (i.e. not at compile-time) specification of `libpq` connection parameters, this change adds a new GUC, `node_conninfo`, which must be a space-separated string of key-value pairs suitable for parsing by `libpq`'s connection establishment methods.

To avoid rebuilding and parsing these values at connection time, this change also adds a cache in front of the configuration params to permit immediate use of any previously-calculated parameters.


src/backend/distributed/shared_library_init.c
```diff
@@ -997,6 +1026,73 @@ NormalizeWorkerListPath(void)
...
+		newval = "";
+	}
+
+	optionArray = PQconninfoParse(newval, NULL);
+	if (optionArray == NULL)
+	{
+		ereport(ERROR, (errmsg("Unexpected")));
+	}
+
+	ResetConnParams();
```
marcocitus: Why isn't the `InitConnParams` call sufficient?

If we're thinking of making this function robust to the value changing if we decide to allow this in the future, shouldn't we also reset the hash?

src/backend/distributed/connection/connection_configuration.c
```diff
@@ -0,0 +1,239 @@
...
+
+
+static Size
+CalculateMaxSize()
+{
+	PQconninfoOption *defaults = PQconndefaults();
+	PQconninfoOption *option = NULL;
+	Size maxSize = 0;
+
+	for (option = defaults, maxSize = 1;
```
marcocitus: I get that we need to allocate the number of possible keywords + 1 because the last entry should be NULL. However, doing the +1 by initialising maxSize to 1 here is way too subtle. Can we just count the number of keywords here and add 1 at the call site?

src/backend/distributed/connection/connection_configuration.c
```diff
@@ -0,0 +1,239 @@
...
+#include "distributed/metadata_cache.h"
+#include "distributed/worker_manager.h"
+#include "distributed/task_tracker.h"
+#include "postmaster/postmaster.h"
+#include "mb/pg_wchar.h"
+#include "utils/builtins.h"
+
+char *NodeConninfo = "";
+
+typedef struct ConnParamsT
```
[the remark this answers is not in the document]
jasonmp85: It's a type, as opposed to the static var here that uses that type.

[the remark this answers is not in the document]
jasonmp85: I've changed it to `ConnParamsInfo`.

src/backend/distributed/connection/connection_configuration.c
```diff
@@ -0,0 +1,239 @@
...
+			  Size whitelistLength, char **errmsg)
+{
+	PQconninfoOption *optionArray = NULL;
+	PQconninfoOption *option = NULL;
+	Index whitelistIdx PG_USED_FOR_ASSERTS_ONLY = 0;
+
+	if (errmsg == NULL)
+	{
+		char *errmsgString;
+		errmsg = &errmsgString;
```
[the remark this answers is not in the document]
jasonmp85: Yup, but I'm not returning it to the user! This is when the user provides a `NULL` `errmsg`, i.e. they don't want an error message back.

I've added a comment to clarify this. Basically if they provide an `errmsg`, then I can assign `*errmsg = "my message";` to get it back to them.

If they _don't_ supply one, they're not looking to get back a message, so `errmsg` is basically a local variable here. I assign it a real pointer so the subsequent `*errmsg = "message"` assignments are valid and I don't have to do `NULL` checks on each one.

src/backend/distributed/connection/connection_configuration.c
```diff
@@ -0,0 +1,239 @@
...
+	}
+
+	if (conninfo == NULL)
+	{
+		return true;
+	}
+
+	if (uri_prefix_length(conninfo) != 0)
+	{
+		return false;
```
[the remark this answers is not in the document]
jasonmp85: Updated.

src/backend/distributed/connection/connection_configuration.c
```diff
@@ -0,0 +1,239 @@
...
+	}
+
+	optionArray = PQconninfoParse(conninfo, NULL);
+
+	if (optionArray == NULL)
+	{
+		return false;
+	}
```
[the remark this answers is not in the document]
jasonmp85: Probably a merge error (I had this in a single branch in enterprise and then meticulously `git rebase`'d it into three separate PRs, this looks like a mistake).

src/backend/distributed/connection/connection_configuration.c
jasonmp85: Well I mean, you figured it out 😆!

Sure, I'll do that.

src/backend/distributed/shared_library_init.c
```diff
@@ -997,6 +1026,73 @@ NormalizeWorkerListPath(void)
...
+			#endif
+		"sslcompression",
+		"sslcrl",
+		"sslmode",
+		"sslrootcert"
+	};
+	char *errmsg = NULL;
+	bool conninfoValid = CheckConninfo(*newval, whitelist, lengthof(whitelist), &errmsg);
+
+	if (!conninfoValid)
```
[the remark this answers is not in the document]
jasonmp85: Sure, I'll change it.

src/backend/distributed/shared_library_init.c
jasonmp85: Still thinking about this one.

jasonmp85: I think being robust to changes is in the realm of e.g. changing `.pgpass` or renaming a role while connections are already cached for the active session. Even if we permit these changes they necessarily are going to only take effect in the next session (so this might be something we'd permit with a SIGHUP or something, but would only take effect with new sessions?)

I've added a comment to say "if we ever allow changing this value at runtime, the conninfo cache should also be cleared".

As for "why separate the logic", I just liked having the initialization of the structure and the initial population of its hardcoded values to be separate. I was sort of modeling this after StringInfo in a way (init/reset/add).

jasonmp85:
I've added a ton of comments in the `connection_configuration` file to clarify the API and lifecycle of this stuff. The history of it is that I started with a bunch of `extern` variables (four, I think) but then realized that would be too error prone so I swapped that for a single (`static`) instance of something more structured, along with functions to manipulate and use it.

I'll give the whole PR another pass in the morning, but in the meanwhile was there anything really pressing bothering you about it? I think I've addressed all places you annotated.

src/backend/distributed/shared_library_init.c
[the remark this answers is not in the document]
jasonmp85: Err, I take it back; I need to have the boolean value to return from the check hook, so it's nicer to save it once then reference it at the `return`.

src/backend/distributed/connection/connection_configuration.c
```diff
@@ -0,0 +1,239 @@
...
+	pg_ltoa(key->port, nodePortString);
+
+	*keywords = MemoryContextAllocZero(context, ConnParams.maxSize * sizeof(char *));
+	*values = MemoryContextAllocZero(context, ConnParams.maxSize * sizeof(char *));
+
+	memcpy(*keywords, ConnParams.keywords, ConnParams.size * sizeof(char *));
+	memcpy(*values, ConnParams.values, ConnParams.size * sizeof(char *));
+
+	memcpy(&((*keywords)[ConnParams.size]), runtimeKeywords, sizeof(runtimeKeywords));
+	memcpy(&(*values)[ConnParams.size], runtimeValues, sizeof(runtimeValues));
```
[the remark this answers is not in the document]
jasonmp85: Yeah, I got a bit lost myself.

I added a ton of comments clarifying that this method basically just builds an array from three inputs (global GUC values, runtime values from the connection key, and auth values (in Enterprise)).

So what's going on in this line in particular is we're memcpy-ing with an offset. We copied the first `ConnParams.size` values from the GUC stuff and now are about to copy host, port, etc. into the subsequent array slots.

I think the `char ***` was the root of the problem here. In particular, precedence rules meant I had to use a bunch of parenthesis here to get what I needed. I've changed the code to assign to two local `char **`s earlier in the function and now do all manipulation using those before ultimately assigning them to the out parameters. I think this makes it much more readable. In fact, when I searched for "memcpy with offset", I got [this](https://stackoverflow.com/a/1163943/179878), and with the `char **` my syntax looks like that now, which is admittedly nicer than what you annotated.

src/backend/distributed/connection/connection_configuration.c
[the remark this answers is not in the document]
jasonmp85: I've added a comment to point out that basically I just need a valid pointer to do assignments and that this case kicks in when the user doesn't want an error message back.

src/backend/distributed/connection/connection_configuration.c
```diff
@@ -0,0 +1,239 @@
...
+	Size size;
+	Size maxSize;
+} ConnParamsT;
+static ConnParamsT ConnParams;
+
+static Size CalculateMaxSize(void);
+static int uri_prefix_length(const char *connstr);
+
+void
+InitConnParams()
```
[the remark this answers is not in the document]
jasonmp85: Added them all (for this file, at least).

src/backend/distributed/connection/connection_configuration.c
```diff
@@ -0,0 +1,335 @@
...
+	PQconninfoOption *option = NULL;
+	Index whitelistIdx PG_USED_FOR_ASSERTS_ONLY = 0;
+
+	/*
+	 * If the user doesn't need a message, just overwrite errmsg with a stack
+	 * variable so we can always safely write to it.
+	 */
+	if (errorMsg == NULL)
+	{
+		char *errorMsgString;
```
marcocitus: This seems wrong. Once this block is over this variable is implicitly popped off the stack. There are other blocks below that declare variables that will overwrite the memory of this one. Maybe move the variable to the top of the function.

src/backend/distributed/connection/connection_configuration.c
```diff
@@ -0,0 +1,335 @@
...
+
+	pg_ltoa(key->port, nodePortString); /* populate node port string with port */
+
+	/* first step: copy global parameters to beginning of array */
+	memcpy(connKeywords, ConnParams.keywords, ConnParams.size * sizeof(char *));
+	memcpy(connValues, ConnParams.values, ConnParams.size * sizeof(char *));
+
+	/* second step: begin at end of global params and copy runtime ones */
+	memcpy(&connKeywords[ConnParams.size], runtimeKeywords, sizeof(runtimeKeywords));
+	memcpy(&connValues[ConnParams.size], runtimeValues, sizeof(runtimeValues));
```
marcocitus: Thanks, this all looks a lot cleaner and readable.

src/backend/distributed/shared_library_init.c
```diff
@@ -997,6 +1026,74 @@ NormalizeWorkerListPath(void)
...
+
+	if (newval == NULL)
+	{
+		newval = "";
+	}
+
+	optionArray = PQconninfoParse(newval, NULL);
+	if (optionArray == NULL)
+	{
+		ereport(ERROR, (errmsg("Unexpected")));
```
marcocitus: The user we'll probably never see this, but for the developer maybe add a comment / message on why it's unexpected. e.g. `"Unexpected because NodeConninfoGucCheckHook prevents invalid values"

src/backend/distributed/shared_library_init.c
marcocitus: For some reason Postgres hash tables don't have a reset function, so it might make more sense to have something like:

```c
FreeConnParams();
InitConnParams();
```

src/backend/distributed/shared_library_init.c
```diff
@@ -997,6 +1026,74 @@ NormalizeWorkerListPath(void)
...
+		"keepalives_count",
+		"keepalives_idle",
+		"keepalives_interval",
+			#if defined(ENABLE_GSS) || defined(ENABLE_SSPI)
+		"krbsrvname",
+			#endif
+		"sslcompression",
+		"sslcrl",
+		"sslmode",
+		"sslrootcert"
```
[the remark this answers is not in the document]
marcocitus: nvm

src/backend/distributed/connection/connection_configuration.c
jasonmp85: Wow talk about a complete slip-up. I was thinking "stack == function call" but you're totally right that automatic storage rules mean this is scoped to this bracketed region. Fixing it.

src/backend/distributed/shared_library_init.c
jasonmp85: Yeah I almost considered making it an `Assert`; what are your thoughts on that?

If I keep it as an `ereport`, what do you think about making it `FATAL`? It seems like a pretty bad state to be in. I'll update the error language regardless.

src/backend/distributed/shared_library_init.c
jasonmp85: I believe they do as of PostgreSQL 10 (or maybe I have 11 checked out; I can't remember, but I looked into it last night)…

That said, the reset function frees all the entries, which is problematic if they contain pointers to allocated memory. At minimum I'd want to make a separate child context of `ConnectionContext` I could reset for this purpose and most of this stuff all seems like premature optimization so long as the GUC requires a restart and must be in the conf file.

src/backend/distributed/shared_library_init.c
marcocitus: I have a slight dislike of asserts because they're not so easy to debug and create slight behavioural changes, FATAL makes sense.

src/backend/distributed/connection/connection_configuration.c
jasonmp85: Fixed.

src/backend/distributed/shared_library_init.c
jasonmp85: Changed.

jasonmp85:
@marcocitus is this good to merge to master?

jasonmp85:
@marcocitus Also, Appveyor is getting a weird error here for `dml_recursive`; it did it in both branches. I don't immediately see why this would be… could there be something non-deterministic about that test?

src/backend/distributed/shared_library_init.c
```diff
@@ -329,15 +346,13 @@ RegisterCitusConfigVariables(void)
 
 	DefineCustomEnumVariable(
 		"citus.sslmode",
-		gettext_noop("SSL mode to use for connections to worker nodes."),
-		gettext_noop("When connecting to a worker node, specify whether the SSL mode"
-					 "mode for the connection is 'disable', 'allow', 'prefer' "
-					 "(the default), 'require', 'verify-ca' or 'verify-full'."),
+		gettext_noop("This GUC variable has been deprecated."),
```
lithp: This is not a useful message. It should have also said something like "please use citus.node_conninfo"

lithp: Opened #2243

mtuncer:
@jasonmp85 @marcocitus do we need a DESCRIPTION for this one ?

marcocitus:
Thanks. Added.

PR #2191, opened 2018-05-29
Add pg_dist_authinfo schema and validation

jasonmp85:
This table will be used by Citus Enterprise to populate authentication-related fields in outbound connections; Citus Community lacks support for this functionality.

### DIFF
--- a/src/backend/distributed/utils/metadata_cache.c
+++ b/src/backend/distributed/utils/metadata_cache.c
@@ -9,6 +9,7 @@
 
 #include "stdint.h"
 #include "postgres.h"
+#include "libpq-fe.h"
 #include "miscadmin.h"
 
 #include "access/genam.h"
@@ -26,6 +27,7 @@
 #include "commands/extension.h"
 #include "commands/trigger.h"
 #include "distributed/colocation_utils.h"
+#include "distributed/connection_management.h"
 #include "distributed/citus_ruleutils.h"
 #include "distributed/master_metadata_utility.h"
 #include "distributed/metadata_cache.h"
@@ -48,6 +50,7 @@
 #include "utils/builtins.h"
 #include "utils/catcache.h"
 #include "utils/datum.h"
+#include "utils/elog.h"
 #include "utils/hsearch.h"
 #include "utils/inval.h"
 #include "utils/fmgroids.h"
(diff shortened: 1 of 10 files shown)


src/backend/distributed/utils/metadata_cache.c
```diff
@@ -3415,3 +3419,26 @@ DistNodeMetadata(void)
...
+}
+
+
+Datum
+authinfo_valid(PG_FUNCTION_ARGS)
+{
+	ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+					errmsg("catalog table unsupported"),
+					errdetail("Citus Community Edition does not support the use of per-"
+							  "host/user authentication options."),
```
marcocitus: `per host/` is probably not relevant here

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -3415,3 +3419,26 @@ DistNodeMetadata(void)
...
+
+	PG_RETURN_BOOL(roleExists);
+}
+
+
+Datum
+authinfo_valid(PG_FUNCTION_ARGS)
+{
+	ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+					errmsg("catalog table unsupported"),
```
marcocitus: Maybe `cannot write to pg_dist_authinfo`. That's a little clearer in case this call happens somewhere deep down.

src/backend/distributed/utils/metadata_cache.c
jasonmp85: Do we want to maybe let Community users at least specify a password here? Using `pgpass` is still kind of annoying and requires filesystem access. I think I wrote this error message with the idea that we might let them set at least a global password.

src/backend/distributed/utils/metadata_cache.c
jasonmp85: Sure.

jasonmp85:
I've addressed all your comments and added a test to exercise the check constraint implementations. We good to :shipit: (i.e. merge this into the other PR?)

jasonmp85:
@marcocitus I don't know what's up with appveyor, but I've addressed any remaining feedback and documented every method; is this good to go, or is anything else needed? I responded to your questions about poolname/poolport in our email thread.

jasonmp85:
@marcocitus Also, Appveyor is getting a weird error here for `dml_recursive`; it did it in both branches. I don't immediately see why this would be… could there be something non-deterministic about that test?

jasonmp85:
I re-triggered the branch build here (but not on #2190) so I suppose we'll see how non-deterministic it is.

PR #2194, opened 2018-05-29
master_modify_multiple_shards() and TRUNCATE honour `citus.multi_shard_modify_mode`

onderkalaci:
Based on #2185 and expands it to cover `TRUNCATE` and `master_modify_multiple_shards()`.

This change is important to support `TRUNCATE reference_table CASCADE` where the reference table cascades into distributed tables. At that point, we should change the modification mode to seq., otherwise Citus ends-up with self-deadlock.

Note that `TRUNCATE table CASCADE` cascades the `TRUNCATE` command to the relations which refer to `table` via a foreign key.

### DIFF
--- a/src/backend/distributed/executor/multi_router_executor.c
+++ b/src/backend/distributed/executor/multi_router_executor.c
@@ -82,7 +82,8 @@ static void AcquireMetadataLocks(List *taskList);
 static ShardPlacementAccess * CreatePlacementAccess(ShardPlacement *placement,
 													ShardPlacementAccessType accessType);
 static int64 ExecuteSingleModifyTask(CitusScanState *scanState, Task *task,
-									 bool failOnError, bool expectResults);
+									 CmdType operation, bool failOnError,
+									 bool expectResults);
 static void ExecuteSingleSelectTask(CitusScanState *scanState, Task *task);
 static List * GetModifyConnections(Task *task, bool markCritical);
 static void ExecuteMultipleTasks(CitusScanState *scanState, List *taskList,
@@ -479,6 +480,7 @@ RouterSequentialModifyExecScan(CustomScanState *node)
 		EState *executorState = scanState->customScanState.ss.ps.state;
 		bool taskListRequires2PC = TaskListRequires2PC(taskList);
 		bool failOnError = false;
+		CmdType operation = scanState->distributedPlan->operation;
 
 		/*
 		 * We could naturally handle function-based transactions (i.e. those using
@@ -521,7 +523,8 @@ RouterSequentialModifyExecScan(CustomScanState *node)
 			Task *task = (Task *) lfirst(taskCell);
 
 			executorState->es_processed +=
-				ExecuteSingleModifyTask(scanState, task, failOnError, hasReturning);
+				ExecuteSingleModifyTask(scanState, task, operation, failOnError,
+										hasReturning);
 		}
 
 		scanState->finishedRemoteScan = true;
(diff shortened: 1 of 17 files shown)


src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -814,9 +814,19 @@ ExecuteSingleModifyTask(CitusScanState *scanState, Task *task, bool multipleTask
...
-	if (task->taskType == MODIFY_TASK)
+	/*
+	 * Prevent replicas of the same shard from diverging. We don't
+	 * need to acquire lock for TRUNCATE_TASK since it already
+	 * acquires AccessExclusiveLock on the relation, and blocks any
+	 * concurrent operation.
+	 *
+	 * Note that master_modify_multiple_shards() does not set
+	 * the operation type, thus does not acquire the executor
+	 * shard locks, which is a known issue.
```
marcocitus: There seems to be less harm it taking a lock on `TRUNCATE` then not taking a lock on `master_modify_multiple_shards`

src/test/regress/sql/multi_truncate.sql
```diff
@@ -221,3 +221,75 @@ DELETE FROM pg_dist_partition WHERE logicalrelid = 'test_local_truncate'::regcla
...
 DROP TABLE test_local_truncate;
+
+CREATE TABLE test_seq_truncate (a int);
+INSERT INTO test_seq_truncate SELECT i FROM generate_series(0, 100) i;
+
+SELECT create_distributed_table('test_seq_truncate', 'a');
+
+-- disable the 2PC recovery for now
+ALTER SYSTEM SET citus.recover_2pc_interval TO -1;
+SELECT pg_reload_conf();
```
marcocitus: we should probably add a note somewhere that this test file cannot be used in parallel with others.

src/backend/distributed/executor/multi_router_executor.c
onderkalaci: I wanted to keep the behavior not changed because `master_modify_multiple_shards()` doesn't do `AcquireExecutorShardLock ()`. I thought that's an intended behavior. 

Do we prefer to fix it now?

src/test/regress/sql/multi_truncate.sql
onderkalaci: I'm planning to move these tests to the new file added in #2189. I'll rename `sequential_ddl.sql` to `sequential_modification.sql`.

src/backend/distributed/executor/multi_router_executor.c
marcocitus: It still goes through `AcquireExecutorMultiShardLocks`, e.g. UPDATEs on RF>1 take exclusive locks.

onderkalaci: That's true, I missed that by adding a break point to `AcquireExecutorShardLock()` and not considering `AcquireExecutorMultiShardLocks()` at all. I should have stopped at `LockShardResource()`.

The minor difference between `AcquireExecutorShardLock ()` and `AcquireExecutorMultiShardLocks ()` is that the former expects `commandType` where `master_modify_multiple_shards` doesn't currently provide when calling `ExecuteSingleModifyTask()`.

We could either explicitly pass `commandType`  to `ExecuteSingleModifyTask()` or leave as is. Are you OK with passing `commandType` to `ExecuteSingleModifyTask` along with `Task`? I'm a bit hesitant doing that since we already have `TaskType` which is unfortuneatly not sufficient for this purpose :/

(I'm avoiding the alternative to add the commandType to task, which sound not a good idea.)

marcocitus: I think adding a parameter is probably ok. It'd be good to gravitate towards taking scanState out of the API, since there might be significant API changes in PG12.

onderkalaci:
Just FYI: #2198 and #2199 are based on this PR, which are approved. I'm not merging those PRs until this is approved not to deal with `git rebase`.

src/backend/distributed/master/master_modify_multiple_shards.c
```diff
@@ -147,7 +148,8 @@ master_modify_multiple_shards(PG_FUNCTION_ARGS)
 #endif
 	modifyQuery = (Query *) linitial(queryTreeList);
 
-	if (modifyQuery->commandType != CMD_UTILITY)
+	operation = modifyQuery->commandType;
+	if (operation != CMD_UTILITY)
```
marcocitus: kind of a strange `if` statement? What do we do for utility commands?

marcocitus: ah, I guess it's for truncate.

PR #2196, opened 2018-06-01
Single DDL execution and add foreign key initials

furkansahin:


onderkalaci:
@furkansahin  overridden this by #2201 so closing this

PR #2197, opened 2018-06-01
Adds SELECT ... FOR UPDATE support for router plannable queries

### LINKED ISSUE #1236: SELECT ... FOR UPDATE is not supported for router-plannable queries
## Test case

```
SELECT "projects".* FROM "projects" WHERE "projects"."account_id" = 20 AND "projects"."id" = 12 LIMIT 1 FOR UPDATE;
```

returns

```
ERROR:  could not run distributed query with FOR UPDATE/SHARE commands
```

Expected behavior is for this to be sent down to the single worker using the router executor. Note that this typically runs within a transaction, with an UPDATE on that same record to follow a bit later.

## Use case

One source of this SQL is `with_lock` in ActiveRecord, which lead me to encounter the problem. Its likely customers using Rails and doing atomic operations on rows will run into this issue. `with_lock` is a common approach to perform complex atomic changes on records in Rails.

## Test schema

```
postgres=# \d projects
                                Table "public.projects"
   Column   |       Type        |                       Modifiers                       
------------+-------------------+-------------------------------------------------------
 id         | integer           | not null default nextval('projects_id_seq'::regclass)
 name       | character varying | 
 account_id | integer           | not null
Indexes:
    "projects_pkey" PRIMARY KEY, btree (id, account_id)
```

```
postgres=# SELECT * from master_get_table_metadata('projects');
 logical_relid | part_storage_type | part_method |  part_key  | part_replica_count | part_max_size | part_placement_policy 
---------------+-------------------+-------------+------------+--------------------+---------------+-----------------------
         23901 | t                 | h           | account_id |                  1 |    1073741824 |                     2
(1 row)
```

---

In chat @marcocitus mentioned that this seems to be due to MultiRouterPlannableQuery returning false for hasForUpdate queries.

velioglu:
Fixes #1236, #1870 

With this PR, `SELECT ... FOR UPDATE` queries will be supported for router plannable queries if it includes hash distributed and reference tables only. Only limitation is not having a distributed table with replication factor >1. 

If reference table is affected by row locking clause, executor shard lock is taken to prevent parallel modifications. Note that, difference between the logic of Postgres and Citus for row locks on reference tables is explained in 'AcquireExecutorShardLock'. For distributed tables, we directly send queries with locking clause to worker nodes. 

Minor remaining items :

- [x] Fixing error logs in real-time planning logic
- [x] Adding isolation tests
- [x] Adding some more regression tests
    - [x] Add tests with all row lock types
    - [x] Add more complex router plannable queries


src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -223,6 +223,44 @@ AcquireExecutorShardLock(Task *task, CmdType commandType)
...
+	 * given in the query.
+	 */
+	if (task->rowLocksOnRelations != NIL)
+	{
+		ListCell *rowLockTypeCell = NULL;
+		ListCell *relationOidCell = NULL;
+		LOCKMODE rowLockMode = NoLock;
+
+		forboth(rowLockTypeCell, task->rowLocksOnRelations, relationOidCell,
+				task->tableOidList)
```
marcocitus: In a similar situation (relationShardList), we opted for introducing a list of structs. Would it make sense to do the same thing here? It's a bit less brittle in terms of relying on the order of these lists being the same.

marcocitus:
There's a subtle issue around reference tables which only have a single replica.

In `AcquireExecutorShardLocks` we have the following snippet:
```c
    if (commandType == CMD_SELECT || list_length(task->taskPlacementList) == 1)
    {
        /*
         * The executor shard lock is used to maintain consistency between
         * replicas and therefore no lock is required for read-only queries
         * or in general when there is only one replica.
         */

        lockMode = NoLock;
    }
```

It's good that we don't take shard resource locks when there is no replication, that saves us quite a bit of overhead in single-replica set-ups like Citus Cloud. However, that breaks the SELECT .. FOR UPDATE logic for reference tables if there is a single node.

To fix it, we should maybe do something like:

```c
    if (commandType == CMD_SELECT)
    {
        /*
         * The executor shard lock is used to maintain consistency between
         * replicas and therefore no lock is required for read-only queries.
         */

        lockMode = NoLock;
    }
    else if (list_length(task->taskPlacementList) == 1)
    {
        if (task->replicationModel == REPLICATION_MODEL_2PC)
        {
            /*
             * While we don't need a lock to ensure writes are applied in
             * a consistent order when there is a single replica. We also use
             * shard resource locks as a crude implementation of SELECT..
             * FOR UPDATE on reference tables, so we should always take
             * a lock that conflicts with the FOR UPDATE/SHARE locks.
             */
            lockMode = RowExclusiveLock;
        }
        else
        {
            /*
             * When there is no replication, the worker itself can decide on
             * on the order in which writes are applied.
             */
            lockMode = NoLock;
        }
    }
```

velioglu:
I have added missing explanations and tests mentioned in the description and also applied the change Marco mentioned above. So, PR is ready for the review.

metdos:
@velioglu, could you also rebase and resolve the conflict?

marcocitus:
Seems there's a bug with views:

```sql
CREATE VIEW test_1 AS SELECT * FROM test WHERE y = 1;                                                                                                                   
INSERT INTO test VALUES (1,1), (2,2);
SELECT * FROM test WHERE x = 1 FOR UPDATE;
 x │ y 
───┼───
 1 │ 1
(1 row)

SELECT * FROM test_1 WHERE x = 1 FOR UPDATE;
 x │ y 
───┼───
(0 rows)
```

same commands on postgre:
```sql
CREATE VIEW test_1 AS SELECT * FROM test WHERE y = 1;                                                                                                                 
INSERT INTO test VALUES (1,1), (2,2);                                                                                                                                      
SELECT * FROM test_1 WHERE x = 1 FOR UPDATE;
 x │ y 
───┼───
 1 │ 1
(1 row)

```

src/test/regress/specs/isolation_select_for_update.spec
```diff
@@ -0,0 +1,145 @@
...
+
+step "s1-select-from-t1-within-cte"
+{
+	WITH first_value AS ( SELECT val_1 FROM test_table_1_rf1 WHERE id = 1 FOR UPDATE)
+	SELECT * FROM first_value;
+}
+
+step "s1-select-from-t1-with-subquery"
+{
+	SELECT * FROM (SELECT * FROM test_table_1_rf1 FOR UPDATE) foo WHERE id = 1;
```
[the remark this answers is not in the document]
velioglu: Added

marcocitus:
There is another place that uses UpdateOrDeleteRTE which leads to some weird bugs:

e.g. only goes to a single replica of the reference table:
```sql
UPDATE ref SET b = 2 WHERE b = 1 AND b IN (SELECT x FROM test WHERE x = 1 FOR UPDATE);
LOG:  issuing BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED;SELECT assign_distributed_transaction_id(0, 3, '2018-06-04 07:57:40.432338+02');
LOG:  issuing UPDATE public.ref_102264 ref SET b = 2 WHERE ((b OPERATOR(pg_catalog.=) 1) AND (b OPERATOR(pg_catalog.=) ANY (SELECT test.x FROM public.test_102746 test WHERE (test.x OPERATOR(pg_catalog.=) 1) FOR UPDATE OF test)))
LOG:  issuing PREPARE TRANSACTION 'citus_0_47906_3_1'
LOG:  issuing COMMIT PREPARED 'citus_0_47906_3_1'
UPDATE 0
```

without FOR UPDATE:
```sql
 UPDATE ref SET b = 2 WHERE b = 1 AND b IN (SELECT x FROM test WHERE x = 1);
ERROR:  cannot perform select on a distributed table and modify a reference table
Time: 0.703 ms
```

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -2768,6 +2789,24 @@ MultiRouterPlannableQuery(Query *query, RelationRestrictionContext *restrictionC
...
+			 * Currently, we don't support tables with replication factor > 1,
+			 * except reference tables with SELECT ... FOR UDPATE queries. It is
+			 * also not supported from MX nodes.
+			 */
+			if (query->hasForUpdate)
+			{
+				uint32 tableReplicationFactor = TableShardReplicationFactor(
+					distributedTableId);
+
+				EnsureCoordinator();
```
marcocitus: We only need to error out if there is a reference table, because then the resource locks are ineffective.

mtuncer:
@velioglu @marcocitus DESCRIPTION ?

PR #2198, opened 2018-06-04
INSERT .. SELECT pushdown honors multi_shard_modification_mode

onderkalaci:
Description: `INSERT .. SELECT` pushdown honors multi_shard_modification_mode

### DIFF
--- a/src/test/regress/sql/sequential_modifications.sql
+++ b/src/test/regress/sql/sequential_modifications.sql
@@ -219,6 +219,30 @@ COMMIT;
 -- see that all the data successfully removed
 SELECT count(*) FROM multi_shard_modify_test;
 
+-- test INSERT ... SELECT queries
+-- with sequential modification mode, we should see #primary worker records
+SET citus.multi_shard_modify_mode TO 'sequential';
+SELECT recover_prepared_transactions();
+INSERT INTO multi_shard_modify_test SELECT * FROM multi_shard_modify_test;
+SELECT distributed_2PCs_are_equal_to_worker_count();
+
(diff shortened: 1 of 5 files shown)


src/test/regress/sql/sequential_modifications.sql
```diff
@@ -206,6 +206,18 @@ SELECT recover_prepared_transactions();
 SELECT master_modify_multiple_shards('DELETE FROM multi_shard_modify_test');
 SELECT distributed_2PCs_are_equal_to_worker_count();
 
+-- test INSERT ... SELECT queries
+-- with sequential modification mode, we should see #primary worker records
+SET citus.multi_shard_modify_mode TO 'sequential';
+SELECT recover_prepared_transactions();
+INSERT INTO multi_shard_modify_test SELECT * FROM multi_shard_modify_test;
+SELECT distributed_2PCs_are_equal_to_worker_count();
```
marcocitus: Now that I saw your multi-row INSERT test in #2199 I kind of like that more. I can live with having these calls, but I like that the multi-row INSERT tests an actual scenario that would go wrong if we didn't switch to sequential mode.

onderkalaci: Sure, I'll try to come up with such a test as well

onderkalaci: Although not 100% applicable, I actually see the tests with `distributed_2PCs_are_equal_to_worker_count ()` etc  as `the unit tests` for this change.  And, the more realistic tx tests as the `system tests`. So, keeping them all for now seems OK?

PR #2199, opened 2018-06-05
Realtime executor honours multi_shard_modify_mode

onderkalaci:
Another (and final!) pre-requisite for foreign keys to reference tables with respect to sequential execution.

After this commit, real-time `SELECT`s will rely on `multi_shard_modify_mode`
GUC . The name of the GUC is unfortunate, but, adding one more GUC
(or renaming the GUC) would make the UX even worse. Given that this
mode is more important for transaction blocks that involve modification
/DDL queries along with real-time SELECTs, we can probably live with the confusion.
Also, we're not going to publicly mention about this feature. This is mostly
going to be useful behind the covers.

I've executed the whole (`check-multi`) regression test suite under the sequential mode, and haven't
observed anything unexpected. We got rid of some of the 
`cannot establish a new connection  .. since` errors, and introduced few new ones. The errors become more common with recursive planning along with `INSERT .. SELECT `where the results 
are `COPY`ied to a distributed table. An example query that doesn't work under seq. mode
but works with parallel mode:

```SQL
WITH raw_data AS (
	DELETE FROM modify_table RETURNING *
)
INSERT INTO summary_table SELECT id, COUNT(*) AS counter FROM raw_data GROUP BY id;
```

Queries like the above fails on the sequential mode mainly because we cannot run `COPY` after any sequential operation.

### DIFF
--- a/src/test/regress/expected/with_transactions.out
+++ b/src/test/regress/expected/with_transactions.out
@@ -105,6 +105,23 @@ SELECT count(*) FROM second_raw_table;
      0
 (1 row)
 
+-- sequential insert followed by a sequential real-time query should be fine
+BEGIN;
+SET LOCAL citus.multi_shard_modify_mode TO 'sequential';
+WITH ids_inserted AS
+(
+  INSERT INTO raw_table (tenant_id) VALUES (11), (12), (13), (14) RETURNING tenant_id
+)
+SELECT income FROM second_raw_table WHERE tenant_id IN (SELECT * FROM ids_inserted) ORDER BY 1 DESC LIMIT 3;
+DEBUG:  data-modifying statements are not supported in the WITH clauses of distributed queries
+DEBUG:  generating subplan 17_1 for CTE ids_inserted: INSERT INTO with_transactions.raw_table (tenant_id) VALUES (11), (12), (13), (14) RETURNING raw_table.tenant_id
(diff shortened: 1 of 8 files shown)


onderkalaci:
> The name of the GUC is unfortunate, but, adding one more GUC
(or renaming the GUC) would make the UX even worse. 

This is frustrating. I'm listing the alternatives that I can see and open to any other suggestion:

1. Deprecate `citus.multi_shard_modify_mode` and introduce `citus.multi_shard_execution_mode`. This is backward incompatible.

2. Live with `citus.multi_shard_modify_mode` affecting real-time SELECTs.

3. Introduce `citus.multi_shard_execution_mode`, whenever this is set, update `citus.multi_shard_modify_mode` behind the covers. This sounds like a confusing behavior.

Anything else?

src/test/regress/expected/with_transactions.out
```diff
@@ -105,6 +105,23 @@ SELECT count(*) FROM second_raw_table;
      0
 (1 row)
 
+-- sequential insert followed by a sequential real-time query should be fine
```
marcocitus: Interesting test. Not for this PR, but now I kind of wonder whether we could make this automatic :)

marcocitus:
> Live with citus.multi_shard_modify_mode affecting real-time SELECTs.

so far this is my favourite option

src/test/regress/expected/with_transactions.out
onderkalaci: I'll keep an eye on this once we introduce automated switches.

PR #2200, opened 2018-06-06
foreign key related code refactoring

furkansahin:
This PR refactors the foreign key related code paths from Citus code base and aims to gather them under foreign_constraint_utils.c


src/backend/distributed/utils/foreign_constraint_utils.c
```diff
@@ -0,0 +1,369 @@
+/*-------------------------------------------------------------------------
+ *
+ * foreign_key_utils.c
```
marcocitus: I don't like when we put stuff in the utils directory that implements core database logic. 

Maybe it's time we introduce a DDL directory?

furkansahin: Yes, this would be good. 
Would you like me to create a ddl directory in the same level with planner, executor, connection etc.? 

Additionally, what should we put in it besides foreign_key_utils.c? If we are going to find and relocate every DDL related functions under this newly created directory, maybe I should do this in another PR.

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -0,0 +1,369 @@
+/*-------------------------------------------------------------------------
+ *
+ * foreign_constraint.c
+ *
+ * This file contains functions to perform useful operations on foreign
+ * keys.
+ *
+ * Copyright (c) 2014-2016, Citus Data, Inc.
```
onderkalaci: Two minor notes:

1. We should probably use 2018.

2. `useful operations` seems too broad. Maybe: `This file contains functions to create, alter and drop foreign constraints on distributed tables.`

src/backend/distributed/executor/multi_utility.c
```diff
@@ -37,6 +37,7 @@
 #include "commands/prepare.h"
 #include "distributed/citus_ruleutils.h"
 #include "distributed/colocation_utils.h"
+#include "distributed/foreign_constraint.h"
 #include "distributed/intermediate_results.h"
```
onderkalaci: Don't we need to remove some of the header includes given that we're removing lots of C code?

src/include/distributed/foreign_constraint.h
```diff
@@ -0,0 +1,24 @@
...
+#include "postgres.h"
+#include "postgres_ext.h"
+#include "utils/relcache.h"
+#include "nodes/primnodes.h"
+
+extern void ErrorIfUnsupportedForeignConstraint(Relation relation, char
+												distributionMethod,
+												Var *distributionColumn, uint32
+												colocationId);
+extern List * GetTableForeignConstraintCommands(Oid relationId);
```
onderkalaci: Minor: Wrt to the coding conventions, we should probably switch the order of these two functions: `GetTableForeignConstraintCommands ()` and `TableReferenced ()`

furkansahin:
This PR is on foreign_master. I am closing this one to merge the changes directly to master.

PR #2202, opened 2018-06-06
create_distributed_table with fk to reference table is implemented 

furkansahin:
This PR adds support for create_distributed_table on tables with foreign keys which references to reference tables.

Additionally, some of the tests from #2201 are changed to use create_distributed_table for tables with already defined foreign keys instead of adding them with `ALTER TABLE`.

### DIFF
--- a/src/backend/distributed/ddl/foreign_constraint.c
+++ b/src/backend/distributed/ddl/foreign_constraint.c
@@ -371,6 +371,62 @@ GetTableForeignConstraintCommands(Oid relationId)
 
 
 /*
+ * HasForeignKeyToReferenceTable function scans the pgConstraint table to
+ * fetch all of the constraints on the given relationId and see if at least one
+ * of them is a foreign key referencing to a reference table.
+ */
+bool
+HasForeignKeyToReferenceTable(Oid relationId)
+{
+	Relation pgConstraint = NULL;
+	SysScanDesc scanDescriptor = NULL;
+	ScanKeyData scanKey[1];
+	int scanKeyCount = 1;
+	HeapTuple heapTuple = NULL;
+	bool hasForeignKeyToReferenceTable = false;
+
+	pgConstraint = heap_open(ConstraintRelationId, AccessShareLock);
+	ScanKeyInit(&scanKey[0], Anum_pg_constraint_conrelid, BTEqualStrategyNumber, F_OIDEQ,
+				relationId);
+	scanDescriptor = systable_beginscan(pgConstraint, ConstraintRelidIndexId, true, NULL,
+										scanKeyCount, scanKey);
+
+	heapTuple = systable_getnext(scanDescriptor);
+	while (HeapTupleIsValid(heapTuple))
+	{
+		Oid referencedTableId = InvalidOid;
+		Form_pg_constraint constraintForm = (Form_pg_constraint) GETSTRUCT(heapTuple);
+
+		if (constraintForm->contype != CONSTRAINT_FOREIGN)
+		{
+			heapTuple = systable_getnext(scanDescriptor);
+			continue;
+		}
+
+		referencedTableId = constraintForm->confrelid;
+
+		if (!IsDistributedTable(referencedTableId))
+		{
+			continue;
+		}
+
+		if (PartitionMethod(referencedTableId) == DISTRIBUTE_BY_NONE)
+		{
+			hasForeignKeyToReferenceTable = true;
+			break;
+		}
+
+		heapTuple = systable_getnext(scanDescriptor);
+	}
+
+	/* clean up scan and close system catalog */
+	systable_endscan(scanDescriptor);
+	heap_close(pgConstraint, NoLock);
+	return hasForeignKeyToReferenceTable;
+}
+
+
+/*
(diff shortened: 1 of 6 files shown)


marcocitus:
Can you write more specific PR titles and descriptions?

create_distributed_table already creates foreign keys on the shards, what's specific about this PR?

src/test/regress/expected/foreign_key_to_reference_table.out
```diff
@@ -342,7 +321,7 @@ SELECT run_command_on_workers($$CREATE TYPE fkey_reference_table.composite AS (k
 (2 rows)
 
 CREATE TABLE referenced_table(test_column composite, PRIMARY KEY(test_column));
-CREATE TABLE referencing_table(id int, referencing_composite composite);
+CREATE TABLE referencing_table(id int, referencing_composite composite, FOREIGN KEY (referencing_composite) REFERENCES referenced_table(test_column) ON DELETE CASCADE);
```
onderkalaci: I'm not a big fan of converting `ALTER TABLE` commands to fkey in `CREATE TABLE`. Our coverage for `ALTER TABLE` is obviously decreasing.

src/test/regress/expected/foreign_key_to_reference_table.out
```diff
@@ -983,6 +1295,125 @@ NOTICE:  drop cascades to constraint fkey_ref on table referencing_table
...
+ 
+(1 row)
+
+  DROP TABLE test_table_1 CASCADE;
+NOTICE:  drop cascades to 2 other objects
+DETAIL:  drop cascades to constraint test_table_2_id_fkey on table test_table_2
+drop cascades to constraint test_table_3_value_1_fkey on table test_table_3
+ROLLBACK;
+-- create_reference_table, create_distributed_table and ALTER TABLE in the same transaction
+-- fails for now
```
onderkalaci: Minor: Just to find easier in the next PR, could you add a prefix  such as: `FIXME: fails for now` or such?

src/test/regress/sql/foreign_key_to_reference_table.sql
```diff
@@ -467,6 +565,23 @@ DROP TABLE referenced_table CASCADE;
 DROP TABLE referencing_table CASCADE;
 DROP TABLE referencing_table2 CASCADE;
 
+-- Check if the same fkeys are created with create_distributed_table
```
onderkalaci: I think you mean `multiple fkeys` instead of `the same` ? Can we at least get the  `count` from  `table_fkeys_in_workers`?

src/test/regress/sql/foreign_key_to_reference_table.sql
```diff
@@ -467,6 +565,23 @@ DROP TABLE referenced_table CASCADE;
...
+CREATE TABLE referenced_table(test_column int, test_column2 int UNIQUE, PRIMARY KEY(test_column));
+CREATE TABLE referencing_table(id int PRIMARY KEY, ref_id int, FOREIGN KEY (id) REFERENCES referenced_table(test_column) ON DELETE CASCADE);
+CREATE TABLE referencing_table2(id int, ref_id int, FOREIGN KEY (ref_id) REFERENCES referenced_table(test_column2) ON DELETE CASCADE, FOREIGN KEY (id) REFERENCES referencing_table(id) ON DELETE CASCADE);
+SELECT create_reference_table('referenced_table');
+BEGIN;
+  SET LOCAL citus.multi_shard_modify_mode TO 'sequential';
```
onderkalaci: why do we set the mode to sequential? I think it's OK to have a single test with `sequential ` mode, but just manually ensure that this test passes without explicit set as well.

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -370,6 +370,59 @@ GetTableForeignConstraintCommands(Oid relationId)
...
+
+		if (constraintForm->contype != CONSTRAINT_FOREIGN)
+		{
+			heapTuple = systable_getnext(scanDescriptor);
+			continue;
+		}
+
+		referencedTableId = constraintForm->confrelid;
+
+		Assert(IsDistributedTable(referencedTableId));
```
onderkalaci: I can see that this assert is always true for today's fkey implementation. But, shall we do something like this:
```C
if !(IsDistributedTable(referencedTableId))
  continue;
```

src/backend/distributed/master/master_stage_protocol.c
```diff
@@ -633,6 +633,13 @@ WorkerCreateShard(Oid relationId, int shardIndex, uint64 shardId, List *ddlComma
 		{
 			referencedShardId = shardId;
 		}
+		else if (PartitionMethod(referencedRelationId) == DISTRIBUTE_BY_NONE)
```
onderkalaci: minor: Can you add a small explanation to the above comment about this case as well?

src/test/regress/sql/foreign_key_to_reference_table.sql
furkansahin: In any case, changed them to count since the order by makes a difference in appveyor..

PR #2203, opened 2018-06-07
Add infrastructure for tracking relation accesses within a transaction block

onderkalaci:
This is a pre-requisite for foreign keys from distributed tables to
reference tables (see #968). This should be implemented very efficiently
since we're going to get the list of relations accessed per query execution
within a transaction block.

I've done some benchmarking with the queries in this gist file: https://gist.github.com/onderkalaci/c43d34c590e659df676b021a3babe59a

There are 5 different test cases, and I've run using pgbench
 (` pgbench -j128 -c 128 -T 60  -f test5.sql`). The function is
manually injected in `FinalizePlan()` function, with the following call `RelationAccessedInTransactionBlock(distributedPlan->relationIdList);`. So
basically its a simulation of what we're planning to implement in the executor.

Note that in the current implementation, `RelationAccessedInTransactionBlock()` returns
as soon as a relation is found. **However, in my benchmarks, I've removed those early 
exists and let the function iterate always over all the locks to measure the worst cases.**

Summary of the results: We're not adding any measurable overhead:

```
test1:

	master:
		latency average = 2.747 ms
		tps = 46591.448168 (including connections establishing)
	branch: 
		latency average = 2.737 ms
		tps = 46770.381859 (including connections establishing)

test2:

	master:
		latency average = 1.918 ms
		tps = 66743.596836 (including connections establishing)

	branch:
		latency average = 1.942 ms
		tps = 65925.051692 (including connections establishing)

test3:

	master:
		latency average = 7.598 ms
		tps = 16847.514636 (including connections establishing)

	branch:
		latency average = 7.724 ms
		tps = 16571.969544 (including connections establishing)
		
test4:

	master:
		latency average = 7.741 ms
		tps = 16536.173865 (including connections establishing)

	branch:
		latency average = 7.836 ms
		tps = 16334.710307 (including connections establishing)

test5:

	master:
		latency average = 782.241 ms
		tps = 163.632361 (including connections establishing)

	branch:
		latency average = 798.815 ms
		tps = 160.237282 (including connections establishing)

```

Do you see any other meaningful tests we want to perform?

### DIFF
--- a/src/backend/distributed/transaction/access_tracking.c
+++ b/src/backend/distributed/transaction/access_tracking.c
@@ -1,0 +1,153 @@
+/*-------------------------------------------------------------------------
+ *
+ * access_tracking.c
+ *
+ *   Transaction access tracking for Citus. The functions in this file
+ *   are intended to track the relation accesses within a transaction. The
+ *   logic here is mostly useful when a reference table is referred by
+ *   a distributed table via a foreign key. Whenever such a pair of tables
+ *   are acccesed inside a transaction, Citus should detect and act
+ *   accordingly.
+ *
+ * Copyright (c) 2018, Citus Data, Inc.
+ *
+ *-------------------------------------------------------------------------
+ */
+
+#include "postgres.h"
+
+#include "miscadmin.h"
+
+
+#include "access/transam.h"
+#include "access/xact.h"
+#include "distributed/access_tracking.h"
+#include "storage/lwlock.h"
+#include "storage/proc.h"
+#include "utils/lsyscache.h"
+
+
+/* the following macros are copied from lmgr/lock.c */
+#define FAST_PATH_BITS_PER_SLOT 3
+#define FAST_PATH_LOCKNUMBER_OFFSET 1
+#define FAST_PATH_MASK ((1 << FAST_PATH_BITS_PER_SLOT) - 1)
+#define FAST_PATH_GET_BITS(proc, n) \
+	(((proc)->fpLockBits >> (FAST_PATH_BITS_PER_SLOT * n)) & FAST_PATH_MASK)
+
+
+/* this function is only exported in the regression tests */
+PG_FUNCTION_INFO_V1(relation_accessed_in_transaction_block);
+
+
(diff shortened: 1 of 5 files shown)


src/backend/distributed/transaction/access_tracking.c
```diff
@@ -0,0 +1,153 @@
+/*-------------------------------------------------------------------------
+ *
+ * access_tracking.c
```
onderkalaci: I might consider renaming  the file to `relation_access_tracking` or such

src/include/distributed/access_tracking.h
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * Copyright (c) 2018, Citus Data, Inc.
+ */
+
+#ifndef ACCESS_TRACKING_H_
+#define ACCESS_TRACKING_H_
+
+#include "nodes/pg_list.h"
+
+extern bool RelationAccessedInTransactionBlock(List *relationIds);
```
onderkalaci: We actually don't need to make this function public.

onderkalaci:
We've decided to close this PR and continue with #2220 for the following reasons:


1) The codes that get the relation locks seems correct in the high-level (e.g., fast path + MyProc locks)
2) There is a high chance that in PG 12 the lock structures can change. So, try avoiding to rely on these very low level data structures
3) Never call list_member_oid(), especially while holding locks. In general, do very minimal things while holding those locks. [That's my bad. I was actually planning to optimize the code more, anyway]
4) LockHashPartitionLockByIndex() could be very expensive in multi-socket systems. Benchmark the changes with high concurrency, high loads and increased max_locks_per_transaction, max_connections on such a system.
5) This approach is likely to become problematic with clusters that have many partitioned tables
6) Althgouh you're reading your own Proc's locks, you still have to call LockHashPartitionLockByIndex(), which might be a problem
7) Not keeping track of the LOCK_LEVELs leads to a more restrictive approach [We already know that]. And, we might eventually want to expand the coverage and would need to change the codes. So better to do it now

PR #2204, opened 2018-06-07
DDLs affecting foreign keys to reference tables are serialized

furkansahin:
This PR serializes some of the DDL operations related to foreign keys from hash distributed to reference tables.

- [x] `DROP CONSTRAINT`
- [x] `DROP COLUMN`
- [x] `ALTER COLUMN id SET DATA TYPE `
- [x] `TRUNCATE ... CASCADE`

We need to serialize these operations because they either `CASCADE` to the distributed table or executes a parallel `SELECT` on the distributed table to do some checks before adding the foreign keys.

### DIFF
--- a/src/backend/distributed/ddl/foreign_constraint.c
+++ b/src/backend/distributed/ddl/foreign_constraint.c
@@ -22,10 +22,73 @@
 #include "distributed/master_protocol.h"
 #include "distributed/multi_join_order.h"
 #include "utils/fmgroids.h"
+#include "utils/lsyscache.h"
 #include "utils/rel.h"
 #include "utils/relcache.h"
 #include "utils/ruleutils.h"
 #include "utils/syscache.h"
+
+
+static bool HeapTupleOfForeignConstraintIncludesColumn(HeapTuple heapTuple, Oid
+													   relationId, int pgConstraintKey,
+													   char *columnName);
+
+/*
+ * ConstraintIsAForeignKeyToReferenceTable function scans the pgConstraint to
+ * fetch all of the constraints on the given relationId and see if at least one
+ * of them is a foreign key referencing to a reference table.
+ */
+bool
+ConstraintIsAForeignKeyToReferenceTable(char *constraintName, Oid relationId)
+{
+	Relation pgConstraint = NULL;
+	SysScanDesc scanDescriptor = NULL;
+	ScanKeyData scanKey[1];
+	int scanKeyCount = 1;
+	HeapTuple heapTuple = NULL;
+	bool foreignKeyToReferenceTable = false;
+
+
+	pgConstraint = heap_open(ConstraintRelationId, AccessShareLock);
+
+	ScanKeyInit(&scanKey[0], Anum_pg_constraint_contype, BTEqualStrategyNumber, F_CHAREQ,
+				CharGetDatum(CONSTRAINT_FOREIGN));
+	scanDescriptor = systable_beginscan(pgConstraint, InvalidOid, false,
+										NULL, scanKeyCount, scanKey);
+
+	heapTuple = systable_getnext(scanDescriptor);
+	while (HeapTupleIsValid(heapTuple))
+	{
+		Oid referencedTableId = InvalidOid;
+		Form_pg_constraint constraintForm = (Form_pg_constraint) GETSTRUCT(heapTuple);
+		char *constraintName = (constraintForm->conname).data;
+
+		if (strncmp(constraintName, constraintName, NAMEDATALEN) != 0 ||
+			constraintForm->conrelid != relationId)
+		{
+			heapTuple = systable_getnext(scanDescriptor);
+			continue;
+		}
+
+		referencedTableId = constraintForm->confrelid;
+
+		Assert(IsDistributedTable(referencedTableId));
+
+		if (PartitionMethod(referencedTableId) == DISTRIBUTE_BY_NONE)
+		{
+			foreignKeyToReferenceTable = true;
+			break;
+		}
+
+		heapTuple = systable_getnext(scanDescriptor);
+	}
+
+	/* clean up scan and close system catalog */
+	systable_endscan(scanDescriptor);
+	heap_close(pgConstraint, AccessShareLock);
+
+	return foreignKeyToReferenceTable;
+}
 
 
 /*
@@ -60,11 +123,11 @@ ErrorIfUnsupportedForeignConstraint(Relation relation, char distributionMethod,
 	uint32 referencedTableColocationId = INVALID_COLOCATION_ID;
 	Var *referencedTablePartitionColumn = NULL;
 
-	Datum referencingColumnsDatum;
-	Datum *referencingColumnArray;
+	Datum referencingColumnsDatum = 0;
+	Datum *referencingColumnArray = NULL;
 	int referencingColumnCount = 0;
-	Datum referencedColumnsDatum;
-	Datum *referencedColumnArray;
+	Datum referencedColumnsDatum = 0;
+	Datum *referencedColumnArray = NULL;
 	int referencedColumnCount = 0;
 	bool isNull = false;
 	int attrIdx = 0;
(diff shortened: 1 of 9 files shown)


src/backend/distributed/executor/multi_utility.c
```diff
@@ -1374,17 +1392,16 @@ PlanAlterTableStmt(AlterTableStmt *alterTableStatement, const char *alterTableCo
...
 	ddlJob->commandString = alterTableCommand;
+	ddlJob->executeSequentially = executeSequentially;
 
 	if (rightRelationId)
 	{
 		/* if foreign key related, use specialized task list function ... */
 		ddlJob->taskList = InterShardDDLTaskList(leftRelationId, rightRelationId,
 												 alterTableCommand);
 
-		if (PartitionMethod(rightRelationId) == DISTRIBUTE_BY_NONE)
```
onderkalaci: This line of change seems harder to read. Can we not do that?

onderkalaci:
I think we should cover `ALTER TABLE ref_table ALTER COLUMN ...` where we touch one of the fkey columns. For example, 

```SQL
ALTER TABLE referece_table_second ALTER COLUMN id SET DATA TYPE bigint;
ERROR:  cannot perform DDL on placement 1230, which has been read over multiple connections
```

since Postgres sends to shards:
```SQL
LOG:  issuing COPY (SELECT fk.value_1 FROM (test.on_delete_fkey_table_103133 fk LEFT JOIN test.referece_table_second_103100 pk ON ((pk.id OPERATOR(pg_catalog.=) fk.value_1))) WHERE ((fk.value_1 IS NOT NULL) AND (pk.id IS NULL))) TO STDOUT
```

onderkalaci:
We're now able to do `TRUNCATE reference_table CASCADE`, by calling the `SetLocalMutltiShardModifyMode()` under the covers. So, we should include in this PR.

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -309,6 +369,94 @@ ErrorIfUnsupportedForeignConstraint(Relation relation, char distributionMethod,
 }
 
 
+bool
+ForeignKeyExistsToReferenceTableOnColumn(char *affectedColumnName,
```
onderkalaci: 1) Missing function comment.
2) maybe `affectedColumnName ` - > `columnName`. 
3) name seems to fit to a single line, so no need to have a new line?

Maybe more importantly, the name seems like not covering all cases, because the function returns true if it finds the key in the reference table as well? Seems like ` `ColumnAppearInForeignKey()` or such more generic names fit better

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -22,12 +22,72 @@
...
+{
+	Relation pgConstraint = NULL;
+	SysScanDesc scanDescriptor = NULL;
+	ScanKeyData scanKey[2];
+	int scanKeyCount = 2;
+	HeapTuple heapTuple = NULL;
+	bool foreignKeyToReferenceTable = false;
+
+	pgConstraint = heap_open(ConstraintRelationId, AccessShareLock);
+	ScanKeyInit(&scanKey[0],
```
onderkalaci: Pretty minor, but we seem to have many unnecessary new lines in `ScanKeyInit ` calls, could we have less lines? (in the other function as well)

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -309,6 +369,94 @@ ErrorIfUnsupportedForeignConstraint(Relation relation, char distributionMethod,
...
+										 Oid relationId)
+{
+	Relation pgConstraint = NULL;
+	SysScanDesc scanDescriptor = NULL;
+	ScanKeyData scanKey[1];
+	int scanKeyCount = 1;
+	HeapTuple heapTuple = NULL;
+	bool foreignKeyToReferenceTableIncludesGivenColumn = false;
+	bool isNull = false;
+	Datum columnsDatum;
```
onderkalaci: Minor convention notes: There are several uninitialized variables here and in the other places. Could you please go over those?

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -309,6 +369,94 @@ ErrorIfUnsupportedForeignConstraint(Relation relation, char distributionMethod,
...
+	int attrIdx;
+
+	pgConstraint = heap_open(ConstraintRelationId, AccessShareLock);
+
+	ScanKeyInit(&scanKey[0],
+				Anum_pg_constraint_connamespace,
+				BTEqualStrategyNumber, F_OIDEQ,
+				get_rel_namespace(relationId));
+
+	scanDescriptor = systable_beginscan(pgConstraint, ConstraintNameNspIndexId, true,
```
onderkalaci: We're not providing the constraint name to the index scan. This seems to have some weird effects (probably wrong results or assertion failures). 

Here it seems safer to use `contype` as the index and skip unrelated relations in the loop below.

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -22,12 +22,72 @@
...
+			break;
+		}
+
+		heapTuple = systable_getnext(scanDescriptor);
+	}
+
+	/* clean up scan and close system catalog */
+	systable_endscan(scanDescriptor);
+	heap_close(pgConstraint, AccessShareLock);
+	return foreignKeyToReferenceTable;
```
onderkalaci: minor: maybe a new line before return

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -309,6 +369,94 @@ ErrorIfUnsupportedForeignConstraint(Relation relation, char distributionMethod,
...
+						  's', &columnArray, NULL, &columnCount);
+
+		if (PartitionMethod(referencedTableId) == DISTRIBUTE_BY_NONE)
+		{
+			for (attrIdx = 0; attrIdx < columnCount; ++attrIdx)
+			{
+				AttrNumber attrNo = DatumGetInt16(columnArray[attrIdx]);
+
+				char *colName = get_relid_attribute_name(relationId, attrNo);
+				if (strcmp(colName, affectedColumnName) == 0)
```
onderkalaci: Let's use `strncmp`

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -309,6 +369,94 @@ ErrorIfUnsupportedForeignConstraint(Relation relation, char distributionMethod,
...
+		if (referencedTableId == relationId)
+		{
+			pgConstraintKey = Anum_pg_constraint_confkey;
+		}
+		else if (referencingTableId == relationId)
+		{
+			pgConstraintKey = Anum_pg_constraint_conkey;
+		}
+
+		columnsDatum = SysCacheGetAttr(CONSTROID, heapTuple,
```
onderkalaci: Why do we run this for `if (PartitionMethod(referencedTableId) != DISTRIBUTE_BY_NONE)`. We should probably add a `continue` before doing these

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -309,6 +369,94 @@ ErrorIfUnsupportedForeignConstraint(Relation relation, char distributionMethod,
...
+									   pgConstraintKey, &isNull);
+		deconstruct_array(DatumGetArrayTypeP(columnsDatum), INT2OID, 2, true,
+						  's', &columnArray, NULL, &columnCount);
```
onderkalaci: I think the calculation for `columnCount` is wrong. It's only the number of columns in the fkey. However, we should get the column count of the relation

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -309,6 +369,94 @@ ErrorIfUnsupportedForeignConstraint(Relation relation, char distributionMethod,
...
+		}
+
+		referencedTableId = constraintForm->confrelid;
+		referencingTableId = constraintForm->conrelid;
+
+		if (referencedTableId == relationId)
+		{
+			pgConstraintKey = Anum_pg_constraint_confkey;
+		}
+		else if (referencingTableId == relationId)
```
onderkalaci: What happens in the `else` clause? I think we should simply `continue`

src/backend/distributed/executor/multi_utility.c
```diff
@@ -335,6 +342,11 @@ multi_ProcessUtility(PlannedStmt *pstmt,
 	if (IsA(parsetree, TruncateStmt))
 	{
 		ErrorIfUnsupportedTruncateStmt((TruncateStmt *) parsetree);
+
+		if (ShouldExecuteTruncateStmtSequential((TruncateStmt *) parsetree))
```
onderkalaci: I think this logic should live in `master_modify_multiple_shards` because we're already doing something related there and if we put it here the logic would be scattered:

```C
f (ShouldExecuteTruncateStmtSequential((TruncateStmt *) parsetree))
{
	SetLocalMultiShardModifyModeToSequential();
}

... and the code continues with
{

}

File: src/backend/distributed/executor/multi_utility.c [hunk 12]
```diff
@@ -3787,3 +3804,78 @@ ProcessDropTableStmt(DropStmt *dropTableStatement)
 		}
 	}
 }
+
+
+/*
+ * ShouldExecuteAlterTableSequentially checks if the given ALTER TABLE
```
onderkalaci: I think we should put all the commands that need sequential execution into here. So, we should move the logic below to here, otherwise, it'd be very hard to follow when we switch to seq, mode

```
		/*
		 * We need to execute the ddls working with reference tables on the
		 * right side sequentially, because parallel ddl operations
		 * relating to one and only shard of a reference table on a worker
		 * may cause self-deadlocks.
		 */
		if (PartitionMethod(rightRelationId) == DISTRIBUTE_BY_NONE)
		{
			ddlJob->executeSequentially = true;
		}
```

I haven't looked into detail on how to move that part here, but, feel free to discuss it is hard to implement.

(Wild guess: Is it simply checking for `AT_AddConstraint` ?)

src/backend/distributed/executor/multi_utility.c
```diff
@@ -3787,3 +3804,78 @@ ProcessDropTableStmt(DropStmt *dropTableStatement)
...
+	else if (alterTableType == AT_DropColumn || alterTableType == AT_AlterColumnType)
+	{
+		char *affectedColumnName = command->name;
+
+		/*
+		 * We need to set multi_shard_modify_mode to sequential since SET DATA TYPE
+		 * command issues one select and one DDL to see if there is incompatible
+		 * data for the new column type.
+		 */
+		if (IsTransactionBlock() && alterTableType == AT_AlterColumnType)
```
onderkalaci: Why do we `SetLocalMultiShardModifyModeToSequential ()` before knowing about `ForeignKeyExistsToReferenceTableOnColumn ()` ?

Also, if I execute the following where there is no reference table/fkey, I get an error

```SQL
 alter table users_table  alter COLUMN value_1 SET DATA TYPE int;
ERROR:  invalid attnum: 0
```

src/test/regress/expected/foreign_key_to_reference_table.out
```diff
@@ -1357,6 +1355,414 @@ ERROR:  cannot establish a new connection for placement 7000388, since DDL has b
...
+ 
+(1 row)
+
+SELECT create_distributed_table('test_table_2', 'id');
+ create_distributed_table 
+--------------------------
+ 
+(1 row)
+
+INSERT INTO test_table_1 VALUES (1,1), (2,2), (3,3);
```
onderkalaci: Can you try a case there `SET DATA TYPE` fails since say there is already an int which is `MAX_INT_64 - 1`

src/backend/distributed/executor/multi_utility.c
```diff
@@ -3787,3 +3804,78 @@ ProcessDropTableStmt(DropStmt *dropTableStatement)
 		}
 	}
 }
+
+
+/*
+ * ShouldExecuteAlterTableSequentially checks if the given ALTER TABLE
+ * statements should be executed sequentially if there is a foreign
```
onderkalaci: `if there is a foreign` ->`when there is a foreign`

src/backend/distributed/ddl/foreign_constraint.c
furkansahin: I don't agree. we should get the number of columns in the key because we use this count as a index while iterating on the columns in the key, not in the table. We get the columnArray from the same function call and use it in line 435.

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -309,6 +363,94 @@ ErrorIfUnsupportedForeignConstraint(Relation relation, char distributionMethod,
...
+		}
+
+		if (pgConstraintKey == 0 || (IsDistributedTable(referencedTableId) &&
+				PartitionMethod(referencedTableId) != DISTRIBUTE_BY_NONE))
+		{
+			heapTuple = systable_getnext(scanDescriptor);
+			continue;
+		}
+
+		columnsDatum = SysCacheGetAttr(CONSTROID, heapTuple,
```
onderkalaci: if possible lets move this logic to a function

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -22,12 +22,66 @@
...
+ConstraintIsAForeignKeyToReferenceTable(char *name, Oid relationId)
+{
+	Relation pgConstraint = NULL;
+	SysScanDesc scanDescriptor = NULL;
+	ScanKeyData scanKey[1];
+	int scanKeyCount = 1;
+	HeapTuple heapTuple = NULL;
+	bool foreignKeyToReferenceTable = false;
+
+	pgConstraint = heap_open(ConstraintRelationId, AccessShareLock);
```
onderkalaci: not to forget: missing param to the scanKey

src/backend/distributed/master/master_modify_multiple_shards.c
```diff
@@ -235,3 +242,34 @@ ModifyMultipleShardsTaskList(Query *query, List *shardIntervalList)
 
 	return taskList;
 }
+
+
+/*
+ * CheckIfNeedsToRunSequentialTruncateStmt decides if the TRUNCATE stmt needs
```
onderkalaci: function name is obselete

mtuncer:
@furkansahin @onderkalaci missing DESCRIPTION

furkansahin:
@mtuncer @onderkalaci I don't think this needs a description. I think we only need description on #2240 for anything related to foreign keys, rest does not need to be in the changelog.

PR #2210, opened 2018-06-12
Added failure test for create index concurrently

mtuncer:


src/test/regress/sql/failure_create_index_concurrently.sql
```diff
@@ -0,0 +1,65 @@
+--
```
onderkalaci: Considering this [comment](https://github.com/citusdata/citus/pull/2244#issuecomment-402990403) and thinking about this PR as the main one for bare DDLs, I could suggest three more tests:

- Test with replication factor 1 as well (I think the default in the regression test suite is two, but, please double check that)
- Add tests for DROP INDEX CONCURRENTLY as well.
- Since the `CREATE/DROP INDEX CONCURRENTLY` uses sequential execution mode, we could kill/cancel the second (or later) commands over a single connection. For example, if you create 8 shards, we'd end up with 4 shards per worker and 4 shards are modified over the same connection per node. So, say if you have `whatevercase().after(2).kill()`, it'd kill the connection after seeing the same command 2 times. An example can be found [here](https://github.com/citusdata/citus/pull/2212/files#diff-1490fb849241ff1456eb927ed102394aR1194)

src/test/regress/expected/failure_create_index_concurrently.out
```diff
@@ -0,0 +1,153 @@
...
+SELECT citus.dump_network_traffic();
+                                                                                                                                                                                                                                                                       dump_network_traffic                                                                                                                                                                                                                                                                        
+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```
onderkalaci: This is a discussion we've made with @lithp  as well, but we haven't concluded the discussion.

I think this `dump_network_traffic ()` is super useful by itself both during development and debugging. But, do you see necessary to add to regression tests? I'm OK with keeping it, but I don't add them to the PRs that I open because they seem too verbose to add to the output. I even hit few cases where the output isn't consistent such as [here](https://github.com/citusdata/citus/pull/2244/commits/c723121d1d4f4962d2aaeb498635f8b979adab5e#r199416333). (I don't expect inconsistencies for this case, but, just wanted to note)

src/test/regress/expected/failure_create_index_concurrently.out
```diff
@@ -0,0 +1,153 @@
...
+-- cancel the connection when create command is issued
+-- network traffic may differ between execution during cancellation
+-- therefore dump_network_traffic() calls are not made
+SELECT citus.mitmproxy('conn.onQuery(query="CREATE").cancel(' || pg_backend_pid() || ')');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT citus.clear_network_traffic();
```
lithp: There's no need to call `clear_network_traffic()` if you're not going to dump afterwards. All it does it empty the buffer of captured packets.

src/test/regress/expected/failure_create_index_concurrently.out
lithp: I think you're right that they're too verbose to add to every line. A style that I think makes sense is to show the dump just once at the top, run a happy-case so future readers can see exactly what the rest of the test is mucking with. These two dumps don't really show me what is supposed to happen, they just show the same truncated log twice.

I don't want to impose that style on anyone though if they don't agree with it.

mtuncer:
- removed clear/dump network traffic calls
- created tables in a schema to allow parallel execution
- added drop index concurrently test
- added after(1) calls to allow some commands to succeed, rest to fail

mtuncer:
@lithp @onderkalaci I am not really enjoying `worker_2_port + 2` to describe `nodeport` Can we define something like `worker_2_proxy` or `worker_2_proxy_port` in the test infrastructure

src/test/regress/sql/failure_create_index_concurrently.sql
```diff
@@ -0,0 +1,112 @@
...
+WHERE nodeport = :worker_2_proxy_port;
+
+RESET SEARCH_PATH;
+DROP SCHEMA index_schema CASCADE;
+
+-- verify index is not at worker 2 upon cleanup
+SELECT * FROM run_command_on_workers($$SELECT count(*) FROM pg_indexes WHERE indexname LIKE 'idx_index_test%' $$)
+WHERE nodeport = :worker_2_proxy_port;
+
+
```
onderkalaci: Pretty minor: extra new lines

src/test/regress/expected/failure_create_index_concurrently.out
```diff
@@ -0,0 +1,200 @@
...
+HINT:  Use DROP INDEX CONCURRENTLY IF EXISTS to remove the invalid index, then retry the original command.
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+-- verify index is not created
+SELECT * FROM run_command_on_workers($$SELECT count(*) FROM pg_indexes WHERE indexname LIKE 'idx_index_test%' $$)
+WHERE nodeport = :worker_2_proxy_port;
```
lithp: Another way would have been to write `WHERE nodeport != :worker_1_port`

lithp: We could also change :worker_2_port and have it always point at the proxy, that sounds like a cleaner option.

src/test/regress/pg_regress_multi.pl
```diff
@@ -417,6 +419,7 @@ sub revert_replace_postgres
 }
 print $fh catfile($bindir, "psql")." ";
 print $fh "--variable=master_port=$masterPort ";
+print $fh "--variable=worker_2_proxy_port=$workerBehindProxyPort ";
```
lithp: We could also change `:worker_2_port` and have it always point at the proxy, that sounds like a cleaner option.

PR #2211, opened 2018-06-12
Create foreign key relation graph and functions to query on it

velioglu:
In order to decide on the type of connection within the transaction that touching reference tables, we need to know which distributed tables are referenced by these reference tables.  

This PR provides a support for getting this information. PR creates both foreign key relation graph and functions to query on this graph. We utilized pg_constraint catalog table to create that graph. Relation can be queried over the graph to find relations referenced by that relation or relations referencing that relation. 

- [x] Add regression tests

Note that we're planning to add this information to distributed metadata cache with a follow-up PR.

### DIFF
--- a/src/backend/distributed/ddl/foreign_constraint.c
+++ b/src/backend/distributed/ddl/foreign_constraint.c
@@ -649,3 +649,97 @@ HeapTupleOfForeignConstraintIncludesColumn(HeapTuple heapTuple, Oid relationId,
 
 	return false;
 }
+
+
+/*
+ * TableReferencing function checks whether given table is referencing by another table
+ * via foreign constraints. If it is referencing, this function returns true. To check
+ * that, this function searches given relation at pg_constraints system catalog. However
+ * since there is no index for the column we searched, this function performs sequential
+ * search, therefore call this function with caution.
+ */
+bool
+TableReferencing(Oid relationId)
+{
+	Relation pgConstraint = NULL;
+	HeapTuple heapTuple = NULL;
+	SysScanDesc scanDescriptor = NULL;
+	ScanKeyData scanKey[1];
+	int scanKeyCount = 1;
+	Oid scanIndexId = InvalidOid;
+	bool useIndex = false;
+
+	pgConstraint = heap_open(ConstraintRelationId, AccessShareLock);
+
+	ScanKeyInit(&scanKey[0], Anum_pg_constraint_conrelid, BTEqualStrategyNumber, F_OIDEQ,
+				relationId);
+	scanDescriptor = systable_beginscan(pgConstraint, scanIndexId, useIndex, NULL,
+										scanKeyCount, scanKey);
+
+	heapTuple = systable_getnext(scanDescriptor);
+	while (HeapTupleIsValid(heapTuple))
+	{
+		Form_pg_constraint constraintForm = (Form_pg_constraint) GETSTRUCT(heapTuple);
+
+		if (constraintForm->contype == CONSTRAINT_FOREIGN)
+		{
+			systable_endscan(scanDescriptor);
+			heap_close(pgConstraint, NoLock);
+
+			return true;
+		}
+
+		heapTuple = systable_getnext(scanDescriptor);
+	}
+
+	systable_endscan(scanDescriptor);
+	heap_close(pgConstraint, NoLock);
+
+	return false;
+}
+
+
+/*
+ * ConstraintIsAForeignKey returns true if the given constraint name
+ * is a foreign key to defined on the relation.
+ */
+bool
+ConstraintIsAForeignKey(char *constraintNameInput, Oid relationId)
+{
+	Relation pgConstraint = NULL;
+	SysScanDesc scanDescriptor = NULL;
+	ScanKeyData scanKey[1];
+	int scanKeyCount = 1;
+	HeapTuple heapTuple = NULL;
+
+	pgConstraint = heap_open(ConstraintRelationId, AccessShareLock);
+
+	ScanKeyInit(&scanKey[0], Anum_pg_constraint_contype, BTEqualStrategyNumber, F_CHAREQ,
+				CharGetDatum(CONSTRAINT_FOREIGN));
+	scanDescriptor = systable_beginscan(pgConstraint, InvalidOid, false,
+										NULL, scanKeyCount, scanKey);
+
+	heapTuple = systable_getnext(scanDescriptor);
+	while (HeapTupleIsValid(heapTuple))
+	{
+		Form_pg_constraint constraintForm = (Form_pg_constraint) GETSTRUCT(heapTuple);
+		char *constraintName = (constraintForm->conname).data;
+
+		if (strncmp(constraintName, constraintNameInput, NAMEDATALEN) == 0 &&
+			constraintForm->conrelid == relationId)
+		{
+			systable_endscan(scanDescriptor);
+			heap_close(pgConstraint, AccessShareLock);
+
+			return true;
+		}
+
+		heapTuple = systable_getnext(scanDescriptor);
+	}
+
+	/* clean up scan and close system catalog */
+	systable_endscan(scanDescriptor);
+	heap_close(pgConstraint, AccessShareLock);
+
+	return false;
+}
--- a/src/backend/distributed/executor/multi_utility.c
+++ b/src/backend/distributed/executor/multi_utility.c
@@ -75,6 +75,7 @@
 #include "utils/builtins.h"
 #include "utils/elog.h"
 #include "utils/errcodes.h"
+#include "utils/fmgroids.h"
 #include "utils/guc.h"
 #include "utils/hsearch.h"
 #include "utils/inval.h"
@@ -85,6 +86,10 @@
 
 
 bool EnableDDLPropagation = true; /* ddl propagation is enabled */
+
+
+static bool shouldInvalidateForeignKeyGraph = false;
+
 
 /*
  * This struct defines the state for the callback for drop statements.
@@ -162,7 +167,11 @@ static void CheckCopyPermissions(CopyStmt *copyStatement);
 static List * CopyGetAttnums(TupleDesc tupDesc, Relation rel, List *attnamelist);
 static void PostProcessUtility(Node *parsetree);
 static List * CollectGrantTableIdList(GrantStmt *grantStmt);
+static char * GetSchemaNameFromDropObject(ListCell *dropSchemaCell);
 static void ProcessDropTableStmt(DropStmt *dropTableStatement);
+static void ProcessDropSchemaStmt(DropStmt *dropSchemaStatement);
+static void InvalidateForeignKeyGraphForDDL(void);
+
 
 /*
  * We need to run some of the commands sequentially if there is a foreign constraint
@@ -371,6 +380,11 @@ multi_ProcessUtility(PlannedStmt *pstmt,
 			{
 				ProcessDropTableStmt(dropStatement);
 			}
+
+			if (dropStatement->removeType == OBJECT_SCHEMA)
+			{
+				ProcessDropSchemaStmt(dropStatement);
+			}
 		}
 
 		if (IsA(parsetree, AlterTableStmt))
@@ -530,6 +544,16 @@ multi_ProcessUtility(PlannedStmt *pstmt,
 		SetUserIdAndSecContext(savedUserId, savedSecurityContext);
 	}
 
+	/*
+	 * Re-forming the foreign key graph relies on the command being executed
+	 * on the local table first. However, in order to decide whether the
+	 * command leads to an invalidation, we need to check before the command
+	 * is being executed since we read pg_constraint table. Thus, we maintain a
+	 * local flag and do the invalidation after multi_ProcessUtility,
+	 * before ExecuteDistributedDDLJob().
+	 */
+	InvalidateForeignKeyGraphForDDL();
+
 	/* after local command has completed, finish by executing worker DDLJobs, if any */
 	if (ddlJobs != NIL)
 	{
@@ -556,8 +580,27 @@ multi_ProcessUtility(PlannedStmt *pstmt,
 
 				if (alterTableType == AT_AddConstraint)
 				{
+					LOCKMODE lockmode = NoLock;
+					Oid relationId = InvalidOid;
+					Constraint *constraint = NULL;
+
 					Assert(list_length(commandList) == 1);
+
 					ErrorIfUnsupportedAlterAddConstraintStmt(alterTableStatement);
+
+					lockmode = AlterTableGetLockLevel(alterTableStatement->cmds);
+					relationId = AlterTableLookupRelation(alterTableStatement, lockmode);
+
+					if (!OidIsValid(relationId))
+					{
+						continue;
+					}
+
+					constraint = (Constraint *) command->def;
+					if (ConstraintIsAForeignKey(constraint->conname, relationId))
+					{
+						InvalidateForeignKeyGraph();
+					}
 				}
 			}
 		}
@@ -614,6 +657,22 @@ multi_ProcessUtility(PlannedStmt *pstmt,
 	 * EXTENSION. This is important to register some invalidation callbacks.
 	 */
 	CitusHasBeenLoaded();
+}
+
+
+/*
+ * InvalidateForeignKeyGraphForDDL simply keeps track of whether
+ * the foreign key graph should be invalidated due to a DDL.
+ */
+static void
+InvalidateForeignKeyGraphForDDL(void)
+{
+	if (shouldInvalidateForeignKeyGraph)
+	{
+		InvalidateForeignKeyGraph();
+
+		shouldInvalidateForeignKeyGraph = false;
+	}
 }
 
 
@@ -2244,8 +2303,25 @@ ErrorIfUnsupportedAlterTableStmt(AlterTableStmt *alterTableStatement)
 			}
 
 #endif
-			case AT_SetNotNull:
 			case AT_DropConstraint:
+			{
+				LOCKMODE lockmode = AlterTableGetLockLevel(alterTableStatement->cmds);
+				Oid relationId = AlterTableLookupRelation(alterTableStatement, lockmode);
+
+				if (!OidIsValid(relationId))
+				{
+					return;
+				}
+
+				if (ConstraintIsAForeignKey(command->name, relationId))
+				{
+					shouldInvalidateForeignKeyGraph = true;
+				}
+
+				break;
+			}
+
+			case AT_SetNotNull:
 			case AT_EnableTrigAll:
 			case AT_DisableTrigAll:
 			case AT_ReplicaIdentity:
@@ -3748,6 +3824,99 @@ RoleSpecString(RoleSpec *spec)
 
 
 /*
+ * ProcessDropSchemaStmt invalidates the foreign key cache if any table created
+ * under dropped schema involved in any foreign key relationship.
+ */
+static void
+ProcessDropSchemaStmt(DropStmt *dropStatement)
+{
+	Relation pgClass = NULL;
+	HeapTuple heapTuple = NULL;
+	SysScanDesc scanDescriptor = NULL;
+	ScanKeyData scanKey[1];
+	int scanKeyCount = 1;
+	Oid scanIndexId = InvalidOid;
+	bool useIndex = false;
+	ListCell *dropSchemaCell;
+
+	if (dropStatement->behavior != DROP_CASCADE)
+	{
+		return;
+	}
+
+	foreach(dropSchemaCell, dropStatement->objects)
+	{
+		char *schemaString = GetSchemaNameFromDropObject(dropSchemaCell);
+		Oid namespaceOid = get_namespace_oid(schemaString, true);
+
+		if (namespaceOid == InvalidOid)
+		{
+			continue;
+		}
+
+		pgClass = heap_open(RelationRelationId, AccessShareLock);
+
+		ScanKeyInit(&scanKey[0], Anum_pg_class_relnamespace, BTEqualStrategyNumber,
+					F_OIDEQ, namespaceOid);
+		scanDescriptor = systable_beginscan(pgClass, scanIndexId, useIndex, NULL,
+											scanKeyCount, scanKey);
+
+		heapTuple = systable_getnext(scanDescriptor);
+		while (HeapTupleIsValid(heapTuple))
+		{
+			Form_pg_class relationForm = (Form_pg_class) GETSTRUCT(heapTuple);
+			char *relationName = NameStr(relationForm->relname);
+			Oid relationId = get_relname_relid(relationName, namespaceOid);
+
+			/* we're not interested in non-valid, non-distributed relations */
+			if (relationId == InvalidOid || !IsDistributedTable(relationId))
+			{
+				heapTuple = systable_getnext(scanDescriptor);
+				continue;
+			}
+
+			/* invalidate foreign key cache if the table involved in any foreign key */
+			if (TableReferenced(relationId) || TableReferencing(relationId))
+			{
+				shouldInvalidateForeignKeyGraph = true;
+
+				systable_endscan(scanDescriptor);
+				heap_close(pgClass, NoLock);
+				return;
+			}
+
+			heapTuple = systable_getnext(scanDescriptor);
+		}
+
+		systable_endscan(scanDescriptor);
+		heap_close(pgClass, NoLock);
+	}
+}
+
+
+/*
+ * GetSchemaNameFromDropObject gets the name of the drop schema from given
+ * list cell. This function is defined due to API change between PG 9.6 and
+ * PG 10.
+ */
+static char *
+GetSchemaNameFromDropObject(ListCell *dropSchemaCell)
+{
+	char *schemaString = NULL;
+
+#if (PG_VERSION_NUM >= 100000)
+	Value *schemaValue = (Value *) lfirst(dropSchemaCell);
+	schemaString = strVal(schemaValue);
+#else
+	List *schemaNameList = (List *) lfirst(dropSchemaCell);
+	schemaString = NameListToString(schemaNameList);
+#endif
+
+	return schemaString;
+}
+
+
+/*
  * ProcessDropTableStmt processes DROP TABLE commands for partitioned tables.
  * If we are trying to DROP partitioned tables, we first need to go to MX nodes
  * and DETACH partitions from their parents. Otherwise, we process DROP command
@@ -3774,10 +3943,20 @@ ProcessDropTableStmt(DropStmt *dropTableStatement)
 
 		Oid relationId = RangeVarGetRelid(tableRangeVar, AccessShareLock, missingOK);
 
-		if (relationId == InvalidOid ||
-			!IsDistributedTable(relationId) ||
-			!ShouldSyncTableMetadata(relationId) ||
-			!PartitionedTable(relationId))
+		/* we're not interested in non-valid, non-distributed relations */
+		if (relationId == InvalidOid || !IsDistributedTable(relationId))
+		{
+			continue;
+		}
+
+		/* invalidate foreign key cache if the table involved in any foreign key */
+		if ((TableReferenced(relationId) || TableReferencing(relationId)))
+		{
+			shouldInvalidateForeignKeyGraph = true;
+		}
+
+		/* we're only interested in partitioned and mx tables */
+		if (!ShouldSyncTableMetadata(relationId) || !PartitionedTable(relationId))
 		{
 			continue;
 		}
--- a/src/backend/distributed/utils/metadata_cache.c
+++ b/src/backend/distributed/utils/metadata_cache.c
@@ -29,6 +29,7 @@
 #include "distributed/colocation_utils.h"
 #include "distributed/connection_management.h"
 #include "distributed/citus_ruleutils.h"
+#include "distributed/foreign_key_relationship.h"
 #include "distributed/master_metadata_utility.h"
 #include "distributed/metadata_cache.h"
 #include "distributed/multi_logical_optimizer.h"
@@ -44,6 +45,8 @@
 #include "distributed/worker_protocol.h"
 #include "executor/executor.h"
 #include "nodes/makefuncs.h"
+#include "nodes/memnodes.h"
+#include "nodes/pg_list.h"
 #include "parser/parse_func.h"
 #include "parser/parse_type.h"
 #include "storage/lmgr.h"
@@ -55,6 +58,7 @@
 #include "utils/inval.h"
 #include "utils/fmgroids.h"
 #include "utils/lsyscache.h"
+#include "utils/palloc.h"
 #include "utils/rel.h"
 #include "utils/relfilenodemap.h"
 #include "utils/relmapper.h"
@@ -182,11 +186,13 @@ static bool HasOverlappingShardInterval(ShardInterval **shardIntervalArray,
 static void InitializeCaches(void);
 static void InitializeDistTableCache(void);
 static void InitializeWorkerNodeCache(void);
+static void RegisterForeignKeyGraphCacheCallbacks(void);
 static void RegisterWorkerNodeCacheCallbacks(void);
 static void RegisterLocalGroupIdCacheCallbacks(void);
 static uint32 WorkerNodeHashCode(const void *key, Size keySize);
 static void ResetDistTableCacheEntry(DistTableCacheEntry *cacheEntry);
 static void CreateDistTableCache(void);
+static void InvalidateForeignRelationGraphCacheCallback(Datum argument, Oid relationId);
 static void InvalidateDistRelationCacheCallback(Datum argument, Oid relationId);
 static void InvalidateNodeRelationCacheCallback(Datum argument, Oid relationId);
 static void InvalidateLocalGroupIdRelationCacheCallback(Datum argument, Oid relationId);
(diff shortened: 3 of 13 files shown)


jasonmp85:
Oh, and there's already some code dealing with edges and adjacency lists in our deadlock detection; is that similar enough to provide something to model this on? I found that code much easier to read.

velioglu:
We are planning to support referencing reference tables from hash distributed ones, yet there is subtle problem occurring while modifying reference tables (referenced by distributed table) and distributed table within the same transaction. Following one is an example of that case. Note that modification on ref_table will be cascaded to dist_table:

BEGIN;
MODIFY ref_table  -> will get lock on both ref_table and shard of dist_table referencing ref_table
MODIFY dist_table -> need to use connections opened by the previous command
COMMIT;

To handle this case, we need to know whether shard of dist_table's can be affected by modifying ref_table. If we can understand such a case, we can warn users about changing their modification logic from parallel to sequential (otherwise there will be a deadlock). Changing the order of modifications also needs care (we need to error out in that case, since parallel connections will be opened before we can understand that we are modifying referenced ref_table)

Initial algorithm was implemented by changing the DFS for finding closure matrix faster than Warshall algorithm for sparse graphs. In order to make it more understandable I have changed it to the well-known Warshall algorithm. Indexes are used to map relationOids to the positions on the adjacency matrix. Hope it is less confusing now.

We won't support incremental updates on the graph. It will be cached by this PR. Invalidation of that will be handled by another PR, which will implemented on top of this. Graph will be used to provide referencing transitivity relation between tables.

onderkalaci:
To add @velioglu's note: 

A single level of foreign key relation information is not enough for us. Say, `dist_table_2` refers to `dist_table_1`, and `dist_table_1` refers to `reference_relation`. In that case, any modification to `reference_relation` could cascade to `dist_table_1`, which eventually cascades to `dist_table_2`. Thus, similar restriction that @velioglu explained should be applied to `dist_table_2` and `reference_relation` as well within a single transaction.

 So, this PR ensures that we get (a) all the relation which could cascade to a given relation (b) all the relations that a relation could cascade to

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -21,11 +22,413 @@
...
+}
+
+
+/*
+ * GetForeignKeyRelation returns the list of oids affected or affecting given
+ * relation id. It uses the transitivity matrix which is computed once in a
+ * session.
+ */
+List *
+GetForeignKeyRelation(Oid relationId, bool isAffecting)
```
onderkalaci: I'm writing some code on top of this PR for adding this information to metadata cache and invalidate the graph. I found this API slightly confusing for the following reasons:

1) Instead of passing a boolean flag, we should have two separate functions with clear names. 

2) `GetForeignKeyRelation()` isn't really a descriptive name. As mentioned in (1) we should probably have the following APIs `ReferencingRelationIdList()` and `ReferencedRelationIdList()`

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -21,11 +22,413 @@
...
+	}
+}
+
+
+/*
+ * CreateForeignKeyRelationGraph creates the foreign key relation graph using
+ * foreign constraint provided by pg_constraint metadata table.
+ */
+void
+CreateForeignKeyRelationGraph()
```
onderkalaci: We also need an API to destroy the graph and release all the allocated memory. Could you please add that? Thanks!

velioglu:
Cache invalidation has also been added to this PR. We may slightly change the logic of `SetRelatedOidsInvalid` (may comment out the if block) after talking about it with Onder tomorrow.

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -21,11 +22,588 @@
...
+				referredNode->adjacencyList = NIL;
+				referredNode->index = curIndex++;
+				referringNode->adjacencyList = lappend(referringNode->adjacencyList,
+													   referredNode);
+			}
+		}
+	}
+
+
+	frelGraph->nodeCount = curIndex;
```
jasonmp85: Strictly, this is unnecessary; the hash knows how many nodes there are: `hash_get_num_entries`

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -21,11 +22,588 @@
...
+		 * Initialize index to oid mapping, that is necessary for accessing node
+		 * elements in O(1) time.
+		 */
+		InitializeIndexToOidMapping();
+
+		/*
+		 * Transitivity matrix will be used to find affected and affecting relations
+		 * for foreign key relation graph.
+		 */
+		CreateTransitivityMatrix();
```
jasonmp85: I didn't like the use of transitivity in the previous PR and I'm still not sure about it here. This should probably be called a _connectivity matrix_. It _represents_ a transitive closure, but it does not represent the relation _transitivity_. Instead, it represents the _connected to_ relation, i.e. if there is a `1` in the matrix for row X, column Y, then X _is connected to_ Y through foreign keys.

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -21,11 +22,588 @@
...
+	}
+
+	frelGraph->validGraph = true;
```
jasonmp85: `validGraph` seems redundant. In a lot of our caching code we use `isValid`; I think we should stick with that.

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -21,11 +22,588 @@
...
+static void
+InitializeIndexToOidMapping()
+{
+	HASH_SEQ_STATUS status;
+	FRelNode *frelNode = NULL;
+
+	frelGraph->indexToOidArray = (Oid *) palloc(sizeof(Oid) * frelGraph->nodeCount);
+
+	hash_seq_init(&status, frelGraph->nodeMap);
+	while ((frelNode = (FRelNode *) hash_seq_search(&status)) != 0)
```
jasonmp85: Our code sometimes compares to `NULL`, sometimes to `0`. I believe `NULL` to be more correct.

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -21,11 +22,588 @@
...
+	{
+		uint32 relationIndex = frelNode->index;
+		frelGraph->indexToOidArray[relationIndex] = frelNode->relationId;
+	}
+}
+
+
+/*
+ * CreateTransitiveClosure creates the transitive closure matrix for the given
+ * graph. It uses Floyd Warshall algorithm for obtaining the transitivity graph.
```
jasonmp85: If this is _just_ to calculate a transitive closure (also, this is a _connectivity matrix_, not a _transitivity graph_), then the algorithm is just called _Warshall's Algorithm_. _Floyd-Warshall_ is for calculating shortest paths in a weighted graph (not what we're doing here).

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -21,11 +22,588 @@
...
+	while ((frelNode = (FRelNode *) hash_seq_search(&status)) != 0)
+	{
+		uint32 relationIndex = frelNode->index;
+		frelGraph->indexToOidArray[relationIndex] = frelNode->relationId;
+	}
+}
+
+
+/*
+ * CreateTransitiveClosure creates the transitive closure matrix for the given
```
jasonmp85: `s/transitive closure matrix/connectivity matrix/`

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -21,11 +22,588 @@
...
+
+	hash_seq_init(&status, frelGraph->nodeMap);
+
+	/*
+	 * Create the initial adjacency matrix, that will be transformed to the
+	 * transitivity matrix dynamically.
+	 */
+	while ((frelNode = (FRelNode *) hash_seq_search(&status)) != 0)
+	{
+		uint32 relationIndex = frelNode->index;
```
jasonmp85: Maybe `referencingIndex`?

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -21,11 +22,588 @@
...
+		ListCell *nodeCell = NULL;
+		List *curAdjacencyList = frelNode->adjacencyList;
+
+
+		foreach(nodeCell, curAdjacencyList)
+		{
+			FRelNode *referencingFrelNode = (FRelNode *) lfirst(nodeCell);
```
jasonmp85: Isn't this actually the `referencedFrelNode`?

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -21,11 +22,588 @@
...
+	 * Add all vertices one by one to the set of intermediate vertices.
+	 *
+	 * Before start of a iteration, we have transitivity values for all
+	 * pairs of vertices such that the transitivity values consider only
+	 * the vertices in set {0, 1, 2, .. k-1} as intermediate vertices.
+	 *
+	 * After the end of a iteration, vertex no. k is added to the set of
+	 * intermediate vertices and the set becomes {0, 1, .. k}
+	 */
+	for (interVertexCounter = 0; interVertexCounter < frelGraph->nodeCount;
```
jasonmp85: We should stick with `vertex` or `node`; it's confusing to use both.

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -21,11 +22,588 @@
...
+		 interVertexCounter++)
+	{
+		/* pick all vertices as source one by one */
+		for (sourceCounter = 0; sourceCounter < frelGraph->nodeCount; sourceCounter++)
```
jasonmp85: Also, `counter` is very weird given these are indexes?

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -21,11 +22,588 @@
...
+		for (indexCounter = 0; indexCounter < frelGraph->nodeCount; indexCounter++)
+		{
+			pfree(frelGraph->transitivityMatrix[indexCounter]);
+		}
+
+		pfree(frelGraph->indexToOidArray);
+	}
+
+	/* free the map holding relation nodes */
+	hash_destroy(frelGraph->nodeMap);
```
jasonmp85: I'm pretty sure this will clear the entire `CacheMemoryContext`, since `hash_destroy` clears the hash's memory context and you've set it to `CurrentMemoryContext` rather than a child.

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -21,11 +22,588 @@
...
+
+	return foreignKeyList;
+}
+
+
+/*
+ * CreateForeignKeyRelationGraph creates the foreign key relation graph using
+ * foreign constraint provided by pg_constraint metadata table.
+ */
+static void
```
jasonmp85: _Everything_ inside this function will be allocated in `CacheMemoryContext`, including anything called during traversal of the `pg_constraint` catalog.

Rather than switching into `CacheMemoryContext`, it might be best to create some sort of `CacheMemoryContext` child specific to this code once and then use it everywhere you need long-lived allocations.

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -2912,6 +2934,51 @@ InvalidateDistRelationCacheCallback(Datum argument, Oid relationId)
...
+			DistTableCacheEntry *neighbourCacheEntry = hash_search(
+				DistTableCacheHash, neighbourHashKey, HASH_FIND, &foundInCache);
+
+			if (foundInCache)
+			{
+				neighbourCacheEntry->isValid = false;
+			}
+		}
+
+		/*if (list_length(connectedComponentListOfRelation) > 0) */
```
jasonmp85: Remove this…

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -2912,6 +2934,51 @@ InvalidateDistRelationCacheCallback(Datum argument, Oid relationId)
...
+	 */
+	if (IsForeignKeyGraphValid())
+	{
+		connectedComponentListOfRelation = ConnectedComponentOfRelationId(
+			relationId);
+
+		foreach(nodeCell, connectedComponentListOfRelation)
+		{
+			Oid neighbourRelationId = lfirst_oid(nodeCell);
+			void *neighbourHashKey = (void *) &neighbourRelationId;
```
jasonmp85: This can probably be done inline? I don't see the point in declaring this.

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -2912,6 +2934,51 @@ InvalidateDistRelationCacheCallback(Datum argument, Oid relationId)
...
+
+			if (foundInCache)
+			{
+				neighbourCacheEntry->isValid = false;
+			}
+		}
+
+		/*if (list_length(connectedComponentListOfRelation) > 0) */
+		/*{ */
+		SetForeignKeyGraphInvalid();
```
jasonmp85: I think it'd be more clear to make this as a separate call after `SetRelatedOidsInvalid`.

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -2912,6 +2934,51 @@ InvalidateDistRelationCacheCallback(Datum argument, Oid relationId)
...
+
+	/*
+	 * We need to invalidate cache entry of each relation which are in the
+	 * same connected component with the given relation. If we do not have
+	 * foreign key relation graph, we don't need to do anything.
+	 *
+	 * We also need to invalidate foreign key relation graph if any of the
+	 * node in it is affected by invalidating given relation ID.
+	 */
+	if (IsForeignKeyGraphValid())
```
jasonmp85: Isn't it possible to get two invalidation messages, and since you invalidate the whole graph on the first message (and don't rebuild it), you'll end up not processing the second invalidation?

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -21,11 +22,588 @@
...
+		{
+			continue;
+		}
+
+		/*
+		 * Check whether we have seen  referring or referred relations in
+		 * constraint table before. If we haven't, add them to the hash map
+		 * of graph.
+		 */
+		referringRelationOid = con->conrelid;
```
jasonmp85: OK, so here's what the simplification could look like. The gist is that we're iterating over an _ordered_ list of `<referencingOid, referencedOid>`, so if we see the same pair in a row, we skip it. This removes any need to check whether we've already added an edge.

In addition, we don't need a node counter or current index variable because we can use the number of entries in the hash for that.

Finally, since we only process each edge once, and in order, we can blindly add them to the adjacency list for a referencing node. The code's just a sketch but I think it's easier than the four-way `if` statement you have in the original.

```c
static void
PopulateAdjacencyLists()
{
	ScanKeyData scanKey[3];
	Oid prevReferencingOid = InvalidOid;
	Oid prevReferencedOid = InvalidOid;

	/*
	 * FIXME: I haven't tested this to figure out how to make SK_ORDER_BY work,
	 * but it should, somehow
	 */
	/* first we'll order by relid */
	ScanKeyInit(&scanKey[0], Anum_pg_constraint_conrelid,
				BTEqualStrategyNumber, F_OIDEQ, NULL);
	scanKey[0].sk_flags |= SK_ORDER_BY;

	/* next by frelid */
	ScanKeyInit(&scanKey[1], Anum_pg_constraint_confrelid,
				BTEqualStrategyNumber, F_OIDEQ, NULL);
	scanKey[1].sk_flags |= SK_ORDER_BY;

	/* finally we only want foreign keys */
	ScanKeyInit(&scanKey[2], Anum_pg_constraint_contype,
				BTEqualStrategyNumber, F_CHAREQ,
				CharGetDatum(CONSTRAINT_FOREIGN));

	fkeyRel = heap_open(ConstraintRelationId, AccessShareLock);
	fkeyScan = systable_beginscan(fkeyRel, ConstraintRelidIndexId, true,
								  NULL, 3, scanKey);

	while (HeapTupleIsValid(tuple = systable_getnext(fkeyScan)))
	{
		Form_pg_constraint con->conrelid = (Form_pg_constraint) GETSTRUCT(tuple);

		/* we just saw this edge, no need to add it twice */
		if (con->conrelid == prevReferencingOid &&
			con->confrelid == prevReferencedOid)
		{
			continue;
		}

		AddEdge(frelGraph->nodeMap, con->conrelid, con->confrelid);

		prevReferencingOid = con->conrelid;
		prevReferencedOid = con->confrelid;
	}

	systable_endscan(fkeyScan);
	heap_close(fkeyRel, AccessShareLock);
}

static void
AddEdge(HTAB *adjacencyLists, Oid referencingOid, Oid referencedOid)
{
	FRelNode *referencing = CreateOrFindNode(adjacencyLists, referencingOid);
	FRelNode *referenced = CreateOrFindNode(adjacencyLists, referencedOid);

	referencing->adjacencyLists = lappend(referencing->adjacencyLists, referenced);
}

static FRelNode *
CreateOrFindNode(HTAB *adjacencyLists, Oid relid)
{
	bool found = false;
	Index newNodeIdx = hash_get_num_entries(adjacencyLists);
	FRelNode *node = (FRelNode *) hash_search(adjacencyLists,
	                                          relid,
											  HASH_ENTER,
											  &found);

	if (!found)
	{
		node->adjacencyList = NIL;
		node->index = newNodeIdx;
	}
}
```

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -21,11 +22,588 @@
...
+				 destinationCounter++)
+			{
+				/*
+				 * If vertex k is on a path from i to j, then make sure that
+				 * the value of transitivityMatrix[i][j] is 1.
+				 */
+				frelGraph->transitivityMatrix[sourceCounter][destinationCounter] =
+					frelGraph->transitivityMatrix[sourceCounter][destinationCounter] ||
+					(frelGraph->transitivityMatrix[sourceCounter][interVertexCounter] &&
+					 frelGraph->transitivityMatrix[interVertexCounter][destinationCounter]);
```
marcocitus: as discussed in yesterday's call, I don't like that we're doing a `O(N^3)` algorithm here to build a transitivity matrix and then at the end of it, we still need a `O(N)` algorithm in `GetReferencedRelationIdHelper` to build the list of referenced/referencing tables. 

We could just let the graph be a graph and perform a BFS in `GetReferencedRelationIdHelper`. The only thing we need is inverted edges, so basically in addition to the `adjacencyList` (`referencedTableList`?), we'd also want a `referencingTableList`.

The BFS has a worst-case complexity of `O(N)`, but in many cases it only needs to look at a small subgraph, and we can skip the `O(N^3)` step altogether.

velioglu:
@onderkalaci I have applied the comments including,

- Invalidate cache after alter table/drop constraint (on dist colocation table)
- Change adjacency list obtaining algorithm (to what Marco suggested)
- Use child of cache context for allocating data of foreign key relation graph
- Stylistic comments

I couldn't apply
- Hash (of foreign key relation graph) create logic (to use what Jason suggest)
- Scan pg_constraint table in order and change node create logic

tomorrow, I will work on them.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -2227,8 +2227,13 @@ ErrorIfUnsupportedAlterTableStmt(AlterTableStmt *alterTableStatement)
 			}
 
 #endif
-			case AT_SetNotNull:
 			case AT_DropConstraint:
+			{
+				CitusInvalidateRelcacheByRelid(DistColocationRelationId());
```
onderkalaci: Only when dropping foreign key

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -2558,6 +2573,8 @@ InitializeDistTableCache(void)
 	/* Watch for invalidation events. */
 	CacheRegisterRelcacheCallback(InvalidateDistRelationCacheCallback,
 								  (Datum) 0);
+	CacheRegisterRelcacheCallback(HoaydaRelationCacheCallback,
```
onderkalaci: rename `HoaydaRelationCacheCallback ` 😄

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -11,21 +11,492 @@
...
+		MemoryContext frelMemoryContext = AllocSetContextCreateExtended(
+			CacheMemoryContext,
+			"FRel Graph Context",
+			ALLOCSET_DEFAULT_MINSIZE,
+			ALLOCSET_DEFAULT_INITSIZE,
+			ALLOCSET_DEFAULT_MAXSIZE);
+
+		ClearForeignKeyRelationGraphContext();
+
+		oldContext = MemoryContextSwitchTo(frelMemoryContext);
```
onderkalaci: Lets not use acronym for `frelMemoryContext `

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -11,21 +11,492 @@
...
+ */
+void
+ClearForeignKeyRelationGraphContext()
+{
+	if (frelGraph == NULL)
+	{
+		return;
+	}
+
+	hash_destroy(frelGraph->nodeMap);
```
onderkalaci: Since we allocate `frelGraph ` under the same memory context, let's set `frelGraph = NULL` after `hash_destroy` given that all the graph will be already freed.

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -2913,6 +2940,22 @@ ResetDistTableCacheEntry(DistTableCacheEntry *cacheEntry)
 }
 
 
+/*
+ * InvalidateDistRelationCacheCallback flushes cache entries when a relation
```
onderkalaci: Comments of the function should be updated.

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -2958,6 +2994,21 @@ InvalidateDistRelationCacheCallback(Datum argument, Oid relationId)
 }
 
 
+static void
+InvalidateEntireDistCache()
```
onderkalaci: We've a very similar function called `FlushDistTableCache()`. I can see that `InvalidateEntireDistCache ()` would be more convenient for our purpose. But, we should document our purpose with some comments

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -11,21 +11,540 @@
...
+ *
+ */
+static void
+GetConnectedListHelper(FRelNode *node, List **adjacentNodeList, bool isReferencing)
+{
+	ListCell *nodeCell = NULL;
+	List *adjacencyList = NIL;
+
+	node->visited = true;
+	*adjacentNodeList = lappend(*adjacentNodeList, node);
```
onderkalaci: @velioglu  Why do we add the node to itself adjacency list?

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -11,21 +11,540 @@
...
+{
+	const FRelEdge *leftEdge = *((const FRelEdge **) leftElement);
+	const FRelEdge *rightEdge = *((const FRelEdge **) rightElement);
+
+	Oid leftReferencingOID = leftEdge->referencingRelationOID;
+	Oid leftReferencedOID = leftEdge->referencedRelationOID;
+	Oid rightReferencingOID = rightEdge->referencingRelationOID;
+	Oid rightReferencedOID = rightEdge->referencedRelationOID;
+
+	if (leftReferencingOID < rightReferencingOID)
```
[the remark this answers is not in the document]
velioglu: First sort in referencing oid and then in referenced oid (I have committed a kind of hacky version)

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -11,21 +11,526 @@
...
+#include "utils/catcache.h"
 #include "utils/fmgroids.h"
+#include "utils/inval.h"
+#include "utils/memutils.h"
 #include "utils/rel.h"
 #include "utils/relcache.h"
 #include "utils/ruleutils.h"
 #include "utils/syscache.h"
+#include "utils/lsyscache.h"
+#include "libpq/libpq-be.h"
```
onderkalaci: @velioglu why do we need `"libpq/libpq-be.h"`. I'm removing that

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -11,21 +11,526 @@
...
+
+/* Helper function to compare two FRelEdges using referencing and referenced ids
+ * respectively */
+static int
+CompareFRelEdges(const void *leftElement, const void *rightElement)
+{
+	const FRelEdge *leftEdge = *((const FRelEdge **) leftElement);
+	const FRelEdge *rightEdge = *((const FRelEdge **) rightElement);
+
+	int referencingDiff = leftEdge->referencingRelationOID -
```
onderkalaci: I'm almost sure that `int = unsigned - unsigned` works fine even if the result is negative.

But, could you please double check that?

(oid is unsigned int, the result of the operation is signed, so make sure that we don't have any subtle bugs because of it)

onderkalaci:
@jasonmp85 

>  Finally, what happens if the invalidation callback is called twice in a row with different relids? I'm not convinced it'll do the right thing. Can that happen?

That cannot happen anymore, because we've re-wrote the cache invalidation logic for the graph. Given that invalidation of the graph is costly, we've restricted the invalidation only for operations: `CREATE/DROP TABLE ` that involve foreign keys, `ALTER TABLE ADD/DROP FOREIGN CONSTRAINT` by a specifc callback (e.g., do not rely on distributed tables' relcache invalidations).

And, @jasonmp85 it seems you're OOTO this week, so we'll have to go ahead and merge the PR. @velioglu applied almost all of the feedback you've provided, which simplified the PR a lot. Thanks!

@velioglu I think @jasonmp85's below comment is also a reasonable idea. Can you do that? Also, let's move `PG_FUNCTION_INFO_V1(get_referencing_relation_id_list);` and `
PG_FUNCTION_INFO_V1(get_referenced_relation_id_list);` under `src/test` folder, which seems like a better place.

>  This should live in a separate file. It's self-contained so it would be nice to see it appropriately separated

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -11,28 +11,530 @@
...
+	if (IsForeignKeyGraphValid())
+	{
+		return;
+	}
+
+	ClearForeignKeyRelationGraphContext();
+
+	foreignRelationMemoryContext = AllocSetContextCreateExtended(
+		CacheMemoryContext,
+		"FRel Graph Context",
```
onderkalaci: We generally do not prefer acronyms. Can you go over the code and rename acronyms, especially`FRel` / `frelGraph ` / `FRelNode ` would too hard to understand for a new reader. Also, using `Relation` could be confusing for the readers since `Relation` means a lot in Postgres 😄  So, maybe use `Relationship`? 

maybe use `ForeignConstraintRelationshipGraph` and `ForeignConstraintRelationshipNode`?

src/backend/distributed/ddl/foreign_constraint.c
onderkalaci: Let's do not forget to double check this.

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -11,28 +11,530 @@
...
+	{
+		return referencingDiff;
+	}
+
+	return referencedDiff;
+}
+
+
+/*
+ * AddEdge adds edge between the nodes having given OIDs.
```
onderkalaci: I think we should expand the comment of this function a bit. We should maybe briefly mention about adding the edges in both ways.

Also, I found passing the `HTAB` here a bit confusing, because we're usually directly accessing the `static` variable directly. But, still acceptable.

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -649,3 +1151,98 @@ HeapTupleOfForeignConstraintIncludesColumn(HeapTuple heapTuple, Oid relationId,
...
+
+/*
+ * TableReferencing function checks whether given table is referencing by another table
+ * via foreign constraints. If it is referencing, this function returns true. To check
+ * that, this function searches given relation at pg_constraints system catalog. However
+ * since there is no index for the column we searched, this function performs sequential
+ * search, therefore call this function with caution.
+ */
+bool
+TableReferencing(Oid relationId)
```
onderkalaci: Let's keep this utility function in this file

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -649,3 +1151,98 @@ HeapTupleOfForeignConstraintIncludesColumn(HeapTuple heapTuple, Oid relationId,
...
+	return false;
+}
+
+
+/*
+ * ConstraintIsAForeignKey returns true if the given constraint name
+ * is a foreign key to defined on the relation.
+ */
+bool
+ConstraintIsAForeignKey(char *constraintNameInput, Oid relationId)
```
onderkalaci: Let's keep this utility function in this file as well

src/backend/distributed/executor/multi_utility.c
```diff
@@ -3774,10 +3842,20 @@ ProcessDropTableStmt(DropStmt *dropTableStatement)
...
+		/* we're not interested in non-valid, non-distributed relations */
+		if (relationId == InvalidOid || !IsDistributedTable(relationId))
+		{
+			continue;
+		}
+
+		/* invalidate foreign key cache if the table involved in any foreign key */
+		if ((TableReferenced(relationId) || TableReferencing(relationId)))
+		{
+			shouldInvalidateForeignKeyGraph = true;
```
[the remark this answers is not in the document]
onderkalaci: I cannot actually see any edge cases with that. Either dropping the referenced or referencing table would invalidate the foreign key graph (depending on the order of the DROP).

What kind of edge do you foresee with that? @velioglu will add a test with the following specs:

```
-- create fkeys in the following form
schema_1.table_1 -> schema_2.table_1 ->schema_1.table_2

-- test1
drop schema schema_2;
check graph

-- test2
drop schema schema_1;
check graph
 
```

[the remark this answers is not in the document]
onderkalaci: Ok, we hit an issue, still looking into that :)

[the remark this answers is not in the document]
marcocitus: An alternative might be to invalidate from `master_drop_distributed_table_metadata` . A drawback is that you cannot know whether there were any foreign keys once you're in the drop handler, so you'd probably always invalidate when a distributed table or reference table is dropped.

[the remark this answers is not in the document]
velioglu: I have added support for invalidating graph when a drop cascades to a table by processing the DROP SCHEMA command in multi_ProcessUtility.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -3747,6 +3822,77 @@ RoleSpecString(RoleSpec *spec)
...
+	ListCell *dropTableCell;
+
+	if (dropStatement->behavior != DROP_CASCADE)
+	{
+		return;
+	}
+
+	foreach(dropTableCell, dropStatement->objects)
+	{
+		Value *schemaValue = (Value *) lfirst(dropTableCell);
```
onderkalaci: We haven't checked whether the drop is on a schema yet

onderkalaci: Sorry, `get_namespace_oid `

src/backend/distributed/executor/multi_utility.c
```diff
@@ -3747,6 +3822,77 @@ RoleSpecString(RoleSpec *spec)
...
+		char *schemaString = schemaValue->val.str;
```
onderkalaci: Maybe use `name = strVal((Value *) object);` as in the postgres source

src/backend/distributed/executor/multi_utility.c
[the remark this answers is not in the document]
onderkalaci: Hmm, I see what you mean. I think we missed that again :/ 

But is it ever possible for any command (except for `DROP SCHEMA` and `DROP DATABASE`) to drop a table? I cannot find any.

I think, say a `DROP TYPE`, can drop a column which is involved in the foreign key, which breaks our logic a bit. Or, `DROP INDEX` on the pkey could `CASCADE` to the foreign key.
Those sound like the only place we're missing the invalidation.

I'm actually not too worried about `DROP` commands because any lack of invalidation could only lead to enforcing some unnecessary restrictions until the graph is invalidated.

And, having support for `ALTER TABLE ADD/DROP fkey`, `CREATE/DROP TABLE with fkey` and `DROP SCHEMA` sounds like covering majority of the cases where invalidation is necessary.

We could maybe quickly add `DROP INDEX CASCADE` support as well cc: @velioglu.

@marcocitus Do you prefer/expect to have a solution with a fundamental change in the logic?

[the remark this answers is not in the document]
onderkalaci: In the mean time, I'll try to write a function that recursively goes trought `pg_depend` on any drop command, and if it finds any foreign key constraint being dropped, invalidates the foreign key.

Actually, almost copy & paste `findDependentObjects()` function in Postgres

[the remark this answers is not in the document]
onderkalaci: Or, given that it is not a simple function to write, I'll probably prefer to open an issue on that for now and leave it for future :/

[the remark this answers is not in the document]
marcocitus: Yeh, I think we can punt on it a bit, since these are unlikely edge cases. 

citus_drop_trigger also gets called when dropping a constraint, which is something we may or may not be able to use.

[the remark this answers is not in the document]
onderkalaci: Opened an issue #2250

PR #2212, opened 2018-06-12
DDL failure testing

onderkalaci:
This commit adds an extensive failure testing, which covers quite
a bit of thing and their combinations:
   - 1PC vs 2PC
   - Replication factor 1 and Replication factor 2
   - Network failures and query cancellations
   - Parallel vs Sequential mode 

I don't think we need to add that many tests for all types of queries. This PR seems to cover a lot of cases in the connection and transaction management.

Missing item:
- [x] Add tests for sequential mode once @lithp rebases `mitmproxy-failure-testing` branch to the master

### DIFF
--- a/src/test/regress/expected/failure_ddl.out
+++ b/src/test/regress/expected/failure_ddl.out
@@ -1,0 +1,1245 @@
+-- 
+-- Test DDL command propagation failures 
+-- Different dimensions we're testing:
+--    Replication factor, 1PC-2PC, sequential-parallel modes
+-- 
+CREATE SCHEMA ddl_failure;
+SET search_path TO 'ddl_failure';
+-- we don't want to see the prepared transaction numbers in the warnings
+SET client_min_messages TO ERROR;
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SET citus.next_shard_id TO 100800;
+-- we'll start with replication factor 1, 1PC and parallel mode
+SET citus.multi_shard_commit_protocol TO '1pc';
+SET citus.shard_count = 4;
+SET citus.shard_replication_factor = 1;
+CREATE TABLE test_table (key int, value int);
+SELECT create_distributed_table('test_table', 'key');
+ create_distributed_table 
+--------------------------
+ 
+(1 row)
+
+-- in the first test, kill just in the first 
+-- response we get from the worker
+SELECT citus.mitmproxy('conn.onAuthenticationOk().kill()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+ALTER TABLE test_table ADD COLUMN new_column INT;
+ERROR:  connection error: localhost:57640
+DETAIL:  server closed the connection unexpectedly
+	This probably means the server terminated abnormally
+	before or while processing the request.
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT array_agg(name::text ORDER BY name::text) FROM public.table_attrs where relid = 'test_table'::regclass;
+  array_agg  
+-------------
+ {key,value}
+(1 row)
+
+-- cancel just in the first 
+-- response we get from the worker
+SELECT citus.mitmproxy('conn.onAuthenticationOk().cancel(' ||  pg_backend_pid() || ')');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+ALTER TABLE test_table ADD COLUMN new_column INT;
+ERROR:  canceling statement due to user request
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT array_agg(name::text ORDER BY name::text) FROM public.table_attrs where relid = 'test_table'::regclass;
+  array_agg  
+-------------
+ {key,value}
+(1 row)
+
+-- kill as soon as the coordinator sends begin
+SELECT citus.mitmproxy('conn.onQuery(query="^BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED").kill()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+ALTER TABLE test_table ADD COLUMN new_column INT;
+ERROR:  failure on connection marked as essential: localhost:57640
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT array_agg(name::text ORDER BY name::text) FROM public.table_attrs where relid = 'test_table'::regclass;
+  array_agg  
+-------------
+ {key,value}
+(1 row)
+
+-- cancel as soon as the coordinator sends begin
+SELECT citus.mitmproxy('conn.onQuery(query="^BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED").cancel(' ||  pg_backend_pid() || ')');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+ALTER TABLE test_table ADD COLUMN new_column INT;
+ERROR:  canceling statement due to user request
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT array_agg(name::text ORDER BY name::text) FROM public.table_attrs where relid = 'test_table'::regclass;
+  array_agg  
+-------------
+ {key,value}
+(1 row)
+
+-- kill as soon as the coordinator sends worker_apply_shard_ddl_command
+SELECT citus.mitmproxy('conn.onQuery(query="worker_apply_shard_ddl_command").kill()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+ALTER TABLE test_table ADD COLUMN new_column INT;
+ERROR:  server closed the connection unexpectedly
+	This probably means the server terminated abnormally
+	before or while processing the request.
+CONTEXT:  while executing command on localhost:57640
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+-- show that we've never commited the changes
+SELECT array_agg(name::text ORDER BY name::text) FROM public.table_attrs where relid = 'test_table'::regclass;
+  array_agg  
+-------------
+ {key,value}
+(1 row)
+
+-- cancel as soon as the coordinator sends worker_apply_shard_ddl_command
+SELECT citus.mitmproxy('conn.onQuery(query="worker_apply_shard_ddl_command").cancel(' ||  pg_backend_pid() || ')');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+ALTER TABLE test_table ADD COLUMN new_column INT;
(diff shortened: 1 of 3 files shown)


src/test/regress/expected/failure_ddl.out
```diff
@@ -0,0 +1,1228 @@
...
+(1 row)
+
+-- show that we've never commited the changes
+SELECT count(*) FROM public.table_attrs where relid = 'test_table'::regclass;
+ count 
+-------
+     2
+(1 row)
+
+-- cance; as soon as the coordinator sends worker_apply_shard_ddl_command
```
furkansahin: `cance;` -> `cancel`

src/test/regress/expected/failure_ddl.out
```diff
@@ -0,0 +1,1228 @@
...
+  run_command_on_placements   
+------------------------------
+ (localhost,57637,100800,t,3)
+ (localhost,57637,100802,t,3)
+ (localhost,57640,100801,t,3)
+ (localhost,57640,100803,t,3)
+(4 rows)
+
+-- killing on command complete of COMMIT PREPARE, we should see that the command succeeds
+-- and all the workers committed
```
lithp: This comment is a little confusing, I wish we could show the `WARNING`s here. From the coordinator's perspective the query "succeeds" in the exact same way that it does below. We give a warning, and can't be sure whether the changes have actually been applied until you run `recover_prepared_transactions()`.

src/test/regress/expected/failure_ddl.out
```diff
@@ -0,0 +1,1228 @@
...
+------------------------------
+ (localhost,57637,100800,t,3)
+ (localhost,57637,100802,t,3)
+ (localhost,57640,100801,t,3)
+ (localhost,57640,100803,t,3)
+(4 rows)
+
+-- but now kill just after the worker sends response to 
+-- ROLLBACK command, so we'll have lots of warnings but the command
+-- should have been rollbacked both on the distributed table and the placements
```
lithp: You probably don't need both of these `ROLLBACK` tests. If we're going to suppress WARNINGs then their behavior is identical. In one of them we explicitly send a ROLLBACK, in the other we just drop the connection and rely on workers to ROLLBACK themselves when the backend is killed.

src/test/regress/expected/failure_ddl.out
```diff
@@ -0,0 +1,1228 @@
...
+SELECT run_command_on_placements('test_table', $$SELECT count(*) FROM public.table_attrs where relid = '%s'::regclass$$) ORDER BY 1;
+  run_command_on_placements   
+------------------------------
+ (localhost,57637,100800,t,3)
+ (localhost,57637,100802,t,3)
+ (localhost,57640,100801,t,3)
+ (localhost,57640,100803,t,3)
+(4 rows)
+
+-- final set of tests with 2PC and replication factor = 2
```
lithp: it's not final now that you've added sequential tests :)

src/test/regress/expected/failure_ddl.out
```diff
@@ -0,0 +1,1228 @@
...
+
+-- show that we've never commited the changes
+SELECT count(*) FROM public.table_attrs where relid = 'test_table'::regclass;
+ count 
+-------
+     2
+(1 row)
+
+-- kill as soon as the coordinator sends COMMIT
+SELECT citus.mitmproxy('conn.onQuery(query="COMMIT").kill()');
```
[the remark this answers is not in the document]
lithp: If you use `^COMMIT`, then you'll get some shards with the new metadata and some with the old metadata.

src/test/regress/expected/failure_ddl.out
```diff
@@ -0,0 +1,1228 @@
...
+SELECT run_command_on_placements('test_table', $$SELECT count(*) FROM public.table_attrs where relid = '%s'::regclass$$) ORDER BY 1;
+  run_command_on_placements   
+------------------------------
+ (localhost,57637,100800,t,3)
+ (localhost,57637,100802,t,3)
+ (localhost,57640,100801,t,2)
+ (localhost,57640,100803,t,2)
+(4 rows)
+
+-- we should be able to revocer the transaction and
```
lithp: "revocer" -> "recover"

lithp: While following along by doing it by hand I accidentally noticed that recover_prepared_transactions() probably has a bug. If I leave in the part which kills connections on `^COMMIT`:

```
brian=# SELECT recover_prepared_transactions();
WARNING:  server closed the connection unexpectedly
WARNING:  connection not open
 recover_prepared_transactions 
                             0
(1 row)
```

It returns the number of transactions which it was able to recover, but doesn't give any indication that there are still transactions left to recover!

src/test/regress/expected/failure_ddl.out
```diff
@@ -0,0 +1,1228 @@
...
+ (localhost,57637,100805,t,2)
+ (localhost,57637,100806,t,2)
+ (localhost,57637,100807,t,2)
+ (localhost,57640,100804,t,2)
+ (localhost,57640,100805,t,2)
+ (localhost,57640,100806,t,2)
+ (localhost,57640,100807,t,2)
+(8 rows)
+
+-- we should be able to revocer the transaction and
```
lithp: "revocer" -> "recover"

src/test/regress/expected/failure_ddl.out
```diff
@@ -0,0 +1,1228 @@
...
+-- see that the command is rollbacked
```
lithp: see that what command is rollbacked? Above all shards have two columns, and below they also have two columns.

src/test/regress/expected/failure_ddl.out
```diff
@@ -0,0 +1,1228 @@
...
+ (localhost,57640,100804,t,2)
+ (localhost,57640,100805,t,2)
+ (localhost,57640,100806,t,2)
+ (localhost,57640,100807,t,2)
+(8 rows)
+
+-- finally, test failing on ROLLBACK with 2CPC
+-- fail just after the coordinator sends the ROLLBACK
+-- so the command can be rollbacked
+SELECT citus.mitmproxy('conn.onQuery(query="ROLLBACK").kill()');
```
lithp: I still think this is not a particularly useful test. We're suppressing WARNING so we're not really resting any of the coordinator's behavior. And by killing connections all we're exercising is the Postgres behavior: automatically rolling back when the backend ends.

src/test/regress/expected/failure_ddl.out
```diff
@@ -0,0 +1,1228 @@
...
+ALTER TABLE test_table ADD COLUMN new_column INT;
+ERROR:  canceling statement due to user request
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SET search_path TO 'public';
+DROP SCHEMA ddl_failure CASCADE;
```
lithp: You should also drop `test_table` :)

src/test/regress/expected/failure_ddl.out
onderkalaci: Yeah, I remember this. But, aren't the `WARNING`s  important indication for the necessity to re-run the recovery?

I remember during the PR for `recover_prepared_transactions ()` we decided to warn the user about the connection errors, but, do not fail the recovery at all since we could potentially recover prepared transactions from the other workers.

Do you see any other approaches for that?

src/test/regress/expected/failure_ddl.out
```diff
@@ -0,0 +1,1228 @@
...
+-----------
+ 
+(1 row)
+
+ALTER TABLE test_table ADD COLUMN new_column INT;
+ERROR:  connection error: localhost:57640
+DETAIL:  server closed the connection unexpectedly
+	This probably means the server terminated abnormally
+	before or while processing the request.
+SELECT citus.mitmproxy('conn.allow()');
```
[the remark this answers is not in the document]
lithp: From `failure_ddl.sql`:

````sql
-- in the first test, kill just in the first 
-- response we get from the worker
SELECT citus.mitmproxy('conn.onAuthenticationOk().kill()');
ALTER TABLE test_table ADD COLUMN new_column INT;
SELECT citus.mitmproxy('conn.allow()'); -- the conn.allow() in question
SELECT count(*) FROM public.table_attrs where relid = 'test_table'::regclass; -- this happens locally

-- cancel just in the first 
-- response we get from the worker
SELECT citus.mitmproxy('conn.onAuthenticationOk().cancel(' ||  pg_backend_pid() || ')'); -- reconfigure
ALTER TABLE test_table ADD COLUMN new_column INT; -- runs under the new rules
SELECT citus.mitmproxy('conn.allow()');
SELECT count(*) FROM public.table_attrs where relid = 'test_table'::regclass;
````

There is another `ALTER TABLE`, but that doesn't happen until we run `citus.mitmproxy` again, the `conn.allow()` never has a change to be used!

This isn't very important, if you think it makes things more clear feel free to leave it in, just wanted to point this out.

src/test/regress/expected/failure_ddl.out
lithp: Opened an issue: https://github.com/citusdata/citus/issues/2282

src/test/regress/expected/failure_ddl.out
lithp: Okay, makes sense

src/test/regress/expected/failure_ddl.out
lithp: I understand the idea but I don't agree that this implementation matches it. Why not error out here? Recovery itself has errored! Warnings are something you can ignore, here it feels like we're suppressing an important failure. Until `recover_prepared_transactions()` runs all your queries on this table are going to block and you won't know why.

This isn't part of this PR though, I guess I should open a ticket.

src/test/regress/expected/failure_ddl.out
lithp: Okay, good points.

src/test/regress/expected/failure_ddl.out
lithp: And another good point, this is a really nice idea, we can create whatever we want and not keep track of it.

mtuncer:
Some views are not cleaned up

\d
             List of relations
 Schema |      Name       | Type |  Owner   
--------+-----------------+------+----------
 public | table_attrs     | view | postgres
 public | table_checks    | view | postgres
 public | table_desc      | view | postgres
 public | table_fkey_cols | view | postgres
 public | table_fkeys     | view | postgres
(5 rows)

onderkalaci:
>Some views are not cleaned up

Those are added via `multi_test_helpers.sql` at the beginning of the regression tests.

Another note from above:

> Just FYI, this is Citus' test helper function see [here](https://github.com/citusdata/citus/blob/master/src/test/regress/sql/multi_test_helpers.sql). I think the main motivation to add that was Postgres' ouput changes between different versions and with this views we could skip adding _0. files for such minor differences.

src/test/regress/expected/failure_ddl.out
onderkalaci: >  If we want to continue and try to recover prepared transactions from the other workers, we can still try to do that before we raise the ERROR.

Yes, that would be a nice improvement.

onderkalaci: Just note that we've got the exact same approach with deadlock detection where we warn the user if we cannot connect one of the nodes. We should probably improve both!

PR #2215, opened 2018-06-12
Always throw errors on failure on critical connection in router executor

### LINKED ISSUE #2214: Reference table mid-transaction failure may cause shardstate 3 and PANIC
It seems we don't check whether a connection is marked as critical in the SELECT logic and could therefore ignore an error that follows a reference table write and ultimately try to mark the placement as inactive due to the failure. In that case, Citus throws an error in the commit handler, which causes a panic, but only after the shard placement state has already been updated to 3.

```
BEGIN;
UPDATE ref SET b = 3;                                                                                                                                                   
-- fail a worker node 
SELECT * FROM ref;                                                                                                                                                      
WARNING:  terminating connection due to administrator command
 a │ b 
───┼───
(0 rows)

END;
WARNING:  AbortTransaction while in COMMIT state
ERROR:  failure on connection marked as essential: localhost:9701
PANIC:  cannot abort transaction 3377, it was already committed
```

marcocitus:
Fixes #2214 

Citus currently doesn't throw an error when a failure occurs in the router executor on a connection that was marked as critical. This can cause transactions to incorrectly proceed after a failure on a connection that's supposed to complete a 2PC due to commands earlier in the transaction. The commit handler is not ready for that situation and panics.

This PR fixes it by checking whether the connection is critical when an error response is returned, or throwing an error immediately if a connection failure occurs.

As a side-effect, this also caused some error messages in queries with CTEs to become more meaningful.

I've also defensively moved the call to `CheckTransactionHealth` out of `CoordinatedRemoteTransactionsCommit` because that created a code path where we could throw an error in the commit handler.

### DIFF
--- a/src/backend/distributed/executor/multi_router_executor.c
+++ b/src/backend/distributed/executor/multi_router_executor.c
@@ -1418,18 +1418,18 @@ SendQueryInSingleRowMode(MultiConnection *connection, char *query,
 
 	if (querySent == 0)
 	{
-		const bool raiseErrors = false;
+		const bool raiseIfTransactionIsCritical = true;
 
-		HandleRemoteTransactionConnectionError(connection, raiseErrors);
+		HandleRemoteTransactionConnectionError(connection, raiseIfTransactionIsCritical);
 		return false;
 	}
 
 	singleRowMode = PQsetSingleRowMode(connection->pgConn);
 	if (singleRowMode == 0)
 	{
-		const bool raiseErrors = false;
+		const bool raiseIfTransactionIsCritical = true;
 
-		HandleRemoteTransactionConnectionError(connection, raiseErrors);
+		HandleRemoteTransactionConnectionError(connection, raiseIfTransactionIsCritical);
 		return false;
 	}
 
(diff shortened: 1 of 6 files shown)


onderkalaci:
Just FYI: This seems like a good candidate to add a test using the new failure testing framework. We have not checked in the failure testing framework, but, after we merge this PR, it should be easy to add a test via a new PR as #2212

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -1581,7 +1585,8 @@ StoreQueryResult(CitusScanState *scanState, MultiConnection *connection,
 			category = ERRCODE_TO_CATEGORY(ERRCODE_INTEGRITY_CONSTRAINT_VIOLATION);
 			isConstraintViolation = SqlStateMatchesCategory(sqlStateString, category);
 
-			if (isConstraintViolation || failOnError)
+			if (isConstraintViolation || failOnError ||
```
pykello: `failOnError` is misnamed here, since there are cases that this can be `false` but we still fail. We should at least update the comments for `StoreQueryResult()` and `ConsumeQueryResult()`.

marcocitus:
Note: I did some renaming based on Hadi's feedback, but only the first 3 commits should be backported.

PR #2218, opened 2018-06-13
refactor grantee serialization for reuse

thanodnl:
Its a common pattern to serialize `RoleSpec *` back to its SQL form. By extracting it to its own function we can reuse it.

Also there might be a difference in the value of `SESSION_USER` between the coordinator and a worker when `SET ROLE ...;` has been used on the coordinator. Therefore we resolve `SESSION_USER` and `CURRENT_USER` to its value on during execution on the coordinator to the actual user.

### DIFF
--- a/src/backend/distributed/executor/multi_utility.c
+++ b/src/backend/distributed/executor/multi_utility.c
@@ -3514,7 +3514,7 @@ PlanGrantStmt(GrantStmt *grantStmt)
 		}
 	}
 
-	/* deparse the privileges */
+	/* deparse the grantees */
 	isFirst = true;
 	foreach(granteeCell, grantStmt->grantees)
 	{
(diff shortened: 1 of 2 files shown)


src/backend/distributed/executor/multi_utility.c
```diff
@@ -3671,6 +3657,46 @@ CollectGrantTableIdList(GrantStmt *grantStmt)
 }
 
 
+/*
+ * RoleSpecString resolves the role specification to its string form that is suitable for transport to a worker node.
+ * This function resolves the following identifiers from the current context so they are safe to transfer.
+ *
+ * CURRENT_USER - resolved to the user name of thecurrent role being used
```
marcocitus: thecurrent -> the current

PR #2220, opened 2018-06-14
Track relation accesses using the connection management infrastructure

onderkalaci:
Similar to #2203.

But this time we use Citus' own infrastructure as opposed to using PostgreSQL's locks. Considering the downsides of using PostgreSQL locks, we'd probably want to go with this approach.

Missing things:

- [x] Inter shard ddl commands should cover left relationId as well as right relation id
- [x] `TRUNCATE` should be `DDL` access considering our purposes
- [x] Test `create_distributed_table()` with data loading
- [x] Test querying partitioned tables. What is the expected thing to do? Should we add its partitions to the list?
- [x] Are there any other queries that I've missed? -- I'm skipping task-tracker at all since they don't work in a transaction block
    - [x] Modifying CTEs w/wout SELECTs
- [x] Fix PG 9.6 tests.


src/backend/distributed/transaction/multi_shard_transaction.c
```diff
@@ -109,9 +109,22 @@ OpenTransactionsForAllTasks(List *taskList, int connectionFlags)
...
-			/* add additional placement accesses for subselects (e.g. INSERT .. SELECT) */
-			placementSelectList = BuildPlacementSelectList(shardPlacement->groupId,
-														   task->relationShardList);
+			if (accessType == PLACEMENT_ACCESS_DDL)
+			{
+				/*
+				 * All relations appearing inter-shard DDL commands should be marked
+				 * with DDL access.
+				 */
+				placementSelectList = BuildPlacementDDLList(shardPlacement->groupId,
```
marcocitus: the `SelectList = ... DDLList` is a little confusing, maybe `placementReadList`

src/backend/distributed/transaction/relation_access_tracking.c
```diff
@@ -0,0 +1,381 @@
...
+{
+	RecordParallelRelationAccess(relationId, PLACEMENT_ACCESS_DDL);
+}
+
+
+/*
+ * RecordParallelRelationAccess records the relation access mode as parallel
+ * for the given access type (e.g., select, dml or ddl) in the RelationAccessHash.
+ *
+ * The function becomes no-op for non-transaction blocks
```
marcocitus: what about modifying CTEs?

src/backend/distributed/transaction/relation_access_tracking.c
```diff
@@ -0,0 +1,381 @@
...
+	hashEntry->relationAccessMode |= (1 << (accessType));
+}
+
+
+/*
+ * RecordRelationMultiShardSelectAccessForTask goes over all the relations
+ * in the relationShardList and records the select access per each table.
+ */
+void
+RecordRelationMultiShardSelectAccessForTask(Task *task)
```
marcocitus: What's the difference between Parallel and MultiShard in the naming?

src/backend/distributed/transaction/relation_access_tracking.c
onderkalaci: That's a very good point. I think in general I missed modifying `CTE`s in this implementation.

At first, I thought we could use `InCoordinatedTransaction()` instead of `IsTransactionBlock ()`, however, that also doesn't solve all of our problems, because if we do so, we'd skip the first `SELECT` statements which are router, because we don't call `BeginOrContinueCoordinatedTransaction()` for router selects.

```SQL
BEGIN;
SELECT count(*) FROM reference_Table;
DELETE from dist_table;
```

Also, using `InCoordinatedTransaction()` instead of `IsTransactionBlock()` would add some overhead for non-cte multi-shard queries, where we'd unnecessarly hit the hash tables.

Overall, seems like we should use `InCoordinatedTransaction() || IsTransactionBlock()`  to preserve the correctness, acknowledging the overhead. (I'm not too worried about the overhead since its a very simple hash lookup for and `Oid->int` mapping.)

Does that sound reasonable? Or, do you have any other ideas?

marcocitus: sounds reasonable

src/backend/distributed/commands/multi_copy.c
```diff
@@ -2311,6 +2313,21 @@ CitusCopyDestReceiverReceive(TupleTableSlot *slot, DestReceiver *dest)
 											   &shardConnectionsFound);
 	if (!shardConnectionsFound)
 	{
+		/*
+		 * Keep track of multi shard accesses before opening connection
+		 * the second shard.
```
marcocitus: maybe add: on the same node

marcocitus: no wait, it's global, nvm

src/backend/distributed/transaction/relation_access_tracking.c
marcocitus: Actually, the `SELECT count(*) FROM reference_Table;` shouldn't really affect anything since it's outside of the transaction so it doesn't hold locks.

onderkalaci: That's true, but, it'd probably a little bit confusing. Because we're explicitly naming things with `RecordRelationAccess*`, and as a reader, the default expectation would probably be to see all the relation accesses. 

(Btw I've done some benchmarks locally, and I haven't observed any performance overhead. I think it's negligible. I'll do some proper benchmarks once we've all the pieces for fkey to reference tables)

(I cannot easily see any future use-cases for this relation accesses infrastructure, but, if we need to use it in the future, this could be confusing for sure).

That said, I have a slight preference to keep this as-is for now. Does that work for you?

marcocitus:
woah, appveyor passed, this must be good :)

src/backend/distributed/transaction/relation_access_tracking.c
```diff
@@ -0,0 +1,529 @@
...
+{
+	hash_delete_all(RelationAccessHash);
+}
+
+
+/*
+ * Allocate RelationAccessHash.
+ */
+void
+AllocateRelationAccessHash()
```
marcocitus: Iirc, some compilers complain without void in between the parentheses

src/backend/distributed/transaction/relation_access_tracking.c
```diff
@@ -0,0 +1,529 @@
...
+{
+	HASHCTL info;
+	uint32 hashFlags = 0;
+
+	memset(&info, 0, sizeof(info));
+	info.keysize = sizeof(RelationAccessHashKey);
+	info.entrysize = sizeof(RelationAccessHashEntry);
+	info.hash = tag_hash;
+	info.hcxt = ConnectionContext;
```
marcocitus: Optional: It might be slightly safer to recreate the hash table for every transaction in a transaction context. It's not necessarily slower because resetting a memory context is faster than deleting individual elements.

src/backend/distributed/transaction/relation_access_tracking.c
```diff
@@ -0,0 +1,529 @@
...
+				ColocatedShardIdInRelation(partitionOid, shardIndex);
+			partitionShardPlacementList = ShardPlacementList(partitionShardId);
+
+			foreach(partitionCell, partitionShardPlacementList)
+			{
+				ShardPlacement *shardPlacement =
+					(ShardPlacement *) lfirst(partitionCell);
+
+				/* recursively record all relation accesses of its partitions */
+				AssociatePlacementAccessWithRelation(shardPlacement, accessType);
```
marcocitus: I'm a bit confused as to why we need to do this for all placements, since we ultimately only care about the relationId.

src/backend/distributed/transaction/relation_access_tracking.c
```diff
@@ -0,0 +1,529 @@
...
+			List *partitionShardPlacementList = NIL;
+			ListCell *partitionCell = NULL;
+
+			/*
+			 * During create_distributed_table, the partitions may not
+			 * have been created yet.
+			 */
+			if (!IsDistributedTable(partitionOid))
+			{
+				continue;
```
marcocitus: Might be worth adding a comment about why not registering is ok (we'll still register when we distribute the partitions).

src/backend/distributed/transaction/relation_access_tracking.c
onderkalaci: That's true, looping on the `placementList` is unnecessary.

But, I preferred to recursively call `AssociatePlacementAccessWithRelation ()` to be able to support multi-level partitioned tables. 

(though we currently don't support those). I thought if we one day decide to implement that, this would probably be forgotten.

I added some comment about that.

src/backend/distributed/transaction/relation_access_tracking.c
onderkalaci: added

src/backend/distributed/transaction/relation_access_tracking.c
```diff
@@ -0,0 +1,532 @@
...
+			partitionShardPlacementList = ShardPlacementList(partitionShardId);
+			partitionInitialShardPlacement =
+				(ShardPlacement *) linitial(partitionShardPlacementList);
+
+			/*
+			 * Recursively record all relation accesses of its partitions. Note that
+			 * we prefer to recursively call AssociatePlacementAccessWithRelation()
+			 * to support multi-level partitioned tables.
+			 */
+			AssociatePlacementAccessWithRelation(partitionInitialShardPlacement,
```
marcocitus: What if we split this function into 2? One that takes a shardId and a wrapper that takes a placement. Then we can call the former recursively without having to look up placement metadata.

src/backend/distributed/transaction/relation_access_tracking.c
```diff
@@ -0,0 +1,539 @@
...
+ * AssociateShardAccessWithRelation associates the shard access to the
+ * distributed relation that the shard belongs to.
+ *
+ * We implemented this function to prevent accessing placement metadata during
+ * recursive calls of the function itself.
+ */
+static void
+AssociateShardAccessWithRelation(uint64 shardId, ShardPlacementAccessType accessType)
+{
+	Oid relationId = RelationIdForShard(shardId);
```
marcocitus: now it strikes me that we could even skip the call to ColocatedShardIdInRelation if this function took a relationId instead

src/backend/distributed/master/master_stage_protocol.c
```diff
@@ -531,8 +547,38 @@ CreateShardsOnWorkers(Oid distributedRelationId, List *shardPlacements,
...
+			parentRelationShard->shardId =
+				ColocatedShardIdInRelation(parentRelationShard->relationId, shardIndex);
+			partitionRelationShard->relationId = distributedRelationId;
+			partitionRelationShard->shardId = shardId;
+
+			relationShardList = list_make2(parentRelationShard, partitionRelationShard);
+			placementAccessList = BuildPlacementDDLList(shardPlacement->groupId,
+														relationShardList);
+
+			connection = GetPlacementListConnection(connectionFlags, placementAccessList,
```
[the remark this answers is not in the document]
marcocitus: (best left to a separate PR if we decide to address it)

src/backend/distributed/transaction/relation_access_tracking.c
onderkalaci: makes sense, done.

src/backend/distributed/master/master_stage_protocol.c
```diff
@@ -518,6 +521,19 @@ CreateShardsOnWorkers(Oid distributedRelationId, List *shardPlacements,
 		CoordinatedTransactionUse2PC();
 	}
 
+	/* mark parallel relation accesses before opening connections */
+	if (ShouldRecordRelationAccess() && useExclusiveConnection)
+	{
+		RecordParallelDDLAccess(distributedRelationId);
+
+		/* we should mark the parent as well */
+		if (alterTableAttachPartitionCommand != NULL)
```
mtuncer: @onder @marcocitus this was to check if the distributed table is a partition of another distributed table. Right ?

src/backend/distributed/master/master_stage_protocol.c
```diff
@@ -531,8 +547,38 @@ CreateShardsOnWorkers(Oid distributedRelationId, List *shardPlacements,
...
 
-		connection = GetPlacementConnection(connectionFlags, shardPlacement,
-											placementOwner);
+		/*
+		 * For partitions, make sure that we mark the parent table relation access
+		 * with DDL. This is only important for parallel relation access in transaction
+		 * blocks, thus check useExclusiveConnection and transaction block as well.
+		 */
+		if ((ShouldRecordRelationAccess() && useExclusiveConnection) &&
+			alterTableAttachPartitionCommand != NULL)
```
mtuncer: same here (alterTableAttachPartitionCommand != NULL) meaning this table is a partition of a distributed partitioned table.

src/backend/distributed/master/master_stage_protocol.c
marcocitus: Yes, this should use the function call instead.

mtuncer:
@onderkalaci @marcocitus do we need DESCRIPTION ?

furkansahin:
@onderkalaci @marcocitus Do we need a DESCRIPTION?

marcocitus:
No, it's infrastructure for foreign keys to reference tables.

PR #2222, opened 2018-06-14
Fix use-after-free that may occur for INSERT..SELECT in prepared statements

### LINKED ISSUE #2217: "Unrecognized node type" error on rollups
Customer has identified a rollup function that reliably throws an "Unrecognized node type" message when attempting to INSERT INTO from a distributed table. 

Client has provided a gist that can automatically reproduce this on 7.3 (client version) and 7.4 (verified to reproduce behavior in my testing). 

https://gist.github.com/tbug/9b8cc6715964569e9dba0f15f690873f

marcocitus:
Fixes #2217 

Basically, we're currently allowing the planner to scribble on the parse tree of the SELECT query (e.g. Citus does this when there are recursively planned subqueries) when executing INSERT..SELECT via the coordinator, but the parse tree may be stored in a prepared plan. The changes to the parse tree will not be in a persistent memory context, hence this may cause crashes or undefined behaviour.

Note: I know I could have added more tests, but I don't have a lot of time.

### DIFF
--- a/src/backend/distributed/executor/insert_select_executor.c
+++ b/src/backend/distributed/executor/insert_select_executor.c
@@ -101,6 +101,7 @@ ExecuteSelectIntoRelation(Oid targetRelationId, List *insertTargetList,
 	int partitionColumnIndex = -1;
 
 	CitusCopyDestReceiver *copyDest = NULL;
+	Query *queryCopy = NULL;
 
 	partitionMethod = PartitionMethod(targetRelationId);
 	if (partitionMethod == DISTRIBUTE_BY_NONE)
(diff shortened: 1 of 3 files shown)


sumedhpathak:
Hey Jason, could you review this change? Thanks!

jasonmp85:
> Hey Jason, could you review this change? Thanks!

Yup; on it.

sumedhpathak:
@marcocitus When would the Insert/Select via coordinator cause the SELECT to be a prepared plan?

marcocitus:
PL/pgSQL typically generates prepared plans.

jasonmp85:
Tested that the test exposed the problem and that the patch fixes it. This class of problems (memory lifetime of objects that might be stored in plans) seems to bite us a lot; not right now, but maybe later, it would be worth investigating more general solutions. Some of our previous solutions have been pretty hacky.

jasonmp85:
@marcocitus merged this, but it'll still need a cherry-pick into the release.

PR #2224, opened 2018-06-18
Implement C interface for setting GUC

onderkalaci:
We need the ability to switch to sequential mode (e.g.,
 SET LOCAL citus.multi_shard_modify_mode = 'sequential'). This
commit enables that.

### DIFF
--- a/src/backend/distributed/executor/multi_executor.c
+++ b/src/backend/distributed/executor/multi_executor.c
@@ -47,8 +47,42 @@
 int MultiShardConnectionType = PARALLEL_CONNECTION;
 
 
-/* ocal function forward declarations */
+/* local function forward declarations */
+static void SetLocalMultiShardModifyModeToSequential(void);
 static Relation StubRelation(TupleDesc tupleDescriptor);
+
+
+/* exports for SQL callable functions */
+PG_FUNCTION_INFO_V1(set_local_multi_shard_modify_mode_to_sequential);
+
+
+/*
+ * set_local_multi_shard_modify_mode_to_sequential is a SQL
+ * interface for testing SetLocalMultiShardModifyModeToSequential().
+ */
+Datum
+set_local_multi_shard_modify_mode_to_sequential(PG_FUNCTION_ARGS)
+{
+	SetLocalMultiShardModifyModeToSequential();
+
+	PG_RETURN_VOID();
+}
+
+
+/*
+ * SetLocalMultiShardModifyModeToSequential simply a C interface for
+ * setting the following:
+ *      SET LOCAL citus.multi_shard_modify_mode = 'sequential';
+ */
+static void
+SetLocalMultiShardModifyModeToSequential()
+{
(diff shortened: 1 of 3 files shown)


marcocitus:
Can't we just use `set_config_option`?

onderkalaci:
> Can't we just use set_config_option?

Actually, `set_config_option` seemed a bit too low level to call (with some slightly hard to understand parameters). 

But seems `set_config_option` cheaper to use and we might call this function often in the near future, so I'll switch to that.

furkansahin:
This is good. :shipit:

src/backend/distributed/executor/multi_executor.c
```diff
@@ -47,10 +47,44 @@
...
+/* exports for SQL callable functions */
+PG_FUNCTION_INFO_V1(set_local_multi_shard_modify_mode_to_sequential);
+
+
+/*
+ * set_local_multi_shard_modify_mode_to_sequential is a SQL
+ * interface for testing SetLocalMultiShardModifyModeToSequential().
+ */
+Datum
+set_local_multi_shard_modify_mode_to_sequential(PG_FUNCTION_ARGS)
```
marcocitus: should this be in the test directory?

onderkalaci: Yes, maybe. But since we're creating the function in the test sql file, I thought it's OK to not create a new file in the test sources that'd only consists of this UDF. And, I've already merged this to master. Do you prefer to revert this change?

onderkalaci: I think in general we don't have any strict policies around such functions. Do you see any drawbacks of just something like in this PR? I prefer to create a new source file in the test whenever there are multiple functions such as #2220.

marcocitus: It's weird to have unit testing code in the main executor logic. Maybe just push a commit to master that moves it to one of the existing test files or a new one.

jasonmp85:
I'm a bit late, but `SetConfigOption`, `SetPGVariable`, and `set_config_by_name` all seem higher-level.

onderkalaci:
> I'm a bit late, but SetConfigOption, SetPGVariable, and set_config_by_name all seem higher-level.

Interesting, I haven't seen those functions. But, we need to use `SET LOCAL ...` inside a transaction, and those APIs seems to miss `GucAction` (`GUC_ACTION_LOCAL`) variable. So, seems slightly higher than what we need 😄 

(`set_config_by_name` seems to have that parameter, but, we probably don't prefer to do a function call here)

src/backend/distributed/executor/multi_executor.c
jasonmp85: Don't we already have a test file like this?

onderkalaci: Yes, we moved this function to `distributed/test` folder with this [PR](https://github.com/citusdata/citus/pull/2229/files) after the discussion here.

PR #2227, opened 2018-06-19
create_distributed_table honors sequential mode

furkansahin:
This PR adds `citus.multi_shard_modify_mode` check to `create_distributed_table`. 
It is important to note that `sequential` mode cannot be used for the relations with data for `create_distributed_table` since we cannot execute `COPY` sequentially.
If the `sequential` mode is used, than `create_distributed_table` does not claim the connections exclusively.

### DIFF
--- a/src/backend/distributed/commands/create_distributed_table.c
+++ b/src/backend/distributed/commands/create_distributed_table.c
@@ -40,6 +40,7 @@
 #include "distributed/metadata_cache.h"
 #include "distributed/metadata_sync.h"
 #include "distributed/multi_copy.h"
+#include "distributed/multi_executor.h"
 #include "distributed/multi_logical_planner.h"
 #include "distributed/multi_partitioning_utils.h"
 #include "distributed/multi_utility.h"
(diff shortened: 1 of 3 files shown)


src/test/regress/expected/sequential_modifications.out
```diff
@@ -491,6 +490,142 @@ SELECT pg_reload_conf();
  t
 (1 row)
 
+-- The following tests are added to test if create_distributed_table honors sequential mode
```
onderkalaci: Why don't we have any tests that'd fail with `cannot distribute relations with data...`

We should have the following test:
```SQL
CREATE TABLE test_create_seq_table (a int);
INSERT INTO test_create_seq_table VALUES (1);

SET citus.multi_shard_modify_mode TO 'sequential';
SELECT create_distributed_table('test_create_seq_table' ,'a');

RESET citus.multi_shard_modify_mode;

BEGIN;
	SET LOCAL citus.multi_shard_modify_mode TO 'sequential';
	select create_distributed_table('test_create_seq_table' ,'a');
ROLLBACK;
```

src/backend/distributed/commands/create_distributed_table.c
```diff
@@ -475,9 +476,34 @@ CreateHashDistributedTableShards(Oid relationId, Oid colocatedTableId,
...
+	/*
+	 * Decide whether to use exclusive connections per placement or not. Note that
+	 * if the local table is not empty, we cannot use sequential mode since the COPY
+	 * operation that'd load the data into shards currently requires exclusive
+	 * connections.
+	 */
 	if (RegularTable(relationId))
 	{
-		useExclusiveConnection = IsTransactionBlock() || !localTableEmpty;
+		if (!localTableEmpty && MultiShardConnectionType == SEQUENTIAL_CONNECTION)
```
onderkalaci: As we've discussed in person, @furkansahin is going to move this logic to a separate function in the next PR, seems acceptable for now.

furkansahin:
@mtuncer , no. It is covered by #2240

PR #2231, opened 2018-06-20
Add connparam invalidation trigger creation logic

jasonmp85:
This needs to live in Community, since we haven't yet added the complication of having divergent upgrade scripts in Enterprise.


src/backend/distributed/citus--7.5-2--7.5-3.sql
```diff
@@ -0,0 +1,28 @@
...
+CREATE FUNCTION task_tracker_conninfo_cache_invalidate()
+    RETURNS trigger
+    LANGUAGE C
+    AS 'citus', $$task_tracker_conninfo_cache_invalidate$$;
+COMMENT ON FUNCTION task_tracker_conninfo_cache_invalidate()
+    IS 'invalidate task-tracker conninfo cache';
+
+CREATE TRIGGER dist_authinfo_cache_invalidate
+    AFTER INSERT OR UPDATE OR DELETE
+    ON pg_catalog.pg_dist_authinfo
```
marcocitus: Could probably add OR TRUNCATE

jasonmp85: Done.

PR #2232, opened 2018-06-21
Use non-data connection for intermediate results

onderkalaci:
Make sure that intermediate results use a connection that is
not associated with any placement. That is useful in two ways:
    - More complex queries can be executed with CTEs
    - Safely use the same connections when there is a foreign key
      to reference table from a distributed table, (which has to
      use the same connection for modifications since the reference
      table might have cascaded to the distributed table).

### DIFF
--- a/src/backend/distributed/connection/connection_management.c
+++ b/src/backend/distributed/connection/connection_management.c
@@ -128,6 +128,50 @@ GetNodeConnection(uint32 flags, const char *hostname, int32 port)
 
 
 /*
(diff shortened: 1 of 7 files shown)


mtuncer:
@onderkalaci @marcocitus DESCRIPTION ?

furkansahin:
@onderkalaci @marcocitus Do we need a DESCRIPTION?

marcocitus:
It's mainly foreign key related so probably covered under that umbrella.

PR #2235, opened 2018-06-25
Add groundwork for citus_stat_statements api

mtuncer:
- Adds descriptions of citus_stat_statements view, citus_query_stats UDF and citus_stat_statements_reset UDF
- Adds C implementation placeholders for citus_query_stats and and citus_stat_statements_reset UDSs
- Provides entry points for shared memory initialization functions
- Adds queryId field to distributed plan struct.

### DIFF
--- a/src/backend/distributed/citus--7.5-3--7.5-4.sql
+++ b/src/backend/distributed/citus--7.5-3--7.5-4.sql
@@ -1,0 +1,47 @@
+/* citus--7.5-3--7.5-4 */
+
+CREATE FUNCTION pg_catalog.citus_query_stats(OUT queryid bigint,
+											 OUT userid oid,
+											 OUT dbid oid,
+											 OUT executor bigint,
+											 OUT partition_key text,
+											 OUT calls bigint)
+RETURNS SETOF record
+LANGUAGE C STRICT
+AS 'MODULE_PATHNAME', $$citus_query_stats$$;
+
+CREATE FUNCTION pg_catalog.citus_stat_statements_reset()
+RETURNS VOID
+LANGUAGE C STRICT
+AS 'MODULE_PATHNAME', $$citus_stat_statements_reset$$;
+
+CREATE FUNCTION pg_catalog.citus_stat_statements(OUT queryid bigint,
+												 OUT userid oid,
+												 OUT dbid oid,
+												 OUT query text,
+												 OUT executor bigint,
+												 OUT partition_key text,
+												 OUT calls bigint)
+RETURNS SETOF record
+LANGUAGE plpgsql
+AS $citus_stat_statements$
+BEGIN
+ IF EXISTS (
+ 	SELECT extname FROM pg_extension
+ 	WHERE extname = 'pg_stat_statements')
+ THEN
+ 	RETURN QUERY SELECT pss.queryid, pss.userid, pss.dbid, pss.query, cqs.executor,
+ 						cqs.partition_key, cqs.calls
+ 				 FROM pg_stat_statements(true) pss
+ 				 	JOIN citus_query_stats() cqs
+ 				 	USING (queryid);
+ ELSE
+    RAISE EXCEPTION 'pg_stat_statements is not installed'
+    	USING HINT = 'install pg_stat_statements extension and try again';
+ END IF;
+END;
+$citus_stat_statements$;
+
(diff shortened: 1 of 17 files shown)


mtuncer:
description field is not added intentionally

src/backend/distributed/citus--7.5-3--7.5-4.sql
```diff
@@ -0,0 +1,47 @@
...
+												 OUT query text,
+												 OUT executor bigint,
+												 OUT partition_key text,
+												 OUT calls bigint)
+RETURNS SETOF record
+LANGUAGE plpgsql
+AS $citus_stat_statements$
+BEGIN
+ IF EXISTS (
+ 	SELECT name FROM pg_available_extensions
```
marcocitus: `pg_available_extensions` are extensions that are installed, but haven't been created. Maybe use `pg_extension` instead.

src/include/distributed/multi_physical_planner.h
```diff
@@ -127,7 +127,7 @@ typedef struct Job
 	bool subqueryPushdown;
 	bool requiresMasterEvaluation; /* only applies to modify jobs */
 	bool deferredPruning;
-	Const *partitionValueConst;
+	Const *partitionKeyConst;
```
marcocitus: I'm a bit confused about this change. The `Const` represents a value, not a key, right?

src/backend/distributed/executor/query_stats.c
```diff
@@ -0,0 +1,56 @@
...
+#include "distributed/query_stats.h"
+
+PG_FUNCTION_INFO_V1(citus_stat_statements_reset);
+PG_FUNCTION_INFO_V1(citus_query_stats);
+
+
+/* placeholder for InitializeCitusQueryStats */
+void
+InitializeCitusQueryStats(void)
+{ }
```
marcocitus: could maybe put a small comment in the function to fix the alignment and tell the reader why it is empty.

src/backend/distributed/citus--7.5-3--7.5-4.sql
mtuncer: done

src/include/distributed/multi_physical_planner.h
mtuncer: yes it is a value.  I am making a naming change in the api. going for partition_key instead of partition_value. Here making it partitionKeyValueConst might be better for readability

src/backend/distributed/citus--7.5-3--7.5-4.sql
```diff
@@ -0,0 +1,47 @@
...
+												 OUT query text,
+												 OUT executor bigint,
+												 OUT partition_key text,
+												 OUT calls bigint)
+RETURNS SETOF record
+LANGUAGE plpgsql
+AS $citus_stat_statements$
+BEGIN
+ IF EXISTS (
+ 	SELECT name FROM pg_extension
```
marcocitus: this gives me an error:
```
ERROR:  column "name" does not exist
LINE 2:   SELECT name FROM pg_extension
```

Should be `SELECT extname`, though since we don't care about the value, event `SELECT 1` is sufficient.

src/backend/distributed/citus--7.5-3--7.5-4.sql
```diff
@@ -0,0 +1,47 @@
...
+ 	WHERE name = 'pg_stat_statements')
```
marcocitus: name -> extname

mtuncer: 😞

src/backend/distributed/citus--7.5-3--7.5-4.sql
```diff
@@ -0,0 +1,47 @@
...
+ THEN
+ 	RETURN QUERY SELECT pss.queryid, pss.userid, pss.dbid, pss.query, cqs.executor,
+ 						cqs.partition_key, cqs.calls
+ 				 FROM pg_stat_statements(true) pss
+ 				 	JOIN citus_query_stats() cqs
```
marcocitus: Slightly funny thing happens here when `pg_stat_statements` is loaded but completely empty (which it rarely is), because then citus_query_stats will never be called and we don't get an error. Seems ok though.

PR #2237, opened 2018-06-27
Enforce foreign key restrictions inside transaction blocks

onderkalaci:
The PR is based on multiple other PRs, but the only commit in this PR covers only the changes I'd like to get merged.

When a hash distributed table have a foreign key to a reference
table, there are few restrictions we have to apply in order to
prevent distributed deadlocks or reading wrong results.

The necessity to apply the restrictions arise from cascading
nature of foreign keys. When a foreign key on a reference table
cascades to a distributed table, a single operation over a single
connection can acquire locks on multiple shards of the distributed
table. Thus, any parallel operation on that distributed table, in the
same transaction should not open parallel connections to the shards.
Otherwise, we'd either end-up with a self-distributed deadlock or
read wrong results.

As briefly described above, the restrictions that we apply is done
by tracking the distributed/reference relation accesses inside 
transaction blocks, and act accordingly when necessary.

The two main rules inside a transaction block are as follows:
   - Whenever a parallel distributed relation access conflicts
     with a consecutive reference relation access, Citus errors
     out
   - Whenever a reference relation access is followed by a 
     conflicting parallel relation access, the execution mode
     is switched to sequential mode.

There are also some other notes to mention: 
   - If the user does SET LOCAL citus.multi_shard_modify_mode 
     TO 'sequential';, all the queries should simply work with
     using one connection per worker and sequentially executing
     the commands. That's obviously a slower approach than Citus'
     usual parallel execution. However, we've at least have a way
     to run all commands successfully.

   - If an unrelated parallel query executed on any distributed
     table, we cannot switch to sequential mode. Because, the essense
     of sequential mode is using one connection per worker. However,
     in the presence of a parallel connection, the connection manager
     picks those connections to execute the commands. That contradicts
     with our purpose, thus we error out.

   - COPY to a distributed table cannot be executed in sequential mode.
     Thus, if we switch to sequential mode and COPY is executed, the 
     operation fails and there is currently no way of implementing that.
     Note that, when the local table is not empty and create_distributed_table
     is used, citus uses COPY internally. Thus, in those cases,
     create_distributed_table() will also fail.

   - There is a GUC called citus.enforce_foreign_key_restrictions 
     to disable all the checks. We added that GUC since the restrictions
     we apply is sometimes a bit more restrictive than its necessary.
     The user might want to relax those. Similarly, if you don't have
     CASCADEing reference tables, you might consider disabling all the 
     checks.

Things to cover:
- [x] better error message - hints
- [x] More comprehensive regression tests (multiple commands inside a tx)
- [x] Try to relax some of the restrictions
    - [x]  Mostly when should `SELECT`s need to conflict with DDLs?
- [x] Fix failing tests once @furkansahin's changes merged into `foreign_key_master' branch
- [x] Improve code quality a bit more
- [x] When cache-invalidation of the graph is fixed, try more tests with `CREATE TABLE / ADD CONSTRAINT` in the same tx.
- [X] Better PR description

### DIFF
--- a/src/backend/distributed/shared_library_init.c
+++ b/src/backend/distributed/shared_library_init.c
@@ -38,6 +38,7 @@
 #include "distributed/multi_utility.h"
 #include "distributed/pg_dist_partition.h"
 #include "distributed/placement_connection.h"
+#include "distributed/relation_access_tracking.h"
 #include "distributed/query_pushdown_planning.h"
 #include "distributed/query_stats.h"
 #include "distributed/remote_commands.h"
@@ -400,6 +401,20 @@ RegisterCitusConfigVariables(void)
 		PGC_USERSET,
 		GUC_NO_SHOW_ALL,
 		NULL, NULL, NULL);
+
+	DefineCustomBoolVariable(
+		"citus.enforce_foreign_key_restrictions",
+		gettext_noop("Enforce restrictions while querying distributed/reference "
+					 "tables with foreign keys"),
+		gettext_noop("When enabled, cascading modifications from reference tables "
+					 "to distributed tables are traced and acted accordingly "
+					 "to avoid creating distributed deadlocks and ensure correctness."),
+		&EnforceForeignKeyRestrictions,
+		true,
+		PGC_USERSET,
+		GUC_NO_SHOW_ALL,
+		NULL, NULL, NULL);
+
 
 	DefineCustomBoolVariable(
 		"citus.subquery_pushdown",
--- a/src/backend/distributed/transaction/relation_access_tracking.c
+++ b/src/backend/distributed/transaction/relation_access_tracking.c
@@ -20,14 +20,27 @@
 #include "access/xact.h"
 #include "distributed/colocation_utils.h"
 #include "distributed/hash_helpers.h"
+#include "distributed/multi_executor.h"
 #include "distributed/multi_join_order.h"
 #include "distributed/multi_partitioning_utils.h"
 #include "distributed/metadata_cache.h"
 #include "distributed/relation_access_tracking.h"
 #include "utils/hsearch.h"
+#include "utils/lsyscache.h"
 
+
+/* Config variables managed via guc.c */
+bool EnforceForeignKeyRestrictions = true;
 
 #define PARALLEL_MODE_FLAG_OFFSET 3
+#define PARALLEL_ACCESS_MASK (int) (0 | \
+									(1 << (PLACEMENT_ACCESS_SELECT + \
+										   PARALLEL_MODE_FLAG_OFFSET)) | \
+									(1 << (PLACEMENT_ACCESS_DML + \
+										   PARALLEL_MODE_FLAG_OFFSET)) | \
+									(1 << (PLACEMENT_ACCESS_DDL + \
+										   PARALLEL_MODE_FLAG_OFFSET)))
+
 
 /*
  * Hash table mapping relations to the
@@ -66,6 +79,7 @@ typedef struct RelationAccessHashEntry
 static HTAB *RelationAccessHash;
 
 
+/* functions related to access recording */
 static void RecordRelationAccess(Oid relationId, ShardPlacementAccessType accessType);
 static void RecordPlacementAccessToCache(Oid relationId,
 										 ShardPlacementAccessType accessType);
@@ -75,6 +89,26 @@ static void RecordParallelRelationAccess(Oid relationId, ShardPlacementAccessTyp
 										 placementAccess);
 static void RecordParallelRelationAccessToCache(Oid relationId,
 												ShardPlacementAccessType placementAccess);
+
+/* functions related to access conflict checks */
+static char * PlacementAccessTypeToText(ShardPlacementAccessType accessType);
+static void CheckConflictingRelationAccesses(Oid relationId,
+											 ShardPlacementAccessType accessType);
+static bool HoldsConflictingLockWithReferencingRelations(Oid relationId,
+														 ShardPlacementAccessType
+														 placementAccess,
+														 Oid *conflictingRelationId,
+														 ShardPlacementAccessType *
+														 conflictingAccessMode);
+static void CheckConflictingParallelRelationAccesses(Oid relationId,
+													 ShardPlacementAccessType
+													 accessType);
+static bool HoldsConflictingLockWithReferencedRelations(Oid relationId,
+														ShardPlacementAccessType
+														placementAccess,
+														Oid *conflictingRelationId,
+														ShardPlacementAccessType *
+														conflictingAccessMode);
 
 
 /*
@@ -132,6 +166,40 @@ AssociatePlacementAccessWithRelation(ShardPlacement *placement,
 
 
 /*
+ * PlacementAccessTypeToText converts ShardPlacementAccessType to
+ * text representation.
+ */
+static char *
+PlacementAccessTypeToText(ShardPlacementAccessType accessType)
+{
+	switch (accessType)
+	{
+		case PLACEMENT_ACCESS_SELECT:
+		{
+			return "SELECT";
+			break;
+		}
+
+		case PLACEMENT_ACCESS_DML:
+		{
+			return "DML";
+		}
+
+		case PLACEMENT_ACCESS_DDL:
+		{
+			return "DDL";
+		}
+
+		default:
+		{
+			return "None";
+			break;
+		}
+	}
+}
+
+
+/*
  * RecordRelationAccess associates the access to the distributed relation. The
  * function takes partitioned relations into account as well.
  *
@@ -142,6 +210,9 @@ AssociatePlacementAccessWithRelation(ShardPlacement *placement,
 static void
 RecordRelationAccess(Oid relationId, ShardPlacementAccessType accessType)
 {
+	/* make sure that this is not a conflicting access */
+	CheckConflictingRelationAccesses(relationId, accessType);
+
 	/*
 	 * If a relation is partitioned, record accesses to all of its partitions as well.
 	 * We prefer to use PartitionedTableNoLock() because at this point the necessary
@@ -374,6 +445,9 @@ RecordParallelRelationAccess(Oid relationId, ShardPlacementAccessType placementA
 		return;
 	}
 
+	/* act accordingly if it's a conflicting access */
+	CheckConflictingParallelRelationAccesses(relationId, placementAccess);
+
 	/*
 	 * If a relation is partitioned, record accesses to all of its partitions as well.
 	 * We prefer to use PartitionedTableNoLock() because at this point the necessary
@@ -431,6 +505,40 @@ RecordParallelRelationAccessToCache(Oid relationId,
 	/* set the bit representing access mode */
 	parallelRelationAccessBit = placementAccess + PARALLEL_MODE_FLAG_OFFSET;
 	hashEntry->relationAccessMode |= (1 << parallelRelationAccessBit);
+}
+
+
+/*
+ * ParallelQueryExecutedInTransaction returns true if any parallel query
+ * is executed in the current transaction.
+ */
+bool
+ParallelQueryExecutedInTransaction(void)
+{
+	HASH_SEQ_STATUS status;
+	RelationAccessHashEntry *hashEntry;
+
+	if (!ShouldRecordRelationAccess() || RelationAccessHash == NULL)
+	{
+		return false;
+	}
+
+	hash_seq_init(&status, RelationAccessHash);
+
+	hashEntry = (RelationAccessHashEntry *) hash_seq_search(&status);
+	while (hashEntry != NULL)
+	{
+		int relationAccessMode = hashEntry->relationAccessMode;
+		if ((relationAccessMode & PARALLEL_ACCESS_MASK))
+		{
+			hash_seq_term(&status);
+			return true;
+		}
+
+		hashEntry = (RelationAccessHashEntry *) hash_seq_search(&status);
+	}
+
+	return false;
 }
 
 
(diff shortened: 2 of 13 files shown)


src/backend/distributed/shared_library_init.c
```diff
@@ -401,6 +402,20 @@ RegisterCitusConfigVariables(void)
 		GUC_NO_SHOW_ALL,
 		NULL, NULL, NULL);
 
+	DefineCustomBoolVariable(
+		"citus.enforce_foreign_key_restrictions",
+		gettext_noop("Enforce restrictions while querying distributed/reference "
+					 "tables with foreign keys"),
+		gettext_noop("When enabled, cascading modifications from reference tables "
+					 "to distributed tables are traced and acted accordingly "
+					 "to avoid creating distributed deadlocks and ensure correctness."),
```
marcocitus: I'm not sure in what situation we would not want to do this. We can easily violate visibility rules if this is disabled.

onderkalaci: I had two main motivations:

1) Foreign keys without `CASCADE` option
2) As you've noted below, some of the operations do not really affect each other and if someone needs to disable the checks, we'd have at least a way to provide that:
> For example, a CREATE INDEX on a reference table shouldn't really affect SELECTs or CREATE INDEX on a distributed table in the same transaction, but currently it would be classified as DDL and therefore everything gets serialised.

onderkalaci: Though I can make this `SHOW_NO_ALL`, would that work?

onderkalaci:
@marcocitus Sorry for the long delay, the PR is now rebased onto all the other PRs related to foreign keys to reference tables and should be ready for you to look again.

It'd be great if you share your thoughts on the error messages / debug messages. I find them slightly more technical then a regular user could understand (e.g., mentionin about parallel  / sequential modes etc). However, I cannot come up with simpler, less technical error messages.

src/backend/distributed/shared_library_init.c
marcocitus: > Though I can make this SHOW_NO_ALL, would that work?

yes, that would work

src/backend/distributed/transaction/relation_access_tracking.c
```diff
@@ -434,6 +500,78 @@ RecordParallelRelationAccessToCache(Oid relationId,
...
+	smallestParallelRelationAccessBit =
+		PLACEMENT_ACCESS_SELECT + PARALLEL_MODE_FLAG_OFFSET;
+	smallestParallelValue = (1 << smallestParallelRelationAccessBit);
+
+	hash_seq_init(&status, RelationAccessHash);
+
+	hashEntry = (RelationAccessHashEntry *) hash_seq_search(&status);
+	while (hashEntry != NULL)
+	{
+		if (hashEntry->relationAccessMode >= smallestParallelValue)
```
marcocitus: This seems slightly dodgy, e.g. it breaks if we start using additional bits with a different meaning. 

Maybe we should have a bit mask `(hashEntry->relationAccessMode & PARALLEL_ACCESS_MASK) != 0`

src/backend/distributed/transaction/relation_access_tracking.c
```diff
@@ -524,10 +662,340 @@ GetRelationAccessMode(Oid relationId, ShardPlacementAccessType accessType)
...
+
+	cacheEntry = DistributedTableCacheEntry(relationId);
+
+	if (!(cacheEntry->partitionMethod == DISTRIBUTE_BY_NONE &&
+		  cacheEntry->referencingRelationsViaForeignKey != NIL))
+	{
+		return;
+	}
+
+	if (HoldsConflictingLockWithReferencingRelations(relationId, accessType,
```
marcocitus: Note for the future: it would be nice if we only considered writes to reference tables that touch columns in the foreign key.

src/backend/distributed/transaction/relation_access_tracking.c
```diff
@@ -524,10 +662,340 @@ GetRelationAccessMode(Oid relationId, ShardPlacementAccessType accessType)
...
+	}
+	else if (cacheEntry->referencingRelationsViaForeignKey != NIL &&
+			 MultiShardConnectionType == PARALLEL_CONNECTION &&
+			 accessType > PLACEMENT_ACCESS_SELECT)
+	{
+		char *relationName = get_rel_name(relationId);
+
+		ereport(DEBUG1, (errmsg("switching to sequential query execution mode"),
+						 errdetail("Reference relation \"%s\" is modified, which might "
+								   "have cascaded to hash distributed relations via "
```
marcocitus: it's not just that it might cascade, but also that updates/inserts on the distributed table that affect columns in the foreign key will need to take a row-level (share) lock on the reference table.

src/test/regress/expected/foreign_key_restriction_enforcement.out
```diff
@@ -0,0 +1,1089 @@
...
+ 
+(1 row)
+
+ALTER TABLE on_update_fkey_table ADD CONSTRAINT fkey FOREIGN KEY(value_1) REFERENCES referece_table(id) ON UPDATE CASCADE;
+INSERT INTO referece_table SELECT i FROM generate_series(0, 100) i;
+INSERT INTO on_update_fkey_table SELECT i, i % 100  FROM generate_series(0, 1000) i;
+INSERT INTO unrelated_dist_table SELECT i, i % 100  FROM generate_series(0, 1000) i;
+-- in order to see when the mode automatically swithces to sequential execution
+SET client_min_messages TO DEBUG1;
+-- case 1.1: SELECT to a reference table is followed by a parallel SELECT to a distributed table
```
marcocitus: is there some method to the numbering?

src/test/regress/expected/foreign_key_restriction_enforcement.out
```diff
@@ -0,0 +1,1089 @@
...
+DETAIL:  Reference relation "referece_table" is modified, which might have cascaded to hash distributed relations via foreign keys. Any parallel modification to those hash distributed relations in the same transaction can only be executed in sequential query execution mode
+	ALTER TABLE on_update_fkey_table ADD COLUMN value_1_X INT;
+ROLLBACK;
+-- case 2.7: UPDATE to a reference table is followed by COPY
+BEGIN;
+	UPDATE referece_table SET id = 101 WHERE id = 99;
+DEBUG:  switching to sequential query execution mode
+DETAIL:  Reference relation "referece_table" is modified, which might have cascaded to hash distributed relations via foreign keys. Any parallel modification to those hash distributed relations in the same transaction can only be executed in sequential query execution mode
+	COPY on_update_fkey_table FROM STDIN WITH CSV;
+ERROR:  cannot execute parallel COPY on relation "on_update_fkey_table" since it holds conflicting locks via foreign keys on reference relation "referece_table" due to a previous "DML" access in the same transaction
```
marcocitus: I think we can still be a bit more informative and provide a hint, e.g.
```
ERROR: cannot execute COPY on relation "on_update_fkey_table, since it has a foreign key to "referece_table" and "referece_table" has been modified in this transaction
DETAIL: COPY to a distributed table uses a separate set of connections which will not be able to see the uncommitted changes to the reference table.
HINT: Perform the COPY in a separate transaction
```

src/test/regress/expected/foreign_key_restriction_enforcement.out
```diff
@@ -0,0 +1,1089 @@
...
+	SELECT count(*) FROM on_update_fkey_table WHERE value_1 = 99;
+ count 
+-------
+    10
+(1 row)
+
+	UPDATE referece_table SET id = 101 WHERE id = 99;
+DEBUG:  switching to sequential query execution mode
+DETAIL:  Reference relation "referece_table" is modified, which might have cascaded to hash distributed relations via foreign keys. Any parallel modification to those hash distributed relations in the same transaction can only be executed in sequential query execution mode
+ERROR:  attempt to set query execution mode to sequential failed since a parallel query has already been executed in the same transaction
```
marcocitus: it's a pity that we have to give up here

anyway, the message seems a hard to digest, especially without the debug information.

Maybe:
```
ERROR: cannot modify reference table "referece_table" because there was a parallel operation on a distributed table with a foreign key to "referece_table" in the same transaction
HINT: ...
```

src/test/regress/expected/foreign_key_restriction_enforcement.out
```diff
@@ -0,0 +1,1089 @@
...
+ count 
+-------
+    10
+(1 row)
+
+	ALTER TABLE referece_table ALTER COLUMN id SET DATA TYPE smallint;
+DEBUG:  rewriting table "referece_table"
+DEBUG:  building index "referece_table_pkey" on table "referece_table"
+DEBUG:  validating foreign key constraint "fkey"
+ERROR:  cannot execute "DDL" on relation "referece_table" since it holds conflicting locks on relation "on_update_fkey_table" due to a previous parallel "SELECT" access in the same transaction
```
marcocitus: the quoting aroudn the operation ("DDL", "SELECT", "DML") looks a bit awkward

src/test/regress/expected/foreign_key_restriction_enforcement.out
```diff
@@ -0,0 +1,1089 @@
...
+BEGIN;
+	SELECT count(*) FROM on_update_fkey_table WHERE value_1 = 99;
+ count 
+-------
+    10
+(1 row)
+
+	TRUNCATE referece_table CASCADE;
+NOTICE:  truncate cascades to table "on_update_fkey_table"
+ERROR:  cannot execute "DDL" on relation "referece_table" since it holds conflicting locks on relation "on_update_fkey_table" due to a previous parallel "SELECT" access in the same transaction
```
marcocitus: > since it holds conflicting locks on relation "on_update_fkey_table" due to a previous parallel "SELECT" access in the same transaction

the conflicting locks bit is probably not going to be meaningful to most users (and it's not necessarily true), maybe:

"because there was a parallel SELECT on "on_update_fkey_table" in the same transaction"

src/test/regress/expected/foreign_key_restriction_enforcement.out
onderkalaci: Yes, but the parallel access doesn't need to be on a relation which has a foreign key to the reference table. The following is a comment from the code:
```
/*
* If there has already been a parallel query executed, the sequential mode
* would still use the already opened parallel connections to the workers,
* thus contradicting our purpose of using sequential mode.
*/
```

So, it could be a completely unrelated parallel access to a colocated table. Are you OK with this restriction? That is probably the most restrictive part of the PR, but, the fix seemed to be touching the connection management, so I've not gone into that for now.

marcocitus: It's not nice, but I wouldn't spend any time trying to fix it in 7.5.

I think we should do brainstorming on how to handle foreign keys to reference tables going forward.

src/backend/distributed/transaction/relation_access_tracking.c
```diff
@@ -524,10 +630,389 @@ GetRelationAccessMode(Oid relationId, ShardPlacementAccessType accessType)
...
+							   "\"%s\" in the same transaction",
+							   accessTypeText, relationName, conflictingAccessTypeText,
+							   conflictingRelationName),
+						errhint("Try re-running the transaction with "
+								"\"SET LOCAL citus.multi_shard_modify_mode TO "
+								"\'sequential\';\"")));
+	}
+	else if (cacheEntry->referencingRelationsViaForeignKey != NIL &&
+			 MultiShardConnectionType == PARALLEL_CONNECTION &&
+			 accessType > PLACEMENT_ACCESS_SELECT)
```
marcocitus: the condition here took me a while to parse , maybe we can add a comment

src/backend/distributed/transaction/relation_access_tracking.c
```diff
@@ -524,10 +630,389 @@ GetRelationAccessMode(Oid relationId, ShardPlacementAccessType accessType)
...
+		{
+			/*
+			 * If there has already been a parallel query executed, the sequential mode
+			 * would still use the already opened parallel connections to the workers,
+			 * thus contradicting our purpose of using sequential mode.
+			 */
+			ereport(ERROR, (errmsg("cannot modify reference table \"%s\" because there "
+								   "was a parallel operation on a distributed table",
+								   relationName),
+							errdetail("attempt to set query execution mode to sequential "
```
marcocitus: I think the detail describes an implementation detail that is probably not relevant to the user. I think one important "detail" here is that the reference table has foreign keys pointing to it.

"DETAIL: When there is a foreign key to a reference table, Citus needs to perform all operations over a single connection per node to ensure consistency."

small note: detail messages use full sentences.

src/test/regress/expected/foreign_key_restriction_enforcement.out
onderkalaci: Yes, I agree with that. We've ended up with many restrictions. Let's first see which piece of these restrictions will bite users more..

src/backend/distributed/transaction/relation_access_tracking.c
```diff
@@ -524,10 +630,389 @@ GetRelationAccessMode(Oid relationId, ShardPlacementAccessType accessType)
...
+													&conflictingReferencingRelationId,
+													&conflictingAccessType))
+	{
+		char *relationName = get_rel_name(relationId);
+		char *conflictingRelationName = get_rel_name(conflictingReferencingRelationId);
+
+		char *conflictingAccessTypeText =
+			PlacementAccessTypeToText(conflictingAccessType);
+
+		ereport(ERROR, (errmsg("cannot execute parallel COPY on relation "
```
marcocitus: I'd omit the conflicting locks bit, since this is a fairly deep down implementation detail. We know that this only becomes a problem in the modification case, so we can say that.

e.g. 
```
ERROR: cannot execute COPY on relation "on_update_fkey_table, since it has a foreign key to "referece_table" and "referece_table" has been modified in this transaction
```

src/backend/distributed/transaction/relation_access_tracking.c
```diff
@@ -524,10 +630,389 @@ GetRelationAccessMode(Oid relationId, ShardPlacementAccessType accessType)
...
+
+		if (ParallelQueryExecutedInTransaction())
+		{
+			/*
+			 * If there has already been a parallel query executed, the sequential mode
+			 * would still use the already opened parallel connections to the workers,
+			 * thus contradicting our purpose of using sequential mode.
+			 */
+			ereport(ERROR, (errmsg("cannot execute parallel %s on relation \"%s\" "
+								   "since it holds conflicting locks on reference "
```
marcocitus: I'd omit the conflicting locks bit, since this is a fairly deep down implementation detail of postgres and citus, and instead be more precise about what the user did that led to this situation:

e.g.
```
ERROR: cannot execute parallel %s on table \"%s\" after a %s command on reference table \"%s\" because there is a foreign key between them and the operation cannot be serialised
```

onderkalaci: I think I'll prefer `has been modified in this transaction` over `the operation cannot be serialized` as you've suggested above. Because that sounds easier to understand for anyone

marcocitus: note that my comments were about different messages. Here I think we don't know that the relation has been modified? (could be DDL after SELECT)

src/backend/distributed/transaction/relation_access_tracking.c
```diff
@@ -524,10 +630,396 @@ GetRelationAccessMode(Oid relationId, ShardPlacementAccessType accessType)
...
+		{
+			/*
+			 * If there has already been a parallel query executed, the sequential mode
+			 * would still use the already opened parallel connections to the workers,
+			 * thus contradicting our purpose of using sequential mode.
+			 */
+			ereport(ERROR, (errmsg("cannot execute parallel %s on relation \"%s\" "
+								   "after %s command on reference relation "
+								   "\"%s\" because there is a foreign key between "
+								   "them and \"%s\" has been modified in this transaction",
```
marcocitus: is "has been modified" is correct? e.g. could this be triggered by parallel SELECT + reference table SELECT + DDL on distributed table?

marcocitus: Yes, that's fine.

mtuncer:
@onderkalaci @marcocitus Do we need DESCRIPTION here ?

marcocitus:
Should be covered by https://github.com/citusdata/citus/pull/2240

Issue #1886, opened 2017-12-18
Support array_agg ( ORDER BY ) in multi-shard queries


marcocitus:
Queries across shards containing an array_agg ( ORDER BY ) at the top level error out:

```
SELECT x, array_agg(y ORDER BY y) FROM test GROUP BY x;
ERROR:  array_agg with order by is unsupported
```

It does work as a single-shard query
```
SELECT x, array_agg(y ORDER BY y) FROM test WHERE x = 1 GROUP BY x;
 x | array_agg 
---+-----------
 1 | {0,1}
(1 row)
```

or as a partitionable subquery:
```
SELECT * FROM (SELECT x, array_agg(y ORDER BY y) FROM test GROUP BY x) a;
 x  | array_agg 
----+-----------
  8 | {64}
  1 | {0,1}
...
(10 rows)
```

marcocitus:
Fixed by #3324

PR #2239, opened 2018-06-27
enable aggregate features (distinct, order by) when group by clause includes partition column

madd512:
This PR not currently working

What I'm trying to do here is enable queries like this:

```sql
SELECT partition_column, array_agg(a ORDER BY b) GROUP BY partition_column;
```

TablePartitioningSupportsDistinct seems to always return false and I don't know why. I don't know C, so this is a tricky thing for me to debug. Would love for someone to point me in the right direction.


marcocitus:
Thanks for looking into #1886 . The way aggregates at the top level of a query are implemented needs some cleaning up before we can address this particular issue. 

If you want to do this right now you can wrap your query in a subquery.
`SELECT * FROM (SELECT partition_column, array_agg(a ORDER BY b) FROM test GROUP BY partition_column) s;`

PR #2240, opened 2018-06-28
Foreign Key support from distributed to reference tables

furkansahin:
This PR adds support for foreign keys from distributed to reference tables. Solves #968 

- [x] FK can be created by either `ALTER TABLE` (https://github.com/citusdata/citus/commit/fef07590e5443ca57ead21a6060aae31222f1c3a) or `create_distributed_table` (https://github.com/citusdata/citus/commit/fe84b8cca971a1147ee86aa9de9717173c3eb013)
- [x] Specific DDLs touching to FK is sequentialized. https://github.com/citusdata/citus/commit/72d604a6c30f6632db2a7ba033ad439a0906c314
- [x] All of the necessary changes are applied to handle the DDL operations inside a transaction block.
- [x] A graph is created to detect the cascading operations through FKs in between relations.

The detailed task list can be seen in #968 .

This PR is the master PR for all of the FK related changes and will be merged into master after a complete last review.

Necessary changes on citus-enterprise for tenant isolation and shard_rebalancer are going to be applied separately on enterprise repository.

### DIFF
--- a/src/backend/distributed/commands/create_distributed_table.c
+++ b/src/backend/distributed/commands/create_distributed_table.c
@@ -759,6 +759,11 @@ EnsureRelationCanBeDistributed(Oid relationId, Var *distributionColumn,
 	ErrorIfUnsupportedConstraint(relation, distributionMethod, distributionColumn,
 								 colocationId);
 
+	if (TableReferenced(relationId) || TableReferencing(relationId))
+	{
+		InvalidateForeignKeyGraph();
+	}
+
 	relation_close(relation, NoLock);
 }
 
--- a/src/backend/distributed/commands/create_distributed_table.c
+++ b/src/backend/distributed/commands/create_distributed_table.c
@@ -100,6 +100,7 @@ static bool LocalTableEmpty(Oid tableId);
 static void CopyLocalDataIntoShards(Oid relationId);
 static List * TupleDescColumnNameList(TupleDesc tupleDescriptor);
 static bool RelationUsesIdentityColumns(TupleDesc relationDesc);
+static bool CanUseExclusiveConnections(Oid relationId, bool localTableEmpty);
 
 /* exports for SQL callable functions */
 PG_FUNCTION_INFO_V1(master_create_distributed_table);
@@ -484,26 +485,8 @@ CreateHashDistributedTableShards(Oid relationId, Oid colocatedTableId,
 	 */
 	if (RegularTable(relationId))
 	{
-		if (!localTableEmpty && MultiShardConnectionType == SEQUENTIAL_CONNECTION)
-		{
-			char *relationName = get_rel_name(relationId);
-
-			ereport(ERROR, (errmsg("cannot distribute \"%s\" in sequential mode "
-								   "because it is not empty", relationName),
-							errhint("If you have manually set "
-									"citus.multi_shard_modify_mode to 'sequential', "
-									"try with 'parallel' option. If that is not the "
-									"case, try distributing local tables when they "
-									"are empty.")));
-		}
-		else if (MultiShardConnectionType == SEQUENTIAL_CONNECTION)
-		{
-			useExclusiveConnection = false;
-		}
-		else if (!localTableEmpty || IsTransactionBlock())
-		{
-			useExclusiveConnection = true;
-		}
+		useExclusiveConnection = CanUseExclusiveConnections(relationId,
+															localTableEmpty);
 	}
 
 	if (colocatedTableId != InvalidOid)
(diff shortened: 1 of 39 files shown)


onderkalaci:
Note that this PR consists of already reviewed commits, and can be considered baseline for the upcoming commits.

src/backend/distributed/commands/create_distributed_table.c
```diff
@@ -776,6 +760,11 @@ EnsureRelationCanBeDistributed(Oid relationId, Var *distributionColumn,
 	ErrorIfUnsupportedConstraint(relation, distributionMethod, distributionColumn,
 								 colocationId);
 
+	if (TableReferenced(relationId) || TableReferencing(relationId))
+	{
+		InvalidateForeignKeyGraph();
```
marcocitus: Is this in the right place?  (EnsureRelationCanBeDistributed)  almost looks like a merge conflict

src/backend/distributed/master/master_repair_shards.c
```diff
@@ -379,19 +416,40 @@ CopyShardForeignConstraintCommandList(ShardInterval *shardInterval)
...
 		referencedSchemaName = get_namespace_name(referencedSchemaId);
 		escapedReferencedSchemaName = quote_literal_cstr(referencedSchemaName);
-		referencedShardId = ColocatedShardIdInRelation(referencedRelationId, shardIndex);
+
+		if (PartitionMethod(referencedRelationId) == DISTRIBUTE_BY_NONE)
+		{
+			List *shardList = LoadShardList(referencedRelationId);
+			uint64 *shardIdPointer = (uint64 *) linitial(shardList);
+
+			referencedShardId = (*shardIdPointer);
```
marcocitus: Might be nice to introduce a `GetFirstShardId(Oid relationId)` function or something along those lines.

src/backend/distributed/transaction/relation_access_tracking.c
```diff
@@ -20,14 +20,27 @@
...
+#include "utils/lsyscache.h"
 
 
+/* Config variables managed via guc.c */
+bool EnforceForeignKeyRestrictions = true;
+
 #define PARALLEL_MODE_FLAG_OFFSET 3
+#define PARALLEL_ACCESS_MASK (int) (0 | \
+									(1 << (PLACEMENT_ACCESS_SELECT + \
+										   PARALLEL_MODE_FLAG_OFFSET)) | \
```
marcocitus: we could probably reduce the indentation a bit to improve readability

src/backend/distributed/commands/create_distributed_table.c
onderkalaci: The rationale was that right after we've passed `ErrorIfUnsupportedConstraint ()`, we're sure that the table is going to be distributed. And, the table is locally created. So, if the foreign keys exist with the table, we could send the invalidations.

However, I see your point. I'll move it to right after we create the shards & add the metadata.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -3760,3 +3982,90 @@ ProcessDropTableStmt(DropStmt *dropTableStatement)
...
+	else if (alterTableType == AT_DropColumn || alterTableType == AT_AlterColumnType)
+	{
+		char *affectedColumnName = command->name;
+
+		if (ColumnAppearsInForeignKeyToReferenceTable(affectedColumnName,
+													  relationId))
+		{
+			if (IsTransactionBlock() && alterTableType == AT_AlterColumnType)
+			{
+				SetLocalMultiShardModifyModeToSequential();
```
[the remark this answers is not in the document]
onderkalaci: Yes, this function seems a bit tricky to me as well. I'll try to simplify it or at least comment better.

`ALTER COLUMN` is tricky because any operation on the columns involving the foreign key runs a `SELECT` on the referencing table. 

So, if we do not switch to the sequential mode in a transaction block, we'd end up with the restrictions where `Parallel SELECT on hash` followed by `DDL` on a reference table, which we cannot support. So, by switching to sequential mode, we ensure that the `SELECT` happens on a single connection per node.

And, note that we don't have to call `SetLocalMultiShardModifyModeToSequential();` for all checks here. For example, `ALTER TABLE ADD FKEY` command itself should be executed sequentially, but, we don't really have to set the multishard mode to sequential which could effect the unrelated consecutive operations in the same transaction. Any dangerous operation is in the transaction is already going to be detected and acted accordingly with the `Enforce Restrictions` PR

src/backend/distributed/transaction/relation_access_tracking.c
onderkalaci: I tried that a bit, but `citus_indent` is constantly produced hard to read things 😄   Added a small comment

marcocitus:
should this PR have a DESCRIPTION?

PR #2241, opened 2018-06-28
Foreign Key tests with partitioning and mx

furkansahin:
This PR adds tests for foreign keys from distributed to reference tables for Citus MX and partitioning.


src/test/regress/sql/foreign_key_to_reference_table.sql
```diff
@@ -909,6 +908,55 @@ BEGIN;
...
+-- should fail since partitioning_test_4 has foreign constraint to referenced_table on value_1 column
+INSERT INTO referencing_table VALUES (0, 5);
+INSERT INTO referenced_table VALUES(5,5);
+-- should succeed since both of the foreign constraints are positive
+INSERT INTO referencing_table VALUES (0, 5);
+
+DROP TABLE referenced_table CASCADE;
+DROP TABLE referencing_table;
+
+-- Check if MX works properly
```
onderkalaci: Well, I said that it's OK to add MX tests here. However, now it seemed unusual to have `start_metadata_sync_to_node ()` (and actually missing `stop_metadata_sync_to_node()`).

Could you simply move these tests to a new file and add it do `mx-schedule`?

Sorry about that :/

PR #2243, opened 2018-06-28
Remove sslmode structs and add more helpful description

lithp:
A small tweak to #2190 to help users migrate.

I made the max 32 so existing installations will keep working, but I think the max should probably be 0. Currently we'll just ignore whatever setting they set, probably we want to give them a hard error so they'll be aware of that, instead of silently changing behavior.


lithp:
@sumedhpathak I remember you saying Jason wasn't working today, is there someone else I should assign this to so it can make it in before the code freeze?

PR #2244, opened 2018-06-29
Add tests for 1PC COPY on append and hash-distributed tables

lithp:


metdos:
> I think one general open question for adding the failure testing is the following:
> 
> Do we need to add tests for all possible scenarios for all tests (e.g., #2212) including
> * Replication factor 1 vs 2
> * Parallel vs Sequential mode (if applicable)
> * 1PC vs 2PC
> * Cancel tests for each step as we do for failures
> 
> If we add all the tests, we'd end up with a huge duplication since we'd be exercising the same code-paths over and over and over again. On the other hand, we're going add these tests only once and we'd be able forget about almost all network failure scenarios if we have all those tests.
> 
> What would be your thoughts on this? I'm also curious what @marcocitus would think about that.

Hey @marcocitus, any ideas on this? We will start to write automated tests next week, so if you have some comments, please let us know.

marcocitus:
> we'd be able forget about almost all network failure scenarios
 
At some point adding a lot more failure tests using the current framework will have diminishing returns, while there are other areas that are not yet covered. In particular network timeouts and network failures behind a pgbouncer (e.g. BEGIN results in a `query_wait_timeout` error). 

More on topic: The way 1PC vs. 2PC works in COPY is not different than in other parts of the code. It might make sense to dedicate a test file specifically to commit/abort failures in 1PC and 2PC, while not doing commit-time fault-injection in other tests.

Some code paths with different failure handling logic:
- COPY
- DDL & multi-shard DML with parallel and sequential execution
- Bare DDL (vacuum, create index concurrently)
- Multi-shard SELECT
- Single-shard SELECT
- Single-shard DML outside a transaction block
- Single-shard DML in a transaction block
- Multi-statement transactions that span across multiple shards
- DML on reference table
- Commit/abort/prepare
- Begin
- Savepoints

Ideally, we would test all of them with replication vs. no replication, and test cancellation. I don't think there are other dimensions we need to consider.

metdos:
@lithp, could you rebase this to the master and make ready to merge?

metdos:
Are these tests checked in #2262?

metdos:
Just noting that other tests are blocked on this for Travis, so it would be good if prioritize this over others :)

src/test/regress/expected/failure_1pc_copy_append.out
```diff
@@ -0,0 +1,285 @@
...
+------+-------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+    0 | coordinator | [initial message]
```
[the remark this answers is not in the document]
lithp: > But, travis sometimes fail due to place of + 0 | coordinator | ['Terminate()'].

This is a bug that should be fixed. I guess it's a timing issue. mitmdump should ignore any packets from connections which existed before the last time the buffer was cleared. I can fix it tomorrow!

src/test/regress/expected/failure_1pc_copy_hash.out
```diff
@@ -0,0 +1,406 @@
...
+ key | value 
+-----+-------
+   0 |     0
+   1 |     1
+   2 |     4
+   3 |     9
+(4 rows)
+
+-- the COPY data fails
+SELECT citus.mitmproxy('conn.onCopyData().killall()');
```
onderkalaci: Maybe add one more step
```
SELECT citus.mitmproxy('conn.onCommandComplete(command="COPY 4").kill()');
```

sumedhpathak:
@lithp Could you follow-up on the action-items and close this PR out?

src/test/regress/expected/failure_1pc_copy_append.out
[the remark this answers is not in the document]
lithp: I've removed the packet trace for now.

src/test/regress/expected/failure_1pc_copy_append.out
```diff
@@ -0,0 +1,285 @@
...
+COPY copy_test FROM PROGRAM 'echo 0, 0 && echo 1, 1 && echo 2, 4 && echo 3, 9' WITH CSV;
+WARNING:  server closed the connection unexpectedly
+	This probably means the server terminated abnormally
+	before or while processing the request.
+CONTEXT:  while executing command on localhost:57640
+WARNING:  connection not open
+CONTEXT:  while executing command on localhost:57640
+ERROR:  failure on connection marked as essential: localhost:57640
+SELECT * FROM pg_dist_shard s, pg_dist_shard_placement p
+  WHERE (s.shardid = p.shardid) AND s.logicalrelid = 'copy_test'::regclass;
```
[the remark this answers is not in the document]
lithp: Good point, pushed a new commit where we always sort by `placementid`.

src/test/regress/expected/failure_1pc_copy_hash.out
lithp: I don't see a lot of value in this test because, from the coordinator's POV, it looks exactly the same as the `conn.onCopyData().killall()` test, but I've just pushed a commit which adds it.

lithp:
@onderkalaci I've rebased on master and addressed, I think, all your comments.

src/test/regress/expected/failure_1pc_copy_hash.out
```diff
@@ -0,0 +1,384 @@
...
+) ORDER BY shardid, placementid;
+ shardid | shardstate | shardlength | nodename  | nodeport | placementid 
+---------+------------+-------------+-----------+----------+-------------
+  100400 |          3 |           0 | localhost |    57637 |         100
+  100400 |          1 |           0 | localhost |    57640 |         101
+(2 rows)
+
+-- ==== okay, run some tests where there's only one active shard ====
+COPY copy_test FROM PROGRAM 'echo 0, 0 && echo 1, 1 && echo 2, 4 && echo 3, 9' WITH CSV;
+SELECT * FROM copy_test;
```
onderkalaci: Minor: Shall we add `ORDER BY`? Though the data comes from a single shard, there is a possibility of ending up with a different order.

src/test/regress/expected/failure_1pc_copy_append.out
```diff
@@ -0,0 +1,265 @@
...
+(1 row)
+
+COPY copy_test FROM PROGRAM 'echo 0, 0 && echo 1, 1 && echo 2, 4 && echo 3, 9' WITH CSV;
+WARNING:  server closed the connection unexpectedly
+	This probably means the server terminated abnormally
+	before or while processing the request.
+CONTEXT:  while executing command on localhost:57640
+WARNING:  connection not open
+CONTEXT:  while executing command on localhost:57640
+ERROR:  failure on connection marked as essential: localhost:57640
```
onderkalaci: Pretty minor: this test took some time for me to digest. Since we've not printed placements just after the successful `COPY` operation 2 lines before, I thought `oh, we're erroring out but also creating the placements successfully, what's happening` 😄 

So, could be useful to print the placements just before starting the test as well.

lithp: good point, added!

src/test/regress/expected/failure_1pc_copy_hash.out
lithp: added!

PR #2245, opened 2018-06-29
Adds locking reference table utils when a shard has a foreign key to it

furkansahin:
This PR is added for rebalancer operations on the enterprise but most of the code path should be in the community as well. Check https://github.com/citusdata/citus-enterprise/pull/253/ for details.

### DIFF
--- a/src/backend/distributed/master/master_repair_shards.c
+++ b/src/backend/distributed/master/master_repair_shards.c
@@ -189,6 +189,16 @@ RepairShardPlacement(int64 shardId, char *sourceNodeName, int32 sourceNodePort,
 	}
 
 	/*
+	 * We take a lock on the referenced table if there is a foreign constraint
+	 * during the copy procedure. If we do not block DMLs on the referenced
+	 * table, we cannot avoid the inconsistency between the two copies of the
+	 * data. Currently, we do not support replication factor > 1 on the tables
+	 * with foreign constraints, so this command will fail for this case anyway.
+	 * However, it is taken as a precaution in case we support it one day.
+	 */
+	LockReferencedReferenceShardDistributionMetadata(shardId, ExclusiveLock);
+
+	/*
 	 * We plan to move the placement to the healthy state, so we need to grab a shard
 	 * metadata lock (in exclusive mode).
 	 */
@@ -393,7 +403,6 @@ CopyShardForeignConstraintCommandList(ShardInterval *shardInterval)
 
 	return copyShardForeignConstraintCommandList;
 }
-
 
 /*
  * ConstuctQualifiedShardName creates the fully qualified name string of the
--- a/src/backend/distributed/master/master_repair_shards.c
+++ b/src/backend/distributed/master/master_repair_shards.c
@@ -348,8 +348,31 @@ CopyShardCommandList(ShardInterval *shardInterval,
 List *
 CopyShardForeignConstraintCommandList(ShardInterval *shardInterval)
 {
-	List *copyShardForeignConstraintCommandList = NIL;
+	List *colocatedShardForeignConstraintCommandList = NIL;
+	List *referenceTableForeignConstraintList = NIL;
 
+	CopyShardForeignConstraintCommandListGrouped(shardInterval,
+												 &
+												 colocatedShardForeignConstraintCommandList,
+												 &referenceTableForeignConstraintList);
+
+	return list_concat(colocatedShardForeignConstraintCommandList,
+					   referenceTableForeignConstraintList);
+}
+
+
+/*
+ * CopyShardForeignConstraintCommandListGrouped generates command lists
+ * to create foreign constraints existing in source shard after copying it to other
+ * node in separate groups for foreign constraints in between hash distributed tables
+ * and from a hash distributed to reference tables.
+ */
+void
+CopyShardForeignConstraintCommandListGrouped(ShardInterval *shardInterval,
+											 List **
+											 colocatedShardForeignConstraintCommandList,
+											 List **referenceTableForeignConstraintList)
+{
 	Oid schemaId = get_rel_namespace(shardInterval->relationId);
 	char *schemaName = get_namespace_name(schemaId);
 	char *escapedSchemaName = quote_literal_cstr(schemaName);
@@ -364,6 +387,9 @@ CopyShardForeignConstraintCommandList(ShardInterval *shardInterval)
 		shardIndex = ShardIndex(shardInterval);
 	}
 
+	*colocatedShardForeignConstraintCommandList = NIL;
+	*referenceTableForeignConstraintList = NIL;
+
 	foreach(commandCell, commandList)
 	{
 		char *command = (char *) lfirst(commandCell);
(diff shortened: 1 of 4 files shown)


src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -649,3 +649,63 @@ HeapTupleOfForeignConstraintIncludesColumn(HeapTuple heapTuple, Oid relationId,
 
 	return false;
 }
+
+
+/*
+ * This function will be removed later
+ */
+List *
+GetReferencedReferenceTableList(Oid relationId)
```
onderkalaci: you can rebase your changes to `foreign_key_relations ` to get rid of this function and rely on 
```C
		List *referencedRelations = cacheEntry->referencedRelationsViaForeignKey;

```

src/backend/distributed/master/master_repair_shards.c
```diff
@@ -488,3 +511,44 @@ WorkerApplyShardDDLCommandList(List *ddlCommandList, int64 shardId)
 
 	return applyDdlCommandList;
 }
+
+
+/*
+ * LockReferencedTablesIfExist takes exclusive lock on the reference tables
+ * which has a foreign key from the given relation. This function is used
+ * for shard_rebalancer and tenant isolation to avoid modifications on the
```
onderkalaci: `shard_rebalancer ` and `tenant isolation` is not defined in this repo. So, we should not explicitly mention those here. And, I think we should have these comments at the point where we call this function.

Also, the name of the function is too generic. We should probably call it something like : `LockReferencedShardDistributionMetadata `.

Finally, for the future, it'd probably be less painful to maintain this function if we pass the `LockMode` as a parameter, instead of hard-coding `ExclusiveLock `

src/backend/distributed/master/master_repair_shards.c
```diff
@@ -488,3 +511,44 @@ WorkerApplyShardDDLCommandList(List *ddlCommandList, int64 shardId)
...
+
+	foreach(referencedTableCell, referencedTableList)
+	{
+		Oid referencedTableId = lfirst_oid(referencedTableCell);
+		List *shardList = NIL;
+		uint64 *shardIdPointer = NULL;
+		Oid referencedShardId = InvalidOid;
+
+		/* check that user has owner rights in referenced table */
+		EnsureTableOwner(referencedTableId);
```
onderkalaci: Future note: We should skip `non-reference` tables since the `foreign_key_relations` does not return only reference tables. And, we should add this information to the function comment.

src/backend/distributed/master/master_repair_shards.c
```diff
@@ -379,7 +390,19 @@ CopyShardForeignConstraintCommandList(ShardInterval *shardInterval)
 		referencedSchemaId = get_rel_namespace(referencedRelationId);
 		referencedSchemaName = get_namespace_name(referencedSchemaId);
 		escapedReferencedSchemaName = quote_literal_cstr(referencedSchemaName);
-		referencedShardId = ColocatedShardIdInRelation(referencedRelationId, shardIndex);
+
```
onderkalaci: Do we have any other places where we call `ColocatedShardIdInRelation ()`? And, are they dangerous as here in the presence of foreign keys to reference tables?

furkansahin: Other than the places we took care of, as I can see, there are some code paths related to Partitioning. Since Postgres does not support foreign keys on partitioned tables, we are not in trouble, yet. Should we dig in and add these checks in case one day we hit such a problem?

src/backend/distributed/master/master_repair_shards.c
```diff
@@ -488,3 +550,44 @@ WorkerApplyShardDDLCommandList(List *ddlCommandList, int64 shardId)
...
+		/*
+		 * Block concurrent reads on the referenced table so that there
+		 * is no inconsistency between before/after copy.
+		 */
+		shardList = LoadShardList(referencedTableId);
+		shardIdPointer = (uint64 *) linitial(shardList);
+		referencedShardId = (*shardIdPointer);
+
+		LockShardDistributionMetadata(referencedShardId, lock);
```
marcocitus: even if it's not yet very useful, it'd be nice to also do `LockShardListMetadataOnWorkers` here to (in the future) block writes from nodes with metadata

src/backend/distributed/ddl/foreign_constraint.c
```diff
@@ -649,3 +649,63 @@ HeapTupleOfForeignConstraintIncludesColumn(HeapTuple heapTuple, Oid relationId,
...
+
+	/*
+	 * Set search_path to NIL so that all objects outside of pg_catalog will be
+	 * schema-prefixed. pg_catalog will be added automatically when we call
+	 * PushOverrideSearchPath(), since we set addCatalog to true;
+	 */
+	OverrideSearchPath *overridePath = GetOverrideSearchPath(CurrentMemoryContext);
+	overridePath->schemas = NIL;
+	overridePath->addCatalog = true;
+	PushOverrideSearchPath(overridePath);
```
marcocitus: why are we changing the search path here? there's no deparsing in this function afaik

src/backend/distributed/utils/resource_lock.c
```diff
@@ -164,6 +166,58 @@ LockShardDistributionMetadata(int64 shardId, LOCKMODE lockMode)
...
+		if (PartitionMethod(relationId) != DISTRIBUTE_BY_NONE)
+		{
+			continue;
+		}
+
+		currentShardList = LoadShardList(relationId);
+		shardList = lappend(shardList, linitial(currentShardList));
+	}
+
+	shardList = SortList(shardList, CompareOids);
```
onderkalaci: `CompareOids` might be tricky. We should probably use `LoadShardIntervalList()` and `CompareShardIntervalsById`.

onderkalaci: But then don't forget to rename the function as `GetSortedReferenceShardIntervals` and fix the loop `foreach(shardCell, shardList)`.

Or, you could write a new function `CompareShardIds`, and use it here. Pick one approach

src/backend/distributed/master/master_repair_shards.c
```diff
@@ -188,6 +188,17 @@ RepairShardPlacement(int64 shardId, char *sourceNodeName, int32 sourceNodePort,
...
 
+	/*
+	 * We take a lock on the referenced table if there is a foreign constraint
+	 * during the copy procedure. If we do not block DMLs on the referenced
+	 * table, we cannot avoid the inconsistency between the two copies of the
+	 * data. Currently, we do not support replication factor > 1 on the tables
+	 * with foreign constraints, so this command will fail for this case anyway.
+	 * However, it is taken as a precaution in case we support it one day.
+	 */
+	LockReferencedReferenceShardDistributionMetadata(shardInterval->shardId,
```
onderkalaci: Very minor: Maybe simply use `shardId ` which is already assigned above

PR #2247, opened 2018-07-02
Add pg_stat_statements to shared_preload_libraries if installed

mtuncer:
We needed pg_stat_statements extension for testing. This change first checks the availability of extension, if it is then adds it to shared_preload_libraries next to citus.


marcocitus:
What if we check whether pg_stat_statements is installed? i.e. if `$(pg_config --sharedir)/extensions/pg_stat_statements.control` exists we automatically add `pg_stat_statements` to `shared_preload_libraries`?

Given that most production clusters (read: Citus Cloud) have pg_stat_statements enabled it might be good to always run our tests with it.

mtuncer:
After some considerations I decided to take suggested approach. PR will be updated accordingly

PR #2248, opened 2018-07-02
Strip implicit coercions when determining partition key value of an INSERT

mtuncer:
Fixed a bug where we fail to determine partition key value when a partition key value is bigint.
We should have stripped the implicit coercions before checking if the expression is a const value.

### DIFF
--- a/src/backend/distributed/planner/multi_router_planner.c
+++ b/src/backend/distributed/planner/multi_router_planner.c
@@ -2661,6 +2661,7 @@ ExtractInsertPartitionKeyValue(Query *query)
 	Var *partitionColumn = NULL;
 	TargetEntry *targetEntry = NULL;
 	Const *singlePartitionValueConst = NULL;
+	Node *targetExpression = NULL;
 
 	char partitionMethod = PartitionMethod(distributedTableId);
 	if (partitionMethod == DISTRIBUTE_BY_NONE)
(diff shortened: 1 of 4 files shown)


mtuncer:
description field is intentionally not added.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -2661,6 +2661,7 @@ ExtractInsertPartitionKeyValue(Query *query)
 	Var *partitionColumn = NULL;
 	TargetEntry *targetEntry = NULL;
 	Const *singlePartitionValueConst = NULL;
+	Expr *targetExpression = NULL;
```
marcocitus: minor: if you make this a `Node *` you can avoid a cast

src/test/regress/expected/multi_router_planner.out
```diff
@@ -2308,6 +2310,7 @@ DETAIL:  distribution column value: 1
 INSERT INTO articles_hash VALUES (51,  1, 'amateus', 1814);
 DEBUG:  Creating router plan
 DEBUG:  Plan is router executable
+DETAIL:  distribution column value: 1
```
marcocitus: might want to add a multi-row insert with the same partition column value in each row to get code coverage for that

mtuncer: Added, but it does not get printed here. We set the partition column on the job only after we realize that multi-column insert is a router executable query.

src/backend/distributed/planner/multi_router_planner.c
mtuncer: done

PR #2249, opened 2018-07-03
Update citus_stat_statements view and regression tests

mtuncer:
UDF citus.citus_executor_name(int) added to convert from executor type to executor name


src/backend/distributed/citus--7.5-4--7.5-5.sql
```diff
@@ -0,0 +1,34 @@
...
+END;
+$function$;
+
+DROP VIEW pg_catalog.citus_stat_statements;
+
+CREATE VIEW citus.citus_stat_statements AS
+SELECT
+  queryid,
+  userid,
+  dbid oid,
```
marcocitus: This renames the column to oid?

mtuncer: unintentional, removed

src/backend/distributed/citus--7.5-4--7.5-5.sql
```diff
@@ -0,0 +1,34 @@
+/* citus--7.5-4--7.5-5 */
+CREATE FUNCTION citus.citus_executor_name(executor_type int)
```
marcocitus: We may need to be careful about this function being in the citus dchema, since regular users don't have access to it.

mtuncer: pg_catalog looked more restrictive. What do you suggest ?

marcocitus: pg_catalog seems fine, all users have usage access to pg_catalog

mtuncer: updated

PR #2254, opened 2018-07-05
create placeholder for policy ddl

thanodnl:
DESCRIPTION: Adds infrastructure to implement DDL commands for policies in the future, warn about being unsupported on distributed tables.

There is a small side effect for people that use policies on distributed tables now as it signals when you  execute `CREATE POLICY` that the operation is not supported on distributed tables. Warns in the same way when distributing tables that already have policies defined that those are unsupported and need to be removed from a table before distributing it.

We need to change our [blog post about RLS](https://www.citusdata.com/blog/2018/04/04/raw-sql-access-with-row-level-security/) accordingly by either add more `SET citus.enable_ddl_propagation to off;` statements around the DDL commands. Or better remove them and reword the blogpost to indicate it is a supported feature in Citus Cloud and Enterprise as `Fine-grained access controls`. The current solution provided by the blogpost has problems around shard movement, repairs and splits as the newly createsd shards will not have the policies set on them

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,7 @@ OBJS = src/backend/distributed/shared_library_init.o \
      src/backend/distributed/connection/placement_connection.o \
      src/backend/distributed/connection/remote_commands.o \
      src/backend/distributed/ddl/foreign_constraint.o \
+     src/backend/distributed/ddl/policy.o \
      src/backend/distributed/executor/citus_custom_scan.o \
      src/backend/distributed/executor/insert_select_executor.o \
      src/backend/distributed/executor/intermediate_results.o \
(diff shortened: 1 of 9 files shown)


mkurz:
@thanodnl @marcocitus 
> We need to change our blog post about RLS accordingly by either add more SET citus.enable_ddl_propagation to off; statements around the DDL commands. Or better remove them and reword the blogpost to indicate it is a supported feature in Citus Cloud and Enterprise as Fine-grained access controls. The current solution provided by the blogpost has problems around shard movement, repairs and splits as the newly createsd shards will not have the policies set on them

That blog post is still out of date.

PR #2255, opened 2018-07-05
Expand count distinct support

mtuncer:
We can now support more complex count distinct operations by
pulling necessary columns to coordinator and evalutating the
aggreage at coordinator.

It supports broad range of expression with the restriction that
the expression must contain a column.


src/test/regress/expected/multi_complex_count_distinct_0.out
```diff
@@ -1,1049 +0,0 @@
...
-(1 row)
-
-SELECT master_create_worker_shards('lineitem_hash', 8, 1);
- master_create_worker_shards 
------------------------------
- 
-(1 row)
-
-\copy lineitem_hash FROM '/Users/mtuncer/dev/citus/features/remove_poll/src/test/regress/data/lineitem.1.data' with delimiter '|'
-\copy lineitem_hash FROM '/Users/mtuncer/dev/citus/features/remove_poll/src/test/regress/data/lineitem.2.data' with delimiter '|'
```
marcocitus: woah, how did this get in there?

src/backend/distributed/planner/extended_op_node_utils.c
```diff
@@ -167,6 +168,9 @@ HasNonPartitionColumnDistinctAgg(List *targetEntryList, Node *havingQual,
 		List *varList = NIL;
 		ListCell *varCell = NULL;
 		bool isPartitionColumn = false;
+		TargetEntry *firstTargetEntry = NULL;
+		Node *firstTargetExprNode = NULL;
+
```
marcocitus: extra newline

src/backend/distributed/planner/extended_op_node_utils.c
```diff
@@ -180,6 +184,22 @@ HasNonPartitionColumnDistinctAgg(List *targetEntryList, Node *havingQual,
...
+		 * evaluated at coordinator level.
+		 */
+		if (list_length(targetAgg->args) > 1 || list_length(targetAgg->aggdistinct) > 1)
+		{
+			return true;
+		}
+
+		firstTargetEntry = linitial_node(TargetEntry, targetAgg->args);
+		firstTargetExprNode = strip_implicit_coercions((Node *) firstTargetEntry->expr);
+		if (!IsA(firstTargetExprNode, Var))
```
marcocitus: The description of this function should be updated to reflect the change.

I think it might also be worth adding a comment that we typically cannot push down expressions because we cannot easily tell whether an expression maps two distinct partition column values to the same value.

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -3513,6 +3516,11 @@ TablePartitioningSupportsDistinct(List *tableNodeList, MultiExtendedOp *opNode,
 			Var *tablePartitionColumn = tableNode->partitionColumn;
 			bool groupedByPartitionColumn = false;
 
+			if (aggregateType == AGGREGATE_COUNT)
+			{
+				tableDistinctSupported = true;
```
marcocitus: At this point, is there any reason not to just pull expressions like sum(distinct) to master?

src/backend/distributed/planner/extended_op_node_utils.c
mtuncer: ok

src/backend/distributed/planner/multi_logical_optimizer.c
mtuncer: Not really, we should be able to pull any distinct expression to coordinator.

There was an issue with pulling up a non-column distinct like count(distinct 1) type expression where I spent most of my time working on this PR. We should also address that with more generalized aggregate distinct work.

src/test/regress/expected/multi_complex_count_distinct_0.out
mtuncer: don't want to tell 😞

furkansahin:
@mtuncer DESCRIPTION?

mtuncer:
@furkansahin added

PR #2257, opened 2018-07-05
Add a GUC to disable opening transaction blocks on workers for SELECT-only xacts

marcocitus:
DESCRIPTION: Adds a citus.select_opens_transaction_block GUC to defer opening transaction blocks on workers

When you run a multi-shard SELECT in a transaction block, we also run the tasks on the workers in transaction blocks. While technically sound, it can dramatically increase idle in transaction time because the transactions won't get committed until all tasks are done and commits get sent out. In a set-up where there is a pgbouncer in between coordinator and worker, this means that a session that could otherwise be used for pending commands will be idle. When there are many such sessions the system will be heavily underutilised (but overloaded from the application perspective).

While the obvious solution is to not run the SELECT commands in transaction blocks when idle in transaction time is concern, there are cases when this is not a realistic option. For example, when the ORM automatically runs queries in a transaction block. Some ORMs always use prepared statements that they immediately execute in the same transaction. For these prepared statements to work when connecting to the coordinator through pgbouncer, they need to be in a transaction block.

This PR adds a GUC that gives the user the option to opt out of opening transaction blocks on the workers for SELECTs, such that the tasks commit immediately. If a write or other operation that requires a distributed transaction (e.g. a CTE) occurs prior to the SELECT it will be executed in a transaction block.

### DIFF
--- a/src/backend/distributed/executor/multi_real_time_executor.c
+++ b/src/backend/distributed/executor/multi_real_time_executor.c
@@ -42,6 +42,13 @@
 #include "utils/timestamp.h"
 
 
+/*
+ * GUC that determines whether a SELECT in a transaction block should also run in
+ * a transaction block on the worker even if no writes have occurred yet.
+ */
+bool SelectOpensTransactionBlock;
+
+
 /* Local functions forward declarations */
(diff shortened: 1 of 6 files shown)


jasonmp85:
:shipit:

PR #2263, opened 2018-07-09
Add "direct" node columns to metadata catalogs

jasonmp85:
When Citus is used in an advanced setup, e.g. with pgbouncer between the coordinator and workers for most queries, it is still the case that certain operations (e.g. logical shard rebalancing) should use "direct" connections.

Because (when both pgbouncer and direct connections are in use) non-direct connections are the "common case", it makes sense to leave the existing nodename and nodeport fields in use for typical queries and to add special-case directname/directport columns for use by operations requiring direct connections. The alternative (changing nodename/port semantics to mean "direct" and adding e.g. poolname/poolport columns to specify the pgbouncer pool) is more complex behaviorally and more surprising to those who upgrade and expect nodename to continue being the name used for typical operations.

As a typical installation may not use pgbouncer, I've allowed these columns to be nullable.

Additionally, pg_dist_authinfo gets an extra column to specify direct-connection–specific authentication information, to extend to this new functionality the same authentication improvements we previously added.

### DIFF
--- a/citus.control
+++ b/citus.control
@@ -1,6 +1,6 @@
 # Citus extension
 comment = 'Citus distributed database'
-default_version = '7.5-5'
+default_version = '7.5-6'
 module_pathname = '$libdir/citus'
 relocatable = false
 schema = pg_catalog
--- a/src/backend/distributed/Makefile
+++ b/src/backend/distributed/Makefile
@@ -16,7 +16,7 @@ EXTVERSIONS = 5.0 5.0-1 5.0-2  \
 	7.2-1 7.2-2 7.2-3 \
 	7.3-1 7.3-2 7.3-3 \
 	7.4-1 7.4-2 7.4-3 \
-	7.5-1 7.5-2 7.5-3 7.5-4 7.5-5
+	7.5-1 7.5-2 7.5-3 7.5-4 7.5-5 7.5-6
 
 # All citus--*.sql files in the source directory
 DATA = $(patsubst $(citus_abs_srcdir)/%.sql,%.sql,$(wildcard $(citus_abs_srcdir)/$(EXTENSION)--*--*.sql))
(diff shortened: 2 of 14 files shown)


jasonmp85:
@marcocitus hey, @sumedhpathak said to assign this to you; it just adds two columns to `pg_dist_node` to specify a "direct connect" override for each node (to be used in e.g. logical rebalancing). In addition a direct authinfo column is added to `pg_dist_authinfo`.

jasonmp85:
(FWIW this is _just_ the metadata changes, no functionality; that's another branch that'll be in Enterprise, but post-release AFAIK).

jasonmp85:
@marcocitus I have no clue why Travis is failing; it has something to do with the latest mitm failure testing but it's not outputting any diff output and the expected/results match exactly. If you can figure something out feel free to update this PR.

marcocitus:
> Because (when both pgbouncer and direct connections are in use) non-direct connections are the "common case", it makes sense to leave the existing nodename and nodeport fields in use for typical queries and to add special-case directname/directport columns for use by operations requiring direct connections. 

That doesn't make much sense to me. I guess you'd have 127.0.0.1, 127.0.0.2 etc. as the nodename. You'd than have to enforce that throughout the cluster, e.g. you could not have worker nodes without pgbouncer or with a different pgbouncer layout. Plus error messages and whatnot would get terribly confusing.

marcocitus:
I think we would want something like `pg_dist_hostinfo` or `pg_dist_poolinfo` that maps a `nodeid` to a connection string that can contain `host`, `port` and maybe `dbname`. We can then configure that on a node-by-node basis (e.g. it may only be populated on the coordinator).

jasonmp85:
> I think we would want something like `pg_dist_hostinfo` or `pg_dist_poolinfo`

So you're saying a whole new table, not just new columns?

marcocitus:
Yes, pg_dist_node gets synced to workers, which may complicate things. A separate table analogous to pg_dist_authinfo seems like the simplest and most flexible solution.

jasonmp85:
Closing in favor of #2268

PR #2264, opened 2018-07-09
Create pull_request_template.md with DESCRIPTION line

marcocitus:
I figured it's easier to remove than to remember the DESCRIPTION line. This change adds "DESCRIPTION:" to the PR description by default.


.github/pull_request_template.md
```diff
@@ -0,0 +1 @@
+DESCRIPTION:
```
velioglu: We might also add a warning about the length of the description, which is 78 character.

PR #2266, opened 2018-07-09
Make citus_stat_statements_reset() super user function

mtuncer:
This PR revokes all rights on citus_stat_statements_reset() UDF from public.

### DIFF
--- a/citus.control
+++ b/citus.control
@@ -1,6 +1,6 @@
 # Citus extension
 comment = 'Citus distributed database'
-default_version = '7.5-5'
+default_version = '7.5-6'
 module_pathname = '$libdir/citus'
 relocatable = false
 schema = pg_catalog
--- a/src/backend/distributed/Makefile
+++ b/src/backend/distributed/Makefile
@@ -16,7 +16,7 @@ EXTVERSIONS = 5.0 5.0-1 5.0-2  \
 	7.2-1 7.2-2 7.2-3 \
 	7.3-1 7.3-2 7.3-3 \
 	7.4-1 7.4-2 7.4-3 \
-	7.5-1 7.5-2 7.5-3 7.5-4 7.5-5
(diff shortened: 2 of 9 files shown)


mtuncer:
does not need description as this is a continuation of landlord work

marcocitus:
Should we do the same for `citus_stat_statements`?

mtuncer:
Do you mean citus_stat_staments view or UDF ?

pg_stat_statements view is available for public. IF we are to limit citus_stat_staments view, we should also limit the UDF.

marcocitus:
> I was thinking we can revoke rights from citus_stat_statements() and citus_query_stats() while granting it for the view. pg_stat_statements sort of handles its own security and after the join with citus_stat_statements() you can only see your own queries since pg_stat_statements hides the queryIds, but you can see all queries in citus_query_stats().

ok, nvm, seems you cannot grant select permission on the view without the ability to execute the function

marcocitus:
I think making the function callable and the view readable only by `pg_read_all_settings` makes sense, though doing it in the extension script creates a hard dependency on `pg_stat_statements`, which is not something we want.

I guess the current PR might be an ok starting point and users will need to adapt to their needs, e.g. Cloud will want to grant execute on `citus_stat_statements_reset()` to `pg_read_all_stats`.

mtuncer:
I am not for this 👎 
role `pg_read_all_stats` is available for PG10+, it is not available in PG9.6 which we still support.

mtuncer:
also current patch to pg_stat_statements has following
```
+-- Don't want this to be available to non-superusers.
+REVOKE EXECUTE ON FUNCTION pg_stat_statements_reset() FROM pg_read_all_stats;
```

my take:
- leave citus_stat_statements view (and related functions) as callable from public
- make citus_state_statements_reset() super user function

PR #2268, opened 2018-07-10
Add pg_dist_poolinfo table

jasonmp85:
For storing nodes' pool host/port overrides.


jasonmp85:
Chatted with @fdr a bit about this. He indicated having separate authentication for direct/pool might be more confusion than it's worth, so I went with this approach: simply making a 1—1 table adding `poolinfo` strings for each node (anticipated to hold host and port, or possibly dbname).

No significant code changes yet, but I've wired up triggers and constraints as needed.

src/backend/distributed/citus--7.5-5--7.5-6.sql
```diff
@@ -0,0 +1,35 @@
...
+CREATE FUNCTION pg_catalog.poolinfo_valid(text)
+	RETURNS boolean
+    LANGUAGE C STRICT
+    AS 'MODULE_PATHNAME', $$poolinfo_valid$$;
+COMMENT ON FUNCTION poolinfo_valid(text) IS 'returns whether a poolinfo is valid';
+
+CREATE TABLE citus.pg_dist_poolinfo (
+    nodeid integer PRIMARY KEY
+                   REFERENCES pg_dist_node(nodeid)
+                              ON DELETE CASCADE,
```
marcocitus: Not relevant to this PR, but just to make note of it: There's a good chance that `master_remove_node` does not invoke triggers.

src/backend/distributed/citus--7.5-5--7.5-6.sql
```diff
@@ -0,0 +1,35 @@
+/* citus--7.5-1--7.5-2 */
+SET search_path = 'pg_catalog';
+
+CREATE FUNCTION pg_catalog.poolinfo_valid(text)
+	RETURNS boolean
+    LANGUAGE C STRICT
+    AS 'MODULE_PATHNAME', $$poolinfo_valid$$;
+COMMENT ON FUNCTION poolinfo_valid(text) IS 'returns whether a poolinfo is valid';
```
marcocitus: Super minor, feel free to ignore: it's a little odd that we're schema-qualifying the `CREATE FUNCTION`, but not in the `COMMENT ON FUNCTION`

jasonmp85: This is largely copy-pasted from other upgrade scripts so not gonna bother to change (to avoid the Travis round-trip).

PR #2272, opened 2018-07-10
Truncate failure tests are added

furkansahin:
Failure test for `TRUNCATE` command is added

### DIFF
--- a/src/test/regress/expected/failure_truncate.out
+++ b/src/test/regress/expected/failure_truncate.out
@@ -1,0 +1,1286 @@
+-- 
+-- Test TRUNCATE command failures 
+-- 
+CREATE SCHEMA truncate_failure;
+SET search_path TO 'truncate_failure';
+SET citus.next_shard_id TO 120000;
+-- we don't want to see the prepared transaction numbers in the warnings
+SET client_min_messages TO ERROR;
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+-- we'll start with replication factor 1, 1PC and parallel mode
+SET citus.multi_shard_commit_protocol TO '1pc';
+SET citus.shard_count = 4;
+SET citus.shard_replication_factor = 1;
+CREATE TABLE test_table (key int, value int);
+SELECT create_distributed_table('test_table', 'key');
+ create_distributed_table 
+--------------------------
+ 
+(1 row)
+
+INSERT INTO test_table SELECT x,x FROM generate_series(1,20) as f(x);
+CREATE VIEW unhealthy_shard_count AS 
+  SELECT count(*) 
+  FROM pg_dist_shard_placement pdsp 
+  JOIN 
+  pg_dist_shard pds 
+  ON pdsp.shardid=pds.shardid 
+  WHERE logicalrelid='truncate_failure.test_table'::regclass AND shardstate != 1;
+-- in the first test, kill just in the first 
+-- response we get from the worker
+SELECT citus.mitmproxy('conn.onAuthenticationOk().kill()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+TRUNCATE test_table;
+ERROR:  connection error: localhost:57640
+DETAIL:  server closed the connection unexpectedly
+	This probably means the server terminated abnormally
+	before or while processing the request.
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT * FROM unhealthy_shard_count;
+ count 
+-------
+     0
+(1 row)
+
+SELECT count(*) FROM test_table;
+ count 
+-------
+    20
+(1 row)
+
+-- cancel just in the first 
+-- response we get from the worker
+SELECT citus.mitmproxy('conn.onAuthenticationOk().cancel(' ||  pg_backend_pid() || ')');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+TRUNCATE test_table;
+ERROR:  canceling statement due to user request
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT * FROM unhealthy_shard_count;
+ count 
+-------
+     0
+(1 row)
+
+SELECT count(*) FROM test_table;
+ count 
+-------
+    20
+(1 row)
+
+-- kill as soon as the coordinator sends begin
+SELECT citus.mitmproxy('conn.onQuery(query="^BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED").kill()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+TRUNCATE test_table;
+ERROR:  failure on connection marked as essential: localhost:57640
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT * FROM unhealthy_shard_count;
+ count 
+-------
+     0
+(1 row)
+
+SELECT count(*) FROM test_table;
+ count 
+-------
+    20
+(1 row)
+
+-- cancel as soon as the coordinator sends begin
+SELECT citus.mitmproxy('conn.onQuery(query="^BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED").cancel(' ||  pg_backend_pid() || ')');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+TRUNCATE test_table;
+ERROR:  canceling statement due to user request
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT * FROM unhealthy_shard_count;
+ count 
+-------
+     0
+(1 row)
+
+SELECT count(*) FROM test_table;
+ count 
+-------
+    20
+(1 row)
+
+-- kill as soon as the coordinator sends TRUNCATE TABLE command
+SELECT citus.mitmproxy('conn.onQuery(query="TRUNCATE TABLE truncate_failure.test_table").kill()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+TRUNCATE test_table;
+ERROR:  server closed the connection unexpectedly
+	This probably means the server terminated abnormally
+	before or while processing the request.
+CONTEXT:  while executing command on localhost:57640
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT * FROM unhealthy_shard_count;
+ count 
+-------
+     0
+(1 row)
+
+SELECT count(*) FROM test_table;
+ count 
+-------
+    20
+(1 row)
+
+-- cancel as soon as the coordinator sends TRUNCATE TABLE command
+SELECT citus.mitmproxy('conn.onQuery(query="TRUNCATE TABLE truncate_failure.test_table").cancel(' ||  pg_backend_pid() || ')');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+TRUNCATE test_table;
+ERROR:  canceling statement due to user request
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT * FROM unhealthy_shard_count;
+ count 
+-------
+     0
+(1 row)
+
+SELECT count(*) FROM test_table;
+ count 
+-------
+    20
+(1 row)
+
+-- kill as soon as the coordinator sends COMMIT
+-- One shard should not get truncated but the other should
+-- since it is sent from another connection. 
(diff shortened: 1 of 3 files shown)


src/test/regress/expected/failure_truncate.out
```diff
@@ -0,0 +1,1094 @@
...
+(1 row)
+
+TRUNCATE test_table;
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT recover_prepared_transactions();
```
onderkalaci: Minor: maybe briefly mention that this is the expected behaviour since some of the placements have been committed and some not. So, we have some pending prepared transactions.

(I think if we had client_min_messages = `WARNING`, we'd have seen the logs but then the prepared tx number changes, that's why we don't want?)

src/test/regress/expected/failure_truncate.out
```diff
@@ -0,0 +1,1096 @@
...
+DETAIL:  server closed the connection unexpectedly
+	This probably means the server terminated abnormally
+	before or while processing the request.
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT * FROM pg_dist_shard_placement ORDER BY 1, 5;
```
onderkalaci: Hmm, there is a likelyhood of `placementId`s changes in the tests. Could we maybe do something `SELECT * FROM pg_dist_shard_placement  WHERE shardstate != 1 ORDER BY 1,5` for all the tests.

And, although we're currently running the tests sequentially in the travis, we've some thoughts on running them in parallel. So, shall we join the `pg_dist_shard_placement ` with `pg_dist_shard` and get the placements that only belong to the `test_table`?

src/test/regress/expected/failure_truncate.out
```diff
@@ -0,0 +1,1096 @@
...
+(1 row)
+
+SELECT count(*) FROM test_table;
+ count 
+-------
+     0
+(1 row)
+
+INSERT INTO test_table SELECT x,x FROM generate_series(1,20) as f(x);
+-- kill as soon as the coordinator sends COMMIT
```
onderkalaci: same here, lets very briefly mention that this is the expected behaviour?

furkansahin:
Hey @onderkalaci , thanks for the review! 
I have made the changes you wanted but I couldn't understand what you want for 
```
(b) Fail on the second PREPARE which is actually the command that CASCADEs to the distributed table.
```

As I can see from the `dump_network_traffic` function, we do not send 2 `PREPARE` commands. We do the both truncate operations in the same transaction and send only 1 `prepare`. 

If you can check [this line](https://github.com/citusdata/citus/pull/2272/files#diff-eba30cd6ebc937c44a5d735e0ee7f5b5R402), I have one question. It is suspicious to me to see 1 recovered prepared transaction as a result of `kill` but not in `cancel`. I cannot find an explanation for it. Can you please clarify for me? What is the difference of cancel and kill inside a transaction block? Do we recover the cascading operation from reference to hash distributed table?

onderkalaci:
> As I can see from the dump_network_traffic function, we do not send 2 PREPARE commands. We do the both truncate operations in the same transaction and send only 1 prepare.

Oh, sorry. It should be the second `TRUNCATE` command. Can you follow what I mean now? The `TRUNCATE` cascades to `hash distributed table`s shards over the same connection and we'd want to cancel during that `TRUNCATE`

> . Can you please clarify for me? What is the difference of cancel and kill inside a transaction block?

Hmm, that's a tricky one.

PostgreSQL (and Citus) holds interrupts (e.g., ignore any signal) during `XACT_EVENT_COMMIT/XACT_EVENT_ROLLBACK`. However, Citus sends `PREPARE TRANSCATION` during `XACT_EVENT_PRE_COMMIT` handler where the interrupts have not been held yet.

So, from Postgres code:

- Call`XACT_EVENT_PRE_COMMIT` event [here](https://github.com/postgres/postgres/blob/master/src/backend/access/transam/xact.c#L1977)
- Hold interrupts [here](https://github.com/postgres/postgres/blob/master/src/backend/access/transam/xact.c#L2018)
- Call `XACT_EVENT_COMMIT` [here](https://github.com/postgres/postgres/blob/master/src/backend/access/transam/xact.c#L2078) while holding interrupts

So, when we cancel at `XACT_EVENT_PRE_COMMIT` postgres/Citus is able to cancel the transaction without any issues given the interrupts haven't been held. The signal is handled properly.

When we `kill` the connection, the connection between the coordinator and the worker is broken manually by us. So, we're not actually sending any signals to the backend. Thus, the packets that come after `PREPARE TRANSACTION` (e.g., `COMMIT PREPARED`) fails and Citus treats that coordinated transaction failed for that connection. So, it needs to recover that transaction. To make sure that the above is true, could you reset the network traffic before running this specific test and then dumping the network traffic right after `TRUNCATE`? It should show what is the latest packet that has been sent over the connection. I mean, we should see something different than `PREPARE TRANSACTION`.

Though it took me some time to, thanks for sharing your observation and please do the test I suggest above, and optionally share the response here.

onderkalaci:
> Thus, the packets that come after `PREPARE TRANSACTION` (e.g., `COMMIT PREPARED`) fails and Citus treats that coordinated transaction failed for that connection. So, it needs to recover that transaction. To make sure that the above is true, could you reset the network traffic before running this specific test and then dumping the network traffic right after `TRUNCATE`? It should show what is the latest packet that has been sent over the connection. I mean, we should see something different than `PREPARE TRANSACTION`.

Sorry, you should see `conn.onCommandComplete(command="PREPARE TRANSACTION")` or its equivalant dump output.

src/test/regress/expected/failure_truncate.out
```diff
@@ -0,0 +1,1178 @@
...
+     0
+(1 row)
+
+SELECT count(*) FROM test_table;
+ count 
+-------
+    20
+(1 row)
+
+-- kill as soon as the coordinator sends COMMIT
```
onderkalaci: @furkansahin see @lithp's important comment for this: https://github.com/citusdata/citus/pull/2212/files#r201830707

Can you fix those?

furkansahin:
@onderkalaci thank you so much for the detailed explanation. Now, I get it clearly and I run the tests by hand to examine what you wanted me to. 
> you should see conn.onCommandComplete(command="PREPARE TRANSACTION") or its equivalant dump output.

You are right, we see that in the output;
```
SELECT * FROM citus.dump_network_traffic();
 conn |   source    |                                                                                                                                                                   message                                                                                                                                                                    
------+-------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
   70 | coordinator | ["Query(query=BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED;SELECT assign_distributed_transaction_id(0, XX, 'XXXX-XX-XX XX:XX:XX.XXXXXX-XX');)"]
   70 | worker      | ['CommandComplete(command=BEGIN)', "RowDescription(fieldcount=1,fields=['F(name=assign_distributed_transaction_id,tableoid=0,colattrnum=0,typoid=2278,typlen=4,typmod=-1,format_code=0)'])", 'DataRow(columncount=1,columns=["C(length=0,value=b\'\')"])', 'CommandComplete(command=SELECT 1)', 'ReadyForQuery(state=in_transaction_block)']
   70 | coordinator | ['Query(query=TRUNCATE TABLE truncate_failure.reference_table_102012 CASCADE)']
   70 | worker      | ['Backend(type=N,body=b\'SNOTICE\\x00VNOTICE\\x00C00000\\x00Mtruncate cascades to table "test_table_102009"\\x00Ftablecmds.c\\x00L1283\\x00RExecuteTruncate\\x00\\x00\')']
   70 | worker      | ['Backend(type=N,body=b\'SNOTICE\\x00VNOTICE\\x00C00000\\x00Mtruncate cascades to table "test_table_102011"\\x00Ftablecmds.c\\x00L1283\\x00RExecuteTruncate\\x00\\x00\')']
   70 | worker      | ['CommandComplete(command=TRUNCATE TABLE)', 'ReadyForQuery(state=in_transaction_block)']
   70 | coordinator | ['Query(query=TRUNCATE TABLE truncate_failure.test_table_102009 CASCADE)']
   70 | worker      | ['CommandComplete(command=TRUNCATE TABLE)', 'ReadyForQuery(state=in_transaction_block)']
   70 | coordinator | ['Query(query=TRUNCATE TABLE truncate_failure.test_table_102011 CASCADE)']
   70 | worker      | ['CommandComplete(command=TRUNCATE TABLE)', 'ReadyForQuery(state=in_transaction_block)']
   70 | coordinator | ["Query(query=PREPARE TRANSACTION 'citus_0_14518_27_6')"]
   70 | worker      | ['CommandComplete(command=PREPARE TRANSACTION)', 'ReadyForQuery(state=idle)']
(12 rows)

```

And I see that cancel sends `ROLLBACK` commands properly.

I applied your review and @lithp 's feedback as well.

src/test/regress/expected/failure_truncate.out
```diff
@@ -0,0 +1,1270 @@
...
+-- One shard should not get truncated but the other should
+-- since it is sent from another connection. 
+-- Thus, we should see a partially successful truncate
```
onderkalaci: Minor: Maybe mention that this is the result of using 1PC and there is no way to recover from it?

src/test/regress/expected/failure_truncate.out
```diff
@@ -0,0 +1,1270 @@
...
+SELECT count(*) FROM test_table;
+ count 
+-------
+    12
+(1 row)
+
+-- refill the table
+TRUNCATE test_table;
+INSERT INTO test_table SELECT x,x FROM generate_series(1,20) as f(x);
+-- cancel as soon as the coordinator sends COMMIT
```
onderkalaci: Minor: The comment seems a little bit off. See [here](https://github.com/citusdata/citus/pull/2212/files#diff-1490fb849241ff1456eb927ed102394aR225)

src/test/regress/expected/failure_truncate.out
```diff
@@ -0,0 +1,1270 @@
...
+SELECT count(*) FROM test_table;
+ count 
+-------
+     0
+(1 row)
+
+-- refill the table
+TRUNCATE test_table;
+INSERT INTO test_table SELECT x,x FROM generate_series(1,20) as f(x);
+-- now kill just after the worker sends response to
```
onderkalaci: Minor: Maybe `SET client_min_messages TO WARNING;` for this test since those warnings are consistent

src/test/regress/expected/failure_truncate.out
```diff
@@ -0,0 +1,1270 @@
...
+SELECT * FROM unhealthy_shard_count;
+ count 
+-------
+     0
+(1 row)
+
+-- we should be able to revocer the transaction and
+-- see that the command is rollbacked
+SELECT recover_prepared_transactions();
+ recover_prepared_transactions
```
onderkalaci: Hard question: Can you explain why we've one prepared transaction here and two in the below test?

onderkalaci: So, we're not sure why there is a pending prepared transaction in the worker.

That is not related to this PR and seems not a critical issue for now. So, @furkansahin will explain this to @marcocitus and open issue if it is a bug.

marcocitus: My expectation is that there are 0 prepared transactions after a cancellation.

What I see in the `cancel()` test is that the `ROLLBACK` is sent before the result for the `PREPARE TRANSACTION` is consumed, which causes it to not roll back.

```
LOG:  issuing ROLLBACK PREPARED 'citus_0_23234_85_15'
DETAIL:  on server localhost:57640
WARNING:  connection error: localhost:57640
WARNING:  failed to roll back prepared transaction 'citus_0_23234_85_15'
HINT:  Run "ROLLBACK PREPARED 'citus_0_23234_85_15'" on localhost:57640
```

That's curious since we do call `ForgetResults` before sending `ROLLBACK PREPARED`, but this might be another `ForgetResults` bug.

marcocitus: After the cancellation, we get to this line:
https://github.com/citusdata/citus/blob/master/src/backend/distributed/connection/remote_commands.c#L740

which causes ForgetResults to return early:
https://github.com/citusdata/citus/blob/master/src/backend/distributed/connection/remote_commands.c#L91

It kind of makes sense to bail out of `ForgetResults` on cancellation, since a query might be stuck waiting for network I/O and you might want to stop waiting, even if you're already in the abort handler, but if there was just a single cancellation we should try to consume the result of the PREPARED TRANSACTION and send the ROLLBACK PREPARED.

marcocitus: btw, this result seems like a race condition that just happens to trigger almost all of the time.

marcocitus: I opened and closed an issue, it's actually expected behaviour because there are 2 cancellations, so the second one ends up cancelling the `ROLLBACK PREPARED`:
https://github.com/citusdata/citus/issues/2287

This still seems like a race condition that occassionally could return a different result. Is there a way to ensure there's only one cancellation?

PR #2275, opened 2018-07-10
Ensure we create shards as the shard owner

### LINKED ISSUE #2271: Incorrect shard owner when creating distributed table as superuser
This has been brought up in #2094 and @thanodnl ran into it separately, but there wasn't an explicit issue. We don't set the table correctly on the shards when running `create_distributed_table` as superuser.

```
$ psql -U foo -c "CREATE TABLE foo (x int, y int)"
CREATE TABLE
$ psql -U marco -c "SELECT create_distributed_table('foo.foo', 'x')"
 create_distributed_table 
──────────────────────────
 
(1 row)
$ psql -U marco -p 9700 -c "SELECT tablename, tableowner FROM pg_tables WHERE tablename LIKE 'foo_%'"
Timing is on.
Line style is unicode.
 tablename  │ tableowner 
────────────┼────────────
 foo_107388 │ marco
 foo_107390 │ marco
 foo_107392 │ marco
 foo_107394 │ marco
 foo_107396 │ marco
 foo_107398 │ marco
 foo_107400 │ marco
 foo_107402 │ marco
 foo_107404 │ marco
 foo_107406 │ marco
 foo_107408 │ marco
 foo_107410 │ marco
 foo_107412 │ marco
 foo_107414 │ marco
 foo_107416 │ marco
 foo_107418 │ marco
(16 rows)
```

`tableowner` should be `foo`, since `foo` ran the original `CREATE TABLE`. 

The reason is that `StartPlacementListConnection` fails to pass down the `userName` parameter and calls `StartNodeConnection` instad of `StartNodeUserDatabaseConnection`:
https://github.com/citusdata/citus/blob/master/src/backend/distributed/connection/placement_connection.c#L314

We're actually connection as superuser, though `placement_connection.c` is acting as if we connected as the table owner, which is why we get the error described in #2094. While the error is expected behaviour, connecting as superuser is not.

marcocitus:
DESCRIPTION: Fixes a bug that might cause shards to have the wrong owner

We currently don't pass the user name from `StartPlacementListConnection` to `StartNodeUserDatabaseConnection`, and end up always connecting as the current user when using `StartPlacementListConnection`. 

When running `create_distributed_table` as superuser on a table that is owned by a regular user, we should be connecting as the table owner, but we end up still connecting as superuser. This PR fixes the issue.

Fixes #2271


metdos:
Do we want this in 7.5?

metdos:
> Do we want this in 7.5?

@marcocitus said yes :)

metdos:
@pykello will you have time to look into this?

pykello:
@metdos Sorry for delay. I will look into this today.

PR #2276, opened 2018-07-10
Drop sequences on workers as current user in master_drop_sequences

### LINKED ISSUE #2274: master_drop_sequences should connect as current user
We use `master_drop_sequences` to drop sequences from worker nodes in an MX set-up, because the sequences are created separately from the tables. Currently, the coordinator connects to the worker nodes as superuser, but it should connect as the current user to avoid dropping a sequence that the user is not allowed to drop.

marcocitus:
DESCRIPTION: Fixes a bug where master_drop_sequences connects as the wrong user

This should only be consequential for MX, since that's the only case in which there are sequences on workers.

We should perhaps consider renaming `SendCommandToWorkers` to `SendCommandToWorkersAsSuperuser` to avoid bugs.

Fixed #2274

### DIFF
--- a/src/backend/distributed/master/master_delete_protocol.c
+++ b/src/backend/distributed/master/master_delete_protocol.c
@@ -305,8 +305,8 @@ master_drop_sequences(PG_FUNCTION_ARGS)
 	{
 		appendStringInfoString(dropSeqCommand, " CASCADE");
 
-		SendCommandToWorkers(WORKERS_WITH_METADATA, DISABLE_DDL_PROPAGATION);
-		SendCommandToWorkers(WORKERS_WITH_METADATA, dropSeqCommand->data);
+		SendCommandToWorkersAsCurrentUser(WORKERS_WITH_METADATA, DISABLE_DDL_PROPAGATION);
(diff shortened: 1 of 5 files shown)


metdos:
Do we want this in 7.5?

onderkalaci:
> We should perhaps consider renaming SendCommandToWorkers to SendCommandToWorkersAsSuperuser to avoid bugs.

While reading the tenant isolation code, I've realized that we're running some commands with the current user and some with the extension owner user. And, to realize that, one should really go deep into the function definitions. So, +1 for such a change such that who writes/reviews the code thinks more on which user to use for the commands that are considered.

One interesting case might be that if someone uses two different users to connect to the workers, it might lead to self deadlocks. But, I cannot come-up with an API that could simplify that. We could maybe consider adding some user related information when `set citus.log_remote_commands TO ON;` in the future so that we could more easily catch any such dangerous cases during development ?

marcocitus:
> Do we want this in 7.5?

Probably not. There's a weird deadlock issue in travis caused by this change, which requires more investigation.

marcocitus:
Not the right fix for now.

PR #2277, opened 2018-07-11
Added failure test add/disable/inactive nodes

mtuncer:
Tests failure cases for
- master_add_node
- master_activate_node
- master_disable_node
- master_add_inactive_node

### DIFF
--- a/src/test/regress/expected/failure_add_disable_node.out
+++ b/src/test/regress/expected/failure_add_disable_node.out
@@ -1,0 +1,387 @@
+--
+-- failure_add_disable_node tests master_add_node, master_remove_node
+-- master_activate_node for failures.
+-- master_disable_node and master_add_inactive_node can not be
+-- tested as they don't create network activity
+--
+ 
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SET citus.next_shard_id TO 200000;
+-- verify we have all worker nodes present
+SELECT * FROM master_get_active_worker_nodes()
+ORDER BY 1, 2;
+ node_name | node_port 
+-----------+-----------
+ localhost |     57637
+ localhost |     57640
+(2 rows)
+
+-- verify there are no tables that could prevent add/remove node operations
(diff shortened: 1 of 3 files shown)


src/test/regress/expected/failure_add_disable_node.out
```diff
@@ -0,0 +1,359 @@
...
+ 
+(1 row)
+
+SELECT citus.clear_network_traffic();
+ clear_network_traffic 
+-----------------------
+ 
+(1 row)
+
+SELECT master_add_node('localhost', :worker_1_port);
```
onderkalaci: I think adding/removing `worker_1_port` tests one side of things. The [proxy](https://github.com/citusdata/citus/pull/2119/files#diff-a7ce4fb174e1acce25b5bcd3121576a9R572) is on `:worker_2_port`, so we should be adding/removing that as well. In these tests, we're catching the `COPY .. STDOUT` command that `worker_2` executes due to `worker_append_table_to_shard` that is being executed on `worker_1`.

 Note that the coordinator sends bunch of other commands to the target node as well:
```SQL
Select master_add_node('localhost', 9702);
NOTICE:  Replicating reference table "ref" to the node localhost:9702
LOG:  issuing BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED;SELECT assign_distributed_transaction_id(0, 22, '2018-07-11 17:46:36.461851+03');
LOG:  issuing SELECT worker_apply_shard_ddl_command (102074, 'DROP TABLE IF EXISTS public.ref CASCADE')
LOG:  issuing SELECT worker_apply_shard_ddl_command (102074, 'CREATE TABLE public.ref (a integer)')
LOG:  issuing SELECT worker_append_table_to_shard ('public.ref_102074', 'public.ref_102074', 'localhost', 9701)
LOG:  issuing COMMIT
                 master_add_node                  
--------------------------------------------------
 (9,9,localhost,9702,default,f,t,primary,default)
```

And, we should probably run these tests in a seperate schema and add a test on failure during `CREATE SCHEMA IF NOT EXISTS` ?

PR #2278, opened 2018-07-11
Adds failure tests for COPY to reference table

furkansahin:
Failure test for copy to reference table is added

### DIFF
--- a/src/test/regress/expected/failure_copy_to_reference.out
+++ b/src/test/regress/expected/failure_copy_to_reference.out
@@ -1,0 +1,461 @@
+-- 
+--  Failure tests for COPY to reference tables 
+-- 
+CREATE SCHEMA copy_reference_failure;
+SET search_path TO 'copy_reference_failure';
+SET citus.next_shard_id TO 130000;
+-- we don't want to see the prepared transaction numbers in the warnings
+SET client_min_messages TO ERROR;
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+CREATE TABLE test_table(id int, value_1 int);
+SELECT create_reference_table('test_table');
+ create_reference_table 
+------------------------
+ 
+(1 row)
+
+CREATE VIEW unhealthy_shard_count AS 
+  SELECT count(*) 
+  FROM pg_dist_shard_placement pdsp 
+  JOIN 
+  pg_dist_shard pds 
+  ON pdsp.shardid=pds.shardid 
+  WHERE logicalrelid='copy_reference_failure.test_table'::regclass AND shardstate != 1;
+-- in the first test, kill just in the first 
+-- response we get from the worker
+SELECT citus.mitmproxy('conn.kill()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+\copy test_table FROM STDIN DELIMITER ','
+ERROR:  connection error: localhost:57640
+DETAIL:  server closed the connection unexpectedly
+	This probably means the server terminated abnormally
+	before or while processing the request.
+CONTEXT:  COPY test_table, line 1: "1,2"
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT * FROM unhealthy_shard_count;
+ count 
+-------
+     0
+(1 row)
+
+SELECT count(*) FROM test_table;
+ count 
+-------
+     0
+(1 row)
+
+-- kill as soon as the coordinator sends begin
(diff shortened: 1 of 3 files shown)


src/test/regress/expected/failure_copy_to_reference.out
```diff
@@ -0,0 +1,434 @@
...
+-- we don't want to see the prepared transaction numbers in the warnings
+SET client_min_messages TO ERROR;
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+-- With one placement COPY should error out and placement should stay healthy.
+SET citus.shard_replication_factor TO 1;
```
onderkalaci: Minor: setting the `shard_replication_factor ` and `shard_count ` is slightly confusing since we don't rely on those

src/test/regress/expected/failure_copy_to_reference.out
```diff
@@ -0,0 +1,434 @@
...
+
+SELECT count(*) FROM test_table;
+ count 
+-------
+     3
+(1 row)
+
+TRUNCATE test_table;
+-- kill as soon as the coordinator sends COMMIT
+SELECT citus.mitmproxy('conn.onQuery(query="COMMIT").kill()');
```
onderkalaci: Missing `^` in `COMMIT`. I think that's why the rest of the tests comments are not accurate?

src/test/regress/expected/failure_copy_to_reference.out
```diff
@@ -0,0 +1,434 @@
...
+
+-- kill as soon as the worker sends CopyComplete
+SELECT citus.mitmproxy('conn.onCommandComplete(command="COPY 3").kill()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+\copy test_table FROM STDIN DELIMITER ','
+ERROR:  failed to COPY to shard 102008 on localhost:57640
```
onderkalaci: Since we rely on shard id, lets set it at the start of the test.

src/test/regress/expected/failure_copy_to_reference.out
```diff
@@ -0,0 +1,434 @@
...
+SELECT count(*) FROM test_table;
+ count 
+-------
+     0
+(1 row)
+
+SET citus.multi_shard_commit_protocol TO '2pc';
+ALTER TABLE test_table ADD CONSTRAINT uniq UNIQUE (value_1);
+BEGIN;
+\copy test_table FROM STDIN DELIMITER ','
```
onderkalaci: Why do we have this test here? I like extra tests but still I'm OK if you remove this 😄

src/test/regress/expected/failure_copy_to_reference.out
```diff
@@ -0,0 +1,434 @@
...
+     0
+(1 row)
+
+SELECT count(*) FROM test_table;
+ count 
+-------
+     0
+(1 row)
+
+-- cancel as soon as the coordinator sends begin
```
onderkalaci: Maybe add one more step
```
+-- ==== kill the connection when we try to start the COPY ====
+-- the query should abort
+SELECT citus.mitmproxy('conn.onQuery(query="FROM STDIN WITH").killall()');
```

src/test/regress/expected/failure_copy_to_reference.out
```diff
@@ -0,0 +1,434 @@
...
+-- finally, test failing on ROLLBACK just after the coordinator
+-- sends the ROLLBACK so the command can be rollbacked
+SELECT citus.mitmproxy('conn.onQuery(query="ROLLBACK").kill()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+BEGIN;
+\copy test_table FROM STDIN DELIMITER ','
```
onderkalaci: Why don't we see any failures here?

src/test/regress/expected/failure_copy_to_reference.out
```diff
@@ -0,0 +1,434 @@
...
+-- ROLLBACK command, command should have been rollbacked
+-- both on the distributed table and the placements
+SELECT citus.mitmproxy('conn.onCommandComplete(command="ROLLBACK").kill()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+BEGIN;
+TRUNCATE test_table;
```
onderkalaci: It should be COPY?

src/test/regress/expected/failure_copy_to_reference.out
furkansahin: What I understand from the other tests such as [this](https://github.com/citusdata/citus/pull/2212/files#diff-1490fb849241ff1456eb927ed102394aR311), it is not that possible to error out during COMMIT/ROLLBACK. For `cancel`, it is hold. For kill, I'm not sure but we give warnings about the problem if the `client_min_messages` is set to `WARNING`.

PR #2279, opened 2018-07-11
Add create_distributed_table (without data) failure tests

velioglu:
Add failure test for creating distributed table without data

### DIFF
--- a/src/test/regress/expected/failure_create_table.out
+++ b/src/test/regress/expected/failure_create_table.out
@@ -1,0 +1,697 @@
+--
+-- failure_create_table adds failure tests for creating table without data.
+--
+CREATE SCHEMA failure_create_table;
+SET search_path TO 'failure_create_table';
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SET citus.shard_replication_factor TO 1;
+SET citus.shard_count to 4;
+CREATE TABLE test_table(id int, value_1 int);
+-- Kill connection before sending query to the worker 
+SELECT citus.mitmproxy('conn.kill()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT create_distributed_table('test_table','id');
+ERROR:  connection error: localhost:9060
+DETAIL:  server closed the connection unexpectedly
+	This probably means the server terminated abnormally
+	before or while processing the request.
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT count(*) FROM pg_dist_shard;
+ count 
+-------
+     0
+(1 row)
+
+SELECT run_command_on_workers($$SELECT count(*) FROM information_schema.tables WHERE table_schema = 'failure_create_table' and table_name LIKE 'test_table%' ORDER BY 1$$);
+ run_command_on_workers 
+------------------------
+ (localhost,9060,t,0)
+ (localhost,57637,t,0)
+(2 rows)
+
+-- kill as soon as the coordinator sends CREATE SCHEMA
+-- Since schemas are created in separate transaction, schema will
+-- be created only on the node which is not behind the proxy.
+-- https://github.com/citusdata/citus/pull/1652
+SELECT citus.mitmproxy('conn.onQuery(query="^CREATE SCHEMA").kill()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT create_distributed_table('test_table', 'id');
+ERROR:  server closed the connection unexpectedly
+	This probably means the server terminated abnormally
+	before or while processing the request.
+CONTEXT:  while executing command on localhost:9060
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT count(*) FROM pg_dist_shard;
+ count 
+-------
+     0
+(1 row)
+
+SELECT run_command_on_workers($$SELECT count(*) FROM information_schema.schemata WHERE schema_name = 'failure_create_table'$$);
+ run_command_on_workers 
+------------------------
+ (localhost,9060,t,0)
+ (localhost,57637,t,1)
+(2 rows)
+
+-- Now, kill the connection while opening transaction on workers.
+SELECT citus.mitmproxy('conn.onQuery(query="^BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED").kill()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT create_distributed_table('test_table','id');
+WARNING:  connection not open
+CONTEXT:  while executing command on localhost:9060
+ERROR:  connection error: localhost:9060
+DETAIL:  connection not open
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT count(*) FROM pg_dist_shard;
+ count 
+-------
+     0
+(1 row)
+
+SELECT run_command_on_workers($$SELECT count(*) FROM information_schema.tables WHERE table_schema = 'failure_create_table' and table_name LIKE 'test_table%' ORDER BY 1$$);
+ run_command_on_workers 
+------------------------
+ (localhost,9060,t,0)
+ (localhost,57637,t,0)
+(2 rows)
+
+-- Now, kill the connection after sending create table command with worker_apply_shard_ddl_command UDF
+SELECT citus.mitmproxy('conn.onQuery(query="^SELECT worker_apply_shard_ddl_command").after(2).kill()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT create_distributed_table('test_table','id');
+ERROR:  server closed the connection unexpectedly
+	This probably means the server terminated abnormally
+	before or while processing the request.
+CONTEXT:  while executing command on localhost:9060
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT count(*) FROM pg_dist_shard;
+ count 
+-------
+     0
+(1 row)
+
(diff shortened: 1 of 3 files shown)


src/test/regress/expected/failure_create_table.out
```diff
@@ -0,0 +1,513 @@
...
+(1 row)
+
+\c - - - :master_port
+SET search_path TO 'failure_create_table';
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
```
onderkalaci: maybe call `recover_prepared_transactions()` to show that there are no pending transactions?

src/test/regress/expected/failure_create_table.out
```diff
@@ -0,0 +1,513 @@
...
+-- workers. Note that, cancel requests will be ignored during
+-- shard creation.
+SELECT citus.mitmproxy('conn.onQuery(query="^BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED").cancel(' || pg_backend_pid() || ')');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+BEGIN;
+SELECT create_distributed_table('test_table','id');
```
onderkalaci: One slightly odd thing about shard creation is that:
- If there are no colocated tables, cancel requests are ignored (see `CreateShardsWithRoundRobinPolicy` -> `HOLD_INTERRUPTS`)
- If there are some colocated tables, `CreateColocatedShards()` there is no `HOLD_INTERRUPTS`. So, could you add a test, say with `colocate_with` option, and cancel and failure during shard creation?

src/test/regress/expected/failure_create_table.out
```diff
@@ -0,0 +1,513 @@
+CREATE SCHEMA failure_create_table;
+SET search_path TO 'failure_create_table';
+SELECT citus.mitmproxy('conn.allow()');
```
onderkalaci: Given that Citus create schemas in a slightly odd way, could you add some failure tests while creating the schema via `EnsureSchemaExistsAllNodes()`?

src/test/regress/expected/failure_create_table.out
```diff
@@ -0,0 +1,513 @@
...
+
+-- Now, kill the connection just after transaction is opened on
+-- workers.
+SELECT citus.mitmproxy('conn.onQuery(query="^BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED").kill()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT create_distributed_table('test_table','id');
```
onderkalaci: It'd also be interesting to fail during distributed table creation where the table has a foreign key to a reference table. Failing on foreign key creation is the one that is a different thing to test. We don't need to add other tests in that scenario.

src/test/regress/expected/failure_create_table.out
velioglu: Done

src/test/regress/expected/failure_create_table.out
```diff
@@ -0,0 +1,597 @@
...
+DETAIL:  server closed the connection unexpectedly
+	This probably means the server terminated abnormally
+	before or while processing the request.
+SELECT count(*) FROM pg_dist_shard;
+ count 
+-------
+     0
+(1 row)
+
+\c - - - :worker_1_port
```
lithp: To make this easier you can also run `run_command_on_workers`, that will test both workers simultaneously and saves you from having to `\connect` and reset `search_path`

Even if you decide not to use `run_command_on_workers`, maybe add a comment that you're testing that _the other_ worker wasn't given any shards. Worker 2 is the one behind the proxy, something which might not be obvious to future readers.

src/test/regress/expected/failure_create_table.out
```diff
@@ -0,0 +1,597 @@
...
+ 
+(1 row)
+
+SELECT count(*) FROM pg_dist_shard;
+ count 
+-------
+     0
+(1 row)
+
+SELECT run_command_on_workers($$SELECT count(*) FROM information_schema.schemata WHERE schema_name = 'failure_create_table'$$);
```
lithp: It's really strange that the `CREATE SCHEMA` isn't created inside a transaction, so the other node is given a schema even though we aren't. I'm not sure, maybe add a comment saying this is expected behavior?

src/test/regress/expected/failure_create_table.out
```diff
@@ -0,0 +1,597 @@
...
+CONTEXT:  while executing command on localhost:57640
+ERROR:  connection error: localhost:57640
+DETAIL:  connection not open
+SELECT count(*) FROM pg_dist_shard;
+ count 
+-------
+     0
+(1 row)
+
+\c - - - :worker_1_port
```
lithp: Again, `run_command_on_workers` might be easier to write and read, though you might need to `conn.allow()` before you run it.

src/test/regress/expected/failure_create_table.out
```diff
@@ -0,0 +1,597 @@
...
+SELECT count(*) FROM pg_class WHERE relname LIKE 'test_table%';
+ count 
+-------
+     0
+(1 row)
+
+\c - - - :master_port
+SET search_path TO 'failure_create_table';
+SELECT citus.mitmproxy('conn.allow()');
```
lithp: This `conn.allow()` is unnecessary, because you're immediately running `citus.mitmproxy` again. This `conn.allow()` configuration is never exercised by any network traffic before you reconfigure.

src/test/regress/expected/failure_create_table.out
```diff
@@ -0,0 +1,597 @@
...
+ count 
+-------
+     4
+(1 row)
+
+\c - - - :worker_1_port
+SELECT count(*) FROM pg_class WHERE relname LIKE 'test_table%';
+ count 
+-------
+     4
```
lithp: Uh oh, there should be 2 of them! 4 shards with replication factor 1 -> 2 shards per worker.

Because you keep using `\connect` you've lost your settings at the top, and we make 4 shards with replication factor 2, I think?

This isn't a bug but it makes the tests harder to read. It'd be nicer if:
- After every `\c - - - :master_port` all the settings were set again
- We didn't use \c, so the settings at the top of the file were maintained.

src/test/regress/expected/failure_create_table.out
```diff
@@ -0,0 +1,597 @@
...
+SET search_path TO 'failure_create_table';
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+-- Now, kill the connection just after transaction is opened on
+-- workers.
+SELECT citus.mitmproxy('conn.onQuery(query="^BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED").kill()');
```
lithp: This test is here twice, once here and once [on line 74](https://github.com/citusdata/citus/pull/2279/files#diff-e491652e5bcabc6bdef5e8003c5ec515R74). What's the difference? This should probably include a comment saying what it is that's being tested.

src/test/regress/expected/failure_create_table.out
```diff
@@ -0,0 +1,597 @@
...
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+-- Now, cancel the connection just after transaction is opened on
+-- workers. Note that, cancel requests will be ignored during
+-- shard creation.
+SELECT citus.mitmproxy('conn.onQuery(query="^BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED").cancel(' || pg_backend_pid() || ')');
```
lithp: This test has also already been one, on line 109.

src/test/regress/expected/failure_create_table.out
```diff
@@ -0,0 +1,597 @@
...
+(1 row)
+
+\c - - - :worker_1_port
+SELECT count(*) FROM pg_class WHERE relname LIKE 'test_table%';
+ count 
+-------
+     0
+(1 row)
+
+\c - - - :master_port
```
lithp: Here you're losing the `citus.multi_shard_commit_protocol` setting.

src/test/regress/expected/failure_create_table.out
```diff
@@ -0,0 +1,597 @@
...
+SET search_path TO 'failure_create_table';
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+-- Kill the connection after worker sends "PREPARE TRANSACTION" ack
+SELECT citus.mitmproxy('conn.onQuery(query="^PREPARE TRANSACTION").kill()');
```
lithp: If you were in 1PC then there would be no `PREPARE TRANSACTION`, there's only a `COMMIT` after all the `worker_shard_apply_ddl_commands`.

src/test/regress/expected/failure_create_table.out
```diff
@@ -0,0 +1,597 @@
...
+(1 row)
+
+SELECT count(*) FROM pg_dist_shard WHERE logicalrelid='test_table'::regclass;
+ count 
+-------
+     0
+(1 row)
+
+DROP TABLE temp_table;
+-- Kill the connection after worker sends "PREPARE TRANSACTION" ack
```
lithp: This comment does not match the command you sent mitmproxy. To kill at an ack you can use a string like `conn.onCommandComplete(command="PREPARE TRANSACTION")`.

src/test/regress/expected/failure_create_table.out
lithp: Here's an example from another PR: https://github.com/citusdata/citus/pull/2280/files#diff-3ba797ec559d43345abd6c6283fd5e2dR484

src/test/regress/expected/failure_create_table.out
```diff
@@ -0,0 +1,695 @@
...
+(2 rows)
+
+-- Cancel the connection after sending prepare transaction in master_create_worker_shards
+SELECT citus.mitmproxy('conn.onCommandComplete(command="PREPARE TRANSACTION").cancel(' ||  pg_backend_pid() || ')');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT master_create_worker_shards('test_table_2', 4, 2);
```
onderkalaci: call recovery UDF right after this one.

Issue #1637, opened 2017-09-07
citus-7 hangs on create_distributed_table() and create_reference_table()


taybin:
There aren't any obvious errors in the logs, although I do see some of these on the citus master which I hadn't before:

```
ERROR:  type "hstore" does not exist at character 8
STATEMENT:  SELECT 'hstore'::regtype::oid
```
Creating the hstore extension silences those errors, but the functions still hang.

The exact same commands worked with citus-6.2.3 and the only difference is the new docker image.

taybin:
I'm not sure if this is related to the new deadlock detection feature (because something seems to be locked up), but I'd like to know how to debug this.  I turned on the citus.log_distributed_deadlock_detection config, but nothing extra was printed out.  Usually there is an error or warning that gives me a hint of what to fix, but with this I'm not seeing anything.

marcocitus:
Are you calling `create_distributed_table` from a function? Is there anything else the function does?

taybin:
It's running inside a migration, which is grouping a number of create_distributed_table() calls inside a single transaction.  Could that be a problem now?

marcocitus:
It shouldn't be a problem unless there's a bug. Can you give a sequence of commands that reproduces the problem?

onderkalaci:
It'd also be useful if you share the number of workers (and the shard count - replication factor if you're not using the defaults).

onderkalaci:
@taybin Any news on the issue? We'd be happy to understand and fix the issues if there are any.

> I turned on the citus.log_distributed_deadlock_detection config, but nothing extra was printed out.

We didn't want to log too many things unless Citus finds any deadlocks. So, the steps to see some logs on the log file:

1. ALTER SYSTEM set citus.log_distributed_deadlock_detection = true;
2. Reload the config (SELECT pg_reload_conf()) since its a [SIGHUP](https://github.com/citusdata/citus/blob/45957e568896cf2f963b13e4fc82e2914efc85cf/src/backend/distributed/shared_library_init.c#L365) GUC variable
3. Generate a deadlock and check the log file, you should see some logs like the following:
```
LOG:  [2017-09-11 19:38:50.845865+03] Distributed deadlock found among the following distributed transactions:
CONTEXT:  Citus maintenance daemon for database 12669 user 10
[...]

PR #2280, opened 2018-07-12
Add create_distributed_table (with data) failure tests

furkansahin:
Adds failure tests for create_distributed_table UDF on non-empty table

### DIFF
--- a/src/test/regress/expected/failure_create_distributed_table_non_empty.out
+++ b/src/test/regress/expected/failure_create_distributed_table_non_empty.out
@@ -1,0 +1,977 @@
+-- 
+--  Failure tests for COPY to reference tables 
+-- 
+CREATE SCHEMA create_distributed_table_non_empty_failure;
+SET search_path TO 'create_distributed_table_non_empty_failure';
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+-- we'll start with replication factor 1 and 2pc
+SET citus.shard_replication_factor TO 1;
+SET citus.shard_count to 4;
+CREATE TABLE test_table(id int, value_1 int);
+INSERT INTO test_table VALUES (1,1),(2,2),(3,3),(4,4);
+-- in the first test, kill the first connection we sent from the coordinator
+SELECT citus.mitmproxy('conn.kill()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT create_distributed_table('test_table', 'id');
+ERROR:  connection error: localhost:9060
+DETAIL:  server closed the connection unexpectedly
+	This probably means the server terminated abnormally
+	before or while processing the request.
+SELECT count(*) FROM pg_dist_shard WHERE logicalrelid='create_distributed_table_non_empty_failure.test_table'::regclass;
+ count 
+-------
+     0
+(1 row)
+
+-- in the first test, cancel the first connection we sent from the coordinator
+SELECT citus.mitmproxy('conn.cancel(' ||  pg_backend_pid() || ')');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT create_distributed_table('test_table', 'id');
+ERROR:  canceling statement due to user request
+SELECT count(*) FROM pg_dist_shard WHERE logicalrelid='create_distributed_table_non_empty_failure.test_table'::regclass;
+ count 
+-------
+     0
+(1 row)
+
+-- kill as soon as the coordinator sends CREATE SCHEMA
+SELECT citus.mitmproxy('conn.onQuery(query="^CREATE SCHEMA").kill()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT create_distributed_table('test_table', 'id');
+ERROR:  server closed the connection unexpectedly
+	This probably means the server terminated abnormally
+	before or while processing the request.
+CONTEXT:  while executing command on localhost:9060
+SELECT count(*) FROM pg_dist_shard WHERE logicalrelid='create_distributed_table_non_empty_failure.test_table'::regclass;
+ count 
+-------
+     0
+(1 row)
+
+SELECT run_command_on_workers($$SELECT count(*) FROM information_schema.schemata WHERE schema_name = 'create_distributed_table_non_empty_failure'$$);
+ run_command_on_workers 
+------------------------
+ (localhost,9060,t,0)
+ (localhost,57637,t,1)
+(2 rows)
+
+-- cancel as soon as the coordinator sends CREATE SCHEMA
+-- Note: Schema should be created in workers because Citus
+-- does not check for interrupts until GetRemoteCommandResult is called.
+-- Since we already sent the command at this stage, the schemas get created in workers
+SELECT citus.mitmproxy('conn.onQuery(query="^CREATE SCHEMA").cancel(' ||  pg_backend_pid() || ')');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT create_distributed_table('test_table', 'id');
+ERROR:  canceling statement due to user request
+SELECT count(*) FROM pg_dist_shard WHERE logicalrelid='create_distributed_table_non_empty_failure.test_table'::regclass;
+ count 
+-------
+     0
+(1 row)
+
+SELECT run_command_on_workers($$SELECT count(*) FROM information_schema.schemata WHERE schema_name = 'create_distributed_table_non_empty_failure'$$);
+ run_command_on_workers 
+------------------------
+ (localhost,9060,t,1)
+ (localhost,57637,t,1)
+(2 rows)
+
+SELECT run_command_on_workers($$DROP SCHEMA IF EXISTS create_distributed_table_non_empty_failure$$);
+      run_command_on_workers       
+-----------------------------------
+ (localhost,9060,t,"DROP SCHEMA")
+ (localhost,57637,t,"DROP SCHEMA")
+(2 rows)
+
+-- kill as soon as the coordinator sends begin
+SELECT citus.mitmproxy('conn.onQuery(query="^BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED").kill()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT create_distributed_table('test_table', 'id');
+WARNING:  connection not open
+CONTEXT:  while executing command on localhost:9060
+ERROR:  connection error: localhost:9060
+DETAIL:  connection not open
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT count(*) FROM pg_dist_shard WHERE logicalrelid='create_distributed_table_non_empty_failure.test_table'::regclass;
+ count 
+-------
+     0
+(1 row)
+
+SELECT run_command_on_workers($$SELECT count(*) FROM information_schema.schemata WHERE schema_name = 'create_distributed_table_non_empty_failure'$$);
+ run_command_on_workers 
+------------------------
+ (localhost,9060,t,1)
+ (localhost,57637,t,1)
+(2 rows)
+
+-- cancel as soon as the coordinator sends begin
+-- shards will be created because we ignore cancel requests during the shard creation 
+-- Interrupts are hold in CreateShardsWithRoundRobinPolicy
+SELECT citus.mitmproxy('conn.onQuery(query="^BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED").cancel(' ||  pg_backend_pid() || ')');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT create_distributed_table('test_table', 'id');
+WARNING:  cancel requests are ignored during shard creation
+NOTICE:  Copying data from local table...
+ create_distributed_table 
+--------------------------
+ 
+(1 row)
+
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT count(*) FROM pg_dist_shard WHERE logicalrelid='create_distributed_table_non_empty_failure.test_table'::regclass;
+ count 
+-------
+     4
+(1 row)
+
+SELECT run_command_on_workers($$SELECT count(*) FROM information_schema.schemata WHERE schema_name = 'create_distributed_table_non_empty_failure'$$);
+ run_command_on_workers 
+------------------------
+ (localhost,9060,t,1)
+ (localhost,57637,t,1)
+(2 rows)
+
+DROP TABLE test_table ;
+CREATE TABLE test_table(id int, value_1 int);
+INSERT INTO test_table VALUES (1,1),(2,2),(3,3),(4,4);
+-- kill as soon as the coordinator sends CREATE TABLE
+SELECT citus.mitmproxy('conn.onQuery(query="CREATE TABLE").kill()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT create_distributed_table('test_table', 'id');
+ERROR:  server closed the connection unexpectedly
+	This probably means the server terminated abnormally
+	before or while processing the request.
+CONTEXT:  while executing command on localhost:9060
+SELECT count(*) FROM pg_dist_shard WHERE logicalrelid='create_distributed_table_non_empty_failure.test_table'::regclass;
+ count 
+-------
+     0
+(1 row)
+
+-- kill as soon as the coordinator sends COPY
+SELECT citus.mitmproxy('conn.onQuery(query="COPY").kill()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT create_distributed_table('test_table', 'id');
+ERROR:  server closed the connection unexpectedly
+	This probably means the server terminated abnormally
+	before or while processing the request.
+CONTEXT:  while executing command on localhost:9060
+SELECT count(*) FROM pg_dist_shard WHERE logicalrelid='create_distributed_table_non_empty_failure.test_table'::regclass;
+ count 
+-------
+     0
+(1 row)
+
+-- kill when the COPY is completed, it should be rollbacked properly
+SELECT citus.mitmproxy('conn.onCommandComplete(command="COPY").kill()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT create_distributed_table('test_table', 'id');
+NOTICE:  Copying data from local table...
+ERROR:  failed to COPY to shard 102052 on localhost:9060
+SELECT count(*) FROM pg_dist_shard WHERE logicalrelid='create_distributed_table_non_empty_failure.test_table'::regclass;
+ count 
+-------
+     0
+(1 row)
+
+-- cancel as soon as the coordinator sends COPY, table 
+-- should not be created and rollbacked properly
+SELECT citus.mitmproxy('conn.onQuery(query="COPY").cancel(' ||  pg_backend_pid() || ')');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT create_distributed_table('test_table', 'id');
+ERROR:  canceling statement due to user request
+SELECT count(*) FROM pg_dist_shard WHERE logicalrelid='create_distributed_table_non_empty_failure.test_table'::regclass;
+ count 
+-------
+     0
+(1 row)
+
+-- cancel when the COPY is completed, it should be rollbacked properly
+SELECT citus.mitmproxy('conn.onCommandComplete(command="COPY").cancel(' ||  pg_backend_pid() || ')');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT create_distributed_table('test_table', 'id');
+NOTICE:  Copying data from local table...
+ERROR:  canceling statement due to user request
+SELECT count(*) FROM pg_dist_shard WHERE logicalrelid='create_distributed_table_non_empty_failure.test_table'::regclass;
+ count 
+-------
+     0
+(1 row)
+
+-- immediately kill when we see prepare transaction to see if the command
+-- successfully rollbacked the created shards
+-- we don't want to see the prepared transaction numbers in the warnings
+SET client_min_messages TO ERROR;
+SELECT citus.mitmproxy('conn.onQuery(query="PREPARE TRANSACTION").kill()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT create_distributed_table('test_table', 'id');
+ERROR:  connection not open
+CONTEXT:  while executing command on localhost:9060
+SELECT count(*) FROM pg_dist_shard WHERE logicalrelid='create_distributed_table_non_empty_failure.test_table'::regclass;
+ count 
+-------
+     0
+(1 row)
+
+-- immediately cancel when we see prepare transaction to see if the command
+-- successfully rollbacked the created shards
+SELECT citus.mitmproxy('conn.onQuery(query="PREPARE TRANSACTION").cancel(' ||  pg_backend_pid() || ')');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT create_distributed_table('test_table', 'id');
+ERROR:  canceling statement due to user request
+SELECT count(*) FROM pg_dist_shard WHERE logicalrelid='create_distributed_table_non_empty_failure.test_table'::regclass;
+ count 
+-------
+     0
+(1 row)
+
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT recover_prepared_transactions();
+ recover_prepared_transactions 
+-------------------------------
+                             1
+(1 row)
+
+-- kill as soon as the coordinator sends COMMIT
+-- shards should be created and kill should not affect
+SELECT citus.mitmproxy('conn.onQuery(query="^COMMIT PREPARED").kill()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT create_distributed_table('test_table', 'id');
+ create_distributed_table 
+--------------------------
+ 
+(1 row)
+
+SELECT count(*) FROM pg_dist_shard WHERE logicalrelid='create_distributed_table_non_empty_failure.test_table'::regclass;
+ count 
+-------
+     4
+(1 row)
+
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT recover_prepared_transactions();
+ recover_prepared_transactions 
+-------------------------------
+                             2
+(1 row)
+
+DROP TABLE test_table ;
+CREATE TABLE test_table(id int, value_1 int);
+INSERT INTO test_table VALUES (1,1),(2,2),(3,3),(4,4);
+-- cancel as soon as the coordinator sends COMMIT
+-- shards should be created and kill should not affect
+SELECT citus.mitmproxy('conn.onQuery(query="^COMMIT PREPARED").cancel(' ||  pg_backend_pid() || ')');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT create_distributed_table('test_table', 'id');
+ create_distributed_table 
+--------------------------
+ 
+(1 row)
+
+SELECT citus.mitmproxy('conn.allow()');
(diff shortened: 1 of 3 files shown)


src/test/regress/expected/failure_create_distributed_table_non_empty.out
```diff
@@ -0,0 +1,520 @@
...
+-----------
+ 
+(1 row)
+
+SELECT create_distributed_table('test_table', 'id');
+ERROR:  connection error: localhost:57640
+DETAIL:  server closed the connection unexpectedly
+	This probably means the server terminated abnormally
+	before or while processing the request.
+SELECT citus.mitmproxy('conn.allow()');
```
lithp: This `conn.allow()` isn't necessary. You only need it if you're going to make more connections to the worker before the next test. Here, all you do is select from `pg_dist_shard` before running the next `citus.mitmproxy` command, so you don't need to tell the proxy to allow connections.

src/test/regress/expected/failure_create_distributed_table_non_empty.out
```diff
@@ -0,0 +1,520 @@
...
+-- in the first test, cancel the first connection we sent from the coordinator
+SELECT citus.mitmproxy('conn.cancel(' ||  pg_backend_pid() || ')');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT create_distributed_table('test_table', 'id');
+ERROR:  canceling statement due to user request
+SELECT citus.mitmproxy('conn.allow()');
```
lithp: this one is also unnecessary

src/test/regress/expected/failure_create_distributed_table_non_empty.out
```diff
@@ -0,0 +1,520 @@
...
+-------
+     0
+(1 row)
+
+SELECT run_command_on_workers($$SELECT count(*) FROM information_schema.schemata WHERE schema_name = 'create_distributed_table_non_empty_failure'$$);
+ run_command_on_workers 
+------------------------
+ (localhost,57637,t,1)
+ (localhost,57640,t,0)
+(2 rows)
```
lithp: If I'm reading this right, it means we created the schema on one worker, but not the other. Is that a bug?

src/test/regress/expected/failure_create_distributed_table_non_empty.out
```diff
@@ -0,0 +1,520 @@
...
+ count 
+-------
+     0
+(1 row)
+
+SELECT run_command_on_workers($$SELECT count(*) FROM information_schema.schemata WHERE schema_name = 'create_distributed_table_non_empty_failure'$$);
+ run_command_on_workers 
+------------------------
+ (localhost,57637,t,1)
+ (localhost,57640,t,1)
```
lithp: But now the schema exists on both, even though the user cancelled. Is that a bug?

src/test/regress/expected/failure_create_distributed_table_non_empty.out
```diff
@@ -0,0 +1,520 @@
...
+ 
+(1 row)
+
+SELECT count(*) FROM pg_dist_shard WHERE logicalrelid='create_distributed_table_non_empty_failure.test_table'::regclass;
+ count 
+-------
+     0
+(1 row)
+
+SELECT run_command_on_workers($$SELECT count(*) FROM information_schema.schemata WHERE schema_name = 'create_distributed_table_non_empty_failure'$$);
```
lithp: Maybe add a comment that this is expected, because the `SCHEMA` part went through but the transaction part failed. Is it a problem that we don't try to create the schema in the same transaction as we create the table in?

lithp: Actually, strangely enough, we first create the schema, and then we again create the schema:

```
 (0,coordinator,"[initial message]")
 (0,worker,"['AuthenticationOk()', 'ParameterStatus(application_name=citus)', 'ParameterStatus(client_encoding=UTF8)', 'ParameterStatus(DateStyle=ISO, MDY)', 'ParameterStatus(integer_datetim
es=on)', 'ParameterStatus(IntervalStyle=postgres)', 'ParameterStatus(is_superuser=on)', 'ParameterStatus(server_encoding=UTF8)', 'ParameterStatus(server_version=XXX)', 'ParameterStatus(sessi
on_authorization=brian)', 'ParameterStatus(standard_conforming_strings=on)', 'ParameterStatus(TimeZone=XXX)', 'BackendKeyData(XXX)', 'ReadyForQuery(state=idle)']")
 **** HERE **** (0,coordinator,"['Query(query=CREATE SCHEMA IF NOT EXISTS test_test AUTHORIZATION brian)']")
 (0,worker,"['CommandComplete(command=CREATE SCHEMA)', 'ReadyForQuery(state=idle)']")
 (3,coordinator,"[""Query(query=BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED;SELECT assign_distributed_transaction_id(0, XX, 'XXXX-XX-XX XX:XX:XX.XXXXXX-XX');)""]")
 (3,worker,"['CommandComplete(command=BEGIN)', ""RowDescription(fieldcount=1,fields=['F(name=assign_distributed_transaction_id,tableoid=0,colattrnum=0,typoid=2278,typlen=4,typmod=-1,format_c
ode=0)'])"", 'DataRow(columncount=1,columns=[""C(length=0,value=b\\'\\')""])', 'CommandComplete(command=SELECT 1)', 'ReadyForQuery(state=in_transaction_block)']")
**** HERE ****  (3,coordinator,"[""Query(query=SELECT worker_apply_shard_ddl_command (100923, 'test_test', 'CREATE SCHEMA IF NOT EXISTS test_test AUTHORIZATION brian'))""]")
 (3,worker,"[""RowDescription(fieldcount=1,fields=['F(name=worker_apply_shard_ddl_command,tableoid=0,colattrnum=0,typoid=2278,typlen=4,typmod=-1,format_code=0)'])"", 'Backend(type=N,body=b\\
'SNOTICE\\\\x00VNOTICE\\\\x00C42P06\\\\x00Mschema ""test_test"" already exists, skipping\\\\x00Fschemacmds.c\\\\x00L122\\\\x00RCreateSchemaCommand\\\\x00\\\\x00\\')']")
 (3,worker,"['DataRow(columncount=1,columns=[""C(length=0,value=b\\'\\')""])', 'CommandComplete(command=SELECT 1)', 'ReadyForQuery(state=in_transaction_block)']")
 (3,coordinator,"[""Query(query=SELECT worker_apply_shard_ddl_command (100923, 'test_test', 'CREATE TABLE test_test.test_table (id integer)'))""]")
 (3,worker,"[""RowDescription(fieldcount=1,fields=['F(name=worker_apply_shard_ddl_command,tableoid=0,colattrnum=0,typoid=2278,typlen=4,typmod=-1,format_code=0)'])"", 'DataRow(columncount=1,c
olumns=[""C(length=0,value=b\\'\\')""])', 'CommandComplete(command=SELECT 1)', 'ReadyForQuery(state=in_transaction_block)']")
 (0,coordinator,"[""Query(query=BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED;SELECT assign_distributed_transaction_id(0, XX, 'XXXX-XX-XX XX:XX:XX.XXXXXX-XX');)""]")
 (0,worker,"['CommandComplete(command=BEGIN)', ""RowDescription(fieldcount=1,fields=['F(name=assign_distributed_transaction_id,tableoid=0,colattrnum=0,typoid=2278,typlen=4,typmod=-1,format_c
ode=0)'])"", 'DataRow(columncount=1,columns=[""C(length=0,value=b\\'\\')""])', 'CommandComplete(command=SELECT 1)', 'ReadyForQuery(state=in_transaction_block)']")
*** HERE *** (0,coordinator,"[""Query(query=SELECT worker_apply_shard_ddl_command (100925, 'test_test', 'CREATE SCHEMA IF NOT EXISTS test_test AUTHORIZATION brian'))""]")
 (0,worker,"[""RowDescription(fieldcount=1,fields=['F(name=worker_apply_shard_ddl_command,tableoid=0,colattrnum=0,typoid=2278,typlen=4,typmod=-1,format_code=0)'])"", 'Backend(type=N,body=b\\
'SNOTICE\\\\x00VNOTICE\\\\x00C42P06\\\\x00Mschema ""test_test"" already exists, skipping\\\\x00Fschemacmds.c\\\\x00L122\\\\x00RCreateSchemaCommand\\\\x00\\\\x00\\')']")
 (0,worker,"['DataRow(columncount=1,columns=[""C(length=0,value=b\\'\\')""])', 'CommandComplete(command=SELECT 1)', 'ReadyForQuery(state=in_transaction_block)']")
 (0,coordinator,"[""Query(query=SELECT worker_apply_shard_ddl_command (100925, 'test_test', 'CREATE TABLE test_test.test_table (id integer)'))""]")
 (0,worker,"[""RowDescription(fieldcount=1,fields=['F(name=worker_apply_shard_ddl_command,tableoid=0,colattrnum=0,typoid=2278,typlen=4,typmod=-1,format_code=0)'])"", 'DataRow(columncount=1,c
olumns=[""C(length=0,value=b\\'\\')""])', 'CommandComplete(command=SELECT 1)', 'ReadyForQuery(state=in_transaction_block)']")
 (0,coordinator,"['Query(query=COPY test_test.test_table_100925 (id) FROM STDIN WITH (FORMAT BINARY))']")
 (0,worker,"[""Backend(type=G,body=b'\\\\x01\\\\x00\\\\x01\\\\x00\\\\x01')""]")
 (3,coordinator,"['Query(query=COPY test_test.test_table_100923 (id) FROM STDIN WITH (FORMAT BINARY))']")
 (3,worker,"[""Backend(type=G,body=b'\\\\x01\\\\x00\\\\x01\\\\x00\\\\x01')""]")
 (0,coordinator,"[""CopyData(data=b'PGCOPY\\\\n\\\\xff\\\\r\\\\n\\\\x00\\\\x00\\\\x00\\\\x00\\\\x00\\\\x00\\\\x00\\\\x00\\\\x00')"", ""CopyData(data=b'\\\\x00\\\\x01\\\\x00\\\\x00\\\\x00\\\\
x04\\\\x00\\\\x00\\\\x00\\\\x02')"", ""CopyData(data=b'\\\\xff\\\\xff')"", 'CopyDone()']")
 (0,worker,"['CommandComplete(command=COPY 1)', 'ReadyForQuery(state=in_transaction_block)']")
 (3,coordinator,"[""CopyData(data=b'PGCOPY\\\\n\\\\xff\\\\r\\\\n\\\\x00\\\\x00\\\\x00\\\\x00\\\\x00\\\\x00\\\\x00\\\\x00\\\\x00')"", ""CopyData(data=b'\\\\x00\\\\x01\\\\x00\\\\x00\\\\x00\\\\
x04\\\\x00\\\\x00\\\\x00\\\\x03')"", ""CopyData(data=b'\\\\xff\\\\xff')"", 'CopyDone()']")
 (3,worker,"['CommandComplete(command=COPY 1)', 'ReadyForQuery(state=in_transaction_block)']")
 (3,coordinator,"[""Query(query=PREPARE TRANSACTION 'citus_0_7238_40_55')""]")
 (0,coordinator,"[""Query(query=PREPARE TRANSACTION 'citus_0_7238_40_57')""]")
 (0,worker,"['CommandComplete(command=PREPARE TRANSACTION)', 'ReadyForQuery(state=idle)']")
 (3,worker,"['CommandComplete(command=PREPARE TRANSACTION)', 'ReadyForQuery(state=idle)']")
 (3,coordinator,"[""Query(query=COMMIT PREPARED 'citus_0_7238_40_55')""]")
 (0,coordinator,"[""Query(query=COMMIT PREPARED 'citus_0_7238_40_57')""]")
 (3,worker,"['CommandComplete(command=COMMIT PREPARED)', 'ReadyForQuery(state=idle)']")
 (0,worker,"['CommandComplete(command=COMMIT PREPARED)', 'ReadyForQuery(state=idle)']")
 (0,coordinator,"['Terminate()']")
```

We do it three times, this is some strange behavior! Probably worth opening a ticket for?

src/test/regress/expected/failure_create_distributed_table_non_empty.out
```diff
@@ -0,0 +1,520 @@
...
+ 
+(1 row)
+
+SELECT create_distributed_table('test_table', 'id');
+NOTICE:  Copying data from local table...
+ERROR:  server closed the connection unexpectedly
+	This probably means the server terminated abnormally
+	before or while processing the request.
+CONTEXT:  while executing command on localhost:57640
+SELECT citus.mitmproxy('conn.allow()');
```
lithp: also unnecessary

src/test/regress/expected/failure_create_distributed_table_non_empty.out
```diff
@@ -0,0 +1,520 @@
...
+-- shards will be created because we ignore cancel requests during the shard creation 
+-- Interrupts are hold in CreateShardsWithRoundRobinPolicy
+SELECT citus.mitmproxy('conn.onQuery(query="^BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED").cancel(' ||  pg_backend_pid() || ')');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT create_distributed_table('test_table', 'id');
+WARNING:  cancel requests are ignored during shard creation
```
lithp: Oh, it's strange that we disallow for just this part, if you try again a little later it works. Can we add that to this message? "please try again soon". Or... can we resurface the cancellation once shard creation finished?

src/test/regress/expected/failure_create_distributed_table_non_empty.out
```diff
@@ -0,0 +1,520 @@
...
+SELECT citus.mitmproxy('conn.onQuery(query="COPY").cancel(' ||  pg_backend_pid() || ')');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT create_distributed_table('test_table', 'id');
+NOTICE:  Copying data from local table...
+ERROR:  canceling statement due to user request
+SELECT citus.mitmproxy('conn.allow()');
```
lithp: this one too

src/test/regress/expected/failure_create_distributed_table_non_empty.out
```diff
@@ -0,0 +1,520 @@
...
+SELECT citus.mitmproxy('conn.onCommandComplete(command="COPY").cancel(' ||  pg_backend_pid() || ')');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT create_distributed_table('test_table', 'id');
+NOTICE:  Copying data from local table...
+ERROR:  canceling statement due to user request
+SELECT citus.mitmproxy('conn.allow()');
```
lithp: and this one

src/test/regress/expected/failure_create_distributed_table_non_empty.out
```diff
@@ -0,0 +1,520 @@
...
+
+SELECT count(*) FROM pg_dist_shard WHERE logicalrelid='create_distributed_table_non_empty_failure.test_table'::regclass;
+ count 
+-------
+     0
+(1 row)
+
+-- immediately cancel after we get prepare transaction complete
+-- to see if the command successfully rollbacked the created shards
+SELECT citus.mitmproxy('conn.onCommandComplete(command="PREPARE TRANSACTION").cancel(' ||  pg_backend_pid() || ')');
```
lithp: When writing failure tests you probably don't need to check *every* step, from the coordinator's point of view failing on `onQuery` and failing on `onCommandComplete` look the exact same.

src/test/regress/expected/failure_create_distributed_table_non_empty.out
```diff
@@ -0,0 +1,520 @@
...
+-- 
+CREATE SCHEMA create_distributed_table_non_empty_failure;
+SET search_path TO 'create_distributed_table_non_empty_failure';
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+-- we'll start with replication factor 1
```
lithp: It seems that you're using 2PC. You should put that somewhere here in the header (with a `SET` command) so that this test doesn't depend on the ordering of tests, and also as documentation. It'd be nice if the file name also included "2pc" somewhere in it, so we know that that's the only thing you're testing.

src/test/regress/expected/failure_create_distributed_table_non_empty.out
furkansahin: That's right. Because we make `EnsureSchemaExists` call before we start the transaction. Therefore, when one of the workers sent the command and the other couldn't, we create the schema in only one worker. I think this is a harmless bug. I don't think this causes a problem in anyway, that's why I did not report it.

src/test/regress/expected/failure_create_distributed_table_non_empty.out
furkansahin: As I understand, for `cancel`, there is a small amount of time that the coordinator gets aware of the `cancel` operation. During that time, coordinator sends the other `CREATE SCHEMA IF NOT EXISTS` command. I think the only reason we hit to that issue is that we make the `EnsureSchemaExists` call before starting the coordinated transaction.

src/test/regress/expected/failure_create_distributed_table_non_empty.out
furkansahin: Ahh, that's I missed. Okay, I will check in detail an open a ticket for this.

src/test/regress/expected/failure_create_distributed_table_non_empty.out
furkansahin: Or, we can even make the cancellations possible during the shard creation. I don't see any reason why we shouldn't. I will open a ticket for this as well.

src/test/regress/expected/failure_create_distributed_table_non_empty.out
furkansahin: OK :D

src/test/regress/expected/failure_create_distributed_table_non_empty.out
onderkalaci: Note that it is not a bug. Actually, it prevents some bugs, see #1637 and #1652. 

So, schema creation is always done one a separate connection and not inside the coordinated transaction. So, it's expected to see that the other worker has the schema. Note that, we're always using `"CREATE SCHEMA IF NOT EXISTS ...` so we can consider it as an idempotent operation.

src/test/regress/expected/failure_create_distributed_table_non_empty.out
onderkalaci: Hmm, this might be counted as a bug. It turns out that Citus calls `CHECK_FOR_INTERRUPTS` on `GetRemoteCommandResult()`. This means that the command has already been sent and (via `SendRemoteCommand`) and the schema has already been created when the cancel is processed.

**I think this is not a problem for queries that are inside the coordinated transaction because we'd already be rollbacking those. @furkansahin and @lithp do you agree with me on this?**

See the bt when I sent the `SIGINT` just before calling `ExecuteCriticalRemoteCommand()`

```SQL
Program received signal SIGINT, Interrupt.

Breakpoint 2, errstart (elevel=20, filename=0x10a48a678 "postgres.c", lineno=2997, funcname=0x10a48a938 "ProcessInterrupts", domain=0x0) at elog.c:251
251			if (CritSectionCount > 0)
(gdb) bt
#0  errstart (elevel=20, filename=0x10a48a678 "postgres.c", lineno=2997, funcname=0x10a48a938 "ProcessInterrupts", domain=0x0) at elog.c:251
#1  0x000000010a0d22be in ProcessInterrupts () at postgres.c:2995
#2  0x000000010a9ccd9d in FinishConnectionIO (connection=0x7f913183cd28, raiseInterrupts=1 '\001') at connection/remote_commands.c:673
#3  0x000000010a9cc091 in GetRemoteCommandResult (connection=0x7f913183cd28, raiseInterrupts=1 '\001') at connection/remote_commands.c:550
#4  0x000000010a9cc948 in ExecuteCriticalRemoteCommand (connection=0x7f913183cd28, command=0x7f913185f930 "CREATE SCHEMA IF NOT EXISTS x AUTHORIZATION onderkalaci") at connection/remote_commands.c:372
#5  0x000000010a9c1447 in EnsureSchemaExistsOnAllNodes (relationId=16824) at commands/create_distributed_table.c:862
#6  0x000000010a9c22bf in CreateHashDistributedTableShards (relationId=16824, colocatedTableId=16820, localTableEmpty=1 '\001') at commands/create_distributed_table.c:489
#7  0x000000010a9c0f53 in CreateDistributedTable (relationId=16824, distributionColumn=0x7f913101da40, distributionMethod=104 'h', colocateWithTableName=0x7f913101e498 "default", viaDeprecatedAPI=0 '\0') 
....
```

But, we should probably discuss this with @marcocitus to understand why `SendRemoteCommand` doesn't check for interrupts? I'm OK with opening an issue for this.

But, in any case, @furkansahin could you add a comment to the test?

src/test/regress/expected/failure_create_distributed_table_non_empty.out
onderkalaci: Yeah, we could open an issue. What I remembered related to that is we left it there because `GetTableDDLEvents()`, which is used in some other places as well, adds the `CREATE SCHEMA IF NOT EXISTS` for each shard creation. So, we should be careful while removing it, making sure that we don't break anyone who relies on that schema creation.

src/test/regress/expected/failure_create_distributed_table_non_empty.out
onderkalaci: See: https://github.com/citusdata/citus/pull/2279/files#r202302595

So, maybe add one more test with colocated tables? (and also open an issue as you've already mentioned)

src/test/regress/expected/failure_create_distributed_table_non_empty.out
furkansahin: opened the issue #2283

furkansahin:
@lithp I added 1pc tests as well, worth checking once more. Thanks!

src/test/regress/expected/failure_create_distributed_table_non_empty.out
lithp: > I think this is not a problem for queries that are inside the coordinated transaction because we'd already be rollbacking those. @furkansahin and @lithp do you agree with me on this?

Yes, I agree.

src/test/regress/expected/failure_create_distributed_table_non_empty.out
```diff
@@ -0,0 +1,890 @@
...
+(1 row)
+
+BEGIN;
+SELECT create_distributed_table('test_table', 'id');
+ create_distributed_table 
+--------------------------
+ 
+(1 row)
+
+ROLLBACK;
```
lithp: You've already written it but this kind of test probably isn't necessary in future tests, as long as _somewhere_ we test transaction management, the failure modes won't be any different if we fail a ROLLBACK in create_distributed_table and one in COPY.

src/test/regress/expected/failure_create_distributed_table_non_empty.out
```diff
@@ -0,0 +1,890 @@
...
+-----------
+ 
+(1 row)
+
+SELECT count(*) FROM pg_dist_shard WHERE logicalrelid='create_distributed_table_non_empty_failure.test_table'::regclass;
+ count 
+-------
+     0
+(1 row)
+
```
lithp: It might be nice to add a big bold comment here saying you've moved on and now you're testing what happens when you're creating co-located tables

lithp: This could maybe even be moved into a separate file?

src/test/regress/expected/failure_create_distributed_table_non_empty.out
```diff
@@ -0,0 +1,890 @@
...
+ 
+(1 row)
+
+SELECT run_command_on_workers($$SELECT count(*) FROM information_schema.tables WHERE table_schema = 'create_distributed_table_non_empty_failure' and table_name LIKE 'test_table%'$$);
+ run_command_on_workers 
+------------------------
+ (localhost,57637,t,0)
+ (localhost,57640,t,0)
+(2 rows)
+
```
lithp: Why did you only test failures during `BEGIN TRANSACTION` when distributing colocated tables? Is there a code path which might be messed up by a failure during one of the CREATE TABLE's?

src/test/regress/expected/failure_create_distributed_table_non_empty.out
```diff
@@ -0,0 +1,890 @@
...
+-- kill as soon as the coordinator sends CREATE SCHEMA
+SELECT citus.mitmproxy('conn.onQuery(query="^CREATE SCHEMA").kill()');
```
lithp: You probably don't need this test, since the first `CREATE SCHEMA` happens before the transaction even begins, the behavior will be the exact same under 1PC and 2PC.

src/test/regress/expected/failure_create_distributed_table_non_empty.out
```diff
@@ -0,0 +1,890 @@
...
+------------------------
+ (localhost,57637,t,0)
+ (localhost,57640,t,0)
+(2 rows)
+
+SELECT run_command_on_workers($$SELECT count(*) FROM information_schema.schemata WHERE schema_name = 'create_distributed_table_non_empty_failure'$$);
+ run_command_on_workers 
+------------------------
+ (localhost,57637,t,1)
+ (localhost,57640,t,1)
```
lithp: This is a little confusing, you told the proxy to kill() but the SCHEMA still appears? I think it's because the SCHEMA already existed before you ran this test. You should probably just remove this check for the schema.

src/test/regress/expected/failure_create_distributed_table_non_empty.out
```diff
@@ -0,0 +1,890 @@
...
+
+SELECT create_distributed_table('test_table', 'id');
+ERROR:  canceling statement due to user request
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT citus.mitmproxy('conn.allow()');
```
lithp: You don't need two of these :)

src/test/regress/expected/failure_create_distributed_table_non_empty.out
```diff
@@ -0,0 +1,890 @@
...
+------------------------
+ (localhost,57637,t,1)
+ (localhost,57640,t,1)
+(2 rows)
+
+-- cancel as soon as the coordinator sends CREATE SCHEMA
+-- Note: Schema should be created in workers because Citus
+-- does not check for interrupts until GetRemoteCommandResult is called.
+-- Since we already sent the command at this stage, the schemas get created in workers
+SELECT citus.mitmproxy('conn.onQuery(query="^CREATE SCHEMA").cancel(' ||  pg_backend_pid() || ')');
```
lithp: Again, you probably don't need this test. Behavior will be the same under 1PC and 2PC.

src/test/regress/expected/failure_create_distributed_table_non_empty.out
```diff
@@ -0,0 +1,890 @@
...
+(1 row)
+
+ROLLBACK;
+SELECT count(*) FROM pg_dist_shard WHERE logicalrelid='create_distributed_table_non_empty_failure.test_table'::regclass;
+ count 
+-------
+     0
+(1 row)
+
+-- cancel as soon as the coordinator sends ROLLBACK
```
lithp: Why no test for `COMMIT`?

furkansahin: added

src/test/regress/expected/failure_create_distributed_table_non_empty.out
furkansahin: I want to keep this test here since it is also related to `create_distributed_table`

src/test/regress/expected/failure_create_distributed_table_non_empty.out
furkansahin: Some more tests are added for `COPY` and `CREATE TABLE`.

src/test/regress/expected/failure_create_distributed_table_non_empty.out
furkansahin: Removed

src/test/regress/expected/failure_create_distributed_table_non_empty.out
furkansahin: Removed

furkansahin:
Review is applied, the PR seems ready for one last look or merge? :)

furkansahin:
hey @lithp , should I rebase and merge or you are going to have some more input?

onderkalaci:
Hey @furkansahin could you please rebase to the master and merge!

Thanks!

PR #2281, opened 2018-07-12
Add create_reference_table failure tests

velioglu:
Add failure tests for creating reference table

### DIFF
--- a/src/test/regress/expected/failure_create_reference_table.out
+++ b/src/test/regress/expected/failure_create_reference_table.out
@@ -1,0 +1,247 @@
+--
+-- Failure tests for creating reference table
+--
+CREATE SCHEMA failure_reference_table;
+SET search_path TO 'failure_reference_table';
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+CREATE TABLE ref_table(id int);
+INSERT INTO ref_table VALUES(1),(2),(3);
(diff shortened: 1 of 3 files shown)


src/test/regress/expected/failure_create_reference_table.out
```diff
@@ -0,0 +1,261 @@
...
+ERROR:  connection error: localhost:57640
+DETAIL:  connection not open
+SELECT count(*) FROM pg_dist_shard_placement;
+ count 
+-------
+     0
+(1 row)
+
+-- Cancel after creating transaction on worker node
+SELECT citus.mitmproxy('conn.allow()');
```
lithp: You don't need to do `conn.allow()` if you're going to immediately call `citus.mitmproxy` again. Every time you call `citus.mitmpeoxy()` the proxy re-configures itself and throws away the previous configuration.

src/test/regress/expected/failure_create_reference_table.out
```diff
@@ -0,0 +1,261 @@
...
+SELECT create_reference_table('ref_table');
+ERROR:  canceling statement due to user request
+SELECT count(*) FROM pg_dist_shard_placement;
+ count 
+-------
+     0
+(1 row)
+
+-- Kill after copying data to worker node
+SELECT citus.mitmproxy('conn.allow()');
```
lithp: Same here, you don't need to call this.

PR #2296, opened 2018-07-20
ALTER TABLE %s ADD COLUMN %s [constraint] constraint checks are implemented

### LINKED ISSUE #2286: foreign table name is not resolved correctly when adding a new column with a foreign key
Repro Steps
```sql
create table rr(a int primary key);
select create_reference_table('rr');
create table tt(a int, b int);
select create_distributed_table('tt', 'a');
alter table tt add column c int references rr(a) on delete set null;
```
Result
```
ERROR:  relation "rr" does not exist
```
Worker log has
```
ERROR:  relation "rr" does not exist
SELECT worker_apply_shard_ddl_command (102300, 'public', 'alter table tt add column c int references rr(a) on delete set null;')
```

### LINKED ISSUE #2289: `ALTER TABLE %s ADD COLUMN` constraint check is missing
I realized this bug while working on #2286
We do not support some of the constraints on distributed table and we check them while `ALTER TABLE %s ADD CONSTRAINT` or `create_distributed_table`. However, some of the constraints can also be added via `ALTER TABLE %s ADD COLUMN %s`, in which case, we do not check anything. Thus, user is able to create some of the constraints which s/he shouldn't. For example,
```
postgres=# CREATE TABLE t(i int);
CREATE TABLE
Time: 5.067 ms
postgres=# SELECT create_distributed_table('t', 'i');
 create_distributed_table
--------------------------

(1 row)

Time: 83.607 ms
postgres=# -- we do not support UNIQUE constraints on non-distribution columns
postgres=# ALTER TABLE t ADD COLUMN j int UNIQUE;
ALTER TABLE
Time: 152.560 ms
postgres=# \d+ t;
                                     Table "public.t"
 Column |  Type   | Collation | Nullable | Default | Storage | Stats target | Description
--------+---------+-----------+----------+---------+---------+--------------+-------------
 i      | integer |           |          |         | plain   |              |
 j      | integer |           |          |         | plain   |              |
Indexes:
    "t_j_key" UNIQUE CONSTRAINT, btree (j)
```

furkansahin:
DESCRIPTION: ALTER TABLE ADD COLUMN constraint check is fixed
fixes #2289 
fixes #2286 

According to the given issues,
 * We do not give a proper error for `ALTER TABLE %s ADD COLUMN %s REFERENCES %s`
    * which we cannot support the creation of fk since a name is not given to the constraint. The reason is the same why we do not allow constraints without a name.
 * We allow people to introduce some constraints that we generally do not support
    * since we never check for the constraints on `alter table add column` we let them to add `unique`, `primary key` constraints to non-distribution columns. 

This PR fixes these issues by detecting if `add column` has a constraint which we cannot execute unless a name is given as in `ALTER TABLE ADD CONSTRAINT` or we mustn't allow for that constraint at all.

### DIFF
--- a/src/backend/distributed/executor/multi_utility.c
+++ b/src/backend/distributed/executor/multi_utility.c
@@ -173,6 +173,8 @@ static void ProcessDropTableStmt(DropStmt *dropTableStatement);
 static void ProcessDropSchemaStmt(DropStmt *dropSchemaStatement);
 static void InvalidateForeignKeyGraphForDDL(void);
 
+static void ErrorUnsupportedAlterTableAddColumn(Oid relationId, AlterTableCmd *command,
+												Constraint *constraint);
 
 /*
  * We need to run some of the commands sequentially if there is a foreign constraint
@@ -618,6 +620,42 @@ multi_ProcessUtility(PlannedStmt *pstmt,
 						InvalidateForeignKeyGraph();
 					}
 				}
+				else if (alterTableType == AT_AddColumn)
+				{
+					List *columnConstraints = NIL;
+					ListCell *columnConstraint = NULL;
+					Oid relationId = InvalidOid;
+					LOCKMODE lockmode = NoLock;
+
+					ColumnDef *columnDefinition = (ColumnDef *) command->def;
+					columnConstraints = columnDefinition->constraints;
+					if (columnConstraints)
+					{
+						ErrorIfUnsupportedAlterAddConstraintStmt(alterTableStatement);
+					}
+
+					lockmode = AlterTableGetLockLevel(alterTableStatement->cmds);
+					relationId = AlterTableLookupRelation(alterTableStatement, lockmode);
+					if (!OidIsValid(relationId))
+					{
+						continue;
+					}
+
+					foreach(columnConstraint, columnConstraints)
+					{
+						Constraint *constraint = (Constraint *) lfirst(columnConstraint);
+
+						if (constraint->conname == NULL &&
+							(constraint->contype == CONSTR_PRIMARY ||
+							 constraint->contype == CONSTR_UNIQUE ||
+							 constraint->contype == CONSTR_FOREIGN ||
+							 constraint->contype == CONSTR_CHECK))
+						{
+							ErrorUnsupportedAlterTableAddColumn(relationId, command,
+																constraint);
+						}
+					}
+				}
 			}
 		}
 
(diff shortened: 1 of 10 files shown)


furkansahin:
Hey @marcocitus, thanks for bringing this up. After your comment on the branch I was working on this issue. I tried to error out when there is a constraint given with `ALTER TABLE %s ADD COLUMN` but this made creating columns with `DEFAULT`, 'NOT NULL` impossible. Then I dived in more to understand which commands we cannot allow and came up with `UNIQUE`, `PRIMARY` and `FOREIGN`. I am going to be testing this branch more. In the mean time, please give me feedback on my approach if you can find time.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -613,11 +613,48 @@ multi_ProcessUtility(PlannedStmt *pstmt,
...
+					{
+						continue;
+					}
+
+					foreach(columnConstraint, columnConstraints)
+					{
+						Constraint *constraint = (Constraint *) lfirst(columnConstraint);
+
+						if (constraint->conname == NULL &&
+								(constraint->contype == CONSTR_PRIMARY ||
```
mtuncer: is this better than allowing a set of commands that are known to work ?

furkansahin: Which commands that we should allow ? The commands that we avoid here are the ones that we can create but cannot drop. I think not allowing such commands is better than allowing the creation of nondestroyable constraints in a straightforward manner

mtuncer: tt2 is a reference table, it has column b of bigserial type

following is a valid for a regular table, but fails with a worker error. 

```sql
 alter table tt2 add column d  bigint default nextval('tt2_b_seq'::regclass);
ERROR:  relation "tt2_b_seq" does not exist
CONTEXT:  while executing command on localhost:9703
```
It is not a big issue and nowhere real being showstopper. just a nuisance;

src/backend/distributed/executor/multi_utility.c
```diff
@@ -613,11 +613,48 @@ multi_ProcessUtility(PlannedStmt *pstmt,
...
+								 constraint->contype == CONSTR_UNIQUE ||
+								 constraint->contype == CONSTR_FOREIGN))
+						{
+							ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+											errmsg("cannot execute ADD COLUMN command with "
+												   "PRIMARY KEY, UNIQUE and FOREIGN constraints"),
+											errhint("You can issue each command separately")));
```
marcocitus: I don't think it's clear from the hint what you need to do. 

Maybe "Add the column without a constraint and then run an ALTER TABLE .. ADD CONSTRAINT command"

We could also add some more context, e.g. `errdetail("Adding a column with a constraint in one command is not supported because all constraints in Citus must have explicit names")`

src/backend/distributed/executor/multi_utility.c
```diff
@@ -613,11 +613,48 @@ multi_ProcessUtility(PlannedStmt *pstmt,
...
+					}
+
+					foreach(columnConstraint, columnConstraints)
+					{
+						Constraint *constraint = (Constraint *) lfirst(columnConstraint);
+
+						if (constraint->conname == NULL &&
+								(constraint->contype == CONSTR_PRIMARY ||
+								 constraint->contype == CONSTR_UNIQUE ||
+								 constraint->contype == CONSTR_FOREIGN))
```
marcocitus: this probably also applies to check constraints

PR #2297, opened 2018-07-20
Fix/insert select onconflict

### LINKED ISSUE #2288: INSERT...SELECT...ON CONFLICT... allows modifying partition column
We don't apply the usual ON CONFLICT checks to distributed INSERT...SELECT queries.
```sql
CREATE TABLE test (x int primary key, y int);                                                                                                                           
CREATE TABLE test2 (x int, y int);
SELECT create_distributed_table('test2', 'x');                                                                                                                          
INSERT INTO test2 VALUES (1,2);

-- First insert, no conflict
INSERT INTO test SELECT * FROM test2 ON CONFLICT (x) DO UPDATE SET x = 3 RETURNING *;

SELECT * FROM test WHERE x = 1;
 x │ y 
───┼───
 1 │ 2
(1 row)

-- Conflicts on the primary key, so update x to 3
INSERT INTO test SELECT * FROM test2 ON CONFLICT (x) DO UPDATE SET x = 3;

-- Row is still there
SELECT * FROM test;
 x │ y 
───┼───
 3 │ 2
(1 row)

-- but it's in the wrong shard
SELECT * FROM test WHERE x = 3;
 x │ y 
───┼───
(0 rows)

```

thanodnl:
DESCRIPTION: fixes bug where INSERT ... SELECT was allowed to update distribution column

Fixes #2288

### DIFF
--- a/src/backend/distributed/planner/insert_select_planner.c
+++ b/src/backend/distributed/planner/insert_select_planner.c
@@ -176,6 +176,14 @@ CreateInsertSelectPlan(Query *originalQuery,
 					   PlannerRestrictionContext *plannerRestrictionContext)
 {
 	DistributedPlan *distributedPlan = NULL;
+	DeferredErrorMessage *deferredError = NULL;
+
+	deferredError = ErrorIfOnConflictNotSupported(originalQuery);
+	if (deferredError != NULL)
+	{
+		/* raising the error as there is no possible solution for the unsupported on conflict statements */
+		RaiseDeferredError(deferredError, ERROR);
+	}
 
 	distributedPlan = CreateDistributedInsertSelectPlan(originalQuery,
 														plannerRestrictionContext);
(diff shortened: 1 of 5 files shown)


src/backend/distributed/planner/insert_select_planner.c
```diff
@@ -176,6 +176,15 @@ CreateInsertSelectPlan(Query *originalQuery,
...
 {
 	DistributedPlan *distributedPlan = NULL;
+	DeferredErrorMessage *deferredError = NULL;
+
+	deferredError = ErrorIfOnConflictNotSupported(originalQuery);
+	if (deferredError != NULL)
+	{
+		distributedPlan = CitusMakeNode(DistributedPlan);
+		distributedPlan->planningError = deferredError;
+		return distributedPlan;
```
marcocitus: It may be worth throwing a hard error here since no amount of recursive planning is going to solve this issue right now.

src/backend/distributed/planner/insert_select_planner.c
```diff
@@ -181,9 +181,7 @@ CreateInsertSelectPlan(Query *originalQuery,
 	deferredError = ErrorIfOnConflictNotSupported(originalQuery);
 	if (deferredError != NULL)
 	{
-		distributedPlan = CitusMakeNode(DistributedPlan);
-		distributedPlan->planningError = deferredError;
-		return distributedPlan;
+		RaiseDeferredError(deferredError, ERROR);
```
marcocitus: maybe add a comment on why we're raising here, since it's not the common thing to do

PR #2299, opened 2018-07-20
Don't try to check unopened connection in EXEC_TASK_FAILED state

marcocitus:
DESCRIPTION: Fix real-time query segfault during online shard moves

### DIFF
--- a/src/backend/distributed/executor/multi_client_executor.c
+++ b/src/backend/distributed/executor/multi_client_executor.c
@@ -294,6 +294,14 @@ MultiClientConnectPoll(int32 connectionId)
 MultiConnection *
 MultiClientGetConnection(int32 connectionId)
 {
+	if (connectionId == INVALID_CONNECTION_ID)
+	{
+		return NULL;
+	}
+
+	Assert(connectionId >= 0);
+	Assert(connectionId < MAX_CONNECTION_COUNT);
+
 	return ClientConnectionArray[connectionId];
 }
 
--- a/src/backend/distributed/executor/multi_real_time_executor.c
+++ b/src/backend/distributed/executor/multi_real_time_executor.c
@@ -425,6 +425,7 @@ ManageTaskExecution(Task *task, TaskExecution *taskExecution,
 		case EXEC_TASK_FAILED:
 		{
 			bool raiseError = false;
+			bool isCritical = false;
 
 			/*
 			 * On task failure, we close the connection. We also reset our execution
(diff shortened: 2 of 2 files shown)


marcocitus:
Reconfirmed that this fixes the issue by running real-time queries via pgbench against a cluster that is running the rebalancer.

src/backend/distributed/executor/multi_real_time_executor.c
```diff
@@ -433,8 +434,29 @@ ManageTaskExecution(Task *task, TaskExecution *taskExecution,
 			 * on this node again.
 			 */
 			int32 connectionId = connectionIdArray[currentIndex];
-			MultiConnection *connection = MultiClientGetConnection(connectionId);
-			bool isCritical = connection->remoteTransaction.transactionCritical;
+			MultiConnection *connection = NULL;
+
+			/* try next worker node next time we look at this task */
+			taskStatusArray[currentIndex] = EXEC_TASK_CONNECT_START;
+			AdjustStateForFailure(taskExecution);
```
jasonmp85: This was previously only called if `isCritical` was false; I was concerned that the side effects it produces might result in some change in behavior, but since the only new behavior is calling it when `isCritical` is (eventually) true, and when `isCritical` is true we're going to fail-fast anyhow, this is only a no-op as far as behavior is concerned (i.e. the old code path was essentially an optimization of sorts; there is no effect calling it during a critical failure).

Chatted with @marcocitus over Slack to confirm this; otherwise the changes in this function are only changes in ordering.

src/backend/distributed/executor/multi_client_executor.c
```diff
@@ -294,6 +294,14 @@ MultiClientConnectPoll(int32 connectionId)
 MultiConnection *
 MultiClientGetConnection(int32 connectionId)
 {
+	if (connectionId == MultiClientGetConnection)
```
jasonmp85: Uuuuhhhh?

jasonmp85:
@marcocitus if you have time, can you look into why the branch build (assert-enabled) is failing? I can reproduce the failure locally but am handling some other stuff right now.

jasonmp85:
(otherwise I can look into it after helping people test some things)

marcocitus:
> if you have time, can you look into why the branch build (assert-enabled) is failing? I can reproduce the failure locally but am handling some other stuff right now.

Not quite sure what was going on. I could initially reproduce it, and then redid the changes in very small steps to figure out which line caused the problem, but can no longer reproduce it.

sumedhpathak:
@marcocitus @jasonmp85 Status on this PR? Is it looking close to merge?

jasonmp85:
> Status on this PR? Is it looking close to merge?

Yeah, didn't see Marco's latest comment. I rebased and pushed up and will merge after checks pass. Still kind of concerned that we both could initially reproduce the error and haven't changed anything to address it.

Issue #2270, opened 2018-07-10
Travis random failure with Failure testing framework


furkansahin:
We time to time hit to an issue in travis with the failure testing automation framework.
The out we see is the following;
```
============== dropping database "regression"         ==============
NOTICE:  database "regression" does not exist, skipping
DROP DATABASE
============== creating database "regression"         ==============
CREATE DATABASE
ALTER DATABASE
============== installing citus                       ==============
CREATE EXTENSION
============== running regression test queries        ==============
test failure_test_helpers     ... ok
test failure_setup            ... 

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

The build has been terminated

```

lithp:
Hmmmm, interesting. This sounds hard to debug.

This seems likely to have been caused by the fifo, you'll block forever when trying to open the fifo if the other side never opens it. So, some theories:
- `mitmproxy` failed to start for some reason (the port was already bound?) and we block forever when attempting to talk to it.
- `mitmproxy` has crashed for some reason. It accepted the command but crashed and never responded?
- Travis runs multiple tests in the same machine? Very weird things happen if you have two proxies and one test process, the proxies often end up spin-looping and talking to each other and blocking the test process.

Some things which would be useful:
- It'd be nice to know which processes are running when we finally give up. `lsof -p `pgrep mitmdump` would be useful.
- It's generally bad that we block forever when attempting to open a fifo. I wish there was a way of setting a timeout from inside sql, but to add a timeout we probably need to write a c-function or switch away from using fifos. (fifos are really nice because they give us free synchronization, by the time the proxy has responded we know for sure that it's configured, but if the cost is that we often "oversynchronize" and deadlock then maybe it's worth switching to something else.

onderkalaci:
This came up several times again :/

mtuncer:
we should really fix this soon. It is way beyond little annoyance now. I would say more annoying than appveyor.

metdos:
@velioglu noted that this also currently fails nightly-merge-build.

@sumedhpathak, @lithp, we may need to disable failure testing framework until we fix this bug.

lithp:
[...]

PR #2300, opened 2018-07-20
Prevent failure tests from hanging by using a port outside the ephemeral port range

lithp:
After running multiple series of tests, I have decided the most likely reason that our tests are hanging is that we were using port 57640, which was sometimes not available. (During a previous test we used that port when opening a connection to the coordinator / the workers, and by the time failure testing started port 57649 was still in the TIME_WAIT state)

This PR addresses #2270 and makes a few changes:
- It uses port 9060 when starting mitmdump, a port which should be unused.
- Updates all the tests to expect the new port, I wish this part wasn't so noisy.
- The `pg_regress_multi.pl` script catches `SIGCHLD` when `mitmdump` dies and kills itself, which should prevent the tests from continuing while `mitmdump` is not running. This will prevent future hangs.
- `pg_regress_multi.pl` has a little more logging, and will also refuse to run the tests if something is already running on the port which `mitmdump` uses.
- It adds fifo-related logging to `mitmdump`, so `proxy.output` has more output which will be useful if this kind of problem appears in the future.


src/test/regress/wrap_test.sh
```diff
@@ -0,0 +1,39 @@
...
+  echo "it's been 60 seconds and regress is still running"
+
+  printf 'is mitmdump running?\n'
+  ps aux | grep mitm
+
+  printf '\nthe proxy output:\n'
+  cat proxy.output
+
+  # what is mitmdump doing?
+  sudo timeout 1 strace -p `pgrep mitmdump`
```
jasonmp85: (a) not a fan of the `sudo`
(b) no `strace` on macOS
(c) no `pgrep` on macOS

lithp: Hmmm, what do you suggest?

This is not intended to work on macOS, it's just there to debug the Travis failures. Hopefully it'll be gone by next week.

Do you know of a way to detect osx from inside a Makefile? I can only wrap if we're running on linux.

jasonmp85: Oh, so this is just for debugging? Maybe hide it behind an environment variable then? I dunno.

metdos:
Hey @jasonmp85, are you reviewing this or you were just taking a look?

jasonmp85:
@metdos reviewing; had to deal with KPI debugging and release stuff yesterday, but I'll be looking into this more seriously today.

src/test/regress/wrap_test.sh
lithp: Just pushed a commit which hides it behind an env variable, so the wrapper is only used on Travis and won't be hit when running things locally. That should make it safer to push to master.

lithp:
Looks like this caught one: https://travis-ci.org/citusdata/citus/jobs/406400622

`Error starting proxy server: OSError(98, 'Address already in use')`

lithp:
That's pretty reasonable output, if the server fails to start it makes sense that we'd block forever on our attempt to talk to it. 
1) Why is something else on port 57640?
2) If the proxy does not exist we should notice and not attempt to talk to it.

lithp:
I've added a change which will notice that the port is already taken and dump out the name of the running process (well, I call lsof which does the dumping). There's a race condition there, so I also notice when mitmdump has failed and abort the tests (so we don't reach the deadlock) and emit the proxy's logfile.

lithp:
The builds have succeeded (4) times.

lithp:
I got it to hang but I'm very confused: https://travis-ci.org/citusdata/citus/jobs/407835794

- Nothing was already on our port when we attempted to spawn `mitmdump`, otherwise `pg_regress_multi.pl` would have thrown an error.
- `mitmdump` didn't die, because `pg_regress_multi.pl` would have caught the signal and quit
- We blocked forever when attmpting to talk to the fifo, indicating that `mitmdump` had not sent an `open()`, or died before we started `failure_setup`.
- mitmdump complained that "address already in use"
- mitmdump was not running when we attempted to strace it.
- When I've run mitmdump locally, if it complains that "address already in use" it quits.

onderkalaci:
You might already see this issue on the mitmproxy: https://github.com/mitmproxy/mitmproxy/issues/782 which actually doesn't help us?

> I got it to hang but I'm very confused: https://travis-ci.org/citusdata/citus/jobs/407835794

jasonmp85:
This came up on HN earlier today and I'm dropping it here as a reference to myself to remember to standardize how we install mitmproxy both as developers and on Travis: https://tech.instacart.com/freezing-pythons-dependency-hell-in-2018-f1076d625241

lithp:
I accidentally committed a change which ran the failure tests first and the tests passed 16 times before I realized the changed it back. Soon after changing it back I got another error: https://travis-ci.org/citusdata/citus/jobs/408255096

Mitmdump immediately failed because the address was already in use...

lithp:
Okay, I added some logging and `TIME_WAIT` indeed seems to be the issue ([this log output](https://travis-ci.org/citusdata/citus/jobs/408279350) proves it). During the earlier tests we use up a bunch of the ephemeral ports, and sometimes, by chance, we happen to use 57640. This locks the port for a short period of time, causing mitmproxy to fail when it starts up.

lithp:
@jasonmp85 I think I've resolved the issue, I've updated the PR description. This is ready for you to review!

lithp:
@onderkalaci I know a couple people in Istanbul have piped up and said this is hurting development velocity, if you want to review it and don't have any suggested changes feel free to merge this.

lithp:
Also, @jasonmp85, cool link but this might not be the best place to leave that link. This has a lower chance of getting lost if you open a ticket?

They mention pipenv once but I think it solves all the problems they complained about? Personally, I think moving to doing all our builds inside docker would be a big improvement.

jasonmp85:
> This has a lower chance of getting lost if you open a ticket?

Yeah but I was actually going to read it _today_ in relation to my testing this PR (to help me figure out whether I have `mitmproxy` set up right in a way I'm unlikely to screw up).

> They mention pipenv once but I think it solves all the problems they complained about?

No clue… as a Ruby person I'd mostly learned all the crazy thing to manage that environment but the Python stuff is a whole other world I need to get right. I have mitmproxy installed via Homebrew, which is probably… not great.

> Personally, I think moving our builds to docker would be a big improvement.

I had the same thought today, though I do think there's some robustness to our _ad hoc_ heterogenous setup.

src/test/regress/sql/failure_setup.sql
```diff
@@ -2,4 +2,4 @@ SELECT citus.mitmproxy('conn.allow()');
 
 -- add the workers
 SELECT master_add_node('localhost', :worker_1_port);  -- the second worker
-SELECT master_add_node('localhost', :worker_2_port + 2);  -- the first worker, behind a mitmproxy
+SELECT master_add_node('localhost', :worker_2_proxy_port);  -- the first worker, behind a mitmproxy
```
jasonmp85: Why does this say _the first worker_ when it's `worker2`?

src/test/regress/pg_regress_multi.pl
```diff
@@ -560,6 +577,20 @@ ()
     die "a file already exists at $mitmFifoPath, delete it before trying again";
   }
 
+  system("lsof -i :$mitmPort");
+  if (! $?) {
+    die "cannot start mitmproxy because a process already exists on port $mitmPort";
+  }
+
+  if ($Config{osname} eq "linux")
```
jasonmp85: How did this ever work in Windows? Does `netstat` do anything there, or am I missing something?

src/test/regress/pg_regress_multi.pl
```diff
@@ -560,6 +577,20 @@ ()
     die "a file already exists at $mitmFifoPath, delete it before trying again";
   }
 
+  system("lsof -i :$mitmPort");
```
jasonmp85: (This isn't intended to work on Windows, right?

src/test/regress/sql/failure_setup.sql
lithp: oops

src/test/regress/pg_regress_multi.pl
lithp: nope, this doesn't work on Windows

PR #2305, opened 2018-07-21
Prototype: INSERT...SELECT via coordinator with ON CONFLICT/RETURNING

marcocitus:
This PR implements ON CONFLICT and RETURNING support for INSERT..SELECT commands which cannot be pushed down. For those INSERT...SELECT commands the coordinator plans and executes the SELECT separately and writes the results a distributed table through by doing a distributed COPY.

Because COPY does not have ON CONFLICT or RETURNING semantics, these features were so far unavailable. However, this PR works around that by reusing the intermediate result infrastructure from https://github.com/citusdata/citus/pull/1829  . It first copies into a set of intermediate results that are partitioned in the same way as the destination table, and then performs a distributed INSERT..SELECT from intermediate result in the corresponding shard.

As it turns out, this only requires minimal changes to the COPY logic. Instead of sending `COPY distributed_table_<shardid> FROM STDIN WITH (format binary)` commands to workers, we simply send  `COPY insert_select_<planid>_<shardid> FROM STDIN WITH (format result)` commands, which has the effect that the rows are not written to the table, but instead written to a set of intermediate results that are named according to the shard into which they should be inserted. 

To construct the INSERT..SELECT queries, we can reuse the infrastructure for replacing CTEs/subqueries with `read_intermediate_result` calls. The bulk of the logic in this PR is in constructing a custom task list for the final INSERT..SELECT. We currently put this task list in a `workerJob` in the `DistributedPlan`, which is otherwise unused for INSERT...SELECT via the coordinator. 

Remaining items:
[ ] Find a clean approach for the hack in `TwoPhaseInsertSelectTaskList` to set the column types in the `read_intermediate_result` call, consider collation etc.
[ ] Consider making the first step a subplan of the second step (optional)
[ ] Careful testing with CTEs/recursively planned subqueries
[ ] Regression tests

### DIFF
--- a/src/backend/distributed/executor/insert_select_executor.c
+++ b/src/backend/distributed/executor/insert_select_executor.c
@@ -16,6 +16,7 @@
 #include "distributed/multi_executor.h"
 #include "distributed/multi_partitioning_utils.h"
 #include "distributed/multi_physical_planner.h"
+#include "distributed/multi_router_executor.h"
 #include "distributed/distributed_planner.h"
 #include "distributed/relation_access_tracking.h"
 #include "distributed/resource_lock.h"
@@ -35,8 +36,9 @@
 #include "utils/snapmgr.h"
 
 
-static void ExecuteSelectIntoRelation(Oid targetRelationId, List *insertTargetList,
-									  Query *selectQuery, EState *executorState);
+static HTAB * ExecuteSelectIntoRelation(Oid targetRelationId, List *insertTargetList,
+										Query *selectQuery, EState *executorState,
+										char *intermediateResultIdPrefix);
 
 
 /*
@@ -58,6 +60,8 @@ CoordinatorInsertSelectExecScan(CustomScanState *node)
 		Query *selectQuery = distributedPlan->insertSelectSubquery;
 		List *insertTargetList = distributedPlan->insertTargetList;
 		Oid targetRelationId = distributedPlan->targetRelationId;
+		char *intermediateResultIdPrefix = distributedPlan->intermediateResultIdPrefix;
+		HTAB *shardConnectionsHash = NULL;
 
 		ereport(DEBUG1, (errmsg("Collecting INSERT ... SELECT results on coordinator")));
 
@@ -71,8 +75,52 @@ CoordinatorInsertSelectExecScan(CustomScanState *node)
 			LockPartitionRelations(targetRelationId, RowExclusiveLock);
 		}
 
-		ExecuteSelectIntoRelation(targetRelationId, insertTargetList, selectQuery,
-								  executorState);
+		shardConnectionsHash = ExecuteSelectIntoRelation(targetRelationId,
+														 insertTargetList, selectQuery,
+														 executorState,
+														 intermediateResultIdPrefix);
+
+		if (distributedPlan->workerJob != NULL)
+		{
+			/*
+			 * If we also have a workerJob that means there is a second step
+			 * to the INSERT...SELECT. This happens when there is a RETURNING
+			 * or ON CONFLICT clause which is implemented as a separate
+			 * distributed INSERT...SELECT from a set of intermediate results
+			 * to the target relation.
+			 */
+			Job *workerJob = distributedPlan->workerJob;
+			ListCell *taskCell = NULL;
+			List *taskList = workerJob->taskList;
+			List *prunedTaskList = NIL;
+			bool hasReturning = distributedPlan->hasReturning;
+			bool isModificationQuery = true;
+
+			/*
+			 * We cannot actually execute INSERT...SELECT tasks that read from
+			 * intermediate results that weren't created because no rows were
+			 * written to them. Prune those tasks out by only including tasks
+			 * on shards with connections.
+			 */
+			foreach(taskCell, taskList)
+			{
+				Task *task = (Task *) lfirst(taskCell);
+				uint64 shardId = task->anchorShardId;
+				bool shardModified = false;
+
+				hash_search(shardConnectionsHash, &shardId, HASH_FIND, &shardModified);
+				if (shardModified)
+				{
+					prunedTaskList = lappend(prunedTaskList, task);
+				}
+			}
+
+			if (prunedTaskList != NIL)
+			{
+				ExecuteMultipleTasks(scanState, prunedTaskList, isModificationQuery,
+									 hasReturning);
+			}
+		}
 
 		scanState->finishedRemoteScan = true;
 	}
@@ -87,10 +135,18 @@ CoordinatorInsertSelectExecScan(CustomScanState *node)
  * ExecuteSelectIntoRelation executes given SELECT query and inserts the
  * results into the target relation, which is assumed to be a distributed
  * table.
+ *
+ * If intermediateResultIdPrefix is not NULL, the data is instead written
+ * to a set of intermediate results that are co-located with the table
+ * for further processing (ON CONFLICT).
+ *
+ * ExecuteSelectIntoRelation returns the hash of connections that were
+ * used to insert into the target relation.
  */
-static void
+static HTAB *
 ExecuteSelectIntoRelation(Oid targetRelationId, List *insertTargetList,
-						  Query *selectQuery, EState *executorState)
+						  Query *selectQuery, EState *executorState,
+						  char *intermediateResultIdPrefix)
 {
 	ParamListInfo paramListInfo = executorState->es_param_list_info;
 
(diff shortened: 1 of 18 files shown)


src/backend/distributed/planner/insert_select_planner.c
```diff
@@ -1408,6 +1408,7 @@ TwoPhaseInsertSelectTaskList(Oid targetRelationId, Query *insertSelectQuery,
 		 * Generate a query string for the query that inserts into a shard and reads
 		 * from an intermediate result.
 		 */
+		insertResultQuery->cteList = NIL;
```
marcocitus: maybe add a comment on why we need to do this

velioglu: Done

velioglu:
@marcocitus I have made the changes about casting and added tests

src/test/regress/expected/multi_insert_select_conflict.out
```diff
@@ -0,0 +1,400 @@
...
+     8 |     0
+     5 |     0
+     1 |     0
+     4 |     0
+     7 |     0
+     3 |     0
+     6 |     0
+     2 |     0
+     9 |     0
+(10 rows)
```
marcocitus: We probably need some ordering here

src/backend/distributed/executor/insert_select_executor.c
```diff
@@ -71,8 +75,52 @@ CoordinatorInsertSelectExecScan(CustomScanState *node)
...
 		}
 
-		ExecuteSelectIntoRelation(targetRelationId, insertTargetList, selectQuery,
-								  executorState);
+		shardConnectionsHash = ExecuteSelectIntoRelation(targetRelationId,
+														 insertTargetList, selectQuery,
+														 executorState,
+														 intermediateResultIdPrefix);
+
+		if (distributedPlan->workerJob != NULL)
```
marcocitus: Do you think it's worth writing a short README (<1 page) on how INSERT..SELECT via the coordinator works?

src/backend/distributed/planner/insert_select_planner.c
```diff
@@ -1295,3 +1325,145 @@ WrapSubquery(Query *subquery)
...
+		if (insertColumn->vartype != attr->atttypid)
+		{
+			CoerceViaIO *coerceExpr = makeNode(CoerceViaIO);
+			coerceExpr->arg = (Expr *) copyObject(insertColumn);
+			coerceExpr->resulttype = attr->atttypid;
+			coerceExpr->resultcollid = attr->attcollation;
+			coerceExpr->coerceformat = COERCE_IMPLICIT_CAST;
+			coerceExpr->location = -1;
+
+			targetEntry->expr = (Expr *) coerceExpr;
```
marcocitus: This is indeed what I had in mind.

src/backend/distributed/executor/insert_select_executor.c
velioglu: Yes, it makes sense. I will update the PR with a README.

marcocitus:
I think we should have a test that trigger text copy format , e.g. by inserting an array of composite types.

src/backend/distributed/executor/insert_select_executor.c
```diff
@@ -35,8 +36,9 @@
 #include "utils/snapmgr.h"
 
 
-static void ExecuteSelectIntoRelation(Oid targetRelationId, List *insertTargetList,
-									  Query *selectQuery, EState *executorState);
+static HTAB * ExecuteSelectIntoRelation(Oid targetRelationId, List *insertTargetList,
```
onderkalaci: We've briefly chatted with @marcocitus on this function, specifically its API.

Basically, the function becomes able to handle two very powerful/important things. Get a select query, execute it, either insert into a distributed table or split the results in a way that the results become co-located intermediate results with the target shards of a given relation.

The latter of the above could be a useful infrastructure to build more sophisticated query planning in the future.

So, the summary is that, @velioglu could you work/think on the API of this function a bit more and potentially make the distinction more explicit? 

Let's discuss the high-level changes you've in mind before pushing any changes for this.

src/test/regress/expected/multi_insert_select_conflict.out
```diff
@@ -0,0 +1,423 @@
...
+     4 |     0
+     5 |     0
+     6 |     0
+     7 |     0
+     8 |     0
+     9 |     0
+    10 |     0
+(10 rows)
+
+-- Get the select part from cte and do nothing on conflict
```
onderkalaci: I've slightly changed the query, and it didn't play nice. Do you know why is this happening? Or is that expected?

```SQL
WITH cte AS
  (SELECT col_1, col_2
   FROM source_table_1)
INSERT INTO target_table
SELECT source_table_1.col_1,
       source_table_1.col_2
FROM cte,
     source_table_1
WHERE cte.col_1 = source_table_1.col_1 ON CONFLICT DO NOTHING;
```

src/test/regress/expected/multi_insert_select_conflict.out
```diff
@@ -0,0 +1,423 @@
+CREATE SCHEMA on_conflict;
+SET search_path TO on_conflict, public;
```
onderkalaci: One test suggestion with foreign keys to reference tables:
```SQL
CREATE TABLE test_ref_table (key int PRIMARY KEY);
SELECT create_Reference_table('test_ref_table');
ALTER TABLE target_table ADD CONSTRAINT fkey FOREIGN KEY (col_1) REFERENCES test_ref_table(key) ON DELETE CASCADE;

BEGIN;
 TRUNCATE test_ref_table CASCADE;

 insert into target_table  SELECT col_2, col_1 from source_table_1 ON CONFLICT (col_1) DO UPDATE set col_2 = 55 RETURNING *;
ERROR:  cannot execute parallel COPY on relation "target_table" after DDL command on reference relation "test_ref_table" because there is a foreign key between them and "test_ref_table" has been modified in this transaction
DETAIL:  COPY to a distributed table uses a separate set of connections which will not be able to see the uncommitted changes to the reference table.
HINT:  Perform the COPY in a separate transaction.

```

And,

```SQL
postgres=# BEGIN;
BEGIN
 DELETE FROM test_ref_table ;
DEBUG:  switching to sequential query execution mode
DETAIL:  Reference relation "test_ref_table" is modified, which might lead to data inconsistencies or distributed deadlocks via parallel accesses to hash distributed relations due to foreign keys. Any parallel modification to those hash distributed relations in the same transaction can only be executed in sequential query execution mode
DELETE 3
insert into target_table  SELECT col_2, col_1 from source_table_1 ON CONFLICT (col_1) DO UPDATE set col_2 = 55 RETURNING *;                                                                      DEBUG:  cannot perform distributed INSERT INTO ... SELECT because the partition columns in the source table and subquery do not match
DETAIL:  The target table's partition column should correspond to a partition column in the subquery.
ERROR:  cannot execute parallel COPY on relation "target_table" after DML command on reference relation "test_ref_table" because there is a foreign key between them and "test_ref_table" has been modified in this transaction
DETAIL:  COPY to a distributed table uses a separate set of connections which will not be able to see the uncommitted changes to the reference table.
HINT:  Perform the COPY in a separate transaction.
```

onderkalaci: And, some passing tests where we only select from target table which has fkey to reference table, and insert into another table:
```SQL
BEGIN;
 TRUNCATE test_ref_table CASCADE;
 insert into source_table_1  SELECT col_2, col_1 from target_table ON CONFLICT (col_1) DO UPDATE set col_2 = 55 RETURNING *;
COMMIT;
```

AND

```SQL
BEGIN;
 DELETE FROM test_ref_table;
 insert into source_table_1  SELECT col_2, col_1 from target_table ON CONFLICT (col_1) DO UPDATE set col_2 = 55 RETURNING *;
COMMIT;
```

You should probably add primary key to source_table_1, or potentially create another table if that fails some of the tests.

src/backend/distributed/planner/INSERT_SELECT_PLANNER_README
```diff
@@ -0,0 +1,16 @@
+Citus supports INSERT ... SELECT queries either by pushing down the whole query to the
```
marcocitus: I think it would be nice if this was a markdown file (`.md`) so it gets nicely rendered in github. It might also be useful to make one file for the whole planner with different sections so you can easily scroll/search through it. So maybe just call it `planner_readme.md` and add a `## INSERT...SELECT via the coordinator` header.

velioglu:
@marcocitus @onderkalaci I have changed the API and added tests.

onderkalaci:
@velioglu This PR breaks appveyor builds. Is that expected? See master branch's appveyor output: https://ci.appveyor.com/project/citus/citus/builds/20720669

PR #2314, opened 2018-07-26
{WIP} Add support for foreign tables in MX

### LINKED ISSUE #2312: can not create cstore_fdw table in MX
Hit this problem when investigating another user issue

steps to reproduce
```sql
select start_metadata_sync_to_node('localhost', 9700);
create foreign table cs (a int, b int) server cstore_server;
select create_distributed_table('cs', 'a');
```

Result:
```
ERROR:  "cs" is a foreign table
DETAIL:  Foreign tables cannot have TRUNCATE triggers.
```

Failing command at worker is
```sql
 SELECT worker_create_truncate_trigger('public.cs')
```

When creating distributed table in coordinator we check [whether the table is a regular table](https://github.com/citusdata/citus/blob/master/src/backend/distributed/commands/create_distributed_table.c#L348)

We fail to perform the same check in [worker_create_truncate_trigger](https://github.com/citusdata/citus/blob/master/src/backend/distributed/worker/worker_truncate_trigger_protocol.c#L36) function.

Fix should be straightforward.

mtuncer:
DESCRIPTION: Add support for foreign tables in MX

- do not attempt to create truncate triggers for foreign tables anymore
- do not blindly try to create a server, use create if exists
- do not drop server when dropping foreign table

Fixes #2312

### DIFF
--- a/src/backend/distributed/metadata/metadata_sync.c
+++ b/src/backend/distributed/metadata/metadata_sync.c
@@ -312,7 +312,6 @@ MetadataCreateCommands(void)
 		List *shardIntervalList = NIL;
 		List *shardCreateCommandList = NIL;
 		char *metadataCommand = NULL;
-		char *truncateTriggerCreateCommand = NULL;
 		Oid clusteredTableId = cacheEntry->relationId;
 
 		/* add the table metadata command first*/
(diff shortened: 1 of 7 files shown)


onderkalaci:
Just FYI: @ozgune mentioned that foreign tables are not prioritized for Citus MX just yet. That's why we haven't worked on this PR more.

marcocitus:
Work is no longer in progress.

PR #2322, opened 2018-08-01
Update mitmproxy version to remove vulnerability warnings

lithp:
Github complains that mitmproxy is too old. This vulnerability doesn't affect us but updating mitmproxy is nicer anyway:

- We now require python 3.6, this should make more environments look like each other.
- The tests now run faster, because we don't wait for the tick() callback before applying changes.


jasonmp85:
Travis says not to depend upon the location of `$HOME`, so I've replaced the `/home/travis` stuff with `$PROJECT_BUILD_DIR`, which should suffice.

PR #2323, opened 2018-08-01
Alternative approach for hiding shards on the MX workers for better UX

onderkalaci:
We're asked to hide shard names behind a schema on MX workers for a much better UX. Although the requirement makes a lot of sense, implementing that approach is quite complicated and would make it harder to maintain. Our current logic for schema support is pretty simple and straight-forward so not touching it would be desirable.

That's why I tried to come up with some ideas. All my ideas gathered around modifying the behavior of `pg_catalog.pg_table_is_visible()`, which is used in Postgres source (and mainly psql) when listing tables.

There are Postgres hooks to modify the behavior of plpgsql functions, however, `pg_catalog` functions are not allowed. That's why I ended-up using a similar approach as we do in `citus.replace_isolation_tester_func()` and `citus.restore_isolation_tester_func()`.

In this case, I've overridden `pg_catalog.pg_table_is_visible()` such that it filters out the tables that are known to be shards in MX workers.

I can see that this approach might be controversial, however, I find breaking the parity of table and its shards controversial as well. And, I find this approach much less controversial.

The results are promising, this PR works expected such that `\d(+)` and other `psql` commands that are supposed to show the tables skips the shards on MX worker nodes. However, you can obviously query those shards if you know the table names.

There are a few things to implement, but before that, I wanted to hear some thoughts on this approach.

Todo list:
- [ ] Add a GUC to control this behavior. Users should be able to disable this behavior. But hiding the names should be the default behaviour
- [ ] Make sure that we don't break anything during upgrades
- [ ] Make sure that we don't break anything on `DROP EXTENSION`. Also, get rid of the version hacks that are in this WIP PR
- [ ] Add some tests 😄 
- [ ] Add some `DEBUG2` or `DEBUG1` messages on the skipped shards
- [ ] Make sure that Postgres' `check-world` runs fine even if Citus extension is loaded (and created as well)


marcocitus:
> There are Postgres hooks to modify the behavior of plpgsql functions, however, pg_catalog functions are not allowed. That's why I ended-up using a similar approach as we do in citus.replace_isolation_tester_func() and citus.restore_isolation_tester_func().

It scares me a little in terms of future pg_upgrade and such. What if walk through the query tree and replace the function OID prior to calling the planner?

src/backend/distributed/citus--8.0-1--8.0-2.sql
```diff
@@ -0,0 +1,62 @@
...
+COMMENT ON FUNCTION citus_table_is_visible(oid)
+	IS 'wrapper on pg_table_is_visible, filtering out tables (and indexes) that are known to be shards';
+
+-- this is the exact same query with what \d
+-- command produces, except pg_table_is_visible
+-- is replaced with pg_catalog.relation_is_a_known_shard(c.oid)
+CREATE VIEW citus.citus_shards_on_worker AS 
+	SELECT n.nspname as "Schema",
+	  c.relname as "Name",
+	  CASE c.relkind WHEN 'r' THEN 'table' WHEN 'v' THEN 'view' WHEN 'm' THEN 'materialized view' WHEN 'i' THEN 'index' WHEN 'S' THEN 'sequence' WHEN 's' THEN 'special' WHEN 'f' THEN 'foreign table' WHEN 'p' THEN 'table' END as "Type",
```
mtuncer: Is there a way to encapsulate this in a function ? Same thing is also used in the next view definition

src/backend/distributed/shared_library_init.c
```diff
@@ -379,6 +380,19 @@ RegisterCitusConfigVariables(void)
 		GUC_NO_SHOW_ALL,
 		NULL, NULL, NULL);
 
+	DefineCustomBoolVariable(
+		"citus.enable_replace_table_visible_function",
```
mtuncer: did not like the name here. `enable` seem to be redundant. `override_table_visibility` or `use _citus_table_visibility` sounds better to me

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -2051,6 +2053,38 @@ CitusTextSendAsJsonbFunctionId(void)
 }
 
 
+/* return oid of the pg_table_is_visible function */
```
mtuncer: we usually start function comment with the function name

src/backend/distributed/worker/worker_data_fetch_protocol.c
```diff
@@ -526,30 +525,46 @@ worker_apply_sequence_command(PG_FUNCTION_ARGS)
...
-/* Extracts shard id from the given table name, and returns it. */
-static uint64
-ExtractShardId(const char *tableName)
+/*
+ * ExtractShardId tries to extract shard id from the given table name,
+ * and returns the shard id if table name is formatted as shard name.
+ * Else, the function returns INVALID_SHARD_ID.
+ */
+uint64
+ExtractShardId(const char *tableName, bool missingOk)
```
mtuncer: `ExtractShardId` is too general consider `ExtractShardIdFromTableName`

src/backend/distributed/worker/worker_shard_visibility.c
```diff
@@ -0,0 +1,299 @@
...
+
+/*
+ * strndup is not part of Standard C and extension of Posix.1-2008.
+ * So, implement our own strndup to make it work on Windows.
+ */
+static char *
+citus_strndup(const char *s, size_t n)
+{
+	char *result;
+	size_t len = strlen(s);
```
mtuncer: is `strnlen` better here ?

src/backend/distributed/worker/worker_shard_visibility.c
```diff
@@ -0,0 +1,299 @@
...
+
+	if (n < len)
+	{
+		len = n;
+	}
+
+	result = (char *) malloc(len + 1);
```
mtuncer: do we need to care about deallocation and memory context ?

src/include/distributed/worker_shard_visibility.h
```diff
@@ -0,0 +1,22 @@
...
+ *-------------------------------------------------------------------------
+ */
+
+#ifndef WORKER_SHARD_VISIBILITY_H
+#define WORKER_SHARD_VISIBILITY_H
+
+#include "nodes/nodes.h"
+
+extern bool EnableReplaceTableVisibleFunction;
+
```
mtuncer: extra line breaks

src/backend/distributed/worker/worker_shard_visibility.c
onderkalaci: Oh, that's a good catch. We should simply use `palloc`, right?

marcocitus: Why not `pstrdup`?

onderkalaci: Interesting, I searched for `pstrndup` in Postgres source, but the function is named  `pnstrdup`. So, I'll switch to that.

(note that I don't want to use `pstrdup ` because the goal here is to get the main table name from the shard name, where we know the exact length we want to copy: `relationName = citus_strndup(shardRelationName, relationNameLength);`)

marcocitus: What if the relation name is truncated because it is very long?

marcocitus: e.g. https://github.com/citusdata/citus/blob/master/src/test/regress/expected/multi_name_lengths.out#L29

onderkalaci: Ops, it is currently broken with this PR and that's something slightly hard to solve. We should do something similar [here](https://github.com/citusdata/citus/commit/f677fadbe686ed370844549e06d9cd3c8c46bbb0#diff-41af17ffe3adc232e63ff6f138e3ab74R648). But, here we should iterate on all distributed relations and find the distributed table with the same hash when the `shardRelationName` long.

marcocitus: Could we iterate over the shards, construct the name, and include them in the result if there is a table with that name?

onderkalaci: I decided to not rely on table/shard names at all. The new algorithm is simpler with some overhead: Extract shard id -> Read `pg_dist_shard` (via index scan on shardId)** -> Make sure that the shard belongs to a distributed table.

This covers the long shard names and allows us to get rid of `citus_strndup ` function.

** note that the overhead is neglicable, I tested locally with 4000 shards on a worker.

src/backend/distributed/citus--8.0-1--8.0-2.sql
onderkalaci: What would be the benefit of that?

src/backend/distributed/worker/worker_shard_visibility.c
```diff
@@ -0,0 +1,240 @@
...
+		/*
+		 * If the input relation is an index we simply replace the
+		 * relationId with the corresponding relation to hide indexes
+		 * as well. See RelationIsAKnownShard() for the details and give
+		 * more meaningful debug message here.
+		 */
+		relKind = get_rel_relkind(relationId);
+		if (relKind == RELKIND_INDEX)
+		{
+			ereport(DEBUG1, (errmsg("skipping index \"%s\" since it belongs to a shard",
```
onderkalaci: Shall we make it `DEBUG2`, what do you think @mtuncer ?

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -3263,10 +3300,11 @@ LookupDistShardTuples(Oid relationId)
 /*
  * LookupShardRelation returns the logical relation oid a shard belongs to.
  *
- * Errors out if the shardId does not exist.
+ * Errors out if the shardId does not exist and missing_ok is true. Returns
```
mtuncer: Comment is not right. Code is doing the opposite.

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -2051,6 +2052,42 @@ CitusTextSendAsJsonbFunctionId(void)
 }
 
 
+/*
+ * PgTableVisibleFuncId returns oid of the pg_table_is_visible function.
```
mtuncer: we usually put short function comments should be in single line like
```c
/* PgTableVisibleFuncId returns oid of the pg_table_is_visible function. */
```
However, I saw many exceptions to this in the code. I guess we are not very strict on this.

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -2051,6 +2052,42 @@ CitusTextSendAsJsonbFunctionId(void)
...
+ */
+Oid
+PgTableVisibleFuncId(void)
+{
+	if (MetadataCache.pgTableIsVisibleFuncId == InvalidOid)
+	{
+		const int argCount = 1;
+
+		MetadataCache.pgTableIsVisibleFuncId =
```
mtuncer: Could this call fail ?

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -3263,10 +3300,11 @@ LookupDistShardTuples(Oid relationId)
...
  * LookupShardRelation returns the logical relation oid a shard belongs to.
  *
- * Errors out if the shardId does not exist.
+ * Errors out if the shardId does not exist and missing_ok is true. Returns
+ * InvalidOid if the shardId does not exist and missing_ok is false.
  */
-static Oid
-LookupShardRelation(int64 shardId)
+Oid
+LookupShardRelation(int64 shardId, bool missing_ok)
```
mtuncer: `missing_ok` or `missingOk` ? 
`ExtractShardIdFromTableName` used `missingOk`

src/backend/distributed/utils/metadata_cache.c
onderkalaci: Yes, it does if someone changes the signature of `pg_table_is_visible`. However, that seems like an unlikely case unless a major Postgres version change. 

Also, we're calling `FunctionOid ()` for several other functions as well.

I already added a regression test (e.g., at the beginning of the sql file), which ensures that the signatures are OK.

I cannot think of doing anything if the function returns `InvalidOid` instead of erroring out. We'll need to error out by ourselves?

src/backend/distributed/worker/worker_shard_visibility.c
mtuncer: DEBUG2 might be better.

src/backend/distributed/worker/worker_data_fetch_protocol.c
```diff
@@ -526,30 +525,46 @@ worker_apply_sequence_command(PG_FUNCTION_ARGS)
...
 	shardIdString++;
 
 	errno = 0;
 	shardId = pg_strtouint64(shardIdString, &shardIdStringEnd, 0);
 
 	if (errno != 0 || (*shardIdStringEnd != '\0'))
 	{
-		ereport(ERROR, (errmsg("could not extract shardId from table name \"%s\"",
-							   tableName)));
+		if (!missingOk)
```
mtuncer: Code repetition makes me feel uncomfortable here. Is there an easy way to avoid this ?

src/backend/distributed/citus--8.0-1--8.0-2.sql
onderkalaci: Note that the queries have some minor differences if you're asking to use that function in both views.

src/backend/distributed/worker/worker_shard_visibility.c
```diff
@@ -0,0 +1,240 @@
+/*
+ * worker_shard_visibility.c
+ *
+ * Implements the functions for hiding shards on the Citus MX
+ * worker (data) nodes. This is mostly required for
+ *
+ *  * Copyright (c) 2018, Citus Data, Inc.
```
mtuncer: extra `  * `

I think our copyright message was 
` Copyright (c) 2012-2018, Citus Data, Inc.`

src/backend/distributed/worker/worker_shard_visibility.c
```diff
@@ -0,0 +1,240 @@
...
+		else
+		{
+			ereport(DEBUG1, (errmsg("skipping relation \"%s\" since it is a shard",
+									get_rel_name(relationId))));
+		}
+
+		return BoolGetDatum(false);
+	}
+
+	return BoolGetDatum(RelationIsVisible(relationId));
```
mtuncer: `PG_RETURN_BOOL` ?

src/backend/distributed/worker/worker_data_fetch_protocol.c
onderkalaci: Hmm, there are two steps involved in extracting the `shardId`.

1. Search for `_` in the table name
2. Try to convert `str` to `uint`

Two `missingOk`  seems should be added for each of the checks above. If you see a simplification, feel free to suggest, I can apply them

src/backend/distributed/utils/metadata_cache.c
onderkalaci: fixed

src/backend/distributed/worker/worker_shard_visibility.c
onderkalaci: Sure, we should fix it and the above.

PR #2324, opened 2018-08-01
make tests pass on pg 11 again

thanodnl:
This diff makes the test suite of pg 11 pass again. It addresses the following items:

1. Changed output for `\d` on indexes due to https://github.com/postgres/postgres/commit/f4a5ce359587257c6d14b8e746889077c649439a. This is fixed by using a view (`index_attrs`) that lists the columns and types of indexes.
2. Add extra output file for `foreign_key_restriction_enforcement` to account for changed debug message in index creation https://github.com/postgres/postgres/commit/9da0cc3 . That commit added the word `serially` to the debug line when creating an index non-parallel. Disabling the debug log output could not happen as it was deliberately turned on to test when the feature kicks in.
3. Fixed compilation of `ddl/foreign_constraint.c` due to changed headers.

This diff depends on a change in citusdata/tools as the master branch for postgres already reflects progress on pg12. The branch changes the installation of pg11 via the `REL_11_STABLE` branch. Separate PR needs to be opened for the change on citusdata/tools which needs to land before these changes can land.


jasonmp85:
> Chatted with @mtuncer and he is owning this PR. Is there any other PR you are reviewing for PG 11?

The last time @mtuncer and I worked on PG11 stuff I had some significant changes I wanted to use to refactor this stuff into a cleaner state. I've got local work from that conversation that we need to use to accomplish this, so wanted to own this.

jasonmp85:
@metdos I'm going to own this and will merge it today.

mtuncer:
@jasonmp85 see https://travis-ci.org/citusdata/citus/jobs/413955974
There is a problem with PG 11 packaged version.  One test hangs/deadlocks. It would be great if you could reproduce that locally and fix.

sumedhpathak:
Just chatted with @jasonmp85 and @mtuncer .  @jasonmp85 will own getting this PR done. He'll chat with @thanodnl on the hanging test and look to resolve it.

jasonmp85:
@thanodnl ok, so I figured out this is due to `jit`. I got a clean compile going without the weird `select` behavior I was seeing (think this was due to me recompiling and maybe not cleaning up all the files between builds?)…

I got tests to pass with a source build similar to Travis'. After that I changed _one thing_, which was to point to `llvm-config` and build PostgreSQL with JIT support. I got a hang exactly where Travis does.

I changed `pg_regress_multi.pl` to have a `jit=off` line and everything got fixed.

thanodnl:
Thanks for looking into this.

It is strange and scary we see a hang due to jit ... is this something we _should_ communicate to pgsql-hackers, Or is this very specific to citus?

mtuncer:
@anarazel may give us better insight here

PR #2325, opened 2018-08-01
Attempt to stabilize packet dumps and add them back it

lithp:


jasonmp85:
@onderkalaci I'll be taking this (I thought it was always assigned to me, but apparently not?)

PR #2330, opened 2018-08-07
Add simple poolinfo test

jasonmp85:
This just exercises the check constraint to ensure it prohibits all
INSERTs in community.


jasonmp85:
@marcocitus I separated this test out from citusdata/citus-enterprise#270. Similar to the authinfo test above it, I'm going to modify the output in Enterprise to verify that the modified check constraint in that codebase works appropriately.

PR #2332, opened 2018-08-07
Support TABLESAMPLE in router queries

### LINKED ISSUE #2331: TABLESAMPLE not deparsed
We allow TABLESAMPLE in router queries, but we do not add it to the queries sent to the workers. Therefore all rows are returned, incorrectly.

The reason is that we don't check for shards in the tablesample deparsing logic:
https://github.com/citusdata/citus/blob/master/src/backend/distributed/utils/ruleutils_10.c#L7138

A shard is represented as an RTE_FUNCTION with the shard ID encoded into the parameters.

marcocitus:
DESCRIPTION: Support for TABLESAMPLE in router queries

We currently allow TABLESAMPLE in router queries, but don't actually add it to the query during deparsing leading to statistically incorrect results.

Fixes #2331

### DIFF
--- a/src/test/regress/sql/multi_simple_queries.sql
+++ b/src/test/regress/sql/multi_simple_queries.sql
@@ -296,4 +296,10 @@ SELECT count(*) FROM (
                   xmax IS NOT NULL
 ) x;
 
+-- tablesample is supported
+SELECT * FROM articles TABLESAMPLE SYSTEM (0) WHERE author_id = 1;
+SELECT * FROM articles TABLESAMPLE BERNOULLI (0) WHERE author_id = 1;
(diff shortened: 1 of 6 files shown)


mtuncer:
did not build for PG11, needed to rebase to current master first

src/test/regress/sql/multi_simple_queries.sql
```diff
@@ -296,4 +296,8 @@ SELECT count(*) FROM (
                   xmax IS NOT NULL
 ) x;
 
+-- tablesample is supported
+SELECT * FROM articles TABLESAMPLE SYSTEM (0) WHERE author_id = 1;
+SELECT * FROM articles TABLESAMPLE BERNOULLI (0) WHERE author_id = 1;
+
```
mtuncer: perhaps add a test with  percentage = 100

PR #2333, opened 2018-08-08
Adds support for writing to reference tables from MX nodes.

### LINKED ISSUE #2252: Citus MX: Better support for referenced tables
Hi,

Currently, any modifications to a reference table has to originate on coordinator. This creates an additional complexity for applications as they have to carry two connections (one for MX and one for the coordinator).

Additionally triggers can't be used to modify referenced tables as direct (table_12321) modifications are not propagated and indirect (table) are not supported.

Any performance optimizations would be also greatly appreciated.

velioglu:
DESCRIPTION: Adds support for applying DML operations on reference tables from MX nodes

Fixes #2252 

Adds support for writing to reference tables from MX nodes. Two essential changes have been applied in order to prevent divergence between placement of reference tables on different MX nodes.

- Get shard resource lock on the first worker node, if the INSERT/UPDATE/DELETE or SELECT FOR UPDATE query affects any reference table. Get that lock even if the query is sent from the coordinator node. We get that lock on the first worker node instead of the coordinator node, since worker nodes don't know the location of coordinator node. Beside, getting that lock on the first one is enough.

- Get shard metadata lock of reference tables on worker nodes while changing the metadata for these tables (by master_add_node, master_remove_node, master_activate_node, master_disable_node)

In order to test getting locks from MX nodes on isolation tests, very simple UDF has also been introduced. 

Remaining tests will be either added to the PR or made by pg_bench.

- [x] Add writing from mx nodes on regression tests
- [x] Test concurrent operations with pg_bench

### DIFF
--- a/src/backend/distributed/commands/multi_copy.c
+++ b/src/backend/distributed/commands/multi_copy.c
@@ -2101,12 +2101,6 @@ CitusCopyDestReceiverStartup(DestReceiver *dest, int operation,
 	copyDest->distributedRelation = distributedRelation;
 	copyDest->tupleDescriptor = inputTupleDescriptor;
 
-	/* we don't support copy to reference tables from workers */
-	if (partitionMethod == DISTRIBUTE_BY_NONE)
-	{
-		EnsureCoordinator();
-	}
-
 	/* load the list of shards and verify that we have shards to copy into */
 	shardIntervalList = LoadShardIntervalList(tableId);
 	if (shardIntervalList == NIL)
@@ -2147,7 +2141,7 @@ CitusCopyDestReceiverStartup(DestReceiver *dest, int operation,
 	 * Prevent concurrent UPDATE/DELETE on replication factor >1
 	 * (see AcquireExecutorMultiShardLocks() at multi_router_executor.c)
 	 */
-	LockShardListResources(shardIntervalList, RowExclusiveLock);
+	SerializeNonCommutativeWrites(shardIntervalList, RowExclusiveLock);
 
 	/* keep the table metadata to avoid looking it up for every tuple */
 	copyDest->tableMetadata = cacheEntry;
--- a/src/backend/distributed/executor/multi_router_executor.c
+++ b/src/backend/distributed/executor/multi_router_executor.c
@@ -32,7 +32,9 @@
 #include "distributed/deparse_shard_query.h"
 #include "distributed/listutils.h"
 #include "distributed/master_metadata_utility.h"
+#include "distributed/master_protocol.h"
 #include "distributed/metadata_cache.h"
+#include "distributed/metadata_sync.h"
 #include "distributed/multi_executor.h"
 #include "distributed/multi_partitioning_utils.h"
 #include "distributed/multi_physical_planner.h"
@@ -248,7 +250,9 @@ AcquireExecutorShardLock(Task *task, CmdType commandType)
 
 	if (shardId != INVALID_SHARD_ID && lockMode != NoLock)
 	{
-		LockShardResource(shardId, lockMode);
+		ShardInterval *shardInterval = LoadShardInterval(shardId);
+
+		SerializeNonCommutativeWrites(list_make1(shardInterval), lockMode);
 	}
 
 	/*
@@ -282,8 +286,6 @@ AcquireExecutorShardLock(Task *task, CmdType commandType)
 			if (PartitionMethod(relationId) == DISTRIBUTE_BY_NONE)
 			{
 				List *shardIntervalList = LoadShardIntervalList(relationId);
-				ShardInterval *referenceTableShardInterval = (ShardInterval *) linitial(
-					shardIntervalList);
 
 				if (rowLockStrength == LCS_FORKEYSHARE || rowLockStrength == LCS_FORSHARE)
 				{
@@ -295,7 +297,7 @@ AcquireExecutorShardLock(Task *task, CmdType commandType)
 					rowLockMode = ExclusiveLock;
 				}
 
-				LockShardResource(referenceTableShardInterval->shardId, rowLockMode);
+				SerializeNonCommutativeWrites(shardIntervalList, rowLockMode);
 			}
 		}
 	}
--- a/src/backend/distributed/transaction/worker_transaction.c
+++ b/src/backend/distributed/transaction/worker_transaction.c
@@ -20,6 +20,7 @@
 
 #include "access/xact.h"
 #include "distributed/connection_management.h"
+#include "distributed/listutils.h"
 #include "distributed/metadata_cache.h"
 #include "distributed/multi_shard_transaction.h"
 #include "distributed/resource_lock.h"
@@ -52,6 +53,30 @@ SendCommandToWorker(char *nodeName, int32 nodePort, char *command)
 	MarkRemoteTransactionCritical(transactionConnection);
 	RemoteTransactionBeginIfNecessary(transactionConnection);
 	ExecuteCriticalRemoteCommand(transactionConnection, command);
+}
+
+
+/*
+ * SendCommandToFirstWorker sends the given command only to the first worker node
+ * sorted by host name and port number using SendCommandToWorker.
+ */
+void
+SendCommandToFirstWorker(char *command)
+{
+	List *workerNodeList = ActivePrimaryNodeList();
+	WorkerNode *firstWorkerNode = NULL;
+
+	workerNodeList = SortList(workerNodeList, CompareWorkerNodes);
+
+	if (list_length(workerNodeList) == 0)
+	{
+		ereport(ERROR, (errmsg("cannot find a worker node")));
+	}
+
+	firstWorkerNode = (WorkerNode *) linitial(workerNodeList);
+
+	SendCommandToWorker(firstWorkerNode->workerName, firstWorkerNode->workerPort,
+						command);
 }
 
 
--- a/src/backend/distributed/utils/reference_table_utils.c
+++ b/src/backend/distributed/utils/reference_table_utils.c
@@ -128,7 +128,9 @@ void
 ReplicateAllReferenceTablesToNode(char *nodeName, int nodePort)
 {
 	List *referenceTableList = ReferenceTableOidList();
+	List *referenceShardIntervalList = NIL;
 	ListCell *referenceTableCell = NULL;
+	ListCell *referenceShardIntervalCell = NULL;
 	List *workerNodeList = ActivePrimaryNodeList();
 	uint32 workerCount = 0;
 	Oid firstReferenceTableId = InvalidOid;
@@ -150,6 +152,20 @@ ReplicateAllReferenceTablesToNode(char *nodeName, int nodePort)
 		Oid referenceTableId = lfirst_oid(referenceTableCell);
 		List *shardIntervalList = LoadShardIntervalList(referenceTableId);
 		ShardInterval *shardInterval = (ShardInterval *) linitial(shardIntervalList);
+
+		referenceShardIntervalList = lappend(referenceShardIntervalList,
+											 shardInterval);
+	}
+
+	if (ClusterHasKnownMetadataWorkers())
+	{
+		BlockWritesToShardList(referenceShardIntervalList);
+	}
+
+	foreach(referenceShardIntervalCell, referenceShardIntervalList)
+	{
+		ShardInterval *shardInterval = (ShardInterval *) lfirst(
+			referenceShardIntervalCell);
 		uint64 shardId = shardInterval->shardId;
 
 		LockShardDistributionMetadata(shardId, ExclusiveLock);
@@ -394,6 +410,7 @@ void
 DeleteAllReferenceTablePlacementsFromNodeGroup(uint32 groupId)
 {
 	List *referenceTableList = ReferenceTableOidList();
+	List *referenceShardIntervalList = NIL;
 	ListCell *referenceTableCell = NULL;
 
 	/* if there are no reference tables, we do not need to do anything */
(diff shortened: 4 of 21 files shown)


src/backend/distributed/utils/resource_lock.c
```diff
@@ -122,6 +123,93 @@ lock_shard_resources(PG_FUNCTION_ARGS)
...
+	if (targetWorkerSet == FIRST_WORKER_WITH_METADATA)
+	{
+		WorkerNode *firstWorkerNode = NULL;
+		workerNodeList = SortList(workerNodeList, CompareWorkerNodesByGroupId);
+
+		firstWorkerNode = (WorkerNode *) linitial(workerNodeList);
+		if (firstWorkerNode->groupId == GetLocalGroupId())
+		{
+			ereport(WARNING, (errmsg(
+								  "Since the query is sent from first worker, won't open connection to itself.")));
```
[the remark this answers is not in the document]
velioglu: Otherwise we can face with self deadlock. Both the transaction will try to get the lock and the distributed transaction opened by that node on itself will try to get that lock.

src/backend/distributed/transaction/worker_transaction.c
```diff
@@ -55,6 +56,59 @@ SendCommandToWorker(char *nodeName, int32 nodePort, char *command)
...
+	WorkerNode *firstWorkerNode = NULL;
+
+	workerNodeList = SortList(workerNodeList, CompareWorkerNodesByGroupId);
+	firstWorkerNode = (WorkerNode *) linitial(workerNodeList);
+
+	if (firstWorkerNode->groupId == GetLocalGroupId())
+	{
+		/*
+		 * Don't get shard resouce lock on the first worker, otherwise it
+		 * results with self-deadlock.
```
marcocitus: This check seems out of place in `SendCommandToFirstWorker`. It's not what I would expect from the name of the function. Maybe we should do all of this in `LockShardListResourcesOnFirstWorker`.

src/include/distributed/worker_transaction.h
```diff
@@ -21,6 +21,7 @@
  */
 typedef enum TargetWorkerSet
 {
+	FIRST_WORKER,
```
marcocitus: seems unused

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -246,6 +248,20 @@ AcquireExecutorShardLock(Task *task, CmdType commandType)
...
+
+		/*
+		 * If the modified shard is a reference table's shard we need to get
+		 * shard resource lock on the first worker node to prevent divergence
+		 * possibility between placements of the reference table.
+		 */
+		if (ReferenceTableShardId(shardId) && ShouldSyncTableMetadata(
+				shardInterval->relationId))
+		{
+			ShardInterval *shardInterval = LoadShardInterval(shardId);
```
onderkalaci: We already have `shardInterval ` loaded, so lets not do it again.

velioglu:
After discussing with @onderkalaci we decided to handle this case by adding a function like `QueryRunOnMXCluster` which controls whether the query is running on the MX cluster by checking two things

1 - If the query is run via coordinator, we need to check whether we have a worker node with metadata.
2 - Is query running via worker

If one of them is correct, then we can say that the query is running on MX cluster.

src/backend/distributed/utils/resource_lock.c
```diff
@@ -122,6 +124,90 @@ lock_shard_resources(PG_FUNCTION_ARGS)
...
+LockShardListResourcesOnFirstWorker(LOCKMODE lockmode, List *shardIntervalList)
+{
+	List *workerNodeList = ActivePrimaryNodeList();
+	WorkerNode *firstWorkerNode = NULL;
+	StringInfo lockCommand = makeStringInfo();
+	ListCell *shardIntervalCell = NULL;
+	int processedShardPlacementCount = 0;
+	int totalShardPlacementCount = list_length(shardIntervalList);
+
+	workerNodeList = SortList(workerNodeList, CompareWorkerNodesByGroupId);
```
onderkalaci: Can we move this logic into a separate function like `IsFirstWorker()` and do

```C
if (IsFirstWorker())
   return;
```

Actually, I think we should probably do

```C
if firstWorker
   LockShardResourceList(..)
  return SHARD_RESOURCES_LOCKED_LOCALLY

...
return SHARD_RESOURCES_LOCKED_REMOTELY
```

Because otherwise the name of the function won't match what its doing. We can later use the return value to decide on calling `LockShardResource` such as : 

```C


	if (list_length(referenceShardIntervalList) != 0)
	{
		if (LockShardListResourcesOnFirstWorker(lockMode, referenceShardIntervalList) != SHARD_RESOURCES_LOCKED_LOCALLY)
		{
			LockShardResourceList();
		}
	}
```

We could even consider not retuning any value and call `LockShardResourceList();` with the cost re-locking the same shard intervals. However, I think I'd prefer returning a value to signal the caller whether we've locked locally or remotely.

src/backend/distributed/utils/resource_lock.c
```diff
@@ -122,6 +124,90 @@ lock_shard_resources(PG_FUNCTION_ARGS)
 }
 
 
+/*
+ * LockShardListResourcesOnFirstWorker acquires the resource locks for the specified
```
onderkalaci: I think it'd be useful to explain the reasoning behind acquiring locks on the first worker here.

src/backend/distributed/utils/resource_lock.c
```diff
@@ -360,11 +464,33 @@ LockShardsInPlacementListMetadata(List *shardPlacementList, LOCKMODE lockMode)
...
+	foreach(shardIntervalCell, shardIntervalList)
+	{
+		ShardInterval *shardInterval = (ShardInterval *) lfirst(shardIntervalCell);
+		int64 shardId = shardInterval->shardId;
+
+		/*
+		 * If it is a reference table shard, lock should be get on worker nodes
+		 * to prevent divergence of shard placement on different worker nodes.
+		 */
+		if (ReferenceTableShardId(shardId) && QueryRunOnMXCluster())
```
onderkalaci: I really don't want to add code complexity to locking functions. They are currently very simple, and lets try to keep them as simple.

I can see that you're first trying to get all the reference table shards that needs to locked on the remote node. That makes sense, but, can we move the logic to a seperate function like:
```C
LockShardResroucesOnFirstWorkerIfNecessary()
     foreach shardId:
     if (ReferenceTableShardId(shardId) && QueryRunOnMXCluster())
          LockShardListResourcesOnFirstWorker(lockMode, list_make1(shardInterval));
```

Later, change here and `AcquireExecutorLocks` to something like:
```C
   if (LockShardResroucesOnFirstWorkerIfNecessary() != SHARD_RESOURCES_LOCKED_LOCALLY)
        LockShardResource()
```

src/test/regress/sql/multi_mx_modifications_to_reference_tables.sql
```diff
@@ -0,0 +1,91 @@
+
+
+ALTER SEQUENCE pg_catalog.pg_dist_shardid_seq RESTART 1470000;
```
marcocitus: Minor: We generally shouldn't rely on the shard ID sequence. If the output contains shard IDs we should use `SET citus.next_shard_id TO 1470000`.

src/backend/distributed/utils/reference_table_utils.c
```diff
@@ -407,6 +424,35 @@ DeleteAllReferenceTablePlacementsFromNodeGroup(uint32 groupId)
...
+				continue;
+			}
+
+			placement = (GroupShardPlacement *) linitial(placements);
+			shardInterval = LoadShardInterval(placement->shardId);
+			referenceShardIntervalList = lappend(referenceShardIntervalList,
+												 shardInterval);
+		}
+
+		BlockWritesToShardList(referenceShardIntervalList);
```
onderkalaci: I think we can also remove `LockShardDistributionMetadata ()` in the below loop since we already have that in `BlockWritesToShardList`.

src/backend/distributed/utils/reference_table_utils.c
```diff
@@ -150,6 +152,20 @@ ReplicateAllReferenceTablesToNode(char *nodeName, int nodePort)
...
+	}
+
+	foreach(referenceShardIntervalCell, referenceShardIntervalList)
+	{
+		ShardInterval *shardInterval = (ShardInterval *) lfirst(
+			referenceShardIntervalCell);
 		uint64 shardId = shardInterval->shardId;
 
 		LockShardDistributionMetadata(shardId, ExclusiveLock);
```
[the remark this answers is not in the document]
velioglu: After talking with @onderkalaci we decided to keep the implementation like this, otherwise non-MX cluster won't acquire the required lock. Beside, we want to keep MX cluster check logic on this function instead of `BlockWritesToShardList`

src/backend/distributed/utils/reference_table_utils.c
velioglu: After talking with @onderkalaci we decided to keep the implementation like this, otherwise non-MX cluster won't acquire the required lock. Beside, we want to keep MX cluster check logic on this function instead of `BlockWritesToShardList`

src/backend/distributed/utils/resource_lock.c
```diff
@@ -122,6 +128,124 @@ lock_shard_resources(PG_FUNCTION_ARGS)
...
+	{
+		return LockShardListResourcesOnFirstWorker(lockmode, shardIntervalList);
+	}
+
+	return SHARD_RESOURCES_NO_LOCK;
+}
+
+
+/*
+ * LockShardListResourcesOnFirstWorker acquires the resource locks for the specified
```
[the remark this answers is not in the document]
velioglu: Done

src/backend/distributed/utils/resource_lock.c
```diff
@@ -122,6 +128,124 @@ lock_shard_resources(PG_FUNCTION_ARGS)
...
+		processedShardPlacementCount++;
+		if (processedShardPlacementCount != totalShardPlacementCount)
+		{
+			appendStringInfo(lockCommand, ", ");
+		}
+	}
+
+	appendStringInfo(lockCommand, "])");
+
+	SendCommandToFirstWorker(lockCommand->data);
```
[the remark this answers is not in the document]
velioglu: Done

src/backend/distributed/utils/resource_lock.c
```diff
@@ -178,14 +275,19 @@ LockReferencedReferenceShardDistributionMetadata(uint64 shardId, LOCKMODE lock)
 
 	DistTableCacheEntry *cacheEntry = DistributedTableCacheEntry(relationId);
 	List *referencedRelationList = cacheEntry->referencedRelationsViaForeignKey;
-
 	List *shardIntervalList = GetSortedReferenceShardIntervals(referencedRelationList);
+
```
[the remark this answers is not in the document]
velioglu: Done

src/backend/distributed/commands/multi_copy.c
```diff
@@ -2145,8 +2139,15 @@ CitusCopyDestReceiverStartup(DestReceiver *dest, int operation,
...
 	 * (see AcquireExecutorMultiShardLocks() at multi_router_executor.c)
+	 *
+	 * Note that, we may need to get shard resource lock on the worker to prevent
+	 * divergence between the placement of reference table shards.
 	 */
-	LockShardListResources(shardIntervalList, RowExclusiveLock);
+	if (LockShardResourcesOnFirstWorkerIfNecessary(RowExclusiveLock, shardIntervalList) !=
+		SHARD_RESOURCES_LOCKED_LOCALLY)
+	{
+		LockShardListResources(shardIntervalList, RowExclusiveLock);
```
marcocitus: This code is super confusing to me, could we maybe move the reference-table-on-mx-specific logic into `LockShardListResources` or a wrapper that does both? It could maybe also be named for its purpose, e.g. `SerialiseNonCommutativeWrites`.

src/backend/distributed/utils/resource_lock.c
```diff
@@ -353,11 +475,43 @@ LockShardsInPlacementListMetadata(List *shardPlacementList, LOCKMODE lockMode)
...
+void
+SerializeNonCommutativeWrites(List *shardIntervalList, LOCKMODE lockMode)
+{
+	ShardInterval *firstShardInterval = (ShardInterval *) linitial(shardIntervalList);
+	int64 firstShardId = firstShardInterval->shardId;
+
+	if (ReferenceTableShardId(firstShardId) && ClusterHasKnownMetadataWorkers() &&
+		!IsFirstWorkerNode())
+	{
+		LockShardListResourcesOnFirstWorker(lockMode, shardIntervalList);
```
marcocitus: nice!

velioglu:
I have made some pgbench tests on cloud (with parameters -c 128 and -j 64)

Followings are the TPS results I have obtained with 4-node cluster. MASTER column means that cluster is formed using the master branch and DML operations are applied via the coordinator node. DML_ON_REF column means that cluster is formed using this branch and DML operations are applied using the data node URL given on the cloud interface.

3 Worker	          MASTER	DML_ON_REF
INSERT	                5753	5845
UPDATE	                   171	192
DELETE	                   181	201

These are the results obtained with 6 node cluster.
5 Worker	           MASTER	DML_ON_REF
INSERT	                 5035	5070
UPDATE	                    139	135
DELETE	                    136	137

With @onderkalaci we think that TPS decreases on 6-node cluster since we are opening more connections and applying these operations on more worker nodes.

metdos:
@velioglu does this fix #2252?

velioglu:
@metdos Yes, it does. I have added it to the comment of the PR.

PR #2335, opened 2018-08-09
Fix DDL execution bug on MX when search_path is used

### LINKED ISSUE #2328: DDL operations on the tables that are in the search_path are broken for Citus MX
`public` schema is out of the scope of this issue given that it doesn't need explicit schema on the table names.

The root cause of this issue is that we're directly relaying the command executed on the coordinator to the worker nodes [here](https://github.com/citusdata/citus/blob/master/src/backend/distributed/executor/multi_utility.c#L3150): `SendCommandToWorkers(WORKERS_WITH_METADATA, (char *) ddlJob->commandString);` 

This seems like a must have for 8.0

Steps to reproduce is very straighforward:

```SQL
CREATE SCHEMA mx_bug;
SET search_path TO mx_bug ;
CREATE TABLE test (a int);
SELECT create_distributed_table('test', 'a');


ALTER TABLE test ADD COLUMN x INT;
LOG:  issuing BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED;SELECT assign_distributed_transaction_id(0, 18, '2018-08-06 16:17:44.491564+03');
LOG:  issuing BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED;SELECT assign_distributed_transaction_id(0, 18, '2018-08-06 16:17:44.491564+03');
DETAIL:  on server localhost:9703
LOG:  issuing BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED;SELECT assign_distributed_transaction_id(0, 18, '2018-08-06 16:17:44.491564+03');
LOG:  issuing BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED;SELECT assign_distributed_transaction_id(0, 18, '2018-08-06 16:17:44.491564+03');
LOG:  issuing SET citus.enable_ddl_propagation TO 'off'
LOG:  issuing SET citus.enable_ddl_propagation TO 'off'
DETAIL:  on server localhost:9703
LOG:  issuing SET citus.enable_ddl_propagation TO 'off'
LOG:  issuing SET citus.enable_ddl_propagation TO 'off'
LOG:  issuing ALTER TABLE test ADD COLUMN x INT;
LOG:  issuing ALTER TABLE test ADD COLUMN x INT;
DETAIL:  on server localhost:9703
LOG:  issuing ALTER TABLE test ADD COLUMN x INT;
LOG:  issuing ALTER TABLE test ADD COLUMN x INT;
LOG:  issuing ROLLBACK
ERROR:  relation "test" does not exist

-- or create index
postgres=# CREATE INDEX i1 ON test (a);
DEBUG:  building index "i1" on table "test"
LOG:  issuing BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED;SELECT assign_distributed_transaction_id(0, 19, '2018-08-06 16:18:04.064978+03');
LOG:  issuing BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED;SELECT assign_distributed_transaction_id(0, 19, '2018-08-06 16:18:04.064978+03');
DETAIL:  on server localhost:9703
LOG:  issuing BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED;SELECT assign_distributed_transaction_id(0, 19, '2018-08-06 16:18:04.064978+03');
LOG:  issuing BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED;SELECT assign_distributed_transaction_id(0, 19, '2018-08-06 16:18:04.064978+03');
LOG:  issuing SET citus.enable_ddl_propagation TO 'off'
LOG:  issuing SET citus.enable_ddl_propagation TO 'off'
DETAIL:  on server localhost:9703
LOG:  issuing SET citus.enable_ddl_propagation TO 'off'
LOG:  issuing SET citus.enable_ddl_propagation TO 'off'
LOG:  issuing CREATE INDEX i1 ON test (a);
LOG:  issuing CREATE INDEX i1 ON test (a);
DETAIL:  on server localhost:9703
LOG:  issuing CREATE INDEX i1 ON test (a);
LOG:  issuing CREATE INDEX i1 ON test (a);
LOG:  issuing ROLLBACK
ERROR:  relation "test" does not exist
```

The workaround is to explicitly add the schema to the query such as:

```SQL
ALTER TABLE mx_bug.test ADD COLUMN x INT;

CREATE INDEX i1 ON mx_bug.test (a);
```

The quickest solution seems to add `search_path` to the points where we send the DDL to the tables with `SendCommandToWorkers(WORKERS_WITH_METADATA, ...)`

onderkalaci:
Fixes #2328 

Make sure that the coordinator sends the commands when the search_path synchronized with the coordinator's search_path. This is only important when Citus sends the commands that are directly relayed to the worker nodes. For example, the DLL commands which are send via `worker_apply_shard_ddl_command` or `SELECT/DML` queries always adds schema qualifications to the queries. So, they do not benefit from this change. This is only useful for the queries that are executed on the actual table, not the shards.

### DIFF
--- a/src/backend/distributed/executor/multi_utility.c
+++ b/src/backend/distributed/executor/multi_utility.c
@@ -158,6 +158,8 @@ static bool IsIndexRenameStmt(RenameStmt *renameStmt);
 static bool AlterInvolvesPartitionColumn(AlterTableStmt *alterTableStatement,
 										 AlterTableCmd *command);
 static void ExecuteDistributedDDLJob(DDLJob *ddlJob);
+static char * SetSearchPathToCurrentSearchPathCommand(void);
+static char * CurrentSearchPath(void);
 static List * CreateIndexTaskList(Oid relationId, IndexStmt *indexStmt);
 static List * DropIndexTaskList(Oid relationId, Oid indexId, DropStmt *dropStmt);
 static List * InterShardDDLTaskList(Oid leftRelationId, Oid rightRelationId,
@@ -3146,7 +3148,19 @@ ExecuteDistributedDDLJob(DDLJob *ddlJob)
 	{
 		if (shouldSyncMetadata)
 		{
+			char *setSearchPathCommand = SetSearchPathToCurrentSearchPathCommand();
+
 			SendCommandToWorkers(WORKERS_WITH_METADATA, DISABLE_DDL_PROPAGATION);
+
+			/*
+			 * Given that we're relaying the query to the worker nodes directly,
+			 * we should set the search path exactly the same when necessary.
+			 */
+			if (setSearchPathCommand != NULL)
+			{
+				SendCommandToWorkers(WORKERS_WITH_METADATA, setSearchPathCommand);
+			}
+
 			SendCommandToWorkers(WORKERS_WITH_METADATA, (char *) ddlJob->commandString);
 		}
 
@@ -3173,8 +3187,19 @@ ExecuteDistributedDDLJob(DDLJob *ddlJob)
 
 			if (shouldSyncMetadata)
 			{
-				List *commandList = list_make2(DISABLE_DDL_PROPAGATION,
-											   (char *) ddlJob->commandString);
+				List *commandList = list_make1(DISABLE_DDL_PROPAGATION);
+				char *setSearchPathCommand = SetSearchPathToCurrentSearchPathCommand();
+
+				/*
+				 * Given that we're relaying the query to the worker nodes directly,
+				 * we should set the search path exactly the same when necessary.
+				 */
+				if (setSearchPathCommand != NULL)
+				{
+					commandList = lappend(commandList, setSearchPathCommand);
+				}
+
+				commandList = lappend(commandList, (char *) ddlJob->commandString);
 
 				SendBareCommandListToWorkers(WORKERS_WITH_METADATA, commandList);
 			}
@@ -3190,6 +3215,77 @@ ExecuteDistributedDDLJob(DDLJob *ddlJob)
 		}
 		PG_END_TRY();
 	}
+}
+
+
+/*
+ * SetSearchPathToCurrentSearchPathCommand generates a command which can
+ * set the search path to the exact same search path that the issueing node
+ * has.
+ *
+ * If the current search path is null (or doesn't have any valid schemas),
+ * the function returns NULL.
+ */
+static char *
+SetSearchPathToCurrentSearchPathCommand(void)
+{
+	StringInfo setCommand = NULL;
+	char *currentSearchPath = CurrentSearchPath();
+
+	if (currentSearchPath == NULL)
+	{
+		return NULL;
+	}
+
+	setCommand = makeStringInfo();
+	appendStringInfo(setCommand, "SET search_path TO %s;", currentSearchPath);
+
+	return setCommand->data;
+}
+
+
+/*
+ * CurrentSearchPath is a C interface for calling current_schemas(bool) that
+ * PostgreSQL exports.
+ *
+ * CurrentSchemas returns all the schemas in the seach_path that are seperated
+ * with comma (,) sign. The returned string can be used to set the search_path.
+ *
+ * The function omits implicit schemas.
+ *
+ * The function returns NULL if there are no valid schemas in the search_path,
+ * mimicing current_schemas(false) function.
+ */
+static char *
+CurrentSearchPath(void)
+{
+	StringInfo currentSearchPath = makeStringInfo();
+	List *searchPathList = fetch_search_path(false);
+	ListCell *searchPathCell;
+	bool schemaAdded = false;
+
+	foreach(searchPathCell, searchPathList)
+	{
+		char *schemaName = get_namespace_name(lfirst_oid(searchPathCell));
+
+		/* watch out for deleted namespace */
+		if (schemaName)
+		{
+			if (schemaAdded)
+			{
+				appendStringInfoString(currentSearchPath, ",");
+				schemaAdded = false;
+			}
+
+			appendStringInfoString(currentSearchPath, quote_identifier(schemaName));
+			schemaAdded = true;
+		}
+	}
+
+	/* fetch_search_path() returns a palloc'd list that we should free now */
+	list_free(searchPathList);
+
+	return (currentSearchPath->len > 0 ? currentSearchPath->data : NULL);
 }
 
 
--- a/src/test/regress/expected/multi_mx_schema_support.out
+++ b/src/test/regress/expected/multi_mx_schema_support.out
@@ -373,3 +373,81 @@ WHERE
 
 -- set task_executor back to real-time
 SET citus.task_executor_type TO "real-time";
+-- connect to the master and do some test
+-- regarding DDL support on schemas where
+-- the search_path is set
+\c - - - :master_port
+CREATE SCHEMA mx_ddl_schema_1;
+CREATE SCHEMA mx_ddl_schema_2;
+CREATE SCHEMA "CiTuS.TeAeN";
+SET citus.shard_count TO 4;
+SET citus.shard_replication_factor TO 1;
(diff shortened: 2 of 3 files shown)


src/backend/distributed/executor/multi_utility.c
```diff
@@ -3146,7 +3148,19 @@ ExecuteDistributedDDLJob(DDLJob *ddlJob)
...
+			/*
+			 * Given that we're relaying the query to the worker nodes directly,
+			 * we should set the search path exactly the same when necessary.
+			 */
+			if (setSearchPathCommand != NULL)
+			{
+				SendCommandToWorkers(WORKERS_WITH_METADATA, setSearchPathCommand);
+			}
+
 			SendCommandToWorkers(WORKERS_WITH_METADATA, (char *) ddlJob->commandString);
```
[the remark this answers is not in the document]
onderkalaci: But `SendBareCommandListToWorkers ()` executes outside of the distributed/coordinated transaction whereas `SendCommandToWorkers ` executes inside the distributed/coordinated transaction. So, that'd break the logic.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -3173,8 +3187,24 @@ ExecuteDistributedDDLJob(DDLJob *ddlJob)
...
+				char *setSearchPathCommand = SetSearchPathToCurrentSearchPathCommand();
+				List *commandList = NIL;
+
+				/*
+				 * Given that we're relaying the query to the worker nodes directly,
+				 * we should set the search path exactly the same when necessary.
+				 */
+				if (setSearchPathCommand != NULL)
+				{
+					commandList = list_make3(DISABLE_DDL_PROPAGATION,
```
mtuncer: I would have preferred
```
list_make1()
if (setSearchPathCommand)
{
    lappend()
}
lappend()
```

for better readability.

src/backend/distributed/executor/multi_utility.c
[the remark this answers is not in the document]
mtuncer: ok

src/backend/distributed/executor/multi_utility.c
```diff
@@ -3193,6 +3223,77 @@ ExecuteDistributedDDLJob(DDLJob *ddlJob)
...
+ * has.
+ *
+ * If the current search path is null (or doesn't have any valid schemas),
+ * the function returns NULL.
+ */
+static char *
+SetSearchPathToCurrentSearchPathCommand(void)
+{
+	StringInfo setCommand = NULL;
+	char *currentSchemas = CurrentSearchPath();
```
mtuncer: `currentSearchPath` ?

mtuncer:
hit some nasty issue, test case is like you are intentionally trying to shoot yourself in the foot but I felt like reporting any way.

- coordinator
  - set replication model to streaming
  - start metadata sync to worker 1
  - create schema s1, s2;
  - create table s1.t1

- worker 1
  - create schema s2
  - create table s2.t1(a int, b int)

- coordinator
  - set search path to s2, s1
  - alter table t1 add column c int
  - \d s1.t1 --> see new column in s1.t1 

- worker 1
  - \d s1.t1   --> column is not added
  - \d s2.t1  --> wrong table is altered

onderkalaci:
> hit some nasty issue, test case is like you are intentionally trying to shoot yourself in the foot but I felt like reporting any way.

We've discussed this with @mtuncer. It looks like these type of nasty things are possible if users create local tables on the worker nodes. I think we should document and discourage users from doing that. They can obviously create local tables, but, tables with the same names as distributed tables are prone to various edge cases such as this one.

So, we're both leaning towards leaving this issue as a known thing and not attempt to fix. Note that the fix seems a little bit involved so it doesn't worth fixing it.

src/backend/distributed/executor/multi_utility.c
onderkalaci: makes sense.

src/backend/distributed/executor/multi_utility.c
onderkalaci: done

src/backend/distributed/executor/multi_utility.c
```diff
@@ -3193,6 +3223,77 @@ ExecuteDistributedDDLJob(DDLJob *ddlJob)
...
+	{
+		char *schemaName = NULL;
+
+		if (schemaAdded)
+		{
+			appendStringInfoString(currentSchemas, ",");
+		}
+
+		/* watch out for deleted namespace */
+		schemaName = get_namespace_name(lfirst_oid(searchPathCell));
```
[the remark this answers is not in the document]
onderkalaci: But I don't want to add `,` at the end of the string. So, I kept the loop but set `schemaAdded ` to false if `schemaName == NULL`. Does that look good?

(I actually copy & paste'd this code from Postgres code. I cannot find a case where `schemaName` is null. But, keeping it here just in case.)

src/backend/distributed/executor/multi_utility.c
```diff
@@ -3193,6 +3218,81 @@ ExecuteDistributedDDLJob(DDLJob *ddlJob)
...
+		if (schemaName)
+		{
+			appendStringInfoString(currentSearchPath, quote_identifier(schemaName));
+
+			schemaAdded = true;
+		}
+		else
```
mtuncer: This makes it a lot less readable. 

How about
```c
	/* watch out for deleted namespace */
	schemaName = get_namespace_name(lfirst_oid(searchPathCell));
	if (schemaName)
	{
		if (schemaAdded)
		{
			appendStringInfoString(currentSearchPath, ",");
		}
		
		appendStringInfoString(currentSearchPath, quote_identifier(schemaName));
		schemaAdded = true;
	}
```

onderkalaci: Yes, this is much better. I'll update the code accordingly

onderkalaci: Please see my new code-style now

src/backend/distributed/executor/multi_utility.c
```diff
@@ -3193,6 +3218,76 @@ ExecuteDistributedDDLJob(DDLJob *ddlJob)
...
+ * The function returns NULL if there are no valid schemas in the search_path,
+ * mimicing current_schemas(false) function.
+ */
+static char *
+CurrentSearchPath(void)
+{
+	StringInfo currentSearchPath = makeStringInfo();
+	List *searchPathList = fetch_search_path(false);
+	ListCell *searchPathCell;
+	bool addComma = false;
```
mtuncer: I would stick with `schemaAdded`, but this is also good

src/backend/distributed/executor/multi_utility.c
```diff
@@ -3193,6 +3218,77 @@ ExecuteDistributedDDLJob(DDLJob *ddlJob)
...
+	{
+		char *schemaName = get_namespace_name(lfirst_oid(searchPathCell));
+
+		/* watch out for deleted namespace */
+		if (schemaName)
+		{
+			if (schemaAdded)
+			{
+				appendStringInfoString(currentSearchPath, ",");
+				schemaAdded = false;
```
furkansahin: Just a very minor note: I think `schemaAdded = false;`is not necessary here and it confused me a bit, tbh.

onderkalaci: I agree that it is a little bit confusing. The name should probably have been `addCommaSeperator`.

But, I think it is necessary to have that variable  because we cannot always get a schema in every iteration. So, how could we reset the variable otherwise?

furkansahin: The thing is, even if we set it to false here, it is immediately set back to true just outside of the curly bracket. I think the check is needed only for the first iteration, don't you think so?

onderkalaci: Ah, I see. Yes, that seems correct.

PR #2339, opened 2018-08-15
Make sure that modifying CTEs always use the correct execution mode

### LINKED ISSUE #2338: Modifying CTEs do not properly switch the execution mode
As the below example show, after `t1` executed, the query execution mode should have been switched to sequential. However, it does not and we got into a self-deadlock.

In general, modifying CTEs may cause such issues given that the decision to use `parallel` vs `sequential` mode has been given before we've executed the subplans.

```SQL
-- set the mode to sequential for the next operations
CREATE TABLE reference_table(id int PRIMARY KEY);

CREATE TABLE distributed_table(id int PRIMARY KEY, value_1 int);
SELECT create_distributed_table('distributed_table', 'id');

ALTER TABLE 
	distributed_table 
ADD CONSTRAINT 
REFERENCES 

INSERT INTO reference_table SELECT i FROM generate_series(0, 10) i;
INSERT INTO distributed_table SELECT i, i % 10  FROM generate_series(0, 100) i;

WITH t1 AS (DELETE FROM reference_table RETURNING id) 
	DELETE FROM distributed_table USING t1 WHERE value_1 = t1.id RETURNING *;

```

onderkalaci:
Fixes #2338

We're simply deferring the execution mode decision right after `ExecuteSubPlans ()`. 

Could be backported but doesn't seem like a very critical fix to have a point release only for this (as @marcocitus also thinks similarly).

### DIFF
--- a/src/backend/distributed/executor/citus_custom_scan.c
+++ b/src/backend/distributed/executor/citus_custom_scan.c
@@ -88,19 +88,10 @@ static CustomExecMethods TaskTrackerCustomExecMethods = {
 	.ExplainCustomScan = CitusExplainScan
 };
 
-static CustomExecMethods RouterSequentialModifyCustomExecMethods = {
-	.CustomName = "RouterSequentialModifyScan",
+static CustomExecMethods RouterModifyCustomExecMethods = {
+	.CustomName = "RouterModifyScan",
 	.BeginCustomScan = CitusModifyBeginScan,
-	.ExecCustomScan = RouterSequentialModifyExecScan,
-	.EndCustomScan = CitusEndScan,
-	.ReScanCustomScan = CitusReScan,
-	.ExplainCustomScan = CitusExplainScan
-};
-
-static CustomExecMethods RouterMultiModifyCustomExecMethods = {
-	.CustomName = "RouterMultiModifyScan",
-	.BeginCustomScan = CitusModifyBeginScan,
-	.ExecCustomScan = RouterMultiModifyExecScan,
+	.ExecCustomScan = RouterModifyExecScan,
 	.EndCustomScan = CitusEndScan,
 	.ReScanCustomScan = CitusReScan,
 	.ExplainCustomScan = CitusExplainScan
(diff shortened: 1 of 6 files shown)


src/backend/distributed/executor/citus_custom_scan.c
```diff
@@ -221,20 +212,7 @@ RouterCreateScan(CustomScan *scan)
...
-			 * to 'sequential' or Multi-row INSERT are executed sequentially
-			 * instead of using parallel connections.
-			 */
-			scanState->customScanState.methods = &RouterSequentialModifyCustomExecMethods;
-		}
-		else
-		{
-			scanState->customScanState.methods = &RouterMultiModifyCustomExecMethods;
-		}
+		scanState->customScanState.methods = &RouterModifyCustomExecMethods;
```
marcocitus: I think the distinction between this block and the one above is no longer relevant.

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -615,6 +583,77 @@ RouterSequentialModifyExecScan(CustomScanState *node)
...
+	{
+		Task *task = (Task *) lfirst(taskCell);
+
+		/*
+		 * Result is expected for SELECT ... FOR UPDATE queries as well.
+		 */
+		executorState->es_processed +=
+			ExecuteSingleModifyTask(scanState, task, operation,
+									alwaysThrowErrorOnFailure,
+									hasReturning || task->relationRowLockList != NIL);
```
marcocitus: As a minor clean-up while we're doing this, it might be nice to declare an `expectResults` variable. It's also not obvious that the comment above applies to this line.

src/backend/distributed/executor/citus_custom_scan.c
onderkalaci: makes sense

onderkalaci:
> @marcocitus I've applied your feedback, could you have a quick look again?

@marcocitus Do you want to have a quick look, or shall I go ahead and merge?

Thanks!

Issue #2336, opened 2018-08-10
Can not test queries from MX nodes with isolation tests


velioglu:
In isolation tests, we can not test queries sent from MX nodes.

In order to catch distributed locks within isolation test logic, we have implemented UDFs to understand that there is a distributed transaction exists which preventing our coordinator session from getting lock. Yet, that logic only works when the distributed transaction is started from coordinator. We have implemented our distributed deadlock detection to run separate processes from each MX node to make detection logic more scalable. Since we can not connect to worker nodes in the isolation tests, we can not understand that two different MX workers blocking each other from the coordinator node.

We are using `get_all_active_transactions` for getting the list of transaction which has been initiated from the coordinator, then by comparing the obtained transaction IDs with the results of `dumb_global_wait_edges` enables us to get the transaction id preventing the session from getting lock. In order to understand that two workers blocking each other, we need to have a kind of `get_global_all_active_transactions` to use it in our isolation tests.

Since we don't have such an infrastructure, we couldn't implemented intended tests in 
- Distributed deadlock PRs
- Writing to reference tables from MX nodes PR (#2333)

Considering we are planning to improve capabilities of our MX nodes, having chance to test new functionalities in our isolation tests will definitely make following changes more robust.

velioglu:
Following tests should be added after implementing this UDF:

- Isolation tests on #2333 
- Basic deadlock detection
- Online rebalancer
- Tenant isolation
- DROP TABLE / SELECT from Worker Node
- TRUNCATE TABLE (both from Worker and Coordinator) / SELECT from Worker Node

PR #2343, opened 2018-08-19
Make sure that table (and metadata) is dropped before shards are dropped on Citus MX

### LINKED ISSUE #1779: concurrent alter/drop table and select may cause undetected deadlock in mx workers
I slowed down alter table command using an event trigger on an mx worker prior to running this test.

- coordinator issues alter table add column command, waits for worker operation to complete
- another connection to coordinator issues drop table command, gets blocked by alter table command
- a worker connection issues select command on the table being dropped, gets blocked.

- alter table command completes after a while, drop and select commands keep blocking each other

- dump_global_wait_edges() shows that node (0) blocked by on node(2) (who issued select) 
- dump_local_wait_edges() on node(2) shows node(0) blocked on node(2).

- killing select query unblocks drop command

onderkalaci:
DESCRIPTION: Prevent a deadlock on concurrent DROP tbl and SELECT on Citus MX

Fixes #1779

This commit fixes a bug where a concurrent DROP TABLE deadlocks
with SELECT (or DML) when the SELECT is executed from the workers.

The problem was that Citus used to remove the metadata before
dropping the table on the workers. That creates a time window
where the SELECT starts running on some of the nodes and DROP
table on some of the other nodes.

The bug can be reproduced either as described on #1779 or
using `gdb`. To repro with `gdb`:

1) Connect to the coordinator, get the pid
2) Attach to the process (1)
3) add a breakpoint on `SendCommandToWorkersParams()`
4) Execute `DROP TABLE ` command
5) After the first `SendRemoteCommandParams()` executed in
`SendCommandToWorkersParams()`, start a new session to any 
(or all) worker node(s)
6) Run `SELECT * FROM table`
7) `continue` on `gdb`. Before this patch, you'd observe the deadlock
after the patch you shouldn't.

### DIFF
--- a/src/backend/distributed/citus--8.0-2--8.0-3.sql
+++ b/src/backend/distributed/citus--8.0-2--8.0-3.sql
@@ -1,0 +1,64 @@
+/* citus--8.0-2--8.0-3 */
+SET search_path = 'pg_catalog';
+
+CREATE FUNCTION master_remove_partition_metadata(logicalrelid regclass,
+                                                   schema_name text,
+                                                   table_name text)
+    RETURNS void
+    LANGUAGE C STRICT
+    AS 'MODULE_PATHNAME', $$master_remove_partition_metadata$$;
+COMMENT ON FUNCTION master_remove_partition_metadata(logicalrelid regclass,
+                                                       schema_name text,
+                                                       table_name text)
+    IS 'deletes the partition metadata of a distributed table';
+
+CREATE OR REPLACE FUNCTION master_remove_distributed_table_metadata_from_workers(logicalrelid regclass,
+                                                   schema_name text,
+                                                   table_name text)
+    RETURNS void
+    LANGUAGE C STRICT
+    AS 'MODULE_PATHNAME', $$master_remove_distributed_table_metadata_from_workers$$;
+COMMENT ON FUNCTION master_remove_distributed_table_metadata_from_workers(logicalrelid regclass,
+                                                       schema_name text,
+                                                       table_name text)
+    IS 'drops the table and removes all the metadata belonging the distributed table in the worker nodes with metadata.';
+
+
+CREATE OR REPLACE FUNCTION pg_catalog.citus_drop_trigger()
+    RETURNS event_trigger
+    LANGUAGE plpgsql
+    SET search_path = pg_catalog
+    AS $cdbdt$
+DECLARE
+    v_obj record;
+    sequence_names text[] := '{}';
+    table_colocation_id integer;
+    propagate_drop boolean := false;
+BEGIN
+    -- collect set of dropped sequences to drop on workers later
+    SELECT array_agg(object_identity) INTO sequence_names
+    FROM pg_event_trigger_dropped_objects()
+    WHERE object_type = 'sequence';
+
+    FOR v_obj IN SELECT * FROM pg_event_trigger_dropped_objects()
+                 WHERE object_type IN ('table', 'foreign table')
+    LOOP
+        -- first drop the table and metadata on the workers
+        -- then drop all the shards on the workers
+        -- finally remove the pg_dist_partition entry on the coordinator
+        PERFORM master_remove_distributed_table_metadata_from_workers(v_obj.objid, v_obj.schema_name, v_obj.object_name);
+        PERFORM master_drop_all_shards(v_obj.objid, v_obj.schema_name, v_obj.object_name);
+        PERFORM master_remove_partition_metadata(v_obj.objid, v_obj.schema_name, v_obj.object_name);
+    END LOOP;
+
+    IF cardinality(sequence_names) = 0 THEN
+        RETURN;
+    END IF;
+
+    PERFORM master_drop_sequences(sequence_names);
+END;
+$cdbdt$;
+COMMENT ON FUNCTION pg_catalog.citus_drop_trigger()
+    IS 'perform checks and actions at the end of DROP actions';
+
+RESET search_path;
--- a/src/backend/distributed/commands/drop_distributed_table.c
+++ b/src/backend/distributed/commands/drop_distributed_table.c
@@ -20,21 +20,46 @@
 #include "utils/lsyscache.h"
 
 
+/* local function forward declarations */
+static void MasterRemoveDistributedTableMetadataFromWorkers(Oid relationId,
+															char *schemaName,
+															char *tableName);
+
+
 /* exports for SQL callable functions */
 PG_FUNCTION_INFO_V1(master_drop_distributed_table_metadata);
+PG_FUNCTION_INFO_V1(master_remove_partition_metadata);
+PG_FUNCTION_INFO_V1(master_remove_distributed_table_metadata_from_workers);
 
 
 /*
- * master_drop_distributed_table_metadata removes the entry of the specified distributed
- * table from pg_dist_partition and drops the table from the workers if needed.
+ * master_drop_distributed_table_metadata UDF is a stub UDF to install Citus flawlessly.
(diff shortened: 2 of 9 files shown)


onderkalaci:
After #2336 we can add a single test for this.

src/backend/distributed/citus--8.0-2--8.0-3.sql
```diff
@@ -0,0 +1,66 @@
+/* citus--8.0-2--8.0-3 */
+SET search_path = 'pg_catalog';
+
+DROP FUNCTION IF EXISTS master_drop_distributed_table_metadata(regclass, text, text);
```
mtuncer: This is the first time we are introducing this function.

src/backend/distributed/citus--8.0-2--8.0-3.sql
```diff
@@ -0,0 +1,66 @@
...
+
+CREATE OR REPLACE FUNCTION master_remove_partition_metadata(logicalrelid regclass,
```
mtuncer: may be just `CREATE FUNCTION`

src/backend/distributed/citus--8.0-2--8.0-3.sql
```diff
@@ -0,0 +1,66 @@
...
+                                                   schema_name text,
+                                                   table_name text)
+    RETURNS void
+    LANGUAGE C STRICT
+    AS 'MODULE_PATHNAME', $$master_remove_partition_metadata$$;
+COMMENT ON FUNCTION master_remove_partition_metadata(logicalrelid regclass,
+                                                       schema_name text,
+                                                       table_name text)
+    IS 'delete the partition metadata of the distributed table';
```
mtuncer: s/delete/deletes

mtuncer: s/the/a

src/backend/distributed/citus--8.0-2--8.0-3.sql
```diff
@@ -0,0 +1,66 @@
...
+CREATE OR REPLACE FUNCTION master_drop_distributed_table_metadata_on_workers(logicalrelid regclass,
+                                                   schema_name text,
+                                                   table_name text)
+    RETURNS void
+    LANGUAGE C STRICT
+    AS 'MODULE_PATHNAME', $$master_drop_distributed_table_metadata_on_workers$$;
+COMMENT ON FUNCTION master_drop_distributed_table_metadata_on_workers(logicalrelid regclass,
+                                                       schema_name text,
+                                                       table_name text)
+    IS 'drops the table and removes all the metadata beloning the distributed table in the worker nodes with metadata.';
```
mtuncer: s/beloning/belonging

src/backend/distributed/citus--8.0-2--8.0-3.sql
```diff
@@ -0,0 +1,66 @@
...
+
+    IF cardinality(sequence_names) = 0 THEN
+        RETURN;
+    END IF;
+
+    PERFORM master_drop_sequences(sequence_names);
+END;
+$cdbdt$;
+COMMENT ON FUNCTION pg_catalog.citus_drop_trigger()
+    IS 'perform checks and actions at the end of DROP actions';
```
mtuncer: s/perform/performs

src/backend/distributed/citus--8.0-2--8.0-3.sql
```diff
@@ -0,0 +1,66 @@
...
+                                                   table_name text)
+    RETURNS void
+    LANGUAGE C STRICT
+    AS 'MODULE_PATHNAME', $$master_remove_partition_metadata$$;
+COMMENT ON FUNCTION master_remove_partition_metadata(logicalrelid regclass,
+                                                       schema_name text,
+                                                       table_name text)
+    IS 'delete the partition metadata of the distributed table';
+
+CREATE OR REPLACE FUNCTION master_drop_distributed_table_metadata_on_workers(logicalrelid regclass,
```
mtuncer: what about
```
master_remove_distributed_table_metadata_from_workers
```

src/backend/distributed/citus--8.0-2--8.0-3.sql
```diff
@@ -0,0 +1,66 @@
...
+    SELECT array_agg(object_identity) INTO sequence_names
+    FROM pg_event_trigger_dropped_objects()
+    WHERE object_type = 'sequence';
+
+    FOR v_obj IN SELECT * FROM pg_event_trigger_dropped_objects()
+                 WHERE object_type IN ('table', 'foreign table')
+    LOOP
+        -- first drop the table and metadata on the workers
+        -- then drop all the shards on the workers
+        -- finally remove the pg_dist_partition entry on the coordinator
```
mtuncer: do we need to ensure this trigger is called from coordinator ?

src/backend/distributed/commands/drop_distributed_table.c
```diff
@@ -20,21 +20,45 @@
...
+ * common operation in the code-base.
+ *
+ * This function is basically replaced with
+ * master_drop_distributed_table_metadata_on_workers() followed by
+ * master_remove_partition_metadata().
  */
 Datum
 master_drop_distributed_table_metadata(PG_FUNCTION_ARGS)
+{
+	ereport(DEBUG2, (errmsg("this function is deprecated and no longer is used")));
```
mtuncer: `WARNING` or `INFO` would have been better. Caller will not see the error message.

mtuncer: Another note. This is a major version. It should be okay to remove functions.

mtuncer:
verified that it fixes the original bug

src/backend/distributed/commands/drop_distributed_table.c
```diff
@@ -20,21 +20,45 @@
...
+	PG_RETURN_VOID();
+}
+
+
+/*
+ * master_remove_partition_metadata removes the entry of the specified distributed
+ * table from pg_dist_partition.
+ */
+Datum
+master_remove_partition_metadata(PG_FUNCTION_ARGS)
```
mtuncer: I did not like `partition` here. It could be confused with logical partitioning.

mtuncer: I wonder if we can merge this functionality with `master_drop_all_shards`. Since these `master_drop_distributed_table_metadata_on_workers` and `master_remove_partition_metadata` are used with that function

mtuncer:
DESCRIPTION is missing.

src/backend/distributed/commands/drop_distributed_table.c
onderkalaci: > I did not like partition here. It could be confused with logical partitioning.

Well, the metadata table is `pg_dist_partition`. I think that's the main issue we should consider fixing in the future 😄  If you have strong opinions on this, feel free to propose a name for the function.

Also, many people refer to partitioning as [`declarative partitioning`](https://www.postgresql.org/docs/10/static/ddl-partitioning.html) in the Postgres community. So, should be OK?

> I wonder if we can merge this functionality with master_drop_all_shards. Since these master_drop_distributed_table_metadata_on_workers and master_remove_partition_metadata are used with that function

I prefer to keep the functions separately for two reasons: 
- This seems easier/clearer to read since otherwise, the reader might have harder time to understand where the `pg_dist_partition` entry is removed
- The flow of the code makes sense: (i) `DROP table` drops the table on the coordinator and that's why we're on the trigger here (ii) Drop the table and the corresponding metadata on the MX workers -- if the cluster is MX (iii) drop the shards (iv) remove `pg_dist_partition` entry on the coordinator

onderkalaci:
@mtuncer Could you have another look? Thanks!

Issue #2320, opened 2018-07-31
start_metadata_sync_to_node() doesn't sync `hasmetadata` flag on existing worker's


onderkalaci:
Seems not critical, opening mostly for reference.

`start_metadata_sync_to_node()` should set the `hasmetadata` flag on the existing worker nodes. Note that `hasmetadata` field is currently only used on the coordinator node, so not syncing to the data nodes is not a problem as of 7.5.

Steps to reproduce:

```SQL
-- connect to the coordinator, add two worker node
SELECT master_add_node('localhost', 9700);
SELECT master_add_node('localhost', 9701);

-- start metadata sync
SELECT start_metadata_sync_to_node('localhost', 9700);
SELECT start_metadata_sync_to_node('localhost', 9701);

-- get the metadata table on the coordinator
SELECT nodename, nodeport, hasmetadata FROM pg_dist_node;

 nodename  | nodeport | hasmetadata 
-----------+----------+-------------
 localhost |     9700 | t
 localhost |     9701 | t
(2 rows)

-- connect to the first worker and get it
\c - - - 9700
SELECT nodename, nodeport, hasmetadata FROM pg_dist_node;
 nodename  | nodeport | hasmetadata 
-----------+----------+-------------
 localhost |     9701 | f
 localhost |     9700 | t
(2 rows)


-- connect to the second worker and get it
\c - - - 9701
SELECT nodename, nodeport, hasmetadata FROM pg_dist_node;
 nodename  | nodeport | hasmetadata 
-----------+----------+-------------
 localhost |     9700 | t
 localhost |     9701 | t
(2 rows)
```

As seen above, the metadata is out of sync for this field on the node with port `9700`.

marcocitus:
I believe this was a conscious design decision to not add unnecessary complexity and exposure to failures of unrelated nodes in `start_metadata_sync_to_node`. 

One fundamental question in the design MX was: If you have HA+MX, how do you recover from multiple node failures, knowing that the only path to recovery is eventually updating the metadata on all nodes to point to the new primaries? Any operation that requires all nodes to be up will not be usable in such a scenario.

onderkalaci:
Just a note that this could be tricky if we want to use `hasmetadata ` flag from the worker nodes. To give an example, we've first implemented #2345 but soon realized that due to this restriction (?), we should be implementing it like #2370 .

marcocitus:
We will need to address this to enable features such as:
- DDL on distributed tables from workers
[...]

PR #2345, opened 2018-08-20
Support TRUNCATE from the MX worker nodes

### LINKED ISSUE #2344: Enable `TRUNCATE distributed/reference table` on MX workers
Currently, Citus errors out:

```SQL
TRUNCATE users_table;
```

This is disabled because it is likely that a concurrent `SELECT/DML` on another node could create a distributed deadlock. The distributed deadlock could happen with the following steps:

```
1. A user connects to node-1 and executes `TRUNCATE table`;
2. `TRUNCATE ...` is executed on the `shard-1`
3. Another user connects to node-2 and executes `BEGIN; SELECT count(*) FROM table`
4. `BEGIN; SELECT ...` is executed on `shard-2`
5. `TRUNCATE ...` is blocked on the `shard-2` since the `SELECT ...` is still active
6. `SELECT ...` is blocked on the `shard-1` since the `TRUNCATE ...` is still active
7. Distributed deadlock is detected and one of the transactions is cancelled
```

onderkalaci:
DESCRIPTION: Enable TRUNCATE command support from MX worker nodes

Fixes #2344

This commit enables support for TRUNCATE on both distributed table and reference tables.
    
The basic idea is to acquire lock on the relation by sending the TRUNCATE command to all metadata worker nodes. We only skip sending the TRUNCATE command to the node that executes the command to prevent a self-distributed-deadlock.

After #2336 we should add an isolation test.

(Note that we've initially considered acquiring an explicit lock on the table via `LOCK TABLE ... `. However, this approach seems more straight-forward and generic enough to be applied in other cases as well.)


src/include/distributed/worker_transaction.h
```diff
@@ -22,6 +22,7 @@
 typedef enum TargetWorkerSet
 {
 	WORKERS_WITH_METADATA,
+	WORKERS_WITH_METADATA_EXCLUDE_CURRENT_WORKER,
```
onderkalaci: I'm open for alternative names for this enum. It seems too long now.

mtuncer:
missing DESCRIPTION

src/include/distributed/worker_transaction.h
mtuncer: `OTHER_WORKERS_WITH_METADATA`  ?

src/backend/distributed/master/master_modify_multiple_shards.c
```diff
@@ -196,6 +203,30 @@ master_modify_multiple_shards(PG_FUNCTION_ARGS)
...
 
+	/*
+	 * We should execute "TRUNCATE table_name;" on the other worker nodes before
+	 * executing the truncate commands on the shards. This is necessary to prevent
+	 * distributed deadlocks where a concurrent operation on the same table (or a
+	 * cascading table) is executed on the other nodes.
+	 *
+	 * Note that we should skip the current node to prevent a self-deadlock.
+	 */
+	if (ShouldSyncTableMetadata(relationId) && truncateOperation)
```
mtuncer: minor, consider reversing arguments here. I think having truncateOperation check first makes it more readable

src/include/distributed/worker_transaction.h
mtuncer: I suggested `OTHER_WORKERS_WITH_METADATA` but it has not come here.

src/backend/distributed/transaction/worker_transaction.c
```diff
@@ -117,7 +117,15 @@ SendBareCommandListToWorkers(TargetWorkerSet targetWorkerSet, List *commandList)
 		int nodePort = workerNode->workerPort;
 		int connectionFlags = FORCE_NEW_CONNECTION;
 
-		if (targetWorkerSet == WORKERS_WITH_METADATA && !workerNode->hasMetadata)
+		if ((targetWorkerSet == WORKERS_WITH_METADATA ||
```
onderkalaci: Ops, this approach looks wrong :/  Workers may not know about the other workers. See #2320. 

I think we need to work on this again. We might consider the initial idea where any node issues something like `LOCK TABLE IF EXISTS x IN EXLUSIVE MODE ` to all other nodes.

And, we should never have something like `OTHER_WORKERS_WITH_METADATA `

PR #2346, opened 2018-08-22
Improve query pushdown planning for very large number of shards and queries with many tables

onderkalaci:
This PR consists of two commits.

The first commit is a performance improvement for queries with many tables. Previously, due to a bug in the `ExtractRangeTableRelationWalker `, Citus pulled too returned unnecessary range table entries. Though doesn't change the correctness, that leads to CPU/memory overhead.

The second commit is a performance improvement for an edge case when a query hits too many shards and the placements of all shards reside on a single node. This case is not a realistic one. However, still fixing it for completeness.

### DIFF
--- a/src/backend/distributed/planner/multi_router_planner.c
+++ b/src/backend/distributed/planner/multi_router_planner.c
@@ -2545,6 +2545,12 @@ IntersectPlacementList(List *lhsPlacementList, List *rhsPlacementList)
 						WORKER_LENGTH) == 0)
 			{
 				placementList = lappend(placementList, rhsPlacement);
+
+				/*
+				 * We don't need to add the same placement over and over again. This
+				 * could happen if both placements of a shard appear on the same node.
+				 */
+				break;
 			}
 		}
 	}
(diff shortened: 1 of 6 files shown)


src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -2545,6 +2545,12 @@ IntersectPlacementList(List *lhsPlacementList, List *rhsPlacementList)
 						WORKER_LENGTH) == 0)
 			{
 				placementList = lappend(placementList, rhsPlacement);
+
```
onderkalaci: You can think the change here as following:
```
Previously: 
First iteration: lhs: A,A   rhs:A,A result: A,A,A,A
Second iteration: lhs: A,A,A,A   rhs:A,A result: A,A,A,A,A,A,A,A
...

After the fix:
First iteration: lhs: A,A   rhs:A,A result: A,A
Second iteration: lhs: A,A   rhs:A,A result: A,A
```

As far as I can see, this change doesn't have any effect if `A`s are distinct. It'll always produce the same result.

Also, this fix aligns with the name of the function.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -2545,6 +2545,12 @@ IntersectPlacementList(List *lhsPlacementList, List *rhsPlacementList)
...
+				/*
+				 * We do need to add the same placement over and over again. This
```
pykello: "We do" -> "We don't"

PR #2356, opened 2018-08-30
Re-enable JIT in PostgreSQL 11 tests

jasonmp85:
This test was just pathologically recompiling the function in its WHERE
clause on every row, making execution time seem like an infinite loop.
A small tweak avoids this.

### DIFF
--- a/src/test/regress/sql/multi_distributed_transaction_id.sql
+++ b/src/test/regress/sql/multi_distributed_transaction_id.sql
@@ -96,8 +96,8 @@ SELECT s AS a, s AS b, s AS c, s AS d, s AS e, s AS f, s AS g, s AS h FROM gener
 INSERT INTO parallel_id_test VALUES (1234567), (1234567), (1234568), (1234568);
 ANALYSE parallel_id_test;
 
-SET LOCAL max_parallel_workers_per_gather TO 8;
-SET LOCAL cpu_tuple_cost TO 1000000;
+SET LOCAL max_parallel_workers_per_gather TO 2;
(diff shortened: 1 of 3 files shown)


src/test/regress/sql/multi_distributed_transaction_id.sql
```diff
@@ -96,8 +96,8 @@ SELECT s AS a, s AS b, s AS c, s AS d, s AS e, s AS f, s AS g, s AS h FROM gener
 INSERT INTO parallel_id_test VALUES (1234567), (1234567), (1234568), (1234568);
 ANALYSE parallel_id_test;
 
-SET LOCAL max_parallel_workers_per_gather TO 8;
-SET LOCAL cpu_tuple_cost TO 1000000;
+SET LOCAL max_parallel_workers_per_gather TO 2;
+SET LOCAL parallel_tuple_cost TO 0;
```
metdos: A comment explaining why we are using this value would be nice.

mtuncer: @metdos it is in commit log

mtuncer:
@jasonmp85 unfortunately, can not verify the fix since Citus does not build with PG11 packages at travis.

jasonmp85:
FWIW, `REL_11_BETA4` was just tagged and should be available tomorrow. In addition, it _disables_ JIT by default, after too many complaints it was making certain plans perform worse, I think?

mtuncer:
Still it does no harm to do our changes.  merging it.

PR #2358, opened 2018-08-31
Evaluate functions and parameters anywhere in query

### LINKED ISSUE #2357: Distributed queries directly using arguments in plpgsql functions error out on 6th execution
**Schema**
```
CREATE TABLE test_distributed(shard_key int, value int);
CREATE UNIQUE INDEX u_indx ON test_distributed(shard_key);

SELECT create_distributed_table('test_distributed','shard_key');
```
**Function with argument (observe is_true) directly used as a part of the WHERE clause:**
```
CREATE OR REPLACE FUNCTION test_func(s_key int,val int, is_true boolean)
  RETURNS void
AS $$
BEGIN
        INSERT INTO test_distributed VALUES (s_key,val) ON CONFLICT(shard_key) DO UPDATE
        SET value=excluded.value where test_distributed.shard_key=excluded.shard_key 
        AND is_true=FALSE;
END;
$$
LANGUAGE plpgsql;

SELECT test_func(1,1,'t');
SELECT test_func(1,1,'t');
SELECT test_func(1,1,'t');
SELECT test_func(1,1,'t');
SELECT test_func(1,1,'t');
SELECT test_func(1,1,'t');
--Error out on 6th execution
/*ERROR:  there is no parameter $3
CONTEXT:  while executing command on ec2-54-227-141-82.compute-1.amazonaws.com:5432
SQL statement "INSERT INTO test_distributed VALUES (s_key,val) ON CONFLICT(shard_key) DO UPDATE
        SET value=excluded.value where test_distributed.shard_key=excluded.shard_key AND is_true=FALSE"
PL/pgSQL function test_func(integer,integer,boolean) line 3 at SQL statement
*/
```

**Rewritten function without argument in the where clause**
```
-- removed the argument (is_true) from the where clause
CREATE OR REPLACE FUNCTION test_func(s_key int,val int, is_true boolean)
  RETURNS void
AS $$
BEGIN
IF is_true= FALSE THEN
        INSERT INTO test_distributed VALUES (s_key,val) ON CONFLICT(shard_key) DO UPDATE
        SET value=excluded.value where test_distributed.shard_key=excluded.shard_key;
ELSE
        INSERT INTO test_distributed VALUES (s_key,val) ON CONFLICT(shard_key) DO NOTHING;
END IF;
END;
$$
LANGUAGE plpgsql;
--runs well
SELECT test_func(1,1,'t');
SELECT test_func(1,1,'t');
SELECT test_func(1,1,'t');
SELECT test_func(1,1,'t');
SELECT test_func(1,1,'t');
SELECT test_func(1,1,'t');
SELECT test_func(1,1,'t');
SELECT test_func(1,1,'t');
SELECT test_func(1,1,'t');
SELECT test_func(1,1,'t');
SELECT test_func(1,1,'t');
SELECT test_func(1,1,'t');
```
I think this is related to handling prepared statements. We fixed prepared statements for arguments (based on [#306](https://github.com/citusdata/citus/issues/306) and  [#1134](https://github.com/citusdata/citus/pull/1134) that directly get inserted/updated etc but this case seems to be not handled.

marcocitus:
DESCRIPTION: Evaluate functions anywhere in query

The logic in `RequiresMasterEvaluation` and `ExecuteMasterEvaluableFunctions` was weirdly selective about which parts of the query tree it would process. In general, there is no reason why we wouldn't want to evaluate a stable/immutable function or parameter anywhere in a DML query (except for volatile functions in update/delete, but those currently throw an error).

There are also cases in which we rely on function evaluation even when there are no function calls. In particular, when there is a parameter in the partition column of an INSERT, we evaluate all parameters in the query on the master. We should generally expect parameters to appear anywhere in the query, especially when the query is executed from a PL/pgSQL function.

This PR removes the selectivity and descends into the full query tree, evaluating node types that can be evaluated but don't have a `Var` (column reference) as a child. 

TODO:
- [x] Regression tests.
- [x] Consider `func(var,param)` cases

Fixed #2357

### DIFF
--- a/src/backend/distributed/utils/citus_clauses.c
+++ b/src/backend/distributed/utils/citus_clauses.c
@@ -26,18 +26,8 @@
 #include "utils/lsyscache.h"
 
 
-typedef struct FunctionEvaluationContext
-{
-	PlanState *planState;
-	bool containsVar;
-} FunctionEvaluationContext;
-
-
 /* private function declarations */
-static void EvaluateValuesListsItems(List *valuesLists, PlanState *planState);
-static Node * EvaluateNodeIfReferencesFunction(Node *expression, PlanState *planState);
-static Node * PartiallyEvaluateExpressionMutator(Node *expression,
-												 FunctionEvaluationContext *context);
+static bool IsVarNode(Node *node);
 static Expr * citus_evaluate_expr(Expr *expr, Oid result_type, int32 result_typmod,
 								  Oid result_collation, PlanState *planState);
 static bool CitusIsVolatileFunctionIdChecker(Oid func_id, void *context);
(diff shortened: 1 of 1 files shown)


marcocitus:
We probably also need to be careful about volatile functions in ON CONFLICT sections.

sumedhpathak:
@jasonmp85 Could you look into/review this change?

jasonmp85:
Rebase had no conflicts, so I did one and pushed; doing some testing locally.

jasonmp85:
@marcocitus can you add a test based on #2357 to ensure this new functionality is working?

src/backend/distributed/utils/citus_clauses.c
```diff
@@ -305,27 +91,52 @@ EvaluateNodeIfReferencesFunction(Node *expression, PlanState *planState)
 		case T_ArrayCoerceExpr:
 		case T_ScalarArrayOpExpr:
 		case T_RowCompareExpr:
-		case T_Param:
 		case T_RelabelType:
 		case T_CoerceToDomain:
 		{
+			if (FindNodeCheck(expression, IsVarNode))
```
jasonmp85: I'm not a huge fan of the style of this function (I'd prefer an explicit return variable and more `break` and a single-return and moving the `T_Const` block down here), but since you're not authoring it, just changing it, I guess this is OK.

jasonmp85:
@metdos… I notice you re-assigned this to me; I reviewed it last week already and gave it approval, pending tests.

metdos:
> @metdos… I notice you re-assigned this to me; I reviewed it last week already and gave it approval, pending tests.

@jasonmp85 I just did that to make it easier for me to track remaining PRs :)

metdos:
> @jasonmp85 I just did that to make it easier for me to track remaining PRs :)

Reviewer does not show up on the PRs page.

jasonmp85:
Hey, uh, we want this in the release, right? Any tests yet?

marcocitus:
@sai reported an issue where the coordinator tries to evaluate generate_series. This issue already existed, but was harder to reach because we skipped over parts of the query where generate_series might typically appear. I pushed a commit to fix the issue and also added some tests I had already worked on. Hope this doesn't conflict with anything you were working on @jasonmp85 .

jasonmp85:
> @Sai reported an issue where the coordinator tries to evaluate generate_series. This issue already existed, but was harder to reach because we skipped over parts of the query where generate_series might typically appear.

I forgot about SRFs in my review; good catch.

> I pushed a commit to fix the issue and also added some tests I had already worked on. Hope this doesn't conflict with anything you were working on @jasonmp85

I was adding DML statements to `multi_function_evaluation.sql`, so I'll just add a few of mine into what you've done and merge.

PR #2360, opened 2018-09-03
Views to provide some insight about the distributed transactions on Citus MX

onderkalaci:
With this commit, we implement two views that are very similar
to pg_stat_activity, but showing queries that are involved in
distributed queries:

    - citus_dist_stat_activity: Shows all the distributed queries
    - citus_worker_stat_activity: Shows all the queries on the shards
                                  that are initiated by distributed queries.

Both views have the same columns in the outputs. In very basic terms, both of the views
are meant to provide some useful insights about the distributed
transactions within the cluster. As the names reveal, both views are similar to pg_stat_activity.
Also note that these views can be pretty useful on Citus MX clusters.

Note that when the views are queried from the worker nodes, they'd not show the distributed
transactions that are initiated from the coordinator node. The reason is that the worker
nodes do not know the host/port of the coordinator. Thus, it is advisable to query the
views from the coordinator.

If we bucket the columns that the views returns, we'd end up with the following:

- Hostnames and ports:
   - query_hostname, query_hostport: The node that the query is running
   - master_query_host_name, master_query_host_port: The node in the cluster
                                                   initiated the query.
    Note that for citus_dist_stat_activity view, the query_hostname-query_hostport
    is always the same with master_query_host_name-master_query_host_port. The
    distinction is mostly relevant for citus_worker_stat_activity. For example,
    on Citus MX, a users starts a transaction on Node-A, which starts worker
    transactions on Node-B and Node-C. In that case, the query hostnames would be
    Node-B and Node-C whereas the master_query_host_name would Node-A.

- Distributed transaction related things:
    This is mostly the process_id, distributed transactionId and distributed transaction
    number.

- pg_stat_activity columns:
    These two views get all the columns from pg_stat_activity. We're basically joining
    pg_stat_activity with get_all_active_transactions on process_id.

Some open things for discussion:

- [x] We currently connect to the nodes with the CurrentUser() so that we get
      the same privilage restrictions that pg_stat_activity provides. But, I
      remember that one of the ideas was to connect with SuperUser(). The rationale
      was that SuperUser's mostly have some reserved connection pools. So, in case of
      an emergency situation where the admin wants to access to the transactions, the
      admin would be able to connect. Do we need to take any action?

- [x] In order to get the local results, we could simply call the function. However, for ease
      of implementation, I preferred to connect to the local node as well. Does it worth to
      skip connections to the localhost?

- [x] `TransactionID` and `timestamptz` could be `NULL` in `pg_stat_activity`. But, we set it to `0` those columns when they are `NULL`. We should probably fix that.

### DIFF
--- a/src/backend/distributed/transaction/citus_dist_stat_activity.c
+++ b/src/backend/distributed/transaction/citus_dist_stat_activity.c
@@ -1,0 +1,1223 @@
+/*-------------------------------------------------------------------------
+ *
+ * citus_dist_stat_activity.c
+ *
+ *	This file contains functions for monitoring the distributed transactions
+ *	accross the cluster.
+ *
+ * Copyright (c) 2018, Citus Data, Inc.
+ *
+ *-------------------------------------------------------------------------
+ */
+
+#include "postgres.h"
+#include "libpq-fe.h"
+#include "miscadmin.h"
+#include "fmgr.h"
+#include "postmaster/postmaster.h"
+
+#include "funcapi.h"
+#include "access/htup_details.h"
+#include "catalog/pg_type.h"
+#include "datatype/timestamp.h"
+#include "distributed/backend_data.h"
+#include "distributed/connection_management.h"
+#include "distributed/listutils.h"
+#include "distributed/lock_graph.h"
+#include "distributed/master_protocol.h"
+#include "distributed/metadata_cache.h"
+#include "distributed/remote_commands.h"
+#include "distributed/transaction_identifier.h"
+#include "executor/spi.h"
+#include "nodes/execnodes.h"
+#include "storage/ipc.h"
+#include "storage/lwlock.h"
+#include "storage/proc.h"
+#include "storage/spin.h"
+#include "storage/s_lock.h"
+#include "utils/builtins.h"
+#if PG_VERSION_NUM >= 100000
+#include "utils/fmgrprotos.h"
+#endif
+#include "utils/inet.h"
+#include "utils/timestamp.h"
+
+
+/*
+ *  citus_dist_stat_activity() and citus_worker_stat_activity() is similar to
+ *  pg_stat_activity. Those functions basically return join of
+ *  pg_stat_activity and get_all_active_transactions() on each node
+ *  in the cluster. The only difference is that citus_dist_stat_activity()
+ *  gets transactions where worker_query = false and citus_worker_stat_activity()
+ *  gets transactions where worker_query = true.
+ *
+ *  In other words, citus_dist_stat_activity returns only the queries that are the
+ *  distributed queries. citus_worker_stat_activity returns only the queries that
+ *  are worker queries (e.g., queries on the shards) initiated by those distributed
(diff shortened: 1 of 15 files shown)


src/backend/distributed/transaction/citus_dist_stat_statements.c
```diff
@@ -0,0 +1,831 @@
+/*-------------------------------------------------------------------------
+ *
+ * citus_dist_stat_statements.c
```
onderkalaci: the file name should have been `citus_dist_stat_activity.c`

src/test/regress/specs/isolation_citus_dist_activity.spec
```diff
@@ -0,0 +1,98 @@
+setup
```
velioglu: We may add some tests using `master_run_from_same_connection` UDF defined in #2363 to check whether we observe the transactions initiated from workers.

src/backend/distributed/citus--8.0-3--8.0-4.sql
```diff
@@ -0,0 +1,45 @@
+/* citus--8.0-3--8.0-4.sql */
+SET search_path = 'pg_catalog';
+
```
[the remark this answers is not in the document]
velioglu: 👍

src/backend/distributed/transaction/citus_dist_stat_activity.c
```diff
@@ -0,0 +1,1223 @@
...
+	List *workerNodeList = ActivePrimaryNodeList();
+	ListCell *workerNodeCell = NULL;
+	char *nodeUser = NULL;
+	List *connectionList = NIL;
+	ListCell *connectionCell = NULL;
+
+	/*
+	 * For the local node, we can avoid opening connections. This might be
+	 * important when we're on the coordinator since it requires configuring
+	 * the authantication for self-connection via any user who calls the citus
```
velioglu: authentication

src/test/regress/specs/isolation_citus_dist_activity.spec
```diff
@@ -0,0 +1,98 @@
...
+{
+	ROLLBACK;
+}
+
+step "s3-view-worker"
+{
+	SELECT query, query_hostname, query_hostport, master_query_host_name, master_query_host_port, state, wait_event_type, wait_event, usename, datname FROM citus_worker_stat_activity ORDER BY query DESC;
+}
+
+step "s3-lock-table"
```
velioglu: Both this step and s2-lock-table haven't been used in the test

marcocitus:
I think it'd be nice to connect as superuser, but it seems like we would have to re-implement permissions, which seems out of scope.

qq: What's the behaviour when a node is down? Do we give a warning or throw an error?

marcocitus:
> We simply warn the user and continue with the other nodes.

Then :ship:!

onderkalaci:
> Then 🚢!

When I get 🚢 from @marcocitus, I merge it within 5 mins 😄

PR #2363, opened 2018-09-04
Adds UDFs for testing MX functionalities with isolation tests

velioglu:
Adds support for testing MX functionalities with isolation tests. 

Since the transactions locking each other are initiated from worker nodes with MX, we need to get these process IDs on worker nodes to check whether any transaction blocking the other one in isolation tests. Since we can not connect to worker nodes, we need to run commands from workers using a UDF on coordinator using a same connection. Beside, in order to check processes on worker nodes within isolation test framework we need to pass process IDs of sessions opened on workers before committing them to another helper process of isolation framework. This helper process detects a lock between sessions.  

We can not connect to worker nodes within isolation tests. So we need to run commands from worker nodes using UDFs on coordinator nodes and commands need to be run from the same connection to mimic the effect of running them from different worker session. Although we have a flag named `SESSION_LIFESPAN` in our connection management API, connections opened with that flag are closed after the command finish if a transaction exists on the worker. So, we have introduced a new flag as a part of the implementation to force connection remain open even if there is a transaction on the worker. We have implemented three different UDFs
  - `start_session_level_connection_to_node(nodename, nodeport)` : Opens a connection to the given node and set the flag `allowNonIdleRemoteTransactionOnXactHandling` to `true`. That allows consecutive commands to use same connection which must be sent by the following UDF.
  - `run_commands_on_session_level_connection_to_node(queryText)`: Runs command using the connection opened with the UDF above.
  - `stop_session_level_connection_to_node()`: Set the flag to false and close the connection.

Since the isolation framework uses an helper process for checking the lock status between different sessions defined in the specs, we need to pass process ID of the session open with `start_session_level_connection_to_node` to that helper process. Since we need to pass that information from an uncommitted transaction, we decided to use GUC variables `isolation_test_session_process_id` and `isolation_test_session_remote_process_id`. First one provides framework to know from which coordinator session `start_session_level_connection_to_node` has been called and `isolation_test_session_remote_process_id` provides the process id of the session on the worker node. Since we need to update that GUC variable on helper process, we need to run `ALTER SYSTEM` command from session defined on the isolation specs. Yet, calling `ALTER SYSTEM` commands is not allowed within a transaction and to not to limit our support to only non-transactional cases we have self-connected to the node and run the `ALTER SYSTEM` command from that self-connected session.

Isolation tests for updating reference tables from workers have also been added.

### DIFF
--- a/src/backend/distributed/citus--8.0-5--8.0-6.sql
+++ b/src/backend/distributed/citus--8.0-5--8.0-6.sql
@@ -1,0 +1,84 @@
+/* citus--8.0-5--8.0-6 */
+SET search_path = 'pg_catalog';
+
+CREATE FUNCTION get_global_active_transactions(OUT datid oid, OUT process_id int, OUT initiator_node_identifier int4, OUT worker_query BOOL, OUT transaction_number int8, OUT transaction_stamp timestamptz)
+  RETURNS SETOF RECORD
+  LANGUAGE C STRICT
+  AS 'MODULE_PATHNAME', $$get_global_active_transactions$$;
+ COMMENT ON FUNCTION get_global_active_transactions(OUT database_id oid, OUT process_id int, OUT initiator_node_identifier int4, OUT transaction_number int8, OUT transaction_stamp timestamptz)
+     IS 'returns distributed transaction ids of active distributed transactions from each node of the cluster';
+
+CREATE OR REPLACE FUNCTION pg_catalog.citus_blocking_pids(pBlockedPid integer)
+RETURNS int4[] AS $$
+  DECLARE
+    mLocalBlockingPids int4[];
+    mRemoteBlockingPids int4[];
+    mLocalTransactionNum int8;
+    workerProcessId integer := current_setting('citus.isolation_test_session_remote_process_id');
+    coordinatorProcessId integer := current_setting('citus.isolation_test_session_process_id');
+  BEGIN
+    SELECT pg_catalog.old_pg_blocking_pids(pBlockedPid) INTO mLocalBlockingPids;
+
+    IF (array_length(mLocalBlockingPids, 1) > 0) THEN
+      RETURN mLocalBlockingPids;
+    END IF;
+
+    -- pg says we're not blocked locally; check whether we're blocked globally.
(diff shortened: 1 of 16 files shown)


src/include/distributed/lock_graph.h
```diff
@@ -14,7 +14,11 @@
 
 
 #include "postgres.h"
+#include "libpq-fe.h"
+
 #include "datatype/timestamp.h"
+#include "distributed/backend_data.h"
+#include "storage/lock.h"
```
onderkalaci: Do we really need `"storage/lock.h"`  and `backend_data.h `?

src/test/regress/specs/isolation_reference_on_mx.spec
```diff
@@ -0,0 +1,125 @@
+# Create and use UDF to send commands from the same connection. Also make the cluster 
+# ready for testing MX functionalities.
+setup
+{	
+	CREATE OR REPLACE FUNCTION start_session_level_connection_to_node(text, integer)
+	    RETURNS void
+	    LANGUAGE C STRICT VOLATILE
+	    AS 'citus', $$start_session_level_connection_to_node$$;
+	COMMENT ON FUNCTION start_session_level_connection_to_node(text, integer)
```
onderkalaci: maybe not add `COMMENT ON` to decrease the line count?

src/backend/distributed/citus--8.0-5--8.0-6.sql
```diff
@@ -0,0 +1,95 @@
...
+    waitingProcessId integer := current_setting('citus.isolation_test_session_process_id');
+  BEGIN
+    SELECT pg_catalog.old_pg_blocking_pids(pBlockedPid) INTO mLocalBlockingPids;
+
+    IF (array_length(mLocalBlockingPids, 1) > 0) THEN
+      RETURN mLocalBlockingPids;
+    END IF;
+
+    -- pg says we're not blocked locally; check whether we're blocked globally.
+    IF (waitingWorkerProcessId != -1) THEN
```
onderkalaci: minor: maybe add a comment that this `IF` is for MX? (If you decide to do so, please update similar places as well)

PR #2366, opened 2018-09-06
Reflect changed index for constraint scans in PG11

mtuncer:
PG11 removed index `pg_constraint_conrelid_index` and created `pg_constraint_conrelid_contypid_conname_index` on pg_constraint. 

With this change we use the new index in PG11, and the old one in pre-11.

### DIFF
--- a/src/include/distributed/version_compat.h
+++ b/src/include/distributed/version_compat.h
@@ -38,6 +38,9 @@
 #define PreventInTransactionBlock PreventTransactionChain
 #define DatumGetJsonbP(d) DatumGetJsonb(d)
 
+/* following is introduced in PG11 */
+#define ConstraintRelidTypidNameIndexId ConstraintRelidIndexId
+
 /* following defines also exist for PG11 */
 #define RELATION_OBJECT_TYPE ACL_OBJECT_RELATION
 #define IndexInfoAttributeNumberArray(indexinfo) (indexinfo->ii_KeyAttrNumbers)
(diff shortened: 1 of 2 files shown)


metdos:
@mtuncer noted that this is a very recent change on the Postgres-11.

src/include/distributed/version_compat.h
```diff
@@ -38,6 +38,9 @@
 #define PreventInTransactionBlock PreventTransactionChain
 #define DatumGetJsonbP(d) DatumGetJsonb(d)
 
+/* following is introduced in PG11 */
+#define ConstraintRelidTypidNameIndexId ConstraintRelidIndexId
```
jasonmp85: OK, so I know this sounds backwards of what we _usually_ want to do, but can you define `ConstraintRelidIndexId` for PostgreSQL 11 rather than defining `ConstraintRelidTypidNameIndexId` for 9.6 and 10?

Yes, we'd remove this `#define` eventually, but the important thing here is that `ConstraintRelidTypidNameIndexId` can be used to look up an index by `relid` (as it's a prefix of the index), but `ConstraintRelidIndexId` is _not_ actually a valid stand-in for `ConstraintRelidTypidNameIndexId` (because it's an index of only a single column, not all three).

Because of this, I think the `#define` you have here is actually misleading (it pretends to backfill some behavior that's actually not implementable, namely this three-column index). Instead, we can make an alias for the name in PostgreSQL 11 to reflect the fact that this new index is still usable in the old manner (looking up by relid).

jasonmp85:
(By the way this is a :shipit: after addressing the feedback)

PR #2367, opened 2018-09-06
Do not recover wrong distributed transactions in MX

### LINKED ISSUE #2095: 2PC recovery might be interfering with active distributed transactions in Citus MX (?)
While working on a large test cluster, we've occasionally observed the following errors on the different worker log files at different times:

The cluster had Citus MX with many worker nodes doing highly concurrent multi-shard commands on 8-9 different distributed tables.

```

ERROR:  prepared transaction with identifier "citus_24_71104_2654_8" does not exist


WARNING:  failed to commit transaction on XXXXX:5432
WARNING:  failed to roll back prepared transaction 'citus_24_65309_836_37'
 HINT:  Run "COMMIT PREPARED 'citus_24_65309_836_37'" on XXXXX:5432



 ERROR:  prepared transaction with identifier "citus_21_63757_589_20" does not exist
[app=citus] STATEMENT:  ROLLBACK PREPARED 'citus_21_63757_589_20'



prepared transaction with identifier "citus_24_21561_24021_24" is busy
STATEMENT:  ROLLBACK PREPARED '

```

Our workaround was just to increase the `citus.recover_2pc_interval` to `30` minutes. After that, all the errors went away. (This obviously cannot be a solution, but mitigate the source of the bug.)

We seem to have a very similar issue in the previous implementation of `recover_prepared_transactions` (e.g., before #1696) and that was fixed with a commit in #1030.

Note: The issue might not be specific to Citus MX, that needs more investigation.

onderkalaci:
DESCRIPTION: Fixes a bug that prevents recovering wrong transactions in MX

Fixes #2095 see the details in that issue.

### DIFF
--- a/src/test/regress/expected/multi_mx_transaction_recovery.out
+++ b/src/test/regress/expected/multi_mx_transaction_recovery.out
@@ -50,15 +50,29 @@ PREPARE TRANSACTION 'citus_12_should_commit';
 BEGIN;
 CREATE TABLE should_be_sorted_into_middle (value int);
 PREPARE TRANSACTION 'citus_12_should_be_sorted_into_middle';
+-- this node (e.g., node id 12) should not touch
+-- transactions with different nodeIds in the gid
+BEGIN;
+CREATE TABLE table_should_do_nothing (value int);
+PREPARE TRANSACTION 'citus_122_should_do_nothing';
 -- Add "fake" pg_dist_transaction records and run recovery
 INSERT INTO pg_dist_transaction VALUES (12, 'citus_12_should_commit');
 INSERT INTO pg_dist_transaction VALUES (12, 'citus_12_should_be_forgotten');
+INSERT INTO pg_dist_transaction VALUES (122, 'citus_122_should_do_nothing');
 SELECT recover_prepared_transactions();
  recover_prepared_transactions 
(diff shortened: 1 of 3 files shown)


src/test/regress/expected/multi_mx_transaction_recovery.out
```diff
@@ -50,15 +50,29 @@ PREPARE TRANSACTION 'citus_12_should_commit';
...
 INSERT INTO pg_dist_transaction VALUES (12, 'citus_12_should_be_forgotten');
+INSERT INTO pg_dist_transaction VALUES (122, 'citus_122_should_do_nothing');
 SELECT recover_prepared_transactions();
  recover_prepared_transactions 
 -------------------------------
                              3
 (1 row)
 
+-- delete the citus_122_should_do_nothing transaction
+DELETE FROM pg_dist_transaction RETURNING *;
```
marcocitus: is it worth adding a filter to make it less dependent on other tests?

onderkalaci: It is probably hard to run this test in parallel with others and also having prepared txs in `pg_dist_transaction` before the test starts. But still, I added the filter.

PR #2368, opened 2018-09-06
Temporarily allow PG11 failures

mtuncer:
This is a temporary commit to unblock PG11 failures at travis
until end of milestone.


metdos:
🚢

PR #2375, opened 2018-09-11
Add regression tests for procedure

mtuncer:
PG11 introduced PROCEDURE concept similar to FUNCTION
Procedure's allow committing/rolling back behavior.

This commmit adds regression tests for procedure calls.


jasonmp85:
A description of some sort would have been handy, since I had to go read documentation to figure out what `PROCEDURE` was, and even then needed several blog posts to determine it permits transaction control statements within certain languages.

PR #2376, opened 2018-09-11
Relax assertion on transaction abort on PREPARE step

### LINKED ISSUE #1966: Assert failure in FinishRemoteTransactionAbort in cancellation test

onderkalaci:
Fixes #1966, and actually very similar to #2031. We should probably
fixed them both at that time.

If you look closely, you'll see that the change doesn't effect 
non-assert builds.

In case a failure happens when a transaction is failed on PREPARE,
we used to hit an assertion for ensuring there is no pending
activity on the connection. However, that's not true after the
changes in #2031. Thus, we've replaced the assertion with a more
generic function call to consume any pending activity, if exists.

### DIFF
--- a/src/backend/distributed/transaction/remote_transaction.c
+++ b/src/backend/distributed/transaction/remote_transaction.c
@@ -528,8 +528,20 @@ FinishRemoteTransactionPrepare(struct MultiConnection *connection)
 		transaction->transactionState = REMOTE_TRANS_PREPARED;
 	}
 
-	result = GetRemoteCommandResult(connection, raiseErrors);
-	Assert(!result);
+	/*
+	 * Try to consume results of PREPARE TRANSACTION command. If we don't
+	 * succeed, rollback the transaction. Note that we've not committed on
+	 * any node yet, and we're not sure about the state of the worker node.
+	 * So rollbacking seems to be the safest action if the worker is
+	 * in a state where it can actually rollback.
+	 */
+	if (!ClearResults(connection, raiseErrors))
+	{
+		ereport(ERROR, (errmsg("failed to prepare transaction '%s' on host %s:%d",
+							   transaction->preparedName, connection->hostname,
+							   connection->port),
+						errhint("Try re-running the command.")));
+	}
 }
 
 


src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -528,8 +528,16 @@ FinishRemoteTransactionPrepare(struct MultiConnection *connection)
 		transaction->transactionState = REMOTE_TRANS_PREPARED;
 	}
 
-	result = GetRemoteCommandResult(connection, raiseErrors);
-	Assert(!result);
+	/*
+	 * Try to consume results of any in-progress commands. In the 1PC case
+	 * this is also where we consume the result of the ROLLBACK.
```
marcocitus: Comment is stale.

onderkalaci: It is actually copy & paste 😄

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -528,8 +528,16 @@ FinishRemoteTransactionPrepare(struct MultiConnection *connection)
...
+	 *
+	 * If we don't succeed the connection will be in a bad state, so we close it.
+	 */
+	if (!ClearResults(connection, raiseErrors))
+	{
+		ShutdownConnection(connection);
```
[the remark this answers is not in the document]
onderkalaci: Hmm, I couldn't find a good answer to this. But, even if we leave any prepared transactions, wouldn't 2PC recovery handle that for us?

marcocitus:
> Hmm, I couldn't find a good answer to this. But, even if we leave any prepared transactions, wouldn't 2PC recovery handle that for us?

Actually, what about throwing an error here? We haven't committed yet and something about the PREPARE TRANSACTION failed, so we're not entirely sure what state the worker is in. Safest course of action seems to be rolling back. The connection might even still be in a state where we can roll it back. 

I realise this would already happen in all cases except replication factor > 1 and we could potentially mark shard placement(s) inactive, but that seems like kind of a poor choice.


Unrelated note: One kind of bad thing about the way Citus handles distributed transaction is that our 2PC logic returns a COMMIT (with warnings) even if we fail to commit a prepared transaction. That breaks read-your-writes consistency. The most correct thing to do would be to wait until recovery completes (potentially forever). A more reasonable, yet correct thing to do would be to close the session entirely, leaving it in an undeterminate state.

onderkalaci:
> Safest course of action seems to be rolling back. 

I agree with that. I updated the PR, though I cannot find a test to add, which seems hard. Do you have any suggestions/ideas for that?

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -528,8 +528,19 @@ FinishRemoteTransactionPrepare(struct MultiConnection *connection)
...
+	 * Try to consume results of PREPARE TRANSACTION command. If we don't
+	 * succeed, rollback the transaction. Note that we've not committed on
+	 * any node yet, and we're not sure about the state of the worker node.
+	 * So rollbacking seems to be the safest action if the worker is
+	 * in a state where it can actually rollback.
+	 */
+	if (!ClearResults(connection, raiseErrors))
+	{
+		ereport(ERROR, (errmsg("failed to prepare transaction '%s'",
+							   transaction->preparedName),
```
marcocitus: could use a host name/port in the message.

PR #2379, opened 2018-09-12
Fixes a bug preventing rollback in stored procedure

### LINKED ISSUE #2371: can not rollback a transaction within procedure
You 
```sql
CREATE TABLE test_table(id integer , org_id integer);
SELECT create_distributed_table('test_table', 'id');
CREATE PROCEDURE test_procedure_rollback(tt_id int, tt_org_id int) LANGUAGE PLPGSQL AS $$
BEGIN
	DELETE FROM test_table;
	COMMIT;
	INSERT INTO test_table VALUES (tt_id, tt_org_id);
	ROLLBACK;
END;
$$;

INSERT INTO test_table VALUES (2,3);

CALL test_procedure_rollback(2,5);
SELECT * FROM test_table ORDER BY 1, 2;
```

Expect:
nothing since delete is commited and insert is rolled back.

Result:
```
 id | org_id 
----+--------
  2 |      5
(1 row)
```

When you run this test on regular (not distributed) tables. insert is rolled back with no problem

Note : It works as expected on reference tables.

marcocitus:
DESCRIPTION: Fixes a bug preventing rollback in CREATE PROCEDURE

Fixes #2371 by setting a global `IsStoredProcedure` for every `CallStmt` in the utility hook. To make sure we reset the flag when an error occurs, we call `standard_ProcessUtility` in a `PG_TRY` block.

### DIFF
--- a/src/backend/distributed/executor/multi_utility.c
+++ b/src/backend/distributed/executor/multi_utility.c
@@ -296,6 +296,37 @@ multi_ProcessUtility(PlannedStmt *pstmt,
 		return;
 	}
 
+#if (PG_VERSION_NUM >= 110000)
+	if (IsA(parsetree, CallStmt))
+	{
+		/*
+		 * Stored procedures are a bit strange in the sense that some statements
+		 * are not in a transaction block, but can be rolled back. We need to
+		 * make sure we send all statements in a transaction block. The
+		 * StoredProcedureLevel variable signals this to the router executor
+		 * and indicates how deep in the call stack we are in case of nested
+		 * stored procedures.
+		 */
+		StoredProcedureLevel += 1;
+
+		PG_TRY();
+		{
+			standard_ProcessUtility(pstmt, queryString, context,
+									params, queryEnv, dest, completionTag);
+
+			StoredProcedureLevel -= 1;
+		}
+		PG_CATCH();
+		{
+			StoredProcedureLevel -= 1;
+			PG_RE_THROW();
+		}
+		PG_END_TRY();
+
+		return;
+	}
+#endif
+
 	/*
 	 * TRANSMIT used to be separate command, but to avoid patching the grammar
 	 * it's no overlaid onto COPY, but with FORMAT = 'transmit' instead of the
--- a/src/backend/distributed/executor/multi_router_executor.c
+++ b/src/backend/distributed/executor/multi_router_executor.c
@@ -80,6 +80,9 @@ bool AllModificationsCommutative = false;
 /* we've deprecated this flag, keeping here for some time not to break existing users */
 bool EnableDeadlockPrevention = true;
 
+/* number of nested stored procedure call levels we are currently in */
+int StoredProcedureLevel = 0;
+
 /* functions needed during run phase */
 static void AcquireMetadataLocks(List *taskList);
 static ShardPlacementAccess * CreatePlacementAccess(ShardPlacement *placement,
@@ -613,7 +616,8 @@ RouterSequentialModifyExecScan(CustomScanState *node)
 	 * customers already use functions that touch multiple shards from within
 	 * a function, so we'll ignore functions for now.
 	 */
-	if (IsTransactionBlock() || multipleTasks || taskListRequires2PC)
+	if (IsTransactionBlock() || multipleTasks || taskListRequires2PC ||
+		StoredProcedureLevel > 0)
 	{
 		BeginOrContinueCoordinatedTransaction();
 
--- a/src/include/distributed/multi_router_executor.h
+++ b/src/include/distributed/multi_router_executor.h
@@ -36,6 +36,10 @@ typedef struct XactShardConnSet
 extern bool AllModificationsCommutative;
 extern bool EnableDeadlockPrevention;
 
+/* number of nested stored procedure call levels we are currently in */
+extern int StoredProcedureLevel;
+
+
 extern void CitusModifyBeginScan(CustomScanState *node, EState *estate, int eflags);
 extern TupleTableSlot * RouterSelectExecScan(CustomScanState *node);
 extern TupleTableSlot * RouterModifyExecScan(CustomScanState *node);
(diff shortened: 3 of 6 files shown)


jasonmp85:
Does `PG_TRY` catch all levels of errors? Is it possible to get something higher than `ERROR` (`FATAL`?) that doesn't kill the backend entirely? I'm a bit skeptical of using try/catch to maintain this state (versus some other form of callback)…

marcocitus:
There is only one level of error which breaks control flow without terminating the session, which is `ERROR`. We use the `PG_TRY` pattern in various places.

from the header:
```
/*----------
 * API for catching ereport(ERROR) exits.  Use these macros like so:
 *
 *      PG_TRY();
 *      {
 *          ... code that might throw ereport(ERROR) ...
 *      }
 *      PG_CATCH();
 *      {
 *          ... error recovery code ...
 *      }
 *      PG_END_TRY();
 *
 * (The braces are not actually necessary, but are recommended so that
 * pgindent will indent the construct nicely.)  The error recovery code
 * can optionally do PG_RE_THROW() to propagate the same error outwards.
 *
 * Note: while the system will correctly propagate any new ereport(ERROR)
 * occurring in the recovery section, there is a small limit on the number
 * of levels this will work for.  It's best to keep the error recovery
 * section simple enough that it can't generate any new errors, at least
 * not before popping the error stack.
 *
 * Note: an ereport(FATAL) will not be caught by this construct; control will
 * exit straight through proc_exit().  Therefore, do NOT put any cleanup
 * of non-process-local resources into the error recovery section, at least
 * not without taking thought for what will happen during ereport(FATAL).
 * The PG_ENSURE_ERROR_CLEANUP macros provided by storage/ipc.h may be
 * helpful in such cases.
 *
 * Note: if a local variable of the function containing PG_TRY is modified
 * in the PG_TRY section and used in the PG_CATCH section, that variable
 * must be declared "volatile" for POSIX compliance.  This is not mere
 * pedantry; we have seen bugs from compilers improperly optimizing code
 * away when such a variable was not marked.  Beware that gcc's -Wclobbered
 * warnings are just about entirely useless for catching such oversights.
 *----------
 */
```

jasonmp85:
> There is only one level of error which breaks control flow without terminating the session, which is `ERROR`. We use the `PG_TRY` pattern in various places.

Ack.

mtuncer:
Verified that it fixes the problem.

Tested with nested procedures 
```sql
drop table if exists test_table cascade;
create table test_table (id int, org_id int);

SELECT create_distributed_table('test_table', 'id');
CREATE OR REPLACE PROCEDURE test_procedure_rollback(tt_id int, tt_org_id int) LANGUAGE PLPGSQL AS $$
BEGIN
    INSERT INTO test_table VALUES (tt_id+12, tt_org_id+12);
    ROLLBACK;
END;
$$;

CREATE OR REPLACE PROCEDURE test_procedure_rollback_2(tt_id int, tt_org_id int) LANGUAGE PLPGSQL AS $$
BEGIN
    INSERT INTO test_table VALUES (tt_id+2, tt_org_id+1);
    ROLLBACK;
END;
$$;

CREATE OR REPLACE PROCEDURE test_procedure(tt_id int, tt_org_id int) LANGUAGE PLPGSQL AS $$
BEGIN
    call test_procedure_rollback(tt_id, tt_org_id);
    call test_procedure_rollback_2(tt_id, tt_org_id);
END;
$$;


SELECT * from test_table;
call test_procedure(1,1);
call test_procedure(20, 20);
SELECT * from test_table;
```

Since both are rolled back, expected nothing at the tables. Found nothing ✅ 

Also verified that tables are left with some rows before this fix

mtuncer:
following test fails to rollback
```sql
drop table if exists test_table cascade;
create table test_table (id int, org_id int);

SELECT create_distributed_table('test_table', 'id');
CREATE OR REPLACE PROCEDURE test_procedure_rollback(tt_id int, tt_org_id int) LANGUAGE PLPGSQL AS $$
BEGIN
    INSERT INTO test_table VALUES (tt_id+12, tt_org_id+12);
    ROLLBACK;
END;
$$;

CREATE OR REPLACE PROCEDURE test_procedure_rollback_2(tt_id int, tt_org_id int) LANGUAGE PLPGSQL AS $$
BEGIN
    INSERT INTO test_table VALUES (tt_id+2, tt_org_id+1);
    ROLLBACK;
END;
$$;

CREATE OR REPLACE PROCEDURE test_procedure(tt_id int, tt_org_id int) LANGUAGE PLPGSQL AS $$
BEGIN
    call test_procedure_rollback(tt_id, tt_org_id);
    call test_procedure_rollback_2(tt_id, tt_org_id);
    INSERT INTO test_table VALUES (tt_id+100, tt_org_id+100);
    ROLLBACK;
END;
$$;

SELECT * from test_table;
call test_procedure(1,1);
call test_procedure(20, 20);
SELECT * from test_table;

```

Problem is that as soon as we exit from call statement execution we reset the flag. We should instead keep count of call statements that we have encountered and decrement int.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -296,6 +296,35 @@ multi_ProcessUtility(PlannedStmt *pstmt,
...
+		 * IsStoredProcedure flag signals this to the router executor.
+		 */
+		IsStoredProcedure = true;
+
+		PG_TRY();
+		{
+			standard_ProcessUtility(pstmt, queryString, context,
+									params, queryEnv, dest, completionTag);
+
+			IsStoredProcedure = false;
```
mtuncer: we should consider a counter instead of a boolean flag here

mtuncer: Otherwise our execution is incorrect when nested procedure calls are used

marcocitus: Wow, I didn't realise procedure calls can be nested. That must get really confusing with commit/rollback. Will change to a counter.

marcocitus:
Changed to a counter and added the new tests. Also, nested procedure calls are insane.

jasonmp85:
> Also, nested procedure calls are insane

I mean short of banning procedures calling one another (which seems overly restrictive by not making them composable whatsoever), isn't this sort of what would be expected?

marcocitus:
> I mean short of banning procedures calling one another (which seems overly restrictive by not making them composable whatsoever), isn't this sort of what would be expected?

True, but the semantics get really weird. A nested stored procedure can roll back or commit changes made in the caller.

PR #2381, opened 2018-09-13
Add citus_lock_waits to show locked distributed queries

velioglu:
DESCRIPTION: Adds `citus_lock_waits` to show which distributed queries are waiting for each other

Adds view named `citus_lock_waits` to show distributed queries waiting for each other within the cluster. View consists of column `blocked_pid`, `blocking_pid`, `blocked_statement` and `current_statement_in_blocking_process`. That view should be queried from the coordinator node since workers are not aware of coordinator node's hostname and port so they can not fetch distributed queries from the coordinator node. 

Note that, some limitation exist while showing distributed queries as a part of this view:
- This only shows queries that are inside distributed transactions. 
    - Single shard modifications that are not inside transaction blocks wouldn't show up in the view. 
    - Real-time selects that are not inside distributed transactions wouldn't show up in the view.
- Task-tracker queries won't show up in the view at all.
- Blockings that are due to table locks wouldn't show up in the view. 
    - For example, if you do ALTER TABLE from the coordinator and then UPDATE/DELETE/SELECT/INSERT from a worker, that wouldn't show up in the view. 
    - Similarly, two DDLs from the coordinator wouldn't show up in the view.

### DIFF
--- a/src/backend/distributed/citus--8.0-6--8.0-7.sql
+++ b/src/backend/distributed/citus--8.0-6--8.0-7.sql
@@ -1,0 +1,46 @@
+/* citus--8.0-6--8.0-7 */
+SET search_path = 'pg_catalog';
+
+CREATE VIEW citus.citus_lock_waits AS
+
+WITH 
+citus_dist_stat_activity AS
+(
+  SELECT * FROM citus_dist_stat_activity
+),
+unique_global_wait_edges AS
+(
+	SELECT DISTINCT ON(waiting_node_id, waiting_transaction_num, blocking_node_id, blocking_transaction_num) * FROM dump_global_wait_edges()
+),
+citus_dist_stat_activity_with_node_id AS
+(
+  SELECT
+  citus_dist_stat_activity.*, (CASE citus_dist_stat_activity.master_query_host_name WHEN 'coordinator_host' THEN 0 ELSE pg_dist_node.nodeid END) as initiator_node_id
+  FROM
+  citus_dist_stat_activity LEFT JOIN pg_dist_node
+  ON
+  citus_dist_stat_activity.master_query_host_name = pg_dist_node.nodename AND 
+  citus_dist_stat_activity.master_query_host_port = pg_dist_node.nodeport
+)
+SELECT
+ waiting.pid AS waiting_pid,
+ blocking.pid AS blocking_pid,
+ waiting.query AS blocked_statement,
+ blocking.query AS current_statement_in_blocking_process,
+ waiting.initiator_node_id AS waiting_node_id,
+ blocking.initiator_node_id AS blocking_node_id,
+ waiting.master_query_host_name AS waiting_node_name,
+ blocking.master_query_host_name AS blocking_node_name,
+ waiting.master_query_host_port AS waiting_node_port,
+ blocking.master_query_host_port AS blocking_node_port
+FROM
+ unique_global_wait_edges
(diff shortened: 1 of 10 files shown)


src/backend/distributed/test/run_from_same_connection.c
```diff
@@ -142,12 +142,12 @@ run_commands_on_session_level_connection_to_node(PG_FUNCTION_ARGS)
 			 "start_session_level_connection_to_node must be called first to open a session level connection");
 	}
 
-	ExecuteCriticalRemoteCommand(connection, queryString);
-
 	appendStringInfo(processStringInfo, ALTER_CURRENT_PROCESS_ID, MyProcPid);
 	appendStringInfo(workerProcessStringInfo, ALTER_CURRENT_WORKER_PROCESS_ID,
 					 GetRemoteProcessId(connection));
 
+	ExecuteCriticalRemoteCommand(connection, queryString);
```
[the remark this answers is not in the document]
velioglu: Otherwise we will get the `ALTER_CURRENT_WORKER_PROCESS_ID` as a last query in isolation tests.

src/backend/distributed/citus--8.0-6--8.0-7.sql
```diff
@@ -0,0 +1,38 @@
...
+(
+  SELECT 
+  citus_dist_stat_activity.*, (CASE citus_dist_stat_activity.master_query_host_name WHEN 'coordinator_host' THEN 0 ELSE pg_dist_node.nodeid END) as initiator_node_id
+  FROM
+  citus_dist_stat_activity LEFT JOIN pg_dist_node
+  ON
+  citus_dist_stat_activity.master_query_host_name = pg_dist_node.nodename AND 
+  citus_dist_stat_activity.master_query_host_port = pg_dist_node.nodeport
+)
+SELECT
```
onderkalaci: This query (and the query in the blog-post) generates duplicates when the commands are multi-shard. The main reason for that is `dump_local_wait_edges()` do not distinguish between master/worker queries and could have multiple entries for the same pair of distributed transactions.

We should probably have some kind of `DISTINCT ON` for `waiting.initiator_node_id, waiting.transaction_number,  blocking.initiator_node_idblocking.transaction_number`.

But, obviously, that's a vague idea that came to my mind right now. So, please ensure that `DISTINCT ON` works for us or not.

One another approach could be to add `worker_query` field to `dump_*_wait_edges()`, which requires some more extensive changes. So, let's first try to re-write the query such that we get rid of duplicates.

src/test/regress/specs/isolation_get_distributed_wait_queries.spec
```diff
@@ -0,0 +1,183 @@
...
+}
+
+session "s3"
+
+step "s3-select-distributed-waiting-queries"
+{
+	SELECT blocked_statement, current_statement_in_blocking_process FROM citus_lock_waits;
+}
+
+permutation "s1-start-session-level-connection" "s1-begin-on-worker" "s1-update-ref-table" "s2-start-session-level-connection" "s2-begin-on-worker" "s2-update-ref-table" "s3-select-distributed-waiting-queries" "s1-commit-worker" "s2-commit-worker" "s1-stop-connection" "s2-stop-connection"
```
onderkalaci: One test request: Could you add a test where one update is executed from the coordinator and one is from the worker?

Issue #2369, opened 2018-09-07
Citus 7.4 can not drop index on parent of partitioned table


zjw5121818:
Hello,all
There is a problem when I try to drop index on parent partitoned table . I've use Citus7.4 + PG11beta2 in my test environment.I'm pretty sure this error comes from Citus because another test environment doesn't cause this problem in the same steps.

```My procedures are as follows:ostgres@centos101 ~]$ psql -d cloud
psql (11beta2)
Type "help" for help.
cloud=# c\dx
                        List of installed extensions
   Name    | Version |   Schema   |               Description                
-----------+---------+------------+------------------------------------------
 citus     | 7.4-3   | pg_catalog | Citus distributed database
 pg_cron   | 1.1     | public     | Job scheduler for PostgreSQL
 plpgsql   | 1.0     | pg_catalog | PL/pgSQL procedural language
 plpythonu | 1.0     | pg_catalog | PL/PythonU untrusted procedural language
(4 rows)

cloud=# create table aaa(id int,name text,cap_date timestamp) partition by range(cap_date);
CREATE TABLE
cloud=# create index aaa_idx_id on aaa(id) ;
CREATE INDEX
cloud=# drop index aaa_idx_id ;
ERROR:  "aaa_idx_id" is not an index
```
I hope contributors can fix the problem soon,Thx.

mtuncer:
verified that with PG11 Beta3. It only occurs when citus extension is present. drop index succeeds when the extension is dropped.

zjw5121818:
@mtuncer We've change this file `src/backend/distributed/executor/multi_real_time_executor.c` and it seems work well.
```
if (classform->relkind != RELKIND_PARTITIONED_INDEX && classform->relkind != relkind)
		ereport(ERROR, (errcode(ERRCODE_WRONG_OBJECT_TYPE),
						errmsg("\"%s\" is not an index", rel->relname)));
```

PR #2383, opened 2018-09-14
Fix drop index bug on PG11 partitioned table

mtuncer:
Fixes a bug on not being able to drop index on a partitioned table.

Reason for the failure is that PG11 introduced a new relation kind `RELKIND_PARTITIONED_INDEX` to be used for partitioned indices. We expanded our check to cover that case.

Fixes : #2369


src/test/regress/expected/multi_partitioning_1.out
```diff
@@ -541,6 +541,18 @@ SELECT tablename, indexname FROM pg_indexes WHERE tablename LIKE 'partitioning_t
...
+-- ie using the same index name
+-- following will fail
+DROP INDEX partitioning_test_2009_id_idx;
+ERROR:  index "partitioning_test_2009_id_idx" does not exist
+-- but dropping index on parent table will succeed
+DROP INDEX partitioning_index;
+ERROR:  index "partitioning_index" does not exist
+-- this index was already created on partition table
+DROP INDEX partitioning_2009_index;
+ERROR:  index "partitioning_2009_index" does not exist
```
marcocitus: fix looks fine, but we haven't covered the original problem in the tests: dropping an index on a non-distributed partitioned table throws an error before this fix.

mtuncer:
@marcocitus ping

PR #2384, opened 2018-09-14
Prevent overflow of memory accesses during deadlock detection

### LINKED ISSUE #2032: Stuck spinlock during MX pgbench
Ran into a PANIC during a pgbench run with tpcb-like workload on an 8-node Citus MX formation:
```
2018-02-28 11:58:28.862 UTC [32124][150/75022] : [3-1] [app=citus] PANIC:  stuck spinlock detected at GetBackendDataForProc, transaction/backend_data.c:587
2018-02-28 11:58:28.862 UTC [32124][150/75022] : [4-1] [app=citus] STATEMENT:  SELECT * FROM dump_local_wait_edges()
2018-02-28 11:58:28.945 UTC [16581][] : [15-1] LOG:  server process (PID 32124) was terminated by signal 6: Aborted
2018-02-28 11:58:28.945 UTC [16581][] : [16-1] DETAIL:  Failed process was running: SELECT * FROM dump_local_wait_edges()
2018-02-28 11:58:28.945 UTC [16581][] : [17-1] LOG:  terminating any other active server processes
```

onderkalaci:
Fixes #2032  and https://github.com/citusdata/citus-enterprise/issues/276

In the distributed deadlock detection design, we concluded that prepared transactions
cannot be part of a distributed deadlock. The idea is that (a) when the transaction
is prepared it already acquires all the locks, so cannot be part of a deadlock
(b) even if some other processes blocked on the prepared transaction,  prepared transactions
 would eventually be committed (or rollbacked) and the system will continue operating.

With the above in mind, we probably had a mistake in terms of memory allocations. For each
backend initialized, we keep a `BackendData` struct. The bug we've introduced is that, we
assumed there would only be `MaxBackend` number of backends. However, `MaxBackends` doesn't
include the prepared transactions and axuliary processes. When you check Postgres' InitProcGlobal`
you'd see that `TotalProcs = MaxBackends + NUM_AUXILIARY_PROCS + max_prepared_xacts;`

This commit aligns with total procs processed with that.

Note that this commit along with #2376 allows me to run the `tpcb-like` benchmark for hours without any issues.

Final note: We could probably tweak the deadlock detection code to completely ignore the prepared statements and auxiliary procs. However, that seems to require some deeper look and a lot more testing. Instead, with approach ends-up with `104` extra backend space which is really too small to bother about.

### DIFF
--- a/src/backend/distributed/transaction/backend_data.c
+++ b/src/backend/distributed/transaction/backend_data.c
@@ -541,9 +541,11 @@ BackendManagementShmemInit(void)
 
 		/*
 		 * We need to init per backend's spinlock before any backend
-		 * starts its execution.
+		 * starts its execution. Note that we initialize TotalProcs (e.g., not
+		 * MaxBackends) since some of the blocking processes could be prepared
+		 * transactions, which aren't covered by MaxBackends.
 		 */
-		for (backendIndex = 0; backendIndex < MaxBackends; ++backendIndex)
+		for (backendIndex = 0; backendIndex < TotalProcs; ++backendIndex)
 		{
 			SpinLockInit(&backendManagementShmemData->backends[backendIndex].mutex);
 		}
@@ -568,7 +570,7 @@ BackendManagementShmemSize(void)
 	Size size = 0;
 
 	size = add_size(size, sizeof(BackendManagementShmemData));
-	size = add_size(size, mul_size(sizeof(BackendData), MaxBackends));
+	size = add_size(size, mul_size(sizeof(BackendData), TotalProcs));
 
 	return size;
 }
--- a/src/include/distributed/backend_data.h
+++ b/src/include/distributed/backend_data.h
@@ -13,12 +13,16 @@
 #define BACKEND_DATA_H
 
 
+#include "access/twophase.h"
 #include "datatype/timestamp.h"
 #include "distributed/transaction_identifier.h"
 #include "nodes/pg_list.h"
 #include "storage/lwlock.h"
 #include "storage/proc.h"
 #include "storage/s_lock.h"
+
+
+#define TotalProcs (MaxBackends + NUM_AUXILIARY_PROCS + max_prepared_xacts)
 
 
 /*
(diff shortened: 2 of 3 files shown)


src/include/distributed/backend_data.h
```diff
@@ -21,6 +22,9 @@
 #include "storage/s_lock.h"
 
 
+#define TotalProcs MaxBackends + NUM_AUXILIARY_PROCS + max_prepared_xacts
```
marcocitus: Maybe put the expression into parentheses. That avoids bugs like `TotalProcs * 2` getting translated to `MaxBackends + NUM_AUXILIARY_PROCS + max_prepared_xacts * 2`

src/backend/distributed/transaction/backend_data.c
```diff
@@ -845,7 +845,7 @@ ActiveDistributedTransactionNumbers(void)
 	int curBackend = 0;
 
 	/* build list of starting procs */
-	for (curBackend = 0; curBackend < MaxBackends; curBackend++)
+	for (curBackend = 0; curBackend < TotalProcs; curBackend++)
```
marcocitus: I'm not entirely against it, but I don't really see a point of continuing beyond `MaxBackends`

src/backend/distributed/transaction/distributed_deadlock_detection.c
```diff
@@ -381,7 +381,7 @@ AssociateDistributedTransactionWithBackendProc(TransactionNode *transactionNode)
 {
 	int backendIndex = 0;
 
-	for (backendIndex = 0; backendIndex < MaxBackends; ++backendIndex)
+	for (backendIndex = 0; backendIndex < TotalProcs; ++backendIndex)
```
marcocitus: same here

src/backend/distributed/transaction/backend_data.c
onderkalaci: True, I also noted that in the issue description. I'll update on some places where we're definitely looking for distributed transactions such as here.

Also added a comment for describing the distinction between usage of `MaxBackends` vs `TotalProcs`

src/include/distributed/backend_data.h
onderkalaci: Oops, that's a nasty bug and already the case on `BuildLocalWaitGraph ()` (though not important at that point).

PR #2386, opened 2018-09-17
Use tree walker instad of mutator in relation visibility

onderkalaci:
This commit changes `*_mutator` to `*_walker` for performance reasons
in relation visibility changing walker.
Given that we're only updating a `functionId` in the tree, the approach
seems fine.

### DIFF
--- a/src/backend/distributed/worker/worker_shard_visibility.c
+++ b/src/backend/distributed/worker/worker_shard_visibility.c
@@ -24,7 +24,7 @@
 bool OverrideTableVisibility = true;
 
 static bool RelationIsAKnownShard(Oid shardRelationId);
-static Node * ReplaceTableVisibleFunctionWalker(Node *inputNode);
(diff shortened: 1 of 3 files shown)


src/backend/distributed/worker/worker_shard_visibility.c
```diff
@@ -235,14 +241,14 @@ ReplaceTableVisibleFunctionWalker(Node *inputNode)
 			 */
 			functionToProcess->funcid = CitusTableVisibleFuncId();
 
-			return (Node *) functionToProcess;
+			return false;
```
marcocitus: Minor: Technically speaking, you could have nested `pg_table_is_visible` function calls. Should we just drop into the expression_tree_walker here?

onderkalaci: Probably not a very likely case but added the call and a test as well.

PR #2389, opened 2018-09-19
Support partitioned tables with replication factor > 1

onderkalaci:
DESCRIPTION: Support partitioned tables with replication factor > 1

Given that we're implementing it with some noticeable restrictions, I prefer not to close the related issue (#2233).

In this iteration, we disallow any kind of modifications on partition tables (including `CREATE INDEX` etc.). Instead, we ask users to run any kind of modification queries on the parent table. With this approach, we're trying to limit the code changes. In summary, this shortcut implementation allows us not to think about shard resource locks to preserve consistency of placements and repairing `INVALID`  shard placements, which involves some edge cases.

TODO:

- [x] Get these commits into enterprise repo, make sure that rebalancer works fine. If not, fix the isues on the enterprise repo. I'm planning to consider `master_move_shard_placement()` only.
    - (We already do not support tenant isolation for partitioned tables. Except for [275](https://github.com/citusdata/citus-enterprise/issues/275))
- [x] Fix the bug on `TRUNCATE`. Given that `TRUNCATE parent;` cascades into partitions, `TRUNCATE parent` also fails. I'll fix it when we introduce utility hook for `TRUNCATE` in this release.
- [x] We do allow `DROP TABLE partition`, which seems fine to me given that we could only mark shard placements of non-existing shards in the worst case. What the reviewer think about it?

### DIFF
--- a/src/backend/distributed/commands/multi_copy.c
+++ b/src/backend/distributed/commands/multi_copy.c
@@ -64,6 +64,7 @@
 #include "distributed/multi_copy.h"
 #include "distributed/multi_partitioning_utils.h"
 #include "distributed/multi_physical_planner.h"
+#include "distributed/multi_router_planner.h"
 #include "distributed/multi_shard_transaction.h"
 #include "distributed/placement_connection.h"
 #include "distributed/relation_access_tracking.h"
@@ -198,6 +199,9 @@ CitusCopyFrom(CopyStmt *copyStatement, char *completionTag)
 	{
 		Oid relationId = RangeVarGetRelid(copyStatement->relation, NoLock, false);
 		char partitionMethod = PartitionMethod(relationId);
+
+		/* disallow modifications to a partition table which have rep. factpr > 1 */
+		EnsurePartitionTableNotReplicated(relationId);
 
 		if (partitionMethod == DISTRIBUTE_BY_HASH || partitionMethod ==
 			DISTRIBUTE_BY_RANGE || partitionMethod == DISTRIBUTE_BY_NONE)
--- a/src/backend/distributed/executor/multi_utility.c
+++ b/src/backend/distributed/executor/multi_utility.c
@@ -152,6 +152,7 @@ static void ErrorIfUnsupportedSeqStmt(CreateSeqStmt *createSeqStmt);
 static void ErrorIfDistributedAlterSeqOwnedBy(AlterSeqStmt *alterSeqStmt);
 static void ErrorIfUnsupportedTruncateStmt(TruncateStmt *truncateStatement);
 static void ProcessTruncateStatement(TruncateStmt *truncateStatement);
+static void EnsurePartitionTableNotReplicatedForTruncate(TruncateStmt *truncateStatement);
 static void LockTruncatedRelationMetadataInWorkers(TruncateStmt *truncateStatement);
 static void AcquireDistributedLockOnRelations(List *relationIdList, LOCKMODE lockMode);
 static bool OptionsSpecifyOwnedBy(List *optionList, Oid *ownedByTableId);
@@ -2931,7 +2932,51 @@ ErrorIfUnsupportedTruncateStmt(TruncateStmt *truncateStatement)
 
 
 /*
- * ProcessTruncateStatement handles distributed locking
+ * ProcessTruncateStatement handles few things that should be
+ * done before standard process utility is called for truncate
+ * command.
+ */
+static void
+ProcessTruncateStatement(TruncateStmt *truncateStatement)
+{
+	EnsurePartitionTableNotReplicatedForTruncate(truncateStatement);
+	LockTruncatedRelationMetadataInWorkers(truncateStatement);
+}
+
+
+/*
+ * EnsurePartitionTableNotReplicatedForTruncate a simple wrapper around
+ * EnsurePartitionTableNotReplicated for TRUNCATE command.
+ */
+static void
+EnsurePartitionTableNotReplicatedForTruncate(TruncateStmt *truncateStatement)
+{
+	ListCell *relationCell = NULL;
+
+	foreach(relationCell, truncateStatement->relations)
+	{
+		RangeVar *relationRV = (RangeVar *) lfirst(relationCell);
+		Relation relation = heap_openrv(relationRV, NoLock);
+		Oid relationId = RelationGetRelid(relation);
+
+		if (!IsDistributedTable(relationId))
+		{
+			heap_close(relation, NoLock);
+			continue;
+		}
+
+		EnsurePartitionTableNotReplicated(relationId);
+
+		heap_close(relation, NoLock);
+	}
+}
+
+
+/*
+ * LockTruncatedRelationMetadataInWorkers determines if distributed
+ * lock is necessary for truncated relations, and acquire locks.
+ *
+ * LockTruncatedRelationMetadataInWorkers handles distributed locking
  * of truncated tables before standard utility takes over.
  *
  * Actual distributed truncation occurs inside truncate trigger.
@@ -2939,17 +2984,6 @@ ErrorIfUnsupportedTruncateStmt(TruncateStmt *truncateStatement)
  * This is only for distributed serialization of truncate commands.
  * The function assumes that there is no foreign key relation between
  * non-distributed and distributed relations.
- */
-static void
-ProcessTruncateStatement(TruncateStmt *truncateStatement)
-{
-	LockTruncatedRelationMetadataInWorkers(truncateStatement);
-}
-
-
-/*
- * LockTruncatedRelationMetadataInWorkers determines if distributed
- * lock is necessary for truncated relations, and acquire locks.
  */
 static void
 LockTruncatedRelationMetadataInWorkers(TruncateStmt *truncateStatement)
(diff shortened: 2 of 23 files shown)


src/backend/distributed/commands/multi_copy.c
```diff
@@ -199,6 +200,9 @@ CitusCopyFrom(CopyStmt *copyStatement, char *completionTag)
 		Oid relationId = RangeVarGetRelid(copyStatement->relation, NoLock, false);
 		char partitionMethod = PartitionMethod(relationId);
 
+		/* disallow modifications to a partition table which have rep. factpr > 1 */
+		EnsurePartitionTableNotReplicated(relationId);
```
mtuncer: This is a major restriction. It will render whole feature useless if we disallow copy.

src/backend/distributed/utils/shardinterval_utils.c
```diff
@@ -414,19 +414,21 @@ SearchCachedShardInterval(Datum partitionColumnValue, ShardInterval **shardInter
...
 SingleReplicatedTable(Oid relationId)
 {
-	List *shardIntervalList = LoadShardList(relationId);
-	ListCell *shardIntervalCell = NULL;
+	List *shardList = LoadShardList(relationId);
+	List *shardPlacementList = NIL;
+	Oid shardId = INVALID_SHARD_ID;
 
-	foreach(shardIntervalCell, shardIntervalList)
+	if (list_length(shardList) <= 1)
```
mtuncer: What is this check for ? reference table ?

src/backend/distributed/utils/shardinterval_utils.c
```diff
@@ -414,19 +414,21 @@ SearchCachedShardInterval(Datum partitionColumnValue, ShardInterval **shardInter
...
-		uint64 shardId = (*shardIdPointer);
-		List *shardPlacementList = ShardPlacementList(shardId);
+		return false;
+	}
 
-		if (list_length(shardPlacementList) != 1)
-		{
-			return false;
-		}
+	/* checking only for the first shard id should suffice */
```
mtuncer: Does this assume a healthy cluster.  What happens if the first shard has a single placement but the rest has multiple ?

src/backend/distributed/commands/multi_copy.c
marcocitus: Can't you still copy into the parent?

mtuncer: Correct,  I take it back.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -3143,6 +3143,7 @@ ExecuteDistributedDDLJob(DDLJob *ddlJob)
 	bool shouldSyncMetadata = ShouldSyncTableMetadata(ddlJob->targetRelationId);
 
 	EnsureCoordinator();
+	EnsurePartitionTableNotReplicated(ddlJob->targetRelationId);
```
mtuncer: might need a comment on function header

src/backend/distributed/utils/shardinterval_utils.c
onderkalaci: To prevent crash for append/range partitioned tables without any shards

src/backend/distributed/utils/shardinterval_utils.c
onderkalaci: > What happens if the first shard has a single placement but the rest has multiple ?

Hmm, that's true for append/range partitioned tables, I overlooked those. My goal was to make this function faster. I should probably do this only for hash tables.

src/backend/distributed/master/master_modify_multiple_shards.c
```diff
@@ -151,6 +152,7 @@ master_modify_multiple_shards(PG_FUNCTION_ARGS)
 			rangeVar->schemaname = schemaName;
 		}
 
+		EnsurePartitionTableNotReplicated(relationId);
```
mtuncer: Can we do without this ? In the end we are only deleting data, it is not different then delete or update command.

src/backend/distributed/master/master_repair_shards.c
```diff
@@ -237,6 +243,17 @@ RepairShardPlacement(int64 shardId, char *sourceNodeName, int32 sourceNodePort,
 								  "not supported.", relationName)));
 	}
 
+	/*
+	 * Let's not allow repairing partitions to prevent any edge cases.
+	 * We're already not allowing any kind of modifications on the partitions
+	 * so their placements are not likely to to be marked as INVALID. The only
+	 * possible case to mark placement of a partition as invalid is
+	 * "ALTER TABLE parent_table DETACH PARTITION partition_table". But,
```
mtuncer: My first understanding of `regular` was `non-distributed` table here. Could you reword this such that it will mean `non-partitioned but still distributed` ?

src/backend/distributed/master/master_repair_shards.c
```diff
@@ -275,6 +328,52 @@ RepairShardPlacement(int64 shardId, char *sourceNodeName, int32 sourceNodePort,
...
+ *
+ * The function returns a list of commands which re-creates all the partitions
+ * of the input shardInterval.
+ */
+static List *
+CopyPartitionShardsCommandList(ShardInterval *shardInterval, char *sourceNodeName,
+							   int32 sourceNodePort)
+{
+	Oid distributedTableId = shardInterval->relationId;
+
```
mtuncer: extra spaces

src/backend/distributed/master/master_repair_shards.c
```diff
@@ -350,11 +449,12 @@ SearchShardPlacementInList(List *shardPlacementList, char *nodeName, uint32 node
...
  * CopyShardCommandList generates command list to copy the given shard placement
- * from the source node to the target node.
+ * from the source node to the target node. Caller could optionally skip copying
+ * the data by the flag includeDataCopy.
  */
 List *
-CopyShardCommandList(ShardInterval *shardInterval,
-					 char *sourceNodeName, int32 sourceNodePort)
+CopyShardCommandList(ShardInterval *shardInterval, char *sourceNodeName,
+					 int32 sourceNodePort, bool includeDataCopy)
```
mtuncer: `copyData` would look better (to me at least)

src/backend/distributed/utils/shardinterval_utils.c
mtuncer: could you add a line describing that ?

src/backend/distributed/worker/worker_data_fetch_protocol.c
```diff
@@ -787,7 +790,25 @@ worker_append_table_to_shard(PG_FUNCTION_ARGS)
 
 	sourceQualifiedName = quote_qualified_identifier(sourceSchemaName, sourceTableName);
 	sourceCopyCommand = makeStringInfo();
-	appendStringInfo(sourceCopyCommand, COPY_OUT_COMMAND, sourceQualifiedName);
+
+	/*
+	 * Partitioned tables do not support "COPY table TO STDOUT". Thus, we use
```
mtuncer: Do we need this ? Even postgres does not support the usage and recommends `copy (select. ...) ...`

src/backend/distributed/commands/multi_copy.c
mtuncer: Only downside would be users won't be able to insert data into `cstore_fdw` partitions  in PG10.

src/backend/distributed/master/master_modify_multiple_shards.c
onderkalaci: The main motivation is to prevent marking placements `INVALID` and locking. So, we could mark placement invalid here as well.

src/backend/distributed/master/master_repair_shards.c
onderkalaci: This is directly coming from enterprise repo. So, I didn't want to diverge. (Btw, this will be merge conflict anyway)

src/backend/distributed/worker/worker_data_fetch_protocol.c
onderkalaci: I think I couldn't get your suggestion here.
If we don't do this, Postgres errors out when we send `COPY partitioned_table_shard ..` command, that's why I added this.

mtuncer: right. my comment was total bs then.

onderkalaci:
>  We do allow DROP TABLE partition, which seems fine to me given that we could only mark shard placements of non-existing shards in the worst case. What the reviewer think about it?

@mtuncer also thinks that these are OK, so leaving as-is.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -2931,25 +2932,59 @@ ErrorIfUnsupportedTruncateStmt(TruncateStmt *truncateStatement)
...
+		Relation relation = heap_openrv(relationRV, NoLock);
+		Oid relationId = RelationGetRelid(relation);
+
+		if (!IsDistributedTable(relationId))
+		{
+			heap_close(relation, NoLock);
+			continue;
+		}
+
+		EnsurePartitionTableNotReplicated(relationId);
```
mtuncer: what about foreign keys ?

mtuncer: no need to worry about those as it appears. can not have foreign keys when replication factor > 1

Issue #2382, opened 2018-09-14
TRUNCATE on multiple tables may fail if there is a foreign key to reference table


onderkalaci:
We could probably improve this behaviour by looking at all the tables that are truncated as opposed to looking each table seperately on the execution time.

```SQL
CREATE TABLE reference_table(id int PRIMARY KEY);
SELECT create_reference_table('reference_table');

CREATE TABLE distributed_table(id int PRIMARY KEY, value_1 int);
SELECT create_distributed_table('distributed_table', 'id');


ALTER TABLE 
	distributed_table 
ADD CONSTRAINT 
	fkey_delete FOREIGN KEY(value_1) 
REFERENCES 
	reference_table(id) ON DELETE CASCADE;
```

```SQL
-- The following command fails 
TRUNCATE distributed_table, reference_table;
ERROR:  cannot execute DDL on reference relation "reference_table" because there was a parallel DDL access to distributed relation "distributed_table" in the same transaction
HINT:  Try re-running the transaction with "SET LOCAL citus.multi_shard_modify_mode TO 'sequential';"
```

If you start the command with the `reference_table` it works just fine because we're automatically switching to `sequential mode`;
```
--
TRUNCATE reference_table, distributed_table;
TRUNCATE TABLE
```

Similarly, explicitly setting the mode to sequential just works
```
BEGIN;
         SET LOCAL citus.multi_shard_modify_mode TO 'sequential';
	TRUNCATE distributed_table, reference_table;
COMMIT;

```

furkansahin:
Just leaving it as a note here for the future owner;
I spent a small amount of time debugging that issue and the solution does not seem that straightforward. The way we use triggers in `truncate` seems like the reason of this issue. For each table in the truncate list, we call the trigger separately and Citus code does not seem to be aware of the list of the tables.

onderkalaci:
> . The way we use triggers in truncate seems like the reason of this issue

@mtuncer is currently working on a completely separate fix, which would move some parts of `TRUNCATE` functionality into `multi_ProcessUtility()`. After that change checked-in, we should be able to quickly fix this as well.

mtuncer:
update : issue still exists after introduced distributed locking to truncate in process utility

PR #2391, opened 2018-09-19
Add distributed locking to truncated mx tables

mtuncer:
DESCRIPTION: Add distributed locking to truncated mx tables

We acquire distributed lock on all mx nodes for truncated
tables before actually doing truncate operation.

This is needed for distributed serialization of the truncate
command without causing a deadlock.

### DIFF
--- a/src/backend/distributed/executor/multi_utility.c
+++ b/src/backend/distributed/executor/multi_utility.c
@@ -36,6 +36,7 @@
 #include "distributed/colocation_utils.h"
 #include "distributed/foreign_constraint.h"
 #include "distributed/intermediate_results.h"
+#include "distributed/listutils.h"
 #include "distributed/maintenanced.h"
 #include "distributed/master_metadata_utility.h"
 #include "distributed/master_protocol.h"
@@ -50,6 +51,7 @@
 #include "distributed/multi_utility.h" /* IWYU pragma: keep */
 #include "distributed/pg_dist_partition.h"
 #include "distributed/policy.h"
+#include "distributed/reference_table_utils.h"
 #include "distributed/relation_access_tracking.h"
 #include "distributed/resource_lock.h"
 #include "distributed/transaction_management.h"
@@ -85,6 +87,9 @@
 #include "utils/memutils.h"
 #include "utils/palloc.h"
 #include "utils/syscache.h"
+
+
+#define LOCK_RELATION_IF_EXISTS "SELECT lock_relation_if_exists('%s', '%s');"
 
 
 bool EnableDDLPropagation = true; /* ddl propagation is enabled */
@@ -146,6 +151,9 @@ static void ErrorIfAlterDropsPartitionColumn(AlterTableStmt *alterTableStatement
 static void ErrorIfUnsupportedSeqStmt(CreateSeqStmt *createSeqStmt);
 static void ErrorIfDistributedAlterSeqOwnedBy(AlterSeqStmt *alterSeqStmt);
 static void ErrorIfUnsupportedTruncateStmt(TruncateStmt *truncateStatement);
+static void ProcessTruncateStatement(TruncateStmt *truncateStatement);
+static void LockTruncatedRelationMetadataInWorkers(TruncateStmt *truncateStatement);
+static void AcquireDistributedLockOnRelations(List *relationIdList, LOCKMODE lockMode);
 static bool OptionsSpecifyOwnedBy(List *optionList, Oid *ownedByTableId);
 static void ErrorIfUnsupportedRenameStmt(RenameStmt *renameStmt);
 static void ErrorIfUnsupportedAlterAddConstraintStmt(AlterTableStmt *alterTableStatement);
@@ -386,6 +394,7 @@ multi_ProcessUtility(PlannedStmt *pstmt,
 	if (IsA(parsetree, TruncateStmt))
 	{
 		ErrorIfUnsupportedTruncateStmt((TruncateStmt *) parsetree);
+		ProcessTruncateStatement((TruncateStmt *) parsetree);
 	}
 
 	/* only generate worker DDLJobs if propagation is enabled */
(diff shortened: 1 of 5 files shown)


src/backend/distributed/executor/multi_utility.c
```diff
@@ -2921,6 +2934,126 @@ ErrorIfUnsupportedTruncateStmt(TruncateStmt *truncateStatement)
...
+ * we have both here is that we needed the ENUM for local locking function, and
+ * string representation for remote locking function.
+ *
+ * Notice that the locking functions is sent to all workers regardless of if
+ * it has metadata or not. This is because a worker node only knows itself
+ * and previous workers that has metadata sync turned on. It does not know
+ * about nodes that have metadata sync turned on afterwards.
+ */
+static void
+AcquireDistributedLockOnRelations(List *relationIdList, LOCKMODE localLockMode,
```
onderkalaci: We have a function which converts text lockmode to `LOCKMODE` : `LockModeTextToLockMode()`

src/backend/distributed/executor/multi_utility.c
```diff
@@ -2921,6 +2934,126 @@ ErrorIfUnsupportedTruncateStmt(TruncateStmt *truncateStatement)
...
+				int nodePort = workerNode->workerPort;
+
+				/* if local node is one of the targets, acquire the lock locally */
+				if (workerNode->groupId == GetLocalGroupId())
+				{
+					LockRelationOid(relationId, localLockMode);
+					continue;
+				}
+
+				SendCommandToWorker(nodeName, nodePort, lockRelationCommand->data);
```
onderkalaci: Given that we send the `lock_relation_if_exists()` here, we should be able to remove the similar code in `master_modify_multiple_shards.c` to get rid of the duplication.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -2921,6 +2934,126 @@ ErrorIfUnsupportedTruncateStmt(TruncateStmt *truncateStatement)
...
+															 referencingRelationId);
+		}
+
+		heap_close(relation, NoLock);
+	}
+
+	if (distributedRelationList != NIL)
+	{
+		BeginOrContinueCoordinatedTransaction();
+		AcquireDistributedLockOnRelations(distributedRelationList, AccessExclusiveLock,
```
onderkalaci: We've a function named `ClusterHasKnownMetadataWorkers()` which returns true if the cluster is an MX cluster.

So, we should not be getting into this logic at all if `!ClusterHasKnownMetadataWorkers()`. Otherwise, we'd be sending these commands even if we're not in an MX cluster.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -2921,6 +2934,126 @@ ErrorIfUnsupportedTruncateStmt(TruncateStmt *truncateStatement)
...
+ * This is only for distributed serialization of truncate commands.
+ * The function assumes that there is no foreign key relation between
+ * non-distributed and distributed relations.
+ */
+static void
+ProcessTruncateStatement(TruncateStmt *truncateStatement)
+{
+	List *distributedRelationList = NIL;
+
+	ListCell *relationCell = NULL;
```
onderkalaci: Shall we move the loop into a separate function like: `LockTableInMetadataWorkers()`. My reasoning is that I'll expand `ProcessTruncateStatement ()` for fixing #2382 such that I can add another function here:
```C
def ProcessTruncateStatement():
      LockTableInMetadataWorkers()
      SetSequentialModeIfNecessary()
      /* and others in the future */
```

Also, we can return asap in `LockTableInMetadataWorkers ()`  when `! ClusterHasKnownMetadataWorkers ()`

src/backend/distributed/utils/listutils.c
```diff
@@ -59,6 +59,11 @@ SortList(List *pointerList, int (*comparisonFunction)(const void *, const void *
 
 	pfree(array);
 
+	if (sortedList != NIL)
```
onderkalaci: I'm surprised that we haven't set this for a long time. I guess there are no other such fields that we need to set?

mtuncer:
tested pgbench with 4 worker cluster in which 3 of them are set as mx nodes (9700, 9701, 9702)

- create pgbench tables at coordinator
```
pgbench -i postgres
```

- Then connect to the coordinator and set up tables
```sql

SELECT create_distributed_table('pgbench_accounts', 'aid');
SELECT create_distributed_table('pgbench_branches', 'bid');
SELECT create_distributed_table('pgbench_history', 'tid');
SELECT create_distributed_table('pgbench_tellers', 'tid');
```

- run the tests from command line
```
pgbench -b tpcb-like -P 3 -c8 -j8 -t 1000 -p 9700 postgres &
pgbench -b tpcb-like -P 3 -c8 -j8 -t 1000 -p 9701 postgres &
pgbench -b tpcb-like -P 3 -c8 -j8 -t 1000 -p 9702 postgres &
```

Result:

We used to get distributed deadlocks before the fix, now we don't get them

src/backend/distributed/executor/multi_utility.c
```diff
@@ -23,10 +23,12 @@
 #include "access/xact.h"
 #include "catalog/catalog.h"
 #include "catalog/dependency.h"
+#include "catalog/heap.h"
 #include "catalog/index.h"
 #include "catalog/indexing.h"
 #include "catalog/pg_attribute.h"
 #include "catalog/pg_class.h"
+#include "catalog/pg_inherits.h"
```
onderkalaci: Do we really need `pg_inherits `?

src/backend/distributed/executor/multi_utility.c
```diff
@@ -87,6 +91,10 @@
 #include "utils/syscache.h"
 
 
+#define LOCK_RELATION_IF_EXISTS "SELECT lock_relation_if_exists('%s', '%s');"
+#define REMOTE_LOCK_MODE_FOR_TRUNCATE "ACCESS EXCLUSIVE"
```
onderkalaci: I don't have a good reason/explanation for this, but,  `REMOTE_LOCK_MODE_FOR_TRUNCATE ` feels like a better fit for the function definition, with `char *remoteLockModeForTruncate = "ACCESS EXCLUSIVE"`. I think that is because we don't have any other macros that defines a lock mode.

Feel free to ignore.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -2921,6 +2934,126 @@ ErrorIfUnsupportedTruncateStmt(TruncateStmt *truncateStatement)
 }
 
 
+/*
+ * ProcessTruncateStatement handles distributed locking
```
onderkalaci: Shall we mention that the actual `TRUNCATE` is done on the trigger and maybe the reasoning behind leaving that there?

src/backend/distributed/executor/multi_utility.c
```diff
@@ -2921,6 +2934,126 @@ ErrorIfUnsupportedTruncateStmt(TruncateStmt *truncateStatement)
...
+static void
+AcquireDistributedLockOnRelations(List *relationIdList, LOCKMODE localLockMode,
+								  char *remoteLockMode)
+{
+	ListCell *relationIdCell = NULL;
+	List *workerNodeList = ActivePrimaryNodeList();
+	relationIdList = SortList(relationIdList, CompareOids);
+	workerNodeList = SortList(workerNodeList, CompareWorkerNodes);
+
+	foreach(relationIdCell, relationIdList)
```
onderkalaci: Hmm, 
```
foreach relation
      for each worker:
```
could be an expensive loop if there are many foreign keys. What if we do the following:

```
foreach relation:
    build the string

for each worker:
     send the command (or acquire the list locally)
```

I think that would even make the window for distributed deadlocks with concurrent `SELECT`s smaller. Because, with the current commit, if you run a concurrent `SELECT`, we're likely to distributed (or local) deadlocks while waiting for relation locks.

(Or even changing the API for `LOCK_RELATION_IF_EXISTS ` to get a list of relations could be simpler?)

src/backend/distributed/executor/multi_utility.c
mtuncer: done, did some code shuffling around

src/backend/distributed/executor/multi_utility.c
mtuncer: done

src/backend/distributed/executor/multi_utility.c
mtuncer: correct, I was only checking `ShouldSyncTableMetadata()` per relation. This is a shortcut to not  to enter that path.

src/backend/distributed/executor/multi_utility.c
mtuncer: gone after ditching lock mode text to get it from enum.

src/backend/distributed/executor/multi_utility.c
mtuncer: I am not fond of this change but added it.

src/backend/distributed/executor/multi_utility.c
mtuncer: added

src/backend/distributed/executor/multi_utility.c
```diff
@@ -2921,6 +2930,142 @@ ErrorIfUnsupportedTruncateStmt(TruncateStmt *truncateStatement)
...
+ * Actual distributed truncation occurs inside truncate trigger.
+ *
+ * This is only for distributed serialization of truncate commands.
+ * The function assumes that there is no foreign key relation between
+ * non-distributed and distributed relations.
+ */
+static void
+ProcessTruncateStatement(TruncateStmt *truncateStatement)
+{
+	LockTruncatedRelationMetadataInWorkers(truncateStatement);
```
onderkalaci: I think one side effect of acquiring locks before process utility is that even if there'll be some errors, we'll send the commands. However, that seems OK, right?
An example:
```SQL
postgres=# TRUNCATE users_table;
LOG:  issuing BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED;SELECT assign_distributed_transaction_id(0, 351, '2018-09-21 11:05:28.847309+03');
LOG:  issuing SELECT lock_relation_if_exists('public.users_table', 'ACCESS EXCLUSIVE');
LOG:  issuing BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED;SELECT assign_distributed_transaction_id(0, 351, '2018-09-21 11:05:28.847309+03');
LOG:  issuing SELECT lock_relation_if_exists('public.users_table', 'ACCESS EXCLUSIVE');
LOG:  issuing BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED;SELECT assign_distributed_transaction_id(0, 351, '2018-09-21 11:05:28.847309+03');
LOG:  issuing SELECT lock_relation_if_exists('public.users_table', 'ACCESS EXCLUSIVE');
LOG:  issuing SELECT lock_relation_if_exists('public.events_table', 'ACCESS EXCLUSIVE');
LOG:  issuing SELECT lock_relation_if_exists('public.events_table', 'ACCESS EXCLUSIVE');
LOG:  issuing SELECT lock_relation_if_exists('public.events_table', 'ACCESS EXCLUSIVE');
LOG:  issuing ROLLBACK
LOG:  issuing ROLLBACK
LOG:  issuing ROLLBACK
ERROR:  cannot truncate a table referenced in a foreign key constraint
DETAIL:  Table "events_table" references "users_table".
HINT:  Truncate table "events_table" at the same time, or use TRUNCATE ... CASCADE.
```

src/backend/distributed/utils/resource_lock.c
```diff
@@ -618,3 +647,120 @@ LockPartitionRelations(Oid relationId, LOCKMODE lockMode)
...
+ * The function errors out if the input lock mode isn't defined in the PostgreSQL's
+ * explicit locking table.
+ */
+LOCKMODE
+LockModeTextToLockMode(const char *lockModeName)
+{
+	int index = 0;
+	LOCKMODE lockMode = -1;
+
+	while (true)
```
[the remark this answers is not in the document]
mtuncer: introduced variable to keep counts

src/backend/distributed/executor/multi_utility.c
```diff
@@ -2921,6 +2930,141 @@ ErrorIfUnsupportedTruncateStmt(TruncateStmt *truncateStatement)
...
+			distributedRelationList = list_append_unique_oid(distributedRelationList,
+															 referencingRelationId);
+		}
+
+		heap_close(relation, NoLock);
+	}
+
+	if (distributedRelationList != NIL)
+	{
+		BeginOrContinueCoordinatedTransaction();
```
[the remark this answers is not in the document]
mtuncer: ok

src/backend/distributed/utils/resource_lock.c
```diff
@@ -618,3 +647,120 @@ LockPartitionRelations(Oid relationId, LOCKMODE lockMode)
...
+    /* ensure that we're in a transaction block */
+	RequireTransactionBlock(true, "lock_relation_if_exists");
+
+	relationId = ResolveRelationId(relationName, true);
+	if (!OidIsValid(relationId))
+	{
+		PG_RETURN_BOOL(false);
+	}
+
+    /* get the lock mode */
```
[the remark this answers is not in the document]
mtuncer: function was moved from another file to here. I did not check the contents. looking into it.

src/backend/distributed/utils/resource_lock.c
```diff
@@ -70,7 +99,7 @@ lock_shard_metadata(PG_FUNCTION_ARGS)
 		ereport(ERROR, (errmsg("no locks specified")));
 	}
 
-	/* we don't want random users to block writes */
+    /* we don't want random users to block writes */
```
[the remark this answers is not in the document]
mtuncer: this is where I was puzzled.
my local citus_indent introduced those. I reverted them before submitting because they were irrelevant to what I was doing.

This time travis rejected it.  I fallen back to what citus_indent did.

src/backend/distributed/executor/multi_utility.c
mtuncer: once you hit `ERROR` transaction is rolled back and locks are released.

PR #2404, opened 2018-09-24
Make sure to use correct execution mode for TRUNCATE

### LINKED ISSUE #2382: TRUNCATE on multiple tables may fail if there is a foreign key to reference table

onderkalaci:
Fixes #2382

We used to set the execution mode in the truncate trigger. However,
when multiple tables are truncated with a single command, we could
set the execution mode very late. Instead, now set the execution mode
on the utility hook.

### DIFF
--- a/src/backend/distributed/executor/multi_utility.c
+++ b/src/backend/distributed/executor/multi_utility.c
@@ -152,6 +152,7 @@ static void ErrorIfUnsupportedSeqStmt(CreateSeqStmt *createSeqStmt);
 static void ErrorIfDistributedAlterSeqOwnedBy(AlterSeqStmt *alterSeqStmt);
 static void ErrorIfUnsupportedTruncateStmt(TruncateStmt *truncateStatement);
 static void ProcessTruncateStatement(TruncateStmt *truncateStatement);
+static void ExecuteTruncateStmtSequentialIfNecessary(TruncateStmt *command);
 static void EnsurePartitionTableNotReplicatedForTruncate(TruncateStmt *truncateStatement);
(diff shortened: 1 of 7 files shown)


onderkalaci:
> However, we switch to the sequential execution mode in this case too.

We talked with @furkansahin and decided to add a `DEBUG1` message because it is basically taking the same precaution that `CheckConflictingRelationAccesses()` does. So, now both produce the same `DEBUG1` message and behaving exactly the same.

PR #2410, opened 2018-09-26
Do not create inherited constraints at worker tables

### LINKED ISSUE #2406: partitions with foreign keys can not be created in mx mode
Following fails when PG11  cluster is configured in MX mode

```sql
create table l1(a int primary key, b int);
select create_distributed_table('l1', 'a');

create table l2(a int references l1(a), b int) partition by range (b);
select create_distributed_table('l2', 'a');
create table l2_1 partition of l2 for values from (0) to (10);
```

Result
```
ERROR:  duplicate key value violates unique constraint "pg_constraint_conrelid_contypid_conname_index"
DETAIL:  Key (conrelid, contypid, conname)=(17476, 0, l2_a_fkey) already exists.
```

It works fine on cluster without mx.

### LINKED ISSUE #2415: foreign keys are created twice in partition tables
This is side issue when we encounter while working on #2406. 

When we attach a partition table a parent table we do
1. create partition table shards as a regular table
2. create constraints at worker shards
3. attach partition shard to a parent shard.

Postgres automatically create foreign tables on parent partitioned table when we attach partitions, therefore we should not create  `inherited` constraints at partition shards when we are creating them.

mtuncer:
PG now allows foreign keys on partitioned tables. Each foreign key constraint on partitioned table is propagated down to partitions.

We used to create all constraints on shards when we are creating a new shards or just simply moving a shard from one worker to another.  We also used the same logic when creating a copy of coordinator table in mx node. 

With this change we create the constraint on worker node only if it is not an inherited constraint.

Fixes #2415

### DIFF
--- a/src/backend/distributed/ddl/foreign_constraint.c
+++ b/src/backend/distributed/ddl/foreign_constraint.c
@@ -493,7 +493,13 @@ GetTableForeignConstraintCommands(Oid relationId)
 	{
 		Form_pg_constraint constraintForm = (Form_pg_constraint) GETSTRUCT(heapTuple);
 
-		if (constraintForm->contype == CONSTRAINT_FOREIGN)
+#if (PG_VERSION_NUM >= 110000)
+		bool inheritedConstraint = OidIsValid(constraintForm->conparentid);
+#else
+		bool inheritedConstraint = false;
+#endif
+
+		if (!inheritedConstraint && constraintForm->contype == CONSTRAINT_FOREIGN)
 		{
 			Oid constraintId = get_relation_constraint_oid(relationId,
 														   constraintForm->conname.data,
--- a/src/test/regress/expected/multi_partitioning.out
+++ b/src/test/regress/expected/multi_partitioning.out
@@ -717,6 +717,76 @@ NOTICE:  Copying data from local table...
 (1 row)
 
 ALTER TABLE partitioning_test ADD CONSTRAINT partitioning_reference_fkey FOREIGN KEY (id) REFERENCES partitioning_test_reference(id) ON DELETE CASCADE;
+CREATE TABLE partitioning_test_foreign_key(id int PRIMARY KEY, value int);
+SELECT create_distributed_table('partitioning_test_foreign_key', 'id');
+ create_distributed_table 
+--------------------------
+ 
+(1 row)
+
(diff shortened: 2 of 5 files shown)


onderkalaci:
I hit yet another issue when I follow the steps on #2406. This PR fixes some part of our main goal of creating partitioned tables with foreign keys on MX, but, one more issue become visible after the fix.

We're hitting the same issue that we already fixed in non-mx case when we implement foreign keys, which is commented in detail [here](https://github.com/citusdata/citus/blob/master/src/backend/distributed/executor/multi_utility.c#L1604).

In MX with partitioning case, when we execute `ALTER TABLE public.l2 ATTACH PARTITION public.l2_1 FOR VALUES FROM (0) TO (10);` on a worker, if the parent has a foreign key, it tries to validate the foreign key and send a query to ensure the validity of the foreign constraint. That query fails because the shards are not visible yet outside the same distributed transaction:
```
...
LOG:  issuing ALTER TABLE public.l2 ATTACH PARTITION public.l2_1 FOR VALUES FROM (0) TO (10);

...
DEBUG:  relation "l2_1_102149" does not exist
DETAIL:  WARNING from localhost:9702
DEBUG:  relation "l2_1_102150" does not exist
DETAIL:  WARNING from localhost:9702
DEBUG:  relation "l2_1_102151" does not exist
DETAIL:  WARNING from localhost:9702
DEBUG:  relation "l2_1_102152" does not exist
DETAIL:  WARNING from localhost:9702
DEBUG:  relation "l2_1_102153" does not exist
DETAIL:  WARNING from localhost:9702
DEBUG:  relation "l2_1_102154" does not exist
DETAIL:  WARNING from localhost:9702
DEBUG:  relation "l2_1_102155" does not exist
DETAIL:  WARNING from localhost:9702
...
CONTEXT:  SQL statement "SELECT fk."a" FROM ONLY "public"."l2_1" fk LEFT OUTER JOIN ONLY "public"."l1" pk ON ( pk."a" OPERATOR(pg_catalog.=) fk."a") WHERE pk."a" IS NULL AND (fk."a" IS NOT NULL)"
while executing command on localhost:9700
```

We should find a way to add the foreign keys invalid and later mark them back as valid. We do a similar trick in online rebalancer + foreign keys to reference tables.

mtuncer:
created a new issue about non-mx tables and marked this PR to resolve that issue. We still need to solve mx one.

PR #2411, opened 2018-09-27
Failure tests for modifying multiple shards in txn

jasonmp85:
Tests various failure points during a multi-shard modification within a transaction with multiple statements. Verifies three cases:

  * Reference tables (single shard, many placements)
  * Normal table with replication factor two
  * Multi-shard table with no replication

In the replication-factor case, we expect shard health to be affected in some transactions; most others fail the transaction entirely and all we need verify is that no effects of the transaction are visible.

Had trouble testing the final PREPARE/COMMIT/ROLLBACK phase of the 2pc, in particular because the error message produced includes the PID of the backend, which is unpredictable.

### DIFF
--- a/src/test/regress/sql/failure_multi_dml.sql
+++ b/src/test/regress/sql/failure_multi_dml.sql
@@ -1,0 +1,179 @@
+SELECT citus.mitmproxy('conn.allow()');
+
+SET citus.shard_count = 2;
+SET citus.shard_replication_factor = 1; -- one shard per worker
+SET citus.next_shard_id TO 103400;
+ALTER SEQUENCE pg_catalog.pg_dist_placement_placementid_seq RESTART 100;
+
+CREATE TABLE dml_test (id integer, name text);
+SELECT create_distributed_table('dml_test', 'id');
+
+COPY dml_test FROM STDIN WITH CSV;
+1,Alpha
+2,Beta
+3,Gamma
+4,Delta
+\.
+
+SELECT citus.clear_network_traffic();
+
+---- test multiple statements spanning multiple shards,
+---- at each significant point. These transactions are 2pc
+
+-- fail at DELETE
+SELECT citus.mitmproxy('conn.onQuery(query="^DELETE").kill()');
+
+BEGIN;
+DELETE FROM dml_test WHERE id = 1;
+DELETE FROM dml_test WHERE id = 2;
+INSERT INTO dml_test VALUES (5, 'Epsilon');
+UPDATE dml_test SET name = 'alpha' WHERE id = 1;
+UPDATE dml_test SET name = 'gamma' WHERE id = 3;
+COMMIT;
+
+--- shouldn't see any changes performed in failed transaction
+SELECT * FROM dml_test ORDER BY id ASC;
+
+-- fail at INSERT
+SELECT citus.mitmproxy('conn.onQuery(query="^INSERT").kill()');
+
+BEGIN;
+DELETE FROM dml_test WHERE id = 1;
+DELETE FROM dml_test WHERE id = 2;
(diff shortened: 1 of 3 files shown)


sumedhpathak:
Assigning this failure automation PR to @thanodnl

src/test/regress/sql/failure_multi_dml.sql
```diff
@@ -0,0 +1,148 @@
...
+DELETE FROM dml_test WHERE id = 2;
+INSERT INTO dml_test VALUES (5, 'Epsilon');
+UPDATE dml_test SET name = 'alpha' WHERE id = 1;
+UPDATE dml_test SET name = 'gamma' WHERE id = 3;
+COMMIT;
+
+--- shouldn't see any changes after failed UPDATE
+SELECT * FROM dml_test ORDER BY id ASC;
+
+-- FIXME: Can't test this because the prepared transaction ID has PID in it
```
thanodnl: I might have an idea here, I will have a look tomorrow

src/test/regress/sql/failure_multi_dml.sql
```diff
@@ -0,0 +1,148 @@
...
+
+-- fail at COMMIT (by failing PREPARE)
+-- SELECT citus.mitmproxy('conn.onQuery(query="^PREPARE").kill()');
```
thanodnl: This makes me assume the query is a 2pc, can we also check to fail at commit, I would expect to see all changes after recovery.

src/test/regress/sql/failure_multi_dml.sql
```diff
@@ -0,0 +1,148 @@
...
+
+COPY dml_test FROM STDIN WITH CSV;
+1,Alpha
+2,Beta
+3,Gamma
+4,Delta
+\.
+
+-- fail at COMMIT (by failing to PREPARE)
+SELECT citus.mitmproxy('conn.onQuery(query="^PREPARE").kill()');
```
thanodnl: Would it be worth to test the behaviour when a commit to a reference table fails? I guess it behaves like a distributed table, eg. recover after the 2pc is completed.

src/test/regress/sql/failure_multi_dml.sql
thanodnl: I have done some tests. I think the pid is in the name of the prepared transaction.

We can prevent that name to show up in the tests by a `SET LOCAL client_min_messages TO 'FATAL';` in the transaction block. This would prevent the error messages to show.

Even though the error message will not show in the log output we can still verify its behaviour by inspecting the contents of the table as we do with the other tests via the commented out `SELECT` statement.

onderkalaci: To give some more context on this, we had a similar discussion around testing this with @lithp  when he first implemented the infrastructure.

Basically, @lithp proposed to add a GUC, which is only set during failure tests running, to not add the PID in the 2PC identifiers. However, I pushed back because (a) that was the only source code change in Citus, and implementing the failure testing automation shouldn't have required any code changes (b) I didn't want to add any complexity to 2PC id generation codes just for testing purposes.

I proposed him another approach for writing those tests, you can see it [here](https://github.com/citusdata/citus/pull/2119#discussion_r190549470)

src/test/regress/sql/failure_multi_dml.sql
jasonmp85: Different test suite (there was a separate task for reference table modifications).

src/test/regress/sql/failure_multi_dml.sql
jasonmp85: Went with the `FATAL` approach for now.

PR #2412, opened 2018-09-27
Show router modify,select and real-time queries on MX views

### LINKED ISSUE #2407: MX views don't show activites that are not from coordinated transactions
Newly introduced MX views `citus_dist_stat_activity`, `citus_worker_stat_activity` and `citus_lock_waits` don't show activities from non-coordinated transactions. As an example single row inserts do not appear on them. 

In order to show these as well, we need to left join pg_stat_activity with all active transactions and assign distributed transaction ids to all activities in the cluster.

velioglu:
Fixes #2407.

This PR adds distributed transaction ID to the router select, router modify and real-time select queries by setting id within custom scan functions and making that function idempotent. Beside, it also changes definition of `citus_dist_stat_activity` and `citus_worker_stat_activity` views to show each transactions in the cluster within that views.


src/backend/distributed/transaction/citus_dist_stat_activity.c
```diff
@@ -154,12 +154,50 @@
...
+		LEFT JOIN \
+		get_all_active_transactions() AS dist_txs(database_id, process_id, initiator_node_identifier, worker_query, transaction_number, transaction_stamp) \
+		ON pg_stat_activity.pid = dist_txs.process_id \
 	WHERE \
-		pg_stat_activity.pid = dist_txs.process_id \
-			AND \
-		dist_txs.worker_query = %s; "
+		(dist_txs.worker_query IS NULL OR dist_txs.worker_query = false) \
+		AND \
+		pg_stat_activity.query NOT ILIKE '%stat_activity%';"
```
marcocitus: Can we add `AND application_name = 'citus'`? we're not so interested in background workers and such.

src/backend/distributed/utils/resource_lock.c
```diff
@@ -752,14 +752,14 @@ lock_relation_if_exists(PG_FUNCTION_ARGS)
...
+	/* get the lock mode */
 	lockMode = LockModeTextToLockMode(lockModeCString);
 
-    /* resolve relationId from passed in schema and relation name */
+	/* resolve relationId from passed in schema and relation name */
 	relationNameList = textToQualifiedNameList(relationName);
 	relation = makeRangeVarFromNameList(relationNameList);
 
-    /* lock the relation with the lock mode */
+	/* lock the relation with the lock mode */
```
marcocitus: are these just citus_indent changes? could we do them as a separate commit?

src/backend/distributed/transaction/citus_dist_stat_activity.c
velioglu: I have added that check for worker stat activities, do we need to add it for dist stat activities as well?

src/backend/distributed/utils/resource_lock.c
onderkalaci: @velioglu Maybe double check your uncrustify version. We had similar issues last week. I think travis uses `uncrustify 0.63`

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -537,6 +538,8 @@ CitusModifyBeginScan(CustomScanState *node, EState *estate, int eflags)
 
 	/* modify tasks are always assigned using first-replica policy */
 	workerJob->taskList = FirstReplicaAssignTaskList(taskList);
+
+	AssignDistributedTransactionId();
```
onderkalaci: I think having this at the beginning of the function makes it easier to read / follow. (at least to me)

And also, we should remove `AssignDistributedTransactionId ()` from `BeginCoordinatedTransaction()`. Because with this change we're decoupling distributed transaction ids from coordinated transactions. And, that's actually a big thing.

I'd even implement a new function called `CitusBeginScan()` which simply calls `AssignDistributedTransactionId();`. And, later I'd call that function in both `CitusSelectBeginScan()` and `CitusModifyBeginScan ()` so that we could potentially add more stuff there in the future. Or, if someone implements another `___Scan__()` which could call that function. Anyway, we could skip that for now.

src/backend/distributed/transaction/backend_data.c
```diff
@@ -68,6 +68,7 @@ static void CheckReturnSetInfo(ReturnSetInfo *returnSetInfo);
 static shmem_startup_hook_type prev_shmem_startup_hook = NULL;
 static BackendManagementShmemData *backendManagementShmemData = NULL;
 static BackendData *MyBackendData = NULL;
+static bool setDistributedTransactionId = false;
```
onderkalaci: Maybe `distributedTransactionIdAlreadySet `

src/backend/distributed/transaction/backend_data.c
```diff
@@ -710,6 +712,12 @@ AssignDistributedTransactionId(void)
 	pg_atomic_uint64 *transactionNumberSequence =
 		&backendManagementShmemData->nextTransactionNumber;
 
+	/* return idempotent transaction id for consecutive requests from a same session */
```
onderkalaci: Minor: update the function comment to say that this function is idempotent. (e.g., the last sentence of the comment now isn't accurate)

src/backend/distributed/transaction/backend_data.c
```diff
@@ -632,6 +633,7 @@ UnSetDistributedTransactionId(void)
 		MyBackendData->transactionId.transactionOriginator = false;
 		MyBackendData->transactionId.transactionNumber = 0;
 		MyBackendData->transactionId.timestamp = 0;
+		setDistributedTransactionId = false;
```
onderkalaci: Minor readability: We don't need to do this within spinlocks. Let's maybe do this right after we release the lock so that it becomes slightly more accurate in terms of its semantic.

src/backend/distributed/transaction/citus_dist_stat_activity.c
```diff
@@ -273,17 +349,10 @@ Datum
...
 
 	CheckCitusVersion(ERROR);
 
-	/* set the workerQuery to false in the query */
-	citusDistStatQuery = makeStringInfo();
-	appendStringInfo(citusDistStatQuery, CITUS_DIST_STAT_ACTIVITY_QUERY,
-					 workerQuery);
-
-	citusDistStatStatements = CitusDistStatActivity(citusDistStatQuery->data);
+	citusDistStatStatements = CitusDistStatActivity(CITUS_DIST_STAT_ACTIVITY_QUERY);
```
onderkalaci: The function name `CitusDistStatActivity ()` seems confusing. (even though I've implemented that). Shall we rename it to `CitusStatActivity()`?

src/backend/distributed/executor/multi_router_executor.c
onderkalaci: > And also, we should remove AssignDistributedTransactionId () from BeginCoordinatedTransaction(). Because with this change we're decoupling distributed transaction ids from coordinated transactions. And, that's actually a big thing.

No, that'd be wrong due to DDLs/TRUNCATE etc. So, not remove from there 😄

src/backend/distributed/transaction/backend_data.c
```diff
@@ -701,19 +704,27 @@ GetCurrentDistributedTransactionId(void)
...
  */
 void
 AssignDistributedTransactionId(void)
 {
 	pg_atomic_uint64 *transactionNumberSequence =
 		&backendManagementShmemData->nextTransactionNumber;
 
-	uint64 nextTransactionNumber = pg_atomic_fetch_add_u64(transactionNumberSequence, 1);
+	uint64 nextTransactionNumber = 0;
 	int localGroupId = GetLocalGroupId();
```
onderkalaci: Minor: why don't we defer other function calls as well?

src/backend/distributed/transaction/backend_data.c
```diff
@@ -701,19 +704,27 @@ GetCurrentDistributedTransactionId(void)
...
 	TimestampTz currentTimestamp = GetCurrentTimestamp();
 
+	/* return idempotent transaction id for consecutive requests from a same session */
+	if (distributedTransactionIdAlreadySet)
+	{
+		return;
+	}
+
+	nextTransactionNumber = pg_atomic_fetch_add_u64(transactionNumberSequence, 1);
```
onderkalaci: Given that we'll call this function a lot more frequently, we're increasing the probability of overflows. Do you see any issues in case the `transactionNumber` overflows?

onderkalaci:
> Hmm, this change adds a lot of overhead to the distributed deadlock detection for processing transactions that could never be part of a distributed deadlock.

Also, note that this change would add considerable overhead to 2PC recovery given that 2PC recovery code relies on `ActiveDistributedTransactionNumbers()` which could yield lots of unnecessary transaction numbers.

onderkalaci:
> Should we add a flag indicating that these are uncoordinated distributed transactions?

So, we have two options:

Option 1: Do not assign distributed transaction id for every backend. Instead, mark the backend as `initializedByCitus` (or something like that). In this case, we couple distributed transaction ids with coordinated transactions.

Option 2: Always assign distributed transaction ids, but also mark the necessary backends as `coordinatedTransaction`s. Here, we decouple distributed transaction id and coordinated transactions.

I think picking `Option 2` creates a better semantic regarding transaction ids and coordinated transaction concepts. 

But, I'm a little bit concerned about `pg_atomic_fetch_add_u64()` becoming a bottleneck for busy servers in that case. Also, that'd make any future callers of distributed transaction ids be careful about the distinction of coordinated vs non-coordinated transactions.

So, my slight preference is still on option 1 unless we ensure that `pg_atomic_fetch_add_u64()` never becomes the bottleneck.

velioglu:
To keep the current logic as much as possible and not to face with the issue Onder introduced, I'm also  inclined to implement first option

src/backend/distributed/executor/citus_custom_scan.c
```diff
@@ -250,12 +251,13 @@ DelayedErrorCreateScan(CustomScan *scan)
 
 
 /*
- * CitusSelectBeginScan is an empty function for BeginCustomScan callback.
+ * CitusSelectBeginScan assigns distributed transaction id for queries using
```
onderkalaci: comment is stale

src/backend/distributed/transaction/backend_data.c
```diff
@@ -720,9 +721,41 @@ AssignDistributedTransactionId(void)
...
+/*
+ * MarkCitusInitiatedBackend set that backend is initiated by Citus.
+ */
+void
+MarkCitusInitiatedBackend(void)
+{
+	SpinLockAcquire(&MyBackendData->mutex);
+
+	MyBackendData->citusInitiatedBackend = true;
+	MyBackendData->transactionId.transactionOriginator = true;
```
onderkalaci: I think setting `MyBackendData->transactionId.transactionOriginator ` is wrong and dangerous. 

At this point, we don't know whether the `transactionId` is valid or not.

Also, why do we need to do that?

src/backend/distributed/transaction/backend_data.c
```diff
@@ -720,9 +721,41 @@ AssignDistributedTransactionId(void)
...
 	MyBackendData->transactionId.transactionOriginator = true;
-	MyBackendData->transactionId.transactionNumber =
-		nextTransactionNumber;
+	MyBackendData->transactionId.transactionNumber = nextTransactionNumber;
 	MyBackendData->transactionId.timestamp = currentTimestamp;
+	MyBackendData->citusInitiatedBackend = true;
+
+	SpinLockRelease(&MyBackendData->mutex);
+
+	MarkCitusInitiatedBackend();
```
onderkalaci: Why do we call it here? You've already marked it above.

src/backend/distributed/transaction/backend_data.c
```diff
@@ -634,6 +636,8 @@ UnSetDistributedTransactionId(void)
 		MyBackendData->transactionId.timestamp = 0;
 
 		SpinLockRelease(&MyBackendData->mutex);
+
+		UnmarkCitusInitiatedBackend();
```
onderkalaci: Can't we do that while the spinlock is already held? 

I think we can introduce a function called `UnmarkCitusInitiatedBackendNoLock()` which doesn't acquire the lock and relies on the called holding the locks.

We could even not have the function and mark the field itself directly.

src/backend/distributed/transaction/backend_data.c
```diff
@@ -65,6 +65,7 @@ typedef struct BackendManagementShmemData
 static void StoreAllActiveTransactions(Tuplestorestate *tupleStore, TupleDesc
 									   tupleDescriptor);
 static void CheckReturnSetInfo(ReturnSetInfo *returnSetInfo);
+static void UnmarkCitusInitiatedBackend();
```
onderkalaci: Minor: let's add a single newline after the function declaration.

src/backend/distributed/transaction/backend_data.c
```diff
@@ -123,6 +124,7 @@ assign_distributed_transaction_id(PG_FUNCTION_ARGS)
 	MyBackendData->transactionId.transactionNumber = PG_GETARG_INT64(1);
 	MyBackendData->transactionId.timestamp = PG_GETARG_TIMESTAMPTZ(2);
 	MyBackendData->transactionId.transactionOriginator = false;
+	MyBackendData->citusInitiatedBackend = true;
```
onderkalaci: I don't think we need this here. We're already having a `LEFT JOIN` on the worker queries so we don't rely on `citusInitiatedBackend` set to true.

 And, that seems the reason you needed to set `transactionOriginator` to true in `MarkCitusInitiatedBackend ()`.

Our goal is to only mark coordinator processes with `citusInitiatedBackend`. Hence, the name `citusInitiatedBackend` sounds not accurate at this point. What about we use `citusInitiatedCoordinatorBackend` and do not set on `assign_distributed_transaction_id ()`? (If we do that, please go over the code and update all related things such as `MarkCitusInitiatedBackend ` -> `MarkCitusInitiatedCoordinatorBackend `)

src/backend/distributed/transaction/backend_data.c
```diff
@@ -701,15 +705,12 @@ GetCurrentDistributedTransactionId(void)
  * sets it for the current backend. It also sets the databaseId and
  * processId fields.
  *
- * This function should only be called on BeginCoordinatedTransaction(). Any other
```
onderkalaci: Let's add this comment back

marcocitus:
Option 1 seems reasonable to me. Distributed transaction IDs that aren't used seem like unnecessary overhead.

src/backend/distributed/transaction/backend_data.c
```diff
@@ -720,14 +723,40 @@ AssignDistributedTransactionId(void)
...
+	SpinLockRelease(&MyBackendData->mutex);
+}
+
+
+/*
+ * UnmarkCitusInitiatedCoordinatorBackendNoLock unset that coordinator backend is
+ * initiated by Citus.
+ */
+static void
+UnmarkCitusInitiatedCoordinatorBackendNoLock(void)
```
onderkalaci: Minor: I'd be OK to get rid of this function given that we only use it in only one place and it doesn't do anything interesting 😄 .

src/backend/distributed/transaction/backend_data.c
```diff
@@ -720,14 +723,40 @@ AssignDistributedTransactionId(void)
...
+/*
+ * MarkCitusInitiatedCoordinatorBackend sets that coordinator backend is
+ * initiated by Citus.
+ */
+void
+MarkCitusInitiatedCoordinatorBackend(void)
+{
+	SpinLockAcquire(&MyBackendData->mutex);
+
+	MyBackendData->citusInitiatedBackend = true;
```
onderkalaci: Minor: Shall we call `citusInitiatedBackend ` -> `citusInitiatedCoordinatorBackend `?

src/test/regress/specs/isolation_citus_dist_activity.spec
```diff
@@ -82,7 +82,5 @@ step "s3-view-worker"
 
 permutation "s1-begin" "s2-begin" "s3-begin" "s1-alter-table" "s2-view-dist" "s3-view-worker" "s2-rollback" "s1-commit" "s3-rollback"
 permutation "s1-begin" "s2-begin" "s3-begin" "s1-insert" "s2-view-dist" "s3-view-worker" "s2-rollback" "s1-commit" "s3-rollback"
-permutation "s1-begin" "s2-begin" "s3-begin" "s1-select" "s2-view-dist" "s3-view-worker" "s2-rollback" "s1-commit" "s3-rollback" 
-
-# router selects don't show up because BEGIN is not sent for performance reasons
+permutation "s1-begin" "s2-begin" "s3-begin" "s1-select" "s2-view-dist" "s3-view-worker" "s2-rollback" "s1-commit" "s3-rollback"
 permutation "s1-begin" "s2-begin" "s3-begin" "s1-select-router" "s2-view-dist" "s3-view-worker" "s2-rollback" "s1-commit" "s3-rollback"
```
onderkalaci: I think it'd be great to have a similar `INSERT` test as well.

src/backend/distributed/transaction/citus_dist_stat_activity.c
```diff
@@ -535,18 +589,9 @@ ParseCitusDistStat(PGresult *result, int64 rowIndex)
...
-		citusDistStat->master_query_host_port = PostPortNumber;
-	}
 	else
 	{
-		/*
-		 * We could only get here if the function is called from metadata workers and
-		 * the query is initiated from the coordinator.
-		 */
-		citusDistStat->master_query_host_name = cstring_to_text(coordinator_host_name);
+		citusDistStat->master_query_host_name = NULL;
```
onderkalaci: We should update the comments (I've sent an example/suggestion in the hangout chat when we talked about this change)

PR #2421, opened 2018-10-01
Add failure tests for VACUUM/ANALYZE

jasonmp85:
VACUUM runs outside of a transaction, so the failure modes for it are somewhat straightforward, though ANALYZE runs in a 1pc transaction and multi-table VACUUM can fail between statements (PG 11 and higher).

### DIFF
--- a/src/test/regress/sql/failure_vacuum.sql
+++ b/src/test/regress/sql/failure_vacuum.sql
@@ -1,0 +1,43 @@
+-- print whether we're using version > 10 to make version-specific tests clear
+SHOW server_version \gset
+SELECT substring(:'server_version', '\d+')::int > 10 AS version_above_ten;
+
+SELECT citus.mitmproxy('conn.allow()');
+
+SET citus.shard_count = 1;
+SET citus.shard_replication_factor = 2; -- one shard per worker
+SET citus.multi_shard_commit_protocol TO '1pc';
+
+CREATE TABLE vacuum_test (key int, value int);
+SELECT create_distributed_table('vacuum_test', 'key');
+
+SELECT citus.clear_network_traffic();
+
+SELECT citus.mitmproxy('conn.onQuery(query="^VACUUM").kill()');
+VACUUM vacuum_test;
+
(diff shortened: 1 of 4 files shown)


src/test/regress/sql/failure_vacuum.sql
```diff
@@ -0,0 +1,43 @@
...
+SELECT citus.mitmproxy('conn.onQuery(query="^VACUUM").kill()');
+VACUUM vacuum_test;
+
+SELECT citus.mitmproxy('conn.onQuery(query="^ANALYZE").kill()');
+ANALYZE vacuum_test;
+
+SELECT citus.mitmproxy('conn.onQuery(query="^COMMIT").kill()');
+ANALYZE vacuum_test;
+
+-- TODO: This feels unnecessary; why are failed ANALYZE transactions critical?
```
thanodnl: interesting find, would it be worth to look into this in a future? In that case an issue on github might be more visible then a TODO.

src/test/regress/sql/failure_vacuum.sql
```diff
@@ -0,0 +1,43 @@
...
+SELECT create_distributed_table('other_vacuum_test', 'key');
+
+SELECT citus.mitmproxy('conn.onQuery(query="^VACUUM.*other").kill()');
+
+VACUUM vacuum_test, other_vacuum_test;
+
+-- ==== Clean up, we're done here ====
+
+SELECT citus.mitmproxy('conn.allow()');
+DROP TABLE vacuum_test;
```
thanodnl: Missing `DROP TABLE other_vacuum_test;`

src/test/regress/sql/failure_vacuum.sql
jasonmp85: Agreed; I'll open an issue.

src/test/regress/sql/failure_vacuum.sql
jasonmp85: Good catch!

PR #2423, opened 2018-10-02
Allow simple DML commands from hot standby

### LINKED ISSUE #1583: Enable DML transactions from secondary coordinator
For the purpose of availability, it would be useful if secondary coordinators (hot standby) could perform DML operations, such that when the primary coordinator fails, clients can continue to write to the distributed table via the secondary.

Currently, most write operations error out in the postgres executor, e.g.:
```
ERROR:  cannot execute INSERT in a read-only transaction
```

To enable write operations, we may need to bring back the executor hooks to avoid the condition in `ExecutorStart` which triggers this error.

Distribute deadlock detection currently uses group ID to identify which node started a distributed transaction. The database on the secondary is identical, so the deadlock detector would not be able to distinguish which distributed transactions were started by the primary and which were started by the secondaries. We should instead use a unique node ID that is assigned to a GUC using `ALTER SYSTEM`. 

To support 2PC recovery, the secondary would have to write transaction recovery records to one of the workers instead of its own `pg_dist_transaction` table, which is not writable.

Replication factor >1 would not be supported since the secondary cannot update `shardstate` in `pg_dist_placement`.

marcocitus:
DESCRIPTION: Add option to allow simple DML commands from hot standby

For availability purposes, it is useful to sometimes switch application traffic to a hot standby of the coordinator. However, a standby coordinator currently errors out on writes, even if nothing is written to the local database.

This PR adds a `citus.writable_standby_coordinator` setting that enables simple DML transactions (not requiring 2PC) from a standby coordinator. In the future, we may support 2PC transactions by writing the 2PC records on a worker and delegating recovery to the worker.

Any attempt to write to the database (e.g. when using a sequence, or 2PC) will result in `ERROR:  cannot assign TransactionIds during recovery`. 

There is no locking for rebalancer purposes. The intention is to only temporarily switch traffic over to a standby.

Fixes #1583


src/backend/distributed/executor/multi_executor.c
```diff
@@ -45,12 +45,114 @@
...
+{
+	PlannedStmt *plannedStmt = queryDesc->plannedstmt;
+
+	if (RecoveryInProgress() && WritableStandbyCoordinator &&
+		IsCitusPlan(plannedStmt->planTree))
+	{
+		PG_TRY();
+		{
+			/* pretend that we're not in read-only mode to allow DML through */
+			XactReadOnly = false;
```
marcocitus: d'oh, this is of those variables that we can't change on Windows.

marcocitus: That'll be hard to fix without re-implementing `standard_ExecutorStart`. For now I'm disabling windows support.

src/backend/distributed/executor/multi_executor.c
```diff
@@ -45,12 +45,124 @@
...
+	{
+		PG_TRY();
+		{
+			/*
+			 * To enable writes from a hot standby we cheat our way through
+			 * the checks in standard_ExecutorStart by temporarily setting
+			 * XactReadOnly to false.
+			 */
+			XactReadOnly = false;
+			standard_ExecutorStart(queryDesc, eflags | EXEC_FLAG_SKIP_TRIGGERS);
```
pykello: What is `EXEC_FLAG_SKIP_TRIGGERS` for? Maybe add a comment.

src/test/regress/sql/multi_follower_dml.sql
```diff
@@ -0,0 +1,57 @@
...
+TRUNCATE the_table;
+ALTER TABLE the_table ADD COLUMN c int;
+
+-- rollback is possible
+BEGIN;
+INSERT INTO the_table (a, b, z) VALUES (1, 2, 2);
+ROLLBACK;
+
+SELECT * FROM the_table;
+
```
pykello: Maybe also add a test to verify that we cannot write into regular postgres tables.

pykello:
I added two suggestions. Everything else looks fine to me.

src/backend/distributed/executor/multi_executor.c
marcocitus: It seemed logical to skip triggers, but there's actually no risk in allowing them and letting them fail if they try to write. It's kind of a weird esoteric case anyway. Removing the flag.

src/test/regress/sql/multi_follower_dml.sql
marcocitus: Added

marcocitus:
@pykello is this ok to :ship:?

pykello:
Sorry, my laptop died yesterday and I was trying to fix it (no luck yet)
before being able to review this one more time. I want to have another look
at this before shipping.

pykello:
I added two more tests, one of them will help us notice a regression if we support selecting from distributed table and inserting into a local table.

I also looked into related postgresql to understand how this change can fail. The following scenario is not working as expected yet:

```
BEGIN;
SET TRANSACTION READ ONLY;
INSERT INTO the_table (a, b, z) VALUES (1, 2, 2);
COMMIT;
```

Normally this fails, but under this patch it succeeds.

I'll spend some time to see how we can fix this. So not ship yet.

marcocitus:
That might not be easily fixable, since `SET TRANSACTION READ ONLY` sets `XactReadOnly` directly, which basically makes it a no-op in recovery mode. I think the long-term and only correct solution to this would be to propagate the `SET TRANSACTION READ ONLY` to the workers.

pykello:
Makes sense. Approving the PR.

PR #2425, opened 2018-10-04
Add failure tests for real time select queries

velioglu:
Adds failure tests for real-time select queries.

### DIFF
--- a/src/test/regress/expected/failure_real_time_select.out
+++ b/src/test/regress/expected/failure_real_time_select.out
@@ -1,0 +1,303 @@
+-- 
+--  Failure tests for real time select queries
+-- 
+CREATE SCHEMA real_time_select_failure;
+SET search_path TO 'real_time_select_failure';
+SET citus.next_shard_id TO 190000;
+-- Preparation
+SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SET citus.shard_count to 4;
+SET citus.shard_replication_factor to 1;
+-- create tables
+CREATE TABLE test_table(id int, value_1 int, value_2 int);
+SELECT create_distributed_table('test_table','id');
+ create_distributed_table 
+--------------------------
+ 
+(1 row)
+
+-- Populate data to the table
+INSERT INTO test_table VALUES(1,1,1),(1,2,2),(2,1,1),(2,2,2),(3,1,1),(3,2,2);
+-- Create a function to make sure that queries returning the same result 
+CREATE FUNCTION raise_failed_execution(query text) RETURNS void AS $$
+BEGIN
+	EXECUTE query;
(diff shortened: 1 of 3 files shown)


src/test/regress/expected/failure_real_time_select.out
```diff
@@ -0,0 +1,198 @@
...
+BEGIN;
+SELECT count(*) FROM test_table;
+WARNING:  could not consume data from worker node
+ count 
+-------
+     6
+(1 row)
+
+COMMIT;
+SELECT citus.dump_network_traffic();
```
onderkalaci: do we need the network dump?  I find those too verbose for adding to the regression test. And, even if you want to add, maybe move to the top of the file as @lithp did in other tests?

src/test/regress/expected/failure_real_time_select.out
```diff
@@ -0,0 +1,198 @@
...
+-- Kill when the first COPY command arrived, since we have a single placement 
+-- it is expected to error out.
+SELECT citus.mitmproxy('conn.onQuery(query="^COPY").kill()');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SET client_min_messages TO ERROR;
+DO $$
```
onderkalaci: Can't we make this a SQL function and pass a query to it? We're repeating it many times. And we'll keep copy & pasting it whenever we need it. 
```SQL
CREATE FUNCTION discard_task_id_from_error(query text)
....

```

src/test/regress/sql/failure_real_time_select.sql
```diff
@@ -0,0 +1,102 @@
...
+	EXCEPTION WHEN OTHERS THEN
+		IF SQLERRM LIKE 'failed to execute task%' THEN
+			RAISE 'Task failed to execute';
+		END IF;
+END;
+$$;
+
+-- Since the outer query uses the connection opened by the CTE,
+-- killing connection after first successful query should break. 
+SELECT citus.mitmproxy('conn.onQuery(query="^COPY").after(1).kill()');
```
onderkalaci: One more test suggestion: Maybe the same test without `after(1)` because we don't have any real-time selects within ctes that is failed in the first step?

src/test/regress/expected/failure_real_time_select.out
```diff
@@ -0,0 +1,291 @@
...
+-- Cancel a real-time executor query - in sequential mode
+SELECT citus.mitmproxy('conn.onQuery(query="^COPY").cancel(' || pg_backend_pid() || ')');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+SELECT count(*) FROM test_table;
+ERROR:  canceling statement due to user request
+-- Cancel a query within the transaction - in sequential mode
```
onderkalaci: One final test request: Can we add sequential cancel test with `after(1)`?

PR #2426, opened 2018-10-04
Add failure tests for insert/select via coordinator

mtuncer:
added failure and cancellation tests for insert/select via coordinator.
failure tests via insert/cte were previously done in failure_cte_subquery test.

### DIFF
--- a/src/test/regress/sql/failure_insert_select_via_coordinator.sql
+++ b/src/test/regress/sql/failure_insert_select_via_coordinator.sql
@@ -1,0 +1,97 @@
+--
+-- failure_insert_select_via_coordinator
+--
+-- performs failure/cancellation test for insert/select executed by coordinator.
+-- test for insert using CTEs are done in failure_cte_subquery, not repeating them here
+--
+
+CREATE SCHEMA coordinator_insert_select;
+SET SEARCH_PATH=coordinator_insert_select;
+SET citus.shard_count to 2;
+SET citus.shard_replication_factor to 1;
+SELECT pg_backend_pid() as pid \gset
+
+CREATE TABLE events_table(user_id int, event_id int, event_type int);
+CREATE TABLE events_summary(event_id int, event_type int, event_count int);
+CREATE TABLE events_reference(event_type int, event_count int);
+CREATE TABLE events_reference_distributed(event_type int, event_count int);
+SELECT create_distributed_table('events_table', 'user_id');
+SELECT create_distributed_table('events_summary', 'event_id');
+SELECT create_reference_table('events_reference');
+SELECT create_distributed_table('events_reference_distributed', 'event_type');
+
+INSERT INTO events_table VALUES (1, 1, 3 ), (1, 2, 1), (1, 3, 2), (2, 4, 3), (3, 5, 1), (4, 7, 1), (4, 1, 9), (4, 3, 2);
+
+SELECT count(*) FROM events_summary; 
+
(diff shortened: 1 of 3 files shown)


src/test/regress/sql/failure_insert_select_via_coordinator.sql
```diff
@@ -0,0 +1,97 @@
...
+SELECT citus.mitmproxy('conn.onQuery(query="^COPY").cancel(' || :pid || ')');
+INSERT INTO events_summary SELECT event_id, event_type, count(*) FROM events_table GROUP BY 1,2;
+
+-- cancel data push
+SELECT citus.mitmproxy('conn.onQuery(query="^COPY coordinator_insert_select").cancel(' || :pid || ')');
+INSERT INTO events_summary SELECT event_id, event_type, count(*) FROM events_table GROUP BY 1,2;
+
+--verify nothing is modified
+SELECT citus.mitmproxy('conn.allow()');
+SELECT count(*) FROM events_summary;
```
thanodnl: I would suggest to have this test after every interrupted transaction. It will give better guidance where a bug might have been introduced.

src/test/regress/sql/failure_insert_select_via_coordinator.sql
```diff
@@ -0,0 +1,97 @@
...
+SELECT citus.mitmproxy('conn.onQuery(query="^COPY").cancel(' || :pid || ')');
+INSERT INTO events_reference SELECT event_type, count(*) FROM events_table GROUP BY 1;
+
+-- cancel data push
+SELECT citus.mitmproxy('conn.onQuery(query="^COPY coordinator_insert_select").cancel(' || :pid || ')');
+INSERT INTO events_reference SELECT event_type, count(*) FROM events_table GROUP BY 1;
+
+--verify nothing is modified
+SELECT citus.mitmproxy('conn.allow()');
+SELECT count(*) FROM events_reference;
```
thanodnl: Same; I would suggest to have this test after every interrupted transaction. It will give better guidance where a bug might have been introduced.

src/test/regress/sql/failure_insert_select_via_coordinator.sql
```diff
@@ -0,0 +1,97 @@
...
+SELECT citus.mitmproxy('conn.onQuery(query="^COPY").cancel(' || :pid || ')');
+INSERT INTO events_reference_distributed SELECT * FROM events_reference;
+
+-- cancel data push
+SELECT citus.mitmproxy('conn.onQuery(query="^COPY coordinator_insert_select").cancel(' || :pid || ')');
+INSERT INTO events_reference_distributed SELECT * FROM events_reference;
+
+--verify nothing is modified
+SELECT citus.mitmproxy('conn.allow()');
+SELECT count(*) FROM events_reference_distributed;
```
thanodnl: Same; I would suggest to have this test after every interrupted transaction. It will give better guidance where a bug might have been introduced.

src/test/regress/sql/failure_insert_select_via_coordinator.sql
mtuncer: I made this way because every `conn.allow()` call costs about `150ms`.  If something goes wrong we can come back here and check where it fails.

thanodnl: ah cool, makes sense

PR #2427, opened 2018-10-04
Add reference table failure tests

jasonmp85:
Fairly straightforward; verified that modifications fail atomically if a worker is down or fails mid-transaction (i.e. all workers need to ack modifications to reference tables in order to persist changes).


src/test/regress/sql/failure_ref_tables.sql
```diff
@@ -0,0 +1,45 @@
...
+SELECT COUNT(*) FROM ref_table WHERE key=5;
+
+-- verify behavior of UPDATE ... RETURNING; should not execute
+SELECT citus.mitmproxy('conn.onQuery(query="^UPDATE").kill()');
+UPDATE ref_table SET key=7 RETURNING value;
+
+SELECT COUNT(*) FROM ref_table WHERE key=7;
+
+-- verify fix to #2214; should raise error and fail to execute
+SELECT citus.mitmproxy('conn.allow()');
```
thanodnl: This statement is not needed here I believe.

src/test/regress/failure_schedule
```diff
@@ -5,17 +5,17 @@ test: failure_test_helpers
...
+# test: failure_add_disable_node
+# test: failure_copy_to_reference
+# test: failure_copy_on_hash
+# test: failure_create_reference_table
+# test: failure_create_distributed_table_non_empty
+# test: failure_create_table
+#
+# test: failure_1pc_copy_hash
+# test: failure_1pc_copy_append
+test: failure_single_select
```
thanodnl: This hunk confuses me a lot,
 - we should not comment out all existing tests
 - not remove existing tests
 - I would expect this to be reduced to only the addition of `failure_ref_tables`.

src/test/regress/sql/failure_ref_tables.sql
```diff
@@ -0,0 +1,45 @@
...
+
+SELECT COUNT(*) FROM ref_table WHERE key=5;
+
+-- verify behavior of UPDATE ... RETURNING; should not execute
+SELECT citus.mitmproxy('conn.onQuery(query="^UPDATE").kill()');
+UPDATE ref_table SET key=7 RETURNING value;
+
+SELECT COUNT(*) FROM ref_table WHERE key=7;
+
+-- verify fix to #2214; should raise error and fail to execute
```
thanodnl: The bug describes a shard ending in shardstate 3, would it make sense to verify that doesn't happen.

src/test/regress/failure_schedule
jasonmp85: I guess by now you've noticed my hygiene around committing or reverting changes to this file is very poor :-P

I comment out other tests when I'm working because I want to avoid the extra time running them, then often forget to revert or commit my new test. Sorry!

src/test/regress/sql/failure_ref_tables.sql
jasonmp85: Sure; I still am not 100% sure where it is and isn't needed, but I'll remove it.

PR #2431, opened 2018-10-09
Add failure test for insert/select pushdown

mtuncer:


mtuncer:
@thanodnl ping

mtuncer:
@thanodnl you are correct, code paths are the same. It should not make a difference. I just could not come up with a different test case and added self insert/select

PR #2433, opened 2018-10-10
Keep track of cached entries in case of interruption.

pykello:
DESCRIPTION: Fix inconsistent metadata error when shard metadata caching gets interrupted.

Previously we set DistTableCacheEntry->sortedShardIntervalArray
and DistTableCacheEntry->shardIntervalArrayLength after we entered
all related shard entries into DistShardCacheHash. The drawback was
that if populating DistShardCacheHash was interrupted,
ResetDistTableCacheEntry() didn't see the shard hash entries created,
so was unable to clean them up.

This patch fixes that by setting sortedShardIntervalArray earlier,
and incrementing shardIntervalArrayLength as we enter shards into
the cache.

We also don't allow interrupts while making the distributed table shard
entries, so setting `statement_timeout` doesn't result in endless query
cancellation.

### DIFF
--- a/src/backend/distributed/utils/metadata_cache.c
+++ b/src/backend/distributed/utils/metadata_cache.c
@@ -860,11 +860,21 @@ LookupDistTableCacheEntry(Oid relationId)
 	memset(((char *) cacheEntry) + sizeof(Oid), 0,
 		   sizeof(DistTableCacheEntry) - sizeof(Oid));
 
+	/*
+	 * We disable interrupts while creating the cache entry because loading
+	 * shard metadata can take a while, and if statement_timeout is too low,
+	 * this will get canceled on each call and we won't be able to run any
+	 * queries on the table.
+	 */
+	HOLD_INTERRUPTS();
+
 	/* actually fill out entry */
 	BuildDistTableCacheEntry(cacheEntry);
 
 	/* and finally mark as valid */
 	cacheEntry->isValid = true;
+
+	RESUME_INTERRUPTS();
 
 	return cacheEntry;
 }
@@ -1170,6 +1180,13 @@ BuildCachedShardList(DistTableCacheEntry *cacheEntry)
 		}
 	}
 
+	/*
+	 * We set these here, so ResetDistTableCacheEntry() can see what has been
+	 * entered into DistShardCacheHash even if the following loop is interrupted
+	 * by throwing errors, etc.
+	 */
+	cacheEntry->sortedShardIntervalArray = sortedShardIntervalArray;
+	cacheEntry->shardIntervalArrayLength = 0;
 
 	/* maintain shardId->(table,ShardInterval) cache */
 	for (shardIndex = 0; shardIndex < shardIntervalArrayLength; shardIndex++)
@@ -1186,6 +1203,7 @@ BuildCachedShardList(DistTableCacheEntry *cacheEntry)
 
 		shardEntry = hash_search(DistShardCacheHash, &shardInterval->shardId, HASH_ENTER,
 								 &foundInCache);
+		cacheEntry->shardIntervalArrayLength++;
 		if (foundInCache)
 		{
 			ereport(ERROR, (errmsg("cached metadata for shard " UINT64_FORMAT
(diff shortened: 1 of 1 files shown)


src/backend/distributed/utils/metadata_cache.c
```diff
@@ -1184,6 +1190,11 @@ BuildCachedShardList(DistTableCacheEntry *cacheEntry)
 		int placementOffset = 0;
 		int numberOfPlacements = 0;
 
+		/*
+		 * Disable interrupts while making changes for this shard to ensure
+		 * the cache stays in a consistent state.
+		 */
+		HOLD_INTERRUPTS();
```
marcocitus: can we actually hit a CHECK_FOR_INTERRUPTS in this code block?

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -1217,13 +1228,15 @@ BuildCachedShardList(DistTableCacheEntry *cacheEntry)
 
 		cacheEntry->arrayOfPlacementArrays[shardIndex] = placementArray;
 		cacheEntry->arrayOfPlacementArrayLengths[shardIndex] = numberOfPlacements;
+		cacheEntry->shardIntervalArrayLength++;
 
 		/* store the shard index in the ShardInterval */
 		shardInterval->shardIndex = shardIndex;
+		RESUME_INTERRUPTS();
+
+		CHECK_FOR_INTERRUPTS();
```
marcocitus: I'm wondering whether we should just put the HOLD_INTERRUPTS around the whole loop

src/backend/distributed/utils/metadata_cache.c
pykello: This took a while to pinpoint. I had trouble using GDB when interrupts happen, and using `elog()` itself causes additional `CHECK_FOR_INTERRUPTS()`, so I just used [`fprintf()` + `backtrace()`](https://www.gnu.org/software/libc/manual/html_node/Backtraces.html) to find this call path.

`BuildShardPlacementList()` -> `systable_getnext()` -> `index_getnext()` -> `index_getnext_tid()` -> `btgettuple()` -> `_bt_next()` -> `_bt_steppage()` -> `_bt_readnextpage()` -> `CHECK_FOR_INTERRUPTS()`.

Anywhere heap access or index access happens, `CHECK_FOR_INTERRUPTS()` is expected.

src/backend/distributed/utils/metadata_cache.c
marcocitus: I think the code restructuring makes sense regardless. We may still have other errors that mess up the cache entry (e.g. out of memory), though we're probably not completely protected against those yet if e.g. what if an error occurs in `BuildShardPlacementList`? Should we increment the `shardIntervalArrayLength` immediately after adding an item to the hash such that we always remove the shard from the hash later, even if an error occurred immediately after adding it?

pykello: Got it. Thanks for the explanation.

Looking at the code, I think we should either move `HOLD_INTERRUPTS` to `LookupDistTableCacheEntry()`, or `CHECK_FOR_INTERRUPTS()` in each iteration of this loop.

If we just do this loop in a `HOLD_INTERRUPTS`, the cancellation can happen much later if shard count is high, and we aren't actually creating anything that we will reuse later, since `DistTableCacheEntry->isValid` will stay false and we will rebuild the cache data again later.

Because if the cancellation is happening because of `statement_timeout`, I'd vote for moving `HOLD_INTERRUPTS` to `LookupDistTableCacheEntry()` so we don't get timeouts over and over again.

What do you think?

marcocitus:
It seems the multi_cache_invalidation regression test crashes.

pykello:
Looking into it ...

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -2950,7 +2956,10 @@ ResetDistTableCacheEntry(DistTableCacheEntry *cacheEntry)
 		bool foundInCache = false;
 
 		/* delete the shard's placements */
-		pfree(placementArray);
+		if (placementArray != NULL)
+		{
+			pfree(placementArray);
```
pykello: This check is probably not needed, since std libraries `free()` just ignores `NULL` pointers, just wanted to make things explicit.

pykello:
Hey @metdos @sumedhpathak , this needs to be backported to versions that use metadata caching.

PR #2435, opened 2018-10-12
Prototype: Unified executor

marcocitus:
Still needs a lot of work, but getting to a stage where it can handle most DML, real-time and router queries with replication and failure handling.

The unified executor uses small pools of connections to be able to schedule multiple tasks on the same connection. The state machine is now architected around connection/transaction state instead of task state.


marcocitus:
I pushed the remaining changes I had on my laptop. It needs a rebase, but there is a conflict because this PR introduces `MultiConnectionState`, but it was also introduced by #2584.

marcocitus:
Superseded by #2798

PR #2437, opened 2018-10-15
Add single-shard modification failure tests

jasonmp85:
I'm pretty sure a lot of this test functionality may be covered in some of our existing regression tests, but I've included them to ensure we put all failure-based tests under our new testing method for that kind of test.

Didn't include lower replication factor, as (for a single-shard mod.), it's indistinguishable from modifying a reference table. So these all test modifications which hit a single, replicated shard.

### DIFF
--- a/src/test/regress/sql/failure_single_mod.sql
+++ b/src/test/regress/sql/failure_single_mod.sql
@@ -1,0 +1,61 @@
+SELECT citus.mitmproxy('conn.allow()');
+SELECT citus.clear_network_traffic();
+
+SET citus.shard_count = 2;
+SET citus.shard_replication_factor = 2;
+
+CREATE TABLE mod_test (key int, value text);
+SELECT create_distributed_table('mod_test', 'key');
+
+-- verify behavior of single INSERT; should mark shard as failed
+SELECT citus.mitmproxy('conn.onQuery(query="^INSERT").kill()');
+INSERT INTO mod_test VALUES (2, 6);
+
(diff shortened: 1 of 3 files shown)


src/test/regress/sql/failure_single_mod.sql
```diff
@@ -0,0 +1,60 @@
...
+
+-- some clean up
+UPDATE pg_dist_shard_placement SET shardstate = 1
+WHERE shardid IN (
+  SELECT shardid FROM pg_dist_shard WHERE logicalrelid = 'mod_test'::regclass
+) AND shardstate = 3 RETURNING placementid;
+TRUNCATE mod_test;
+
+-- verify behavior of UPDATE ... RETURNING; should mark as failed
+INSERT INTO mod_test VALUES (2, 6);
```
thanodnl: This insert fails already and marks the shard on the mitm-ed machine as failed. causing this test to not verify the update behaviour

```suggestion
SELECT citus.mitmproxy('conn.allow()');
INSERT INTO mod_test VALUES (2, 6);
```

sumedhpathak:
@thanodnl Jason is out on leave for some time. Could you get this change merged in on his behalf? Thanks!

thanodnl:
@sumedhpathak I have addressed my feedback, due to github issues today tests are not running. Once tests pass I will squash the commits and merge them to master. Hope travis picks up tonight once the webhooks work again.

PR #2444, opened 2018-10-18
Attempt to address planner context crashes

jasonmp85:
Originally added in #2080. I believe what happened was we added this to 7.1 in order to see whether it would fix an issue in Cloud, and held off adding the change to `master` until we got a thumbs up from them, but didn't follow up.

Since then, the surrounding code has changed significantly, so I had to modify the patch a bit (I used `git cherry-pick` and `git mergetool`, but still needed some manual editing to get the right code flow. The tests pass, but I'm still not really willing to assert this definitively fixes the issue (reproduction was very hard).

> Both of these are a bit of a shot in the dark. In one case, we noticed a stack trace where a caller received a null pointer and attempted to dereference the memory context field (at 0x010). In the other, I saw that any error thrown from within AdjustParseTree could keep the stack from being cleaned up (presumably if we push we should always pop).

> Both stack traces were collected during times of high memory pressure and locally reproducing the problem locally or otherwise has been very tricky (i.e. it hasn't been reproduced reliably at all).

### DIFF
--- a/src/backend/distributed/planner/distributed_planner.c
+++ b/src/backend/distributed/planner/distributed_planner.c
@@ -157,6 +157,10 @@ distributed_planner(Query *parse, int cursorOptions, ParamListInfo boundParams)
 
 			result = CreateDistributedPlannedStmt(planId, result, originalQuery, parse,
 												  boundParams, plannerRestrictionContext);
+
+			setPartitionedTablesInherited = true;
+			AdjustPartitioningForDistributedPlanning(parse,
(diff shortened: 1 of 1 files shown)


sumedhpathak:
@marcocitus Since you originally reviewed #2080, could you also take a look at this one? (And merge if it looks good?) Thanks!

metdos:
I asked @mtuncer to backport this only to 8.0 for the time being. We should also add this to 7.x branches in the case of point releases.

PR #1162, opened 2017-01-23
Allow start_metadata_sync_to_node in a transaction block

marcocitus:
After the worker_transaction conversion to the connection API, we no longer need to prevent transaction blocks.

PR #2446, opened 2018-10-19
make PG11 builds mandatory again

mtuncer:
We made PG11 builds optional when we had an issue with mx isolation test that we could not solve back then.

This PR solves the issue with a workaround by running start_metadats_sync_to_node outside the transaction block.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,8 +21,6 @@ matrix:
     - env: PGVERSION=9.6
     - env: PGVERSION=10
     - env: PGVERSION=11
-  allow_failures:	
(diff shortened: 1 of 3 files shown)


onderkalaci:
> 1- The reason we cannot run start_metadata_sync_to_node inside a transaction block is not commented at all. Is the reason rollback of such an operation? In case of a failure, we have to open a connection to worker and make metadata changes? This is the only thing pops-up in my mind. I would be glad if you can help me understand the reason.

See related PR:  https://github.com/citusdata/citus/pull/1162

PR #2450, opened 2018-10-23
Don't throw error for DROP DATABASE IF EXISTS

pykello:


pykello:
@metdos I was confused by this when doing release testing. This is a simple change. Please assign a reviewer.

src/test/regress/sql/multi_utility_statements.sql
```diff
@@ -146,3 +146,7 @@ FETCH ABSOLUTE 5 FROM noHoldCursor;
 FETCH BACKWARD noHoldCursor;
 COMMIT;
 FETCH ABSOLUTE 5 FROM noHoldCursor;
+
+-- Test we don't throw an error for DROP IF EXISTS
+DROP DATABASE IF EXISTS not_existing_database;
+DROP TABLE IF EXISTS not_existing_table;
```
pykello: We could add an exhaustive list of object types here, but I didn't want to make this PR too hard to review, as an exhaustive list would need the reviewer that we are not missing anything. So that'd be another PR if we need to.

src/backend/distributed/executor/multi_utility.c
```diff
@@ -549,11 +549,15 @@ multi_ProcessUtility(PlannedStmt *pstmt,
 	 */
 	if (IsA(parsetree, DropdbStmt))
 	{
+		const bool missingOK = true;
 		DropdbStmt *dropDbStatement = (DropdbStmt *) parsetree;
 		char *dbname = dropDbStatement->dbname;
-		Oid databaseOid = get_database_oid(dbname, false);
+		Oid databaseOid = get_database_oid(dbname, missingOK);
```
onderkalaci: Maybe rely on `dropDbStatement ->missing_ok` to avoid any edge cases where the user drops an existing database which doesn't have a valid oid (if that's ever possible 😄 ):

```suggestion
		Oid databaseOid = get_database_oid(dbname, dropDbStatement ->missing_ok);
```

pykello: I think `dbcommands.c` -> `dropdb()` should be the single point we throw errors for `DROP DATABASE`, otherwise debugging becomes difficult. Looking at `dbcommands.c`, it seems that all databases have a valid oid.

metdos:
@pykello @onderkalaci do we want to cherry-pick this to release-8.0?

onderkalaci:
> @pykello @onderkalaci do we want to cherry-pick this to release-8.0?

@metdos yes I'm getting the commit to both community and enterprise now.

metdos:
Needs a description, such as Fixes a bug which cause errors on DROP DATABASE IF EXISTS

PR #2452, opened 2018-10-23
Processes that are blocked on advisory locks show up in wait edges

### LINKED ISSUE #2451: citus_lock_waits does not show waiting locks on multi-shard update
Hit this during release testing.  citus_lock_waits failed to show waiting process id when multi-shard update command is waiting for another multi-shard update.

[PR description](https://github.com/citusdata/citus/pull/2381) states it could not catch single shard modifications outside of transaction block, however this is nothing like it.

Have two tables t1(a int, b int), r1(a int, b int). t1 is hash distributed r1 is references.

# Test 1

### Session 1:
```sql
begin;
update t1 set b=1 where b = 2;
```

### Session 2:
```sql
begin;
update t1 set b = 3 where b = 2;
--> blocked
```

### Session 3:
```sql
select * from citus_lock_waits;
--> shows nothing
```

# Test 2: repeat of test 1 on a reference table

### Session 1:
```sql
begin;
update r1 set b=1 where b = 2;
```

### Session 2:
```sql
begin;
update r1 set b = 3 where b = 2;
--> blocked
```

### Session 3:
```sql
select * from citus_lock_waits;
--> shows session 2 query blocked on session 1
```

onderkalaci:
(Doesn't need a description since it is a bugfix for #2381)

Fixes #2451

Assign the distributed transaction id before trying to acquire the
executor advisory locks. This is useful to show this backend in citus
lock graphs (e.g., dump_global_wait_edges() and citus_lock_waits).

### DIFF
--- a/src/backend/distributed/executor/multi_router_executor.c
+++ b/src/backend/distributed/executor/multi_router_executor.c
@@ -1352,6 +1352,13 @@ ExecuteModifyTasks(List *taskList, bool expectResults, ParamListInfo paramListIn
 	}
 
 	/*
+	 * Assign the distributed transaction id before trying to acquire the
+	 * executor advisory locks. This is useful to show this backend in citus
+	 * lock graphs (e.g., dump_global_wait_edges() and citus_lock_waits).
+	 */
+	BeginOrContinueCoordinatedTransaction();
+
+	/*
 	 * Ensure that there are no concurrent modifications on the same
 	 * shards. In general, for DDL commands, we already obtained the
 	 * appropriate locks in ProcessUtility. However, we still prefer to
@@ -1361,8 +1368,6 @@ ExecuteModifyTasks(List *taskList, bool expectResults, ParamListInfo paramListIn
 	 * tables.
 	 */
 	AcquireExecutorMultiShardLocks(taskList);
-
-	BeginOrContinueCoordinatedTransaction();
 
 	if (MultiShardCommitProtocol == COMMIT_PROTOCOL_2PC ||
 		firstTask->replicationModel == REPLICATION_MODEL_2PC)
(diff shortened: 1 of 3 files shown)


mtuncer:
Verified that PR fixes the original problem

mtuncer:
@onderkalaci could you cherry-pick this into `release-8.0` and related enterprise branches.

onderkalaci:
> verified the fix

Thanks for the review!

> I am seeing a weird behavior here

Isn't the behaviour the same for `pg_stat_activity` or [here](https://wiki.postgresql.org/wiki/Lock_Monitoring#.D0.A1ombination_of_blocked_and_blocking_activity)? We're simply joining `pt_stat_activity` with `dump_global_wait_edges` on `process_id`.

I don't think there anything we can do for that.

> I have following output when I run select * from citus_lock_waits in the transaction that acquired the lock.

I think in general people use these kind of views from separate connections. But, its good that we documented the behaviour here.

> @onderkalaci could you cherry-pick this into release-8.0 and related enterprise branches.

I'll wait @marcocitus for some time before merging it in case he wants to have a look. But, after that I'll do that.

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -1351,6 +1351,13 @@ ExecuteModifyTasks(List *taskList, bool expectResults, ParamListInfo paramListIn
 		LockPartitionRelations(firstShardInterval->relationId, RowExclusiveLock);
 	}
 
+	/*
+	 * Assign the distributed transaction id before trying to acquire the
+	 * executor advisory locks. This is useful to show this backend in citus
+	 * lock graphs (e.g., dump_global_wait_edges() and citus_lock_waits).
+	 */
+	BeginOrContinueCoordinatedTransaction();
```
marcocitus: Maybe a bigger change, but should we do this before taking metadata locks?

PR #2464, opened 2018-10-30
Use heap_deform_tuple() instead of calling heap_getattr().

### LINKED ISSUE #2453: Entries from pg_worker_list.conf aren't picked up in PostgreSQL 11.
When using postgres 11 and master branch of citus, entries from `pg_worker_list.conf` aren't picked up properly. I can see them in `pg_dist_node`, but `master_get_active_worker_nodes()` returns empty set. This doesn't happen when using postgres 10 and master branch of citus.

I used [this script](https://gist.github.com/pykello/ec9d391b5d61b6352d47ba6f8e93c5cc) for testing.

The output when using postgres 11 is:

```
Setting up data.9700 ...
+ Removing data dir
+ postgres version is: PostgreSQL 11.0
+ initdb
+ Setting config
+ Adding pg_worker_list.conf
+ Starting postgres
+ pg_dist_node:
 nodeid | groupid | nodename  | nodeport | noderack | hasmetadata | isactive | noderole | nodecluster 
--------+---------+-----------+----------+----------+-------------+----------+----------+-------------
      1 |       1 | localhost |     9701 | default  | f           | t        | primary  | default
      2 |       2 | localhost |     9702 | default  | f           | t        | primary  | default
(2 rows)

+ master_get_active_worker_nodes()
 node_name | node_port 
-----------+-----------
(0 rows)

                                            citus_version                                            
-----------------------------------------------------------------------------------------------------
 Citus 8.0devel on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 7.3.0-27ubuntu1~18.04) 7.3.0, 64-bit
(1 row)
```

when using postgres 10, the output is as expected:

```
Setting up data.9700 ...
+ Shutting down previously running postgres
+ Removing data dir
+ postgres version is: PostgreSQL 10.5
+ initdb
+ Setting config
+ Adding pg_worker_list.conf
+ Starting postgres
+ pg_dist_node:
 nodeid | groupid | nodename  | nodeport | noderack | hasmetadata | isactive | noderole | nodecluster 
--------+---------+-----------+----------+----------+-------------+----------+----------+-------------
      1 |       1 | localhost |     9701 | default  | f           | t        | primary  | default
      2 |       2 | localhost |     9702 | default  | f           | t        | primary  | default
(2 rows)

+ master_get_active_worker_nodes()
 node_name | node_port 
-----------+-----------
 localhost |      9701
 localhost |      9702
(2 rows)

                                            citus_version                                            
-----------------------------------------------------------------------------------------------------
 Citus 8.0devel on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 7.3.0-27ubuntu1~18.04) 7.3.0, 64-bit
(1 row)
```

This breaks some of our test scripts, so I think we should fix this. It is late here right now, so if no one finds the cause by Wednesday morning, I'll look into it.

pykello:
After [Fast ALTER TABLE ADD COLUMN with a non-NULL default](https://github.com/postgres/postgres/commit/16828d5c0273b4fe5f10f42588005f16b415b2d8) in PG11, physical heaps might not contain all attributes after a `ALTER TABLE ADD COLUMN` happens. `heap_getattr()` returns `NULL` when the physical tuple doesn't contain an attribute. So we should use `heap_deform_tuple()` in these cases, which fills in the missing attributes.

Our catalog tables evolve over time, and an upgrade might involve some  `ALTER TABLE ADD COLUMN` commands.

Note that we don't need to worry about postgres catalog tables and we can use `heap_getattr()` for them, because they only change between major versions.

This also fixes #2453.


src/backend/distributed/utils/node_metadata.c
```diff
@@ -1525,26 +1525,27 @@ TupleToWorkerNode(TupleDesc tupleDescriptor, HeapTuple heapTuple)
 	WorkerNode *workerNode = NULL;
 	bool isNull = false;
 
-	Datum nodeId = heap_getattr(heapTuple, Anum_pg_dist_node_nodeid,
+	HeapTuple expandedTuple = ExpandTupleIfNeeded(heapTuple, tupleDescriptor);
+	Datum nodeId = heap_getattr(expandedTuple, Anum_pg_dist_node_nodeid,
```
anarazel: I think this is the wrong fix. I think you rather should just deform the tuple and access the columns via that. That'll take care of the expansion *and* be faster.

pykello: Makes sense. Just for my understanding, is it just that tuple deformation is the recommended way, or are there any problems with this approach?

src/backend/distributed/master/master_metadata_utility.c
```diff
@@ -1090,6 +1088,19 @@ DeletePartitionRow(Oid distributedRelationId)
...
 
 
+/*
+ * PartitionTupleRelationId returns the logicalrelid column in the given
+ * pg_dist_partition heap tuple.
+ */
+Oid
+PartitionTupleRelationId(HeapTuple heapTuple)
+{
+	Form_pg_dist_partition data = (Form_pg_dist_partition) GETSTRUCT(heapTuple);
```
pykello: We don't need to use `heap_deform_tuple()` here since `logicalrelid` has been in this catalog table since beginning of time, and deforming the whole tuple because of one field seems wasteful.

pykello:
This is still in progress, I still need to look at how we use `GETSTRUCT()`.

src/backend/distributed/master/master_metadata_utility.c
marcocitus: Maybe add that in a comment. We should be very careful doing this from now on, because we can never be sure whether an ALTER EXTENSION was done pre-PG11 or in PG11.

src/backend/distributed/transaction/transaction_recovery.c
```diff
@@ -245,15 +245,17 @@ RecoverWorkerTransactions(WorkerNode *workerNode)
...
 		bool isTransactionInProgress = false;
 		bool foundPreparedTransactionBeforeCommit = false;
 		bool foundPreparedTransactionAfterCommit = false;
+		char *transactionName = NULL;
 
-		Datum transactionNameDatum = heap_getattr(heapTuple,
-												  Anum_pg_dist_transaction_gid,
-												  tupleDescriptor, &isNull);
-		char *transactionName = TextDatumGetCString(transactionNameDatum);
+		heap_deform_tuple(heapTuple, tupleDescriptor, datumArray, isNullArray);
```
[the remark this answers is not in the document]
onderkalaci: > SELECT attrelid::regclass, attname, atthasmissing, attmissingval FROM pg_attribute WHERE atthasmissing;


Another random thought I've: Do you think would adding the following query (and a comment like you've in your PR description) to `multi_extension.sql` help us to be cautious about such changes when we add a new column:

```SQL
SELECT attrelid::regclass, attname, atthasmissing, attmissingval FROM pg_attribute WHERE atthasmissing AND attrelid::regclass::text LIKE '%pg_dist%;

```

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -889,13 +889,13 @@ BuildDistTableCacheEntry(DistTableCacheEntry *cacheEntry)
...
 	HeapTuple distPartitionTuple = NULL;
 	Relation pgDistPartition = NULL;
-	Form_pg_dist_partition partitionForm = NULL;
 	Datum partitionKeyDatum = 0;
 	Datum replicationModelDatum = 0;
 	MemoryContext oldContext = NULL;
 	TupleDesc tupleDescriptor = NULL;
-	bool isNull = false;
 	bool partitionKeyIsNull = false;
+	Datum datumArray[Natts_pg_dist_partition];
```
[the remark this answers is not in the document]
pykello: I'll compare the performance before merging, but note that in cases we get all values for the tuple (like this case), reading their implementations, the performance of `heap_deform_tuple()` should be as good as `heap_getattr()` or better.

I'll do a benchmark before merge though.

src/backend/distributed/master/master_metadata_utility.c
pykello: Removed this because it wasn't very related to this patch. I'll do it as part of another refactoring PR.

src/backend/distributed/transaction/transaction_recovery.c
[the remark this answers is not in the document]
pykello: I think @onderkalaci 's suggestion that we do `heap_deform_tuple()` only when decoding whole tuples, and using `heap_getattr()` when we are accessing a single attribute and we are sure it was there pre-PG11 makes sense.

Just please note that the fact the postgres uses `heap_getattr()` doesn't mean we can use it in extensions worry-free. Postgres catalog only changes between major versions, so they are guaranteed to be `ALTER COLUMN` free.

src/test/regress/sql/multi_extension.sql
```diff
@@ -165,6 +165,14 @@ WHERE pgd.refclassid = 'pg_extension'::regclass AND
 	  pge.extname    = 'citus' AND
 	  pgio.schema    NOT IN ('pg_catalog', 'citus', 'test');
 
+-- if the output of following query changes, we might need to change
```
onderkalaci: I think maintaining an extra output file for `multi_extension.sql` could be very painful because we need to update it very frequently, whenever we add/remove/change UDFs. 

What if we add a new test file, `multi_metadata_attributes.sql` just before `multi_extension.sql` and only includes the followin:

```SQL
CREATE EXTENSION citus;

 -- the query you've -- 

DROP EXTENSION citus;
```

so that no one / no test actually is affected by this on regular basis?

PR #2424, opened 2018-10-02
Make sure not to leak intermediate result folders on the workers

onderkalaci:
Fixes #2420 

To ensure that Citus never leaks any intermediate results (e.g., recursive planning, real-time or task-tracker), I added the test at the end of the regression test suite.

But it seems like we've some more edge cases in real-time (or task-tracker?) because the tests sometimes fail with the following output indicating that we've one folder leaking at the end of the regression tests:
```SQL
 SELECT run_command_on_workers($$SELECT pg_ls_dir('base/pgsql_job_cache')$$);
  run_command_on_workers 
-------------------------
+--------------------------------------
  (localhost,57637,t,"")
- (localhost,57638,t,"")
+ (localhost,57638,t,job_238202912770)
 (2 rows)
```

And, that seems to be caused by not handling `task-tracker` failures well, because if I execute `multi_task_string_size.sql` multiple times, I'd see one intermediate result per call.

So, I could overlook the task-tracker failures and only test recursive planning. 

In any case, I'm still investigating the task-tracker issue some more.

PR #2413, opened 2018-09-28
Fix memory leak in FinishRemoteTransactionPrepare

mtuncer:
We seem to be leaking some memory in remote transaction preparation. This works reclaims memory kept in PGresult structure.

We should do more extensive review and testing for similar usages in the codebase.

Fixes #2405

PR #2466, opened 2018-10-31
Add changelog entry for 8.0.0

velioglu:
Add changelog entry for 8.0.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,49 @@
+### citus v8.0.0 (October 31, 2018) ###
+
+* Adds support for PostgreSQL 11
+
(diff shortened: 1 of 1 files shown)


metdos:
I think we group MX related items and use capitalized always (MX instead of mx)

metdos:
PG11 support is missing.

metdos:
Description is missing here: https://github.com/citusdata/citus/pull/2450

metdos:
Maybe also for this https://github.com/citusdata/citus/pull/2424?

metdos:
Maybe for this also https://github.com/citusdata/citus/pull/2413?

PR #2468, opened 2018-11-05
Expand who is using Citus section in readme

mtuncer:
- add new users (Pex, Algolia, Copper) to who is using Citus part
- add a reference to more use cases in Citus web page

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,18 +129,22 @@ technology start-ups to large enterprises. Here are some examples:
 
 * [MixRank](https://mixrank.com/) uses Citus to efficiently collect
 and analyze vast amounts of data to allow inside B2B sales teams
-to find new customers. [Case
-Study](https://www.citusdata.com/solutions/case-studies/mixrank-case-study)
+to find new customers. [User Story](https://www.citusdata.com/solutions/case-studies/mixrank-case-study)
 * [Neustar](https://www.neustar.biz/) builds and maintains scalable
 ad-tech infrastructure that counts billions of events per day using
 Citus and HyperLogLog.
 * [Agari](https://www.agari.com/) uses Citus to secure more than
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -141,6 +141,12 @@ Study](https://www.citusdata.com/solutions/case-studies/agari-case-study)
 funnel, segmentation, and cohort queries across billions of users
 and tens of billions of events. [Watch
 Video](https://www.youtube.com/watch?v=NVl9_6J1G60&list=PLixnExCn6lRpP10ZlpJwx6AuU3XIgNWpL)
+* [Pex](https://pex.com/) uses Citus to implement very high volume data creation and search database for audio/video data on a 20+ node cluster on Google Cloud [Case Study](https://www.citusdata.com/customers/pex)
```
ozgune: Two small edit proposals. Please evaluate and incorporate as you see fit.

1. "Pex uses Citus to ingest 30B data points per day and analyze that data in real-time. They use a 20+ node cluster on Google Cloud."
2. "Algolia uses Citus to provide real-time analytics for over 1B searches per day. For faster insights, they also use TopN and HLL extensions."

tgcmn1: This is my first ever github comment. :-) Here are a few small things:

1. please change most, but not all, references to "Case Study" to "Customer Story"
2. at the bottom, please change to: "You can read more customer stories about how users employ Citus to scale Postgres for both multi-tenant SaaS applications as well as real-time analytics dashboards.
3. I've sent a note off to Rasty from Pex to confirm the size of his cluster. I think he upgraded since we wrote the story.

tgcmn1: Rasty responds quickly. Pex is still using a 20-node cluster.

mtuncer: @tgcmn1 Can we not use "Customer" here. "User" seems to be a better option

PR #2469, opened 2018-11-06
Refactor the UtilityProcess function into its own module

thanodnl:
Description: Refactor code that handles DDL commands from one file into a module

The file handling the utility functions (DDL) for citus organically grew over time and became unreasonably large. This refactor takes that file and refactored the functionality into separate files per command. Initially modeled after the directory and file layout that can be found in postgres.

Although the size of the PR is quite big there are barely any code changes. Only one two functions have been added for readability purposes:
 - `PostProcessIndexStmt` which is extracted from `PostProcessUtility`
 - `PostProcessAlterTableStmt` which is extracted from `multi_ProcessUtility`

Extra attention can be given to those functions to make sure no changes have been introduced.

I have added a `README.md` to `src/backend/distributed` describing the contents of the module and every file in the module. I think we need more documentation around the overloading of the `COPY` command. I have added the boilerplate for others with better knowledge to fill in.


src/backend/distributed/commands/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+| File                         | Description |
+|------------------------------|-------------|
+| `create_distributed_table.c` | Implementation of UDF's for creating distributed tables |
+| `drop_distributed_table.c`   | Implementation for dropping metadata for partitions of distributed tables |
+| `extension.c`                | Implementation of `CREATE EXTENSION` commands for citus specific checks |
+| `foreign_constraint.c`       | Implementation of and helper functions for foreign key constraints |
+| `grant.c`                    | Placeholder for code granting users access to relations, implemented as enterprise feature |
+| `indexcmds.c`                | Implementation of commands specific to indices on distributed tables |
```
mtuncer: some files have postfix `cmds` some do not. Shall we have a standard on this ?

src/backend/distributed/commands/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+The copy command is overloaded for a couple of use-cases specific to citus. The syntax of the command stays the same, however the implementation might slightly differ from the stock implementation. The overloading is mostly done via extra options that Citus uses to indicate how to operate the copy. The options used are described below.
+
+## FORMAT transmit
+
+Implemented in `transmit.c`
+
+TODO: to be written by someone with enough knowledge to write how, when and why it is used.
+
+## FROMAT result
```
mtuncer: s/FROMAT/FORMAT

src/include/distributed/version_compat.h
```diff
@@ -12,7 +12,7 @@
 #define VERSION_COMPAT_H
 
 #include "postgres.h"
-#include "commands/explain.h"
+#include <commands/explain.h>
```
mtuncer: looks odd, can we go back to quotation notation ?

src/include/distributed/commands/extension.h
```diff
@@ -0,0 +1,22 @@
+/*-------------------------------------------------------------------------
+ *
+ * extension.c
```
mtuncer: extension.h

src/include/distributed/citus_ruleutils.h
```diff
@@ -17,7 +17,7 @@
 #if (PG_VERSION_NUM >= 100000)
 #include "catalog/pg_sequence.h"
 #endif
-#include "commands/sequence.h"
+#include <commands/sequence.h>
```
mtuncer: switch to quoatation

src/include/distributed/commands/grant.h
```diff
@@ -0,0 +1,19 @@
+/*-------------------------------------------------------------------------
+ *
+ * grant.c
```
mtuncer: grant.h

mtuncer: is it better to group these tiny header files together ?  We literally have a single function declared in many of those.

src/include/distributed/commands/utility_hook.h
```diff
@@ -10,10 +10,18 @@
...
+#include "postgres.h"
+
+#include "utils/relcache.h"
 #include "tcop/utility.h"
 
 extern bool EnableDDLPropagation;
-extern bool EnableVersionChecks;
+
+
+#if (PG_VERSION_NUM < 100000)
```
mtuncer: could this be moved to version compat ?

mtuncer:
@thanodnl I am guessing there would be additional work on enterprise side after we merge this. right ?

src/include/distributed/version_compat.h
thanodnl: TL;DR: Unfortunately we can't. There is a difference in the search path for quotes and the bracket signs.

We need to include the `commands/explain.h` from postgres explicitly here. By using brackets we force the pre-compiler to only look in the include directories passed to the compiler. With quotes the first directory it searches for the include is the directory relative to this file (`src/include/distributed/`). Since there is a `commands/explain.h` in that directory it will include that file instead of postgres' header file.

src/include/distributed/commands/grant.h
thanodnl: That is a good question. I don't know if there are any real downsides of having multiple small header files vs a combined one.

Benefit of having them separated is that all functions related to this concept should be implemented here. It might be that I couldn't find all the functions just yet for granting permission to a table as I have only focused on the utilities file.

For other concepts you need to implement multiple scenarios. Creation before table distribution, creation after table distribution and new shard creation. I expect such functions to reside all together in this file in the end. At what point would it warrant to split them in their own header file?

src/backend/distributed/commands/README.md
thanodnl: Yeah, I started with making parallels to how the postgres repository is laid out. I already started to divert as most of what we do is related to tables but often would be better suited to be in their own files.

I will change the files with the `cmds` postfix to be different in a way that covers what can be expected inside.

src/backend/distributed/commands/extension.c
```diff
@@ -0,0 +1,117 @@
+/*-------------------------------------------------------------------------
+ *
+ * extension.c
```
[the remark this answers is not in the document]
thanodnl: well, as above I am planning on removing the `cmds` postfix :). I think the functions are all very much related to ddl's that upgrade the citus extension. So either `upgrade_extension` or the other way around.

src/backend/distributed/commands/extension.c
```diff
@@ -0,0 +1,117 @@
...
+	if (IsA(parsetree, CreateExtensionStmt))
+	{
+		extensionName = ((CreateExtensionStmt *) parsetree)->extname;
+	}
+	else if (IsA(parsetree, AlterExtensionStmt))
+	{
+		extensionName = ((AlterExtensionStmt *) parsetree)->extname;
+	}
+
+	return (strcmp(extensionName, "citus") == 0);
```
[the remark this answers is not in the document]
thanodnl: I think it makes more sense on its own after code reorganisation. We need to find all the places where the hard coded string has the same meaning as here and extract it to the same macro. Only doing it for this occurrence makes this code less readable.

grepping the source I see it mentioned at more places, but never in a macro define:

```
$ git grep '"citus"'
src/backend/distributed/connection/connection_configuration.c:  AddConnParam("fallback_application_name", "citus");
src/backend/distributed/executor/multi_utility.c:   return (strcmp(extensionName, "citus") == 0);
src/backend/distributed/install.pl:my $EXTENSION = "citus";
src/backend/distributed/master/master_metadata_utility.c:   citusExtensionAddr.objectId = get_extension_oid("citus", false);
src/backend/distributed/shared_library_init.c:  EmitWarningsOnPlaceholders("citus");
src/backend/distributed/utils/maintenanced.c:       sprintf(worker.bgw_library_name, "citus");
src/backend/distributed/utils/maintenanced.c:   Oid extensionOid = get_extension_oid("citus", true);
src/backend/distributed/utils/maintenanced.c:   recheckExtensionOid = get_extension_oid("citus", true);
src/backend/distributed/utils/metadata_cache.c:     Oid extensionOid = get_extension_oid("citus", true);
src/backend/distributed/utils/metadata_cache.c:     if (strcmp(extensionName, "citus") == 0)
src/backend/distributed/utils/metadata_cache.c:             CStringGetDatum("citus"));
src/backend/distributed/utils/metadata_cache.c:     Oid citusExtensionOid = get_extension_oid("citus", false);
src/backend/distributed/utils/metadata_cache.c:             CStringGetDatum("citus"));
src/backend/distributed/worker/task_tracker.c:  snprintf(worker.bgw_library_name, BGW_MAXLEN, "citus");
src/test/regress/pg_regress_multi.pl:my $sharedPreloadLibraries = "citus";
```

src/backend/distributed/commands/extension.c
```diff
@@ -0,0 +1,117 @@
...
+	else
+	{
+		/* input must be one of the two above types */
+		Assert(false);
+	}
+
+	foreach(optionsCell, optionsList)
+	{
+		DefElem *defElement = (DefElem *) lfirst(optionsCell);
+		if (strncmp(defElement->defname, "new_version", NAMEDATALEN) == 0)
```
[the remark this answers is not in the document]
thanodnl: I can try and figure out what it is doing. I didn't pay much attention to functionality of code during the refactor. Only where it was required to figure out in which file to place the code.

thanodnl:
Yes, once we settle on the details it requires some work with enterprise as well. I haven't tried merging it in there just yet.

Once that all works I think we should move the implementation for grant in enterprise and stub the functions out here in the same way we do with policy.

src/include/distributed/version_compat.h
mtuncer: can we rename the file so that there won't be a conflict ?

src/test/regress/.gitignore
```diff
@@ -9,3 +9,4 @@
 # Regression test output
 /regression.diffs
 /regression.out
+/proxy.output
```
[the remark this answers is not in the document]
thanodnl: This file keeps lingering untracked after you have run a failure test locally.

Its irrelevant to the refactor I will remove it here and open a separate pr with some other small developer productivity things.

src/include/distributed/version_compat.h
```diff
@@ -15,6 +15,7 @@
 #include "commands/explain.h"
 #include "catalog/namespace.h"
 #include "nodes/parsenodes.h"
+#include "postmaster/bgworker.h"
```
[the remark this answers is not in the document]
thanodnl: it is required for builds on 9.6 iirc. a macro below in this depends on this file being included. It became a compilation problem now that utility hook is including version_compat for the querystruct.

I will move this to the ifdef where the macro is overloaded. There is a small include block there as well.

src/backend/distributed/commands/table.c
```diff
@@ -0,0 +1,1380 @@
+/*-------------------------------------------------------------------------
+ *
+ * table.c
+ *    Commands for altering and creating distributed tables.
```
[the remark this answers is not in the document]
thanodnl: makes sense, I think I have this pattern all over the place, will revisit before landing

PR #2471, opened 2018-11-07
Update topn and hll versions used in travis

mtuncer:
- changed hll  version 2.12 from 2.10.2
- changed topn version to 2.2.0 from 2.1.0
- both topn and hll were not installed for PG11, now they are installed in travis builds


furkansahin:
This makes sense to me, :shipit: after the tests pass.

PR #2472, opened 2018-11-08
Round robin queries to reference tables with task_assignment_policy set to `round-robin`

thanodnl:
Description: Support round-robin `task_assignment_policy` for queries to reference tables.

This PR allows users to query multiple placements of shards in a round robin fashion. When `citus.task_assignment_policy` is set to `'round-robin'` the planner will use a round robin scheduling feature when multiple shard placements are available.

The primary use-case is spreading the load of reference table queries to all the nodes in the cluster instead of hammering only the first placement of the reference table. Since reference tables share the same path for selecting the shards with single shard queries that have multiple placements (`citus.shard_replication_factor > 1`) this setting also allows users to spread the query load on these shards.

For modifying queries we do not apply a round-robin strategy. This would be negated by an extra reordering step in the executor for such queries where a `first-replica` strategy is enforced.

### DIFF
--- a/src/backend/distributed/planner/multi_physical_planner.c
+++ b/src/backend/distributed/planner/multi_physical_planner.c
@@ -178,7 +178,6 @@ static List * ReorderAndAssignTaskList(List *taskList,
 static int CompareTasksByShardId(const void *leftElement, const void *rightElement);
 static List * ActiveShardPlacementLists(List *taskList);
 static List * ActivePlacementList(List *placementList);
-static List * LeftRotateList(List *list, uint32 rotateCount);
 static List * FindDependedMergeTaskList(Task *sqlTask);
 static List * AssignDualHashTaskList(List *taskList);
 static int CompareTasksByTaskId(const void *leftElement, const void *rightElement);
@@ -5311,7 +5310,7 @@ ActivePlacementList(List *placementList)
  * repeatedly moves the list's first element to the end of the list, and
  * then returns the newly rotated list.
  */
-static List *
+List *
 LeftRotateList(List *list, uint32 rotateCount)
 {
 	List *rotatedList = list_copy(list);
--- a/src/backend/distributed/planner/multi_physical_planner.c
+++ b/src/backend/distributed/planner/multi_physical_planner.c
@@ -171,13 +171,13 @@ static bool HasMergeTaskDependencies(List *sqlTaskList);
 static List * GreedyAssignTaskList(List *taskList);
 static Task * GreedyAssignTask(WorkerNode *workerNode, List *taskList,
 							   List *activeShardPlacementLists);
-static List * RoundRobinAssignTaskList(List *taskList);
 static List * RoundRobinReorder(Task *task, List *placementList);
 static List * ReorderAndAssignTaskList(List *taskList,
 									   List * (*reorderFunction)(Task *, List *));
 static int CompareTasksByShardId(const void *leftElement, const void *rightElement);
 static List * ActiveShardPlacementLists(List *taskList);
 static List * ActivePlacementList(List *placementList);
+static List * LeftRotateList(List *list, uint32 rotateCount);
 static List * FindDependedMergeTaskList(Task *sqlTask);
 static List * AssignDualHashTaskList(List *taskList);
 static int CompareTasksByTaskId(const void *leftElement, const void *rightElement);
@@ -5097,7 +5097,7 @@ FirstReplicaAssignTaskList(List *taskList)
  * by the number of active shard placements, and ensure that we rotate between
  * these placements across subsequent queries.
  */
-static List *
+List *
 RoundRobinAssignTaskList(List *taskList)
 {
 	taskList = ReorderAndAssignTaskList(taskList, RoundRobinReorder);
@@ -5310,7 +5310,7 @@ ActivePlacementList(List *placementList)
  * repeatedly moves the list's first element to the end of the list, and
  * then returns the newly rotated list.
  */
-List *
+static List *
 LeftRotateList(List *list, uint32 rotateCount)
 {
 	List *rotatedList = list_copy(list);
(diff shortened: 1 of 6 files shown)


src/test/regress/expected/multi_task_assignment_policy_0.out
```diff
@@ -176,3 +176,84 @@ RESET client_min_messages;
...
+
+EXPLAIN SELECT * FROM task_assignment_reference_table;
+                                                              QUERY PLAN                                                              
+--------------------------------------------------------------------------------------------------------------------------------------
+ Custom Scan (Citus Router)  (cost=0.00..0.00 rows=0 width=0)
+   Task Count: 1
+   Tasks Shown: All
+   ->  Task
+         Node: host=localhost port=57638 dbname=regression
+         ->  Seq Scan on task_assignment_reference_table_880000 task_assignment_reference_table  (cost=0.00..35.50 rows=2550 width=4)
```
marcocitus: we always disable costs when using EXPLAIN in regression tests, since the cost estimate may change over time / different postgres versions

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -1757,7 +1755,7 @@ ChildNodeList(MultiNode *multiNode)
  * When citus.enable_unique_job_ids is off then only the local counter is
  * included to get repeatable results.
  */
-static uint64
+uint64
 UniqueJobId(void)
```
marcocitus: I wonder whether this function should now be in a separate file.

thanodnl: I was thinking the same, but since the Job struct also is defined in `multi_physical_planner` I found no better place to put it. I can extract everything related to jobs :)

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1625,34 +1629,66 @@ RouterJob(Query *originalQuery, PlannerRestrictionContext *plannerRestrictionCon
...
+												  shardId);
+
+		/*
+		 * Queries to reference tables, or distributed tables with multiple replica's have
+		 * their task placements reordered according to the configured
+		 * task_assignment_policy. This is only applicable to select queries as the modify
+		 * queries will be reordered to _always_ use the first-replica policy during
+		 * execution.
+		 */
+		ReorderTaskPlacementsByTaskAssignmentPolicy(job, TaskAssignmentPolicy);
```
onderkalaci: With replication factor = 1 and reference table usage, many users would hit this and need to change the setting to spread the load on the reference tables across the cluster. (For replication factor == 1, task assignment policies don't have any affect since there is only a single placement and we have to assign it 😄 ).

Do you think would it make sense to make this behavior default? If yes, I think we need to discuss about how to make it :)

marcocitus: I don't think it should be the default because it means consecutive SELECTs may flip back-and-forth between different versions of the reference table depending on whether a commit has happened on the shard placement or not (no monotonic read consistency). Using greedy/first-replica doesn't entirely prevent that, but at least makes it quite unlikely that you'll go back to an older version since it can only happen in case of failure.

onderkalaci: Do you mean that if we do something like the following concurrently we're more likely to get inconsistent results on session 2 and/or session 3?

```SQL
-- session 1
INSERT INTO ref_table .. ();
```

```SQL
-- session 2
SELECT count(*) FROM ref_table;

-- another select from the same session
SELECT count(*) FROM ref_table;
```


```SQL
-- session 3
SELECT count(*) FROM ref_table;
```

Or, what you've mind is something else?

marcocitus: Yes, the second count(*) could see a lower value than the first count(*) when the task assignment policy is round-robin. That's an anomaly that we normally would not expect. 

Another potential downside is a worse cache hit ratio, especially because reference tables end up being redundantly cached on every node.

mtuncer: @marcocitus that should only be possible if the query comes in just between commit times to workers right ?

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1625,34 +1629,66 @@ RouterJob(Query *originalQuery, PlannerRestrictionContext *plannerRestrictionCon
...
+ * ReorderTaskPlacementsByTaskAssignmentPolicy applies selective reordering for supported
+ * TaskAssignmentPolicyTypes.
+ *
+ * Supported Types
+ * - TASK_ASSIGNMENT_ROUND_ROBIN round robin schedule queries among placements
+ *
+ * By default it does not reorder the task list, implying a first-replica strategy.
+ */
+static void
+ReorderTaskPlacementsByTaskAssignmentPolicy(Job *job, TaskAssignmentPolicyType
```
mtuncer: This function looked more like it belonged to multi_physical_planner with the rest of task assignment functions

src/test/regress/expected/multi_task_assignment_policy_0.out
```diff
@@ -176,3 +176,84 @@ RESET client_min_messages;
...
+ Custom Scan (Citus Router)
+   Task Count: 1
+   Tasks Shown: All
+   ->  Task
+         Node: host=localhost port=57638 dbname=regression
+         ->  Seq Scan on task_assignment_reference_table_880000 task_assignment_reference_table
+(6 rows)
+
+SET citus.task_assignment_policy TO 'round-robin';
+EXPLAIN (COSTS FALSE) SELECT * FROM task_assignment_reference_table;
```
mtuncer: Is this deterministic ? If we end up adding any new query above, we would have to get here and fix it. It may get worse if we have parallel tasks doing this.

src/backend/distributed/planner/multi_router_planner.c
thanodnl: I thought about this as well, however this function is a very router planner specific version to `AssignAnchorShardTaskList `: https://github.com/citusdata/citus/blob/89bc3ad5cbd2c4616840439f6de6d244f31a5f0d/src/backend/distributed/planner/multi_physical_planner.c#L4903-L4924

Unfortunately there were all kinds of side effects when reusing that function when the policy was not set to `round-robin`. Given that that function was not reusable for this feature and this is very specific to the router planner I am inclined to keep it here.

src/test/regress/expected/multi_task_assignment_policy_0.out
thanodnl: you are right, it could be undeterministic if there is an odd amount of calls to `UniqueJobId` in between which can be cause by queries, however the tests above in this file suffer the same.

I have changed the assertion to be more inline with the tests above.

PR #2473, opened 2018-11-13
Use function permissions to guard node metadata functions 

marcocitus:
We currently enforce permissions for functions that manipulate node metadata through `EnsureSuperuser()` calls, but allow all users to call them.

This has two downsides:
- it encourages bugs, since we may forget to check permissions
- it encourages poor security practices, since you always need superuser (e.g. you cannot create a user which can manipulate cluster metadata, but not access user data)

This PR revokes permissions for all node metadata manipulating functions from public and removes the `EnsureSuperuser()` checks, such that administrators can decide who can call the functions and by default only superusers can.

I did not bump the schema version to 8.1 yet. Since we don't have other schema changes in the current development cycle yet we could still elect to backport to 8.0.


marcocitus:
> The changes are nice, but given the description I think we need to add regression tests to make sure if the super user granted permission to a user to call these functions they can actually be called. This ensures we have removed all the required EnsureSuperUser calls, which I believe is the case.

Added. The node ID / group ID sequences are called 2 extra times, which caused some output changes in the MX tests because MX internals depend heavily on group ID. For now I fixed it by updating the MX tests.

PR #2475, opened 2018-11-13
Fix permissions checks in UDFs called from the drop table trigger

### LINKED ISSUE #2274: master_drop_sequences should connect as current user

onderkalaci:
Not fixes #2274 as described in the issue, but addresses in another way, similar to how drop table is handled.

In MX, we allow unauthorized users to drop (or at least mess up) tables/sequences. This PR prevents that.

I'm assigning @marcocitus since he asked for the change. But, feel to re-assign to @metdos if you're not avaliable.

### DIFF
--- a/src/backend/distributed/master/master_delete_protocol.c
+++ b/src/backend/distributed/master/master_delete_protocol.c
@@ -58,6 +58,9 @@
 #include "utils/elog.h"
 #include "utils/errcodes.h"
 #include "utils/lsyscache.h"
+#if (PG_VERSION_NUM >= 100000)
+#include "utils/varlena.h"
+#endif
 
 
 /* Local functions forward declarations */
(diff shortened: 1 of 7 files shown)


src/backend/distributed/master/master_delete_protocol.c
```diff
@@ -268,12 +269,13 @@ master_drop_sequences(PG_FUNCTION_ARGS)
 	Datum sequenceText = 0;
 	bool isNull = false;
 	StringInfo dropSeqCommand = makeStringInfo();
-	bool coordinator = IsCoordinator();
 
 	CheckCitusVersion(ERROR);
 
-	/* do nothing if DDL propagation is switched off or this is not the coordinator */
-	if (!EnableDDLPropagation || !coordinator)
+	EnsureCoordinator();
```
marcocitus: I think this function always gets called when anything is dropped on Postgres, so that's why we don't want an explicit error.

onderkalaci: Makes sense, I'll undo that and add a test.

onderkalaci: Should be ready for review again.

PR #2479, opened 2018-11-15
Cleanup create_distributed_table.c.

pykello:
Removing some unnecessary code while I'm looking at this file to debug a bug.

### DIFF
--- a/src/backend/distributed/commands/create_distributed_table.c
+++ b/src/backend/distributed/commands/create_distributed_table.c
@@ -323,8 +323,6 @@ CreateDistributedTable(Oid relationId, Var *distributionColumn, char distributio
 	Oid colocatedTableId = InvalidOid;
 	bool localTableEmpty = false;
 
-	Relation colocatedRelation = NULL;
-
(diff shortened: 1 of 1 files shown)


pykello:
@mtuncer can you please take a quick look? This doesn't change any behaviour, it just removes a variable that is not used, but was confusing when reading the code.

src/backend/distributed/commands/create_distributed_table.c
```diff
@@ -358,13 +356,6 @@ CreateDistributedTable(Oid relationId, Var *distributionColumn, char distributio
 	 */
 	if (viaDeprecatedAPI)
 	{
-		/*
-		 * We exit early but there is no need to close colocatedRelation. Because
-		 * if viaDeprecatedAPI is true, we never open colocatedRelation in the first
-		 * place.
-		 */
-		Assert(colocatedRelation == NULL);
```
onderkalaci: For this line, instead of removing we should maybe change the assert `colocateWithTableName == NULL` here?

pykello: done

PR #2483, opened 2018-11-19
Fix permissions checks in lesser-known UDFs

marcocitus:
DESCRIPTION: Fixes permission checks in mark_tables_colocated, upgrade_to_reference_table, lock_relation_if_exists

This PR adds permissions checks to lesser-known UDFs:
- upgrade_to_reference_table (upgrade single shard tables to reference tables)
- mark_tables_colocated (mark tables as co-located if conditions are met)
- lock_relation_if_exists (internal function that is used to implement TRUNCATE in MX)

The `upgrade_to_reference_table` and `mark_tables_colocated` functions now throw an error unless the caller is the table owner (for all tables in case of `mark_tables_colocated`) or superuser.

The `lock_relation_if_exists` check uses the `RangeVarGetRelidExtended` callback from `lockcmds.c`. The callback in `RangeVarGetRelidExtended` is meant to check permissions just before taking the lock, but just after resolving the name in a safe manner.

### DIFF
--- a/src/backend/distributed/utils/resource_lock.c
+++ b/src/backend/distributed/utils/resource_lock.c
@@ -19,6 +19,7 @@
 
 #include "access/xact.h"
 #include "catalog/namespace.h"
+#include "commands/tablecmds.h"
 #include "distributed/colocation_utils.h"
 #include "distributed/listutils.h"
 #include "distributed/master_metadata_utility.h"
@@ -36,6 +37,7 @@
 #include "distributed/version_compat.h"
 #include "storage/lmgr.h"
 #include "utils/builtins.h"
+#include "utils/lsyscache.h"
 #if (PG_VERSION_NUM >= 100000)
 #include "utils/varlena.h"
 #endif
@@ -73,6 +75,9 @@ static void LockShardListResources(List *shardIntervalList, LOCKMODE lockMode);
 static void LockShardListResourcesOnFirstWorker(LOCKMODE lockmode,
 												List *shardIntervalList);
 static bool IsFirstWorkerNode();
+static void CitusRangeVarCallbackForLockTable(const RangeVar *rangeVar, Oid relationId,
+											  Oid oldRelationId, void *arg);
+static AclResult CitusLockTableAclCheck(Oid relationId, LOCKMODE lockmode, Oid userId);
 
 
 /* exports for SQL callable functions */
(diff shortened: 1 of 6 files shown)


src/backend/distributed/utils/resource_lock.c
```diff
@@ -760,7 +765,88 @@ lock_relation_if_exists(PG_FUNCTION_ARGS)
...
+
+	relkind = get_rel_relkind(relationId);
+	if (!relkind)
+	{
+        /* table was concurrently dropped */
+		return;
+	}
+
+    /* we only allow tables and views to be locked */
+	if (relkind != RELKIND_RELATION && relkind != RELKIND_PARTITIONED_TABLE &&
```
onderkalaci: PG 9.6 compile error. We've `RegularTable(oid)` which handles the first two.

src/backend/distributed/utils/resource_lock.c
```diff
@@ -760,7 +765,88 @@ lock_relation_if_exists(PG_FUNCTION_ARGS)
...
+
+/*
+ * CitusLockTableAclCheck checks whether a user has permission to lock a relation
+ * in the given lock mode.
+ *
+ * This function is a copy of LockTableAclCheck in lockcmds.c adapted to Citus
+ * code style.
+ */
+static AclResult
+CitusLockTableAclCheck(Oid reloid, LOCKMODE lockmode, Oid userId)
```
onderkalaci: Minor once we're already changing to Citus style: `reloid ` -> `relationId` or such?

src/backend/distributed/utils/resource_lock.c
onderkalaci: We have `RegularTable(Oid)` which could be useful for getting rid of pg 96 warnings for the first two.

src/backend/distributed/utils/resource_lock.c
```diff
@@ -760,7 +765,88 @@ lock_relation_if_exists(PG_FUNCTION_ARGS)
...
+{
+	AclResult aclResult;
+	AclMode aclMask;
+
+    /* Verify adequate privilege */
+	if (lockmode == AccessShareLock)
+	{
+		aclMask = ACL_SELECT;
+	}
+	else if (lockmode == RowExclusiveLock)
```
onderkalaci: I can see that this is copy & pasted from Postgres. But still, I couldn't understand the logic in this `else if`/`else` block.

src/backend/distributed/utils/resource_lock.c
marcocitus: Changing it to use only `RegularTable(oid)` since we don't care about views here.

PR #2486, opened 2018-11-21
Fix create_distributed_table as non-table-owner

### LINKED ISSUE #2094: create_distributed_table gives misleading error when called as non-table-owner
When a table that contains data is distributed, `create_distributed_table` may give an error such as:
```
cannot establish a new connection for placement 374, since DDL has been executed on a connection that is in use
```

What happens is that Citus connects as the table owner to create the shards (the "DDL") to make sure the shard has the same owner, but then Citus tries to copy data into the shards as the current user, which would require a new connection, but that new connection would not be able to see the new table yet.

We could require that `create_distributed_table` is always called by the table owner, but that might break some existing workflows. We could only require it when the table contains data, such that we just change the error message into something more meaningful.

thanodnl:
Descrption: Fix create_distributed_table as non-table-owner

Fixes #2094

This change started as adding an error to the creation of a distributed table when the super user tries to distribute a table of an other owner. After going through the code it is actually non-trivial to test this as the super user is the owner internally of everything, even when not the owner. Meaning, all ownership tests functions exposed by postgres are short circuited to return true when called by the super user.

Instead I have changed the logic that creates a distributed table. This logic used to connect as the owner of the table instead of the current user. This is contrary to most of our distribution logic where we always connect as the current user.

Now we connect as the current user and add an `ALTER TABLE ... OWNER TO ...` command to the creation of a remote table. This is in line with how `pg_dump` recreates a table owned by an other user.

Most of this diff is changing existing tests due to the added ddl command. I have thought long on what to do here and settled on this approach. Things considered are:
 - Only add the command when the table is not owned by the current user.
    Downside: when the command is called before a user switch this could cause issues that are hard to debug.
 - move all occurrences of `master_get_table_ddl_events` to one file that is responsible for testing it
   Downside: big change and hard to enforce in the future.
 - not fixing it and just error out
   Downside: just not good UX to disallow a super user to distribute a table.

I am still trying to figure out what extra tests to add, there is already 1 test that covers the ownership of distributed tables. Things to add are below

 - [x] Add regression test for distributing a table with data as non-owner
 - [x] Add test as non-owner that is not allowed to distribute the table


src/backend/distributed/utils/citus_ruleutils.c
```diff
@@ -676,6 +676,26 @@ pg_get_tablecolumnoptionsdef_string(Oid tableRelationId)
 }
 
 
+/*
+ * pg_get_tableowner_string
+ */
+char *
+pg_get_tableowner_string(Oid relationId)
```
onderkalaci: Minor note: There is `TableOwnerResetCommand()` implemented for Citus MX. Can we use it here, or is there a reason for not doing that?

src/backend/distributed/master/master_node_protocol.c
```diff
@@ -666,6 +666,14 @@ GetTableCreationCommands(Oid relationId, bool includeSequenceDefaults)
 		tableDDLEventList = lappend(tableDDLEventList, tableColumnOptionsDef);
 	}
 
+	{
+		char *tableOwnerDef = pg_get_tableowner_string(relationId);
+		if (tableOwnerDef != NULL)
+		{
+			tableDDLEventList = lappend(tableDDLEventList, tableOwnerDef);
```
marcocitus: We should probably verify whether a user can always make itself owner:

> To alter the owner, you must also be a direct or indirect member of the new owning role, and that role must have CREATE privilege on the table's schema. (These restrictions enforce that altering the owner doesn't do anything you couldn't do by dropping and recreating the table. However, a superuser can alter ownership of any table anyway.) 

It seems that I can do `ALTER TABLE me TO me` in PG11 even if I don't have create privileges on the schema, but we should have a test that makes sure this is also true for 9.6 and 10 and remains true in future versions.

velioglu:
After fixing the error on failure tests and rebasing, it looks good to me. Please keep the commit replacing `table_ddl_command_array` with `master_get_table_ddl_events` as separate one.

PR #2487, opened 2018-11-21
Execute SQL tasks using worker_execute_sql_task UDF when using task-tracker

marcocitus:
DESCRIPTION: Execute SQL tasks using worker_execute_sql_task UDF when using task-tracker

We currently use `COPY (SELECT ...) TO 'base/pgsql_job_cache/...'` in the task-tracker to write the results of SQL tasks to files.  This PR changes the behaviour to a) use a `worker_execute_sql_task` UDF instead to be able to avoid hacks in the COPY hook in the future and b) calls the same logic from the COPY hook to allow a more seamless upgrade path.

A lot of the code is borrowed from `intermediate_results.c`. Long-term we should replace the task files with intermediate results which are tied to a particular user and a distributed transaction. However, those files are cleaned up at commit/abort, so we cannot use them in task-tracker which does all tasks as individual transactions.

There is one follow-up task that will be in a separate PR:
- Add user names to file names to prevent users from overwriting each other's files


src/backend/distributed/worker/worker_sql_task_protocol.c
```diff
@@ -0,0 +1,301 @@
+/*-------------------------------------------------------------------------
```
onderkalaci: we also probably need to add this file to `./Makefile` for appveyor to pass

PR #2488, opened 2018-11-21
Validate Constraint Support

### LINKED ISSUE #2441: VALIDATE is not supported
Citus currently does not support `VALIDATE` command. We are able to add `not valid` constraints which is good but these kind of constraints are added to skip the burden of validating them initially **to be validated** later. It is also mentioned in this [comment](https://github.com/citusdata/citus/issues/968#issuecomment-391362278) by me for foreign key constraints to reference tables. 

I believe this is a nice candidate as a warm-up task though. @marcocitus should I create a label and add it to this issue as it is discussed in the dev discussions last week?

hanefi:
DESCRIPTION: Add validate constraint sql query support

Citus allows creation of Foreign Key and Check Constraints as `not valid`. This PR grants Citus the capability to run `VALIDATE CONSTRAINT` queries.

The queries that can have `NOT VALID` option are of the form:
```SQL
ALTER TABLE [ IF EXISTS ] [ ONLY ] name [ * ] ADD table_constraint NOT VALID [, ...]
```
Where table_constraint 
```SQL
[ CONSTRAINT constraint_name ]
{ CHECK ( expression ) [ NO INHERIT ] |
UNIQUE ( column_name [, ... ] ) index_parameters |
PRIMARY KEY ( column_name [, ... ] ) index_parameters |
EXCLUDE [ USING index_method ] ( exclude_element WITH operator [, ... ] ) index_parameters [ WHERE ( predicate ) ] |
FOREIGN KEY ( column_name [, ... ] ) REFERENCES reftable [ ( refcolumn [, ... ] ) ]
  [ MATCH FULL | MATCH PARTIAL | MATCH SIMPLE ] [ ON DELETE action ] [ ON UPDATE action ] }
[ DEFERRABLE | NOT DEFERRABLE ] [ INITIALLY DEFERRED | INITIALLY IMMEDIATE ]
```

On the other hand `NOT VALID` option is only allowed for foreign key and CHECK constraints right now. I added the necessary failing tests for the other constraint types so that when Postgres supports `NOT VALID` option for them in the future, we will be notified.

Closes #2441

### DIFF
--- a/src/backend/distributed/commands/table.c
+++ b/src/backend/distributed/commands/table.c
@@ -1060,11 +1060,13 @@ ErrorIfUnsupportedAlterTableStmt(AlterTableStmt *alterTableStatement)
 			case AT_EnableTrigAll:
 			case AT_DisableTrigAll:
 			case AT_ReplicaIdentity:
+			case AT_ValidateConstraint:
 			{
 				/*
 				 * We will not perform any special check for ALTER TABLE DROP CONSTRAINT
 				 * , ALTER TABLE .. ALTER COLUMN .. SET NOT NULL and ALTER TABLE ENABLE/
-				 * DISABLE TRIGGER ALL, ALTER TABLE .. REPLICA IDENTITY ..
+				 * DISABLE TRIGGER ALL, ALTER TABLE .. REPLICA IDENTITY .., ALTER TABLE
+				 * .. VALIDATE CONSTRAINT ..
 				 */
 				break;
 			}
@@ -1083,7 +1085,7 @@ ErrorIfUnsupportedAlterTableStmt(AlterTableStmt *alterTableStatement)
 						(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
 						 errmsg("alter table command is currently unsupported"),
 						 errdetail("Only ADD|DROP COLUMN, SET|DROP NOT NULL, "
-								   "SET|DROP DEFAULT, ADD|DROP CONSTRAINT, "
+								   "SET|DROP DEFAULT, ADD|DROP|VALIDATE CONSTRAINT, "
 								   "SET (), RESET (), "
 								   "ATTACH|DETACH PARTITION and TYPE subcommands "
 								   "are supported.")));
--- a/src/test/regress/expected/validate_constraint.out
+++ b/src/test/regress/expected/validate_constraint.out
@@ -1,0 +1,150 @@
+--
+--  VALIDATE_CONSTRAINT
+--
+-- PostgreSQL 11 Docs Excerpt:
+-- The queries that can have 'NOT VALID' option are of the form:
+-- ALTER TABLE [ IF EXISTS ] [ ONLY ] name [ * ] ADD table_constraint NOT VALID [, ...]
+--
+-- where table_constraint is:
+--   [ CONSTRAINT constraint_name ]
+--   { CHECK ( expression ) [ NO INHERIT ] |
+--     UNIQUE ( column_name [, ... ] ) index_parameters |
+--     PRIMARY KEY ( column_name [, ... ] ) index_parameters |
+--     EXCLUDE [ USING index_method ] ( exclude_element WITH operator [, ... ] ) index_parameters [ WHERE ( predicate ) ] |
+--     FOREIGN KEY ( column_name [, ... ] ) REFERENCES reftable [ ( refcolumn [, ... ] ) ]
+--       [ MATCH FULL | MATCH PARTIAL | MATCH SIMPLE ] [ ON DELETE action ] [ ON UPDATE action ] }
+--   [ DEFERRABLE | NOT DEFERRABLE ] [ INITIALLY DEFERRED | INITIALLY IMMEDIATE ]
+-- This form adds a new constraint to a table using the same syntax as CREATE TABLE, plus the option NOT VALID, which is
+-- currently only allowed for foreign key and CHECK constraints. If the constraint is marked NOT VALID, the
+-- potentially-lengthy initial check to verify that all rows in the table satisfy the constraint is skipped.
+-- The constraint will still be enforced against subsequent inserts or updates (that is, they'll fail unless there is a
+-- matching row in the referenced table, in the case of foreign keys; and they'll fail unless the new row matches the
+-- specified check constraints). But the database will not assume that the constraint holds for all rows in the table,
+-- until it is validated by using the VALIDATE CONSTRAINT option.
+CREATE SCHEMA validate_constraint;
+SET search_path TO 'validate_constraint';
+SET citus.shard_replication_factor TO 1;
+SET citus.shard_count TO 8;
+SET citus.next_shard_id TO 8000000;
+SET citus.next_placement_id TO 8000000;
+CREATE TYPE constraint_validity AS (name text, validated bool);
+SELECT run_command_on_workers(
+         $$CREATE TYPE constraint_validity AS (name text, validated bool)$$);
+      run_command_on_workers       
+-----------------------------------
+ (localhost,57637,t,"CREATE TYPE")
+ (localhost,57638,t,"CREATE TYPE")
+(2 rows)
+
+CREATE VIEW constraint_validations_in_workers AS
+  SELECT (json_populate_record(NULL :: constraint_validity,
+                               json_array_elements_text((run_command_on_workers($$
+    SELECT
+      COALESCE(json_agg(row_to_json(d)), '[]'::json)
+    FROM
+      (
+        SELECT conname as name,
+        convalidated as validated
+        FROM pg_catalog.pg_constraint AS con
+        JOIN pg_catalog.pg_namespace AS ns
+        ON ns.oid = con.connamespace
+        WHERE nspname = 'validate_constraint'
+        AND contype = 'c'
+        ORDER BY 1,2
+      )
+      d $$)).RESULT :: json) :: json)).*
+    ORDER BY 1,2;
+CREATE VIEW constraint_validations AS
+  SELECT conname as "Constraint",
+         convalidated as "Validated?"
+  FROM pg_catalog.pg_constraint AS con
+         JOIN pg_catalog.pg_namespace AS ns ON ns.oid = con.connamespace
+  WHERE nspname = 'validate_constraint'
+    AND contype = 'c';
+CREATE TABLE referenced_table (id int UNIQUE, test_column int);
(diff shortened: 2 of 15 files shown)


src/backend/distributed/commands/table.c
```diff
@@ -1060,11 +1060,13 @@ ErrorIfUnsupportedAlterTableStmt(AlterTableStmt *alterTableStatement)
 			case AT_EnableTrigAll:
 			case AT_DisableTrigAll:
 			case AT_ReplicaIdentity:
+			case AT_ValidateConstraint:
 			{
 				/*
 				 * We will not perform any special check for ALTER TABLE DROP CONSTRAINT
 				 * , ALTER TABLE .. ALTER COLUMN .. SET NOT NULL and ALTER TABLE ENABLE/
-				 * DISABLE TRIGGER ALL, ALTER TABLE .. REPLICA IDENTITY ..
+				 * DISABLE TRIGGER ALL, ALTER TABLE .. REPLICA IDENTITY .., ALTER TABLE
```
onderkalaci: Should update the `errdetail ` below to add `VALIDATE `?

hanefi: I updated the `errdetail` to include `VALIDATE CONSTRAINT` support

src/test/regress/expected/validate_constraint.out
```diff
@@ -0,0 +1,254 @@
...
+ALTER TABLE constrained_table
+  VALIDATE CONSTRAINT fk_constraint;
+ROLLBACK;
+-- It is possible that some other table_constraint commands will support NOT VALID option in the future
+-- These should fail as Postgres does not yet allow these constraints to be NOT VALID
+-- If one of these queries are not failing on a future Postgres version, we will need to test corresponding VALIDATE CONSTRAINT queries
+BEGIN;
+ALTER TABLE constrained_table
+  ADD CONSTRAINT unique_constraint UNIQUE (constrained_column) NOT VALID;
+ERROR:  UNIQUE constraints cannot be marked NOT VALID
```
onderkalaci: I think we should remove the constraints which cannot be marked as NOT VALID, such as UNIQUE

src/backend/distributed/commands/table.c
hanefi: changing the error messages changed the outputs of other tests. I need to make changes to several other test outputs to sort this out.

src/test/regress/multi_schedule
```diff
@@ -173,7 +173,8 @@ test: multi_modifications
 test: multi_distribution_metadata
 test: multi_generate_ddl_commands multi_create_shards multi_prune_shard_list multi_repair_shards 
 test: multi_upsert multi_simple_queries multi_create_insert_proxy multi_data_types
-test: multi_utilities foreign_key_to_reference_table validate_constraint
```
hanefi: Setting different values to `citus.next_shard_id` and `citus.next_placement_id` in `foreign_key_to_reference_table` and `validate_constraint` tests are not causing any problem in my local, but I am guessing they are the main reason why one of these two tests fails occasionally on CI.

Either CI gives a timeout in this parallel group, or we have a different number of rows printed in a test query

src/test/regress/expected/validate_constraint.out
onderkalaci: I think what disturbs me here is not having the commands that might be supported in the future. But, mostly trying to execute ` VALIDATE CONSTRAINT ` command after a failure in the transaction. Because, that only adds some verbose output to the test and hard to track. We could even remove `BEGIN/ROLLBACK` steps in the failing tests and only keep `ALTER TABLE .. ADD CONSTRAINT .. NOT VALID` commands.

src/test/regress/expected/validate_constraint.out
```diff
@@ -0,0 +1,256 @@
...
+from generate_series(1, 1000) as f (x);
+-- a constraint that can be validated
+ALTER TABLE constrained_table
+  ADD CONSTRAINT validatable_constraint CHECK (constrained_column < 10000) NOT VALID;
+ALTER TABLE constrained_table
+  VALIDATE CONSTRAINT validatable_constraint;
+-- a constraint that can not be validated
+ALTER TABLE constrained_table
+  ADD CONSTRAINT failing_constraint CHECK (constrained_column > 100) NOT VALID;
+ALTER TABLE constrained_table
```
onderkalaci: I remember suggesting adding such a test where the validation fails. But, now I'm a little bit skeptical to this. The reason is that, the output of this test might change randomly. In this test, probably the first shard is complaining about the violation. But, its possible that the same test errors out on the second shard and these error messages change.

So, given that you've `constraint_validations_in_workers ` view, I'd suggest to remove these test.

If you really want to have such a test, you should hide the shard names in the error messages. An example you can see is here: Define a function which absorbs the message as [1](https://github.com/citusdata/citus/pull/2475/files#diff-ddf4b11e6f488ae78926f36e2f9fb864R13) and call the command within that function [2](https://github.com/citusdata/citus/pull/2475/files#diff-ddf4b11e6f488ae78926f36e2f9fb864R316).

But, as I mentioned, given the view, we don't need it. I mostly pointed out for future reference.

src/test/regress/expected/validate_constraint.out
```diff
@@ -0,0 +1,256 @@
...
+ validatable_constraint_8000015 | t
+ validatable_constraint_8000016 | t
+(24 rows)
+
+DROP TABLE constrained_table;
+CREATE TABLE referenced_table (
+  id          int UNIQUE,
+  test_column int
+);
+SELECT create_distributed_table('referenced_table', 'id');
```
onderkalaci: Creating and removing the tables are costly in terms of the regression test execution times. Can't we use the same tables throughout the test?

src/test/regress/expected/validate_constraint.out
```diff
@@ -0,0 +1,256 @@
...
+INSERT INTO referencing_table
+SELECT x, x
+from generate_series(1, 100) as f (x);
+ALTER TABLE referencing_table
+  ADD CONSTRAINT fk_constraint FOREIGN KEY (ref_id) REFERENCES referenced_table (id) NOT VALID;
+-- should fail as the foreign keys are not present
+ALTER TABLE referencing_table
+  VALIDATE CONSTRAINT fk_constraint;
+ERROR:  insert or update on table "referencing_table" violates foreign key constraint "fk_constraint"
+DETAIL:  Key (ref_id)=(1) is not present in table "referenced_table".
```
onderkalaci: Similar here. Validation output may likely to change in some executions and end-up with random test failures, which is not good. 

I think, I can see your goal of this test. You're trying to test whether the constraint is applied afterwards even if it is not valid. Well, that's something PostgreSQL enforces, so I'd be OK to remove these tests. Because, other parts of the tests are already covered by the above tests.

src/test/regress/multi_schedule
```diff
@@ -174,6 +174,7 @@ test: multi_distribution_metadata
 test: multi_generate_ddl_commands multi_create_shards multi_prune_shard_list multi_repair_shards 
 test: multi_upsert multi_simple_queries multi_create_insert_proxy multi_data_types
 test: multi_utilities foreign_key_to_reference_table
+test: validate_constraint
```
onderkalaci: As we've talked in person, we should be able to run the tests concurrently with others. I'd be OK to run the tests with some others if it fails in the above bucket.

src/test/regress/expected/validate_constraint.out
hanefi: Valid point. I am removing the statements that are not executed in aborted transactions, and those `BEGIN/ROLLBACK` steps.

I guess I was a little bit confused about when it is appropriate to use those transaction blocks in a test. This task helped me get a better understanding. 

Correct me if I am wrong, but I think that it is better to use a `BEGIN/ROLLBACK` block when we wish to run a query that makes a successful change in the schema but we wish to revert it to test some other thing afterwards. (Which is not the case here, so removing them makes perfect sense)

src/test/regress/multi_schedule
hanefi: I wanted to understand why `foreign_key_to_reference_table` tests fail when run in parallel with mine. Made some little changes to a helper view in `foreign_key_to_reference_table` so that that view contains the foreign keys only in its associated schema and not the fks on other parallel tests

src/test/regress/expected/validate_constraint.out
hanefi: After removing the tests that are not needed as they are enforced by Postgres, I arrived at the following:

- insert invalid references
- add not valid constraint
- ~fail to validate~
- ~fail to insert invalid references as Postgres enforces not valid constraints on insertion~
- create missing references
- validate

Now these steps does not really mean much, so I am removing these steps all together

src/test/regress/expected/validate_constraint.out
hanefi: I made the necessary changes to reuse the 3 tables I created in the beginning of the tests, and drop them at the very end.

src/test/regress/expected/validate_constraint.out
```diff
@@ -0,0 +1,150 @@
...
+ALTER TABLE constrained_table
+  ADD CONSTRAINT exclude_constraint EXCLUDE USING gist (id WITH =, constrained_column WITH <>) NOT VALID;
+ERROR:  EXCLUDE constraints cannot be marked NOT VALID
+INSERT INTO constrained_table
+SELECT x, x
+from generate_series(1, 1000) as f (x);
+-- a constraint that can be validated
+ALTER TABLE constrained_table
+  ADD CONSTRAINT validatable_constraint CHECK (constrained_column < 10000) NOT VALID;
+ALTER TABLE constrained_table
```
onderkalaci: I think, before executing `VALIDATE CONSTRAINT`, we should query the views `constraint_validations ` and `constraint_validations_in_workers ` to ensure that we're successfully added invalid constraints.

src/test/regress/specs/isolation_validate_vs_insert.spec
```diff
@@ -0,0 +1,39 @@
+#
+# How we organize this isolation test spec, is explained at README.md file in this directory.
+#
+
+# create distributed table to test behavior of VALIDATE in concurrent operations
```
onderkalaci: Minor: It'd be great to mention that `ALTER TABLE % VALIDATE CONSTRAINT` doesn't block inserts and selects on the table with this comment. 

Because, in general `ALTER TABLE` commands blocks all activity on the tables, and this one is unusual in that sense.

PR #2489, opened 2018-11-21
Add user ID suffix to intermediate files in task-tracker

marcocitus:
DESCRIPTION: Add user ID suffix to intermediate files generated by task-tracker

### Design

The task-tracker functions that run on the worker can currently be called by any user. While it's not possible to read intermediate files belonging to another user directly, it would be technically possible to overwrite another user's intermediate files by guessing the job ID and the task ID while the user is running a re-partition query, which could affect the outcome of the query or prevent it from finishing.

This PR adds the user ID as a suffix to all files generated by task-tracker-functions, which allows the `worker_merge_*` functions to verify whether a file was generated (via `worker_*_partition_table`) and fetched (via `worker_fetch_partition_file`) by the user doing the merge and makes sure the executor never fetches results generated by a different user.

The `COPY ... WITH (format 'transmit')` command is always executed over a superuser connection and therefore it does not know the user ID to append. Instead, this PR allows you to add a `user '<username>'` attribute to the `WITH` section. The `worker_fetch_partition_file` sets this attribute to the user that called the function and the executor sets it to the current user.

### Alternatives considered

The reason that we're not appending to the filename in the COPY statement directly is that the OIDs may differ across nodes.

I've also considered other designs such as a per-user directory or prefix, but this was a far bigger change and breaks the protocol, which did not seem worth doing. 

Long-term, we should replace task files with intermediate results, which are isolated per user and cleaned up at the end of the transaction. However, we currently cannot use intermediate results in task-tracker because every task runs as a separate transaction.


src/backend/distributed/commands/multi_copy.c
```diff
@@ -2625,6 +2626,14 @@ ProcessCopyStmt(CopyStmt *copyStatement, char *completionTag)
 			}
 
 			query = (Query *) linitial(queryTreeList);
+
+			/*
+			 * Add a username suffix to prevent other users from reading/writing
+			 * the same file. We do this consistently in all functions that interact
+			 * with task files.
+			 */
```
onderkalaci: I think we should quote the user name on all places. For example, the following user fails to run repartition queries now ` CREATE user "on'ender";`

src/backend/distributed/commands/transmit.c
```diff
@@ -359,6 +359,30 @@ IsTransmitStmt(Node *parsetree)
...
 
 
+/*
+ * TransmitStatementUser extracts the user attribute from a
+ * COPY ... (format 'transmit', user '...') statement.
+ */
+char *
+TransmitStatementUser(CopyStmt *copyStatement)
+{
+	ListCell *optionCell = NULL;
```
onderkalaci: Shall we add `AssertArg(IsTransmitStmt((Node *) copyStatement))` to ensure that all callers of this function are transmit statements? 

This is mostly for future callers not to expected non-transmist statements to have `user` field?

src/include/distributed/worker_protocol.h
```diff
@@ -45,6 +45,8 @@
 /* Defines used for fetching files and tables */
 /* the tablename in the overloaded COPY statement is the to-be-transferred file */
 #define TRANSMIT_REGULAR_COMMAND "COPY \"%s\" TO STDOUT WITH (format 'transmit')"
```
onderkalaci: Do we have any callers of `TRANSMIT_REGULAR_COMMAND `? Maybe remove `TRANSMIT_WITH_USER_COMMAND ` and add `user '%s'` to `TRANSMIT_REGULAR_COMMAND` for future callers?

src/backend/distributed/worker/worker_partition_protocol.c
```diff
@@ -606,7 +607,14 @@ TaskDirectoryName(uint64 jobId, uint32 taskId)
 }
 
 
-/* Constructs a standardized partition file path for given directory and id. */
+/*
+ * PartitionFilename returns a partition file path for given directory and id
```
onderkalaci: I cannot follow the difference between `PartitionFilename ()` and `UserPartitionFilename ()` 😞 

My main question is that, why do we still need `PartitionFilename ()` ?

src/backend/distributed/commands/multi_copy.c
onderkalaci: Oh, is it? As far as I followed, the `ereport(WARNING` check relies on the user name being the same? I'll read the code a bit more.

src/include/distributed/worker_protocol.h
marcocitus: Removing the `TRANSMIT_REGULAR_COMMAND`. I kind of like the `WITH_USER` part since it's a bit dangerous if you forget.

src/backend/distributed/worker/worker_partition_protocol.c
marcocitus: PartitionFilename:
`base/pgsql_job_cache/job_201010/task_101106/p_00000`

UserPartitionfilename:
`base/pgsql_job_cache/job_201010/task_101106/p_00000.marco`

`PartitionFilename` is called to contruct a transmit statement for a partition file in `worker_fetch_partition_file`. On the wire, the paths do not have the username suffixes. The suffix is only added before reading or writing a file.

marcocitus:
> Oh, is it? As far as I followed, the ereport(WARNING check relies on the user name being the same? I'll read the code a bit more.

It checks the local file name (`taskFilename`) that was used in `worker_fetch_partition_file`, but that could also use the local OID instead of the name.

src/backend/distributed/worker/worker_partition_protocol.c
onderkalaci: Ok, that makes more sense now. Thanks!

onderkalaci:
@marcocitus are you going to push another commit (name to Oid)? I'll be doing some high-level review until that changes comes in and do my final review / testing after that. Thanks!

src/backend/distributed/worker/worker_partition_protocol.c
```diff
@@ -606,7 +608,14 @@ TaskDirectoryName(uint64 jobId, uint32 taskId)
...
+ * which is suitable for use in worker_fetch_partition_file and tranmsit.
+ *
+ * It excludes the username part at the end of the filename, since that is
```
onderkalaci: Minor: stale comment ` the username part`

src/test/regress/expected/worker_hash_partition.out
```diff
@@ -25,10 +25,10 @@ SELECT worker_hash_partition_table(:JobId, :TaskId, :Select_Query_Text,
  
 (1 row)
 
-COPY :Table_Part_00 FROM 'base/pgsql_job_cache/job_201010/task_101103/p_00000';
-COPY :Table_Part_01 FROM 'base/pgsql_job_cache/job_201010/task_101103/p_00001';
-COPY :Table_Part_02 FROM 'base/pgsql_job_cache/job_201010/task_101103/p_00002';
-COPY :Table_Part_03 FROM 'base/pgsql_job_cache/job_201010/task_101103/p_00003';
+COPY :Table_Part_00 FROM 'base/pgsql_job_cache/job_201010/task_101103/p_00000.postgres';
```
onderkalaci: Ok, you need to find Oid of postgres, which seems to be always `10`?

marcocitus: Made it automatic using `SELECT usesysid AS userid FROM pg_user WHERE usename = 'postgres' \gset`

src/backend/distributed/worker/worker_merge_protocol.c
```diff
@@ -482,14 +487,20 @@ CreateStatement(RangeVar *relation, List *columnDefinitionList)
  * CopyTaskFilesFromDirectory finds all files in the given directory, except for
  * those having an attempt suffix. The function then copies these files into the
  * database table identified by the given schema and table name.
+ *
+ * The function makes sure all files were generated by the current user by checking
```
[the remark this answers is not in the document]
marcocitus: In `worker_merge_files_into_table` this function is called after the current user is changed to superuser, which is necessary for the `DoCopy` to be able to read from a file. That means that `CurrentUserId()` would return the superuser's ID, but that's not what we want.

Not related to this PR, but I did notice some weirdness in this code:
- `worker_merge_files_and_run_query` does not call the function as superuser, which suggests that this code path never worked for non-superusers.
- `worker_append_table_to_shard` has a slightly different implementation for reading files into a table (calling the ProcessUtility from the top)

PR #2492, opened 2018-11-23
Clean up UDFs and revoke unnecessary permissions

marcocitus:
DESCRIPTION: Clean up UDFs and revoke unnecessary permissions

Some general clean-up of (mostly harmless) UDFs that we'd rather not have regular users calling.

### DIFF
--- a/src/backend/distributed/citus--8.0-10--8.0-11.sql
+++ b/src/backend/distributed/citus--8.0-10--8.0-11.sql
@@ -1,0 +1,20 @@
+/* citus--8.0-10--8.0-11 */
+SET search_path = 'pg_catalog';
+
+-- Deprecated functions
+DROP FUNCTION IF EXISTS worker_foreign_file_path(text);
+DROP FUNCTION IF EXISTS worker_find_block_local_path(bigint,text[]);
+DROP FUNCTION IF EXISTS worker_fetch_query_results_file(bigint,integer,integer,text,integer);
+DROP FUNCTION IF EXISTS master_drop_distributed_table_metadata(regclass,text,text);
+REVOKE ALL ON FUNCTION create_insert_proxy_for_table(regclass,regclass) FROM PUBLIC;
+
+-- Testing functions
+REVOKE ALL ON FUNCTION citus_blocking_pids(integer) FROM PUBLIC;
+REVOKE ALL ON FUNCTION citus_isolation_test_session_is_blocked(integer,integer[]) FROM PUBLIC;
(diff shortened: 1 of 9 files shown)


src/backend/distributed/citus--8.0-10--8.0-11.sql
```diff
@@ -0,0 +1,19 @@
+/* citus--8.0-10--8.0-11 */
+SET search_path = 'pg_catalog';
+
+-- Deprecated functions
+DROP FUNCTION IF EXISTS worker_foreign_file_path(text);
```
onderkalaci: Should we drop ` worker_hash_partition_table(bigint, integer, text, text, oid, integer);` as well? That probably requires some more code clean-up as well, so I'd be ok not to do it here. [Reference](https://github.com/citusdata/citus/pull/2075/files#diff-eb0a1157a70a245b3717f65df05279e1R210) for the change.

src/backend/distributed/citus--8.0-10--8.0-11.sql
```diff
@@ -0,0 +1,19 @@
...
+DROP FUNCTION IF EXISTS worker_find_block_local_path(bigint,text[]);
+DROP FUNCTION IF EXISTS worker_fetch_query_results_file(bigint,integer,integer,text,integer);
+DROP FUNCTION IF EXISTS master_drop_distributed_table_metadata(regclass,text,text);
+REVOKE ALL ON FUNCTION create_insert_proxy_for_table(regclass,regclass) FROM PUBLIC;
+
+-- Testing functions
+REVOKE ALL ON FUNCTION check_distributed_deadlocks() FROM PUBLIC;
```
onderkalaci: Should we revoke all for `recover_prepared_transactions();` as well? Because, if I do `BEGIN; SELECT recover_prepared_transactions()` from any user, it blocks whole maintenance daemon due to this [lock](https://github.com/citusdata/citus/blob/master/src/backend/distributed/transaction/transaction_recovery.c#L188) blocking on the maintenance daemon.

src/backend/distributed/citus--8.0-10--8.0-11.sql
marcocitus: Removed it

src/backend/distributed/citus--8.0-10--8.0-11.sql
marcocitus: Makes sense

Issue #2408, opened 2018-09-26
Parallel index creation on distributed tables?


colton-citus:
If you connect to the coordinator using multiple simultaneous connections and issue a series of CREATE INDEX statements on the same table, the indexes will run in series, where index B only starts being created on all workers once index A is finished on all workers. 

The PostgreSQL 10 documentation makes it clear that it is possible to create multiple indexes on the same table in parallel: 

https://www.postgresql.org/docs/current/static/sql-createindex.html 

This issue is being opened to clarify expected behavior and/or request parallel index creation if this is not currently expected behavior.

marcocitus:
`CREATE INDEX` runs in parallel across all shard placements. `CREATE INDEX `CONCURRENTLY` does not, in part because Postgres does not allow `CREATE INDEX CONCURRENTLY` to run in parallel.

We could run `CREATE INDEX CONCURRENTLY` in parallel across nodes, but currently don't due to executor limitations.

Should we rename this issue to "Parallelise CREATE INDEX CONCURRENTLY across nodes"?

colton-citus:
I haven't been able to get CREATE INDEX to run in parallel across shard placements in my testing so far- all the workers have finished creating the first index before starting on the second. I'll do some further testing to make sure that this isn't a problem with my environment and nail down reproduction steps and update shortly. 

I am not using CONCURRENTLY in my testing.

marcocitus:
How are you checking this? I cannot reproduce a case where the CREATE INDEX is not executed parallel.

seamusabshere:
@marcocitus I initially reported this bug by running `CREATE INDEX` on 3 separate connections. On the coordinator, it shows up as them all running.

If you look at the worker nodes, though, only one set of `CREATE INDEX` run at a time.

seamusabshere:
@marcocitus here's an example from literally right now, citus cloud with citus 7.5 and postgres 10.4

each worker node is shown with all of its active queries

<img width="855" alt="screen shot 2018-10-17 at 4 44 51 pm" src="https://user-images.githubusercontent.com/15787/47115438-91203080-d22c-11e8-9f1d-0cb20a40af54.png">

marcocitus:
We had some confusion around the meaning of parallel.

Internally, we tend to use:
Parallel CREATE INDEX - CREATE INDEX is executed in parallel across shards 
Concurrent CREATE INDEX - Multiple CREATE INDEX statements issued by the client concurrently

[...]

PR #2494, opened 2018-11-26
Relax multi-shard modify locks when enable_deadlock_prevention is disabled

marcocitus:
DESCRIPTION: Relax multi-shard modify locks when enable_deadlock_prevention is disabled

Several users have complained about multi-shard commands blocking each other. For example #2408 discusses CREATE INDEX, but other users have encountered the problem with concurrent multi-shard DELETE and INSERT...SELECT and it forced them to change their app even though there was no risk of deadlock in their particular workload.

Multi-shard commands that run in parallel and affect the same rows are likely to form a distributed deadlock. While we can detect these deadlocks, it will cause errors and unlike deadlocks that are caused by the order of statements in a transaction block, they cannot easily be avoided by the application.

This PR couples the lock level with the `citus.enable_deadlock_prevention` flag to allow users to relax the locks and allow concurrent, parallel modifications. If `citus.enable_deadlock_prevention` is enabled (the default), we use ShareUpdateExclusive locks to prevent deadlocks. If disabled, we use RowExclusive locks to allow commands to run concurrently, but if another backend runs a command with the flag enabled then the commands do still block each other.

Note that the `citus.enable_deadlock_prevention` flag was previously used to prevent transaction blocks from expanding to new nodes and deprecated with the introduction of distributed deadlock detection, but since we leave GUCs around and it has the right name I thought it made sense to reuse it.

### DIFF
--- a/src/backend/distributed/executor/multi_router_executor.c
+++ b/src/backend/distributed/executor/multi_router_executor.c
@@ -369,9 +369,21 @@ AcquireExecutorMultiShardLocks(List *taskList)
 			 * In either case, ShareUpdateExclusive has the desired effect, since
 			 * it conflicts with itself and ExclusiveLock (taken by non-commutative
 			 * writes).
+			 *
+			 * However, some users find this too restrictive, so we allow them to
+			 * reduce to a RowExclusiveLock when citus.enable_deadlock_prevention
+			 * is enabled, which lets multi-shard modifications run in parallel as
+			 * long as they all disable the GUC.
 			 */
 
-			lockMode = ShareUpdateExclusiveLock;
+			if (EnableDeadlockPrevention)
+			{
+				lockMode = ShareUpdateExclusiveLock;
+			}
+			else
+			{
+				lockMode = RowExclusiveLock;
+			}
 		}
 		else
 		{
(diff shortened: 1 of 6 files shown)


marcocitus:
> I cannot come-up with simple approach to overcome what I've mind. Do you think is the above a realistic concern or that's not a common scenario?

When considering UPDATE/DELETE/upsert, the behaviour of having to set both GUCs seems correct, e.g. if your modifications are not commutative with other updates then we should definitely take aggressive locks to preserve consistency, regardless of whether or not it's a multi-shard command.

CREATE INDEX is a special case because it is basically always commutative. Maybe we should have a special task type for CREATE INDEX?

onderkalaci:
> I think with this simple PR we're solving majority of the complaints we've heard so far. So, it seems OK to open an issue for that and track separately? It might be a very uncommon case?

Though it'd be good to document the requirement for replication > 1 at somewhere, potentially in the GUC description?

marcocitus:
> I think with this simple PR we're solving majority of the complaints we've heard so far. So, it seems OK to open an issue for that and track separately? It might be a very uncommon case?

Agree, I figured we'll just leave #2408 open after merging this PR.

seamusabshere:
ok so... if i temporary disable `citus.enable_deadlock_prevention`, then will my `CREATE INDEX`es run in parallel?

Issue #2347, opened 2018-08-22
EXPLAIN ANALYZE error while inserting in a distributed or reference table 


louiseGrandjonc:
I'm working with the dataset of the tutorials and while trying to use `EXPLAIN ANALYZE` in `INSERT` queries I ran into errors.

```
postgres=# SELECT * FROM companies WHERE id=5005;
 id | name | image_url | created_at | updated_at | something_unique | country
----+------+-----------+------------+------------+------------------+---------
(0 rows)

postgres=# EXPLAIN ANALYZE INSERT INTO companies (id, name, created_at, updated_at) VALUES (5005, 'citus', NOW(), NOW());
WARNING:  duplicate key value violates unique constraint "companies_pkey_102024"
DETAIL:  Key (id)=(5005) already exists.
WARNING:  could not commit transaction for shard 102024 on any active node
ERROR:  could not commit transaction on any active node
postgres=# SELECT * FROM companies WHERE id=5005;
  id  | name  | image_url |         created_at         |         updated_at         | something_unique | country
------+-------+-----------+----------------------------+----------------------------+------------------+---------
 5005 | citus |           | 2018-08-22 18:51:10.193422 | 2018-08-22 18:51:10.193422 |                  |
(1 row)
```

In this case the row is still inserted.

I ran into a similar error while inserting in a reference table. Here are the steps to reproduce:

```
CREATE TABLE countries (id serial, name varchar(250), CONSTRAINT countries_pkey PRIMARY KEY (id));

INSERT INTO countries (name) VALUES ('France');

select create_reference_table('countries');
```

Here is the error

``` 
postgres=# EXPLAIN ANALYZE INSERT INTO countries (name) VALUES ('Germany');
WARNING:  duplicate key value violates unique constraint "countries_pkey_102104"
DETAIL:  Key (id)=(3) already exists.
WARNING:  prepared transaction with identifier "citus_0_2762_6_67" does not exist
WARNING:  failed to commit transaction on citus_worker_1:5432
WARNING:  failed to roll back prepared transaction 'citus_0_2762_6_67'
HINT:  Run "COMMIT PREPARED 'citus_0_2762_6_67'" on citus_worker_1:5432
                                               QUERY PLAN
--------------------------------------------------------------------------------------------------------
 Custom Scan (Citus Router)  (cost=0.00..0.00 rows=0 width=0) (actual time=0.914..0.914 rows=0 loops=1)
   Task Count: 1
   Tasks Shown: All
   ->  Task
[...]

PR #2495, opened 2018-11-27
{WIP} Change behavior for explaining distributed modify queries

mtuncer:


mtuncer:
@anarazel  did not like this approach because it is prone to other issues like execute, prepared statements etc. Closing this PR in favor of not using explain hook and doing explain analyze during BeginCustomScan and before ExecutorRun

Longer term solution would be more invasive to avoid duplicate execution of explain analyze queries.

mtuncer:
btw this was for https://github.com/citusdata/citus/issues/2347

PR #2496, opened 2018-11-27
Only allow transmit from pgsql_job_cache directory

marcocitus:


src/backend/distributed/worker/worker_partition_protocol.c
```diff
@@ -643,7 +643,7 @@ CacheDirectoryElement(const char *filename)
 	appendStringInfo(directoryPath, "base/%s/", PG_JOB_CACHE_DIR);
 
 	directoryPathFound = strstr(filename, directoryPath->data);
-	if (directoryPathFound != NULL)
+	if (directoryPathFound == filename)
```
marcocitus: this confirms that the `base/pgsql_job_cache/` appears at the start of the string, rather than anywhere in the string

mtuncer:
we could do all those nasty things with pg_read_file. like
```sql
select pg_read_file('/etc/passwd');
```

src/backend/distributed/worker/worker_partition_protocol.c
mtuncer: I would have liked it better if this comment in the code

Issue #1862, opened 2017-12-07
TPC-H query support


marcocitus:
When creating 'lineitem' and 'orders' as distributed tables and the remaining tables as reference tables, the following queries are unsupported:

- Q13 reference table in outer part of outer join
- Q15 non-co-located join across subqueries
- Q17 correlated subquery
- Q18 reference table in FROM when there's a subquery in WHERE
- Q20 reference table in FROM when there's a subquery in WHERE
- Q21 reference table in FROM when there's a subquery in WHERE
- Q22 non-co-located join across subqueries

On the `sql_monster` branch, this changes to:
- Q13 outer join with re-partitioning
- Q17 correlated subquery
- Q18 reference table in FROM when there's a subquery in WHERE
- Q20 reference table in FROM when there's a subquery in WHERE
- Q21 reference table in FROM when there's a subquery in WHERE
- Q22 reference table in FROM when there's a subquery in WHERE

The "reference table in FROM" error is partly due to an overly aggressive for check in `DeferErrorIfUnsupportedSublinkAndReferenceTable`. Fixing that check reduces the unsupported queries on `sql_monster` to:

- Q13 outer join with re-partitioning
- Q17 correlated subquery
- Q20 reference table in FROM when there's a subquery in WHERE
- Q22 reference table in FROM when there's a subquery in WHERE

The remaining queries require improvements to the join order planner (e.g. use join restriction information) and recursive planning.

Q13: 
Join order planner should recognise that `<reference table> LEFT JOIN <distributed table>` requires re-partitioning both tables by the join column.

Alternatively, recursive planning could recursively plan a query of the form `(SELECT .. FROM <reference table> INNER JOIN <distributed table> ..)` and then perform `<reference table> LEFT JOIN <intermediate result>`.

Q17: 
Join order planner should recognise that `lineitem` (a distributed table) and `part` (a reference table) both need to be re-partitioned by `partkey` (semi-join clause) in order to handle the correlated subquery.

Q20: 
Recursive planning should recognise the subquery in the WHERE clause on `part` should be recursively planned because of the additional, correlated subquery in the WHERE clause on `lineitem`

Join order planner should recognise that `lineitem` (a distributed table) and `part`  (a reference table) and `partsupp` (a reference table) all need to be re-partitioned by `partkey` (semi-join clause) in order to handle the join in the correlated subquery.

Q22: 
[...]

PR #2504, opened 2018-11-28
Expand recursive planning for distributed tables and recurring outer/semi joins

onderkalaci:
This PR is a prototype, and requires several improvements before merging. 

It enables good chunk of complex SQL queries that used be not supported. The improvements are mostly around subquery/cte joins with tables and outer joins with recurring tuples (e.g., reference table or intermediate results).

Though I started with a single improvement in my mind, I ended up implementing the second one as well to complement the first one which is by itself a useful feature.

- Expand recursive planning for tables
- Expand recursive planning for recursively planning inner parts of recurring tuple joins

The PR itself may not be performant for many cases where all the distributed table (or even some of the columns) needs to pulled to the coordinator** and pushed back. However, this can be considered a step towards future improvements. This PR, in combination with #2481 and an enhanced version of the implementation in #2305 could enable `Repartition via Coordinator`. Coupled with Citus MX, we might become closer to replace the existing repartitioning support. Though, we should find a way to execute `push` part of `pull & push` in parallel to compete with the current repartitioning implementation. Can we implement `Parallel COPY` in Postgres and use it?

 The coverage of this PR is tightly coupled with #2481, which enables (recurring tuple OUTER JOIN recurring tuple). And, this PR could yield lots of those joins.

 Basic algorithm to implement the changes:

 - Expand recursive planning for tables 

    - Simply convert "table" into `(SELECT required_columns FROM table)`
    - required_columns: The columns that the query refers
    - push all the filters that exists on the table in to the subquery, so actually: `(SELECT required_columns FROM table WHERE filters)`
    - Postgres pushes as much filters as possible, and we use those information to generate the wrapper subquery
    - Finally, call `RecursivelyPlanSubquery()` on the newly created subquery
    - Do this for failed non-colocated joins and failed recurrig tuple checks due to the table

 - Expand recursive planning for recursively planning inner parts of recurring tuple joins
    - During recursive planning, call `DeferredErrorIfUnsupportedRecurringTuplesJoin()`. If the function returns error, we know that there is no other way of executing this query.
    - Thus, get the inner part of the join that led to the error
    - If it is a relation, follow the above algorithm. If it is a subquery, we already know what to do via `RecursivelyPlanSubquery()`.

TODOs:
- [ ] Fix 9.6 regression tests
- [ ] Add more specific tests
- [ ] Probably merge it via two different PRs
- [ ] Improve the algorithm for calculating required_columns. For example, if a column appears only as a filter, we shouldn't pull it, but we do it now.
- [ ] Even though we don't need a column at all, we're currently adding it to the target list like `NULL::int as user_id`. Instead, we could only add the required columns and travers the remaining of query tree to update the references to those columns.
- [ ] We currently ignore FULL OUTER JOINS, which should be simple to add
- [ ] There are two regressions in the tests. Make sure that those are fixed with #2481
- [ ] Improve logging. We currently do not mention about recursive relation planning, which should be mentioned. Otherwirse, it'd be pretty confusing for the users.
- [ ] Restriction eq. codes rely on RTE to be `RTE_SUBQUERY`. However, we've now enough references to `RTE_RELATION`, so make those functions operate on `RTE_RELATION` as well
- [ ] Test with edge cases such as dropped columns, prepared statements, insert..select via coordinator etc.
- [ ] Consider LATERAL or any other correlated query cases some more

Any feedback is welcome!

** luckily we already have 1GB max limit of intermediate results

### DIFF
--- a/src/backend/distributed/planner/distributed_planner.c
+++ b/src/backend/distributed/planner/distributed_planner.c
@@ -61,7 +61,8 @@ static DistributedPlan * CreateDistributedPlan(uint64 planId, Query *originalQue
 											   Query *query, ParamListInfo boundParams,
 											   bool hasUnresolvedParams,
 											   PlannerRestrictionContext *
-											   plannerRestrictionContext);
+											   plannerRestrictionContext,
+											   List *previousSubPlanList);
 static DeferredErrorMessage * DeferErrorIfPartitionTableNotSingleReplicated(Oid
 																			relationId);
 static Node * ResolveExternalParams(Node *inputNode, ParamListInfo boundParams);
@@ -497,7 +498,8 @@ CreateDistributedPlannedStmt(uint64 planId, PlannedStmt *localPlan, Query *origi
 
 	distributedPlan =
 		CreateDistributedPlan(planId, originalQuery, query, boundParams,
-							  hasUnresolvedParams, plannerRestrictionContext);
+							  hasUnresolvedParams, plannerRestrictionContext,
+							  NIL);
 
 	/*
 	 * If no plan was generated, prepare a generic error to be emitted.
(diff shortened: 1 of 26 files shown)


marcocitus:
In terms of [TPC-H](https://github.com/citusdata/citus/issues/1862), this PR adds support for Q13 and Q22.

marcocitus:
> So, as far as I remember, Q21 is definitely not solved by de-correlation, but Q20 is still a candidate because the correlations are eq. operators.

Q21 is already supported if only lineitem and orders are distributed.

saimicrosoft:
@onderkalaci  +1. Customer ran into this in a technical session today. Overall SQL coverage has improved tremendously but still this error comes up often.

onderkalaci:
Some of the codes/tests here are already merged to main branch via: https://github.com/citusdata/citus/pull/4358

The rest is mostly going to be overridden via (or a follow up): https://github.com/citusdata/citus/pull/6219

Issue #2387, opened 2018-09-18
Regression testing sometimes fails randomly 


qytang777:
# Original Description

`regression.diffs`:
```
src/test/regress/expected/multi_utilities_0.out
src/test/regress/expected/multi_utilities.out
-334,21 +334,21
-- send a VACUUM FREEZE after a new row
\c  - - - :master_port
INSERT INTO dustbunnies VALUES(5,'peter');
VACUUM (FREEZE) dustbunnies;
-- verify that relfrozenxid increased
\c - - - :worker_1_port
SELECT relfrozenxid::text::integer > :frozenxid AS frozenperformed FROM pg_class
WHERE oid='dustbunnies_990002'::regclass;
frozen_performed
------------------------+
 -t
+f
(1 row)

-- check there are no nulls in either colum
SELECT attname,null_frac FROM pg_stats
WHERE tablename = 'dustbunnies_990002' ORDER BY attname;
attname | null_frac
---------+----------
age       +            1
id          +            0
name    +            0
```

# On <relative-time datetime="2018-11-21T14:29:26Z" title="Nov 21, 2018, 7:29 AM MST">Nov 21, 2018</relative-time> @onderkalaci commented

I've been re-running the regression tests for a while, and the random failure ratio is pretty high (%40-50).

The following list is sorted with respect to the frequency of the failures:

- [ ] **Failure 1**

Failure automation somehow doesn't continue. I don't think that this is similar with #2300 because this is only happening in  `failure_create_distributed_table_non_empty` test, which is 10th test or so in the list. So, we're already got the port. 
```
test failure_create_distributed_table_non_empty …

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated

```

- [ ] **Failure 2**

The same test above. So, this is a very suspicious test.

```SQL
/home/travis/build/citusdata/citus/src/test/regress/results/failure_create_distributed_table_non_empty.out	2018-11-19 11:05:36.449666424 +0000 (count 2)
@@ -595,25 +595,24 @@
SELECT run_command_on_workers($$SELECT count(*) FROM information_schema.tables WHERE table_schema = 'create_distributed_table_non_empty_failure' and table_name LIKE 'test_table%'$$);
 run_command_on_workers
------------------------
 (localhost,9060,t,0)
[...]

PR #2506, opened 2018-11-29
Fix explain outputs on some tests

onderkalaci:
In the scope of #2387.

In general, I thought adding `VACUUM ANALYZE ` should fix non-consistent `EXPLAIN` outputs. 

However, I realized that we've a random failure which is executed just after `VACUUM ANALYZE` as this [one](https://github.com/citusdata/citus/blob/master/src/test/regress/expected/multi_view.out#L797-L823). 

So, we should probably find a better solution.

I think there is no harm merging this in since it decreases the frequency of these issues.


onderkalaci:
Ok, this doesn't solve the problem, closing it for a better solution.

Issue #2219, opened 2018-06-13
Feature request: GROUPING SETS, CUBE, ROLLUP


colton-citus:
This issue is being created to request support for following features: 

GROUPING SETS
CUBE
ROLLUP

See ticket # 593 for original request

pastures:
This has been here for a while, any progress on fixing it?

cooltnt:
hope the distributed table can support

PR #2507, opened 2018-11-29
Error out when a subquery has grouping set clause

### LINKED ISSUE #2256: Grouping sets in pushdownable subqueries incorrectly allowed
GROUPING SETS are not supported (see #2219), but they can incorrectly pass through in subqueries.

Top-level grouping sets error out:

```sql
SELECT x, y FROM test GROUP BY GROUPING SETS ((x), (y));
ERROR:  could not run distributed query with GROUPING SETS, CUBE, or ROLLUP
```

but subqueries may give incorrect results:
```sql 
INSERT INTO test SELECT s, 1 FROM generate_series(1,10) s;                                                                                                             
SELECT * FROM (SELECT x, y FROM test GROUP BY GROUPING SETS ((x), (y))) s;
 x  │ y 
────┼───
  8 │  
    │ 1
  1 │  
    │ 1
 10 │  
    │ 1
  5 │  
    │ 1
  4 │  
  7 │  
    │ 1
  3 │  
    │ 1
  6 │  
    │ 1
  2 │  
    │ 1
  9 │  
    │ 1
(19 rows)
```

The correct result has 11 rows:
```sql
CREATE TABLE local AS SELECT * FROM test;                                                                                                                               
SELECT * FROM (SELECT x, y FROM local GROUP BY GROUPING SETS ((x), (y))) s;                                                                                             
 x  │ y 
────┼───
  9 │  
  3 │  
  5 │  
  4 │  
 10 │  
  6 │  
  2 │  
  7 │  
  1 │  
  8 │  
    │ 1
(11 rows)
```

hanefi:
Description: Do not allow GROUPING SET clauses in subqueries

We currently do not support GROUPING SETS (see #2219) but they can incorrectly pass through in subqueries.

Closes #2256

### DIFF
--- a/src/test/regress/expected/subqueries_not_supported.out
+++ b/src/test/regress/expected/subqueries_not_supported.out
@@ -157,6 +157,17 @@ WHERE
 ;
 ERROR:  cannot push down this subquery
 DETAIL:  Having qual without group by on partition column is currently unsupported when a subquery references a column from another query
+-- We do not support GROUPING SETS in subqueries
+-- This also includes ROLLUP or CUBE clauses
+SELECT * FROM (SELECT user_id, value_1 FROM users_table GROUP BY GROUPING SETS ((user_id), (value_1))) s;
+ERROR:  could not run distributed query with GROUPING SETS, CUBE, or ROLLUP
+HINT:  Consider using an equality filter on the distributed table's partition column.
+SELECT * FROM (SELECT user_id, value_1 FROM users_table GROUP BY ROLLUP (user_id, value_1)) s;
+ERROR:  could not run distributed query with GROUPING SETS, CUBE, or ROLLUP
+HINT:  Consider using an equality filter on the distributed table's partition column.
+SELECT * FROM (SELECT user_id, value_1 FROM users_table GROUP BY CUBE (user_id, value_1)) s;
+ERROR:  could not run distributed query with GROUPING SETS, CUBE, or ROLLUP
+HINT:  Consider using an equality filter on the distributed table's partition column.
 SET client_min_messages TO DEFAULT;
 DROP SCHEMA not_supported CASCADE;
 NOTICE:  drop cascades to table users_table_local
--- a/src/backend/distributed/planner/query_pushdown_planning.c
+++ b/src/backend/distributed/planner/query_pushdown_planning.c
@@ -807,6 +807,14 @@ DeferErrorIfCannotPushdownSubquery(Query *subqueryTree, bool outerMostQueryHasLi
 		}
 	}
 
+	/* grouping sets are not allowed in subqueries*/
+	if (subqueryTree->groupingSets)
+	{
+		preconditionsSatisfied = false;
+		errorDetail = "could not run distributed query with GROUPING SETS, CUBE, "
+					  "or ROLLUP";
+	}
+
 	/*
 	 * We support window functions when the window function
 	 * is partitioned on distribution column.
--- a/src/test/regress/sql/subqueries_not_supported.sql
+++ b/src/test/regress/sql/subqueries_not_supported.sql
@@ -149,6 +149,13 @@ WHERE
     )
 ;
 
+
+-- We do not support GROUPING SETS in subqueries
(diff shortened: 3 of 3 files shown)


marcocitus:
Needs a regression test.

Something worth thinking about: Which grouping sets are actually acceptable in subquery pushdown?

hanefi:
I know that grouping sets are supported on single shard, but currently I do not care whether a subquery hits a single shard or not. Other than checking for the number of shard hits, nothing comes to my mind that would make subquery pushdown acceptable

I am trying to write some GROUPING SET queries that give wrong results on master instead of returning errors. I planned to reuse an existing table in a regression test for this

I will add regression tests to `multi_router_planner.sql` today.

marcocitus:
> I know that grouping sets are supported on single shard, but currently I do not care whether a subquery hits a single shard or not. 

The test is to confirm that we actually throw the error and don't accidentally remove the error.

> Other than checking for the number of shard hits, nothing comes to my mind that would make subquery pushdown acceptable

For example, the following seems perfectly pushdownable: `SELECT x, y, count(*) FROM test GROUP BY GROUPING SETS ((x), (x,y));` . I guess the rule is that every set needs to contain the distribution column.

Not necessarily for this PR, but it's worth understanding what we could/should be doing. 

> I will add regression tests to multi_router_planner.sql today.

No need, we already have those.

src/test/regress/expected/subqueries_not_supported.out
```diff
@@ -157,6 +157,17 @@ WHERE
 ;
 ERROR:  cannot push down this subquery
 DETAIL:  Having qual without group by on partition column is currently unsupported when a subquery references a column from another query
+-- We do not support GROUPING SETS in subqueries
+-- This also includes ROLLUP or CUBE clauses
+SELECT * FROM (SELECT user_id, value_1 FROM users_table GROUP BY GROUPING SETS ((user_id), (value_1))) s;
```
onderkalaci: These tests are good for now. But, I've a note here for when you touch similar parts of the code in the future or if you have seen `random()` calls in the tests.

In general, PostgreSQL has the ability to pull up subqueries. So. say we have the following query:
```SQL
SELECT * FROM (SELECT user_id FROM users_table) as foo;
```

It simply converts (e.g., pulls-up the subquery) the query to the following and plans in that form for various performance reasons:
```SQL
SELECT user_id FROM users_table;
```

For that to happen, the subquery should conform to the rules in [`is_simple_subquery()`](https://doxygen.postgresql.org/prepjointree_8c_source.html#l01442).

As you can see, if `subquery->groupingSets` the function returns false, so we're safe in these tests. 

But, in general, to ensure that Postgresql doesn't do that, we add a `random()` function call since there is the following check in the subquery, and that is less likely to change in the future:

```C
 if (contain_volatile_functions((Node *) subquery->targetList))
         return false;
```

So, if I were to write these queries, I'd ended-up with the following to prevent subquery being pulled:
```SQL

SELECT * FROM (SELECT random() * user_id, value_1 FROM users_table GROUP BY GROUPING SETS ((user_id), (value_1))) s;

```

hanefi:
> I guess the rule is that every set needs to contain the distribution column.

I have one small thing to add to this. We discussed which queries we can safely pushdown with @onderkalaci . I can say that it would be better if we say every *nonempty* group needs to contain distribution column.

There is a strange case where we can possibly pushdown the query even though there is an empty group (therefore not including the distribution column):
```
CUBE ((x,y), (x,z));
```
corresponds to these 4 groups in a grouping set:
```
GROUPING SETS (
    (x, y, z),
    (x, y,  ),
    (x,    z),
    (       )
)
```
The first three groups will hit one shard, but the empty group will need to process all the shards AFAIK. I assume that it will still be pushdownable though.

src/test/regress/expected/subqueries_not_supported.out
hanefi: I read the tests in this file to get a general understanding, but I didn't pay attention to random number generation in subqueries.

It is really nice for you to point these things out as it helps me learn how Postgres internals work.
Thanks

PR #2509, opened 2018-11-30
Make sure the explain output for partition wise join is stable

onderkalaci:
Related with #2387

We disable bunch of planning options on the workers. This might be
risky if any concurrent test relies on EXPLAIN OUTPUT as well. Still,
we want to keep this test, so we should try to not parallelize this
test with such test.

### DIFF
--- a/src/test/regress/sql/multi_partitioning.sql
+++ b/src/test/regress/sql/multi_partitioning.sql
@@ -986,7 +986,6 @@ ORDER BY
 COMMIT;
 
 -- test partition-wise join
-
 CREATE TABLE partitioning_hash_join_test(id int, subid int) PARTITION BY HASH(subid);
 CREATE TABLE partitioning_hash_join_test_0 PARTITION OF partitioning_hash_join_test FOR VALUES WITH (MODULUS 3, REMAINDER 0);
 CREATE TABLE partitioning_hash_join_test_1 PARTITION OF partitioning_hash_join_test FOR VALUES WITH (MODULUS 3, REMAINDER 1);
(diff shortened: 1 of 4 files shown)


src/test/regress/sql/multi_partitioning.sql
```diff
@@ -986,16 +986,21 @@ ORDER BY
...
 CREATE TABLE partitioning_hash_join_test(id int, subid int) PARTITION BY HASH(subid);
 CREATE TABLE partitioning_hash_join_test_0 PARTITION OF partitioning_hash_join_test FOR VALUES WITH (MODULUS 3, REMAINDER 0);
 CREATE TABLE partitioning_hash_join_test_1 PARTITION OF partitioning_hash_join_test FOR VALUES WITH (MODULUS 3, REMAINDER 1);
 CREATE TABLE partitioning_hash_join_test_2 PARTITION OF partitioning_hash_join_test FOR VALUES WITH (MODULUS 3, REMAINDER 2);
 
 SELECT create_distributed_table('partitioning_hash_join_test', 'id');
 
--- see the query plan without partition-wise join
-EXPLAIN
+SELECT success FROM run_command_on_workers('alter system set enable_mergejoin to off');
```
mtuncer: These are really limiting postgres planner's options.

mtuncer:
We discussed this offline with @onderkalaci I am happy to keep this test as opposed to removing it altogether.

🚢

src/test/regress/sql/multi_partitioning.sql
onderkalaci: That's my goal 😃

jasonmp85: I mean it's not like we're doing this in our code, just in the test… seems fine to me.

PR #2510, opened 2018-11-30
Sort task list for multi-task explain outputs

onderkalaci:
part of #2387

This is purely for ensuring that regression tests do not randomly fail.

### DIFF
--- a/src/backend/distributed/planner/multi_explain.c
+++ b/src/backend/distributed/planner/multi_explain.c
@@ -23,6 +23,7 @@
 #include "distributed/citus_nodefuncs.h"
 #include "distributed/connection_management.h"
 #include "distributed/insert_select_planner.h"
+#include "distributed/listutils.h"
 #include "distributed/multi_client_executor.h"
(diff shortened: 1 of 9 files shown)


jasonmp85:
The failures here are weird and related to not finding HLL in packagecloud… Looking into it.

src/backend/distributed/planner/multi_explain.c
```diff
@@ -358,6 +359,9 @@ ExplainTaskList(List *taskList, ExplainState *es)
 	ListCell *remoteExplainCell = NULL;
 	List *remoteExplainList = NIL;
 
+	/* make sure that the output is consistent */
+	taskList = SortList(taskList, CompareTasksByTaskId);
```
jasonmp85: This is contingent upon the planning and `taskId` assignment being deterministic, but if it is, this is an improvement over the status quo, so I'm fine merging this without further inspection (AFAIK there is not a strong reason to assume task id assignment is ever non-deterministic, but just bringing it up).

PR #2512, opened 2018-11-30
Add a column to pg_dist_node to enable metadata-only workers

marcocitus:
DESCRIPTION: Add a column to pg_dist_node to enable metadata-only workers

This PR adds an `isdatanode` column to `pg_dist_node`, which allows you to specify whether a node should have shards created on it (default is `true`). This opens the door for having metadata-only/query nodes.

I did not add a more official API for it yet. I imagine it would be something like `add_query_node`. We can use this to gain some experience with it and figure out naming before adding more official support.

(probably not an 8.1 item)

### DIFF
--- a/citus.control
+++ b/citus.control
@@ -1,6 +1,6 @@
 # Citus extension
 comment = 'Citus distributed database'
-default_version = '8.0-12'
+default_version = '8.0-13'
 module_pathname = '$libdir/citus'
 relocatable = false
 schema = pg_catalog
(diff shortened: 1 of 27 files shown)


pykello:
@marcocitus Is this a 9.0 item? If so, then part of this and https://github.com/citusdata/citus/pull/2928 can be done together to avoid altering pg_dist_node multiple times.

marcocitus:
@JelteF would this column be compatible with drain node?

JelteF:
@marcocitus thanks for the ping on this. I'll reuse part of this in my next PR.

JelteF:
Closing this since my next PR will incorporate this as well.

PR #2515, opened 2018-12-03
Use `conn.killall()` instead of `conn.kill()` in the failure tests

onderkalaci:
Mostly opened to trigger travis failure tests, yet.

There seems to be an issue with `flow.kill`, which is fixed in never versions of mitmproxy but not the one that we use. Here is the link to that [issue](https://github.com/mitmproxy/mitmproxy/issues/2724).

I realized Citus regression tests randomly fails on right after `conn.kill()`, and trying to understand whether the change in this PR helps us.


onderkalaci:
this doesn't fix the issue

PR #2516, opened 2018-12-03
Review some strcpy/memcpy/strprintf/sscanf usage in the code.

DimCitus:
In answer to a security audit, we double check buffer sizes and avoid known-dangerous operations such as sscanf.

### DIFF
--- a/src/backend/distributed/connection/connection_configuration.c
+++ b/src/backend/distributed/connection/connection_configuration.c
@@ -260,15 +260,37 @@ GetConnParams(ConnectionHashKey *key, char ***keywords, char ***values,
 	/* auth keywords will begin after global and runtime ones are appended */
 	Index authParamsIdx = ConnParams.size + lengthof(runtimeKeywords);
 
+	int paramIndex = 0;
+	int runtimeParamIndex = 0;
+
+	if (ConnParams.size + lengthof(runtimeKeywords) > ConnParams.maxSize)
+	{
+		/* unexpected, intended as developers rather than users */
+		ereport(ERROR, (errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+						errmsg("too many connParams entries")));
+	}
+
 	pg_ltoa(key->port, nodePortString); /* populate node port string with port */
 
 	/* first step: copy global parameters to beginning of array */
-	memcpy(connKeywords, ConnParams.keywords, ConnParams.size * sizeof(char *));
-	memcpy(connValues, ConnParams.values, ConnParams.size * sizeof(char *));
+	for (paramIndex = 0; paramIndex < ConnParams.size; paramIndex++)
+	{
+		/* copy the keyword&value pointers to the new array */
+		connKeywords[paramIndex] = ConnParams.keywords[paramIndex];
+		connValues[paramIndex] = ConnParams.values[paramIndex];
+	}
 
 	/* second step: begin at end of global params and copy runtime ones */
-	memcpy(&connKeywords[ConnParams.size], runtimeKeywords, sizeof(runtimeKeywords));
-	memcpy(&connValues[ConnParams.size], runtimeValues, sizeof(runtimeValues));
+	for (runtimeParamIndex = 0;
+		 runtimeParamIndex < lengthof(runtimeKeywords);
+		 runtimeParamIndex++)
+	{
+		/* copy the keyword&value pointers to the new array */
+		connKeywords[ConnParams.size + runtimeParamIndex] =
+			(char *) runtimeKeywords[runtimeParamIndex];
+		connValues[ConnParams.size + runtimeParamIndex] =
+			(char *) runtimeValues[runtimeParamIndex];
+	}
 
 	/* final step: add terminal NULL, required by libpq */
 	connKeywords[authParamsIdx] = connValues[authParamsIdx] = NULL;
--- a/src/backend/distributed/relay/relay_event_utility.c
+++ b/src/backend/distributed/relay/relay_event_utility.c
@@ -692,12 +692,12 @@ void
 AppendShardIdToName(char **name, uint64 shardId)
 {
 	char extendedName[NAMEDATALEN];
-	uint32 extendedNameLength = 0;
 	int nameLength = strlen(*name);
 	char shardIdAndSeparator[NAMEDATALEN];
 	int shardIdAndSeparatorLength;
 	uint32 longNameHash = 0;
 	int multiByteClipLength = 0;
+	int neededBytes = 0;
 
 	if (nameLength >= NAMEDATALEN)
 	{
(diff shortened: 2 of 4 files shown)


src/backend/distributed/connection/connection_configuration.c
```diff
@@ -260,15 +260,30 @@ GetConnParams(ConnectionHashKey *key, char ***keywords, char ***values,
...
 
+	int paramIndex = 0;
+	int runtimeParamIndex = 0;
+
 	pg_ltoa(key->port, nodePortString); /* populate node port string with port */
 
 	/* first step: copy global parameters to beginning of array */
-	memcpy(connKeywords, ConnParams.keywords, ConnParams.size * sizeof(char *));
-	memcpy(connValues, ConnParams.values, ConnParams.size * sizeof(char *));
+	for (paramIndex = 0; paramIndex < ConnParams.size; paramIndex++)
```
mtuncer: just asking, how different is this from previous code. I don't see any difference wrt security here.

src/backend/distributed/relay/relay_event_utility.c
```diff
@@ -751,11 +751,20 @@ AppendShardIdToName(char **name, uint64 shardId)
 				 SHARD_NAME_SEPARATOR, longNameHash,
 				 shardIdAndSeparator);
 	}
-	extendedNameLength = strlen(extendedName) + 1;
-	Assert(extendedNameLength <= NAMEDATALEN);
 
-	(*name) = (char *) repalloc((*name), extendedNameLength);
-	snprintf((*name), extendedNameLength, "%s", extendedName);
+	(*name) = (char *) repalloc((*name), NAMEDATALEN);
+	writtenBytes = snprintf((*name), NAMEDATALEN, "%s", extendedName);
```
mtuncer: naming suggest it was returning actually written bytes, but it is not.
something along neededBytes, actualLength, requestedBytes ...  would look better

src/backend/distributed/connection/connection_configuration.c
marcocitus: There's no known security issue here, but there's a list of functions like sscanf and memcpy that are regarded unsafe. That doesn't mean we need to fix every instance (postgres is full of them), but the memcpy calls in this function were marked as suspicious, most likely because of their use of global variables (which an attacker might be able to control if there's another issue).

src/backend/distributed/connection/connection_configuration.c
```diff
@@ -260,15 +260,30 @@ GetConnParams(ConnectionHashKey *key, char ***keywords, char ***values,
...
 
 	/* second step: begin at end of global params and copy runtime ones */
-	memcpy(&connKeywords[ConnParams.size], runtimeKeywords, sizeof(runtimeKeywords));
-	memcpy(&connValues[ConnParams.size], runtimeValues, sizeof(runtimeValues));
+    for (runtimeParamIndex = 0;
+         runtimeParamIndex < lengthof(runtimeKeywords);
+         runtimeParamIndex++)
+    {
+        /* copy the keyword&value pointers to the new array */
+        connKeywords[ConnParams.size + runtimeParamIndex] =
```
marcocitus: In the spirit of these changes, should we add some ConnParams.size < ConnParams.maxSize check somewhere?

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -1321,21 +1321,86 @@ Assign2PCIdentifier(MultiConnection *connection)
...
 {
-	const int expectedFieldCount = 4;
-	int parsedFieldCount = 0;
-	bool nameValid = false;
+	char *ptr = preparedTransactionName;
 
-	parsedFieldCount = sscanf(preparedTransactionName, PREPARED_TRANSACTION_NAME_FORMAT,
-							  groupId, procId, transactionNumber, connectionNumber);
-	if (parsedFieldCount == expectedFieldCount)
+	for (int i=0; i<4; i++)
```
mtuncer: shall we drop the loop and do it in sequential manner.
looks like loop is only saving us from repeating 
```c
        ptr = strchr(ptr, '_');
 		if (ptr == NULL)
		{
			return false;
		}
 		/* step ahead of the current '_' character */
		++ptr;
```

But reducing readability significantly.

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -1321,21 +1321,86 @@ Assign2PCIdentifier(MultiConnection *connection)
...
+					|| (*procId == INT_MAX && errno == ERANGE))
+				{
+					return false;
+				}
+				break;
+			}
+
+			case 2:
+			{
+				*transactionNumber = strtoull(ptr, NULL, 10);
```
mtuncer: can we use `pg_strtouint64` instead

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -1321,21 +1321,86 @@ Assign2PCIdentifier(MultiConnection *connection)
...
+		}
+
+		/* step ahead of the current '_' character */
+		++ptr;
+
+		switch (i)
+		{
+			case 0:
+			{
+				*groupId = strtol(ptr, NULL, 10);
```
mtuncer: pg_atoi ?

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -1321,21 +1321,86 @@ Assign2PCIdentifier(MultiConnection *connection)
...
+					|| (*groupId == INT_MAX && errno == ERANGE))
+				{
+					return false;
+				}
+				break;
+			}
+
+			case 1:
+			{
+				*procId = strtol(ptr, NULL, 10);
```
mtuncer: pg_atoi

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -1321,21 +1321,86 @@ Assign2PCIdentifier(MultiConnection *connection)
...
+				{
+					return false;
+				}
+				break;
+			}
+
+			case 3:
+			{
+				uint64_t ulNumber = 0;
+				ulNumber = strtoul(ptr, NULL, 10);
```
mtuncer: pg_strtouint64

src/backend/distributed/transaction/remote_transaction.c
mtuncer: may be not, we probably don't want to fail with error. If erroring out is fine please use pg_atoi

src/backend/distributed/transaction/remote_transaction.c
DimCitus: Nope, we want to avoid `ereport(ERROR ...` on bad input. If we fail to parse the prepared transaction name it just means it's not a Citus _internal_ one, and we have another processing ahead.

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -1321,21 +1322,84 @@ Assign2PCIdentifier(MultiConnection *connection)
...
+
+	ptr = strchr(ptr, '_');
+	if (ptr == NULL)
+	{
+		return false;
+	}
+	/* step ahead of the current '_' character */
+	++ptr;
+
+	if ((*transactionNumber == 0 && errno != 0)
```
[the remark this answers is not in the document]
DimCitus: Oops, fixing.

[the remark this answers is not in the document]
DimCitus: Fixed now.

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -1321,21 +1322,79 @@ Assign2PCIdentifier(MultiConnection *connection)
...
+	{
+		return false;
+	}
+
+	/* step ahead of the current '_' character */
+	++currentCharPointer;
+
+	*procId = strtol(currentCharPointer, NULL, 10);
+	if ((*procId == 0 && errno == EINVAL) ||
+		(*procId == INT_MAX && errno == ERANGE))
```
pykello: Sorry, late here. If we are checking for `errno` anywhere, we should reset it before making the function calls. Looking at `strtol`'s specification and its implementation, it doesn't change `errno` if it succeeds.

src/backend/distributed/connection/connection_configuration.c
jasonmp85: Yup, no difference with security, huge difference with subtle bug that was introduced and took time to find and fix.

jasonmp85: Anyways, this is the root cause of #2611

jasonmp85: Ok, I may have been premature and should have taken my own advice here: https://github.com/citusdata/citus/pull/2611#issuecomment-473487391

Still digging…

jasonmp85: Yup, I thought you replaced a `memcpy` of the _strings_ with just directly copying the pointers to the strings; didn't realize the `memcpy` of an array of pointers was just replaced by a loop copying the pointers. No problem there… it's my code.

PR #2521, opened 2018-12-04
Simplify make file for citus sql files

thanodnl:
Adding a change to the citus schema currently requires the copying of two lines substituting some version numbers and adding the new version to a list of all versions.

This change aims to simplify the addition of file changing the schema to only adding the new version to the list of all versions in which order it needs to appear. This removes the necessity of copying and changing the lines responsible for concatenating the sql files.

How it actually works is less trivial then I'd liked, but trying the same with a rule looking up the previous version based on a name caused make to segfault 😞 .

Assigning @marcocitus as he was positive on opening a PR on this. Feel free to unassign if you don't have the bandwidth or if you think someone else is better suited to review.


DimCitus:
It seems non-idiomatic in a Makefile to have an explicit loop to generate rules that way. I wonder if we can obtain our effect with either https://www.gnu.org/software/make/manual/make.html#Secondary-Expansion or some other technique?

marcocitus:
Since we dropped 9.6 support, I removed all the auto-generated SQL files and renamed `citus.sql` to `citus--5.0.sql`.

PR #2523, opened 2018-12-06
Adds support for disabling hash agg with hll functions on coordinator 

velioglu:
Checks whether the master query created by the Citus contains an hll aggregate function. We need to implement it in this way instead of using hooks provided by Postgres, since we create the query for the coordinator from scratch.


src/backend/distributed/planner/multi_master_planner.c
```diff
@@ -313,8 +384,12 @@ BuildAggregatePlan(Query *masterQuery, Plan *subPlan)
 		 * in group and order by with aggregate operations.
 		 * see nodeAgg.c:build_pertrans_for_aggref(). In that case we use
 		 * sorted agg strategy, otherwise we use hash strategy.
+		 *
+		 * If the master query contains hll aggregate functions and the client set
+		 * hll.enable_hashagg to off, then we choose to use group aggregation.
 		 */
-		if (!enable_hashagg || !groupingIsHashable || hasDistinctAggregate)
+		if (!enable_hashagg || !groupingIsHashable || hasDistinctAggregate ||
+			UseGroupAggregateWithHLL(masterQuery))
```
mtuncer: I would rather see this next to  `groupingIsHashable` and `hasDistinctAggregate` calculations.

src/backend/distributed/planner/multi_master_planner.c
```diff
@@ -244,6 +250,71 @@ BuildSelectStatement(Query *masterQuery, List *masterTargetList, CustomScan *rem
...
+ * QueryContainsAggregateWithHLL returns true if the query has an hll aggregate
+ * function in it's target list.
+ */
+static bool
+QueryContainsAggregateWithHLL(Query *query)
+{
+	List *varList = NIL;
+	ListCell *varCell = NULL;
+
+	varList = pull_var_clause((Node *) query->targetList, PVC_INCLUDE_AGGREGATES);
```
mtuncer: could you make sure hll functions in having clause are handled here

src/backend/distributed/planner/multi_master_planner.c
```diff
@@ -244,6 +250,71 @@ BuildSelectStatement(Query *masterQuery, List *masterTargetList, CustomScan *rem
 }
 
 
+/*
+ * QueryContainsAggregateWithHLL returns true if the query has an hll aggregate
+ * function in it's target list.
+ */
+static bool
+QueryContainsAggregateWithHLL(Query *query)
```
mtuncer: Our function definition logic is that
1 - we declare the function
2 - we use the function 
3 - we provide the implementation of the function at the first available place.

Thus function definition order should look like
```
BuildAggregatePlan
HasDistinctAggregate
UseGroupAggregateWithHLL
QueryContainsAggregateWithHLL
```

src/backend/distributed/planner/multi_master_planner.c
```diff
@@ -244,6 +250,71 @@ BuildSelectStatement(Query *masterQuery, List *masterTargetList, CustomScan *rem
...
+	const char *gucStrValue = NULL;
+
+	/* If HLL extension is not loaded, return false */
+	if (hllId == InvalidOid)
+	{
+		return false;
+	}
+
+	/* If HLL is loaded but related GUC is not set, return false */
+	gucStrValue = GetConfigOption(HLL_ENABLE_HASH_GUC_NAME, false, false);
```
mtuncer: Can we have hll extension defined in a different schema and still has the guc with hll prefix ?

marcocitus: Should be fine. GUC prefixes are unrelated to schemas.

src/include/distributed/multi_logical_optimizer.h
```diff
@@ -38,6 +38,7 @@
 #define HLL_ADD_AGGREGATE_NAME "hll_add_agg"
 #define HLL_UNION_AGGREGATE_NAME "hll_union_agg"
 #define HLL_CARDINALITY_FUNC_NAME "hll_cardinality"
+#define HLL_ENABLE_HASH_GUC_NAME "hll.enable_hashagg"
```
byucesoy: I think the name enable_hashagg can be confused with PostgreSQL's enable_hashagg, We wouldn't want to explain all users the difference between the enable_hashagg and hll.enable_hashagg. Can we name it something else; force_groupagg maybe?

src/backend/distributed/planner/multi_master_planner.c
velioglu: Added in that way to not decrease the performance of aggregates without hll type

PR #2553, opened 2018-12-17
Citus 8.1.0 changelog 1545053355

hanefi:
I want to make sure Changelog has all the necessary items. 

I :
- retrieved all PR descriptions
- shortened them when necessary
- dropped 2 (Fixing failing tests & Refactoring)
- reordered them


onderkalaci:
what is `1545053355 ` in the title?

CHANGELOG.md
```diff
@@ -1,3 +1,39 @@
+### citus v8.1.0 (December 17, 2018) ###
+
+* Turns on ssl by default for new installations of citus
+
+* Restricts SSL Ciphers to TLS1.2 and above
+
+* Adds support for INSERT INTO SELECT..ON CONFLICT/RETURNING via coordinator
+
+* Adds support for round-robin task_assignment_policy for queries
```
onderkalaci: This item should include `reference tables`, because this is already a Citus feature, we're only expanding it for reference tables. If we don't have it, it'd be confusing.

Maybe: `Support round-robin task assignment for queries to reference tables.`

CHANGELOG.md
```diff
@@ -1,3 +1,39 @@
...
+
+* Cleans up UDFs and revokes unnecessary permissions
+
+* Improves security in task-tracker protocol
+
+* Improves permission checks in internal DROP TABLE functions
+
+* Improves permission checks in cluster management functions
+
+* Fixes permission checks
```
onderkalaci: This sounds too generic. Maybe use the title itself? `Fix permissions checks in lesser-known UDFs`

CHANGELOG.md
```diff
@@ -1,3 +1,39 @@
+### citus v8.1.0 (December 17, 2018) ###
```
onderkalaci: Maybe add `* Fixes a bug that could cause maintenance daemon panic` which we've packported to 8.0.2, but merged in 8.1.0 as well. (@velioglu pointed that we should add these type of patches to here as well.)

onderkalaci: Also, maybe add an entry for https://github.com/citusdata/citus/pull/2496.

`Only allow transmit from pgsql_job_cache directory`

CHANGELOG.md
hanefi: I needed to shorten the description, and your _shortened_ version makes much more sense here

CHANGELOG.md
hanefi: The original description was `Fixes permission checks in mark_tables_colocated, upgrade_to_reference_table, lock_relation_if_exists` which is way too long. I am using your idea instead

hanefi: Should I merge these two and say `Cleans up UDFs and fixes permissions` ? What do you think @onderkalaci ?

velioglu: Merging them into one makes sense

velioglu:
Make sure you add them as a single commit

PR #2559, opened 2018-12-19
Execute CREATE INDEX CONCURRENTLY in parallel

marcocitus:
DESCRIPTION: Execute CREATE INDEX CONCURRENTLY in parallel

As of https://github.com/postgres/postgres/commit/1dec82068b3b59b621e6b04040c150241f5060f3  the risk of deadlocks when running multiple CREATE INDEX CONCURRENTLY commands in parallel is gone. Switch to using parallel execution to improve performance.


jasonmp85:
Looks good to me! :shipit:

PR #2564, opened 2018-12-21
Speed-up run_command_on_shards().

pykello:
We were establishing connections synchronously. Establishing
connections asynchronously results in some parallelization, saving
hundreds of milliseconds.

In a test I did, this decreased the query time from 200-300ms to 110-150ms.

DESCRIPTION: Speed-up run_command_on_shards().


src/backend/distributed/master/master_citus_tools.c
```diff
@@ -229,28 +229,37 @@ ExecuteCommandsInParallelAndStoreResults(StringInfo *nodeNameArray, int *nodePor
 		palloc0(commmandCount * sizeof(MultiConnection *));
 	int finishedCount = 0;
 
-	/* establish connections */
+	/* start connections asynchronously */
 	for (commandIndex = 0; commandIndex < commmandCount; commandIndex++)
 	{
 		char *nodeName = nodeNameArray[commandIndex]->data;
 		int nodePort = nodePortArray[commandIndex];
 		int connectionFlags = FORCE_NEW_CONNECTION;
```
mtuncer: while you are at it. Could you also check if  we can safely re-use existing connections, or let the connection stay open afterwards.

this could enable setting session level gucs at workers

pykello: I talked with @onderkalaci about this. Making this session-wide will encourage people to use it more, and `run_command_on_shards` doesn't work well with transactional semantics, deadlock detection, etc.

Also since we are unaware what is sent as the command, this will break `FOR_DML`, `FOR_DDL`, etc.

mtuncer:
looks good but I am not convinced that we needed the performance. These functions exist to enable users to perform utility tasks that are not supported in citus. We did not even provide (back then) documentation for those to not to promote widespread use.

marcocitus:
@mtuncer I think Hadi has a very specific use case in mind that justifies this change.

PR #2565, opened 2018-12-21
Git ignore LLVM bitcode files.

pykello:
Compiling with `--with-llvm` results in lots of `*.bc` files which should be ignored when committing to repository.


jasonmp85:
Oops! I use an out-of-tree build so didn't realize the need for this during PG11 development and testing.

PR #2568, opened 2018-12-24
Make sure spinlock is not left unreleased when an exception is thrown

mtuncer:
DESCRIPTION: Fix maintenance daemon panic due to unreleased spinlock

A spinlock is not released when an exception is thrown after spinlock is acquired. This has caused infinite wait and eventual crash in maintenance daemon.

This work moves the code than can fail to the outside of spinlock scope so that in the case of failure spinlock is not left locked since it was not locked in the first place.


mtuncer:
If we don't want to backport, I prefer we do it with a separate PR. 
I don't like PRs with multiple commits doing similar things. In this case we would need to coordinate which commit needs to be backported and which not.

onderkalaci:
> If we don't want to backport, I prefer we do it with a separate PR.
I don't like PRs with multiple commits doing similar things. In this case we would need to coordinate which commit needs to be backported and which not.

That works too.

mtuncer:
I will do it (PR) right away. I have already did some checks. It should be a quick one

onderkalaci:
> @onderkalaci why we don't want to backport other commits?

Three things mainly: 

(a) Those are really minor things, which are only possible if the user calls two external `UDF`s  `get_global_active_transactions()` and `get_all_active_transactions()` which are only useful for cluster monitoring/debugging purposes. So, it is very unlikely that a user would use those functions. And, even if they do, we can tell them either upgrade or do not use.

(b)  To backport those functions, we may need multiple PRs since as far as I remember the code has changed for those functions between 7.5 to 8.1. (Note that this part exists from 7.5, not 8.0 as this PR)

(c) And, those functions actually do tiny memory allocations while holding the spinlocks. So, the likelyhood of this bug happening seems too low.

PR #2569, opened 2018-12-24
 Fix having clause bug for complex joins

### LINKED ISSUE #2558: Aggregate in HAVING might use the wrong column in joins
```
CREATE TABLE descriptions (key int, description text);
CREATE TABLE amounts (key int, amount int);
SELECT create_distributed_table('descriptions','key');
SELECT create_distributed_table('amounts','key');

SELECT amounts.key, sum(amounts.amount)
FROM descriptions d1 JOIN descriptions d2 USING (key) JOIN amounts USING (key) 
GROUP BY amounts.key HAVING sum(amounts.amount) > 200;
...
WARNING:  function sum(text) does not exist
ERROR:  failed to execute task 16
```

Under the covers it's sending `sum(d1.description)` instead of `sum(amounts.amount)`:
```
COPY (SELECT amounts.key, sum(amounts.amount) AS sum, sum(amounts.amount) AS worker_column_3 FROM ((descriptions_103537 d1 JOIN descriptions_103537 d2 ON ((d1.key OPERATOR(pg_catalog.=) d2.key))) JOIN amounts_103569 amounts ON ((d1.key OPERATOR(pg_catalog.=) amounts.key))) WHERE true GROUP BY amounts.key HAVING (sum(d1.description) OPERATOR(pg_catalog.>) 200)) TO STDOUT
```

mtuncer:
We update column attributes of various clauses for a query
including target columns, select clauses when we introduce
new range table entries in the query.

It seems having clause column attributes were not updated.

This fix resolves the issue

DESCRIPTION: Fixes an issue with having clause when used with complex joins

Fixes #2558

### DIFF
--- a/src/backend/distributed/planner/multi_physical_planner.c
+++ b/src/backend/distributed/planner/multi_physical_planner.c
@@ -688,6 +688,18 @@ BuildJobQuery(MultiNode *multiNode, List *dependedJobList)
 		}
 	}
 
+	/* set correct column attributes for having columns */
+	if (updateColumnAttributes)
+	{
+		columnCell = NULL;
+		columnList = pull_var_clause_default((Node *) havingQual);
+		foreach(columnCell, columnList)
+		{
+			Var *column = (Var *) lfirst(columnCell);
+			UpdateColumnAttributes(column, rangeTableList, dependedJobList);
+		}
+	}
+
 	/*
 	 * Build the From/Where construct. We keep the where-clause list implicitly
 	 * AND'd, since both partition and join pruning depends on the clauses being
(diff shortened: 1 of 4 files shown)


mtuncer:
@metdos sending your way for reviewer assignment

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -688,6 +688,18 @@ BuildJobQuery(MultiNode *multiNode, List *dependedJobList)
...
 
+	/* set correct column attributes for having columns */
+	if (updateColumnAttributes)
+	{
+		columnCell = NULL;
+		columnList = pull_var_clause_default((Node *) havingQual);
+		foreach(columnCell, columnList)
+		{
+			Var *column = (Var *) lfirst(columnCell);
+			UpdateColumnAttributes(column, rangeTableList, dependedJobList);
```
marcocitus: This code looks almost identical to the code above.

Maybe:
```c
if (updateColumnAttributes)
{
    UpdateAllColumnAttributes(selectClauseList, rangeTableList, dependedJobList);
    UpdateAllColumnAttributes(havingQual, rangeTableList, dependedJobList);
}
```

mtuncer: code is also duplicated earlier in the function for target entries

mtuncer: will update all

PR #2571, opened 2019-01-02
Relax reference table checks in subquery union pushdown

### LINKED ISSUE #2518: Subquery pushdown planning with UNION/UNION ALL fails with reference tables
Reported by @saicitus and @marcocitus. I wanted to open an issue to not track of it.

```SQL

set client_min_messages TO DEBUG;

-- unnecessarily switch to recursive planning 
 SELECT * FROM (SELECT x, y FROM test JOIN ref ON (x = a) UNION ALL SELECT x, y FROM test) a;
DEBUG:  generating subplan 1_1 for subquery SELECT test.x, test.y FROM (public.test JOIN public.ref ON ((test.x OPERATOR(pg_catalog.=) ref.a)))
DEBUG:  generating subplan 1_2 for subquery SELECT x, y FROM public.test
DEBUG:  Creating router plan
DEBUG:  generating subplan 1_3 for subquery SELECT intermediate_result.x, intermediate_result.y FROM read_intermediate_result('1_1'::text, 'binary'::citus_copy_format) intermediate_result(x integer, y integer) UNION ALL SELECT intermediate_result.x, intermediate_result.y FROM read_intermediate_result('1_2'::text, 'binary'::citus_copy_format) intermediate_result(x integer, y integer)
DEBUG:  Plan 1 query after replacing subqueries and CTEs: SELECT x, y FROM (SELECT intermediate_result.x, intermediate_result.y FROM read_intermediate_result('1_3'::text, 'binary'::citus_copy_format) intermediate_result(x integer, y integer)) a
DEBUG:  Creating router plan
 x | y 
---+---
(0 rows)
```

There is no reason for this query not to pushdown. We seem to have aggressive checks in `DeferErrorIfUnsupportedUnionQuery()` such that it errors out whenever we see a reference table or recurring tuple. The check seems to be added for the following types of queries, which is still valid, so we shouldn't forget those cases:

```SQL
SELECT * FROM 
(
     SELECT * FROM reference_table 
UNION
    SELECT * FROM distributed_table
);
```

mtuncer:
DESCRIPTION: Relax subquery union pushdown restrictions for reference tables

Fixes #2518

### DIFF
--- a/src/backend/distributed/planner/query_pushdown_planning.c
+++ b/src/backend/distributed/planner/query_pushdown_planning.c
@@ -53,6 +53,11 @@ typedef enum RecurringTuplesType
 	RECURRING_TUPLES_RESULT_FUNCTION
 } RecurringTuplesType;
 
+typedef struct
+{
+	PlannerRestrictionContext *plannerRestrictionContext;
+	RecurringTuplesType recurringTupleType;
+} HasRecurringTuplesContext;
 
 /* Config variable managed via guc.c */
 bool SubqueryPushdown = false; /* is subquery pushdown enabled */
@@ -75,8 +80,15 @@ static bool RelationInfoContainsRecurringTuples(PlannerInfo *plannerInfo,
 												RecurringTuplesType *recurType);
 static bool IsRecurringRTE(RangeTblEntry *rangeTableEntry,
 						   RecurringTuplesType *recurType);
-static bool IsRecurringRangeTable(List *rangeTable, RecurringTuplesType *recurType);
-static bool HasRecurringTuples(Node *node, RecurringTuplesType *recurType);
+static bool IsRecurringRangeTable(List *rangeTable,
+								  HasRecurringTuplesContext *hasRecurringTuplesContext);
+static void DetermineReferenceTableJoinRelation(PlannerInfo *plannerInfo, int
+												sourceRTEIdentity,
+												RelOptInfo *relinfo,
+												bool *hasReferenceTable,
+												bool *hasDistributedTable);
+static bool HasRecurringTuples(Node *node,
+							   HasRecurringTuplesContext *hasRecurringTuplesContext);
 static MultiNode * SubqueryPushdownMultiNodeTree(Query *queryTree);
 static void FlattenJoinVars(List *columnList, Query *queryTree);
 static void UpdateVarMappingsForExtendedOpNode(List *columnList,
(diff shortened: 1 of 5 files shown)


mtuncer:
I am not 100% happy with the fix

onderkalaci:
> Wouldn't it be sufficient to verify that the subquery contains "not just recurring tuples"? Things like ref_table LEFT JOIN dist_table will already be checked elsewhere.

Agreed, as I had already pointed related functions [here](https://github.com/citusdata/citus/issues/2518#issuecomment-449856272) plus [here](https://github.com/citusdata/citus/issues/2518#issuecomment-449857074).

mtuncer:
@onderkalaci I don't see the problem as the same

We were errorring out as soon as we see a reference table inside a subquery within union. We wanted to relax this such that we would allow cases where a reference table is joined by a distributed table, but not in the outer part.

Only simplification I could think of is to remove checks regarding if the reference table is on the outer side. Still we need to check if the other table in the join is a distributed table. Single reference table in the query will still need to be disallowed.

marcocitus:
> We wanted to relax this such that we would allow cases where a reference table is joined by a distributed table, but not in the outer part.

Our old understanding was: Recurring tuples are dangerous. If we saw recurring tuples in a part of a join or union we generated an error.

We came to a new understanding while working on #2481: An RTE/subquery is only "recurring" (dangerous) if it does not contain a distributed table. There are exceptions to this rule, such as `FROM <ref table> WHERE x IN (SELECT * FROM <dist table>)`, but we check all of those exceptions explicitly for each subquery in `DeferErrorIfUnsupportedSubqueryPushdown` and generate an appropriate error.

After `DeferErrorIfUnsupportedUnionQuery` we'll still do `DeferredErrorIfUnsupportedRecurringTuplesJoin` which looks at all joins in all subqueries, so you don't need to check whether joins are valid in `DeferErrorIfUnsupportedUnionQuery`. If they're not, we'll error out anyway.

Per our new understanding, you only need to check whether or not there is a distributed table. We might want something like a `HasOnlyRecurringTuples` instead of `HasRecurringTuples`. If `HasOnlyRecurrsingTuples`, generate a relevant error. Otherwise, let other checks deal like `DeferredErrorIfUnsupportedRecurringTuplesJoin` and `DeferErrorIfCannotPushdownSubquery` with the distributed table in the subquery.

#2481 introduces a `RelationInfoContainsOnlyRecurringTuples` which has the same goal, though I don't think you can use it here since it works on `RelOptInfo`.

mtuncer:
Closing this in favor of #2590

Issue #2519, opened 2018-12-04
Handle function calls in joins


marcocitus:
This would be a nice warm-up task to become familiar with the Citus planner and Postgres data structures.

Currently, the following fails on Citus because it tries to do the function call on the workers:
```
CREATE TABLE test (x int, y int);
SELECT create_distributed_table('test','x');
CREATE SEQUENCE numbers;
SELECT * FROM test JOIN nextval('numbers') n ON (x = n);
...
WARNING:  relation "public.numbers" does not exist
WARNING:  relation "public.numbers" does not exist
ERROR:  failed to execute task 1
```

There is an easy workaround, which is to wrap the function call in a subquery.
```
SELECT * FROM test JOIN (SELECT * FROM nextval('numbers') n)  nv ON (x = n);
```

It would be nice if it just worked.

hanefi:
I started working on this task ( see #2572 ). I realized that this proposed workaround does not always work, and for some cases some more action needs to be taken.


Here are some cases of function calls that are quite hard to tackle along with their complications and my ideas on how to tackle them:
- Function calls that are in the right side of a `LATERAL JOIN`
  - The inputs to the function that reference cloumns of previous tables are stored in `varno` entries. When a subquery is introduced that wraps the function, this input reference now needs to point to a `var` in the parent query and AFAIK there is no easy way to do this.
  - If we use a column of a table expression as input to a function, we modify the query tree to include `LATERAL JOIN`s, therefore some CTE's do not work.
  - Proposed solution:
    - incrementing `varlevelsup` attribute of all `Var` nodes. (for subquery `Var`s referencing outer relations, this shows how many levels up one must travel to find the referenced column)
    - Note: I was not successful in my attempts to do this. Either I am doing something wrong or this does not solve our problem.
- Joining user-defined immutable functions
  - Currently we treat immutable functions as if they are reference tables ( see [here](https://github.com/citusdata/citus/blob/1656b519c49062dee5bb5fc8174963821dda9f17/src/backend/distributed/planner/query_pushdown_planning.c#L998) ). If this function is not defined in workers, we fail to execute queries.
  - Proposed solution alternatives:
    1. Put `OFFSET 0` at the subqueries to force pushdown planning and not treat the function like a reference table.
    2. Change the behaviour of `query_pushdown_planning` and defer an error if the immutable function is a user defined one
- Functions returning a table/a record/setof records
[...]

PR #2572, opened 2019-01-03
Wrap single column/value functions in subqueries

hanefi:
DESCRIPTION: Added support for function calls in joins

If a function call is used in a join with a distributed table, Citus tries to execute that said function in workers. If said function utilizes some resources that are available only in the coordinator node, the query fails.

Currently, the following fails on Citus because it tries to do the function call on the workers:
```
CREATE SEQUENCE numbers;
SELECT * FROM test JOIN nextval('numbers') n ON (x = n);
...
WARNING:  relation "public.numbers" does not exist
WARNING:  relation "public.numbers" does not exist
```

There is an easy workaround, which is to wrap the function call in a subquery.
```
SELECT * FROM test JOIN (SELECT * FROM nextval('numbers') n)  nv ON (x = n);
```

We currently do not support all the possible functions. These functions are NOT supported yet:
- User defined `IMMUTABLE` functions that are absent on workers
  - The class of subqueries `SELECT * FROM immutable_func()` will fail to execute on the workers if used in a join. AFAIK, Citus decides to run these queries in Advising the user to create said immutable function makes more sense to me.
- Functions that return a single record / `SETOF` records / tables.
  - Wrapping these functions inside `(SELECT * FROM ...)` is not enough in this case. Possibly a subset of the columns should be present in the query target list. If we wish to support this class of functions I need to have a deeper understanding on the possible use cases that needs to be covered.

See #2519


hanefi:
We have discussed this PR with @onderkalaci and come up with some action items.

- [ ] Improve proposed solution
  - [ ] Do not wrap a built-in immutable function
  - [ ] Whitelist `read_intermediate_result()` calls and make sure it is never wrapped
- [ ] Cleaner Code
  - [ ] Cleanup repetitive checks (`IsRteRowType()` calls in both `WrapFunctionsInQuery()` and `ShouldWrapFunctionsInQuery()`)
  - [ ] Rename functions ( e.g. WrapRteFunctionIntoRteSubquery )
  - [ ] Remove one pointless for loop (We already asserted that it will have one element)
  - [ ] Do not follow a chain of references in a single statement. e.g.
         `((RangeTblFunction *) linitial(rte->functions))->funcexpr;`
  - [ ] Rename the parameter `rteFunction` in `WrapRteFunctionIntoRteSubquery()` that stores a Subquery once we are finished with it
  - [ ] Remove redundant indentations (e.g. else block in `if(..) { return } else {}`)
  - [ ] Rename `IsRteRowType()`
- [ ] Add some more tests
  - [ ] Multiple functions in one RTE. e.g. 
` SELECT * FROM ROWS FROM (f(1), f(2)) as f1(a, b), users_table WHERE user_id = a;`
  - [ ] `WITH ORDINALITY`
  - [ ] CTEs
    - [ ] Functions in joins inside the CTE
    - [ ] Functions joined with the CTE result columns
  - [ ] `OFFSET 0` clauses to enforce recursive planning
  - [ ] Give aliases to some function columns
- [ ] Some more ideas
  - [ ] Generate random data. Run queries on Postgres and compare results with citus
  - [ ] Move the tests in the schedule (Travis is slow when 5 tests are run in parallel)

marcocitus:
> Wrapping these functions inside (SELECT * FROM ...) is not enough in this case. Possibly a subset of the columns should be present in the query target list. If we wish to support this class of functions I need to have a deeper understanding on the possible use cases that needs to be covered.

Mostly user-defined functions that return the result of a stored procedure.

As a dummy example: `SELECT * FROM test JOIN dblink('dbname=postgres', 'SELECT 1,2') AS res(a int, b int) ON (x = a);`

(requires `CREATE EXTENSION dblink`)

hanefi:
> As a dummy example: `SELECT * FROM test JOIN dblink('dbname=postgres', 'SELECT 1,2') AS res(a int, b int) ON (x = a);`

I see that wrapping this function in a `SELECT *` subquery works. The necessary changes I need to make in the parsed query tree is not so clear to me now. I need to look into this scenario a bit more.

For reference I tried to wrap the function like this:
```sql
SELECT *
FROM test JOIN (
SELECT * FROM dblink('dbname=postgres', 'SELECT 1,2') AS res (a int, b int)
               ) sub
         ON (x = a);
```

hanefi:
Had another chat with Onder and we came up with some more items:

- [x] Renaming WrapFunctionInSubquery -> TransformFunctionInSubquery
- [x] Move all checks in a function (`ShouldDoThisThing()`)
- [x] Put some assertions when needed
- [x] Add some DEBUG level logs when a function is being wrapped

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -1305,6 +1312,140 @@ ContainsReferencesToOuterQueryWalker(Node *node, VarLevelsUpWalkerContext *conte
...
 
+/*
+ * WrapFunctionsInQuery iterates over all the immediate Range Table Entries
+ * of a query and wraps the functions inside (SELECT * FROM fnc() f) subqueries.
+ *
+ * We currently wrap only those functions that return a single value. If a
+ * function returns records or a table we leave it as it is
+ * */
+static void
+WrapFunctionsInQuery(Query *query)
```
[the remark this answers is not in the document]
hanefi: Makes sense. I changed the name of the function

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -1305,6 +1312,217 @@ ContainsReferencesToOuterQueryWalker(Node *node, VarLevelsUpWalkerContext *conte
 }
 
 
+/*
+ * WrapFunctionsInSubqueries iterates over all the immediate Range Table Entries
+ * of a query and wraps the functions inside (SELECT * FROM fnc() f) subqueries.
+ *
+ * We currently wrap only those functions that return a single value. If a
+ * function returns records or a table we leave it as it is
```
marcocitus: comment is stale

marcocitus: I think it might also be nice to explain why we want to wrap things in subqueries.

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -1305,6 +1312,217 @@ ContainsReferencesToOuterQueryWalker(Node *node, VarLevelsUpWalkerContext *conte
...
+ * */
+static void
+WrapFunctionsInSubqueries(Query *query)
+{
+	List *rangeTableList = query->rtable;
+	ListCell *rangeTableCell = NULL;
+
+	/* there needs to be at least two RTEs for a join operation */
+	if (list_length(rangeTableList) < 2)
```
marcocitus: The comment could elaborate a bit about why we only care about joins.

src/backend/distributed/planner/recursive_planning.c
hanefi: I have rewritten my stale comments, and added explanations as to why we need to wrap function calls in subqueries

src/backend/distributed/planner/recursive_planning.c
hanefi: I added some explanations to why we only wrap functions that are joined. I did not go into so much detail, but I think this is enough

PR #2576, opened 2019-01-04
Add make target to run regression tests in isolation with vagrant

thanodnl:
DESCRIPTION: Allow vagrant tests to run on a small number of regression tests

This PR adds a new target to the Makefile for our regression tests; `check-base-vg`.

It runs the same tests as it would with `check-base` while enabling vagrant in the same way as `check-multi-vg` would. This can be used when debugging a failure observed with vagrant in a specific test.


metdos:
@thanodnl it seems @velioglu approved this PR. Any reason to not merge this?

thanodnl:
nothing other then that I missed it :), merging now

PR #2581, opened 2019-01-09
Add changelog entry for 8.0.3

hanefi:
Add changelog entry for 8.0.3

The items listed are the same as 8.1.1 since the patch is the same

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+### citus v8.0.3 (January 9, 2019) ###
+
+* Fixes maintenance daemon panic due to unreleased spinlock
+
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,12 @@
 
 * Fixes an issue with having clause when used with complex joins
 
+### citus v8.0.3 (January 9, 2019) ###
```
velioglu: Can we keep the versions in time-sorted? I mean putting that version at the top.

hanefi: Makes sense. Ordering by the version numbers looked like the better option to me. I guess this is not my decision to make though.

Issue #2556, opened 2018-12-19
Running valgrind tests with SSL enabled leads to regression test failures


onderkalaci:
Version: Citus 8.1, PostgreSQL REL_11_STABLE (I guess 11.1). Using our tools repo to run the valgrind tests.

The following tests fails consistently when SSL is enabled (not when the SSL is disabled).

The test that fails [here](https://github.com/citusdata/citus/blob/master/src/test/regress/expected/multi_alter_table_add_constraints.out#L21-L28). There seems to be some other unrelated failures as well. Also note that we got #2552 while running these tests, but that's not consistently happening.

The full regressions.diff we get is below. Note that the other 
```
--- /home/ubuntu/citus/src/test/regress/expected/multi_alter_table_add_constraints.out  2018-12-14 08:52:13.728245525 +0000
+++ /home/ubuntu/citus/src/test/regress/results/multi_alter_table_add_constraints.out   2018-12-17 13:11:25.167047965 +0000
@@ -18,34 +18,35 @@
  
 (1 row)
 
 -- Can only add primary key constraint on distribution column (or group of columns
 -- including distribution column)
 -- Command below should error out since 'name' is not a distribution column
 ALTER TABLE products ADD CONSTRAINT p_key PRIMARY KEY(name);
 ERROR:  cannot create constraint on "products"
 DETAIL:  Distributed relations cannot have UNIQUE, EXCLUDE, or PRIMARY KEY constraints that do not include the partition column (with an equality operator if EXCLUDE).
 ALTER TABLE products ADD CONSTRAINT p_key PRIMARY KEY(product_no);
+ERROR:  connection error: localhost:57638
+DETAIL:  server closed the connection unexpectedly
+       This probably means the server terminated abnormally
+       before or while processing the request.
 INSERT INTO products VALUES(1, 'product_1', 1);
 -- Should error out, since we are trying to add a new row having a value on p_key column
 -- conflicting with the existing row.
 INSERT INTO products VALUES(1, 'product_1', 1);
-ERROR:  duplicate key value violates unique constraint "p_key_1450001"
-DETAIL:  Key (product_no)=(1) already exists.
-CONTEXT:  while executing command on localhost:57638
 ALTER TABLE products DROP CONSTRAINT p_key;
+ERROR:  constraint "p_key" of relation "products" does not exist
 INSERT INTO products VALUES(1, 'product_1', 1);
 -- Can not create constraint since it conflicts with the existing data
[...]

PR #2584, opened 2019-01-10
Finish connection establishment in parallel for multiple connections

thanodnl:
DESCRIPTION: Finish connection establishment in parallel for multiple connections

This change came out of #2556.

When running the test mentioned in the ticket in valgrind it tries to establish 64 connections over SSL. Due to old implementation of `FinishConnectionListEstablishment` the connections are effectively established in parallel. As SSL has more roundtrips and more compute per roundtrip compared to non-ssl it takes longer then ~60 seconds in valgrind which seem to be a magic time that sockets stay in `ESTAB` state on linux.

By processing the roundtrips in parallel connnection establishment should take less time, both in valgrind and in actual workloads.

Current state:
 - [x] regression tests pass
 - [x] failure tests pass
 - [x] running tests in valgrind pass*

Suspicion of short commings:
 - [x] timeouts on connection establishment seem to be handled in the single connection function and not here

* there is one (1) valgrind report on the use of uninitialized variable use in postgres, this has been observed before, but internal discussions died.

### DIFF
--- a/src/backend/distributed/connection/connection_management.c
+++ b/src/backend/distributed/connection/connection_management.c
@@ -9,11 +9,7 @@
  */
 
 #include "postgres.h"
-
-#ifdef HAVE_POLL_H
-#include <poll.h>
-#endif
-
+#include "pgstat.h"
 
 #include "libpq-fe.h"
 
@@ -23,6 +19,7 @@
 #include "commands/dbcommands.h"
 #include "distributed/connection_management.h"
 #include "distributed/errormessage.h"
+#include "distributed/memutils.h"
 #include "distributed/metadata_cache.h"
 #include "distributed/hash_helpers.h"
 #include "distributed/placement_connection.h"
@@ -47,6 +44,30 @@ static void AfterXactHostConnectionHandling(ConnectionHashEntry *entry, bool isC
 static void DefaultCitusNoticeProcessor(void *arg, const char *message);
 static MultiConnection * FindAvailableConnection(dlist_head *connections, uint32 flags);
 static bool RemoteTransactionIdle(MultiConnection *connection);
+static int EventSetSizeForConnectionList(List *connections);
+
+/* types for async connection management */
+enum MultiConnectionPhase
+{
+	MULTI_CONNECTION_PHASE_CONNECTING,
+	MULTI_CONNECTION_PHASE_CONNECTED,
+	MULTI_CONNECTION_PHASE_ERROR,
+};
+typedef struct MultiConnectionState
+{
+	enum MultiConnectionPhase phase;
+	MultiConnection *connection;
+	PostgresPollingStatusType pollmode;
+} MultiConnectionState;
+
+
+/* helper functions for async connection management */
+static bool MultiConnectionStatePoll(MultiConnectionState *connectionState);
+static WaitEventSet * WaitEventSetFromMultiConnectionStates(List *connections,
+															int *waitCount);
+static long DeadlineTimestampTzToTimeout(TimestampTz deadline);
+static void CloseNotReadyMultiConnectionStates(List *connectionStates);
+static uint32 MultiConnectionStateEventMask(MultiConnectionState *connectionState);
 
 
 static int CitusNoticeLogLevel = DEFAULT_CITUS_NOTICE_LEVEL;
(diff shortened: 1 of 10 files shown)


src/backend/distributed/connection/connection_management.c
```diff
@@ -437,6 +453,148 @@ ShutdownConnection(MultiConnection *connection)
...
+	}
+
+	if (waitCount)
+	{
+		*waitCount = 0;
+	}
+
+	/* allocate pending connections + 2 for the signal latch and postmaster death */
+	/* (CreateWaitEventSet makes room for pgwin32_signal_event automatically) */
+	waitEventSet = CreateWaitEventSet(CurrentMemoryContext, maxEvents + 2);
```
marcocitus: should this be numConnections + 2?

src/backend/distributed/connection/connection_management.c
```diff
@@ -445,15 +603,102 @@ ShutdownConnection(MultiConnection *connection)
...
+		if (!connections[connectionIndex].ready)
+		{
+			waitCount++;
+		}
+
+		connectionIndex++;
+	}
 
-		/* TODO: consider making connection establishment fully in parallel */
-		FinishConnectionEstablishment(multiConnection);
```
marcocitus: Can we replace `FinishConnectionEstablishment` with the logic in this function?

marcocitus: e.g. do `FinishConnectionListEstablishment(list_make1(connection));`

src/backend/distributed/connection/connection_management.c
```diff
@@ -445,15 +603,102 @@ ShutdownConnection(MultiConnection *connection)
...
+
+			if (waitCount <= 0)
+			{
+				break;
+			}
+		}
+
+		eventCount = WaitEventSetWait(waitEventSet, timeout, events, waitCount,
+									  WAIT_EVENT_CLIENT_READ);
+
```
marcocitus: should probably check for a 0 return value and then check the timeout

src/backend/distributed/connection/connection_management.c
```diff
@@ -445,15 +603,102 @@ ShutdownConnection(MultiConnection *connection)
...
+				if (InterruptHoldoffCount > 0 && (QueryCancelPending || ProcDiePending))
+				{
+					return;
+				}
+
+				continue;
+			}
+
+			MultiConnectionReadyExecutePoll(connection);
+
```
marcocitus: It would be nice to do a `ModifyWaitEvent` when the list of ready connections has not changed, but the pollmode has.

src/backend/distributed/connection/connection_management.c
thanodnl: There is a limit of events that can be added to the `WaitEventSet` as described in `FD_SETSIZE`.

Most of the time this number will be numConnections as `maxEvents` gets initialized to `numConnections`. However, if the number of connections becomes to big (~1024 on osx) you can't add them all to the `WaitEventSet`.

Instead we keep track of how many connections we can add via `maxEvents`. When this reaches 0 we stop adding them.

src/backend/distributed/connection/connection_management.c
thanodnl: I was thinking how to do this. Your suggestion makes a lot of sense, running tests with this now locally. Will push after they pass.

src/backend/distributed/connection/connection_management.c
thanodnl: Makes sense, I think we should also set the `timeout` in that case. Currently its -1, but I think we could calculate a deadline based on the citus connection timeout setting and compute the timeout based on the deadline.

src/backend/distributed/connection/connection_management.c
thanodnl: Yeah, there are roughly 3 outcomes after this function.

1. connection is not ready and pollmode stays the same, eg. nothing to do
2. poll mode changed, we can `ModifyWaitEvent` with the new pollmode
3. connection is ready, rebuild the waitset

This would prevent a lot of rebuilds in the case of ssl (it changes state ~7 times before the connection is ready).

src/backend/distributed/connection/connection_management.c
marcocitus: the naming here is confusing, maxEvents appears to not be the maximum number of events?

thanodnl: How do you mean? It is the maximum number of `WaitEvent`s that can be added to the set, hence `maxEvents`. We can make it more verbose by `maxWaitEvents` or `maxWaitEventsToAdd` as it eventually guards the number of times `AddWaitEventToSet` can be called.

marcocitus: It might help to introduce some additional variables to make the code more self-documenting, e.g.
- `maxEvents` is the maximum number of events we can add
- `numEventsAdded` is the number of events we added
- `waitEventSetSize = Min(maxEvents, numConnections) + 2` is the size of the wait event set

src/backend/distributed/connection/connection_management.c
```diff
@@ -437,187 +458,347 @@ ShutdownConnection(MultiConnection *connection)
...
-			int pollResult = 0;
+			waitEventSet = WaitEventSetFromMultiConnectionReadyArray(connections,
+																	 connectionCount,
+																	 &waitCount);
 
-			/* we use poll(2) if available, otherwise select(2) */
-#ifdef HAVE_POLL
-			struct pollfd pollFileDescriptor;
+			/* TODO use resource that only holds the last pointer and frees it on swap */
+			EnsureReleaseResource((MemoryContextCallbackFunction) (&FreeWaitEventSet),
```
marcocitus: This call might come too late if there is an exception while constructing the wait event set. We should probably do this at the beginning (and resolve the TODO).

src/backend/distributed/connection/connection_management.c
```diff
@@ -437,184 +462,383 @@ ShutdownConnection(MultiConnection *connection)
...
+	{
+		MultiConnectionState *connectionState = lfirst(connectionCell);
+		MultiConnection *connection = connectionState->connection;
+		if (connectionState->ready)
+		{
+			continue;
+		}
+
+		/* close connection, otherwise we take up resource on the other side */
+		PQfinish(connection->pgConn);
```
marcocitus: I think we should be calling `CloseConnection` here.

src/test/regress/Makefile
```diff
@@ -50,6 +50,11 @@ check-base: all
 	$(pg_regress_multi_check) --load-extension=citus \
 	-- $(MULTI_REGRESS_OPTS) --schedule=$(citus_abs_srcdir)/base_schedule $(EXTRA_TESTS)
 
+check-base-vg: all
+	$(pg_regress_multi_check) --load-extension=citus \
+	--valgrind --pg_ctl-timeout=360 --connection-timeout=500000 --valgrind-path=valgrind --valgrind-log-file=$(VALGRIND_LOG_FILE) \
+	-- $(MULTI_REGRESS_OPTS) --schedule=$(citus_abs_srcdir)/base_schedule $(EXTRA_TESTS)
```
marcocitus: probably should be in a different PR

src/backend/distributed/connection/connection_management.c
```diff
@@ -437,184 +462,383 @@ ShutdownConnection(MultiConnection *connection)
...
+			waitCount++;
 		}
+	}
 
-		/* Loop, to handle poll() being interrupted by signals (EINTR) */
-		while (true)
-		{
-			int pollResult = 0;
+	/* prepapre space for socket events */
+	#define MIN(a, b) ((a) > (b) ? (b) : (a))
```
marcocitus: there's a `Min` in `c.h`, which gets included via `postgres.h`.

src/backend/distributed/connection/connection_management.c
marcocitus: ping

src/backend/distributed/connection/connection_management.c
```diff
@@ -437,184 +457,378 @@ ShutdownConnection(MultiConnection *connection)
...
+
+		/* close connection, otherwise we take up resource on the other side */
+		PQfinish(connection->pgConn);
+		connection->pgConn = NULL;
+	}
+}
+
+
+static bool
+CheckMultiConnectionStateTimeouts(List *connections)
```
marcocitus: Needs function comments

src/backend/distributed/connection/connection_management.c
```diff
@@ -437,184 +457,378 @@ ShutdownConnection(MultiConnection *connection)
...
+
+		if (eventCount == 0)
+		{
+			/*
+			 * timeout has occured, lets check there is an actual timeout, report if that
+			 * is the case and close all non-connected sockets
+			 */
+
+			bool timeoutOccured = CheckMultiConnectionStateTimeouts(connectionStates);
+			if (timeoutOccured)
```
marcocitus: Checking all the different timers seems a bit weird. What if we just have one timer for this function?

Theoretically the connections could start at different times, but practically...

marcocitus:
Ping.

src/backend/distributed/connection/connection_management.c
```diff
@@ -437,184 +457,378 @@ ShutdownConnection(MultiConnection *connection)
...
 	{
-		MultiConnection *multiConnection = (MultiConnection *) lfirst(
-			multiConnectionCell);
+		connectionState->ready = true;
+		return Rebuild;
+	}
+	else if (status == CONNECTION_BAD)
+	{
+		/* FIXME: retries? */
+		connectionState->ready = true;
```
pykello: Was this meant to be set to `false`? If not, I'm a bit confused why we're setting this to `true`.

src/backend/distributed/connection/connection_management.c
```diff
@@ -437,184 +457,378 @@ ShutdownConnection(MultiConnection *connection)
...
+	connectionState->pollmode = PQconnectPoll(connection->pgConn);
 
-		/* TODO: consider making connection establishment fully in parallel */
-		FinishConnectionEstablishment(multiConnection);
+	/*
+	 * FIXME: Do we want to add transparent retry support here?
+	 */
+	if (connectionState->pollmode == PGRES_POLLING_FAILED)
+	{
+		connectionState->ready = true;
```
pykello: Why is this being set to `true`?

src/backend/distributed/connection/connection_management.c
marcocitus: I think because we're no longer waiting for the connection. Perhaps this should be an enum instead of a bool to avoid confusion.

src/backend/distributed/connection/connection_management.c
```diff
@@ -437,184 +457,378 @@ ShutdownConnection(MultiConnection *connection)
...
+				/* we are done waiting for the sockets */
+				break;
+			}
+		}
+	}
+	MemoryContextDelete(MemoryContextSwitchTo(oldContext));
+}
+
+
+static int
```
pykello: This should probably return `long`.

metdos:
Hi  @thanodnl, what is the status of this PR?

thanodnl:
It fell off my radar before the ping of marco last week, however I was busy fighting with the packaging for citusha.

Now that those are out I will have a look at this today. Last I remembered of it was that the scope became quite broad for the change as it is not only making it async, but basically changing how the whole connections are connected. I think this needs more testing to make sure it doesn't break anything.

With the comments of Hadi I think I have enough to work on, but I might need some help 'breaking' the code. As the code will be impacting every and all queries in its core (connecting) this can't really be shipped without confidence.

src/backend/distributed/connection/connection_management.c
thanodnl: It looked like a thing making sense, but the testsuite fails when using `CloseConnection`, so I guess the connection is not yet put in the hashmap during the connection phase.

metdos:
@marcocitus do we want these changes in 8.2?

src/backend/distributed/connection/connection_management.c
marcocitus: That seems like a separate bug. `StartNodeUserDatabaseConnection` puts the connection into the hash.

marcocitus:
> @marcocitus do we want these changes in 8.2?

Yes, as in "should have", not "must have". We've gotten some complaints about the latency in small multi-shard update/deletes, plus the original motivation for this PR was that we got timeouts in valgrind.

thanodnl:
I should have addressed the feedback that was left in earlier rounds, @pykello do you have time for a review?

src/backend/distributed/connection/connection_management.c
```diff
@@ -438,183 +465,373 @@ ShutdownConnection(MultiConnection *connection)
...
+	ListCell *connectionCell = NULL;
+
+	/*
+	 * maxEvents is the number of events we can add to the WaitEventSet.
+	 * -3 is used to have room for the following events outside of our connection events
+	 *  - WL_POSTMASTER_DEATH
+	 *  - WL_LATCH_SET
+	 *  - pgwin32_signal_event (automatically added by CreateWaitEventSet on WIN32)
+	 */
+	const int maxEvents = FD_SETSIZE - 3;
```
pykello: Why are we using `FD_SETSIZE` here? It is a `select()` related constant. Wait event sets doesn't seem to use selects at all.  In Windows it uses `WaitForMultipleObjects()`, which can wait for `MAXIMUM_WAIT_OBJECTS` unless the suggestions at https://docs.microsoft.com/en-us/windows/desktop/api/synchapi/nf-synchapi-waitformultipleobjects are followed. `latch.c` doesn't seem to follow those suggestions, so I'd suggest adding something like:

```c
/*
 * WaitEventSets use WaitForMultipleObjects() in windows, which can wait for up to
 * MAXMIUM_WAIT_OBJECTS events. CreateWaitEventSet() internally uses one
 * of these slots for pgwin32_signal_event, so we have (MAXMIUM_WAIT_OBJECTS - 1)
 * slots to use.
 */
#ifdef WIN32
#define MAX_EVENTSET_SIZE (MAXMIUM_WAIT_OBJECTS - 1)
#else
#define MAX_EVENTSET_SIZE 1024
#endif
```

which abstracts away the `WaitEventSet` implementation related parts, and using `MAX_EVENTSET_SIZE - 2` here.

 What do you think?

src/backend/distributed/connection/connection_management.c
```diff
@@ -438,183 +465,373 @@ ShutdownConnection(MultiConnection *connection)
...
-		{
-			return;
-		}
-		else if (pollmode == PGRES_POLLING_OK)
+		eventMask = MultiConnectionStateEventMask(connectionState);
+
+		AddWaitEventToSet(waitEventSet, eventMask, socket, NULL, connectionState);
+		numEventsAdded++;
+
+		/* increment the waitCount if non NULL */
```
pykello: the comment seem to be exactly what the code is, so we can remove the code.

pykello: I mean the comment :D

src/backend/distributed/connection/connection_management.c
thanodnl: I tried finding where I got this from, must have been an example somewhere, but couldn't find it. I went for your suggestion here.

Do you think it is safe to use the 1k hardcoded default? Or is there a header/define in postgres that would make more sense? Looking for one didn't yield much results for me.

src/backend/distributed/connection/connection_management.c
```diff
@@ -438,183 +479,377 @@ ShutdownConnection(MultiConnection *connection)
...
+			/*
+			 * timeout has occured on waitset, double check the timeout since
+			 * connectionStart and if passed close all non-finished connections
+			 */
+
+			TimestampTz now = GetCurrentTimestamp();
+			if (TimestampDifferenceExceeds(connectionStart, now, NodeConnectionTimeout))
+			{
+				CloseNotReadyMultiConnectionStates(connectionStates);
+				ereport(ERROR, (errmsg("could not establish connection after %u ms",
```
pykello: 1. Previously we were emitting this message as a `WANRING`, why did we change it to an `ERROR`?
2. If this is intended to be an `ERROR`, do we have need to reset the temporary memory context before emitting the error?

src/backend/distributed/connection/connection_management.c
```diff
@@ -46,6 +56,37 @@ static void AfterXactHostConnectionHandling(ConnectionHashEntry *entry, bool isC
...
+	enum MultiConnectionPhase phase;
+	MultiConnection *connection;
+	PostgresPollingStatusType pollmode;
+} MultiConnectionState;
+
+
+/* helper functions for async connection management */
+enum EventSetResult
+{
+	REBUILD,
```
pykello: Maybe `REBUILD_WAITEVENTSET`?

src/backend/distributed/connection/connection_management.c
```diff
@@ -46,6 +56,37 @@ static void AfterXactHostConnectionHandling(ConnectionHashEntry *entry, bool isC
...
+	UPDATE,
```
pykello: Maybe `UPDATE_EVENT_MASK`?

src/backend/distributed/connection/connection_management.c
pykello: WaitEventSets use either `poll()` or `epoll()` in non-Windows, and these functions don't seem to have any limits unlike `select()` or `WaitForMultipleObjects()`. `1024` is rather arbitrary, we can make it a much larger number. We didn't have to do this if the implementation in `latch.c` could work around the limits of `WaitForMultipleObjects()`.

pykello:
Much of the complexity in this change is to work around the shortcomings of WaitEventSets in Windows. I think if we could refactor those workarounds into a separate file, we could replace `poll()`/`select()` in rest of Citus code with WaitEventSets. But since that'd require a good amount of design and thinking and testing, I'm ok with the current PR and leaving the refactor to another PR.

pykello:
@thanodnl based on the outcome of an internal chat, it is ok to simplify the code and assume WaitEventSets don't have event count limits and remove the rebuild and `MAX_EVENTSET_SIZE` related parts, and a put comment that we need to fix this for `> MAXIMUM_WAIT_OBJECTS` events in Windows. I think this should make the code much simpler.

marcocitus:
Note that the rebuild is necessary because there is no way to stop waiting for a connection once it's done/failed. ModifyWaitEvent has an assert that disallows 0 as an event mask, presumably because the underlying implementation may not support disabling a particular socket. Would be nice if the WaitEventSet code itself did the rebuild though.

pykello:
> Note that the rebuild is necessary because there is no way to stop waiting for a connection once it's done/failed. ModifyWaitEvent has an assert that disallows 0 as an event mask, presumably because the underlying implementation may not support disabling a particular socket. Would be nice if the WaitEventSet code itself did the rebuild though.

OK. Makes sense.

src/backend/distributed/connection/connection_management.c
thanodnl: An explanation is on the previous discussion: https://github.com/citusdata/citus/pull/2584#discussion_r266419735
TLDR: not reporting it as an error causes a segmentation fault.

Connections are already added to the cache so we nee to call `CloseConnection` on all connections to remove them from the cache. `CloseConnection` free's the data of the `MultiConnection`. Functions calling `FinishConnection(List)Establishment` assume the connections are established.

By changing it to an error we break out from that codepath which makes it safe to call `CloseConnection`, not reporting it as an error causes a segmentation fault.

Besides removing the connection from the cache, the error message becomes more clear as the timeout is not reported as an error opposed to a generic `connection error` from the coordinator with a warning above hinting that a timeout occurred.

thanodnl:
I followed the suggestion to remove the upper bound on the events we add to the WaitEventSet. As per comment of Marco, we still need to have the rebuild logic as we cannot disable wait events once they are in the set, hence the code didn't become less complicated :(. Theoretically we could swap established connections to the end of the list and reduce the number of events to wait from in the list, but that feels more complicated for a user to understand what is going on then rebuilding the set.

I did do some quick tests to see if it is safe to add an 'unbounded' list to the `WaitEventSet`, it is surprisingly complicated to test due to the process limit of both postgres and my OS, most often I do get fork issues due to running out of PID's when I query 1200 shards at the same time. But it doesn't seem to cause problems with the number of connections when it doesn't error on forks.

src/backend/distributed/connection/connection_management.c
pykello: Ok, makes sense. Thanks for the explanation. What's your take on my 2nd comment? i.e. calling before reporting errors `MemoryContextDelete(MemoryContextSwitchTo(oldContext));`? Will throwing an error automatically take care of that?

src/backend/distributed/connection/connection_management.c
```diff
@@ -438,183 +466,372 @@ ShutdownConnection(MultiConnection *connection)
...
 }
 
 
 /*
- * Synchronously finish connection establishment of an individual connection.
+ * eventSetSizeForConnectionList calculates the space needed for a WaitEventSet based on a
+ * list of connections.
+ */
+inline static int
+eventSetSizeForConnectionList(List *connections)
```
marcocitus: I'm not sure this still adds much, it might be helpful to actually inline because the number of extra events depends on what the caller does.

src/backend/distributed/connection/connection_management.c
```diff
@@ -438,183 +466,372 @@ ShutdownConnection(MultiConnection *connection)
...
-				/*
-				 * IO possible, continue connection establishment. We could
-				 * check for timeouts here as well, but if there's progress
-				 * there seems little point.
-				 */
-				break;
-			}
-			else
+
+			switch (MultiConnectionStatePoll(connectionState))
```
marcocitus: The naming of `MultiConnectionStatePoll`  does not suggest it returns a value that is relevant to a wait event set. I'm also wondering whether this code could be simplified, e.g. 

```
stateChanged = MultiConnectionStatePoll(connectionState)
if (stateChanged)
{
    if (connectionState->phase == MULTI_CONNECTION_PHASE_CONNECTED ||
        connectionState->phase == MULTI_CONNECTION_PHASE_ERROR)
    {
        /* we cannot stop waiting for a connection, so rebuild the event set */
        waitEventSetRebuild = true;
    }
    else
    {
        /* connection state changed, reset the event mask */
        int eventMask = MultiConnectionStateEventMask(connectionState);
        ModifyWaitEvent(waitEventSet, event->pos, eventMask, NULL);
    }
}
```

src/backend/distributed/connection/connection_management.c
marcocitus: Yes, assuming this function is called within `TopTransactionContext` the memory context will be deleted on abort.

src/backend/distributed/connection/connection_management.c
thanodnl: As discussed, we will keep it as a function (altough with a capitalised name) since it is used on 2 occasions.

 - `WaitEventSetFromMultiConnectionStates`
 - `FinishConnectionListEstablishment`

src/backend/distributed/connection/connection_management.c
thanodnl: I like this approach a lot, implemented it with slight changes.

src/backend/distributed/connection/connection_management.c
thanodnl: I was trying to look for documentation around this, but was hard to find. Thanks Marco for the explanation.

I did have a look at the implementation of `ereport(ERROR,..)` and found that one of the first things they do is switching memory contexts to the ErrorContext, so I guess we don't even need to switch the context back to the `oldContext` before calling.

src/backend/distributed/connection/connection_management.c
```diff
@@ -438,186 +459,370 @@ ShutdownConnection(MultiConnection *connection)
...
+		{
+			/*
+			 * timeout has occured on waitset, double check the timeout since
+			 * connectionStart and if passed close all non-finished connections
+			 */
+
+			TimestampTz now = GetCurrentTimestamp();
+			if (TimestampDifferenceExceeds(connectionStart, now, NodeConnectionTimeout))
+			{
+				CloseNotReadyMultiConnectionStates(connectionStates);
```
marcocitus: I kind of looked over this before. Why do we need it? We close all connections at transaction end in `AfterXactConnectionHandling`.

thanodnl: This is ported from the original code around this comment https://github.com/citusdata/citus/pull/2584/files#diff-0b04fef38ab380718daadfcf21d40eb0L579
```
/* close connection, otherwise we take up resource on the other side */
```

I did not look deeply into the reasoning, what you are suggesting is that the comment is not correct? Given the claim that it would be closed in `AfterXactConnectionHandling`.

marcocitus: Hmz, there are places where we try to be tolerant to connection failures, for example when querying a reference table in the router executor. If we time out on the first replica we will then fail over to the second replica. There are also particular functions that open many connections, but can tolerate some of them failing (e.g. `get_global_active_transactions`).

If we continue with the transaction after a timeout it makes sense to close the connection.

We have various functions that allow caller to specify whether it can tolerate ERRORs by passing a boolean flag (`raiseErrors`). What if we do that here as well? When in doubt, we can default to false since that's the existing behaviour.

thanodnl: I am not fond of adding the `raiseErrors`, would suggest to use `PG_TRY` and `PG_CATCH` if you can tolerate ERRORs.

On the other hand, the reason I changed the code to raise an error is to be able to call CloseConnection instead of PQfinish on the connection. From your previous comment and testing of a `DDL` and `Router query` I was able to confirm that the connection got closed via `AfterXactConnectionHandling`. That would suggest we can change the message back to a warning and have the cases described above still behave the same.

I will try and add tests for the timeout on reference table query and the `get_global_active_transactions` given tests for those are most likely missing in the failure tests. Given your description it sounds like essential tests to have to make this change confidently.

marcocitus: In general, PG_TRY and PG_CATCH should be used sparingly because weird things happen to your variables when you jump through C code https://linux.die.net/man/3/longjmp . 

Changing back to a warning seems fine.

src/backend/distributed/connection/connection_management.c
```diff
@@ -438,183 +458,343 @@ ShutdownConnection(MultiConnection *connection)
...
+				 * Queries that can't will error then and there
+				 */
+				ereport(WARNING, (errmsg("could not establish connection after %u ms",
+										 NodeConnectionTimeout)));
+
+				/*
+				 * we are done waiting for the sockets, any sockets that are in limbo
+				 * will be closed by AfterXactConnectionHandling at the end of the
+				 * transaction
+				 */
```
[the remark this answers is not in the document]
marcocitus: We brought back the function to remain consistent with previous behaviour. This also resolved some strange output in case of failure.

src/backend/distributed/connection/connection_management.c
thanodnl: For future readers; we went full circle on this one.

PQfinish is used here since we want to free resources in case this function is used in a long running transaction. PQfinish will release the resources before reaching the end of the transaction.

Even though it is suboptimal to keep the MultiConnection in the connection cache it is the only possibility.

 - We can't raise an error since there are at least 2 use-cases that can happily continue with non-established connections (`get_global_active_transactions` and queries on reference tablels).
 - We can't call `CloseConnection` as this would cause a use after free on most of the callers of `FinishConnectionListEstablishment`

By calling PQfinish here and setting the `pgConn` to `NULL` we release resources early. The connection will be removed from the cache at the end of the transaction via the transaction callback.

marcocitus:
A few notes from offline discussion:

- Calling `PQfinish` early in case of failure is not necessary, but reasonable because it avoids keeping a connection open that timed out, but did not actually fail. 
- We cannot really remove the connection from the hash during the transaction, because it is allocated in a session-lifetime memory context and we need to make sure it gets free-d. We therefore `PQfinish` and set `pgConn` to `NULL`.
- We should start considering dead connections in `FindAvailableConnection`. Currently, it's possible that a connection fails, but is later returned from the cache when there are additional statements in the same transaction.

PR #2585, opened 2019-01-10
Reset planner context instead of popping with recursive planning

velioglu:
DESCRIPTION: Fixes a problem that could cause segmentation fault with recursive planning

Make recursive planning logic exception safe to not to have segmentation fault in case of insufficient memory by resetting the current planner context instead of pop&push.

### DIFF
--- a/src/backend/distributed/planner/distributed_planner.c
+++ b/src/backend/distributed/planner/distributed_planner.c
@@ -84,6 +84,8 @@ static List * CopyPlanParamList(List *originalPlanParamList);
 static PlannerRestrictionContext * CreateAndPushPlannerRestrictionContext(void);
 static PlannerRestrictionContext * CurrentPlannerRestrictionContext(void);
 static void PopPlannerRestrictionContext(void);
+static void ResetPlannerRestrictionContext(
+	PlannerRestrictionContext *plannerRestrictionContext);
 static bool HasUnresolvedExternParamsWalker(Node *expression, ParamListInfo boundParams);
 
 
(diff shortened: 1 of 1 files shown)


src/backend/distributed/planner/distributed_planner.c
```diff
@@ -703,11 +703,20 @@ CreateDistributedPlan(uint64 planId, Query *originalQuery, Query *query, ParamLi
...
 
-		/* remove the pre-transformation planner restrictions context */
-		PopPlannerRestrictionContext();
+		PG_TRY();
+		{
+			/* remove the pre-transformation planner restrictions context */
+			PopPlannerRestrictionContext();
+
+			/* create a fresh new planner context */
+			plannerRestrictionContext = CreateAndPushPlannerRestrictionContext();
```
marcocitus: In principle this works, but I think it would be nicer to define a function to reset the current planner restriction context, e.g.:
```c
static void
ResetPlannerRestrictionContext(PlannerRestrictionContext *plannerRestrictionContext)
{
    plannerRestrictionContext->relationRestrictionContext =
        palloc0(sizeof(RelationRestrictionContext));

    plannerRestrictionContext->joinRestrictionContext =
        palloc0(sizeof(JoinRestrictionContext));

    plannerRestrictionContext->relationRestrictionContext->allReferenceTables = true;
}
```

velioglu:
@marcocitus I have added the reset function.

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -1536,6 +1537,24 @@ PopPlannerRestrictionContext(void)
 }
 
 
+/*
+ * ResetPlannerRestrictionContext resets the element of the given planner
+ * restriction context.
+ */
+static void
+ResetPlannerRestrictionContext(PlannerRestrictionContext *plannerRestrictionContext)
```
marcocitus: It might make sense to also call this in `CreateAndPushPlannerRestrictionContext`

PR #2587, opened 2019-01-11
Treat functions as transaction blocks

### LINKED ISSUE #659: Support function-delimited transaction blocks
As part of #627, @anarazel and I discussed methods to detect transactions initiated as part of a function. The core of the problem is that `IsTransactionBlock` (which is how that PR detects transactions) will only return `true` if the session is literally within a lexical transaction block (i.e. between a `BEGIN` and a `COMMIT`). Since functions automatically take part in the snapshot of the command that runs them, a simple one-line call to a function _should_ take part in a transaction even if no `BEGIN`/`END` wraps it.

@anarazel found a way to detect this: simply look at the `queryDesc->dest` `CommandType`. If it is _not_ anything remote, it means we're sending results back to some sort of language handler locally rather than to an end user. By looking for this (in addition to `IsTransactionBlock`) we could cover both the `BEGIN`\- and function-triggered cases. However…

It turns out some customers _already_ use modification commands within a function. Previously (i.e. `pg_shard`) we detected and rejected modification commands within a transaction block (as we could not apply them atomically), but did _not_ make use of the above logic to detect functions. So customers used this as a workaround to have "multi-statement" calls (even if not atomic). Adding the correct transactional semantics to function calls would break all their code.

That said, we might still want to add transaction handling to function calls. By providing a GUC, we could permit existing customers to opt-out of this handling, so they could continue to have functions which modify many shards in one call (perhaps even on a per-function level).

We didn't want this to block #627, but are noting it here for future prioritization.

marcocitus:
DESCRIPTION: Treat functions as transaction blocks

Had this sitting around on my laptop.

Currently we don't always open a transaction block on workers for commands that are executed from PL/pgSQL, which means single-shard writes in a function may be prematurely committed, unless the function itself is called from within a transaction block or one of the commands requires 2PC. 

This PR changes the default behaviour to treat PL/pgSQL functions (or more generally, SPI) as transactions and introduces a GUC to revert to the old behaviour.

To determine whether we are in a function call we look at the tuple destination in the `ExecutorRun` hook. A command that is being run from a function will have `DestSPI` as its destination type. This also covers other languages that run queries via SPI.

The new behaviour breaks `create_insert_proxy_for_table`, which has been deprecated for years and is therefore removed in this PR. 

The change also has some side-effects in DDL commands that affect foreign keys, since Postgres checks whether DDL command will violate a foreign key by running a SELECT query SPI, which ends up going through distributed planning. This change mostly improves correctness by making sure these SELECT queries run in the same transaction as the DDL, though at some point we should figure out how to skip these checks in the coordinator altogether since they also happen on the workers.

Fixes #659


metdos:
@jasonmp85 when you have time, could you review this?

src/backend/distributed/executor/multi_executor.c
```diff
@@ -96,6 +97,48 @@ CitusExecutorStart(QueryDesc *queryDesc, int eflags)
...
+ * are executed through SPI, we can check the DestReceiver here.
+ */
+void
+CitusExecutorRun(QueryDesc *queryDesc, ScanDirection direction, uint64 count,
+				 bool executeOnce)
+{
+	DestReceiver *destReceiver = queryDesc->dest;
+	int originalLevel = FunctionCallLevel;
+
+	if (destReceiver->mydest == DestSPI)
```
jasonmp85: Is checking for `DestSPI` sufficient here? In particular, `DestSQLFunction` appears relevant (i.e. does this patch work with pure SQL functions (not PL/pgSQL?)…

I know @anarazel's comment on #659 is pretty old at this point, but at least the way I wrote it up sounded as though we wanted to check if `destReceiver->mydest >= DestSPI` (this would be bad code by our standards, I think, but you get the drift).

Is this patch meant to be comprehensive, or could there still be certain multi-statement transactions which do not do the right thing?

src/backend/distributed/transaction/transaction_management.c
```diff
@@ -545,3 +553,17 @@ SwallowErrors(void (*func)())
...
+/*
+ * IsMultiStatementTransaction determines whether the current statement is
+ * part of a bigger multi-statement transaction. This is the case when the
+ * statement is wrapped in a transaction block (comes after BEGIN), or it
+ * is called from a stored procedure or function.
+ */
+bool
+IsMultiStatementTransaction(void)
+{
+	return IsTransactionBlock() || StoredProcedureLevel > 0 || (FunctionCallLevel > 0 &&
```
jasonmp85: I know what this disjunction means right now, but I'm not sure I'll know in a year or so. I'm struggling to think of a more comprehensible way to write it; the following is very verbose, but I feel it'll remain scrutable in the future?

```c
if (IsTransactionBlock())
{
    /* we're in a BEGIN ... END block, could have multi-stmt */
    return true;
}
else if (StoreProcedureLevel > 0)
{
    /* in a stored procedure */
    return true;
}
else if (FunctionCallLevel > 0)
{
    /* in a language-handler function call, open a transaction if configured to do so */
    return FunctionOpensTransactionBlock;
}
else
{
    /* apparently not in any sort of multi-stmt txn */
    return false;
}
```

The above seems a little crazy, but I do worry about that long boolean thing aging well.

src/backend/distributed/executor/multi_executor.c
marcocitus: I mostly had PL/pgSQL in mind while writing this patch, but including `DestSQLFunction` makes sense. The other tuple destinations don't indicate that there is a multi-statement transaction.

```
    DestSPI          - PL/pgSQL function and other languages
    DestTuplestore   - Mainly for things like RETURNING
    DestIntoRel      - CREATE TABLE tmp AS SELECT ...
    DestCopyOut      - COPY (SELECT ...) TO STDOUT
    DestSQLFunction  - SQL function output
    DestTransientRel - CREATE MATERIALIZED VIEW
    DestTupleQueue   - Parallel worker sends back to parent
```

The only other case I can think of is when there are multiple statements in a single query string. I'm not sure how to detect that reliably. Postgres does a `BeginImplicitTransactionBlock` to set the transaction state to `TBLOCK_IMPLICIT_INPROGRESS`, but there seems to be no way of getting that state. On the other hand, I expect that doing transactions in this manner is somewhat unusual so we can probably live with reduced support.

src/backend/distributed/transaction/transaction_management.c
marcocitus: Makes sense. Will change it

src/backend/distributed/executor/multi_executor.c
marcocitus: Ah, `DestSQLFunction` is only set for the last statement in the function, not the first one :-/

marcocitus: I think covering only PL/pgSQL for now is probably the way to go for now. It at least reduces the hazard quite a bit.

src/backend/distributed/citus--8.2-1--8.2-2.sql
```diff
@@ -0,0 +1,3 @@
+/* citus--8.2-1--8.2-2.sql */
+
+DROP FUNCTION IF EXISTS pg_catalog.create_insert_proxy_for_table(regclass,regclass);
```
onderkalaci: Today I realized that we still have this UDF and the tests. Happy to see that we're removing it within an open PR 😄

sumedhpathak:
@jasonmp85 Ping on this one as well. Thanks.

jasonmp85:
> @jasonmp85 Ping on this one as well. Thanks.

No problem; I'll rebase it, too, since it's out of date now.

jasonmp85:
This is a :shipit: from me. Merging if tests pass, figuring out why not if they don't!

src/test/regress/sql/multi_modifying_xacts.sql
```diff
@@ -1157,4 +1157,25 @@ SELECT * FROM users JOIN usergroups ON (user_group = gid) WHERE id = 2;
 SELECT * FROM users JOIN usergroups ON (user_group = gid) WHERE id = 4;
 END;
 
+-- make sure functions that throw an error roll back propertly
```
pykello: What do you think about adding tests for the following cases?
1. Call to another plpgsql function within a function.
2. Exception handling within a plplgsql function.

src/backend/distributed/shared_library_init.c
```diff
@@ -572,6 +572,21 @@ RegisterCitusConfigVariables(void)
 		GUC_NO_SHOW_ALL,
 		NULL, NULL, NULL);
 
+	DefineCustomBoolVariable(
+		"citus.function_opens_transaction_block",
+		gettext_noop("Open transaction blocks for function calls"),
+		gettext_noop("When enabled, Citus will always send a BEGIN to workers when "
+					 "running distributed queres in a function. When disabled, the "
```
pykello: `queres` -> `queries`

jasonmp85:
@pykello … didn't see your comments before merging; I'll fix the typo in my own work branch and post some output from manually running your suggestions here.

PR #2590, opened 2019-01-15
Relax reference table restrictions in subquery union pushdowns

### LINKED ISSUE #2518: Subquery pushdown planning with UNION/UNION ALL fails with reference tables

mtuncer:
DESCRIPTION: Relax subquery union pushdown restrictions for reference tables

Fixes #2518

### DIFF
--- a/src/backend/distributed/planner/query_pushdown_planning.c
+++ b/src/backend/distributed/planner/query_pushdown_planning.c
@@ -65,6 +65,7 @@ static bool IsNodeQuery(Node *node);
 static bool IsOuterJoinExpr(Node *node);
 static bool WindowPartitionOnDistributionColumn(Query *query);
 static DeferredErrorMessage * DeferErrorIfFromClauseRecurs(Query *queryTree);
+static RecurringTuplesType FromClauseRecurringTupleType(Query *queryTree);
 static DeferredErrorMessage * DeferredErrorIfUnsupportedRecurringTuplesJoin(
 	PlannerRestrictionContext *plannerRestrictionContext);
 static DeferredErrorMessage * DeferErrorIfUnsupportedTableCombination(Query *queryTree);
@@ -572,30 +573,7 @@ DeferErrorIfFromClauseRecurs(Query *queryTree)
 		return NULL;
 	}
 
-	if (FindNodeCheckInRangeTableList(queryTree->rtable, IsDistributedTableRTE))
-	{
-		/*
-		 * There is a distributed table somewhere in the FROM clause.
-		 *
-		 * In the typical case this means that the query does not recur,
-		 * but there are two exceptions:
-		 *
-		 * - outer joins such as reference_table LEFT JOIN distributed_table
-		 * - FROM reference_table WHERE .. (SELECT .. FROM distributed_table) ..
-		 *
-		 * However, we check all subqueries and joins separately, so we would
-		 * find such conditions in other calls.
-		 */
-		return NULL;
-	}
-
-
-	/*
-	 * Try to figure out which type of recurring tuples we have to produce a
-	 * relevant error message. If there are several we'll pick the first one.
-	 */
-	IsRecurringRangeTable(queryTree->rtable, &recurType);
-
+	recurType = FromClauseRecurringTupleType(queryTree);
 	if (recurType == RECURRING_TUPLES_REFERENCE_TABLE)
 	{
 		return DeferredError(ERRCODE_FEATURE_NOT_SUPPORTED,
@@ -641,6 +619,50 @@ DeferErrorIfFromClauseRecurs(Query *queryTree)
 	 */
 
 	return NULL;
+}
+
+
+/*
+ * FromClauseRecurringTupleType returns tuple recurrence information
+ * in query result based on range table entries in from clause.
+ *
+ * Returned information is used to prepare appropriate deferred error
+ * message for subquery pushdown checks.
+ */
+static RecurringTuplesType
+FromClauseRecurringTupleType(Query *queryTree)
+{
+	RecurringTuplesType recurType = RECURRING_TUPLES_INVALID;
+
+	if (queryTree->rtable == NIL)
+	{
+		return RECURRING_TUPLES_EMPTY_JOIN_TREE;
+	}
+
+	if (FindNodeCheckInRangeTableList(queryTree->rtable, IsDistributedTableRTE))
+	{
+		/*
+		 * There is a distributed table somewhere in the FROM clause.
+		 *
+		 * In the typical case this means that the query does not recur,
+		 * but there are two exceptions:
+		 *
+		 * - outer joins such as reference_table LEFT JOIN distributed_table
+		 * - FROM reference_table WHERE .. (SELECT .. FROM distributed_table) ..
+		 *
+		 * However, we check all subqueries and joins separately, so we would
+		 * find such conditions in other calls.
+		 */
+		return RECURRING_TUPLES_INVALID;
+	}
+
+	/*
+	 * Try to figure out which type of recurring tuples we have to produce a
+	 * relevant error message. If there are several we'll pick the first one.
+	 */
+	IsRecurringRangeTable(queryTree->rtable, &recurType);
+
+	return recurType;
 }
 
 
@@ -1062,11 +1084,11 @@ DeferErrorIfUnsupportedUnionQuery(Query *subqueryTree)
 
 		if (IsA(leftArg, RangeTblRef))
 		{
-			Node *leftArgSubquery = NULL;
+			Query *leftArgSubquery = NULL;
 			leftArgRTI = ((RangeTblRef *) leftArg)->rtindex;
-			leftArgSubquery = (Node *) rt_fetch(leftArgRTI,
-												subqueryTree->rtable)->subquery;
-			if (HasRecurringTuples(leftArgSubquery, &recurType))
+			leftArgSubquery = rt_fetch(leftArgRTI, subqueryTree->rtable)->subquery;
+			recurType = FromClauseRecurringTupleType(leftArgSubquery);
+			if (recurType != RECURRING_TUPLES_INVALID)
 			{
 				break;
 			}
@@ -1074,11 +1096,11 @@ DeferErrorIfUnsupportedUnionQuery(Query *subqueryTree)
 
 		if (IsA(rightArg, RangeTblRef))
 		{
-			Node *rightArgSubquery = NULL;
+			Query *rightArgSubquery = NULL;
 			rightArgRTI = ((RangeTblRef *) rightArg)->rtindex;
-			rightArgSubquery = (Node *) rt_fetch(rightArgRTI,
-												 subqueryTree->rtable)->subquery;
-			if (HasRecurringTuples(rightArgSubquery, &recurType))
+			rightArgSubquery = rt_fetch(rightArgRTI, subqueryTree->rtable)->subquery;
+			recurType = FromClauseRecurringTupleType(rightArgSubquery);
+			if (recurType != RECURRING_TUPLES_INVALID)
 			{
 				break;
 			}
--- a/src/test/regress/expected/multi_subquery_complex_reference_clause.out
+++ b/src/test/regress/expected/multi_subquery_complex_reference_clause.out
@@ -150,15 +150,22 @@ WHERE subquery_1.user_id = subquery_2.user_id ;
      4
 (1 row)
 
-  -- Should be able to push down since reference tables are inner joined
-  -- with hash distributed tables, the results of those joins are the parts of
-  -- an outer join
-SELECT subquery_2.id FROM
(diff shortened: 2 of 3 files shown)


src/backend/distributed/planner/query_pushdown_planning.c
```diff
@@ -1367,6 +1367,11 @@ HasRecurringTuples(Node *node, RecurringTuplesType *recurType)
 			return true;
 		}
 
+		if (FindNodeCheckInRangeTableList(query->rtable, IsDistributedTableRTE))
```
marcocitus: I like how small this change is, but this function is becoming a little odd. If I look at the `if (IsA(node, RangeTblEntry))` block it appears as though we return true if any RTE is a reference table. In reality, we already look at all the RTEs here so we would return false if there is a reference table and a distributed table. The function would also behave quite differently if we passed an RTE or rtable directly. It would help to do some refactoring / write this function from scratch.

onderkalaci: Here is my take: I think with this change, we're actually breaking `HasRecurringTuples()` function.

I think the right way to do is refactor `DeferErrorIfFromClauseRecurs()` and its callers such that `if (!queryTree->hasSubLinks)` is not there anymore. And, finally, call `DeferErrorIfFromClauseRecurs()` in `DeferErrorIfUnsupportedUnionQuery()`  instead of `HasRecurringTuples()`.

 `DeferErrorIfFromClauseRecurs()` is already broken (e.g., it doesn't do what its name suggests due to the `if` statement) and if we merge this, we'd break `HasRecurringTuples()`.

I think refactoring  is straight forward, simply pull  `if (!queryTree->hasSubLinks)` to the places where `DeferErrorIfFromClauseRecurs()`. 

Does that make sense to you as well?

mtuncer: Complication of what we are trying to do is causing the issue
We want HasRecurringTuples to return
1 - true : if the query only has a reference table
2 - true : if the query does not reference any table (function, values etc)
3 - false : if the query references at least one distributed table.

Originally, I want to restrict the function such that it will check if the reference table is involved in a join with a distributed table. But we decided to go with presence of a distributed table instead.

`HasRecurringTuples` is called from 3 external places. 1 call from `IsRecurringRangeTable` and 2 calls from  `DeferErrorIfUnsupportedUnionQuery`. Our starting point was to fix the second call.

`IsRecurringRangeTable` is indirectly called from many other places through `RelationInfoContainsOnlyRecurringTuples` and  `RelationInfoContainsRecurringTuples`. Thus, even though the changes are small, it is affects more places it is intended to do.

Given that, it would have been nicer limiting the change with subquery union pushdown.

@onderkalaci I am not following comment. 

> Here is my take: I think with this change, we're actually breaking HasRecurringTuples() function.

Function was too restrictive. It had (still has) many false positives. This change removes some restrictions and allows some false negatives relying that these would be caught later by join checks.

Let me spend some time on your suggestion on not using HasRecurringTuples.

mtuncer: @onderkalaci There are 2 issues with that approach
1 - `DeferErrorIfFromClauseRecurs` only checks distributed table presence in immediate range table entries. Thus if union includes a join wrapped inside another subquery, we won't be able to detect that. This would cause future regressions on this matter.

2 - Less of an issue, we would need to do some more refactoring to return correct error message.

onderkalaci:
> 1 - DeferErrorIfFromClauseRecurs only checks distributed table presence in immediate range table entries. Thus if union includes a join wrapped inside another subquery, we won't be able to detect that. This would cause future regressions on this matter.

But we call it for every single subquery already?
```C
	ExtractQueryWalker((Node *) originalQuery, &subqueryList);
	subqueryList = list_delete(subqueryList, originalQuery);

	/* iterate on the subquery list and error out accordingly */
	foreach(subqueryCell, subqueryList)
	{
		Query *subquery = lfirst(subqueryCell);
		error = DeferErrorIfCannotPushdownSubquery(subquery,
												   outerMostQueryHasLimit);
		if (error)
		{
			return error;
		}
	}
```

marcocitus:
Maybe we should split the problem into two steps, one for correctness and one for a nice error message? 

e.g.
```c
DeferErrorIfUnsupportedUnionQuery
...
if (!FindNodeCheckInRangeTableList(query->rtable, IsDistributedTableRTE))
{
    recurType = FindRecurringTuplesType(leftArgSubquery);
}
...
```

`FindRecurringTuplesType` would basically just be a wrapper around `HasRecurringTuples`. 

Or, if we want to be lazier:
```c

DeferErrorIfUnsupportedUnionQuery
...
if (!FindNodeCheckInRangeTableList(query->rtable, IsDistributedTableRTE))
{
    HasRecurringTuples(leftArgSubquery, &recurType);
}
...
```

onderkalaci:
> Maybe we should split the problem into two steps, one for correctness and one for a nice error message?

We talked with @mtuncer and decided something similar to your suggestion above

src/backend/distributed/planner/query_pushdown_planning.c
```diff
@@ -644,6 +622,50 @@ DeferErrorIfFromClauseRecurs(Query *queryTree)
 }
 
 
+/*
+ * DetermineQueryReturnsRecurringTuples returns tuple recurrence information
```
onderkalaci: Hmm, I think having `Query` in the function name is slightly confusing because we're not actually checking all parts of the query.

name suggestion: `FromClauseRecurringTupleType ()`
function comment suggestion: 
```
FromClauseRecurringTupleType recursively checks the query's range table list
 and returns a valid recurring tuple type if there are no distributed tables 
exist. Note that the function recursively checks all subqueries as well. 

If there are several candidate recurring tuples, we'll simply pick the first one.
```

mtuncer:
that looks like the only solution. I will do that once I got a reliable internet connection.

src/backend/distributed/planner/query_pushdown_planning.c
```diff
@@ -644,6 +622,50 @@ DeferErrorIfFromClauseRecurs(Query *queryTree)
...
+ *
+ * Returned information is used to prepare appropriate deferred error
+ * message for subquery pushdown checks.
+ */
+static RecurringTuplesType
+DetermineQueryReturnsRecurringTuples(Query *queryTree)
+{
+	RecurringTuplesType recurType = RECURRING_TUPLES_INVALID;
+
+	if (queryTree->rtable == NIL)
```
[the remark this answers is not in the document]
mtuncer: if `queryTree->rtable` is NIL walker function [does not call](https://github.com/postgres/postgres/blob/master/src/backend/nodes/nodeFuncs.c#L2313) `IsDistributedTableRTE`

src/backend/distributed/planner/query_pushdown_planning.c
```diff
@@ -644,6 +622,50 @@ DeferErrorIfFromClauseRecurs(Query *queryTree)
...
+		 * find such conditions in other calls.
+		 */
+		return RECURRING_TUPLES_INVALID;
+	}
+
+	/*
+	 * Try to figure out which type of recurring tuples we have to produce a
+	 * relevant error message. If there are several we'll pick the first one.
+	 */
+	IsRecurringRangeTable(queryTree->rtable, &recurType);
```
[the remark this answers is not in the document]
mtuncer: Not mandatory, but good pre-caution. added.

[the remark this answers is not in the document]
mtuncer: Nothing sets `recurType` prior to that point. You probably meant after the function call.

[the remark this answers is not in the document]
mtuncer: I don't think assert is a good idea there.

src/backend/distributed/planner/query_pushdown_planning.c
```diff
@@ -572,30 +573,7 @@ DeferErrorIfFromClauseRecurs(Query *queryTree)
...
-	}
-
-
-	/*
-	 * Try to figure out which type of recurring tuples we have to produce a
-	 * relevant error message. If there are several we'll pick the first one.
-	 */
-	IsRecurringRangeTable(queryTree->rtable, &recurType);
-
+	recurType = DetermineQueryReturnsRecurringTuples(queryTree);
```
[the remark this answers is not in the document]
mtuncer: Think we talked about moving this outside when we wanted to call `DeferErrorIfFromClauseRecurs` for union pushdown too. But later decided to refactor search code into `DetermineQueryReturnsRecurringTuples` and call that function from both sides instead.

So I am more inclined to keep it this way.

src/backend/distributed/planner/query_pushdown_planning.c
```diff
@@ -1367,6 +1389,7 @@ HasRecurringTuples(Node *node, RecurringTuplesType *recurType)
 			return true;
 		}
 
+
```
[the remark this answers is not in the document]
mtuncer: removed

src/backend/distributed/planner/query_pushdown_planning.c
mtuncer: We are also interested in the recurrence type to provide a customized error message.  So I am not in favor for having a bool function. 

Agreed that it could have a better name

src/test/regress/expected/multi_subquery_complex_reference_clause.out
```diff
@@ -150,30 +150,46 @@ WHERE subquery_1.user_id = subquery_2.user_id ;
...
-  -- Should be able to push down since reference tables are inner joined
-  -- with hash distributed tables, the results of those joins are the parts of
-  -- an outer join
-SELECT subquery_2.id FROM
-  (SELECT tt1.user_id, random() FROM user_buy_test_table AS tt1 JOIN users_return_test_table as tt2
-    ON tt1.user_id = tt2.user_id) subquery_1
-  RIGHT JOIN
-  (SELECT tt1.user_id, ref.id, random() FROM user_buy_test_table as tt1 JOIN users_ref_test_table as ref
-   ON tt1.user_id = ref.id) subquery_2 ON subquery_1.user_id = subquery_2.user_id ORDER BY 1 DESC LIMIT 5;
+-- Should be able to push down since reference tables are inner joined
```
onderkalaci: As far as I can see, we don't have tests that triggers the following two cases:

- We're recursively planning an unrelated part of the query, but still be able to pushdown the UNION query
- We're recursively planning an unrelated part of the query, and cannot pushdown the UNION because there is no distribution key equality.

I can hear your complaints that these cases are not directly related to your changes. That's true, they are semi-related since they involve reference table JOIN dist table, and unrelated intermediate results. Still, I'd be happy to have these tests as well.

I'm giving two example queries based on the tables [here](https://github.com/citusdata/citus/issues/2518#issue-387260734). 


```SQL

SELECT *
FROM
  (SELECT test.x,
          test.y
   FROM REF
   JOIN test ON (test.x = ref.a)
   WHERE a IN
       (SELECT y
        FROM test)
   UNION SELECT x,
                y
   FROM test
   WHERE x IN
       (SELECT a
        FROM REF) ) AS fpp;

        
SELECT *
FROM
  (SELECT test.x,
          test.y
   FROM REF
   JOIN test ON (test.x = ref.a)
   WHERE a IN
       (SELECT y
        FROM test)
   UNION SELECT y,
                x
   FROM test
   WHERE x IN
       (SELECT a
        FROM REF) ) AS fpp;
```

PR #2597, opened 2019-01-22
Fix full outer join with subquery pushdown

### LINKED ISSUE #2588: Assert failure on FULL JOIN .. USING
As #2481 is merged we broke some `OUTER JOIN`s that have `USING` clauses. 

When the following queries are run [this assertion](https://github.com/citusdata/citus/blob/1656b519c49062dee5bb5fc8174963821dda9f17/src/backend/distributed/planner/query_pushdown_planning.c#L1654) fails and backend crashes. If assertions are not enabled, we receive a `bogus varno:2` error.

Note: it may be a good idea to add a similar regression test case

```sql
CREATE TABLE t1 ( id int );
CREATE TABLE t2 ( id int );
SELECT create_distributed_table('t1', 'id');
SELECT create_distributed_table('t2', 'id');

-- works
SELECT * FROM t1 FULL JOIN t2 ON (t1.id = t2.id);

-- crashes
SELECT * FROM t1 FULL JOIN t2 USING (id);
```

velioglu:
DESCRIPTION: Fixes a bug that could prevent planning full outer joins with using clause

In the current subquery pushdown logic, we were assuming that each column of the target list of the subquery will be converted to the Var with the `FlattenJoinVars` function. Yet, by utilizing that logic for outer join we have faced with `CoaelesceExpression` in the result of the `FlatterJoinVars` function (you can check the https://github.com/postgres/postgres/blob/master/src/backend/parser/parse_clause.c#L1568-L1610 to understand why `CoalesceExpression` exist), when the tables (full) joined with the using clause. 

This PR handles that problem by adding `CoalesceExpression` awareness to the helper function used as a part of the subquery pushdown logic. 

Fixes #2588


src/backend/distributed/planner/query_pushdown_planning.c
```diff
@@ -1583,13 +1586,28 @@ FlattenJoinVars(List *columnList, Query *queryTree)
...
 			 */
-			memcpy(column, normalizedVar, sizeof(Var));
+
+			if (IsA(normalizedNode, Var))
+			{
+				memcpy(column, normalizedNode, sizeof(Var));
+			}
+			else if (IsA(normalizedNode, CoalesceExpr))
+			{
+				column = repalloc(column, sizeof(CoalesceExpr));
```
thanodnl: As discussed in slack; I think this need to be reworked where the MemoryContext is set to the context that column comes from before calling `flatten_join_alias_vars`.

That function goes into `flatten_join_alias_vars_mutator` which recurses through the object graph and makes a deep copy besides some mutations.

Since `CoalesceExpr` contains a `List *` if will hit:https://github.com/postgres/postgres/blob/97c39498e5ca9208d3de5a443a2282923619bf91/src/backend/nodes/nodeFuncs.c#L2906-L2924 which deep copies the list into the `CurrentMemoryContext`.

So even though we memcpy the contents of `CoalesceExpr` into a struct that lives in the original MemoryContext it points to a `List *` in the `CurrentMemoryContext`.

src/test/regress/expected/multi_subquery_full_join.out
```diff
@@ -0,0 +1,197 @@
...
+  5 |       | 
+(5 rows)
+
+-- Join subqueries using multiple columns
+SELECT * FROM 
+	(SELECT test_table_1.id, test_table_1.val1 FROM test_table_1 FULL JOIN test_table_2 using(id)) as j1
+		FULL JOIN
+	(SELECT test_table_2.id, test_table_2.val1 FROM test_table_1 FULL JOIN test_table_2 using(id)) as j2
+	USING(id, val1)
+	ORDER BY 1;
```
onderkalaci: We need `ORDER BY 1,2` here. Could you go over others as well?

src/backend/distributed/planner/query_pushdown_planning.c
onderkalaci: This `repalloc` was outcome of an optimization/enhancement to the query plan. However, given that it complicates things a lot, I'd prefer not to have `repalloc` or anything equivalent at all, and keep `CoalesceExpr ` on the extentedOpNode as well.

So, in the end, with this repalloc, we're targeting to get something like:

```SQL
SELECT worker_column_1  as id FROM (SELECT Coalesce(table_1.id, table_2.id)) FROM table_1 FULL JOIN t2 USING(id))
```

Without re-palloc, we'd have a redundant `Coalesce` on the master query/top level worker query, which doesn't affect the correctness, but mostly better for readability/debuggibility:
```SQL
SELECT Coalesce(worker_column_1, worker_column_1)  as id FROM (SELECT Coalesce(table_1.id, table_2.id)) FROM table_1 FULL JOIN t2 USING(id))
```

So, @velioglu unless you find a simple implementation for re-allocating the `Coalesce` to `Var`, I'd rather not go that router.

marcocitus: Could we build a new list and return it? I don't think the `repalloc` is correct since it may change the pointer and we're not doing anything with that pointer.

onderkalaci: > Could we build a new list and return it? 

I think no, because the items in the list are actually the `Var`s in the `targetList` and `havingClause` entries, which are used later on the extended op node as well. So, it seems safer to keep `Coalesce ` and get rid of `repalloc`, only use `memcpy` after creating the correct `Coalesce `.

velioglu:
By checking the comments on github & slack and also discussing with @onderkalaci. I have changed the implementation of `FlattenJoinVars` to not to update `columnList` on place but create a new `flattenedExprList`.

src/backend/distributed/planner/query_pushdown_planning.c
```diff
@@ -1583,15 +1593,43 @@ FlattenJoinVars(List *columnList, Query *queryTree)
...
 			 * We need to copy values over existing one to make sure it is updated on
 			 * respective places.
 			 */
-			memcpy(column, normalizedVar, sizeof(Var));
+
+			if (IsA(normalizedNode, Var))
+			{
+				Var *var = makeNode(Var);
+
+				memcpy(var, normalizedNode, sizeof(Var));
```
marcocitus: Is this doing the same as `Var *var = copyObject(normalizedNode);`?

src/test/regress/sql/full_join.sql
```diff
@@ -0,0 +1,91 @@
...
+-- Join subqueries using multiple columns
+SELECT * FROM 
+	(SELECT test_table_1.id, test_table_1.val1 FROM test_table_1 FULL JOIN test_table_3 using(id)) as j1
+		FULL JOIN
+	(SELECT test_table_1.id, test_table_1.val1 FROM test_table_1 FULL JOIN test_table_3 using(id)) as j2
+	USING(id, val1)
+	ORDER BY 1;
+
+-- Full join using multiple columns
+SELECT * FROM test_table_1 FULL JOIN test_table_3 USING(id, val1) ORDER BY 1;
```
onderkalaci: Maybe change one of the tests to have more complicated target list:
```SQL
SELECT count(DISTINCT id), avg(test_table_1.val1) + id * id, id::numeric IS NOT NULL  FROM test_table_1 FULL JOIN test_table_3 using(id) WHERE id::bigint < 55 GROUP BY id ORDER BY 2 ASC LIMIT 3;
```

Or, anything such.

src/test/regress/sql/full_join.sql
```diff
@@ -0,0 +1,91 @@
...
+-- Full outer join with different distribution column types, should error out
+SELECT * FROM test_table_1 full join test_table_2 using(id);
+
+-- Test when the non-distributed column has the value of NULL
+INSERT INTO test_table_1 VALUES(7, NULL);
+INSERT INTO test_table_2 VALUES(7, NULL);
+INSERT INTO test_table_3 VALUES(7, NULL);
+
+-- Get all columns as the result of the full join
+SELECT * FROM test_table_1 FULL JOIN test_table_3 using(id) ORDER BY 1;
```
onderkalaci: For completeness, could you add a single test with `ON (test_table_1.id = test_table_3.id)`

src/test/regress/sql/full_join.sql
```diff
@@ -0,0 +1,91 @@
+--
+-- Full join with subquery pushdown support 
+--
+
+SET citus.next_shard_id TO 9000000;
+
+CREATE TABLE test_table_1(id int, val1 int);
```
onderkalaci: Minor: Given that you're using a generic name for the tables, it might make sense to create them in a separate schema.

src/test/regress/multi_schedule
```diff
@@ -66,7 +66,7 @@ test: multi_basic_queries multi_complex_expressions multi_subquery multi_subquer
 test: multi_subquery_complex_reference_clause multi_subquery_window_functions multi_view multi_sql_function multi_prepare_sql
 test: sql_procedure multi_function_in_join
 test: multi_subquery_in_where_reference_clause
-test: multi_subquery_union multi_subquery_in_where_clause multi_subquery_misc
+test: multi_subquery_union multi_subquery_in_where_clause multi_subquery_misc full_join
```
onderkalaci: Very minor: Could you maybe add the test above line? This line has 3, the above is only 1 currently.

src/backend/distributed/planner/query_pushdown_planning.c
```diff
@@ -1661,28 +1672,85 @@ CreateSubqueryTargetEntryList(List *columnList)
...
+				column->varno = 1;
+				column->varattno = targetEntry->resno;
+				break;
+			}
+		}
+		else if (IsA(targetEntry->expr, CoalesceExpr))
 		{
-			TargetEntry *targetEntry = (TargetEntry *) lfirst(targetEntryCell);
-			Var *targetColumn = NULL;
+			CoalesceExpr *targetCoalesceExpr = (CoalesceExpr *) targetEntry->expr;
```
onderkalaci: Minor: Maybe 1 line comment to remind why we could have `CoalesceExpr ` here

src/test/regress/expected/full_join.out
```diff
@@ -88,6 +90,32 @@ SELECT * FROM test_table_1 FULL JOIN test_table_3 USING(id, val1) ORDER BY 1;
...
+ORDER BY 2
+ASC LIMIT 3;
+ count | avg_value | not_null 
+-------+-----------+----------
+     1 |         2 | t
+     1 |         6 | t
+     1 |        12 | t
+(3 rows)
+
+SELECT max(val1)
```
onderkalaci: Note that the issue I brought up happens on non-full outer joins such as left join. Does this commit consider that?

velioglu: Yes it does, test with left join is added as well.

Issue #2592, opened 2019-01-21
Cannot create a foreign key to a reference table from a partitioned table


mtuncer:
This has started to occur today due to a recent change in PG11. Travis was used to be fine until Jan 18th. Some change afterwards broke the master branch.

Smallest repro I could produce at the moment

```sql
DROP TABLE IF EXISTS partitioning_test, partitioning_test_2009, partitioning_test_2012, partitioning_test_reference CASCADE;


CREATE TABLE partitioning_test(id int, time date) PARTITION BY RANGE (time);
SELECT create_distributed_table('partitioning_test', 'id');

CREATE TABLE partitioning_test_2009 PARTITION OF partitioning_test FOR VALUES FROM ('2009-01-01') TO ('2010-01-01');

-- create a reference table for foreign key test
CREATE TABLE partitioning_test_reference(id int PRIMARY KEY, subid int);
SELECT create_reference_table('partitioning_test_reference');

ALTER TABLE partitioning_test ADD CONSTRAINT partitioning_reference_fkey FOREIGN KEY (id) REFERENCES partitioning_test_reference(id) ON DELETE CASCADE;
```

`ALTER TABLE` command fails with
```
ERROR:  cannot perform DDL on placement 3208, which has been read over multiple connections
```

Creating the foreign key from a partition to a reference table works fine. Problem is with the parent table. This works fine on PG10 but fails on PG11.

This is causing all PRs on travis to fail. We can't merge anything else unless we fix this.

We should also backport the fix to earlier versions that support PG11.

mtuncer:
Looks like  this is a side effect of Alvaro Herrera's recent commit https://github.com/postgres/postgres/commit/0325d7a5957ba39a0dce90835ab54a08ab8bf762 

I haven't confirmed this yet

mtuncer:
Findings so far
- issue occurs only if partitioned table has a partition that does not have the same foreign key created.
- PG tries to validate the foreign key by running a select query on the partition table and the reference to see if there is any violation.
- This query gets to Citus via SPI call when we let the standard_executor run during in process https://github.com/citusdata/citus/blob/master/src/backend/distributed/commands/utility_hook.c#L422
- We end up executing the select query, connection is marked as run a select query
- When we get the control back to actually perform the DDL operation. We detect that a select is already performed in this connection and error out.

We did not have this problem because, postgres did not propagate foreign key creation commands to partitions in the past.

Still thinking about how we should fix this.

mtuncer:
[...]

PR #2599, opened 2019-01-23
Fix partitioned table operations involving foreign key to reference table

mtuncer:
Partial fix for https://github.com/citusdata/citus/issues/2592

No description needed as the issue is not surfaced yet.

### DIFF
--- a/src/backend/distributed/commands/table.c
+++ b/src/backend/distributed/commands/table.c
@@ -1185,6 +1185,26 @@ SetupExecutionModeForAlterTable(Oid relationId, AlterTableCmd *command)
 			{
 				executeSequentially = true;
 			}
+
+			/*
+			 * Postgres performs additional selects when creating constraints
+			 * on partitioned tables. We need to set execution mode to
+			 * sequential for the select query so that ddl connections
+			 * we open does not fail due to previous select.
+			 */
+			if (executeSequentially && PartitionedTable(relationId))
+			{
+				SetLocalMultiShardModifyModeToSequential();
+			}
+		}
+	}
+	else if (alterTableType == AT_DetachPartition)
+	{
+		/* check if there are foreign constraints to reference tables */
+		if (HasForeignKeyToReferenceTable(relationId))
+		{
+			executeSequentially = true;
+			SetLocalMultiShardModifyModeToSequential();
 		}
 	}
 


mtuncer:
This is the minimal changes to make regression tests pass.

However, there is a wider problem with foreign constraints such as attaching a partition to a table with existing foreign keys.

marcocitus:
> However, there is a wider problem with foreign constraints such as attaching a partition to a table with existing foreign keys.

Since it's blocking other PRs, should we get this one merged first?

metdos:
> Since it's blocking other PRs, should we get the current changes merged first?

That's was what we discussed with @mtuncer. Who wants to review it @marcocitus or @onderkalaci :)

src/backend/distributed/commands/table.c
```diff
@@ -1185,6 +1185,26 @@ SetupExecutionModeForAlterTable(Oid relationId, AlterTableCmd *command)
 			{
 				executeSequentially = true;
 			}
+
+			/*
+			 * Postgres performs additional selects when creating constraints
```
onderkalaci: I think this comment is misleading because there nothing really special about partitioned tables, Postgres does this check for any fkey creation (unless created as INVALID).

And, although this change fixes the issue, I see some dark points. For example, in `PlanAlterTableStmt()` function, we're marking this constraint so that it skips the validation just to fix this issue, see the code and the comment:
```
				/*
				 * Foreign constraint validations will be done in workers. If we do not
				 * set this flag, PostgreSQL tries to do additional checking when we drop
				 * to standard_ProcessUtility. standard_ProcessUtility tries to open new
				 * connections to workers to verify foreign constraints while original
				 * transaction is in process, which causes deadlock.
				 */
				constraint->skip_validation = true;
``` 

(Well, it used to lead deadlocks, but today the error you see is expected as far as I can see).

The question is that why does Postgres perform the validation even tough we're trying to skip the validation. If you set the debug2, you'd see `DEBUG:  validating foreign key constraint "partitioning_reference_fkey"`.

And the subtle thing is that we see the error is happening on the partition table even if we create the fkey on the partitioned table (e.g., the SELECT is on `partitioning_test_2009`):
```SQL
ALTER TABLE partitioning_test ADD CONSTRAINT partitioning_reference_fkey FOREIGN KEY (id) REFERENCES partitioning_test_reference(id) ON DELETE CASCADE;
DEBUG:  validating foreign key constraint "partitioning_reference_fkey"
LOG:  issuing COPY (SELECT worker_column_1 AS id FROM (SELECT fk.id AS worker_column_1 FROM (public.partitioning_test_2009_102106 fk LEFT JOIN public.partitioning_test_reference_102138 pk ON ((pk.id OPERATOR(pg_catalog.=) fk.id))) WHERE ((pk.id IS NULL) AND (fk.id IS NOT NULL))) worker_subquery) TO STDOUT
...
```

I think it'd probably makes more sense to understand the change and its affect a bit more before merging this PR.

mtuncer: We do set `skip_validation` flag before we call standard process utility. That is why validation is not kicked in for the partitioned table itself.

However, when creating restrictions on partitions, PG code does not consider that. Instead the foreign keys are read from catalog. see  where [foreign key cloning is initiated](https://github.com/postgres/postgres/blob/REL_11_STABLE/src/backend/commands/tablecmds.c#L7775) and where [FK descriptions are actually read](https://github.com/postgres/postgres/blob/REL_11_STABLE/src/backend/utils/cache/relcache.c#L4097). 

At this time, we can not merge a single PR due to travis failure unless we fix this. 

We should consider this PR as an interim fix.

PG folks still change the code regarding this. We would have to come up with a more proper fix. But the timing is not right now.

pykello: > And, although this change fixes the issue, I see some dark points. For example, in `PlanAlterTableStmt()` function, we're marking this constraint so that it skips the validation just to fix this issue, see the code and the comment:
> 
> ```
> 				/*
> 				 * Foreign constraint validations will be done in workers. If we do not
> 				 * set this flag, PostgreSQL tries to do additional checking when we drop
> 				 * to standard_ProcessUtility. standard_ProcessUtility tries to open new
> 				 * connections to workers to verify foreign constraints while original
> 				 * transaction is in process, which causes deadlock.
> 				 */
> 				constraint->skip_validation = true;
> ```
> 
> (Well, it used to lead deadlocks, but today the error you see is expected as far as I can see).
> 
> The question is that why does Postgres perform the validation even tough we're trying to skip the validation. If you set the debug2, you'd see `DEBUG: validating foreign key constraint "partitioning_reference_fkey"`.

`NOT VALID` is not supported for partitioned tables, so setting `skip_validation` doesn't work. 

```
postgres=# CREATE TABLE ref(a int primary key);
postgres=# CREATE TABLE t1(a int) PARTITION BY LIST (a);
postgres=# CREATE TABLE t1_p1 PARTITION OF t1 FOR VALUES IN (1);
postgres=# ALTER TABLE t1 ADD CONSTRAINT t1_fk FOREIGN KEY (a) REFERENCES ref NOT VALID;

ERROR:  cannot add NOT VALID foreign key on partitioned table "t1" referencing relation "ref" 
DETAIL:  This feature is not yet supported on partitioned tables. 
```

PR #2601, opened 2019-02-01
Add support for column aliases on join clauses

### LINKED ISSUE #2012: distributed joins fail when join alias and column aliases are used
Following query
```sql
select * from (select k1, random()
from (users_table join events_table using (user_id)) k (k1, k2, k3)) l;
```
fails with
```sql
column "k1" specified in USING clause does not exist in left table
```
offending worker query is
```sql
SELECT
    worker_column_1 AS k1,
    worker_column_2 AS random
FROM (
    SELECT
        l.k1 AS worker_column_1,
        l.random AS worker_column_2
    FROM (
        SELECT
            k.k1, random() AS random
        FROM (
            public.users_table_102011 users_table
         JOIN
            public.events_table_102015 events_table
         USING (k1)
         ) k(k1, k2, k3, value_2, value_3, val_4, "time", event_type, value_2_1, value_3_1, value_4)
    ) l
) worker_subquery
WHERE true
```

Notes:
1 - join is given an alias
2 - joined columns are also given aliases
3 - join uses ```using``` clause not ``on``
4 - it occurs on both router and multi-shard queries

Preliminary investigation resulted that this is more like a deparsing problem on Postgres side.

mtuncer:
DESCRIPTION: Add support for column aliases on join clauses

We can now support queries like below where join is given an alias with column names are renamed

```sql
SELECT * FROM (
    SELECT k1, random() FROM (users_table JOIN events_table USING (user_id)) k (k1, k2, k3)) l;
```

Fixes #2012

### DIFF
--- a/src/backend/distributed/utils/ruleutils_10.c
+++ b/src/backend/distributed/utils/ruleutils_10.c
@@ -7134,6 +7134,12 @@ get_from_clause_item(Node *jtnode, Query *query, deparse_context *context)
 			/* Else print column aliases as needed */
 			get_column_alias_list(colinfo, context);
 		}
+		/* check if column's are given aliases in distributed tables */
+		else if (colinfo->parentUsing != NIL)
+		{
+			Assert(colinfo->printaliases);
+			get_column_alias_list(colinfo, context);
+		}
 
 		/* Tablesample clause must go after any alias */
 		if ((rteKind == CITUS_RTE_RELATION || rteKind == CITUS_RTE_SHARD) &&
--- a/src/test/regress/sql/multi_subquery_complex_queries.sql
+++ b/src/test/regress/sql/multi_subquery_complex_queries.sql
@@ -2400,3 +2400,31 @@ FROM
   USING (user_id)
 GROUP BY a.user_id
 ORDER BY 1;
+
+-- queries where column aliases are used
+-- the query is not very complex. join is given an alias with aliases
+-- for each output column
+SELECT k1 
+FROM (
+	SELECT k1, random()
+	FROM (users_table JOIN events_table USING (user_id)) k (k1, k2, k3)) l
+ORDER BY k1
+LIMIT 5;
+
+SELECT DISTINCT k1 
+FROM (
+	SELECT k1, random()
+	FROM (users_table JOIN events_table USING (user_id)) k (k1, k2, k3)) l
+ORDER BY k1
+LIMIT 5;
+
+SELECT x1, x3, value_2
+FROM (users_table u FULL JOIN events_table e ON (u.user_id = e.user_id)) k(x1, x2, x3, x4, x5)
+ORDER BY 1, 2, 3
+LIMIT 5;
+
+SELECT x1, x3, value_2
+FROM (users_table u FULL JOIN events_table e USING (user_id)) k(x1, x2, x3, x4, x5)
+ORDER BY 1, 2, 3
+LIMIT 5;
+
--- a/src/backend/distributed/utils/ruleutils_11.c
+++ b/src/backend/distributed/utils/ruleutils_11.c
@@ -7151,6 +7151,12 @@ get_from_clause_item(Node *jtnode, Query *query, deparse_context *context)
 			/* Else print column aliases as needed */
 			get_column_alias_list(colinfo, context);
 		}
+		/* check if column's are given aliases in distributed tables */
+		else if (colinfo->parentUsing != NIL)
+		{
+			Assert(colinfo->printaliases);
+			get_column_alias_list(colinfo, context);
+		}
 
 		/* Tablesample clause must go after any alias */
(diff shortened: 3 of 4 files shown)


src/backend/distributed/utils/ruleutils_10.c
```diff
@@ -7134,6 +7134,12 @@ get_from_clause_item(Node *jtnode, Query *query, deparse_context *context)
 			/* Else print column aliases as needed */
 			get_column_alias_list(colinfo, context);
 		}
+		/* check if column's are given aliases in distributed tables */
+		else if (colinfo->parentUsing != NIL)
+		{
+			Assert(colinfo->printaliases);
+			get_column_alias_list(colinfo, context);
```
marcocitus: I was wondering why we need the `if (GetRangeTblKind(rte) != CITUS_RTE_SHARD)` condition at all, but that seems to be for merge tables during re-partition joins. The deparsing logic would add more column aliases than there are columns in the merge table. It seems like there is a bug in the physical planner that we're trying to work around here, but I couldn't find a quick fix.

Making an exception for USING seems acceptable, since re-partition joins never have USING.

src/test/regress/sql/multi_subquery_complex_queries.sql
```diff
@@ -2400,3 +2400,20 @@ FROM
   USING (user_id)
 GROUP BY a.user_id
 ORDER BY 1;
+
+-- queries where column aliases are used
+-- the query is not very complex. join is given an alias with aliases
+-- for each output column
+SELECT k1 
+FROM (
```
onderkalaci: now non-subquery outer joins also goes through from query pushdown planning (#2481). So, I think it is valuable to add one more test with such a query.

onderkalaci: Two example queries I've used which might save some time for you. 
The tables will be in the regression test after #2597: 
```SQL
SELECT *
FROM (test_table_1 AS t1
      FULL JOIN test_table_3 AS t2 ON (t1.id = t2.id
                                       AND t1.val1=t2.val1)) k(x, y, z, t)
ORDER BY 1, 2, 3, 4;


SELECT *
FROM (test_table_1 AS t1
      FULL JOIN test_table_3 AS t2 USING (val1,
                                          id)) k(x, y)
ORDER BY 1,
         2;
```

mtuncer: added

onderkalaci:
@mtuncer and @metdos  I hit the following error:

```SQL
SELECT c_custkey
FROM   (users_table LEFT OUTER JOIN events_table ON (users_table.user_id = events_table.user_id)) AS
       test(c_custkey, c_nationkey)
       INNER JOIN users_table as u2 ON (test.c_custkey = u2.user_id)
LIMIT 10;

WARNING:  invalid reference to FROM-clause entry for table "users_table"
HINT:  There is an entry for table "users_table", but it cannot be referenced from this part of the query.
WARNING:  invalid reference to FROM-clause entry for table "users_table"
HINT:  There is an entry for table "users_table", but it cannot be referenced from this part of the query.
WARNING:  invalid reference to FROM-clause entry for table "users_table"
HINT:  There is an entry for table "users_table", but it cannot be referenced from this part of the query.
WARNING:  invalid reference to FROM-clause entry for table "users_table"
HINT:  There is an entry for table "users_table", but it cannot be referenced from this part of the query.
ERROR:  failed to execute task 2
```

And the worker query is
```SQL
SELECT worker_column_1 AS c_custkey
FROM
  (SELECT users_table.user_id AS worker_column_1
   FROM ((public.users_table_102020 users_table
          LEFT JOIN public.events_table_102052 events_table ON ((users_table.user_id OPERATOR(pg_catalog.=) events_table.user_id))) test(c_custkey, c_nationkey, value_1, value_2, value_3, value_4, user_id, "time", event_type, value_2_1, value_3_1, value_4_1)
         JOIN public.users_table_102020 u2 ON ((test.c_custkey OPERATOR(pg_catalog.=) u2.user_id)))) worker_subquery
LIMIT '10'::bigint
```

Where the error comes from `SELECT users_table.user_id AS worker_column_1` since it should have been `test` which is the alias for the join clause. 

Just to give some more context, the following query works fine where the target list directly picks from the join rte:
```SQL

SELECT test.c_custkey
FROM (users_table
      LEFT OUTER JOIN events_table ON (users_table.user_id = events_table.user_id)) AS test(c_custkey, c_nationkey)
INNER JOIN users_table AS u2 ON (test.c_custkey = u2.user_id)
LIMIT 10;
```

In general, when there are multiple joins, our logic for `FlattenJoinVars()` looks broken. And, also our  `FlattenJoinVars()` is too fragile, we should maybe consider finding a better way of handling this problem.

@metdos Is there anyone avaliable to look into this?

metdos:
> @metdos Is there anyone avaliable to look into this?

I think @mtuncer can take a look given that he has some the context on this.

mtuncer:
we should have the same problem with `Var`s in having clause or window clauses

Issue #1848, opened 2017-12-01
Technical debt reduction items


marcocitus:
Projects that reduce technical debt and make Citus code easier to understand:

- [x] Unified executor (#1273): Get rid of router / multi-shard / real-time / task-tracker / other executors and provide one code path instead.

- [x] Query pushdown (#1706): Get rid of fixed join order planner.

- [x] Fix OR shard pruning bug (#1537): This code is fast, but hard to understand. We may want to make it more understandable as part of fixing the OR problem.

- [x] Implement prepared statement support: We currently have 4 different ways of handling prepared statements and they mostly fail to provide prepared statement benefits. We should be able to reason about query parameters (e.g. does this query prune down to 1 shard once the parameter is known?).

- [x] Merge shard rebalancer extension into Citus enterprise.

- [x] Improve code organisation, especially within the planner. The boundaries between the logical planner, optimiser, physical planner, subquery pushdown planner, router planner, and INSERT...SELECT planner have diffused over time and functions are scattered across files. It makes the code harder to understand.

ozgune:
I'll add a few items to this thread. These items made us waste notable amounts of time as we were trying Citus on different platforms:

* `FinishConnectionEstablishment()` says that it synchronously closes a connection. However, the function loops until a connection is established and polls the socket. In general, most functions related to `connection_management.c` were hard to follow. When opening one connection, one needs to go down 9-10 levels down the call stack. Why can't we just call the related libpq functions? -- don't know, just asking

* We explicitly allocate and use `LWLockTranche`. Example code from pg_stat_statements is simpler and doesn't make those extra allocations. We can significantly simplify our code by following what pg_stat_statements does.

marcocitus:
> Why can't we just call the related libpq functions?

Those don't support cancellation, but I agree that functions like `FinishConnectionEstablishment` are convoluted.

marcocitus:
One more item that came up:

- Remove distinction between original query and rewritten query in logical planner. Subquery pushdown uses original query, logical planner uses rewritten query. The code would be much simpler if we just used one (probably original).

marcocitus:
Most of these items are addressed.

PR #2606, opened 2019-02-15
Introduce fast path router planning

onderkalaci:
DESCRIPTION: Skip standard_planner() for trivial queries

In this context, we define "Fast Path Planning for SELECT" as trivial
queries where Citus can skip relying on the standard_planner() and
handle all the planning.

For router planner, standard_planner() is mostly important to generate
the necessary restriction information. Later, the restriction information
generated by the standard_planner is used to decide whether all the shards
that a distributed query touches reside on a single worker node. However,
standard_planner() does a lot of extra things such as cost estimation and
execution path generations which are completely unnecessary in the context
of distributed planning.

There are certain types of queries where Citus could skip relying on
standard_planner() to generate the restriction information. For queries
in the following format, Citus does not need any information that the
standard_planner() generates:

  SELECT ... FROM single_table WHERE distribution_key = X;  or
  DELETE FROM single_table WHERE distribution_key = X; or
  UPDATE single_table SET value_1 = value_2 + 1 WHERE distribution_key = X;

Note that the queries might not be as simple as the above such that
GROUP BY, WINDOW FUNCIONS, ORDER BY or HAVING etc. are all acceptable. The
only rule is that the query is on a single distributed (or reference) table
and there is a "distribution_key = X;" in the WHERE clause. With that, we
could use to decide the shard that a distributed query touches reside on
a worker node.

Benchmark Results (1 c5.2xlarge +3 c5.4xlarge cluster, scale factor 100, test run: 30 mins each test, 1 c5.9xkarge driver node). Basically, prepare the benchmarks such that the coordinator is CPU bound:

branch | select-only | simple-update | tpcb-like
------------ | -------------| -------------| -------------
master |38354.800621| 9661.9| 4825.449080
fast path router | 47820.992931|10696.494726 |5266.452572
ratio          | 0.80 | 0.90 | 0.91

Remaining Tasks:
- [x] Update Planner's README.md
- [x] Add some tests with reference tables
- [x] Add tests with partitioned tables
- [x] Test with serial Column


src/backend/distributed/planner/fast_path_router_planner.c
```diff
@@ -0,0 +1,391 @@
...
+
+	rangeTableEntry = (RangeTblEntry *) linitial(query->rtable);
+	if (rangeTableEntry->rtekind != RTE_RELATION)
+	{
+		return false;
+	}
+
+	/*
+	 * We don't want to deal with potentially overlapping
+	 * append/range distributed tables.
```
marcocitus: Minor note: range-distributed tables shouldn't overlap. Skipping them makes sense, but comment is a bit misleading.

src/backend/distributed/planner/fast_path_router_planner.c
```diff
@@ -0,0 +1,391 @@
...
+		return false;
+	}
+
+	/*
+	 * Pull all range table entries. We prefer this to make sure that
+	 * there is no subqueries in the any part of the query, including
+	 * the from clause, where clause, target list or HAVING clause etc.
+	 */
+	ExtractRangeTableEntryWalker((Node *) query, &rangeTableList);
+	if (list_length(rangeTableList) != 1)
```
marcocitus: This seems a bit inefficient for what it's trying to do. In the spirit of this PR, I wonder whether there's a faster implementation.

src/backend/distributed/planner/fast_path_router_planner.c
```diff
@@ -0,0 +1,391 @@
...
+	/*
+	 * Distribution column must be used in a simple equality match check and it must be
+	 * place at top level conjustion operator. In simple words, we should have
+	 *	    WHERE dist_key = VALUE [AND  ....];
+	 *
+	 * We also skip this check for reference tables.
+	 */
+	distributionKey = PartitionColumn(distributedTableId, 1);
+	if (distributionKey != NULL &&
+		!ColumnMatchExpressionAtTopLevelConjunction(quals, distributionKey))
```
marcocitus: Wouldn't it be simpler to iterate over a list of implict ANDs and find a `<distribution column> = <Const>` / `<Const> = <distribution column>` clause?

src/backend/distributed/planner/fast_path_router_planner.c
```diff
@@ -0,0 +1,391 @@
...
+	}
+
+	if (IsA(node, OpExpr))
+	{
+		OpExpr *opExpr = (OpExpr *) node;
+		bool simpleExpression = SimpleOpExpression((Expr *) opExpr);
+		bool simpleOpExpressionWithParam = SimpleOpExpressionWithParam((Expr *) opExpr);
+		bool columnInExpr = false;
+
+		if (!(simpleExpression || simpleOpExpressionWithParam))
```
marcocitus: is this distinction relevant? could it one function?

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1886,11 +1889,41 @@ PlanRouterQuery(Query *originalQuery,
 
 	*placementList = NIL;
 
-	prunedRelationShardList = TargetShardIntervalsForQuery(originalQuery,
-														   plannerRestrictionContext->
-														   relationRestrictionContext,
-														   &isMultiShardQuery,
-														   partitionValueConst);
+	if (FastPathRouterQuery(originalQuery))
```
marcocitus: A comment on what's special about this path would be good.

src/backend/distributed/planner/fast_path_router_planner.c
```diff
@@ -0,0 +1,391 @@
...
+		return false;
+	}
+
+	/*
+	 * We want to deal with only very simple select queries. Some of the
+	 * checks might be too restrictive, still we prefer this way.
+	 */
+	if (query->cteList != NIL || query->returningList != NIL ||
+		query->hasSubLinks || query->setOperations != NULL ||
+		query->hasForUpdate || query->hasTargetSRFs ||
```
marcocitus: worth confirming whether supporting FOR UPDATE makes a difference in benchmarks

src/backend/distributed/planner/fast_path_router_planner.c
marcocitus: Also, hasSublinks should catch subqueries in the where/target list/having, shouldn't it?

onderkalaci:
> I'm a bit concerned about potential inefficiencies in `FastPathRouterQuery`, which is now called for all queries (might slow down local queries), and twice for (slow path) distributed queries.

Makes sense.

As we've talked in person, I've simplified the checks to decrease the inefficiencies. I'll also do one benchmark where the query fails in the last step of the `FastPathRouterQuery()` check such as logically incorrect query `SELECT count(*) FROM table WHERE dist_key = 5 AND (unrelated_key = 5 AND dist_key = 10)`;

src/backend/distributed/planner/fast_path_router_planner.c
onderkalaci: We've talked this in person. Noting for readers: `make_ands_implicit()` doesn't recurse into the expression tree. So, if we have `a AND (b and (c and d))` we wouldn't get `a,b,d,c`. Thus, we need to recurse anyway. So, that approach would also have a similar implementation.

onderkalaci:
> As we've talked in person, I've simplified the checks to decrease the inefficiencies. I'll also do one benchmark where the query fails in the last step of the `FastPathRouterQuery()` check such as logically incorrect query `SELECT count(*) FROM table WHERE dist_key = 5 AND (unrelated_key = 5 AND dist_key = 10)`;

The difference is almost unnoticeable:

branch | select-only 
------------ | -------------
master |29604.635222
fast path router | 29172.805833
ratio          | 1.014


For early exists in the checks (`WITH t1 AS(
SELECT count(*) FROM pgbench_accounts WHERE aid=:aid AND (bid=:bid AND aid=:cid)) SELECT * FROM t1;`), even less noticeable, probably noise:

branch | select-only 
------------ | -------------
master |26672.030728
fast path router | 26468.319651
ratio          | 1.007

src/backend/distributed/planner/fast_path_router_planner.c
onderkalaci: For pgbench tests it didn't have any affect (as expected). I'll double check with TPC-C tests, which I'll run next.

onderkalaci:
tpcc test results with scale factor 10, driverCount 128, (1 c5.2xlarge +3 c5.4xlarge cluster, scale factor 100, test run: 5 mins each test, 1 c5.9xkarge driver node).


branch | total transactions
------------ | -------------
master | 89016
fast path router | 95855
fast path router with FOR UPDATE enabled | 101318

src/backend/distributed/planner/fast_path_router_planner.c
```diff
@@ -0,0 +1,409 @@
...
+}
+
+
+/*
+ * ColumnMatchExpressionAtTopLevelConjunction returns true if the query contains an exact
+ * match (equal) expression on the provided column. The function returns true only
+ * if the match expression has an AND relation with the rest of the expression tree.
+ */
+static bool
+ColumnMatchExpressionAtTopLevelConjunction(Node *node, Var *column)
```
marcocitus: Some ideas: `ExpressionFiltersColumnByValue`, `ConjunctionContainsColumnFilter`, `ConjunctionHasSimpleOpExpressionForColumn`

src/backend/distributed/planner/fast_path_router_planner.c
```diff
@@ -0,0 +1,409 @@
...
+	}
+	else if (IsA(node, BoolExpr))
+	{
+		BoolExpr *boolExpr = (BoolExpr *) node;
+		List *argumentList = boolExpr->args;
+		ListCell *argumentCell = NULL;
+
+		if (boolExpr->boolop != AND_EXPR)
+		{
+			return false;
```
marcocitus: maybe small comment on happens here, e.g. We do not descend into boolean expressions other than AND. If the column filter appears in an OR clause, we do not consider it even if it is logically the same as a single value comparison (e.g. `<column> = <Const> OR false`).

src/backend/distributed/planner/fast_path_router_planner.c
```diff
@@ -0,0 +1,409 @@
...
+
+	if (paramClause && paramClause->paramkind != PARAM_EXTERN)
+	{
+		/* we can only handle param_externs */
+		return false;
+	}
+	else if (constantClause && constantClause->constisnull)
+	{
+		/* we can only handle param_externs */
+		return false;
```
marcocitus: stale comment

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1886,11 +1889,46 @@ PlanRouterQuery(Query *originalQuery,
...
+		List *shardIntervalList =
+			TargetShardIntervalForFastPathQuery(originalQuery, partitionValueConst,
+												&isMultiShardQuery);
+
+		/*
+		 * This could only happen when there is a parameter on the distribution key.
+		 * We defer error here, later the planner is forced to use a generic plan
+		 * by assigning arbitrarily high cost to the plan.
+		 */
+		if (UpdateOrDeleteQuery(originalQuery) && isMultiShardQuery)
```
marcocitus: I kind of feel like we should be catching this case earlier (unresolved parameter + UPDATE/DELETE).

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1886,11 +1889,46 @@ PlanRouterQuery(Query *originalQuery,
...
+	 */
+	if (FastPathRouterQuery(originalQuery))
+	{
+		List *shardIntervalList =
+			TargetShardIntervalForFastPathQuery(originalQuery, partitionValueConst,
+												&isMultiShardQuery);
+
+		/*
+		 * This could only happen when there is a parameter on the distribution key.
+		 * We defer error here, later the planner is forced to use a generic plan
```
marcocitus: it's forced to NOT use a generic plan

src/backend/distributed/planner/fast_path_router_planner.c
onderkalaci: I like `ConjunctionContainsColumnFilter()`, if reader is interested in the details, can read the code.

src/backend/distributed/planner/multi_router_planner.c
onderkalaci: What about this new commit: https://github.com/citusdata/citus/pull/2606/commits/1625c32158c9136c516707e2522058c4dde35536

The trick is that `FastPath` queries fail on `QueryPushdownSqlTaskList ()` since that function relies heavily on restrictInfos. So, I ended-up with this hack

marcocitus: I like it less than the previous approach.

onderkalaci: So few options that I can come up:

- Previous commit so that return as early as possible 
- This commit
- Similar to this commit, but get rid of the 'else if (isMultiShardModifyQuery && FastPathRouterQuery(originalQuery))' and change 'QueryPushdownSqlTaskList' so that it doesn't rely on restrict info. Note that non-fast path router update/deletes with parameter on the dist key goes through that path on the 6th execution 

Any other option can you think of?

marcocitus: I think previous commit is probably ok for now. We might want to do some refactoring in a separate PR.

onderkalaci: Makes sense. 

This is the first time I'm touching `dist_key = param` parts and in general I found it hard to follow. This PR is also adding one more edge to it. Already discussed on #1848 so just linking for reference.

src/backend/distributed/planner/fast_path_router_planner.c
onderkalaci: So, it is clear that allowing `hasForUpdate` improves the performance and we should do it.

One edge case is that `MultiRouterPlannableQuery ` usees restrictInfos to disable replicationFactor > 1 and hasForUpdate (except for reference table). 

If I simply remove `query->hasForUpdate` from there, we'd silently allow it.

So, I'd either (a) disable query->hasForUpdate with replication factor > 1 here, (b) refactor `MultiRouterPlannableQuery` which might require query traversal (e.g., call `ExtractRangeTableRelationWalker`), (c) enable hasForUpdate with replication > 1 which I need to look a bit and probably would take more time.

Any preference @marcocitus ?

onderkalaci: > (a) disable query->hasForUpdate with replication factor > 1 here,

For now, I pushed this.

src/backend/distributed/planner/fast_path_router_planner.c
```diff
@@ -0,0 +1,431 @@
...
+	/*
+	 * hasForUpdate is tricky because Citus does support only when
+	 * replication = 1 or reference tables.
+	 */
+	if (query->hasForUpdate)
+	{
+		if (cacheEntry->partitionMethod == DISTRIBUTE_BY_NONE ||
+			SingleReplicatedTable(distributedTableId))
+		{
+			return true;
```
marcocitus: shouldn't we do the checks below?

onderkalaci: Ops, definitely this is wrong. I'll fix it.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -2065,19 +2103,59 @@ GetInitialShardId(List *relationShardList)
...
+
+	/* we're only expecting single shard from a single table */
+	Assert(FastPathRouterQuery(query));
+
+	if (list_length(prunedShardIntervalList) > 1)
+	{
+		*isMultiShardQuery = true;
+	}
+	else if (list_length(prunedShardIntervalList) == 1 &&
+			 queryPartitionValueConst != NULL)
```
marcocitus: should this be `partitionValueConst != NULL`? meaning the caller asked for a partition column value?

src/backend/distributed/planner/fast_path_router_planner.c
marcocitus: Looking at this again. I'd have a slight preference for refactoring `MultiRouterPlannableQuery`, because we're quietly skipping a bunch of checks which will not be obvious to readers of that function and may lead to bugs down the road.

Issue #2608, opened 2019-02-20
Invalid casts for workerNode->groupId


pykello:
We use signed 32 bit integer for `groupId` in Citus metadata tables, but there are some places in code that we cast it to the wrong data type. This issue is to track some of these and fix them.

- Almost anywhere in the code, we cast it to an __unsigned__ 32-bit integer. This is probably ok, I am not sure if there are any commonly used architectures that this might cause a problem.
- In master_metadata_utility.c, we wrongly use `Int64GetDatum()` instead of `Int32GetDatum()` or `UInt32GetDatum()`. This causes problems in ARM Cortex A53 cpus, as reported [here](https://groups.google.com/forum/#!topic/citus-users/4SVx3famE5M).
- In metadata_cache.c -> `GetLocalGroupId()` we use `Oid` data type instead of a 32-bit integer for group Id.

We should fix the last two items, first item is probably low priority.

PR #2609, opened 2019-02-20
Fix some of the casts for groupId

pykello:
A small change which partially addresses #2608.


src/backend/distributed/utils/metadata_cache.c
```diff
@@ -160,7 +160,7 @@ static int WorkerNodeCount = 0;
 static bool workerNodeHashValid = false;
 
 /* default value is -1, for coordinator it's 0 and for worker nodes > 0 */
-static int LocalGroupId = -1;
+static int32 LocalGroupId = -1;
```
pykello: We use `-1` for "not set yet", so I kept this as signed, although in most other places we used unsigned group Id.

marcocitus: I'm kind of wondering whether we should standardize on `int32` for group ID, since integers are always signed at the SQL level.

pykello: done

sumedhpathak:
@pykello How's this PR looking? Thanks.

PR #2610, opened 2019-02-21
Add transactionId based round robin policy 

onderkalaci:
DESCRIPTION: Round-robin task assignment policy relies on local transaction id

    Before this commit, round-robin task assignment policy was relying
    on the taskId. Thus, even inside a transaction, the tasks were
    assigned to different nodes. This was especially problematic
    while reading from reference tables within transaction blocks.
    Because, we had to expand the distributed transaction to many
    nodes that are not necessarily already in the distributed transaction.

### DIFF
--- a/src/backend/distributed/planner/multi_physical_planner.c
+++ b/src/backend/distributed/planner/multi_physical_planner.c
@@ -29,6 +29,7 @@
 #include "catalog/pg_type.h"
 #include "commands/defrem.h"
 #include "commands/sequence.h"
+#include "distributed/backend_data.h"
 #include "distributed/listutils.h"
 #include "distributed/citus_nodefuncs.h"
 #include "distributed/citus_nodes.h"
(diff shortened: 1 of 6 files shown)


src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -5091,6 +5092,34 @@ RoundRobinReorder(Task *task, List *placementList)
...
+
+/*
+ * RoundRobinPerTransactionReorder implements the core of the round-robin assignment policy.
+ * It takes a task and placement list and rotates a copy of the placement list
+ * based on the task's jobId. The rotated copy is returned.
+ */
+static List *
+RoundRobinPerTransactionReorder(Task *task, List *placementList)
+{
+	TransactionId transactionId = GetStableLatestTransactionId();
```
onderkalaci: I don't want to rely on distributed transactionId since it is only set on the execution. And, changing that would probably end up with lots of edge cases.

marcocitus: Makes sense.

src/test/regress/sql/multi_task_assignment_policy.sql
```diff
@@ -9,6 +9,28 @@ SET citus.next_shard_id TO 880000;
...
+       shardOfTheTask text;
+begin
+  for r in execute qry loop
+       IF r LIKE '%port%' THEN
+       portOfTheTask = substring(r, '([0-9]{1,10})');
+    END IF;
+
+    IF r LIKE '%' || table_name || '%' THEN
+       shardOfTheTask = substring(r, '([0-9]{1,10})');
+       return QUERY SELECT  shardOfTheTask || '@' || portOfTheTask ;
```
marcocitus: That's some pretty nasty hacking, but I'll allow it because debug messages are bad.

(would it make sense to use JSON output? or would that make this function super verbose?)

onderkalaci: > That's some pretty nasty hacking, but I'll allow it because debug messages are bad.

The problem is that we cannot use DEBUG outputs because they might change -- because we're now relying on the Postgres' transactionId.

> That's some pretty nasty hacking,

I completely agree this. But, I really couldn't find anything better.

onderkalaci: Any suggestion to simplify it? What I try is simply make sure that `count(DISTINCT)` of `shardId@port` is 1 or 2 depending on the test.

marcocitus: My first suggestion would be to use JSON output in the explain and then cast it to jsonb, but the queries might get pretty nasty and then still require some parsing.

The line parsing using regexes might break if explain output changes (or even if the table name contains a number) and it may not be immediately obvious from the regression test failure that this is the problem. It's probably helpful to at least show the actual shard IDs and ports in the output.

marcocitus: (but I'm fine with merging this as is)

PR #2611, opened 2019-02-22
Make sure to allocate connection keys/values in the connection context

### LINKED ISSUE #2607: NodeConninfoGucAssignHook() might be freeing connectionParams that are in use
This is happening in CitusCloud (if you spin up a dev cluster with a branch from Citus repo -- not the enterprise) very often while doing benchmarks. I guess CitusCloud repeatedly updates  `citus.node_conninfo` which triggers   `NodeConninfoGucAssignHook()`.

Steps to repro:

session-1:
```SQL
SELECT logicalrelid, pg_size_pretty(citus_relation_size(logicalrelid)) from pg_dist_partition ;
 \watch 0.001
```
session-2:
```SQL
alter system set citus.node_conninfo to "sslmode=prefer"; select pg_reload_conf();
```

session-1:
```SQL
DETAIL:  invalid connection option ""
postgres=# 
```

or

```
WARNING:  connection error: ec2-3-90-126-107.compute-1.amazonaws.com:5432
DETAIL:  invalid sslmode value: "citus"
```

onderkalaci:
DESCRIPTION: Fixes an invalidation bug  after setting citus.node_conninfo

Fixes #2607

This is required, otherwise we'd end-up in use-after-free bugs when
an invalidation comes in.

We already do the exact same things for runTime keywords, but not
connValues. And, this patch fixes it.

### DIFF
--- a/src/backend/distributed/connection/connection_configuration.c
+++ b/src/backend/distributed/connection/connection_configuration.c
@@ -280,9 +280,14 @@ GetConnParams(ConnectionHashKey *key, char ***keywords, char ***values,
 	/* first step: copy global parameters to beginning of array */
 	for (paramIndex = 0; paramIndex < ConnParams.size; paramIndex++)
 	{
-		/* copy the keyword&value pointers to the new array */
-		connKeywords[paramIndex] = ConnParams.keywords[paramIndex];
-		connValues[paramIndex] = ConnParams.values[paramIndex];
+		/*
+		 * Copy the keyword&value strings to the new array since they might
+		 * be free()-d when settings are reloaded.
+		 */
+		connKeywords[paramIndex] =
+			MemoryContextStrdup(context, ConnParams.keywords[paramIndex]);
+		connValues[paramIndex] =
+			MemoryContextStrdup(context, ConnParams.values[paramIndex]);
 	}
 
 	/* second step: begin at end of global params and copy runtime ones */
--- a/src/backend/distributed/connection/connection_configuration.c
+++ b/src/backend/distributed/connection/connection_configuration.c
@@ -11,6 +11,7 @@
 #include "postgres.h"
 
 #include "distributed/connection_management.h"
+#include "distributed/hash_helpers.h"
 #include "distributed/metadata_cache.h"
 #include "distributed/worker_manager.h"
 #include "distributed/task_tracker.h"
@@ -112,6 +113,39 @@ AddConnParam(const char *keyword, const char *value)
 	ConnParams.size++;
 
 	ConnParams.keywords[ConnParams.size] = ConnParams.values[ConnParams.size] = NULL;
+}
+
+
+/*
+ * DeallocateConnectionParamHash goes over the ConnParamsHash and deallocates
+ * all parameters and values. Finally, all of the entries in the hash table is
+ * deleted.
+ */
+void
+DeallocateConnectionParamHash(void)
+{
+	ConnParamsHashEntry *entry = NULL;
+	HASH_SEQ_STATUS status;
+	hash_seq_init(&status, ConnParamsHash);
+
+	while ((entry = (ConnParamsHashEntry *) hash_seq_search(&status)) != NULL)
+	{
+		int paramIndex = 0;
+
+		while (!(entry->keywords[paramIndex] == NULL ||
+				 entry->keywords[paramIndex] == '\0'))
+		{
+			pfree(entry->keywords[paramIndex]);
+			pfree(entry->values[paramIndex]);
+
+			++paramIndex;
+		}
+
+		pfree(entry->keywords);
+		pfree(entry->values);
+	}
+
+	hash_delete_all(ConnParamsHash);
 }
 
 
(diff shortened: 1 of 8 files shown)


onderkalaci:
I ensured that the issue https://github.com/citusdata/citus/issues/2607 is fixed with this patch.

src/backend/distributed/connection/connection_configuration.c
```diff
@@ -281,8 +281,10 @@ GetConnParams(ConnectionHashKey *key, char ***keywords, char ***values,
 	for (paramIndex = 0; paramIndex < ConnParams.size; paramIndex++)
 	{
 		/* copy the keyword&value pointers to the new array */
```
marcocitus: stale comment, we're now copying the strings since the originals might be free()-d when settings are reloaded.

src/backend/distributed/connection/connection_configuration.c
```diff
@@ -280,9 +280,14 @@ GetConnParams(ConnectionHashKey *key, char ***keywords, char ***values,
...
 	{
-		/* copy the keyword&value pointers to the new array */
-		connKeywords[paramIndex] = ConnParams.keywords[paramIndex];
-		connValues[paramIndex] = ConnParams.values[paramIndex];
+		/*
+		 * Copy the keyword&value strings to the new array since they might
+		 * be free()-d when settings are reloaded.
+		 */
+		connKeywords[paramIndex] =
+			MemoryContextStrdup(context, ConnParams.keywords[paramIndex]);
```
marcocitus: I'm a bit concerned that we are introducing / adding to an already-existing memory leak because this context is `ConnectionContext`, which is never reset.

onderkalaci: Hmm, that looks like a bigger problem in `GetConnParams()` because everything is allocated in `ConnectionContext`. I'll look into this.

onderkalaci: In master branch, if we repeatedly execute the following commands, we can reproduce memory leak:
```SQL
SELECT pg_reload_conf();
SELECT citus_total_relation_size('poc_country_1');
```

This PR, with its current form, expands the memory leak to include all fields 😄 

I think I'll keep this commit as-is and fix the problem on top of this. I'm still thinking for a simple fix.

onderkalaci: I'd like to give an update on this issue.

First of all, `citus_relation_size()` has a completely unrelated memory leak, which confused me a lot. Here(#2618) is the fix for it.

Initially, I thought that the logic for `AddConnParam()` and/or `ResetConnParams()` in combination with SIGHUP/pg_reload_conf() is broken. However, that logic seems sane now. I was doing my tests with `citus_relation_size()` and as I mentioned above, that confused my on this topic.

Note that the fix in this PR could be seen as a completely separate issue. As I mentioned above, this fix expands the memory leak, but at least makes all elements allocated in the same context. Another fix on top of this should fix the memory leak.

The actual memory leak is happening on `ConnParamsHash` invalidation, in the enterprise repo. We keep allocating connection keywords and values in `ConnectionContext` and never deallocate them. Thus, any invalidation on `ConnParamsHash` makes the leak visible.

On citus-enterprise, if you open two sessions, and execute the followings, you'd see the leak:
```SQL
-- session 1
SELECT count(*) FROM table;
\watch 0.001
```

```SQL
-- session 2
UPDATE pg_dist_poolinfo SET poolinfo = '';
\watch 0.001
```
You'll soon see that session 1 uses a lot of memory, and keeps increasing.

Well, the fix does not seem trivial. We probably need a relatively large refactoring. My current suggestions:

- Per `MultiConnection` keep `connKeywords` and `connValues` in a per connection context (no such context exists yet). Make sure to use those variables for starting the pg connection. While `free` ing the `MultiConnetion`, delete this internal  memory context as well. (Probably a wrapper function doing both.)
- Whenever an invalidation comes, mark the entries invalid. Later, at the beginning or end of `StartConnectionEstablishment()` get rid of the invalid entries. Note that invalidation might come during `StartConnectionEstablishment()` so do not forget and break things further :) 

Or, we could allow memory leaks, but only when the values actually change by checking the values in the callbacks. Though, I don't quite like that idea.

@jasonmp85 did you assigned yourself for the review or do you want to take-over the fixes? Both works fine for me.

jasonmp85: I assigned myself to look over it because I was the author and I remember this code being somewhat delicately and intentionally written. It was more a reminder for myself to look over it and see whether I agree with the reasoning and fixes, since both the feature set and performance goals of this code were pretty precise (needed to be ready to create connections easily, but also resettable for changes in config, adding new users and certs, etc.)

sumedhpathak:
@jasonmp85 Are you reviewing this issue? Could you provide feedback (if any) soon? I'd like to close the lingering PR's

jasonmp85:
> Could you provide feedback (if any) soon? I'd like to close the lingering PRs

Thanks for the reminder; I'd been heads-down migrating to my new laptop and workstation and forgot about some of these.

I'll spend the day doing PRs and see what I can clean up, starting with the two you tagged me in.

jasonmp85:
Looks like this also needs a rebase. Guess I'll find out if my `git` credentials transferred… 😉

jasonmp85:
I got to the bottom of this and probably should take a moment before commenting because this code was correct before.

jasonmp85:
> this code was correct before

Hm, that was an arrogant assumption. I see my mistakes now. I think a simpler fix is possible.

jasonmp85:
Hey @onderkalaci… after looking over this, I found the two root problems in my original code, both stemming from my misunderstanding of signal handling in PostgreSQL.

Basically, I didn't think the hook could be called more than once in a backend; I thought reload only affected future connections (I'm not sure why I thought this, but it explains the code).

By ensuring all runtime connection keys and values are in `ConnectionContext`, and keeping track of _where_ the runtime entries begin, we can easily free them all if we know an entry is invalid.

Additionally, since the global config are _already_ being freed, we need to ensure all hash entries are marked as invalid when a reset occurs.

#2632 addressed both of the above, fixing the use-after-free and the memory leak, I believe. Can you take a look at it today and let me know whether something like that could work to address #2607 and the memory leak? It's not a huge change (I didn't port over your tests yet, was just getting to a point where `check-multi` passed on my local machine).

PR #2616, opened 2019-02-27
Enable CircleCI

jasonmp85:
The configuration for the build is in the YAML file; the changes to the regression runner are backward-compatible with Travis and just add the logic to detect whether our custom (isolation- and vanilla enabled) pkg is present.

DESCRIPTION: Switch CI solution to CircleCI


jasonmp85:
I mostly opened this PR for posterity. I've run like 250 CircleCI workflow runs at this point, and this PR doesn't break anything with Travis, so merging…

Issue #2614, opened 2019-02-27
Reference table (or subquery) left join distributed table error out


colton-citus:
Messages about the following are seen with certain subqueries and CTEs:
```
ERROR:  cannot pushdown the subquery
DETAIL:  There exist a reference table in the outer part of the outer join
```

This behavior can only be reliably reproduced so far when both of the following conditions are met:
1. There are LEFT JOINs between 2 different reference tables
2. There is a subquery or CTE of any sort in the query

The following minimum reproduction script contains multiple queries that each trigger this behavior: 

```
drop table if exists distributed_1;
drop table if exists distributed_2;
drop table if exists reference_1;
drop table if exists reference_2;

create table distributed_1 (col1 int, col2 int, distrib_col int);
create table distributed_2 (col1 int, col2 int, distrib_col int);
create table reference_1 (col1 int, col2 int);
create table reference_2(col1 int, col2 int);


select create_distributed_table('distributed_1','distrib_col');
select create_distributed_table('distributed_2','distrib_col');
select create_reference_table('reference_1');
select create_reference_table('reference_2');



select r1.* from distributed_1 AS d1
LEFT JOIN reference_1 AS r1 ON d1.col2=r1.col2
LEFT JOIN reference_2 AS r2 ON r2.col1 = r1.col1
join (select distrib_col,count(*) from distributed_2 group by distrib_col) d2 ON d2.distrib_col=d1.distrib_col
;

--If there's no aggregate in the join, it complains about subqueries instead


with d2 AS (select distrib_col,count(*) from distributed_2 group by distrib_col)
select r1.* from distributed_1 AS d1
LEFT JOIN reference_1 AS r1 ON d1.col2=r1.col2
LEFT JOIN reference_2 AS r2 ON r2.col1 = r1.col1
join d2 ON d2.distrib_col=d1.distrib_col
;

with d2 AS (select distrib_col,col1 from distributed_2)
select r1.* from distributed_1 AS d1
LEFT JOIN reference_1 AS r1 ON d1.col2=r1.col2
LEFT JOIN reference_2 AS r2 ON r2.col1 = r1.col1
join d2 ON d2.distrib_col=d1.distrib_col
;

with cte_1 AS (select col1 from reference_1)
select r1.* from distributed_1 AS d1
LEFT JOIN reference_1 AS r1 ON d1.col2=r1.col2
LEFT JOIN reference_2 AS r2 ON r2.col1 = r1.col1
[...]

PR #2617, opened 2019-02-27
Add some more regression tests for outer join pushdown

onderkalaci:
I've realized that we haven't explicitly added enough tests in #2481 regarding this [change](https://github.com/citusdata/citus/pull/2481/files#diff-659a9573986a07963c9b7bdcaf12fac3R689), and I've used the queries that are mentioned in #2614.

### DIFF
--- a/src/test/regress/expected/with_join.out
+++ b/src/test/regress/expected/with_join.out
@@ -266,6 +266,84 @@ LIMIT
           1 |        
 (2 rows)
 
+-- some more tests for more complex outer-joins
+-- with reference tables
+CREATE TABLE distributed_1 (col1 int, col2 int, distrib_col int);
+CREATE TABLE distributed_2 (col1 int, col2 int, distrib_col int);
+CREATE TABLE reference_1 (col1 int, col2 int);
+CREATE TABLE reference_2(col1 int, col2 int);
+SELECT create_distributed_table('distributed_1','distrib_col');
+ create_distributed_table 
+--------------------------
+ 
+(1 row)
+
+SELECT create_distributed_table('distributed_2','distrib_col');
+ create_distributed_table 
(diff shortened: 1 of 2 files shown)


src/test/regress/expected/outer_join_pushdown.out
```diff
@@ -0,0 +1,83 @@
+-- ===================================================================
+-- test complex outer join trees
```
mtuncer: we also have multi_outer_join test. would it make sense to add these tests there ?
outer join pushdown tests already spread out to multiple files. since we only introduce 4 queries here. we might consider appending these tests to another file such as multi_outer_join.

onderkalaci: makes sense, that file was not there when I opened this PR.

src/test/regress/expected/with_join.out
```diff
@@ -266,6 +266,84 @@ LIMIT
...
+select count(*) from distributed_1 AS d1
+LEFT JOIN reference_1 AS r1 ON d1.col2=r1.col2
+LEFT JOIN reference_2 AS r2 ON r2.col1 = r1.col1
+join d2 ON d2.distrib_col=d1.distrib_col;
+ count 
+-------
+  1734
+(1 row)
+
+with d2 AS (select distrib_col,col1 from distributed_2)
```
mtuncer: I did not get the the last 2 CTEs relation with the change. The previous CTE query is useful to compare that CTE query and subquery generates the same result. We could do without the latter two.

It is not really important though.

onderkalaci: Given that CTEs become intermediate results, I'm trying to ensure that reference tables and intermediate results behave the same.

PR #2618, opened 2019-02-27
Make sure to clear `PGresult` in missing places

onderkalaci:
This leads to a memory leak otherwise.

DESCRIPTION: Fixes a bug that could lead to memory leak on `citus_relation_size`

To repro:
```SQL
SELECT citus_relation_size('table');
\watch 0.001
```

That seems like a minor issue, so I'm not tagging with backport.


pykello:
This looks good. There's also two other instances of the same issue at:
- `run_from_same_connection.c` -> `GetRemoteProcessId()`
- `worker_data_fetch_protocol.c` -> `TableDDLCommandList()`

PR #2619, opened 2019-03-01
Fix memory leak on connection param invalidations

onderkalaci:
Change memory allocations for to prevent memory leaks discussed [here](https://github.com/citusdata/citus/pull/2611#discussion_r259832277).

There are 4 commits which are small enough to follow. 

I'm not very happy with this PR since now we're allocating up to 1KB per connection and also adding few CPU cycles that happens every time a new connection is established. 

@jasonmp85 and @marcocitus, do you have any alternatives in mind? I'll do some benchmarks to see the overhead the PR adds.

(We also need to add one more commit to enterprise to change the invalidation auth info and pool info )

### DIFF
--- a/src/backend/distributed/connection/connection_management.c
+++ b/src/backend/distributed/connection/connection_management.c
@@ -702,6 +702,8 @@ StartConnectionEstablishment(ConnectionHashKey *key)
 	bool found = false;
 	MultiConnection *connection = NULL;
 	ConnParamsHashEntry *entry = NULL;
+	char **keywords = NULL;
+	char **values = NULL;
 
 	connection = MemoryContextAllocZero(ConnectionContext, sizeof(MultiConnection));
 	connection->internalConnectionContext =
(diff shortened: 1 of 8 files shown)


src/backend/distributed/connection/connection_management.c
```diff
@@ -689,27 +703,40 @@ StartConnectionEstablishment(ConnectionHashKey *key)
...
 
-	connection = MemoryContextAllocZero(ConnectionContext, sizeof(MultiConnection));
-
 	strlcpy(connection->hostname, key->hostname, MAX_NODE_LENGTH);
 	connection->port = key->port;
 	strlcpy(connection->database, key->database, NAMEDATALEN);
 	strlcpy(connection->user, key->user, NAMEDATALEN);
 
+	CopyConnectionParams(connection->internalConnectionContext, &keywords, &values,
+						 entry->keywords, entry->values);
```
marcocitus: It seems strange that we're trying to keep this memory for the lifetime of the connection. Do we use keywords/values at all after `PQconnectStartParams`?

onderkalaci: That's a fair question. Yes, after `PQconnectStartParams ()` we do not need them because they are already [copied](https://github.com/postgres/postgres/blob/554ca6954ecbfc54e7159a609b5ffc7921670185/src/interfaces/libpq/fe-connect.c#L832) to [`PGConn`](https://github.com/postgres/postgres/blob/554ca6954ecbfc54e7159a609b5ffc7921670185/src/interfaces/libpq/fe-connect.c#L720-L727).

I'm undoing that change.

onderkalaci:
I'm closing this draft PR since it is small enough to be reviewed inside #2611. Already pushed the necessary commits there.

PR #2621, opened 2019-03-06
Minor code refactoring to skip certain operations for fast path queries

onderkalaci:
We get %2.0 - 2.5 more throughput on select-only pgbench, data in memory and the coordinator is CPU bottlenecked. One of the tests with 1+2 c5.2xlarge cluster:

master: 44868 tps
branch: 45954 tps

### DIFF
--- a/src/backend/distributed/planner/distributed_planner.c
+++ b/src/backend/distributed/planner/distributed_planner.c
@@ -97,6 +97,7 @@ distributed_planner(Query *parse, int cursorOptions, ParamListInfo boundParams)
 	Query *originalQuery = NULL;
 	PlannerRestrictionContext *plannerRestrictionContext = NULL;
 	bool setPartitionedTablesInherited = false;
+	bool fastPathRouterQuery = false;
 
 	if (cursorOptions & CURSOR_OPT_FORCE_DISTRIBUTED)
 	{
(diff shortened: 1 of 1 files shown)


src/backend/distributed/planner/distributed_planner.c
```diff
@@ -105,13 +106,15 @@ distributed_planner(Query *parse, int cursorOptions, ParamListInfo boundParams)
...
+		fastPathRouterQuery = FastPathRouterQuery(parse);
+
 		/*
 		 * Inserting into a local table needs to go through the regular postgres
 		 * planner/executor, but the SELECT needs to go through Citus. We currently
 		 * don't have a way of doing both things and therefore error out, but do
 		 * have a handy tip for users.
 		 */
-		if (InsertSelectIntoLocalTable(parse))
+		if (!fastPathRouterQuery && InsertSelectIntoLocalTable(parse))
```
marcocitus: wondering whether there is a nicer place for the InsertSelectIntoLocalTable check

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -126,12 +129,24 @@ distributed_planner(Query *parse, int cursorOptions, ParamListInfo boundParams)
...
+		if (!fastPathRouterQuery)
+		{
+			AssignRTEIdentities(parse);
+			originalQuery = copyObject(parse);
 
-		setPartitionedTablesInherited = false;
-		AdjustPartitioningForDistributedPlanning(parse, setPartitionedTablesInherited);
+			setPartitionedTablesInherited = false;
+			AdjustPartitioningForDistributedPlanning(parse,
+													 setPartitionedTablesInherited);
```
marcocitus: I'm slightly concerned that we might introduce subtle bugs in the future because the query tree might be in different states in the router planner.

Worth noting that these functions can be a lot faster:
https://github.com/citusdata/citus/commit/83f81e1fe215c3e79a49927d4d7234e4866f229d

onderkalaci: > Worth noting that these functions can be a lot faster:
83f81e1

Agreed, I think if we decide to get `faster_all_the_things` branch, merging this would not be as affective mentioned in the PR description.

onderkalaci:
> Agreed, I think if we decide to get faster_all_the_things branch, merging this would not be as affective mentioned in the PR description.

Closing this PR in favor of the other [branch](https://github.com/citusdata/citus/compare/faster_all_the_things) since that is simpler.

PR #2623, opened 2019-03-08
Implement direct COPY distributed_table TO STDOUT (better pg_dump)

### LINKED ISSUE #2306: Fast COPY distributed_table TO STDOUT
We currently implement `COPY distributed_table TO STDOUT` by replacing the table name with `SELECT * FROM distributed_table` and running the query through the regular distributed planner and (real-time) executor. A much faster method would be to run commands like `COPY distributed_table_102008 TO STDOUT` against the shards and writing the rows returned by the commands directly to the client. We could do something similar for COPY commands with queries that don't require any aggregation.

A benefit is that it would make `pg_dump` on Citus much more usable, since the data doesn't have to be written to the coordinator to return it.

marcocitus:
DESCRIPTION: Implement direct COPY distributed_table TO STDOUT (better pg_dump)

To dump all the data in a table, `pg_dump` issues `COPY table TO STDOUT` commands. Citus rewrites `COPY distributed_table TO STDOUT` to `COPY (SELECT * FROM distributed_table) TO STDOUT` and then executes it via the real-time executor. That means the contents of each shard are written to a file on the coordinator, then to a tuple store, and finally to the client. When there is a large distributed table, that will cause the coordinator to run out of disk space.

This PR fixes the problem by propagating the `COPY distributed_table TO STDOUT` command to each shard (sequentially) and forwarding the result byte-by-byte.

This can make pg_dump 5-10x faster in typical situations and has a much lower memory and CPU footprint. Most importantly, it does not use any disk space on the coordinator.

This approach does have some downsides:
- Snapshots are further apart
- Loss of (I/O) parallelism
- Binary COPY, and COPY .. TO file/program fall back to real-time executor because the added complexity did not seem worth it

The first item could be resolved by using repeatable read, but that will require some more time.

Fixes #2306

### DIFF
--- a/src/backend/distributed/commands/multi_copy.c
+++ b/src/backend/distributed/commands/multi_copy.c
@@ -80,6 +80,7 @@
 #include "distributed/worker_protocol.h"
 #include "executor/executor.h"
 #include "foreign/foreign.h"
+#include "libpq/libpq.h"
 #include "libpq/pqformat.h"
 #include "nodes/makefuncs.h"
 #include "tsearch/ts_locale.h"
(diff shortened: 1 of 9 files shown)


pykello:
@marcocitus why are we closing this?

seamusabshere:
@marcocitus what version of citus is this available in?

pykello:
@seamusabshere this is not available in any of Citus releases yet. I will send an email to our product management to see if it gets prioritized.

PR #2629, opened 2019-03-13
Don't execute ALTER TABLE constraint checks in coordinator.

### LINKED ISSUE #2600: Fix foreign key constraints to reference tables in partitioned tables
We have recently faced https://github.com/citusdata/citus/issues/2592 . Due to a change in PGs regarding foreign keys in partitioned tables.

PG used to skip foreign key push down to partitions when a FK is created on the parent partition. Now it propagates FK constraint creation down to individual partitions.

A side effect of this is the validation part. When we handle FKs, we set `skip_validation` flag before letting standard processing to execute. However, after the change, PG tries to validate each partition against any violations via running a select query.

When this happens, we try to perform a DDL operation on a shard that has already seen a select on a different connection, therefore we bail out preventing the DDL operation on a partitioned tables.

We partially fixed this by forcing sequential execution. It let us perform FK creation and detach partition operations. However, attach partition is still broken due to same issue. 

Longer term fix is to have a flag that would block such distributed select queries that come from standard processing during DDL (alter table) operations.

Note that there is still some work going on PG side when we open this issue.  I am opening this issue to not to loose track and provide the proper fix.

pykello:
These constraints will be validated in worker nodes, so no need to run them on coordinator.

Also fixes distributed deadlock for `ALTER TABLE ... ATTACH PARTITION`.

Fixes #2600

DESCRIPTION: Don't do redundant ALTER TABLE consistency checks at coordinator.

### DIFF
--- a/src/backend/distributed/commands/utility_hook.c
+++ b/src/backend/distributed/commands/utility_hook.c
@@ -57,6 +57,7 @@
 
 bool EnableDDLPropagation = true; /* ddl propagation is enabled */
 static bool shouldInvalidateForeignKeyGraph = false;
+static int activeAlterTables = 0;
 
 
 /* Local functions forward declarations for helper functions */
@@ -419,8 +420,32 @@ multi_ProcessUtility(PlannedStmt *pstmt,
 	}
 
 	pstmt->utilityStmt = parsetree;
-	standard_ProcessUtility(pstmt, queryString, context,
-							params, queryEnv, dest, completionTag);
+
+	PG_TRY();
+	{
+		if (IsA(parsetree, AlterTableStmt))
+		{
+			activeAlterTables++;
+		}
+
+		standard_ProcessUtility(pstmt, queryString, context,
+								params, queryEnv, dest, completionTag);
+
+		if (IsA(parsetree, AlterTableStmt))
+		{
+			activeAlterTables--;
+		}
+	}
+	PG_CATCH();
+	{
+		if (IsA(parsetree, AlterTableStmt))
+		{
+			activeAlterTables--;
+		}
+
+		PG_RE_THROW();
+	}
+	PG_END_TRY();
 
 	/*
 	 * We only process CREATE TABLE ... PARTITION OF commands in the function below
@@ -748,4 +773,15 @@ DDLTaskList(Oid relationId, const char *commandString)
 	}
 
 	return taskList;
+}
+
+
+/*
+ * AlterTableInProgress returns true if we're processing an ALTER TABLE command
+ * right now.
+ */
+bool
+AlterTableInProgress(void)
+{
+	return activeAlterTables > 0;
 }
--- a/src/test/regress/expected/foreign_key_restriction_enforcement.out
+++ b/src/test/regress/expected/foreign_key_restriction_enforcement.out
@@ -511,14 +511,14 @@ ERROR:  cannot execute DDL on reference relation "referece_table" because there 
 HINT:  Try re-running the transaction with "SET LOCAL citus.multi_shard_modify_mode TO 'sequential';"
 ROLLBACK;
 -- case 5.4: Parallel UPDATE on distributed table follow by a related DDL on reference table
--- FIXME: Can we do better?
 BEGIN;
 	UPDATE on_update_fkey_table SET value_1 = 16 WHERE value_1 = 15;
 	ALTER TABLE referece_table ALTER COLUMN id SET DATA TYPE smallint;
 DEBUG:  rewriting table "referece_table"
 DEBUG:  building index "referece_table_pkey" on table "referece_table"
 DEBUG:  validating foreign key constraint "fkey"
-ERROR:  cannot perform DDL on placement 2380001, which has been read over multiple connections
+ERROR:  cannot execute DDL on reference relation "referece_table" because there was a parallel DML access to distributed relation "on_update_fkey_table" in the same transaction
+HINT:  Try re-running the transaction with "SET LOCAL citus.multi_shard_modify_mode TO 'sequential';"
 ROLLBACK;
 -- case 6:1: Unrelated parallel DDL on distributed table followed by SELECT on ref. table
 BEGIN;
--- a/src/test/regress/sql/foreign_key_restriction_enforcement.sql
+++ b/src/test/regress/sql/foreign_key_restriction_enforcement.sql
@@ -303,7 +303,6 @@ BEGIN;
 ROLLBACK;
 
 -- case 5.4: Parallel UPDATE on distributed table follow by a related DDL on reference table
--- FIXME: Can we do better?
 BEGIN;
 	UPDATE on_update_fkey_table SET value_1 = 16 WHERE value_1 = 15;
 	ALTER TABLE referece_table ALTER COLUMN id SET DATA TYPE smallint;
--- a/src/test/regress/sql/multi_partitioning.sql
+++ b/src/test/regress/sql/multi_partitioning.sql
@@ -1127,6 +1127,58 @@ GROUP BY
 ORDER BY
 	1,2;
 
+
+-- test we don't deadlock when attaching and detaching partitions from partitioned
+-- tables with foreign keys
+CREATE TABLE reference_table(id int PRIMARY KEY);
+SELECT create_reference_table('reference_table');
+
+CREATE TABLE reference_table_2(id int PRIMARY KEY);
+SELECT create_reference_table('reference_table_2');
+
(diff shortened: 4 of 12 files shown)


src/test/regress/sql/multi_partitioning.sql
```diff
@@ -1127,6 +1127,23 @@ GROUP BY
 ORDER BY
 	1,2;
 
+
+-- test attaching and detaching partitions from partitioned tables with foreign keys
```
pykello: These are to check we don't have a distributed deadlock in these cases. I should add an explicit comment for that.

src/test/regress/sql/foreign_key_restriction_enforcement.sql
```diff
@@ -303,7 +303,6 @@ BEGIN;
 ROLLBACK;
 
 -- case 5.4: Parallel UPDATE on distributed table follow by a related DDL on reference table
--- FIXME: Can we do better?
```
pykello: We're getting a better error message now, so I guess this is fixed.

src/test/regress/expected/foreign_key_restriction_enforcement.out
```diff
@@ -511,14 +511,14 @@ ERROR:  cannot execute DDL on reference relation "referece_table" because there
...
 -- case 5.4: Parallel UPDATE on distributed table follow by a related DDL on reference table
--- FIXME: Can we do better?
 BEGIN;
 	UPDATE on_update_fkey_table SET value_1 = 16 WHERE value_1 = 15;
 	ALTER TABLE referece_table ALTER COLUMN id SET DATA TYPE smallint;
 DEBUG:  rewriting table "referece_table"
 DEBUG:  building index "referece_table_pkey" on table "referece_table"
 DEBUG:  validating foreign key constraint "fkey"
-ERROR:  cannot perform DDL on placement 2380001, which has been read over multiple connections
+ERROR:  cannot execute DDL on reference relation "referece_table" because there was a parallel DML access to distributed relation "on_update_fkey_table" in the same transaction
```
pykello: This seems to be a better error message, so I think it is ok?

pykello:
I'll split this into two commits after the code review process is done: (1) Fixes the `ATTACH PARTITION` distributed deadlock, (2) Don't run distributed constraint checks.

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -97,12 +105,51 @@ distributed_planner(Query *parse, int cursorOptions, ParamListInfo boundParams)
...
+		 * no need to run expensive distributed queries from coordinator.
+		 *
+		 * For example, ALTER TABLE ... ATTACH PARTITION checks that the new
+		 * partition doesn't violate constraints of the parent table, which
+		 * might involve running some SELECT queries. Ideally we'd completely
+		 * skip these checks in the coordinator, but we don't have any means to
+		 * tell postgres to skip the checks. So the best we can do is to execute
+		 * these queries locally.
+		 *
+		 * See the comments for distributedPlannerEnabled for more information.
```
pykello: This is an outdated comment. I'll remove this.

pykello:
Hey @mtuncer , since you have worked on this issue before, can you please take a look?

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -97,12 +98,49 @@ distributed_planner(Query *parse, int cursorOptions, ParamListInfo boundParams)
...
 	PlannerRestrictionContext *plannerRestrictionContext = NULL;
 	bool setPartitionedTablesInherited = false;
+	bool constraintCheckQuery = false;
 
 	if (cursorOptions & CURSOR_OPT_FORCE_DISTRIBUTED)
 	{
 		needsDistributedPlanning = true;
 	}
 
+	if (AlterTableInProgress() && parse->commandType == CMD_SELECT &&
```
onderkalaci: could we move this logic into a function maybe? It is in the `distributed_planner ()` and looks like too much information for readers of this function.

src/backend/distributed/commands/utility_hook.c
```diff
@@ -419,8 +420,32 @@ multi_ProcessUtility(PlannedStmt *pstmt,
 	}
 
 	pstmt->utilityStmt = parsetree;
-	standard_ProcessUtility(pstmt, queryString, context,
-							params, queryEnv, dest, completionTag);
+
+	PG_TRY();
+	{
+		if (IsA(parsetree, AlterTableStmt))
```
onderkalaci: Is it ever possible that `activeAlterTables > 1`?

If not, in order to get rid of pg_try/pg_catch, can we simply use a boolean flag which is set to false at the beginning of the function, and set to true here?

mtuncer: usually we should not. But if we have case of cascading alter table's we could. I am in favor of keeping this

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -97,12 +98,49 @@ distributed_planner(Query *parse, int cursorOptions, ParamListInfo boundParams)
...
+		 * no need to run expensive distributed queries from coordinator.
+		 *
+		 * For example, ALTER TABLE ... ATTACH PARTITION checks that the new
+		 * partition doesn't violate constraints of the parent table, which
+		 * might involve running some SELECT queries. Ideally we'd completely
+		 * skip these checks in the coordinator, but we don't have any means to
+		 * tell postgres to skip the checks. So the best we can do is to execute
+		 * these queries locally.
+		 */
+		needsDistributedPlanning = false;
```
marcocitus: do we need both this and the LIMIT 0 or would LIMIT 0 be sufficient? (could we simplify the code?)

src/backend/distributed/commands/utility_hook.c
pykello: > Is it ever possible that `activeAlterTables > 1`?
> 

Yes, if there is a trigger that gets fired on some `ALTER TABLE` command which itself does another `ALTER TABLE`.

> If not, in order to get rid of pg_try/pg_catch, can we simply use a boolean flag which is set to false at the beginning of the function, and set to true here?

I didn't get it. It seems to me even with a boolean variable we need to reset it to `false` after this function is done, so we need to `PG_TRY()`/`PG_CATCH()`?

src/backend/distributed/planner/distributed_planner.c
marcocitus: Alternatively, it would be nice if we could handle this case in ExecutorRun somehow. We have a similarly nasty hack in ExecutorStart.

`distributed_planner` is the main entry point into Citus so it would be nice to keep it free of clutter.

src/backend/distributed/planner/distributed_planner.c
pykello: This is sufficient, but we would get `push down of limit count: 0` debug messages after `ALTER TABLE`s, which will be confusing.

I am trying to see if we can do this in `ExecutorRun()` as you suggested.

marcocitus:
I'm happy about the ExecutorRun change.

src/test/regress/sql/multi_partitioning.sql
```diff
@@ -1127,6 +1127,24 @@ GROUP BY
 ORDER BY
 	1,2;
 
+
+-- test we don't deadlock when attaching and detaching partitions from partitioned
+-- tables with foreign keys
+DROP TABLE reference_table CASCADE;
```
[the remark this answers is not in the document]
pykello: done

src/test/regress/sql/multi_partitioning.sql
```diff
@@ -1127,6 +1127,24 @@ GROUP BY
...
+CREATE TABLE reference_table(id int PRIMARY KEY);
+SELECT create_reference_table('reference_table');
+
+CREATE TABLE partitioning_test(id int, time date) PARTITION BY RANGE (time);
+CREATE TABLE partitioning_test_2009 (LIKE partitioning_test);
+
+ALTER TABLE partitioning_test ADD CONSTRAINT partitioning_reference_fkey FOREIGN KEY (id) REFERENCES reference_table(id) ON DELETE CASCADE;
+ALTER TABLE partitioning_test_2009 ADD CONSTRAINT partitioning_reference_fkey_2009 FOREIGN KEY (id) REFERENCES reference_table(id) ON DELETE CASCADE;
+
+ALTER TABLE partitioning_test ATTACH PARTITION partitioning_test_2009 FOR VALUES FROM ('2009-01-01') TO ('2010-01-01');
```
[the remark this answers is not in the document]
pykello: done

PR #2631, opened 2019-03-15
Fix column references to aliased joins

mtuncer:
This is  continuation of https://github.com/citusdata/citus/pull/2601 

Initial fix did not consider aliased join expressions participating in other joins. Now we walk over join tree trying to find a join with alias defined. If none found we can go deep to the lowest level (relation range table entry) as before.

### DIFF
--- a/src/backend/distributed/planner/query_pushdown_planning.c
+++ b/src/backend/distributed/planner/query_pushdown_planning.c
@@ -80,6 +80,7 @@ static bool IsRecurringRangeTable(List *rangeTable, RecurringTuplesType *recurTy
 static bool HasRecurringTuples(Node *node, RecurringTuplesType *recurType);
 static MultiNode * SubqueryPushdownMultiNodeTree(Query *queryTree);
 static List * FlattenJoinVars(List *columnList, Query *queryTree);
+static Node * FlattenJoinVarsMutator(Node *node, Query *queryTree);
 static void UpdateVarMappingsForExtendedOpNode(List *columnList,
 											   List *flattenedColumnList,
 											   List *subqueryTargetEntryList);
@@ -1582,52 +1583,71 @@ SubqueryPushdownMultiNodeTree(Query *queryTree)
 static List *
 FlattenJoinVars(List *columnList, Query *queryTree)
 {
-	ListCell *columnCell = NULL;
-	List *rteList = queryTree->rtable;
 	List *flattenedExprList = NIL;
 
+	ListCell *columnCell = NULL;
 	foreach(columnCell, columnList)
 	{
-		Var *column = (Var *) lfirst(columnCell);
-		RangeTblEntry *columnRte = NULL;
-		PlannerInfo *root = NULL;
-
-		Assert(IsA(column, Var));
-
-		/*
-		 * if join does not have an alias, it is copied over join rte.
-		 * There is no need to find the JoinExpr to check whether it has
-		 * an alias defined.
-		 *
-		 * We use the planner's flatten_join_alias_vars routine to do
-		 * the flattening; it wants a PlannerInfo root node, which
-		 * fortunately can be mostly dummy.
-		 */
-		columnRte = rt_fetch(column->varno, rteList);
-		if (columnRte->rtekind == RTE_JOIN && columnRte->alias == NULL)
-		{
-			Node *normalizedNode = NULL;
-
-			if (root == NULL)
-			{
-				root = makeNode(PlannerInfo);
-				root->parse = (queryTree);
-				root->planner_cxt = CurrentMemoryContext;
-				root->hasJoinRTEs = true;
-			}
-
-			normalizedNode = strip_implicit_coercions(flatten_join_alias_vars(root,
-																			  (Node *)
-																			  column));
-			flattenedExprList = lappend(flattenedExprList, copyObject(normalizedNode));
-		}
-		else
-		{
-			flattenedExprList = lappend(flattenedExprList, copyObject(column));
-		}
+		Node *column = strip_implicit_coercions(
+			FlattenJoinVarsMutator((Node *) lfirst(columnCell), queryTree));
+		flattenedExprList = lappend(flattenedExprList, copyObject(column));
 	}
 
 	return flattenedExprList;
+}
+
+
+/*
+ * FlattenJoinVarsMutator flattens a single column var as outlined in the caller
+ * function (FlattenJoinVars). It iterates the join tree to find the
+ * lowest Var it can go. This is usually the relation range table var. However
+ * if a join operation is given an alias, iteration stops at that level since the
+ * query can not reference the inner RTE by name if the join is given an alias.
+ */
+static Node *
+FlattenJoinVarsMutator(Node *node, Query *queryTree)
+{
+	if (node == NULL)
+	{
+		return NULL;
+	}
+
+	if (IsA(node, Var))
+	{
+		Var *column = (Var *) node;
+		RangeTblEntry *rte = rt_fetch(column->varno, queryTree->rtable);
+		if (rte->rtekind == RTE_JOIN)
+		{
+			Node *newColumn = NULL;
+
+			/*
+			 * if join has an alias, it is copied over join RTE. We should
+			 * reference this RTE.
+			 */
+			if (rte->alias != NULL)
+			{
+				return (Node *) column;
+			}
+
+			/* join RTE does not have and alias defined at this level, deeper look is needed */
+			Assert(column->varattno > 0);
+			newColumn = (Node *) list_nth(rte->joinaliasvars, column->varattno - 1);
+			Assert(newColumn != NULL);
+
+			/*
+			 * Ideally we should use expression_tree_mutator here. But it does not call
+			 * mutate function for Vars, thus we make a recursive call to make sure
+			 * not to miss Vars in nested joins.
+			 */
+			return FlattenJoinVarsMutator(newColumn, queryTree);
+		}
+		else
+		{
+			return node;
+		}
+	}
+
+	return expression_tree_mutator(node, FlattenJoinVarsMutator, (void *) queryTree);
 }
 
 
@@ -1733,13 +1753,12 @@ UpdateColumnToMatchingTargetEntry(Var *column, Node *flattenedExpr, List *target
 		else if (IsA(targetEntry->expr, CoalesceExpr))
 		{
 			/*
-			 * flatten_join_alias_vars() flattens full oter joins' columns that is
+			 * FlattenJoinVars() flattens full oter joins' columns that is
 			 * in the USING part into COALESCE(left_col, right_col)
 			 */
-			CoalesceExpr *targetCoalesceExpr = (CoalesceExpr *) targetEntry->expr;
 
 			if (IsA(flattenedExpr, CoalesceExpr) && equal(flattenedExpr,
-														  targetCoalesceExpr))
+														  targetEntry->expr))
 			{
 				Oid expressionType = exprType(flattenedExpr);
 				int32 expressionTypmod = exprTypmod(flattenedExpr);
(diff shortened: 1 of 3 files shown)


mtuncer:
having issues with CircleCI run. same test works as expected on local machine and on travis, but fails on CircleCI.

metdos:
> having issues with CircleCI run. same test works as expected on local machine and on travis, but fails on CircleCI.

It seems to work fine now. Do you know the reason for previous failures?

pykello:
> > having issues with CircleCI run. same test works as expected on local machine and on travis, but fails on CircleCI.
> 
> It seems to work fine now. Do you know the reason for previous failures?

I'm not sure if Murat's failure was the same, but one of commits few days ago failed at `multi_select_distinct`. I think restarting the build might fix it. The difference a slightly different plan for a join. I think I got the same failure when working on #2600.

mtuncer:
@pykello correct. multi_select_distinct was the failing test. It looks good now.

mtuncer:
@onderkalaci my initial evaluation regarding the failing query is that it looks like it is not directly relevant to this part of the code. But I haven't debugged it in detail.  I should have some information soon.

mtuncer:
Verified that the change is not related to failing query.  We only considered top level var and did not go any deep. This looks more like deparsing issue caused by working on modified parse tree due to standard planning.

src/backend/distributed/planner/query_pushdown_planning.c
```diff
@@ -53,6 +53,16 @@ typedef enum RecurringTuplesType
...
 } RecurringTuplesType;
 
+/*
+ * FlattenJoinVarsMutatorContext stores context variable needed by
+ * FlattenJoinVarsMutator function.
+ */
+typedef struct FlattenJoinVarsMutatorContext
+{
+	Query *queryTree;
+	PlannerInfo *plannerInfo;
```
onderkalaci: We don't use the planner info anymore. To simplify the code, can't we get rid of the context and only pass the `queryTree`?

src/backend/distributed/planner/query_pushdown_planning.c
```diff
@@ -1718,45 +1756,61 @@ UpdateColumnToMatchingTargetEntry(Var *column, Node *flattenedExpr, List *target
...
+					int32 expressionTypmod = exprTypmod(flattenedExpr);
+					Oid expressionCollation = exprCollation(flattenedExpr);
+
+					column->varno = 1;
+					column->varattno = targetEntry->resno;
+					column->vartype = expressionType;
+					column->vartypmod = expressionTypmod;
+					column->varcollid = expressionCollation;
+
+					doneProcessing = true;
```
onderkalaci: Minor readability question: Is there a reason not to return here (or break)? We could get rid of this intermediate variable?

src/test/regress/sql/multi_subquery_complex_queries.sql
```diff
@@ -2428,3 +2428,67 @@ FROM (users_table u FULL JOIN events_table e USING (user_id)) k(x1, x2, x3, x4,
...
+LIMIT 10;
+
+SELECT c_custkey, date_trunc('minute', max(c_nationkey))
+FROM  (users_table LEFT OUTER JOIN events_table ON (users_table.user_id = events_table.user_id)) AS test(c_custkey, c_nationkey)
+	INNER JOIN users_table as u2 ON (test.c_custkey = u2.user_id)
+GROUP BY 1
+ORDER BY 2, 1
+LIMIT 10;
+
+SELECT c_custkey, date_trunc('minute', max(c_nationkey))
```
onderkalaci: I think you forgot to add the FULL JOIN test that you've pushed the fix for?
```SQL
SELECT user_id
FROM (users_table
      JOIN events_table USING (user_id)) AS test(user_id, c_nationkey)
FULL JOIN users_table AS u2 USING (user_id)
LIMIT 10;
```

src/backend/distributed/planner/query_pushdown_planning.c
```diff
@@ -1718,45 +1756,61 @@ UpdateColumnToMatchingTargetEntry(Var *column, Node *flattenedExpr, List *target
...
-			/*
-			 * flatten_join_alias_vars() flattens full oter joins' columns that is
-			 * in the USING part into COALESCE(left_col, right_col)
-			 */
-			CoalesceExpr *targetCoalesceExpr = (CoalesceExpr *) targetEntry->expr;
 
-			if (IsA(flattenedExpr, CoalesceExpr) && equal(flattenedExpr,
-														  targetCoalesceExpr))
+			case T_CoalesceExpr:
+			case T_FuncExpr:
```
onderkalaci: Do we really need `T_FuncExpr `? I think we shouldn't need that. 

The reason you had to add seems like on `FlattenJoinVarsMutator ()` we return the node itself for two cases (e.g., `return (Node *) column` and `return node;`). Instead, we should return copies of them (e.g., `copyObject(node)`). Otherwise, we're losing the modifications on the node, which is `strip_implicit_coercions ()` in this case.

I think, if you do that, we should be able to undo the changes on `UpdateColumnToMatchingTargetEntry ` as well

mtuncer: I tried to remove the need for FuncExpr.  For some reason `pull_var_clause_default` did not reduce `max(var)` into a `var` but kept it as function reference in one of the existing test cases. 

regarding copyObject, `expression_tree_mutator` already does copying itself before calling mutator function.  I will try copyObject and update it accordingly.

src/backend/distributed/planner/query_pushdown_planning.c
mtuncer: makes sense

src/backend/distributed/planner/query_pushdown_planning.c
mtuncer: returning looks better. since it is inside switch, break (already there) only exits from switch block, and still stay at loop block.

src/backend/distributed/planner/query_pushdown_planning.c
mtuncer: query forcing me to have funcexpr in full_join test is
```sql
SELECT max(val1)
FROM test_table_1 LEFT JOIN test_table_3 USING(id, val1)


`val1` has conflicting declarations in joined tables. test_table_1 defines is as int, test_table_3 uses it as bigint.
GROUP BY test_table_1.id
ORDER BY 1;
```
However following is just fine
```sql
SELECT max(val1)
FROM test_table_1 FULL JOIN test_table_3 USING(id, val1)
GROUP BY test_table_1.id
ORDER BY 1;
```

onderkalaci: > regarding copyObject, expression_tree_mutator already does copying itself before calling mutator function. I will try copyObject and update it accordingly.

Oops, I realized something that actually could be dangerous/wrong with the expression mutation logic we've implemented. `expression_tree_mutator()` does not [mutate](https://github.com/postgres/postgres/blob/master/src/backend/nodes/nodeFuncs.c#L2479-L2488) `Var`s. That'd actually break our logic, no? Because you've switch from `while (Var belongs to join){..}`, now we'd exit on the first loop.

So, should we do something like the following?
```SQL
diff --git a/src/backend/distributed/planner/query_pushdown_planning.c b/src/backend/distributed/planner/query_pushdown_planning.c
index 3bba6f25..a12a8452 100644
--- a/src/backend/distributed/planner/query_pushdown_planning.c
+++ b/src/backend/distributed/planner/query_pushdown_planning.c
@@ -1656,8 +1656,8 @@ FlattenJoinVarsMutator(Node *node, FlattenJoinVarsMutatorContext *context)
                        Assert(column->varattno > 0);
                        newColumn = (Node *) list_nth(rte->joinaliasvars, column->varattno - 1);
                        Assert(newColumn != NULL);
-                       return expression_tree_mutator(newColumn, FlattenJoinVarsMutator,
-                                                                                  (void *) context);
+                       /* add a clarifying comment */
+                       return FlattenJoinVarsMutator(newColumn, context);
                }
                else
                {
@@ -1775,7 +1775,7 @@ UpdateColumnToMatchingTargetEntry(Var *column, Node *flattenedExpr, List *target
                        }
 
                        case T_CoalesceExpr:
-                       case T_FuncExpr:
+                       //case T_FuncExpr:
 
                                /*
                                 * FlattenJoinVars() flattens full oter joins' columns that is
```

(And, we'd add `copyObject` before returning the `Var`s because that might not come from expression_tree_mutator with the above logic.)

src/backend/distributed/planner/query_pushdown_planning.c
```diff
@@ -53,6 +53,15 @@ typedef enum RecurringTuplesType
...
 } RecurringTuplesType;
 
+/*
+ * FlattenJoinVarsMutatorContext stores context variable needed by
+ * FlattenJoinVarsMutator function.
+ */
+typedef struct FlattenJoinVarsMutatorContext
+{
+	Query *queryTree;
+} FlattenJoinVarsMutatorContext;
```
onderkalaci: Do we need the struct at all? I thought we'd only pass `queryTree ` as the parameter

src/backend/distributed/planner/query_pushdown_planning.c
```diff
@@ -1582,52 +1592,71 @@ SubqueryPushdownMultiNodeTree(Query *queryTree)
...
 	List *flattenedExprList = NIL;
+	ListCell *columnCell = NULL;
+	FlattenJoinVarsMutatorContext context = { queryTree };
 
 	foreach(columnCell, columnList)
 	{
-		Var *column = (Var *) lfirst(columnCell);
-		RangeTblEntry *columnRte = NULL;
-		PlannerInfo *root = NULL;
+		Node *column = strip_implicit_coercions(
```
onderkalaci: Do we need `strip_implicit_coercions ()` here as well? What is your thought on my [comment](https://github.com/citusdata/citus/pull/2631#discussion_r268566482)? I think if we do that, we shouldn't need this as well.

src/backend/distributed/planner/query_pushdown_planning.c
mtuncer: I was about to remove that. done

src/backend/distributed/planner/query_pushdown_planning.c
mtuncer: we need `strip_jmplicit_coercions` either before calling `FlattenJoinVarsMutator` or after. 
Problem is with `some_var::int8` type casting. It appears as a function. If we provide the node the mutator function we got function expression back with inner var mutated (if needed). So we still need to do `strip_jmplicit_coercions` 

I felt like doing it afterwards. I have no strong opinion against doing it before.

mtuncer: Addition, I still do copyObject after strip....

src/backend/distributed/planner/query_pushdown_planning.c
```diff
@@ -1582,52 +1583,68 @@ SubqueryPushdownMultiNodeTree(Query *queryTree)
...
+ */
+static Node *
+FlattenJoinVarsMutator(Node *node, Query *queryTree)
+{
+	if (node == NULL)
+	{
+		return NULL;
+	}
+
+	node = strip_implicit_coercions(node);
```
mtuncer: guess we might skip this

src/test/regress/sql/multi_subquery_complex_queries.sql
mtuncer: added

src/backend/distributed/planner/query_pushdown_planning.c
onderkalaci: > What is your thought on my comment? 

What I'm saying is that `expression_tree_mutator ` doesn't call mutate function on `Var`s. Don't you think that's a problem?

mtuncer: we handle top level var's in the function. So mutator function is not called with Var node. For everything else it does mutating of the node via FLATCOPY and MUTATE on the child.

we should be fine here.

onderkalaci: Just noting for future readers, we had a quick call with @mtuncer and he agreed to change the call to `FlattenJoinVarsMutator ()`, so marking as resolved.

PR #2632, opened 2019-03-18
Fix conninfo memory bugs

### LINKED ISSUE #2607: NodeConninfoGucAssignHook() might be freeing connectionParams that are in use

jasonmp85:
DESCRIPTION: Address some memory issues in connection config

This basically replaces #2611, addressing both the use-after-free bug in global parameters as well as the memory leak in runtime ones.

The root of the problem was my mental model of signal handling in PostgreSQL was wrong; I thought the config changes were only updated in the postmaster and (therefore) any subsequent forks; I didn't realize children were also signaled and would call e.g. GUC hooks at the top of their main loop.

I've addressed the issues by (a) always allocating runtime parameters in `ConnectionContext`, making it simple for us to `pfree` them in a loop, so long as we keep track of where they start, and (b) adding an invalidation loop into the GUC hook, ensuring any hash entries which may include references to previous global config keywords/values are marked as invalid, forcing their repopulation with valid references before they are used again.

Fixes #2607 
Closes #2611 
Closes #2619

### DIFF
--- a/src/backend/distributed/connection/connection_configuration.c
+++ b/src/backend/distributed/connection/connection_configuration.c
@@ -85,20 +85,9 @@ ResetConnParams()
 		ConnParams.keywords[paramIdx] = ConnParams.values[paramIdx] = NULL;
 	}
 
-	if (ConnParamsHash != NULL)
-	{
-		ConnParamsHashEntry *entry = NULL;
-		HASH_SEQ_STATUS status;
-
-		hash_seq_init(&status, ConnParamsHash);
-		while ((entry = (ConnParamsHashEntry *) hash_seq_search(&status)) != NULL)
-		{
-			entry->isValid = false;
-		}
-	}
-
-
 	ConnParams.size = 0;
+
+	InvalidateConnParamsHashEntries();
 
 	AddConnParam("fallback_application_name", "citus");
 }
@@ -238,7 +227,7 @@ CheckConninfo(const char *conninfo, const char **whitelist,
  */
 void
 GetConnParams(ConnectionHashKey *key, char ***keywords, char ***values,
-			  Index *nonGlobalParamStart, MemoryContext context)
+			  Index *runtimeParamStart, MemoryContext context)
 {
 	/* make space for the port as a string: sign, 10 digits, NUL */
 	char *nodePortString = MemoryContextAlloc(context, 12 * sizeof(char *));
@@ -304,6 +293,9 @@ GetConnParams(ConnectionHashKey *key, char ***keywords, char ***values,
 		connValues[paramIndex] = ConnParams.values[paramIndex];
 	}
 
+	/* remember where global/GUC params end and runtime ones start */
+	*runtimeParamStart = ConnParams.size;
+
 	/* second step: begin at end of global params and copy runtime ones */
 	for (runtimeParamIndex = 0;
 		 runtimeParamIndex < lengthof(runtimeKeywords);
@@ -321,7 +313,6 @@ GetConnParams(ConnectionHashKey *key, char ***keywords, char ***values,
 
 	*keywords = connKeywords;
 	*values = connValues;
-	*nonGlobalParamStart = ConnParams.size;
 }
 
 
--- a/src/backend/distributed/connection/connection_configuration.c
+++ b/src/backend/distributed/connection/connection_configuration.c
@@ -77,15 +77,26 @@ void
 ResetConnParams()
 {
 	Index paramIdx = 0;
-
 	for (paramIdx = 0; paramIdx < ConnParams.size; paramIdx++)
 	{
-		/* FIXME: People still have references to these! */
 		free((void *) ConnParams.keywords[paramIdx]);
 		free((void *) ConnParams.values[paramIdx]);
 
 		ConnParams.keywords[paramIdx] = ConnParams.values[paramIdx] = NULL;
 	}
+
+	if (ConnParamsHash != NULL)
+	{
+		ConnParamsHashEntry *entry = NULL;
+		HASH_SEQ_STATUS status;
+
+		hash_seq_init(&status, ConnParamsHash);
+		while ((entry = (ConnParamsHashEntry *) hash_seq_search(&status)) != NULL)
+		{
+			entry->isValid = false;
+		}
+	}
+
 
 	ConnParams.size = 0;
 
--- a/src/backend/distributed/connection/connection_configuration.c
+++ b/src/backend/distributed/connection/connection_configuration.c
@@ -80,6 +80,7 @@ ResetConnParams()
 
 	for (paramIdx = 0; paramIdx < ConnParams.size; paramIdx++)
 	{
+		/* FIXME: People still have references to these! */
 		free((void *) ConnParams.keywords[paramIdx]);
 		free((void *) ConnParams.values[paramIdx]);
 
@@ -94,15 +95,14 @@ ResetConnParams()
 
 /*
  * AddConnParam adds a parameter setting to the global libpq settings according
- * to the provided keyword and value. Under assert-enabled builds, array bounds
- * checking is performed.
+ * to the provided keyword and value.
  */
 void
 AddConnParam(const char *keyword, const char *value)
 {
 	if (ConnParams.size + 1 >= ConnParams.maxSize)
 	{
-		/* we expect developers to see that error messages */
+		/* hopefully this error is only seen by developers */
 		ereport(ERROR, (errcode(ERRCODE_INSUFFICIENT_RESOURCES),
 						errmsg("ConnParams arrays bound check failed")));
 	}
@@ -227,7 +227,7 @@ CheckConninfo(const char *conninfo, const char **whitelist,
  */
 void
 GetConnParams(ConnectionHashKey *key, char ***keywords, char ***values,
-			  MemoryContext context)
+			  Index *nonGlobalParamStart, MemoryContext context)
 {
 	/* make space for the port as a string: sign, 10 digits, NUL */
 	char *nodePortString = MemoryContextAlloc(context, 12 * sizeof(char *));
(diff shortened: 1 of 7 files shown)


src/backend/distributed/connection/connection_management.c
```diff
@@ -694,8 +694,27 @@ StartConnectionEstablishment(ConnectionHashKey *key)
 	entry = hash_search(ConnParamsHash, key, HASH_ENTER, &found);
 	if (!found || !entry->isValid)
 	{
-		/* if they're not found, compute them from GUC, runtime, etc. */
-		GetConnParams(key, &entry->keywords, &entry->values, ConnectionContext);
+		if (found && !entry->isValid)
+		{
+			char **keyword = &entry->keywords[entry->nonGlobalParamStart];
```
onderkalaci: But if we start from `nonGlobalParamStart`, wouldn't we hit the issue that Marco raised here: https://github.com/citusdata/citus/pull/2611/files#r259832277

On any invalidation, we'd re-allocate items that are `nonGlobalParams` in the `ConnectionContext`.

I had an example case to repro in that thread, which seem to have the same memory leak affect on this branch:
```SQL
-- session 1
SELECT count(*) FROM table;
\watch 0.001
```
and:

```SQL
-- session 2
alter system set citus.node_conninfo TO 'connect_timeout=20';
SELECT pg_reload_conf();
\watch 0.001 
```

(Well, I've created a script with different values for connect timeout and used it, so not sure about setting the same value always works as I expect ):
```
SELECT pg_reload_conf();
alter system set citus.node_conninfo TO 'connect_timeout= 6 ';
SELECT pg_reload_conf();
alter system set citus.node_conninfo TO 'connect_timeout= 7 ';
SELECT pg_reload_conf();
alter system set citus.node_conninfo TO 'connect_timeout= 8 ';
SELECT pg_reload_conf();
alter system set citus.node_conninfo TO 'connect_timeout= 9 ';
SELECT pg_reload_conf();
....
```

jasonmp85: I don't quite understand the issue. It's odd that there's still a memory leak; I _didn't_ change my code to allocate the global settings in the `ConnectionContext`; they're still managed by the configuration methods, and `malloc`ed/`free`d there, no?

jasonmp85: Oh dur, I see the problem now; the arrays aren't `pfree`d. I'll see whether that fixes your memory issue.

jasonmp85: Ok, so @onderkalaci… I think I've removed any leaks in _this_ code, but the _side effects_ of this code may have some leaks? Like, shouldn't we somehow tear down the existing cached connections if their parameters change?

I'm starting to lean towards "maybe this GUC should require a restart, or at least a new backend"?

onderkalaci: > Like, shouldn't we somehow tear down the existing cached connections if their parameters change?

I think yes, that'd be ideal. But given those are too rare, the assumption in the code looks like `we can live with that`.

(But the memory leak might not be too rare because any SIGHUP could trigger that)

> I'm starting to lean towards "maybe this GUC should require a restart, or at least a new backend"?

I think that used to be case, but we recently had to change it to enable SSL by default. See the related change [here](https://github.com/citusdata/citus/commit/4af40eee76d3e17260e0a4af8febb18039d6c9eb).

As far as I can see, that's only required for upgrades from non-ssl-enabled clusters to later versions of Citus where ssl-enabled.

Is there a way to enable SIGHUP only during the upgrades?

jasonmp85:
> I think that used to be case, but we recently had to change it to enable SSL by default. See the related change here.

Oh. My. God. I was feeling _so_ responsible for this huge oversight in this code and I didn't actually put it there! I feel a lot better now, I was wondering how in the world I overlooked this: turns out someone violated my assumptions out from under me!

I'm now even more convinced we should have handled the SSL stuff in packaging and not in core.

> I can verify that there is no memory leak on this branch related to connection params. It might be useful to move the test in #2611 (or something you'd find useful).

Great; thanks! I'll move the test.

> In order to simplify the code, I proposed to copy even ConnParams to the connection context and not have nonGlobalParamStart in #2611.

The more I think about it, the more I think there should be some sort of child context (maybe `ConnParamsContext`) that we reset rather than calling `free` or `pfree` directly. I'll try to simplify by doing that.

> since we always invalidate all entries in the ConnParamsHash , I used a single flag and set it whenever necessary instead of looping all the entries (and removed entry->isValid). But, I'm fine with this change as well, just wanted to raise in case you've not realized that.

I wrote it this way so if we want in the future we can only invalidate parameters that have changed (e.g. based on a single row changing in the parameter table)

> I'd like to have a final look after you consider my comments.

Where are you working from right now, time-zone-wise?

> And a final question: Do we need any changes on the enterprise repo? As far as I can see no, this should work fine, we may need to handle some merge conflicts only.

I'll merge and run the tests, but I don't think there are any changes needed. `pfree`ing starting at the global param start point should take care of `pfree`ing the auth parameters as well, whose lifetime rules are identical to the per-host ones (the only major difference is around the security of the table they come from).

src/backend/distributed/connection/connection_configuration.c
```diff
@@ -302,6 +321,7 @@ GetConnParams(ConnectionHashKey *key, char ***keywords, char ***values,
 
 	*keywords = connKeywords;
 	*values = connValues;
+	*nonGlobalParamStart = ConnParams.size;
```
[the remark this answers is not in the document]
jasonmp85: Changed to `runtimeParamStart`, since this code consistently uses `runtime` and `global` to talk about these things. Also… moved it.

src/backend/distributed/connection/connection_configuration.c
```diff
@@ -86,6 +85,19 @@ ResetConnParams()
 		ConnParams.keywords[paramIdx] = ConnParams.values[paramIdx] = NULL;
 	}
 
+	if (ConnParamsHash != NULL)
```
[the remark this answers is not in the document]
jasonmp85: Done.

src/backend/distributed/connection/connection_configuration.c
```diff
@@ -227,7 +238,7 @@ CheckConninfo(const char *conninfo, const char **whitelist,
  */
 void
 GetConnParams(ConnectionHashKey *key, char ***keywords, char ***values,
-			  MemoryContext context)
+			  Index *nonGlobalParamStart, MemoryContext context)
 {
 	/* make space for the port as a string: sign, 10 digits, NUL */
 	char *nodePortString = MemoryContextAlloc(context, 12 * sizeof(char *));
```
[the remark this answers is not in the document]
jasonmp85: Yeah, I also don't like it, but our "assign at declaration and don't interleave declarations and code" style rule meant I wrote it this way (note the big assignments up top are declarations, too).

Gonna leave it here, since I think the alternative would have too much unrelated code churn.

src/backend/distributed/connection/connection_management.c
```diff
@@ -694,8 +694,30 @@ StartConnectionEstablishment(ConnectionHashKey *key)
...
+			char **value = &entry->values[entry->nonGlobalParamStart];
```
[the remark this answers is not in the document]
jasonmp85: Agreed; done. I think it sort of mirrors e.g. `PQconninfoFree` or similar methods, so I extracted it out into something named like that.

jasonmp85:
Ok, feedback all addressed.

src/backend/distributed/connection/connection_management.c
```diff
@@ -91,6 +92,26 @@ InitializeConnectionManagement(void)
 }
 
 
+/*
+ * InvalidateConnParamsHashEntries sets every hash entry's isValid flag to false.
+ */
+void
+InvalidateConnParamsHashEntries()
```
marcocitus: I think some compilers complain when there is no `(void)` (though not sure which)

src/backend/distributed/connection/connection_management.c
```diff
@@ -694,8 +715,15 @@ StartConnectionEstablishment(ConnectionHashKey *key)
 	entry = hash_search(ConnParamsHash, key, HASH_ENTER, &found);
 	if (!found || !entry->isValid)
 	{
-		/* if they're not found, compute them from GUC, runtime, etc. */
-		GetConnParams(key, &entry->keywords, &entry->values, ConnectionContext);
+		/* avoid leaking memory in the keys and values arrays */
+		if (found && !entry->isValid)
+		{
+			FreeConnParamsHashEntryFields(entry);
```
marcocitus: This probably shouldn't hold up the PR, but I kind of wonder what happens if there is an out of memory error thrown in `GetConnParams`. We'll re-enter this code, and then most likely segfault by trying to free twice.

onderkalaci: >  but I kind of wonder what happens if there is an out of memory error thrown in GetConnParams.

Just to clarify: If we get out of memory error, wouldn't the execution end (or broadly transaction fail)? How could we re-enter this code?

marcocitus: on the next transaction?

src/backend/distributed/connection/connection_management.c
jasonmp85: Hm… I know that's true of declarations, but I thought it was OK to put in definitions. I'll fix it.

src/backend/distributed/connection/connection_management.c
jasonmp85: I attempted a mitigation for this but am worried about edge cases. Giving up and I think we need to resurrect my `pg_fiu` extension to start injecting OOM errors into our codebase to shake this stuff out.

jasonmp85:
Merging after tests, then I'll update Enterprise, too.

onderkalaci:
shall we backport this to 8.1? @marcocitus and @jasonmp85 what do you think? The use-after-free part of this fix might be relevant, though I'm not concerned about the memory leak part of the fix since it is too rare. Overall, I'd be in favor of backporting it.

PR #2638, opened 2019-03-22
Ignore test_times.log

pykello:
Running tests locally generates `test_times.log` which shouldn't be committed, but is easy to miss when there are a lot of changes.

### DIFF
--- a/src/test/regress/.gitignore
+++ b/src/test/regress/.gitignore
@@ -10,5 +10,8 @@
 /regression.diffs
 /regression.out
 
+# Regression test timing
+/test_times.log
+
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -26,6 +26,7 @@ win32ver.rc
 lib*dll.def
 lib*.pc
 *.bc
+src/test/regress/test_times.log
```
marcocitus: Maybe put it in https://github.com/citusdata/citus/blob/master/src/test/regress/.gitignore

pykello: done

PR #2639, opened 2019-03-22
Warn we are not propagating GRANT/REVOKE to workers

### LINKED ISSUE #2637: Citus Community edition should warn that it is not propagating GRANT to workers
Based on #2129, it seems that we are not going to propagate `GRANT` to workers, but we also don't print any kind of warning messages, which makes the situation a bit confusing.

pykello:
Fixes #2637 

DESCRIPTION: Warn we are not propagating GRANT/REVOKE to workers in community edition.

### DIFF
--- a/src/backend/distributed/commands/grant.c
+++ b/src/backend/distributed/commands/grant.c
@@ -8,12 +8,67 @@
  *-------------------------------------------------------------------------
  */
 
+#include "postgres.h"
+
+#include "catalog/namespace.h"
 #include "distributed/commands.h"
+#include "distributed/metadata_cache.h"
 
 
 /* placeholder for PlanGrantStmt */
 List *
 PlanGrantStmt(GrantStmt *grantStmt)
 {
+	bool showPropagationWarning = false;
+
+
+	if (grantStmt->targtype == ACL_TARGET_ALL_IN_SCHEMA)
+	{
+		showPropagationWarning = true;
+	}
+	else if (grantStmt->targtype == ACL_TARGET_OBJECT)
+	{
+		switch (grantStmt->objtype)
+		{
+			case OBJECT_SCHEMA:
+			case OBJECT_DATABASE:
+			{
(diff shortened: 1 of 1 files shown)


src/backend/distributed/commands/grant.c
```diff
@@ -8,12 +8,82 @@
...
+#endif
+				{
+					ListCell *rangeVarCell = NULL;
+
+					foreach(rangeVarCell, grantStmt->objects)
+					{
+						RangeVar *rangeVar = (RangeVar *) lfirst(rangeVarCell);
+
+						Oid relationId = RangeVarGetRelid(rangeVar, NoLock, false);
+						if (OidIsValid(relationId) && IsDistributedTable(relationId))
```
marcocitus: can you add comments to the parts that do propagate (need to be removed) in enterprise? That might make future merge conflicts easier.

metdos:
Do we want this in 8.3?

metdos:
@thanodnl do we still need this PR? What do you think?

mkurz:
Please merge this, I when I started to use Citus Community edition it took a while to figure out what's going on. If there is a warning this would definitely improve things.

thanodnl:
I think this is a question **we should discuss with product**.

For enterprise these messages will not make sense as we actually do support propagation of grants and revokes for users that are distributed.

In community we do provide a warning when creating a role. That message might require some rewording to make it clear users will not be managed by citus. It might be nice to add warnings for anything that touches users on distributed objects to inform the user this is not supported in the community editions as per @mkurz comments. If you overlook or misunderstand the warning when creating users it can take a while for people to figure this out on their own.

quickly looking at the age of the PR and other comments I think we can't use this as is anymore given the changed world in enterprise.

onderkalaci:
This is not relevant as we now propagate GRANT/REVOKE commands via https://github.com/citusdata/citus/pull/6008

PR #2640, opened 2019-03-23
Address unsafe coercion removal in pruning logic

### LINKED ISSUE #2605: Shard pruning may prune wrong shards when explicit or implicit coercion is on the distribution key
This seems to be happening since we're passing a `numeric` type `hashint8` function on `PruneShards()`

Steps to repro, note that shardids change on each execution:

```SQL

CREATE TABLE articles_hash (
	id bigint NOT NULL,
	author_id bigint NOT NULL,
	title varchar(20) NOT NULL,
	word_count integer
);
SELECT create_distributed_table('articles_hash', 'author_id');
SET client_min_messages TO DEBUG2;
SET citus.log_remote_commands TO ON; 
SELECT *
FROM articles_hash
WHERE author_id = 1.0::numeric;
DEBUG:  Creating router plan
DETAIL:  distribution column value: 1.0
LOG:  issuing SELECT id, author_id, title, word_count FROM public.articles_hash_102399 articles_hash WHERE ((author_id)::numeric OPERATOR(pg_catalog.=) 1.0)
 id | author_id | title | word_count 
----+-----------+-------+------------
(0 rows)

SELECT *
FROM articles_hash
WHERE author_id = 1.0::numeric;
DEBUG:  Creating router plan
DETAIL:  distribution column value: 1.0
LOG:  issuing SELECT id, author_id, title, word_count FROM public.articles_hash_102398 articles_hash WHERE ((author_id)::numeric OPERATOR(pg_catalog.=) 1.0)
 id | author_id | title | word_count 
----+-----------+-------+------------
(0 rows)

```

Implicit coercion, again :

```SQL
insert into articles_hash VALUES (1,1,1,1);
SELECT *
FROM articles_hash
WHERE author_id = 1;
DEBUG:  Creating router plan
DETAIL:  distribution column value: 1
LOG:  issuing SELECT id, author_id, title, word_count FROM public.articles_hash_102433 articles_hash WHERE (author_id OPERATOR(pg_catalog.=) 1)
 id | author_id | title | word_count 
----+-----------+-------+------------
(0 rows)

SELECT *
FROM articles_hash
WHERE author_id = 1.0;
DEBUG:  Creating router plan
DETAIL:  distribution column value: 1.0
LOG:  issuing SELECT id, author_id, title, word_count FROM public.articles_hash_102455 articles_hash WHERE ((author_id)::numeric OPERATOR(pg_catalog.=) 1.0)
 id | author_id | title | word_count 
----+-----------+-------+------------
  1 |         1 | 1     |          1
(1 row)

```

On Postgres, it always returns
```SQL
SELECT *
FROM articles_hash
WHERE author_id = 1;
 id | author_id | title | word_count 
----+-----------+-------+------------
  1 |         1 | 1     |          1
(1 row)

SELECT *
FROM articles_hash
WHERE author_id = 1.0;
 id | author_id | title | word_count 
----+-----------+-------+------------
  1 |         1 | 1     |          1
(1 row)
```

jasonmp85:
The root cause of this bug is what I believe is an assumption that `strip_implicit_coercions` only removes e.g. `RelabelType` nodes and (mostly) returns a pointer to a binary-compatible type. This is _not_ true; _any_ implicit coercion will be stripped, even e.g. a `FuncExpr`.

The provided repro case had a query attempting to SELECT based on an equality constraint between an `integer` column and `numeric` constant. While one might assume PostgreSQL would cast the `numeric` to the type of the column, that's not what happens. It's _possible_ there is actually an equality operator that can directly compare the two types. If so, PostgreSQL would have used that. In this case, such an operator does not exist, so PostgreSQL looks for candidates that can compare a type both sides of the clause can be (losslessly) converted into. There _does not seem to be any preference for the `Var`'s type_.

So PostgreSQL decided `numeric_eq` was a good pick and wrapped the `integer` in a cast to `numeric`, rather than the other way around. We come in and strip all implicit coercions and then _pass the `Const`'s `Datum` directly to our pruning comparison function without checking types_, essentially handing a binary `numeric` to a function expecting `integer`.

I've addressed this by (a) adding some `Assert`s to ensure we're always calling these functions with types that are at least binary-coercible into the expected argument types, and (b) pruning logic will proactively coerce any restriction values into the partition column type before doing any function dispatches. I believe the issues is resolved.

DESCRIPTION: Fixes a bug related to pruning shards using a coerced value

Fixes #2605

### DIFF
--- a/src/backend/distributed/planner/shard_pruning.c
+++ b/src/backend/distributed/planner/shard_pruning.c
@@ -534,8 +534,7 @@ PrunableExpressionsWalker(Node *node, ClauseWalkerContext *context)
 
 		if (!prune->addedToPruningInstances)
 		{
-			context->pruningInstances = lappend(context->pruningInstances,
-												prune);
+			context->pruningInstances = lappend(context->pruningInstances, prune);
 			prune->addedToPruningInstances = true;
 		}
 
@@ -565,8 +564,8 @@ PrunableExpressionsWalker(Node *node, ClauseWalkerContext *context)
 			 * Found a restriction on the partition column itself. Update the
 			 * current constraint with the new information.
 			 */
-			AddPartitionKeyRestrictionToInstance(context,
-												 opClause, varClause, constantClause);
+			AddPartitionKeyRestrictionToInstance(context, opClause, varClause,
+												 constantClause);
 		}
 		else if (constantClause && varClause &&
 				 varClause->varattno == RESERVED_HASHED_COLUMN_ID)
@@ -597,8 +596,7 @@ PrunableExpressionsWalker(Node *node, ClauseWalkerContext *context)
 		 */
 		if (!prune->addedToPruningInstances)
 		{
-			context->pruningInstances = lappend(context->pruningInstances,
-												prune);
+			context->pruningInstances = lappend(context->pruningInstances, prune);
 			prune->addedToPruningInstances = true;
 		}
 
@@ -740,8 +738,7 @@ AddPartitionKeyRestrictionToInstance(ClauseWalkerContext *context, OpExpr *opCla
 	/* at this point, we'd better be able to pass binary Datums to comparison functions */
 	Assert(IsBinaryCoercible(constantClause->consttype, partitionColumn->vartype));
 
-	btreeInterpretationList =
-		get_op_btree_interpretation(opClause->opno);
+	btreeInterpretationList = get_op_btree_interpretation(opClause->opno);
 	foreach(btreeInterpretationCell, btreeInterpretationList)
 	{
 		OpBtreeInterpretation *btreeInterpretation =
@@ -846,8 +843,7 @@ AddPartitionKeyRestrictionToInstance(ClauseWalkerContext *context, OpExpr *opCla
 
 	if (!matchedOp)
 	{
-		prune->otherRestrictions =
-			lappend(prune->otherRestrictions, opClause);
+		prune->otherRestrictions = lappend(prune->otherRestrictions, opClause);
 	}
 	else
 	{
--- a/src/backend/distributed/planner/shard_pruning.c
+++ b/src/backend/distributed/planner/shard_pruning.c
@@ -174,7 +174,7 @@ static void AddPartitionKeyRestrictionToInstance(ClauseWalkerContext *context,
 												 OpExpr *opClause, Var *varClause,
 												 Const *constantClause);
 static Const * TransformPartitionRestrictionValue(Var *partitionColumn,
-												  Node *restrictionValue);
+												  Const *restrictionValue);
 static void AddSAOPartitionKeyRestrictionToInstance(ClauseWalkerContext *context,
 													ScalarArrayOpExpr *
 													arrayOperatorExpression);
@@ -724,15 +724,19 @@ AddPartitionKeyRestrictionToInstance(ClauseWalkerContext *context, OpExpr *opCla
 	ListCell *btreeInterpretationCell = NULL;
 	bool matchedOp = false;
 
-	/* we want our restriction value in terms of the type of the partition column */
-	constantClause = TransformPartitionRestrictionValue(partitionColumn,
-														(Node *) constantClause);
-	if (constantClause == NULL)
+	/* only have extra work to do if const isn't same type as partition column */
+	if (constantClause->consttype != partitionColumn->vartype)
 	{
-		/* couldn't coerce the value, so we note this as a restriction we don't grok */
-		prune->otherRestrictions = lappend(prune->otherRestrictions, opClause);
+		/* we want our restriction value in terms of the type of the partition column */
+		constantClause = TransformPartitionRestrictionValue(partitionColumn,
+															constantClause);
+		if (constantClause == NULL)
+		{
+			/* couldn't coerce value, so we note this as a restriction we don't grok */
+			prune->otherRestrictions = lappend(prune->otherRestrictions, opClause);
 
-		return;
+			return;
+		}
 	}
 
 	/* at this point, we'd better be able to pass binary Datums to comparison functions */
@@ -860,10 +864,10 @@ AddPartitionKeyRestrictionToInstance(ClauseWalkerContext *context, OpExpr *opCla
  * we will simply fail to prune partitions based on this clause.
  */
 static Const *
-TransformPartitionRestrictionValue(Var *partitionColumn, Node *restrictionValue)
+TransformPartitionRestrictionValue(Var *partitionColumn, Const *restrictionValue)
 {
-	Node *transformedValue = coerce_to_target_type(NULL, restrictionValue,
-												   exprType(restrictionValue),
+	Node *transformedValue = coerce_to_target_type(NULL, (Node *) restrictionValue,
+												   restrictionValue->consttype,
 												   partitionColumn->vartype,
 												   partitionColumn->vartypmod,
 												   COERCION_ASSIGNMENT,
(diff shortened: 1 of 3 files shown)


src/backend/distributed/planner/shard_pruning.c
```diff
@@ -825,6 +852,45 @@ AddPartitionKeyRestrictionToInstance(ClauseWalkerContext *context, OpExpr *opCla
...
+	/* if NULL, no implicit coercion is possible between the types */
+	if (transformedValue == NULL)
+	{
+		return NULL;
+	}
+
+	/* if still not a constant, evaluate coercion */
+	if (!IsA(transformedValue, Const))
+	{
+		transformedValue = (Node *) expression_planner((Expr *) transformedValue);
```
onderkalaci: Nice, I haven't seen this function before! I'll probably update [here](https://github.com/citusdata/citus/blob/master/src/backend/distributed/planner/fast_path_router_planner.c#L85-L92) to use it instead of `eval_const_expressions`.

src/backend/distributed/planner/shard_pruning.c
```diff
@@ -803,6 +818,19 @@ AddPartitionKeyRestrictionToInstance(ClauseWalkerContext *context, OpExpr *opCla
 
 			case ROWCOMPARE_NE:
 			{
+				/*
```
onderkalaci: I couldn't follow this change. Is this related to the fix?

src/backend/distributed/planner/shard_pruning.c
```diff
@@ -825,6 +852,45 @@ AddPartitionKeyRestrictionToInstance(ClauseWalkerContext *context, OpExpr *opCla
...
+ * than the Const; to deal with this, we strip the coercions from both and
+ * manually coerce the Const into the type of our partition column. It is
+ * conceivable that in some instances, this may not be possible; in those cases
+ * we will simply fail to prune partitions based on this clause.
+ */
+static Const *
+TransformPartitionRestrictionValue(Var *partitionColumn, Node *restrictionValue)
+{
+	Node *transformedValue = coerce_to_target_type(NULL, restrictionValue,
+												   exprType(restrictionValue),
```
onderkalaci: I'm wondering whether should we have a a short circuit for `if exprType(restrictionValue) == partitionColumn->vartype`? That's going to be the vast majority of the time and I'd like to avoid any extra overhead.

I can see that `coerce_to_target_type ` does some very little things if that's already the case. But, given that this function will be called too many times on busy servers, I'd have a slight preference to have the short-circuit (or maybe run a quick select-only benchmark to see if it has any noticeable overhead or not.).

(And, that'd probably easier to follow: `If the types are the same return else transform`)

jasonmp85:
> I've only a minor concern on the overhead that might be added by this change. We've had hard times for finding lots of optimizations during this release and I don't want to create any new overheads on the fast-path router queries 😄

While true, I wanted to rely on the PostgreSQL logic for when to short-circuit, or to make determinations about what was safe. This bug is a bit worrisome from a correctness standpoint. I'll look into it, though.

src/backend/distributed/planner/shard_pruning.c
```diff
@@ -715,15 +717,28 @@ AddNewConjuction(ClauseWalkerContext *context, OpExpr *op)
...
 	List *btreeInterpretationList = NULL;
 	ListCell *btreeInterpretationCell = NULL;
 	bool matchedOp = false;
 
-	btreeInterpretationList =
-		get_op_btree_interpretation(opClause->opno);
+	/* we want our restriction value in terms of the type of the partition column */
+	constantClause = TransformPartitionRestrictionValue(partitionColumn,
+														(Node *) constantClause);
+	if (constantClause == NULL)
```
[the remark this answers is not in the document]
jasonmp85: Yeah, the case I'm worried about is this. It's _theoretically_ possible to have a…

  * Partition column `foo` of type `Foo`
  * Type `Bar` with a `CAST` function defined from `Foo` to `Bar`, but not the other way around
  * Query like `SELECT * FROM table WHERE foo = 'baz'::Bar`

In this case, PostgreSQL will wrap the `foo` `Var` in a cast to `Bar` and rely on `Bar`'s equality operator. We'll step in and go "waitasec, we really want a `Foo` here" and try to bring the `const` back to `Foo`, but the `CAST` has no inverse function. I poked through a lot of PostgreSQL's built-in casts and operator classes/families/access methods, but that proved to be enough of a goose chase I just gave up and implemented the logic to handle it (I'm pretty sure there are cases that might act like this).

src/test/regress/expected/multi_prune_shard_list.out
```diff
@@ -203,3 +203,27 @@ SELECT print_sorted_shard_intervals('pruning_range');
...
+ 
+(1 row)
+
+-- PostgreSQL will cast id to numeric rather than 1.0 to bigint...
+-- We used to blindly pass the RHS' Datum to our comparison func.,
+-- resulting in inaccurate pruning. An Assert now verifies type-
+-- compatibility; the following would crash the server in an Assert-
+-- before the underlying issue was addressed. It looks like a boring
+-- test now, but if the old behavior is restored, it should crash again.
+SELECT * FROM coerce_hash WHERE id = 1.0::numeric;
```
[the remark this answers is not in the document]
jasonmp85: I just hesitated to add a "do a bunch of `SELECTs`" test because (a) it's not guaranteed to fail in the case of a regression, and (b) I don't really like adding tests that do things multiple times (our tests take long enough as it is). But I figure it won't actually take that long. I'll add it.

src/backend/distributed/planner/shard_pruning.c
```diff
@@ -825,6 +852,45 @@ AddPartitionKeyRestrictionToInstance(ClauseWalkerContext *context, OpExpr *opCla
...
 
+/*
+ * Sometimes PostgreSQL chooses to try to wrap our Var in a coercion rather
+ * than the Const; to deal with this, we strip the coercions from both and
+ * manually coerce the Const into the type of our partition column. It is
+ * conceivable that in some instances, this may not be possible; in those cases
+ * we will simply fail to prune partitions based on this clause.
+ */
+static Const *
+TransformPartitionRestrictionValue(Var *partitionColumn, Node *restrictionValue)
```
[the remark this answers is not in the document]
jasonmp85: I actually _did_ have that but enough of PostgreSQL's functions require a `Node *` pointer that I changed it. I'll change it back.

src/backend/distributed/planner/shard_pruning.c
jasonmp85: Yeah it seemed like a _slightly_ higher-level abstraction.

src/backend/distributed/planner/shard_pruning.c
jasonmp85: No, not directly; just while reading the logic I saw that `ROWCOMPARE_NE` was some sort of magic value returned because PostgreSQL _consider[s] a `<>` operator to be a "member" of an opfamily if its negator is an equality operator of the opfamily_ and returns _ROWCOMPARE_NE […] for this case_.

Since we're iterating over a list of these things, either (a) the list should contain no `ROWCOMPARE_NE` elements, or (b) it should contain _only_ `ROWCOMPARE_NE` elements. But _nothing_ in our code actually reflected this behavior of the underlying code, and I was worried enough about further correctness issues in this pruning stuff that I threw in some logic to `Assert` or short-circuit in those cases (we don't understand `ROWCOMPARE_NE` at anyhow).

src/backend/distributed/planner/shard_pruning.c
jasonmp85: Sure, this seems reasonable. Also, I think I probably am pushed toward blindly calling a function and assigning it during declaration because of the weird style requirement around that.

jasonmp85: OK, I actually put the check _before_ the call to `TransformPartitionRestrictionValue`, as that way we can actually save the function call overhead, too. I think that this will be the _overwhelmingly_ common case, and as such branch prediction should _probably_ make all of this moot, but just in case.

PR #2647, opened 2019-04-02
Ensure that stack resizing logic works expected

onderkalaci:
This commit has two goals:

(a) Ensure to access both edges of the allocated stack since it might depend on architecture
(b) Ensure that any compiler optimizations to prevent the function optimized away.

Stack size after the patch:
```
 sudo grep -A 1 stack /proc/2119/smaps
7ffe305a6000-7ffe307a9000 rw-p 00000000 00:00 0                          [stack]
Size:               2060 kB
```
Stack size before the patch:
```
 sudo grep -A 1 stack /proc/3610/smaps
7fff09957000-7fff09978000 rw-p 00000000 00:00 0                          [stack]
Size:                132 kB
```
DESCRIPTION: Fixes a bug that prevents stack size to be adjusted

### DIFF
--- a/src/backend/distributed/shared_library_init.c
+++ b/src/backend/distributed/shared_library_init.c
@@ -52,6 +52,7 @@
 #include "distributed/worker_manager.h"
 #include "distributed/worker_protocol.h"
 #include "distributed/worker_shard_visibility.h"
+#include "port/atomics.h"
 #include "postmaster/postmaster.h"
 #include "optimizer/planner.h"
 #include "optimizer/paths.h"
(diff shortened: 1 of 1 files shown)


onderkalaci:
Documentation for [pg_memory_barrier()](https://github.com/postgres/postgres/blob/master/src/backend/storage/lmgr/README.barrier) if anyone interested.

onderkalaci:
> Tested with my HLL crash reproducer and confirmed this fixes the issue.

Just for future reference: @byucesoy was able to crash Postgres on the Citus master branch with -O2 optimizations enabled. And, on this branch, the crash was not reproduced.

PR #2650, opened 2019-04-03
Address constraint naming inconsistencies

### LINKED ISSUE #2484: Non-index CHECK constraints from CREATE TABLE cannot be dropped from coordinator
A customer ran into an issue while trying to drop a constraint.

# Repro Steps

```sql
CREATE TABLE test_constraint (id int CHECK (id>0));
-- CREATE TABLE
SELECT create_distributed_table('test_constraint', 'id');
-- create_distributed_table
-- --------------------------
-- 
-- (1 row)

ALTER TABLE test_constraint DROP CONSTRAINT test_constraint_id_check;
-- ERROR:  constraint "test_constraint_id_check_102168" of relation "test_constraint_102168" does not exist
-- CONTEXT:  while executing command on ec2-23-22-154-48.compute-1.amazonaws.com:5432
```

# Notes

On the worker, here is the constraint that it created 

```sql
-- \d+ test_constraint_102169
--                           Table "public.test_constraint_102169"
--  Column |  Type   | Collation | Nullable | Default | Storage | Stats target | Description
-- --------+---------+-----------+----------+---------+---------+--------------+-------------
--  id     | integer |           |          |         | plain   |              |
-- Check constraints:
--     "test_constraint_id_check" CHECK (id > 0)
```

As you can see, the constraint doesn't include the `shardid`…

# Workaround

I was able to drop the constraint using the following:

```sql
SELECT run_command_on_shards('test_constraint', 'ALTER TABLE %s DROP CONSTRAINT test_constraint_id_check');
```

jasonmp85:
Fixes #2484

### DIFF
--- a/src/test/regress/sql/multi_create_table_constraints.sql
+++ b/src/test/regress/sql/multi_create_table_constraints.sql
@@ -230,6 +230,9 @@ SELECT "Column", "Type", "Definition" FROM index_attrs WHERE
 SELECT "Constraint", "Definition" FROM table_checks WHERE relid='public.check_example_365056'::regclass;
 \c - - - :master_port
 
+-- try dropping a CHECK constraint that was created without explicit name
+ALTER TABLE check_example DROP CONSTRAINT check_example_other_col_check;
+
 -- drop unnecessary tables
 DROP TABLE pk_on_non_part_col, uq_on_non_part_col CASCADE;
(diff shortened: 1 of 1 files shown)


jasonmp85:
Oops; was using `hub` CLI and meant to create this as a draft.

PR #2651, opened 2019-04-03
Address constraint naming issue

### LINKED ISSUE #2484: Non-index CHECK constraints from CREATE TABLE cannot be dropped from coordinator

jasonmp85:
DESCRIPTION: Make DROP/VALIDATE CONSTRAINT tolerant of ambiguous shard extension

I'll update this task tomorrow morning with more explanation. The comment in my SQL test file should give some basis for a review. @pykello, I've assigned to you. This change addresses the bug in a backportable way, but I think we should probably do something to clean up the ambiguous shard-extension issue (which silently changed at some point in the recent past).

I've got a bunch of GitHub windows open with various commits pointing to where this stuff was introduced; I've got to assume this was done unknowingly since it's quite undesirable.

Fixes #2484

# Tasks

  - [x] Write failing repro test
  - [x] Determine how implicitly-named `CHECK` constraints are deparsed
  - [x] Determine how `ALTER TABLE ... DROP CONSTRAINT ...` statements are deparsed
  - [x] Change `ALTER TABLE` deparse to be more tolerant of missing `shardid`
  - [x] Investigate similar occurrences of this problem
  - [x] Consider additional changes or creating additional issues

### DIFF
--- a/src/backend/distributed/relay/relay_event_utility.c
+++ b/src/backend/distributed/relay/relay_event_utility.c
@@ -51,7 +51,6 @@
 #include "utils/relcache.h"
 
 /* Local functions forward declarations */
-static void AppendShardIdToConstraintName(AlterTableCmd *command, uint64 shardId);
 static bool UpdateWholeRowColumnReferencesWalker(Node *node, uint64 *shardId);
 
 /* exports for SQL callable functions */
@@ -85,6 +84,7 @@ RelayEventExtendNames(Node *parseTree, char *schemaName, uint64 shardId)
 			 */
 
 			AlterTableStmt *alterTableStmt = (AlterTableStmt *) parseTree;
+			Oid relationId = InvalidOid;
 			char **relationName = &(alterTableStmt->relation->relname);
 			char **relationSchemaName = &(alterTableStmt->relation->schemaname);
 
@@ -104,6 +104,7 @@ RelayEventExtendNames(Node *parseTree, char *schemaName, uint64 shardId)
 				if (command->subtype == AT_AddConstraint)
 				{
 					Constraint *constraint = (Constraint *) command->def;
+					char **constraintName = &(constraint->conname);
 
 					if (constraint->contype == CONSTR_PRIMARY && constraint->indexname)
 					{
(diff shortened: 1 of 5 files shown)


pykello:
Why are we doing this instead of making `CREATE TABLE` and `ADD CONSTRAINT` assign consistent names? Is it because we want to backport?

jasonmp85:
> Why are we doing this instead of making `CREATE TABLE` and `ADD CONSTRAINT` assign consistent names? Is it because we want to backport?

Read the pull request description! Quoth…

> This change addresses the bug in a backportable way, but I think we should probably do something to clean up the ambiguous shard-extension issue

jasonmp85:
(ambiguous in the sense that there is presently no unambiguous mapping from _constraint name on coordinator_ to _constraint name on worker_)

PR #2654, opened 2019-04-04
Search all outer node levels for lateral join params.

### LINKED ISSUE #2641: Lateral-join pushdown fails when the join clause contains a column from two levels up
This was a known issue, wanted to open an issue to track. We're only considering `Param`s
that are one level above on `multi_relation_restriction_hook`:
```C
	/* see comments on GetVarFromAssignedParam() */
	if (relationRestriction->parentPlannerInfo)
	{
		relationRestriction->parentPlannerParamList =
			CopyPlanParamList(root->parent_root->plan_params);
	}
```

Instead, we should probably consider the `parent_root`s with a `while` loop and go up to the
upper levels and gather all parameters. Otherwise, we're only considering the `Param` that are avaliable immediately one level up.

Minimals steps to reproduce:

```SQL
-- works fine
SELECT
  DISTINCT event_type
FROM (
  SELECT
    user_id,
    event_type
  FROM events_table
) a,
LATERAL (
    SELECT
      user_id 
    FROM events_table WHERE
      events_table.user_id = a.user_id
) b   
LIMIT 10;
 event_type 
------------
          1
          5
          4
          2
          6
          0
          3
(7 rows)
```

```SQL
-- the same query, but the LATERAL part is wrapped into 
-- an extra subquery
SELECT
    DISTINCT event_type
FROM (
  SELECT
    user_id,
    event_type
  FROM events_table
) a,
LATERAL (
  SELECT
    user_id, random() -- prevent postgresql planner to pull-up the subquery
  FROM (
    SELECT
      user_id , random() -- prevent postgresql planner to pull-up the subquery
    FROM events_table WHERE
      events_table.user_id = a.user_id
  ) events
) b   
LIMIT 10;
```

The workaround is to add join clauses on the distribution key on the upper :

```SQL
SELECT
    DISTINCT event_type
FROM (
  SELECT
    user_id,
    event_type
  FROM events_table
) a,
LATERAL (
  SELECT
    user_id, random() -- prevent postgresql planner to pull-up the subquery
  FROM (
    SELECT
      user_id , random() -- prevent postgresql planner to pull-up the subquery
    FROM events_table WHERE
      events_table.user_id = a.user_id
  ) events
) b 
-- THIS IS NECESSARY
WHERE a.user_id = b.user_id   
-- ADD JOIN BETWEEN SUBQUERIES
LIMIT 10;
 event_type 
------------
          1
          5
          4
          2
          6
          0
          3
(7 rows)
```

pykello:
A planner node can have parameters that reference variables in outer level planner nodes. This can happen when doing a lateral join. Previously we only searched the parent node to resolve these parameters, but this parameters can be references to any of the outer levels. This patch fixes the code to search all of the outer levels.

DESCRIPTION: Search all outer node levels for lateral join params.

fixes #2641

### DIFF
--- a/src/backend/distributed/planner/distributed_planner.c
+++ b/src/backend/distributed/planner/distributed_planner.c
@@ -79,6 +79,7 @@ static void CheckNodeIsDumpable(Node *node);
 static Node * CheckNodeCopyAndSerialization(Node *node);
 static void AdjustReadIntermediateResultCost(RangeTblEntry *rangeTableEntry,
 											 RelOptInfo *relOptInfo);
+static List * OuterPlanParamsList(PlannerInfo *root);
 static List * CopyPlanParamList(List *originalPlanParamList);
 static PlannerRestrictionContext * CreateAndPushPlannerRestrictionContext(void);
 static PlannerRestrictionContext * CurrentPlannerRestrictionContext(void);
@@ -1299,15 +1300,10 @@ multi_relation_restriction_hook(PlannerInfo *root, RelOptInfo *relOptInfo, Index
 	relationRestriction->relOptInfo = relOptInfo;
 	relationRestriction->distributedRelation = distributedTable;
 	relationRestriction->plannerInfo = root;
-	relationRestriction->parentPlannerInfo = root->parent_root;
 	relationRestriction->prunedShardIntervalList = NIL;
 
 	/* see comments on GetVarFromAssignedParam() */
-	if (relationRestriction->parentPlannerInfo)
-	{
-		relationRestriction->parentPlannerParamList =
-			CopyPlanParamList(root->parent_root->plan_params);
-	}
+	relationRestriction->outerPlanParamsList = OuterPlanParamsList(root);
 
 	relationRestrictionContext = plannerRestrictionContext->relationRestrictionContext;
 	relationRestrictionContext->hasDistributedRelation |= distributedTable;
@@ -1459,6 +1455,36 @@ AdjustReadIntermediateResultCost(RangeTblEntry *rangeTableEntry, RelOptInfo *rel
 	path = (Path *) linitial(pathList);
 	path->rows = rowCountEstimate;
 	path->total_cost = rowCountEstimate * rowCost + ioCost;
+}
+
+
+/*
+ * OuterPlanParamsList creates a list of RootPlanParams for outer nodes of the
+ * given root. The first item in the list corresponds to parent_root, and the
+ * last item corresponds to the outer most node.
+ */
+static List *
+OuterPlanParamsList(PlannerInfo *root)
+{
+	List *planParamsList = NIL;
+	PlannerInfo *outerNodeRoot = NULL;
+
+	for (outerNodeRoot = root->parent_root; outerNodeRoot != NULL;
+		 outerNodeRoot = outerNodeRoot->parent_root)
+	{
+		RootPlanParams *rootPlanParams = palloc0(sizeof(RootPlanParams));
+		rootPlanParams->root = outerNodeRoot;
+
+		/*
+		 * TODO: In SearchPlannerParamList() we are only interested in Var plan
+		 * params, consider copying just them here.
+		 */
+		rootPlanParams->plan_params = CopyPlanParamList(outerNodeRoot->plan_params);
+
+		planParamsList = lappend(planParamsList, rootPlanParams);
+	}
+
+	return planParamsList;
 }
 
 
--- a/src/backend/distributed/planner/relation_restriction_equivalence.c
+++ b/src/backend/distributed/planner/relation_restriction_equivalence.c
@@ -98,8 +98,9 @@ static void AddRteRelationToAttributeEquivalenceClass(AttributeEquivalenceClass 
 													  attrEquivalenceClass,
 													  RangeTblEntry *rangeTableEntry,
 													  Var *varToBeAdded);
-static Var * GetVarFromAssignedParam(List *parentPlannerParamList,
-									 Param *plannerParam);
+static Var * GetVarFromAssignedParam(List *outerPlanParamsList, Param *plannerParam,
+									 PlannerInfo **rootContainingVar);
+static Var * SearchPlannerParamList(List *plannerParamList, Param *plannerParam);
 static List * GenerateAttributeEquivalencesForJoinRestrictions(JoinRestrictionContext
 															   *joinRestrictionContext);
 static bool AttributeClassContainsAttributeClassMember(AttributeEquivalenceClassMember *
@@ -737,15 +738,15 @@ AttributeEquivalenceClassForEquivalenceClass(EquivalenceClass *plannerEqClass,
 
 		if (IsA(strippedEquivalenceExpr, Param))
 		{
-			List *parentParamList = relationRestriction->parentPlannerParamList;
+			PlannerInfo *outerNodeRoot = NULL;
 			Param *equivalenceParam = (Param *) strippedEquivalenceExpr;
 
-			expressionVar = GetVarFromAssignedParam(parentParamList,
-													equivalenceParam);
+			expressionVar =
+				GetVarFromAssignedParam(relationRestriction->outerPlanParamsList,
+										equivalenceParam, &outerNodeRoot);
 			if (expressionVar)
 			{
-				AddToAttributeEquivalenceClass(&attributeEquivalance,
-											   relationRestriction->parentPlannerInfo,
+				AddToAttributeEquivalenceClass(&attributeEquivalance, outerNodeRoot,
 											   expressionVar);
 			}
 		}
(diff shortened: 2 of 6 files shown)


src/backend/distributed/planner/distributed_planner.c
```diff
@@ -1462,6 +1458,29 @@ AdjustReadIntermediateResultCost(RangeTblEntry *rangeTableEntry, RelOptInfo *rel
...
 
 
+/*
+ * CopyOuterPlanParamLists deep-copies plan_params lists of the outer nodes
+ * of the given root, and returns them in a list of lists of PlannerParamItem.
+ * First item in this list corresponds to the parent node, and the last
+ * item corresponds to the outer most node.
+ */
+static List *
+CopyOuterPlanParamLists(PlannerInfo *root)
```
pykello: Naming PlannerInfo variables as `root` to match PostgreSQL's convention. See the comments for `PlannerInfo` in `relation.h`.

pykello: Also note that keeping a copy of plan_params lists seems necessary, for some reason that I'm not sure of, all `PlannerInfo->planner_params` lists are empty in `relation_restriction_equivalence.c`.

onderkalaci: > Also note that keeping a copy of plan_params lists seems necessary

Just to clarify: Before the patch, we already keep the copies that are only one level up, right? So, we're only keeping more stuff when necessary?

> , for some reason that I'm not sure of, all PlannerInfo->planner_params lists are empty in relation_restriction_equivalence.c

I think that's because we're passing `baseRelOptInfo->subroot` specific per subquery, not the `root` that we see in the relation restriction hook. It looks like we've somewhat related discussion sometime [ago](https://github.com/citusdata/citus/pull/1268/files#r108345915).

src/backend/distributed/planner/relation_restriction_equivalence.c
```diff
@@ -737,15 +736,37 @@ AttributeEquivalenceClassForEquivalenceClass(EquivalenceClass *plannerEqClass,
...
+				}
+
+				outerNodeRoot = outerNodeRoot->parent_root;
+			}
+
 			if (expressionVar)
 			{
-				AddToAttributeEquivalenceClass(&attributeEquivalance,
-											   relationRestriction->parentPlannerInfo,
+				AddToAttributeEquivalenceClass(&attributeEquivalance, outerNodeRoot,
```
onderkalaci: Hmm, this is tricky to follow and expand in the future because we're trying to match the `expressionVar ` and its corresponding `root` via the loop. I think it is OK for this change, but, for any future users of these fields, it'd be pretty hard to follow the dependency and prone to bugs.

What I had in my mind was to have a new struct and pass the params and its corresponding `root` together and use them together. 

```C
typedef struct RootParams
{
   PlannerInfo *root;
  List *plan_params
};

```

What do you think @pykello ?


(the struct name and field names are placeholders, feel free to improve.)
(Also note that we don't need to copy `root` field, only the params as you did.)

src/backend/distributed/planner/relation_restriction_equivalence.c
```diff
@@ -737,15 +736,37 @@ AttributeEquivalenceClassForEquivalenceClass(EquivalenceClass *plannerEqClass,
...
+			ListCell *outerPlanParamListCell = NULL;
 			Param *equivalenceParam = (Param *) strippedEquivalenceExpr;
 
-			expressionVar = GetVarFromAssignedParam(parentParamList,
-													equivalenceParam);
+			/*
+			 * A Param can reference a Var in any of the outer plan nodes. Search
+			 * them until we find the corresponding Var.
+			 */
+			foreach(outerPlanParamListCell, relationRestriction->outerPlannerParamLists)
```
onderkalaci: I'd be in favor of moving this logic to a separate function, especially if you agree with my below comment.

src/backend/distributed/planner/relation_restriction_equivalence.c
pykello: makes sense, done.

src/backend/distributed/planner/distributed_planner.c
pykello: > > Also note that keeping a copy of plan_params lists seems necessary
> 
> Just to clarify: Before the patch, we already keep the copies that are only one level up, right? So, we're only keeping more stuff when necessary?

Yes, before this patch we were keeping one level up, now we keep all the levels. I'm not sure if we can decide at this point if a levels are necessary or not, so just keeping all levels up.

I'm looking to see if we can avoid keeping all levels and only the necessary ones ...

pykello: Today I looked into if it is possible to not copy any `plan_params` lists and make the code simpler. I haven't found a solution which works on all cases yet.

Using `PlannerInfo->simple_rel_array[idx]->subplan_params` was close, but it was failing one of the tests.

If this change isn't urgent, maybe we can keep this on hold for few more days so I can come back and look into it again.

onderkalaci: > Today I looked into if it is possible to not copy any plan_params lists and make the code simpler.
> If this change isn't urgent, maybe we can keep this on hold for few more days so I can come back and look into it again.

Well, I don't mind copying those params :) But, still getting this change to 8.3 should be sufficient (as far as I can see), so feel free to look later

metdos:
What is the status of this PR? Does it need more work or just clean up?

onderkalaci:
> What is the status of this PR? Does it need more work or just clean up?


Mostly clean-up

metdos:
> Mostly clean-up

Then, shall we try to merge it before you guys forget all the context :)

pykello:
> > Mostly clean-up
> 
> Then, shall we try to merge it before you guys forget all the context :)

I'll work on this when rest of team is travelling next week.

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -1299,15 +1300,10 @@ multi_relation_restriction_hook(PlannerInfo *root, RelOptInfo *relOptInfo, Index
 	relationRestriction->relOptInfo = relOptInfo;
 	relationRestriction->distributedRelation = distributedTable;
 	relationRestriction->plannerInfo = root;
-	relationRestriction->parentPlannerInfo = root->parent_root;
 	relationRestriction->prunedShardIntervalList = NIL;
 
 	/* see comments on GetVarFromAssignedParam() */
```
serprex: You renamed `GetVarFromAssignedParam` to `GetVarFromAssignedVar`

pykello:
@onderkalaci I think I won't find time to look into if we can make this simpler, so I think I'll just make this ready for commit. Since you said you were ok with the current approach, can you please take another look?

src/backend/distributed/planner/distributed_planner.c
pykello: Reverted the rename.

src/backend/distributed/planner/distributed_planner.c
onderkalaci: > I'm looking to see if we can avoid keeping all levels and only the necessary ones ...

We could maybe only copy the params when `IsA(originalParamItem->item, Var)` because we only need them. But, I don't think it is worth doing that at the moment, you could maybe consider adding a comment if you want.

But, we typically have very little params, so I'm fine with what we have here. My main motivation to avoid optimizing here is that as the number of params increase, the execution times typically increases a lot. So, saving very little here isn't very tempting to me.

pykello: > I'm looking to see if we can avoid keeping all levels and only the necessary ones ...
> 
> We could maybe only copy the params when IsA(originalParamItem->item, Var) because we only need them. But, I don't think it is worth doing that at the moment, you could maybe consider adding a comment if you want.
> But, we typically have very little params, so I'm fine with what we have here. My main motivation to avoid optimizing here is that as the number of params increase, the execution times typically increases a lot. So, saving very little here isn't very tempting to me.

Added a comment.

metdos:
@pykello I still get the same error for the query in the issue Onder created. Is this expected?

```SQL
SELECT
    DISTINCT event_type
FROM (
  SELECT
    user_id,
    event_type
  FROM events_table
) a,
LATERAL (
  SELECT
    user_id, random() -- prevent postgresql planner to pull-up the subquery
  FROM (
    SELECT
      user_id , random() -- prevent postgresql planner to pull-up the subquery
    FROM events_table WHERE
      events_table.user_id = a.user_id
  ) events
) b   
LIMIT 10;
```

Here is the error message:

```
```

pykello:
> @pykello I still get the same error for the query in the issue Onder created (#2654). Is this expected?
> SELECT
>     DISTINCT event_type
> FROM (
>   SELECT
>     user_id,
>     event_type
>   FROM events_table
> ) a,
> LATERAL (
>   SELECT
>     user_id, random() -- prevent postgresql planner to pull-up the subquery
>   FROM (
>     SELECT
>       user_id , random() -- prevent postgresql planner to pull-up the subquery
>     FROM events_table WHERE
>       events_table.user_id = a.user_id
>   ) events
> ) b   
> LIMIT 10;
> Here is the error message:
> ERROR:  complex joins are only supported when all distributed tables are joined on their distribution columns with equal operator

No, I will look into this tomorrow.

metdos:
> No, I will look into this tomorrow.

Hey @pykello, this looks like a false alarm. In my tests 'user_id' was not distribution key, if I make it distribution key, the query works fine.

PR #2657, opened 2019-04-05
Make sure that the regression tests are resistant to execution order changes

onderkalaci:
Mostly a preparation for #2435 where execution order on the shards might change. So, add missing `ORDER BY`s and be less verbose if the worker node host/port is printed in the error messages, especially when replication factor > 1.

There might be some missing ones, which I'll keep opening PRs as I find more.

### DIFF
--- a/src/test/regress/expected/multi_name_lengths.out
+++ b/src/test/regress/expected/multi_name_lengths.out
@@ -81,14 +81,14 @@ ERROR:  cannot create constraint without a name on a distributed table
 ALTER TABLE name_lengths ADD CHECK (date_col_12345678901234567890123456789012345678901234567890 > '2014-01-01'::date);
 ERROR:  cannot create constraint without a name on a distributed table
 \c - - - :worker_1_port
-SELECT "Column", "Type", "Modifiers" FROM table_desc WHERE relid='public.name_lengths_225002'::regclass;
+SELECT "Column", "Type", "Modifiers" FROM table_desc WHERE relid='public.name_lengths_225002'::regclass ORDER BY 1 DESC, 2 DESC;
                             Column                            |       Type       | Modifiers 
 --------------------------------------------------------------+------------------+-----------
- col1                                                         | integer          | not null
- col2                                                         | integer          | not null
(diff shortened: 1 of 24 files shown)


onderkalaci:
@hanefi I'm planning to squash the commits before merging, leaving it as is if you want to see each individual change separately.

src/test/regress/expected/multi_name_lengths.out
```diff
@@ -81,14 +81,14 @@ ERROR:  cannot create constraint without a name on a distributed table
 ALTER TABLE name_lengths ADD CHECK (date_col_12345678901234567890123456789012345678901234567890 > '2014-01-01'::date);
 ERROR:  cannot create constraint without a name on a distributed table
 \c - - - :worker_1_port
-SELECT "Column", "Type", "Modifiers" FROM table_desc WHERE relid='public.name_lengths_225002'::regclass;
+SELECT "Column", "Type", "Modifiers" FROM table_desc WHERE relid='public.name_lengths_225002'::regclass ORDER BY 1 DESC, 2 DESC;
```
hanefi: There is no `ORDER BY` on the 3rd column. Is this really intended?

hanefi:
> @hanefi I'm planning to squash the commits before merging, leaving it as is if you want to see each individual change separately.

I am ok with it. 

FYI, You can also use the `Squash and Merge` option on Github and it will show up as a single squash commit on master, but the individual commits will be visible on the PR. In the end it will be the same, so you can choose whichever you want.

onderkalaci:
> FYI, You can also use the Squash and Merge option on Github and it will show up as a single squash commit on master, but the individual commits will be visible on the PR. In the end it will be the same, so you can choose whichever you want.

I personally find it confusing when the commit SHAs on the PR differs from the commit SHAs on the master branch. So I always prefer to squash locally and merge those commits.

metdos:
> FYI, You can also use the `Squash and Merge` option on Github and it will show up as a single squash commit on master, but the individual commits will be visible on the PR. In the end it will be the same, so you can choose whichever you want.

This may also have problems while creating the changelog. Check with Burak V. for details.

PR #2660, opened 2019-04-10
Ensure that master_modify_multiple_shards() returns the affectedTupleCount

onderkalaci:
The PR is against `unified_executor` branch. Our goal is to make sure that we fix issues on the branch one by one.

This change fixes 4 tests directly, so we have 33-35 left 😄 

```
multi_shard_modify
multi_schema_support
multi_outer_join_reference
multi_prepare_sql
```

I didn't like accessing `session->currentTask->shardCommandExecution`, but the other way around is passing `gotResults ` flag as well, which is not nice as well. So, I'm open for suggestion if you have any.

### DIFF
--- a/src/backend/distributed/executor/executor.c
+++ b/src/backend/distributed/executor/executor.c
@@ -91,8 +91,16 @@ typedef struct DistributedExecution
 	/* indicates whether distributed execution has failed */
 	bool failed;
 
-	/* total number of rows received from shard commands */
-	uint64 rowCount;
+	/*
+	 * For SELECT commands or INSERT/UPDATE/DELETE commands with RETURNING,
+	 * the total number of rows received from the workers. For
+	 * INSERT/UPDATE/DELETE commands without RETURNING, the total number of
+	 * tuples modified.
+	 *
+	 * Note that for replicated tables (e.g., reference tables), we only consider
+	 * a single replica's rows that are processed.
+	 */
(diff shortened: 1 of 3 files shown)


src/backend/distributed/executor/executor.c
```diff
@@ -1609,19 +1611,25 @@ ReceiveResults(WorkerSession *session, bool storeRows)
...
 
-			execution->rowCount += currentAffectedTupleCount;
+			/* no more results */
 			return true;
 		}
 		if (resultStatus == PGRES_TUPLES_OK)
 		{
-			execution->rowCount += PQntuples(result);
+			/* we've already consumed all the tuples, no more results */
+			Assert (PQntuples(result) == 0);
```
pykello: `DistributedExecution->rowCount` is a bit confusing. The comment in the struct definition says `total number of rows received from shard commands`, which is:
* correct in the case of `SELECT` (for example, if the shard count is 32, then row count will be `32` for `SELECT count(*) from t`), but then how is this useful?
* incorrect in the case of DML. In this case it is number of rows affected.

If it is not useful for `SELECT`s, then what do you think about renaming it to `rowsAffected` or `rowsProcessed`, and updating its comment, and not incrementing it for `SELECT`?

pykello: With the name/comment updated, it looks good to me.

PR #2661, opened 2019-04-10
Add order by subquery_complex_target_list

onderkalaci:
Similar to #2657 and #2659. The subquery has a `LIMIT` without an `ORDER BY`, which might have inconsistent output.


src/test/regress/expected/subquery_complex_target_list.out
```diff
@@ -310,6 +310,7 @@ SELECT * FROM
 	    WINDOW my_win AS (PARTITION BY user_id ORDER BY time DESC)
 	    ORDER BY rnk DESC
 		) as foo_inner
+     ORDER BY user_id DESC
```
onderkalaci: note that adding `ORDER BY user_id LIMIT 4` is sufficient because we're `PARTITION BY user_id` and joining on `user_id` later.

PR #2662, opened 2019-04-10
Adjust intermediate result size limits to pass the regression tests

onderkalaci:
This PR is against `unified_executor`.

The real-time executor uses COPY command to get the results
from the worker nodes. Unified executor avoids that which
results in less data transfer. Simply adjust the tests to lower
sizes.

### DIFF
--- a/src/test/regress/expected/limit_intermediate_size.out
+++ b/src/test/regress/expected/limit_intermediate_size.out
@@ -1,5 +1,5 @@
 SET citus.enable_repartition_joins to ON;
-SET citus.max_intermediate_result_size TO 3;
+SET citus.max_intermediate_result_size TO 2;
 -- should fail because the copy size is ~4kB for each cte
 WITH cte AS 
 (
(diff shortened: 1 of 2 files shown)


src/test/regress/expected/limit_intermediate_size.out
```diff
@@ -39,42 +39,33 @@ ERROR:  the intermediate result size exceeds citus.max_intermediate_result_size
 DETAIL:  Citus restricts the size of intermediate results of complex subqueries and CTEs to avoid accidentally pulling large result sets into once place.
 HINT:  To run the current query, set citus.max_intermediate_result_size to a higher value or -1 to disable.
 -- router queries should be able to get limitted too
-SET citus.max_intermediate_result_size TO 3;
+SET citus.max_intermediate_result_size TO 2;
```
onderkalaci: Note that the query before the change was pulling all the data to together in the last step (subquery `a`). So, with this commit I made the test comply with the comment. Now, each cte is small, the top level query is only on the aggregates of the results.

Issue #2665, opened 2019-04-12
Get adaptive executor to production ready state


onderkalaci:
We've a prototype for unified executor (#2435 and the main issue: https://github.com/citusdata/citus/issues/1273). The branch implements all the main required features. However, it still requires a lot of effort to get into the production ready state.

We split the whole task into several sub-tasks to get things done in parallel and have smaller granularities to review the sub-tasks.

As we dive deeper into the tasks, we may update the issue to reflect our investigations.

#### Many of the failures are as simple as adjusting the regression tests to produce consistent results over different executions. Some other failures are just because we've forked the branch few releases before and thus some features (like foreign keys to reference tables) are not considered in the branch. So, make sure to fix those smallish integration issues. There are few errors that needs closer look that I've not able to understand in the first glance.

- [x] Regression tests may fail just because execution order of the queries on the shards changes. Real-time executor pretty much always follows the same execution order. So, even if a query misses an `ORDER BY`, the results don't change on every execution.
   - [x] Simply switch to [unified_executor](https://github.com/citusdata/citus/compare/unified_executor?expand=1) branch and run regression tests few times to gather randomly changing regression test outputs. Related PRs #2659, #2661, #2657. There might be some more of these.

   - [x] The harder case is to get consistent output for queries with `RETURNING` clause (`RETURNING` doesn't support `ORDER BY`). Two approaches we could think of: (a) Change the data/queries such that the results returns a single row or multiple of the same rows (b) Wrap the query into a CTE, and add `ORDER BY` to the cte. Note that the latter sounds like changing the query execution semantics so be cautious. Similarly, for the former, we should make sure that we gather results from multiple shards (e.g., don't return distribution key only) to keep the execution semantics the same. (c) Sort tuplestore #2677

- [x] Unified executor does not create intermediate files. `multi_query_directory_cleanup.sql` test has the assumption of intermediate folders/files. We should adjust the tests so that they don't fail (#2728).

- [x] With the new executor, the intermediate data sizes change since we're avoiding `COPY`. See `limit_intermediate_size.sql` test. Related PR: #2662

- [x] `master_modify_multiple_shards()` seems to have a bug. It is probably as simple as not setting the number of tuples that are modified. Related PR: #2660

- [x] Marco has started working on this branch few releases before. Thus, the current prototype does not consider few of the newest Citus features. We should make sure to enable them: #2688 and #2692 
[...]

PR #2666, opened 2019-04-16
    Force one connection per placement (or co-located placements) when requested

onderkalaci:
This PR is against unified executor branch. This needs some more work, (a) add more tests that'd be fixed by this (b) when all the regression tests are executed with this flag `+push(@pgOptions, '-c', "citus.force_max_query_parallelization=on");`, one test crashes -- which might be directly related to this or maybe amplified by this (`multi_modifying_xacts.sql`)

    
    The existing executors (real-time and router) always open 1 connection per
    placement when parallel execution is requested.
    
    That might be useful under certain circumstances:
    
    (a) User wants to utilize as much as CPUs on the workers per
    distributed query
    (b) User has a transaction block which involves COPY command
    
    Also, lots of regression tests rely on this execution semantics.
    So, we'd enable few of the tests with this change as well.

### DIFF
--- a/src/backend/distributed/executor/executor.c
+++ b/src/backend/distributed/executor/executor.c
@@ -156,6 +156,9 @@ struct TaskPlacementExecution;
  */
 typedef struct WorkerSession
 {
+	/* only useful for debugging */
+	uint64 sessionId;
+
 	/* worker pool of which this session is part */
 	WorkerPool *workerPool;
 
@@ -170,6 +173,12 @@ typedef struct WorkerSession
 
 	/* task the worker should work on or NULL */
 	struct TaskPlacementExecution *currentTask;
+
+	/*
+	 * The number of commands sent to the worker over the session. Excludes
+	 * distributed transaction related commands such as BEGIN/COMMIT etc.
+	 */
+	uint64 commandsSent;
 } WorkerSession;
 
 
@@ -280,6 +289,10 @@ typedef struct TaskPlacementExecution
 } TaskPlacementExecution;
 
 
+/* GUC, determining whether Citus opens 1 connection per task */
+bool ForceMaxQueryParallelization = false;
+
+
 /* local functions */
 static DistributedExecution * CreateDistributedExecution(DistributedPlan *distributedPlan,
 														 CitusScanState *scanState,
@@ -290,6 +303,9 @@ static void FinishDistributedExecution(DistributedExecution *execution);
 
 static bool DistributedStatementRequiresRollback(DistributedPlan *distributedPlan);
 static void AssignTasksToConnections(DistributedExecution *execution);
+static bool ShouldForceOpeningConnection(TaskPlacementExecutionState
+										 placementExecutionState,
+										 PlacementExecutionOrder placementExecutionOrder);
 static PlacementExecutionOrder ExecutionOrderForTask(CmdType operation, Task *task);
 static WorkerPool * FindOrCreateWorkerPool(DistributedExecution *execution,
 										   WorkerNode *workerNode);
@@ -348,6 +364,14 @@ CitusExecScan(CustomScanState *node)
 		if (MultiShardConnectionType == SEQUENTIAL_CONNECTION)
 		{
 			targetPoolSize = 1;
+		}
+		else if (ForceMaxQueryParallelization)
+		{
+			Job *job = distributedPlan->workerJob;
+			List *taskList = job->taskList;
+
+			/* we might at most have the length of tasks pool size */
+			targetPoolSize = list_length(taskList);
 		}
 
 		execution = CreateDistributedExecution(distributedPlan, scanState,
@@ -531,6 +555,9 @@ FinishDistributedExecution(DistributedExecution *execution)
 		RemoteTransaction *transaction = &(connection->remoteTransaction);
 		RemoteTransactionState transactionState = transaction->transactionState;
 
+		elog(DEBUG4, "Total number of commands sent over the session %ld: %ld",
+			 session->sessionId, session->commandsSent);
+
 		UnclaimConnection(connection);
 
 		if (connection->connectionState == MULTI_CONNECTION_CONNECTING)
@@ -641,6 +668,26 @@ AssignTasksToConnections(DistributedExecution *execution)
 			connection = GetPlacementListConnectionIfCached(connectionFlags,
 															placementAccessList,
 															NULL);
+
+			/*
+			 * If there is no cached connection that we have to use and we're instructed
+			 * to force max query parallelization, we try to do it here.
+			 *
+			 * In the end, this is the user's choice and we should try to open all the necessary
+			 * connections. Doing it here the code a lot so we prefer to do it.
+			 */
+			if (connection == NULL &&
+				ShouldForceOpeningConnection(placementExecution->executionState,
+											 shardCommandExecution->executionOrder))
+			{
+				int connectionFlags = CONNECTION_PER_PLACEMENT;
+
+				/* ConnectionStateMachine will finish the connection establishment */
+				connection =
+					StartPlacementListConnection(connectionFlags, placementAccessList,
+												 NULL);
+			}
+
 			if (connection != NULL)
 			{
 				RemoteTransaction *transaction = &(connection->remoteTransaction);
@@ -669,12 +716,6 @@ AssignTasksToConnections(DistributedExecution *execution)
 					dlist_push_tail(&session->pendingTaskQueue,
 									&placementExecution->sessionPendingQueueNode);
 				}
-
-				/*
-				 * Not all parts of the code set connection state. For now we set
-				 * it explicitly here.
-				 */
-				connection->connectionState = MULTI_CONNECTION_CONNECTED;
 
 				/* always poll the connection in the first round */
 				connection->waitFlags = WL_SOCKET_READABLE | WL_SOCKET_WRITEABLE;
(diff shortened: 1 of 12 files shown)


src/backend/distributed/executor/executor.c
```diff
@@ -817,6 +869,7 @@ FindOrCreateWorkerSession(WorkerPool *workerPool, MultiConnection *connection)
 	DistributedExecution *execution = workerPool->distributedExecution;
 	WorkerSession *session = NULL;
 	ListCell *sessionCell = NULL;
+	static int sessionId = 1;
```
onderkalaci: make this uint64

src/backend/distributed/shared_library_init.c
```diff
@@ -604,6 +604,21 @@ RegisterCitusConfigVariables(void)
 		GUC_NO_SHOW_ALL,
 		NULL, NULL, NULL);
 
+	DefineCustomBoolVariable(
+		"citus.force_max_query_parallelization",
+		gettext_noop("Open as many connections as possible to maximize query "
+					 "parallelization"),
+		gettext_noop("When enabled, Citus will force the executor to use "
```
onderkalaci: mention that sequential mode is favored over this.

src/backend/distributed/executor/executor.c
```diff
@@ -637,6 +664,21 @@ AssignTasksToConnections(DistributedExecution *execution)
...
+			 * to force max query parallelization, we do it here with a blocking fashion,
+			 * contrary to the usual approach of non-blocking connection establishment.
+			 *
+			 * In the end, this is the user's choice and we have to open all the necessary
+			 * connections. Doing it here with blocking API simplifies the code a lot so we
+			 * prefer to do it.
+			 */
+			if (connection == NULL && ForceMaxQueryParallelization)
+			{
+				connection = GetPlacementListConnection(0, placementAccessList, NULL);
```
onderkalaci: Make it SESSION_LIFESPAN as done in the other place for completeness.

src/backend/distributed/executor/executor.c
```diff
@@ -348,6 +361,17 @@ CitusExecScan(CustomScanState *node)
 		if (MultiShardConnectionType == SEQUENTIAL_CONNECTION)
 		{
 			targetPoolSize = 1;
+
+			/* we favor MultiShardConnectionType over ForceMaxQueryParallelization */
+			ForceMaxQueryParallelization = false;
```
onderkalaci: either (a) create a local copy of `ForceMaxQueryParallelization ` in the execution struct. Only set this locally in the execution (b) remove this setting here and act accordingly when  `MultiShardConnectionType == SEQUENTIAL_CONNECTION`

Otherwise, it is easy to mass things up by changing both settings at the same time.

src/backend/distributed/executor/executor.c
```diff
@@ -1844,14 +1902,19 @@ PlacementExecutionDone(TaskPlacementExecution *placementExecution, bool succeede
...
 
 		/* find a placement execution that is not yet marked as failed */
 		do {
 			int nextPlacementExecutionIndex =
 				placementExecution->placementExecutionIndex + 1;
 
+			if (nextPlacementExecutionIndex == placementExecutionCount)
+			{
+				/* adjust the index if the placement execution is the last one */
+				nextPlacementExecutionIndex = 0;
```
marcocitus: Don't we always start at 0 and count up (meaning we shouldn't have to wrap around)?

Line above this `if` statement is a bit suspicious seems `placementExecution doesn't change in this loop.

onderkalaci: As far as I understand, if the first placement execution fails, we might start from > 0?

marcocitus: But when do we need to wrap around?

src/backend/distributed/executor/executor.c
marcocitus: `GetPlacementListConnection` synchronously establishes the connection, we should change this to `StartPlacementListConnection`

src/backend/distributed/executor/executor.c
```diff
@@ -681,6 +723,12 @@ AssignTasksToConnections(DistributedExecution *execution)
 				{
 					workerPool->idleConnectionCount++;
 				}
+
+				/* we don't want any connection to be used by any other session */
+				if (ForceMaxQueryParallelization)
+				{
+					ClaimConnectionExclusively(connection);
```
marcocitus: This seems a little off. The preferred way of getting a new connection for every placement is to use the `CONNECTION_PER_PLACEMENT` flag in `StartPlacementListConnection`. That also ensures that, if we modify shard 1 and then query shard 1 and 2,  we avoid using the connection that modified shard 1 to read from shard 2, which would leave shard 1 without a suitable connection.

src/backend/distributed/executor/executor.c
marcocitus: That assumption that's encoded into this function is that we only reach this code if the task has not failed yet. That means that the placement execution that has failed cannot be the last one and therefore placementExecutionIndex + 1 cannot exceed the count.

However, this assumption is violated when a worker pool fails and we mark all placement executions on that worker as failed.

The `if` statement that checks the execution order originally had an additional condition which checked that the state of placementExecution was RUNNING at the start of the function, which means it must have been the "current" placement execution. 

In general, this is my least favourite part of the code because it's built on implicit assumptions and that is already leading to bugs. It could use some refactoring.

src/backend/distributed/executor/executor.c
```diff
@@ -1844,14 +1902,19 @@ PlacementExecutionDone(TaskPlacementExecution *placementExecution, bool succeede
...
 	{
 		TaskPlacementExecution *nextPlacementExecution = NULL;
-		int placementExecutionCount PG_USED_FOR_ASSERTS_ONLY =
-			shardCommandExecution->placementExecutionCount;
+		int placementExecutionCount = shardCommandExecution->placementExecutionCount;
 
 		/* find a placement execution that is not yet marked as failed */
 		do {
 			int nextPlacementExecutionIndex =
 				placementExecution->placementExecutionIndex + 1;
```
marcocitus: I'm a little puzzled by this line, since I would expect `nextPlacementExecutionIndex` to keep incrementing as we iterate the loop.

src/backend/distributed/executor/executor.c
pykello: Did you discuss this internally? I have similar concerns to Marco. Isn't that the case because we try the placements sequentially, we have already tried all indexes before `placementExecutionIndex`, so we don't need to wrap around?

src/test/regress/sql/multi_real_time_transaction.sql
```diff
@@ -220,14 +224,17 @@ SET LOCAL citus.multi_shard_modify_mode TO 'sequential';
...
+
 SET client_min_messages TO DEFAULT;
 alter system set deadlock_timeout TO DEFAULT;
 SELECT pg_reload_conf();
 
 BEGIN;
 SET citus.select_opens_transaction_block TO off;
 -- This query would self-deadlock if it ran in a distributed transaction
-SELECT id, pg_advisory_lock(15) FROM test_table ORDER BY id;
+-- we use a different advisory lock because previous tests
```
onderkalaci: A sad story that took me really long to understand what is going on. Basically, since the executor does not disconnect (e.g., due to `SESSION_LIFESPAN`), the lock acquired by the above test leads to blocking this one.

So, an example of what I hit where `ROLLBACK` doesn't release `pg_advisory_lock `. Well, that's obvious from the documentation, but still it was very confusing for me because previously real-time executor always disconnects so we haven't had this issue at all. I mostly thought that this is a bug in the new executor, which turned out to be wrong...



```SQL
BEGIN
SELECT  pg_advisory_lock(15);
 pg_advisory_lock 
------------------
 
(1 row)

SELECT locktype, objid, mode, granted FROM pg_locks where pid = pg_backend_pid();
  locktype  | objid |      mode       | granted 
------------+-------+-----------------+---------
 relation   |       | AccessShareLock | t
 relation   |       | AccessShareLock | t
 virtualxid |       | ExclusiveLock   | t
 advisory   |    15 | ExclusiveLock   | t
(4 rows)

ROLLBACK;

SELECT locktype, objid, mode, granted FROM pg_locks where pid = pg_backend_pid();
  locktype  | objid |      mode       | granted 
------------+-------+-----------------+---------
 relation   |       | AccessShareLock | t
 virtualxid |       | ExclusiveLock   | t
 advisory   |    15 | ExclusiveLock   | t
(3 rows)
```

src/test/regress/expected/with_dml.out
```diff
@@ -106,6 +106,10 @@ DEBUG:  Plan 16 query after replacing subqueries and CTEs: SELECT tenant_id FROM
 -- not a very meaningful query
 -- but has two modifying CTEs along with another
 -- modify statement
+-- We need to force 1 connection per placement
+-- otherwise the coordinator insert select fails
+-- since COPY cannot be executed
+SET citus.force_max_query_parallelization TO on;
```
onderkalaci: Note that this is a regression that the new executor leads to. I added an item to the main issue #2665:

```
The unified executor may regress for certain types of distributed queries. 
The most notable ones is that modifying ctes followed by an coordinator insert/select. The 
reason is that the previous modifications might use the same connections for modifying
 the same placements. However, the final coordinator INSERT/SELECT would require 1
 connection per placement. We should probably automatically detect these cases and 
force 1 connection per placement in advance.
```
However, forcing here to (a) make more regression tests pass (b) test the new GUC with a real use-case for it.

Let me know if this doesn't look good to you.

PR #2669, opened 2019-04-16
Fix self strncmp

jasonmp85:
DESCRIPTION: Fix bad foreign constraint name search

See the commit messages for more details.

### DIFF
--- a/configure
+++ b/configure
@@ -4008,6 +4008,41 @@ if test x"$citusac_cv_prog_cc_cflags__Wmissing_prototypes" = x"yes"; then
   CITUS_CFLAGS="$CITUS_CFLAGS -Wmissing-prototypes"
 fi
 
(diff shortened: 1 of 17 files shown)


metdos:
This PR seems approved, any reason to keep this open?

jasonmp85:
> This PR seems approved, any reason to keep this open?

Mostly my own paranoia that I changed behavior somewhere in all these variable renames… just wanted time to go over it thoroughly a second time. Doing so now and will merge.

jasonmp85:
> Not sure whether this is worth adding a regression test. Fine with the PR as is.

I mostly considered the compiler warning its own kind of test to prevent the old behavior. I don't think this is necessarily worth another runtime test. I'll leave it as-is and merge.

jasonmp85:
Double-checked everything; looks good. Merging.

PR #2672, opened 2019-04-17
Resolve parameters before using them

onderkalaci:
The PR is against `unified_executor` branch.

For the details, see PostgreSQL's copyParamList().

This fixes several query failures, but not the failing tests yet. There are some minor unrelated failures on the same files, which will be handled separately.


onderkalaci:
> This fixes several query failures, but not the failing tests yet

Oh, that's wrong. This already fixes 3 tests as well: `fast_path_router_modify`, `multi_prepare_plsql` and `sql_procedure`.

PR #2673, opened 2019-04-17
Fix savepoint rollback after multi-shard modify/copy failure.

### LINKED ISSUE #2668: Savepoint rollback after failed multi-shard update doesn't work properly
For example, the following case:

```
postgres=# DROP TABLE IF EXISTS t;
DROP TABLE
Time: 37.887 ms
postgres=#
postgres=# SET citus.shard_count TO 2;
SET
postgres=#
postgres=# CREATE TABLE t(a int, b int CHECK (b > 0));
CREATE TABLE
postgres=# SELECT create_distributed_table('t', 'a');
 create_distributed_table
--------------------------

(1 row)
postgres=# BEGIN;
BEGIN
postgres=# INSERT INTO t SELECT i, i+1 FROM generate_series(1, 5) i;
INSERT 0 5
postgres=# SAVEPOINT sp1;
SAVEPOINT
postgres=# UPDATE t SET b = b - 2;
WARNING:  connection error: localhost:5433
ERROR:  new row for relation "t_1330003" violates check constraint "t_b_check"
DETAIL:  Failing row contains (1, 0).
CONTEXT:  while executing command on localhost:5433
postgres=# ROLLBACK TO SAVEPOINT sp1;
ROLLBACK
postgres=# INSERT INTO t VALUES (10, 11);
WARNING:  savepoint "savepoint_3" does not exist
ERROR:  cannot establish a new connection for placement 608, since DML has been executed on a connection that is in use
postgres=# COMMIT;
ROLLBACK
```

I'll add more to here as I understand the reason better.

pykello:
To fix this, we needed to do:
1. Clear & cancel any queries on the connections we want to send rollback to.
2. Unclaim connections after multi-shard modify fails.

Fixes #2668.

DESCRIPTION: Fix savepoint rollback after multi-shard modify/copy failure.

### DIFF
--- a/src/backend/distributed/commands/multi_copy.c
+++ b/src/backend/distributed/commands/multi_copy.c
@@ -154,6 +154,8 @@ static void CopySendInt32(CopyOutState outputState, int32 val);
 static void CopySendInt16(CopyOutState outputState, int16 val);
 static void CopyAttributeOutText(CopyOutState outputState, char *string);
 static inline void CopyFlushOutput(CopyOutState outputState, char *start, char *pointer);
+static bool CitusSendTupleToPlacements(TupleTableSlot *slot,
+									   CitusCopyDestReceiver *copyDest);
 
 /* CitusCopyDestReceiver functions */
 static void CitusCopyDestReceiverStartup(DestReceiver *copyDest, int operation,
@@ -2251,8 +2253,33 @@ CitusCopyDestReceiverStartup(DestReceiver *dest, int operation,
 static bool
 CitusCopyDestReceiverReceive(TupleTableSlot *slot, DestReceiver *dest)
 {
+	bool result = false;
 	CitusCopyDestReceiver *copyDest = (CitusCopyDestReceiver *) dest;
 
+	PG_TRY();
+	{
+		result = CitusSendTupleToPlacements(slot, copyDest);
+	}
+	PG_CATCH();
+	{
+		HTAB *shardConnectionHash = copyDest->shardConnectionHash;
+		UnclaimAllShardConnections(shardConnectionHash);
+
+		PG_RE_THROW();
+	}
+	PG_END_TRY();
+
+	return result;
+}
+
+
+/*
+ * CitusSendTupleToPlacements sends the given TupleTableSlot to the appropriate
+ * shard placement(s).
+ */
+static bool
+CitusSendTupleToPlacements(TupleTableSlot *slot, CitusCopyDestReceiver *copyDest)
+{
 	int partitionColumnIndex = copyDest->partitionColumnIndex;
 	TupleDesc tupleDescriptor = copyDest->tupleDescriptor;
 	CopyStmt *copyStatement = copyDest->copyStatement;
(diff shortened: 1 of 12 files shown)


src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -1499,28 +1500,38 @@ ExecuteModifyTasks(List *taskList, bool expectResults, ParamListInfo paramListIn
 				SetCitusNoticeLevel(INFO);
 			}
 
-			/*
-			 * If caller is interested, store query results the first time
-			 * through. The output of the query's execution on other shards is
-			 * discarded if we run there (because it's a modification query).
-			 */
-			if (placementIndex == 0 && expectResults)
+			PG_TRY();
```
pykello: `ConsumeQueryResult()` might raise errors with multiple `errcontext()`s. If we wanted to change DefferedErrors to support this, change would be huge, so I am using a `PG_TRY()` block instead.

marcocitus: Are there other execution paths we should consider doing this for? e.g. COPY,

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -1068,6 +1068,12 @@ CoordinatedRemoteTransactionsSavepointRollback(SubTransactionId subId)
 		MultiConnection *connection = dlist_container(MultiConnection, transactionNode,
 													  iter.cur);
 		RemoteTransaction *transaction = &connection->remoteTransaction;
+
+		/* cancel any ongoing queries before issuing rollback */
+		ClearResultsIfReady(connection);
```
marcocitus: Probably not necessary to do this before the cancellation. The cancellation actually goes over a separate connection.

src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -1068,6 +1068,12 @@ CoordinatedRemoteTransactionsSavepointRollback(SubTransactionId subId)
...
+		SendCancelationRequest(connection);
+		ForgetResults(connection);
```
marcocitus: We once planned to deprecate `ForgetResults`. This should maybe be `ClearResults(connection, false);`

src/backend/distributed/commands/multi_copy.c
```diff
@@ -2251,8 +2253,33 @@ CitusCopyDestReceiverStartup(DestReceiver *dest, int operation,
 static bool
 CitusCopyDestReceiverReceive(TupleTableSlot *slot, DestReceiver *dest)
 {
+	bool result = false;
 	CitusCopyDestReceiver *copyDest = (CitusCopyDestReceiver *) dest;
 
+	PG_TRY();
+	{
+		result = CitusSendTupleToPlacements(slot, copyDest);
```
pykello: Moved all of ex-`CitusCopyDestReceiverReceive()` to `CitusSendTupleToPlacements()`, so it is easier to surround it with `PG_TRY()`.

src/backend/distributed/transaction/remote_transaction.c
pykello: removed.

src/backend/distributed/transaction/remote_transaction.c
pykello: done.

src/backend/distributed/executor/multi_router_executor.c
pykello: Added `COPY`, still looking for other execution paths ...

pykello:
> I guess there is an implicit assumption that savepoints are only used to continue after SQL-level failures (e.g. constraint violation) and not network failures. I think that assumption is ok, but can we prevent rollback to savepoint after a network failure somehow?

PostgreSQL inserts an implicit `SAVEPOINT` command after a `ROLLBACK TO SAVEPOINT` command. Current behavior is that if there is a network failure, the implicit `SAVEPOINT` command errors out, so the rollback to savepoint is effectively prevented. I am not sure if we can error at the `ROLLBACK TO SAVEPOINT` itself, from what I remember that might make PostgreSQL's internal state inconsistent. Can we keep this to a separate PR as it might need some more inspection?

pykello:
> Can you add some comments above the call to UnclaimAllSahrdConnection on why we need to do this? It's not clear from the code that it has to do with savepoints.

done.

PR #2676, opened 2019-04-18
Fix savepoints in unified executor.

pykello:
It turns out that all we need to do is #2673 + reset state before throwing errors.

### DIFF
--- a/src/backend/distributed/executor/executor.c
+++ b/src/backend/distributed/executor/executor.c
@@ -303,6 +303,7 @@ static void FinishDistributedExecution(DistributedExecution *execution);
 
 static bool DistributedStatementRequiresRollback(DistributedPlan *distributedPlan);
 static void AssignTasksToConnections(DistributedExecution *execution);
+static void UnclaimAllSessionConnections(List *sessionList);
 static bool ShouldForceOpeningConnection(TaskPlacementExecutionState
 										 placementExecutionState,
 										 PlacementExecutionOrder placementExecutionOrder);
@@ -602,6 +603,23 @@ FinishDistributedExecution(DistributedExecution *execution)
 	}
 }
 
+
+/*
+ * UnclaimAllSessionConnections unclaims all of the connections for the given
+ * sessionList.
+ */
+static void
+UnclaimAllSessionConnections(List *sessionList)
+{
+	ListCell *sessionCell = NULL;
+	foreach(sessionCell, sessionList)
+	{
+		WorkerSession *session = lfirst(sessionCell);
+		MultiConnection *connection = session->connection;
+
+		UnclaimConnection(connection);
+	}
+}
 
 /*
  * AssignTasksToConnections goes through the list of tasks to determine whether any
@@ -1039,6 +1057,12 @@ RunDistributedExecution(DistributedExecution *execution)
 	}
 	PG_CATCH();
 	{
+		/*
+		 * We can still recover from error using ROLLBACK TO SAVEPOINT,
+		 * unclaim all connections to allow that.
+		 */
+		UnclaimAllSessionConnections(execution->sessionList);
+
 		/* make sure the epoll file descriptor is always closed */
 		if (waitEventSet != NULL)
 		{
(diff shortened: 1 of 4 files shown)


pykello:
@onderkalaci please review https://github.com/citusdata/citus/pull/2676/commits/7d8463ac75024c27086e469f25d799326611d472. The other commit is just simply #2673, which is to be reviewed separately.

Right now all tests in `multi_subtransactions.sql` except one test pass, and the failure in that test is not because of the savepoint code. The following snippet which doesn't contain a savepoint fails for the same reason:

```sql
DROP TABLE IF EXISTS artists;

CREATE TABLE artists (
    id bigint NOT NULL,
    name text NOT NULL
);
SELECT create_distributed_table('artists', 'id');

BEGIN;
DELETE FROM artists;
INSERT INTO artists SELECT i, NULL FROM generate_series(1, 5) i;
ERROR:  cannot establish a new connection for placement 301, since DML has been executed on a connection that is in use
```

I will look into this separately.

src/backend/distributed/executor/executor.c
```diff
@@ -538,8 +538,15 @@ FinishDistributedExecution(DistributedExecution *execution)
...
-
-		/* TODO: can we be in SENT_COMMAND? */
+		else if (transactionState == REMOTE_TRANS_SENT_COMMAND)
+		{
+			/*
+			 * Result of a command contains errors. So we change the state
+			 * to clear results and reset state, so we can process future
+			 * ROLLBACK TO SAVEPOINTs.
+			 */
+			transaction->transactionState = REMOTE_TRANS_CLEARING_RESULTS;
```
marcocitus: As discussed, I think the main effect of this is that it puts the transaction back into a state from which it is safe to re-enter the unified executor, whereas `REMOTE_TRANS_SENT_COMMAND` is not safe. In case of a rollback to savepoint we already clear the results in `FinishRemoteTransactionSavepointRollback` so we do not actually need to re-enter that state. However, `FinishRemoteTransactionSavepointRollback` currently neglects to update `transactionState` to `REMOTE_TRANS_STARTED`.

src/backend/distributed/executor/executor.c
```diff
@@ -1643,6 +1650,12 @@ ReceiveResults(WorkerSession *session, bool storeRows)
 		}
 		else if (resultStatus != PGRES_SINGLE_TUPLE)
 		{
+			/*
+			 * We can still recover from error using ROLLBACK TO SAVEPOINT,
+			 * cleanup state to allow that.
+			 */
+			FinishDistributedExecution(execution);
```
marcocitus: Same comment as https://github.com/citusdata/citus/pull/2673#pullrequestreview-230549746

We should make sure you cannot roll back to a savepoint after a connection failure.

src/backend/distributed/executor/executor.c
pykello: Removing this and changing the transaction state in `FinishRemoteTransactionSavepointRollback`.

src/backend/distributed/executor/executor.c
pykello: From what I remember, @anarazel said it is not safe to error out in `ROLLBACK TO SAVEPOINT`. However postgres inserts an implicit `SAVEPOINT` command right after the rollback, which errors out if there is a connection error, so effectively not allowing rolling back to savepoints in case of connection errors.

I need to do more postgres code reading/inspection to see if we can have a better solution.

pykello:
@marcocitus rebased and cleaned-up. Please take a look.

src/backend/distributed/executor/executor.c
marcocitus: I'm still slightly concerned about other points we could error out and continue (E.g. in `ErrorSizeLimitIsExceeded`). Any thoughts?

marcocitus: ping

pykello: Makes sense. Moved the unclaim to the `PG_CATCH()` of `RunDistributedExecution()` to cover all errors, what do you think?

PR #2677, opened 2019-04-19
Sort output of RETURNING

onderkalaci:
The feature is only intended for getting consistent outputs for the regression tests. So, do not consider this as a user-facing feature.

RETURNING does not have any ordering gurantees and with unified executor (#2665), the ordering
of query executions on the shards are also becoming unpredictable. Thus, we're enforcing
ordering when a GUC is set.

We implicitly add an `ORDER BY` something equivalent of
```SQL
RETURNING expr1, expr2, .. ,exprN
ORDER BY expr1, expr2, .. ,exprN
```


src/backend/distributed/shared_library_init.c
```diff
@@ -679,6 +679,20 @@ RegisterCitusConfigVariables(void)
...
+		"citus.sort_returning",
+		gettext_noop("Sorts the RETURNING clause to get consistent test output"),
+		gettext_noop("This feature is not intended for users. It is developed "
+					 "to get consistent regression test outputs. When enabled, "
+					 "the RETURNING clause returns the tuples sorted. The sort "
+					 "is done for all the entries, starting from the first one."
+					 "Finally, the sorting is done in ASC order."),
+		&SortReturning,
+		false,
+		PGC_SUSET,
```
onderkalaci: @marcoslot I'm curious what you think about `PGC_SUSET ` for this setting. Given that the code is not efficient, I'd suggest to keep it super-user only guc for not.

onderkalaci:
> RETURNING does not have any ordering gurantees and with unified executor (#2665), the ordering
of query executions on the shards are also becoming unpredictable

Once we merge it to the master, I'll simple open another PR against `unified_executor` and call this function.

src/backend/distributed/shared_library_init.c
onderkalaci: > @marcoslot I'm curious what you think about PGC_SUSET for this setting. Given that the code is not efficient, I'd suggest to keep it super-user only guc for not.

I'm planning to merge this as-is for now. We can always revisit the decision.

onderkalaci: > I'm planning to merge this as-is for now. We can always revisit the decision.

Ah, sorry @pykello hasn't approved this yet 😄 So, I'm not merging now.

pykello: Looks ok.

> > I'm planning to merge this as-is for now. We can always revisit the decision.
> 
> Ah, sorry @pykello hasn't approved this yet smile So, I'm not merging now.

Looks ok, with some minor comments. Doing another pass for more minor feedback.

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -603,6 +614,122 @@ RouterModifyExecScan(CustomScanState *node)
...
+	Oid *collations = (Oid *) palloc(numberOfSortKeys * sizeof(Oid));
+	bool *nullsFirst = (bool *) palloc(numberOfSortKeys * sizeof(bool));
+
+	ListCell *targetCell = NULL;
+	int sortKeyIndex = 0;
+
+	Tuplesortstate *tuplesortstate = NULL;
+
+	/*
+	 * Itereate on the returning target list and generate the necessary information
```
pykello: `Itereate` -> `Iterate`.

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -603,6 +614,122 @@ RouterModifyExecScan(CustomScanState *node)
...
+		TargetEntry *returningEntry = (TargetEntry *) lfirst(targetCell);
+		Oid sortop = 0;
+		Oid eqop;
+		bool hashable;
+
+		/* determine the eqop and optional sortop */
+		get_sort_group_operators(exprType((Node *) returningEntry->expr),
+								 true, false, false,
+								 &sortop, &eqop, NULL,
+								 &hashable);
```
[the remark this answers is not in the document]
pykello: We can pass `NULL` if we are not using `hashable`.

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -603,6 +614,122 @@ RouterModifyExecScan(CustomScanState *node)
...
+	Tuplesortstate *tuplesortstate = NULL;
+
+	/*
+	 * Itereate on the returning target list and generate the necessary information
+	 * for sorting the tuples.
+	 */
+	foreach(targetCell, targetList)
+	{
+		TargetEntry *returningEntry = (TargetEntry *) lfirst(targetCell);
+		Oid sortop = 0;
```
pykello: `0` -> `InvalidOid`.

src/backend/distributed/shared_library_init.c
pykello: done with the review :smile:.

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -603,6 +614,122 @@ RouterModifyExecScan(CustomScanState *node)
...
+		/* determine the eqop and optional sortop */
+		get_sort_group_operators(exprType((Node *) returningEntry->expr),
+								 true, false, false,
+								 &sortop, &eqop, NULL,
+								 &hashable);
+
+		sortColIdx[sortKeyIndex] = sortKeyIndex + 1;
+		sortOperators[sortKeyIndex] = sortop;
+		collations[sortKeyIndex] = exprCollation((Node *) returningEntry->expr);
+		nullsFirst[sortKeyIndex] = false;
```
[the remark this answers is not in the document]
onderkalaci: Hmm, I think we cannot have resjunk entries with `RETURNING`. Still, I don't see any meaningful reason to call `ExecCleanTargetListLength ()`. That's probably a copy & paste error. Fixing it.

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -603,6 +614,122 @@ RouterModifyExecScan(CustomScanState *node)
...
+
+		if (TupIsNull(slot))
+		{
+			break;
+		}
+
+		newSlot = MakeSingleTupleTableSlot(tupleDescriptor);
+		ExecCopySlot(newSlot, slot);
+
+		tuplesort_puttupleslot(tuplesortstate, newSlot);
```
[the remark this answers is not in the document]
onderkalaci: Makes sense for both here and the below loop.

onderkalaci:
> As described in the code comments as well, this is probably not the most
performant approach we could implement

Thanks to @pykello's comment's, that's not the case anymore. I update the code comments and removing it from the PR description.

PR #2682, opened 2019-04-24
Ensure that unified executor doesn't ignore sequential execution of DDLJob's

onderkalaci:
Certain DDL commands, mainly creating foreign keys to reference tables,
should be executed sequentially. Otherwise, we'd end up with a self
distributed deadlock.

To overcome this situation, we set a flag `DDLJob->executeSequentially`
and execute it sequentially. Note that we have to do this because
the command might not be called within a transaction block, and
we cannot call `SetLocalMultiShardModifyModeToSequential()`.

This fixes at least two test: `multi_insert_select_on_conflit.sql` and
`multi_foreign_key.sql`

Also, I wouldn't mind scattering local `targetPoolSize` variables within
the code. The reason is that we'll soon have a GUC (or a global
variable based on a GUC) that'd set the pool size. In that case, we'd
simply replace `targetPoolSize` with the global variables.

### DIFF
--- a/src/backend/distributed/commands/utility_hook.c
+++ b/src/backend/distributed/commands/utility_hook.c
@@ -535,6 +535,7 @@ static void
 ExecuteDistributedDDLJob(DDLJob *ddlJob)
 {
 	bool shouldSyncMetadata = ShouldSyncTableMetadata(ddlJob->targetRelationId);
+	int targetPoolSize = ddlJob->executeSequentially ? 1 : DEFAULT_POOL_SIZE;
 
 	EnsureCoordinator();
 	EnsurePartitionTableNotReplicated(ddlJob->targetRelationId);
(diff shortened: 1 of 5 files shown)


src/backend/distributed/commands/utility_hook.c
```diff
@@ -535,6 +535,7 @@ static void
 ExecuteDistributedDDLJob(DDLJob *ddlJob)
 {
 	bool shouldSyncMetadata = ShouldSyncTableMetadata(ddlJob->targetRelationId);
+	int targetPoolSize = ddlJob->executeSequentially ? 1 : 4;
```
pykello: Maybe this PR is a good one to make 4 a constant? Then in future if we want to make this a GUC we will have an easier time finding all occurrences of it.

onderkalaci: I don't want to it just right now because we might pick a different API such that instead of exposing `targetPoolSize ` we might consider something else. See some [ideas](https://docs.google.com/document/d/1MuJ6V6E7X_25J4kFN0uXFiFwLMcq8h4I0PGJd-XqqKA/edit?usp=drive_web&ouid=103285309091993434178), which requires some more discussion.

Does that sound OK to you?

PR #2687, opened 2019-04-25
Enable connection reuse during INSERT...SELECT via coordinator

marcocitus:
DESCRIPTION: Enable connection reuse during INSERT...SELECT via coordinator

This PR should probably only be considered once the unified executor is merged. It extends the two-phase INSERT..SELECT logic we use for RETURNING and ON CONFLICT to cases where multiple shards (or rather, sets of co-located shard placements) of the same table (or rather, set of co-located tables) were modified over a single connection. The goal is to avoid regressions in the unified executor such as:

```sql
BEGIN;
-- Multi-shard command, intelligently selects connections
DELETE FROM table;
-- Hidden COPY, needs one connection per shard placement
INSERT INTO table SELECT * FROM local_table;
```

When shards were modified over the same connection, then instead copying directly into a shard placement, which is only possible using a dedicated connection, we first copy into an intermediate result. Because intermediate results can be used across connections, copying into a set of intermediate results can be done independently from any other operations in the transaction. 

In the second phase we do an insert...select which selects from each intermediate result into the corresponding shard placement. The second phase can be handled by the unified executor, which can appropriately assign tasks to connections based on preceding commands in the transaction.

Currently, the decision to do the insert..select in two phases is made in the planner, but the state of connections isn't known until execution time. Therefore, this PR moves it and the corresponding code to the executor.

One of the regression tests is failing. This is because the current unified_executor finishes execution before all connections are fully established, which needs to be resolved per the item in:
https://github.com/citusdata/citus/issues/2665

This approach could be extended to COPY itself, though it will require some additional refactoring.


src/backend/distributed/executor/insert_select_executor.c
```diff
@@ -147,6 +174,279 @@ CoordinatorInsertSelectExecScan(CustomScanState *node)
...
+	return selectQuery;
+}
+
+
+/*
+ * WrapSubquery wraps the given query as a subquery in a newly constructed
+ * "SELECT * FROM (...subquery...) citus_insert_select_subquery" query.
+ */
+static Query *
+WrapSubquery(Query *subquery)
```
marcocitus: This function was moved from insert_select_planner.c and left unchanged

src/backend/distributed/executor/insert_select_executor.c
```diff
@@ -147,6 +174,279 @@ CoordinatorInsertSelectExecScan(CustomScanState *node)
...
+}
+
+
+/*
+ * TwoPhaseInsertSelectTaskList generates a list of task for a query that
+ * inserts into a target relation and selects from a set of co-located
+ * intermediate results.
+ */
+static List *
+TwoPhaseInsertSelectTaskList(Oid targetRelationId, Query *insertSelectQuery,
```
marcocitus: This function was moved from insert_select_planner.c and left unchanged

marcocitus:
Superseded by #2758

PR #2688, opened 2019-04-25
Refactor Parallel Relation Access Recording

onderkalaci:
Instead of scattering the code around, we move all the
logic into a single function.

This will help supporting foreign keys to reference tables
in the unified executor with a single line of change, just
calling this function.


src/backend/distributed/transaction/relation_access_tracking.c
```diff
@@ -281,11 +284,68 @@ RecordPlacementAccessToCache(Oid relationId, ShardPlacementAccessType accessType
...
 
+/*
+ * RecordParallelAccessForTaskList gets a task list and records
+ * the necessary parallel relation accesses for the task list.
+ *
+ * This function is used to enforce foreign keys from distributed
+ * tables to reference tables.
+ */
+void
+RecordParallelAccessForTaskList(List *taskList)
```
onderkalaci: As @marcocitus noted in #2692 we should change the name of the function

> I think keeping the RecordParallel -> RecordRelationParallel convention would be nice. Then it's "obvious" why we only care about the first task.

src/backend/distributed/transaction/relation_access_tracking.c
```diff
@@ -281,11 +284,68 @@ RecordPlacementAccessToCache(Oid relationId, ShardPlacementAccessType accessType
...
+	{
+		/* single shard task doesn't mean parallel access in our definition */
+		return;
+	}
+
+	firstTask = linitial(taskList);
+
+	if (firstTask->taskType == SQL_TASK)
+	{
+		RecordRelationParallelSelectAccessForTask(firstTask);
```
pykello: In the previous version of code, where this is done in `MultiRealTimeExecute()`, we loop over all tasks and call this function for each of the tasks. Is calling it only for the first task ok?

I'm assuming all tasks are on the same set of relations and this should be ok. just wanted to make sure.

src/backend/distributed/transaction/relation_access_tracking.c
```diff
@@ -281,11 +284,68 @@ RecordPlacementAccessToCache(Oid relationId, ShardPlacementAccessType accessType
...
+			 * We prefer to mark with all remaining multi-shard modifications
+			 * with both modify and select accesses.
+			 */
+			RecordRelationParallelModifyAccessForTask(firstTask);
+			RecordRelationParallelSelectAccessForTask(firstTask);
+		}
+	}
+	else
+	{
+		RecordRelationParallelDDLAccessForTask(firstTask);
```
pykello: In the previous version of the code we are skipping this for reference tables. Why are we not doing that anymore?

src/backend/distributed/transaction/relation_access_tracking.c
pykello: Some comment explaining that all tasks contains the same set of relations, so we are only processing the first task, can be useful.

src/backend/distributed/transaction/relation_access_tracking.c
onderkalaci: Yes, that's fine. In the end, we're only tracking relation accesses and each task accesses the same relations.

src/backend/distributed/transaction/relation_access_tracking.c
onderkalaci: I made some clean-up. In the previous code, we've the following comment, which is actually true:
```
		/*	
		 * Even single task DDLs hit here, so we'd prefer	
		 * not to record for reference tables.	
		 */
```

But, with this function we return early for that case `if (list_length(taskList) < 2)` which is also easier to read for me. What do you think?

Issue #1374, opened 2017-05-01
Consider making multi-tenant sharding explicit


marcocitus:
We currently require users to add a `tenant_id` column to all their tables, which significantly complicates the migration process. The client needs to specify the `tenant_id` in order to efficiently determine which shard(s) to use. However, that does not mean the tenant_id needs to be a column.

We could introduce an explicit notion of a tenant that can be set before running a query. A table could 
be distributed by tenant without having it as a column.

A possible implementation could then be to use a GUC to set the tenant.

```
CREATE TABLE hits (
  page_id int, 
  ip inet,
  hit_time timestamptz default now()
);
SELECT distribute_by_tenant('hits');

BEGIN;
SET citus.tenant TO 'citusdata.com';
INSERT INTO hits (page_id, ip) VALUES (3, '123.0.0.1');
END;
```

This means the client needs some specific logic to set the GUC at the start of the transaction, but it needs no other logic for migration or application conversion. The advantage is that the `citus.tenant` GUC can be coupled with a PostgreSQL role, which makes it mostly transparent to the application.

lfittl:
I think this is a very valuable train of thought, since application changes are currently our biggest blocker / effort required for multi-tenant use cases.

I wonder if we could do something similar for SELECTs as well, so you don't need to rewrite your queries to include the tenant_id anymore..

marcocitus:
I think once you set `citus.tenant`, all queries prune down to the shard for that tenant, so additional filters or join requirements become unnecessary. Sharing by citus.tenant is like a lightweight way of sharding by schema, without all the downsides of sharing by schema.

The only case in which we wouldn't prune based on `citus.tenant` is DDL, but that would typically be done during a migration anyway.

COPY across multiple shards gets a bit trickier. You'd either have to do separate COPY commands for each tenant, or we could allow you to specify a tenant_id column in the COPY that is only used for pruning. This seems acceptable.

INSERT..SELECT across tenants and more generally colocated multi-shard joins also get a bit trickier. We could assume that all tables are implicitly joined on tenant ID, which is simple, but changes SQL semantics. Alternatively, we could require foreign keys on join columns, such that colocation can be inferred. For pure multi-tenant use-cases, not having INSERT..SELECT or colocated multi-shard joins might also be acceptable.

aamederen:
In my opinion this would make things super easy on the user end. Instead of fixing things on the application side like multitenant gem, this could fix it once and for all. Looks super promising.

[...]

Issue #1327, opened 2017-04-17
Add Infrastructure for Syncing Configuration from Coordinator to Workers 


anarazel:
There's a fair amount of GUCs that can usefully be set on a per-session, query, user, ... basis.  Currently in Citus such GUCs will only take effect on the coordinator, because connection to workers won't sync them.  I suggest we change things so that newly established connections sync one set of GUCs, and possibly other GUCs get synced more frequently.

GUCs that are worthwhile to be synced are (please expand):

GUC | Affects Correctness | Performance | Behaviour | Debugging | Comment
----|---------------------|-------------|-----------|-----------|-------------
TimeZone			 | x |   | x |   | How timestamps are parsed
standard_conforming_string	 | x |   |   |   |
lock_timeout			 |   |   | x |   | Deadlock avoidance
synchronous_commit		 | x | x | x |   | Affects Durability, Availability & Perf
statement_timeout		 |   |   | x |   |
maintenance_work_mem		 |   | x |   |   | Index Build Speed
work_mem			 |   | x |   |   | Query Speed, Highly Workload Dependant
temp_file_limit			 |   |	 | x |   | Avoid Overload due to Bad Plan
max_parallel_workers_per_gather	 |   | x | x |	 | Control over parallelism, different use-cases
track_io_timing			 |   |	 |   | x | Useful for performance evaluation, display IO in EXPLAIN (ANALYZE, BUFFERS)
track_functions			 |   |	 |   | x |
effective_io_concurrency	 |   |   |   | x | can drastically effect query execution speed
effective_cache_size		 |   | x |   | x | can influence query plans
enable_$plantype		 |   | x |   | x | influences types of plan chosen, important for debugging presumably bad plans
cpu_*_cost			 |   | x |   | x | to influence planning / debug badplans
cursor_tuple_fraction		 |   | x |   | x | to get correct plans for cursors
from_collapse_limit		 |   | x |   | x |
from_collapse_limit		 |   | x |   | x |
min_parallel_relation_size	 |   | x |   | x | Controls when to use parallelism
parallel_*_cost			 |   | x |   | x | Controls when to use parallelism
geqo*				 |   | x |   | x |
random_page_cost		 |   | x |   | x | Controls Planning Cost
seq_page_cost			 |   | x |   | x | Controls Planning Cost

[...]

PR #2689, opened 2019-04-25
Add logic to propagate SET LOCAL at xact start

jasonmp85:
Partially addresses #1327 (which discusses _all_ config, not just `SET LOCAL`)

DESCRIPTION: Add logic to propagate SET LOCAL at xact start

# Tasks
  - [x] Add failing SQL test
  - [x] Add GUC to whitelist propagated settings (for now); do not allow PostgreSQL ones (for now)
  - [x] During transactions, detect matching `SET LOCAL` commands and…
  - [x] Push down to any already-open remote transactions, and…
  - [x] Append to a list of needed local settings for any further remote transactions, and…
  - [x] During any subsequent remote transactions, add the `SET` commands after the initial `BEGIN` (but in the same network hop)
  - [x] Show test passes

### DIFF
--- a/src/backend/distributed/commands/utility_hook.c
+++ b/src/backend/distributed/commands/utility_hook.c
@@ -56,6 +56,7 @@
 
 
 bool EnableDDLPropagation = true; /* ddl propagation is enabled */
+PropSetCmdBehavior PropagateSetCommands = PROPSETCMD_NONE; /* SET prop off */
 static bool shouldInvalidateForeignKeyGraph = false;
 static int activeAlterTables = 0;
 
@@ -175,6 +176,24 @@ multi_ProcessUtility(PlannedStmt *pstmt,
 		return;
 	}
 #endif
+
+	/* process SET LOCAL stmts of whitelisted GUCs in multi-stmt xacts */
+	if (IsA(parsetree, VariableSetStmt))
+	{
+		VariableSetStmt *setStmt = (VariableSetStmt *) parsetree;
+		bool propagateSetVar = (PropagateSetCommands == PROPSETCMD_LOCAL &&
+								setStmt->is_local);
+		bool setVarIsValid = SetCommandTargetIsValid(setStmt);
+
+		/* at present, we only implement the NONE and LOCAL behaviors */
+		AssertState(PropagateSetCommands == PROPSETCMD_NONE ||
+					PropagateSetCommands == PROPSETCMD_LOCAL);
+
+		if (propagateSetVar && setVarIsValid && IsMultiStatementTransaction())
+		{
+			ProcessVariableSetStmt(setStmt, queryString);
+		}
+	}
 
 	/*
 	 * TRANSMIT used to be separate command, but to avoid patching the grammar
--- a/src/backend/distributed/commands/variableset.c
+++ b/src/backend/distributed/commands/variableset.c
@@ -1,0 +1,133 @@
+/*-------------------------------------------------------------------------
+ *
+ * variableset.c
+ *    Support for propagation of SET (commands to set variables)
+ *
+ * Copyright (c) 2019, Citus Data, Inc.
+ *
+ *-------------------------------------------------------------------------
+ */
+
+#include "postgres.h"
+#include "c.h"
+
+#include "common/string.h"
+#include "distributed/commands.h"
+#include "distributed/commands/utility_hook.h"
+#include "distributed/metadata_cache.h"
+#include "distributed/multi_router_executor.h"
+#include "distributed/resource_lock.h"
+#include "distributed/transaction_management.h"
+#include "distributed/version_compat.h"
+#include "storage/lmgr.h"
+#include "utils/builtins.h"
+#include "utils/lsyscache.h"
+#include "lib/ilist.h"
+#include "utils/varlena.h"
+#include "distributed/remote_commands.h"
+
+
+/*
+ * Checks whether a SET command modifies a variable which might violate assumptions
+ * made by Citus. Because we generally don't want a connection to be destroyed on error,
+ * and because we eagerly ensure the stack can be fully allocated (at backend startup),
+ * permitting changes to these two variables seems unwise. Also, ban propagating the
+ * SET command propagation setting (not for correctness, more to avoid confusion).
+ */
+bool
+SetCommandTargetIsValid(VariableSetStmt *setStmt)
+{
+	/* if this list grows considerably, switch to bsearch */
+	const char *blacklist[] = {
+		"citus.propagate_set_commands",
+		"client_encoding",
+		"exit_on_error",
+		"max_stack_depth"
+	};
+	Index idx = 0;
+
+	for (idx = 0; idx < lengthof(blacklist); idx++)
+	{
+		if (pg_strcasecmp(blacklist[idx], setStmt->name))
+		{
+			return true;
+		}
+	}
+
+	return false;
+}
+
+
+/*
+ * ProcessVariableSetStmt actually does the work of propagating a provided SET stmt
+ * to currently-participating worker nodes and adding the SET command test to a string
+ * keeping track of all propagated SET commands since (sub-)xact start.
+ */
+void
+ProcessVariableSetStmt(VariableSetStmt *setStmt, const char *setStmtString)
+{
+	dlist_iter iter;
+	const bool raiseInterrupts = true;
+	List *connectionList = NIL;
+
+	/* at present we only support SET LOCAL */
+	AssertArg(setStmt->is_local);
+
+	/* haven't seen any SET stmts so far in this (sub-)xact: initialize StringInfo */
+	if (activeSetStmts == NULL)
+	{
+		MemoryContext old_context = MemoryContextSwitchTo(CurTransactionContext);
+		activeSetStmts = makeStringInfo();
+		MemoryContextSwitchTo(old_context);
+	}
+
+	/* send text of SET stmt to participating nodes... */
+	dlist_foreach(iter, &InProgressTransactions)
+	{
+		MultiConnection *connection = dlist_container(MultiConnection, transactionNode,
+													  iter.cur);
+		RemoteTransaction *transaction = NULL;
+
+		transaction = &connection->remoteTransaction;
+		if (transaction->transactionFailed)
+		{
+			continue;
+		}
+
+		if (!SendRemoteCommand(connection, setStmtString))
+		{
+			const bool raiseErrors = true;
+			HandleRemoteTransactionConnectionError(connection, raiseErrors);
+		}
+
+		connectionList = lappend(connectionList, connection);
+	}
+
+	WaitForAllConnections(connectionList, raiseInterrupts);
+
+	/* ... and wait for the results */
+	dlist_foreach(iter, &InProgressTransactions)
+	{
+		MultiConnection *connection = dlist_container(MultiConnection, transactionNode,
+													  iter.cur);
+		RemoteTransaction *transaction = NULL;
+		const bool raiseErrors = true;
+
+		transaction = &connection->remoteTransaction;
+		if (transaction->transactionFailed)
+		{
+			continue;
+		}
+
+		ClearResults(connection, raiseErrors);
+	}
+
+	/* SET propagation successful: add to active SET stmt string */
+	appendStringInfoString(activeSetStmts, setStmtString);
+
+	/* ensure semicolon on end to allow appending future SET stmts */
+	if (!pg_str_endswith(setStmtString, ";"))
+	{
+		appendStringInfoChar(activeSetStmts, ';');
+	}
+}
--- a/src/backend/distributed/shared_library_init.c
+++ b/src/backend/distributed/shared_library_init.c
@@ -21,6 +21,7 @@
 #include "executor/executor.h"
 #include "distributed/backend_data.h"
 #include "distributed/citus_nodefuncs.h"
+#include "distributed/commands.h"
 #include "distributed/commands/multi_copy.h"
 #include "distributed/commands/utility_hook.h"
 #include "distributed/connection_management.h"
@@ -59,6 +60,7 @@
 #include "tcop/tcopprot.h"
 #include "utils/guc.h"
 #include "utils/guc_tables.h"
+#include "utils/varlena.h"
 
 /* marks shared object as one loadable by the postgres version compiled against */
 PG_MODULE_MAGIC;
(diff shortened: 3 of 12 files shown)


jasonmp85:
# New Settings

FWIW I think the original suggestion we add some sort of GUC like `citus.gucs_synced_at_begin` isn't a bad one. In fact, it might not be a bad idea to straight-up only allow non-PostgreSQL GUCs in there to start (rather than delicately crafting a whitelist of what might be safe to send).

# MVP: Transaction Blocks

We'll want to keep track of any `SET LOCAL` things in the local transaction and push them down during any `BEGIN` to appropriate workers. Maybe "slow" (though I'm not actually sure of that, honestly), but correct/sufficient for a lot of stuff to become possible (including @mkurz's RLS policies).

# Additional Enhancements

OK, so in addition to explicit `SET LOCAL` commands within `BEGIN`, it appears there are two other places this could work…

## Functions

Declarations of functions can include configuration that is set/reset as the functions are entered/exited, respectively. Now that functions use transactions, propagating this should be straightforward.

## Multi-Statement Commands

This is a bit arcane, but it _is_ possible to use `SET LOCAL` without an explicit transaction block or function call. In particular, look at this…

```
psql 'SET LOCAL DateStyle TO SQL; SELECT now()' -c 'SHOW DateStyle;' postgres
┌───────────────────────────────┐
│              now              │
├───────────────────────────────┤
│ 04/24/2019 12:24:34.95311 MDT │
└───────────────────────────────┘
(1 row)

Time: 0.556 ms
┌───────────┐
│ DateStyle │
├───────────┤
│ ISO, MDY  │
└───────────┘
(1 row)

Time: 0.118 ms
```

The custom `DateStyle` appears to take affect for the `SELECT`. This means we could (in theory) preface SQL we generate (going to workers) with various `SET LOCAL` statements in order to force those single-SQL commands (not in transaction blocks) to run with the desired settings. As you can see above, the date style returns during the next transaction.

_BUT JASON ARE YOU SURE EACH `-c` COMMAND RUNS IN THE SAME CONNECTION?_

```
psql -c 'SELECT pg_backend_pid();' -c 'SELECT pg_backend_pid();' postgres
┌────────────────┐
│ pg_backend_pid │
├────────────────┤
│          16050 │
└────────────────┘
(1 row)

Time: 0.522 ms
┌────────────────┐
│ pg_backend_pid │
├────────────────┤
│          16050 │
└────────────────┘
(1 row)

Time: 0.103 ms
```

Yes, I am.

marcocitus:
> do not allow PostgreSQL ones

We should support propagating settings like enable_hashagg and work_mem. The fact that we don't propagate those is a frequent pain point.

mkurz:
So maybe a whitelist of allowed "PostgreSQL ones"?

hanefi:
> We should support propagating settings like enable_hashagg and work_mem

If we propagate changes to `work_mem` will that mean that the setting will be the same across the workers and the coordinator? AFAIK it may be better to have higher `work_mem` on the coordinator, and we should be careful in how we support this

marcocitus:
> If we propagate changes to work_mem will that mean that the setting will be the same across the workers and the coordinator? AFAIK it may be better to have higher work_mem on the coordinator, and we should be careful in how we support this

Yes, but only for the current transaction. You could use this to hint to the planner on each worker session that it should pick less memory hungry plan to prevent OOMs. work_mem does not affect the query plan on the coordinator in distributed queries.

sumedhpathak:
@jasonmp85 What is the status on this PR?

src/backend/distributed/commands/variableset.c
```diff
@@ -0,0 +1,123 @@
+/*-------------------------------------------------------------------------
+ *
+ * vacuum.c
+ *    Commands for vacuuming distributed tables.
```
marcocitus: stale comment

src/backend/distributed/commands/utility_hook.c
```diff
@@ -176,6 +176,20 @@ multi_ProcessUtility(PlannedStmt *pstmt,
...
+		VariableSetStmt *setStmt = (VariableSetStmt *) parsetree;
+
+		/* we don't process anything other than SET LOCAL */
+		/* need to be in a transaction to propagate SET LOCAL */
+		if (setStmt->is_local && IsMultiStatementTransaction())
+		{
+			ProcessVariableSetStmt(setStmt, queryString);
+		}
+
+		/* TODO: warn if not possible to propagate? */
```
marcocitus: I guess we probably shouldn't emit anything by default, since this potentially happens very frequently.

src/backend/distributed/commands/variableset.c
```diff
@@ -0,0 +1,123 @@
...
+	{
+		MultiConnection *connection = dlist_container(MultiConnection, transactionNode,
+													  iter.cur);
+		RemoteTransaction *transaction = &connection->remoteTransaction;
+		if (transaction->transactionFailed)
+		{
+			continue;
+		}
+
+		FinishRemoteTransactionVariableModify(connection);
```
marcocitus: We don't really care about the result very much except it not being an error. Could we do `ClearResults(connection, true)` here?

src/backend/distributed/commands/variableset.c
jasonmp85: Yeah, I always copy-paste these when adding a new file; fixing…

src/backend/distributed/commands/utility_hook.c
jasonmp85: I'll remove the `TODO`, but I'm _mostly_ worried about the case where a user is modifying a GUC that's part of the whitelist (admittedly not yet part of the PR when you looked at it) but _isn't_ doing so in a transaction so we don't propagate its effects. That seems like it might be counterintuitive enough for us to warn them.

src/backend/distributed/commands/variableset.c
jasonmp85: Sure, I was mostly cargo-culting another function pair here.

jasonmp85:
I left function comments for me to write tomorrow, but otherwise I've addressed your comments and added the whitelist GUC (for now I'm just parsing it out of a string on each `SET LOCAL`; that _could_ obviously be improved if you want, before merge).

What was the behavior you wanted for a whitelist: a single list with all GUCs that should be permitted, or a hardcoded one of some PostgreSQL GUCs we know can work, or…?

Let me know and I'll clean that up/document the functions/merge tomorrow, approval permitting.

mkurz:
Great to see you work on this!

I have one question: Will the example I gave in [this comment](https://github.com/citusdata/citus/issues/1327#issuecomment-384768838) work with this pull request?
I mean will the coordinator correctly figure out **to which worker send the query to when using RLS**?

E.g. when having a transaction like:
```sql
BEGIN;
SET LOCAL my_app.tenant_id=2; -- or SELECT set_config('my_app.tenant_id', '2', true);
SELECT * from events; -- fetches ONLY tenant_id with id 2 because of RLS rules
COMMIT; -- or ROLLBACK, doesn't matter for this example
```
Will the coordinator know about the RLS rules (only events with tenant_id = 2) and then, based on this restriction, send the query down to the worker that contains tenant with id 2, and then this worker will also set the GUC and apply the RLS?
I just want to make sure this works. If it does, this pull request is amazing :+1: and would also fix #1374 IMHO.

metdos:
Hi @marcocitus, is this on your plate now?

src/backend/distributed/commands/variableset.c
```diff
@@ -0,0 +1,149 @@
...
+
+	/* only perform the following if we appear to be the coordinator for a txn */
+	if (activeSetStmts == NULL)
+	{
+		MemoryContext old_context = MemoryContextSwitchTo(CurTransactionContext);
+		activeSetStmts = makeStringInfo();
+		MemoryContextSwitchTo(old_context);
+	}
+
+	/* asynchronously send SAVEPOINT */
```
marcocitus: stale comment

src/backend/distributed/commands/variableset.c
```diff
@@ -0,0 +1,149 @@
...
+
+/*
+ * TODO: document
+ */
+static void
+FinishRemoteTransactionVariableModify(MultiConnection *connection)
+{
+	const bool raiseErrors = true;
+
+	ClearResults(connection, raiseErrors);
```
marcocitus: I'd probably just inline these Start/Finish functions.

src/backend/distributed/transaction/transaction_management.c
```diff
@@ -156,6 +140,25 @@ InitializeTransactionManagement(void)
 }
 
 
+/*
+ * BeginCoordinatedTransaction begins a coordinated transaction. No
+ * pre-existing coordinated transaction may be in progress./
+ */
+static void
+BeginCoordinatedTransaction(void)
```
marcocitus: Maybe move this back to not get `git blame`d

src/backend/distributed/shared_library_init.c
```diff
@@ -1119,6 +1122,18 @@ RegisterCitusConfigVariables(void)
 		GUC_NO_SHOW_ALL,
 		NULL, NULL, NULL);
 
+	DefineCustomStringVariable(
+		"citus.set_local_whitelist",
+		gettext_noop("Sets whitelist of GUCs permitted in SET LOCAL."),
+		NULL,
+		&SetLocalWhitelistString,
+		"",
+		PGC_POSTMASTER,
```
marcocitus: Can we make this PGC_USERSET? Otherwise it isn't very useful in cloud.

src/backend/distributed/transaction/transaction_management.c
```diff
@@ -483,8 +495,12 @@ static void
 PopSubXact(SubTransactionId subId)
 {
 	MemoryContext old_context = MemoryContextSwitchTo(CurTransactionContext);
-	Assert(linitial_int(activeSubXacts) == subId);
-	activeSubXacts = list_delete_first(activeSubXacts);
+	SubXactState *state = linitial(activeSubXactStates);
+
+	Assert(state->subId == subId);
+	activeSetStmts = state->setLocalCmd;
```
marcocitus: I'm not sure I follow this logic. Should we not keep separate SET LOCAL lists for each subxact level?

src/backend/distributed/shared_library_init.c
jasonmp85: Sure; did it (pushing soon).

src/backend/distributed/transaction/transaction_management.c
jasonmp85: Done.

src/backend/distributed/commands/variableset.c
jasonmp85: Sure; Done.

src/backend/distributed/commands/variableset.c
jasonmp85: Yeah, copypasta.

src/backend/distributed/transaction/transaction_management.c
jasonmp85: Writing an explanation in code comments, since I anticipated this would be the confusing part (I somewhat deviously/lazily like to see what code reviewers are confused by before explaining things in comments to see what I need to focus on).

src/backend/distributed/transaction/transaction_management.c
jasonmp85: Actually, I'm going to leave this (though I'll keep it in a separate commit. During coding I saw this function was only called from within this file, so I moved it down with the other static functions and marked it as such.

By keeping it in a separate commit with a good message, I doubt we'll have any mistaken attribution (and more advanced git users would probably be using flags to ignore moves anyhow).

jasonmp85:
OK, @marcocitus… I've brushed up the documentation and (I think) addressed your question about the implementation. Basically:

  * During a transaction there's this `StringInfo` I append `SET LOCAL` stmts to
  * When we push into a subxact, the current value of that string is saved on the context stack and the string ref itself is reset to empty to keep track of subsequent `SET LOCAL` commands

Then:

  * Nodes already part of a transaction just get the `SET LOCAL` commands as the transaction progresses
  * Nodes coming into a transaction at a later point piggyback on the `SAVEPOINT` stacking behavior: instead of just sending them `SAVEPOINT`s, we send (pseudo-regex) `((SET LOCAL;)* SAVEPOINT)+(SET LOCAL;)*`

The `ROLLBACK` behavior of `SET LOCAL` ensures all this works in a stack-like way. The command strings live in the current transaction context so all that Just Works™, too.

**Let me know what i need to do to get this merged, i'd like to do so tomorrow and don't know what else is required now that all feedback is addressed**.

jasonmp85:
@samay-sharma: Marco wanted to know what you thought about the use of a whitelist for specifying which GUC variables are subject to propagation. Do you think this works for customers? From what I've seen, the use-cases typically have specific variables they'd like propagated, so if they can identify and whitelist them they should be OK, IMO.

jasonmp85:
> Looks fine.

Ok. But no PR approval? Ask I asked before: what’s needed for that?

> I'm wondering whether we should force router SELECTs to open a transaction block when there is a SET LOCAL.

Isn’t this scope creep? I’d really like to move on to something else unless this doesn’t address the issue it was written to address. Let me know.

jasonmp85:
@marcocitus this is ready for a final review.

@samay-sharma (I think? Maybe it was you) mentioned offhandedly that an `enum` GUC would be more forward-looking, so I went with that instead of a `bool`; it'll eventually allow settings to permit propagation of all types of `SET` commands, but for now it's just `none` or `local` (default `none`).

The only things I thought might be worth including on a blacklist of "illegal" variables to propagate were `exit_on_error` (kills the session on any error, which would be bad for assumptions around connection reuse, I suppose) and `max_stack_depth` (we read this during initialization to eagerly attempt to allocate the stack (and fail fast) rather than letting the kernel do later so as it grows (and fail in a more roundabout/obtuse way)). I also threw in the `propagate_set_commands` GUC for good measure.

If you could look over the settings to see whether _you_ think anything needs banning, that would be good (I just read over them all using e.g. `SELECT name, short_desc  FROM pg_settings WHERE context IN ('user', 'superuser');`). Thanks!

src/backend/distributed/transaction/transaction_management.c
```diff
@@ -46,8 +46,18 @@ XactModificationType XactModificationLevel = XACT_MODIFICATION_NONE;
 /* list of connections that are part of the current coordinated transaction */
 dlist_head InProgressTransactions = DLIST_STATIC_INIT(InProgressTransactions);
 
-/* stack of active sub-transactions */
-static List *activeSubXacts = NIL;
+/* string to keep track of SET LOCAL statements executed within current subxact */
+/* set to NULL when pushing into new subxact or ending top xact */
```
marcocitus: slightly unusual comment formatting

marcocitus:
Note: I pushed a commit to fix the compilation issue, maybe squash it first.

src/backend/distributed/transaction/transaction_management.c
jasonmp85: I think I added one line in one commit and the other in another; I'll combine.

jasonmp85:
> This currently gives me a compile error

I really gotta figure out what's different between dev and CI; it's odd that the CI build doesn't similarly fail…

I don't write normal `for` loops very often, but when I do I _really_ dislike not being able to restrict the loop variable's scope to, well, that of the loop. We should address that someday. But for now I'll fix it.

jasonmp85:
I think `session_preload_libraries` isn't going to do much given the user is already in a session (i.e. it's too late to change that at this point, no?)…

`application_name` is a good one. I'll add that.

PR #2692, opened 2019-04-26
Support foreign keys to reference tables in unified executor

onderkalaci:
The idea for supporting foreign keys to reference tables is simple:
Keep track of the relation accesses within a transaction block.
* If a parallel access happens on a distributed table which
          has a foreign key to a reference table, one cannot modify
          the reference table in the same transaction. Otherwise,
          we're very likely to end-up with a self-distributed deadlock.
* If an access to a reference table happens, and then a parallel
          access to a distributed table (which has a fkey to the reference
          table) happens, we switch to sequential mode.
    
Unified executor misses the function calls that marks the relation
    accesses during the execution. Thus, simply add the necessary calls
    and let the logic kick in.

This PR includes a commit from #2688 so would be good if both
PRs are reviewed together.

The PR fixes several regression tests:`foreign_key_restriction_enforcement`, `relation_access_tracking`, `foreign_key_to_reference_table`, `validate_constraint` and `multi_reference_table`.  Well, few of the tests mentioned above still fails due to very minor test output differences, which will be fixed altogether later.

### DIFF
--- a/src/backend/distributed/executor/executor.c
+++ b/src/backend/distributed/executor/executor.c
@@ -471,14 +471,14 @@ CreateDistributedExecution(DistributedPlan *distributedPlan,
 void
 StartDistributedExecution(DistributedExecution *execution)
 {
+	DistributedPlan *distributedPlan = execution->plan;
+	Job *job = distributedPlan->workerJob;
+	List *taskList = job->taskList;
+
 	if (MultiShardCommitProtocol != COMMIT_PROTOCOL_BARE)
 	{
-		DistributedPlan *distributedPlan = execution->plan;
 		if (DistributedStatementRequiresRollback(distributedPlan))
 		{
-			Job *job = distributedPlan->workerJob;
-			List *taskList = job->taskList;
-
 			BeginOrContinueCoordinatedTransaction();
 
 			if (TaskListRequires2PC(taskList))
@@ -489,6 +489,21 @@ StartDistributedExecution(DistributedExecution *execution)
 	}
 
 	execution->isTransaction = InCoordinatedTransaction();
+
+	/*
+	 * We should not record parallel access if the target pool size is less than 2.
+	 * The reason is that we define parallel access as at least two connections
+	 * accessing established to worker node.
+	 *
+	 * It is not ideal to have this check here, it'd have been better if we simply passed
+	 * DistributedExecution directly to the RecordParallelAccess*() function. However,
+	 * since we have two other executors that rely on the function, we had to only pass
+	 * the tasklist to have a common API.
+	 */
+	if (execution->targetPoolSize > 1)
+	{
+		RecordParallelRelationAccessForTaskList(taskList);
+	}
 
 	AssignTasksToConnections(execution);
 }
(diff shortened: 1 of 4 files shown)


src/backend/distributed/transaction/relation_access_tracking.c
```diff
@@ -281,11 +284,68 @@ RecordPlacementAccessToCache(Oid relationId, ShardPlacementAccessType accessType
...
+/*
+ * RecordParallelAccessForTaskList gets a task list and records
+ * the necessary parallel relation accesses for the task list.
+ *
+ * This function is used to enforce foreign keys from distributed
+ * tables to reference tables.
+ */
+void
+RecordParallelAccessForTaskList(List *taskList)
+{
```
marcocitus: I think keeping the RecordParallel -> RecordRelationParallel convention would be nice. Then it's "obvious" why we only care about the first task.

src/test/regress/expected/foreign_key_restriction_enforcement.out
```diff
@@ -546,7 +546,7 @@ ROLLBACK;
 BEGIN;
 	ALTER TABLE on_update_fkey_table ADD COLUMN X int;
 	UPDATE referece_table SET id = 160 WHERE id = 15;
-ERROR:  cannot execute SELECT on reference relation "referece_table" because there was a parallel DDL access to distributed relation "on_update_fkey_table" in the same transaction
+ERROR:  cannot execute DML on reference relation "referece_table" because there was a parallel DDL access to distributed relation "on_update_fkey_table" in the same transaction
```
onderkalaci: It looks like we've also fixed a minor bug and hence the output is more accurate now :)

src/backend/distributed/executor/executor.c
```diff
@@ -1675,6 +1687,37 @@ PlacementAccessListForTask(Task *task, ShardPlacement *taskPlacement)
 		placementAccessList = lappend(placementAccessList, placementAccess);
 	}
 
+	/*
+	 * We've already added anchor shardId's placement access to the list. Now,
+	 * add the other placements in the relationShardList.
+	 */
```
pykello: I'm a bit confused. Function comments says this function returns placement accesses for the given task and **task placement**, but here we are adding placement accesses for the other shards' placements?

Also aren't we already looping over all tasks and their placements in `AssignTasksToConnections()`, which should cover all shards? Why do we need to do that here again?

onderkalaci: >  but here we are adding placement accesses for the other shards' placements?

`anchorShardId` is the anchor one 😄  However, a task could access to many placements. For example, `SELECT FROM t1 JOIN t2 USING(id)` accesses two tables, and each task actually accesses two placements, each coming from different tables. So, here `relationShardList ` is used to keep track of these things. And, it makes sense to add all placements to the access list so that the connection/transaction mngmt keeps track of it.

(For example, if you create foreign key, we add the table that is referenced to the `relationShardList` where as the referencing one is on the anchor part.)

> Also aren't we already looping over all tasks and their placements in AssignTasksToConnections(), which should cover all shards? Why do we need to do that here again?

I think we're looping over anchor shards, but we should still loop over `relationShards`

pykello: I see. It seems that `task->relationShardList` already contains `task->anchorShardId`, so what is that `addAnchorAccess` logic in this function for? I mean the code that you added always seem to add `anchorShardId` access, so we can remove the `if (addAnchorAccess)` part?

pykello: Ok, it contains it for computeTasks, but not for DDL or DML tasks.

onderkalaci: There might be some code paths which skips building `relationShardList ` because it is a single table - single shard access. For example, do we set it for single row inserts? I need to do some code reading to make sure that it is OK.

pykello: > There might be some code paths which skips building relationShardList  because it is a single table - single shard access. For example, do we set it for single row inserts? I need to do some code reading to make sure that it is OK.

It seems that removing `if (addAnchorAccess)` is not Ok, I removed it and tests failed. But this function is overall a bit not easy to understand.

I am still trying to make sure I understand every part of it and then make suggestions, if I have.

src/backend/distributed/executor/executor.c
```diff
@@ -1664,9 +1679,6 @@ PlacementAccessListForTask(Task *task, ShardPlacement *taskPlacement)
 		accessType = PLACEMENT_ACCESS_SELECT;
```
pykello: What is an example of `SELECT` queries that does not touch any shard placements?

src/backend/distributed/executor/executor.c
onderkalaci: I think non-foreign-key DDLs do not set `relationShardList `

src/backend/distributed/executor/executor.c
onderkalaci: Reading from intermediate results could be an example:

```SQL
SELECT * FROM (SELECT * FROM users_table OFFSET 0) as foo LIMIT 1;
```

src/backend/distributed/executor/executor.c
onderkalaci: > But this function is overall a bit not easy to understand.

For example, there is no `else if DML ` and we always use `BuildPlacementSelectList ()`. THat's because there cannot be a (distributed) query which modifies multiple tables. Even if a say INSERT .. SELECT, only one table is modified, the others are read. I'll update the comment for this.

onderkalaci: > I'll update the comment for this.

Ah, sorry. I already have a comment for it:

```
		/*
		 * In case of SELECTs or DML's, we add SELECT placement accesses to the
		 * elements in relationShardList. For SELECT queries, it is trivial, since
		 * the query is literally accesses the relationShardList in the same query.
		 *
		 * For DMLs, create placement accesses for placements that appear in a
		 * subselect.
```

mkurz:
Can #968 be closed now that this pull request got merged?

PR #2694, opened 2019-04-28
Improve closing connections that are not fully established in unified executor

onderkalaci:
This PR is against `unified_executor`.

It looks like we're calling wrong function to close the connections that have not been fully established. `ShutdownConnection ()` keeps the `MultiConnection` around with a closed connection, which ends up with `connection errors`, especially with recursively planned queries (or sometimes transaction blocks). The reason is that `ConnectionHash` would end-up with connections that are closed, and the next executions try to use those connections.

This doesn't implement what I had in mind where we'd have the options to forcefully close (like this) or wait until the connection is fully established. 

Without this fix, we see lots of intermittent/random connection errors, I'd propose to merge this and implement the actual solution later (and tick the item in #2665 later as well). 

(Note that we have one more use of `ShutdownConnection `, which looks fine given that we call it only when the transaction fails as we typically call `ShutdownConnection`)

### DIFF
--- a/src/backend/distributed/executor/executor.c
+++ b/src/backend/distributed/executor/executor.c
@@ -1256,6 +1256,17 @@ ConnectionStateMachine(WorkerSession *session)
 
 				/* remove the connection */
 				UnclaimConnection(connection);
+
+				/*
+				 * We forcefully close the underlying libpq connection because
+				 * we don't want any subsequent execution (either subPlan executions
+				 * or new command executions within a transaction block) use the
+				 * connection.
+				 *
+				 * However, we prefer to keep the MultiConnection around until
+				 * the end of FinishDistributedExecution() to simplify the code.
+				 * Thus, we prefer ShutdownConnection() over CloseConnection().
+				 */
 				ShutdownConnection(connection);
 
 				/* remove connection from wait event set */
--- a/src/backend/distributed/executor/executor.c
+++ b/src/backend/distributed/executor/executor.c
@@ -560,8 +560,13 @@ FinishDistributedExecution(DistributedExecution *execution)
 
 		UnclaimConnection(connection);
 
-		if (connection->connectionState == MULTI_CONNECTION_CONNECTING)
+		if (connection->connectionState == MULTI_CONNECTION_CONNECTING ||
+			connection->connectionState == MULTI_CONNECTION_FAILED)
 		{
+			/*
+			 * We want the MultiConnection go away and not used in
+			 * the subsequent executions.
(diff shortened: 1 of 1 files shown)


src/backend/distributed/executor/executor.c
```diff
@@ -562,7 +562,7 @@ FinishDistributedExecution(DistributedExecution *execution)
 
 		if (connection->connectionState == MULTI_CONNECTION_CONNECTING)
 		{
-			ShutdownConnection(connection);
```
marcocitus: ShutdownConnection is used in a few other places. Are those usages also incorrect?

marcocitus:
> (Note that we have one more use of ShutdownConnection , which looks fine given that we call it only when the transaction fails as we typically call ShutdownConnection)

Is that true for savepoints?

onderkalaci:
> Is that true for savepoints?

Good question, I'm not sure about that, I'll look into that.

onderkalaci:
> Is that true for savepoints?

We use `ShutdownConnection()` in 4 places other than the change above: 

- `StartRemoteTransactionAbort()` and  `FinishRemoteTransactionAbort()`: When there is a savepoint in the transaction, failures wouldn't trigger calling `CoordinatedRemoteTransactionsAbort()`. In other words, to get here, the user either has to issue `ROLLBACK` or the worker query should fail when there is no proceeding savepoint(s).  So, we should be fine for these calls.
- `AfterXactHostConnectionHandling()`: Similar to the above, we only call this at the end of transactions on non-session lifespan connection. So, should be fine here as well.
- `ConnectionStateMachine()`: This is interesting. It looks like we're actually doing the correct thing. We call `ShutdownConnection()` on session failures because we expect the `MultiConnection` to be around during the execution. However, we probably wouldn't want the `multiConnection` in the subsequent executions. Please see my latest commit on this PR as well. 

After getting check-multi pass, I'll add (or ask someone else to add) lost of failure tests. Hopefully, I'll trigger each and every failure scenarios. Still, I think merging this unblocks us because otherwise we'd have lots of random failures.

marcocitus:
>  This is interesting. It looks like we're actually doing the correct thing. We call ShutdownConnection() on session failures because we expect the MultiConnection to be around during the execution. However, we probably wouldn't want the multiConnection in the subsequent executions. Please see my latest commit on this PR as well.

Could we add some comments to this extent?

onderkalaci:
> Could we add some comments to this call?

Is that better now? I think writing failure tests would be fun for this executor :)

PR #2697, opened 2019-05-02
Force a connection per placement for coordinator INSERT .. SELECT with modifying CTEs

onderkalaci:
This PR is against unified_executor branch.

Fixes quite a few queries in `with_modifying.sql` and some others
as well.

The new executor can re-use the same connection for modifying
multiple placements. However, INSERT .. SELECT via coordinator,
which internally implements COPY, is not capable of doing
that (yet).

When modifying CTEs are followed by a coordinator INSERT .. SELECT,
it is very likely to error out. Thus, we detect those cases and
force one connection per placement for the modifying CTEs.

I think we should discuss this along with #2687 because that could also
work for this case, but, probably a lot slower. So, we may or may not
decide to merge this PR.

I think @pykello will be the primary reviewer, but, still getting @marcocitus's
feedback regarding this vs #2687 would be nice.

### DIFF
--- a/src/backend/distributed/executor/insert_select_executor.c
+++ b/src/backend/distributed/executor/insert_select_executor.c
@@ -36,6 +36,8 @@
 #include "utils/snapmgr.h"
 
 
+static char * GetForceMaxQueryParallelization(void);
+static void SetLocalMaxQueryParallelization(char *value);
 static void ExecuteSelectIntoRelation(Oid targetRelationId, List *insertTargetList,
 									  Query *selectQuery, EState *executorState);
 static HTAB * ExecuteSelectIntoColocatedIntermediateResults(Oid targetRelationId,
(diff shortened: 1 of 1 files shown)


pykello:
I am a bit hesitant about merging this, since although it fixes one problem, but might surface some other problems. For example, if a modifying CTE is followed by a DDL/DML in a transaction.

I think we should fix INSERT ... INTO SELECT before merging unified_executor, otherwise this, #2702, and probably some other regression tests will need fixes which although fix some problem, but just create some other problems.

What do you think?

marcocitus:
I think if we can cleanly merge #2702 to master without significant extra effort that seems like a good approach. Note that on the master branch it wouldn't have any benefit if the INSERT..SELECT part goes through the multi-shard executor, although we could do sequential execution to at least be able to test the behavior.

onderkalaci:
I agree with your comments. The main motivation for me to open this PR is to get all the regression tests pass -- with less `citus.force_max_query_parallelization` in the tests -- since we've initially considered #2702 as a change on top of the new executor on the master branch. Since we've decided to prioritize that change for 8.3, I'm happy to close this in favor of #2702

PR #2700, opened 2019-05-06
Rebuild WaitEventSet only when connection set changes.

pykello:
We should rebuild the wait event set only when the set of connections or the events we are interested in changes.

### DIFF
--- a/src/backend/distributed/executor/executor.c
+++ b/src/backend/distributed/executor/executor.c
@@ -57,10 +57,21 @@ typedef struct DistributedExecution
 	List *sessionList;
 
 	/*
-	 * Flag to indiciate that the number of connections has changed
-	 * and the WaitEventSet needs to be rebuilt.
+	 * Flag to indiciate that the set of connections we are interested
+	 * in has changed and waitEventSet needs to be rebuilt.
 	 */
-	bool rebuildWaitEventSet;
+	bool connectionSetChanged;
+
+	/*
+	 * Flag to indiciate that the set of wait events we are interested
+	 * in might have changed and waitEventSet needs to be updated.
+	 * 
+	 * Note that we set this flag whenever we assign a value to waitFlags,
+	 * but we don't check that the waitFlags is actually different from the
+	 * previous value. So we might have some false positives for this flag,
+	 * which is OK, because in this case ModifyWaitEvent() is noop.
+	 */
+	bool waitFlagsChanged;
 
 	/*
 	 * The number of connections we aim to open per worker.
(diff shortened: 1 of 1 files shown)


src/backend/distributed/executor/executor.c
```diff
@@ -600,6 +604,12 @@ FinishDistributedExecution(DistributedExecution *execution)
 		/* prevent copying shards in same transaction */
 		XactModificationLevel = XACT_MODIFICATION_DATA;
 	}
+
+	if (execution->waitEventSet != NULL)
+	{
+		FreeWaitEventSet(execution->waitEventSet);
```
marcocitus: Would it be sufficient to do this at the end of `RunDistributedExecution`. If so, can `waitEventSet` remain a local variable?

src/backend/distributed/executor/executor.c
```diff
@@ -1211,9 +1231,11 @@ ConnectionStateMachine(WorkerSession *session)
 
 					connection->waitFlags = WL_SOCKET_WRITEABLE;
 					connection->connectionState = MULTI_CONNECTION_CONNECTED;
-					break;
 				}
 
+				/* connection->waitFlags has changed, we need to rebuild the WaitEventSet */
+				execution->rebuildWaitEventSet = true;
```
marcocitus: As long as we're not adding or removing (in case of a failure) a connection, we could use `ModifyWaitEvent` and avoid rebuilding the `WaitEventSet`

pykello: done

src/backend/distributed/executor/executor.c
pykello: done

src/backend/distributed/executor/executor.c
```diff
@@ -57,10 +57,16 @@ typedef struct DistributedExecution
...
+	 * in has changed and waitEventSet needs to be rebuilt.
 	 */
-	bool rebuildWaitEventSet;
+	bool connectionSetChanged;
+
+	/*
+	 * Flag to indiciate that the set of wait events we are interested
+	 * in has changed and waitEventSet needs to be updated.
+	 */
+	bool waitFlagsChanged;
```
marcocitus: I think we set this flag somewhat liberally, since we don't check whether the flags actually changed, but that seems mostly fine because `ModifyWaitEvent` is a noop when the flags don't change. Maybe we can add a comment to that effect somewhere?

pykello: done. update the comments above `waitFlagsChanged`.

src/backend/distributed/executor/executor.c
```diff
@@ -1063,17 +1088,14 @@ RunDistributedExecution(DistributedExecution *execution)
...
-		}
-
+		FreeWaitEventSet(waitEventSet);
 		pfree(events);
 
 		PG_RE_THROW();
 	}
 	PG_END_TRY();
+
+	FreeWaitEventSet(waitEventSet);
```
marcocitus: this could maybe just be after `pfree(events);`

pykello: done

PR #2701, opened 2019-05-06
Fix mixed declarations and code warnings

pykello:
This fixes some build fails.


jasonmp85:
So… these aren't failing in CircleCI? That's odd… I thought our `CFLAGS` were set up to flag these.

PR #2702, opened 2019-05-07
Use ExecuteTaskList() for CreateShardsOnWorkers()

pykello:
This will only work after COPY goes through unified executor, but I thought it can be reviewed individually.


pykello:
This requires multiple copy commands be able to share one connection, so it is blocked on #2687.

src/backend/distributed/master/master_stage_protocol.c
```diff
@@ -556,41 +528,35 @@ CreateShardsOnWorkers(Oid distributedRelationId, List *shardPlacements,
...
 
-		if (useExclusiveConnection)
+		commands = WorkerCreateShardCommandList(distributedRelationId, shardIndex,
+												shardId, ddlCommandList,
+												foreignConstraintCommandList);
+		foreach(commandCell, commands)
 		{
-			ClaimConnectionExclusively(connection);
-			claimedConnectionList = lappend(claimedConnectionList, connection);
+			const char *command = lfirst(commandCell);
```
onderkalaci: Hmm, I'm still thinking on this. We probably never have other places with multiple commands in the `task->queryString`.  I'm trying to understand the implications of this.

I think it is also not a good idea to make each command as a separate task because there is no guarantee of execution of the tasks. So we might try to create fkey before the shard is created etc.

We're implicitly always going from 0th to last element in the task list. But still sounds dangerous. 

In sequential execution PR, @marcocitus mentioned about creating dependencies among the tasks. Still, that probably require a big change.

Any other alternatives can we think of?

marcocitus: I think concatenating query strings is acceptable and will help improve performance. The TransactionStateMachine will happily loop around in REMOTE_TRANSACTION_CLEARING_RESULTS after sending it. 

There are other cases in the Citus code base where we concatenate commands. The task-tracker executor does it a lot.

onderkalaci: Thanks for the note! Then I'll stick to this solution

marcocitus: We also do this in the unified executor already whenever we send BEGIN.

src/backend/distributed/master/master_stage_protocol.c
```diff
@@ -556,41 +528,35 @@ CreateShardsOnWorkers(Oid distributedRelationId, List *shardPlacements,
...
-	 * We need to unclaim all connections to make them usable again for the copy
-	 * command, otherwise copy going to open new connections to placements and
-	 * can not see uncommitted changes.
-	 */
-	foreach(connectionCell, claimedConnectionList)
-	{
-		MultiConnection *connection = (MultiConnection *) lfirst(connectionCell);
-		UnclaimConnection(connection);
-	}
+	poolSize = InCoordinatedTransaction() ? DEFAULT_POOL_SIZE : 1;
```
marcocitus: I think we could just as well let it use multiple connections.

marcocitus: oh, maybe not because of the long name issues.

pykello:
Closing, because we have https://github.com/citusdata/citus/pull/2770 now.

PR #2703, opened 2019-05-09
Fix DistShardCacheHash initialization

mtuncer:
DESCRIPTION: Fix a crash that can occur under high memory load

InitializeCaches() method may prematurely set performedInitialization without actually creating DistShardCacheHash . Fix makes sure flag is set only if DistShardCacheHash is created successfully.

Also introduced a new memory context to allocate aforementioned hash tables. If allocation/initialization fails for any reason we make sure memory is reclaimed by deleting the memory context.

### DIFF
--- a/src/backend/distributed/utils/metadata_cache.c
+++ b/src/backend/distributed/utils/metadata_cache.c
@@ -60,6 +60,7 @@
 #include "utils/inval.h"
 #include "utils/fmgroids.h"
 #include "utils/lsyscache.h"
+#include "utils/memutils.h"
 #include "utils/palloc.h"
 #include "utils/rel.h"
 #include "utils/relfilenodemap.h"
@@ -155,6 +156,7 @@ static HTAB *DistTableCacheHash = NULL;
 
 /* Hash table for informations about each shard */
 static HTAB *DistShardCacheHash = NULL;
+static MemoryContext MetadataCacheMemoryContext = NULL;
 
 /* Hash table for informations about worker nodes */
 static HTAB *WorkerNodeHash = NULL;
@@ -927,7 +929,7 @@ BuildDistTableCacheEntry(DistTableCacheEntry *cacheEntry)
 	{
 		Node *partitionNode = NULL;
 
-		oldContext = MemoryContextSwitchTo(CacheMemoryContext);
+		oldContext = MemoryContextSwitchTo(MetadataCacheMemoryContext);
 
 		/* get the string representation of the partition column Var */
 		cacheEntry->partitionKeyString = TextDatumGetCString(partitionKeyDatum);
@@ -979,10 +981,11 @@ BuildDistTableCacheEntry(DistTableCacheEntry *cacheEntry)
 		typeEntry = lookup_type_cache(partitionColumn->vartype,
 									  TYPECACHE_HASH_PROC_FINFO);
 
-		hashFunction = MemoryContextAllocZero(CacheMemoryContext,
+		hashFunction = MemoryContextAllocZero(MetadataCacheMemoryContext,
 											  sizeof(FmgrInfo));
 
-		fmgr_info_copy(hashFunction, &(typeEntry->hash_proc_finfo), CacheMemoryContext);
+		fmgr_info_copy(hashFunction, &(typeEntry->hash_proc_finfo),
+					   MetadataCacheMemoryContext);
 
 		cacheEntry->hashFunction = hashFunction;
 
@@ -996,7 +999,7 @@ BuildDistTableCacheEntry(DistTableCacheEntry *cacheEntry)
 		cacheEntry->hashFunction = NULL;
 	}
 
-	oldContext = MemoryContextSwitchTo(CacheMemoryContext);
+	oldContext = MemoryContextSwitchTo(MetadataCacheMemoryContext);
 
 	cacheEntry->referencedRelationsViaForeignKey = ReferencedRelationIdList(
 		cacheEntry->relationId);
@@ -1044,16 +1047,16 @@ BuildCachedShardList(DistTableCacheEntry *cacheEntry)
 		ListCell *distShardTupleCell = NULL;
 		int arrayIndex = 0;
 
-		shardIntervalArray = MemoryContextAllocZero(CacheMemoryContext,
+		shardIntervalArray = MemoryContextAllocZero(MetadataCacheMemoryContext,
 													shardIntervalArrayLength *
 													sizeof(ShardInterval *));
 
 		cacheEntry->arrayOfPlacementArrays =
-			MemoryContextAllocZero(CacheMemoryContext,
+			MemoryContextAllocZero(MetadataCacheMemoryContext,
 								   shardIntervalArrayLength *
 								   sizeof(GroupShardPlacement *));
 		cacheEntry->arrayOfPlacementArrayLengths =
-			MemoryContextAllocZero(CacheMemoryContext,
+			MemoryContextAllocZero(MetadataCacheMemoryContext,
 								   shardIntervalArrayLength *
 								   sizeof(int));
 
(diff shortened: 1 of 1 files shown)


src/backend/distributed/utils/metadata_cache.c
```diff
@@ -2621,6 +2630,9 @@ InitializeDistTableCache(void)
 {
 	HASHCTL info;
 
+	/* make sure DistShardCacheHash was not initialized before */
+	Assert(DistShardCacheHash == NULL);
```
marcocitus: I think this can happen in case of OOM.

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -2598,19 +2598,28 @@ InitializeCaches(void)
...
 			CreateCacheMemoryContext();
 		}
 
 		InitializeDistTableCache();
+
+		/*
+		 * Make sure DistShardCacheHash is created successfully before setting
+		 * performedInitialization flag and registering callbacks.
+		 */
+		if (DistShardCacheHash != NULL)
```
marcocitus: I don't think this can really happen. What could happen is that we throw an OOM before we get to this point that leaves the cache in a partially initialised state. Moving the boolean assignment should cause us to reinitialise.

What we might also consider doing is cleaning up the memory when the cache is partially initialised before re-initialising it, e.g. we could use a separate memory context (child of CacheMemoryContext) and reset that in a PG_CATCH. That might be overkill since this should be a very rare event.

src/backend/distributed/utils/metadata_cache.c
mtuncer: I don't follow how this could happen. This function is the only one making memory allocation. Once DistShardCacheHash is set, we only make few more calls that sets part of previously allocated memory.

If we are really hitting OOM due to hash_create calls within this function DistShardCacheHash should remain unset.

Could you explain more on how it could be not null ?

src/backend/distributed/utils/metadata_cache.c
mtuncer: One way partial initialization can happen is DistTableCacheHash could be created successfully [here](https://github.com/citusdata/citus/blob/master/src/backend/distributed/utils/metadata_cache.c#L2646) but following [call](https://github.com/citusdata/citus/blob/master/src/backend/distributed/utils/metadata_cache.c#L2653) could fail.

I still think DistShardCacheHash remains unset here.

marcocitus: If the call fails I would expect an error to be thrown, hence we don't continue with this function.

src/backend/distributed/utils/metadata_cache.c
marcocitus: An error might be thrown after DistShardCacheHash is set, but before initialisation is complete. In that case, we'll do initialisation again.

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -2598,19 +2598,28 @@ InitializeCaches(void)
...
+			InitializeDistTableCache();
+			RegisterForeignKeyGraphCacheCallbacks();
+			RegisterWorkerNodeCacheCallbacks();
+			RegisterLocalGroupIdCacheCallbacks();
+		}
+		PG_CATCH();
 		{
-			CreateCacheMemoryContext();
+			performedInitialization = false;
+			PG_RE_THROW();
```
marcocitus: I'm not sure this PG_CATCH adds anything over moving the assignment to the end of the function such that it is only reached when we manage to complete all function calls above.

What we could use a PG_CATCH for is to reset the cache if we fail to initialise it.

src/backend/distributed/utils/metadata_cache.c
mtuncer: done

src/backend/distributed/utils/metadata_cache.c
mtuncer: put a try catch block outside, and did a memory reset if something bad happens

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -2598,26 +2599,47 @@ InitializeCaches(void)
 
 	if (!performedInitialization)
 	{
-		/* set first, to avoid recursion dangers */
-		performedInitialization = true;
+		MemoryContext shardCacheMemoryContext = NULL;
 
-		/* make sure we've initialized CacheMemoryContext */
-		if (CacheMemoryContext == NULL)
+		PG_TRY();
```
marcocitus: Would be good to add a comment on why we need the PG_TRY

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -2636,7 +2658,7 @@ InitializeDistTableCache(void)
 
 	fmgr_info_cxt(F_OIDEQ,
 				  &DistShardScanKey[0].sk_func,
-				  CacheMemoryContext);
+				  shardCacheMemoryContext);
```
marcocitus: To make deleting the memory context effective, we should probably also use shardCacheMemoryContext for the hash tables (currently they use `TopMemoryContext`, which is the default). Now it only allocates these two functions in the memory context.

marcocitus:
Verified that throwing an error in `InitializeDistTableCache` causes a crash and that this fixes the crash.

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -155,6 +156,7 @@ static HTAB *DistTableCacheHash = NULL;
 
 /* Hash table for informations about each shard */
 static HTAB *DistShardCacheHash = NULL;
+static MemoryContext ShardCacheMemoryContext = NULL;
```
marcocitus: Maybe `MetadataCacheMemoryContext`? the hash table used to cache distributed table information is also allocated in this memory context.

Optional, but it would be good to use this in place of CacheMemoryContext whenever we allocate something into the cache.

src/backend/distributed/utils/metadata_cache.c
mtuncer: done

src/backend/distributed/utils/metadata_cache.c
mtuncer: done

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -1720,7 +1723,7 @@ InstalledExtensionVersion(void)
 		}
 
 		/* we will cache the result of citus version to prevent catalog access */
-		oldMemoryContext = MemoryContextSwitchTo(CacheMemoryContext);
+		oldMemoryContext = MemoryContextSwitchTo(MetadataCacheMemoryContext);
```
marcocitus: We might be using this one before `MetadataCacheMemoryContext` is initialised because it is called from CitusHasBeenLoaded().

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -1654,7 +1657,7 @@ AvailableExtensionVersion(void)
 			Datum availableVersion = slot_getattr(tupleTableSlot, 2, &isNull);
 
 			/* we will cache the result of citus version to prevent catalog access */
-			oldMemoryContext = MemoryContextSwitchTo(CacheMemoryContext);
+			oldMemoryContext = MemoryContextSwitchTo(MetadataCacheMemoryContext);
```
marcocitus: We might be using this one before `MetadataCacheMemoryContext` is initialised because it is called from CitusHasBeenLoaded().

src/backend/distributed/utils/metadata_cache.c
marcocitus: Upon further inspection, we always call `InitializeCaches()` in `AvailableExtensionVersion()`, which always gets called before this function. That's way too subtle though, we should at least add a call to `InitializeCaches()`

src/backend/distributed/utils/metadata_cache.c
mtuncer: InitializeCaches() is called earlier in the function

src/backend/distributed/utils/metadata_cache.c
mtuncer: right, added InitializeCaches() call

PR #2707, opened 2019-05-14
Prevent anchoring reference table shard ids when distributed tables are in join clauses

### LINKED ISSUE #2680: Round robin with a distributed sub query bug
Hi, 

When activating the round robin task assignment policy, this query will result in unexpected behavior.
```sql
select * from (select * from ditributed_table where distributed_key = $1 ) as dist
left join reference_table as ref  on dist.ref_id = ref.id
where 1=1
```
It seems that for some reason citus sends the query to a wrong node and this warning is thrown in the master:
`WARNING:  relation "public.ditributed_table_102542" does not exist`

In one of the workers, we will find the following error:

```sql
ERROR:  relation "public.ditributed_table_102542" does not exist at character XXX
STATEMENT:  
SELECT *
 FROM ((SELECT * 
 	FROM public.ditributed_table_102542 dist1 
 	WHERE (dist1.distributed_key OPERATOR(pg_catalog.=) $1)) dist 
 LEFT JOIN public.reference_table_102446 ref ON ((dist.ref_id OPERATOR(pg_catalog.=) ref.id)))
  WHERE (1 OPERATOR(pg_catalog.=) 1)
```

Fortunately, we found that using a CTE query will solve the problem but we need to know if this is expected behavior or it is a possible bug.

Thanks,

hanefi:
DESCRIPTION: Fixes a bug that selects wrong worker when using round-robin assignment

When using the round-robin task assignment policy, citus needs to find the worker that stores all the shards that are accessed in a query/subquery. 
- If the query accesses shards of reference tables only, it can be run on any workers that have the reference table.
- If there is a distributed table shard in the `FROM` clause, it is the deciding factor in which worker Citus should execute this query.

fixes #2680

### DIFF
--- a/src/test/regress/expected/multi_task_assignment_policy.out
+++ b/src/test/regress/expected/multi_task_assignment_policy.out
@@ -184,7 +184,7 @@ RESET client_min_messages;
 -- which might change and we don't have any control over it.
 -- the important thing that we look for is that round-robin policy 
 -- should give the same output for executions in the same transaction 
--- and different output for executions that are not insdie the
+-- and different output for executions that are not inside the
 -- same transaction. To ensure that, we define a helper function
 BEGIN;
 SET LOCAL citus.explain_distributed_queries TO on;
@@ -201,12 +201,11 @@ SELECT count(DISTINCT value) FROM explain_outputs;
      1
 (1 row)
 
-DROP TABLE explain_outputs;
+TRUNCATE explain_outputs;
 COMMIT;
 -- now test round-robin policy outside
--- a transaction, we should see the assignements
+-- a transaction, we should see the assignments
 -- change on every execution
-CREATE TEMPORARY TABLE explain_outputs (value text);
 SET citus.task_assignment_policy TO 'round-robin';
 SET citus.explain_distributed_queries TO ON;
 INSERT INTO explain_outputs 
@@ -248,12 +247,11 @@ SELECT count(DISTINCT value) FROM explain_outputs;
      1
 (1 row)
 
-DROP TABLE explain_outputs;
+TRUNCATE explain_outputs;
 COMMIT;
 -- now test round-robin policy outside
--- a transaction, we should see the assignements
+-- a transaction, we should see the assignments
 -- change on every execution
-CREATE TEMPORARY TABLE explain_outputs (value text);
 SET citus.task_assignment_policy TO 'round-robin';
 SET citus.explain_distributed_queries TO ON;
 INSERT INTO explain_outputs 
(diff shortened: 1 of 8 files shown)


src/test/regress/expected/multi_task_assignment_policy.out
```diff
@@ -268,6 +266,45 @@ SELECT count(DISTINCT value) FROM explain_outputs;
...
+EXPLAIN SELECT *
+FROM (SELECT * FROM task_assignment_nonreplicated_hash WHERE test_id = 3) AS dist
+       LEFT JOIN task_assignment_reference_table ref
+                 ON dist.ref_id = ref.test_id
+$cmd$, 'task_assignment_nonreplicated_hash');
+INSERT INTO explain_outputs
+SELECT parse_explain_output($cmd$
+EXPLAIN SELECT *
+FROM (SELECT * FROM task_assignment_nonreplicated_hash WHERE test_id = 3) AS dist
+       LEFT JOIN task_assignment_reference_table ref
```
marcocitus: Did you verify whether this test fails before the change? Wondering under what conditions the reference table appears first in the relation shard list.

hanefi: Yes, I verified that it fails on master branch. Checking it once again. The second `SELECT` fails with these extra outputs:
```
WARNING:  relation "public.task_assignment_nonreplicated_hash_880006" does not exist
PL/pgSQL function parse_explain_output(text,text) line 6 at FOR over EXECUTE statement
```

It should have hit the worker running on port 57638

src/test/regress/sql/multi_task_assignment_policy.sql
```diff
@@ -212,7 +206,35 @@ INSERT INTO explain_outputs
 -- given that we're in the same transaction, the count should be 2
 -- since there are two different worker nodes
 SELECT count(DISTINCT value) FROM explain_outputs;
+TRUNCATE explain_outputs;
 
+-- test that the round robin policy detects the anchor shard correctly
```
marcocitus: Might be worth adding something like "we should not pick a reference table shard as the anchor shard when joining with a distributed table".

hanefi: I added this explanation and revised my other comments in this test file. Thanks for the advice

hanefi:
I wanted to squash all my commits into one before merging. I guess I picked one extra commit (that was already merged into master) and thus the Github UI seems really messy now.

`git diff HEAD~1` shows only my changes, but not the Github UI: https://github.com/citusdata/citus/pull/2707/files

PR #2708, opened 2019-05-15
Copy goes through executor.

pykello:


src/backend/distributed/executor/executor.c
```diff
@@ -1679,7 +1688,14 @@ StartPlacementExecutionOnSession(TaskPlacementExecution *placementExecution,
 	/* connection is going to be in use */
 	workerPool->idleConnectionCount--;
 
-	if (paramListInfo != NULL)
+
+	if (task->copyTask)
+	{
+		uint64 shardId = taskPlacement->shardId;
+		ExecuteCopyTask(connection, task->copyReceiver, shardId, task->copyData);
```
onderkalaci: I think doing it prevents any parallelism across task executions. Shouldn't we call different pieces of the function on different parts of `TransactionStateMachine`?

For example, `ConstructCopyStatement` might be good in `StartPlacementExecutionOnSession`. Later, on `REMOTE_TRANS_SENT_COMMAND`, do `SendCopyDataToPlacement` and on `REMOTE_TRANS_CLEARING_RESULTS` do `PutRemoteCopyEnd` etc?

Because, we're currently only utilizing 1 connection per node. If you set the debug level to `DEBUG4`, you'd see something like:
```
DEBUG:  Total number of commands sent over the session 253: 10
DEBUG:  Total number of commands sent over the session 254: 0
DEBUG:  Total number of commands sent over the session 255: 0
DEBUG:  Total number of commands sent over the session 256: 0
DEBUG:  Total number of commands sent over the session 257: 11
DEBUG:  Total number of commands sent over the session 258: 0
DEBUG:  Total number of commands sent over the session 259: 0
DEBUG:  Total number of commands sent over the session 260: 0
DEBUG:  Total number of commands sent over the session 261: 11
DEBUG:  Total number of commands sent over the session 262: 0
DEBUG:  Total number of commands sent over the session 263: 0
DEBUG:  Total number of commands sent over the session 264: 0
```

pykello: Thanks. I agree, I will address this.

marcocitus:
This change is making me a little nervous in terms of adding a lot more complexity (and development time) to both the COPY implementation and the unified executor at a time when we're already making the biggest-ever change to Citus internals. Onder's comments made me realise doing a proper non-blocking (parallelised) implementation of COPY is tricky because you have to poll all the sockets, send bytes on the ones that are writable, consume notices, handle failures, plus you want to be as fast as possible, and not have bugs.

The existing COPY handles this primarily by buffering the data in libpq when we're sending data faster than the socket can handle and blocking when a buffer gets too big or the COPY is finishing. Instead of polling all the sockets continuously, we wait until we have a row to send to a shard before we consume and flush bytes on the socket. This is not perfect, but relatively simple and performant because libpq handles a lot of the complexity. It has also worked well for many large production deployments, though it took a while to get the kinks out.

What if we hold off on this change and focus on other items for merging the unified executor for the moment?

marcocitus:
Discussing with Onder a little more, I realise holding off on COPY is not really an option. 

Let's think about the "right" design a bit more. I might be overestimating how complex it would be to integrate the asynchronous sending/receiving into the executor. I am also wondering whether we can do something better/smarter if we don't try to fit into the executor. Maybe we can dispatch copy tasks in the background?

Some possible designs:

1. Collect rows in "copy tasks". Interrupt reading rows once task reaches 8MB and execute all tasks greater than 4MB via the unified executor. Inherits asynchronous connection establishment from unified executor, but requires executor changes and there is no parallelism between coordinator and workers.
2. Collect rows in "copy tasks" and execute them in the background once they reach 4MB (?) and block if they reach 8MB. Possibly more complex to implement, but preserves parallelism between processing rows on the coordinator and processing rows on the worker.
3. Multiplex copy into multiple shards over a single connection, e.g. `COPY "multiplexer" FROM STDIN WITH (multiplex = '[{"min":"-120986568", "max":"-1","destination":"test_102008"},]')`. Quite optimal, but requires a huge amount of extra code to handle the writes on the workers.
4. Copy into intermediate results and finish with an INSERT..SELECT via the unified executor. We need to do this anyway for INSERT..SELECT..ON CONFLICT and it's relatively simple to implement, but also gives up parallelism between coordinator and workers. The workers only do meaningful work once the cooridnator is done.

pykello:
Closing this, because https://github.com/citusdata/citus/pull/2758 is the solution we have consensus on.

PR #2709, opened 2019-05-15
Getting a diff generated

jasonmp85:


metdos:
Do we want this in 8.3?

jasonmp85:
> Do we want this in 8.3?

Given this is mostly testing code, it won't matter whether it's in the release; it can always be backported into the release branfhes.

marcocitus:
I assume this is abandoned.

PR #2710, opened 2019-05-15
Propagate more ALTER FOREIGN TABLE to workers, fixes #2548

### LINKED ISSUE #2548: ALTER FOREIGN TABLE … OWNER TO isn't propagated to workers.
`ALTER TABLE ... OWNER TO ...` is propagated to workers, but the foreign table equivalent (`ALTER FOREIGN TABLE ... OWNER TO ...` isn't propagated to workers. This makes it difficult to change the owner of distributed cstore tables.

serprex:
Propagate ALTER FOREIGN TABLE commands to workers

### DIFF
--- a/src/backend/distributed/commands/table.c
+++ b/src/backend/distributed/commands/table.c
@@ -524,12 +524,14 @@ IsAlterTableRenameStmt(RenameStmt *renameStmt)
 {
 	bool isAlterTableRenameStmt = false;
 
-	if (renameStmt->renameType == OBJECT_TABLE)
+	if (renameStmt->renameType == OBJECT_TABLE ||
+		renameStmt->renameType == OBJECT_FOREIGN_TABLE)
 	{
 		isAlterTableRenameStmt = true;
 	}
 	else if (renameStmt->renameType == OBJECT_COLUMN &&
-			 renameStmt->relationType == OBJECT_TABLE)
+		(renameStmt->relationType == OBJECT_TABLE ||
+			renameStmt->relationType == OBJECT_FOREIGN_TABLE))
 	{
 		isAlterTableRenameStmt = true;
 	}
--- a/src/test/regress/base_schedule
+++ b/src/test/regress/base_schedule
@@ -5,3 +5,4 @@ test: multi_cluster_management
 test: multi_test_helpers
 test: multi_create_table multi_behavioral_analytics_create_table
 test: multi_load_data
+test: multi_create_fdw
--- a/src/test/regress/sql/multi_generate_ddl_commands.sql
+++ b/src/test/regress/sql/multi_generate_ddl_commands.sql
@@ -106,15 +106,18 @@ CREATE FOREIGN TABLE foreign_table (
 	full_name text not null default ''
 ) SERVER fake_fdw_server OPTIONS (encoding 'utf-8', compression 'true');
 
-SELECT master_get_table_ddl_events('foreign_table');
-
 SELECT create_distributed_table('foreign_table', 'id');
 ALTER FOREIGN TABLE foreign_table rename to renamed_foreign_table;
 ALTER FOREIGN TABLE renamed_foreign_table rename full_name to rename_name;
 ALTER FOREIGN TABLE renamed_foreign_table alter rename_name type char(8);
 \c - - - :worker_1_port
-\d renamed_foreign_table_*
+select table_name, column_name, data_type
+from information_schema.columns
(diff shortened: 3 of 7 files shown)


serprex:
This PR is still WIP. Currently using a custom testing harness I hacked together over past couple days; need to create tests using existing testing infrastructure. Also my tests covered 4 cases: renaming column, renaming table, changing column type, dropping table. The last was already working. Whereas #2548 explicitly mentions propagating altering owner, & I haven't actually tested that these changes cover that alter variant

serprex:
Tests added. Spoke with @pykello & will not be adding an `OWNER TO` test

marcocitus:
With DDL commands we have to be careful that we also apply them when (re)creating shards (e.g. when moving shard). 

It might be worth taking a look at `TableOwnerResetCommand`, though some of the code paths for foreign tables might already be broken.

serprex:
Looked for where shard movement occurs, found `master_move_shard_placement` stub. In citus-enterprise it errors out on foreign tables

src/test/regress/sql/multi_generate_ddl_commands.sql
```diff
@@ -108,14 +108,22 @@ CREATE FOREIGN TABLE foreign_table (
 
 SELECT master_get_table_ddl_events('foreign_table');
 
+SELECT create_distributed_table('foreign_table', 'id');
+ALTER FOREIGN TABLE foreign_table rename to renamed_foreign_table;
+ALTER FOREIGN TABLE renamed_foreign_table rename full_name to rename_name;
+ALTER FOREIGN TABLE renamed_foreign_table alter rename_name type char(8);
+\c - - - :worker_1_port
+\d renamed_foreign_table_*
```
marcocitus: We (now) try to avoid \d in regression tests because its output tends to change heavily across major releases.

serprex: I've updated the test to select from `information_schema.columns`

serprex:
Ran a test. Postgres is smart enough to not require `FOREIGN` keyword when using `ALTER TABLE tbl OWNER TO user`

src/test/regress/base_schedule
```diff
@@ -5,3 +5,4 @@ test: multi_cluster_management
 test: multi_test_helpers
 test: multi_create_table multi_behavioral_analytics_create_table
 test: multi_load_data
+test: multi_create_fdw
```
marcocitus: can this be parallelised with another test? (tests on the same line run in parallel)

marcocitus:
Seems like a reasonable set of tests and fixes.

Note that we usually rebase onto master rather than merging master (e.g. `git rebase origin/master`)

src/test/regress/base_schedule
serprex: `multi_create_fdw` starts with `SET citus.next_shard_id TO 390000;` but otherwise shouldn't conflict with anything

pykello:
@jasonmp85 do you know why tests for this PR are not running automatically?

pykello:
@serprex can you make this a branch of citus repo so tests would run?

src/backend/distributed/commands/table.c
```diff
@@ -524,12 +524,14 @@ IsAlterTableRenameStmt(RenameStmt *renameStmt)
...
-	if (renameStmt->renameType == OBJECT_TABLE)
+	if (renameStmt->renameType == OBJECT_TABLE ||
+		renameStmt->renameType == OBJECT_FOREIGN_TABLE)
 	{
 		isAlterTableRenameStmt = true;
 	}
 	else if (renameStmt->renameType == OBJECT_COLUMN &&
-			 renameStmt->relationType == OBJECT_TABLE)
+		(renameStmt->relationType == OBJECT_TABLE ||
+			renameStmt->relationType == OBJECT_FOREIGN_TABLE))
```
pykello: When you move this repo to a branch of citus, check-style test will fail. You need to run `make reindent` to reformat the changes. To be able to run that, you need to install https://github.com/citusdata/tools.

jasonmp85:
> do you know why tests for this PR are not running automatically? Is it because this is a PR from a clone rather than a branch?

Yeah it's definitely that it's from a fork; builds (Travis before and now CircleCI) can have secrets in them and there are risks of users opening a fork and exfiltrating those secrets; I probably have fork builds turned off out of caution for CircleCI, but I don't think it's super critical they be disabled any longer; I'll make a note to look into it tomorrow.

serprex:
Duplicated by #2726

PR #2713, opened 2019-05-17
Improve failure handling on unified executor

onderkalaci:
This is a draft PR against unified_executor, I wanted to share my proposal here. @marcocitus and @pykello please let me know what you think about it. I want to add/adjust lots of tests, but before going after that I wanted to hear your thoughts in the mean time.

With this PR, we're defining the failure behavior of the new executor for DML and DDLs.

|  Failure Case										| Reference Table  | Distributed Table with replication == 1 | Distributed Table with replication > 1 | Applies for|
| ------------------------------------------- 		| ------------- | ------------- |------------- | ------ |
| Failure during connection establishment    		| Distributed query fails  | Distributed query fails  | Mark the placements what are assigned to the session (e.g., the session) with state INVALID**   | DML/DDL*|
| Failure during shard query execution one of the replicas  | Distributed query fails  | Distributed query fails  | Distributed query fails   |  DML/DDL/SELECT|
| Failure on a worker session where at least one successful shard query execution happened | Distributed query fails | Distributed query fails| Distributed query fails|DML/DDL/SELECT|

\*  `SELECT`s retry on a second placement
** We could even improve this to make it only happen on worker pools. But, here we prefer simplicity since replication > 1 is not very common use-case anymore. 

*** I also considered VACUUM/VACUUM ANALYZE as DDL in the above table since it was the case in the existing executors. 

The above table differs from the existing executors. The main difference is that the new executor throws an error on any failures during query execution. For example, if a long running SELECT fails, the new executor throws an error whereas the real-time executor tries the next replica and successfully finishes the query. Another difference is that, the new executor throws an error if a DML/DDL query in one of the shard replicas fails whereas the router executor marks the placement as INVALID. 

In summary, the new executor marks placements INACTIVE only when connection establishment for DML/DDL fails on a replicated distributed table (e.g., replication factor > 1). The design decision for this is two folded (a) It is very unlikely that a query fails on one of the placements of the same shard during the query execution (b) This implementation is a lot simpler

### DIFF
--- a/src/backend/distributed/executor/executor.c
+++ b/src/backend/distributed/executor/executor.c
@@ -29,6 +29,7 @@
 #include "distributed/remote_commands.h"
 #include "distributed/resource_lock.h"
 #include "distributed/subplan_execution.h"
+#include "distributed/transaction_management.h"
 #include "distributed/worker_protocol.h"
 #include "distributed/version_compat.h"
 #include "lib/ilist.h"
@@ -70,7 +71,7 @@ typedef struct DistributedExecution
 	/*
 	 * Flag to indiciate that the set of wait events we are interested
 	 * in might have changed and waitEventSet needs to be updated.
-	 * 
+	 *
 	 * Note that we set this flag whenever we assign a value to waitFlags,
 	 * but we don't check that the waitFlags is actually different from the
 	 * previous value. So we might have some false positives for this flag,
@@ -118,6 +119,9 @@ typedef struct DistributedExecution
 
 	/* indicates whether distributed execution has failed */
 	bool failed;
+
+	/* set to true when we prefer to bail out early */
+	bool errorOnAnyFailure;
 
 	/*
 	 * For SELECT commands or INSERT/UPDATE/DELETE commands with RETURNING,
@@ -264,6 +268,8 @@ typedef struct ShardCommandExecution
 	 * after we got results from the first placements.
 	 */
 	bool gotResults;
+
+	TaskExecutionState executionState;
 } ShardCommandExecution;
 
 /*
@@ -323,7 +329,7 @@ bool ForceMaxQueryParallelization = false;
 
 /* local functions */
 static DistributedExecution * CreateDistributedExecution(DistributedPlan *distributedPlan,
-						   								 ParamListInfo paramListInfo,
+														 ParamListInfo paramListInfo,
 														 TupleDesc tupleDescriptor,
 														 Tuplestorestate *tupleStore,
 														 int targetPoolSize);
@@ -331,6 +337,7 @@ static void StartDistributedExecution(DistributedExecution *execution);
 static void RunDistributedExecution(DistributedExecution *execution);
 static void FinishDistributedExecution(DistributedExecution *execution);
 
+static bool DistributedPlanModifiesDatabase(DistributedPlan *plan);
 static bool DistributedStatementRequiresRollback(DistributedPlan *distributedPlan);
 static void AssignTasksToConnections(DistributedExecution *execution);
 static void UnclaimAllSessionConnections(List *sessionList);
@@ -352,6 +359,8 @@ static bool StartPlacementExecutionOnSession(TaskPlacementExecution *placementEx
 											 WorkerSession *session);
 static List * PlacementAccessListForTask(Task *task, ShardPlacement *taskPlacement);
 static void ConnectionStateMachine(WorkerSession *session);
+static void Activate2PCIfModifyingTransactionExpandsToNewNode(WorkerSession *session);
+static bool TransactionModifiedDistributedTable(DistributedExecution *execution);
 static void TransactionStateMachine(WorkerSession *session);
 static bool CheckConnectionReady(MultiConnection *connection);
 static bool ReceiveResults(WorkerSession *session, bool storeRows);
@@ -359,9 +368,12 @@ static void WorkerSessionFailed(WorkerSession *session);
 static void WorkerPoolFailed(WorkerPool *workerPool);
 static void PlacementExecutionDone(TaskPlacementExecution *placementExecution,
 								   bool succeeded);
+static void ScheduleNextPlacementExecution(TaskPlacementExecution *placementExecution,
+										  bool succeeded);
+static bool ShouldMarkPlacementsInvalidOnFailure(DistributedExecution *execution);
 static void PlacementExecutionReady(TaskPlacementExecution *placementExecution);
-static TaskExecutionState GetTaskExecutionState(
-	ShardCommandExecution *shardCommandExecution);
+static TaskExecutionState TaskExecutionStateMachine(ShardCommandExecution *
+													shardCommandExecution);
 
 
 /*
@@ -411,7 +423,8 @@ CitusExecScan(CustomScanState *node)
 		}
 
 		execution = CreateDistributedExecution(distributedPlan, paramListInfo,
-											   tupleDescriptor, tupleStore, targetPoolSize);
+											   tupleDescriptor, tupleStore,
+											   targetPoolSize);
 
 		StartDistributedExecution(execution);
 		RunDistributedExecution(execution);
@@ -461,7 +474,7 @@ ExecuteTaskList(CmdType operation, List *taskList, int targetPoolSize)
  */
 uint64
 ExecuteTaskListExtended(CmdType operation, List *taskList,
-						TupleDesc tupleDescriptor, Tuplestorestate *tupleStore, 
+						TupleDesc tupleDescriptor, Tuplestorestate *tupleStore,
 						bool hasReturning, int targetPoolSize)
 {
 	DistributedPlan *distributedPlan = NULL;
@@ -549,9 +562,35 @@ StartDistributedExecution(DistributedExecution *execution)
 
 			if (TaskListRequires2PC(taskList))
 			{
+				/*
+				 * Although using two phase commit protocol is an independent decision than
+				 * failing on any error, we prefer to couple them. Our motivation is that
+				 * the failures are rare, and we prefer to avoid marking placements invalid
+				 * in case of failures.
+				 */
 				CoordinatedTransactionUse2PC();
+
+				execution->errorOnAnyFailure = true;
+			}
+			else if (MultiShardCommitProtocol != COMMIT_PROTOCOL_2PC &&
+					 list_length(taskList) > 1 &&
+					 DistributedPlanModifiesDatabase(distributedPlan))
+			{
+				/*
+				 * Even if we're not using 2PC, we prefer to error out
+				 * on any failures during multi shard modifications/DDLs.
+				 */
+				execution->errorOnAnyFailure = true;
 			}
 		}
+	}
+	else
+	{
+		/*
+		 * We prefer to error on any failures for CREATE INDEX
+		 * CONCURRENTLY or VACUUM//VACUUM ANALYZE (e.g., COMMIT_PROTOCOL_BARE).
+		 */
+		execution->errorOnAnyFailure = true;
 	}
 
 	execution->isTransaction = InCoordinatedTransaction();
@@ -572,6 +611,39 @@ StartDistributedExecution(DistributedExecution *execution)
 	}
 
 	AssignTasksToConnections(execution);
+}
+
+
+/*
+ *  DistributedPlanModifiesDatabase returns true if the plan modifies the data
+ *  or the schema.
+ */
+static bool
+DistributedPlanModifiesDatabase(DistributedPlan *plan)
+{
+	CmdType operation = plan->operation;
+	List *taskList = NIL;
+	Task *firstTask = NULL;
+
+	if (operation == CMD_INSERT || operation == CMD_UPDATE || operation == CMD_DELETE)
+	{
+		return true;
+	}
+
+	/*
+	 * If we cannot decide by only checking the operation, we should look closer
+	 * to the tasks.
+	 */
+	taskList = plan->workerJob->taskList;
+	if (list_length(taskList) < 1)
+	{
+		/* does this ever possible? */
+		return false;
+	}
+
+	firstTask = (Task *) linitial(taskList);
+
+	return !ReadOnlyTask(firstTask->taskType);
 }
 
 
@@ -597,7 +669,7 @@ DistributedStatementRequiresRollback(DistributedPlan *distributedPlan)
 		return false;
 	}
 
-	if (IsTransactionBlock())
+	if (IsMultiStatementTransaction())
 	{
 		return true;
 	}
@@ -607,10 +679,33 @@ DistributedStatementRequiresRollback(DistributedPlan *distributedPlan)
 		return true;
 	}
 
+	/*
+	 * Checking the first task's placement list is not sufficient for all purposes since
+	 * for append/range distributed tables we might have unequal number of placements for
+	 * shards. However, it is safe to do here, because we're searching for a reference
+	 * table. All other cases return false for this purpose.
+	 */
 	task = (Task *) linitial(taskList);
 	if (list_length(task->taskPlacementList) > 1)
 	{
-		return true;
+		/*
+		 * Some tasks don't set replicationModel thus we only
+		 * rely on the anchorShardId, not replicationModel.
+		 *
+		 * TODO: Do we ever need replicationModel in the Task structure?
+		 * Can't we always rely on anchorShardId?
+		 */
+		uint64 anchorShardId = task->anchorShardId;
+		if (anchorShardId != INVALID_SHARD_ID && ReferenceTableShardId(anchorShardId))
+		{
+			return true;
+		}
+
+		/*
+		 * Single DML/DDL tasks with replicated tables (non-reference)
+		 * should not require BEGIN/COMMIT/ROLLBACK.
+		 */
+		return false;
 	}
 
 	return false;
@@ -686,6 +781,7 @@ UnclaimAllSessionConnections(List *sessionList)
 	}
 }
 
+
 /*
  * AssignTasksToConnections goes through the list of tasks to determine whether any
  * task placements need to be assigned to particular connections because of preceding
@@ -719,12 +815,14 @@ AssignTasksToConnections(DistributedExecution *execution)
 		shardCommandExecution->task = task;
 		shardCommandExecution->job = job;
 		shardCommandExecution->executionOrder = ExecutionOrderForTask(operation, task);
+		shardCommandExecution->executionState = TASK_EXECUTION_NOT_FINISHED;
 		shardCommandExecution->placementExecutions =
 			(TaskPlacementExecution **) palloc0(placementExecutionCount *
 												sizeof(TaskPlacementExecution *));
 		shardCommandExecution->placementExecutionCount = placementExecutionCount;
 		shardCommandExecution->expectResults = distributedPlan->hasReturning ||
 											   distributedPlan->operation == CMD_SELECT;
+
 
 		foreach(taskPlacementCell, task->taskPlacementList)
 		{
@@ -821,6 +919,12 @@ AssignTasksToConnections(DistributedExecution *execution)
 				/* always poll the connection in the first round */
 				connection->waitFlags = WL_SOCKET_READABLE | WL_SOCKET_WRITEABLE;
 				execution->waitFlagsChanged = true;
+
+				/*
+				 * If the connections are already avaliable, make sure to activate
+				 * 2PC when necessary.
+				  */
+				Activate2PCIfModifyingTransactionExpandsToNewNode(session);
 			}
 			else
 			{
@@ -1044,8 +1148,7 @@ RunDistributedExecution(DistributedExecution *execution)
 	int maxWaitEventCount = execution->totalTaskCount * workerCount + 2;
 
 	/* allocate events for the maximum number of connections to avoid realloc */
-	WaitEvent *events = palloc0(execution->totalTaskCount * workerCount *
-								sizeof(WaitEvent));
+	WaitEvent *events = palloc0(maxWaitEventCount * sizeof(WaitEvent));
 
 	PG_TRY();
 	{
@@ -1081,7 +1184,7 @@ RunDistributedExecution(DistributedExecution *execution)
 			}
 
 			/* we should always have more (or equal) waitEvents */
-			Assert (maxWaitEventCount >= connectionCount + 2);
+			Assert(maxWaitEventCount >= connectionCount + 2);
 
 			/* wait for I/O events */
 #if (PG_VERSION_NUM >= 100000)
@@ -1172,10 +1275,10 @@ ManageWorkerPool(WorkerPool *workerPool)
 	int connectionIndex = 0;
 
 	/* we should always have more (or equal) active connections than idle connections */
-	Assert (activeConnectionCount >= idleConnectionCount);
+	Assert(activeConnectionCount >= idleConnectionCount);
 
 	/* we should never have less than 0 connections ever */
-	Assert (activeConnectionCount >= 0 && idleConnectionCount >= 0);
+	Assert(activeConnectionCount >= 0 && idleConnectionCount >= 0);
 
 	if (failedConnectionCount >= 1)
 	{
@@ -1363,7 +1466,7 @@ ConnectionStateMachine(WorkerSession *session)
 					WorkerPoolFailed(workerPool);
 				}
 
-				if (execution->failed)
+				if (execution->failed || execution->errorOnAnyFailure)
 				{
 					/* a task has failed due to this connection failure */
 					ReportConnectionError(connection, ERROR);
@@ -1391,6 +1494,18 @@ ConnectionStateMachine(WorkerSession *session)
 
 				/* remove connection from wait event set */
 				execution->connectionSetChanged = true;
+
+				/*
+				 * Reset the transaction state machine since CloseConnection()
+				 * relies on it and even if we're not inside a distributed transaction
+				 * we set the transaction state (e.g., REMOTE_TRANS_SENT_COMMAND).
+				 */
+				if (!connection->remoteTransaction.beginSent)
+				{
+					connection->remoteTransaction.transactionState =
+						REMOTE_TRANS_INVALID;
+				}
+
 				break;
 			}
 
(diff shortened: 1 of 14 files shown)


src/backend/distributed/executor/executor.c
```diff
@@ -499,7 +503,15 @@ StartDistributedExecution(DistributedExecution *execution)
 
 			if (TaskListRequires2PC(taskList))
 			{
+				/*
```
onderkalaci: FYI: This logic comes from [here](https://github.com/citusdata/citus/blob/febe41210879fd601da52ac4bf7a2178a789d531/src/backend/distributed/executor/multi_router_executor.c#L755-L766)

marcocitus:
I think it makes sense and seems a lot cleaner than today's approach setting the shard state. We mostly need to be careful about what happens if we do more commands in the same transaction block if we continue after a failure.

onderkalaci:
> I think it makes sense and seems a lot cleaner than today's approach setting the shard state. We mostly need to be careful about what happens if we do more commands in the same transaction block if we continue after a failure.

The following currently crashes, and I'm actually working on that before updating the PR to non-draft PR. The main issue is that `PlacementExecutionDone` does not seem to be idempotent which puts the same placement execution to the ready queue twice [here](https://github.com/citusdata/citus/blob/unified_executor/src/backend/distributed/executor/executor.c#L1295-L1307)

```SQL
BEGIN;
INSERT replicated_table (1);

-- terminate first worker

-- do one more insert, update the first placement to INVALID
INSERT replicated_table (1);

-- this crashes
INSERT replicated_table (1);
```

onderkalaci:
I've fixed the issues around you've raised (and I had already been working on those)

> We mostly need to be careful about what happens if we do more commands in the same transaction block if we continue after a failure.

Also, we've fixed 3 more tests with this PR: `multi_modifications`, `multi_modifying_xacts` and `multi_router_planner`.

Finally,

> I think it makes sense and seems a lot cleaner than today's approach setting the shard state.

As far as I can see, we'll always have to keep the current code that sets the shard state in the transaction callbacks as well. The reason is that we might have failures on commit time, where the executor doesn't have control over. We even have a test case for that as [here](https://github.com/citusdata/citus/blob/master/src/test/regress/sql/multi_modifying_xacts.sql#L488-L500). Still, the current code covers quite a lot of cases and easier to follow. Any comments on that?

onderkalaci:
I think this PR is ready for review now.

With @marcocitus's suggestion, I've decided to look into failure tests as well within the scope of the PR. The PR considers various failure cases in the `check-failure` regression test. Note that this doesn't mean that those tests pass after this PR, they fail because we need to update various error messages in the outputs and adjust `SESSION_LIFESPAN` connections (e.g., sometimes do not have any cached connections, sometimes have 1 cached connection since we [explicitly](https://github.com/citusdata/citus/blob/cd00e92cbc01003229d1679a9ee4f3f1af4f8154/src/test/regress/sql/failure_single_select.sql#L60) test those). So, I'm now thinking that #2716 would help us to get the failure regression tests pass and we should merge that. 

As far as I can see, all the failure cases are covered, except for one: [Failures](https://github.com/citusdata/citus/blob/cd00e92cbc01003229d1679a9ee4f3f1af4f8154/src/backend/distributed/commands/utility_hook.c#L603) of `CREATE INDEX CONCURRENTLY` which I prefer to handle separately in a different PR since it looks like a different code-path and this PR is already big enough.

src/backend/distributed/executor/executor.c
```diff
@@ -499,7 +505,29 @@ StartDistributedExecution(DistributedExecution *execution)
...
 				CoordinatedTransactionUse2PC();
+
+				execution->errorOnAnyFailure = true;
+			}
+			else if (MultiShardCommitProtocol != COMMIT_PROTOCOL_2PC &&
+					 list_length(taskList) > 1)
+			{
+				Task *initialTask = (Task *) linitial(taskList);
+
+				if (ModifyTask(initialTask->taskType))
```
marcocitus: Might be cleaner to use `IsModifyDistributedPlan`

src/backend/distributed/executor/executor.c
```diff
@@ -560,7 +588,24 @@ DistributedStatementRequiresRollback(DistributedPlan *distributedPlan)
...
+		if (anchorShardId != INVALID_SHARD_ID && ReferenceTableShardId(anchorShardId))
+		{
+			return true;
+		}
+
+		/*
+		 * Single DML/DDL tasks with replicated tables (non-reference)
+		 * should not require BEGIN/COMMIT/ROLLBACK.
+		 */
+		return false;
```
marcocitus: I'm really tempted to change this, but will suppress my desire to change too many things at once. :)

src/backend/distributed/executor/executor.c
```diff
@@ -1346,6 +1407,40 @@ ConnectionStateMachine(WorkerSession *session)
...
+static void
+Use2PCIfParticipatingNodesExpanded(WorkerSession *session)
+{
+	DistributedExecution *execution = NULL;
+
+	if (DoesCoordinatedTransactionUse2PC() ||
+		MultiShardCommitProtocol != COMMIT_PROTOCOL_2PC)
+	{
+		/* already using or doesn't require 2PC */
+		return;
```
marcocitus: Given how fast the operation below is, I don't think it's really worth having this code block and the extra function.

src/backend/distributed/executor/executor.c
```diff
@@ -1346,6 +1407,40 @@ ConnectionStateMachine(WorkerSession *session)
...
+	 * If we're expanding the set nodes that participate in the distributed
+	 * transaction, conform to MultiShardCommitProtocol.
+	 */
+	execution = session->workerPool->distributedExecution;
+	if (execution->isTransaction &&
+		XactModificationLevel == XACT_MODIFICATION_DATA)
+	{
+		MultiConnection *connection = session->connection;
+		RemoteTransaction *transaction = &connection->remoteTransaction;
+		if (transaction->transactionState == REMOTE_TRANS_INVALID)
```
marcocitus: we may need to be careful here because we've started opening transaction blocks for SELECTs in a transaction block, so something like `BEGIN; SELECT * FROM test WHERE x = 2; INSERT INTO test VALUES (1,2); INSERT INTO test VALUES (2,1);` might not activate 2PC because the `transactionState` is not invalid (did not verify yet).

src/backend/distributed/executor/executor.c
```diff
@@ -1346,6 +1407,40 @@ ConnectionStateMachine(WorkerSession *session)
 }
 
 
+/*
+ * Use2PCIfParticipatingNodesExpanded ensures that the execution uses
+ * 2PC if the transaction is expanded to access a new worker node.
+ */
+static void
+Use2PCIfParticipatingNodesExpanded(WorkerSession *session)
```
marcocitus: We're doing a bit of important magic in this function. I think it would help readability/naming if we split this function up and replace the call to `Use2PCIfParticipatingNodesExpanded` with something very explicit:

```c
if (IsModifyDistributedPlan(execution->distributedPlan) && TransactionModifiedDistributedTable() && 
    !ConnectionModifiedDistributedTable(session->connection))
{
    /* 
     * We already did a modification, but not on the connection that we just opened, which means we're
     * now going to make modifications over multiple connections. Activate 2PC! 
     */
    CoordinatedTransactionUse2PC();
}
```

src/backend/distributed/executor/executor.c
```diff
@@ -1267,6 +1312,9 @@ ConnectionStateMachine(WorkerSession *session)
 
 			case MULTI_CONNECTION_CONNECTED:
 			{
+				/* if we're expanding the nodes in a transaction, use 2PC */
+				Use2PCIfParticipatingNodesExpanded(session);
```
marcocitus: Should this be here, or later when we start using the connection?

src/backend/distributed/executor/executor.c
```diff
@@ -1334,6 +1382,19 @@ ConnectionStateMachine(WorkerSession *session)
 
 				/* remove connection from wait event set */
 				execution->connectionSetChanged = true;
+
+				/*
+				 * Reset the transaction state machine since CloseConnection()
+				 * relies on it and even if we're not inside a distributed transaction
+				 * we set the transaction state (e.g., REMOTE_TRANS_SENT_COMMAND).
+				 */
+				if (!execution->isTransaction)
```
marcocitus: This doesn't feel like the right solution. I guess we've overloaded the meaning of transactionState a bit. Perhaps the right solution would be to have a commandState. A less wrong solution would be to have a flag on the RemoteTransaction indicating whether or not we sent BEGIN.

src/backend/distributed/executor/executor.c
```diff
@@ -2022,6 +2117,20 @@ PlacementExecutionDone(TaskPlacementExecution *placementExecution, bool succeede
...
 	{
+		if (ShouldMarkPlacementsInvalidOnFailure(execution))
+		{
+			ShardPlacement *shardPlacement = placementExecution->shardPlacement;
+
+			/*
+			 * We only set shard state if its current state is FILE_FINALIZED, which
+			 * prevents overwriting shard state if it is already set at somewhere else.
+			 */
+			if (shardPlacement->shardState == FILE_FINALIZED)
```
marcocitus: Minor unrelated note: it would be nice to rename FILE_FINALIZED and FILE_INACTIVE to something more  accurate (e.g. PLACEMENT_ACTIVE and PLACEMENT_INACTIVE). Probably not worth doing right now, but on a lazy day.

src/backend/distributed/executor/executor.c
marcocitus: That function is probably not quite accurate because it does not consider DDL. Anyway, I'm guilty of using `initialTask` as well, but it's a bit of an anti-pattern.

src/backend/distributed/executor/executor.c
```diff
@@ -2086,6 +2195,30 @@ PlacementExecutionDone(TaskPlacementExecution *placementExecution, bool succeede
...
+ * ShouldMarkPlacementsInvalidOnFailures returns true if the failure
+ * should trigger marking placements invalid.
+ */
+static bool
+ShouldMarkPlacementsInvalidOnFailure(DistributedExecution *execution)
+{
+	DistributedPlan *distributedPlan = execution->plan;
+	CmdType operation = distributedPlan->operation;
+
+	if (operation == CMD_SELECT || execution->errorOnAnyFailure)
```
marcocitus: It would be nice if we could generalise this a bit. I could see us wanting to push down router queries with modifying CTEs in the near future, which are technically SELECTs. We also don't want to make placements as inactive when ANALYZE fails.

In general, it seems we need more a reliable version of `IsModifyDistributedPlan` that can distinguish which plans modify the database.

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -841,6 +842,23 @@ TaskListRequires2PC(List *taskList)
...
 
+/*
+ * ModifyTask returns true if the input task type modifies the
+ * database such as DML, DDL and Vacuum Analyze.
+ */
+bool
+ModifyTask(TaskType taskType)
+{
+	if (taskType == MODIFY_TASK || taskType == DDL_TASK ||
+		taskType == VACUUM_ANALYZE_TASK)
```
marcocitus: Probably `VACUUM_ANALYZE_TASK` is not really a modify task.

src/test/regress/expected/multi_modifying_xacts.out
```diff
@@ -532,27 +531,25 @@ DEFERRABLE INITIALLY IMMEDIATE
...
 \set VERBOSITY terse
 BEGIN;
 INSERT INTO objects VALUES (1, 'apple');
 INSERT INTO objects VALUES (2, 'BAD');
-WARNING:  illegal value
-INSERT INTO labs VALUES (7, 'E Corp');
+ERROR:  illegal value
 COMMIT;
--- data should be persisted
+-- so the data should noy be persisted
```
marcocitus: noy -> not

src/test/regress/expected/multi_modifying_xacts.out
```diff
@@ -684,9 +672,9 @@ INSERT INTO labs VALUES (8, 'Aperture Science');
 INSERT INTO labs VALUES (9, 'BAD');
 COMMIT;
 WARNING:  illegal value
-WARNING:  failed to commit transaction on localhost:57637
+WARNING:  failed to commit critical transaction on localhost:57637, metadata is likely out of sync
 WARNING:  illegal value
-WARNING:  failed to commit transaction on localhost:57638
+WARNING:  failed to commit critical transaction on localhost:57638, metadata is likely out of sync
```
marcocitus: unrelated note: this "metadata is likely out of sync"  warning is really vague

src/backend/distributed/executor/executor.c
onderkalaci: As far as I can see, it doesn't make any difference.

If we've opened a connection which expands the participating nodes, even if we don't use this specific connection, we have to activate 2PC because the command/execution needs that anyway. 

Since `Use2PCIfParticipatingNodesExpanded ` (or anything equivalent) is idempotent, it seems OK to have it here (because it seems easier to read code in this way instead of moving it to `TransactionStateMachine`?).

src/backend/distributed/executor/executor.c
onderkalaci: > This doesn't feel like the right solution. 

Yeah, I was expecting to get feedback to this change :)  But, are you OK with `connection->remoteTransaction.transactionState = REMOTE_TRANS_INVALID;` after we *somehow* decide to set the transaction state?

> Perhaps the right solution would be to have a commandState.

Do you mean not having `REMOTE_TRANS_SENT_BEGIN, REMOTE_TRANS_SENT_COMMAND, REMOTE_TRANS_FETCHING_RESULTS` and `REMOTE_TRANS_CLEARING_RESULTS` and instead have a `CommandState` in `RemoteTransaction`? Or, keep those states and have `commandState` in `WorkerSession`? The former seems a big change to me.

src/backend/distributed/executor/executor.c
onderkalaci: I haven't verified that as well, but as far as I can see, we're resetting the transaction state after successful command executions [here](https://github.com/citusdata/citus/compare/unified_executor#diff-4b24978830320f290897555d61ec2af3R1462)

src/backend/distributed/executor/executor.c
onderkalaci: > Do you mean not having REMOTE_TRANS_SENT_BEGIN, REMOTE_TRANS_SENT_COMMAND, REMOTE_TRANS_FETCHING_RESULTS and REMOTE_TRANS_CLEARING_RESULTS and instead have a CommandState in RemoteTransaction? 

I assume you pointed me to above.

I can see that `CommandState ` approach would make sense. But, that'd make the changes in this PR pretty big. So, for now, I'm adding the `beginSent` flag, which I'll revisit in a separate on top of this to implement `commandState`

src/backend/distributed/executor/executor.c
onderkalaci: Ok, I see what you mean now. So, with `ConnectionModifiedPlacement` we should be fine?

src/backend/distributed/executor/executor.c
onderkalaci: Btw, I pushed one more commit, which moved this logic into `TransactionStateMachine`. I think we can revert if you don't like this way.

src/backend/distributed/executor/executor.c
```diff
@@ -572,7 +642,24 @@ DistributedStatementRequiresRollback(DistributedPlan *distributedPlan)
 	task = (Task *) linitial(taskList);
 	if (list_length(task->taskPlacementList) > 1)
```
pykello: I think it is possible for a relation to have shards with non-equal replication factors, e.g. if we create shards with replication factor 2, then decomission a node, then some shards will have replication factor 2 and some other replication factor 1. In that, is checking the number of placements of one task enough?

src/backend/distributed/executor/executor.c
```diff
@@ -1286,6 +1373,20 @@ ConnectionStateMachine(WorkerSession *session)
...
+					TransactionModifiedDistributedTable(execution) &&
+					DistributedPlanModifiesDatabase(execution->plan) &&
+					!ConnectionModifiedPlacement(session->connection))
+				{
+					/*
+					 * We already did a modification, but not on the connection that we
+					 * just opened, which means we're now going to make modifications
+					 * over multiple connections. Activate 2PC!
+					 */
+					CoordinatedTransactionUse2PC();
```
pykello: Why is the check in `StartDistributedExecution()` for activating 2PC not enough? What are some examples? Is this for modify tasks with single shard but multiple placements? Otherwise I guess both `DistributedStatementRequiresRollback()` and `TaskListRequires2PC()` will return `true` if the checks here are `true` and we have already enabled 2PC in `StartDistributedExecution()`?

In any case, some comments on why this is necessary might be useful.

src/backend/distributed/executor/executor.c
onderkalaci: There is a function called `SingleReplicatedTable()`, which I can use here. However, I'm not sure whether it is worth using that.

Because, at this point we're already passed `list_length(taskList) > 1`, in the sense that there is only a single task we're dealing with for the distributed plan. And, (a) We'll only return true for `reference tables` after this point, otherwise always false (b) `SingleReplicatedTable()` or any other such function has to iterate through all the shards/placements to decide, which might be too large for append partitioned tables.

I'm more inclined to keep this as is and maybe add a comment. What do you think @pykello

src/backend/distributed/executor/executor.c
```diff
@@ -2390,20 +2466,24 @@ GetTaskExecutionState(ShardCommandExecution *shardCommandExecution)
...
-		return TASK_EXECUTION_FINISHED;
+		currentTaskExecutionState = TASK_EXECUTION_FINISHED;
 	}
 	else
 	{
-		return TASK_EXECUTION_NOT_FINISHED;
+		currentTaskExecutionState = TASK_EXECUTION_NOT_FINISHED;
 	}
+
+	shardCommandExecution->executionState = currentTaskExecutionState;
```
marcocitus: Should we rename this function to something other than `Get`?

onderkalaci: Maybe only `TaskExecutionState()` ?

marcocitus: It kind of seems like we're actually moving the task execution state machine forward here. Maybe `TaskExecutionStateMachine` to be in line with the other functions.

src/backend/distributed/executor/executor.c
```diff
@@ -2209,14 +2469,21 @@ PlacementExecutionReady(TaskPlacementExecution *placementExecution)
  * finished or failed according to its execution order.
  */
 static TaskExecutionState
-GetTaskExecutionState(ShardCommandExecution *shardCommandExecution)
+TaskExecutionStateMachine(ShardCommandExecution *shardCommandExecution)
```
pykello: Update comment.

src/backend/distributed/executor/executor.c
```diff
@@ -1403,6 +1512,84 @@ ConnectionStateMachine(WorkerSession *session)
...
+		DoesCoordinatedTransactionUse2PC())
+	{
+		/* we don't need 2PC or we're already using 2PC, so no need to continue */
+		return;
+	}
+
+	execution = session->workerPool->distributedExecution;
+	if (TransactionModifiedDistributedTable(execution) &&
+		DistributedPlanModifiesDatabase(execution->plan) &&
+		!ConnectionModifiedPlacement(session->connection))
```
marcocitus: Calling `ConnectionModifiedPlacement` gets inefficient for long transaction blocks that do mostly SELECTs. That might be acceptable, but still wondering whether there is a better place to call this function.

onderkalaci: But this is called after `TransactionModifiedDistributedTable(execution) && DistributedPlanModifiesDatabase(execution->plan)`, so shouldn't affect select only workloads?

onderkalaci: Oh, sorry you're mentioning `mostly single-tenant SELECT`, let me think a bit more

src/backend/distributed/executor/executor.c
```diff
@@ -1435,6 +1622,8 @@ TransactionStateMachine(WorkerSession *session)
 					StartRemoteTransactionBegin(connection);
 
 					transaction->transactionState = REMOTE_TRANS_CLEARING_RESULTS;
+
+					transaction->beginSent = true;
```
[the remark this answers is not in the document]
onderkalaci: Looks like I forgot this, I'll do it once I push the next commit.

src/backend/distributed/executor/executor.c
```diff
@@ -2130,6 +2362,31 @@ PlacementExecutionDone(TaskPlacementExecution *placementExecution, bool succeede
 }
 
 
+/*
+ * ShouldMarkPlacementsInvalidOnFailures returns true if the failure
```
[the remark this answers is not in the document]
onderkalaci: I'll do this as well.

src/backend/distributed/executor/executor.c
marcocitus: I think we should look for a better place to call it from, so that it doesn't get called quite so often. e.g. in StartPlacementExecutionOnSession?

onderkalaci: `StartPlacementExecutionOnSession ` is a little bit trick, because when `SET citus.force_max_query_parallelization TO ON;`, we're calling `StartPlacementListConnection`, which sets the placement accesses even before the connection is established.

Can you check where I added the call now?

src/backend/distributed/executor/executor.c
```diff
@@ -1431,6 +1615,9 @@ TransactionStateMachine(WorkerSession *session)
 			{
 				if (execution->isTransaction)
 				{
+					/* if we're expanding the nodes in a transaction, use 2PC */
+					Activate2PCIfModifyingTransactionExpandsToNewNode(session);
```
marcocitus: I think this is probably the perfect place.

src/backend/distributed/executor/executor.c
```diff
@@ -1403,6 +1509,84 @@ ConnectionStateMachine(WorkerSession *session)
...
+ *          -- activate 2PC since we're only interested in modifications/DDLs
+ *          INSERT INTO distributed_table (dist_key) VALUES (2);
+ */
+static void
+Activate2PCIfModifyingTransactionExpandsToNewNode(WorkerSession *session)
+{
+	DistributedExecution *execution = NULL;
+
+	if (MultiShardCommitProtocol != COMMIT_PROTOCOL_2PC ||
+		DoesCoordinatedTransactionUse2PC())
```
marcocitus: I think we could now afford to drop this block for simplicity.

src/backend/distributed/executor/executor.c
```diff
@@ -2066,23 +2254,51 @@ PlacementExecutionDone(TaskPlacementExecution *placementExecution, bool succeede
...
+	}
+	else
 	{
-		execution->unfinishedTaskCount--;
-		execution->failed = true;
+		/*
+		 * Task execution has already been finished, no need to continue the
+		 * next placement.
+		 */
 		return;
```
marcocitus: Wondering if this code could be structured a little more nicely, e.g..

```c
if (executionState != TASK_EXECUTION_NOT_FINISHED)
{
   /*
    * Task execution has already been finished, no need to continue the
    * next placement.
    */
    return;
}

newExecutionState = TaskExecutionStateMachine(shardCommandExecution);
if (newExecutionState == TASK_EXECUTION_FINISHED)
{
    // should this be in TaskExecutionStateMachine maybe?
    execution->unfinishedTaskCount--;
    return;
}
else if (newExecutionState == TASK_EXECUTION_FAILED)
{
    // should this be in TaskExecutionStateMachine maybe?
    execution->unfinishedTaskCount--;
    execution->failed = true;
    return;
}
else
{
    DoTheExecutionOrderThingWeDoBelow();
}
```

src/backend/distributed/executor/executor.c
```diff
@@ -549,10 +559,36 @@ StartDistributedExecution(DistributedExecution *execution)
...
 			if (TaskListRequires2PC(taskList))
 			{
+				/*
+				 * Although using two phase commit protocol is an independent decision than
+				 * failing on any error, we prefer to couple them. Our motivation is that
+				 * the failures are rare, and we prefer to avoid marking placements invalid
+				 * in case of failures.
+				 */
 				CoordinatedTransactionUse2PC();
+
```
marcocitus: Kind of wondering whether we actually still need this line. Would `Activate2PCIfModifyingTransactionExpandsToNewNode` be sufficient?

src/backend/distributed/executor/executor.c
```diff
@@ -2066,23 +2254,51 @@ PlacementExecutionDone(TaskPlacementExecution *placementExecution, bool succeede
...
+		{
+			execution->unfinishedTaskCount--;
+			return;
+		}
 
-	if (executionState == TASK_EXECUTION_FAILED)
+		if (newExecutionState == TASK_EXECUTION_FAILED)
+		{
+			execution->unfinishedTaskCount--;
+			execution->failed = true;
```
marcocitus: probably could use a comment explaining why the whole execution just failed

src/backend/distributed/executor/executor.c
```diff
@@ -575,6 +611,39 @@ StartDistributedExecution(DistributedExecution *execution)
...
+	taskList = plan->workerJob->taskList;
+	if (list_length(taskList) < 1)
+	{
+		/* does this ever possible? */
+		return false;
+	}
+
+	firstTask = (Task *) linitial(taskList);
+
+	return !ReadOnlyTask(firstTask->taskType);
```
marcocitus: Should we return false for vacuum/analyze?

src/backend/distributed/executor/executor.c
onderkalaci: makes sense

src/backend/distributed/executor/executor.c
onderkalaci: Ah, this actually doesn't work well with the following:

```SQL
select count(*) from users_table where user_id = 1;
select count(*) from users_table where user_id = 2;
insert into users_table (user_id) VALUES (1);
insert into users_table (user_id) VALUES (2);
```

Because the transaction is not `REMOTE_TRANS_INVALID ` but `REMOTE_TRANS_SENT_BEGIN`

src/backend/distributed/executor/executor.c
marcocitus: Agreed

src/backend/distributed/executor/executor.c
```diff
@@ -2066,67 +2255,125 @@ PlacementExecutionDone(TaskPlacementExecution *placementExecution, bool succeede
...
+	else
 	{
-		TaskPlacementExecution *nextPlacementExecution = NULL;
-		int placementExecutionCount = shardCommandExecution->placementExecutionCount;
+		/*
+		 * If the query needs to be executed on any or all placements in order
+		 * and there is a placement left, then make that placement ready-to-start
+		 * by adding it to the appropriate queue.
+		 */
+		executionOrder = shardCommandExecution->executionOrder;
```
marcocitus: could this block be in a separate function?

marcocitus:
I get some weird error when inserting into a reference table when a worker is down:

```sql
BEGIN;
INSERT INTO ref VALUES (1,2);
ERROR:  epoll_ctl() failed: No such file or directory
```

src/backend/distributed/executor/executor.c
onderkalaci: Well, the main issue that we cannot a nice place for this is the `ForceMaxQueryParallelization` because with that the placements are assigned before the connection establishment is done whereas in the regular execution path, the placements are assigned after the connection establishment is done. Otherwise, calling the function on `StartPlacementExecutionOnSession()` would be logical and fix the issue at all.

I really cannot come up with a place other than `MULTI_CONNECTION_CONNECTED`. I can undo the flag check and move back to there. Otherwise, I had to end-up with calling the function in multiple places, which I don't like.

Or, do you have any other suggestions?

src/backend/distributed/executor/executor.c
```diff
@@ -2066,32 +2257,73 @@ PlacementExecutionDone(TaskPlacementExecution *placementExecution, bool succeede
...
+
+
+/*
+ * MovePlacementExecutionsToReady is triggered if the query needs to be
+ * executed on any or all placements in order and there is a placement on
+ * which the execution has not happened yet. If so make that placement
+ * ready-to-start by adding it to the appropriate queue.
+ */
+static void
+MovePlacementExecutionToReady(TaskPlacementExecution *placementExecution, bool succeeded)
```
marcocitus: Maybe `ScheduleNextPlacementExecution`

src/backend/distributed/executor/executor.c
marcocitus: I much prefer calling it in several places over calling it in the main loop. Where could we call it to support `ForceMaxQueryParallelization`?

onderkalaci: makes sense, please see my latest commit(s). I think the second point where I added the call fixed both the edge cases because of `ForceMaxQueryParallelization ` and selects opening remote transaction blocks. I added the tests as well

marcocitus:
Minor note: I found this behaviour surprising, but it's the same on the master branch.
```
postgres=# BEGIN;                                                                                                                                                                  
BEGIN
Time: 0.144 ms
postgres=# INSERT INTO ref VALUES (1,2);                                                                                                                                           
DEBUG:  query before rebuilding: (null)
DEBUG:  query after rebuilding:  INSERT INTO public.ref_102040 (x, y) VALUES (1, 2)
DEBUG:  Creating router plan
DEBUG:  assigned task 0 to node localhost:9700
DEBUG:  opening 1 new connections to localhost:9700
DEBUG:  established connection to localhost:9700
DEBUG:  opening 1 new connections to localhost:9701
DEBUG:  established connection to localhost:9701
DEBUG:  Total number of commands sent over the session 3: 1
DEBUG:  Total number of commands sent over the session 4: 1
INSERT 0 1
Time: 24.519 ms
postgres=# COMMIT;
WARNING:  could not cancel connection: PQcancel() -- connect() failed: Connection refused

ERROR:  terminating connection due to administrator command
```

We're trying to cancel the `PREPARE TRANSACTION`, which means we're still seeing the connection as CONNECTION_OK in `ShutdownConnection`.

src/backend/distributed/executor/executor.c
```diff
@@ -607,10 +679,33 @@ DistributedStatementRequiresRollback(DistributedPlan *distributedPlan)
...
 	task = (Task *) linitial(taskList);
 	if (list_length(task->taskPlacementList) > 1)
 	{
-		return true;
+		/*
+		 * Some tasks don't set replicationModel thus we only
+		 * rely on the anchorShardId, not replicationModel.
+		 *
+		 * TODO: Do we ever need replicationModel in the Task structure?
+		 * Can't we always rely on anchorShardId?
```
marcocitus: can we resolve this todo?

onderkalaci: That's copied from `master`, so we can consider fixing it on [master](https://github.com/citusdata/citus/blob/674b7ce29a9822329a39367db7f1e68207ba745e/src/backend/distributed/executor/multi_router_executor.c#L817-L828).

I think that's also a candidate task for new joiner? I can open an issue for that?

onderkalaci: If that's not a blocker, I'd like to merge this PR @marcocitus

marcocitus: ah, no worries then. :ship:

PR #2723, opened 2019-05-22
Simplify round robin logic on router queries

### LINKED ISSUE #2595: Implement "round-robin" task assignment policy for intermediate results
With #2472, we've implemented round-robin task assignment policy for router queries with only reference tables. We could implement that for router queries with only intermediate results as well for the same performance considerations. 

However, the challenge is that the code that does task assignment currently very tightly coupled with shardIds and placements. We should probably first decouple task assignments/orderings from placements/shardids. 

Related to #2594 as well.

hanefi:
DESCRIPTION: Improves round robin logic on router queries

This PR includes 2 minor improvements to Router queries with Round-Robin policy.

- If a query is router executable, we create a Task list that contains a single task. Therefore there is no need to sort it.
- For creating a router plan, the list of active shard placements are already calculated, no need to repeat this logic in calls to `RoundRobinAssignTaskList()` in physical planner. Thus I replaced this function call with a simpler alternative

The changes here are advised by @marcocitus https://github.com/citusdata/citus/issues/2595#issuecomment-492248397. Closes #2595

### DIFF
--- a/src/backend/distributed/planner/multi_router_planner.c
+++ b/src/backend/distributed/planner/multi_router_planner.c
@@ -155,8 +155,10 @@ static bool RowLocksOnRelations(Node *node, List **rtiLockList);
 static List * SingleShardModifyTaskList(Query *query, uint64 jobId,
 										List *relationShardList, List *placementList,
 										uint64 shardId);
-static void ReorderTaskPlacementsByTaskAssignmentPolicy(Job *job, TaskAssignmentPolicyType
-														taskAssignmentPolicy);
+static void ReorderTaskPlacementsByTaskAssignmentPolicy(Job *job,
+														TaskAssignmentPolicyType
+														taskAssignmentPolicy,
+														List *placementList);
 
 
 /*
@@ -1640,7 +1642,8 @@ RouterJob(Query *originalQuery, PlannerRestrictionContext *plannerRestrictionCon
 		 */
 		if (shardId != INVALID_SHARD_ID)
 		{
-			ReorderTaskPlacementsByTaskAssignmentPolicy(job, TaskAssignmentPolicy);
+			ReorderTaskPlacementsByTaskAssignmentPolicy(job, TaskAssignmentPolicy,
+														placementList);
 		}
 	}
 	else if (isMultiShardModifyQuery)
@@ -1673,12 +1676,31 @@ RouterJob(Query *originalQuery, PlannerRestrictionContext *plannerRestrictionCon
  * By default it does not reorder the task list, implying a first-replica strategy.
  */
 static void
-ReorderTaskPlacementsByTaskAssignmentPolicy(Job *job, TaskAssignmentPolicyType
-											taskAssignmentPolicy)
+ReorderTaskPlacementsByTaskAssignmentPolicy(Job *job,
+											TaskAssignmentPolicyType taskAssignmentPolicy,
+											List *placementList)
 {
 	if (taskAssignmentPolicy == TASK_ASSIGNMENT_ROUND_ROBIN)
 	{
-		job->taskList = RoundRobinAssignTaskList(job->taskList);
+		Task *task = NULL;
+		List *reorderedPlacementList = NIL;
+		ShardPlacement *primaryPlacement = NULL;
+
+		/*
+		 * We hit a single shard on router plans, and there should be only
+		 * one task in the task list
+		 */
+		Assert(list_length(job->taskList) == 1);
+		task = (Task *) linitial(job->taskList);
+
+		/* reorder the placement list */
+		reorderedPlacementList = RoundRobinReorder(task, placementList);
+		task->taskPlacementList = reorderedPlacementList;
+
+		primaryPlacement = (ShardPlacement *) linitial(reorderedPlacementList);
+		ereport(DEBUG3, (errmsg("assigned task %u to node %s:%u", task->taskId,
+								primaryPlacement->nodeName,
+								primaryPlacement->nodePort)));
 	}
 }
 
--- a/src/test/regress/sql/multi_task_assignment_policy.sql
+++ b/src/test/regress/sql/multi_task_assignment_policy.sql
@@ -18,17 +18,16 @@ DECLARE
        shardOfTheTask text;
 begin
   for r in execute qry loop
-       IF r LIKE '%port%' THEN
-       portOfTheTask = substring(r, '([0-9]{1,10})');
+    IF r LIKE '%port%' THEN
+      portOfTheTask = substring(r, '([0-9]{1,10})');
     END IF;
 
     IF r LIKE '%' || table_name || '%' THEN
-       shardOfTheTask = substring(r, '([0-9]{1,10})');
-       return QUERY SELECT  shardOfTheTask || '@' || portOfTheTask ;
+      shardOfTheTask = substring(r, '([0-9]{5,10})');
     END IF;
 
   end loop;
-  return;
+  return QUERY SELECT shardOfTheTask || '@' || portOfTheTask;
 end; $$ language plpgsql;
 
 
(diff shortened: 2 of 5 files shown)


src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1673,12 +1676,29 @@ RouterJob(Query *originalQuery, PlannerRestrictionContext *plannerRestrictionCon
...
 {
 	if (taskAssignmentPolicy == TASK_ASSIGNMENT_ROUND_ROBIN)
 	{
-		job->taskList = RoundRobinAssignTaskList(job->taskList);
+		ShardPlacement *primaryPlacement = NULL;
+
+		/*
+		 * We hit a single shard on router plans, and there should be only
+		 * one task in the TaskList
+		 * */
```
pykello: minor: last line of comment needs formatting.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1673,12 +1676,29 @@ RouterJob(Query *originalQuery, PlannerRestrictionContext *plannerRestrictionCon
...
+		Assert(list_length(job->taskList) == 1);
```
pykello: When built with `--enable-cassert', produces the following warning:

```
warning: ISO C90 forbids mixed declarations and code [-Wdeclaration-after-statement]
```

hanefi: Thanks for pointing this out. I should fine a way to enable this warning flag when I run `make` on my Mac.

I wonder if the CI should fail on this warning.

src/backend/distributed/planner/multi_router_planner.c
hanefi: I believe it is ok now

src/backend/distributed/planner/multi_router_planner.c
hanefi: I fixed the order of declarations and statements in this function block. Now there is one new warning on the CI build. It is on a file I did not touch in this PR, and I believe it is safe to say everything is ok.
```
/home/circleci/project/src/backend/distributed/relay/relay_event_utility.c:132:22: warning: implicit declaration of function 'get_relation_constraint_oid' [-Wimplicit-function-declaration]
```

src/test/regress/sql/multi_task_assignment_policy.sql
```diff
@@ -18,17 +18,16 @@ DECLARE
...
   for r in execute qry loop
-       IF r LIKE '%port%' THEN
-       portOfTheTask = substring(r, '([0-9]{1,10})');
+    IF r LIKE '%port%' THEN
+      portOfTheTask = substring(r, '([0-9]{1,10})');
     END IF;
 
     IF r LIKE '%' || table_name || '%' THEN
-       shardOfTheTask = substring(r, '([0-9]{1,10})');
-       return QUERY SELECT  shardOfTheTask || '@' || portOfTheTask ;
```
hanefi: On some queries the regex for `shardOfTheTask` captures a wrong number when there is a line containing the table name, but no shard ids. For example `([0-9]{1,10})` captures `0` in the following query plan.

I moved the return statement after the end of the loop, so that all regex captures will overwrite the values of variables `portOfTheTask` and `shardOfTheTask` and the latest ones will be returned by the function.

```
                                                         QUERY PLAN
-----------------------------------------------------------------------------------------------------------------------------
 Custom Scan (Citus Router)  (cost=0.00..0.00 rows=0 width=0)
   ->  Distributed Subplan 1_1
         ->  Seq Scan on task_assignment_nonreplicated_hash  (cost=0.00..38.25 rows=11 width=8)
               Filter: (test_id = 3)
   Task Count: 1
   Tasks Shown: All
   ->  Task
         Node: host=localhost port=9702 dbname=postgres
         ->  Merge Left Join  (cost=239.61..435.86 rows=12750 width=12)
               Merge Cond: (intermediate_result.ref_id = ref.test_id)
               ->  Sort  (cost=59.83..62.33 rows=1000 width=8)
                     Sort Key: intermediate_result.ref_id
                     ->  Function Scan on read_intermediate_result intermediate_result  (cost=0.00..10.00 rows=1000 width=8)
               ->  Sort  (cost=179.78..186.16 rows=2550 width=4)
                     Sort Key: ref.test_id
                     ->  Seq Scan on task_assignment_reference_table_102042 ref  (cost=0.00..35.50 rows=2550 width=4)
(16 rows)
```

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1673,12 +1676,31 @@ RouterJob(Query *originalQuery, PlannerRestrictionContext *plannerRestrictionCon
...
 	{
-		job->taskList = RoundRobinAssignTaskList(job->taskList);
+		Task *task = NULL;
+		List *reorderedPlacementList = NIL;
+		ShardPlacement *primaryPlacement = NULL;
+
+		/*
+		 * We hit a single shard on router plans, and there should be only
+		 * one task in the task list
+		 * */
```
mtuncer: `* */` should be `*/` only

src/test/regress/sql/multi_task_assignment_policy.sql
```diff
@@ -236,19 +235,31 @@ $cmd$, 'task_assignment_nonreplicated_hash');
...
 CREATE TABLE task_assignment_test_table_2 (test_id integer);
 SELECT create_distributed_table('task_assignment_test_table_2', 'test_id');
 
-WITH q1 AS (SELECT * FROM task_assignment_test_table_2) SELECT * FROM q1;
-SET LOCAL citus.task_assignment_policy TO 'round-robin';
-WITH q1 AS (SELECT * FROM task_assignment_test_table_2) SELECT * FROM q1;
-ROLLBACK;
+SET citus.task_assignment_policy TO 'round-robin';
+
+-- run the query two times to make sure that it hits two different worrkers
```
mtuncer: s/worrkers/workers

src/test/regress/sql/multi_task_assignment_policy.sql
```diff
@@ -236,19 +235,31 @@ $cmd$, 'task_assignment_nonreplicated_hash');
...
+-- on consecutive runs
+INSERT INTO explain_outputs
+SELECT parse_explain_output($cmd$
+EXPLAIN WITH q1 AS (SELECT * FROM task_assignment_test_table_2) SELECT * FROM q1
```
mtuncer: This does not test the code.
1 - it needs to be router query, this is not
2 - you need a replication factor > 1
3 - or test with a reference table

hanefi: The CTE here is replaced with an `intermediate_result` and is rewritten into:
```sql
SELECT test_id
FROM (
  SELECT intermediate_result.test_id
  FROM read_intermediate_result('13_1'::text, 'binary'::citus_copy_format)
    intermediate_result(test_id integer)) q1;
```

The resulting query is router executable, and the it is executed on different workers if using round-robin. Here are 2 `EXPLAIN` outputs:

```
                                                                 QUERY PLAN
--------------------------------------------------------------------------------------------------------------------------------------------
 Custom Scan (Citus Router)  (cost=0.00..0.00 rows=0 width=0)
   ->  Distributed Subplan 15_1
         ->  Custom Scan (Citus Real-Time)  (cost=0.00..0.00 rows=0 width=0)
               Task Count: 32
               Tasks Shown: One of 32
               ->  Task
                     Node: host=localhost port=9701 dbname=postgres
                     ->  Seq Scan on task_assignment_test_table_2_102043 task_assignment_test_table_2  (cost=0.00..35.50 rows=2550 width=4)
   Task Count: 1
   Tasks Shown: All
   ->  Task
         Node: host=localhost port=9702 dbname=postgres
         ->  Function Scan on read_intermediate_result intermediate_result  (cost=0.00..10.00 rows=1000 width=4)
(13 rows)
```
```
                                                                 QUERY PLAN
--------------------------------------------------------------------------------------------------------------------------------------------
 Custom Scan (Citus Router)  (cost=0.00..0.00 rows=0 width=0)
   ->  Distributed Subplan 17_1
         ->  Custom Scan (Citus Real-Time)  (cost=0.00..0.00 rows=0 width=0)
               Task Count: 32
               Tasks Shown: One of 32
               ->  Task
                     Node: host=localhost port=9701 dbname=postgres
                     ->  Seq Scan on task_assignment_test_table_2_102043 task_assignment_test_table_2  (cost=0.00..35.50 rows=2550 width=4)
   Task Count: 1
   Tasks Shown: All
   ->  Task
         Node: host=localhost port=9701 dbname=postgres
         ->  Function Scan on read_intermediate_result intermediate_result  (cost=0.00..10.00 rows=1000 width=4)
(13 rows)
```

I should have explained this better in the comments. I may also add a simpler router query in this test to make it easier to read&understand

mtuncer:
while you are improving things could you also look at this line
https://github.com/citusdata/citus/blob/master/src/backend/distributed/planner/multi_router_planner.c#L2037

It is very unlikely but we should consider restarting the counter when we hit some upper bound, like worker_count, etc

hanefi:
> while you are improving things could you also look at this line
> https://github.com/citusdata/citus/blob/master/src/backend/distributed/planner/multi_router_planner.c#L2037
> 
> It is very unlikely to overflow this counter but we should consider restarting the counter when we hit some upper bound, like worker_count, etc

@pykello had a similar remark when `zeroShardQueryRoundRobin` variable was first introduced.
https://github.com/citusdata/citus/pull/1835#discussion_r153610941

1. It is unsigned, so it will reveret to zero in case of overflow.
2. The only time it is used is here:
https://github.com/citusdata/citus/blob/af16ed7308f98eb3f3b8073abe07e7bc5f44f247/src/backend/distributed/planner/multi_router_planner.c#L2025
Finding the modulus with worker count makes things perfect. Even if it overflows, things are still fine

Issue #2630, opened 2019-03-14
TRUNCATE on partitioned tables very slow


marcocitus:
Because we handle TRUNCATE from a trigger, we end up truncating all partitions in a partitioned distributed table separately, which can take a long time when there are hundreds of partitions.

onderkalaci:
Similar for partitioned tables, dropping a simple table with 365 partitions took:
```SQL
drop table collections_list;
DROP TABLE
Time: 111391.475 ms (01:51.391)

```

PR #2724, opened 2019-05-22
Stop using master_modify_multiple_shards in TRUNCATE

marcocitus:
We currently rely on `master_modify_multiple_shards` in truncate, but I plan to deprecate and remove that function. This PR changes the truncate implementation to construct a task list and execute it directly. I considered calling `ExecuteDistributedDDLJob`, but we do not want the truncate to propagate to workers with metadata in MX.

We should clean up truncate further and handle it fully in the utility hook, since the current trigger-based implementation complicates upgrades and causes issues like #2630 .

### DIFF
--- a/src/backend/distributed/master/master_truncate.c
+++ b/src/backend/distributed/master/master_truncate.c
@@ -19,10 +19,15 @@
 #include "distributed/master_metadata_utility.h"
 #include "distributed/master_protocol.h"
 #include "distributed/multi_join_order.h"
+#include "distributed/multi_router_executor.h"
 #include "distributed/pg_dist_partition.h"
+#include "distributed/resource_lock.h"
 #include "utils/builtins.h"
 #include "utils/lsyscache.h"
 #include "utils/rel.h"
+
+
+static List * TruncateTaskList(Oid relationId);
 
 
 /* exports for SQL callable functions */
@@ -39,9 +44,6 @@ citus_truncate_trigger(PG_FUNCTION_ARGS)
 	TriggerData *triggerData = NULL;
 	Relation truncatedRelation = NULL;
 	Oid relationId = InvalidOid;
-	char *relationName = NULL;
-	Oid schemaId = InvalidOid;
-	char *schemaName = NULL;
 	char partitionMethod = 0;
 
 	if (!CALLED_AS_TRIGGER(fcinfo))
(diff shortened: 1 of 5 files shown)


src/backend/distributed/master/master_truncate.c
```diff
@@ -65,22 +67,79 @@ citus_truncate_trigger(PG_FUNCTION_ARGS)
...
+	{
+		ShardInterval *shardInterval = (ShardInterval *) lfirst(shardIntervalCell);
+		uint64 shardId = shardInterval->shardId;
+		StringInfo shardQueryString = makeStringInfo();
+		Task *task = NULL;
+		char *shardName = pstrdup(relationName);
+
+		AppendShardIdToName(&shardName, shardId);
+
+		appendStringInfo(shardQueryString, "TRUNCATE TABLE %s CASCADE",
```
onderkalaci: Very minor, not related to this PR, feel free to ignore: I've realized that we don't have a TRUNCATE test on schemas with quotes. You might consider adding one TRUNCATE test on one of [these tables](https://github.com/citusdata/citus/blob/master/src/test/regress/sql/multi_schema_support.sql#L793-L799)

src/backend/distributed/master/master_truncate.c
```diff
@@ -18,13 +18,21 @@
 #include "distributed/commands/utility_hook.h"
 #include "distributed/master_metadata_utility.h"
 #include "distributed/master_protocol.h"
+#include "distributed/multi_executor.h"
```
onderkalaci: As far as I can see, we don't need `#include "distributed/multi_executor.h"`, `#include "distributed/multi_physical_planner.h"` and `#include "distributed/relay_utility.h"`

src/backend/distributed/master/master_truncate.c
```diff
@@ -65,22 +67,79 @@ citus_truncate_trigger(PG_FUNCTION_ARGS)
...
-		char *qualifiedTableName = quote_qualified_identifier(schemaName, relationName);
+		List *taskList = TruncateTaskList(relationId);
 
-		appendStringInfo(truncateStatement, "TRUNCATE TABLE %s CASCADE",
-						 qualifiedTableName);
-
-		DirectFunctionCall1(master_modify_multiple_shards,
-							CStringGetTextDatum(truncateStatement->data));
+		if (MultiShardConnectionType == SEQUENTIAL_CONNECTION)
+		{
```
onderkalaci: Just brainstorming on this: I think one interesting note here is that with trigger approach we're *in a sense* mixing planning and execution.  `TRUNCATE parent_table CASCADE`, would first create list of tasks, then execute it for the parent table, then create list of tasks for child_1 -> execute it, create list of tasks for child_n -> execute it.

I think for `TRUNCATE` it doesn't matter because we're always acquiring access exclusive lock on the table. Anything interesting to you here?

marcocitus: I think the current implementation is correct, but very slow. We don't really want to cascade on the coordinator if we are already going to cascade on the workers.

src/backend/distributed/master/master_truncate.c
marcocitus: added

PR #2725, opened 2019-05-22
Deprecate master_modify_multiple_shards

marcocitus:
DESCRIPTION: Deprecate master_modify_multiple_shards

`master_modify_multiple_shards` is starting to incur maintainenance costs that we'd rather not pay as we have full support for multi-shard DML and the function should no longer be used.

This PR replaces the internals of `master_modify_multiple_shards` by a call to `ExecuteQueryStringIntoDestReceiver`, which means the query now goes through the regular planner and executor and we no longer need to consider locking and such. As a side-effect, `master_modify_multiple_shards` no longer returns the number of affected rows, which did not seem worth investing in.

It also throws a deprecation warning whenever it is called and it is replaced or removed it in all regression tests except `multi_master_modify` just to verify it is still backwards compatible (apart from the output).

### DIFF
--- a/src/backend/distributed/master/master_modify_multiple_shards.c
+++ b/src/backend/distributed/master/master_modify_multiple_shards.c
@@ -59,188 +59,35 @@
 #include "utils/memutils.h"
 
 
-static List * ModifyMultipleShardsTaskList(Query *query, List *shardIntervalList, TaskType
-										   taskType);
-
-
 PG_FUNCTION_INFO_V1(master_modify_multiple_shards);
 
 
 /*
  * master_modify_multiple_shards takes in a DELETE or UPDATE query string and
- * pushes the query to shards. It finds shards that match the criteria defined
- * in the delete command, generates the same delete query string for each of the
- * found shards with distributed table name replaced with the shard name and
- * sends the queries to the workers. It uses one-phase or two-phase commit
- * transactions depending on citus.copy_transaction_manager value.
+ * executes it. This is mainly provided for backwards compatibility, users
+ * should use regular UPDATE and DELETE commands.
  */
 Datum
 master_modify_multiple_shards(PG_FUNCTION_ARGS)
 {
 	text *queryText = PG_GETARG_TEXT_P(0);
 	char *queryString = text_to_cstring(queryText);
-	List *queryTreeList = NIL;
-	Oid relationId = InvalidOid;
-	Index tableId = 1;
-	Query *modifyQuery = NULL;
-	Node *queryTreeNode;
-	List *restrictClauseList = NIL;
-	bool failOK = false;
-	List *prunedShardIntervalList = NIL;
-	List *taskList = NIL;
-	int32 affectedTupleCount = 0;
-	CmdType operation = CMD_UNKNOWN;
-	TaskType taskType = TASK_TYPE_INVALID_FIRST;
-	bool truncateOperation = false;
 	RawStmt *rawStmt = (RawStmt *) ParseTreeRawStmt(queryString);
-	queryTreeNode = rawStmt->stmt;
+	Node *queryTreeNode = rawStmt->stmt;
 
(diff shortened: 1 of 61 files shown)


src/backend/distributed/master/master_modify_multiple_shards.c
```diff
@@ -79,30 +75,14 @@ master_modify_multiple_shards(PG_FUNCTION_ARGS)
 {
```
onderkalaci: Minor: We'd probably want to update the function comment

src/backend/distributed/executor/multi_executor.c
```diff
@@ -442,7 +442,7 @@ ExecuteQueryIntoDestReceiver(Query *query, ParamListInfo params, DestReceiver *d
 
 /*
  * ExecuteIntoDestReceiver plans and executes a query and sends results to the given
- * DestReceiver.
+ * DestReceiver and returns the affected tuple count.
```
onderkalaci: Does it really return anything?

src/backend/distributed/master/master_modify_multiple_shards.c
```diff
@@ -79,30 +75,14 @@ master_modify_multiple_shards(PG_FUNCTION_ARGS)
...
-	List *queryTreeList = NIL;
 	Oid relationId = InvalidOid;
-	Index tableId = 1;
-	Query *modifyQuery = NULL;
-	Node *queryTreeNode;
-	List *restrictClauseList = NIL;
 	bool failOK = false;
-	List *prunedShardIntervalList = NIL;
-	List *taskList = NIL;
 	int32 affectedTupleCount = 0;
```
onderkalaci: Minor: Maybe get rid of this variable as well

src/backend/distributed/master/master_modify_multiple_shards.c
```diff
@@ -115,132 +95,19 @@ master_modify_multiple_shards(PG_FUNCTION_ARGS)
 		relationId = RangeVarGetRelid(updateStatement->relation, NoLock, failOK);
 		EnsureTablePermissions(relationId, ACL_UPDATE);
```
onderkalaci: I guess we don't need these (`EnsureTablePermissions`) as well, bu who cares :)

src/backend/distributed/master/master_modify_multiple_shards.c
```diff
@@ -115,132 +95,19 @@ master_modify_multiple_shards(PG_FUNCTION_ARGS)
...
-	if (MultiShardConnectionType == SEQUENTIAL_CONNECTION)
-	{
-		affectedTupleCount =
-			ExecuteModifyTasksSequentiallyWithoutResults(taskList, operation);
-	}
-	else
-	{
-		affectedTupleCount = ExecuteModifyTasksWithoutResults(taskList);
-	}
+	ExecuteQueryStringIntoDestReceiver(queryString, NULL, None_Receiver);
```
onderkalaci: Just to clarify for myself, we cannot have parameters passed to `master_modify_multiple_shards` so we can safely pass NULL?

marcocitus: correct

src/test/regress/expected/multi_partitioning.out
```diff
@@ -1395,24 +1331,6 @@ SELECT relation::regclass, locktype, mode FROM pg_locks WHERE relation::regclass
  partitioning_locks_for_select | relation | AccessShareLock
 (5 rows)
 
-COMMIT;
--- test locks on master_modify_multiple_shards
-BEGIN;
-SELECT master_modify_multiple_shards('UPDATE partitioning_locks SET time = ''2009-03-01''');
```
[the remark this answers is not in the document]
marcocitus: I'll just bring them back as multi-shard as DML.

PR #2727, opened 2019-05-22
Fix misc typos (2)

serprex:
Fix typos in comments & internal READMEs


pykello:
@serprex before merging, can you:

1. Merge two commits into one?
2. Configure your git so we see the correct author for your commits here?

Thanks.

PR #2728, opened 2019-05-23
Update the regression test outputs in unified executor

onderkalaci:
This PR is on top of #2713 and fixes several minor regression tests. Each commit is self explanatory. But basically, we're either supporting a complicated transaction block or producing much better error messages.

After this one, we only have **3** remaining failing tests in check-multi and check-multi-mx. Those tests include the two known issues: (a) COPY cannot re-use connections yet (b) We're not properly acquiring the partitioning locks yet

### DIFF
--- a/src/test/regress/expected/subquery_and_cte.out
+++ b/src/test/regress/expected/subquery_and_cte.out
@@ -470,7 +470,7 @@ FOR i IN 1..3 LOOP
 			GROUP BY
 				user_id, cte.value_2);
 	EXCEPTION WHEN OTHERS THEN
-		IF SQLERRM LIKE 'failed to execute task%' THEN
+		IF SQLERRM LIKE 'more than one row returned by a subquery%%' THEN
 			errors_received := errors_received + 1;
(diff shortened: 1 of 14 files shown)


src/test/regress/expected/subquery_and_cte.out
```diff
@@ -470,7 +470,7 @@ FOR i IN 1..3 LOOP
 			GROUP BY
 				user_id, cte.value_2);
 	EXCEPTION WHEN OTHERS THEN
-		IF SQLERRM LIKE 'failed to execute task%' THEN
+		IF SQLERRM LIKE 'more than one row returned by a subquery%%' THEN
```
marcocitus: mainly just wondering: do we return a useful error code here?

onderkalaci: Yes, we return the same error code that Postgres does:
```
ERROR:  21000: more than one row returned by a subquery used as an expression
```

which is 

`21000 | cardinality_violation`

PR #2733, opened 2019-05-24
Fix join alias resolution

### LINKED ISSUE #2642: Two join aliases with the same name fails to deparse accurately
PostgreSQL parser already disallows the same join alias on the same level of the query such as:

```SQL
SELECT *
FROM
      (events_table
          INNER JOIN 
         users_table_ref ON (events_table.user_id = users_table_ref.value_2)) AS join_alias(user_id_deep),
     (events_table
      INNER JOIN 
     users_table_ref ON (events_table.user_id = users_table_ref.value_2)) AS join_alias(user_id_deep)
WHERE (bar.user_id_deep = join_alias.user_id_deep);
ERROR:  table name "join_alias" specified more than once
```

In this case, we're discussing this even the aliases appear in different queries/subqueries.

```SQL
SELECT r
FROM
  (SELECT 
  	user_id_deep, random() as r -- prevent pulling up the subquery
   FROM (events_table
         INNER JOIN 
         users_table_ref ON (events_table.user_id = users_table_ref.value_2)) AS join_alias(user_id_deep)) AS bar,
     (events_table
      INNER JOIN 
     users_table_ref ON (events_table.user_id = users_table_ref.value_2)) AS join_alias(user_id_deep)
WHERE (bar.user_id_deep = join_alias.user_id_deep);
```

Worker query:
```SQL
...
LOG:  issuing COPY (SELECT worker_column_1 AS r FROM (SELECT bar.r AS worker_column_1 FROM (SELECT join_alias_1.user_id_deep, random() AS r FROM (public.events_table_102050 events_table_1 JOIN public.users_table_ref_102072 users_table_ref_1 ON ((events_table_1.user_id OPERATOR(pg_catalog.=) users_table_ref_1.value_2))) join_alias(user_id_deep, "time", event_type, value_2, value_3, value_4, user_id, time_1, value_1, value_2_1, value_3_1, value_4_1)) bar, (public.events_table_102050 events_table JOIN public.users_table_ref_102072 users_table_ref ON ((events_table.user_id OPERATOR(pg_catalog.=) users_table_ref.value_2))) join_alias(user_id_deep, "time", event_type, value_2, value_3, value_4, user_id, time_1, value_1, value_2_1, value_3_1, value_4_1) WHERE (bar.user_id_deep OPERATOR(pg_catalog.=) join_alias.user_id_deep)) worker_subquery) TO STDOUT
...
WARNING:  missing FROM-clause entry for table "join_alias_1"
...
ERROR:  failed to execute task 11
```

Note that when you create such a rule (e.g., `CREATE RULE r1 AS ON INSERT table DO query`) on non-distributed tables, the rule is saved correctly and works fine. So, it makes me think that the issue might be related to Citus' additional deparsing rules.

serprex:
Fixes #2642

FROM (query) alias ignored renaming
In nested subqueries the select list would rename, while the join alias would not respect that

DESCRIPTION: fix error caused by joins with shadowed aliases

### DIFF
--- a/src/backend/distributed/utils/ruleutils_11.c
+++ b/src/backend/distributed/utils/ruleutils_11.c
@@ -7263,8 +7263,16 @@ get_from_clause_item(Node *jtnode, Query *query, deparse_context *context)
 		/* Yes, it's correct to put alias after the right paren ... */
 		if (j->alias != NULL)
 		{
+			/*
+			 * Note that it's correct to emit an alias clause if and only if
+			 * there was one originally.  Otherwise we'd be converting a named
+			 * join to unnamed or vice versa, which creates semantic
+			 * subtleties we don't want.  However, we might print a different
+			 * alias name than was there originally.
+			 */
 			appendStringInfo(buf, " %s",
-							 quote_identifier(j->alias->aliasname));
+							 quote_identifier(get_rtable_name(j->rtindex,
+															  context)));
 			get_column_alias_list(colinfo, context);
 		}
 	}
--- a/src/test/regress/sql/multi_name_resolution.sql
+++ b/src/test/regress/sql/multi_name_resolution.sql
@@ -1,0 +1,29 @@
+--
+-- MULTI_NAME_RESOLUTION
+--
+
+-- There was a failure compiling queries with shadowed subquery aliases
+-- https://github.com/citusdata/citus/issues/2642
+
+CREATE SCHEMA multi_name_resolution;
+SET search_path TO multi_name_resolution;
+
+create table namenest1 (id integer primary key, user_id integer);
+create table namenest2 (id integer primary key, value_2 integer);
+select * from create_distributed_table('namenest1', 'id');
+select * from create_reference_table('namenest2');
+
+SELECT r
+FROM (
+	SELECT id_deep, random() as r -- prevent pulling up the subquery
+	FROM (
+		namenest1
+		JOIN namenest2 ON (namenest1.user_id = namenest2.value_2)
+	) AS join_alias(id_deep)
+) AS bar, (
+	namenest1
+	JOIN namenest2 ON (namenest1.user_id = namenest2.value_2)
+) AS join_alias(id_deep)
+WHERE bar.id_deep = join_alias.id_deep;
+
+DROP SCHEMA multi_name_resolution CASCADE;
--- a/src/backend/distributed/utils/ruleutils_10.c
+++ b/src/backend/distributed/utils/ruleutils_10.c
@@ -7246,8 +7246,16 @@ get_from_clause_item(Node *jtnode, Query *query, deparse_context *context)
 		/* Yes, it's correct to put alias after the right paren ... */
 		if (j->alias != NULL)
 		{
+			/*
+			 * Note that it's correct to emit an alias clause if and only if
+			 * there was one originally.  Otherwise we'd be converting a named
+			 * join to unnamed or vice versa, which creates semantic
+			 * subtleties we don't want.  However, we might print a different
+			 * alias name than was there originally.
+			 */
 			appendStringInfo(buf, " %s",
-							 quote_identifier(j->alias->aliasname));
+							 quote_identifier(get_rtable_name(j->rtindex,
+															  context)));
 			get_column_alias_list(colinfo, context);
 		}
 	}
--- a/src/test/regress/expected/multi_name_resolution.out
+++ b/src/test/regress/expected/multi_name_resolution.out
@@ -1,0 +1,41 @@
+--
+-- MULTI_NAME_RESOLUTION
+--
+-- There was a failure compiling queries with shadowed subquery aliases
+-- https://github.com/citusdata/citus/issues/2642
+CREATE SCHEMA multi_name_resolution;
+SET search_path TO multi_name_resolution;
+create table namenest1 (id integer primary key, user_id integer);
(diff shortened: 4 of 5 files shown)


src/backend/distributed/utils/ruleutils_11.c
```diff
@@ -7264,7 +7264,7 @@ get_from_clause_item(Node *jtnode, Query *query, deparse_context *context)
 		if (j->alias != NULL)
 		{
 			appendStringInfo(buf, " %s",
-							 quote_identifier(j->alias->aliasname));
+							 quote_identifier(get_rtable_name(j->rtindex, context)));
```
pykello: This file (and `ruleutils_10.c`) is mostly copy/pasted versions of PostgreSQL's `postgres/backend/utils/adt/ruleutils.c`, with the purpose of just making some of the functions we need public. Are there any solutions to this issue that doesn't involve changing `ruleutils` files?

src/test/regress/sql/multi_name_nested.sql
```diff
@@ -0,0 +1,20 @@
+create table namenest1 (id integer primary key, user_id integer);
```
pykello: 1. We usually put a comment explaining what category of features/issues each file is testing. Look at other `.sql` files to see what style we use. Without that, it is really difficult to figure out what this file tests.
2. Why are you not including this in `multi_subquery.sql`?

src/test/regress/sql/multi_name_nested.sql
```diff
@@ -0,0 +1,20 @@
...
+create table namenest2 (id integer primary key, value_2 integer);
+select * from create_distributed_table('namenest1', 'id');
+select * from create_reference_table('namenest2');
+
+SELECT r
```
pykello: Please add a comment explaining what this specific query is testing. It is very difficult to figure that out without having much context about the issue and by just looking at the query. People shouldn't need to spend more than a minute or two to figure out what each query is testing. See other tests for examples of comments, and the style we use.

src/test/regress/sql/multi_name_nested.sql
serprex: 1. Updated file
2. Bug is a flaw in name resolution which subqueries were able to expose. If this should be in `multi_subquery.sql` let me know

src/backend/distributed/utils/ruleutils_11.c
serprex: Alternative would be to rely on shadowing & have both selects resolve as `join_alias` as opposed to the inner resolving `join_alias_1`

serprex:
I've confirmed this is an upstream bug
```
postgres=# create view asdf as SELECT r
FROM
  (SELECT  user_id_deep, random() as r -- prevent pulling up the subquery
   FROM (events_table
         INNER JOIN
         users_table_ref ON (events_table.user_id = users_table_ref.value_2)) AS join_alias(user_id_dee
p)) AS bar,
     (events_table
      INNER JOIN
     users_table_ref ON (events_table.user_id = users_table_ref.value_2)) AS join_alias(user_id_deep)
WHERE (bar.user_id_deep = join_alias.user_id_deep);
CREATE VIEW
postgres=# \d+ asdf
                                 View "public.asdf"
 Column |       Type       | Collation | Nullable | Default | Storage | Description 
--------+------------------+-----------+----------+---------+---------+-------------
 r      | double precision |           |          |         | plain   | View definition:
 SELECT bar.r
   FROM ( SELECT join_alias_1.user_id_deep,
            random() AS r
           FROM (events_table events_table_1
             JOIN users_table_ref users_table_ref_1 ON events_table_1.user_id = users_table_ref_1.value_2) join_alias(user_id_deep, user_id, id, value_2)) bar,
    (events_table
     JOIN users_table_ref ON events_table.user_id = users_table_ref.value_2) join_alias(user_id_deep, user_id, id, value_2)
  WHERE bar.user_id_deep = join_alias.user_id_deep;
```

serprex:
Upstream: https://www.postgresql.org/message-id/CY4PR2101MB080246F2955FF58A6ED1FEAC98140%40CY4PR2101MB0802.namprd21.prod.outlook.com

src/test/regress/sql/multi_name_resolution.sql
```diff
@@ -0,0 +1,27 @@
...
+		JOIN namenest2 ON (namenest1.user_id = namenest2.value_2)
+	) AS join_alias(user_id_deep)
+) AS bar, (
+	namenest1
+	JOIN namenest2 ON (namenest1.user_id = namenest2.value_2)
+) AS join_alias(user_id_deep)
+WHERE (bar.user_id_deep = join_alias.user_id_deep);
+
+drop table namenest1;
+drop table namenest2;
```
marcocitus: for cleanliness and to prevent naming conflicts for tests that run concurrently, we often do something like:

```sql
CREATE SCHEMA multi_name_resolution;
SET search_path TO multi_name_resolution;
...
DROP SCHEMA multi_name_resolution CASCADE;
```

serprex: Updated

serprex:
[Postgres has committed this fix to their ruleutils](https://git.postgresql.org/gitweb/?p=postgresql.git;a=commit;h=3d99a81397abb8bb7b95aee794d6644e174e174f)

onderkalaci:
> Postgres has committed this fix to their ruleutils
Updating branch to match Tom's commit

Great!

PR #2735, opened 2019-05-28
Implement true sequential execution for multi-row INSERTS

onderkalaci:
This PR is against `unified_executor`

Execute the individual tasks one by one. Note that this is different than
MultiShardConnectionType == SEQUENTIAL_CONNECTION case (e.g., sequential execution
mode). In that case, running the tasks across the nodes in parallel is acceptable
and implemented in that way.

However, the executions that are qualified here would perform poorly if the
tasks across the workers are executed in parallel. We currently qualify only
one class of distributed queries here, multi-row INSERTs. If we do not enforce
true sequential execution, concurrent multi-row upserts could easily form
a distributed deadlock when the upserts touch the same rows.

Another interesting case here is that setting
	"citus.force_max_query_parallelization=on"
forces the execution to still be one-by-one, but opening #placement
connections. I could probably prefer retuning false in
`ShouldRunTasksSequentially()`, any thoughts?

### DIFF
--- a/src/backend/distributed/executor/executor.c
+++ b/src/backend/distributed/executor/executor.c
@@ -45,8 +45,13 @@
  */
 typedef struct DistributedExecution
 {
-	/* distributed query plan */
-	DistributedPlan *plan;
+	/* the corresponding distributed plan's operation */
+	CmdType operation;
+
+	List *tasksToExecute;
+
+	/* the corresponding distributed plan has RETURNING */
+	bool hasReturning;
 
 	/* Parameters for parameterized plans. Can be NULL. */
 	ParamListInfo paramListInfo;
@@ -262,9 +267,6 @@ typedef struct ShardCommandExecution
 	/* description of the task */
 	Task *task;
 
-	/* job of which the task is part */
-	Job *job;
-
 	/* order in which the command should be replicated on replicas */
 	PlacementExecutionOrder executionOrder;
 
@@ -340,18 +342,26 @@ bool ForceMaxQueryParallelization = false;
 
 
 /* local functions */
-static DistributedExecution * CreateDistributedExecution(DistributedPlan *distributedPlan,
+static DistributedExecution * CreateDistributedExecution(CmdType operation,
+														 List *taskList,
+														 bool hasReturning,
 														 ParamListInfo paramListInfo,
 														 TupleDesc tupleDescriptor,
 														 Tuplestorestate *tupleStore,
 														 int targetPoolSize);
 static void StartDistributedExecution(DistributedExecution *execution);
 static void RunDistributedExecution(DistributedExecution *execution);
-static void FinishDistributedExecution(DistributedExecution *execution);
+static bool ShouldRunTasksSequentially(List *taskList);
+static void SequentialRunDistributedExecution(DistributedExecution *execution);
 
-static bool DistributedPlanModifiesDatabase(DistributedPlan *plan);
-static bool DistributedStatementRequiresRollback(DistributedPlan *distributedPlan);
+static void FinishDistributedExecution(DistributedExecution *execution);
+static void CleanUpSessions(DistributedExecution *execution);
+
 static void LockPartitionsForDistributedPlan(DistributedPlan *distributedPlan);
+static bool DistributedExecutionModifiesDatabase(DistributedExecution *execution);
+static bool DistributedPlanModifiesDatabase(DistributedPlan *plan);
+static bool TaskListModifiesDatabase(CmdType operation, List *taskList);
+static bool DistributedExecutionRequiresRollback(DistributedExecution *execution);
 static void AssignTasksToConnections(DistributedExecution *execution);
 static void UnclaimAllSessionConnections(List *sessionList);
 static bool ShouldForceOpeningConnection(TaskPlacementExecutionState
@@ -414,6 +424,9 @@ CitusExecScan(CustomScanState *node)
 		bool interTransactions = false;
 		int targetPoolSize = DEFAULT_POOL_SIZE;
 
+		Job *job = distributedPlan->workerJob;
+		List *taskList = job->taskList;
+
 		/*
 		 * PostgreSQL takes locks on all partitions in the executor. It's not entirely
 		 * clear why this is necessary (instead of locking the parent during DDL), but
@@ -433,26 +446,29 @@ CitusExecScan(CustomScanState *node)
 		}
 		else if (ForceMaxQueryParallelization)
 		{
-			Job *job = distributedPlan->workerJob;
-			List *taskList = job->taskList;
-
 			/* we might at most have the length of tasks pool size */
 			targetPoolSize = list_length(taskList);
 		}
 
-		execution = CreateDistributedExecution(distributedPlan, paramListInfo,
-											   tupleDescriptor, tupleStore,
-											   targetPoolSize);
+		execution = CreateDistributedExecution(distributedPlan->operation, taskList,
+											   distributedPlan->hasReturning,
+											   paramListInfo, tupleDescriptor,
+											   tupleStore, targetPoolSize);
 
 		StartDistributedExecution(execution);
-		RunDistributedExecution(execution);
+
+		if (ShouldRunTasksSequentially(execution->tasksToExecute))
+		{
+			SequentialRunDistributedExecution(execution);
+		}
+		else
+		{
+			RunDistributedExecution(execution);
+		}
 
 		if (distributedPlan->operation != CMD_SELECT)
 		{
 			executorState->es_processed = execution->rowsProcessed;
-
-			/* prevent copying shards in same transaction */
-			XactModificationLevel = XACT_MODIFICATION_DATA;
 		}
 
 		FinishDistributedExecution(execution);
@@ -495,24 +511,17 @@ ExecuteTaskListExtended(CmdType operation, List *taskList,
 						TupleDesc tupleDescriptor, Tuplestorestate *tupleStore,
 						bool hasReturning, int targetPoolSize)
 {
-	DistributedPlan *distributedPlan = NULL;
 	DistributedExecution *execution = NULL;
 	ParamListInfo paramListInfo = NULL;
-
-	distributedPlan = CitusMakeNode(DistributedPlan);
-	distributedPlan->operation = operation;
-	distributedPlan->workerJob = CitusMakeNode(Job);
-	distributedPlan->workerJob->taskList = taskList;
-	distributedPlan->hasReturning = hasReturning;
 
 	if (MultiShardConnectionType == SEQUENTIAL_CONNECTION)
 	{
 		targetPoolSize = 1;
 	}
 
-	execution = CreateDistributedExecution(distributedPlan, paramListInfo,
-										   tupleDescriptor, tupleStore,
-										   targetPoolSize);
+	execution =
+		CreateDistributedExecution(operation, taskList, hasReturning, paramListInfo,
+								   tupleDescriptor, tupleStore, targetPoolSize);
 
 	StartDistributedExecution(execution);
 	RunDistributedExecution(execution);
@@ -527,16 +536,17 @@ ExecuteTaskListExtended(CmdType operation, List *taskList,
  * a distributed plan.
  */
 DistributedExecution *
-CreateDistributedExecution(DistributedPlan *distributedPlan,
+CreateDistributedExecution(CmdType operation, List *taskList, bool hasReturning,
 						   ParamListInfo paramListInfo, TupleDesc tupleDescriptor,
 						   Tuplestorestate *tupleStore, int targetPoolSize)
 {
-	DistributedExecution *execution = NULL;
-	Job *job = distributedPlan->workerJob;
-	List *taskList = job->taskList;
+	DistributedExecution *execution =
+		(DistributedExecution *) palloc0(sizeof(DistributedExecution));
 
-	execution = (DistributedExecution *) palloc0(sizeof(DistributedExecution));
-	execution->plan = distributedPlan;
+	execution->operation = operation;
+	execution->tasksToExecute = taskList;
+	execution->hasReturning = hasReturning;
+
 	execution->executionStats =
 		(DistributedExecutionStats *) palloc0(sizeof(DistributedExecutionStats));
 	execution->paramListInfo = paramListInfo;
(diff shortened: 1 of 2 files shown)


src/backend/distributed/executor/executor.c
```diff
@@ -994,6 +1005,149 @@ FindOrCreateWorkerSession(WorkerPool *workerPool, MultiConnection *connection)
...
+		 * individual singleTaskExecution execution here.
+		 */
+
+		/* simply call the regular execution function */
+		RunDistributedExecution(singleTaskExecution);
+
+		/* remember the total number of rows processed */
+		execution->rowsProcessed += singleTaskExecution->rowsProcessed;
+
+		FinishDistributedExecution(singleTaskExecution);
```
onderkalaci: I'm still thinking whether this line is necessary/correct. Happy to get feedback on this.

src/backend/distributed/executor/executor.c
```diff
@@ -994,6 +1005,149 @@ FindOrCreateWorkerSession(WorkerPool *workerPool, MultiConnection *connection)
...
+	 * plans for a restricted set of operations (see ShouldRunTasksSequentially()).
+	 * So, we don't expect other plans go through this code, unless implemented properly.
+	 */
+	Assert (singleTaskPlan->operation == CMD_INSERT);
+	Assert (singleTaskPlan->subPlanList == NIL);
+	Assert (singleTaskPlan->insertSelectSubquery == NULL);
+	Assert (singleTaskPlan->insertTargetList == NIL);
+	Assert (singleTaskPlan->intermediateResultIdPrefix == NULL);
+	Assert (singleTaskPlan->masterQuery == NULL);
+	Assert (list_length(originalPlan->workerJob->taskList) > 1);
```
pykello: It is not really clear from the name of the function that these `Assert()`s should be done here. Also this creates a tight coupling between `ShouldRunTasksSequentially()` and this function (i.e. whenever we modify one, we should also modify the other), while it doesn't seem very necessary.

What do you think about moving these `Assert()`s to the `if (initialTask->rowValuesLists != NIL)` block in `ShouldRunTasksSequentially()`?

src/backend/distributed/executor/executor.c
pykello: Some parts of `FinishDistributedExecution()` is necessary (e.g. most of the `foreach()` loop in there), and other parts seem harmless, so my take is that this is correct. I need to think more though.

src/backend/distributed/executor/executor.c
```diff
@@ -994,6 +1005,149 @@ FindOrCreateWorkerSession(WorkerPool *workerPool, MultiConnection *connection)
...
+
+
+/*
+ * CreateSingleTaskDistributedPlan gets a task and a distributed plan. The
+ * function returns a newly allocated distributed plan where the new plan
+ * contains only the task that is provided to the function along with
+ * some of the necessary fields set.
+ */
+static DistributedPlan *
+CreateSingleTaskDistributedPlan(Task *task, DistributedPlan *originalPlan)
```
pykello: If you agree with rest of my feedback for the implementation of this function, which would reduce this to 2 lines, what do you think about just inlining those two lines instead of creating a function for that? It is not clear from the name what this function does, so readers will need to read the function anyway, i.e. this function doesn't abstract much. Also, the comments aren't very helpful at explaining what the function does.

src/backend/distributed/executor/executor.c
```diff
@@ -532,8 +544,6 @@ StartDistributedExecution(DistributedExecution *execution)
 	{
 		RecordParallelRelationAccessForTaskList(taskList);
 	}
-
-	AssignTasksToConnections(execution);
```
pykello: Now that we have remove this line, we should update comments for `StartDistributedExecution()`.

src/backend/distributed/executor/executor.c
```diff
@@ -994,6 +1005,149 @@ FindOrCreateWorkerSession(WorkerPool *workerPool, MultiConnection *connection)
...
+		singleTaskExecution->targetPoolSize = 1;
+		singleTaskExecution->unfinishedTaskCount = 1;
+		singleTaskExecution->totalTaskCount = 1;
+
+		/* RunDistributedExecution() will set the sessions, be explict here */
+		singleTaskExecution->sessionList = NIL;
+		singleTaskExecution->waitEventSet = NULL;
+
+		/* we're going to execute a single task only */
+		singleTaskExecution->plan = singleTaskDistributedPlan;
```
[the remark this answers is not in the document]
onderkalaci: Makes sense, but we need to set to fields from the main execution:
- `isTransaction`: To make sure that new connections  starts remote transaction when necessary
- `executionStats`: To make sure that the total intermediate result size doesn't exceed across sequential executions.

I've explicitly set those fields and added comment. Does that look good?

src/backend/distributed/executor/executor.c
```diff
@@ -994,6 +1005,149 @@ FindOrCreateWorkerSession(WorkerPool *workerPool, MultiConnection *connection)
...
+/*
+ * CreateSingleTaskDistributedPlan gets a task and a distributed plan. The
+ * function returns a newly allocated distributed plan where the new plan
+ * contains only the task that is provided to the function along with
+ * some of the necessary fields set.
+ */
+static DistributedPlan *
+CreateSingleTaskDistributedPlan(Task *task, DistributedPlan *originalPlan)
+{
+	DistributedPlan *singleTaskPlan = palloc0(sizeof(DistributedPlan));
```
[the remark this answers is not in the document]
onderkalaci: > (One question could be why not to look into what's going on with the crash, I think I should do that irrespective of this task..)

Please ignore the above. I've probably got confused because of another crash. I've tried hard to repro that, but no luck. So, I'm not confident that we can copy the plan.

onderkalaci:
> Thanks for the changes. This looks good to me, but as you suggested before, let's wait for @marcocitus 's feedback before merging.

Yes, @marcocitus please give a 👍 or 👎 to the approach implemented here once you're back.

src/backend/distributed/executor/executor.c
```diff
@@ -995,21 +1005,136 @@ FindOrCreateWorkerSession(WorkerPool *workerPool, MultiConnection *connection)
...
+	DistributedPlan *distributedPlan = execution->plan;
+	Job *job = distributedPlan->workerJob;
+	List *taskList = job->taskList;
+
+	ListCell *taskCell = NULL;
+
+	foreach(taskCell, taskList)
+	{
+		Task *taskToExecute = (Task *) lfirst(taskCell);
+		DistributedPlan *singleTaskDistributedPlan = copyObject(distributedPlan);
```
marcocitus: This is quite expensive to do on every iteration.

src/backend/distributed/executor/executor.c
marcocitus: This does seem prone to bugs. If I'm making changes to these functions I don't expect create-start-create-run-finish-create-run-finish as a possible execution order.

src/backend/distributed/executor/executor.c
onderkalaci: I guess that's mostly because the tasks could be giant in this case, right? I'll push a commit which does only copy the required parts of the plan for multi-row inserts.

src/backend/distributed/executor/executor.c
```diff
@@ -995,21 +1022,177 @@ FindOrCreateWorkerSession(WorkerPool *workerPool, MultiConnection *connection)
...
+		/* found a multi-row INSERT */
+		return true;
+	}
+
+	return false;
+}
+
+
+/*
+ * SequentialRunDistributedExecution gets a distributed execution and
```
onderkalaci: Just to share my thoughts for the reviewer and future reference: 
I think this function (and the logic in general) is pretty rigid. But, with the asserts and comments, I'm OK for now. If we want to make it more flexible in the future , we'd hit the asserts immediately. I don't want to over engineer this for cases that we currently don't support and will probably never need to support.

onderkalaci:
And, one more note for future reference. I used the following query to test the changes on my local machine (which used to lead a distributed deadlock on unified executor):

```SQL
create table test_a(key int PRIMARY KEY, value bigint);
select create_distributed_table('test_a','key');

INSERT INTO test_a VALUES (1,1), (2,2), (3,3), (4,4),(5,5),(6,6),(7,7),(8,8), (9,9),(10,10) ON CONFLICT(key) DO UPDATE SET value=test_a.value+EXCLUDED.value RETURNING key, value/key;
```

And, running it with pgbench: ` pgbench -T 100 -P 2 -j4 -c4  -f /tmp/f.sql postgres`

The performance comparison with the master branch seems fine. Once we do real-benchmarks, I'll make sure to have various tests on multi-row inserts/upserts.

src/backend/distributed/executor/executor.c
```diff
@@ -993,21 +1020,96 @@ FindOrCreateWorkerSession(WorkerPool *workerPool, MultiConnection *connection)
...
+		/* we'll only need one connection */
+		execution->targetPoolSize = 1;
+
+		/* execute each task one by one */
+		execution->tasksToExecute = list_make1(taskToExecute);
+		execution->totalTaskCount = 1;
+		execution->unfinishedTaskCount = 1;
+
+		/* simply call the regular execution function */
+		RunDistributedExecution(execution);
```
onderkalaci: @marcocitus Does that fit what into what you had in mind?

src/backend/distributed/executor/executor.c
```diff
@@ -1080,93 +1064,28 @@ ShouldRunTasksSequentially(DistributedPlan *plan)
...
-									   targetPoolSize);
-
-		/*
-		 * Use some fields as the same which are safe to share across executions
-		 * and required to to have across single task executions.
-		 */
-		singleTaskExecution->isTransaction = execution->isTransaction;
-		singleTaskExecution->executionStats = execution->executionStats;
+		/* we'll only need one connection */
+		execution->targetPoolSize = 1;
```
marcocitus: probably redundant given that we'll only do one task at a time.

src/backend/distributed/executor/executor.c
marcocitus: This definitely seems a lot cleaner, could you rebase on unified_executor to make the changes easier to read?

onderkalaci: should be ready to have a look.

src/backend/distributed/executor/executor.c
```diff
@@ -530,20 +542,19 @@ CreateDistributedExecution(DistributedPlan *distributedPlan,
...
 {
-	DistributedPlan *distributedPlan = execution->plan;
-	Job *job = distributedPlan->workerJob;
-	List *taskList = job->taskList;
+	List *taskList = execution->tasksToExecute;
 
 	if (MultiShardCommitProtocol != COMMIT_PROTOCOL_BARE)
 	{
-		if (DistributedStatementRequiresRollback(distributedPlan))
+		if (DistributedStatementRequiresRollback(execution->operation, taskList))
```
onderkalaci: Maybe I should just rename this function to `DistributedExecutionRequiresRollback()` and pass `execution` itself?

src/backend/distributed/executor/executor.c
```diff
@@ -622,16 +629,34 @@ DistributedStatementRequiresRollback(DistributedPlan *distributedPlan)
...
+static void
 FinishDistributedExecution(DistributedExecution *execution)
 {
-	DistributedPlan *distributedPlan = execution->plan;
+	UnsetCitusNoticeLevel();
+
+	if (execution->operation != CMD_SELECT)
+	{
+		/* prevent copying shards in same transaction */
+		XactModificationLevel = XACT_MODIFICATION_DATA;
```
marcocitus: We also do this in CitusExecScan, we should probably remove it from there.

src/backend/distributed/executor/executor.c
```diff
@@ -622,16 +629,34 @@ DistributedStatementRequiresRollback(DistributedPlan *distributedPlan)
...
  * clears their state.
  */
-void
+static void
 FinishDistributedExecution(DistributedExecution *execution)
 {
-	DistributedPlan *distributedPlan = execution->plan;
+	UnsetCitusNoticeLevel();
+
+	if (execution->operation != CMD_SELECT)
```
marcocitus: Would be good to put this in a function.

src/backend/distributed/executor/executor.c
onderkalaci: We should probably remove it from `CitusExecScan` because of `ExecuteTaskList`?

marcocitus: Right

src/backend/distributed/executor/executor.c
marcocitus: seems fine

src/backend/distributed/executor/executor.c
```diff
@@ -1080,16 +1179,25 @@ RunDistributedExecution(DistributedExecution *execution)
 				execution->waitFlagsChanged = false;
 			}
 
+			if (MultiShardConnectionType == SEQUENTIAL_CONNECTION)
+			{
+				connectionsInUse = list_length(execution->workerList);
```
marcocitus: Is there a way to avoid this if statement? It seems a bit hacky.

onderkalaci: I think we could consider keeping this for performance optimization (as the commit message says as well). Because, the `latch.c` has the following loop:
```C
	for (cur_event = set->events, cur_pollfd = set->pollfds;
		 cur_event < (set->events + set->nevents) &&
		 returned_events < nevents;
		 cur_event++, cur_pollfd++)
```

Where `connectionsInUse` == `nevents` and  `set->nevents` == `list_length(sessionList)` in `BuildWaitEventSet`.

So, we're instructing the loop to continue for `connectionsInUse` times. I don't have any numbers on the performance effect of this change.

If we're to remove this, we need to remove `Assert (maxWaitEventCount >= connectionCount + 2);` as well, because that doesn't hold true for sequential executions, where there might be more sessions than we need.

If it feels hacky, I can remove now. Once we're doing the benchmarks, I can check the effect of this change.

onderkalaci: removed the if/else

src/test/regress/expected/failure_multi_row_insert.out
```diff
@@ -43,10 +43,10 @@ SELECT citus.mitmproxy('conn.onQuery(query="^INSERT").kill()');
 (1 row)
 
 INSERT INTO distributed_table VALUES (1,1), (1,2), (1,3);
-ERROR:  server closed the connection unexpectedly
+ERROR:  connection error: localhost:9060
```
onderkalaci: now that failure PRs merged, I pushed 1 more commit that updates the error messages here.

src/backend/distributed/executor/executor.c
```diff
@@ -1177,8 +1277,14 @@ RunDistributedExecution(DistributedExecution *execution)
...
-			/* we should always have more (or equal) waitEvents */
-			Assert(maxWaitEventCount >= connectionCount + 2);
+
+			/*
+			 * maxWaitEventCount depends to workerCount in sequential execution
+			 * else to session count.
+			 */
+			Assert(MultiShardConnectionType == SEQUENTIAL_CONNECTION ?
+				   (maxWaitEventCount >= list_length(execution->workerList) + 2) :
+				   (maxWaitEventCount >= connectionCount + 2));
```
marcocitus: I feel like this assert is mostly noise.

src/backend/distributed/executor/executor.c
marcocitus: Bump

onderkalaci: Sorry, missed that. Do you mean moving these two lines into a function like
```C
MarkAsModifyingXacts()
```
etc?

src/backend/distributed/executor/executor.c
onderkalaci: Will remove

src/backend/distributed/executor/executor.c
marcocitus: I meant using the `DistributedExecutionModifiesDatabase` function instead of `!= CMD_SELECT`

onderkalaci:
we've few more greens with this PR, and one failing failure test:`failure_savepoints `

PR #2736, opened 2019-05-28
Make sure that the regression tests are resistant to execution order changes

onderkalaci:
Same as #2657 but mostly for `check-multi-mx` tests.

Since the changes are very trivial and only in the regression tests, I'll go ahead and merge once the tests pass. Otherwise, maintaining these branches becomes a burden.

@pykello Feel free to go over the PR and revert it if you see anything unusual/wrong etc.


pykello:
Looks good to me.

Issue #2430, opened 2018-10-09
Should a failed ANALYZE roll back the transaction?


jasonmp85:
While writing failure tests, I noticed that injecting a failure on one node during an ANALYZE transaction would result in a failure.

Do we really need to mark connections during ANALYZE as "critical"? It seems like we could get away with a warning in this case.

PR #2738, opened 2019-05-29
Update failure test regression test outputs

onderkalaci:
We've changed the failure behaviour a bit, and also the error messages
that show up to the user. This PR covers majority of the updates.

4 failure tests are failing:
- `failure_savepoints` crashing, @pykello will look into that.
- `multi_row_insert` will just work on top of #2735 
- `failure_vacuum` is failing, I need to discuss the failure behaviour for `VACUUM` with @pykello and @marcocitus 
- `failure_connection_establishment` is failing because we used to have `could not establish connection after XXX msecs` error. Should we have that on the unified executor as well?

Anyway, this is ready for a quick review, we can handle the above one-by-one separately.

### DIFF
--- a/src/test/regress/expected/failure_real_time_select.out
+++ b/src/test/regress/expected/failure_real_time_select.out
@@ -26,18 +26,20 @@ INSERT INTO test_table VALUES(1,1,1),(1,2,2),(2,1,1),(2,2,2),(3,1,1),(3,2,2);
 -- Kill when the first COPY command arrived, since we have a single placement 
 -- it is expected to error out.
 SET client_min_messages TO ERROR;
-SELECT citus.mitmproxy('conn.onQuery(query="^COPY").kill()');
+SELECT citus.mitmproxy('conn.onQuery(query="^SELECT").kill()');
  mitmproxy 
 -----------
  
 (1 row)
 
-SELECT public.raise_failed_execution('SELECT count(*) FROM test_table');
-ERROR:  Task failed to execute
-CONTEXT:  PL/pgSQL function public.raise_failed_execution(text) line 6 at RAISE
+SELECT count(*) FROM test_table;
+ERROR:  connection error: localhost:9060
+DETAIL:  server closed the connection unexpectedly
+	This probably means the server terminated abnormally
+	before or while processing the request.
 SET client_min_messages TO DEFAULT;
 -- Kill the connection with a CTE
-SELECT citus.mitmproxy('conn.onQuery(query="^COPY").kill()');
+SELECT citus.mitmproxy('conn.onQuery(query="^SELECT").kill()');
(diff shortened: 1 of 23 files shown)


src/test/regress/expected/failure_real_time_select.out
```diff
@@ -125,27 +131,23 @@ results AS (SELECT * FROM test_table)
 SELECT * FROM test_table
 WHERE test_table.id > (SELECT id FROM results);
 ERROR:  canceling statement due to user request
--- Since Citus opens a new connection after a failure within the real time
--- execution and after(1).kill() kills connection after a successful execution
--- for each connection, following transaciton does not fail.
+-- Citus fails if the connection that is already used fails afterwards
```
onderkalaci: This is a behavior change we implemented in the unified executor. If any query on a shard over a connection succeeds, further failures end with errors. See here for a more detailed [discussion](https://github.com/citusdata/citus/pull/2713#issue-279801940)

onderkalaci:
> failure_vacuum is failing, I need to discuss the failure behaviour for VACUUM with @pykello and @marcocitus

I think I have a better idea on this. I think #2430 was confusing to me, but now `failure_vacuum` is passing as well. Let me know if you think otherwise.

onderkalaci:
> failure_connection_establishment is failing because we used to have could not establish connection after XXX msecs error. Should we have that on the unified executor as well?

I think the implementation of `citus.node_connection_timeout` should change with the unified executor. In the existing executors, we errored out whenever **any** connection failed  to be established within `citus.node_connection_timeout`.


With unified executor, it should probably be only valid when at least one connection cannot be established within `citus.node_connection_timeout`.

Let me see if that's simple enough to implement.

src/test/regress/expected/multi_query_directory_cleanup_0.out
```diff
@@ -0,0 +1,273 @@
...
+(1 row)
+
+DECLARE c_19 CURSOR FOR SELECT sum(l_extendedprice * l_discount) as revenue FROM lineitem;
+FETCH 1 FROM c_19;
+    revenue    
+---------------
+ 22770844.7654
+(1 row)
+
+SELECT * FROM pg_ls_dir('base/pgsql_job_cache') f ORDER BY f;
```
onderkalaci: Well, we need `multi_query_directory_cleanup_0.out ` because of task-tracker still uses temp tables and we've task tracker schedule which runs this test (and makes sense to run it on tt)

onderkalaci:
Finally `ci/circleci: test-10_check-tt-van-mx — Your tests passed on CircleCI!` and same for `ci/circleci: test-11_check-tt-van-mx `.

`check-failure` and `check-multi` missing few tests.

PR #2741, opened 2019-05-31
Unified executor honors citus.node_connection_timeout

onderkalaci:
This PR is on top of multiple PRs #2713, #2738 and #2740, so might make sense to review those first. Due to some rebase issue, I had to include 2 commits, where the first commit is coming from #2740. So, if you want to review this PR, please only review the second one here.
   
```
With this commit, unified executor errors out if even
a single connection cannot be established within
`citus.node_connection_timeout.`
    
And, as a side effect this fixes failure_connection_establishment
test.
```

### DIFF
--- a/src/backend/distributed/executor/executor.c
+++ b/src/backend/distributed/executor/executor.c
@@ -178,6 +178,18 @@ typedef struct WorkerPool
 	 */
 	dlist_head readyTaskQueue;
 	int readyTaskCount;
+
+	/*
+	 * We keep this for enforcing the connection timeouts. In our definition, a pool
+	 * starts when the first connection establishment starts.
+	 */
+	TimestampTz poolStartTime;
+
+	/*
+	 * This is only set in WorkerPoolFailed() function. Once a pool fails, we do not
+	 * use it anymore.
+	 */
+	bool failed;
 } WorkerPool;
 
 struct TaskPlacementExecution;
@@ -351,6 +363,7 @@ static WorkerPool * FindOrCreateWorkerPool(DistributedExecution *execution,
 static WorkerSession * FindOrCreateWorkerSession(WorkerPool *workerPool,
 												 MultiConnection *connection);
 static void ManageWorkerPool(WorkerPool *workerPool);
+static void CheckConnectionTimeout(WorkerPool *workerPool);
 static WaitEventSet * BuildWaitEventSet(List *sessionList);
 static TaskPlacementExecution * PopPlacementExecution(WorkerSession *session);
 static TaskPlacementExecution * PopAssignedPlacementExecution(WorkerSession *session);
@@ -925,8 +938,8 @@ AssignTasksToConnections(DistributedExecution *execution)
 				 * FindOrCreateWorkerSession ensures that we only have one session per
 				 * connection.
 				 */
-				WorkerSession *session = FindOrCreateWorkerSession(workerPool,
-																   connection);
+				WorkerSession *session =
+					FindOrCreateWorkerSession(workerPool, connection);
 
 				elog(DEBUG4, "%s:%d has an assigned task", connection->hostname,
 					 connection->port);
@@ -963,6 +976,7 @@ AssignTasksToConnections(DistributedExecution *execution)
 					/* task is ready to execute on any session */
 					dlist_push_tail(&workerPool->readyTaskQueue,
 									&placementExecution->workerReadyQueueNode);
+
 					workerPool->readyTaskCount++;
 				}
 				else
@@ -1107,6 +1121,7 @@ FindOrCreateWorkerPool(DistributedExecution *execution, WorkerNode *workerNode)
 
 	workerPool = (WorkerPool *) palloc0(sizeof(WorkerPool));
 	workerPool->node = workerNode;
+	workerPool->poolStartTime = 0;
 	workerPool->distributedExecution = execution;
 	dlist_init(&workerPool->pendingTaskQueue);
 	dlist_init(&workerPool->readyTaskQueue);
@@ -1155,6 +1170,15 @@ FindOrCreateWorkerSession(WorkerPool *workerPool, MultiConnection *connection)
 		workerPool->idleConnectionCount++;
 	}
 
+	/*
+	 * Record the first connection establishment time to the pool. We need this
+	 * to enforce NodeConnectionTimeout.
+	 */
+	if (list_length(workerPool->sessionList) == 0)
+	{
+		workerPool->poolStartTime = GetCurrentTimestamp();
+	}
+
 	workerPool->sessionList = lappend(workerPool->sessionList, session);
 	execution->sessionList = lappend(execution->sessionList, session);
 
@@ -1170,6 +1194,8 @@ FindOrCreateWorkerSession(WorkerPool *workerPool, MultiConnection *connection)
 void
 RunDistributedExecution(DistributedExecution *execution)
 {
+	const TimestampTz deadline = TimestampTzPlusMilliseconds(GetCurrentTimestamp(),
+															 NodeConnectionTimeout);
 	int workerCount = list_length(execution->workerList);
 
 	/* additional 2 is for postmaster and latch */
@@ -1186,7 +1212,7 @@ RunDistributedExecution(DistributedExecution *execution)
 
 		while (execution->unfinishedTaskCount > 0 && !cancellationReceived)
 		{
-			long timeout = 1000;
+			long timeout = DeadlineTimestampTzToTimeout(deadline);
 			int connectionCount = list_length(execution->sessionList);
 			int eventCount = 0;
 			int eventIndex = 0;
@@ -1289,7 +1315,7 @@ ManageWorkerPool(WorkerPool *workerPool)
 	DistributedExecution *execution = workerPool->distributedExecution;
 	WorkerNode *workerNode = workerPool->node;
 	int targetPoolSize = execution->targetPoolSize;
-	int currentConnectionCount = list_length(workerPool->sessionList);
+	int initiatedConnectionCount = list_length(workerPool->sessionList);
 	int activeConnectionCount = workerPool->activeConnectionCount;
 	int failedConnectionCount = workerPool->failedConnectionCount;
 	int idleConnectionCount = workerPool->idleConnectionCount;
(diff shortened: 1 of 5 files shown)


marcocitus:
The real-time executor tracks timeouts per connection.

For connections that are opened after initial pool creation, are we ok not having a timeout?

It would probably make sense for the timeout check to appear under `MULTI_CONNECTION_CONNECTING` in the state machine.

onderkalaci:
> It would probably make sense for the timeout check to appear under MULTI_CONNECTION_CONNECTING in the state machine.

My reasoning was that even the whole execution could be finished with a single connection. So, checking per connection would prevent some queries to finish compared to this approach?

onderkalaci:
> It would probably make sense for the timeout check to appear under MULTI_CONNECTION_CONNECTING in the state machine.


To be more explicit, erroring out on the timeout is a behaviour change compared to the existing executors for replication > 1. For example, in real-time executor, if the connection establishment fails within `NodeConnectionTimeout `, we warn the user, close the connection and try the replica if exists. (Similar for reference tables in router executor)

However, to me the behaviour here seems more in-line with #2713 where we error out if any query fails during  [execution](https://github.com/citusdata/citus/pull/2713#issue-279801940). Happy to revisit the decision(s), if you think that this is a common scenario (where we expect the unified executor to make those scenarios less common :) )

Still, instead of erroring out, I could use `WorkerPoolFailed()` to continue the execution, which probably makes more sense. But, I prefer this approach over tracking per connection for simplicity and fitting into the architecture.


-- 

Also one more minor note for readers, this is different than when the node is down. If a node is down, we'd give a `WARNING` on one placement and continue with the other one. This case is more when the node and/or the pgbouncers are very busy and so we cannot get even **a single** connection. So, overall I'd expect to hit this very infrequently.

src/backend/distributed/executor/executor.c
```diff
@@ -1230,6 +1233,28 @@ ManageWorkerPool(WorkerPool *workerPool)
...
+	/* we should always have more (or equal) active connections than idle connections */
+	Assert (activeConnectionCount >= idleConnectionCount);
+
+	/* we should never have less than 0 connections ever */
+	Assert (activeConnectionCount >= 0 && idleConnectionCount >= 0);
+
+	TimestampTz poolCreationTime = workerPool->poolCreationTime;
+	TimestampTz now = GetCurrentTimestamp();
+
+	if (activeConnectionCount == 0 &&
```
onderkalaci: Hmm, we should also consider the case when `force_max_query_parallelization = true`. In that case, we should probably check `len(workerPool->sessionList) != activeConnectionCount ` instead of `activeConnectionCount == 0`.

src/backend/distributed/executor/executor.c
```diff
@@ -888,6 +898,17 @@ AssignTasksToConnections(DistributedExecution *execution)
 					/* task is ready to execute on any session */
 					dlist_push_tail(&workerPool->readyTaskQueue,
 									&placementExecution->workerReadyQueueNode);
+
+					/*
+					 * Record the first task assignment time to the pool. We need this
+					 * to enforce NodeConnectionTimeout.
+					 */
+					if (workerPool->activeConnectionCount == 0 &&
+						workerPool->readyTaskCount == 0)
```
marcocitus: I'm a bit confused why we care about the task assignment. Why not first connection establishment?

src/backend/distributed/executor/executor.c
```diff
@@ -1301,6 +1327,75 @@ ManageWorkerPool(WorkerPool *workerPool)
...
+ * number of connections. In the end, the executor is capable of using one connection
+ * to execute multiple tasks.
+ */
+static void
+CheckConnectionTimeout(WorkerPool *workerPool)
+{
+	DistributedExecution *execution = workerPool->distributedExecution;
+	TimestampTz firstTaskAssignmentTime = workerPool->firstTaskAssignmentTime;
+	TimestampTz now = GetCurrentTimestamp();
+
```
marcocitus: might be nice to return early if we already managed to establish a connection before the timeout

src/backend/distributed/executor/executor.c
onderkalaci: > Why not first connection establishment?

Do you mean the pool creation as I used to do?

When replication factor > 1 and single task query, we create the pools at the same time but only try the first pool to access the placement until the timeout. So, the time that actually matters seems when the first task becomes ready to execute on the pool.

src/backend/distributed/executor/executor.c
onderkalaci: Hmm, due to `ForceMaxQueryParallelization ` checks, it becomes slightly complicated. But I think compilers could optimize this fairly well, so shall I keep as-is?

src/backend/distributed/executor/executor.c
marcocitus: but we don't open a connection if there are no ready tasks, right?

I was expecting it to be somewhere around here:
https://github.com/citusdata/citus/blob/32678f8f871e2324dcb2446010946d92457a81aa/src/backend/distributed/executor/executor.c#L1298

I'm also wondering whether it's worth marking sessions as "critical" to make the distinction between critical and non-critical connection establishment a bit more obvious to the reader.

onderkalaci: > I was expecting it to be somewhere around here:

Ok, that makes sense. I agree that it is not easier to follow/understand the code.

> I'm also wondering whether it's worth marking sessions as "critical" to make the distinction between critical and non-critical connection establishment a bit more obvious to the reader.

Do you mean on #2713 get rid of (or rely less on) `errorOnAnyFailure ` and instead mark connections critical? Let me also think about that a bit more if that's what you're asking.

src/backend/distributed/executor/executor.c
onderkalaci: I added the early exit check now.

src/test/regress/expected/failure_connection_establishment.out
```diff
@@ -81,13 +80,13 @@ SELECT citus.mitmproxy('conn.delay(500)');
 
 -- we cannot control which replica of the reference table will be queried and there is
 -- only one specific client we can control the connection for.
--- by using round-robin task_assignment_policy we can force to hit both machines. We will
--- use two output files to match both orders to verify there is 1 that times out and falls
--- through to read from the other machine
+-- by using round-robin task_assignment_policy we can force to hit both machines.
```
onderkalaci: Clarification on the change here:
we talked on this with @thanodnl who has added the original test. The two output file is unnecessary (and not exist in the repo as well), since (a) We made the log level ERROR and ignore WARNINGs (b) via round-robin we enforce that we'll hit the node behind the proxy once. And, output of `dump_network_traffic ` ensures that.

So, the comment seems stale and we update it now.

src/backend/distributed/executor/executor.c
onderkalaci: Oh no, the check I added is different than what you ask. I could maybe move `CheckConnectionTimeout(workerPool);` after the first 3 if checks in `ManageWorkerPool` instead of where it is now. That'd decrease the frequency of calling the function a lot?

src/backend/distributed/executor/executor.c
```diff
@@ -1301,6 +1331,80 @@ ManageWorkerPool(WorkerPool *workerPool)
...
+	/*
+	 * This is a special case where we assign tasks to sessions even before
+	 * the connections are established. So, make sure to apply similar
+	 * restrictions. In this case, make sure that we get all the connections
+	 * established.
+	 */
+	if (ForceMaxQueryParallelization)
+	{
+		requiredActiveConnectionCount = list_length(workerPool->sessionList);
```
marcocitus: I'm kind of wondering whether it would be cleaner to distinguish between critical and non-critical connections, but it would probably be more trouble than it's worth. I think the long-term solution is to remove the flag.

src/backend/distributed/executor/executor.c
```diff
@@ -846,6 +856,13 @@ AssignTasksToConnections(DistributedExecution *execution)
 				connection =
 					StartPlacementListConnection(connectionFlags, placementAccessList,
 												 NULL);
+
+				/*
+				 * We'll assign tasks almost at the same time, in the loop right now.
+				 * So, it is acceptable to have a tiny difference between the first and
+				 * the last assignments.
+				 */
+				workerPool->connectionStart = connection->connectionStart;
```
marcocitus: I think this is the right place to put this, but wondering whether we should only do this when we first establish a new connection. Right now it probably doesn't matter, but if we keep adding a new connection every 10ms the time-out will keep resetting.

src/backend/distributed/executor/executor.c
```diff
@@ -173,6 +173,14 @@ typedef struct WorkerPool
...
 	dlist_head readyTaskQueue;
 	int readyTaskCount;
+
+	/*
+	 * We keep this for enforcing the connection timeouts. In our definition, a pool
+	 * starts when the first connection establishment starts.
+	 */
+	TimestampTz poolStartTime;
+
+	bool failed;
```
marcocitus: Would be good to have a comment.

src/backend/distributed/executor/executor.c
```diff
@@ -1343,6 +1375,78 @@ ManageWorkerPool(WorkerPool *workerPool)
...
+
+	/*
+	 * This is a special case where we assign tasks to sessions even before
+	 * the connections are established. So, make sure to apply similar
+	 * restrictions. In this case, make sure that we get all the connections
+	 * established.
+	 */
+	if (ForceMaxQueryParallelization)
+	{
+		requiredActiveConnectionCount = initiatedConnectionCount;
```
marcocitus: Is this affected by #2764?

marcocitus: (I think not)

onderkalaci: As far as I can see, no. Because, even with #2764 we're still initiating all the connections at the same time, so it is either `initiatedConnectionCount == 0` or `requiredActiveConnectionCount = initiatedConnectionCount;`.

But, it would be effected if decide to open connections gradually, as you suggested such as with `10ms` wait loop. But, still, for  `ForceMaxQueryParallelization ` we probably wouldn't implement such thing and this should be fine.

src/backend/distributed/executor/executor.c
```diff
@@ -2234,6 +2341,35 @@ WorkerPoolFailed(WorkerPool *workerPool)
 
 	/* we do not want more connections in this pool */
 	workerPool->readyTaskCount = 0;
+	workerPool->failed = true;
+
+	/*
+	 * When each task is forced to be executed over a single connection,
+	 * we should consider the worker failover (e.g., replication > 1) and
+	 * restart the connection timeout.
```
marcocitus: Maybe elaborate on why, e.g.

The reason is that when replication factor is > 1 and we are performing a SELECT, then we only establish connections for the specific placements that we will read from. However, when a worker pool fails, we will need to establish multiple new connection to other workers and the query can only succeed if all those connections are established.

PR #2742, opened 2019-05-31
Also check rewrittenQuery jointree for outer join

### LINKED ISSUE #2739: Outer JOINs wrapped as a subquery error out
```
create table test_a(id int);
CREATE TABLE
citus=> create table test_b(id int);
CREATE TABLE
citus=> select create_distributed_table('test_a','id');
 create_distributed_table
--------------------------

(1 row)

citus=> select create_distributed_table('test_b','id');
 create_distributed_table
--------------------------

(1 row)

citus=> select * from test_a a left join test_b b on a.id=b.id;
 id | id
----+----
(0 rows)

citus=> select * from (select * from test_a a left join test_b b on a.id=b.id)lo;
DETAIL:  Cartesian products are currently unsupported
```
Such queries could be generated by an ORM

serprex:
Fixes #2739 

DESCRIPTION: Fix a regression in outer joining subqueries introduced in 0.8.2

This is currently a straight forward fix based on the comment made by @onderkalaci. I will look to investigate further as to how this regression was introduced & whether there's a deeper issue here

Aside: to backport this to 8.2 do I make a pull request to release-8.2?

### DIFF
--- a/src/backend/distributed/planner/query_pushdown_planning.c
+++ b/src/backend/distributed/planner/query_pushdown_planning.c
@@ -143,6 +143,19 @@ ShouldUseSubqueryPushDown(Query *originalQuery, Query *rewrittenQuery)
 	}
 
 	/*
+	 * Original query may not have an outer join while rewritten query does.
+	 * We should push down in this case.
+	 * An example of this is https://github.com/citusdata/citus/issues/2739
+	 * where postgres pulls-up the outer-join in the subquery.
+	 */
+	if (FindNodeCheck((Node *) rewrittenQuery->jointree, IsOuterJoinExpr))
+	{
+		/* Assert what _should_ be only situation this occurs in. */
+		Assert(JoinTreeContainsSubquery(originalQuery));
+		return true;
+	}
+
+	/*
 	 * Some unsupported join clauses in logical planner
 	 * may be supported by subquery pushdown planner.
 	 */
--- a/src/test/regress/expected/multi_subquery.out
+++ b/src/test/regress/expected/multi_subquery.out
@@ -462,6 +462,25 @@ SELECT DISTINCT ON (t1.user_id) t1.user_id, t2.value_1, t2.value_2, t2.value_3
  ORDER BY 1 DESC, 2 DESC, 3 DESC, 4 DESC 
  LIMIT 5;
 ERROR:  complex joins are only supported when all distributed tables are joined on their distribution columns with equal operator
+-- outer joins as subqueries should work
+-- https://github.com/citusdata/citus/issues/2739
+SELECT user_id, value_1, event_type
+FROM (
+	SELECT a.user_id, a.value_1, b.event_type
+	FROM users_table a
+	LEFT JOIN events_table b ON a.user_id = b.user_id
+) lo
+ORDER BY 1, 2, 3
+LIMIT 5;
+ user_id | value_1 | event_type 
+---------+---------+------------
+       1 |       1 |          0
+       1 |       1 |          0
+       1 |       1 |          1
+       1 |       1 |          1
+       1 |       1 |          2
+(5 rows)
+
 -- inner joins on reference tables with functions works
 SELECT DISTINCT ON (t1.user_id) t1.user_id, t2.value_1, t2.value_2, t2.value_3
 FROM events_table t1
--- a/src/test/regress/expected/non_colocated_subquery_joins.out
+++ b/src/test/regress/expected/non_colocated_subquery_joins.out
@@ -957,7 +957,7 @@ $$);
 DEBUG:  skipping recursive planning for the subquery since it contains references to outer queries
 DEBUG:  skipping recursive planning for the subquery since it contains references to outer queries
 DEBUG:  skipping recursive planning for the subquery since it contains references to outer queries
-ERROR:  cannot perform distributed planning on this query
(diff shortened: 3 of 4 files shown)


serprex:
I've pin pointed the error as having been introduced by 1656b519c49062dee5bb5fc8174963821dda9f17 so pinging @marcoslot in case they can shed some more light on how best to move forward

src/backend/distributed/planner/query_pushdown_planning.c
```diff
@@ -141,6 +141,10 @@ ShouldUseSubqueryPushDown(Query *originalQuery, Query *rewrittenQuery)
 	{
 		return true;
 	}
+	if (FindNodeCheck((Node *) rewrittenQuery->jointree, IsOuterJoinExpr))
```
serprex: So this is nitpicking, but I'd like to get feedback on how well I'm emulating Citus code style

If I were writing this for my own project, I'd've combined this 2nd `FindNodeCheck` into the preceding condition with `||`. My guess is either is acceptable here, but which way is preferable?

onderkalaci:
> Aside: to backport this to 8.2 do I make a pull request to release-8.2?

We typically add the `backport` label to the PR (as I did now), and the person who is managing the release generally cherry-picks the necessary commits to the related branches, so you don't need to do it now.

src/backend/distributed/planner/query_pushdown_planning.c
onderkalaci: I'd also go with `||`. But, to make the reason more explicit, I'd probably do something like [here](https://github.com/citusdata/citus/compare/fix_outer_join_issue?expand=1) where the reader can understand why we're doing this hack.

To give some background information, in #1848 we discuss this [item](https://github.com/citusdata/citus/issues/1848#issuecomment-350216272) as well. We had some existing planner code-paths which rely on `rewrittenQuery ` and some newer planner code-paths which rely `originalQuery `.


And the commit you've mentioned (1656b51) actually moves outer logic planning logic from the existing/older planning code-paths to one of the newer code-path. And, we've been beaten by the technical debt with this bug.

I'd also recommend you to have a quick look to PostgreSQL's [`pull_up_subqueries`](https://doxygen.postgresql.org/prepjointree_8c.html#a9634db36695e0a4eb89c892472dfaf56) function to understand what's the distinction between (that we're mentioning here) about pulling up the subquery or not.

Basically, if the subquery is a very simple one defined [here](https://doxygen.postgresql.org/prepjointree_8c.html#a133463f44522aaed9274d3ffe811e8da), we end-up with this discrepancy between the original query and the re-written query. 

For example, if you add something that'd fail on `simple_subquery()`, Citus can plan the same query -- see `random()`, which is a volatile function and make the query !simple_subquery():
```SQL
select * from (select *, random() from test_a a left join test_b b on a.id=b.id)lo;
```

marcocitus: I think the conditions are separate primarily to add separate comments on each of them, e.g. it would be worth explaining why we need to check `rewrittenQuery` above this condition.

marcocitus:
> so pinging @marcoslot in case they can shed some more light on how best to move forward

I think the fix makes sense. 

The fact that we're still dealing with the rewrittenQuery is annoying and prone to bugs. It would be good to do some refactoring to reduce our reliance on `rewrittenQuery`, but in any case that would not be something we would want to backport. 

For context: Subquery pushdown can handle subqueries, but not re-partition joins, and needs `originalQuery` to be able to deparse the original subquery (rewritten queries cannot be deparsed), while the logical planner can handle re-partition joins and does not directly deparse the original query, but cannot handle "complex" subqueries, and relies on `rewrittenQuery` to handle re-partition joins in very simple subqueries.

marcocitus:
> and the person who is managing the release generally cherry-picks the necessary commits to the related branches

To be honest, it's a bit nicer if the PR author does this after merging to be able to resolve conflicts.

src/backend/distributed/planner/query_pushdown_planning.c
pykello: @onderkalaci I talked with @serprex and he was concerned that irrespective of what the reason was for this situation (original query didn't have an outer join but the rewritten one had an outer join), we should push down. Subquery pull-up can be one of the reasons, but are we sure this is the only one?

Checking `subqueryPulledUp && FindNodeCheck((Node *) rewrittenQuery, IsOuterJoinExpr)` raises the question of why we checked for `subqueryPulledUp`? Why shouldn't we push-down when `!subqueryPulledUp`? Our goal might have been making things more explicit, but it might raise new questions which might add to the confusion.

So we thought Philip's original change with some comments to make it make it more explicit.

Does the current changes look good to you?

pykello: Put it another way, even if "subquery pull-up" is the only case where the original query doesn't have an outer join node but the rewritten one has, this is not clear from `subqueryPulledUp && FindNodeCheck((Node *) rewrittenQuery, IsOuterJoinExpr) `. For someone without much context of this discussion, it might read like "subquery pull up is not the only case, and if we shouldn't push-down when we have an outer join in the original query and no subquery pull-up", which is not what we wanted to mean from what I understand.

marcocitus: I think the comment is sufficient explanation. You could maybe add an `Assert(JoinTreeContainsSubquery(originalQuery))` to confirm our understanding.

PR #2746, opened 2019-06-02
insert_select_executor goes through unified executor.

pykello:
This is a step towards COPY through executor.

### DIFF
--- a/src/backend/distributed/executor/executor.c
+++ b/src/backend/distributed/executor/executor.c
@@ -47,8 +47,13 @@ typedef struct DistributedExecution
 	/* distributed query plan */
 	DistributedPlan *plan;
 
-	/* custom scan state */
-	CitusScanState *scanState;
+	/* Parameters for parameterized plans. Can be NULL. */
+	ParamListInfo paramListInfo;
+
+	/* Tuple descriptor and destination for result. Can be NULL. */
+	TupleDesc tupleDescriptor;
+	Tuplestorestate *tupleStore;
+
 
 	/* list of workers involved in the execution */
 	List *workerList;
@@ -318,7 +323,9 @@ bool ForceMaxQueryParallelization = false;
 
 /* local functions */
 static DistributedExecution * CreateDistributedExecution(DistributedPlan *distributedPlan,
-														 CitusScanState *scanState,
+						   								 ParamListInfo paramListInfo,
+														 TupleDesc tupleDescriptor,
+														 Tuplestorestate *tupleStore,
 														 int targetPoolSize);
 static void StartDistributedExecution(DistributedExecution *execution);
 static void RunDistributedExecution(DistributedExecution *execution);
@@ -373,7 +380,10 @@ CitusExecScan(CustomScanState *node)
 	{
 		DistributedPlan *distributedPlan = scanState->distributedPlan;
 		DistributedExecution *execution = NULL;
-		EState *executorState = scanState->customScanState.ss.ps.state;
+		EState *executorState = ScanStateGetExecutorState(scanState);
+		ParamListInfo paramListInfo = executorState->es_param_list_info;
+		Tuplestorestate *tupleStore = NULL;
+		TupleDesc tupleDescriptor = ScanStateGetTupleDescriptor(scanState);
 		bool randomAccess = true;
 		bool interTransactions = false;
 		int targetPoolSize = DEFAULT_POOL_SIZE;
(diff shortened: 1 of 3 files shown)


src/backend/distributed/executor/executor.c
```diff
@@ -1821,6 +1830,15 @@ ReceiveResults(WorkerSession *session, bool storeRows)
 			scanState->customScanState.ss.ps.ps_ResultTupleSlot->tts_tupleDescriptor;
 		attributeInputMetadata = TupleDescGetAttInMetadata(tupleDescriptor);
 		expectedColumnCount = tupleDescriptor->natts;
+	
+		if (scanState->tuplestorestate == NULL)
+		{
+			bool randomAccess = true;
+			bool interTransactions = false;
+			scanState->tuplestorestate =
+				tuplestore_begin_heap(randomAccess, interTransactions, work_mem);
```
pykello: I've used the same argument values as `StoreQueryResult()` in `multi_router_executor.c`.

marcocitus: I've been wondering, is there a way to remove `ScanState` from the executor and instead keep a tuple store and descriptor in the `DistributedExecution`?

It would be useful to have a function that can just write to a tuple store supplied by the caller instead of having to supply a whole `ScanState`. That way, it becomes easier to call `ExecuteTaskListExtended` from other contexts (e.g. a UDF).

pykello: > I've been wondering, is there a way to remove `ScanState` from the executor and instead keep a tuple store and descriptor in the `DistributedExecution`?
> 
> It would be useful to have a function that can just write to a tuple store supplied by the caller instead of having to supply a whole `ScanState`. That way, it becomes easier to call `ExecuteTaskListExtended` from other contexts (e.g. a UDF).

I think that makes sense. I made the changes.

src/backend/distributed/executor/insert_select_executor.c
```diff
@@ -128,8 +128,20 @@ CoordinatorInsertSelectExecScan(CustomScanState *node)
 
 			if (prunedTaskList != NIL)
 			{
-				ExecuteMultipleTasks(scanState, prunedTaskList, isModificationQuery,
-									 hasReturning);
+				TupleDesc tupleDescriptor =
+					scanState->customScanState.ss.ps.ps_ResultTupleSlot->tts_tupleDescriptor;
+
+				if (scanState->tuplestorestate == NULL)
```
pykello: Can this be not-NULL here?

marcocitus: No, this is our own variable, assigned by `CoordinatorInsertSelectExecScan` itself.

src/backend/distributed/executor/executor.c
```diff
@@ -429,23 +441,49 @@ CitusExecScan(CustomScanState *node)
...
+uint64
+ExecuteTaskListExtended(CmdType operation, List *taskList,
+						TupleDesc tupleDescriptor, Tuplestorestate *tupleStore, 
+						bool hasReturning, int targetPoolSize)
 {
 	DistributedPlan *distributedPlan = NULL;
 	DistributedExecution *execution = NULL;
 
+	/* TODO: Should we promote this as an argument to ExecuteTaskListExtended()? */
+	ParamListInfo paramListInfo = NULL;
```
marcocitus: I think we can afford to wait until someone needs it.

src/backend/distributed/executor/insert_select_executor.c
```diff
@@ -128,8 +128,17 @@ CoordinatorInsertSelectExecScan(CustomScanState *node)
...
+				TupleDesc tupleDescriptor = ScanStateGetTupleDescriptor(scanState);
+				bool randomAccess = true;
+				bool interTransactions = false;
+
+				Assert(scanState->tuplestorestate == NULL);
+				scanState->tuplestorestate =
+					tuplestore_begin_heap(randomAccess, interTransactions, work_mem);
+
+				ExecuteTaskListExtended(CMD_INSERT, prunedTaskList,
+										tupleDescriptor, scanState->tuplestorestate,
```
onderkalaci: @pykello Could you please make sure that we sort the tuplestore as we do in other executors? After rebasing my changes, I get some insert/select queries to have indeterministic returnings. Thanks!

pykello: Sure, https://github.com/citusdata/citus/pull/2754

PR #2748, opened 2019-06-04
Refactor some scan state info into their own functions.

pykello:
Repeated enough times to deserve their own functions. Unified executor will add couple more calls to these functions.

### DIFF
--- a/src/backend/distributed/executor/citus_custom_scan.c
+++ b/src/backend/distributed/executor/citus_custom_scan.c
@@ -308,7 +308,7 @@ static void
 CitusReScan(CustomScanState *node)
 {
 	CitusScanState *scanState = (CitusScanState *) node;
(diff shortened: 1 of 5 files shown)


src/backend/distributed/executor/citus_custom_scan.c
```diff
@@ -318,3 +318,25 @@ CitusReScan(CustomScanState *node)
 							   "parameters are currently unsupported")));
 	}
 }
+
+
+/*
+ * ScanStateGetTupleDescriptor returns the tuple descriptor for the given
+ * scan state.
+ */
+extern TupleDesc
```
onderkalaci: Minor: We typically don't have `extern` qualifier on the source files.

pykello: Fixed. Thanks.

PR #2750, opened 2019-06-05
enable_statistics_collection defaults to off (opt-in)

marcocitus:
DESCRIPTION: citus.enable_statistics_collection defaults to off (opt-in)

Users can opt-in to statistics collection by setting `citus.enable_statistics_collection = on`. Package maintainers can choose to make it opt-out by using `-DENABLE_CITUS_STATISTICS_COLLECTION` during compilation.

### DIFF
--- a/src/backend/distributed/shared_library_init.c
+++ b/src/backend/distributed/shared_library_init.c
@@ -1073,13 +1073,13 @@ RegisterCitusConfigVariables(void)
 					 "and operating system name. This configuration value controls "
 					 "whether these reports are sent."),
 		&EnableStatisticsCollection,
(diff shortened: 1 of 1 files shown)


ozgune:
hey @marcocitus, @pykello -- I have a quick question. After we disable anonymous stats collection, I don't expect users to go through the effort of enabling it back on. Should we just remove code related to stats collection from Citus? This way, we'd have less code and dependencies to maintain.

We always have the version history. If we want to look at the implementation later, we can always refer to the code history in Git.

marcocitus:
We should consider whether there are particular places where we still want it (e.g. enterprise?), but probably makes sense.

Issue #2749, opened 2019-06-05
update mx_table  error  (anchorShardId was assigned ref table's shardid, but should mx_table'shardid)


ifquant:
Reproduction script:   2 worker node, 1 cn node

```sql
SET citus.shard_replication_factor TO 1;
SET citus.replication_model TO streaming;

CREATE TABLE reference_raw_table1 (id bigint, value bigint);
SELECT create_reference_table('reference_raw_table1');
INSERT INTO reference_raw_table1 VALUES (1, 100);

CREATE TABLE summary_table1 (
    id bigint,
    min_value numeric,
    average_value numeric,
    count int,
    uniques int);

SELECT create_distributed_table('summary_table1', 'id');
INSERT INTO summary_table1 VALUES (1);

UPDATE summary_table1 SET average_value = average_query.average FROM (
    SELECT avg(value) AS average FROM reference_raw_table1 WHERE id = 1
    ) average_query
WHERE id = 1;

ERROR:  relation "public.summary_table1_540067" does not exist
CONTEXT:  while executing command on localhost:57638
```

code below may have problem;
```c
PlanRouterQuery(Query *originalQuery,
				PlannerRestrictionContext *plannerRestrictionContext,
				List **placementList, uint64 *anchorShardId, List **relationShardList,
				bool replacePrunedQueryWithDummy, bool *multiShardModifyQuery,
				Const **partitionValueConst)
{
    shardId = GetInitialShardId(prunedRelationShardList);     // reference_raw_table1 shardid
    workerList = WorkersContainingAllShards(prunedRelationShardList);
    *placementList = workerList;                                            // summary_table1  placement
    *anchorShardId = shardId;
}
```

in `CitusModifyBeginScan->FirstReplicaAssignTaskList`, `task->placementList` will be replaced with
`reference_raw_table1`'s 2 placement, then citus will send the update sql to two worker,
but in fact, `public.summary_table1_540067` not existed in another worker.

Reproduction script is grabbed from test sql `multi_modifications.sql` . but the test could work because the `shard_replication_factor` is 2.

-----

below is log_remote_command:

```sql
BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED;SELECT assign_distributed_transaction_id                              work1
BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED;SELECT assign_distributed_transaction_id                              work2
[...]

PR #2751, opened 2019-06-06
add case for issue 2749

ifquant:
https://github.com/citusdata/citus/issues/2749#

### DIFF
--- a/src/test/regress/sql/multi_task_assignment_policy.sql
+++ b/src/test/regress/sql/multi_task_assignment_policy.sql
@@ -261,5 +261,43 @@ SELECT count(DISTINCT value) FROM explain_outputs;
 RESET citus.task_assignment_policy;
 RESET client_min_messages;
 
+--  make sure we set the right anchor shardid for task
+-- citus/issues/2749 , anchor_shardId should be assigned as dis_table'shardid 
+SET citus.shard_replication_factor TO 1;
+SET citus.replication_model TO streaming;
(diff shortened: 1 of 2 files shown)


src/test/regress/sql/multi_task_assignment_policy.sql
```diff
@@ -261,5 +261,43 @@ SELECT count(DISTINCT value) FROM explain_outputs;
...
 RESET client_min_messages;
 
+--  make sure we set the right anchor shardid for task
+-- citus/issues/2749 , anchor_shardId should be assigned as dis_table'shardid 
+SET citus.shard_replication_factor TO 1;
+SET citus.replication_model TO streaming;
+CREATE TABLE reference_table (id bigint, value bigint);
+SELECT create_reference_table('reference_table');
+
+CREATE TABLE dis_table (
```
onderkalaci: We typically try to create/drop as less distributed tables as possible in the regression tests, mostly to prevent the regression test times keep increasing.

So, instead of creating new tables, could you please use the existing tables: instead of `dist_table` could use `task_assignment_nonreplicated_hash ` and instead of `reference_table` could use `task_assignment_reference_table `. 

Thanks!

ifquant:
because the schema of reference_table  is not  equal to  task_assignment_reference_table,
it's not convenient to reuse the task_assignment_reference_table.  total sql will be changed ...,

PR #2752, opened 2019-06-06
Add executor locks to unified executor

marcocitus:
This change adds the executor locks to prevent concurrent non-commutative changes to replicas, and avoid deadlocks between multi-shard DML to the unified executor.

### DIFF
--- a/src/backend/distributed/executor/executor.c
+++ b/src/backend/distributed/executor/executor.c
@@ -361,6 +361,7 @@ static void FinishDistributedExecution(DistributedExecution *execution);
 static void CleanUpSessions(DistributedExecution *execution);
 
 static void LockPartitionsForDistributedPlan(DistributedPlan *distributedPlan);
+static void AcquireExecutorShardLocks(DistributedExecution *execution);
 static bool DistributedExecutionModifiesDatabase(DistributedExecution *execution);
 static bool DistributedPlanModifiesDatabase(DistributedPlan *plan);
 static bool TaskListModifiesDatabase(CmdType operation, List *taskList);
@@ -632,6 +633,9 @@ StartDistributedExecution(DistributedExecution *execution)
 	{
 		RecordParallelRelationAccessForTaskList(taskList);
 	}
+
+	/* prevent unsafe concurrent modifications */
+	AcquireExecutorShardLocks(execution);
 }
 
 
@@ -772,6 +776,57 @@ LockPartitionsForDistributedPlan(DistributedPlan *distributedPlan)
 	 * have a stronger lock this doesn't do any harm.
 	 */
 	LockPartitionsInRelationList(distributedPlan->relationIdList, AccessShareLock);
+}
+
+
+/*
+ * AcquireExecutorShardLocks acquires advisory lock on shard IDs to prevent
+ * unsafe concurrent modifications of shards.
+ *
+ * We prevent concurrent modifications of shards in two cases:
+ * 1. Any non-commutative writes to a replicated table
+ * 2. Multi-shard writes that are executed in parallel
+ *
+ * The first case ensures we do not apply updates in different orders on
+ * different replicas (e.g. of a reference table), which could lead the
+ * replicas to diverge.
+ *
+ * The second case prevents deadlocks due to out-of-order execution.
+ *
+ * We do not take executor shard locks for utility commands such as
+ * TRUNCATE because the table locks already prevent concurrent access.
+ */
+static void
+AcquireExecutorShardLocks(DistributedExecution *execution)
+{
+	CmdType operation = execution->operation;
+	List *taskList = execution->tasksToExecute;
+
+	if (!(operation == CMD_INSERT || operation == CMD_UPDATE || operation == CMD_DELETE))
+	{
+		/* executor locks only apply to DML commands */
+		return;
+	}
+
+	/*
+	 * When executing in sequential mode or only executing a single task, we
+	 * do not need multi-shard locks.
+	 */
+	if (list_length(taskList) == 1 || ShouldRunTasksSequentially(taskList))
+	{
+		ListCell *taskCell = NULL;
+
+		foreach(taskCell, taskList)
+		{
+			Task *task = (Task *) lfirst(taskCell);
+
+			AcquireExecutorShardLock(task, operation);
+		}
+	}
+	else if (list_length(taskList) > 1)
+	{
+		AcquireExecutorMultiShardLocks(taskList);
+	}
 }
 
 
--- a/src/backend/distributed/executor/executor.c
+++ b/src/backend/distributed/executor/executor.c
@@ -697,21 +697,35 @@ DistributedExecutionRequiresRollback(DistributedExecution *execution)
 {
 	CmdType operation = execution->operation;
 	List *taskList = execution->tasksToExecute;
+	int taskCount = list_length(taskList);
 	Task *task = NULL;
+	bool selectForUpdate = false;
 
 	if (MultiShardCommitProtocol == COMMIT_PROTOCOL_BARE)
 	{
 		return false;
 	}
 
-	if (operation == CMD_SELECT)
-	{
-		return SelectOpensTransactionBlock && IsTransactionBlock();
-	}
-
-	if (taskList == NIL)
+	if (taskCount == 0)
 	{
 		return false;
+	}
+
+	task = (Task *) linitial(taskList);
+
+	selectForUpdate = task->relationRowLockList != NIL;
+	if (selectForUpdate)
+	{
+		/*
+		 * Do not check SelectOpensTransactionBlock, always open transaction block
+		 * if SELECT FOR UPDATE is executed inside a distributed transaction.
+		 */
+		return IsTransactionBlock();
+	}
+
+	if (ReadOnlyTask(task->taskType))
+	{
+		return SelectOpensTransactionBlock && IsTransactionBlock();
 	}
 
 	if (IsMultiStatementTransaction())
--- a/src/backend/distributed/executor/executor.c
+++ b/src/backend/distributed/executor/executor.c
@@ -366,6 +366,7 @@ static bool DistributedExecutionModifiesDatabase(DistributedExecution *execution
 static bool DistributedPlanModifiesDatabase(DistributedPlan *plan);
 static bool TaskListModifiesDatabase(CmdType operation, List *taskList);
 static bool DistributedExecutionRequiresRollback(DistributedExecution *execution);
+static bool SelectForUpdateOnReferenceTable(CmdType operation, List *taskList);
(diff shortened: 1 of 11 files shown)


src/backend/distributed/executor/executor.c
```diff
@@ -530,17 +530,47 @@ CreateDistributedExecution(DistributedPlan *distributedPlan,
...
  */
 void
 StartDistributedExecution(DistributedExecution *execution)
 {
 	DistributedPlan *distributedPlan = execution->plan;
+	CmdType operation = distributedPlan->operation;
 	Job *job = distributedPlan->workerJob;
 	List *taskList = job->taskList;
 
+	if (operation == CMD_INSERT || operation == CMD_UPDATE || operation == CMD_DELETE)
```
marcocitus: will change this to `DistributedPlanModifiesDatabase` once it is merged.

marcocitus:
Note that this currently causes multi-row INSERTs to block each other, but this will be resolved in #2735 .

onderkalaci:
I've tried to look deeper into the isolation test failures on top of this branch and all the PRs that are open against unified_executor. I think we're in a pretty good spot with this simple change:

| Failing Test  | Fix difficulty | Notes | 
| ------------- | ------------- | ------------- |
| isolation_concurrent_dml and isolation_insert_vs_all  |Simple | Failing due to multi row inserts, but note that in #2735 we do `StartExecution` once so it  this PR does not directly solve it |
| isolation_distributed_transaction_id  | Trivial  | Just update the transaction number in the tests |
|isolation_insert_select_conflict | Trivial| Just an RETURNING order issue, wait until #2754 merged |
| isolation_citus_dist_activity and isolation_get_distributed_wait_queries | Trivial |Two things: (a) Real-time worker query should be updated from COPY to SELECT  (b) Router SELECT with BEGIN opens transaction block now|
|isolation_dump_global_wait_edges| Medium | In the master branch  AssignDistributedTransactionId() precedes AcquireExecutorShardLock() where as it is different on this branch . I simply tried moving the lock acquisition right after `BeginOrContinueCoordinatedTransaction` and it worked. But need to think more  |


Apart from the above, I still see the following warnings, which I'll look a bit more details this time. `WARNING:  problem in alloc set MessageContext: detected write past chunk end in block 0x7facfb858200, chunk 0x7facfb859000
WARNING:  problem in alloc set MessageContext: detected write past chunk end in block 0x7facfb858200, chunk 0x7facfb859000`

src/backend/distributed/executor/executor.c
```diff
@@ -530,17 +530,47 @@ CreateDistributedExecution(DistributedPlan *distributedPlan,
...
+		 * 1. Any non-commutative writes to a replicated table
+		 * 2. Multi-shard writes that are executed in parallel
+		 *
+		 * The first case ensures we do not apply updates in different orders on
+		 * different replicas (e.g. of a reference table), which could lead the
+		 * replicas to diverge.
+		 *
+		 * The second case prevents deadlocks due to out-of-order execution.
+		 */ 
+		if (list_length(taskList) == 1)
```
onderkalaci: As I already noted, this doesn't immediately fixes multi row inserts on top of #2735 because the `StartDistributedExecution ` would qualify those s a multi-shard/multi-task operation.

src/backend/distributed/executor/executor.c
onderkalaci: Maybe move the whole logic to `AcquireExecutorLocks()` or something like that?

onderkalaci:
> Apart from the above, I still see the following warnings, which I'll look a bit more details this time. WARNING: problem in alloc set MessageContext: detected write past chunk end in block 0x7facfb858200, chunk 0x7facfb859000 WARNING: problem in alloc set MessageContext: detected write past chunk end in block 0x7facfb858200, chunk 0x7facfb859000

That might be related to skipping `make-clean` but still suspicious and still looking to reproduce it...

src/backend/distributed/executor/executor.c
onderkalaci: I think we should also add `|| operation == CMD_SELECT` because of `SELECT FOR UPDATE` on reference tables acquire exclusive locks (and in MX does some more stuff as well). I'm actually disappointed that none of the tests failed because of this.

I think that's because of using [replace_isolation_tester_func](https://github.com/citusdata/citus/pull/2197/files#diff-09efd54a1e80012e8973bb60022f196fR4) which switches sessions because of the locks on the workers, not on the coordinator. I think we need that for distributed tables since they do not acquire executor locks on `SELECT FOR UPDATE`, but, it also works for reference tables, which hides the missing executor locks on the coordinator.

A (semi)-related note: On the master branch, we're treating `select for update` on reference tables as modification in the sense that we're running the query on all the replicas. Shall we implement that in this PR? A test would be nice for that, we can use a function similar to [this](https://github.com/citusdata/citus/blob/master/src/test/regress/sql/sequential_modifications.sql#L16) to ensure that `SELECT` opens a transaction block.

onderkalaci:
> In the master branch AssignDistributedTransactionId() precedes AcquireExecutorShardLock() where as it is different on this branch . I simply tried moving the lock acquisition right after BeginOrContinueCoordinatedTransaction and it worked. But need to think more

It looks like we've discussed/fixed a similar thing https://github.com/citusdata/citus/pull/2452/files while back. I think `isolation_get_distributed_wait_queries` test is also failing due to the same thing.

src/backend/distributed/executor/executor.c
onderkalaci: Another note:

After merging onto a temp branch, I realized that one of the tests is failing [here](https://github.com/citusdata/citus/blob/master/src/test/regress/expected/multi_partitioning.out#L1467-L1499).

That is basically `TRUNCATE` on a partition should acquire locks on the parent test. 

I think the main reason is that we're not acquiring these locks on `TRUNCATE`. 

Once we change the check to `DistributedPlanModifiesDatabase ()` should fix that.  

Maybe we should also return true for `SELECT FOR UPDATE` in that function, or have a separate check. Just thinking about yet.

src/backend/distributed/executor/executor.c
```diff
@@ -779,6 +783,57 @@ LockPartitionsForDistributedPlan(DistributedPlan *distributedPlan)
...
+	{
+		/* executor locks only apply to DML commands */
+		return;
+	}
+
+	/*
+	 * When executing in sequential mode or only executing a single task, we
+	 * do not need multi-shard locks.
+	 */
+	if (list_length(taskList) == 1 || ShouldRunTasksSequentially(taskList))
```
onderkalaci: Self note to myself (and please confirm that my understanding is correct about our conversations). Later, once I compile all the behavioral changes in a document, I'd like to add this as well.

We're slightly changing two behaviours which are also related to here:

- When ` SET citus.multi_shard_modify_mode TO sequential`, the new executor executes the multi shard UPDATE/DELETEs in parallel across worker nodes. On the other hand,  router executor serializes the execution across worker nodes as well. Router executor's behaviour is not intended for multi shard DMLs, mostly it was simpler to implement at that time and we haven't considered this use-case useful ever (for multi-row INSERTs it was intended and we're already enforcing it here). 
- With router executor, we're not acquiring the `AcquireExecutorMultiShardLocks ` for the above case. But, now we're acquiring the locks. The reason is actually above. Because with this new behaviour, the user could get into distributed deadlocks across concurrent multi-shard DMLs.

But, given that running multi-shard DMLs with sequential mode is mostly** relevant to supporting foreign keys from distributed tables, we actually do not care much about it right now. We can revisit this idea if anyone complains about it.

Finally, with fkey to reference table case, the new approach would increase the performance in many cases and doesn't lead to any potential distributed deadlock. One super edge case is that if the executor switches to sequential mode and then a distributed DML comes in the transaction block, it'd acquire the executor locks and the other concurrent parallel DMLs on the same table might need to wait for the locks. But, that seems very unlikely situation. 


** we've not heard any user trying to execute multi-shard DMLs in sequential mode. In fact, that's not a good idea.

src/backend/distributed/executor/executor.c
```diff
@@ -779,6 +783,57 @@ LockPartitionsForDistributedPlan(DistributedPlan *distributedPlan)
...
+ * We do not take executor shard locks for utility commands such as
+ * TRUNCATE because the table locks already prevent concurrent access.
+ */
+static void
+AcquireExecutorShardLocks(DistributedExecution *execution)
+{
+	CmdType operation = execution->operation;
+	List *taskList = execution->tasksToExecute;
+
+	if (!(operation == CMD_INSERT || operation == CMD_UPDATE || operation == CMD_DELETE))
```
onderkalaci: I think `SELECT FOR UPDATE` on reference tables should go through `AcquireExecutorShardLock `.  So, we should probably not have this check.

But then `AcquireExecutorMultiShardLocks ` should not be called for `SELECTs` because it doesn't have any check for it.

onderkalaci: We could maybe just ignore SELECT FOR UPDATE in this PR, and have it via #2768

onderkalaci:
Two more greens, isolation test pass as of now!

ifquant:
some question: how distributed lock is  avoided below in ShouldRunTasksSequentially case.
we need sort the task by shardid?  i will do some test. but there is some complicated about 
AcquireExecutorShardLock, for example relationRowLockList,RequiresConsistentSnapshot
...,is there some case that we can not prevent distributed lock  while not in mx,just cn.

AcquireExecutorShardLocks {
		foreach(taskCell, taskList)
		{
			AcquireExecutorShardLock(task, operation);
		}
}

i am some pullzed by the design goal about the lock of citus. 
can we get the  guarantee from citus as below:
 A: run transcations  through just only one cn node.(not mx case)
 B: run transcations throught just originally postgres
if transcations will not cause dead lock in case B, it will also not case distributed
dead lock in case A.

ifquant:
below code is some strange. is there need change the order of return ?

  ExecutionOrderForTask
  {
  		if (operation == CMD_INSERT && !task->upsertQuery)
		{
			return EXECUTION_ORDER_SEQUENTIAL;
                       //return EXECUTION_ORDER_PARALLEL;
		}
		else
		{
			return EXECUTION_ORDER_PARALLEL;
                        //return EXECUTION_ORDER_SEQUENTIAL;
		}
  }

onderkalaci:
> some question: how distributed lock is avoided below in ShouldRunTasksSequentially case.
we need sort the task by shardid? 

As far as I can see, we already sort the tasks. 

`ShouldRunTasksSequentially()` returns true only for multi-row INSERTs. Those tasks are sorted via `RouterInsertTaskList()` -> `BuildRoutesForInsert()` -> `GroupInsertValuesByShardId()` -> `SortList()` -> `CompareInsertValuesByShardId()`.  So, the `taskList` that the executor gets is already sorted.

PR #2753, opened 2019-06-06
Remove Statistics Collection related code

pykello:
DESCRIPTION: Removes statistics collection.


src/backend/distributed/citus--8.2-2--8.2-3.sql
```diff
@@ -0,0 +1,3 @@
+/* citus--8.2-2--8.2-3.sql */
+
+DROP FUNCTION IF EXISTS pg_catalog.citus_server_id();
```
pykello: We also need to drop `pg_dist_node_metadata`, but currently few other tools (docker, cloud, etc.) depend on existence of that. We will remove it after updating those.

ozgune:
hey @pykello -- Minor question.

Let's say I'm a user who previously disabled anonymous stats collection (through GUC). If I upgrade to Citus 8.3 (and this guc is still in my config file), what's my upgrade experience?

pykello:
> hey @pykello -- Minor question.
> 
> Let's say I'm a user who previously disabled anonymous stats collection (through GUC). If I upgrade to Citus 8.3 (and this guc is still in my config file), what's my upgrade experience?

Good question. It will emit a warning in server logs when starting the server:

```
WARNING:  unrecognized configuration parameter "citus.enable_statistics_collection"
```

Nothing in client logs though. It is also documented in https://www.postgresql.org/docs/11/runtime-config-custom.html:

> Because custom options may need to be set in processes that have not loaded the relevant extension module, PostgreSQL will accept a setting for any two-part parameter name. Such variables are treated as placeholders and have no function until the module that defines them is loaded. When an extension module is loaded, it will add its variable definitions, convert any placeholder values according to those definitions, and issue warnings for any unrecognized placeholders that begin with its extension name.

marcocitus:
I feel like the decision to remove the code was a bit rushed. e.g. I'm a bit sad to the see the libcurl dependency go, since keeping it would make it easier to experiment with things like blob storage integration.

metdos:
> I feel like the decision to remove the code was a bit rushed. e.g. I'm a bit sad to the see the libcurl dependency go, since keeping it would make it easier to experiment with things like blob storage integration.

Yes, let's not merge this until we get an OK from the product. There is an email thread on this.

PR #2756, opened 2019-06-07
Take partition locks correctly in unified executor

marcocitus:
Take RowExclusiveLock for DML. We previously took AccessShareLock for TRUNCATE, but PostgreSQL does not do this, removed it from the test.

### DIFF
--- a/src/backend/distributed/executor/executor.c
+++ b/src/backend/distributed/executor/executor.c
@@ -339,6 +339,7 @@ static void FinishDistributedExecution(DistributedExecution *execution);
 
 static bool DistributedPlanModifiesDatabase(DistributedPlan *plan);
 static bool DistributedStatementRequiresRollback(DistributedPlan *distributedPlan);
+static void LockPartitionsForDistributedPlan(DistributedPlan *distributedPlan);
 static void AssignTasksToConnections(DistributedExecution *execution);
 static void UnclaimAllSessionConnections(List *sessionList);
 static bool ShouldForceOpeningConnection(TaskPlacementExecutionState
@@ -400,8 +401,12 @@ CitusExecScan(CustomScanState *node)
 		bool interTransactions = false;
 		int targetPoolSize = DEFAULT_POOL_SIZE;
 
-		/* we are taking locks on partitions of partitioned tables */
-		LockPartitionsInRelationList(distributedPlan->relationIdList, AccessShareLock);
+		/*
+		 * PostgreSQL takes locks on all partitions in the executor. It's not entirely
+		 * clear why this is necessary (instead of locking the parent during DDL), but
+		 * We do the same for consistency.
+		 */
+		LockPartitionsForDistributedPlan(distributedPlan);
 
 		ExecuteSubPlans(distributedPlan);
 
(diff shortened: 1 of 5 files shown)


src/backend/distributed/executor/executor.c
```diff
@@ -388,8 +389,19 @@ CitusExecScan(CustomScanState *node)
...
 		int targetPoolSize = DEFAULT_POOL_SIZE;
 
-		/* we are taking locks on partitions of partitioned tables */
-		LockPartitionsInRelationList(distributedPlan->relationIdList, AccessShareLock);
+		/*
+		 * PostgreSQL takes locks on all partitions in the executor. It's not entirely
+		 * clear why this is necessary (instead of locking the parent during DDL), but
+		 * We do the same for consistency.
+		 */
+		if (distributedPlan->operation == CMD_SELECT)
```
marcocitus: @onderkalaci This is kind of a recurring theme. I wonder whether we should have a plan type e.g. `PLAN_READ` `PLAN_WRITE` `PLAN_SCHEMA_CHANGE` `PLAN_VACUUM`.

(here we can only have PLAN_READ and PLAN_WRITE)

src/backend/distributed/executor/executor.c
```diff
@@ -388,8 +390,12 @@ CitusExecScan(CustomScanState *node)
...
 		int targetPoolSize = DEFAULT_POOL_SIZE;
 
-		/* we are taking locks on partitions of partitioned tables */
-		LockPartitionsInRelationList(distributedPlan->relationIdList, AccessShareLock);
+		/*
+		 * PostgreSQL takes locks on all partitions in the executor. It's not entirely
+		 * clear why this is necessary (instead of locking the parent during DDL), but
+		 * We do the same for consistency.
+		 */
+		LockPartitionsForDistributedPlan(distributedPlan);
```
onderkalaci: Just to make sure: I think for DDLs we expect Postgres to acquire the locks on the utility hook so we don't care about them on master branch or here as well. Can you please confirm that?

src/backend/distributed/executor/executor.c
```diff
@@ -617,6 +623,24 @@ DistributedStatementRequiresRollback(DistributedPlan *distributedPlan)
...
+LockPartitionsForDistributedPlan(DistributedPlan *distributedPlan)
+{
+	if (distributedPlan->operation != CMD_SELECT)
+	{
+		Oid targetRelationId = distributedPlan->targetRelationId;
+
+		LockPartitionsInRelationList(list_make1_oid(targetRelationId), RowExclusiveLock);
+	}
+
+	LockPartitionsInRelationList(distributedPlan->relationIdList, AccessShareLock);
```
onderkalaci: Minor: Might be good to carry over the comment from the router executor for `AccessShareLock`s :
```
	/*
	 * We are taking locks on partitions of partitioned tables. These locks are
	 * necessary for locking tables that appear in the SELECT part of the query.
	 */
```

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -228,6 +228,7 @@ CreateModifyPlan(Query *originalQuery, Query *query,
 	distributedPlan->masterQuery = NULL;
 	distributedPlan->routerExecutable = true;
 	distributedPlan->hasReturning = false;
+	distributedPlan->targetRelationId = ExtractFirstDistributedTableId(query);
```
onderkalaci: This is probably not an issue, but if we do modifying CTEs or modifies with subqueries with router executor (or something fancier), would that always give the correct target relationId?

I wonder whether we should write a new function which gets the  `	rte = rt_fetch(query->resultRelation, query->rtable);  	Assert(rte->rtekind == RTE_RELATION); return rte->relid`  or such

src/backend/distributed/executor/executor.c
marcocitus: Correct.

PR #2758, opened 2019-06-11
Unified Executor compatible copy

pykello:
This doesn't work yet, just opening so others can see what kind of changes I have in mind.

Goals are:
1. Be compatible with unified-executor,
2. When using a pool-size >= shard count, behave same as old copy with the same performance.

Notes:
1. This can be against the master branch, but for testing purposes opening it against unified_executor branch first. After it is approved here, I'll open it against the master branch for final review.
2. Keeping changes to `CopyToNewShards()` for later, to keep the PR not too large.

### DIFF
--- a/src/backend/distributed/commands/multi_copy.c
+++ b/src/backend/distributed/commands/multi_copy.c
@@ -12,9 +12,15 @@
  * parses the master node copy options and handles communication with the master
  * node.
  *
- * It opens a new connection for every shard placement and uses the PQputCopyData
- * function to copy the data. Because PQputCopyData transmits data, asynchronously,
- * the workers will ingest data at least partially in parallel.
+ * If this is the first command in the transaction, we open a new connection for
+ * every shard placement. Otherwise we open as many connections as we can to
+ * not conflict with previous commands in transactions, in which case some shards
+ * may share connections. See the comments of CopyConnectionState for how we
+ * operate in that case.
+ *
+ * We use the PQputCopyData function to copy the data. Because PQputCopyData
+ * transmits data asynchronously, the workers will ingest data at least partially
+ * in parallel.
  *
  * For hash-partitioned tables, if it fails to connect to a worker, the master
  * marks the placement for which it was trying to open a connection as inactive,
@@ -70,6 +76,7 @@
 #include "distributed/multi_physical_planner.h"
 #include "distributed/multi_router_planner.h"
 #include "distributed/multi_shard_transaction.h"
+#include "distributed/multi_executor.h"
 #include "distributed/placement_connection.h"
 #include "distributed/relation_access_tracking.h"
 #include "distributed/remote_commands.h"
@@ -95,6 +102,86 @@ static const char BinarySignature[11] = "PGCOPY\n\377\r\n\0";
 
 /* use a global connection to the master node in order to skip passing it around */
 static MultiConnection *masterConnection = NULL;
+
+/*
+ * Data size threshold to switch over the active placement for a connection.
+ * If this is too low, overhead of starting COPY commands will hurt the
+ * performance. If this is too high, buffered data will use lots of memory.
+ * 4MB is a good balance between memory usage and performance. Note that this
+ * is irrelevant in the common case where we open one connection per placement.
+ */
+#define COPY_SWITCH_OVER_THRESHOLD (4 * 1024 * 1024)
+
+typedef struct CopyShardState CopyShardState;
+typedef struct CopyPlacementState CopyPlacementState;
+
+/*
+ * Multiple shard placements can share one connection. Each connection has one
+ * of those placements as the activePlacementState, and others in the
+ * bufferedPlacementList. When we want to send a tuple to a CopyPlacementState,
+ * we check if it is the active one in its connectionState, and in this case we
+ * directly put data on wire. Otherwise, we buffer it so we can put it on wire
+ * later, when copy ends or a switch-over happens. See CitusSendTupleToPlacements()
+ * for more details.
+ *
+ * This is done so we are compatible with executor.c. If a previous command
+ * in the current transaction has been executed using executor.c, then
+ * CopyGetPlacementConnection() might return the same connection for multiple
+ * placements. We support that case by the buffering mechanism described above.
+ *
+ * If no previous command in the current transaction has used executor.c, then
+ * CopyGetPlacementConnection() returns one connection per placement and no
+ * buffering happens and we put the copy data directly on connection.
+ */
+typedef struct CopyConnectionState
+{
+	/* Used as hash key. Equal to PQsocket(connection->pgConn). */
+	int socket;
+
+	MultiConnection *connection;
+
+	/*
+	 * Placement for which we have an active COPY going on over connection.
+	 * Can be NULL.
+	 */
+	CopyPlacementState *activePlacementState;
+
+	/*
+	 * Other placements that we are buffering data for. Later when a switch-over
+	 * happens, we remove an item from this list and set it to activePlacementState.
+	 * In this case, old activePlacementState isn't NULL, is added to this list.
+	 */
+	dlist_head bufferedPlacementList;
+} CopyConnectionState;
+
+
+struct CopyPlacementState
+{
+	/* Connection state to which the placemement is assigned to. */
+	CopyConnectionState *connectionState;
+
+	/* State of shard to which the placement belongs to. */
+	CopyShardState *shardState;
+
+	/*
+	 * Buffered COPY data. When the placement is activePlacementState of
+	 * some connection, this is empty. Because in that case we directly
+	 * send the data over connection.
+	 */
+	StringInfo data;
+
+	/* List node for CopyConnectionState->bufferedPlacementList. */
+	dlist_node bufferedPlacementNode;
+};
+
+struct CopyShardState
+{
+	/* Used as hash key. */
+	uint64 shardId;
+
+	/* List of CopyPlacementStates for all active placements of the shard. */
+	List *placementStateList;
+};
 
 
 /* Local functions forward declarations */
@@ -145,6 +232,27 @@ static bool CopyStatementHasFormat(CopyStmt *copyStatement, char *formatName);
 static bool IsCopyFromWorker(CopyStmt *copyStatement);
 static NodeAddress * MasterNodeAddress(CopyStmt *copyStatement);
 static void CitusCopyFrom(CopyStmt *copyStatement, char *completionTag);
+static HTAB * CreateConnectionStateHash(MemoryContext memoryContext);
+static HTAB * CreateShardStateHash(MemoryContext memoryContext);
+static CopyConnectionState * GetConnectionState(HTAB *connectionStateHash,
+												MultiConnection *connection);
+static CopyShardState * GetShardState(uint64 shardId, HTAB *shardStateHash,
+									  HTAB *connectionStateHash, bool stopOnFailure,
+									  bool *found);
+static MultiConnection * CopyGetPlacementConnection(ShardPlacement *placement,
+													bool stopOnFailure);
+static List * ConnectionStateList(HTAB *connectionStateHash);
+static void InitializeCopyShardState(CopyShardState *shardState,
+									 HTAB *connectionStateHash,
+									 uint64 shardId, bool stopOnFailure);
+static void StartPlacementStateCopyCommand(CopyPlacementState *placementState,
+										   CopyStmt *copyStatement,
+										   CopyOutState copyOutState);
+static void EndPlacementStateCopyCommand(CopyPlacementState *placementState,
+										 CopyOutState copyOutState);
+static void UnclaimCopyConnections(List *connectionStateList);
+static void ShutdownCopyConnectionState(CopyConnectionState *connectionState,
+										CitusCopyDestReceiver *copyDest);
 
 /* Private functions copied and adapted from copy.c in PostgreSQL */
 static void CopySendData(CopyOutState outputState, const void *databuf, int datasize);
@@ -828,15 +936,6 @@ OpenCopyConnections(CopyStmt *copyStatement, ShardConnections *shardConnections,
 		uint32 connectionFlags = FOR_DML;
 		StringInfo copyCommand = NULL;
 		PGresult *result = NULL;
-
-		/*
-		 * Make sure we use a separate connection per placement for hash-distributed
-		 * tables in order to allow multi-shard modifications in the same transaction.
-		 */
-		if (placement->partitionMethod == DISTRIBUTE_BY_HASH)
-		{
-			connectionFlags |= CONNECTION_PER_PLACEMENT;
-		}
 
 		connection = GetPlacementConnection(connectionFlags, placement, nodeUser);
 
@@ -2237,7 +2336,8 @@ CitusCopyDestReceiverStartup(DestReceiver *dest, int operation,
 	copyStatement->filename = NULL;
 	copyDest->copyStatement = copyStatement;
 
-	copyDest->shardConnectionHash = CreateShardConnectionHash(TopTransactionContext);
+	copyDest->shardStateHash = CreateShardStateHash(TopTransactionContext);
+	copyDest->connectionStateHash = CreateConnectionStateHash(TopTransactionContext);
 }
 
 
@@ -2262,8 +2362,8 @@ CitusCopyDestReceiverReceive(TupleTableSlot *slot, DestReceiver *dest)
 		 * We might be able to recover from errors with ROLLBACK TO SAVEPOINT,
 		 * so unclaim the connections before throwing errors.
 		 */
-		HTAB *shardConnectionHash = copyDest->shardConnectionHash;
-		UnclaimAllShardConnections(shardConnectionHash);
+		List *connectionStateList = ConnectionStateList(copyDest->connectionStateHash);
+		UnclaimCopyConnections(connectionStateList);
 
 		PG_RE_THROW();
 	}
@@ -2283,10 +2383,13 @@ CitusSendTupleToPlacements(TupleTableSlot *slot, CitusCopyDestReceiver *copyDest
 	TupleDesc tupleDescriptor = copyDest->tupleDescriptor;
 	CopyStmt *copyStatement = copyDest->copyStatement;
 
-	HTAB *shardConnectionHash = copyDest->shardConnectionHash;
+	CopyShardState *shardState = NULL;
 	CopyOutState copyOutState = copyDest->copyOutState;
 	FmgrInfo *columnOutputFunctions = copyDest->columnOutputFunctions;
 	CopyCoercionData *columnCoercionPaths = copyDest->columnCoercionPaths;
+	ListCell *placementStateCell = NULL;
+	bool cachedShardStateFound = false;
+	bool firstTupleInShard = false;
 
 	bool stopOnFailure = copyDest->stopOnFailure;
 
@@ -2294,9 +2397,6 @@ CitusSendTupleToPlacements(TupleTableSlot *slot, CitusCopyDestReceiver *copyDest
 	bool *columnNulls = NULL;
 
 	int64 shardId = 0;
-
-	bool shardConnectionsFound = false;
-	ShardConnections *shardConnections = NULL;
 
 	EState *executorState = copyDest->executorState;
 	MemoryContext executorTupleContext = GetPerTupleMemoryContext(executorState);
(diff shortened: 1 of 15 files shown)


src/backend/distributed/commands/multi_copy.c
```diff
@@ -2311,47 +2361,58 @@ CitusSendTupleToPlacements(TupleTableSlot *slot, CitusCopyDestReceiver *copyDest
...
+		}
 
-		/* send copy binary headers to shard placements */
-		if (copyOutState->binary)
+		if (placementState == connectionState->activePlacementState)
 		{
-			SendCopyBinaryHeaders(copyOutState, shardId,
-								  shardConnections->connectionList);
+			/* no need to buffer, put data directly on line */
+			SendCopyDataToPlacement(copyOutState->fe_msgbuf, shardId,
```
marcocitus: Note that when we just do the switchover the TCP buffer will probably not fit everything. In this case, the data will be buffered in the PGconn and we'll send the next batch the next time `PQflush` is called. We need to make sure the internal buffer doesn't grow too big. In the past, we've had many issues with COPY running out of memory and crashing for this reason.

src/backend/distributed/commands/multi_copy.c
```diff
@@ -2923,3 +2984,216 @@ CopyGetAttnums(TupleDesc tupDesc, Relation rel, List *attnamelist)
...
+
+static void
+EndPlacementStateCopyCommand(CopyPlacementState *placementState,
+							 CopyOutState copyOutState, bool binaryCopy)
+{
+	MultiConnection *connection = placementState->connectionState->connection;
+	uint64 shardId = placementState->shardState->shardId;
+
+	/* flush buffered data */
+	SendCopyDataToPlacement(placementState->data, shardId, connection);
```
marcocitus: We're not guaranteed to flush all the data that is buffered in the PGconn 's`outBuffer` here because we're using non-blocking sockets. We should be calling FinishConnectionIO to flush all the data, otherwise `pqEndcopy3` will fail.

pykello: The following `EndRemoteCopy()` calls `PutRemoteCopyEnd()` which calls `FinishConnectionIO()`.

src/backend/distributed/commands/multi_copy.c
pykello: `SendCopyDataToPlacement()` calls `PutRemoteCopyData()` which flushes the internal buffer by calling `FinishConnectionIO()` if internal buffer passes 8MB.

src/backend/distributed/commands/multi_copy.c
```diff
@@ -2923,3 +3041,308 @@ CopyGetAttnums(TupleDesc tupDesc, Relation rel, List *attnamelist)
...
+													NULL);
+	if (connection != NULL)
+	{
+		return connection;
+	}
+
+	/*
+	 * Make sure we use a separate connection per placement for hash-distributed
+	 * tables in order to allow multi-shard modifications in the same transaction.
+	 */
```
marcocitus: We should probably update the comment to say that we now do this to maximize performance for large COPY batches.

src/backend/distributed/commands/multi_copy.c
marcocitus: True, we should probably add a call to `FinishConnectionIO` to `PutRemoteCopyEnd` as well.

pykello: `PutRemoteCopyEnd()` already calls `FinishConnecitonIO()`.

src/backend/distributed/commands/multi_copy.c
```diff
@@ -2311,47 +2397,67 @@ CitusSendTupleToPlacements(TupleTableSlot *slot, CitusCopyDestReceiver *copyDest
...
+			if (connectionState->activePlacementState != NULL)
+			{
+				EndPlacementStateCopyCommand(connectionState->activePlacementState,
+											 copyOutState);
+				dlist_push_head(&connectionState->bufferedPlacementList,
+								&connectionState->activePlacementState->
+								bufferedPlacementNode);
+			}
+			StartPlacementStateCopyCommand(placementState, copyStatement,
+										   copyOutState);
```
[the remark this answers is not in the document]
pykello: `connectionState->activePlacementState = placementState` is part of the switch-over operation, but `StartPlacementStateCopyCommand` doesn't know anything about switch-over. It just starts a binary/non-binary COPY command on the connection. I think we can pass a connection and shardId instead of `placementState` to make this more explicit.

Also, I think having all of the logic for switch-over helps readability.

src/backend/distributed/commands/multi_copy.c
```diff
@@ -96,6 +96,79 @@ static const char BinarySignature[11] = "PGCOPY\n\377\r\n\0";
...
+ *
+ * If no previous command in the current transaction has used executor.c, then
+ * CopyGetPlacementConnection() returns one connection per placement and no
+ * buffering happens and we put the copy data directly on connection.
+ */
+typedef struct CopyConnectionState
+{
+	/* Used as hash key. Equal to PQsocket(connection->pgConn). */
+	int socket;
+	MultiConnection *connection;
```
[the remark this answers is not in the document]
pykello: done

src/backend/distributed/commands/multi_copy.c
pykello: done

src/backend/distributed/commands/multi_copy.c
```diff
@@ -96,6 +96,79 @@ static const char BinarySignature[11] = "PGCOPY\n\377\r\n\0";
 /* use a global connection to the master node in order to skip passing it around */
 static MultiConnection *masterConnection = NULL;
 
+/* data size threshold to switch over the active placement for a connection */
+#define COPY_SWITCH_OVER_THRESHOLD 4194304
```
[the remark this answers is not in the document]
pykello: done

src/backend/distributed/commands/multi_copy.c
```diff
@@ -2923,3 +3041,308 @@ CopyGetAttnums(TupleDesc tupDesc, Relation rel, List *attnamelist)
...
+ * a connection has already been assigned the placement in the current transaction
+ * then it reuses the connection. Otherwise, it requests a connection for placement.
+ */
+static MultiConnection *
+CopyGetPlacementConnection(ShardPlacement *placement)
+{
+	MultiConnection *connection = NULL;
+	uint32 connectionFlags = FOR_DML;
+	char *nodeUser = CurrentUserName();
+	List *placementAccessList = NIL;
```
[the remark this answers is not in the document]
pykello: removed

src/backend/distributed/commands/multi_copy.c
```diff
@@ -96,6 +96,79 @@ static const char BinarySignature[11] = "PGCOPY\n\377\r\n\0";
 /* use a global connection to the master node in order to skip passing it around */
```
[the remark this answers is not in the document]
pykello: Makes sense. I'll update this after fixing other parts of the PR in few days, in case we make changes to the logic.

src/test/regress/bin/diff
```diff
@@ -12,7 +12,8 @@
 
 file1="${@:(-2):1}"
 file2="${@:(-1):1}"
-test=$(basename $file1 .out)
+# $file1 is the expected file, which can have _0 suffix, so we use results file
+test=$(basename $file2 .out)
```
pykello: I opened a PR for this: https://github.com/citusdata/citus/pull/2765

src/test/regress/bin/diff
```diff
@@ -12,7 +12,10 @@
 
 file1="${@:(-2):1}"
 file2="${@:(-1):1}"
-test=$(basename $file1 .out)
+# $file1 is the expected file, which can have _0 suffix, so we use results file
+# as the test name. Passing it through sed to remove _0 in case in future order
+# of args are swapped.
+test=$(basename $file2 .out | sed "s/_0//")
```
pykello: I opened a PR for this: #2765

src/backend/distributed/commands/multi_copy.c
```diff
@@ -2923,3 +3070,380 @@ CopyGetAttnums(TupleDesc tupDesc, Relation rel, List *attnamelist)
...
+/*
+ * GetConnectionState finds existing CopyConnectionState for a connection in the
+ * provided hash. If not found, then a default structure is returned.
+ */
+static CopyConnectionState *
+GetConnectionState(HTAB *connectionStateHash, MultiConnection *connection)
+{
+	CopyConnectionState *connectionState = NULL;
+	bool found = false;
+	int socket = PQsocket(connection->pgConn);
```
onderkalaci: Minor: Maybe move this line inside `!found` block.

src/backend/distributed/commands/multi_copy.c
```diff
@@ -2311,46 +2408,84 @@ CitusSendTupleToPlacements(TupleTableSlot *slot, CitusCopyDestReceiver *copyDest
 	/* connections hash is kept in memory context */
 	MemoryContextSwitchTo(copyDest->memoryContext);
 
-	/* get existing connections to the shard placements, if any */
-	shardConnections = GetShardHashConnections(shardConnectionHash, shardId,
-											   &shardConnectionsFound);
-	if (!shardConnectionsFound)
+	shardState = GetShardState(shardId, copyDest->shardStateHash,
```
onderkalaci: Minor: Instead of passing `found `, can't we use `shardState->placementStateList != NIL`? This is slightly hard to follow

src/backend/distributed/commands/multi_copy.c
```diff
@@ -2923,3 +3070,380 @@ CopyGetAttnums(TupleDesc tupDesc, Relation rel, List *attnamelist)
...
+	}
+
+	/*
+	 * For placements that haven't been assigned a connection by a previous command
+	 * in the current transaction, we use a separate connection per placement for
+	 * hash-distributed tables in order to get the maximum performance.
+	 */
+	if (placement->partitionMethod == DISTRIBUTE_BY_HASH)
+	{
+		connectionFlags |= CONNECTION_PER_PLACEMENT;
```
onderkalaci: I guess we should not use connection per placement for sequential mode. We're setting the flag, but, we're still using the same connection to COPY multiple shards. 

Probably not a big deal, just while reading that I felt a bit unusual.

src/backend/distributed/commands/multi_copy.c
```diff
@@ -2923,3 +3070,380 @@ CopyGetAttnums(TupleDesc tupDesc, Relation rel, List *attnamelist)
...
+	return connection;
+}
+
+
+/*
+ * GetWorkerConnectionIfCached finds a connection for the given hostname/port from
+ * the given connectionStateHash. If not found, it returns NULL.
+ */
+static MultiConnection *
+GetWorkerConnectionIfCached(const char *nodeName, int nodePort, HTAB *connectionStateHash)
```
onderkalaci: I think instead of this trick, we could consider using `GetPlacementListConnection()` without `CONNECTION_PER_PLACEMENT `. In that case, we should get the same connection and have the placement connection associations done. That's how sequential router executor does.

marcocitus: Note that the connection is also claimed exclusively, which means GetPlacementListConnection will not return it again.

src/backend/distributed/commands/multi_copy.c
```diff
@@ -2923,3 +3070,380 @@ CopyGetAttnums(TupleDesc tupDesc, Relation rel, List *attnamelist)
...
+			continue;
+		}
+
+		connectionState = GetConnectionState(connectionStateHash, connection);
+		placementState = palloc0(sizeof(CopyPlacementState));
+		placementState->shardState = shardState;
+		placementState->data = makeStringInfo();
+		placementState->connectionState = connectionState;
+		dlist_push_head(&connectionState->bufferedPlacementList,
+						&placementState->bufferedPlacementNode);
```
marcocitus: When there is not yet an active placement, it's slightly confusing to me that we add the placement to `bufferedPlacementList` only to remove it right after.

Could we make the logic more explicit by adding something like:
```c
if (connectionState->activePlacementState == NULL)
{
     connectionState->activePlacementState = placementState;
}
else
{
    dlist_push_head(&connectionState->bufferedPlacementList, &placementState->bufferedPlacementNode);
}
```

src/backend/distributed/commands/multi_copy.c
```diff
@@ -2923,3 +3070,380 @@ CopyGetAttnums(TupleDesc tupDesc, Relation rel, List *attnamelist)
...
+													NULL);
+	if (connection != NULL)
+	{
+		return connection;
+	}
+
+	/* if execution mode set to sequential, reuse worker connection if available */
+	if (MultiShardConnectionType == SEQUENTIAL_CONNECTION)
+	{
+		connection = GetWorkerConnectionIfCached(placement->nodeName,
```
marcocitus: We also need to make sure `AssignPlacementListToConnection` is called.

src/backend/distributed/commands/multi_copy.c
marcocitus: For context: The CONNECTION_PER_PLACEMENT flag solved a specific problem that, after this PR, no longer really exists.

e.g. if distribution column values 1 and 2 are in different shards on the same node and we did not use CONNECTION_PER_PLACEMENT:
```sql
BEGIN;
INSERT INTO test VALUES (2,3);
\COPY test FROM STDIN WITH CSV
1,2 -- uses the already opened connection that was used for the insert
2,3 -- needs to use the connection that was used for the insert, but that's now in use
\.
```

The same problem existed for multi-shard queries and DML in the old executors. 

Going forward, we could consider removing the flag but it might still be beneficial to use it for COPY because it would avoid multi-plexing in some scenarios.

w.r.t. the sequential case I think neither using connection exclusively, nor the CONNECTION_PER_PLACEMENT flag could work.

src/backend/distributed/commands/multi_copy.c
```diff
@@ -2311,46 +2408,84 @@ CitusSendTupleToPlacements(TupleTableSlot *slot, CitusCopyDestReceiver *copyDest
...
+		bool switchOver = false;
 
-		/* send copy binary headers to shard placements */
-		if (copyOutState->binary)
+		if (placementState != connectionState->activePlacementState)
 		{
-			SendCopyBinaryHeaders(copyOutState, shardId,
-								  shardConnections->connectionList);
+			if (placementState->data->len > COPY_SWITCH_OVER_THRESHOLD ||
+				connectionState->activePlacementState == NULL)
```
marcocitus: if we make the change in `InitializeCopyShardState` then we can remove `connectionState->activePlacementState == NULL`

src/backend/distributed/commands/multi_copy.c
marcocitus: is it possible that `PQsocket` returns -1 when a connection has failed?

src/backend/distributed/commands/multi_copy.c
pykello: `shardState->placementStateList` is always initiated with the placement list when a `CopyShardState` created, so it is not NIL here.

I changed the names to make it a bit more readable.

src/backend/distributed/commands/multi_copy.c
```diff
@@ -2311,46 +2408,84 @@ CitusSendTupleToPlacements(TupleTableSlot *slot, CitusCopyDestReceiver *copyDest
...
-							copyOutState->binary);
+	foreach(placementStateCell, shardState->placementStateList)
+	{
+		CopyPlacementState *placementState = lfirst(placementStateCell);
+		CopyConnectionState *connectionState = placementState->connectionState;
+		bool switchOver = false;
 
-		/* send copy binary headers to shard placements */
-		if (copyOutState->binary)
+		if (placementState != connectionState->activePlacementState)
```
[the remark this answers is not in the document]
pykello: Your suggestion looks good. I had to make some modifications to your suggestion to make it correct. See if  the current version looks readable to you.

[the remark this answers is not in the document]
pykello: Your suggestion looks good. I had to make some modifications to your suggestion to make it correct. See if  the current version looks readable to you.

src/backend/distributed/commands/multi_copy.c
```diff
@@ -2440,30 +2575,42 @@ CitusCopyDestReceiverShutdown(DestReceiver *destReceiver)
...
+	connectionStateList = ConnectionStateList(connectionStateHash);
 
 	PG_TRY();
 	{
-		foreach(shardConnectionsCell, shardConnectionsList)
+		foreach(connectionStateCell, connectionStateList)
 		{
-			ShardConnections *shardConnections = (ShardConnections *) lfirst(
-				shardConnectionsCell);
+			CopyConnectionState *connectionState =
```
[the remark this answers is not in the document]
pykello: done.

src/backend/distributed/commands/multi_copy.c
pykello: > Minor: Maybe move this line inside `!found` block.

This is used in the `hash_search()` call before the `!found` block.

pykello: > is it possible that `PQsocket` returns -1 when a connection has failed?

Looking into this. If `GetPlacementListConnectionIfCached()` can return bad connections, then maybe, otherwise no. I will look into this.

src/backend/distributed/commands/multi_copy.c
pykello: > I think instead of this trick, we could consider using `GetPlacementListConnection()` without `CONNECTION_PER_PLACEMENT `. In that case, we should get the same connection and have the placement connection associations done. That's how sequential router executor does.

Removed this function and used the method you and @marcocitus suggested.

src/backend/distributed/commands/multi_copy.c
pykello: > We also need to make sure `AssignPlacementListToConnection` is called.

I removed `GetWorkerConnectionIfCached()`, so I guess calls to `AssignPlacementListToConnection` are now handled in placement_connection.c.

src/backend/distributed/commands/multi_copy.c
```diff
@@ -2311,46 +2408,99 @@ CitusSendTupleToPlacements(TupleTableSlot *slot, CitusCopyDestReceiver *copyDest
...
+		if (currentActivePlacementState == NULL)
 		{
-			SendCopyBinaryHeaders(copyOutState, shardId,
-								  shardConnections->connectionList);
+			switchOver = true;
+		}
+		else if (placementState != currentActivePlacementState &&
+				 placementState->data->len > COPY_SWITCH_OVER_THRESHOLD)
+		{
+			switchOver = true;
```
marcocitus: To clarify what we are switching to, maybe we can call this `switchToTargetShard` and rename `shardState` to something like `targetShardState`?

marcocitus: or `switchToCurrentShard` and `currentShardState` and rename `currentActivePlacementState` to `activePlacementState`.

src/backend/distributed/commands/multi_copy.c
pykello: > if we make the change in `InitializeCopyShardState` then we can remove `connectionState->activePlacementState == NULL`

The reason for initializing this to `NULL` was that we need to call `StartPlacementStateCopyCommand()` for the first placement state somewhere. I could either do this or check the value of `cachedShardStateFound`. I thought defining `activePlacementState` as "the placement for which the COPY has been started" makes sense, and COPY hasn't started before `StartPlacementStateCopyCommand()` is called, so setting it to NULL until here made more sense to me.

src/backend/distributed/commands/multi_copy.c
pykello: > is it possible that `PQsocket` returns -1 when a connection has failed?

Postgres docs say "This will not change during normal operation, but could change during connection setup or reset", so I think this shouldn't be -1 here. I added an assert.

src/backend/distributed/commands/multi_copy.c
pykello: > or `switchToCurrentShard` and `currentShardState` and rename `currentActivePlacementState` to `activePlacementState`.

done.

src/backend/distributed/commands/multi_copy.c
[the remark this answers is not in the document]
pykello: updated file comment.

src/backend/distributed/commands/multi_copy.c
```diff
@@ -2923,3 +3103,365 @@ CopyGetAttnums(TupleDesc tupDesc, Relation rel, List *attnamelist)
...
+ * adds them to shardState->placementStateList.
+ */
+static void
+InitializeCopyShardState(CopyShardState *shardState,
+						 HTAB *connectionStateHash, uint64 shardId,
+						 bool stopOnFailure)
+{
+	List *finalizedPlacementList = NIL;
+	ListCell *placementCell = NULL;
+	bool failedPlacementCount = 0;
```
onderkalaci: this should be int

src/backend/distributed/commands/multi_copy.c
```diff
@@ -96,6 +103,80 @@ static const char BinarySignature[11] = "PGCOPY\n\377\r\n\0";
 /* use a global connection to the master node in order to skip passing it around */
 static MultiConnection *masterConnection = NULL;
 
+/* data size threshold to switch over the active placement for a connection */
+#define COPY_SWITCH_OVER_THRESHOLD (4 * 1024 * 1024)
```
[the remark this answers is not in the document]
marcocitus: It'll probably just become something like `citus.max_assign_task_batch_size`, which no one every users.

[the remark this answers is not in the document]
onderkalaci: I think that makes sense, not adding the GUC.

But, maybe add a single line comment on why we pick 4MB as the threshold?

src/test/regress/bin/normalize.sed
```diff
@@ -13,11 +13,20 @@ s/shard [0-9]+/shard xxxxx/g
 
 # In foreign_key_to_reference_table, normalize shard table names, etc in
 # the generated plan
-s/"(fkey_ref_|referenced_table_|referencing_table_)[0-9]+"/"\1xxxxxxx"/g
+s/"(foreign_key_2_|fkey_ref_to_dist_|fkey_ref_)[0-9]+"/"\1xxxxxxx"/g
```
onderkalaci: I think we can also update `with_modifying.out`, there is only a minor error message update is missing.

src/backend/distributed/commands/multi_copy.c
```diff
@@ -829,15 +937,6 @@ OpenCopyConnections(CopyStmt *copyStatement, ShardConnections *shardConnections,
 		StringInfo copyCommand = NULL;
 		PGresult *result = NULL;
 
-		/*
```
onderkalaci: I think we should rename this function and update the function comment. I'd even add an `assert (placement->partitionMethod != DISTRIBUTE_BY_HASH)`. Otherwise it is becoming confusing to have two different ways of handling connection opening.

Maybe `OpenCopyConnectionsForNewShards`

src/test/regress/expected/with_modifying.out
```diff
@@ -193,9 +193,8 @@ WITH cte AS (
 	SELECT user_id, value_2 FROM users_table WHERE value_2 IN (1, 2)
 )
 INSERT INTO modify_table (SELECT (SELECT value_2 FROM cte GROUP BY value_2));
-WARNING:  more than one row returned by a subquery used as an expression
+ERROR:  more than one row returned by a subquery used as an expression
```
onderkalaci: Hmm, when you open the PR against master, do you think would the error message change back? Similar for the below test.

Anyway, I'll need to handle bunch of those things on #2762

onderkalaci:
> This doesn't work yet, just opening so others can see what kind of changes I have in mind.

Maybe update the PR description. I think copy & pasting some of the existing comments in the code would be fine

src/backend/distributed/commands/multi_copy.c
pykello: done

src/backend/distributed/commands/multi_copy.c
```diff
@@ -2311,46 +2414,98 @@ CitusSendTupleToPlacements(TupleTableSlot *slot, CitusCopyDestReceiver *copyDest
...
+		hash_get_num_entries(copyDest->shardStateHash) == 2)
+	{
+		Oid relationId = copyDest->distributedRelationId;
 
-			/* error out of conflicting COPY */
-			CheckConflictingParallelCopyAccesses(relationId);
+		/* mark as multi shard to skip doing the same thing over and over */
+		copyDest->multiShardCopy = true;
 
+		/* error out of conflicting COPY */
```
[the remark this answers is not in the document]
pykello: Still looking into see if this is already covered or do we need to add new tests ...

pykello:
Merged into master via https://github.com/citusdata/citus/pull/2780

PR #2762, opened 2019-06-11
Manually change the pool size and enable/disable the unified executor

onderkalaci:
This PR adds a new GUC `citus.max_pool_size` which controls the pool size. When set to 0, the unified executor is disabled and the existing executors is used.

The PR is far from its final shape, but still very useful to test the unified executor with different pool sizes and compare with the existing executor.

**I prefer to hold on this PR because there are quite a bit of open PRs that'd conflict with this.**

The todo list:

- [ ] Finalize the name(s) for the new executor
- [ ] Make sure that `ExecuteTaskList` also considers the case where the `max_pool_size = 0`. DDLs currently get stuck because they go through the unified executor with pool size = 0 :) . 
- [ ] Make sure that regression tests pass when `max_pool_size = 0` so that if we decide not to make the unified executor the default one
- [ ] `JobExecutorType()` is unnecessarily complicated.   
- [ ] `CoordinatorInsertSelect` related functions doesn't fit well into the picture. Try to make them look more native.


src/backend/distributed/shared_library_init.c
```diff
@@ -720,6 +720,18 @@ RegisterCitusConfigVariables(void)
 		GUC_UNIT_KB,
 		NULL, NULL, NULL);
 
+	DefineCustomIntVariable(
+		"citus.max_pool_size",
```
marcocitus: when we remove the old executors, what will be the meaning of 0?

onderkalaci: I guess the min value would be 1?

onderkalaci:
Closing this in favor of #2791

PR #2763, opened 2019-06-12
Always rebuild the wait event sets

onderkalaci:
In the previous implementation, on waitFlag changes, we were only
modifying the wait events via `ModifyWaitEvent `. However, we've 
realized that it might be an over optimization since 
    (a) we couldn't see any performance benefits 
    (b) we see some [errors](https://github.com/citusdata/citus/pull/2713#issuecomment-500875482) on failures and because of (a) we prefer to disable it now.

(As noted below, I couldn't observe any performance improvements with pgbench-select/tpcb/simple-update benchmarks).

### DIFF
--- a/src/backend/distributed/executor/executor.c
+++ b/src/backend/distributed/executor/executor.c
@@ -1632,7 +1632,12 @@ TransactionStateMachine(WorkerSession *session)
 					placementExecution = PopPlacementExecution(session);
 					if (placementExecution == NULL)
 					{
-						/* no tasks are ready to be executed at the moment */
+						/*
+						 * No tasks are ready to be executed at the moment. But we
+						 * still mark the socket readable to get any notices if exists.
+						 */
+						UpdateConnectionWaitFlags(session, WL_SOCKET_READABLE);
+
 						break;
 					}
 
--- a/src/backend/distributed/executor/executor.c
+++ b/src/backend/distributed/executor/executor.c
@@ -351,7 +351,6 @@ static WorkerSession * FindOrCreateWorkerSession(WorkerPool *workerPool,
 												 MultiConnection *connection);
 static void ManageWorkerPool(WorkerPool *workerPool);
 static WaitEventSet * BuildWaitEventSet(List *sessionList);
-static void UpdateWaitEventSetFlags(WaitEventSet *waitEventSet, List *sessionList);
 static TaskPlacementExecution * PopPlacementExecution(WorkerSession *session);
 static TaskPlacementExecution * PopAssignedPlacementExecution(WorkerSession *session);
 static TaskPlacementExecution * PopUnassignedPlacementExecution(WorkerPool *workerPool);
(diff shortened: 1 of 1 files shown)


src/backend/distributed/executor/executor.c
```diff
@@ -2427,8 +2458,8 @@ PlacementExecutionReady(TaskPlacementExecution *placementExecution)
 			 * If the connection is idle, wake it up by checking whether
 			 * the connection is writeable.
 			 */
-			connection->waitFlags = WL_SOCKET_READABLE | WL_SOCKET_WRITEABLE;
-			distributedExecution->waitFlagsChanged = true;
+			UpdateConnectionWaitFlags(session, WL_SOCKET_READABLE | WL_SOCKET_WRITEABLE,
+									  true);
```
marcocitus: the overwrite parameter is usually true. What if we remove the parameter and use `session->waitFlags | WL_SOCKET_***` in the non-overwrite cases?

src/backend/distributed/executor/executor.c
```diff
@@ -1646,8 +1643,9 @@ TransactionStateMachine(WorkerSession *session)
 					transaction->transactionState = REMOTE_TRANS_SENT_COMMAND;
```
marcocitus: Not changed by this PR, but I think we should be setting the wait flags to WL_SOCKET_READABLE above when there are no tasks to only check for notices.

src/backend/distributed/executor/executor.c
onderkalaci: Do you mean not calling `UpdateConnectionWaitFlags `? 

I think expecting future users to always keep in mind the decency between `connection->waitFlags ` and  `execution->waitFlagsChanged` is hard.  So, I'd still keep this to prevent bugs in the future, but I do not have strong opinions either. So, if you want, I can do that.

src/backend/distributed/executor/executor.c
onderkalaci: Shall I add a commit to this PR?

marcocitus: yes

src/backend/distributed/executor/executor.c
marcocitus: I mean remove the overwrite flag and in cases where you want to add to the existing flags call `UpdateConnectionWaitFlags(session, session->waitFlags | WL_SOCKET_READABLE)` instead.

src/backend/distributed/executor/executor.c
```diff
@@ -1757,19 +1755,51 @@ TransactionStateMachine(WorkerSession *session)
...
+	{
+		connection->waitFlags = waitFlags;
+	}
+	else
+	{
+		connection->waitFlags |= waitFlags;
+	}
+
+	/* without signalling the execution, the flag changes won't be reflected */
+	execution->waitFlagsChanged = true;
```
marcocitus: we should check whether the flags actually changed.

marcocitus:
> An unrelated thing we should figure it out is why we emit the message in the DETAIL twice:

Never mind, seems once for IPv4 and once for IPv6. Probably makes sense.

PR #2764, opened 2019-06-12
Alternative implementation for force_max_query_parallelization

marcocitus:
If `citus.force_max_query_parallelization` is enabled, `AssignTasksToConnections` currently opens connections via `StartPlacementListConnection` to assign the task to a new connection. This complicates the notion of assigning tasks to connections, which is normally only done to handle past accesses but now also done to force future accesses onto separate connections, and this has some undesirable side-effects. For instance, it breaks our ability to determine whether a connection modified a placement prior to a distributed execution using the `hadDML` flag.

This PR changes the handling of `citus.force_max_query_parallelization`. If `force_max_query_parallelization` is enabled and we are not in "sequential" mode:

1. Sessions do not take a new task from the `readyTaskQueue` of the `WorkerPool` if a command has previously been sent.
2. `ManageWorkerPool` increases the pool when there are more tasks in the `readyTaskQueue` of the worker pool than there are unused connections in the worker pool. An unused connection is one that has not sent a command yet.

The net effect is that each unassigned command will be sent over a separate connection, because we will open as many connection as we have commands to send and each connection will only send one command.

### DIFF
--- a/src/backend/distributed/executor/executor.c
+++ b/src/backend/distributed/executor/executor.c
@@ -167,6 +167,9 @@ typedef struct WorkerPool
 	 */
 	int idleConnectionCount;
 
+	/* number of connections that did not send a command */
+	int unusedConnectionCount;
+
 	/* number of failed connections */
 	int failedConnectionCount;
 
@@ -364,9 +367,7 @@ static bool TaskListModifiesDatabase(CmdType operation, List *taskList);
 static bool DistributedExecutionRequiresRollback(DistributedExecution *execution);
 static void AssignTasksToConnections(DistributedExecution *execution);
 static void UnclaimAllSessionConnections(List *sessionList);
-static bool ShouldForceOpeningConnection(TaskPlacementExecutionState
-										 placementExecutionState,
-										 PlacementExecutionOrder placementExecutionOrder);
+static bool UseConnectionPerPlacement(void);
 static PlacementExecutionOrder ExecutionOrderForTask(CmdType operation, Task *task);
 static WorkerPool * FindOrCreateWorkerPool(DistributedExecution *execution,
 										   WorkerNode *workerNode);
@@ -443,11 +444,6 @@ CitusExecScan(CustomScanState *node)
 		if (MultiShardConnectionType == SEQUENTIAL_CONNECTION)
 		{
 			targetPoolSize = 1;
-		}
-		else if (ForceMaxQueryParallelization)
-		{
-			/* we might at most have the length of tasks pool size */
-			targetPoolSize = list_length(taskList);
 		}
 
 		execution = CreateDistributedExecution(distributedPlan->operation, taskList,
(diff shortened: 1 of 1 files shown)


src/backend/distributed/executor/executor.c
```diff
@@ -1183,25 +1132,45 @@ ManageWorkerPool(WorkerPool *workerPool)
 		return;
 	}
 
-	if (currentConnectionCount >= targetPoolSize)
+	if (UseSeparateConnectionPerPlacement())
 	{
-		/* already reached the minimal pool size */
-		return;
+		int unusedConnectionCount = workerPool->unusedConnectionCount;
```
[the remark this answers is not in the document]
marcocitus: Makes sense, though as a counter point: it introduces a new constraint that `ManageWorkerPool` always finishes at once. In the future, we'll introduce things like 10ms space between connection attempts. We probably would not want to do that in the force case, but it'll make the code slightly harder to reason about if we make different assumptions in different code paths.

src/backend/distributed/executor/executor.c
```diff
@@ -401,14 +402,6 @@ CitusExecScan(CustomScanState *node)
 		{
 			targetPoolSize = 1;
 		}
-		else if (ForceMaxQueryParallelization)
```
onderkalaci: comment is not targeting this line, but more generic: Should we replace all `ForceMaxQueryParallelization` with  `UseSeparateConnectionPerPlacement` . We have 1-2 extra uses in the `unified_executor` branch

src/backend/distributed/executor/executor.c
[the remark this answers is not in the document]
marcocitus: Discussed this off-line, but this would complicate the logic for failover as we may suddenly have a greater number of ready tasks. It would also require a separate flag, because `initiatedConnectionCount` may be greater than 0 from the beginning if there is a separate task.

src/backend/distributed/executor/executor.c
```diff
@@ -1463,25 +1415,45 @@ ManageWorkerPool(WorkerPool *workerPool)
 		return;
```
onderkalaci: Not very much related with this PR.

I sometimes hit `Assert(activeConnectionCount >= idleConnectionCount);` if I randomly fail worker nodes while running SELECT queries with the flag on.

I think that's mainly because we decrement `activeConnectionCount ` on `MULTI_CONNECTION_LOST`. But, if the same connection is idle at that time, we do not decrement `idleConnectionCount`.

I think we should have probably have something like this (which fixes the issue I observe):
```
diff --git a/src/backend/distributed/executor/executor.c b/src/backend/distributed/executor/executor.c
index 66101df2..365580b1 100644
--- a/src/backend/distributed/executor/executor.c
+++ b/src/backend/distributed/executor/executor.c
@@ -1658,6 +1658,9 @@ ConnectionStateMachine(WorkerSession *session)
                                /* managed to connect, but connection was lost */
                                workerPool->activeConnectionCount--;
 
+                               if (session->currentTask == NULL)
+                                       workerPool->idleConnectionCount--;
+
                                connection->connectionState = MULTI_CONNECTION_FAILED;
                                break;
                        }
```

I can do that in another PR if you think it's not nice to have such thing here.

marcocitus: Let's do it in a separate PR.

PR #2765, opened 2019-06-13
Fix test name detection in bin/diff

pykello:
Noticed this while working on https://github.com/citusdata/citus/pull/2758

### DIFF
--- a/src/test/regress/bin/diff
+++ b/src/test/regress/bin/diff
@@ -12,7 +12,12 @@
 
 file1="${@:(-2):1}"
 file2="${@:(-1):1}"
-test=$(basename $file1 .out)
+# pg_regress passes the expected file as the first argument ($file1 above),
+# and results file as the second argument ($file2 above). We take the base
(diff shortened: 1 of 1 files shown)


src/test/regress/bin/diff
```diff
@@ -12,7 +12,10 @@
 
 file1="${@:(-2):1}"
 file2="${@:(-1):1}"
-test=$(basename $file1 .out)
+# $file1 is the expected file, which can have _0 suffix, so we use results file
```
onderkalaci: We also have an expected file with `custom_aggregate_support_1.out` and could potentially have more, so maybe use any integer regex instead of `_0`?

onderkalaci: And, also the sentence is a bit hard to follow. We're discussion only `file1 ` in the text, but sed uses `file2`. Is it a typo?

pykello: Changed the regex to match any sequence of digits, and updated the comment.

PR #2766, opened 2019-06-14
Make sure to allocate enough sized waitEventSet

onderkalaci:
With multi-row INSERTs, we might have more sessions than
task*workerCount after few calls of RunDistributedExecution()
because the previous sessions would also be alive.

Instead, re-allocate events when the connectino set changes.

### DIFF
--- a/src/backend/distributed/executor/executor.c
+++ b/src/backend/distributed/executor/executor.c
@@ -647,7 +647,6 @@ static bool
 DistributedExecutionModifiesDatabase(DistributedExecution *execution)
 {
 	return TaskListModifiesDatabase(execution->operation, execution->tasksToExecute);
-
 }
 
 
@@ -1310,8 +1309,6 @@ void
 RunDistributedExecution(DistributedExecution *execution)
 {
 	TimestampTz deadline = 0;
-	int workerCount = 0;
-	int maxWaitEventCount = 0;
 	WaitEvent *events = NULL;
 
 	deadline =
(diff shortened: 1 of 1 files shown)


src/backend/distributed/executor/executor.c
```diff
@@ -1350,18 +1341,28 @@ RunDistributedExecution(DistributedExecution *execution)
...
 				FreeWaitEventSet(execution->waitEventSet);
+
 				execution->waitEventSet = BuildWaitEventSet(execution->sessionList);
+
+				if (execution->connectionSetChanged)
+				{
+					/* recalculate since the sessions have changed */
+					eventSetSize = EventSetSizeForSessionList(execution->sessionList);
+
+					events = palloc0(eventSetSize * sizeof(WaitEvent));
```
marcocitus: I think we should do a `pfree` of events here.

src/backend/distributed/executor/executor.c
```diff
@@ -1418,7 +1419,10 @@ RunDistributedExecution(DistributedExecution *execution)
 		 */
 		UnclaimAllSessionConnections(execution->sessionList);
 
-		pfree(events);
+		/*
+		 * We don't need to (and probably shouldn't) free events because
+		 * it might have never been allocated.
+		 */
```
marcocitus: I don't think we need to explain that we don't do this, because I don't think the reader is necessarily expecting it. Freeing things is relatively unusual in postgres.

src/backend/distributed/executor/executor.c
```diff
@@ -1310,33 +1310,24 @@ void
...
-	/* additional 2 is for postmaster and latch */
-	maxWaitEventCount = execution->totalTaskCount * workerCount + 2;
-
-	/* allocate events for the maximum number of connections to avoid realloc */
-	events = palloc0(maxWaitEventCount * sizeof(WaitEvent));
-
 	PG_TRY();
 	{
 		bool cancellationReceived = false;
+		int eventSetSize = EventSetSizeForSessionList(execution->sessionList);
```
marcocitus: I'd prefer this to be more explicit to show what we're doing and why:
```
/* additional 2 is for postmaster and latch */
int eventSetSize = list_length(execution->sessionList) + 2;
```

src/backend/distributed/executor/executor.c
```diff
@@ -2465,7 +2469,7 @@ WorkerPoolFailed(WorkerPool *workerPool)
 	ListCell *sessionCell = NULL;
 
 	/* a pool cannot fail multiple times */
-	Assert (!workerPool->failed);
+	Assert(!workerPool->failed);
```
marcocitus: should maybe be in separate commits.

onderkalaci: Just removed those, I'll have another PR for indent etc.

PR #2767, opened 2019-06-14
Decrement idleConnectionCount when idle connection is lost

marcocitus:
As discussed in #2764


onderkalaci:
Verified that the assert failure is gone.

PR #2768, opened 2019-06-16
Implement executor locks SELECT FOR UPDATE on reference tables

onderkalaci:
On master branch, we do two extra things on SELECT FOR UPDATE
queries that touches any rows of reference tables:
   - Acquire executor locks
   - Execute the query on all replicas

With this commit, we're only adding `Acquire executor locks` piece to
unified executor, because we think that executing the query on all
replicas doesn't add any benefits over the executor locks.

### DIFF
--- a/src/backend/distributed/executor/executor.c
+++ b/src/backend/distributed/executor/executor.c
@@ -697,21 +697,35 @@ DistributedExecutionRequiresRollback(DistributedExecution *execution)
 {
 	CmdType operation = execution->operation;
 	List *taskList = execution->tasksToExecute;
+	int taskCount = list_length(taskList);
 	Task *task = NULL;
+	bool selectForUpdate = false;
 
 	if (MultiShardCommitProtocol == COMMIT_PROTOCOL_BARE)
 	{
 		return false;
 	}
 
-	if (operation == CMD_SELECT)
-	{
-		return SelectOpensTransactionBlock && IsTransactionBlock();
-	}
-
-	if (taskList == NIL)
+	if (taskCount == 0)
 	{
 		return false;
+	}
+
+	task = (Task *) linitial(taskList);
+
+	selectForUpdate = task->relationRowLockList != NIL;
+	if (selectForUpdate)
+	{
+		/*
+		 * Do not check SelectOpensTransactionBlock, always open transaction block
+		 * if SELECT FOR UPDATE is executed inside a distributed transaction.
+		 */
+		return IsTransactionBlock();
+	}
+
+	if (ReadOnlyTask(task->taskType))
+	{
+		return SelectOpensTransactionBlock && IsTransactionBlock();
 	}
 
 	if (IsMultiStatementTransaction())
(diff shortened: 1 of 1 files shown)


onderkalaci:
I updated the PR and the description according to our discussions.

src/backend/distributed/executor/executor.c
```diff
@@ -706,7 +708,7 @@ DistributedExecutionRequiresRollback(DistributedExecution *execution)
 		return false;
 	}
 
-	if (execution->operation == CMD_SELECT)
+	if (operation == CMD_SELECT)
 	{
 		return SelectOpensTransactionBlock && IsTransactionBlock();
```
marcocitus: Should we ignore this GUC when there is FOR UPDATE?

onderkalaci: I guess in general without a transaction block, `SELECT FOR UPDATE` doesn't make much sense. Am I missing anything here?

I considered/tests changing here as well, but then (a) we'd be slightly converging from how Postgres behaves (b) could slightly confuse the users. 

So, my slight preference is to keep as-is. But if you prefer to change, I can do that.

marcocitus: The GUC is specifically meant for situations where you do `BEGIN; SELECT`. We give the users to not open a transaction block to reduce idle in transaction time, but we should never do that in case of FOR UPDATE.

onderkalaci: Oh, do you mean ignore the GUC in the sense that if we're in a TX block, have `SELECT FOR UPDATE` and the GUC is OFF, we still open the tx block on the workers?

marcocitus: right

onderkalaci: This makes a lot of sense.

src/backend/distributed/executor/executor.c
```diff
@@ -698,22 +699,29 @@ TaskListModifiesDatabase(CmdType operation, List *taskList)
...
 
-	if (execution->operation == CMD_SELECT)
+	if (taskCount == 0)
 	{
-		return SelectOpensTransactionBlock && IsTransactionBlock();
+		return false;
 	}
 
-	if (taskList == NIL)
+	if (operation == CMD_SELECT)
```
marcocitus: wondering if we can avoid explicitly checking for operations (given upcoming CTE changes)

```c
task = (Task *) linitial(taskList);

selectForUpdate = task->relationRowLockList != NIL;
if (selectForUpdate)
{
    return IsTransactionBlock();
}

if (ReadOnlyTask(task))
{
    return SelectOpensTransactionBlock && IsTransactionBlock();
}
```

src/backend/distributed/executor/executor.c
```diff
@@ -806,9 +851,13 @@ AcquireExecutorShardLocks(DistributedExecution *execution)
 	CmdType operation = execution->operation;
 	List *taskList = execution->tasksToExecute;
 
-	if (!(operation == CMD_INSERT || operation == CMD_UPDATE || operation == CMD_DELETE))
+	if (!(operation == CMD_INSERT || operation == CMD_UPDATE || operation == CMD_DELETE ||
+		  SelectForUpdateOnReferenceTable(operation, taskList)))
```
marcocitus: Would be good to turn this into a function.

PR #2769, opened 2019-06-17
Refactor shard creation logic

onderkalaci:
This is a preparation for the new executor, where creating shards
would go through the executor. So, explicitly generate the commands
for further processing.

### DIFF
--- a/src/backend/distributed/master/master_stage_protocol.c
+++ b/src/backend/distributed/master/master_stage_protocol.c
@@ -553,11 +553,7 @@ CreateShardsOnWorkers(Oid distributedRelationId, List *shardPlacements,
 		{
 			List *placementAccessList = NIL;
 
-			/* we only need to calculate this once per shardInterval, not placement */
-			if (relationShardList == NIL)
-			{
-				relationShardList = RelationShardListForShardCreate(shardInterval);
-			}
+			relationShardList = RelationShardListForShardCreate(shardInterval);
 
 			placementAccessList = BuildPlacementDDLList(shardPlacement->groupId,
 														relationShardList);
@@ -617,6 +613,7 @@ RelationShardListForShardCreate(ShardInterval *shardInterval)
 	List *allForeignKeyRelations =
 		list_concat_unique_oid(referencedRelationList, referencingRelationList);
 	int shardIndex = -1;
+	ListCell *fkeyRelationIdCell = NULL;
 
 	/* record the placement access of the shard itself */
 	relationShard = CitusMakeNode(RelationShard);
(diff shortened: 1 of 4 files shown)


src/backend/distributed/master/master_stage_protocol.c
```diff
@@ -627,7 +638,9 @@ WorkerCreateShard(Oid relationId, int shardIndex, uint64 shardId, List *ddlComma
 							 escapedDDLCommand);
 		}
 
-		ExecuteCriticalRemoteCommand(connection, applyDDLCommand->data);
+		appendStringInfoChar(applyDDLCommand, ';');
```
onderkalaci: I'll remove all `appendStringInfoChar(command, ';');` from this PR.

src/backend/distributed/master/master_stage_protocol.c
```diff
@@ -544,18 +551,14 @@ CreateShardsOnWorkers(Oid distributedRelationId, List *shardPlacements,
 		 */
 		if (ShouldRecordRelationAccess() && useExclusiveConnection && partitionTable)
 		{
-			RelationShard *parentRelationShard = CitusMakeNode(RelationShard);
```
onderkalaci: I think before this change we were missing few relationShards, especially it is not a partitioned table: (a) the placement itself (b) fkey relationships. This PR makes sure to cover those as well.

src/backend/distributed/master/master_stage_protocol.c
```diff
@@ -544,18 +551,14 @@ CreateShardsOnWorkers(Oid distributedRelationId, List *shardPlacements,
...
-			List *relationShardList = NIL;
 			List *placementAccessList = NIL;
 
-			parentRelationShard->relationId = PartitionParentOid(distributedRelationId);
-			parentRelationShard->shardId =
-				ColocatedShardIdInRelation(parentRelationShard->relationId, shardIndex);
-			partitionRelationShard->relationId = distributedRelationId;
-			partitionRelationShard->shardId = shardId;
+			/* we only need to calculate this once per shardInterval, not placement */
+			if (relationShardList == NIL)
```
pykello: This is always `true`. I don't know if there is an easy way to do what your comment says here.

src/backend/distributed/master/master_stage_protocol.c
```diff
@@ -595,14 +602,109 @@ CreateShardsOnWorkers(Oid distributedRelationId, List *shardPlacements,
...
+		relationShardList = lappend(relationShardList, parentRelationShard);
+	}
+
+	return relationShardList;
+}
+
+
+/*
+ * WorkerCreateShardCommandList returns a list of DDL commands for the given
+ * shardId  to create the shard on the worker node.
```
pykello:
```suggestion
 * shardId to create the shard on the worker node.
```

src/backend/distributed/master/master_stage_protocol.c
```diff
@@ -595,14 +602,109 @@ CreateShardsOnWorkers(Oid distributedRelationId, List *shardPlacements,
...
+	relationShard->shardId = shardInterval->shardId;
+	relationShardList = list_make1(relationShard);
+
+	if (cacheEntry->partitionMethod == DISTRIBUTE_BY_HASH &&
+		cacheEntry->colocationId != INVALID_COLOCATION_ID)
+	{
+		shardIndex = ShardIndex(shardInterval);
+	}
+
+	/* all foregin constraint relations */
```
pykello:
```suggestion
	/* all foreign key constraint relations */
```

src/backend/distributed/master/master_stage_protocol.c
```diff
@@ -595,14 +602,109 @@ CreateShardsOnWorkers(Oid distributedRelationId, List *shardPlacements,
...
+	}
+
+	/* if  partitioned table, make sure to record the parent table */
+	if (PartitionTable(relationId))
+	{
+		RelationShard *parentRelationShard = CitusMakeNode(RelationShard);
+
+		parentRelationShard->relationId = PartitionParentOid(relationId);
+		parentRelationShard->shardId =
+			ColocatedShardIdInRelation(parentRelationShard->relationId, shardIndex);
```
pykello: Do we support partitioned reference tables? If yes, then this line will probably have some kind of fault.

src/backend/distributed/master/master_stage_protocol.c
```diff
@@ -595,14 +602,109 @@ CreateShardsOnWorkers(Oid distributedRelationId, List *shardPlacements,
...
+	relationShardList = list_make1(relationShard);
+
+	if (cacheEntry->partitionMethod == DISTRIBUTE_BY_HASH &&
+		cacheEntry->colocationId != INVALID_COLOCATION_ID)
+	{
+		shardIndex = ShardIndex(shardInterval);
+	}
+
+	/* all foregin constraint relations */
+	if (allForeignKeyRelations != NIL)
```
pykello: Minor: We could also remove this `if` and unconditionally do the `foreach`. If you agree, then rename `relationIdCell` to `fkeyRelationIdCell`.

src/backend/distributed/master/master_stage_protocol.c
pykello: > Do we support partitioned reference tables? If yes, then this line will probably have some kind of fault (because shardIndex is -1)

I think we don't, so maybe just add some asserts here.

pykello: > > Do we support partitioned reference tables? If yes, then this line will probably have some kind of fault (because shardIndex is -1)
> 
> I think we don't, so maybe just add some asserts here.

Maybe just `Assert(shardIndex != -1)`.

src/backend/distributed/master/master_stage_protocol.c
onderkalaci: I think there is not way of doing that. Removing it.

src/backend/distributed/master/master_stage_protocol.c
onderkalaci: Makes sense added comment as well.

PR #2773, opened 2019-06-18
Feature: optionally force master_update_node during failover

thanodnl:
DESCRIPTION: Feature: optionally force master_update_node during failover

When `master_update_node` is called to update a node's location it waits for appropriate locks to become available. This is useful during normal operation as new operations will be blocked till after the metadata update while running operations have time to finish.

When `master_update_node` is called after a node failure it is less useful to wait for running operations to finish as they can't. The lock being held indicates an operation that once attempted to commit will fail as the machine already failed. Now the downside is the failover is postponed till the termination point of the operation. This has been observed by users to take a significant amount of time causing the rest of the system to be observed unavailable.

With this patch it is possible in such situations to invoke `master_update_node` with 2 optional arguments:
 - `force` (bool defaults to `false`): When called with true the update of the metadata will be forced to proceed by terminating conflicting backends. A cancel is not enough as the backend might be in idle time (eg. an interactive session, or going back and forth between an appliaction), therefore a more intrusive solution of termination is used here.
 - `lock_cooldown` (int defaults to `10000`): This is the time in milliseconds before conflicting backends are terminated. This is to allow the backends to finish cleanly before terminating them. This allows the user to set an upperbound to the expected time to complete the metadata update, eg. performing the failover.

The functionality is implemented by spawning a background worker that has the task of helping a certain backend in acquiring its locks. The backend is either terminated on successful execution of the metadata update, or once the memory context of the expression gets reset, eg. on a cancel of the statement.

### DIFF
--- a/src/backend/distributed/citus--8.2-2--8.2-3.sql
+++ b/src/backend/distributed/citus--8.2-2--8.2-3.sql
@@ -20,4 +20,6 @@ COMMENT ON FUNCTION master_update_node(node_id int,
                                        force bool)
   IS 'change the location of a node';
 
+REVOKE ALL ON FUNCTION master_update_node(int,text,int,bool) FROM PUBLIC;
+
 RESET search_path;
--- a/src/backend/distributed/citus--8.2-2--8.2-3.sql
+++ b/src/backend/distributed/citus--8.2-2--8.2-3.sql
@@ -9,7 +9,8 @@ DROP FUNCTION master_update_node(node_id int,
 CREATE OR REPLACE FUNCTION master_update_node(node_id int,
                                               new_node_name text,
                                               new_node_port int,
-                                              force bool DEFAULT false)
+                                              force bool DEFAULT false,
+                                              lock_cooldown int DEFAULT 10000)
   RETURNS void
   LANGUAGE C STRICT
   AS 'MODULE_PATHNAME', $$master_update_node$$;
@@ -17,9 +18,10 @@ CREATE OR REPLACE FUNCTION master_update_node(node_id int,
 COMMENT ON FUNCTION master_update_node(node_id int,
                                        new_node_name text,
                                        new_node_port int,
-                                       force bool)
-  IS 'change the location of a node';
+                                       force bool,
+                                       lock_cooldown int)
+  IS 'change the location of a node. when force => true it will wait lock_cooldown ms before killing competing locks';
 
-REVOKE ALL ON FUNCTION master_update_node(int,text,int,bool) FROM PUBLIC;
+REVOKE ALL ON FUNCTION master_update_node(int,text,int,bool,int) FROM PUBLIC;
 
 RESET search_path;
--- a/src/backend/distributed/citus--8.2-2--8.2-3.sql
+++ b/src/backend/distributed/citus--8.2-2--8.2-3.sql
@@ -1,0 +1,23 @@
+/* citus--8.2-2--8.2-3 */
+
+SET search_path = 'pg_catalog';
+
+DROP FUNCTION master_update_node(node_id int,
+                                 new_node_name text,
+                                 new_node_port int);
+
+CREATE OR REPLACE FUNCTION master_update_node(node_id int,
+                                              new_node_name text,
+                                              new_node_port int,
+                                              force bool DEFAULT false)
+  RETURNS void
+  LANGUAGE C STRICT
+  AS 'MODULE_PATHNAME', $$master_update_node$$;
+
+COMMENT ON FUNCTION master_update_node(node_id int,
+                                       new_node_name text,
+                                       new_node_port int,
+                                       force bool)
+  IS 'change the location of a node';
+
+RESET search_path;
--- a/src/backend/distributed/utils/acquire_lock.c
+++ b/src/backend/distributed/utils/acquire_lock.c
@@ -222,7 +222,15 @@ LockAcquireHelperMain(Datum main_arg)
 	/* connecting to the database */
 	BackgroundWorkerInitializeConnectionByOid(args->DatabaseId, InvalidOid, 0);
 
-	/* TODO explain sql below */
+	/*
+	 * The query below does a self join on pg_locks to find backends that are granted a
+	 * lock our target backend (backendPid) is waiting for. Once it found such a backend
+	 * it terminates the backend with pg_terminate_pid.
+	 *
+	 * The result is are rows of pid,bool indicating backend that is terminated and the
+	 * success of the termination. These will be logged accordingly below for an
+	 * administrator to correlate in the logs with the termination message.
+	 */
 	initStringInfo(&sql);
 	appendStringInfo(&sql,
 					 "SELECT \n"
@@ -260,8 +268,6 @@ LockAcquireHelperMain(Datum main_arg)
 		{
 			for (row = 0; row < SPI_processed; row++)
 			{
-				/* TODO count the number of backends canceled and log about it */
-
 				int terminatedPid = 0;
 				bool isTerminated = false;
 				bool isnull = false;
--- a/src/backend/distributed/utils/acquire_lock.c
+++ b/src/backend/distributed/utils/acquire_lock.c
@@ -29,6 +29,7 @@
 #include "utils/snapmgr.h"
 
 #include "distributed/citus_acquire_lock.h"
+#include "distributed/version_compat.h"
 
 /* forward declaration of background worker entrypoint */
 extern void LockAcquireHelperMain(Datum main_arg);
@@ -69,7 +70,9 @@ StartLockAcquireHelperBackgroundWorker(int backendToHelp)
 	snprintf(worker.bgw_name, BGW_MAXLEN,
 			 "Citus Lock Acquire Helper: %d/%u",
 			 backendToHelp, MyDatabaseId);
+#if PG_VERSION_NUM >= 110000
 	snprintf(worker.bgw_type, BGW_MAXLEN, "citus_lock_aqcuire");
+#endif
 
 	/* TODO verify we need both */
 	worker.bgw_flags = BGWORKER_SHMEM_ACCESS | BGWORKER_BACKEND_DATABASE_CONNECTION;
--- a/src/backend/distributed/utils/acquire_lock.c
+++ b/src/backend/distributed/utils/acquire_lock.c
@@ -24,6 +24,7 @@
 
 
 #include "access/xact.h"
+#include "catalog/pg_type.h"
 #include "executor/spi.h"
 #include "miscadmin.h"
 #include "pgstat.h"
@@ -196,6 +197,11 @@ LockAcquireHelperMain(Datum main_arg)
 	const TimestampTz deadline = TimestampTzPlusMilliseconds(connectionStart,
 															 args->lock_cooldown);
 
+	/* parameters for sql query to be executed */
+	const int paramCount = 1;
+	Oid paramTypes[1] = { INT4OID };
+	Datum paramValues[1];
+
 	pqsignal(SIGTERM, lock_acquire_helper_sigterm);
 
 	BackgroundWorkerUnblockSignals();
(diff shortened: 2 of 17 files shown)


src/backend/distributed/citus--8.2-2--8.2-3.sql
```diff
@@ -0,0 +1,27 @@
+/* citus--8.2-2--8.2-3 */
+
+SET search_path = 'pg_catalog';
+
+DROP FUNCTION master_update_node(node_id int,
```
thanodnl: We are replacing the function, since the new arguments are optional I dropped this function. The usage should be the same as before, the only downside would be if a user has granted extra roles with the privilege to invoke this function, those changes are lost.

Wonder if it is expected that someone would grant other roles with this permission and we should prevent the grants to be lost on update.

src/backend/distributed/utils/acquire_lock.c
```diff
@@ -0,0 +1,303 @@
...
+
+	got_sigterm = true;
+	SetLatch(MyLatch);
+
+	errno = save_errno;
+}
+
+
+static bool
+ShouldAcquireLock(long sleepms)
```
thanodnl: needs comments

src/backend/distributed/utils/node_metadata.c
```diff
@@ -8,6 +8,8 @@
 #include "miscadmin.h"
 #include "funcapi.h"
 
+#include <unistd.h>
```
thanodnl: unused

src/test/regress/expected/isolation_master_update_node.out
```diff
@@ -0,0 +1,57 @@
...
+ <waiting ...>
+step s2-update-node-1-force: <... completed>
+master_update_node
+
+               
+step s2-abort: ABORT;
+step s1-abort: ABORT;
+WARNING: this step had a leftover error message
+FATAL:  terminating connection due to administrator command
+server closed the connection unexpectedly
```
thanodnl: The tests are racy on what kind of error they return during the termination. Sometimes it is an ssl error, other times its this error.

I created two acceptable output versions due to that, which is suboptimal, but the best I could come up with.

src/test/regress/expected/isolation_citus_dist_activity.out
```diff
@@ -45,16 +45,16 @@ step s3-view-worker:
 
 query          query_hostname query_hostport master_query_host_namemaster_query_host_portstate          wait_event_typewait_event     usename        datname        
 
-SELECT worker_apply_shard_ddl_command (105941, 'public', '
+SELECT worker_apply_shard_ddl_command (105949, 'public', '
```
marcocitus: We might be missing a `SET citus.next_shard_id`

src/backend/distributed/utils/node_metadata.c
```diff
@@ -524,12 +538,31 @@ master_update_node(PG_FUNCTION_ARGS)
 	 */
```
marcocitus: The comment above might need to be rephrased a bit, mainly the drawback.

src/include/distributed/citus_acquire_lock.h
```diff
@@ -0,0 +1,20 @@
...
+#ifndef CITUS_CITUS_ACQUIRE_LOCK_H
+#define CITUS_CITUS_ACQUIRE_LOCK_H
+
+
+#include "postmaster/bgworker.h"
+
+BackgroundWorkerHandle * StartLockAcquireHelperBackgroundWorker(int backendToHelp,
+																int32 lock_cooldown);
+
+#endif /*CITUS_CITUS_ACQUIRE_LOCK_H */
```
marcocitus: spurious search and replace

src/backend/distributed/utils/node_metadata.c
```diff
@@ -524,12 +538,31 @@ master_update_node(PG_FUNCTION_ARGS)
...
 	if (WorkerNodeIsPrimary(workerNode))
 	{
+		/*
+		 * before acquiring the locks check if we want a background worker to help us to
+		 * aggressively obtain the locks.
+		 */
+		if (force)
+		{
+			handle = StartLockAcquireHelperBackgroundWorker(MyProcPid, lock_cooldown);
+			ereport(NOTICE, (errmsg("force the update to the node")));
```
marcocitus: could be a bit more descriptive on what that means, e.g.: `cancelling pending transactions on node %d in %d milliseconds`

src/backend/distributed/utils/acquire_lock.c
```diff
@@ -0,0 +1,311 @@
...
+					 "             AND conflicting.objid = blocked.objid)\n"
+					 " WHERE conflicting.granted = true\n"
+					 "   AND blocked.granted = false\n"
+					 "   AND blocked.pid = %d;",
+					 backendPid);
+
+	while (ShouldAcquireLock(100))
+	{
+		int i = 0;
+		int ret = 0;
```
marcocitus: Maybe spiStatus.

src/backend/distributed/utils/acquire_lock.c
```diff
@@ -0,0 +1,311 @@
...
+				   sleepms * 1L, PG_WAIT_EXTENSION);
+	ResetLatch(MyLatch);
+
+	/* emergency bailout if postmaster has died */
+	if (rc & WL_POSTMASTER_DEATH)
+	{
+		proc_exit(1);
+	}
+
+	CHECK_FOR_INTERRUPTS();
```
marcocitus: I wonder if we should just do a graceful proc_exit here instead of throwing an error.

thanodnl: I am not completely sure what happens in `CHECK_FOR_INTERRUPTS` that would cause the ungraceful `proc_exit`. This logic is extracted from an example in the postgres source code of [`src/test/modules/worker_spi`](https://github.com/postgres/postgres/blob/414cca40d506dd3f17b49ae3139853139192c2ba/src/test/modules/worker_spi/worker_spi.c#L228-L243).

This code is based on that as I assumed that is best practice on the signal handling in a worker (albeit I skipped sighup for now as we don't use any guc's).
Let me know if you have reasons to believe the `CHECK_FOR_INTERRUPTS` here would cause an issue.

marcocitus: CHECK_FOR_INTERRUPT() throws an error here: https://github.com/postgres/postgres/blob/master/src/backend/tcop/postgres.c#L3137 .  This is normally used to interrupt queries and abort the transaction.

If an error is thrown in a background worker and not caught anywhere then that is seen as a failure and the worker exits with error code 1 and is restarted by postmaster, though that can be prevented by `BGW_NEVER_RESTART`. 

However, in this case getting a cancellation is a perfectly normal, expected outcome, so we might as well exit gracefully.

src/backend/distributed/utils/node_metadata.c
thanodnl: actually, I think I will remove this.

I used this early in the development to make sure I was hitting the codepath in a quick and dirty way :). Given that the user invokes the function via `force => true` to get here I see little value in keeping this `NOTICE`

src/include/distributed/citus_acquire_lock.h
thanodnl: I assume you meant the double `CITUS` in the identifier, removed one.

src/test/regress/expected/isolation_citus_dist_activity.out
thanodnl: sounds plausible, I will dive a bit deeper and use an unused range for these tests to make them more deterministic in their execution.

src/backend/distributed/utils/acquire_lock.c
thanodnl: During normal operation we call `TerminateBackgroundWorker` which asks postmaster to send the background worker a `SIGTERM`. This signal is trapped in via `lock_acquire_helper_sigterm` per [`pqsignal`](https://github.com/citusdata/citus/pull/2773/files#diff-2e185cd6be0ac05904cbedc2db8803d1R193).

any other signal is unexpected behaviour I guess and suffices a non clean exit.

I am not sure about the semantics of `pqsignal` with `CHECK_FOR_INTERRUPTS` though, are you saying the latter will handle `SIGTERM` differently in the case where `MyLatch` has stopped due to a timeout and a `SIGTERM` has been been received between line 162 and the `CHECK_FOR_INTERRUPTS`?

During development I have not witnessed any non graceful shutdown of the background worker. For safety reasons the background worker is actually started with `BGW_NEVER_RESTART` as a defensive way of running this background worker.

src/backend/distributed/utils/acquire_lock.c
```diff
@@ -0,0 +1,311 @@
...
+	/*
+	 * this loop waits till the deadline is reached (eg. lock_cooldown has passed) OR we
+	 * no longer need to acquire the lock due to the termination of this backend.
+	 * Only after the timeout the code will continue with the section that will acquire
+	 * the lock.
+	 */
+	do {
+		timeout = DeadlineTimestampTzToTimeout(deadline);
+	} while (timeout > 0 && ShouldAcquireLock(timeout));
+
```
marcocitus: Maybe add an early exit in case of got_sigterm.

src/backend/distributed/utils/acquire_lock.c
```diff
@@ -0,0 +1,311 @@
...
+					 "SELECT \n"
+					 "    DISTINCT conflicting.pid,\n"
+					 "    pg_terminate_backend(conflicting.pid)\n"
+					 "  FROM pg_locks AS blocked\n"
+					 "       JOIN pg_locks AS conflicting\n"
+					 "         ON (conflicting.database = blocked.database\n"
+					 "             AND conflicting.objid = blocked.objid)\n"
+					 " WHERE conflicting.granted = true\n"
+					 "   AND blocked.granted = false\n"
+					 "   AND blocked.pid = %d;",
```
marcocitus: Probably more hassle than it's worth, but this seems like a candidate for `SPI_execute_with_args`.

thanodnl: `with_args` makes a lot of sense, will rewrite it in such a way. Even though I think it will be safe to interpolate an int in there, it always feels wrong to munge sql with string interpolation.

src/backend/distributed/utils/acquire_lock.c
thanodnl: `ShouldAcquireLock(...)` escapes on `got_sigterm` albeit not early. We could change it in such a way that before entering the wait we can already check for `got_sigterm`. Or is your point to have an explicit `&& !got_sigterm` in the condition?

src/backend/distributed/citus--8.2-2--8.2-3.sql
marcocitus: Good question. We routinely drop functions like master_add_node, so I think we can continue doing it here. 

We once took the position that these functions should not explicitly require superuser, because it would be more secure to call them only via an unprivileged role, though is difficult to maintain in practice. I think we should perhaps maintain such a role ourselves. That role can then be granted to a user.

src/backend/distributed/utils/acquire_lock.c
marcocitus: Either way seems fine, currently it waits 100ms for no particular reason.

src/backend/distributed/utils/node_metadata.c
thanodnl: I have added some extra information here, could you have a look if that is what you were looking for?

src/backend/distributed/utils/acquire_lock.c
thanodnl: added an early escape in `ShouldAcquireLock` as it is also used in the loop below.

src/include/distributed/citus_acquire_lock.h
```diff
@@ -0,0 +1,20 @@
+/*-------------------------------------------------------------------------
+ *
+ * citus_acquire_lock.h
+ *	  Background worker to help with acquiering locks by canceling competing backends.
+ *
+ * Copyright (c) 2017, Citus Data, Inc.
```
thanodnl: cc year

src/test/regress/expected/isolation_citus_dist_activity.out
thanodnl: Setting the shard id causes all bit the first permutation to fail due to the following messages in the output:
```
ERROR:  cached metadata for shard 107000 is inconsistent
```

I left a comment in the test file for the next soul that will run into this.

src/backend/distributed/citus--8.2-2--8.2-3.sql
thanodnl: Like the idea of the citus managed role, will not implement it now though

PR #2774, opened 2019-06-19
Fix subxact release crash.

pykello:
Before this patch, distributed `RELEASE SAVEPOINT` crashed if worker `RELEASE SAVEPOINT` failed and we were in a critical transaction. Before this patch, state of our subxact stack and postgres transaction changed as follows:

```
SAVEPOINT s1;
postgres->push(s1);
citus->push(s1);
RELEASE s1;
citus->pop(s1);
-- Worker RELEASE fails, we throw an ERROR, so postgres doesn't pop
ROLLBACK s1; -- inserted implicitly, because RELEASE failed
citus->pop(s1); -- Crash, since we have already popped s1 before
```

With this fix,
```
SAVEPOINT s1;
postgres->push(s1);
citus->push(s1);
RELEASE s1;
-- Worker RELEASE fails, we throw an ERROR, so postgres doesn't pop,
-- and citus doesn't pop.
ROLLBACK s1; -- inserted implicitly, because RELEASE failed
citus->pop(s1); 
postgres->pop(s1);
```

### DIFF
--- a/src/backend/distributed/transaction/transaction_management.c
+++ b/src/backend/distributed/transaction/transaction_management.c
@@ -396,6 +396,16 @@ CoordinatedSubTransactionCallback(SubXactEvent event, SubTransactionId subId,
 {
 	switch (event)
 	{
+		/*
+		 * Our subtransaction stack should be consistent with postgres' internal
+		 * transaction stack. In case of subxact begin, postgres calls our
+		 * callback after it has pushed the transaction into stack, so we have to
+		 * do the same even if worker commands fail, so we PushSubXact() first.
+		 * In case of subxact commit, callback is called before pushing subxact to
+		 * the postgres transaction stack, so we call PopSubXact() after making sure
+		 * worker commands didn't fail. Otherwise, Postgres would roll back that
+		 * would cause us to call PopSubXact again.
+		 */
 		case SUBXACT_EVENT_START_SUB:
 		{
 			PushSubXact(subId);
@@ -408,21 +418,21 @@ CoordinatedSubTransactionCallback(SubXactEvent event, SubTransactionId subId,
 
 		case SUBXACT_EVENT_COMMIT_SUB:
 		{
-			PopSubXact(subId);
 			if (InCoordinatedTransaction())
 			{
 				CoordinatedRemoteTransactionsSavepointRelease(subId);
 			}
+			PopSubXact(subId);
 			break;
 		}
 
 		case SUBXACT_EVENT_ABORT_SUB:
 		{
-			PopSubXact(subId);
 			if (InCoordinatedTransaction())
 			{
 				CoordinatedRemoteTransactionsSavepointRollback(subId);
 			}
+			PopSubXact(subId);
 
 			UnsetCitusNoticeLevel();
 			break;
--- a/src/test/regress/expected/failure_savepoints.out
+++ b/src/test/regress/expected/failure_savepoints.out
@@ -38,8 +38,6 @@ CONTEXT:  while executing command on localhost:9060
 WARNING:  connection not open
 CONTEXT:  while executing command on localhost:9060
 DELETE FROM artists WHERE id=4;
-WARNING:  could not issue cancel request
-DETAIL:  Client error: PQcancel() -- no cancel object supplied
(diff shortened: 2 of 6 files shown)


src/test/regress/expected/failure_savepoints.out
```diff
@@ -66,16 +66,26 @@ SELECT citus.mitmproxy('conn.onQuery(query="^RELEASE").kill()');
 (1 row)
 
 BEGIN;
-INSERT INTO artists VALUES (5, 'Asher Lev');
+UPDATE artists SET name='a';
 SAVEPOINT s1;
 DELETE FROM artists WHERE id=4;
 RELEASE SAVEPOINT s1;
+WARNING:  AbortSubTransaction while in COMMIT state
```
pykello: This means `RELEASE SAVEPOINT` started, but didn't finish, so this warning is expected.

src/test/regress/sql/failure_savepoints.sql
```diff
@@ -33,11 +33,11 @@ SELECT * FROM artists WHERE id IN (4, 5);
 SELECT citus.mitmproxy('conn.onQuery(query="^RELEASE").kill()');
 
 BEGIN;
-INSERT INTO artists VALUES (5, 'Asher Lev');
+UPDATE artists SET name='a';
```
pykello: This is to make the transaction critical.

src/test/regress/expected/failure_savepoints.out
```diff
@@ -66,16 +66,26 @@ SELECT citus.mitmproxy('conn.onQuery(query="^RELEASE").kill()');
...
+WARNING:  could not issue cancel request
+DETAIL:  Client error: PQcancel() -- no cancel object supplied
```
marcocitus: Not caused by this PR, but this is the first time I'm seeing this warning. Maybe we can suppress it in `SendCancelationRequest`?:
```c
    PGcancel *cancelObject = PQgetCancel(connection->pgConn);
    if (cancelObject == NULL)
    {
        /* connection failed, nothing to cancel */
        return false;
    }
```

src/backend/distributed/transaction/transaction_management.c
```diff
@@ -408,11 +419,11 @@ CoordinatedSubTransactionCallback(SubXactEvent event, SubTransactionId subId,
 
 		case SUBXACT_EVENT_COMMIT_SUB:
 		{
-			PopSubXact(subId);
 			if (InCoordinatedTransaction())
 			{
 				CoordinatedRemoteTransactionsSavepointRelease(subId);
 			}
+			PopSubXact(subId);
```
marcocitus: What about the code below?  (case SUBXACT_EVENT_ABORT_SUB)

src/test/regress/expected/failure_savepoints.out
pykello: done

src/backend/distributed/transaction/transaction_management.c
pykello: `ROLLBACK TO SAVEPOINT` never errors out, so in  `case SUBXACT_EVENT_ABORT_SUB` order doesn't matter, but changed it after our discussion to keep the order of pops consistent.

src/backend/distributed/transaction/transaction_management.c
```diff
@@ -396,6 +396,17 @@ CoordinatedSubTransactionCallback(SubXactEvent event, SubTransactionId subId,
...
 	switch (event)
 	{
+		/*
+		 * Our subtransaction stack should be consistent with postgres' internal
+		 * transaction stack. In case of subxact begin, postgres calls our
+		 * callback after it has pushed the transaction into stack, so we have to
+		 * do the same even if worker commands fail, so we PushSubXact() first.
+		 * In case of subxact commit, callback is called before pushing subxact to
+		 * the postgres transaction stack, so we call PopSubXact() after making sure
+		 * worker commands didn't fail. For subxact abort, we don't care about the
```
marcocitus: Maybe add: "Otherwise, Postgres would roll back that would cause us to call PopSubXact again".

The last sentence can probably be removed.

PR #2778, opened 2019-06-20
Support CTEs in router planner for modification queries

### LINKED ISSUE #2400: CTEs in modifying router queries do recursive planning
**Schema:**
```
CREATE TABLE test_id(id int,tenant_id int);
SELECT create_distributed_table('test_id','tenant_id');
```
**Query that does distributed planning:**
```
EXPLAIN WITH t AS (SELECT id from test_id where tenant_id=1)
UPDATE test_id ti SET id=1 FROM t where ti.id=t.id and ti.tenant_id=1;
                                                          QUERY PLAN
------------------------------------------------------------------------------------------------------------------------------
 Custom Scan (Citus Router)  (cost=0.00..0.00 rows=0 width=0)
   ->  Distributed Subplan 3_1
         ->  Custom Scan (Citus Router)  (cost=0.00..0.00 rows=0 width=0)
               Task Count: 1
               Tasks Shown: All
               ->  Task
                     Node: host=ec2-54-175-83-159.compute-1.amazonaws.com port=5432 dbname=citus
                     ->  Seq Scan on test_id_102201 test_id  (cost=0.00..38.25 rows=11 width=4)
                           Filter: (tenant_id = 1)
   Task Count: 1
   Tasks Shown: All
   ->  Task
         Node: host=ec2-54-175-83-159.compute-1.amazonaws.com port=5432 dbname=citus
         ->  Update on test_id_102201 ti  (cost=38.39..52.69 rows=55 width=42)
               ->  Hash Join  (cost=38.39..52.69 rows=55 width=42)
                     Hash Cond: (intermediate_result.id = ti.id)
                     ->  Function Scan on read_intermediate_result intermediate_result  (cost=0.00..10.00 rows=1000 width=32)
                     ->  Hash  (cost=38.25..38.25 rows=11 width=14)
                           ->  Seq Scan on test_id_102201 ti  (cost=0.00..38.25 rows=11 width=14)
                                 Filter: (tenant_id = 1)
```

We don't need to do recursive planning for the query, it can be directly pushed down. The concern is recursive planning generates connection to each worker because of reference table logic.

serprex:
Fixes #2400 

DESCRIPTION: Support CTEs in router planner for modification queries


src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -3002,13 +3039,14 @@ MultiRouterPlannableQuery(Query *query)
 
 				if (tableReplicationFactor > 1 && partitionMethod != DISTRIBUTE_BY_NONE)
 				{
-					return false;
+					distributedPlan->planningError = DeferredError(
+						ERRCODE_FEATURE_NOT_SUPPORTED,
+						"Table replication factor > 1 not supported for non-reference tables.",
```
marcocitus: Maybe "SELECT FOR UPDATE with table replication factor > 1 ..."

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -2941,14 +2966,15 @@ ExtractInsertPartitionKeyValue(Query *query)
...
+ * setting distributedPlan->planningError if not.
  * The query is router plannable if it is a modify query, or if its is a select
  * query issued on a hash partitioned distributed table. Router plannable checks
  * for select queries can be turned off by setting citus.enable_router_execution
  * flag to false.
  */
-static bool
-MultiRouterPlannableQuery(Query *query)
+static void
+MultiRouterPlannableQuery(DistributedPlan *distributedPlan, Query *query)
```
marcocitus: I think it would be a little cleaner to return a DeferredError directly.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -163,24 +165,27 @@ static void ReorderTaskPlacementsByTaskAssignmentPolicy(Job *job,
...
-	if (MultiRouterPlannableQuery(query))
+	DistributedPlan *distributedPlan = CitusMakeNode(DistributedPlan);
+
+	MultiRouterPlannableQuery(distributedPlan, query);
+
+	if (distributedPlan->planningError == NULL)
 	{
-		return CreateSingleTaskRouterPlan(originalQuery, query,
-										  plannerRestrictionContext);
+		distributedPlan->planningError = ErrorIfQueryHasModifyingCTE(query);
```
marcocitus: I know the TODO suggested moving it here, but what about putting it in `MultiRouterPlannableQuery` now that it neatly returns an error?

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -163,24 +165,27 @@ static void ReorderTaskPlacementsByTaskAssignmentPolicy(Job *job,
...
+			CreateSingleTaskRouterPlan(distributedPlan, originalQuery, query,
+									   plannerRestrictionContext);
+		}
 	}
 
-	/*
-	 * TODO: Instead have MultiRouterPlannableQuery set an error describing
-	 * why router cannot support the query.
-	 */
-	return NULL;
```
marcocitus: Thanks for cleaning this up!

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -603,18 +599,47 @@ ModifyQuerySupported(Query *queryTree, Query *originalQuery, bool multiShardQuer
...
+			{
+				return cteError;
+			}
+		}
 	}
 
 	/* extract range table entries */
 	ExtractRangeTableEntryWalker((Node *) queryTree, &rangeTableList);
 
+	/* TODO is this checking CTEs? requiresMasterEvaluation check above shouldn't be necessary if so */
```
marcocitus: Per [range_table_walker])(https://github.com/postgres/postgres/blob/master/src/backend/nodes/nodeFuncs.c#L2366), it does not descend into CTEs, while it does descend into subqueries.

I'm wondering whether we should fix this in `ExtractRangeTableEntryWalker` (mostly, how dangerous that would be).

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -603,18 +599,47 @@ ModifyQuerySupported(Query *queryTree, Query *originalQuery, bool multiShardQuer
...
+			requiresMasterEvaluation = RequiresMasterEvaluation(cteQuery);
+			if (requiresMasterEvaluation)
+			{
+				return DeferredError(ERRCODE_FEATURE_NOT_SUPPORTED,
+									 "Router planner doesn't support common table expressions requiring master evaluation.",
+									 NULL, NULL);
+			}
+
+			/* TODO can we reduce redundancy here by directly calling PlanRouterQuery? */
+			cteError = CreateRouterPlan(cteQuery, cteQuery,
```
marcocitus: It's a little odd that we're doing all of `CreateRouterPlan` here rather than just the error checks. Maybe we could just do `MultiRouterPlannableQuery` after we move `ErrorIfQueryHasModifyingCTE`?

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -603,18 +599,47 @@ ModifyQuerySupported(Query *queryTree, Query *originalQuery, bool multiShardQuer
...
+			bool requiresMasterEvaluation;
+
+			if (cteQuery->commandType != CMD_SELECT)
+			{
+				return DeferredError(ERRCODE_FEATURE_NOT_SUPPORTED,
+									 "Router planner doesn't support non-select common table expressions.",
+									 NULL, NULL);
+			}
+
+			requiresMasterEvaluation = RequiresMasterEvaluation(cteQuery);
```
marcocitus: Hmz, I was looking at how we handled function calls in updates with subqueries and it turns out there is a bug there:

```
# UPDATE test SET y = 3 WHERE x IN (SELECT x FROM test WHERE x = 3 AND y > random());
LOG:  issuing UPDATE public.test_102023 test SET y = 3 WHERE (x OPERATOR(pg_catalog.=) ANY (SELECT test_1.x FROM public.test_102023 test_1 WHERE ((test_1.x OPERATOR(pg_catalog.=) 3) AND ((test_1.y)::double precision OPERATOR(pg_catalog.>) '0.3421865622513'::double precision))))
```

i.e. we're evaluating the random() on the coordinator, while we should be re-evaluating it for every row on the worker (and actually, this should be recursively planned).

It seems we're primarily doing function evaluation in UPDATE/DELETE to support stable functions like now(), which are expected to return the same result throughout the lifetime of the query. The function evaluation logic does not distinguish between stable and volatile functions. The assumption is that UPDATE and DELETE queries do not contain volatile functions because those are rejected, but it seems we're neglecting to do so in subqueries.

Basically, we should do function evaluation if required, but we need to reject volatile functions in both CTEs and subqueries (and leave them to recursive planning).

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -2980,14 +3009,22 @@ MultiRouterPlannableQuery(Query *query)
...
 			}
 
 			partitionMethod = PartitionMethod(distributedTableId);
 			if (!(partitionMethod == DISTRIBUTE_BY_HASH || partitionMethod ==
 				  DISTRIBUTE_BY_NONE || partitionMethod == DISTRIBUTE_BY_RANGE))
 			{
-				return false;
+				distributedPlan->planningError = DeferredError(
+					ERRCODE_FEATURE_NOT_SUPPORTED,
+					"Unsupported partition method.",
```
marcocitus: Maybe: "Router planner does not support append-partitioned tables" or "Router planner only supports hash- and range-distributed tables and reference tables"

src/backend/distributed/planner/multi_router_planner.c
serprex: `ExtractRangeTableEntryWalker` already descends into CTEs

[https://github.com/citusdata/citus/blob/6741ffd716593f09b22a01dab70f64d7c62922cd/src/backend/distributed/planner/multi_logical_planner.c#L2037] this logic goes to [`query_tree_walker`](https://github.com/postgres/postgres/blob/20e1cc898da6f82b5680973c107e8a96d46caf01/src/backend/nodes/nodeFuncs.c#L2299) without `QTW_IGNORE_CTE_SUBQUERIES`

src/backend/distributed/planner/multi_router_planner.c
serprex: It was running into "Complex subqueries and CTEs" type of errors without checking if PlanRouterQuery wasn't going to bail either. Might be related to range table scan seeming to be skipping CTEs

My other understanding is that I need information from PlanRouterQuery in order to make sure the CTE belongs to the same shard (or, more broadly, worker) as the query. Need to test `with cte1 as (query-using-shard-1), cte2 as (query-using-shard-2) select * from (query-using-ctes)`

marcocitus: `with cte1 as (query-using-shard-1), cte2 as (query-using-shard-2) select * from (query-using-ctes)` is only router plannable if shard 1 and 2 are co-located shards from different tables. Otherwise we should be using recursive planning like we do now.

My expectation is that `PlanRouterQuery` only needs to be called once for the whole query, since it checks whether all of the shards that are used in the query belong to the same worker.

src/backend/distributed/planner/multi_router_planner.c
serprex: This seems like an issue to be looked into as a separate PR

marcocitus: Agree that we can defer fixing it for subqueries, though we should figure out how to get it right in CTEs.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -603,13 +594,30 @@ ModifyQuerySupported(Query *queryTree, Query *originalQuery, bool multiShardQuer
...
+				return DeferredError(ERRCODE_FEATURE_NOT_SUPPORTED,
+									 "Router planner doesn't support non-select common table expressions.",
+									 NULL, NULL);
+			}
+
+			cteError = MultiRouterPlannableQuery(cteQuery);
+			if (cteError)
+			{
+				return cteError;
+			}
```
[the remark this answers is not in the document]
serprex: [Added bail for SELECT FOR UPDATE.](https://github.com/citusdata/citus/pull/2778/commits/6619e3d0f772c902d5ff75ca0d24c1438a1a8882) Not sure what "specialised logic" you were referring to, as the isolation test I added passes even when SELECT FOR UPDATE isn't bailed on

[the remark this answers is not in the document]
serprex: So woke up & think I can recognize where this would've been lacking: need to make sure that update tasks get put on the same connection as the SELECT FOR UPDATE task

marcocitus:
Any update on the volatile function issue?

serprex:
@marcocitus added bail out on volatile functions in CTEs. Should've figured this out yesterday, but got lost trying to figure out how to script gdb printing backtraces when errors occur in test suite

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -603,13 +594,46 @@ ModifyQuerySupported(Query *queryTree, Query *originalQuery, bool multiShardQuer
...
+									 "Router planner doesn't support SELECT FOR UPDATE"
+									 " in common table expressions.",
+									 NULL, NULL);
+			}
+
+			if (FindNodeCheck((Node *) cteQuery, CitusIsVolatileFunction))
+			{
+				return DeferredError(ERRCODE_FEATURE_NOT_SUPPORTED,
+									 "Router planner doesn't support VOLATILE functions"
+									 " in common table expressions.",
```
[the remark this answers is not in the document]
serprex: Good idea, will do. My initial thoughts were bad ideas about selecting on the result of `EXPLAIN` to figure out of random was replaced or not

PR #2779, opened 2019-06-20
Add slow start interval to the executor

marcocitus:
This PR adds an `citus.executor_slow_start_interval` setting that defaults to 10ms and determines the number of milliseconds to wait between connection attempts to the same node. At every connection attempt, the number of connections that `ManageWorkerPool` is allowed to open doubles to rapidly build up the number of connections when tasks are taking a long time. This is inspired by TCP slow start, which doubles the congestion window every RTT.


src/backend/distributed/executor/executor.c
```diff
@@ -1565,6 +1575,26 @@ ManageWorkerPool(WorkerPool *workerPool)
...
+			TimestampTz now = GetCurrentTimestamp();
+
+			if (TimestampDifferenceExceeds(workerPool->lastConnectionOpenTime, now,
+										   ExecutorSlowStartInterval))
+			{
+				newConnectionCount = Min(newConnectionCount,
+										 workerPool->maxNewConnectionsPerCycle);
+
+				/* exponentially increase the open rate (like TCP slow start) */
+				workerPool->maxNewConnectionsPerCycle *= 2;
```
marcocitus: wondering if we should start doubling from the second attempt.

e.g. if we do 1+1+2+4+8+16+32, can start utilising 32 cores after 64 cores after 60ms. If we do 1+2+4+8+16+32 we get this slightly odd result of having 63 connections after 50ms and 64 after 60ms.

marcocitus:
The combination of these commits works reasonably well.

src/backend/distributed/executor/executor.c
```diff
@@ -1576,6 +1592,26 @@ ManageWorkerPool(WorkerPool *workerPool)
...
+			TimestampTz now = GetCurrentTimestamp();
+
+			if (TimestampDifferenceExceeds(workerPool->lastConnectionOpenTime, now,
+										   ExecutorSlowStartInterval))
+			{
+				newConnectionCount = Min(newConnectionCount,
+										 workerPool->maxNewConnectionsPerCycle);
+
+				/* exponentially increase the open rate (like TCP slow start) */
+				workerPool->maxNewConnectionsPerCycle += 1;
```
pykello: I'm not sure if I am understanding this correctly. How is this exponential?

marcocitus: It isn't, but it was initially. the comment is stale. Will fix it.

onderkalaci:
could you please rebase the PR? Thanks!

src/include/distributed/connection_management.h
```diff
@@ -129,6 +129,7 @@ typedef struct ConnectionHashEntry
 {
 	ConnectionHashKey key;
 	dlist_head *connections;
+	int connectionCount;
```
onderkalaci: I guess we added this variable because there is no easy way of getting the length of `dlist`? Well, considering the gain we'd have as the connectionCount increases, I'd probably prefer to iterate over the list to count over adding a new field here :)

PR #2780, opened 2019-06-20
Make COPY adapt to connection use behaviour of previous commands in transaction

pykello:
DESCRIPTION: Make COPY adapt to connection use behaviour of previous commands in transaction

This is reviewed at https://github.com/citusdata/citus/pull/2758

Goals are:
1. Be compatible with unified-executor,
2. When using a pool-size >= shard count, behave same as old copy with the same performance.

Notes:
Keeping changes to `CopyToNewShards()` for later, to keep the PR not too large.


src/backend/distributed/connection/placement_connection.c
```diff
@@ -432,6 +432,31 @@ StartPlacementListConnection(uint32 flags, List *placementAccessList,
 }
 
 
+MultiConnection *
+GetPlacementListConnectionIfCached(int flags, List *placementAccessList,
```
onderkalaci: Missing function comment

src/test/regress/expected/foreign_key_restriction_enforcement.out
```diff
@@ -290,10 +290,8 @@ BEGIN;
 DEBUG:  switching to sequential query execution mode
 DETAIL:  Reference relation "referece_table" is modified, which might lead to data inconsistencies or distributed deadlocks via parallel accesses to hash distributed relations due to foreign keys. Any parallel modification to those hash distributed relations in the same transaction can only be executed in sequential query execution mode
 	COPY on_update_fkey_table FROM STDIN WITH CSV;
-ERROR:  cannot execute parallel COPY on relation "on_update_fkey_table" after DML command on reference relation "referece_table" because there is a foreign key between them and "referece_table" has been modified in this transaction
-DETAIL:  COPY to a distributed table uses a separate set of connections which will not be able to see the uncommitted changes to the reference table.
-HINT:  Perform the COPY in a separate transaction.
-CONTEXT:  COPY on_update_fkey_table, line 2: "1002,99"
+ERROR:  insert or update on table "on_update_fkey_table_2380001" violates foreign key constraint "fkey_2383883"
```
onderkalaci: maybe adjust the test so that the data goes into the table via COPY? Because this currently looks slightly confusing for the readers. If you check the regression test file, we typically fail only if we cannot support the tx block.

src/test/regress/expected/multi_insert_select_conflict.out
```diff
@@ -325,9 +321,9 @@ BEGIN;
 		col_2, 
 		col_1 
 	FROM source_table_1 ON CONFLICT (col_1) DO UPDATE SET col_2 = 55 RETURNING *;
-ERROR:  cannot execute parallel COPY on relation "target_table" after DML command on reference relation "test_ref_table" because there is a foreign key between them and "test_ref_table" has been modified in this transaction
-DETAIL:  COPY to a distributed table uses a separate set of connections which will not be able to see the uncommitted changes to the reference table.
-HINT:  Perform the COPY in a separate transaction.
+ERROR:  insert or update on table "target_table_1900001" violates foreign key constraint "fkey_1900001"
```
onderkalaci: Similar here, can we make the query work?

src/backend/distributed/connection/placement_connection.c
onderkalaci: Maybe use `GetConnectionIfPlacementAccessedInXact()`
`GetConnectionIfAnyPlacementsAccessedInXact()` as we discussed. @marcocitus any comment?

marcocitus: Sounds good, slight preference for  GetConnectionIfPlacementAccessedInXact`

PR #2782, opened 2019-06-20
 Fix the newConnectionCount calculation in ManageWorkerPool

marcocitus:
The current calculation only works correctly the first time.

### DIFF
--- a/src/backend/distributed/executor/executor.c
+++ b/src/backend/distributed/executor/executor.c
@@ -1547,27 +1547,30 @@ ManageWorkerPool(WorkerPool *workerPool)
 	}
 	else
 	{
-		if (initiatedConnectionCount >= targetPoolSize)
-		{
-			/* already reached the minimal pool size */
-			return;
-		}
+		/* cannot open more than targetPoolSize connections */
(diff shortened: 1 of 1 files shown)


marcocitus:
@onderkalaci  pushed a fix to the issues.

src/backend/distributed/executor/executor.c
```diff
@@ -1547,27 +1547,30 @@ ManageWorkerPool(WorkerPool *workerPool)
 	}
 	else
 	{
-		if (initiatedConnectionCount >= targetPoolSize)
-		{
-			/* already reached the minimal pool size */
-			return;
-		}
+		/* cannot open more than targetPoolSize connections */
+		int maxNewConnectionCount = targetPoolSize - initiatedConnectionCount;
```
onderkalaci: Maybe `maxNewConnectionCount ` to `maxAllowedConnectionCount `

marcocitus: I actually changed it from `maxAllowedConnectionCount` to `maxNewConnectionCount`  :)

I figured the maximum allowed connection count is the pool size, while here we're calculated the maximum number of new connections that we're allowed to open according to the pool size.

PR #2783, opened 2019-06-20
Fix wait flags in CheckConnectionReady

marcocitus:
`CheckConnectionReady` currently sets `WL_SOCKET_WRITEABLE` in case of `PQisBusy`, but sockets are always writeable unless they are blocked on writes, which means `WaitEventSetWait` wakes up much more often than necessary. This PR fixes the issue by only setting the wait flags to `WL_SOCKET_READABLE | WL_SOCKET_WRITEABLE` when `PQflush` returns `1` and setting the waitFlags to `WL_SOCKET_READABLE` otherwise. 

We also use `WL_SOCKET_WRITEABLE` as signaling mechanism for waking up `WaitEventSetWait` and this change exposes a few cases where we do not set the waitFlags appropriately. Fix those as well.

### DIFF
--- a/src/backend/distributed/executor/executor.c
+++ b/src/backend/distributed/executor/executor.c
@@ -1729,6 +1729,9 @@ ConnectionStateMachine(WorkerSession *session)
 					workerPool->activeConnectionCount++;
 					workerPool->idleConnectionCount++;
 
+					UpdateConnectionWaitFlags(session,
+											  WL_SOCKET_READABLE | WL_SOCKET_WRITEABLE);
+
 					connection->connectionState = MULTI_CONNECTION_CONNECTED;
 					break;
 				}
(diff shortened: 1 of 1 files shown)


src/backend/distributed/executor/executor.c
```diff
@@ -2177,17 +2182,14 @@ CheckConnectionReady(WorkerSession *session)
 		return false;
 	}
 
-	if (PQisBusy(connection->pgConn))
```
onderkalaci: I might be missing something but I cannot exactly match the PR description to the code changes here, so I wanted to double check my understanding:

```
CheckConnectionReady currently sets WL_SOCKET_WRITEABLE in case of PQisBusy, but sockets are always writeable unless they are blocked on writes, which means WaitEventSetWait wakes up much more often than necessary.
```

When I read the core, I see that we set the socket readable when `PQisBusy`, not to writable as the description says.

Is it that  being in `CheckConnectionReady` means that the `session->connection->waitFlags` already contains `WL_SOCKET_WRITEABLE ` so we kept setting it again and again on every call. 

Now, we only set `WL_SOCKET_WRITEABLE ` `if (sendStatus == 1)`, which is what the postgres docs suggest?

marcocitus: I was describing the situation before this PR. Updated the PR description with
> This PR fixes the issue by only setting the wait flags to `WL_SOCKET_READABLE | WL_SOCKET_WRITEABLE` when `PQflush` returns `1` and setting the waitFlags to `WL_SOCKET_READABLE` otherwise.

PR #2784, opened 2019-06-21
failure_savepoints output for unified executor.

pykello:
Before unified executor we didn't error out in "SAVEPOINT" and "RELEASE" commands if the transaction wasn't critical, and instead we marked the placements as invalid. Now we error out in those commands, so output changes quite a bit.

### DIFF
--- a/src/test/regress/expected/failure_savepoints.out
+++ b/src/test/regress/expected/failure_savepoints.out
@@ -1,3 +1,8 @@
+-- We have two different output files for this failure test because the
+-- failure behaviour of SAVEPOINT and RELEASE commands are different if
+-- we use the executor. If we use it, these commands error out if any of
+-- the placement commands fail. Otherwise, we might mark the placement
+-- as invalid and continue with a WARNING.
 SELECT citus.mitmproxy('conn.allow()');
  mitmproxy 
 -----------
--- a/src/test/regress/expected/failure_savepoints_9.out
+++ b/src/test/regress/expected/failure_savepoints_9.out
@@ -3,64 +3,125 @@
 -- we use the executor. If we use it, these commands error out if any of
 -- the placement commands fail. Otherwise, we might mark the placement
 -- as invalid and continue with a WARNING.
-
 SELECT citus.mitmproxy('conn.allow()');
+ mitmproxy 
+-----------
+ 
+(1 row)
 
 SET citus.shard_count = 2;
 SET citus.shard_replication_factor = 1; -- one shard per worker
 SET citus.next_shard_id TO 100950;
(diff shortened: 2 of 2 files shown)


src/test/regress/expected/failure_savepoints_0.out
```diff
@@ -0,0 +1,348 @@
+SELECT citus.mitmproxy('conn.allow()');
```
onderkalaci: I typically add comment to the sql file explaining why we have two output files. Given that this is a new case, it'd be nice to have such a comment.

src/test/regress/expected/failure_savepoints_0.out
```diff
@@ -0,0 +1,348 @@
...
+-----------
+ 
+(1 row)
+
+BEGIN;
+UPDATE artists SET name='a';
+SAVEPOINT s1;
+DELETE FROM artists WHERE id=4;
+RELEASE SAVEPOINT s1;
+WARNING:  AbortSubTransaction while in COMMIT state
```
onderkalaci: This looks suspicious. Could it be a bug in our savepoint implementation? (We have one more of this.)

It looks like we're first sending `RELEASE SAVEPOINT savepoint_2`. Later, after a failure, we try to rollback to `savepoint_2`

```SQL
LOG:  issuing RELEASE SAVEPOINT savepoint_2
LOG:  issuing RELEASE SAVEPOINT savepoint_2
LOG:  issuing RELEASE SAVEPOINT savepoint_2
LOG:  issuing RELEASE SAVEPOINT savepoint_2
LOG:  issuing RELEASE SAVEPOINT savepoint_2
WARNING:  AbortSubTransaction while in COMMIT state
LOG:  issuing ROLLBACK TO SAVEPOINT savepoint_2
WARNING:  could not issue cancel request
DETAIL:  Client error: PQcancel() -- connect() failed: Connection refused

WARNING:  connection not open
```

onderkalaci: I'm OK (and prefer) to not fix it right now. But we should open an issue if you think so as well.

pykello: COMMIT state means RELEASE started, but it didn't end. It is more like "COMMITTING" not "COMMITTED". So it doesn't look an issue to me.

onderkalaci: Noting what we've talked in person, the warning is coming from the coordinator, so look inevitable.

src/test/regress/expected/failure_savepoints_0.out
pykello: done.

PR #2785, opened 2019-06-21
failure_vacuum output (pg 10) for unified executor.

pykello:


src/test/regress/expected/failure_vacuum.out
```diff
@@ -35,37 +38,34 @@ SELECT citus.mitmproxy('conn.onQuery(query="^VACUUM").kill()');
 (1 row)
 
 VACUUM vacuum_test;
-ERROR:  connection error: localhost:9060
```
pykello: Note that the changes in this fail are just reverting it to the same version as master.

PR #2786, opened 2019-06-21
Improve one regression test

hanefi:
The test `multi_real_time_transaction` fails on enterprise. Instead of changing the test results on enterprise, I wanted to update the test slightly to make the community and enterprise version output the same.

## The problem
GRANT queries are propagated on Enterprise. If a user attempts to create a user and run a GRANT query before creating the user on workers, we fail. This issue does not happen in community as the user needs to run the GRANTs on the workers manually.

## Note
I also removed the PG9 test output as it is no longer needed.


hanefi:
I realize that I needed this PR only 13 hours after https://github.com/citusdata/citus/pull/2689 is merged.

@jasonmp85 For information sharing purposes, let me summarize what I changed here: 
- There was `GRANT` queries for a user that was not yet created on the workers. As the `GRANT`s are propagated to workers on enterprise, this test was failing. I changed the order of operations so that this test still works fine on enterprise as well
- I deleted the `multi_real_time_transaction_0.out` file as the outputs of PG11 and PG10 are the same, and there is no need to have it here. (Please correct me if I am missing something)

jasonmp85:
> I realize that I needed this PR only 13 hours after #2689 is merged.

Yeah, I hadn't merged to enterprise yet; sorry.

PR #2787, opened 2019-06-21
Fix WaitEventSetWait timeout in unified executor

marcocitus:
We currently determine the `WaitEventSetWait` timeout using `DeadlineTimestampTzToTimeout` where the deadline is the start time of `RunDistributedExecution` plus the node connection timeout. However, this has several issues a) the `RunDistributedExecution` timeout may not be aligned with the actual worker pool timeout, b) the timeout value converges to <= 0 which effectively disable the timeout (i.e. we might not wake up when the pool actually times out) and c) it ignores node connection timeouts that start later (after a failover).

This PR fixes the issue by going through all worker pools and determining the number of milliseconds until the connection timeout. There could be a slightly simpler solution, but this one can easily be extended to consider other time-based events such as slow start.

### DIFF
--- a/src/backend/distributed/executor/executor.c
+++ b/src/backend/distributed/executor/executor.c
@@ -193,6 +193,9 @@ typedef struct WorkerPool
 	 */
 	TimestampTz poolStartTime;
 
+	/* indicates whether to check for the connection timeout */
+	bool checkForPoolTimeout;
+
 	/*
 	 * This is only set in WorkerPoolFailed() function. Once a pool fails, we do not
 	 * use it anymore.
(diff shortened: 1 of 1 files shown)


src/backend/distributed/executor/executor.c
```diff
@@ -1658,36 +1665,110 @@ CheckConnectionTimeout(WorkerPool *workerPool)
...
+			ereport(logLevel, (errcode(ERRCODE_CONNECTION_FAILURE),
+							   errmsg("could not establish any connections to the node "
+									  "%s:%d after %u ms", workerPool->node->workerName,
+									  workerPool->node->workerPort,
+									  NodeConnectionTimeout)));
+		}
+		else
+		{
+			/* stop interrupting WaitEventSetWait for timeouts */
+			workerPool->survivedTimeout = true;
```
onderkalaci: Totally optional: I find slightly hard to follow the calculation and the name of `survivedTimeout `. It might be easier to read if we rename the flag to `checkForPoolTimeout` (or such) and revert the value of the flag.

In that case, we could even use that flag to early return in this function. Because with this change we're actually executing some of the checks of the function more often.

src/backend/distributed/executor/executor.c
```diff
@@ -1524,12 +1525,18 @@ ManageWorkerPool(WorkerPool *workerPool)
...
+	if (workerPool->failed)
+	{
+		/* connection pool failed */
+		return;
+	}
+
 	/* we might fail the execution or warn the user about connection timeouts */
 	CheckConnectionTimeout(workerPool);
 
 	if (failedConnectionCount >= 1)
```
onderkalaci: shall we move this check above
```C

if (workerPool->failed || failedConnectionCount >= 1)
{
    return
}
```

marcocitus: I figured there might be situations in which we want to check for timeout after one connection has failed, e.g. we open 4 connections and 1 fails and the second never responds.

PR #2789, opened 2019-06-21
Fix a null pointer caused by partial initialization of ConnParamsHashEntry

thanodnl:
DESCRIPTION: Fix null pointer caused by partial initialization of ConnParamsHashEntry

It has been reported a null pointer dereference could be triggered in `FreeConnParamsHashEntryFields`. Likely cause is an error in `GetConnParams` which will leave the cached `ConnParamsHashEntry` in a state that would cause the null pointer dereference in a subsequent connection establishment to the same server. This has been simulated by inserting `ereport(ERROR, ...)` at certain places in the code.

Not only would `ConnParamsHashEntry` be in a state that would cause a crash, it was also leaking memory in the `ConnectionContext` due to the loss of pointers as they are only stored on the `ConnParamsHashEntry` at the end of the function.

This patch rewrites both the `GetConnParams` to store pointers 'durably' at every point in the code so that an error would not lose the pointer as well as `FreeConnParamsHashEntryFields` in a way that it can clear half initialised `ConnParamsHashEntry`'s in a safer manner.

### DIFF
--- a/src/backend/distributed/connection/connection_configuration.c
+++ b/src/backend/distributed/connection/connection_configuration.c
@@ -270,7 +270,7 @@ GetConnParams(ConnectionHashKey *key, char ***keywords, char ***values,
 	 *
 	 * assignment is done directly to not loose the pointers if any of the later
 	 * allocations cause an error. FreeConnParamsHashEntryFields knows about the
-	 * possibility of half initialized keywords of values and correctly reclaims them when
+	 * possibility of half initialized keywords or values and correctly reclaims them when
 	 * the cache is reused.
 	 *
 	 * Need to zero enough space for all possible libpq parameters.
@@ -305,7 +305,7 @@ GetConnParams(ConnectionHashKey *key, char ***keywords, char ***values,
 
 	/*
 	 * remember where global/GUC params end and runtime ones start, all entries after this
-	 * point should be allocated in context and will be freed uppon
+	 * point should be allocated in context and will be freed upon
 	 * FreeConnParamsHashEntryFields
 	 */
 	*runtimeParamStart = ConnParams.size;
--- a/src/backend/distributed/connection/connection_configuration.c
+++ b/src/backend/distributed/connection/connection_configuration.c
@@ -229,12 +229,16 @@ void
 GetConnParams(ConnectionHashKey *key, char ***keywords, char ***values,
 			  Index *runtimeParamStart, MemoryContext context)
 {
-	/* make space for the port as a string: sign, 10 digits, NUL */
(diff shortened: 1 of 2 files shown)


src/backend/distributed/connection/connection_configuration.c
```diff
@@ -247,28 +251,34 @@ GetConnParams(ConnectionHashKey *key, char ***keywords, char ***values,
...
+		GetDatabaseEncodingName()
 	};
 
 	/*
-	 * Declare local params for readability; we'll assign to outparams at end.
+	 * Declare local params for readability;
+	 *
+	 * assignment is done directly to not loose the pointers if any of the later
+	 * allocations cause an error. FreeConnParamsHashEntryFields knows about the
+	 * possibility of half initialized keywords of values and correctly reclaims them when
```
thanodnl: s/keywords of values/keywords or values/

src/backend/distributed/connection/connection_configuration.c
```diff
@@ -293,26 +303,27 @@ GetConnParams(ConnectionHashKey *key, char ***keywords, char ***values,
 		connValues[paramIndex] = ConnParams.values[paramIndex];
 	}
 
-	/* remember where global/GUC params end and runtime ones start */
+	/*
+	 * remember where global/GUC params end and runtime ones start, all entries after this
+	 * point should be allocated in context and will be freed uppon
```
marcocitus: uppon -> upon

src/backend/distributed/connection/connection_management.c
```diff
@@ -958,23 +958,42 @@ StartConnectionEstablishment(ConnectionHashKey *key)
...
+	char **keyword = NULL;
+	char **value = NULL;
 
-	while (*keyword != NULL)
+	/*
+	 * if there was a memory error during the initialization of ConnParamHashEntry in
+	 * GetConnParams the keywords or values might not have been initialized completely.
+	 * We check if they have been initialized before freeing them.
+	 *
+	 * We only recursively free the lists starting at the index pointed to by
```
marcocitus: recursively -> iteratively?

src/backend/distributed/connection/connection_configuration.c
```diff
@@ -247,28 +251,34 @@ GetConnParams(ConnectionHashKey *key, char ***keywords, char ***values,
...
+		key->database,
+		key->user,
+		GetDatabaseEncodingName()
 	};
 
 	/*
-	 * Declare local params for readability; we'll assign to outparams at end.
+	 * Declare local params for readability;
+	 *
+	 * assignment is done directly to not loose the pointers if any of the later
```
serprex: s/loose/lose/

PR #2790, opened 2019-06-23
Change the order of placement access added to the placement access list

onderkalaci:
This is to make sure that the error messages related to foreign keys
to reference tables shows the exact placement access name instead of
SELECT.

This only changes the error message, doesn't have any other side effects.
The reason for the PR is to make sure that we give the same error messages
between the adaptive executor and the router executor.

### DIFF
--- a/src/test/regress/expected/foreign_key_restriction_enforcement.out
+++ b/src/test/regress/expected/foreign_key_restriction_enforcement.out
@@ -538,7 +538,7 @@ ROLLBACK;
 BEGIN;
 	ALTER TABLE on_update_fkey_table ADD COLUMN X int;
 	UPDATE referece_table SET id = 160 WHERE id = 15;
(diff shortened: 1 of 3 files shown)


src/test/regress/expected/foreign_key_restriction_enforcement.out
```diff
@@ -538,7 +538,7 @@ ROLLBACK;
 BEGIN;
 	ALTER TABLE on_update_fkey_table ADD COLUMN X int;
 	UPDATE referece_table SET id = 160 WHERE id = 15;
-ERROR:  cannot execute SELECT on reference relation "referece_table" because there was a parallel DDL access to distributed relation "on_update_fkey_table" in the same transaction
+ERROR:  cannot execute DML on reference relation "referece_table" because there was a parallel DDL access to distributed relation "on_update_fkey_table" in the same transaction
```
onderkalaci: Note that the error message is fixed now

marcocitus:
:-O All green

PR #2791, opened 2019-06-23
Switch between adaptive executor and the existing executors, also set the pool size

onderkalaci:
With this PR, we're adding the ability to switch between the executors. Also, making the pool size configurable.

We're enabling CircleCI for the existing executors as well. 

Although we've tried as much as possible to not have different output files by adjusting the tests, we had to add some, especially for failure tests. The convention we follow for the test output file names: Use `test_name.out` for the existing executors and use `test_table_9.out` for the adaptive executor.


src/test/regress/expected/failure_1pc_copy_append.out
```diff
@@ -161,24 +161,21 @@ SELECT * FROM pg_dist_shard s, pg_dist_shard_placement p
...
 -----------
  
 (1 row)
 
 SELECT count(1) FROM copy_test;
-WARNING:  connection error: localhost:9060
-DETAIL:  server closed the connection unexpectedly
-	This probably means the server terminated abnormally
-	before or while processing the request.
+WARNING:  could not consume data from worker node
```
onderkalaci: Note that the changes on `failure_test_name.out` files are actually setting the output files back to work fine with the existing executor since we had previously changed those to work OK with the adaptive executor.

src/backend/distributed/executor/executor.c
```diff
@@ -404,83 +405,78 @@ static TaskExecutionState TaskExecutionStateMachine(ShardCommandExecution *
...
- * On the first call, it executes the distributed query and writes the
- * results to a tuple store. The postgres executor calls this function
- * repeatedly to read tuples from the tuple store.
+ * AdaptiveExecutorExecScan is called via CitusExecScan on the
+ * first call of CitusExecScan. The function fills the tupleStore
+ * of the input scanScate.
  */
 TupleTableSlot *
-CitusExecScan(CustomScanState *node)
+AdaptiveExecutorExecScan(CustomScanState *node)
```
marcocitus: this should probably have a different name since its semantics are different from other ExecScan functions

src/backend/distributed/shared_library_init.c
```diff
@@ -719,6 +720,18 @@ RegisterCitusConfigVariables(void)
 		GUC_UNIT_KB,
 		NULL, NULL, NULL);
 
+	DefineCustomIntVariable(
+		"citus.max_pool_size",
```
marcocitus: should this be max_adaptive_executor_pool_size?

src/backend/distributed/shared_library_init.c
```diff
@@ -923,7 +936,7 @@ RegisterCitusConfigVariables(void)
 					 "queries that touch thousands of shards and/or that involve table "
 					 "repartitioning."),
 		&TaskExecutorType,
-		MULTI_EXECUTOR_REAL_TIME,
+		MULTI_ADAPTIVE_EXECUTOR,
```
marcocitus: do we still need to support pool size 0 if we you can disable the executor here?

src/backend/distributed/commands/utility_hook.c
```diff
@@ -573,7 +590,24 @@ ExecuteDistributedDDLJob(DDLJob *ddlJob)
...
+					ddlJob->executeSequentially)
+				{
+					ExecuteModifyTasksSequentiallyWithoutResults(ddlJob->taskList,
+																 CMD_UTILITY);
+				}
+				else
+				{
+					ExecuteModifyTasksWithoutResults(ddlJob->taskList);
+				}
+			}
```
marcocitus: the same block appears twice, can we wrap this into a function?

src/backend/distributed/executor/multi_server_executor.c
```diff
@@ -89,13 +112,13 @@ JobExecutorType(DistributedPlan *distributedPlan)
 	taskCount = list_length(job->taskList);
 	tasksPerNode = taskCount / ((double) workerNodeCount);
 
-	if (executorType == MULTI_EXECUTOR_REAL_TIME)
+	if (adaptiveExecutorEnabled || executorType == MULTI_EXECUTOR_REAL_TIME)
```
marcocitus: This seems a bit confusing given that most of the if statements below do not trigger when `adaptiveExecutorEnabled` is false.

.circleci/config.yml
```diff
@@ -130,3 +159,7 @@ workflows:
       - {test-11_check-iso-work-fol: {requires: [build]}}
       - {test-11_check-fol: {requires: [build]}}
       - {test-11_check-failure: {requires: [build]}}
+
```
onderkalaci: With these 3, I guess we're hitting our concurrency limit on CircleCI, so some tests wait until the others are finished.

I tried merging these 3 into a single step, then the whole step takes really long time.

So, I'm open for suggestions here.

src/backend/distributed/shared_library_init.c
marcocitus: Wouldn't MaxPoolSize == 0 become obsolete at the same time or sooner?

onderkalaci: I think there is a minor misunderstanding here.

You cannot disable adaptive executor with setting the taskExecutor to realtime with the current code. The only way is to set the pool size. This is just to make things to look a bit more compatible.

Do you think should we disable when this GUC is not set to adaptive?

marcocitus: I'd have a slight preference for that, but I don't feel so strongly about it that I want to delay things.

onderkalaci:
>I'm not a huge fan of using pool size 0 to disable instead of the task_executor_type, but otherwise this seems ok.

@marcocitus I changed the logic, the poolsize can be at least 1, and the executorType is adaptive. Do you want to have another look before merging?

src/test/regress/expected/with_transactions.out
```diff
@@ -85,17 +85,18 @@ UPDATE raw_table SET created_at = '2001-02-10 20:00:00' WHERE tenant_id IN (SELE
 DEBUG:  common table expressions are not supported in distributed modifications
```
pykello: Why did this file change? I assume this is with non-adaptive executor and its output should stay the same?

onderkalaci:
> @marcocitus I changed the logic, the poolsize can be at least 1, and the executorType is adaptive. Do you want to have another look before merging?

Wait, I think that'd make the regression tests change very badly. Because we've quite a lot of explicit `SET citus.task_executor_type TO 'real-time';`, so I'm not thinking to revert my last commit and keep the logic as -is.

Even if we want to do, I'll do it with a separate PR, this is already huge.

onderkalaci:
> Even if we want to do, I'll do it with a separate PR, this is already huge.


I'm merging this since @marcocitus already approved. I think @marcocitus's suggestion about enabling/disabling the executor makes sense. I'll do it with a separate PR.

PR #2792, opened 2019-06-24
Add executor.c comment

marcocitus:
Initial top comment for unified executor


src/backend/distributed/executor/executor.c
```diff
@@ -2,6 +2,86 @@
  *
  * executor.c
  *
+ * The adaptive executor executes a list of tasks (queries on shards) over
```
onderkalaci: In general, I really like how you described the executor here. I've some notes/questions:

- Given that I'm very familiar with the code, I can quickly follow the text here. Do you think would it make sense to get a quick review from someone else as well?
- Should we consider renaming `executor.c` to `adaptive_executor.c`. We'll at least have the task tracker for a while :) 
- Should we very briefly mention about the unusual interpretation of `SEQUENTIAL_CONNECTION` vs `SequentialRunDistributedExecution`? (Maybe only even refer to `ShouldRunTasksSequentially` function where we explain this to some extent)
- Would it make sense to mention about the failure handling here? We're handling failures differently for some cases compared to the other executions, and we probably have a very little comment that explains failure handling in general.

src/backend/distributed/executor/executor.c
```diff
@@ -2,6 +2,86 @@
...
+ * a connection pool per worker node. The results of the queries, if any,
+ * are written to a tuple store.
+ *
+ * Each connection pool (WorkerPool) and each connection (WorkerSession)
```
pykello: What do you think about adding couple of sentences before this paragraph to give high level definition of WorkerPool and WorkerSession and their relation briefly? Otherwise "connection pool == WorkerPool" is a bit confusing and people will need to look at how these structs are defined and used to make sense of it.

pykello: Also there is no reference to the pool size part of execution. Is that just a detail or is it one defining characteristic of this executor?

src/backend/distributed/executor/executor.c
pykello: >     * Should we very briefly mention about the unusual interpretation of `SEQUENTIAL_CONNECTION` vs `SequentialRunDistributedExecution`? (Maybe only even refer to `ShouldRunTasksSequentially` function where we explain this to some extent)

@onderkalaci What do you think about renaming `SEQUENTIAL_CONNECTION` to `SINGLE_CONNECTION_PER_WORKER` in another PR so it doesn't need explanation?

marcocitus: Added some note on sequential execution, failure handling, and did the rename.

src/backend/distributed/executor/executor.c
marcocitus: Pushed some fixes.

PR #2793, opened 2019-06-24
Show just coordinator plan in some test outputs.

pykello:
This is to make output of multi_select_distinct consistent.


marcocitus:
I like this way better!

PR #2794, opened 2019-06-24
Add ANALYZE to multi_select_distinct

marcocitus:
Trying to see if this stabilizes multi_select_distinct.


onderkalaci:
what about #2793? In those tests, we're only interested in the coordinator plan, so should be fine?

onderkalaci:
I think analyze wouldn't (always) help here, we've tried other things like loading more data or creating index, which failed as well.

PR #2796, opened 2019-06-25
Fix memory issues in ReceiveResults().

pykello:
Unlike StoreQueryResult() in router executor which is called once per execution, adaptive executor's
ReceiveResults() is called repeatedly in the execution. So we should free the memory used there
to avoid memory leak.

### DIFF
--- a/src/backend/distributed/executor/adaptive_executor.c
+++ b/src/backend/distributed/executor/adaptive_executor.c
@@ -2680,11 +2680,13 @@ PlacementAccessListForTask(Task *task, ShardPlacement *taskPlacement)
 
 /*
  * ReceiveResults reads the result of a command or query and writes returned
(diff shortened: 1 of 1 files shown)


src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -2694,7 +2696,8 @@ ReceiveResults(WorkerSession *session, bool storeRows)
 		if (result == NULL)
 		{
 			/* no more results */
-			return true;
+			fetchDone = true;
```
onderkalaci: maybe `no more results` -> `no more results, do not immediately return because we should free some memory before` (or probably a better comment)

because it is not obvious that why we're deferring the return here.

pykello: done

PR #2781, opened 2019-06-20
Refactor ExecuteModifyTasksSequentially.

pykello:
1. Introduce `ExecuteModifyTasksSequentially`, for benefit of https://github.com/citusdata/citus/pull/2780
2. Make `RouterSequentialModifyExecScan` use `ExecuteModifyTasksSequentially`.
3. Make `CreatePlacementAccess` public, for benefit of #2780.

PR #2770, opened 2019-06-17
create_distributed_table goes through the executor

onderkalaci:
Based on #2702. Mostly do two extra things (a) Move as many changes as possible to master (b) Make it simple to switch between executor create_distributed_table and command based create distributed table so that #2762 wouldn't be too large.

PR #2798, opened 2019-06-25
Introduce the adaptive executor

onderkalaci:
DESCRIPTION: Adds a new distributed executor: Adaptive Executor

(copy & paste from the source code)

```
 *
 * adaptive_executor.c
 *
 * The adaptive executor executes a list of tasks (queries on shards) over
 * a connection pool per worker node. The results of the queries, if any,
 * are written to a tuple store.
 *
 * The concepts in the executor are modelled in a set of structs:
 *
 * - DistributedExecution:
 *     Execution of a Task list over a set of WorkerPools.
 * - WorkerPool
 *     Pool of WorkerSessions for the same worker which opportunistically
 *     executes "unassigned" tasks from a queue.
 * - WorkerSession:
 *     Connection to a worker that is used to execute "assigned" tasks
 *     from a queue and may execute unasssigned tasks from the WorkerPool.
 * - ShardCommandExecution:
 *     Execution of a Task across a list of placements.
 * - TaskPlacementExecution:
 *     Execution of a Task on a specific placement.
 *     Used in the WorkerPool and WorkerSession queues.
 *
 * Every connection pool (WorkerPool) and every connection (WorkerSession)
 * have a queue of tasks that are ready to execute (readyTaskQueue) and a
 * queue/set of pending tasks that may become ready later in the execution
 * (pendingTaskQueue). The tasks are wrapped in a ShardCommandExecution,
 * which keeps track of the state of execution and is referenced from a
 * TaskPlacementExecution, which is the data structure that is actually
 * added to the queues and describes the state of the execution of a task
 * on a particular worker node.
 *
 * When the task list is part of a bigger distributed transaction, the
 * shards that are accessed or modified by the task may have already been
 * accessed earlier in the transaction. We need to make sure we use the
 * same connection since it may hold relevant locks or have uncommitted
 * writes. In that case we "assign" the task to a connection by adding
 * it to the task queue of specific connection (in
 * AssignTasksToConnections). Otherwise we consider the task unassigned
 * and add it to the task queue of a worker pool, which means that it
 * can be executed over any connection in the pool.
 *
 * A task may be executed on multiple placements in case of a reference
 * table or a replicated distributed table. Depending on the type of
 * task, it may not be ready to be executed on a worker node immediately.
 * For instance, INSERTs on a reference table are executed serially across
 * placements to avoid deadlocks when concurrent INSERTs take conflicting
 * locks. At the beginning, only the "first" placement is ready to execute
 * and therefore added to the readyTaskQueue in the pool or connection.
 * The remaining placements are added to the pendingTaskQueue. Once
 * execution on the first placement is done the second placement moves
 * from pendingTaskQueue to readyTaskQueue. The same approach is used to
 * fail over read-only tasks to another placement.
 *
 * Once all the tasks are added to a queue, the main loop in
 * RunDistributedExecution repeatedly does the following:
 *
 * For each pool:
 * - ManageWorkPool evaluates whether to open additional connections
 *   based on the number unassigned tasks that are ready to execute
 *   and the targetPoolSize of the execution.
 *
 * Poll all connections:
 * - We use a WaitEventSet that contains all (non-failed) connections
 *   and is rebuilt whenever the set of active connections or any of
 *   their wait flags change.
 *
 *   We almost always check for WL_SOCKET_READABLE because a session
 *   can emit notices at any time during execution, but it will only
 *   wake up WaitEventSetWait when there are actual bytes to read.
 *
 *   We check for WL_SOCKET_WRITEABLE just after sending bytes in case
 *   there is not enough space in the TCP buffer. Since a socket is
 *   almost always writable we also use WL_SOCKET_WRITEABLE as a
 *   mechanism to wake up WaitEventSetWait for non-I/O events, e.g.
 *   when a task moves from pending to ready.
 *
 * For each connection that is ready:
 * - ConnectionStateMachine handles connection establishment and failure
 *   as well as command execution via TransactionStateMachine.
 *
 * When a connection is ready to execute a new task, it first checks its
 * own readyTaskQueue and otherwise takes a task from the worker pool's
 * readyTaskQueue (on a first-come-first-serve basis).
 *
 * In cases where the tasks finish quickly (e.g. <1ms), a single
 * connection will often be sufficient to finish all tasks. It is
 * therefore not necessary that all connections are established
 * successfully or open a transaction (which may be blocked by an
 * intermediate pgbouncer in transaction pooling mode). It is therefore
 * essential that we take a task from the queue only after opening a
 * transaction block.
 *
 * When a command on a worker finishes or the connection is lost, we call
 * PlacementExecutionDone, which then updates the state of the task
 * based on whether we need to run it on other placements. When a
 * connection fails or all connections to a worker fail, we also call
 * PlacementExecutionDone for all queued tasks to try the next placement
 * and, if necessary, mark shard placements as inactive. If a task fails
 * to execute on all placements, the execution fails and the distributed
 * transaction rolls back.
 *
 * For multi-row INSERTs, tasks are executed sequentially by
 * SequentialRunDistributedExecution instead of in parallel, which allows
 * a high degree of concurrency without high risk of deadlocks.
 * Conversely, multi-row UPDATE/DELETE/DDL commands take aggressive locks
 * which forbids concurrency, but allows parallelism without high risk
 * of deadlocks. Note that this is unrelated to SEQUENTIAL_CONNECTION,
 * which indicates that we should use at most one connection per node, but
 * can run tasks in parallel across nodes. This is used when there are
 * writes to a reference table that has foreign keys from a distributed
 * table.
 *
 * Execution finishes when all tasks are done, the query errors out, or
 * the user cancels the query.
 *
 *-------------------------------------------------------------------------
 */
```


src/test/regress/pg_regress_multi.pl
```diff
@@ -354,6 +356,11 @@ sub revert_replace_postgres
   push(@pgOptions, '-c', "wal_level=replica");
 }
 
+if ($maxPoolSize == 0)
+{
+  push(@pgOptions, '-c', "citus.max_pool_size=0");
```
onderkalaci: ops, I've changed the GUC name but not changed here. I realized from the code-cov tool, so I'll fix it. So, the code-coverage is pretty important tool :)

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -0,0 +1,3302 @@
...
+	}
+
+	firstTask = (Task *) linitial(taskList);
+
+	return !ReadOnlyTask(firstTask->taskType);
+}
+
+
+static bool
+DistributedExecutionRequiresRollback(DistributedExecution *execution)
```
marcocitus: Missing function comment

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -0,0 +1,3302 @@
...
+	AcquireExecutorShardLocks(execution);
+}
+
+
+/*
+ *  DistributedExecutionModifiesDatabase returns true if the execution modifies the data
+ *  or the schema.
+ */
+static bool
+DistributedExecutionModifiesDatabase(DistributedExecution *execution)
```
marcocitus: It might be nice to move these utility functions to the bottom of the file, or a separate file where applicable.

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -0,0 +1,3302 @@
...
+
+			case MULTI_CONNECTION_CONNECTING:
+			{
+				PostgresPollingStatusType pollMode;
+
+				ConnStatusType status = PQstatus(connection->pgConn);
+				if (status == CONNECTION_OK)
+				{
+					elog(DEBUG4, "established connection to %s:%d", connection->hostname,
+						 connection->port);
```
marcocitus: it would be useful to add the session ID here so we can relate it to other debug messages.

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -0,0 +1,3302 @@
...
+				/*
+				 * Note: We may get the same connection for multiple task placements.
+				 * FindOrCreateWorkerSession ensures that we only have one session per
+				 * connection.
+				 */
+				WorkerSession *session =
+					FindOrCreateWorkerSession(workerPool, connection);
+
+				elog(DEBUG4, "%s:%d has an assigned task", connection->hostname,
+					 connection->port);
```
marcocitus: maybe also add the session ID

src/backend/distributed/executor/adaptive_executor.c
onderkalaci: I'm not sure if we're still following the conventions, but didn't we have a convention of defining a function right after its first use?

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -0,0 +1,3410 @@
...
+			 */
+			CloseConnection(connection);
+		}
+		else if (transactionState == REMOTE_TRANS_CLEARING_RESULTS)
+		{
+			/* TODO: should we keep this in the state machine? or cancel? */
+			ClearResults(connection, false);
+		}
+
+		/* TODO: can we be in SENT_COMMAND? */
```
marcocitus: Do you think we should do anything about the TODO's here?

marcocitus:
failure_ddl is a little wobbly.

marcocitus:
It would be good to list all relevant PRs somewhere since the discussion here will be weirdly short. Not urgent.

onderkalaci:
> It would be good to list all relevant PRs somewhere since the discussion here will be weirdly short. Not urgent.

Makes sense, I think I'll do it once I'm back because the list is really long :)

src/backend/distributed/executor/adaptive_executor.c
onderkalaci: I don't think so, I think I'll push a commit to remove those comments unless you have further  comments.

The reason the execution could only finish:
    - Connection State of `MULTI_CONNECTION_CONNECTING` or `MULTI_CONNECTION_FAILED` or `MULTI_CONNECTION_CONNECTED`.
    - In the first `if` statement we handle `MULTI_CONNECTION_CONNECTING` and `MULTI_CONNECTION_FAILED` 
    - If the state is  `MULTI_CONNECTION_CONNECTED` the `TransactionStateMachine` ensures that we're on `REMOTE_TRANS_STARTED` or `REMOTE_TRANS_INVALID` in the end of the state transitions (note that on any failures/unexpected things we'd be back on `MULTI_CONNECTION_FAILED` state)

onderkalaci: (also, check-full doesn't hit anything other lines/states that I explained above.)

onderkalaci: Hmm, wait, the transaction might be in `REMOTE_TRANS_CLEARING_RESULTS` because if the session has sent `BEGIN` but there is not enough tasks to assign to that, and so never actually got to the point where it is fully cleared. So, we need to keep the  `REMOTE_TRANS_CLEARING_RESULTS`. 

Do you see any other such edge cases?

onderkalaci:
> It would be good to list all relevant PRs somewhere since the discussion here will be weirdly short. Not urgent.

Hi all,

We've merged all the changes for the adaptive executor. There are quite a lot of PRs went into either the master branch or the adaptive executor development branch. 

### Feature implementations for the adaptive executor:
•	#2660: Make sure that master_modify_multiple_shards() relies on the adaptive executor
•	#2666, #2764, #2800: Make sure that the adaptive executor can mimic real-time/router-executors by using 1 connection per placement. Useful for testing and potentially some customers might want for certain cases where the first command in a transaction block does require very few resources whereas the following commands require a lot more. In this case, we give the opportunity to mimic the old executors by opening 1 connection per placement
•	 #2672: Make sure that parametrized queries (e.g., prepared statements) works fine with the adaptive executor
•	#2682: Sequential execution for adaptive executor
•	#2688, #2692, #2790: Foreign keys to reference table for adaptive executor
•	#2724: Stop using master_modify_multiple_shards in TRUNCATE
•	#2725: Deprecate master_modify_multiple_shards
•	#2734: Make sure that TRUNCATE relies on the adaptive executor
•	#2735: Implement multi-row INSERTs for adaptive executor 
•	#2741: Adaptive executor honors citus.node_connection_timeout
•	#2746: INSERT .. SELECT via coordinator for adaptive executor
•	#2752: Implement executor locks for adaptive executor
•	#2756: Implement partitioning locks for adaptive executor
•	#2768: Implement SELECT FOR UPDATE for adaptive executor
•	#2769, #2770: create_distributed_table() relies on adaptive executor
•	#2779: Add slow start interval to the executor
•	#2791, #2799, #2804: Switch between adaptive executor and the existing executors, also set the pool size
•	#2798: Introduce the adaptive executor [The main PR]

### Failure handling for adaptive executor:
•	#2664: Don't schedule tasks on inactive nodes
•	#2673, #2676, #2774: Fix some issues with failures while using SAVEPOINTs
•	#2685: 2PC related improvements
•	#2706: Improve failure scenarios for COPY
•	#2713: Main PR for implementing failure handling for adaptive executor, touches quite a bit of things
•	#2795: Some improvements for replication > 1
•	#2809: Update warnings in FinishRemoteTransactionCommit()

### COPY for adaptive executor:
•	#2745: Refactor code to get ready for implementing COPY to work fine with adaptive executor
•	#2780: Make COPY compatible with adaptive executor

### Improve connection management for adaptive executor:
•	#2694: Improve closing connections
•	#2700, #2730, #2763, #2766, #2783, #2787: WaitEventSet related improvements
•	#2716: Make the number of cached connections configurable
•	#2772: Use SendCancelationRequest() in ShutdownConnection()
•	#2808: Refactor session cleanup

### Misc changes:
•	#2701: Get rid of compiler warnings
•	#2719, #2792, #2801: Improve code comments
•	#2722, #2748, #2781: Some code refactoring
•	#2740, #2767, #2782, #2812, #2816: Some bugfixes that are introduced during development of adaptive executor
•	#2656: Rename a field to prevent conflicts with the adaptive executor concepts

### Ensure regression test produce consistent outputs – mostly add ORDER BYs or minor test adjustments
•	#2657, #2569, #2661, #2671, #2678, #2683, #2695, #2696, #2720, #2736, #2810: Purely adding ORDER BYs to the regression tests to get consistent test results
•	#2662, #2721, #2728, #2729, #2738, #2788, #2793, #2802, #2805, #2815, #2817: Various improvements in the regression tests
•	#2677, #2681, #2754: Add ability to sort the output of RETURNING, only used for getting consistent output results for regression tests
•	#2690, #2765, #2775, #2797, #2838: Add ability to normalize shardIds/taskIds/placementIds etc for the regression tests

PR #2799, opened 2019-06-26
Use TaskExecutorType to switch between executors

onderkalaci:
Previously we were checking MaxAdaptiveExecutorPoolSize to switch to adaptive executor. Instead, check `citus.task_executor_type`.

As a result, the min value of `MaxAdaptiveExecutorPoolSize` could be 1.


src/test/regress/expected/multi_mx_explain.out
```diff
@@ -66,7 +66,7 @@ Sort
   Sort Key: COALESCE((pg_catalog.sum((COALESCE((pg_catalog.sum(remote_scan.count_quantity))::bigint, '0'::bigint))))::bigint, '0'::bigint), remote_scan.l_quantity
   ->  HashAggregate
         Group Key: remote_scan.l_quantity
-        ->  Custom Scan (Citus Real-Time)
+        ->  Custom Scan (Citus Adaptive)
```
onderkalaci: We already ignore the executor names in the explain outputs. Just made it adaptive since that's the default

PR #2800, opened 2019-06-26
Force max parallelization to create shards if asked to use exclusive connections

pykello:
We are doing so to make creating distributed tables fast when they contain local data or inside a transaction block.

### DIFF
--- a/src/backend/distributed/executor/adaptive_executor.c
+++ b/src/backend/distributed/executor/adaptive_executor.c
@@ -3394,3 +3394,17 @@ BuildWaitEventSet(List *sessionList)
 
 	return waitEventSet;
 }
+
+
+/*
+ * SetLocalForceMaxQueryParallelization simply a C interface for
+ * setting the following:
+ *      SET LOCAL citus.multi_shard_modify_mode TO on;
+ */
+void
+SetLocalForceMaxQueryParallelization(void)
+{
+	set_config_option("citus.force_max_query_parallelization", "on",
+					  (superuser() ? PGC_SUSET : PGC_USERSET), PGC_S_SESSION,
+					  GUC_ACTION_LOCAL, true, 0, false);
+}
--- a/src/backend/distributed/master/master_stage_protocol.c
+++ b/src/backend/distributed/master/master_stage_protocol.c
@@ -534,8 +534,7 @@ CreateShardsOnWorkersViaExecutor(Oid distributedRelationId, List *shardPlacement
 
 	int taskId = 1;
 	List *taskList = NIL;
-
-	int poolSize = useExclusiveConnection ? MaxAdaptiveExecutorPoolSize : 1;
+	int poolSize = 1;
 
 	foreach(shardPlacementCell, shardPlacements)
 	{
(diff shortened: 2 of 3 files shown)


pykello:
Looking into why failure tests fail ...

src/backend/distributed/master/master_stage_protocol.c
```diff
@@ -562,7 +562,37 @@ CreateShardsOnWorkersViaExecutor(Oid distributedRelationId, List *shardPlacement
...
+		poolSize = MaxAdaptiveExecutorPoolSize;
+	}
+
+	PG_TRY();
+	{
+		ExecuteTaskList(CMD_UTILITY, taskList, poolSize);
+	}
+	PG_CATCH();
+	{
+		ForceMaxQueryParallelization = prevForceMaxQueryParallelization;
```
marcocitus: I think we can avoid the PG_TRY by using set_config_option (SET LOCAL), see SetLocalMultiShardModifyModeToSequential.

onderkalaci: Maybe double check that `useExclusiveConnection ` is also set true if `StoredProcedureLevel>1` or `FunctionCallLevel>1`. Can we safely use set local in those cases?

pykello: > Maybe double check that `useExclusiveConnection ` is also set true if `StoredProcedureLevel>1` or `FunctionCallLevel>1`. Can we safely use set local in those cases?

I tried `SET LOCAL` in stored procedures and functions and they work as expected.

src/backend/distributed/master/master_stage_protocol.c
```diff
@@ -570,6 +569,24 @@ CreateShardsOnWorkersViaExecutor(Oid distributedRelationId, List *shardPlacement
 		taskList = lappend(taskList, task);
 	}
 
+	if (useExclusiveConnection)
+	{
+		/*
+		 * When the table has local data, we force max parallelization so data
+		 * copy is done efficiently.
+		 */
+		SetLocalForceMaxQueryParallelization();
```
onderkalaci: We could get here even if we're not inside a transaction block. And, actually `SET LOCAL` gives a WARNING executed outside of a transaction block (`WARNING:  SET LOCAL can only be used in transaction blocks`). However, the warning is given from one level higher function of `set_config_option `.

Still, reading the postgres code, I think there is no harm to call `set_config_option ` even if we're not inside a transaction block. It also does what we want.

@marcocitus any opinions on this? Note that with `SetLocalMultiShardModifyMode()` we're always in a transaction block.

marcocitus: I think it should be ok.

src/backend/distributed/master/master_stage_protocol.c
```diff
@@ -570,6 +569,24 @@ CreateShardsOnWorkersViaExecutor(Oid distributedRelationId, List *shardPlacement
...
+
+		/*
+		 * TODO: we shouldn't need to set poolSize when setting ForceMaxQueryParallelization.
+		 *
```
marcocitus: so should we remove this or ...?

pykello: I updated the comment to make it more comprehensible. Currently adaptive executor ignores ForceMaxQueryParallelization for recording parallel access and uses poolSize > 1 check instead. Making adaptive executor to take into account ForceMaxQueryParallelization isn't as easy as thought, since it seems that then some TRUNCATE related issues arise. So we just set poolSize here to force recording parallel access, which is a hacky way to do it, and should be removed in future once we have the proper checks in adaptive executor.

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -3394,3 +3394,17 @@ BuildWaitEventSet(List *sessionList)
...
+ * setting the following:
+ *      SET LOCAL citus.multi_shard_modify_mode TO on;
+ */
+void
+SetLocalForceMaxQueryParallelization(void)
+{
+	set_config_option("citus.force_max_query_parallelization", "on",
+					  (superuser() ? PGC_SUSET : PGC_USERSET), PGC_S_SESSION,
+					  GUC_ACTION_LOCAL, true, 0, false);
+}
```
onderkalaci: Maybe add a new line at the end of the file?

src/backend/distributed/master/master_stage_protocol.c
```diff
@@ -570,6 +569,26 @@ CreateShardsOnWorkersViaExecutor(Oid distributedRelationId, List *shardPlacement
 		taskList = lappend(taskList, task);
 	}
 
+	if (useExclusiveConnection)
+	{
+		/*
+		 * When the table has local data, we force max parallelization so data
```
onderkalaci: I think we're doing it in two cases: local table has data or we're inside a transaction block.

So, maybe add `We also prefer to use max parallelization when we're inside a transaction block because the user might execute compute heavy commands later in the transaction block.`

PR #2801, opened 2019-06-26
Improve comments & debug messages

onderkalaci:
Expand some debug messages to include the sessionId and add some missing function comments.

### DIFF
--- a/src/backend/distributed/executor/adaptive_executor.c
+++ b/src/backend/distributed/executor/adaptive_executor.c
@@ -879,6 +879,12 @@ TaskListModifiesDatabase(CmdType operation, List *taskList)
 }
 
 
+/*
+ * DistributedExecutionRequiresRollback returns true if the distributed
+ * execution should start a CoordinatedTransaction. In other words, if the
+ * function returns true, the execution sends BEGIN; to every connection
+ * involved in the distributed execution.
+ */
 static bool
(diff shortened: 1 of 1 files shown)


src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -2530,6 +2536,18 @@ PopUnassignedPlacementExecution(WorkerPool *workerPool)
 }
 
 
+/*
+ * StartPlacementExecutionOnSession gets a TaskPlacementExecition and
+ * WorkerSession, and as the name suggests, the task's query is sent
+ * to the worker via the session.
+ *
+ * The function does some bookkeeping such as associating the placement
+ * accesses with the connection and updating sesion's local variables. For
```
marcocitus: Session

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -2530,6 +2536,18 @@ PopUnassignedPlacementExecution(WorkerPool *workerPool)
 }
 
 
+/*
+ * StartPlacementExecutionOnSession gets a TaskPlacementExecition and
+ * WorkerSession, and as the name suggests, the task's query is sent
```
marcocitus: I think the "and as the name suggests" is more redundant than just repeating the name in words :)

Issue #2776, opened 2019-06-19
Modifying CTEs in router queries do recursive planning


serprex:
Router planner should support things like
```
EXPLAIN
WITH t AS (SELECT id from test_id where tenant_id=1),
 t2 as (UPDATE test_id ti SET id=1 FROM t where ti.id=t.id and ti.tenant_id=1 returning ti.id, t.id)
	select * from t2;
```
without breaking query up through recursive planning. See https://github.com/citusdata/citus/issues/2400#issuecomment-425837423 for things to think about

See https://github.com/citusdata/citus/blob/6741ffd716593f09b22a01dab70f64d7c62922cd/src/backend/distributed/planner/multi_router_planner.c#L3077

marcocitus:
As a first step we should improve function evaluation (citus_clauses.c) to only evaluate volatile functions in contexts where they only need to be evaluated once (e.g. VALUES). That way, we can safely run function evaluation on a SELECT even if it contains modifying CTEs.

PR #2807, opened 2019-06-27
DistributedPlan: replace operation with modLevel

serprex:
Also rename ExtactInsertRangeTableEntry to ExtractResultRelationRTE,
as the source of this function didn't match the documentation

This causes no behaviorial changes, only organizes better to implement modifying CTEs (#2776)

### DIFF
--- a/src/backend/distributed/executor/adaptive_executor.c
+++ b/src/backend/distributed/executor/adaptive_executor.c
@@ -158,8 +158,8 @@
  */
 typedef struct DistributedExecution
 {
-	/* the corresponding distributed plan's operation */
-	CmdType operation;
+	/* the corresponding distributed plan's modLevel */
+	RowModifyLevel modLevel;
 
 	List *tasksToExecute;
 
@@ -504,7 +504,7 @@ int ExecutorSlowStartInterval = 10;
 
 
 /* local functions */
-static DistributedExecution * CreateDistributedExecution(CmdType operation,
+static DistributedExecution * CreateDistributedExecution(RowModifyLevel modLevel,
 														 List *taskList,
 														 bool hasReturning,
 														 ParamListInfo paramListInfo,
@@ -520,16 +520,16 @@ static void FinishDistributedExecution(DistributedExecution *execution);
 static void CleanUpSessions(DistributedExecution *execution);
 
 static void LockPartitionsForDistributedPlan(DistributedPlan *distributedPlan);
-static void AcquireExecutorShardLocks(DistributedExecution *execution);
+static void AcquireExecutorShardLocksForExecution(DistributedExecution *execution);
 static bool DistributedExecutionModifiesDatabase(DistributedExecution *execution);
 static bool DistributedPlanModifiesDatabase(DistributedPlan *plan);
-static bool TaskListModifiesDatabase(CmdType operation, List *taskList);
+static bool TaskListModifiesDatabase(RowModifyLevel modLevel, List *taskList);
 static bool DistributedExecutionRequiresRollback(DistributedExecution *execution);
-static bool SelectForUpdateOnReferenceTable(CmdType operation, List *taskList);
+static bool SelectForUpdateOnReferenceTable(RowModifyLevel modLevel, List *taskList);
 static void AssignTasksToConnections(DistributedExecution *execution);
 static void UnclaimAllSessionConnections(List *sessionList);
 static bool UseConnectionPerPlacement(void);
-static PlacementExecutionOrder ExecutionOrderForTask(CmdType operation, Task *task);
+static PlacementExecutionOrder ExecutionOrderForTask(RowModifyLevel modLevel, Task *task);
 static WorkerPool * FindOrCreateWorkerPool(DistributedExecution *execution,
 										   WorkerNode *workerNode);
 static WorkerSession * FindOrCreateWorkerSession(WorkerPool *workerPool,
@@ -610,7 +610,7 @@ AdaptiveExecutor(CustomScanState *node)
 		targetPoolSize = 1;
 	}
 
-	execution = CreateDistributedExecution(distributedPlan->operation, taskList,
+	execution = CreateDistributedExecution(distributedPlan->modLevel, taskList,
 										   distributedPlan->hasReturning,
 										   paramListInfo, tupleDescriptor,
 										   tupleStore, targetPoolSize);
@@ -626,7 +626,7 @@ AdaptiveExecutor(CustomScanState *node)
 		RunDistributedExecution(execution);
 	}
 
-	if (distributedPlan->operation != CMD_SELECT)
+	if (distributedPlan->modLevel != ROW_MODIFY_READONLY)
 	{
 		executorState->es_processed = execution->rowsProcessed;
 	}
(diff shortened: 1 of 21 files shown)


src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -158,106 +158,106 @@ AcquireMetadataLocks(List *taskList)
...
-		 * A RowExclusiveLock does not conflict with itself and therefore allows
-		 * multiple commutative commands to proceed concurrently. It does
-		 * conflict with ExclusiveLock, which may still be obtained by another
-		 * session that executes an UPDATE/DELETE/UPSERT command with
-		 * citus.all_modifications_commutative disabled.
-		 */
 
-		lockMode = RowExclusiveLock;
-	}
-	else if (task->upsertQuery || commandType == CMD_UPDATE || commandType == CMD_DELETE)
```
serprex: Need to make sure `modLevel == MODLEVEL_NONCOMMUTATIVE` when `task->upsertQuery`

src/include/distributed/multi_physical_planner.h
```diff
@@ -110,6 +110,15 @@ typedef enum
 } BoundaryNodeJobType;
 
 
+/* Enumeration that specifies extent of DML modifications */
+typedef enum ModifyLevel
+{
+	MODLEVEL_NONE = 0,
```
marcocitus: We tend to shun acronyms and shorthands, so maybe MODIFY_LEVEL_.

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -158,106 +158,106 @@ AcquireMetadataLocks(List *taskList)
...
  */
 static void
-AcquireExecutorShardLock(Task *task, CmdType commandType)
+AcquireExecutorShardLock(Task *task, ModifyLevel modLevel)
 {
-	LOCKMODE lockMode = NoLock;
 	int64 shardId = task->anchorShardId;
 
-	if (commandType == CMD_SELECT)
+	if (shardId != INVALID_SHARD_ID)
```
marcocitus: Could we do an early return if `shardId == INVALID_SHARD_ID`? Long code blocks are hard to follow and this makes it seem like we may have a task without a shard ID that does have a relevant `relationRowLockList` or for which `RequiresConsistentSnapshot` is true, which I don't think can be the case.

src/backend/distributed/executor/multi_router_executor.c
marcocitus: Hmz, seems like we're missing some test coverage.

src/backend/distributed/master/master_truncate.c
```diff
@@ -79,7 +79,7 @@ citus_truncate_trigger(PG_FUNCTION_ARGS)
 
 		if (MultiShardConnectionType == SEQUENTIAL_CONNECTION)
 		{
-			ExecuteModifyTasksSequentiallyWithoutResults(taskList, CMD_UTILITY);
+			ExecuteModifyTasksSequentiallyWithoutResults(taskList, MODLEVEL_NONE);
```
marcocitus: This reads a little strangely, since we're clearly doing a modification here. I wonder whether we should have a MODIFY_LEVEL_DDL to indicate that this is a non-commutative modification, but we don't need special locks because it takes an AccessExclusiveLock on the table.

src/backend/distributed/executor/multi_router_executor.c
marcocitus: Discussed this offline, it seems we may want to leverage `relationRowLockList` for CTEs.

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -4310,6 +4309,41 @@ GenerateSyntheticShardIntervalArray(int partitionCount)
...
+	switch (query->commandType)
+	{
+		case CMD_SELECT:
+		{
+			return ROW_MODIFY_READONLY;
+		}
+
+		case CMD_INSERT:
+		{
+			if (query->onConflict != NULL)
```
pykello: I'm a bit confused here. Why is an `INSERT ... ON CONFLICT ...` is commutative? Am I misunderstanding this condition?

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -4310,6 +4309,41 @@ GenerateSyntheticShardIntervalArray(int partitionCount)
...
+			{
+				return ROW_MODIFY_COMMUTATIVE;
+			}
```
pykello: Minor: without an `else` and with the `FALLTHROUGH` I need to scan more lines to understand this. I'd have preferred an `else` block here.

src/backend/distributed/planner/multi_physical_planner.c
serprex: Right, originally the condition was `== NULL` & then in a daze looking into why things weren't working I read it thinking "That's not right" & flipped to `!= NULL`

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -457,19 +457,12 @@ IsMultiTaskPlan(DistributedPlan *distributedPlan)
...
 {
-	CmdType commandType = distributedPlan->operation;
-
-	if (commandType == CMD_UPDATE || commandType == CMD_DELETE)
-	{
-		return true;
-	}
-
-	return false;
+	return distributedPlan->modLevel >= ROW_MODIFY_NONCOMMUTATIVE;
```
marcocitus: I think the only caller of this function is `IsMultiShardModifyPlan`, which is quite specifically looking for multi-shard UPDATE/DELETE and not multi-shard upserts. It might be ok to continue checking command type for now, at least until we can push down CTEs in multi-shard statements.

For context: `IsMultiShardModifyPlan` function is called in `CreateDistributedPlannedStmt` to avoid planning prepared update/delete command that has a `distribution column = $1` filter as a multi-shard update/delete when it could be a single-shard update/delete if the parameter was resolved. For multi-shard upserts we can tolerate parameters in the distribution column since we can defer shard pruning to the executor, so you could see a slight performance improvement when using a prepared statement.

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -968,16 +968,11 @@ DistributedExecutionRequiresRollback(DistributedExecution *execution)
  * that contains FOR UPDATE clause that locks any reference tables.
  */
 static bool
-SelectForUpdateOnReferenceTable(CmdType operation, List *taskList)
+SelectForUpdateOnReferenceTable(RowModifyLevel modLevel, List *taskList)
 {
 	Task *task = NULL;
 	ListCell *rtiLockCell = NULL;
 
-	if (operation != CMD_SELECT)
```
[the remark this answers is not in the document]
serprex: I'd like to change `SELECT FOR UPDATE` to be `ROW_MODIFY_NONCOMMUTATIVE` but don't currently know enough to know if that'd make sense in the end

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -1402,7 +1401,7 @@ ExecutionOrderForTask(CmdType operation, Task *task)
 
 		case MODIFY_TASK:
 		{
-			if (operation == CMD_INSERT && !task->upsertQuery)
+			if (modLevel >= ROW_MODIFY_NONCOMMUTATIVE)
```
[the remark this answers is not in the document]
serprex: Reading over it, this didn't make sense at first. Found comment at top explaining that the weaker locks made by inserts open them up to deadlock, so it makes sense to change this to `< ROW_MODIFY_COMMUTATIVE` as opposed to `== ROW_MODIFY_COMMUTATIVE`. Just wanting to make sure the mental model of what `RowModifyLevel` isn't breaking down here

serprex:
Squashed. This is ready to merge now

PR #2808, opened 2019-06-28
Improve code structure of session clean up at the end of execution

onderkalaci:
Explicitly list all the states that the execution might end,
    otherwise warn.


src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -1130,13 +1128,45 @@ CleanUpSessions(DistributedExecution *execution)
...
+				 * on MULTI_CONNECTION_FAILED state.
+				 */
+				elog(WARNING, "unexpected transaction state at the end of execution: %d",
+					 transactionState);
+			}
+		}
+		else
+		{
+			elog(WARNING, "unexpected connection state at the end of execution: %d",
+				 connection->connectionState);
```
marcocitus: I guess we usually handle it immediately, but I'm a little worried that we can let the connection sit in a state like `MULTI_CONNECTION_LOST` and do not notice it during tests.

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -1130,13 +1128,45 @@ CleanUpSessions(DistributedExecution *execution)
...
 
-		/* TODO: can we be in SENT_COMMAND? */
+			if (transactionState == REMOTE_TRANS_CLEARING_RESULTS)
+			{
+				/*
+				 * We might have established the connection, and even sent BEGIN, but not
+				 * get to the point where we assigned a task to this specific connection
+				 * (because other connections in the pool already finished all the tasks).
+				 */
+				Assert(session->commandsSent == 0);
```
marcocitus: It may be useful to spell out the important assumption we're making in the function comment: That we only reach `CleanUpSessions` after successfully completing all the tasks and we expect no tasks are still in progress. If we ever decide to change our logic, this function will need to be revised. (crazy example: run SELECT tasks in parallel and see which returns first)

src/backend/distributed/executor/adaptive_executor.c
onderkalaci: Sorry, I couldn't follow your concern for this. `MULTI_CONNECTION_LOST ` is immediately goes to state `MULTI_CONNECTION_FAILED`, also `MULTI_CONNECTION_INITIAL` goes to `MULTI_CONNECTION_CONNECTED`.

There is one another state `MULTI_CONNECTION_WAIT_RETRY` which is never used in the code and I don't know what is that.  Do you think would that be a problem?

marcocitus: I don't think there is a problem here. It's just to make the code more robust. The loop in the StateMachine functions is just an optimisation that I'd rather not rely on.

Some other code path could start using the `MULTI_CONNECTION_LOST` without realising that the adaptive executor is not tolerant to getting a connection in that state from the connection API. Perhaps we already have that possibility somewhere (e.g. an OOM causes a rollback to savepoint), and users get flushed with warnings in production which we never see in testing.

I think we can probably remove `MULTI_CONNECTION_WAIT_RETRY` from the unified_executor branch. It does not exist on master.

onderkalaci: Then, I guess it is safer to `CloseConnection` for `MULTI_CONNECTION_LOST`, similar to `MULTI_CONNECTION_FAILED` ?

PR #2809, opened 2019-06-28
Update warnings in FinishRemoteTransactionCommit()

pykello:
To fix inconsistent failure_ddl outputs

### DIFF
--- a/src/backend/distributed/transaction/remote_transaction.c
+++ b/src/backend/distributed/transaction/remote_transaction.c
@@ -303,17 +303,8 @@ FinishRemoteTransactionCommit(MultiConnection *connection)
 
 		if (transaction->transactionState == REMOTE_TRANS_1PC_COMMITTING)
 		{
-			if (transaction->transactionCritical)
-			{
-				ereport(WARNING, (errmsg("failed to commit critical transaction "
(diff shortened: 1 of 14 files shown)


src/backend/distributed/transaction/remote_transaction.c
```diff
@@ -303,17 +303,8 @@ FinishRemoteTransactionCommit(MultiConnection *connection)
...
-				ereport(WARNING, (errmsg("failed to commit critical transaction "
-										 "on %s:%d, metadata is likely out of sync",
-										 connection->hostname, connection->port)));
-			}
-			else
-			{
-				ereport(WARNING, (errmsg("failed to commit transaction on %s:%d",
-										 connection->hostname, connection->port)));
-			}
+			ereport(WARNING, (errmsg("failed to commit transaction on %s:%d",
```
marcocitus: I wonder, are we supposed to set transactionFailed here? Not sure we want to change it right now, but it's a little odd that we don't.

pykello: We seem to handle it in the `HandleRemoteTransactionResultError()` called few lines before this change.

marcocitus: Thanks! Missed that.

PR #2810, opened 2019-06-28
Add more ORDER BYs to multi_mx_partitioning

pykello:
Tom make its output consistent


marcocitus:
I think at this point we can probably save some headaches by piling up on the unified_executor branch.

PR #2812, opened 2019-07-02
Fix a use after free in adaptive executor

pykello:
`connection` is freed when it is failed in which case we shouldn't access it anymore. Also, we are done with the execution, so no need to set `waitFlagsChanged`.

### DIFF
--- a/src/backend/distributed/executor/adaptive_executor.c
+++ b/src/backend/distributed/executor/adaptive_executor.c
@@ -1171,16 +1171,15 @@ CleanUpSessions(DistributedExecution *execution)
 				elog(WARNING, "unexpected transaction state at the end of execution: %d",
 					 transactionState);
 			}
+
+			/* get ready for the next executions if we need use the same connection */
+			connection->waitFlags = WL_SOCKET_READABLE | WL_SOCKET_WRITEABLE;
 		}
 		else
 		{
 			elog(WARNING, "unexpected connection state at the end of execution: %d",
 				 connection->connectionState);
 		}
-
-		/* get ready for the next executions if we need use the same connection */
-		connection->waitFlags = WL_SOCKET_READABLE | WL_SOCKET_WRITEABLE;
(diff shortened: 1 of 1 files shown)


src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -1171,16 +1171,15 @@ CleanUpSessions(DistributedExecution *execution)
 				elog(WARNING, "unexpected transaction state at the end of execution: %d",
 					 transactionState);
 			}
+
+			/* get ready for the next executions if we need use the same connection */
+			connection->waitFlags = WL_SOCKET_READABLE | WL_SOCKET_WRITEABLE;
 		}
 		else
 		{
 			elog(WARNING, "unexpected connection state at the end of execution: %d",
```
pykello: This is unreachable since we never set connection state to `MULTI_CONNECTION_INITIAL`. I'd prefer to replace with an `Assert()`, but maybe we should postpone that to another PR.

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -1171,16 +1171,15 @@ CleanUpSessions(DistributedExecution *execution)
...
 				 connection->connectionState);
 		}
-
-		/* get ready for the next executions if we need use the same connection */
-		connection->waitFlags = WL_SOCKET_READABLE | WL_SOCKET_WRITEABLE;
-		execution->waitFlagsChanged = true;
```
pykello: No need for this, we are done with the execution.

marcocitus: That's not quite true for multi-row inserts. We don't need to set waitFlagsChanged because we will rebuild the WaitEventSet anyway. I do wonder where we set the wait flags to writeable before the next execution.

marcocitus: Ah, it's the line you added :)

Issue #2820, opened 2019-07-04
Worker node cache related memory issues


pykello:
In valgrind reports we see:

```
==11659== VALGRINDERROR-BEGIN
==11659== Invalid read of size 1
==11659==    at 0x4C317F9: __strncmp_sse42 (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==11659==    by 0x5F8A9F7: WorkerNodeCompare (worker_node_manager.c:499)
==11659==    by 0x5F74572: FindOrCreateWorkerPool (adaptive_executor.c:1447)
==11659==    by 0x5F74901: AssignTasksToConnections (adaptive_executor.c:1254)
==11659==    by 0x5F75FD6: RunDistributedExecution (adaptive_executor.c:1616)
==11659==    by 0x5F76332: SequentialRunDistributedExecution (adaptive_executor.c:1595)
==11659==    by 0x5F7643D: AdaptiveExecutor (adaptive_executor.c:622)
==11659==    by 0x5F76867: CitusExecScan (citus_custom_scan.c:192)
==11659==    by 0x64A09F: ExecCustomScan (nodeCustom.c:114)
==11659==    by 0x63BA9E: ExecProcNodeFirst (execProcnode.c:445)
==11659==    by 0x63444A: ExecProcNode (executor.h:247)
==11659==    by 0x63444A: ExecutePlan (execMain.c:1723)
==11659==    by 0x635003: standard_ExecutorRun (execMain.c:364)
==11659==  Address 0x5f097a0 is 31,504 bytes inside a block of size 32,712 free'd
==11659==    at 0x4C2EDEB: free (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==11659==    by 0x8C1400: AllocSetReset (aset.c:610)
==11659==    by 0x8C9828: MemoryContextResetOnly (mcxt.c:174)
==11659==    by 0x8C0FF2: AllocSetDelete (aset.c:652)
==11659==    by 0x8C9B8E: MemoryContextDelete (mcxt.c:245)
==11659==    by 0x8A9AF4: hash_destroy (dynahash.c:828)
==11659==    by 0x5FBCBF1: InitializeWorkerNodeCache (metadata_cache.c:2911)
==11659==    by 0x5FBCFC7: PrepareWorkerNodeCache (metadata_cache.c:2823)
==11659==    by 0x5FBE9E1: GetWorkerNodeHash (metadata_cache.c:2794)
[...]

PR #2821, opened 2019-07-04
Don't keep a pointer to a worker node cache entry in adaptive executor

pykello:
Worker node cache entries can be invalidated when looking up for other worker nodes in next steps of execution. Calling `FindWorkerNode()` can invalidate previous results of `FindWorkerNode()`.

That is, `FindWorkerNode()` calls `GetWorkerNodeHash()` which calls `PrepareWorkerNodeCache()` which calls `AcceptInvalidationMessages()` which causes the `WorkerNodeCache` to be destroyed and all pointers to its entries are invalid.

Fixes https://github.com/citusdata/citus/issues/2820

I double checked that we are not keeping `WorkerNode` pointers in this manner in other places.

### DIFF
--- a/src/backend/distributed/executor/adaptive_executor.c
+++ b/src/backend/distributed/executor/adaptive_executor.c
@@ -290,7 +290,7 @@ typedef struct WorkerPool
 	DistributedExecution *distributedExecution;
 
 	/* worker node on which we have a pool of sessions */
-	WorkerNode *node;
+	WorkerNode node;
 
 	/* all sessions on the worker that are part of the current execution */
 	List *sessionList;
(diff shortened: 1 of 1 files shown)


src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -1444,14 +1444,14 @@ FindOrCreateWorkerPool(DistributedExecution *execution, WorkerNode *workerNode)
...
-		if (WorkerNodeCompare(workerPool->node, workerNode, 0) == 0)
+		if (WorkerNodeCompare(&workerPool->node, workerNode, 0) == 0)
 		{
 			return workerPool;
 		}
 	}
 
 	workerPool = (WorkerPool *) palloc0(sizeof(WorkerPool));
-	workerPool->node = workerNode;
+	memcpy(&workerPool->node, workerNode, sizeof(WorkerNode));
```
marcocitus: I'm a little concerned that other (perhaps less frequent) calls to FindWorkerNode are similarly broken and that if we ever add a pointer to WorkerNode we'll need to do so in many places. What do you think about always copying the data structure?

https://github.com/citusdata/citus/pull/new/fix_find_worker_node

pykello: > I'm a little concerned that other (perhaps less frequent) calls to FindWorkerNode are similarly broken and that if we ever add a pointer to WorkerNode we'll need to copy it in many places. What do you think about always copying the data structure? e.g. #2822

Sounds good. I had checked that there is no other broken call to FindWorkerNode.

PR #2829, opened 2019-07-09
Prototype: Delegate procedure calls to worker nodes

marcocitus:
Citus currently always executes stored procedures on the coordinator node, which results in high response times and low throughput since every command in the procedure involves one or more network round-trip between the coordinator and workers and incurs planning/execution overhead, even if all the commands go to the same worker node.

This PR implements a basic mechanism for defining distributed functions and procedures using `distribute_procedure`. The procedure is initally replicated to all nodes and has a distribution argument which is used for routing calls to worker nodes with metadata (à la MX). The worker node is chosen based on a the value in the "distribution argument" according to the distribution of a co-located distributed table.

The idea is that the stored procedure would have a tenant ID argument and the node executing the procedure is the one storing the data for that tenant, to avoid network round-trips. Because the worker node has the full metadata, it is still able to handle multi-shard queries within the procedure in case not all queries have distribution column filters or there are writes to a reference table. That way, we avoid having to do careful analysis of the function to ensure it only touches co-located shards.

I'm mainly pushing this branch to do some benchmarks and opening the PR for information sharing purposes.

Currently only CALL is delegated to the worker nodes, the function is not yet created on `master_activate_node`, there is no way to undistribute a function, and the metadata is not updated after changes to a function (e.g. DROP FUNCTION).

I'm also experimenting a bit with storing the metadata in an upgrade-safe manner, since the OIDs of the functions may not be preserved during pg_upgrade. The `proargtypes` still need to be added to uniquely identify the function and renames and schema changes need te be handled. We may want to consider a simpler solution (e.g. a pre-upgrade script).

### DIFF
--- a/src/backend/distributed/metadata/procedure_metadata.c
+++ b/src/backend/distributed/metadata/procedure_metadata.c
@@ -1,0 +1,412 @@
+/*
+ * procedure_metadata.c
+ *   Functions to load metadata on distributed stored procedures.
+ */
+#include "postgres.h"
+#include "miscadmin.h"
+#include "funcapi.h"
+
+
+#include "access/genam.h"
+#include "access/heapam.h"
+#include "access/htup.h"
+#include "access/htup_details.h"
+#include "access/skey.h"
+#include "access/tupmacs.h"
+#include "access/xact.h"
+#include "catalog/indexing.h"
+#include "catalog/namespace.h"
+#include "catalog/pg_namespace.h"
+#include "catalog/pg_proc.h"
+#include "commands/sequence.h"
+#include "distributed/citus_procedures.h"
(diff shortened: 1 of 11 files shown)


src/backend/distributed/metadata/procedure_metadata.c
```diff
@@ -0,0 +1,412 @@
...
+		heap_close(pgDistColocation, ExclusiveLock);
+	}
+
+	return colocationId;
+}
+
+
+/*
+ * InsertDistributedProcedureRecord inserts a record into citus.procedures recording
+ * that the function should be distributed and
```
serprex: and...

marcocitus:
Superseded by #2965

PR #2830, opened 2019-07-09
Prototype: Execute query locally if shards are local

marcocitus:
In #2829 we're delegating function calls to worker nodes, but we're still making connections to localhost. incurring the overhead of another backend and significantly lowering the achievable concurrency and throughput.

This PR plans and executes tasks that are local to the current node via the regular planner after rewriting the query tree to have shard names instead of table names (without re-parsing the query). This needs to happen after initial function evaluation (to do shard pruning) and locking. 

Currently the local tasks are executed prior to running `AdaptiveExecutor`. Strictly speaking they should be executed during `AdaptiveExecutor` to preserve execution order to avoid deadlock in multi-row inserts. On the other hand, we do not strictly enforce deadlock prevention in MX for multi-shard UPDATE/DELETE either and deadlocks in multi-row insert should be rare.

Pushing this branch mainly to do some benchmarks alongside #2829 and opening the PR for information sharing purposes.

### DIFF
--- a/src/backend/distributed/commands/multi_copy.c
+++ b/src/backend/distributed/commands/multi_copy.c
@@ -345,6 +345,7 @@ CitusCopyFrom(CopyStmt *copyStatement, char *completionTag)
 	}
 
 	XactModificationLevel = XACT_MODIFICATION_DATA;
+	LocalTaskExecutionLevel = LOCAL_EXECUTION_DISALLOWED;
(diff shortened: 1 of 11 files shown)


onderkalaci:
Should we close this since #2938 merged?

marcocitus:
Superseded by #2938

PR #2831, opened 2019-07-09
Fix an issue with subquery map merge jobs as non-root

thanodnl:
DESCRIPTION: Fix an issue with subquery map merge jobs as non-root

Also automated all manual tests around multi user isolation for internal citus udf's

commits:
- automate upgrade_to_reference_table tests
- add negative tests for lock_relation_if_exists
- add tests for permissions on worker_cleanup_job_schema_cache
- add tests for worker_fetch_partition_file
- add tests for worker_merge_files_into_table
- fix problem with worker_merge_files_and_run_query when run as non-super user and add tests for behaviour

### DIFF
--- a/src/backend/distributed/worker/worker_merge_protocol.c
+++ b/src/backend/distributed/worker/worker_merge_protocol.c
@@ -175,6 +175,8 @@ worker_merge_files_and_run_query(PG_FUNCTION_ARGS)
 	int createMergeTableResult = 0;
 	int createIntermediateTableResult = 0;
 	int finished = 0;
+	Oid savedUserId = InvalidOid;
+	int savedSecurityContext = 0;
 	Oid userId = GetUserId();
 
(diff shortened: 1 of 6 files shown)


src/backend/distributed/worker/worker_merge_protocol.c
```diff
@@ -221,11 +223,17 @@ worker_merge_files_and_run_query(PG_FUNCTION_ARGS)
 							   createMergeTableQuery)));
 	}
 
+	/* need superuser to copy from files */
```
thanodnl: @marcocitus can you have a look at this? I ran into issues running this udf as non-root and it differed here compared to the call to `CopyTaskFilesFromDirectory` in [`worker_merge_files_into_table`](https://github.com/citusdata/citus/blob/master/src/backend/distributed/worker/worker_merge_protocol.c#L130-L137)

PR #2842, opened 2019-07-11
(WIP) Refactor COPY implementation.

pykello:
Trying to remove duplicate code from `multi_copy.c` ....


pykello:
I don't like this refactoring which is bunch of conditionals here and there. I'll think about this more.

Issue #2834, opened 2019-07-10
Collapse SQL scripts to previous major release


marcocitus:
We currently keep the full history of ALTER EXTENSION patches in the master branch, but there is no need for keeping anything up to the last major release that supported one of the current PostgreSQL versions. The 6.x line only supported PostgreSQL <=9.6 and the 8.x line only supports PostgreSQL>=10, so there is no world in which a user would upgrade from 6.x to 8.2 without first upgrading to a 7.0 release branch. It is therefore sufficient for the 7.x branches to have the upgrade scripts from 6.0 onwards. The 8.x branches only need upgrade scripts from 7.0 onwards, and so on.

Collapsing all scripts up to 6.0 would allow us to clean up the main script, collapse function re-definitions, and remove functions that are no longer used. This would allow us to go from 107 files to 53.

We could also consider collapsing the upgrade scripts for older minor releases into a single file per release. We keep multiple files mainly in case someone runs an intermediate version (built from master), which only really applies to our developers while working on the current release.

PR #2844, opened 2019-07-15
Postgres 12

### LINKED ISSUE #2755: Citus and PostgreSQL 12
Hi,

PostgreSQL 12 beta is out, and I started rolling out RPMs against it. Citus currently does not support v12. Do you have a timeline for that?

Thanks!

Regards, Devrim

serprex:
DESCRIPTION: Add support for Postgres 12

Closes #2755 #2834

Failing 3 tests because of #2873


serprex:
nb while this is passing tests, that's only showing that my shims are successfully not breaking 10/11. Version 12 is still chock full of errors (though with the last commit, we do get past `CREATE EXTENSION`)

src/test/regress/sql/multi_extension.sql
```diff
@@ -55,62 +55,7 @@ DROP EXTENSION citus;
...
-ALTER EXTENSION citus UPDATE TO '6.1-13';
-ALTER EXTENSION citus UPDATE TO '6.1-14';
-ALTER EXTENSION citus UPDATE TO '6.1-15';
-ALTER EXTENSION citus UPDATE TO '6.1-16';
-ALTER EXTENSION citus UPDATE TO '6.1-17';
-ALTER EXTENSION citus UPDATE TO '6.2-1';
-ALTER EXTENSION citus UPDATE TO '6.2-2';
-ALTER EXTENSION citus UPDATE TO '6.2-3';
-ALTER EXTENSION citus UPDATE TO '6.2-4';
-ALTER EXTENSION citus UPDATE TO '7.0-1';
```
pykello: Why are we making these changes?

src/backend/distributed/relay/relay_event_utility.c
```diff
@@ -755,8 +756,10 @@ AppendShardIdToName(char **name, uint64 shardId)
 	neededBytes = snprintf((*name), NAMEDATALEN, "%s", extendedName);
 	if (neededBytes < 0)
 	{
+		char *strerrno = strerror(errno);
+
 		ereport(ERROR, (errcode(ERRCODE_OUT_OF_MEMORY),
-						errmsg("out of memory: %s", strerror(errno))));
+						errmsg("out of memory: %s", strerrno)));
```
pykello: We can use `%m` here.
```suggestion
						errmsg("out of memory: %m")));
```

src/backend/distributed/utils/citus_readfuncs.c
```diff
@@ -45,9 +45,15 @@ CitusSetTag(Node *node, int tag)
 	nodeTypeName *local_node = (nodeTypeName *) CitusSetTag((Node *) node, T_##nodeTypeName)
 
 /* And a few guys need only the pg_strtok support fields */
+#if PG_VERSION_NUM >= 120000
 #define READ_TEMP_LOCALS()	\
-	char	   *token;		\
+	const char	*token;		\
```
pykello: What happens if we use `const` also in pre-12?

src/backend/distributed/worker/worker_sql_task_protocol.c
```diff
@@ -251,7 +252,8 @@ TaskFileDestReceiverReceive(TupleTableSlot *slot, DestReceiver *dest)
 static void
 WriteToLocalFile(StringInfo copyData, File fileDesc)
 {
-	int bytesWritten = FileWrite(fileDesc, copyData->data, copyData->len, PG_WAIT_IO);
+	int bytesWritten = FileWriteCompat(fileDesc, copyData->data, copyData->len, 0,
```
pykello: Passing `offset = 0` doesn't seem right here. In `TaskFileDestReceiverReceive` we might not be in the beginning of file.

src/test/regress/sql/multi_extension.sql
pykello: > Why are we making these changes?

I had missed that we are also addressing #2834 here. What do you think moving these changes to a separate PR? My reason being that we can do better than concatenating files, and avoiding many of the `ALTER … ` commands there, which can also simplify some of the C code. For example some UDFs are called when some columns are added, so we needed to special handle those.

src/include/distributed/version_compat.h
```diff
@@ -240,5 +244,67 @@ RangeVarGetRelidInternal(const RangeVar *relation, LOCKMODE lockmode, uint32 fla
...
+
+#define fcSetArg(fc, n, argval) \
+	((fc)->args[n].isnull = false, (fc)->args[n].value = (argval))
+#define fcSetArgNull(fc, n) \
+	((fc)->args[n].isnull = true, (fc)->args[n].value = (Datum) 0)
+#define InitFunctionCallInfoDataCompat InitFunctionCallInfoData
+
+#else /* pre PG12 */
+#define QTW_EXAMINE_RTES_BEFORE QTW_EXAMINE_RTES
+#define MakeSingleTupleTableSlotCompat(tupleDesc, tts_opts) \
```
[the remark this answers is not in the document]
serprex: There's code which throws errors when we don't pass `&TTSOpsMinimalTuple` (such as the extension check). However I'm still getting errors about minimal tuples being expected when I replaced all callers to specifying `TTSOpsMinimalTuple` so missing a part of this puzzle

src/backend/distributed/utils/citus_readfuncs.c
serprex: In 12 we get warnings that token is receiving `const`. Before 12 we get warnings that `const` is being discarded

src/test/regress/sql/multi_extension.sql
serprex: Splitting up this PR makes sense. I can split this to another branch/PR, & then the optimizations you have in mind can be follow up commits

serprex:
At this point there are 69 failing tests for `check-multi`. I've summarized the regressions:

ignore:
`TABLESPACE pg_default` omitted from `CREATE INDEX` -- regex
Floating point representation changed -- clever regex?
Version changed -- ignore how?
\d+ says "Partitioned table" for "Table" -- regex
'invalid input syntax for XX' became 'for type XX' -- regex
sql_procedure: drop cascade list ordering changed -- ignore how?
foreign key generated names now include '_ref' -- regex

investiate:
Complex window function test has different dates
`pg_indexes` has new entry `partitioning_test | partitioning_index`
missing a result in set_operations_and_local_tables
2 DEBUG messages omitted in subquery_in_where
multi_subquery_complex_reference_clause, multi_subquery_in_where_reference_clause: 'cannot pushdown query' now gets computed
'failed to execute task N' lots of these, don't know what they are
multi_remove_node_reference_table: colocationid doesn't seem to've taken to be set
multi_multiuser: received copy data after EOF marker

doubling issue:
non_colocated_join_order went from 9 to 36
multi_complex_expression: quadrupling behavior
multi_view et al: doubling behavior

other:
limit_intermediate_size: 'intermediate result size exceeds' -- I think this isn't triggering because CTEs are subqueries? Either that or pg's using less memory. If CTEs are subqueries, can say MATERIALIZE or use cte multiple times to prevent materialize
WITH OIDS -- remove tests?
intermediate_results: `EXPLAIN` costs varying -- shouldn't use `COSTS OFF`? otherwise maybe I broke some cost computation code, I did modify an cost function

mkurz:
Out of interest: PostgreSQL 12 will support foreign keys that reference partitioned tables, [see article](https://www.depesz.com/2019/04/24/waiting-for-postgresql-12-support-foreign-keys-that-reference-partitioned-tables/). How does this influence Citus? Will Citus make use of this (somehow)?

serprex:
@mkurz that's outside the scope of this task

serprex:
Overview of `_2.out` tests
test|reason|howtonot2
-+-+-
`multi_create_shards`|Uses `WITH OIDS`|Remove `WITH OIDS` from tests
`multi_metadata_sync`|`fk_test_col1_fkey` becomes `fk_test_col1_col2_fkey`|Give constraint explicit name, or add regex to make table spacing the same
`multi_null_minmax_value_pruning`|selects pg version
`multi_partitioning`|selects pg version
`multi_partitioning_utils`|selects pg version
`multi_reference_table`|Uses `WITH OIDS`|Remove `WITH OIDS` from tests
`multi_task_assignment_policy`|Task assignment ordering changed|Don't log task number (or regex it out)
`multi_utility_statement`|Uses `WITH OIDS`|Remove `WITH OIDS` from tests
`replicated_partitioned_table`|selects pg version|Exactly the same otherwise
`sql_procedure`|drop cascades list order changed

To be removed:
* `multi_master_protocol` TABLESPACE regex should fix

A possible fix to these reordering things: add to the postprocessor logic which reorders adjacent lines if they share some prefix length & their reordering would fix diff

src/include/distributed/version_compat.h
```diff
@@ -240,5 +244,67 @@ RangeVarGetRelidInternal(const RangeVar *relation, LOCKMODE lockmode, uint32 fla
...
+	(((fc)->argnull[n] = false), ((fc)->arg[n] = (value)))
+#define fcSetArgNull(fc, n) \
+	(((fc)->argnull[n] = true), ((fc)->arg[n] = (Datum) 0))
+#define InitFunctionCallInfoDataCompat(fc, fn, nargs, collation, ctx, result) \
+	InitFunctionCallInfoData(fc, fn, nargs, collation, ctx, result)
+
+#define FileReadCompat(file, buffer, amount, offset, wait_event_info) \
+	FileRead(file, buffer, amount, wait_event_info)
+#define FileWriteCompat(file, buffer, amount, offset, wait_event_info) \
+	FileWrite(file, buffer, amount, wait_event_info)
```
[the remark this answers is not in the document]
serprex: Waiting for tests to be passing before thinking about this

src/include/distributed/version_compat.h
```diff
@@ -240,5 +244,67 @@ RangeVarGetRelidInternal(const RangeVar *relation, LOCKMODE lockmode, uint32 fla
...
+#define GetSysCacheOid2Compat(cacheId, oidcol, key1, key2) \
+	GetSysCacheOid2(cacheId, key1, key2)
+#define GetSysCacheOid3Compat(cacheId, oidcol, key1, key2, key3) \
+	GetSysCacheOid3(cacheId, key1, key2, key3)
+#define GetSysCacheOid4Compat(cacheId, oidcol, key1, key2, key3, key4) \
+	GetSysCacheOid4(cacheId, key1, key2, key3, key4)
+#define LOCAL_FCINFO(name, nargs) \
+	FunctionCallInfoData name ## data; \
+	FunctionCallInfoData *name = &name ## data
+#define make_aggCompat(tlist, qual, aggstrategy, aggsplit, numGroupCols, grpColIdx, \
```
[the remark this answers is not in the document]
serprex: grpCollations introduced in PG12 with this commit: https://github.com/postgres/postgres/commit/5e1963fb764e9cc092e0f7b58b28985c311431d9#diff-bcead21a2e56eb3d56947f8f7aa8f622

serprex:
At this point the remaining test failures have underlying issues in pg11 that happen to not get hit by test suite

src/include/distributed/version_compat.h
```diff
@@ -240,5 +244,67 @@ RangeVarGetRelidInternal(const RangeVar *relation, LOCKMODE lockmode, uint32 fla
...
+	NextCopyFrom(cstate, econtext, values, nulls, NULL)
+#define GetSysCacheOid1Compat(cacheId, oidcol, key1) \
+	GetSysCacheOid1(cacheId, key1)
+#define GetSysCacheOid2Compat(cacheId, oidcol, key1, key2) \
+	GetSysCacheOid2(cacheId, key1, key2)
+#define GetSysCacheOid3Compat(cacheId, oidcol, key1, key2, key3) \
+	GetSysCacheOid3(cacheId, key1, key2, key3)
+#define GetSysCacheOid4Compat(cacheId, oidcol, key1, key2, key3, key4) \
+	GetSysCacheOid4(cacheId, key1, key2, key3, key4)
+#define LOCAL_FCINFO(name, nargs) \
```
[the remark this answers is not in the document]
serprex: ## data appends `data` to the identifier passed as `name`. This identifier mashup matches how PG12's [LOCAL_FCINFO](https://github.com/postgres/postgres/blob/07b39083c2aca003c4b1f289d7dc2368b5e2297a/src/include/fmgr.h#L110) works

src/include/distributed/version_compat.h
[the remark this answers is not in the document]
serprex: Implemented. The only place we weren't using `O_TRUNC` happens to be directly after creating a directory with a random number in the name

serprex:
This currently includes commits from #2858 & #2859 so that the only failing tests are due to `pg_dist_colocation` not being distinct on (shards, replicas, type) & thus pg12 picks a different colocationid than pg11

src/backend/distributed/commands/vacuum.c
```diff
@@ -22,9 +22,162 @@
...
+ */
+static VacOptTernaryValue
+get_vacopt_ternary_value(DefElem *def)
+{
+	return defGetBoolean(def) ? VACOPT_TERNARY_ENABLED : VACOPT_TERNARY_DISABLED;
+}
+
+
+static int
+VacuumStmt_options(VacuumStmt *vacstmt)
```
pykello: The order we follow is usually declarations -> definitions, and we have forward declarations for every function. In older times that we were stricter in style, the definitions followed their use. That is `VacuumStmt_options()`, etc. should be defined after the place it is used. The idea being that when we read the file from top to down, we first read the higher level concepts firsts and then continue to the lower level details.

src/backend/distributed/commands/create_distributed_table.c
```diff
@@ -1230,7 +1237,7 @@ CopyLocalDataIntoShards(Oid distributedRelationId)
 
 	/* get the table columns */
 	tupleDescriptor = RelationGetDescr(distributedRelation);
-	slot = MakeSingleTupleTableSlot(tupleDescriptor);
+	slot = MakeSingleTupleTableSlotCompat(tupleDescriptor, &TTSOpsHeapTuple);
```
onderkalaci: Do we need `TTSOpsHeapTuple ` or is it OK to use `TTSOpsMinimalTuple `?

Would you mind describing the difference between both?

src/backend/distributed/commands/vacuum.c
```diff
@@ -22,9 +22,162 @@
...
+		params.freeze_table_age = -1;
+		params.multixact_freeze_min_age = -1;
+		params.multixact_freeze_table_age = -1;
+	}
+
+	/* user-invoked vacuum is never "for wraparound" */
+	params.is_wraparound = false;
+
+	/* user-invoked vacuum never uses this parameter */
+	params.log_min_duration = -1;
```
marcocitus: Setting all these fields seems a little odd given that we only return one of them.

src/backend/distributed/commands/create_distributed_table.c
serprex: `ExecStoreHeapTuple` errors if it doesn't get `TTSOpsHeapTuple`

src/backend/distributed/executor/citus_custom_scan.c
```diff
@@ -162,6 +162,11 @@ CitusBeginScan(CustomScanState *node, EState *estate, int eflags)
 	MarkCitusInitiatedCoordinatorBackend();
 
 	scanState = (CitusScanState *) node;
+
+#if PG_VERSION_NUM >= 120000
```
serprex: This is necessary because pg12 gives us a `TTSOpsVirtual` which doesn't use with tuplestore

src/backend/distributed/executor/multi_executor.c
```diff
@@ -138,7 +138,9 @@ CitusExecutorRun(QueryDesc *queryDesc,
 		EState *estate = queryDesc->estate;
 
 		estate->es_processed = 0;
+#if PG_VERSION_NUM < 120000
 		estate->es_lastoid = InvalidOid;
```
serprex: Removed in pg12

src/include/distributed/version_compat.h
```diff
@@ -16,7 +16,11 @@
 #include "catalog/namespace.h"
 #include "nodes/parsenodes.h"
 
-#if (PG_VERSION_NUM >= 100000 && PG_VERSION_NUM < 110000)
+#if (PG_VERSION_NUM >= 120000)
+#include "optimizer/optimizer.h"
```
serprex: PG12 consolidated many declarations to `optimizer/optimizer.h`

src/include/distributed/version_compat.h
```diff
@@ -240,5 +244,132 @@ RangeVarGetRelidInternal(const RangeVar *relation, LOCKMODE lockmode, uint32 fla
...
+#define GetSysCacheOid3Compat GetSysCacheOid3
+#define GetSysCacheOid4Compat GetSysCacheOid4
+#define make_aggCompat make_agg
+
+#define fcSetArg(fc, n, argval) \
+	(((fc)->args[n].isnull = false), ((fc)->args[n].value = (argval)))
+	#define fcSetArgNull(fc, n) \
+	(((fc)->args[n].isnull = true), ((fc)->args[n].value = (Datum) 0))
+
+typedef struct
```
serprex: PG12 uses `pread` & `pwrite`. These APIs require the consumer to track offset themselves

src/include/distributed/version_compat.h
```diff
@@ -240,5 +244,132 @@ RangeVarGetRelidInternal(const RangeVar *relation, LOCKMODE lockmode, uint32 fla
...
+#define GetSysCacheOid1Compat(cacheId, oidcol, key1) \
+	GetSysCacheOid1(cacheId, key1)
+#define GetSysCacheOid2Compat(cacheId, oidcol, key1, key2) \
+	GetSysCacheOid2(cacheId, key1, key2)
+#define GetSysCacheOid3Compat(cacheId, oidcol, key1, key2, key3) \
+	GetSysCacheOid3(cacheId, key1, key2, key3)
+#define GetSysCacheOid4Compat(cacheId, oidcol, key1, key2, key3, key4) \
+	GetSysCacheOid4(cacheId, key1, key2, key3, key4)
+
+#define LOCAL_FCINFO(name, nargs) \
```
serprex: See [LOCAL_FCINFO in postgres](https://github.com/postgres/postgres/blob/12afc7145c03c212f26fea3a99e016da6a1c919c/src/include/fmgr.h#L110) for reference

src/backend/distributed/planner/query_pushdown_planning.c
```diff
@@ -194,6 +198,29 @@ JoinTreeContainsSubquery(Query *query)
...
+bool
+JoinTreeContainsWithoutFrom(Query *query)
+{
+	if (query->rtable == NIL)
+	{
+		return true;
+	}
+
+#if PG_VERSION_NUM >= 120000
+	else if (list_length(query->rtable) == 1)
```
onderkalaci: Hmm, I think I misguided you on this. It looks like for simple  `INSERT INTO tbl VALUES ();` type of queries would also end up with `RTE_RESULT ` appended next to RTE_RELATION.  I want to test/play with that to see if anything is broken on Citus with such a change. 

But as I mentioned I'm blocked on `CREATE EXTENSION citus` is failing on my local machine.

For reference, the query tree of an INSERT after RTE_RESULT changes:

```
   {QUERY 
   :commandType 3 
   :querySource 0 
   :canSetTag true 
   :utilityStmt <> 
   :resultRelation 1 
   :hasAggs false 
   :hasWindowFuncs false 
   :hasTargetSRFs false 
   :hasSubLinks false 
   :hasDistinctOn false 
   :hasRecursive false 
   :hasModifyingCTE false 
   :hasForUpdate false 
   :hasRowSecurity false 
   :cteList <> 
   :rtable (
      {RTE 
      :alias <> 
      :eref 
         {ALIAS 
         :aliasname users_table 
         :colnames ("user_id" "time" "value_1" "value_2" "value_3" "value_4")
         }
      :rtekind 0 
      :relid 16393 
      :relkind r 
      :rellockmode 3 
      :tablesample <> 
      :lateral false 
      :inh false 
      :inFromCl false 
      :requiredPerms 1 
      :checkAsUser 0 
      :selectedCols (b)
      :insertedCols (b 8)
      :updatedCols (b)
      :extraUpdatedCols (b)
      :securityQuals <>
      }
      {RTE 
      :alias <> 
      :eref 
         {ALIAS 
         :aliasname *RESULT* 
         :colnames <>
         }
      :rtekind 8 
      :lateral false 
      :inh false 
      :inFromCl false 
      :requiredPerms 0 
      :checkAsUser 0 
      :selectedCols (b)
      :insertedCols (b)
      :updatedCols (b)
      :extraUpdatedCols (b)
      :securityQuals <>
      }
   )
   :jointree 
      {FROMEXPR 
      :fromlist (
         {RANGETBLREF 
         :rtindex 2
         }
      )
      :quals <>
      }
   :targetList (
      {TARGETENTRY 
      :expr 
         {CONST 
         :consttype 23 
         :consttypmod -1 
         :constcollid 0 
         :constlen 4 
         :constbyval true 
         :constisnull false 
         :location 42 
         :constvalue 4 [ 8 0 0 0 0 0 0 0 ]
         }
      :resno 1 
      :resname user_id 
      :ressortgroupref 0 
      :resorigtbl 0 
      :resorigcol 0 
      :resjunk false
      }
   )
   :override 0 
   :onConflict <> 
   :returningList <> 
   :groupClause <> 
   :groupingSets <> 
   :havingQual <> 
   :windowClause <> 
   :distinctClause <> 
   :sortClause <> 
   :limitOffset <> 
   :limitCount <> 
   :rowMarks <> 
   :setOperations <> 
   :constraintDeps <> 
   :withCheckOptions <> 
   :stmt_location 0 
   :stmt_len 44
   }
```

src/backend/distributed/commands/vacuum.c
```diff
@@ -339,3 +343,74 @@ DeparseVacuumColumnNames(List *columnNameList)
...
+			freeze = defGetBoolean(opt);
+		}
+		else if (strcmp(opt->defname, "full") == 0)
+		{
+			full = defGetBoolean(opt);
+		}
+		else if (strcmp(opt->defname, "disable_page_skipping") == 0)
+		{
+			disable_page_skipping = defGetBoolean(opt);
+		}
```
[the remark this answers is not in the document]
pykello: > minor: maybe add
> else
> {
> /* postgres wouldn't allow unrecognized options to pass through */
> Assert (true);
> }

`Assert(false)`?

src/backend/distributed/utils/ruleutils_12.c
```diff
@@ -0,0 +1,7913 @@
+/*-------------------------------------------------------------------------
+ *
+ * ruleutils_11.c
```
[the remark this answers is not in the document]
serprex: I copied `ruleutils_11.c` to `ruleutils_12.c` & then compared `ruleutils.c` from REL_11_STABLE to `ruleutils.c` from REL_12_STABLE to guide updating `ruleutils_12.c`

src/backend/distributed/commands/vacuum.c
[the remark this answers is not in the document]
onderkalaci: > Did you mean this branch cannot be taken?

Ah, yes sure, `Assert(false)`. I actually noted this for readability, I'm not expecting this branch to be taken unless we forget any options or have a typo in the if/else statements etc.

src/backend/distributed/commands/vacuum.c
```diff
@@ -339,3 +343,74 @@ DeparseVacuumColumnNames(List *columnNameList)
...
+		else if (strcmp(opt->defname, "freeze") == 0)
+		{
+			freeze = defGetBoolean(opt);
+		}
+		else if (strcmp(opt->defname, "full") == 0)
+		{
+			full = defGetBoolean(opt);
+		}
+		else if (strcmp(opt->defname, "disable_page_skipping") == 0)
```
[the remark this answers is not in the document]
serprex: I've revised the vacuum commit to use a `CitusVacuumParams` which carries these options forward to being deparsed. `SKIP_LOCKED` is also new

src/backend/distributed/commands/vacuum.c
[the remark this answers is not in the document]
serprex: I went with keeping the error messages from Postgres

src/backend/distributed/commands/vacuum.c
```diff
@@ -339,3 +343,74 @@ DeparseVacuumColumnNames(List *columnNameList)
...
+static int
+VacuumStmt_options(VacuumStmt *vacstmt)
+{
+	bool verbose = false;
+	bool skip_locked = false;
+	bool analyze = false;
+	bool freeze = false;
+	bool full = false;
+	bool disable_page_skipping = false;
+	ListCell *lc;
```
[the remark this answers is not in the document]
serprex: I'd prefer to keep this function as close to the original pg function as possible, naming conventions included

src/backend/distributed/planner/query_pushdown_planning.c
serprex: Checking for explicit `length == 1` seems all the more correct then. It's more that we want the code to be for-all rather than for-any, which my earlier code was the latter. The length check will give the incorrect result if we end up with rtable having length greater than 1 & containing only `RTE_RESULT`

src/backend/distributed/planner/query_pushdown_planning.c
```diff
@@ -194,6 +198,29 @@ JoinTreeContainsSubquery(Query *query)
 }
 
 
+bool
+JoinTreeContainsWithoutFrom(Query *query)
```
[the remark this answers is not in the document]
serprex: `HasEmptyJoinTree`

src/backend/distributed/commands/vacuum.c
```diff
@@ -299,6 +320,29 @@ DeparseVacuumStmtPrefix(int vacuumFlags)
 		appendStringInfoString(vacuumPrefix, "VERBOSE,");
 	}
 
+#if PG_VERSION_NUM >= 120000
```
onderkalaci: This is crashing. I think we should either (a) add tests for the new features or (b) error out saying that we don't support it. Not only for vacuum, but also for any new thing that we allow.
 

```SQL
VACUUM (SKIP_LOCKED true )users_table;
Time: 1611.956 ms (00:01.612)
```

src/backend/distributed/commands/vacuum.c
```diff
@@ -299,6 +320,29 @@ DeparseVacuumStmtPrefix(int vacuumFlags)
...
+	if (vacuumFlags & VACOPT_SKIP_LOCKED)
+	{
+		appendStringInfoString(vacuumPrefix, "SKIP_LOCKED,");
+	}
+
+	if (vacuumParams.truncate != VACOPT_TERNARY_DEFAULT)
+	{
+		appendStringInfoString(vacuumPrefix,
+							   vacuumParams.truncate == VACOPT_TERNARY_ENABLED ?
```
onderkalaci: Minor: can't we make `0` -> `false`? Because these might show up in the logs, and `false` looks better.

src/backend/distributed/commands/vacuum.c
```diff
@@ -233,14 +250,13 @@ VacuumTaskList(Oid relationId, int vacuumOptions, List *vacuumColumnList)
...
 static StringInfo
-DeparseVacuumStmtPrefix(int vacuumFlags)
+DeparseVacuumStmtPrefix(CitusVacuumParams vacuumParams)
 {
+	int vacuumFlags = vacuumParams.options;
 	StringInfo vacuumPrefix = makeStringInfo();
 	const int unsupportedFlags PG_USED_FOR_ASSERTS_ONLY = ~(
+		VACOPT_VACUUM |
 		VACOPT_ANALYZE |
-		VACOPT_DISABLE_PAGE_SKIPPING |
```
onderkalaci: Why do we end-up with changing this with this commit? Is there anything changing related to `VACOPT_FULL ` or `VACOPT_FREEZE ` ?

src/backend/distributed/utils/ruleutils_12.c
```diff
@@ -0,0 +1,7913 @@
+/*-------------------------------------------------------------------------
```
onderkalaci: I'm done reviewing `ruleutils_12.c`, the way you created this file makes sense and I've double checked if anything missed. As I noted in the other issue, adding some tests with `MATERIALIZED` and `NOT MATERIALIZED` would be nice.

src/backend/distributed/commands/vacuum.c
serprex: Later in the code, after the assert, we have handling for those options. Meanwhile we don't handle those options before the assert

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -1413,9 +1423,19 @@ AdjustReadIntermediateResultCost(RangeTblEntry *rangeTableEntry, RelOptInfo *rel
...
 		/* add the cost of parsing a column */
-		rowCost += get_func_cost(inputFunctionId) * cpu_operator_cost;
+#if PG_VERSION_NUM >= 120000
+		add_function_cost(NULL, inputFunctionId, NULL, &funcCost);
+#else
+		funcCost += get_func_cost(inputFunctionId);
+#endif
 	}
+#if PG_VERSION_NUM >= 120000
+	rowCost += funcCost.per_tuple * cpu_operator_cost;
```
onderkalaci: Two things, `add_function_cost ()` seems to be already multiplying with `cpu_operator_cost`, so we should not do it again.
```add_function_cost()
...
	/* No support function, or it failed, so rely on procost */
	cost->per_tuple += procform->procost * cpu_operator_cost;
```
Also:
Should we need to add the following?
```C
rowCost += funcCost.startup
```

serprex: I assume `funcCost.startup` is a one-time cost, as opposed to `per_tuple` which would be the per-row portion

src/backend/distributed/commands/vacuum.c
serprex: Ideally we should be testing new features even if we error out on them. I'm considering adding a test that specifically goes through pg12 features, that way it'll be straight forward for having the pg11/pg10 results which spew errors

serprex: Looking closer, `ANALYZE` & `VERBOSE` are both applicable to `VACUUM` while also being handled later on in the code. So this assert is straight forward to hit with `VACUUM (ANALYZE)`

I've removed the assertion

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -1413,9 +1423,19 @@ AdjustReadIntermediateResultCost(RangeTblEntry *rangeTableEntry, RelOptInfo *rel
...
 		/* add the cost of parsing a column */
-		rowCost += get_func_cost(inputFunctionId) * cpu_operator_cost;
+#if PG_VERSION_NUM >= 120000
+		add_function_cost(NULL, inputFunctionId, NULL, &funcCost);
+#else
+		funcCost += get_func_cost(inputFunctionId);
+#endif
 	}
+#if PG_VERSION_NUM >= 120000
+	rowCost += funcCost.per_tuple;
```
onderkalaci: What I meant is shouldn't we need do add `startup` cost  as below:

```C
diff --git a/src/backend/distributed/planner/distributed_planner.c b/src/backend/distributed/planner/distributed_planner.c
index 312024b1c..c4feffcb5 100644
--- a/src/backend/distributed/planner/distributed_planner.c
+++ b/src/backend/distributed/planner/distributed_planner.c
@@ -1448,7 +1448,13 @@ AdjustReadIntermediateResultCost(RangeTblEntry *rangeTableEntry, RelOptInfo *rel
        /* tell the planner about the cost and row count of the function */
        path = (Path *) linitial(pathList);
        path->rows = rowCountEstimate;
+
+#if PG_VERSION_NUM >= 120000
+       path->total_cost = rowCountEstimate * rowCost + ioCost + funcCost.startup;
+#else
        path->total_cost = rowCountEstimate * rowCost + ioCost;
+#endif
+
 }
```

src/backend/distributed/commands/vacuum.c
onderkalaci: I don't think so, shouldn't we just need to add `VACOPT_SKIP_LOCKED ` to `unsupportedFlags `? We're already negating during the assignment? ```const int unsupportedFlags PG_USED_FOR_ASSERTS_ONLY = ~(...```

So, I'd rather add the assert back.

src/include/distributed/hash_helpers.h
```diff
@@ -11,6 +11,8 @@
 
 #include "utils/hsearch.h"
 
+#if PG_VERSION_NUM < 120000
+
```
onderkalaci: Why don't we need this function > 120000?

src/include/compat/nodes/relation.h
```diff
@@ -0,0 +1,5 @@
+#if PG_VERSION_NUM >= 120000
```
onderkalaci: Do we really need these `compat/...` header files. Can't we do as we used to, have these ifdefs in the code?

I can see that the number of such files could increase, and the cost of maintaining would increase. So, I'd go for just inlining these changes in the places where they are used.

src/backend/distributed/worker/worker_merge_protocol.c
```diff
@@ -17,7 +17,17 @@
 #include "funcapi.h"
 #include "miscadmin.h"
 
+#if PG_VERSION_NUM >= 120000
+#include "access/genam.h"
+#endif
 #include "access/htup_details.h"
+#if PG_VERSION_NUM >= 120000
+#include "access/table.h"
+#define heap_beginscan_catalog table_beginscan_catalog
```
onderkalaci: We should be consistent about these changes. For some other function in other files, we inline such changes in the code and here we're doing it here. We should try to do all the same, and I actually prefer the following, it makes it easy to navigate on eclipse.

```C
#if PG_VERSION_NUM >= 120000
	table_endscan(scan);
#else
	heap_endscan(scan);
#endif
```

src/include/distributed/version_compat.h
```diff
@@ -240,5 +244,132 @@ RangeVarGetRelidInternal(const RangeVar *relation, LOCKMODE lockmode, uint32 fla
...
+
+static inline int
+FileReadCompat(FileCompat *file, char *buffer, int amount, uint32 wait_event_info)
+{
+	return FileRead(file->fd, buffer, amount, wait_event_info);
+}
+
+
+static inline FileCompat
+FileCompatFromFileStart(File fileDesc)
```
onderkalaci: If you can move everything related to FileCompat into the same commit (e.g., remove it from this commit and move it to FileComant commit), it'd be easier for us to review.

src/include/distributed/hash_helpers.h
serprex: Postgres now exports it. Will add comment

src/backend/distributed/commands/vacuum.c
serprex: No, because the idea behind `unsupportedFlags` is that any of the flags before then will be removed or we'll've raised an error. It turns out the only flag that shouldn't remain by time we get to the assert is VACOPT_VACUUM

src/backend/distributed/planner/distributed_planner.c
serprex: I've also included `relOptInfo->baserestrictcost.startup`

src/include/distributed/version_compat.h
serprex: So think this should be done. Brought back a change that'd been deferred until collations commit & split up `version_compat.h` to be updated bit by bit. This does introduce a kind of linear dependency on the commits, but I'm fine with that, since in the end none of these commits should really be approved with out each other. Just makes it preferable if we don't mix up their order by cherry picking them in order of approval to `pg12-master`

pykello:
Today I looked at 

1. FileCompat, ok with merging it, maybe style needs some clean-up.

2. collations. I made a suggestion in an internal email to @serprex about `make_aggCompat` to make ignoring collations in pre-12 less confusing. I am ok with rest of the changes.

src/backend/distributed/planner/query_pushdown_planning.c
```diff
@@ -194,6 +200,29 @@ JoinTreeContainsSubquery(Query *query)
 }
 
 
+bool
```
onderkalaci: Missing function comment. Also, There is no regression test that triggers `if (rte->rtekind == RTE_RESULT)`, Can you find/add one?

src/test/regress/bin/normalized_tests.lst
```diff
@@ -1,40 +1,42 @@
...
-multi_name_lengths
-multi_partition_pruning
-multi_subtransactions
-multi_modifying_xacts
-multi_insert_select
-sql_procedure
-multi_reference_table
-multi_create_table_constraints
-
-# the following tests' output are
```
onderkalaci: These were splitted in a way that the following were only added here because of EXPLAIN output differences of the executor type. (`Adaptive` vs `Real-time`  or `Adaptive` vs 'Router'). It was done because it'd simplify things when we remove the support for real-time/router queries. 

Is there a reason you changed the order?

src/test/regress/expected/multi_partitioning_2.out
```diff
@@ -0,0 +1,1941 @@
+--
```
onderkalaci: I think we should not need this file. The diff between this and `multi_partitioning_1`.out is very, very small. I can see that few locks has changed which might be related to #2855 and its follow-ups.

I'm OK to remove this for now, and look the differences in detail later with a separate PR after we fix the lock issues.

src/test/regress/expected/multi_subquery_complex_reference_clause.out
```diff
@@ -349,7 +349,7 @@ ON user_buy_test_table.item_id = users_ref_test_table.id;
      4
 (1 row)
 
--- subquery without FROM cannot be the outer relationship in an outer join
+-- subquery without FROM cannot be the outer relationship in an outer join before pg12
```
onderkalaci: Adding an extra test output for just one test seems aggressive to me. The issue is that we should change  `SELECT 5 AS id` such that it is also qualified as `HasEmptyJoinTree()`. If you can do that, this test should not change between the versions.

src/test/regress/expected/multi_metadata_sync_2.out
```diff
@@ -0,0 +1,1451 @@
+--
+-- MULTI_METADATA_SYNC
```
onderkalaci: Again the diff is pretty minor between this and `multi_metadata_sync_2.out`. 

To avoid adding this file: (a) Add ORDER BY to `SELECT unnest(master_metadata_snapshot()); ORDER BY 1`

(b) Add a normalization rule to get rid of ` CREATE INDEX mx_index ON public.mx_test_table USING btree (col_2) TABLESPACE pg_default`  the `TABLESPACE pg_default` part (which I guess you already do).

So, do we really need this?

src/test/regress/expected/multi_master_protocol_2.out
```diff
@@ -0,0 +1,34 @@
+--
+-- MULTI_MASTER_PROTOCOL
```
onderkalaci: Again, the diff between this and `multi_master_protocol.out` is extremely small:
```
18d17
<  CREATE INDEX lineitem_time_index ON public.lineitem USING btree (l_shipdate) TABLESPACE pg_default
19a19
>  CREATE INDEX lineitem_time_index ON public.lineitem USING btree (l_shipdate)
```

Why wouldn't the normalization rule fixed it?

src/test/regress/expected/multi_create_shards_2.out
```diff
@@ -0,0 +1,238 @@
+SET citus.next_shard_id TO 370000;
```
onderkalaci: Again, the diff is only with OIDS,
```
43a44,46
> ERROR:  syntax error at or near "WITH"
> LINE 1: ALTER TABLE table_to_distribute SET WITH OIDS;
>                                             ^
45,46c48,49
< ERROR:  cannot distribute relation: table_to_distribute
< DETAIL:  Distributed relations must not specify the WITH (OIDS) option in their definitions.
---
> ERROR:  cannot create constraint on "table_to_distribute"
> DETAIL:  Distributed relations cannot have UNIQUE, EXCLUDE, or PRIMARY KEY constraints that do not include the partition column (with an equality operator if EXCLUDE).

```

Just removing the test that errors out is much better compared to adding a new file. The test already does a trivial check, so we can live without that test.

src/test/regress/expected/multi_reference_table_2.out
```diff
@@ -0,0 +1,1606 @@
+SET citus.next_shard_id TO 1250000;
```
onderkalaci: Again, the diff is trivial and just removing the test with OIDS should be good enough:
```
< ERROR:  alter table command is currently unsupported
< DETAIL:  Only ADD|DROP COLUMN, SET|DROP NOT NULL, SET|DROP DEFAULT, ADD|DROP|VALIDATE CONSTRAINT, SET (), RESET (), ATTACH|DETACH PARTITION and TYPE subcommands are supported.
---
> ERROR:  syntax error at or near "WITH"
> LINE 1: ...ER TABLE reference_schema.reference_table_ddl SET WITH OIDS;
>        
```

Citus doesn't support WITH OIDS anyway, so I don't care the test much.

src/test/regress/expected/multi_utility_statements_2.out
```diff
@@ -0,0 +1,363 @@
+--
```
onderkalaci: I cannot even understand the necessity for adding this test, the diff is suspicious:

```
diff expected/multi_utility_statements.out expected/multi_utility_statements_2.out 
195a196
> ERROR:  tables declared WITH OIDS are not supported
198,206c199
<  customer_key |       c_name       |                c_address                
< --------------+--------------------+-----------------------------------------
<             3 | Customer#000000003 | MG9kdTD2WBHm
<            14 | Customer#000000014 | KXkletMlL2JQEA 
<            30 | Customer#000000030 | nJDsELGAavU63Jl0c5NKsKfL8rIJQQkQnYL2QJY
<            59 | Customer#000000059 | zLOCP0wh92OtBihgspOGl4
<           106 | Customer#000000106 | xGCOEAUjUNG
< (5 rows)
< 
---
> ERROR:  current transaction is aborted, commands ignored until end of transaction block
```

So, how could pg12 error out with ` tables declared WITH OIDS are not supported`? 

Could this be an artifact of some intermediate works you've done?

src/test/regress/expected/multi_subquery_in_where_reference_clause.out
```diff
@@ -87,7 +87,7 @@ DETAIL:  Functions are not allowed in FROM clause when the query has subqueries
 SELECT
   user_id
 FROM
-  (SELECT  5 AS user_id) users_reference_table
+  (SELECT  5 AS user_id UNION ALL SELECT 6) users_reference_table
```
onderkalaci: Same here, if you can find a query where `HasEmptyJoinTree` returns true instead of `SELECT  5 AS user_id`, you wouldn't need to add a UNION.

src/test/regress/expected/pg12_0.out
```diff
@@ -0,0 +1,126 @@
+SHOW server_version \gset
```
onderkalaci: Well, the diff between pg12_0 and pg12_1 should be able to become empty by some normalization.
```
3c3
<  server_version_above_eleven 
---
>  server_version_above_eleven
16c16
< ERROR:  could not find function "blackhole_am_handler" in file "/home/pdube/pg11/lib/postgresql/citus.so"
---
> ERROR:  could not find function "blackhole_am_handler" in file "/home/pdube/pg10/lib/postgresql/citus.so"
85c85
< ERROR:  syntax error at or near "skip_locked"
---
> ERROR:  syntax error at or near "("
87c87
<                  ^
---
>                 ^
```
Besides, you have you local path in the error message:

> ERROR:  could not find function "blackhole_am_handler" in file "/home/pdube/pg10/lib/postgresql/citus.so"

So, we should hide this error message for sure

src/test/regress/bin/normalize.sed
```diff
@@ -73,6 +73,9 @@ s/(job_[0-9]+\/task_[0-9]+\/p_[0-9]+\.)[0-9]+/\1xxxx/g
 /^LINE [0-9]+:.*$/d
 /^\s*\^\s*$/d
 
+# Ignore file paths
+s,"/(home|usr|lib)/(.+/)*([^/]*\.so)","\3",
```
onderkalaci: I don't think this is generic enough. We should hide the error message all-together. You can use something like this function: https://github.com/citusdata/citus/blob/master/src/test/regress/sql/multi_test_helpers.sql#L112

src/test/regress/bin/normalize.sed
```diff
@@ -65,3 +69,19 @@ s/ERROR:  failed to execute task [0-9]+/ERROR:  failed to execute task X/g
...
+/^LINE [0-9]+:.*$/d
+/^\s*\^\s*$/d
+
+# Ignore file paths
+s,"/(home|usr|lib)/(.+/)*([^/]*\.so)","\3",
+
+# pg12 changes
+s/Partitioned table "/Table "/g
+s/\) TABLESPACE pg_default$/\)/g
+s/invalid input syntax for type /invalid input syntax for /g
```
onderkalaci: There seems to be one more occurrence of `type` diffs on `worker_check_invalid_arguments.out`, so we could add that.

```
-ERROR:  invalid input syntax for integer: "\x0b50"
+ERROR:  invalid input syntax for type integer: "\x0b50"
```

src/backend/distributed/planner/query_pushdown_planning.c
onderkalaci: Ping on this

src/test/regress/bin/normalized_tests.lst
onderkalaci: Ping on this as well.

src/test/regress/expected/multi_subquery_complex_reference_clause.out
onderkalaci: I noted on the e-mail as well. It turns out that the test outputs are the same between pg 11 and pg 12. Can you double check please? If so, lets remove _2.out

src/test/regress/expected/pg12_0.out
onderkalaci: Ping on this as well.

src/test/regress/expected/pg12.out
```diff
@@ -0,0 +1,90 @@
...
+	val2 int GENERATED ALWAYS AS (val1 + 2) STORED
+);
+create table gen2 (
+	id int,
+	val1 int,
+	val2 int GENERATED ALWAYS AS (val1 + 2) STORED
+);
+insert into gen1 (id, val1) values (1,4),(3,6),(5,2),(7,2);
+insert into gen2 (id, val1) values (1,4),(3,6),(5,2),(7,2);
+select * from create_distributed_table('gen1', 'id');
```
onderkalaci: I think if `create_distributed_table ` errors out, it'd be a nicer user experience. As we do `ERROR:  only heap AM is supported`

src/test/regress/expected/pg12.out
```diff
@@ -0,0 +1,90 @@
...
+vacuum (index_cleanup 1) gen1;
+-- COPY FROM
+create table cptest (id int, val int);
+select create_distributed_table('cptest', 'id');
+ create_distributed_table 
+--------------------------
+ 
+(1 row)
+
+copy cptest from STDIN with csv where val < 4;
```
onderkalaci: So, the result is wrong. Can't we error out in this PR when COPY has `WHERE` clause?

serprex: I'll bring #2880 into this PR now that we've decided not to implement support

src/test/regress/bin/normalized_tests.lst
serprex: There were duplicates & adding to a list is easier when it's alphabetical. I agree with @pykello that we should remove `normalized_tests.lst` & have a branch that does that, but it needs to figure out how to exclude vanillatest from normalization

src/backend/distributed/planner/shard_pruning.c
```diff
@@ -138,6 +139,17 @@ typedef struct PendingPruningInstance
 	Node *continueAt;
 } PendingPruningInstance;
 
+#if PG_VERSION_NUM >= 120000
+typedef union \
+{ \
+	FunctionCallInfoBaseData fcinfo; \
+	/* ensure enough space for nargs args is available */ \
+	char fcinfo_data[SizeForFunctionCallInfo(2)]; \
+} FunctionCall2InfoData;
```
pykello: This looks fine to me. Please add the flag to disable the warning in clang.

src/backend/distributed/planner/query_pushdown_planning.c
serprex: No tests in `check-multi` run into this. I'm wondering if postgres has optimized out `RTE_RESULT` by time it gets to us

src/backend/distributed/executor/multi_task_tracker_executor.c
```diff
@@ -447,7 +447,7 @@ MultiTaskTrackerExecute(Job *job)
 	}
 	else if (clusterFailed)
 	{
-		ereport(ERROR, (errmsg("failed to execute task %u", failedTaskId)));
+		ereport(ERROR, (errmsg("cluster failed")));
```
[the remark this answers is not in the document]
serprex: It isn't related to pg12, was an undefined variable usage I came across while debugging file compat changes. Will separate

src/test/regress/expected/pg12.out
```diff
@@ -0,0 +1,366 @@
...
+SELECT substring(:'server_version', '\d+')::int > 11 AS server_version_above_eleven
+\gset
+\if :server_version_above_eleven
+\else
+\q
+\endif
+set citus.shard_replication_factor to 1;
+create schema test_pg12;
+set search_path to test_pg12;
+CREATE FUNCTION blackhole_am_handler(internal)
```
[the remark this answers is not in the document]
serprex: It's important to test that we error on custom access handlers still. `blackhole_am` is the simplest access handler available

src/backend/distributed/commands/create_distributed_table.c
```diff
@@ -1355,3 +1395,48 @@ RelationUsesIdentityColumns(TupleDesc relationDesc)
...
+}
+
+
+/*
+ * Returns whether given relation uses default access method
+ */
+static bool
+RelationUsesHeapAccessMethod(Oid relationId)
+{
+#if PG_VERSION_NUM >= 120000
```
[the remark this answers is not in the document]
serprex: Foreign tables also have `InvalidOid` for amhandler. I think it's best to let `InvalidOid` & leave it to the rest of the function to reject the relation is necessary

src/test/regress/multi_schedule
```diff
@@ -277,3 +277,7 @@ test: multi_task_string_size
 # ---------
 test: ssl_by_default
 
+# ---------
+# test features introduced in postgres 12
+# ---------
+test: pg12
```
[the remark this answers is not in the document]
serprex: I'd like to look into addressing test times as a separate task, using the output of pg12's regression tests now including timing info to prioritize which tests to look into. I think there's much more time to be saved than the second this test takes up on my machine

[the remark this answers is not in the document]
onderkalaci: Still, we could prevent this test to have use more time now :)

src/backend/distributed/commands/create_distributed_table.c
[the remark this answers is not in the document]
pykello: I think a function named `RelationUsesHeapAccessMethod()` returning true for relations with no storage can be confusing. I'd suggest the following which covers both of your concerns:

```c
diff --git a/src/backend/distributed/commands/create_distributed_table.c b/src/backend/distributed/commands/create_distributed_table.c
index 87b3034d..151a404b 100644
--- a/src/backend/distributed/commands/create_distributed_table.c
+++ b/src/backend/distributed/commands/create_distributed_table.c
@@ -104,7 +104,7 @@ static void CopyLocalDataIntoShards(Oid relationId);
 static List * TupleDescColumnNameList(TupleDesc tupleDescriptor);
 static bool RelationUsesIdentityColumns(TupleDesc relationDesc);
 static bool RelationUsesGeneratedStoredColumns(TupleDesc relationDesc);
-static bool RelationUsesHeapAccessMethod(Relation relation);
+static void EnsureSupportedAccessMethod(Relation relation);
 static bool CanUseExclusiveConnections(Oid relationId, bool localTableEmpty);
 
 /* exports for SQL callable functions */
@@ -651,11 +651,7 @@ EnsureRelationCanBeDistributed(Oid relationId, Var *distributionColumn,
 	relationDesc = RelationGetDescr(relation);
 	relationName = RelationGetRelationName(relation);
 
-	if (!RelationUsesHeapAccessMethod(relation))
-	{
-		ereport(ERROR, (errmsg(
-							"cannot distribute relations using non-heap access methods")));
-	}
+	EnsureSupportedAccessMethod(relation);
 
 #if PG_VERSION_NUM < 120000
 
@@ -1423,17 +1419,21 @@ RelationUsesGeneratedStoredColumns(TupleDesc relationDesc)
 
 
 /*
- * Returns whether given relation uses default access method
+ * Errors out if relation has an access method that we don't support.
+ * Table access methods were introduced in postgres 12, so this is noop
+ * for version before that.
  */
-static bool
-RelationUsesHeapAccessMethod(Relation relation)
+static void
+EnsureSupportedAccessMethod(Relation relation)
 {
 #if PG_VERSION_NUM >= 120000
 
 	/* Foreign & partitioned tables lack access methods. Let them by. */
-	return relation->rd_amhandler == HEAP_TABLE_AM_HANDLER_OID ||
-		   relation->rd_amhandler == InvalidOid;
-#else
-	return true;
+	if (relation->rd_rel->relkind == RELKIND_RELATION &&
+		relation->rd_amhandler == HEAP_TABLE_AM_HANDLER_OID)
+	{
+		ereport(ERROR, (errmsg("cannot distribute relations using non-heap"
+							   " access methods")));
+	}
 #endif
 }
```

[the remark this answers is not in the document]
onderkalaci: Btw, the above suggestion makes sense to, if you agree please make the change @serprex.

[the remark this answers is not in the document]
serprex: I renamed `RelationUsesHeapAccessMethod` to `RelationUseHeapAccessMethodOrNone`

PR #2845, opened 2019-07-15
Squash migrations for versions 5/6, don't use WITH OIDS

### LINKED ISSUE #2834: Collapse SQL scripts to previous major release

serprex:
DESCRIPTION: Squash migration scripts prior to version 7

`WITH OIDS` is a syntax error in postgres 12, so blocks #2844 (in that once this is merged I'll remove the commit that does this from that branch)

Closes #2834


src/backend/distributed/citus--7.0-1.sql
```diff
@@ -0,0 +1,1735 @@
...
+\echo Use "CREATE EXTENSION citus" to load this file. \quit
+
+CREATE SCHEMA citus;
+
+-- Ensure CREATE EXTENSION is not run against an old citus data
+-- directory, we're not compatible (due to the builtin functions/tables)
+DO $$
+BEGIN
+   IF EXISTS(SELECT * FROM pg_proc WHERE proname = 'worker_apply_shard_ddl_command') THEN
+      RAISE 'cannot install citus extension in Citus 4 data directory';
```
pykello: Update the check and the error message.

serprex: Removed altogether since postgres versions shouldn't match up anyways

src/backend/distributed/citus--7.0-1.sql
```diff
@@ -0,0 +1,1431 @@
...
+    IS 'convert the textual Var representation to a column name';
+
+RESET search_path;
+/* citus--6.1-4--6.1-5.sql */
+
+SET search_path = 'pg_catalog';
+
+DROP FUNCTION create_distributed_table(regclass, text, citus.distribution_type);
+
+CREATE FUNCTION create_distributed_table(table_name regclass,
```
marcocitus: Defined twice

src/backend/distributed/citus--7.0-1.sql
```diff
@@ -0,0 +1,1431 @@
...
+    LANGUAGE C STRICT
+    AS 'MODULE_PATHNAME', $$worker_fetch_regular_table$$;
+COMMENT ON FUNCTION pg_catalog.worker_fetch_regular_table(text, text, bigint, text[], integer[])
+    IS 'fetch PostgreSQL table from remote node';
+CREATE OR REPLACE FUNCTION pg_catalog.worker_apply_shard_ddl_command(bigint, text)
+    RETURNS void
+    LANGUAGE sql
+AS $worker_apply_shard_ddl_command$
+    SELECT pg_catalog.worker_apply_shard_ddl_command($1, 'public', $2);
+$worker_apply_shard_ddl_command$;
```
marcocitus: These seem like some wrappers were introduced for backwards compatibility when schema support was introduced (for when the workers are ahead of the coordinator). Those wrappers can be removed.

src/backend/distributed/citus--7.0-1.sql
```diff
@@ -0,0 +1,1431 @@
...
+    IS 'get total disk space used by the specified table';
+
+RESET search_path;
+/* citus--6.2-2--6.2-3.sql */
+
+SET search_path = 'pg_catalog';
+
+DROP FUNCTION IF EXISTS master_add_node(text, integer);
+
+CREATE FUNCTION master_add_node(nodename text,
```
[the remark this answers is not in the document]
pykello: He is going to clean this file up and merge all `DROP/CREATE' or `ALTER … ADD/DROPs` and reorganize code and remove unnecessary comments.

src/backend/distributed/citus--7.0-1.sql
serprex: I've removed the previous 2-argument variant of this function, but the wrapper should be removed in a separate migration in 8.3. There should be no difference in schema between upgrading to `7.0-1` or installing to `7.0-1` through this script

PR #2848, opened 2019-07-22
Adaptive executor performance improvements

marcocitus:
- Avoid a call to FreeWaitEventSet by not creating an empty set
- Avoid calls to PQconsumeInput by checking the wait events
- Revert the change that removed use of `ModifyWaitEvent``
- Remove use of NodeConnectionCount (it's expensive)
- Remove calls to FindWorkerNode
- Skip timeout checks when checkForPoolTimeout is false
- Replace elog with errmsg

### DIFF
--- a/src/backend/distributed/connection/connection_management.c
+++ b/src/backend/distributed/connection/connection_management.c
@@ -333,7 +333,6 @@ StartNodeUserDatabaseConnection(uint32 flags, const char *hostname, int32 port, 
 	connection = StartConnectionEstablishment(&key);
 
 	dlist_push_tail(entry->connections, &connection->connectionNode);
-	entry->connectionCount++;
 
 	ResetShardPlacementAssociation(connection);
 
@@ -1065,8 +1064,6 @@ AfterXactHostConnectionHandling(ConnectionHashEntry *entry, bool isCommit)
 			cachedConnectionCount++;
 		}
 	}
-
-	entry->connectionCount = cachedConnectionCount;
 }
 
 
@@ -1171,30 +1168,4 @@ TrimLogLevel(const char *message)
 	} while (n < strlen(chompedMessage) && chompedMessage[n] == ' ');
 
 	return chompedMessage + n;
-}
-
-
-/*
- * NodeConnectionCount gets the number of connections to the given node
- * for the current username and database.
- */
-int
-NodeConnectionCount(char *hostname, int port)
-{
-	ConnectionHashKey key;
-	ConnectionHashEntry *entry = NULL;
-	bool found = false;
-
-	strlcpy(key.hostname, hostname, MAX_NODE_LENGTH);
-	key.port = port;
-	strlcpy(key.user, CurrentUserName(), NAMEDATALEN);
-	strlcpy(key.database, CurrentDatabaseName(), NAMEDATALEN);
-
-	entry = (ConnectionHashEntry *) hash_search(ConnectionHash, &key, HASH_FIND, &found);
-	if (!found)
-	{
-		return 0;
-	}
-
-	return entry->connectionCount;
 }
--- a/src/backend/distributed/executor/adaptive_executor.c
+++ b/src/backend/distributed/executor/adaptive_executor.c
@@ -290,7 +290,8 @@ typedef struct WorkerPool
 	DistributedExecution *distributedExecution;
 
 	/* worker node on which we have a pool of sessions */
-	WorkerNode *node;
+	char *nodeName;
+	int nodePort;
 
 	/* all sessions on the worker that are part of the current execution */
 	List *sessionList;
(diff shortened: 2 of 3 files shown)


src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -2567,11 +2593,13 @@ CheckConnectionReady(WorkerSession *session)
 		waitFlags = waitFlags | WL_SOCKET_WRITEABLE;
 	}
 
-	/* if reading fails, there's not much we can do */
-	if (PQconsumeInput(connection->pgConn) == 0)
+	if ((session->latestWaitEvents & WL_SOCKET_READABLE) != 0)
```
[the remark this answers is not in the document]
onderkalaci: And, shouldn't we also need to do it for `WL_SOCKET_WRITEABLE`? [`PQconsumeInput()`](https://doxygen.postgresql.org/fe-exec_8c.html#a98c667c1d76df63c6639c5f4b3b42f0d) also tries to flush for non-blocking connections.

So, maybe change `latestWaitEvents ` to a boolean field like `consumeInput` and do not check specifically for `WL_SOCKET_READABLE `?

src/backend/distributed/connection/connection_management.c
```diff
@@ -1172,29 +1172,3 @@ TrimLogLevel(const char *message)
...
-	strlcpy(key.user, CurrentUserName(), NAMEDATALEN);
-	strlcpy(key.database, CurrentDatabaseName(), NAMEDATALEN);
-
-	entry = (ConnectionHashEntry *) hash_search(ConnectionHash, &key, HASH_FIND, &found);
-	if (!found)
-	{
-		return 0;
-	}
-
-	return entry->connectionCount;
```
onderkalaci: I guess `entry->connectionCount` becomes unused with this change, so maybe remove?

src/backend/distributed/executor/adaptive_executor.c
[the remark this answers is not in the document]
marcocitus: > And, shouldn't we also need to do it for WL_SOCKET_WRITEABLE? PQconsumeInput() also tries to flush for non-blocking connections.

We already do a PQflush flush above.

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -3444,6 +3475,47 @@ BuildWaitEventSet(List *sessionList)
...
+	foreach(sessionCell, sessionList)
+	{
+		WorkerSession *session = lfirst(sessionCell);
+		MultiConnection *connection = session->connection;
+		int socket = 0;
+
+		if (connection->pgConn == NULL)
+		{
+			/* connection died earlier in the transaction */
+			continue;
```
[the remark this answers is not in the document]
marcocitus: Hmz, true, I copy-pasted this, but this seems off.

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -3444,6 +3478,49 @@ BuildWaitEventSet(List *sessionList)
...
 
 
+/*
+ * UpdateWaitEventSetFlags modifies the given waitEventSet with the wait flags
+ * for connections in the sessionList.
+ */
+static void
+UpdateWaitEventSetFlags(WaitEventSet *waitEventSet, List *sessionList)
+{
+	int connectionIndex = -1;
```
onderkalaci: We get some errors `+ERROR:  cannot modify postmaster death event`. I think we should be starting from 1 ` additional 2 is for postmaster and latch` ?

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -2581,6 +2612,9 @@ CheckConnectionReady(WorkerSession *session)
 
 	UpdateConnectionWaitFlags(session, waitFlags);
 
+	/* don't consume input redundantly if we cycle back into CheckConnectionReady */
+	session->latestWaitEvents = 0;
```
onderkalaci: Very minor, feel free to ignore: The name might be misleading since we're reseting the value right after the first iteration. Maybe `latestUnconsumedWaitEvents`?

src/backend/distributed/executor/adaptive_executor.c
onderkalaci: > I think we should be starting from 1 additional 2 is for postmaster and latch ?

Oh, no. Those are added to the end of the set.  And, we're not adding the connections that we skip here in `BuildWaitEventSet()`. So, we should not be incrementing `connectionIndex ` when we skip the session in the below loop. Sorry for misguide with my previous comment.

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -386,6 +387,13 @@ typedef struct WorkerSession
 	 * distributed transaction related commands such as BEGIN/COMMIT etc.
 	 */
 	uint64 commandsSent;
+
+	/* index in the wait event set */
+	int waitEventSetIndex;
+
+	/* events reported by the latest call to WaitEventSetWait */
+	int latestUnconsumedWaitEvents;
+>>>>>>> 73ab31ab... Avoid unnecessary calls to PQconsumeInput
```
onderkalaci: git conflict

PR #2849, opened 2019-07-23
Update workerNodeList after sorting

hanefi:
`SortList()` does not mutate the list given in the parameters, and the original list should be replaced with the sorted one here.

I checked all other usages of the said `SortList()`.


serprex:
It's questionable whether this actually makes things more deterministic since you didn't end up having to update any tests

Still, I'm currently dealing with non deterministic test results being a pain with pg12, & the code should definitely be either your fixed line or no line at all

PR #2851, opened 2019-07-23
(WIP) Add missing locks.

pykello:
These are unveiled by PostgreSQL 12's `Assert(CheckRelationLockedByMe(...))` checks. It might be worthwhile we add similar assertions in Citus.

### DIFF
--- a/src/backend/distributed/commands/create_distributed_table.c
+++ b/src/backend/distributed/commands/create_distributed_table.c
@@ -60,6 +60,7 @@
 #include "parser/parse_node.h"
 #include "parser/parse_relation.h"
 #include "parser/parser.h"
+#include "storage/lmgr.h"
 #include "tcop/pquery.h"
 #include "tcop/tcopprot.h"
 #include "utils/builtins.h"
@@ -152,6 +153,12 @@ master_create_distributed_table(PG_FUNCTION_ARGS)
 	{
 		ereport(ERROR, (errmsg("could not create distributed table: "
 							   "relation does not exist")));
+	}
+
+	/* We need to lock partitions too */
+	if (PartitionedTable(relationId))
+	{
+		LockRelationPartitions(relationId, ExclusiveLock);
 	}
 
 	/*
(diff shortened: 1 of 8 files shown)


src/backend/distributed/utils/node_metadata.c
```diff
@@ -1331,9 +1331,15 @@ DeleteNodeRow(char *nodeName, int32 nodePort)
...
 	ScanKeyData scanKey[2];
-
 	Relation pgDistNode = heap_open(DistNodeRelationId(), RowExclusiveLock);
 
+	/*
+	 * simple_heap_delete() expects that the caller has at least an
+	 * AccessShareLock on replica identity index.
+	 */
+	Relation replicaIndex = index_open(RelationGetReplicaIndex(pgDistNode),
+									   AccessShareLock);
```
pykello: I talked with Andres about this, and he was ok with this approach.

Normally deletions go through postgres executor which acquires the necessary locks, but it seems that we cannot do that here because the user might not have enough permissions for executing DML on pg_dist_node so going through executor might error out.

src/backend/distributed/commands/create_distributed_table.c
```diff
@@ -1355,3 +1362,22 @@ RelationUsesIdentityColumns(TupleDesc relationDesc)
 
 	return false;
 }
+
+
+/* Locks all paritions of the given relation with the given lock mode. */
+void
+LockRelationPartitions(Oid relationId, LOCKMODE lockMode)
```
onderkalaci: I think we have a very similar function `LockPartitionRelations()`

src/backend/distributed/commands/create_distributed_table.c
```diff
@@ -154,6 +155,12 @@ master_create_distributed_table(PG_FUNCTION_ARGS)
 							   "relation does not exist")));
 	}
 
+	/* We need to lock partitions too */
+	if (PartitionedTable(relationId))
```
onderkalaci: Ah, `master_create_distributed_table ` again :/ With 9.0, we should remove all of these all APIs..

src/backend/distributed/commands/table.c
```diff
@@ -155,6 +155,14 @@ ProcessCreateTableStmtPartitionOf(CreateStmt *createStatement)
 			char *parentRelationName = generate_qualified_relation_name(parentRelationId);
 			bool viaDeprecatedAPI = false;
 
+
+			/*
+			 * relation_access_tracking assumes we acquire locks on all partitions
+			 * of the parent relation.
+			 */
+			LockRelationOid(parentRelationId, ExclusiveLock);
+			LockRelationPartitions(parentRelationId, ExclusiveLock);
```
onderkalaci: This change looks surprising to me. We're getting here after `standardProcess_Utility` executed, right? So, doesn't Postgres already acquire the necessary relation locks? If not, are the locks really necessary?

Also, in what sense `relation_access_tracking ` requires the locks? Is it the regression tests of it requires or the actual tracking?

PR #2852, opened 2019-07-24
Update two tests, useful fallout from pg12 branch

serprex:
multi_colocation_utils: sort by nodeport, not placementid
multi_copy: replace smgr with aclitem, smgr is removed in pg12


src/test/regress/expected/multi_colocation_utils.out
```diff
@@ -716,7 +716,7 @@ ORDER BY
     logicalrelid,
     shardmaxvalue::integer,
     shardid,
-    placementid;
+    nodeport;
```
onderkalaci: I don't have strong objections but just wanted to understand more. How could this change help for pg 12 compatibility? 

If the output of the test is not consistent across pg versions, I cannot see how this would fix it.

serprex: The projection doesn't use `placementid` so the ordering had ties

PR #2854, opened 2019-07-25
CompareShardIntervals: if intervals are equal, compare id

serprex:
Works around sort being unstable

In pg10/pg11, returning 0 when both shards had null intervals caused `multi_prune_shard_list` to print equal shardids in descending order (whereas the test has them in ascending). Meanwhile in pg12 the currently broken null comparison was returning in descending order & with return 0 began returning in ascending order. All this to say that when you're using an unstable sort & you have an id, it's probably a good idea to fallback on the id when the comparison thinks the elements are equal

`multi_task_assignment_policy` was also impacted by this, so including normalization

### DIFF
--- a/src/backend/distributed/utils/shardinterval_utils.c
+++ b/src/backend/distributed/utils/shardinterval_utils.c
@@ -64,39 +64,40 @@ CompareShardIntervals(const void *leftElement, const void *rightElement,
 {
 	ShardInterval *leftShardInterval = *((ShardInterval **) leftElement);
 	ShardInterval *rightShardInterval = *((ShardInterval **) rightElement);
-	Datum leftDatum = 0;
-	Datum rightDatum = 0;
-	Datum comparisonDatum = 0;
 	int comparisonResult = 0;
+	bool leftHasNull = (!leftShardInterval->minValueExists ||
+						!leftShardInterval->maxValueExists);
+	bool rightHasNull = (!rightShardInterval->minValueExists ||
+						 !rightShardInterval->maxValueExists);
 
 	Assert(typeCompareFunction != NULL);
 
-	/*
-	 * Left element should be treated as the greater element in case it doesn't
-	 * have min or max values.
-	 */
-	if (!leftShardInterval->minValueExists || !leftShardInterval->maxValueExists)
+	if (leftHasNull && rightHasNull)
+	{
+		comparisonResult = 0;
+	}
+	else if (leftHasNull)
 	{
 		comparisonResult = 1;
-		return comparisonResult;
 	}
-
-	/*
-	 * Right element should be treated as the greater element in case it doesn't
-	 * have min or max values.
-	 */
-	if (!rightShardInterval->minValueExists || !rightShardInterval->maxValueExists)
+	else if (rightHasNull)
 	{
 		comparisonResult = -1;
-		return comparisonResult;
+	}
+	else
+	{
+		/* if both shard interval have min/max values, calculate comparison result */
+		Datum leftDatum = leftShardInterval->minValue;
+		Datum rightDatum = rightShardInterval->minValue;
+		Datum comparisonDatum = CompareCall2(typeCompareFunction, leftDatum, rightDatum);
+		comparisonResult = DatumGetInt32(comparisonDatum);
 	}
 
-	/* if both shard interval have min/max values, calculate the comparison result */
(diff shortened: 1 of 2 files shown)


src/backend/distributed/utils/shardinterval_utils.c
```diff
@@ -74,29 +74,31 @@ CompareShardIntervals(const void *leftElement, const void *rightElement,
...
-	if (!rightShardInterval->minValueExists || !rightShardInterval->maxValueExists)
+	if (!leftShardInterval->minValueExists || !leftShardInterval->maxValueExists ||
+		!rightShardInterval->minValueExists || !rightShardInterval->maxValueExists)
 	{
-		comparisonResult = -1;
-		return comparisonResult;
+		comparisonResult = (!leftShardInterval->minValueExists ||
+							!leftShardInterval->maxValueExists) -
+						   (!rightShardInterval->minValueExists ||
+							!rightShardInterval->maxValueExists);
```
pykello: Is this correct? This returns 0 in cases that two shard intervals aren't equal. For example, when left is `(min=NULL, max=100)` and right is `(min=101, max=NULL)`.

serprex: Old code would return 1 regardless of which of your examples are left/right. There might be an argument to only compare minValue & not care about maxValue being null. I'm trying to make the most conservative change necessary for my ends

src/backend/distributed/utils/shardinterval_utils.c
```diff
@@ -74,29 +74,31 @@ CompareShardIntervals(const void *leftElement, const void *rightElement,
...
-		comparisonResult = 1;
-		return comparisonResult;
-	}
-
-	/*
 	 * Right element should be treated as the greater element in case it doesn't
 	 * have min or max values.
 	 */
-	if (!rightShardInterval->minValueExists || !rightShardInterval->maxValueExists)
+	if (!leftShardInterval->minValueExists || !leftShardInterval->maxValueExists ||
```
onderkalaci: I find it slightly hard to follow the conditions and the calculation. What about something like the following for ease of read:
```C
bool leftHasNull = (!leftShardInterval->minValueExists || !leftShardInterval->maxValueExists);
bool rightHasNull = (!rightShardInterval->minValueExists || !rightShardInterval->maxValueExists);

if (leftHasNull && rightHasNull)
{
    /* rely on shardId for order to get consistent results */
	comparisonResult = 0;
}
else if (leftHasNull)
{
	comparisonResult = 1;
}
else if (rightHasNull)
{
	comparisonResult = -1;
}
else
{
	/* if both shard interval have min/max values, calculate the comparison result */
	...

}

...
```

src/backend/distributed/utils/shardinterval_utils.c
```diff
@@ -74,29 +74,31 @@ CompareShardIntervals(const void *leftElement, const void *rightElement,
 	/*
 	 * Left element should be treated as the greater element in case it doesn't
 	 * have min or max values.
-	 */
```
onderkalaci: I guess the function comment of `CompareShardIntervals ` should change slightly to reflect the changes. I'm OK to remove the second sentence and have something like `See the comments in the function for shard intervals with empty min/max values`

serprex: Given that we explain the desired behavior in the function's top level comment, I went with removing the inline comment altogether

onderkalaci: Not that important in this case, just sharing my thoughts on this in general. Typically function comments become obsolete as we change the functions and we forget to update the function comments. So, I prefer to keep more detailed comments in the function instead of the top.

src/backend/distributed/utils/shardinterval_utils.c
```diff
@@ -68,35 +68,40 @@ CompareShardIntervals(const void *leftElement, const void *rightElement,
...
-	if (!rightShardInterval->minValueExists || !rightShardInterval->maxValueExists)
+	else if (rightHasNull)
 	{
 		comparisonResult = -1;
-		return comparisonResult;
 	}
+	else
+	{
+		/* if both shard interval have min/max values, calculate the comparison result */
+		leftDatum = leftShardInterval->minValue;
```
onderkalaci: super minor: I think we can declare `leftDatum ` and `rightDatum ` inside the else block.

PR #2855, opened 2019-07-25
Add some missing locks.

pykello:
Pieces of https://github.com/citusdata/citus/pull/2851 which seem to be non-controversial. The plan is to resolve rest of the locking issues after we merge postgres 12 support.

### DIFF
--- a/src/backend/distributed/commands/truncate.c
+++ b/src/backend/distributed/commands/truncate.c
@@ -48,9 +48,32 @@ static void AcquireDistributedLockOnRelations(List *relationIdList, LOCKMODE loc
 void
 ProcessTruncateStatement(TruncateStmt *truncateStatement)
 {
+	ListCell *rangeVarCell = NULL;
+	ListCell *relationCell = NULL;
+	List *lockedRelations = NIL;
+
+	/*
+	 * Lock relations while calling next three functions, because they expect
+	 * relations to be locked. We release them then to avoid distributed
+	 * deadlock in MX.
+	 */
+	foreach(rangeVarCell, truncateStatement->relations)
+	{
+		RangeVar *rangeVar = (RangeVar *) lfirst(rangeVarCell);
+		Relation relation = heap_openrv(rangeVar, AccessShareLock);
+		lockedRelations = lappend(lockedRelations, relation);
+	}
+
 	ErrorIfUnsupportedTruncateStmt(truncateStatement);
 	EnsurePartitionTableNotReplicatedForTruncate(truncateStatement);
 	ExecuteTruncateStmtSequentialIfNecessary(truncateStatement);
+
+	foreach(relationCell, lockedRelations)
+	{
+		Relation relation = (Relation) lfirst(relationCell);
+		heap_close(relation, AccessShareLock);
+	}
+
 	LockTruncatedRelationMetadataInWorkers(truncateStatement);
 }
 
@@ -67,7 +90,7 @@ ErrorIfUnsupportedTruncateStmt(TruncateStmt *truncateStatement)
 	foreach(relationCell, relationList)
 	{
 		RangeVar *rangeVar = (RangeVar *) lfirst(relationCell);
-		Oid relationId = RangeVarGetRelid(rangeVar, NoLock, true);
+		Oid relationId = RangeVarGetRelid(rangeVar, NoLock, false);
 		char relationKind = get_rel_relkind(relationId);
 		if (IsDistributedTable(relationId) &&
 			relationKind == RELKIND_FOREIGN_TABLE)
@@ -93,19 +116,15 @@ EnsurePartitionTableNotReplicatedForTruncate(TruncateStmt *truncateStatement)
 
 	foreach(relationCell, truncateStatement->relations)
 	{
-		RangeVar *relationRV = (RangeVar *) lfirst(relationCell);
-		Relation relation = heap_openrv(relationRV, NoLock);
-		Oid relationId = RelationGetRelid(relation);
+		RangeVar *rangeVar = (RangeVar *) lfirst(relationCell);
+		Oid relationId = RangeVarGetRelid(rangeVar, NoLock, false);
 
 		if (!IsDistributedTable(relationId))
 		{
-			heap_close(relation, NoLock);
 			continue;
 		}
 
 		EnsurePartitionTableNotReplicated(relationId);
-
-		heap_close(relation, NoLock);
 	}
 }
 
@@ -181,22 +200,19 @@ LockTruncatedRelationMetadataInWorkers(TruncateStmt *truncateStatement)
 
 	foreach(relationCell, truncateStatement->relations)
 	{
-		RangeVar *relationRV = (RangeVar *) lfirst(relationCell);
-		Relation relation = heap_openrv(relationRV, NoLock);
-		Oid relationId = RelationGetRelid(relation);
+		RangeVar *rangeVar = (RangeVar *) lfirst(relationCell);
+		Oid relationId = RangeVarGetRelid(rangeVar, NoLock, false);
 		DistTableCacheEntry *cacheEntry = NULL;
 		List *referencingTableList = NIL;
 		ListCell *referencingTableCell = NULL;
 
 		if (!IsDistributedTable(relationId))
 		{
-			heap_close(relation, NoLock);
 			continue;
 		}
 
 		if (list_member_oid(distributedRelationList, relationId))
 		{
-			heap_close(relation, NoLock);
 			continue;
 		}
 
@@ -212,8 +228,6 @@ LockTruncatedRelationMetadataInWorkers(TruncateStmt *truncateStatement)
 			distributedRelationList = list_append_unique_oid(distributedRelationList,
 															 referencingRelationId);
 		}
-
-		heap_close(relation, NoLock);
 	}
 
 	if (distributedRelationList != NIL)
--- a/src/backend/distributed/transaction/relation_access_tracking.c
+++ b/src/backend/distributed/transaction/relation_access_tracking.c
@@ -222,45 +222,14 @@ PlacementAccessTypeToText(ShardPlacementAccessType accessType)
 static void
 RecordRelationAccessBase(Oid relationId, ShardPlacementAccessType accessType)
 {
+	/*
+	 * We call this only for reference tables, and we don't support partitioned
+	 * reference tables.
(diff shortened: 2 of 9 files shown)


src/backend/distributed/commands/create_distributed_table.c
```diff
@@ -154,6 +155,12 @@ master_create_distributed_table(PG_FUNCTION_ARGS)
 							   "relation does not exist")));
 	}
 
+	/* we need to lock partitions too */
```
onderkalaci: We don't support partitioned tables with `master_create_distributed_table()`, so do we really need this lock(s)?

```SQL
SELECT master_create_distributed_table('collections_list', 'key', 'hash');
ERROR:  distributing partitioned tables in only supported with create_distributed_table UDF                                                                                                                                 
```

src/backend/distributed/utils/node_metadata.c
```diff
@@ -1331,9 +1331,15 @@ DeleteNodeRow(char *nodeName, int32 nodePort)
 	HeapTuple heapTuple = NULL;
 	SysScanDesc heapScan = NULL;
 	ScanKeyData scanKey[2];
-
 	Relation pgDistNode = heap_open(DistNodeRelationId(), RowExclusiveLock);
 
+	/*
+	 * simple_heap_delete() expects that the caller has at least an
```
onderkalaci: What about the other usages of `simple_heap_delete()` on Citus code-base, do you think do we need the same pattern there as well?

As far as I remember, replica identity index is the primary key or the unique index (if explicitly specified) on the table, and all other tables that we do `simple_heap_delete` also operates on tables with pkey or unique index. So, to make the code more roboust, shall we do it for all? If we do, we should make sure to skip incase of `invalidOid`s?

 In that sense, maybe use `GetRelationIdentityOrPK()` function?


 (I guess the reason that only this crashes with pg 12 is that Pkeys are automatically set as the replica identity, but not unique indexes.)

marcocitus: There are places where Postgres complains that our catalog tables don't have a replica identity. Should we change the unique indexes to primary keys in SQL?

src/backend/distributed/commands/create_distributed_table.c
pykello: Removed this to have a closer in next release.

src/backend/distributed/utils/node_metadata.c
pykello: > There are places where Postgres complains that our catalog tables don't have a replica identity. Should we change the unique indexes to primary keys in SQL?

@marcocitus @onderkalaci  Agreed, but since we are tight on time for this release, what do you think about postponing this to 9.0?

I added the lock acquisition for two cases in which we are doing delete from tables with primary key.

src/backend/distributed/transaction/relation_access_tracking.c
```diff
@@ -227,10 +227,12 @@ RecordRelationAccessBase(Oid relationId, ShardPlacementAccessType accessType)
...
 	 * If a relation is partitioned, record accesses to all of its partitions as well.
-	 * We prefer to use PartitionedTableNoLock() because at this point the necessary
-	 * locks on the relation has already been acquired.
+	 *
+	 * TODO: We should use PartitionedTableNoLock() and PartitionTableNoLock() here
+	 * after making sure all of the calls that end up here acquire proper locks.
+	 * See https://github.com/citusdata/citus/issues/2888.
 	 */
-	if (PartitionedTableNoLock(relationId))
+	if (PartitionedTable(relationId))
```
pykello: @onderkalaci @marcocitus what do you think postponing a proper solution for this to 9.0? I'm suggesting this because we are tight on time for this release and these issues are issues we had in previous releases too, but were just unveiled by postgres 12's assertions, so no new issues.

If we get free time before the release we can in look into this more properly though.

src/backend/distributed/commands/truncate.c
```diff
@@ -67,7 +67,7 @@ ErrorIfUnsupportedTruncateStmt(TruncateStmt *truncateStatement)
 	foreach(relationCell, relationList)
 	{
 		RangeVar *rangeVar = (RangeVar *) lfirst(relationCell);
-		Oid relationId = RangeVarGetRelid(rangeVar, NoLock, true);
+		Oid relationId = RangeVarGetRelid(rangeVar, AccessShareLock, true);
```
pykello: We could also use `RangeVarGetRelid()`, I don't know if we want to really keep relations locked at this stage?

src/backend/distributed/transaction/relation_access_tracking.c
onderkalaci: Actually, as of #2866 we're only calling `RecordRelationAccessBase ()` for reference tables. So, they cannot be partitioned tables anyway -- for now. 

And, for reference tables the performance bottleneck is not usually related anything happening on the coordinator, plus we're only acquiring `AccessShareLock`.

I'm OK to keep as-is, or even remove the logic for partition/partitioned tables here.

src/backend/distributed/commands/truncate.c
```diff
@@ -48,6 +48,18 @@ static void AcquireDistributedLockOnRelations(List *relationIdList, LOCKMODE loc
 void
 ProcessTruncateStatement(TruncateStmt *truncateStatement)
 {
+	ListCell *relationCell = NULL;
+
+	/*
+	 * Lock relations to avoid them being dropped or their structure being
```
onderkalaci: Hmm, isn't `AccessShareLock ` is too weak for TRUNCATE? And, also the issue with this approach that we do not capture relations that are truncated via `TRUNCATE .. CASCADE` and partitioned relations. I guess that's one of main reasons that we use truncate triggers so that it cascades to related relations.

So, my question is that, do we really need to have these locks? I cannot see anything related to this in https://github.com/citusdata/citus/issues/2888 as well

src/backend/distributed/transaction/relation_access_tracking.c
```diff
@@ -523,10 +525,12 @@ RecordParallelRelationAccess(Oid relationId, ShardPlacementAccessType placementA
 
 	/*
 	 * If a relation is partitioned, record accesses to all of its partitions as well.
-	 * We prefer to use PartitionedTableNoLock() because at this point the necessary
-	 * locks on the relation has already been acquired.
+	 *
+	 * TODO: We should use PartitionedTableNoLock() and PartitionTableNoLock() here
```
onderkalaci: I think in parallel access case, we don't need to be careful about performance, so I'm even up for removing the comment here.

The main idea for `NoLockPartitionedTable` is that we could call it too many times (as it used to be) and this could be a performance issue. For parallel queries, I think that's not going to be a performance issue for sure because we're calling `RecordParallelRelationAccess ` mostly once per distributed query/table.

src/backend/distributed/utils/multi_partitioning_utils.c
```diff
@@ -40,9 +40,15 @@ static char * PartitionBound(Oid partitionId);
 bool
 PartitionedTable(Oid relationId)
 {
-	Relation rel = heap_open(relationId, AccessShareLock);
+	Relation rel = try_relation_open(relationId, AccessShareLock);
 	bool partitionedTable = false;
 
+	/* don't error out for tables that are dropped */
+	if (rel == NULL)
```
onderkalaci: I couldn't see the reason that we're doing this for. How could this change lead to getting a NULL relation here?

src/backend/distributed/transaction/relation_access_tracking.c
pykello: Removed this checks and added an assert.

src/backend/distributed/transaction/relation_access_tracking.c
pykello: Removed the comment.

src/backend/distributed/commands/truncate.c
pykello: Removed this lock. Changed all `heap_open()` calls to `RangeVarGetRelid(..., NoLock, ...)` which doesn't have the offending assertion. This keeps the behaviour as before for this release without failing assertions in pg12. We will look into this closer in next release.

src/backend/distributed/commands/truncate.c
```diff
@@ -93,19 +93,15 @@ EnsurePartitionTableNotReplicatedForTruncate(TruncateStmt *truncateStatement)
...
+		RangeVar *rangeVar = (RangeVar *) lfirst(relationCell);
+		Oid relationId = RangeVarGetRelid(rangeVar, NoLock, false);
 
 		if (!IsDistributedTable(relationId))
 		{
-			heap_close(relation, NoLock);
 			continue;
 		}
 
 		EnsurePartitionTableNotReplicated(relationId);
```
onderkalaci: Unfortunately this is not behaving exactly the same. This function acquires `AccessShareLock`, and then if we do `TRUNCATE` from the MX workers at the same time, we get into distributed deadlocks:

```SQL

cat /tmp/f.sql 
TRUNCATE users_table;

pgbench -h localhost -p 9700 -f /tmp/f.sql -T 10 -P 5 postgres &
pgbench -h localhost -p 9701 -f /tmp/f.sql -T 10 -P 5 postgres &
pgbench -h localhost -p 9702 -f /tmp/f.sql -T 10 -P 5 postgres &

lient 0 aborted in command 0 (SQL) of script 0; ERROR:  canceling the transaction since it was involved in a distributed deadlock

client 0 aborted in command 0 (SQL) of script 0; ERROR:  canceling the transaction since it was involved in a distributed deadlock

```

I guess, with heap_close, we used to release all the locks on the relation.

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -794,7 +794,7 @@ EnsurePartitionTableNotReplicated(Oid relationId)
 static DeferredErrorMessage *
 DeferErrorIfPartitionTableNotSingleReplicated(Oid relationId)
 {
-	if (PartitionTableNoLock(relationId) && !SingleReplicatedTable(relationId))
+	if (PartitionTable(relationId) && !SingleReplicatedTable(relationId))
```
onderkalaci: Maybe just switch back to NoLock to fix the issue?

pykello: With `NoLock` we will get assertion errors in PG12. I will look into how to avoid the deadlock today.

Issue #2823, opened 2019-07-05
One task failure causes whole query to fail in adaptive executor even when replication factor > 1


mtuncer:
Had a 64 shard table t3(a,b,c) with replication factor = 2 distributed on column a, total 100M rows

run 
```sql
select count(distinct a) from t3;
```

using default settings (adaptive executor). and killed a random worker process.
```
postgres=# select count(distinct a) from t3;
ERROR:  terminating connection due to administrator command
CONTEXT:  while executing command on localhost:9701
```

set `set citus.task_executor_type to 'real-time'` and run the same query again.
killed a random worker task, this time query completed.
```
postgres=# select count(distinct a) from t3b;
WARNING:  terminating connection due to administrator command
WARNING:  SSL connection has been closed unexpectedly

  count  
---------
 1000001
(1 row)
```

marcocitus:
This change in behaviour is by design.  We decided to give hard errors for failures that occur after a command was sent because a) for performance reasons we no longer store results in intermediate files which means we can no longer tolerate getting back the same rows from multiple placements due to a failover, and b) we can give more meaningful error messages to the client by throwing a hard error immediately.

Issue #2460, opened 2018-10-26
Query cancellation is partially broken for single shard modification with replication factor > 1


onderkalaci:
On query cancellation, even if the query errors out, the modifications are done on the placements for single shard modification with replication factor > 1.

I hit this while trying to add cancellation tests for single shard modification queries using mitmproxy. This bug probably have existed since from pg_shard because I cannot see anything that could have prevented this (unless we had `HOLD_INTERRUPTS()` during the execution).

Basically, the following happens:

- User executes the query on the coordinator
	- Query is executed on the first replica and since it is not inside a 2PC, the query is already committed in the first replica
	- Query is executed on the second replica and since it is not inside a 2PC, the query is already committed in the second replica
	- User cancels the query just after the query is sent to the second worker
	- The coordinator query rollbacks with an `ERROR:  canceling statement due to user request`
	- Since there is no 2PC, there is not way to cancel the queries on the worker.

Reproducing this interactively is pretty hard. But, we can either use mitmproxy or gdb to repro the bug.

```SQL

SET citus.shard_count = 2;
SET citus.shard_replication_factor = 2;
CREATE TABLE mod_test (key int, value text);
SELECT create_distributed_table('mod_test', 'key');

SELECT COUNT(*) FROM mod_test WHERE key=2;
 count 
-------
     0
(1 row)

-- cancel right after the query is sent to the 2nd worker
SELECT citus.mitmproxy('conn.onQuery(query="^INSERT").cancel(' ||  pg_backend_pid() || ')');
 mitmproxy 
-----------
 
(1 row)

SET client_min_messages TO DEBUG2;
SET citus.log_remote_commands TO ON;
INSERT INTO mod_test VALUES (2, 6);
DEBUG:  Creating router plan
DEBUG:  Plan is router executable
DETAIL:  distribution column value: 2
LOG:  issuing INSERT INTO public.mod_test_102270 (key, value) VALUES (2, 6)
DETAIL:  on server localhost:57637
LOG:  issuing INSERT INTO public.mod_test_102270 (key, value) VALUES (2, 6)
DETAIL:  on server localhost:9060
ERROR:  canceling statement due to user request


SELECT COUNT(*) FROM mod_test WHERE key=2;
DEBUG:  Creating router plan
DEBUG:  Plan is router executable
DETAIL:  distribution column value: 2
[...]

Issue #1032, opened 2016-12-09
Consider running all replicated modifications in a transaction block


marcocitus:
When performing individual DML commands, the router executor currently modifies and commits replicas one by one, rather than opening a transaction block on each replica and committing them after modification succeeds on all of them. While this has latency benefits in a few situations, it is not a sustainable model, complicates the code, and lowers reliability and uptime.

Currently, if we detect an application-level failure after the first replica has committed, there is no way to roll back and thus we must mark the replica that exhibited the failure as inactive, even if the same failure occurs on all remaining replicas. The subsequent cost of repairing the inactive placements is high, since writes are blocked while the placements are copied.

Consider a scenario in which the user manually adds a trigger with application-level error checks to shard 1 with placements A B and C. For some reason, placement A became inactive, the user then runs a shard repair operation for placement A, but neglects to also re-create the trigger. Then, a DML command is issued and succeeds on placement A, but the trigger fails on placements B and C. Now placements B and C, which are the "good" placements with the appropriate trigger, are marked as inactive, and only placement A, which is the "bad" placement without the trigger remains. In the end, the trigger logic is not honoured and an expensive repair operation is necessary.

Another scenario is when a network blip happens after a change was committed on a single replica. If connections to the other placements fail, they will have to be marked as inactive, otherwise they would be out-of-sync. In the end, an expensive repair operation is necessary.

Many scenarios like the ones described above exist and they would be prevented by rolling back the transaction on placement A after the failure on B by performing them in a transaction block that is committed in a 1PC or 2PC. Doing this would allow us to be much more conservative about changing shardstate, which can avoid unnecessary repairs or failures. Inactive shards and problems caused by them is a problem many Citus users have encountered.

The drawback of using a 1PC/2PC would be an additional round-trip for sending a commit to all the nodes after the write succeeded, but the commit step can be performed in parallel. In addition, this technique also allows us to perform update/delete/upsert and non-conflicting insert in parallel across the replicas, which actually lowers latency.

marcocitus:
This has been implemented

PR #2858, opened 2019-07-30
Use 2PC in adaptive executor when dealing with replication factors above 1

serprex:
DESCRIPTION: Use 2PC in adaptive executor when dealing with replication factors above 1

While this passes `check-multi`, with `check-base EXTRA_TESTS="multi_modifications"` we end up in a state where both shards are healthy, yet worker1 has kept the insert & worker2 is missing it

May be relevant to #2823


serprex:
I've updated this PR with a fix

src/test/regress/expected/multi_modifications_9.out
```diff
@@ -356,6 +356,20 @@ ERROR:  relation "public.limit_orders_750000" does not exist
 -- Second: Move aside limit_orders shard on the second worker node
 ALTER TABLE renamed_orders RENAME TO limit_orders_750000;
 -- Connect back to master node
```
pykello: Move the comment

marcocitus:
I'm not entirely against this change (see #1032), but it could have substantial performance impact on users with replication factor >1.

onderkalaci:
Marking this as related to #2460, and actually fixes the part related to replication factor > 1.

> I'm not entirely against this change (see #1032), but it could have substantial performance impact on users with replication factor >1.

A quick benchmark on my local machine showed that single statement insert throughput dropped %50.

serprex:
@marcocitus alternative is marking shards unhealthy. Feel 1PC should do that regardless of whether we keep this choice or not

onderkalaci:
>@marcocitus alternative is marking shards unhealthy. Feel 1PC should do that regardless of whether we keep this choice or not

@marcocitus @serprex I'd actually vote for doing that, if there is a bug we should fix it. This change might hurt some users really badly and there is no turn back from this via a GUC etc.

PR #2861, opened 2019-08-01
PlanRouterQuery: don't store list of list of shard intervals in relationShardList

serprex:
Also don't call shard intervals shards or list of lists lists


pykello:
Do you have any suggestion on refactoring this function (and this final in general) to make it a bit easier to follow?

serprex:
Honestly once I saw that it was returning a differently typed list based on `multiShardModifyQuery`, I didn't even think it was too bad that it was returning a differently typed list. The function seems quite linear with defined bail out points

PR #2864, opened 2019-08-03
Foreign key between reference tables

pykello:
DESCRIPTION: Support Foreign Key constraints between reference tables.

The changes are:
1. Modify `ErrorIfUnsupportedForeignConstraint()` to not error out for foreign keys for reference tables.
2. Modify `SerializeNonCommutativeWrites()` to lock referenced (and transitively referenced) tables to avoid concurrency issues like https://github.com/citusdata/citus/pull/2864#issuecomment-520020075
3. Update `foreign_key_restriction_enforcement.sql` with tests for dist -> ref -> ref foreign keys. Behaviour should be similar to dist -> ref case.
4. Add `isolation_ref2ref_foreign_keys_on_mx.spec` to make sure that we are getting the correct locks in the first worker to avoid replica divergence in MX.

### DIFF
--- a/src/backend/distributed/commands/foreign_constraint.c
+++ b/src/backend/distributed/commands/foreign_constraint.c
@@ -35,14 +35,19 @@
 static bool HeapTupleOfForeignConstraintIncludesColumn(HeapTuple heapTuple, Oid
 													   relationId, int pgConstraintKey,
 													   char *columnName);
+static void ForeignConstraintFindDistKeys(HeapTuple pgConstraintTuple,
+										  Var *referencingDistColumn,
+										  Var *referencedDistColumn,
+										  int *referencingAttrIndex,
+										  int *referencedAttrIndex);
 
 /*
- * ConstraintIsAForeignKeyToReferenceTable function scans the pgConstraint to
- * fetch all of the constraints on the given relationId and see if at least one
- * of them is a foreign key referencing to a reference table.
+ * ConstraintIsAForeignKeyToReferenceTable checks if the given constraint is a
+ * foreign key constraint from the given relation to a reference table. It does
+ * that by scanning pg_constraint for foreign key constraints.
  */
 bool
-ConstraintIsAForeignKeyToReferenceTable(char *constraintNameInput, Oid relationId)
+ConstraintIsAForeignKeyToReferenceTable(char *constraintName, Oid relationId)
 {
 	Relation pgConstraint = NULL;
 	SysScanDesc scanDescriptor = NULL;
@@ -64,9 +69,9 @@ ConstraintIsAForeignKeyToReferenceTable(char *constraintNameInput, Oid relationI
 	{
 		Oid referencedTableId = InvalidOid;
 		Form_pg_constraint constraintForm = (Form_pg_constraint) GETSTRUCT(heapTuple);
-		char *constraintName = (constraintForm->conname).data;
+		char *tupleConstraintName = (constraintForm->conname).data;
 
-		if (strncmp(constraintNameInput, constraintName, NAMEDATALEN) != 0 ||
+		if (strncmp(constraintName, tupleConstraintName, NAMEDATALEN) != 0 ||
 			constraintForm->conrelid != relationId)
 		{
 			heapTuple = systable_getnext(scanDescriptor);
@@ -95,7 +100,7 @@ ConstraintIsAForeignKeyToReferenceTable(char *constraintNameInput, Oid relationI
 
 
 /*
- * ErrorIfUnsupportedForeignConstraint runs checks related to foreign constraints and
+ * ErrorIfUnsupportedForeignConstraintExists runs checks related to foreign constraints and
  * errors out if it is not possible to create one of the foreign constraint in distributed
  * environment.
  *
@@ -109,11 +114,13 @@ ConstraintIsAForeignKeyToReferenceTable(char *constraintNameInput, Oid relationI
  * - If referenced table is a reference table
  *      - ON DELETE/UPDATE SET NULL, ON DELETE/UPDATE SET DEFAULT and ON UPDATE CASCADE options
  *        are not used on the distribution key of the referencing column.
- * - If referencing table is a reference table, error out
+ * - If referencing table is a reference table, error out if the referenced table is not a
+ *   a reference table.
  */
 void
-ErrorIfUnsupportedForeignConstraint(Relation relation, char distributionMethod,
-									Var *distributionColumn, uint32 colocationId)
+ErrorIfUnsupportedForeignConstraintExists(Relation relation, char referencingDistMethod,
+										  Var *referencingDistKey,
+										  uint32 referencingColocationId)
 {
 	Relation pgConstraint = NULL;
 	SysScanDesc scanDescriptor = NULL;
@@ -123,21 +130,20 @@ ErrorIfUnsupportedForeignConstraint(Relation relation, char distributionMethod,
 
 	Oid referencingTableId = relation->rd_id;
 	Oid referencedTableId = InvalidOid;
-	uint32 referencedTableColocationId = INVALID_COLOCATION_ID;
-	Var *referencedTablePartitionColumn = NULL;
-
-	Datum referencingColumnsDatum = 0;
-	Datum *referencingColumnArray = NULL;
-	int referencingColumnCount = 0;
-	Datum referencedColumnsDatum = 0;
-	Datum *referencedColumnArray = NULL;
-	int referencedColumnCount = 0;
-	bool isNull = false;
-	int attrIdx = 0;
-	bool foreignConstraintOnPartitionColumn = false;
+	uint32 referencedColocationId = INVALID_COLOCATION_ID;
 	bool selfReferencingTable = false;
-	bool referencedTableIsAReferenceTable = false;
-	bool referencingColumnsIncludeDistKey = false;
+	bool referencingNotReplicated = true;
+
+	if (IsDistributedTable(referencingTableId))
+	{
+		/* ALTER TABLE command is applied over single replicated table */
+		referencingNotReplicated = SingleReplicatedTable(referencingTableId);
+	}
+	else
+	{
+		/* Creating single replicated table with foreign constraint */
+		referencingNotReplicated = (ShardReplicationFactor == 1);
+	}
 
 	pgConstraint = heap_open(ConstraintRelationId, AccessShareLock);
 	ScanKeyInit(&scanKey[0], Anum_pg_constraint_conrelid, BTEqualStrategyNumber, F_OIDEQ,
@@ -149,7 +155,15 @@ ErrorIfUnsupportedForeignConstraint(Relation relation, char distributionMethod,
 	while (HeapTupleIsValid(heapTuple))
 	{
 		Form_pg_constraint constraintForm = (Form_pg_constraint) GETSTRUCT(heapTuple);
-		bool singleReplicatedTable = true;
+		bool referencedIsDistributed = false;
+		char referencedDistMethod = 0;
+		Var *referencedDistKey = NULL;
+		bool referencingIsReferenceTable = false;
+		bool referencedIsReferenceTable = false;
+		int referencingAttrIndex = -1;
+		int referencedAttrIndex = -1;
+		bool referencingColumnsIncludeDistKey = false;
+		bool foreignConstraintOnDistKey = false;
 
 		if (constraintForm->contype != CONSTRAINT_FOREIGN)
 		{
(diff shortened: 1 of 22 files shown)


marcocitus:
Do we need any changes to the locking? e.g. if an update cascades from reference table A to reference table B while reference table B is concurrently being updated, could the updates on table B be applied in a different order on different replicas?

pykello:
> Do we need any changes to the locking? e.g. if an update cascades from reference table A to reference table B while reference table B is concurrently being updated, could the updates on table B be applied in a different order on different replicas?

Just to document a case that doesn't work without the locks:

Preparation:

```sql
create table ref(a int primary key, b int);
select create_reference_table('ref');
create table t(a int, b int);
select create_reference_table('t');
alter table t add constraint tfk foreign key (b) references ref(a) on update cascade;
insert into ref values (1, 1), (3, 3);
insert into t values (1, 1), (3, 3);
```

1.sql:
```sql
update ref set a = case when a % 2 == 1 then a + 1 else a - 1 end;
```

2.sql:
```
update t set a = (a + 1) % 1000 where b = 4;
```

Concurrently run 1.sql and 2.sql using pgbench. 2.sql will produce warnings like `WARNING:  modified 1 tuples, but expected to modify 0`, and after pgbench is done, we see that the replicas of `t` have diverged.

I'll add the necessary locks to avoid this.

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -938,13 +956,6 @@ ExecuteSingleSelectTask(CitusScanState *scanState, Task *task)
 			placementAccessList = list_make1(placementAccess);
 		}
 
-		if (placementAccessList == NIL)
```
pykello: Not possible, because of preceding checks.

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -1316,9 +1327,6 @@ GetModifyConnections(Task *task, bool markCritical)
 		multiConnectionList = lappend(multiConnectionList, multiConnection);
 	}
 
-	/* then finish in parallel */
-	FinishConnectionListEstablishment(multiConnectionList);
```
pykello: We are not doing async connection establishment here.

src/test/regress/expected/foreign_key_restriction_enforcement.out
```diff
@@ -249,6 +420,35 @@ DETAIL:  Reference relation "reference_table" is modified, which might lead to d
...
 
+ROLLBACK;
+BEGIN;
+	UPDATE transitive_reference_table SET id = 101 WHERE id = 99;
+DEBUG:  switching to sequential query execution mode
+DETAIL:  Reference relation "transitive_reference_table" is modified, which might lead to data inconsistencies or distributed deadlocks via parallel accesses to hash distributed relations due to foreign keys. Any parallel modification to those hash distributed relations in the same transaction can only be executed in sequential query execution mode
+	SELECT count(*) FROM on_update_fkey_table WHERE value_1 = 101 AND id = 99;
+ count 
+-------
+     0
```
pykello: Maybe we need to create the tables in such a way that changes to `transitive_reference_table` cascades to `on_update_fkey_table` ...

src/backend/distributed/executor/multi_router_executor.c
marcocitus: The error below suggests a condition under which it is possible (FindShardPlacementOnGroup in BuildPlacementAccessList returns NULL because a shard move completed in between planning and execution). We should remove `Assert(list_length(placementAccessList) == list_length(relationShardList));` instead.

src/backend/distributed/utils/resource_lock.c
```diff
@@ -336,21 +338,58 @@ LockReferencedReferenceShardDistributionMetadata(uint64 shardId, LOCKMODE lock)
...
+		!IsFirstWorkerNode())
+	{
+		LockShardListResourcesOnFirstWorker(lockMode, shardIntervalList);
+	}
+
+	foreach(shardIntervalCell, shardIntervalList)
+	{
+		ShardInterval *shardInterval = (ShardInterval *) lfirst(shardIntervalCell);
+		LockShardResource(shardInterval->shardId, lockMode);
+	}
```
marcocitus: would it make sense to use `LockShardListResources` here or did you avoid it to skip the extra sort?

src/backend/distributed/commands/foreign_constraint.c
```diff
@@ -150,65 +150,65 @@ ErrorIfUnsupportedForeignConstraint(Relation relation, char distributionMethod,
...
+			referencedIsReferenceTable =
+				(PartitionMethod(referencedTableId) == DISTRIBUTE_BY_NONE);
+		}
+		else if (selfReferencingTable)
+		{
+			/*
+			 * If the table is just being distributed, IsDistributedTable() returns
+			 * false, so we need a special check for self-referencing tables.
+			 */
+			referencedIsReferenceTable = (distributionMethod == DISTRIBUTE_BY_NONE);
```
marcocitus: could we fill in the distributionMethod at the start if the table already exists to avoid the extra call to PartitionMethod above?

src/backend/distributed/utils/resource_lock.c
```diff
@@ -534,12 +573,21 @@ SerializeNonCommutativeWrites(List *shardIntervalList, LOCKMODE lockMode)
...
 	int64 firstShardId = firstShardInterval->shardId;
 
-	if (ReferenceTableShardId(firstShardId) && ClusterHasKnownMetadataWorkers() &&
-		!IsFirstWorkerNode())
+	if (ReferenceTableShardId(firstShardId))
 	{
-		LockShardListResourcesOnFirstWorker(lockMode, shardIntervalList);
+		if (ClusterHasKnownMetadataWorkers() && !IsFirstWorkerNode())
+		{
+			LockShardListResourcesOnFirstWorker(lockMode, shardIntervalList);
```
marcocitus: are we doing this twice now? (again in LockReferencedReferenceShardResources)

src/test/regress/expected/foreign_key_restriction_enforcement.out
```diff
@@ -249,6 +420,35 @@ DETAIL:  Reference relation "reference_table" is modified, which might lead to d
      1
 (1 row)
 
+ROLLBACK;
+BEGIN;
+	UPDATE transitive_reference_table SET id = 101 WHERE id = 99;
```
marcocitus: was this meant to update id or value_1?

src/test/regress/specs/isolation_ref2ref_foreign_keys_on_mx.spec
```diff
@@ -0,0 +1,153 @@
+setup
+{
+    SELECT citus.replace_isolation_tester_func();
+  	SELECT citus.refresh_isolation_tester_prepared_statement();
+
+	-- start_metadata_sync_to_node can not be run inside a transaction block
+	-- following is a workaround to overcome that
+	-- port numbers are hard coded at the moment
```
marcocitus: could use punctuation :)

marcocitus:
Do we need non-MX isolation tests as well?

src/backend/distributed/utils/resource_lock.c
pykello: > are we doing this twice now? (again in LockReferencedReferenceShardResources)

Here it is for the relation itself, in `LockReferencedReferenceShardResources` it is for referenced relations. I am changing the variable names there to avoid confusion.

src/backend/distributed/commands/foreign_constraint.c
pykello: > could we fill in the distributionMethod at the start if the table already exists to avoid the extra call to PartitionMethod above?

`PartitionMethod` was called once per referenced relation. Since this function was very confusing, I did a total overhaul of it after your comment. Trying to make things a bit more obvious. The logic is divided into three functions:

1. ErrorIfUnsupportedForeignConstraintExists -> Does the high level looping and information preparation.
2. ErrorIfUnsupportedForeignConstraint -> Our main logic
3. ForeignConstraintFindDistKeys -> A low-level detail which took too much space, so I thought it can be refactored out.

The logic is same as before, but in 3 separate functions.

src/backend/distributed/utils/resource_lock.c
pykello: > would it make sense to use `LockShardListResources` here or did you avoid it to skip the extra sort?

Yeah, just wanted to skip the extra sort.

pykello:
> Do we need non-MX isolation tests as well?

Yes. working on it. I'll add them soon.

src/backend/distributed/commands/foreign_constraint.c
```diff
@@ -149,237 +150,291 @@ ErrorIfUnsupportedForeignConstraint(Relation relation, char distributionMethod,
...
-		/*
-		 * If columns in the foreign key includes the distribution key from the
-		 * referencing side, we do not allow update/delete operations through
-		 * foreign key constraints (e.g. ... ON UPDATE SET NULL)
-		 */
+	/*
+	 * We support foreign keys between reference tables. No more checks
+	 * are necessary.
+	 */
+	if (referencingIsReferenceTable && referencedIsReferenceTable)
```
pykello: Except this line + removal of the related error, everything else in this file is just moving things around and renaming variables so code can be understood a bit easier.

pykello:
> > Do we need non-MX isolation tests as well?
> 
> Yes. working on it. I'll add them soon.

Added non-MX isolation tests and fixed a cache invalidation related bug. We need to guarantee `MetadataCache.distColocationRelationId` is initialized to be able to get foreign key graph invalidation messages. Previously this could be 0 when invalidation messages arrived, hence the issue.

marcocitus:
Minor issues with our current implementation of locking in case of foreign keys that do are beyond the scope of this PR:
- A multi-shard update/delete on a distributed table with a foreign key to a reference table could result in a deadlock if there is a concurrent update on the reference table. The solution would be for the update on the reference table to take multi-shard locks.
- Shard moves can always trigger a deadlock because transactions can take metadata locks on co-located shards and reference tables in an arbitrary order, while the shard move takes metadata locks on all shards in the foreign key graph

src/backend/distributed/commands/foreign_constraint.c
```diff
@@ -149,237 +150,291 @@ ErrorIfUnsupportedForeignConstraint(Relation relation, char distributionMethod,
...
-							errmsg("cannot create foreign key constraint because "
-								   "reference tables are not supported as the "
-								   "referencing table of a foreign constraint"),
-							errdetail("Reference tables are only supported as the "
-									  "referenced table of a foreign key when the "
-									  "referencing table is a hash distributed "
-									  "table")));
+			ereport(ERROR, (errcode(ERRCODE_INVALID_TABLE_DEFINITION),
+							errmsg("cannot create foreign key constraint"),
+							errdetail("Referenced table must be a distributed table.")));
```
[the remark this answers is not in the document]
pykello: done

src/backend/distributed/commands/foreign_constraint.c
```diff
@@ -149,237 +150,291 @@ ErrorIfUnsupportedForeignConstraint(Relation relation, char distributionMethod,
...
+	 * We do not allow to create foreign constraints if shard replication factor is
+	 * greater than 1. Because in our current design, multiple replicas may cause
+	 * locking problems and inconsistent shard contents.
+	 *
+	 * Note that we allow referenced table to be a reference table (e.g., not a
+	 * single replicated table). This is allowed since (a) we are sure that
+	 * placements always be in the same state (b) executors are aware of reference
+	 * tables and handle concurrency related issues accordingly.
+	 */
+	if (!referencingSingleReplicated)
```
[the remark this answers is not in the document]
pykello: done

src/backend/distributed/commands/foreign_constraint.c
```diff
@@ -149,237 +150,291 @@ ErrorIfUnsupportedForeignConstraint(Relation relation, char distributionMethod,
...
+ */
+static void
+ErrorIfUnsupportedForeignConstraint(char referencingDistMethod,
+									Var *referencingDistColumn,
+									uint32 referencingColocationId,
+									bool referencingSingleReplicated,
+									char referencedDistMethod,
+									Var *referencedDistColumn,
+									uint32 referencedColocationId,
+									HeapTuple pgConstraintTuple)
```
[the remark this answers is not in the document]
pykello: inlined this function. I think renaming the variables and reordering the code a bit is enough for making the code more understandable.

src/test/regress/specs/isolation_ref2ref_foreign_keys.spec
```diff
@@ -0,0 +1,114 @@
...
+permutation "s2-begin" "s2-delete-table-2" "s1-view-locks" "s2-rollback" "s1-view-locks"
+# Case 3. Modifying ref_table_3 should also lock ref_table_1 and ref_table_2 shards in Exclusive mode.
+permutation "s2-begin" "s2-update-table-3" "s1-begin" "s1-view-locks" "s1-rollback" "s2-rollback" "s1-view-locks"
+permutation "s2-begin" "s2-delete-table-3" "s1-begin" "s1-view-locks" "s1-rollback" "s2-rollback" "s1-view-locks"
+# Case 4. Inserting into ref_table_1 should only lock its own shard in RowExclusive mode.
+permutation "s2-begin" "s2-insert-table-1" "s1-view-locks" "s2-rollback" "s1-view-locks"
+# Case 5. Modifying ref_table_2 should also lock ref_table_1 in RowExclusive mode.
+permutation "s2-begin" "s2-insert-table-2" "s1-view-locks" "s2-rollback" "s1-view-locks"
+# Case 6. Modifying ref_table_2 should also lock ref_table_1 in RowExclusive mode.
+permutation "s2-begin" "s2-insert-table-3" "s1-view-locks" "s2-rollback" "s1-view-locks"
```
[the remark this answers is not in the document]
pykello: Added some more tests.

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -2637,6 +2637,14 @@ InitializeCaches(void)
...
 			RegisterWorkerNodeCacheCallbacks();
 			RegisterLocalGroupIdCacheCallbacks();
+
+			/*
+			 * This needs to be initialized so we can receive foreign relation graph
+			 * invalidation messages in InvalidateForeignRelationGraphCacheCallback().
+			 * See the comments of InvalidateForeignKeyGraph for more context.
+			 */
+			MetadataCache.distColocationRelationId =
+				get_relname_relid("pg_dist_colocation", PG_CATALOG_NAMESPACE);
```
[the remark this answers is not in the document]
pykello: I need to think about this a bit more. It seems that this `InitializeCaches()` can be called even before creation of extension in which case `DistPartitionRelationId()` will error out. What I added will set `distColocationRelationId` to `InvalidOid`, I am not sure if this is ok. I just need to think more.

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -3089,7 +3097,11 @@ ResetDistTableCacheEntry(DistTableCacheEntry *cacheEntry)
 static void
 InvalidateForeignRelationGraphCacheCallback(Datum argument, Oid relationId)
 {
-	/* when invalidation happens simply set the LocalGroupId to the default value */
+	/*
+	 * We shouldn't call DistColocationRelationId() here since cache
+	 * invalidation can happen before postgres' relation cache being
+	 * initialized, in which case get_relname_relid() fails.
```
pykello: this comment is not accurate. I'll update it after spending time on understanding things.

src/backend/distributed/commands/foreign_constraint.c
```diff
@@ -149,132 +155,110 @@ ErrorIfUnsupportedForeignConstraint(Relation relation, char distributionMethod,
...
-							errmsg("cannot create foreign key constraint because "
-								   "reference tables are not supported as the "
-								   "referencing table of a foreign constraint"),
-							errdetail("Reference tables are only supported as the "
-									  "referenced table of a foreign key when the "
-									  "referencing table is a hash distributed "
-									  "table")));
+			ereport(ERROR, (errcode(ERRCODE_INVALID_TABLE_DEFINITION),
+							errmsg("cannot create foreign key constraint"),
+							errdetail("Referenced table must be a distributed table"
```
marcocitus: maybe "a distributed table or a reference table"

src/backend/distributed/utils/metadata_cache.c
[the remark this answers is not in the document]
marcocitus: Could we do this in CitusHasBeenLoaded?

src/backend/distributed/commands/foreign_constraint.c
pykello: > maybe "a distributed table or a reference table"

` a reference table` part is in the next line

PR #2865, opened 2019-08-05
Update dependencies in CONTRIBUTING.md

hanefi:
- Updated `Contributing.md` with recent package versions.
- Updated rpm repo description because RPM no longer has separate repos for different Postgres versions.
- Added libcurl and autoconf to required dependencies

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -81,7 +81,7 @@ why we ask this as well as instructions for how to proceed, see the
   sudo yum groupinstall -y 'Development Tools'
   sudo yum install -y postgresql11-devel postgresql11-server    \
                       libxml2-devel libxslt-devel openssl-devel \
-                      pam-devel readline-devel git
+                      pam-devel readline-devel git libcurl-devel
 
   git clone https://github.com/citusdata/citus.git
   cd citus
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -47,7 +47,7 @@ why we ask this as well as instructions for how to proceed, see the
        sudo apt-key add -
   sudo apt-get update
 
-  sudo apt-get install -y postgresql-server-dev-10 postgresql-10 \
+  sudo apt-get install -y postgresql-server-dev-11 postgresql-11 \
                           libedit-dev libselinux1-dev libxslt-dev  \
                           libpam0g-dev git flex make libssl-dev    \
                           libkrb5-dev
@@ -67,7 +67,7 @@ why we ask this as well as instructions for how to proceed, see the
 
 #### Red Hat-based Linux (RHEL, CentOS, Fedora)
 
-1. Find the PostgreSQL 10 RPM URL for your repo at [yum.postgresql.org](http://yum.postgresql.org/repopackages.php#pg10)
+1. Find the RPM URL for your repo at [yum.postgresql.org](http://yum.postgresql.org/repopackages.php)
 2. Register its contents with Yum:
 
   ```bash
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -79,13 +79,13 @@ why we ask this as well as instructions for how to proceed, see the
...
-  sudo yum install -y postgresql96-devel postgresql96-server    \
+  sudo yum install -y postgresql11-devel postgresql11-server    \
                       libxml2-devel libxslt-devel openssl-devel \
                       pam-devel readline-devel git
 
   git clone https://github.com/citusdata/citus.git
   cd citus
-  PG_CONFIG=/usr/pgsql-9.6/bin/pg_config ./configure
+  PG_CONFIG=/usr/pgsql-11/bin/pg_config ./configure
   make
```
thanodnl: make fails as well:
```
# make -sj8
make[1]: /opt/rh/llvm-toolset-7/root/usr/bin/clang: Command not found
make[1]: /opt/rh/llvm-toolset-7/root/usr/bin/clang: Command not found
make[1]: *** [shared_library_init.bc] Error 127
make[1]: *** Waiting for unfinished jobs....
make[1]: *** [commands/cluster.bc] Error 127
make: *** [extension] Error 2
```

Seems to be linking with clang :S, a compile (without `-s`) gives better output.

CONTRIBUTING.md
```diff
@@ -79,13 +79,13 @@ why we ask this as well as instructions for how to proceed, see the
...
   sudo yum groupinstall -y 'Development Tools'
-  sudo yum install -y postgresql96-devel postgresql96-server    \
+  sudo yum install -y postgresql11-devel postgresql11-server    \
                       libxml2-devel libxslt-devel openssl-devel \
                       pam-devel readline-devel git
 
   git clone https://github.com/citusdata/citus.git
   cd citus
-  PG_CONFIG=/usr/pgsql-9.6/bin/pg_config ./configure
+  PG_CONFIG=/usr/pgsql-11/bin/pg_config ./configure
```
thanodnl: centos also needs libcurl:
```
$ PG_CONFIG=/usr/pgsql-11/bin/pg_config ./configure
...
configure: error: libcurl not found
```

```
yum install libcurl-devel
```
solved it for me

CONTRIBUTING.md
hanefi: The yum postgresql package is compiled with llvm to enable JIT features. To build an extension with JIT support one needs to install Clang/LLVM which is not trivial.

Problems:
- Clang/LLVM is not on usual rpm repositories.
- RHEL and CentOS users need to set up different repos for setting up clang and llvm. The official recommended steps for RHEL are quite long: https://developers.redhat.com/blog/2018/07/07/yum-install-gcc7-clang/

My proposed solution:
(i) Put some links up for setting up clang/llvm on different distros  using RPM

hanefi: Some more info:
When I install `postgresql11.x86_64` rpm package on a CentOS 7 docker container, I see the following in the `pg_config` output:
```
LDFLAGS = -Wl,--as-needed -L/usr/lib64/llvm5.0/lib ....
```

The problem is llvm5.0 package is not on the default YUM repo or the PGDG one. One needs to set it up using another repository

thanodnl: That is unfortunate, guess it is non-trivial to disable llvm.

In that sense we can keep it like this, although it would be great if we have the steps documented here. As people will run into this problem when they are reading this doc.

CONTRIBUTING.md
thanodnl: once libcurl-devel is added we can approve this PR.

CONTRIBUTING.md
hanefi: I checked out how our packaging images are setup for RHEL/CentOS packages. Added one more step to install EPEL repos. During my research I came across this explanation in [PostgreSQL Rpm repositories](https://yum.postgresql.org/repopackages.php):

`
Please note that PostgreSQL YUM repository depends on EPEL repository for some packages. RHEL/CentOS/, etc. users should install EPEL repo RPM along with PGDG repo RPMs to satisfy dependencies.
`

PR #2866, opened 2019-08-05
Do not record relation accessess unnecessarily

onderkalaci:
DESCRIPTION: Record single connection relation accesses only for reference tables

Before this commit, we recorded the relation accesses in 3 different
places
    - FindPlacementListConnection         -- applies to all executors in tx block
    - StartPlacementExecutionOnSession()  -- adaptive executor only
    - StartPlacementListConnection()      -- router/real-time only

This is different than Citus 8.2, and could lead to query execution times
increase considerably on multi-shard commands in transaction block
that are on partitioned tables.

When we implemented foreign keys to reference tables, we first implemented
a relation access tracking infrastructure (#2220). That was generic enough
to follow all the relation accesses.

However, for the purposes of foreign keys to reference tables, we only need
to track relation accesses to reference tables.

This change especially important because the relation access tracking is
very inefficient for partitioned tables with many partitions. So, we decided
to only record relevant tables, which are reference tables only.

(Note that we're already tracking the Parallel relation accesses via
RecordParallelRelationAccessForTaskList() once per distributed query.)

### DIFF
--- a/src/backend/distributed/commands/multi_copy.c
+++ b/src/backend/distributed/commands/multi_copy.c
@@ -2346,6 +2346,8 @@ CitusCopyDestReceiverStartup(DestReceiver *dest, int operation,
 
 	copyDest->shardStateHash = CreateShardStateHash(TopTransactionContext);
 	copyDest->connectionStateHash = CreateConnectionStateHash(TopTransactionContext);
+
+	RecordRelationAccessIfReferenceTable(tableId, PLACEMENT_ACCESS_DML);
 }
 
 
--- a/src/backend/distributed/commands/table.c
+++ b/src/backend/distributed/commands/table.c
@@ -418,11 +418,15 @@ PlanAlterTableStmt(AlterTableStmt *alterTableStatement, const char *alterTableCo
 															   command);
 	}
 
+	if (executeSequentially)
+	{
+		SetLocalMultiShardModifyModeToSequential();
+	}
+
 	ddlJob = palloc0(sizeof(DDLJob));
 	ddlJob->targetRelationId = leftRelationId;
 	ddlJob->concurrentIndexCmd = false;
 	ddlJob->commandString = alterTableCommand;
-	ddlJob->executeSequentially = executeSequentially;
 
 	if (rightRelationId)
 	{
--- a/src/backend/distributed/commands/utility_hook.c
+++ b/src/backend/distributed/commands/utility_hook.c
@@ -553,7 +553,7 @@ static void
 ExecuteDistributedDDLJob(DDLJob *ddlJob)
 {
 	bool shouldSyncMetadata = ShouldSyncTableMetadata(ddlJob->targetRelationId);
-	int targetPoolSize = ddlJob->executeSequentially ? 1 : MaxAdaptiveExecutorPoolSize;
+	int targetPoolSize = MaxAdaptiveExecutorPoolSize;
 
 	EnsureCoordinator();
 	EnsurePartitionTableNotReplicated(ddlJob->targetRelationId);
@@ -579,8 +579,7 @@ ExecuteDistributedDDLJob(DDLJob *ddlJob)
 		}
 
 		/* use adaptive executor when enabled */
-		ExecuteUtilityTaskListWithoutResults(ddlJob->taskList, targetPoolSize,
-											 ddlJob->executeSequentially);
+		ExecuteUtilityTaskListWithoutResults(ddlJob->taskList, targetPoolSize);
 	}
 	else
 	{
@@ -592,8 +591,7 @@ ExecuteDistributedDDLJob(DDLJob *ddlJob)
 		PG_TRY();
 		{
 			/* use adaptive executor when enabled */
-			ExecuteUtilityTaskListWithoutResults(ddlJob->taskList, targetPoolSize,
-												 ddlJob->executeSequentially);
+			ExecuteUtilityTaskListWithoutResults(ddlJob->taskList, targetPoolSize);
 
 			if (shouldSyncMetadata)
 			{
--- a/src/backend/distributed/commands/utility_hook.c
+++ b/src/backend/distributed/commands/utility_hook.c
@@ -553,7 +553,6 @@ static void
 ExecuteDistributedDDLJob(DDLJob *ddlJob)
 {
 	bool shouldSyncMetadata = ShouldSyncTableMetadata(ddlJob->targetRelationId);
-	int targetPoolSize = MaxAdaptiveExecutorPoolSize;
 
 	EnsureCoordinator();
 	EnsurePartitionTableNotReplicated(ddlJob->targetRelationId);
@@ -579,7 +578,7 @@ ExecuteDistributedDDLJob(DDLJob *ddlJob)
 		}
 
 		/* use adaptive executor when enabled */
-		ExecuteUtilityTaskListWithoutResults(ddlJob->taskList, targetPoolSize);
+		ExecuteUtilityTaskListWithoutResults(ddlJob->taskList);
 	}
 	else
 	{
@@ -591,7 +590,7 @@ ExecuteDistributedDDLJob(DDLJob *ddlJob)
 		PG_TRY();
 		{
 			/* use adaptive executor when enabled */
-			ExecuteUtilityTaskListWithoutResults(ddlJob->taskList, targetPoolSize);
+			ExecuteUtilityTaskListWithoutResults(ddlJob->taskList);
 
 			if (shouldSyncMetadata)
 			{
--- a/src/backend/distributed/connection/placement_connection.c
+++ b/src/backend/distributed/connection/placement_connection.c
@@ -366,6 +366,8 @@ AssignPlacementListToConnection(List *placementAccessList, MultiConnection *conn
 		ConnectionPlacementHashEntry *placementEntry = NULL;
 		ConnectionReference *placementConnection = NULL;
 
+		Oid relationId = InvalidOid;
+
 		if (placement->shardId == INVALID_SHARD_ID)
 		{
 			/*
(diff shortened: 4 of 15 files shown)


src/backend/distributed/commands/multi_copy.c
```diff
@@ -2468,6 +2468,10 @@ CitusSendTupleToPlacements(TupleTableSlot *slot, CitusCopyDestReceiver *copyDest
 
 		if (switchToCurrentPlacement)
 		{
+			List *placementList = ShardPlacementList(shardId);
```
onderkalaci: This is necessary because otherwise COPY code-path wouldn't call `AssociatePlacementAccessWithRelation ()` at all. I think this is a decent place to have this call.

src/backend/distributed/commands/table.c
```diff
@@ -418,11 +418,15 @@ PlanAlterTableStmt(AlterTableStmt *alterTableStatement, const char *alterTableCo
 															   command);
 	}
 
+	if (executeSequentially)
```
onderkalaci: Deferring this decision leads to weird edge cases on adaptive executor because there is an important reference to `SEQUENTIAL_CONNECTION` on `UseConnectionPerPlacement()`.

I think this is easier to follow and makes things more robust.

src/backend/distributed/commands/utility_hook.c
```diff
@@ -553,7 +553,7 @@ static void
 ExecuteDistributedDDLJob(DDLJob *ddlJob)
 {
 	bool shouldSyncMetadata = ShouldSyncTableMetadata(ddlJob->targetRelationId);
-	int targetPoolSize = ddlJob->executeSequentially ? 1 : MaxAdaptiveExecutorPoolSize;
+	int targetPoolSize = MaxAdaptiveExecutorPoolSize;
```
marcocitus: is it still worth keeping this as a parameter?

src/backend/distributed/commands/multi_copy.c
marcocitus: Wouldn't this happen quite often?

onderkalaci: For COPY without in tx block, once per placement. Because, we'll be only switching to the current placement once at the beginning of the COPY command.

For COPY (or anything that copies into a distributed table) inside a tx, if the previous command multiplexed some commands over a single connection, then we'll get to here once in 4MB COPY into the placement as well.

I guess this is okayish, because we don't have many choices here since we don't know which placement are going to be accessed by the COPY. 

Especially with #2867, this'll be almost a no-op so I didn't want to add any extra logic to keep track of which placements are already accessed.

src/backend/distributed/commands/utility_hook.c
onderkalaci: Good point,  we can remove the parameter on `ExecuteUtilityTaskListWithoutResults`.

src/backend/distributed/commands/multi_copy.c
onderkalaci: `ShardPlacementList ()` is not a free function even if `AssociatePlacementAccessWithRelation ` could be. One thing I considered was to have another api for `AssociatePlacementAccessWithRelation ()`, which gets the shard itself. In fact, we're only using the `shardId` there. 

Do you have any preference here?

src/backend/distributed/commands/multi_copy.c
```diff
@@ -2480,6 +2484,12 @@ CitusSendTupleToPlacements(TupleTableSlot *slot, CitusCopyDestReceiver *copyDest
 
 			/* additionaly, we need to send the current tuple too */
 			sendTupleOverConnection = true;
+
+			/* make sure that we keep track of relation accesses */
+			placement =
+				SearchShardPlacementInList(placementList, connection->hostname,
+										   connection->port, false);
+			AssociatePlacementAccessWithRelation(placement, PLACEMENT_ACCESS_DML);
```
marcocitus: This function is actually kind of dumb. It does the same thing for every placement in a relation. Can we just do this once for the whole statement?

src/backend/distributed/connection/placement_connection.c
```diff
@@ -700,9 +700,6 @@ FindPlacementListConnection(int flags, List *placementAccessList, const char *us
 			Assert(!placementConnection->hadDML);
 			Assert(accessType != PLACEMENT_ACCESS_DDL);
 		}
-
-		/* record the relation access mapping */
-		AssociatePlacementAccessWithRelation(placement, accessType);
```
marcocitus: What about the one in `AssignPlacementListToConnection`? Can we move that somewhere else to not do it per placement?

In fact, can we eliminate `AssociatePlacementAccessWithRelation` to make sure we don't start calling it again?

src/backend/distributed/commands/multi_copy.c
onderkalaci: We do it for parallel queries via `RecordParallelRelationAccessForTaskList()`. I think we don't like that approach much as well because that requires any code-refactoring to keep that in mind explicitly calling the function. Could be an option with some larger refactoring that touches router/real-time executors as well.

Besides, I'm thinking that #2867 is the complementary change for this. With that, (a) we're already guarding against disasters like this -- I might consider adding more comments (b) giving a chance to have the infrastructure to record all relation accesses pretty easily in case any future change needs it. Obviously we should improve the performance first.

marcocitus: ok, but why would we do it per placement in the COPY logic instead of per relation in the COPY logic. Wouldn't the latter be less code to maintain?

The naming and signature of `AssociatePlacementAccessWithRelation` also seems quite misleading, which is likely what led to this bug.

onderkalaci: > ok, but why would we do it per placement in the COPY logic instead of per relation in the COPY logic. Wouldn't the latter be less code to maintain?

Makes sense.

> The naming and signature of AssociatePlacementAccessWithRelation also seems quite misleading, which is likely what led to this bug.

That's also true. I think the intention for fkey to reference tables was to only record it for reference tables. But since we implemented the relation accesses as a separate infrastructure, we ended up with.

src/backend/distributed/connection/placement_connection.c
onderkalaci: I think the current changes should meet your expectations, which I also agree btw.

src/test/regress/expected/foreign_key_restriction_enforcement_0.out
```diff
@@ -1064,8 +1064,6 @@ ADD CONSTRAINT
 	fkey_delete FOREIGN KEY(value_1) 
 REFERENCES 
 	reference_table(id) ON DELETE CASCADE;
-DEBUG:  switching to sequential query execution mode
-DETAIL:  Reference relation "reference_table" is modified, which might lead to data inconsistencies or distributed deadlocks via parallel accesses to hash distributed relations due to foreign keys. Any parallel modification to those hash distributed relations in the same transaction can only be executed in sequential query execution mode
```
pykello: Can we add a `SHOW citus.multi_shard_modify_mode;` here to check that connection is sequential?

src/backend/distributed/commands/multi_copy.c
```diff
@@ -2346,6 +2346,8 @@ CitusCopyDestReceiverStartup(DestReceiver *dest, int operation,
 
 	copyDest->shardStateHash = CreateShardStateHash(TopTransactionContext);
 	copyDest->connectionStateHash = CreateConnectionStateHash(TopTransactionContext);
+
+	RecordRelationAccessIfReferenceTable(tableId, PLACEMENT_ACCESS_DML);
```
pykello: Do we also need to set the connection type to sequential if this is a reference table? Maybe that can be a separate PR with its own tests though.

pykello: It seems that we

> Do we also need to set the connection type to sequential if this is a reference table? Maybe that can be a separate PR with its own tests though.

Never mind, calling `RecordRelationAccessIfReferenceTable()` does the switch.

src/test/regress/expected/relation_access_tracking.out
```diff
@@ -577,20 +593,20 @@ BEGIN;
...
 
 	SELECT * FROM relation_acesses  WHERE table_name IN ('table_6', 'table_1') ORDER BY 1,2;
- table_name |  select_access  |  dml_access  |  ddl_access  
-------------+-----------------+--------------+--------------
- table_1    | parallel_access | not_accessed | not_accessed
- table_6    | parallel_access | not_accessed | not_accessed
+ table_name |  select_access  |      dml_access       |      ddl_access       
+------------+-----------------+-----------------------+-----------------------
+ table_1    | parallel_access | not_parallel_accessed | not_parallel_accessed
+ table_6    | parallel_access | not_accessed          | not_accessed
```
pykello: `table_6` is a reference table. Why is it marked as `parallel_access` and not `reference_table_access`?

onderkalaci: I also realized this. 

The query is reference table join with a distributed table. That seems like a minor bug in the implementation before this change. We could fix it, but seems not appropriate for backport?

src/test/regress/expected/foreign_key_restriction_enforcement_0.out
onderkalaci: We're not inside a transaction block, and as far as I understand this test aims to make sure that adding fkey to reference table works without in a tx block.

If we cannot switch to sequential mode, we'd simply get into a self distributed-deadlock. So, is that good enough for you?

src/backend/distributed/transaction/relation_access_tracking.c
```diff
@@ -148,25 +148,35 @@ AllocateRelationAccessHash(void)
...
-	relationId = RelationIdForShard(shardId);
+	/*
+	 * We keep track of relation accesses for the purposes of foreign keys to
+	 * reference tables. So, other distributed tables are not relevant for now.
+	 * Additionally, partitioned tables with lots of partitions might require
+	 * recursively calling RecordRelationAccessBase(), so becareful about
+	 * removing this check.
+	 */
+	cache = DistributedTableCacheEntry(relationId);
+	if (cache->partitionMethod != DISTRIBUTE_BY_NONE)
```
marcocitus: We probably have a PartitionMethod function somewhere.

PR #2871, opened 2019-08-07
Update tests in preparation for pg12

serprex:
This is the result of dropping 9ac4478 & 7c4e654 from postgres12 branch of #2844


src/test/regress/sql/multi_repair_shards.sql
```diff
@@ -62,15 +62,12 @@ SELECT master_copy_shard_placement(:newshardid, 'localhost', :worker_1_port, 'lo
 INSERT INTO customer_engagements VALUES (4, '04-01-2015', 'fourth event');
 ROLLBACK;
 
--- add a fake healthy placement for the tests
-
-INSERT INTO pg_dist_placement (groupid, shardid, shardstate, shardlength)
-							 VALUES (:worker_2_group, :newshardid, 1, 0);
+-- deactivate placement
```
onderkalaci: good catch!

src/test/regress/expected/multi_partitioning_1.out
```diff
@@ -0,0 +1,1939 @@
+--
```
onderkalaci: I think we should drop adding the file right now, and do it whenever we actually need it on pg12 branch?

src/test/regress/pg_regress_multi.pl
```diff
@@ -289,6 +289,7 @@ sub revert_replace_postgres
 # not required, and we don't necessarily have access to the default directory
 push(@pgOptions, '-c', "unix_socket_directories=");
 push(@pgOptions, '-c', "fsync=off");
+push(@pgOptions, '-c', "extra_float_digits=0");
```
onderkalaci: This is already the default, why do we set it again?

src/test/regress/expected/subquery_in_where.out
```diff
@@ -39,17 +39,18 @@ ERROR:  cannot pushdown the subquery
 DETAIL:  Complex subqueries and CTEs are not allowed in the FROM clause when the query has subqueries in the WHERE clause and it references a column from another query
 -- Recurring tuples as empty join tree 
 SELECT * 
-FROM   (SELECT 1 AS id, 
-               2 AS value_1, 
-               3 AS value_3) AS tt1 
+FROM   (SELECT 1 AS id, 2 AS value_1, 3 AS value_3
```
onderkalaci: Can we leave this change when we work on `RESULT_EMPTY_RTE` (or whatever the name)?

src/test/regress/bin/diff
```diff
@@ -37,13 +37,13 @@ then
...
 
-    $DIFF $args $file1.modified $file2.modified
+    $DIFF -w $args $file1.modified $file2.modified
     exitcode=$?
 
     rm -f $file1.modified $file2.modified
 else
     # if test is not in normalized_tests.lst, just diff without normalizing.
-    $DIFF $args $file1 $file2
+    $DIFF -w $args $file1 $file2
```
onderkalaci: This change looks fine to me, ignoring white spaces actually makes sense.

But I'd like to get @pykello's opinion on this as well. Do you see any issues with this? Thanks!

src/test/regress/bin/normalize.sed
```diff
@@ -65,3 +68,17 @@ s/ERROR:  failed to execute task [0-9]+/ERROR:  failed to execute task X/g
 # normalize file names for partitioned files
 s/(task_[0-9]+\.)[0-9]+/\1xxxx/g
 s/(job_[0-9]+\/task_[0-9]+\/p_[0-9]+\.)[0-9]+/\1xxxx/g
+
+# Line info varies between versions
+/^LINE [0-9]+:.*$/d
+/^\s*\^\s*$/d
+
+# pg12 changes
+s/Partitioned table "/Table "/g
```
onderkalaci: Do you remember why we need `/Partitioned table "/Table "/g` and `invalid input syntax for type /invalid input syntax for /g`? 

If that's necessary for future changes, can we remove it for now and add whenever necessary?

(normilize.sed is super powerful and could potentially hide some bugs/issues etc, so I'm trying to be cautious while using it. )

I'd probably prefer to defer `Normalize for pg12` commit to future changes.

src/test/regress/pg_regress_multi.pl
serprex: https://www.postgresql.org/docs/12/release-12.html

> Improve performance by changing the default number of trailing digits output for REAL and DOUBLE PRECISION values (Andrew Gierth)
> Previously, float values were output rounded to 6 or 15 decimals by default. Now, only the number of digits required to preserve the exact binary value is output. The previous behavior can be restored by setting extra_float_digits to zero.

src/test/regress/bin/diff
serprex: Primary reason for adding this is to deal with normalization rules which clear an entire line leaving a newline difference

src/test/regress/bin/normalize.sed
serprex: These are changes in `\dt` & postgres error messages. It can make sense to separate the pg12 rules to the pg12 PR

src/test/regress/expected/multi_subquery_in_where_reference_clause.out
```diff
@@ -87,7 +87,7 @@ DETAIL:  Functions are not allowed in FROM clause when the query has subqueries
 SELECT
   user_id
 FROM
-  (SELECT  5 AS user_id) users_reference_table
+  (SELECT  5 AS user_id UNION ALL SELECT 6) users_reference_table
```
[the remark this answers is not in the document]
serprex: pg12 has gotten a lot better at optimizing basic `select x` queries, so that without these complication tactics we end up with changed subquery/task ids & constant propagation

src/test/regress/expected/multi_modifying_xacts.out
```diff
@@ -1186,11 +1186,13 @@ COPY reference_failure_test FROM STDIN WITH (FORMAT 'csv');
 ERROR:  connection error: localhost:57637
 COMMIT;
 -- show that no data go through the table and shard states are good
+SET client_min_messages to 'ERROR';
```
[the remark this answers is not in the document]
serprex: In pg12 the order of workers changed, so that we get a warning about the select failing on the first worker before the second worker gives us the correct result. In pg11 we happen to hit the healthy worker first, but would warn if we changed the order & tried the unhealthy worker first

src/test/regress/expected/foreign_key_to_reference_table.out
```diff
@@ -1214,24 +1213,24 @@ SELECT create_distributed_table('referencing_referencing_table', 'id');
...
+ fkey_ref_7000301                                     | fkey_reference_table.referencing_table_7000301             | fkey_reference_table.referenced_table_7000298
+ fkey_ref_7000302                                     | fkey_reference_table.referencing_table_7000302             | fkey_reference_table.referenced_table_7000298
+ fkey_ref_7000303                                     | fkey_reference_table.referencing_table_7000303             | fkey_reference_table.referenced_table_7000298
+ fkey_ref_7000304                                     | fkey_reference_table.referencing_table_7000304             | fkey_reference_table.referenced_table_7000298
+ fkey_ref_7000305                                     | fkey_reference_table.referencing_table_7000305             | fkey_reference_table.referenced_table_7000298
+ fkey_ref_7000306                                     | fkey_reference_table.referencing_table_7000306             | fkey_reference_table.referenced_table_7000298
+ referencing_referencing_table_id_ref_id_fkey_7000307 | fkey_reference_table.referencing_referencing_table_7000307 | fkey_reference_table.referencing_table_7000299
```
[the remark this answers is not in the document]
serprex: https://github.com/postgres/postgres/commit/f177660ab01e53dd5597b195dcc8526baa5cfcbd

src/test/regress/expected/window_functions_2.out
```diff
@@ -0,0 +1,1144 @@
+-- ===================================================================
```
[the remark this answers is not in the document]
serprex: Not sure how we can deal with the differing order of sort keys? I'd suggested to @pykello that we could sort all the lines of a file when diffing

src/test/regress/expected/multi_partitioning_1.out
serprex: This is an expected result for pg11

src/test/regress/expected/subquery_in_where.out
serprex: pg12 flattens the previous subquery. Some of these tests might be off, but the `RTE_RESULT` change is part of what is allowing pg12 to more aggressively handle these types of queries. It might be correct to create an `_2.out` to keep the single row subquery

src/test/regress/bin/normalize.sed
onderkalaci: > These are changes in \dt & postgres error messages. It can make sense to separate the pg12 rules to the pg12 PR

Thanks, than let's remove those from this change before merging in.

src/test/regress/expected/foreign_key_to_reference_table.out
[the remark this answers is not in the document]
onderkalaci: This is not backported to REL_11_STABLE, right? So how does CIRCLE CI pass the tests with with pg11 when we change this?

src/test/regress/expected/multi_subquery_in_where_reference_clause.out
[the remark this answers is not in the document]
onderkalaci: I'm concerned about this change because it changes the query in a way that Citus behaves (e.g., plans) differently. So, we're not testing the same thing anymore from Citus' perspective. 

Can we find another way of having a subquery without FROM clause? From example, does defining function which always returns 5, and calling it help?

I'm happy to remove it now and leave it pg12 change btw.

src/test/regress/expected/subquery_in_where.out
onderkalaci: No, I think it is fine to keep as is for now.

src/test/regress/expected/window_functions_2.out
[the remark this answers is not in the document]
onderkalaci: I think the SortKey has changed because you've added an ORDER BY to the window function, right? So, shouldn't we simply update `window_functions.out` file instead of adding a new one?

src/test/regress/expected/multi_partitioning_1.out
onderkalaci: Wait, then why do we need to add now?

src/test/regress/expected/foreign_key_to_reference_table.out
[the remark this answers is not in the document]
serprex: `normalize.sed`

src/test/regress/expected/window_functions_2.out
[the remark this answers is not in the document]
serprex: It changed because pg12 likes to do things in different orders than pg11 when the order doesn't matter. The query wasn't changed

src/test/regress/expected/multi_partitioning_1.out
serprex: Fixed. This isn't an addition, it's renaming `multi_partitioning_9.out` to `multi_partitioning_0.out`. This file's a bit of a mess since testing framework changes leaves us having 4 versions once we add in `_2.out`

src/test/regress/expected/window_functions_2.out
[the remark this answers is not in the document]
onderkalaci: I think adding a new test output is an extreme action for a case where only the explain output has changed. The cost of maintaining extra files is really high.

I think we should find those queries and consider re-writing them in a way that they give consistent outputs. Because in most cases, we're really not interested in the explain output of the worker nodes/shards. If that's the case, lets just do this: https://github.com/citusdata/citus/pull/2793

src/test/regress/expected/subquery_in_where.out
serprex: By "as is" do you mean with the `UNION` or as it is before this PR?

PR #2877, opened 2019-08-09
Add UDF citus_(prepare|finish)_pg_upgrade to aid with upgrading citus

thanodnl:
DESCRIPTION: Add functions to help with postgres upgrades

Currently there is [a list of manual steps](https://docs.citusdata.com/en/v8.2/admin_guide/upgrading_citus.html?highlight=upgrade#upgrading-postgresql-version-from-10-to-11) to perform during a postgres upgrade. These steps guarantee our catalog tables are kept and counter values are maintained across upgrades.

Having more than 1 command in our docs for users to manually execute during upgrades is error prone for both the user, and our docs. There are already 2 catalog tables that have been introduced to citus that have not been added to our docs for backing up during upgrades (`pg_authinfo` and `pg_dist_poolinfo`).

As we add more functionality to citus we run into situations where there are more steps required either before or after the upgrade. At the same time, when we move catalog tables to a place where the contents will be maintained automatically during upgrades we could have less steps in our docs. This will come to a hard to maintain matrix of citus versions and steps to be performed.

Instead we could take ownership of these steps within the extension itself. This PR introduces two new functions for the user to use instead of long lists of error prone instructions to follow.
 - `citus_prepare_pg_upgrade`
    This function should be called by the user right before shutting down the cluster. This will ensure all citus catalog tables are backed up in a location where the information will be retained during an upgrade.
- `citus_finish_pg_upgrade`
    This function should be called right after a pg_upgrade of the cluster. This will restore the catalog tables to the state before the upgrade happend.

Both functions need to be executed both on the coordinator and on all the workers, in the same fashion our current documentation instructs to do.

There are two known problems with this function in its current form, which is also a problem with our docs. We should schedule time in the future to improve on this, but having it automated now is better as we are about to add extra steps to take after upgrades.
 - When you install citus in a clean cluster we do enable ssl for communication between the coordinator and the workers. If an upgrade to a clean cluster is performed we do not setup ssl on the new cluster causing the communication to fail.
 - There are no automated tests added in this PR to execute an upgrade test durning every build. 
    Our current test infrastructure does not allow for 2 versions of postgres to exist in the same environment. We will need to invest time to create a new testing harness that could run the following scenario:
      1. Create cluster
      2. Run extensible scripts to execute arbitrary statements on this cluster
      3. Perform an upgrade by preparing, upgrading and finishing
      4. Run extensible scripts to verify all objects created by earlier scripts exists in correct form in the upgraded cluster

    Given the non trivial amount of work involved for such a suite I'd like to land this before we have 
automated testing.

On a side note; As the reviewer noticed, the tables created in the public namespace are not visible in `psql` with `\d`. The backup catalog tables have the same name as the tables in `pg_catalog`. Due to postgres internals `pg_catalog` is first in the search path and therefore the non-qualified name would alwasy resolve to `pg_catalog.pg_dist_*`. Internally this is called a non-visible table as it would resolve to a different table without a qualified name. Only visible tables are shown with `\d`.

### DIFF
--- a/src/backend/distributed/citus--8.3-1--8.4-1.sql
+++ b/src/backend/distributed/citus--8.3-1--8.4-1.sql
@@ -7,17 +7,20 @@ CREATE OR REPLACE FUNCTION pg_catalog.citus_prepare_pg_upgrade()
     SET search_path = pg_catalog
     AS $cppu$
 BEGIN
+    --
     -- backup citus catalog tables
-    CREATE TABLE public.pg_dist_authinfo AS SELECT * FROM pg_catalog.pg_dist_authinfo;
-    CREATE TABLE public.pg_dist_colocation AS SELECT * FROM pg_catalog.pg_dist_colocation;
-    CREATE TABLE public.pg_dist_local_group AS SELECT * FROM pg_catalog.pg_dist_local_group;
-    CREATE TABLE public.pg_dist_node AS SELECT * FROM pg_catalog.pg_dist_node;
-    CREATE TABLE public.pg_dist_node_metadata AS SELECT * FROM pg_catalog.pg_dist_node_metadata;
+    --
     CREATE TABLE public.pg_dist_partition AS SELECT * FROM pg_catalog.pg_dist_partition;
-    CREATE TABLE public.pg_dist_placement AS SELECT * FROM pg_catalog.pg_dist_placement;
-    CREATE TABLE public.pg_dist_poolinfo AS SELECT * FROM pg_catalog.pg_dist_poolinfo;
     CREATE TABLE public.pg_dist_shard AS SELECT * FROM pg_catalog.pg_dist_shard;
+    CREATE TABLE public.pg_dist_placement AS SELECT * FROM pg_catalog.pg_dist_placement;
+    CREATE TABLE public.pg_dist_node_metadata AS SELECT * FROM pg_catalog.pg_dist_node_metadata;
+    CREATE TABLE public.pg_dist_node AS SELECT * FROM pg_catalog.pg_dist_node;
+    CREATE TABLE public.pg_dist_local_group AS SELECT * FROM pg_catalog.pg_dist_local_group;
     CREATE TABLE public.pg_dist_transaction AS SELECT * FROM pg_catalog.pg_dist_transaction;
+    CREATE TABLE public.pg_dist_colocation AS SELECT * FROM pg_catalog.pg_dist_colocation;
+    -- enterprise catalog tables
+    CREATE TABLE public.pg_dist_authinfo AS SELECT * FROM pg_catalog.pg_dist_authinfo;
+    CREATE TABLE public.pg_dist_poolinfo AS SELECT * FROM pg_catalog.pg_dist_poolinfo;
 END;
 $cppu$;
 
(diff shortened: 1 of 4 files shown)


src/backend/distributed/citus--8.3-1--8.4-1.sql
```diff
@@ -0,0 +1,95 @@
...
+
+/* bump version to 8.4-1 */
+CREATE OR REPLACE FUNCTION pg_catalog.citus_prepare_pg_upgrade()
+    RETURNS void
+    LANGUAGE plpgsql
+    SET search_path = pg_catalog
+    AS $cppu$
+BEGIN
+    -- backup citus catalog tables
+    CREATE TABLE public.pg_dist_authinfo AS SELECT * FROM pg_catalog.pg_dist_authinfo;
```
onderkalaci: To make this work on community version as well, shall we add another function, `citus_enterprise_prepare_pg_upgrade()`, which is a no-op here, but contains the accesses to enterprise-only metadata tables on the same file for citus-enterprise repo? And, call `citus_enterprise_prepare_pg_upgrade()` at the end of this function, remove pg_dist_authinfo/ pg_dist_poolinfo from this function.

src/backend/distributed/citus--8.3-1--8.4-1.sql
```diff
@@ -0,0 +1,95 @@
...
+    LANGUAGE plpgsql
+    SET search_path = pg_catalog
+    AS $cppu$
+DECLARE
+    table_name regclass;
+    command text;
+    trigger_name text;
+BEGIN
+    -- restore citus catalog tables
+    INSERT INTO pg_catalog.pg_dist_authinfo SELECT * FROM public.pg_dist_authinfo;
```
onderkalaci: Minor: I think the order or these statements doesn't matter at the moment. Still, I like the order that is on the documentation for some related tables, because if we one day add foreign keys between metadata tables, that'd be robust. 

(Also, the flow of that feels/sounds more natural for `pg_dist_partition ` -> `pg_dist_shard ` -> `pg_dist_placement ` ) 

In any case, I think it'd be nice to bucket enterprise-only tables at the end would be easier to read/maintain.

src/backend/distributed/citus--8.3-1--8.4-1.sql
```diff
@@ -0,0 +1,95 @@
...
+    INSERT INTO pg_catalog.pg_dist_node SELECT * FROM public.pg_dist_node;
+    INSERT INTO pg_catalog.pg_dist_node_metadata SELECT * FROM public.pg_dist_node_metadata;
+    INSERT INTO pg_catalog.pg_dist_partition SELECT * FROM public.pg_dist_partition;
+    INSERT INTO pg_catalog.pg_dist_placement SELECT * FROM public.pg_dist_placement;
+    INSERT INTO pg_catalog.pg_dist_poolinfo SELECT * FROM public.pg_dist_poolinfo;
+    INSERT INTO pg_catalog.pg_dist_shard SELECT * FROM public.pg_dist_shard;
+    INSERT INTO pg_catalog.pg_dist_transaction SELECT * FROM public.pg_dist_transaction;
+
+    -- drop backup tables
+    DROP TABLE public.pg_dist_authinfo;
```
onderkalaci: Do you think would it make sense to have smaller functions for each just for readability and potentially maintainability in the future?

```SQL
CREATE OR REPLACE FUNCTION pg_catalog.citus_finish_pg_upgrade()
BEGIN;
      PERFORM load_metadata_from_backup_tables();
      PERFORM drop_metadata_backup_tables();
      PERFORM reset_metadata_sequences();
      PERFORM create_truncate_triggers_for_distributed_tables();
      PERFORM set_dependency();
END;
```

I think this could end up with too many functions if we decide to have `enterprise` counterparts functions for the metadata tables.

src/backend/distributed/citus.control
```diff
@@ -1,6 +1,6 @@
 # Citus extension
 comment = 'Citus distributed database'
```
onderkalaci: minor: we typically also bump the version in `sql/multi_extension.sql` regression test. I think it might even make sense to simply call these two functions 

```SQL
ALTER EXTENSION citus UPDATE TO '8.4-1';
.. some new lines to allow other future ALTER EXTENSIONs


-- just call the functions for sanity check, 
-- showing that nothing errors out is good enough for now
SELECT citus_prepare_pg_upgrade();
SELECT citus_finish_pg_upgrade();

```

onderkalaci: Ah, this probably wouldn't work because the trigger creation and setting the dependency would fail as they are already there.

src/backend/distributed/citus--8.3-1--8.4-1.sql
marcocitus: Downside is that these would all be very confusing user-facing functions.

onderkalaci: That's also true. My intention here is to make each block easily traceable while skimming though. We could maybe simply make the comments between blocks more visible by adding `--` before and after the comment:

```SQL
 --
 -- drop backup tables
 --
```

If that makes sense to you as well.

src/backend/distributed/citus--8.3-1--8.4-1.sql
thanodnl: `pg_dist_authinfo` and `pg_dist_poolinfo` are already available in community, only used in enterprise.

IIRC there is no SQL difference between community and enterprise. The only difference is the binary. Meaning, if we only want to restore these catalogs in enterprise we need to migrate the logics to C code. Until then we need to have the backup and restore of those tables here I guess.

src/backend/distributed/citus--8.3-1--8.4-1.sql
thanodnl: Your reasoning makes sense, didn't take FK's into account. For these blocks to be easy to maintain I always sort them alphabetically. Now with your reasoning I will change the order back to the order in the docs and separate the enterprise catalogs to the bottom.

src/backend/distributed/citus--8.3-1--8.4-1.sql
thanodnl: Comments make sense, will expand.

src/backend/distributed/citus.control
thanodnl: Thanks for pointing out `sql/multi_extension.sql`, will add the bump there.

Great thinking about the invocation of the functions, I will see if and how they will fail if you call them outside of an upgrade. My guess is that you are correct and it will fail on the triggers and dependencies.

We need to come back to these changes for testing in the short future anyway. Hanefi will send some pointers next week on how he tested the upgrades. Until then we will need to keep following our old testing documents to verify it works.

src/backend/distributed/citus--8.3-1--8.4-1.sql
marcocitus: I think fixing the order makes sense. 

Small note on foreign keys: The one foreign key we added (pg_dist_poolinfo -> pg_dist_node) is causing some issues. pg_dist_poolinfo gets truncated during metadata syncing, but it's supposed to be managed by the control plane. We should probably remove this key.

src/backend/distributed/citus.control
thanodnl: calling the functions without upgrading causes even earlier failures:

```
postgres=# SELECT citus_prepare_pg_upgrade();
┌──────────────────────────┐
│ citus_prepare_pg_upgrade │
├──────────────────────────┤
│                          │
└──────────────────────────┘
(1 row)

Time: 29.438 ms
postgres=# SELECT citus_finish_pg_upgrade();
ERROR:  duplicate key value violates unique constraint "pg_dist_node_pkey"
DETAIL:  Key (nodeid)=(1) already exists.
CONTEXT:  SQL statement "INSERT INTO pg_catalog.pg_dist_node SELECT * FROM public.pg_dist_node"
PL/pgSQL function citus_finish_pg_upgrade() line 14 at SQL statement
Time: 6.338 ms
```

Don't think it is good to drop all the things we currently replicate to test these functions, rather have a good upgrade test that will be easier to maintain.

thanodnl:
I have added the note of `\d` not showing the tables to the PR description, which I will copy into the commit message of the merge.

PR #2879, opened 2019-08-09
Pg12 generatedcolumns

serprex:
DESCRIPTION: Support GENERATE ALWAYS AS STORED

Blocked by #2844

Lacks adding tests

We support using the generated column as the distribution key, but it brings up weird issues like `INSERT INTO tbl (val) VALUES (1)` complaining that we need to include the distribution key, which if we do then it'll error that you can't explicitly insert values into generated columns

### DIFF
--- a/src/test/regress/expected/pg12.out
+++ b/src/test/regress/expected/pg12.out
@@ -33,37 +33,41 @@ create table gen2 (
 );
 insert into gen1 (id, val1) values (1,4),(3,6),(5,2),(7,2);
 insert into gen2 (id, val1) values (1,4),(3,6),(5,2),(7,2);
-select * from create_distributed_table('gen1', 'id');
-ERROR:  cannot distribute relation: gen1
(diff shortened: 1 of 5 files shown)


marcocitus:
This all makes sense to me, are there any open items?

serprex:
There are not

src/test/regress/expected/pg12.out
```diff
@@ -33,37 +33,41 @@ create table gen2 (
 );
 insert into gen1 (id, val1) values (1,4),(3,6),(5,2),(7,2);
 insert into gen2 (id, val1) values (1,4),(3,6),(5,2),(7,2);
-select * from create_distributed_table('gen1', 'id');
-ERROR:  cannot distribute relation: gen1
-DETAIL:  Distributed relations must not use GENERATED ALWAYS AS (...) STORED.
-select * from create_distributed_table('gen2', 'val2');
+select create_distributed_table('gen1', 'id');
+NOTICE:  Copying data from local table...
```
marcocitus: This seems to cover the COPY changes.

Issue #2878, opened 2019-08-09
COPY FROM WHERE in pg12 isn't straight forward to support


serprex:
https://github.com/postgres/postgres/commit/31f3817402da736b29014ace470cf70aeb126ac5#diff-2593bfcd8f00a702b24d894bc081f314R863

This mangling of the expression occurs in code we bypass, preferring our own `CheckCopyPermissions` which doesn't have parseState

CopyState is an opaque pointer for us, so we can't assign `copyState->whereClause`

Actual filtering happens here: https://github.com/postgres/postgres/blob/31f3817402da736b29014ace470cf70aeb126ac5/src/backend/commands/copy.c#L2728

marcocitus:
Can we push down the WHERE clause to the worker? It's a lot less efficient, but at least correct.

pykello:
> Can we push down the WHERE clause to the worker? It's a lot less efficient, but at least correct.

I think this makes sense. In the meeting @serprex and @onderkalaci we decided to postpone implementing this to next release and just error out in 8.4.

PR #2880, opened 2019-08-09
Pg12 nocopyfromwhere

serprex:
Without this our custom `COPY FROM` implementation silently ignores the predicate

#2878 outlines the challenge of implementing support


serprex:
#2844 now contains this

PR #2881, opened 2019-08-12
Test SET LOCAL propagation when GUC is used in RLS policy

mkurz:
DESCRIPTION: Test `SET LOCAL` propagation: GUC in RLS policy filters data spread across nodes

This PR just adds a test:
We build a proof-of-concept where we set a GUC which then is used in a row-level-security policy to filter tenants. We were not 100% sure if this works in Citus when we filter for multiple tenants which are located on different workers/nodes/machines, so I just went ahead and wrote a test to make sure this works (and will so in future). So the interesing part of this test actually is:
```sql
SET LOCAL app.tenant_id TO '1,3';
SELECT * FROM co_test_table ORDER BY id, col_1;
```
which filters for tenant 1 and 3, but which are located on shards on different workers.

(I did use [the test added in #2689](https://github.com/citusdata/citus/pull/2689/files#diff-1476e43e842655c3ff614534db0bcf95) as "template" for this pull request here)

### DIFF
--- a/src/test/regress/expected/multi_real_time_transaction.out
+++ b/src/test/regress/expected/multi_real_time_transaction.out
@@ -430,6 +430,169 @@ SELECT COUNT(*) FROM test_table;
 
 COMMIT;
 RESET ROLE;
+-- Test GUC propagation of SET LOCAL in combination with a RLS policy
+-- that uses a GUC to filter tenants. Tenant data is spread across nodes.
+-- First, as a non-superuser, we'll see all rows because RLS is not in place yet.
+SET ROLE rls_user;
+SET search_path = 'multi_real_time_transaction';
+SELECT * FROM co_test_table ORDER BY id, col_1;
+ id | col_1 | col_2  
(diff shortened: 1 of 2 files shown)


mkurz:
@citus-bot I did sign the CLA.

mkurz:
Tests are not running...? Anything I can do to get this merged?

serprex:
@mkurz it's a configuration issue on our end which is preventing forks from running CI. Worst case I can pull your fork into a branch on our repo to trigger tests running. I'm currently wanting to wait on merging pg12 support & then I can test this against pg12 with minimal effort & figure out some way to merge this at that point

mkurz:
@serprex Sounds good to me. No hurry, it's not urgent. It's just another test.

mkurz:
@serprex Postgres 12 support got merged. I think you can just cherry-pick my commit into a new branch of yours and then just provide a new pull request to trigger a build (?)

PR #2883, opened 2019-08-12
Some cleanup

pykello:
Remove unnecessary code, add comments to struct fields, reduce nesting level.


src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -1316,9 +1316,6 @@ GetModifyConnections(Task *task, bool markCritical)
 		multiConnectionList = lappend(multiConnectionList, multiConnection);
 	}
 
-	/* then finish in parallel */
-	FinishConnectionListEstablishment(multiConnectionList);
```
onderkalaci: I think it is safe to remove this because we're already doing `GetPlacementListConnection()`. What I'm wondering is that why haven't we do `StartPlacementListConnection` instead of  `GetPlacementListConnection()` before your change. Do you have any clue?

PR #2884, opened 2019-08-12
Avoid multiple pg_dist_colocation records being created for reference tables

### LINKED ISSUE #2873: pg_dist_colocation doesn't maintain uniqueness on replication/shardcount/type
This is causing the last 3 test failures in #2844

```
In multi_metadata_sync I added some select * from pg_dist_colocation lines

CREATE TABLE mx_ref (col_1 int, col_2 text);
SELECT create_reference_table('mx_ref');
+WARNING:  COLO SEARCH 1 1 0
+WARNING:  CREATECOLO
+WARNING:  reftblcolo 10005 1 1 0
  create_reference_table
---------------------------------------------------------------------

 (1 row)

+select * from pg_dist_colocation;
+ colocationid | shardcount | replicationfactor | distributioncolumntype
+---------------------------------------------------------------------
+        10001 |          7 |                 1 |                     23
+        10002 |          3 |                 1 |                     23
+        10003 |          4 |                 1 |                     23
+        10004 |          1 |                 2 |                      0
+        10005 |          1 |                 1 |                      0
+(5 rows)
+
SELECT shardid, nodename, nodeport
FROM pg_dist_shard NATURAL JOIN pg_dist_shard_placement
WHERE logicalrelid='mx_ref'::regclass;
  shardid | nodename  | nodeport
---------+-----------+----------
  1310072 | localhost |    57637
(1 row)

 \c - - - :worker_1_port
SELECT shardid, nodename, nodeport
@@ -1436,10 +1487,20 @@
(1 row)

 RESET citus.shard_count;
RESET citus.shard_replication_factor;
RESET citus.replication_model;
RESET citus.multi_shard_commit_protocol;
ALTER SEQUENCE pg_catalog.pg_dist_groupid_seq RESTART :last_group_id;
ALTER SEQUENCE pg_catalog.pg_dist_node_nodeid_seq RESTART :last_node_id;
ALTER SEQUENCE pg_catalog.pg_dist_colocationid_seq RESTART :last_colocation_id;
ALTER SEQUENCE pg_catalog.pg_dist_placement_placementid_seq RESTART :last_placement_id;
+select * from pg_dist_colocation;
+ colocationid | shardcount | replicationfactor | distributioncolumntype
+---------------------------------------------------------------------
+        10001 |          7 |                 1 |                     23
+        10002 |          3 |                 1 |                     23
+        10003 |          4 |                 1 |                     23
+        10004 |          1 |                 2 |                      0
+        10005 |          1 |                 2 |                      0
+(5 rows)
+

This duplicate 1/2/0 entry between 10004/10005 stick around for the test suite, causing differences when pg11 picks 10004 & pg12 picks 10005. Both selects are happening on master_port

Same goes for multi_remove_node_reference though haven’t dug into why we’re creating 1370004/1370005 ids when 10004/10005 are available
```

This state happens here: https://github.com/citusdata/citus/blob/e88cb8335fc0c692cf514505125959804abad7ae/src/backend/distributed/utils/reference_table_utils.c#L185

In test we update the 1-shard 1-replication to 1-shard 2-replication, meanwhile such a colocation already exists

Possible approaches:
- Introduce a new function which deduplicates colocations (ie deletes duplicates & updates all references to refer to the surviving record)
- Have UpdateColocationGroupReplicationFactor check for conflict & if there’s a conflict leave the existing colocation as is while updating everything to point to existing record
- Never update. Insert-if-not-there a colocation with the desired replication factor & update references to the new record
- Have reference tables use a colocation entry with a NULL replication factor
- @onderkalaci suggests we shouldn't have reference tables use colocation

serprex:
DESCRIPTION: Prevent pg_dist_colocation from having multiple records for reference tables

master_deactivate_node is updated to decrement the replication factor
Otherwise deactivation could have create_reference_table produce a second record

UpdateColocationGroupReplicationFactor is renamed UpdateColocationGroupReplicationFactorForReferenceTables
& the implementation looks up the record based on distributioncolumntype == InvalidOid, rather than by id
Otherwise the record's replication factor fails to be maintained when there are no reference tables

One downside to this change is that there's no index for looking up by distributiontype. `pg_dist_colocation_configuration_index` could serve that purpose if the order of columns was revised

Fixes #2873

### DIFF
--- a/src/backend/distributed/master/master_metadata_utility.c
+++ b/src/backend/distributed/master/master_metadata_utility.c
@@ -1253,61 +1253,63 @@ UpdateShardPlacementState(uint64 placementId, char shardState)
 
 
 /*
- * UpdateColocationGroupReplicationFactor finds colocation group record for given
- * colocationId and updates its replication factor to given replicationFactor value.
+ * UpdateColocationGroupReplicationFactorForReferenceTables updates the
+ * replicationFactor for the pg_dist_colocation record for reference tables.
  * Since we do not cache pg_dist_colocation table, we do not need to invalidate the
  * cache after updating replication factor.
  */
 void
-UpdateColocationGroupReplicationFactor(uint32 colocationId, int replicationFactor)
+UpdateColocationGroupReplicationFactorForReferenceTables(int replicationFactor)
 {
 	Relation pgDistColocation = NULL;
 	SysScanDesc scanDescriptor = NULL;
 	ScanKeyData scanKey[1];
 	int scanKeyCount = 1;
-	bool indexOK = true;
+	bool indexOK = false;
 	HeapTuple heapTuple = NULL;
-	HeapTuple newHeapTuple = NULL;
 	TupleDesc tupleDescriptor = NULL;
 
-	Datum values[Natts_pg_dist_colocation];
-	bool isnull[Natts_pg_dist_colocation];
-	bool replace[Natts_pg_dist_colocation];
-
-	/* we first search for colocation group by its colocation id */
+	/*
+	 * All reference tables share a colocation entry with:
+	 * shardCount = 1, replicationFactor = activeWorkerCount, distributiontype = InvalidOid
+	 * Find the record based on distributiontype = InvalidOid, as this uniquely identifies the group.
+	 */
 	pgDistColocation = heap_open(DistColocationRelationId(), RowExclusiveLock);
 	tupleDescriptor = RelationGetDescr(pgDistColocation);
-	ScanKeyInit(&scanKey[0], Anum_pg_dist_colocation_colocationid, BTEqualStrategyNumber,
-				F_OIDEQ, ObjectIdGetDatum(colocationId));
+	ScanKeyInit(&scanKey[0], Anum_pg_dist_colocation_distributioncolumntype,
+				BTEqualStrategyNumber, F_OIDEQ, ObjectIdGetDatum(InvalidOid));
 
 	scanDescriptor = systable_beginscan(pgDistColocation,
-										DistColocationColocationidIndexId(), indexOK,
+										InvalidOid, indexOK,
 										NULL, scanKeyCount, scanKey);
 
 	heapTuple = systable_getnext(scanDescriptor);
-	if (!HeapTupleIsValid(heapTuple))
+	if (HeapTupleIsValid(heapTuple))
 	{
-		ereport(ERROR, (errcode(ERRCODE_UNDEFINED_OBJECT),
-						errmsg("could not find valid entry for colocation group "
-							   "%d", colocationId)));
+		/* after finding the group, update its replication factor */
+		/* if it doesn't exist, no worries, it'll be created when needed */
+		HeapTuple newHeapTuple = NULL;
+		Datum values[Natts_pg_dist_colocation];
+		bool isnull[Natts_pg_dist_colocation];
+		bool replace[Natts_pg_dist_colocation];
+
+		memset(replace, false, sizeof(replace));
+		memset(isnull, false, sizeof(isnull));
+		memset(values, 0, sizeof(values));
+
+		values[Anum_pg_dist_colocation_replicationfactor - 1] = Int32GetDatum(
+			replicationFactor);
+		replace[Anum_pg_dist_colocation_replicationfactor - 1] = true;
+
+		newHeapTuple = heap_modify_tuple(heapTuple, tupleDescriptor, values, isnull,
+										 replace);
+
+		CatalogTupleUpdate(pgDistColocation, &newHeapTuple->t_self, newHeapTuple);
+
+		CommandCounterIncrement();
+
+		heap_freetuple(newHeapTuple);
 	}
-
-	/* after we find colocation group, we update it with new values */
-	memset(replace, false, sizeof(replace));
-	memset(isnull, false, sizeof(isnull));
-	memset(values, 0, sizeof(values));
-
-	values[Anum_pg_dist_colocation_replicationfactor - 1] = Int32GetDatum(
-		replicationFactor);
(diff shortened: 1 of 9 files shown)


src/backend/distributed/master/master_metadata_utility.c
```diff
@@ -1277,37 +1277,36 @@ UpdateColocationGroupReplicationFactor(uint32 colocationId, int replicationFacto
...
 	tupleDescriptor = RelationGetDescr(pgDistColocation);
-	ScanKeyInit(&scanKey[0], Anum_pg_dist_colocation_colocationid, BTEqualStrategyNumber,
-				F_OIDEQ, ObjectIdGetDatum(colocationId));
+	ScanKeyInit(&scanKey[0], Anum_pg_dist_colocation_distributioncolumntype,
+				BTEqualStrategyNumber,
+				F_OIDEQ, ObjectIdGetDatum(InvalidOid));
 
 	scanDescriptor = systable_beginscan(pgDistColocation,
-										DistColocationColocationidIndexId(), indexOK,
+										InvalidOid, indexOK,
```
onderkalaci: Why have we changed `DistColocationColocationidIndexId ()` to  `InvalidOid ` ? That's the Oid of the index relation that Postgres uses for the systable scan. Is that intentional? If so, we should consider making `indexOK = false` as well. But, I think we should just add back `DistColocationColocationidIndexId ()`.

src/backend/distributed/master/master_metadata_utility.c
```diff
@@ -1277,37 +1277,36 @@ UpdateColocationGroupReplicationFactor(uint32 colocationId, int replicationFacto
 	/* we first search for colocation group by its colocation id */
 	pgDistColocation = heap_open(DistColocationRelationId(), RowExclusiveLock);
 	tupleDescriptor = RelationGetDescr(pgDistColocation);
-	ScanKeyInit(&scanKey[0], Anum_pg_dist_colocation_colocationid, BTEqualStrategyNumber,
-				F_OIDEQ, ObjectIdGetDatum(colocationId));
+	ScanKeyInit(&scanKey[0], Anum_pg_dist_colocation_distributioncolumntype,
```
onderkalaci: I think it might make sense to have an intermediate variable for `InvalidOid` just for readability purposes and create the opportunity to explain the hack we're doing:

```C


/* 
 * Reference tables don't have a distribution column, the distribution column type is invalid. 
  * All reference tables should have the same colocation group with distribution column type is invalid.
  *

  */
referenceTableDistributionColumnType = InvalidOid;

...

F_OIDEQ, ObjectIdGetDatum(
referenceTableDistributionColumnType));

```

If you don't prefer to do it in the function itself, maybe add such a comment to the function comment part. My goal here is to explicitly mention that there is a single entry in the pg_dist_colocation for reference tables, and you can get that via ` distributioncolumntype = 0;`

src/backend/distributed/master/master_metadata_utility.c
```diff
@@ -1277,37 +1277,36 @@ UpdateColocationGroupReplicationFactor(uint32 colocationId, int replicationFacto
...
-		ereport(ERROR, (errcode(ERRCODE_UNDEFINED_OBJECT),
-						errmsg("could not find valid entry for colocation group "
-							   "%d", colocationId)));
-	}
-
-	/* after we find colocation group, we update it with new values */
-	memset(replace, false, sizeof(replace));
-	memset(isnull, false, sizeof(isnull));
-	memset(values, 0, sizeof(values));
+		/* after we find colocation group, we update it with new values */
```
onderkalaci: Minor: We could move declarations of `values/isnull/replace` and `newHeapTuple` inside the if statement.

src/backend/distributed/utils/reference_table_utils.c
```diff
@@ -133,56 +133,48 @@ ReplicateAllReferenceTablesToNode(char *nodeName, int nodePort)
...
-	 * We sort the reference table list to prevent deadlocks in concurrent
-	 * ReplicateAllReferenceTablesToAllNodes calls.
-	 */
-	referenceTableList = SortList(referenceTableList, CompareOids);
-	foreach(referenceTableCell, referenceTableList)
-	{
-		Oid referenceTableId = lfirst_oid(referenceTableCell);
-		List *shardIntervalList = LoadShardIntervalList(referenceTableId);
-		ShardInterval *shardInterval = (ShardInterval *) linitial(shardIntervalList);
+		/*
```
onderkalaci: Minor: we could move declarations of `referenceTableList/referenceShardIntervalList/referenceShardIntervalCell/referenceTableCell` in to the if block.

src/test/regress/expected/multi_remove_node_reference_table.out
```diff
@@ -141,7 +141,7 @@ WHERE colocationid IN
      WHERE logicalrelid = 'remove_node_reference_table'::regclass);
```
onderkalaci: Shall we add one test to `multi_metadata_sync ` so that we won't get into such weird cases again? We already know the exact point where to add the test, so should be easy.

```SQL
-- THIS IS THE SECOND TIME WE CREATE mx_ref
 CREATE TABLE mx_ref (col_1 int, col_2 text);
 SELECT create_reference_table('mx_ref');

-- make sure that adding/removing nodes wouldn't end up
-- with multiple colocation entries for reference tables
SELECT count(*) FROM pg_dist_colocation WHERE distributioncolumntype = 0;
 count 
-------
     1
(1 row)
```

src/backend/distributed/master/master_metadata_utility.c
serprex: Because I changed the search to be based on a column which is not indexed

onderkalaci: Ok, makes sense. Let's set `indexOK  = false` in that case.

PR #2886, opened 2019-08-13
Attempt to improve slow start algorithm

onderkalaci:
The idea is that if the task executions take long enough, we could speed up the slow start algorithm. The rationale is that we'll need all the connections anyway, so there is no need to wait while building up all the connections.

In practice, I really couldn't see any performance benefits. Irrespective of this change, setting  `citus.executor_slow_start_interval TO 0`  doesn't help much either. Not the when `citus.executor_slow_start_interval TO 0`, we're already by-passing the slow start algorithm at all.

Some tests results for the query `select count(*), pg_sleep(0.2) from users_table;` where we ensure that each task takes at least 200 msecs.

master + citus.executor_slow_start_interval TO 0;    `315ms`
master + citus.executor_slow_start_interval TO 10;    `322ms`
patch: `320ms`

I even tried with 128 shards, but still the improvements are in the noise level.

I'm guessing that postgres is somehow responding slower when Citus opens too many connection at the same time. 

So my conclusion is that this PR requires more investigation. @marcocitus and @pykello, given that we're short in-time for other tasks, I'm planning to leave this as a draft PR, and potentially come back in the next release. Any thoughts/preference?

### DIFF
--- a/src/backend/distributed/executor/adaptive_executor.c
+++ b/src/backend/distributed/executor/adaptive_executor.c
@@ -1764,10 +1764,10 @@ RunDistributedExecution(DistributedExecution *execution)
 	PG_END_TRY();
 }
 
+
 static void
 SpeedUpSlowStartIfNecessary(DistributedExecution *execution)
 {
-
 	/*
 	 * - No tasks has finished yet
 	 * - The slow start interval has not been already adjusted (or already disable by the user)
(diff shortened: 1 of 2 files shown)


pykello:
@onderkalaci I think postponing this to next release makes sense.

onderkalaci:
I'm closing this, as we're not planning to implement this anytime soon & the code is very simple anyway.

PR #2890, opened 2019-08-15
Avoid invalid array accesses to partitionFileArray

### LINKED ISSUE #2828: backend crash by user command
During testing I ran into this query that will crash the backend by any user.

```
postgres=# SELECT worker_hash_partition_table(42,1,'SELECT a FROM generate_series(1,100) AS a', 'a', 23, ARRAY[0]);
Time: 8.579 ms
!>
```

Crux is probably in the length of the array.
Low priority

serprex:
DESCRIPTION: Prevent segfault in worker_partition_protocol edgecase

Closes #2828

### DIFF
--- a/src/backend/distributed/worker/worker_partition_protocol.c
+++ b/src/backend/distributed/worker/worker_partition_protocol.c
@@ -917,8 +917,13 @@ FilterAndPartitionTable(const char *filterQuery,
 	if (SPI_processed > 0)
 	{
 		TupleDesc rowDescriptor = SPI_tuptable->tupdesc;
-		partitionColumnIndex = ColumnIndex(rowDescriptor, partitionColumnName);
 
+		if (fileCount == 0)
+		{
+			ereport(ERROR, (errmsg("no partition to read into")));
+		}
+
+		partitionColumnIndex = ColumnIndex(rowDescriptor, partitionColumnName);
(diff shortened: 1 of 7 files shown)


src/backend/distributed/worker/worker_partition_protocol.c
```diff
@@ -942,6 +942,12 @@ FilterAndPartitionTable(const char *filterQuery,
 	while (SPI_processed > 0)
 	{
 		int rowIndex = 0;
+
+		if (fileCount == 0)
```
serprex: Deferred error until here so that we don't mind a lack of shards if there's no results

serprex:
The test failure is caused because now when `intermediate_results` test runs before `multi_query_directory_cleanup` test, there's a left over job directory from having an error raised

I lean towards adding a test function to clear PG_JOB_CACHE_DIR

thanodnl:
Change all in all looks good, but I guess the test failures are due to this change (possibly by the added test).

They are super unstable now, too unstable to merge. If you think your described function is easy to implement lets go ahead with that.

serprex:
@thanodnl implemented. Not sure if it's considered a bug that job directories are getting left around when errors are raised, I think this is something that happens when calling `worker_hash_partition_table` directly, whereas usually the caller cleans up afterwards. But that isn't a regression introduced here, since a) stray directories are better than segfaults & b) some other inputs could cause errors to be raised already

PR #2891, opened 2019-08-15
Introduce citus.single_shard_commit_protocol for if users want 1PC on writes to replicas

serprex:
DESCRIPTION: Add option `citus.single_shard_commit_protocol`

### DIFF
--- a/src/backend/distributed/shared_library_init.c
+++ b/src/backend/distributed/shared_library_init.c
@@ -128,7 +128,7 @@ static const struct config_enum_entry use_secondary_nodes_options[] = {
 	{ NULL, 0, false }
 };
 
-static const struct config_enum_entry multi_shard_commit_protocol_options[] = {
+static const struct config_enum_entry shard_commit_protocol_options[] = {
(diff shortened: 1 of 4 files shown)


src/backend/distributed/shared_library_init.c
```diff
@@ -930,7 +930,22 @@ RegisterCitusConfigVariables(void)
...
 		&MultiShardCommitProtocol,
 		COMMIT_PROTOCOL_2PC,
-		multi_shard_commit_protocol_options,
+		shard_commit_protocol_options,
+		PGC_USERSET,
+		0,
+		NULL, NULL, NULL);
+
+	DefineCustomEnumVariable(
+		"citus.single_shard_commit_protocol",
```
onderkalaci: Minor: I was thinking to make it `GUC_NO_SHOW_ALL ` since this is a GUC we expect to be used very rare (and hopefully never).

If you agree with me, please mark the GUC with `GUC_NO_SHOW_ALL`.

PR #2882, opened 2019-08-12
Refactor Ensure Schema Exists to Ensure Dependecies Exists

thanodnl:
DESCRIPTION: Refactor ensure schema exists to dependency exists

Historically we only supported schema's as table dependencies to be created on the workers before a table gets distributed. This PR puts infrastructure in place to walk `pg_depend` to figure out which dependencies to create on the workers. Currently only schema's are supported as objects to create before creating a table.

We also keep track of dependencies that have been created in the cluster. When we add a new node to the cluster we use this catalog to know which objects need to be created on the worker.

Side effect of knowing which objects are already distributed is that we don't have debug messages anymore when creating schema's that are already created on the workers.

PR #665, opened 2016-07-22
Remove warnings on schema creation

byucesoy:
Since now we support schema related operations, there is no need to warn user about
schema usage.

PR #2893, opened 2019-08-16
Distribute Types to worker nodes

thanodnl:
DESCRIPTION: Distribute Types to worker nodes

Blocked by #2882, draft is against that branch

When to propagate
==============

There are two logical moments that types could be distributed to the worker nodes
 - When they get used ( just in time distribution )
 - When they get created ( proactive distribution )

The just in time distribution follows the model used by how schema's get created right before we are going to create a table in that schema, for types this would be when the table uses a type as its column.

The proactive distribution is suitable for situations where it is benificial to have the type on the worker nodes directly. They can later on be used in queries where an intermediate result gets created with a cast to this type.

Just in time creation is always the last resort, you cannot create a distributed table before the type gets created. A good example use case is; you have an existing postgres server that needs to scale out. By adding the citus extension, add some nodes to the cluster, and distribute the table. The type got created before citus existed. There was no moment where citus could have propagated the creation of a type.

Proactive is almost always a good option. Types are not resource intensive objects, there is no performance overhead of having 100's of types. If you want to use them in a query to represent an intermediate result (which happens in our test suite) they just work.

There is however a moment when proactive type distribution is not beneficial; in transactions where the type is used in a distributed table.

Lets assume the following transaction:

```sql
BEGIN;
CREATE TYPE tt1 AS (a int, b int);
CREATE TABLE t1 AS (a int PRIMARY KEY, b tt1);
SELECT create_distributed_table('t1', 'a');
\copy t1 FROM bigdata.csv
```

Types are node scoped objects; meaning the type exists once per worker. Shards however have best performance when they are created over their own connection. For the type to be visible on all connections it needs to be created and committed before we try to create the shards. Here the just in time situation is most beneficial and follows how we create schema's on the workers. Outside of a transaction block we will just use 1 connection to propagate the creation.

How propagation works
=================

Just in time
-----------

Just in time propagation hooks into the infrastructure introduced in #2882. It adds types as a supported object in `SupportedDependencyByCitus`. This will make sure that any object being distributed by citus that depends on types will now cascade into types. When types are depending them self on other objects they will get created first.

Creation later works by getting the ddl commands to create the object by its `ObjectAddress` in `GetDependencyCreateDDLCommands` which will dispatch types to `CreateTypeDDLCommandsIdempotent`.

For the correct walking of the graph we follow array types, when later asked for the ddl commands for array types we return `NIL` (empty list) which makes that the object will not be recorded as distributed, (its an internal type, dependant on the user type).

Proactive distribution
---------------------

When the user creates a type (composite or enum) we will have a hook running in `multi_ProcessUtility` after the command has been applied locally. Running after running locally makes that we already have an `ObjectAddress` for the type. This is required to mark the type as being distributed.

Keeping the type up to date
====================

For types that are recorded in `pg_dist_object` (eg. `IsObjectDistributed` returns true for the `ObjectAddress`) we will intercept the utility commands that alter the type.
 - `AlterTableStmt` with `relkind` set to `OBJECT_TYPE` encapsulate changes to the fields of a composite type.
 - `DropStmt` with removeType set to `OBJECT_TYPE` encapsulate `DROP TYPE`.
 - `AlterEnumStmt` encapsulates changes to enum values.
    Enum types can not be changed transactionally. When the execution on a worker fails a warning will be shown to the user the propagation was incomplete due to worker communication failure. An idempotent command is shown for the user to re-execute when the worker communication is fixed.

~~Keeping types up to date is done prior to applying the change to the coordinator. This is primarily important for a `DROP TYPE` command which would not be able to resolve the `ObjectAddress` of the type after it has been dropped.~~

Keeping types up to date is done via the executor. Before the statement is executed locally we create a plan on how to apply it on the workers. This plan is executed after we have applied the statement locally.

All changes to types need to be done in the same transaction for types that have already been distributed and will fail with an error if parallel queries have already been executed in the same transaction. Much like foreign keys to reference tables.

TODO:
 - [x] Propagate `RENAME TYPE` for distributed types
 - [x] ALTER TYPE XXX OWNER TO XXXX;
 - [x] ALTER TYPE name RENAME ATTRIBUTE attribute_name TO new_attribute_name [ CASCADE | RESTRICT ]
 - [x] ALTER TYPE name RENAME TO new_name
 - [x] ALTER TYPE name SET SCHEMA new_schema
 - [x] ALTER TABLE ADD COLUMN composite_type
 - [x] ALTER TYPE name SET DATA TYPE ATTRIBUTE
 - [ ] ~~CREATE RANGE TYPE~~ (depends on function propagation)


src/backend/distributed/catalog/dependency.c
```diff
@@ -29,13 +31,14 @@ static bool IsObjectAddressInList(const ObjectAddress *findAddress, List *addres
...
+							  void (*apply)(void *context, Form_pg_depend row),
 							  void *context);
-static bool follow_order_object_address(void *context, const Form_pg_depend pg_depend);
-static bool follow_get_dependencies_for_object(void *context,
-											   const Form_pg_depend pg_depend);
-static void apply_add_to_target_list(void *context, const Form_pg_depend pg_depend);
+static bool follow_order_object_address(void *context, Form_pg_depend pg_depend);
+static bool follow_get_dependencies_for_object(void *context, Form_pg_depend pg_depend);
+static void apply_add_to_target_list(void *context, Form_pg_depend pg_depend);
+static List * epxand_citus_supported_types(void *context, Form_pg_depend pg_depend);
```
hanefi: `s/epxand_citus_supported_types/expand_citus_supported_types/`

src/backend/distributed/commands/utility_hook.c
```diff
@@ -282,27 +287,50 @@ multi_ProcessUtility(PlannedStmt *pstmt,
...
-				ddlJobs = PlanDropPolicyStmt(dropStatement, queryString);
+				case OBJECT_POLICY:
+				{
+					ddlJobs = PlanDropPolicyStmt(dropStatement, queryString);
+					break;
+				}
+
+				case OBJECT_TYPE:
+				{
+					ddlJobs = PlanDropTypeStmt(dropStatement, queryString);
```
hanefi: We are missing a break statement here.

(I did not realize this when I added a new case for functions after `OBJECT_TYPE` and broke things)

src/backend/distributed/deparser/deparse_type_stmts.c
```diff
@@ -0,0 +1,427 @@
+/*-------------------------------------------------------------------------
+ *
+ * deparse_type_stmts.c
+ *	  All routines to deparse type statements.
+ *	  This file contains all entry points specific for type statement deparsing as well as
+ *	  functions that are currently only used for deparsing of the type statements.
+ *
+ *	  Functions that could move later are appendColumnDef, appendColumnDefList, etc. These
+ *	  should be reused across multimple statements and should live in their own deparse
```
hanefi: `s/multimple/multiple`

src/backend/distributed/deparser/deparse_type_stmts.c
```diff
@@ -0,0 +1,427 @@
...
+ *	  file.
+ *
+ * Copyright (c) 2019, Citus Data, Inc.
+ *
+ * $Id$
```
hanefi: `$Id$`?

src/backend/distributed/catalog/dependency.c
```diff
@@ -249,7 +287,42 @@ recurse_pg_depend(const ObjectAddress *target,
...
+			ListCell *expandedObjectAddress = NULL;
+			foreach(expandedObjectAddress, expansion)
+			{
+				ObjectAddress *expandedAddress = (ObjectAddress *) lfirst(
+					expandedObjectAddress);
+				recurse_pg_depend(expandedAddress, follow, expand, apply, context);
+
+				if (apply != NULL)
+				{
+					/*
```
onderkalaci: Can we move this logic into a function just for readability purpose? I want `recurse_pg_depend ` be as simple as the initial commit :)

src/backend/distributed/catalog/dependency.c
```diff
@@ -371,11 +440,54 @@ follow_order_object_address(void *context, const Form_pg_depend pg_depend)
...
+/*
+ * epxand_citus_supported_types base on supported types by citus we might want to expand
+ * the list of objects to visit in pg_depend.
+ *
+ * An example where we want to expand is for types. Their dependencies are not captured
+ * with an entry in pg_depend from their object address, but by the object address of the
+ * relation describing the type.
+ */
+static List *
+epxand_citus_supported_types(void *context, Form_pg_depend pg_depend)
```
onderkalaci: epxand -> expand.

Also, we don't seem to use `context`, so remove ?

src/backend/distributed/worker/worker_create_if_not_exist.c
```diff
@@ -0,0 +1,210 @@
...
+	{
+		DropStmt *dropStmtParseTree = NULL;
+
+		/*
+		 * there might be dependencies left on the worker on this type, these are not
+		 * managed by citus anyway so it should be ok to drop, thus we cascade to any such
+		 * dependencies
+		 */
+		dropStmtParseTree = drop_stmt_from_object_create(parseTree);
+		dropStmtParseTree->behavior = DROP_CASCADE;
```
hanefi: Would not this fail if `dropStmtParseTree` is `NULL`? Should we move this line after the NULL-check below?

src/backend/distributed/commands/utility_hook.c
thanodnl: Thanks, missed that one

src/backend/distributed/deparser/deparse_type_stmts.c
thanodnl: probably a left over of a copy somewhere

src/backend/distributed/catalog/dependency.c
thanodnl: I have refactored `expand` a bit to not return an `ObjectAddres *` but a `Form_pg_depend`, as if you are adding records to `pg_depend` based on the rules. This made this body easier to digest.

Do you think this maakes it easy enough again?

src/backend/distributed/catalog/dependency.c
thanodnl: `context` is not used here in this expansion. For now this is the only expansion so the context is not used at all.

given it fits as pluggable function to `recurse_pg_depend` I think it is benificial to keep its signature the same. As with `follow` and `apply` the purpose of `context` is to control its behaviour from the initial call site of `recurse_pg_depend`.

If you feel strongly about it I can remove, but for consistency I think it is beneficial to keep `context` here.

src/backend/distributed/worker/worker_create_if_not_exist.c
thanodnl: makes sense

this function however will change significantly, where it will probably not drop anymore, but instead rename it (to prevent accidental data destruction) if the type is not already the same as the type you are creating.

src/backend/distributed/commands/type.c
```diff
@@ -0,0 +1,852 @@
...
+	return NULL;
+}
+
+
+/*
+ * RecreateTypeStatement returns a parsetree for the CREATE TYPE statement to recreate the
+ * type by its oid.
+ */
+Node *
+RecreateTypeStatement(Oid typeOid)
```
thanodnl: Refactor to `const ObjectAddress *` for simpler signature matching.

src/backend/distributed/catalog/dependency.c
```diff
@@ -215,15 +278,35 @@ OrderObjectAddressListInDependencyOrder(List *objectAddressList)
...
 {
 	Relation depRel = NULL;
 	ScanKeyData key[2] = { 0 };
 	SysScanDesc depScan = NULL;
 	HeapTuple depTup = NULL;
 
+	/********************
+	* first apply on expanded entries of pg_depend
+	********************/
+	if (expand != NULL)
```
onderkalaci: As discussed, let's capture the contents of the table and concat the expanded entries in a list and iterate over it.

src/backend/distributed/commands/utility_hook.c
```diff
@@ -59,6 +60,8 @@ bool EnableDDLPropagation = true; /* ddl propagation is enabled */
 PropSetCmdBehavior PropagateSetCommands = PROPSETCMD_NONE; /* SET prop off */
 static bool shouldInvalidateForeignKeyGraph = false;
 static int activeAlterTables = 0;
+static int activeCreateExtension = 0;
```
onderkalaci: Postgres already exports `creating_extension` under `commands/extension.h`, which seems to cover both createExtension and alterExtension. Can we rely on those?

src/include/distributed/dist_catalog/namespace.h
```diff
@@ -0,0 +1,15 @@
+/* */
+
+/* Created by Nils Dijk on 2019-08-07. */
```
onderkalaci: Let's update the comment

src/backend/distributed/catalog/dependency.c
```diff
@@ -101,6 +112,52 @@ SupportedDependencyByCitus(const ObjectAddress *address)
...
+					/*
+					 * array types should be followed but not created, as they get created
+					 * by the original type.
+					 */
+					return type_is_array(address->objectId);
+				}
+
+				default:
+				{
+					/* type not supported */
```
onderkalaci: I wonder whether we should warn the user that we're not going to support the other types. For example, see the following:

```SQL
postgres=# CREATE TYPE floatrange AS RANGE (
postgres(#     subtype = float8,
postgres(#     subtype_diff = float8mi
postgres(# );
CREATE TYPE
Time: 2.742 ms
postgres=# -- we don't support
postgres=# CREATE TABLE test_table_2(a int, c floatrange);
CREATE TABLE
Time: 5.713 ms
postgres=# SELECT create_distributed_table('test_table_2', 'a');
ERROR:  type "public.floatrange" does not exist
CONTEXT:  while executing command on localhost:9702
Time: 30.242 ms
```

As a user who has distributed many types which worked well, seeing this error was confusing. I can argue that WARNING would also be confusing. For example, I remember that creating schemas  was very confusing to the user before https://github.com/citusdata/citus/pull/665. So, I cannot push for a warning either.

So, maybe one step back, is there a reason that we do not support range types?

src/backend/distributed/commands/type.c
```diff
@@ -0,0 +1,852 @@
...
+	ObjectAddress typeAddress = { 0 };
+
+	/*
+	 * by not propagating in a transaction block we allow for parallelism to be used when
+	 * this type will be used as a column in a table that will be created and distributed
+	 * in this same transaction.
+	 */
+	if (IsTransactionBlock())
+	{
+		return NIL;
```
onderkalaci: How common would the following be where the user creates a type and use it in a query: not any other table

```SQL
BEGIN;
CREATE TYPE tt10 AS (a int, b int);

select (user_id, value_1)::tt10 FROM users_table LIMIT 1;
DETAIL:  on server onderkalaci@localhost:9700
ERROR:  type "public.tt10" does not exist
```

@marcocitus any comments here?

src/backend/distributed/catalog/dependency.c
thanodnl: good question, range types rely on functions and operators which we don't support at this moment but is about to be solved as well. Then we definitely should push for range types (and operators) to be added.

I have been thinking about adding a warning when we do the dependency traversal and run into an unsupported dependency. But I think that would give too many warnings at the moment.

That said, the primary reason is that ENUM and COMPOSITE types were the primary scope for this PR, so I haven't looked deep at range types. Can you actually use PLPGSQL functions, or are these only C functions that are supported?

Looking at the docs I read this about one of the functions
```
Creating a canonical function is a bit tricky, ... To do this, you must first create a shell type
```

Which comes close to a base type. This causes circular dependencies which are not covered at the moment and will probably stay a hard problem as we need to generate the right commands in the right order:
 - shell type
 - distribute function
 - upgrade type
 Not impossible, but definitely not easy.

src/backend/distributed/commands/type.c
marcocitus: I would guess not very common since it already errors out unless you also do run_command_on_workers, in which case it still works.

src/backend/distributed/catalog/dependency.c
onderkalaci: > Can you actually use PLPGSQL functions, or are these only C functions that are supported?

It looks like we can use plpgsql functions:

```SQL
 CREATE OR REPLACE FUNCTION onder_float8mi(float, float)
 RETURNS float AS 
 $$
 BEGIN
    RETURN $1 * $2;
 END;
 $$ LANGUAGE plpgsql IMMUTABLE;

CREATE TYPE onder_floatrange AS RANGE (
   subtype = float8,
   subtype_diff = onder_float8mi
 );

```

src/backend/distributed/catalog/dependency.c
```diff
@@ -335,7 +432,7 @@ follow_get_dependencies_for_object(void *context, const Form_pg_depend pg_depend
  *    the object we will not try and distribute it.
  */
 static bool
-follow_order_object_address(void *context, const Form_pg_depend pg_depend)
+follow_order_object_address(void *context, Form_pg_depend pg_depend)
```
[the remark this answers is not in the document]
onderkalaci: I'm OK with that, just curiosity.

src/backend/distributed/commands/type.c
```diff
@@ -0,0 +1,852 @@
+/*-------------------------------------------------------------------------
+ *
+ * type.c
+ *    Commands for TYPE statements.
+ *    The following types are supported in citus
+ *     - Composite Types
+ *     - Enum Types
+ *     - Array Types
+ *
+ *    Base types are more complex and often involve c code from extensions.
```
onderkalaci: Maybe we should mention about range types here as well?

src/backend/distributed/commands/type.c
```diff
@@ -0,0 +1,852 @@
...
+#include "nodes/makefuncs.h"
+#include "parser/parse_type.h"
+#include "utils/builtins.h"
+#include "utils/fmgroids.h"
+#include "utils/lsyscache.h"
+#include "utils/regproc.h"
+#include "utils/syscache.h"
+#include "utils/typcache.h"
+
+#include "distributed/dist_catalog/distobject.h"
```
onderkalaci: Again minor: We typically inline `distributed/` headers to above with alphabetical order

src/backend/distributed/commands/type.c
```diff
@@ -0,0 +1,734 @@
...
+	EnsureDependenciesExistsOnAllNodes(&typeAddress);
+
+	/* reconstruct creation statement in a portable fashion */
+	createEnumStmtSql = deparse_create_enum_stmt(stmt);
+	ereport(DEBUG3, (errmsg("deparsed enum type statement"),
+					 errdetail("sql: %s", createEnumStmtSql)));
+	createEnumStmtSql = wrap_in_sql(CREATE_OR_REPLACE_COMMAND, createEnumStmtSql);
+
+	/* to prevent recursion with mx we disable ddl propagation */
+	/* TODO, mx expects the extension owner to be used here, this requires an alter owner statement as well */
```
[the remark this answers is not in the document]
onderkalaci: `SendCommandToWorkers` uses `CitusExtensionOwnerName()`, we should maybe go for that

src/backend/distributed/commands/type.c
```diff
@@ -0,0 +1,852 @@
...
+							   "transaction"),
+						errdetail("When creating or altering a type, Citus needs to "
+								  "perform all operations over a single connection per "
+								  "node to ensure consistency."),
+						errhint("Try re-running the transaction with "
+								"\"SET LOCAL citus.multi_shard_modify_mode TO "
+								"\'sequential\';\"")));
+	}
+
+	ereport(DEBUG1, (errmsg("switching to sequential query execution mode"),
```
onderkalaci: Minor: We should switch to sequential mode if we're inside a transaction block. If we don't do it, we see lots of DEBUG1 messages, which are actually not useful. And, also calling `SetLocalMultiShardModifyModeToSequential` unnecessarily, which is minor again.

src/backend/distributed/worker/worker_create_if_not_exist.c
```diff
@@ -0,0 +1,210 @@
...
+PG_FUNCTION_INFO_V1(worker_create_or_replace);
+PG_FUNCTION_INFO_V1(type_recreate_command);
+
+
+static bool object_from_create_exists(Node *parseTree);
+static DropStmt * drop_stmt_from_object_create(Node *createStmt);
+
+
+static bool
+object_from_create_exists(Node *parseTree)
```
onderkalaci: Document functions in this file

src/backend/distributed/catalog/dependency.c
thanodnl: lets punt range times to when we have functions supported.
most of the code is already in place now so could be a nice task for a new joiner

src/backend/distributed/catalog/dependency.c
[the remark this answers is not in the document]
thanodnl: `Form_pg_depend` is a typedef to a `FormData_pg_depend *`, the const applies therefore to the pointer and not to the content of the pointer.

`const Form_pg_depend` will be interpreted as `FormData_pg_depend const *` instead of `const FormData_pg_depend *` which was the intention. As a matter of fact my editor was warning me about it, otherwise I would not have known :)

src/backend/distributed/commands/type.c
thanodnl: Guess you mean, switch _only_ when we are in a transaction block?

Thought it was harmless to switch during a single statement, but apparently there are a lot of logs then. Will change

src/backend/distributed/commands/type.c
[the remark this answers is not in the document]
thanodnl: That could work, but I really like the idea of sending it as the current user in most situations. It makes ACL's applied and we don't have to add an extra command that changes the owner.

It could be that since we don't propagate when in xact it already works with MX as expected now. I will have a look.

onderkalaci:
>  Propagate RENAME TYPE for distributed types

Yeah, otherwise it could easily lead to shard being diverged.

onderkalaci:
>  Propagate RENAME TYPE for distributed types

Yeah, otherwise it could easily lead to shard being diverged.

src/backend/distributed/commands/type.c
```diff
@@ -0,0 +1,734 @@
...
+	/*
+	 * We swap the list of objects to remove during deparse so we need a reference back to
+	 * the old list to put back
+	 */
+	List *oldTypes = stmt->objects;
+	List *distributedTypes = NIL;
+	const char *dropStmtSql = NULL;
+	ListCell *addressCell = NULL;
+	List *distributedTypeAddresses = NIL;
+
```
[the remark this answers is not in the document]
thanodnl: Note to self; `DROP ... CASCADE` could cause changes to happen to other objects that need to be visible in the transaction. Lets keep the drop propagated.

src/backend/distributed/commands/type.c
```diff
@@ -0,0 +1,852 @@
...
+ *
+ * As types are node scoped objects there exists only 1 instance of the type used by
+ * potentially multiple shards. To make sure all shards in the transaction can interact
+ * with the type the type needs to be visible on all connections used by the transaction,
+ * meaning we can only use 1 connection per node.
+ */
+static void
+EnsureSequentialModeForTypeDDL(void)
+{
+	if (ParallelQueryExecutedInTransaction())
```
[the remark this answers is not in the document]
thanodnl: ah, that is a good insight, didn't expect this to be tied to Foreign Keys so tightly. Was thinking this to be a generic test on the current xact.

Any reason this function is tightly coupled to the FK functionality? Could we perhaps untie it?

src/backend/distributed/commands/type.c
```diff
@@ -0,0 +1,865 @@
...
+	 * temporary swap the lists of objects to delete with the distributed objects and
+	 * deparse to an executable sql statement for the workers
+	 */
+	stmt->objects = distributedTypes;
+	dropStmtSql = deparse_drop_type_stmt(stmt);
+	stmt->objects = oldTypes;
+
+	/* to prevent recursion with mx we disable ddl propagation */
+	EnsureSequentialModeForTypeDDL();
+	SendCommandToWorkersAsUser(ALL_WORKERS, DISABLE_DDL_PROPAGATION, NULL);
```
onderkalaci: One side effect of running the commands on the shards before the coordinator is that the errors comes from the worker node. For example, all DDLs first executed on the coordinator. If anything is wrong, the error is thrown. Else, we'll run the commands on the shards as well.

```SQL
CREATE TYPE drop_type_test AS (user_name text, user_id bigint);
CREATE TABLE drop_type_table (a int, b drop_type_test);
SELECT create_distributed_table('drop_type_table', 'a');

 DROP TYPE  drop_type_test;

ERROR:  cannot drop type drop_type_test because other objects depend on it
DETAIL:  column b of table drop_type_table_102623 depends on type drop_type_test
column b of table drop_type_table_102625 depends on type drop_type_test
column b of table drop_type_table_102627 depends on type drop_type_test
column b of table drop_type_table_102629 depends on type drop_type_test
column b of table drop_type_table_102631 depends on type drop_type_test
column b of table drop_type_table_102633 depends on type drop_type_test
column b of table drop_type_table_102635 depends on type drop_type_test
column b of table drop_type_table_102637 depends on type drop_type_test
column b of table drop_type_table_102639 depends on type drop_type_test
column b of table drop_type_table_102641 depends on type drop_type_test
column b of table drop_type_table_102643 depends on type drop_type_test
column b of table drop_type_table_102645 depends on type drop_type_test
column b of table drop_type_table_102647 depends on type drop_type_test
column b of table drop_type_table_102649 depends on type drop_type_test
column b of table drop_type_table_102651 depends on type drop_type_test
column b of table drop_type_table_102653 depends on type drop_type_test
HINT:  Use DROP ... CASCADE to drop the dependent objects too.
Time: 21.078 ms
```

We've this behaviour in other places as well such as `alter type test_enum add value ` (and probably others as well)


Also, the function name starts with `Plan..`, and we also execute it.

thanodnl: Yeah, would be great to actually 'plan' for the drop statements.
All other calls seem to rely on the changes be visible locally before constructing the sql to propagate as they sometimes rely on oid's to resolve correctly.

For drop we actually need to construct the sql before we execute locally as the type is not visible anymore after the drop, which causes names to not resolve correctly.

Problem I had with planning ddl jobs is that they seemed to be very shard centric, will have a new look at it to see if we can use it.

I agree that names should reflect correctly in any case.

src/backend/distributed/commands/type.c
```diff
@@ -0,0 +1,734 @@
...
+	}
+
+	EnsureCoordinator();
+
+	/* reconstruct alter statement in a portable fashion */
+	alterTypeStmtSql = deparse_alter_type_stmt(stmt);
+	ereport(DEBUG3, (errmsg("deparsed alter type statement"),
+					 errdetail("sql: %s", alterTypeStmtSql)));
+
+	/* TODO, mx expects the extension owner to be used here, this requires an alter owner statement as well */
```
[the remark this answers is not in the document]
marcocitus: Using the current user seems right because otherwise we would lock the type in a separate connection, worth adding a comment to that effect.

src/backend/distributed/commands/type.c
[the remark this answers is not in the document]
marcocitus: We should probably hurry to propagate schema grants to avoid situations where we cannot actually create the type on the worker.

src/backend/distributed/commands/type.c
```diff
@@ -0,0 +1,734 @@
...
+	if (AlterEnumIsAddValue(stmt))
+	{
+		/*
+		 * ADD VALUE can't be executed in a transaction, we will execute optimistically
+		 * and on an error we will advise to fix the issue with the worker and rerun the
+		 * query with the IF NOT EXTISTS modifier. The modifier is needed as the value
+		 * might already be added to some nodes, but not all.
+		 */
+
+		/* TODO function name is unwieldly long, and runs serially which is not nice */
```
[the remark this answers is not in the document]
marcocitus: The adaptive executor does not require a shard ID. You should be able to use ExecuteTaskList.

e.g.
```c
        targetPlacement = CitusMakeNode(ShardPlacement);
        targetPlacement->nodeName = workerNode->workerName;
        targetPlacement->nodePort = workerNode->workerPort;
        targetPlacement->groupId = workerNode->groupId;

        task = CitusMakeNode(Task);
        task->taskType = DDL_TASK;
        task->queryString = concatenatedCommands;
        task->taskPlacementList = list_make1(targetPlacement);
```

src/backend/distributed/deparser/deparse.c
```diff
@@ -0,0 +1,115 @@
...
+ * limited support. Check support before using, and add support for new statements as
+ * required.
+ *
+ * Currently supported:
+ *  - CREATE TYPE
+ *  - ALTER TYPE
+ *  - DROP TYPE
+ */
+const char *
+DeparseTreeNode(Node *stmt)
```
marcocitus: At some point we will likely want to pass a deparsing context (e.g. to store the shard ID to use when deparsing table names). Not on the critical path, but it might make sense to wire that in early.

src/backend/distributed/commands/type.c
```diff
@@ -0,0 +1,923 @@
...
+}
+
+
+/*
+ * CreateEnumStmtToDrop returns, given a CREATE TYPE ... AS ENUM statement, a
+ * corresponding statement to drop the type that is to be created. The type does not need
+ * to exists in this postgres for this function to succeed.
+ */
+DropStmt *
+CreateEnumStmtToDrop(CreateEnumStmt *stmt)
```
marcocitus: Naming gets a bit confusing here. Maybe `DropStmtForEnum`?

src/backend/distributed/deparser/deparse_type_stmts.c
```diff
@@ -0,0 +1,480 @@
...
+appendColumnDef(StringInfo str, ColumnDef *columnDef)
+{
+	Oid typeOid = LookupTypeNameOid(NULL, columnDef->typeName, false);
+	Oid collationOid = GetColumnDefCollation(NULL, columnDef, typeOid);
+
+	Assert(!columnDef->is_not_null); /* not null is not supported on composite types */
+
+	if (columnDef->colname)
+	{
+		appendStringInfo(str, "%s ", columnDef->colname);
```
pykello: Does this need `quote_identifier()`?

src/backend/distributed/deparser/deparse.c
marcocitus: Discussed offline, we can probably do these translations in the tree because utility commands have raw strings rather than OIDs.

src/backend/distributed/deparser/qualify_type_stmt.c
```diff
@@ -0,0 +1,160 @@
+/*-------------------------------------------------------------------------
+ *
+ * qualify_type_stmt.c
+ *	  Functins specialized in fully qualifying all type statements
```
marcocitus: Functions

src/backend/distributed/deparser/qualify_type_stmt.c
```diff
@@ -0,0 +1,160 @@
...
+ *
```
marcocitus: if we're not adding function comments, it would be good to give a more detailed description of what this file is for, in particular what does it mean to fully qualify a type statement?

src/backend/distributed/metadata/dependency.c
```diff
@@ -55,6 +57,7 @@ static void recurse_pg_depend(const ObjectAddress *target,
 static bool FollowAllSupportedDependencies(void *context, Form_pg_depend pg_depend);
 static bool FollowNewSupportedDependencies(void *context, Form_pg_depend pg_depend);
 static void ApplyAddToDependencyList(void *context, Form_pg_depend pg_depend);
+static List * EpxandCitusSupportedTypes(void *context, const ObjectAddress *target);
```
marcocitus: Expand

src/backend/distributed/commands/dependencies.c
```diff
@@ -230,6 +260,9 @@ ReplicateAllDependenciesToNode(const char *nodeName, int nodePort)
 		return;
 	}
 
+	/* since we are executing ddl commands lets disable propagation, primarily for mx */
+	ddlCommands = list_concat(list_make1(DISABLE_DDL_PROPAGATION), ddlCommands);
+
```
marcocitus: `ExecuteCriticalRemoteCommandList` does not seem to run things in a transactino block, so this will not work correctly through a pgbouncer. Maybe use `SendCommandListToWorkerInSingleTransaction`.

src/backend/distributed/commands/utility_hook.c
```diff
@@ -446,29 +521,57 @@ multi_ProcessUtility(PlannedStmt *pstmt,
 
 	PG_TRY();
 	{
-		if (IsA(parsetree, AlterTableStmt))
-		{
-			activeAlterTables++;
-		}
+		trackStatementDepth(parsetree, true);
```
marcocitus: can be removed I think

src/backend/distributed/commands/type.c
thanodnl: This is solved now that the executor is used

src/backend/distributed/commands/dependencies.c
thanodnl: Guess the same is true for `EnsureDependenciesExistsOnAllNodes`
It refactored the way we created schema's, but that happened in a single statement.

For our testing track we might want to consider a test suite with pg_bouncer to catch these subtleties.

marcocitus:
One more bug (replaying Onder's tests):
```sql
CREATE TYPE mood AS ENUM ('sad', 'ok', 'happy');
CREATE TABLE people_mood (user_id int, b mood);
SELECT create_distributed_table('people_mood', 'user_id');

BEGIN;
  -- should switch to sequential mode
  DROP TYPE mood CASCADE;

  -- should only see user_id
  SELECT * FROM people_mood;
  CREATE TYPE mood AS ENUM ('super_sad', 'sad', 'ok', 'happy', 'super_happy');
  ALTER TABLE people_mood ADD COLUMN mood_of_user mood;
```

src/backend/distributed/commands/type.c
```diff
@@ -0,0 +1,1366 @@
...
+
+	EnsureCoordinator();
+
+	QualifyTreeNode((Node *) stmt);
+	sql = DeparseTreeNode((Node *) stmt);
+
+	EnsureSequentialModeForTypeDDL();
+	commands = list_make3(DISABLE_DDL_PROPAGATION,
+						  (void *) sql,
+						  ENABLE_DDL_PROPAGATION);
```
marcocitus: Minor: Maybe turn this into a function.

onderkalaci:
> Keeping types up to date is done prior to applying the change to the coordinator. This is primarily important for a DROP TYPE command which would not be able to resolve the ObjectAddress of the type after it has been dropped.

Is this comment on the PR description still valid? As far as I read the code, it is happening on the coordinator first. (As far as I can see `PlanAlterEnumStmt ` is exception, which is OK)


Also, lets change the PR from Draft PR to regular one :)

src/backend/distributed/deparser/qualify_type_stmt.c
```diff
@@ -0,0 +1,194 @@
...
+#include "catalog/namespace.h"
+#include "catalog/objectaddress.h"
+#include "catalog/pg_type.h"
+#include "distributed/deparser.h"
+#include "distributed/metadata/namespace.h"
+#include "nodes/makefuncs.h"
+#include "parser/parse_type.h"
+#include "utils/lsyscache.h"
+
+static Oid
```
onderkalaci: in general, missing function comments in the file. And, why do we use Postgres function name conventions here (and some other places as well)?

I think especially these  kinds of utility functions should be named with Citus convention to cause less confusion. When I see a function name, it is mostly useful to understand whether that's a Citus or Postgres function.

src/backend/distributed/commands/type.c
```diff
@@ -0,0 +1,1366 @@
...
+	 * transaction by taking a RowShareLock on pg_dist_node, which conflicts with the
+	 * ExclusiveLock taken by master_add_node.
+	 * This guarantees that all active nodes will have the object, because they will
+	 * either get it now, or get it in master_add_node after this transaction finishes and
+	 * the pg_dist_object record becomes visible.
+	 */
+	LockRelationOid(DistNodeRelationId(), RowShareLock);
+
+	/* fully qualify before lookup and later deparsing */
+	QualifyTreeNode((Node *) stmt);
```
onderkalaci: One thing that I'm hesitant is to modify the `stmt` before going through `standardProcess_Utility()` via `QualifyTreeNode() `. That might break things.

For example, `PlanIndexStmt` works on the copy of the parsetree since it is going to modify it.
```C
			/* copy parse tree since we might scribble on it to fix the schema name */
			parsetree = copyObject(parsetree);
```

Should we also work on copies of the objects for the deparse?

src/backend/distributed/deparser/deparse.c
```diff
@@ -0,0 +1,191 @@
+
```
onderkalaci: missing file comment

src/backend/distributed/deparser/objectaddress.c
```diff
@@ -0,0 +1,180 @@
+
```
onderkalaci: Missing file comment and function comments.

I can see that functions are pretty simple/straightforward and the names are descriptive. Still, it'd be good to stick to the conventions.

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -413,8 +413,11 @@ AcquireExecutorMultiShardLocks(List *taskList)
 		 * If we are dealing with a partition we are also taking locks on parent table
 		 * to prevent deadlocks on concurrent operations on a partition and its parent.
 		 */
-		LockParentShardResourceIfPartition(task->anchorShardId, lockMode);
-		LockShardResource(task->anchorShardId, lockMode);
+		if (task->anchorShardId != INVALID_SHARD_ID)
```
onderkalaci: The function name is `AcquireExecutorMultiShardLocks ` and we've `task->anchorShardId != INVALID_SHARD_ID`. This might be confusing and could lead to future bugs via other callers. 

Can't we add a  check in the adaptive executor `AcquireExecutorShardLocksForExecution` and return if the tasks doesn't have valid shardIds?

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -1479,11 +1481,15 @@ ExecuteModifyTasks(List *taskList, bool expectResults, ParamListInfo paramListIn
 	 * ProcessUtility, so we only need to do this for DML commands.
 	 */
 	firstTask = (Task *) linitial(taskList);
-	firstShardInterval = LoadShardInterval(firstTask->anchorShardId);
-	if (PartitionedTable(firstShardInterval->relationId) &&
-		firstTask->taskType == MODIFY_TASK)
+	if (firstTask->taskType == MODIFY_TASK)
```
onderkalaci: Do we really hit these lines? I guess this is done before we decide  to disable router executor for type creation?

src/backend/distributed/transaction/relation_access_tracking.c
```diff
@@ -439,7 +439,10 @@ RecordRelationParallelDDLAccessForTask(Task *task)
 		lastRelationId = currentRelationId;
 	}
 
-	RecordParallelDDLAccess(RelationIdForShard(task->anchorShardId));
+	if (task->anchorShardId != INVALID_SHARD_ID)
```
onderkalaci: We should have a single task, and that's why `RecordParallelRelationAccessForTaskList()` should return early. So, do we really need this change?

src/backend/distributed/worker/worker_create_or_replace.c
```diff
@@ -0,0 +1,179 @@
...
+		Node *localCreateStmt = NULL;
+		const char *localSqlStatement = NULL;
+		DropStmt *dropStmtParseTree = NULL;
+
+		localCreateStmt = CreateStmtByObjectAddress(address);
+		localSqlStatement = DeparseTreeNode(localCreateStmt);
+		if (strcmp(sqlStatement, localSqlStatement) == 0)
+		{
+			/*
+			 * TODO string compare is a poor mans comparison, but calling equal on the
```
onderkalaci: Is it possible that we fail to detect this because of `QualifyTreeNode ` on the coordinator?

src/backend/distributed/worker/worker_create_or_replace.c
```diff
@@ -0,0 +1,179 @@
...
+ * create the object.
+ *
+ * Renaming has two purposes
+ *  - free the identifier for creation
+ *  - non destructive if there is data store that would be destroyed if the object was
+ *    used in a table on this node, eg. types. If the type would be dropped with a cascade
+ *    it would drop any column holding user data for this type.
+ */
+Datum
+worker_create_or_replace(PG_FUNCTION_ARGS)
```
onderkalaci: Minor: `worker_create_or_replace ` sounds too generic. Maybe `worker_create_or_replace_object() ` ? I'm OK to  keep as-is as well.

src/backend/distributed/worker/worker_create_or_replace.c
```diff
@@ -0,0 +1,179 @@
...
+			 * parseTree sent by the coordinator is the same as we would create for our
+			 * object, therefore we can omit the create statement locally and not create
+			 * the object as it already exists.
+			 *
+			 * We let the coordinator know we didn't create the object.
+			 */
+			PG_RETURN_BOOL(false);
+		}
+
+		/* TODO don't drop, instead rename as described in documentation */
```
onderkalaci: As far as I remember from in person discussions, you're working on this, right?

marcocitus: Let's punt on that until after merge to unblock others.

thanodnl: I have the test written, implementation almost, I can leave it out for now but is essential to add before 9.0

src/backend/distributed/commands/type.c
thanodnl: My choice was deliberate to change the names on the parsetree before passing it to postgres' handler. That is to make sure the command applied locally is exactly the same as the command applied locally.

If we make a mistake now it is visible on the coordinator and the coordinator+workers are in sync. If we make a mistake with qualifying there is a divergence between coordinator and workers and it is invisible on the coordinator.

src/backend/distributed/commands/type.c
```diff
@@ -0,0 +1,1366 @@
...
+/*
+ * PlanAlterEnumStmt handles ALTER TYPE ... ADD VALUE for enum based types. Planning
+ * happens before the statement has been applied locally.
+ *
+ * Since it is an alter of an existing type we actually have the ObjectAddress. This is
+ * used to check if the type is distributed, if so the alter will be executed on the
+ * workers directly to keep the types in sync accross the cluster.
+ */
+List *
+PlanAlterEnumStmt(AlterEnumStmt *stmt, const char *queryString)
```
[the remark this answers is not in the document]
thanodnl: Most of the execution happens after it has been applied locally, except for adding values to the enum. Reason being that postgres < 12 does not allow this in a transaction block.

As pg12 will start allowing it my plan was to put the direct execution part of adding enum values in an `#if pgversion < 12`. Splitting that out in plan and process is non-trivial but I can make it work

src/backend/distributed/commands/type.c
thanodnl: IMHO it will be less readable if we abstract this to a function.
Now it is very visible we disable and enable ddl propagation around the remote command to be applied. I think that will be less clear if we hide it in a function

src/backend/distributed/commands/type.c
```diff
@@ -0,0 +1,1366 @@
...
+	ddlCommands = lappend(ddlCommands, buf.data);
+
+	return ddlCommands;
+}
+
+
+/*
+ * wrap_in_sql
+ *
+ * TODO specialize info create_or_replace command
```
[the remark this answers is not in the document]
thanodnl: yeah, this function is only used to wrap a ddl command in a call to `create_or_replace`. Makes more sense to specialize the function instead of having this generic one which is very specific in how to use it.

src/backend/distributed/commands/type.c
```diff
@@ -0,0 +1,1366 @@
...
+		 */
+		return false;
+	}
+
+	/*
+	 * by not propagating in a transaction block we allow for parallelism to be used when
+	 * this type will be used as a column in a table that will be created and distributed
+	 * in this same transaction.
+	 */
+	if (IsTransactionBlock())
```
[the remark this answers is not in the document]
thanodnl: Looking at its implementation it makes sense.

src/backend/distributed/commands/type.c
```diff
@@ -0,0 +1,1366 @@
...
+ *    the extension on the worker
+ *  - During a transaction block; if types are used in a distributed table in the same
+ *    block we can only provide parallelism on the table if we do not change to sequentia
+ *    mode. Types will be propagated outside of this transaction to the workers so that
+ *    the transaction can use 1 connection per shard and fully utilize citus' parallelism
+ */
+static bool
+ShouldPropagateTypeCreate()
+{
+	if (creating_extension)
```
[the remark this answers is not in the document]
thanodnl: I think it could be valuable, but I guess we want this guc to guard more then the creation of a type.

It should also prevent the dependency traversal to types on distributing a table, adding a node etc. as well as alter type statements.

My understanding was that this guc should be able to NOOP this whole change

src/backend/distributed/deparser/qualify_type_stmt.c
thanodnl: re missing docs, this function is the only one where we can add useful documentation (will add) the other would be either very repetitive or very trivial. This is compensated by having extensive documentation on this file and even more so on `qualify.c` as qualification of a tree is an important concept.

these functions are all in a form of QualifyNameOfStmtStruct where the documentation would be
```
/*
 * QualifyNameOfStmtStruct qualifies a NameOfStmtStruct
 */
```

By not having that on there the hope is people actually look at the top of the file to get a sense of what these functions are doing.

src/backend/distributed/executor/multi_router_executor.c
thanodnl: To have this check earlier is complicated for the following reasons:
 - the list is passed, so there might be some tasks that are anchored to a shard and some that are not
 - there are multiple call sites to this, so the check on the list needs to be done in a loop, multiple times.

instead I can move the check up in this look and `continue;` the loop if there is no `anchorShardId` on the task it is currently trying to take a lock on

src/backend/distributed/executor/multi_router_executor.c
thanodnl: I hit this case with a `DDLJob` that has tasks that are not anchored to a shard.

By having the task type check earlier we prevent calling `LoadShardInterval` with `INVALID_SHARD_ID`.

src/backend/distributed/transaction/relation_access_tracking.c
thanodnl: Might be I hit this code when I was creating separate tasks per node. Then again, if this value is set to invalid there is no reason to call the function below. Any concerns having this check here?

src/backend/distributed/worker/worker_create_or_replace.c
thanodnl: `CreateStmtByObjectAddress` is supposed to fully qualify the statement.

src/backend/distributed/commands/type.c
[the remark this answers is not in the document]
onderkalaci: I'm OK with a rename of the function, we don't need to change anything.

src/backend/distributed/commands/type.c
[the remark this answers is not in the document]
onderkalaci: > My understanding was that this guc should be able to NOOP this whole change

Yes, and also we can do it after merging this, but before 9.0 with a separate PR?

src/backend/distributed/commands/type.c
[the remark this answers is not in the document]
thanodnl: already changed it around

src/backend/distributed/commands/type.c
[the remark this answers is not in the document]
thanodnl: makes sense

src/backend/distributed/worker/worker_create_or_replace.c
thanodnl: issue for tracking https://github.com/citusdata/citus/issues/2953

src/backend/distributed/commands/type.c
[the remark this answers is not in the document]
thanodnl: issue for tracking: https://github.com/citusdata/citus/issues/2954

src/backend/distributed/executor/multi_router_executor.c
thanodnl: Looking at the code again, there is a codepath below that would be skipped as well with a continue. This seems to be closely related to shards as well and I assume this would always be an empty list when not anchored to a shard.

marcocitus:
> One more bug (replaying Onder's tests):

Discussed this offline. This seems to be a somewhat fundamental problem. One improvement would be to immediately propagate CREATE TYPE when in sequential mode using the same connection as the one that did the DROP TYPE. That would solve the particular case in Onder's test. However, if the next statement is a create_distributed_table it would bring back the issue that not propagating CREATE TYPE in a transaction block tries to avoid: namely that MX will create a table as superuser, and it would not be able to see the uncommitted type.

I think adding the sequential mode clause makes sense, though would prefer to do it post-merge. For MX, the solution is ultimately to separate metadata syncing into a superuser and non-super user part, such that the table is created by non-superuser and can see uncommitted writes.

thanodnl:
The bug itself is logged in https://github.com/citusdata/citus/issues/2956

src/backend/distributed/commands/type.c
marcocitus: The problem is that these trees are sometimes cached by Postgres, which means that they are allocated in a session-level memory context. If we change them by adding pointers to memory in a transaction-level context we end up with use-after-free situations which results in a crash / undefined behaviour,

```sql
DROP TYPE IF EXISTS foo;
CREATE TYPE foo AS (x int, y int);

CREATE OR REPLACE FUNCTION public.type_crash()
RETURNS void
LANGUAGE plpgsql
AS $function$
DECLARE
BEGIN
        ALTER TYPE foo OWNER TO postgres;
END;
$function$;

SELECT type_crash();
SELECT type_crash();
psql:test.sql:15: server closed the connection unexpectedly
psql:test.sql:15: connection to server was lost
```

marcocitus: Changing the memory context here is likely not the right thing to do because we'll leak a bunch of stuff into a session-level context. Instead we should do something similar as we do in index.c:
```c
        /* ensure we copy string into proper context */
        relationContext = GetMemoryChunkContext(createIndexStatement->relation);
        namespaceName = MemoryContextStrdup(relationContext, namespaceName);
        createIndexStatement->relation->schemaname = namespaceName;
```

To achieve that, it might make sense to pass the memory context all the way down into the tree via a context struct.

marcocitus: We should be extra careful in cases where we're doing a list_make2, which also allocates memory. That memory should be allocated in the memory context of the existing list. More importantly we should free the old list to prevent a memory leak.

onderkalaci: >If we make a mistake now it is visible on the coordinator and the coordinator+workers are in sync. If we make a mistake with qualifying there is a divergence between coordinator and workers and it is invisible on the coordinator.

Is that really a possibility? I mean, can we come-up with steps to repro such thing? We seem to be not doing it for anything else. So, maybe moving `QualifyTreeNode ` to the deparsing logic, and copying it before should be good enough.

thanodnl: if caching in a session scoped context is a possibility I lean more towards Onder's suggestion of copying the stmt before scribbling. It is also a smaller change baring that we do not rely on qualified stmt's in the Process step for commands. Will recheck those and copy the trees instead.

Some statement subtrees rely on postgres make function to create the correct datastructure, eg. TypeName and retain pointers to lists passed in. Hence it will become cumbersome to keep everything in the right memory context while at the same time not leaking too much. Hence copying is probably the safest option.

onderkalaci:
> The bug itself is logged in #2956


Given that it is fixed in this issue via that, can you add a regression that Marco shared which crashes the backend? Thanks!

marcocitus:
Note that #2956 is a different bug that we won't fix in this PR, but indeed it would be good to add the CREATE TYPE in PL/pgSQL.

thanodnl:
I have added a test in the same line that Marco showed above, went for a rename of the type. This was easier to verify and showed the same behavior.

Without the patch for CREATE INDEX it crashed, now that we copy the `PlannedStmt` it doesn't crash nor leaks the schema name into the cache.

src/backend/distributed/citus--8.3-1--8.4-1.sql
```diff
@@ -42,6 +42,11 @@ CREATE OR REPLACE FUNCTION pg_catalog.master_unmark_object_distributed(classid o
 COMMENT ON FUNCTION pg_catalog.master_unmark_object_distributed(classid oid, objid oid, objsubid int)
     IS 'remove an object address from citus.pg_dist_object once the object has been deleted';
 
+CREATE OR REPLACE FUNCTION worker_create_or_replace_object(statement text)
+  RETURNS bool
+  LANGUAGE C STRICT
+  AS 'MODULE_PATHNAME', $$worker_create_or_replace_object$$;
+
```
marcocitus: Could use a COMMENT

src/backend/distributed/deparser/qualify_type_stmt.c
```diff
@@ -0,0 +1,197 @@
...
+	Assert(stmt->objectType == OBJECT_TYPE);
+
+	names = (List *) stmt->object;
+	if (list_length(names) == 1)
+	{
+		/* not qualified with schema, lookup type and its schema s*/
+		TypeName *typeName = makeTypeNameFromNameList(names);
+		Oid typeOid = LookupTypeNameOid(NULL, typeName, false);
+		Oid namespaceOid = TypeOidGetNamespaceOid(typeOid);
+		char *nspname = get_namespace_name_or_temp(namespaceOid);
```
marcocitus: These 4 lines are repeated a lot, should probably be a separate function.

src/backend/distributed/commands/dependencies.c
```diff
@@ -129,7 +141,13 @@ EnsureDependenciesExistsOnAllNodes(const ObjectAddress *target)
 	foreach(connectionCell, connections)
 	{
 		MultiConnection *connection = (MultiConnection *) lfirst(connectionCell);
+
+		MarkRemoteTransactionCritical(connection);
+		RemoteTransactionBegin(connection);
+
 		ExecuteCriticalRemoteCommandList(connection, ddlCommands);
+
+		RemoteTransactionCommit(connection);
```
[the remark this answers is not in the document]
thanodnl: I tried this, but there is check in `SendCommandListToWorkerInSingleTransaction` that triggered when used for this code path.

```
cannot open new connections after the first modification command within a transaction
```

src/backend/distributed/commands/type.c
```diff
@@ -0,0 +1,1415 @@
...
+ * potentially multiple shards. To make sure all shards in the transaction can interact
+ * with the type the type needs to be visible on all connections used by the transaction,
+ * meaning we can only use 1 connection per node.
+ */
+static void
+EnsureSequentialModeForTypeDDL(void)
+{
+	if (!IsTransactionBlock())
+	{
+		/* we do not need to switch to sequential mode if we are not in a transaction */
```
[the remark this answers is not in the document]
thanodnl: This was added after Onder noticed the `DEBUG1` message from line 1373 to be printed even when not in a xact block. It should probably change to `IsMultiStatementTransaction` now that the check for when to execute has changed as well.

src/backend/distributed/commands/utility_hook.c
```diff
@@ -607,12 +774,33 @@ multi_ProcessUtility(PlannedStmt *pstmt,
...
+		ddlJob->targetRelationId == InvalidOid)
+	{
+		/*
+		 * Some ddl jobs can only be run by the adaptive executor and not our legacy ones.
+		 *
+		 * These are tasks that are not pinned to any relation nor shards. We can execute
+		 * these very naively with a simple for loop that sends them to the target worker.
+		 */
+
+		ExecuteNodeBaseDDLCommands(ddlJob->taskList);
```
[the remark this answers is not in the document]
thanodnl: Our `check-multi` test suite is still running on the old executors. Which is a good thing since we have not removed them.

Without implementing this it would cause a big difference in output files which I like less then supporting it like this.

PR #2894, opened 2019-08-16
Fix distributed deadlock in TRUNATE

pykello:
This is to fix two issues at the same time:
1. Not to get distributed deadlock for concurrent TRUNCATE from MX nodes
2. Not hit `Assert(CheckRelationLockedByMe(...))` in postgres 12.

Getting any locks in `ProcessTruncateStatement()` (even if we release it right after acquisition)
causes distributed deadlock when doing `TRUNCATE` from MX workers. To reproduce the
deadlock before this commit:

1. Create a MX distributed table `test`.
2. `echo "TRUNCATE test;" > f.sql`
3. Run:

```
pgbench -f f.sql -p 5433 -P 1 -c4 -j4 -T 25 postgres &
pgbench -f f.sql -p 5434 -P 1 -c4 -j4 -T 25 postgres &
pgbench -f f.sql -p 5435 -P 1 -c4 -j4 -T 25 postgres &
```

You will notice that some transactions are cancelled because of distributed deadlock.

### DIFF
--- a/src/backend/distributed/utils/multi_partitioning_utils.c
+++ b/src/backend/distributed/utils/multi_partitioning_utils.c
@@ -24,6 +24,7 @@
 #include "distributed/shardinterval_utils.h"
 #include "lib/stringinfo.h"
 #include "nodes/pg_list.h"
+#include "pgstat.h"
 #include "utils/builtins.h"
 #include "utils/fmgroids.h"
 #include "utils/lsyscache.h"
@@ -32,6 +33,7 @@
 
 
 static char * PartitionBound(Oid partitionId);
+static Relation try_relation_open_nolock(Oid relationId);
 
 
 /*
@@ -69,7 +71,7 @@ PartitionedTable(Oid relationId)
 bool
 PartitionedTableNoLock(Oid relationId)
 {
-	Relation rel = try_relation_open(relationId, NoLock);
+	Relation rel = try_relation_open_nolock(relationId);
 	bool partitionedTable = false;
 
 	/* don't error out for tables that are dropped */
@@ -122,7 +124,7 @@ PartitionTable(Oid relationId)
 bool
 PartitionTableNoLock(Oid relationId)
 {
-	Relation rel = try_relation_open(relationId, NoLock);
+	Relation rel = try_relation_open_nolock(relationId);
 	bool partitionTable = false;
 
 	/* don't error out for tables that are dropped */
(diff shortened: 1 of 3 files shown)


onderkalaci:
> I tested that this doesn't deadlock TRUNCATE and it fixes the assertion failures in pg12.

Hmm, I get the distributed deadlocks on this PR, both on partitioned and non-partitioned tables. But, it turns out that master also have the deadlocks. When I revert https://github.com/citusdata/citus/commit/c582eb89c8ab1947aa8f19e1af6a09cdbd04512c, we don't see the deadlocks. (We should probably have added some tests around that)

Could you please look into that again? Thanks

pykello:
> > I tested that this doesn't deadlock TRUNCATE and it fixes the assertion failures in pg12.
> 
> Hmm, I get the distributed deadlocks on this PR, both on partitioned and non-partitioned tables. But, it turns out that master also have the deadlocks. When I revert [c582eb8](https://github.com/citusdata/citus/commit/c582eb89c8ab1947aa8f19e1af6a09cdbd04512c), we don't see the deadlocks. (We should probably have added some tests around that)
> 
> Could you please look into that again? Thanks

Sure, I'll look into this.

src/backend/distributed/utils/multi_partitioning_utils.c
```diff
@@ -140,6 +142,32 @@ PartitionTableNoLock(Oid relationId)
...
+try_relation_open_nolock(Oid relationId)
+{
+	Relation relation = NULL;
+	if (!SearchSysCacheExists1(RELOID, ObjectIdGetDatum(relationId)))
+	{
+		return NULL;
+	}
+
+	relation = RelationIdGetRelation(relationId);
+	if (!RelationIsValid(relation))
```
onderkalaci: * NB: caller should already have at least AccessShareLock on the
 * relation ID, else there are nasty race conditions.
 
`RelationIdGetRelation` has the above comment. However, Citus was not doing it properly for the last few years anyway as @pykello convinced me on a private chat.

src/backend/distributed/utils/multi_partitioning_utils.c
```diff
@@ -140,6 +142,32 @@ PartitionTableNoLock(Oid relationId)
 }
 
 
+/*
+ * try_relation_open_nolock opens a relation with given relationId without
+ * acquiring locks. PostgreSQL's try_relation_open() asserts that caller
+ * has already acquired a lock on the relation, which we don't always do.
+ */
```
onderkalaci: Maybe add a half a sentence comment saying that this should be in sync with postgres' `try_relation_open` function.. In case anyone has any issues with comes to that?

pykello: done

PR #2896, opened 2019-08-18
single_shard_commit_protocol: GUC_NO_SHOW_ALL

serprex:
After #2891 was merged @onderkalaci suggested we mark it `GUC_NO_SHOW_ALL`

### DIFF
--- a/src/backend/distributed/shared_library_init.c
+++ b/src/backend/distributed/shared_library_init.c
@@ -947,7 +947,7 @@ RegisterCitusConfigVariables(void)
 		COMMIT_PROTOCOL_2PC,
 		shard_commit_protocol_options,
 		PGC_USERSET,
-		0,
+		GUC_NO_SHOW_ALL,
 		NULL, NULL, NULL);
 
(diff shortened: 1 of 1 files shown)


src/backend/distributed/utils/citus_ruleutils.c
```diff
@@ -473,8 +473,7 @@ EnsureRelationKindSupported(Oid relationId)
 							relationKind == RELKIND_FOREIGN_TABLE;
 
 	/*
-	 * Citus doesn't support bare inherited tables (i.e., not a partition or
-	 * partitioned table)
+	 * Citus doesn't support inherited tables
```
onderkalaci: I don't think this new sentence is accurate. Partitioned tables are also classified as inherited tables, and there is also another concept of [inherited tables](https://www.postgresql.org/docs/11/tutorial-inheritance.html) which Citus doesn't support. And, this check is for that.

Also, this doesn't seem to be related to the PR, so can you remove it?

PR #2897, opened 2019-08-19
Task Tracker: fix error being copy pasted from above block

serprex:
DESCRIPTION: Fix error message when cluster fails for task tracker


src/backend/distributed/executor/multi_task_tracker_executor.c
```diff
@@ -447,7 +447,7 @@ MultiTaskTrackerExecute(Job *job)
 	}
 	else if (clusterFailed)
 	{
-		ereport(ERROR, (errmsg("failed to execute task %u", failedTaskId)));
+		ereport(ERROR, (errmsg("cluster failed")));
```
marcocitus: maybe `majority of nodes failed`

PR #2898, opened 2019-08-20
Move tuplestore setup to a helper function

JelteF:
DESCRIPTION: Move tuplestore setup to a helper function


src/backend/distributed/utils/tuplestore.c
```diff
@@ -0,0 +1,98 @@
...
+ */
+ReturnSetInfo *
+CheckTuplestoreReturn(FunctionCallInfo fcinfo, TupleDesc *tupdesc)
+{
+	/* IMPORTANT can this be undefined behaviour. Casting to ReturnSetInfo before
+	 * we checked the type with IsA?
+	 * In case the type actual type is more strictly aligned than ReturnSetInfo
+	 * it is it seems:
+	 * https://wiki.sei.cmu.edu/confluence/display/c/EXP36-C.+Do+not+cast+pointers+into+more+strictly+aligned+pointer+types */
+	ReturnSetInfo *returnSetInfo = (ReturnSetInfo *) fcinfo->resultinfo;
```
JelteF: If this is indeed a problem it's easily fixable by moving it after the check. I mainly want to know if this is indeed a possible problem, because it was done like this everywhere before.

src/backend/distributed/utils/tuplestore.c
```diff
@@ -0,0 +1,98 @@
...
+Tuplestorestate *
+SetupTuplestore(FunctionCallInfo fcinfo, TupleDesc *tupleDescriptor)
+{
+	MemoryContext perQueryContext = NULL;
+	MemoryContext oldContext = NULL;
+	Tuplestorestate *tupstore = NULL;
+	ReturnSetInfo *resultSet = CheckTuplestoreReturn(fcinfo, tupleDescriptor);
+
+	/* NOTE: Is this copy actually needed? */
+	*tupleDescriptor = CreateTupleDescCopy(*tupleDescriptor);
```
JelteF: This was done here: https://github.com/citusdata/citus/pull/2898/files#diff-77e7988c2a4f80c856387d90e08229cfL773

So I coppied it, but in all the other places it's not done. Is there a reason for this? It seems not needed to me.

src/backend/distributed/utils/tuplestore.c
marcocitus: Doing the IsA check later seems safe. We're just not allowed to look at any fields that are not in Node until we did the check.

The problem in the examples is that an "int" is 4 bytes while a char is 1 byte. Originally, the pointer was pointing at a 1 byte value, while after the cast it's pointing at 4 byte value and that changes the way the pointer is resolved.

src/backend/distributed/utils/tuplestore.c
marcocitus: seems unnecessary

src/backend/distributed/utils/tuplestore.c
JelteF: The way I read it, was that dereferencing/accessing fields on the pointer doesn't even matter. In the first example the pointer casting alone is already a making the pointers unequal. So I would assume you can still get the same problem in this case if fields further down the struct definitions of the two types don't align the same.

e.g.:
```c
struct A {
   uint64 shared,
   char a,
}
struct B {
   uint64 shared,
   int b,
}

void main() {
    struct A = struct A{
}
```

src/backend/distributed/utils/tuplestore.c
JelteF: I'll remove it then.

src/backend/distributed/utils/tuplestore.c
marcocitus: That seems kind of a silly thing for a compiler to do even for big endian architectures. Anyway, pointers to structs point to the first member of the struct, which does not change here. Postgres uses this pattern all over the place.

JelteF: Alright, I'll keep it like this since Postgres also does it like this too. Modern compilers make it a habbit of doing silly things when you hit undefined behaviour though. See this for a horrible example: https://gcc.godbolt.org/z/d9eCOh

One thing I could imagine happening is that it optimizes away the `IsA` check here, because everytime when this specific layout is created it also sets the field to `ReturnSetInfo`.

PR #2899, opened 2019-08-20
Propagate REINDEX TABLE & REINDEX INDEX

serprex:
DESCRIPTION: Propagate reindex on tables & index

I wrote this because @onderkalaci wanted `REINDEX CONCURRENTLY` to error out, & then I found that the reindex logic I assumed existed wasn't there, so I wrote what I assumed

This doesn't include logic to handle changes in pg12

Not sure what desirable behavior is for reindex schema/database/system, whether that should go on to include worker nodes

Since this is a bit late in the release cycle, I'll be creating a `reindex-error` branch (#2900) which eschews doing the right thing in favor of raising an error, which doesn't require thinking about locking scenarios

### DIFF
--- a/src/backend/distributed/commands/index.c
+++ b/src/backend/distributed/commands/index.c
@@ -19,6 +19,7 @@
 #include "catalog/index.h"
 #include "catalog/namespace.h"
 #include "catalog/pg_class.h"
+#include "commands/tablecmds.h"
 #include "distributed/citus_ruleutils.h"
 #include "distributed/commands.h"
 #include "distributed/commands/utility_hook.h"
@@ -39,8 +40,12 @@
 
 /* Local functions forward declarations for helper functions */
 static List * CreateIndexTaskList(Oid relationId, IndexStmt *indexStmt);
+static List * CreateReindexTaskList(Oid relationId, ReindexStmt *reindexStmt);
 static void RangeVarCallbackForDropIndex(const RangeVar *rel, Oid relOid, Oid oldRelOid,
 										 void *arg);
+static void RangeVarCallbackForReindexIndex(const RangeVar *rel, Oid relOid, Oid
+											oldRelOid,
+											void *arg);
 static void ErrorIfUnsupportedIndexStmt(IndexStmt *createIndexStatement);
 static void ErrorIfUnsupportedDropIndexStmt(DropStmt *dropIndexStatement);
 static List * DropIndexTaskList(Oid relationId, Oid indexId, DropStmt *dropStmt);
@@ -55,6 +60,18 @@ struct DropRelationCallbackState
 	char relkind;
 	Oid heapOid;
 	bool concurrent;
+};
+
+/*
+ * This struct defines the state for the callback for reindex statements.
+ * It is copied as it is from commands/indexcmds.c in Postgres source.
+ */
+struct ReindexIndexCallbackState
+{
+#if PG_VERSION_NUM >= 120000
+	bool concurrent;
+#endif
+	Oid locked_table_oid;
 };
 
 
@@ -101,7 +118,6 @@ PlanIndexStmt(IndexStmt *createIndexStatement, const char *createIndexCommand)
 		Relation relation = NULL;
 		Oid relationId = InvalidOid;
 		bool isDistributedRelation = false;
-		char *namespaceName = NULL;
 		LOCKMODE lockmode = ShareLock;
 		MemoryContext relationContext = NULL;
 
@@ -126,19 +142,22 @@ PlanIndexStmt(IndexStmt *createIndexStatement, const char *createIndexCommand)
 
 		isDistributedRelation = IsDistributedTable(relationId);
 
-		/*
-		 * Before we do any further processing, fix the schema name to make sure
-		 * that a (distributed) table with the same name does not appear on the
-		 * search path in front of the current schema. We do this even if the
-		 * table is not distributed, since a distributed table may appear on the
-		 * search path by the time postgres starts processing this statement.
-		 */
-		namespaceName = get_namespace_name(RelationGetNamespace(relation));
+		if (createIndexStatement->relation->schemaname == NULL)
+		{
+			/*
+			 * Before we do any further processing, fix the schema name to make sure
+			 * that a (distributed) table with the same name does not appear on the
+			 * search path in front of the current schema. We do this even if the
+			 * table is not distributed, since a distributed table may appear on the
+			 * search path by the time postgres starts processing this statement.
+			 */
+			char *namespaceName = get_namespace_name(RelationGetNamespace(relation));
 
-		/* ensure we copy string into proper context */
-		relationContext = GetMemoryChunkContext(createIndexStatement->relation);
-		namespaceName = MemoryContextStrdup(relationContext, namespaceName);
-		createIndexStatement->relation->schemaname = namespaceName;
+			/* ensure we copy string into proper context */
+			relationContext = GetMemoryChunkContext(createIndexStatement->relation);
+			createIndexStatement->relation->schemaname = MemoryContextStrdup(
+				relationContext, namespaceName);
+		}
 
 		heap_close(relation, NoLock);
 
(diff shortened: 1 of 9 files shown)


src/backend/distributed/utils/citus_ruleutils.c
```diff
@@ -692,6 +693,76 @@ deparse_shard_index_statement(IndexStmt *origStmt, Oid distrelid, int64 shardid,
...
+		{
+			appendStringInfo(buffer, "SYSTEM %s",
+							 reindexStmt->name);
+			break;
+		}
+
+		case REINDEX_OBJECT_DATABASE:
+		{
+			appendStringInfo(buffer, "DATABASE %s",
+							 reindexStmt->name);
```
marcocitus: we may need quotes here as well

src/backend/distributed/commands/index.c
```diff
@@ -168,6 +169,93 @@ PlanIndexStmt(IndexStmt *createIndexStatement, const char *createIndexCommand)
...
+	 * open the relation. To prevent race conditions with later lookups, lock the table,
+	 * and modify the rangevar to include the schema.
+	 */
+	if (ReindexStatement->relation != NULL)
+	{
+		Relation relation = NULL;
+		Oid relationId = InvalidOid;
+		bool isDistributedRelation = false;
+		char *namespaceName = NULL;
+		LOCKMODE lockmode = ShareLock;
```
marcocitus: using a ShareLock here might lead to deadlocks, since Postgres itself will take an AccessExclusiveLock, so if we go ShareLock->AccessExclusiveLock we might deadlock with a transaction that goes AccessShareLock->RowExclusiveLock (e.g. SELECT + INSERT).

It probably makes sense to copy the code from ReindexIndex.

src/backend/distributed/commands/index.c
```diff
@@ -172,65 +173,94 @@ PlanIndexStmt(IndexStmt *createIndexStatement, const char *createIndexCommand)
...
+		/*
+		 * XXX: Consider using RangeVarGetRelidExtended with a permission
+		 * checking callback. Right now we'll acquire the lock before having
+		 * checked permissions, and will only fail when executing the actual
+		 * index statements.
+		 */
+		if (ReindexStatement->kind == REINDEX_OBJECT_INDEX)
+		{
+			Oid indOid = RangeVarGetRelid(ReindexStatement->relation,
+										  NoLock, false);
```
marcocitus: Could we copy-paste `RangeVarCallbackForReindexIndex` from postgres and use `RangeVarGetRelidExtended ` here?
https://github.com/postgres/postgres/blob/master/src/backend/commands/indexcmds.c#L2329

The reason that Postgres does the permissions check so early is that it would otherwise allow arbitrary users to get in the queue for an access exclusive lock. That means that if there is a long-running SELECT (possibly initiated by the user himself), all other queries that use the index will block. To prevent arbitrary users from completely messing up the database, aggressive locks are guarded with `RangeVarGetRelidExtended`.

We don't always do this in Citus, but we should.

marcocitus: Definitely lower priority than procedure call delegation tasks, though it seems small and close so if you have some time remaining (e.g. while waiting for review or at the end of the release cycle), this would be a nice thing to merge.

serprex: Updated to use `RangeVarGetRelidExtended` for `REINDEX INDEX` & `EnsureTableOwner` for `REINDEX TABLE`

src/backend/distributed/commands/index.c
```diff
@@ -172,65 +188,108 @@ PlanIndexStmt(IndexStmt *createIndexStatement, const char *createIndexCommand)
...
 
-	if (isDistributedRelation)
-	{
-		ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
-						errmsg(
-							"REINDEX is not implemented for distributed relations")));
+		/* ensure we copy string into proper context */
+		relationContext = GetMemoryChunkContext(ReindexStatement->relation);
+		namespaceName = MemoryContextStrdup(relationContext, namespaceName);
+		ReindexStatement->relation->schemaname = namespaceName;
```
marcocitus: I think we can skip this if schemaname != NULL. (otherwise this creates a small memory leak)

src/backend/distributed/commands/index.c
```diff
@@ -172,65 +188,108 @@ PlanIndexStmt(IndexStmt *createIndexStatement, const char *createIndexCommand)
...
+											  lockmode, 0,
+											  RangeVarCallbackForReindexIndex,
+											  &state);
+			relation = index_open(indOid, NoLock);
+			relationId = IndexGetRelation(indOid, false);
+		}
+		else
+		{
+			Oid tblOid = RangeVarGetRelid(ReindexStatement->relation,
+										  NoLock, false);
```
marcocitus: I suppose this should be
```
   tblOid = RangeVarGetRelidExtended(relation, ShareLock, 0,
                                       RangeVarCallbackOwnsTable, NULL);
```

This one should be simpler since `RangeVarCallbackOwnsTable` is a public function in `commands/tablecmds.h`.

src/backend/distributed/commands/index.c
serprex: Does that also apply to `CREATE INDEX` where I copied this code from?

src/backend/distributed/commands/index.c
serprex: Had checked that `EnsureTableOwner` achieves the same thing, but updated to use `RangeVarCallbackOwnsTable` instead

src/backend/distributed/commands/index.c
marcocitus: Yes. We actually added a call to copyObject before these functions are called in utility_hook.c recently, making all the memory context switching and copying redundant.

serprex: CircleCI was segfaulting when I added the null check. I can look into this more tomorrow

serprex: I think the segfaults were caused by my not noticing that CreateIndexStatement uses `namespaceName` later on, causing it to be left `NULL`. Also noticed we were now copying `ReindexStatement` twice, so fixed that too

src/backend/distributed/commands/index.c
```diff
@@ -585,6 +695,90 @@ RangeVarCallbackForDropIndex(const RangeVar *rel, Oid relOid, Oid oldRelOid, voi
...
+	 * lock.
+	 */
+	if (relId != oldRelId && OidIsValid(oldRelId))
+	{
+		UnlockRelationOid(state->locked_table_oid, table_lockmode);
+		state->locked_table_oid = InvalidOid;
+	}
+
+	/* If the relation does not exist, there's nothing more to do. */
+	if (!OidIsValid(relId))
```
thanodnl: Just glanced over this code and noticed there are some if statements without enclosing braces `{` `}`.

https://github.com/citusdata/wiki/wiki/Development-Team-&-Process#coding-conventions

I realise now that the coding conventions are still in a google doc which not everyone might have access to.

relevant section:

```
We add curly braces after each if / else. Even if the block is only one statement:

if (queryResultList != NIL)
{
	succeededPlacementList = lappend(succeededPlacementList, 
                                       shardPlacement);
}
else
{
	failedPlacementList = lappend(failedPlacementList,
                                    shardPlacement);
}
```

Lets make sure we move this document to OneDrive.

No priority now, but I think there are a couple of action items for the team here:
 - uncrustify should catch this, there is no benefit in humans checking coding style which computers can do
 - go back through the code to make sure everything conforms, which is easily done if uncrustify has a rule for this.
 - once everyone has access to the styleguide promote reading it :)

serprex: Note that the code in question has INDENT-OFF because it's copied from postgres

marcocitus: We should add a note on copying postgres code to the style guide. We do so ad verbatim with INDENT-OFF to be able to diff later.

thanodnl: ah, makes sense

PR #2900, opened 2019-08-20
Raise an error when REINDEX TABLE or INDEX is invoked on a distributed relation

serprex:
DESCRIPTION: Raise an error when reindexing a distributed table or an index of a distributed table

Propagation is implemented by #2899, but this is more straight forward to fit into the next release

### DIFF
--- a/src/backend/distributed/commands/index.c
+++ b/src/backend/distributed/commands/index.c
@@ -169,6 +169,69 @@ PlanIndexStmt(IndexStmt *createIndexStatement, const char *createIndexCommand)
 
 
 /*
+ * ErrorIfReindexOnDistributedTable determines whether a given REINDEX
+ * involves a distributed table, & raises an error if so.
+ */
+void
+ErrorIfReindexOnDistributedTable(ReindexStmt *ReindexStatement)
+{
+	Relation relation = NULL;
+	Oid relationId = InvalidOid;
+	bool isDistributedRelation = false;
+	LOCKMODE lockmode = AccessShareLock;
+
+	/*
+	 * We first check whether a distributed relation is affected. For that, we need to
+	 * open the relation.
+	 */
+	if (ReindexStatement->relation == NULL)
+	{
+		/* ignore REINDEX SCHEMA, REINDEX SYSTEM, and REINDEX DATABASE */
+		return;
+	}
+
+	Assert(ReindexStatement->relkind == REINDEX_OBJECT_INDEX ||
+		   ReindexStatement->relkind == REINDEX_OBJECT_TABLE);
+
+	/*
+	 * XXX: Consider using RangeVarGetRelidExtended with a permission
+	 * checking callback. Right now we'll acquire the lock before having
+	 * checked permissions.
+	 */
+	if (ReindexStatement->kind == REINDEX_OBJECT_INDEX)
+	{
+		Oid indOid = RangeVarGetRelid(ReindexStatement->relation,
+									  NoLock, false);
+		relation = index_open(indOid, lockmode);
+		relationId = IndexGetRelation(indOid, false);
+	}
+	else
+	{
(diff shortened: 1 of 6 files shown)


src/backend/distributed/commands/utility_hook.c
```diff
@@ -279,6 +279,19 @@ multi_ProcessUtility(PlannedStmt *pstmt,
 			ddlJobs = PlanIndexStmt((IndexStmt *) parsetree, queryString);
 		}
 
+		if (IsA(parsetree, ReindexStmt))
+		{
+			MemoryContext oldContext = MemoryContextSwitchTo(GetMemoryChunkContext(
```
onderkalaci: 1) Why do we copy the parse tree? 

2) Also, it is a little bit confusing that `PlanReindexStmt ` returns a list. Maybe re-name it to `ErrorIfReindexOnDistributedTable()` etc and return void?

src/backend/distributed/commands/index.c
```diff
@@ -168,6 +168,67 @@ PlanIndexStmt(IndexStmt *createIndexStatement, const char *createIndexCommand)
...
+List *
+PlanReindexStmt(ReindexStmt *ReindexStatement, const char *ReindexCommand)
+{
+	List *ddlJobs = NIL;
+
+	/*
+	 * We first check whether a distributed relation is affected. For that, we need to
+	 * open the relation. To prevent race conditions with later lookups, lock the table.
+	 */
+	if (ReindexStatement->relation != NULL)
```
onderkalaci: Minor suggestion in order to decrease the indentation level:

```
if (ReindexStatement->relation == NULL)
{

   /* we don't do anything for REINDEX_OBJECT_SCHEMA/ REINDEX_OBJECT_SYSTEM/REINDEX_OBJECT_DATABASE at the moment */
     return;
}
```

In the future, we might consider erroring out if the schema/database/system has any distributed tables

src/backend/distributed/commands/index.c
```diff
@@ -168,6 +168,67 @@ PlanIndexStmt(IndexStmt *createIndexStatement, const char *createIndexCommand)
...
+	{
+		Relation relation = NULL;
+		Oid relationId = InvalidOid;
+		bool isDistributedRelation = false;
+		LOCKMODE lockmode = ShareLock;
```
onderkalaci: I think it should be OK to get `AccessShareLock` here. If Citus won't error out, we'll already go through the `standardProcess_Utility`, which will acquire the accurate locks. Here we're only checking a property of a table, which should be fine with `AccessShareLock`

src/backend/distributed/commands/index.c
serprex: We definitely should not error out if there are distributed tables. It'd only make sense if there were only distributed tables

onderkalaci: To me, it makes more sense to error out (or at least WARNING) with a clear message. Because the user expects all the tables in the schema/db/system to be reindexed once the command succeeds. 

The user could reindex the local tables one by one if needed. But we can discuss that when we need it :)

src/backend/distributed/commands/index.c
```diff
@@ -168,6 +168,69 @@ PlanIndexStmt(IndexStmt *createIndexStatement, const char *createIndexCommand)
...
+	 * We first check whether a distributed relation is affected. For that, we need to
+	 * open the relation.
+	 */
+	if (ReindexStatement->relation == NULL)
+	{
+		/* ignore REINDEX SCHEMA, REINDEX SYSTEM, and REINDEX DATABASE */
+		return;
+	}
+
+	Assert(ReindexStatement->relkind == REINDEX_OBJECT_INDEX ||
```
pykello: This fails in postgres 11. Make sure test with `--enable-cassert` before merging.

PR #2904, opened 2019-08-21
Sort load_shard_placement_array by worker name/port

pykello:
multi_distribution_metadata was failing randomly, e.g. in https://circleci.com/gh/citusdata/citus/24222.

Issue was that load_shard_placement_array() sorted placements by placement id, and different worker nodes took different placement ids from test to test.

### DIFF
--- a/src/backend/distributed/test/distribution_metadata.c
+++ b/src/backend/distributed/test/distribution_metadata.c
@@ -40,6 +40,11 @@
 #include "utils/palloc.h"
 
 
+/* forward declaration of local functions */
+static int CompareShardPlacementsByWorker(const void *leftElement,
+										  const void *rightElement);
+
+
 /* declarations for dynamic loading */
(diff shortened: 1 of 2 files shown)


serprex:
Shard placement comparison is pretty scattered. We have `CompareShardPlacementsByShardId` in `shardinterval_utils`, `CompareShardPlacements` in `multi_physical_planner`, `CompareShardPlacementsByNode` in `colocation_utils` (nb this compares nodes by nodeId), & now this adds `CompareShardPlacementsByWorker` in `multi_physical_planner` which is only used in `distribution_metadata`

`CompareShardPlacementsByWorker` should be defined in `distribution_metadata` if ignoring divergence of placement ids is the route we're taking. Otherwise `CompareShardPlacementsByNode` looks like a pretty stable choice for the test case, if an existing comparison function is preferable

pykello:
@serprex moved the the definition to `distribution_metadata`

PR #2907, opened 2019-08-23
pg_dist_colocation: distributioncolumncollation

serprex:
DESCRIPTION: Respect collations on distribution columns. Do not allow non-deterministic collations to be hash distributed

Initially started with wanting to test non deterministic collations, found out we don't properly support collations on partition columns at all. ICU's hashing changes with every version, so this rejects using non deterministic collations for hash distribution

Tried coming up with tests for https://github.com/citusdata/citus/commit/be3285828f17c8841d649361edc2aa3dd094f0cb

This causes tests to require compiling postgres with `--with-icu`

`CREATE COLLATION` is not currently being propagated. I'd like to wait for #2882 to properly implement that

Involved having tables with different collations on their distribution keys not be colocated

### DIFF
--- a/src/backend/distributed/planner/multi_physical_planner.c
+++ b/src/backend/distributed/planner/multi_physical_planner.c
@@ -144,6 +144,7 @@ static Task * QueryPushdownTaskCreate(Query *originalQuery, int shardIndex,
 									  TaskType taskType,
 									  bool modifyRequiresMasterEvaluation);
 static bool ShardIntervalsEqual(FmgrInfo *comparisonFunction,
+								Oid collation,
 								ShardInterval *firstInterval,
 								ShardInterval *secondInterval);
 static List * SqlTaskList(Job *job);
@@ -2469,15 +2470,27 @@ QueryPushdownTaskCreate(Query *originalQuery, int shardIndex,
 bool
 CoPartitionedTables(Oid firstRelationId, Oid secondRelationId)
 {
-	bool coPartitionedTables = true;
 	DistTableCacheEntry *firstTableCache = DistributedTableCacheEntry(firstRelationId);
 	DistTableCacheEntry *secondTableCache = DistributedTableCacheEntry(secondRelationId);
+
 	ShardInterval **sortedFirstIntervalArray = firstTableCache->sortedShardIntervalArray;
 	ShardInterval **sortedSecondIntervalArray =
 		secondTableCache->sortedShardIntervalArray;
 	uint32 firstListShardCount = firstTableCache->shardIntervalArrayLength;
 	uint32 secondListShardCount = secondTableCache->shardIntervalArrayLength;
 	FmgrInfo *comparisonFunction = firstTableCache->shardIntervalCompareFunction;
+
+	/* reference tables are always & only copartitioned with reference tables */
+	if (firstTableCache->partitionMethod == DISTRIBUTE_BY_NONE &&
+		secondTableCache->partitionMethod == DISTRIBUTE_BY_NONE)
+	{
+		return true;
+	}
+	else if (firstTableCache->partitionMethod == DISTRIBUTE_BY_NONE ||
+			 secondTableCache->partitionMethod == DISTRIBUTE_BY_NONE)
+	{
+		return false;
+	}
 
 	if (firstListShardCount != secondListShardCount)
 	{
@@ -2517,6 +2530,18 @@ CoPartitionedTables(Oid firstRelationId, Oid secondRelationId)
 
 
 	/*
+	 * Don't compare unequal types
+	 */
+	Oid collation = firstTableCache->partitionColumn->varcollid;
+	if (firstTableCache->partitionColumn->vartype !=
+		secondTableCache->partitionColumn->vartype ||
+		collation != secondTableCache->partitionColumn->varcollid)
+	{
+		return false;
+	}
+
+
+	/*
 	 * If not known to be colocated check if the remaining shards are
 	 * anyway. Do so by comparing the shard interval arrays that are sorted on
 	 * interval minimum values. Then it compares every shard interval in order
@@ -2529,17 +2554,17 @@ CoPartitionedTables(Oid firstRelationId, Oid secondRelationId)
 		ShardInterval *secondInterval = sortedSecondIntervalArray[intervalIndex];
 
 		bool shardIntervalsEqual = ShardIntervalsEqual(comparisonFunction,
+													   collation,
 													   firstInterval,
 													   secondInterval);
 		if (!shardIntervalsEqual || !CoPlacedShardIntervals(firstInterval,
 															secondInterval))
 		{
-			coPartitionedTables = false;
-			break;
+			return false;
 		}
 	}
 
-	return coPartitionedTables;
+	return true;
 }
 
 
@@ -2585,8 +2610,8 @@ CoPlacedShardIntervals(ShardInterval *firstInterval, ShardInterval *secondInterv
  * ShardIntervalsEqual checks if given shard intervals have equal min/max values.
  */
 static bool
-ShardIntervalsEqual(FmgrInfo *comparisonFunction, ShardInterval *firstInterval,
-					ShardInterval *secondInterval)
+ShardIntervalsEqual(FmgrInfo *comparisonFunction, Oid collation,
+					ShardInterval *firstInterval, ShardInterval *secondInterval)
 {
 	bool shardIntervalsEqual = false;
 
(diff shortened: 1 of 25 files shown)


src/test/regress/expected/pg12.out
```diff
@@ -361,8 +361,60 @@ SELECT DISTINCT y FROM test;
  25
 (1 row)
 
+-- non deterministic collations
+CREATE COLLATION case_insensitive (
+	provider = icu,
+	locale = 'und-u-ks-level2',
+	deterministic = false
+);
+\c - - - :worker_1_port
```
pykello: We can use `run_command_on_master_and_workers` defined in `multi_test_helpers.sql`.

src/test/regress/expected/pg12.out
```diff
@@ -361,8 +361,60 @@ SELECT DISTINCT y FROM test;
...
+ create_distributed_table 
+--------------------------
+ 
+(1 row)
+
+insert into col_test values
+	('aSdf','vALue', 4), ('AsDf','vAluE', 5), ('asdF','value', 6);
+select count(*)
+from col_test
+where id = 'asdf' COLLATE case_insensitive;
```
pykello: This should also use the collation for the column if we omit the `COLLATE case_insensitive` from the `SELECT`. It seems that it doesn't.

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -2462,6 +2463,7 @@ CoPartitionedTables(Oid firstRelationId, Oid secondRelationId)
 	uint32 firstListShardCount = firstTableCache->shardIntervalArrayLength;
 	uint32 secondListShardCount = secondTableCache->shardIntervalArrayLength;
 	FmgrInfo *comparisonFunction = firstTableCache->shardIntervalCompareFunction;
+	Oid collation = firstTableCache->partitionColumn->varcollid;
```
pykello: What if the tables don't have the same collation Id? I think we should return false in that case since we really don't know their ordering match? Maybe also add a test for this?

Other option is to handle this check in call sites of this function and add an assertion here.

src/include/distributed/shardinterval_utils.h
```diff
@@ -26,9 +26,21 @@ typedef struct ShardIntervalCompareFunctionCacheEntry
 	FmgrInfo *functionInfo;
 } ShardIntervalCompareFunctionCacheEntry;
 
+/*
+ * SortShardIntervalContext is the context parameter in SortShardIntervalArray
+ */
+typedef struct SortShardIntervalContext
+{
+	FmgrInfo *finfo;
```
pykello: comparisonFunction or something similar?

src/backend/distributed/citus--8.4-1--8.4-2.sql
```diff
@@ -0,0 +1,13 @@
+/* citus--8.3-1--8.4-1 */
+
+ALTER TABLE pg_catalog.pg_dist_colocation ADD distributioncolumncollation oid;
+UPDATE pg_catalog.pg_dist_colocation dc SET distributioncolumncollation = t.typcollation
```
pykello: If we were ignoring collation before, this will break upgraded clusters since we are not resharding tables when upgrading. Can you look into this? Maybe also add a comment here explaining what we are doing here is the correct decision for upgraded tables.

src/backend/distributed/utils/colocation_utils.c
```diff
@@ -439,24 +442,28 @@ CompareShardPlacementsByNode(const void *leftElement, const void *rightElement)
  * colocation id, otherwise it returns INVALID_COLOCATION_ID.
  */
 uint32
-ColocationId(int shardCount, int replicationFactor, Oid distributionColumnType)
+ColocationId(int shardCount, int replicationFactor, Oid distributionColumnType, Oid
+			 distributionColumnCollation)
```
pykello: also update the function comment.

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -1381,9 +1358,10 @@ HasOverlappingShardInterval(ShardInterval **shardIntervalArray,
 		Assert(lastShardInterval->minValueExists && lastShardInterval->maxValueExists);
 		Assert(curShardInterval->minValueExists && curShardInterval->maxValueExists);
 
-		comparisonDatum = CompareCall2(shardIntervalSortCompareFunction,
-									   lastShardInterval->maxValue,
-									   curShardInterval->minValue);
+		comparisonDatum = FunctionCall2Coll(shardIntervalSortCompareFunction,
+											DEFAULT_COLLATION_OID,
```
pykello: why are we using the `DEFAULT_COLLATION_OID` here and not the distribution column's collation?

src/backend/distributed/planner/multi_physical_planner.c
serprex: I figured it was correct since we assume `comparisonFunction` is compatible between the table caches. It doesn't appear so, so I've added a commit that checks that both columns have the same type oid & collation oid

src/backend/distributed/citus--8.4-1--8.4-2.sql
marcocitus: Does the collation affect hash distribution as well?

src/test/regress/expected/pg12.out
```diff
@@ -361,8 +361,61 @@ SELECT DISTINCT y FROM test;
...
+);
+$$);
+ run_command_on_master_and_workers 
+-----------------------------------
+ 
+(1 row)
+
+select worker_hash('asdf'::text) asdf, worker_hash('aSdf'::text) "aSdf",
+	worker_hash('asdf'::text collate case_insensitive) "Casdf",
+	worker_hash('aSdf'::text collate case_insensitive) "CaSdf";
```
marcocitus: Ah, I guess it also affects hash distribution

serprex: One thing is that this test is using a non-deterministic collation. For deterministic collations, it should not impact hash distribution. Not sure how best to implement discerning the two, so current logic will not recognize different deterministic collations as colocated

marcocitus:
This seems like the right thing to do (or a necessary evil). The collations are starting to get weaved through Postgres as well and the longer we wait the more likely it is that someone will get incorrectly distributed tables, though to avoid rocking the boat in this release cycle I'm inclined to punt until 9.1.

src/backend/distributed/citus.control
```diff
@@ -1,6 +1,6 @@
 # Citus extension
 comment = 'Citus distributed database'
-default_version = '9.1-1'
+default_version = '9.1-collate'
```
marcocitus: This should just be 9.1-1.

serprex: With unknown timelines I've been having sql scripts be `X.Y-featurename` which can then be renamed when we're closer to merging. This avoids merge conflicts

marcocitus: We've switched to adding all changes to a single upgrade script per major release.

src/backend/distributed/utils/shardinterval_utils.c
```diff
@@ -292,6 +321,7 @@ FindShardIntervalIndex(Datum searchedValue, DistTableCacheEntry *cacheEntry)
 	bool useBinarySearch = (partitionMethod != DISTRIBUTE_BY_HASH ||
 							!cacheEntry->hasUniformHashDistribution);
 	int shardIndex = INVALID_SHARD_INDEX;
+	Oid shardIntervalCollation = cacheEntry->partitionColumn->varcollid;
```
marcocitus: Here and various other places: partitionColumn may be NULL in case of reference tables.

src/test/regress/expected/pg12.out
serprex: Issue: this test now fails because ICU changes hash values between versions (Ctrl+F for 'Sort keys will change from one ICU version to another' on [docs](http://userguide.icu-project.org/collation/api))

This only matters for non deterministic collations for hashing. Postgres stores version info for collations to warn user when collation version doesn't match their ICU version: https://www.postgresql.org/docs/10/sql-altercollation.html#SQL-ALTERCOLLATION-NOTES

Ideally we'd have a UDF to go through all records & move them to their appropriate shard. Would be useful also for use cases like changing ranges of range partition tables. Not sure how much this overlaps with shard rebalancer

marcocitus:
Seems like there's a relevant test failure.

src/test/regress/expected/pg12.out
marcocitus: > I think the best option is to reject hash distribution on columns with non deterministic collations. I expect realistic use cases would want range distribution anyways

I think that's acceptable.

src/backend/distributed/citus--8.4-1--8.4-2.sql
serprex: Users should not've distributed non-default collated columns before this change, as until only recently we weren't even setting up workers to have collations on columns. It may make sense to have a check which errors upgrade if the user has a table's distribution column declared with a non default collation, just in case

src/backend/distributed/citus.control
serprex: Tried changing to `9.2-1` but tests need to be updated for new version, so that should be a different PR which is blocking this

src/test/regress/sql/pg12.sql
```diff
@@ -253,6 +253,36 @@ UPDATE test SET y = 40;
...
+	val text collate case_insensitive,
+	key int
+);
+
+insert into col_test values
+	('asdF','vaLue', 1), ('Asdf','vAlue', 2), ('asDF','valUE', 3);
+
+select create_distributed_table('col_test', 'id');
+
+insert into col_test values
```
marcocitus: we did not distribute the table yet.

serprex:
Waiting on #3196 before I add a collated range table test

PR #2917, opened 2019-08-30
Use gcc .d file support to recompile files when header files change

JelteF:
DESCRIPTION: Recompile .c files automatically when an included .h file changes

This uses the `-MMD` and `-MP` flags to create `.d` files when compiling a
`.c` file as well. These `.d` files are tiny makefiles that record all the
dependencies of the respective `.c` file. This way a `.c` file will be
recompiled whenever you change one of the `.h` files that it includes. So, the
main advantage of this is not having to run `make clean` whenever you change an
`.h` file .

These flags are supported both by GCC and Clang. I'm not sure if we support any
other compilers. If so, I'm not sure what the right way is to enable them
conditionally.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 # Global excludes across all subdirectories
+*.d
 *.o
 *.so
 *.so.[0-9]
--- a/Makefile.global.in
+++ b/Makefile.global.in
@@ -68,8 +68,8 @@ ifneq (@FLEX@,)
 endif
 
 # Add options passed to configure or computed therein, to CFLAGS/CPPFLAGS/...
-override CFLAGS += @CFLAGS@ @CITUS_CFLAGS@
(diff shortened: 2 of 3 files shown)


JelteF:
I just tried out that flag and indeed it works automatically for citus as well when that's enabled. I think it's much nicer to do this with this postgres functionality than my slightly hacky solution. Is there any place we document these useful flags?

The list I have now is this:
```
--enable-debug
--enable-depend
--enable-cassert
CFLAGS=-ggdb -Og -g3 -fno-omit-frame-pointer
--with-openssl
--with-libxml
--with-libxslt
```

Probably good to document that somewhere, e.g. CONTRIBUTING.md

marcocitus:
I don't know of any documentation that has the configure command. Maybe update the PR to add it to CONTRIBUTING.md? You might want to include --with-icu (see #2907).

pykello:
@JelteF @marcocitus  is this to be closed or to be merged?

marcocitus:
Closed I think

Issue #2397, opened 2018-09-20
FIRST() and LAST() aggregates


seamusabshere:
## What

As far as I understand it [0], citus ⭐️ **hardcodes a whitelist** 🌟  of allowed aggregates - please add 2 more, `FIRST()` and `LAST()`

```
CREATE OR REPLACE FUNCTION public.first_agg ( anyelement, anyelement )
RETURNS anyelement AS $$
        SELECT CASE WHEN $1 IS NULL THEN $2 ELSE $1 END;
$$ LANGUAGE SQL STABLE;

CREATE AGGREGATE public.first (
        sfunc    = public.first_agg,
        basetype = anyelement,
        stype    = anyelement
);

CREATE OR REPLACE FUNCTION public.last_agg ( anyelement, anyelement )
RETURNS anyelement AS $$
        SELECT $2;
$$ LANGUAGE SQL STABLE;

CREATE AGGREGATE public.last (
        sfunc    = public.last_agg,
        basetype = anyelement,
        stype    = anyelement
);
```

(thanks to Mike Rylander [1])

## Why

First of all, because you added pg_cron [2] to Citus Cloud and then listed all the other clouds that _didn't_ provide this incredibly useful system tool, so you have a history of one-upping non-citus situations :)

<img width="356" alt="screen shot 2018-09-20 at 2 33 05 pm" src="https://user-images.githubusercontent.com/15787/45839353-1ffb6500-bce2-11e8-976b-cfbbbcae854b.png">

Second, because the list of approved aggregates is hardcoded, so if you don't do it, nobody can - not even if I apply it to each of my worker nodes manually.

Third, because `FIRST()` is _faster_ and _much_ simpler to use than the alternatives that Tom Lane lists here [3]. Faster because ... it's complicated. Simpler because you don't have to come up with a complex `OVER` to use it.

## Horrible alternatives

Here's one way to implement `LAST(foo)` if you don't have it already: `(jsonb_object_agg('a', foo) ->> 'a')::text`. Gross!

## Footnotes

[0] https://github.com/citusdata/citus/blob/master/src/backend/distributed/planner/multi_logical_optimizer.c#L1702
[1] https://www.postgresql.org/message-id/b918cf3d0603311735v5dde21d4pfcd613c09d80cf57@mail.gmail.com
[2] https://github.com/citusdata/pg_cron
[3] https://www.postgresql.org/message-id/10345.1508510311%40sss.pgh.pa.us

seamusabshere:
[...]

PR #2921, opened 2019-09-02
Support unary aggregates with COMBINEFUNC & non-pseudotype STYPE

### LINKED ISSUE #120: Allow Custom Aggregates

serprex:
DESCRIPTION: Implement support for aggregates which supply combinefunc & use a non-internal transition type

Currently doesn't support more than 1 argument to aggregate. For PostGIS this impacts 3 aggregates:
- https://postgis.net/docs/manual-dev/ST_ClusterWithin.html
- https://postgis.net/docs/manual-dev/ST_AsMVT.html
- https://postgis.net/docs/manual-dev/ST_AsGeobuf.html

*NB most PostGIS aggregates use internal transition types, making them already not supported by this PR*

Still todo:
- [x] cleanup commits
- [x] fix things
- [x] general testing, as of now I've only had a successful test using `avg` & `hll_union_agg`, need to review how I'm passing oid to workers for aggregate
- [x] regression tests

Closes #120
Doesn't apply to #2397 until support for anyelement stype is implemented *(their FIRST/LAST declarations would need to include `COMBINEFUNC=public.first_agg` (or public.last_agg for `LAST`))*


marcocitus:
> there's no good solution that's going to happen here, use pg_dist_object to list which aggregates support this distribution method

would this include aggregates in extensions?

serprex:
Currently working on a branch https://github.com/citusdata/citus/commits/distribute-aggregate

Due to the nature of `pg_dist_object` being used by various other things which might distribute an aggregate that doesn't support being parallelized across workers (such as `CREATE EXTENSION` distribution) it'll likely be best to include a bool column for whether the aggregate can be parallelized

marcocitus:
Will PostGIS aggregates work out of the box?

serprex:
I haven't tested with PostGIS as of yet, but I expect it to work for single argument aggregates at least. More than one argument aggregates can use the existing C code but will need to include a method of generating the higher order aggregate aggregates (`coord_combine_agg` & `worker_partial_agg`)

hanefi:
On Citus public Slack channel, Mohamed Bakli wrote:

> Hello again, I am one of the main contributers of MobilityDB, it is a moving object extension of PostgreSQL and PostGIS. I am working now on understanding Citus from inside. Actullay our extension works with Citus but we need to extend the distributed planner to support our operators. In addition, the parallel execution of our aggregate functions. As I know Citus uses a white list of the aggregate functions and If I have a new aggregate function  and I want to make Citus support this function, which direction should I follow? I mean should I extend your code or you make major changes every release. Thank you for your patience and time.

Original message is [here](https://citus-public.slack.com/archives/C0XRHT1KJ/p1570637355024900)

serprex:
This branch has made a pivot to only distribute the execution of aggregates with `COMBINEFUNC` & an `STYPE` which is not `INTERNAL`. This allows us to punt on how the user tells us it's safe to serialize/deserialize for now. Unfortunately most (but not all) PostGIS aggregates use `INTERNAL`

serprex:
Aggregates which we could remove from being explicitly supported, as this PR would implicitly support:

1. `min`
1. `max`
1. `bit_and`
1. `bit_or`
1. `bool_and`
1. `bool_or`
1. `every`

The various concat aggregates don't implement COMBINEFUNC (as they make for inefficient parallel queries). `count` exists in this weird place where it takes 0 arguments, while this PR only supports 1 argument aggregates. Many variants of `avg` work, but `avg(int8)` & `avg(numeric)` use internal transition types. Same goes for `sum` using internal for `sum(int8)` & `sum(numeric)`

Built in aggregates we'll support which weren't previously supported:
1. `var_pop(float4)`, `var_pop(float8)`
1. `var_samp(float4)`, `var_samp(float8)`
1. `variance(float4)`, `variance(float8)`
1. `stddev_pop(float4)`, `stddev_pop(float8)`
1. `stddev_samp(float4)`, `stddev_samp(float8)`
1. `stddev(float4)`, `stddev(float8)`
1. SQL2003 binary regression aggregates: `regr_count`, `regr_sxx`, `regr_syy`, `regr_sxy`, `regr_avgx`, `regr_avgy`, `regr_r2`, `regr_slope`, `regr_intercept`, `covar_pop`, `covar_samp`, `corr`

src/backend/distributed/utils/aggregate_utils.c
```diff
@@ -0,0 +1,590 @@
...
+ * return box.agg.ffunc(box.value)
+ */
+Datum
+coord_combine_agg_ffunc(PG_FUNCTION_ARGS)
+{
+	Datum ret;
+	StypeBox *box = (StypeBox *) (PG_ARGISNULL(0) ? NULL : PG_GETARG_POINTER(0));
+	LOCAL_FCINFO(inner_fcinfo, FUNC_MAX_ARGS);
+	FmgrInfo info;
+	int inner_nargs;
```
marcocitus: Should follow style guide in terms of naming, etc.

https://docs.google.com/document/d/104XKGvh8Z_AamjtUEmbzNKgx0F6qsKyqIOJD_QRDJzI/edit?usp=sharing

marcocitus:
ST_MemCollect gave a segmentation fault in `SELECT citus.worker_partial_agg('public.st_memcollect(public.geometry)'::regprocedure, geom) AS st_astext FROM geometries_102054 geometries WHERE true`

serprex:
> I tried creating the last aggregate, but when I use it fails with:
> 
> ```sql
> postgres=# SELECT last(y) FROM test;
> ERROR:  could not determine actual result type for function declared to return type anyelement
> CONTEXT:  while executing command on localhost:9700
> SQL function "last_agg" during startup
> ```

I'm relying on STYPE to lookup text serialize/deserialize functions. I lean towards rejecting all aggregates which use a pseudotype as STYPE for phase 1

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -2942,6 +3079,44 @@ AggregateArgumentType(Aggref *aggregate)
 }
 
 
+static bool
+AggregateEnabledCustom(Oid aggregateOid)
```
marcocitus: Missing function comment

src/backend/distributed/utils/aggregate_utils.c
```diff
@@ -0,0 +1,609 @@
...
+										  typioparam, -1);
+		pfree(strInitVal);
+
+		MemoryContextSwitchTo(oldContext);
+	}
+}
+
+
+/*
+ * HandleTransition replicates logic used in nodeAgg for handling
```
marcocitus: Maybe refer to advance_transition_function

src/backend/distributed/utils/aggregate_utils.c
```diff
@@ -0,0 +1,609 @@
...
+#include "fmgr.h"
+#include "miscadmin.h"
+#include "pg_config_manual.h"
+
+PG_FUNCTION_INFO_V1(worker_partial_agg_sfunc);
+PG_FUNCTION_INFO_V1(worker_partial_agg_ffunc);
+PG_FUNCTION_INFO_V1(coord_combine_agg_sfunc);
+PG_FUNCTION_INFO_V1(coord_combine_agg_ffunc);
+
+typedef struct StypeBox
```
marcocitus: Could use a comment explaining what it is for.

src/backend/distributed/utils/aggregate_utils.c
```diff
@@ -0,0 +1,609 @@
...
+	box->valueNull = false;
+	box->valueInit = true;
+}
+
+
+/*
+ * (box, agg, ...) -> box
+ * box.agg = agg;
+ * box.value = agg.sfunc(box.value, ...);
+ * return box
```
marcocitus: Comment could use some clarification

src/backend/distributed/utils/aggregate_utils.c
```diff
@@ -0,0 +1,609 @@
...
+	HandleTransition(box, fcinfo, inner_fcinfo);
+
+	PG_RETURN_POINTER(box);
+}
+
+
+/*
+ * (box) -> text
+ * return box.agg.stype.output(box.value)
+ */
```
marcocitus: Comment could use some clarification

marcocitus:
> I wonder whether we should check for unsupported aggregates in ShouldUseSubqueryPushDown.

though that would probably lead to some weird error messages for all unsupported aggregates.

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -3004,6 +3183,54 @@ AggregateFunctionOid(const char *functionName, Oid inputType)
 }
 
 
+/*
+ * AggregateFunctionOid performs a reverse lookup on aggregate function name,
```
pykello: Comment needs to be updated.

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -3004,6 +3183,54 @@ AggregateFunctionOid(const char *functionName, Oid inputType)
...
+	procRelation = heap_open(ProcedureRelationId, AccessShareLock);
+
+	ScanKeyInit(&scanKey[0], Anum_pg_proc_proname,
+				BTEqualStrategyNumber, F_NAMEEQ, CStringGetDatum(functionName));
+
+	scanDescriptor = systable_beginscan(procRelation,
+										ProcedureNameArgsNspIndexId, true,
+										NULL, scanKeyCount, scanKey);
+
+	/* loop until we find the right function */
```
pykello: where is the "loop"?

src/backend/distributed/utils/aggregate_utils.c
```diff
@@ -0,0 +1,626 @@
...
+{
+	MemoryContext aggregateContext;
+	MemoryContext oldContext;
+	void *result;
+	if (!AggCheckCallContext(fcinfo, &aggregateContext))
+	{
+		elog(ERROR, "Aggregate function called without an aggregate context");
+	}
+
+	oldContext = MemoryContextSwitchTo(aggregateContext);
```
pykello: Maybe use `MemoryContextAlloc(aggregateContext, size)`?

src/backend/distributed/utils/aggregate_utils.c
pykello: > Should follow style guide in terms of naming, etc.
> 
> https://docs.google.com/document/d/104XKGvh8Z_AamjtUEmbzNKgx0F6qsKyqIOJD_QRDJzI/edit?usp=sharing

@serprex what was the resolution on this?

src/backend/distributed/planner/multi_logical_optimizer.c
serprex: This whole function was thrown together too hastily; it'd easily grab the wrong thing if a user created functions with colliding names. Updated to lookup with qualified name & input types

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -1506,6 +1510,9 @@ MasterAggregateExpression(Aggref *originalAggregate,
 	const Index columnLevelsUp = 0;  /* normal column */
 	const AttrNumber argumentId = 1; /* our aggregates have single arguments */
 	AggClauseCosts aggregateCosts;
+	HeapTuple aggTuple;
+	Form_pg_aggregate aggform;
+	Oid combine;
```
marcocitus: These should be declared in the block within which they are used.

src/backend/distributed/utils/aggregate_utils.c
marcocitus: Bump

PR #2927, opened 2019-09-03
ActivePrimaryNodeList: Lock DistNodeRelationId()

### LINKED ISSUE #2909: Reference table may not be replicated if created during master_add_node
Creating a reference table and master_add_node currently do not block each other, which means that reference tables created while adding a node may not be replicated to the new worker node.

Session 1:
```sql
BEGIN;
set citus.multi_shard_modify_mode to 'sequential';
```

Session 2:
```sql
SELECT master_add_node(...);
```

Session 1:
```sql
SELECT * FROM test JOIN ref ON (a = x);
ERROR:  relation "ref_102072" does not exist
```

serprex:
DESCRIPTION: Avoid race condition between `create_reference_table` & `master_add_node`

TODO:
- [x] [review all usage of `ActivePrimaryNodeList`](https://github.com/citusdata/citus/issues/2909#issuecomment-527660926) (root is `GetWorkerNodeHash` which I looked over, seems like elsewhere these kinds of race conditions don't occur)
- [x] isolation tests (there's room for more: reference table colocation id, send command to workers is a broader category that doesn't matter as much without #2928)

Closes #2909


src/backend/distributed/commands/create_distributed_table.c
```diff
@@ -301,6 +302,12 @@ create_reference_table(PG_FUNCTION_ARGS)
 	 */
 	EnsureRelationKindSupported(relationId);
 
+	/*
+	 * Take an exclusive lock on pg_dist_node to serialize pg_dist_node changes
+	 * with reference table distribution
+	 */
+	LockRelationOid(DistNodeRelationId(), ShareLock);
```
pykello: Multiple reference tables should be able to be created concurrently, so I'd use `RowExclusiveLock`?

marcocitus: ShareLock might make sense considering that it blocks all writes, but not itself.  Comment is stale.

serprex: I've updated it to `AccessShareLock` because our use is only reading from the list

src/backend/distributed/commands/create_distributed_table.c
```diff
@@ -301,6 +302,12 @@ create_reference_table(PG_FUNCTION_ARGS)
 	 */
 	EnsureRelationKindSupported(relationId);
 
+	/*
+	 * Take an exclusive lock on pg_dist_node to serialize pg_dist_node changes
+	 * with reference table distribution
+	 */
+	LockRelationOid(DistNodeRelationId(), AccessShareLock);
```
marcocitus: AccessShareLock does not seem like it would resolve the issue, since it only conflicts with AccessExclusiveLock.
https://www.postgresql.org/docs/current/explicit-locking.html

In master_add_node we take an ExclusiveLock. Here we want a lock that conflicts with ExclusiveLock, but not with itself.

src/test/regress/specs/isolation_add_node_vs_reference_table_operations.spec
```diff
@@ -110,3 +132,4 @@ permutation "s1-begin" "s1-add-second-worker" "s2-insert-to-reference-table" "s1
 permutation "s2-begin" "s2-insert-to-reference-table" "s1-add-second-worker" "s2-commit" "s2-print-content"
 permutation "s1-begin" "s1-add-second-worker" "s2-ddl-on-reference-table" "s1-commit" "s2-print-index-count"
 permutation "s2-begin" "s2-ddl-on-reference-table" "s1-add-second-worker" "s2-commit" "s2-print-index-count"
+permutation "s2-begin" "s2-create-reference-table-2" "s1-add-second-worker" "s2-commit" "s2-print-content-2" "s2-drop-reference-table-2"
```
marcocitus: Can you add another permutation where node addition comes first? I believe that's the main issue.

serprex:
Part of me thinks that the issue should be master_add_node should take an AccessExclusiveLock instead of ExclusiveLock, since then ActivePrimaryNodes can lock with AccessShare since it's essentially a select. Whereas instead we're saying master_add_node takes ExclusiveLock & anyone who doesn't want their select to be inconsistent should take a ShareLock with AccessPrimaryNodes. All that said, I'd like to make the fix be to pass a lockMode parameter to AccessPrimaryNodes, which would either be NoLock (or AccessShareLock?) or ShareLock. Or we could have AccessPrimaryNodes just always take ShareLock. Kind of went on a tangent there with this update

marcocitus:
I don't like ActiveExclusiveLock because it unnecessarily blocks even simple SELECT * FROM pg_dist_node by the user, potentially for hours while we copy reference tables.

Passing a lock mode probably makes sense. This function is used in a variety of scenarios, many of which are much more tolerant to concurrent master_add_node.

src/backend/distributed/master/master_stage_protocol.c
```diff
@@ -167,14 +168,15 @@ master_create_empty_shard(PG_FUNCTION_ARGS)
 	shardId = GetNextShardId();
 
 	/* if enough live groups, add an extra candidate node as backup */
-	{
-		uint32 primaryNodeCount = ActivePrimaryNodeCount();
+	workerNodeList = ActivePrimaryNodeList(NoLock);
```
serprex: afaict there's no need for a lock, because it should be okay for shard placements to ignore some newly added node. But it is important that `list_length(workerNodeList) == primaryNodeCount` so avoid having two separate calls to `ActivePrimaryNodeList`

src/backend/distributed/transaction/worker_transaction.c
```diff
@@ -63,7 +63,7 @@ SendCommandToWorker(char *nodeName, int32 nodePort, char *command)
 void
 SendCommandToFirstWorker(char *command)
 {
-	List *workerNodeList = ActivePrimaryNodeList();
+	List *workerNodeList = ActivePrimaryNodeList(NoLock);
```
serprex: I believe this function is used when MX nodes need to agree on something? It seems it'd be better to have a singleton table `pg_dist_mx_lock_node` referencing a nodeid. Then one node is elected as the "first node" by looking up the value of that table. Then locks can be on that table when it comes to needing a consistent elected node

serprex:
With these changes I'm still hitting a race condition:

- s1: begin; create index
- s2: start_metadata_sync_to_node
- s1: commit

New node lacks index, but pg_dist_node says it has metadata

src/test/regress/expected/isolation_add_node_vs_reference_table_operations.out
```diff
@@ -457,7 +475,95 @@ nodeport       success        result
...
+		nodeport;
+
+nodeport       success        result         
+
+57637          t              0              
+57638          t              0              
+step s2-drop-reference-table-2: 
+	DROP TABLE test_reference_table_2;
+
+ERROR:  table "test_reference_table_2" does not exist
```
marcocitus: This seems unexpected.

src/backend/distributed/metadata/metadata_sync.c
```diff
@@ -1096,7 +1096,7 @@ SchemaOwnerName(Oid objectId)
 static bool
 HasMetadataWorkers(void)
 {
-	List *workerNodeList = ActivePrimaryNodeList();
+	List *workerNodeList = ActivePrimaryNodeList(ShareLock);
```
serprex: May be best to leave as `NoLock` for now, leave issues here for #2928. `start_metadata_sync_to_node` doesn't appear to currently acquire necessary locks

src/test/regress/expected/multi_follower_dml.out
```diff
@@ -74,7 +74,8 @@ SELECT * FROM the_table ORDER BY a;
 DELETE FROM the_table;
 -- DDL is not possible
 TRUNCATE the_table;
-ERROR:  cannot execute TRUNCATE TABLE in a read-only transaction
+ERROR:  cannot acquire lock mode ShareLock on database objects while recovery is in progress
+HINT:  Only RowExclusiveLock or less can be acquired on database objects during recovery.
```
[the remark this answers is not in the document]
serprex: Fixed by making `EnsureDependenciesExistOnAllNodes` use `RowShareLock`

src/backend/distributed/metadata/metadata_sync.c
```diff
@@ -1096,7 +1096,7 @@ SchemaOwnerName(Oid objectId)
 static bool
 HasMetadataWorkers(void)
 {
-	List *workerNodeList = ActivePrimaryNodeList();
+	List *workerNodeList = ActivePrimaryNodeList(NoLock);
```
serprex: Reverted this to `NoLock` as the whole metadata story isn't currently well flushed out to call here

Issue #2915, opened 2019-08-29
Propagate metadata in master_update_node


marcocitus:
Currently `master_update_node` does not propagate workers with metadata (`start_metadata_sync_to_node` has been called), because it is meant to resolve situations in which one or more workers are down. 

We should add a new column in pg_dist_node that tracks whether the metadata is up-to-date. If we fail to propagate the change to a worker we should mark the worker as stale and try to update it from the maintenance daemon. That way, the coordinator has an authoritative view of which workers have up-to-date metadata and can consider that in routing decisions (specifically stored procedure call delegation).

marcocitus:
Note that the design space might be a little bigger here. We mainly want master_update_node to propagate in a way that tolerates multiple workers failing over simultaneously (concurrent master_update_node calls that could both fail because the old nodes are down), and in such a way that we can decide whether to use the worker or not when we do stored procedure call delegation.

marcocitus:
Closed by #2928

PR #2928, opened 2019-09-04
Propagate metadata for master_update_node

pykello:
DESCRIPTION: Make master_update_node MX compatible

Addresses https://github.com/citusdata/citus/issues/2915

Previously we were not propagating changes done by `master_update_node()` to other nodes, now we do. The common use-case for `master_update_node()` is by HA tools like pg_auto_failover. 

Ususaly the new node is in standby mode, so we cannot update the metadata in the new node. To address this, a new `pg_dist_node.metadatasynced` column is added which tracks if a metadata node's metadata tables are in sync with the coordinator node.

`master_update_node()` sets the `metadatasynced` of metadata nodes to false. Maintenance daemon checks for out-of-sync metadata nodes and tries syncing them. We don't want to wait too long after `master_update_node()` for the next sync, so `master_update_node()` calls `TriggerMetadataSync()` to trigger the maintenance daemon to do a sync asap.

It is possible that the HA tool calls multiple `master_update_node()`s around the same time, which means we have multiple stand-by nodes in the cluster, in which case maintenance daemon will eventually sync the metadata to all of the newly promoted nodes.

In the tests we needed a mean for waiting until next sync. To enable this, we use postgresqls' `LISTEN/NOTIFY` commands. The maintenance daemon does a `NOTIFY metadata_sync` after doing a metadata sync. `wait_until_metadata_sync()` does a `LISTEN metadata_sync()`, so it can get the notification.

The test files have couple of other interesting UDFs:
 - `verify_metadata(nodename, nodeport)` verifies that metadata tables in a worker have the same entries as the coordinator. Right now, it just compares the pg_dist_node and pg_dist_placement entries.
 - `mark_node_readonly` simulates a readonly node by setting `default_transaction_read_only` in that node.

Todo:
- [x] Test with pg_auto_failover

### DIFF
--- a/src/backend/distributed/metadata/metadata_sync.c
+++ b/src/backend/distributed/metadata/metadata_sync.c
@@ -38,11 +38,13 @@
 #include "distributed/multi_join_order.h"
 #include "distributed/multi_partitioning_utils.h"
 #include "distributed/pg_dist_node.h"
+#include "distributed/remote_commands.h"
 #include "distributed/worker_manager.h"
 #include "distributed/worker_transaction.h"
 #include "distributed/version_compat.h"
 #include "foreign/foreign.h"
 #include "nodes/pg_list.h"
+#include "storage/lmgr.h"
 #include "utils/builtins.h"
 #include "utils/fmgroids.h"
 #include "utils/lsyscache.h"
@@ -52,6 +54,8 @@
 
 static char * LocalGroupIdUpdateCommand(int32 groupId);
 static void MarkNodeHasMetadata(char *nodeName, int32 nodePort, bool hasMetadata);
+static void UpdateDistNodeBoolAttr(char *nodeName, int32 nodePort, int attrNum,
+								   bool value);
 static List * SequenceDDLCommandsForTable(Oid relationId);
 static void EnsureSupportedSequenceColumnType(Oid sequenceOid);
 static Oid TypeOfColumn(Oid tableId, int16 columnId);
@@ -59,18 +63,15 @@ static char * TruncateTriggerCreateCommand(Oid relationId);
 static char * SchemaOwnerName(Oid objectId);
 static bool HasMetadataWorkers(void);
 static List * DetachPartitionCommandList(void);
+static bool SyncMetadataSnapshotToNode(WorkerNode *workerNode, bool raiseOnError);
 
 PG_FUNCTION_INFO_V1(start_metadata_sync_to_node);
 PG_FUNCTION_INFO_V1(stop_metadata_sync_to_node);
 
 
 /*
- * start_metadata_sync_to_node function creates the metadata in a worker for preparing the
- * worker for accepting queries. The function first sets the localGroupId of the worker
- * so that the worker knows which tuple in pg_dist_node table represents itself. After
- * that, SQL statements for re-creating metadata of MX-eligible distributed tables are
- * sent to the worker. Finally, the hasmetadata column of the target node in pg_dist_node
- * is marked as true.
+ * start_metadata_sync_to_node function sets hasmetadata column of the given
+ * node to true, and then synchronizes the metadata on the node.
  */
 Datum
 start_metadata_sync_to_node(PG_FUNCTION_ARGS)
@@ -78,14 +79,12 @@ start_metadata_sync_to_node(PG_FUNCTION_ARGS)
 	text *nodeName = PG_GETARG_TEXT_P(0);
 	int32 nodePort = PG_GETARG_INT32(1);
 	char *nodeNameString = text_to_cstring(nodeName);
-	char *extensionOwner = CitusExtensionOwnerName();
 	char *escapedNodeName = quote_literal_cstr(nodeNameString);
 
 	WorkerNode *workerNode = NULL;
-	char *localGroupIdUpdateCommand = NULL;
-	List *recreateMetadataSnapshotCommandList = NIL;
-	List *dropMetadataCommandList = NIL;
-	List *createMetadataCommandList = NIL;
+
+	/* fail if metadata synchronization doesn't succeed */
+	bool raiseInterrupts = true;
 
 	EnsureCoordinator();
 	EnsureSuperUser();
@@ -93,6 +92,8 @@ start_metadata_sync_to_node(PG_FUNCTION_ARGS)
 	CheckCitusVersion(ERROR);
 
 	PreventInTransactionBlock(true, "start_metadata_sync_to_node");
+
+	LockRelationOid(DistNodeRelationId(), ExclusiveLock);
 
 	workerNode = FindWorkerNode(nodeNameString, nodePort);
 	if (workerNode == NULL)
@@ -123,31 +124,8 @@ start_metadata_sync_to_node(PG_FUNCTION_ARGS)
 		PG_RETURN_VOID();
 	}
 
-	/* generate and add the local group id's update query */
-	localGroupIdUpdateCommand = LocalGroupIdUpdateCommand(workerNode->groupId);
-
-	/* generate the queries which drop the metadata */
-	dropMetadataCommandList = MetadataDropCommands();
-
-	/* generate the queries which create the metadata from scratch */
-	createMetadataCommandList = MetadataCreateCommands();
-
-	recreateMetadataSnapshotCommandList = lappend(recreateMetadataSnapshotCommandList,
-												  localGroupIdUpdateCommand);
-	recreateMetadataSnapshotCommandList = list_concat(recreateMetadataSnapshotCommandList,
-													  dropMetadataCommandList);
-	recreateMetadataSnapshotCommandList = list_concat(recreateMetadataSnapshotCommandList,
-													  createMetadataCommandList);
-
-	/*
-	 * Send the snapshot recreation commands in a single remote transaction and
-	 * error out in any kind of failure. Note that it is not required to send
-	 * createMetadataSnapshotCommandList in the same transaction that we send
-	 * nodeDeleteCommand and nodeInsertCommand commands below.
-	 */
-	EnsureNoModificationsHaveBeenDone();
-	SendCommandListToWorkerInSingleTransaction(nodeNameString, nodePort, extensionOwner,
-											   recreateMetadataSnapshotCommandList);
+	SyncMetadataSnapshotToNode(workerNode, raiseInterrupts);
+	MarkNodeMetadataSynced(workerNode->workerName, workerNode->workerPort, true);
 
 	PG_RETURN_VOID();
 }
@@ -170,6 +148,8 @@ stop_metadata_sync_to_node(PG_FUNCTION_ARGS)
 	EnsureSuperUser();
 	CheckCitusVersion(ERROR);
 
+	LockRelationOid(DistNodeRelationId(), ExclusiveLock);
+
 	workerNode = FindWorkerNode(nodeNameString, nodePort);
 	if (workerNode == NULL)
 	{
@@ -178,6 +158,7 @@ stop_metadata_sync_to_node(PG_FUNCTION_ARGS)
 	}
 
 	MarkNodeHasMetadata(nodeNameString, nodePort, false);
+	MarkNodeMetadataSynced(nodeNameString, nodePort, false);
 
 	PG_RETURN_VOID();
 }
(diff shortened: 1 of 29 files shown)


src/backend/distributed/metadata/metadata_sync.c
```diff
@@ -862,10 +871,36 @@ LocalGroupIdUpdateCommand(int32 groupId)
 
 /*
  * MarkNodeHasMetadata function sets the hasmetadata column of the specified worker in
- * pg_dist_node to true.
+ * pg_dist_node to hasNodeMetadata.
```
serprex: `hasMetadata` not `hasNodeMetadata`

src/backend/distributed/utils/maintenanced.c
```diff
@@ -356,6 +361,29 @@ CitusMaintenanceDaemonMain(Datum main_arg)
 		}
 #endif
 
+		if (TimestampDifferenceExceeds(lastMetadataSyncTime,
+									   GetCurrentTimestamp(),
+									   METADATA_SYNC_TIMEOUT) && false)
```
serprex: `&& false` should be removed

src/backend/distributed/utils/maintenanced.c
```diff
@@ -356,6 +361,29 @@ CitusMaintenanceDaemonMain(Datum main_arg)
...
+		{
+			InvalidateMetadataSystemCache();
```
serprex: CAn we check if any nodes need sync before invalidation?

src/backend/distributed/utils/maintenanced.c
pykello: Currently `SyncMetadataToNodes()` creates a deadlock when run concurrently with `start_metadata_sync_to_node()`, so I've disabled this block temporarily so I can come back to it after fixing other problems.

src/backend/distributed/utils/maintenanced.c
pykello: Note that in `maintenanced.c` we don't get cache invalidation messages, so to have the most up-to-date view of `pg_dist_node`, etc. we need to invalidate the metadata cache. Otherwise we cannot check if any nodes need sync.

src/backend/distributed/metadata/metadata_sync.c
```diff
@@ -236,6 +215,61 @@ ShouldSyncTableMetadata(Oid relationId)
...
+ * RecreateMetadataSnapshot does the following:
+ *  1. Sets the localGroupId on the worker so the worker knows which tuple in
+ *     pg_dist_node represents itself.
+ *  2. Recreates the distributed metadata on the given worker.
+ *  3. If synchronization succeeds, it sets the pg_dist_node.metadatasynced for
+ *     this node to true.
+ * If raiseOnError is true, it errors out if synchronization fails.
+ */
+bool
+RecreateMetadataSnapshot(WorkerNode *workerNode, bool raiseOnError)
```
marcocitus: Maybe SyncMetadataSnapshotToNodes to indicate that this actually sends the metadata as well.

src/backend/distributed/transaction/worker_transaction.c
```diff
@@ -215,10 +236,25 @@ SendCommandToWorkersParams(TargetWorkerSet targetWorkerSet, char *command,
 												parameterTypes, parameterValues);
 		if (querySent == 0)
 		{
-			ReportConnectionError(connection, ERROR);
+			if (targetWorkerSet == WORKERS_WITH_METADATA)
+			{
+				ReportConnectionError(connection, WARNING);
+				MarkNodeMetadataSynced(connection->hostname, connection->port, false);
```
marcocitus: Maybe we can leave these functions as they are and just have some alternative implementation for master_update_node?

Most of the time, having atomic metadata updates (i.e. erroring out on failure) seems preferable to setting metadatasynced to false. We also need to preserve a working, fault-tolerant MX implementation (MX = clients connect directly to workers). master_update_node is different because we need to make progress while a node is down for the system to recover. We'll need to figure out a separate solution for clients connecting to stale nodes (e.g. take node out of load balancer until synced), but at least that will be limited to failover scenarios.

src/backend/distributed/utils/node_metadata.c
```diff
@@ -1608,3 +1622,27 @@ DatumToString(Datum datum, Oid dataType)
...
 	return outputString;
 }
+
+
+/*
+ * SyncDistNodeEntry synchronizes the corresponding entry for the given workerNode
+ * in pg_dist_node metadata table of all workers.
+ */
+static void
+SyncDistNodeEntry(WorkerNode *workerNode)
```
marcocitus: Maybe `SyncDistNodeEntryToNodes`

pykello: done

src/backend/distributed/metadata/metadata_sync.c

src/backend/distributed/utils/node_metadata.c
```diff
@@ -579,6 +611,81 @@ master_update_node(PG_FUNCTION_ARGS)
...
+	char *extensionOwner = CitusExtensionOwnerName();
+	char *nodeDeleteCommand = NodeDeleteCommand(nodeToSync->nodeId);
+	char *nodeInsertCommand = NodeListInsertCommand(list_make1(nodeToSync));
+	List *commandList = list_make2(nodeDeleteCommand, nodeInsertCommand);
+
+	foreach(workerCell, metadataWorkers)
+	{
+		WorkerNode *workerNode = lfirst(workerCell);
+		bool success =
+			SendOptionalCommandListToWorkerInTransaction(workerNode->workerName,
```
pykello: Using this is not correct. It breaks transaction semantics. Instead I am creating a `SendOptionalCommandToWorkers()` with which is similar to `SendCommandToWorkers()` but uses SAVEPOINTs to rollback failures. Couldn't get it done today.

src/test/regress/sql/multi_mx_master_update_node.sql
```diff
@@ -0,0 +1,52 @@
...
+BEGIN TRANSACTION ISOLATION LEVEL REPEATABLE READ;
+SELECT nodeid, nodename, nodeport, hasmetadata, metadatasynced FROM pg_dist_node;
+SELECT 1 FROM master_update_node((SELECT nodeid FROM pg_dist_node),
+                                 'localhost', :worker_1_port);
+SELECT nodeid, nodename, nodeport, hasmetadata, metadatasynced FROM pg_dist_node;
+END;
+
+-- wait until maintenance daemon does the next metadata sync, and then
+-- check if metadata is synced again
+SELECT pg_sleep(2);
```
pykello: We shouldn't do this. Instead we should have a test UDF which waits for the next sync event.

src/backend/distributed/transaction/worker_transaction.c
pykello: Makes sense. Today I couldn't get this done (maintenance daemon related issues took time). Our changes to `SendCommandToWorkers()` weren't even correct, since on error it just left the worker connection in error state. I'll add a `SendOptionalCommandToWorkers()` which uses savepoints to rollback errors.

src/backend/distributed/utils/node_metadata.c
```diff
@@ -566,6 +579,25 @@ master_update_node(PG_FUNCTION_ARGS)
...
 
+	strlcpy(workerNode->workerName, newNodeNameString, WORKER_LENGTH);
+	workerNode->workerPort = newNodePort;
+
+	/*
+	 * Propagate the updated pg_dist_node entry to all metadata workers.
+	 * The usual case is that the new node is in stand-by mode, so we don't
+	 * sync the changes to the new node, and instead schedule it for being
+	 * synced by the maintenance daemon.
+	 */
```
marcocitus: Awesome.

src/test/regress/expected/multi_mx_master_update_node.out
```diff
@@ -0,0 +1,216 @@
...
+SELECT nodeid, nodename, nodeport, hasmetadata, metadatasynced FROM pg_dist_node;
+ nodeid | nodename  | nodeport | hasmetadata | metadatasynced 
+--------+-----------+----------+-------------+----------------
+      2 | localhost |    12345 | t           | f
+(1 row)
+
+END;
+-- maintenace daemon metadata sync should fail, because node is still unwriteable.
+SELECT wait_until_metadata_sync();
+ERROR:  could not start maintenance background worker
```
pykello: Still looking into this ...

onderkalaci:
> Addresses #2915

@pykello I think it'd be great if you can summarize what are the main steps we're taking in the PR description. Thanks!

pykello:
> > Addresses #2915
> 
> @pykello I think it'd be great if you can summarize what are the main steps we're taking in the PR description. Thanks!

@onderkalaci sure, once I'm done with the changes. Still couple of minor things that are left to figure out.

pykello:
@marcocitus @onderkalaci I need to add few isolation tests too, but otherwise I think this is ready for the final review.

src/test/regress/expected/multi_mx_master_update_node.out
```diff
@@ -0,0 +1,378 @@
...
+SELECT nextval('pg_catalog.pg_dist_groupid_seq') AS last_group_id \gset
+SELECT nextval('pg_catalog.pg_dist_node_nodeid_seq') AS last_node_id \gset
+SELECT nextval('pg_catalog.pg_dist_colocationid_seq') AS last_colocation_id \gset
+SELECT nextval('pg_catalog.pg_dist_shardid_seq') AS last_shard_id \gset
+SET citus.replication_model TO streaming;
+SET citus.shard_count TO 8;
+SET citus.shard_replication_factor TO 1;
+ALTER SYSTEM citus.metadata_sync_interval TO 4000;
+ERROR:  syntax error at or near "citus"
+LINE 1: ALTER SYSTEM citus.metadata_sync_interval TO 4000;
```
pykello: Fix this

src/test/regress/expected/multi_mx_master_update_node.out
```diff
@@ -0,0 +1,378 @@
...
+                     ^
+ALTER SYSTEM citus.metadata_sync_retry_interval TO 1000;
+ERROR:  syntax error at or near "citus"
+LINE 1: ALTER SYSTEM citus.metadata_sync_retry_interval TO 1000;
```
pykello: Fix this

src/backend/distributed/utils/node_metadata.c
```diff
@@ -566,6 +583,29 @@ master_update_node(PG_FUNCTION_ARGS)
 
 	UpdateNodeLocation(nodeId, newNodeNameString, newNodePort);
 
+	strlcpy(workerNode->workerName, newNodeNameString, WORKER_LENGTH);
```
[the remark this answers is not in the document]
marcocitus: I ran master_update_node and it gave a slightly gibberish error. Fixed now.

src/test/regress/expected/multi_mx_master_update_node.out
```diff
@@ -0,0 +1,373 @@
...
+SELECT nextval('pg_catalog.pg_dist_placement_placementid_seq') AS last_placement_id
+\gset
+SELECT nextval('pg_catalog.pg_dist_groupid_seq') AS last_group_id \gset
+SELECT nextval('pg_catalog.pg_dist_node_nodeid_seq') AS last_node_id \gset
+SELECT nextval('pg_catalog.pg_dist_colocationid_seq') AS last_colocation_id \gset
+SELECT nextval('pg_catalog.pg_dist_shardid_seq') AS last_shard_id \gset
+SET citus.replication_model TO streaming;
+SET citus.shard_count TO 8;
+SET citus.shard_replication_factor TO 1;
+-- set sync intervals to less than 5s so wait_until_metadata_sync never times out
```
marcocitus: This often breaks when running tests valgrind, maybe make the difference a bit more aggressive..

src/test/regress/expected/multi_mx_master_update_node.out
```diff
@@ -0,0 +1,373 @@
...
+    AS $$
+    SELECT master_run_on_worker(ARRAY[hostname], ARRAY[port],
+           ARRAY['ALTER SYSTEM SET default_transaction_read_only TO ' || isreadonly::TEXT], false);
+    SELECT result FROM
+        master_run_on_worker(ARRAY[hostname], ARRAY[port],
+                             ARRAY['SELECT pg_reload_conf()'], false);
+$$;
+-- add a node to the cluster
+SELECT master_add_node('localhost', :worker_1_port) As nodeid_1 \gset
+SELECT * FROM pg_dist_node;
```
marcocitus: This will probably break when we add more columns.

src/test/regress/expected/multi_mx_master_update_node.out
pykello: Done. Increased the timeout to 15s and interval to 3s. Tested it for couple of times with valgrind.

src/test/regress/expected/multi_mx_master_update_node.out
pykello: done

pykello: Also updated multi_read_from_secondaries to not fail by changes to pg_dist_node. I think it should be good to see the whole pg_dist_node entries in multi_cluster_management though.

src/backend/distributed/shared_library_init.c
```diff
@@ -605,6 +606,30 @@ RegisterCitusConfigVariables(void)
...
+	DefineCustomIntVariable(
+		"citus.metadata_sync_interval",
+		gettext_noop("Sets the time to wait between metadata syncs."),
+		gettext_noop("metadata sync needs to run every so often "
+					 "to synchronize metadata to metadata nodes "
+					 "that are out of sync."),
+		&MetadataSyncInterval,
+		60000, 1, 7 * 24 * 3600 * 1000,
+		PGC_SIGHUP,
+		GUC_UNIT_MS,
```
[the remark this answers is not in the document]
pykello: done

src/backend/distributed/metadata/metadata_sync.c
```diff
@@ -237,6 +218,112 @@ ShouldSyncTableMetadata(Oid relationId)
...
+bool
+SendOptionalCommandListToWorkerInTransaction(char *nodeName, int32 nodePort,
+											 char *nodeUser, List *commandList)
+{
+	MultiConnection *workerConnection = NULL;
+	ListCell *commandCell = NULL;
+	int connectionFlags = FORCE_NEW_CONNECTION;
+	bool failed = false;
+
+	if (XactModificationLevel > XACT_MODIFICATION_NONE)
```
marcocitus: If this check is still relevant maybe move it to the caller such that this becomes a generic utility function.

src/backend/distributed/utils/node_metadata.c
```diff
@@ -591,6 +625,170 @@ master_update_node(PG_FUNCTION_ARGS)
...
+			}
+
+			commandResult = GetRemoteCommandResult(connection, raiseInterrupts);
+			responseOK = IsResponseOK(commandResult);
+			PQclear(commandResult);
+			ForgetResults(connection);
+
+			if (!responseOK)
+			{
+				ReportResultError(connection, commandResult, WARNING);
```
marcocitus: We've already cleared the result here.

src/backend/distributed/metadata/metadata_sync.c
pykello: I think it can be removed, since `start_metadata_sync_to_node` cannot be run in a transaction.

src/backend/distributed/utils/node_metadata.c
pykello: fixed.

src/backend/distributed/utils/node_metadata.c
```diff
@@ -591,6 +625,176 @@ master_update_node(PG_FUNCTION_ARGS)
 }
 
 
+/* MetadataNodeList returns list of all synced metadata workers. */
```
pykello:
```suggestion
/* SyncedMetadataNodeList returns list of all synced metadata workers. */
```

pykello:
> Looks good. I guess we'll have some performance blips during master_update_node, but I can't think of an easy way to fix that.

I thought about scheduling metadata propagation to be done in commit time, but implementing that correctly wouldn't be very easy. Maybe we can attempt that in the next release.

PR #2929, opened 2019-09-04
get_catalog_object_by_oid requires an extra parameter in pg12

serprex:
While this fixes the compile error, pg12 still fails to create the schema in `multi_metadata_sync`


thanodnl:
Thanks for finding this, turns out to be a bug that was not caught in #2882 .
The fix is subtile, but I added an explanation to the git commit. Given its importance I have created #2930 as well. If we would land this to master before can we keep the history commit of the fix separate? This to make the history easier to track.

thanodnl:
The patch has been landed with #2930 to make sure this bug is fixed on master.

PR #2930, opened 2019-09-05
Fix: Remove early escape in walking pg_depend

thanodnl:
DESCRIPTION: Fix bug in pg_depend_recurse introduced in #2882

This is a bug that got in when we inlined the body of a function into this loop. Earlier revisions had two loops, hence a function that would be reused.

With a return instead of a continue the list of dependencies being walked is dependent on the order in which we find them in pg_depend. This became apparent during pg12 compatibility. The order of entries in pg12 was luckily different causing a random test to fail due to this return.

By changing it to a continue we only skip the entries that we don’t want to follow instead of skipping all entries that happen to be found later.

### DIFF
--- a/src/test/regress/expected/isolation_ensure_dependency_activate_node.out
+++ b/src/test/regress/expected/isolation_ensure_dependency_activate_node.out
@@ -8,7 +8,7 @@ step s1-print-distributed-objects:
     SELECT nodename, nodeport, isactive FROM master_add_node('localhost', 57638);
 
     -- print an overview of all distributed objects
-    SELECT pg_identify_object_as_address(classid, objid, objsubid) FROM citus.pg_dist_object;
+    SELECT pg_identify_object_as_address(classid, objid, objsubid) FROM citus.pg_dist_object ORDER BY 1;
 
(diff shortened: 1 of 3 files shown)


src/test/regress/expected/isolation_ensure_dependency_activate_node.out
```diff
@@ -8,7 +8,7 @@ step s1-print-distributed-objects:
     SELECT nodename, nodeport, isactive FROM master_add_node('localhost', 57638);
 
     -- print an overview of all distributed objects
-    SELECT pg_identify_object_as_address(classid, objid, objsubid) FROM citus.pg_dist_object;
+    SELECT pg_identify_object_as_address(classid, objid, objsubid) FROM citus.pg_dist_object GROUP BY 1;
```
marcocitus: What's the benefit of the GROUP BY?

thanodnl: should have been an `ORDER BY`, thanks.

The test failed once due to different ordering.

PR #2932, opened 2019-09-05
Specialize SendCommandToWorkers into SendCommandToMetadataWorkers

serprex:
We only ever call `SendCommandToWorkers` with `WORKERS_WITH_METADATA`

& in https://github.com/citusdata/citus/commit/9929ced071753f1a7005ae2ec86dfcf10a964005 I started adding logic specific to `WORKERS_WITH_METADATA`

So this avoids creating dead code paths


pykello:
Sorry for the suggestion, #2893 uses this with `ALL_WORKERS`.

thanodnl:
That is due for refactor, as it is suggested to use the adaptive executor instead. I will let you know if I require the function with `ALL_WORKERS` in the end or not.

PR #2934, opened 2019-09-06
Add shard rebalancer stubs

JelteF:
DESCRIPTION: Add shard rebalancer stubs

NOTE: The macro for for the not supported functions could be used in a lot more
places. I haven't done for two reasons:
- I don't know what the right place in the codebase would be to put a macro
  definition like this.
- I didn't want to clutter this PR.


src/backend/distributed/master/shard_rebalancer.c
```diff
@@ -0,0 +1,28 @@
...
+ * $Id$
+ *
+ *-------------------------------------------------------------------------
+ */
+
+#include "postgres.h"
+#include "fmgr.h"
+
+
+#define NOT_SUPPORTED_IN_COMMUNITY(name) \
```
marcocitus: That sees like a useful macro. Is there an appropriate header file?

JelteF: Yeah I was thinking the same, but I don't know the codebase well enough to know where it should be.

PR #2935, opened 2019-09-06
Refactor part of pg_dist_node and make it easier to update

JelteF:
DESCRIPTION: Refactor part of pg_dist_node and make it easier to update

This PR contains two separate changes that I thought were useful when I was adding a new column to `pg_dist_node`. If needed they can be separated to two different PRs (or only a single one could be merged).

### Refactor `InsertNodeRow` and `AddNodeMetadata` to not have 9 arguments
I needed to add another column which would require to add another argument to an already 9 argument function signature. In this case it would be a boolean flag and there were already two boolean flags in there. In my experience it becomes really easy to mess up the order of these flags at that point. Especially because the type system doesn't distinguish between the 3 different booleans with completely different meanings.

So I refactored these signatures to receive a struct containing most of these arguments. Like that you don't mess up orderening, because the meaning of the boolean is not order dependent but fieldname dependent. It also makes it possible to set good shared defaults for this struct.

### Place UDFs in separate files and `#include` them in sql migration files
@thanodnl told me it was a bit of a problem that it's impossible to see the history of a UDF in git. The only way to do so is by reading all the sql migration files from new to old. Another problem is that it's also hard to review the changed UDF during code review, because to find out what changed you have to do the same. I thought of a IMHO better (but not perfect) way to handle this.

We keep the definition of a UDF in `sql/udfs/{name_of_udf}/latest.sql`. That file we change whenever we need to make a change to the the UDF. On top of that you also make a snapshot of the file in `sql/udfs/{name_of_udf}/{migration-version}.sql` (e.g. `8.4-1.sql`) by copying the contents. This way you can easily view what the actual changes where by looking at the `latest.sql` file.

There's still the question on how to use these files then. Sadly postgres doesn't allow inclusion of other sql files in the migration sql file (it does in psql using `\i`). So instead I used the C preprocessor+ `make` to compile a `sql/xxx.sql` to a `build/sql/xxx.sql` file. This final `build/sql/xxx.sql` file has every occurence of `#include "somefile.sql"` in `sql/xxx.sql` replaced by the contents of `somefile.sql`.

### Example usage of changes

This PR shows how to use the following changes from the refactor: https://github.com/citusdata/citus/pull/2937

### TODO if this idea is deemed good
- [x] hide copied files on github to avoid clutter
- [x] Use C preprocessor
- [ ] Add a test in CI to make sure that the contents of `latest.sql` are the same as the contents of `{highest_version_number}.sql`. Not super hard to do, simple shell script with `diff` should suffice. The most complex part is parsing the version number so 10.1-1 is newer than 9.2-3 (i.e. split on `.` and `-` and do numerical compares from left to right)


src/backend/distributed/template.awk
```diff
@@ -0,0 +1,24 @@
+# https://stackoverflow.com/a/51808896/2570866
```
serprex: An alternative that avoids our own build script is to use the C preprocessor & `#include`. This also opens up being able to reference `PG_VERSION_NUM`

See https://github.com/citusdata/postgresql-hll/pull/73/commits/8d4b4c7fda9a3c62cbbc42407707d999f880705d where I implemented this for postgresql-hll

JelteF: That's a good idea. Definitely open to do the same for this PR.

marcocitus:
Should we stop returning a record from master_add_node? Maybe just node ID? As far as I know none of the current control plane implementations really care what master_add_node returns.

JelteF:
> 
> 
> Should we stop returning a record from master_add_node? Maybe just node ID? As far as I know none of the current control plane implementations really care what master_add_node returns.

That's possible, but when adding a node with psql I think it's quite nice to see the record. That way you can see directly if the node is added the way you meant it, instead of having to look it up in `pg_dist_node`. Not sure if it's worth the hassle of having to update it every time though.

src/backend/distributed/template.awk
JelteF: Done

JelteF:
Closing this in favor of the split up PRs #2952 and #2951

PR #2938, opened 2019-09-06
Introduce the concept of Local Execution

onderkalaci:
DESCRIPTION: Introduce local query execution for Citus MX

Based on #2830. 

local_executor.c

The scope of the local execution is locally executing the queries on the
shards. In other words, local execution does not deal with any local tables
that are not shards on the node that the query is being executed. In that sense,
the local executor is only triggered if the node has both the metadata and the
shards (e.g., only Citus MX worker nodes).

The goal of the local execution is to skip the unnecessary network round-trip
happening on the node itself. Instead, identify the locally executable tasks and
simply call PostgreSQL's planner and executor.

The local executor is an extension of the adaptive executor. So, the executor uses
adaptive executor's custom scan nodes.

One thing to note that Citus MX is only supported with replication factor = 1, so
keep that in mind while continuing the comments below.

On the high level, there are 3 slightly different ways of utilizing local execution:

(1) Execution of local single shard queries of a distributed table

   This is the simplest case. The executor kicks at the start of the adaptive
   executor, and since the query is only a single task the execution finishes
   without going to the network at all.

   Even if there is a transaction block (or recursively planned CTEs), as long
   as the queries hit the shards on the same, the local execution will kick in.

(2) Execution of local single queries and remote multi-shard queries

   The rule is simple. If a transaction block starts with a local query execution,
   all the other queries in the same transaction block that touch any local shard
   have to use the local execution. Although this sounds restrictive, we prefer to
   implement in this way, otherwise we'd end-up with as complex scenarious as we
   have in the connection managements due to foreign keys.

   See the following example:
   BEGIN;
       -- assume that the query is executed locally
       SELECT count() FROM test WHERE key = 1;

       -- at this point, all the shards that reside on the
       -- node is executed locally one-by-one. After those finishes
       -- the remaining tasks are handled by adaptive executor
       SELECT count() FROM test;

(3) Modifications of reference tables

		Modifications to reference tables have to be executed on all nodes. So, after the
		local execution, the adaptive executor keeps continuing the execution on the other
		nodes.

		Note that for read-only queries, after the local execution, there is no need to
		kick in adaptive executor.

There are also few limitations/trade-offs that is worth mentioning. First, the
local execution on multiple shards might be slow because the execution has to
happen one task at a time (e.g., no parallelism). Second, if a transaction
block/CTE starts with a multi-shard command, we do not use local query execution
since local execution is sequential. Basically, we do not want to lose parallelism
across local tasks by switching to local execution. Third, the local execution
currently only supports queries. In other words, any utility commands like TRUNCATE,
fails if the command is executed after a local execution inside a transaction block.
Forth, the local execution cannot be mixed with the executors other than adaptive,
namely task-tracker, real-time and router executors. Finally, related with the
previous item, COPY command cannot be mixed with local execution in a transaction.
The implication of that any part of INSERT..SELECT via coordinator cannot happen
via the local execution.

TODO:

- [ ] Run some benchmarks
- [ ] Should we call it `local_shard_executor.c` ?

### DIFF
--- a/src/backend/distributed/connection/connection_management.c
+++ b/src/backend/distributed/connection/connection_management.c
@@ -161,6 +161,41 @@ AfterXactConnectionHandling(bool isCommit)
 
 
 /*
+ * AnyConnectionAccessedPlacements goes over all the connections in the ConnectionHash
+ * and returns true if any of the connection has any referenced placements. This is
+ * useful to detect whether we're in a distirbuted transaction and already executed
+ * at least one command that accessed to a placement.
+ */
+bool
+AnyConnectionAccessedPlacements(void)
+{
+	HASH_SEQ_STATUS status;
+	ConnectionHashEntry *entry;
+
+	hash_seq_init(&status, ConnectionHash);
+	while ((entry = (ConnectionHashEntry *) hash_seq_search(&status)) != 0)
+	{
+		dlist_iter iter;
+
+		dlist_foreach(iter, entry->connections)
+		{
+			MultiConnection *connection =
+				dlist_container(MultiConnection, connectionNode, iter.cur);
+
+			if (!dlist_is_empty(&connection->referencedPlacements))
+			{
+				hash_seq_term(&status);
+
+				return true;
+			}
+		}
+	}
+
+	return false;
+}
+
+
+/*
  * GetNodeConnection() establishes a connection to remote node, using default
  * user and database.
  *
--- a/src/backend/distributed/executor/adaptive_executor.c
+++ b/src/backend/distributed/executor/adaptive_executor.c
@@ -131,6 +131,7 @@
 #include "commands/dbcommands.h"
 #include "distributed/citus_custom_scan.h"
 #include "distributed/connection_management.h"
+#include "distributed/local_executor.h"
 #include "distributed/multi_client_executor.h"
 #include "distributed/multi_executor.h"
 #include "distributed/multi_physical_planner.h"
@@ -528,9 +529,10 @@ static void CleanUpSessions(DistributedExecution *execution);
 
 static void LockPartitionsForDistributedPlan(DistributedPlan *distributedPlan);
 static void AcquireExecutorShardLocksForExecution(DistributedExecution *execution);
+static void AdjustDistributedExecutionAfterLocalExecution(DistributedExecution *execution,
+														  List *remoteTaskList);
 static bool DistributedExecutionModifiesDatabase(DistributedExecution *execution);
 static bool DistributedPlanModifiesDatabase(DistributedPlan *plan);
-static bool TaskListModifiesDatabase(RowModifyLevel modLevel, List *taskList);
 static bool DistributedExecutionRequiresRollback(DistributedExecution *execution);
 static bool SelectForUpdateOnReferenceTable(RowModifyLevel modLevel, List *taskList);
 static void AssignTasksToConnections(DistributedExecution *execution);
@@ -588,11 +590,11 @@ AdaptiveExecutor(CustomScanState *node)
 	DistributedExecution *execution = NULL;
 	EState *executorState = ScanStateGetExecutorState(scanState);
 	ParamListInfo paramListInfo = executorState->es_param_list_info;
-	Tuplestorestate *tupleStore = NULL;
 	TupleDesc tupleDescriptor = ScanStateGetTupleDescriptor(scanState);
 	bool randomAccess = true;
 	bool interTransactions = false;
 	int targetPoolSize = MaxAdaptiveExecutorPoolSize;
+
 
 	Job *job = distributedPlan->workerJob;
 	List *taskList = job->taskList;
@@ -609,21 +611,58 @@ AdaptiveExecutor(CustomScanState *node)
 
 	ExecuteSubPlans(distributedPlan);
 
-	scanState->tuplestorestate =
-		tuplestore_begin_heap(randomAccess, interTransactions, work_mem);
-	tupleStore = scanState->tuplestorestate;
-
 	if (MultiShardConnectionType == SEQUENTIAL_CONNECTION)
 	{
+		/* defer decision after ExecuteSubPlans() */
 		targetPoolSize = 1;
 	}
 
-	execution = CreateDistributedExecution(distributedPlan->modLevel, taskList,
-										   distributedPlan->hasReturning,
-										   paramListInfo, tupleDescriptor,
-										   tupleStore, targetPoolSize);
+	scanState->tuplestorestate =
+		tuplestore_begin_heap(randomAccess, interTransactions, work_mem);
 
+	execution = CreateDistributedExecution(distributedPlan->modLevel, taskList,
+										   distributedPlan->hasReturning, paramListInfo,
+										   tupleDescriptor,
+										   scanState->tuplestorestate, targetPoolSize);
+
+	/*
+	 * Make sure that we acquire the appropriate locks even if the local tasks
+	 * are going to be executed with local execution.
+	 */
 	StartDistributedExecution(execution);
+
+	if (ShouldExecuteTasksLocally(distributedPlan))
+	{
+		bool localExecutionHappenedInCommand = false;
+		List *remoteTaskList = NULL;
+
+		localExecutionHappenedInCommand = ExecuteLocalTasks(scanState, &remoteTaskList);
+
+		if (list_length(remoteTaskList) == 0)
+		{
+			if (SortReturning && distributedPlan->hasReturning)
+			{
+				SortTupleStore(scanState);
+			}
+
+			Assert(localExecutionHappenedInCommand);
+
+			/*
+			 * Local execution have executed all the required tasks,
+			 * in that case do not need continue.
+			 */
+			return resultSlot;
+		}
+
+		/*
+		 * If there are more tasks/placements to be executed, make sure that the execution
+		 * struct is updated accurately.
+		 */
+		if (localExecutionHappenedInCommand)
+		{
+			AdjustDistributedExecutionAfterLocalExecution(execution, remoteTaskList);
+		}
+	}
 
 	if (ShouldRunTasksSequentially(execution->tasksToExecute))
 	{
@@ -636,7 +675,21 @@ AdaptiveExecutor(CustomScanState *node)
 
 	if (distributedPlan->modLevel != ROW_MODIFY_READONLY)
 	{
-		executorState->es_processed = execution->rowsProcessed;
+		if (!LocalExecutionHappened)
+		{
+			Assert(executorState->es_processed == 0);
+
+			executorState->es_processed = execution->rowsProcessed;
+		}
+		else if (PartitionMethod(distributedPlan->targetRelationId) != DISTRIBUTE_BY_NONE)
+		{
+			/*
+			 * For reference tables we already add rowsProcessed on the local execution,
+			 * this is required to ensure that mixed local/remote executions reports
+			 * the accurate number of rowsProcessed to the user.
+			 */
+			executorState->es_processed += execution->rowsProcessed;
+		}
 	}
 
 	FinishDistributedExecution(execution);
@@ -647,6 +700,28 @@ AdaptiveExecutor(CustomScanState *node)
 	}
 
 	return resultSlot;
+}
+
+
+/*
+ * AdjustDistributedExecutionAfterLocalExecution simply updates the necessary fields of
+ * the distributed execution.
+ */
+static void
+AdjustDistributedExecutionAfterLocalExecution(DistributedExecution *execution,
+											  List *remoteTaskList)
+{
+	/*
+	 * Local execution already stores the  tuples for returning, so we should not
+	 * store again.
+	 */
+	execution->hasReturning = false;
+
+	/* we only need to execute the remote tasks */
+	execution->tasksToExecute = remoteTaskList;
+
+	execution->totalTaskCount = list_length(remoteTaskList);
+	execution->unfinishedTaskCount = list_length(remoteTaskList);
 }
 
 
@@ -704,6 +779,12 @@ ExecuteTaskListExtended(RowModifyLevel modLevel, List *taskList,
 {
 	DistributedExecution *execution = NULL;
 	ParamListInfo paramListInfo = NULL;
+
+	/*
+	 * The code-paths that rely on this function do not know how execute
+	 * commands locally.
+	 */
+	ErrorIfLocalExecutionHappened();
 
 	if (MultiShardConnectionType == SEQUENTIAL_CONNECTION)
 	{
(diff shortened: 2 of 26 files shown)


src/backend/distributed/executor/local_executor.c
```diff
@@ -0,0 +1,579 @@
...
+	uint64 totalRowsProcessed = 0;
+
+	foreach(taskCell, taskList)
+	{
+		Task *task = (Task *) lfirst(taskCell);
+
+		PlannedStmt *localPlan = NULL;
+		int cursorOptions = 0;
+		const char *shardQueryString = task->queryString;
+		Query *shardQuery = ParseQueryString(shardQueryString, paramOids, numParams);
```
onderkalaci: Note: `ParseQueryString ()` performs slightly (~15-25%) better than this [code-block](https://github.com/citusdata/citus/pull/2830/files#diff-aff2f55a09f9f9619f28774e6fb18965R363-R378) and seems easier to read. Let me know if you think otherwise.

src/backend/distributed/executor/local_executor.c
```diff
@@ -0,0 +1,579 @@
...
+		}
+		else
+		{
+			param = &paramList->params[numParamIndex];
+		}
+
+		if (param->isnull)
+		{
+			paramListTypes[numParamIndex] = InvalidOid;
+			continue;
```
onderkalaci: Question: Do we really want to skip at this point? Or, could we ever get a NULL param?

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -774,11 +855,20 @@ StartDistributedExecution(DistributedExecution *execution)
 
 	if (MultiShardCommitProtocol != COMMIT_PROTOCOL_BARE)
 	{
-		if (DistributedExecutionRequiresRollback(execution))
+		/*
```
onderkalaci: Question: Do we really need this `|| LocalExecutionHappened `? I added this during the development  because in my first iterations, `StartDistributedExecution ()` was called after local execution. But if we call   `StartDistributedExecution ()`, it should take care  of setting 2PC etc.

src/backend/distributed/executor/local_executor.c
```diff
@@ -0,0 +1,579 @@
...
+
+	readOnlyPlan = !TaskListModifiesDatabase(distributedPlan->modLevel, taskList);
+
+	SplitLocalAndRemoteTasks(readOnlyPlan, taskList, &localTaskList, remoteTaskList);
+	rowProcessed = ExecuteLocalTaskList(node, localTaskList);
+
+	/* we already filtered SELECTs when calculating rowProcessed */
+	executorState->es_processed = rowProcessed;
+
+	LocalExecutionHappened = true;
```
onderkalaci: This doesn't look good. What if `list_length(localTaskList) == 0`? Should we still set `LocalExecutionHappened `?

src/backend/distributed/executor/local_executor.c
```diff
@@ -0,0 +1,579 @@
...
+
+	if (!EnableLocalExecution)
+	{
+		return false;
+	}
+
+	if (GetLocalGroupId() == 0)
+	{
+		/* currently local execution is only supported on metadata workers */
+		return false;
```
marcocitus: why do we need this one?

src/backend/distributed/executor/local_executor.c
```diff
@@ -0,0 +1,579 @@
...
+	ExecutorStart(queryDesc, eflags);
+	ExecutorRun(queryDesc, scanDirection, 0L, true);
+
+	/*
+	 * We'll set the executorState->es_processed later, for now only remember
+	 * the count.
+	 */
+	if (taskPlan->commandType != CMD_SELECT)
+	{
+		totalRowsProcessed = queryDesc->estate->es_processed;
```
marcocitus: I think the reason you didn't reuse `ExecutePlanIntoDestReceiver` is to get the number of rows, though maybe `ExecutePlanIntoDestReceiver` can return the number using portal->queryDesc?

src/backend/distributed/executor/local_executor.c
```diff
@@ -0,0 +1,579 @@
...
+	return false;
+}
+
+
+/*
+ * TaskAccessLocalNode returns true if any placements of the task reside on the
+ * node that we're executing the query.
+ */
+bool
+TaskAccessLocalNode(Task *task)
```
marcocitus: TaskAccessesLocalNode?

src/backend/distributed/executor/local_executor.c
```diff
@@ -0,0 +1,579 @@
...
+
+	if (!singleTask)
+	{
+		/*
+		 * For multi-task executions, switching to local execution would likely to
+		 * perform poorly, because we'd lose the parallelizm. Note that the local
+		 * execution is happening one task at a time (e.g., similar to sequential
+		 * distributed execution).
+		 */
+		Assert(!LocalExecutionHappened);
```
marcocitus: Given the `if (LocalExecutionHappened) return true;` above, this assert may be a little redundant.

src/backend/distributed/connection/connection_management.c
```diff
@@ -160,6 +160,41 @@ AfterXactConnectionHandling(bool isCommit)
...
 
 
+/*
+ * AnyConnectionAccessedPlacements goes over all the connections in the ConnectionHash
+ * and returns true if any of the connection has any referenced placements. This is
+ * useful to detect whether we're in a distirbuted transaction and already executed
+ * at least one command that accessed to a placement.
+ */
+bool
+AnyConnectionAccessedPlacements(void)
```
marcocitus: this function is potentially quite expensive, is there any faster way we can determine this?

src/backend/distributed/executor/local_executor.c
```diff
@@ -0,0 +1,579 @@
...
+	DistributedPlan *distributedPlan = node->distributedPlan;
+	bool readOnlyPlan = false;
+	Job *job = distributedPlan->workerJob;
+	List *taskList = job->taskList;
+	List *localTaskList = NIL;
+	uint64 rowProcessed = 0;
+
+	*remoteTaskList = NIL;
+
+	readOnlyPlan = !TaskListModifiesDatabase(distributedPlan->modLevel, taskList);
```
marcocitus: Maybe DistributedPlanModifiesDatabase.

src/backend/distributed/executor/local_executor.c
```diff
@@ -0,0 +1,579 @@
...
+
+	SplitLocalAndRemoteTasks(readOnlyPlan, taskList, &localTaskList, remoteTaskList);
+	rowProcessed = ExecuteLocalTaskList(node, localTaskList);
```
marcocitus: I feel like it's a bit strange that these 2 lines are in the same function, while remote execution is handled elsewhere. What if we call `SplitLocalAndRemoteTasks` from `StartDistributedExecution` and maybe keep track of both lists in the `DistributedExecution`?

src/backend/distributed/executor/local_executor.c
```diff
@@ -0,0 +1,579 @@
...
+/*
+ * ExecuteLocalTasks gets a CitusScanState node and executes any tasks that are
+ * local to the node executing the query.
+ *
+ * The function returns true if at least one task has been locally executed. The
+ * function also fills the remoteTaskList so that the caller (e.g., adaptive executor)
+ * can continue with executing the remote tasks.
+ */
+bool
+ExecuteLocalTasks(CitusScanState *node, List **remoteTaskList)
```
marcocitus: I find the naming here a bit confusing. I kind of expect we execute some set of local tasks (like ExecuteLocalTaskList), but it's rather that we'll first determine what work can be done locally and construct a task list from that and then also execute it.

Anyway, would be nice to move the building of separate task lists into StartDistributedExecution.

src/backend/distributed/executor/local_executor.c
```diff
@@ -0,0 +1,579 @@
...
+ * SplitLocalAndRemoteTasks essentially gets a taskList and generates two
+ * task lists namely localTaskList and remoteTaskList. The function goes
+ * over the input taskList and puts the tasks that are local to the node
+ * into localTaskList and the remaining to the remoteTaskList. One slightly
+ * different case is modifications to replicated tables (e.g., reference
+ * tables) where a single task ends in two seperate tasks and the local
+ * task is added to localTaskList and the remanings to the remoteTaskList.
+ */
+static void
+SplitLocalAndRemoteTasks(bool readOnly, List *taskList, List **localTaskList,
```
marcocitus: Not sure if Split is the right word since we don't always Split.

src/backend/distributed/shared_library_init.c
```diff
@@ -509,6 +521,17 @@ RegisterCitusConfigVariables(void)
 		0,
 		NULL, NULL, NULL);
 
+	DefineCustomBoolVariable(
+		"citus.log_local_commands",
+		gettext_noop("Log queries that are executed locally, can be overriden by "
+					 "citus.log_remote_commands"),
```
marcocitus: Maybe just rely on the 1 GUC? The extra GUC just seems like a pain.

src/backend/distributed/executor/local_executor.c
```diff
@@ -0,0 +1,579 @@
...
+
+	ereport(LOG, (errmsg("executing the command locally: %s",
+						 ApplyLogRedaction(command))));
+}
+
+
+/*
+ * SetLocalForceMaxQueryParallelization simply a C interface for
+ * setting the following:
+ *      SET LOCAL citus.multi_shard_modify_mode TO on;
```
marcocitus: stale comment

src/backend/distributed/executor/local_executor.c
marcocitus: I guess it makes sense for simple queries, which is the common case that we're trying to optimise. We could probably do even better if we replaced table OIDs directly with shard OIDs. (not necessary)

src/backend/distributed/executor/local_executor.c
onderkalaci: We can remove this, not necessary

src/backend/distributed/executor/local_executor.c
onderkalaci: I kept it just in case someone tries to re-order the function in the future. I can remove if this looks weird.

onderkalaci:
> - Test 3: Pgbench, 1 worker node, -c64 -c256 -T 120
            - (a) Local execution enabled (tps): `select-only`: 56202   `simple-update`:  11771 `tpcb-like`: 7796
            - (a) Local execution disabled (tps): `select-only`:  24524 `simple-update`: 5077  `tpcb-like`:   3510 (some connection errors for tpcb-like)


FYI: pgbench results look very promising

src/backend/distributed/executor/adaptive_executor.c
onderkalaci: Oh, yes  we need this. I had already added an example which needs  this:

```SQL
-- the same transaction block, but this time
-- enable local execution as well. The first
-- command is locally executed, the second
-- is remote, so 1 entry is expected
BEGIN;
INSERT INTO test_recovery VALUES ('hello');
INSERT INTO test_recovery VALUES ('world');
COMMIT;
SELECT count(*) FROM pg_dist_transaction;
 count 
-------
     1
(1 row)

 recover_prepared_transactions 
                             0
(1 row)
```

src/backend/distributed/executor/local_executor.c
onderkalaci: How do you see the new code organization? I moved the task split (or I may change the term "split") to `CreateDistributedExecution`, which seems to fit nicely.

src/backend/distributed/shared_library_init.c
onderkalaci: I agree, two notes:

(a) `citus.log_remote_commands` overrules `citus.log_local_commands` when it is true. So, when you have `citus.log_remote_commands` set to ON, you'll see both local and remote commands.

(b) The tricky part is that there is no easy way of showing that the execution is done locally. And, if I set `citus.log_remote_commands`, then the regression output becomes extremely verbose. That's why I came up with this trick, which is probably only useful for testing.

If you still think this is not good, I can find another way for understanding the local query executions on the regression tests.

onderkalaci: (And, it is `GUC_NO_SHOW_ALL`)

src/backend/distributed/executor/local_executor.c
onderkalaci: That's a good idea, tried with it but failed. Postgres seem set queryDesc not for all types of [queries](https://github.com/postgres/postgres/blob/eb43f3d19324d7e5376b1f57fc2e5c142a6b5f3d/src/backend/tcop/pquery.c#L482-L537). So, I'm skipping this suggestion?

src/backend/distributed/connection/connection_management.c
onderkalaci: I guess the following should work:  `return hash_get_num_entries(ConnectionPlacementHash) > 0;`

onderkalaci:
>  Should we call it local_shard_executor.c ?

I guess no because we'd want to use the infrastructure for intermediate results on the coordinator

src/backend/distributed/shared_library_init.c
marcocitus: fair

src/test/regress/sql/local_shard_execution.sql
```diff
@@ -0,0 +1,670 @@
...
+INSERT INTO distributed_table VALUES (1, '1', 20);
+INSERT INTO second_distributed_table VALUES (1, '1');
+
+-- a simple test for 
+CREATE TABLE collections_list (
+	key int,
+	ts timestamptz,
+	collection_id integer,
+	value numeric,
+	PRIMARY KEY(key, collection_id)
```
marcocitus: maybe add a bigserial column

src/backend/distributed/executor/local_executor.c
```diff
@@ -0,0 +1,490 @@
...
+		 * perform poorly, because we'd lose the parallelizm. Note that the local
+		 * execution is happening one task at a time (e.g., similar to sequential
+		 * distributed execution).
+		 */
+		Assert(!LocalExecutionHappened);
+
+		return false;
+	}
+
+	return false;
```
marcocitus: Doesn't this kind of obsolete the last `if` block?

src/backend/distributed/executor/local_executor.c
```diff
@@ -0,0 +1,490 @@
...
+		List *localTaskPlacementList = NULL;
+		List *remoteTaskPlacementList = NULL;
+
+		SplitLocalAndRemotePlacements(task->taskPlacementList, &localTaskPlacementList,
+									  &remoteTaskPlacementList);
+
+		/* either the local or the remote should be non-nil */
+		Assert(!(localTaskPlacementList == NIL && remoteTaskPlacementList == NIL));
+
+		if (localTaskPlacementList == NIL || remoteTaskPlacementList == NIL)
```
marcocitus: I find this a bit more understandable if it says `list_length(task->taskPlacementList) == 1`

PR #2940, opened 2019-09-09
Add upgrade postgres version test

SaitTalhaNisanci:
The script does the following:
- Temporary folder `tmp_upgrade` is created in `src/test/regress/`, if one exists it is removed first.
- Database is initialized and citus cluster is created(1 coordinator + 2 workers) with old postgres.
- `before_upgrade_schedule` is run with `pg_regress`. This schedule does not drop any tables or data so that we can verify upgrade.
- `citus_prepare_pg_upgrade` is run in coordinators and workers.
- Old database is stopped.
- A new database is initialized with new postgres under `tmp_upgrade`.
- Postgres upgrade is performed.
- New database is started in both coordinators and workers.
- `citus_finish_pg_upgrade` is run in coordinators and workers to finalize the upgrade step.
- `after_upgrade_schedule` is run with `pg_regress` to verify that the previously created tables, and data still exist. Router and realtime queries are used to verify this.

Currently the schedules to test the upgrade have only basic `create/select` statements, we can add more complex queries later.

The script takes around **24** seconds to execute on my local machine.
If `pg_regress` finds any difference then the script exits with a nonzero code.


src/test/regress/upgrade/upgrade_test.py
```diff
@@ -0,0 +1,139 @@
+"""upgrade_test
+Usage:
+    upgrade_test --old-bindir=<old-bindir> --new-bindir=<new-bindir> --postgres-srcdir=<postgres-srcdir> [--citus-path=<citus-path>]
+
+Options:
+    --old-bindir=<old-bindir>              The old PostgreSQL executable directory;(ex: '~/.pgenv/pgsql/bin')
+    --new-bindir=<new-bindir>              New postgres binary absolute path(ex: '~/.pgenv/pgsql-11.3/bin')
+    --postgres-srcdir=<postgres-srcdir>    Path to postgres build directory(ex: ~/.pgenv/src/postgresql-11.3/src/test/regress)
+    --citus-path=<citus-path>              Absolute path for citus directory (default: ~/citus)
```
thanodnl: What do we use the `--citus-path` for? Quick search doesn't show any usage

src/test/regress/upgrade/upgrade_test.py
```diff
@@ -0,0 +1,139 @@
+"""upgrade_test
+Usage:
+    upgrade_test --old-bindir=<old-bindir> --new-bindir=<new-bindir> --postgres-srcdir=<postgres-srcdir> [--citus-path=<citus-path>]
+
+Options:
+    --old-bindir=<old-bindir>              The old PostgreSQL executable directory;(ex: '~/.pgenv/pgsql/bin')
+    --new-bindir=<new-bindir>              New postgres binary absolute path(ex: '~/.pgenv/pgsql-11.3/bin')
+    --postgres-srcdir=<postgres-srcdir>    Path to postgres build directory(ex: ~/.pgenv/src/postgresql-11.3/src/test/regress)
```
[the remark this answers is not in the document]
thanodnl: seems the example is wrong anyway:
```
Traceback (most recent call last):
  File "upgrade/upgrade_test.py", line 139, in <module>
    main()
  File "upgrade/upgrade_test.py", line 124, in main
    NODE_PORTS[COORDINATOR_NAME], BEFORE_UPGRADE_SCHEDULE)
  File "upgrade/upgrade_test.py", line 75, in run_pg_regress
    exit_code = utils.run(command)
  File "/Users/nilsdijk/Projects/citus/src/test/regress/upgrade/utils.py", line 7, in run
    return subprocess.call(command)
  File "/Users/nilsdijk/.pyenv/versions/3.6.8/lib/python3.6/subprocess.py", line 287, in call
    with Popen(*popenargs, **kwargs) as p:
  File "/Users/nilsdijk/.pyenv/versions/3.6.8/lib/python3.6/subprocess.py", line 729, in __init__
    restore_signals, start_new_session)
  File "/Users/nilsdijk/.pyenv/versions/3.6.8/lib/python3.6/subprocess.py", line 1364, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: '/Users/nilsdijk/.pgenv/src/postgresql-11.5/src/test/regress/src/test/regress/pg_regress': '/Users/nilsdijk/.pgenv/src/postgresql-11.5/src/test/regress/src/test/regress/pg_regress'
```

src/test/regress/upgrade/config.py
```diff
@@ -0,0 +1,41 @@
+OLD_BINDIR = 'OLD_BINDIR'
+NEW_BINDIR = 'NEW_BINDIR'
+TEMP_DIR = 'TEMP_DIR'
+NEW_DATADIR = 'NEW_DATADIR'
+OLD_DATADIR = 'OLD_DATADIR'
+PG_SRCDIR = 'PG_SRCDIR'
+BEFORE_UPGRADE_SCHEDULE = './before_upgrade_schedule'
```
thanodnl: for the normal tests we pass these in via arguments as well. Together with `EXTRA_TESTS='...'`.

This combination makes it easy to have simpler base schedules as well and some `EXTRA_TESTS` to run on top of the base schedule. This is really helpful when you are creating a new test, instead of running everything you can run a bare minimum and your new test without running the complete suite.

Did you have a look if we could have the same workflow here?

src/test/regress/upgrade/config.py
```diff
@@ -0,0 +1,41 @@
...
+    config[TEMP_DIR] = './tmp_upgrade'
+    config[NEW_DATADIR] = config[TEMP_DIR] + '/newData'
+    config[OLD_DATADIR] = config[TEMP_DIR] + '/oldData'    
+
+
+
+USER = 'postgres'
+DBNAME = 'postgres'
+
+COORDINATOR_NAME = 'coordinator'
```
thanodnl: any reason the coordinator is in an variable like this and the workers are not?

src/test/regress/upgrade/config.py
```diff
@@ -0,0 +1,41 @@
+OLD_BINDIR = 'OLD_BINDIR'
+NEW_BINDIR = 'NEW_BINDIR'
+TEMP_DIR = 'TEMP_DIR'
+NEW_DATADIR = 'NEW_DATADIR'
+OLD_DATADIR = 'OLD_DATADIR'
+PG_SRCDIR = 'PG_SRCDIR'
+BEFORE_UPGRADE_SCHEDULE = './before_upgrade_schedule'
+AFTER_UPGRADE_SCHEDULE = './after_upgrade_schedule'
+
+config = {
```
thanodnl: I can see this dictionary works, but it is a bit fragile.
Would it be much work to use a class and members instead of a dict?

src/test/regress/upgrade/upgrade_test.py
```diff
@@ -0,0 +1,139 @@
...
+    start_databases(config[OLD_BINDIR], config[OLD_DATADIR])
+    create_citus_extension(config[OLD_BINDIR])
+    add_workers(config[OLD_BINDIR])
+
+
+def main():
+    initialize_temp_dir()
+    initialize_citus_cluster()
+
+    run_pg_regress(config[OLD_BINDIR], config[PG_SRCDIR],
```
thanodnl: If you provide a path to pg_regress that doesn't exist there are two things that happen:
 - the test suite crashes
 - postgres keeps running

The crash we can live with, but it would be great if postgres gets shut down when our testing application crashes.

src/test/regress/upgrade/upgrade_test.py
```diff
@@ -0,0 +1,139 @@
...
+    utils.run(['mkdir', rel_data_path])
+    for node_name in NODE_NAMES:
+        abs_data_path = os.path.abspath(os.path.join(rel_data_path, node_name))
+        command = [os.path.join(pg_path, 'initdb'),
+                   '--pgdata', abs_data_path, '--username', USER]
+        utils.run(command)
+        add_citus_to_shared_preload_libraries(abs_data_path)
+
+
+def get_add_citus_to_shared_preload_library_cmd(abs_data_path):
```
thanodnl: unused?

src/test/regress/upgrade/upgrade_test.py
```diff
@@ -0,0 +1,139 @@
...
+    return 'echo "shared_preload_libraries = \'citus\'" >> {conf_path}'.format(
+        conf_path=os.path.join(abs_data_path, 'postgresql.conf'))
+
+
+def add_citus_to_shared_preload_libraries(abs_data_path):
```
thanodnl: given that we now also want to write other (`citus.node_conninfo`) settings before we start postgres it might be nice to make this function a bit more robust.

Let is take a dict of settings to write.
Append it to a file with nice python calls instead of echo :) : https://stackoverflow.com/questions/4706499/how-do-you-append-to-a-file-in-python

src/test/regress/upgrade/upgrade_test.py
```diff
@@ -0,0 +1,139 @@
...
+    conf_path = os.path.join(abs_data_path, 'postgresql.conf')
+    with open(conf_path, 'a') as conf_file:
+        utils.run_with_stdout(
+            ['echo', "shared_preload_libraries = \'citus\'"], conf_file)
+
+
+def start_databases(pg_path, rel_data_path):
+    for node_name in NODE_NAMES:
+        abs_data_path = os.path.abspath(os.path.join(rel_data_path, node_name))
+        command = [os.path.join(pg_path, 'pg_ctl'), '--pgdata', abs_data_path, '-U',
```
thanodnl: can we change the formatting of this slightly:
```python
 [ "command", "subcommand",
    "--arg", "value",
    "--arg2", "value2",
   ... ]
```

src/test/regress/upgrade/upgrade_test.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+def create_citus_extension(pg_path):
+    for port in NODE_PORTS.values():
+        utils.psql(pg_path, port, "CREATE EXTENSION citus;")
+
+
+def add_workers(pg_path):
+    for port in WORKER_PORTS:
+        command = "SELECT * from master_add_node('localhost', {port});".format(
```
thanodnl: does this formatting allow us to specify that `port` is expected to be a number?

src/test/regress/upgrade/upgrade_test.py
```diff
@@ -0,0 +1,139 @@
...
+            port=port)
+        utils.psql(pg_path, NODE_PORTS[COORDINATOR_NAME], command)
+
+
+def run_pg_regress(pg_path, PG_SRCDIR, port, schedule):
+    command = [os.path.join(PG_SRCDIR, 'src/test/regress/pg_regress'), '--port', str(port),
```
thanodnl: same formatting as above

src/test/regress/upgrade/upgrade_test.py
```diff
@@ -0,0 +1,139 @@
...
+               '--schedule', schedule, '--bindir', pg_path, '--user', USER, '--dbname', DBNAME,
+               '--use-existing']
+    exit_code = utils.run(command)
+    if exit_code != 0:
+        sys.exit(exit_code)
```
thanodnl: guess that the database will stay running here.

If we could trap our application on exit and use that to shut down postgres would be great.

src/test/regress/upgrade/upgrade_test.py
```diff
@@ -0,0 +1,139 @@
...
+
+def citus_prepare_pg_upgrade(pg_path):
+    for port in NODE_PORTS.values():
+        utils.psql(pg_path, port, "SELECT citus_prepare_pg_upgrade();")
+
+
+def stop_databases(pg_path, rel_data_path):
+    for node_name in NODE_NAMES:
+        abs_data_path = os.path.abspath(os.path.join(rel_data_path, node_name))
+        command = [os.path.join(pg_path, 'pg_ctl'), '--pgdata', abs_data_path, '-U',
```
thanodnl: formatting of `command`

src/test/regress/upgrade/utils.py
```diff
@@ -0,0 +1,31 @@
+import subprocess
+import os
+from config import USER
+
+
+def run(command):
+    return subprocess.call(command)
```
thanodnl: I don't see a reason to have this a utility, we can just use `subrpocess.call` instead of `utils.run`

thanodnl:
Missing notes explaining the ssl issue

When I run the upgrade test locally it fails after the upgrade on ssl issues:
```
--- 1,78 ----
  SET search_path TO upgrade_before, public;
  SELECT * FROM t ORDER BY a;
! ERROR:  connection error: localhost:9701
! DETAIL:  server does not support SSL, but SSL was required
  SELECT * FROM t WHERE a = 1;
! ERROR:  connection error: localhost:9702
! DETAIL:  server does not support SSL, but SSL was required
  DROP SCHEMA upgrade_before CASCADE;
  NOTICE:  drop cascades to table t
+ WARNING:  connection error: localhost:9701
+ DETAIL:  server does not support SSL, but SSL was required
```

You might have postgres compiled locally without openssl. To compile postgres with openssl is done with the `--with-openssl` flag to `./configure` of postgres.

The reason it fails after an upgrade is that Citus enables ssl (when compiled with openssl) during its installation if not configured to not do this. Due to missing upgrade tests we have not caught this problem during the implementation of this ;).

For now you can prevent Citus to use ssl by setting the following setting before you `CREATE EXTENSION citus;`:

```
citus.node_conninfo = 'sslmode=prefer'
```

With this setting set it will not forcefully turn on ssl on the coordinator and workers, but might use it if configured. By default it will be assumed to be
```
citus.node_conninfo = 'sslmode=require'
```

src/test/regress/upgrade/upgrade_test.py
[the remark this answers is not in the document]
SaitTalhaNisanci: Yeah the example is wrong, it is postgres src dir, not regress. I will make it `pgxsdir` to align with `pg_regress_multi.pl`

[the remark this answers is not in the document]
SaitTalhaNisanci: It seems that we have 3 flags for postgres src directory:
https://github.com/citusdata/citus/blob/master/src/test/regress/pg_regress_multi.pl#L41
https://github.com/citusdata/citus/blob/master/src/test/regress/pg_regress_multi.pl#L40
https://github.com/citusdata/citus/blob/master/src/test/regress/pg_regress_multi.pl#L42

Only the last one is used, are the others leftovers possibly then?

src/test/regress/upgrade/upgrade_test.py
SaitTalhaNisanci: It is a leftover, I will remove it.

src/test/regress/upgrade/upgrade_test.py
SaitTalhaNisanci: To resolve this, we can add a try/catch/finally block as follows:
```python
    exit_code = FAIL_CODE # a nonzero code
    try:
        # here run our script
        ...
        # if we come here it means we did not get any exception and pg_regress run successfully so upgrade was successful
        exit_code = SUCCESS_CODE # zero code
    except SystemExit as code:
        # pg_regress returned nonzero code so sys.exit is called, which throws SystemExit exception.
        exit_code = code
    finally:
        # here we can stop the databases that we started and exit with our exit_code
        sys.exit(exit_code)
```

If we have a block like this, we can just check the return code of the script without having to worry about side effects. What do you think @thanodnl?

src/test/regress/upgrade/utils.py
SaitTalhaNisanci: initially I was passing `shell=True`, after removing it, it doesnt really have any utility as we are passing only one argument without any additional things so I will remove this method and use `subprocess.call` directly

src/test/regress/upgrade/upgrade_test.py
thanodnl: makes sense, however this needs to be around every step in the chain:
 - initdb
 - setup citus
 - pg_regress before schedule
 - prepare_upgrade
 - initdb
 - pg_upgrade
 - pg_regress after schedule

So I wonder if python has a more generic hook that we can use on exit to clean up our databases.

thanodnl: possibly something like https://docs.python.org/3/library/atexit.html

src/test/regress/upgrade/upgrade_test.py
SaitTalhaNisanci: I have looked at it, but couldnt find a way to specify that `port` is expected to be a number in this formatting.

src/test/regress/Pipfile.lock
```diff
@@ -131,45 +116,53 @@
...
+                "sha256:8d10113ca826a4c29d5b85b2c4e045ffa8bad74fb525ee0eceb1d38d4c70dfd6",
+                "sha256:be495b8ec5a939a7605274b6e59fbc35e76f5ad814ae010eb679529671c9e119",
+                "sha256:dc2d3f3b1548f4d11786616cf0f4415e25b0fbecb8a1d2cd8c07568f13fdde38",
+                "sha256:e4aecdd9d5a3d06c337894c9a6e2961898d3f64fe54ca920a72234a3de0f9cb3",
+                "sha256:e79ab4485b99eacb2166f3212218dd858258f374855e1568f728462b0e6ee0d9",
+                "sha256:f995d3667301e1754c57b04e0bae6f0fa9d710697a9f8d6712e8cca02550910f"
+            ],
+            "version": "==2.3.1"
+        },
+        "docopt": {
```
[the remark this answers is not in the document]
SaitTalhaNisanci: This seems related: https://github.com/pypa/pipenv/issues/2024
As far as I understand it is not a problem.

src/test/regress/Makefile
```diff
@@ -120,6 +122,9 @@ check-failure-base: all
 	$(pg_regress_multi_check) --load-extension=citus --mitmproxy \
 	-- $(MULTI_REGRESS_OPTS) --schedule=$(citus_abs_srcdir)/failure_base_schedule $(EXTRA_TESTS)
 
+check-upgrade:
+	$(pg_upgrade_check) --old-bindir=$(old-bindir) --new-bindir=$(new-bindir) --pgxsdir=$(pgxsdir)
```
SaitTalhaNisanci: I think if we are going to have a `base_upgrade_schedule` and use `EXTRA_TESTS` argument we will need to add it here. We can give schedules from here as well instead of putting them to the script.

src/test/regress/upgrade/README.md
```diff
@@ -16,15 +16,16 @@ Before running the script, make sure that:
     pipenv install
     pipenv shell
 ```
-- Finally run the script in `citus/src/test/regress`:
+
+- Finally run upgrade test in `citus/src/test/regress`:
 ```
-    pipenv run python upgrade/upgrade_test.py --old-bindir=<old-bindir> --new-bindir=<new-bindir> --postgres-srcdir=<postgres-srcdir>
+    pipenv run make check-upgrade old-bindir=<old-bindir> new-bindir=<new-bindir> pgxsdir=<pgxsdir>
```
thanodnl: Do we need to pass in `pgxsdir`? I thought it was available as a variable in the `Makefile` as an argument.

src/test/regress/upgrade/config.py
```diff
@@ -1,41 +1,32 @@
...
 
+class Config():
+    def __init__(self, arguments):
+        self.old_bindir = arguments['--old-bindir']
+        self.new_bindir = arguments['--new-bindir']
+        self.pg_srcdir = arguments['--pgxsdir']
+        self.temp_dir = './tmp_upgrade'
+        self.old_datadir = self.temp_dir + '/oldData'
+        self.new_datadir = self.temp_dir + '/newData'
+        self.settings = {
```
thanodnl: 👍 like this one!

src/test/regress/upgrade/utils.py
```diff
@@ -3,16 +3,9 @@
...
-    return subprocess.call(command)
-
-def run_with_stdout(command, file):
-    return subprocess.call(command, stdout = file)
-
-
 def psql(pg_path, port, command):
-    return run([os.path.join(pg_path, 'psql'), '-U', USER, '-p', str(port),
-                '-c {}'.format(command)])
+    return subprocess.call([os.path.join(pg_path, 'psql'), '-U', USER, '-p', str(port),
```
thanodnl: can you also format the arguments to psql here the same as the others? It makes it easier to glance an understand what its doing.

src/test/regress/upgrade/utils.py
```diff
@@ -3,16 +3,9 @@
...
+                            '-c {}'.format(command)])
```
thanodnl: wouldn't it be possible to do this without formatting?
```python
[...
'-c', command
]
```

src/test/regress/upgrade/upgrade_test.py
```diff
@@ -0,0 +1,181 @@
...
+    upgrade_test --old-bindir=<old-bindir> --new-bindir=<new-bindir> --pgxsdir=<pgxsdir>
+
+Options:
+    --old-bindir=<old-bindir>              The old PostgreSQL executable directory;(ex: '~/.pgenv/pgsql/bin')
+    --new-bindir=<new-bindir>              New postgres binary absolute path(ex: '~/.pgenv/pgsql-11.3/bin')
+    --pgxsdir=<pgxsdir>           	       Path to the PGXS directory(ex: ~/.pgenv/src/postgresql-11.3)
+"""
+
+import utils
+import atexit
```
thanodnl: given that `atexit` is not in the `Pipfile` I guess it came from the standard library?

src/test/regress/upgrade/upgrade_test.py
```diff
@@ -0,0 +1,181 @@
...
+def start_databases(pg_path, rel_data_path):
+    for node_name in NODE_NAMES:
+        abs_data_path = os.path.abspath(os.path.join(rel_data_path, node_name))
+        command = [
+            os.path.join(pg_path, 'pg_ctl'),
+            '--pgdata', abs_data_path,
+            '-U', USER,
+            "-o '-p {}'".format(NODE_PORTS[node_name]),
+            '--log', os.path.join(pg_path, 'logfile_' + node_name),
+            'start'
```
thanodnl: nit; I thought this one could also be at the beginning `pg_ctl start ...` might be slightly more readable if that is possible.

src/test/regress/upgrade/upgrade_test.py
```diff
@@ -0,0 +1,181 @@
...
+def stop_databases(pg_path, rel_data_path):
+    for node_name in NODE_NAMES:
+        abs_data_path = os.path.abspath(os.path.join(rel_data_path, node_name))
+        command = [
+            os.path.join(pg_path, 'pg_ctl'),
+            '--pgdata', abs_data_path,
+            '-U', USER,
+            "-o '-p {}'".format(NODE_PORTS[node_name]),
+            '--log', os.path.join(pg_path, 'logfile_' + node_name),
+            'stop'
```
thanodnl: same as `start`

src/test/regress/upgrade/upgrade_test.py
```diff
@@ -0,0 +1,181 @@
...
+    start_databases(config.new_bindir, config.new_datadir)
+    citus_finish_pg_upgrade(config.new_bindir)
+
+    run_pg_regress(config.new_bindir, config.pg_srcdir,
+                   NODE_PORTS[COORDINATOR_NAME], AFTER_UPGRADE_SCHEDULE)
+
+
+if __name__ == '__main__':
+    config = Config(docopt(__doc__, version='upgrade_test'))
+    #atexit.register(stop_all_databases, config.old_bindir,
```
thanodnl: commented out? didn't this work?

SaitTalhaNisanci: Oh no, it worked, I was just testing it and clearly forgot to remove the comment before committing :)

src/test/regress/upgrade/README.md
SaitTalhaNisanci: Yeah, it is available. I will update the readme

src/test/regress/upgrade/utils.py
SaitTalhaNisanci: it is definitely possible :)

src/test/regress/Makefile
thanodnl: I have been thinking a bit more on `EXTRA_TESTS` and it will be non-trivial. Normally we run 1 schedule which would in this case be a base schedule, or even an empty on given the setup is done outside of the regression test.

However, we need 2 extra tests
- one that runs before the upgrade
- one that runs after the upgrade

we either need to supply both, or enforce strict naming scheme. Both of the solutions I don't like at the moment. So I guess we can punt that till we have too many upgrade tests and need a quicker way during development to run a selective upgrade test.

src/test/regress/upgrade/upgrade_test.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+def stop_databases(pg_path, rel_data_path):
+    for node_name in NODE_NAMES:
+        abs_data_path = os.path.abspath(os.path.join(rel_data_path, node_name))
+        command = [
+            os.path.join(pg_path, 'pg_ctl'), 'stop',
+            '--pgdata', abs_data_path,
+            '-U', USER,
+            "-o '-p {}'".format(NODE_PORTS[node_name]),
```
thanodnl: nit; this will require a format, but can we separate the flag to `pg_ctl` from the flag that will be passed to the `postgres` process? line would become:
```python
    '-o', '-p {}'.format(...)
```

This makes it so the second part will be correctly encapsulated by the os instead of our encapsulation (which causes the mix of `"` and `'`). I am not 100% sure this works as intended, but can you have a look?

src/test/regress/upgrade/upgrade_test.py
```diff
@@ -0,0 +1,179 @@
...
+
+def start_databases(pg_path, rel_data_path):
+    for node_name in NODE_NAMES:
+        abs_data_path = os.path.abspath(os.path.join(rel_data_path, node_name))
+        command = [
+            os.path.join(pg_path, 'pg_ctl'), 'start',
+            '--pgdata', abs_data_path,
+            '-U', USER,
+            "-o '-p {}'".format(NODE_PORTS[node_name]),
+            '--log', os.path.join(pg_path, 'logfile_' + node_name)
```
thanodnl: can we put these in the `abs_data_path`? I'd rather not have them in the `pg_path`.
Funfact is, I think they are ending up in my working directory at the moment.

src/test/regress/upgrade/config.py
```diff
@@ -0,0 +1,32 @@
...
+            'shared_preload_libraries': 'citus',
+            'citus.node_conninfo': 'prefer'
```
thanodnl: ⚠️ This needs to be changed to fix sslmode, after that its ready to merge.

```python
  'citus.node_conninfo': 'sslmode=prefer'
```

The `node_conninfo` guc consists of key/value pairs that are parts of postrges' connection string: https://www.postgresql.org/docs/10/libpq-connect.html#LIBPQ-CONNSTRING

src/test/regress/upgrade/upgrade_test.py
SaitTalhaNisanci: Yeah what you wrote works.

SaitTalhaNisanci:
> Code and everything is fine for me, but I can't run the tests locally on my machine till completion.
> Still having some issues with ssl, did you compile postgres `--with-openssl`? Would be great if you could reproduce on your machine :). After the upgrade my coordinator is not able to talk to the workers.
> 
> I have marked the line which needs to be changed, I have verified locally that I can finish the test successfully with that change.

Yeah I have received the same error with openssl and updated the code afterwards, now it should work as you said.

PR #2942, opened 2019-09-10
Make sure that lost connections are handled properly in adaptive executor

onderkalaci:
Before this patch, when a connection is lost, we'd have the following
situation:

    - Pop a task execution from readyQueue
    - Lost connection
    - Fail the session/pool. -> This step was not acting properly
      because we've popped the task, but not set to session->currentTask
      yet

After the patch:

    - Pop a task execution from readyQueue
    - Immediately set it to session->currentTask
    - Lost connection
    - Fail the session/pool. -> At this step, failing the
      session would trigger query failures (or failovers)
      properly.

DESCRIPTION: Fixes a bug that leads to various issues when a connection is lost

### DIFF
--- a/src/backend/distributed/executor/adaptive_executor.c
+++ b/src/backend/distributed/executor/adaptive_executor.c
@@ -2412,6 +2412,7 @@ TransactionStateMachine(WorkerSession *session)
 				else
 				{
 					TaskPlacementExecution *placementExecution = NULL;
+					bool placementExecutionStarted = false;
 
 					placementExecution = PopPlacementExecution(session);
 					if (placementExecution == NULL)
@@ -2425,7 +2426,17 @@ TransactionStateMachine(WorkerSession *session)
 						break;
 					}
 
-					StartPlacementExecutionOnSession(placementExecution, session);
+					placementExecutionStarted =
+						StartPlacementExecutionOnSession(placementExecution, session);
+					if (!placementExecutionStarted)
+					{
+						/* no need to continue, connection is lost */
+						Assert(session->connection->connectionState ==
+							   MULTI_CONNECTION_LOST);
+
+						return;
+					}
+
 					transaction->transactionState = REMOTE_TRANS_SENT_COMMAND;
 				}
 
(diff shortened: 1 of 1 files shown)


src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -2748,6 +2768,8 @@ StartPlacementExecutionOnSession(TaskPlacementExecution *placementExecution,
 
 	/* connection is going to be in use */
 	workerPool->idleConnectionCount--;
+	session->currentTask = placementExecution;
+	placementExecution->executionState = PLACEMENT_EXECUTION_RUNNING;
```
onderkalaci: Adding test to this PR is extremely hard. Even manually triggering the situation is very hard -- I wasn't able to do yet.

onderkalaci: I mean, I can only do it with code changes:
```C
index ae3f23ae0..560f5b8e6 100644
--- a/src/backend/distributed/executor/adaptive_executor.c
+++ b/src/backend/distributed/executor/adaptive_executor.c
@@ -1799,7 +1799,7 @@ ManageWorkerPool(WorkerPool *workerPool)
        Assert(initiatedConnectionCount >= activeConnectionCount);
 
        /* we should never have less than 0 connections ever */
-       Assert(activeConnectionCount >= 0 && idleConnectionCount >= 0);
+       //Assert(activeConnectionCount >= 0 && idleConnectionCount >= 0);
 
        if (workerPool->failed)
        {
@@ -2679,6 +2679,7 @@ PopAssignedPlacementExecution(WorkerSession *session)
 }
 
 
+
 /*
  * PopAssignedPlacementExecution finds an executable task from the queue of assigned tasks.
  */
@@ -2749,6 +2750,7 @@ StartPlacementExecutionOnSession(TaskPlacementExecution *placementExecution,
        /* connection is going to be in use */
        workerPool->idleConnectionCount--;
 
+       ShutdownConnection(connection);
        if (paramListInfo != NULL)
        {
                int parameterCount = paramListInfo->numParams;
```

onderkalaci:
> Let's open an issue to add a failure test that covers the return paths in StartPlacementExecutionOnSession.

Opened #2943

PR #2944, opened 2019-09-11
Use variables instead of hardcoded tmp dirs

SaitTalhaNisanci:
This PR refactors some hard coded temporary directories.


SaitTalhaNisanci:
@marcocitus  I have updated the error messages too.

PR #2951, opened 2019-09-11
Change SQL migration build process for easier reviews

JelteF:
DESCRIPTION: Change SQL migration build process for easier reviews

@thanodnl told me it was a bit of a problem that it's impossible to see the history of a UDF in git. The only way to do so is by reading all the sql migration files from new to old. Another problem is that it's also hard to review the changed UDF during code review, because to find out what changed you have to do the same. I thought of a IMHO better (but not perfect) way to handle this.

We keep the definition of a UDF in `sql/udfs/{name_of_udf}/latest.sql`. That file we change whenever we need to make a change to the the UDF. On top of that you also make a snapshot of the file in `sql/udfs/{name_of_udf}/{migration-version}.sql` (e.g. `8.4-1.sql`) by copying the contents. This way you can easily view what the actual changes where by looking at the `latest.sql` file.

There's still the question on how to use these files then. Sadly postgres doesn't allow inclusion of other sql files in the migration sql file (it does in psql using `\i`). So instead I used the C preprocessor+ `make` to compile a `sql/xxx.sql` to a `build/sql/xxx.sql` file. This final `build/sql/xxx.sql` file has every occurence of `#include "somefile.sql"` in `sql/xxx.sql` replaced by the contents of `somefile.sql`.


.gitattributes
```diff
@@ -30,3 +30,6 @@ src/backend/distributed/utils/ruleutils_10.c -citus-style
 src/backend/distributed/utils/ruleutils_11.c -citus-style
 src/backend/distributed/utils/ruleutils_12.c -citus-style
 src/include/distributed/citus_nodes.h -citus-style
+
+# Hide diff on github by default for copied udfs
+src/backend/distributed/udfs/*/[123456789]*.sql linguist-generated=true
```
thanodnl: This doesn't seem to have the desired effect.

The format is `x.y-z--x.y-z.sql` I guess, you are only matching numbers it seems.

src/test/regress/expected/multi_unsupported_worker_operations.out
```diff
@@ -269,7 +269,7 @@ DROP TABLE mx_table;
 ERROR:  operation is not allowed on this node
 HINT:  Connect to the coordinator and run it again.
 CONTEXT:  SQL statement "SELECT master_remove_distributed_table_metadata_from_workers(v_obj.objid, v_obj.schema_name, v_obj.object_name)"
-PL/pgSQL function citus_drop_trigger() line 19 at PERFORM
+PL/pgSQL function citus_drop_trigger() line 18 at PERFORM
```
thanodnl: fascinating, how did this happen?

.gitattributes
serprex: The * is a blob, so this matches any file starting with a digit

thanodnl: I mean `x.y-z.sql` anyway, I see the diff in github :)

src/test/regress/expected/multi_unsupported_worker_operations.out
JelteF: empty new lines and lines with c comments are stripped out now by the preprocessor. But before they were counted by postgres for the line count.

.gitattributes
JelteF: it's not regex but git matching, so it matches all sql files where the first character is a number. I didn't update the path to `.../sql/udfs/...` though so that's why the matching is not working

src/backend/distributed/sql/udfs/master_add_secondary_node/latest.sql
```diff
@@ -0,0 +1,23 @@
...
+                                          OUT noderole noderole,
+                                          OUT nodecluster name)
+  RETURNS record
+  LANGUAGE C STRICT
+  AS 'MODULE_PATHNAME', $$master_add_secondary_node$$;
+COMMENT ON FUNCTION master_add_secondary_node(nodename text, nodeport integer,
+                                              primaryname text, primaryport integer,
+                                              nodecluster name)
+  IS 'add a secondary node to the cluster';
+REVOKE ALL ON FUNCTION master_add_secondary_node(text,int,text,int,name) FROM PUBLIC;
```
marcocitus: What happens to the search_path in these files?

CONTRIBUTING.md
```diff
@@ -136,3 +136,39 @@ Once you've done that, you can run the `citus_indent` command to recursively che
...
+this function, if not create one. Then change or create the file called
+`latest.sql` in that directory to match how it should create the function. This
+should be including any DROP (IF EXISTS), COMMENT and REVOKE statements for this
+function.
+
+Then copy the `latest.sql` file to `{version}.sql`, where `{version}` is the
+version for which this sql change is, e.g. `{9.0-1.sql}`. Now that you've
+created this stable snapshot of the function definition for your version you
+should use it in your actual sql file, .e.g.
+`src/backend/distributed/sql/citus--8.3-1--9.0-1.sql`. You do this by using C
```
onderkalaci: I think I need a clarification on this. When I look at the mentioned [file](https://github.com/citusdata/citus/blob/master/src/backend/distributed/sql/citus--8.3-1--9.0-1.sql), I see `worker_create_or_replace_object ()`. 

For example, why is that not included via a header and explicitly added? (Or may other function (re)-definitions on the file)

marcocitus: Mostly because I pushed back on this a bit. I think there are some risks to liberally separating UDFs into separate SQL files such as bugs around permissions, search paths, and ambiguity around how to declare types that are used in a function argument. We should form a more comprehensive policy around how we do SQL scripts.

On the other hand, the lack of history tracking for PL/pgSQL has been very annoying and this release introduces lengthy upgrade scripts. It seems better not to wait with those since they are likely to change in the near future.

PR #2952, opened 2019-09-11
Refactor 9 argument function to use a struct

JelteF:
DESCRIPTION: Refactor 9 argument function to use a struct

For another PR I needed to add another column which would require to add another argument to an already 9 argument function signature. In this case it would be a boolean flag and there were already two boolean flags in there. In my experience it becomes really easy to mess up the order of these flags at that point. Especially because the type system doesn't distinguish between the 3 different booleans with completely different meanings.

So I refactored these signatures to receive a struct containing most of these arguments. Like that you don't mess up orderening, because the meaning of the boolean is not order dependent but fieldname dependent. It also makes it possible to set good shared defaults for this struct.

### DIFF
--- a/src/backend/distributed/utils/node_metadata.c
+++ b/src/backend/distributed/utils/node_metadata.c
@@ -53,19 +53,27 @@ int GroupSize = 1;
 /* config variable managed via guc.c */
 char *CurrentCluster = "default";
 
+typedef struct NodeMetadata
+{
+	int32 groupId;
+	char *nodeRack;
+	bool hasMetadata;
+	bool isActive;
+	Oid nodeRole;
+	char *nodeCluster;
+} NodeMetadata;
+
 /* local function forward declarations */
 static int ActivateNode(char *nodeName, int nodePort);
 static void RemoveNodeFromCluster(char *nodeName, int32 nodePort);
-static int AddNodeMetadata(char *nodeName, int32 nodePort, int32 groupId,
-						   char *nodeRack, bool hasMetadata, bool isActive,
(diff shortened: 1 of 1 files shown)


src/backend/distributed/utils/node_metadata.c
```diff
@@ -84,6 +92,17 @@ PG_FUNCTION_INFO_V1(master_initialize_node_metadata);
 PG_FUNCTION_INFO_V1(get_shard_id_for_distribution_column);
 
 
+static NodeMetadata
+DefaultNodeMetadata()
+{
+	NodeMetadata nodeMetadata = {
+		.nodeRole = InvalidOid,
+		.nodeRack = WORKER_DEFAULT_RACK,
```
marcocitus: can we fill in the rest here?

JelteF: The rest is all set to 0/NULL by default when initializing like this: https://stackoverflow.com/a/330867/2570866

Do you want me to make those 0/NULL cases explicit?

marcocitus: Then .nodeRole can be left out.

src/backend/distributed/utils/node_metadata.c
```diff
@@ -84,6 +92,17 @@ PG_FUNCTION_INFO_V1(master_initialize_node_metadata);
 PG_FUNCTION_INFO_V1(get_shard_id_for_distribution_column);
 
 
+static NodeMetadata
+DefaultNodeMetadata()
```
marcocitus: Needs a function comment.

PR #2960, opened 2019-09-12
Add shouldhaveshards to pg_dist_node

JelteF:
DESCRIPTION: Add shouldhaveshards to pg_dist_node

This is an improvement over #2512.

This adds the boolean `shouldhaveshards` column to `pg_dist_node`. When it's `false`, create_distributed_table for new collocation groups will not create shards on that node. Reference tables will still be created on nodes where it is `false`.


src/backend/distributed/utils/node_metadata.c
```diff
@@ -924,7 +962,7 @@ RemoveNodeFromCluster(char *nodeName, int32 nodePort)
 
 	DeleteNodeRow(nodeName, nodePort);
 
-	if (WorkerNodeIsPrimary(workerNode))
+	if (WorkerNodeIsPrimaryCurrentDataNode(workerNode))
 	{
 		UpdateColocationGroupReplicationFactorForReferenceTables(
 			ActivePrimaryNodeCount());
```
thanodnl: It feels like this count needs to be reflecting the change in how we decide which nodes have a reference table on them.

src/test/regress/expected/multi_cluster_management.out
```diff
@@ -676,8 +676,122 @@ SELECT master_update_node(:worker_1_node, 'localhost', :worker_1_port);
...
+SELECT nodeport, count(*)
+FROM pg_dist_shard JOIN pg_dist_shard_placement USING (shardid)
+WHERE logicalrelid = 'test_dist_colocated'::regclass GROUP BY nodeport;
+ nodeport | count 
+----------+-------
+    57637 |     2
+    57638 |     2
+(2 rows)
+
+-- non colocated tables should still be placed on 'marked for removal' nodes for safety
```
thanodnl: comment seems to be copied from above.
I would expect a non colocated table to _not_ be placed on `marked for removal` nodes.

src/test/regress/expected/multi_cluster_management.out
```diff
@@ -676,8 +676,122 @@ SELECT master_update_node(:worker_1_node, 'localhost', :worker_1_port);
...
+SELECT nodeport, count(*)
+FROM pg_dist_shard JOIN pg_dist_shard_placement USING (shardid)
+WHERE logicalrelid = 'test_dist_non_colocated'::regclass GROUP BY nodeport;
+ nodeport | count 
+----------+-------
+    57637 |     4
+(1 row)
+
+-- tihs table should be colocated with the test_dist_non_colocated table
```
thanodnl: nit: s/tihs/this/

src/test/regress/expected/multi_cluster_management.out
```diff
@@ -676,8 +676,122 @@ SELECT master_update_node(:worker_1_node, 'localhost', :worker_1_port);
...
+-- reference tables should be placed on 'marked for removal' nodes for safety
+SELECT nodeport, count(*)
+FROM pg_dist_shard JOIN pg_dist_shard_placement USING (shardid)
+WHERE logicalrelid = 'test_ref'::regclass GROUP BY nodeport;
+ nodeport | count 
+----------+-------
+    57637 |     1
+    57638 |     1
+(2 rows)
+
```
thanodnl: Can we add a test here that tries to set `isdatanode` to `false`?

We need to make sure a trigger refuses to do so as long as there are shards of distributed tables on a node. It should allow to change to `false` if the only placements are from reference tables. Although this would require an update to the active placements of the reference table to not be on this node anymore. future reference tables should not be placed there either anymore as tested below.

src/test/regress/expected/multi_metadata_attributes.out
```diff
@@ -1,15 +1,14 @@
...
 -- some heap_getattr() calls to heap_deform_tuple(). This errors out in
--- postgres versions before 11.
+-- postgres versions before 11. If this test fails check out
+-- https://github.com/citusdata/citus/pull/2464 for an explanation of what to
+-- do. Once you used the new code for the table you can add it to the NOT IN
+-- part of the query so new changes to it won't affect this test.
 SELECT attrelid::regclass, attname, atthasmissing, attmissingval
 FROM pg_attribute
-WHERE atthasmissing
+WHERE atthasmissing AND attrelid NOT IN ('pg_dist_node'::regclass)
```
thanodnl: I have a though time understanding if we can safely do this.
The query got added for pg11 and new non-null columns to our catalogs.
I guess our new non-null value for `isdatanode` changed the output of this test.

We are now not only hiding that column, but any future changes to `pg_dist_node`. Is that allowed because we have already added all required deform calls for the use of `pg_dist_node` in #2464 and therefore don't need to see any changes to this table anymore?

src/test/regress/expected/multi_mx_hide_shard_names.out
```diff
@@ -139,6 +139,7 @@ SELECT pg_table_is_visible('test_table_1130000'::regclass);
 SET search_path TO 'mx_hide_shard_names';
 SET citus.shard_count TO 4;
 SET citus.shard_replication_factor TO 1;
+SET citus.next_shard_id TO 1131000;
```
thanodnl: :+1:

src/test/regress/expected/multi_read_from_secondaries.out
```diff
@@ -24,10 +24,10 @@ INSERT INTO dest_table (a, b) VALUES (2, 1);
 INSERT INTO source_table (a, b) VALUES (10, 10);
 -- simluate actually having secondary nodes
 SELECT * FROM pg_dist_node;
- nodeid | groupid | nodename  | nodeport | noderack | hasmetadata | isactive | noderole | nodecluster 
---------+---------+-----------+----------+----------+-------------+----------+----------+-------------
-      1 |       1 | localhost |    57637 | default  | f           | t        | primary  | default
-      2 |       2 | localhost |    57638 | default  | f           | t        | primary  | default
+ nodeid | groupid | nodename  | nodeport | noderack | hasmetadata | isactive | noderole | nodecluster | isdatanode
```
thanodnl: this test needs to be updated anytime we change `pg_dist_node`, can we instead only select fields relevant for this test.

src/backend/distributed/master/worker_node_manager.c
```diff
@@ -355,6 +356,73 @@ ActivePrimaryNodeList(LOCKMODE lockMode)
 }
 
 
+/*
+ * ActivePrimaryDesiredDataNodeList returns a list of all active, primary
+ * worker nodes that can store new data, i.e isdatanode is 'true'.
+ */
+List *
+ActivePrimaryDesiredDataNodeList(LOCKMODE lockMode)
```
thanodnl: This function and the function below share 99% if the code except for a function call to either `WorkerNodeIsPrimaryDesiredDataNode` and `WorkerNodeIsPrimaryCurrentDataNode`.

Any way we can share more of the code instead of copying?

src/test/regress/expected/multi_metadata_attributes.out
JelteF: As far as I understood the answer to your last question is yes. i.e. this test only exists to notify us when we add a column to tables that we haven't added the deform calls to yet. @pykello is that indeed the case?

src/backend/distributed/utils/node_metadata.c
```diff
@@ -264,13 +274,150 @@ master_remove_node(PG_FUNCTION_ARGS)
...
+
+
+/*
+ * ModifiableWorkerNode gets the requested WorkerNode and also gets locks
+ * required for modifying it. This fails if the node does not exist.
+ */
+static WorkerNode *
+ModifiableWorkerNode(text *nodeNameText, int32 nodePort)
+{
+	char *nodeName = text_to_cstring(nodeNameText);
```
thanodnl: as said earlier, use `char *`'s to pass the name instead of this late postgres to c translation.

src/backend/distributed/utils/node_metadata.c
```diff
@@ -922,53 +1100,19 @@ ReadWorkerNodes(bool includeNodesFromOtherClusters)
  * replication factor will be equal to worker count.
  */
 static void
-RemoveNodeFromCluster(char *nodeName, int32 nodePort)
+RemoveNodeFromCluster(text *nodeNameText, int32 nodePort)
```
thanodnl: same here, we generally pass `char *` instead of `text`.

src/backend/distributed/utils/node_metadata.c
```diff
@@ -1128,13 +1280,49 @@ SetNodeState(char *nodeName, int32 nodePort, bool isActive)
...
+}
+
+
+/*
+ * SetIsDataNode function sets the isdatanode column of the specified worker in
+ * pg_dist_node.
+ * It returns the new worker node after the modification.
+ */
+static WorkerNode *
+SetIsDataNode(WorkerNode *workerNode, Oid isDataNode)
```
marcocitus: the name suggests isDataNode is supposed to be a boolean

src/backend/distributed/sql/citus--8.3-1--9.0-1.sql
```diff
@@ -170,4 +170,46 @@ CREATE FUNCTION worker_apply_sequence_command(create_sequence_command text,
 COMMENT ON FUNCTION worker_apply_sequence_command(text,regtype)
     IS 'create a sequence which produces globally unique values';
 
+CREATE TYPE isdatanode AS ENUM('true', 'marked for draining', 'false');
```
marcocitus: isdatanode seems like a strange name for a type

maybe 'marked for draining' -> 'draining'.

src/backend/distributed/sql/citus--8.3-1--9.0-1.sql
```diff
@@ -170,4 +170,46 @@ CREATE FUNCTION worker_apply_sequence_command(create_sequence_command text,
...
+COMMENT ON TYPE isdatanode IS
+    'enum that contains the different possible states of pg_dist_node.isdatanode';
+ALTER TABLE pg_dist_node ADD isdatanode isdatanode NOT NULL DEFAULT 'true';
```
marcocitus: isdatanode suggests this should be a boolean.

src/backend/distributed/utils/node_metadata.c
```diff
@@ -1088,38 +1240,38 @@ AddNodeMetadata(char *nodeName, int32 nodePort,
...
- * SetNodeState function sets the isactive column of the specified worker in
- * pg_dist_node to isActive.
+ * SetIsDataNode function sets the column with the specified index on the
+ * worker in pg_dist_node.
+ * It returns the new worker node after the modification.
  */
-static void
-SetNodeState(char *nodeName, int32 nodePort, bool isActive)
+static WorkerNode *
+SetWorkerColumn(WorkerNode *workerNode, int columnIndex, Datum value, char *updateCommand)
```
marcocitus: I read the function signature and comment several times, but couldn't understand what this does. It seems more like a hack to save a few lines of code rather than a useful abstraction that helps make the code more maintainable.

Maybe something like: 
```c
static void
UpdateSingleNodeProperty(int nodeId, NodeProperty property, Datum newValue)
{
    ... 
    switch (property)
    {
        case NODE_PROPERTY_IS_DATA_NODE:
        {
            columnIndex = Anum_pg_dist_node_isdatanode;
            metadataSyncCommand = IsDataNodeUpdateCommand(nodeId, DatumGetObjectId(newValue));
            break;
        }
        ....
}

```

src/backend/distributed/utils/node_metadata.c
```diff
@@ -286,30 +439,41 @@ master_disable_node(PG_FUNCTION_ARGS)
...
+ * ModifiableWorkerNode gets the requested WorkerNode and also gets locks
+ * required for modifying it. This fails if the node does not exist.
+ *
+ * It also fails if the node is a primary with active distributed tables. To
+ * check this it removes all the reference table placements from the node.
+ * SO ONLY USE THIS when reference tables should be deleted from the node after
+ * a succesful transaction.
+ */
+static WorkerNode *
+ModifiableWorkerNodeWithoutPrimaryPlacements(const char *nodeName, int32 nodePort)
```
marcocitus: Not very clear to me what this function does.

src/backend/distributed/utils/node_metadata.c
```diff
@@ -286,30 +439,41 @@ master_disable_node(PG_FUNCTION_ARGS)
...
-								"SELECT master_activate_node('%s', %d) to activate this "
-								"node back.",
-								nodeName, nodePort, nodeName, nodePort)));
+		return workerNode;
 	}
 
-	SetNodeState(nodeName, nodePort, isActive);
+	/* Delete refrence table placements so they are not taken into account for
+	 * the check if there are placements after this */
+	DeleteAllReferenceTablePlacementsFromNodeGroup(workerNode->groupId);
```
marcocitus: Should this really be here? This function seems to be somewhat random.

src/backend/distributed/sql/citus--8.3-1--9.0-1.sql
marcocitus: Do we really need 3 states? It seems to complicate the code a lot. Draining could also be (isdatanode=false, placements exist).

src/backend/distributed/commands/create_distributed_table.c
```diff
@@ -309,7 +309,7 @@ create_reference_table(PG_FUNCTION_ARGS)
 	 */
 	EnsureRelationKindSupported(relationId);
 
-	workerNodeList = ActivePrimaryNodeList(ShareLock);
+	workerNodeList = ActivePrimaryCurrentDataNodeList(ShareLock);
```
marcocitus: I'm wondering whether isdatanode should really apply to reference tables. We can remove a node at any time once it no longer has shards. We'll soon start replicating the reference table to the coordinator (by adding coordinator to pg_dist_node), coordinator should not be marked as a data node. We also have the notion of active/inactive.

To clear up confusion, perhaps the column should be called `hasshards`?

marcocitus:
I think making this a `shouldhaveshards bool` column would significantly improve things, since that avoids ambiguity e.g. around whether or not a node should have reference tables, or currently has shards, but shouldn't.

src/test/regress/expected/multi_metadata_attributes.out
pykello: This should be safe since the only place in the code that should convert a tuple to a worker node is `TupleToWorkerNode` and other places should just read the cached version. Can you please double check that there is no other place that reads from `pg_dist_node` tuples before merging?

JelteF: There's no other place that reads from pg_dist_node

src/backend/distributed/utils/node_metadata.c
```diff
@@ -264,52 +273,160 @@ master_remove_node(PG_FUNCTION_ARGS)
...
+ */
+static void
+CleanUpReferenceTables(WorkerNode *workerNode)
+{
 	if (WorkerNodeIsPrimary(workerNode))
 	{
-		DeleteAllReferenceTablePlacementsFromNodeGroup(workerNode->groupId);
+		UpdateColocationGroupReplicationFactorForReferenceTables(
+			ActivePrimaryNodeCount());
 	}
```
marcocitus: The "CleanUpReferenceTables" name does not help me understand what this function does. Maybe just inline this code?

src/backend/distributed/utils/node_metadata.c
```diff
@@ -264,52 +273,160 @@ master_remove_node(PG_FUNCTION_ARGS)
...
+	PG_RETURN_VOID();
+}
+
+
+/*
+ * master_make_data_node sets the shouldhaveshards column in pg_dist_node to true,
+ * i.e new distributed tables will get placements on this node.
+ */
+Datum
+master_make_data_node(PG_FUNCTION_ARGS)
```
marcocitus: Do we need this UDF? It seems redundant and is not covered by tests.

src/test/regress/expected/multi_cluster_management.out
```diff
@@ -691,8 +691,196 @@ SELECT master_update_node(:worker_1_node, 'localhost', :worker_1_port);
...
+-- non colocated tables should be placed on nodedes that were switched to
+-- shouldhaveshards true
+SELECT nodeport, count(*)
+FROM pg_dist_shard JOIN pg_dist_shard_placement USING (shardid)
+WHERE logicalrelid = 'test_dist_non_colocated'::regclass GROUP BY nodeport;
+ nodeport | count 
+----------+-------
+    57637 |     2
+    57638 |     2
```
marcocitus: should have an ORDER BY nodeport

src/backend/distributed/utils/node_metadata.c
```diff
@@ -264,52 +273,160 @@ master_remove_node(PG_FUNCTION_ARGS)
...
+
+	if (strcmp(text_to_cstring(propertyText), "shouldhaveshards") == 0)
 	{
-		ereport(ERROR, (errmsg("node at \"%s:%u\" does not exist", nodeName, nodePort)));
+		SetShouldHaveShards(workerNode, value);
 	}
+	else
+	{
+		ereport(ERROR, (errmsg(
+							"It's only supported to set 'shouldhaveshards' using this function"
```
marcocitus: This sentence is a little strange.
"only the 'shouldhaveshards' property can be set using this function"

src/backend/distributed/utils/node_metadata.c
```diff
@@ -264,52 +273,160 @@ master_remove_node(PG_FUNCTION_ARGS)
...
-
 	char *nodeName = text_to_cstring(nodeNameText);
+	WorkerNode *workerNode = ModifiableWorkerNode(nodeName, nodePort);
 	bool isActive = false;
+	bool onlyConsiderActivePlacements = false;
 
-	WorkerNode *workerNode = NULL;
+	if (WorkerNodeIsPrimary(workerNode))
+	{
+		/* Delete refrence table placements so they are not taken into account for
```
marcocitus: refrence -> reference.

Our usual commenting format is
```c
/*
 * Delete ..
 * the ..
 */
```

src/backend/distributed/utils/node_metadata.c
```diff
@@ -264,52 +273,160 @@ master_remove_node(PG_FUNCTION_ARGS)
...
+		 * the check if there are placements after this */
+		DeleteAllReferenceTablePlacementsFromNodeGroup(workerNode->groupId);
```
marcocitus: I kind of wonder why we bother doing this instead of just leaving some stale copies around that will be replaced on re-activation, but that's an existing issue.

src/backend/distributed/utils/node_metadata.c
```diff
@@ -1296,6 +1411,8 @@ InsertNodeRow(int nodeid, char *nodeName, int32 nodePort, NodeMetadata *nodeMeta
 	values[Anum_pg_dist_node_isactive - 1] = BoolGetDatum(nodeMetadata->isActive);
 	values[Anum_pg_dist_node_noderole - 1] = ObjectIdGetDatum(nodeMetadata->nodeRole);
 	values[Anum_pg_dist_node_nodecluster - 1] = nodeClusterNameDatum;
+	values[Anum_pg_dist_node_shouldhaveshards - 1] = ObjectIdGetDatum(
```
marcocitus: Should be Bool

src/backend/distributed/utils/node_metadata.c
```diff
@@ -1562,6 +1679,9 @@ TupleToWorkerNode(TupleDesc tupleDescriptor, HeapTuple heapTuple)
 		DatumGetBool(datumArray[Anum_pg_dist_node_metadatasynced - 1]);
 	workerNode->isActive = DatumGetBool(datumArray[Anum_pg_dist_node_isactive - 1]);
 	workerNode->nodeRole = DatumGetObjectId(datumArray[Anum_pg_dist_node_noderole - 1]);
+	workerNode->shouldHaveShards = DatumGetObjectId(
```
marcocitus: Should be Bool

src/backend/distributed/utils/node_metadata.c
marcocitus: not resolved

JelteF: Seems that comment was in two places and I only changed it in one of them.

src/backend/distributed/utils/node_metadata.c
marcocitus: Not yet fixed.

src/backend/distributed/utils/node_metadata.c
```diff
@@ -1088,38 +1154,62 @@ AddNodeMetadata(char *nodeName, int32 nodePort,
...
+	char *metadataSyncCommand = NULL;
 
-	char *nodeStateUpdateCommand = NULL;
-	WorkerNode *workerNode = NULL;
+
+	switch (columnIndex)
+	{
+		case Anum_pg_dist_node_isactive:
+		{
+			metadataSyncCommand = ShouldHaveShardsUpdateCommand(workerNode->nodeId,
```
pykello: Probably a typo. I am addressing this issue in https://github.com/citusdata/citus/pull/3164.

src/backend/distributed/utils/node_metadata.c
```diff
@@ -1088,38 +1154,62 @@ AddNodeMetadata(char *nodeName, int32 nodePort,
...
+		case Anum_pg_dist_node_shouldhaveshards:
+		{
+			metadataSyncCommand = ShouldHaveShardsUpdateCommand(workerNode->nodeId,
+																DatumGetBool(value));
+			break;
+		}
+
+		default:
+		{
+			ereport(ERROR, (errmsg("could not find valid entry for node \"%s:%d\"",
```
pykello: This error seems strange for the invalid column index. I think this should be probably replaced with an `Assert(false)`, since this case seems a programming error and should be unreachable.

src/backend/distributed/utils/node_metadata.c
JelteF: Very weird that the metadatasync tests from this PR didn't catch that then. Testing that this sync command is working is the exact reason why I added them.

PR #2965, opened 2019-09-13
Distribute CALL

### LINKED ISSUE #2945: Add a cache for efficiently looking up pg_dist_object rows
Once we add procedures to pg_dist_object we need to be able to efficiently determine whether a function is distributed and what its additional properties are (distribution argument index, colocation ID). This could also be useful in the future to replicate other objects on demand if they appear in queries.

It appears we can reuse the catalog cache by initialising it with a magic number that is not used for any of the Postgres catalog tables:
https://github.com/citusdata/citus/pull/2829/files#diff-801332ee4319b28e7065fb6c9c32154eR389

### LINKED ISSUE #2946: Delegate call commands to worker nodes
We should add a distribution argument (e.g. tenant_id) and an co-location ID to `pg_dist_object` to identify distributed procedures.

When a "distributed" procedure is called through CALL and the data for the distribution argument value is on a worker node that has up-to-date metadata, we should forward it to the worker node.

An example of handling the CALL statement is at:
https://github.com/citusdata/citus/pull/2829/files#diff-82b34f694db38c26921dbd5d12890974R36

The prototype does not return the results of the CALL. We should probably build a task list and run it through the adaptive executor and then emit the tuples into a `DestReceiver`. We should also add additional checks for strings that cannot be pushed down (e.g. a call statement can contain a variable because it is called from a different procedure, though that may be covered by not delegating calls inside a multi-statement transaction).

Ideally, we would refactor the adaptive executor to emit directly into a DestReceiver, and use a `CreateTuplestoreDestReceiver()` when writing to a tuple store. An alternative would be to iterate over the tuple store in `CallFuncExprRemotely`.

serprex:
DESCRIPTION: implement infrastructure for routing `CALL` to MX workers

Mostly copied code from #2829 except it uses metadata_cache & pg_dist_object & etc

Closes #2945 
Closes #2946

### DIFF
--- a/src/backend/distributed/commands/call.c
+++ b/src/backend/distributed/commands/call.c
@@ -39,7 +39,7 @@ static bool CallFuncExprRemotely(CallStmt *callStmt,
 								 DestReceiver *dest);
 
 /*
- * CallDistributedProcedure calls a stored procedure on the worker if possible.
+ * CallDistributedProcedureRemotely calls a stored procedure on the worker if possible.
  */
 bool
 CallDistributedProcedureRemotely(CallStmt *callStmt, const char *queryString,
@@ -71,7 +71,6 @@ CallFuncExprRemotely(CallStmt *callStmt, DistObjectCacheEntry *procedure,
 	ShardInterval *shardInterval = NULL;
 	List *placementList = NIL;
 	ListCell *argCell = NULL;
-	WorkerNode *preferredWorkerNode = NULL;
 	DistTableCacheEntry *distTable = NULL;
 	ShardPlacement *placement = NULL;
 	WorkerNode *workerNode = NULL;
@@ -128,13 +127,7 @@ CallFuncExprRemotely(CallStmt *callStmt, DistObjectCacheEntry *procedure,
 	placement = (ShardPlacement *) linitial(placementList);
 	workerNode = FindWorkerNode(placement->nodeName, placement->nodePort);
 
-	if (workerNode->hasMetadata)
-	{
-		/* we can execute this procedure on the worker! */
-		preferredWorkerNode = workerNode;
-	}
-
-	if (preferredWorkerNode == NULL)
+	if (workerNode == NULL || !workerNode->hasMetadata || !workerNode->metadataSynced)
 	{
 		ereport(DEBUG2, (errmsg("there is no worker node with metadata")));
 		return false;
--- a/src/backend/distributed/commands/call.c
+++ b/src/backend/distributed/commands/call.c
@@ -1,7 +1,9 @@
 /*-------------------------------------------------------------------------
  *
  * call.c
- *    Commands for call remote stored procedures.
+ *    Commands for distributing CALL for distributed procedures.
+ *
+ *    Procedures can be distributed with create_distributed_function.
  *
  * Copyright (c) 2019, Citus Data, Inc.
  *
@@ -169,6 +171,7 @@ CallFuncExprRemotely(CallStmt *callStmt, DistObjectCacheEntry *procedure,
 		TupleDesc tupleDesc = CallStmtResultDesc(callStmt);
 		TupleTableSlot *slot = MakeSingleTupleTableSlotCompat(tupleDesc,
 															  &TTSOpsMinimalTuple);
+		bool hasReturning = true;
 		Task *task = CitusMakeNode(Task);
 
 		task->jobId = INVALID_JOB_ID;
@@ -182,7 +185,7 @@ CallFuncExprRemotely(CallStmt *callStmt, DistObjectCacheEntry *procedure,
 		task->taskPlacementList = placementList;
 
 		ExecuteTaskListExtended(ROW_MODIFY_NONE, list_make1(task),
-								tupleDesc, tupleStore, true,
+								tupleDesc, tupleStore, hasReturning,
 								MaxAdaptiveExecutorPoolSize);
 
 		while (tuplestore_gettupleslot(tupleStore, true, false, slot))
--- a/src/backend/distributed/commands/call.c
+++ b/src/backend/distributed/commands/call.c
@@ -14,6 +14,7 @@
 
 #include "catalog/pg_proc.h"
 #include "commands/defrem.h"
+#include "distributed/citus_ruleutils.h"
 #include "distributed/colocation_utils.h"
 #include "distributed/commands.h"
 #include "distributed/commands/multi_copy.h"
@@ -38,15 +39,13 @@
 
 static bool CallFuncExprRemotely(CallStmt *callStmt,
 								 DistObjectCacheEntry *procedure,
-								 FuncExpr *funcExpr, const char *queryString,
-								 DestReceiver *dest);
+								 FuncExpr *funcExpr, DestReceiver *dest);
 
 /*
  * CallDistributedProcedureRemotely calls a stored procedure on the worker if possible.
  */
 bool
-CallDistributedProcedureRemotely(CallStmt *callStmt, const char *queryString,
-								 DestReceiver *dest)
+CallDistributedProcedureRemotely(CallStmt *callStmt, DestReceiver *dest)
 {
 	DistObjectCacheEntry *procedure = NULL;
 	FuncExpr *funcExpr = callStmt->funcexpr;
@@ -58,7 +57,7 @@ CallDistributedProcedureRemotely(CallStmt *callStmt, const char *queryString,
 		return false;
 	}
 
-	return CallFuncExprRemotely(callStmt, procedure, funcExpr, queryString, dest);
+	return CallFuncExprRemotely(callStmt, procedure, funcExpr, dest);
 }
 
 
@@ -67,7 +66,7 @@ CallDistributedProcedureRemotely(CallStmt *callStmt, const char *queryString,
  */
 static bool
 CallFuncExprRemotely(CallStmt *callStmt, DistObjectCacheEntry *procedure,
-					 FuncExpr *funcExpr, const char *queryString, DestReceiver *dest)
+					 FuncExpr *funcExpr, DestReceiver *dest)
 {
 	Oid colocatedRelationId = InvalidOid;
 	Const *partitionValue = NULL;
@@ -78,6 +77,7 @@ CallFuncExprRemotely(CallStmt *callStmt, DistObjectCacheEntry *procedure,
 	Var *partitionColumn = NULL;
 	ShardPlacement *placement = NULL;
 	WorkerNode *workerNode = NULL;
+	StringInfo callCommand = NULL;
 
 	if (IsMultiStatementTransaction())
 	{
@@ -160,6 +160,10 @@ CallFuncExprRemotely(CallStmt *callStmt, DistObjectCacheEntry *procedure,
 
 	ereport(DEBUG1, (errmsg("pushing down the procedure")));
 
+	/* build remote command with fully qualified names */
+	callCommand = makeStringInfo();
+	appendStringInfo(callCommand, "CALL %s", pg_get_rule_expr((Node *) funcExpr));
+
 	{
 		Tuplestorestate *tupleStore = tuplestore_begin_heap(true, false, work_mem);
 		TupleDesc tupleDesc = CallStmtResultDesc(callStmt);
@@ -170,7 +174,7 @@ CallFuncExprRemotely(CallStmt *callStmt, DistObjectCacheEntry *procedure,
 		task->jobId = INVALID_JOB_ID;
 		task->taskId = 0;
 		task->taskType = DDL_TASK;
-		task->queryString = pstrdup(queryString);
+		task->queryString = callCommand->data;
 		task->replicationModel = REPLICATION_MODEL_INVALID;
 		task->dependedTaskList = NIL;
 		task->anchorShardId = placement->shardId;
--- a/src/backend/distributed/commands/call.c
+++ b/src/backend/distributed/commands/call.c
@@ -1,0 +1,181 @@
+/*-------------------------------------------------------------------------
+ *
+ * call.c
+ *    Commands for call remote stored procedures.
+ *
+ * Copyright (c) 2019, Citus Data, Inc.
+ *
+ *-------------------------------------------------------------------------
+ */
+
+#include "postgres.h"
+
+#if PG_VERSION_NUM >= 110000
+
+#include "catalog/pg_proc.h"
+#include "commands/defrem.h"
+#include "distributed/colocation_utils.h"
+#include "distributed/commands.h"
+#include "distributed/connection_management.h"
+#include "distributed/master_metadata_utility.h"
+#include "distributed/metadata_cache.h"
+#include "distributed/multi_executor.h"
+#include "distributed/multi_physical_planner.h"
+#include "distributed/remote_commands.h"
+#include "distributed/shard_pruning.h"
+#include "distributed/version_compat.h"
+#include "distributed/worker_manager.h"
+#include "nodes/nodeFuncs.h"
+#include "nodes/parsenodes.h"
+#include "nodes/primnodes.h"
+#include "miscadmin.h"
+#include "tcop/dest.h"
+#include "utils/lsyscache.h"
+#include "utils/syscache.h"
+
+static bool CallFuncExprRemotely(CallStmt *callStmt,
+								 DistObjectCacheEntry *procedure,
+								 FuncExpr *funcExpr, const char *queryString,
+								 DestReceiver *dest);
+
+/*
+ * CallDistributedProcedure calls a stored procedure on the worker if possible.
+ */
+bool
+CallDistributedProcedureRemotely(CallStmt *callStmt, const char *queryString,
+								 DestReceiver *dest)
+{
+	DistObjectCacheEntry *procedure = NULL;
+	FuncExpr *funcExpr = callStmt->funcexpr;
+	Oid functionId = funcExpr->funcid;
+
+	procedure = LookupDistObjectCacheEntry(ProcedureRelationId, functionId, 0);
+	if (procedure == NULL)
+	{
+		return false;
+	}
+
+	return CallFuncExprRemotely(callStmt, procedure, funcExpr, queryString, dest);
+}
+
+
+/*
+ * CallFuncExprRemotely calls a procedure of function on the worker if possible.
+ */
+static bool
+CallFuncExprRemotely(CallStmt *callStmt, DistObjectCacheEntry *procedure,
+					 FuncExpr *funcExpr, const char *queryString, DestReceiver *dest)
+{
+	Oid colocatedRelationId = InvalidOid;
+	Const *partitionValue = NULL;
+	ShardInterval *shardInterval = NULL;
+	List *placementList = NIL;
+	ListCell *argCell = NULL;
+	WorkerNode *preferredWorkerNode = NULL;
+	DistTableCacheEntry *distTable = NULL;
+	ShardPlacement *placement = NULL;
+	WorkerNode *workerNode = NULL;
+
+	if (IsMultiStatementTransaction())
+	{
+		ereport(DEBUG2, (errmsg("cannot push down CALL in multi-statement transaction")));
+		return false;
+	}
+
+	colocatedRelationId = ColocatedTableId(procedure->colocationId);
+	if (colocatedRelationId == InvalidOid)
+	{
+		ereport(DEBUG2, (errmsg("stored procedure does not have co-located tables")));
+		return false;
+	}
+
+	if (procedure->distributionArgIndex < 0 ||
+		procedure->distributionArgIndex >= list_length(funcExpr->args))
+	{
+		ereport(DEBUG2, (errmsg("cannot push down invalid distribution_argument_index")));
+		return false;
+	}
+
+	foreach(argCell, funcExpr->args)
+	{
+		Node *argNode = (Node *) lfirst(argCell);
+		if (!IsA(argNode, Const))
+		{
+			ereport(DEBUG2, (errmsg("cannot push down non-constant argument value")));
+			return false;
+		}
+	}
+
+	partitionValue = (Const *) list_nth(funcExpr->args, procedure->distributionArgIndex);
+	distTable = DistributedTableCacheEntry(colocatedRelationId);
+	shardInterval = FindShardInterval(partitionValue->constvalue, distTable);
+
+	if (shardInterval == NULL)
+	{
+		ereport(DEBUG2, (errmsg("cannot push down call, failed to find shard interval")));
+		return false;
+	}
+
+	placementList = FinalizedShardPlacementList(shardInterval->shardId);
+	if (list_length(placementList) != 1)
+	{
+		/* punt on reference tables for now */
+		ereport(DEBUG2, (errmsg(
+							 "cannot push down CALL for reference tables or replicated distributed tables")));
+		return false;
+	}
+
+	placement = (ShardPlacement *) linitial(placementList);
+	workerNode = FindWorkerNode(placement->nodeName, placement->nodePort);
+
+	if (workerNode->hasMetadata)
+	{
+		/* we can execute this procedure on the worker! */
+		preferredWorkerNode = workerNode;
+	}
+
+	if (preferredWorkerNode == NULL)
+	{
+		ereport(DEBUG2, (errmsg("there is no worker node with metadata")));
+		return false;
+	}
+
+	{
+		Tuplestorestate *tupleStore = tuplestore_begin_heap(true, false, work_mem);
+		TupleDesc tupleDesc = CallStmtResultDesc(callStmt);
+		TupleTableSlot *slot = MakeSingleTupleTableSlotCompat(tupleDesc,
+															  &TTSOpsMinimalTuple);
+
+		Task *task = CitusMakeNode(Task);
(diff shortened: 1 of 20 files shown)


src/backend/distributed/citus--8.3-1--8.4-1.sql
```diff
@@ -209,13 +214,20 @@ BEGIN
...
+		classid, objid, objsubid,
+		distribution_arg_index,
+		colocation_id)
     SELECT
         address.classid,
         address.objid,
-        address.objsubid
+        address.objsubid,
+		address.distribution_arg_index,
+		address.colocation_id
```
marcocitus: We should probably move this statement into its own function in case we need to change it again later.

src/backend/distributed/commands/call.c
```diff
@@ -0,0 +1,127 @@
...
+	WorkerNode *preferredWorkerNode = NULL;
+	int connectionFlags = 0;
+	MultiConnection *connection = NULL;
+
+	if (IsMultiStatementTransaction())
+	{
+		ereport(DEBUG2, (errmsg("cannot push down CALL in multi-statement transaction")));
+		return false;
+	}
+
```
marcocitus: some other things to maybe watch out for (not sure if these are actual issues) 
- Expressions in the parameters that cannot be evaluated on the worker 
- Implicit casts in the distribution argument

src/backend/distributed/planner/shard_pruning.c
```diff
@@ -1563,3 +1563,24 @@ ExhaustivePruneOne(ShardInterval *curInterval,
...
 	return false;
 }
+
+
+/*
+ * ShardsForPartitionColumnValue gets the shards for a single partition column value.
+ */
+List *
+ShardsForPartitionColumnValue(Oid relationId, Node *partitionColumnValue)
+{
```
onderkalaci: Minor: Might be cheaper to call `FindShardInterval()` and avoids creating the `opExpr` etc. You can see `get_shard_id_for_distribution_column()` UDF for how to use it.
```C	
    ... calculate distributionValueDatum ...
	shardInterval = FindShardInterval(distributionValueDatum, cacheEntry);
```

src/backend/distributed/citus--8.3-1--8.4-1.sql
```diff
@@ -209,13 +214,20 @@ BEGIN
         RETURNING
             type,
             object_names,
-            object_args
+            object_args,
+            distribution_arg_index,
+            colocation_id
     )
-    INSERT INTO citus.pg_dist_object (classid, objid, objsubid)
+    INSERT INTO citus.pg_dist_object (
```
marcocitus: We'll move this function in its own file per #2951, so ignore the comment about moving this into its own function.

src/backend/distributed/sql/udfs/citus_finish_pg_upgrade/9.0-1.sql
```diff
@@ -73,22 +73,16 @@ BEGIN
...
-    INSERT INTO citus.pg_dist_object (classid, objid, objsubid)
-    SELECT
-        address.classid,
-        address.objid,
-        address.objsubid
-    FROM
-        old_records naming,
-        pg_get_object_address(naming.type, naming.object_names, naming.object_args) address;
+    UPDATE citus.pg_dist_object
+    SET classid = address.classid, objid = address.objid, objsubid = address.objsubid
```
marcocitus: We didn't do an update here because it can cause a primary key violation when a new OID matches an old OID. We also couldn't use deferred constraints in combination with systable functions.

src/backend/distributed/commands/call.c
```diff
@@ -0,0 +1,128 @@
...
+	foreach(placementCell, placementList)
+	{
+		ShardPlacement *placement = (ShardPlacement *) lfirst(placementCell);
+		WorkerNode *workerNode = FindWorkerNode(placement->nodeName, placement->nodePort);
+
+		if (workerNode->hasMetadata)
+		{
+			/* we can execute this procedure on the worker! */
+			preferredWorkerNode = workerNode;
```
serprex: I'd think we'd offer round robin as an option here

marcocitus: Would make sense when co-locating with a reference table.

serprex:
So as to where this task is right now: I'm testing things

Right now there's a segfault related to INOUT parameters, found while testing the example `triple` in docs https://www.postgresql.org/docs/11/plpgsql-control-structures.html#PLPGSQL-STATEMENTS-CALLING-PROCEDURE

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -896,6 +922,104 @@ LookupDistTableCacheEntry(Oid relationId)
...
+	 * here.
+	 */
+	if (!CitusHasBeenLoaded())
+	{
+		return NULL;
+	}
+
+	InitializeCaches();
+
+	cacheEntry = hash_search(DistTableCacheHash, &hashKey, HASH_ENTER, &foundInCache);
```
pykello: `DistObjectCacheHash`?

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -3565,7 +3757,8 @@ TupleToShardInterval(HeapTuple heapTuple, TupleDesc tupleDescriptor, Oid interva
 	 */
 	heap_deform_tuple(heapTuple, tupleDescriptor, datumArray, isNullArray);
 
-	relationId = DatumGetObjectId(datumArray[Anum_pg_dist_shard_logicalrelid - 1]);
+	relationId = DatumGetObjectId(datumArray[Anum_pg_dist_shard_logicalrelid -
+											 1]);
```
pykello: Why are all of these reindent changes? Do you have different config for reindent by any chance?

serprex: I think my version is different

src/backend/distributed/utils/metadata_cache.c
serprex: That was indeed the issue

src/include/distributed/metadata_cache.h
```diff
@@ -89,6 +89,24 @@ typedef struct
...
+typedef struct DistObjectCacheEntry
+{
+	/* lookup key - must be first. */
+	DistObjectCacheEntryKey key;
+
+	bool isValid;
+
+	int distributionArgIndex;
+	int colocationId;
+} DistObjectCacheEntry;
```
pykello: Why are the rest of pg_dist_object's columns not cached?

serprex: The other 3 should only be used for upgrades

src/backend/distributed/commands/call.c
```diff
@@ -0,0 +1,143 @@
...
+
+	colocatedRelationId = ColocatedTableId(procedure->colocationId);
+	if (colocatedRelationId == InvalidOid)
+	{
+		ereport(DEBUG2, (errmsg("stored procedure does not have co-located tables")));
+		return false;
+	}
+
+	partitionValue = list_nth(funcExpr->args, procedure->distributionArgIndex);
+	partitionValueConst = (Const *) strip_implicit_coercions(partitionValue);
```
pykello: 1. This probably should be moved to after the `!IsA(...)` check
2. If node type is `Const`, `strip_implicit_coercions()` just returns it as it is.

src/backend/distributed/commands/call.c
```diff
@@ -0,0 +1,143 @@
...
+/*
+ * CallFuncExprRemotely calls a procedure of function on the worker if possible.
+ */
+static bool
+CallFuncExprRemotely(DistObjectCacheEntry *procedure, FuncExpr *funcExpr,
+					 const char *queryString, DestReceiver *dest)
+{
+	Oid colocatedRelationId = InvalidOid;
+	Node *partitionValue = NULL;
+	ShardInterval *shard = NULL;
```
pykello: shardInterval?

src/backend/distributed/commands/call.c
```diff
@@ -0,0 +1,143 @@
...
+
+	placementList = FinalizedShardPlacementList(shard->shardId);
+	if (list_length(placementList) != 1)
+	{
+		/* punt on reference tables for now */
+		ereport(DEBUG2, (errmsg("cannot push down CALL without a distribution column")));
+		return false;
+	}
+
+	foreach(placementCell, placementList)
```
pykello: We have already checked that we have only one placement?

src/backend/distributed/commands/call.c
```diff
@@ -0,0 +1,143 @@
...
+	{
+		ereport(DEBUG2, (errmsg("cannot push down call, failed to find shard interval")));
+		return false;
+	}
+
+	placementList = FinalizedShardPlacementList(shard->shardId);
+	if (list_length(placementList) != 1)
+	{
+		/* punt on reference tables for now */
+		ereport(DEBUG2, (errmsg("cannot push down CALL without a distribution column")));
```
pykello: "cannot push down CALL for reference tables or replicated distributed tables"?

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -3222,6 +3384,24 @@ InvalidateEntireDistCache(void)
 }
 
 
+/*
+ * InvalidateDistObjectCache marks all DistObjectCacheHash entries invalid.
+ */
+static void
+InvalidateDistObjectCache(void)
```
marcocitus: Should we put a trigger on pg_dist_object that calls a UDF that calls this function? That's what we do for other metadata tables.

src/backend/distributed/commands/call.c
serprex: The logic now only allows for cases when all arguments are `Const` nodes

src/backend/distributed/utils/metadata_cache.c
serprex: From what I can tell we track relation changes in `InvalidateDistRelationCacheCallback` which is registered in `InitializeDistCache`

marcocitus: That's listening for catalog changes, which we use to invalidate the distributed table cache when someone adds a column to the distributed table. Individual insert/update/delete commands are handled via triggers (e.g. `master_dist_node_cache_invalidate`)

src/backend/distributed/commands/call.c
```diff
@@ -0,0 +1,204 @@
...
+
+static void PipePgResultToDestReceiver(TupleDesc resultDesc, PGresult *result,
+									   DestReceiver *dest);
+static bool CallFuncExprRemotely(CallStmt *callStmt,
+								 DistObjectCacheEntry *procedure,
+								 FuncExpr *funcExpr, const char *queryString,
+								 DestReceiver *dest);
+
+/*
+ * CallDistributedProcedure calls a stored procedure on the worker if possible.
```
pykello: Update the function name in comment.

src/backend/distributed/commands/call.c
```diff
@@ -0,0 +1,204 @@
...
+	if (list_length(placementList) != 1)
+	{
+		/* punt on reference tables for now */
+		ereport(DEBUG2, (errmsg(
+							 "cannot push down CALL for reference tables or replicated distributed tables")));
+		return false;
+	}
+
+	placement = (ShardPlacement *) linitial(placementList);
+	workerNode = FindWorkerNode(placement->nodeName, placement->nodePort);
```
pykello: This can return `NULL`.

src/test/regress/expected/multi_mx_call.out
```diff
@@ -0,0 +1,53 @@
+-- Test passing off CALL to mx workers
```
pykello: These tests failed probably with a worker crash, when postgres/citus is compiled with `--enable-cassert`. Can you double check that?

src/backend/distributed/commands/call.c
```diff
@@ -0,0 +1,204 @@
...
+	return CallFuncExprRemotely(callStmt, procedure, funcExpr, queryString, dest);
+}
+
+
+/*
+ * PipePgResultToDestReceiver outputs a PGresult into a DestReceiver
+ * We assume DestReceiver expects a TTSOpsVirtual tuple slot
+ */
+static void
+PipePgResultToDestReceiver(TupleDesc resultDesc, PGresult *result, DestReceiver *dest)
```
pykello: Alternative approach is https://github.com/citusdata/citus/pull/2998/, just in case.

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -2643,6 +2771,31 @@ master_dist_local_group_cache_invalidate(PG_FUNCTION_ARGS)
...
+	{
+		ereport(ERROR, (errcode(ERRCODE_E_R_I_E_TRIGGER_PROTOCOL_VIOLATED),
+						errmsg("must be called as trigger")));
+	}
+
+	CheckCitusVersion(ERROR);
+
+	CitusInvalidateRelcacheByRelid(DistObjectRelationId());
+
+	PG_RETURN_DATUM(PointerGetDatum(NULL));
```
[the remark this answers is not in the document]
pykello: If by any chance `PG_RETURN_NULL()` doesn't work here, can you read the docs to see what `PG_RETURN_DATUM(PointerGetDatum(NULL));` indicates?

src/test/regress/sql/multi_mx_call.sql
```diff
@@ -0,0 +1,35 @@
...
+create table mx_call_table (val int);
+insert into mx_call_table values (2);
+$$);
+
+CREATE TABLE mx_call_dist_table(id int);
+select create_distributed_table('mx_call_dist_table', 'id');
+
+CREATE PROCEDURE mx_call_proc(x int, INOUT y int) LANGUAGE plpgsql AS $$
+BEGIN
+    y := x + (select val from mx_call_table);
```
marcocitus: You could use the `groupid` value in `pg_dist_local_group` to distinguish workers.

src/test/regress/expected/multi_mx_call.out
marcocitus: for me too, https://github.com/citusdata/citus/pull/2998/ fixes it

src/backend/distributed/utils/metadata_cache.c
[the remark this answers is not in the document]
serprex: > A trigger function must return either NULL or a record/row value having exactly the structure of the table the trigger was fired for.

From https://www.postgresql.org/docs/9.2/plpgsql-trigger.html

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -2670,18 +2823,12 @@ InitializeCaches(void)
 			/* set first, to avoid recursion dangers */
 			performedInitialization = true;
 
-			/* make sure we've initialized CacheMemoryContext */
```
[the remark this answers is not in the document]
serprex: `CacheMemoryContext` comes from `catcache.h`, so was taken out when I removed use of `SearchCatCache3`. Reverted removal

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -805,6 +812,26 @@ DistributedTableCacheEntry(Oid distributedRelationId)
 }
 
 
+/*
+ * DistributedProcedureRecord loads a record from citus.pg_dist_object.
+ */
+DistObjectCacheEntry *
+DistributedObjectCacheEntry(Oid classid, Oid objid, int32 objsubid)
```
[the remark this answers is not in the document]
serprex: Yes

src/include/distributed/remote_commands.h
```diff
@@ -43,7 +43,8 @@ extern void LogRemoteCommand(MultiConnection *connection, const char *command);
 extern void ExecuteCriticalRemoteCommandList(MultiConnection *connection,
 											 List *commandList);
 extern void ExecuteCriticalRemoteCommand(MultiConnection *connection,
-										 const char *command);
+										 const char *command,
+										 struct pg_result **result);
```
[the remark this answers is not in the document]
serprex: So code originally came with a `struct pg_result;` for use to not have `PGresult` in header, but Citus still compiles if we reference `PGresult` so `PGresult` we'll use

src/backend/distributed/commands/call.c
```diff
@@ -0,0 +1,170 @@
...
+								MaxAdaptiveExecutorPoolSize);
+
+		while (tuplestore_gettupleslot(tupleStore, true, false, slot))
+		{
+			if (!dest->receiveSlot(slot, dest))
+			{
+				break;
+			}
+		}
+	}
```
pykello: Maybe we should also call `tuplestore_end(tupleStore)` here.

serprex: I've included a comment which outlines 2 potential reasons not to call `tuplestore_end`

src/backend/distributed/commands/call.c
```diff
@@ -0,0 +1,174 @@
...
+		/* punt on reference tables for now */
+		ereport(DEBUG2, (errmsg(
+							 "cannot push down CALL for reference tables or replicated distributed tables")));
+		return false;
+	}
+
+	placement = (ShardPlacement *) linitial(placementList);
+	workerNode = FindWorkerNode(placement->nodeName, placement->nodePort);
+
+	if (workerNode == NULL || !workerNode->hasMetadata)
```
pykello: Also check for `metadatasynced`?

src/test/regress/sql/multi_mx_call.sql
```diff
@@ -0,0 +1,30 @@
+-- Test passing off CALL to mx workers
```
pykello: Maybe extend the tests to:
1. Check for cases that we shouldn't route? Set `client_min_messages`, and maybe create a case for each of the `return false`s in `CallFuncExprRemotely`.
2. Can procedures return set of records, or other kinds of sets? Maybe check for that.
3. Add a real MX table test.
4. Maybe add some cache related tests. For example, verify that UPDATE/DELETEs invalidate cache correctly.
...

pykello:
Other than the comment on testing, rest of the PR looks fine to me.

marcocitus:
Some suggested fixes:
https://github.com/citusdata/citus/commit/eff1753162766d21ef2c570f53dcb829ee04d896

This mainly resolves two issues for commands like:
https://github.com/TPC-Council/HammerDB/blob/master/src/postgresql/pgoltp.tcl#L2399

In that case, the type of the distribution column argument does not match the type of the distribution column and there are expressions in the argument list.

src/backend/distributed/commands/call.c
```diff
@@ -0,0 +1,193 @@
...
+			if (!dest->receiveSlot(slot, dest))
+			{
+				break;
+			}
+		}
+
+		/* Don't call tuplestore_end(tupleStore). It'll be freed soon enough in a top level CALL,
+		 * & dest->receiveSlot could conceivably rely on slots being long lived.
+		 */
+	}
```
marcocitus: Maybe wrap this block in a function, e.g. `ExecuteTaskListIntoDestReceiver`

src/backend/distributed/commands/call.c
```diff
@@ -0,0 +1,193 @@
...
+		task->taskId = 0;
+		task->taskType = DDL_TASK;
+		task->queryString = pstrdup(queryString);
+		task->replicationModel = REPLICATION_MODEL_INVALID;
+		task->dependedTaskList = NIL;
+		task->anchorShardId = placement->shardId;
+		task->relationShardList = NIL;
+		task->taskPlacementList = placementList;
+
+		ExecuteTaskListExtended(ROW_MODIFY_COMMUTATIVE, list_make1(task),
```
marcocitus: I think we can effectively treat this as a read-only query because all the necessary locking (including shard metadata locks) will be done on the worker.

src/backend/distributed/connection/remote_commands.c
```diff
@@ -459,6 +433,33 @@ ExecuteOptionalRemoteCommand(MultiConnection *connection, const char *command,
 }
 
 
+/*
+ * ExecuteCriticalRemoteCommand executes a remote command that is critical
+ * to the transaction. If the command fails then the transaction aborts.
+ */
+void
+ExecuteCriticalRemoteCommand(MultiConnection *connection, const char *command,
+							 PGresult **result)
```
marcocitus: I think modifying this function is no longer necessary for this PR. Maybe leave it to another PR (or not since it'll be dead code).

marcocitus:
Overall, this looks good me. I'll let @pykello sign off on it.

One high level observation is that we don't process notices nicely. That isn't generally a problem because we don't really expect many NOTICE or even WARNING messages when executing a query, but in procedures we can easily have a RAISE NOTICE and these emitted with DEBUG1 instead of NOTICE in `DefaultCitusNoticeProcessor`.

src/test/regress/sql/multi_mx_call.sql
```diff
@@ -0,0 +1,160 @@
...
+SELECT create_distributed_function('mx_call_add(int,int)');
+
+-- non-const distribution parameters cannot be pushed down
+call mx_call_proc(2, mx_call_add(3, 4));
+
+-- non-const parameter can be pushed down
+call mx_call_proc(mx_call_add(3, 4), 2);
+
+-- volatile parameter cannot be pushed down
+call mx_call_proc(random()::int, 2);
```
pykello: Output of this is not consistent. Maybe do `random()::int % 1` instead.

src/backend/distributed/commands/call.c
serprex: Updated to use `ROW_MODIFY_NONE` as utility statements are apt to do

src/backend/distributed/connection/remote_commands.c
serprex: Reverted. Personally think the implementation being shared between Critical & Optional is correct, but the third parameter would be omitted in a PR

src/backend/distributed/commands/call.c
serprex: I think it'd be better to wait until there's a clearer idea of how much of this logic is parametric

src/backend/distributed/utils/maintenanced.c
```diff
@@ -374,6 +374,7 @@ CitusMaintenanceDaemonMain(Datum main_arg)
 
 			InvalidateMetadataSystemCache();
 			StartTransactionCommand();
+			PushActiveSnapshot(GetTransactionSnapshot());
```
marcocitus: We already do this on the line below.

src/backend/distributed/commands/call.c
```diff
@@ -0,0 +1,205 @@
...
+	{
+		ereport(DEBUG1, (errmsg("there is no worker node with metadata")));
+		return false;
+	}
+
+	ereport(DEBUG1, (errmsg("pushing down the procedure")));
+
+	/* build remote command with fully qualified names */
+	callCommand = makeStringInfo();
+	appendStringInfo(callCommand, "CALL %s", pg_get_rule_expr((Node *) funcExpr));
```
marcocitus: Pushed a commit to deparse the query string instead of using the original one.

src/backend/distributed/commands/call.c
```diff
@@ -0,0 +1,205 @@
...
+		task->taskId = 0;
+		task->taskType = DDL_TASK;
+		task->queryString = callCommand->data;
+		task->replicationModel = REPLICATION_MODEL_INVALID;
+		task->dependedTaskList = NIL;
+		task->anchorShardId = placement->shardId;
+		task->relationShardList = NIL;
+		task->taskPlacementList = placementList;
+
+		ExecuteTaskListExtended(ROW_MODIFY_NONE, list_make1(task),
```
onderkalaci: super minor: instead of passing `true`, it'd be nice to pass `hasReturning` variable, because I've trying to understand how does this return tuples with `ROW_MODIFY_NONE ` and `DDL_TASK`

src/test/regress/sql/multi_mx_call.sql
```diff
@@ -0,0 +1,194 @@
...
+
+-- We still don't route them to the workers, because they aren't
+-- colocated with any distributed tables.
+SET client_min_messages TO DEBUG1;
+call multi_mx_call.mx_call_proc(2, 0);
+call multi_mx_call.mx_call_proc_custom_types('S', 'A');
+
+-- Mark them as colocated with a table. Now we should route them to workers.
+call multi_mx_call.colocate_proc_with_table('mx_call_proc', 'mx_call_dist_table_1'::regclass, 1);
+call multi_mx_call.colocate_proc_with_table('mx_call_proc_custom_types', 'mx_call_dist_table_enum'::regclass, 1);
```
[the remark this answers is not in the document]
serprex: Current UDF also is a clearer test of cache invalidation

src/test/regress/sql/multi_mx_call.sql
```diff
@@ -0,0 +1,194 @@
+-- Test passing off CALL to mx workers
+
+-- Create worker-local tables to test procedure calls were routed
+
+set citus.shard_replication_factor to 2;
+set citus.replication_model to 'statement';
+
+-- This table requires specific settings, create before getting into things
+create table mx_call_dist_table_replica(id int, val int);
+select create_distributed_table('mx_call_dist_table_replica', 'id');
```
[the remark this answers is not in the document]
serprex: I'd rather keep it unless there's an alternative way to hit the replica error message

PR #2967, opened 2019-09-13
MX isolation test insert

halilozanakgul:
Created MX isolation tests for insert operation.

### DIFF
--- a/src/test/regress/expected/isolation_insert_vs_all_on_mx.out
+++ b/src/test/regress/expected/isolation_insert_vs_all_on_mx.out
@@ -1,0 +1,783 @@
+Parsed test spec with 3 sessions
+
+starting permutation: s1-start-session-level-connection s1-begin-on-worker s1-insert s2-start-session-level-connection s2-begin-on-worker s2-insert s1-commit-worker s2-commit-worker s3-select-count s1-stop-connection s2-stop-connection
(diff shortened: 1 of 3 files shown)


halilozanakgul:
> One question I've is the reference tables. With this test(s) we're triggering distributed tables, but we should probably have some coverage for reference tables as well. We can definitely handle that later, separately.
> This looks good, and #2972 makes it pretty simple to follow the progress.

There is already a test file for reference tables (isolation_reference_on_mx.spec)

onderkalaci:
> There is already a test file for reference tables (isolation_reference_on_mx.spec)

Does it cover all the cases as you listed #2972?

halilozanakgul:
> There is already a test file for reference tables (isolation_reference_on_mx.spec)
> 
> Does it cover all the cases as you listed #2972?

  | update | delete | insert | copy | select for update | select from | alter table
-- | -- | -- | -- | -- | -- | -- | --
update | + |   |   |   |   |   |  
delete | + |   |   |   |   |   |  
insert | + |   | + |   |   |   | +
copy | + |   | + | + |   |   |  
select for update | + |   |   |   |   |   |  
select from |   |   |   |   |   |   | +
alter table |   |   |   |   |   |   |  

It only covers these

PR #2969, opened 2019-09-13
Ignore .vscode

SaitTalhaNisanci:


serprex:
Is it best to ignore `.vscode`? My experience in the past has been that it's useful for storing formatting configuration, but that was on a javascript codebase

SaitTalhaNisanci:
@serprex if we want to share formatting configuration, extensions etc we can probably exclude them from gitignore. We can use something like:
```
.vscode/*
!.vscode/settings.json # this contains formatting configuration etc
!.vscode/tasks.json
!.vscode/launch.json
!.vscode/extensions.json
```

what do you think?

serprex:
I don't use vscode for Citus development, so I don't know precisely which files shouldn't be ignored, but that's a good solution if auto formatting is in there

SaitTalhaNisanci:
Well that is the file that contains auto formatting but for now I think it is better to ignore vscode because that file can have other user specific things too. If that is a concern we can just use the solution I provided later(not ignoring the settings file etc).

PR #2973, opened 2019-09-16
Add postgres upgrade job

### LINKED ISSUE #2974: Add postgres upgrade job to CI
Now that we have a script to check postgres upgrades, we should add that to our CI.

SaitTalhaNisanci:
This PR adds postgres upgrade job. The upgrade job takes `--old-pg-version` and `--new-pg-version` parameters and it performs the upgrade test: https://github.com/citusdata/citus/blob/master/src/test/regress/Makefile#L144

You can see the image [here](https://github.com/citusdata/the-process/tree/upgradeTester/circleci/images/pgupgradetester)

Instead of putting the parameters to the [script file](https://github.com/citusdata/the-process/blob/upgradeTester/circleci/images/pgupgradetester/files/sbin/install-and-test-ext) that is run on docker I wanted to put them outside so that if we want to test different versions of postgres we dont need to:
- Change the parameters in the script
- Build the new image
- Push the new image

Currently we have postgres upgrade tests for:
- `pg10 -> pg11`
- `pg11 -> pg12`

For `build` job I am using my own image that has citus artifacts for postgres 12. I will push the same image to `citus` docker hub when the PR is approved(meaning other parts are agreed on.)

The upgrade job takes [39](https://circleci.com/gh/citusdata/citus/38512) seconds. 

Fixes #2974.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@ orbs:
 jobs:
   build:
     docker:
-      - image: 'citusdata/extbuilder:werror'
+      - image: 'citus/extbuilder:latest'
     steps:
       - checkout
       - run: 
@@ -204,9 +204,30 @@ jobs:
     steps:
       - attach_workspace: 
           at: .
-      - run:
+      - run: 
           name: 'Install and Test (check-failure)'
           command: 'install-and-test-ext check-failure'
+  test-10-11_check-pg-upgrade:
+    docker:
+      - image: 'citus/pgupgradetester:latest'
+    working_directory: /home/circleci/project     
+    steps:
+      - attach_workspace:
+          at: .
+      - run: 
+          name: 'Install and test postgres upgrade'
+          command: 'install-and-test-ext --target check-upgrade --old-pg-version 10 --new-pg-version 11'
+  test-11-12_check-pg-upgrade:
+    docker:
+      - image: 'citus/pgupgradetester:latest'
+    working_directory: /home/circleci/project     
+    steps:
+      - attach_workspace:
+          at: .
+      - run: 
+          name: 'Install and test postgres upgrade'
+          command: 'install-and-test-ext --target check-upgrade --old-pg-version 11 --new-pg-version 12'        
+
 workflows:
   version: 2
   build_and_test:
@@ -237,9 +258,14 @@ workflows:
       - test-11_check-failure:
           requires: [build]
 
-      - test-11_check-non-adaptive-multi:
+      - test-11_check-non-adaptive-multi: 
           requires: [build]
       - test-11_check-non-adaptive-failure:
           requires: [build]
       - test-11_check-non-adaptive-isolation:
           requires: [build]
+
+      - test-10-11_check-pg-upgrade: 
+          requires: [build]
+      - test-11-12_check-pg-upgrade: 
+          requires: [build]    


metdos:
We should also add an upgrade test from PG11 to PG12.

.circleci/config.yml
```diff
@@ -147,6 +147,13 @@ jobs:
     steps:
       - {attach_workspace: {at: .}}
       - {run: {name: 'Install and Test (check-failure)', command: 'install-and-test-ext check-failure'}}
+  test_check-pg-upgrade:
+    docker:
+      - {image: 'citus/pgupgradetester:latest'}
```
thanodnl: the other images have their pg versions in the image name. I think we should either put it in the image name or use a tag other then `latest` with the pg versions we are performing the upgrade for.

SaitTalhaNisanci: If we put the pg versions in the image name, it would mean that we will have two different images if we want to have `(pg10->pg11)` and `(pg11->pg12)`. I thought that instead of doing that if we give the postgres versions as parameters we can just perform the same tests `(pg10->pg11)` and `(pg11->pg12)` in the same image.  So when we add `pg12` we can have something like:  
```yaml
- run: 
     name: 'Install and test postgres upgrade'
     command: 
          'install-and-test-ext --target check-upgrade --old-pg-version 10 --new-pg-version 11'
          'install-and-test-ext --target check-upgrade --old-pg-version 11 --new-pg-version 12'
```

if we create new images for each postgres combination then we will have more jobs. With citus upgrade tests we will have an additional job too. I think it makes sense to have separate images for different postgres versions when the job is running slow, because with 2 jobs it can be parallelized. 

What is the downside of having parameters?

marcocitus: We could have one image that contains pg10, pg11, pg12, ... and perhaps even use it for all other tests as well.

.circleci/config.yml
```diff
@@ -5,7 +5,7 @@ orbs:
 jobs:
   build:
     docker:
-      - image: 'citusdata/extbuilder:werror'
+      - image: 'heisenberg302/build:latest'
```
thanodnl: please use the images in either `citusdata` or `citus`

.circleci/config.yml
```diff
@@ -204,9 +204,21 @@ jobs:
     steps:
       - attach_workspace: 
           at: .
-      - run:
+      - run: 
           name: 'Install and Test (check-failure)'
           command: 'install-and-test-ext check-failure'
+  test_check-pg-upgrade:
+    docker:
+      - image: 'heisenberg302/pgupgradetester:latest'
```
thanodnl: same

.circleci/config.yml
```diff
@@ -204,9 +204,21 @@ jobs:
...
+    working_directory: /home/circleci/project     
+    steps:
+      - attach_workspace:
+          at: .
+      - run: 
+          name: 'Install and test postgres upgrade'
+          command: |
+            install-and-test-ext --target check-upgrade --old-pg-version 10 --new-pg-version 11
+            install-and-test-ext --target check-upgrade --old-pg-version 11 --new-pg-version 12
```
thanodnl: can we please create to jobs in the matrix? This helps in attributing faults easier, and also allows for more parallelism.

.circleci/config.yml
SaitTalhaNisanci: I will update them once we agree on the other details because I want to put the final image to `citus`.

PR #2979, opened 2019-09-16
 Add columns to pg_dist_object for distributed functions

onderkalaci:
Both my work and #2965 relies on the changes here. So, let's merge this to master and do not block/rely on each other.

@serprex This change is already based on #2965, so can you please review this? 

Note: The internal data structures always rely on the index (even if we might consider using parameter name), because that's cheaper to keep and process.

### DIFF
--- a/src/backend/distributed/sql/citus--8.3-1--9.0-1.sql
+++ b/src/backend/distributed/sql/citus--8.3-1--9.0-1.sql
@@ -38,6 +38,11 @@ CREATE TABLE citus.pg_dist_object (
     object_names text[] DEFAULT NULL,
     object_args text[] DEFAULT NULL,
 
+    -- fields that are only valid for distributed
+    -- functions/procedures
+    distribution_argument_index int,
+    colocationid int,
+
     CONSTRAINT pg_dist_object_pkey PRIMARY KEY (classid, objid, objsubid)
 );
 
--- a/src/include/distributed/pg_dist_object.h
+++ b/src/include/distributed/pg_dist_object.h
@@ -1,0 +1,29 @@
+/*-------------------------------------------------------------------------
+ *
+ * pg_dist_object.h
+ *	  definition of the relation that holds the object information on the
+ *	  cluster (pg_dist_object).
+ *
+ * Copyright (c) 2019, Citus Data, Inc.
+ *
+ *-------------------------------------------------------------------------
+ */
+
+#ifndef PG_DIST_OBJECT_H
+#define PG_DIST_OBJECT_H
+
+/* ----------------
+ *      compiler constants for pg_dist_object
+ * ----------------
+ */
+#define Natts_pg_dist_object 8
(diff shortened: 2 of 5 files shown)


serprex:
@onderkalaci I think having 2 functions makes more sense

onderkalaci:
> @onderkalaci I think having 2 functions makes more sense

After reading my comments again, that also makes more sense to me actually.

src/backend/distributed/sql/citus--8.3-1--9.0-1.sql
```diff
@@ -33,6 +33,11 @@ CREATE TABLE citus.pg_dist_object (
     objid oid NOT NULL,
     objsubid integer NOT NULL,
 
+    -- fields that are only valid for distributed
+    -- functions/procedures
+    distribution_argument_index int,
+    colocationid int,
```
marcocitus: It would make a bit more sense for these to be at the end, since we might add more of these class-specific fields later.

src/backend/distributed/sql/udfs/create_distributed_function/latest.sql
```diff
@@ -1,7 +1,12 @@
-CREATE OR REPLACE FUNCTION create_distributed_function(function_name regprocedure)
+CREATE OR REPLACE FUNCTION create_distributed_function(function_name regprocedure,
+						       colocate_with text DEFAULT 'default',
+						       distribution_argument_name text DEFAULT '',
```
marcocitus: Seems fine. I would have expected that the arguments have the same order as create_distributed_table in the happy case, but could be convinced otherwise.

src/backend/distributed/sql/udfs/create_distributed_function/latest.sql
```diff
@@ -1,7 +1,12 @@
-CREATE OR REPLACE FUNCTION create_distributed_function(function_name regprocedure)
+CREATE OR REPLACE FUNCTION create_distributed_function(function_name regprocedure,
```
marcocitus: Should this change be in a separate commit or PR? This doesn't seem relevant to other PRs without an actual implementation.

We should also make sure to use qualified names in these scripts or move them into the upgrade script such that the search path is clear.

onderkalaci: Ok, makes sense. We can discuss the API on another PR.

src/include/distributed/pg_dist_object.h
```diff
@@ -0,0 +1,29 @@
+/*-------------------------------------------------------------------------
```
pykello: This file should be merged into src/include/distributed/metadata/pg_dist_object.h. Two issues:
1. Both start with `#ifndef PG_DIST_OBJECT_H #define PG_DIST_OBJECT_H`,
2. Both define `Anum_pg_dist_object...`.

Issue #2954, opened 2019-09-11
Add feature flag to disable new behaviour for type distribution


thanodnl:
Type distribution is a complex feature, if we run into issues in production it is helpful if there is a feature flag that we can use to disable its behaviour.

A guc with `GUC_NO_SHOW_ALL` can be used to create a hidden feature flag that can be checked in `commands/type.c` before we decide to propagate type statement.

thanodnl:
The following GUC's have been added
 - `citus.enable_create_type_propagation` default value `true`
 - `citus.enable_object_propagation` default value `true`

PR #2982, opened 2019-09-17
Add feature flag to turn off create type propagation

thanodnl:
DESCRIPTION: Add feature flag to turn off create type propagation

Solves #2954 

When `citus.enable_create_type_propagation` is set to `false` citus will not propagate `CREATE TYPE` statements to the workers. Types are still distributed when tables that depend on these types are distributed.


thanodnl:
Makes sense to have a more generic switch as well, which is not trivial given the schema's still need to be created. Will work an a patch like that.

Will merge this after rebase and passed tests.

PR #2985, opened 2019-09-17
Create previously distributed functions in new workers

### LINKED ISSUE #2984: Adding a node fails if a distributed function exists
If assertions are enabled, the following steps lead to craashes.

```SQL
CREATE EXTENSION citus;

 CREATE OR REPLACE FUNCTION add_without_param_names(integer, integer) RETURNS integer
    AS 'select $1 + $2;'
    LANGUAGE SQL
    IMMUTABLE
    RETURNS NULL ON NULL INPUT;

SELECT create_distributed_function('add_without_param_names(int, int)');

select master_add_node('localhost', 9700);
ERROR:  unsupported object function for distribution by citus
DETAIL:  citus tries to recreate an unsupported object on its workers
HINT:  please report a bug as this should not be happening
Time: 2772.641 ms (00:02.773)
```

This doesn't need to be the first node, any node fails:
```SQL
TRUNCATE citus.pg_dist_object ;
SELECT create_distributed_function('add_without_param_names(int, int)');
 select master_add_node('localhost', 9701);
ERROR:  unsupported object function for distribution by citus
DETAIL:  citus tries to recreate an unsupported object on its workers
HINT:  please report a bug as this should not be happening
Time: 2623.082 ms (00:02.623)
```

hanefi:
When we add a new node, we wish to create all the functions that are already distributed. This PR adds the missing logic in `GetDependencyCreateDDLCommands()` to include distributed function creation queries.

fixes #2984

### DIFF
--- a/src/backend/distributed/commands/function.c
+++ b/src/backend/distributed/commands/function.c
@@ -20,6 +20,7 @@
 #include "access/xact.h"
 #include "catalog/namespace.h"
 #include "catalog/pg_proc.h"
+#include "distributed/commands.h"
 #include "distributed/metadata_sync.h"
 #include "distributed/metadata/distobject.h"
 #include "distributed/multi_executor.h"
@@ -29,7 +30,7 @@
 #include "utils/fmgrprotos.h"
 
 /* forward declaration for helper functions*/
-static const char * GetFunctionDDLCommand(Oid funcOid);
+static char * GetFunctionDDLCommand(const RegProcedure funcOid);
 static void EnsureSequentialModeForFunctionDDL(void);
 
 PG_FUNCTION_INFO_V1(create_distributed_function);
@@ -75,15 +76,31 @@ create_distributed_function(PG_FUNCTION_ARGS)
 
 
 /*
+ * CreateFunctionDDLCommandsIdempotent returns a list of DDL statements (const char *) to be
+ * executed on a node to recreate the function addressed by the functionAddress.
+ */
+List *
+CreateFunctionDDLCommandsIdempotent(const ObjectAddress *functionAddress)
+{
+	char *ddlCommand = NULL;
+
+	Assert(functionAddress->classId == ProcedureRelationId);
+
+	ddlCommand = GetFunctionDDLCommand(functionAddress->objectId);
+	return list_make1(ddlCommand);
+}
+
+
+/*
  * GetFunctionDDLCommand returns the complete "CREATE OR REPLACE FUNCTION ..." statement for
  * the specified function.
  */
-static const char *
-GetFunctionDDLCommand(RegProcedure funcOid)
+static char *
+GetFunctionDDLCommand(const RegProcedure funcOid)
 {
 	OverrideSearchPath *overridePath = NULL;
 	Datum sqlTextDatum = 0;
-	const char *sql = NULL;
+	char *sql = NULL;
 
 	/*
 	 * Set search_path to NIL so that all objects outside of pg_catalog will be
--- a/src/test/regress/expected/isolation_ensure_dependency_activate_node.out
+++ b/src/test/regress/expected/isolation_ensure_dependency_activate_node.out
@@ -18,6 +18,10 @@ step s1-print-distributed-objects:
 	SELECT count(*) FROM pg_type where typname = 'tt1';
     SELECT run_command_on_workers($$SELECT count(*) FROM pg_type where typname = 'tt1';$$);
 
+    -- print if the function has been created
+    SELECT count(*) FROM pg_proc WHERE proname='add';
+    SELECT run_command_on_workers($$SELECT count(*) FROM pg_proc WHERE proname='add';$$);
+
     SELECT master_remove_node('localhost', 57638);
 
 ?column?       
@@ -25,6 +29,13 @@ step s1-print-distributed-objects:
 1              
 pg_identify_object_as_address
 
+count          
+
+0              
+run_command_on_workers
+
+(localhost,57637,t,0)
+(localhost,57638,t,0)
 count          
 
 0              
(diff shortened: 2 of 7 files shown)


marcocitus:
Cool! Probably needs a test.

onderkalaci:
>Cool! Probably needs a test.

Yes, after adding a test, this looks good to me as well. Instead of inventing a new test with `master_add_node()`, you might consider adding such a test to an already existing `master_add_node()` test.

hanefi:
> Instead of inventing a new test with `master_add_node()`, you might consider adding such a test to an already existing `master_add_node()` test.

I added some test cases to `isolation_ensure_dependency_activate_node`. @onderkalaci can you check if it looks good enough?

src/backend/distributed/commands/function.c
```diff
@@ -74,16 +75,32 @@ create_distributed_function(PG_FUNCTION_ARGS)
 }
 
 
+/*
+ * CreateFunctionDDLCommandsIdempotent returns a list of DDL statements (const char *) to be
+ * executed on a node to recreate the function addressed by the typeAddress.
```
onderkalaci: `typeAddress ` -> `functionAddress `.

src/test/regress/expected/isolation_ensure_dependency_activate_node.out
```diff
@@ -1446,6 +1725,389 @@ count
...
+	SELECT 1 FROM master_add_node('localhost', 57638);
+
+?column?       
+
+1              
+step s2-public-schema: 
+    SET search_path TO public;
+
+step s2-create-distributed-function: 
+	CREATE TYPE tt1 AS (a int, b int);
```
onderkalaci: Why do we have this type in the test? We can have it, but I what I want to learn is can we still have `<waiting ...>` even if there is no  type? If not, we should probably be able  to see it, otherwise,  we could have nodes without the function via  concurrent master_add_node and create_distributed_function

hanefi: If I remove `CREATE TYPE ..` statement from `s2-create-distributed-function` step, and later use `s2-create-type` before `s2-create-distributed-function`, we indefinitely wait. I guess the type creation waits until s1 transaction is committed to move forward.

Should I create a new permutation for creating a distributed function that does not depend on a custom type? Does that sound good?

onderkalaci: Why not we remove `s2-create-type` at all for the permutations that are related to functions? We've some other tests which follows the dependency graph, ad this point we're mostly targeting to show that function creation works fine with add_node?

src/backend/distributed/commands/function.c
hanefi: fixed it

src/test/regress/expected/isolation_ensure_dependency_activate_node.out
hanefi: We discussed it online. I changed the test steps to distribute simple functions without any custom type dependencies

src/test/regress/expected/isolation_ensure_dependency_activate_node.out
```diff
@@ -1448,6 +1727,384 @@ run_command_on_workers
...
+0              
+run_command_on_workers
+
+(localhost,57637,t,0)
+(localhost,57638,t,0)
+count          
+
+0              
+run_command_on_workers
+
```
onderkalaci: Aren't we supposed to see `1` here? I guess the function name in the query is `add_tt1 ` and the function we're actually distributing is `add`?

src/test/regress/expected/isolation_ensure_dependency_activate_node.out
```diff
@@ -1448,6 +1727,384 @@ run_command_on_workers
...
+    BEGIN;
+
+step s2-public-schema: 
+    SET search_path TO public;
+
+step s2-distribute-function: 
+    CREATE FUNCTION add (INT,INT) RETURNS INT AS $$ SELECT $1 + $2 $$ LANGUAGE SQL;
+	SELECT create_distributed_function('add(INT,INT)');
+
+ERROR:  function "add" already exists with same argument types
```
onderkalaci: Maybe use `CREATE OR REPLACE`

hanefi: Fixed all the broken function references in the file. Now all the function references are to `add(int,int)` that do not depend on any custom types

src/test/regress/expected/isolation_ensure_dependency_activate_node.out
hanefi: Yes, you are right. I fixed all `add_tt1` references and now we see 1 here.

src/test/regress/expected/isolation_ensure_dependency_activate_node.out
```diff
@@ -1448,6 +1727,383 @@ run_command_on_workers
...
+0              
+run_command_on_workers
+
+(localhost,57637,t,0)
+(localhost,57638,t,0)
+count          
+
+0              
+run_command_on_workers
+
```
onderkalaci: I think we should not see `1` here  as well. But, since dropping the function doesn't remove item from `pg_dist_object`, we end-up with `1` because `master_add_node`  which is called from `print-distributed-objects` -- that's weird btw -- replicates the function.

Maybe for now, just remove it  manually from `pg_dist_object`, and once your DROP FUNCTION pr would do it automatically, we'd remove the manual removal.



So, have the following next to `DROP FUNCTION` call
```SQL

DELETE FROM citus.pg_dist_object WHERE objid = 'add(int,int)'::regprocedure;
```

src/test/regress/specs/isolation_ensure_dependency_activate_node.spec
```diff
@@ -20,6 +20,21 @@ teardown
     DROP TABLE IF EXISTS t2 CASCADE;
     DROP TABLE IF EXISTS t3 CASCADE;
     DROP TYPE IF EXISTS tt1 CASCADE;
+    DROP FUNCTION IF EXISTS add(INT,INT) CASCADE;
+
+    -- Remove the pg_dist_object record manually as we do not yet hook into DROP FUNC
```
thanodnl: did you verify this was required?

the citus drop trigger should already remove the entry from `pg_dist_object`

hanefi: > did you verify this was required?

Yes, I needed to do this as the next permutations increased the total number of entries in the `pg_dist_object` table after `create_distributed_function()` calls for the same function in the test.

> the citus drop trigger should already remove the entry from pg_dist_object

I do not think it did.

> Also, is there any moment the casting doesn't fail? The function just has been deleted one line above

Tough question, will check it out and report it here

PR #2986, opened 2019-09-17
Feature disable object propagation

thanodnl:
DESCRIPTION: Provide a GUC to turn of the new dependency propagation functionality

Expansion on #2954 

In the case the dependency propagation functionality introduced in 9.0 causes issues to a cluster of a user they can turn it off almost completely. The only dependency that will still be propagated and kept track of is the schema to emulate the old behaviour.

GUC to change is `citus.enable_object_propagation`. When set to `false` the functionality will be mostly turned off. Be aware that objects marked as distributed in `pg_dist_object` will still be kept in the catalog as a distributed object. Alter statements to these objects will not be propagated to workers and may cause desynchronisation.

### DIFF
--- a/src/backend/distributed/commands/type.c
+++ b/src/backend/distributed/commands/type.c
@@ -99,6 +99,7 @@ static CreateEnumStmt * RecreateEnumStmt(Oid typeOid);
 static List * EnumValsList(Oid typeOid);
 
 static bool ShouldPropagateTypeCreate(void);
+static bool ShouldPropagateAlterType(const ObjectAddress *address);
 
 
 /*
@@ -209,20 +210,8 @@ PlanAlterTypeStmt(AlterTableStmt *stmt, const char *queryString)
 
 	Assert(stmt->relkind == OBJECT_TYPE);
 
-	if (creating_extension)
-	{
-		/*
-		 * extensions should be created separately on the workers, types cascading from an
-		 * extension should therefor not be propagated.
-		 */
-		return NIL;
-	}
-
-	/*
-	 * Only distributed types should be propagated
-	 */
 	typeAddress = GetObjectAddressFromParseTree((Node *) stmt, false);
-	if (!IsObjectDistributed(typeAddress))
+	if (!ShouldPropagateAlterType(typeAddress))
 	{
 		return NIL;
 	}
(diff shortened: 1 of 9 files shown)


src/backend/distributed/commands/type.c
```diff
@@ -209,20 +210,8 @@ PlanAlterTypeStmt(AlterTableStmt *stmt, const char *queryString)
...
-		 */
-		return NIL;
-	}
-
-	/*
-	 * Only distributed types should be propagated
-	 */
 	typeAddress = GetObjectAddressFromParseTree((Node *) stmt, false);
-	if (!IsObjectDistributed(typeAddress))
+	if (!ShouldPropagateAlterType(typeAddress))
```
onderkalaci: Should we use this flag on `ReplicateAllDependenciesToNode()` and `create_distributed_function()`? I wonder if we can find a point to add this check such that any future object also not propagated?

thanodnl: > I wonder if we can find a point to add this check such that any future object also not propagated?

I have been thinking about this, and is not feasible in the current form of our code. `EnableDDLPropagation` did this somewhat right for its purpose as we putt all calls for `Plan*` and `Process*` functions in a block guarded by that flag.

This will not work for new objects being propagated as they are often routed through overloaded statements by Postgres, eg. `ALTER TYPE ... ADD ATTRIBUTE ...` is parsed into a `AlterTableStmt` struct which is reused for tables and types, and possibly more. We don't want to put the whole handling of `AlterTableStmt` in a block guarded by the `EnableObjectPropagation` flag as that would intervene with changes on distributed tables. For now we need to be diligent and keep adding this check to the objects we want to exclude from distribution such as the upcoming functions.

> Should we use this flag on ReplicateAllDependenciesToNode() and create_distributed_function()?

`ReplicateAllDependenciesToNode()` is a complicated one. Currently it will behave very strange when there are objects in `pg_dist_object`. It will most likely fail to order the dependencies already recorded before the flag was turned off. A solution to this might be to filter `pg_dist_object` by with `SupportedDependenciesByCitus` as that will fail all but the schemas. Those are a requirement to still create on the new node and be able to distribute tables to the node.

Since `create_distributed_function` is a deliberate call by the user, I wonder if we should execute, error or potentially warn when object propagation is turned off. I think executing without a warning is potentially dangerous, the warning should be something like:
```
distributing functions when object propagation is turned off will prevent citus from keeping the function on the coordinator and the workers in sync
```

Completely erroring on the UDF feels funky. I don't think we give any warning or error on `create_distributed_table` when ddl propagation is turned off, so maybe executing without any warning is a sane thing to do.

onderkalaci: > distributing functions when object propagation is turned off will prevent citus from keeping the function on the coordinator and the workers in sync

I think this sounds good.

serprex:
client_min_messages should be set to `error`, not `fatal`. `fatal` produces an error in pg12

thanodnl:
Thanks for the heads up, will create a PR to fix

thanodnl:
https://github.com/citusdata/citus/pull/3000

Issue #2972, opened 2019-09-16
MX Isolation Test Automation


halilozanakgul:
### Reference Tables

| OPERATIONS | INSERT | SELECT | INSERT SELECT | UPDATE, DELETE, UPSERT | COPY | DROP, ALTER, INDEX | TRUNCATE | SELECT FOR UPDATE | CREATE INDEX CONCURRENTLY
|-|-|-|-|-|-|-|-|-|-|
| **INSERT** | ✔️||| ✔️<sub>UPDATE</sub> | ✔️ | ✔️ ||✔️<sup>2</sup>|
| **SELECT** ||||✔️<sup>1</sup><sub>UPSERT</sub>|| ✔️ ||✔️<sup>2</sup>|
| **INSERT SELECT** ||||✔️<sup>1</sup><sub>DELETE</sub>||||✔️<sup>2</sup>|
| **UPDATE, DELETE, UPSERT** | ✔️<sub>UPDATE</sub> |✔️<sup>1</sup><sub>UPSERT</sub>|✔️<sup>1</sup><sub>DELETE</sub>| ✔️<sub>UPDATE</sub> | ✔️<sub>UPDATE</sub> |✔️<sup>1</sup><sub>UPSERT/DROP</sub>|✔️<sup>1</sup><sub>DELETE</sub>| ✔️<sub>UPDATE</sub> |❌[<sup>🔗</sup>](https://github.com/citusdata/citus/issues/2966)
| **COPY** | ✔️ ||| ✔️<sub>UPDATE</sub> | ✔️ |||✔️<sup>2</sup>|
| **DROP, ALTER, INDEX** | ✔️ | ✔️ ||✔️<sup>1</sup><sub>UPSERT/DROP</sub>||||✔️<sup>2</sup><sub>ALTER</sub>|
| **TRUNCATE** ||||✔️<sup>1</sup><sub>DELETE</sub>||||✔️<sup>2</sup>|
| **SELECT FOR UPDATE** |✔️<sup>2</sup>|✔️<sup>2</sup>|✔️<sup>2</sup>|✔️<sub>UPDATE</sub>|✔️<sup>2</sup>|✔️<sup>2</sup><sub>ALTER</sub>|✔️<sup>2</sup>|✔️<sup>2</sup>|✔️<sup>2</sup>

The numbers in the superscript show the file the test is in:
1. [isolation_ref_update_delete_upsert_vs_all_on_mx.spec](https://github.com/citusdata/citus/pull/3044)
[...]

PR #2989, opened 2019-09-18
MX Isolation Tests for Update, Delete and Upsert

halilozanakgul:
Created Isolation tests for update, delete and upsert on MX.
For more info on MX isolation tests checkout [here](https://github.com/citusdata/citus/issues/2972
).

### DIFF
--- a/src/test/regress/specs/isolation_update_delete_upsert_vs_all_on_mx.spec
+++ b/src/test/regress/specs/isolation_update_delete_upsert_vs_all_on_mx.spec
@@ -1,0 +1,146 @@
+# Create and use UDF to send commands from the same connection. Also make the cluster
+# ready for testing MX functionalities.
+setup
+{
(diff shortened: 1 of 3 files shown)


src/test/regress/specs/isolation_update_delete_upsert_vs_all_on_mx.spec
```diff
@@ -0,0 +1,153 @@
...
+        SELECT run_commands_on_session_level_connection_to_node('BEGIN');
+}
+
+step "s2-delete"
+{
+	SELECT run_commands_on_session_level_connection_to_node('DELETE FROM dist_table WHERE id=5');
+}
+
+step "s2-copy"
+{
```
onderkalaci: I think COPY should also touch the same rows, so could you add `5,50` row as well to `COPY` command?

halilozanakgul:
Since this pr is not primarily about INSERT operation. We decided to replace UPSERT with UPDATE instead of INSERT, on in-person discussion.

PR #2990, opened 2019-09-18
Add arguments to `create_distributed_function()`

onderkalaci:
With this commit, we're changing the API for create_distributed_function()
such that users can provide the distribution argument and the colocation
information.

Note: This only implements the subset of what is required. However, this part seems straight-forward to be implemented separately, to keep the overall review easier.

We're adding **2 optional parameters**:
```SQL

CREATE OR REPLACE FUNCTION create_distributed_function(function_name regprocedure,
						       distribution_arg_name text DEFAULT NULL,
						       colocate_with text DEFAULT 'default')
```
- `distribution_arg_name`: Supply the parameter on which you want your function to be distributed. It is either the parameter name, or its index as it'd be used in the function itself ('$1' for the first parameter). If not supplied, the function is distributed, but the distribution argument is left NULL to indicate that this is not distributed on an argument.
- `colocate_with`: Very similar to `create_distributed_table()`'s `colocate_with` parameter.  

DESCRIPTION: Add dist. arg. and colocation parameters to create_distributed_function()

### DIFF
--- a/src/backend/distributed/commands/function.c
+++ b/src/backend/distributed/commands/function.c
@@ -16,21 +16,43 @@
  */
 
 #include "postgres.h"
+#include "funcapi.h"
 
+#include "access/htup_details.h"
 #include "access/xact.h"
 #include "catalog/namespace.h"
 #include "catalog/pg_proc.h"
 #include "distributed/commands.h"
+#include "catalog/pg_type.h"
+#include "distributed/colocation_utils.h"
+#include "distributed/master_protocol.h"
 #include "distributed/metadata_sync.h"
 #include "distributed/metadata/distobject.h"
+#include "distributed/metadata/pg_dist_object.h"
 #include "distributed/multi_executor.h"
 #include "distributed/relation_access_tracking.h"
 #include "distributed/worker_transaction.h"
 #include "utils/builtins.h"
 #include "utils/fmgrprotos.h"
+#include "utils/fmgroids.h"
+#include "utils/lsyscache.h"
+#include "utils/syscache.h"
+
+#define argumentStartsWith(arg, prefix) \
+	(strncmp(arg, prefix, strlen(prefix)) == 0)
 
 /* forward declaration for helper functions*/
 static char * GetFunctionDDLCommand(const RegProcedure funcOid);
+static int GetDistributionArgIndex(Oid functionOid, char *distributionArgumentName,
+								   Oid *distributionArgumentOid);
+static int GetFunctionColocationId(Oid functionOid, char *colocateWithName, Oid
+								   distributionArgumentOid);
+static void EnsureFunctionCanBeColocatedWithTable(Oid functionOid, Oid
+												  distributionColumnType, Oid
+												  sourceRelationId);
+static void UpdateFunctionDistributionInfo(const ObjectAddress *distAddress,
+										   int *distribution_argument_index,
+										   int *colocationId);
 static void EnsureSequentialModeForFunctionDDL(void);
 
 PG_FUNCTION_INFO_V1(create_distributed_function);
@@ -43,18 +65,56 @@ Datum
 create_distributed_function(PG_FUNCTION_ARGS)
 {
 	RegProcedure funcOid = PG_GETARG_OID(0);
+
+	text *distributionArgumentNameText = NULL; /* optional */
+	text *colocateWithText = NULL; /* optional */
+
 	const char *ddlCommand = NULL;
 	ObjectAddress functionAddress = { 0 };
+
+	int distributionArgumentIndex = -1;
+	Oid distributionArgumentOid = InvalidOid;
+	int colocationId = -1;
+
+	char *distributionArgumentName = NULL;
+	char *colocateWithTableName = NULL;
 
 	/* if called on NULL input, error out */
 	if (funcOid == InvalidOid)
 	{
-		ereport(ERROR, (errmsg("create_distributed_function() requires a single "
-							   "parameter that is a valid function or procedure name "
+		ereport(ERROR, (errmsg("the first parameter for create_distributed_function() "
+							   "should be a single a valid function or procedure name "
 							   "followed by a list of parameters in parantheses"),
 						errhint("skip the parameters with OUT argtype as they are not "
 								"part of the signature in PostgreSQL")));
 	}
+
+	if (PG_ARGISNULL(1))
+	{
+		/*
+		 * Using the  default value, so distribute the function but do not set
+		 * the  distribution argument.
+		 */
+		distributionArgumentName = NULL;
+	}
+	else
+	{
+		distributionArgumentNameText = PG_GETARG_TEXT_P(1);
+		distributionArgumentName = text_to_cstring(distributionArgumentNameText);
+	}
+
+	if (PG_ARGISNULL(2))
+	{
+		ereport(ERROR, (errmsg("colocate_with parameter should not be NULL"),
+						errhint("To use the default value, set colocate_with option "
+								"to \"default\"")));
+	}
+	else
+	{
+		colocateWithText = PG_GETARG_TEXT_P(2);
+		colocateWithTableName = text_to_cstring(colocateWithText);
+	}
+
 
 	ObjectAddressSet(functionAddress, ProcedureRelationId, funcOid);
 
@@ -70,6 +130,42 @@ create_distributed_function(PG_FUNCTION_ARGS)
 	SendCommandToWorkers(ALL_WORKERS, ddlCommand);
 
 	MarkObjectDistributed(&functionAddress);
+
+	if (distributionArgumentName == NULL)
+	{
+		/* cannot provide colocate_with without distribution_arg_name */
+		if (pg_strncasecmp(colocateWithTableName, "default", NAMEDATALEN) != 0)
+		{
+			char *functionName = get_func_name(funcOid);
+
+
+			ereport(ERROR, (errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+							errmsg("cannot distribute the function \"%s\" since the "
+								   "distribution argument is not valid ", functionName),
+							errhint("To provide \"colocate_with\" option, the"
+									" distribution argument parameter should also "
+									"be provided")));
+		}
+
+		/* set distribution argument and colocationId to NULL */
+		UpdateFunctionDistributionInfo(&functionAddress, NULL, NULL);
+	}
+	else if (distributionArgumentName != NULL)
+	{
+		/* get the argument index, or error out if we cannot find a valid index */
+		distributionArgumentIndex =
+			GetDistributionArgIndex(funcOid, distributionArgumentName,
+									&distributionArgumentOid);
+
+		/* get the colocation id, or error out if we cannot find an appropriate one */
+		colocationId =
+			GetFunctionColocationId(funcOid, colocateWithTableName,
+									distributionArgumentOid);
+
+		/* if provided, make sure to record the distribution argument and colocationId */
+		UpdateFunctionDistributionInfo(&functionAddress, &distributionArgumentIndex,
+									   &colocationId);
+	}
 
 	PG_RETURN_VOID();
 }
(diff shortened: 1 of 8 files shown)


src/backend/distributed/commands/function.c
```diff
@@ -70,10 +107,297 @@ create_distributed_function(PG_FUNCTION_ARGS)
...
+	distributionArgumentName = text_to_cstring(distributionArgumentNameText);
+	for (argIndex = 0; argIndex < numberOfArgs; ++argIndex)
+	{
+		char *argNameOnIndex = argNames != NULL ? argNames[argIndex] : NULL;
+		Oid argOidOnIndex = argTypes[argIndex];
+
+		/* if user specified distribution_arg_index OR distribution_arg_name */
+		if (distributionArgumentIndex == argIndex ||
+			(argNameOnIndex != NULL &&
+			 strncmp(argNameOnIndex, distributionArgumentName, NAMEDATALEN) == 0))
```
onderkalaci: we should probably use `pg_strncasecmp` instead of `strncmp `

src/backend/distributed/commands/function.c
```diff
@@ -70,10 +107,297 @@ create_distributed_function(PG_FUNCTION_ARGS)
...
+						errmsg("cannot distribute the function \"%s\" since the "
+							   "distribution argument is not valid ", functionName),
+						errhint("Provide a valid distribution_arg_index to "
+								"create_distributed_function()")));
+	}
+	else if (distributionArgumentIndexInput != -1)
+	{
+		/*
+		 * Internal representation for the distributionArgumentIndex
+		 * starts from 0 whereas user facing API starts from 1.
```
marcocitus: Why do we start the user-facing one from 1?

onderkalaci: I couldn't find any Citus UDF that asks for an index, do you remember any?
Picked this mostly because arrays starts from `1` in Postgres, but I can change it back if it feels confusing.

marcocitus:
Do you think we start metadata syncing in this function, or do that in a separate PR?

src/test/regress/expected/distributed_functions.out
```diff
@@ -52,6 +52,190 @@ SELECT * FROM run_command_on_workers('SELECT function_tests.dup(42);') ORDER BY
...
+ERROR:  relation replicated_table_func_test_3 is not distributed
+-- a function cannot be colocated with a reference table
+SELECT create_reference_table('replicated_table_func_test_3');
+ create_reference_table 
+------------------------
+ 
+(1 row)
+
+SELECT create_distributed_function('add_with_param_names(int, int)', distribution_arg_index:=1, colocate_with:='replicated_table_func_test_3');
+ERROR:  cannot colocate function "add_with_param_names" and table "replicated_table_func_test_3" because colocate_with option is only supported for hash distributed tables.
```
marcocitus: What happens if we allow this?

onderkalaci:
> Do you think we start metadata syncing in this function, or do that in a separate PR?

My plan is to do it with a separate PR, just because that seems easier for reviewers. And also that seems a separate problem. To test it now with metadata on workers, I just sync the metadata manually.

src/test/regress/expected/distributed_functions.out
onderkalaci: I think there are two things to consider:

- Hash tables with rep > 1, append/range tables: We do not sync metadata for such distributed tables anyway. So, it doesn't help much to allow, and could be confusing.

- For reference tables: I think, it makes sense to distribute `reference function`. But, to keep the scope smaller and also follow our naming conventions (e.g., we should probably have `create_reference_function()`). Do you think should we go ahead and implement it?

marcocitus: Reference function sounds pretty odd. Anyway, it also complicates the CALL logic a bit so let's maybe punt on this.

onderkalaci:
> We're adding 3 optional parameters:

Thanks to @thanodnl's suggestion, we'd have a single optional parameter, which is `colocate_with`

src/backend/distributed/commands/function.c
onderkalaci: I think, with the new commit, starting from `1` becomes more intuitive because `$1` is the way to access the first parameter in the functions.

src/backend/distributed/commands/function.c
```diff
@@ -91,6 +131,277 @@ CreateFunctionDDLCommandsIdempotent(const ObjectAddress *functionAddress)
...
+		if (colocationId == INVALID_COLOCATION_ID)
+		{
+			colocationId =
+				CreateColocationGroup(ShardCount, ShardReplicationFactor,
+									  distributionArgumentOid);
+
+			createdColocationGroup = true;
+		}
+	}
+	else if (pg_strncasecmp(colocateWithTableName, "none", NAMEDATALEN) == 0)
```
onderkalaci: Well, I'm not sure if `none` makes sense for a function. But, we might keep it to have the same API with `create_distributed_table`

src/backend/distributed/commands/function.c
```diff
@@ -91,6 +131,277 @@ CreateFunctionDDLCommandsIdempotent(const ObjectAddress *functionAddress)
...
+
+	proctup = SearchSysCache1(PROCOID, ObjectIdGetDatum(functionOid));
+	if (!HeapTupleIsValid(proctup))
+	{
+		elog(ERROR, "cache lookup failed for function %u", functionOid);
+	}
+
+	numberOfArgs = get_func_arg_info(proctup, &argTypes, &argNames, &argModes);
+
+	distributionArgumentName = text_to_cstring(distributionArgumentNameText);
```
marcocitus: are we doing this twice? usually I think it's a bit nicer to use char * in internal functions and convert the text early on in the UDF.

src/backend/distributed/commands/function.c
marcocitus: I'd be in favour of removing it. More code more problems.

src/backend/distributed/commands/function.c
```diff
@@ -43,9 +62,20 @@ Datum
 create_distributed_function(PG_FUNCTION_ARGS)
 {
 	RegProcedure funcOid = PG_GETARG_OID(0);
+
+	text *distributionArgumentNameText = PG_GETARG_TEXT_P(1);
```
marcocitus: Maybe this should also be optional, such that you can use create_distributed_function to just use a function in queries without other magic?

onderkalaci: Yeah, that also sounds good. Let me think a bit more

onderkalaci: I've done that, looks good to me as well.

src/backend/distributed/sql/udfs/create_distributed_function/latest.sql
```diff
@@ -1,7 +1,11 @@
-CREATE OR REPLACE FUNCTION create_distributed_function(function_name regprocedure)
+CREATE OR REPLACE FUNCTION create_distributed_function(function_name regprocedure,
+						       distribution_arg_name text DEFAULT '',
```
marcocitus: why not DEFAULT NULL?

src/backend/distributed/commands/function.c
```diff
@@ -91,6 +156,305 @@ CreateFunctionDDLCommandsIdempotent(const ObjectAddress *functionAddress)
...
+
+	char *updateQuery =
+		"UPDATE citus.pg_dist_object SET distribution_argument_index = NULL, "
+		"colocationid = NULL WHERE classid = $1 AND objid=$2 AND objsubid=$3";
+
+	/* UPDATE query can only be useful if the entry is already in the pg_dist_object */
+	Assert(IsObjectDistributed(distAddress));
+
+	spiStatus = ExecuteCommandAsSuperuser(updateQuery, paramCount, paramTypes,
+										  paramValues);
```
marcocitus: We probably should not overuse this pattern. In general it is a bit dangerous to enter SQL land as superuser. If an administrator sets up a trigger on citus.pg_dist_object then a user might be able to exploit the fact that the trigger is called as superuser. I don't have any specific concerns here, but I'm weary of dangerous patterns being repeated later on. We should be adding some warning signs.

src/backend/distributed/sql/udfs/create_distributed_function/latest.sql
onderkalaci: Just because it complicates the `C` code a bit, you'd need to use `PG_ARGISNULL` before accessing the value. But, that's  actually a good practice, otherwise if the users passes `NULL` the function crashes.  So, fixing it.

src/backend/distributed/commands/function.c
onderkalaci: Good to know about this case. I used it just for simplicity (with copy & paste), now switched to `heap_tuple_modify()`

onderkalaci: Also added comment to `ExecuteCommandAsSuperuser `

src/backend/distributed/commands/function.c
```diff
@@ -91,6 +187,309 @@ CreateFunctionDDLCommandsIdempotent(const ObjectAddress *functionAddress)
...
+	*distributionArgumentOid = InvalidOid;
+
+	proctup = SearchSysCache1(PROCOID, ObjectIdGetDatum(functionOid));
+	if (!HeapTupleIsValid(proctup))
+	{
+		elog(ERROR, "cache lookup failed for function %u", functionOid);
+	}
+
+	numberOfArgs = get_func_arg_info(proctup, &argTypes, &argNames, &argModes);
+
```
marcocitus: Can we move `ReleaseSysCache(proctup);` up here?

src/backend/distributed/commands/function.c
```diff
@@ -91,6 +187,309 @@ CreateFunctionDDLCommandsIdempotent(const ObjectAddress *functionAddress)
...
+		 * Internal representation for the distributionArgumentIndex
+		 * starts from 0 whereas user facing API starts from 1.
+		 */
+		distributionArgumentIndex -= 1;
+	}
+
+	/*
+	 * If the user provided "$paramIndex", we're unnecessarily looping, but we prefer
+	 * this for the simplicity of the code.
+	 */
```
marcocitus: When I read this I didn't understand why we needed to go into the loop at all. Maybe add that we still need to find the OID of the argument.

btw, wouldn't something like:
```c
*distributionArgumentOid = argTypes[distributionArgumentIndex];

/*
 * Internal representation for the distributionArgumentIndex
 * starts from 0 whereas user facing API starts from 1.
 */
return distributionArgumentIndex  - 1;
```

do the trick?

src/test/regress/expected/distributed_functions.out
```diff
@@ -39,7 +39,7 @@ CREATE TYPE dup_result AS (f1 int, f2 text);
 CREATE FUNCTION dup(int) RETURNS dup_result
     AS $$ SELECT $1, CAST($1 AS text) || ' is text' $$
     LANGUAGE SQL;
-SELECT create_distributed_function('dup(int)');
+SELECT create_distributed_function('dup(int)', '$1');
```
marcocitus: I think we no longer have tests without distribution arguments, maybe restore these.

onderkalaci:
> Apart from tests without distribution argument index and minor comments, this seems fine.

We have the test at the [bottom](https://github.com/citusdata/citus/pull/2990/files#diff-d5180900d1a6ab45152d77180bad5560R217) of the file, checking for NULL in the `pg_dist_object.distribution_argument_index` and `pg_dist_object.colocationid` after successful call to the function

src/backend/distributed/commands/function.c
onderkalaci: I think we probably should not  because we're  still using  `argTypes ` and `argNames ` below. Other callers of `get_func_arg_info` does keep until they're done processing with the `argTypes ` and `argNames `. Let's follow the postgres convention unless you have further objection?

src/backend/distributed/commands/function.c
onderkalaci: > do the trick?

I guess no, so there are two cases:  (a) the user types in like function parameter: `$1` (b) gives the name of the parameter : `val1`

For (a) we don't need to loop over array, we can just use `argTypes[distributionArgumentIndex];`. However, for (b) we need to iterate over the arguments  and check whether the name exists in the parameters of the function. If we find, we can use `argTypes[distributionArgumentIndex]`, otherwise we error out.

So, the comment actually aims to  explain that. I expanded the comment a bit more to clarify.

src/backend/distributed/commands/function.c
marcocitus: `get_func_arg_info` explicitly takes the trouble of allocating new memory and fresh new datums for these arrays, so they aren't tied to the tuple.

https://github.com/postgres/postgres/blob/REL_11_STABLE/src/backend/utils/fmgr/funcapi.c#L909

src/backend/distributed/commands/function.c
marcocitus: I meant in the `if (argumentStartsWith(distributionArgumentName, "$"))` block above .

marcocitus:
> We have the test at the bottom of the file, checking for NULL in the pg_dist_object.distribution_argument_index and pg_dist_object.colocationid after successful call to the function

Fair, but we no longer check whether we can run a query as a before.

src/backend/distributed/commands/function.c
onderkalaci: I see, yes, that seem to simplify the code. Let me do that =, thanks

onderkalaci:
>Fair, but we no longer check whether we can run a query as a before.

> I meant in the if (argumentStartsWith(distributionArgumentName, "$")) block above .

I applied your feedback on these two, so I'll wait for you to re-approve before merging in case you have any further comments on top

PR #2991, opened 2019-09-18
Feature/drop function

thanodnl:
DESCRIPTION: Add DROP FUNCTION propagation for distributed functions

includes #2986 

For functions that have been distributed and later dropped we want to send the drop commands to the workers to clean up the function. As one might drop multiple functions at once we filter the list to only contain distributed functions.

### DIFF
--- a/src/backend/distributed/commands/function.c
+++ b/src/backend/distributed/commands/function.c
@@ -26,21 +26,23 @@
 #include "access/xact.h"
 #include "catalog/namespace.h"
 #include "catalog/pg_proc.h"
-#include "distributed/commands.h"
 #include "catalog/pg_type.h"
+#include "commands/extension.h"
 #include "distributed/colocation_utils.h"
-#include "distributed/master_protocol.h"
+#include "distributed/commands.h"
+#include "distributed/commands/utility_hook.h"
 #include "distributed/maintenanced.h"
-#include "distributed/metadata_sync.h"
+#include "distributed/master_protocol.h"
 #include "distributed/metadata/distobject.h"
 #include "distributed/metadata/pg_dist_object.h"
+#include "distributed/metadata_sync.h"
 #include "distributed/multi_executor.h"
 #include "distributed/relation_access_tracking.h"
 #include "distributed/worker_transaction.h"
 #include "storage/lmgr.h"
 #include "utils/builtins.h"
-#include "utils/fmgrprotos.h"
 #include "utils/fmgroids.h"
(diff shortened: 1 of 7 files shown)


src/backend/distributed/commands/function.c
```diff
@@ -157,3 +162,155 @@ EnsureSequentialModeForFunctionDDL(void)
...
+	 */
+	QualifyTreeNode((Node *) stmt);
+
+	/*
+	 * map all functions to their object addresses and step through them in lock step to
+	 * create two lists with all distributed functions and their addresses.
+	 */
+	objectAddresses = FunctionListToObjectAddresses(oldObjectWithArgsList,
+													stmt->missing_ok);
+	forboth(objectWithArgsListCell, oldObjectWithArgsList, addressCell, objectAddresses)
```
marcocitus: Maybe just make a `FunctionToObjectAddress` function that we call in the loop below. The current structure is neither more efficient nor more readable.

src/backend/distributed/commands/function.c
```diff
@@ -157,3 +162,155 @@ EnsureSequentialModeForFunctionDDL(void)
...
+	}
+
+	return true;
+}
+
+
+/*
+ * PlanDropFunctionStmt gets called during the planning phase of a DROP FUNCTION statement
+ * and returns a list of DDLJob's that will drop any distributed functions from the
+ * workers.
```
marcocitus: It's worth noting that this handles DROP FUNCTION statements in which a subset of functions are distributed.

src/backend/distributed/commands/function.c
```diff
@@ -157,3 +162,155 @@ EnsureSequentialModeForFunctionDDL(void)
...
+	/* remove the entries for the distributed objects on dropping */
+	foreach(addressCell, distributedFunctionAddresses)
+	{
+		ObjectAddress *address = (ObjectAddress *) lfirst(addressCell);
+		UnmarkObjectDistributed(address);
+	}
+
+	stmt->objects = distributedObjectWithArgsList;
+	dropStmtSql = DeparseTreeNode((Node *) stmt);
+	stmt->objects = oldObjectWithArgsList;
```
marcocitus: Copying the object before setting the objects seems cleaner. In that case we can have an explicit `distributedDropFunction` vs. `overallDropFunction`, meaning the code clearly documents what it is trying to achieve.

thanodnl:

PR #2992, opened 2019-09-18
Feature/distributed function set schema

thanodnl:
DESCRIPTION: Implement ALTER FUNCTION ... SET SCHEMA ... for distributed functions

TODO
 - [x] Deparser logic for `ALTER FUNCTION ... SET SCHEMA ...`
 - [x] Update tests after success
 - [ ] Add test that triggers the codepath for a function that does not exist, citus will error out before postgres will see the statement

### DIFF
--- a/src/backend/distributed/commands/function.c
+++ b/src/backend/distributed/commands/function.c
@@ -31,6 +31,7 @@
 #include "distributed/colocation_utils.h"
 #include "distributed/commands.h"
 #include "distributed/commands/utility_hook.h"
+#include "distributed/deparser.h"
 #include "distributed/maintenanced.h"
 #include "distributed/master_protocol.h"
 #include "distributed/metadata/distobject.h"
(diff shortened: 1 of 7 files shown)


src/backend/distributed/commands/function.c
```diff
@@ -157,3 +162,151 @@ EnsureSequentialModeForFunctionDDL(void)
...
+ * statements targeting the function should be propagated.
+ */
+static bool
+ShouldPropagateAlterFunction(const ObjectAddress *address)
+{
+	if (creating_extension)
+	{
+		/*
+		 * extensions should be created separately on the workers, functions cascading
+		 * from an extension should therefor not be propagated.
```
SaitTalhaNisanci: should **therefore** not be ...

thanodnl: Thanks!

thanodnl:

PR #2995, opened 2019-09-18
Feature: distributed function alter actions

thanodnl:
DESCRIPTION: Implement ALTER FUNCTION ... action [...] for distributed functions

TODO
 - [ ] Test for changes being applied remotely
 - [ ] Add tests for SET actions

### DIFF
--- a/src/backend/distributed/commands/function.c
+++ b/src/backend/distributed/commands/function.c
@@ -31,6 +31,7 @@
 #include "distributed/colocation_utils.h"
 #include "distributed/commands.h"
 #include "distributed/commands/utility_hook.h"
+#include "distributed/deparser.h"
 #include "distributed/maintenanced.h"
 #include "distributed/master_protocol.h"
 #include "distributed/metadata/distobject.h"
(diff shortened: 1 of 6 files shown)


src/backend/distributed/commands/function.c
```diff
@@ -157,3 +161,90 @@ EnsureSequentialModeForFunctionDDL(void)
...
+
+const ObjectAddress *
+AlterFunctionStmtObjectAddress(AlterFunctionStmt *stmt, bool missing_ok)
+{
+	Oid funcOid = InvalidOid;
+	ObjectAddress *address = NULL;
+
+	funcOid = LookupFuncWithArgs(OBJECT_FUNCTION, stmt->func, missing_ok);
+	address = palloc0(sizeof(ObjectAddress));
+	ObjectAddressSet(*address, ProcedureRelationId, funcOid);
```
marcocitus: These 3 lines seem to repeat in several PRs for different calling contexts. That should just be a utility function that we used in multiple places.

src/backend/distributed/commands/function.c
```diff
@@ -157,3 +161,90 @@ EnsureSequentialModeForFunctionDDL(void)
...
+	commands = list_make3(DISABLE_DDL_PROPAGATION,
+						  (void *) sql,
+						  ENABLE_DDL_PROPAGATION);
+
+	return NodeDDLTaskList(ALL_WORKERS, commands);
+}
+
+
+const ObjectAddress *
+AlterFunctionStmtObjectAddress(AlterFunctionStmt *stmt, bool missing_ok)
```
marcocitus: Needs a function comment. We do not expect readers to open this file with full knowledge of the dependency system.

src/backend/distributed/commands/function.c
thanodnl: This has been implemented as a side effect of the comment on `forboth`, will use this function throughout.

src/backend/distributed/commands/function.c
thanodnl: Will go through and comment all functions

thanodnl:

PR #2998, opened 2019-09-19
Use Adaptive Executor for call.c

pykello:
@serprex just for the sake of comparison, this is how it would look if we had used adaptive executor for call.c.


src/backend/distributed/utils/maintenanced.c
```diff
@@ -374,6 +374,7 @@ CitusMaintenanceDaemonMain(Datum main_arg)
 
 			InvalidateMetadataSystemCache();
 			StartTransactionCommand();
+			PushActiveSnapshot(GetTransactionSnapshot());
```
pykello: These are none-related, but I needed these to be able to run tests.

marcocitus:
This seems cleaner.

PR #2999, opened 2019-09-19
Disable flappy test

JelteF:
This test caused random failures because it was non deterministic which of the
create_distributed_table statements finished the earliest. Locally this test
would randomly deadlock the sessions because of this. Commiting session 1
earlier fixed this. 

Example of failing test in CI:
https://circleci.com/gh/citusdata/citus/34188

### DIFF
--- a/src/test/regress/expected/isolation_citus_dist_activity.out
+++ b/src/test/regress/expected/isolation_citus_dist_activity.out
@@ -41,16 +41,16 @@ step s3-view-worker:
 
 query          query_hostname query_hostport master_query_host_namemaster_query_host_portstate          wait_event_typewait_event     usename        datname        
 
-SELECT worker_apply_shard_ddl_command (102217, 'public', '
+SELECT worker_apply_shard_ddl_command (1300004, 'public', '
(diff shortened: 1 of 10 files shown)


thanodnl:
Looking back how this permutation got to be is this commit: https://github.com/citusdata/citus/pull/2882/commits/18282dab90b19ac5ea7ee6a88069191461afa8af

The issue is a postgres bug where two concurrent
```sql
CREATE SCHEMA IF NOT EXISTS x;
```

Could error in a unique constraint violation on the catalog as the existence check is done earlier compared to when the insert will happen, without rechecking its existence after the more aggressive lock is acquired.

I do think we need all 3 sessions to block in a deterministic order to trigger that issue were we not to upsert our `pg_dist_object` entry correctly. I don't think this patch will execute that scenario anymore.

So we could
 - merge this patch and call it a day
 - remove the 4th session from this permutation (and the rest of the suite as I think this is the only permutation that uses it)
 - go back to that commit and recreate the situation

Since this is essentially a concurrency bug of postgres I opt for removing the 4th session and this permutation.

thanodnl:
misclick on the closing :(

PR #3002, opened 2019-09-20
Enhance pg upgrade tests

SaitTalhaNisanci:
This PR adds some more tests for pg upgrade tests:
- Basic distributed type test
- Truncate/drop table
- We store the index of distribution column, and when a column with an
index that is smaller than distribution column index is dropped before
a pg upgrade, the index should still match the distribution column after
the upgrade

### DIFF
--- a/src/test/regress/expected/upgrade_distributed_table_after.out
+++ b/src/test/regress/expected/upgrade_distributed_table_after.out
@@ -28,72 +28,6 @@ SELECT * FROM t WHERE a = 11;
  11
 (1 row)
 
--- test task-tracker executor
-SET citus.task_executor_type TO 'task-tracker';
-SELECT * FROM t WHERE a = 10;
- a  
-----
- 10
-(1 row)
-
-SELECT * FROM t WHERE a = 11;
- a  
-----
- 11
-(1 row)
-
-SELECT * FROM t WHERE a < 3 ORDER BY a;
- a 
----
- 1
- 2
-(2 rows)
-
-RESET citus.task_executor_type;
--- test adaptive executor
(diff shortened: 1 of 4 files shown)


src/test/regress/expected/upgrade_distributed_table_after.out
```diff
@@ -15,5 +15,125 @@ SELECT * FROM t WHERE a = 1;
...
+ a 
+---
+ 1
+ 2
+(2 rows)
+
+RESET citus.task_executor_type;
+-- test real-time executor
+SET citus.task_executor_type to "real-time";
+SELECT * FROM t WHERE a = 10;
```
marcocitus: I think testing the different executors after upgrade is probably not very interesting. We're mainly interested in what happens to metadata/catalogs during the upgrade.

src/test/regress/expected/upgrade_distributed_table_after.out
```diff
@@ -15,5 +15,125 @@ SELECT * FROM t WHERE a = 1;
...
+ 11
+ 22
+ 33
+(3 rows)
+
+SELECT * FROM t_ab WHERE b = 11;
+ b  
+----
+ 11
+(1 row)
```
marcocitus: Good to see this works!

src/test/regress/expected/upgrade_distributed_table_after.out
SaitTalhaNisanci: @marcocitus I thought that we could use this test both in postgres and citus upgrades. But if it doesnt make sense to include this for postgres upgrades I can put this to only citus upgrade tests.

PR #3003, opened 2019-09-20
Add Citus upgrade tests with its job

### LINKED ISSUE #3042: Add citus upgrade test and job
We should have citus upgrade tests as we have postgres upgrade tests. This will eliminate the upgrade tests that we do in release tests.

SaitTalhaNisanci:
This PR adds citus upgrade tests with a job. There is only one image for different postgres versions. The format to run a citus upgrade test is:

```bash
install-and-test-ext --target check-citus-upgrade --citus-pre-tar /install-pg11-citusv8.0.0.tar --citus-post-tar /home/circleci/project/install-11.tar
```
It is currently test for pg11, once we release 9.0 pg12 can also be added.

Currently the image in in my personal docker hub account and that will be moved to `citus` docker hub, until then this should not be merged.

As there is a lot of common functionality with citus and pg upgrades, I have created a `upgrade_common.py` which contains common things for upgrade such as starting database.

Tests are run both in mixed mode and normal mode.

Fixes #3042.


.circleci/config.yml
```diff
@@ -228,6 +228,38 @@ jobs:
           name: 'Install and test postgres upgrade'
           command: 'install-and-test-ext --target check-upgrade --old-pg-version 11 --new-pg-version 12'        
 
+
+
+  test-10_check-citus-upgrade:
+    docker:
+      - image: 'heisenberg302/citusupgradetester:latest'
```
SaitTalhaNisanci: I will change this before merging.

src/test/regress/upgrade/citus_upgrade_test.py
```diff
@@ -0,0 +1,154 @@
...
+    else:
+        restart_databases(config.bindir, config.datadir)
+        run_alter_citus(config.bindir)
+        verify_upgrade(config)
+
+    run_test_on_coordinator(config, AFTER_CITUS_UPGRADE_COORD_SCHEDULE)
+    remove_citus_master_artifacts(config.pg_version)
+
+def main(config):
+    run_citus_upgrade_test(config=config, mixed_mode=False)
```
JelteF: It never runs `mixed_mode`?

src/test/regress/upgrade/citus_upgrade_test.py
```diff
@@ -0,0 +1,154 @@
...
+                       port, AFTER_CITUS_UPGRADE_SCHEDULE)
+
+
+def install_citus(citus_version, pg_version):
+    with utils.cd('/'):
+        subprocess.call(
+            ['tar', 'xvf', './install-pg{}-citus{}.tar'.format(pg_version, citus_version)])
+
+def install_citus_master(pg_version):
+    with utils.cd('~/project'):
```
thanodnl: I don't think we need to cd into `~/project`, can't we directly extract `~/project/install-{}.tar` from the root?

src/test/regress/upgrade/citus_upgrade_test.py
```diff
@@ -0,0 +1,154 @@
...
+        run_pg_regress(config.bindir, config.pg_srcdir,
+                       port, AFTER_CITUS_UPGRADE_SCHEDULE)
+
+
+def install_citus(citus_version, pg_version):
+    with utils.cd('/'):
+        subprocess.call(
+            ['tar', 'xvf', './install-pg{}-citus{}.tar'.format(pg_version, citus_version)])
+
+def install_citus_master(pg_version):
```
thanodnl: the functions `install_citus`, `install_citus_master`, `remove_citus_artifacts ` and `remove_citus_master_artifacts` seem to be sharing a lot of code. Can we reduce this to 2 functions:|
- `install_citus`
- `remove_citus`

that take a version argument where `master` is a valid version. We would just run a different path into the tar function extracting the files.

src/test/regress/sql/after_citus_upgrade.sql
```diff
@@ -0,0 +1 @@
+\dx;
```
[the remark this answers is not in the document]
JelteF: This PR hopfully fixes that in the future: https://github.com/citusdata/citus/pull/3051

.circleci/config.yml
SaitTalhaNisanci: We only need 1 postgres version to execute a citus upgrade but I test both `citus_initial_version pg10 -> citus_master pg10` and `citus_initial_version pg11 -> citus_master pg11`. For example, currently in `citusupgradetester` image we have citus artifacts built for:
```
citus v8.0.0 pg10
citus v8.1.0 pg10
citus v8.2.0 pg10
citus v8.3.0 pg10
citus v8.0.0 pg11
citus v8.1.0 pg11
citus v8.2.0 pg11
citus v8.3.0 pg11
```

They are already in the image therefore we dont spend any time installing them during testing. Otherwise we will need to:
- Download citus with the given version
- Install citus

Doing this for each combination is not fast. I have not recorded the time it takes to do these steps but from my observation it is worth generating the citus artifacts during build time of images.

There was a branch in the repository, which is https://github.com/citusdata/the-process/tree/citusUpgradeTester. I have created a PR now but our images have changed therefore the PR will need to be updated, yet you can still take a look at it to see what is going on there, https://github.com/citusdata/the-process/pull/31

src/test/regress/upgrade/README.md
```diff
@@ -1,44 +1,88 @@
...
+
+## Citus Upgrade Test
 
+Citus upgrade test is used for testing citus version upgrades from specific version to master. The purpose of this test is to ensure that a newly made change does not result in unexpected upgrade errors.
 
+The citus upgrade test is designed to be run on a docker image for CircleCI, so we won't explain here how to run it in your local.
+
+Currently the citus upgrade test assumes that:
+
+- In the root directory("\\") there are citus installation tar files for postgres versions. The tar files are used to install specific citus version with a postgres version. The tar files format is `install-pg{pg-major}-citus{citus-version}.tar` such as install-pg11-citusv8.0.0.tar.
```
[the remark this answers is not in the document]
SaitTalhaNisanci: Yeah I dont know why I tried to escape that in a README :)

.circleci/config.yml
thanodnl: makes sense to package the citus artifacts, can imagine that will be faster compared to installing from source in the test.

But we don't need the pg11 + its citus versions when running the pg10 test, could we instead make 2 images? Separate images for `pg10` and `pg11`.

For the artifacts, would it make sense to instead have them in tar balls to download the debs from our repo during the build of the image? That way we still only have to download them once (image build time) but we have guarantees that we are actually testing to our released versions and it is probably easier keeping it up to date.

downloading debs: https://unix.stackexchange.com/questions/408346/how-to-download-package-not-install-it-with-apt-get-command

src/test/regress/sql/after_citus_upgrade.sql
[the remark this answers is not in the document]
SaitTalhaNisanci: Yeah I remember this from your previous reviews but it is possible to miss adding this for some files. So good to add that to our editor config.

.circleci/config.yml
SaitTalhaNisanci: Yeah technically we dont need pg11 while testing pg10. But I feel like having more images is harder to maintain. What is the downside of having pg11 in the image while testing pg10? It increases the size of image, but it is very little compared to the total size. And downloading images is not a bottleneck in our tests.

I think we can download artifacts from our repo. I am not sure how much of a change it is, but that can also work technically.

src/test/regress/upgrade/citus_upgrade_test.py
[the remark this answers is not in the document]
SaitTalhaNisanci: We run `run_citus_upgrade_test` for mixed_mode = `False` and `True`. Therefore we cant put the body of function here. However currently I get an error when I enable the mixed mode. You can see the error in the PR description. I will look into the error so that we can test `mixed_mode` enabled as well.

src/test/regress/sql/before_citusv8.0.0_upgrade.sql
```diff
@@ -0,0 +1 @@
+\dx;
```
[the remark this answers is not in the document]
SaitTalhaNisanci: I have removed the verify version step from schedules, which is now done in the python script.

src/test/regress/upgrade/citus_upgrade_test.py
```diff
@@ -0,0 +1,154 @@
...
+    with utils.cd('~/project'):
+        abs_tar_path = os.path.abspath('./install-{}.tar'.format(pg_version))
+        with utils.cd('/'):
+            subprocess.call(['tar', 'xvf', abs_tar_path])
+
+def remove_citus_master_artifacts(pg_version):
```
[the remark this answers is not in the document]
SaitTalhaNisanci: I removed this file and use 'remove_citus'.

.circleci/config.yml
```diff
@@ -271,6 +271,32 @@ jobs:
...
+    docker:
+      - image: 'citus/citusupgradetester-11:dev'
+    working_directory: /home/circleci/project     
+    steps:
+      - {attach_workspace: {at: .}}
+      - run: 
+          name: 'Install and test citus upgrade'
+          command: |
+           chown -R circleci:circleci /home/circleci
+           install-and-test-ext --target check-citus-upgrade --citus-pre-tar /install-pg11-citusv8.0.0.tar --citus-post-tar /home/circleci/project/install-11.tar
```
thanodnl: `--citus-post-tar /home/circleci/project/install-11.tar` is very redundant, I wonder if that one can be deduced in the `install-and-test-ext` script based on the PG major version

src/test/regress/upgrade/README.md
```diff
@@ -1,44 +1,88 @@
...
+- `install-{pg_major}.tar` tars exist in `$HOME/project` folder. These tars should contain the checked out citus installation files. Currently build job creates these tars.
+
+How the citus upgrade test work:
+
+- The script takes postgres version and the initial citus version.
+- It installs the given citus version from `install-pg{pg-major}-citus{citus-version}.tar`.
+- It creates a citus cluster(1 coordinator 2 workers).
+- It runs `before_citus{citus_version}_upgrade` schedule to check that the version is installed correctly.
```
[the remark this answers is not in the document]
thanodnl: This part needs to be updated after changes.
Could you review the whole README to reflect latest changes?

src/test/regress/upgrade/config.py
```diff
@@ -23,10 +54,13 @@ def __init__(self, arguments):
 WORKER1 = 'worker1'
 WORKER2 = 'worker2'
 NODE_NAMES = [COORDINATOR_NAME, WORKER1, WORKER2]
+COORDINATOR_PORT = 9700
```
thanodnl: can we use the same ports as in our normal tests

src/test/regress/upgrade/upgrade_common.py
```diff
@@ -0,0 +1,97 @@
...
+            '--log', os.path.join(abs_data_path, 'logfile_' + node_name)
+        ]
+        subprocess.call(command)
+
+
+def initialize_citus_cluster(old_bindir, old_datadir, settings):
+    initialize_db_for_cluster(old_bindir, old_datadir, settings)
+    start_databases(old_bindir, old_datadir)
+    create_citus_extension(old_bindir)
+    add_workers(old_bindir)
```
thanodnl: nit; missing newline

src/test/regress/upgrade/citus_upgrade_test.py
```diff
@@ -0,0 +1,117 @@
...
+import utils
+
+from docopt import docopt
+
+from config import (
+    CitusUpgradeConfig, NODE_PORTS, COORDINATOR_NAME, CITUS_VERSION_SQL, MASTER_VERSION,
+    NODE_NAMES, USER, WORKER1PORT, MASTER, HOME,
+    AFTER_CITUS_UPGRADE_COORD_SCHEDULE, BEFORE_CITUS_UPGRADE_COORD_SCHEDULE
+)
+from upgrade_common import initialize_temp_dir, initialize_citus_cluster, run_pg_regress, stop_databases
```
thanodnl: maybe import as common and use `common.<name>` in the code. Makes it slightly easier to follow while reading where the code is located when called.

src/test/regress/sql/after_citus_upgrade_coord.sql
```diff
@@ -0,0 +1,17 @@
...
+SELECT * FROM t WHERE a = 1;
+
+INSERT INTO t SELECT * FROM generate_series(10, 15);
+
+SELECT * FROM t WHERE a = 10;
+SELECT * FROM t WHERE a = 11;
+
+TRUNCATE TABLE t;
+
+SELECT * FROM T;
```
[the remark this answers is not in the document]
thanodnl: we can add tests later I think, unless @SaitTalhaNisanci has some extra time now.

.circleci/config.yml
SaitTalhaNisanci: Yeah we can do that.

Issue #2326, opened 2018-08-02
Provide a UDF for cleaning up metadata on nodes for MX


aamederen:
When a user issues `start_metadata_sync_to_node` command, the coordinator sends metadata clearing commands, followed by metadata creating commands to the target node. 

When a user issues `stop_metadata_sync_to_node`, coordinator just marks the node and does nothing else. 

It would be good to have a way to clear the metadata after `stop_metadata_sync_to_node` command. 

This can be done via wrapping the clear steps of `start_metadata_sync_to_node` and expose it via a UDF.

onderkalaci:
One idea that came-up with @marcocitus is to add a `clear bool DEFAULT false` parameter to the `stop_metadata_sync_to_node `. This seems to cause less trouble compared to having another UDF because otherwise we'd need to think about the edges of what happens operation X happens in between these calls etc.

PR #3005, opened 2019-09-20
Sync metadata to worker nodes after create_distributed_function

onderkalaci:
Since the distributed functions are useful when the workers have
metadata, we automatically sync it.

Also, after master_add_node(). We do it lazily and let the deamon
sync it. That's mainly because the metadata syncing cannot be done
in transaction blocks, and we don't want to add lots of transactional
limitations to master_add_node() and create_distributed_function().

- [x] One test is almost consistently failing on PG 10, still opened the PR
to get some feedback.


src/backend/distributed/test/metadata_sync.c
```diff
@@ -80,8 +80,36 @@ wait_until_metadata_sync(PG_FUNCTION_ARGS)
 	uint32 timeout = PG_GETARG_UINT32(0);
 	int waitResult = 0;
 
-	MultiConnection *connection = GetNodeConnection(FORCE_NEW_CONNECTION,
-													"localhost", PostPortNumber);
+	List *workerList = ActivePrimaryNodeList(NoLock);
```
onderkalaci: @pykello FYI on the change here. Let me know if there is anything that doesn't make sense to you.

src/backend/distributed/utils/node_metadata.c
```diff
@@ -472,9 +473,22 @@ ActivateNode(char *nodeName, int nodePort)
...
+
 		EnsureNoModificationsHaveBeenDone();
-		ReplicateAllDependenciesToNode(nodeName, nodePort);
+		ReplicateAllDependenciesToNode(nodeName, nodePort,
+									   &clusterHasDistributedFunction);
 		ReplicateAllReferenceTablesToNode(nodeName, nodePort);
+
+		/*
+		 * Let the maintanince deamon do the hard work of syncing the metadata. We prefer
+		 * this because otherwise node activation mig
```
serprex: mig?

src/backend/distributed/test/metadata_sync.c
```diff
@@ -80,8 +80,36 @@ wait_until_metadata_sync(PG_FUNCTION_ARGS)
...
+		if (workerNode->hasMetadata && !workerNode->metadataSynced)
+		{
+			waitNotifications = true;
+			break;
+		}
+	}
+
+	/*
+	 * If all the metadata nodes have already been synced, we should not wait.
+	 * That's primarily because the maintenance deamon might have already sent
```
pykello:
```suggestion
	 * That's primarily because the maintenance daemon might have already sent
```

src/backend/distributed/commands/function.c
```diff
@@ -467,3 +480,34 @@ EnsureSequentialModeForFunctionDDL(void)
...
+		{
+			/*
+			 * Let the maintanince deamon do the hard work of syncing the metadata. We prefer
+			 * this because otherwise node activation might fail withing transaction blocks.
+			 */
+			LockRelationOid(DistNodeRelationId(), ExclusiveLock);
+			MarkNodeHasMetadata(workerNode->workerName, workerNode->workerPort, true);
+
+			/* let the maintanince deamon know about the metadata sync */
+			TriggerMetadataSync(MyDatabaseId);
```
pykello: Calling this more than once is wasteful.

src/backend/distributed/commands/function.c
```diff
@@ -113,6 +119,13 @@ create_distributed_function(PG_FUNCTION_ARGS)
 	UpdateFunctionDistributionInfo(&functionAddress, distributionArgumentIndex,
 								   colocationId);
 
+	/*
+	 * Once we have at least one distributed function/procedure, we sync the
+	 * metadata to nodes so that the function/procedure delegation can be handled
+	 * locally on the nodes.
+	 */
+	TriggerSyncMetadataToPrimaryNodes();
```
pykello: 1. We already do `SendCommandToWorkers()` few lines above, so doing metadata sync here (and not postponing it to maintenanced) won't add any transactional limitations.
2. In addition to that, this is wasteful, since it causes **all** metadata on workers to be dropped and recreated from scratch.

onderkalaci: >We already do SendCommandToWorkers() few lines above, so doing metadata sync here (and not delegating it to maintenanced) won't add any transactional limitations.

We don't support:
```SQL
BEGIN;
SELECT start_metadata_Sync_to_node('localhost', 9700);
ERROR:  start_metadata_sync_to_node cannot run inside a transaction block
```

So, if we sync metadata within `create_distributed_function()/master_add_node()`, we'd impose these limitations on those UDFs as well. That's why, doing it lazily seems a better idea.

>n addition to that, this is wasteful, since it causes all metadata on workers to be dropped and recreated from scratch.

Is it? The code looks like we only to `if (workerNode->hasMetadata && !workerNode->metadataSynced)` [here](https://github.com/citusdata/citus/blob/ff100b2720e3efed2241b969ca560c86859851a6/src/backend/distributed/metadata/metadata_sync.c#L1334). Am I missing anything?

Note: `TriggerSyncMetadataToPrimaryNodes` only syncs when `if (!workerNode->hasMetadata)`, so maybe I should rename the function to reflect it?

onderkalaci:
>Why not just propagate metadata by default?

As I mentioned in the above command, initial syncing cannot be done in a transaction block. 


> master_add_node() -> master_update_node().

Sorry, I couldn't follow this comment. The idea for this PR is that, if you do `create_distributed_function()` we trigger metadata sync to nodes, because that's what CALL/SELECT propagation relies on. Similarly, if a `master_add_node` happens in the cluster when there is already a distributed function, we trigger metadata sync to that node as well.


> Metadata propagation can be done in transaction blocks. It cannot be done in a "prepared transaction", because Citus doesn't support distributed prepared transactions.

Initial sync cannot be done in transaction blocks.

pykello:
@onderkalaci your answers make sense, I had misunderstood some of the points.

marcocitus:
> Why not just propagate metadata by default?

I think we should try to get there eventually, but it's likely to have far reaching consequences and a little risky just before the 9.0 release.

src/backend/distributed/commands/dependencies.c
```diff
@@ -245,6 +248,12 @@ ReplicateAllDependenciesToNode(const char *nodeName, int nodePort)
 		ObjectAddress *dependency = (ObjectAddress *) lfirst(dependencyCell);
 		ddlCommands = list_concat(ddlCommands,
 								  GetDependencyCreateDDLCommands(dependency));
+
+		if (!*clusterHasDistributedFunction &&
```
marcocitus: This API is kind of weird. Maybe just add a separate function that looks whether there is any tuple with `ProcedureRelationId` in the `classid` column in `pg_dist_object`.

src/backend/distributed/commands/function.c
marcocitus: It's possible that the transaction restriction is there only for historical reasons. It might be ok to sync the metadata snapshot as part of a distributed transaction. That said, I'm fine with either approach.

src/backend/distributed/commands/dependencies.c
onderkalaci: I didn't want to scan through the `pg_dist_object` again, but that's OK as well.

src/backend/distributed/utils/node_metadata.c
```diff
@@ -475,6 +476,16 @@ ActivateNode(char *nodeName, int nodePort)
 		EnsureNoModificationsHaveBeenDone();
 		ReplicateAllDependenciesToNode(nodeName, nodePort);
 		ReplicateAllReferenceTablesToNode(nodeName, nodePort);
+
+		/*
+		 * Let the maintanince deamon do the hard work of syncing the metadata. We prefer
+		 * this because otherwise node activation mig
+		 */
+		if (ClusterHasDistributedFunction())
```
marcocitus: It's worth trying to run the regression tests with metadata syncing always on to check for MX bugs.

e.g. I spotted a crash in `ALTER TABLE collections ATTACH PARTITION collections_4 FOR VALUES IN ( 4 );`

marcocitus: We should probably also check the replication model here.

src/backend/distributed/commands/function.c
onderkalaci: I don't want to remove that check now because we'd need to think/test a bit more. For example, we should switch to sequential mode. Or, what happens if `create_distributed_function` is followed by `SELECT` call in the same transaction? For now, we seem fine what we have. 


Now, we're just not optimizing some `SELECT/CALL` pushdowns, so seems safer.

src/backend/distributed/metadata/distobject.c
```diff
@@ -254,6 +255,41 @@ IsObjectDistributed(const ObjectAddress *address)
...
+ */
+bool
+ClusterHasDistributedFunction(void)
+{
+	Relation pgDistObjectRel = NULL;
+	SysScanDesc pgDistObjectScan = NULL;
+	HeapTuple pgDistObjectTup = NULL;
+
+	pgDistObjectRel = heap_open(DistObjectRelationId(), AccessShareLock);
+	pgDistObjectScan = systable_beginscan(pgDistObjectRel, InvalidOid, false, NULL, 0,
```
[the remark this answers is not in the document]
marcocitus: Not a lot, unless you have a ton of schemas.

We do have an index (primary key):
```
postgres=# \d citus.pg_dist_object
                       Table "citus.pg_dist_object"
┌─────────────────────────────┬─────────┬───────────┬──────────┬─────────┐
│           Column            │  Type   │ Collation │ Nullable │ Default │
├─────────────────────────────┼─────────┼───────────┼──────────┼─────────┤
│ classid                     │ oid     │           │ not null │         │
│ objid                       │ oid     │           │ not null │         │
│ objsubid                    │ integer │           │ not null │         │
│ type                        │ text    │           │          │         │
│ object_names                │ text[]  │           │          │         │
│ object_args                 │ text[]  │           │          │         │
│ distribution_argument_index │ integer │           │          │         │
│ colocationid                │ integer │           │          │         │
└─────────────────────────────┴─────────┴───────────┴──────────┴─────────┘
Indexes:
    "pg_dist_object_pkey" PRIMARY KEY, btree (classid, objid, objsubid)
```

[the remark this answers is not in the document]
onderkalaci: Hmm, that's true. It looks like `systable_beginscan ` expects all the index columns to scankeys:
```C
 * If we choose to index, we reset them to 1..n to reference the index
 * columns.  Note this means there must be one scankey qualification per
 * index column! 
```

But, reading the `systable_beginscan` code, we may only pass `classid `, since it is the first column in the index and it should(?) work fine. Though, we probably wouldn't want to rely on that?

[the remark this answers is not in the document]
marcocitus: It can probably, but seems fine to leave it as is.

src/backend/distributed/utils/node_metadata.c
onderkalaci: When I cherry-pick the commit in #3008 and #3009 and run the regression tests with MX enabled, the crashes are gone. The output difference in the failing tests are just expected.

src/test/regress/sql/distributed_functions.sql
```diff
@@ -132,18 +166,29 @@ FROM pg_dist_partition, citus.pg_dist_object as objects
...
+SELECT stop_metadata_sync_to_node(nodename,nodeport) FROM pg_dist_node WHERE isactive AND noderole = 'primary';
 
--- also show that we can use the function
-SELECT * FROM run_command_on_workers('SELECT function_tests.add_mixed_param_names(2,3);') ORDER BY 1,2;
+SET client_min_messages TO error; -- suppress cascading objects dropping
+DROP SCHEMA function_tests CASCADE;
 
--- clear objects
+-- This is hacky, but we should clean-up the resources as below
+
```
marcocitus: Why wasn't the DROP SCHEMA sufficient?

src/test/regress/sql/distributed_functions.sql
```diff
@@ -132,18 +166,29 @@ FROM pg_dist_partition, citus.pg_dist_object as objects
...
+TRUNCATE pg_dist_node;
 DROP SCHEMA function_tests CASCADE;
-SELECT run_command_on_workers($$DROP SCHEMA function_tests CASCADE;$$);
+
+\c - - - :worker_2_port
+SET client_min_messages TO error; -- suppress cascading objects dropping
+UPDATE pg_dist_local_group SET groupid = 0;
+SELECT worker_drop_distributed_table(logicalrelid::text) FROM pg_dist_partition WHERE logicalrelid::text ILIKE '%replicated_table_func_test%';
+TRUNCATE pg_dist_node;
+DROP SCHEMA function_tests CASCADE;
```
marcocitus: We should add an option to stop_metadata_sync_to_node to clear the metadata.

src/test/regress/sql/distributed_functions.sql
marcocitus: Discussed offline. It does not actually remove the metadata, so leaves a lot of stuff for other tests.

src/test/regress/sql/distributed_functions.sql
onderkalaci: Seems like we already have an issue #2326

Issue #2434, opened 2018-10-10
DISTINCT() AS with 2+ columns gives an "anonymous composite type" error message


colton-citus:
Following error occurs when attempting to SELECT DISTINCT() AS on a distributed table with more than one column within the DISTINCT clause: 

ERROR:  input of anonymous composite types is not implemented 

Minimum reproduction SQL script: 

```
create table t1 (col1 int,col2 int);
select create_distributed_table('t1','col1');
insert into t1(col1,col2) VALUES (1,1),(2,2),(1,2);
select distinct (col1,col2) AS foo FROM t1;
drop table t1;
```

Please note that a regular DISTINCT clause works fine here.

marcocitus:
Somehow this query deparses to `SELECT DISTINCT ROW(col1, col2) AS foo FROM t1_102393 t1 WHERE true`, which seems like a bug in ruleutils.

marcocitus:
this also applies to router queries

marcocitus:
It's probably not a bug in ruleutils, but rather the inability of Postgres to parse anonymous composite types when the coordinator reads the results from the workers.

marcocitus:
Minimal test case:
```
SELECT (col1,col2) FROM t1 WHERE col1 = 1;
ERROR:  input of anonymous composite types is not implemented
```

marcocitus:
In Postgres itself, this limitation also manifests itself when the type of the record needs to be resolved.
```
CREATE TABLE foo AS SELECT (x,y) FROM local;
ERROR:  column "row" has pseudo-type record
```

onderkalaci:
Looks like related with #2184 as well.

A nasty workaround to have the output have a type and skip the error: `select distinct ARRAY[col1,col2] AS foo FROM t1;`

serprex:
Fixed by #3191

Issue #2184, opened 2018-05-21
Table returning functions are not supported


furkansahin:
While I was working on #2175, I realized that Citus gives the following error when there is a table returning function;

```
ERROR:  0A000: input of anonymous composite types is not implemented
LOCATION:  record_in, rowtypes.c:103
```

To reproduce the problem, let's create a distributed table;
```
create table test_tab(i int, j int);
select create_distributed_table('test_tab', 'i'); 
insert into test_tab values(1,2),(3,4),(5,6);
```

Now, let's create our table returning function;
```
CREATE OR REPLACE FUNCTION foo(INT) RETURNS TABLE(name text, id INT) AS $$ 
DECLARE 
BEGIN
   RETURN query 
   SELECT
      $1::text,
      $1;
END
;
$$ language plpgsql;
```

Now, we can simply run the following query and see that it fails;
```
select foo(3) from test_tab;
...
...
LOG:  00000: issuing COPY (SELECT public.foo(3) AS foo FROM test_tab_102178 test_tab WHERE true) TO STDOUT
DETAIL:  on server localhost:9701
LOCATION:  LogRemoteCommand, remote_commands.c:348
ERROR:  0A000: input of anonymous composite types is not implemented
LOCATION:  record_in, rowtypes.c:103
Time: 73.038 ms
```

The error we get happens when Citus starts to copy the result of the query to the coordinator. Since we do not know the resulting row type, copy simply fails in `record_in` call as you can see from the backtrace.
```
#0  errstart (elevel=20, filename=0x107c93e5c "rowtypes.c", lineno=103, funcname=0x107c93e67 "record_in", domain=0x0) at elog.c:251
#1  0x00000001079ef766 in record_in (fcinfo=0x7fff587eb5f0) at rowtypes.c:101
#2  0x0000000107a96d43 in InputFunctionCall (flinfo=0x7fa7c9102130, str=0x7fa7c9101900 "(3,3)", typioparam=2249, typmod=-1) at fmgr.c:1640
#3  0x0000000107613f9b in NextCopyFrom (cstate=0x1084bb7f0, econtext=0x7fa7c90cee58, values=0x7fa7c90c5818, nulls=0x7fa7c90c5838 "\001~\020", tupleOid=0x0) at copy.c:3416
[...]

PR #3006, opened 2019-09-20
Support anonymous composite types on the target list in router queries

marcocitus:
DESCRIPTION: Support anonymous composite types on the target list in router queries

To support SELECT function call delegation, we need to be able to parse anonymous composite types. In addition, we have several bugs involving row expressions (#2434, #2184).

This PR adds support for functions returning records and row expressions on the target list of router queries. I looked into supporting multi-shard queries as well, but the master planner is a little trickier to fix.

The approach is to "bless" the tuple descriptor of the function result or row expression, which adds it to the type cache and assigns it a "type mod". The type mod is stored in the vartypmod in the target list of the custom scan, which is eventually translated into the TupleDesc that is used by the executor to parse rows coming back from workers.


src/test/regress/sql/row_types.sql
```diff
@@ -0,0 +1,43 @@
...
+
+CREATE OR REPLACE FUNCTION record_returner(INOUT id int, OUT name text)
+RETURNS record
+AS $$
+BEGIN
+   id := id + 1;
+   name := 'returned';
+END;
+$$ language plpgsql;
+SELECT create_distributed_function('record_returner(int)');
```
serprex: Until we merge specifying distribution argument, I've been manually updating pg_dist_object in multi_mx_call test

marcocitus: There's no distribution argument here, I just used create_distributed_function to be able to call the function from a query.

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -1089,6 +1091,51 @@ FinalizeRouterPlan(PlannedStmt *localPlan, CustomScan *customScan)
...
+					newVar->vartypmod = resultTupleDesc->tdtypmod;
+				}
+			}
+			else if (IsA(targetEntry->expr, RowExpr))
+			{
+				/*
+				 * Handle row expressions, e.g. SELECT (1,2);
+				 */
+				RowExpr *rowExpr = (RowExpr *) targetEntry->expr;
+				TupleDesc resultTupleDesc = ExecTypeFromExprList(rowExpr->args);
```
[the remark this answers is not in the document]
marcocitus: Makes sense, unwrapped `ExecTypeFromExprList` and added support for nested row types, though table functions are handled a bit differently by the planner so those cannot be nested yet.

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -1120,6 +1135,78 @@ FinalizeRouterPlan(PlannedStmt *localPlan, CustomScan *customScan)
...
+
+			TupleDescInitEntry(rowTupleDesc, currentResno, NULL,
+							   rowArgTypeId, rowArgTypeMod, 0);
+			TupleDescInitEntryCollation(rowTupleDesc, currentResno,
+										exprCollation(rowArg));
+
+			currentResno++;
+		}
+
+		BlessTupleDesc(rowTupleDesc);
```
onderkalaci: Note for other readers: I was worried about `blessing` too many records, but it turns out that Postgres is smart enough to create one entry for multiple calls, see `assign_record_type_typmod`

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -1120,6 +1135,78 @@ FinalizeRouterPlan(PlannedStmt *localPlan, CustomScan *customScan)
...
+
+		rowTupleDesc = CreateTemplateTupleDesc(list_length(rowExpr->args), false);
+
+		foreach(argCell, rowExpr->args)
+		{
+			Node *rowArg = (Node *) lfirst(argCell);
+			Oid rowArgTypeId = exprType(rowArg);
+			int rowArgTypeMod = exprTypmod(rowArg);
+
+			if (rowArgTypeId == RECORDOID)
```
onderkalaci: Similarly, we could consider adding `RECORARRAYDOID `

src/test/regress/expected/row_types.out
```diff
@@ -0,0 +1,151 @@
...
+--------------------
+ (1,"(2,returned)")
+ (1,"(2,returned)")
+(2 rows)
+
+-- table functions in row expressions are not supported
+SELECT (x,table_returner(x)) FROM test WHERE x = 1 ORDER BY x, y;
+ERROR:  input of anonymous composite types is not implemented
+-- try prepared statements
+PREPARE rec(int) AS SELECT (x,y) FROM test WHERE x = $1  ORDER BY x, y;
```
onderkalaci: Minor: Another prepared statement test could be use the parameter in the target list `SELECT row(x,$1 * y) FROM test ..`, which works fine btw

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -1120,6 +1135,78 @@ FinalizeRouterPlan(PlannedStmt *localPlan, CustomScan *customScan)
 }
 
 
+/*
+ * BlessRecordExpression ensures we can parse an anonymous composite type on the
```
onderkalaci: This function is doing some allocations within the current memory context. That all seems fine because `assign_record_type_typmod` uses the cache context when it needs to store the tupleDesc, and we're not relying on any memory generated via this function.  

Do you see any issues with the memory contexts?

marcocitus: Should be fine. Fortunately we're only flipping an integer in the plan so we don't need to allocate new memory.

`get_expr_result_type` explicitly returns a copy of the TupleDesc in the current context.

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -1089,6 +1092,18 @@ FinalizeRouterPlan(PlannedStmt *localPlan, CustomScan *customScan)
 
 		/* build target entry pointing to remote scan range table entry */
 		newVar = makeVarFromTargetEntry(customScanRangeTableIndex, targetEntry);
+
+		if (newVar->vartype == RECORDOID)
```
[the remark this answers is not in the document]
marcocitus: Good idea, though we should probably figure out a more general implementation that covers multi-shard queries as well before adding a lot more weight.

PR #3007, opened 2019-09-20
commands/type: combine FilterNameListForDistributedTypes & TypeNameListToObjectAddresses

serprex:
These functions are both only used once, & the latter repeats work the former already does

Noticed while trying to follow the type propagation code to implement [propagation for collations](https://github.com/citusdata/citus/tree/propagate_create_collation)

### DIFF
--- a/src/backend/distributed/commands/type.c
+++ b/src/backend/distributed/commands/type.c
@@ -85,8 +85,8 @@
 bool EnableCreateTypePropagation = true;
 
 /* forward declaration for helper functions*/
-static List * FilterNameListForDistributedTypes(List *objects, bool missing_ok);
-static List * TypeNameListToObjectAddresses(List *objects);
+static List * FilterNameListForDistributedTypes(List *objects, bool missing_ok,
+												List **objectAddresses);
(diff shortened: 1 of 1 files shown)


src/backend/distributed/commands/type.c
```diff
@@ -493,7 +494,6 @@ PlanDropTypeStmt(DropStmt *stmt, const char *queryString)
 	/*
 	 * remove the entries for the distributed objects on dropping
 	 */
-	distributedTypeAddresses = TypeNameListToObjectAddresses(distributedTypes);
 	foreach(addressCell, distributedTypeAddresses)
 	{
 		ObjectAddress *address = (ObjectAddress *) lfirst(addressCell);
```
marcocitus: Maybe we could iterate over `distributedTypes` and use a `TypeNameToObjectAddress` function here and in `FilterNameListForDistributedTypes`? Seems cleaner overall and the double work is easily forgiven in a DROP TYPE statement.

serprex: Potentially, still means translating name-to-address twice when copying this code to a new object type

serprex:
Closing this as there isn't clear support that this is better

PR #3009, opened 2019-09-22
Support serial and smallserial when syncing metadata

marcocitus:
DESCRIPTION: Support serial and smallserial when syncing metadata

The limitation that you cannot have a table with serial or smallserial when syncing metadata may cause unexpected errors as we start syncing metadata more aggressively. This PR takes away the limitation by applying the logic we use for bigserial to serial and smallserial as well.

We normally reserve the first 8 bits of generated values for the group ID when using bigserial, which allows distinct ranges for up to 65535 nodes. For serial and smallserial we reserve the first 4 bits, which allows distinct ranges for up to 15 nodes, and leaves ~268 million unique values per node for serial, and 4096 unique values per node for smallserial. 

In general, users should either use bigserial or insert via the coordinator.

### DIFF
--- a/src/backend/distributed/sql/citus--8.3-1--9.0-1.sql
+++ b/src/backend/distributed/sql/citus--8.3-1--9.0-1.sql
@@ -147,4 +147,12 @@ ALTER TABLE pg_dist_node ADD COLUMN metadatasynced BOOLEAN DEFAULT FALSE;
 COMMENT ON COLUMN pg_dist_node.metadatasynced IS
     'indicates whether the node has the most recent metadata';
 
+CREATE FUNCTION worker_apply_sequence_command(create_sequence_command text,
+                                              sequence_type_id regtype DEFAULT 'bigint'::regtype)
+    RETURNS VOID
+    LANGUAGE C STRICT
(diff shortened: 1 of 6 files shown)


src/backend/distributed/sql/citus--8.3-1--9.0-1.sql
```diff
@@ -147,4 +147,12 @@ ALTER TABLE pg_dist_node ADD COLUMN metadatasynced BOOLEAN DEFAULT FALSE;
 COMMENT ON COLUMN pg_dist_node.metadatasynced IS
     'indicates whether the node has the most recent metadata';
 
+CREATE FUNCTION worker_apply_sequence_command(create_sequence_command text,
```
onderkalaci: So, we'll have two functions on the workers: `worker_apply_sequence_command (text)` and `worker_apply_sequence_command (text, regtype)`. Is that intentional?

marcocitus: It was intentional in the sense that I was thinking of rolling upgrade scenarios where workers are ahead of the coordinator.

PR #3012, opened 2019-09-23
Update all docker images with pg 11.5

### LINKED ISSUE #3004: Some of docker images not using the latest pg minor version
In some of our docker images we do not have the latest pg minor version. For example for failtester we are using pg11.2
https://github.com/citusdata/the-process/blob/master/circleci/images/failtester/Dockerfile-11#L80

I have heard that this could cause a problem, for example if a change in pg11.5 results in an unexpected behavior in the linked image in citus, we cannot detect it currently.

Would it make sense to have a script that updates the images daily possibly? So that if there is a new minor version for postgres, it will be installed and we will be testing with that from now on instead of the old one.

This is true for other images as well, for example for `exttester` we install postgres at the time of creating the image. https://github.com/citusdata/the-process/blob/master/circleci/images/exttester/Dockerfile#L3

SaitTalhaNisanci:
This PR updates our test images to use the latest pg11.

Once this is approved I will push the images to `citus` docker hub account and update the repositories here.

Fixes #3004.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -100,7 +100,7 @@ jobs:
           command: 'install-and-test-ext check-failure'
   test-11_check-multi:
     docker:
-      - image: 'citusdata/exttester-11:latest'
+      - image: 'citus/exttester-11:latest'
     working_directory: /home/circleci/project
     steps:
       - attach_workspace:
@@ -113,7 +113,7 @@ jobs:
 
   test-11_check-non-adaptive-multi:
     docker:
-      - image: 'citusdata/exttester-11:latest'
+      - image: 'citus/exttester-11:latest'
     working_directory: /home/circleci/project
     steps:
       - attach_workspace:
@@ -126,7 +126,7 @@ jobs:
 
   test-11_check-non-adaptive-failure:
     docker:
-      - image: 'citusdata/failtester-11:latest'
+      - image: 'citus/failtester-11:latest'
     working_directory: /home/circleci/project
     steps:
       - attach_workspace: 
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -5,7 +5,7 @@ orbs:
 jobs:
   build:
     docker:
-      - image: 'citusdata/extbuilder:werror'
```
JelteF: is `latest` now `werror`?

SaitTalhaNisanci: Yeah, I heard that `werror` tag wasnt necessary anymore anyways since it was merged

JelteF:
seems good to me if we push all images to the `citus` docker hub

JelteF:
Can you change this PR to also use the new PG10 images? To be sure that the new docker file works fine for those?

mtuncer:
@thanodnl Use of a different docker hub account is new. We did not have admin rights to existing account while @jasonmp85 was on leave. Now he is back, or soon to be back perhaps we can continue with the old account. 

@SaitTalhaNisanci  I have push rights to existing  docker hub account.  I could help with updating images if needed.

thanodnl:
My block is not about the docker hub account, but about the packagecloud account. I have never seen this account and I don't know how people have access to it.

Since it is about package delivery an operational security I need to have a look into the account, why and since when it is there, and what the difference with our normal packagecloud account is.

I guess I know why @SaitTalhaNisanci doesn't have access to the citusdata pacakagecloud account, that has todo with operational security etc. Creating a new account is not an answer here. We either need to get him on our original account or possibly need to onboard citus-bot on there.

I don't have time tonight to dive deep on the account hence me blocking this PR, I will have a chat offline with involved parties and resolve tomorrow.

PR #2994, opened 2019-09-18
Feature: distributed function rename

thanodnl:
DESCRIPTION: Implement ALTER FUNCTION ... RENAME TO ... for distributed functions

TODO
 - [x] Deparse
 - [x] Qualify

PR #3014, opened 2019-09-23
Deparsing and qualifiying {ALTER|DROP} .. {FUNCTION|PROCEDURE} queries

hanefi:
This PR aims to add all the necessary logic to qualify and deparse all possible `{ALTER|DROP} .. {FUNCTION|PROCEDURE}` queries.

As Procedures are introduced in PG11, the code contains many PG version checks. I tried my best to make it easy to clean up once we drop PG10 support.

Here are some caveats:
- I assumed that the parse tree is a valid one. There are some queries that are not allowed, but still are parsed successfully by postgres planner. Such queries will result in errors in execution time. (e.g. `ALTER PROCEDURE p STRICT` -> `STRICT` action is valid for functions but not procedures. Postgres decides to parse them nevertheless.)

### DIFF
--- a/src/backend/distributed/deparser/deparse.c
+++ b/src/backend/distributed/deparser/deparse.c
@@ -166,8 +166,8 @@ DeparseAlterTableStmt(AlterTableStmt *stmt)
 /*
  * DeparseRenameStmt deparses an RenameStmt to its SQL command.
  * Contrary to its name these statements are not covering all ALTER .. RENAME
- * statements. 
- * 
+ * statements.
+ *
  * e.g. ALTER TYPE name RENAME VALUE old TO new is an AlterEnumStmt
  *
  * Currently with limited support. Check support before using, and add support for new
--- a/src/backend/distributed/deparser/deparse.c
+++ b/src/backend/distributed/deparser/deparse.c
@@ -137,9 +137,10 @@ DeparseDropStmt(DropStmt *stmt)
 
 
 /*
- * DeparseAlterTableStmt aims to deparse
- * ALTER TABLE ..
- * statements.
+ * DeparseAlterTableStmt deparses an AlterTableStmt to its SQL command.
+ * Contrary to its name these statements are covering not only ALTER TABLE ...
+ * statements but are used for almost all relation-esque objects in postgres,
+ * including but not limited to, Tables, Types, ...
  *
  * Currently with limited support. Check support before using, and add support for new
  * statements as required.
@@ -163,9 +164,11 @@ DeparseAlterTableStmt(AlterTableStmt *stmt)
 
 
 /*
- * DeparseRenameStmt aims to deparse
- * ALTER .. RENAME TO ..
- * statements.
+ * DeparseRenameStmt deparses an RenameStmt to its SQL command.
+ * Contrary to its name these statements are not covering all ALTER .. RENAME
+ * statements. 
+ * 
+ * e.g. ALTER TYPE name RENAME VALUE old TO new is an AlterEnumStmt
  *
  * Currently with limited support. Check support before using, and add support for new
  * statements as required.
@@ -201,14 +204,6 @@ DeparseRenameStmt(RenameStmt *stmt)
 }
 
 
-/*
- * DeparseRenameAttributeStmt aims to deparse
- * ALTER TYPE .. RENAME ATTRIBUTE .. TO ..
- * statements.
- *
- * Currently with limited support. Check support before using, and add support for new
- * statements as required.
- */
 static const char *
 DeparseRenameAttributeStmt(RenameStmt *stmt)
 {
--- a/src/backend/distributed/deparser/deparse.c
+++ b/src/backend/distributed/deparser/deparse.c
@@ -35,7 +35,11 @@ static const char * DeparseAlterObjectDependsStmt(AlterObjectDependsStmt *stmt);
  *  - ALTER TYPE
  *  - DROP TYPE
  *
+ *  - ALTER FUNCTION
  *  - DROP FUNCTION
+ *
+ *  - ALTER PROCEDURE
+ *  - DROP PROCEDURE
  */
 const char *
 DeparseTreeNode(Node *stmt)
--- a/src/backend/distributed/deparser/deparse.c
+++ b/src/backend/distributed/deparser/deparse.c
@@ -23,7 +23,7 @@ static const char * DeparseRenameStmt(RenameStmt *stmt);
 static const char * DeparseRenameAttributeStmt(RenameStmt *stmt);
 static const char * DeparseAlterObjectSchemaStmt(AlterObjectSchemaStmt *stmt);
 static const char * DeparseAlterOwnerStmt(AlterOwnerStmt *stmt);
-
+static const char * DeparseAlterObjectDependsStmt(AlterObjectDependsStmt *stmt);
 
 /*
  * DeparseTreeNode aims to be the inverse of postgres' ParseTreeNode. Currently with
@@ -85,6 +85,11 @@ DeparseTreeNode(Node *stmt)
 		case T_AlterOwnerStmt:
 		{
 			return DeparseAlterOwnerStmt(castNode(AlterOwnerStmt, stmt));
+		}
+
+		case T_AlterObjectDependsStmt:
+		{
+			return DeparseAlterObjectDependsStmt(castNode(AlterObjectDependsStmt, stmt));
 		}
 
 		default:
@@ -151,6 +156,11 @@ DeparseRenameStmt(RenameStmt *stmt)
 			return DeparseRenameAttributeStmt(stmt);
 		}
 
+		case OBJECT_FUNCTION:
+		{
+			return DeparseRenameFunctionStmt(stmt);
+		}
+
 		default:
 		{
 			ereport(ERROR, (errmsg("unsupported rename statement for deparsing")));
@@ -190,6 +200,11 @@ DeparseAlterObjectSchemaStmt(AlterObjectSchemaStmt *stmt)
 			return DeparseAlterTypeSchemaStmt(stmt);
 		}
 
+		case OBJECT_FUNCTION:
+		{
+			return DeparseAlterFunctionSchemaStmt(stmt);
+		}
+
 		default:
 		{
 			ereport(ERROR, (errmsg("unsupported rename statement for deparsing")));
@@ -208,9 +223,34 @@ DeparseAlterOwnerStmt(AlterOwnerStmt *stmt)
 			return DeparseAlterTypeOwnerStmt(stmt);
 		}
 
+		case OBJECT_FUNCTION:
+		case OBJECT_PROCEDURE:
+		{
+			return DeparseAlterFunctionOwnerStmt(stmt);
+		}
+
 		default:
 		{
 			ereport(ERROR, (errmsg("unsupported alter owner statement for deparsing")));
+		}
+	}
+}
+
+
+static const char *
+DeparseAlterObjectDependsStmt(AlterObjectDependsStmt *stmt)
+{
+	switch (stmt->objectType)
+	{
+		case OBJECT_FUNCTION:
+		case OBJECT_PROCEDURE:
+		{
+			return DeparseAlterFunctionDependsStmt(stmt);
+		}
+
+		default:
+		{
+			ereport(ERROR, (errmsg("unsupported alter depends statement for deparsing")));
 		}
 	}
 }
--- a/src/backend/distributed/deparser/deparse.c
+++ b/src/backend/distributed/deparser/deparse.c
@@ -111,6 +111,9 @@ DeparseDropStmt(DropStmt *stmt)
 		}
 
 		case OBJECT_FUNCTION:
+#if PG_VERSION_NUM >= 110000
+		case OBJECT_PROCEDURE:
+#endif
 		{
 			return DeparseDropFunctionStmt(stmt);
 		}
@@ -157,6 +160,9 @@ DeparseRenameStmt(RenameStmt *stmt)
 		}
 
 		case OBJECT_FUNCTION:
+#if PG_VERSION_NUM >= 110000
+		case OBJECT_PROCEDURE:
+#endif
 		{
 			return DeparseRenameFunctionStmt(stmt);
 		}
@@ -201,6 +207,9 @@ DeparseAlterObjectSchemaStmt(AlterObjectSchemaStmt *stmt)
 		}
 
 		case OBJECT_FUNCTION:
+#if PG_VERSION_NUM >= 110000
+		case OBJECT_PROCEDURE:
+#endif
 		{
 			return DeparseAlterFunctionSchemaStmt(stmt);
 		}
@@ -224,7 +233,9 @@ DeparseAlterOwnerStmt(AlterOwnerStmt *stmt)
 		}
 
 		case OBJECT_FUNCTION:
+#if PG_VERSION_NUM >= 110000
 		case OBJECT_PROCEDURE:
+#endif
 		{
 			return DeparseAlterFunctionOwnerStmt(stmt);
 		}
@@ -243,7 +254,9 @@ DeparseAlterObjectDependsStmt(AlterObjectDependsStmt *stmt)
 	switch (stmt->objectType)
 	{
 		case OBJECT_FUNCTION:
+#if PG_VERSION_NUM >= 110000
 		case OBJECT_PROCEDURE:
+#endif
 		{
 			return DeparseAlterFunctionDependsStmt(stmt);
 		}
--- a/src/backend/distributed/deparser/deparse.c
+++ b/src/backend/distributed/deparser/deparse.c
@@ -120,13 +120,13 @@ DeparseDropStmt(DropStmt *stmt)
 			return DeparseDropTypeStmt(stmt);
 		}
 
-		case OBJECT_FUNCTION:
 #if PG_VERSION_NUM >= 110000
 		case OBJECT_PROCEDURE:
 #endif
-			{
-				return DeparseDropFunctionStmt(stmt);
-			}
+		case OBJECT_FUNCTION:
+		{
+			return DeparseDropFunctionStmt(stmt);
+		}
 
 		default:
 		{
@@ -188,13 +188,13 @@ DeparseRenameStmt(RenameStmt *stmt)
 			return DeparseRenameAttributeStmt(stmt);
 		}
 
-		case OBJECT_FUNCTION:
 #if PG_VERSION_NUM >= 110000
 		case OBJECT_PROCEDURE:
 #endif
-			{
-				return DeparseRenameFunctionStmt(stmt);
-			}
+		case OBJECT_FUNCTION:
+		{
+			return DeparseRenameFunctionStmt(stmt);
+		}
 
 		default:
 		{
@@ -243,13 +243,13 @@ DeparseAlterObjectSchemaStmt(AlterObjectSchemaStmt *stmt)
 			return DeparseAlterTypeSchemaStmt(stmt);
 		}
 
-		case OBJECT_FUNCTION:
 #if PG_VERSION_NUM >= 110000
 		case OBJECT_PROCEDURE:
 #endif
-			{
-				return DeparseAlterFunctionSchemaStmt(stmt);
-			}
+		case OBJECT_FUNCTION:
+		{
+			return DeparseAlterFunctionSchemaStmt(stmt);
+		}
 
 		default:
 		{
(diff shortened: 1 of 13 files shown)


src/backend/distributed/deparser/deparse_function_stmts.c
```diff
@@ -0,0 +1,527 @@
...
+}
+
+
+const char *
+DeparseDropFunctionStmt(DropStmt *stmt)
+{
+	StringInfoData str = { 0 };
+	initStringInfo(&str);
+
+#if (PG_VERSION_NUM < 110000)
```
onderkalaci: Do we need these asserts in `Deparse*()` functions? We're only getting here if these conditions hit and we're already guarding every deparse with a test. 

These are very repetitive and ifdefs make it ugly :)

src/test/regress/expected/multi_deparse_function.out
```diff
@@ -0,0 +1,311 @@
...
+CREATE SCHEMA function_tests;
+SET search_path TO function_tests;
+SET citus.shard_count TO 4;
+SET client_min_messages TO INFO;
+CREATE FUNCTION deparse_test(text)
+	RETURNS text
+	AS 'citus'
+ 	LANGUAGE C STRICT;
+-- Create a simple function
+CREATE FUNCTION add(integer, integer) RETURNS integer
```
onderkalaci: Let's use a lot more complicated function in the tests:

a) Use a function name that needs to be quoted
b) Use a schema name that needs to be quoted (see [here](https://github.com/citusdata/citus/blob/master/src/test/regress/sql/multi_schema_support.sql#L773) for examples)
c) Have an argument with DEFAULT value
d) Have OUT argument
e) Have an INOUT argument
f) Have an VARIADIC argument
g) Use a different type than INT
h) Execute some tests with search path `SET search_path TO public` for sanity checks
i) Sanity check with one function using plpgsql instead of sql
j) Sanity check with RETURNS TABLE -- do we support that?

src/backend/distributed/deparser/deparse_function_stmts.c
```diff
@@ -0,0 +1,527 @@
...
+	Assert(stmt->objectType == OBJECT_FUNCTION);
+#else
+	Assert(stmt->objectType == OBJECT_FUNCTION || stmt->objectType == OBJECT_PROCEDURE);
+#endif
+
+	AppendAlterFunctionSchemaStmt(&str, stmt);
```
onderkalaci: Minor: In general, I think it'd be easier to move `AppendXXXStmt ` right after `DeparseXXXStmt `. Otherwise, the readers need to navigate to see the actual logic.

src/backend/distributed/deparser/deparse_function_stmts.c
hanefi: I will create a new commit to drop all these assertions in `Deparse*()` functions. I guess I will still need to test that all the callers already make sure that these conditions are hit.

hanefi: You were right, and there is no harm in dropping these assertions. Thanks for this catch

src/backend/distributed/deparser/deparse_function_stmts.c
hanefi: I ordered them locally. I hope it will work out this time.

src/backend/distributed/deparser/deparse.c
```diff
@@ -196,9 +236,38 @@ DeparseAlterOwnerStmt(AlterOwnerStmt *stmt)
...
+
 		default:
 		{
 			ereport(ERROR, (errmsg("unsupported alter owner statement for deparsing")));
 		}
 	}
 }
+
+
+static const char *
```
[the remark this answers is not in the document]
hanefi: I added docs in many places. Feel free to point me in places where you think we need more comments.

src/backend/distributed/deparser/deparse.c
```diff
@@ -23,7 +23,7 @@ static const char * DeparseRenameStmt(RenameStmt *stmt);
 static const char * DeparseRenameAttributeStmt(RenameStmt *stmt);
 static const char * DeparseAlterObjectSchemaStmt(AlterObjectSchemaStmt *stmt);
 static const char * DeparseAlterOwnerStmt(AlterOwnerStmt *stmt);
-
+static const char * DeparseAlterObjectDependsStmt(AlterObjectDependsStmt *stmt);
```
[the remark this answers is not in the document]
hanefi: #3020 seems to fix this problem. After rebasing it compiles without warnings

src/test/regress/expected/multi_deparse_function.out
```diff
@@ -0,0 +1,311 @@
...
+SELECT deparse_test($cmd$
+DROP FUNCTION IF EXISTS missing_schema.missing_function(int,float);
+$cmd$);
+                                        deparse_test                                         
+---------------------------------------------------------------------------------------------
+ DROP FUNCTION IF EXISTS missing_schema.missing_function(pg_catalog.int4,pg_catalog.float8);
+(1 row)
+
+-- clear objects
+SET client_min_messages TO FATAL; -- suppress cascading objects dropping
```
onderkalaci: can you add a test with multiple options: `alter function xxx  volatile leakproof  SECURITY DEFINER PARALLEL unsafe;
`

src/backend/distributed/deparser/deparse_function_stmts.c
```diff
@@ -0,0 +1,596 @@
...
+}
+
+
+/*
+ * AppendDefElem appends a string representing the DefElem to a buffer
+ */
+static void
+AppendDefElem(StringInfo buf, DefElem *def)
+{
+	if (strcmp(def->defname, "strict") == 0)
```
onderkalaci: Are we sure that we'd always have lower case-letters in defelem? If not, we could consider using `pg_strncasecmp`. If yes, can you please change some of the tests to have CapiTal lETters in the elems?

src/backend/distributed/deparser/deparse_function_stmts.c
```diff
@@ -0,0 +1,596 @@
...
+		appendStringInfoString(buf, "DROP PROCEDURE ");
+	}
+#endif
+
+	if (stmt->missing_ok)
+	{
+		appendStringInfoString(buf, "IF EXISTS ");
+	}
+
+	AppendFunctionNameList(buf, stmt->objects);
```
onderkalaci: Can you add a test which drops if exists multiple functions?`DROP FUNCTION IF EXISTS add(int,int), add(int);`

src/test/regress/expected/multi_deparse_function.out
onderkalaci: maybe one argument is a composite type: `CREATE TYPE intpair AS (x int, y int);
`

src/test/regress/expected/multi_deparse_function.out
```diff
@@ -0,0 +1,311 @@
...
+    AS 'select $1 + $2;'
+    LANGUAGE SQL
+    IMMUTABLE
+    RETURNS NULL ON NULL INPUT;
+SELECT deparse_test($cmd$
+ALTER FUNCTION  add CALLED ON NULL INPUT
+$cmd$);
+                       deparse_test                        
+-----------------------------------------------------------
+ ALTER FUNCTION function_tests.add() CALLED ON NULL INPUT;
```
onderkalaci: one problem across all the tests is that the deparsed query cannot be executed:
```SQL
ALTER FUNCTION function_tests.add() CALLED ON NULL INPUT;
ERROR:  function function_tests.add() does not exist
Time: 0.612 ms
```
Which should not happen because we're planning to execute this queries on the workers.

So, we should add the parameter types to the output. 

And, more importantly, maybe let `deparse_test ` to connect to the node, and execute the command so that we'd not have any such cases. Otherwise, you'll need to manually ensure that the commands work fine

src/backend/distributed/deparser/deparse_function_stmts.c
hanefi: This is exactly how it is used in Postgres codebase. So I think it is perfetly safe to use them this way.

For reference:
https://github.com/postgres/postgres/blob/f5daf7f3266ff5a92f1bf8d386bd5ac3d7d042d6/src/backend/commands/functioncmds.c#L494

hanefi: I will however add some tests with mixed letters like 
`ALTER FUNCTION "CiTuS.TeeN"."TeeNFuNc.1!?!"() StrIct`
 to be safe

src/test/regress/expected/multi_deparse_function.out
thanodnl: Connecting to the workers and executing the commands is best done in an end to end fashion by hooking the deparser to the utility hook as it done in (#2995, #2994, #2993, #2992, #2991). Doing so in this test seems redundant and a lot of work.

For now I think it is our job as author and reviewers to critically inspect the output of the tests.

src/test/regress/expected/multi_deparse_function.out
```diff
@@ -0,0 +1,311 @@
...
+SELECT deparse_test($cmd$
+ALTER FUNCTION  add SET log_min_messages TO DEFAULT
+$cmd$);
+                                     deparse_test                                     
+--------------------------------------------------------------------------------------
+ ALTER FUNCTION function_tests.add(integer, integer) SET log_min_messages TO DEFAULT;
+(1 row)
+
+SELECT deparse_test($cmd$
+ALTER FUNCTION  add SET log_min_messages FROM CURRENT
```
thanodnl: what does from current entail, is this portable to remote workers that might have set the value to a different value?
From documentation:
```
SET FROM CURRENT saves the value of the parameter that is current when ALTER FUNCTION is executed as the value to be applied when the function is entered.
```

the qualification logic should resolve this and instead change the statement to a `ALTER FUNCTION ... SET ... TO current_value`.

src/test/regress/expected/multi_deparse_function.out
```diff
@@ -0,0 +1,311 @@
+--
+-- Regression tests for deparsing ALTER/DROP FUNCTION Queries
+--
+-- This test implements all the possible queries as of Postgres 11:
+-- 
+-- ALTER FUNCTION name [ ( [ [ argmode ] [ argname ] argtype [, ...] ] ) ]
+--     action [ ... ] [ RESTRICT ]
```
thanodnl: can we add some tests with multiple actions to make sure those are correctly deparsed?

src/test/regress/expected/multi_deparse_function.out
```diff
@@ -0,0 +1,311 @@
...
+--     [ EXTERNAL ] SECURITY INVOKER | [ EXTERNAL ] SECURITY DEFINER
+--     PARALLEL { UNSAFE | RESTRICTED | SAFE }
+--     COST execution_cost
+--     ROWS result_rows
+--     SET configuration_parameter { TO | = } { value | DEFAULT }
+--     SET configuration_parameter FROM CURRENT
+--     RESET configuration_parameter
+--     RESET ALL
+-- 
+-- DROP FUNCTION [ IF EXISTS ] name [ ( [ [ argmode ] [ argname ] argtype [, ...] ] ) ] [, ...]
```
thanodnl: we seem to miss a test for `DROP FUNCTION` without the `IF EXISTS` modifier. Also one that has the `CASCADE` modifier

src/test/regress/expected/multi_deparse_procedure.out
```diff
@@ -0,0 +1,316 @@
+--
```
thanodnl: same comments as the function test file

missing:
 - DROP PROCEDURE ...
 - DROP PROCEDURE ... CASCADE
 - ALTER PROCEDURE ... SET ... FROM CURRENT
 - ALTER PROCEDURE ... multiple actions

src/test/regress/expected/multi_deparse_function.out
onderkalaci: > For now I think it is our job as author and reviewers to critically inspect the output of the tests.

Why not let Postgres do the testing for us :)  I don't think that it is a lot of work, simply call into StandardProcessUtility.  Besides, testing the deparsing logic here would help us to keep scope the smaller on the mentioned PRs. I'm not saying that we should not test end-to-end in those PRs, but I cannot understand the pushback that'd prevent us to be confident with deparsing in this PR.

Anyway, I'll leave it to you guys.

thanodnl: First of all, I now see the initial problem you have pointed out here, it got overshadowed by the suggestion to 'connect' to a node to execute it. @hanefi As onder pointed out the deparsing logic has a flaw which causes it to fail on this statement.
`add` and `add()` have different meanings in postgres. The first would resolve to a function as long as there is only 1 named function `add`. It ignores the input arguments as it can unambiguously resolve `add` to `add(int,int)`. However the deparsed function name has the parenthesis so postgres assumes you would be looking for a function `add` with no arguments which is not what we want. Can you make sure the deparser doesn't output parenthesis for these kinds of deparses? Make sure to also add a test that would target a function without arguments as the parenthesis would need to be added if it has been overloaded. eg.
```sql
magic_function()
magic_function(int,int)

ALTER FUNCTION magic_function ... -- should deparse without parenthesis, even though it will fail during execution
ALTER FUNCTION magic_function() ... -- sholuld deparse with an empty argument list
ALTER FUNCTION magic_function(int, int) ... -- should deparse with its full list of arguments
```

@onderkalaci On the point of executing the commands. Although the idea of executing the commands locally is appealing, more so than connecting to a worker as I understood from your initial message, however it is not as clear cut as it seems.

Executing the commands have side effects on the target function that we would not want to have during the deparsing tests. This could be prevented by executing the command in a transaction that rolls back after it confirmed the change applied correctly.

The definition of correctly is hard. In the bug you pointed out here it would have sufficed to assert no error has been raised by executing the command locally. Given postgres would not be able to find `add()` would cause the exception to be raised. Other test cases explicitly test non-existing schema's and non-existing function names to verify those can be deparsed as well. Executing these statements locally would always fail, which is desired behaviour.

One could argue we could create a savepoint, execute the original command locally, record its output, restore the savepoint, execute the deparsed statement verify the output is the same. Now we are getting somewhere, but still it will not catch subtile issues.

To correctly test the correct execution of the statement we now need to also inspect all the side effects; did the deparsed statement address the exact same object, did it apply the same changes, etc. To get to a hardened version of this that would work throughout the test suite is a though job which I don't think we could afford at the moment.

hanefi: In a commit I pushed yesterday (717b41f53fbd5645cdcfd6fa3f8c407525bf84c5) I resolved the erronous `add()` pieces in the tests.

I will however add some test cases with several functions with the same name but with a different list of parameters to make sure that we address the same function in the workers as well.

I have also created a utility SQL function that I can possibly use to test that the queries are able to run on the workers.

```SQL
CREATE FUNCTION deparse_and_run_on_workers(text)
    RETURNS SETOF record AS $fnc$
    	WITH deparsed_query AS ( SELECT deparse_test($1) qualified_query )
    	SELECT run_command_on_workers(qualified_query) FROM deparsed_query 
    $fnc$ LANGUAGE SQL;
```

Even though I do not check that the resulting changes in the worker are the same as that in the coordinator, I think seeing that the deparsed query can successfully run on the worker should be enough for now.

thanodnl: actually they don't have to be run on the worker, you can execute them locally.
multiple renames, and changes of schema's are a bit tricky if you apply them, either locally or remote. Hence my suggestion to create a savepoint to rollback to. This will execute the query but not persist the side effects, meaning subsequent calls can all work on the same source.

src/backend/distributed/deparser/deparse.c
```diff
@@ -106,6 +136,14 @@ DeparseDropStmt(DropStmt *stmt)
 }
 
 
+/*
+ * DeparseAlterTableStmt aims to deparse
+ * ALTER TABLE ..
```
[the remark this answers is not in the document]
hanefi: This makes sense. I will look at the postgres parser to make sure all my comments are valid in these `Deparse*()` functions.

e.g. here is `AlterTableStmt` definition:  https://github.com/postgres/postgres/blob/a083657896c739909a25190ebd0032c01f6c8109/src/backend/parser/gram.y#L1829

src/backend/distributed/deparser/deparse.c
```diff
@@ -147,6 +201,14 @@ DeparseRenameStmt(RenameStmt *stmt)
 }
 
 
+/*
+ * DeparseRenameAttributeStmt aims to deparse
```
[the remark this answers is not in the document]
hanefi: I will remove this comment, as this is getting out of scope. I should not have touched this function in the first place.

src/backend/distributed/deparser/deparse.c
```diff
@@ -186,6 +264,14 @@ DeparseAlterObjectSchemaStmt(AlterObjectSchemaStmt *stmt)
 }
 
 
+/*
+ * DeparseAlterOwnerStmt aims to deparse
```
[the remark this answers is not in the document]
hanefi: A valid Statement is an `AlterOwnerStmt` if and only if it conforms to the expression `ALTER.*OWNER.*`

When I check out postgres parser rules, and search for `ALTER.*OWNER.*`:
- I see that all of them fall under `AlterOwnerStmt:`
- All the rules are covered under `AlterOwnerStmt:`

https://github.com/postgres/postgres/blob/a083657896c739909a25190ebd0032c01f6c8109/src/backend/parser/gram.y#L9324

src/backend/distributed/deparser/deparse.c
```diff
@@ -196,9 +282,46 @@ DeparseAlterOwnerStmt(AlterOwnerStmt *stmt)
...
 		default:
 		{
 			ereport(ERROR, (errmsg("unsupported alter owner statement for deparsing")));
 		}
 	}
 }
+
+
+/*
+ * DeparseAlterObjectDependsStmt aims to deparse
```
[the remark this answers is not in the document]
hanefi: A valid statement is an `AlterObjectDependsStmt` if it is of the form `ALTER THING name DEPENDS ON EXTENSION name`

Therefore I think my comment is still valid.

src/test/regress/expected/multi_deparse_function.out
hanefi: I completed all of these items except for one.

> h) Execute some tests with search path SET search_path TO public for sanity checks

I do not get what I should be testing here. If I set the search_path TO public before altering a function, I need to specify the original schema name in the query.

src/backend/distributed/deparser/deparse_function_stmts.c
hanefi: side note: I updated the logic to uppercase all the DefElem values in the final deparsed query. So this `strict` keyword will be `STRICT` in the final query string

src/test/regress/expected/multi_deparse_function.out
```diff
@@ -0,0 +1,311 @@
...
+----------------------------------------------------------------------
+ ALTER FUNCTION function_tests.add(integer, integer) PARALLEL unsafe;
+(1 row)
+
+SELECT deparse_test($cmd$
+ALTER FUNCTION add PARALLEL RESTRICTED
+$cmd$);
+                               deparse_test                               
+--------------------------------------------------------------------------
+ ALTER FUNCTION function_tests.add(integer, integer) PARALLEL restricted;
```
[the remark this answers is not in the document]
hanefi: I capitalize all DefElem actions now, so it should be ok

src/test/regress/expected/multi_deparse_function.out
hanefi: I solved this in deparsing logic for now. Need to move it into the associated `Qualify*()` funcs

src/backend/distributed/deparser/deparse.c
```diff
@@ -98,6 +120,14 @@ DeparseDropStmt(DropStmt *stmt)
 			return DeparseDropTypeStmt(stmt);
 		}
 
+		case OBJECT_FUNCTION:
+#if PG_VERSION_NUM >= 110000
```
thanodnl: nit; if you move this block above `case OBJECT_FUNCTION` the indent of the body is a bit more consistent, there are a couple more in this PR.

src/test/regress/expected/multi_deparse_function.out
thanodnl: As discussed; please revert to deparse `SET FROM CURRENT` and open an issue to solve this later.
In #3041 I raise an error when this syntax is applied on a distributed function with some information on how the user can work around this limitation.

src/test/regress/expected/multi_deparse_function.out
hanefi: I added a test for this here: https://github.com/citusdata/citus/blob/47adafe69adf4ad503e5f34f77cc5aec049115d7/src/test/regress/sql/multi_deparse_function.sql#L219

src/test/regress/expected/multi_deparse_function.out
hanefi: I reverted my changes and I will deparse the `SET FROM CURRENT` and I will open an issue and link it to here

src/test/regress/expected/multi_deparse_function.out
thanodnl: Executing them on the workers has been implemented

src/test/regress/expected/multi_deparse_function.out
thanodnl: will punt on `h` for now.

PR #3016, opened 2019-09-23
Swap pg_dist_node groupid and nodeid sequences

### LINKED ISSUE #1081: pg_dist_node uses wrong sequence names for the associated columns
This is a bug but nothing critical. I'm opening the issue to discuss solutions and whether we want to fix it before 6.1. or not.

With #798, we introduced `pg_dist_node` as a new metadata table. The definition of it:

```SQL
CREATE TABLE citus.pg_dist_node(
        nodeid int NOT NULL DEFAULT nextval('pg_dist_groupid_seq') PRIMARY KEY,
        groupid int NOT NULL DEFAULT nextval('pg_dist_node_nodeid_seq'),
        nodename text NOT NULL,
        nodeport int NOT NULL DEFAULT 5432,
        noderack text NOT NULL DEFAULT 'default',
        UNIQUE (nodename, nodeport)
);
```

As seen above, we assign `group_id` sequence to `nodeid` column. Similarly, 'nodeid_seq' to `groupid` column.
The output as expected:
```SQL
                            Table "pg_catalog.pg_dist_node"
   Column    |  Type   |                           Modifiers                           
-------------+---------+---------------------------------------------------------------
 nodeid      | integer | not null default nextval('pg_dist_groupid_seq'::regclass)
 groupid     | integer | not null default nextval('pg_dist_node_nodeid_seq'::regclass)
 nodename    | text    | not null
 nodeport    | integer | not null default 5432
 noderack    | text    | not null default 'default'::text
 hasmetadata | boolean | not null default false
Indexes:
    "pg_dist_node_pkey" PRIMARY KEY, btree (nodeid)
    "pg_dist_node_nodename_nodeport_key" UNIQUE CONSTRAINT, btree (nodename, nodeport)
Triggers:
    
```

marcocitus:
DESCRIPTION: Fix sequence names in pg_dist_node

The groupid and nodeid sequences in pg_dist_node are currently swapped. That does not mean that we use the groupid sequence to generate a node ID, because functions like master_add_node draw directly from the sequence instead of using the default expression. However, if you do an `INSERT INTO pg_dist_node` and don't specify the node ID or group ID, you would use the wrong sequence.

This PR swaps the defaults on the columns to clear up confusion.

Fixes #1081

### DIFF
--- a/src/backend/distributed/sql/citus--8.3-1--9.0-1.sql
+++ b/src/backend/distributed/sql/citus--8.3-1--9.0-1.sql
@@ -1,6 +1,9 @@
 /* citus--8.3-1--9.0-1 */
 
-/* bump version to 9.0-1 */
+-- We swapped the groupid and nodeid sequences when creating pg_dist_node
+ALTER TABLE pg_dist_node ALTER COLUMN groupid SET DEFAULT nextval ('pg_dist_groupid_seq');
+ALTER TABLE pg_dist_node ALTER COLUMN nodeid SET DEFAULT nextval('pg_dist_node_nodeid_seq');
+
 CREATE SCHEMA IF NOT EXISTS citus_internal;
 
 -- move citus internal functions to citus_internal to make space in the citus schema for
--- a/src/test/regress/expected/multi_cluster_management.out
+++ b/src/test/regress/expected/multi_cluster_management.out
@@ -608,7 +608,7 @@ CONTEXT:  PL/pgSQL function citus_internal.pg_dist_node_trigger_func() line 18 a
 INSERT INTO pg_dist_node (nodename, nodeport, groupid, noderole, nodecluster)
   VALUES ('localhost', 5000, 1000, 'primary', 'olap');
 ERROR:  new row for relation "pg_dist_node" violates check constraint "primaries_are_only_allowed_in_the_default_cluster"
-DETAIL:  Failing row contains (19, 1000, localhost, 5000, default, f, t, primary, olap, f).
+DETAIL:  Failing row contains (24, 1000, localhost, 5000, default, f, t, primary, olap, f).
 UPDATE pg_dist_node SET nodecluster = 'olap'
   WHERE nodeport = :worker_1_port;
 ERROR:  new row for relation "pg_dist_node" violates check constraint "primaries_are_only_allowed_in_the_default_cluster"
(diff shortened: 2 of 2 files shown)


src/backend/distributed/sql/citus--8.3-1--9.0-1.sql
```diff
@@ -1,6 +1,9 @@
 /* citus--8.3-1--9.0-1 */
 
-/* bump version to 9.0-1 */
+-- We swapped the groupid and nodeid sequences when creating pg_dist_node
+ALTER TABLE pg_dist_node ALTER COLUMN groupid SET DEFAULT nextval ('pg_dist_groupid_seq');
+ALTER TABLE pg_dist_node ALTER COLUMN nodeid SET DEFAULT nextval('pg_dist_node_nodeid_seq');
```
pykello: Do you know if this can cause problems for upgrades? If so, maybe also `ALTER SEQUENCE ... RENAME ...` or something so `groupid` and `nodeid` assignemnt continue gracefully?

marcocitus: `master_add_node` already uses the right sequences (e.g. `pg_dist_node_nodeid_seq` for `nodeid`) so if we rename them it will start using a different sequence for the nodeid. The ALTER TABLE is there to fix direct inserts and for not looking weird.

onderkalaci:
>  However, if you do an INSERT INTO pg_dist_node and don't specify the node ID or group ID, you would use the wrong sequence.

Thanks, I actually haven't realized that. So, it turns out that this is a pretty minor issue. The regression test output changes confirms this as well.

src/test/regress/expected/multi_cluster_management.out
```diff
@@ -608,7 +608,7 @@ CONTEXT:  PL/pgSQL function citus_internal.pg_dist_node_trigger_func() line 18 a
 INSERT INTO pg_dist_node (nodename, nodeport, groupid, noderole, nodecluster)
   VALUES ('localhost', 5000, 1000, 'primary', 'olap');
 ERROR:  new row for relation "pg_dist_node" violates check constraint "primaries_are_only_allowed_in_the_default_cluster"
-DETAIL:  Failing row contains (19, 1000, localhost, 5000, default, f, t, primary, olap, f).
+DETAIL:  Failing row contains (24, 1000, localhost, 5000, default, f, t, primary, olap, f).
```
onderkalaci: Note for future readers: The other outputs that have changed incremented by `2`, and this one by `5`, this is expected. 

The reason is that the other changes  done via `master_add_node ` or other UDFs, whereas this it done via a plain `INSERT` to pg_dist_node. For the difference, read the PR description. 

(Finally, see the `multi_cluster_management.sql` file and see that there are 2 proceeding `INSERT INTO pg_dist_node` calls, which is the source for `+2`)

PR #3020, opened 2019-09-23
Fix pg12

serprex:
Include ifdefs for pg12 API changes, update local_shard_executiuon test to avoid CTE inlining

### DIFF
--- a/src/backend/distributed/commands/function.c
+++ b/src/backend/distributed/commands/function.c
@@ -19,6 +19,9 @@
 #include "miscadmin.h"
 #include "funcapi.h"
 
+#if PG_VERSION_NUM >= 120000
+#include "access/genam.h"
+#endif
 #include "access/htup_details.h"
 #include "access/xact.h"
 #include "catalog/namespace.h"
--- a/src/backend/distributed/planner/distributed_planner.c
+++ b/src/backend/distributed/planner/distributed_planner.c
@@ -1176,7 +1176,11 @@ BlessRecordExpression(Expr *expr)
 		ListCell *argCell = NULL;
 		int currentResno = 1;
 
+#if PG_VERSION_NUM >= 120000
+		rowTupleDesc = CreateTemplateTupleDesc(list_length(rowExpr->args));
+#else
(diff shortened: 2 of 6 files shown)


onderkalaci:
> update local_shard_executiuon test to avoid CTE inlining

@serprex  Why do we need this? The query you've changed has a comment:

-- multi-shard CTE is followed by a query which could be executed locally, but
-- since the query started with a parallel query, it doesn't use local execution

In any case, I'm curious the reason to change this test

serprex:
In pg12 the query is executed locally because the CTE is inlined, using a CTE twice prevents inlining

onderkalaci:
> In pg12 the query is executed locally because the CTE is inlined (appears to us as a subexpression), using a CTE twice prevents inlining

Oh, so to make the test be compatible with the comment above? I'm just being extra careful not to overlook any integration bugs across too  many new components  in Citus 9.0 :) 

I can see that Postgres inlines the CTE and pushes down the filters to the CTE and that makes it a router query. But, I think to preserve the test as-is, we should probably do something like where Postgres can inline, but cannot pushdown the filters so we preserve the test we want:
```SQL
+WITH all_data AS (SELECT age FROM distributed_table)
 SELECT 
        count(*) 
 FROM 
        distributed_table, all_data 
 WHERE 
-       distributed_table.key = all_data.key AND distributed_table.key = 1
-       -- the following is to avoid CTE inlining
-       AND EXISTS (SELECT * FROM all_data);
+       distributed_table.key = all_data.age AND distributed_table.key = 1;

```

I'll look a bit on this today, and send the the review to you. Thanks for running things on pg 12  and catching earlier!

onderkalaci:
> In pg12 the query is executed locally because the CTE is inlined (appears to us as a subexpression), using a CTE twice prevents inlining

Thinking a bit on this, your change also seems a valuable test for pg 12. My suggestion above is to make sure that even if Postgres does the CTE inlining, we'd be able to prevent local execution due to the new subquery (old  CTE). I might end-up keeping this change and add one more as the above

PR #3021, opened 2019-09-23
Distribute select function

pykello:
DESCRIPTION: Delegate `SELECT function();` to a worker node.

When a function is marked as colocated with a distributed table,
we try delegating queries of kind "SELECT func(...)" to workers.

We currently only support this simple form, and don't delegate
forms like "SELECT f1(...), f2(...)", "SELECT f1(...) FROM ...",
or function calls inside transactions.

As a side effect, we also fix the transactional semantics of DO blocks.
Previously we didn't consider a DO block a multi-statement transaction.
Now we do.

### DIFF
--- a/src/backend/distributed/planner/distributed_planner.c
+++ b/src/backend/distributed/planner/distributed_planner.c
@@ -18,6 +18,7 @@
 #include "catalog/pg_type.h"
 #include "distributed/citus_nodefuncs.h"
 #include "distributed/citus_nodes.h"
+#include "distributed/function_call_delegation.h"
 #include "distributed/insert_select_planner.h"
 #include "distributed/intermediate_results.h"
 #include "distributed/metadata_cache.h"
@@ -191,6 +192,14 @@ distributed_planner(Query *parse, int cursorOptions, ParamListInfo boundParams)
 			setPartitionedTablesInherited = true;
 			AdjustPartitioningForDistributedPlanning(rangeTableList,
 													 setPartitionedTablesInherited);
+		}
+		else
+		{
+			DistributedPlan *delegatePlan = TryToDelegateFunctionCall(parse);
+			if (delegatePlan != NULL)
+			{
+				result = FinalizePlan(result, delegatePlan);
+			}
 		}
 	}
 	PG_CATCH();
--- a/src/backend/distributed/planner/function_call_delegation.c
+++ b/src/backend/distributed/planner/function_call_delegation.c
@@ -1,0 +1,287 @@
+/*-------------------------------------------------------------------------
+ *
+ * function_call_delegation.c
+ *    Planning logic for delegating a function call to a worker when the
+ *    function was distributed with a distribution argument and the worker
+ *    has metadata.
+ *
+ * Copyright (c), Citus Data, Inc.
+ *
+ *-------------------------------------------------------------------------
+ */
+
+#include "postgres.h"
+
+#include "catalog/pg_proc.h"
+#include "catalog/pg_type.h"
+#include "commands/defrem.h"
+#include "distributed/citus_custom_scan.h"
+#include "distributed/citus_ruleutils.h"
+#include "distributed/colocation_utils.h"
+#include "distributed/commands.h"
+#include "distributed/commands/multi_copy.h"
+#include "distributed/connection_management.h"
+#include "distributed/function_call_delegation.h"
+#include "distributed/master_metadata_utility.h"
+#include "distributed/master_protocol.h"
+#include "distributed/metadata_cache.h"
+#include "distributed/multi_executor.h"
+#include "distributed/multi_physical_planner.h"
+#include "distributed/remote_commands.h"
+#include "distributed/shard_pruning.h"
+#include "distributed/version_compat.h"
+#include "distributed/worker_manager.h"
+#include "nodes/makefuncs.h"
+#include "nodes/nodeFuncs.h"
+#include "nodes/parsenodes.h"
+#include "nodes/primnodes.h"
+#include "optimizer/clauses.h"
+#include "parser/parse_coerce.h"
+#if PG_VERSION_NUM >= 120000
+#include "parser/parsetree.h"
+#endif
+#include "miscadmin.h"
+#include "tcop/dest.h"
+#include "utils/lsyscache.h"
+#include "utils/syscache.h"
+
+static bool contain_param_walker(Node *node, void *context);
+
+/*
+ * contain_param_walker scans node for Param nodes.
+ * returns whether any such nodes found.
+ */
+static bool
+contain_param_walker(Node *node, void *context)
+{
+	return IsA(node, Param);
+}
+
+
+/*
+ * TryToDelegateFunctionCall calls a function on the worker if possible.
+ * We only support delegating the SELECT func(...) form for distributed
+ * functions colocated by distributed tables, and not more complicated
+ * forms involving multiple function calls, FROM clauses, WHERE clauses,
+ * ... Those complex forms are handled in the coordinator.
+ */
+DistributedPlan *
+TryToDelegateFunctionCall(Query *query)
+{
+	FromExpr *joinTree = NULL;
+	List *targetList = NIL;
+	TargetEntry *targetEntry = NULL;
+	FuncExpr *funcExpr = NULL;
+	DistObjectCacheEntry *procedure = NULL;
+	Oid colocatedRelationId = InvalidOid;
+	Const *partitionValue = NULL;
+	Datum partitionValueDatum = 0;
+	ShardInterval *shardInterval = NULL;
+	List *placementList = NIL;
+	DistTableCacheEntry *distTable = NULL;
+	Var *partitionColumn = NULL;
+	ShardPlacement *placement = NULL;
+	WorkerNode *workerNode = NULL;
+	StringInfo queryString = NULL;
+	Task *task = NULL;
+	Job *job = NULL;
+	DistributedPlan *distributedPlan = NULL;
+	int32 groupId = 0;
+
+	if (!CitusHasBeenLoaded() || !CheckCitusVersion(DEBUG4))
+	{
+		/* Citus is not ready to determine whether function is distributed */
+		return NULL;
+	}
+
+	groupId = GetLocalGroupId();
+	if (groupId != 0 || groupId == GROUP_ID_UPGRADING)
+	{
+		/* do not delegate from workers, or while upgrading */
+		return NULL;
+	}
+
+	if (query == NULL)
+	{
+		/* no query (mostly here to be defensive) */
+		return NULL;
+	}
+
+	joinTree = query->jointree;
+	if (joinTree == NULL)
+	{
+		/* no join tree (mostly here to be defensive) */
+		return NULL;
+	}
+
+	if (joinTree->quals != NULL)
+	{
+		/* query has a WHERE section */
+		return NULL;
+	}
+
+	if (joinTree->fromlist != NIL)
+	{
+		/* query has a FROM section */
+#if PG_VERSION_NUM >= 120000
+
+		/* in pg12 empty FROMs are represented with an RTE_RESULT */
+		if (list_length(joinTree->fromlist) == 1)
+		{
+			RangeTblRef *reference = linitial(joinTree->fromlist);
+			RangeTblEntry *rtentry = rt_fetch(reference->rtindex, query->rtable);
+			if (rtentry->rtekind != RTE_RESULT)
+			{
+				return NULL;
+			}
+		}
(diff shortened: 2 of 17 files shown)


src/test/regress/sql/multi_mx_function_call_delegation.sql
```diff
@@ -1,69 +1,147 @@
...
-CREATE FUNCTION mx_call_add(int, int) RETURNS int
-    AS 'select $1 + $2;' LANGUAGE SQL IMMUTABLE;
+--
+-- Utility UDFs
+--
 
-CREATE FUNCTION mx_call_dist_object(procname text, tablerelid regclass, argument_index int) RETURNS void LANGUAGE plpgsql AS $$
+-- 1. Marks the given function as colocated with the given table.
+-- 2. Marks the argument index with which we route the function.
+CREATE FUNCTION colocate_func_with_table(procname text, tablerelid regclass, argument_index int)
```
pykello: Maybe move this as `PROCEDURE` to `multi_test_helpers.sql`, since it is used in more than one test files. Similar to https://github.com/citusdata/citus/pull/3032/files

src/test/regress/sql/multi_mx_function_call_delegation.sql
```diff
@@ -1,69 +1,147 @@
...
-select mx_call_dist_object('mx_call_func', 'mx_call_dist_table'::regclass, -1);
+select colocate_func_with_table('mx_call_func', 'mx_call_dist_table_1'::regclass, -1);
 select mx_call_func(2, 0);
-select mx_call_dist_object('mx_call_func', 'mx_call_dist_table'::regclass, 2);
+select colocate_func_with_table('mx_call_func', 'mx_call_dist_table_1'::regclass, 2);
 select mx_call_func(2, 0);
-select mx_call_dist_object('mx_call_func', 'mx_call_dist_table'::regclass, 1);
 
--- test non Const distribution parameter
-select mx_call_func(2, mx_call_add(3, 4));
```
pykello: Also add these, similar to https://github.com/citusdata/citus/pull/3032/files, but also make sure the behaviour is as we expect.

pykello:
I would also add the following tests:
1. A table returning function:

```sql
CREATE FUNCTION dup(int) RETURNS TABLE(f1 int, f2 text)
    AS $$ SELECT $1, CAST($1 AS text) || ' is text' $$
    LANGUAGE SQL;

SELECT dup(3);
```
2. A set returning function:
```sql
CREATE FUNCTION squares(int) RETURNS SETOF RECORD
    AS $$ SELECT i, i * i FROM generate_series(1, $1) i $$
    LANGUAGE SQL;

SELECT squares(4);
```
3. void returning function
4. null returning function

src/test/regress/expected/multi_mx_function_call_delegation.out
```diff
@@ -322,26 +283,214 @@ PL/pgSQL function mx_call_func(integer,integer) line 8 at assignment
...
+PL/pgSQL function mx_call_func(integer,integer) line 8 at assignment
+ mx_call_func 
+--------------
+           29
+(1 row)
+
+-- non-const parameter can be pushed down
+select mx_call_func((select x + 1 from mx_call_add(3, 4) x), 2);
+DEBUG:  pushing down the function call
+ERROR:  there is no parameter $0
```
serprex: This is wrong

src/test/regress/expected/multi_mx_function_call_delegation.out
```diff
@@ -322,26 +283,214 @@ PL/pgSQL function mx_call_func(integer,integer) line 8 at assignment
...
+ mx_call_func | mx_call_func 
+--------------+--------------
+           29 |           27
 (1 row)
 
+do $$ begin select mx_call_func(2, 0); END; $$;
+DEBUG:  pushing down the function call
+CONTEXT:  SQL statement "select mx_call_func(2, 0)"
+PL/pgSQL function inline_code_block line 1 at SQL statement
+ERROR:  unexpected plan node type: 34
```
serprex: This is wrong

src/test/regress/expected/multi_mx_function_call_delegation.out
```diff
@@ -322,26 +283,214 @@ PL/pgSQL function mx_call_func(integer,integer) line 8 at assignment
...
+DEBUG:  Plan 33 query after replacing subqueries and CTEs: SELECT (3 OPERATOR(pg_catalog.+) (SELECT intermediate_result.sum FROM read_intermediate_result('33_1'::text, 'binary'::citus_copy_format) intermediate_result(sum bigint)))
+CONTEXT:  SQL statement "SELECT y + (select sum(t1.val + t2.val) from multi_mx_function_call_delegation.mx_call_dist_table_1 t1 join multi_mx_function_call_delegation.mx_call_dist_table_2 t2 on t1.id = t2.id)"
+PL/pgSQL function mx_call_func(integer,integer) line 8 at assignment
+ y  
+----
+ 29
+(1 row)
+
+select mx_call_func(2, 0) from mx_call_dist_table_1;
+ERROR:  cached metadata for shard 102033 is inconsistent
```
serprex: Is this wrong?

pykello: Still looking into this, but this also happens earlier in multi_mx_call too: https://github.com/citusdata/citus/pull/3021/files#diff-0dfffc23c4864f2642ff50f1cb57e3a7R445. It happens in one of the workers. It didn't happen before. Still looking into that.

pykello: I think this is due to `stop_metadata_sync_to_node()`/`start_metadata_sync_to_node()` calls, which recreate metadata but doesn't invalidate the caches. Still looking.

src/test/regress/expected/multi_mx_function_call_delegation.out
```diff
@@ -0,0 +1,530 @@
...
+select mx_call_func_custom_types('S', 'A');
+DEBUG:  pushing down the function call
+ mx_call_func_custom_types 
+---------------------------
+ (S,S)
+(1 row)
+
+select squares(4);
+DEBUG:  pushing down the function call
+ERROR:  input of anonymous composite types is not implemented
```
serprex: A future PR should look into blessing this through

src/test/regress/expected/multi_mx_function_call_delegation.out
```diff
@@ -0,0 +1,530 @@
...
+DEBUG:  switching to sequential query execution mode
+DETAIL:  A distributed function is created. To make sure subsequent commands see the type correctly we need to make sure to use only one connection for all future commands
+ create_distributed_function 
+-----------------------------
+ 
+(1 row)
+
+-- subquery parameters cannot be pushed down
+select mx_call_func((select x + 1 from mx_call_add(3, 4) x), 2);
+DEBUG:  arguments in a distributed function must not contain subqueries
```
serprex: A future PR may be able to get around this with `ResolveExternalParams`

src/backend/distributed/planner/function_call_delegation.c
```diff
@@ -0,0 +1,288 @@
...
+	if (!CitusHasBeenLoaded() || !CheckCitusVersion(DEBUG4))
+	{
+		/* Citus is not ready to determine whether function is distributed */
+		return NULL;
+	}
+
+	groupId = GetLocalGroupId();
+	if (groupId != 0 || groupId == GROUP_ID_UPGRADING)
+	{
+		/* do not delegate from workers, or while upgrading */
```
onderkalaci: We don't have to do anything right now, but why we do not delegate from workers? That should help in MX case. As I said, it is OK to not do it now, just wanted to learn if are there any specific concerns you have?

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -192,6 +193,14 @@ distributed_planner(Query *parse, int cursorOptions, ParamListInfo boundParams)
 			AdjustPartitioningForDistributedPlanning(rangeTableList,
 													 setPartitionedTablesInherited);
 		}
+		else
```
onderkalaci: I wasn't expecting to see that our logic is in `needsDistributedPlanning  == false`. Instead, I thought we'd do it in `RouterPlanner`, or even `FastPathRouterPlanner`. 

I'm OK with merging like this, but we should probably discuss a follow-up PR to move this to somewhere else.

src/backend/distributed/planner/function_call_delegation.c
```diff
@@ -0,0 +1,288 @@
...
+	}
+
+	if (contain_volatile_functions((Node *) funcExpr->args))
+	{
+		ereport(DEBUG1, (errmsg("arguments in a distributed function must "
+								"be constant expressions")));
+		return NULL;
+	}
+
+	if (expression_tree_walker((Node *) funcExpr->args, contain_param_walker, NULL))
```
onderkalaci: Do we ever need to check for `PARAM_EXEC` as well? Can you double check?

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -3143,18 +3143,19 @@ GetLocalGroupId(void)
...
 
 		groupId = DatumGetInt32(groupIdDatum);
+
+		/* set the local cache variable */
+		LocalGroupId = groupId;
 	}
 	else
 	{
-		elog(ERROR, "could not find any entries in pg_dist_local_group");
+		/* upgrade is happening */
```
onderkalaci: When does that happen? With a concurrent metadata sync?

src/backend/distributed/planner/function_call_delegation.c
```diff
@@ -0,0 +1,288 @@
...
+
+/*
+ * TryToDelegateFunctionCall calls a function on the worker if possible.
+ * We only support delegating the SELECT func(...) form for distributed
+ * functions colocated by distributed tables, and not more complicated
+ * forms involving multiple function calls, FROM clauses, WHERE clauses,
+ * ... Those complex forms are handled in the coordinator.
+ */
+DistributedPlan *
+TryToDelegateFunctionCall(Query *query)
```
onderkalaci: Minor: We might consider splitting this function into two `ShouldDelegateFunctionCall() `, and `DelegateFunctionCall()` just for making this function smaller.

marcocitus:
> pykello requested a review from marcocitus about 11 hours ago

I won't be near a laptop until Monday. Any specific concerns with merging?

onderkalaci:
> I won't be near a laptop until Monday. Any specific concerns with merging?

I'll send an update via e-mail, but in terms of compatibility with benchmarks, this works without issues.

src/backend/distributed/planner/function_call_delegation.c
marcocitus: Would probably cause a lot of duplicate work though (e.g. looking at metadata to get shard placements).

src/backend/distributed/planner/function_call_delegation.c
marcocitus: Delegation loops. Benefit is also unclear because connections are a scarce resource in mx.

src/backend/distributed/utils/metadata_cache.c
pykello: When upgrading postgres, it seems this table is empty before `citus_finish_pg_upgrade()` finishes exeuction (https://github.com/citusdata/citus/blob/4f60e3a149fb83204278f50df9315828cffa8a28/src/backend/distributed/sql/udfs/citus_finish_pg_upgrade/9.0-1.sql#L1)

src/backend/distributed/planner/function_call_delegation.c
pykello: Do you mean we should just not push down for `PARAM_EXEC` and push-down other kinds of params? (paramkind is `PARAM_EXEC` in our subquery test).

I also added a prepared statement test. It seems that by the time the code arrives here parameters have been replaced.

So it seems that we can just reject any param type for now as we are doing now.

src/backend/distributed/utils/metadata_cache.c
pykello: also updated the comment.

src/backend/distributed/planner/function_call_delegation.c
```diff
@@ -0,0 +1,288 @@
...
+	{
+		/* multiple target list items */
+		return NULL;
+	}
+
+	targetEntry = (TargetEntry *) linitial(targetList);
+	if (!IsA(targetEntry->expr, FuncExpr))
+	{
+		/* target list item is not a function call */
+		ereport(DEBUG4, (errmsg("query is not a function call")));
```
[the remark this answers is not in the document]
pykello: done

PR #3024, opened 2019-09-24
CALL propagation: set search_path

serprex:
Adaptive executor also updated to not only store first resultset to tuplestore
This works in our case because SET doesn't produce any tuples

### DIFF
--- a/src/backend/distributed/commands/call.c
+++ b/src/backend/distributed/commands/call.c
@@ -79,7 +79,6 @@ CallFuncExprRemotely(CallStmt *callStmt, DistObjectCacheEntry *procedure,
 	Var *partitionColumn = NULL;
 	ShardPlacement *placement = NULL;
 	WorkerNode *workerNode = NULL;
-	StringInfo callCommand = NULL;
 
 	if (IsMultiStatementTransaction())
 	{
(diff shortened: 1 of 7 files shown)


marcocitus:
We tend to have pgbouncers sit in between coordinator and worker, which might end up sending the SET over a different connection than the CALL.

This search_path issue sucks a bit. I'm wondering whether we should introduce something like `SELECT worker_execute_with_search_path('CALL ...', 'myschema, public')`

I don't think this is a 9.0 blocker. Users can always mess around with `ALTER ROLE .. SET search_path` if they really need to.

serprex:
Perhaps having a reset to `search_path` at end of list, similar to how we have `DISABLE_DDL_PROPAGATION` / `ENABLE_DDL_PROPAGATION` wrapping elsewhere

marcocitus:
The DDL propagation wrapping is always done in a transaction block. Technically we could do CALL in a transaction block, though I don't know how a commit within the CALL affects SET LOCAL. On the other hand, SET is undone in case of rollback so wrapping the whole thing in BEGIN; SET;..;RESET;END; seems ok.

serprex:
This is not the correct solution, a future PR will know better

PR #3026, opened 2019-09-24
Improve some tests around local execution and CTE inlining on pg 12

onderkalaci:
As we've discussed on #3020 a minor change on the tests


onderkalaci:
> Surprising that * & age produce such different results

The idea is that if you do the `inner join` between `distributed_table.key = all_data.key `, and when CTE inlining happens, postgres pushes down `distributed_table.key = 1` filter to `all_data` as well (implicitly adds `all_data.key = 1`). In that case, router planner kicks in.

But if do the join `distributed_table.key = all_data.age`, then Postgres cannot push down filter to `all_data`, and so the router planner doesn't kick in.

PR #3028, opened 2019-09-24
MX Isolation Tests for Drop, Alter, Index and Select For Update

halilozanakgul:
Created isolation tests for DROP TABLE, ALTER TABLE, CREATE INDEX and SELECT FOR UPDATE on MX.
For more info on MX isolation tests checkout [here](https://github.com/citusdata/citus/issues/2972)

### DIFF
--- a/src/test/regress/expected/isolation_drop_alter_index_select_for_update_on_mx.out
+++ b/src/test/regress/expected/isolation_drop_alter_index_select_for_update_on_mx.out
@@ -1,0 +1,218 @@
+Parsed test spec with 3 sessions
+
+starting permutation: s1-start-session-level-connection s1-begin-on-worker s1-insert s2-start-session-level-connection s2-begin-on-worker s2-alter s1-commit-worker s2-commit-worker s1-stop-connection s2-stop-connection s3-select-count
+step s1-start-session-level-connection: 
+        SELECT start_session_level_connection_to_node('localhost', 57637);
+
+start_session_level_connection_to_node
+
+               
+step s1-begin-on-worker: 
+        SELECT run_commands_on_session_level_connection_to_node('BEGIN');
+
+run_commands_on_session_level_connection_to_node
+
+               
+step s1-insert: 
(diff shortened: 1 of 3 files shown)


src/test/regress/expected/isolation_drop_alter_index_select_for_update_on_mx.out
```diff
@@ -0,0 +1,218 @@
...
+
+               
+step s2-begin-on-worker: 
+        SELECT run_commands_on_session_level_connection_to_node('BEGIN');
+
+run_commands_on_session_level_connection_to_node
+
+               
+step s2-select-for-update: 
+	SELECT run_commands_on_session_level_connection_to_node('SELECT * FROM dist_table WHERE id = 5 FOR UPDATE');
```
onderkalaci: On Postgres, can you concurrently do `CREATE INDEX on TABLE ...` and `SELECT * FROM table FOR UPDATE` when there is data on the table and SELECT accesses the rows?

I was expecting to see `waiting`  here, so can you try on Postgres first?
And, can you also do the same test on Postgres with `UPDATE`  instead of `SELECT  FOR  UPDATE`? I'm pretty sure that UPDATE will be blocked, but let's see `SELECT FOR UPDATE`

halilozanakgul: The same happens without Citus. CREATE INDEX doesn't block SELECT FOR UPDATE but blocks UPDATE.
CREATE INDEX acquires a SHARE lock. UPDATE acquires a ROW EXCLUSIVE lock, which conflicts with SHARE but SELECT FOR UPDATE acquires a ROW SHARE lock, which doesn't conflict with SHARE.

onderkalaci: thanks, good to know. I always thought that 'UPDATE' == 'SELECT FOR UPDATE'.

PR #3030, opened 2019-09-24
Use synchronous replication to make follower tests reliable

JelteF:
`check-follower-cluster` tests are currently flaky: https://circleci.com/gh/citusdata/citus/37681

The reason seems to be that it doesn't use synchronous replication with
`synchronous_commit=remote_apply`. This causes commits to not necessarily be
visible on the follower. Which in turn causes errors about non existing
tables/databases.

This PR changes the test perl script such that it sets the right synchronous
replication options.

More info on these options can be found here: https://www.2ndquadrant.com/en/blog/evolution-fault-tolerance-postgresql-synchronous-commit/

### DIFF
--- a/src/test/regress/pg_regress_multi.pl
+++ b/src/test/regress/pg_regress_multi.pl
@@ -504,15 +504,6 @@ if ($followercluster)
 system(catfile("$bindir", "initdb"), ("--nosync", "-U", $user, "--encoding", "UTF8", catfile($TMP_CHECKDIR, $MASTERDIR, "data"))) == 0
     or die "Could not create $MASTERDIR data directory";
 
-if ($followercluster)
-{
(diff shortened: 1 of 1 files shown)


src/test/regress/pg_regress_multi.pl
```diff
@@ -654,10 +646,16 @@ END
 # Signal that servers should be shutdown
 $serversAreShutdown = "FALSE";
 
+# enable synchronous replication if needed
+if ($followercluster)
+{
+    $synchronousReplication = "-c synchronous_standby_names='FIRST 1 (*)' -c synchronous_commit=remote_apply";
```
marcocitus: `synchronous_standby_names='*'` is a lot more common, but I guess this works.

PR #3032, opened 2019-09-25
Hadi's take on SELECT func() tests.

pykello:
I did this in parallel with Philip's latest commit, I will close this without merging.

### DIFF
--- a/src/test/regress/expected/multi_mx_function_call_delegation.out
+++ b/src/test/regress/expected/multi_mx_function_call_delegation.out
@@ -1,148 +1,342 @@
--- Test passing off function call to mx workers
--- Create worker-local tables to test procedure calls were routed
-CREATE SCHEMA multi_mx_function_call_delegation;
-SET search_path TO multi_mx_function_call_delegation;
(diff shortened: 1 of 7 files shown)


src/test/regress/expected/multi_mx_function_call_delegation.out
```diff
@@ -1,148 +1,342 @@
...
+ create_distributed_function 
+-----------------------------
  
 (1 row)
 
--- test non Const distribution parameter
-select mx_call_func(2, mx_call_add(3, 4));
+-- non-const distribution parameters cannot be pushed down
+SELECT multi_mx_select_func.mx_dist_func(2, mx_call_add(3, 4));
 DEBUG:  pushing down the function call
```
pykello: @serprex is this expected?

serprex: No, think this might be due to changing from 0-based indexing to 1-based indexing

marcocitus: In the select case the planner evaluates constant (immutable) expressions.

PR #3041, opened 2019-09-26
Hookup function/procedure deparsing to our utility hook

thanodnl:
DESCRIPTION: Propagate ALTER FUNCTION statements for distributed functions

Replaces #2995 #2994 #2993 #2992 #2991 
Blocked by #3014 

Using the implemented deparser for function statements to propagate changes to both functions and procedures that are previously distributed.

TODO:
 - [x] Implement `ErrorIfUnsupportedAlterFunctionStmt` to error for `SET ... TO CURRENT`.


src/backend/distributed/commands/function.c
```diff
@@ -615,3 +632,492 @@ TriggerSyncMetadataToPrimaryNodes(void)
...
+	EnsureCoordinator();
+
+	ErrorIfUnsupportedAlterFunctionStmt(stmt);
+
+	/* reconstruct alter statement in a portable fashion */
+	QualifyTreeNode((Node *) stmt);
+	sql = DeparseTreeNode((Node *) stmt);
+
+	/*
+	 * all functions that are distributed will need their alter statements propagated
```
onderkalaci: Super minor: Maybe move `EnsureSequentialModeForFunctionDDL ` above `/* reconstruct alter statement in a portable fashion */` block, the flow would look nicer. (In general, this applies to other functions as well, but feel free to ignore.)

src/backend/distributed/commands/function.c
```diff
@@ -615,3 +632,492 @@ TriggerSyncMetadataToPrimaryNodes(void)
...
+ * workers.
+ *
+ * The DropStmt could have multiple objects to drop, the list of objects will be filtered
+ * to only keep the distributed functions for deletion on the workers. Non-distributed
+ * functions will still be dropped locally but not on the workers.
+ */
+List *
+PlanDropFunctionStmt(DropStmt *stmt, const char *queryString)
+{
+	List *oldObjectWithArgsList = stmt->objects;
```
onderkalaci: minor: oldObjectWithArgsList -> probably better to use `droppedObjectWithArgsList`

src/backend/distributed/deparser/qualify_function_stmt.c
```diff
@@ -171,6 +171,7 @@ QualifyFunctionSchemaName(ObjectWithArgs *func, ObjectType type)
 		procform = (Form_pg_proc) GETSTRUCT(proctup);
 		schemaName = get_namespace_name(procform->pronamespace);
 		functionName = NameStr(procform->proname);
+		functionName = pstrdup(functionName);
```
onderkalaci: The name of `makeString ` function that we use below led me think that it'd allocate. But, it turns out that it doesn't, so good catch here.

I've checked other the place that we use `proctup `, but that seems to be fine because we use `quote_qualified_identifier`, which does allocate memory. Can you double check?

src/backend/distributed/commands/function.c
```diff
@@ -615,3 +632,492 @@ TriggerSyncMetadataToPrimaryNodes(void)
...
+	}
+
+	/*
+	 * Swap the list of objects before deparsing and restore the old list after. This
+	 * ensures we only have distributed functions in the deparsed drop statement.
+	 */
+	oldObjectWithArgsList = stmt->objects;
+	stmt->objects = distributedObjectWithArgsList;
+	dropStmtSql = DeparseTreeNode((Node *) stmt);
+	stmt->objects = oldObjectWithArgsList;
```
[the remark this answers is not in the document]
thanodnl: Guess you refer to https://github.com/citusdata/citus/pull/2991#discussion_r326899326

I thought you were pointing out that the backup of the pointer was taken at the top of the function and restored here. I have now added a assignment of the old value right above this block.

Your comment here suggests this is not what you referred to, can you elaborate on what you want to see changed?

src/backend/distributed/commands/function.c
```diff
@@ -615,3 +632,492 @@ TriggerSyncMetadataToPrimaryNodes(void)
...
+	QualifyTreeNode((Node *) stmt);
+	sql = DeparseTreeNode((Node *) stmt);
+
+	EnsureSequentialModeForFunctionDDL();
+
+	commands = list_make3(DISABLE_DDL_PROPAGATION,
+						  (void *) sql,
+						  ENABLE_DDL_PROPAGATION);
+
+	return NodeDDLTaskList(ALL_WORKERS, commands);
```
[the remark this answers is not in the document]
thanodnl: I do agree that it is some amount of repetition but I think abstracting 2 logical instructions (make list, transform to ddl task list) into a function makes it less readable.

For most of the Alter/Drop/Create statements throughout types and functions I found a lot of repetition (eg. the bodies of the functions are practically the same). This caught me once the GetAddress, Qualify and Deparsing logic took form. I think it makes sense to DRY it up in subsequent release. Will keep it here for now.

src/backend/distributed/commands/function.c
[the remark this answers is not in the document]
marcocitus: I meant that instead of modifying the original statement and having to change it back we could instead modify a throwaway copy (with a name that identifies its purpose).

thanodnl:
> I don't like the strategy of introducing huge amounts of infrastructure in one PR and then getting blocked in the PR that uses it, because a) the quality of review and thus code quality deteriorates significantly when the PR is over 1000 lines, and b) it prevents appropriate prioritisation. DROP and CREATE OR REPLACE are far more important than ALTER, while the latter is far more time-consuming. We should be focused on merging the former.

Cannot agree more, would have loved it differently, but this was the easiest way to make quick progress now. Lets discuss on Monday.

src/backend/distributed/deparser/qualify_function_stmt.c
thanodnl: I will have a look into the use of `proctup`. The use-after-free was a problem here during the hooking up, hence the quick fix. Others aren't a problem or didn't came up just yet. I think the safe thing is to _just_ `pstrdup` them all if the ptr is used after the tuple release.

onderkalaci:
>I don't like the strategy of introducing huge amounts of infrastructure in one PR

I think otherwise, we've found issues in the infrastructure PR because it just focused on one thing. If we had that logic here, we'd likely not be able find those.

> the quality of review and thus code quality deteriorates significantly when the PR is over 1000 lines, 

But if we have the infrastructure here, wouldn't this PR become huge and makes things worse? 

I'm always in favor of splitting PRs as much as possible. I think we can question the timelines here. If we had some more reasonable timelines, we'd then first focus on the deparsing and later this. I agree that this time we've not done a great job in pipelining/parallelizing the tasks, but that doesn't mean that merging the infrastructures first is a bad idea. I can easily list lots of projects where this worked pretty well.

src/backend/distributed/commands/function.c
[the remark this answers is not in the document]
thanodnl: check, will do so!

src/backend/distributed/commands/function.c
```diff
@@ -615,3 +632,472 @@ TriggerSyncMetadataToPrimaryNodes(void)
...
+		DefElem *action = castNode(DefElem, lfirst(actionCell));
+		if (strcmp(action->defname, "set") == 0)
+		{
+			VariableSetStmt *setStmt = castNode(VariableSetStmt, action->arg);
+			if (setStmt->kind == VAR_SET_CURRENT)
+			{
+				/* check if the set action is a SET ... FROM CURRENT */
+				ereport(ERROR, (errmsg("unsupported ALTER FUNCTION ... SET ... FROM "
+									   "CURRENT for a distributed function"),
+								errdetail("SET FROM CURRENT is not supported for "
```
onderkalaci: I think this fits slightly better as `errhint` instead of `errdetail `

src/test/regress/expected/distributed_functions.out
```diff
@@ -106,6 +110,170 @@ SELECT * FROM run_command_on_workers('SELECT function_tests.add(2,3);') ORDER BY
...
+$$);
+                 run_command_on_workers                  
+---------------------------------------------------------
+ (localhost,57637,t,"(functionuser,function_tests,add)")
+ (localhost,57638,t,"(functionuser,function_tests,add)")
+(2 rows)
+
+-- change the schema of the function and verify the old schema doesn't exist anymore while
+-- the new schema has the function.
+ALTER FUNCTION add(int,int) SET SCHEMA function_tests2;
```
onderkalaci: Should we not create `function_tests2 ` above and let the dependency logic create it? This is also fine, but that looks fancier :)

src/backend/distributed/commands/function.c
```diff
@@ -26,21 +26,25 @@
...
 #include "distributed/metadata/distobject.h"
 #include "distributed/metadata/pg_dist_object.h"
+#include "distributed/metadata_sync.h"
 #include "distributed/multi_executor.h"
 #include "distributed/relation_access_tracking.h"
 #include "distributed/worker_transaction.h"
+#include "parser/parse_type.h"
 #include "storage/lmgr.h"
 #include "utils/builtins.h"
-#include "utils/fmgrprotos.h"
```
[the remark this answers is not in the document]
onderkalaci: But not added back?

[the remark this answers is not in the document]
thanodnl: it is actually back, but ordered on alphabet so it appears 2 lines below

src/test/regress/expected/distributed_functions.out
thanodnl: The schema is actually created on the workers via the dependency logic at this point in time. We do create the schema on the coordinator at the top of the file. Having it here might make it more clear it is propagated to the workers here, however it is hard to read which schema's a test uses if they are not all created at the top, hence the `CREATE SCHEMA` statement all the way at the top

src/test/regress/expected/distributed_procedure_0.out
```diff
@@ -0,0 +1,206 @@
...
+SET citus.shard_count TO 4;
+-- Create and distribute a simple function
+CREATE OR REPLACE PROCEDURE raise_info(text)
+    LANGUAGE PLPGSQL AS $proc$
+BEGIN
+    RAISE INFO 'information message %', $1;
+END;
+$proc$;
+ERROR:  syntax error at or near "PROCEDURE"
+LINE 1: CREATE OR REPLACE PROCEDURE raise_info(text)
```
onderkalaci: shall we do something like here: https://github.com/citusdata/citus/blob/master/src/test/regress/expected/pg12_1.out

PR #3043, opened 2019-09-26
Propagate CREATE OR REPLACE FUNCTION to workers for distributed functions

thanodnl:
DESCRIPTION: Propagate CREATE OR REPLACE FUNCTION

Distributed functions could be replaced, which should be propagated to the workers to keep the function in sync between all nodes.

Due to the complexity of deparsing the `CreateFunctionStmt` we actually produce the plan during the processing phase of our utilityhook. Since the changes have already been made in the catalog tables we can reuse `pg_get_functiondef` to get us the generated `CREATE OR REPLACE` sql.

### DIFF
--- a/src/backend/distributed/commands/function.c
+++ b/src/backend/distributed/commands/function.c
@@ -66,6 +66,7 @@ static void UpdateFunctionDistributionInfo(const ObjectAddress *distAddress,
 										   int *colocationId);
 static void EnsureSequentialModeForFunctionDDL(void);
 static void TriggerSyncMetadataToPrimaryNodes(void);
+static bool ShouldPropagateCreateFunction(CreateFunctionStmt *stmt);
 static bool ShouldPropagateAlterFunction(const ObjectAddress *address);
 static ObjectAddress * FunctionToObjectAddress(ObjectType objectType,
 											   ObjectWithArgs *objectWithArgs,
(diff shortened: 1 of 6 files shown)


src/backend/distributed/commands/function.c
```diff
@@ -665,6 +717,102 @@ ShouldPropagateAlterFunction(const ObjectAddress *address)
...
+
+	if (!ShouldPropagateCreateFunction(stmt))
+	{
+		return NIL;
+	}
+
+	address = GetObjectAddressFromParseTree((Node *) stmt, false);
+	EnsureDependenciesExistsOnAllNodes(address);
+
+	sql = GetFunctionDDLCommand(address->objectId);
```
onderkalaci: I like this approach so much that, I couldn't remember the reason that we have all the logic for function deparsing. Can't we always use this approach for all function related things? We don't have to do that, but I think if we can do it, we'd be able to remove a lot of code. I wouldn't mind sending `CREATE OR REPLACE FUNCTION` to the workers on `ALTER FUNCTION` on the coordinator. Are there any specific reasons/concerns for not doing it?

src/backend/distributed/commands/function.c
```diff
@@ -631,6 +632,57 @@ TriggerSyncMetadataToPrimaryNodes(void)
...
+
+	if (!EnableDependencyCreation)
+	{
+		/*
+		 * we are configured to disable object propagation, should not propagate anything
+		 */
+		return false;
+	}
+
+	if (!stmt->replace)
```
onderkalaci: I guess we shouldn't need this check because `if (!stmt->replace)`, it cannot be a distributed function -- so the below check would return false.

On the other hand, this makes it easier to follow the code -- more explicit code. So, keeping it seems like a good option as well

src/backend/distributed/commands/function.c
marcocitus: I think the main uses of ALTER FUNCTION (ownership, rename, schema change) probably cannot be done via CREATE OR REPLACE. It could make sense for SET, but not sure if CREATE takes more aggressive locks.

thanodnl: Marco's response for ownership, rename and schema changes are the exact reasons.

Looking at the actions base syntax (which includes) `SET` I think we could replace that one with this version. However due to time constraints now I think we should keep our current version and refactor it at a later stage when time is not a limiting factor.

src/backend/distributed/commands/function.c
thanodnl: One could do a create without a replace on a distributed function, this would error during local execution though. I think having this explicitly here is more clear towards the intention of the function.

PR #3049, opened 2019-09-27
Disallow distributed functions for functions depending on an extension

thanodnl:
DESCRIPTION: Disallow distributed functions for functions depending on an extension

Functions depending on an extension cannot (yet) be distributed by citus. If we would allow this it would cause issues with our dependency following mechanism as we stop following objects depending on an extension.

By not allowing functions to be distributed when they depend on an extension as well as not allowing to make distributed functions depend on an extension we won't break the ability to add new nodes. Allowing functions depending on extensions to be distributed at the moment could cause problems in that area.

### DIFF
--- a/src/backend/distributed/commands/function.c
+++ b/src/backend/distributed/commands/function.c
@@ -1072,6 +1072,17 @@ PlanDropFunctionStmt(DropStmt *stmt, const char *queryString)
 }
 
 
+/*
+ * PlanAlterFunctionDependsStmt is called during the planning phase of an
+ * ALTER FUNCION ... DEPENDS ON EXTENSION ... statement. Since functions depending on
+ * extensions are assumed to be Owned by an extension we assume the extension to keep the
+ * function in sync.
+ *
+ * If we would allow users to create a dependency between a distributed function and an
+ * extension our pruning logic for which objects to distribute as dependencies of other
+ * objects will change significantly which could cause issues adding new workers. Hence we
+ * don't allow this dependency to be created.
+ */
 List *
 PlanAlterFunctionDependsStmt(AlterObjectDependsStmt *stmt, const char *queryString)
 {
@@ -1119,6 +1130,11 @@ PlanAlterFunctionDependsStmt(AlterObjectDependsStmt *stmt, const char *queryStri
 }
 
 
+/*
+ * AlterFunctionDependsStmtObjectAddress resolves the ObjectAddress of the function that
+ * is the subject of an ALTER FUNCTION ... DEPENS ON EXTENSION ... statement. If
+ * missing_ok is set to false the lookup will raise an error.
+ */
 const ObjectAddress *
 AlterFunctionDependsStmtObjectAddress(AlterObjectDependsStmt *stmt, bool missing_ok)
 {
--- a/src/backend/distributed/commands/function.c
+++ b/src/backend/distributed/commands/function.c
@@ -1111,11 +1111,10 @@ PlanAlterFunctionDependsStmt(AlterObjectDependsStmt *stmt, const char *queryStri
 	 * workers
 	 */
 
-	functionName = get_func_name(address->objectId);
-
+	functionName = getObjectIdentity(address);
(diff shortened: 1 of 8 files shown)


src/backend/distributed/commands/function.c
```diff
@@ -146,6 +146,17 @@ create_distributed_function(PG_FUNCTION_ARGS)
 
 	ObjectAddressSet(functionAddress, ProcedureRelationId, funcOid);
 
+	if (IsObjectAddressOwnedByExtension(&functionAddress))
+	{
+		char *functionName = get_func_name(funcOid);
+		ereport(ERROR, (errmsg("unable to create a distributed function from functions "
```
onderkalaci: can we find the extension name as well? Printing that would be helpful in the error messages.

Also, super minor, to keep this function smaller and avoid verbose error lines, it'd have a slight preference to move this check to `ErrorIfFunctionDependsOnExtension()`. But, feel free to ignore this if you don't feel similar.

src/backend/distributed/commands/function.c
```diff
@@ -1061,6 +1072,63 @@ PlanDropFunctionStmt(DropStmt *stmt, const char *queryString)
 }
 
 
+List *
+PlanAlterFunctionDependsStmt(AlterObjectDependsStmt *stmt, const char *queryString)
```
onderkalaci: Missing function comment, few others as well.

We also don't use `queryString` so maybe not pass at all.

src/backend/distributed/commands/utility_hook.c
```diff
@@ -852,6 +860,33 @@ PlanAlterOwnerStmt(AlterOwnerStmt *stmt, const char *queryString)
...
 
 
+/*
+ * PlanAlterObjectDependsStmt gets called during the planning phase for
+ * ALTER ... DEPENDS ON EXTENSION ... statements. Based on the object type we call out to
+ * a specialized implementation. If no implementation is available we do nothing, eg. we
+ * allow the local node to execute.
+ */
+static List *
+PlanAlterObjectDependsStmt(AlterObjectDependsStmt *stmt, const char *queryString)
```
onderkalaci: I find it little confusing that a function that is either going to error out or do nothing returns a `List`.

In general, we make this functions in a way that `ErrorIfAlterDistributedFunctionDependsOnExtension()` to make things a lot more explicit.

I think your point would be that this is consistent with other similar functions, which is fair. Still, I've a slight preference for changing the function name/return value such that it becomes easier to follow the code.

What's your take on this?

src/backend/distributed/commands/function.c
thanodnl: will add the comments.

will respond on the `queryString` part in the subsequent comment as they are very much aligned.

src/backend/distributed/commands/utility_hook.c
thanodnl: I agree that only looking at this function might make it a bit strange that the function signature is `List * Plan...(... stmt, const char *queryString)` if its only function is to return an error if it is a distributed function.

But if you look at it in the broader scope of how the utility hook is structured it is a prelude to a structured way of responding to statements from citus' perspective.

```C
ddlJobs = PlanStmtType(stmt, queryString);
ExecuteOriginalUtilityHook(...)
ProcessStmtType(stmt, queryString);
ExecuteDDLJobsOnWorkers(ddlJobs);
```

Currently the logic is not as clear as above due to the history of how the utility hook came to be. However there is a clear interface forming now where based on the type of statement we need to retrieve two (potentially `NULL`) function pointers, eg. `PlanStmt` and `ProcessStmt`. Hence the current setup where it both returns a `List *` and takes a `queryString`.

Even if we don't have the abstraction described here it becomes easier to read and understand the body of our utility hook if the functions we call look a lot alike. Once you understand the flow of one you understand them all and it becomes natural to navigate. If we would tailer the names of the functions and the input arguments specific to each situation my fear would be that the utility hook is not understandable to the same degree.

PR #3050, opened 2019-09-30
Update CircleCI codecov orb

JelteF:
Hopefully this fixes the long runtime for coverage uploads that we sometimes have (> 1 minute).


SaitTalhaNisanci:
why do we expect this to fix the problem?

JelteF:
I don't expect it to help, but if we still have the problem afterwards. We can try to email them about it.

PR #3051, opened 2019-09-30
Add editorconfig settings for UNIX line endings, utf8, newline at end of file and trimming trailing whitespace

JelteF:
Just some sane settings


.editorconfig
```diff
@@ -7,6 +7,10 @@ root = true
 indent_style = tab
 indent_size = 4
 tab_width = 4
+end_of_line = lf
+insert_final_newline = true
+charset = utf-8
+trim_trailing_whitespace = true
```
thanodnl: Trimming trailing whitespace is non-desirable for `*.out` files. IIRC there are some tests that output whitespace at the end, and the test will fail without a diff showing where.

JelteF: Fixed

PR #3053, opened 2019-09-30
Fix pg12 empty from clauses

hanefi:
PG12 introduced some changes in the way Query objects store the `jointree` of the query. This PR aims to take this into account and make necessary checks during processing queries without any `FROM` or `WHERE` clauses.

## Changes in tests:

Outputs of these tests are slightly different than PG11:
- multi_mx_call
- multi_mx_function_call_delegation
- multi_unsupported_worker_operations
	
I created new regression test outputs for these files and rotated all necessary test outputs. For example here are the expected regression outputs for `multi_mx_call`:

| test output         	| PG version 	|
|---------------------	|------------	|
| multi_mx_call.out   	| PG12       	|
| multi_mx_call_0.out 	| PG11       	|
| multi_mx_call_1.out 	| PG11       	|

Note: all the changes introduced in this PR is inside a `#if PG_VERSION_NUM >= 120000` block, so there is no CI job testing these commits. I ran `make check` to test citus master branch running on `PG 12rc1`

### DIFF
--- a/src/backend/distributed/planner/function_call_delegation.c
+++ b/src/backend/distributed/planner/function_call_delegation.c
@@ -156,24 +156,45 @@ TryToDelegateFunctionCall(Query *query, bool *hasExternParam)
 
 	if (joinTree->fromlist != NIL)
 	{
-		/* query has a FROM section */
 #if PG_VERSION_NUM >= 120000
 
-		/* in pg12 empty FROMs are represented with an RTE_RESULT */
+		/*
+		 * In pg12's planning phase empty FROMs are represented with an RTE_RESULT.
+		 * When we arrive here, standard_planner has already been called which calls
+		 * replace_empty_jointree() which replaces empty fromlist with a list of
+		 * single RTE_RESULT RangleTableRef node.
+		 */
 		if (list_length(joinTree->fromlist) == 1)
 		{
 			RangeTblRef *reference = linitial(joinTree->fromlist);
-			RangeTblEntry *rtentry = rt_fetch(reference->rtindex, query->rtable);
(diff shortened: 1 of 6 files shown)


src/test/regress/expected/multi_mx_call_1.out
```diff
@@ -0,0 +1,337 @@
+-- Test passing off CALL to mx workers
+create schema multi_mx_call;
```
onderkalaci: can you do the trick similar to here: https://github.com/citusdata/citus/pull/3041/files#diff-3b203d78d59c5fe3d63034e6461ce142R2 

where the output files for not supported versions is minimal?

hanefi: Procedures are supported in PG11 and PG12, but they produce different `CONTEXT` messages if a plpgsql function/procedure raises an exception. That's why I think that it makes sense to have 3 different outputs.

```diff
--CONTEXT:  while executing command on localhost:57638
--PL/pgSQL function multi_mx_call.mx_call_proc_raise(integer) line 4 at RAISE
++CONTEXT:  PL/pgSQL function multi_mx_call.mx_call_proc_raise(integer) line 4 at RAISE
++while executing command on localhost:57638
```

onderkalaci: Does the following fix it `\set VERBOSITY terse`?

onderkalaci: ping on this

serprex: Terse verbosity indeed resolved the difference

PR #3057, opened 2019-09-30
Force custom plan for function call delegation

### LINKED ISSUE #3056: Function calls with prepared statements are not delegated after 5th execution
See below to repro the issue, after the 5th execution, we cannot see `DEBUG:  pushing down the function call`

```SQL
DROP TABLE IF EXISTS test;
CREATE TABLE test (x int);

DROP FUNCTION IF EXISTS delegated_function(int);
CREATE OR REPLACE FUNCTION public.delegated_function(a int)
RETURNS int
LANGUAGE plpgsql
AS $function$
DECLARE
BEGIN
    INSERT INTO test VALUES (a);
    INSERT INTO test VALUES (a+1);
    RETURN a+2;
END;
$function$;
SELECT create_distributed_function('delegated_function(int)', 'a');

PREPARE p1 (int) AS SELECT delegated_function($1);
SET client_min_messages TO DEBUG1;
postgres=# EXECUTE p1(1);
DEBUG:  pushing down the function call
┌────────────────────┐
│ delegated_function │
├────────────────────┤
│                  3 │
└────────────────────┘
(1 row)

Time: 16.100 ms
postgres=# EXECUTE p1(1);
DEBUG:  pushing down the function call
┌────────────────────┐
│ delegated_function │
├────────────────────┤
│                  3 │
└────────────────────┘
(1 row)

Time: 2.072 ms
postgres=# EXECUTE p1(1);
DEBUG:  pushing down the function call
┌────────────────────┐
│ delegated_function │
├────────────────────┤
│                  3 │
└────────────────────┘
(1 row)

Time: 1.964 ms
postgres=# EXECUTE p1(1);
DEBUG:  pushing down the function call
┌────────────────────┐
│ delegated_function │
├────────────────────┤
│                  3 │
└────────────────────┘
(1 row)

Time: 2.653 ms
postgres=# EXECUTE p1(1);
DEBUG:  pushing down the function call
┌────────────────────┐
│ delegated_function │
├────────────────────┤
│                  3 │
└────────────────────┘
(1 row)

Time: 3.295 ms
postgres=# EXECUTE p1(1);
DEBUG:  arguments in a distributed function must not contain subqueries
┌────────────────────┐
│ delegated_function │
├────────────────────┤
│                  3 │
└────────────────────┘
(1 row)

Time: 3.509 ms
postgres=# EXECUTE p1(1);
┌────────────────────┐
│ delegated_function │
├────────────────────┤
│                  3 │
└────────────────────┘
(1 row)

Time: 2.762 ms
postgres=# EXECUTE p1(1);
┌────────────────────┐
│ delegated_function │
├────────────────────┤
│                  3 │
└────────────────────┘
(1 row)

Time: 4.822 ms
postgres=# 
```

serprex:
Fixes #3056

### DIFF
--- a/src/backend/distributed/planner/distributed_planner.c
+++ b/src/backend/distributed/planner/distributed_planner.c
@@ -195,10 +195,20 @@ distributed_planner(Query *parse, int cursorOptions, ParamListInfo boundParams)
 		}
 		else
 		{
-			DistributedPlan *delegatePlan = TryToDelegateFunctionCall(parse);
+			bool hasExternParam = false;
+			DistributedPlan *delegatePlan = TryToDelegateFunctionCall(parse,
+																	  &hasExternParam);
 			if (delegatePlan != NULL)
 			{
 				result = FinalizePlan(result, delegatePlan);
+			}
+			else if (hasExternParam)
+			{
+				/*
+				 * As in CreateDistributedPlannedStmt, try dissuade planner when planning
+				 * potentially failed due to unresolved prepared statement parameters.
+				 */
+				result->planTree->total_cost = FLT_MAX / 100000000;
 			}
 		}
 	}
--- a/src/backend/distributed/planner/function_call_delegation.c
+++ b/src/backend/distributed/planner/function_call_delegation.c
@@ -48,16 +48,43 @@
 #include "utils/lsyscache.h"
 #include "utils/syscache.h"
 
+struct ParamWalkerContext
+{
+	bool hasParam;
+	ParamKind paramKind;
+};
+
 static bool contain_param_walker(Node *node, void *context);
 
 /*
  * contain_param_walker scans node for Param nodes.
- * returns whether any such nodes found.
+ * Ignore the return value, instead check context afterwards.
+ *
+ * context is a struct ParamWalkerContext*.
+ * hasParam is set to true if we find a Param node.
+ * paramKind is set to the paramkind of the Param node if any found.
+ * paramKind is set to PARAM_EXEC if both PARAM_EXEC & PARAM_EXTERN are found.
+ *
+ * By time we walk, Param nodes are either PARAM_EXTERN or PARAM_EXEC.
  */
 static bool
 contain_param_walker(Node *node, void *context)
 {
-	return IsA(node, Param);
+	if (IsA(node, Param))
+	{
+		Param *paramNode = (Param *) node;
+		struct ParamWalkerContext *pwcontext =
+			(struct ParamWalkerContext *) context;
+
+		pwcontext->hasParam = true;
+		pwcontext->paramKind = paramNode->paramkind;
+		if (paramNode->paramkind == PARAM_EXEC)
+		{
+			return true;
+		}
+	}
+
+	return false;
 }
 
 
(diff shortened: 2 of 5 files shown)


src/test/regress/expected/multi_mx_function_call_delegation.out
```diff
@@ -533,6 +533,42 @@ DEBUG:  pushing down the function call
...
+
+EXECUTE call_plan(2, 0);
+DEBUG:  pushing down the function call
+ mx_call_func 
+--------------
+           28
+(1 row)
+
+EXECUTE call_plan(2, 0);
+DEBUG:  arguments in a distributed function must not contain subqueries
```
pykello: Why is this happening?

src/backend/distributed/planner/function_call_delegation.c
```diff
@@ -189,7 +192,9 @@ TryToDelegateFunctionCall(Query *query)
 		return NULL;
 	}
 
-	if (expression_tree_walker((Node *) funcExpr->args, contain_param_walker, NULL))
+	*hasParam = expression_tree_walker((Node *) funcExpr->args, contain_param_walker,
```
pykello: I think this should get the param kind so it can differentiate between `PARAM_EXTERN`, `PARAM_SUBLINK`, ...

The returned hasParam also needs to be renamed to `hasExternParam` (if my understanding is correct that prepared statements use `PARAM_EXTERN`. Please double check)

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -195,11 +195,20 @@ distributed_planner(Query *parse, int cursorOptions, ParamListInfo boundParams)
...
 			{
 				result = FinalizePlan(result, delegatePlan);
 			}
+			else if (hasParam)
+			{
+				/*
+				 * As in CreateDistributedPlannedStmt, try dissuade planner when planning
+				 * potentially failed due to unresolved prepared statement parameters.
+				 */
+				result->planTree->total_cost = FLT_MAX / 100000000;
```
marcocitus: Wouldn't this apply to all regular  postgres plans as well?

marcocitus: Ah, no, the check is done quite late. Maybe isDistributedFunctionCallWithParam?

serprex: We only set `hasParam` to true if we get rather deep into `TryToDelegateFunctionCall`

src/test/regress/expected/multi_mx_function_call_delegation.out
serprex: Fixed now

src/backend/distributed/planner/function_call_delegation.c
serprex: Updated

src/backend/distributed/planner/function_call_delegation.c
```diff
@@ -257,6 +292,24 @@ TryToDelegateFunctionCall(Query *query)
...
 	}
 
+	(void) expression_tree_walker((Node *) funcExpr->args, contain_param_walker,
+								  &walkerParamContext);
+	if (walkerParamContext.hasParam)
+	{
+		if (walkerParamContext.paramKind == PARAM_EXTERN)
+		{
+			*hasExternParam = true;
+			ereport(DEBUG1, (errmsg("external parameter found in arguments to "
```
pykello: Can you also add a test to hit this case?

pykello: I think this is unreachable, since if this is parametrized then distribution column is also parametrized and we have returned earlier. Maybe we can change this to an Assert() in future after careful consideration, but this is ok for now

PR #3058, opened 2019-09-30
PG_VERSION_NUM > 110000 should be PG_VERSION_NUM >= 110000

serprex:
Fix minor typos from #3041


pykello:
index.c has a `PG_VERSION_NUM > 12000`, which is weird. It is missing the equality sign and a 0.

PR #3060, opened 2019-10-01
MX Isolation Tests for Foreign keys from Distributrion to Reference tables vs DML

halilozanakgul:
Created isolation tests for distribution table to reference table foreign keys vs DML operations on MX.

### DIFF
--- a/src/test/regress/expected/isolation_dis2ref_foreign_keys_on_mx.out
+++ b/src/test/regress/expected/isolation_dis2ref_foreign_keys_on_mx.out
@@ -1,0 +1,544 @@
+Parsed test spec with 3 sessions
+
+starting permutation: s1-start-session-level-connection s1-begin-on-worker s1-delete s2-start-session-level-connection s2-begin-on-worker s2-insert s1-rollback-worker s2-commit-worker s1-stop-connection s2-stop-connection s3-display
+step s1-start-session-level-connection: 
+	SELECT start_session_level_connection_to_node('localhost', 57637);
+
+start_session_level_connection_to_node
+
+               
+step s1-begin-on-worker: 
+	SELECT run_commands_on_session_level_connection_to_node('BEGIN'); 
+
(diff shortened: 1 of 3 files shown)


src/test/regress/expected/isolation_dis2ref_foreign_keys_on_mx.out
```diff
@@ -0,0 +1,532 @@
...
+	SELECT run_commands_on_session_level_connection_to_node('BEGIN'); 
+
+run_commands_on_session_level_connection_to_node
+
+               
+step s2-insert: 
+        SELECT run_commands_on_session_level_connection_to_node('INSERT INTO dist_table VALUES (1, 2), (2, 1)');
+ <waiting ...>
+step s1-rollback-worker: 
+	SELECT run_commands_on_session_level_connection_to_node('ROLLBACK');
```
onderkalaci: I think instead of `ROLLBACK` we should prefer `COMMIT`, just that's a lot more common in real-world, so the tests are triggering those cases more commonly. 


I can see that `INSERT` fails if you commit on the s1 as-is.

So, why not we do `INSERT INTO dist_table ... ON CONFLICT DO UPDATE set value = value + 1`, and on `s3`, we print the content instead of the `count(*)` so that we see everything went smoothly?

Can you also consider doing that for the other tests that we do `ROLLBACK`?

src/test/regress/specs/isolation_dis2ref_foreign_keys_on_mx.spec
```diff
@@ -0,0 +1,166 @@
...
+}
+
+step "s2-stop-connection"
+{
+	SELECT stop_session_level_connection_to_node();
+}
+
+session "s3"
+
+step "s3-select-count"
```
onderkalaci: As I noted somewhere as well, I think in this test it is more interesting to see `SELECT * FROM dist_table ORDER BY x,y`

PR #3062, opened 2019-10-02
Refactor isolation tests

SaitTalhaNisanci:
DESCRIPTION: Refactor isolation tests.

Currently in mx isolation tests the setup is the same except the creation of tables. Isolation framework lets us define multiple `setup` stages, therefore I thought that we can put the `mx_setup` to one file and prepend this prior to running tests. 

How the structure works:
- cpp is used before running isolation tests to preprocess spec files. This way we can include any file we want to. Currently this is used to include mx common part.
- spec files are put to `/build/specs` for clear separation between generated files and template files
- a symbolic link is created for `/expected` in `build/expected/`.
- when running isolation tests, as the `inputdir`, `build` is passed so it runs the spec files from `build/specs` and checks the expected output from `build/expected`.

`/specs` is renamed as `/spec` because postgres first look at the `specs` file under current directory, so this is renamed to avoid that since we are running the isolation tests from `build/specs` now.

Note: now we use `//` instead of `#` in comments in spec files, because cpp interprets `#` as a directive and it ignores `//`.

This can be extended to put more steps to the common file, but as a general structure does this make sense? Can we benefit from this?


hanefi:
I wonder if we can do this using a make target for preprocessing, instead of creating a new python dependency.

Here is how Jelte created changes for preprocessing sql files: https://github.com/citusdata/citus/blob/a2eef99a6d56a0671a1ddb6eec611fb4d322cec1/src/backend/distributed/Makefile#L9-L12

SaitTalhaNisanci:
@JelteF I have looked at the commit https://github.com/citusdata/citus/commit/a2eef99a6d56a0671a1ddb6eec611fb4d322cec1. To me it seems that this is less readable and maintainable. When someone is looking at that, they will need to understand what those flags are etc. To me the python script seems more readable. We can see what it is doing and it is easy to maintain for someone else too. 

I feel like automatically regenerating the sql files isnt that useful here because what the script is doing is just to prepend the MX setup part. So we are probably not interested in the target file because we care about the core part of isolation tests, not MX setup part if I understand correctly. And if we want to see the target files we can just run tests and they will be generated.

JelteF:
> To me it seems that this is less readable and maintainable.

I understand what you mean but in this case I don't think it's true, because:
- We already have this "weird" command, so there's not really any extra
- The `man` pages quite clearly explain what the flags do
- All the flags containing an `M` are not actually needed to reproduce the behaviour with the script, just to create makefile dependency files. Without them the command becomes: `cpp -undef -w -P $@ $< > $@`, which I think is quicker to understand than a whole script.

I do definitely agree that the command could use a bit more comments explaining what the -M flags do.

The main advantages of using the C preprocessor to me are:
- Automatic regeneration of the final files when you change any of the input ones.
- We don't have to change the script when we want to use this for other common setup steps. 
- We can use it to include smaller pieces than a whole setup section.
- There's no magic filename endings. Instead you do `#include "somebasefile.spec"`, which is quite clear.

src/test/regress/Makefile
```diff
@@ -48,6 +50,14 @@ check: check-full
 # check-full triggers all tests that ought to be run routinely
 check-full: check-multi check-multi-mx check-multi-task-tracker-extra check-worker check-follower-cluster check-failure
 
+# this can be used to print a value of variable
+# ex: make print-generated_mx_files
+print-%  : ; @echo $* = $($*)
+
+$(generated_mx_files): $(citus_abs_srcdir)/specs/%: $(citus_abs_srcdir)/specs/base_%
+	echo $@
+	cd $(citus_abs_srcdir) && cpp -undef -w -P -MMD -MP -MT $@ $< > $@
```
JelteF: Can you use the `-MMF` flag to store the makefile dependency files in a separate directory.

src/test/regress/Makefile
```diff
@@ -48,6 +50,14 @@ check: check-full
...
+
```
JelteF: You have not `include`d the dependency files in the make file. Equivalent to this: https://github.com/citusdata/citus/blob/a2eef99a6d56a0671a1ddb6eec611fb4d322cec1/src/backend/distributed/Makefile#L44-L47

You can test that it works by checking changing `isolation_mx_common.spec` rebuilds the generated files.

src/test/regress/Makefile
```diff
@@ -26,6 +26,8 @@ MULTI_REGRESS_OPTS = --inputdir=$(citus_abs_srcdir) $(pg_regress_locale_flags) -
 pg_upgrade_check = $(citus_abs_srcdir)/upgrade/pg_upgrade_test.py
 citus_upgrade_check = $(citus_abs_srcdir)/upgrade/citus_upgrade_test.py
 
+template_mx_files = $(shell find $(citus_abs_srcdir)/specs/ -name 'base*mx.spec')
+generated_mx_files = $(patsubst $(citus_abs_srcdir)/specs/base_isolation%,$(citus_abs_srcdir)/specs/isolation%,$(template_mx_files))
```
JelteF: I think it's nice to run this on all spec files and write them to `src/test/regress/build/specs/` that way there's a clear separation between source files and non source files, and no magic file prefixes (only magic directories).

Looking at the code of the isolation tester in the postgres repo you can pass an `--inputdir` flag, which should be changed from the default `.` to `./build`. We actually already set this flag for the vanilla tester. However, it seems that it will first try to find them in `{outputdir}/specs`. Since this would be our source specs directory I think you would have to rename that directory as well, easiest would be to rename to `spec` (without the s)

SaitTalhaNisanci: In spec files, `#` means comment which is a directive in cpp. Currently we use `//` for comments in the base isolation files for mx(since cpp ignores `//`). I guess we can use `//` in other spec files too(it might be even better than having mixed comment symbol). But would that be hard for people to switch from `#` to `//` ?

JelteF: I think it's fine to switch comment symbol and indeed an advantage if it's the same symbol everywhere.

src/test/regress/Makefile
```diff
@@ -23,9 +23,13 @@ MULTI_INSTALLDIR=$(CURDIR)/tmp_check/install
 pg_regress_multi_check = $(PERL) $(citus_abs_srcdir)/pg_regress_multi.pl --pgxsdir="$(pgxsdir)" --bindir="$(bindir)" --libdir="$(libdir)" --majorversion="$(MAJORVERSION)" --postgres-builddir="$(postgres_abs_builddir)"  --postgres-srcdir="$(postgres_abs_srcdir)"
 MULTI_REGRESS_OPTS = --inputdir=$(citus_abs_srcdir) $(pg_regress_locale_flags) --launcher="$(citus_abs_srcdir)/log_test_times"
 
+MX_DEPDIR := .deps
```
JelteF: can be removed

src/test/regress/Makefile
```diff
@@ -23,9 +23,13 @@ MULTI_INSTALLDIR=$(CURDIR)/tmp_check/install
...
+
 pg_upgrade_check = $(citus_abs_srcdir)/upgrade/pg_upgrade_test.py
 citus_upgrade_check = $(citus_abs_srcdir)/upgrade/citus_upgrade_test.py
 
+template_mx_files = $(shell find $(citus_abs_srcdir)/spec/ -name '*.spec')
+generated_mx_files = $(patsubst $(citus_abs_srcdir)/spec/%,$(citus_abs_srcdir)/build/specs/%,$(template_mx_files))
```
JelteF: -> template_isolation_files
-> generated_isolation_files

src/test/regress/Makefile
```diff
@@ -48,6 +52,25 @@ check: check-full
...
+
+ISOLATION_DEPDIR=.deps/isolation
+ISOLATION_BUILDDIR=build/specs
+
+# this can be used to print a value of variable
+# ex: make print-generated_mx_files
+print-%  : ; @echo $* = $($*)
+
+$(ISOLATION_DEPDIR) $(ISOLATION_BUILDDIR):
+	mkdir -p $(citus_abs_srcdir)/$@
```
JelteF: Directories as targets mess make up. I suspect because modified time of directories can change even though we only care about the fact that it exists. So simply move this to the generated_isolation_fiels rule: https://github.com/citusdata/citus/pull/2977/files

src/test/regress/.gitignore
```diff
@@ -6,6 +6,7 @@
 /tmp_upgrade/
 /tmp_citus_upgrade/
 /tmp_citus_tarballs/
+/build/
 /results/
```
JelteF: Should this also contain `.deps`?

SaitTalhaNisanci: I guess we already have it here https://github.com/citusdata/citus/blob/master/.gitignore#L15

marcocitus:
I'm fine with this. @JelteF Maybe update your review status.

src/test/regress/Makefile
```diff
@@ -48,6 +50,29 @@ check: check-full
...
+# ex: make print-generated_isolation_files
+print-%  : ; @echo $* = $($*)
+
+.PHONY: create-symbolic-link
+
+create-symbolic-link:
+	mkdir -p $(citus_abs_srcdir)/build
+	ln -fsn $(citus_abs_srcdir)/expected $(citus_abs_srcdir)/build/
+
+$(generated_isolation_files): $(citus_abs_srcdir)/build/specs/%: $(citus_abs_srcdir)/spec/% create-symbolic-link
```
JelteF: the dependency here on `create-symbolic-link` doesn't really make sense. Also, because it's a `.PHONY` target I believe it will always cause a rebuild of the generated_isolation_files.

A better solution would be to do something like this:
```make
isolation_test_files=$(generated_isolation_files) create-symbolic-link
......
# Use this instead of generated_isolation_files
check-isolation: all tempinstall-main $(isolation_test_files)
```

src/test/regress/Makefile
```diff
@@ -48,6 +50,29 @@ check: check-full
...
+	@mkdir -p $(citus_abs_srcdir)/$(ISOLATION_DEPDIR) $(citus_abs_srcdir)/$(ISOLATION_BUILDDIR)
+	cd $(citus_abs_srcdir) && cpp -undef -w -P -MMD -MP -MF$(ISOLATION_DEPDIR)/$(*F).Po -MT$@ $< > $@
```
JelteF: Let's add a comment explaining these `cpp` options for future readers.

src/test/regress/spec/isolation_add_remove_node.spec
```diff
@@ -107,43 +107,43 @@ step "s2-commit"
 	COMMIT;
 }
 
-# session 1 adds a node, session 2 removes it, should be ok
+// session 1 adds a node, session 2 removes it, should be ok
```
onderkalaci: @JelteF and @SaitTalhaNisanci, I get the following error when I execute isolation tests locally:

```
syntax error at line 114: unexpected character "/"
```

Do you have any idea why? All the isolation tests with `//` comment fails. What was the reason to change this?

SaitTalhaNisanci: @onderkalaci `make check-isolation` works fine in my local. 

The reason for this change was that we have many common code in isolation tests and we thought of using c preprocessor to handle this. We use CPP to generate actual isolation files before testing. And `//` are ignored since it is a comment in CPP.

What do you run locally?

onderkalaci:
```
gcc --version
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/c++/4.2.1
Apple clang version 11.0.0 (clang-1100.0.33.17)
Target: x86_64-apple-darwin18.7.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

And, MacOS Majove 10.14.6 (18G2022)

JelteF: I think he meant what command to actually run the tests? Also `make check-isolation`?

onderkalaci: > I think he meant what command to actually run the tests? Also make check-isolation?

Yes, make check-isolation

JelteF: A quick workaround would probably be installing the gcc version of cpp and using that one in the Makefile here: https://github.com/citusdata/citus/pull/3062/files#diff-f89c636652a95930b80fdfc4290d15c1R84. I guess it's something weird about the `clang` version of `cpp` that it doesn't always strip comments away. The reason it was changed from `#` comments to `//` is that `#` is interpreted as a preprocessor command, like `#include`.

thanodnl: Ran into this today as well.

Pretty frustrating we break dev infra on a regular basis for OSX.

I think when we are changing build infra someone with a Mac should run it at least locally before we merge these things.

Would be great if we can treat this as an outage and fix it ASAP as it is non-trivial to change test output without being able to run these tests in your dev env. Especially with a code freeze upcoming.

Asking an annoying question, but what would it take to revert this PR?

JelteF: This is fixed now in #3706. 
> I think when we are changing build infra someone with a Mac should run it at least locally before we merge these things.

Yes I agree

> Would be great if we can treat this as an outage and fix it ASAP

Yes I agree. I think this was not done because it was weekend when this was first reported. Then during the week it was not picked up. I guess because it wasn't a problem for @onderkalaci at that point anymore.

thanodnl: Thanks @JelteF @SaitTalhaNisanci and @onderkalaci for chipping in.

PR #3063, opened 2019-10-02
Update certifi in regress pipenv

JelteF:
For some reason pipenv really wanted me to update this when I ran `pipenv install`.
To improve future `pipenv install` runs it's probably best to save this change.

### DIFF
--- a/src/test/regress/Pipfile.lock
+++ b/src/test/regress/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "705f2b4bd5a49fcd8d7a7d0b4f1bc90f92f8fa3640764b36689296365d41aefe"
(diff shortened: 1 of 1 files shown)


SaitTalhaNisanci:
I think here you can find a lot of useful informaton: https://github.com/pypa/pipenv/issues/2024

As far as I understand the hashes here are not used for verifying the version but rather act like a checksum.

From the link:

```
It works sort like how you check you downloaded the right file from, say, python.org. You download a file listed in the table, and after you’re done, check if the file’s actual hash matches the one listed in the table. pipenv lock saves all the hashes in the table locally ahead of time, so it doesn’t need to check online every time you create a new environment (and in some circumstances it can’t), but the idea is the same. It ensures this current download matches the information it got when you lock.
```
So I think it is better to send this as part of a main PR. Like if you have got this while enhancing update tests, you can include this change there. What do you think?

JelteF:
It actually updated certifi as well to a new version, so it makes sense that the hashes changed. Closing this in favor of #3069

Issue #3065, opened 2019-10-02
maintenance daemon might get stuck waiting for pg_dist_node lock


marcocitus:
The maintenance daemon currently takes an exclusive lock on pg_dist_node on every iteration:
https://github.com/citusdata/citus/blob/ec9fee1c922cf3690fb2bb5af699c0e045c653a5/src/backend/distributed/metadata/metadata_sync.c#L1281

It can easily get stuck waiting for a large transaction that takes a somewhat aggressive `pg_dist_node` lock such as `create_distributed_function`. If that transaction is in a distributed deadlock, the maintenance daemon is unable to resolve it.
https://github.com/citusdata/citus/blob/29f1ea079b975816b0248788bc9a0d70beaa273f/src/backend/distributed/commands/function.c#L607

We should perhaps add some timeout to the lock and maybe use a less aggressive lock.

serprex:
To reproduce issue, interleave these sessions
```
session 1:
create table t2(a int, b int);select create_distributed_table('t2', 'a');insert into t2 values (1, 2), (2, 3);begin;update pg_dist_node set metadatasynced = metadatasynced;
commit;begin;update pg_dist_node set metadatasynced = metadatasynced;
update t2 set b = 3 where a = 1;
update t2 set b = 3 where a = 2;

session 2:
select start_metadata_sync_to_node('localhost', 5434);
begin;update t2 set b = 4 where a = 2;
update t2 set b = 4 where a = 1;
```

pykello:
> https://github.com/citusdata/citus/blob/29f1ea079b975816b0248788bc9a0d70beaa273f/src/backend/distributed/commands/function.c#L607
> 
> We should perhaps add some timeout to the lock and maybe use a less aggressive lock.

I think the idea was this we wanted to get a lock that allows concurrent reads, but blocks on updates. I am not sure if I understand what we should do when this times out. Just ignore doing this or schedule it for maintenanced? What do you think instead of doing this, always schedule it for maintenanced?

pykello:
> > https://github.com/citusdata/citus/blob/29f1ea079b975816b0248788bc9a0d70beaa273f/src/backend/distributed/commands/function.c#L607
> > 
> > We should perhaps add some timeout to the lock and maybe use a less aggressive lock.
> 
[...]

PR #3068, opened 2019-10-03
Don't block for locks in SyncMetadataToNodes()

pykello:
Use less aggressive locks + don't wait for locks in `SyncMetadataToNodes()`. The idea
is that other processes shouldn't be able to block maintenance daemon, so it is able to
check for distributed deadlocks regularly.

Addresses part of #3065.

TODO:
- [x] Understand and fix the distributed_function failures in check

### DIFF
--- a/src/backend/distributed/metadata/metadata_sync.c
+++ b/src/backend/distributed/metadata/metadata_sync.c
@@ -1278,7 +1278,12 @@ SyncMetadataToNodes(void)
 		return METADATA_SYNC_SUCCESS;
 	}
 
-	if (!ConditionalLockRelationOid(DistNodeRelationId(), AccessShareLock))
+	/*
+	 * Request a RowExclusiveLock so we don't run concurrently with other
+	 * functions updating pg_dist_node, but allow concurrency with functions
+	 * which are just reading from pg_dist_node.
+	 */
+	if (!ConditionalLockRelationOid(DistNodeRelationId(), RowExclusiveLock))
 	{
 		return METADATA_SYNC_FAILED_LOCK;
 	}
@@ -1293,17 +1298,7 @@ SyncMetadataToNodes(void)
 		{
 			bool raiseInterrupts = false;
 
-			/*
-			 * Request a RowExclusiveLock so we don't run concurrently with other
-			 * functions updating pg_dist_node, but allow concurrency with functions
-			 * which are just reading from pg_dist_node.
-			 */
-			if (!ConditionalLockRelationOid(DistNodeRelationId(), RowExclusiveLock))
-			{
-				result = METADATA_SYNC_FAILED_LOCK;
-				break;
-			}
-			else if (!SyncMetadataSnapshotToNode(workerNode, raiseInterrupts))
+			if (!SyncMetadataSnapshotToNode(workerNode, raiseInterrupts))
 			{
 				result = METADATA_SYNC_FAILED_SYNC;
 			}
--- a/src/backend/distributed/metadata/metadata_sync.c
+++ b/src/backend/distributed/metadata/metadata_sync.c
@@ -1263,24 +1263,28 @@ DetachPartitionCommandList(void)
 
 /*
  * SyncMetadataToNodes tries recreating the metadata snapshot in the
- * metadata workers that are out of sync. Returns false if synchronization
- * to at least one of the workers fails.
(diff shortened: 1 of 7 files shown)


src/backend/distributed/transaction/worker_transaction.c
```diff
@@ -97,7 +97,7 @@ TargetWorkerSetNodeList(TargetWorkerSet targetWorkerSet, LOCKMODE lockMode)
 	{
 		WorkerNode *workerNode = (WorkerNode *) lfirst(workerNodeCell);
 		if (targetWorkerSet == WORKERS_WITH_METADATA &&
-			!workerNode->hasMetadata)
+			(!workerNode->hasMetadata || !workerNode->metadataSynced))
```
pykello: Otherwise some of the depended objects might not exist yet and this will error out.

src/backend/distributed/metadata/metadata_sync.c
```diff
@@ -1290,7 +1300,21 @@ SyncMetadataToNodes(void)
...
 			{
-				result = false;
+				result = METADATA_SYNC_FAILED_SYNC;
+			}
+
+			/*
+			 * Try locking pg_dist_node so we can update it. MarkNodeMetadataSynced()
+			 * will get row locks on the updated row, so RowExclusiveLock is good
+			 * enough here. The only other function which updates rows with metadataSynced
+			 * equal to false is start_metadata_sync_to_node() which gets an ExclusiveLock,
```
pykello: Also `create_distributed_function()`, which also gets an `ExclusiveLock` before updating.

src/backend/distributed/metadata/metadata_sync.c
```diff
@@ -1263,24 +1263,34 @@ DetachPartitionCommandList(void)
...
 	}
 
-	LockRelationOid(DistNodeRelationId(), ExclusiveLock);
+	/*
+	 * We request a ShareLock here, because we want to avoid running at
+	 * the same time master_update_node() or some other function is updating
+	 * pg_dist_node, but we should be able to run concurrently with other
+	 * functions that are scanning pg_dist_node.
+	 */
+	if (!ConditionalLockRelationOid(DistNodeRelationId(), ShareLock))
```
marcocitus: maybe `RowExclusiveLock`? (which is a lot lighter than ShareLock, but still blocked by ExclusiveLock)

src/backend/distributed/metadata/metadata_sync.c
```diff
@@ -1290,7 +1300,21 @@ SyncMetadataToNodes(void)
...
+
+			/*
+			 * Try locking pg_dist_node so we can update it. MarkNodeMetadataSynced()
+			 * will get row locks on the updated row, so RowExclusiveLock is good
+			 * enough here. The only other function which updates rows with metadataSynced
+			 * equal to false is start_metadata_sync_to_node() which gets an ExclusiveLock,
+			 * so getting a RowExclusiveLock here is enough to protect us from concurrent
+			 * execution with that.
+			 */
+			else if (!ConditionalLockRelationOid(DistNodeRelationId(), RowExclusiveLock))
```
marcocitus: I don't think we need this if we already grabbed a heavier lock before.

src/backend/distributed/transaction/worker_transaction.c
marcocitus: Should it be in a separate commit?

src/backend/distributed/metadata/metadata_sync.c
marcocitus: or even AccessShareLock just to make sure it never blocks.

serprex: Yeah, I was mixing up `ShareLock` with `RowShareLock` in discussion

marcocitus: Looking into it more, I think we should first do some basic check using an AccessShareLock to see if there is any work to do and only take more aggressive locks if there is. `master_add_node` takes an exclusive lock and can take a really long time if it's copying reference tables to the new node and we don't want deadlock detection to be blocked during that period.

src/backend/distributed/transaction/worker_transaction.c
pykello: done

src/backend/distributed/metadata/metadata_sync.c

PR #3070, opened 2019-10-03
Add more upgrade tests

JelteF:
This PR improves the upgrade tests by adding more of them.

It also fixes a bug found during the adding of these tests. The
citus_truncate_trigger was added as an `AFTER TRUNCATE` trigger as opposed to a
`BEFORE TRUNCATE` trigger.

TODO:
- [x] Test for append tables


src/backend/distributed/sql/udfs/citus_finish_pg_upgrade/latest.sql
```diff
@@ -52,7 +52,7 @@ BEGIN
     FOR table_name IN SELECT logicalrelid FROM pg_catalog.pg_dist_partition
     LOOP
         trigger_name := 'truncate_trigger_' || table_name::oid;
-        command := 'create trigger ' || trigger_name || ' after truncate on ' || table_name || ' execute procedure pg_catalog.citus_truncate_trigger()';
```
JelteF: @marcocitus This change brings triggers created by the finsh_pg_upgrade UDF in line with the ones created by `create_distributed_table`: https://github.com/citusdata/citus/blob/492d1b2cbad626a5e3cfee01e972dc922f18dcfb/src/backend/distributed/commands/create_distributed_table.c#L1163

I'm not sure which one is better though, or if it really matters. I could also change the C code to create an AFTER trigger instead if that's what we want.

src/test/regress/sql/upgrade_distributed_table_after.sql
```diff
@@ -1,28 +1,99 @@
...
+  WHERE sequence_name LIKE 'pg_dist_%'
+  AND sequence_name NOT IN (
+    -- these ones are restored above
+    'pg_dist_shardid_seq',
+    'pg_dist_placement_placementid_seq',
+    'pg_dist_groupid_seq',
+    'pg_dist_node_nodeid_seq',
+    'pg_dist_colocationid_seq',
+    -- jobid is not actually stored in tables, so it does not need to be restored
+    'pg_dist_jobid_seq'
```
JelteF: @marcocitus Currently we don't restore jobid after an upgrade. Looking at it's definition that seems reasonable:
https://github.com/citusdata/citus/blob/master/src/backend/distributed/sql/citus--7.0-1.sql#L110-L118

 I wanted to double check that it's indeed fine that we don't restore it.

src/test/regress/expected/upgrade_ref2ref_after.out
```diff
@@ -0,0 +1,169 @@
+SET search_path TO upgrade_ref2ref, public;
+BEGIN;
+SET LOCAL citus.multi_shard_modify_mode TO 'sequential';
+\d ref_table_1
```
onderkalaci: Minor: `\d` may not be consistent across postgres versions. Is there  any view that you can utilize  in `multi_test_helpers`  to get this information?

src/test/regress/expected/upgrade_type_after.out
```diff
@@ -0,0 +1,30 @@
...
+\d tt
+              Table "upgrade_type.tt"
+ Column |  Type   | Collation | Nullable | Default 
+--------+---------+-----------+----------+---------
+ a      | integer |           | not null | 
+ b      | type1   |           |          | 
+Indexes:
+    "tt_pkey" PRIMARY KEY, btree (a)
+
+\dt type1
```
onderkalaci: Similarly, can you use something like `SELECT count(*) FROM pg_type WHERE typename ='type1' ` in order to guard us from future changes?

JelteF: It's not actually that important, since that would be testing pg_upgrade instead of the citus upgrade. So I removed it. The tests below still indirectly ensure that it exists.

src/test/regress/sql/upgrade_distributed_table_after.sql
marcocitus: That's fine. We don't actually use that sequence for anything. It should be dropped.

src/backend/distributed/sql/udfs/citus_finish_pg_upgrade/latest.sql
marcocitus: Be careful about changing these files. We can no longer change 9.0.

marcocitus: I don't know whether it matters, that's worth figuring out (e.g. do tests still pass if we do AFTER triggers in C, does truncate MX concurrently with other commands still behave as expected)

JelteF: Changed the C code to create an AFTER trigger and replace all old truncate triggers with an AFTER trigger.

src/test/regress/sql/upgrade_distributed_table_after.sql
JelteF: I added a commit that drops this sequence.

src/test/regress/upgrade/README.md
```diff
@@ -10,12 +10,12 @@ Before running the script, make sure that:
 - Citus is installed to both of the postgres versions. For each postgres version:
   - In citus source directory run:
 
-  ```bash
-    make clean
-    ./configure PG_CONFIG=<your path to postgres pg config>
-    PG_CONFIG=<your path to postgres pg config> make
-    sudo PG_CONFIG=<your path to postgres pg config> make install
-  ```
+```bash
```
thanodnl: removing this indent messes up the indenting on github, could you revert this?

src/test/regress/upgrade/README.md
```diff
@@ -42,7 +42,10 @@ How the postgres upgrade test works:
 
 - Temporary folder `tmp_upgrade` is created in `src/test/regress/`, if one exists it is removed first.
 - Database is initialized and citus cluster is created(1 coordinator + 2 workers) with old postgres.
-- `before_pg_upgrade_schedule` is run with `pg_regress`. This schedule does not drop any tables or data so that we can verify upgrade.
+- `before_pg_upgrade_schedule` is run with `pg_regress`. This schedule sets up any
+  objects and data that will be tested for preservation after the upgrade. It
+  also runs all the tests that will be run after the upgrade, to make sure their
```
thanodnl: this line is not up to date with our changes that the after schedule is now executed twice.

src/backend/distributed/sql/citus--8.3-1--9.0-1.sql
```diff
@@ -174,4 +174,41 @@ COMMENT ON FUNCTION worker_apply_sequence_command(text,regtype)
...
+        EXECUTE command;
+    END LOOP;
+END;
+$$;
+
+SELECT ensure_truncate_trigger_is_after();
+DROP FUNCTION ensure_truncate_trigger_is_after;
+
+-- This sequence is unused
+DROP SEQUENCE pg_catalog.pg_dist_jobid_seq;
```
hanefi: @jeltef this is indeed unused on community. Otoh, there are some references to this sequence in Enterprise tests, and that's why merges to Enterprise fail now.

It is easy to fix, but it took a bit of my time to understand what is going on. I will update those tests first thing in the morning.

My point is that it is a really nice practice to merge changes into Enterprise after merging your PR in community. No matter how small and simple your change is.

PR #3071, opened 2019-10-03
Check command type in TryToDelegateFunctionCall

marcocitus:
We do a lot of checks in `TryToDelegateFunctionCall` to see if the query has the right format for a `SELECT distributed_function(...)`, but we neglect to check whether the query is actually a SELECT.

### DIFF
--- a/src/backend/distributed/planner/function_call_delegation.c
+++ b/src/backend/distributed/planner/function_call_delegation.c
@@ -141,6 +141,12 @@ TryToDelegateFunctionCall(Query *query, bool *hasExternParam)
 		return NULL;
 	}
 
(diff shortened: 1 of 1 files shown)


src/backend/distributed/planner/function_call_delegation.c
```diff
@@ -141,6 +141,12 @@ TryToDelegateFunctionCall(Query *query, bool *hasExternParam)
 		return NULL;
 	}
 
+	if (query->commandType != CMD_SELECT)
```
onderkalaci: Well, I'm also good with removing the above check. `if (query == NULL)` we'd have already crashed :)

PR #3073, opened 2019-10-03
Multiuser testing 9.1: fix distributed function ownership

serprex:
While doing multi user testing I found that distributed functions' owners were incorrect on workers

### DIFF
--- a/src/backend/distributed/commands/function.c
+++ b/src/backend/distributed/commands/function.c
@@ -53,7 +53,7 @@
 	(strncmp(arg, prefix, strlen(prefix)) == 0)
 
 /* forward declaration for helper functions*/
-static char * GetFunctionDDLCommand(const RegProcedure funcOid);
+static List * GetFunctionDDLCommand(const RegProcedure funcOid);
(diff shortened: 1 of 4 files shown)


src/backend/distributed/commands/function.c
```diff
@@ -157,7 +164,11 @@ create_distributed_function(PG_FUNCTION_ARGS)
 	EnsureDependenciesExistsOnAllNodes(&functionAddress);
 
 	ddlCommand = GetFunctionDDLCommand(funcOid);
-	SendCommandToWorkers(ALL_WORKERS, ddlCommand);
+	foreach(ddlCommandCell, ddlCommand)
+	{
+		char *ddlCommandString = lfirst(ddlCommandCell);
+		SendCommandToWorkers(ALL_WORKERS, ddlCommandString);
```
marcocitus: We should probably create this as the current user to ensure transactional semantics when running as non-superuser. We should also check ownership (or superuser) in this function before doing anything else.

marcocitus:
Superseded by #3076

PR #3076, opened 2019-10-04
Make sure that distributed functions always have the correct user

onderkalaci:
Alternative (and based on) to #3073 

Objectives:

(a) both super user and regular user should have the correct owner for the function on the worker
(b) The transactional semantics should work fine for both super user and regular user when user creates a function inside a transaction block
(c) non-super-user and non-function owner should get a reasonable error message if tries to distribute the function

Co-authored-by: serprex (should I also have @serprex)

### DIFF
--- a/src/backend/distributed/commands/function.c
+++ b/src/backend/distributed/commands/function.c
@@ -33,6 +33,7 @@
 #include "distributed/commands/utility_hook.h"
 #include "distributed/deparser.h"
 #include "distributed/maintenanced.h"
+#include "distributed/master_metadata_utility.h"
 #include "distributed/master_protocol.h"
 #include "distributed/metadata/distobject.h"
 #include "distributed/metadata/pg_dist_object.h"
@@ -54,6 +55,7 @@
 
 /* forward declaration for helper functions*/
 static char * GetFunctionDDLCommand(const RegProcedure funcOid);
+static char * GetFunctionAlterOwnerCommand(const RegProcedure funcOid);
 static int GetDistributionArgIndex(Oid functionOid, char *distributionArgumentName,
 								   Oid *distributionArgumentOid);
 static int GetFunctionColocationId(Oid functionOid, char *colocateWithName, Oid
@@ -144,6 +146,7 @@ create_distributed_function(PG_FUNCTION_ARGS)
 		colocateWithTableName = text_to_cstring(colocateWithText);
 	}
 
+	EnsureFunctionOwner(funcOid);
 
 	ObjectAddressSet(functionAddress, ProcedureRelationId, funcOid);
 	ErrorIfFunctionDependsOnExtension(&functionAddress);
@@ -157,7 +160,8 @@ create_distributed_function(PG_FUNCTION_ARGS)
 	EnsureDependenciesExistsOnAllNodes(&functionAddress);
 
 	ddlCommand = GetFunctionDDLCommand(funcOid);
-	SendCommandToWorkers(ALL_WORKERS, ddlCommand);
+
+	SendCommandToWorkersAsUser(ALL_WORKERS, CurrentUserName(), ddlCommand);
 
 	MarkObjectDistributed(&functionAddress);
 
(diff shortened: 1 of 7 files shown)


src/backend/distributed/commands/function.c
```diff
@@ -555,8 +563,94 @@ GetFunctionDDLCommand(const RegProcedure funcOid)
 	/* revert back to original search_path */
 	PopOverrideSearchPath();
 
-	sql = TextDatumGetCString(sqlTextDatum);
-	return sql;
+	createFunctionSQL = TextDatumGetCString(sqlTextDatum);
+	alterFunctionOwnerSQL = GetFunctionAlterUserCommand(funcOid);
+
+	appendStringInfo(ddlCommand, "%s;%s", createFunctionSQL, alterFunctionOwnerSQL);
```
onderkalaci: I think it always makes sense to append `alterFunctionOwnerSQL ` to `createFunctionSQL ` when `GetFunctionDDLCommand ` is called. Because we rely ob this on two other cases: (a) create or replace function (b) add node  In both cases, we should append the owner of the function.

Let mw know if you think otherwise

src/test/regress/expected/multi_multiuser.out
```diff
@@ -428,7 +440,72 @@ INSERT INTO full_access_user_schema.t1 VALUES (1),(2),(3);
...
+(2 rows)
+
+SELECT wait_until_metadata_sync();
+ wait_until_metadata_sync 
+--------------------------
+ 
+(1 row)
+
+-- now, make sure that even if we use super-user, we should be able to
+-- use the function created in the transaction
```
marcocitus: I think the role is still `usage_access` here.

src/backend/distributed/commands/function.c
```diff
@@ -555,8 +563,94 @@ GetFunctionDDLCommand(const RegProcedure funcOid)
...
+		ReleaseSysCache(proctup);
+	}
+	else if (!OidIsValid(funcOid) || !HeapTupleIsValid(proctup))
+	{
+		ereport(ERROR, (errmsg("cannot find function with oid: %d", funcOid)));
+	}
+
+	qualifiedFunctionName = quote_qualified_identifier(schemaName, functionName);
+
+	{
```
marcocitus: I don't think this scope is needed.

src/backend/distributed/commands/function.c
```diff
@@ -555,8 +563,94 @@ GetFunctionDDLCommand(const RegProcedure funcOid)
...
+
+		PushOverrideSearchPath(overridePath);
+
+		/*
+		 * If the function exists we want to use pg_get_function_identity_arguments to
+		 * serialize its canonical arguments
+		 */
+		qualifiedFunctionArgumentsSQL =
+			DirectFunctionCall1(pg_get_function_identity_arguments,
+								ObjectIdGetDatum(funcOid));
```
marcocitus: Can we use `regprocedureout` here?

src/backend/distributed/commands/function.c
```diff
@@ -555,8 +563,84 @@ GetFunctionDDLCommand(const RegProcedure funcOid)
...
+	return ddlCommand->data;
+}
+
+
+/*
+ * GetFunctionAlterUserCommand returns "ALTER FUNCTION .. SET OWNER .." statement for
+ * the specified function.
+ */
+static char *
+GetFunctionAlterUserCommand(const RegProcedure funcOid)
```
marcocitus: Maybe AlterOwner

src/test/regress/expected/multi_multiuser.out
```diff
@@ -428,7 +440,72 @@ INSERT INTO full_access_user_schema.t1 VALUES (1),(2),(3);
...
+ 
+(1 row)
+
+SELECT usage_access_func_second(1, 2,3,4,5) FROM full_access_user_schema.t1 LIMIT 1; 
+ usage_access_func_second 
+--------------------------
+ usage_access
+(1 row)
+
+ROLLBACK;
```
marcocitus: Should we add a another test to confirm that superuser can call create_distributed_function for a function created by a regular user?

onderkalaci: makes sense, I'll add

src/backend/distributed/commands/function.c
```diff
@@ -555,8 +563,84 @@ GetFunctionDDLCommand(const RegProcedure funcOid)
...
+	 * serialize its canonical arguments
+	 */
+	functionSignatureDatum =
+		DirectFunctionCall1(regprocedureout, ObjectIdGetDatum(funcOid));
+
+	/* revert back to original search_path */
+	PopOverrideSearchPath();
+
+	/* regprocedureout returns cstring */
+	functionSignature = DatumGetPointer(functionSignatureDatum);
```
marcocitus: Maybe `DatumGetCString` (same thing)

PR #3077, opened 2019-10-04
Use citus docker repositories instead of citusdata in CI jobs

SaitTalhaNisanci:
We will also need to change `stylechecker` image to use `citus` repo in all of our images. I will build that image and update it soon.


hanefi:
This is a good place to practice merging changes into enterprise. Let's have a quick chat after this one is merged into master

SaitTalhaNisanci:
Yeah we dont have `werror` tag in the new built images anyways :)

PR #3078, opened 2019-10-05
Prototype: INSERT...SELECT with repartitioning

marcocitus:
DESCRIPTION: INSERT...SELECT with repartitioning

Opening a PR for visibility. Still very buggy, does not build on PG12.

We currently have two methods of executing an INSERT...SELECT: 
- in parallel between co-located shards, used when the distribution column value is preserved between source and target table
- via the coordinator using COPY, used in all other cases

The first method scales horizontally, but the second method can get bottlenecked on single core performance on the coordinator, and may cause the coordinator to run out of disk space.

This PR adds a third method:
- distributed repartitioning of the select result, used in cases where the SELECT is a distributed query that does not require a merge on the coordinator

The INSERT..SELECT runs in 3 phases:
1. Run SELECT tasks wrapped in worker_predistribute_query_result to partition query results into fragments which correspond to the hash ranges of the shards in the destination table
2. Fetch fragments to the corresponding destination shard using fetch_intermediate_result using one call per node pair
3. Run INSERT INTO .. SELECT .. FROM read_intermediate_result(..) into the shards of the destination table, where read_intermediate_result reads the concatenation of fragments that correspond to the hash range of the shard

In each phase, the tasks are executed by the adaptive executor. Empty tasks are pruned away during execution.

### DIFF
--- a/src/backend/distributed/executor/partition_intermediate_results.c
+++ b/src/backend/distributed/executor/partition_intermediate_results.c
@@ -107,16 +107,16 @@ static void PartitionedResultDestReceiverDestroy(DestReceiver *destReceiver);
 
 
 /* exports for SQL callable functions */
-PG_FUNCTION_INFO_V1(create_hash_partitioned_intermediate_result);
+PG_FUNCTION_INFO_V1(worker_predistribute_query_result);
 
 
 /*
- * create_hash_partitioned_intermediate_result executes a query and writes the results
+ * worker_predistribute_query_result executes a query and writes the results
  * into a set of local files according to the partition scheme and the partition
  * column.
  */
 Datum
-create_hash_partitioned_intermediate_result(PG_FUNCTION_ARGS)
+worker_predistribute_query_result(PG_FUNCTION_ARGS)
 {
 	ReturnSetInfo *resultInfo = (ReturnSetInfo *) fcinfo->resultinfo;
 
@@ -173,7 +173,7 @@ create_hash_partitioned_intermediate_result(PG_FUNCTION_ARGS)
 	 * Intermediate results will be stored in a directory that is derived
 	 * from the distributed transaction ID.
 	 */
-	BeginOrContinueCoordinatedTransaction();
+	//BeginOrContinueCoordinatedTransaction();
 
 	/* parse the query */
 	query = ParseQueryString(queryString);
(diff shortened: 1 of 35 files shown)


src/backend/distributed/executor/partition_intermediate_results.c
```diff
@@ -0,0 +1,639 @@
...
+		StringInfo resultId = makeStringInfo();
+		StringInfo filePathStringInfo = makeStringInfo();
+
+		appendStringInfo(resultId, "%s.%d", resultIdPrefix, partitionIndex);
+
+		filePath = QueryResultFileName(resultId->data);
+		appendStringInfoString(filePathStringInfo, filePath);
+
+		/* initialize the entry but do not create the file yet */
+		partitionFileArray[partitionIndex].fileDescriptor = 0;
```
marcocitus: The FileOutputStream API changed in PG12, causing builds to fail on PG12.

pykello:
Just for sake of documentation, couple of failed ideas:

1.
As an alternative to `worker_predistribute_query_result()`, we can also create an aggregate `write_intermediate_result(resultId, tuple)` (which is somehow `COPY` as a UDF) and do the predistribution in a query like:

```sql
SELECT write_intermediate_result(get_result_id(...), q)
FROM ( $selectQuery ) q
GROUP BY shard_id_func(...);
```

This can work perfectly, and is simpler (my implementation of `write_intermediate_result` was around 100 LOC and removed the need for some of the structures in `worker_predistribute_query_result`), except that its performance isn't great. A simple benchmark showed that it was 10-20% slower than the `COPY` command, and I think the reason was how postgres does function calls and aggregation, which we cannot optimize. So abandoning this idea, although looked pretty interesting to me.

2. 
Besides that, to support both hash partitioning and range partitioning, we can either pass a bool or enum arg to `worker_predistribute_query_result`, or alternatively replace partition column index arg with a `pg_node_tree` value which can represent the expression to use as sharding value. Although this is also pretty interesting, we don't require that flexibility and it might have some performance regressions, so also abandoning this idea too and sticking with the new enum/bool arg.

marcocitus:
For range-partitioning, one challenge might be workloads with a very large number of ranges. I don't know if we should do something special, or just write a really big query string that contains the full array. We also need to consider ranges of composite types.

marcocitus:
> A simple benchmark showed that it was 10-20% slower than the COPY command,

Was it also slower than worker_predistribute_query_result?

marcocitus:
Replaced by #3376

PR #3082, opened 2019-10-07
Fix enum add value order and pg12

thanodnl:
DESCRIPTION: Fix order for enum values and correctly support pg12

PG 12 introduces `ALTER TYPE ... ADD VALUE ...` during transactions. Earlier versions would error out when called in a transaction, hence we connect to workers outside of the transaction which could cause inconsistencies on pg12 now that postgres doesn't error with this syntax anymore.

During the implementation of this fix it became apparent there was an error with the ordering of enum labels when the type was recreated. A patch and test have been included.

### DIFF
--- a/src/test/regress/expected/distributed_types_0.out
+++ b/src/test/regress/expected/distributed_types_0.out
@@ -100,6 +100,20 @@ SELECT * FROM t4;
 
 -- ALTER TYPE ... ADD VALUE does not work in transactions
 COMMIT;
+-- verify order of enum labels
+SELECT string_agg(enumlabel, ',' ORDER BY enumsortorder ASC) FROM pg_enum WHERE enumtypid = 'type_tests.te2'::regtype;
+ string_agg 
+------------
+ yes,no
+(1 row)
+
+SELECT run_command_on_workers($$SELECT string_agg(enumlabel, ',' ORDER BY enumsortorder ASC) FROM pg_enum WHERE enumtypid = 'type_tests.te2'::regtype;$$);
+    run_command_on_workers    
+------------------------------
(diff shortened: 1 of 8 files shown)


src/test/regress/expected/distributed_types_0.out
```diff
@@ -100,6 +100,20 @@ SELECT * FROM t4;
 
 -- ALTER TYPE ... ADD VALUE does not work in transactions
 COMMIT;
+-- verify order of enum labels
+SELECT string_agg(enumlabel, ',' ORDER BY enumsortorder ASC) FROM pg_enum WHERE enumtypid = 'type_tests.te2'::regtype;
+ string_agg 
+------------
+ yes,no
```
marcocitus: is it worth making this list longer? or is the order deterministic?

thanodnl: I think it is deterministic, I scan the index that contains the order column as the second column. But am not super familiar with how to scan indices. Didn't see a ASC/DESC option for a ScanKey.

src/test/regress/sql/distributed_types_xact_add_enum_value.sql
```diff
@@ -0,0 +1,33 @@
+SET citus.next_shard_id TO 20040000;
+
```
hanefi: Now that we have differing test outputs, I think we should print PG versions in these tests now:

```
SHOW server_version \gset
SELECT substring(:'server_version', '\d+')::int > 11 AS version_above_eleven;
```

thanodnl: Good idea, have it added

PR #3083, opened 2019-10-07
add support to run citus upgrade tests locally

SaitTalhaNisanci:
DESCRIPTION: add support to run citus upgrade tests locally

Because of the logic of how we run our tests on circleci, it wasnt possible to run citus upgrade tests locally. This PR adds the necessary logic to run citus upgrades locally. 

There is a script, which is also used in our docker images, to generate citus tarballs. If `citus-old-version` arg is passed to our citus upgrade script:
- Citus tarballs for `citus-old-version` and `current code` are created.
- `--citus-pre-tar` and `--citus-post-tar` are updated with the generated citus tarballs.
- the rest of the execution is the same.

I have created two targets for local run, `check-citus-upgrade-mixed-local`, and `check-citus-upgrade-local`. 

When using these targets the only thing user should do is to supply the citus-old-version, for example to run citus upgrade tests with v8.0.0 -> current code:

```bash
make check-citus-upgrade-local citus-old-version=v8.0.0
```

or for mixed mode:

```bash
make check-citus-upgrade-mixed-local citus-old-version=v8.0.0
```

### DIFF
--- a/src/test/regress/upgrade/generate_citus_tarballs.sh
+++ b/src/test/regress/upgrade/generate_citus_tarballs.sh
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+set -euxo pipefail
 
 citus_old_version=$1
 
@@ -16,7 +17,8 @@ install_citus_and_tar() {
   make "-j$(nproc)" && mkdir -p "${installdir}" && make DESTDIR="${installdir}" install
 
   cd "${installdir}" && find . -type f -print > "${builddir}/files.lst"
-  tar cvf "${basedir}/install-citus${citus_version}.tar" `cat ${builddir}/files.lst`
+
+  tar cvf "${basedir}/install-citus${citus_version}.tar" $(cat "${builddir}"/files.lst)
   mv "${basedir}/install-citus${citus_version}.tar" "${base}/install-citus${citus_version}.tar"
 
   cd "${builddir}" && rm -rf install files.lst && make clean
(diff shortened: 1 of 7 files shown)


src/test/regress/upgrade/generate_citus_tarballs.sh
```diff
@@ -0,0 +1,35 @@
...
+  git clone --branch $citus_version https://github.com/citusdata/citus.git --depth 1 citus_$citus_version  
+  builddir="${basedir}/build"
+
+  # do everything in a subdirectory to avoid clutter in current directory
+  mkdir -p "${builddir}" && cd "${builddir}"
+
+  "${citus_dir}/configure" --without-libcurl
+
+  installdir="${builddir}/install"
+  make && mkdir -p "${installdir}" && make DESTDIR="${installdir}" install
```
JelteF: Let's use those cores
```suggestion
  make "-j$(nproc)" && mkdir -p "${installdir}" && make DESTDIR="${installdir}" install
```

src/test/regress/upgrade/generate_citus_tarballs.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
```
JelteF: Good to fix the warnings that `shellcheck` reports on this file

src/test/regress/upgrade/generate_citus_tarballs.sh
```diff
@@ -0,0 +1,58 @@
...
+
```
JelteF: Let's add `set -euxo pipefail`, so the script will fail fast.

PR #3089, opened 2019-10-08
Move C files into the appropriate directories

marcocitus:
Super minor thing that keeps bugging me.


serprex:
Somewhat related: what's going on with the header file layout? Feels like for the most part they exist in a flat `include/distributed/header.h` but now some are existing in subdirectories. Is it desirable that header files should exist in a subdirectory matching the subdirectory they exist in in backend?

serprex:
codecov issue is because this needs to be updated: https://github.com/citusdata/citus/blob/622ee54c95c172ca54047d296f69d440910cd3d4/.codecov.yml#L11

marcocitus:
> Is it desirable that header files should exist in a subdirectory matching the subdirectory they exist in in backend?

I think yes.

serprex:
@marcocitus you never addressed my note about `.codecov.yml`

marcocitus:
> @marcocitus you never addressed my note about .codecov.yml

oops, fixed it on master

PR #3092, opened 2019-10-08
Remove Postgres 10 support

### LINKED ISSUE #2970: Drop Postgres 10 support
We generally support 2 major versions. 9.0 will add Postgres 12 support, meaning we can drop Postgres 10.

serprex:
DESCRIPTION: Remove Postgres 10 support

Current issues:

- Circle CI images fail build because I don't know how to prevent
- There may be some _0.out files in last commit which should not be removed-- fixing CI will verify this

Fixes #2970


marcocitus:
@SaitTalhaNisanci any thoughts on how to stop building for PG10?

SaitTalhaNisanci:
Sorry I saw this now, this is because we had pg10->pg11 upgrade test therefore we were still building pg10 in `build` job. I am removing it now.

src/test/regress/expected/multi_view.out
```diff
@@ -913,7 +905,7 @@ EXPLAIN (COSTS FALSE) SELECT et.* FROM recent_10_users JOIN events_table et USIN
  Limit
    ->  Sort
          Sort Key: remote_scan."time" DESC
-         ->  Custom Scan (Citus Real-Time)
+         ->  Custom Scan (Citus Adaptive)
```
pykello: Do you know why were these Real-Time previously?

serprex: I generated the file with check-multi, it's been left with Real-Time since adaptive executor was brought in due to normalization

PR #3093, opened 2019-10-09
(NO-MERGE, POC) Add backtrace to test outputs

pykello:
Build with `--enable-debug`, run `make -C src/test/regress check-base`, look at the end of `src/test/regress/results/multi_cluster_management.out` or `regression.diffs`. You should see the backtrace. This is what I get: https://gist.github.com/pykello/b5741b1311bf0bff7bb3f30d1826669c#file-multi_cluster_management-out-L94

Assumption is that current user is passwordless sudoer so it can run `sudo gdb`.

### DIFF
--- a/src/test/regress/bin/psql-bt.py
+++ b/src/test/regress/bin/psql-bt.py
@@ -1,0 +1,36 @@
+#!/usr/bin/env python
+
+import subprocess, sys, os
+
+args = " ".join(map(lambda s: '"%s"' % s if ' ' in s else s, sys.argv[1:]))
+
+psql = subprocess.Popen([args + " 2>&1"], stdin=subprocess.PIPE, stdout=subprocess.PIPE, env=os.environ, shell=True)
+
+psql.stdin.write("SELECT pg_backend_pid();\n")
+psql.stdout.readline() # SELECT pg_backend_pid();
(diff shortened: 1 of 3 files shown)


src/test/regress/bin/psql-bt.py
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env python
+
+import subprocess, sys, os
+
+args = " ".join(map(lambda s: '"%s"' % s if ' ' in s else s, sys.argv[1:]))
+
+psql = subprocess.Popen([args + " 2>&1"], stdin=subprocess.PIPE, stdout=subprocess.PIPE, env=os.environ, shell=True)
```
serprex: `2>&1` can instead be `stderr=subprocess.STDOUT` https://stackoverflow.com/questions/11495783/redirect-subprocess-stderr-to-stdout

pykello: Yeah, the problem was that logs are output to stderr and everything else are output to stdout. In the final output of the psql process we need to output both, and getting ordering of two correct manually was more difficult than `2>&1`.

In the gdb process the ordering doesn't matter for us, so I did as you suggested.

pykello:
Closing, #4044 is solving this more broadly with a better approach.

PR #3096, opened 2019-10-09
CALL delegation: apply strip_implicit_coercions to distribution argument

### LINKED ISSUE #3085: CALL not delegated when distribution argument is bigint
An implicit cast from bigint to int prevents delegation:

```sql
CREATE TYPE invite_resp AS ENUM ('yes', 'no', 'maybe'); 

CREATE TABLE event_responses (
  event_id bigint,
  user_id bigint,
  response invite_resp,
  primary key (event_id, user_id)
);
SELECT create_distributed_table('event_responses', 'event_id');
CREATE OR REPLACE PROCEDURE register_for_event(p_event_id bigint, p_user_id bigint, p_choice invite_resp)
LANGUAGE plpgsql AS $fn$
BEGIN
   INSERT INTO event_responses VALUES (p_event_id, p_user_id, p_choice)
   ON CONFLICT (event_id, user_id)
   DO UPDATE SET response = EXCLUDED.response;
END;
$fn$;

SELECT create_distributed_function('register_for_event(bigint,bigint,invite_resp)', 'p_event_id', 'event_responses');

SET client_min_messages TO DEBUG;
CALL register_for_event(16, 1, 'yes');
DEBUG:  distribution argument value must be a constant
DEBUG:  Creating router plan
LOG:  issuing BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED;SELECT assign_distributed_transaction_id(0, 21, '2019-10-07 15:57:24.951449+02');
DETAIL:  on server marco@localhost:9700
LOG:  issuing INSERT INTO public.event_responses_102054 AS citus_table_alias (event_id, user_id, response) VALUES ($1, $2, $3) ON CONFLICT(event_id, user_id) DO UPDATE SET response = excluded.response
DETAIL:  on server marco@localhost:9700
LOG:  issuing COMMIT
DETAIL:  on server marco@localhost:9700
CALL
```

We should add a `strip_implict_coercions` call around `partitionValue` to remove the implicit FuncExpr (cast) that casts int to bigint.

serprex:
Fixes #3085 

Function call delegation not impacted since planner handles this for us. Still included test to verify that planner does so

### DIFF
--- a/src/backend/distributed/commands/call.c
+++ b/src/backend/distributed/commands/call.c
@@ -71,6 +71,7 @@ CallFuncExprRemotely(CallStmt *callStmt, DistObjectCacheEntry *procedure,
 					 FuncExpr *funcExpr, DestReceiver *dest)
 {
 	Oid colocatedRelationId = InvalidOid;
+	Node *partitionValueNode = NULL;
 	Const *partitionValue = NULL;
 	Datum partitionValueDatum = 0;
 	ShardInterval *shardInterval = NULL;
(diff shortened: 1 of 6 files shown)


src/backend/distributed/commands/call.c
```diff
@@ -118,7 +118,9 @@ CallFuncExprRemotely(CallStmt *callStmt, DistObjectCacheEntry *procedure,
 		return false;
 	}
 
-	partitionValue = (Const *) list_nth(funcExpr->args, procedure->distributionArgIndex);
+	partitionValue = (Const *) strip_implicit_coercions(list_nth(funcExpr->args,
+																 procedure->
+																 distributionArgIndex));
```
marcocitus: would be nice to split this into two lines

PR #3097, opened 2019-10-09
isolation_multiuser_locking

serprex:
Introduce a test which checks that locks are only acquired when a user has necessary permissions
Currently tests REINDEX, CREATE INDEX, TRUNCATE


pykello:
Forgot to update the `isolation_schedule`?

PR #3099, opened 2019-10-10
create_distributed_function: accept aggregates

serprex:
DESCRIPTION: allow distributing aggregate definitions to workers

This only distributes aggregates to worker nodes. It does not implement support for aggregates in distributed queries, which is what #2921 is for

It may be correct to not allow distribution argument parameters to be supplied for aggregates

### DIFF
--- a/src/backend/distributed/commands/function.c
+++ b/src/backend/distributed/commands/function.c
@@ -24,10 +24,12 @@
 #endif
 #include "access/htup_details.h"
 #include "access/xact.h"
+#include "catalog/pg_aggregate.h"
 #include "catalog/namespace.h"
 #include "catalog/pg_proc.h"
 #include "catalog/pg_type.h"
 #include "commands/extension.h"
+#include "distributed/citus_ruleutils.h"
 #include "distributed/colocation_utils.h"
 #include "distributed/commands.h"
 #include "distributed/commands/utility_hook.h"
@@ -40,7 +42,9 @@
 #include "distributed/metadata_sync.h"
 #include "distributed/multi_executor.h"
 #include "distributed/relation_access_tracking.h"
+#include "distributed/worker_create_or_replace.h"
 #include "distributed/worker_transaction.h"
+#include "nodes/makefuncs.h"
 #include "parser/parse_coerce.h"
 #include "parser/parse_type.h"
 #include "storage/lmgr.h"
@@ -49,12 +53,13 @@
 #include "utils/fmgrprotos.h"
 #include "utils/lsyscache.h"
 #include "utils/syscache.h"
+#include "utils/regproc.h"
 
 #define argumentStartsWith(arg, prefix) \
 	(strncmp(arg, prefix, strlen(prefix)) == 0)
 
 /* forward declaration for helper functions*/
-static char * GetFunctionDDLCommand(const RegProcedure funcOid);
+static char * GetAggregateDDLCommand(const RegProcedure funcOid, bool useCreateOrReplace);
 static char * GetFunctionAlterOwnerCommand(const RegProcedure funcOid);
 static int GetDistributionArgIndex(Oid functionOid, char *distributionArgumentName,
 								   Oid *distributionArgumentOid);
@@ -75,12 +80,10 @@ static ObjectAddress * FunctionToObjectAddress(ObjectType objectType,
 											   bool missing_ok);
 static void ErrorIfUnsupportedAlterFunctionStmt(AlterFunctionStmt *stmt);
 static void ErrorIfFunctionDependsOnExtension(const ObjectAddress *functionAddress);
+static char * quote_qualified_func_name(Oid funcOid);
 
 
 PG_FUNCTION_INFO_V1(create_distributed_function);
-
-#define AssertIsFunctionOrProcedure(objtype) \
-	Assert((objtype) == OBJECT_FUNCTION || (objtype) == OBJECT_PROCEDURE)
 
 
 /*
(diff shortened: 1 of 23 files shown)


src/backend/distributed/deparser/deparse_function_stmts.c
```diff
@@ -90,14 +90,17 @@ AppendAlterFunctionStmt(StringInfo buf, AlterFunctionStmt *stmt)
 	{
 		appendStringInfo(buf, "ALTER FUNCTION ");
 	}
-	else
+	else if (stmt->objtype == OBJECT_PROCEDURE)
```
thanodnl: Given the third addition I am inclined to make this a switch statement now.

src/backend/distributed/commands/function.c
```diff
@@ -537,27 +542,33 @@ GetFunctionDDLCommand(const RegProcedure funcOid)
...
-	/*
-	 * Set search_path to NIL so that all objects outside of pg_catalog will be
-	 * schema-prefixed. pg_catalog will be added automatically when we call
-	 * PushOverrideSearchPath(), since we set addCatalog to true;
-	 */
-	overridePath = GetOverrideSearchPath(CurrentMemoryContext);
-	overridePath->schemas = NIL;
-	overridePath->addCatalog = true;
-	PushOverrideSearchPath(overridePath);
+	createFunctionSQL = GetAggregateDDLCommand(funcOid);
```
thanodnl: As noted below;

This function will be more readable if we use `get_func_prokind` to get the function type and differentiate between calling `GetAggregateDDLCommand` and the body below. Instead of using the `NULL` output here to fallback to the other behaviour.

src/backend/distributed/commands/function.c
```diff
@@ -628,14 +646,395 @@ GetFunctionAlterOwnerCommand(const RegProcedure funcOid)
...
 	return alterCommand->data;
 }
 
 
+/*
+ * GetAggregateDDLCommand returns a string for creating an aggregate.
+ * Returns NULL if funcOid is not an aggregate.
+ */
+static char *
+GetAggregateDDLCommand(const RegProcedure funcOid)
```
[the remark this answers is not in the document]
serprex: Mostly created by hand

src/backend/distributed/commands/function.c
```diff
@@ -628,14 +646,395 @@ GetFunctionAlterOwnerCommand(const RegProcedure funcOid)
...
+		initStringInfo(&resultbuf);
+
+		appendStringInfoChar(&delim, '$');
+		while (strstr(buf.data, delim.data) != NULL)
+		{
+			appendStringInfoChar(&delim, 'z');
+		}
+
+		appendStringInfo(&resultbuf,
+						 "DO %s$BEGIN %s;EXCEPTION WHEN duplicate_function THEN NULL;END%s$",
```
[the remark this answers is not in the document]
serprex: It doesn't seem like much work. The main adjustment would be that worker_create_or_replace_object should know the difference between `CREATE` & `CREATE OR REPLACE` so that it can skip rename in the latter case

metdos:
@serprex, @thanodnl to merge this, what do we need?

serprex:
@metdos this is essentially ready to merge. Merging #3136 into this branch first would be nice tho

metdos:
> @metdos this is essentially ready to merge. Merging #3136 into this branch first would be nice tho

great! then whenever #3136 is ready, you can merge both

metdos:
I see #3136 is merged, shall we also merge this one?

src/backend/distributed/deparser/deparse_function_stmts.c
```diff
@@ -78,6 +80,37 @@ DeparseAlterFunctionStmt(AlterFunctionStmt *stmt)
...
+		{
+			return "PROCEDURE";
+		}
+
+		case OBJECT_AGGREGATE:
+		{
+			return "AGGREGATE";
+		}
+
+		default:
```
thanodnl: missing `{}` for the body

src/backend/distributed/worker/worker_create_or_replace.c
```diff
@@ -68,18 +84,15 @@ worker_create_or_replace_object(PG_FUNCTION_ARGS)
 	address = GetObjectAddressFromParseTree(parseTree, true);
 	if (ObjectExists(address))
 	{
-		Node *localCreateStmt = NULL;
-		const char *localSqlStatement = NULL;
 		char *newName = NULL;
-		RenameStmt *renameStmt = NULL;
 		const char *sqlRenameStmt = NULL;
+		RenameStmt *renameStmt;
```
thanodnl: `NULL` initialise

src/test/regress/expected/custom_aggregate_support.out
```diff
@@ -119,29 +119,29 @@ GROUP BY(1);
...
-   ->  Custom Scan (Citus Real-Time)
+   ->  Custom Scan (Citus Adaptive)
          Task Count: 4
          Tasks Shown: All
          ->  Task
                Node: host=localhost port=57637 dbname=regression
                ->  HashAggregate
                      Group Key: day
-                     ->  Seq Scan on daily_uniques_360289 daily_uniques
+                     ->  Seq Scan on daily_uniques_360453 daily_uniques
```
thanodnl: Would be a good moment to add `SET citus.next_shard_id TO ...;` at the top of the test to make shard identifiers stable.

src/backend/distributed/commands/function.c
```diff
@@ -549,38 +560,39 @@ UpdateFunctionDistributionInfo(const ObjectAddress *distAddress,
  * GetFunctionDDLCommand returns the complete "CREATE OR REPLACE FUNCTION ..." statement for
  * the specified function followed by "ALTER FUNCTION .. SET OWNER ..".
  */
-static char *
-GetFunctionDDLCommand(const RegProcedure funcOid)
+char *
+GetFunctionDDLCommand(const RegProcedure funcOid, bool wrapWithCreateOrReplace)
```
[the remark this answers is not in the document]
serprex: Since we use string comparisons for equality, the test case relies on worker_create_or_replace comparing to a CREATE, not a CREATE OR REPLACE

PR #3102, opened 2019-10-11
remove pg10 from pg upgrade tests

SaitTalhaNisanci:
DESCRIPTION: remove pg10 from pg upgrade tests


serprex:
The followup here is to remove building pg10 from build step in circleci

JelteF:
@serprex that should already be done now

Issue #2046, opened 2018-03-08
Improve view DML support


jasonmp85:
While Citus has greatly improved its support for views referencing distributed tables, users' ability to issue modifications against these views remains essentially non-existant. The relevant places in our planner code _do_ detect the use of a "relation […] [that is] not distributed", so it's not like this is an unhandled case; it's just unimplemented.

# Simple _Updatable Views_

For _simple_ views (essentially those defined using non-repeated direct references to columns of a single table), PostgreSQL has included `INSERT`/`UPDATE`/`DELETE` support since [PostgreSQL 9.3](https://www.postgresql.org/docs/9.3/static/sql-createview.html#SQL-CREATEVIEW-UPDATABLE-VIEWS). For instance:

```sql
CREATE TABLE employees (id integer, name text, salary integer);
-- CREATE TABLE

INSERT INTO employees VALUES (1, 'Bob Maiden', 123456789);
-- INSERT 0 1

CREATE VIEW employee_directory AS SELECT id, name FROM employees;
-- CREATE VIEW

UPDATE employee_directory SET name='Bob Married' WHERE id = 1;
-- UPDATE 1

SELECT * FROM employees;
-- ┌────┬─────────────┬───────────┐
-- │ id │    name     │  salary   │
-- ├────┼─────────────┼───────────┤
-- │  1 │ Bob Married │ 123456789 │
-- └────┴─────────────┴───────────┘
-- (1 row)
```

But let's try distributing this table in Citus:

```sql
SELECT create_distributed_table('employees', 'id');
-- NOTICE:  00000: Copying data from local table...
-- LOCATION:  CopyLocalDataIntoShards, create_distributed_table.c:1257
-- ┌──────────────────────────┐
-- │ create_distributed_table │
-- ├──────────────────────────┤
-- │                          │
-- └──────────────────────────┘
-- (1 row)

UPDATE employee_directory SET name='Bob Married' WHERE id = 1;
-- ERROR:  XX000: relation employee_directory is not distributed
-- LOCATION:  DistributedTableCacheEntry, metadata_cache.c:711
```

So step one is supporting the plain "updatable views" supported by PostgreSQL since 9.3. I used an `UPDATE` as my example, but the same error occurs even for `COPY`.

# Trigger Workarounds

As in pre-9.3 PostgreSQL, users can often work around Citus' VIEW limitations using a trigger function. For the above case, we could do something like this:

```sql
[...]

PR #3104, opened 2019-10-14
Fix view is not distributed error when view is used in modify statements

### LINKED ISSUE #2374: Views don't get recognized in UPDATE statements that can use router executor
**Schema**
```
CREATE TABLE large (id int, tenant_id int);
CREATE TABLE small (id int, tenant_id int);

SELECT create_distributed_table('large','tenant_id');
SELECT create_distributed_table('small','tenant_id');

CREATE VIEW small_view AS SELECT * from small;
```
**Below query (router executable) errors out:**
```
UPDATE large SET id=1 FROM small_view WHERE small_view.id=large.id and small_view.tenant_id=1 and large.tenant_id=1;
ERROR:  relation small_view is not distributed
```
**Below query (uses distributed planning and real-time executor) passes:**
```
UPDATE large SET id=1 FROM small_view WHERE small_view.id=large.id;
```

**A simple workaround for the router executable query is to use a CTE:**
```
explain WITH small_view AS (SELECT * FROM small_view where tenant_id=1) 
UPDATE large SET id=1 FROM small_view WHERE small_view.id=large.id and 
small_view.tenant_id=1 and large.tenant_id=1;
```

onurctirtir:
DESCRIPTION: Fixes a bug when view is used in modify statements.

Fixes #2374

### DIFF
--- a/src/backend/distributed/planner/multi_router_planner.c
+++ b/src/backend/distributed/planner/multi_router_planner.c
@@ -132,6 +132,7 @@ static bool TargetEntryChangesValue(TargetEntry *targetEntry, Var *column,
 static Job * RouterInsertJob(Query *originalQuery, Query *query,
 							 DeferredErrorMessage **planningError);
 static void ErrorIfNoShardsExist(DistTableCacheEntry *cacheEntry);
+static DeferredErrorMessage * DeferErrorIfModifyView(Query *queryTree);
 static bool CanShardPrune(Oid distributedTableId, Query *query);
 static Job * CreateJob(Query *query);
 static Task * CreateTask(TaskType taskType);
@@ -567,6 +568,12 @@ ModifyQuerySupported(Query *queryTree, Query *originalQuery, bool multiShardQuer
 	uint32 queryTableCount = 0;
 	CmdType commandType = queryTree->commandType;
 
+	deferredError = DeferErrorIfModifyView(queryTree);
+	if (deferredError != NULL)
+	{
+		return deferredError;
+	}
+
 	/*
 	 * Here, we check if a recursively planned query tries to modify
 	 * rows based on the ctid column. This is a bad idea because ctid of
@@ -656,29 +663,40 @@ ModifyQuerySupported(Query *queryTree, Query *originalQuery, bool multiShardQuer
 
 		if (rangeTableEntry->rtekind == RTE_RELATION)
 		{
-			Oid relationId = rangeTableEntry->relid;
-
-			if (!IsDistributedTable(relationId))
-			{
-				StringInfo errorMessage = makeStringInfo();
-				char *relationName = get_rel_name(rangeTableEntry->relid);
-
-				appendStringInfo(errorMessage, "relation %s is not distributed",
-								 relationName);
-
-				return DeferredError(ERRCODE_FEATURE_NOT_SUPPORTED,
-									 errorMessage->data, NULL, NULL);
-			}
-
-			queryTableCount++;
-
 			/* we do not expect to see a view in modify query */
 			if (rangeTableEntry->relkind == RELKIND_VIEW)
 			{
+				/*
+				 * we already check if modify is run on a view in DeferErrorIfModifyView
+				 * function call. In addition, since Postgres replaced views in FROM
+				 * clause with subqueries, encountering with a view should not be a problem here.
+				 */
+			}
+			else if (rangeTableEntry->relkind == RELKIND_MATVIEW)
+			{
 				return DeferredError(ERRCODE_FEATURE_NOT_SUPPORTED,
-									 "cannot modify views over distributed tables",
+									 "materialized views in modify queries are not supported",
 									 NULL, NULL);
 			}
+			/* for other kinds of relations, check if its distributed */
+			else
+			{
+				Oid relationId = rangeTableEntry->relid;
+
+				if (!IsDistributedTable(relationId))
+				{
+					StringInfo errorMessage = makeStringInfo();
+					char *relationName = get_rel_name(rangeTableEntry->relid);
+
+					appendStringInfo(errorMessage, "relation %s is not distributed",
+									 relationName);
+
+					return DeferredError(ERRCODE_FEATURE_NOT_SUPPORTED,
+										 errorMessage->data, NULL, NULL);
+				}
+			}
+
+			queryTableCount++;
 		}
 		else if (rangeTableEntry->rtekind == RTE_VALUES
 #if PG_VERSION_NUM >= 120000
@@ -889,6 +907,39 @@ ModifyQuerySupported(Query *queryTree, Query *originalQuery, bool multiShardQuer
 	if (deferredError != NULL)
 	{
 		return deferredError;
+	}
+
+	return NULL;
+}
+
+
+/*
+ * Modify statements on simple updetable views are not supported yet.
+ * Actually, we need the original query (the query before postgres
+ * pg_rewrite_query) to detect if the view sitting in rtable is to
+ * be updated or just to be used in FROM clause.
+ * Hence, tracing the postgres source code, we deduced that postgres
+ * puts the relation to be modified to the first entry of rtable.
+ * If first element of the range table list is a simple updatable
+ * view and this view is not coming from FROM clause (inFromCl = False),
+ * then update is run "on" that view.
+ */
+static DeferredErrorMessage *
+DeferErrorIfModifyView(Query *queryTree)
+{
+	if (queryTree->rtable != NIL)
+	{
+		RangeTblEntry *firstRangeTableElement = (RangeTblEntry *) linitial(
+			queryTree->rtable);
+
+		if (firstRangeTableElement->rtekind == RTE_RELATION &&
+			firstRangeTableElement->relkind == RELKIND_VIEW &&
+			firstRangeTableElement->inFromCl == false)
+		{
+			return DeferredError(ERRCODE_FEATURE_NOT_SUPPORTED,
+								 "cannot modify views over distributed tables", NULL,
+								 NULL);
+		}
 	}
 
 	return NULL;
--- a/src/test/regress/expected/materialized_view.out
+++ b/src/test/regress/expected/materialized_view.out
@@ -200,3 +200,125 @@ SELECT count(*) FROM materialized_view;
 DROP MATERIALIZED VIEW materialized_view;
 DROP SCHEMA materialized_view CASCADE;
 NOTICE:  drop cascades to view air_shipped_lineitems
+-- modify statements on/with views
+-- create two tables and a view
+CREATE TABLE large (id int, tenant_id int);
+CREATE TABLE small (id int, tenant_id int);
+SELECT create_distributed_table('large','tenant_id');
+ create_distributed_table 
+--------------------------
+ 
+(1 row)
+
+SELECT create_distributed_table('small','tenant_id');
+ create_distributed_table 
+--------------------------
+ 
+(1 row)
+
+\copy small FROM STDIN DELIMITER ','
+CREATE MATERIALIZED VIEW small_view AS SELECT * from small where id < 100;
+\copy large FROM STDIN DELIMITER ','
+-- running any kind of modify statements "on" materialized views is not supported by postgres
+UPDATE small_view SET id = 1;
+ERROR:  cannot change materialized view "small_view"
+-- for now, using materialized views in modify statements' FROM / WHERE clauses is not supported
+UPDATE large SET id=20 FROM small_view WHERE small_view.id=large.id;
+ERROR:  materialized views in modify queries are not supported
+SELECT * FROM large ORDER BY 1, 2;
+ id | tenant_id 
+----+-----------
+  1 |         2
+  2 |         3
+  5 |         4
+  6 |         5
+(4 rows)
+
+-- test on a router executable update statement, this will also fail
+UPDATE large SET id=28 FROM small_view WHERE small_view.id=large.id and small_view.tenant_id=2 and large.tenant_id=2;
+ERROR:  materialized views in modify queries are not supported
+SELECT * FROM large ORDER BY 1, 2;
+ id | tenant_id 
+----+-----------
+  1 |         2
+  2 |         3
+  5 |         4
+  6 |         5
+(4 rows)
+
+-- delete statement on large with subquery, this should succeed
+DELETE FROM large WHERE tenant_id in (SELECT tenant_id FROM small_view);
+SELECT * FROM large ORDER BY 1, 2;
+ id | tenant_id 
+----+-----------
+  6 |         5
+(1 row)
+
+-- INSERT INTO views is already not supported by PostgreSQL
+INSERT INTO small_view VALUES(3, 3);
+ERROR:  cannot change materialized view "small_view"
+DROP TABLE small CASCADE;
+NOTICE:  drop cascades to materialized view small_view
+DROP TABLE large;
+-- now, run the same modify statement tests on a partitioned table
+CREATE TABLE small (id int, tenant_id int);
+CREATE TABLE large_partitioned (id int, tenant_id int) partition by range(tenant_id);
+CREATE TABLE large_partitioned_p1 PARTITION OF large_partitioned FOR VALUES FROM (1) TO (10);
+CREATE TABLE large_partitioned_p2 PARTITION OF large_partitioned FOR VALUES FROM (10) TO (20);
+CREATE TABLE large_partitioned_p3 PARTITION OF large_partitioned FOR VALUES FROM (20) TO (100);
+SELECT create_distributed_table('large_partitioned','tenant_id');
+ create_distributed_table 
+--------------------------
+ 
+(1 row)
+
+SELECT create_distributed_table('small','tenant_id');
+ create_distributed_table 
+--------------------------
+ 
+(1 row)
+
+\copy small FROM STDIN DELIMITER ','
+CREATE MATERIALIZED VIEW small_view AS SELECT * from small where id < 100;
+\copy large_partitioned FROM STDIN DELIMITER ','
+-- running modify statements "on" views is still not supported, hence below two statements will fail
+UPDATE small_view SET id = 1;
+ERROR:  cannot change materialized view "small_view"
+DELETE FROM small_view;
+ERROR:  cannot change materialized view "small_view"
+-- using mat. view in modify statements' FROM / WHERE clauses is not valid yet
+UPDATE large_partitioned SET id=20 FROM small_view WHERE small_view.id=large_partitioned.id;
+ERROR:  materialized views in modify queries are not supported
+SELECT * FROM large_partitioned ORDER BY 1, 2;
+ id | tenant_id 
+----+-----------
+  1 |         2
+  2 |         3
+  5 |         4
+  6 |         5
+ 26 |        32
+ 29 |        15
+ 60 |        51
+(7 rows)
+
+-- delete statement on large_partitioned
+DELETE FROM large_partitioned WHERE id in (SELECT id FROM small_view);
+SELECT * FROM large_partitioned ORDER BY 1, 2;
+ id | tenant_id 
+----+-----------
+  2 |         3
+  5 |         4
+ 26 |        32
+ 29 |        15
+ 60 |        51
+(5 rows)
+
+-- we should still have identical rows for next test statement, then insert new rows to both tables
+INSERT INTO large_partitioned VALUES(14, 14);
+INSERT INTO small VALUES(14, 14);
+-- delete statement with CTE
+WITH all_small_view_ids AS (SELECT id FROM small_view)
+DELETE FROM large_partitioned WHERE id in (SELECT * FROM all_small_view_ids);
+DROP TABLE large_partitioned;
+DROP TABLE small CASCADE;
+NOTICE:  drop cascades to materialized view small_view
--- a/src/test/regress/expected/multi_view.out
+++ b/src/test/regress/expected/multi_view.out
@@ -883,3 +883,404 @@ DROP VIEW recent_selected_users;
 DROP VIEW selected_users;
 DROP VIEW recent_events;
 DROP VIEW recent_users;
+-- modify statements on/with views
+-- create two tables and a view
+CREATE TABLE large (id int, tenant_id int);
+-- constraint id to be unique for "insert into on conflict" test
+CREATE TABLE small (id int, tenant_id int, unique(tenant_id));
+SELECT create_distributed_table('large','tenant_id');
+ create_distributed_table 
+--------------------------
+ 
+(1 row)
+
+SELECT create_distributed_table('small','tenant_id');
+ create_distributed_table 
+--------------------------
+ 
+(1 row)
+
+CREATE VIEW small_view AS SELECT * from small where id < 100;
+\copy small FROM STDIN DELIMITER ','
+\copy large FROM STDIN DELIMITER ','
+-- running modify statements "on" views is still not supported, hence below two statements will fail
+UPDATE small_view SET id = 1;
+ERROR:  cannot modify views over distributed tables
+DELETE FROM small_view;
+ERROR:  cannot modify views over distributed tables
+INSERT INTO small_view VALUES(8, 5) ON CONFLICT(tenant_id) DO UPDATE SET tenant_id=99;
+ERROR:  cannot modify views over distributed tables
+-- using views in modify statements' FROM / WHERE clauses is still valid 
+UPDATE large SET id=20 FROM small_view WHERE small_view.id=large.id;
+SELECT * FROM large order by 1, 2;
+ id | tenant_id 
+----+-----------
+  2 |         3
+  5 |         4
+ 20 |         2
+ 20 |         5
+(4 rows)
+
+-- we should still have identical rows for next test statements, then insert new rows to both tables
+INSERT INTO large VALUES(14, 14);
+INSERT INTO small VALUES(14, 14);
+-- using views in subqueries within modify statements is still valid 
+UPDATE large SET id=23 FROM (SELECT *, id*2 from small_view ORDER BY 1,2 LIMIT 5) as small_view WHERE small_view.id=large.id;
+SELECT * FROM large order by 1, 2;
+ id | tenant_id 
+----+-----------
+  2 |         3
+  5 |         4
+ 20 |         2
+ 20 |         5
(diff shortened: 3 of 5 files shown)


serprex:
Don't reference issue numbers in commit messages. It makes for weird things on enterprise repo

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -647,7 +647,7 @@ ModifyQuerySupported(Query *queryTree, Query *originalQuery, bool multiShardQuer
 	}
 
 	/* extract range table entries */
-	ExtractRangeTableEntryWalker((Node *) queryTree, &rangeTableList);
+	ExtractRangeTableRelationWalker((Node *) queryTree, &rangeTableList);
```
marcocitus: Looks good, but needs tests.

onurctirtir:
tests will be added soon

src/test/regress/expected/multi_view.out
```diff
@@ -251,6 +251,75 @@ SELECT * FROM mode_counts WHERE l_shipmode = 'AIR' ORDER BY 2 DESC, 1 LIMIT 10;
...
+--------------------------
+ 
+(1 row)
+
+CREATE VIEW small_view AS SELECT * from small where id < 100;
+\copy small FROM STDIN DELIMITER ','
+\copy large FROM STDIN DELIMITER ','
+-- running modify statements "on" views still is not supported, hence below two statements will fail
+UPDATE small_view SET id = 1;
+ERROR:  table name "small" specified more than once
```
onderkalaci: This error is not expected. We get the error from the from the workers because we send the following query:
```
 DELETE FROM public.small_102330 small USING public.small_102330 small WHERE (small.id OPERATOR(pg_catalog.<) 100)

```

I think the issue is that we're now calling `ExtractRangeTableRelationWalker `, which only pulls the relations. But, in  `ModifyQuerySupported` , we check for `if (rangeTableEntry->relkind == RELKIND_VIEW)`, which is not in the list.

We should fix this

src/test/regress/expected/multi_view.out
```diff
@@ -251,6 +251,75 @@ SELECT * FROM mode_counts WHERE l_shipmode = 'AIR' ORDER BY 2 DESC, 1 LIMIT 10;
...
+ 
+(1 row)
+
+SELECT create_distributed_table('small','tenant_id');
+ create_distributed_table 
+--------------------------
+ 
+(1 row)
+
+CREATE VIEW small_view AS SELECT * from small where id < 100;
```
onderkalaci: Can you also add similar tests with materialized views?

src/test/regress/expected/multi_view.out
onderkalaci: Note that similarly we're losing all the coverage for the `else` part of `if (rangeTableEntry->rtekind == RTE_RELATION)` in [here](https://github.com/citusdata/citus/blob/870b743b938121b52e0967e14bf41347f1926d54/src/backend/distributed/planner/multi_router_planner.c#L691-L751)

marcocitus: Would be good to improve this error message. We may be able to detect this case in originalQuery.

onurctirtir: I attempted to fix the bad error message issue in e844b596e45fed88fcef37fd64540adddaadbcb3

src/test/regress/expected/multi_view.out
onurctirtir: I will add them asap

src/test/regress/expected/multi_view.out
onurctirtir: Update: I guess I found a better way to solve it, I will try on it soon

onurctirtir: I tried to check if `resultRelation`th entry  of `rtable`  field of  `originalQuery/ ModifyQuerySupported` is a view or not. However, I guess it does not correspond to the query before rewrite process of postgress. This is because, `ModifyQuerySupported` is called via `distributed_planner` hook,  which is run at function call at `exec_simple_query:1143`. However, postgres's query rewrite is run at `exec_simple_query:1140`. After query rewrite, postgres appends the relation that the view refers to into the `rtable` and sets `resultRelation` to the index of that referred relation.

It would be better if we had the original query (the query itself before the rewrite) hence we can deduce if the modify statement is run on a view or not.

onurctirtir: For `update small_view set id=1` statement:

Before postgres rewrite, we have:
```
   {QUERY 
   :commandType 2 
   :querySource 0 
   :canSetTag true 
   :utilityStmt <> 
   :resultRelation 1 
   :hasAggs false 
   :hasWindowFuncs false 
   :hasTargetSRFs false 
   :hasSubLinks false 
   :hasDistinctOn false 
   :hasRecursive false 
   :hasModifyingCTE false 
   :hasForUpdate false 
   :hasRowSecurity false 
   :cteList <> 
   :rtable (
      {RTE 
      :alias <> 
      :eref 
         {ALIAS 
         :aliasname small_view 
         :colnames ("id" "tenant_id")
         }
      :rtekind 0 
      :relid 16774 
      :relkind v 
      :rellockmode 3 
      :tablesample <> 
      :lateral false 
      :inh true 
      :inFromCl false 
      :requiredPerms 4 
      :checkAsUser 0 
      :selectedCols (b)
      :insertedCols (b)
      :updatedCols (b 8)
      :extraUpdatedCols (b)
      :securityQuals <>
      }
   )
   :jointree 
      {FROMEXPR 
      :fromlist (
         {RANGETBLREF 
         :rtindex 1
         }
      )
      :quals <>
      }
   :targetList (
      {TARGETENTRY 
      :expr 
         {CONST 
         :consttype 23 
         :consttypmod -1 
         :constcollid 0 
         :constlen 4 
         :constbyval true 
         :constisnull false 
         :location 25 
         :constvalue 4 [ 1 0 0 0 0 0 0 0 ]
         }
      :resno 1 
      :resname id 
      :ressortgroupref 0 
      :resorigtbl 0 
      :resorigcol 0 
      :resjunk false
      }
   )
   :override 0 
   :onConflict <> 
   :returningList <> 
   :groupClause <> 
   :groupingSets <> 
   :havingQual <> 
   :windowClause <> 
   :distinctClause <> 
   :sortClause <> 
   :limitOffset <> 
   :limitCount <> 
   :rowMarks <> 
   :setOperations <> 
   :constraintDeps <> 
   :withCheckOptions <> 
   :stmt_location 0 
   :stmt_len 26
   }
```

.After postgres query rewrite:

`originalQuery/ModifyQuerySupported` is 

```
   {QUERY 
   :commandType 2 
   :querySource 0 
   :canSetTag true 
   :utilityStmt <> 
   :resultRelation 2 
   :hasAggs false 
   :hasWindowFuncs false 
   :hasTargetSRFs false 
   :hasSubLinks false 
   :hasDistinctOn false 
   :hasRecursive false 
   :hasModifyingCTE false 
   :hasForUpdate false 
   :hasRowSecurity false 
   :cteList <> 
   :rtable (
      {RTE 
      :alias <> 
      :eref 
         {ALIAS 
         :aliasname small_view 
         :colnames ("id" "tenant_id")
         }
      :rtekind 0 
      :relid 16774 
      :relkind v 
      :rellockmode 3 
      :tablesample <> 
      :lateral false 
      :inh true 
      :inFromCl false 
      :requiredPerms 4 
      :checkAsUser 0 
      :selectedCols (b)
      :insertedCols (b)
      :updatedCols (b 8)
      :extraUpdatedCols (b)
      :securityQuals <>
      }
      {RTE 
      :alias <> 
      :eref 
         {ALIAS 
         :aliasname small 
         :colnames ("id" "tenant_id")
         }
      :rtekind 0 
      :relid 16769 
      :relkind r 
      :rellockmode 3 
      :tablesample <> 
      :lateral false 
      :inh true 
      :inFromCl true 
      :requiredPerms 4 
      :checkAsUser 10 
      :selectedCols (b 8 9)
      :insertedCols (b)
      :updatedCols (b 8)
      :extraUpdatedCols (b)
      :securityQuals <>
      }
   )
   :jointree 
      {FROMEXPR 
      :fromlist (
         {RANGETBLREF 
         :rtindex 2
         }
      )
      :quals <>
      }
   :targetList (
      {TARGETENTRY 
      :expr 
         {CONST 
         :consttype 23 
         :consttypmod -1 
         :constcollid 0 
         :constlen 4 
         :constbyval true 
         :constisnull false 
         :location 25 
         :constvalue 4 [ 1 0 0 0 0 0 0 0 ]
         }
      :resno 1 
      :resname id 
      :ressortgroupref 0 
      :resorigtbl 0 
      :resorigcol 0 
      :resjunk false
      }
   )
   :override 0 
   :onConflict <> 
   :returningList <> 
   :groupClause <> 
   :groupingSets <> 
   :havingQual <> 
   :windowClause <> 
   :distinctClause <> 
   :sortClause <> 
   :limitOffset <> 
   :limitCount <> 
   :rowMarks <> 
   :setOperations <> 
   :constraintDeps <> 
   :withCheckOptions <> 
   :stmt_location 0 
   :stmt_len 26
   }
``` 
and `queryTree/ModifyQuerySupported` is 

```
   {QUERY 
   :commandType 2 
   :querySource 0 
   :canSetTag true 
   :utilityStmt <> 
   :resultRelation 2 
   :hasAggs false 
   :hasWindowFuncs false 
   :hasTargetSRFs false 
   :hasSubLinks false 
   :hasDistinctOn false 
   :hasRecursive false 
   :hasModifyingCTE false 
   :hasForUpdate false 
   :hasRowSecurity false 
   :cteList <> 
   :rtable (
      {RTE 
      :alias <> 
      :eref 
         {ALIAS 
         :aliasname small_view 
         :colnames ("id" "tenant_id")
         }
      :rtekind 0 
      :relid 16774 
      :relkind v 
      :rellockmode 3 
      :tablesample <> 
      :lateral false 
      :inh false 
      :inFromCl false 
      :requiredPerms 4 
      :checkAsUser 0 
      :selectedCols (b)
      :insertedCols (b)
      :updatedCols (b 8)
      :extraUpdatedCols (b)
      :securityQuals <>
      }
      {RTE 
      :alias <> 
      :eref 
         {ALIAS 
         :aliasname small 
         :colnames ("id" "tenant_id")
         }
      :rtekind 0 
      :relid 16769 
      :relkind r 
      :rellockmode 3 
      :tablesample <> 
      :lateral false 
      :inh false 
      :inFromCl true 
      :requiredPerms 4 
      :checkAsUser 10 
      :selectedCols (b 8 9)
      :insertedCols (b)
      :updatedCols (b 8)
      :extraUpdatedCols (b)
      :securityQuals <>
      }
   )
   :jointree 
      {FROMEXPR 
      :fromlist (
         {RANGETBLREF 
         :rtindex 2
         }
      )
      :quals <>
      }
   :targetList (
      {TARGETENTRY 
      :expr 
         {CONST 
         :consttype 23 
         :consttypmod -1 
         :constcollid 0 
         :constlen 4 
         :constbyval true 
         :constisnull false 
         :location 25 
         :constvalue 4 [ 1 0 0 0 0 0 0 0 ]
         }
      :resno 1 
      :resname id 
      :ressortgroupref 0 
      :resorigtbl 0 
      :resorigcol 0 
      :resjunk false
      }
      {TARGETENTRY 
      :expr 
         {VAR 
         :varno 2 
         :varattno -1 
         :vartype 27 
         :vartypmod -1 
         :varcollid 0 
         :varlevelsup 0 
         :varnoold 2 
         :varoattno -1 
         :location -1
         }
      :resno 2 
      :resname ctid 
      :ressortgroupref 0 
      :resorigtbl 0 
      :resorigcol 0 
      :resjunk true
      }
   )
   :override 0 
   :onConflict <> 
   :returningList <> 
   :groupClause <> 
   :groupingSets <> 
   :havingQual <> 
   :windowClause <> 
   :distinctClause <> 
   :sortClause <> 
   :limitOffset <> 
   :limitCount <> 
   :rowMarks <> 
   :setOperations <> 
   :constraintDeps <> 
   :withCheckOptions <> 
   :stmt_location 0 
   :stmt_len 26
   }
```

src/test/regress/expected/materialized_view.out
```diff
@@ -200,3 +200,74 @@ SELECT count(*) FROM materialized_view;
...
+CREATE MATERIALIZED VIEW small_view AS SELECT * from small where id < 100;
+\copy small FROM STDIN DELIMITER ','
+\copy large FROM STDIN DELIMITER ','
+-- running modify statements "on" views is still not supported, hence below two statements will fail
+UPDATE small_view SET id = 1;
+ERROR:  cannot change materialized view "small_view"
+DELETE FROM small_view;
+ERROR:  cannot change materialized view "small_view"
+-- using views in modify statements' FROM / WHERE clauses is stll valid 
+UPDATE large SET id=20 FROM small_view WHERE small_view.id=large.id; --rows with id=6 and id=1 now become id=20 in large table
```
onderkalaci: what happens if the view contains join of two tables and users updates the view? First, can you check what does postgres do?

src/test/regress/expected/multi_view.out
onderkalaci: Thanks for sharing this, so the summary is that after postgresql's rewrite (which happens before `distributed_planner` is called), the view is not on the `targetRelation`. 

What other options do we have:

1) Why do we prevent when target relation is a view? It turns out that the issue is on deparsing of the query. Can we fix that? We need to consider some extra things, so probably not easy.

2) The magic happens on postgresql's `rewriteTargetView` function. What I see is that, the function doesn't remove the view from the `rtable`, but the range table not used in `jointree`. Can we change the code such that we only error out if there is a `rtable` which is a view and it doesn't show-up in the `jointree`?  

3) This is probably easier than 2. So, Postgres only rewrites target views when the view is created based on a single table (see `view_query_is_auto_updatable`). So, can't we simply check if the `list_length(rtable) == 2` and `linitial(rtable) is a view` and `jointree doesn't have reference to 1` -- see `ExtractRangeTableIndexWalker`. So, basically, move the check for update/delete on the view from the loop, and have a separate check.

Well, 2 and 3 heavily depends on postgresql internals, but these seems very rare to change (as commented in Postgres code these are SQL standard checks). So, I'm OK with implementing 3 with some sufficient comments.

src/test/regress/expected/materialized_view.out
```diff
@@ -200,3 +200,74 @@ SELECT count(*) FROM materialized_view;
 DROP MATERIALIZED VIEW materialized_view;
 DROP SCHEMA materialized_view CASCADE;
 NOTICE:  drop cascades to view air_shipped_lineitems
+-- modify statements on/with views
```
onderkalaci: I think `rewriteTargetView` is a good resource to check what to test. Some examples:

- [ ] RETURNING
- [ ] ON CONFLICT DO UPDATE .. SET ..
- [ ] DELETE/UPDATE/INSERT to the view itself
- [ ]  Junk columns (`resjunk` in the target list)
- [ ] subqueries in the WHERE clause

can you make sure to add tests for the above as well?

src/test/regress/expected/multi_view.out
onurctirtir: Actually, I implemented a similar solution to 3 which relies in e844b596e45fed88fcef37fd64540adddaadbcb3. But, according to postgresql docs, it is legal for update to take many tables as argument    
`UPDATE [ ONLY ] table [ * ] [ [ AS ] alias ]`-. Then implementing 3 becomes a bit harder than we expected, if I am not wrong.

Having not much context on 1, when I observed the problem in https://github.com/citusdata/citus/pull/3104#discussion_r336479086, I initially thought that "can implementing view support be less complex than un-supporting it ?"

onderkalaci: I don't think you can update multiple tables with a single command in PostgreSQL. What you're referring is something else:

> table_name
The name (optionally schema-qualified) of the table to update. If ONLY is specified before the table name, matching rows are updated in the named table only. If ONLY is not specified, matching rows are also updated in any tables inheriting from the named table. Optionally, * can be specified after the table name to explicitly indicate that descendant tables are included.

well, its good to remember. We should also consider testing with partitioned tables which are inherited internally :) 


So, the commit you mention is similar to what I've in mind. Did you have any issues with that?

onurctirtir: Now it makes sense and I have no issues with that. So I will implement the solution 3. Then I will add the missing tests as well

onurctirtir:
To remind myself, I have the current todo's for testing the expected behavior in comments:

- [ ] https://github.com/citusdata/citus/pull/3104#discussion_r337009839
- [ ] https://github.com/citusdata/citus/pull/3104#discussion_r336984045
- [ ] https://github.com/citusdata/citus/pull/3104#discussion_r337003861

src/test/regress/expected/multi_view.out
onderkalaci: To summarize: 

We really couldn't find a robust way of detecting the `targetRelation` when it is a view. Because, the view rewriting is happening before Citus kicks in.

Our proposed solution (` if rtable[0] is a view AND view.inFromCl == false: throw error`) would be heuristic. Reading through the postgres code, I cannot see any counter examples. The target relation is always added to the `rtable` as the first item. Similarly, the relation that is replaced in place of the view is appended to the `rtable`, so it is the `second` item.

(The more intuitive approach that we considered was to trace back from the `targetRelation` to the view. But, that's also a heuristic and our proposal seems a lot simpler and looks like very similar to what this would do.)

The worst case scenario is that the query on the shards would fail, because of [this](https://github.com/citusdata/citus/issues/2046#issuecomment-465136620) explanation.

That said, happy to get any input here.

onurctirtir: The third solution we mentioned in https://github.com/citusdata/citus/pull/3104#discussion_r336999409 is implemented in commit 0847cce

onurctirtir: The third solution we mentioned in https://github.com/citusdata/citus/pull/3104#discussion_r336999409 is implemented in commit d72f17f

src/test/regress/expected/materialized_view.out
onurctirtir:
```sql
ERROR:  cannot update view "localtablejoin"
DETAIL:  Views that do not select from a single table or view are not automatically updatable.
HINT:  To enable updating the view, provide an INSTEAD OF UPDATE trigger or an unconditional ON UPDATE DO INSTEAD rule.
```

As I see, we even cannot update views that contain join of two local tables in postgresql automatically.

onurctirtir:
For the followers of this pr, I wanted to share another errornous case that I encountered yesterday. I was testing modify statements with partitioned tables and views (see https://github.com/citusdata/citus/pull/3104#discussion_r337009839) and I observed the following behaviour:

```sql
postgres=# CREATE TABLE small (id int, tenant_id int);
CREATE TABLE
postgres=# CREATE VIEW small_view AS SELECT * from small where id < 100;
CREATE VIEW
postgres=# CREATE TABLE large_partitioned (id int, tenant_id int) partition by range(tenant_id);
CREATE TABLE
postgres=# CREATE TABLE large_partitioned_p1 PARTITION OF large_partitioned FOR VALUES FROM (1) TO (10);
CREATE TABLE
postgres=# SELECT create_distributed_table('large_partitioned','tenant_id');
 create_distributed_table
--------------------------

(1 row)

postgres=# SELECT create_distributed_table('small','tenant_id');
 create_distributed_table
--------------------------

(1 row)

postgres=# insert into large_partitioned values (1, 9);
INSERT 0 1
-- while the below does not work
postgres=# UPDATE large_partitioned SET id=20 FROM small_view WHERE small_view.id=large_partitioned.id;
-- below does work
postgres=# UPDATE large_partitioned_p1 SET id=20 FROM small_view WHERE small_view.id=large_partitioned_p1.id;
UPDATE 1
```

src/test/regress/expected/materialized_view.out
onurctirtir: In commits 2b8e10a1c4087455878805a7528025e8eaacd0cf and f214815e3045e70c8f6dc6c6b401cabedbd14da2, I added tests for these items

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -656,29 +694,31 @@ ModifyQuerySupported(Query *queryTree, Query *originalQuery, bool multiShardQuer
 
 		if (rangeTableEntry->rtekind == RTE_RELATION)
 		{
-			Oid relationId = rangeTableEntry->relid;
-
-			if (!IsDistributedTable(relationId))
+			if (rangeTableEntry->relkind == RELKIND_RELATION)
```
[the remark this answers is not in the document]
onurctirtir: Yes, that makes sense to me

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -895,6 +904,37 @@ ModifyQuerySupported(Query *queryTree, Query *originalQuery, bool multiShardQuer
...
+ * Hence, tracing the postgres source code, we deduced that postgres
+ * puts the relation to be modified to the first entry of rtable.
+ * If first element of the range table list is a simple updatable
+ * view and this view is not coming from FROM clause (inFromCl = False),
+ * then update is run "on" that view.
+ */
+static DeferredErrorMessage *
+DeferErrorIfModifyView(Query *queryTree)
+{
+	RangeTblEntry *firstRangeTableElement = (RangeTblEntry *) lfirst(
```
onderkalaci: you should use `linitial()`, search Citus code, you'd see the examples of it

onurctirtir: Replaced `linitial `with `lfirst `in 7128dbe000fcd412d7c51fb17865496cea9322f9

src/test/regress/expected/materialized_view.out
```diff
@@ -200,3 +200,182 @@ SELECT count(*) FROM materialized_view;
...
+SELECT create_distributed_table('small','tenant_id');
+ create_distributed_table 
+--------------------------
+ 
+(1 row)
+
+\copy small FROM STDIN DELIMITER ','
+CREATE MATERIALIZED VIEW small_view AS SELECT * from small where id < 100;
+\copy large_partitioned FROM STDIN DELIMITER ','
+-- make "eye-inspection" easier by printing the initial tables, for readers of this test
```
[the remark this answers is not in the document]
onurctirtir: I removed those statements in commit 1c9fb101941d8b420663619d1704c8de4314a60c

src/test/regress/expected/multi_view.out
```diff
@@ -883,3 +883,261 @@ DROP VIEW recent_selected_users;
...
+
+-- running modify statements "on" views is still not supported, hence below two statements will fail
+UPDATE small_view SET id = 1;
+ERROR:  cannot modify views over distributed tables
+DELETE FROM small_view;
+ERROR:  cannot modify views over distributed tables
+INSERT INTO small_view VALUES(8, 5) ON CONFLICT(tenant_id) DO UPDATE SET tenant_id=99;
+ERROR:  cannot modify views over distributed tables
+-- using views in modify statements' FROM / WHERE clauses is still valid 
+UPDATE large SET id=20 FROM small_view WHERE small_view.id=large.id;
```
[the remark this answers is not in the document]
onurctirtir: I added missing tests for new review items in c44d0b9f68aca0502e68c4a8598ac46899b0d759

src/backend/distributed/planner/multi_router_planner.c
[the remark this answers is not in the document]
onurctirtir: I fixed the logic in commit 75d5943543b166a29d05ec22fe5ed7ee1a5fedc2

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -544,6 +545,37 @@ IsTidColumn(Node *node)
...
+ * pg_rewrite_query) to detect if the view sitting in rtable is to
+ * be updated or just to be used in FROM clause.
+ * Hence, tracing the postgres source code, we deduced that postgres
+ * puts the relation to be modified to the first entry of rtable.
+ * If first element of the range table list is a simple updatable
+ * view and this view is not coming from FROM clause (inFromCl = False),
+ * then update is run "on" that view.
+ */
+static DeferredErrorMessage *
+ErrorIfModifySimpleView(Query *queryTree)
```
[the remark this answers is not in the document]
onurctirtir: In commit eb1d7924e4945386dc33b5d891a079b1931baa6f, I applied necessary changes

src/test/regress/expected/multi_view.out
```diff
@@ -883,3 +883,404 @@ DROP VIEW recent_selected_users;
...
+  5 |         4
+ 20 |         2
+ 20 |         5
+(4 rows)
+
+-- we should still have identical rows for next test statements, then insert new rows to both tables
+INSERT INTO large VALUES(14, 14);
+INSERT INTO small VALUES(14, 14);
+-- using views in subqueries within modify statements is still valid 
+UPDATE large SET id=23 FROM (SELECT *, id*2 from small_view) as small_view WHERE small_view.id=large.id;
```
onderkalaci: This subquery is too simple. Postgres has a logic to pull this subquery into a join (e.g., get rid of the subquery, see `pull_up_subqueries` function in postgres source). To prevent postgres to pull up the subquery, can you add `ORDER BY ... LIMIT ..` to the subquery itself?

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -895,6 +909,37 @@ ModifyQuerySupported(Query *queryTree, Query *originalQuery, bool multiShardQuer
...
+ * Hence, tracing the postgres source code, we deduced that postgres
+ * puts the relation to be modified to the first entry of rtable.
+ * If first element of the range table list is a simple updatable
+ * view and this view is not coming from FROM clause (inFromCl = False),
+ * then update is run "on" that view.
+ */
+static DeferredErrorMessage *
+DeferErrorIfModifyView(Query *queryTree)
+{
+	RangeTblEntry *firstRangeTableElement = (RangeTblEntry *) linitial(
```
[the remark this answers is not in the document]
onurctirtir: Actually, it does not sound possible to me but I could not find any useful information in postgres code. It would be better to add a `queryTree->rtable == NIL` check to this function.

src/test/regress/expected/multi_view.out
onurctirtir: Just read the leading comment of `pull_up_subqueries`: `If the subquery has no special features like grouping/aggregation then we can merge it into the parent's jointree.`

I am adding these extra clauses now

PR #2918, opened 2019-09-02
Remove copyright years

SaitTalhaNisanci:
This PR removes copyright year.

The following command from the root is used for updating:
`grep -rli '2012-2015' * | xargs -i@ sed -i 's/2012-2015/2012-2019/g' @`
`grep -rli '2012-2016' * | xargs -i@ sed -i 's/2012-2016/2012-2019/g' @`
`grep -rli '2012-2017' * | xargs -i@ sed -i 's/2012-2017/2012-2019/g' @`
`grep -rli '2012-2018' * | xargs -i@ sed -i 's/2012-2018/2012-2019/g' @`
`grep -rli ' 2015, Ci' * | xargs -i@ sed -i 's/ 2015, Ci/ 2012-2019, Ci/g' @`
`grep -rli ' 2016, Ci' * | xargs -i@ sed -i 's/ 2016, Ci/ 2012-2019, Ci/g' @`
`grep -rli ' 2017, Ci' * | xargs -i@ sed -i 's/ 2017, Ci/ 2012-2019, Ci/g' @`
`grep -rli ' 2018, Ci' * | xargs -i@ sed -i 's/ 2018, Ci/ 2012-2019, Ci/g' @`
`grep -rli '2016-2016' * | xargs -i@ sed -i 's/2016-2016/2012-2019/g' @`
`grep -rli '2017-2017' * | xargs -i@ sed -i 's/2017-2017/2012-2019/g' @`

And finally:
`grep -rli '2012-2019, Citus Data, Inc.' * | xargs -i@ sed -i 's/2012-2019, Citus Data, Inc./Citus Data, Inc./g' @`

PR #3109, opened 2019-10-15
Refactor range table walkers

onurctirtir:
We currently have two widely used functions for getting the range table entries from a query: `ExtractRangeTableRelationWalker` and `ExtractRangeTableEntryWalker`. They are subtly different since the former excludes views (i.e. only returns relations), however their implementation is almost identical.

This PR refactors them to use the same underlying implementation.


marcocitus:
We can omit the DESCRIPTION for refactoring, since that automatically shows up in the changelog.

src/include/distributed/query_utils.h
```diff
@@ -0,0 +1,40 @@
...
+
+#include "postgres.h"
+#include "nodes/pg_list.h"
+
+/* Enum to define execution flow of ExtractRangeTableList */
+typedef enum ExtractMode
+{
+	EXT_RELATION_ENTRIES,
+	EXT_INDEX_ENTRIES,
+	EXT_ALL_ENTRIES
```
marcocitus: We try to avoid acronyms (We as in the developers of this codebase, not Microsoft ;). 

So maybe `EXTRACT_` instead of `EXT_`

src/include/distributed/query_utils.h
```diff
@@ -0,0 +1,40 @@
...
+} ExtractMode;
```
marcocitus: this name might be a little generic, maybe `ExtractRangeTableMode`

src/include/distributed/query_utils.h
```diff
@@ -0,0 +1,40 @@
...
+
+/* Struct to pass rtable list and execution flow flag to query_walker_tree */
+typedef struct ExtractParams
+{
+	List **rangeTableList;
+	ExtractMode walkerMode;
+} ExtractParams;
```
marcocitus: This type of struct is often referred to as a (walker) context in Postgres.

e.g. `ExtractRangeTableWalkerContext`

src/include/distributed/query_utils.h
```diff
@@ -0,0 +1,40 @@
+/*-------------------------------------------------------------------------
+ *
+ * query_utils.h
+ *
+ * Declarations for query-walker utility functions and related types.
+ *
+ * Copyright (c) 2019, Citus Data, Inc.
```
marcocitus: maybe omit the year https://github.com/citusdata/citus/pull/2918

src/backend/distributed/utils/query_utils.c
```diff
@@ -0,0 +1,133 @@
+/*-------------------------------------------------------------------------
+ *
+ * query_utils.c
+ *
+ * Query-walker utility functions to be used to construct a logical plan
+ * tree from the given query tree structure.
+ *
+ * Copyright (c) 2019, Citus Data, Inc.
+ *
+ * $Id$
```
marcocitus: this seems like some historical thing, can be omitted.

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1177,34 +1178,6 @@ HasComplexRangeTableType(Query *queryTree)
...
-	bool walkerResult = false;
-	if (node == NULL)
-	{
-		return false;
-	}
-
-	if (IsA(node, RangeTblRef))
-	{
-		int rangeTableIndex = ((RangeTblRef *) node)->rtindex;
-		(*rangeTableIndexList) = lappend_int(*rangeTableIndexList, rangeTableIndex);
```
marcocitus: We probably shouldn't try to merge this function in because it does something quite different (returns indexes and does not descend into query).

src/backend/distributed/utils/query_utils.c
```diff
@@ -0,0 +1,133 @@
...
+	else if (IsA(node, RangeTblRef) && walkerMode == EXT_INDEX_ENTRIES)
+	{
+		int rangeTableIndex = ((RangeTblRef *) node)->rtindex;
+		(*rangeTableRelationList) = lappend_int(*rangeTableRelationList, rangeTableIndex);
+	}
+	else if (IsA(node, Query) && walkerMode != EXT_INDEX_ENTRIES)
+	{
+		Query *query = (Query *) node;
+
+		if (walkerMode == EXT_ALL_ENTRIES || query->hasSubLinks || query->cteList ||
```
marcocitus: I don't think we should have the walkerMode comparison here. The if statement is an optimisation, because we know the query tree does not contain RTEs other than the ones in rangeTableList if these conditions hold. It would be nice to apply that optimisation in all cases.

src/backend/distributed/utils/query_utils.c
```diff
@@ -0,0 +1,133 @@
...
+		if (walkerMode == EXT_ALL_ENTRIES ||
+			(rangeTable->rtekind == RTE_RELATION && rangeTable->relkind != RELKIND_VIEW))
+		{
+			(*rangeTableRelationList) = lappend(*rangeTableRelationList, rangeTable);
+		}
+	}
+	else if (IsA(node, RangeTblRef) && walkerMode == EXT_INDEX_ENTRIES)
+	{
+		int rangeTableIndex = ((RangeTblRef *) node)->rtindex;
+		(*rangeTableRelationList) = lappend_int(*rangeTableRelationList, rangeTableIndex);
```
marcocitus: Per the other comment, this one should be omitted.

onurctirtir:
@marcocitus, I will fix all the above items in my next commit, along with the cause of the build error

src/backend/distributed/planner/multi_logical_planner.c
onurctirtir: it makes sense. Another question is, should I move that function to util file as it is, without merging it ?

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1178,6 +1178,34 @@ HasComplexRangeTableType(Query *queryTree)
 }
 
 
+/*
+ * ExtractRangeTableIndexWalker walks over a join tree, and finds all range
+ * table indexes in that tree.
+ */
+bool
+ExtractRangeTableIndexWalker(Node *node, List **rangeTableIndexList)
```
marcocitus: it could still be moved to query_utils.c.

src/backend/distributed/utils/query_utils.c
```diff
@@ -0,0 +1,109 @@
+/*-------------------------------------------------------------------------
+ *
+ * query_utils.c
+ *
+ * Query-walker utility functions to be used to construct a logical plan
+ * tree from the given query tree structure.
+ *
+ * Copyright (c), Citus Data, Inc.
+ *
+ ****-------------------------------------------------------------------------
```
marcocitus: some extra ***

src/backend/distributed/utils/query_utils.c
```diff
@@ -0,0 +1,109 @@
...
+ */
+
+#include "postgres.h"
+
+#include "nodes/nodeFuncs.h"
+#include "distributed/query_utils.h"
+#include "distributed/query_pushdown_planning.h"
+
+/* This function will be used to walk over a tree to gather entries.
+ * Execution is parameterized by passing walkerMode flag via ExtractRangeTableWalkerContext
```
marcocitus: Our typical commenting style starts with the name of the function:
```c
/*
 * ExtractRangeTableList walks over a tree ...
 */
```

src/backend/distributed/utils/query_utils.c
```diff
@@ -0,0 +1,109 @@
...
+	{
+		Query *query = (Query *) node;
+
+		if (query->hasSubLinks || query->cteList || query->setOperations)
+		{
+			/* descend into all parts of the query */
+			walkIsComplete = query_tree_walker(query,
+											   ExtractRangeTableList,
+											   context,
+											   QTW_EXAMINE_RTES_BEFORE);
```
marcocitus: This is defined in `distributed/version_compat.h`

src/backend/distributed/utils/query_utils.c
```diff
@@ -0,0 +1,109 @@
...
+	{
+		return false;
+	}
+
+	if (IsA(node, RangeTblEntry))
+	{
+		RangeTblEntry *rangeTable = (RangeTblEntry *) node;
+
+		if (walkerMode == EXTRACT_ALL_ENTRIES ||
+			(rangeTable->rtekind == RTE_RELATION && rangeTable->relkind != RELKIND_VIEW))
```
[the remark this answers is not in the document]
onderkalaci: I like the refactoring. But, what about `RELKIND_MATVIEW`? I think we should probably skip that as well as `RELKIND_VIEW ` ?

In general, we seem to be only interested** in `RELKIND_RELATION`, maybe we  should change the if statement to  `rangeTable->rtekind == RTE_RELATION && rangeTable->relkind == RELKIND_RELATION)`. It also makes the `EXTRACT_RELATION_ENTRIES ` flag fit into its name better.

In case anyone interested in other `relkind`s, they can add `EXTRACT_` flags on top of this?

** check-full passes with this [diff](https://github.com/citusdata/citus/compare/onder_test_2) , besides I don't see any code that'd need the other `relkind`s explicitly in Citus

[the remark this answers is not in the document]
onurctirtir: That makes sense, working on the changes

PR #3110, opened 2019-10-16
Change ExtractRangeTable function to make it extendable for further walker types

onurctirtir:
Small changes on ExtractRangeTableList to make it extendable for further walker types

PR #3111, opened 2019-10-16
Refresh Materialized View with Subquery

### LINKED ISSUE #2401: Cannot refresh router materialized view containing subquery
Attempting to REFRESH some materialized views on distributed tables gives the following error, even though the views can be queried and created without an error. 

ERROR:  relation router_view is not distributed

The following is an attempt at a minimal reproduction that reliably exhibits this behavior. 

```
drop table if exists t1 cascade;
drop table if exists t2 cascade;
create table t1 (col1 int,col2 int);
create table t2 (col1 int,col2 int);
select create_distributed_table('t1','col1');
select create_distributed_table('t2','col1');
drop materialized view if exists router_view;
create materialized view router_view AS
select t1.col2,temp_sum.target
from t1,t2,(select sum(col1) AS target from t2) AS temp_sum
where t1.col1=t2.col1;

refresh materialized view router_view;
```

halilozanakgul:
DESCRIPTION: Fixes a bug preventing refresh materialized view

Fixes #2401


src/test/regress/expected/multi_materialized_view.out
```diff
@@ -0,0 +1,125 @@
+---
+--- MULTI_MATERIALIZED_VIEW
+---
+-- This file contains test cases for materialized view support. It verifies
+-- various Citus features: 
+-- materialized views work
+-- insert into... select works with views
+CREATE VIEW air_shipped_lineitems AS SELECT * FROM lineitem_hash_part WHERE l_shipmode = 'AIR';
```
marcocitus: In newer test files we usually do something like: 
```sql
CREATE SCHEMA multi_materialized_view;
SET search_path TO multi_materialized_view;
...
DROP SCHEMA multi_materialized_view CASCADE;
``` 

This ensures the test doesn't leave stuff behind, makes sure the names don't collide with other tests, and also helps detect schema-related issues.

src/test/regress/expected/multi_materialized_view.out
```diff
@@ -0,0 +1,125 @@
+---
+--- MULTI_MATERIALIZED_VIEW
```
marcocitus: The multi_ is there only for historical reasons (Citus used to be in the Postgres source tree and multi_ was used to distinguish Citus files). Newer test files omit it.

src/test/regress/multi_schedule
```diff
@@ -83,6 +83,7 @@ test: multi_task_assignment_policy multi_cross_shard
 test: multi_utility_statements
 test: multi_dropped_column_aliases foreign_key_restriction_enforcement
 test: multi_binary_master_copy_format
+test: multi_materialized_view
```
marcocitus: by adding it to an existing line it will run in parallel with other tests, maybe the same line as multi_view?

src/test/regress/sql/multi_materialized_view.sql
```diff
@@ -0,0 +1,90 @@
...
+
+-- This file contains test cases for materialized view support. It verifies
+-- various Citus features: 
+
+
+-- materialized views work
+-- insert into... select works with views
+CREATE VIEW air_shipped_lineitems AS SELECT * FROM lineitem_hash_part WHERE l_shipmode = 'AIR';
+CREATE TABLE temp_lineitem(LIKE lineitem_hash_part);
+SELECT create_distributed_table('temp_lineitem', 'l_orderkey', 'hash', 'lineitem_hash_part');
```
marcocitus: I'm kind of curious about what happens if the materialized view is initially on a table that was local, but then the table is distributed.

SaitTalhaNisanci:
Just as a small note, you can use imperative language in commit messages rather than past tense. The commit messages should be meaningful when put to the following:

`If applied, this commit will <your commit message goes here>`

git itself uses imperative language for commit messages too.

src/backend/distributed/utils/query_utils.c
```diff
@@ -41,7 +41,8 @@ ExtractRangeTableList(Node *node, ExtractRangeTableWalkerContext *context)
 		RangeTblEntry *rangeTable = (RangeTblEntry *) node;
 
 		if (walkerMode == EXTRACT_ALL_ENTRIES ||
-			(rangeTable->rtekind == RTE_RELATION && rangeTable->relkind != RELKIND_VIEW))
+			(rangeTable->rtekind == RTE_RELATION && rangeTable->relkind != RELKIND_VIEW &&
```
onderkalaci: I think if you rebase onto master after #3110 is merged, can remove this change

onderkalaci:
>Fixes the refresh materialized view with subquery bug and adds extra tests that detect this bug

Can you make sure to include `fixes #issuenumber` so that the issue is automatically closed when this PR is merged

src/test/regress/expected/multi_materialized_view.out
```diff
@@ -0,0 +1,125 @@
...
+-- following is a where false query, should not be inserting anything
+INSERT INTO temp_lineitem SELECT * FROM air_shipped_lineitems WHERE l_shipmode = 'MAIL';
+SELECT count(*) FROM temp_lineitem;
+ count 
+-------
+  1706
+(1 row)
+
+-- can create and query materialized views
+CREATE MATERIALIZED VIEW mode_counts
```
onderkalaci: Can you make sure to add tests with all the  options described in the  postgres docs: https://www.postgresql.org/docs/11/sql-refreshmaterializedview.html

For example, add a test with `refresh materialized view CONCURRENTLY ..`. You may need to create unique index

src/backend/distributed/utils/query_utils.c
onurctirtir: It is merged now

marcocitus:
Also needs a DESCRIPTION: that will appear in the changelog.

PR #3113, opened 2019-10-17
Move coordinator insert..select logic into executor

marcocitus:
Some basic refactoring to prepare for #3078: Move the generation of tasks for INSERT...SELECT via the coordinator into the executor. I previously needed this in #2687 so it seems like a useful thing to do in general.

### DIFF
--- a/src/backend/distributed/executor/insert_select_executor.c
+++ b/src/backend/distributed/executor/insert_select_executor.c
@@ -11,6 +11,7 @@
 #include "postgres.h"
 #include "miscadmin.h"
 
+#include "distributed/citus_ruleutils.h"
 #include "distributed/commands/multi_copy.h"
 #include "distributed/distributed_execution_locks.h"
 #include "distributed/insert_select_executor.h"
@@ -19,7 +20,9 @@
 #include "distributed/multi_executor.h"
 #include "distributed/multi_partitioning_utils.h"
 #include "distributed/multi_physical_planner.h"
+#include "distributed/multi_router_planner.h"
 #include "distributed/distributed_planner.h"
+#include "distributed/recursive_planning.h"
 #include "distributed/relation_access_tracking.h"
 #include "distributed/resource_lock.h"
 #include "distributed/transaction_management.h"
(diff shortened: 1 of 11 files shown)


src/backend/distributed/executor/insert_select_executor.c
```diff
@@ -220,6 +240,231 @@ CoordinatorInsertSelectExecScanInternal(CustomScanState *node)
...
+	return selectQuery;
+}
+
+
+/*
+ * WrapSubquery wraps the given query as a subquery in a newly constructed
+ * "SELECT * FROM (...subquery...) citus_insert_select_subquery" query.
+ */
+static Query *
+WrapSubquery(Query *subquery)
```
marcocitus: Moved ad verbatim from the planner.

src/backend/distributed/executor/insert_select_executor.c
```diff
@@ -220,6 +240,231 @@ CoordinatorInsertSelectExecScanInternal(CustomScanState *node)
...
+}
+
+
+/*
+ * TwoPhaseInsertSelectTaskList generates a list of tasks for a query that
+ * inserts into a target relation and selects from a set of co-located
+ * intermediate results.
+ */
+static List *
+TwoPhaseInsertSelectTaskList(Oid targetRelationId, Query *insertSelectQuery,
```
marcocitus: Moved ad verbatim from the planner.

onderkalaci:
> I previously needed this in #2678 so it seems like a useful thing to do in general.

I guess 2678 is typo?

marcocitus:
Yes, should be #2687

marcocitus:
Somehow I remembered your feedback as: We should do the planning of the SELECT in the planner, which probably makes sense.

> I don't have strong opinions/objections, but it is not very common to create the tasks during the execution.

For (multi-row) inserts we typically generate them in the executor once we know the actual distribution column values. This is a similar situation where we only know which shards we will write to in the executor. 

It probably does not make sense to generate all of them and then prune. Will try to fix that.

src/backend/distributed/executor/insert_select_executor.c
```diff
@@ -97,12 +105,16 @@ CoordinatorInsertSelectExecScanInternal(CustomScanState *node)
 
 	if (!scanState->finishedRemoteScan)
 	{
-		EState *executorState = ScanStateGetExecutorState(scanState);
+		EState *executorState = scanState->customScanState.ss.ps.state;
```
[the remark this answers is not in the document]
marcocitus: Fixed

PR #3114, opened 2019-10-21
Fix memory leak on ReceiveResults

onderkalaci:
Fix memory leak on ReceiveResults

DESCRIPTION: Fixes a memory leak in adaptive executor when query returns many columns

It turns out that `TupleDescGetAttInMetadata()` allocates quite a lot
of memory and we ignore it. And, if the target list is long and 
there are too many rows returning (e.g., `ReceiveResults is called 
frequently during the execution`), the leak becomes apparent.

I considered [this](https://github.com/citusdata/citus/compare/fix_leak?expand=1) fix as well, but I like this more because it could
prevent any future bugs as well.

You can reproduce the issue wout the fix with the following commands:

```SQL

CREATE TABLE users_table (user_id int, time timestamp, value_1 int, value_2 int, value_3 float, value_4 bigint);
SELECT create_distributed_table('users_table', 'user_id');

insert into users_table SELECT i, now(), i, i, i, i FROM generate_series(0,99999)i;

-- load faster

-- 200,000
INSERT INTO users_table SELECT * FROM users_table;

-- 400,000
INSERT INTO users_table SELECT * FROM users_table;

-- 800,000
INSERT INTO users_table SELECT * FROM users_table;

-- 1,600,000
INSERT INTO users_table SELECT * FROM users_table;

-- 3,200,000
INSERT INTO users_table SELECT * FROM users_table;

-- 6,400,000
INSERT INTO users_table SELECT * FROM users_table;

-- 12,800,000
INSERT INTO users_table SELECT * FROM users_table;


-- making the target list entry wider speeds up the leak to show up
 select *,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,* 
 FROM users_table ;

 ```

 TODO:

 - [x] Do a quick benchmark to see if the performance is the same for pgbench 
-  [ ] Can we add a test?

### DIFF
--- a/src/backend/distributed/executor/adaptive_executor.c
+++ b/src/backend/distributed/executor/adaptive_executor.c
@@ -262,6 +262,17 @@ typedef struct DistributedExecution
 
 	/* statistics on distributed execution */
 	DistributedExecutionStats *executionStats;
+
+	/*
+	 * The following fields are used while receiving results from remote nodes.
+	 * We store this information here to avoid re-allocating it every time.
+	 *
+	 * columnArray field is reset/calculated per row, so might be useless for other
+	 * contexts. The benefit of keeping it here is to avoid allocating the array
+	 * over and over again.
+	 */
+	AttInMetadata *attributeInputMetadata;
+	char **columnArray;
 } DistributedExecution;
 
 /*
@@ -849,6 +860,19 @@ CreateDistributedExecution(RowModifyLevel modLevel, List *taskList, bool hasRetu
 
 	execution->connectionSetChanged = false;
 	execution->waitFlagsChanged = false;
+
+	/* allocate execution specific data once, on the ExecutorState memory context */
+	if (tupleDescriptor != NULL)
+	{
+		execution->attributeInputMetadata = TupleDescGetAttInMetadata(tupleDescriptor);
+		execution->columnArray =
+			(char **) palloc0(tupleDescriptor->natts * sizeof(char *));
+	}
+	else
+	{
+		execution->attributeInputMetadata = NULL;
+		execution->columnArray = NULL;
+	}
 
 	if (ShouldExecuteTasksLocally(taskList))
 	{
@@ -3016,23 +3040,27 @@ ReceiveResults(WorkerSession *session, bool storeRows)
 	DistributedExecution *execution = workerPool->distributedExecution;
 	DistributedExecutionStats *executionStats = execution->executionStats;
 	TupleDesc tupleDescriptor = execution->tupleDescriptor;
-	AttInMetadata *attributeInputMetadata = NULL;
+	AttInMetadata *attributeInputMetadata = execution->attributeInputMetadata;
 	uint32 expectedColumnCount = 0;
-	char **columnArray = NULL;
+	char **columnArray = execution->columnArray;
 	Tuplestorestate *tupleStore = execution->tupleStore;
+	MemoryContext ioContext = NULL;
 
-	MemoryContext ioContext = AllocSetContextCreate(CurrentMemoryContext,
-													"ReceiveResults",
-													ALLOCSET_DEFAULT_MINSIZE,
-													ALLOCSET_DEFAULT_INITSIZE,
-													ALLOCSET_DEFAULT_MAXSIZE);
 	if (tupleDescriptor != NULL)
 	{
-		attributeInputMetadata = TupleDescGetAttInMetadata(tupleDescriptor);
 		expectedColumnCount = tupleDescriptor->natts;
-		columnArray = (char **) palloc0(expectedColumnCount * sizeof(char *));
 	}
 
+	/*
+	 * We use this context while converting each row fetched from remote node
+	 * into tuple. The context is reseted on every row, thus we create it at the
+	 * start of the loop and reset on every iteration.
+	 */
+	ioContext = AllocSetContextCreate(CurrentMemoryContext,
+									  "IoContext",
+									  ALLOCSET_DEFAULT_MINSIZE,
+									  ALLOCSET_DEFAULT_INITSIZE,
+									  ALLOCSET_DEFAULT_MAXSIZE);
 
 	while (!PQisBusy(connection->pgConn))
 	{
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/backend/distributed/executor/adaptive_executor.c [hunk 1]
```diff
@@ -3020,19 +3020,45 @@ ReceiveResults(WorkerSession *session, bool storeRows)
...
-	MemoryContext ioContext = AllocSetContextCreate(CurrentMemoryContext,
-													"ReceiveResults",
-													ALLOCSET_DEFAULT_MINSIZE,
-													ALLOCSET_DEFAULT_INITSIZE,
-													ALLOCSET_DEFAULT_MAXSIZE);
 	if (tupleDescriptor != NULL)
 	{
 		attributeInputMetadata = TupleDescGetAttInMetadata(tupleDescriptor);
 		expectedColumnCount = tupleDescriptor->natts;
 		columnArray = (char **) palloc0(expectedColumnCount * sizeof(char *));
```
marcocitus: could we just store these in the DistributedExecution to avoid re-allocating them at all?

onderkalaci: makes sense, I added it. Well, the extra memory context (`ReceiveResults`) now seems useless. Still, I've a slight preference to keep it to prevent such cases all together. What is your take?

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -262,6 +262,18 @@ typedef struct DistributedExecution
...
+	/*
+	 * Used while receiving results from remote nodes. We store this information
+	 * here to avoid re-allocating it every time.
+	 *
+	 * columnArray field is reset/calculated per row, so might be useless for other
+	 * contextes. The benefit of keeping it here is to avoid allocating the array
+	 * over and over again.
+	 */
+	AttInMetadata *attributeInputMetadata;
+	uint32 expectedColumnCount;
```
marcocitus: Do we need to store this separately?

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -3015,24 +3042,50 @@ ReceiveResults(WorkerSession *session, bool storeRows)
...
+	 * times per distributed execution, so we're guarding against any memory leaks with
+	 * using a specific memory context. This context is deleted at the end of the function,
+	 * so be careful to add early returns to the function or relying on anything
+	 * allocated within the function.
+	 *
+	 * Note that tuplestore_puttuple() copies the tuples that are stored in the tuplestore
+	 * in the tuplestore's own memory context. So, we don't need to anything special
+	 * for that.
+	 */
+	oldContext = MemoryContextSwitchTo(receiveResultsContext);
```
marcocitus: Is this still used for anything? How much extra overhead does creating a memory context incur?

onderkalaci: >Is this still used for anything?

No it is not,  see [here](https://github.com/citusdata/citus/pull/3114#discussion_r336956442)

I haven't measured it. But I'm now leaning towards  removing it because this is  a performance critical code-path. Do you also think similar?

src/backend/distributed/executor/adaptive_executor.c
onderkalaci: No, we can use `tupleDesc->nattr`, I'll switch  to that

src/backend/distributed/executor/adaptive_executor.c
marcocitus: Yes, let's remove it.

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -3020,19 +3044,31 @@ ReceiveResults(WorkerSession *session, bool storeRows)
...
 	Tuplestorestate *tupleStore = execution->tupleStore;
+	MemoryContext ioContext = NULL;
+
+	/*
+	 * These fields are computed once at the start of the execution for performance
+	 * reasons. So, do not (p)free these fields, these are set to local variables for
+	 * readability purpose.
+	 */
+	attributeInputMetadata = execution->attributeInputMetadata;
+	columnArray = execution->columnArray;
```
marcocitus: This could probably move to the initialisation section. I'm not sure the comments will add much. It's kind of obvious that you should not pfree things from the execution.

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -262,6 +262,17 @@ typedef struct DistributedExecution
 
 	/* statistics on distributed execution */
 	DistributedExecutionStats *executionStats;
+
+	/*
+	 * Used while receiving results from remote nodes. We store this information
+	 * here to avoid re-allocating it every time.
+	 *
+	 * columnArray field is reset/calculated per row, so might be useless for other
+	 * contextes. The benefit of keeping it here is to avoid allocating the array
```
marcocitus: contexts

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -262,6 +262,17 @@ typedef struct DistributedExecution
 
 	/* statistics on distributed execution */
 	DistributedExecutionStats *executionStats;
+
+	/*
+	 * Used while receiving results from remote nodes. We store this information
+	 * here to avoid re-allocating it every time.
```
marcocitus: Could use a comment on what these fields are for.

onderkalaci:
>  Do a quick benchmark to see if the performance is the same for pgbench

pgbench performance is the same as expected.

PR #3131, opened 2019-10-25
Do not assign InvalidOid for local execution while extracting parameters

### LINKED ISSUE #3084: Local execution in function with custom types errors out
Found a weird bug in local execution + custom types.

On coordinator:
```sql
CREATE TYPE invite_resp AS ENUM ('yes', 'no', 'maybe'); 

CREATE TABLE event_responses (
  event_id int,
  user_id int,
  response invite_resp,
  primary key (event_id, user_id)
);
SELECT create_distributed_table('event_responses', 'event_id');
CREATE OR REPLACE PROCEDURE register_for_event(p_event_id int, p_user_id int, p_choice invite_resp)
LANGUAGE plpgsql AS $fn$
BEGIN
   INSERT INTO event_responses VALUES (p_event_id, p_user_id, p_choice)
   ON CONFLICT (event_id, user_id)
   DO UPDATE SET response = EXCLUDED.response;
END;
$fn$;

SELECT create_distributed_function('register_for_event(int,int,invite_resp)', 'p_event_id', 'event_responses');

CALL register_for_event(16, 1, 'yes');
ERROR:  there is no parameter $3
PL/pgSQL function register_for_event(integer,integer,invite_resp) line 3 at SQL statement
```

If I remove p_choice from the insert then it works as expected.

SaitTalhaNisanci:
While extracting parameters from `ParamListInfo` we shouldnt assign 0 for custom types. 0 is an `InvalidOid` therefore when resolving the parameters, postgres sees that the parameter has an invalid oid, so it gives the error in #3084.

After running 5 times, postgres binds the parameters so the error does not happen.

Fixes #3084.

### DIFF
--- a/src/backend/distributed/executor/multi_router_executor.c
+++ b/src/backend/distributed/executor/multi_router_executor.c
@@ -1767,8 +1767,9 @@ ExtractParametersForRemoteExecution(ParamListInfo paramListInfo, Oid **parameter
 									const char ***parameterValues)
 {
 	ExtractParametersFromParamList(paramListInfo, parameterTypes,
-							  parameterValues, false);
+								   parameterValues, false);
 }
+
 
 /*
  * ExtractParametersFromParamList extracts parameter types and values from
@@ -1777,9 +1778,9 @@ ExtractParametersForRemoteExecution(ParamListInfo paramListInfo, Oid **parameter
  */
 void
 ExtractParametersFromParamList(ParamListInfo paramListInfo,
-						  Oid **parameterTypes,
-						  const char ***parameterValues, bool
-						  useOriginalCustomTypeOids)
+							   Oid **parameterTypes,
+							   const char ***parameterValues, bool
+							   useOriginalCustomTypeOids)
 {
 	int parameterIndex = 0;
 	int parameterCount = paramListInfo->numParams;
--- a/src/backend/distributed/executor/multi_router_executor.c
+++ b/src/backend/distributed/executor/multi_router_executor.c
@@ -117,10 +117,10 @@ static bool StoreQueryResult(CitusScanState *scanState, MultiConnection *connect
 static bool ConsumeQueryResult(MultiConnection *connection, bool
 							   alwaysThrowErrorOnFailure, int64 *rows);
 
-void ExtractParametersFromParamListInfoInternal(ParamListInfo paramListInfo,
-												Oid **parameterTypes,
-												const char ***parameterValues, bool
-												isLocalExecution);
+static void ExtractParametersInternal(ParamListInfo paramListInfo,
+									  Oid **parameterTypes,
+									  const char ***parameterValues, bool
+									  isLocalExecution);
 
 /*
  * AcquireMetadataLocks acquires metadata locks on each of the anchor
@@ -1729,8 +1729,8 @@ SendQueryInSingleRowMode(MultiConnection *connection, char *query,
 		/* force evaluation of bound params */
 		paramListInfo = copyParamList(paramListInfo);
 
-		ExtractParametersFromParamListInfo(paramListInfo, &parameterTypes,
-										   &parameterValues);
+		ExtractParametersForRemoteExecution(paramListInfo, &parameterTypes,
+											&parameterValues);
 
 		querySent = SendRemoteCommandParams(connection, query, parameterCount,
 											parameterTypes, parameterValues);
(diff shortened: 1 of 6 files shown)


src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -1783,7 +1800,7 @@ ExtractParametersFromParamListInfo(ParamListInfo paramListInfo, Oid **parameterT
 		 * the master and worker nodes. Therefore, the worker nodes can
 		 * infer the correct oid.
 		 */
-		if (parameterData->ptype >= FirstNormalObjectId)
+		if (parameterData->ptype >= FirstNormalObjectId && !isLocalExecution)
```
onderkalaci: what happens if you always assign `(*parameterTypes)[parameterIndex] = parameterData->ptype;`

onderkalaci: I mean, what is the difference in passing `InvalidOid` to the remote nodes vs `wrongOid`. I'm not saying we should do it, just curious.

If postgres is ignoring the type, we might do that.

SaitTalhaNisanci: @onderkalaci, with that we dont get the same behavior:

```
 -- test parameterized inserts
 PREPARE prepared_insert(varchar(20)) AS
 	INSERT INTO router_executor_table VALUES (1, $1, $2);
 EXECUTE prepared_insert('comment-1', '(1, 10)');
+ERROR:  cache lookup failed for type 22949
```

https://circleci.com/gh/citusdata/citus/51604

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -1764,6 +1768,30 @@ SendQueryInSingleRowMode(MultiConnection *connection, char *query,
...
+void
+ExtractParametersFromParamListInfoLocal(ParamListInfo paramListInfo, Oid **parameterTypes,
+										const char ***parameterValues)
+{
+	ExtractParametersFromParamListInfoInternal(paramListInfo, parameterTypes,
+											   parameterValues, true);
+}
+
+
+void
```
onderkalaci: Needs function comment, and could use `static` qualifier

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -1764,6 +1768,30 @@ SendQueryInSingleRowMode(MultiConnection *connection, char *query,
 void
 ExtractParametersFromParamListInfo(ParamListInfo paramListInfo, Oid **parameterTypes,
 								   const char ***parameterValues)
+{
+	ExtractParametersFromParamListInfoInternal(paramListInfo, parameterTypes,
+											   parameterValues, false);
+}
+
+
+/*
```
onderkalaci: The comment is the same with `ExtractParametersFromParamListInfo `, I think it'd be good if we can have different comments for the functions.

Also, the function name(s) sounds a little bit hard to understand, what about we rename to so that it is pretty clear to the readers : `ExtractParametersFromParamListInfoLocal ` > `ExtractParametersForLocalExecution`
`ExtractParametersFromParamListInfo ` -> `ExtractParametersForRemoteExecution `

SaitTalhaNisanci:
In `ExtractParametersFromParamList` function we extract two things:
- ParamTypes
- ParamValues
In remote execution, when we send the `parameterValues` along with `parameterTypes`. If a type does not exist on the other side of the connection it gives an error. So when we create a custom type, we might have different oids in different nodes. Say we have oid `x` in coordinator and oid `y` in worker1. If we send a query with a parameter type `x`, postgres will error saying that `x` does not exist, however it already has the value of parameter which has a type of `x`, so we dont use this to get the value.

However in local execution, we use a different mechanism. We do not use the `parameterValues` directly. We have a reference to parameters, and if an `InvalidOid` exists, postgres gives an error. https://github.com/postgres/postgres/blob/9555cc8d2b02c4191d67ba39f589b39b01735518/src/backend/parser/parse_param.c#L107

onderkalaci:
> If we send a query with a parameter type x, postgres will error saying that x does not exist, however it already has the value of parameter which has a type of x, so we dont use this to get the value.

This explains some part of it. But, in that case, wouldn't you expect it should work for all types (including base types like `int`), if we set the `oid = 0` in the `parameterTypes` array for remote execution? But, as far as I can see, it fails for base types.

So, we still don't have an answer on what's different for custom types on remote execution in terms of `parameterType` value. Am I missing anything?

src/backend/distributed/executor/multi_router_executor.c
```diff
@@ -1758,12 +1762,44 @@ SendQueryInSingleRowMode(MultiConnection *connection, char *query,
...
+ * ExtractParametersInternal extracts parameter types and values from
+ * the given ParamListInfo structure, and fills parameter type and value arrays.
+ * If isLocalExecution is true, it does not change the oid of custom types.
+ */
+void
+ExtractParametersInternal(ParamListInfo paramListInfo,
+						  Oid **parameterTypes,
+						  const char ***parameterValues, bool
+						  isLocalExecution)
```
marcocitus: The name of this variable describes something about the caller, rather than about what its effect is. Maybe something along the lines of `setCustomTypeOids`? Possible uses can be described in the comments.

SaitTalhaNisanci:
@onderkalaci, it doesnt fail for base types, it only fails for custom types. Base types have the same oid in coordinator and workers, so it is not a problem. We do not assign InvalidOid to base types, we do this only for the custom types.

If you look at the circleci job that fails(https://circleci.com/gh/citusdata/citus/51604), you might be tempted to think that it fails for `int` type, but it doesnt if you look at it carefully(the failing query is with the custom type and that is because we do not have the sent oid in worker)

onderkalaci:
> So, we still don't have an answer on what's different for custom types on remote execution in terms of parameterType value. Am I missing anything?

I'd like to conclude this discussion. I think Citus' logic is slightly incorrect/inconsistent for base types.

If Citus sends only base types to the remote node, in the first 5 executions, the parameters in the query are fully replaced with constants and after 6th time it is a parameter:
```SQL
PREPARE base_type_filter(int) AS SELECT count(*) from users_Table WHERE value_1 = $1 AND user_id = 5; 
-- first 5 times
EXECUTE base_type_filter(5);
 SELECT count(*) AS count FROM public.users_table_102014 users_table WHERE ((value_1 OPERATOR(pg_catalog.=) 5) AND (user_id OPERATOR(pg_catalog.=) 5))
...
-- sixth time the parameter is not replaced
EXECUTE base_type_filter(5);
SELECT count(*) AS count FROM public.users_table_102014 users_table WHERE ((value_1 OPERATOR(pg_catalog.=) $1) AND (user_id OPERATOR(pg_catalog.=) 5))
```

For composite types, the query always includes the parameter:
```SQL
PREPARE prepared_custom_type(varchar(20)) AS
INSERT INTO router_executor_table VALUES (1, $1, $2);

EXECUTE prepared_custom_type('comment-3', '(1, 10)');
INSERT INTO public.router_executor_table_102074 (id, comment, stats) VALUES (1, $1, $2)
...
-- 6th time
EXECUTE prepared_custom_type('comment-3', '(1, 10)');
INSERT INTO public.router_executor_table_102074 (id, comment, stats) VALUES (1, $1, $2)
```

And, on all cases, Citus sends the `parameterType` over the wire. In case of the first five executions of base-types, Postgres errors out if (a) There is a parameter value/type (b) there is no parameter in the query so that Postgres cannot infer the data type. The error message would be something like `could not determine data type of parameter`. This happens even if the parameter is not actually used in the query.

src/backend/distributed/executor/multi_router_executor.c
SaitTalhaNisanci: I have changed it to `useOriginalCustomTypeOids`

SaitTalhaNisanci:
@onderkalaci I think I found how it works with `InvalidOid` (0).

From postgres https://www.postgresql.org/docs/current/libpq-exec.html#LIBPQ-EXEC-MAIN


paramTypes[]
Specifies, by OID, the data types to be assigned to the parameter symbols. If paramTypes is NULL, or **any particular element in the array is zero**, the server infers a data type for the parameter symbol in the same way it would do for an untyped literal string.

metdos:
> Specifies, by OID, the data types to be assigned to the parameter symbols. If paramTypes is NULL, or **any particular element in the array is zero**, the server infers a data type for the parameter symbol in the same way it would do for an untyped literal string.

That was my assumption too, if we keep that code part, maybe you can add this as a comment to there...

onderkalaci:
> That was my assumption too, if we keep that code part, maybe you can add this as a comment to there...

We've also talked with @SaitTalhaNisanci on this, using `InvalidOid` for base types doesn't work because we don't send them as string literals. I presume if we send them as strings, we can avoid sending the `Oid`, but definitely very low priority experiment to do now.

PR #3132, opened 2019-10-25
Don't warn for declaration-after-statement since we only support GNU99

JelteF:
This change was actually already intended in #3124. However, pg_config
manually adds this warning too. This way we undo that.

To confirm that it works one function was changed to make use of not
having the warning anymore.


hanefi:
Does this mean that we will be able to mix declarations and statements in our code base?

Will the coders & reviewers be responsible to have them seperate in the future?

JelteF:
> Does this mean that we will be able to mix declarations and statements in our code base?

Yes you will be able to.

> Will the coders & reviewers be responsible to have them seperate in the future?

What would be the reason to do so?
There are a lot of readability advantages to not declaring variables at the start of the function:
1. [No boilerplate NULL declarations](https://github.com/citusdata/citus/pull/3132/files#diff-72d969594f1dc954d8d2167d15f65ddcL369-L375), because you can [initialize the declaration with a useful value instead](https://github.com/citusdata/citus/pull/3132/files#diff-72d969594f1dc954d8d2167d15f65ddcR380). 
2. [Having Assert statements or other checks on arguments at the top of the function](https://github.com/citusdata/citus/pull/3132/files#diff-72d969594f1dc954d8d2167d15f65ddcR368), this way it's instantly clear what the function invariants are.
3. [Having declaration and usage close together](https://github.com/citusdata/citus/pull/3132/files#diff-72d969594f1dc954d8d2167d15f65ddcR368), so it's clear what the type of something is or (the other way around) what a variable is used for.

hanefi:
Thanks for listing the advantages so clearly. I used to be indifferent to the removal of that flag, now I feel like this is a really nice thing to do.

PR #3134, opened 2019-10-25
Clean old citus install before installing extension

JelteF:
One case where this would have been useful recently was when we changed
the `citus--9.0-1--9.1-1.sql` file to `citus--9.0-2--9.1-1.sql`. If you
had the old file installed it would never be cleaned. Any changes in
the 9.0-2 version would not be applied, because postgres tries to find
the shortest migration path, so it would always run the 9.0-1--9.1-1
version, because then it could skip 9.0-1--9.0-2.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@
 
 citus_subdir = .
 citus_top_builddir = .
+extension_dir = $(shell $(PG_CONFIG) --sharedir)/extension
 
 # Hint that configure should be run first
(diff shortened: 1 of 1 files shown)


marcocitus:
The user might not have permission to remove files from the installation directory. Should this be done on make install?

JelteF:
@marcocitus, that's a good point. Updated the PR to this every install. It's actually even nicer, since you don't even have to do `make clean` and destroy your build cache.

PR #3135, opened 2019-10-25
Include all relevant relations in the ExtractRangeTableRelationWalker

onderkalaci:
We've changed the logic for pulling `RTE_RELATION`s in #3109 and #3110, and it turns out that #3110 breaks some edge cases with non-colocated subquery joins and partitioned tables. @onurctirtir found this [steps](https://github.com/citusdata/citus/pull/3104#issuecomment-545787394) where I traced back and found the issues.

While looking into it in more detail, we decided to expand the list in a way that the callers get all the relevant `RTE_RELATION`s `RELKIND_RELATION`, `RELKIND_PARTITIONED_TABLE` , `RELKIND_FOREIGN_TABLE`  and `RELKIND_MATVIEW`. These are all relation kinds that Citus planner is aware of.

### DIFF
--- a/src/backend/distributed/utils/query_utils.c
+++ b/src/backend/distributed/utils/query_utils.c
@@ -17,6 +17,10 @@
 #include "distributed/version_compat.h"
 #include "nodes/nodeFuncs.h"
 
+
+static bool CitusQueryableRangeTableRelation(RangeTblEntry *rangeTableEntry);
+
+
 /*
  * ExtractRangeTableList walks over a tree to gather entries.
  * Execution is parameterized by passing walkerMode flag via ExtractRangeTableWalkerContext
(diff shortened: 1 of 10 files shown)


src/backend/distributed/utils/query_utils.c
```diff
@@ -40,12 +40,18 @@ ExtractRangeTableList(Node *node, ExtractRangeTableWalkerContext *context)
...
+		 * the code, so fetch them as well.
+		 */
 		if (walkerMode == EXTRACT_ALL_ENTRIES || (walkerMode ==
 												  EXTRACT_RELATION_ENTRIES &&
 												  rangeTable->rtekind == RTE_RELATION &&
-												  rangeTable->relkind ==
-												  RELKIND_RELATION))
+												  (rangeTable->relkind ==
+												   RELKIND_RELATION ||
+												   rangeTable->relkind ==
```
marcocitus: What about foreign tables and such?

The alignment in this if statement makes it very hard to read.

src/backend/distributed/utils/query_utils.c
```diff
@@ -40,12 +44,14 @@ ExtractRangeTableList(Node *node, ExtractRangeTableWalkerContext *context)
...
-		if (walkerMode == EXTRACT_ALL_ENTRIES || (walkerMode ==
-												  EXTRACT_RELATION_ENTRIES &&
-												  rangeTable->rtekind == RTE_RELATION &&
-												  rangeTable->relkind ==
-												  RELKIND_RELATION))
+		/*
+		 * Make sure that we are extracting only relation entries if walkerMode is set to
+		 * EXTRACT_RELATION_ENTRIES. We treat partitioned tables as relations thoughout
+		 * the code, so fetch them as well.
+		 */
```
onderkalaci: @marcocitus what do you think about this change overall? I think #3110 was a good first step, but clearly missed some pieces.

This seems to be the way to go for me, `ExtractRangeTableRelationWalker ` returns all the RTE_RELATIONs that Citus might be interested in. 

I need to change few callers, which expects only distributed tables but relying on this function.

marcocitus:
Maybe update PR description.

PR #3137, opened 2019-10-28
refactor some of hard coded values in citus gucs

SaitTalhaNisanci:
DESCRIPTION: refactor some of hard coded values in citus gucs

For gucs we have flags, and `0` means, the guc does not have any filter. We are using variables for all flags but `0`, I thought that it makes more sense to explicitly write that. Also when a unit flag is given, if there is no filter there is a hidden `0` flag, I have made that explicit. 

Time constants are added for better readability of values.


src/backend/distributed/shared_library_init.c
```diff
@@ -371,9 +372,9 @@ RegisterCitusConfigVariables(void)
 		gettext_noop("Sets the maximum duration to connect to worker nodes."),
 		NULL,
 		&NodeConnectionTimeout,
-		5000, 10, 60 * 60 * 1000,
+		5 * MS_PER_SECOND, 10 * MS, MS_PER_HOUR,
 		PGC_USERSET,
-		GUC_UNIT_MS,
+		GUC_UNIT_MS | GUC_ALLOW_ALL,
```
marcocitus: ALLOW_ALL suggests this field is about permissions, but it's about a few different things. I do find it useful to make the code more self-documenting. Maybe `GUC_STANDARD`?

SaitTalhaNisanci: Yeah, I forgot to include that in the description but I wasnt sure if ALLOW_ALL was a good choice, thanks for the suggestion.

PR #3142, opened 2019-10-30
Remove real time router executors

onderkalaci:
DESCRIPTION: Drop support for deprecated real-time and router executors

As of Citus 8.3, adaptive executor is the default executor. We continued to support real-time/router executors as a fallback mechanism in case of an issue/bug in adaptive executor. We now feel comfortable enough to remove the existing executors to reduce the technical debt in the code.


src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -1118,6 +1112,86 @@ DistributedExecutionRequiresRollback(DistributedExecution *execution)
...
+/*
+ * TaskListRequires2PC determines whether the given task list requires 2PC
+ * because the tasks provided operates on a reference table or there are multiple
+ * tasks and the commit protocol is 2PC.
+ *
+ * Note that we currently do not generate tasks lists that involves multiple different
+ * tables, thus we only check the first task in the list for reference tables.
+ */
+static bool
+TaskListRequires2PC(List *taskList)
```
onderkalaci: copy & paste from previous location

src/backend/distributed/executor/distributed_execution_locks.c
```diff
@@ -0,0 +1,449 @@
+/*-------------------------------------------------------------------------
```
onderkalaci: copy & paste from previous location (all file)

src/backend/distributed/executor/multi_task_tracker_executor.c
```diff
@@ -3033,3 +3036,19 @@ TaskTrackerExecScan(CustomScanState *node)
 
 	return resultSlot;
 }
+
+
+/*
```
onderkalaci: copy & paste from previous location

src/backend/distributed/shared_library_init.c
```diff
@@ -1066,7 +1069,7 @@ RegisterCitusConfigVariables(void)
 		task_executor_type_options,
 		PGC_USERSET,
 		GUC_STANDARD,
-		NULL, NULL, NULL);
+		WarnIfDeprecatedExecutorUsed, NULL, NULL);
```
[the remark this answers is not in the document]
onderkalaci: I think that variable is misnamed. If you disable it, Citus skips router planning as well. I think it is still not clear to me after reading the Pr that enabled it .(https://github.com/citusdata/citus/commit/c3a60bff7049dd88fd001baa553dd62ee444d698)

I think we should drop that GUC with a major release

src/backend/distributed/executor/multi_server_executor.c
```diff
@@ -2,7 +2,7 @@
  *
  * multi_server_executor.c
```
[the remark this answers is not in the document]
onderkalaci: thanks for catching those!  I think I've checked wrong code-cov. report, otherwise I shouldn't have missed this :/

onderkalaci:
> I think some branches of FindPlacementListConnection() are unreachable now, i.e. else if (placementConnection->hadDDL) and else if (placementConnection->hadDML). If that is correct, maybe remove those.

That's a good one (and tricky) to remove. I've double checked with @marcocitus on this, here is the commit which removes the [code](https://github.com/citusdata/citus/pull/3142/commits/3da013b9c471fdf881a72ec0cb726002c8ccf3eb), I wanted to share explicitly since these checks used to be at the center of enforcing transaction visibility rules/preventing self-deadlocks in TX blocks.

Let me know if you see anything wrong in removing that piece. I'm excited to remove this pretty complex code-block, thanks for reminding!

PR #3147, opened 2019-11-01
Fix debug1 message in JobExecutorType

SaitTalhaNisanci:
When citus.enable_repartition_joins guc is set to on, and we have
adaptive executor, there was a typo in the debug message, which was
saying realtime executor no adaptive executor.

### DIFF
--- a/src/backend/distributed/executor/multi_server_executor.c
+++ b/src/backend/distributed/executor/multi_server_executor.c
@@ -105,7 +105,7 @@ JobExecutorType(DistributedPlan *distributedPlan)
 			}
 
 			ereport(DEBUG1, (errmsg(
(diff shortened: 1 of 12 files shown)


src/backend/distributed/executor/multi_server_executor.c
```diff
@@ -105,7 +105,7 @@ JobExecutorType(DistributedPlan *distributedPlan)
 			}
 
 			ereport(DEBUG1, (errmsg(
```
onderkalaci: should update the comment above as well

SaitTalhaNisanci: Oh I saw this after merging, I will create another PR and update this.

PR #3149, opened 2019-11-01
add IsLoggableLevel utility function

SaitTalhaNisanci:
This PR adds log utils file to separate log utilities.

### DIFF
--- a/src/backend/distributed/planner/recursive_planning.c
+++ b/src/backend/distributed/planner/recursive_planning.c
@@ -68,6 +68,7 @@
 #include "distributed/query_pushdown_planning.h"
 #include "distributed/recursive_planning.h"
 #include "distributed/relation_restriction_equivalence.h"
+#include "distributed/log_utils.h"
 #include "distributed/version_compat.h"
 #include "lib/stringinfo.h"
 #include "optimizer/planner.h"
@@ -222,8 +223,7 @@ GenerateSubplansForSubqueriesAndCTEs(uint64 planId, Query *originalQuery,
 		RaiseDeferredError(error, ERROR);
 	}
 
-	if (context.subPlanList && (log_min_messages <= DEBUG1 || client_min_messages <=
-								DEBUG1))
+	if (context.subPlanList && IsLoggableLevel(DEBUG1))
 	{
 		StringInfo subPlanString = makeStringInfo();
 		pg_get_query_def(originalQuery, subPlanString);
@@ -730,7 +730,7 @@ RecursivelyPlanCTEs(Query *query, RecursivePlanningContext *planningContext)
 
 		subPlanId = list_length(planningContext->subPlanList) + 1;
 
-		if (log_min_messages <= DEBUG1 || client_min_messages <= DEBUG1)
+		if (IsLoggableLevel(DEBUG1))
 		{
 			StringInfo subPlanString = makeStringInfo();
 			pg_get_query_def(subquery, subPlanString);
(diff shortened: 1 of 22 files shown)


onderkalaci:
> DESCRIPTION: add IsLoggableLevel utility function

The text written in the DESCRIPTION fields show up in the release notes. I think we shouldn't have these code level changes in the change log. We'd mostly want to have user-facing changes.

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -1151,7 +1151,7 @@ RecursivelyPlanSubquery(Query *subquery, RecursivePlanningContext *planningConte
 	 */
 	resultQuery = BuildSubPlanResultQuery(subquery->targetList, NIL, resultId);
 
-	if (log_min_messages <= DEBUG1 || client_min_messages <= DEBUG1)
+	if (IsLoggableLevel(DEBUG1))
```
onderkalaci: To me this change is not an improvement `IsLoggableLevel` doesn't reveal much about what we're trying to do. When I see `log_min_messages ` / `client_min_messages ` I can understand what we're trying to do, but with `IsLoggableLevel ` not.

Besides, Postgres mostly follows the convention before the change, I'd prefer to stick to that

SaitTalhaNisanci: Why do you think it doesnt reveal what we are trying to do? When I ask the question `what does this if check do?` the answer will probably be: `It checks if DEBUG1 is a level that we can log, so basically it answers the question: `isLoggableLevel`.

onderkalaci: > It checks if DEBUG1 is a level that we can log

In postgres, `log` can be configured differently for [clients](https://www.postgresql.org/docs/9.5/runtime-config-client.html#GUC-CLIENT-MIN-MESSAGES) like psql, and for [server logs](https://www.postgresql.org/docs/9.5/runtime-config-logging.html#GUC-LOG-MIN-MESSAGES).  When you use `log` in the function name, you're slightly hiding `client_min_messages` from this context. So, the term `loggable` is not as clear to me as you think.

Also, for such a simple function, do we really need source/header file(s)? What else are we planning to put there in the future?

I'm not hardly against merging this to master, what I'm just saying is that I couldn't find this change as an improvement over what we have, so I'm neutral, I'm not planning to approve. If someone else thinks that it is a nice improvement, let's go ahead and merge.

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -1446,6 +1446,7 @@ TransformFunctionRTE(RangeTblEntry *rangeTblEntry)
 			subquery->targetList = lappend(subquery->targetList, targetEntry);
 		}
 	}
+
```
onderkalaci: If we decide to merge this, let's don't forget to remove these extra new-lines

src/backend/distributed/utils/log_utils.c
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+#include "postgres.h"
+
+#include "utils/guc.h"
+#include "distributed/log_utils.h"
+
+
+/*
+ * In postgres, log can be configured differently for clients and servers.
```
marcocitus: Should start with `IsLoggableLevel returns ...` per the usual style.

src/backend/distributed/utils/log_utils.c
```diff
@@ -0,0 +1,25 @@
+/*-------------------------------------------------------------------------
+ *
+ * log_utils.c
+ *	  Utilities regarding logs
```
marcocitus: It would kind of make sense to move ApplyLogRedaction over here, though it would complicate merge into enterprise.

marcocitus:
Can you add a short PR description that describes why we are making this change?

src/backend/distributed/utils/log_utils.c
SaitTalhaNisanci: I guess we should move it now since this is the right file for it and at some point we will need to do that merge. (Assuming that we want methods to be in the right file)

SaitTalhaNisanci: Also in community we still have that method as a "dummy" one. So I am updating the headers in this PR, I would only need to move the function in enterprise I guess.

PR #3150, opened 2019-11-01
create a utility method to mark tasks as failed

SaitTalhaNisanci:

### DIFF
--- a/src/backend/distributed/worker/task_tracker.c
+++ b/src/backend/distributed/worker/task_tracker.c
@@ -92,6 +92,7 @@ static void ManageWorkerTasksHash(HTAB *WorkerTasksHash);
 static void ManageWorkerTask(WorkerTask *workerTask, HTAB *WorkerTasksHash);
 static void RemoveWorkerTask(WorkerTask *workerTask, HTAB *WorkerTasksHash);
 static void CreateJobDirectoryIfNotExists(uint64 jobId);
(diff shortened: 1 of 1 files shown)


src/backend/distributed/worker/task_tracker.c
```diff
@@ -1065,6 +1062,14 @@ ManageWorkerTask(WorkerTask *workerTask, HTAB *WorkerTasksHash)
 }
 
 
+static void
+MarkAsFailed(WorkerTask *workerTask)
```
marcocitus: Maybe MarkTaskAsFailed?

Missing function comment

SaitTalhaNisanci: I prefer self documenting code, so I try to avoid comments as much as possible. `MarkWorkerTaskAsFailed` is a better name for sure and if we rename it to that, would we need a comment?

marcocitus: Function comments are required in the Citus (and PostgreSQL) code base, with good reason.

It's not at all clear from reading this function what we achieve by modifying these 2 fields / marking a task as failed. The function comment is an opportunity to explain that to the reader.

The primary goal of a function in our code base is to break a problem down into well-defined subproblems. Repetitive code is a sign of a common subproblem, but not in itself a reason to write a function. We should be able to describe what subproblem this code solves in a function comment. Otherwise, this function just hides important state changes from the reader without further explanation.

PR #3152, opened 2019-11-01
Don't maintain replicationfactor of reference tables in pg_dist_colocation

pykello:
We don't use this value anywhere. So always set it to -1 so we don't have to maintain it in different cases.


serprex:
Looks good. Do we need a migration with `UPDATE pg_dist_colocation SET replicationfactor = -1 WHERE distributioncolumntype = 0`?

pykello:
> Looks good. Do we need a migration with `UPDATE pg_dist_colocation SET replicationfactor = -1 WHERE distributioncolumntype = 0`?

It might be nice, but it is not necessary, since we don't use this value.

pykello:
> > Looks good. Do we need a migration with `UPDATE pg_dist_colocation SET replicationfactor = -1 WHERE distributioncolumntype = 0`?
> 
> It might be nice, but it is not necessary, since we don't use this value.

As communicated in an internal chat, I was wrong. I will add the UPDATE to the migration script.

src/backend/distributed/utils/reference_table_utils.c
```diff
@@ -389,9 +382,8 @@ uint32
 CreateReferenceTableColocationId()
 {
 	uint32 colocationId = INVALID_COLOCATION_ID;
-	List *workerNodeList = ActivePrimaryNodeList(ShareLock);
 	int shardCount = 1;
-	int replicationFactor = list_length(workerNodeList);
+	int replicationFactor = -1;
```
marcocitus: Could use a comment (or a constant?)

marcocitus:
One step further would be to eliminate the pg_dist_colocation record entirely and replace it with something like -1, but this is already a lot less clunky.

PR #3155, opened 2019-11-04
Replicate reference tables to coordinator, except planner changes

pykello:
DESCIPTION: Replicate reference tables to the coordinators.

After this change, users can add the coordinator to pg_dist_node by `SELECT master_add_node(hostname, port, groupId => 0)`, and once they do that all reference tables will also have a shard on the coordinator.

### DIFF
--- a/src/backend/distributed/commands/utility_hook.c
+++ b/src/backend/distributed/commands/utility_hook.c
@@ -65,6 +65,7 @@ bool EnableDDLPropagation = true; /* ddl propagation is enabled */
 PropSetCmdBehavior PropagateSetCommands = PROPSETCMD_NONE; /* SET prop off */
 static bool shouldInvalidateForeignKeyGraph = false;
 static int activeAlterTables = 0;
+static int activeDropSchemaOrDBs = 0;
 
 
 /* Local functions forward declarations for helper functions */
@@ -76,6 +77,7 @@ static List * PlanRenameAttributeStmt(RenameStmt *stmt, const char *queryString)
 static List * PlanAlterOwnerStmt(AlterOwnerStmt *stmt, const char *queryString);
 static List * PlanAlterObjectDependsStmt(AlterObjectDependsStmt *stmt,
 										 const char *queryString);
+static bool IsDropSchemaOrDB(Node *parsetree);
 
 
 /*
@@ -655,6 +657,11 @@ multi_ProcessUtility(PlannedStmt *pstmt,
 			activeAlterTables++;
 		}
 
+		if (IsDropSchemaOrDB(parsetree))
+		{
+			activeDropSchemaOrDBs++;
+		}
+
 		standard_ProcessUtility(pstmt, queryString, context,
 								params, queryEnv, dest, completionTag);
 
@@ -678,12 +685,22 @@ multi_ProcessUtility(PlannedStmt *pstmt,
 		{
 			activeAlterTables--;
 		}
+
+		if (IsDropSchemaOrDB(parsetree))
+		{
+			activeDropSchemaOrDBs--;
+		}
 	}
 	PG_CATCH();
 	{
 		if (IsA(parsetree, AlterTableStmt))
 		{
 			activeAlterTables--;
+		}
+
+		if (IsDropSchemaOrDB(parsetree))
+		{
+			activeDropSchemaOrDBs--;
 		}
 
 		PG_RE_THROW();
@@ -793,6 +810,26 @@ multi_ProcessUtility(PlannedStmt *pstmt,
 	 * EXTENSION. This is important to register some invalidation callbacks.
 	 */
 	CitusHasBeenLoaded();
+}
+
+
+/*
+ * IsDropSchemaOrDB returns true if parsetree represents DROP SCHEMA ...or
+ * a DROP DATABASE.
+ */
+static bool
+IsDropSchemaOrDB(Node *parsetree)
+{
+	DropStmt *dropStatement = NULL;
+
+	if (!IsA(parsetree, DropStmt))
+	{
+		return false;
+	}
+
+	dropStatement = (DropStmt *) parsetree;
+	return (dropStatement->removeType == OBJECT_SCHEMA) ||
+		   (dropStatement->removeType == OBJECT_DATABASE);
 }
 
 
@@ -1197,4 +1234,15 @@ bool
 AlterTableInProgress(void)
 {
 	return activeAlterTables > 0;
+}
+
+
+/*
+ * DropSchemaOrDBInProgress returns true if we're processing a DROP SCHEMA
+ * or a DROP DATABASE command right now.
+ */
+bool
+DropSchemaOrDBInProgress(void)
+{
+	return activeDropSchemaOrDBs > 0;
 }
--- a/src/backend/distributed/executor/subplan_execution.c
+++ b/src/backend/distributed/executor/subplan_execution.c
@@ -56,7 +56,7 @@ ExecuteSubPlans(DistributedPlan *distributedPlan)
 	 */
 	BeginOrContinueCoordinatedTransaction();
 
-	nodeList = ActiveReadableNodeList();
+	nodeList = ActiveReadableWorkerNodeList();
 
 	foreach(subPlanCell, subPlanList)
 	{
--- a/src/backend/distributed/master/master_delete_protocol.c
+++ b/src/backend/distributed/master/master_delete_protocol.c
@@ -422,7 +422,21 @@ DropShards(Oid relationId, char *schemaName, char *relationName,
 								 quotedShardName);
 			}
 
-			connection = GetPlacementConnection(connectionFlags, shardPlacement, NULL);
+			/*
+			 * The active DROP SCHEMA/DATABASE ... CASCADE will drop the shard, if we
+			 * try to drop it over another connection, we will get into a distributed
+			 * deadlock.
+			 */
+			if (shardPlacement->groupId == COORDINATOR_GROUP_ID &&
+				IsCoordinator() &&
+				DropSchemaOrDBInProgress())
+			{
+				DeleteShardPlacementRow(shardPlacement->placementId);
+				continue;
+			}
+
+			connection = GetPlacementConnection(connectionFlags, shardPlacement,
+												NULL);
 
 			RemoteTransactionBeginIfNecessary(connection);
 
--- a/src/backend/distributed/master/worker_node_manager.c
+++ b/src/backend/distributed/master/worker_node_manager.c
@@ -44,6 +44,9 @@ static List * PrimaryNodesNotInList(List *currentList);
 static WorkerNode * FindRandomNodeFromList(List *candidateWorkerNodeList);
 static bool OddNumber(uint32 number);
 static bool ListMember(List *currentList, WorkerNode *workerNode);
+static bool NodeIsPrimaryWorker(WorkerNode *node);
+static bool NodeCanHaveDistTablePlacements(WorkerNode *node);
+static bool NodeIsReadableWorker(WorkerNode *node);
 
 
 /* ------------------------------------------------------------
@@ -294,12 +297,12 @@ WorkerGetNodeWithName(const char *hostname)
 
 
 /*
- * ActivePrimaryNodeCount returns the number of groups with a primary in the cluster.
+ * ActivePrimaryWorkerNodeCount returns the number of groups with a primary in the cluster.
  */
 uint32
-ActivePrimaryNodeCount(void)
+ActivePrimaryWorkerNodeCount(void)
 {
-	List *workerNodeList = ActivePrimaryNodeList(NoLock);
+	List *workerNodeList = ActivePrimaryWorkerNodeList(NoLock);
 	uint32 liveWorkerCount = list_length(workerNodeList);
 
 	return liveWorkerCount;
@@ -307,15 +310,25 @@ ActivePrimaryNodeCount(void)
 
 
 /*
- * ActiveReadableNodeCount returns the number of groups with a node we can read from.
+ * ActiveReadableWorkerNodeCount returns the number of groups with a node we can read from.
  */
 uint32
-ActiveReadableNodeCount(void)
+ActiveReadableWorkerNodeCount(void)
 {
-	List *workerNodeList = ActiveReadableNodeList();
+	List *workerNodeList = ActiveReadableWorkerNodeList();
 	uint32 liveWorkerCount = list_length(workerNodeList);
 
 	return liveWorkerCount;
+}
+
+
+/*
+ * NodeIsCoordinator returns true if the given node represents the coordinator.
+ */
+bool
+NodeIsCoordinator(WorkerNode *node)
+{
+	return node->groupId == COORDINATOR_GROUP_ID;
 }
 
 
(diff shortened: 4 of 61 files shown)


src/backend/distributed/master/worker_node_manager.c
```diff
@@ -365,8 +381,23 @@ FilterActiveNodeListFunc(LOCKMODE lockMode, bool (*checkFunction)(WorkerNode *))
 List *
 ActivePrimaryNodeList(LOCKMODE lockMode)
 {
+	bool excludeCoordinator = true;
+	EnsureModificationsCanRun();
+	return FilterActiveNodeListFunc(lockMode, WorkerNodeIsPrimary, excludeCoordinator);
```
onderkalaci: I know it is more code to write, but having both the filter function and `excludeCoordinator ` parameter makes it little unintuitive. How much work is to have different filter functions?

src/backend/distributed/metadata/metadata_sync.c
```diff
@@ -124,6 +124,13 @@ StartMetadatSyncToNode(char *nodeNameString, int32 nodePort)
 								escapedNodeName, nodePort)));
 	}
 
+	if (workerNode->groupId == 0)
+	{
+		ereport(WARNING, (errmsg("%s:%d is the coordinator and already contains "
+								 "metadata, skipping.", nodeNameString, nodePort)));
```
onderkalaci: I think `NOTICE` might be more accurate. Is there a reason to warn the user for this? 

And, minor: `skipping ` -> `skipping to sync the metadata` might be little more useful

src/backend/distributed/transaction/worker_transaction.c
```diff
@@ -118,6 +118,11 @@ TargetWorkerSetNodeList(TargetWorkerSet targetWorkerSet, LOCKMODE lockMode)
 	foreach(workerNodeCell, workerNodeList)
 	{
 		WorkerNode *workerNode = (WorkerNode *) lfirst(workerNodeCell);
+		if (workerNode->groupId == 0)
```
onderkalaci: Using `NodeIsCoordinator ` could make it easier to read the function. Otherwise, a comment could be useful.

src/backend/distributed/utils/node_metadata.c
```diff
@@ -1091,11 +1093,17 @@ AddNodeMetadata(char *nodeName, int32 nodePort,
...
-	if (nodeMetadata->groupId == 0)
+	if (nodeMetadata->groupId == INVALID_GROUP_ID)
 	{
 		nodeMetadata->groupId = GetNextGroupId();
 	}
 
+	/* if this is a coordinator, we shouldn't place shards on it */
+	if (nodeMetadata->groupId == 0)
+	{
+		nodeMetadata->shouldHaveShards = false;
```
onderkalaci: the field name `shouldHaveShards ` sounds incomplete now. We'll have reference table shards on the coordinator. But I think it is not easy to come up with a nicer name ?

src/backend/distributed/worker/worker_data_fetch_protocol.c
```diff
@@ -670,12 +670,18 @@ worker_append_table_to_shard(PG_FUNCTION_ARGS)
...
+	 *
+	 * We don't acquire lock on the shard id to avoid deadlock when doing
+	 * upgrade_to_reference_table(). upgrade_to_reference_table() acquires
+	 * a lock on shard resource on the coordinator, and opens connections
+	 * to all nodes and calls worker_append_table_to_shard(). This connection
+	 * might be a connection to the coordinator.
 	 */
 	shardId = ExtractShardIdFromTableName(shardTableName, false);
-	LockShardResource(shardId, AccessExclusiveLock);
+	LockRelationOid(RelnameGetRelid(shardTableName), AccessExclusiveLock);
```
onderkalaci: given the `upgrade_to_reference_table` is a lot less commonly used, can't we acquire `ExclusiveLock` (or even `AccessExclusive`) on the reference table itself while doing `upgrade_to_reference_table`? I'd rather not change this since it touches a lot more places?

src/backend/distributed/master/worker_node_manager.c
pykello: My primary reason was that I couldn't find a good name for such filter functions. Now that I think, `NodeIsPrimaryWorker` and `NodeIsPrimary` should work.

src/test/regress/expected/foreign_key_to_reference_table.out
```diff
@@ -527,12 +527,9 @@ SELECT count(*) FROM referencing_schema.referencing_table;
    800
 (1 row)
 
+DROP TABLE referenced_schema.referenced_table, referencing_schema.referencing_table;
```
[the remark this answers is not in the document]
pykello: no, reverted.

src/backend/distributed/master/master_delete_protocol.c
```diff
@@ -422,28 +422,38 @@ DropShards(Oid relationId, char *schemaName, char *relationName,
 								 quotedShardName);
 			}
 
-			connection = GetPlacementConnection(connectionFlags, shardPlacement, NULL);
+			/*
+			 * The active DROP SCHEMA ... CASCADE will drop the shard, if we try to drop
+			 * it over another connection, we will get into a distributed deadlock.
+			 */
+			if (!(shardPlacement->groupId == 0 &&
```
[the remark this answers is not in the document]
pykello: done

src/backend/distributed/commands/utility_hook.c
```diff
@@ -653,6 +655,11 @@ multi_ProcessUtility(PlannedStmt *pstmt,
 			activeAlterTables++;
 		}
 
+		if (IsDropSchema(parsetree))
```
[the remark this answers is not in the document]
pykello: done

src/backend/distributed/metadata/metadata_sync.c

src/backend/distributed/transaction/worker_transaction.c
pykello: done.

src/backend/distributed/worker/worker_data_fetch_protocol.c
pykello: done

src/backend/distributed/utils/node_metadata.c
pykello: Maybe `shouldHaveDistTablePlacements`, but changing this will probably need a change in `pg_dist_node`, so I think I'd keep for a separate PR.

onderkalaci: Makes sense, let's keep it for now. There is already a comment next to the field definition.

src/backend/distributed/transaction/worker_transaction.c
```diff
@@ -111,13 +111,18 @@ SendCommandToWorkers(TargetWorkerSet targetWorkerSet, const char *command)
...
 {
-	List *workerNodeList = ActivePrimaryNodeList(lockMode);
+	List *workerNodeList = ActivePrimaryWorkerNodeList(lockMode);
 	ListCell *workerNodeCell = NULL;
 	List *result = NIL;
 
 	foreach(workerNodeCell, workerNodeList)
 	{
 		WorkerNode *workerNode = (WorkerNode *) lfirst(workerNodeCell);
+		if (NodeIsCoordinator(workerNode))
```
onderkalaci: I guess we don't need  this anymore

src/backend/distributed/transaction/distributed_deadlock_detection.c
```diff
@@ -108,7 +108,7 @@ CheckForDistributedDeadlocks(void)
 	TransactionNode *transactionNode = NULL;
 	int edgeCount = 0;
 	int localGroupId = GetLocalGroupId();
-	List *workerNodeList = ActiveReadableNodeList();
+	List *workerNodeList = ActiveReadableWorkerNodeList();
```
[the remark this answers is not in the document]
pykello: Distributed deadlock detection runs on the coordinator and I think it also checks the local locks, so we are not skipping the coordinator. I added an isolation test for this. Am I missing something?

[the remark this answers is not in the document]
onderkalaci: The deadlock detection logic only considers transactions that are originated on the node doing the detection: https://github.com/citusdata/citus/blob/94a7e6475ccdb63b88bcce4472ad0eb9fda04059/src/backend/distributed/transaction/distributed_deadlock_detection.c#L145-L150

So, if queries fired from worker nodes deadlock with each other including the coordinator, we cannot detect by the deadlock detector on the coordinator.

src/backend/distributed/utils/resource_lock.c
```diff
@@ -233,7 +233,7 @@ LockShardListResourcesOnFirstWorker(LOCKMODE lockmode, List *shardIntervalList)
 static bool
 IsFirstWorkerNode()
 {
-	List *workerNodeList = ActivePrimaryNodeList(NoLock);
+	List *workerNodeList = ActivePrimaryWorkerNodeList(NoLock);
```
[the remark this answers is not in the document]
pykello: I agree.

src/backend/distributed/transaction/backend_data.c
```diff
@@ -216,7 +216,7 @@ get_global_active_transactions(PG_FUNCTION_ARGS)
 {
 	TupleDesc tupleDescriptor = NULL;
 	Tuplestorestate *tupleStore = NULL;
-	List *workerNodeList = ActivePrimaryNodeList(NoLock);
+	List *workerNodeList = ActivePrimaryWorkerNodeList(NoLock);
```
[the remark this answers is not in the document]
pykello: added tests.

src/backend/distributed/transaction/distributed_deadlock_detection.c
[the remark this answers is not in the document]
marcocitus: We should include the coordinator, both in the locks (does it become the first worker now?) and in the deadlock detection.

We do not prevent shards from being placed on the coordinator, and inserts can conflict, and all_modifications_commutative may be enabled.

src/backend/distributed/commands/truncate.c
```diff
@@ -228,7 +228,7 @@ static void
 AcquireDistributedLockOnRelations(List *relationIdList, LOCKMODE lockMode)
 {
 	Oid relationId = InvalidOid;
-	List *workerNodeList = ActivePrimaryNodeList(NoLock);
+	List *workerNodeList = ActivePrimaryWorkerNodeList(NoLock);
```
marcocitus: should this include the coordinator? I assume the truncate also happens there?

src/test/regress/expected/multi_mx_add_coordinator.out
```diff
@@ -0,0 +1,76 @@
+CREATE SCHEMA mx_add_coordinator;
```
onderkalaci: minor: I think we do not prefer `multi` prefix anymore

src/backend/distributed/transaction/distributed_deadlock_detection.c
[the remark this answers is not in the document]
onderkalaci: > We should include the coordinator, both in the locks (does it become the first worker now?)

We're still acquiring the lock on the first worker node. Does it really matter where we acquire the lock? Note that the writes on the coordinator also acquires the lock on the first worker node.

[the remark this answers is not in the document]
onderkalaci: > We do not prevent shards from being placed on the coordinator,

The coordinator node is marked `shouldhaveshards = false`. Doesn't that prevent non-reference table shards on the coordinator?

src/backend/distributed/commands/truncate.c
onderkalaci: Oh, yes, it is true.

src/backend/distributed/executor/subplan_execution.c
```diff
@@ -56,7 +56,7 @@ ExecuteSubPlans(DistributedPlan *distributedPlan)
 	 */
 	BeginOrContinueCoordinatedTransaction();
 
-	nodeList = ActiveReadableNodeList();
+	nodeList = ActiveReadableWorkerNodeList();
```
onderkalaci: I think this is only true if we merge #3170, otherwise we could fail to support the following because the anchor shard might be the coordinator for the reference table:

```SQL
SET citus.task_assignment_policy TO "round-robin";
WITH cte1 AS (SELECT value_1, random() FROM users_table) SELECT count(*) FROM users_table_ref JOIN cte1 USING (value_1);
```

Anyway, just wanted to mention so that we don't forget to test on the main PR

src/backend/distributed/transaction/distributed_deadlock_detection.c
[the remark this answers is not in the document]
onderkalaci: I'd be happy if we can conclude the discussion because this seems to the blocker for now:

1) I think as @marcocitus we should acquire the locks on the coordinator for `TRUNCATE reference_table` at `AcquireDistributedLockOnRelations `

2) As @marcocitus suggested, we might get into distributed deadlocks happening on the coordinator because of `citus.all_modifications_commutative`  (though I couldn't reproduce, it should be possible)

3) On MX, citus acquires executor locks on the first worker node for reference table modifications (`AcquireExecutorShardLocks()`->`AcquireExecutorShardLockForRowModify()` -> ` AcquireExecutorShardLockForRowModify()` -> `SerializeNonCommutativeWrites()` -> `LockShardListResourcesOnFirstWorker()` -> `SELECT lock_shard_resources()`). **The modifications on the coordinator still acquires the executor lock on the first worker node. I don't think we need to change that, it should be fine that we still acquire the lock on the first worker.**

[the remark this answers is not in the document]
marcocitus: > We're still acquiring the lock on the first worker node. Does it really matter where we acquire the lock? Note that the writes on the coordinator also acquires the lock on the first worker node.

No. I was mostly saying that we should make sure workers include the coordinator if it can be the first worker, but excluding the coordinator from this list is fine.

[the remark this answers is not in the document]
marcocitus: > The coordinator node is marked shouldhaveshards = false. Doesn't that prevent non-reference table shards on the coordinator?

No, "shoudhaveshards=false" means it should not have shards, not that it does not have shards.

src/backend/distributed/metadata/metadata_sync.c
```diff
@@ -124,9 +124,17 @@ StartMetadatSyncToNode(char *nodeNameString, int32 nodePort)
...
+		ereport(NOTICE, (errmsg("%s:%d is the coordinator and already contains "
+								"metadata, skipping syncing the metadata",
+								nodeNameString, nodePort)));
+		return;
+	}
+
 	MarkNodeHasMetadata(nodeNameString, nodePort, true);
 
-	if (!WorkerNodeIsPrimary(workerNode))
+	if (!NodeIsPrimary(workerNode))
```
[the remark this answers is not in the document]
pykello: > Do we actually need this change given that we returned for the coordinator?
> 
> I think it is easier to read the code with the change, just wanted to to make sure not to miss any details for myself.

I am not sure if I understand your point. This is not a change, we just renamed `WorkerNodeIsPrimary()` to `NodeIsPrimary()`, and it is to not sync metadata directly to secondaries since they receive the changes using streaming replication from the primary.

pykello:
Updated the following:
1. Include the coordinator placement for `citus_..._size(reference_table)`.
2. Change `master_create_empty_shard` to include only shouldhaveshard nodes.
3. Include coordinator in deadlock detection.
4. Lock coordinator for distributed truncate.

src/backend/distributed/executor/subplan_execution.c
pykello: Makes sense. I'll merge it to master after merging this PR to master.

src/backend/distributed/transaction/lock_graph.c
```diff
@@ -91,7 +91,7 @@ dump_global_wait_edges(PG_FUNCTION_ARGS)
 WaitGraph *
 BuildGlobalWaitGraph(void)
 {
-	List *workerNodeList = ActiveReadableNodeList();
+	List *workerNodeList = ActiveReadableWorkerNodeList();
```
onderkalaci: I  think according to the discussions,  we should involve coordinator to  the deadlock detection, so shouldn't this be `ActivePrimaryNodeList` ?

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -2119,7 +2119,7 @@ PlanRouterQuery(Query *originalQuery,
 	}
 	else if (replacePrunedQueryWithDummy)
 	{
-		List *workerNodeList = ActiveReadableNodeList();
+		List *workerNodeList = ActiveReadableWorkerNodeList();
```
onderkalaci: Self-note: we might consider adding only using localNode here as we have local execution.

src/backend/distributed/transaction/lock_graph.c
pykello: done

src/backend/distributed/commands/create_distributed_table.c
```diff
@@ -306,10 +305,7 @@ create_reference_table(PG_FUNCTION_ARGS)
 	 */
 	EnsureRelationKindSupported(relationId);
 
-	workerNodeList = ActivePrimaryNodeList(ShareLock);
-	workerCount = list_length(workerNodeList);
-
-	/* if there are no workers, error out */
+	workerCount = ActivePrimaryWorkerNodeCount();
 	if (workerCount == 0)
```
marcocitus: This is create_reference_table, shouldn't we include the coordinator?

marcocitus: Do we still need to error out here?

pykello: done

PR #3157, opened 2019-11-05
Fix queries with repartition joins and group by unique column

JelteF:
DESCRIPTION: Fix queries with repartition joins and group by unique column

Postgres doesn't require you to add all columns that are in the target list to
the GROUP BY when you group by a unique column (or columns). It even actively
removes these group by clauses when you do.

This is normally fine, but for repartition joins it is not. The reason for this
is that the temporary tables don't have these primary key columns. So when the
worker executes the query it will complain that it is missing columns in the
group by.

This PR fixes that by adding an `ANY_VALUE` aggregate around each variable in
the target list that does is not contained in the group by or in an aggregate.
This is done only for repartition joins.

### DIFF
--- a/src/backend/distributed/planner/multi_logical_optimizer.c
+++ b/src/backend/distributed/planner/multi_logical_optimizer.c
@@ -1847,6 +1847,7 @@ MasterAggregateExpression(Aggref *originalAggregate,
 		 */
 		Var *column = NULL;
 		TargetEntry *columnTargetEntry = NULL;
+		Aggref *newMasterAggregate = NULL;
 
 		/* worker aggregate and original aggregate have the same return type */
 		Oid workerReturnType = exprType((Node *) originalAggregate);
@@ -1857,7 +1858,16 @@ MasterAggregateExpression(Aggref *originalAggregate,
 		Oid aggregateFunctionId = AggregateFunctionOid(aggregateName, workerReturnType);
 		Oid masterReturnType = get_func_rettype(aggregateFunctionId);
 
-		Aggref *newMasterAggregate = copyObject(originalAggregate);
+		/*
+		 * If return type aggregate is anyelement, it's actual return type is
+		 * determined on the type of its argument. So we replace it with the
+		 * argument type in that case.
+		 */
+		if (masterReturnType == ANYELEMENTOID)
+		{
+			masterReturnType = workerReturnType;
+		}
+		newMasterAggregate = copyObject(originalAggregate);
 		newMasterAggregate->aggdistinct = NULL;
 		newMasterAggregate->aggfnoid = aggregateFunctionId;
 		newMasterAggregate->aggtype = masterReturnType;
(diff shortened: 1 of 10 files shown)


src/backend/distributed/utils/metadata_cache.c
```diff
@@ -136,6 +136,7 @@ typedef struct MetadataCacheData
 	Oid readIntermediateResultFuncId;
 	Oid extraDataContainerFuncId;
 	Oid workerHashFunctionId;
+	Oid firstFunctionId;
```
marcocitus: anyValue

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -955,6 +967,56 @@ TargetEntryList(List *expressionList)
 }
 
 
+static Node *
+AddAnyValueAggregates(Node *node, void *context)
```
marcocitus: Needs a function comment, especially since we're doing some unusual magic here.

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -672,6 +676,14 @@ BuildJobQuery(MultiNode *multiNode, List *dependedJobList)
 	/* build group clauses */
 	groupClauseList = QueryGroupClauseList(multiNode);
 
+	if (groupClauseList != NIL && isRepartitionJoin)
+	{
+		targetList = (List *) expression_tree_mutator((Node *) targetList,
```
marcocitus: Maybe add:
```c
/* 
 * Group by on primary key allows all columns to appear in the target
 * list, but after re-partitioning we will be querying an intermediate
 * table that does not have the primary key. We therefore wrap all the
 * columns that do not appear in the GROUP BY in an any_value aggregate.
 */
```

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -2216,6 +2217,21 @@ CitusWorkerHashFunctionId(void)
 }
 
 
+/* return oid of the worker_hash function */
```
marcocitus: stale comment

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -1856,8 +1857,11 @@ MasterAggregateExpression(Aggref *originalAggregate,
 		const char *aggregateName = AggregateNames[aggregateType];
 		Oid aggregateFunctionId = AggregateFunctionOid(aggregateName, workerReturnType);
 		Oid masterReturnType = get_func_rettype(aggregateFunctionId);
-
-		Aggref *newMasterAggregate = copyObject(originalAggregate);
+		if (masterReturnType == ANYELEMENTOID)
+		{
+			masterReturnType = workerReturnType;
```
marcocitus: This line is not covered by a test, when is it used? Could also use a comment with an example.

JelteF: Fixed now

src/backend/distributed/sql/udfs/any_value/latest.sql
```diff
@@ -0,0 +1,13 @@
+CREATE OR REPLACE FUNCTION pg_catalog.any_value_agg ( anyelement, anyelement )
+RETURNS anyelement AS $$
+        SELECT CASE WHEN $1 IS NULL THEN $2 ELSE $1 END;
+$$ LANGUAGE SQL STABLE;
+
+CREATE AGGREGATE pg_catalog.any_value (
+        sfunc    = pg_catalog.any_value_agg,
+        basetype = anyelement,
+        stype    = anyelement
```
marcocitus: Can you add combinefunc = pg_catalog.any_value_agg? Then it can go through the new aggregate logic.

PR #3159, opened 2019-11-05
Replicate reference tables to coordinator: rename OpenConnectionToAllNodes

serprex:
create_restore_point creates a restore point on coordinator before having rest of nodes create restore point. No reason to have it create on coordinator again

This is already the case. This PR renames a function to be clear about that & adds testing reference tables with create_restore_point. Actual testing of create_restore_point is lacking

### DIFF
--- a/src/backend/distributed/master/citus_create_restore_point.c
+++ b/src/backend/distributed/master/citus_create_restore_point.c
@@ -30,7 +30,7 @@
 
 
 /* local functions forward declarations */
-static List * OpenConnectionsToAllNodes(LOCKMODE lockMode);
+static List * OpenConnectionsToAllWorkerNodes(LOCKMODE lockMode);
 static void BlockDistributedTransactions(void);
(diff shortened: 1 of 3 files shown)


src/backend/distributed/master/citus_create_restore_point.c
```diff
@@ -171,28 +174,35 @@ CreateRemoteRestorePoints(char *restoreName, List *connectionList)
 	{
 		MultiConnection *connection = (MultiConnection *) lfirst(connectionCell);
 
-		int querySent = SendRemoteCommandParams(connection, CREATE_RESTORE_POINT_COMMAND,
-												parameterCount, parameterTypes,
-												parameterValues);
-		if (querySent == 0)
+		if (GroupForNode(connection->hostname, connection->port) != 0)
```
pykello: Probably not necessary, since `ActivePrimaryNodeList()` doesn't include the coordinator. Maybe just rename `OpenConnectionsToAllNodes()` to `OpenConnectionsToAllWorkerNodes()` and add some test.

PR #3160, opened 2019-11-06
Local execution should not change hasReturning for distributed tables

onderkalaci:
It looks like the logic to prevent RETURNING in reference tables to
have duplicate entries that comes from local and remote executions
leads to missing some tuples for distributed tables.

With this PR, we're ensuring to kick in the logic for reference tables
only.

DESCRIPTION: Fixes a bug in local execution that could cause some missing rows in RETURNING

### DIFF
--- a/src/backend/distributed/executor/adaptive_executor.c
+++ b/src/backend/distributed/executor/adaptive_executor.c
@@ -738,12 +738,6 @@ RunLocalExecution(CitusScanState *scanState, DistributedExecution *execution)
 static void
 AdjustDistributedExecutionAfterLocalExecution(DistributedExecution *execution)
 {
-	/*
-	 * Local execution already stores the  tuples for returning, so we should not
-	 * store again.
-	 */
-	execution->hasReturning = false;
-
 	/* we only need to execute the remote tasks */
 	execution->tasksToExecute = execution->remoteTaskList;
 
@@ -1467,8 +1461,9 @@ AssignTasksToConnections(DistributedExecution *execution)
 												sizeof(TaskPlacementExecution *));
 		shardCommandExecution->placementExecutionCount = placementExecutionCount;
 
-		shardCommandExecution->expectResults = hasReturning ||
-											   modLevel == ROW_MODIFY_READONLY;
+		shardCommandExecution->expectResults =
+			(hasReturning && !task->partiallyLocalOrRemote) ||
+			modLevel == ROW_MODIFY_READONLY;
 
 		foreach(taskPlacementCell, task->taskPlacementList)
 		{
--- a/src/backend/distributed/executor/local_executor.c
+++ b/src/backend/distributed/executor/local_executor.c
@@ -241,7 +241,7 @@ ExtractLocalAndRemoteTasks(bool readOnly, List *taskList, List **localTaskList,
 		}
 		else
 		{
-			Task *localTask = copyObject(task);
+			Task *localTask = NULL;
 			Task *remoteTask = NULL;
 
 			/*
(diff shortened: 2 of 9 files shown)


src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -738,17 +738,31 @@ RunLocalExecution(CitusScanState *scanState, DistributedExecution *execution)
...
 	execution->unfinishedTaskCount = list_length(execution->remoteTaskList);
+
+	/*
+	 * For reference tables, local execution already stores the  tuples for
+	 * returning, so we should not store again.
+	 *
+	 * We typically expect to have modifications to reference tables when the
+	 * unfinishedTaskCount is one. However, we still explicitly check for
+	 * reference tables because there are some edge cases for distributed
+	 * tables with 2 shards.
```
marcocitus: What are the edge cases?

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -738,17 +738,31 @@ RunLocalExecution(CitusScanState *scanState, DistributedExecution *execution)
...
+	 */
+	if (execution->hasReturning && execution->unfinishedTaskCount == 1)
+	{
+		Task *initialTask = (Task *) linitial(execution->tasksToExecute);
+
+		if (initialTask->anchorShardId != INVALID_SHARD_ID &&
+			ReferenceTableShardId(initialTask->anchorShardId))
```
marcocitus: This fix seems messy. Could we keep track of which tasks were split up? e.g. set a flag on the Task in 	`ExtractLocalAndRemoteTasks`?

marcocitus: An alternative is that we create the ShardCommandExecution states before the local execution and keep track in there.

src/backend/distributed/executor/adaptive_executor.c
onderkalaci:
```SQL


SELECT start_metadata_sync_to_node(nodename, nodeport) FROM pg_dist_node;

SET citus.shard_replication_model TO 'streaming';

SET citus.shard_count TO 2;

CREATE TABLE test_table (a INT);
SELECT create_distributed_table('test_table', 'a');


-- connect to a worker
psql -p 9700

-- do a remote insert
INSERT INTO test_table VALUES (2);

BEGIN;
	-- do a local insert
	INSERT INTO test_table VALUES (1);

    -- if we don't check for reference tables, this would skip returning 2
	DELETE FROM test_table RETURNING *;
```

src/backend/distributed/executor/adaptive_executor.c
onderkalaci: > This fix seems messy. Could we keep track of which tasks were split up? e.g. set a flag on the Task in ExtractLocalAndRemoteTasks?

I considered that as well, which makes sense. But isn't it already equivalent for `ReferenceTableShardId(initialTask->anchorShardId) && taskType == ROUTER_TASK`? Since we could have both local and remote tasks only for modifications, I skipped checking  `taskType == ROUTER_TASK`. I think `hasReturning` check already also ensures that this is a modification.

(I'm trying to avoid adding things to `Task` which is only relevant for a very specific case, but if you think that's OK, I can do that)

marcocitus: I think making the notion of "this task is part local, part remote" explicit would be helpful in making code more understandable.

onderkalaci: Added the explicit field, and moved the decision to `ShardCommandExecution `, which makes more sense?

PR #3161, opened 2019-11-06
CTE pushdown via CTE inlining in distributed planning

onderkalaci:
Before this patch, Citus used to always recursively plan CTEs.
In PostgreSQL 12, there is a [logic](https://git.postgresql.org/gitweb/?p=postgresql.git;a=commitdiff;h=608b167f9f9c4553c35bb1ec0eab9ddae643989b) for inlining CTEs, which is basically converting certain CTEs to subqueries.

With this patch, citus becomes capable of doing the same, can get rid of
recursively planning all the CTEs. Instead, the pushdown-able ones would
simply be converted to subquery pushdown. If the inlined CTE query cannot
be pushed down, it'd simply follow the recursive planning logic.

See an example below:

```SQL
-- the query that users pass
WITH some_users AS
(SELECT
        users_table.user_id
FROM
 users_table JOIN events_table USING (user_id) WHERE event_type = 5)
SELECT count(*) FROM users_table JOIN some_users USING (user_id);

-- worker query
SELECT count(*) AS COUNT
FROM ((users_table_102039 users_table
       JOIN users_table_102039 users_table_1 ON ((users_table_1.user_id OPERATOR(pg_catalog.=) users_table.user_id)))
      JOIN events_table_102071 events_table ON ((users_table.user_id OPERATOR(pg_catalog.=) events_table.user_id)))
WHERE (events_table.event_type OPERATOR(pg_catalog.=) 5)
```

There are few things to call-out for future reference and
help the reviewer(s) to understand the patch easier:

1) On top of Postgres' restrictions to inline CTEs, Citus enforces one more.
This is to prevent regressing on the SQL support. For example, the following
cte is OK to inline by Postgres. However, if inlined, Citus cannot plan the whole
query, so we prefer to skip inlining that cte:

```SQL
-- Citus should not inline this CTE because otherwise it cannot
-- plan the query
WITH cte_1 AS (SELECT * FROM test_table) 
SELECT 
	*, row_number() OVER () 
FROM 
	cte_1;
```

2) Some exotic queries with multiple colocation groups involved
   could become repartition joins. Basically, after the CTE inlining
   happens, ShouldRecursivelyPlanNonColocatedSubqueries() fails to
   detect that the query is a non-colocated subquery. We should improve
   there to fix it. But, since we fall-back to planning again, the query is
   successfully executed by Citus.
```SQL
SET citus.shard_count TO 4;
CREATE TABLE  colocation_1 (key int, value int);
SELECT create_distributed_table('colocation_1', 'key');

SET citus.shard_count TO 8;
CREATE TABLE  colocation_2 (key int, value int);
SELECT create_distributed_table('colocation_2', 'key');

-- which used to work because the cte was recursively planned
-- now the cte becomes a repartition join since
    --- (a) the cte is replaced to a subquery
    --- (b) since the subquery is very simple, postgres pulled it to become
    ---     a simple join
WITH cte AS (SELECT * FROM colocation_1)
SELECT count(*) FROM cte JOIN colocation_2 USING (key);
...
message: the query contains a join that requires repartitioning
detail: 
hint: Set citus.enable_repartition_joins to on to enable repartitioning
...
┌───────┐
│ count │
├───────┤
│     0 │
└───────┘
(1 row)
```

3) We decided to implement inlining CTEs even after standard planner. 
In Postgres 12+, the restriction information in CTEs are generated
because the CTEs are actually treated as subqueries via Postgres' inline
capabilities.

In Postgres 11-, the restriction information is not generated for CTEs. Because
of that, some queries work differently on pg 11 vs pg 12. To see such queries,
see cte_inline.sql file, where the file has two output files.

4) As a side-effect of (2), we're now able to inline CTEs for INSERT .. SELECT
queries as well. Postgres prevents it, I cannot see a reason to prevent it. With
this capability, some of the INSERT ... SELECT queries where the cte is in the
SELECT query could become pushdownable. See an example:

```SQL
INSERT INTO test_table
WITH fist_table_cte AS
  (SELECT * FROM test_table)
    SELECT
      key, value
    FROM
      fist_table_cte;
```

5) A class of queries now could be supported. Previously, if a CTE is used in the
outer part of an outer join, Citus would complained about that.

So, the following query:

```SQL
WITH cte AS (
  SELECT * FROM users_table WHERE user_id = 1 ORDER BY value_1
)
SELECT
  cte.user_id, cte.time, events_table.event_type
FROM
  cte
LEFT JOIN
  events_table ON cte.user_id = events_table.user_id
ORDER BY
  1,2,3
LIMIT
  5;
ERROR:  cannot pushdown the subquery
DETAIL:  Complex subqueries and CTEs cannot be in the outer part of the outer join
```

Becomes
```SQL
-- cte LEFT JOIN distributed_table should error out
WITH cte AS (
  SELECT * FROM users_table WHERE user_id = 1 ORDER BY value_1
)
SELECT
  cte.user_id, cte.time, events_table.event_type
FROM
  cte
LEFT JOIN
  events_table ON cte.user_id = events_table.user_id
ORDER BY
  1,2,3
LIMIT
  5;
 user_id |              time               | event_type
---------+---------------------------------+------------
       1 | Wed Nov 22 22:51:43.132261 2017 |          0
       1 | Wed Nov 22 22:51:43.132261 2017 |          0
       1 | Wed Nov 22 22:51:43.132261 2017 |          1
       1 | Wed Nov 22 22:51:43.132261 2017 |          1
       1 | Wed Nov 22 22:51:43.132261 2017 |          2
(5 rows)
```

## We need to add lots of tests as well

- [x] CTE becomes subquery pushdown / recursively planned subquery / non-colocated subquery join
     - [x] In FROM clause (INNER/OUTER joins)
     - [x] In WHERE clause
     - [x] In set operations
     - [x] In a view
     - [x] INSERT .. SELECT pushdown (with ON CONFLICT/RETURNING)
     - [x] CTE calls in Having/ORDER BY/Target List
- [x] CTE Joins
    - [x] Becomes subquery pushdown joins
    - [x] Becomes non-colocated subquery joins
    - [x] The inlined CTEs are small enough, they become joins between tables
- [x] Nested CTEs
    - [x] All the CTEs can be inlined
    - [x] Some CTEs in the nested CTEs can be inlined
- [x] CTEs define in (e.g.,  the query doesn't start WITH CTE ..., but SELECT .. FROM (WITH cte ....)) 
    - [x] FROM clause
    - [x] Having clause
    - [x] WHERE clause
- [x] CTE with local tables
- [x] Make sure router queries works just fine
    - [x] Router queries already relies on Postgres to inline CTEs, make sure that we do not break ever
    - [x] Fast-path router queries never get affected by this
- [x] CTE alias: WITH my_cte (col_name_1, col_name_2)...
- [x] Make sure recursive CTEs, modifying CTEs, CTEs referred multiple times are always ignored
- [x] Make sure that we do not break anything with correlated subqueries/CTEs

- [x] Quite a lot of tests are affected by the change. Should we change those tests to have `random()` to prevent CTE inlining and add similar set of tests for CTE inlining?


src/backend/distributed/planner/distributed_planner.c
```diff
@@ -99,6 +100,19 @@ static void ResetPlannerRestrictionContext(
 static bool HasUnresolvedExternParamsWalker(Node *expression, ParamListInfo boundParams);
 
 
+typedef struct inline_cte_walker_context
```
onderkalaci: `inline_cte_walker_context `, `inline_cte_walker `, `contain_dml ` and `contain_dml_walker ` copy paste from postgres.


`inline_cte ` and `InlineCTEs ` are heavily based on Postgres code

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -794,6 +811,171 @@ CreateDistributedPlan(uint64 planId, Query *originalQuery, Query *query, ParamLi
...
+	ListCell *cteCell = NULL;
+	CmdType cmdType = query->commandType;
+
+	List *copiedList = list_copy(query->cteList);
+
+	foreach(cteCell, copiedList)
+	{
+		CommonTableExpr *cte = (CommonTableExpr *) lfirst(cteCell);
+
+		if ((cte->ctematerialized == CTEMaterializeNever ||
```
onderkalaci: I think at this point we can do `SubqueryColocated(query, cte->ctequery)` to prevent unncessary cte inlining. Because if the query wouldn't yield a colocated one, there is no need to inline the CTE.

That could even convert some intermediate result joins to repartition joins, which we probably wouldn't want at this point.

serprex:
Is there any benefit to having this logic when built against pg12? If not, this should all be wrapped in `#if PG_VERSION_NUM < 120000`

onderkalaci:
> Is there any benefit to having this logic when built against pg12? If not, this should all be wrapped in #if PG_VERSION_NUM < 120000

This is completely isolated than Postgres versions, mostly inspired from the CTE inlining logic that Postgres implemented. 

Currently, Citus always recursively plans the subqueries in every CTE for multi-shard queries (e.g., non-router queries). By re-using the inline logic, we can utilize a lot more advanced subquery logic that Citus has (e.g., subquery pushdown).

onderkalaci:
> Is there any benefit to having this logic when built against pg12? If not, this should all be wrapped in #if PG_VERSION_NUM < 120000


It turns out to be this patch is relying on `QTW_EXAMINE_RTES_AFTER` which is avaliable only in PG 12. Well, we can consider  doing it only for pg 12, or  copy & paste some  code to enable it for pg 11 as well.

marcocitus:
This PR breaks an important class of queries, namely those that use CTEs to work around distributed query planning limitations:

The second query works on master:
```sql
SELECT *, row_number() OVER () FROM test;
ERROR:  could not run distributed query because the window function that is used cannot be pushed down
HINT:  Window functions are supported in two ways. Either add an equality filter on the distributed tables' partition column or use the window functions with a PARTITION BY clause

WITH a AS (SELECT * FROM test) SELECT *, row_number() OVER () FROM a;
ERROR:  could not run distributed query because the window function that is used cannot be pushed down
HINT:  Window functions are supported in two ways. Either add an equality filter on the distributed tables' partition column or use the window functions with a PARTITION BY clause
```

onderkalaci:
> This PR breaks an important class of queries, namely those that use CTEs to work around distributed query planning limitations:


Ops, thanks for finding out quickly.

My proposal is to skip CTE inlining if any of the queries in the queryTree returns non-null for `DeferErrorIfQueryNotSupported()`. Note that we cannot directly check the (sub)query that uses the CTE for `DeferErrorIfQueryNotSupported()`, because that subquery might be pulled up as well.

I know that it is a little bit limiting the pushdown capabilities, but still probably the pushdowns that we prevent would not be realistic queries. This is the rough code for such approach(pushed already): https://github.com/citusdata/citus/pull/3161/commits/0090f87edfc61f039574232f59ae6db3da908bf1

Note that `DeferErrorIfQueryNotSupported()` is preventing any ctes/set operations as well. But, that's too aggressive for our purpose, what I want is the queries that Citus cannot plan in anyway. That's why, I've changed `DeferErrorIfQueryNotSupported() ` to signal such cases, though the code is not good enough yet.

what do you think?

onderkalaci:
>  I know that it is a little bit limiting the pushdown capabilities, but still probably the pushdowns that we prevent would not be realistic queries. 

We can  maybe add one more check that if the  query return non-null for  `DeferErrorIfQueryNotSupported()`  has  at least one of the CTEs  in `cteList` used in it, anywhere in that query tree. If we do that, we'd be allowing inlining CTEs if the non-supported query is at a completely unrelated part of the query. But, to me  that sounds too much code a  pretty small  set of queries that users would run. What do  you think?

src/backend/distributed/planner/cte_inline.c
```diff
@@ -0,0 +1,644 @@
...
+RecursivelyInlineCtesInQueryTree(Query *query)
+{
+	InlineCTEsInQueryTree(query);
+
+	query_tree_walker(query, RecusrivelyInlineCteWalker, NULL, 0);
+}
+
+
+/*
+ * RecusrivelyInlineCteWalker recursively finds all the Query nodes and
```
marcocitus: typo

onderkalaci:
I found another issue, looking into now:
```SQL
-- even if the CTE is not used immediately
-- on a query that Citus could not support,
-- it skips inlining, otherwise the planner
-- might fail
WITH cte_1 AS
  (SELECT *
   FROM test_table)
SELECT row_number() OVER ()
FROM
  (SELECT *
   FROM cte_1) AS foo;
```


In the current patch, we're  only checking the immediate query where the CTE is used. That's not enough such as the above query

marcocitus:
One very conservative option would be to only do the inlining if the user specifies NOT MATERIALIZED.

onderkalaci:
> One very conservative option would be to only do the inlining if the user specifies NOT MATERIALIZED.

I think that wouldn't work in PG 11 because the MATERIALIZED syntax is avaliable in pg 12.

One another conservative option I'm currently working on is to prevent inlining if there is any unsupported query in the query tree. More optimal would  be trace back the query tree from where  the CTE is used in range table, skip inlining  if any is an unsupported query. The latter is harder but  optimizes the number of CTEs inlined. The former is easier to communicate.

Any reservations on doing the former?

onderkalaci:
>  More optimal would be trace back the query tree from where the CTE is used in range table, skip inlining if any is an unsupported query.

I've implemented this  approach, looks simple enough for me. Let me know if you think otherwise.

onderkalaci:
The following queries doesn't work with the patch
```SQL
WITH b AS (SELECT * FROM dist2)
SELECT * FROM  ref LEFT JOIN b ON (a.x = b.x);

WITH a AS (SELECT * FROM dist1),
b AS (SELECT * FROM dist2)
SELECT * FROM  a LEFT JOIN b ON (a.y = b.y);

WITH a AS (SELECT * FROM dist1),
b AS (SELECT * FROM dist2)
SELECT * FROM  a JOIN b ON (a.y > b.y);

```

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -527,14 +543,44 @@ IsModifyDistributedPlan(DistributedPlan *distributedPlan)
...
+		 *
+		 * We also should fallback to distributed planning with non-inlined CTEs
+		 * if the distributed planning fails with inlined CTEs, because recursively
+		 * planning CTEs can provide full SQL coverage, although it might be slow.
+		 */
+		result = InlineCtesAndCreateDistributedPlannedStmt(planId, localPlan,
+														   originalQuery,
+														   query, boundParams,
+														   plannerRestrictionContext);
+		if (result != NULL)
```
marcocitus: might be nicer to return a DistributedPlan with a planning error (the one that was caught)

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -616,6 +662,94 @@ CreateDistributedPlannedStmt(uint64 planId, PlannedStmt *localPlan, Query *origi
...
+											  plannerRestrictionContext);
+	}
+	PG_CATCH();
+	{
+		MemoryContextSwitchTo(savedContext);
+		ErrorData *edata = CopyErrorData();
+		FlushErrorState();
+
+		/* don't try to intercept PANIC or FATAL, let those breeze past us */
+		if (edata->elevel != ERROR)
```
marcocitus: We should probably add error codes to our planner errors such that we could check them here.

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -527,14 +543,44 @@ IsModifyDistributedPlan(DistributedPlan *distributedPlan)
...
+		 * planning some (or all) of the CTEs. In other words, the inlined
+		 * CTEs could become part of query pushdown planning, which is much
+		 * more efficient than recursively planning. So, first try distributed
+		 * planning on the inlined CTEs in the query tree.
+		 *
+		 * We also should fallback to distributed planning with non-inlined CTEs
+		 * if the distributed planning fails with inlined CTEs, because recursively
+		 * planning CTEs can provide full SQL coverage, although it might be slow.
+		 */
+		result = InlineCtesAndCreateDistributedPlannedStmt(planId, localPlan,
```
marcocitus: This looks kind of ok. I don't know whether the feature is worth the extra complexity, but it's something we could live with for a while.

src/test/regress/sql/with_set_operations.sql
```diff
@@ -25,14 +25,14 @@ FROM (
 
 -- cte with unions of other ctes
 WITH
-cte_1 AS (SELECT user_id FROM users_table),
-cte_2 AS (SELECT user_id FROM events_table),
+cte_1 AS (SELECT user_id FROM users_table OFFSET 0),
+cte_2 AS (SELECT user_id FROM events_table OFFSET 0),
```
marcocitus: Why did we need these?

marcocitus:
Would be nice to make separate commits for code changes and tests. My browser has trouble rendering the PR.

onderkalaci:
> Would be nice to make separate commits for code changes and tests. My browser has trouble rendering the PR.

Makes sense, I'll clean the PR and have commits explain why we need to change the tests. 

> Why did we need these?

Once I push the separate commits, I'll explain -- forgot the exact reason per test

src/backend/distributed/planner/distributed_planner.c
onderkalaci: For some users, this change will help a lot, so as far as I can see, it is worth the complexity. I've seen quite a lot of performance gains for some cases, I'll summarize the results as well.

onderkalaci:
> Would be nice to make separate commits for code changes and tests. My browser has trouble rendering the PR.

Done, the commits make more sense now. Just fyi, the code-cov extension might be the reason for that, it slows down mine as well.

src/test/regress/expected/multi_insert_select_conflict.out
```diff
@@ -300,23 +304,23 @@ ALTER TABLE target_table ADD CONSTRAINT fkey FOREIGN KEY (col_1) REFERENCES test
...
+		target_table
+	SELECT
 		col_2,
 		col_1
 	FROM source_table_1 ON CONFLICT (col_1) DO UPDATE SET col_2 = 55 RETURNING *;
-ERROR:  insert or update on table "target_table_1900000" violates foreign key constraint "fkey_1900000"
-DETAIL:  Key (col_1)=(1) is not present in table "test_ref_table_1900012".
-CONTEXT:  while executing command on localhost:57637
+ERROR:  insert or update on table "target_table_1900001" violates foreign key constraint "fkey_1900001"
+DETAIL:  Key (col_1)=(3) is not present in table "test_ref_table_1900012".
```
JelteF: Strange that this detail changes

onderkalaci: That's not related to this change, but how adaptive executor behaves. Depending from which shard the error comes, the error message changes (e.g., the execution order is not deterministic). 

We already ignore shard ids in the regression test output diffs (see src/test/regress/bin/normalize.sed)

 We have `multi_insert_select_conflict.out` test output changes not for this lines, but for the other lines above. However, while copying test output file  on to expected folder, we get such changes as well.

But, you're right, in general, it might be  nicer  not to include these changes in the diffs, but I cannot find an easy way of doing that.

src/test/regress/bin/normalize.sed
```diff
@@ -65,6 +65,14 @@ s/"(ref_table_[0-9]_|ref_table_[0-9]_value_fkey_)[0-9]+"/"\1xxxxxxx"/g
 /^LINE [0-9]+:.*$/d
 /^ *\^$/d
 
+# PG 11 and PG 12 differs in terms of CTE inlining
```
JelteF: let's add `PG_VERSION_NUM < 120000` somewhere in this comment. That way we know we can remove this when we stop supporting PG 11.

src/backend/distributed/shared_library_init.c
```diff
@@ -754,6 +755,20 @@ RegisterCitusConfigVariables(void)
 		GUC_STANDARD,
 		NULL, NULL, NULL);
 
+	DefineCustomBoolVariable(
```
JelteF: Let's also add a comment here to remove this GUC when we stop supporting PG 12, should contain `PG_VERSION_NUM < 120000` somewhere for easy grepping.

src/test/regress/bin/normalize.sed
```diff
@@ -65,6 +65,14 @@ s/"(ref_table_[0-9]_|ref_table_[0-9]_value_fkey_)[0-9]+"/"\1xxxxxxx"/g
...
+# so shadow anything related to subplan ids
+s/generating subplan [0-9]+\_[0-9]+/generating subplan XXX\_XX/g
```
JelteF: Based on the test output, I think the second number is quite stable. Which makes it probably interesting if it changes. So I think this would be better normalization rule:
```suggestion
s/generating subplan [0-9]+\_/generating subplan XXX\_/g
```

Same for this rule below:
```
s/Subplan [0-9]+\_[0-9]+/Subplan XX\_XX/g 
```

JelteF: Actually more of these rules match more than we want. For instance removing InitPlan numbers explain plans, because it matches `Plan [0-9]+]`. So I suggest this change to the normalize script instead.
```sed
# Needed because PG12 and PG11 generated different output, can probably be
# removed when we don't support PG11 anymore, so PG_VERSION_NUM < 120000
s/for CTE/for subquery/g

# Plan numbers are not very stable, subplan numbers are so we keep those
s/DEBUG:  Plan [0-9]+/DEBUG:  Plan XXX/g
s/generating subplan [0-9]+\_/generating subplan XXX\_/g
s/read_intermediate_result\('[0-9]+_/read_intermediate_result('XXX_/g
s/Subplan [0-9]+\_/Subplan XXX\_/g
```

src/backend/distributed/planner/distributed_planner.c
onderkalaci: Do you suggest that as a future improvement or a  required change for this PR?

As far as I can see, the cost of not doing it is to go through distributed planning twice for queries that Citus doesn't support. Are there any additional benefits?

src/test/regress/bin/normalize.sed
onderkalaci: This is already merged into master via your PR, so skipping

src/test/regress/expected/with_modifying.out
```diff
@@ -596,11 +596,7 @@ raw_data AS (
 	DELETE FROM modify_table WHERE id = 1 AND val IN (SELECT val FROM select_data) RETURNING *
 )
 SELECT COUNT(*) FROM raw_data;
- count 
--------
-     2
-(1 row)
-
+ERROR:  set-valued function called in context that cannot accept a set
```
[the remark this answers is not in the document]
onderkalaci: We're hitting an already existing issue by query transformation: https://github.com/citusdata/citus/issues/3189#issuecomment-571497051

I'll fix the query as suggested above for now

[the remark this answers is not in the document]
JelteF: This doesn't really fix the underlying issue right? This PR still has the regression that we stop supporting this pattern in CTEs, so people their queries might stop working if we merge this and they update. 
A workaround for those people would be to disable cte inlining for such a query with `citus.enable_cte_inlining`. 

If we want to go this way, then that should be a concious decision. If not, we should not merge this until we fix #3189 or find a workaround in this PR so we don't hit that bug (e.g. not doing CTE inlining for CTEs with `now` or other functions like that in there)

Changing the query in the test isn't really either of those. It simply means we're ignoring the issue.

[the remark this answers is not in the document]
onderkalaci: > This doesn't really fix the underlying issue right?

No, it doesn't  fix  the issue.

> If we want to go this way, then that should be a concious decision. If not, we should not merge this until we fix #3189 or find a workaround in this PR so we don't hit that bug (e.g. not doing CTE inlining for CTEs with now or other functions like that in there)

To me, the  issue we  hit is an edge case, so I wouldn't mind merging the PR and prioritize fixing the other later on. 

As far as I can  tell, the benefits of this PR is large enough  to ignore this edge case, where there are multiple workarounds to suggest, in case anyone hit by this.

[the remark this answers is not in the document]
onderkalaci: >Changing the query in the test isn't really either of those. It simply means we're ignoring the issue.

I just didn't want to change an  existing  test written for some other purpose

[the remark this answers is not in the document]
JelteF: > As far as I can tell, the benefits of this PR is large enough to ignore this edge case, where there are multiple workarounds to suggest, in case anyone hit by this.

Alright that sounds reasonable. @marcocitus do you agree with that?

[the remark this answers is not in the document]
onderkalaci: > Alright that sounds reasonable. @marcocitus do you agree with that?

I'm also curious about what Marco would think about this.

src/backend/distributed/planner/distributed_planner.c
marcocitus: We don't want to continue in case of e.g. out of memory.

src/test/regress/expected/with_modifying.out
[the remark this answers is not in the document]
JelteF: Like we discussed in person with marco, this seems acceptable. Let's confirm that the workaround works though, by adding the old test back, but with `SET citus.enable_cte_inlining TO FALSE`.

src/test/regress/expected/cte_inline.out
```diff
@@ -0,0 +1,1011 @@
...
+DEBUG:  Router planner cannot handle multi-shard select queries
+ count
+---------------------------------------------------------------------
+(0 rows)
+
+EXECUTE inlined_cte_has_parameter_on_non_dist_key(6);
+DEBUG:  CTE cte_1 is going to be inlined via distributed planning
+DEBUG:  Router planner cannot handle multi-shard select queries
+DEBUG:  CTE cte_1 is going to be inlined via distributed planning
+DEBUG:  Router planner cannot handle multi-shard select queries
```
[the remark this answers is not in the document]
JelteF: Discussed this with marco and it is indeed expected.

src/test/regress/sql/cte_inline.sql
```diff
@@ -0,0 +1,470 @@
...
+WITH cte_1 AS (SELECT * FROM test_table)
+	DELETE FROM test_table WHERE key NOT IN (SELECT key FROM cte_1);
+
+-- we don't inline CTEs if they are modifying CTEs
+WITH cte_1 AS (DELETE FROM test_table RETURNING key)
+SELECT * FROM cte_1;
+
+-- cte with column aliases
+SELECT * FROM test_table,
+(WITH cte_1 (x,y) AS (SELECT * FROM test_table),
```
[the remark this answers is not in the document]
onderkalaci: yes it is on purpose, to make sure that even if Citus doesn't involve in planning a CTE, the CTE could have aliases. Other CTEs have already aliases that are complicated enough

PR #3163, opened 2019-11-07
Escape transaction names

serprex:
While `Assign2PCIdentifier` included *NB: we rely on the fact that we don't need to do full escaping on the names*, it seems best to escape the transaction name regardless

We could end up with a quote slipping through if a user manually inserted data to `pg_dist_transaction` (ParsePreparedTransactionName would at least let it by), though users inserting data into `pg_dist_transaction` is probably outside our threat model

After noticing this, did a search for `'%s'` in codebase. Found an instance in `commands/truncate.c` that seemed suspect. Otherwise all seems fine assuming `workerNode->workerName` can't contain a quote

### DIFF
--- a/src/backend/distributed/commands/truncate.c
+++ b/src/backend/distributed/commands/truncate.c
@@ -25,11 +25,12 @@
 #include "distributed/transaction_management.h"
 #include "distributed/worker_transaction.h"
 #include "storage/lmgr.h"
+#include "utils/builtins.h"
 #include "utils/lsyscache.h"
 #include "utils/rel.h"
 
 
-#define LOCK_RELATION_IF_EXISTS "SELECT lock_relation_if_exists('%s', '%s');"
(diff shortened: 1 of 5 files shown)


src/backend/distributed/commands/truncate.c
```diff
@@ -25,11 +25,12 @@
 #include "distributed/transaction_management.h"
 #include "distributed/worker_transaction.h"
 #include "storage/lmgr.h"
+#include "utils/builtins.h"
 #include "utils/lsyscache.h"
 #include "utils/rel.h"
 
 
-#define LOCK_RELATION_IF_EXISTS "SELECT lock_relation_if_exists('%s', '%s');"
+#define LOCK_RELATION_IF_EXISTS "SELECT lock_relation_if_exists(%s, '%s');"
```
marcocitus: Thanks for catching this!

PR #3164, opened 2019-11-07
Propagate isactive to metadata nodes.

pykello:
We use `isactive` in planners, so not propagating to the worker nodes can result in wrong behaviour. Without this, we can get into replica inconsistency easily. Start metadata sync to node 1, add node 2, then insert into a reference table from node 1. We didn't replicate the change to the replica in node 2 in that case, because node 1 didn't see node 2 as active.

### DIFF
--- a/src/backend/distributed/utils/node_metadata.c
+++ b/src/backend/distributed/utils/node_metadata.c
@@ -273,29 +273,51 @@ master_disable_node(PG_FUNCTION_ARGS)
 	WorkerNode *workerNode = ModifiableWorkerNode(nodeName, nodePort);
 	bool isActive = false;
 	bool onlyConsiderActivePlacements = false;
+	MemoryContext savedContext = CurrentMemoryContext;
 
-	if (WorkerNodeIsPrimary(workerNode))
+	PG_TRY();
 	{
-		/*
-		 * Delete reference table placements so they are not taken into account
-		 * for the check if there are placements after this
-		 */
-		DeleteAllReferenceTablePlacementsFromNodeGroup(workerNode->groupId);
-
-		if (NodeGroupHasShardPlacements(workerNode->groupId,
-										onlyConsiderActivePlacements))
+		if (WorkerNodeIsPrimary(workerNode))
 		{
-			ereport(NOTICE, (errmsg(
-								 "Node %s:%d has active shard placements. Some queries "
-								 "may fail after this operation. Use "
-								 "SELECT master_activate_node('%s', %d) to activate this "
-								 "node back.",
-								 workerNode->workerName, nodePort, workerNode->workerName,
-								 nodePort)));
-		}
-	}
+			/*
+			 * Delete reference table placements so they are not taken into account
+			 * for the check if there are placements after this.
+			 */
+			DeleteAllReferenceTablePlacementsFromNodeGroup(workerNode->groupId);
 
-	SetNodeState(nodeName, nodePort, isActive);
+			if (NodeGroupHasShardPlacements(workerNode->groupId,
+											onlyConsiderActivePlacements))
+			{
+				ereport(NOTICE, (errmsg(
+									 "Node %s:%d has active shard placements. Some queries "
+									 "may fail after this operation. Use "
+									 "SELECT master_activate_node('%s', %d) to activate this "
+									 "node back.",
+									 workerNode->workerName, nodePort,
+									 workerNode->workerName,
+									 nodePort)));
+			}
+		}
+
+		SetNodeState(nodeName, nodePort, isActive);
+	}
+	PG_CATCH();
+	{
+		ErrorData *edata = NULL;
+
+		/* CopyErrorData() requires (CurrentMemoryContext != ErrorContext) */
+		MemoryContextSwitchTo(savedContext);
+		edata = CopyErrorData();
+
+		ereport(ERROR, (errcode(ERRCODE_OBJECT_NOT_IN_PREREQUISITE_STATE),
+						errmsg("Disabling %s:%d failed", workerNode->workerName,
+							   nodePort),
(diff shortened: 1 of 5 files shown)


src/backend/distributed/utils/node_metadata.c
```diff
@@ -1237,8 +1238,35 @@ static WorkerNode *
 SetNodeState(char *nodeName, int nodePort, bool isActive)
 {
 	WorkerNode *workerNode = FindWorkerNodeAnyCluster(nodeName, nodePort);
-	return SetWorkerColumn(workerNode, Anum_pg_dist_node_isactive,
-						   BoolGetDatum(isActive));
+	WorkerNode *newWorkerNode = SetWorkerColumn(workerNode,
+												Anum_pg_dist_node_isactive,
+												BoolGetDatum(isActive));
+	PropagateNodeStateMetadata(workerNode->nodeId, isActive);
```
onderkalaci: Hmm, this feels similar to #2320.

If we do this, we won't be able to add a new node when any node is down in the cluster. I wonder whether we should do it similar to `TriggerMetadataSync()`, which tries later and skips as long as the nodes are down.
 
I think until the metadata synced, we'd not be able to utilize the reference table shard placements on the coordinator, which is very much accepted? 

Are there anything else that'd not work until we sync `isactive `?

pykello: > Hmm, this feels similar to #2320.

Delaying `hasmetadata` propagation wouldn't result in data inconsistency. Delaying `isactive` propagation creates a window for data inconsistency if we do DML from metadata nodes. It seems that we don't route DML to inactive nodes.

>  If we do this, we won't be able to add a new node when any node is down in the cluster. 

This is already disallowed because `master_add_node()` and `master_add_inactive_node()` send `INSERT INTO pg_dist_node` commands to workers right away.

> I think until the metadata synced, we'd not be able to utilize the reference table shard placements on the coordinator, which is very much accepted?

This is not reference table or coordinator specific. If we add a node and then create a distributed table, and then insert into it from the worker node, then I think INSERT from metadata worker won't modify the placements on the not-yet-active node.

pykello: An example is:

```sql
-- coordinator:
select start_metadata_sync_to_node('localhost', 5433);                             
select master_activate_node('localhost', 5434);
create table t (a int, b int);
select create_distributed_table('t', 'a');

-- localhost:5433, before isactive is synced
postgres=# insert into t values (3);
ERROR:  no active placements were found for shard 102055
```

You might be fine for single-replicated tables, but imagine what happens for reference tables.

marcocitus: Does master_disable_node work if a node is down? It is generally used for nodes that are down.

I agree that master_activate_node is similar to master_add_node, so requiring nodes to be up seems acceptable.

src/backend/distributed/utils/node_metadata.c
```diff
@@ -549,7 +550,7 @@ PrimaryNodeForGroup(int32 groupId, bool *groupContainsNodes)
  * given node.
  */
 static int
-ActivateNode(char *nodeName, int nodePort)
+ActivateNode(char *nodeName, int nodePort, int nodeId)
```
onderkalaci: Minor: `ActivateNode ()` doesn't use `nodeId` so we can undo this change

src/backend/distributed/utils/node_metadata.c
onderkalaci: > This is already disallowed because master_add_node() and master_add_inactive_node() send INSERT INTO pg_dist_node commands to workers right away.

Yes, sorry I missed that but thanks marco for pointing the below:

> Does master_disable_node work if a node is down?

No, it'd fail if the cluster is MX and there is at least one reference table. `master_disable_node()` -> `DeleteAllReferenceTablePlacementsFromNodeGroup()` -> `BlockWritesToShardList()` and `master_disable_node()` -> `DeleteAllReferenceTablePlacementsFromNodeGroup()`-> `SendCommandToWorkers(DELETE FROM pg_dist_placement WHERE placementid)`

I think with this PR, we're adding one more command to the above. `master_disable_node()` -> `SetNodeState()` -> `PropagateNodeStateMetadata()` -> `SendCommandToWorkers()`

@marcocitus how critical do you see that?

src/backend/distributed/utils/node_metadata.c
pykello: using it again.

src/backend/distributed/utils/node_metadata.c
pykello: Fixed this case but not doing the pg_dist_placement changes synchronously. Also added some tests.

I'm sending a meeting invite to discuss what else we could do if you don't agree with these changes.

pykello:
@marcocitus @onderkalaci as per meeting conclusion, `master_disable_node()` also does the metadata propagation synchronously, except that it hints the users that they should `stop_metadata_sync_to_node()` if an error happens.

src/backend/distributed/utils/node_metadata.c
```diff
@@ -273,29 +273,50 @@ master_disable_node(PG_FUNCTION_ARGS)
...
+			 * Delete reference table placements so they are not taken into account
+			 * for the check if there are placements after this.
+			 */
+			DeleteAllReferenceTablePlacementsFromNodeGroup(workerNode->groupId);
+
+			if (NodeGroupHasShardPlacements(workerNode->groupId,
+											onlyConsiderActivePlacements))
+			{
+				ereport(NOTICE, (errmsg(
+									 "Node %s:%d has active shard placements. Some queries "
```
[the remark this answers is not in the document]
pykello: keeping this for another PR.

PR #3165, opened 2019-11-07
remove duplicate code in citus_dist_stat_activity

SaitTalhaNisanci:


src/backend/distributed/transaction/citus_dist_stat_activity.c
```diff
@@ -253,6 +253,8 @@ static TimestampTz ParseTimestampTzFieldFromHeapTuple(HeapTuple tuple, TupleDesc
 													  int colIndex);
 static TransactionId ParseXIDFieldFromHeapTuple(HeapTuple tuple, TupleDesc tupdesc, int
 												colIndex);
+static void ReplaceInitiatorNodeIdentifier(int initiator_node_identifier,
```
onderkalaci: Super minor coding convention notes:

We usually prefer to declare & define the function right after the first function that uses them. In this case, it'd be good to move the declaration right after `ParseCitusDistStat `. As far as I can see, the definition is already on the correct place.

src/backend/distributed/transaction/citus_dist_stat_activity.c
```diff
@@ -487,55 +489,11 @@ ParseCitusDistStat(PGresult *result, int64 rowIndex)
...
-	{
-		citusDistStat->master_query_host_name = cstring_to_text(coordinator_host_name);
-		citusDistStat->master_query_host_port = 0;
-	}
-	else
-	{
-		citusDistStat->master_query_host_name = NULL;
-		citusDistStat->master_query_host_port = 0;
-	}
+	ReplaceInitiatorNodeIdentifier(initiator_node_identifier, citusDistStat);
```
onderkalaci: Again super minor convention note, feel free to ignore: We generally try to have a single empty line between code-blocks that are logically different than others.

For example, when you use  `ReplaceInitiatorNodeIdentifier ` is feels a little bit different than `PQgetisnull ` / `ParseIntFieldFromHeapTuple `, so might be good to have a single line for readability purpose.

Try it out, if that doesn't feel similar for you, skip doing it.

src/backend/distributed/transaction/citus_dist_stat_activity.c
SaitTalhaNisanci: putting definition right after the usage makes it easy to follow the flow of the code, but what benefit does having declaration in the same order?

(I have updated the declaration too)

onderkalaci: I think you can trace the related functions a bit easier

PR #3168, opened 2019-11-07
Remove citus.worker_list_file & master_initialize_node_metadata

### LINKED ISSUE #848: Remove pg_worker_list.conf related Code and GUC
With #786 is closed by #798, we no longer use pg_worker_list.conf. However, for upgrade purposes, we left some portions of the code about parsing pg_worker_list.conf file:
- `citus.worker_list_file` GUC
- `master_initialize_node_metadata` UDF (called automatically after update to 6.0)

These should be removed from the code after we decide not to support upgrades from Citus 5.

serprex:
DESCRIPTION: remove `citus.worker_list_file` GUC

Also removes `master_initialize_node_metadata`

Not sure best way to remove a function. I've updated previous version scripts, could cause issue if someone were to upgrade to 8.1 from 7.0. Other option I thought of would be leaving an empty function definition to keep the symbol around

Fixes #848

### DIFF
--- a/src/backend/distributed/shared_library_init.c
+++ b/src/backend/distributed/shared_library_init.c
@@ -81,7 +81,6 @@ static void RegisterCitusConfigVariables(void);
 static bool ErrorIfNotASuitableDeadlockFactor(double *newval, void **extra,
 											  GucSource source);
 static bool WarnIfDeprecatedExecutorUsed(int *newval, void **extra, GucSource source);
-static void NormalizeWorkerListPath(void);
 static bool NodeConninfoGucCheckHook(char **newval, void **extra, GucSource source);
 static void NodeConninfoGucAssignHook(const char *newval, void *extra);
 static bool StatisticsCollectionGucCheckHook(bool *newval, void **extra, GucSource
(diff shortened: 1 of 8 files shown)


src/backend/distributed/shared_library_init.c
```diff
@@ -380,18 +379,6 @@ RegisterCitusConfigVariables(void)
 		GUC_UNIT_MS | GUC_STANDARD,
 		NULL, NULL, NULL);
 
-	/* keeping temporarily for updates from pre-6.0 versions */
```
onderkalaci: I'm not the best person to ask, but I think we prefer not to remove GUCs on minor release versions. @metdos is that correct?

src/backend/distributed/sql/citus--9.0-2--9.1-1.sql
```diff
@@ -8,3 +8,7 @@ COMMENT ON COLUMN pg_catalog.pg_dist_node.shouldhaveshards IS
 -- we don't maintain replication factor of reference tables anymore and just
 -- use -1 instead.
 UPDATE pg_dist_colocation SET replicationfactor = -1 WHERE distributioncolumntype = 0;
+
+-- drop function which was used for upgrading from 6.0
+-- creation was removed from citus--7.0-1.sql
+DROP FUNCTION IF EXISTS master_initialize_node_metadata;
```
marcocitus: Should to be careful about search_path here. We've made mistakes where because we use "IF EXISTS", we did not notice that we were not dropping the function. That caused pg_upgrade to fail, since pg_upgrade will try to re-create the function the new database, but that will fail if the C implementation is missing.

src/backend/distributed/shared_library_init.c
marcocitus: I think this GUC has had enough time to no longer be used.

src/backend/distributed/sql/citus--9.0-2--9.1-1.sql
serprex: Updated to qualified, `DROP FUNCTION IF EXISTS pg_catalog.master_initialize_node_metadata;`

PR #3169, opened 2019-11-08
Add adaptive executor support for repartition joins

SaitTalhaNisanci:
Description: Add adaptive executor support for repartition joins

This PR implements repartitioning support for adaptive executor. Repartition queries were supported by only Task tracker. This PR does not implement the logic for replication factor > 1. In that case Task tracker will be chosen by the planner.

Adaptive executor repartition's main entry point is ExecuteDependentTasks.
ExecuteDependentTasks takes a list of top level tasks. Its logic is as follows:
 - It generates all the tasks by descending in the tasks tree. Note that each task
has a dependedTaskList.
- It creates schemas in each worker in a single transaction to store intermediate results. For this, it finds the unique `jobIds` and and creates the schemas in one network latency per worker.
- It iterates all tasks and finds the ones whose dependencies are already executed, and executes them with adaptive executor logic.
- It removes temporary job directories and schemas.

When executing repartition queries, same connections are used if they are not in a transaction, otherwise new connections are opened. Also repartition queries do not begin a transaction even if we are in a transaction block. As we dont begin a transaction, they wont see the DDLs that happened earlier in the transaction because we dont have that transaction id with repartition queries. Therefore we error in this case.

When replication factor is 2, two tests fails in our test suite, which says that the relation( which is created to store the temporary results) already exists. Currently if replication factor is greater than 1, task tracker executor is chosen. This could be a nice to have to fix(Only 2 tests fail with replication factor > 1). [ ] ( We wont implement replication factor > 2 in this PR)

~~Currently it is draft because~~:
- [x]  Some tests fail
- [x] Some function comments are missing
- [x] Memory leak exists( No freeing is done currently) Well it turns out that in this case we dont need to do freeing :) Every allocation in this PR is done at the CurrentContextMemory therefore it doesnt leak anything and we do not allocate big things.
- [x] Repartition join tests for adaptive executor are missing
- [x] `citus.enable_repartition_joins` guc should be removed? (No we wont remove it, repartition joins with adaptive executor will only work when `citus.enable_repartition_joins` guc is set to true)
- [x] This description should be updated with how it works.


src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -1170,7 +1193,9 @@ TaskListRequires2PC(List *taskList)
 static bool
 ReadOnlyTask(TaskType taskType)
 {
-	if (taskType == ROUTER_TASK || taskType == SQL_TASK)
+	if (taskType == ROUTER_TASK || taskType == SQL_TASK || taskType ==
+		MAP_OUTPUT_FETCH_TASK ||
+		taskType == MAP_TASK || taskType == MERGE_TASK)
```
marcocitus: Maybe make it a case statement.

Minor: I could see us end up wanting to re-partition the output of a DELETE .. RETURNING at some point. I wonder whether we should start keeping an explicit readOnly flag.

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -597,9 +607,11 @@ static bool ShouldMarkPlacementsInvalidOnFailure(DistributedExecution *execution
 static void PlacementExecutionReady(TaskPlacementExecution *placementExecution);
 static TaskExecutionState TaskExecutionStateMachine(ShardCommandExecution *
 													shardCommandExecution);
+static bool DoesHaveDependedJobs(Job *mainJob);
 static void ExtractParametersForRemoteExecution(ParamListInfo paramListInfo,
 												Oid **parameterTypes,
 												const char ***parameterValues);
+static void LogSessionId(uint64 sessionId);
```
SaitTalhaNisanci: I found this useful during debugging. We know what command is sent over what host-port but we dont know what session is used for that.

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -2100,11 +2205,21 @@ ManageWorkerPool(WorkerPool *workerPool)
 		/* experimental: just to see the perf benefits of caching connections */
 		int connectionFlags = 0;
```
SaitTalhaNisanci: Probably here we should have a method that takes execution and returns connection flags.

src/backend/distributed/executor/multi_server_executor.c
```diff
@@ -103,12 +105,11 @@ JobExecutorType(DistributedPlan *distributedPlan)
...
 			}
-
-			ereport(DEBUG1, (errmsg(
-								 "cannot use adaptive executor with repartition jobs"),
-							 errhint("Since you enabled citus.enable_repartition_joins "
-									 "Citus chose to use task-tracker.")));
-			return MULTI_EXECUTOR_TASK_TRACKER;
+			if (ShardReplicationFactor > 1)
+			{
+				return MULTI_EXECUTOR_TASK_TRACKER;
```
SaitTalhaNisanci: Here we should probably add a log message

src/backend/distributed/worker/worker_merge_protocol.c
```diff
@@ -53,6 +57,49 @@ static void CopyTaskFilesFromDirectory(StringInfo schemaName, StringInfo relatio
...
+Datum
+worker_create_schema(PG_FUNCTION_ARGS)
+{
+	uint64 jobId = PG_GETARG_INT64(0);
+
+	StringInfo jobSchemaName = JobSchemaName(jobId);
+	bool schemaExists = false;
+	CheckCitusVersion(ERROR);
+
+	LockJobResource(jobId, AccessExclusiveLock);
```
marcocitus: I think this lock is no longer useful because we explicitly only call this function once per job.

src/backend/distributed/worker/worker_merge_protocol.c
```diff
@@ -53,6 +57,49 @@ static void CopyTaskFilesFromDirectory(StringInfo schemaName, StringInfo relatio
...
+ */
+Datum
+worker_remove_jobdir(PG_FUNCTION_ARGS)
+{
+	uint64 jobId = PG_GETARG_INT64(0);
+	StringInfo jobDirectoryName = JobDirectoryName(jobId);
+
+	CheckCitusVersion(ERROR);
+
+	CitusRemoveDirectory(jobDirectoryName);
```
marcocitus: This is kind of dangerous, since any user can call it. We should check whether the user owns the schema.

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -277,8 +281,13 @@ typedef struct DistributedExecution
 	 */
 	AttInMetadata *attributeInputMetadata;
 	char **columnArray;
+
+	bool isRepartition;
+
+	List *forcedConnections;
```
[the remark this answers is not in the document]
SaitTalhaNisanci: I guess for an execution, we will have only one type of task and they will be executed either inside or outside a transaction, so when it is outside a transaction session list will be equal to forcedConnections. That is why you find it unnecessary right?

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -758,6 +789,22 @@ ExecuteUtilityTaskListWithoutResults(List *taskList)
 }
 
 
+/*
+ * ExecuteTaskListRepartiton is a proxy to ExecuteTaskListExtended() with defaults
+ * for some of the arguments for a repartition query.
+ */
+uint64
+ExecuteTaskListRepartition(RowModifyLevel modLevel, List *taskList, int targetPoolSize)
```
marcocitus: Maybe move this into the re-partitioning-specific file, or give it a more generic name.

src/backend/distributed/executor/adaptive_executor_repartition.c
```diff
@@ -0,0 +1,519 @@
+/*-------------------------------------------------------------------------
+ *
+ * adaptive_executor_repartition.c
```
marcocitus: I think the adaptive executor is a bit of an implementation detail since you're only interacting with it via ExecuteTaskList (which originally comes from the router executor) and are also sending commands outside of the adaptive executor. 

Is there a better way to describe what we're doing here? Is it even specific to re-partitioning?

src/backend/distributed/executor/adaptive_executor.c
SaitTalhaNisanci: @marcocitus What do you think of this?

src/backend/distributed/executor/adaptive_executor.c
SaitTalhaNisanci: `ExecuteTaskListOutsideTransaction` seems like a better name

src/backend/distributed/executor/adaptive_executor_repartition.c
SaitTalhaNisanci: umm currently this file is kind of specific to repartitioning because it has some repartition specific things such as generating FETCH_TASK query string from MAP_TASK( however currently this could well be done in the planner step, it is currently left empty because we dont know in which placement the map task succeeded in case of replication factor > 1, however we dont even use adaptive executor for repartition joins if that is the case.) 

There are other things such as creating schemas etc which are also repartition specific, so I guess in general I would say this file is repartition specific. So I will try to give a better description here

marcocitus: I'm wondering whether we can split up the file into "directed acyclic graph execution logic" and "re-partitioning logic". Also, I think we probably do not need really call it "adaptive" because that's an implementation detail and soon-ish it will be the only way to execute re-partition joins.

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -2982,6 +3092,13 @@ StartPlacementExecutionOnSession(TaskPlacementExecution *placementExecution,
 }
 
 
+static void
+LogSessionId(uint64 sessionId)
+{
+	ereport(DEBUG4, (errmsg("Command will be sent on session id: %ld", sessionId)));
```
marcocitus: It would be nicer if this was part of the same line. Maybe we should use "connection ID" (stored in MultiConnection) instead of a session ID everywhere?

src/backend/distributed/executor/adaptive_executor_repartition.c
SaitTalhaNisanci: Makes sense.

src/backend/distributed/executor/adaptive_executor.c
SaitTalhaNisanci: You mean we should create connectionID for MultiConnection? Because as far as I see, it doesnt a connection does not have an id.

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -277,8 +281,11 @@ typedef struct DistributedExecution
 	 */
 	AttInMetadata *attributeInputMetadata;
 	char **columnArray;
+
+	bool isOutsideTransaction;
```
marcocitus: Needs a comment

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -596,9 +604,11 @@ static bool ShouldMarkPlacementsInvalidOnFailure(DistributedExecution *execution
 static void PlacementExecutionReady(TaskPlacementExecution *placementExecution);
 static TaskExecutionState TaskExecutionStateMachine(ShardCommandExecution *
 													shardCommandExecution);
+static bool DoesHaveDependedJobs(Job *mainJob);
```
marcocitus: HasDependedJobs

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -618,6 +628,7 @@ AdaptiveExecutor(CitusScanState *scanState)
 	bool randomAccess = true;
 	bool interTransactions = false;
 	int targetPoolSize = MaxAdaptiveExecutorPoolSize;
+	static bool hasDependedJobsInAllTree = false;
```
marcocitus: What does this variable do? In general, we should not keep state in static variables except when it relates to the process itself (e.g. a counter since the start of the process).

src/backend/distributed/sql/udfs/worker_create_schema/latest.sql
```diff
@@ -0,0 +1,13 @@
+CREATE FUNCTION pg_catalog.worker_create_schema(bigint)
+    RETURNS void
+    LANGUAGE C STRICT
+    AS 'MODULE_PATHNAME', $$worker_create_schema$$;
+COMMENT ON FUNCTION pg_catalog.worker_create_schema(bigint)
+    IS 'create schema in remote node';
+
+CREATE FUNCTION pg_catalog.worker_remove_jobdir(bigint)
```
marcocitus: Either move these into the upgrade script or use 1 file per function.

src/backend/distributed/planner/multi_physical_planner.c
```diff
@@ -4462,11 +4462,22 @@ MergeTaskList(MapMergeJob *mapMergeJob, List *mapTaskList, uint32 taskIdIndex)
 		{
 			Task *mapTask = (Task *) lfirst(mapTaskCell);
 
-			/* we need node names for the query, and we'll resolve them later */
-			char *undefinedQueryString = NULL;
+			/* find the node name/port for map task's execution */
+			List *mapTaskPlacementList = mapTask->taskPlacementList;
+
+			ShardPlacement *mapTaskPlacement = linitial(mapTaskPlacementList);
```
marcocitus: This could use more elaboration. Specifically that we're optimistically assuming that we'll be able to use the first placement. In the task-tracker executor we may fail over to a different placement and rebuild the query string.

src/backend/distributed/executor/repartition_logic.c
```diff
@@ -0,0 +1,252 @@
...
+/*
+ * SendCommandToAllWorkers sends the given command to all workers in
+ * a single transaction.
+ */
+static void
+SendCommandToAllWorkers(List *commandList)
+{
+	ListCell *workerNodeCell = NULL;
+	char *extensionOwner = CitusExtensionOwnerName();
+	List *workerNodeList = ActiveReadableNodeList();
```
marcocitus: This should probably be ActivePrimaryWorkerNodeList. We cannot create schemas on read-only nodes.

src/backend/distributed/executor/repartition_logic.c
```diff
@@ -0,0 +1,252 @@
...
+	SendCommandToAllWorkers(commandList);
+}
+
+
+/*
+ * SendCommandToAllWorkers sends the given command to all workers in
+ * a single transaction.
+ */
+static void
+SendCommandToAllWorkers(List *commandList)
```
marcocitus: Should maybe be in worker_transaction.c

src/backend/distributed/executor/repartition_logic.c
```diff
@@ -0,0 +1,252 @@
...
+
+/*
+ * GenerateCommand returns concatenated commands with the given template
+ * command for each job id from the given job ids. The returned command is
+ * exactly list_length(jobIds) subcommands.
+ *  E.g create_schema(jobId1); create_schema(jobId2); ...
+ * This way we can send the command in just one latency to a worker.
+ */
+static char *
+GenerateCommand(List *jobIds, char *templateCommand)
```
marcocitus: This name seems too generic. Maybe GenerateJobCommands.

src/backend/distributed/executor/repartition_logic.c
```diff
@@ -0,0 +1,252 @@
...
+	List *commandList = list_make1(GenerateDeleteJobsCommand(jobIds));
+	SendCommandToAllWorkers(commandList);
+}
+
+
+/*
+ * GenerateDeleteJobsCommand returns concatanated remove job dir commands.
+ */
+static char *
+GenerateDeleteJobsCommand(List *jobIds)
```
marcocitus: Would be nice to move this next to GenerateCreateSchemasCommand

src/backend/distributed/executor/repartition_logic.c
```diff
@@ -0,0 +1,252 @@
...
+
+	RemoveTempJobDirs(jobIds);
+}
+
+
+/*
+ * FillTaskGroups iterates all tasks and creates a group for mergeTasks.
+ */
+static void
+FillTaskGroups(List **allTasks, List **mergeTasks)
```
marcocitus: This API seems very weird. What about `List * ExtractMergeTasks(List *allTasks)`? If we need to concatenate the result then use list_concat.

src/backend/distributed/executor/repartition_logic.c
```diff
@@ -0,0 +1,252 @@
...
+static List *
+CreateJobIds(List *mergeTasks)
+{
+	ListCell *taskCell = NULL;
+	List *jobIds = NIL;
+
+	foreach(taskCell, mergeTasks)
+	{
+		Task *task = (Task *) lfirst(taskCell);
+		if (!DoesJobIDExist(jobIds, task->jobId))
```
marcocitus: jobIds = list_append_unique_int(jobIds, task->jobIds);

src/backend/distributed/executor/repartition_logic.c
```diff
@@ -0,0 +1,252 @@
...
+	return jobIds;
+}
+
+
+/*
+ * CreateJobIds returns a list of unique job ids that will be used
+ * in mergeTasks.
+ */
+static List *
+CreateJobIds(List *mergeTasks)
```
marcocitus: Maybe `ExtractJobIdsFromTasks`

src/backend/distributed/executor/repartition_logic.c
```diff
@@ -0,0 +1,252 @@
...
+	}
+}
+
+
+/*
+ * CreateTemporarySchemas creates the necessary schemas that will be used
+ * later in each worker. Single transaction is used to create the schemas.
+ */
+static List *
+CreateTemporarySchemas(List *mergeTasks)
```
marcocitus: Maybe `CreateTemporarySchemasForMergeTasks`

src/backend/distributed/executor/repartition_logic.c
```diff
@@ -0,0 +1,252 @@
...
+
+/*
+ * CreateSchemasOnAllWorkers creates schemas in all workers.
+ */
+static void
+CreateSchemasOnAllWorkers(char *createSchemasCommand)
+{
+	List *commandList = list_make1(createSchemasCommand);
+
+	SendCommandToAllWorkers(commandList);
```
marcocitus: The implementation on the function doesn't match its name. This seems to be an `SendCommandToAllWorkers` whereas `SendCommandToAllWorkers` should be `SendCommandListToAllWorkers`.

src/backend/distributed/executor/directed_acylic_graph_execution_logic.c
```diff
@@ -0,0 +1,228 @@
...
+}
+
+
+static uint32
+TaskHash(const void *key, Size keysize)
+{
+	TaskHashKey *taskKey = (TaskHashKey *) key;
+	uint32 hash = 0;
+
+	hash = hash_combine(hash, hash_uint32((uint32) taskKey->jobId));
```
marcocitus: Should we try to preserve the 64-bitness here?

src/backend/distributed/executor/directed_acylic_graph_execution_logic.c
```diff
@@ -0,0 +1,228 @@
...
+	}
+	return true;
+}
+
+
+static void
+FillTaskKey(TaskHashKey *taskKey, Task *task)
+{
+	taskKey->jobId = task->jobId;
+	taskKey->taskId = task->taskId;
```
marcocitus: Why not just inline this instead of having the extra code?

TaskHashKey taskKey = { task->jobId, task->taskId };

src/backend/distributed/executor/repartition_logic.c
marcocitus: ah, I guess the problem is that it's int64

src/backend/distributed/executor/adaptive_executor.c
SaitTalhaNisanci: If we have depended jobs in any part of the tree then we know that we created temporary schemas. Therefore I have used this to determine that, if so we run the cleanup logic at the top level.

src/backend/distributed/executor/repartition_logic.c
SaitTalhaNisanci: I have used `ActivePrimaryWorkerNodeList(NoLock);`

src/backend/distributed/executor/repartition_logic.c
SaitTalhaNisanci: I have placed it after `RemoveTempJobDirs` since it is called in `RemoveTempJobDirs`. I think when someone is looking at `RemoveTempJobDirs` they will want to look at its methods and it is easier to see the methods it calls if they are placed right after it.

We will probably not want to look at `GenerateDeleteJobsCommand` when we are looking at `GenerateCreateSchemasCommand`.

What do you think?

src/backend/distributed/executor/repartition_logic.c
SaitTalhaNisanci: My bad, it is a leftover name from a previous version. Now It makes no sense. I will change it to ExtractMergeTasks.

src/backend/distributed/executor/repartition_logic.c
SaitTalhaNisanci: I guess we cannot use that here because jobIds are uint64 not int.

SaitTalhaNisanci: I guess it would be nice if Github had a feature like "there is a new comment" if I am sending a comment and a new one is added since the last time I refreshed the page.

SaitTalhaNisanci: Well I have added a utility function for uint64, might be useful in future.

src/backend/distributed/executor/directed_acylic_graph_execution_logic.c
SaitTalhaNisanci: hmm at first I couldnt find a method like hash_uint64 but I guess we do it this way :

```
hash = hash_combine(hash, hash_any((unsigned char *) &taskKey->jobId,
									   sizeof(int64)));
```

src/backend/distributed/executor/repartition_logic.c
```diff
@@ -142,28 +138,7 @@ CreateSchemasOnAllWorkers(char *createSchemasCommand)
...
-	List *workerNodeList = ActiveReadableNodeList();
-
-	foreach(workerNodeCell, workerNodeList)
-	{
-		WorkerNode *workerNode = (WorkerNode *) lfirst(workerNodeCell);
-		SendCommandListToWorkerInSingleTransaction(workerNode->workerName,
-												   workerNode->workerPort, extensionOwner,
-												   commandList);
-	}
+	SendCommandListToAllWorkers(commandList);
```
marcocitus: This function sends commands as superuser, which is undesirable. We may want to add a user argument to it.

src/backend/distributed/executor/repartition_logic.c
```diff
@@ -142,28 +138,7 @@ CreateSchemasOnAllWorkers(char *createSchemasCommand)
 {
```
marcocitus: Maybe rename to SendCommandToAllWorkers.

src/backend/distributed/executor/repartition_logic.c
```diff
@@ -238,7 +194,7 @@ static void
 RemoveTempJobDirs(List *jobIds)
 {
 	List *commandList = list_make1(GenerateDeleteJobsCommand(jobIds));
-	SendCommandToAllWorkers(commandList);
+	SendCommandListToAllWorkers(commandList);
```
marcocitus: Maybe `SendCommandToAllWorkers(GenerateDeleteJobsCommand(jobIds));`

src/backend/distributed/transaction/worker_transaction.c
```diff
@@ -117,6 +117,27 @@ SendCommandToWorkersWithMetadata(const char *command)
...
 
+/*
+ * SendCommandListToAllWorkers sends the given command to all workers in
+ * a single transaction.
+ */
+void
+SendCommandListToAllWorkers(List *commandList)
+{
+	ListCell *workerNodeCell = NULL;
+	char *extensionOwner = CitusExtensionOwnerName();
```
marcocitus: Note: This is always superuser, so our ownership check in worker_remove_jobdir would not work.

src/backend/distributed/executor/repartition_logic.c
marcocitus: Thinking about it more, sending this as superuser is useful because it allows us to revoke all access to the UDFs.

src/backend/distributed/executor/adaptive_executor.c
marcocitus: This seems like a hack. Can you come up with a different way of doing that?

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -707,6 +736,13 @@ AdaptiveExecutor(CitusScanState *scanState)
 }
 
 
+static bool
+HasDependedJobs(Job *mainJob)
```
marcocitus: Missing comment. What's a depended job?

(minor: this seems like a misspelling of dependentJob, we should probably stop misspelling it)

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -1346,8 +1440,14 @@ CleanUpSessions(DistributedExecution *execution)
 			 * We cannot get MULTI_CONNECTION_LOST via the ConnectionStateMachine,
 			 * but we might get it via the connection API and find us here before
 			 * changing any states in the ConnectionStateMachine.
+			 *
+			 * If execution is outside a transaction, then the connection will be shutdown later so
+			 * we can skip it here.
 			 */
-			CloseConnection(connection);
+			if (!execution->isOutsideTransaction)
+			{
```
marcocitus: Could we just let it close here? (We should perhaps introduce a MULTI_CONNECTION_CLOSED)

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -1308,6 +1382,26 @@ FinishDistributedExecution(DistributedExecution *execution)
...
+{
+	ListCell *sessionCell = NULL;
+
+	foreach(sessionCell, sessionList)
+	{
+		WorkerSession *session = lfirst(sessionCell);
+		MultiConnection *con = session->connection;
+
+		UnclaimConnection(con);
+		con->remoteTransaction.transactionState = REMOTE_TRANS_INVALID;
```
marcocitus: do we need this line?

src/backend/distributed/executor/adaptive_executor.c
marcocitus: yes, could be useful to add some static counter to StartConnectionEstablishment

src/backend/distributed/executor/directed_acylic_graph_execution_logic.c
```diff
@@ -0,0 +1,218 @@
+/*-------------------------------------------------------------------------
+ *
+ * directed_acylic_graph_execution_logic.c
```
marcocitus: Maybe remove the "_logic". Presumably most C files contain some kind of logic.

src/backend/distributed/executor/repartition_logic.c
```diff
@@ -0,0 +1,208 @@
...
+ * different jobs.
+ */
+void
+ExecuteDependedTasks(List *topLevelTasks)
+{
+	EnsureNoModificationsHaveBeenDone();
+
+	List *allTasks = TaskAndExecutionList(topLevelTasks);
+
+	List *mergeTasks = ExtractMergeTasks(allTasks);
```
marcocitus: I think the point of this is ultimately just to get a set of job IDs. If we passed the Job to ExecuteDependedTasks. Would we be able to get the job IDs in a cheaper way?

src/backend/distributed/sql/udfs/worker_remove_jobdir/latest.sql
```diff
@@ -0,0 +1,6 @@
+CREATE FUNCTION pg_catalog.worker_remove_jobdir(bigint)
+    RETURNS void
+    LANGUAGE C STRICT
+    AS 'MODULE_PATHNAME', $$worker_remove_jobdir$$;
+COMMENT ON FUNCTION pg_catalog.worker_remove_jobdir(bigint)
+    IS 'remove job in remote node';
```
marcocitus: since we're running these as superuser, we can revoke access from all other users

src/test/regress/expected/adaptive_executor_repartition.out
```diff
@@ -0,0 +1,61 @@
...
+
+ROLLBACK;
+BEGIN;
+INSERT INTO ab values(1, 2);
+-- DDL happened before repartition query in a transaction block, so this should error.
+SELECT count(*) FROM (SELECT k.a FROM ab k, ab l WHERE k.a = l.b) first, (SELECT * FROM ab) second WHERE first.a = second.b;
+ERROR:  cannot open new connections after the first modification command within a transaction
+ROLLBACK;
+DROP SCHEMA adaptive_executor CASCADE;
+NOTICE:  drop cascades to table ab
```
marcocitus: It would be useful to add some single hash repartition joins.

citus.enable_single_hash_repartition_joins is disabled by default.

src/backend/distributed/executor/repartition_logic.c
SaitTalhaNisanci: Rename what to `SendCommandToAllWorkers` ?

src/backend/distributed/executor/adaptive_executor.c
SaitTalhaNisanci: https://github.com/citusdata/citus/pull/3244/files

src/backend/distributed/executor/repartition_logic.c
SaitTalhaNisanci: IIRC jobs do not contain any dependent jobs. Tasks contain dependent tasks. This can be fixed in the planner part I guess. Then we can get the job ids from by traversing the top job.

src/backend/distributed/executor/adaptive_executor.c
SaitTalhaNisanci: What value would introducing `MULTI_CONNECTION_CLOSED`  have? We are freeing the connection in https://github.com/citusdata/citus/blob/master/src/backend/distributed/connection/connection_management.c#L458 while closing the connection. So even if we set that value it will not be `MULTI_CONNECTION_CLOSED`.

src/backend/distributed/executor/adaptive_executor.c
SaitTalhaNisanci: The reason I have added this line is because when closing the connection there is this check:
https://github.com/citusdata/citus/blob/master/src/backend/distributed/transaction/remote_transaction.c#L763

Since we skip some parts in TransactionState for repartition queries, the state is not set to REMOTE_TRANS_INVALID, yet we try to delete something that does not exist so this was to bypass that if check. I realize that there could be a better solution, or if we keep this solution I should document that I guess.

src/backend/distributed/executor/adaptive_executor.c
SaitTalhaNisanci: Do you see any problem in doing the cleanup at the top level ? This also speeds up the process a bit because we use only one network round per worker to do cleanup.

src/backend/distributed/executor/adaptive_executor.c
SaitTalhaNisanci: I think in that case we can just add this to LogRemoteCommand:

```
LogRemoteCommand(MultiConnection *connection, const char *command)
{
	if (!LogRemoteCommands)
	{
		return;
	}

	ereport(LOG, (errmsg("issuing %s", ApplyLogRedaction(command)),
				  errdetail("on server %s@%s:%d connectionId: %ld", connection->user, connection->hostname,
							connection->port, connection->connectionId)));
}

```

src/backend/distributed/executor/adaptive_executor.c
marcocitus: I mean that we shouldn't have to use a static/global variable for this purpose.

src/backend/distributed/executor/adaptive_executor.c
marcocitus: As far as I can tell, we only do CloseConnections when we are done executing tasks. I would expect the transaction state to be REMOTE_TRANS_INVALID unless something is wrong. Perhaps this should be an assert?

src/backend/distributed/executor/adaptive_executor.c
marcocitus: Let me rephrase: What happens if we just call CloseConnection as before instead of doing this check?

src/backend/distributed/executor/adaptive_executor.c
marcocitus: yes, that what I was thinking

src/backend/distributed/executor/repartition_logic.c
marcocitus: I don't think that's true. The EXPLAIN code relies on the dependedJobList

src/backend/distributed/executor/repartition_logic.c
marcocitus: CreateSchemasOnAllWorkers

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -698,6 +721,12 @@ AdaptiveExecutor(CitusScanState *scanState)
 
 	FinishDistributedExecution(execution);
 
+	if (hasDependedJobsInAllTree && SubPlanLevel == 0)
+	{
+		CleanUpSchemas();
```
marcocitus: This is only meant to happen at system start-up since it removes *all* job schemas.

src/backend/distributed/executor/repartition_logic.c
SaitTalhaNisanci: Umm, makes sense I will move this to worker_transaction.c

src/backend/distributed/executor/adaptive_executor.c
SaitTalhaNisanci: As far as I see we do not remove a pointer from the `sessionList`. So when we close a connection it gets freed but the pointer is still there pointing to something random. In that case when we iterate the `sessionList` again later we still have the connections in the session list. 

I guess the problem happens when the connection status is `connecting` so in that case after we are done with the execution, adaptive executor tries to close the connections that are either `connecting`, `lost`, or `failed`. I am not sure why we are not removing the session from `sessionList` when its connection is closed. Is there a reason for that? Maybe I am missing where it is being removed. But if it was we wouldnt get the error I described.

src/backend/distributed/executor/repartition_logic.c
SaitTalhaNisanci: I guess you are right, I have debugged some queries and it seems that they have dependentJobList. I have changed it to use a DFS to find job ids starting from the top level job.

src/backend/distributed/executor/adaptive_executor.c
SaitTalhaNisanci: what is the problem with removing all job schemas? Is that because another repartition query could be in progress while this one is removing all job schemas?

marcocitus: yes

src/backend/distributed/worker/worker_merge_protocol.c
```diff
@@ -53,6 +57,50 @@ static void CopyTaskFilesFromDirectory(StringInfo schemaName, StringInfo relatio
...
+	uint64 jobId = PG_GETARG_INT64(0);
+	StringInfo jobDirectoryName = JobDirectoryName(jobId);
+	StringInfo jobSchemaName = JobSchemaName(jobId);
+
+	CheckCitusVersion(ERROR);
+
+	Oid schemaId = get_namespace_oid(jobSchemaName->data, false);
+
+	EnsureSchemaOwner(schemaId);
+	CitusRemoveDirectory(jobDirectoryName);
```
marcocitus: We should probably do RemoveJobSchema here as well

src/include/distributed/multi_server_executor.h
```diff
@@ -36,6 +36,12 @@
 #define JOB_CLEANUP_QUERY "SELECT task_tracker_cleanup_job("UINT64_FORMAT ")"
 #define JOB_CLEANUP_TASK_ID INT_MAX
 
+/* Adaptive executor repartioning related defines */
+#define WORKER_CREATE_SCHEMA_QUERY "SELECT worker_create_schema \
+ ("UINT64_FORMAT ");"
```
marcocitus: the string formatting is a little strange here. Should this be?

```c
"SELECT worker_create_schema" \
"(" UINT64_FORMAT ")";
```

(unless it fits on a single line?)

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -1050,7 +1106,8 @@ DistributedExecutionRequiresRollback(DistributedExecution *execution)
 
 	if (ReadOnlyTask(task->taskType))
 	{
-		return SelectOpensTransactionBlock && IsTransactionBlock();
+		return !execution->isOutsideTransaction && SelectOpensTransactionBlock &&
+			   IsTransactionBlock();
```
marcocitus: Maybe add an `if (execution->isOutsideTransaction) return false;` somewhere above to make it more explicit and robust to other types of tasks.

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -277,8 +281,16 @@ typedef struct DistributedExecution
 	 */
 	AttInMetadata *attributeInputMetadata;
 	char **columnArray;
+
+	/*
+	 * isOutsideTransaction is used to determine if we should begin a transaction
+	 * or not. For example, we do not begin a transaction for repartition specific
+	 * transactions.
+	 */
+	bool isOutsideTransaction;
```
marcocitus: After some discussion with Onder, we agreed that we effectively have 3 situations:
- We need tasks to be executed in transaction blocks (isTransaction is true)
- We need tasks to be executed outside transaction blocks (isOutsideTransaction is true)
- We don't care whether or not tasks are executed in transaction blocks (both false)

It would make sense for these 3 situations to be captured in a single enum, rather than 2 bools.

For instance `RemoteTransactionBlocksUsage useRemoteTransactionBlocks` with `REMOTE_TRANSACTION_BLOCKS_REQUIRED`, `REMOTE_TRANSACTION_BLOCKS_ALLOWED`, `REMOTE_TRANSACTION_BLOCKS_DISALLOWED`.

src/backend/distributed/worker/worker_merge_protocol.c
SaitTalhaNisanci: We use the schemas in the SQL_TASK queries so we should remove them only at the end I think.

For example for `select count(*) from ab k, ab l where k.a = l.b;`
```
LOG:  issuing SELECT count(*) AS count FROM (pg_merge_job_469577498631.task_000033 "pg
_merge_job_469577498631.task_000033" JOIN pg_merge_job_469577498632.task_000033 "pg_me
rge_job_469577498632.task_000033" ON (("pg_merge_job_469577498631.task_000033".interme
diate_column_469577498631_0 OPERATOR(pg_catalog.=) "pg_merge_job_469577498632.task_000
033".intermediate_column_469577498632_0))) WHERE true
```

SaitTalhaNisanci: I guess we should do cleanup after the whole execution not after `executeDependentTasks`. Then this wouldnt be a problem

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -277,8 +283,20 @@ typedef struct DistributedExecution
 	 */
 	AttInMetadata *attributeInputMetadata;
 	char **columnArray;
+
+	/*
+	 * useRemoteTransactionBlocks is used to manage remote transactions.
```
marcocitus: Comment is unclear. What does this variable mean?

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -635,6 +654,12 @@ AdaptiveExecutor(CitusScanState *scanState)
 
 	ExecuteSubPlans(distributedPlan);
 
+	bool hasDependentJobs = HasDependentJobs(job);
+	if (hasDependentJobs)
+	{
```
marcocitus: Just to keep this function clean, could the `if` be moved into `ExecuteDependentTasks`?

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -945,7 +1011,11 @@ StartDistributedExecution(DistributedExecution *execution)
 	 * If the current or previous execution in the current transaction requires
 	 * rollback then we should use transaction blocks.
 	 */
-	execution->isTransaction = InCoordinatedTransaction();
+	if (InCoordinatedTransaction() && execution->useRemoteTransactionBlocks ==
+		REMOTE_TRANSACTION_BLOCKS_ALLOWED)
```
marcocitus: what if it is DISALLOWED?

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -1048,9 +1118,15 @@ DistributedExecutionRequiresRollback(DistributedExecution *execution)
 		return IsTransactionBlock();
 	}
 
+	if (execution->useRemoteTransactionBlocks == REMOTE_TRANSACTION_BLOCKS_DISALLOWED)
+	{
+		return false;
```
marcocitus: Should this be up higher among the other `return false` conditions? It's a bit weird that we could theoretically return true when it is disallowed.

Could use a comment, e.g. `/* tasks cannot run in a transaction block */`

src/backend/distributed/executor/repartition.c
```diff
@@ -0,0 +1,169 @@
...
+
+
+/*
+ * DoRepartitionCleanup removes the temporary job directories and schemas that are
+ * used for repartition queries for the given job ids.
+ */
+void
+DoRepartitionCleanup(List *jobIds)
+{
+	SendCommandListToAllWorkers(list_make1(GenerateDeleteJobsCommand(jobIds)));
```
marcocitus: We need to make sure not to generate any errors here, since it's called from a PG_CATCH. `SendCommandListToAllWorkers` may throw errors iirc.

src/backend/distributed/sql/udfs/worker_repartition_cleanup/9.1-1.sql
```diff
@@ -0,0 +1,8 @@
+CREATE FUNCTION pg_catalog.worker_repartition_cleanup(bigint)
```
marcocitus: maybe worker_job_cleanup

src/backend/distributed/utils/listutils.c
```diff
@@ -155,6 +157,44 @@ ListToHashSet(List *itemList, Size keySize, bool isStringList)
 }
 
 
+/*
+ * ListAppendUniqueUint64 appends the given uint64 if it does not
+ * exist in the list.
+ */
+List *
+ListAppendUniqueUint64(List *list, uint64 datum)
```
marcocitus: Is this code no longer used? It shows up as not covered.

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -635,6 +654,12 @@ AdaptiveExecutor(CitusScanState *scanState)
 
 	ExecuteSubPlans(distributedPlan);
 
+	bool hasDependentJobs = HasDependentJobs(job);
+	if (hasDependentJobs)
+	{
+		jobList = ExecuteDependentTasks(taskList, job);
```
marcocitus: Returning the jobList here seems a little arbitrary. Maybe rebuild it in DoRepartitionCleanup? It's a very cheap operation.

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -846,6 +906,12 @@ CreateDistributedExecution(RowModifyLevel modLevel, List *taskList, bool hasRetu
 	execution->connectionSetChanged = false;
 	execution->waitFlagsChanged = false;
 
+	execution->useRemoteTransactionBlocks = REMOTE_TRANSACTION_BLOCKS_ALLOWED;
+	if (isOutsideTransaction)
```
onderkalaci: suggestion: replace `isOutsideTransaction ` with `excludeFromTransaction`, in the sense that the caller chooses to exclude the execution from the transaction block.

`isOutsideTransaction` feels a bit vague to me.

src/backend/distributed/executor/adaptive_executor.c
SaitTalhaNisanci: We use `hasDependentJobs` boolean to clean schemas as well. So if since we want to call the function(hasDependentJobs) once it makes sense to keep it here I think. Though maybe we can execute dependent jobs after creating a distributed execution, and that maybe the function could be more clean. I will take a look at it

src/backend/distributed/executor/adaptive_executor.c
SaitTalhaNisanci: we have three states:
- by default a transaction can be opened so it is **allowed**.
- If there are dependent jobs then a transaction should not be opened, so it is **disallowed**.
- If we are in a coordinated transaction and there are no dependent jobs, then we should open a transaction, so it is **required**. 

If we are in a coordinated transaction but we have dependent jobs then we do not open transaction so that is why the check is here.

I will add it as a comment here.

src/backend/distributed/sql/udfs/worker_repartition_cleanup/9.1-1.sql
SaitTalhaNisanci: I thought that as what we are doing is specific to repartition, it is better to be explicitly state it here. Like we cleanup repartition specific things. What do you think?

src/backend/distributed/utils/listutils.c
SaitTalhaNisanci: Yeah after changing the way we found jobids, this is no longer necessary.

src/backend/distributed/executor/adaptive_executor.c
SaitTalhaNisanci: In order to rebuild it, we need the top level job, as far as I see we do not have it in execution. So I will put the top level job to the executor then to be able to reach to it from PG_CATCH where we only have execution.

src/backend/distributed/executor/adaptive_executor.c
SaitTalhaNisanci: I think that fits more into what we are doing, thanks for the suggestion

src/backend/distributed/executor/repartition.c
SaitTalhaNisanci: you mean we should have a pg_catch inside SendCommandListToAllWorkers  as well?

src/backend/distributed/executor/adaptive_executor.c
marcocitus: True, did not consider that. In that case the `jobIdList` is probably a simpler thing to keep than the top-level job.

src/backend/distributed/executor/adaptive_executor.c
marcocitus: I meant:

What if we are in a coordinated transaction and useRemoteTransactionBlocks is DISALLOWED? We should throw an error or have an assert.

src/backend/distributed/executor/adaptive_executor.c
marcocitus: makes sense, ok keeping it this way

src/backend/distributed/sql/udfs/worker_repartition_cleanup/9.1-1.sql
marcocitus: Actually, could we reuse task_tracker_cleanup_job here? What's the difference between those functions?

src/backend/distributed/executor/adaptive_executor.c
SaitTalhaNisanci: when we have a query like:

```sql
begin;
a repartition query;
end;
```

do we want to throw an error? I dont think we do.

src/backend/distributed/sql/udfs/worker_repartition_cleanup/9.1-1.sql
SaitTalhaNisanci: `task_tracker_cleanup_job ` has task tracker specific hash table and it removes tasks from the hash table which was filled by task tracker but later it removes the directory and schema as well. So they have common parts so I think we can use the `worker_repartition_cleanup` to do remove directory and schema in task tracker as well, but using `task_tracker_cleanup_job ` seems wrong to me because it is task tracker specific.

src/backend/distributed/executor/adaptive_executor.c
marcocitus: but `InCoordinatedTransaction()` wouldn't be true in that case, since `DistributedExecutionRequiresRollback` returns false for the repartition query. 

This code is pretty confusing to me. The outcome of DistributedExecutionRequiresRollback depends on the value of useRemoteTransactionBlocks, but now the value of useRemoteTransactionBlocks depends on the outcome of DistributedExecutionRequiresRollback.

src/backend/distributed/executor/repartition.c
marcocitus: we should have an "optional" variant, e.g. like SendOptionalCommandListToWorkerInTransaction

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -277,8 +283,24 @@ typedef struct DistributedExecution
...
+	 * - opening a remote transaction is disallowed
+	 * - opening a remote transaction does not matter, so it is allowed but not required.
+	 */
+	enum RemoteTransactionBlocksUsage useRemoteTransactionBlocks;
+
+	/*
+	 * jobList contains all jobs in the job tree, this is used to
+	 * do cleanup for repartition queries.
+	 */
+	List *jobList;
```
marcocitus: maybe jobIdList

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -707,6 +744,13 @@ AdaptiveExecutor(CitusScanState *scanState)
 }
 
 
+static bool
+HasDependentJobs(Job *mainJob)
```
marcocitus: missing function comment

src/backend/distributed/executor/adaptive_executor.c
SaitTalhaNisanci: In `InCoordinatedTransaction()` we check if the transaction state is NONE or IDLE.

When we have a subquery we start a transaction so the transaction state will be STARTED, after that if we have a repartition query in the subquery then the DistributedExecutionRequiresRollback will return false because this is a repartition query however our transaction state is STARTED, so in that case `InCoordinatedTransaction()` would return true. 

I have moved the decision of whether to open a transaction or not to the creation of execution.

src/backend/distributed/executor/repartition.c
SaitTalhaNisanci: is this possibly what you meant https://github.com/citusdata/citus/pull/3169/commits/3dfc24f7a80bbcc8c7a7672f603016fe18c44b7b

src/backend/distributed/transaction/worker_transaction.c
```diff
@@ -142,13 +155,37 @@ SendCommandListToAllWorkers(List *commandList)
 	foreach(workerNodeCell, workerNodeList)
 	{
 		WorkerNode *workerNode = (WorkerNode *) lfirst(workerNodeCell);
-		SendCommandListToWorkerInSingleTransaction(workerNode->workerName,
-												   workerNode->workerPort, extensionOwner,
-												   commandList);
+		PG_TRY();
+		{
+			SendCommandListToWorkerInSingleTransaction(workerNode->workerName,
```
marcocitus: note that there's already a `SendOptionalCommandListToWorkerInTransaction`

SaitTalhaNisanci: oh my bad, didnt know that. I will use it.

SaitTalhaNisanci: I thought you wanted me to create such a method :)

src/backend/distributed/executor/multi_server_executor.c
```diff
@@ -104,12 +106,11 @@ JobExecutorType(DistributedPlan *distributedPlan)
...
 										"to enable repartitioning")));
 			}
-
-			ereport(DEBUG1, (errmsg(
-								 "cannot use adaptive executor with repartition jobs"),
-							 errhint("Since you enabled citus.enable_repartition_joins "
-									 "Citus chose to use task-tracker.")));
-			return MULTI_EXECUTOR_TASK_TRACKER;
+			if (ShardReplicationFactor > 1)
+			{
```
marcocitus: This comparison uses the current value of the GUC, not the actual replication factor. What happens if the GUC changes to 1 after creating a table with replication factor 2?

src/backend/distributed/executor/repartition.c
```diff
@@ -0,0 +1,178 @@
+/*-------------------------------------------------------------------------
+ *
+ * repartition.c
```
marcocitus: Maybe `repartition_join_execution.c` (kind of like `subplan_execution.c`)

src/backend/distributed/executor/multi_server_executor.c
SaitTalhaNisanci: Then it would work incorrectly. We should use the replication factor at the creation time here.

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -704,6 +724,11 @@ AdaptiveExecutor(CitusScanState *scanState)
 
 	FinishDistributedExecution(execution);
 
+	if (hasDependentJobs)
+	{
+		DoRepartitionCleanup(jobIdList, true);
```
marcocitus: Why is it ok to error here? We still want to make sure we clean up as much as possible.

SaitTalhaNisanci: Yeah, that is true.

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -646,7 +665,8 @@ AdaptiveExecutor(CitusScanState *scanState)
 		tuplestore_begin_heap(randomAccess, interTransactions, work_mem);
 
 	TransactionProperties xactProperties =
-		DecideTransactionPropertiesForTaskList(distributedPlan->modLevel, taskList);
+		DecideTransactionPropertiesForTaskList(distributedPlan->modLevel, taskList,
```
marcocitus: Since it does not depend on current state, maybe just initialize xactProperties directly?

src/backend/distributed/executor/directed_acylic_graph_execution.c
```diff
@@ -0,0 +1,218 @@
...
+	return hash;
+}
+
+
+static int
+TaskHashCompare(const void *key1, const void *key2, Size keysize)
+{
+	TaskHashKey *taskKey1 = (TaskHashKey *) key1;
+	TaskHashKey *taskKey2 = (TaskHashKey *) key2;
+	return taskKey1->jobId != taskKey2->jobId || taskKey1->taskId != taskKey2->taskId;
```
marcocitus: This does not seem like a correct compare function. It's supposed to return -1, 0 or 1.

src/backend/distributed/executor/multi_server_executor.c
```diff
@@ -131,6 +134,35 @@ JobExecutorType(DistributedPlan *distributedPlan)
...
 
 
+/*
+ * HasReplicatedNonReferenceTable returns true if there is any
+ * table in the given list that is:
+ * - not a reference table
+ * - has replication factor > 1
+ */
+static bool
+HasReplicatedNonReferenceTable(List *relationOids)
```
marcocitus: Maybe HasReplicatedDistributedTable

src/backend/distributed/transaction/worker_transaction.c
```diff
@@ -117,6 +119,73 @@ SendCommandToWorkersWithMetadata(const char *command)
...
+/*
+ * SendCommandListToAllWorkersInternal sends the given command to all workers in a single
+ * transaction. If failOnError is false, then it continues sending the commandList to other
+ * workers even if it fails in one of them.
+ */
+static void
+SendCommandListToAllWorkersInternal(List *commandList, bool failOnError)
+{
+	ListCell *workerNodeCell = NULL;
+	char *extensionOwner = CitusExtensionOwnerName();
```
marcocitus: We've had some bugs from calling these functions with the expectation that they sent a command as the current user. It would be nice to either pass the user as a parameter, or add something like AsSuperuser to the name.

src/backend/distributed/transaction/worker_transaction.c
```diff
@@ -117,6 +119,73 @@ SendCommandToWorkersWithMetadata(const char *command)
...
+/*
+ * SendCommandListToAllWorkersInternal sends the given command to all workers in a single
+ * transaction. If failOnError is false, then it continues sending the commandList to other
+ * workers even if it fails in one of them.
+ */
+static void
+SendCommandListToAllWorkersInternal(List *commandList, bool failOnError)
+{
+	ListCell *workerNodeCell = NULL;
+	char *extensionOwnerAsSuperuser = CitusExtensionOwnerName();
```
marcocitus: Would be good to either add AsSuperuser to the name of these functions, or pass the user as a parameter (preferable).

Issue #2582, opened 2019-01-09
Support for CHbenCHmark queries


marcocitus:
[CHbenCHmark](https://db.in.tum.de/research/projects/CHbenCHmark/?lang=en) is a pretty interesting benchmark for Citus because it mixes an OLTP workload (TPC-C) with analytical queries that are based on TPC-H.

It is one of the benchmarks supported by [oltbench](https://github.com/oltpbenchmark/oltpbench).

Out of the 22 analytical queries, 5 currently give an error: Q7, Q8, Q9, Q11, Q20.

Using the following schema:
```sql
DROP TABLE IF EXISTS order_line;
CREATE TABLE order_line (
  ol_w_id int NOT NULL,
  ol_d_id int NOT NULL,
  ol_o_id int NOT NULL,
  ol_number int NOT NULL,
  ol_i_id int NOT NULL,
  ol_delivery_d timestamp NULL DEFAULT NULL,
  ol_amount decimal(6,2) NOT NULL,
  ol_supply_w_id int NOT NULL,
  ol_quantity decimal(2,0) NOT NULL,
  ol_dist_info char(24) NOT NULL,
  PRIMARY KEY (ol_w_id,ol_d_id,ol_o_id,ol_number)
);

DROP TABLE IF EXISTS new_order;
CREATE TABLE new_order (
  no_w_id int NOT NULL,
  no_d_id int NOT NULL,
  no_o_id int NOT NULL,
  PRIMARY KEY (no_w_id,no_d_id,no_o_id)
);

DROP TABLE IF EXISTS stock;
CREATE TABLE stock (
  s_w_id int NOT NULL,
  s_i_id int NOT NULL,
  s_quantity decimal(4,0) NOT NULL,
  s_ytd decimal(8,2) NOT NULL,
  s_order_cnt int NOT NULL,
  s_remote_cnt int NOT NULL,
  s_data varchar(50) NOT NULL,
  s_dist_01 char(24) NOT NULL,
  s_dist_02 char(24) NOT NULL,
  s_dist_03 char(24) NOT NULL,
  s_dist_04 char(24) NOT NULL,
  s_dist_05 char(24) NOT NULL,
  s_dist_06 char(24) NOT NULL,
  s_dist_07 char(24) NOT NULL,
  s_dist_08 char(24) NOT NULL,
  s_dist_09 char(24) NOT NULL,
  s_dist_10 char(24) NOT NULL,
  PRIMARY KEY (s_w_id,s_i_id)
);

[...]

PR #3171, opened 2019-11-08
Make simple in queries on unique columns work with repartion join

JelteF:
DESCRIPTION: Support simple in queries on unique columns with repartition joins

This is necassery to support Q20 of the CHbenCHmark: #2582.

To summarize the fix: The subquery is converted into an INNER JOIN on a
table. This fixes the issue, since an INNER JOIN on a table is already
supported by the repartion planner.

The way this replacement is happening.:
1. Postgres replaces `col in (subquery)` with a SEMI JOIN (subquery) on col = subquery_result
2. If this subquery is simple enough Postgres will replace it with a
   regular read from a table
3. If the subquery returns unique results (e.g. a primary key) Postgres
   will convert the SEMI JOIN into an INNER JOIN during the planning. It
   will not change this in the rewritten query though.
4. We check if Postgres sends us any SEMI JOINs during its join order
   planning, if it doesn't we replace all SEMI JOINs in the rewritten
   query with INNER JOIN (which we already support).


src/backend/distributed/planner/query_pushdown_planning.c
```diff
@@ -160,7 +178,7 @@ ShouldUseSubqueryPushDown(Query *originalQuery, Query *rewrittenQuery)
 	if (FindNodeCheck((Node *) rewrittenQuery->jointree, IsOuterJoinExpr))
 	{
 		/* Assert what _should_ be only situation this occurs in. */
-		Assert(JoinTreeContainsSubquery(originalQuery));
+		/* Assert(JoinTreeContainsSubquery(originalQuery)); */
```
marcocitus: should be removed?

src/test/regress/expected/ch_bench_subquery_repartition.out
```diff
@@ -0,0 +1,217 @@
...
+          s_name           |              s_address              
+---------------------------+-------------------------------------
+ Supplier#000000033        | gfeKpYw3400L0SDywXA6Ya1Qmq1w6YB9f3R
+ Supplier#000000044        | kERxlLDnlIZJdN66zAPHklyL
+(2 rows)
+
+SET client_min_messages TO WARNING;
+DROP SCHEMA ch_bench_subquery_repartition CASCADE;
+COMMIT;
+WARNING:  there is no transaction in progress
```
marcocitus: COMMIT; can be removed

src/backend/distributed/planner/query_pushdown_planning.c
JelteF: Yes, did that now.

PR #3174, opened 2019-11-10
Remove failure tests related to real-time executor

onderkalaci:
Since we've removed the executor, we don't need the specific tests.
The test is already using adaptive executor, so it was passing.
But, we've plenty of extra tests for adaptive executor, so seems safe
to remove.


src/test/regress/expected/failure_single_mod.out
```diff
@@ -44,7 +44,7 @@ WHERE shardid IN (
 ) AND shardstate = 3 RETURNING placementid;
  placementid 
 -------------
-         137
+         125
```
marcocitus: can this be made deterministic? / is the placementid relevant?

onderkalaci: > / is the placementid relevant?

I think not at all, it is used to show that there is one placement with `shardstate = 3 `.

I'm currently having some issues on my local machine with pipenv/python versions, so couldn't run the failure tests locally and fix this. I'll do it today/tomorrow, merging the PR for now

PR #3178, opened 2019-11-11
Return early in CitusHasBeenLoaded when creating a different extension

marcocitus:
DESCRIPTION: Fixes a bug that caused other CREATE EXTENSION commands to take longer

When `creating_extension` is true, we always go into the logic in `CitusHasBeenLoaded` regardless of whether Citus has been loaded. This is primarily to reset the `citusVersionKnownCompatible` flag after an `ALTER EXTENSION`. However, this resulted in `CheckAvailableVersion` being called for every DDL command during a `CREATE EXTENSION notcitus`, which made it significantly slower.

This PR refactors `CitusHasBeenLoaded` and makes the following changes to its behaviour: 
- We do not reset `citusVersionKnownCompatible`. In practice the Citus versions are generally not compatible before an `ALTER EXTENSION`, especially since we changed to one upgrade script per major release.
- We only start the maintenance daemon after creating the citus extension, since starting it earlier might lead to issues if the maintenance daemon tries to read from not-yet-existent tables
- We no longer have special behaviour or additional overhead during `CREATE EXTENSION notcitus`.

Before this change, CREATE EXTENSION postgis took 50-500 seconds on my laptop. After this change, CREATE EXTENSION postgis takes 1.2 seconds.

### DIFF
--- a/src/backend/distributed/utils/metadata_cache.c
+++ b/src/backend/distributed/utils/metadata_cache.c
@@ -202,6 +202,7 @@ static char * InstalledExtensionVersion(void);
 static bool HasOverlappingShardInterval(ShardInterval **shardIntervalArray,
 										int shardIntervalArrayLength,
 										FmgrInfo *shardIntervalSortCompareFunction);
+static bool CitusHasBeenLoadedInternal(void);
 static void InitializeCaches(void);
 static void InitializeDistCache(void);
 static void InitializeDistObjectCache(void);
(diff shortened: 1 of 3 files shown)


src/test/regress/expected/isolation_dump_global_wait_edges.out
```diff
@@ -29,11 +29,11 @@ step detector-dump-wait-edges:
 
 waiting_transaction_numblocking_transaction_numblocking_transaction_waiting
 
-290            289            f              
+289            288            f
```
marcocitus: Caused by the change in master_drop_sequences, previously a DROP SEQUENCE during CREATE EXTENSION set off a distributed transaction.

serprex:
Was wondering why PostGIS was taking so long in my aggregate tests. Confirmed that this resolved that

src/backend/distributed/utils/metadata_cache.c
```diff
@@ -1576,26 +1563,58 @@ CitusHasBeenLoaded(void)
...
+{
+	Oid citusExtensionOid = InvalidOid;
+
+	if (IsBinaryUpgrade)
+	{
+		/* never use Citus logic during pg_upgrade */
+		return false;
+	}
+
+	citusExtensionOid = get_extension_oid("citus", true);
```
pykello: Caching this seems to drop another 10% overhead. Without your PR, it takes 12s for me, with your PR 950ms, with caching this 860ms.

marcocitus:
Anything needed for a :ship:?

PR #3180, opened 2019-11-11
Feature: Expression in reference join

thanodnl:
DESCRIPTION: Expression in reference join

Fixed: #2582

This patch allows arbitrary expressions in the join clause when joining to a reference table. An example of such joins could be found in CHbenCHmark queries 7, 8, 9 and 11; `mod((s_w_id * s_i_id),10000) = su_suppkey` and `ascii(substr(c_state,1,1)) = n2.n_nationkey`. Since the join is on a reference table these queries are able to be pushed down to the workers.

To implement these queries we will widen the `IsJoinClause` predicate to not check if the expressions are a type `Var` after stripping the implicit coerciens. Instead we define a join clause when the `Var`'s in a clause come from more than 1 table.

This allows more clauses to pass into the logical planner's `MultiNodeTree(...)` planning function. To compensate for this we tighten down the `LocalJoin`, `SinglePartitionJoin` and `DualPartitionJoin` to check for direct column references when planning. This allows the planner to work with arbitrary join expressions on reference tables.

This causes some previously unplannable queries to give a slightly less straightforward error message.

Take a query
```SQL
SELECT *
FROM dist1, dist2
WHERE dist1.distcolumn + 1 = dist2.distcolumn;
```

Instead of 
```
```
It now gives the following error
```
DETAIL:  Cartesian products are currently unsupported
```

Reason is that before the join was not recognised as a join clause by `IsJoinClause` since the left hand side was not a column reference. With our new logic this clause is recognised as a join clause because the columns in the expression are from multiple different tables. Since there is no join operation that can execute this the planner falls back to a cartesian product that we do not support. Such queries would be solvable if we update the Singe and Dual Partition Join operations to work on expressions instead of columns.


src/backend/distributed/planner/multi_join_order.c
```diff
@@ -1121,6 +1132,46 @@ MakeJoinOrderNode(TableEntry *tableEntry, JoinRuleType joinRuleType,
 }
 
 
+/*
+ * IsApplicableJoinClause tests if the current joinClause is applicable to the join at
+ * hand.
+ *
+ *   Given a list of left hand tables and a candidate right hand table the join clause is
+ *   valid if atleast 1 column is from the right hand table AND all columns can be found
+ *   in either the list of tables on the left *or* in the right hand table.
```
JelteF: Seems this has some unintended extra indentation?

src/backend/distributed/planner/multi_join_order.c
```diff
@@ -1157,12 +1198,20 @@ ApplicableJoinClauses(List *leftTableIdList, uint32 rightTableId, List *joinClau
 }
 
 
+/* Returns the left expression in the given join clause. */
+Node *
+LeftHandExpr(OpExpr *joinClause)
```
JelteF: This function can be static (and moved after it's first usage). (same for `RightHandExpr`)

src/backend/distributed/planner/multi_join_order.c
```diff
@@ -1157,12 +1198,20 @@ ApplicableJoinClauses(List *leftTableIdList, uint32 rightTableId, List *joinClau
...
+{
+	List *argumentList = joinClause->args;
+	return (Node *) linitial(argumentList);
+}
+
+
 /* Returns the left column in the given join clause. */
 Var *
 LeftColumn(OpExpr *joinClause)
```
JelteF: There's only two usages of this function left (in `multi_logical_planner.c`). Can these be changed to the new `orNULL` variants as well?

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -1387,62 +1388,52 @@ ExtractFromExpressionWalker(Node *node, QualifierWalkerContext *walkerContext)
...
-	argumentList = operatorExpression->args;
+	operatorExpression = castNode(OpExpr, clause);
+	equalsOperator = OperatorImplementsEquality(operatorExpression->opno);
 
-	/* join clauses must have two arguments */
-	if (list_length(argumentList) != 2)
+	if (!equalsOperator)
 	{
+		/* Not an equi-join */
 		return false;
```
marcocitus: Maybe add a comment on why we still need this restriction.

It would be nice if we could lift the it at some point, because it does not seem relevant whether or not the expression contains an "=" symbol. For instance `distance(dist.point, ref.point) < 20` is not fundamentally different from `mod((dist.x * dist.y),10000) = ref.z` in terms of the conditions in which it can be planned and executed.

src/test/regress/sql/ch_bench_join_expr.sql
```diff
@@ -0,0 +1,70 @@
...
+SELECT create_distributed_table('order_line','ol_w_id');
+SELECT create_distributed_table('stock','s_w_id');
+SELECT create_reference_table('supplier');
+
+INSERT INTO supplier SELECT c, 'abc', 'def', c, 'ghi', c, 'jkl' FROM generate_series(0,10) AS c;
+INSERT INTO stock SELECT c,c,c,c,c,c, 'abc','abc','abc','abc','abc','abc','abc','abc','abc','abc','abc' FROM generate_series(1,3) AS c;
+INSERT INTO stock SELECT c, 5000,c,c,c,c, 'abc','abc','abc','abc','abc','abc','abc','abc','abc','abc','abc' FROM generate_series(1,3) AS c; -- mod(2*5000,10000) == 0
+INSERT INTO order_line SELECT c, c, c, c, c, NULL, c, c, c, 'abc' FROM generate_series(0,10) AS c;
+
+-- MVP for CH-BenCHmark queries 7/8/9 where a join with an expression is done between a distributed table and a reference table.
```
marcocitus: Could maybe add queries from https://gist.github.com/marcocitus/02a8068a65850549995e8e385bce0018

src/test/regress/expected/multi_subquery.out
```diff
@@ -377,7 +378,8 @@ JOIN
 ON (l_orderkey = o_orderkey + 1)
 ORDER BY l_orderkey DESC
 LIMIT 10;
-ERROR:  complex joins are only supported when all distributed tables are joined on their distribution columns with equal operator
+ERROR:  cannot perform distributed planning on this query
```
marcocitus: I wonder if we could improve this error message, at least for this particular case. We've previously gotten a lot of feedback from users to provide better error messages in unsupported cases.

src/backend/distributed/planner/multi_logical_planner.c
marcocitus: Note that removing doesn't seem to lead to any issues, a few queries in regression tests start working.

thanodnl: I like the idea of supporting more, however we need to add this check back in Local and Partition joins as they still would rely on equality of both sides.

Biggest problem is testing the support of such joins, as it opens up a whole class of new queries that might work, or not :).

src/test/regress/sql/ch_bench_join_expr.sql
thanodnl: Makes sense, and touches on our conversation yesterday :)

Would it be enough to add these queries without data? At least we know they are planned and executed, but we won't have assurance they provide the correct result.

Ideally we would have data that makes sense for these queries, which so far has been non-trivial in even the reference setup. But I can start working on that.

src/test/regress/expected/multi_subquery.out
thanodnl: Yeah, I am not happy about these errors and have looked around how to make these errors better.

Biggest problem is that we changed what we do recognised as a join. Before the `(l_orderkey = o_orderkey + 1)` clause was not recognised as a join, but now that we do not require `Var`'s to be directly on either side the clause is recognised as a join.

This causes `MultiNodeTree(...)` to be called which _always_ returns a plan, as everything will fall back to a cartesian product for joins that do not satisfy any other join order planner. The 'correct' approach would be to error in the `DualPartitionJoin` as this should be handled in the end by that planner. However it is not expected in the join order planner to error.

I wonder if removing the cartesian product planner and providing the `complex joins are only supported ....` error if the join order planner can't find anything without it. But that will change the output of a dozen or so test suites as the Cartesian product error is throughout our suite :(.

Any ideas how to provide a correct error here?

src/backend/distributed/planner/multi_join_order.c
thanodnl: it was actually intended, albeit not used in our code base often it is often used in postgres.

I will remove the extra indent.

src/backend/distributed/planner/multi_join_order.c
thanodnl: Now it could be static yes, but I hope we can soon fix the partition joins to allow for repartitions on expressions instead of `Var`'s which would need this function again.

That said, `LeftColumn` and `RightColumn` could probably be removed as you stated below. That would make these lines prime for inlining in the `...OrNULL` variants of both, causing these functions to be removed.

src/test/regress/expected/multi_subquery.out
thanodnl: I have cooked up something that works: https://github.com/citusdata/citus/pull/3180/commits/1e084ea85b0c2f61500a536defcdf441f1f3e9d7

Instead of always falling back to a cartesian product it will only apple the cartesian product if there are no applicable join clauses between tables. This will prevent queries like above to be planned with a cartesian product, but instead fail to plan.

For this to work the join order planner is now allowed to come up with a plan for a certain permutation. This cascades down the call chain and will give the complex query error if there is no candidate plan found to pick a best plan from. This seems to be a good tradeoff as this error reverted back while other cartesian errors remained.

src/backend/distributed/planner/multi_logical_planner.c
thanodnl: removing the clause is tracked in https://github.com/citusdata/citus/issues/3198

marcocitus: For now, a possible workaround is `(<expression>)::int = 1`

src/test/regress/sql/chbenchmark_all_queries.sql
```diff
@@ -0,0 +1,722 @@
+SET citus.next_shard_id TO 1650000;
+CREATE SCHEMA chbenchmark_all_queries;
+SET search_path TO chbenchmark_all_queries;
+
+SET citus.enable_repartition_joins TO on;
```
marcocitus: Wonder whether we should turn this on by default, or wait for the adaptive executor implementation.

src/backend/distributed/planner/multi_join_order.c
```diff
@@ -905,6 +941,8 @@ SinglePartitionJoin(JoinOrderNode *currentJoinNode, TableEntry *candidateTable,
 	/*
 	 * If we previously dual-hash re-partitioned the tables for a join or made
 	 * cartesian product, we currently don't allow a single-repartition join.
+	 *
+	 * TODO: verify what happens when there was an other join in between our current evaluation and below joins, this pattern has caused crashes before.
```
marcocitus: can this todo be resolved?

thanodnl: I now have the knowledge of what kind of query to create to trigger the case that would verify its behaviour. I will add it to the test suite and remove the TODO

thanodnl: Removed the TODO.

There is a bit of weird behaviour but not crashing, there are queries in our test suite covering this.

The side effect of `DUAL_PARTITION_JOIN` is that it loses the `currentPartitionColumn` since there is no common column after a dual repartition. `SinlgePartitionJoinClause` will always return `NULL` when there is not `currentPartitionColumn`. It could still enter a single partition clause  afterwards if it could partition to the candidate column.

`CARTESIAN_PRODUCT` strangely retains its current partition column which sounds like something it should not, but given the cartesian product will never execute it doesn't matter much. I wonder if we at some point shoulld remove the cartesian join order planner from the planner as it will never end up with an executable path, and the error message is quite vague.

src/test/regress/sql/chbenchmark_all_queries.sql
thanodnl: Don't know the reason for it being turned off, I always though this was to warn users about the slow performance once repartitions come in play.

Now the user needs to enable repartitions which might make them think about their sharding strategy more.

PR #3181, opened 2019-11-12
Automatically convert useless declarations using regex replace

JelteF:
One pattern we have in our codebase is a lot of useless 0/NULL declarations at the top of the function scope, so we have this code:
```c
int a = 0;
int b = 2;
Assert(b == 2);
a = b + b;
```
instead of the shorter:
```c
int b = 2;
Assert(b == 2);
int a = b + b;
```
After failing to find a tool that merges declarations. I decided that a regex
should be able to do it as well. The idea is that should find a declaration and if the next occurunce of the variable is an assignment, the declaration will be moved there.

After some trial and error I came up with this bash script:
```bash
#!/bin/bash
shopt -s globstar
while true; do
    perl -i -p0e 's/\n\t(?!return )(?P<type>(\w+ )+\**)(?>(?P<variable>\w+)( = *[\w>\s\n-]*?)?;\n(?P<code_between>(?>(?P<comment_or_string_or_not_preprocessor>\/\*.*?\*\/|"(?>\\"|.)*?"|[^#]))*?)(\t)?(?=\b(?P=variable)\b))(?<=\n\t)(?P=variable) =(?![^;]*?[^>_]\b(?P=variable)\b[^_])/\n$+{code_between}\t$+{type}$+{variable} =/sg' ./**/*.c;
    # The following are simply the same regex, but repeated for different tab sizes
    # (this is needed because variable sized backtracking is not supported in perl)
    perl -i -p0e 's/\n\t\t(?!return )(?P<type>(\w+ )+\**)(?>(?P<variable>\w+)( = *[\w>\s\n-]*?)?;\n(?P<code_between>(?>(?P<comment_or_string_or_not_preprocessor>\/\*.*?\*\/|"(?>\\"|.)*?"|[^#]))*?)(\t\t)?(?=\b(?P=variable)\b))(?<=\n\t\t)(?P=variable) =(?![^;]*?[^>_]\b(?P=variable)\b[^_])/\n$+{code_between}\t\t$+{type}$+{variable} =/sg' ./**/*.c;
    perl -i -p0e 's/\n\t\t\t(?!return )(?P<type>(\w+ )+\**)(?>(?P<variable>\w+)( = *[\w>\s\n-]*?)?;\n(?P<code_between>(?>(?P<comment_or_string_or_not_preprocessor>\/\*.*?\*\/|"(?>\\"|.)*?"|[^#]))*?)(\t\t\t)?(?=\b(?P=variable)\b))(?<=\n\t\t\t)(?P=variable) =(?![^;]*?[^>_]\b(?P=variable)\b[^_])/\n$+{code_between}\t\t\t$+{type}$+{variable} =/sg' ./**/*.c;
    perl -i -p0e 's/\n\t\t\t\t(?!return )(?P<type>(\w+ )+\**)(?>(?P<variable>\w+)( = *[\w>\s\n-]*?)?;\n(?P<code_between>(?>(?P<comment_or_string_or_not_preprocessor>\/\*.*?\*\/|"(?>\\"|.)*?"|[^#]))*?)(\t\t\t\t)?(?=\b(?P=variable)\b))(?<=\n\t\t\t\t)(?P=variable) =(?![^;]*?[^>_]\b(?P=variable)\b[^_])/\n$+{code_between}\t\t\t\t$+{type}$+{variable} =/sg' ./**/*.c;
    perl -i -p0e 's/\n\t\t\t\t\t(?!return )(?P<type>(\w+ )+\**)(?>(?P<variable>\w+)( = *[\w>\s\n-]*?)?;\n(?P<code_between>(?>(?P<comment_or_string_or_not_preprocessor>\/\*.*?\*\/|"(?>\\"|.)*?"|[^#]))*?)(\t\t\t\t\t)?(?=\b(?P=variable)\b))(?<=\n\t\t\t\t\t)(?P=variable) =(?![^;]*?[^>_]\b(?P=variable)\b[^_])/\n$+{code_between}\t\t\t\t\t$+{type}$+{variable} =/sg' ./**/*.c;
    perl -i -p0e 's/\n\t\t\t\t\t\t(?!return )(?P<type>(\w+ )+\**)(?>(?P<variable>\w+)( = *[\w>\s\n-]*?)?;\n(?P<code_between>(?>(?P<comment_or_string_or_not_preprocessor>\/\*.*?\*\/|"(?>\\"|.)*?"|[^#]))*?)(\t\t\t\t\t\t)?(?=\b(?P=variable)\b))(?<=\n\t\t\t\t\t\t)(?P=variable) =(?![^;]*?[^>_]\b(?P=variable)\b[^_])/\n$+{code_between}\t\t\t\t\t\t$+{type}$+{variable} =/sg' ./**/*.c;
    git diff --quiet && break;
    git add .;
done
```
For a better view of the first regex try this:
https://www.debuggex.com/r/_6Ajy7jetf4QdUaE

The way it works is roughly like this:
1. find a declaration + initialization
2. if the initializer is not a single variable/constant, don't consider replacing it
2. if it is a simple initializer, take the type and variable from that declaration
3. find the next use of this variable
4. if the next use is not an assignment, don't do anything (the value of the original initialization is used)
5. if the next use is an assignment, remove the old declaration + initialization and prefix it with the type:
    - unless the variable is also used in the same line of the new initialization, e.g:
      ```c
      int a = 1;
      a = a + a;
      ```

Source: https://xkcd.com/1171/

This could be as a check and automatic fixer to `citus_indent`. It does depend on the fact that it's run through citus_indent before though. And it also depends on the fact that variables are actually used (which the compiler guarantees for us with `-Werror` in CI), right now this:
```c
int
main(int argc, char *argv[])
{
	int a = 1;
	if (1)
	{
		int a = 2;
	}
	else
	{
		a = 3;
	}

	return 0;
}
```
becomes this:
```c
int
main(int argc, char *argv[])
{
	int a = 1;
	if (1)
	{
	}
	else
	{
		int a = 3;
	}

	return 0;
}
```
This should be pretty easy to fix though if we care about that, because scope is pretty well defined.

Note: This regex currently has to known false negatives for replacement (i.e. replacement is not done when it would be possible):
1. ~~comments containing the the variable name~~ (fixed now)
```c
int foo = 0;
/* setting foo */
foo = 4;
```
2. code in `#if`/`#ifdef`
```c
int foo = 0;
#ifdef SOMETHING
foo = 1
#else
foo = 2
#endif
```
This failing on `#if`/`#ifdef` one is done by stop trying to replace whenever a # character is found. Outside of strings and comments these are really only for preprocessor statements, so that should be pretty fine.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,6 +22,12 @@ jobs:
       - run:
           name: 'Check Style'
           command: citus_indent --check
+      - run:
+          name: 'Remove useless declarations'
+          command: ci/remove_useless_declarations.sh
+      - run:
+          name: 'Check if changed'
+          command: git diff --cached --exit-code
   check-sql-snapshots:
     docker:
       - image: 'citus/extbuilder:latest'
--- a/ci/remove_useless_declarations.sh
+++ b/ci/remove_useless_declarations.sh
@@ -1,0 +1,23 @@
+#!/bin/sh
+
+set -eu
+
+files=$(find src -iname '*.c' | git check-attr --stdin citus-style | grep -v ': unset$' | sed 's/: citus-style: set$//')
+while true; do
+    # shellcheck disable=SC2086
+    perl -i -p0e 's/\n\t(?!return )(?P<type>(\w+ )+\**)(?>(?P<variable>\w+)( = *[\w>\s\n-]*?)?;\n(?P<code_between>(?>(?P<comment_or_string_or_not_preprocessor>\/\*.*?\*\/|"(?>\\"|.)*?"|[^#]))*?)(\t)?(?=\b(?P=variable)\b))(?<=\n\t)(?P=variable) =(?![^;]*?[^>_]\b(?P=variable)\b[^_])/\n$+{code_between}\t$+{type}$+{variable} =/sg' $files
+    # The following are simply the same regex, but repeated for different tab sizes
+    # (this is needed because variable sized backtracking is not supported in perl)
+    # shellcheck disable=SC2086
+    perl -i -p0e 's/\n\t\t(?!return )(?P<type>(\w+ )+\**)(?>(?P<variable>\w+)( = *[\w>\s\n-]*?)?;\n(?P<code_between>(?>(?P<comment_or_string_or_not_preprocessor>\/\*.*?\*\/|"(?>\\"|.)*?"|[^#]))*?)(\t\t)?(?=\b(?P=variable)\b))(?<=\n\t\t)(?P=variable) =(?![^;]*?[^>_]\b(?P=variable)\b[^_])/\n$+{code_between}\t\t$+{type}$+{variable} =/sg' $files
+    # shellcheck disable=SC2086
+    perl -i -p0e 's/\n\t\t\t(?!return )(?P<type>(\w+ )+\**)(?>(?P<variable>\w+)( = *[\w>\s\n-]*?)?;\n(?P<code_between>(?>(?P<comment_or_string_or_not_preprocessor>\/\*.*?\*\/|"(?>\\"|.)*?"|[^#]))*?)(\t\t\t)?(?=\b(?P=variable)\b))(?<=\n\t\t\t)(?P=variable) =(?![^;]*?[^>_]\b(?P=variable)\b[^_])/\n$+{code_between}\t\t\t$+{type}$+{variable} =/sg' $files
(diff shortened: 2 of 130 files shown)


JelteF:
Discussed this with @onderkalaci and this will be done after all the PRs for 9.1 are merged, but right before code freeze. That way there will be no merge conflicts for the PRs and also no merge conflicts when doing backports to the 9.1 branch.

PR #3183, opened 2019-11-13
Introduce intermediate result broadcasting

hanefi:
DESCRIPTION: Prunes intermediate result broadcasting

copy-pasted from source code
```c
 *
 * intermediate_result_pruning.c
 *   Functions for pruning intermediate result broadcasting.
 *
 * In plain words, each distributed plan pulls the necessary intermediate
 * results to the worker nodes that the plan hits. This is primarily useful
 * in three ways. First, if the distributed plan that uses intermediate
 * result(s) is a router query, then the intermediate results are only
 * broadcasted to a single node. Second, if a distributed plan consists
 * of only intermediate results, which is not uncommon, the intermediate
 * results are broadcasted to a single node only. Finally, if a distributed
 * query hits a sub-set of the shards in multiple workers, the intermediate
 * results will be broadcasted to the relevant node(s). The final item becomes
 * crucial for append/range distributed tables where typically the distributed
 * queries hit a small subset of shards/workers.
 *
 * To do this, for each query that Citus creates a distributed plan, we keep
 * track of the subPlans used in the queryTree, and save it in the distributed
 * plan. Just before Citus executes each subPlan, Citus first keeps track of
 * every worker node that the distributed plan hits, and marks every subPlan
 * should be broadcasted to these nodes. Later, for each subPlan which is a
 * distributed plan, Citus does this operation recursively since these
 * distributed plans may access to different subPlans, and those have to be
 * recorded as well.
 *
 *-------------------------------------------------------------------------
 */
```

See `Next Steps` section in https://github.com/citusdata/citus/issues/1804#issuecomment-348509606


src/backend/distributed/utils/node_metadata.c
```diff
@@ -895,6 +896,31 @@ FindWorkerNodeAnyCluster(const char *nodeName, int32 nodePort)
 }
 
 
+/*
+ * GetWorkerNodeByNodeId iterates over worker nodes and returns a pointer to the
+ * one that has the given id
+ */
+WorkerNode *
+GetWorkerNodeByNodeId(int nodeId)
```
hanefi: There seems to be a better alternative to this function here: https://github.com/citusdata/citus/blob/eb35743c3fd937e1b96c1ee39979aa9dead6d222/src/backend/distributed/utils/metadata_cache.c#L578-L583

The main difference is I used the index on `pg_dist_node`, whereas this one uses `WorkerNodeArray` that is already in a global variable in memory.

I guess I better remove my function and use this one here

src/backend/distributed/executor/subplan_execution.c
```diff
@@ -81,3 +124,99 @@ ExecuteSubPlans(DistributedPlan *distributedPlan)
...
+		entry = hash_search(intermediateResultsHash, &key, HASH_ENTER, &found);
+
+		if (!found)
+		{
+			entry->nodeList = NIL;
+		}
+
+		if (list_length(entry->nodeList) == GetWorkerNodeCount())
+		{
+			entry->containsAllNodes = true;
```
marcocitus: We should just break here, also when this condition occurs in AppendAllAccessedWorkerNodes.

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -556,6 +558,12 @@ CreateDistributedPlannedStmt(uint64 planId, PlannedStmt *localPlan, Query *origi
 		RaiseDeferredError(distributedPlan->planningError, ERROR);
 	}
 
+	/* traverse distributedPlan and extract the list of used subplans */
+	if (ShouldFindSubPlansUsedInPlan(distributedPlan))
+	{
+		FindSubPlansUsedInPlan(distributedPlan);
```
onderkalaci: What happens if we use the originalQuery instead of `distributedPlan == ->jobQuery`, I don't want to rely on `jobQuery` field, which is very much dependent how the distributed planning happens.


Do you see anything that prevents us using `originalQuery` to find the intermediate results?

src/backend/distributed/executor/subplan_execution.c
hanefi: makes sense. Removed the unused flag, and added early returns in 2 places in commit 5df2a6c471421e9c9a5a4d526860bae9d89a7968

src/backend/distributed/executor/subplan_execution.c
```diff
@@ -81,3 +117,148 @@ ExecuteSubPlans(DistributedPlan *distributedPlan)
...
+SearchIntermediateResult(HTAB *intermediateResultHash, char *resultId, bool *found)
+{
+	IntermediateResultHashKey key;
+	IntermediateResultHashEntry *entry = NULL;
+
+	strlcpy(key.intermediate_result_id, resultId, NAMEDATALEN);
+	entry = hash_search(intermediateResultHash, &key, HASH_ENTER, found);
+
+	/* if the key was not found, update the entry with sane defaults */
+	if (!found)
```
onderkalaci: if (!found) > if (!*found)

src/backend/distributed/executor/subplan_execution.c
```diff
@@ -81,3 +117,148 @@ ExecuteSubPlans(DistributedPlan *distributedPlan)
...
+}
+
+
+/**
+ * SearchIntermediateResult searches for an entry in a hash table and creates a
+ * new entry if the search key is not present.
+ *
+ */
+static IntermediateResultHashEntry *
+SearchIntermediateResult(HTAB *intermediateResultHash, char *resultId, bool *found)
```
onderkalaci: why do we pass found? Can't we check `entry->nodeList` instead of `found` in the above functions?

src/backend/distributed/executor/subplan_execution.c
```diff
@@ -81,3 +117,148 @@ ExecuteSubPlans(DistributedPlan *distributedPlan)
...
+/*
+ * AppendAllAccessedWorkerNodes iterates over all the tasks in a distributed plan
+ * to create the list of worker nodes that can be accessed when this plan is executed.
+ *
+ * If there are multiple placements of a Shard, all of them are considered.
+ */
+static List *
+AppendAllAccessedWorkerNodes(List *workerNodeList, DistributedPlan *distributedPlan)
+{
+	List *taskList = distributedPlan->workerJob->taskList;
```
onderkalaci: Instead of calling `GetWorkerNodeCount()` too many times, maybe set to a variable (here and the other usages as well)

src/backend/distributed/executor/subplan_execution.c
```diff
@@ -69,6 +74,44 @@ ExecuteSubPlans(DistributedPlan *distributedPlan)
...
+
+		strlcpy(key.intermediate_result_id, resultId, NAMEDATALEN);
+		entry = hash_search(intermediateResultHash, &key, HASH_ENTER, &found);
+
+		if (found)
+		{
+			ListCell *lcInner = NULL;
+			foreach(lcInner, entry->nodeList)
+			{
+				WorkerNode *w = LookupNodeByNodeId(lfirst_int(lcInner));
```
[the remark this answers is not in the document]
hanefi: Replaced them with `workerNode` and `nodeIdCell` respectively. Thanks for pointing it out

src/include/distributed/intermediate_results.h
```diff
@@ -27,6 +28,19 @@ extern DestReceiver * CreateRemoteFileDestReceiver(char *resultId, EState *execu
 extern void ReceiveQueryResultViaCopy(const char *resultId);
 extern void RemoveIntermediateResultsDirectory(void);
 extern int64 IntermediateResultSize(char *resultId);
+extern HTAB * makeIntermediateResultHTAB(void);
+
+typedef struct IntermediateResultHashKey
```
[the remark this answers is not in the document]
hanefi: moved them

src/include/distributed/intermediate_results.h
```diff
@@ -27,6 +28,19 @@ extern DestReceiver * CreateRemoteFileDestReceiver(char *resultId, EState *execu
...
+{
+	char intermediate_result_id[NAMEDATALEN];
+} IntermediateResultHashKey;
+
+typedef struct IntermediateResultHashEntry
+{
+	IntermediateResultHashKey key;
+	List *nodeList;
```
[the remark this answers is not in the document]
hanefi: replaced all occurences of this field, thanks

src/backend/distributed/executor/subplan_execution.c
```diff
@@ -69,6 +74,44 @@ ExecuteSubPlans(DistributedPlan *distributedPlan)
 
 		char *resultId = GenerateResultId(planId, subPlanId);
 
+		IntermediateResultHashKey key;
```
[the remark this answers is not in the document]
hanefi: Created `SearchIntermediateResult()` that searches and places sane defaults if the entry is not found and hence newly created.

src/backend/distributed/executor/subplan_execution.c
```diff
@@ -69,6 +74,44 @@ ExecuteSubPlans(DistributedPlan *distributedPlan)
...
+				WorkerNode *w = LookupNodeByNodeId(lfirst_int(lcInner));
+
+				nodeList = lappend(nodeList, w);
+				elog(DEBUG4, "%s is send to %s:%d", key.intermediate_result_id,
+					 w->workerName, w->workerPort);
+			}
+		}
+		else
+		{
+			/* TODO: Think more on this. Consider HAVING. */
```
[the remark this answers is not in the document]
hanefi: I read through the codebase, and tested with some queries with `HAVING` clauses. I will need to revisit this

src/backend/distributed/executor/subplan_execution.c
```diff
@@ -56,18 +66,49 @@ ExecuteSubPlans(DistributedPlan *distributedPlan)
...
+
+		if (list_length(entry->nodeIdList) > 0)
+		{
+			ListCell *nodeIdCell = NULL;
+			foreach(nodeIdCell, entry->nodeIdList)
+			{
+				WorkerNode *workerNode = LookupNodeByNodeId(lfirst_int(nodeIdCell));
+
+				nodeList = lappend(nodeList, workerNode);
+				elog(DEBUG4, "%s is send to %s:%d", resultId, workerNode->workerName,
```
onderkalaci: Can you make this as `DEBUG1`(and the below), otherwise adding regression tests become super hard.

hanefi: done

src/backend/distributed/executor/subplan_execution.c
hanefi: The variable is declared as `bool found = false;` above. So I think this one is actually correct.

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -582,6 +590,76 @@ CreateDistributedPlannedStmt(uint64 planId, PlannedStmt *localPlan, Query *origi
...
+
+			if (resultIdConst == NULL)
+			{
+				continue;
+			}
+
+			resultIdDatum = resultIdConst->constvalue;
+			resultId = TextDatumGetCString(resultIdDatum);
+
+			elog(DEBUG4, "resultId: %s", resultId);
```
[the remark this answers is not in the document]
hanefi: changed it to:
`elog(DEBUG4, "appending %s subplan to usedSubPlanNodeList of plan %lu",
				 resultId, plan->planId);`

src/backend/distributed/executor/subplan_execution.c
hanefi: no longer passing `found`. now that the function also places sane defaults ( a `NIL` in `nodeList` field ) it is no longer necessary

src/backend/distributed/executor/subplan_execution.c
```diff
@@ -56,18 +66,49 @@ ExecuteSubPlans(DistributedPlan *distributedPlan)
...
 		DistributedSubPlan *subPlan = (DistributedSubPlan *) lfirst(subPlanCell);
 		PlannedStmt *plannedStmt = subPlan->plan;
 		uint32 subPlanId = subPlan->subPlanId;
+
 		DestReceiver *copyDest = NULL;
 		ParamListInfo params = NULL;
 		EState *estate = NULL;
 
 		char *resultId = GenerateResultId(planId, subPlanId);
+		IntermediateResultsHashEntry *entry = SearchIntermediateResult(
```
onderkalaci: I think my comment is lost: Can we move  this logic to a function where we give `resultId ` and get `nodeIdList`? If `nodeIdList == NIL`, we write to a local  file

src/backend/distributed/executor/subplan_execution.c
```diff
@@ -25,6 +26,15 @@ int MaxIntermediateResult = 1048576; /* maximum size in KB the intermediate resu
 int SubPlanLevel = 0;
 
 
+static List * AppendAllAccessedWorkerNodes(List *workerNodeList,
+										   DistributedPlan *distributedPlan);
+static void RecordSubplanExecutionsOnNodes(HTAB *intermediateResultHash,
+										   DistributedPlan *distributedPlan);
+static HTAB * MakeIntermediateResultHTAB(void);
+static IntermediateResultHashEntry * SearchIntermediateResult(
```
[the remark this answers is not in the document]
hanefi: I started using `citus_indent` in git commit hook. I guess it is not so safe to use hooks 🤷‍♂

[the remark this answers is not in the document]
hanefi: The only alternative `ciitus_indent` is happy with is this:

```C
static IntermediateResultsHashEntry * SearchIntermediateResult(HTAB
															   *intermediateResultsHash,
															   char *resultId);
```

I guess I'll update the indentation with this

src/backend/distributed/executor/subplan_execution.c
hanefi: No longer calling it multiple times. setting it to a variable and sending it to `AppendAllAccessedWorkerNodes()` as a param

src/backend/distributed/executor/subplan_execution.c
hanefi: I misunderstood your previous comment. I created `SearchIntermediateResult` to address that comment. As I see it know there is some more room for improvement here. 

What you are suggesting here seems like an elegant and cleaner way to implement this. I will move the logic here to a seperate function

src/test/regress/expected/multi_function_in_join.out
```diff
@@ -176,6 +196,8 @@ $$ language plpgsql;
 SELECT * FROM table1 JOIN max_and_min() m ON (m.maximum = data OR m.minimum = data) ORDER BY 1,2,3,4;
 DEBUG:  generating subplan 14_1 for subquery SELECT minimum, maximum FROM functions_in_joins.max_and_min() m(minimum, maximum)
 DEBUG:  Plan 14 query after replacing subqueries and CTEs: SELECT table1.id, table1.data, m.minimum, m.maximum FROM (functions_in_joins.table1 JOIN (SELECT intermediate_result.minimum, intermediate_result.maximum FROM read_intermediate_result('14_1'::text, 'binary'::citus_copy_format) intermediate_result(minimum integer, maximum integer)) m ON (((m.maximum OPERATOR(pg_catalog.=) table1.data) OR (m.minimum OPERATOR(pg_catalog.=) table1.data)))) ORDER BY table1.id, table1.data, m.minimum, m.maximum
+DEBUG:  14_1 is send to localhost:57637
```
onderkalaci: Should this end with `t` -> `sent`?

hanefi: `14_1 is broadcasted to localhost:57637` sounds better to me. I guess I will update all those 30+ test outputs one more time

src/backend/distributed/executor/subplan_execution.c
hanefi: Ended up having
`if (workerNodeList == NIL) { .. }`
as we have a list of `WorkerNode`s at this point.

src/backend/distributed/executor/subplan_execution.c
```diff
@@ -66,13 +77,27 @@ ExecuteSubPlans(DistributedPlan *distributedPlan)
...
 		char *resultId = GenerateResultId(planId, subPlanId);
+		List *workerNodeList =
+			FindAllWorkerNodesUsingSubplan(intermediateResultsHash,
+										   resultId);
+
+		/*
+		 * Write intermediate results to local file only if there is no worker
+		 * node that receives them
+		 */
+		if (workerNodeList == NIL)
```
onderkalaci: This  could happen in  two cases: (a) Subquery in  the having (b) The intermediate result is not used, such as RETURNING of a modifying CTE is not used -- for SELECT, Postgres/Citus is cleaver enough to not execute the CTE if it is not used at all, but for modifications we have to execute the queries.

So, please add a comment how this could happen

src/test/regress/expected/multi_function_in_join.out
onderkalaci: sorry about that : /

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -71,6 +71,8 @@ static DistributedPlan * CreateDistributedPlan(uint64 planId, Query *originalQue
 											   bool hasUnresolvedParams,
 											   PlannerRestrictionContext *
 											   plannerRestrictionContext);
+static void FindSubPlansUsedInPlan(DistributedPlan *plan);
```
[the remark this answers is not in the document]
hanefi: I created `intermediate_result_pruning.c` and `intermediate_result_pruning.h` and moved all the pruning logic in executor and planner to there.

See my two commits
- distributed planner : 9a8f66f76cba85de6d06c2bab2b6d486e6da6d84 
- executor : 05ef4e1fa8413a892f8e4fa60f93506cc89c7339

src/backend/distributed/executor/subplan_execution.c
hanefi: improved the comments above.


		/*
		 * Write intermediate results to local file only if there is no worker
		 * node that receives them.
		 *
		 * This could happen in two cases:
		 * (a) Subquery in the having
		 * (b) The intermediate result is not used, such as RETURNING of a
		 *     modifying CTE is not used
		 *
		 * For SELECT, Postgres/Citus is clever enough to not execute the CTE
		 * if it is not used at all, but for modifications we have to execute
		 * the queries.
		 */

src/backend/distributed/planner/intermediate_result_pruning.c
```diff
@@ -0,0 +1,257 @@
...
+
+			if (resultIdConst == NULL)
+			{
+				continue;
+			}
+
+			plan->usedSubPlanNodeList = list_append_unique(plan->usedSubPlanNodeList,
+														   resultIdConst);
+
+			resultIdDatum = resultIdConst->constvalue;
```
onderkalaci: we should use to skip doing things if the debug level is not high enough

```C
if (IsLoggable(DEBUG4))
{
resultIdDatum = resultIdConst->constvalue;
			resultId = TextDatumGetCString(resultIdDatum);
elog(DEBUG4, "appending %s subplan to usedSubPlanNodeList of plan %lu",
				 resultId, plan->planId);

}

File: src/backend/distributed/planner/intermediate_result_pruning.c [hunk 20]
```diff
@@ -0,0 +1,257 @@
+/*-------------------------------------------------------------------------
+ *
+ * intermediate_result_pruning.c
```
onderkalaci: Let's add the main logic here:

In plain words, each distributed plan pulls the necessary intermediate results to the worker nodes that the plan hits.

For each query that Citus creates a distributed plan, we keep track of the subPlans used in the queryTree, and save it in the distributed plan. Just before Citus executes each subPlan, Citus first keeps track of every worker node that the distributed plan hits, and marks every subPlan should be broadcasted to these nodes. Later, for each subPlan which is a distributed plan, Citus does this operation recursively since these distributed plans may access to different subPlans, and those have to be recorded as well.

src/backend/distributed/planner/intermediate_result_pruning.c
```diff
@@ -0,0 +1,257 @@
...
+ * plan->usedSubPlanNodeList field. Simply traversing the range table entries in the plan.
+ */
+void
+FindSubPlansUsedInPlan(DistributedPlan *plan)
+{
+	Query *jobQuery = NULL;
+	List *rangeTableList = NIL;
+	ListCell *rangeTableCell = NULL;
+
+	jobQuery = plan->workerJob == NULL ? plan->insertSelectSubquery :
```
onderkalaci: Don't forget to use `originalQuery` instead of `jobQuery`

src/backend/distributed/planner/multi_logical_planner.c
```diff
@@ -807,6 +808,33 @@ IsReadIntermediateResultFunction(Node *node)
 }
 
 
+/*
+ * FindResultIdOfIntermediateResultFunction extracts the id of the intermediate result
+ * if the given RTE contains a read_intermediate_results function, NULL otherwise
+ */
+Const *
+FindResultIdOfIntermediateResultFunction(RangeTblEntry *rte)
```
onderkalaci: The name is a little bit confusing. When I read the name, I expect the input to be intermediate result function, but that's not the case.

Maybe: `FindIntermediateResultIdIfExists()` or such could be better

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -556,6 +557,9 @@ CreateDistributedPlannedStmt(uint64 planId, PlannedStmt *localPlan, Query *origi
 		RaiseDeferredError(distributedPlan->planningError, ERROR);
 	}
 
+	/* traverse distributedPlan and extract the list of used subplans */
+	FindSubPlansUsedInPlan(distributedPlan);
```
marcocitus: This API is confusing, maybe we can do `distributedPlan->usedSubPlanNodeList = FindSubPlansUsedInPlan(distributedPlan);`?

src/backend/distributed/planner/intermediate_result_pruning.c
```diff
@@ -0,0 +1,257 @@
...
+				continue;
+			}
+
+			plan->usedSubPlanNodeList = list_append_unique(plan->usedSubPlanNodeList,
+														   resultIdConst);
+
+			resultIdDatum = resultIdConst->constvalue;
+			resultId = TextDatumGetCString(resultIdDatum);
+			elog(DEBUG4, "appending %s subplan to usedSubPlanNodeList of plan %lu",
+				 resultId, plan->planId);
```
marcocitus: minor: it's a little confusing that we say "appending...", but we actually already appended. Maybe move the append below?

src/backend/distributed/planner/distributed_planner.c
onderkalaci: I suggested the same, but `distributedPlan->usedSubPlanNodeList = findSubPlansUsedInQuery(originalQuery)`, because that seems safer than looking into `jobQuery` or `insertSelectQuery` etc

src/backend/distributed/executor/citus_custom_scan.c
```diff
@@ -398,3 +398,99 @@ ScanStateGetExecutorState(CitusScanState *scanState)
...
+	{
+		return false;
+	}
+
+	if (IsCitusCustomScan(plan))
+	{
+		return true;
+	}
+
+	if (plan->lefttree != NULL && IsCitusPlan(plan->lefttree))
```
SaitTalhaNisanci: At the beginning of the function you have the null check, so you instead of the following part you can just return this:

```
return IsCitusPlan(plan->leftTree) || isCitusPlan(plan->rightTree)
```

src/test/regress/expected/multi_function_in_join.out
SaitTalhaNisanci: @onderkalaci I think we should use `sent` or something similar because `broadcast` means sending to every node, so to me it seems incorrect to use that when we are not actually `broadcasting` that. We are just sending it to some nodes.

src/backend/distributed/planner/intermediate_result_pruning.c
```diff
@@ -0,0 +1,257 @@
...
+			entry->nodeIdList = AppendAllAccessedWorkerNodes(entry->nodeIdList,
+															 distributedPlan,
+															 workerNodeCount);
+
+			elog(DEBUG4, "subplan %s is used in %lu", resultId, distributedPlan->planId);
+		}
+	}
+
+	/* descend into the subPlans */
+	foreach(subPlanCell, subPlanList)
```
marcocitus: is this still necessary if we send to all nodes?

src/backend/distributed/planner/intermediate_result_pruning.c
```diff
@@ -0,0 +1,257 @@
...
+	ListCell *taskCell = NULL;
+
+	foreach(taskCell, taskList)
+	{
+		Task *task = lfirst(taskCell);
+		ListCell *placementCell = NULL;
+		foreach(placementCell, task->taskPlacementList)
+		{
+			ShardPlacement *placement = lfirst(placementCell);
+			workerNodeList = list_append_unique_int(workerNodeList, placement->nodeId);
```
marcocitus: ideally, we'd use a hash table for sets.

src/backend/distributed/planner/intermediate_result_pruning.c
```diff
@@ -0,0 +1,257 @@
...
+	{
+		Const *resultIdConst = lfirst(usedSubPlanNodeCell);
+		Datum resultIdDatum = resultIdConst->constvalue;
+		char *resultId = TextDatumGetCString(resultIdDatum);
+
+		IntermediateResultsHashEntry *entry = SearchIntermediateResult(
+			intermediateResultsHash, resultId);
+
+		/* no need to traverse the whole plan if all the workers are hit */
+		if (list_length(entry->nodeIdList) == workerNodeCount)
```
marcocitus: maybe move before the previous statement to avoid doing any unnecessary work

src/backend/distributed/planner/intermediate_result_pruning.c
```diff
@@ -0,0 +1,257 @@
...
+	ListCell *usedSubPlanNodeCell = NULL;
+	List *subPlanList = distributedPlan->subPlanList;
+	ListCell *subPlanCell = NULL;
+	int workerNodeCount = GetWorkerNodeCount();
+
+	foreach(usedSubPlanNodeCell, usedSubPlanNodeList)
+	{
+		Const *resultIdConst = lfirst(usedSubPlanNodeCell);
+		Datum resultIdDatum = resultIdConst->constvalue;
+		char *resultId = TextDatumGetCString(resultIdDatum);
```
marcocitus: could we store a string in the list instead of Const? seems a bit odd to store the Const

src/test/regress/expected/multi_function_in_join.out
onderkalaci: By no means I'm an english expert, and you might be right. But,  `broadcast` does not necessarily feel `to every`, it feels closer to `spread` to me. And, in the code there is no reference to `send` in this context, so I'd rather use something else.

Given that this is a `DEBUG` message, I don't want to ask updating too many files again. If that's disturbing you, feel free to open a PR on top of this to fix it.

src/backend/distributed/planner/intermediate_result_pruning.c
onderkalaci: We considered that as well, but couldn't make sure that it'd always (say more common case , worker count < 8) perform better.

Do you prefer that? If we had some more time, we wanted to benchmark this a bit, but couldn't get there :/

src/test/regress/expected/multi_function_in_join.out
SaitTalhaNisanci: I think we can use `multicast`. From https://techdifferences.com/difference-between-broadcast-and-multicast.html :

```
The transmission technologies broadcast and multicast are different from each other in a
context that in the broadcast, the packet is forwarded to all the hosts connected to the
network whereas, in multicast, the packet is forwarded only to intended recipients.
```

I guess we are doing multicasting, so I would suggest to use `multicast`. I can update it if you guys agree

src/backend/distributed/executor/subplan_execution.c
```diff
@@ -66,13 +64,35 @@ ExecuteSubPlans(DistributedPlan *distributedPlan)
...
+		 *     modifying CTE is not used
+		 *
+		 * For SELECT, Postgres/Citus is clever enough to not execute the CTE
+		 * if it is not used at all, but for modifications we have to execute
+		 * the queries.
+		 */
+		if (workerNodeList == NIL)
+		{
+			writeLocalFile = true;
+			elog(DEBUG1, "%s is not broadcasted to any node, write to local only",
```
SaitTalhaNisanci: At this point it is not written to a local file. I think in the future this might make debugging harder because we will see this debugging message, thinking that it is written to a local file, but maybe there is a failure so it is not written to a local file. Maybe we can have something like `will write to only local`

src/backend/distributed/planner/intermediate_result_pruning.c
```diff
@@ -0,0 +1,257 @@
...
+			}
+		}
+	}
+
+	return workerNodeList;
+}
+
+
+/*
+ * MakeIntermediateResultHTAB is a helper method that creates a Hash Table that store information on the intermediate result.
```
SaitTalhaNisanci: `store` -> `stores`

src/backend/distributed/planner/intermediate_result_pruning.c
```diff
@@ -0,0 +1,257 @@
...
+	IntermediateResultsHashEntry *entry =
+		SearchIntermediateResult(intermediateResultsHash, resultId);
+
+	ListCell *nodeIdCell = NULL;
+	foreach(nodeIdCell, entry->nodeIdList)
+	{
+		WorkerNode *workerNode = LookupNodeByNodeId(lfirst_int(nodeIdCell));
+
+		workerNodeList = lappend(workerNodeList, workerNode);
+		elog(DEBUG1, "%s is broadcasted to %s:%d", resultId, workerNode->workerName,
```
SaitTalhaNisanci: Maybe I am missing a point but at this point, are the results broadcasted?

src/backend/distributed/planner/distributed_planner.c
marcocitus: bump

hanefi: I am working on this one now. Will get it working before merging the PR

src/backend/distributed/executor/citus_custom_scan.c
hanefi: makes sense. thanks for the advice

src/backend/distributed/planner/intermediate_result_pruning.c
```diff
@@ -0,0 +1,259 @@
...
+			if (resultIdConst == NULL)
+			{
+				continue;
+			}
+
+			subPlanList = list_append_unique(subPlanList, resultIdConst);
+
+			resultIdDatum = resultIdConst->constvalue;
+			resultId = TextDatumGetCString(resultIdDatum);
+			elog(DEBUG4, "Results of SubPlan %s is used in Plan %lu",
```
marcocitus: Results -> result

src/backend/distributed/planner/intermediate_result_pruning.c
marcocitus: bump

src/backend/distributed/planner/intermediate_result_pruning.c
marcocitus: bump

src/backend/distributed/planner/intermediate_result_pruning.c
hanefi: We can not use `list_append_unique()` for strings as the function checks for simple equality for pointers whereas deep equality is checked for numerous PostgreSQL structs such as `Const`

I am not so proud of this piece of code, but I found it hard to do it elegantly.

src/backend/distributed/planner/intermediate_result_pruning.c
hanefi: entry pointer is updated in the previous statement. I am afraid it is not possible to change the ordering here

src/backend/distributed/planner/intermediate_result_pruning.c
hanefi: This is actually not the part of code that does the broadcasting. Here we only determine which worker will receive the intermediate results as this is part of distributed planner.

It might have been more appropriate to say `will be broadcasted to` or `will be multicasted to` here

src/backend/distributed/planner/intermediate_result_pruning.c
hanefi: Done

src/backend/distributed/planner/multi_logical_planner.c
hanefi: done

src/backend/distributed/planner/intermediate_result_pruning.c
hanefi: `RecordSubplanExecutionsOnNodes()` is recursively called for every subplan of a plan. Currently at every call we check node count with `GetWorkerNodeCount()`

I can easily add a new parameter to this recursive function to keep the number of nodes, and possibly create a wrapper that can initialize the node count and pass it to the first call of the original function.

Do you think increasing the complexity here is worth it?

src/backend/distributed/planner/intermediate_result_pruning.c
hanefi: `list_append_unique` does not work well with pointers. If I were to use strings here, I would also need several helper functions that check for uniqueness before appending, and I did not want to do that.

src/test/regress/expected/multi_function_in_join.out
hanefi: I am not really familiar with the word `multicast` but it seems like a better fit in this scenario.

One more idea @SaitTalhaNisanci suggested makes sense as well: using future tense instead of past tense. As this is part of planner, it can be helpful to stress that the data is not actually sent to the worker at the time of logging.

src/backend/distributed/planner/intermediate_result_pruning.c
marcocitus: Hmz, that's not ideal. Maybe we should try to change to a hash table.

src/backend/distributed/executor/subplan_execution.c
marcocitus: Agreed, this should be rephrased, or maybe just omitted entirely.

src/backend/distributed/planner/distributed_planner.c
hanefi: We lose some of the information, and end up not sending the intermediate results to some of the workers.

You can see one failed attempt here: https://github.com/citusdata/citus/tree/hanefi-failing-originalquery

src/backend/distributed/planner/intermediate_result_pruning.c
hanefi: We lose some of the information when we use originalQuery, and end up not sending the intermediate results to some of the workers.

You can see one failed attempt here: https://github.com/citusdata/citus/tree/hanefi-failing-originalquery

src/backend/distributed/planner/distributed_planner.c
hanefi: Marco's suggestion is completed yesterday. I was not able to use `originalQuery` instead ob `distributedPlan->jobQuery` or `distributedPlan->insertSelectQuery` though.

src/backend/distributed/executor/subplan_execution.c
hanefi: This debug message is really helpful in the regression tests. I am more inclined to rephrase it. How does `Subplan %s will not broadcasted to any node, will write to local only` sound?

hanefi: Updated all my `DEBUG1` messages. Now they are in two different forms:
```
DEBUG:  Subplan 18_1 will not broadcasted to any node, will write to local only
DEBUG:  Subplan 18_2 will be broadcasted to localhost:57637
```

SaitTalhaNisanci: It should be `will not be broadcasted`.

src/backend/distributed/executor/subplan_execution.c
```diff
@@ -66,14 +64,35 @@ ExecuteSubPlans(DistributedPlan *distributedPlan)
...
+		 *
+		 * For SELECT, Postgres/Citus is clever enough to not execute the CTE
+		 * if it is not used at all, but for modifications we have to execute
+		 * the queries.
+		 */
+		if (workerNodeList == NIL)
+		{
+			writeLocalFile = true;
+			elog(DEBUG1, "Subplan %s will not broadcasted to any node, "
+						 "will write to local only", resultId);
```
marcocitus: I don't think these debug messages add a lot of long-term value. Maybe it should be DEBUG5?

broadcasted -> sent

src/backend/distributed/planner/intermediate_result_pruning.c
```diff
@@ -0,0 +1,293 @@
...
+		SearchIntermediateResult(intermediateResultsHash, resultId);
+
+	ListCell *nodeIdCell = NULL;
+	foreach(nodeIdCell, entry->nodeIdList)
+	{
+		WorkerNode *workerNode = LookupNodeByNodeId(lfirst_int(nodeIdCell));
+
+		workerNodeList = lappend(workerNodeList, workerNode);
+		elog(DEBUG1, "Subplan %s will be broadcasted to %s:%d", resultId,
+			 workerNode->workerName, workerNode->workerPort);
```
marcocitus: same here, broadcasted->sent and DEBUG5

onderkalaci: If we do that, we won't have any explicit tests for showing the pruning works correctly. So, I prefer to keep it as `DEBUG1`, not for usefulness for users, but for testing.

src/backend/distributed/executor/subplan_execution.c
hanefi:
```diff
-DEBUG:  Subplan 184_1 will not broadcasted to any node, will write to local only
+DEBUG:  Subplan 184_1 will not be broadcasted to any node, will be written to local only
```

src/backend/distributed/planner/intermediate_result_pruning.c
onderkalaci: @marcocitus are you OK with that? The debug message is also not irrelevant, when the log level is DEBUG1, we already print bunch of things for recursive planning.

marcocitus: No, let's just change client_min_messages and show it in 3-4 tests.

onderkalaci:
> Please reduce the debug messages to a few select tests.

I think we should definitely keep all the tests in the newly added test file which tests various edge cases. So the option I can come up with is adding a hidden GUC, and print this log either DEBUG5 ( DEBUG4 sounds better to me) or that GUC is set to true. And, we only set it on that file.

marcocitus:
> I think we should definitely keep all the tests in the newly added test file which tests various edge cases. So the option I can come up with is adding a hidden GUC, and print this log either DEBUG5 ( DEBUG4 sounds better to me) or that GUC is set to true. And, we only set it on that file.

Works for me.

src/backend/distributed/executor/subplan_execution.c
hanefi: I hid the message behind a GUC `log_intermediate_results`

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -556,6 +558,12 @@ CreateDistributedPlannedStmt(uint64 planId, PlannedStmt *localPlan, Query *origi
 		RaiseDeferredError(distributedPlan->planningError, ERROR);
 	}
 
+	/* traverse distributedPlan and extract the list of used subplans */
+	if (ShouldFindSubPlansUsedInPlan(distributedPlan))
```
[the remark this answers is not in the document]
hanefi: 👍 done

src/backend/distributed/planner/intermediate_result_pruning.c
```diff
@@ -0,0 +1,301 @@
...
+ * If an entry is not found, creates a new entry with sane defaults.
+ */
+static IntermediateResultsHashEntry *
+SearchIntermediateResult(HTAB *intermediateResultsHash, char *resultId)
+{
+	IntermediateResultsHashKey key;
+	IntermediateResultsHashEntry *entry = NULL;
+	bool found = false;
+
+	strlcpy(key.intermediate_result_id, resultId, NAMEDATALEN);
```
marcocitus: This seems to reimplement logic that is an integral part of the hash table.

src/test/regress/expected/multi_function_in_join.out
hanefi: We hid new Debug messages behind a GUC `log_intermediate_results`

The final log messages are as follows:

`Subplan %s will be written to local file`
and 
`Subplan %s will be sent to %s:%d`

They will appear only when `the GUC is set, or the min_messages is `DEBUG4` or below

src/backend/distributed/planner/intermediate_result_pruning.c
hanefi: I created a hidden GUC `log_intermediate_results` that handles the debug outputs for this case.

By default the messages will be shown if `client_min_messages` is `DEBUG4` or below. If the GUC is set, they will be shown for `DEBUG1` and below.

marcocitus:
Pushed a commit that changes Const to Value, which is a little more standard.

src/backend/distributed/planner/intermediate_result_pruning.c
```diff
@@ -0,0 +1,275 @@
...
+ */
+HTAB *
+MakeIntermediateResultHTAB()
+{
+	HTAB *intermediateResultsHash = NULL;
+	uint32 hashFlags = 0;
+	HASHCTL info = { 0 };
+	int initialNumberOfElements = 16;
+
+	info.keysize = sizeof(NAMEDATALEN);
```
marcocitus: should be `= NAMEDATALEN;`

src/include/distributed/subplan_execution.h
```diff
@@ -19,5 +19,20 @@ extern int SubPlanLevel;
...
+ * the list of nodes that can possibly run a query that will use the
+ * intermediate results.
+ *
+ * The nodeIdList contains a set of unique WorkerNode ids that have placements
+ * that can be used in non-colocated subquery joins with the intermediate result
+ * given in the key.
+ */
+typedef struct IntermediateResultsHashEntry
+{
+	char *key;
```
marcocitus: actually, it's probably more convenient to put `char key[NAMEDATALEN];` here

Otherwise you would have to not use HASH_ELEM and allocate IntermediateResultsHashEntry yourself.

onderkalaci: yes, that seems more common, I was writing a similar comment :) 

Check `plperl_query_entry` @hanefi for an example

src/backend/distributed/planner/intermediate_result_pruning.c
```diff
@@ -0,0 +1,275 @@
...
+ */
+HTAB *
+MakeIntermediateResultHTAB()
+{
+	HTAB *intermediateResultsHash = NULL;
+	uint32 hashFlags = 0;
+	HASHCTL info = { 0 };
+	int initialNumberOfElements = 16;
+
+	info.keysize = sizeof(char *);
```
marcocitus: Right, that's better. Thanks!

src/backend/distributed/planner/intermediate_result_pruning.c
hanefi: No, It should have been `sizeof(char*)` here.
`= NAMEDATALEN;` ends up in backend crashes

src/backend/distributed/planner/intermediate_result_pruning.c
```diff
@@ -0,0 +1,275 @@
...
+ * results to the worker nodes that the plan hits. This is primarily useful
+ * in three ways. First, if the distributed plan that uses intermediate
+ * result(s) is a router query, then the intermediate results are only
+ * broadcasted to a single node. Second, if a distributed plan consists
+ * of only intermediate results, which is not uncommon, the intermediate
+ * results are broadcasted to a single node only. Finally, if a distributed
+ * query hits a sub-set of the shards in multiple workers, the intermediate
+ * results will be broadcasted to the relevant node(s). The final item becomes
+ * crucial for append/range distributed tables where typically the distributed
+ * queries hit a small subset of shards/workers.
```
marcocitus: This is a bit of a wall of text, and will need maintenance, e.g. "Second, if a distributed plan consists of only intermediate results, which is not uncommon, the intermediate results are broadcasted to a single node only." hopefully won't be true for long. 

I think the explanation in the paragraph below is useful, but here maybe just say something like: "We only send intermediate results of subqueries and CTEs to worker nodes that use them in the remainder of the distributed plan to avoid unnecessary network traffic. "

PR #3184, opened 2019-11-13
Alter Role Propagation

halilozanakgul:
DESCRIPTION: Add ALTER ROLE propagation and role propagation for master_add_node

### DIFF
--- a/src/backend/distributed/commands/role.c
+++ b/src/backend/distributed/commands/role.c
@@ -1,0 +1,281 @@
+/*-------------------------------------------------------------------------
+ *
+ * role.c
+ *    Commands for ALTER ROLE statements.
+ *
+ * Copyright (c) Citus Data, Inc.
+ *
+ *-------------------------------------------------------------------------
+ */
+
+#include "postgres.h"
+
+#include "access/heapam.h"
+#include "access/htup_details.h"
+#if PG_VERSION_NUM >= 120000
+#include "access/table.h"
+#endif
+#include "catalog/catalog.h"
+#include "catalog/pg_authid.h"
+#include "distributed/citus_ruleutils.h"
+#include "distributed/commands.h"
+#include "distributed/commands/utility_hook.h"
+#include "distributed/deparser.h"
+#include "distributed/master_protocol.h"
+#include "distributed/worker_transaction.h"
+#include "nodes/makefuncs.h"
+#include "nodes/parsenodes.h"
+#include "utils/acl.h"
+#include "utils/builtins.h"
+#include "utils/rel.h"
+#include "utils/syscache.h"
+
+static const char * ExtractEncryptedPassword(Oid roleOid);
+static const char * CreateAlterRoleIfExistsCommand(AlterRoleStmt *stmt);
+static DefElem * makeDefElemInt(char *name, int value);
+
+/* controlled via GUC */
+bool EnableAlterRolePropagation = false;
+
+/*
+ * ProcessAlterRoleStmt actually creates the plan we need to execute for alter
+ * role statement. We need to do it this way because we need to use the encrypted
+ * password, which is, in some cases, created at standardProcessUtility.
+ */
+List *
+ProcessAlterRoleStmt(AlterRoleStmt *stmt, const char *queryString)
+{
+	ListCell *optionCell = NULL;
+	List *commands = NIL;
+
+	if (!EnableAlterRolePropagation || !IsCoordinator())
+	{
+		return NIL;
+	}
+
+	/*
+	 * Make sure that no new nodes are added after this point until the end of the
+	 * transaction by taking a RowShareLock on pg_dist_node, which conflicts with the
+	 * ExclusiveLock taken by master_add_node.
+	 */
+	LockRelationOid(DistNodeRelationId(), RowShareLock);
+
+	foreach(optionCell, stmt->options)
+	{
+		DefElem *option = (DefElem *) lfirst(optionCell);
+
+		if (strcasecmp(option->defname, "password") == 0)
+		{
+			Oid roleOid = get_rolespec_oid(stmt->role, true);
+			const char *encryptedPassword = ExtractEncryptedPassword(roleOid);
+
+			if (encryptedPassword != NULL)
+			{
+				Value *encryptedPasswordValue = makeString((char *) encryptedPassword);
+				option->arg = (Node *) encryptedPasswordValue;
+			}
+			else
+			{
+				option->arg = NULL;
+			}
+
+			break;
+		}
+	}
+	commands = list_make1((void *) CreateAlterRoleIfExistsCommand(stmt));
+
+	return NodeDDLTaskList(ALL_WORKERS, commands);
+}
+
+
+/*
+ * CreateAlterRoleIfExistsCommand creates ALTER ROLE command, from the alter role node
+ *  using the alter_role_if_exists() UDF.
+ */
+static const char *
+CreateAlterRoleIfExistsCommand(AlterRoleStmt *stmt)
+{
+	StringInfoData alterRoleQueryBuffer = { 0 };
+	const char *roleName = RoleSpecString(stmt->role, false);
+	const char *alterRoleQuery = DeparseTreeNode((Node *) stmt);
+
+	initStringInfo(&alterRoleQueryBuffer);
+	appendStringInfo(&alterRoleQueryBuffer,
+					 "SELECT alter_role_if_exists(%s, %s)",
+					 quote_literal_cstr(roleName),
+					 quote_literal_cstr(alterRoleQuery));
+
+	return alterRoleQueryBuffer.data;
+}
+
+
+/*
+ * ExtractEncryptedPassword extracts the encrypted password of a role. The function
+ * gets the password from the pg_authid table.
+ */
+static const char *
+ExtractEncryptedPassword(Oid roleOid)
+{
+	Relation pgAuthId = heap_open(AuthIdRelationId, AccessShareLock);
+	TupleDesc pgAuthIdDescription = RelationGetDescr(pgAuthId);
+	HeapTuple tuple = SearchSysCache1(AUTHOID, roleOid);
+	bool isNull = true;
+	Datum passwordDatum;
+
+	if (!HeapTupleIsValid(tuple))
+	{
+		return NULL;
+	}
+
+	passwordDatum = heap_getattr(tuple, Anum_pg_authid_rolpassword,
+								 pgAuthIdDescription, &isNull);
+
+	heap_close(pgAuthId, AccessShareLock);
+	ReleaseSysCache(tuple);
+
+	if (isNull)
+	{
+		return NULL;
+	}
+
+	return pstrdup(TextDatumGetCString(passwordDatum));
+}
+
+
+/*
+ * GenerateAlterRoleIfExistsCommand generate ALTER ROLE command that copies a role from
+ * the pg_authid table.
+ */
+static const char *
+GenerateAlterRoleIfExistsCommand(HeapTuple tuple, TupleDesc pgAuthIdDescription)
+{
+	char *rolPassword = "";
+	char *rolValidUntil = "infinity";
+	Datum rolValidUntilDatum;
+	Datum rolPasswordDatum;
+	bool isNull = true;
+	Form_pg_authid role = ((Form_pg_authid) GETSTRUCT(tuple));
+	AlterRoleStmt *stmt = makeNode(AlterRoleStmt);
+	const char *rolename = NameStr(role->rolname);
+
+	stmt->role = makeNode(RoleSpec);
+	stmt->role->roletype = ROLESPEC_CSTRING;
+	stmt->role->location = -1;
+	stmt->role->rolename = pstrdup(rolename);
+	stmt->action = 1;
+	stmt->options = NIL;
+
+	stmt->options =
+		lappend(stmt->options,
+				makeDefElemInt("superuser", role->rolsuper));
+
+	stmt->options =
+		lappend(stmt->options,
+				makeDefElemInt("createdb", role->rolcreatedb));
+
+	stmt->options =
+		lappend(stmt->options,
+				makeDefElemInt("createrole", role->rolcreaterole));
+
+	stmt->options =
+		lappend(stmt->options,
+				makeDefElemInt("inherit", role->rolinherit));
+
+	stmt->options =
+		lappend(stmt->options,
+				makeDefElemInt("canlogin", role->rolcanlogin));
+
+	stmt->options =
+		lappend(stmt->options,
+				makeDefElemInt("isreplication", role->rolreplication));
+
+	stmt->options =
+		lappend(stmt->options,
+				makeDefElemInt("bypassrls", role->rolbypassrls));
+
+
+	stmt->options =
+		lappend(stmt->options,
+				makeDefElemInt("connectionlimit", role->rolconnlimit));
+
+
+	rolPasswordDatum = heap_getattr(tuple, Anum_pg_authid_rolpassword,
+									pgAuthIdDescription, &isNull);
+	if (!isNull)
+	{
+		rolPassword = pstrdup(TextDatumGetCString(rolPasswordDatum));
+		stmt->options = lappend(stmt->options, makeDefElem("password",
+														   (Node *) makeString(
+															   rolPassword),
+														   -1));
+	}
+	else
+	{
+		stmt->options = lappend(stmt->options, makeDefElem("password", NULL, -1));
+	}
+
+	rolValidUntilDatum = heap_getattr(tuple, Anum_pg_authid_rolvaliduntil,
+									  pgAuthIdDescription, &isNull);
+	if (!isNull)
+	{
+		rolValidUntil = pstrdup((char *) timestamptz_to_str(rolValidUntilDatum));
+	}
+
+	stmt->options = lappend(stmt->options, makeDefElem("validUntil",
+													   (Node *) makeString(rolValidUntil),
+													   -1));
+
+	return CreateAlterRoleIfExistsCommand(stmt);
+}
+
+
+/*
+ * GenerateAlterRoleIfExistsCommandAllRoles creates ALTER ROLE commands
+ * that copies all roles from the pg_authid table.
+ */
+List *
+GenerateAlterRoleIfExistsCommandAllRoles()
+{
+	Relation pgAuthId = heap_open(AuthIdRelationId, AccessShareLock);
+	TupleDesc pgAuthIdDescription = RelationGetDescr(pgAuthId);
+	HeapTuple tuple = NULL;
+	List *commands = NIL;
+	const char *alterRoleQuery = NULL;
+
+#if PG_VERSION_NUM >= 120000
+	TableScanDesc scan = table_beginscan_catalog(pgAuthId, 0, NULL);
+#else
+	HeapScanDesc scan = heap_beginscan_catalog(pgAuthId, 0, NULL);
+#endif
+
+	while ((tuple = heap_getnext(scan, ForwardScanDirection)) != NULL)
+	{
+		const char *rolename = NameStr(((Form_pg_authid) GETSTRUCT(tuple))->rolname);
+
+		/*
+		 * The default roles are skipped, because reserved roles
+		 * cannot be altered.
+		 */
+		if (IsReservedName(rolename))
+		{
+			continue;
+		}
+		alterRoleQuery = GenerateAlterRoleIfExistsCommand(tuple, pgAuthIdDescription);
+		commands = lappend(commands, (void *) alterRoleQuery);
+	}
+
+	heap_endscan(scan);
+	heap_close(pgAuthId, AccessShareLock);
+
+	return commands;
+}
+
+
+/*
+ * makeDefElemInt creates a DefElem with integer typed value with -1 as location.
+ */
+static DefElem *
+makeDefElemInt(char *name, int value)
+{
+	return makeDefElem(name, (Node *) makeInteger(value), -1);
+}
--- a/src/backend/distributed/deparser/deparse_role_stmts.c
+++ b/src/backend/distributed/deparser/deparse_role_stmts.c
@@ -1,0 +1,139 @@
+/*-------------------------------------------------------------------------
+ *
+ * deparse_role_stmts.c
+ *	  All routines to deparse role statements.
+ *	  This file contains all entry points specific for ALTER ROLE statement
+ *    deparsing as well as functions that are currently only used for deparsing
+ *    ALTER ROLE statements.
+ *
+ * Copyright (c), Citus Data, Inc.
+ *
+ *-------------------------------------------------------------------------
+ */
+
+#include "postgres.h"
+
+#include "distributed/citus_ruleutils.h"
+#include "distributed/deparser.h"
+#include "lib/stringinfo.h"
+#include "nodes/parsenodes.h"
+#include "utils/builtins.h"
+
+static void AppendAlterRoleStmt(StringInfo buf, AlterRoleStmt *stmt);
+
+
+/*
+ * DeparseAlterRoleStmt builds and returns a string representing of the
+ * AlterRoleStmt for application on a remote server.
+ */
+const char *
+DeparseAlterRoleStmt(AlterRoleStmt *stmt)
+{
+	StringInfoData buf = { 0 };
+	initStringInfo(&buf);
+
+	AppendAlterRoleStmt(&buf, stmt);
+
+	return buf.data;
+}
+
+
+/*
+ * AppendAlterRoleStmt generates the string representation of the
+ * AlterRoleStmt and appends it to the buffer.
+ */
+static void
+AppendAlterRoleStmt(StringInfo buf, AlterRoleStmt *stmt)
+{
+	ListCell *optionCell = NULL;
+	RoleSpec *role = stmt->role;
+
+	/*
+	 * If the role_specification used is CURRENT_USER or SESSION_USER,
+	 * it will be converted to thats roles role name.
+	 */
+	appendStringInfo(buf, "ALTER ROLE %s", RoleSpecString(role, true));
+
+	foreach(optionCell, stmt->options)
+	{
+		DefElem *option = (DefElem *) lfirst(optionCell);
+
+		if (strcmp(option->defname, "superuser") == 0 && intVal(option->arg))
+		{
+			appendStringInfo(buf, " SUPERUSER");
+		}
+		else if (strcmp(option->defname, "superuser") == 0 && !intVal(option->arg))
+		{
+			appendStringInfo(buf, " NOSUPERUSER");
+		}
+		else if (strcmp(option->defname, "createdb") == 0 && intVal(option->arg))
+		{
+			appendStringInfo(buf, " CREATEDB");
+		}
+		else if (strcmp(option->defname, "createdb") == 0 && !intVal(option->arg))
+		{
+			appendStringInfo(buf, " NOCREATEDB");
+		}
+		else if (strcmp(option->defname, "createrole") == 0 && intVal(option->arg))
+		{
+			appendStringInfo(buf, " CREATEROLE");
+		}
+		else if (strcmp(option->defname, "createrole") == 0 && !intVal(option->arg))
+		{
+			appendStringInfo(buf, " NOCREATEROLE");
+		}
+		else if (strcmp(option->defname, "inherit") == 0 && intVal(option->arg))
+		{
+			appendStringInfo(buf, " INHERIT");
+		}
+		else if (strcmp(option->defname, "inherit") == 0 && !intVal(option->arg))
+		{
+			appendStringInfo(buf, " NOINHERIT");
+		}
+		else if (strcmp(option->defname, "canlogin") == 0 && intVal(option->arg))
+		{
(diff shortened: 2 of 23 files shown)


src/backend/distributed/commands/role.c
```diff
@@ -0,0 +1,216 @@
...
+
+
+const char *
+prepareAlterRoleQueryWithUDF(AlterRoleStmt *stmt)
+{
+	StringInfoData alterRoleQueryBuffer = { 0 };
+
+	initStringInfo(&alterRoleQueryBuffer);
+	appendStringInfo(&alterRoleQueryBuffer,
+					 "SELECT run_utility_query_if_role_exists(%s, %s)",
```
onderkalaci: `run_utility_query_if_role_exists ` -> `alter_role_if_exists`

src/backend/distributed/commands/role.c
```diff
@@ -0,0 +1,216 @@
...
+	if (!IsCoordinator())
+	{
+		return NIL;
+	}
+
+	foreach(optionCell, stmt->options)
+	{
+		DefElem *option = (DefElem *) lfirst(optionCell);
+
+		if (EnableAlterRolePropagation && strcasecmp(option->defname, "password") == 0)
```
onderkalaci: we should now pull `EnableAlterRolePropagation ` to above, before the loop starts

src/backend/distributed/commands/role.c
```diff
@@ -0,0 +1,221 @@
...
+	if (list_length(commands) <= 0)
+	{
+		return NIL;
+	}
+	return NodeDDLTaskList(ALL_WORKERS, commands);
+}
+
+
+const char *
+prepareAlterRoleQueryWithUDF(AlterRoleStmt *stmt)
```
onderkalaci: Convention note: (a) Add function comment (b) Start capital letter
 
`prepareAlterRoleQueryWithUDF `  -> `CreateAlterRoleIfExistsCommand()`

src/backend/distributed/commands/role.c
```diff
@@ -0,0 +1,221 @@
...
+{
+	StringInfoData alterRoleQueryBuffer = { 0 };
+
+	initStringInfo(&alterRoleQueryBuffer);
+	appendStringInfo(&alterRoleQueryBuffer,
+					 "SELECT run_utility_query_if_role_exists(%s, %s)",
+					 quote_literal_cstr(stmt->role->rolename),
+					 quote_literal_cstr(DeparseTreeNode((Node *) stmt)));
```
onderkalaci: Minor: assigning this to  `char *alterRoleStmtString = DeparseTreeNode((Node *) stmt))`

src/backend/distributed/commands/role.c
```diff
@@ -0,0 +1,221 @@
...
+
+	return pstrdup(alterRoleQueryBuffer.data);
+}
+
+
+void
```
onderkalaci: Missing comment.

Rename function  to  `ExtractEncryptedPassword()`. Update option on the caller  side.

src/backend/distributed/commands/role.c
```diff
@@ -0,0 +1,221 @@
...
+	}
+	else
+	{
+		((Value *) (option->arg))->val.str = TextDatumGetCString(passwordDatum);
+	}
+}
+
+
+static const char *
+CreateAlterRoleQueryForCopyingUser(HeapTuple tuple)
```
onderkalaci: -`GenerateAlterRoleIfExistsCommand()`
-  Pass description, do not re-open heap
- Missing comment

src/backend/distributed/commands/role.c
```diff
@@ -0,0 +1,221 @@
...
+	return prepareAlterRoleQueryWithUDF(stmt);
+}
+
+
+/*
+ * CreateAlterRoleQueriesForCopyingAllUsers creates the set password queries
+ * for users whose password exists in the pg_authid table.
+ */
+List *
+CreateAlterRoleQueriesForCopyingAllUsers()
```
onderkalaci: `User` -> `Role`
Reflect the function name change of the above as well

src/backend/distributed/commands/role.c
```diff
@@ -0,0 +1,221 @@
...
+
+#if PG_VERSION_NUM >= 120000
+	scan = table_beginscan_catalog(pgAuthId, 0, NULL);
+#else
+	scan = heap_beginscan_catalog(pgAuthId, 0, NULL);
+#endif
+
+	while ((tuple = heap_getnext(scan, ForwardScanDirection)) != NULL)
+	{
+		if (IsReservedName(NameStr((Form_pg_authid) GETSTRUCT(tuple))->rolname))
```
onderkalaci: `NameStr((Form_pg_authid) GETSTRUCT(tuple))->rolname)` -> move to a  variable

src/backend/distributed/utils/node_metadata.c
```diff
@@ -364,6 +365,22 @@ SetUpDistributedTableDependencies(WorkerNode *newWorkerNode)
 }
 
 
+/*
+ * SetPasswords sets the password of users with passwords in the new node.
+ */
+static void
+SetPasswords(WorkerNode *newWorkerNode)
```
onderkalaci: `AlterRolePasswordsIfExists()`

src/backend/distributed/commands/role.c
```diff
@@ -0,0 +1,221 @@
...
+	HeapTuple tuple = NULL;
+	List *commands = NIL;
+	const char *alterRoleQuery = NULL;
+
+	pgAuthId = heap_open(AuthIdRelationId, AccessShareLock);
+
+#if PG_VERSION_NUM >= 120000
+	scan = table_beginscan_catalog(pgAuthId, 0, NULL);
+#else
+	scan = heap_beginscan_catalog(pgAuthId, 0, NULL);
```
marcocitus: maybe put the declaration here to avoid one of the ifdefs.

src/backend/distributed/commands/role.c
halilozanakgul: Done 👍

src/backend/distributed/commands/role.c
halilozanakgul: Done 👍

src/backend/distributed/commands/role.c
halilozanakgul: Done 👍

src/backend/distributed/commands/role.c
halilozanakgul: Done 👍

src/backend/distributed/utils/node_metadata.c
halilozanakgul: Done 👍

src/backend/distributed/commands/role.c
```diff
@@ -0,0 +1,216 @@
...
+	const char *alterRoleQuery = NULL;
+
+#if PG_VERSION_NUM >= 120000
+	TableScanDesc scan = table_beginscan_catalog(pgAuthId, 0, NULL);
+#else
+	HeapScanDesc scan = heap_beginscan_catalog(pgAuthId, 0, NULL);
+#endif
+
+	while ((tuple = heap_getnext(scan, ForwardScanDirection)) != NULL)
+	{
```
onderkalaci: minor: assingning `NameStr(((Form_pg_authid) GETSTRUCT(tuple))->rolname` to a  local  variable  might b e useful

src/backend/distributed/commands/role.c
```diff
@@ -0,0 +1,231 @@
...
+		const char *rolename = NameStr(((Form_pg_authid) GETSTRUCT(tuple))->rolname);
+		if (IsReservedName(rolename) ||
```
onderkalaci: might be useful to have a comment on why we're skipping these roles. 

What about citus user? I'll think a bit on this once I see the comment/reasoning behind this check

src/backend/distributed/commands/role.c
```diff
@@ -0,0 +1,231 @@
+#include "postgres.h"
```
onderkalaci: Minor: we typically have `file comments` at the start of the files with some very basic explanations of the files. Can you check some examples in the source code? And also apply it to all newly added files in this PR?

src/backend/distributed/deparser/deparse_role_stmts.c
```diff
@@ -0,0 +1,139 @@
+/*
+ *
```
onderkalaci: as I noted earlier, missing file comment.

src/backend/distributed/deparser/deparse_role_stmts.c
```diff
@@ -0,0 +1,139 @@
...
+
+#include "lib/stringinfo.h"
+#include "nodes/parsenodes.h"
+#include "utils/builtins.h"
+#include "distributed/citus_ruleutils.h"
+
+static void AppendAlterRoleStmt(StringInfo buf, AlterRoleStmt *stmt);
+const char * DeparseAlterRoleStmt(AlterRoleStmt *stmt);
+
+const char *
```
onderkalaci: missing function comments in the file.

src/backend/distributed/deparser/deparse_role_stmts.c
```diff
@@ -0,0 +1,139 @@
...
+
+#include "postgres.h"
+
+#include "lib/stringinfo.h"
+#include "nodes/parsenodes.h"
+#include "utils/builtins.h"
+#include "distributed/citus_ruleutils.h"
+
+static void AppendAlterRoleStmt(StringInfo buf, AlterRoleStmt *stmt);
+const char * DeparseAlterRoleStmt(AlterRoleStmt *stmt);
```
onderkalaci: As I noted in other place, we typically do not declare external/public functions here as well.

src/backend/distributed/deparser/deparse_role_stmts.c
```diff
@@ -0,0 +1,139 @@
...
+			appendStringInfo(buf, "CURRENT_USER");
+			break;
+		}
+
+		case ROLESPEC_SESSION_USER:
+		{
+			appendStringInfo(buf, "SESSION_USER");
+			break;
+		}
+
```
onderkalaci: Minor: Typically we add `{}` for default case as well
```C
		default:
		{
			appendStringInfo(buf, "%s", RoleSpecString(role));
			break;
		}
```

src/backend/distributed/deparser/deparse_role_stmts.c
```diff
@@ -0,0 +1,139 @@
...
+	foreach(optionCell, stmt->options)
+	{
+		DefElem *option = (DefElem *) lfirst(optionCell);
+
+		if (strcmp(option->defname, "superuser") == 0 && intVal(option->arg))
+		{
+			appendStringInfo(buf, " SUPERUSER");
+		}
+		if (strcmp(option->defname, "superuser") == 0 && !intVal(option->arg))
```
onderkalaci: should these be `else if` instead of `if`?

src/backend/distributed/utils/node_metadata.c
```diff
@@ -364,6 +366,22 @@ SetUpDistributedTableDependencies(WorkerNode *newWorkerNode)
 }
 
 
+/*
+ * PropagateRolesToNewNode sets the password of users with passwords in the new node.
```
onderkalaci: I think we're not only doing passwords anymore, so the comment is outdated?

src/backend/distributed/utils/role.c
```diff
@@ -0,0 +1,48 @@
+/*-------------------------------------------------------------------------
+ *
+ * role.c
+ *
+ * Utilities for role passwords.
```
onderkalaci: out-dated comment

src/backend/distributed/deparser/deparse_role_stmts.c
```diff
@@ -0,0 +1,139 @@
...
+{
+	ListCell *optionCell = NULL;
+	RoleSpec *role = stmt->role;
+
+	appendStringInfo(buf, "ALTER ROLE ");
+	switch (role->roletype)
+	{
+		case ROLESPEC_CURRENT_USER:
+		{
+			appendStringInfo(buf, "CURRENT_USER");
```
onderkalaci: Well,  because Citus using a separate user while sending these commands to the remote nodes, writing `CURRENT_USER `  or `SESSION_USER ` wouldn't be correct. I think `RoleSpecString`  already resolves the names to actual role names. Can't we simply use that for all?

halilozanakgul: I feel like this would not be the expected behaviour of a deparser. However, I still changed it and added a comment explaining this.

onderkalaci: Well, that's true. thanks for adding the comments.

src/backend/distributed/commands/role.c
```diff
@@ -0,0 +1,258 @@
...
+	Relation pgAuthId = heap_open(AuthIdRelationId, AccessShareLock);
+	TupleDesc pgAuthIdDescription = RelationGetDescr(pgAuthId);
+	HeapTuple tuple = SearchSysCache1(AUTHOID, roleOid);
+	bool isNull = true;
+	Datum passwordDatum = heap_getattr(tuple, Anum_pg_authid_rolpassword,
+									   pgAuthIdDescription, &isNull);
+
+	heap_close(pgAuthId, AccessShareLock);
+	ReleaseSysCache(tuple);
+
```
onderkalaci: we should return NULL when `isNull` is set, because `TextDatumGetCString ` doesn't expect a NULL datum, I actually observed a crash because of this

src/backend/distributed/commands/role.c
```diff
@@ -0,0 +1,258 @@
...
+/*
+ * ExtractEncryptedPassword extracts the encrypted password of a role. The function
+ * gets the password from the pg_authid table.
+ */
+static const char *
+ExtractEncryptedPassword(Oid roleOid)
+{
+	Relation pgAuthId = heap_open(AuthIdRelationId, AccessShareLock);
+	TupleDesc pgAuthIdDescription = RelationGetDescr(pgAuthId);
+	HeapTuple tuple = SearchSysCache1(AUTHOID, roleOid);
```
onderkalaci: We can happily pass InvalidOid here as the roleId. So, we should be careful, and return NULL
```C
if (!HeapTupleIsValid(tuple))
{
		retun NULL
}

File: src/backend/distributed/commands/role.c [hunk 24]
```diff
@@ -0,0 +1,258 @@
...
+	const char *rolename = NameStr(role->rolname);
+
+	stmt->role = makeNode(RoleSpec);
+	stmt->role->roletype = ROLESPEC_CSTRING;
+	stmt->role->location = -1;
+	stmt->role->rolename = pstrdup(rolename);
+	stmt->action = 1;
+	stmt->options = NIL;
+
+	stmt->options =
```
onderkalaci: I know this is annoying, but  the indentation is still not very convenient here :/ Can you check this: https://github.com/citusdata/citus/compare/alter_role_propagation...minor_improvement?expand=1

src/backend/distributed/commands/role.c
```diff
@@ -0,0 +1,258 @@
...
+ProcessAlterRoleStmt(AlterRoleStmt *stmt, const char *queryString)
+{
+	ListCell *optionCell = NULL;
+	List *commands = NIL;
+
+	if (!EnableAlterRolePropagation || !IsCoordinator())
+	{
+		return NIL;
+	}
+
```
onderkalaci: To prevent concurrent node addition, we should have something like this here:
```C
        /* don't allow concurrent node list changes that require an exclusive lock */
        LockRelationOid(DistNodeRelationId(), RowShareLock);
```

Otherwise, in one session, you can add a node, and on the other you can ALTER ROLE, and that ALTER ROLE might not get propagated to the newly added node.

src/backend/distributed/sql/udfs/alter_role_if_exists/9.1-1.sql
```diff
@@ -0,0 +1,12 @@
+CREATE OR REPLACE FUNCTION alter_role_if_exists(
```
marcocitus: Should include pg_catalog in the name

src/backend/distributed/commands/role.c
```diff
@@ -0,0 +1,271 @@
...
+		DefElem *option = (DefElem *) lfirst(optionCell);
+
+		if (strcasecmp(option->defname, "password") == 0)
+		{
+			Oid roleOid = get_rolespec_oid(stmt->role, true);
+			const char *encryptedPassword = ExtractEncryptedPassword(roleOid);
+
+			if (encryptedPassword != NULL)
+			{
+				Value *encriptedPasswordValue = makeString((char *) encryptedPassword);
```
marcocitus: encripted -> encrypted

src/backend/distributed/commands/role.c
```diff
@@ -0,0 +1,278 @@
...
+
+	stmt->options =
+		lappend(stmt->options,
+				makeDefElem("superuser",
+							(Node *) makeInteger(role->rolsuper), -1));
+
+	stmt->options =
+		lappend(stmt->options,
+				makeDefElem("createdb",
+							(Node *) makeInteger(role->rolcreatedb), -1));
```
marcocitus: Could make a makeDefElemInt utility function to shorten this a bit.

src/backend/distributed/commands/type.c
```diff
@@ -82,7 +82,7 @@
 
 
 /* guc to turn of the automatic type distribution */
-bool EnableCreateTypePropagation = true;
+bool EnableCreateTypePropagation = false;
```
marcocitus: why change this?

serprex:
Is there a good reason to not bring the `CREATE ROLE` logic from enterprise to community? It feels with the object propagation development that this will become a nuisance. It already costs us in terms of regression test maintenance

src/backend/distributed/deparser/deparse_role_stmts.c
```diff
@@ -0,0 +1,139 @@
...
+	 * If the role_specification used is CURRENT_USER or SESSION_USER,
+	 * it will be converted to thats roles role name.
+	 */
+	appendStringInfo(buf, "ALTER ROLE %s", RoleSpecString(role));
+
+	foreach(optionCell, stmt->options)
+	{
+		DefElem *option = (DefElem *) lfirst(optionCell);
+
+		if (strcmp(option->defname, "superuser") == 0 && intVal(option->arg))
```
SaitTalhaNisanci: I think here you can create a char array and use that instead of having many if elses.

```
char* defNames[7] = {"superuser", "createdb", "createrole", "inherit", "login", "replication", "bypassrls"}
```

And you can iterate the char array to see if `option->defname` is equal to the current element, if so it means that you will append the current element if `intVal(option->arg)` returns true, or you will prepend `NO` if it returns false.

marcocitus:
There is no CREATE ROLE logic in enterprise, only in cloud environments.

src/test/regress/sql/propagate_alter_role.sql
```diff
@@ -0,0 +1,86 @@
...
+SELECT row(rolname, rolsuper, rolinherit,  rolcreaterole, rolcreatedb, rolcanlogin, rolreplication, rolbypassrls, rolconnlimit, rolpassword, rolvaliduntil) FROM pg_authid WHERE rolname = 'alter_role_1';
+SELECT run_command_on_workers($$SELECT row(rolname, rolsuper, rolinherit,  rolcreaterole, rolcreatedb, rolcanlogin, rolreplication, rolbypassrls, rolconnlimit, rolpassword, rolvaliduntil) FROM pg_authid WHERE rolname = 'alter_role_1'$$);
+
+-- make sure that we propagate all options accurately
+ALTER ROLE alter_role_1 WITH NOSUPERUSER NOCREATEDB NOCREATEROLE NOINHERIT NOLOGIN NOREPLICATION NOBYPASSRLS CONNECTION LIMIT 0 VALID UNTIL '2052-01-01 00:00:00+00'; 
+SELECT row(rolname, rolsuper, rolinherit,  rolcreaterole, rolcreatedb, rolcanlogin, rolreplication, rolbypassrls, rolconnlimit, rolpassword, rolvaliduntil) FROM pg_authid WHERE rolname = 'alter_role_1';
+SELECT run_command_on_workers($$SELECT row(rolname, rolsuper, rolinherit,  rolcreaterole, rolcreatedb, rolcanlogin, rolreplication, rolbypassrls, rolconnlimit, rolpassword, rolvaliduntil) FROM pg_authid WHERE rolname = 'alter_role_1'$$);
```
onderkalaci: Please update all the occurrences of `rolvaliduntil` in the SQL test files -- including isolation tests -- with `left(rolvaliduntil::text, 10)`  to ignore the hour,  only get the `YYY-MM-DD`.
Otherwise, the tests are dependent on the timezone

src/test/regress/isolation_schedule
```diff
@@ -6,6 +6,7 @@ test: isolation_create_table_vs_add_remove_node
 test: isolation_master_update_node
 test: isolation_ensure_dependency_activate_node
 test: isolation_shouldhaveshards
+test: isolation_alter_role_propagation
```
onderkalaci: can you move this after `isolation_dump_global_wait_edges ` because  that test's output is changing -- shame on that test being dependent on the previous tests

src/backend/distributed/commands/type.c
onderkalaci: I guess during testing he changed and  forgot? @halilozanakgul let's not forget to undo this

src/backend/distributed/deparser/deparse_role_stmts.c
onderkalaci: > if so it means that you will append the current element

note that few of elements have different defName  than the actual sql string like   `canlogin `   > `LOGIN `    or `isreplication ` -> `REPLICATION`. 

So, that approach might still  require some if statements,  I'd rather keep this, which is long but at least very straightforward

src/test/regress/expected/isolation_alter_role_propagation.out
```diff
@@ -0,0 +1,108 @@
...
+step s1-begin: 
+    SET citus.enable_alter_role_propagation to ON;
+    SELECT run_command_on_workers($$SET citus.enable_alter_role_propagation to ON;$$);
+    BEGIN;
+
+run_command_on_workers
+
+(localhost,57637,t,SET)
+step s1-add-node: 
+    SELECT 1 FROM master_add_node('localhost', 57637);
```
onderkalaci: This node already exists in the cluster, so `master_add_node ` does nothing -- no-op. 

To make this test work,  we should remove the node on `setup`, and add it like this

src/test/regress/specs/isolation_alter_role_propagation.spec
```diff
@@ -0,0 +1,70 @@
...
+    SET citus.enable_alter_role_propagation to OFF;
+    SELECT run_command_on_workers($$SET citus.enable_alter_role_propagation to OFF;$$);
+}
+
+session "s1"
+
+step "s1-begin"
+{
+    SET citus.enable_alter_role_propagation to ON;
+    SELECT run_command_on_workers($$SET citus.enable_alter_role_propagation to ON;$$);
```
onderkalaci: we don't need to set the GUC on the workers. actually, that makes it confusing,  can you remove that?

src/backend/distributed/commands/role.c
```diff
@@ -0,0 +1,278 @@
...
+
+static const char * ExtractEncryptedPassword(Oid roleOid);
+static const char * CreateAlterRoleIfExistsCommand(AlterRoleStmt *stmt);
+
+/* controlled via GUC */
+bool EnableAlterRolePropagation = false;
+
+/*
+ * ProcessAlterRoleStmt actually creates the plan we need to execute for alter
+ * role statement.
```
onderkalaci: Typically, the `Plan...` corresponding of the statement creates the tasks to execute. However, for some specific cases -- like this -- we do it on `Process...` part.  So, maybe add the reason for that to the comment (which is because we need to read the encrypted password from pg_authid, and that is true after `standardProcessUtility` is called)

src/backend/distributed/utils/node_metadata.c
```diff
@@ -388,6 +390,27 @@ SetUpDistributedTableDependencies(WorkerNode *newWorkerNode)
 }
 
 
+/*
+ * PropagateRolesToNewNode copies the roles' attributes in the new node.
```
onderkalaci: maybe add that the roles that do  not exist skipped

src/test/regress/expected/propagate_alter_role.out
```diff
@@ -0,0 +1,227 @@
+CREATE SCHEMA alter_role;
+SET citus.enable_alter_role_propagation to ON;
+SELECT run_command_on_workers($$SET citus.enable_alter_role_propagation to ON;$$);
+ run_command_on_workers  
+-------------------------
+ (localhost,57637,t,SET)
+ (localhost,57638,t,SET)
+(2 rows)
+
+CREATE ROLE alter_role_1 WITH LOGIN;
```
onderkalaci: it would be nice to use a role name that has some escaped characters like `"alter_role'1"` or such to test the edge cases

src/test/regress/expected/propagate_alter_role.out
```diff
@@ -0,0 +1,227 @@
+CREATE SCHEMA alter_role;
+SET citus.enable_alter_role_propagation to ON;
+SELECT run_command_on_workers($$SET citus.enable_alter_role_propagation to ON;$$);
```
onderkalaci: why do we enable on the workers?

src/test/regress/expected/propagate_alter_role.out
```diff
@@ -0,0 +1,227 @@
...
+(1 row)
+
+SELECT run_command_on_workers($$SELECT row(rolname, rolsuper, rolinherit,  rolcreaterole, rolcreatedb, rolcanlogin, rolreplication, rolbypassrls, rolconnlimit, rolpassword, rolvaliduntil) FROM pg_authid WHERE rolname = 'alter_role_1'$$);
+                                               run_command_on_workers                                                
+---------------------------------------------------------------------------------------------------------------------
+ (localhost,57637,t,"(alter_role_1,f,f,f,f,f,f,f,0,md5be308f25c7b1a2d50c85cf7e6f074df9,""2052-01-01 03:00:00+03"")")
+ (localhost,57638,t,"(alter_role_1,f,f,f,f,f,f,f,0,md5be308f25c7b1a2d50c85cf7e6f074df9,""2052-01-01 03:00:00+03"")")
+(2 rows)
+
+-- TODO: we should not propagate scram-sha-256 passwords
```
onderkalaci: let's remove the TODO, we agreed to support scram-sha-256

src/test/regress/sql/propagate_alter_role.sql
onderkalaci: Or, maybe  just use ` EXTRACT (year FROM rolvaliduntil)`, and set to different `years` in the tests. So,  seeing the correct year is  enough

src/backend/distributed/deparser/deparse_role_stmts.c
SaitTalhaNisanci: Is there a reason why they are different in the first place? I think we can make them the same right? Also we have them hardcoded in two different places, I think ideally we should use a constant variable/define that holds the hard coded strings, so that we use them only from one place and they are not hard coded.

onderkalaci: that's the output of Postgres parser, we don't have a control on that

src/backend/distributed/sql/udfs/alter_role_if_exists/latest.sql
```diff
@@ -0,0 +1,12 @@
+CREATE OR REPLACE FUNCTION pg_catalog.alter_role_if_exists(
+    role_name text,
+    utility_query text)
+    RETURNS void
```
onderkalaci: the function actually returns bool, and it might be useful. So can you update here to return bool as well?

src/backend/distributed/deparser/deparse_role_stmts.c
SaitTalhaNisanci: cant we choose the same output as Postgres?

onderkalaci: all the strings used in this logic directly related to postgres,

option->defname- > Postgres parser gives us
appendStringInfo(buf, XXX) -> Postgres parser expects us

So, that doesn't seem feasible

SaitTalhaNisanci: I see, thanks for the explanation :) 

I think we should at least use a variable instead of having the same hard coded variables in two different places.

src/backend/distributed/commands/role.c
```diff
@@ -0,0 +1,282 @@
...
+	while ((tuple = heap_getnext(scan, ForwardScanDirection)) != NULL)
+	{
+		const char *rolename = NameStr(((Form_pg_authid) GETSTRUCT(tuple))->rolname);
+
+		/*
+		 * The default roles and "postgres" is skipped, because reserved roles
+		 * cannot be altered.
+		 */
+		if (IsReservedName(rolename) ||
+			strcmp(rolename, "postgres") == 0)
```
onderkalaci: I'm still not convinced that we need anything special for `postgres` user. All the tests pass even if I remove that check. So, what's the reason we added? If you don't have a good reason, can you remove the check?

PR #3185, opened 2019-11-13
Add extensions to distributed object propagation infrastructure

onurctirtir:
DESCRIPTION:  Add extensions to distributed object propagation infrastructure

### DIFF
--- a/src/backend/distributed/commands/extension.c
+++ b/src/backend/distributed/commands/extension.c
@@ -11,35 +11,32 @@
 #include "postgres.h"
 
 #include "citus_version.h"
+#include "catalog/pg_extension_d.h"
+#include "commands/extension.h"
+#include "distributed/citus_ruleutils.h"
 #include "distributed/commands.h"
-#include "distributed/metadata_cache.h"
-#include "nodes/parsenodes.h"
+#include "distributed/commands/utility_hook.h"
+#include "distributed/deparser.h"
+#include "distributed/metadata_sync.h"
+#include "distributed/metadata/distobject.h"
+#include "distributed/multi_executor.h"
+#include "distributed/relation_access_tracking.h"
+#include "distributed/transaction_management.h"
+#include "nodes/makefuncs.h"
+#include "utils/lsyscache.h"
+#include "utils/builtins.h"
+
 
 /* Local functions forward declarations for helper functions */
-static char * ExtractNewExtensionVersion(Node *parsetree);
-
-
-/*
- * IsCitusExtensionStmt returns whether a given utility is a CREATE or ALTER
- * EXTENSION statement which references the citus extension. This function
- * returns false for all other inputs.
- */
-bool
-IsCitusExtensionStmt(Node *parsetree)
-{
-	char *extensionName = "";
-
-	if (IsA(parsetree, CreateExtensionStmt))
-	{
-		extensionName = ((CreateExtensionStmt *) parsetree)->extname;
-	}
-	else if (IsA(parsetree, AlterExtensionStmt))
-	{
-		extensionName = ((AlterExtensionStmt *) parsetree)->extname;
-	}
-
-	return (strcmp(extensionName, "citus") == 0);
-}
+static char * ExtractNewExtensionVersion(Node *parseTree);
+static void AddSchemaFieldIfMissing(CreateExtensionStmt *stmt);
+static List * FilterDistributedExtensions(List *extensionObjectList);
+static List * ExtensionNameListToObjectAddressList(List *extensionObjectList);
+static void EnsureSequentialModeForExtensionDDL(void);
+static bool ShouldPropagateExtensionCommand(Node *parseTree);
+static bool IsDropCitusStmt(Node *parseTree);
+static bool IsAlterExtensionSetSchemaCitus(Node *parseTree);
+static Node * RecreateExtensionStmt(Oid extensionOid);
 
 
 /*
@@ -49,9 +46,9 @@ IsCitusExtensionStmt(Node *parsetree)
  * out. It ignores the schema version.
  */
 void
-ErrorIfUnstableCreateOrAlterExtensionStmt(Node *parsetree)
+ErrorIfUnstableCreateOrAlterExtensionStmt(Node *parseTree)
 {
-	char *newExtensionVersion = ExtractNewExtensionVersion(parsetree);
+	char *newExtensionVersion = ExtractNewExtensionVersion(parseTree);
 
 	if (newExtensionVersion != NULL)
 	{
@@ -78,24 +75,24 @@ ErrorIfUnstableCreateOrAlterExtensionStmt(Node *parsetree)
 
 
 /*
- * ExtractNewExtensionVersion returns the new extension version specified by
- * a CREATE or ALTER EXTENSION statement. Other inputs are not permitted. This
- * function returns NULL for statements with no explicit version specified.
+ * ExtractNewExtensionVersion returns the palloc'd new extension version specified
+ * by a CREATE or ALTER EXTENSION statement. Other inputs are not permitted.
+ * This function returns NULL for statements with no explicit version specified.
  */
 static char *
-ExtractNewExtensionVersion(Node *parsetree)
+ExtractNewExtensionVersion(Node *parseTree)
 {
-	char *newVersion = NULL;
-	List *optionsList = NIL;
-	ListCell *optionsCell = NULL;
+	Value *newVersionValue = NULL;
 
-	if (IsA(parsetree, CreateExtensionStmt))
+	List *optionsList = NIL;
+
+	if (IsA(parseTree, CreateExtensionStmt))
 	{
-		optionsList = ((CreateExtensionStmt *) parsetree)->options;
+		optionsList = ((CreateExtensionStmt *) parseTree)->options;
 	}
-	else if (IsA(parsetree, AlterExtensionStmt))
+	else if (IsA(parseTree, AlterExtensionStmt))
 	{
-		optionsList = ((AlterExtensionStmt *) parsetree)->options;
+		optionsList = ((AlterExtensionStmt *) parseTree)->options;
 	}
 	else
 	{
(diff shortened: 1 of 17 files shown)


src/backend/distributed/deparser/deparse_extension_stmts.c
```diff
@@ -0,0 +1,127 @@
...
+	List *optionsList = createExtensionStmt->options;
+
+	const char *newVersion = GetCreateExtensionOption(optionsList, "new_version");
+	const char *schemaName = GetCreateExtensionOption(optionsList, "schema");
+
+	newVersion = quote_identifier(newVersion);
+	schemaName = quote_identifier(schemaName);
+
+	appendStringInfo(str,
+					 "CREATE EXTENSION IF NOT EXISTS %s WITH SCHEMA %s VERSION %s CASCADE",
```
marcocitus: Not sure we should include a version, e.g. Citus would throw errors if the version that's installed on the workers is different (not uncommon).

onderkalaci: So, I guess we should only do if the user actually adds the `WITH version`? In that case, it probably still makes sense to send the version. If it fails, user should take action because he/she wants a specific version?

src/backend/distributed/deparser/deparse_extension_stmts.c
```diff
@@ -0,0 +1,127 @@
...
+}
+
+
+/*
+ * AppendCreateExtensionStmt appends a string representing the CreateExtensionStmt to a buffer
+ */
+static void
+AppendCreateExtensionStmt(StringInfo str, CreateExtensionStmt *createExtensionStmt)
+{
+	/*
```
onderkalaci: we typically do not start functions with comments. Can't we move relevant pieces of the comment into the function?

src/backend/distributed/deparser/objectaddress.c
```diff
@@ -257,3 +267,15 @@ AlterObjectDependsStmtObjectAddress(AlterObjectDependsStmt *stmt, bool missing_o
 		}
 	}
 }
+
+
+static const ObjectAddress *
```
onderkalaci: missing comment

src/backend/distributed/deparser/deparse_extension_stmts.c
marcocitus: Yes.

src/backend/distributed/deparser/deparse_extension_stmts.c
onurctirtir: I will work on it

src/backend/distributed/deparser/deparse_extension_stmts.c
onurctirtir: I will make the necessary changes

src/backend/distributed/commands/extension.c
```diff
@@ -115,3 +110,516 @@ ExtractNewExtensionVersion(Node *parsetree)
...
+	if (!ShouldPropagateExtensionCommand((Node *) createExtensionStmt))
+	{
+		return NIL;
+	}
+
+	/* Extension management can only be done via coordinator node */
+	EnsureCoordinator();
+
+	/*
+	 * Here we append "new_version" and "schema" fields to the "options" list
```
onderkalaci: Let's not forget to removing the code for always adding the version and updating the comment. 

In that case, it might be good to rename the function to `AddSchemaFieldIfMissing(createExtensionStmt)` or such

src/backend/distributed/commands/extension.c
```diff
@@ -115,3 +110,516 @@ ExtractNewExtensionVersion(Node *parsetree)
...
+		char *extensionName = createExtensionStmt->extname;
+
+		/*
+		 * Get the latest available version of the extension from
+		 * pg_available_extensions() to ensure version consistency
+		 * between worker nodes and coordinator.
+		 */
+		char *extensionDefaultVersion = GetDefaultExtensionVersion(extensionName);
+
+		Node *extensionVersionArg = (Node *) (makeString(extensionDefaultVersion));
```
onderkalaci: Minor: we don't need the `()`, so use `makeString(extensionDefaultVersion))`

src/backend/distributed/commands/extension.c
```diff
@@ -115,3 +110,516 @@ ExtractNewExtensionVersion(Node *parsetree)
...
+	{
+		char *currentSchemaName = GetCurrentSchema();
+
+		Node *schemaNameArg = (Node *) (makeString(currentSchemaName));
+
+		/* Set location to -1 as it is unknown */
+		int location = -1;
+
+		DefElem *newDefElement = makeDefElem("schema", schemaNameArg, location);
+
```
onderkalaci: Let's add a comment here. Why are we adding `schema` here?

src/backend/distributed/commands/extension.c
```diff
@@ -115,3 +110,516 @@ ExtractNewExtensionVersion(Node *parsetree)
...
+		if (IsA(defElement, DefElem) && strncmp(defElement->defname, defname,
+												NAMEDATALEN) == 0)
+		{
+			targetStr = strVal(defElement->arg);
+			break;
+		}
+	}
+
+	return targetStr;
```
onderkalaci: to be on the safe side: `return pstrdup(targetStr)`

src/backend/distributed/commands/extension.c
```diff
@@ -115,3 +110,516 @@ ExtractNewExtensionVersion(Node *parsetree)
...
+	if (list_length(distributedExtensions) <= 0)
+	{
+		/* no distributed extensions to drop */
+		return NIL;
+	}
+
+	/* Extension management can only be done via coordinator node */
+	EnsureCoordinator();
+
+	distributedExtensionAddresses = ExtensionNameListToObjectAddresses(
```
onderkalaci: `ExtensionNameListToObjectAddresses` -> `ExtensionNameListToObjectAddressList`  and please update the variables/comments related to this

src/backend/distributed/commands/extension.c
```diff
@@ -115,3 +110,516 @@ ExtractNewExtensionVersion(Node *parsetree)
...
+	bool missingOk = true;
+	ListCell *objectCell = NULL;
+
+	foreach(objectCell, objects)
+	{
+		const char *extensionName = strVal(lfirst(objectCell));
+
+		ObjectAddress *address = palloc0(sizeof(ObjectAddress));
+
+		Oid extensionoid = get_extension_oid(extensionName, missingOk);
```
onderkalaci: extensionoid -> extensionOid

src/backend/distributed/commands/extension.c
```diff
@@ -115,3 +110,516 @@ ExtractNewExtensionVersion(Node *parsetree)
...
+/*
+ * Given the "objects" list of a DropStmt, return the object addresses in
+ * an ObjectAddress list.
+ * Callers of this function should ensure that all the objects in the list
+ * are valid and distributed.
+ */
+static List *
+ExtensionNameListToObjectAddresses(List *objects)
+{
+	List *result = NIL;
```
onderkalaci: result -> extensionAddressList

src/backend/distributed/commands/extension.c
```diff
@@ -115,3 +110,516 @@ ExtractNewExtensionVersion(Node *parsetree)
...
+		}
+
+		ObjectAddressSet(*address, ExtensionRelationId, extensionoid);
+
+		if (!IsObjectDistributed(address))
+		{
+			continue;
+		}
+
+		result = lappend(result, makeString(extensionName));
```
onderkalaci: why do we use `makeString` here? Can't we simply add `extensionName` to list.

Because we do it here, we have to use `strVal` later on to get the `char *`, let's simplify by removing both

src/backend/distributed/commands/extension.c
```diff
@@ -115,3 +110,516 @@ ExtractNewExtensionVersion(Node *parsetree)
...
+
+/*
+ * IsCreateAlterCitusStmt returns whether a given utility is a CREATE or ALTER
+ * EXTENSION statement which references the citus extension. This function
+ * returns false for all other inputs.
+ */
+bool
+IsCreateAlterCitusStmt(Node *parsetree)
+{
+	char *extensionName = "";
```
onderkalaci: = `NULL` is more common in the conventions

src/backend/distributed/commands/extension.c
```diff
@@ -115,3 +110,516 @@ ExtractNewExtensionVersion(Node *parsetree)
...
+ * in worker nodes if distributed before.
+ * We first ensure that we keep only the distributed ones before propagating
+ * the statement to worker nodes.
+ * If no extensions in the drop list are distributed, then no calls will
+ * be made to the workers.
+ */
+List *
+PlanDropExtensionStmt(DropStmt *dropStmt, const char *queryString)
+{
+	List *oldObjects = dropStmt->objects;
```
onderkalaci: `oldObjects` doesn't sound very descriptive. Maybe `allDroppedExtensions`

src/backend/distributed/commands/extension.c
```diff
@@ -115,3 +110,516 @@ ExtractNewExtensionVersion(Node *parsetree)
...
+		return NIL;
+	}
+
+	/* Extension management can only be done via coordinator node */
+	EnsureCoordinator();
+
+	distributedExtensionAddresses = ExtensionNameListToObjectAddresses(
+		distributedExtensions);
+
+	/* Unmark each distributed extension */
```
onderkalaci: Super minor: In general, if the comment is a single line comment, we start with lower character. Can you go over the code to fix this minor convention issue

src/backend/distributed/commands/extension.c
```diff
@@ -115,3 +110,516 @@ ExtractNewExtensionVersion(Node *parsetree)
...
+
+
+/*
+ * Given the "objects" list of a DropStmt, return the object addresses in
+ * an ObjectAddress list.
+ * Callers of this function should ensure that all the objects in the list
+ * are valid and distributed.
+ */
+static List *
+ExtensionNameListToObjectAddresses(List *objects)
```
onderkalaci: `objects` -> `extensionObjectList`

src/backend/distributed/deparser/deparse_extension_stmts.c
```diff
@@ -0,0 +1,127 @@
...
+	 * append "CASCADE" and "IF NOT EXISTS" clauses regardless of
+	 * the content of the statement.
+	 * Here we only need to fetch "objects" list that is storing the
+	 * object names to be deleted.
+	 */
+	appendStringInfoString(str, "DROP EXTENSION IF EXISTS ");
+
+	AppendExtensionNameList(str, dropStmt->objects);
+
+	appendStringInfoString(str, " CASCADE;");
```
onderkalaci: Why do we always add `CASCADE` here? I can see that `IF EXISTS` is required, but I couldn't follow the `CASCADE` part

src/test/regress/expected/propagate_extension_commands.out
```diff
@@ -0,0 +1,166 @@
...
+
+	-- now, create a reference table relying on the data types
+	CREATE TABLE ref_table (a public.issn);
+	SELECT create_reference_table('ref_table');
+ create_reference_table 
+------------------------
+ 
+(1 row)
+
+	-- and the extension is created on the workers -- TODO: they return 0 as we are in the transaction block
```
onderkalaci: As noted in the comment, this test is not useful. So, remove this select

src/test/regress/expected/propagate_extension_commands.out
```diff
@@ -0,0 +1,166 @@
...
+ (localhost,57637,t,0)
+ (localhost,57638,t,0)
+(2 rows)
+
+	-- now, update to a newer version
+	ALTER EXTENSION isn UPDATE TO '1.2';
+	-- now change the schema
+	ALTER EXTENSION isn SET SCHEMA "extension'test";
+	-- show that the ALTER EXTENSION commands are propagated
+	SELECT run_command_on_workers($$SELECT extnamespace FROM pg_extension WHERE extname = 'isn'$$);
```
onderkalaci: Similarly, `run_command_on_workers` doesn't work in transaction block.

Let's move this `SELECT run_command_on_workers($$SELECT extnamespace FROM pg_extension WHERE extname = 'isn'$$);` and the following  `run_command_on_workers ` outside of the transaction block. But, instead of `ROLLBACK `, do a commit, otherwise, we'd not see the results again

src/backend/distributed/commands/extension.c
```diff
@@ -115,3 +110,516 @@ ExtractNewExtensionVersion(Node *parsetree)
...
+	const ObjectAddress *extensionAddress = NULL;
+
+	if (!ShouldPropagateExtensionCommand((Node *) createExtensionStmt))
+	{
+		return;
+	}
+
+	extensionAddress = GetObjectAddressFromParseTree((Node *) createExtensionStmt, false);
+
+	EnsureDependenciesExistsOnAllNodes(extensionAddress);
```
onderkalaci: We should do something like but for extensions (please look at the source of the function below):
```C
EnsureSequentialModeForFunctionDDL();
```

Otherwise:
```SQL
CREATE EXTENSION isn ;
CREATE TABLE t1 (a int, b issn);
select create_distributed_table('t1', 'a');
...
DETAIL:  on server onderkalaci@localhost:9701
ERROR:  type "public.issn" does not exist
```

src/backend/distributed/commands/extension.c
onurctirtir: I guess I should extend the leading comment of this function

src/backend/distributed/commands/extension.c
onurctirtir: Actually, I aimed to implement `ExtensionNameListToObjectAddresses` `(ExtensionNameListToObjectAddressList)` function to be a generic one so it takes an `object` list of a `DropStmt`. This is why I am returning a list of `makeString(char*)` (which returns `Value`) in `FilterDistributedExtensions`.

Does this approach seem to be a valid one ?

src/backend/distributed/deparser/deparse_extension_stmts.c
onurctirtir: Actually, I aimed to automatically drop the objects that depend on the extension in worker nodes. I cannot predict what kind of objects can depend on an extension though.

Should I change this logic ?

onurctirtir:
I applied many changes in 2ec66bba48c7b8de44cfe7b5833e5a9b848c6ff4 for most items mentioned in above comments (that I reacted with thumbs up). I will work on the other review items soon

src/backend/distributed/commands/extension.c
onurctirtir: Implemented EnsureSequentialModeForExtensionDDL  and utilized it in plan extension command functions in 359526463cf6a8c3559297b20839aba83800b2e7. Tests will be added soon

onurctirtir:
Extended master_add_node functionality for extension recreation when a new node is added in 3f3e295e01243b992210207d2981e90c49e1c775. Tests for this will be added soon

src/backend/distributed/commands/extension.c
```diff
@@ -113,5 +109,523 @@ ExtractNewExtensionVersion(Node *parsetree)
...
+ProcessCreateExtensionStmt(CreateExtensionStmt *createExtensionStmt, const
+						   char *queryString)
+{
+	const ObjectAddress *extensionAddress = NULL;
+
+	if (!ShouldPropagateExtensionCommand((Node *) createExtensionStmt))
+	{
+		return;
+	}
+
```
onderkalaci: To prevent concurrent node addition, we should have something like this here:
```C
        /* don't allow concurrent node list changes that require an exclusive lock */
        LockRelationOid(DistNodeRelationId(), RowShareLock);
```

Otherwise, in one session, you can add a node, and on the other you can add an extension, and that the extension might not get propagated to the newly added node.


We should verify this  with an  isolation test

src/test/regress/expected/propagate_extension_commands.out
```diff
@@ -0,0 +1,166 @@
...
+ (localhost,57638,t,0)
+(2 rows)
+
+-- try setting the schema outside of the transaction block, TODO: alters in coordinator but not in workers as it is not implemented yed
+ALTER EXTENSION hstore SET SCHEMA "public";
+-- TODO: give a notice for the following commands saying that it is not
+-- propagated to the workers. the user should run it manually on the workers 
+-- ALTER EXTENSION name ADD member_object
+-- ALTER EXTENSION name DROP member_object
+DROP EXTENSION hstore CASCADE;
```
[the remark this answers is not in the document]
onurctirtir: changes can be seen at 7e956de0c46dc5f986da224d7ff6d3d25ad069c3

src/test/regress/expected/propagate_extension_commands.out
onurctirtir: changes can be seen at 7e956de0c46dc5f986da224d7ff6d3d25ad069c3

src/backend/distributed/commands/extension.c
onurctirtir: tests for this are added in ae7bfdad2076bf2f55c150d9af7a830a13e8e18c but ineffective because of some other problems

onurctirtir:
I guess, changes in 3f3e295e01243b992210207d2981e90c49e1c775 were not sufficient for `1) Include the changes required for master_add_node to work`. I will work on it

onurctirtir:
I think I fixed the problem with master_add_node at 26347c8b80c86d9d7076a550460aa2ec66959d80

src/backend/distributed/commands/extension.c
onurctirtir: Implemented in bddad5780f18cfc01cba610e3f0c78175372cb93, I will add related isolation tests soon

src/test/regress/expected/propagate_extension_commands.out
```diff
@@ -0,0 +1,285 @@
...
+CREATE TABLE ref_table_2 (x hstore);
+SELECT create_reference_table('ref_table_2');
+ create_reference_table 
+------------------------
+ 
+(1 row)
+
+-- and add the other node
+SELECT master_add_node('localhost', 57638);
+NOTICE:  Replicating reference table "ref_table_2" to the node localhost:57638
```
onderkalaci: We need to make sure that this block works in a transaction block as well:
```SQL
BEGIN;
CREATE EXTENSION hstore;
CREATE TYPE test_type AS (a int, b hstore); 
CREATE TABLE t (a int, b test_type );
SELECT create_distributed_table('t', 'a');
SELECT master_add_node(...);
COMMIT;
```

which unfortunately fails, I'll look a bit more

src/backend/distributed/commands/extension.c
```diff
@@ -113,5 +110,661 @@ ExtractNewExtensionVersion(Node *parsetree)
...
+	commands = list_make3(DISABLE_DDL_PROPAGATION,
+						  (void *) createExtensionStmtSql,
+						  ENABLE_DDL_PROPAGATION);
+
+	return NodeDDLTaskList(ALL_WORKERS, commands);
+}
+
+
+/*
+ * Add DefElem item for "schema" (if not specified in statement) to "options"
```
onderkalaci: Minor: We typically start the function comment with the functions name  `AddSchemaFieldIfMissing  adds ...`

(please apply the  same to all functions)

src/backend/distributed/commands/schema.c
```diff
@@ -128,6 +128,11 @@ PlanAlterObjectSchemaStmt(AlterObjectSchemaStmt *stmt, const char *queryString)
 			return PlanAlterFunctionSchemaStmt(stmt, queryString);
 		}
 
+		case OBJECT_EXTENSION:
+		{
```
onderkalaci: What about the other `ALTER EXTENSION` commands? 
We should have a higher level function like `PlanAlterExtensionStmt()`, in which we get into `PlanAlterExtensionSchemaStmt `.

And, in `PlanAlterExtensionStmt()` we handle others. As I've noted, we should give `NOTICE`, on `ADD/DROP` members since we're not going to implement them yet.

Also, we should plan/process ALTER EXTENSION UPDATE, which is probably the most important one

src/backend/distributed/deparser/deparse_extension_stmts.c
onderkalaci: Yes, we should only CASCADE if user specified it in `dropStmt`.

src/backend/distributed/metadata/dependency.c
```diff
@@ -315,6 +377,11 @@ SupportedDependencyByCitus(const ObjectAddress *address)
 				return true;
 			}
 
+			case OCLASS_EXTENSION:
```
onderkalaci: This is wrong place to add. We should add outside of `if (!EnableDependencyCreation)`  below switch

src/backend/distributed/commands/schema.c
onurctirtir: I think, we should only have an entry for PlanAlterExtensionSchemaStmt in PlanAlterObjectSchemaStmt, not for the other kind of AlterExtension commands because of the code base structure.

src/backend/distributed/metadata/dependency.c
onurctirtir: Done in 876af1603cbed2741240088ce6418d3f26a5e215

src/backend/distributed/commands/extension.c
onurctirtir: Done in 26e1acfcd1144e0ee2ed3527e51e8122ee340143

src/backend/distributed/commands/extension.c
```diff
@@ -113,5 +110,661 @@ ExtractNewExtensionVersion(Node *parsetree)
...
+	CreateExtensionStmt *stmt = makeNode(CreateExtensionStmt);
+
+	char *extensionName = get_extension_name(extensionOid);
+
+	/* set extension name and if_not_exists fields */
+	stmt->extname = extensionName;
+	stmt->if_not_exists = true;
+
+	/* append the missing schema name DefElem */
+	AddSchemaFieldIfMissing(stmt);
```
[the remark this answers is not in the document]
onurctirtir: Done in 7c6e2f615eb3eff90d3db66f14b0d86183b7b48b

src/backend/distributed/deparser/deparse_extension_stmts.c
onurctirtir: Changed the logic in 4b594c01afc5fa3d8638f47b2f541b8b0c0ecda4

src/backend/distributed/commands/extension.c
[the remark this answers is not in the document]
onurctirtir: This fixed my problem, thanks :)

onurctirtir:
Addressed new review items in 157d505fb31b6ed2406847280f7303d18d6216b0 and marked them as resolved

src/backend/distributed/commands/extension.c
```diff
@@ -78,40 +76,801 @@ ErrorIfUnstableCreateOrAlterExtensionStmt(Node *parsetree)
...
+		 * it does not matter if the it is about citus
+		 */
+		return false;
+	}
+
+	/*
+	 * Now that we have CreateExtensionStmt or AlterExtensionStmt,
+	 * check if it is run for/on citus
+	 */
+	return (strncmp(extensionName, CITUS_APPLICATION_NAME, NAMEDATALEN) == 0);
```
onderkalaci: Citus application name might change in the future, let's use
```C
strncasecmp(extensionName, CITUS_NAME, NAMEDATALEN) == 0
```

onurctirtir: But CITUS_NAME is equal to "Citus", if I am not missing something ?

onurctirtir: I just aware that you suggested me to use `strncasecmp` for this reason, sorry for this

src/backend/distributed/deparser/deparse_extension_stmts.c
```diff
@@ -0,0 +1,266 @@
...
+
+	/* "new_version" may not be specified in CreateExtensionStmt */
+	if (newVersion)
+	{
+		newVersion = quote_identifier(newVersion);
+
+		appendStringInfo(buf, " VERSION %s", newVersion);
+	}
+
+	appendStringInfoString(buf, " CASCADE;");
```
onderkalaci: I think we should only include CASCADE if the user includes it. Can you change that?

onurctirtir: done at 41be1d8

src/test/regress/expected/propagate_extension_commands.out
```diff
@@ -0,0 +1,399 @@
...
+-------
+     0
+(1 row)
+
+-- restore client_min_messages after DROP EXTENSION
+RESET client_min_messages;
+-- SET client_min_messages TO WARNING before executing a DROP EXTENSION statement
+SET client_min_messages TO WARNING;
+-- drop extension in a transaction block, should just work
+--BEGIN;
```
onderkalaci: undo the comments for `BEGIN/COMMIT`, and update the comment below -- TODO:: There is a bug now

src/test/regress/expected/propagate_extension_commands.out
```diff
@@ -0,0 +1,399 @@
...
+------------------------
+ 
+(1 row)
+
+COMMIT;
+-- add the node back
+SELECT master_add_node('localhost', :worker_2_port);
+NOTICE:  Replicating reference table "t3" to the node localhost:57638
+ master_add_node 
+-----------------
```
onderkalaci: master_add_node()'s output could change if someone adds a node before this test. so could you do something like

SELECT 1 FROM master_add_node()

please check all the tests in the file

src/test/regress/expected/propagate_extension_commands.out
onderkalaci: added this test

src/backend/distributed/commands/extension.c
```diff
@@ -78,40 +76,804 @@ ErrorIfUnstableCreateOrAlterExtensionStmt(Node *parsetree)
...
+	if (!schemaNameValue)
+	{
+		char *currentSchemaName = GetCurrentSchema();
+
+		Node *schemaNameArg = (Node *) makeString(currentSchemaName);
+
+		/* set location to -1 as it is unknown */
+		int location = -1;
+
+		DefElem *newDefElement = makeDefElem("schema", schemaNameArg, location);
```
marcocitus: Does this work correctly for extensions with a schema defined in their control file (like citus)?

onurctirtir: Actually, I didn't test my code with such an extension. Maybe I should test such an extension

onurctirtir: Actually, I didn't test my code with such an extension. Maybe I should test such an extension

onurctirtir: Now I tried it on hll and it worked

onderkalaci: Verified that creating Citus extension works fine (by changing some code). Also, other extensions as well

marcocitus:
One thing I don't quite understand: We sometimes create extension as a result of using a an object from the extension by going down the dependency tree. However, users generally do not have permissions to create extensions, only superuser does. Is it still useful to do this dependency resolution for non-superusers? Could it lead to issues?

onderkalaci:
> One thing I don't quite understand: We sometimes create extension as a result of using a an object from the extension by going down the dependency tree. However, users generally do not have permissions to create extensions, only superuser does. Is it still useful to do this dependency resolution for non-superusers? Could it lead to issues?

If the extension is already created, the dependency would just skip it. If the extension doesn't exists, any types depending on the extension wouldn't be avaliable, so the user cannot create any dependency to the extension?

marcocitus:
Actually, the CREATE EXTENSION is done as superuser. Makes sense now.

onderkalaci:
>  Could it lead to issues? (incl. confusing errors)

One edge case I come up with is the following -- which also seems to be working fine.

The superuser creates an extension, but with enable_ddl_propagation TO Off;
Later, another user create a table that depends on the a type that the extension provides. In that case, `EnsureDependencies()` would send `CREATE EXTENSION` to the workers, but that is happening with super user
```C

		SendCommandListToWorkerInSingleTransaction(nodeName, nodePort,
												   CitusExtensionOwnerName(),
												   ddlCommands);
```

So, works without any issues. Can you see any other edge cases?

marcocitus:
After fixing the duplicate enable_ddl_propagation issue this all looks good to me.

onurctirtir:
As a last comment, for the ones curios about the last issue, here is how I solved:

Actualy, this look like my bad. In in `EnsureDependenciesExistsOnAllNodes`, we call `GetDependencyCreateDDLCommands` for each dependency of the object, which calls `CreateExtensionDDLCommand` in turn. In that function, I was wrapping the resulting ddl command by `DISABLE/ENABLE_DDL_COMMAND` without noticing that `EnsureDependenciesExistsOnAllNodes` already does.

PR #3186, opened 2019-11-13
Strip trailing whitespace and add final newline

serprex:
Our `.editorconfig` directs files to not end with whitespace

First commit is the result of:
```
for f in `find src -iname '*.sql' -o -iname '*.c' -o -iname '*.h' -o -iname '*.py'`;do sed -e 's/[[:space:]]*$//' -i "$f";done
```

Second commit updates expected results of tests


JelteF:
I was talking with @onderkalaci this evening regarding the automatic refactor in this PR: https://github.com/citusdata/citus/pull/3181
He suggested to wait with running it until right after code freeze for 9.1, to not impact merges, but before creating the release, to make backports to the 9.1 release branch easy.

I think the same should be done with this one, because I think it is indeed good to do this change.

JelteF:
Also, I think it makes sense to make this a check in CI as well. And probably apart from not having trailing whitespace also make sure a final newline is part of all files.

JelteF:
@serprex Updated this PR to add a check to CI and and also update all final newlines.

PR #3188, opened 2019-11-14
Plan reference<->local table joins locally

pykello:
To do this, in the planner we check if the command is a SELECT which contains both local tables and reference tables, and if so replace all reference table rtes with the shard table and plan it locally.

In the executor, we make sure that we don't do these joins in transaction blocks because we might do the changes to the reference table over MultiConnections which won't be visible to the locally planned queries. This also saves us from worrying about locking, etc. for now.


src/backend/distributed/executor/multi_executor.c
```diff
@@ -631,3 +648,36 @@ AlterTableConstraintCheck(QueryDesc *queryDesc)
...
+
+	if (plannedStmt->commandType != CMD_SELECT)
+	{
+		return false;
+	}
+
+	foreach(oidCell, plannedStmt->relationOids)
+	{
+		Oid relationId = lfirst_oid(oidCell);
+		if (RelationIsAKnownShard(relationId))
```
pykello: No need to check if this is a reference table or not here. We wouldn't get here for non-reference distributed tables.

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -238,10 +256,31 @@ distributed_planner(Query *parse, int cursorOptions, ParamListInfo boundParams)
 						errhint("Consider using PL/pgSQL functions instead.")));
 	}
 
+	if (localReferenceTableJoin)
+	{
+		/* See comments for IsLocalReferenceTableJoinPlan */
+		result->utilityStmt = (Node *) parse;
```
pykello: This is pretty hacky and ugly way to do this, alternatives are (1) using CustomScan nodes (2) checking the parse tree in the executor again. 1 might be a bit more complicated to implement, 2 might slow down all of the worker queries quite a bit. This is a temporary solution and we will remove this after fixing the limitations of local execution. Let's chat about this more tomorrow.

marcocitus:
This doesn't work correctly for `WITH ins AS (INSERT INTO ref VALUES (1,2) RETURNING *) SELECT * FROM ins, local;`

src/backend/distributed/executor/multi_executor.c
```diff
@@ -70,6 +73,22 @@ CitusExecutorStart(QueryDesc *queryDesc, int eflags)
 {
 	PlannedStmt *plannedStmt = queryDesc->plannedstmt;
 
+	if (CitusHasBeenLoaded())
+	{
+		bool hasReferenceTableReplica = false;
+
+		PrimaryNodeForGroup(GetLocalGroupId(), &hasReferenceTableReplica);
```
marcocitus: This is clever, but really needs a comment.

pykello:
> WITH ins AS (INSERT INTO ref VALUES (1,2) RETURNING *) SELECT * FROM ins, local;

Fixed.

src/include/distributed/multi_logical_planner.h
```diff
@@ -213,7 +213,6 @@ extern List * TableEntryList(List *rangeTableList);
 extern List * UsedTableEntryList(Query *query);
 extern List * pull_var_clause_default(Node *node);
 extern bool OperatorImplementsEquality(Oid opno);
-extern bool FindNodeCheck(Node *node, bool (*check)(Node *));
```
pykello: there were two of these in the same file.

src/backend/distributed/executor/multi_executor.c
```diff
@@ -70,6 +73,28 @@ CitusExecutorStart(QueryDesc *queryDesc, int eflags)
...
+		 * All groups that have pg_dist_node entries, also have reference
+		 * table replicas.
+		 */
+		PrimaryNodeForGroup(GetLocalGroupId(), &hasReferenceTableReplica);
+
+		if (IsCoordinator() &&
+			hasReferenceTableReplica &&
+			IsLocalReferenceTableJoinPlan(plannedStmt))
+		{
+			if (IsMultiStatementTransaction())
```
marcocitus: maybe merge into above

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -118,7 +122,25 @@ distributed_planner(Query *parse, int cursorOptions, ParamListInfo boundParams)
...
+		 * table replicas.
+		 */
+		PrimaryNodeForGroup(GetLocalGroupId(), &hasReferenceTableReplica);
+
+		if (IsCoordinator() &&
+			hasReferenceTableReplica &&
+			IsLocalReferenceTableJoin(parse, rangeTableList))
+		{
+			needsDistributedPlanning = false;
+			UpdateReferenceTablesWithShard((Node *) parse, NULL);
```
marcocitus: would appreciate a comment here

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -118,7 +122,25 @@ distributed_planner(Query *parse, int cursorOptions, ParamListInfo boundParams)
...
+
+		/*
+		 * All groups that have pg_dist_node entries, also have reference
+		 * table replicas.
+		 */
+		PrimaryNodeForGroup(GetLocalGroupId(), &hasReferenceTableReplica);
+
+		if (IsCoordinator() &&
+			hasReferenceTableReplica &&
+			IsLocalReferenceTableJoin(parse, rangeTableList))
```
marcocitus: Maybe we can put these checks in a function?

pykello: done

src/backend/distributed/planner/distributed_planner.c
pykello: done

src/backend/distributed/executor/multi_executor.c
pykello: done

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -1770,3 +1787,172 @@ HasUnresolvedExternParamsWalker(Node *expression, ParamListInfo boundParams)
...
+	return hasLocalTable && hasReferenceTable;
+}
+
+
+/*
+ * QueryIsNotSimpleSelect returns true if node is a query which modifies or
+ * marks for modifications.
+ */
+static bool
+QueryIsNotSimpleSelect(Node *node)
```
onderkalaci: what about modifying CTEs?

pykello: Modifying CTEs contain queries which are not-select. In `IsLocalReferenceTableJoin` we recursively search for those by using `FindNodeCheck`.

PR #3191, opened 2019-11-14
Improve support for RECORD

serprex:
Copied bless logic from router planner

Support/tests for `RECORD[]` would be a different PR


src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -1412,6 +1412,11 @@ MasterExtendedOpNode(MultiExtendedOp *originalOpNode,
 			column->varoattno = walkerContext->columnId;
 			walkerContext->columnId++;
 
+			if (column->vartype == RECORDOID)
```
serprex: Should check if we're sometimes having redundant entry into this logic (ie having called `BlessRecordExpression` already earlier in planning), perhaps add a check to skip blessing when vartypmod already set

serprex: At least within the test suite the two blessing paths are mutually exclusive

PR #3195, opened 2019-11-15
Make distributed object dependency logic follow upto extensions

onderkalaci:
With this commit, we're slightly changing the dependency traversal
logic to enable extension propagation.

The main idea is to "follow" the extension dependencies, but do not
"apply" them.

Since some extension dependencies are base types, and base types
could have circular dependencies, we implement a logic to prevent
revisiting an already visited object.

### DIFF
--- a/src/backend/distributed/metadata/dependency.c
+++ b/src/backend/distributed/metadata/dependency.c
@@ -30,14 +30,17 @@
  * ObjectAddressCollector keeps track of collected ObjectAddresses. This can be used
  * together with recurse_pg_depend.
  *
- * We keep two different datastructures for the following reasons
+ * We keep three different datastructures for the following reasons
  *  - A List ordered by insert/collect order
  *  - A Set to quickly O(1) check if an ObjectAddress has already been collected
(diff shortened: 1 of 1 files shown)


src/backend/distributed/metadata/dependency.c
```diff
@@ -251,6 +273,46 @@ InitObjectAddressCollector(ObjectAddressCollector *collector)
...
+/*
+ * TargetObjectVisited returns true if the input target has been visited while
+ * traversing pg_depend.
+ */
+static bool
+TargetObjectVisited(ObjectAddressCollector *collector, const ObjectAddress *target)
+{
+	bool found = false;
+
+	/* add to set */
```
thanodnl: this comment should probably not be `add to set`, but more like `find in set`.

Guess this might have been copied from the function `IsObjectAddressCollected`. Let's fix there as well

SaitTalhaNisanci: Minor: I think we shouldn't have a comment here. The function is already self explanatory. I dont see any value in adding this comment, for me it makes the code dirty.

PR #3196, opened 2019-11-15
Propagate collations

serprex:
DESCRIPTION: Add propagation support for collations, along with `DROP COLLATION` & `ALTER COLLATION`

This makes #2907 less awkward, but is mostly unrelated to that PR

Todo:
- [x] `create_or_replace_object` support
- [x] qualify code

### DIFF
--- a/src/backend/distributed/commands/collation.c
+++ b/src/backend/distributed/commands/collation.c
@@ -1,0 +1,601 @@
+/*-------------------------------------------------------------------------
+ * collation.c
+ *
+ * This file contains functions to create, alter and drop policies on
+ * distributed tables.
+ *
+ * Copyright (c) 2019, Citus Data, Inc.
+ *
+ *-------------------------------------------------------------------------
+ */
+#include "postgres.h"
+
+#include "access/htup_details.h"
+#include "access/xact.h"
+#include "catalog/pg_collation.h"
+#include "distributed/commands/utility_hook.h"
+#include "distributed/commands.h"
+#include "distributed/deparser.h"
+#include "distributed/master_metadata_utility.h"
+#include "distributed/metadata/distobject.h"
+#include "distributed/metadata_sync.h"
+#include "distributed/multi_executor.h"
+#include "distributed/relation_access_tracking.h"
+#include "distributed/worker_create_or_replace.h"
+#include "distributed/worker_manager.h"
+#include "parser/parse_type.h"
+#include "utils/builtins.h"
+#include "utils/lsyscache.h"
+#include "utils/syscache.h"
+#include "miscadmin.h"
+
+
+static char * CreateCollationDDLInternal(Oid collationId, Oid *collowner,
+										 char **quotedCollationName);
+static List * FilterNameListForDistributedCollations(List *objects, bool missing_ok,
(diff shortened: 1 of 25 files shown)


serprex:
@thanodnl should I create a `create_distributed_collation` UDF? Right now if users want to force propagation, they have to distribute a table with a column that has 'col text COLLATE collation'

serprex:
@thanodnl https://app.circleci.com/jobs/github/citusdata/citus/62064 it seems that by having `distributed_functions` & `distributed_collations` regression tests run in parallel we can sometimes hit deadlocks with the object distribution logic

thanodnl:
re: the `create_distributed_collation` UDF, I'd rather refrain from it but if we need it we might indeed want to go the route of the more generic `create_distributed_object` or `distribute_object`.

In what cases would one want to have the collation distributed outside of a column in a table? I thought the collation is part of the type in a table.

I guess one of the reasons to force distribution could be if collations could be used in queries. Marco was contemplating a just in time distributions for aggregates earlier. It might be that we would like to hook the planner into the distributed objects at some point to verify and distribute if not already done so, objects required in queries. For this we need to improve the speed at which we can lookup if an object is distributed, possibly via syscache infra structure.

Could you shed some light on the situations in which a collation is required to be distributed outside of the use in a column?

serprex:
@thanodnl no light to shed, use in queries is the use case

pykello:
> Could you shed some light on the situations in which a collation is required to be distributed outside of the use in a column?

Collations that are created using the [CREATE COLLATION](https://www.postgresql.org/docs/current/sql-createcollation.html) commands by the user might be easier to distributed/managed similar to types, ... A user might want to use a collation which is not installed by `initdb` by default.

For example when I tried, initdb didn't install the German collation with phone book ordering and I had to create it manually:

```
CREATE COLLATION german_phonebook (provider = icu, locale = 'de-u-co-phonebk');
```

I guess the usecase for having them in `pg_dist_object` would be distributing these kinds of collations.

serprex:
> For types we do the initial propagation during CREATE TYPE (if it's not in a transaction block). Would it be worth doing the same here to make sure user-defined collations can be used in queries?

Implemented

src/backend/distributed/commands/collation.c
```diff
@@ -0,0 +1,601 @@
...
+	}
+
+	/* fully qualify */
+	QualifyTreeNode((Node *) stmt);
+
+	/* deparse sql*/
+	char *renameStmtSql = DeparseTreeNode((Node *) stmt);
+
+	/* to prevent recursion with mx we disable ddl propagation */
+	EnsureSequentialModeForCollationDDL();
```
marcocitus: comment does not match line

serprex: this matches the comments in `commands/type.c`

serprex: Also updated instances of this comment in `commands/type.c`

serprex:
Agreed that there seems to be a lot more code than necessary for deparsing currently. I have some ideas about how to consolidate the code so that adding new objects to deparse will be less error prone & more straightforward

PR #3197, opened 2019-11-18
Prevent Citus extension becoming distributed object

onderkalaci:
Because that could complicate the role management. There could be some
other edge cases, so I'd prefer to keep Citus extension outside the picture for now.

DESCRIPTION: Prevent Citus extension becoming distributed object


onderkalaci:
The patch coverage is low, because we'll trigger this case with #3185

src/backend/distributed/commands/dependencies.c
```diff
@@ -67,7 +67,19 @@ EnsureDependenciesExistsOnAllNodes(const ObjectAddress *target)
...
 		ObjectAddress *dependency = (ObjectAddress *) lfirst(dependencyCell);
-		List *dependencyCommands = GetDependencyCreateDDLCommands(dependency);
+		List *dependencyCommands = NIL;
+
+		if (CitusExtensionObject(dependency))
+		{
+			/*
+			 * We currently do not prefer to mark Citus extension as distributed
+			 * because it could prevent some use use-cases like upgrading the
+			 * coordinator before the workers.
```
marcocitus: note that upgrading the coordinator before the workers is never supported

src/backend/distributed/metadata/distobject.c
```diff
@@ -141,6 +148,16 @@ MarkObjectDistributed(const ObjectAddress *distAddress)
 	char *insertQuery = "INSERT INTO citus.pg_dist_object (classid, objid, objsubid) "
 						"VALUES ($1, $2, $3) ON CONFLICT DO NOTHING";
 
+	if (CitusExtensionObject(distAddress))
```
marcocitus: I wouldn't expect this behaviour in this function. Can we move the check to `EnsureDependenciesExistsOnAllNodes`?

thanodnl:
Good catch. I missed that citus distributed tables will end up depending on citus!

I think I would prefer to not 'follow' the citus extension during recursing of `pg_depend`.  This will solve the problem at the root. Besides that I would suggest to keep the current checks in there, although it should be dead code. Having the citus extension as a distributed object for now does create a lot of complications.

onderkalaci:
> I think I would prefer to not 'follow' the citus extension during recursing of pg_depend. This will solve the problem at the root. 

I considered that, but do we really want to stop traversing the dependencies when we see the Citus extension? Can't we have objects that depend on Citus, but becomes distributed objects?

Should we maybe add this check in "apply" @thanodnl ?

thanodnl:
> Can't we have objects that depend on Citus, but becomes distributed objects?

Those objects can still become distributed. In the suggestion above we would stop the traversal when we already reached the citus extension, causing the extension and all its dependencies not to be traversed. If a user's object depends on a common dependency with citus those objects would still be found and distributed if required.

Terminating the traversal via a `false` returned from the `follow` function makes it so the citus extension will never show in a list of dependencies of anything.

onderkalaci:
> causing the extension and all its dependencies not to be traversed

Yes, that's what I was considering. Say,  as a user, you create a type or function, which depends on Citus extension. If we do not "follow" citus,  wouldn't we ignore the type/function as a distributed objects?

I'm OK with moving the check to "follow", just wanted to make sure that you've considered the above as well.

(And, remember that, in  any case we rely on the fact that the Citus extension to already exists on the workers, that's why "apply" sounds OK to me)

src/backend/distributed/metadata/dependency.c
```diff
@@ -635,6 +635,10 @@ ApplyAddToDependencyList(ObjectAddressCollector *collector, Form_pg_depend pg_de
 	{
 		return;
 	}
+	else if (CitusExtensionObject(&address))
```
thanodnl: When in the `apply` function it will still resolve any dependencies the Citus extension might have which is undesirable here.

PR #3204, opened 2019-11-19
Rely on original query for intermediate result pruning

onderkalaci:
Basically, on every successful distributed plan creation, make sure
to record the subPlans used in the plan, by checking the originalQuery.

### DIFF
--- a/src/backend/distributed/planner/distributed_planner.c
+++ b/src/backend/distributed/planner/distributed_planner.c
@@ -26,6 +26,7 @@
 #include "distributed/metadata_cache.h"
 #include "distributed/multi_executor.h"
 #include "distributed/distributed_planner.h"
+#include "distributed/query_pushdown_planning.h"
 #include "distributed/multi_logical_optimizer.h"
 #include "distributed/multi_logical_planner.h"
 #include "distributed/multi_partitioning_utils.h"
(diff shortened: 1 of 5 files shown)


src/backend/distributed/planner/distributed_planner.c
```diff
@@ -810,13 +817,43 @@ CreateDistributedPlan(uint64 planId, Query *originalQuery, Query *query, ParamLi
...
+/*
+ * RecordSubPlansUsedInPlan gets a distributed plan a queryTree, and
+ * updates the usedSubPlanNodeList of the distributed plan.
+ *
+ * The function simply pulls all the subPlans that are used in the queryTree
+ * with one exception: subPlans in the HAVING clause. The reason is explained
+ * in the function.
+ */
+static void
+RecordSubPlansUsedInPlan(DistributedPlan *plan, Query *queryTree)
```
marcocitus: maybe preserve `originalQuery` for consistency, since `queryTree` sometimes refers to the rewritten query.

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -779,6 +784,8 @@ CreateDistributedPlan(uint64 planId, Query *originalQuery, Query *query, ParamLi
 	 */
 	if (IsModifyCommand(originalQuery))
 	{
+		RecordSubPlansUsedInPlan(distributedPlan, originalQuery);
```
marcocitus: Maybe create a wrapper called `FinalizeDistributedPlan` that just calls this function. This is mainly to make it easier to understand what's going on here and that for instance, the call to `RecordSubPlansUsedInPlan` is not related to `IsModifyCommand`.

PR #3207, opened 2019-11-19
Fix potential segfault from standard_planner inlining functions

serprex:
Calling `GetRTEIdentity` on a `RangeTblEntry` which was added midway through `distributed_planner` segfaults, as we collect a list of `RangeTblEntry`s at the very beginning of `distributed_planner`

This could happen if `standard_planner` inlined a function call to something like `SELECT * FROM table`, since we function definitions aren't part of the initial query tree

### DIFF
--- a/src/backend/distributed/planner/distributed_planner.c
+++ b/src/backend/distributed/planner/distributed_planner.c
@@ -80,7 +80,7 @@ static void RecordSubPlansUsedInPlan(DistributedPlan *plan, Query *originalQuery
 static DeferredErrorMessage * DeferErrorIfPartitionTableNotSingleReplicated(Oid
 																			relationId);
 
-static void AssignRTEIdentities(List *rangeTableList);
+static int AssignRTEIdentities(List *rangeTableList, int rteIdCounter);
 static void AssignRTEIdentity(RangeTblEntry *rangeTableEntry, int rteIdentifier);
 static void AdjustPartitioningForDistributedPlanning(List *rangeTableList,
 													 bool setPartitionedTablesInherited);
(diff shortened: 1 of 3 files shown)


pykello:
@serprex Can you update the description with an example of a query that caused the crashed, and what was the backtrace for the crash, and what was the reason for crash?

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -112,6 +112,7 @@ distributed_planner(Query *parse, int cursorOptions, ParamListInfo boundParams)
 	PlannerRestrictionContext *plannerRestrictionContext = NULL;
 	bool setPartitionedTablesInherited = false;
 	List *rangeTableList = ExtractRangeTableEntryList(parse);
+	int rteIdentifier = 1;
```
pykello: Maybe `rteIdCounter`

src/test/regress/expected/multi_subquery_union.out
```diff
@@ -1237,5 +1237,14 @@ ORDER BY types;
...
 (4 rows)
 
+-- Previously this produced a segfault from standard_planner introducing a subquery after we'd called AssignRTEIdentities
+CREATE OR REPLACE FUNCTION users_udf()
+RETURNS TABLE(user_id int)
+AS $$SELECT user_id FROM users_reference_table;$$
+LANGUAGE sql stable;
+SELECT user_id FROM users_table
+UNION SELECT u.user_id FROM users_table, users_udf() u;
+ERROR:  cannot perform distributed planning on this query because parameterized queries for SQL functions referencing distributed tables are not supported
```
marcocitus: Why do we hit this error?

serprex: `distributed_planner` ends up being called on `SELECT user_id FROM public.users_udf() u(user_id)` while planning this query

PR #3209, opened 2019-11-20
refactor some parts in adaptive executor

SaitTalhaNisanci:


src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -2347,60 +2357,7 @@ ConnectionStateMachine(WorkerSession *session)
...
-											connection->hostname, connection->port,
-											session->sessionId)));
-
-					workerPool->activeConnectionCount++;
-					workerPool->idleConnectionCount++;
-
-					UpdateConnectionWaitFlags(session,
-											  WL_SOCKET_READABLE | WL_SOCKET_WRITEABLE);
-
-					connection->connectionState = MULTI_CONNECTION_CONNECTED;
```
marcocitus: I find it useful for statement machines to explicitly show their state changes. The switch already organizes the code under different headers. I'm not sure what we gain from this change.

SaitTalhaNisanci: When the block is small in case statements then it might be easier to see it. When it is long, I find it harder to understand because it is harder to navigate in my opinion. Also while refactoring I removed some duplicate code.

marcocitus: This code defines a state machine, rather than an imperative algorithm. To be able to understand the correctness of the code, it's important that you can easily map it to the logical state machine transitions. In that sense it's useful to keep all the state changes close to each other in the same switch statement, rather than dispersed across functions. The case statements already provide a clean syntactic grouping of code blocks. 

As the core execution logic in Citus, ConnectionStateMachine and TransactionStateMachine have been quite carefully crafted over many months to look the way they do. There is room for improvement, but from the code review perspective this is probably going to be the hardest piece of code in Citus to refactor. On the other hand, they are in serious need of some comments (and maybe diagrams).

SaitTalhaNisanci: I see, a diagram could be useful I guess. Are you okay with creating a function for the duplicate code for this PR?

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -2507,6 +2488,28 @@ ConnectionStateMachine(WorkerSession *session)
 }
 
 
+/*
+ * HandleMultiConnectionSuccess logs the established connection and updates connection's state.
+ */
+static void
+HandleMultiConnectionSuccess(WorkerPool *workerPool, MultiConnection *connection,
+							 WorkerSession *session)
```
marcocitus: I think you could reduce this to just the session, and obtain the connection and pool from there.

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -2507,6 +2488,28 @@ ConnectionStateMachine(WorkerSession *session)
...
+							connection->hostname, connection->port,
+							session->sessionId)));
+
+	workerPool->activeConnectionCount++;
+	workerPool->idleConnectionCount++;
+
+	UpdateConnectionWaitFlags(session,
+							  WL_SOCKET_READABLE | WL_SOCKET_WRITEABLE);
+
+	connection->connectionState = MULTI_CONNECTION_CONNECTED;
```
marcocitus: I'd have a strong preference for keeping the last line in the state machine to clearly show the state change. I'd have a slight preference for keeping these last 2 lines in the state machine, mostly to keep the UpdateConnectionWaitFlags calls together.

SaitTalhaNisanci: That makes sense, I will keep the last 2 lines in the state machine.

PR #3211, opened 2019-11-20
Implement support for RECORD[] where we support RECORD

serprex:
DESCRIPTION: Support `RECORD[]` columns in queries where we support `RECORD`

TODO:
- [x] test `RECORD[]` being returned by function *(sql & plpgsql functions cannot return `RECORD[]`)*
- [x] test multidimensional/nested array

Future PR can address: *(started a branch, `support-more-record-expressions`)*
- support other expression types are not covered by `FuncExpr`/`RowExpr`/`ArrayExpr`
  - `OpExpr` & `NullIfExpr` *(the latter for `NULLIF((1,2),(3,4))`)*
  - `FuncExpr` or `OpExpr` with function whose return type is `ANYELEMENT`
  - `CoalesceExpr`, `MinMaxExpr`, `CaseExpr`

### DIFF
--- a/src/backend/distributed/planner/distributed_planner.c
+++ b/src/backend/distributed/planner/distributed_planner.c
@@ -1169,7 +1169,7 @@ FinalizeRouterPlan(PlannedStmt *localPlan, CustomScan *customScan)
 		/* build target entry pointing to remote scan range table entry */
 		Var *newVar = makeVarFromTargetEntry(customScanRangeTableIndex, targetEntry);
 
-		if (newVar->vartype == RECORDOID)
+		if (newVar->vartype == RECORDOID || newVar->vartype == RECORDARRAYOID)
 		{
 			/*
 			 * Add the anonymous composite type to the type cache and store
@@ -1263,7 +1263,7 @@ BlessRecordExpression(Expr *expr)
 			Oid rowArgTypeId = exprType(rowArg);
 			int rowArgTypeMod = exprTypmod(rowArg);
 
-			if (rowArgTypeId == RECORDOID)
+			if (rowArgTypeId == RECORDOID || rowArgTypeId == RECORDARRAYOID)
 			{
 				/* ensure nested rows are blessed as well */
 				rowArgTypeMod = BlessRecordExpression((Expr *) rowArg);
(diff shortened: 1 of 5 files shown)


src/test/regress/sql/row_types.sql
```diff
@@ -29,13 +29,15 @@ INSERT INTO test VALUES (1,2), (1,3), (2,2), (2,3);
 -- multi-shard queries support row types
 SELECT (x,y) FROM test ORDER BY x, y;
 SELECT (x,y) FROM test GROUP BY x, y ORDER BY x, y;
+SELECT ARRAY[(x,(y,x)),(y,(x,y))] FROM test ORDER BY x, y;
```
onderkalaci:
```SQL
 SELECT ARRAY[(x,(y,x)),(y,(x,y),x)] FROM test WHERE x != 2 ORDER BY x, y;
```

The following query works fine on Postgres, but  fails on this patch with Citus.  Can we make it work?

Also, what about ARRAY's inside ARRAY, such as ` ARRAY[ARRAY[(x,(y,x)), (x,(y,x))],ARRAY[(y,(x,y))]]`, do they work with this patch?

serprex: Nested arrays work, will include test case. I was not aware that arrays could have varying shapes for RECORDs, will look into further

serprex: Don't believe arrays with varying RECORD shape can be supported. `array_in` expects all its elements to take the same typmod as it has

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -1281,6 +1281,32 @@ BlessRecordExpression(Expr *expr)
...
+			Node *elemArg = (Node *) lfirst(elemCell);
+			int32 arrayTypeMod = BlessRecordExpression((Expr *) elemArg);
+
+			if (typeMod == -1)
+			{
+				typeMod = arrayTypeMod;
+			}
+			else if (arrayTypeMod != typeMod)
+			{
+				return -1;
```
marcocitus: Could use a comment.

serprex:
Issue: we currently reject `ARRAY[(x,y),NULL]`. Will be fixed in next `RECORD` PR

Issue #3216, opened 2019-11-22
CREATE EXTENSION propagation does not escape extension name


marcocitus:
Saw the following error:
```sql
2019-11-22 14:58:07.288 UTC [97] ERROR:  syntax error at or near "-"
2019-11-22 14:58:07.288 UTC [97] CONTEXT:  while executing command on 10.0.0.6:5432
2019-11-22 14:58:07.288 UTC [97] STATEMENT:  SELECT * from master_add_node($1, $2)
```

Caused by not escaping the extension name in the new extension propagation feature:
```sql
2019-11-22 15:05:35.963 UTC [153] ERROR:  syntax error at or near "-" at character 36
2019-11-22 15:05:35.963 UTC [153] STATEMENT:  CREATE EXTENSION IF NOT EXISTS uuid-ossp WITH SCHEMA public ;
```

serprex:
There's more escaping issues, such as use of `regprocedureout` in `GetFunctionAlterOwnerCommand`. Will look into

metdos:
@onurctirtir did we fix this with #3218? If yes, could you close the issue?

onurctirtir:
Fixed in #3218 hence closing

PR #3218, opened 2019-11-22
Escape extension name in extension command propagation

onurctirtir:
This PR aims to fix the bug identified in https://github.com/citusdata/citus/issues/3216
* Now, we escape the extension name in deparse step to propagate the extension command to worker nodes properly.

### DIFF
--- a/src/test/regress/expected/propagate_extension_commands.out
+++ b/src/test/regress/expected/propagate_extension_commands.out
@@ -183,7 +183,7 @@ SELECT create_reference_table('ref_table_2');
 (1 row)
 
 -- and add the other node
-SELECT 1 from master_add_node('localhost', 57638);
+SELECT 1 from master_add_node('localhost', :worker_2_port);
 NOTICE:  Replicating reference table "ref_table_2" to the node localhost:57638
  ?column? 
 ----------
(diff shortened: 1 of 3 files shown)


serprex:
`uuid-ossp` should be added as a test case

onurctirtir:
Added a regression test for the fix as well

src/test/regress/expected/propagate_extension_commands.out
```diff
@@ -385,6 +385,25 @@ SELECT run_command_on_workers($$SELECT count(*) FROM pg_extension WHERE extname
  (localhost,57638,t,2)
 (2 rows)
 
+-- test if citus can escape the extension name
+CREATE EXTENSION "uuid-ossp";
```
onderkalaci: before merging in, please make sure that it works with master_add_node as well?

It's good that the regression test file has a schema  name which requires escaping, so we've tested that.

onurctirtir: Updated propagate_extension_commands.sql to test the fix with master_add_node as well

marcocitus:
Thanks. Can you cherry pick onto the release branches as well?

Minor note: We might want to add alternative output in case uuid-ossp is not installed.

SaitTalhaNisanci:
@marcocitus @onurctirtir Should we add an alternative output in case `uuid-ossp` is not installed. Otherwise this requires us installing this extension in some other places. For example to run valgrind tests in test automation we will need to add this extension.

marcocitus:
No objections. There are other contrib/ extensions in these test files, but I guess uuid-ossp is a bit harder to install than the others.

fwiw, I compile postgres with `--with-uuid=e2fs` to compile uuid-ossp.

PR #3220, opened 2019-11-23
Test coordinator coherence

serprex:
Extended regression tests to cover some items in release testing

- local/reference table joins
- `ALTER ROLE` propagation with coordinator in pg_dist_node, & no propagation from MX workers

`IsMultiStatementTransaction` check was moved in `multi_executor.c` because the `FunctionCallLevel` gets reset between statements, so it's only valid to check in `CitusExecutorRun`

### DIFF
--- a/src/test/regress/expected/replicate_reference_tables_to_coordinator.out
+++ b/src/test/regress/expected/replicate_reference_tables_to_coordinator.out
@@ -113,11 +113,70 @@ SELECT local_table.a, numbers.a FROM local_table NATURAL JOIN numbers ORDER BY 1
  20 | 20
 (1 row)
 
+-- test non equijoin
+SELECT lt.a, sq.a, sq.b
+FROM local_table lt
+JOIN squares sq ON sq.a > lt.a and sq.b > 90
+ORDER BY 1,2,3;
+ a | a  |  b  
+---+----+-----
+ 2 | 10 | 100
+ 4 | 10 | 100
+ 7 | 10 | 100
+(3 rows)
+
 -- error if in transaction block
 BEGIN;
 SELECT local_table.a, numbers.a FROM local_table NATURAL JOIN numbers ORDER BY 1;
 ERROR:  cannot join local tables and reference tables in a transaction block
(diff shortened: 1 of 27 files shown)


serprex:
@pykello has pointed out that we should be raising an error for the local/reference joins which occur inside function calls

src/test/regress/expected/replicate_reference_tables_to_coordinator.out
```diff
@@ -113,11 +113,91 @@ SELECT local_table.a, numbers.a FROM local_table NATURAL JOIN numbers ORDER BY 1
...
+$$;
+ERROR:  cannot join local tables and reference tables in a transaction block
+CONTEXT:  SQL statement "SELECT local_table.a, numbers.a FROM local_table NATURAL JOIN numbers"
+PL/pgSQL function inline_code_block line 3 at PERFORM
+-- test from within function call
+CREATE FUNCTION test_reference_local_join_func()
+RETURNS RECORD AS $$
+SELECT local_table.a, numbers.a FROM local_table NATURAL JOIN numbers ORDER BY 1;
+$$ LANGUAGE sql;
+SELECT test_reference_local_join_func();
```
pykello: This should also error out, but is unrelated to this PR. I created an issue: https://github.com/citusdata/citus/issues/3256

To hit the issue, maybe you can replace your test with:

```sql
CREATE or replace FUNCTION test_reference_local_join_func()
RETURNS SETOF RECORD AS $$
SET LOCAL citus.enable_local_execution to false;
INSERT INTO numbers VALUES (2);
SELECT local_table.a, numbers.a FROM local_table NATURAL JOIN numbers ORDER BY 1;
$$ LANGUAGE sql;
SELECT test_reference_local_join_func();
```

We should get two rows, but currently it returns one row.

I'd wait on merging this until https://github.com/citusdata/citus/issues/3256 is fixed.

pykello: Added this test to #3257

Issue #3080, opened 2019-10-07
Move all qualifying logic in deparse* funcs to qualify* funcs


hanefi:
When deparsing `{ALTER|DROP} .. {FUNCTION|PROCEDURE}` queries, we qualify the function/procedure and parameter type names. This logic should be implemented in `Qualify*FunctionStmt()` functions.

As @onderkalaci suggested today, `regprocedureout` can be handy in solving this
https://github.com/citusdata/citus/blob/bffd1104464127690864ea6034399e908aee7aa7/src/backend/distributed/commands/function.c#L622-L627

PR #3221, opened 2019-11-24
Review identifier escaping

serprex:
Reviewed all code matching `[A-Z] %s`

* GetFunctionAlterOwnerCommand: use format_procedure_qualified
* distributed_functions: test a function with a quote in name
* AppendDefElemSet: escape keys & values

`regprocedureout` uses `format_procedure` under the hood. `format_procedure` & `format_procedure_qualified` share their implementation, the latter passing a parameter to achieve what we were previously achieving with search_path juggling

### DIFF
--- a/src/backend/distributed/deparser/deparse_function_stmts.c
+++ b/src/backend/distributed/deparser/deparse_function_stmts.c
@@ -274,25 +274,27 @@ AppendDefElemSet(StringInfo buf, DefElem *def)
 	{
 		case VAR_SET_VALUE:
 		{
-			appendStringInfo(buf, " SET %s = %s", setStmt->name, setVariableArgs);
+			appendStringInfo(buf, " SET %s = %s", quote_identifier(setStmt->name),
+							 setVariableArgs);
 			break;
 		}
 
 		case VAR_SET_CURRENT:
 		{
-			appendStringInfo(buf, " SET %s FROM CURRENT", setStmt->name);
+			appendStringInfo(buf, " SET %s FROM CURRENT", quote_identifier(
+								 setStmt->name));
 			break;
 		}
 
 		case VAR_SET_DEFAULT:
 		{
-			appendStringInfo(buf, " SET %s TO DEFAULT", setStmt->name);
+			appendStringInfo(buf, " SET %s TO DEFAULT", quote_identifier(setStmt->name));
 			break;
 		}
 
 		case VAR_RESET:
 		{
-			appendStringInfo(buf, " RESET %s", setStmt->name);
+			appendStringInfo(buf, " RESET %s", quote_identifier(setStmt->name));
 			break;
 		}
 
--- a/src/backend/distributed/deparser/deparse_function_stmts.c
+++ b/src/backend/distributed/deparser/deparse_function_stmts.c
@@ -58,6 +58,7 @@ static void AppendDefElemCost(StringInfo buf, DefElem *def);
 static void AppendDefElemRows(StringInfo buf, DefElem *def);
 static void AppendDefElemSet(StringInfo buf, DefElem *def);
 
+static void AppendVarSetValue(StringInfo buf, VariableSetStmt *setStmt);
 static void AppendRenameFunctionStmt(StringInfo buf, RenameStmt *stmt);
 static void AppendAlterFunctionSchemaStmt(StringInfo buf, AlterObjectSchemaStmt *stmt);
 static void AppendAlterFunctionOwnerStmt(StringInfo buf, AlterOwnerStmt *stmt);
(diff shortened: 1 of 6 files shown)


src/backend/distributed/deparser/deparse_function_stmts.c
```diff
@@ -274,25 +274,27 @@ AppendDefElemSet(StringInfo buf, DefElem *def)
 	{
 		case VAR_SET_VALUE:
 		{
-			appendStringInfo(buf, " SET %s = %s", setStmt->name, setVariableArgs);
+			appendStringInfo(buf, " SET %s = %s", quote_identifier(setStmt->name),
```
marcocitus: Can we add a test case for this?

Note that you can always SET prefix.whatever to something and read it using current_setting.

onderkalaci:
marking #3080 as related issue

src/backend/distributed/deparser/deparse_function_stmts.c
```diff
@@ -274,25 +274,27 @@ AppendDefElemSet(StringInfo buf, DefElem *def)
...
+							 setVariableArgs);
```
marcocitus: Seems this needs to be escaped as well:
```
ALTER FUNCTION default_setting() SET "citus.setting;'" TO 'hello '' world';
LOG:  issuing BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED;SELECT assign_distributed_transaction_id(0, 1, '2019-11-23 10:51:37.252897+01');
DETAIL:  on server marco@localhost:9701
LOG:  issuing BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED;SELECT assign_distributed_transaction_id(0, 1, '2019-11-23 10:51:37.252897+01');
DETAIL:  on server marco@localhost:9700
LOG:  issuing SET citus.enable_ddl_propagation TO 'off';ALTER FUNCTION public.default_setting() SET "citus.setting;'" = hello ' world;;SET citus.enable_ddl_propagation TO 'on';
DETAIL:  on server marco@localhost:9701
LOG:  issuing SET citus.enable_ddl_propagation TO 'off';ALTER FUNCTION public.default_setting() SET "citus.setting;'" = hello ' world;;SET citus.enable_ddl_propagation TO 'on';
DETAIL:  on server marco@localhost:9700
ERROR:  syntax error at or near "' world;;SET citus.enable_ddl_propagation TO '"
```

marcocitus: I suspect we'll need to copy-paste this block from postgres, including simple_quote_literal:
```c
                /*
                 * Variables that are marked GUC_LIST_QUOTE were already fully
                 * quoted by flatten_set_variable_args() before they were put
                 * into the proconfig array.  However, because the quoting
                 * rules used there aren't exactly like SQL's, we have to
                 * break the list value apart and then quote the elements as
                 * string literals.  (The elements may be double-quoted as-is,
                 * but we can't just feed them to the SQL parser; it would do
                 * the wrong thing with elements that are zero-length or
                 * longer than NAMEDATALEN.)
                 *
                 * Variables that are not so marked should just be emitted as
                 * simple string literals.  If the variable is not known to
                 * guc.c, we'll do that; this makes it unsafe to use
                 * GUC_LIST_QUOTE for extension variables.
                 */
                if (GetConfigOptionFlags(configitem, true) & GUC_LIST_QUOTE)
                {
                    List       *namelist;
                    ListCell   *lc;

                    /* Parse string into list of identifiers */
                    if (!SplitGUCList(pos, ',', &namelist))
                    {
                        /* this shouldn't fail really */
                        elog(ERROR, "invalid list syntax in proconfig item");
                    }
                    foreach(lc, namelist)
                    {
                        char       *curname = (char *) lfirst(lc);

                        simple_quote_literal(&buf, curname);
                        if (lnext(lc))
                            appendStringInfoString(&buf, ", ");
                    }
                }
                else
                    simple_quote_literal(&buf, pos);
```

serprex: Another error: `ALTER FUNCTION add(int, int) SET TIME ZONE INTERVAL '-08:00' HOUR TO MINUTE;`

Will update PR with regression tests & fixes today

marcocitus: Pushed a few tests that fail, but should succeed after the fix.

marcocitus: If the fix gets too complicated, then failing on some of these seems ok as long as it only fails for distributed functions.

src/backend/distributed/deparser/deparse_function_stmts.c
```diff
@@ -268,31 +268,128 @@ static void
...
 
 	switch (setStmt->kind)
 	{
 		case VAR_SET_VALUE:
 		{
-			appendStringInfo(buf, " SET %s = %s", setStmt->name, setVariableArgs);
+			ListCell *varArgCell = NULL;
+			ListCell *firstCell = list_head(setStmt->args);
+
+			/* see flatten_set_variable_args in postgres's utils/misc/guc.c */
```
marcocitus: This should probably live in its own function with a comment to clarify the quoting difference with flatten_set_variable_args.

PR #3224, opened 2019-11-25
Fix underflow initialization of default values

fereidani:
Initialization of queryWindowClause and queryOrderByLimit "memset" underflow these variables.
It's possible due to invalid usage of sizeof this part of the program cause buffer overflow and function return data corruption in future changes.


marcocitus:
Thanks @fereidani ! Did you use a particular tool to find the issue?

PR #3225, opened 2019-11-25
Fix distributed aggregation for non superuser roles

### LINKED ISSUE #3222: Custom aggregates throw permission denied error when run as non-superuser
When using a custom aggregate as non-superuser, Citus throws a permission denied error:
```
ALTER TABLE test ADD COLUMN z geometry;
SELECT ST_MemUnion(z) FROM test; 
ERROR:  permission denied for schema citus
```

serprex:
Moves support functions to pg_catalog for now. We'd prefer a different solution
for when we're creating these support functions dynamically

Fixes #3222

### DIFF
--- a/src/test/regress/expected/aggregate_support.out
+++ b/src/test/regress/expected/aggregate_support.out
@@ -162,5 +162,30 @@ select array_collect_sort(val) from aggdata;
  {0,2,2,3,4,5,8,NULL,NULL,NULL,NULL}
 (1 row)
 
+-- Test multiuser scenario
+create user notsuper;
+NOTICE:  not propagating CREATE ROLE/USER commands to worker nodes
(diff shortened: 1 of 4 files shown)


src/test/regress/expected/aggregate_support.out
```diff
@@ -162,5 +162,24 @@ select array_collect_sort(val) from aggdata;
  {0,2,2,3,4,5,8,NULL,NULL,NULL,NULL}
 (1 row)
 
+-- Test multiuser scenario
+select run_command_on_master_and_workers($$
+create user notsuper;
+grant all on schema aggregate_support to notsuper;
+grant all on table aggdata to notsuper;
+$$);
+ERROR:  function run_command_on_master_and_workers(unknown) does not exist
```
serprex: This is wrong

Issue #3182, opened 2019-11-12
Async metadata sync can cause replica inconsistency


pykello:
Some functions like `master_update_node()` mark `metadatasynced=false` so maintenanced asynchronously propagates metadata to worker nodes. If during this period we activate a node from the coordinator, the change won't be available immediately in the metadata workers. So DML from metadata nodes can result in inconsistent replicas. I think the fix is to disallow users from activating a node when some metadata nodes are out of sync.

marcocitus:
Can this be closed after #3226?

PR #3226, opened 2019-11-25
Error for metadata commands if any metadata node is out-of-sync

pykello:
Previously for metadata commands we just didn't propagate them to out-of-sync nodes, assuming that because these nodes will get the metadata state shortly after, we should be ok.

But problems arise if user does DML from metadata nodes while it is out-of-sync. Because it doesn't have the most recent image of the cluster, the DML might be routed to the wrong nodes and we might get replica inconsistency.

Address https://github.com/citusdata/citus/issues/3182

### DIFF
--- a/src/backend/distributed/transaction/worker_transaction.c
+++ b/src/backend/distributed/transaction/worker_transaction.c
@@ -32,6 +32,15 @@
 #include "utils/memutils.h"
 
 
+static void SendCommandToMetadataWorkersParams(const char *command,
+											   const char *user, int parameterCount,
+											   const Oid *parameterTypes, const
+											   char *const *parameterValues);
+static void SendCommandToWorkersParamsInternal(TargetWorkerSet targetWorkerSet,
+											   const char *command, const char *user,
+											   int parameterCount, const
+											   Oid *parameterTypes,
+											   const char *const *parameterValues);
 static void ErrorIfAnyMetadataNodeOutOfSync(List *metadataNodeList);
 
 
@@ -57,11 +66,6 @@ SendCommandToWorkersAsUser(TargetWorkerSet targetWorkerSet, const char *nodeUser
 {
 	List *workerNodeList = TargetWorkerSetNodeList(targetWorkerSet, ShareLock);
 	ListCell *workerNodeCell = NULL;
-
-	if (targetWorkerSet == WORKERS_WITH_METADATA)
-	{
-		ErrorIfAnyMetadataNodeOutOfSync(workerNodeList);
-	}
 
 	/* run commands serially */
 	foreach(workerNodeCell, workerNodeList)
@@ -106,10 +110,10 @@ SendCommandToWorkerAsUser(char *nodeName, int32 nodePort, const char *nodeUser,
  * owner to ensure write access to the Citus metadata tables.
  */
 void
-SendCommandToWorkers(TargetWorkerSet targetWorkerSet, const char *command)
+SendCommandToWorkersWithMetadata(const char *command)
 {
-	SendCommandToWorkersParams(targetWorkerSet, command, CitusExtensionOwnerName(),
-							   0, NULL, NULL);
+	SendCommandToMetadataWorkersParams(command, CitusExtensionOwnerName(),
+									   0, NULL, NULL);
 }
 
 
@@ -146,24 +150,22 @@ TargetWorkerSetNodeList(TargetWorkerSet targetWorkerSet, LOCKMODE lockMode)
 
 
 /*
- * SendBareCommandListToWorkers sends a list of commands to a set of target
+ * SendBareCommandListToMetadataWorkers sends a list of commands to metadata
  * workers in serial. Commands are committed immediately: new connections are
  * always used and no transaction block is used (hence "bare"). The connections
  * are made as the extension owner to ensure write access to the Citus metadata
  * tables. Primarly useful for INDEX commands using CONCURRENTLY.
  */
 void
-SendBareCommandListToWorkers(TargetWorkerSet targetWorkerSet, List *commandList)
+SendBareCommandListToMetadataWorkers(List *commandList)
 {
+	TargetWorkerSet targetWorkerSet = WORKERS_WITH_METADATA;
 	List *workerNodeList = TargetWorkerSetNodeList(targetWorkerSet, ShareLock);
 	ListCell *workerNodeCell = NULL;
 	char *nodeUser = CitusExtensionOwnerName();
 	ListCell *commandCell = NULL;
 
-	if (targetWorkerSet == WORKERS_WITH_METADATA)
-	{
-		ErrorIfAnyMetadataNodeOutOfSync(workerNodeList);
-	}
+	ErrorIfAnyMetadataNodeOutOfSync(workerNodeList);
 
 	/* run commands serially */
 	foreach(workerNodeCell, workerNodeList)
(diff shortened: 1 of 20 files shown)


src/backend/distributed/transaction/worker_transaction.c
```diff
@@ -55,6 +58,11 @@ SendCommandToWorkersAsUser(TargetWorkerSet targetWorkerSet, const char *nodeUser
 	List *workerNodeList = TargetWorkerSetNodeList(targetWorkerSet, ShareLock);
 	ListCell *workerNodeCell = NULL;
 
+	if (targetWorkerSet == WORKERS_WITH_METADATA)
+	{
+		ErrorIfAnyMetadataNodeOutOfSync(workerNodeList);
```
marcocitus: `SendCommandToWorkersAsUser` doesn't seem like quite the right place to put this logic. I wonder whether we should be wrapping these functions into something like "SendCommandToWorkersWithMetadata" to have a clear separation of concerns.

onderkalaci: I'm considering another alternative. Put the check to `SendRemoteCommandParams() / SendRemoteCommand()`.

Moving a similar check to those functions requires reading the worker node from workernodeash on every command, which might be expensive to do. 

So, I'm considering to keep a global state of `bool HasWorkerWithOutOfSyncMetadata`, and in `SendRemoteCommandParams() / SendRemoteCommand()` we can do:
```C
if (HasWorkerWithOutOfSyncMetadata)
{
    WorkerNode *w = ...; /* only read the worker when necessary */
     if (!w-> metadataSynced)
        error
}
```

In this patch, keeping track of where to add `ErrorIfAnyMetadataNodeOutOfSync() ` is hard, in my proposal maintaining `HasWorkerWithOutOfSyncMetadata` is slightly hard, but at least we'd be ensuring any future command sending API is safe to use.

What do you think for that?

marcocitus: My problem is mainly that these are generic functions for sending commands and we're putting logic related to metadata syncing into them. I'm proposing we do some refactoring to instead call wrappers that are specific to metadata syncing.

src/backend/distributed/transaction/worker_transaction.c
```diff
@@ -244,6 +260,11 @@ SendCommandToWorkersParams(TargetWorkerSet targetWorkerSet, const char *command,
...
```
marcocitus: Is there a way to move this into SendCommandToWorkersWithMetadata?

src/backend/distributed/transaction/worker_transaction.c
```diff
@@ -139,20 +147,23 @@ TargetWorkerSetNodeList(TargetWorkerSet targetWorkerSet, LOCKMODE lockMode)
...
- * SendBareCommandListToWorkers sends a list of commands to a set of target
+ * SendBareCommandListToMetadataWorkers sends a list of commands to metadata
  * workers in serial. Commands are committed immediately: new connections are
  * always used and no transaction block is used (hence "bare"). The connections
  * are made as the extension owner to ensure write access to the Citus metadata
  * tables. Primarly useful for INDEX commands using CONCURRENTLY.
  */
 void
-SendBareCommandListToWorkers(TargetWorkerSet targetWorkerSet, List *commandList)
+SendBareCommandListToMetadataWorkers(List *commandList)
```
marcocitus: Minor: It would kind of make sense for metadata syncing functions to live in metadata_sync.c, but seems ok to keep them here for the moment to keep the change small.

src/backend/distributed/transaction/worker_transaction.c
```diff
@@ -185,14 +196,19 @@ SendBareCommandListToWorkers(TargetWorkerSet targetWorkerSet, List *commandList)
...
 {
+	TargetWorkerSet targetWorkerSet = WORKERS_WITH_METADATA;
 	List *workerNodeList = TargetWorkerSetNodeList(targetWorkerSet, ShareLock);
 	ListCell *workerNodeCell = NULL;
 	ListCell *commandCell = NULL;
 	int maxError = RESPONSE_OKAY;
 
+	if (targetWorkerSet == WORKERS_WITH_METADATA)
+	{
+		ErrorIfAnyMetadataNodeOutOfSync(workerNodeList);
```
marcocitus: Not sure this correct, given that this `if` statement is always true, and SendBareOptionalCommandListToAllWorkersAsUser is used for ALTER ENUM, which also applies to shards.

src/backend/distributed/transaction/worker_transaction.c
```diff
@@ -180,19 +196,37 @@ SendBareCommandListToWorkers(TargetWorkerSet targetWorkerSet, List *commandList)
...
+static int
+SendBareOptionalCommandListInternal(TargetWorkerSet targetWorkerSet, List *commandList,
+									const char *user)
 {
 	List *workerNodeList = TargetWorkerSetNodeList(targetWorkerSet, ShareLock);
 	ListCell *workerNodeCell = NULL;
 	ListCell *commandCell = NULL;
 	int maxError = RESPONSE_OKAY;
 
+	if (targetWorkerSet == WORKERS_WITH_METADATA)
```
marcocitus: I think this can be removed now?

src/backend/distributed/transaction/worker_transaction.c
onderkalaci: I'm still thinking about this, there is no caller of `SendBareOptionalCommandListInternal` with metadata workers. I don't want any future caller to skip this.

In that sense, I'm planning to have an `SendBareOptionalCommandListToMetadataWorkersAsUser ()` and do not have any callers for that function. Or, would it be more confusing than not having `ErrorIfAnyMetadataNodeOutOfSync ` check in `SendBareOptionalCommandList*()` API?

src/backend/distributed/transaction/worker_transaction.c
onderkalaci: Should I create `SendBareCommandListInternal()` for consistency with the other APIs where `SendBareCommandListToMetadataWorkers ` would only call `SendBareCommandListInternal` with `WORKERS_WITH_METADATA `?

src/backend/distributed/transaction/worker_transaction.c
marcocitus: Maybe just remove the TargetWorkerSet argument?

src/backend/distributed/transaction/worker_transaction.c
```diff
@@ -180,19 +196,37 @@ SendBareCommandListToWorkers(TargetWorkerSet targetWorkerSet, List *commandList)
...
+/*
+ * SendBareOptionalCommandListInternal sends a list of commands to a set of target
  * workers in serial. Commands are committed immediately: new connections are
  * always used and no transaction block is used (hence "bare").
  */
-int
-SendBareOptionalCommandListToWorkersAsUser(TargetWorkerSet targetWorkerSet,
-										   List *commandList, const char *user)
+static int
+SendBareOptionalCommandListInternal(TargetWorkerSet targetWorkerSet, List *commandList,
```
marcocitus: Maybe remove the TargetWorkerSet  parameter.

src/backend/distributed/transaction/worker_transaction.c
```diff
@@ -180,19 +196,37 @@ SendBareCommandListToWorkers(TargetWorkerSet targetWorkerSet, List *commandList)
...
+									const char *user)
 {
 	List *workerNodeList = TargetWorkerSetNodeList(targetWorkerSet, ShareLock);
```
marcocitus: maybe pass in ALL_WORKERS and remove the targetWorkerSet argument.

src/backend/distributed/transaction/worker_transaction.c
marcocitus: Could be nice, don't feel strongly about it.

src/test/regress/sql/distributed_functions.sql
```diff
@@ -449,6 +452,7 @@ TRUNCATE pg_dist_node;
 \c - - - :worker_2_port
 SET client_min_messages TO error; -- suppress cascading objects dropping
 UPDATE pg_dist_local_group SET groupid = 0;
+UPDATE pg_dist_node SET hasmetadata=false;
```
marcocitus: Do we need this? Pg_dist_node gets truncated a few lines down and I'd expect it to not matter on the worker.

onderkalaci: I think we shouldn't need. This test is very complicated, we first do stop_metadata_sync on the coordinator, and trying to drop schema on the workers.

So, we could do `SET citus.enable_ddl_propagation TO off;` instead of  `UPDATE pg_dist_node SET hasmetadata=false;`, or I can try to clean-up the tests a bit.

I'll fix it tomorrow

onderkalaci: This hack is mostly because we're not dropping the schema on the workers when `DROP SCHEMA` is executed, and MX errors if you try to drop schema from the workers. I pushed a commit to improve a bit

onderkalaci:
> We should probably figure out some kind of metadata cleansing function.

Marking #2326 as related

PR #3234, opened 2019-11-27
Add a script that fixes style related things

SaitTalhaNisanci:
It is kind of tedious that we need to make sure that every style check
passes with any change we make. A script is added, which does all the
things for us so that we dont have to run separate commands.

### DIFF
--- a/ci/fix_style.sh
+++ b/ci/fix_style.sh
@@ -1,0 +1,14 @@
+#!/bin/sh
+
+# fail if trying to reference a variable that is not set.
+set -u / set -o nounset
+# exit immediately if a command fails
+set -e
+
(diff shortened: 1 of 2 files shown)


SaitTalhaNisanci:
I didnt know we had reindent. Yeah I have added that there.

ci/fix_style.sh
```diff
@@ -0,0 +1,14 @@
...
+# exit immediately if a command fails
+set -e
+
+cidir="${0%/*}"
+cd ${cidir}/..
+
+ci/editorconfig.sh
+ci/remove_useless_declarations.sh
+
+citus_indent .
```
marcocitus: Maybe add back --quiet

PR #3237, opened 2019-11-27
RECORD: Add support for more expression types

serprex:
- OpExpr
- NullIfExpr
- MinMaxExpr
- CoalesceExpr
- CaseExpr

Also fix case where `ARRAY[(1,2), NULL]` was rejected

DESCRIPTION: support more expressions with `RECORD` result

### DIFF
--- a/src/test/regress/sql/row_types.sql
+++ b/src/test/regress/sql/row_types.sql
@@ -23,32 +23,61 @@ END;
 $$ language plpgsql;
 SELECT create_distributed_function('record_returner(int)');
 
+CREATE OR REPLACE FUNCTION identity_returner(x anyelement)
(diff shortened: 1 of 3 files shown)


src/test/regress/sql/row_types.sql
```diff
@@ -66,4 +95,5 @@ EXECUTE rec(1);
 EXECUTE rec(1);
 EXECUTE rec(1);
 
+SET client_min_messages TO error;
```
marcocitus: usually no point in doing a SET at the end of the file, since the session ends immediately after

serprex: Stops `DROP CASCADE` listing everything

PR #3239, opened 2019-11-28
Remove useless checks

onderkalaci:
We already support CTEs and Set Operations via recursive planning. Whenever we call `DeferErrorIfQueryNotSupported ()`, we error out anyway, so no callers rely on the checks. And, it is a little confusing because we support those.

And, this can be useful in #3161

### DIFF
--- a/src/backend/distributed/planner/multi_logical_planner.c
+++ b/src/backend/distributed/planner/multi_logical_planner.c
@@ -859,25 +859,10 @@ DeferErrorIfQueryNotSupported(Query *queryTree)
 					"clause containing the distribution column";
 	}
 
-	if (queryTree->setOperations)
-	{
-		preconditionsSatisfied = false;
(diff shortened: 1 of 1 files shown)


marcocitus:
Is there any case in which we don't use recursive planning (e.g. subquery pushdown setting?)

onderkalaci:
Good point:

- `GenerateSubplansForSubqueriesAndCTEs() -> RecursivelyPlanCTEs()` kicks in before any subquery pushdown logic (or any multi-shard query planning logic). So, that seems fine.
- For set operations, you're right. If we remove the check, the error message becomes weird:
```SQL
SET citus.subquery_pushdown TO ON;
(SELECT user_id FROM users_Table) UNION (SELECT user_id FROM users_table);
```

Anyway, I'll try to find something better for what I need in #3161, closing this for now

PR #3243, opened 2019-11-28
Remove distinction between SQL_TASK and ROUTER_TASK

marcocitus:
This distinction is no longer useful and the naming was more mysterious than it should have been.


src/include/distributed/multi_physical_planner.h
```diff
@@ -78,13 +78,12 @@ typedef enum
...
 {
 	TASK_TYPE_INVALID_FIRST = 0,
-	SQL_TASK = 1,
+	SELECT_TASK = 1,
 	MAP_TASK = 2,
 	MERGE_TASK = 3,
 	MAP_OUTPUT_FETCH_TASK = 4,
 	MERGE_FETCH_TASK = 5,
 	MODIFY_TASK = 6,
-	ROUTER_TASK = 7,
```
serprex: Decrement `DDL_TASK` & `VACUUM_ANALYZE_TASK`?

marcocitus: Removed the values.

PR #3246, opened 2019-11-29
add gitref to the output of citus_version

thanodnl:
DESCRIPTION: add gitref to the output of citus_version

During debugging of custom builds it is hard to know the exact version of the citus build you are using. This patch will add a human readable/understandable git reference to the build of citus which can be retrieved by calling `citus_version();`.

```sql
# SELECT citus_version();
┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│                                                                    citus_version                                                                    │
├─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Citus 9.1devel on x86_64-apple-darwin19.0.0, compiled by Apple clang version 11.0.0 (clang-1100.0.33.12), 64-bit gitref: v6.1.0-rc2-2107-g229115582 │
└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
(1 row)

Time: 36.899 ms
```

For the first commit in this PR (at time of writing the gitref is: `v6.1.0-rc2-2107-g229115582`.
This tells the developer the following things
 - closest tag on the tree `v6.1.0-rc2`
 - commits on top of the tag `2107`
 - short git sha of commit: `229115582`

If a build is from an exact tag only the tag name will be shown. Since we build our release versions from tags named eg. `v9.1.0` that will be in the build as the gitref.

During development it looks a bit strange at the moment that it seems to be a `v6.1.0-rc2` build. We can solve this to merge the last released version into the `master` branch after releasing. This normally is a simple merge without any conflicts as the release branch and `master` are pretty up-to-date.

An example of the output after merging `v9.1.0` into this branch:
```sql
# SELECT citus_version();
┌────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│                                                               citus_version                                                                │
├────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Citus 9.1.0 on x86_64-apple-darwin19.0.0, compiled by Apple clang version 11.0.0 (clang-1100.0.33.12), 64-bit gitref: v9.1.0-19-g8ac40740c │
└────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
(1 row)

Time: 312.196 ms
```

This would make it more intuitive to the reader of the gitref we compiled a newer version than `v9.1.0` which makes sense during development of `9.2.0`, however this would require a small change to our release runbook with the addition of merging the tagged release into master.

As a safe guard the gitref is _only_ added if the build environment has git installed and has the `.git` directory available next to the source. Meaning we don't break any build pipeline that doesn't have it, but we won't see the git ref in the final build.

Lastly a nice feature of `git describe`; it will add `...-dirty` to the gitref if there are uncommitted changes in the working directory during compilation. Giving even more information about the actual build you are looking at.

### DIFF
--- a/src/backend/distributed/Makefile
+++ b/src/backend/distributed/Makefile
@@ -19,6 +19,12 @@ SUBDIRS = . commands connection ddl deparser executor master metadata planner pr
 OBJS += \
 	$(patsubst $(citus_abs_srcdir)/%.c,%.o,$(foreach dir,$(SUBDIRS), $(sort $(wildcard $(citus_abs_srcdir)/$(dir)/*.c))))
 
+# make sure citus_version.o is always recompiled to reflect the correct gitref for every build
+# idea came from https://stackoverflow.com/a/3304967
+utils/citus_version.o: FORCE
+FORCE:
+.PHONY: FORCE
+
 # be explicit about the default target
 all:
 
--- a/src/backend/distributed/Makefile
+++ b/src/backend/distributed/Makefile
@@ -19,12 +19,6 @@ SUBDIRS = . commands connection ddl deparser executor master metadata planner pr
 OBJS += \
 	$(patsubst $(citus_abs_srcdir)/%.c,%.o,$(foreach dir,$(SUBDIRS), $(sort $(wildcard $(citus_abs_srcdir)/$(dir)/*.c))))
 
-# make sure citus_version.o is always recompiled to reflect the correct gitref for every build
-# idea came from https://stackoverflow.com/a/3304967
-utils/citus_version.o: FORCE
-FORCE:
-.PHONY: FORCE
-
 # be explicit about the default target
 all:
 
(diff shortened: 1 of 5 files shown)


src/backend/distributed/Makefile
```diff
@@ -19,6 +19,12 @@ SUBDIRS = . commands connection ddl deparser executor master metadata planner pr
 OBJS += \
 	$(patsubst $(citus_abs_srcdir)/%.c,%.o,$(foreach dir,$(SUBDIRS), $(sort $(wildcard $(citus_abs_srcdir)/$(dir)/*.c))))
 
+# make sure citus_version.o is always recompiled to reflect the correct gitref for every build
+# idea came from https://stackoverflow.com/a/3304967
+utils/citus_version.o: FORCE
```
marcocitus: Small downside is that `make install` now rebuilds citus_version.o and citus.so twice. I tend to do `make -j 8 && make install`, which rebuilds it 3 times.

What if we add a rule like:
`utils/citus_version.o: $(OBJS)`

That way, we rebuild it whenever a .o file changes.

I believe it needs to be below `include $(citus_top_builddir)/Makefile.global`.

(Bonus points for removing the circular dependency, though `make` does not seem to mind to much)

hanefi:
> During development it looks a bit strange at the moment that it seems to be a `v6.1.0-rc2` build. We can solve this to merge the last released version into the `master` branch after releasing. This normally is a simple merge without any conflicts as the release branch and `master` are pretty up-to-date.

This has only one down-side I can think of: The handling version on master gets tricky. As you can see on your examples, the version changed from `Citus 9.1devel` to `Citus 9.1.0`.

What if we merge `release-9.x` branch into master (`9.xdevel` -> `9.x.0`), immediately followed by a commit that fixes the version (`9.x.0 -> 9.x+1devel`). Does that make sense?

thanodnl:
I think that makes a lot of sense, however I have never done a release for citus, but for pg_auto_failover I made sure to always merge releases back into master. This has two benefits;
 1.  The changelog can be made on the release branch and will end up on master afterwards
 2. Any patches made during releases end up in master. We solve this differently by solving issues on master and cherry-picking them into the release branch.

The second one requires a change in how we manage our releases. We can omit the latter and still get the benefit of the changelog and the tags ending back in master. The merge of a release in master is mostly a clean merge in my experience, as it has the same units of changes.

PR #3248, opened 2019-11-29
Bump repo version to 9.2devel

marcocitus:
Updating configure, citus.control, tests, and upgrade script.


metdos:
Are ditching runstatedir on purpose?

marcocitus:
Configure is auto-generated by autoconf. Whether or not you get a runstatedir argument depends on your version of autoconf, though both claim to be 2.69.

PR #3249, opened 2019-11-30
aggregate_support test: test DISTINCT, ORDER BY, FILTER, & no intermediate results

serprex:
Previously,
- we'd push down ORDER BY, but this doesn't order intermediate results between workers
- we'd keep FILTER on master aggregate, which would raise an error about unexpected cstrings

### DIFF
--- a/src/test/regress/expected/aggregate_support.out
+++ b/src/test/regress/expected/aggregate_support.out
@@ -58,8 +58,7 @@ select create_distributed_table('aggdata', 'id');
 (1 row)
 
 insert into aggdata (id, key, val, valf) values (1, 1, 2, 11.2), (2, 1, NULL, 2.1), (3, 2, 2, 3.22), (4, 2, 3, 4.23), (5, 2, 5, 5.25), (6, 3, 4, 63.4), (7, 5, NULL, 75), (8, 6, NULL, NULL), (9, 6, NULL, 96), (10, 7, 8, 1078), (11, 9, 0, 1.19);
-select key, sum2(val), sum2_strict(val), stddev(valf)
-from aggdata group by key order by key;
+select key, sum2(val), sum2_strict(val), stddev(valf) from aggdata group by key order by key;
  key | sum2 | sum2_strict |      stddev      
 -----+------+-------------+------------------
    1 |      |           4 | 6.43467170879758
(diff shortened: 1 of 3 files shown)


src/test/regress/expected/aggregate_support.out
```diff
@@ -71,6 +70,28 @@ from aggdata group by key order by key;
...
+               ^
+-- DISTINCT unsupported
+select key, sum2(distinct val), sum2_strict(distinct val) aggdata group by key order by key;
+ERROR:  column "key" does not exist
+LINE 1: select key, sum2(distinct val), sum2_strict(distinct val) ag...
+               ^
+-- ORDER BY unsupported
+select key, sum2(val order by valf), sum2_strict(val order by valf) aggdata group by key order by key;
+ERROR:  column "key" does not exist
+LINE 1: select key, sum2(val order by valf), sum2_strict(val order b...
```
marcocitus: These tests are missing FROM

serprex: fixed

src/test/regress/expected/aggregate_support.out
```diff
@@ -123,8 +148,7 @@ create aggregate sumstring(text) (
 );
 select sumstring(valf::text order by id)
 from aggdata where valf is not null;
-ERROR:  function "aggregate_support.sumstring(text)" does not exist
-CONTEXT:  while executing command on localhost:57637
+ERROR:  unsupported aggregate function sumstring
```
serprex: these tests should be updated to something that doesn't need `order by`

src/test/regress/expected/aggregate_support.out
```diff
@@ -71,6 +70,49 @@ from aggdata group by key order by key;
...
+   5 |      |            
+   6 |      |            
+   7 |      |            
+   9 |    0 |           0
+(7 rows)
+
+-- DISTINCT unsupported, unless grouped by partition key
+select key, sum2(distinct val), sum2_strict(distinct val) from aggdata group by key order by key;
+ERROR:  cannot compute aggregate (distinct)
+DETAIL:  table partitioning is unsuitable for aggregate (distinct)
```
marcocitus: Error messages could use improvement.

PR #3257, opened 2019-12-03
Detect SQL UDF Calls.

### LINKED ISSUE #3256: "LANGUAGE SQL" UDF calls aren't detected as function calls
In `CitusExecutorRun()` we use the `dest->mydest == DestSPI` to check whether we are in a function call or not. This works fine for PLs, but for SQL functions `dest->mydest` is `DestSQLFunction`, so we don't detect them as UDF calls.

We use `FunctionCallLevel > 0` as part of `IsMultiStatementTransaction()`, and SQL UDFs can have multiple statements, so we should add this check.

pykello:
Fixes #3256 

We treat statements inside a function as a multistatement transaction, so we start 2PC and roll back multi-statement UDFs properly. Previously we checked `dest == DestSPI` in the executor to check if we are in a function call or not. This is not enough for SQL UDFs, since they don't use the SPI executor.

SQL UDFs start postgres executors recursively. This PR uses this fact to detect SQL UDF calls.

This is not enough for UDFs with utility commands, because postgres doesn't use the executor for utility commands. So this only fixes the case for UDFs which are sequence of DML and read queries only.

### DIFF
--- a/src/backend/distributed/executor/multi_executor.c
+++ b/src/backend/distributed/executor/multi_executor.c
@@ -56,6 +56,13 @@ bool WritableStandbyCoordinator = false;
 /* sort the returning to get consistent outputs, used only for testing */
 bool SortReturning = false;
 
+/*
+ * How many nested executors have we started? This can happen for SQL
+ * UDF calls. The outer query starts an executor, then postgres opens
+ * another executor to run the SQL UDF.
+ */
+int ExecutorLevel = 0;
+
 
 /* local function forward declarations */
 static Relation StubRelation(TupleDesc tupleDescriptor);
@@ -114,18 +121,9 @@ CitusExecutorRun(QueryDesc *queryDesc,
 				 ScanDirection direction, uint64 count, bool execute_once)
 {
 	DestReceiver *dest = queryDesc->dest;
-	int originalLevel = FunctionCallLevel;
+	int originalLevel = ExecutorLevel;
 
-	if (dest->mydest == DestSPI)
-	{
-		/*
-		 * If the query runs via SPI, we assume we're in a function call
-		 * and we should treat statements as part of a bigger transaction.
-		 * We reset this counter to 0 in the abort handler.
-		 */
-		FunctionCallLevel++;
-	}
-
+	ExecutorLevel++;
 	if (CitusHasBeenLoaded())
 	{
 		if (IsLocalReferenceTableJoinPlan(queryDesc->plannedstmt) &&
@@ -139,7 +137,8 @@ CitusExecutorRun(QueryDesc *queryDesc,
 			 */
 			ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
 							errmsg("cannot join local tables and reference tables in "
-								   "a transaction block")));
+								   "a transaction block, udf block, or distributed "
+								   "CTE subquery")));
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


src/backend/distributed/executor/multi_executor.c
```diff
@@ -131,16 +152,16 @@ CitusExecutorRun(QueryDesc *queryDesc,
...
 {
 	DestReceiver *dest = queryDesc->dest;
-	int originalLevel = FunctionCallLevel;
+	int originalLevel = SPILevel;
 
 	if (dest->mydest == DestSPI)
 	{
 		/*
-		 * If the query runs via SPI, we assume we're in a function call
+		 * If the query runs via SPI, we assume we're in a function call.
```
serprex: This period doesn't make sense when the following line begins with 'and', comma?

src/backend/distributed/executor/multi_executor.c
```diff
@@ -182,7 +203,7 @@ CitusExecutorRun(QueryDesc *queryDesc,
 		 * the value because exceptions might cause us to go back a few
 		 * levels at once.
 		 */
```
marcocitus: Does the same apply to ExecutorLevel?

src/backend/distributed/executor/multi_executor.c
```diff
@@ -131,16 +152,16 @@ CitusExecutorRun(QueryDesc *queryDesc,
...
 	if (dest->mydest == DestSPI)
 	{
 		/*
-		 * If the query runs via SPI, we assume we're in a function call
+		 * If the query runs via SPI, we assume we're in a function call.
 		 * and we should treat statements as part of a bigger transaction.
 		 * We reset this counter to 0 in the abort handler.
 		 */
-		FunctionCallLevel++;
+		SPILevel++;
```
marcocitus: is SPILevel still necessary if we have ExecutorLevel? I assume ExecutorLevel gets raised for SPI as well?

src/backend/distributed/transaction/transaction_management.c
```diff
@@ -691,3 +693,17 @@ IsMultiStatementTransaction(void)
...
+ * ExecutingUDF returns true if we are executing a function call.
+ */
+static bool
+ExecutingUDF(void)
+{
+	/*
+	 * SQL UDFs start another executor, and PL UDFs use SPI to
+	 * execute statements.
+	 */
+	return ExecutorLevel > 1 || SPILevel > 0;
```
marcocitus: This probably also triggers for other things other than UDFs (e.g. recursively planned CTEs).

src/backend/distributed/executor/multi_executor.c
pykello: Removed SPILevel.

src/backend/distributed/transaction/transaction_management.c
pykello: Renamed the function to `MaybeExecutingUDF`, I guess in other cases we also create remote transaction blocks too, so false positives here should be safe. I need to do some testing before  merging though.

src/backend/distributed/executor/multi_executor.c
pykello: Currently I increment/decrement in ExecutorStart/End functions. Given your point, maybe I should do it ExecutorRun. I'll update the PR with this after finding a failing test.

PR #3262, opened 2019-12-03
Fix error in DML with NULL expression in where clause

### LINKED ISSUE #3236: planner bug: could not find valid entry for shard 0
## versions
PostgreSQL 12.1 ( postgres/postgres@bf3cef24a319666 )
Citus 9.0.1 ( citusdata/citus@a853a824a5bc89 ) 

## schema
``` sql
CREATE TABLE stores_purchase (
    id integer NOT NULL,
    quantity integer NOT NULL,
    product_id integer NOT NULL,
    store_id integer NOT NULL
);
SELECT create_distributed_table('stores_purchase', 'store_id')
```

No data is needed in the table. 

### query 
```
postgres=# explain delete from stores_purchase where null > jsonb_build_array();
ERROR:  could not find valid entry for shard 0
```

No traffic makes it to the workers.

marcocitus:
DESCRIPTION: Fixes a bug that caused some noop DML statements to fail

We generated a task without an anchor shard ID for the query in #3236, which later causes a failure in the executor when looking up placements. We should always have an empty task list for DML that prunes to 0 shards, since it cannot return a result.

Fixes #3236

### DIFF
--- a/src/backend/distributed/planner/multi_router_planner.c
+++ b/src/backend/distributed/planner/multi_router_planner.c
@@ -1715,6 +1715,11 @@ RouterJob(Query *originalQuery, PlannerRestrictionContext *plannerRestrictionCon
 												 MODIFY_TASK,
 												 requiresMasterEvaluation);
 	}
+	else if (shardId == INVALID_SHARD_ID)
+	{
+		/* modification that prunes to 0 shards */
+		job->taskList = NIL;
+	}
 	else
 	{
 		job->taskList = SingleShardModifyTaskList(originalQuery, job->jobId,
@@ -2068,7 +2073,6 @@ PlanRouterQuery(Query *originalQuery,
 		}
 
 		Assert(UpdateOrDeleteQuery(originalQuery));
-
 		planningError = ModifyQuerySupported(originalQuery, originalQuery,
 											 isMultiShardQuery,
 											 plannerRestrictionContext);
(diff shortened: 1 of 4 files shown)


src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1705,6 +1705,11 @@ RouterJob(Query *originalQuery, PlannerRestrictionContext *plannerRestrictionCon
 												 MODIFY_TASK,
 												 requiresMasterEvaluation);
 	}
+	else if (shardId == INVALID_SHARD_ID)
+	{
+		/* modification that prunes to 0 shards */
```
onderkalaci: Given that the `if/else` block is not very simple, and can change in  future,  maybe add `Assert (originalQuery->commandType != CMD_SELECT)`,  because in `SELECT` case it is where `intermediate result only` router queries.

src/test/regress/expected/multi_modifications.out
```diff
@@ -874,6 +874,10 @@ INSERT INTO raw_table VALUES (2, 400);
 INSERT INTO raw_table VALUES (2, 500);
 INSERT INTO summary_table VALUES (1);
 INSERT INTO summary_table VALUES (2);
+-- test noop deletes
+DELETE FROM summary_table WHERE false;
```
onderkalaci: could be nice to have one test for update,  in case we one  day have different ways of  planning  it

PR #3263, opened 2019-12-03
Fix strange errors in DML with unreachable sublinks

### LINKED ISSUE #3235: planner bug: WindowFunc found in non-WindowAgg plan node
## versions
PostgreSQL 12.1 ( postgres/postgres@bf3cef24a319666 )
Citus 9.0.1 ( citusdata/citus@a853a824a5bc89 ) 

## schema
``` sql
CREATE TABLE stores_purchase (
    id integer NOT NULL,
    quantity integer NOT NULL,
    product_id integer NOT NULL,
    store_id integer NOT NULL
);
SELECT create_distributed_table('stores_purchase', 'store_id')
```

## broken queries
No data is needed in the table. In neither case does any traffic make it to the workers.

### query 1
``` sql
explain delete from stores_purchase
where
case when
  EXISTS (
    select
      (select action_statement from information_schema.triggers)
    from
      pg_catalog.pg_statio_sys_sequences
  )
then null
end = null
;
```
#### result 
`ERROR:  WindowFunc found in non-WindowAgg plan node`
from https://github.com/postgres/postgres/blob/bf3cef24a31966686dfaae3085ce7c545c0019a4/src/backend/executor/execExpr.c#L862

### query 2
```sql 
explain delete from stores_purchase
where
case when
  EXISTS (
    select 1
    from pg_catalog.pg_statio_sys_sequences
  )
then null
end = null
;
```
#### result 
`ERROR:  could not find valid entry for shard 0`

marcocitus:
DESCRIPTION: Fix strange errors in DML with unreachable sublinks

We were checking the relations in ModifyQuerySupported using the rewritten query, but the rewritten query omits unreachable sublinks. Since we eventually use the original query, this causes strange errors when the unreachable sublinks are not supported by other parts of the code.

Fixes #3235

### DIFF
--- a/src/backend/distributed/planner/multi_router_planner.c
+++ b/src/backend/distributed/planner/multi_router_planner.c
@@ -640,7 +640,7 @@ ModifyQuerySupported(Query *queryTree, Query *originalQuery, bool multiShardQuer
 	}
 
 	/* extract range table entries */
-	ExtractRangeTableEntryWalker((Node *) queryTree, &rangeTableList);
+	ExtractRangeTableEntryWalker((Node *) originalQuery, &rangeTableList);
 
 	foreach(rangeTableCell, rangeTableList)
 	{
--- a/src/test/regress/expected/multi_modifications.out
+++ b/src/test/regress/expected/multi_modifications.out
@@ -1285,6 +1285,18 @@ HINT:  Try rewriting your queries with 'INSERT INTO ... SELECT' syntax.
 INSERT INTO summary_table (id) VALUES ((SELECT id FROM reference_summary_table));
 ERROR:  subqueries are not supported within INSERT queries
 HINT:  Try rewriting your queries with 'INSERT INTO ... SELECT' syntax.
+-- subqueries that would be eliminated by = null clauses
+DELETE FROM summary_table WHERE (
+    SELECT 1 FROM pg_catalog.pg_statio_sys_sequences
+) = null;
+DELETE FROM summary_table WHERE (
+    SELECT (select action_statement from information_schema.triggers)
+    FROM pg_catalog.pg_statio_sys_sequences
+) = null;
+ERROR:  relation pg_namespace is not distributed
+DELETE FROM summary_table WHERE id < (
+    SELECT 0 FROM pg_dist_node
+);
 DROP TABLE raw_table;
 DROP TABLE summary_table;
 DROP TABLE reference_raw_table;
--- a/src/test/regress/sql/multi_modifications.sql
+++ b/src/test/regress/sql/multi_modifications.sql
@@ -858,6 +858,19 @@ INSERT INTO summary_table (id) VALUES (5), ((SELECT id FROM summary_table));
 INSERT INTO reference_summary_table (id) VALUES ((SELECT id FROM summary_table));
 INSERT INTO summary_table (id) VALUES ((SELECT id FROM reference_summary_table));
 
+-- subqueries that would be eliminated by = null clauses
+DELETE FROM summary_table WHERE (
+    SELECT 1 FROM pg_catalog.pg_statio_sys_sequences
+) = null;
+DELETE FROM summary_table WHERE (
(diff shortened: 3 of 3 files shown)


src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -640,7 +640,7 @@ ModifyQuerySupported(Query *queryTree, Query *originalQuery, bool multiShardQuer
 	}
 
 	/* extract range table entries */
-	ExtractRangeTableEntryWalker((Node *) queryTree, &rangeTableList);
+	ExtractRangeTableEntryWalker((Node *) originalQuery, &rangeTableList);
```
onderkalaci: I think this change makes sense in general. But, by doing that, we're now treating pulled up subqueries a little different.

Before the patch:
```SQL

DELETE FROM stores_purchase WHERE  store_id > (SELECT logicalrelid  FROM pg_dist_partition);
ERROR:  relation pg_dist_partition is not distributed
```
After the patch:
```SQL
 DELETE FROM stores_purchase WHERE  store_id > (SELECT logicalrelid  FROM pg_dist_partition);
DEBUG:  generating subplan 3_1 for subquery SELECT logicalrelid FROM pg_dist_partition
DEBUG:  Plan 3 query after replacing subqueries and CTEs: DELETE FROM public.stores_purchase WHERE ((store_id)::oid OPERATOR(pg_catalog.>) ((SELECT intermediate_result.logicalrelid FROM read_intermediate_result('3_1'::text, 'binary'::citus_copy_format) intermediate_result(logicalrelid regclass)))::oid)
DELETE 0
```

Do you see any case where we'd regress? I think no, but wanted to ask in case you can think of any edges?

marcocitus: That error happened in a pretty awkward place because we allowed the local table to by-pass our checks:
```
#0  errstart (elevel=elevel@entry=20, filename=filename@entry=0x7f84f1887f88 "metadata/metadata_cache.c", lineno=lineno@entry=782, 
    funcname=funcname@entry=0x7f84f1889150 <__func__.29196> "DistributedTableCacheEntry", domain=domain@entry=0x0) at elog.c:251
#1  0x00007f84f183680b in DistributedTableCacheEntry (distributedRelationId=distributedRelationId@entry=16799) at metadata/metadata_cache.c:782
#2  0x00007f84f1852a55 in TargetShardIntervalsForRestrictInfo (restrictionContext=<optimized out>, multiShardQuery=multiShardQuery@entry=0x7ffeb3c7a96f, 
    partitionValueConst=partitionValueConst@entry=0x7ffeb3c7a9d0) at planner/multi_router_planner.c:2294
#3  0x00007f84f18533a8 in PlanRouterQuery (originalQuery=originalQuery@entry=0xeeb968, plannerRestrictionContext=plannerRestrictionContext@entry=0xf60568, 
    placementList=placementList@entry=0x7ffeb3c7a9f0, anchorShardId=anchorShardId@entry=0x7ffeb3c7a9f8, relationShardList=relationShardList@entry=0x7ffeb3c7a9e8, 
    prunedShardIntervalListList=prunedShardIntervalListList@entry=0x7ffeb3c7a9e0, replacePrunedQueryWithDummy=true, multiShardModifyQuery=0x7ffeb3c7a9df, 
    partitionValueConst=0x7ffeb3c7a9d0) at planner/multi_router_planner.c:2039
#4  0x00007f84f1853754 in RouterJob (originalQuery=originalQuery@entry=0xeeb968, plannerRestrictionContext=plannerRestrictionContext@entry=0xf60568, 
    planningError=planningError@entry=0x1054b20) at planner/multi_router_planner.c:1649
#5  0x00007f84f1853bff in CreateModifyPlan (originalQuery=originalQuery@entry=0xeeb968, query=query@entry=0xeeba80, 
```

In that sense this is another bug that is fixed. 

I don't think we would regress since the rewritten query contains a subset of RTEs of the original query (since views are already expanded in the original query).

PR #3267, opened 2019-12-04
(WIP) INSERT...SELECT with repartitioning (not ready for review)

pykello:
DESCRIPTION: INSERT...SELECT with repartitioning

Copy of #3078, rebased + compile with pg12. Keeping  #3078 in case I miss anything.

We currently have two methods of executing an INSERT...SELECT: 
- in parallel between co-located shards, used when the distribution column value is preserved between source and target table
- via the coordinator using COPY, used in all other cases

The first method scales horizontally, but the second method can get bottlenecked on single core performance on the coordinator, and may cause the coordinator to run out of disk space.

This PR adds a third method:
- distributed repartitioning of the select result, used in cases where the SELECT is a distributed query that does not require a merge on the coordinator

The INSERT..SELECT runs in 3 phases:
1. Run SELECT tasks wrapped in worker_predistribute_query_result to partition query results into fragments which correspond to the hash ranges of the shards in the destination table
2. Fetch fragments to the corresponding destination shard using fetch_intermediate_result using one call per node pair
3. Run INSERT INTO .. SELECT .. FROM read_intermediate_result(..) into the shards of the destination table, where read_intermediate_result reads the concatenation of fragments that correspond to the hash range of the shard

In each phase, the tasks are executed by the adaptive executor. Empty tasks are pruned away during execution.

### DIFF
--- a/src/backend/distributed/executor/distributed_intermediate_results.c
+++ b/src/backend/distributed/executor/distributed_intermediate_results.c
@@ -1,0 +1,818 @@
+/*-------------------------------------------------------------------------
+ *
+ * distributed_intermediate_results.c
+ *   Functions for reading and writing distributed intermediate results.
+ *
+ * Copyright (c) 2017, Citus Data, Inc.
+ *
+ *-------------------------------------------------------------------------
+ */
+#include <sys/stat.h>
+#include <unistd.h>
+
+#include "postgres.h"
+#include "funcapi.h"
+#include "libpq-fe.h"
+#include "miscadmin.h"
+#include "pgstat.h"
+#include "port.h"
+
+#include "access/hash.h"
+#include "access/nbtree.h"
+#include "access/tupdesc.h"
+#include "catalog/pg_am.h"
+#include "catalog/pg_enum.h"
+#include "catalog/pg_type.h"
+#include "commands/copy.h"
+#include "distributed/citus_ruleutils.h"
+#include "distributed/colocation_utils.h"
+#include "distributed/commands/multi_copy.h"
+#include "distributed/connection_management.h"
+#include "distributed/intermediate_results.h"
+#include "distributed/master_metadata_utility.h"
+#include "distributed/metadata_cache.h"
+#include "distributed/multi_executor.h"
+#include "distributed/recursive_planning.h"
+#include "distributed/redistribution.h"
+#include "distributed/remote_commands.h"
+#include "distributed/sharding.h"
+#include "distributed/transmit.h"
+#include "distributed/transaction_identifier.h"
+#include "distributed/tuplestore.h"
+#include "distributed/worker_protocol.h"
+#include "nodes/makefuncs.h"
+#include "nodes/parsenodes.h"
+#include "nodes/primnodes.h"
+#include "storage/fd.h"
+#include "tcop/pquery.h"
+#include "tcop/tcopprot.h"
+#include "utils/builtins.h"
+#include "utils/lsyscache.h"
+#include "utils/memutils.h"
+#include "utils/portal.h"
+#include "utils/syscache.h"
+
+
+/*
+ * NodePair contains the source and destination node in a NodeToNodeFragmentsTransfer.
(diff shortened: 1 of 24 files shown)


marcocitus:
It might make sense to turn the commits in #3078 into separate PRs given its size.

This PR represents one of the commits:
https://github.com/citusdata/citus/pull/3113

src/backend/distributed/planner/recursive_planning.c
```diff
@@ -1541,6 +1542,75 @@ ShouldTransformRTE(RangeTblEntry *rangeTableEntry)
...
+	resultIdConst->constisnull = false;
+	resultIdConst->location = -1;
+
+	return BuildReadIntermediateResultsQuery(targetEntryList, columnAliasList,
+											 resultIdConst);
+}
+
+
+/*
+ * BuildSubPlanResultQuery returns a query of the form:
```
serprex: Comment needs to be updated

src/backend/distributed/utils/colocation_utils.c
```diff
@@ -1036,3 +1038,111 @@ DeleteColocationGroup(uint32 colocationId)
...
+	PartitioningScheme *partitioningScheme = (PartitioningScheme *) palloc0(
+		sizeof(PartitioningScheme));
+
+	LoadPartitioningSchemeForRelationId(partitioningScheme, relationId);
+
+	return partitioningScheme;
+}
+
+
+static void
```
serprex: Lacks a top level comment

src/backend/distributed/utils/colocation_utils.c
```diff
@@ -1036,3 +1038,111 @@ DeleteColocationGroup(uint32 colocationId)
...
+	for (shardIndex = 0; shardIndex < shardCount; shardIndex++)
+	{
+		GroupShardPlacement *placementArray =
+			distTableCacheEntry->arrayOfPlacementArrays[shardIndex];
+		int placementCount =
+			distTableCacheEntry->arrayOfPlacementArrayLengths[shardIndex];
+		int placementIndex = 0;
+
+		distributionScheme->groupIds[shardIndex] =
+			(int *) palloc0(sizeof(int) * (placementCount + 1));
```
serprex: Why `+ 1`?

src/test/regress/sql/insert_select_repartition.sql
```diff
@@ -0,0 +1,15 @@
...
+CREATE SCHEMA insert_select_repartition;
+SET search_path TO 'insert_select_repartition';
+
+SET citus.shard_count TO 8;
+CREATE TABLE src_table (a int);
+SELECT create_distributed_table('src_table', 'a');
+INSERT INTO src_table SELECT i FROM generate_series(1, 100) i;
+
+SET citus.shard_count TO 3;
+CREATE TABLE trg_table (a int);
```
serprex: `trg` as opposed to `dst` or `tgt` is misleading since `trg` could be read as `trigger` as much as `target`

src/backend/distributed/executor/distributed_intermediate_results.c
```diff
@@ -0,0 +1,714 @@
...
+/*
+ * ExecutePredistributionTasks executes a list of tasks that were generated using
+ * WrapTasksForPredistribution.
+ */
+static PredistributionStats *
+ExecutePredistributionTasks(List *taskList, DistributionScheme *targetDistribution)
+{
+	TupleDesc resultDescriptor = NULL;
+	Tuplestorestate *resultStore = NULL;
+	int resultColumnCount = 5;
```
serprex: 4

src/backend/distributed/executor/distributed_intermediate_results.c
```diff
@@ -0,0 +1,714 @@
+/*-------------------------------------------------------------------------
+ *
+ * distributed_intermediate_results.c
+ *   Functions for reading and writing distributed intermediate results.
+ *
+ * Copyright (c) 2017, Citus Data, Inc.
```
serprex: Omit years

PR #3280, opened 2019-12-10
Fix the way we check for local/reference table joins in the executor

### LINKED ISSUE #3279: Cannot drop a column from a reference table on the coordinator
The checks for reference + local table joins might be too aggressive as they can show up in the drop trigger.

```sql
SELECT master_add_node('localhost', 5432, groupid := 0);
ALTER TABLE ref DROP COLUMN z;
ERROR:  cannot join local tables and reference tables in a transaction block
CONTEXT:  while executing command on 127.0.0.1:5432
SQL statement "SELECT master_unmark_object_distributed(v_obj.classid, v_obj.objid, v_obj.objsubid)"
PL/pgSQL function citus_drop_trigger() line 28 at PERFORM
```

```sql
SELECT worker_apply_shard_ddl_command (102040, 'public', 'ALTER TABLE ref DROP COLUMN z;');
ERROR:  cannot join local tables and reference tables in a transaction block
CONTEXT:  SQL statement "SELECT master_unmark_object_distributed(v_obj.classid, v_obj.objid, v_obj.objsubid)"
PL/pgSQL function citus_drop_trigger() line 28 at PERFORM
```

pykello:
Fixes #3279 

Previously we used `plannedStmt->releationOids` which is the set of relations query depends on to check for the relations "involved" in the query. But this is not correct. For example, citus plans query of the following form in the `citus_drop_trigger()` to have `relationOids = [pg_class, reference_table]`, so we assumed it is a join between reference table and local table.

```sql
SELECT master_unmark_object_distributed(v_obj.classid, v_obj.objid, v_obj.objsubid)
```

This PR changes the behaviour to use the range table entries instead.

### DIFF
--- a/src/backend/distributed/executor/multi_executor.c
+++ b/src/backend/distributed/executor/multi_executor.c
@@ -580,7 +580,7 @@ IsLocalReferenceTableJoinPlan(PlannedStmt *plan)
 {
 	bool hasReferenceTable = false;
 	bool hasLocalTable = false;
-	ListCell *oidCell = NULL;
+	ListCell *rangeTableCell = NULL;
 	bool hasReferenceTableReplica = false;
 
 	/*
@@ -617,12 +617,24 @@ IsLocalReferenceTableJoinPlan(PlannedStmt *plan)
 		return false;
 	}
 
-	foreach(oidCell, plan->relationOids)
+	/*
+	 * plan->rtable contains the flattened RTE lists of the plan tree, which
+	 * includes rtes in subqueries, CTEs, ...
+	 *
+	 * It doesn't contain optimized away table accesses (due to join optimization),
+	 * which is fine for our purpose.
+	 */
+	foreach(rangeTableCell, plan->rtable)
 	{
-		Oid relationId = lfirst_oid(oidCell);
+		RangeTblEntry *rangeTableEntry = (RangeTblEntry *) lfirst(rangeTableCell);
 		bool onlySearchPath = false;
 
-		if (RelationIsAKnownShard(relationId, onlySearchPath))
+		if (rangeTableEntry->rtekind != RTE_RELATION)
+		{
+			continue;
+		}
+
+		if (RelationIsAKnownShard(rangeTableEntry->relid, onlySearchPath))
 		{
 			/*
 			 * We don't allow joining non-reference distributed tables, so we
--- a/src/backend/distributed/executor/multi_executor.c
+++ b/src/backend/distributed/executor/multi_executor.c
@@ -14,6 +14,7 @@
 
 #include "access/xact.h"
 #include "catalog/dependency.h"
+#include "catalog/pg_class.h"
 #include "catalog/namespace.h"
 #include "distributed/citus_custom_scan.h"
 #include "distributed/commands/multi_copy.h"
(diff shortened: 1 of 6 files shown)


src/backend/distributed/executor/multi_executor.c
```diff
@@ -617,12 +617,22 @@ IsLocalReferenceTableJoinPlan(PlannedStmt *plan)
 		return false;
 	}
 
-	foreach(oidCell, plan->relationOids)
+	foreach(rangeTableCell, plan->rtable)
```
marcocitus: Does this contain RTEs for the entire query? (incl. CTEs, subqueries)

src/backend/distributed/executor/multi_executor.c
```diff
@@ -617,12 +617,22 @@ IsLocalReferenceTableJoinPlan(PlannedStmt *plan)
...
 	{
-		Oid relationId = lfirst_oid(oidCell);
+		RangeTblEntry *rangeTableEntry = (RangeTblEntry *) lfirst(rangeTableCell);
 		bool onlySearchPath = false;
 
-		if (RelationIsAKnownShard(relationId, onlySearchPath))
+		if (rangeTableEntry->rtekind == RTE_FUNCTION)
+		{
+			return false;
```
marcocitus: why return false here? What if we a join between a local table, a reference table and a function?

marcocitus:
if I try without any works (just coordinator), I get:
```sql
postgres=# ALTER TABLE ref ADD COLUMN z int;
WARNING:  could not find any shard placements for shardId 102040
... stuck forever
```

marcocitus:
> if I try without any works (just coordinator), I get:

that actually seems to be a separate bug unrelated to having the coordinator in the metadata

src/backend/distributed/executor/multi_executor.c
pykello: To match distributed_planner.c's `IsLocalReferenceTableJoin()`. Joins between local, reference, and functions will error out in planner. We can remove that restriction in both planner and executor, but maybe in another PR.

Removing the check here, although I think it won't change any behaviour.

src/backend/distributed/executor/multi_executor.c
pykello: Yes, standard_planner stores the flattened rte lists in plannedStmt->rtable. `set_plan_references` concatenates all these lists in `glob->finalrtable`, and then standard_planner does `plannedStmt->rtable = glob->finalrtable`.

Added a comment.

pykello:
> > if I try without any works (just coordinator), I get:
> 
> that actually seems to be a separate bug unrelated to having the coordinator in the metadata

I can look into this tomorrow.

src/backend/distributed/executor/multi_executor.c
```diff
@@ -617,12 +617,24 @@ IsLocalReferenceTableJoinPlan(PlannedStmt *plan)
...
+	foreach(rangeTableCell, plan->rtable)
 	{
-		Oid relationId = lfirst_oid(oidCell);
+		RangeTblEntry *rangeTableEntry = (RangeTblEntry *) lfirst(rangeTableCell);
 		bool onlySearchPath = false;
 
-		if (RelationIsAKnownShard(relationId, onlySearchPath))
+		if (rangeTableEntry->rtekind != RTE_RELATION)
+		{
+			continue;
```
marcocitus: Since views are RTE_RELATION they are also treated as local tables.

```sql
CREATE VIEW v AS SELECT * FROM ref WHERE a = 1;
BEGIN;
SELECT * FROM v JOIN ref ON (v.a = ref.a);
ERROR:  cannot join local tables and reference tables in a transaction block, udf block, or distributed CTE subquery
```

pykello: Fixed + tests. Also added tests and comments for the function join.

PR #3282, opened 2019-12-11
Add missing ORDER BY in a CTE

onderkalaci:
Otherwise, the query output might not be consistent.


onderkalaci:
Since this is a very simple regression test fix by adding an ORDER BY for consistent outputs, I'll go ahead and merge it without review.

PR #3284, opened 2019-12-11
update contributing

SaitTalhaNisanci:
For style and some other style related checks we have 3 checks. `citus_indent` is just one of them. We should run `make reindent` which runs all of the checks.  This PR updates the contributing.md with this information.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -132,10 +132,9 @@ tool.
   popd
   ```
 
-Once you've done that, you can run the `citus_indent` command to recursively check and
-correct the style of any source files in the current directory. You can also run `make
-reindent` from within the Citus repo to correct the style of all source files in the
-repository.
+Once you've done that, you can run the `make reindent` command from the top directory to recursively check and
+correct the style of any source files in the current directory. Under the hood, `make reindent` will run `citus_indent` and some
+other style corrections for you.
 
 ### Making SQL changes
 


onderkalaci:
I get the following error on Mac:
```
make reindent
/Users/onderkalaci/Documents/citus_code/citus/ci/fix_style.sh
sed: -i may not be used with stdin
make: *** [reindent] Error 1

```

SaitTalhaNisanci:
I think it is because of this line https://github.com/citusdata/citus/blob/master/ci/editorconfig.sh#L10

The script I wrote uses `editorconfig` as well, I thought it would already work on Mac as well but it seems that it doesnt.

So if you get a problem with editorconfig, how do you fix it then as you cannot run this script on Mac, do you fix it manually? Or you never get an error because you already have your editor configured to fix style?

onderkalaci:
> So if you get a problem with editorconfig, how do you fix it then as you cannot run this script on Mac, do you fix it manually? Or you never get an error because you already have your editor configured to fix style?

it happened few times, I fixed them manually by removing `-i` and doing some  more tricks. I agree that this is not the problem of `make reindent` but it wouldn't work until we fix the issue. 
I can look that later

SaitTalhaNisanci:
I will merge this PR but it would be useful you can do a patch for that. I dont have a macbook so I dont know how to fix it.

PR #3285, opened 2019-12-11
rename REMOTE_TRANS_INVALID to REMOTE_TRANS_IDLE

SaitTalhaNisanci:
REMOTE_TRANS_INVALID seems like a state where we should not be. Because it is an `invalid` state, however from the usage it seems that it is like the state where there is no transaction, so it is `idle`. I feel like the wording REMOTE_TRANS_IDLE better describes what this state is.

### DIFF
--- a/src/backend/distributed/executor/adaptive_executor.c
+++ b/src/backend/distributed/executor/adaptive_executor.c
@@ -1365,7 +1365,7 @@ CleanUpSessions(DistributedExecution *execution)
 
 				ClearResults(connection, false);
 			}
-			else if (!(transactionState == REMOTE_TRANS_INVALID ||
+			else if (!(transactionState == REMOTE_TRANS_NOT_STARTED ||
 					   transactionState == REMOTE_TRANS_STARTED))
 			{
 				/*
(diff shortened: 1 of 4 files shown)


src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -2677,7 +2677,7 @@ TransactionStateMachine(WorkerSession *session)
 				}
 				else
 				{
-					transaction->transactionState = REMOTE_TRANS_INVALID;
```
marcocitus: We should watch out for this one. We could (soon) get into situations where we're resetting the transaction state on a session that's actually in a transaction block.

SaitTalhaNisanci:
REMOTE_TRANS_READY
REMOTE_TRANS_NOT_STARTED
REMOTE_TRANS_IDLE
REMOTE_TRANS_FREE

Which one do you think is better? Or if you have any other suggestion we can use that too.

marcocitus:
`REMOTE_TRANS_NOT_STARTED` conveys the meaning best I think.

src/backend/distributed/executor/adaptive_executor.c
marcocitus: Fixed in https://github.com/citusdata/citus/pull/3288/files

PR #3288, opened 2019-12-11
Clean up transaction block usage logic in adaptive executor

marcocitus:
Currently StartDistributedExecution tries to infer how the transaction should be executed based on the current and preceding executions. That is somewhat useful, since it means we can blindly call ExecuteTaskList and have it mostly do the right thing. However, we sometimes want more precise control over the execution. 

This PR separates the logic for deciding on how to execute the transaction from StartDistributedExecution. ExtendedTaskListExtended now allows the caller to specify whether to use transaction blocks and error out on failure.

### DIFF
--- a/src/backend/distributed/executor/adaptive_executor.c
+++ b/src/backend/distributed/executor/adaptive_executor.c
@@ -242,17 +242,11 @@ typedef struct DistributedExecution
 	 */
 	bool raiseInterrupts;
 
-	/*
-	 * Flag to indicate whether the query is running in a distributed
-	 * transaction.
-	 */
-	bool isTransaction;
+	/* transactional properties of the current execution */
+	TransactionProperties *transactionProperties;
 
 	/* indicates whether distributed execution has failed */
 	bool failed;
-
-	/* set to true when we prefer to bail out early */
-	bool errorOnAnyFailure;
 
 	/*
 	 * For SELECT commands or INSERT/UPDATE/DELETE commands with RETURNING,
@@ -541,7 +535,12 @@ static DistributedExecution * CreateDistributedExecution(RowModifyLevel modLevel
 														 ParamListInfo paramListInfo,
 														 TupleDesc tupleDescriptor,
 														 Tuplestorestate *tupleStore,
-														 int targetPoolSize);
+														 int targetPoolSize,
+														 TransactionProperties *
+														 xactProperties);
+static TransactionProperties DecideTransactionPropertiesForTaskList(RowModifyLevel
+																	modLevel,
+																	List *taskList);
 static void StartDistributedExecution(DistributedExecution *execution);
 static void RunLocalExecution(CitusScanState *scanState, DistributedExecution *execution);
 static void RunDistributedExecution(DistributedExecution *execution);
@@ -556,8 +555,9 @@ static void AcquireExecutorShardLocksForExecution(DistributedExecution *executio
 static void AdjustDistributedExecutionAfterLocalExecution(DistributedExecution *
 														  execution);
 static bool DistributedExecutionModifiesDatabase(DistributedExecution *execution);
+static bool IsMultiShardModification(RowModifyLevel modLevel, List *taskList);
 static bool TaskListModifiesDatabase(RowModifyLevel modLevel, List *taskList);
-static bool DistributedExecutionRequiresRollback(DistributedExecution *execution);
+static bool DistributedExecutionRequiresRollback(List *taskList);
 static bool TaskListRequires2PC(List *taskList);
 static bool SelectForUpdateOnReferenceTable(RowModifyLevel modLevel, List *taskList);
 static void AssignTasksToConnections(DistributedExecution *execution);
@@ -621,7 +621,6 @@ AdaptiveExecutor(CitusScanState *scanState)
 	bool interTransactions = false;
 	int targetPoolSize = MaxAdaptiveExecutorPoolSize;
 
-
 	Job *job = distributedPlan->workerJob;
 	List *taskList = job->taskList;
 
@@ -646,13 +645,18 @@ AdaptiveExecutor(CitusScanState *scanState)
 	scanState->tuplestorestate =
 		tuplestore_begin_heap(randomAccess, interTransactions, work_mem);
 
+	TransactionProperties xactProperties =
+		DecideTransactionPropertiesForTaskList(distributedPlan->modLevel, taskList);
+
 	DistributedExecution *execution = CreateDistributedExecution(
-		distributedPlan->modLevel, taskList,
-		distributedPlan->
-		hasReturning, paramListInfo,
+		distributedPlan->modLevel,
+		taskList,
+		distributedPlan->hasReturning,
+		paramListInfo,
 		tupleDescriptor,
-		scanState->
-		tuplestorestate, targetPoolSize);
+		scanState->tuplestorestate,
+		targetPoolSize,
+		&xactProperties);
 
 	/*
 	 * Make sure that we acquire the appropriate locks even if the local tasks
@@ -771,8 +775,32 @@ ExecuteTaskList(RowModifyLevel modLevel, List *taskList, int targetPoolSize)
 	Tuplestorestate *tupleStore = NULL;
 	bool hasReturning = false;
 
+	TransactionProperties xactProperties =
+		DecideTransactionPropertiesForTaskList(modLevel, taskList);
+
 	return ExecuteTaskListExtended(modLevel, taskList, tupleDescriptor,
-								   tupleStore, hasReturning, targetPoolSize);
+								   tupleStore, hasReturning, targetPoolSize,
+								   &xactProperties);
+}
+
+
+/*
+ * ExecuteTaskListIntoTupleStore is a proxy to ExecuteTaskListExtended() with defaults
+ * for some of the arguments.
+ */
+uint64
+ExecuteTaskListIntoTupleStore(RowModifyLevel modLevel, List *taskList,
+							  TupleDesc tupleDescriptor, Tuplestorestate *tupleStore,
+							  bool hasReturning)
+{
+	int targetPoolSize = MaxAdaptiveExecutorPoolSize;
+
+	TransactionProperties xactProperties =
+		DecideTransactionPropertiesForTaskList(modLevel, taskList);
+
+	return ExecuteTaskListExtended(modLevel, taskList, tupleDescriptor,
+								   tupleStore, hasReturning, targetPoolSize,
+								   &xactProperties);
 }
 
 
@@ -783,7 +811,8 @@ ExecuteTaskList(RowModifyLevel modLevel, List *taskList, int targetPoolSize)
 uint64
 ExecuteTaskListExtended(RowModifyLevel modLevel, List *taskList,
 						TupleDesc tupleDescriptor, Tuplestorestate *tupleStore,
-						bool hasReturning, int targetPoolSize)
+						bool hasReturning, int targetPoolSize,
+						TransactionProperties *xactProperties)
 {
 	ParamListInfo paramListInfo = NULL;
 
@@ -800,7 +829,8 @@ ExecuteTaskListExtended(RowModifyLevel modLevel, List *taskList,
 
 	DistributedExecution *execution =
 		CreateDistributedExecution(modLevel, taskList, hasReturning, paramListInfo,
-								   tupleDescriptor, tupleStore, targetPoolSize);
+								   tupleDescriptor, tupleStore, targetPoolSize,
+								   xactProperties);
 
 	StartDistributedExecution(execution);
 	RunDistributedExecution(execution);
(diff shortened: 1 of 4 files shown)


src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -2671,7 +2763,7 @@ TransactionStateMachine(WorkerSession *session)
 				UpdateConnectionWaitFlags(session,
 										  WL_SOCKET_READABLE | WL_SOCKET_WRITEABLE);
 
-				if (execution->isTransaction)
+				if (useTransactionBlocks == TRANSACTION_BLOCKS_REQUIRED)
 				{
 					transaction->transactionState = REMOTE_TRANS_STARTED;
 				}
```
marcocitus: Setting the state back to invalid below is wrong in the ALLOWED case.

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -872,64 +922,112 @@ CreateDistributedExecution(RowModifyLevel modLevel, List *taskList, bool hasRetu
...
+ * that affect the remainder of the transaction.
+ *
+ * Additionally, it returns whether the current execution should error out on any failure
+ * and whether to use transaction blocks, which may depend on previous statements in the
+ * transaction.
  */
-void
-StartDistributedExecution(DistributedExecution *execution)
+static void
+InitialiseCoordinatedTransactionForTaskList(RowModifyLevel modLevel, List *taskList,
```
SaitTalhaNisanci: This function doesnt necessarily start a coordinated transaction, the method name implies that it starts a transaction always. This is more like Initializing the steps for execution, so I feel like we should choose a different name

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -872,64 +922,112 @@ CreateDistributedExecution(RowModifyLevel modLevel, List *taskList, bool hasRetu
...
+	 * The primary motivation is that at this point we're definitely expanding
+	 * the nodes participated in the transaction. And, by re-generating the
+	 * remote task lists during local query execution, we might prevent the adaptive
+	 * executor to kick-in 2PC (or even start coordinated transaction, that's why
+	 * we prefer adding this check here instead of
+	 * Activate2PCIfModifyingTransactionExpandsToNewNode()).
+	 */
+	if (LocalExecutionHappened)
+	{
+		BeginOrContinueCoordinatedTransaction();
```
SaitTalhaNisanci: Maybe I am missing something but to me it seems kind of counter intuitive that we start a transaction when `useTransactionBlocks` is not required then later we set it to `REQUIRED`. Shouldnt the reason we are starting the transaction be the other way around?

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -872,64 +922,112 @@ CreateDistributedExecution(RowModifyLevel modLevel, List *taskList, bool hasRetu
...
+			 * the failures are rare, and we prefer to avoid marking placements invalid
+			 * in case of failures.
+			 */
+			CoordinatedTransactionUse2PC();
+
+			*errorOnAnyFailure = true;
+			*useTransactionBlocks = TRANSACTION_BLOCKS_REQUIRED;
+			return;
+		}
+		else if (MultiShardCommitProtocol != COMMIT_PROTOCOL_2PC &&
```
SaitTalhaNisanci: For better readability we can have a method like `IsMultiShardModification(modLevel, taskList)`

src/include/distributed/multi_executor.h
```diff
@@ -26,6 +26,14 @@ typedef enum
 	SEQUENTIAL_CONNECTION = 1
 } MultiShardConnectionTypes;
 
+/* whether to use remote transaction blocks */
+typedef enum TransactionBlocksUsage
```
SaitTalhaNisanci: These are specifically remote transactions, so should we have a `REMOTE` prefix maybe?

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -815,14 +861,18 @@ ExecuteTaskListExtended(RowModifyLevel modLevel, List *taskList,
...
+						   TransactionBlocksUsage useRemoteTransactionBlocks)
 {
 	DistributedExecution *execution =
 		(DistributedExecution *) palloc0(sizeof(DistributedExecution));
 
 	execution->modLevel = modLevel;
 	execution->tasksToExecute = taskList;
 	execution->hasReturning = hasReturning;
+	execution->errorOnAnyFailure = errorOnAnyFailure;
+	execution->useRemoteTransactionBlocks = useRemoteTransactionBlocks;
```
SaitTalhaNisanci: At this point we can already determine what we want this value to be, right?. However, we might change this in InitialiseCoordinatedTransactionForTaskList, I feel like we should assign the last value (which will be used during the execution) here. Since there is only one method to create an execution, this will be the single source of truth. What do you think?

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -242,10 +242,10 @@ typedef struct DistributedExecution
 	bool raiseInterrupts;
 
 	/*
-	 * Flag to indicate whether the query is running in a distributed
-	 * transaction.
+	 * The transaction blocks usage indicates whether the current execution
```
SaitTalhaNisanci: /*
	 * useRemoteTransactionBlocks is used to manage remote transactions. It
	 * could state one of the followings:
	 * - opening a remote transaction is required
	 * - opening a remote transaction is disallowed
	 * - opening a remote transaction does not matter, so it is allowed but not required.
	 */

my comment was like this, you can use it if you find it more descriptive.

src/include/distributed/multi_executor.h
marcocitus: It messed with alignment a lot, so was looking for a way to make it shorter.

src/backend/distributed/executor/adaptive_executor.c
marcocitus: Not sure I follow, this is called after InitialiseCoordinatedTransactionForTaskList.

src/backend/distributed/executor/adaptive_executor.c
marcocitus: Prior to this function we are undecided on the value of useTransactionBlocks. Is the main issue that we set useTransactionBlocks to ALLOWED at the start of this function? I could move it from there.

SaitTalhaNisanci: No I am not talking about setting it at the beginning of function.

If useTransactionBlocks is required then we should start a coordinated transaction. So intuitively I would expect that it is set to required before this function is called, not after. Does that make sense?

src/backend/distributed/executor/adaptive_executor.c
SaitTalhaNisanci: my bad, I thought it was called afterwards

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -872,64 +923,111 @@ CreateDistributedExecution(RowModifyLevel modLevel, List *taskList, bool hasRetu
...
+DecideTransactionProperties(RowModifyLevel modLevel, List *taskList,
+							bool *errorOnAnyFailure,
+							TransactionBlocksUsage *useTransactionBlocks)
 {
-	List *taskList = execution->tasksToExecute;
+	*errorOnAnyFailure = false;
+	*useTransactionBlocks = TRANSACTION_BLOCKS_ALLOWED;
 
-	if (MultiShardCommitProtocol != COMMIT_PROTOCOL_BARE)
+	if (taskList == NIL)
```
SaitTalhaNisanci: From the coverage report it seems that taskList can be NIL, how can it be NIL? wouldnt it be better to return early in `AdaptiveExecutor` method so that we dont have to worry about this case in the rest of the method.

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -242,10 +242,14 @@ typedef struct DistributedExecution
 	bool raiseInterrupts;
 
 	/*
-	 * Flag to indicate whether the query is running in a distributed
-	 * transaction.
+	 * useRemoteTransactionBlocks is used to manage remote transactions according
+	 * to one of the following policies:
+	 *
+	 * - opening a remote transaction is required
```
onderkalaci: super minor: The order of the comments doesn't match with the order in the enum definition. So, maybe fix the order or add the enum name in parenthesis here?

(it confused me at first)

src/backend/distributed/executor/adaptive_executor.c
onderkalaci: Is  that we should not call `BeginOrContinueCoordinatedTransaction()`/`CoordinatedTransactionUse2PC()` in `DecideTransactionProperties()`,  but on `StartDistributedExecution()`? But then we should also get 1PC/2PC from this function as well, which makes sense btw.

src/backend/distributed/executor/adaptive_executor.c
onderkalaci: > From the coverage report it seems that taskList can be NIL, how can it be NIL?

tip: Run the regression tests by adding an error message, you'd  see which tests fail so you'd have a better idea on what types of queries/commands are affected by it
```C
diff --git a/src/backend/distributed/executor/adaptive_executor.c b/src/backend/distributed/executor/adaptive_executor.c
index abb9d351b..81e90edd7 100644
--- a/src/backend/distributed/executor/adaptive_executor.c
+++ b/src/backend/distributed/executor/adaptive_executor.c
@@ -941,6 +941,7 @@ DecideTransactionProperties(RowModifyLevel modLevel, List *taskList,
 
        if (taskList == NIL)
        {
+               elog(ERROR, "which queries fail");
                /* nothing to do */
                return;
        }
```
>  wouldnt it be better to return early in AdaptiveExecutor method so that we dont have to worry about this case in the rest of the method.

There seem to be some side effects  like acquiring some locks `LockPartitionsForDistributedPlan()`, or  things in `FinishDistributedExecution()`.

It might be possible to early return in adaptive executor,  but we should be careful  about  those side-effects, and I would probably not do  it in this PR, maybe with a PR focusing on that.

marcocitus:
> The change makes sense, but I think it'd be good to get 2PC decision out of that function as well.

Where should we put it? I'm trying to keep these decisions (e.g. 2PC and errorOnAnyFailure) together since they are tightly coupled. That's not nice and elegant, but that's because the actual behaviour is not nice and elegant, and this way we can easily see that.

onderkalaci:
> Where should we put it? 

I  thought of another output variable on `DecideTransactionProperties ()` ?

marcocitus:
Makes sense, I think I'll introduce a new struct that contains all the decisions to use as a return value.

marcocitus:
@onderkalaci is this roughly what you have in mind? Global state changes are now deferred until StartDistributedExecution, which is kind of nice in the sense that it allows us to reason about what will happen without actually doing it (e.g. could augment explain output).

src/include/distributed/multi_executor.h
```diff
@@ -26,6 +26,40 @@ typedef enum
...
+/*
+ * TransactionBlocksUsage indicates whether to use remote transaction
+ * blocks according to one of the following policies:
+ * - opening a remote transaction is required
+ * - opening a remote transaction does not matter, so it is allowed but not required.
+ * - opening a remote transaction is disallowed
+ */
+typedef enum TransactionBlocksUsage
+{
+	TRANSACTION_BLOCKS_REQUIRED,
```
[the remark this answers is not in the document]
marcocitus: No, we should be able to execute re-partition joins (disallowed) within a query with a CTE (required).

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -874,62 +906,114 @@ CreateDistributedExecution(RowModifyLevel modLevel, List *taskList, bool hasRetu
...
+		{
+			/*
+			 * Even if we're not using 2PC, we prefer to error out
+			 * on any failures during multi shard modifications/DDLs.
+			 */
+			xactProperties.errorOnAnyFailure = true;
 		}
 	}
-	else
+	else if (InCoordinatedTransaction())
```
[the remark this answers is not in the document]
marcocitus: What if we open a new connection? If we are already in a coordinated transaction then transaction blocks are required for the new connections.

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -874,62 +906,114 @@ CreateDistributedExecution(RowModifyLevel modLevel, List *taskList, bool hasRetu
...
 		 * the nodes participated in the transaction. And, by re-generating the
 		 * remote task lists during local query execution, we might prevent the adaptive
 		 * executor to kick-in 2PC (or even start coordinated transaction, that's why
 		 * we prefer adding this check here instead of
 		 * Activate2PCIfModifyingTransactionExpandsToNewNode()).
 		 */
-		if (DistributedExecutionRequiresRollback(execution) || LocalExecutionHappened)
-		{
-			BeginOrContinueCoordinatedTransaction();
+		xactProperties.errorOnAnyFailure = true;
```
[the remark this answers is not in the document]
marcocitus: Good catch!

PR #3289, opened 2019-12-12
not continue in sequential execution if a cancellation is received

### LINKED ISSUE #3023: Cancelling multi-row inserts in adaptive executor.
We use `SequentialRunDistributedExecution()` for multi-row inserts which calls `RunDistributedExecution()` with the same execution object multiple times. A cancel terminates `RunDistributedExecution()`, but we don't check for it in `SequentialRunDistributedExecution()`, and make the next call to `RunDistributedExecution()` regardless, which might have an unexpected behaviour because the execution object was not cleaned-up in the previous round.

We should handle cancellation in `SequentialRunDistributedExecution()`.

SaitTalhaNisanci:
If there is a cancellation in sequential execution, we should not continue execution and return. This PR adds that.

Fixes #3023.

### DIFF
--- a/src/backend/distributed/executor/adaptive_executor.c
+++ b/src/backend/distributed/executor/adaptive_executor.c
@@ -1808,6 +1808,13 @@ SequentialRunDistributedExecution(DistributedExecution *execution)
 		execution->totalTaskCount = 1;
 		execution->unfinishedTaskCount = 1;
 
+		CHECK_FOR_INTERRUPTS();
+
+		if (InterruptHoldoffCount > 0 && (QueryCancelPending || ProcDiePending))
+		{
+			break;
+		}
+
 		/* simply call the regular execution function */
 		RunDistributedExecution(execution);
 	}
--- a/src/test/regress/expected/failure_multi_row_insert.out
+++ b/src/test/regress/expected/failure_multi_row_insert.out
@@ -134,6 +134,23 @@ SELECT citus.mitmproxy('conn.onQuery(query="^INSERT").cancel(' || :pid || ')');
 
 INSERT INTO reference_table VALUES (1), (2), (3), (4);
 ERROR:  canceling statement due to user request
+SELECT citus.mitmproxy('conn.onQuery(query="^INSERT").cancel(' || :pid || ')');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+INSERT INTO distributed_table VALUES (1,1), (2,2), (3,3), (4,2), (5,2), (6,2), (7,2);
+ERROR:  canceling statement due to user request
+-- cancel the second insert over the same connection
+SELECT citus.mitmproxy('conn.onQuery(query="^INSERT").after(1).cancel(' || :pid || ')');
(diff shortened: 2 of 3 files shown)


src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -1808,6 +1808,13 @@ SequentialRunDistributedExecution(DistributedExecution *execution)
 		execution->totalTaskCount = 1;
 		execution->unfinishedTaskCount = 1;
 
+		CHECK_FOR_INTERRUPTS();
+
+		if (InterruptHoldoffCount > 0 && (QueryCancelPending || ProcDiePending))
```
marcocitus: Do we ever run this code in a HOLD_INTERRUPTS block?

src/test/regress/expected/failure_multi_row_insert.out
```diff
@@ -134,6 +134,14 @@ SELECT citus.mitmproxy('conn.onQuery(query="^INSERT").cancel(' || :pid || ')');
 
 INSERT INTO reference_table VALUES (1), (2), (3), (4);
 ERROR:  canceling statement due to user request
+SELECT citus.mitmproxy('conn.onQuery(query="^INSERT").cancel(' || :pid || ')');
+ mitmproxy 
+-----------
+ 
+(1 row)
+
+INSERT INTO distributed_table VALUES (1,1), (2,2), (3,3), (4,2), (5,2), (6,2), (7,2);
```
onderkalaci: Maybe  add one  more test with `conn.onQuery(query="^INSERT").after(1).cancel(' || :pid || ')` as well,  which is basically cancel on the second INSERT over  the same connection

onderkalaci: And, what  happens  if you run the tests when you revert the  code changes? I'm curious

src/backend/distributed/executor/adaptive_executor.c
marcocitus: Seems this is also on RunDistributedExecution and mostly to be defensive.

src/test/regress/expected/failure_multi_row_insert.out
marcocitus: Looking at it more this issue seems a bit theoretical. We normally just handle cancellations in CHECK_FOR_INTERRUPTS, which jumps out of SequentialRunDistributedExecution.

SaitTalhaNisanci: Yeah after talking with Onder, it seems that this is theoretical because we would normally return in CHECK_FOR_INTERRUPTS and the problem would happen only if the execution is run and prior to the execution run HOLD_INTERRUPTS was called. This is not the case and it is not easy to reproduce. So adding the check in SequentialRunExecution is to prevent future problems, currently there is no problem with that because we do not call HOLD_INTERRUPTS prior to `RunDistributedExecution`.

SaitTalhaNisanci:
I have added the test.

PR #3290, opened 2019-12-12
add IsCancellationReceived utility function

SaitTalhaNisanci:
This PR adds `IsHoldOffCancellationReceived` utility function to remove duplicate code in different parts of the codebase to understand cancellation when HOLD_INTERRUPTS is called.

The motivation behind this method is that we want to take immediate action in parts of execution that could take a long time. So even if HOLD_INTERRUPTS is set, we do the cancellation.

### DIFF
--- a/src/backend/distributed/connection/connection_management.c
+++ b/src/backend/distributed/connection/connection_management.c
@@ -25,6 +25,7 @@
 #include "distributed/hash_helpers.h"
 #include "distributed/placement_connection.h"
 #include "distributed/run_from_same_connection.h"
+#include "distributed/cancel_utils.h"
 #include "distributed/remote_commands.h"
 #include "distributed/version_compat.h"
 #include "mb/pg_wchar.h"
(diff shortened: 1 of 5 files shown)


src/backend/distributed/connection/connection_management.c
```diff
@@ -738,7 +739,7 @@ FinishConnectionListEstablishment(List *multiConnectionList)
 
 				CHECK_FOR_INTERRUPTS();
 
-				if (InterruptHoldoffCount > 0 && (QueryCancelPending || ProcDiePending))
+				if (IsCancellationReceived())
```
marcocitus: This catches a specific situation when we are in a HOLD_INTERRUPTS block to avoid stack unwinding  / errors. I wonder if that should be reflected in the name. e.g IsHoldOffCancellationReceived

SaitTalhaNisanci: @marcocitus I have also updated the comment for the method to be more clear why we have that.

PR #3293, opened 2019-12-12
replace `Begin` words in coordinated transactions with `Use`

SaitTalhaNisanci:
This PR replaces the action words in coordinated transactions. When I read the method `BeginOrContinueCoordinatedTransaction` for a very long time I assumed that It actually begins a coordinated transaction. Like it sends the `BEGIN` command. If it was sending a `BEGIN` command, I would name it just like this, so I find it confusing. Instead we do set up some state for later use. I have replaced `Begin` with `Use`.  Let me know if you have any other suggestion.

### DIFF
--- a/src/backend/distributed/transaction/transaction_management.c
+++ b/src/backend/distributed/transaction/transaction_management.c
@@ -96,7 +96,6 @@ bool FunctionOpensTransactionBlock = true;
 
 
 /* transaction management functions */
-static void BeginCoordinatedTransaction(void);
 static void CoordinatedTransactionCallback(XactEvent event, void *arg);
 static void CoordinatedSubTransactionCallback(SubXactEvent event, SubTransactionId subId,
 											  SubTransactionId parentSubid, void *arg);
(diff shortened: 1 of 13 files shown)


src/backend/distributed/transaction/transaction_management.c
```diff
@@ -170,11 +170,12 @@ InitializeTransactionManagement(void)
...
 /*
- * BeginCoordinatedTransaction begins a coordinated transaction. No
- * pre-existing coordinated transaction may be in progress./
+ * UseCoordinatedTransaction sets up the necessary variables to use
+ * a coordinated transaction. No pre-existing coordinated
+ * transaction may be in progress.
  */
 static void
-BeginCoordinatedTransaction(void)
+UseCoordinatedTransaction(void)
```
marcocitus: Do we really need this function?

src/backend/distributed/transaction/transaction_management.c
```diff
@@ -111,18 +111,18 @@ static bool MaybeExecutingUDF(void);
...
 
 /*
- * BeginOrContinueCoordinatedTransaction starts a coordinated transaction,
- * unless one already is in progress.
+ * UseOrKeepCoordinatedTransaction sets up the necessary variables to use
+ * a coordinated transaction, unless one is already in progress.
  */
 void
-BeginOrContinueCoordinatedTransaction(void)
+UseOrKeepCoordinatedTransaction(void)
```
marcocitus: If we remove the other function, could we omit "OrKeep"? Use seems descriptive enough.

SaitTalhaNisanci: Yeah I will remove the other function and remove `OrKeep`

SaitTalhaNisanci:
I will wait for @onderkalaci approval as well for this one.

Issue #3116, opened 2019-10-22
Auto-adjust the pool size in adaptive executor


onderkalaci:
Already listed in #2665 but probably better to have a separate issue. 

One of the most important settings of the adaptive executor is `citus.max_adaptive_executor_pool_size`, which is the max. limit of the connections that the executor is allowed to open per worker node. Setting it low decreases the parallelism (and hence the performance) on the workers. Setting it too high puts a lot of pressure on the workers, especially when there are concurrent queries running on the coordinator.

To avoid performance problems, we've implemented an algorithm called `Slow-Start` (#2779), where we start with a small number of connections per node, and open new connections when they are necessary in `10 msecs` (a.k.a, `citus.executor_slow_start_interval `). 

For example, if the queries on the shards take a short time (~`0.5 msec`), say index-only scan, the adaptive executor can use the same connection for executing all the shard queries on that worker node. That's great, it saves opening connections to the workers, which is pretty expensive operation.

The problem with that approach is that, if the workers are busy because of many concurrent queries, they'd execute queries slower. In response, the adaptive executor would open even more connections, putting a lot more pressure on the workers. Instead, it'd be great if the adaptive executor can realize that the workers are busy, so there is no point in opening more connections (e.g., decrease the pool size) - pushback.

Some alternatives we've discussed earlier, which makes some sense but requires further discussion, is the following. All of the above only applies for multi-shard queries. Single shard queries would always get the cached connection, so that's OK.

1) While (or at the same UDF?) the coordinator is sending `SELECT assign_distributed_transaction_id(0, 1, '2019-10-22 12:18:48.436979+02')`, it also sends one more command, which returns a metric to identify the busyness of the worker node. Depending on that, a heuristic on the coordinator decides the pool size. We could rely on `backend_data` infrastructure that we use for deadlock detection. This prevents having one more round-trip to the worker nodes.

2) The coordinator counts the number of concurrent queries per node in the shared memory. This doesn't work well on MX and could have performance impact because of shared memory accesses.

3) The coordinator periodically asks for the busyness of the worker nodes, similar to deadlock detection or 2PC recovery, and caches that information in a shared memory?. The problem with this is the interval to check this information. If the interval is long, it won't be accurate. If it is low, it'd add extra pressure to the workers, especially with MX, this'd be even more problematic.

Any further ideas? My favorite is 1, which adds little overhead, pretty accurate and scales well with MX.

SaitTalhaNisanci:
[...]

PR #3298, opened 2019-12-13
separate the logic in ManageWorkerPool

SaitTalhaNisanci:
This PR separates the logic in ManageWorkerPool.

We have 3 separate logics in this method:
- We check if we can currently open new connections
- We calculate how many new connections we can open
- We open the calculated amount of new connections.

I have separated these 3 in `ManageWorkerPool` so that we clearly see the separation and it is more clear what is going on in this function. I have also refactored some parts in the function for better readability.

### DIFF
--- a/src/backend/distributed/executor/adaptive_executor.c
+++ b/src/backend/distributed/executor/adaptive_executor.c
@@ -603,6 +603,10 @@ static WorkerPool * FindOrCreateWorkerPool(DistributedExecution *execution,
 static WorkerSession * FindOrCreateWorkerSession(WorkerPool *workerPool,
 												 MultiConnection *connection);
 static void ManageWorkerPool(WorkerPool *workerPool);
+static bool ShouldWaitForSlowStart(WorkerPool *workerPool);
+static int CalculateNewConnectionCount(WorkerPool *workerPool);
+static void OpenNewConnections(WorkerPool *workerPool, int newConnectionCount,
+							   TransactionProperties *transactionProperties);
 static void CheckConnectionTimeout(WorkerPool *workerPool);
 static int UsableConnectionCount(WorkerPool *workerPool);
 static long NextEventTimeout(DistributedExecution *execution);
@@ -615,6 +619,7 @@ static bool StartPlacementExecutionOnSession(TaskPlacementExecution *placementEx
 											 WorkerSession *session);
 static void ConnectionStateMachine(WorkerSession *session);
 static void HandleMultiConnectionSuccess(WorkerSession *session);
+static bool HasAnyConnectionFailure(WorkerPool *workerPool);
 static void Activate2PCIfModifyingTransactionExpandsToNewNode(WorkerSession *session);
 static bool TransactionModifiedDistributedTable(DistributedExecution *execution);
 static void TransactionStateMachine(WorkerSession *session);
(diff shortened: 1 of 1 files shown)


SaitTalhaNisanci:
This PR has been open for a few months, maybe someone can review it

onderkalaci:
I may change `ManageWorkerPool` in #3116. So, can we hold off this PR for a while? I may incorporate this changes there (or adjust this PR as necessary for #3116)

SaitTalhaNisanci:
TPCC gave the following result:

This branch: 425619
master: 408761

I don't think this PR would increase performance. It might be that the region was loaded during the test of master therefore the network traffic was slower maybe.

I will rerun it.

SaitTalhaNisanci:
Another run:

this branch: 412678
master: 412435

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -2361,38 +2427,38 @@ ManageWorkerPool(WorkerPool *workerPool)
 		 * Number of additional connections we would need to run all ready tasks in
 		 * parallel.
 		 */
-		int newConnectionsForReadyTasks = readyTaskCount - usableConnectionCount;
+		int newConnectionsForReadyTasks = Max(0, readyTaskCount - usableConnectionCount);
```
SaitTalhaNisanci, line 2430: Here there is an additional Max because newConnectionsForReadyTasks cannot be negative. If it is set to zero we won't open any new connection.

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -2361,38 +2436,38 @@ ManageWorkerPool(WorkerPool *workerPool)
...
 		/*
 		 * Open enough connections to handle all tasks that are ready, but no more
 		 * than the target pool size.
 		 */
 		newConnectionCount = Min(newConnectionsForReadyTasks, maxNewConnectionCount);
 
-		if (newConnectionCount > 0 && ExecutorSlowStartInterval != SLOW_START_DISABLED)
+		if (newConnectionCount > 0)
 		{
-			if (MillisecondsPassedSince(workerPool->lastConnectionOpenTime) >=
```
SaitTalhaNisanci, line 2374: here we check if it is too early to open a new connection after calculating the new connection count. This is now done in IsTooEarlyToOpenConnections

SaitTalhaNisanci:
@onderkalaci can we merge this to 9.5 maybe?

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -2334,9 +2349,69 @@ ManageWorkerPool(WorkerPool *workerPool)
 	if (failedConnectionCount >= 1)
 	{
 		/* do not attempt to open more connections after one failed */
-		return;
+		return false;
 	}
 
+	if (IsTooEarlyToOpenConnections(workerPool))
```
marcocitus, line 2355: this function name is confusing me. This is effectively part of the slow start logic, which should be called out more explicitly.

Maybe rename to WaitForSlowStart and move out of CanOpenNewConnection and into ManageWorkerPool?

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -2418,11 +2452,67 @@ ManageWorkerPool(WorkerPool *workerPool)
...
+	}
+
+	/* if slow start is disabled, we can open new connections */
+	if (ExecutorSlowStartInterval == SLOW_START_DISABLED)
+	{
+		return false;
+	}
+
+	return MillisecondsPassedSince(workerPool->lastConnectionOpenTime) <
+		   ExecutorSlowStartInterval;
```
marcocitus, line 2485: Could maybe expand into a variable + if statement just for readability.

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -2387,29 +2393,57 @@ static void
...
+ * for the given workerPool.
+ */
+static bool
+CanOpenNewConnection(WorkerPool *workerPool)
+{
+	if (HasAnyConnectionFailure(workerPool))
+	{
+		return false;
+	}
+	if (ShouldWaitForSlowStart(workerPool))
```
marcocitus, line 2427: I'd rather inline these 2 checks in ManageWorkerPool so I don't need to wonder what "can open new connection" means, or how it is different from CalculateNewConnectionCount returning 0.

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -2387,29 +2392,41 @@ static void
...
-		workerPool->idleConnectionCount;
-	int failedConnectionCount = workerPool->failedConnectionCount;
-	int readyTaskCount = workerPool->readyTaskCount;
-	int newConnectionCount = 0;
 
-	/* we should always have more (or equal) active connections than idle connections */
-	Assert(activeConnectionCount >= idleConnectionCount);
+	if (HasAnyConnectionFailure(workerPool))
+	{
+		return;
```
marcocitus, line 2398: Could use a comment to explain why we return:

`/* we do not expand the pool further if there was any failure */`

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -2387,29 +2392,41 @@ static void
...
+	}
+	if (ShouldWaitForSlowStart(workerPool))
+	{
+		return;
```
marcocitus, line 2402: Could use a comment to explain why we return:

`/* we wait until a slow start interval has passed before expanding the pool */`

PR #3301, opened 2019-12-13
Use any available non-data connection for intermediate results

### LINKED ISSUE #3300: Repeated call to a function with CTEs may exhaust connections
Calling a function with a query with CTEs many times may cause us to leak connections and eventually run out of connections on the worker.

```sql
CREATE OR REPLACE FUNCTION run_ctes(id int)
RETURNS text
LANGUAGE 'plpgsql' 
AS $BODY$
DECLARE
  value text;
BEGIN
	WITH 
	dist AS (SELECT x FROM public.test)
	SELECT count(*) INTO value FROM dist WHERE x = id;

	RETURN value ;
END;
$BODY$;

SELECT run_ctes(s) FROM generate_series(1,1000) s;
```

Note that this works without the `x = id` filter.

marcocitus:
DESCRIPTION: Fix an issue that caused CTEs to sometimes leak connections

When we send intermediate results to workers we pick a connection that did not access any placements. The reason is that the connection might otherwise occupy a connection that is needed during the query. However, the logic in StartNonDataAccessConnection picked an arbitrary connection and decided to open a new one if it accessed data, rather than using any of the available connections that did not access data. Under some circumstances, this would cause repeated CTEs in the same transaction to keep opening new connections because a connection that is inspected in StartNonDataAccessConnection is always one that accessed a placement.

This PR fixes the issue by inspecting all available connections when asking for a non-data access connection, and marking it as non-data-access to avoid using it for placement access, such that there will always be a connection available for sending intermediate results.

Fixes #3300

### DIFF
--- a/src/backend/distributed/connection/connection_management.c
+++ b/src/backend/distributed/connection/connection_management.c
@@ -50,6 +50,7 @@ static bool ShouldShutdownConnection(MultiConnection *connection, const int
 static void ResetConnection(MultiConnection *connection);
 static void DefaultCitusNoticeProcessor(void *arg, const char *message);
 static MultiConnection * FindAvailableConnection(dlist_head *connections, uint32 flags);
+static void GivePurposeToConnection(MultiConnection *connection, int flags);
 static bool RemoteTransactionIdle(MultiConnection *connection);
 static int EventSetSizeForConnectionList(List *connections);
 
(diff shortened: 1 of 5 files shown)


src/backend/distributed/connection/connection_management.c
```diff
@@ -386,6 +345,28 @@ FindAvailableConnection(dlist_head *connections, uint32 flags)
 			continue;
 		}
 
+		if ((flags & NO_DATA_ACCESS_CONNECTION) != 0)
+		{
+			if (!connection->isNonDataAccess &&
```
SaitTalhaNisanci: Should this be `||` instead of `&&` ?

marcocitus: No, but it does seem slightly unintuitive. Maybe I should swap the if-else.

SaitTalhaNisanci: So If a connection is used for data access but it is not used for any placements, we want to use that connection for non data access query?

Or If a connection is non data access, but it is used for some placement, do we want to return that connection?

src/backend/distributed/connection/connection_management.c
```diff
@@ -386,6 +345,28 @@ FindAvailableConnection(dlist_head *connections, uint32 flags)
...
+				ConnectionUsedForAnyPlacements(connection))
+			{
+				/* cannot use this connection for intermediate results */
+				continue;
+			}
+			else
+			{
+				connection->isNonDataAccess = true;
```
onderkalaci: Is there a reason not to mark  `isNonDataAccess`  after  calling `StartConnectionEstablishment()` instead of here? Otherwise, we might still end-up having more than one non-data access connections for certain CTE patterns?

I think it  could  also help simplifying this logic a bit, where `connection->isNonDataAccess == false`,  we don't need  to  check `ConnectionUsedForAnyPlacements(connection)`, because it'd be true by default.
Or, we could  assert `ConnectionUsedForAnyPlacements(connection) ==  false` when `connection->isNonDataAccess = true` etc.

(Plus, we should be able to remove `ClaimConnectionExclusively()` in  `RemoteFileDestReceiverStartup`  as long  as  we send  a  single intermediate result per node per query)

marcocitus: Refactored it. I also found the double negation  (!isNon) confusing, so I renamed it to sideChannel.

marcocitus: A downside I'm still looking to address is that we cannot reuse a cached connection.

marcocitus: Introduced a connection purpose to distinguish between any (cached), data access and sidechannel.

marcocitus:
There's a crash in Circle CI in multi_explain:
```
 SELECT true AS valid FROM explain_json($$
 	SELECT count(*)
 	FROM lineitem, orders, customer_append, supplier_single_shard
 	WHERE l_orderkey = o_orderkey
 	AND o_custkey = c_custkey
 	AND l_suppkey = s_suppkey$$);
-t
```

PR #3302, opened 2019-12-13
Fix a crash when calling a distributed function from PL/pgSQL

### LINKED ISSUE #3291: Using return value of a delegated function inside  another function errors
Reported on slack, I didn't have much time to look,  just created steps to repro:

```SQL
CREATE TABLE statement_table(id int, name text);
SELECT create_distributed_table('statement_table','id');

CREATE OR REPLACE FUNCTION increment(int)
RETURNS NUMERIC AS $$
DECLARE ret_val NUMERIC;
BEGIN
        SELECT max(id)::numeric INTO ret_val  FROM statement_table WHERE id = $1+1;
        RETURN ret_val;
END;
$$  LANGUAGE plpgsql
    SECURITY DEFINER
    -- Set a secure search_path: trusted schema(s), then 'pg_temp'.
    SET search_path = admin, public;


SELECT create_distributed_function('increment(int)', '$1', colocate_with := 'statement_table');


CREATE OR REPLACE FUNCTION test_func_calls_dist_func()
RETURNS NUMERIC AS $$
DECLARE incremented_val NUMERIC;
BEGIN
        SELECT INTO incremented_val increment(1);
        RETURN incremented_val;
END;
$$  LANGUAGE plpgsql
    SECURITY DEFINER
    -- Set a secure search_path: trusted schema(s), then 'pg_temp'.
    SET search_path = admin, public;

SELECT test_func_calls_dist_func();

SELECT test_func_calls_dist_func();
ERROR:  unexpected plan node type: 34
CONTEXT:  PL/pgSQL function test_func_calls_dist_func() line 4 at SQL statement
Time: 1.305 ms


SELECT test_func_calls_dist_func();
Time: 341.933 ms
!> \q
```

marcocitus:
DESCRIPTION: Fix a crash when calling a distributed function from PL/pgSQL

We currently do not always correctly detect being in PL/pgSQL, which causes issues when calling a distributed function from PL/pgSQL since we might incorrectly delegate it.

Fixes #3291

### DIFF
--- a/src/backend/distributed/planner/distributed_planner.c
+++ b/src/backend/distributed/planner/distributed_planner.c
@@ -63,6 +63,9 @@ static List *plannerRestrictionContextList = NIL;
 int MultiTaskQueryLogLevel = MULTI_TASK_QUERY_INFO_OFF; /* multi-task query log level */
 static uint64 NextPlanId = 1;
 
+/* keep track of planner call stack levels */
+int PlannerLevel = 0;
+
 
 static bool ListContainsDistributedTableRTE(List *rangeTableList);
 static bool IsUpdateOrDelete(Query *query);
@@ -189,6 +192,14 @@ distributed_planner(Query *parse, int cursorOptions, ParamListInfo boundParams)
 	PG_TRY();
 	{
 		/*
+		 * We keep track of how many times we've recursed into the planner, primarily
+		 * to detect whether we are in a function call. We need to make sure that the
+		 * PlannerLevel is decremented exactly once at the end of this PG_TRY block,
+		 * both in the happy case and when an error occurs.
+		 */
+		PlannerLevel++;
+
+		/*
 		 * For trivial queries, we're skipping the standard_planner() in
 		 * order to eliminate its overhead.
 		 *
@@ -242,13 +253,19 @@ distributed_planner(Query *parse, int cursorOptions, ParamListInfo boundParams)
 				result->planTree->total_cost = FLT_MAX / 100000000;
 			}
 		}
+
+		PlannerLevel--;
 	}
 	PG_CATCH();
 	{
 		PopPlannerRestrictionContext();
+
+		PlannerLevel--;
+
 		PG_RE_THROW();
 	}
 	PG_END_TRY();
+
 
 	/* remove the context from the context list */
 	PopPlannerRestrictionContext();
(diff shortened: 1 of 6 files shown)


src/backend/distributed/transaction/transaction_management.c
```diff
@@ -695,9 +697,12 @@ IsMultiStatementTransaction(void)
...
  * CTEs, etc. which also start nested executors.
+ *
+ * If the planner is being called from the executor, then we may also be in
+ * a UDF.
  */
 static bool
 MaybeExecutingUDF(void)
 {
-	return ExecutorLevel > 1;
+	return ExecutorLevel > 1 || (PlannerLevel > 0 && ExecutorLevel > 0);
```
thanodnl:
```suggestion
	return ExecutorLevel > 1 || (ExecutorLevel == 1  && PlannerLevel > 0);
```

By rephrasing it like suggested it will be slightly easier to read at once without changing the behaviour.

By having `ExecutorLevel` at the end and the check which has a lot of overlap it made it hard to understand we always return true for `ExecutorLevel` 2 and above. By getting the `ExecutorLevel` check in the front and a hard equal on `1` it reads more intuitive of what we are checking.

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -242,14 +248,24 @@ distributed_planner(Query *parse, int cursorOptions, ParamListInfo boundParams)
 				result->planTree->total_cost = FLT_MAX / 100000000;
 			}
 		}
+
+		PlannerLevel = originalLevel;
```
thanodnl: Other places [1] where we keep counters like this we perform a decrement both here and in the `PG_CATCH`. I have two questions:

1. Is the use of  `originalLevel` safe in `PG_CATCH`? I thought all local variables are unsafe to use during rollback, hence we were decrementing global variables in other places:
2. Is it unsafe to decrement? Would there be a possibility of missing a decrement?

[1] https://github.com/citusdata/citus/blob/2829c601dd48dcdbbbf3c748777f0b1b056484fa/src/backend/distributed/commands/utility_hook.c#L269

src/backend/distributed/transaction/transaction_management.c
```diff
@@ -302,6 +303,7 @@ CoordinatedTransactionCallback(XactEvent event, void *arg)
 			activeSetStmts = NULL;
 			CoordinatedTransactionUses2PC = false;
 			ExecutorLevel = 0;
+			PlannerLevel = 0;
```
thanodnl: Would it be a bug if this is not reset to `0`? Wonder if an `Assert` is in place.

PR #3304, opened 2019-12-14
Implement read_intermediate_results

pykello:
Implements a UDF for reading multiple result files in the same call, like:

```sql
SELECT * FROM 
  read_intermediate_results(ARRAY['squares_1', 'squares_2', 'squares_3']::text[], 'binary') 
    AS res (x int, x2 int);
```

This is a step towards implementing `INSERT INTO ... SELECT`

### DIFF
--- a/src/backend/distributed/planner/distributed_planner.c
+++ b/src/backend/distributed/planner/distributed_planner.c
@@ -99,6 +99,13 @@ static void CheckNodeIsDumpable(Node *node);
 static Node * CheckNodeCopyAndSerialization(Node *node);
 static void AdjustReadIntermediateResultCost(RangeTblEntry *rangeTableEntry,
 											 RelOptInfo *relOptInfo);
+static void AdjustReadIntermediateResultArrayCost(RangeTblEntry *rangeTableEntry,
+												  RelOptInfo *relOptInfo);
+static void AdjustReadIntermediateResultsCostInternal(RelOptInfo *relOptInfo,
+													  List *columnTypes,
+													  int resultIdCount,
+													  Datum *resultIds,
+													  Const *resultFormatConst);
 static List * OuterPlanParamsList(PlannerInfo *root);
 static List * CopyPlanParamList(List *originalPlanParamList);
 static PlannerRestrictionContext * CreateAndPushPlannerRestrictionContext(void);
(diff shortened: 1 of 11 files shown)


pykello:
> Would be nice to adjust the cost in the hook in distributed planner as we do for the regular read_intermediate_result call. Probably not necessary for insert..select, but will be helpful for avoiding really bad plans if we start using the function more.

Done. Do you know why `COSTS` were off for `read_intermediate_result()` cost estimation tests in `intermediate_results.sql`? Maybe because they were not consistent, but that could probably solved with a UDF which parsed the `EXPLAIN` output rather than turning it off.

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -1620,41 +1604,135 @@ AdjustReadIntermediateResultCost(RangeTblEntry *rangeTableEntry, RelOptInfo *rel
...
 
+	Datum resultFormatDatum = resultFormatConst->constvalue;
+	Oid resultFormatId = DatumGetObjectId(resultFormatDatum);
 	if (resultFormatId == BinaryCopyFormatId())
 	{
 		binaryFormat = true;
 
 		/* subtract 11-byte signature + 8 byte header + 2-byte footer */
-		resultSize -= 21;
+		totalResultSize -= 21;
```
marcocitus: Shouldn't this be per result file?

marcocitus:
> Do you know why COSTS were off for read_intermediate_result() cost estimation tests in intermediate_results.sql? Maybe because they were not consistent, but that could probably solved with a UDF which parsed the EXPLAIN output rather than turning it off.

Not sure, but sounds like a reasonable explanation and approach.

src/backend/distributed/planner/distributed_planner.c
pykello: Yes. done.

src/test/regress/sql/intermediate_results.sql
```diff
@@ -148,4 +148,54 @@ select broadcast_intermediate_result('a', 'create table foo(int serial)');
...
+-- error behaviour, and also check that results are deleted on rollback
+BEGIN;
+SELECT create_intermediate_result('squares_1', 'SELECT s, s*s FROM generate_series(1,3) s');
+SAVEPOINT s1;
+SELECT * FROM read_intermediate_results(ARRAY['notexistingfile', 'squares_1'], 'binary') AS res (x int, x2 int);
+ROLLBACK TO SAVEPOINT s1;
+SELECT * FROM read_intermediate_results(ARRAY['squares_1', 'notexistingfile'], 'binary') AS res (x int, x2 int);
+ROLLBACK TO SAVEPOINT s1;
+SELECT * FROM read_intermediate_results(ARRAY['squares_1', NULL], 'binary') AS res (x int, x2 int);
+END;
```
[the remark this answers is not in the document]
pykello: There is a test with empty array in the previous `BEGIN/END` block. Added one more here after roll back to savepoint to make sure rolling back doesn't mess up the state.

PR #3306, opened 2019-12-14
Fix union (all) pushdown issue

### LINKED ISSUE #3303: Views that UNION differently typed distribution columns error out
A VIEW that unions the distribution columns of 2 tables that have different colocation groups errors out with the following message. The SQL query from the view works correctly when run directly. 

```
DETAIL:  Shards of relations in subquery need to have 1-to-1 shard partitioning
```

Reproduction SQL:
```
create table viewtest1(col1 int, col2 int);
create table viewtest2(col1 float, col2 int);
select create_distributed_table('viewtest1','col1');
select create_distributed_table('viewtest2','col1');

create view test_view_1
AS
select col1 from viewtest1
union all
select col1 from viewtest2;

```

marcocitus:
DESCRIPTION: Fixes an issue causing some queries with UNION (ALL) to fail

Our co-location checks currently don't do a great job of actually checking for co-location. That causes unions on non-co-located tables to pass all the equivalence checks and fail in the logical planner, rather than being recursively planned.

As a side-effect, some error messages in other non-co-located join cases improved.

Fixes #3303

### DIFF
--- a/src/backend/distributed/planner/relation_restriction_equivalence.c
+++ b/src/backend/distributed/planner/relation_restriction_equivalence.c
@@ -10,7 +10,9 @@
  */
 #include "postgres.h"
 
+#include "distributed/colocation_utils.h"
 #include "distributed/distributed_planner.h"
+#include "distributed/listutils.h"
 #include "distributed/metadata_cache.h"
 #include "distributed/multi_logical_planner.h"
 #include "distributed/multi_logical_optimizer.h"
@@ -133,6 +135,8 @@ static void ListConcatUniqueAttributeClassMemberLists(AttributeEquivalenceClass 
 													  secondClass);
 static Index RelationRestrictionPartitionKeyIndex(RelationRestriction *
 												  relationRestriction);
+static bool AllRelationsInRestrictionContextColocated(RelationRestrictionContext *
+													  restrictionContext);
 static RelationRestrictionContext * FilterRelationRestrictionContext(
 	RelationRestrictionContext *relationRestrictionContext,
 	Relids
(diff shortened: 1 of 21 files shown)


src/backend/distributed/planner/relation_restriction_equivalence.c
```diff
@@ -626,13 +628,27 @@ EquivalenceListContainsRelationsEquality(List *attributeEquivalenceList,
 		RelationRestriction *relationRestriction =
 			(RelationRestriction *) lfirst(relationRestrictionCell);
 		int rteIdentity = GetRTEIdentity(relationRestriction->rte);
+		int colocationId = TableColocationId(relationRestriction->relationId);
```
onderkalaci: I think this works too, but as noted below, some test outputs sounds less accurate now.
And, also this place seems too deep inside the planning to make the colocation checks to me.

One suggestion could be to make the colocation checks more explicit  on `SafeToPushdownUnionSubquery()` and `RestrictionEquivalenceForPartitionKeys`. To do  that, we  could consider  refactoring `ErrorIfUnsupportedShardDistribution()` function to  defer error, and on these functions return false if the relations are not colocated. 

To improve the  error messages ,  we can use the same  check on `DeferErrorIfUnsupportedSubqueryPushdown()`, before getting into the other checks.

I can see that we'd need  to enforce the same check in multiple places, but still wanted to push this comment in case you can propose  something simpler.

Otherwise, we can merge  this  patch

marcocitus: I agree, would make sense to have a separate check.

src/test/regress/expected/full_join.out
```diff
@@ -130,8 +130,7 @@ ORDER BY 1;
 
 -- Full outer join with different distribution column types, should error out
 SELECT * FROM test_table_1 full join test_table_2 using(id);
-ERROR:  cannot push down this subquery
-DETAIL:  Shards of relations in subquery need to have 1-to-1 shard partitioning
+ERROR:  complex joins are only supported when all distributed tables are joined on their distribution columns with equal operator
```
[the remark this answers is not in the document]
marcocitus: User have complained about not the "1-to-1" message being hard to understand. I adjusted the complex joins message to be more generally applicable.

PR #3307, opened 2019-12-14
Do not repeat GROUP BY distribution_column on coordinator

### LINKED ISSUE #3292: Redundant GROUP BY distribution_column on the coordinator
We push down `GROUP BY distribution_column` to the workers, but we also evaluate it on the coordinator:

```sql
postgres=# explain SELECT x FROM test GROUP BY x;
┌────────────────────────────────────────────────────────────────────────────────────────────┐
│                                         QUERY PLAN                                         │
├────────────────────────────────────────────────────────────────────────────────────────────┤
│ HashAggregate  (cost=0.00..0.00 rows=0 width=0)                                            │
│   Group Key: remote_scan.x                                                                 │
│   ->  Custom Scan (Citus Adaptive)  (cost=0.00..0.00 rows=0 width=0)                       │
│         Task Count: 32                                                                     │
│         Tasks Shown: One of 32                                                             │
│         ->  Task                                                                           │
│               Node: host=127.0.0.1 port=5432 dbname=postgres                               │
│               ->  HashAggregate  (cost=38.25..40.25 rows=200 width=4)                      │
│                     Group Key: x                                                           │
│                     ->  Seq Scan on test_102041 test  (cost=0.00..32.60 rows=2260 width=4) │
└────────────────────────────────────────────────────────────────────────────────────────────┘
(10 rows)

```

We should stop doing it on the coordinator, since the results can simply be concatenated. That would also allow us to use re-partitioning after a GROUP BY on the source distribution column in #3267.

marcocitus:
DESCRIPTION: Improves performance for GROUP BY on the distribution column

This change omits the GROUP BY and HAVING clause from the master MultiExtendedOpNode if the GROUP BY is on the distribution column. In case of re-partition joins, we try to determine if the GROUP BY is on the join column of the outer-most join. 

Previously, `groupedByDisjointPartitionColumn` was true when the GROUP BY matched the partition column of any table, but that is not entirely correct. Since we're now relying more heavily on `groupedByDisjointPartitionColumn`, I rewrote the function that determines this value to terminate at MultiPartition, since anything below a MultiPartition will not be in the same Job.

Before:
```sql
EXPLAIN SELECT x, count(*) FROM test GROUP BY x;
┌────────────────────────────────────────────────────────────────────────────────────────────┐
│                                         QUERY PLAN                                         │
├────────────────────────────────────────────────────────────────────────────────────────────┤
│ HashAggregate  (cost=0.00..0.00 rows=0 width=0)                                            │
│   Group Key: remote_scan.x                                                                 │
│   ->  Custom Scan (Citus Adaptive)  (cost=0.00..0.00 rows=0 width=0)                       │
│         Task Count: 32                                                                     │
│         Tasks Shown: One of 32                                                             │
│         ->  Task                                                                           │
│               Node: host=localhost port=9700 dbname=postgres                               │
│               ->  HashAggregate  (cost=43.90..45.90 rows=200 width=12)                     │
│                     Group Key: x                                                           │
│                     ->  Seq Scan on test_102008 test  (cost=0.00..32.60 rows=2260 width=4) │
└────────────────────────────────────────────────────────────────────────────────────────────┘
(10 rows)
```

After:
```sql
EXPLAIN SELECT x, count(*) FROM test GROUP BY x;
┌──────────────────────────────────────────────────────────────────────────────────────┐
│                                      QUERY PLAN                                      │
├──────────────────────────────────────────────────────────────────────────────────────┤
│ Custom Scan (Citus Adaptive)  (cost=0.00..0.00 rows=0 width=0)                       │
│   Task Count: 32                                                                     │
│   Tasks Shown: One of 32                                                             │
│   ->  Task                                                                           │
│         Node: host=localhost port=9700 dbname=postgres                               │
│         ->  HashAggregate  (cost=43.90..45.90 rows=200 width=12)                     │
│               Group Key: x                                                           │
│               ->  Seq Scan on test_102008 test  (cost=0.00..32.60 rows=2260 width=4) │
└──────────────────────────────────────────────────────────────────────────────────────┘
(8 rows)
```

We should have probably done this a few years earlier since the extra aggregation on the coordinator can take up a lot of memory and CPU cycles, but it becomes extra relevant for INSERT..SELECT with re-partitioning (#3267 #3078) because it would otherwise prevent `INSERT INTO a SELECT b.distcol, ... FROM b GROUP BY b.distcol, ...` type queries from being re-partitioned, and those seems useful for aggregation. 

We could consider relaxing the checks in `ErrorIfContainsUnsupportedAggregate` when grouping by distribution column to allow any kind of aggregate to be used in queries that are grouped by the distribution column at the top level (workaround is to wrap the query in a subquery), but that's beyond the scope of this PR.

Fixed #3292


src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -2213,11 +2225,15 @@ ProcessTargetListForWorkerQuery(List *targetEntryList,
 		workerAggContext->createGroupByClause = false;
 
 		/*
-		 * If the expression uses aggregates inside window function contain agg
-		 * clause still returns true. We want to make sure it is not a part of
-		 * window function before we proceed.
+		 * If the query has a window function, we currently assuem it's safe to push
```
serprex: s/assuem/assume/

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -2213,11 +2225,15 @@ ProcessTargetListForWorkerQuery(List *targetEntryList,
...
+		 * down the target list.
+		 *
+		 * If there are aggregates and without a GROUP BY on the distribution column
```
serprex: Drop 'and', insert comma after 'column'

src/backend/distributed/planner/extended_op_node_utils.c
```diff
@@ -83,41 +81,86 @@ BuildExtendedOpNodeProperties(MultiExtendedOp *extendedOpNode)
...
 	}
+	else if (CitusIsA(node, MultiPartition))
+	{
+		MultiPartition *partitionNode = (MultiPartition *) node;
 
-	return result;
+		if (GroupedByColumn(opNode->groupClauseList, opNode->targetList,
+							partitionNode->partitionColumn))
+		{
+			/* only range- and hash-distributed tables are strictly partitioned  */
```
pykello: does this comment belong to here?

src/backend/distributed/planner/extended_op_node_utils.c
```diff
@@ -83,41 +81,86 @@ BuildExtendedOpNodeProperties(MultiExtendedOp *extendedOpNode)
...
+			return false;
 		}
 
 		if (GroupedByColumn(opNode->groupClauseList, opNode->targetList,
 							tableNode->partitionColumn))
 		{
-			result = true;
-			break;
+			/* only range- and hash-distributed tables are strictly partitioned  */
```
pykello: does this comment belong to here?

serprex: I don't think so. I lean towards removing these two comments without replacement

PR #3309, opened 2019-12-16
add code coverage badge

SaitTalhaNisanci:
This PR adds code coverage badge so that the master code coverage will be displayed in README.

You can see how that looks like here: 
https://github.com/citusdata/citus/blob/58ed931d7a04b2a98e1adcba9911497814b33e4a/README.md


onderkalaci:
Nice!

PR #3311, opened 2019-12-16
FIX: Make sure to correctly decrement ExecutorLevel

thanodnl:
DESCRIPTION: Fix counter that keeps track of internal depth in executor

While reviewing #3302 I ran into the `ExecutorLevel` variable which used a variable to keep the original value to restore on successful exit. I haven't explored the full space and if it is possible to get into an inconsistent state. However using `PG_TRY`/`PG_CATCH` seems generally more correct.

Given very bad things will happen if this level is not reset, I kept the failsafe of setting the variiable back to 0 on the `XactCallback` but I did add an assert to treat it as a developer bug.

### DIFF
--- a/src/backend/distributed/transaction/transaction_management.c
+++ b/src/backend/distributed/transaction/transaction_management.c
@@ -290,6 +290,13 @@ CoordinatedTransactionCallback(XactEvent event, void *arg)
 			dlist_init(&InProgressTransactions);
 			activeSetStmts = NULL;
 			CoordinatedTransactionUses2PC = false;
+
+			/*
+			 * Getting here without ExecutorLevel 0 is a bug, however it is such a big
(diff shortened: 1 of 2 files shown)


src/backend/distributed/transaction/transaction_management.c
```diff
@@ -290,6 +290,13 @@ CoordinatedTransactionCallback(XactEvent event, void *arg)
 			dlist_init(&InProgressTransactions);
 			activeSetStmts = NULL;
 			CoordinatedTransactionUses2PC = false;
+
```
onderkalaci: @thanodnl I get into this assertion failure: https://github.com/citusdata/citus/issues/3535

Are you OK to remove the assert as Marco suggested?

PR #3314, opened 2019-12-17
Split propagate extension test and add alternative output

onurctirtir:
* This pr aims to add an alternative output file for "escape extension name" tests.  
* Before doing this, we first picked those tests from existing regression test file (`propagate_extension_commands.sql`) and moved them to a new regression test file.

This is because, we need `uuid` library installed to perform "escape extension name tests" but it is not maintained well according to [PostgreSQL Documentation](https://www.postgresql.org/docs/12/uuid-ossp.html)

### DIFF
--- a/src/test/regress/expected/escape_extension_name_0.out
+++ b/src/test/regress/expected/escape_extension_name_0.out
@@ -1,0 +1,83 @@
+CREATE SCHEMA "extension'test";
+-- use  a schema name with escape character
+SET search_path TO "extension'test";
+-- test if citus can escape the extension name
+-- this output will help us to understand why we have alternative outputs for this test
+-- print true if uuid-ossp is available false otherwise
+SELECT CASE WHEN COUNT(*) > 0
+  THEN 'CREATE EXTENSION "uuid-ossp"'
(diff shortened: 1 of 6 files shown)


src/test/regress/expected/escape_extension_name_0.out
```diff
@@ -0,0 +1,72 @@
...
+\gset
+:uuid_present_command;
+ uuid_ossp_present 
+-------------------
+ f
+(1 row)
+
+-- test if citus can escape the extension name
+CREATE EXTENSION "uuid-ossp";
+ERROR:  could not open extension control file "/home/onurctirtir/.pgenv/pgsql-12.0/share/extension/uuid-ossp.control": No such file or directory
```
marcocitus: Given that your home directory is in here, will this actually work for anyone else? Maybe use a BEGIN .. EXCEPTION block to hide the error message.

onurctirtir: that exactly makes sense, sorry for this

onurctirtir: Updated the test and output files like we did in `custom_aggregate_support.sql`

PR #3316, opened 2019-12-17
Fix handling of empty intermediate results when distributing custom aggregates

### LINKED ISSUE #3219: Aggregate result of empty set should be initcond
Postgres returns `initcond`'s value when there are no input tuples, but our custom aggregate implementation returns `NULL`.

serprex:
DESCRIPTION: Fix empty intermediate results returning `NULL` when distributing custom aggregates

This leverages `AggGetAggref` to pull the first argument's value. If a user is directly using these aggregates they may not be passing a constant, in that case we continue to return `NULL`

The issue went deeper than I'd initially assessed. `NULL`s were being returned from shards without intermediate results, which could taint the whole result set. This is what causes the change to the `FILTER` test results

Fixes #3219

### DIFF
--- a/src/backend/distributed/utils/aggregate_utils.c
+++ b/src/backend/distributed/utils/aggregate_utils.c
@@ -55,8 +55,10 @@ static HeapTuple GetProcForm(Oid oid, Form_pg_proc *form);
 static HeapTuple GetTypeForm(Oid oid, Form_pg_type *form);
 static void * pallocInAggContext(FunctionCallInfo fcinfo, size_t size);
 static void aclcheckAggregate(ObjectType objectType, Oid userOid, Oid funcOid);
+static Datum GetAggInitVal(Datum textInitVal, Oid transtype);
 static void InitializeStypeBox(FunctionCallInfo fcinfo, StypeBox *box, HeapTuple aggTuple,
 							   Oid transtype);
+static StypeBox * TryCreateStypeBoxFromFcinfoAggref(FunctionCallInfo fcinfo);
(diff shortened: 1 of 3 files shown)


src/backend/distributed/utils/aggregate_utils.c
```diff
@@ -171,27 +194,55 @@ InitializeStypeBox(FunctionCallInfo fcinfo, StypeBox *box, HeapTuple aggTuple, O
...
+ * when we receive no intermediate rows.
+ *
+ * Returns NULL if the Aggref isn't our expected shape.
+ */
+static StypeBox *
+TryCreateStypeBoxFromFcinfoAggref(FunctionCallInfo fcinfo)
+{
+	Aggref *aggref = AggGetAggref(fcinfo);
+	if (aggref != NULL && aggref->args != NIL)
+	{
```
marcocitus: Would be nice to do early returns.

src/test/regress/expected/aggregate_support.out
```diff
@@ -75,11 +75,11 @@ select key, sum2(val) filter (where valf < 5), sum2_strict(val) filter (where va
  key | sum2 | sum2_strict 
 -----+------+-------------
    1 |      |            
-   2 |      |          10
-   3 |      |            
-   5 |      |            
-   6 |      |            
-   7 |      |            
+   2 |   10 |          10
```
marcocitus: hmz, why was this NULL before?

serprex: Workers returned `NULL` instead of `0` when there were no rows passing filter, which then caused `NULL` to be combined instead of `0`

PR #3318, opened 2019-12-18
Implement fetch_intermediate_results

pykello:
Implements a UDF to fetch multiple result files from a node over a single connection:

```sql
SELECT fetch_intermediate_results(ARRAY['result_id_1', 'result_id_2', ...]::text[], nodehost, nodeport);
```

Also implements the `COPY result_id TO ... WITH (format result);` syntax to aid the above UDF.

This is a step towards implementing `INSERT INTO ... SELECT` with repartitiong.

### DIFF
--- a/src/backend/distributed/executor/intermediate_results.c
+++ b/src/backend/distributed/executor/intermediate_results.c
@@ -23,6 +23,7 @@
 #include "distributed/intermediate_results.h"
 #include "distributed/master_metadata_utility.h"
 #include "distributed/metadata_cache.h"
+#include "distributed/multi_client_executor.h"
 #include "distributed/multi_executor.h"
 #include "distributed/remote_commands.h"
 #include "distributed/transmit.h"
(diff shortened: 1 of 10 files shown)


src/test/regress/expected/intermediate_results.out
```diff
@@ -413,6 +418,90 @@ EXPLAIN (COSTS ON) SELECT * FROM read_intermediate_results(ARRAY['stored_squares
...
+ERROR:  result "squares_1" does not exist
+ROLLBACK TO SAVEPOINT s1;
+-- fetch from worker 2 should fail
+SELECT * FROM fetch_intermediate_results(ARRAY['squares_1', 'squares_2']::text[], 'localhost', :worker_2_port);
+ERROR:  could not open file "base/pgsql_job_cache/10_0_200/squares_1.data": No such file or directory
+CONTEXT:  while executing command on localhost:57638
+ROLLBACK TO SAVEPOINT s1;
+-- still, results aren't available on coordinator yet
+SELECT * FROM read_intermediate_results(ARRAY['squares_1', 'squares_2']::text[], 'binary') AS res (x int, x2 int);
+ERROR:  COPY file signature not recognized
```
pykello: strange error, I think because of an empty file.

src/backend/distributed/executor/intermediate_results.c
```diff
@@ -769,3 +786,158 @@ ReadIntermediateResultsIntoFuncOutput(FunctionCallInfo fcinfo, char *copyFormat,
...
+
+	CheckCitusVersion(ERROR);
+
+	if (resultCount == 0)
+	{
+		PG_RETURN_INT64(0);
+	}
+
+	if (!IsMultiStatementTransaction())
+	{
```
marcocitus: Should this also check for a distributed transaction ID?

src/backend/distributed/executor/intermediate_results.c
```diff
@@ -769,3 +786,158 @@ ReadIntermediateResultsIntoFuncOutput(FunctionCallInfo fcinfo, char *copyFormat,
...
+	}
+
+	PQclear(result);
+
+	char *localPath = QueryResultFileName(resultId);
+	File fileDesc = FileOpenForTransmit(localPath, fileFlags, fileMode);
+	FileCompat fileCompat = FileCompatFromFileStart(fileDesc);
+
+	while (true)
+	{
```
marcocitus: Might be nice to move this loop into a separate function.

PR #3323, opened 2019-12-18
Support any inner join on a reference table

### LINKED ISSUE #3079: Cartesian product problem
Hello, thank you for this amazing work. I have a question, suppose that we have two tables: one table contains 10 rows and it is the distributed table . The other table contains 5 rows and it is a reference table. So, when I use this query: select * from T1, T2 it should work because If we have 2 worker nodes, then every worker will take part of the rows that are different than the other node. So the Cartesian product can be applicable, all what you need is to push the results to the caller. Correct me if I have wrong understanding about this issue. Thanks for your help.

### LINKED ISSUE #3198: Support non-equi-joins to reference tables for repartition joins
Currently we only support equi-joins for repartition joins. This restriction can be removed for reference joins since any join can be evaluated for reference joins.

Non-equi-joins are not recognised as a join by `IsJoinClause`. The following planners need to be changed to not allow planning for non-equi-joins:
 - `SingleRepartitionJoin`
 - `DualRepartitionJoin`
 - `LocalJoin`

This would allow the reference join's to be able to plan the join correctly.

JelteF:
DESCRIPTION: Support any inner join on a reference table

This PR works by doing two things:
1. Expand the notion of a join condition to any expression that contains
   columns from two or more tables.
2. Support cartesian products on reference tables.

Cartesian products on reference tables are considered in the join order planner
as the least desirable join (except for normal cartesian products). That way
they will be done at the end of the join. This is preferable since the
cartesian product multiplies the rows. By doing it at the end at least these
multiplications of rows will not be sent over the network when doing
repartitioning, only when sending to the master.

Fixes #3079
Fixes #3198

### DIFF
--- a/src/backend/distributed/planner/multi_join_order.c
+++ b/src/backend/distributed/planner/multi_join_order.c
@@ -212,7 +212,6 @@ ExtractLeftMostRangeTableIndex(Node *node, int *rangeTableIndex)
 static bool
 JoinOnColumns(Var *currentColumn, Var *candidateColumn, List *joinClauseList)
 {
-	ListCell *joinClauseCell = NULL;
 	if (currentColumn == NULL || candidateColumn == NULL)
 	{
 		/*
@@ -222,15 +221,16 @@ JoinOnColumns(Var *currentColumn, Var *candidateColumn, List *joinClauseList)
 		return false;
 	}
 
-	foreach(joinClauseCell, joinClauseList)
+	Node *joinClause = NULL;
+	foreach_ptr(joinClause, joinClauseList)
 	{
-		OpExpr *joinClause = castNode(OpExpr, lfirst(joinClauseCell));
-		Var *leftColumn = LeftColumnOrNULL(joinClause);
-		Var *rightColumn = RightColumnOrNULL(joinClause);
-		if (!OperatorImplementsEquality(joinClause->opno))
+		if (!NodeIsEqualsOpExpr(joinClause))
 		{
 			continue;
 		}
+		OpExpr *joinClauseOpExpr = castNode(OpExpr, joinClause);
+		Var *leftColumn = LeftColumnOrNULL(joinClauseOpExpr);
+		Var *rightColumn = RightColumnOrNULL(joinClauseOpExpr);
 
 		/*
 		 * Check if both join columns and both partition key columns match, since the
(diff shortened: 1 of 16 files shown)


src/backend/distributed/planner/multi_join_order.c
```diff
@@ -781,17 +797,38 @@ static JoinOrderNode *
...
-	char leftPartitionMethod = PartitionMethod(currentJoinNode->tableEntry->relationId);
-	bool performReferenceJoin = false;
-
 	if (applicableJoinCount <= 0)
 	{
 		return NULL;
 	}
 
+	JoinOrderNode *nextJoinNode =
+		CartesianProductReferenceJoin(currentJoinNode, candidateTable,
```
thanodnl: Can we refactor this so that `ReferenceJoin` is not calling to `CartesianProductReferenceJoin`? It is a bit strange to read and understand the logic here.

Instead I would reuse `IsReferenceJoin` to derive `performReferenceJoin` and call the correct `MakeJoinOrderNode` call.

src/backend/distributed/planner/multi_join_order.c
```diff
@@ -1006,17 +1043,25 @@ SinglePartitionJoin(JoinOrderNode *currentJoinNode, TableEntry *candidateTable,
...
-		Var *rightColumn = RightColumnOrNULL(applicableJoinClause);
+		if (!IsA(applicableJoinClause, OpExpr))
+		{
+			continue;
+		}
+		OpExpr *applicableJoinOpExpr = castNode(OpExpr, applicableJoinClause);
+		if (!OperatorImplementsEquality(applicableJoinOpExpr->opno))
+		{
+			continue;
+		}
```
thanodnl: This portion is twice in this PR, without any good documentation.

Instead; could we extract this to a function `IsEquiJoinClause`? This makes the code readable as
```C
if (!IsEquiJoinClause(applicableJoinClause))
{
    continue;
}
```

src/backend/distributed/planner/multi_join_order.c
```diff
@@ -222,11 +225,20 @@ JoinOnColumns(Var *currentColumn, Var *candidateColumn, List *joinClauseList)
...
+	foreach_ptr(joinClause, joinClauseList)
 	{
-		OpExpr *joinClause = castNode(OpExpr, lfirst(joinClauseCell));
-		Var *leftColumn = LeftColumnOrNULL(joinClause);
-		Var *rightColumn = RightColumnOrNULL(joinClause);
+		if (!IsA(joinClause, OpExpr))
+		{
+			continue;
+		}
+		OpExpr *joinClauseOpExpr = castNode(OpExpr, lfirst(joinClauseCell));
```
thanodnl:
```suggestion
		OpExpr *joinClauseOpExpr = castNode(OpExpr, joinClause);
```

PR #3324, opened 2019-12-18
Pull up intermediate rows to coordinator for aggregates we cannot push down

### LINKED ISSUE #3212: ERROR: XX000: unsupported aggregate function string_agg
Hi,

I am recently migrating my company's single-server Postgres database to servers with a Citus deployment. 

One of our old queries uses `string_agg()` and it works well in a single-server Postgres environment. The query looks like this:

```
SELECT nw.id, nw.word, STRING_AGG(nwam.author_name, '、、' ORDER BY nwam.id) AS customers, COUNT(*) AS num, STRING_AGG(nwam.dialogue_id::text, '、、' ORDER BY nwam.id) AS dialogue_ids, STRING_AGG(nwam.content_id::text, '、、' ORDER BY nwam.id) AS content_ids 
FROM conf.negative_word AS nw 
INNER JOIN conf.negative_word_author_map AS nwam 
ON nw.id = nwam.nw_id 
WHERE nw.status = 1 AND nw.is_dangerous = 1 AND nwam.author_type = 1 AND nwam.publish_time <= 1574265599 AND nwam.publish_time >= 1571587200 
GROUP BY nw.id, nw.word ORDER BY num DESC;
```

In our Citus environment, `conf.negative_word_author_map` is a distributed table and `conf.negative_word` is a reference table. 

When I tried to execute this query on Citus coordinator node, an `unsupported aggregate function string_agg` error was thrown. The full DEBUG log is copied below:

```
DEBUG:  00000: StartTransaction(1) name: unnamed; blockState: DEFAULT; state: INPROGRESS, xid/subid/cid: 0/1/0
LOCATION:  ShowTransactionStateRec, xact.c:5167
LOG:  00000: statement: SELECT nw.id, nw.word, string_agg(nwam.author_name, '、、' ORDER BY nwam.id) AS customers, COUNT(*) AS num, string_agg(nwam.dialogue_id::text, '、、' ORDER BY nwam.id) AS dialogue_ids, string_agg(nwam.content_id::text, '、、' ORDER BY nwam.id) AS content_ids 
FROM conf.negative_word AS nw 
INNER JOIN conf.negative_word_author_map AS nwam 
ON nw.id = nwam.nw_id 
WHERE nw.status = 1 AND nw.is_dangerous = 1 AND nwam.author_type = 1 AND nwam.publish_time <= 1574265599 AND nwam.publish_time >= 1571587200 
GROUP BY nw.id, nw.word ORDER BY num DESC;
LOCATION:  exec_simple_query, postgres.c:975
DEBUG:  0A000: Router planner cannot handle multi-shard select queries
LOCATION:  PlanRouterQuery, multi_router_planner.c:2008
ERROR:  XX000: unsupported aggregate function string_agg
LOCATION:  GetAggregateType, multi_logical_optimizer.c:2907
```

To my understanding, `string_agg()` should not be a concern of Citus and can be passed down to the workers and handled by Postgres. This is why I do not understand it is "unsupported", as shown in the error message.

I will fully trace the source code and try to understand what was happening later when I have more time. But I am still a rookie on Citus, if anyone had any clue about what was causing this issue, or noticed I was making careless mistakes, please let me know. Thank you so much!

serprex:
DESCRIPTION: `SET citus.coordinator_aggregation_strategy to 'row-gather'` will enable pulling up intermediate rows to the coordinator for unsupported aggregates

There were two aggregation strategies suggested for coordinator aggregation. I'll refer to them as *row-gather* & *array-fold*. **This PR only implements row-gather**, but I'll detail both

row-gather detects when we would otherwise fail to push down an aggregate & disables aggregate push down for the entire query, instead removing `GROUP BY` from the worker query in favor of selecting each argument as a separate row. Then on the coordinator we execute the query with args set to use the results from the worker. We implement `FILTER` by having it computed on the worker as if it were a boolean arg

`array-fold` pushes down `agg(x, y order by z) filter (where expr)` as `array_agg(ROW(x,y,z)) filter (where expr)`, then on the coordinator executes `fold_agg_over_array(agg, array_cat_agg(rows), false, {true, true, false}, {order-by-z}, NULL::returntypeofagg)` where the `false` tells us whether to dedupe rows, `{true, true, false}` tells us which columns are arguments to the aggregate, & `{order-by-z}` is an array of descriptions on how to sort the input. We use this to effectively apply the aggregate to an array of the intermediate rows

`array-fold` is mostly implemented on another branch: https://github.com/citusdata/citus/commits/aggregate-on-coordinator

Performance differences:

- array-fold pushes down grouping to workers, row-gather does not
- array-fold pushes down `FILTER`, row-gather computes filtered rows
- array-fold doesn't disable other aggregates from being pushed down
- array-fold falls apart once it starts to hit memory bounds, row-gather smoothly taps into postgres's external memory management
- array-fold cannot handle [ordered-set aggregates](https://www.postgresql.org/docs/9.4/functions-aggregate.html#FUNCTIONS-ORDEREDSET-TABLE) unless we start copying over private structs from postgres. This is related to the `AggGetAggref` machinery which we can only fake so much
- array-fold's code is more complicated than row-gather

Given the trade-offs, there are two potential options for implementing coordinator aggregation:

- only implement row-gather
- implement both row-gather & array-fold as options for `citus.coordinator_aggregation_strategy`, having row-gather used for ordered-set aggregates when `array-fold` is opted for

Both options implement row-gather, so it seems best to implement row-gather for now & then benchmark further to decide if array-fold is worth implementing. My initial testing showed that array-fold had definite advantages when `GROUP BY` was used with small groups. However when I increased the group size by a factor of 10, row-gather took a bit more than 10x longer while array-fold took 100x longer before crashing my laptop

My test case:

```
create table entries(
  id serial primary key,
  amount float8 not null,
  amount2 float8 not null
);
insert into entries(amount, amount2)
select x * 0.5, x * 2.5
from generate_series(1, 10000) x; -- increase generate_series to add more groups
select create_distributed_table('entries', 'id');
select floor(e1.amount/100), corr(e1.amount * 2, e1.amount2 - 2), corr(e1.amount + 1, e1.amount2 * 2 order by e2), corr(e1.amount, e1.amount2)
FROM entries e1, generate_series(1, 1000) e2 GROUP BY floor(e1.amount/100); -- increase generate_series to increase group size
```

Fixes #3212

### DIFF
--- a/src/backend/distributed/planner/extended_op_node_utils.c
+++ b/src/backend/distributed/planner/extended_op_node_utils.c
@@ -42,7 +42,8 @@ static bool ShouldPullDistinctColumn(bool repartitionSubquery,
  * value should be used in a read-only manner.
  */
 ExtendedOpNodeProperties
-BuildExtendedOpNodeProperties(MultiExtendedOp *extendedOpNode)
+BuildExtendedOpNodeProperties(MultiExtendedOp *extendedOpNode, bool
+							  pullUpIntermediateRows)
 {
 	ExtendedOpNodeProperties extendedOpNodeProperties;
 
@@ -75,6 +76,7 @@ BuildExtendedOpNodeProperties(MultiExtendedOp *extendedOpNode)
 		hasNonPartitionColumnDistinctAgg;
 	extendedOpNodeProperties.pullDistinctColumns = pullDistinctColumns;
 	extendedOpNodeProperties.pushDownWindowFunctions = pushDownWindowFunctions;
+	extendedOpNodeProperties.pullUpIntermediateRows = pullUpIntermediateRows;
 
 	return extendedOpNodeProperties;
 }
--- a/src/backend/distributed/planner/multi_logical_optimizer.c
+++ b/src/backend/distributed/planner/multi_logical_optimizer.c
@@ -27,8 +27,10 @@
 #include "distributed/citus_nodes.h"
 #include "distributed/citus_ruleutils.h"
 #include "distributed/colocation_utils.h"
+#include "distributed/errormessage.h"
 #include "distributed/extended_op_node_utils.h"
 #include "distributed/function_utils.h"
+#include "distributed/listutils.h"
 #include "distributed/metadata_cache.h"
 #include "distributed/multi_logical_optimizer.h"
 #include "distributed/multi_logical_planner.h"
@@ -59,19 +61,19 @@
 /* Config variable managed via guc.c */
 int LimitClauseRowFetchCount = -1; /* number of rows to fetch from each task */
 double CountDistinctErrorRate = 0.0; /* precision of count(distinct) approximate */
-
+int CoordinatorAggregationStrategy = COORDINATOR_AGGREGATION_ROW_GATHER;
 
 typedef struct MasterAggregateWalkerContext
 {
+	const ExtendedOpNodeProperties *extendedOpNodeProperties;
 	AttrNumber columnId;
-	bool pullDistinctColumns;
 } MasterAggregateWalkerContext;
 
 typedef struct WorkerAggregateWalkerContext
 {
+	const ExtendedOpNodeProperties *extendedOpNodeProperties;
 	List *expressionList;
 	bool createGroupByClause;
-	bool pullDistinctColumns;
 } WorkerAggregateWalkerContext;
 
 
@@ -189,7 +191,8 @@ static void ParentSetNewChild(MultiNode *parentNode, MultiNode *oldChildNode,
 static void ApplyExtendedOpNodes(MultiExtendedOp *originalNode,
 								 MultiExtendedOp *masterNode,
 								 MultiExtendedOp *workerNode);
-static void TransformSubqueryNode(MultiTable *subqueryNode);
+static void TransformSubqueryNode(MultiTable *subqueryNode, bool
+								  requiresIntermediateRowPullUp);
 static MultiExtendedOp * MasterExtendedOpNode(MultiExtendedOp *originalOpNode,
 											  ExtendedOpNodeProperties *
 											  extendedOpNodeProperties);
@@ -267,12 +270,18 @@ static Const * MakeIntegerConst(int32 integerValue);
 static Const * MakeIntegerConstInt64(int64 integerValue);
 
 /* Local functions forward declarations for aggregate expression checks */
-static void ErrorIfContainsUnsupportedAggregate(MultiNode *logicalPlanNode);
-static void ErrorIfUnsupportedArrayAggregate(Aggref *arrayAggregateExpression);
-static void ErrorIfUnsupportedJsonAggregate(AggregateType type,
-											Aggref *aggregateExpression);
-static void ErrorIfUnsupportedAggregateDistinct(Aggref *aggregateExpression,
-												MultiNode *logicalPlanNode);
+static bool RequiresIntermediateRowPullUp(MultiNode *logicalPlanNode);
+static DeferredErrorMessage * DeferErrorIfContainsNonPushdownableAggregate(
+	MultiNode *logicalPlanNode);
+static DeferredErrorMessage * DeferErrorIfUnsupportedArrayAggregate(
+	Aggref *arrayAggregateExpression);
+static DeferredErrorMessage * DeferErrorIfUnsupportedJsonAggregate(AggregateType type,
+																   Aggref *
+																   aggregateExpression);
+static DeferredErrorMessage * DeferErrorIfUnsupportedAggregateDistinct(
+	Aggref *aggregateExpression,
+	MultiNode *
+	logicalPlanNode);
 static Var * AggregateDistinctColumn(Aggref *aggregateExpression);
 static bool TablePartitioningSupportsDistinct(List *tableNodeList,
 											  MultiExtendedOp *opNode,
@@ -315,16 +324,29 @@ MultiLogicalPlanOptimize(MultiTreeRoot *multiLogicalPlan)
 	ListCell *collectNodeCell = NULL;
 	ListCell *tableNodeCell = NULL;
 	MultiNode *logicalPlanNode = (MultiNode *) multiLogicalPlan;
-
+	bool requiresIntermediateRowPullUp = RequiresIntermediateRowPullUp(logicalPlanNode);
 	List *extendedOpNodeList = FindNodesOfType(logicalPlanNode, T_MultiExtendedOp);
 	MultiExtendedOp *extendedOpNode = (MultiExtendedOp *) linitial(extendedOpNodeList);
 	ExtendedOpNodeProperties extendedOpNodeProperties = BuildExtendedOpNodeProperties(
-		extendedOpNode);
+		extendedOpNode, requiresIntermediateRowPullUp);
 
-	if (!extendedOpNodeProperties.groupedByDisjointPartitionColumn)
+	if (!extendedOpNodeProperties.groupedByDisjointPartitionColumn &&
+		!extendedOpNodeProperties.pullUpIntermediateRows)
 	{
-		/* check that we can optimize aggregates in the plan */
-		ErrorIfContainsUnsupportedAggregate(logicalPlanNode);
+		DeferredErrorMessage *aggregatePushdownError =
+			DeferErrorIfContainsNonPushdownableAggregate(logicalPlanNode);
+
+		if (aggregatePushdownError != NULL)
+		{
+			if (CoordinatorAggregationStrategy == COORDINATOR_AGGREGATION_DISABLED)
+			{
+				RaiseDeferredError(aggregatePushdownError, ERROR);
+			}
+			else
+			{
+				extendedOpNodeProperties.pullUpIntermediateRows = true;
+			}
+		}
 	}
 
 	/*
@@ -382,7 +404,6 @@ MultiLogicalPlanOptimize(MultiTreeRoot *multiLogicalPlan)
 	 * clause list to the worker operator node. We then push the worker operator
 	 * node below the collect node.
 	 */
-
 	MultiExtendedOp *masterExtendedOpNode =
 		MasterExtendedOpNode(extendedOpNode, &extendedOpNodeProperties);
 	MultiExtendedOp *workerExtendedOpNode =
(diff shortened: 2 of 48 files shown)


src/include/distributed/extended_op_node_utils.h
```diff
@@ -28,6 +28,7 @@ typedef struct ExtendedOpNodeProperties
 	bool hasNonPartitionColumnDistinctAgg;
 	bool pullDistinctColumns;
 	bool pushDownWindowFunctions;
+	bool pullUpIntermediateRows;
```
serprex: This seems like the wrong place to store this since it isn't being set by `BuildExtendedOpNodeProperties`. May make sense to push this logic further back in planner, especially since it seems like this can handle window function pull up, but the decision is currently tied to aggregation push down support, which is all in multi_logical_optimization

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -1494,16 +1546,55 @@ static Expr *
...
+			targetEntry->expr = (Expr *)
+								makeVar(masterTableId, walkerContext->columnId,
+										exprType((Node *) targetEntry->expr),
+										exprTypmod((Node *) targetEntry->expr),
+										exprCollation((Node *) targetEntry->expr),
+										columnLevelsUp);
+			walkerContext->columnId++;
+		}
+
+		aggregate->aggdirectargs = NIL;
```
serprex: Look into whether we can skip pushing down aggdirectargs in favor of having it all kept up in coordinator

src/include/distributed/extended_op_node_utils.h
marcocitus: For window functions we would still want to push down GROUP BY.

src/backend/distributed/shared_library_init.c
```diff
@@ -1006,6 +1012,17 @@ RegisterCitusConfigVariables(void)
 		GUC_STANDARD,
 		NULL, NULL, NULL);
 
+	DefineCustomEnumVariable(
+		"citus.coordinator_aggregation_strategy",
+		gettext_noop("Sets the strategy for when an aggregate cannot be pushed down."),
+		NULL,
+		&CoordinatorAggregationStrategy,
+		COORDINATOR_AGGREGATION_DISABLED,
```
marcocitus: If row-gather works reliably then users will likely be happier if we make it the default.

serprex: Bit worried that users will stumble into this when they try to do something like `sum(distinct)`, but updated default

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -380,6 +406,8 @@ MultiLogicalPlanOptimize(MultiTreeRoot *multiLogicalPlan)
 	ExtendedOpNodeProperties extendedOpNodeProperties = BuildExtendedOpNodeProperties(
 		extendedOpNode);
 
+	extendedOpNodeProperties.pullUpIntermediateRows = requiresIntermediateRowPullUp;
```
marcocitus: I wonder whether we should have an enum that defines where the boundary is, e.g.

```c
/* push down only the FROM .. WHERE .. clauses */
PUSH_DOWN_JOIN_TREE,

/* partially push down GROUP BY and/or aggregates, merge on the coordinator */
PUSH_DOWN_AGGREGATION_PARTIAL,

/* fully push down GROUP BY */
PUSH_DOWN_AGGREGATION_FULL,

/* fully push down window functions */
PUSH_DOWN_WINDOW_FUNCTIONS,

/* fully push down ORDER BY and/or LIMIT */
PUSH_DOWN_ORDER_BY_LIMIT
```

Otherwise this might get pretty clunky in combination with #3307.

For reference: https://jvns.ca/blog/2019/10/03/sql-queries-don-t-start-with-select/

serprex: That'd be a good follow up PR. Would like to get #3307 merged & then hopefully get pushing down arbitrary aggregates when `GROUP BY` is on a partition key so that people at least avoid this fallback when they're grouping by something which intuitively should be pushed down

src/backend/distributed/shared_library_init.c
marcocitus: You're probably right, but would be preferable to use max_intermediate_result_size. That has so far proven to not be a major annoyance.

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -315,9 +324,26 @@ MultiLogicalPlanOptimize(MultiTreeRoot *multiLogicalPlan)
 	ListCell *collectNodeCell = NULL;
 	ListCell *tableNodeCell = NULL;
 	MultiNode *logicalPlanNode = (MultiNode *) multiLogicalPlan;
+	bool requiresIntermediateRowPullUp = RequiresIntermediateRowPullUp(logicalPlanNode);
```
marcocitus: Other, similar properties (e.g. pullDistinctColumns) are determined in BuildExtendedOpNodeProperties. Should we call this from there for consistency?

src/backend/distributed/planner/extended_op_node_utils.c
```diff
@@ -77,6 +78,7 @@ BuildExtendedOpNodeProperties(MultiExtendedOp *extendedOpNode)
 		hasNonPartitionColumnDistinctAgg;
 	extendedOpNodeProperties.pullDistinctColumns = pullDistinctColumns;
 	extendedOpNodeProperties.pushDownWindowFunctions = pushDownWindowFunctions;
+	extendedOpNodeProperties.pullUpIntermediateRows = pullUpIntermediateRows;
```
marcocitus: Do we still need pullDistinctColumns if we have this property? They seem to implement a similar thing.

marcocitus:
We should perhaps exclude constants from pull up if possible:

```sql
SELECT percentile_disc(0.5) within group (order by x) filter (where x > 50) FROM test;
LOG:  issuing SELECT x AS percentile_disc, '0.5'::double precision AS percentile_disc, (x OPERATOR(pg_catalog.>) 50) AS percentile_disc FROM test_102039 test WHERE true
DETAIL:  on server marco@localhost:9701 connectionId: 2
┌─────────────────┐
│ percentile_disc │
├─────────────────┤
│              75 │
└─────────────────┘
(1 row)

Time: 13.080 ms

```

src/backend/distributed/planner/multi_logical_optimizer.c
serprex: No, since this call alone doesn't determine the value

src/backend/distributed/planner/extended_op_node_utils.c
serprex: Created #3343 to remove `pullDistinctColumns`

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -59,19 +61,19 @@
 /* Config variable managed via guc.c */
 int LimitClauseRowFetchCount = -1; /* number of rows to fetch from each task */
 double CountDistinctErrorRate = 0.0; /* precision of count(distinct) approximate */
-
+int CoordinatorAggregationStrategy = COORDINATOR_AGGREGATION_DISABLED;
```
marcocitus: Maybe ROW_GATHER to indicate the default.

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -315,16 +324,29 @@ MultiLogicalPlanOptimize(MultiTreeRoot *multiLogicalPlan)
...
+		extendedOpNode, requiresIntermediateRowPullUp);
 
-	if (!extendedOpNodeProperties.groupedByDisjointPartitionColumn)
+	if (!extendedOpNodeProperties.groupedByDisjointPartitionColumn &&
+		!extendedOpNodeProperties.pullUpIntermediateRows)
 	{
-		/* check that we can optimize aggregates in the plan */
-		ErrorIfContainsUnsupportedAggregate(logicalPlanNode);
+		DeferredErrorMessage *error =
+			DeferErrorIfContainsUnsupportedAggregate(logicalPlanNode);
```
marcocitus: maybe Unsupported -> NonPushdownable or something along those lines?

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -315,16 +324,29 @@ MultiLogicalPlanOptimize(MultiTreeRoot *multiLogicalPlan)
...
-		extendedOpNode);
+		extendedOpNode, requiresIntermediateRowPullUp);
 
-	if (!extendedOpNodeProperties.groupedByDisjointPartitionColumn)
+	if (!extendedOpNodeProperties.groupedByDisjointPartitionColumn &&
+		!extendedOpNodeProperties.pullUpIntermediateRows)
 	{
-		/* check that we can optimize aggregates in the plan */
-		ErrorIfContainsUnsupportedAggregate(logicalPlanNode);
+		DeferredErrorMessage *error =
```
marcocitus: maybe error -> aggregatePushdownError to clarify the intent of the code (what does it mean when it is not NULL)

src/backend/distributed/planner/multi_logical_optimizer.c
marcocitus: Can we move all the logic that determines the value into BuildExtendedOpNodeProperties?

src/backend/distributed/planner/multi_logical_optimizer.c
```diff
@@ -2682,7 +2782,7 @@ GenerateWorkerTargetEntry(TargetEntry *targetEntry, Expr *workerExpression,
 	 */
 	if (targetEntry)
 	{
-		newTargetEntry = copyObject(targetEntry);
+		newTargetEntry = flatCopyTargetEntry(targetEntry);
```
marcocitus: Is this mostly for performance?

src/test/regress/sql/aggregate_support.sql
```diff
@@ -154,5 +158,41 @@ set role notsuper;
...
+select key, sum2(val order by valf), sum2_strict(val order by valf) from aggdata group by key order by key;
+select string_agg(distinct floor(val/2)::text, '|' order by floor(val/2)::text) from aggdata;
+select string_agg(distinct floor(val/2)::text, '|' order by floor(val/2)::text) filter (where val < 5) from aggdata;
+select mode() within group (order by floor(val/2)) from aggdata;
+select percentile_cont(0.5) within group(order by valf) from aggdata;
+select key, percentile_cont(key/10.0) within group(order by val) from aggdata group by key;
+select floor(val/2), corr(valf, valf + val) from aggdata group by floor(val/2) order by 1;
+select floor(val/2), corr(valf, valf + val) from aggdata group by floor(val/2) having corr(valf + val, val) < 1 order by 1;
+select array_agg(val order by valf) from aggdata;
+
```
marcocitus: Could use some tests with user-defined aggregates.

src/backend/distributed/planner/multi_logical_optimizer.c
serprex: Yes

src/test/regress/sql/aggregate_support.sql
serprex: `sum2` & `sum2_strict` are user defined

src/backend/distributed/planner/multi_logical_optimizer.c
serprex: Yes, but that's as far reaching as `GetAggregateType`

src/backend/distributed/shared_library_init.c
```diff
@@ -1006,6 +1012,17 @@ RegisterCitusConfigVariables(void)
 		GUC_STANDARD,
 		NULL, NULL, NULL);
 
+	DefineCustomEnumVariable(
+		"citus.coordinator_aggregation_strategy",
+		gettext_noop("Sets the strategy for when an aggregate cannot be pushed down."),
+		NULL,
```
marcocitus: Could use some more explanation on what the strategies are.

PR #3329, opened 2019-12-20
Partitioned intermediate results

pykello:
Introduces the `partition_query_result` UDF which partitions results of a query into a set of intermediate result files. The signature of this function is:

```sql
SELECT partition_query_result(
    result_prefix text,
    query text,
    partition_column_index int,
    partition_method citus.distribution_type,
    partition_min_values text[],
    partition_max_values text[],
    binaryCopy boolean);
```

It executes the given query and partitions its results into result files named `result_prefix_0`, `result_prefix_1`, ... Number of partitions is decided by length min/max value arrays. `partition_min_values` and `partition_max_values` correspond to values in `minvalue` and `maxvalue` columns of `pg_dist_shard`. If the `partition_method` is `hash`, then min/max value arrays denote the hash ranges, and otherwise they denote value ranges.

This PR mostly glues together the machinery already available in other parts of code for: 1. which shard should contain a given tuple? 2. file management and copy protocol to write intermediate result files, 3. ...

### Performance
Partitioning result of query=`SELECT i, i FROM generate_series(1, 10000000)` (10M rows)
- partition_query_results (4 hash partitioned partitions): 3.6s
- COPY (query) TO '/local/file': 3.8s
- INSERT INTO distributed_table (query) (4 shards): 6.5s

TODO:
- [x] Add more tests

### DIFF
--- a/src/backend/distributed/executor/partitioned_intermediate_results.c
+++ b/src/backend/distributed/executor/partitioned_intermediate_results.c
@@ -1,0 +1,519 @@
+/*-------------------------------------------------------------------------
+ *
+ * partition_intermediate_results.c
+ *   Functions for writing partitioned intermediate results.
+ *
+ * Copyright (c), Citus Data, Inc.
+ *
+ *-------------------------------------------------------------------------
+ */
+#include <sys/stat.h>
+#include <unistd.h>
+
+#include "postgres.h"
+#include "funcapi.h"
+#include "libpq-fe.h"
+#include "miscadmin.h"
+#include "port.h"
+
+#include "access/nbtree.h"
+#include "catalog/pg_am.h"
+#include "catalog/pg_type.h"
+#include "distributed/intermediate_results.h"
+#include "distributed/master_metadata_utility.h"
+#include "distributed/metadata_cache.h"
+#include "distributed/multi_executor.h"
+#include "distributed/pg_dist_shard.h"
+#include "distributed/remote_commands.h"
+#include "distributed/tuplestore.h"
+#include "distributed/worker_protocol.h"
+#include "nodes/makefuncs.h"
+#include "nodes/primnodes.h"
+#include "tcop/pquery.h"
+#include "tcop/tcopprot.h"
+#include "utils/typcache.h"
+
+
+/*
+ * PartitionedResultDestReceiver is used for streaming tuples into a set of
+ * partitioned result files.
+ */
+typedef struct PartitionedResultDestReceiver
+{
+	/* public DestReceiver interface */
+	DestReceiver pub;
+
+	/* partition file $i is stored at file named $resultIdPrefix_$i. */
+	char *resultIdPrefix;
+
+	/* use binary copy or just text copy format? */
+	bool binaryCopy;
+
+	/* used for deciding which partition a shard belongs to. */
+	DistTableCacheEntry *shardSearchInfo;
+
+	MemoryContext perTupleContext;
+
+	/* how does stream tuples look like? */
+	TupleDesc tupleDescriptor;
+
+	/* which column of streamed tuples to use as partition column? */
+	int partitionColumnIndex;
+
+	/* how many partitions do we have? */
+	int partitionCount;
+
+	/*
+	 * Tuples for partition[i] are sent to partitionDestReceivers[i], which
+	 * writes it to a result file.
+	 */
+	DestReceiver **partitionDestReceivers;
+} PartitionedResultDestReceiver;
+
+static Portal StartPortalForQueryExecution(const char *queryString);
+static DistTableCacheEntry * QueryTupleShardSearchInfo(ArrayType *minValuesArray,
+													   ArrayType *maxValuesArray,
+													   char partitionMethod,
+													   Var *partitionColumn);
+static PartitionedResultDestReceiver * CreatePartitionedResultDestReceiver(char *resultId,
+																		   int
+																		   partitionColumnIndex,
+																		   int
+																		   partitionCount,
+																		   TupleDesc
+																		   tupleDescriptor,
+																		   bool binaryCopy,
+																		   DistTableCacheEntry
+																		   *
+																		   shardSearchInfo,
+																		   MemoryContext
+																		   perTupleContext);
+static void PartitionedResultDestReceiverStartup(DestReceiver *dest, int operation,
+												 TupleDesc inputTupleDescriptor);
+static bool PartitionedResultDestReceiverReceive(TupleTableSlot *slot,
+												 DestReceiver *dest);
+static void PartitionedResultDestReceiverShutdown(DestReceiver *destReceiver);
+static void PartitionedResultDestReceiverDestroy(DestReceiver *destReceiver);
+
+/* exports for SQL callable functions */
+PG_FUNCTION_INFO_V1(worker_partition_query_result);
+
+
+/*
+ * worker_partition_query_result executes a query and writes the results into a
+ * set of local files according to the partition scheme and the partition column.
+ */
+Datum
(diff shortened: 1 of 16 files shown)


src/backend/distributed/executor/partitioned_intermediate_results.c
```diff
@@ -0,0 +1,428 @@
...
+	Portal portal = StartPortalForQueryExecution(queryString);
+
+	/* extract the partition column */
+	TupleDesc tupleDescriptor = portal->tupDesc;
+	if (tupleDescriptor == NULL)
+	{
+		elog(ERROR, "no tuple descriptor");
+	}
+
+	if (partitionColumnIndex >= tupleDescriptor->natts)
```
serprex: `>=` should be `>`

src/backend/distributed/executor/partitioned_intermediate_results.c
```diff
@@ -0,0 +1,428 @@
...
+															   partitionColumnIndex);
+	Var *partitionColumn = makeVar(partitionColumnIndex, partitionColumnIndex,
+								   partitionColumnAttr->atttypid,
+								   partitionColumnAttr->atttypmod,
+								   partitionColumnAttr->attcollation, 0);
+	DistTableCacheEntry *shardSearchInfo =
+		QueryTupleShardSearchInfo(minValuesArray, maxValuesArray,
+								  partitionMethod, partitionColumn);
+
+	if (partitionColumnIndex >= tupleDescriptor->natts)
```
serprex: Thsi check was made 15 lines ago

src/backend/distributed/executor/partitioned_intermediate_results.c
```diff
@@ -0,0 +1,428 @@
...
+
+	return (DestReceiver *) resultDest;
+}
+
+
+static void
+PartitionedResultDestReceiverStartup(DestReceiver *dest, int operation,
+									 TupleDesc inputTupleDescriptor)
+{
+	/* nothing to do here */
```
serprex: Call `rStartup` on any non-null entries, since the interface is such that this could be coming from after a call to Shutdown

src/backend/distributed/executor/partitioned_intermediate_results.c
```diff
@@ -0,0 +1,428 @@
...
+	PartitionedResultDestReceiver *partitionedDest =
+		(PartitionedResultDestReceiver *) copyDest;
+	for (int index = 0; index < partitionedDest->partitionCount; index++)
+	{
+		DestReceiver *partitionDest = partitionedDest->partitionDestReceivers[index];
+		if (partitionDest != NULL)
+		{
+			partitionDest->rDestroy(partitionDest);
+		}
+	}
```
serprex: `pfree(partitionDest->partitionDestReceivers)`

src/backend/distributed/executor/partitioned_intermediate_results.c
```diff
@@ -0,0 +1,428 @@
...
+		ereport(ERROR, (errcode(ERRCODE_OBJECT_NOT_IN_PREREQUISITE_STATE),
+						errmsg("could not find shard for partition column "
+							   "value")));
+	}
+
+	int partitionIndex = shardInterval->shardIndex;
+	DestReceiver *partitionDest = partitionedDest->partitionDestReceivers[partitionIndex];
+	if (partitionDest == NULL)
+	{
+		StringInfo resultId = makeStringInfo();
```
serprex: `StringInfoData` avoids an allocation per partition

src/backend/distributed/metadata/metadata_cache.c
```diff
@@ -1330,6 +1309,32 @@ BuildCachedShardList(DistTableCacheEntry *cacheEntry)
...
 
+/*
+ * ErrorIfInconsistentShardIntervals checks if shard intervals are consistent with
+ * our expectations.
+ */
+void
+ErrorIfInconsistentShardIntervals(DistTableCacheEntry *cacheEntry)
+{
+	/*
+	 * If table is hash-partitioned and has shards, there never should be
```
serprex: This comment can be a line shorter now that it has less indentation

src/backend/distributed/executor/partitioned_intermediate_results.c
pykello: > Is there a reason we don't want to create the files eagerly?

We can have a lot of partitions for range partitioned tables, and we don't want to lots of empty files if most of partitions are empty.

pykello:
@serprex @marcocitus I need to add more tests and do some more manual testing, but I think the code itself seems to be what I intended. So please feel free to put comments.

src/backend/distributed/executor/partitioned_intermediate_results.c
```diff
@@ -0,0 +1,515 @@
...
+	}
+
+	FormData_pg_attribute *partitionColumnAttr = TupleDescAttr(tupleDescriptor,
+															   partitionColumnIndex);
+	Var *partitionColumn = makeVar(partitionColumnIndex, partitionColumnIndex,
+								   partitionColumnAttr->atttypid,
+								   partitionColumnAttr->atttypmod,
+								   partitionColumnAttr->attcollation, 0);
+	DistTableCacheEntry *shardSearchInfo =
+		QueryTupleShardSearchInfo(minValuesArray, maxValuesArray,
```
marcocitus: Maybe add `/* construct an artificial DistTableCacheEntry for shard pruning */`

src/backend/distributed/executor/partitioned_intermediate_results.c
```diff
@@ -0,0 +1,515 @@
...
+/*
+ * PartitionedResultDestReceiverDestroy implements the rDestroy interface of
+ * PartitionedResultDestReceiver.
+ */
+static void
+PartitionedResultDestReceiverDestroy(DestReceiver *copyDest)
+{
+	PartitionedResultDestReceiver *partitionedDest =
+		(PartitionedResultDestReceiver *) copyDest;
+	for (int index = 0; index < partitionedDest->partitionCount; index++)
```
marcocitus: we usually give descriptive names like partitionIndex

src/backend/distributed/sql/udfs/partition_query_result/latest.sql
```diff
@@ -0,0 +1,16 @@
+CREATE OR REPLACE FUNCTION pg_catalog.partition_query_result(
+    result_prefix text,
+    query text,
+    partition_column_index int,
+    partition_method citus.distribution_type,
```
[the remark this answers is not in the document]
serprex: Collation is picked up later with `partitionColumnAttr->attcollation` which should be able to get the right thing from `query` https://github.com/citusdata/citus/pull/3329/files#diff-631a9e0934602ba7a290578ef10e4fddR194

src/backend/distributed/executor/partitioned_intermediate_results.c
```diff
@@ -0,0 +1,515 @@
...
+}
+
+
+/*
+ * QueryTupleShardSearchInfo returns a DistTableCacheEntry which has enough
+ * information so that FindShardInterval() can find the shard corresponding
+ * to a tuple.
+ */
+static DistTableCacheEntry *
+QueryTupleShardSearchInfo(ArrayType *minValuesArray, ArrayType *maxValuesArray,
```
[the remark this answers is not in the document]
pykello: I think this might take about a day. I can do it before code-freeze if everything else is done by then. Created an issue: https://github.com/citusdata/citus/issues/3341

pykello:
Addressed feedback + fixed a bug (I was parsing min/max values with partition column type instead of interval type) + added more tests.

PR #3332, opened 2019-12-21
Use fewer CPU cycles on fast-path planning

onderkalaci:
Fast-path queries are introduced with #2606. The basic idea is that for very simple queries like `SELECT count(*) FROM table WHERE dist_key = X`, we can skip some parts of the distributed planning. The most notable thing to skip is `standard_planner()`, which was already done in #2606.

With this commit, we do some further optimizations. First, we used to call the function which decides whether the query is fast path twice, which can be reduced to one. Second, we used to do shard pruning for every query, now we'll optimize it for some cases. Finally, since the definition of fast-path queries are very strict, we can skip some query traversals.

This is the first of many PRs that'll work on improving the performance of fast-path queries, we prefer to get them in smaller patches like this.

### DIFF
--- a/src/backend/distributed/planner/distributed_planner.c
+++ b/src/backend/distributed/planner/distributed_planner.c
@@ -79,8 +79,7 @@ static DistributedPlan * CreateDistributedPlan(uint64 planId, Query *originalQue
 											   bool hasUnresolvedParams,
 											   PlannerRestrictionContext *
 											   plannerRestrictionContext);
-static void FinalizeDistributedPlan(DistributedPlan *plan, Query *originalQuery,
-									bool fastPathRouterQuery);
+static void FinalizeDistributedPlan(DistributedPlan *plan, Query *originalQuery);
 static void RecordSubPlansUsedInPlan(DistributedPlan *plan, Query *originalQuery);
 static DeferredErrorMessage * DeferErrorIfPartitionTableNotSingleReplicated(Oid
 																			relationId);
@@ -685,8 +684,6 @@ CreateDistributedPlan(uint64 planId, Query *originalQuery, Query *query, ParamLi
 {
 	DistributedPlan *distributedPlan = NULL;
 	bool hasCtes = originalQuery->cteList != NIL;
-	bool fastPathRouterQuery =
-		plannerRestrictionContext->fastPathRestrictionContext->fastPathRouterQuery;
 
 	if (IsModifyCommand(originalQuery))
 	{
@@ -722,7 +719,7 @@ CreateDistributedPlan(uint64 planId, Query *originalQuery, Query *query, ParamLi
 
 		if (distributedPlan->planningError == NULL)
 		{
-			FinalizeDistributedPlan(distributedPlan, originalQuery, fastPathRouterQuery);
+			FinalizeDistributedPlan(distributedPlan, originalQuery);
 
 			return distributedPlan;
 		}
@@ -744,7 +741,7 @@ CreateDistributedPlan(uint64 planId, Query *originalQuery, Query *query, ParamLi
 										   plannerRestrictionContext);
 		if (distributedPlan->planningError == NULL)
 		{
-			FinalizeDistributedPlan(distributedPlan, originalQuery, fastPathRouterQuery);
+			FinalizeDistributedPlan(distributedPlan, originalQuery);
 
 			return distributedPlan;
 		}
@@ -838,7 +835,7 @@ CreateDistributedPlan(uint64 planId, Query *originalQuery, Query *query, ParamLi
 												plannerRestrictionContext);
 		distributedPlan->subPlanList = subPlanList;
 
-		FinalizeDistributedPlan(distributedPlan, originalQuery, fastPathRouterQuery);
+		FinalizeDistributedPlan(distributedPlan, originalQuery);
 
 		return distributedPlan;
 	}
@@ -850,7 +847,7 @@ CreateDistributedPlan(uint64 planId, Query *originalQuery, Query *query, ParamLi
 	 */
 	if (IsModifyCommand(originalQuery))
 	{
-		FinalizeDistributedPlan(distributedPlan, originalQuery, fastPathRouterQuery);
+		FinalizeDistributedPlan(distributedPlan, originalQuery);
 
 		return distributedPlan;
 	}
@@ -883,7 +880,7 @@ CreateDistributedPlan(uint64 planId, Query *originalQuery, Query *query, ParamLi
 	/* distributed plan currently should always succeed or error out */
 	Assert(distributedPlan && distributedPlan->planningError == NULL);
 
-	FinalizeDistributedPlan(distributedPlan, originalQuery, fastPathRouterQuery);
+	FinalizeDistributedPlan(distributedPlan, originalQuery);
 
 	return distributedPlan;
 }
@@ -894,14 +891,13 @@ CreateDistributedPlan(uint64 planId, Query *originalQuery, Query *query, ParamLi
  * currently only implements some optimizations for intermediate result(s) pruning.
  */
 static void
-FinalizeDistributedPlan(DistributedPlan *plan, Query *originalQuery,
-						bool fastPathRouterQuery)
+FinalizeDistributedPlan(DistributedPlan *plan, Query *originalQuery)
 {
 	/*
 	 * Fast path queries, we cannot have any subplans by their definition,
 	 * so skip expensive traversals.
 	 */
-	if (!fastPathRouterQuery)
+	if (!plan->fastPathRouterPlan)
 	{
 		RecordSubPlansUsedInPlan(plan, originalQuery);
 	}
--- a/src/backend/distributed/planner/distributed_planner.c
+++ b/src/backend/distributed/planner/distributed_planner.c
@@ -79,7 +79,8 @@ static DistributedPlan * CreateDistributedPlan(uint64 planId, Query *originalQue
 											   bool hasUnresolvedParams,
 											   PlannerRestrictionContext *
 											   plannerRestrictionContext);
-static void FinalizeDistributedPlan(DistributedPlan *plan, Query *originalQuery);
+static void FinalizeDistributedPlan(DistributedPlan *plan, Query *originalQuery,
+									bool fastPathRouterQuery);
 static void RecordSubPlansUsedInPlan(DistributedPlan *plan, Query *originalQuery);
 static DeferredErrorMessage * DeferErrorIfPartitionTableNotSingleReplicated(Oid
 																			relationId);
@@ -168,7 +169,7 @@ distributed_planner(Query *parse, int cursorOptions, ParamListInfo boundParams)
 		 * don't have a way of doing both things and therefore error out, but do
 		 * have a handy tip for users.
 		 */
-		if (InsertSelectIntoLocalTable(parse))
+		if (!fastPathRouterQuery && InsertSelectIntoLocalTable(parse))
 		{
 			ereport(ERROR, (errmsg("cannot INSERT rows from a distributed query into a "
 								   "local table"),
@@ -183,13 +184,27 @@ distributed_planner(Query *parse, int cursorOptions, ParamListInfo boundParams)
 		 * set, which doesn't break our goals, but, prevents us keeping an extra copy
 		 * of the query tree. Note that we copy the query tree once we're sure it's a
 		 * distributed query.
+		 *
+		 * Since fast-path queries do not through standard planner, we skip unnecessary
+		 * parts in that case.
 		 */
-		rteIdCounter = AssignRTEIdentities(rangeTableList, rteIdCounter);
-		originalQuery = copyObject(parse);
+		if (!fastPathRouterQuery)
+		{
+			rteIdCounter = AssignRTEIdentities(rangeTableList, rteIdCounter);
+			originalQuery = copyObject(parse);
 
-		setPartitionedTablesInherited = false;
-		AdjustPartitioningForDistributedPlanning(rangeTableList,
-												 setPartitionedTablesInherited);
+			setPartitionedTablesInherited = false;
+			AdjustPartitioningForDistributedPlanning(rangeTableList,
+													 setPartitionedTablesInherited);
+		}
+		else
+		{
+			/*
+			 *  We still need to copy the parse tree because the FastPathPlanner
+			 *  modifies it.
+			 */
+			originalQuery = copyObject(parse);
+		}
 	}
 
 	/*
@@ -249,9 +264,12 @@ distributed_planner(Query *parse, int cursorOptions, ParamListInfo boundParams)
 			result = CreateDistributedPlannedStmt(planId, result, originalQuery, parse,
 												  boundParams, plannerRestrictionContext);
 
-			setPartitionedTablesInherited = true;
-			AdjustPartitioningForDistributedPlanning(rangeTableList,
-													 setPartitionedTablesInherited);
+			if (!fastPathRouterQuery)
+			{
+				setPartitionedTablesInherited = true;
+				AdjustPartitioningForDistributedPlanning(rangeTableList,
+														 setPartitionedTablesInherited);
+			}
 		}
 		else
 		{
@@ -667,7 +685,8 @@ CreateDistributedPlan(uint64 planId, Query *originalQuery, Query *query, ParamLi
 {
 	DistributedPlan *distributedPlan = NULL;
 	bool hasCtes = originalQuery->cteList != NIL;
-
+	bool fastPathRouterQuery =
+		plannerRestrictionContext->fastPathRestrictionContext->fastPathRouterQuery;
 
 	if (IsModifyCommand(originalQuery))
 	{
@@ -703,7 +722,7 @@ CreateDistributedPlan(uint64 planId, Query *originalQuery, Query *query, ParamLi
 
 		if (distributedPlan->planningError == NULL)
 		{
-			FinalizeDistributedPlan(distributedPlan, originalQuery);
+			FinalizeDistributedPlan(distributedPlan, originalQuery, fastPathRouterQuery);
 
 			return distributedPlan;
 		}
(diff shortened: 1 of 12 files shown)


src/backend/distributed/planner/distributed_planner.c
```diff
@@ -864,9 +894,17 @@ CreateDistributedPlan(uint64 planId, Query *originalQuery, Query *query, ParamLi
  * currently only implements some optimizations for intermediate result(s) pruning.
  */
 static void
-FinalizeDistributedPlan(DistributedPlan *plan, Query *originalQuery)
+FinalizeDistributedPlan(DistributedPlan *plan, Query *originalQuery,
+						bool fastPathRouterQuery)
```
marcocitus: Could the fastPathRouterQuery by a property of the DistributedPlan to avoid cluttering the signatures?

onderkalaci: I think it makes sense, but keeping it as a part of `restrictionContext` still useful to pass the information during the planning.

Does that work if I keep it in both places? I think having it on the `distributedPlan` can help me in the next PRs (at least to have extra asserts etc)

marcocitus: Yes, keeping in both places seems fine.

onderkalaci:
Reference for future: This increase the performance around ~%4-%5 on top of master when the workers CPU bottlenecked

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -238,9 +263,12 @@ distributed_planner(Query *parse, int cursorOptions, ParamListInfo boundParams)
 			result = CreateDistributedPlannedStmt(planId, result, originalQuery, parse,
 												  boundParams, plannerRestrictionContext);
 
-			setPartitionedTablesInherited = true;
-			AdjustPartitioningForDistributedPlanning(rangeTableList,
-													 setPartitionedTablesInherited);
+			if (!fastPathRouterQuery)
+			{
```
marcocitus: There are a lot of fastPathQuery conditions in this function. Can we write a different variant of distributed_planner that excludes all the non-fast-path bits?

onderkalaci: That sounds a big change with some edge cases like handling `PlannerLevel` or `plannerRestrictionContext` etc.

One simple approach could be to change the following and get rid of the other `if` checks
```C
if (needsDistributedPlanning && FastPathRouterQuery(parse, &distributionKeyValue))
{
	fastPathRouterQuery = true;
 
    /*  we don't need any further checks on the rangeTableList */
    rangeTableList =  NIL;
}
```

And, since `rangeTableList` becomes `NIL`, we would avoid many checks/loops. 

If  we  can find a  better name `rangeTableList` to make it explicit, it'd be even better.

How does that sound?

marcocitus: My main concern is the understandability of this function, which is already hard to follow. Setting `rangeTableList = NIL` does not help understandability.

onderkalaci: My concern on separating the code-path for fast-path is that we'd end-up with handling the same things twice like the hacks for prepared statements, or maintaining `PlannerLevel` needs to be doubled. Or,  we're now putting some information to restriction context in both cases, which we may need to change because in fast-path Push/Pop restriction context doesn't actually make sense.

One approach I  can think of, we can have  a function like `PrepareForDistributedPlanning()` which captures the first ~50 lines of `distributed_planner`  and `PostDistributedPlanning()` to clean-up some of the logic.

How does that sound to you?

(And, one another thing  that I  don't like about this function is that `TryToDelegateFunctionCall` is called in `!needsDistributedPlanning`, which  adds complexity in  my point  of  view.)

JelteF: I refactored this now to have a clearer separation between the 3 planning logic: 
- distributed fast path
- distributed non fast path
- non distributed plan

I think that makes much easier to understand what is happening.

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -66,14 +66,20 @@ static uint64 NextPlanId = 1;
 /* keep track of planner call stack levels */
 int PlannerLevel = 0;
 
+typedef struct DistributedStmtOptions
```
marcocitus: Options doesn't seem like the right name since it captures the whole distributed statement (twice!). Maybe `DistributedPlanningContext` or ..?

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -550,30 +544,111 @@ IsModifyDistributedPlan(DistributedPlan *distributedPlan)
 }
 
 
+/*
+ * FastPathDistributedStmt creates a distributed planned statement using the
+ * FastPathPlanner.
+ */
+static PlannedStmt *
+FastPastDistributedStmt(DistributedStmtOptions *opts, Const *distributionKeyValue)
```
marcocitus: Maybe `PlanFastPathDistributedStmt` or `PlanDistributedStmtViaFastPath`

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -117,17 +123,28 @@ static bool HasUnresolvedExternParamsWalker(Node *expression, ParamListInfo boun
...
 	bool setPartitionedTablesInherited = false;
 	List *rangeTableList = ExtractRangeTableEntryList(parse);
 	int rteIdCounter = 1;
+	bool fastPathRouterQuery = false;
+	Const *distributionKeyValue = NULL;
+	DistributedStmtOptions opts = {
+		.parse = parse,
+		.cursorOptions = cursorOptions,
+		.boundParams = boundParams,
+	};
```
marcocitus: To clean things up a bit further, maybe we can have an early return for `if (!CitusHasBeenLoaded())`?

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -151,10 +168,22 @@ distributed_planner(Query *parse, int cursorOptions, ParamListInfo boundParams)
...
 			needsDistributedPlanning = ListContainsDistributedTableRTE(rangeTableList);
+			if (needsDistributedPlanning)
+			{
+				fastPathRouterQuery = FastPathRouterQuery(parse, &distributionKeyValue);
+			}
 		}
 	}
 
-	if (needsDistributedPlanning)
+	if (fastPathRouterQuery)
```
marcocitus: If found this a bit confusing since we also do need to copy the parse tree for all other distributed queries. 

Maybe:
```c
if (needsDistributedPlanning)
{
    opts.originalQuery = copyObject(parse);
}

if (!fastPathRouterQuery)
{
    ...
}
```

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -550,30 +544,111 @@ IsModifyDistributedPlan(DistributedPlan *distributedPlan)
...
+ */
+static PlannedStmt *
+NonDistributedStmt(DistributedStmtOptions *opts)
+{
+	PlannedStmt *result = standard_planner(opts->parse, opts->cursorOptions,
+										   opts->boundParams);
+
+	bool hasExternParam = false;
+	DistributedPlan *delegatePlan = TryToDelegateFunctionCall(opts->parse,
+															  &hasExternParam);
```
marcocitus: Would it make sense to capture function call delegation as a separate `else if` in `distributed_planner`? 

e.g.

```c
		if (fastPathRouterQuery)
		{
			result = FastPastDistributedStmt(&opts, distributionKeyValue);
		}
		else if (needsDistributedPlanning)
		{
			result = NonFastPathDistributedStmt(&opts, rangeTableList, rteIdCounter);
		}
		else if (TryToDelegateFunctionCall(opts->parse, &hasExternParams, &delegatePlan))
		{
			result = PlanFunctionCallDelegation(result, hasExternParams, delegatePlan);
		}
		else
		{
			result = standard_planner(opts.parse, opts.cursorOptions, opts.boundParams);
		}
```

marcocitus: Would it make sense to capture function call delegation as a separate `else if` in `distributed_planner`? 

e.g.

```c
		if (fastPathRouterQuery)
		{
			result = FastPastDistributedStmt(&opts, distributionKeyValue);
		}
		else if (needsDistributedPlanning)
		{
			result = NonFastPathDistributedStmt(&opts, rangeTableList, rteIdCounter);
		}
		else if (TryToDelegateFunctionCall(opts->parse, &hasExternParams, &delegatePlan))
		{
			result = PlanFunctionCallDelegation(result, hasExternParams, delegatePlan);
		}
		else
		{
			result = standard_planner(opts.parse, opts.cursorOptions, opts.boundParams);
		}
```

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -2022,11 +2041,34 @@ PlanRouterQuery(Query *originalQuery,
 	 * not been called. Thus, restriction information is not avaliable and we do the
 	 * shard pruning based on the distribution column in the quals of the query.
 	 */
-	if (FastPathRouterQuery(originalQuery))
+	if (fastPathRouterQuery)
```
onderkalaci: I think there is some room for improvement here. The function `TargetShardIntervalForFastPathQuery ()` could cover the case when `distributionKeyValue ` exists as well.

In the end, both cases are `FastPath`

src/include/distributed/distributed_planner.h
```diff
@@ -85,10 +85,31 @@ typedef struct JoinRestriction
 	RelOptInfo *outerrel;
 } JoinRestriction;
 
+typedef struct FastPathRestrictionContext
+{
+	bool fastPathRouterQuery;
+
+	/*
```
onderkalaci: This comment only mentions prepared statements, however we also do not set it if the dist key column type doesn't match the const's type. So, either expand the comment or do not mention about the specific cases where distributionKeyValue == NULL

src/backend/distributed/planner/distributed_planner.c
JelteF: This is actually needed like this, because for non fast path queries AssignRTEIdentities needs to be called before the copy. I've improved the comments to make this reasoning more clear.

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -550,30 +562,120 @@ IsModifyDistributedPlan(DistributedPlan *distributedPlan)
...
+	PlannedStmt *result = standard_planner(ctx->parse, ctx->cursorOptions,
+										   ctx->boundParams);
+
+	if (hasExternParam)
+	{
+		/*
+		 * As in CreateDistributedPlannedStmt, try dissuade planner when planning
+		 * potentially failed due to unresolved prepared statement parameters.
+		 */
+		result->planTree->total_cost = FLT_MAX / 100000000;
```
marcocitus: One of the reasons I wanted a separate `PlanDelegatedFunctionCall` is that this hasExternParam variable is misleading. It is only set to true deep down in `TryToDelegateFunctionCall` when we've already established the statement is a distributed function call, but couldn't delegate because of a parameter. Hence this block should move to `PlanDelegatedFunctionCall`.

There's similar logic for cases where we could not use the router planner because of a parameter in `CreateDistributedPlannedStmt`.

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -193,75 +240,39 @@ distributed_planner(Query *parse, int cursorOptions, ParamListInfo boundParams)
...
-		 *
-		 * Otherwise, call into standard planner. This is required because the Citus
-		 * planner relies on both the restriction information per table and parse tree
-		 * transformations made by postgres' planner.
-		 */
+		DistributedPlan *delegatedPlan = NULL;
+		bool hasExternParam = false;
 
-		if (needsDistributedPlanning && FastPathRouterQuery(originalQuery))
+		if (fastPathRouterQuery)
```
onderkalaci: To make the flow a little more explicit, what if we do something like:

```C
	if (fastPathRouterQuery)
	{
		result = PlanFastPathDistributedStmt(&ctx, distributionKeyValue);
	}
	else
	{
		PlannedStmt *localPlan = standard_planner(ctx->parse, ctx->cursorOptions,
											   ctx->boundParams);

		else if (needsDistributedPlanning)
		{
			result = PlanDistributedStmt(&ctx, localPlan, rangeTableList, rteIdCounter);
		}
		else if ((delegatedPlan = TryToDelegateFunctionCall(ctx.parse, 
															&hasExternParam)))
		{
			result = PlanDelegatedFunctionCall(&ctx, localPlan, delegatedPlan, hasExternParam);
		}
		else
		{
			result = localPlan;
		}
	}
``` 

I don't have strong preference, but still wanted to share if you find it easier to follow as well, maybe do it

src/backend/distributed/planner/multi_router_planner.c
onderkalaci: Still valid note, in case you missed

src/include/distributed/distributed_planner.h
```diff
@@ -85,10 +85,31 @@ typedef struct JoinRestriction
...
 	RelationRestrictionContext *relationRestrictionContext;
 	JoinRestrictionContext *joinRestrictionContext;
+
+	/*
+	 * When the query is qualified for fast path, we don't have
+	 * the RelationRestrictionContext and JoinRestrictionContext
+	 * since those are dependent to calling standard_planner.
+	 * Instead, we keep this struct to pass some extra information.
+	 */
+	FastPathRestrictionContext *fastPathRestrictionContext;
```
onderkalaci: If we're going to pass `planningContext` down to the lower functions like `CreateRouterPlan()`, than I'd suggest to move `FastPathRestrictionContext ` into `planningContext`, with the name `FastPathPlanningContext` or such.

I've two points for this:

1) `PlannerRestrictionContext ` is required because `standard_planner` is called and we get into restrictions hooks. In fast-path case, we're not calling standard_planner but still  calling `Push/Pop` planner contexts. We could improve it later, and makes it easier to read.

2) `FastPathRestrictionContext ` does not fit into `PlannerRestrictionContext ` well because they are generated completely differently and could live as different structs.

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -66,14 +66,34 @@ static uint64 NextPlanId = 1;
...
+	 * that change that is made for non fast path router queries though, which
+	 * is the assigning of RTE identities using AssignRTEIdentities. This is
+	 * NULL for non distributed plans, since those don't need it. */
+	Query *originalQuery;
+
+	/* the cursor options given to the planner */
+	int cursorOptions;
+
+	/* the ParamListInfo that is given to the planner */
+	ParamListInfo boundParams;
```
onderkalaci: I guess not,  but should we also add `hasUnresolvedParams` here? Do you have any comments on this?

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -550,30 +562,120 @@ IsModifyDistributedPlan(DistributedPlan *distributedPlan)
...
 	}
 
-	plannerRestrictionContext->joinRestrictionContext =
+	ctx->plannerRestrictionContext->joinRestrictionContext =
 		RemoveDuplicateJoinRestrictions(joinRestrictionContext);
 
 	DistributedPlan *distributedPlan =
-		CreateDistributedPlan(planId, originalQuery, query, boundParams,
-							  hasUnresolvedParams, plannerRestrictionContext);
+		CreateDistributedPlan(planId, ctx->originalQuery, ctx->parse,
```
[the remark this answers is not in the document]
JelteF: I mainly didn't want to go too crazy on code that was so far unrelated to this PR.

src/backend/distributed/planner/distributed_planner.c
JelteF: I'm not sure what you mean with this, what would we return in that case?

src/backend/distributed/planner/distributed_planner.c
JelteF: I changed it differently now then we discussed before in person.
I've added a `plan` field to `DistributedPlanningContext` and I'm setting the total_cost now directly in the `TryToDelegateFunctionCall` function. I think this makes more sense, since `TryToDelegateFunctionCall` can fail for a variety of reasons and in all these cases it should simply fallback to non distributed execution of the function.

src/backend/distributed/planner/distributed_planner.c
```diff
@@ -117,17 +137,31 @@ static bool HasUnresolvedExternParamsWalker(Node *expression, ParamListInfo boun
 static bool IsLocalReferenceTableJoin(Query *parse, List *rangeTableList);
 static bool QueryIsNotSimpleSelect(Node *node);
 static bool UpdateReferenceTablesWithShard(Node *node, void *context);
+static PlannedStmt * PlanFastPathDistributedStmt(DistributedPlanningContext *ctx,
```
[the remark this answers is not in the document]
JelteF: I think it makes sense indeed, done (although I call it the slightly shorter `planContext`)

src/backend/distributed/planner/distributed_planner.c
JelteF: changed it to something close to this now, because I store the localPlan in the planContext.

src/backend/distributed/planner/distributed_planner.c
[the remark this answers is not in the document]
onderkalaci: I'm fine with that, in the future we can create something like [CreateDistributedExecution()](https://github.com/citusdata/citus/blob/51a7e661f9e73f85356e0f798ae5aa2b4f64a4eb/src/backend/distributed/executor/adaptive_executor.c#L902-L963) for the planner where we do common initializations there and use it during planning.

src/include/distributed/distributed_planner.h
```diff
@@ -108,6 +129,33 @@ typedef struct RelationRowLock
...
+	 * NULL for non distributed plans, since those don't need it. */
+	Query *originalQuery;
+
+	/* the cursor options given to the planner */
+	int cursorOptions;
+
+	/* the ParamListInfo that is given to the planner */
+	ParamListInfo boundParams;
+
+	/* Plan created either by standard_planner or by FastPathPlanner */
```
onderkalaci: I think this comment is out-dated, we only  store plans generated by standard_planner?

JelteF: No, the plan created by FastPathPlanner is also put in there. `CreateDistributedPlannedStmt` will finalize whatever plan is in there.

PR #3333, opened 2019-12-21
Make sure to update shard states of partitions on failures

### LINKED ISSUE #3331: query using shards in state==3 for partitioned, sharded tables.
A few months ago I started noticing inconsistent results when querying a partitioned and sharded table between using the parent table with time range limits and querying the underlying partition directly.

for example:
SELECT
  count(*) AS n
FROM myschema.parenttable
WHERE ts >= '2019-12-01 00:00:00'
AND ts < '2019-12-02 00:00:00' 
returns 4377277

When I select by a partition of the parent table I get roughly half the results
SELECT
  count(*) AS n
FROM myschema.parenttable_20191201 
returns 2045170

I reported this on slack on Dec 5th 2:42pm pacific time

Given some clues by Frederic's similar sounding issue (Dec 17th, 7:34am) I realized that I was likely suffering from the same problem.

Further testing shows conclusively that when doing queries against the partitions of a partitioned and sharded table, the system is mistakenly using shards in state 3.

I am running postgresql 11.1 and citusdata 8.1

Manually cleaned up the state 3 shards (so they are all state 1) and both queries now return the correct result.

onderkalaci:
Fixes #3331

In #2389, we've implemented support for partitioned tables with rep > 1.
The implementation is limiting the use of modification queries on the
partitions. In fact, we error out when any partition is modified via
EnsurePartitionTableNotReplicated().

However, we seem to forgot an important case, where the parent table's
partition is marked as INVALID. In that case, at least one of the partition
becomes INVALID. However, we do not mark partitions as INVALID ever.

If the user queries the partition table directly, Citus could happily send
the query to INVALID placements -- which are not marked as INVALID.

This PR fixes it by marking the placements of the partitions as INVALID
as well.

The shard placement repair logic already re-creates all the partitions,
so should be fine in that front.

DESCRIPTION:Prevent wrong results for replicated partitioned tables after failure 

- [x] Add a test
- [ ] Discuss how to backport the change.

### DIFF
--- a/src/backend/distributed/connection/placement_connection.c
+++ b/src/backend/distributed/connection/placement_connection.c
@@ -18,6 +18,7 @@
 #include "distributed/master_protocol.h"
 #include "distributed/metadata_cache.h"
 #include "distributed/distributed_planner.h"
+#include "distributed/multi_partitioning_utils.h"
 #include "distributed/placement_connection.h"
 #include "distributed/relation_access_tracking.h"
 #include "utils/hsearch.h"
@@ -1065,8 +1066,7 @@ CheckShardPlacements(ConnectionShardHashEntry *shardEntry)
 		{
 			uint64 shardId = shardEntry->key.shardId;
 			uint64 placementId = placementEntry->key.placementId;
-			GroupShardPlacement *shardPlacement =
-				LoadGroupShardPlacement(shardId, placementId);
+			ShardPlacement *shardPlacement = LoadShardPlacement(shardId, placementId);
 
 			/*
 			 * We only set shard state if its current state is FILE_FINALIZED, which
@@ -1074,7 +1074,7 @@ CheckShardPlacements(ConnectionShardHashEntry *shardEntry)
 			 */
 			if (shardPlacement->shardState == FILE_FINALIZED)
 			{
-				UpdateShardPlacementState(placementEntry->key.placementId, FILE_INACTIVE);
+				MarkShardPlacementInactive(shardPlacement);
 			}
 		}
 	}
--- a/src/backend/distributed/master/master_metadata_utility.c
+++ b/src/backend/distributed/master/master_metadata_utility.c
@@ -28,6 +28,7 @@
 #include "catalog/pg_namespace.h"
 #include "catalog/pg_type.h"
 #include "commands/extension.h"
+#include "distributed/colocation_utils.h"
 #include "distributed/connection_management.h"
 #include "distributed/citus_nodes.h"
 #include "distributed/listutils.h"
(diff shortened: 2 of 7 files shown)


onderkalaci:
Unfortunately the failure tests are not counted in the code cov, that's why it is failing.

src/backend/distributed/connection/placement_connection.c
```diff
@@ -1075,6 +1075,17 @@ CheckShardPlacements(ConnectionShardHashEntry *shardEntry)
 			if (shardPlacement->shardState == FILE_FINALIZED)
 			{
 				UpdateShardPlacementState(placementEntry->key.placementId, FILE_INACTIVE);
```
marcocitus: Could we move this into a separate function? e.g. `MarkShardPlacementInactive(..)` which also implements the partitioning-specific logic?

src/backend/distributed/master/master_metadata_utility.c
```diff
@@ -1178,6 +1181,66 @@ DeleteShardPlacementRow(uint64 placementId)
...
+ShardPlacementOnNode(ShardInterval *shardInterval, int nodeId)
+{
+	List *placementList = ShardPlacementList(shardInterval->shardId);
+	ListCell *placementCell = NULL;
+
+	foreach(placementCell, placementList)
+	{
+		ShardPlacement *placement = (ShardPlacement *) lfirst(placementCell);
+
+		if (placement->nodeId == nodeId)
```
marcocitus: Would be nicer to use groupId here. Placements do not have a node ID  in the metadata. It just so happens that we applied the same groupId->nodeId mapping to both of these placements.

src/backend/distributed/master/master_metadata_utility.c
```diff
@@ -1178,6 +1181,66 @@ DeleteShardPlacementRow(uint64 placementId)
...
+}
+
+
+/*
+ * ShardPlacementOnNode gets a shardInterval and a nodeId, returns a placement
+ * of the shard on the given node. If no such placement exists, the function
+ * return NULL.
+ */
+static ShardPlacement *
+ShardPlacementOnNode(ShardInterval *shardInterval, int nodeId)
```
marcocitus: Maybe use a shard ID instead of a shard interval, since we don't use any other properties.

src/backend/distributed/master/master_metadata_utility.c
marcocitus: Would perhaps be nice to have a GroupShardPlacementList instead of ShardPlacementList.

PR #3334, opened 2019-12-23
Fix inconsistent shard metadata issue

### LINKED ISSUE #2853: cached metadata for shard 1450000 is inconsistent: DistShardCacheHash  and connection cache  Problem,need more check
reproduce sql.
in citus' test env.( 1 cn (57636), 2 worker(57637,57638)
```
start_metadata_sync_to_node('localhoust',57637);

-- in session 1,run coude below:
ALTER SEQUENCE pg_catalog.pg_dist_shardid_seq RESTART 1450000;
ALTER SEQUENCE pg_catalog.pg_dist_placement_placementid_seq RESTART 1450000;
SET citus.replication_model TO streaming;
CREATE TABLE products_ref (
    product_no integer,
    name text,
    price numeric
);
SELECT create_reference_table('products_ref');
ALTER TABLE products_ref ADD CONSTRAINT p_key PRIMARY KEY(name);
drop table products_ref;

-- now still in that session 1,run below again:
ALTER SEQUENCE pg_catalog.pg_dist_shardid_seq RESTART 1450000;
ALTER SEQUENCE pg_catalog.pg_dist_placement_placementid_seq RESTART 1450000;
SET citus.replication_model TO streaming;
CREATE TABLE products_ref (
    product_no integer,
    name text,
    price numeric
);
SELECT create_reference_table('products_ref');
ALTER TABLE products_ref ADD CONSTRAINT p_key PRIMARY KEY(name);
```

some error:
"cached metadata for shard 1450000 is inconsistent"

i debug the processs. and find some interesting.
when drop the table, cn will run worker_drop_distributed_table on mx node.
but it  will clear and fill , clear and fill  the DistShardCacheHash.
when the drop is finished, DistTableCacheEntry(shardIntervalArrayLength=1, isValid=false)
DistShardCacheHash still hash the entry of the droped table.

now when we resart pg_dist_shardid_seq and reuse the connection. and create the table again.
now he have a new DistTableCacheEntry (isValid=false,shardIntervalArrayLength=0),
so now we can't ResetDistTableCacheEntry because shardIntervalArrayLength=0,we have no
info about the old relid's cache. but when we use the shardid,we could find some into in DistShardCacheHash. then lead to "cached metadata for shard 1450000 is inconsistent"

i think there may be some hidden bug, the code need more check.

### LINKED ISSUE #3038: start/stop_sync_metadata_to_node() cause inconsistent metadata cache in worker nodes.
$title, as demonstrated in the following example. I think `start_sync_metadata_to_node()` should invalidate the metadata on worker nodes.

```sql
SELECT start_metadata_sync_to_node('localhost', 5433);
SELECT start_metadata_sync_to_node('localhost', 5434);

CREATE TABLE t1(a int);
SELECT create_distributed_table('t1', 'a');

CREATE PROCEDURE proc1(x int) LANGUAGE SQL AS $$
DELETE FROM t1;
$$;

select create_distributed_function('proc1(int)');

-- colocate proc1 and t1
update citus.pg_dist_object
set distribution_argument_index = 0, colocationid = pg_dist_partition.colocationid
from pg_proc, pg_dist_partition
where proname = 'proc1' and oid = objid and pg_dist_partition.logicalrelid = 't1'::regclass::oid;

-- the following is routed to worker node, which causes the metadata to be
-- loaded.
CALL proc1(1);

-- Recreate metadata
SELECT stop_metadata_sync_to_node('localhost', 5433);
SELECT stop_metadata_sync_to_node('localhost', 5434);
SELECT start_metadata_sync_to_node('localhost', 5433);
SELECT start_metadata_sync_to_node('localhost', 5434);

DROP TABLE t1;
ERROR:  cached metadata for shard 102008 is inconsistent
HINT:  Reconnect and try again.
CONTEXT:  while executing command on localhost:5434
SQL statement "SELECT master_remove_distributed_table_metadata_from_workers(v_obj.objid, v_obj.schema_name, v_obj.object_name)"
PL/pgSQL function citus_drop_trigger() line 18 at PERFORM

```

marcocitus:
DESCRIPTION: Fixes cached metadata for shard is inconsistent issue

Metadata syncing routinely changes the relation IDs of a shard, which caused an error for open sessions. The main problem is that we do not clear the table cache entry of the old relation, because we will never do a cache lookup for this entry again. 

This PR clears the DistTableCacheEntry of the old relation when first looking up the new relation. More details in the comments.

Fixes #3038
Fixes #2853

### DIFF
--- a/src/backend/distributed/metadata/metadata_cache.c
+++ b/src/backend/distributed/metadata/metadata_cache.c
@@ -1158,6 +1158,25 @@ BuildCachedShardList(DistTableCacheEntry *cacheEntry)
 		}
 
 		heap_close(distShardRelation, AccessShareLock);
+
+		ShardInterval *firstShardInterval = shardIntervalArray[0];
+		bool foundInCache = false;
+		ShardCacheEntry *shardEntry = hash_search(DistShardCacheHash,
+												  &firstShardInterval->shardId, HASH_FIND,
+												  &foundInCache);
+		if (foundInCache && shardEntry->tableEntry != cacheEntry)
+		{
+			/*
+			 * Normally, all shard cache entries for a given DistTableEntry are removed
+			 * before we get here. There is one exception: When a shard changes from
+			 * one relation ID to another. That typically happens during metadata
+			 * syncing when the distributed table is dropped and re-created without
+			 * changing the shard IDs. That means that old relation no longer exists
+			 * and we can safely wipe its entry, which will remove all corresponding
+			 * shard cache entries.
+			 */
+			ResetDistTableCacheEntry(shardEntry->tableEntry);
+		}
 	}
 
 	/* look up value comparison function */
--- a/src/test/regress/sql/multi_cache_invalidation.sql
+++ b/src/test/regress/sql/multi_cache_invalidation.sql
@@ -1,13 +1,23 @@
-SET citus.next_shard_id TO 1601000;
-SET citus.shard_count TO 1;
-SET citus.shard_replication_factor TO 1;
-CREATE TABLE tab9 (test_id integer NOT NULL, data int);
-CREATE TABLE tab10 (test_id integer NOT NULL, data int);
-SELECT create_distributed_table('tab9', 'test_id', 'hash');
-SELECT master_create_distributed_table('tab10', 'test_id', 'hash');
-TRUNCATE tab9;
-UPDATE pg_dist_shard SET logicalrelid = 'tab10'::regclass WHERE logicalrelid = 'tab9'::regclass;
-TRUNCATE tab10;
+-- test that we are tolerant to the relation ID of a shard being changed
+-- and do not cache invalid metadata
+CREATE SCHEMA mci_1;
+CREATE SCHEMA mci_2;
 
-DROP TABLE tab9;
-DROP TABLE tab10;
+SET citus.next_shard_id TO 1601000;
+CREATE TABLE mci_1.test (test_id integer NOT NULL, data int);
+CREATE TABLE mci_2.test (test_id integer NOT NULL, data int);
+SELECT create_distributed_table('mci_1.test', 'test_id');
+SELECT create_distributed_table('mci_2.test', 'test_id', 'append');
+
+INSERT INTO mci_1.test VALUES (1,2), (3,4);
+
+-- move shards into other append-distributed table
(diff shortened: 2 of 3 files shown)


src/test/regress/sql/multi_cache_invalidation.sql
```diff
@@ -1,13 +1,23 @@
...
+
+INSERT INTO mci_1.test VALUES (1,2), (3,4);
+
+-- move shards into other append-distributed table
+SELECT run_command_on_placements('mci_1.test', 'ALTER TABLE %s SET SCHEMA mci_2');
+UPDATE pg_dist_shard
+SET logicalrelid = 'mci_2.test'::regclass, shardminvalue = NULL, shardmaxvalue = NULL
+WHERE logicalrelid = 'mci_1.test'::regclass;
+
+SELECT * FROM mci_2.test ORDER BY test_id;
```
marcocitus: I changed this test because it started hitting a different error (table does not exists).

src/backend/distributed/metadata/metadata_cache.c
```diff
@@ -1164,6 +1164,25 @@ BuildCachedShardList(DistTableCacheEntry *cacheEntry)
...
+			/*
+			 * Normally, all shard cache entries for a given DistTableEntry are removed
+			 * before we get here. There is one exception: When a shard changes from
+			 * one relation ID to another. That typically happens during metadata
+			 * syncing when the distributed table is dropped and re-created without
+			 * changing the shard IDs. That means that old relation no longer exists
+			 * and we can safely wipe its entry, which will remove all corresponding
+			 * shard cache entries.
+			 */
+			ResetDistTableCacheEntry(shardEntry->tableEntry);
```
onderkalaci: Hmm, I couldn't find a way to trigger any bugs, but this seems a little bit dangerous to me.

We're in the middle of `BuildDistTableCacheEntry()`, and wouldn't calling `ResetDistTableCacheEntry ()` makes the `DistTableCacheEntry *` partially broken?

For example, `cacheEntry->partitionKeyString` is build before `BuildCachedShardList ()`, and now with this call we're invalidating `cacheEntry->partitionKeyString`.

Or, am I missing something?

marcocitus: The tableEntry in the shardEntry is the table entry for the old table, while we're rebuilding the table entry for the new table. I can add an extra check to confirm this is the case. That way, we'll fall back to the old error if there's a bug, plus it makes the intent of the code slightly more clear.

PR #3336, opened 2019-12-23
Commit normalized test output for better diffs

JelteF:
We have a `normalize.sed` script that before diffing test output normalizes the
expected file and the actual file. This makes sure that we don't have random
test failures and that we have to update our test output all the time. This PR
takes that one step further and actually commits the normalized files. That way
whenever we DO have to update our test output files only relevant changes will
be visible in the diff.

The other change that this PR does it that it strips trailing whitespace during 
normalization. This works well with our editorconfig settings.

As an added benefit of committing these files it's also much more visible what
new normalization rules will result in. The original changes that were proposed
here were a bit to wide and were changing output that was not intentended to
be changed: https://github.com/citusdata/citus/pull/3161#discussion_r360928922
Because these changes are now in the diff of the commit they are much easier to
spot.

Finally the Plan number normalization rules were also added to this PR, because
they are useful even without the CTE inlining PR.


onderkalaci:
@pykello feel free to have a look

src/test/regress/bin/diff
```diff
@@ -29,13 +30,14 @@ then
     DIFF=/usr/bin/diff
 fi
 
-if test -z "$VANILLATEST"
+if test -z "${VANILLATEST:-}"
 then
-	sed -Ef $BASEDIR/normalize.sed < $file1 > $file1.modified
-	sed -Ef $BASEDIR/normalize.sed < $file2 > $file2.modified
-	$DIFF -w $args $file1.modified $file2.modified
+	sed -Ef $BASEDIR/normalize.sed < $file1 > "$file1.modified"
```
serprex: `sed -i` can be used to modify files in-place, removing need for this `< > mv` dance

src/test/regress/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Isolation testing
+See `src/test/regress/spec/README.md`
+
+## Upgrade testing
+See `src/test/regress/upgrade/README.md`
+
+## Failure testing
+
+TODO: Add some text here
```
onderkalaci: maybe link to https://github.com/citusdata/citus/blob/master/src/test/regress/mitmscripts/README.md

which is very well commented

JelteF: Done

pykello:
Looks good to me.

PR #3337, opened 2019-12-23
Skip parse/deparse of local fast-path queries

onderkalaci:
If we're dealing with local fast-path queries, we can skip the expensive
operations of parse/deparse of the queries.

Given that the planner cannot know in advance whether the query is going
to be executed locally, it tries its best, skips expensive operations.

However, the executor could end-up with doing the expensive operations
if needed.

**This increase the performance around %13 on top of #3332  when the workers CPU bottlenecked**

The changes are inspired from #2830 and in-person discussions with @marcocitus

### DIFF
--- a/src/backend/distributed/commands/utility_hook.c
+++ b/src/backend/distributed/commands/utility_hook.c
@@ -48,6 +48,7 @@
 #include "distributed/metadata_cache.h"
 #include "distributed/metadata_sync.h"
 #include "distributed/multi_executor.h"
+#include "distributed/multi_explain.h"
 #include "distributed/resource_lock.h"
 #include "distributed/transmit.h"
 #include "distributed/version_compat.h"
@@ -164,6 +165,8 @@ multi_ProcessUtility(PlannedStmt *pstmt,
 	if (IsA(parsetree, ExplainStmt) &&
 		IsA(((ExplainStmt *) parsetree)->query, Query))
 	{
+		ExplainStatementRunning = true;
+
 		ExplainStmt *explainStmt = (ExplainStmt *) parsetree;
 
 		if (IsTransactionBlock())
@@ -693,6 +696,7 @@ multi_ProcessUtility(PlannedStmt *pstmt,
 			activeDropSchemaOrDBs++;
 		}
 
+
 		standard_ProcessUtility(pstmt, queryString, context,
 								params, queryEnv, dest, completionTag);
 
@@ -721,6 +725,13 @@ multi_ProcessUtility(PlannedStmt *pstmt,
 		{
 			activeDropSchemaOrDBs--;
 		}
+
+
+		if (IsA(parsetree, ExplainStmt) &&
+			IsA(((ExplainStmt *) parsetree)->query, Query))
+		{
+			ExplainStatementRunning = false;
+		}
 	}
 	PG_CATCH();
 	{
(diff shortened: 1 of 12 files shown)


src/backend/distributed/commands/utility_hook.c
```diff
@@ -721,6 +725,13 @@ multi_ProcessUtility(PlannedStmt *pstmt,
 		{
 			activeDropSchemaOrDBs--;
 		}
+
+
+		if (IsA(parsetree, ExplainStmt) &&
+			IsA(((ExplainStmt *) parsetree)->query, Query))
+		{
+			ExplainStatementRunning = false;
```
marcocitus: I suppose we could technically have an explain in explain if it's an EXPLAIN ANALYZE that calls a function that does an explain. Usually it's better to use levels.

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1925,9 +1940,21 @@ SingleShardModifyTaskList(Query *query, uint64 jobId, List *relationShardList,
...
+	 * fast path queries. With this, we can avoid deparsing during the execution
+	 * as well.
+	 */
+	if (localFastPathQuery)
+	{
+		queryString = NULL;
+	}
+	else
+	{
+		pg_get_query_def(query, queryString);
```
marcocitus: High level question: Is there any benefit we get from generating the query string here as opposed to in the executor?

src/backend/distributed/executor/local_executor.c
```diff
@@ -123,53 +133,124 @@ ExecuteLocalTaskList(CitusScanState *scanState, List *taskList)
...
+	 * references to the actual shard relations.
+	 *
+	 * Otherwise, like multi-shard queries going through the local execution,
+	 * we'd need to parse the queryString back to Query before passing it
+	 * to the planner,
+	 */
+	if (task->queryString == NULL)
 	{
-		Task *task = (Task *) lfirst(taskCell);
+		ReplaceShardReferencesWalker((Node *) shardQuery, task);
```
marcocitus: Is there a way to skip a few more steps here and replace the relation ID of the distributed table with that of the shard without doing the intermediate step of wrapping information in a function call like we do when joining reference tables with local tables? We're now in the unique situation that we can actually deparse the shard names without magic, since they are local tables.

(not sure, just asking)

src/backend/distributed/planner/multi_router_planner.c
JelteF: I added a commit that tries to do this by storing a copy of the query in the Task, instead of a string. Overal it seems to work pretty fine. The only thing that breaks badly seems are prepared statements without parameters. I expect it's related to the string memory being in the wrong memory context.

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -959,6 +964,21 @@ CreateDistributedExecution(RowModifyLevel modLevel, List *taskList, bool hasRetu
 		ExtractLocalAndRemoteTasks(readOnlyPlan, taskList, &execution->localTaskList,
 								   &execution->remoteTaskList);
 	}
+	else if (jobQuery && list_length(execution->tasksToExecute) == 1)
+	{
+		Task *task = (Task *) linitial(execution->tasksToExecute);
+
+		if (TaskAccessesLocalNode(task))
+		{
```
marcocitus: Do we need this condition given that GenerateShardQueryStringIfMissing has an inherent condition? Also, can't we have multiple local tasks?

JelteF:
Closing in favor of #3368

PR #3338, opened 2019-12-23
Skip when local

onderkalaci:
I'll clean this up a bit more.

**This increase the performance around ~%1.8 on top of #3337 when the workers CPU bottlenecked**

Given the trick I use in the executor part, I'll reconsider not doing this in the executor. The trick/hack is that we use `instrumentOptions` for something unrelated. We should double check not to break anything

### DIFF
--- a/src/backend/distributed/executor/local_executor.c
+++ b/src/backend/distributed/executor/local_executor.c
@@ -148,7 +148,7 @@ ExecuteLocalTaskList(CitusScanState *scanState, List *taskList)
 		LogLocalCommand(workerJob, task);
 
 		totalRowsProcessed +=
-			ExecuteLocalTaskPlan(scanState, localPlan, TaskQueryString(task));
+			ExecuteLocalTaskPlan(scanState, localPlan, task->queryString);
 	}
 
 	return totalRowsProcessed;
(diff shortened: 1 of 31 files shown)


src/backend/distributed/executor/local_executor.c
```diff
@@ -196,15 +196,7 @@ LocalTaskPlannedStmt(Query *workerJobQuery, Task *task, ParamListInfo boundParam
...
-	 * Altough the shardQuery is local to this node, we prefer planner()
-	 * over standard_planner(). The primary reason for that is Citus itself
-	 * is not very tolarent standard_planner() calls that doesn't go through
-	 * distributed_planner() because of the way that restriction hooks are
-	 * implemented. So, let planner to call distributed_planner() which
-	 * eventually calls standard_planner().
-	 */
-	int cursorOptions = 0;
+	int cursorOptions = CURSOR_OPT_FORCE_LOCAL;
 	PlannedStmt *localPlan = planner(shardQuery, cursorOptions, boundParams);
```
marcocitus: what would we need to do to safely call standard_planner here?

onderkalaci: I think that's a good idea. I see two options:

1) Relax the assertions in `multi_join_restriction_hook` and join_hook, so that we simply skip those when `plannerRestrictionContextList == NULL`

2) Call `CreateAndPushPlannerRestrictionContext()` and the pop functions around here.


I think (1) makes more sense

onderkalaci:
This doesn't have any prominent effect after #3369 and #3388, as we skip quite a lot with those two PRs

PR #3339, opened 2019-12-23
Always use NOTICE in log_remote_commands and avoid redaction when possible

marcocitus:
DESCRIPTION: Changes the citus.log_remote_commands level to NOTICE

It seems not very useful to issue these messages with LOG level since we almost always want to see these in the client session, plus they are currently unusable on Hyperscale.

### DIFF
--- a/src/include/distributed/log_utils.h
+++ b/src/include/distributed/log_utils.h
@@ -9,7 +9,22 @@
 #ifndef LOG_UTILS_H
 #define LOG_UTILS_H
 
+
+#include "utils/guc.h"
+
+
 extern bool IsLoggableLevel(int logLevel);
-extern char * ApplyLogRedaction(const char *text);
+extern char * HashLogMessage(const char *text);
+
+#define ApplyLogRedaction(text) \
+	(log_min_messages <= ereport_loglevel ? HashLogMessage(text) : text)
+
+#undef ereport
+#define ereport(elevel, rest) \
+	do { \
+		int ereport_loglevel = elevel; \
+		(void) (ereport_loglevel); \
+		ereport_domain(elevel, TEXTDOMAIN, rest); \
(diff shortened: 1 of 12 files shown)


serprex:
Why shouldn't ApplyLogRedaction take an elevel parameter so that it does the `log_min_messages` check?

marcocitus:
> Why shouldn't ApplyLogRedaction take an elevel parameter so that it does the log_min_messages check?

That seems a little clunky and prone to bugs. I'm wondering whether we should define a wrapper around ereport for these cases. We could change ApplyLogRedaction into a macro that only does stuff if the message is going to the log.

marcocitus:
>  I'm wondering whether we should define a wrapper around ereport for these cases. We could change ApplyLogRedaction into a macro that only does stuff if the message is going to the log.

Pushed a commit that does that by re-defining ereport.

src/include/distributed/log_utils.h
```diff
@@ -9,7 +9,22 @@
...
-extern char * ApplyLogRedaction(const char *text);
+extern char * HashLogMessage(const char *text);
+
+#define ApplyLogRedaction(text) \
+	(log_min_messages <= ereport_loglevel ? HashLogMessage(text) : text)
+
+#undef ereport
+#define ereport(elevel, rest) \
+	do { \
+		int ereport_loglevel = elevel; \
```
serprex: maybe `int ereport_loglevel = (elevel)` & then `ereport_domain(ereport_loglevel, TEXTDOMAIN, rest)` which also removes need for void cast to avoid usage warnings

marcocitus: We need elevel to be a compile time constant for the macro to inject pg_unreachable. Otherwise we get a bunch of compile errors.

PR #3342, opened 2019-12-24
Makefile fix DESTDIR together with cleanup

JelteF:
DESCRIPTION: Fix Makefile trying to cleanup PG directory during install

This should fix this build issue: https://redmine.postgresql.org/issues/5032


metdos:
@JelteF do we need to backport this to 9.1 release branch?

JelteF:
If we do another 9.1 release I think it makes sense to include this as well, but doing 9.1.2 only for this seems not very useful. The workaround is very simple.

metdos:
@JelteF I think you can cherry-pick this to 9.1 release branch. If we later decide to release another patch release it would be there.

JelteF:
Done

devrimgunduz:
Hi,

> If we do another 9.1 release I think it makes sense to include this as well, but doing 9.1.2 only for this seems not very useful. The workaround is very simple, i.e. run as `sudo` or remove citus files from your PG directory manually before.

Your use base prefer prebuilt packages. Building and patching the source file is not preferred. So, can I get a new point release, please, to build the RPMs?

Thx! Devrim

JelteF:
@devrimgunduz We tagged a new version: 9.1.2

devrimgunduz:
Hi @JelteF , already pushed them to the RPM repos. Thanks!

PR #3343, opened 2019-12-24
Remove exact count distinct

serprex:
DESCRIPTION: remove specialized `count(distinct)` support when approximation disabled

It's redundant alongside row-gather, which works for everything

It does potentially perform worse when there's lots of duplicates, as row-gather doesn't deduplicate on workers, whereas `count(distinct)` logic does. `count(distinct)` also doesn't disable partial aggregation for other aggregates

Not sure if tests should be updated to use row-gather for the cases they're now failing, or if we should remove tests. Probably a mix of both. Note that we're only failing tests where I'd explicitly disabled row-gather, as it's on by default


marcocitus:
There are a few test failures.

serprex:
Reviewed failing tests, this is my assessment:
```
set_operations: update expected
subquery_complex_target_list: same
subqueries_not_supported: same
multi_agg_approximate_distinct: same


multi_subquery:
	update test comments to not say it works. Include test repeated with row-gather

multi_agg_distinct:
	enable row-gather for failing tests

multi_single_relation_subquery:
	investigate why majority of nodes failed

multi_complex_count_distinct:
	remove failing tests (maybe remove this whole test, or run it with row-gather)
```

serprex:
The remaining test failure is a separate issue: #3371 

@pykello was considering that row-gather will have worse performance for cases when the existing pullUpDistinct logic would deduplicate on workers since it works by pushing down the arguments as `GROUP BY`. I feel like that strategy could be implemented as an `AGGREGATE_CUSTOM_DISTINCT`. Leaves two options: either I update this PR with a second commit implementing `AGGREGATE_CUSTOM_DISTINCT`, or we leave that as a future feature whose value can be determined during release planning

marcocitus:
Hmz, makes sense, the existing code makes a pretty useful optimisation. Let's maybe punt on it for now.

PR #3344, opened 2019-12-24
Fix tests when hll/topn installed

serprex:
Since `CREATE EXTENSION` is propagated now, we don't need to create extensions on workers in tests

Previously check-multi has been failing locally for me


marcocitus:
Had the same issue. I guess CircleCI doesn't have them installed?

PR #3347, opened 2019-12-30
Do not go through standard_planner() for INSERTs

onderkalaci:
Do not go through standard_planner() for INSERTs
    
That seems unnecessary. We already have the notion of FastPath queries,
simply add it there.

It is probably more logical to merge this PR before #3332, #3337 and #3338 so that
we make sure to build all the logic for INSERTs as well (@JelteF let me know what you think) 
    
Especially re-consider #3337 for INSERTs where there is  a parameter on the dist. key like
```SQL
PREPARE p1(int) AS INSERT INTO dist_table (dist_key) VALUES ($1)
```
the above can still skip parse/deparse for local queries.

### DIFF
--- a/src/backend/distributed/planner/fast_path_router_planner.c
+++ b/src/backend/distributed/planner/fast_path_router_planner.c
@@ -35,6 +35,7 @@
 #include "postgres.h"
 
 #include "distributed/distributed_planner.h"
+#include "distributed/insert_select_planner.h"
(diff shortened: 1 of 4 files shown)


onderkalaci:
I'll respond to your questions, this PR is not complete yet, `FastPathRouterQuery` should be adjusted a bit more because it checks for joinTree explicitly, where INSERTs don't have any.

PR #3348, opened 2019-12-31
Do not go through standard_planner() for INSERTs

onderkalaci:
Do not go through standard_planner() for INSERTs
    
That seems unnecessary. We already have the notion of FastPath queries,
simply add it there.

It is probably more logical to merge this PR before #3332, #3337 and #3338 so that
we make sure to build all the logic for INSERTs as well (@JelteF let me know what you think) 
    
Especially re-consider #3337 for INSERTs where there is  a parameter on the dist. key like
```SQL
PREPARE p1(int) AS INSERT INTO dist_table (dist_key) VALUES ($1)
```
the above can still skip parse/deparse for local queries. I added regression tests for that, so that those PRs cannot break it.

**I'll run some benchmarks before marking the PR as non-draft PR, if the results are promising, we can consider merging this.**

### DIFF
--- a/src/test/regress/sql/local_shard_execution.sql
+++ b/src/test/regress/sql/local_shard_execution.sql
@@ -485,6 +485,65 @@ BEGIN;
 	EXECUTE remote_prepare_param(1);
 COMMIT;
 
+PREPARE local_insert_prepare_no_param AS INSERT INTO distributed_table VALUES (1+0*random(), '11',21::int) ON CONFLICT(key) DO UPDATE SET value = '29' || '28' RETURNING *, key + 1, value || '30', age * 15;
+PREPARE local_insert_prepare_param (int) AS INSERT INTO distributed_table VALUES ($1+0*random(), '11',21::int) ON CONFLICT(key) DO UPDATE SET value = '29' || '28' RETURNING *, key + 1, value || '30', age * 15;
+BEGIN;
+	-- 6 local execution without params
+	EXECUTE local_insert_prepare_no_param;
+	EXECUTE local_insert_prepare_no_param;
+	EXECUTE local_insert_prepare_no_param;
+	EXECUTE local_insert_prepare_no_param;
+	EXECUTE local_insert_prepare_no_param;
+	EXECUTE local_insert_prepare_no_param;
+
+	-- 6 local executions with params
+	EXECUTE local_insert_prepare_param(1);
+	EXECUTE local_insert_prepare_param(5);
+	EXECUTE local_insert_prepare_param(6);
+	EXECUTE local_insert_prepare_param(1);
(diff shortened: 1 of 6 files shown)


onderkalaci:
Moving the discussion from the old [PR](https://github.com/citusdata/citus/pull/3347#pullrequestreview-337097153) -- @JelteF fyi

> Very nice that it's such a simple PR. Could you explain a bit more why no inserts have to go through standard_planner? (except INSERT ... SELECT)

For SELECT/UPDATE/DELETE queries, we heaviliy relay on standard_planner(), because we need the restriction information via `multi_join_restriction_hook` and and `multi_relation_restriction_hook`. There are also some other things that Citus relies on, like `eval_const_expressions()` on the certain parts of the query or certain jointree rewrite. 

However, for INSERTs, we really don't need any of these things. Just a note, we don't support INSERT with `query->hasSublinks` at the momemt. We might consider running standard_planner in that case when we support.



> I'm interested mostly in why multirow inserts work out of the box. Do we split those up somewhere else already?

Yes, that's true. `RouterInsertJob()` is the function that handles both types of `INSERT` commands. 

In the end, this PR only skips `standard_planner()` for the purposes of distributed planning. We're still going through router planning, and the branching between INSERT vs SELECT/UPDATE/DELETE already happens at the start of router planning.

The only thing that I'm still thinking is that, with this PR, fast-path router queries might include multi-task distributed queries with multi-row INSERTs. I don't think that's a problem, but it changes the scope/definition of fast-path planning. In that sense, we could consider not going through fast-path for multi-row INSERTs, in order to make the overall design simple to reason about.

onderkalaci:
> I'll run some benchmarks before marking the PR as non-draft PR, if the results are promising, we can consider merging this.

This helps to improve the benchmark results ~%4.8, so I'm in favor of merging this first.

```SQL
15322 NOPM -> #3338 
16070 NOPM -> this commit cherry-picked on top of #3338
```

src/test/regress/sql/local_shard_execution.sql
```diff
@@ -482,6 +482,29 @@ BEGIN;
 	EXECUTE remote_prepare_param(1);
 COMMIT;
 
+PREPARE local_insert_prepare_no_param AS INSERT INTO distributed_table VALUES (1+0*random(), '11',21::int) ON CONFLICT(key) DO UPDATE SET value = '29' || '28' RETURNING *, key + 1, value || '30', age * 15;
```
onderkalaci: Add more multi-row INSERTs

onderkalaci:
added few more tests with local multi-row INSERTs, either a single task or multiple tasks but still both local tasks for completeness as @JelteF requested, merging now

PR #3349, opened 2019-12-31
End regression tests with ensure_no_intermediate_data_leak

### LINKED ISSUE #2429: Citus may leak intermediate results
With #2424, we've fixed a critical intermediate result leak bug. However, it turned out to be few more cases which are not too common.

1. Failures on task-tracker execution: See the discussion [here](https://github.com/citusdata/citus/pull/2424#issue-219729944) 
2. Citus on Windows, see [here](https://github.com/citusdata/citus/pull/2424#issuecomment-426563454)

Given that we don't have many users relaying on the above, we prefer to fix those later.

Once we fix both, we should be able to move `ensure_no_intermediate_data_leak .sql` file to the end of the all different regression test schedules like `check-multi` and `check-failure`.

serprex:
All of the listed issues in #2429 are resolved. However, `ensure_no_intermediate_data_leak` is not passing on all regression tests

Current leaks:
- [x] `multi_multiuser`
- [x] `worker_range_partition_complex`
- [x] `task_tracker_assign_task`

Fixes #2429

### DIFF
--- a/src/test/regress/expected/ensure_no_intermediate_data_leak.out
+++ b/src/test/regress/expected/ensure_no_intermediate_data_leak.out
@@ -1,20 +1,17 @@
 ------
--- THIS TEST SHOULD IDEALLY BE EXECUTED AT THE END OF 
--- THE REGRESSION TEST SUITE  TO MAKE SURE THAT WE 
--- CLEAR ALL INTERMEDIATE RESULTS ON BOTH THE COORDINATOR 
+-- THIS TEST SHOULD IDEALLY BE EXECUTED AT THE END OF
+-- THE REGRESSION TEST SUITE TO MAKE SURE THAT WE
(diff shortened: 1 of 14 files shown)


src/test/regress/expected/ensure_no_intermediate_data_leak_0.out
```diff
@@ -0,0 +1,19 @@
...
+------
+SELECT pg_ls_dir('base/pgsql_job_cache') WHERE citus_version() NOT ILIKE '%windows%';
+ pg_ls_dir 
+-----------
+(0 rows)
+
+SELECT run_command_on_workers($$SELECT pg_ls_dir('base/pgsql_job_cache') WHERE citus_version() NOT ILIKE '%windows%'$$);
+ run_command_on_workers 
+------------------------
+ (localhost,9060,t,"")
```
serprex: We could update this test to only select cases when there are directories, so that we only have 1 expected result, as opposed to 3

PR #3350, opened 2020-01-02
Lazy query deparsing executable queries

JelteF:
Deparsing and parsing a query can be heavy on CPU. When locally executing 
the query we don't need to do this in theory most of the time.

This PR is the first step in allowing to skip deparsing and parsing
the query in these cases, by lazily creating the query string and
storing the query in the task. Future commits will make use of this and
not deparse and parse the query anymore, but use the one from the task
directly.


src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1842,26 +1843,17 @@ SingleShardSelectTaskList(Query *query, uint64 jobId, List *relationShardList,
...
-	{
-		queryString = NULL;
-	}
-	else
-	{
-		pg_get_query_def(query, queryString);
-	}
-
-	task->queryString = queryString ? queryString->data : NULL;
+	task->query = copyObject(query);
```
marcocitus: Copying the query tree seems like an expensive operation. Can it be avoided? The Job also has a query tree.

JelteF: We can at least use the job query tree in some cases. And possibly we don't need to copy the query if we already copy it later on. Since we don't actually modify the query in the task. I'm mainly copying it currently to rule it out as a source for the errors.

JelteF:
SELF NOTE: The following query string might break on local execution, because the actualy query stored in the task got corrupted there.

```
PREPARE prepared_test_2 AS
WITH users_events AS(
  SELECT
      *
  FROM
      events_table
)

SELECT
    users_events.event_type
FROM
    users_events
    join
    (SELECT 1 event_type, 1 count) event_attendee_count
    on (users_events.event_type = event_attendee_count.event_type)
;
```

src/backend/distributed/executor/local_executor.c
```diff
@@ -124,53 +136,158 @@ ExecuteLocalTaskList(CitusScanState *scanState, List *taskList)
...
 	EState *executorState = ScanStateGetExecutorState(scanState);
 	ParamListInfo paramListInfo = copyParamList(executorState->es_param_list_info);
-	int numParams = 0;
-	Oid *parameterTypes = NULL;
 	ListCell *taskCell = NULL;
 	uint64 totalRowsProcessed = 0;
+	Job *workerJob = scanState->distributedPlan->workerJob;
 
-	if (paramListInfo != NULL)
+	foreach(taskCell, taskList)
```
onderkalaci: I thought we'd be  excluding the  changes related  to local execution from this PR. Can't we simply call `TaskQueryString`  followed by `ParseQueryString ` in this PR, and implement local-fast-path specific  ones in #3337?

The idea that I thought we'd be to generalize lazy queryString generation in this  PR,  and skip generating the `queryString` in #3337? Or, am I missing something?

onderkalaci: We've talked with @JelteF in person, decided to go with the above approach to simplify the code reviews. I'll push a commit that does it soon

src/backend/distributed/planner/multi_router_planner.c
JelteF: I've stopped copying the querytree and all the tests keep passing, so I think it's fine to not copy them. I'll investigate a bit more though to be more sure.

src/backend/distributed/executor/local_executor.c
JelteF: Done

src/backend/distributed/executor/local_executor.c
```diff
@@ -124,53 +136,158 @@ ExecuteLocalTaskList(CitusScanState *scanState, List *taskList)
...
-		 * eventually calls standard_planner().
-		 */
-		PlannedStmt *localPlan = planner(shardQuery, cursorOptions, paramListInfo);
+		if (boundParams != NULL)
+		{
+			const char **parameterValues = NULL; /* not used anywhere, so decleare here */
 
-		LogLocalCommand(shardQueryString);
+			ExtractParametersForLocalExecution(boundParams, &parameterTypes,
+											   &parameterValues);
```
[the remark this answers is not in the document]
JelteF: Not relevant to this PR anymore.

src/backend/distributed/executor/local_executor.c
```diff
@@ -506,3 +507,31 @@ DisableLocalExecution(void)
 					  (superuser() ? PGC_SUSET : PGC_USERSET), PGC_S_SESSION,
 					  GUC_ACTION_LOCAL, true, 0, false);
 }
+
+
+/*
+ * GenerateShardQueryStringIfMissing gets the workerJob's query and the task.
```
onderkalaci: Minor: outdated comment

src/backend/distributed/executor/local_executor.c
```diff
@@ -506,3 +507,31 @@ DisableLocalExecution(void)
...
+ * re-generate the queryString.
+ */
+char *
+TaskQueryString(Task *task)
+{
+	if (task->queryStringLazy != NULL)
+	{
+		return task->queryStringLazy;
+	}
+	Assert(task->query != NULL);
```
onderkalaci: Also, might be useful to have a comment explaining why we're doing the context switch as it might not be obvious for the readers

src/backend/distributed/executor/adaptive_executor.c
```diff
@@ -132,8 +132,11 @@
 #include "catalog/pg_type.h"
 #include "commands/dbcommands.h"
 #include "distributed/citus_custom_scan.h"
+#include "distributed/citus_ruleutils.h"
```
onderkalaci: Minor: do we really need all these headers in this PR?

src/backend/distributed/executor/local_executor.c
```diff
@@ -506,3 +507,35 @@ DisableLocalExecution(void)
 					  (superuser() ? PGC_SUSET : PGC_USERSET), PGC_S_SESSION,
 					  GUC_ACTION_LOCAL, true, 0, false);
 }
+
+
+/*
+ * TaskQueryString generates task->queryStringLazy if missing.
+ *
+ * For performance reasons, the queryString is generated lazily. For example
+ * for local queries it is always needed to generate it.
```
onderkalaci: Minor: `For example
 * for local queries it is always needed to generate it.` this part of the comment sounds a little confusing to me. For some  of the local queries, we skip generating the query string?

src/backend/distributed/planner/multi_router_planner.c
```diff
@@ -1837,17 +1838,20 @@ RemoveCoordinatorPlacement(List *placementList)
...
 	RowLocksOnRelations((Node *) query, &relationRowLockList);
-	pg_get_query_def(query, queryString);
 
-	task->queryString = queryString->data;
+	/*
+	 * For performance reasons, we skip generating the queryString. For local
+	 * execution this is not needed, so we wait until the executor determines
+	 * that the query cannot be executed locally.
+	 */
+	task->query = query;
```
onderkalaci: I think the  down side of doing this is that now the `task` is heavier, in the sense that when we need to copy the task, we'd be copying  the queryTree, instead of a string. (Similar for `SingleShardModifyTaskList `). So, is it worth *always* doing it for single shard queries?

I don't want to over complicate things, just sharing some potential ways of doing it less commonly.

So, is it possible that we only do this when there is a possibility of local execution? Something like:


```C

/* best effort to defer query string generation */
if (TaskAccessesLocalNode())
{
task->query = query;
task->queryStringLazy = NULL;
}
else
{
pg_get_query_def(query, queryString);
task->queryString = queryString->data;
}
```

We're not doing deferring query generation for multi-shard queries anyway. So, I wouldn't mind doing it only for `tasks that could potentially use local execution`

What is your take on this?


(There is already a function named `TaskAccessesLocalNode` for checking this, we could change the api to get a placement list instead of task)

JelteF: Last commit implements this in a helper function.

src/backend/distributed/planner/deparse_shard_query.c
```diff
@@ -105,13 +106,13 @@ RebuildQueryStrings(Query *originalQuery, List *taskList)
 		}
 
 		ereport(DEBUG4, (errmsg("query before rebuilding: %s",
-								task->queryString == NULL ? "(null)" :
-								ApplyLogRedaction(task->queryString))));
+								TaskQueryString(task) == NULL ? "(null)" :
+								ApplyLogRedaction(TaskQueryString(task)))));
 
 		UpdateTaskQueryString(query, relationId, valuesRTE, task);
```
onderkalaci: Is it possible to defer query string generation here as well? This is the code-path called for all INSERTs, so we're not lazily deferring for INSERTs?

JelteF: Should be resolved in the latest commit

src/include/distributed/multi_physical_planner.h
```diff
@@ -182,15 +182,23 @@ typedef struct Task
...
+	 * assign a different query for each placement. We need this flexibility
+	 * when a query should return node specific values. For example, on which
+	 * node did we succeed storing some result files?
+	 *
+	 * In almost all cases queryStringLazy should be read by using
+	 * TaskQueryString(). This will populate the field if only the query field
+	 * is not NULL.
 	 */
-	char *queryString;
+	Query *query;
```
marcocitus: Why do we need this in addition to jobQuery + relationShardList?

JelteF: The main reason is that we don't have access to the job in all places where we currently use `task->query`.

src/backend/distributed/executor/local_executor.c
```diff
@@ -506,3 +505,72 @@ DisableLocalExecution(void)
 					  (superuser() ? PGC_SUSET : PGC_USERSET), PGC_S_SESSION,
 					  GUC_ACTION_LOCAL, true, 0, false);
 }
+
+
+void
+SetTaskQueryAndPlacementList(Task *task, Query *query, List *placementList)
```
marcocitus: missing function comment

src/backend/distributed/planner/deparse_shard_query.c
```diff
@@ -139,30 +141,65 @@ UpdateTaskQueryString(Query *query, Oid distributedTableId, RangeTblEntry *value
...
+		 */
 		List *relationShardList = task->relationShardList;
 		UpdateRelationToShardNames((Node *) query, relationShardList);
-
-		pg_get_query_def(query, queryString);
+		if (!localExecution)
+		{
+			pg_get_query_def(query, queryString);
+		}
 	}
```
marcocitus: This function is a little dizzying. Would it be worth writing a specialized one at least for inserts?

src/include/distributed/multi_physical_planner.h
marcocitus: Can we change that? This blows up the data structure quite a bit.

src/backend/distributed/executor/local_executor.c
marcocitus: The AndPlacementList part does not seem to contribute a whole lot since it's just an assignment and distracts from the main purpose of this function.

src/backend/distributed/executor/local_executor.c
```diff
@@ -506,3 +505,72 @@ DisableLocalExecution(void)
...
+
+/*
+ * TaskQueryString generates task->queryStringLazy if missing.
+ *
+ * For performance reasons, the queryString is generated lazily. For example
+ * for local queries it is usually not needed to generate it, so this way we
+ * can skip the expensive deparsing+parsing.
+ */
+char *
+TaskQueryString(Task *task)
```
marcocitus: It seems more appropriate to have this function (and maybe SetTaskQueryAndPlacementList) in deparse_shard_query.c, since it's more of general deparsing function.

src/include/distributed/multi_physical_planner.h
JelteF: as discussed in person this is not possible or at least not easy in some cases. So this will be kept like it is

src/include/distributed/multi_physical_planner.h
```diff
@@ -182,15 +182,23 @@ typedef struct Task
...
+	char *queryStringLazy;
 	List *perPlacementQueryStrings;
 
+	Oid distributedTableId;     /* only applies to insert tasks that are locally executed */
```
onderkalaci: I can see that this'd be useful and many devs will use this. How hard is it to make this generic for all tasks?

src/backend/distributed/utils/citus_readfuncs.c
```diff
@@ -377,7 +377,8 @@ ReadTask(READFUNC_ARGS)
 	READ_ENUM_FIELD(taskType, TaskType);
 	READ_UINT64_FIELD(jobId);
 	READ_UINT_FIELD(taskId);
-	READ_STRING_FIELD(queryString);
+	READ_STRING_FIELD(queryStringLazy);
```
onderkalaci: Is it intentional to not have `perPlacementQueryStrings ` here?

src/backend/distributed/planner/deparse_shard_query.c
```diff
@@ -139,30 +141,65 @@ UpdateTaskQueryString(Query *query, Oid distributedTableId, RangeTblEntry *value
...
+	bool localExecution = TaskAccessesLocalNode(task);
+
 	if (query->commandType == CMD_INSERT)
 	{
-		deparse_shard_query(query, distributedTableId, task->anchorShardId, queryString);
+		/*
+		 * For INSERT queries, we only have one relation to update, so we can
+		 * use deparse_shard_query().
+		 */
+		if (localExecution)
```
onderkalaci: Even if taskAccessesLocalNode, users can disable local execution by `citus.enable_local_execution` (or Citus sets this within a TX block when necessary).

Do we produce accurate results when the GUC is off?

src/backend/distributed/planner/deparse_shard_query.c
JelteF: Simplified it now. Is this better?

src/include/distributed/multi_physical_planner.h
JelteF: removed it and replaced it with relationShardList, like discussed in person

src/backend/distributed/planner/deparse_shard_query.c
JelteF: Since you have made tests for this on your branch I will resolve this.

src/backend/distributed/executor/local_executor.c
JelteF: Makes sense, moved it there

src/backend/distributed/planner/deparse_shard_query.c
JelteF: I have minimised my changes to this function and I think it's as easy to understand as it was before now.

src/include/distributed/multi_physical_planner.h
marcocitus: I'd still like a better name and comment.

JelteF: Done, now called `queryForLocalExecution`

src/include/distributed/multi_physical_planner.h
JelteF: Turns out my solution with relationShardList was not working for `INSERT ... SELECT` queries. I'v changed it back to distributedTableId now, but renamed `distributedTableId` to `anchorDistributedTableId`. I think this is the easiest solution.

src/backend/distributed/utils/citus_readfuncs.c
pykello: Yes, that was unintentional. I will create a PR for this.

PR #3352, opened 2020-01-02
Allow RETURNING in fast-path queries

onderkalaci:
Because there is no specific reason for that.

There are already bunch of regression test with UPDATE/DELETE/INSERT

### DIFF
--- a/src/test/regress/sql/fast_path_router_modify.sql
+++ b/src/test/regress/sql/fast_path_router_modify.sql
@@ -53,8 +53,11 @@ UPDATE modify_fast_path SET key = 1::float WHERE key = 1;
 UPDATE modify_fast_path SET key = 2 WHERE key = 1;
 UPDATE modify_fast_path SET key = 2::numeric WHERE key = 1;
 
--- returning is not supported via fast-path
+-- returning is supported via fast-path
+INSERT INTO modify_fast_path (key, value_1) VALUES (1,1);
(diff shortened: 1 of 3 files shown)


src/test/regress/sql/fast_path_router_modify.sql
```diff
@@ -53,8 +53,11 @@ UPDATE modify_fast_path SET key = 1::float WHERE key = 1;
 UPDATE modify_fast_path SET key = 2 WHERE key = 1;
 UPDATE modify_fast_path SET key = 2::numeric WHERE key = 1;
 
--- returning is not supported via fast-path
+-- returning is supported via fast-path
+INSERT INTO modify_fast_path (key, value_1) VALUES (1,1);
 DELETE FROM modify_fast_path WHERE key = 1 RETURNING *;
+INSERT INTO modify_fast_path (key, value_1) VALUES (2,1);
```
marcocitus: maybe add returning to the inserts as well

onderkalaci: added the test

onderkalaci:
> This line should be removed: https://github.com/citusdata/citus/pull/3352/files#diff-021095bb9d8ba99274a746314d429aacR165

> Tested that we still error when RETURNING random() in fast path. Including that in a test might be good

both done

onderkalaci:
@serprex given that your requests were minor, and this PR is approved, I'm merging it. Feel free to comment on again if you think anything should be improved.

PR #3353, opened 2020-01-02
Partitioned task list results.

pykello:
Implements `PartitionTasklistResults()`, which partitions results of given SELECT tasks based on shard ranges of a given relation. Comments from that function:

```c
/*
 * PartitionTasklistResults executes the given task list, and partitions results
 * of each task based on targetRelation's distribution method and intervals.
 * Each of the result partitions are stored in the node where task was executed,
 * and are named as $resultIdPrefix_from_$sourceShardId_to_$targetShardIndex.
 *
 * Result is list of DistributedResultFragment, each of which represents a
 * partition of results. Empty results are omitted. Therefore, if we have N tasks
 * and target relation has M shards, we will have NxM-(number of empty results)
 * fragments.
 */
List *
PartitionTasklistResults(char *resultIdPrefix, List *selectTaskList,
						 DistTableCacheEntry *targetRelation,
						 bool binaryFormat)
```

This will be used by a higher level function `CreateDistributedResult()`, which moves the fragments created by `PartitionTasklistResults()` between nodes so they are colocated with the shards of `targetRelation`. Then `CreateDistributedResult()` will be used to implement INSERT/SELECT with re-partitioning.

### DIFF
--- a/src/backend/distributed/executor/distributed_intermediate_results.c
+++ b/src/backend/distributed/executor/distributed_intermediate_results.c
@@ -1,0 +1,320 @@
+/*-------------------------------------------------------------------------
+ *
+ * distributed_intermediate_results.c
+ *   Functions for reading and writing distributed intermediate results.
+ *
+ * Copyright (c), Citus Data, Inc.
+ *
+ *-------------------------------------------------------------------------
+ */
+#include <sys/stat.h>
+#include <unistd.h>
+
+#include "postgres.h"
+#include "funcapi.h"
+#include "miscadmin.h"
+#include "port.h"
+
+#include "access/tupdesc.h"
+#include "catalog/pg_type.h"
+#include "distributed/intermediate_results.h"
+#include "distributed/metadata_cache.h"
+#include "distributed/multi_executor.h"
+#include "distributed/transaction_management.h"
+#include "distributed/tuplestore.h"
+#include "distributed/worker_protocol.h"
+#include "tcop/pquery.h"
+#include "tcop/tcopprot.h"
+#include "utils/builtins.h"
+#include "utils/lsyscache.h"
+
+
+/* forward declarations of local functions */
+static void WrapTasksForPartitioning(char *resultIdPrefix, List *selectTaskList,
+									 DistTableCacheEntry *targetRelation,
+									 bool binaryFormat);
+static List * ExecutePartitionTaskList(List *partitionTaskList,
+									   DistTableCacheEntry *targetRelation);
+static ArrayType * CreateArrayFromDatums(Datum *datumArray, bool *nullsArray, int
+										 datumCount, Oid typeId);
+static void ShardMinMaxValueArrays(ShardInterval **shardIntervalArray, int shardCount,
+								   Oid intervalTypeId, ArrayType **minValueArray,
+								   ArrayType **maxValueArray);
+static char * SourceShardPrefix(char *resultPrefix, uint64 shardId);
+static DistributedResultFragment * TupleToDistributedResultFragment(
+	TupleTableSlot *tupleSlot, DistTableCacheEntry *targetRelation);
+static Tuplestorestate * ExecuteSelectTasksIntoTupleStore(List *taskList, TupleDesc
+														  resultDescriptor);
+
+
+/*
+ * PartitionTasklistResults executes the given task list, and partitions results
+ * of each task based on targetRelation's distribution method and intervals.
+ * Each of the result partitions are stored in the node where task was executed,
+ * and are named as $resultIdPrefix_from_$sourceShardId_to_$targetShardIndex.
+ *
+ * Result is list of DistributedResultFragment, each of which represents a
+ * partition of results. Empty results are omitted. Therefore, if we have N tasks
+ * and target relation has M shards, we will have NxM-(number of empty results)
+ * fragments.
+ */
+List *
+PartitionTasklistResults(char *resultIdPrefix, List *selectTaskList,
+						 DistTableCacheEntry *targetRelation,
+						 bool binaryFormat)
+{
+	/*
+	 * Make sure that this transaction has a distributed transaction ID.
+	 *
+	 * Intermediate results will be stored in a directory that is derived
+	 * from the distributed transaction ID.
+	 */
+	UseCoordinatedTransaction();
+
+	WrapTasksForPartitioning(resultIdPrefix, selectTaskList, targetRelation,
+							 binaryFormat);
+	return ExecutePartitionTaskList(selectTaskList, targetRelation);
+}
+
+
+/*
+ * WrapTasksForPartitioning wraps the query for each of the tasks by a call
+ * to worker_partition_query_result(). Target list of the wrapped query should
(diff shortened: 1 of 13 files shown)


src/backend/distributed/test/distributed_intermediate_results.c
```diff
@@ -0,0 +1,91 @@
...
+
+#include "distributed/commands/multi_copy.h"
+#include "distributed/connection_management.h"
+#include "distributed/intermediate_results.h"
+#include "distributed/multi_executor.h"
+#include "distributed/remote_commands.h"
+#include "distributed/tuplestore.h"
+#include "tcop/tcopprot.h"
+
+PG_FUNCTION_INFO_V1(parition_task_list_results);
```
serprex: `partition` is being consistently spelt as `parition`?

src/backend/distributed/executor/distributed_intermediate_results.c
```diff
@@ -0,0 +1,322 @@
...
+		if (!maxValueNulls[shardIndex])
+		{
+			Datum maxValue = shardIntervalArray[shardIndex]->maxValue;
+			char *maxValueStr = DatumGetCString(OidFunctionCall1(intervalTypeOutFunc,
+																 maxValue));
+			maxValues[shardIndex] = CStringGetTextDatum(maxValueStr);
+		}
+	}
+
+	(*minValueArray) = CreateArrayFromDatums(minValues, minValueNulls, shardCount,
```
serprex: Not a fan of excessive parentheses

src/backend/distributed/executor/intermediate_results.c
```diff
@@ -848,9 +847,6 @@ fetch_intermediate_results(PG_FUNCTION_ARGS)
 		totalBytesWritten += FetchRemoteIntermediateResult(connection, resultId);
 	}
 
-	RemoteTransactionCommit(connection);
```
serprex: What prompted this change?

pykello: In `fetch_intermediate_results()` we were not forcing new connections, so we could be using remote connections which had already executed some commands. Committing and closing connections would be wrong in that case. For example,

```sql
BEGIN;
UPDATE dist_table a = a + 1;
SELECT fetch_intermediate_results(...);
DELETE FROM dist_table;
END;
```

wouldn't work, because `fetch_intermediate_results()` would have committed and closed the connections and connections weren't open to send `DELETE`s.

Also commit and close would delete remote intermediate files, since we delete intermediate files at commit.

src/test/regress/sql/distributed_intermediate_results.sql
```diff
@@ -0,0 +1,103 @@
...
+  FROM distributed_result_info GROUP BY nodeport ORDER BY nodeport;
+-- Read all fetched result files. Sum(x) should be 4550, verified by
+-- racket -e '(for/sum ([i (range 1 101)]) (modulo (* 3 i i) 100))'
+SELECT count(*), sum(x) FROM
+  read_intermediate_results((SELECT array_agg(resultId) FROM distributed_result_info),
+                            'text') AS res (x int);
+END;
+
+DROP TABLE source_table, target_table, distributed_result_info;
+
```
serprex: Are we able to test that intermediate result files are deleted in the end?

src/backend/distributed/executor/distributed_intermediate_results.c
```diff
@@ -0,0 +1,320 @@
...
+	StringInfo maxValuesString = ArrayObjectToString(maxValueArray, TEXTOID,
+													 intervalTypeMod);
+
+	foreach(taskCell, selectTaskList)
+	{
+		Task *selectTask = (Task *) lfirst(taskCell);
+		StringInfo wrappedQuery = makeStringInfo();
+		List *shardPlacementList = selectTask->taskPlacementList;
+
+		ShardPlacement *shardPlacement = linitial(shardPlacementList);
```
serprex: `Assert(list_length(shardPlacementList) == 1)`?

src/test/regress/sql/distributed_intermediate_results.sql
```diff
@@ -0,0 +1,103 @@
...
+--
+
+-- partition_task_list_results tests the internal PartitionTasklistResults function
+CREATE OR REPLACE FUNCTION pg_catalog.partition_task_list_results(resultIdPrefix text,
+                                                                  query text,
+                                                                  target_table regclass,
+                                                                  binaryFormat bool DEFAULT true)
+    RETURNS TABLE(resultId text,
+                  nodeId int,
+                  rowCount int,
```
marcocitus: should rowCount be bigint?

marcocitus: minor: This is bigint in other places that use row count.

src/backend/distributed/executor/distributed_intermediate_results.c
marcocitus: Would we need to support replication here? We should at least be able to generate a task per placement and have it execute on whichever placement is available. Not sure about the next step.

marcocitus: We may have to return groupId instead of nodeId to be able to use the same query.

src/test/regress/sql/distributed_intermediate_results.sql
marcocitus: Note that these intermediate results are cleaned up when the transaction that created them commits/aborts, so it's quite unlikely that they will leak. 

Task-tracker queries generate files that are not automatically cleaned up.

src/backend/distributed/executor/distributed_intermediate_results.c
```diff
@@ -0,0 +1,320 @@
...
+									  "hash" : "range";
+		const char *binaryFormatString = binaryFormat ? "true" : "false";
+
+		appendStringInfo(wrappedQuery,
+						 "SELECT %d, partition_index"
+						 ", %s || '_' || partition_index::text "
+						 ", rows_written "
+						 "FROM worker_partition_query_result"
+						 "(%s,%s,%d,%s,%s,%s,%s) WHERE rows_written > 0",
+						 shardPlacement->nodeId,
```
marcocitus: I wonder whether there's another way in which can fill in the node ID, such that we can fail over to another replica if needed and return a different node ID in the fragment list.

One way could be to look at the task execution state in the DistributedExecution. We'd probably have to include the anchor shard ID in the result and map that back to a ShardCommandExecution to figure out which node the task was executed on, which probably requires a bunch of refactoring since the ShardCommandExecutions are not easily accessible from the DistributedExecution data structure. 

Anyway, we should probably error out here when there are multiple placements if we don't properly support it.

pykello: What is your take on something like https://github.com/citusdata/citus/pull/3363? I haven't tested it much yet, just opened it to get early feedback.

src/backend/distributed/executor/distributed_intermediate_results.c
pykello: Added support for replication here in https://github.com/citusdata/citus/pull/3363

PR #3355, opened 2020-01-04
Redistribute task list results to correspond to a target relation's distribution

pykello:
In https://github.com/citusdata/citus/pull/3353 we implement `PartitionTasklistResults()` which partitions results of a query task list using distribution ranges of a target relation. To be able to insert those partitioned results into the target relation, we need to co-locate result file fragments with the shard placements of the target relation. This PR implements `RedistributeTaskListResult()` which does that.

`RedistributeTaskListResult()`'s return value contains a list of result ids to be inserted for each shard of target relation, which INSERT/SELECT will use to construct `read_intermediate_results()` calls which replace the original SELECT subqueries.

### DIFF
--- a/src/backend/distributed/executor/distributed_intermediate_results.c
+++ b/src/backend/distributed/executor/distributed_intermediate_results.c
@@ -18,6 +18,7 @@
 #include "access/tupdesc.h"
 #include "catalog/pg_type.h"
 #include "distributed/intermediate_results.h"
+#include "distributed/master_metadata_utility.h"
 #include "distributed/metadata_cache.h"
 #include "distributed/multi_executor.h"
 #include "distributed/transaction_management.h"
@@ -27,6 +28,28 @@
 #include "tcop/tcopprot.h"
 #include "utils/builtins.h"
 #include "utils/lsyscache.h"
+
+
+/*
+ * NodePair contains the source and destination node in a NodeToNodeFragmentsTransfer.
+ * It is a separate struct to use it as a key in a hash table.
+ */
+typedef struct NodePair
+{
+	int sourceNodeId;
+	int targetNodeId;
+} NodePair;
+
+
+/*
+ * NodeToNodeFragmentsTransfer contains all fragments that need to be fetched from
+ * the source node to the destination node in the NodePair.
+ */
+typedef struct NodeToNodeFragmentsTransfer
+{
+	NodePair nodes;
+	List *fragmentList;
+} NodeToNodeFragmentsTransfer;
 
 
 /* forward declarations of local functions */
@@ -46,6 +69,44 @@ static DistributedResultFragment * TupleToDistributedResultFragment(
 static Tuplestorestate * ExecuteSelectTasksIntoTupleStore(List *taskList,
 														  TupleDesc resultDescriptor,
 														  bool errorOnAnyFailure);
+static List ** ColocateFragmentsWithRelation(List *fragmentList,
+											 DistTableCacheEntry *targetRelation);
+static List * ColocationTransfers(List *fragmentList,
+								  DistTableCacheEntry *targetRelation);
+static List * FragmentTransferTaskList(List *fragmentListTransfers);
+static char * QueryStringForFragmentsTransfer(
+	NodeToNodeFragmentsTransfer *fragmentsTransfer);
+static void ExecuteFetchTaskList(List *fetchTaskList);
+
+
+/*
+ * RedistributeTaskListResults partitions the results of given task list using
+ * shard ranges and partition method of given targetRelation, and then colocates
+ * the result files with shards.
+ *
+ * If a shard has a replication factor > 1, corresponding result files are copied
+ * to all nodes containing that shard.
+ *
+ * returnValue[shardIndex] is list of cstrings each of which is a resultId which
+ * correspond to targetRelation->sortedShardIntervalArray[shardIndex].
+ */
+List **
+RedistributeTaskListResults(char *resultIdPrefix, List *selectTaskList,
+							DistTableCacheEntry *targetRelation,
+							bool binaryFormat)
+{
+	/*
+	 * Make sure that this transaction has a distributed transaction ID.
+	 *
+	 * Intermediate results will be stored in a directory that is derived
+	 * from the distributed transaction ID.
+	 */
+	UseCoordinatedTransaction();
+
+	List *fragmentList = PartitionTasklistResults(resultIdPrefix, selectTaskList,
+												  targetRelation, binaryFormat);
+	return ColocateFragmentsWithRelation(fragmentList, targetRelation);
+}
 
 
 /*
@@ -64,6 +125,14 @@ PartitionTasklistResults(char *resultIdPrefix, List *selectTaskList,
 						 DistTableCacheEntry *targetRelation,
 						 bool binaryFormat)
 {
+	if (targetRelation->partitionMethod != DISTRIBUTE_BY_HASH &&
+		targetRelation->partitionMethod != DISTRIBUTE_BY_RANGE)
+	{
+		ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+						errmsg("repartitioning results of a tasklist is only supported "
+							   "when target relation is hash or range partitioned.")));
+	}
+
 	/*
 	 * Make sure that this transaction has a distributed transaction ID.
 	 *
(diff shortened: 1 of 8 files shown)


src/backend/distributed/executor/distributed_intermediate_results.c
```diff
@@ -29,6 +30,29 @@
...
+
+/*
+ * NodeToNodeFragmentsTransfer contains all fragments that need to be fetched from
+ * the source node to the destination node in the NodePair.
+ */
+typedef struct NodeToNodeFragmentsTransfer
+{
+	NodePair nodes;
+	List *fragmentList;
+	uint64 anchorShardId;
```
pykello: do we need this?

src/backend/distributed/executor/distributed_intermediate_results.c
```diff
@@ -45,6 +69,39 @@ static DistributedResultFragment * TupleToDistributedResultFragment(
...
+														   DistTableCacheEntry *
+														   targetRelation);
+static List * FragmentTransferTaskList(List *fragmentListTransfers);
+static char * QueryStringForFragmentsTransfer(
+	NodeToNodeFragmentsTransfer *fragmentsTransfer);
+static uint64 ExecuteFetchTaskList(List *fetchTaskList);
+
+
+DistributedResult *
+RedistributeTaskListResults(char *resultIdPrefix, List *selectTaskList,
```
pykello: comment

src/test/regress/expected/distributed_intermediate_results.out
```diff
@@ -150,10 +220,111 @@ SELECT count(*), sum(x) FROM
...
+CALL public.create_range_partitioned_shards('colocated_with_target', '{"(0,a)","(25,a)"}','{"(24,z)","(49,z)"}');
+-- one value per shard, so we can route calls to read_intermediate_shards
+INSERT INTO colocated_with_target VALUES ((0,'a')), ((25, 'a'));
+BEGIN;
+CREATE TABLE distributed_result_info AS
+  SELECT * FROM redistribute_task_list_results('test', $$ SELECT mapped_key, value FROM source_table $$, 'target_table');
+SELECT * FROM distributed_result_info ORDER BY shardid;
+ shardid |                                       colocated_results                                       
+---------+-----------------------------------------------------------------------------------------------
+ 4213602 | {test_from_4213600_to_0,test_from_4213600_to_0,test_from_4213601_to_0,test_from_4213601_to_0}
```
pykello: 1. this shouldn't have duplicates
2. log which transfers happened to make sure we handled replication correctly.

src/backend/distributed/transaction/transaction_management.c
```diff
@@ -130,7 +130,16 @@ UseCoordinatedTransaction(void)
 
 	CurrentCoordinatedTransactionState = COORD_TRANS_STARTED;
 
-	AssignDistributedTransactionId();
+	/*
+	 * This might be part of bigger distributed transaction originating from
+	 * another node, in which case transaction id has already been assigned
+	 * by a assign_distributed_transaction_id() call.
```
marcocitus: I wonder if we should have a new value for `CurrentCoordinatedTransactionState` to reflect that the current transaction is part of a coordinated transaction, but did not start it.

src/backend/distributed/executor/distributed_intermediate_results.c
```diff
@@ -318,3 +379,227 @@ ExecuteSelectTasksIntoTupleStore(List *taskList, TupleDesc resultDescriptor)
...
+		appendStringInfoString(fragmentNamesArrayString,
+							   quote_literal_cstr(fragmentName));
+
+		fragmentCount++;
+	}
+
+	appendStringInfoString(fragmentNamesArrayString, "]::text[]");
+
+	appendStringInfo(queryString,
+					 "SELECT fetch_intermediate_results(%s,%s,%d) bytes",
```
marcocitus: I think doing `SELECT * FROM fetch_intermediate_results` is a little nicer. If we add multiple return values in the future and some workers are ahead of the coordinator, then it won't break this query.

src/backend/distributed/executor/distributed_intermediate_results.c
```diff
@@ -318,3 +379,227 @@ ExecuteSelectTasksIntoTupleStore(List *taskList, TupleDesc resultDescriptor)
...
+	resultStore = ExecuteSelectTasksIntoTupleStore(taskList, resultDescriptor);
+
+	TupleTableSlot *slot = MakeSingleTupleTableSlotCompat(resultDescriptor,
+														  &TTSOpsMinimalTuple);
+
+	uint64 byteCount = 0;
+	while (tuplestore_gettupleslot(resultStore, true, false, slot))
+	{
+		bool isNull = false;
+		byteCount += DatumGetInt64(slot_getattr(slot, 1, &isNull));
```
marcocitus: Are we ever planning to use the return value of this function?

src/backend/distributed/executor/distributed_intermediate_results.c
```diff
@@ -318,3 +379,227 @@ ExecuteSelectTasksIntoTupleStore(List *taskList, TupleDesc resultDescriptor)
...
+	transferHashInfo.entrysize = sizeof(NodeToNodeFragmentsTransfer);
+	transferHashInfo.hcxt = CurrentMemoryContext;
+	HTAB *transferHash = hash_create("Fragment Transfer Hash", 32, &transferHashInfo,
+									 HASH_ELEM | HASH_CONTEXT | HASH_BLOBS);
+
+	ListCell *fragmentCell = NULL;
+	foreach(fragmentCell, fragmentList)
+	{
+		DistributedResultFragment *fragment = lfirst(fragmentCell);
+		List *placementList = ShardPlacementList(fragment->targetShardId);
```
marcocitus: Would it make sense to use FinalizedShardPlacementList?

src/backend/distributed/executor/distributed_intermediate_results.c
```diff
@@ -318,3 +379,227 @@ ExecuteSelectTasksIntoTupleStore(List *taskList, TupleDesc resultDescriptor)
...
+ * targetRelation->sortedShardIntervalArray[shardIndex] after fetch tasks are
+ * done.
+ */
+static List **
+ColocateFragmentsWithRelation(List *fragmentList, DistTableCacheEntry *targetRelation)
+{
+	List *fragmentListTransfers = ColocationTransfers(fragmentList, targetRelation);
+	List *fragmentTransferTaskList = FragmentTransferTaskList(fragmentListTransfers);
+
+	ExecuteFetchTaskList(fragmentTransferTaskList);
```
marcocitus: There is a question of whether we want/need to be able to tolerate failure on the write path. In that case, we may want to have fetch tasks with multiple placements, but that seems complex and requires us to have different queries for different placements.

My inclination would be to opt for simplicity and require all target placements to be available during the insert..select.

src/backend/distributed/executor/distributed_intermediate_results.c
```diff
@@ -318,3 +379,227 @@ ExecuteSelectTasksIntoTupleStore(List *taskList, TupleDesc resultDescriptor)
...
+	transferHashInfo.hcxt = CurrentMemoryContext;
+	HTAB *transferHash = hash_create("Fragment Transfer Hash", 32, &transferHashInfo,
+									 HASH_ELEM | HASH_CONTEXT | HASH_BLOBS);
+
+	ListCell *fragmentCell = NULL;
+	foreach(fragmentCell, fragmentList)
+	{
+		DistributedResultFragment *fragment = lfirst(fragmentCell);
+		List *placementList = ShardPlacementList(fragment->targetShardId);
+
```
marcocitus: Do we want to do any additional pruning based on rowCount? I suppose it's somewhat hypothetical.

src/backend/distributed/executor/distributed_intermediate_results.c
pykello: > There is a question of whether we want/need to be able to tolerate failure on the write path. In that case, we may want to have fetch tasks with multiple placements, but that seems complex and requires us to have different queries for different placements.
> 
> My inclination would be to opt for simplicity and require all target placements to be available during the insert..select (keeping it the way it is now).

going to look into this today and tomorrow to see if there is a simple solution.

src/backend/distributed/executor/distributed_intermediate_results.c
pykello: > Optional: Do we want to do any additional pruning based on rowCount?

Do you mean when the rowCount == 0? The query in `WrapTasksForPartitioning()` prunes out `rows_written = 0`.

src/backend/distributed/transaction/transaction_management.c
pykello: > I wonder if we should have a new value for `CurrentCoordinatedTransactionState` to reflect that the current transaction is part of a coordinated transaction, but did not start it.

Added `COORD_TRANS_STARTED_ON_WORKER`.

src/backend/distributed/executor/distributed_intermediate_results.c
pykello: done

src/backend/distributed/executor/distributed_intermediate_results.c
pykello: I don't think so. It returns void now.

pykello:
> I get a crash when the target table is a reference table:
> 
> ```
> SELECT * FROM redistribute_task_list_results('foo', $$ SELECT y+1,x FROM test $$, 'ref');
> server closed the connection unexpectedly
>         This probably means the server terminated abnormally
>         before or while processing the request.
> The connection to the server was lost. Attempting reset: Failed.
> Time: 537.323 ms
> ```

Added a check to disallow append/hash partitioned tables for now. I think we can support this in future though.

src/backend/distributed/executor/distributed_intermediate_results.c
pykello: done

PR #3356, opened 2020-01-06
(demo) multi row insert not going through

pykello:
In the multi_row_insert.sql added by this PR, the multi-row INSERT doesn't take effect. It also causes `ensure_no_intermediate_data_leak` to fail, which might be a clue.


serprex:
Updated this branch with a fix

